Issue #7877, opened 2015-03-25
add property yii\widgets\ActiveField::renderContent = null


lynicidn:
here https://github.com/yiisoft/yii2/blob/master/framework/widgets/ActiveField.php#L156 we do render without args - i propose add new property for active field with `null` value as default.
Bonus: we can configure default behavior for generate `label`, `hint` and `input` of `template` `parts`

``` php
'fieldConfig' => [
    'renderContent' => function($field) {
        if (!isset($field->parts['{input}'])) {
            $field->parts['{input}'] = myLogicForGenerateInputAsDefault($field->model, $field->attribute, $field->inputOptions);
        }
        return strtr($field->template, $field->parts);
    }
]
```

now it need do for every attribute:

``` php
$form->field($model, 'attr')->inputText()->render(function($field){/*...*/});
//or
$form->field($model, 'attr')->render(function($field){/*...*/});
```

cebe:
why would you need this?
We have template https://github.com/yiisoft/yii2/blob/master/framework/widgets/ActiveField.php#L55
and parts https://github.com/yiisoft/yii2/blob/master/framework/widgets/ActiveField.php#L144
which lets you control everything already.

lynicidn:
@cebe i want change default behavior for generate input part of template in extension. Without it i should extend yii\bootstrap\ActiveField and yii\widgets\ActiveField
with this changes i can write extension unknowing what user is use

lynicidn:
@cebe also with it i can set function only one place in code

lynicidn:
problem in bootstrap extension - i can't write extensions for form or write for concrete bootstrap or yii\widgets

lynicidn:
reopen it please and mark as `under discussion`

lynicidn:
may be it help:
```php
use myVendor\form\ActiveField;

/**
 * @group form
 */
class ActiveFieldTest extends \yii\codeception\TestCase
{
    use \Codeception\Specify;

    public $appConfig = '@tests/codeception/config/console/unit.php';
    /**
     * @var \UnitTester
     */
    protected $tester;

    public $htmlOptions = ['class' => 'html'];
    public $label = 'label';
    public $labelOptions = ['class' => 'label'];
    public $hint = 'hint';
    public $hintOptions = ['class' => 'hint'];

    /**
     * @var \yii\widgets\ActiveForm
     */
    protected $form;
    /**
     * @var \yii\base\Model
     */
    protected $model;
    /**
[...]

PR #7897, opened 2015-03-27
Update ActiveField.php

lynicidn:
#7877


cebe:
This is outdated, lets find a solution in #7877 before implementing something.

PR #7898, opened 2015-03-27
Fix #7896: Intelligent and more correct `Html::addCssClass`

### LINKED ISSUE #7896: Intelligent and more correct Html `addCssClass`
I will be submitting a PR for improving and making the `Html::addCssClass` more intelligent and more practical to use for all use cases - similar or better than jquery addClass.

## Enhancements:

- Parse the original `$options['class']` correctly irrespective of whitespaces  (tabs, newlines or spaces)
- Parse the `$class` param correctly irrespective of whitespaces  (tabs, newlines or spaces)
- If the `$class` property is separated by spaces... it will validate each class separately already added and 
- Intelligently do not append css class if it already exists irrespective of the whitespace

### Example
```php
$options = ['class' => '    kartik']; // note the whitespace
Html::addCssClass($options, ' hide    kartik  robot');
Html::addCssClass($options, "\tkartik   hide fancy    box     holy");
```
### Output
```php
// generated output
$options['class'] =  '    kartik  hide    kartik  robot         kartik   hide fancy    box     holy';
```

### Example
```php
$options = ['class' => '    kartik']; // note the whitespace
Html::addCssClass($options, ' hide    kartik  robot');
Html::addCssClass($options, "\tkartik   hide fancy    box     holy");
```
### Output
```php
// generated output
$options['class'] =  'kartik hide robot fancy box holy';
```

kartik-v:
Enhanced parsing of whitespaces in both `options['class']` and the `$class` param to  `Html::addCssClass` method.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -16,6 +16,7 @@ Yii Framework 2 Change Log
 - Enh #7636: `yii\web\Session::getHasSessionId()` uses a more lenient way to check if session ID is provided in URL (robsch)
 - Enh #7850: Added `yii\filters\PageCache::cacheCookies` and `cacheHeaders` to allow selectively caching cookies and HTTP headers (qiangxue)
 - Enh: Added `yii\helper\Console::wrapText()` method to wrap indented text by console window width and used it in `yii help` command (cebe)
+- Enh #7896: Intelligent and more correct Html::addCssClass to allow more correct parsing of whitespaces (kartikv)
 - Chg: Updated dependency to `cebe/markdown` to version `1.1.x` (cebe)
 
 
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1715,7 +1715,7 @@ class BaseHtml
         if (isset($options['class'])) {
             $options['class'] = trim(preg_replace('/\s+/', ' ', $options['class']));
             $classes = ' ' . $options['class'] . ' ';
-            if (strpos( $classes, ' ' . $class . ' ') === false) {
+            if (strpos($classes, ' ' . $class . ' ') === false) {
                 $options['class'] .= ' ' . $class;
             }
         } else {
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1698,22 +1698,35 @@ class BaseHtml
      * Adds a CSS class (or a set of space / tab separated classes) to the specified options.
      * If the CSS class already exists in the options, it will not be added again.
      * @param array $options the options to be modified.
-     * @param string $class the CSS class(es) to be added. You can add multiple
-     * classes separated with a whitespace (space / tab).
+     * @param string|array $class the CSS class(es) to be added. You can add multiple
+     * classes either as an array OR separated with a whitespace (space / tab).
+     * @param bool whether to clean and sanitize whitespaces in the input. If set to true,
+     * the method will replace all whitespaces (tab, space, newlines) to single space and trim
+     * whitespaces from beginning and end of the generated options['class'].
      * @author Kartik Visweswaran <kartikv2@gmail.com>
      */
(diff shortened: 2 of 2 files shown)


qiangxue:
Here are my opinions:

* The original goal of the method is to add a single css class quickly
* The new code all of a sudden introduces several regex calls
* Not quite sure if we really need to handle TABs
* Perhaps supporting array of new css classes is more practical and efficient?

kartik-v:
> The original goal of the method is to add a single css class quickly

Yes understand. But that can be a limitation as the use cases can be often when one wants to add more than CSS classes in the code.

> The new code all of a sudden introduces several regex calls

Yes. But not all will be parsed unless needed. There will be an initial regex check to see if a whitespace exists (you could use strpos here as well) - else the functionality as in original implementation proceeds. A couple of regex is for input sanitization to check for any whitespace (i.e. space, tab, line feed, new line). If we do not want to make it intelligent enough to parse HTML like a browser would, then these sanitizations can be removed. 
From a performance perspective there will be similar impact when adding more than one CSS class. Current implementation will anyway involve calling this method N number of times.

> Not quite sure if we really need to handle TABs

That's the reason regex was introduced... for any whitespace parsing. If we want to be less strict and force only space parsing, then this check can be removed as mentioned previously.

> Perhaps supporting array of new css classes is more practical and efficient?

True - if we do not have any whitespace parsing to be done. The current implementation was to also maintain consistency with `Html::addCssStyle` that supports string separated css styles.

Will modify the PR and send by simplifying with an array (it will not involve

kartik-v:
I pushed in an update and here are the changes:

- the addCssClass method now accepts `class` param as either a string or an array
- regex parsing is minimized unless a third param `clean` is specified as true.
- the only additional validation is that the method once checks for a whitespace and then proceeds to split the classes else it is exactly same to the original implementation

Example Tests:

```php
$options = ['class' => '    kartik'];

//
// TEST 1 (clean = true)
//
echo Html::addCssClass($options, ' hide kartik robot', true);
//
// TEST 1 OUTPUT =   'kartik hide robot'
//

//
// TEST 2 (clean = true) - extending above test 1 options
//
echo Html::addCssClass($options, ' kartik   hide fancy    box     holy');
//
// TEST 2 OUTPUT =   'kartik hide fancy box holy'
//

// TEST 3 pass class params as array and clean = false
$options = ['class' => 'kartik'];
echo Html::addCssClass($options, ['hide', 'kartik',  'robot']);
//
// TEST 3 OUTPUT =   'kartik hide robot'
//

// TEST 1a variation of 1 with class param as string with whitespaces and clean = false
$options = ['class' => '    kartik'];
echo Html::addCssClass($options, ' hide kartik robot');
//
// TEST 1a OUTPUT =   '    kartik hide robot'
//

// TEST 2a variation of 2 with class param as string with whitespaces and clean = false
echo Html::addCssClass($options, '  kartik   hide fancy    box     holy');
//
// TEST 2a OUTPUT =   '    kartik hide fancy box holy'
//
```

qiangxue:
> However, note that even if the input had ONLY spaces - we would hit errors in parsing CSS classes with the original implementation - when we have more than ONE space / extra spaces in between or in the beginning/end. In these cases we may end up having duplicate CSS classes appended.

This is not true. The original implementation has considered this situation and can work correctly.

If we only support whitespace separator, allow single class as string and multiple classes as array, then we don't need any regex.

kartik-v:
Yes, true the error would happen is if a tab character was pressed somehow in the begin or in between classes.

Is the new approach acceptable to have the CSS sent as either string or array or space separated? Much like `Html::addCssStyle` which allows both string or an array method to add styles.

qiangxue:
I think 99% of the time users won't use tabs or newlines to separate css classes. We can rely on documentation to explicitly specify that we only handle whitespace-separated css classes.

The nested `addCssClass()` call at line 1722 should be rewritten as it is causing calling the regex method twice unnecessarily.

PR #7910, opened 2015-03-29
Include ruleTable in list of customizable tables [ci skip]

smichae:
With the addition of ruleTable in 2.0, including it in the list of customizable tables in the class-level doc makes it easier to spot the difference from 1.1.


cebe:
thanks!

PR #7912, opened 2015-03-30
Add aria-hidden="true" => yii\grid\ActionColumn

lav45:
http://getbootstrap.com/components/#accessible-icons

### DIFF
--- a/framework/grid/ActionColumn.php
+++ b/framework/grid/ActionColumn.php
@@ -110,28 +110,31 @@ class ActionColumn extends Column
     {
         if (!isset($this->buttons['view'])) {
             $this->buttons['view'] = function ($url, $model, $key) {
-                return Html::a('<span class="glyphicon glyphicon-eye-open" aria-hidden="true"></span>', $url, array_merge([
+                $options = array_merge([
                     'title' => Yii::t('yii', 'View'),
                     'data-pjax' => '0',
-                ], $this->buttonOptions));
+                ], $this->buttonOptions);
+                return Html::a('<span class="glyphicon glyphicon-eye-open" aria-label="' . $options['title'] . '"></span>', $url, $options);
             };
         }
         if (!isset($this->buttons['update'])) {
             $this->buttons['update'] = function ($url, $model, $key) {
-                return Html::a('<span class="glyphicon glyphicon-pencil" aria-hidden="true"></span>', $url, array_merge([
+                $options = array_merge([
                     'title' => Yii::t('yii', 'Update'),
                     'data-pjax' => '0',
-                ], $this->buttonOptions));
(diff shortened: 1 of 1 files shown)


framework/grid/ActionColumn.php
```diff
@@ -110,7 +110,7 @@ protected function initDefaultButtons()
     {
         if (!isset($this->buttons['view'])) {
             $this->buttons['view'] = function ($url, $model, $key) {
-                return Html::a('<span class="glyphicon glyphicon-eye-open"></span>', $url, array_merge([
+                return Html::a('<span class="glyphicon glyphicon-eye-open" aria-hidden="true"></span>', $url, array_merge([
```
samdark: Won't it completely hide the button from screen readers?

cebe:
@LAV45 you should read the full explaination of what you linked. For this case, the following applies instead: 

> If you're creating controls with no other text (such as a <button> that only contains an icon), you should always provide alternative content to identify the purpose of the control, so that it will make sense to users of assistive technologies. In this case, you could add an aria-label attribute on the control itself.

LAV45:
Excuse me, didn't read :)

cebe:
can you apply the correct changes?

samdark:
Should be aria-label.

samdark:

samdark:
@LAV45 it's better to use the same email for commits as you're using for GitHub so commit and profile will be linked together.

framework/grid/ActionColumn.php
```diff
@@ -110,28 +110,31 @@ protected function initDefaultButtons()
     {
         if (!isset($this->buttons['view'])) {
             $this->buttons['view'] = function ($url, $model, $key) {
-                return Html::a('<span class="glyphicon glyphicon-eye-open"></span>', $url, array_merge([
+                $options = array_merge([
                     'title' => Yii::t('yii', 'View'),
                     'data-pjax' => '0',
-                ], $this->buttonOptions));
+                ], $this->buttonOptions);
+                return Html::a('<span class="glyphicon glyphicon-eye-open" aria-label="' . $options['title'] . '"></span>', $url, $options);
```
qiangxue: `$options['title']` should be HTML-encoded. Similar thing for the code below.

LAV45:
View, Update and Delete - it's a simple word in which no characters need to be escaped using HTML-encode.

cebe:
it is translated by Yii::t and thus may be anything.

samdark:
What was merged is *adjusted* version ;)

cebe:
good point :-) But shouldn't the label be on the span rather than the link?

samdark:
Logically it should be on the link.

PR #7919, opened 2015-03-30
fix oci\Schema::findConstraints() and pgsql\Schema::findTableNames()

nineinchnick:
Also added missing view to oci tests.

@samdark I left some bugs when rebasing my recent PRs, sorry about that. Should have run the tests.


samdark:
Thanks :)

PR #7921, opened 2015-03-30
Valid function name in example of optimisticLock [skip ci]

bethrezen:
Replace activeHiddenField with activeHiddenInput in code example


cebe:
Thanks!

PR #7922, opened 2015-03-30
Little timesaver tip for optimisticLock use [skip ci]

bethrezen:
Those who follow this 4 simple steps may forget about need to validate 'version' field. Let us save their time.


qiangxue:
Thanks!

PR #7925, opened 2015-03-31
added ability to disable formSelector and linkSelector

### LINKED ISSUE #7918: Pjax widget suggestion
Suggestion one: If we using pjax widget in the loop - there are too many useless js code. For exapmle, we need handle only one click on the button per iteration. Is this possible to make $linkSelector and $formSelector oprional? In the case above we should disable $formSelector and it would reduce the generated code twice.

Suggestion second: it would be great to define ability to have only one handler for the same operation. For example, we have list of the items and there is only one button in the each item. I want update only item's block when i click to the button. But in this case i should define the same handler for each item's button.

usualdesigner:
fixes #7918

### DIFF
--- a/framework/widgets/Pjax.php
+++ b/framework/widgets/Pjax.php
@@ -55,11 +55,19 @@ class Pjax extends Widget
      */
     public $linkSelector;
     /**
+     * @var boolean whether to enable auto generated link selector.
+     */
+    public $enableLinkSelector = true;
+    /**
      * @var string the jQuery selector of the forms whose submissions should trigger pjax requests.
      * If not set, all forms with `data-pjax` attribute within the enclosed content of Pjax will trigger pjax requests.
      * Note that if the response to the pjax request is a full page, a normal request will be sent again.
      */
     public $formSelector;
+    /**
+     * @var boolean whether to enable auto generated form selector.
+     */
+    public $enableFormSelector = true;
     /**
      * @var boolean whether to enable push state.
      */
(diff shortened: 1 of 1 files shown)


Alex-Code:
Instead of introducing new properties I would have just allowed setting `formSelector` and `linkSelector` to `false`.

usualdesigner:
Core team, what do you think? Should we introduce new properties or just use the exist?

framework/widgets/Pjax.php
```diff
@@ -55,12 +55,20 @@ class Pjax extends Widget
      */
     public $linkSelector;
     /**
+     * @var boolean whether to enable auto generated link selector.
```
samdark: It's not generating just a selector. It's pjax handler for all links matching selector.

framework/widgets/Pjax.php
```diff
@@ -55,12 +55,20 @@ class Pjax extends Widget
...
+    /**
      * @var string the jQuery selector of the forms whose submissions should trigger pjax requests.
      * If not set, all forms with `data-pjax` attribute within the enclosed content of Pjax will trigger pjax requests.
      * Note that if the response to the pjax request is a full page, a normal request will be sent again.
      */
     public $formSelector;
     /**
+     * @var boolean whether to enable auto generated form selector.
+     */
+    public $enableFormSelector = true;
```
samdark: Same here.

cebe:
yes, should be as @Alex-Code said. This is common usage in other places.

samdark:
Fixed with adjusted version https://github.com/yiisoft/yii2/commit/36a0e8384a88697b7a3e64d3269a5b70c775bddd

Thanks!

PR #7937, opened 2015-03-31
PK UNSIGNED

thiagotalma:
As with auto-incrementing primary keys will never be negative, let's define as "unsigned", which doubles the maximum number.


pana1990:
#1032 here is the yii2 position about it. I didn't agreed at first but I agree now

thiagotalma:
Ok! Let's close

PR #7943, opened 2015-04-01
BaseListView render <tag>

lav45:
In the current implementation it is not possible to use correctly [panels-tables](http://getbootstrap.com/components/#panels-tables) because by default, the table is wrapped in a <div> which disrupts the structure of classes [.panel>.table](https://github.com/twbs/bootstrap/blob/master/less/panels.less#L104)

I added the ability to insert arbitrary html code in the block in which the table is positioned

``` php
<?php GridView::begin([
    'options' => [
        'class' => 'panel panel-default'
    ],
    'columns' => [  ],
]); ?>

<!-- Default panel contents -->
<div class="panel-heading">Panel heading</div>
<div class="panel-body">
    <p>Some default panel content here.</p>
</div>

<?php GridView::end() ?>
```

### DIFF
--- a/framework/widgets/BaseListView.php
+++ b/framework/widgets/BaseListView.php
@@ -93,6 +93,10 @@ abstract class BaseListView extends Widget
      */
     public $layout = "{summary}\n{items}\n{pager}";
 
+    /**
+     * @var string html tag which is list view
+     */
+    private $_tag;
 
     /**
      * Renders the data models.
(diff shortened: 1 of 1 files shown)


LAV45:
Can be closed, problem solved
```php
<div class="panel panel-default">
    <!-- Default panel contents -->
    <div class="panel-heading">Panel heading</div>
    <div class="panel-body">
        <p>Some default panel content here.</p>
    </div>

    <?php GridView::widget([
        'options' => [
            'class' => 'table-responsive'
        ],
    ]); ?>

</div>
```

PR #7945, opened 2015-04-01
Change fields and extraFields

shcherbanich:
Adding getter and setter for fields and extraFields.

Бывает необходимо устанавливать на лету fields и extraFields. С подходом, который был ранее 

```
public function fields()
    {
        return ['field1', 'field2', 'field3'];
    }
```

это не возможно, поля определялись один раз. Хочется, чтобы можно было делать это на лету.

С данным изменением можно использовать как прежний подход, так и ряд новых, с условием, что метот fields() не будет переопределен. Например, можно делать так:

```
public function init(){

    $this->setFields(['field1', 'field2', 'field3']);

}

public function setUserFields(){
    $this->setFields(['field1', 'field2']);
}


public function setAdminFields(){
    $this->setFields(['field1', 'field2', 'field3', 'field4']);
}

public function getCurrentFields(){
    return $this->getFields();   // or $this->fields();
}
```

или поля можно определять так: 

```
    protected fields = [
        'field1',
        'field2'
    ];
```

Прошу рассмотреть данные изменения, они могут быть весьма полезны

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -90,7 +90,37 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      * @var array related models indexed by the relation names
      */
     private $_related = [];
+    
+    protected $fields = [];
+    
+    protected $extraFields = [];
+    
 
+    public function getExtraFields(){
+        
+        return $this->extraFields;
+    }
+    
+    public function setExtraFields(array $fields){
+        
+        $this->extraFields = $fields;
+        
+        return $this;
+    }
+    
+    
+    public function getFields(){
+        
+        return $this->fields;
+    }
+    
+    
+    public function setFields(array $fields){
+        
+        $this->fields = $fields;
+        
+        return $this;
+    }
 
     /**
      * @inheritdoc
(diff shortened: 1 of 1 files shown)


samdark:
@shcherbanich your English seems to be OK. Please use it so core team members other than me and @klimov-paul undestand you correctly.

klimov-paul:
Methods `fields()` and `extraFields()` are public and you may implement they as you pleased. if you wish them to be configurable you may do so in your particular class.
Sinse `fields()` and `extraFields()` are declared by `Arrayable` interface and regard to it, I don't think setters for them must exist anywhere at the core.
You may compose your change into a trait, which can be added to any of your Active Record classes or even to any other class implementing `Arrayable` interface. In this way you will avoid core duplication if your need this feature.

shcherbanich:
@klimov-paul, why not to make this default behavior? In the remaining places there are getters and setters, and you should not write a wrapper

klimov-paul:
For example: this change will have no effect on some common cases like `Serializer::serializeDataProvider()`:
https://github.com/yiisoft/yii2/blob/master/framework/rest/Serializer.php#L162

Here default 'fields' and 'extraFields' will be used anyway.

Methods of `Arrayable` interface are ment to serve as 'static' (although this is not necessary), and the whole interface is treated around the framework in this way.

klimov-paul:
Also for the model it could be more common to vary 'fields' and 'extraFields' depending on the scenario:
```php
public function extraFields()
{
    if ($this->scenario == 'rest') {
        return ['slaveModels'];
    } else {
        return [];
    }
}
```
So such change definetely can not be considered as generic.
You may still compose it into an extension using trait approach as I recommended.

PR #7957, opened 2015-04-01
do not extra convert to floats in CompareValidator client code

CthulhuDen:
See https://github.com/yiisoft/yii2/blob/1d32f4d269765d75a413ea026b22922f81990f71/framework/assets/yii.validation.js#L277, conversion is already done if needed. For me this bug stops from validating a date (comparing two dates in php:Y-m-d format).


qiangxue:
Thanks!

PR #7990, opened 2015-04-05
Update structure-assets.md

maksimchuk:


samdark:
It's OK as is. ts = TypeScript.

PR #7998, opened 2015-04-06
Update db-active-record.md

pixlmonstr:
Fixed syntax error in PHP example

### DIFF
--- a/docs/guide/db-active-record.md
+++ b/docs/guide/db-active-record.md
@@ -667,7 +667,7 @@ public function actionUpdate($id)
 
     try {
         if ($model->load(Yii::$app->request->post()) && $model->save()) {
-            return $this->redirect(['view', $urlParams]);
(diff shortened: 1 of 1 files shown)


docs/guide/db-active-record.md
```diff
@@ -667,7 +667,7 @@ public function actionUpdate($id)
 
     try {
         if ($model->load(Yii::$app->request->post()) && $model->save()) {
-            return $this->redirect(['view', <?= $urlParams ?>]);
+            return $this->redirect(['view', $urlParams]);
```
cebe: should be something like `['id' => $model->primaryKey]` instead of `$urlParams` because this variable is not defined.

pixlmonstr: Yes, you are right. I just starred on the failed syntax highlighting.

qiangxue:
Thanks!

Issue #4899, opened 2014-09-01
Default ODM mapper for MongoDB?


docsolver:
The documentation of the `yii2-mongodb` extension states:

> This extension does not provide any special way to work with embedded documents (sub-documents). General recommendation is avoiding it if possible

I understand how including an Object Document Mapper for embedded documents would be very complex, but to me its absence defeats the whole purpose of using MongoDB with ActiveRecord. After all, due to its NoSQL-nature, MongoDB relies heavily on embedded documents as an alternative to relational joins. I was very excited to see MongoDB support out-of-the-box in Yii2, but I believe that this omission will severely hinder Yii2+MongoDB adoption for most people.

Instead of building a native solution, there are some existing tools that might be plugged in to do this job. It seems that the most important ones are [Doctrine Mongodb ODM](http://www.doctrine-project.org/projects/mongodb-odm.html) and [Mandango](http://mandango.org/).

I am eager to adopt MongoDB, since it would solve some of the constraints I am facing with a relational approach. Unfortunately my experience with MongoDB and hacking the internals of Yii2 is insufficient to build a decent solution myself.

However, I assume that I am not the only one facing this problem, so maybe someone else has already successfully combined one of these tools (or something similar) with Yii2 ActiveRecord to enable MongoDB-based ActiveRecord models with embedded models as subdocuments? If so, maybe this approach could be included in the `yii2-mongodb` extension or in the documentation?

alexandernst:
@cebe I'm considering spending some time on this and implement it, but I'm not really sure where to start nor how exactly to do it. Do you have any guide/plans for this that you could share with me?

cebe:
Well, for elasticsearch all sub documents are already represented as PHP arrays because the communication is in json and we simply json_encode/decode the values so there is nothing to be implemented additionally. For mongodb I have no idea what has to be done to implement this. How is the data of sub-documents retrieved currently?

klimov-paul:
For the MongoDB is it the same.

Sammaye:
One thing that is being forgotten her is validation, manipulation, and most importantly (for performance) population of said subdocuments.

All of these points, population being most significant, is one reason why I dropped making this myself and made a pesonalised suboducment loader specifically designed for my needs.

alexandernst:
@Sammaye Care to explain a little further the problem? (maybe with actual pseudo-code examples). Maybe we all can get to a solution (or throw away this entire bug if a solution is not possible).

Sammaye:
[...]

PR #8002, opened 2015-04-07
Allow indirect modifications of overloaded properties when using MongoDB

alexandernst:
Take for example this data structure saved in a MongoDB database:

```
_id: '3456345b6serbts34sdfbtsdf',
type: [
    "admin",
    "client"
],
personal_info: {
    name: 'Bob',
    surnames: ['', '', '', ''....],
    bday: ISODate(....),
},
credentials: {
    email: 'qqq@www.com',
    pass: '123123',
    salt: '2345c2345v345wq'
}
```

Getting the entire document is easy, `$person = Person::find(["_id" => "....."]);`

But modifying any subdocument values is not possible:

```
var_dump($person->credentials);  // <--- works
var_dump($person->credentials['password']); // <--- works
$person->credentials["password"] == "new password"; // fails with "Indirect modification of overloaded property"
```

This patch fixes this and, as far as I'm aware of, it doesn't introduce any bugs or any API breaks.

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -233,7 +233,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      * @return mixed property value
      * @see getAttribute()
      */
-    public function __get($name)
+    public function &__get($name)
     {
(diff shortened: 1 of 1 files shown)


klimov-paul:
See explanation at #4899: such change will cause more trouble then it solves.

PR #8005, opened 2015-04-07
Added grid\DetailView

hiqsol:
[[yii\grid\DetailView]] is similar to [[yii\widgets\DetailView]] but uses [[Column]]
to show header and data part of given [[Model]].

Helps reuse your code by defining [[Column]]s once and reusing them for
both GridView and DetailView, like this:

``` php
$columns = [
    [
        'class' => SomeMyColumn::actionClass(),
    ],
    [
        'class' => ActionColumn::actionClass(),
    ],
];

GridView::widget([
    'filterModel'  => $searchModel,
    'columns'      => $columns,
]);

DetailView::widget([
    'model'        => $model,
    'columns'      => $columns,
]);
```

### DIFF
--- a/framework/grid/ColumnViewTrait.php
+++ b/framework/grid/ColumnViewTrait.php
@@ -1,0 +1,167 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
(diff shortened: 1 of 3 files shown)


qiangxue:
Thank you for your contribution. This is better maintained as a user extension rather than a core one. The main reason is this will introduce a lot of confusion with the existing `yii\widgets\DetailView`.

Issue #3376, opened 2014-05-06
Values array validation (from multiple select)


RomeroMsk:
I have model Order with attribute status_id. This attribute must have only one of this values: [1, 2, 3]. I want to make search model and form with **multiple** select tag, which correspond to status_id attribute. What validation rule can I use for status_id to allow multiple values (to call **Order::find()->andFilterWhere(['in', 'status_id', $this->status_id])** in search model) and validate each of them against an array [1, 2, 3]?
I think this is common usecase, so if Yii doesn't have an out-of-box solution for this, it needs an enhancement.

tecnologiaterabyte:
I have the same problem described, which was the best found solution, since I have searched and read and I can not find solution to the issue, I opened the question in the forum and facebook and no answer, I would like to be able to establish this enhancement, or another proposal would allow filtering of rows by '\yii\grid\CheckboxColumn'.
![gridview-03](https://cloud.githubusercontent.com/assets/4957021/5962725/0ed2da44-a7b8-11e4-9f58-c325e4c85603.jpg)

lynicidn:
i solve via `rules`
```php
['name', 'exist', 'allowArray' => true, 'when' => function ($model, $attribute) {return is_array($model->$attribute);}],
```
and in `search` if string do explode by comma and trim and use `$query->andFilterWhere(['name' => $name])` else `$query->andFilterWhere(['like', 'name', $this->name])`

Notice: need do implode by comma to string before render value into input filter

tecnologiaterabyte:
thank you very much for the reply, I commented that the extension because besides making a multiple filter in the gridview, I can also filter the checkboxcolumn to delete selected records, download selected records.

View:

    <?= ActionBar::widget([
        'grid' => 'kv-grid',
        'templates' => [
            '{bulk-actions}' => ['class' => 'col-xs-4'],
            '{filter}' => ['class' => 'col-xs-8 text-right'],
        ],
        'bulkActionsItems' => [
            'General' => [
                'general-filter' => 'Filter Rows Select',
            ],
        ],
        'bulkActionsOptions' => [
            'options' => [
                'general-filter' => [
                    'url' => \yii\helpers\Url::to(),
                ],
            ],
        ],
    ]) ?>

Controller:

    public function actionIndex()
    {
        $ids = Yii::$app->request->post('ids');
        if ($ids) {
            Yii::$app->request->setQueryParams([
                'HorariosSearch' => ['id' => implode(', ', $ids)],
            ]);
[...]

PR #8023, opened 2015-04-09
Fix for #3376: `yii\validators\EachValidator`

klimov-paul:
Fix for #3376: Added `yii\validators\EachValidator`, which allows validation of the array attributes

### DIFF
--- a/framework/validators/EachValidator.php
+++ b/framework/validators/EachValidator.php
@@ -30,7 +30,9 @@ use yii\base\Model;
  * }
  * ~~~
  *
- * Note: this validator will not work with validation declared via model inline method.
+ * Note: this validator will not work with validation declared via model inline method. If you declare inline
+ * validation rule for attribute, you should avoid usage of this validator and iterate over array attribute
+ * values manually inside your code.
  *
  * @property Validator $validator related validator instance. This property is read only.
  *
--- a/framework/validators/EachValidator.php
+++ b/framework/validators/EachValidator.php
@@ -1,0 +1,148 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\validators;
+
+use yii\base\InvalidConfigException;
+use Yii;
+use yii\base\Model;
+
+/**
+ * EachValidator serves validation of the array attributes.
+ * It perform validation of each array element using any other validator specified by [[rule]].
+ *
+ * ~~~php
+ * class MyModel extends Model
+ * {
+ *     public $arrayAttribute = [];
+ *
+ *     public function rules()
+ *     {
+ *         return [
+ *             ['arrayAttribute', 'each', 'rule' => ['trim']],
+ *             ['arrayAttribute', 'each', 'rule' => ['integer']],
+ *         ]
+ *     }
+ * }
+ * ~~~
+ *
+ * Note: this validator will not work with validation declared via model inline method.
+ *
+ * @property Validator $validator related validator instance. This property is read only.
+ *
+ * @author Paul Klimov <klimov.paul@gmail.com>
+ * @since 2.0.4
+ */
+class EachValidator extends Validator
+{
+    /**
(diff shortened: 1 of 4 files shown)


framework/validators/EachValidator.php
```diff
@@ -0,0 +1,148 @@
...
+ *     {
+ *         return [
+ *             ['arrayAttribute', 'each', 'rule' => ['trim']],
+ *             ['arrayAttribute', 'each', 'rule' => ['integer']],
+ *         ]
+ *     }
+ * }
+ * ~~~
+ *
+ * Note: this validator will not work with validation declared via model inline method.
```
samdark: Why not?

klimov-paul: It can not be created.
Inline validator from model method created here:
https://github.com/yiisoft/yii2/blob/master/framework/validators/Validator.php#L189

While creating slave validator I simply don't have an owner model instance at hand, so I have to pass a dummy model instead.
So if someone pass inline method name into a `rule` it will simply fail with message 'unable to find class "myInlineValidator"'.

samdark: That's technical reason which could be solved. Inability to use inline method in this case doesn't sound good overall.

klimov-paul: Why?
If you write your validation rule as model inline method, you may as well handle attribute array value inside it.
It does not make much sense to use such complex manipulation as `EachValidator` to actually do so.
Also, note, that `Closure` based validation will work just fine.

samdark: Right. Then suggestion to handle arrays yourself should be added to docs.

klimov-paul: Agreed. Doc comments have been updated.

samdark:
Now it looks good to me. @yiisoft/core-developers opinions? Comments?

framework/validators/EachValidator.php
```diff
@@ -0,0 +1,150 @@
...
+     *
+     * Please refer to [[yii\base\Model::rules()]] for more details.
+     */
+    public $rule;
+    /**
+     * @var boolean whether to use error message composed by validator declared via [[rule]] if its validation fails.
+     * If enabled, error message specified for this validator itself will appear only if attribute value is not an array.
+     * If disabled, own error message value will be used always.
+     */
+    public $allowMessageFromRule = true;
```
qiangxue: Perhaps `useRuleMessage` because `allow` means you may or may not use the message from the rule, although it allows you to do so.

qiangxue:
Looks good to me. Could you also add this rule to the guide `input-validation.md`? Thanks.

framework/validators/EachValidator.php
klimov-paul: If the value is not an array: message from the 'rule' validator will not be used. Instead own message will be displayed.
So I think 'allow' is more appropriate.

framework/validators/EachValidator.php
```diff
@@ -0,0 +1,150 @@
...
+
+
+    /**
+     * @inheritdoc
+     */
+    public function init()
+    {
+        parent::init();
+        if ($this->message === null) {
+            $this->message = $this->allowMessageFromRule ? Yii::t('yii', '{attribute} should be an array.') : Yii::t('yii', '{attribute} is invalid.');
```
qiangxue: Why the message varies according to the value of `allowMessageFromRule`?

klimov-paul: Validator may produce 2 different error messages:
1) if value is not an array
2) if value does not pass slave validation

If developer chooses `allowMessageFromRule = true` he whould expect see message 'attribute should be an array' at first case instead of some generic 'attribute is invalid'. This seems logical considering slave validator may produce message like 'attribute should be an integer'.
However if developer chooses `allowMessageFromRule = false` more generic message should be used to fit both 1 and 2 cases.

qiangxue: I think this is a bit twisted. Perhaps we should simply use the generic message. This is also because for end users a message like `'{attribute} should be an array.'` doesn't really make sense (normal users have no idea what is an array).

klimov-paul: OK, agreed.

Issue #5042, opened 2014-09-16
Use RETURNING PostgreSQL syntax for inserts, instead of Connection::getLastInsertID


aotd1:
I have PostgreSql 9.3 table:
```sql
CREATE EXTENSION "uuid-ossp";
CREATE TABLE Favorites
(
    id UUID PRIMARY KEY DEFAULT uuid_generate_v1mc() NOT NULL,
    user_id INT NOT NULL,
    created_at TIMESTAMP DEFAULT now() NOT NULL,
    ...
);
```
And simple php code:
```php
    $f = new Favorite();
    $f->attributes = ['user_id' => 22599];
    $f->save();
    var_dump($f->attributes);
```
Returns:
```php
array(7) {
  'id' =>  NULL
  'user_id' =>  int(22599)
  'created_at' =>  NULL
  ...
}
```
Table has no sequences, so saved model has empty `id` property (look at [ActiveRecord::insertInternal](https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveRecord.php#L439) logic).

How I can fix that?
Best way as I think - rewrite ActiveRecord::insertInternal (or Command::insert) for PostgreSQL to use insert with returning statement:
```sql
INSERT INTO favorites (user_id) VALUES (22599) RETURNING id;
```
 
PHP: 5.5.15
Yii2:  2.0.0-beta, dev-master (d83d340)

EagleMoor:
It's very need!

samdark:
Do you want to try implementing it and sending a pull request?

nineinchnick:
I do.

EagleMoor:
If we use custom id name like "uuid" and if 32-bit string, RETURNING uuid don't give result.

Fix it we can by 

```php
beforeInsert() {
  $this->uuid = 'samdoasda-asdasmodasda-sadasda'; // generate value
}
```

But if id generate on db, it's not job.

nineinchnick:
@EagleMoor this is a different issue, please create a new issue or search existing ones for something related to that.

EagleMoor:
@nineinchnick why different? 

```sql
id UUID PRIMARY KEY DEFAULT uuid_generate_v1mc() NOT NULL,
```

id generated by function, they can create string pk (UUID type)
It's but, what after insert with no auto increment index, this index not return.

psrustik:
Issue also actual for Oracle database.
```sql
CREATE TABLE mytable (
[...]

PR #8030, opened 2015-04-09
Insert into returning

nineinchnick:
WIP, requesting a review.

Three DBMS support the `INSERT INTO RETURNING` statement but all three do it differently:
- PostgreSQL returns rows as a normal SELECT would
- Oracle requires binding params
- MSSQL got `OUTPUT` but I haven't looked into it

I was struggling to keep the code in ActiveRecord::insertInternal() dbms independent and I'm not entirely satisfied.

Also, the new params to `Command::insert()` could be moved to a new method, like `Command::insertReturning()` if it's going to help BC in any way.

I've left one `@todo` in the code, I don't know yet how to resolve mapping a column type to a PDO type constant.

Concerns #5042.


framework/db/ActiveRecord.php
```diff
@@ -455,18 +455,37 @@ protected function insertInternal($attributes = null)
...
         }
         $db = static::getDb();
-        $command = $db->createCommand()->insert($this->tableName(), $values);
-        if (!$command->execute()) {
-            return false;
+        // depending on the underlying DBMS, either $returnParams will be not empty, the command will return results
+        // or lastInsertId has to be used to fetch primary key values
+        $returnParams = [];
+        $command = $db->createCommand()->insert($this->tableName(), $values, $this->primaryKey(), $returnParams);
+        // force preparing as a command for writing, not reading
```
samdark: Why it's required to force it?

framework/db/Command.php
```diff
@@ -408,14 +408,27 @@ public function queryColumn()
      *
      * @param string $table the table that new rows will be inserted into.
      * @param array $columns the column data (name => value) to be inserted into the table.
+     * @param array $returnColumns the column names which values to be returned after inserting rows.
+     * @param array $returnParams if not empty after calling this method, each value is bound to the query.
      * @return Command the command object itself
+     * @throws NotSupportedException if $returnColumns is set but the underlying DBMS doesn't support returning
      */
-    public function insert($table, $columns)
+    public function insert($table, $columns, $returnColumns = null, &$returnParams = null)
```
samdark: It's public API and it's changed. Could it be avoided?

framework/db/QueryBuilder.php
```diff
@@ -151,8 +140,37 @@ public function insert($table, $columns, &$params)
...
+     *
+     * @param string $table the table that new rows will be inserted into.
+     * @param array $columns the column data (name => value) to be inserted into the table.
+     * @param array $params the binding parameters that will be generated by this method.
+     * They should be bound to the DB command later.
+     * @param array $returnColumns used only when if underlying DBMS support returning data from an INSERT statement
+     * @param array $returnParams used only when if underlying DBMS support returning data from an INSERT statement
+     * @return string the INSERT SQL
+     */
+    public function insert($table, $columns, &$params, $returnColumns = null, &$returnParams = null)
```
samdark: Same here. Public method signature changed.

samdark:
@yiisoft/core-developers this one needs review from everyone.

framework/db/ActiveRecord.php
nineinchnick: Because `queryOne` is called instead of `execute` which would prepare it as for reading and use slave connection.

framework/db/Command.php
nineinchnick: Yes, I've wrote about it in the PR description. A new `insertReturning` method could be created out of this. Is it a good idea?

samdark: Yes. This way we'll keep current signature so it will be backwards compatible and could be applied to 2.0.x instead of 2.1.x.

framework/db/QueryBuilder.php
```diff
@@ -151,13 +140,58 @@ public function insert($table, $columns, &$params)
...
+     * They should be bound to the DB command later.
+     * @param array $returnColumns used only when if underlying DBMS support returning data from an INSERT statement
+     * @param array $returnParams used only when if underlying DBMS support returning data from an INSERT statement
+     * They should be bound to the DB command later.
+     * @return string the INSERT SQL
+     */
+    public function insertReturning($table, $columns, &$params, $returnColumns = null, &$returnParams = null)
+    {
+        // clear returnParams
+        $returnParams = null;
```
nineinchnick: `insertReturning` behaves just like `insert` for DBMS not supporting this. It could throw an exception, but I wanted to avoid that because it would mean that AR would have to catch it each time it saves a record when using MySQL or SQLite, which is very common.

I'm not sure if that's the better solution or an exception should be thrown anyway.

framework/db/Command.php
```diff
@@ -419,6 +419,39 @@ public function insert($table, $columns)
...
+    {
+        $params = [];
+        $sql = $this->db->getQueryBuilder()->insertReturning($table, $columns, $params, $returnColumns, $returnParams);
+
+        $this->setSql($sql)->bindValues($params);
+
+        if ($returnParams !== null) {
+            $this->prepare(false);
+            foreach ($returnParams as $param => &$value) {
+                //! @todo type has to be determined from column schema
```
nineinchnick: Anybody got any idea how to implement this? Actually we could detect only integer/numeric and just use string for everything else, but maybe some valid `column type => pdo type` map could be created.

framework/db/oci/QueryBuilder.php
```diff
@@ -162,6 +162,27 @@ public function addForeignKey($name, $table, $columns, $refTable, $refColumns, $
     }
 
     /**
+     * For each return column a param will be added to $returnParams.
+     * They should have variables bound before executing the query.
+     * @inheritdoc
+     */
+    public function insertReturning($table, $columns, &$params, $returnColumns = null, &$returnParams = null)
+    {
+        $schema = $this->db->getSchema();
```
klimov-paul: Can't you just generate `$returnParams` inside this method?
As far as I can see it can be composed from column names.
This way method will not allow full flexibility of the Oracle, but will remain more generic.

nineinchnick: No, primary key usually isn't passed in $columns at all.

nineinchnick: I don't get it, I do generate `$returnParams` here.

klimov-paul: Is it necessary for `$returnParams` to be passed as method argument?
Can't you use column name as param binding name for these?
So its value will be:
```php
[
    ':id' => null
]
```
Then `queryOne()` will return row set just the same as Postgres.

We may check if param ':id' already present amoung query params (which is unlikely for the insert composed by query builder) and throw an exception on it or try to replace it.

nineinchnick: I was looking for a way to simplify this too.

First, now since I've moved this to a separate method we don't have to worry that `Command::insert()` calls `bindValues`.

But how a bound param (using `bindParam`) can be returned in `queryOne()`? I actually tried to run in PostgreSQL `INSERT INTO RETURNING col INTO :param` yesterday but it doesn't work as sql, it's only valid for plpgsql.

nineinchnick: Oh, you meant the other way, trying to do `queryOne` for Oracle. I didn't try that yet.

klimov-paul: Currently `$returnParams` is redundant: it make sense only for Oracle. It whould be nice if we can avoid it.
This may simplify code at `ActiveRecord` as well.

nineinchnick: I agree, I'm going to look into it more.

klimov-paul: Should be something like:
```php
public function insertReturning($table, $columns, &$params, $returnColumns = null)
    {
        $schema = $this->db->getSchema();
        $returning = [];
        if ($returnColumns !== null) {
            $returnParams = [];
            foreach ($returnColumns as $name) {
                $returnParams[':' . $name] = null;
                $returning[] = $name === '*' ? $name : $schema->quoteColumnName($name);
            }
        }
        return $this->insert($table, $columns, $params)
        . (empty($returning) ? '' : ' RETURNING ' . implode(', ', $returning) . ' INTO ' . implode(', ', array_keys($returnParams)));
    }
```

nineinchnick: I don't think it's possible to return bound params by any `query*` method.

There would have to be a private property in `Command` class to track and return such params instead of results. Is it a good idea to add them?

klimov-paul: I am unsure: this is the first time I see any results are fetched from SQL server using bound params.

nineinchnick:
Tests pass, it might me ready for merging but should be reviewed again.

framework/db/Command.php
```diff
@@ -90,6 +90,11 @@ class Command extends Component
      */
     private $_pendingParams = [];
     /**
+     * @var mixed INSERT/UPDATE return parameters to be bound to the current PDO statement
+     * or boolean true if return parameters are the query result.
```
samdark: Where's this `boolean true` is used?

framework/db/Command.php
```diff
@@ -268,6 +274,22 @@ protected function bindPendingParams()
...
+        if (is_array($this->_returningParams)) {
+            foreach ($this->_returningParams as $name => &$value) {
+                $value = array_merge([
+                    'value' => null,
+                    'dataType' => \PDO::PARAM_STR,
+                    'length' => null,
+                    'driverOptions' => null,
+                    'bound' => false,
+                ], $value);
+                if ($value['bound'] !== false) {
```
samdark: `=== true`?

framework/db/Command.php
nineinchnick: In `Command::queryInternal`. The comment here is rather unclear. I didn't want to introduce another argument to the method so I just used this as a special value.

There are other quirks, all the steps necessary to call it are not intuitive:

* create a command using insertReturning
* prepare for writing using prepare(false)
* don't just execute, use queryOne

framework/db/Command.php
```diff
@@ -419,6 +441,31 @@ public function insert($table, $columns)
...
+     *
+     * Note that the created command is not executed until [[execute()]] is called.
+     *
+     * @param string $table the table that new rows will be inserted into.
+     * @param array $columns the column data (name => value) to be inserted into the table.
+     * @param array $returnColumns the column names which values to be returned after inserting rows.
+     * @return Command the command object itself
+     * @throws NotSupportedException if $returnColumns is set but the underlying DBMS doesn't support returning
+     */
+    public function insertReturning($table, $columns, $returnColumns = null)
```
samdark: Need to add `@since 2.0.4`

framework/db/QueryBuilder.php
```diff
@@ -151,13 +140,56 @@ public function insert($table, $columns, &$params)
...
             . implode(', ', $placeholders) . ')';
     }
 
     /**
+     * Creates an INSERT INTO RETURNING SQL statement or plain INSERT if the underlying DBMS does not support it.
+     * @see insert()
+     *
+     * @param string $table the table that new rows will be inserted into.
+     * @param array $columns the column data (name => value) to be inserted into the table.
+     * @param array $params the binding parameters that will be generated by this method.
```
samdark: It's not clear what binding parameters are, why/how they should be bound to the DB command and where/when "later" is.

framework/db/QueryBuilder.php
```diff
@@ -151,13 +140,56 @@ public function insert($table, $columns, &$params)
...
+     * They should be bound to the DB command later.
+     * @param array $returnColumns used only when if underlying DBMS support returning data from an INSERT statement
```
samdark: I think description should contain info about what to pass in there.

framework/db/QueryBuilder.php
```diff
@@ -151,13 +140,56 @@ public function insert($table, $columns, &$params)
...
+     * @param array $returnParams used only when if underlying DBMS support returning data from an INSERT statement
+     * They should be bound to the DB command later.
```
samdark: Same as previous 2 comments.

framework/db/QueryBuilder.php
nineinchnick: This is the original comment. I did not write it.

framework/db/QueryBuilder.php
```diff
@@ -213,25 +245,14 @@ public function batchInsert($table, $columns, $rows)
...
-     *
+     * Creates assignment expression for the SET part of the UPDATE command.
      * @param string $table the table to be updated.
      * @param array $columns the column data (name => value) to be updated.
-     * @param array|string $condition the condition that will be put in the WHERE part. Please
-     * refer to [[Query::where()]] on how to specify condition.
      * @param array $params the binding parameters that will be modified by this method
      * so that they can be bound to the DB command later.
-     * @return string the UPDATE SQL
+     * @return array lines of columnName = param
```
samdark: lines of columnName = param → columnName => param pairs?

framework/db/oci/QueryBuilder.php
```diff
@@ -162,6 +162,41 @@ public function addForeignKey($name, $table, $columns, $refTable, $refColumns, $
     }
 
     /**
+     * For each return column a param will be added to $returnParams.
+     * They should have variables bound before executing the query.
+     * @inheritdoc
+     */
+    public function insertReturning($table, $columns, &$params, $returnColumns = null, &$returnParams = null)
```
samdark: Need to add `@since 2.0.4`

framework/db/QueryBuilder.php
samdark: It's a messy one anyway.

framework/db/Command.php
samdark: Type could be specified as `array|true`.

samdark: Comment could be `Either an array of return parameters to be bound to the current PDO statement
or a special value "true" if return parameters are the query result.`

framework/db/oci/QueryBuilder.php
nineinchnick: Added to parent, it should be included by `@inheritdoc`

samdark: OK.

framework/db/ActiveRecord.php
```diff
@@ -455,18 +455,28 @@ protected function insertInternal($attributes = null)
...
         }
         $db = static::getDb();
-        $command = $db->createCommand()->insert($this->tableName(), $values);
-        if (!$command->execute()) {
+        // depending on the underlying DBMS, the command may return results
+        // or lastInsertId has to be used to fetch primary key values
+        $command = $db->createCommand()->insertReturning($this->tableName(), $values, $this->primaryKey());
+        // force preparing as a command for writing, not reading
+        $command->prepare(false);
+        $primaryKeys = $command->queryOne();
```
qiangxue: Does mysql support INSERT returning? If not, what will this statement do? An extra query? Any error?

samdark:
Perfect. I have no more comments. @yiisoft/core-developers any issues with the current code?

framework/db/ActiveRecord.php
samdark: MySQL doesn't support the syntax itself. It has `last_insert_id` instead.

nineinchnick: That's why the base `QueryBuilder` returns a normal `INSERT` ignoring `returnColumns`.

If you execute an `INSERT` statement in MySQL and then try to fetch results from the PDO statement it throws a general error. There's a check in `Command::queryInternal` so it doesn't do that, making it behave like `execute`.

qiangxue: > There's a check in Command::queryInternal so it doesn't do that, making it behave like execute.

Could you point to me which LOC does this? It seems to me an exception will be thrown in this case.

framework/db/ActiveRecord.php
```diff
@@ -455,18 +455,28 @@ protected function insertInternal($attributes = null)
             }
         }
         $db = static::getDb();
-        $command = $db->createCommand()->insert($this->tableName(), $values);
-        if (!$command->execute()) {
+        // depending on the underlying DBMS, the command may return results
+        // or lastInsertId has to be used to fetch primary key values
+        $command = $db->createCommand()->insertReturning($this->tableName(), $values, $this->primaryKey());
+        // force preparing as a command for writing, not reading
+        $command->prepare(false);
```
qiangxue: Is this line really needed? `$forRead` defaults to null, which means the command will check the SQL to determine which it is a read SQL.

framework/db/ActiveRecord.php
nineinchnick: https://github.com/nineinchnick/yii2/blob/5042-insert-returning/framework/db/Command.php#L880

`Command::_returningParams` defaults to null, so `queryInternal` behavior doesn't change and `pgsql\QueryBuilder` sets it to `true` which also restores default behavior of `queryInternal`.

Default `insertReturning` sets it to an empty array, which makes `queryInternal` iterate over it, which does nothing (no fetching).

framework/db/ActiveRecord.php
nineinchnick: `Command::execute()` calls it with `false` and `Command::queryInternal()` calls it with `true`.

klimov-paul:
Changes look a little bit messy. Most of all due to the Oracle specifics with param binding.
Perhaps such functionality should not be provided by `QueryBuilder`, but made more abstract.
I whould like to see entire functionality of inserting new record and retrieving the primary key encapsulated into single function without exposing if it using 'returning' statement or not.
Something like:
```php
/**
 * @return array primary key value for new row.
 */
public function newRecord($table, array $fields);
```

In order to vary actual implementation such method can be placed inside `yii\db\Schema` class.

So general implementation will be:
```php
class Schema extends Object
{
    public function newRecord($table, array $fields)
    {
        $sql = $this->getQueryBuilder()->insert($table, $fields, $params);
        $this->db->createCommand($sql, $params)->execute();
        return $this->getLastInsertID();
    }
}
```

Then all schema specifics, like using 'returning', `queryOne()`, binding params and so on, will stay inside this method.

Inside the Active Record it will always be simple:
```php
$pks = $this->schema->newRecord();
```

nineinchnick:
I had the same feeling there should be a method to wrap it all up. But I think the new `insertReturning` is abstract enough. It's usage together with `queryOne()` is consistent across all databases.

klimov-paul:
Currently is not consistent across the all database. The most proving example is Oracle, which demands bound params to return the values.
Also note the way you need to process the results of calling this method: at some case you need to call `getLastInsertID()`, at some - use `queryOne()` and at some iterate of `$returnParams`.
This eliminates any 'generic' natute of the method.

nineinchnick:
@klimov-paul I was refering to `Command::insertReturning`, not `QueryBuilder::insertReturning`. And I'd say that its as consistent as `getLastInsertID()` gets.

You want me to make any specific changes?

klimov-paul:
I am afraid it is not.
`Command::insertReturning()` knows nothing about how the returning valu should be actually fetched: via `getLastInsertID()`, `queryOne()` or `$returnParams` - and this is a problem.
It only composes the SQL, but does handles any other differences.
Also, the fact `Command::insertReturning()` creates a normal `insert` statement for MySQL and SQLite does not add any consistency either. It should throw an `NotSupportedException` for these dbms to be conistent.
With the current implementation the whole idea of `QueryBuilder` abstraction does not actually work. You will need a special treatment for the created command depending on actual DB usage.
I whould rather use `Schema` or even `TableSchema` to provide solid method, which inserts record and return the PK value. This is of course will not provide the ability of actual composition of the 'returning' statement, but still will solve the #5042 issue.

nineinchnick:
> This is of course will not provide the ability of actual composition of the 'returning' statement

I agree that `insertReturning` should throw an exception if it's not supported. But when it is, `Command::insertReturning` usage is constistent. You call it, execute the command using `queryOne` and you always get the result.

I did not make it throw an exception in the first place to avoid always catching it in AR for MySQL. With your proposed wrapper it's no longer necessary.

Should I add the wrapper method now or should we first discuss its naming? I'd go for `insert`.

klimov-paul:
It is the best to open an another PR and so things from scratch there.
Then we will be able to compare 2 solutions.
`insert` should be fine, I think. We can disscuss the name later if needed.

nineinchnick:
I'm not going to make another PR.

qiangxue:
I like the latest version (with `Schema::insert()`) better. Perhaps we can get rid of the changes made to `Command` too?

nineinchnick:
@qiangxue I'd agree that `QueryBuilder::insertReturning` is messy, but check out `Command::insertReturning` usage in `pgsql\Schema::insert` and the same in `oci\Schema::insert`. What's wrong with that?

I'd rather remove/hide the interface in `QueryBuilder` and leave it in `Command` as is. But it's you guys decision finally.

klimov-paul:
Alternative implementation #8071

klimov-paul:
So, @yiisoft/core-developers, what is our decision here?

qiangxue:
@klimov-paul I like your PR better. It's consistent with the implementation of savePoint in Schema.

Issue #8032, opened 2015-04-10
rbac/PhpManager throws an exception when you change an item's name


ChristopheBrun:
- Enable PhpManager in your authentication config.
- Create some permissions in the GUI (I did it using the yii2-admin extension) then edit one of them and change its name. 
- Save the form. 

Your page shows the following error message : 
> Invalid Parameter – yii\base\InvalidParamException
Unable to change the item name. The name 'Delete Usersaaaa' is already used by another item.
1. in D:\Dev\PHP\yii2\advanced\vendor\yiisoft\yii2\rbac\PhpManager.php
[...]

The exception is thrown in the *updateItem()* method.

PR #8033, opened 2015-04-10
Fixes issue #8032

ChristopheBrun:
- Unit test added -> allows at least to reproduce the bug within the original project
- updated -> PhpManager::updateItem()
- composer.lock optionnaly added into the root .gitignore file


samdark:
Merged with some adjustments: https://github.com/yiisoft/yii2/commit/82033125c402a43cc17a77e9520054c8e42ad553

Thanks for the fix and especially for test verifying it.

qiangxue:
The change is incorrect - now you can't update the item whose name is not changed.

framework/rbac/PhpManager.php
```diff
@@ -567,15 +567,23 @@ protected function addRule($rule)
...
     /**
+     * @param string $name The name of the item being updated (aka : the 'old' name)
+     * @param Item $item The updated item
+     * @return bool
      * @inheritdoc
      */
     protected function updateItem($name, $item)
     {
+        // You can't overload an item with an existing name
+        if (isset($this->items[$item->name])) {
```
qiangxue: If the name is not changed, an exception will be thrown.

samdark: Right. It seems I haven't put enough attention. Fixing it.

PR #8034, opened 2015-04-11
Saving AR without values / calling Command::insert without columns

### LINKED ISSUE #7374: Inserting new ActiveRecord attempts to set primary key to NULL
When inserting a new ActiveRecord corresponding to a database table consisting only of one autoincremening primary key column, it attemps to set the primary key to NULL. This causes issues in postgresql. It should instead insert a row of DEFAULT VALUES.

nineinchnick:
When `Command::insert` is called without any column values defaults are used. Resolves #7374.

This PR should wait for #8030 to be accepted to make `QueryBuilder::insert()` easier to overload.


nineinchnick:
This is ready for adoption, any comments?

samdark:
@nineinchnick is it adjusted to deal with the fact that #8030 wasn't merged and alternative implementation was?

samdark:
Changelog is missing.

samdark:
Also I think this one fixes something...

nineinchnick:
@samdark yes, I think I've updated it after #8030 has been resolved.

samdark:
#7374

samdark:
Thanks!

PR #8044, opened 2015-04-11
force dot as decimal separator when casting float values of db columns to string

### LINKED ISSUE #7847: set_locale() and db/ColumnSchema decimal bug while using comma as a decimal separator with DECIMAL columns
We are currently working on a multilingual website with Yii2. Everything is great except that we are running into some database issues when using setlocale(LC_ALL, $language). 

The problems occur when we are using a locale that uses the comma as decimal seperator (nl_NL for instance). The problems do not occur when a country is used that has a point as decimal seperator. For example:

````php
// The BAD
setlocale(LC_ALL, 'nl_NL');

$model = new Model();
$model->decimal_column = 0.1; // DECIMAL(9,8)
$model->save();

$model->refresh();

var_dump($model->decimal_column); // 0.00000000
````

VS

````php
// The GOOD
setlocale(LC_ALL, 'en_US');

$model = new Model();
$model->decimal_column = 0.1; // DECIMAL(9,8)
$model->save();

$model->refresh();

var_dump($model->decimal_column); // 0.10000000
````

The problem seems to be that db/ColumnSchema->phpTypecast is using (string) to cast a DECIMAL MySQL column.

/vendor/yiisoft/yii2/db/ColumnSchema.php:97
````php
        // $this->type = 'decimal'
        // $this->phpType = 'string'

        switch ($this->phpType) {
            case 'resource':
            case 'string':
                return is_resource($value) ? $value : (string) $value;
            case 'integer':
                return (int) $value;
            case 'boolean':
                return (bool) $value;
            case 'double':
                return (double) $value;
        }
````

When the save is called on an activerecord, the db/ColumnSchema->phpTypecast parses a column that needs to be DECIMAL with (string). The value coming out of the STRING parse is directly used in the MySQL INSERT (or UPDATE). When 0.1 is parsed as string when en_US is the locale, the result is 0.1 - which MySQL processes corretly, but when nl_NL is used as the locale, the result is 0,1 - which MySQL processes as 0. So data is corrupted. See:

````php
setlocale(LC_ALL, 'en_US');
var_dump((string) 0.1); // 0.1

setlocale(LC_ALL, 'nl_NL');
var_dump((string) 0.1); // 0,1
````

Is this expected behaviour and shouldn't I use set_locale()? Or is this a bug that needs to be fixed.
PS. This is before the refactoring of ColumnSchema->setType()

nineinchnick:
Resolves #7847. The SQL standard requires all numeric to have dot as the decimal separator. All locales use only either a dot or a comma.

### DIFF
--- a/framework/db/ColumnSchema.php
+++ b/framework/db/ColumnSchema.php
@@ -122,7 +122,13 @@ class ColumnSchema extends Object
         switch ($this->phpType) {
             case 'resource':
             case 'string':
-                return is_resource($value) ? $value : (string) $value;
+                if (is_resource($value)) {
+                    return $value;
+                }
+                if (is_float($value)) {
+                    return str_replace(',', '.', (string)$value);
+                }
+                return (string)$value;
             case 'integer':
                 return (int) $value;
             case 'boolean':


samdark:
@nineinchnick are you sure all DBs we support are doing it according to the standard?

nineinchnick:
I'm actually writing a PR to improve loading column default values and I'm looking at the docs about constants:

* [Oracle literals](http://docs.oracle.com/cd/B19306_01/server.102/b14200/sql_elements003.htm)
* [PostgreSQL constants](http://www.postgresql.org/docs/9.1/static/sql-syntax-lexical.html)
* [MySQL literals](https://dev.mysql.com/doc/refman/5.0/en/literals.html)
* [SQL-T constants](https://msdn.microsoft.com/en-us/library/ms179899.aspx)
* [SQLite literals in expressions](https://www.sqlite.org/lang_expr.html)
* [CUBRID literals](http://www.cubrid.org/manual/821/en/Literal)

Seems all of them use dot as decimal separator.

samdark:
Merged manually: https://github.com/yiisoft/yii2/commit/698bfddeed478c98190f0603e93f0156c7e46953

samdark:
Thanks for taking care of it.

PR #8048, opened 2015-04-12
find mysql fk constraints from information_schema tables instead of 'show create table'

### LINKED ISSUE #6351: mysql\Schema::findConstraints matches foreign keys both in comments and MySQL syntax
During creation of migrate files for my existing database, I found it strange, that one of tables had two foreign key declarations, from Yii point of view, while there was only one in MySQL workbench. During debug process, I added following lines in `mysql\Schema::findConstraints` method (https://github.com/yiisoft/yii2/blob/master/framework/db/mysql/Schema.php#L250)
```
	        if ($table->name == 'user__oauth') {
		        var_dump($sql);
		        var_dump($matches);
		        die();
	        }
```
And I got following output
```
string 'CREATE TABLE `user__oauth` (
  `user_id` int(11) NOT NULL COMMENT 'CONSTRAINT FOREIGN KEY (user_id) REFERENCES user(id)',
  `uid` int(11) NOT NULL,
  `provider` varchar(100) CHARACTER SET utf8 NOT NULL,
  PRIMARY KEY (`user_id`,`uid`,`provider`),
  KEY `uid` (`uid`,`provider`),
  CONSTRAINT `user_id` FOREIGN KEY (`user_id`) REFERENCES `user` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci' (length=447)
array (size=2)
  0 => 
    array (size=4)
      0 => string 'FOREIGN KEY (user_id) REFERENCES user(id)' (length=41)
      1 => string 'user_id' (length=7)
      2 => string 'user' (length=4)
      3 => string 'id' (length=2)
  1 => 
    array (size=4)
      0 => string 'FOREIGN KEY (`user_id`) REFERENCES `user` (`id`)' (length=48)
      1 => string '`user_id`' (length=9)
      2 => string '`user`' (length=6)
      3 => string '`id`' (length=4)
```
It turns out, that the regex match (https://github.com/yiisoft/yii2/blob/master/framework/db/mysql/Schema.php#L249) finds either comment and foreign key in MySQL syntax, which is unexpected.

I am used to "old style" yii1, where I would add in MySQL table column comments foreign key declarations, to aid GII model generation process, and I don't know, if this behavior is feature or bug. Should I avoid adding "constraint" comments for tables, if I use foreign keys in MySQL?

I think, there should be some sort of algorithm that ignores stuff written in comments about constraints, if there exists foreign key declaration in MySQL syntax for that same field.

nineinchnick:
Resolves #6351.

### DIFF
--- a/framework/db/mysql/Schema.php
+++ b/framework/db/mysql/Schema.php
@@ -246,19 +246,32 @@ class Schema extends \yii\db\Schema
      */
     protected function findConstraints($table)
     {
-        $sql = $this->getCreateTableSql($table);
+        $sql = <<<SQL
+SELECT
+    kcu.constraint_name,
+    kcu.column_name,
+    kcu.referenced_table_name,
+    kcu.referenced_column_name
+FROM information_schema.referential_constraints AS rc
+JOIN information_schema.key_column_usage AS kcu ON
+    kcu.constraint_catalog = rc.constraint_catalog AND
+    kcu.constraint_schema = rc.constraint_schema AND
+    kcu.constraint_name = rc.constraint_name
+WHERE rc.table_name = :tableName
+SQL;
 
-        $regexp = '/FOREIGN KEY\s+\(([^\)]+)\)\s+REFERENCES\s+([^\(^\s]+)\s*\(([^\)]+)\)/mi';
-        if (preg_match_all($regexp, $sql, $matches, PREG_SET_ORDER)) {
-            foreach ($matches as $match) {
-                $fks = array_map('trim', explode(',', str_replace('`', '', $match[1])));
-                $pks = array_map('trim', explode(',', str_replace('`', '', $match[3])));
-                $constraint = [str_replace('`', '', $match[2])];
-                foreach ($fks as $k => $name) {
-                    $constraint[$name] = $pks[$k];
-                }
-                $table->foreignKeys[] = $constraint;
-            }
+        $rows = $this->db->createCommand($sql, [':tableName' => $table->name])->queryAll();
+        $constraints = [];
+        foreach ($rows as $row) {
+            $constraints[$row['constraint_name']]['referenced_table_name'] = $row['referenced_table_name'];
+            $constraints[$row['constraint_name']]['columns'][$row['column_name']] = $row['referenced_column_name'];
+        }
+        $table->foreignKeys = [];
+        foreach ($constraints as $constraint) {
+            $table->foreignKeys[] = array_merge(
+                [$constraint['referenced_table_name']],
+                $constraint['columns']
+            );
         }
     }
 


qiangxue:
Is there any mysql version compatibility issue using this approach?

nineinchnick:
I've ran it on MySQL 5.5 [using sqlfiddle](http://sqlfiddle.com/#!2/eb70d/2/0). MySQL 3.23.58 doesn't have those tables, but it's ancient. What's the oldest supported version?

nineinchnick:
This is ready for adoption, any comments?

cebe:
will check this and run tests on different versions.

nineinchnick:
This is related to fixes to mssql in #9773. Might be a good idea to move it to base Schema class.

nineinchnick:
Also fixes #8750.

samdark:
@nineinchnick any idea about MySQL 4 compatibility?

nineinchnick:
_INFORMATION_SCHEMA_ was added in MySQL 5.02.

nineinchnick:
I've wasted 1 hour on installing Debian 3.1 with MySQL 4.1 just to see that you can't even connect to anything older than 5.0 with default PHP mysql drivers.

samdark:
Oh. So that's not a problem. Great you've found it out.

samdark:
Merged https://github.com/yiisoft/yii2/commit/191185f5e301dbea93783870d3b650916c304719. Thank you!

dynasource:
This merge leads to a critical issue for anyone with MYSQL <5.1.10.

For me this commit breaks all of my Yii installations on my server with the following message:

```php
SQLSTATE[42S02]: Base table or view not found: 1109 Unknown table 'referential_constraints' in information_schema
The SQL being executed was: SELECT
kcu.constraint_name,
kcu.column_name,
kcu.referenced_table_name,
kcu.referenced_column_name
FROM information_schema.referential_constraints AS rc
JOIN information_schema.key_column_usage AS kcu ON
kcu.constraint_catalog = rc.constraint_catalog AND
kcu.constraint_schema = rc.constraint_schema AND
kcu.constraint_name = rc.constraint_name
WHERE rc.table_name = 'identity'
```

Analyzing further, the mysql table 'referential_constraints' is added from 5.1.10, (https://dev.mysql.com/doc/refman/5.1/en/referential-constraints-table.html) causing into a critical error for anyone having MYSQL <5.1.10. My server uses 5.0.67 and I guess I am not alone in this.

To conclude, either everybody has to upgrade with a new minimum Yii2 mysql requirement or this commit (unfortunately as it is) has to be reverted

samdark:
See comments in https://github.com/yiisoft/yii2/commit/191185f5e301dbea93783870d3b650916c304719

PR #8055, opened 2015-04-12
[REST] extraPatterns should take precedence over patterns

Agrumas:
According to docs extraPatterns should take precedence over patterns.

### DIFF
--- a/framework/rest/UrlRule.php
+++ b/framework/rest/UrlRule.php
@@ -166,7 +166,7 @@ class UrlRule extends CompositeUrlRule
     {
         $only = array_flip($this->only);
         $except = array_flip($this->except);
-        $patterns = array_merge($this->patterns, $this->extraPatterns);
+        $patterns = $this->extraPatterns + $this->patterns;
         $rules = [];
         foreach ($this->controller as $urlName => $controller) {
             $prefix = trim($this->prefix . '/' . $urlName, '/');
--- a/tests/framework/rest/UrlRuleTest.php
+++ b/tests/framework/rest/UrlRuleTest.php
@@ -1,0 +1,197 @@
+<?php
+
+namespace yiiunit\framework\rest;
+
+use yii\web\UrlManager;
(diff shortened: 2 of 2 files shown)


cebe:
the original code is correct, left most parameter of array_merge takes precedence.

Agrumas:
@cebe could you check again, because array_merge appends items to previous
ones.
2015 bal. 12 23:41 "Carsten Brandt" <notifications@github.com> rašė:

cebe:
http://php.net/manual/en/function.array-merge.php

see the example code, array2 takes precedence over array1 for key 'color'.

cebe:
Hm... I see what you mean now. There is a duplicate meaning of `precedence` here. It could mean:

- extraPatterns overwrite patterns or
- extraPatterns will be in the list of patterns before patterns

@qiangxue what is the purpose of having extraPatterns and how should this work?

Agrumas:
@cebe oh, i interpreted precedence as matching order.
I use MongoDB, so my id pattern accepts letters. When i try to access custom route (module/test) defined in extraPatterns, router invokes actionView with id test. I could make id regex more precise, but main issue still exits - i could not add higher priority route without redefining all rests rules.
So should extraPatterns cover this case or not?

cebe:
I am not sure, lets wait for @qiangxue for a decision.

Agrumas:
This covers both cases:
```php
$patterns = $this->extraPatterns + $this->patterns
```

qiangxue:
By "taking precedence", I mean patterns specified in `extraPatterns` will overwrite those in `patterns` if they are the same. This gives you chance to customize the route. So the original code was written for this purpose.

I understand your problem about specifying a higher priority rule via `extraPatterns`. I think your proposal `$this->extraPatterns + $this->patterns` should be fine. Would be great if you could add some unit test.

Agrumas:
@qiangxue @cebe here you go

qiangxue:
Thanks!

Issue #7883, opened 2015-03-26
yii\i18n\PhpMessageSource::loadMessages


yccphp:
log:

```
The message file for category 'yii' does not exist: /Applications/MAMP/htdocs/phpcto/trunk/application/admin/vendor/yiisoft/yii2/messages/zh/yii.php
```

```
web.php 
    'language' => 'zh_CN',

```

`What reason is this error?`

```
PhpMessageSource 

:72 line
protected function loadMessages($category, $language) 

// Why the truncation
$fallbackLanguage = substr($language, 0, 2);

zh_CN => zh 

message /  
zh 
 This folder is not useful

```

samdark:
It tries `zh_CN` then if it doesn't exist it tries to fallback to `zh`. Same for all languages: `en_US` -> `en`, `en_AU` => `en` etc. It's strange that it goes to fallback for you. Probably messages for `zh_CN` are a bit outdated.

yccphp:
zh_CN This file is present

lynicidn:
what is sourceLanguage u have? also zh?

cebe:
Note that the file in yii2\messages is `zh-CN` not `zh_CN`, maybe thats a problem? Guess we need a normalization for the locale. Same fix might need to be applied here: https://github.com/yiisoft/yii2-jui/blob/master/DatePicker.php#L152

yccphp:
@cebe  yes good !

samdark:
@cebe could you handle normalization?

samdark:
@cebe ICU locale could be quite precise such as `fr@collation=phonebook;calendar=islamic-civil`. Any suggestion on where to put method extracting language from it?

cebe:
`yii\i18n\I18N` would be a good place imo.

PR #8057, opened 2015-04-13
Implemented I18N::normalizeLocale() that noramlizes ICU or BCP47 locale into Yii translation language

samdark:
Part of #7883 fix.

TODO:
- [x] Better method name?
- [x] Introduce method for getting fallback language to take script into account.
- [x] Adjust `BaseFileHelper::localize()`.
- [x] Adjust JUI DatePicker asset (https://github.com/yiisoft/yii2-jui/pull/15)
- [ ] Adjust `DbMessageSource`.
- [ ] Adjust `PhpMessageSource`.
- [ ] Adjust `GettextMessageSource`.


framework/i18n/I18N.php
```diff
@@ -199,4 +199,86 @@ public function getMessageSource($category)
...
+     * - Language tag according to https://www.rfc-editor.org/bcp/bcp47.txt
+     * - ICU locale string http://userguide.icu-project.org/locale
+     *
+     * @return string normalized language ID
+     *
+     * @since 2.0.4
+     */
+    public function normalizeLocale($locale)
+    {
+        if (!class_exists('\Locale')) {
```
cebe: should not trigger autoload if it does not exist.

cebe:
Looks good to me and also fine with the name.

framework/i18n/I18N.php
```diff
@@ -199,4 +199,108 @@ public function getMessageSource($category)
...
+     * @param string $language language string. Could be either:
+     *
+     * - Language tag according to https://www.rfc-editor.org/bcp/bcp47.txt
+     * - ICU locale string http://userguide.icu-project.org/locale
+     *
+     * @return string normalized language ID
+     *
+     * @since 2.0.4
+     */
+    protected function fallbackNormalizeLocale($locale)
```
klimov-paul: Are you sure it should be 'protected'? Why not 'private'?

cebe: I think it should be private too, this is only a fallback if intl is not available which should not be exposed.

framework/i18n/I18N.php
```diff
@@ -199,4 +199,108 @@ public function getMessageSource($category)
...
+     * "zh-Hant-CN" ? "zh-CN" ? "zh"
+     *
+     * @param string $languageId normalized language ID. Either "zh-Hant-CN", "zh-CN" or "CN".
+     * @return string fallback language ID or null if there's no fallback
+     */
+    public function getFallbackLanguageId($languageId)
+    {
+        $parts = explode('-', $languageId);
+
+        switch(count($parts)) {
```
klimov-paul: Missing space after 'switch'

framework/i18n/I18N.php
```diff
@@ -199,4 +199,108 @@ public function getMessageSource($category)
 
         throw new InvalidConfigException("Unable to locate message source for category '$category'.");
     }
+
+    /**
+     * Normalizes language string to "zh-Hant-CN", "zh-CN" or "CN" format ready
+     * to be used to specify translation language.
+     *
+     * @param string $language language string. Could be either:
```
klimov-paul: Param name missmatch: '$language' at docs vs '$locale' at definition.

tests/framework/i18n/ExposedI18N.php
```diff
@@ -0,0 +1,22 @@
...
+ * @package yiiunit\framework\i18n
+ */
+class ExposedI18N extends I18N
+{
+    /**
+     * @inheritdoc
+     */
+    public function fallbackNormalizeLocale($locale)
+    {
+        return parent::fallbackNormalizeLocale($locale);
```
klimov-paul: Bypass via Reflection can be used instead.

framework/i18n/I18N.php
```diff
@@ -199,4 +199,108 @@ public function getMessageSource($category)
...
+     *
+     * - Language tag according to https://www.rfc-editor.org/bcp/bcp47.txt
+     * - ICU locale string http://userguide.icu-project.org/locale
+     *
+     * @return string normalized language ID
```
klimov-paul: Perhaps doc comment should use 'locale' term instead of 'language'?

klimov-paul:
> Adjust BaseFileHelper::localize().

Whould not this break a BC?
Developer may not stumble on the #7883 issue, while he might plac his localized files under 'en_us' instead of 'en-us'. Change will make them unavailbale.

tests/framework/i18n/ExposedI18N.php
samdark: How?

samdark:
Yes, that could break things in that case.

tests/framework/i18n/ExposedI18N.php
klimov-paul: See for example:
https://github.com/yiisoft/yii2-bootstrap/pull/31/files#diff-2f1760bba264cf4f4b0167dbbe1599e3R21

samdark:
Won't be able to finish it before 2.0.4. Moving to 2.0.5.

samdark:
Adjusted PR according to suggestions but not sure how to deal with BC in this case...

samdark:
I'm for adding it into upgrade notes and proceeding with merging it. Any objections?

cebe:
not sure, maybe it is time to start the 2.1 branch?

PowerGamer1:
@samdark 
https://github.com/yiisoft/yii2-jui/pull/15/files#diff-b21f181a007dbc219c5eca1f1e93d2d0R46

You normalize the language specified by user for the widget (`$this->language`) but not the language in the filename - where is the logic in this? In particular, your code won't find "/ui/i18n/datepicker-en_US.js" file if the user specifies "en_US" language for the widget.

samdark:
Won't be merged into 2.0.

PR #8069, opened 2015-04-13
Fix contaner tag is set to false

### LINKED ISSUE #8064: Bug with menu container when tag set to false
Hi.
I've got a custom menu extended from basic yii\widgets\Menu
In the widget we have $options array with tag option:

> * - tag: string, defaults to "ul", the tag name of the item container tags. Set to false to disable container tag.

Okay, then what we do in our new widget:

```
$this->options['tag'] = false;
```

But now we see "<>" string in page at the begining of menu widget, what's going on? Let's have a look in parent widget:

```
echo Html::tag($tag, $this->renderItems($items), $options);
```

No conditions about $tag is equal to false, so it renders  '<>' right before widget layout

kirsenn:
Do not add any empty tags when options['tag'] is set to false in Menu widget
Fixes #8064

### DIFF
--- a/framework/widgets/Menu.php
+++ b/framework/widgets/Menu.php
@@ -175,7 +175,12 @@ class Menu extends Widget
         if (!empty($items)) {
             $options = $this->options;
             $tag = ArrayHelper::remove($options, 'tag', 'ul');
-            echo Html::tag($tag, $this->renderItems($items), $options);
+            
+            if ($tag !== false) {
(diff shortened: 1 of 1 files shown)


samdark:
Added readme, tests, changelog, phpdoc comments and merged: https://github.com/yiisoft/yii2/commit/cafd135022716b50863a70efafb4d3e6cba05e6c

samdark:
Thanks!

PR #8071, opened 2015-04-13
Insert into returning alternative

klimov-paul:
Alternative implementation of #8030


framework/db/oci/Schema.php
```diff
@@ -422,4 +422,51 @@ protected function extractColumnSize($column, $dbType, $precision, $scale, $leng
...
+        }
+
+        $command = $this->db->createCommand($sql, $params);
+        $command->prepare(false);
+
+        foreach ($returnParams as $name => &$value) {
+            $command->pdoStatement->bindParam($name, $value['value']);
+        }
+
+        $command->queryOne();
```
nineinchnick: `execute()` should be called instead of `queryOne()`, which would make the `$command->prepare(false);` line above obsolete.

klimov-paul: What about further check `$command->pdoStatement->rowCount()`?
Should not it be different then?

klimov-paul:
Note: I am unable to run unit tests for Postgre or Oracle.
Someone should confirm this code is actually working.

framework/db/oci/Schema.php
nineinchnick: `execute()` returns `rowCount()` so you could use it.

klimov-paul: OK, done

framework/db/oci/Schema.php
```diff
@@ -422,4 +422,50 @@ protected function extractColumnSize($column, $dbType, $precision, $scale, $leng
...
+                $returning[] = $name === '*' ? $name : $this->quoteColumnName($name);
+            }
+            $sql .= ' RETURNING ' . implode(', ', $returning) . ' INTO ' . implode(', ', array_keys($returnParams));
+        }
+
+        $command = $this->db->createCommand($sql, $params);
+        $command->prepare(false);
+
+        foreach ($returnParams as $name => &$value) {
+            $command->pdoStatement->bindParam($name, $value['value']);
```
nineinchnick: This should be:
~~~php
$command->pdoStatement->bindParam($name, $value['value'], $value['dataType']);
~~~

framework/db/pgsql/Schema.php
```diff
@@ -450,4 +450,33 @@ protected function loadColumnSchema($info)
...
+     */
+    public function insert($table, $columns)
+    {
+        $params = [];
+        $sql = $this->db->getQueryBuilder()->insert($table, $columns, $params);
+        $returnColumns = $this->getTableSchema($table)->primaryKey;
+        if (!empty($returnColumns)) {
+            $returning = [];
+            foreach ((array)$returnColumns as $name) {
+                if ($name === '*') {
```
nineinchnick: This condition can be removed, a PK cannot be named *.

framework/db/pgsql/Schema.php
```diff
@@ -450,4 +450,33 @@ protected function loadColumnSchema($info)
...
+                    $returning[] = $name;
+                } elseif (preg_match('/^(.*?)(?i:\s+as\s+|\s+)([\w\-_\.]+)$/', $name, $matches)) {
```
nineinchnick: This also is invalid here, PKs can't have aliases.

framework/db/oci/Schema.php
```diff
@@ -422,4 +422,50 @@ protected function extractColumnSize($column, $dbType, $precision, $scale, $leng
...
+                $returnParams[$phName] = [
+                    'column' => $name,
+                    'value' => null,
+                ];
+                if (!isset($columnSchemas[$name]) || $columnSchemas[$name]->phpType !== 'integer') {
+                    $returnParams[$phName]['dataType'] = \PDO::PARAM_STR;
+                } else {
+                    $returnParams[$phName]['dataType'] = \PDO::PARAM_INT;
+                }
+                $returning[] = $name === '*' ? $name : $this->quoteColumnName($name);
```
nineinchnick: The `$name === '*'` condition should be removed.

framework/db/oci/Schema.php
```diff
@@ -422,4 +422,50 @@ protected function extractColumnSize($column, $dbType, $precision, $scale, $leng
...
+            $command->pdoStatement->bindParam($name, $value['value']);
+        }
+
+        if (!$command->execute()) {
+            return false;
+        }
+
+        $result = [];
+        foreach ($returnParams as $value) {
+            $result[$value['column']] = $value['value'];
```
nineinchnick: Should this include `$this->getTableSchema($table)->columns[$value['column']]->phpTypecast($value['value'])`? It was missing in my PR too.

framework/db/pgsql/Schema.php
```diff
@@ -450,4 +450,33 @@ protected function loadColumnSchema($info)
...
+                } else {
+                    $returning[] = $this->quoteColumnName($name);
+                }
+            }
+            $sql .= ' RETURNING ' . implode(', ', $returning);
+        }
+
+        $command = $this->db->createCommand($sql, $params);
+        $command->prepare(false);
+        $result = $command->queryOne();
```
nineinchnick: Same as in `oci`, should `$result` be casted using `phpTypecast()`?

framework/db/oci/Schema.php
klimov-paul: removed

framework/db/oci/Schema.php
klimov-paul: Fixed

framework/db/oci/Schema.php
klimov-paul: No: typecast code belongs to `ActiveRecord`.
I have restored it there.

framework/db/pgsql/Schema.php
klimov-paul: removed

framework/db/pgsql/Schema.php
klimov-paul: removed

framework/db/pgsql/Schema.php
klimov-paul: Typecase done at `ActiveRecord`

framework/db/ActiveRecord.php
```diff
@@ -454,20 +454,15 @@ protected function insertInternal($attributes = null)
...
-        if ($table->sequenceName !== null) {
-            foreach ($table->primaryKey as $name) {
-                if ($this->getAttribute($name) === null) {
-                    $id = $table->columns[$name]->phpTypecast($db->getLastInsertID($table->sequenceName));
-                    $this->setAttribute($name, $id);
-                    $values[$name] = $id;
-                    break;
-                }
+        foreach ($primaryKeys as $name => $value) {
+            if ($this->getAttribute($name) === null) {
```
nineinchnick: Just a question, why is this condition necessary and shouldn't it be `!isset($values[$name])`?

framework/db/Schema.php
```diff
@@ -399,6 +399,34 @@ public function setTransactionIsolationLevel($level)
...
+            return false;
+        }
+        $tableSchema = $this->getTableSchema($table);
+        if ($tableSchema->sequenceName === null) {
+            return [];
+        }
+        $result = [];
+        foreach ($tableSchema->primaryKey as $name) {
+            if ($tableSchema->columns[$name]->autoIncrement) {
+                $id = $tableSchema->columns[$name]->phpTypecast($this->db->getLastInsertID($tableSchema->sequenceName));
```
nineinchnick: Since you added the typecast to AR it should be removed here.

framework/db/ActiveRecord.php
klimov-paul: To be honest I am unsure. This is original implemetation.
I assume this check cuts off not incremental primary keys, which should be filled up manually before saving the record.

framework/db/Schema.php
klimov-paul: Removed

framework/db/ActiveRecord.php
nineinchnick: I've improved that check when moving code from AR to `Schema::insert` in my PR. Previously it assigned last insert id to first null PK column. I've changed it to assign it to first auto incremented PK column.

klimov-paul: Well, I can not decide such change on my own.

klimov-paul:
@nineinchnick, can you confirm the unit tests are passing for Postres and Oracle?

nineinchnick:
@klimov-paul  I only ran AR tests but they do pass. I can run the rest tomorrow.

framework/db/Schema.php
```diff
@@ -399,6 +399,33 @@ public function setTransactionIsolationLevel($level)
...
+     */
+    public function insert($table, $columns)
+    {
+        $command = $this->db->createCommand()->insert($table, $columns);
+        if (!$command->execute()) {
+            return false;
+        }
+        $tableSchema = $this->getTableSchema($table);
+        if ($tableSchema->sequenceName === null) {
+            return [];
```
qiangxue: Why returning an empty array in this case? Shouldn't it return the primary key value(s) as described in the comment?

nineinchnick: I agree, when there's no `sequenceName` it's probable that `$columns` contain new PK values anyway.

This is also related to my comment about the null check in AR, see above.

klimov-paul:
> Why returning an empty array in this case? Shouldn't it return the primary key value(s) as described in the comment?

Fixed.

qiangxue:
:+1:

framework/db/ActiveRecord.php
```diff
@@ -454,21 +454,13 @@ protected function insertInternal($attributes = null)
...
-                    $id = $table->columns[$name]->phpTypecast($db->getLastInsertID($table->sequenceName));
-                    $this->setAttribute($name, $id);
-                    $values[$name] = $id;
-                    break;
-                }
-            }
+        foreach ($primaryKeys as $name => $value) {
+            $id = $this->getTableSchema()->columns[$name]->phpTypecast($value);
+            $this->setAttribute($name, $id);
+            $values[$name] = $id;
```
cebe: before, the attribute will only be overwritten if it was `null`, now we will overwrite it any time.

cebe: added a test: 91777ae seems to be fine...

PR #8099, opened 2015-04-16
Simplified tests directory structure a bit

samdark:


cebe:
what about the namespaces? this will also affect AR tests in extensions as they copy some classes from here.

klimov-paul:
> what about the namespaces? 

What do you mean? They are not affected anyhow.

> this will also affect AR tests in extensions as they copy some classes from here.

yes, those tests should be adjusted later.

klimov-paul:
Migrated to #8102

Issue #8104, opened 2015-04-16
Memcache Expire Regression


wilwade:
https://github.com/yiisoft/yii2/commit/fcf2d0a3f2e4371be6fe09d9aa435fd006bed07b#diff-f9f025d17dc839d144e378e0dda1414cL202

Memcached uses timestamp expiration, while Memcache uses seconds (this is only for set however. For add it uses expiration)

http://php.net/manual/en/memcache.set.php
http://php.net/manual/en/memcached.expiration.php

I'll get a pull request in

qiangxue:
PHP manual says "Expiration time of the item. If it's equal to zero, the item will never expire. You can also use **Unix timestamp** or a number of seconds starting from current time, ..."

PR #8105, opened 2015-04-16
Issue #8104 Fixed. Memcache set() expiration to seconds

wilwade:
Memcached uses timestamp expiration, while Memcache uses seconds (this is only for set() however. For add() it uses expiration which is correct in the code)

http://php.net/manual/en/memcache.set.php
http://php.net/manual/en/memcached.expiration.php

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.4 under development
 -----------------------
 
+- Bug #8104: Fixed regression that had set Memcache expiration to time plus seconds instead of just seconds (wilwade)
 - Bug #6642: Fixed the bug that using confirmation dialog via `data-confirm` in an `ActiveForm` may cause the dialog to appear twice (pana1990, qiangxue)
 - Bug #6871: Fixed the bug that using defaults and hostnames in URL rules may cause an out-of-range index issue (qiangxue)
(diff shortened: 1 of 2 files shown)


qiangxue:
PHP manual says it can take either UNIX timestamp or number of seconds. So the original code is code.

wilwade:
@qiangxue I see that now. That is odd however as it did not expire correctly. Perhaps I have a system clock off. I will look into it more.

longmayar:
```php
        $memcache = new Memcache;
        $memcache->connect('localhost', 11211);
        $tmp_object = new \stdClass();
        $tmp_object->str_attr = "1";
        $memcache->set('not-yii', $tmp_object , false, 7);
        var_dump($memcache->get('not-yii'));  //  <--- It expires after 7 sec

        $yiimemcache = Yii::$app->cache;
        $tmp_object = new \stdClass();
        $tmp_object->str_attr = "1";
        $yiimemcache->set('yii', $tmp_object, 7);
        var_dump($yiimemcache->get('yii')); // <--- It NOT expires after 7 sec

        die();
```
Solution:
```php
class MemCache {

     protected function setValue($key, $value, $duration)
    {
        $expire = $duration; // > 0 ? $duration + time() : 0;

        return $this->useMemcached ? $this->_cache->set($key, $value, $expire) : $this->_cache->set($key, $value, 0, $expire);
    }

}
```

It works fine.

PR #8119, opened 2015-04-18
Separate IdentityInterface in two interfaces basic and common that exten...

Shkarbatov:
Separate IdentityInterface in two interfaces BasicIdentityInterface and IdentityInterface that extends BasicIdentityInterface.

If we need to create simple authorization, without token, we don't need all methods from IdentityInterface like: findIdentityByAccessToken(), getAuthKey() and validateAuthKey().

Example Yii2 basic:
https://github.com/yiisoft/yii2-app-basic

Wee need to implement all methods from IdentityInterface, but we dont use them all.
https://github.com/yiisoft/yii2-app-basic/blob/master/models/User.php

So it turns redundancy.

In my pull request we will need implement only BasicIdentityInterface, so we don't need to implement redundancy methods.

### DIFF
--- a/docs/guide-ru/intro-upgrade-from-v1.md
+++ b/docs/guide-ru/intro-upgrade-from-v1.md
@@ -495,12 +495,12 @@ class MyBehavior extends Behavior
 ```
 
 
-User и IdentityInterface
+User и BasicIdentityInterface|IdentityInterface
 ---------------------------
 
 Класс `CWebUser` из версии 1.1 теперь заменён классом [[yii\web\User]]. Также больше не существует класса `CUserIdentity`.
-Вы должны реализовать интерфейс [[yii\web\IdentityInterface]], что гораздо проще. Пример реализации представлен в шаблоне
-приложения advanced.
+Вы должны реализовать интерфейс [[yii\web\BasicIdentityInterface]] либо [[yii\web\IdentityInterface]], что
+гораздо проще. Пример реализации представлен в шаблоне приложения advanced.
 
 Более подробная информация представлена в разделах «[Аутентификация](security-authentication.md)»,
 «[Авторизация](security-authorization.md)» и «[Шаблон приложения advanced](tutorial-advanced-app.md)».
--- a/docs/guide/intro-upgrade-from-v1.md
+++ b/docs/guide/intro-upgrade-from-v1.md
@@ -498,12 +498,13 @@ class MyBehavior extends Behavior
 ```
 
 
-User and IdentityInterface
+User and BasicIdentityInterface|IdentityInterface
 --------------------------
 
 The `CWebUser` class in 1.1 is now replaced by [[yii\web\User]], and there is no more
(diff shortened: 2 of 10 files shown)


Shkarbatov:
Do I need to write unit tests for new interface?

klimov-paul:
Such separation does not make sense and may confuse users.
For example: stripping `getAuthKey()` from identity will ruin `User::enableAutoLogin` feature.
No one forces you to implement all the methods, you can leave thier declarations blank if you are sure you don't need them.

Shkarbatov:
> No one forces you to implement all the methods, you can leave thier declarations blank if you are sure you don't need them.

Yes, but it's not pretty.
Why should I describe the methods that should be empty, if they are not needed?

> For example: stripping getAuthKey() from identity will ruin User::enableAutoLogin feature.

If User needed getAuthKey() he must implement IdentityInterface.

This is our conversation.
http://yiiframework.ru/forum/viewtopic.php?f=19&t=23474&p=140218#p140218

klimov-paul:
Those methods are needed for your Identity class to satisfy framework demands to it.
We can of course split `IdentityInterface` into 2, but it does not worth the effots. We will need to bestore extra checks around the core code to handle this situation, while it is not so common to have an identity stripped from `getAuthKey()`.

Shkarbatov:
> Those methods are needed for your Identity class to satisfy framework demands to it.

You are wrong. I write at the begining:
> Wee need to implement all methods from IdentityInterface, but we dont use them all.
> https://github.com/yiisoft/yii2-app-basic/blob/master/models/User.php

It was example from Yii2 basic, and for everyone who will needed simple authorization, not only for my own Identity.

> We will need to bestore extra checks around the core code to handle this situation
It`s true. But if we think so, we will never improve Yii2 =)

klimov-paul:
Interface splitting can be done until single interface contains only single method.
Take your `BasicIdentityInterface` for instance. I can imagine situation, when I don't need `getId()` method at all. Actually it is not necessary to get ID of the identity to be able identify the user. I may need just `findIdentity()`.

Also what about other interfaces we introduced? 
For example, lets take `yii\mail\MessageInterface`. It introduces `attach()` method:
https://github.com/yiisoft/yii2/blob/master/framework/mail/MessageInterface.php#L165
But I can create my own simple implementation of the mail message, which does not attaches any files, because I don't need it (as you say for your case). So should we not split `MessageInterface` into several ones?

`IdentityInterface` provides contact for the Identity in the term how framework introduces it and relies on it. Matching it garantees that any feature provided by core framework code will work for it.

Shkarbatov:
> I can imagine situation, when I don't need getId() method at all. 

I said only about common approach but you said about the special case.

> Also what about other interfaces we introduced? 
> So should we not split MessageInterface into several ones?

Yes we should, if it will needed in common approach.

I think you just don't want to do checks around the core code to handle this situation, as you said before.

llfm:
> Yes, but it's not pretty.

Simplicity over complexity.

PR #8121, opened 2015-04-19
Add HTML Imports to Yii2's AssetBundle system

navarr:
This pull request is a Feature Request for supporting HTML imports.  This is a browser technology that is currently being implemented and has available polyfills.

This PR adds HTML Import specific knowledge to asset bundles so that they may declare HTML imports.  This is especially useful for any platform that wishes to make use of [Polymer](http://polymer-project.org/)

[Status of native browser implementation of the feature](https://www.chromestatus.com/feature/5144752345317376)

---

Imports are last on the list so that the required webcomponents.js can be included before the imports.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -47,6 +47,7 @@ Yii Framework 2 Change Log
 - Enh: Implement batchInsert for oci (nineinchnick)
 - Enh: Detecting IntegrityException for oci (nineinchnick)
 - Enh: `yii\widgets\LinkPager::$firstPageLabel` and `yii\widgets\LinkPager::$lastPageLabel` now could be set to true in order to use page number as label (samdark)
+- Enh: #8121: Added ability to include HTML imports in AssetBundles using `yii\web\AssetBundle::$imports`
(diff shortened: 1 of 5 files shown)


samdark:
That's [Chrome-specific](http://caniuse.com/#search=HTML%20Imports) and as far as I understand, has no specification. Not sure if it will be ever implemented everywhere and if it's a good idea to have it in the framework core.

navarr:
It is currently Chrome specific in native implementation, however the use case I created this for works with all Evergreen browsers (Polymer)

There is a spec: http://w3c.github.io/webcomponents/spec/imports/

It is implemented in Firefox behind a flag and is under consideration by IE. 

It's also in Opera

navarr:
Been "under discussion" for 22 days.  Is anything else needed from me to advance this?

samdark:
I tend to discard the pull request since it's not generic enough. @yiisoft/core-developers opinions?

qiangxue:
> I tend to discard the pull request since it's not generic enough. @yiisoft/core-developers opinions?
 Show all checks

Agree.

PR #8126, opened 2015-04-19
Fixed empty($class->method()) style for 5.4

zhdanovartur:
Fixed empty($class->method()) style for 5.4, because this feature allowed for php5.5 or high.


samdark:
Thanks!

PR #8130, opened 2015-04-19
rest-quick-start.md : adding a simple <?php opening for UserController.php example and objectiving filename

gautiermichelin:
rest-quick-start.md : adding a simple <?php opening for UserController.php example and objectiving filename

As this is a beginner tuto, better to be simpler than simple.


klimov-paul:
No, such change does not make sense.

PR #8133, opened 2015-04-20
Added EVENT_INIT event to Component

is7116:


samdark:
How you're going to use it?

is7116:
Use case for me is to statically attach behaviors to any component. This could be helpfull for extending components from another module.
```
Event::on(SomeComponent::className(), Component::EVENT_INIT, function($event){
    $event->sender->attachBehavior('foo', FooBehavior::className());
});
```

unclead:
@samdark i think it can be useful when i need to attach behavior which will implement some logic after initialization.
For example i have a menu in main module and filter in rbac module which remove all disallowed items before rendering
For solving this i implement my own event, but if component had AFTER_INIT event i would used it

qiangxue:
This will have big performance hit.

samdark:
Yep. I tend to reject such addition...

klimov-paul:
Requested use case can be resolved using `DI`: `yii\di\Container::set()`.
Sinse introduction of proposed event will produce perfromance impact, issue rejected.

Issue #8139, opened 2015-04-20
ENH: Unsafe Validator


mdmunir:
Now, to exclude some attribute from mass assigned. We must override method `Model::scenarios()`. For me, its not simple and potentially mistake. E.g, we add new attribute and forgot to add it to `Model::scenarios()`, the attribute will never be validate.
For simplify, i hope we have unsafe validator. Such as yii1 has. http://www.yiiframework.com/doc/api/1.1/CUnsafeValidator

cebe:
we already have `SafeValidator` so we could add unsafe as well...

mdmunir:
Of course i can submit PR for this.

cebe:
sure, go ahead.

SilverFire:
https://github.com/yiisoft/yii2/pull/8145#issuecomment-158827726

cebe:
reopening as there is still discussion going on in the PR.

cebe:
A different solution is in #10234

SilverFire:
Fixed in https://github.com/yiisoft/yii2/commit/da82ff3fb1625597706d9da3c76ca5a0a1ee7ad5

PR #8145, opened 2015-04-21
Enh #8139: Added unsafe validator

mdmunir:
#8139

I hope i am doing right :D


mdmunir:
[ASK]
How to setting environtment for test. Can i use only one database? where i must setup?
After merge i got too many error when execute `phpunit`.
```
'PDOException' with message 'SQLSTATE[28000] [1045] Access denied for user 'travis'@'localhost' (using passwor
```

cebe:
you have to set up your database credentials for testing in the config file: https://github.com/yiisoft/yii2/blob/master/tests/data/config.php#L4-L14
the db must already exist in dbms.

framework/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ Yii Framework 2 Change Log
 - Enh #6043: Specification for 'class' and 'style' in array format added to `yii\helpers\Html` (klimov-paul)
 - Enh #7169: `yii\widgets\ActiveField` now uses corresponding methods for default parts rendering (klimov-paul)
 - Enh #8070: `yii\console\controllers\MessageController` now sorts created messages, even if there is no new one, while saving to PHP file (klimov-paul)
+- Enh #8139: Added unsafe validator (mdmunir)
```
AlexCollin: When???

AlexCollin:
Hello! Do you can make patch for this feature? Please...

SilverFire:
I'm not sure this implementation is good.
We do not have `!attribute` notation in framework yet. I guess this thing gives a way to crash the app by submitting `<input name="!attribute" value="haax">` for the scenario that is not safe.

Also the code should be reformatted according to Yii2 code style.
@samdark, @cebe your thoughts?

samdark:
I think we should not have unsafe validator at all. I can't find any use case for it.

arash26:
@samdark 
In this case unsafe validator required
```php
[
                'password_hash', 
                'filter', 
                'when' => function(){
                    return !empty($this->password);
                }, 
                'filter' => function($model){
                    return Yii::$app->security->generatePasswordHash($this->password);
                }
],
[
                'password_hash', 
                'unsafe', 
                'when' => function(){
                    return empty($this->password);
                },                
],

**Comment by samdark** [comment]
@arash26 hash generation should not be handled via validation.

**Comment by arash26** [comment]
@samdark 
password_hash is just an example.
based on documents $when is A PHP callable whose return value determines whether this validator should be applied or not.
in this case password_hash will be safe regardless of $when

**Comment by SilverFire** [comment]
@arash26 I agree with @samdark 
Password hashing is not a task of validators. I'd prefer to use behavior or event a simple event. You need to check, whether the `password` was affected in the model by `$model->isAttributeChanged('password')` and update the hash, if it was.

**Comment by arash26** [comment]
@SilverFire 
My point is when $when is false, attribute must not considered safe.

```php
public function rules()
    {
        return [            
            [
                'name', 
                'required',
                'when' => function(){
                    return $this->checkSomeFlags();
                }
            ]            
        ];
    }
```
"name" is safe even $this->checkSomeFlags() return false.

SilverFire:
@arash26 ok, I got it. This use case makes sense.
@samdark ?

samdark:
@SilverFire, @arash26 could you provide a real use case for it which can't be solved otherwise?

arash26:
I can not find any other use case for it which can't be solved otherwise, but it's logical when $when is false the whole rule must be ignored.
I spent hours to find out $when don't behave like $on.

samdark:
https://github.com/yiisoft/yii2/pull/8145#issuecomment-158671953 makes sense but how's `unsafe` validator helpful in this case?

samdark:
OK. https://github.com/yiisoft/yii2/pull/8145#issuecomment-158671953 makes sense.

arash26:
I don't vote for unsafe validator.
I suggest $when behavior correction.

samdark:
Yeah, that's much better suggestion. Do you want to take care of the code for it?

samdark:
@mdmunir what do you think about it?

Faryshta:
@arash26 I think you are confusing 'safe' with 'validated'. Safe means it can be massively assigned using methods like `Model::load()`. An attributes is safe regardless of the value of `when` since the model doesn't contain data yet and its the purpose of the safes attributes to know which ones can be assigned and which ones can't.

After that a validation is performed when its decided if the value provided to your attribute is fills the expected format.

In your case you need to do

`['password_hash', 'required']`

mdmunir:
I am open any suggestion.
Back to topic. I just want to simple way to exclude some attribute from `safe`.
Currently, we must override method `scenarios()`.

Faryshta:
@mdmunir if you start the attribute name with `!` it will not be considered safe

`['!password_has', 'required']`

means the password is required but its not safe for autoloading.

mdmunir:
@Faryshta Oh... OK. 
If that so, case close for me.

arash26:
@Faryshta Thanks for your explanation.

mdmunir:
@SilverFire But, its not mension in doc.

SilverFire:
@mdmunir truth. Could you make another PR and describe the use case and solution in the guide, please?

mdmunir:
@SilverFire 
The @Faryshta says is not work as expected. I add this to unit test
```php
// tests/framework/base/ModelTest.php -> testSafeScenarios()

        $model = new RulesModel();
        $model->rules = [
            [['name','!email'], 'required'],
        ];
        $this->assertEquals(['name'], $model->safeAttributes());
        $this->assertEquals(['name', 'email'], $model->activeAttributes());
        $model->attributes = ['name'=>'mdmunir','email'=>'mdm@mun.com'];
        $this->assertNull($model->email);
        $this->assertFalse($model->validate());
```
Test not pass at line `$this->assertFalse($model->validate());`.
Its caused by this https://github.com/yiisoft/yii2/blob/master/framework/validators/Validator.php#L234

SilverFire:
@mdmunir you are setting attributes directly. Try `$model->load(...)` instead and assert that model's email is null.

mdmunir:
Look at this
```php
$this->assertNull($model->email);
```
This test pass. Its mean `email` is unsafe. OK, it is as expected. 
But model validation become true. its mean email attribute not be validated.

SilverFire:
Agree. Debugged, it really doesn't work as expected.

Faryshta:
Its explained in the docs here http://www.yiiframework.com/doc-2.0/yii-base-model.html#scenarios%28%29-detail but yes its hard to find maybe we should include it on the docs for the `rules()` method and the `validating-input.md` docs

arash26:
@Faryshta +1

DrDeath72:
guys, add priority to unsafe attribute in scenario
```php
['bla', 'foo', '!bla']
```
must be unsafe not safe, it's easy to add unsafe to scenario end by condition than clear safe and than add

SilverFire:
@DrDeath72 I didn't get it.

DrDeath72:
@SilverFire what?

mdmunir:
Agree with @DrDeath72 says.

DrDeath72:
```php
    public function safeAttributes()
    {
        $scenario = $this->getScenario();
        $scenarios = $this->scenarios();
        if (!isset($scenarios[$scenario])) {
            return [];
        }
        $attributes = [];
        foreach ($scenarios[$scenario] as $attribute) {
//            if ($attribute[0] !== '!') {
            if ($attribute[0] !== '!' && !in_array('!' . $attribute, $scenarios[$scenario]) {
                $attributes[] = $attribute;
            }
        }

        return $attributes;
    }
```

mdmunir:
https://github.com/mdmunir/yii2/commit/cda075602f8d5f0faa96b9929c9afee1024f53bb

PR #8156, opened 2015-04-22
Update ActiveField.php

udokmeci:
the option uncheck is not used by /yii/helpers/BaseHtml::activeRadioList instead checked against unselect .

### DIFF
--- a/framework/widgets/ActiveField.php
+++ b/framework/widgets/ActiveField.php
@@ -452,7 +452,7 @@ class ActiveField extends Component
      * This method will generate the "checked" tag attribute according to the model attribute value.
      * @param array $options the tag options in terms of name-value pairs. The following options are specially handled:
      *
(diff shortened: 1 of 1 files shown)


cebe:
Can one of the admins verify this patch?

klimov-paul:
No, original text is correct.

udokmeci:
Ok the problem is with ActiveField::radioList beacuse [BaseHtml::activeRadioList](https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1598) calls [BaseHtml::activeListInput](https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1599) and checks for unselect. I was working right on it now and it works if unselect option passed not uncheck.

klimov-paul:
So?

udokmeci:
Ok my fault I assumed they are using same method for generation radio inputs.

PR #8158, opened 2015-04-22
add isset check on index which might not be set

EliasZ:


samdark:
How to reproduce the error you're fixing?

cebe:
Can one of the admins verify this patch?

EliasZ:
A colleague of mine got the error on a Windows 7 machine, upon inspecting the `$output` variable, it was consistently empty. I don't know the exact environment of his machine. A quick Google reveals a [similar issue].

cebe:
check looks valid, it should not crash if no output is received.

cebe:
Merged. Thanks!

samdark:
Changelog?

cebe:
not sure it is needed for such a small thing but I can add it if you insist

samdark:
Worth adding, I think.

cebe:
done.

PR #8171, opened 2015-04-24
Enh: allow user to send files which are not seekable (S3StreamWrapper, ...)

pgaultier:
When a stream is not seekable, the Response::sendFile() cannot be used.
Without this modification, it's not possible to send file (opened using S3StreamWrapper) which are located on Amazon S3.


samdark:
Merged. Thanks!

PR #8174, opened 2015-04-24
Fixes #8036 : `yii\log\Logger` unable to export session id

### LINKED ISSUE #8036: yii session is not active when log message to file
Hi,
I find that yii session component is inactive when the work flow process to log messages. In file yii/web/log/Target.php, line 286:

     $sessionID = $session && $session->getIsActive() ? $session->getId() : '-';

The asset $session->getIsActive() always returns false and thus sessionID is '-' in the log files. Where in controllers, function getIsActive returns true. I guest that session component calls close() method before logger logs messages.

klimov-paul:
Fixes #8036 : `yii\log\Logger` unable to export session id


samdark:
Looks OK.

cebe:
:+1:

gregos22:
Hi,

I have the same problem as tlanyan in yii2 version 2.0.7.
I resolved the bug by modify the file Logger.php, init() function, line no 120 like this :
$this->flush(**true**);

Is it a good solution ?

cebe:
@gregos22 please open a new issue for this as it should be fixed since 2.0.4 already. also provide more details on how to reproduce the issue.

PR #8177, opened 2015-04-24
Compare it equals

thiagotalma:
If the comparison was made as identical, all integers from forms submitted by the user will be marked as changed.

Even though this PR is not the best solution, this inconsistency needs to be solved.


cebe:
this has been discussed in detail many times already: e.g. #2790

PR #8179, opened 2015-04-24
Compare it equals

thiagotalma:
@cebe , how about stop this old thread #2790?

A parameter solves the problem.

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -519,7 +519,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
     public function isAttributeChanged($name)
     {
         if (isset($this->_attributes[$name], $this->_oldAttributes[$name])) {
-            return $this->_attributes[$name] !== $this->_oldAttributes[$name];
+            return $this->_attributes[$name] != $this->_oldAttributes[$name];
         } else {
             return isset($this->_attributes[$name]) || isset($this->_oldAttributes[$name]);
         }
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -514,12 +514,13 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
     /**
      * Returns a value indicating whether the named attribute has been changed.
      * @param string $name the name of the attribute
-     * @return boolean whether the attribute has been changed
+     * @param bool $identical if the comparison is made as identical
+     * @return bool whether the attribute has been changed
      */
-    public function isAttributeChanged($name)
+    public function isAttributeChanged($name, $identical = true)
     {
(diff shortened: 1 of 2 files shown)


thiagotalma:
@qiangxue , Please give your opinion here.

framework/db/BaseActiveRecord.php
```diff
@@ -514,12 +514,13 @@ public function markAttributeDirty($name)
...
-     * @return boolean whether the attribute has been changed
+     * @param bool $identical if the comparison is made as identical
+     * @return bool whether the attribute has been changed
      */
-    public function isAttributeChanged($name)
+    public function isAttributeChanged($name, $identical = true)
     {
         if (isset($this->_attributes[$name], $this->_oldAttributes[$name])) {
-            return $this->_attributes[$name] !== $this->_oldAttributes[$name];
+            return $identical ? $this->_attributes[$name] !== $this->_oldAttributes[$name] : $this->_attributes[$name] != $this->_oldAttributes[$name];
```
cebe: code looks confusing on one line, should be a normal if `if ($identical) ... else ...`

qiangxue:
What's the use of this parameter? It won't fix #2790.

framework/db/BaseActiveRecord.php
thiagotalma: Ok @cebe

thiagotalma:
@qiangxue the intention is not to solve the issue, but provide a way to bypass it when necessary.

qiangxue:
> ...but provide a way to bypass it when necessary.

In which use case?

thiagotalma:
```php
public function beforeSave($insert)
{
    if (!$insert && $this->isAttributeChanged('user_id', false)) {
        // Do something
    }

    return parent::beforeSave($insert);
}
```

@qiangxue 
Consider that user_id is a field indicating a relationship. You need to perform a procedure if the relationship changes.

Without modification of PR, even if you do not change a field it will always be considered changed because the comparison as identical.

qiangxue:
Convinced. Could you add a line to the CHANGELOG file? Thanks.

qiangxue:
thanks!

samdark:
But isn't it breaking backwards compatibility?

klimov-paul:
No, by default `!==` used just like it was before.

samdark:
I mean signature break.

klimov-paul:
Yes, if someone is overriding this method he will receive E_STRICT.

samdark:
Should be mentioned in upgrade guide then.

cebe:
I'm afraid we are missing this kind of break in some other places. e.g. here: https://github.com/yiisoft/yii2-framework/compare/2.0.3...master#diff-966df45f15a870463a3012e393f2c25dR247 have not scrolled further yet.

PR #8180, opened 2015-04-24
get Pagination data before Models

lav45:
``` php
    public function actionIndex()
    {
        $_GET['page'] = 3;

        $dataProvider = new ArrayDataProvider([
            'allModels' => [
                ['id' => 1],
                ['id' => 2],
                ['id' => 3],
                ['id' => 4],
                ['id' => 5],
                ['id' => 6],
                ['id' => 7],
            ],
            'pagination' => [
                'pageSize' => 2,
            ],
        ]);

        var_dump($dataProvider->getPagination()->getPage());
        // return => int 0

        var_dump($dataProvider->getModels());
        // return => array (size=2)
        //  0 => 
        //    array (size=1)
        //      'id' => int 1
        //  1 => 
        //    array (size=1)
        //      'id' => int 2

        return 'OK';
    }
```

``` php
    public function actionIndex()
    {
        $_GET['page'] = 3;

        $dataProvider = new ArrayDataProvider([
            'allModels' => [
                ['id' => 1],
                ['id' => 2],
                ['id' => 3],
                ['id' => 4],
                ['id' => 5],
                ['id' => 6],
                ['id' => 7],
            ],
            'pagination' => [
                'pageSize' => 2,
            ],
        ]);

        var_dump($dataProvider->getModels());
        // return => array (size=2)
        //  4 => 
        //    array (size=1)
        //      'id' => int 5
        //  5 => 
        //    array (size=1)
        //      'id' => int 6

        var_dump($dataProvider->getPagination()->getPage());
        // return => int 2

        return 'OK';
    }
```

### DIFF
--- a/tests/framework/data/ActiveDataProviderTest.php
+++ b/tests/framework/data/ActiveDataProviderTest.php
@@ -7,8 +7,10 @@
 
 namespace yiiunit\framework\data;
 
+use Yii;
 use yii\data\ActiveDataProvider;
 use yii\db\Query;
+use yii\web\Request;
 use yiiunit\data\ar\ActiveRecord;
 use yiiunit\data\ar\Customer;
 use yiiunit\data\ar\Item;
@@ -168,7 +170,6 @@ class ActiveDataProviderTest extends DatabaseTestCase
             'query' => $query->from('order')->orderBy('id'),
         ]);
         $pagination = $provider->getPagination();
-        $this->assertEquals(0, $pagination->getPageCount());
         $this->assertCount(3, $provider->getModels());
         $this->assertEquals(1, $pagination->getPageCount());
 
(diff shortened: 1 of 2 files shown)


tests/framework/data/ActiveDataProviderTest.php
```diff
@@ -168,7 +170,6 @@ public function testPaginationBeforeModels()
             'query' => $query->from('order')->orderBy('id'),
         ]);
         $pagination = $provider->getPagination();
-        $this->assertEquals(0, $pagination->getPageCount());
```
LAV45: This line also points to this problem

LAV45:
And here is the source of the problem https://github.com/LAV45/yii2/blob/master/framework/data/Pagination.php#L191

xjflyttp:
I think you call before $dataProvider->prepare();

qiangxue:
The change is not correct. It will make the pagination configuration to be sensitive about whether pagination is configured first or not.

As described in the API comment, you should call `prepare()` or `getModels()` before calling `getPagination()`.

LAV45:
@qiangxue,  If the api class has methods, procedure call, which may affect the result obtained, it may be worthwhile to protect such methods from external access?

PR #8187, opened 2015-04-25
Adjusted masked input to correctly generate the hash variable to store the plugin options

### LINKED ISSUE #7808: \yii\widgets\MaskedInput no generates special data attributes "data-plugin-name" and "data-plugin-options"
The error is in the method "run". The call to "registerClientScript" should be before creating the "input".

Change it:

```php
public function run()
{
    if ($this->hasModel()) {
        echo Html::activeTextInput($this->model, $this->attribute, $this->options);
    } else {
        echo Html::textInput($this->name, $this->value, $this->options);
    }
    $this->registerClientScript();
}
```

To

```php
public function run()
{
    $this->registerClientScript();
    if ($this->hasModel()) {
        echo Html::activeTextInput($this->model, $this->attribute, $this->options);
    } else {
        echo Html::textInput($this->name, $this->value, $this->options);
    }
}
```

wbraganca:
resolve #7808

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.4 under development
 -----------------------
 
+- Enh #7808: Adjusted Masked input to correctly generate the hash variable to store the plugin options (wbraganca)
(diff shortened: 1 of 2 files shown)


qiangxue:
As explained by @lynicidn, the code is correct because it uses js to register the inputmask plugin rather than relying on data attributes.

PR #8188, opened 2015-04-25
Updated path in header "Location" after creating a new model

zhdanovartur:
Before: http://hostname/resource/view?id={id}
After: Before: http://hostname/resource/{id}


cebe:
this change is not correct. If you see the url above you most likely have a problem with your URL rule configuration.

PR #8189, opened 2015-04-25
Improve performance by using static variables for class-level data

laszlovl:
While profiling a Yii2 application, I discovered some interesting results: a few (yii-internal) functions that were called a very large amount of times, while they were essentially calculating static data.

A good example is BaseActiveRecord's __get() function: almost every invocation results in a call to BaseActiveRecord.hasAttribute(), which leads to Model.attributes(), which uses reflection on the object's class to determine its set of properties.

Right now, the entire process of reflection is re-calculated (almost) every time __get() is called on an object, while the return value of all these intermediary functions (hasAttribute(), attributes()) could be cached not only within an object instance, but within an object's class: while not all of the affected functions are static*, for all of them it's impossible that invocations on two different instances of the same class would ever return different results.

PHP's "static variables" (http://php.net/manual/en/language.variables.scope.php) are perfect for this situation: it remembers the contents of a variable in a function's scope across all invocations to that function within the same class.

With this commit, this strategy is added to all applicable functions that are likely to be called many times within the same request. The performance gains can be huge, especially in situations where many instances of the same ActiveRecord are instanciated in a single request. These are the profiling statistics for a real-world application where around 1350 ActiveRecord instances - spread over mostly 4 different classes - are used on a single page:

Before:

```
yii\db\Connection::getSchema    10,229 calls

Total Incl. Wall Time (microsec):   1,704,351 microsecs
Total Incl. CPU (microsecs):    1,688,537 microsecs
Total Incl. MemUse (bytes): 22,212,712 bytes
Total Incl. PeakMemUse (bytes): 23,626,296 bytes
Number of Function Calls:       245,465
```

After:

```
yii\db\Connection::getSchema    28 calls

Total Incl. Wall Time (microsec):   1,205,574 microsecs
Total Incl. CPU (microsecs):    1,198,505 microsecs
Total Incl. MemUse (bytes): 22,247,936 bytes
Total Incl. PeakMemUse (bytes): 23,598,200 bytes
Number of Function Calls:   168,656
```

.. an improvement in execution time of **30%**.

*We could later refactor these functions to become static, by using `new ReflectionClass(get_called_class())` instead of `new ReflectionClass($this)`.

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -239,9 +239,15 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      */
     public function formName()
     {
-        $reflector = new ReflectionClass($this);
+        static $formName = null;
 
-        return $reflector->getShortName();
+        if ($formName === null) {
+            $reflector = new ReflectionClass($this);
+
+            $formName = $reflector->getShortName();
+        }
+
+        return $formName;
     }
 
     /**
@@ -252,15 +258,21 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      */
     public function attributes()
     {
-        $class = new ReflectionClass($this);
-        $names = [];
-        foreach ($class->getProperties(\ReflectionProperty::IS_PUBLIC) as $property) {
-            if (!$property->isStatic()) {
-                $names[] = $property->getName();
+        static $attributes = null;
+
+        if ($attributes === null) {
+            $class = new ReflectionClass($this);
+            $names = [];
+            foreach ($class->getProperties(\ReflectionProperty::IS_PUBLIC) as $property) {
+                if (!$property->isStatic()) {
+                    $names[] = $property->getName();
+                }
             }
+
+            $attributes = $names;
         }
 
-        return $names;
+        return $attributes;
     }
 
     /**
@@ -917,9 +929,15 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      */
     public function fields()
     {
-        $fields = $this->attributes();
+        static $fields = null;
 
-        return array_combine($fields, $fields);
+        if ($fields === null) {
+            $attributes = $this->attributes();
+
+            $fields = array_combine($attributes, $attributes);
+        }
+
+        return $fields;
     }
 
     /**
--- a/framework/db/ActiveRecord.php
+++ b/framework/db/ActiveRecord.php
@@ -114,7 +114,13 @@ class ActiveRecord extends BaseActiveRecord
      */
     public function loadDefaultValues($skipIfSet = true)
     {
-        foreach ($this->getTableSchema()->columns as $column) {
+        static $columns = null;
+
+        if ($columns === null) {
+            $columns = $this->getTableSchema()->columns;
+        }
+        
+        foreach ($columns as $column) {
             if ($column->defaultValue !== null && (!$skipIfSet || $this->{$column->name} === null)) {
                 $this->{$column->name} = $column->defaultValue;
             }
@@ -281,7 +287,13 @@ class ActiveRecord extends BaseActiveRecord
      */
     public static function tableName()
     {
-        return '{{%' . Inflector::camel2id(StringHelper::basename(get_called_class()), '_') . '}}';
+        static $tableName = null;
+
+        if ($tableName === null) {
+            $tableName = '{{%' . Inflector::camel2id(StringHelper::basename(get_called_class()), '_') . '}}';
+        }
+
+        return $tableName;
     }
 
     /**
@@ -291,12 +303,18 @@ class ActiveRecord extends BaseActiveRecord
      */
     public static function getTableSchema()
     {
-        $schema = static::getDb()->getSchema()->getTableSchema(static::tableName());
-        if ($schema !== null) {
-            return $schema;
-        } else {
-            throw new InvalidConfigException("The table does not exist: " . static::tableName());
+        static $tableSchema = null;
+
+        if ($tableSchema === null) {
+            $schema = static::getDb()->getSchema()->getTableSchema(static::tableName());
+            if ($schema !== null) {
+                $tableSchema = $schema;
+            } else {
+                throw new InvalidConfigException("The table does not exist: " . static::tableName());
+            }
         }
+
+        return $tableSchema;
     }
 
     /**
@@ -314,7 +332,13 @@ class ActiveRecord extends BaseActiveRecord
      */
     public static function primaryKey()
     {
-        return static::getTableSchema()->primaryKey;
+        static $primaryKey = null;
+
+        if ($primaryKey === null) {
+            $primaryKey = static::getTableSchema()->primaryKey;
+        }
+
+        return $primaryKey;
     }
 
     /**
@@ -324,7 +348,13 @@ class ActiveRecord extends BaseActiveRecord
      */
     public function attributes()
     {
-        return array_keys(static::getTableSchema()->columns);
+        static $attributes = null;
+
+        if ($attributes === null) {
+            $attributes = array_keys(static::getTableSchema()->columns);
+        }
+
+        return $attributes;
     }
 
     /**
@@ -364,7 +394,12 @@ class ActiveRecord extends BaseActiveRecord
      */
     public static function populateRecord($record, $row)
     {
-        $columns = static::getTableSchema()->columns;
+        static $columns = null;
+
+        if ($columns === null) {
+            $columns = static::getTableSchema()->columns;
+        }
+        
         foreach ($row as $name => $value) {
             if (isset($columns[$name])) {
                 $row[$name] = $columns[$name]->phpTypecast($value);
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -418,7 +418,13 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      */
     public function hasAttribute($name)
     {
-        return isset($this->_attributes[$name]) || in_array($name, $this->attributes());
+        static $attributes = [];
+
+        if (!array_key_exists($name, $attributes)) {
+            $attributes[$name] = isset($this->_attributes[$name]) || in_array($name, $this->attributes());
+        }
+
+        return $attributes[$name];
     }
 
     /**
--- a/framework/db/Command.php
+++ b/framework/db/Command.php
@@ -310,12 +310,14 @@ class Command extends Component
             return $this;
         }
 
+        $schema = $this->db->getSchema();
+
         foreach ($values as $name => $value) {
             if (is_array($value)) {
                 $this->_pendingParams[$name] = $value;
                 $this->params[$name] = $value[0];
             } else {
-                $type = $this->db->getSchema()->getPdoType($value);
+                $type = $schema->getPdoType($value);
                 $this->_pendingParams[$name] = [$value, $type];
                 $this->params[$name] = $value;
             }
--- a/tests/framework/db/pgsql/PostgreSQLActiveRecordTest.php
+++ b/tests/framework/db/pgsql/PostgreSQLActiveRecordTest.php
@@ -168,6 +168,11 @@ class PostgreSQLActiveRecordTest extends ActiveRecordTest
         $record->save(false);
         $this->assertEquals(5, $record->primaryKey);
     }
+    
+    public function getCustomerClass()
+    {
+        return Customer::className();
+    }
 }
 
 class BoolAR extends ActiveRecord
@@ -195,4 +200,8 @@ class UserAR extends ActiveRecord
             TimestampBehavior::className(),
         ];
     }
+}
+
+class Customer extends \yiiunit\data\ar\Customer
+{
 }


klimov-paul:
Overall this make sense.
However we need to be careful in such refactoring.
Keep in mind #1318 for example.

framework/base/Model.php
```diff
@@ -239,9 +239,15 @@ public function scenarios()
...
+        static $formName = null;
 
-        return $reflector->getShortName();
+        if ($formName === null) {
+            $reflector = new ReflectionClass($this);
+
+            $formName = $reflector->getShortName();
+        }
+
+        return $formName;
```
klimov-paul: Make sense

framework/base/Model.php
```diff
@@ -252,15 +258,21 @@ public function formName()
...
+        if ($attributes === null) {
+            $class = new ReflectionClass($this);
+            $names = [];
+            foreach ($class->getProperties(\ReflectionProperty::IS_PUBLIC) as $property) {
+                if (!$property->isStatic()) {
+                    $names[] = $property->getName();
+                }
             }
+
+            $attributes = $names;
```
klimov-paul: Make sense

framework/base/Model.php
```diff
@@ -917,9 +929,15 @@ public static function validateMultiple($models, $attributeNames = null)
      */
     public function fields()
     {
-        $fields = $this->attributes();
+        static $fields = null;
+
+        if ($fields === null) {
+            $attributes = $this->attributes();
```
klimov-paul: No, if `attributes()` return varyon the model state, which can be (see #1318) method will produce invalid result.

framework/db/ActiveRecord.php
```diff
@@ -114,7 +114,13 @@ class ActiveRecord extends BaseActiveRecord
      */
     public function loadDefaultValues($skipIfSet = true)
     {
-        foreach ($this->getTableSchema()->columns as $column) {
+        static $columns = null;
+
+        if ($columns === null) {
+            $columns = $this->getTableSchema()->columns;
+        }
```
klimov-paul: Unlikely to male sense.
Here you avoiding just single call of `getTableSchema` by the cost of extra memory usage, which stores exact(!) copy of `TableSchema::columns`.

framework/db/ActiveRecord.php
```diff
@@ -281,7 +287,13 @@ public static function find()
...
     public static function tableName()
     {
-        return '{{%' . Inflector::camel2id(StringHelper::basename(get_called_class()), '_') . '}}';
+        static $tableName = null;
+
+        if ($tableName === null) {
+            $tableName = '{{%' . Inflector::camel2id(StringHelper::basename(get_called_class()), '_') . '}}';
+        }
+
+        return $tableName;
```
klimov-paul: Does not really make sense.
Method `tableName()` is meant to be overriden by the child class, which should declare table name explicitely.
Still can be kept.

framework/db/ActiveRecord.php
```diff
@@ -291,12 +303,18 @@ public static function tableName()
...
     {
-        $schema = static::getDb()->getSchema()->getTableSchema(static::tableName());
-        if ($schema !== null) {
-            return $schema;
-        } else {
-            throw new InvalidConfigException("The table does not exist: " . static::tableName());
+        static $tableSchema = null;
+
+        if ($tableSchema === null) {
+            $schema = static::getDb()->getSchema()->getTableSchema(static::tableName());
```
klimov-paul: This eliminates the ability of refreshing table schema in case some changes to table structure have been made.

klimov-paul: Besides `Schema::getTableSchema()` already provide an internal cache.

framework/db/ActiveRecord.php
```diff
@@ -314,7 +332,13 @@ public static function getTableSchema()
...
     public static function primaryKey()
     {
-        return static::getTableSchema()->primaryKey;
+        static $primaryKey = null;
+
+        if ($primaryKey === null) {
+            $primaryKey = static::getTableSchema()->primaryKey;
+        }
+
+        return $primaryKey;
```
klimov-paul: Unlikely to make sense.
Here you avoiding just single call of getTableSchema by the cost of extra memory usage, which stores exact(!) copy of TableSchema::primaryKey.

framework/db/ActiveRecord.php
```diff
@@ -324,7 +348,13 @@ public static function primaryKey()
      */
     public function attributes()
     {
-        return array_keys(static::getTableSchema()->columns);
+        static $attributes = null;
+
+        if ($attributes === null) {
+            $attributes = array_keys(static::getTableSchema()->columns);
```
klimov-paul: Same problem: if table structure changes there is no way to refresh attributes.

framework/db/ActiveRecord.php
```diff
@@ -364,7 +394,12 @@ public function transactions()
      */
     public static function populateRecord($record, $row)
     {
-        $columns = static::getTableSchema()->columns;
+        static $columns = null;
+
+        if ($columns === null) {
+            $columns = static::getTableSchema()->columns;
```
klimov-paul: And here again.

framework/db/BaseActiveRecord.php
```diff
@@ -418,7 +418,13 @@ public function getRelatedRecords()
...
     public function hasAttribute($name)
     {
-        return isset($this->_attributes[$name]) || in_array($name, $this->attributes());
+        static $attributes = [];
+
+        if (!array_key_exists($name, $attributes)) {
+            $attributes[$name] = isset($this->_attributes[$name]) || in_array($name, $this->attributes());
+        }
+
+        return $attributes[$name];
```
klimov-paul: Once agian see #1318

framework/db/Command.php
```diff
@@ -310,12 +310,14 @@ public function bindValues($values)
             return $this;
         }
 
+        $schema = $this->db->getSchema();
```
klimov-paul: Make sense

framework/db/ActiveRecord.php
laszlovl: @klimov-paul Don't those two comments contradict eachother?

Schema::getTableSchema() indeed already has an [internal cache](https://github.com/yiisoft/yii2/blob/master/framework/db/Schema.php#L117), so if the table structure were to change in the middle of a request, the changes wouldn't automatically be picked up before this change either. If I see it correctly, this causes no functional changes, it just prevents the overhead of function calls to `getDb()`, `getSchema()`, `getTableSchema` and `tableName`.

If someone wanted to explicitly refresh the schema data, they could still call Schema::getTableSchema($refresh=true).

Or am I misunderstanding something?

framework/db/ActiveRecord.php
laszlovl: It does matter if loadDefaultValues is called multiple times.

```php
for(...) {
    $model = new Model;
    $model->loadDefaultValues();
}
```

framework/base/Model.php
laszlovl: Ah, you're right, I didn't consider someone overriding Model.attributes(), I'll have to revert this one. (.. continued below)

framework/db/BaseActiveRecord.php
laszlovl: This might be a different matter though, because here we are calling BaseActiveRecord.attributes() instead of Model.attributes().

In a Model, it might make sense to override `attributes()` based on information in the current instance. But in ActiveRecord, `attributes()` is tied to the definition of the database backend, so I can't think of a reason to do so there. (You would use scenarios instead)

framework/db/ActiveRecord.php
klimov-paul: > If someone wanted to explicitly refresh the schema data, they could still call Schema::getTableSchema($refresh=true).

Yes, but your `static $tableSchema` will still contain the old instance of the `TableSchema` so `Schema::getTableSchema('mytable', true)` will have NO effect on the Active Record.

framework/db/ActiveRecord.php
klimov-paul: Such thing may be told about any method. We can in similar improvements at many places. But it does not make much sense. Here the performance can be saved for multiple calls of `loadDefaultValues()`, which is actually a rare use case, while calling `loadDefaultValues()` just a single time, which is common for 'CRUD create action' perfomance decreses. Also you reserve an extra memory for the static variable, which can not be released anyhow untill program termination.

framework/db/BaseActiveRecord.php
klimov-paul: No, it not different. `ActiveRecord::attributes()` are bound to the table schema by default, but this does not mean someone can not override and adjust this method.

klimov-paul: No, it not different. `ActiveRecord::attributes()` are bound to the table schema by default, but this does not mean someone can not override and adjust this method.

framework/db/ActiveRecord.php
laszlovl: Right, of course. You can still refresh the schema in Schema, but you can't propagate it to any already instanciated class.

We could fix this by adding the $refresh parameter to ActiveRecord::getTableSchema() as well, like this.

```php
public static function getTableSchema($refresh = false)
{
    static $tableSchema = null;
    
    if ($refresh === true) {
        $tableSchema = null;
    }

    if ($tableSchema === null) {
        $schema = static::getDb()->getSchema()->getTableSchema(static::tableName(), $refresh);
        if ($schema !== null) {
            $tableSchema = $schema;
        } else {
            throw new InvalidConfigException("The table does not exist: " . static::tableName());
        }
    }

    return $tableSchema;
}
```

I understand that the schema data is already cached in Schema::getTableSchema(), but even then ActiveRecord::getTableSchema() is called very frequently and as you can see in my profiling results the overhead of all these small function calls does add up, while for the vast majority of applications a class' schema would never change in the middle of a request.

laszlovl:
I'll wait a while for more feedback, and otherwise perhaps I should move the parts that are not under discussion to a separate PR?

framework/db/BaseActiveRecord.php
laszlovl: Right, it's technically possible, regardless of whether it makes sense to do so.

For these cases, perhaps we could place this logic in a separate class, document it and write about it in the Performance Tuning guide. (Or even in a separate extension if it's not wanted in the core)

```php
    public function attributes()
    {
        static $attributes = null;

        if ($attributes === null) {
            $attributes = parent::attributes();
        }

        return $attributes;
    }
```

framework/base/Model.php
samdark: :+1:

framework/base/Model.php
samdark: :+1:

framework/db/ActiveRecord.php
samdark: Changes behavior significantly. I don't think we should do it.

patlecat:
I feel this discussion is sometimes getting way too academic when people say things like "makes no sense because it's only called rarely", etc.. The way PHP defines **static** variables they **should be used in every function and loop that is short enough not to loose control over** it! Make it a **standard in Yii2** and then look at the places where it could pose a problem to introduce them, not the other way round. Use the engineering approach not the academic one!

framework/db/ActiveRecord.php
samdark: Any actual benchmark done for this change?

framework/base/Model.php
```diff
@@ -239,9 +239,15 @@ public function scenarios()
      */
     public function formName()
     {
-        $reflector = new ReflectionClass($this);
+        static $formName = null;
```
mikehaertl: Isn't it a kind of a de-facto convention to rather use protected class properties in all these cases?

```
protected $_var;

public function getFoo()
{
    if ($this->_var === null) {
        $this->_var = ...
    }
    return $this->_var
}
```

I think it's effectively doing the same. But I'd vote to use one pattern only. This pattern seems to be more established in the current code base.

laszlovl: @mikehaertl There's quite a difference: class properties work on the instance level, static variables in the function scope on the class level.

If you have a class `Car` and `Car::findAll()` returns 10 different models that you call `->formName()` on, the first variant would execute the logic in `formName` 10 times (once for each instance), the latter only a single time.

mikehaertl: @laszlovl Thanks, that's quite interesting. I've never really thought about what happens to `static` variables in class methods. I wonder if this behavior is documented in the PHP manual somewhere. I would have expected it [here](http://php.net/manual/en/language.variables.scope.php#language.variables.scope.static), but it's missing.

mikehaertl: @laszlovl So wouldn't an alternative be a static class method?

```
protected static $_var;

public function getFoo()
{
    if (static::$_var === null) {
        static::$_var = ...
    }
    return static::$_var
}
```

By using `static::$_var` instead of `self::$_var` it would even allow child classes to keep their own values (at least that's what I think it does). Maybe also a feasible approach?

laszlovl:
I found a few more which have can a huge impact during (bulk) updates: for example `Model::scenarios()` could be safely cached amongst all instances of a class in most codebases, as long as the logic in all of your `rules()` is static and doesn't reference `$this`.

Perhaps I should create a separate extension with a "faster but less flexible" ActiveRecord that includes all these performance tweaks and carefully explain the caveats.

klimov-paul:
Validation rules and thus scenarios can be added dynamically via `Model::getValidators()`.
Performance pursuit here unlikely to make sense as usually only single instance of the same model is validated during single application run.

laszlovl:
Thanks for all the feedback. I moved the uncontended changes to #8247, and the remaining optimizations to a separate extension: https://github.com/laszlovl/yii2-staticactiverecord

I think this can be closed now.

PR #8191, opened 2015-04-26
Added ability to render errors with the custom layout

mervick:
I think this is a good idea to add a simple way to configure error layout.


klimov-paul:
You can change layout inside the view file if you needed.
Such addition is redundant.

PR #8194, opened 2015-04-26
Cache which rule belongs to a given route+params

laszlovl:
Something else that stood out while profiling a Yii2 application was the huge amount of calls to `UrlRule::createUrl`.

Currently, every time a url is generated from a route (for example, by `Html::a()`), UrlManager iterates over all existing url rules until a matching rule is found. Imagine your application has 50 url rules, and you have 100 links on a page: in the worst case this would result in 5000 different calls to `createUrl`.

This doesn't seem neccessary at all. Once we have determined that a certain url rule matches with the route `['car/view', 'id' => $id]`, all subsequent calls to `UrlManager::createUrl` with that same signature (route name + parameter names) can instantly use the rule we matched earlier.

This patch does just that, by using `$route . ',' . implode(',', array_keys($params));` as cache key and storing the matched rule (or the fact that none of the url rules match) with that key.

Profiling results for a page containing around 1000 different links: a few extra kilobytes of memory to store the matches, but a **16%** performance increase.

```
Before:
yii\web\UrlRule::createUrl  43,309  calls
Total Incl. Wall Time (microsec):   1,671,670 microsecs
Total Incl. CPU (microsecs):    1,662,765 microsecs
Total Incl. MemUse (bytes): 20,533,208 bytes
Total Incl. PeakMemUse (bytes): 21,317,528 bytes
Number of Function Calls:   237,350

After:
yii\web\UrlRule::createUrl  1,202   calls
Total Incl. Wall Time (microsec):   1,399,313 microsecs
Total Incl. CPU (microsecs):    1,374,664 microsecs
Total Incl. MemUse (bytes): 20,541,008 bytes
Total Incl. PeakMemUse (bytes): 21,324,616 bytes
Number of Function Calls:   183,039
```

### DIFF
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -129,6 +129,7 @@ class UrlManager extends Component
     private $_baseUrl;
     private $_scriptUrl;
     private $_hostInfo;
+    private $_ruleMatchCache;
 
 
     /**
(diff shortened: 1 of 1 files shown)


klimov-paul:
This will produce unexpected results for the following set of rules:
```php
[
    [
        'pattern' => 'posts',
        'route' => 'user/activity',
        'defaults' => ['type' => 'post'],
    ],
    [
        'pattern' => 'photos',
        'route' => 'user/activity',
        'defaults' => ['type' => 'photo'],
    ],
]
```

cebe:
@klimov-paul is this a valid use case? The second rule should never be used, neither when creating nor when parsing urls.

klimov-paul:
Why?
These rules mean:
- for url '/posts' use controller 'User', action 'activity' and fill 'type' param to be 'post'
- for url '/photos' use controller 'User', action 'activity' and fill 'type' param to be 'photo'

Or have I written something wrong?

klimov-paul:
More reliable version created at #8205

PR #8196, opened 2015-04-26
Change way to get

shirokovroma:
Using http may call issue, changet to https. More https://getcomposer.org/doc/00-intro.md#installation-linux-unix-osx


samdark:
Thanks!

PR #8205, opened 2015-04-27
Fix #8194 create url cache

klimov-paul:
Fix #8194 create url cache


framework/web/UrlManager.php
```diff
@@ -309,18 +310,38 @@ public function createUrl($params)
...
+                        $this->_createUrlRuleMatches[$cacheKey][] = $rule;
+                        break;
+                    }
+                }
+            }
+
+            if ($url !== false) {
+                if (strpos($url, '://') !== false) {
+                    if ($baseUrl !== '' && ($pos = strpos($url, '/', 8)) !== false) {
+                        return substr($url, 0, $pos) . $baseUrl . substr($url, $pos);
```
samdark: What about non-English URLs?

klimov-paul: This is not my addition. This code is just moved out from `foreach`.

samdark: OK.

framework/web/UrlManager.php
```diff
@@ -309,18 +310,38 @@ public function createUrl($params)
...
                     } else {
-                        return "$baseUrl/{$url}{$anchor}";
+                        return $url . $baseUrl . $anchor;
```
qiangxue: Is this change correct?

klimov-paul: This is not a change. GIT just compose a 'smart' diff, so it is look that way.
This is actually a former line 319 (see above). It just have been moved.
Original line now is line 344 - it is intact.

qiangxue:
Thanks! https://github.com/yiisoft/yii2/commit/ebc089e6c26ba85642e5b32eaae017c6b637be79

PR #8207, opened 2015-04-27
Clear 'static' from method signature

ufocoder:
This method [is called as not static method](https://github.com/yiisoft/yii2/blob/ff71eb2251cac033522e7c54d1bc6cbb027b39da/framework/log/Target.php#L102)


klimov-paul:
No orignal declaration is correct.
Any static method can be called from object (via `->`).

ufocoder:
Yes, but this method is called inside \yii\log\Target class only and it is called as non-static :alien:

cebe:
it may be used in other places too.

PR #8218, opened 2015-04-28
Update Schema.php

vbelogai:
fixed: $columnSchemas is not initialized; 
added: size for binding to $returnParams


klimov-paul:
Thank you.

PR #8226, opened 2015-04-29
Update core-code-style.md

lmuzinic:
Fixed Typo


cebe:
original is correct, StudlyCaps is more about arbitrary casing while CamelCase is ucfirst() per word.

PR #8244, opened 2015-04-30
docs/guide* update copyright year to 2015

ghost:


samdark:
I don't think the change is correct. (c) usage is about year content was initially created.

vchenin:
i think it's reasonable to change year, because content of the guides was particularly changed at this year already

cebe:
it does not really matter, year is mainly updated to keep copyright after 100 years to let it not expire as public domain. I do not think anyone cares if the docs become public domain in 100 years.

vchenin:
ok

PR #8247, opened 2015-05-01
Performance tweaks by caching static data

laszlovl:
New PR based on #8189, only with the parts that were not under discussion.

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -239,9 +239,15 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      */
     public function formName()
     {
-        $reflector = new ReflectionClass($this);
+        static $formName = null;
 
-        return $reflector->getShortName();
+        if ($formName === null) {
+            $reflector = new ReflectionClass($this);
+
+            $formName = $reflector->getShortName();
+        }
+
+        return $formName;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


samdark:
@yiisoft/core-developers should be reviewed once again.

qiangxue:
The proposed optimization approach is problematic. See this gist: https://gist.github.com/qiangxue/9c6635ffeacef3eff8a9
You will get "BB" rather than "BA".

mikehaertl:
@qiangxue You can handle this with a static map array of `className => cachedValue` in your class `A`:

```php
class A {
    protected static $a = [];
    function t() {
        $className = get_called_class();
        if (!array_key_exists($className, self::$a)) {
            self::$a[$className] = $className;
        }
        return self::$a[$className];
    }
}
```

This also has the advantage that it doesn't use static function vars (which is not used anywhere in Yii AFAIK), but static class vars (which is used a lot).

qiangxue:
@mikehaertl Yes, using static map is the only reliable way we have found so far in such situation.

Overall I think we should be extremely careful when doing such kind of micro-optimizations because it may bring in unexpected side effects. We should mainly do this to deal with performance bottlenecks on the critical path. If we take a look at the three places being optimized in this PR, we may find they are not really called frequently (`formName()` is only used when you take form inputs, which usually happens once; `tableName()` is usually overridden with a plain string; `attributes()` is mainly used when setting/getting attributes in batch, which again isn't happening frequently.)

I am closing this PR for the above reasons. Thank you for your work.

laszlovl:
Ok. I have to admit that the changes in this PR aren't highly likely to cause a performance impact. The ones that do are in ActiveRecord, but aren't safe to apply in every single codebase, so they are now in a seperate extension here: https://github.com/laszlovl/yii2-staticactiverecord

Do check out this small change though, it's completely harmless and saves a lot of function calls for queries with a lot of parameters (as often happens in eager loading): https://github.com/yiisoft/yii2/pull/8247/files#diff-3898926ea9b3c7307c9d94389ab91b7bR313

cebe:
picked it from the commit, thanks! https://github.com/yiisoft/yii2/commit/8c222705917624719e2a6e73d787a2e34897732d

klimov-paul:
#8315

dynasource:
one thing to add the conversation here. Many problems can arise during PHP functional/integration testing. Static variables in methods stay 'alive' which can have unexpected side effects, like with the following example:

```php
private function getSomething(){
  static $a;
  if(!isset($a)){
    $a = $changingVariablePerRequest;
  }
  
}
```

A more a robust approach is that one of @mikehaertl https://github.com/yiisoft/yii2/pull/8247#issuecomment-98325925

PR #8249, opened 2015-05-01
enh: docs/guide/intro-yii add id to link

vchenin:
Because translators must translate anchor text, it's better to use id for reference,
so they shouldn't translate it twice. Also not all translators puts same text in this case.


cebe:
Thanks!

PR #8262, opened 2015-05-04
add gzcompress for FileCache

msajko:


klimov-paul:
This does not make sense. Cache will loose performce wit this change.

msajko:
This was the first step for enabling Gzip compression in PageCache. By enabling GZIP compression on your server, you can save around 50% on your bandwidth usage. Sadly, many people still forget to implement it on the web servers.

klimov-paul:
Response GZIP compression does not relates to the `cache`.
Also, the best way to achieve such feature will be proper configuration of the web server.

PR #8268, opened 2015-05-04
Added suport of everithing convertable to unsigned int to limit and offs...

### LINKED ISSUE #7023: \yii\db\QueryBuilder limit and offset rejects integer values in double type
PHP 5.6, Yii2 2.0.2

Ceil result passed to limit method will ignore value:
```php
\yii\db\ActiveRecord::find()->limit(ceil(5/2))->all();
```
Because ceil return type is double: 

> The return value of ceil() is still of type float as the value range of float is usually bigger than that of integer.

### LINKED ISSUE #8266: Offset and limit ignored in subquery
When I use subquery, offset and limit arent in generated sql (postgresql):

Original query:

```sql
SELECT * FROM tag WHERE id IN (
    SELECT tag_id FROM (
        SELECT * FROM view LIMIT :limit OFFSET :offset
    ) AS cluster
    GROUP BY tag_id
    ORDER BY COUNT(*)
 )
LIMIT :limit_per
```

PHP code:
```php
$querySlice = (new Query)->select('*')
    ->from('view')
    ->limit($cluster_size)
    ->offset($i);

$topOfSlice = (new Query)->select('tag_id')
    ->from(['cluster' => $querySlice])
    ->groupBy('tag_id')
    ->orderBy(new Expression('COUNT(*)'));

$tags = Tag::find()
    ->where(['id' => $topOfSlice])
    ->limit($per_cluster_top)
    ->all();
```

Generated query from debugger:

```sql
SELECT * FROM "tag" WHERE "id" IN (
    SELECT "tag_id" FROM (
        SELECT * FROM "view"
    ) "cluster" 
    GROUP BY "tag_id" 
    ORDER BY COUNT(*)
) 
LIMIT 10
```

tino415:
fix #8266
fix #7023

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -846,7 +846,7 @@ class QueryBuilder extends \yii\base\Object
     {
         // Check if limit is not object convertable to string
         if(is_object($offset) && method_exists($offset, '__toString'))
-            $limit = "$limit";
+            $offset = "$offset";
 
         // Check if limit is unsigned integer bigger then 0
         return is_numeric($offset) && $offset > 0;
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -834,7 +834,7 @@ class QueryBuilder extends \yii\base\Object
             $limit = "$limit";
             
         // Check if limit is unsigned integer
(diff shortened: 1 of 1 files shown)


qiangxue:
Checked in a simplified version. Thanks!

PR #8272, opened 2015-05-04
added m4a to the list of extensions => mimeTypes available.

dmill-bz:
The title says it all. I added `audio/mp4` for the `.m4a` extensions. This is quite crucial when streaming the files to players.

### DIFF
--- a/framework/helpers/mimeTypes.php
+++ b/framework/helpers/mimeTypes.php
@@ -414,6 +414,7 @@ return [
     'm3a' => 'audio/mpeg',
     'm3u' => 'audio/x-mpegurl',
     'm3u8' => 'application/vnd.apple.mpegurl',
+    'm4a' => 'audio/mp4',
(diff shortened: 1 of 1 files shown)


PommeVerte:
For clarification. Some ressources may show that the appropriate mimeType is `audio/m4a` or any odd combination of both `mp4a*` or `x-m4a` however the file is still basically an mpeg-4 file. Some browsers (such as IE9) will even refuse to play the files if they aren't `audio/mp4`. Hence why that's the mimeType in this pull request.

cebe:
We are generating this file from http://svn.apache.org/viewvc/httpd/httpd/trunk/docs/conf/mime.types?view=markup please send your request to the apache project, otherwise it would be overwritten after next update. Thanks!

PommeVerte:
Thanks for the info. I've submitted a bug and patch to apache.

PR #8274, opened 2015-05-04
Fix FragmentCache when enabled is false.

### LINKED ISSUE #8273: FragmentCache Layout Issue? ob_start unpaired?
I am trying to use FragmentCache for a footer, but I want to be able to turn it off using the `$enabled` option http://www.yiiframework.com/doc-2.0/yii-widgets-fragmentcache.html#$enabled-detail (perhaps via a GET param or YII_DEBUG).

So in the main layout I have:
```
if ($this->beginCache('main-footer', ['enabled' => false])):
	echo $this->render('_footer', ['brand' => $brand]);
	$this->endCache();
endif;
```

This leaves me with `<![CDATA[YII-BLOCK-HEAD]]>` and such in my source HTML instead of what that is supposed to be replaced.

It appears that the issue is that https://github.com/yiisoft/yii2/blob/master/framework/widgets/FragmentCache.php#L89 runs `ob_start()` but there is not a corresponding `ob_get_clean()` to make for the correct number when `endPage()` is finally called.

Perhaps an else `ob_get_clean()` is needed at the end of the run? https://github.com/yiisoft/yii2/blob/master/framework/widgets/FragmentCache.php#L117

nkovacs:
Fixes #8273


cebe:
Thanks!

PR #8297, opened 2015-05-06
Added index to source_message table

rhertogh:
In the function loadMessagesFromDb the column sourceMessageTable.category is used in the where clause but is not indexed which might cause performance issues when the table gets big.
This patch adds the missing index.


samdark:
Are you sure there's an issue?

rhertogh:
@samdark Hi, I ran some profiling and the difference is much less than I expected, virtually none actualy:
```
without index: 0.04784300 sec
with index:    0.04442675 sec

MySQL Server version: 5.6.23
source records: 5570
translation records: 17899
```
Perhaps the difference is be bigger on slower machines (tested with ssd) but for the moment it looks like it's no issue.

PR #8311, opened 2015-05-07
Added Alert component example

joshijitendra:
Flash message example with Alert bootstrap component.


cebe:
Merged with some adjustments, thanks!

PR #8312, opened 2015-05-07
Updated comments in ManagerInterface

githubjeka:
It is necessary that triggered tips of result on viewing variables.


cebe:
Thanks!

Issue #7280, opened 2015-02-13
UploadedFile::getInstance() and formName()


oherych:
I'm working on web-service. I'm using next code for have model without form name:

``` PHP
$model->load(Yii::$app->request->post(), '')
```

 But this solution is not affect to

``` PHP
UploadedFile::getInstance($this, 'file');
```

Perhaps will be great have setFormName() method. This help in my case and also if developer want change form name. What is you opinion? Have sense prepare pull request for this? Or override method will be batter way?

DieselDraft:
I used temporary

```PHP

<?php
namespace common\classes;

trait ModelTrait  {

	private $_formName;

	public function formName() {
		if(isset($this->_formName)) {
			return $this->_formName;
		} else {
			return parent::formName();
		}
	}

	public function setFormName($name) {
		$this->_formName = $name;
	}

}

```

qiangxue:
Would `getInstanceByName()` work for you?

DieselDraft:
I'm using this form in several places. In web page and in API. On Web page I need have form Name. On API of course no

```PHP

class AppealForm extends Appeal {

    /**
     * @var UploadedFile
     */
    public $file;

    public function rules() {
        return ArrayHelper::merge(parent::rules(), [
        	[['file'], 'file', 'extensions' => 'gif, jpg', 'skipOnEmpty' => true],
        ]);
    }

    public function save($runValidation = true, $attributeNames = NULL) {
    	$transaction = $this->getDb()->beginTransaction();

    	if(parent::save($runValidation, $attributeNames)) {
    		$this->file = UploadedFile::getInstance($this, 'file');

    		if($this->file) {
    		   $photo = new Photo();
	            $photo->appeal_id = $this->getPrimaryKey();
	            $photo->path = $photo->appeal_id . '.' . $this->file->extension;

				if($this->file->saveAs(Yii::getAlias('@mediaPath/' . $photo->path))) {

		            if($photo->save()) {

			    		$transaction->commit();
			    		return true;
		            }
	        	}

                $this->addErrors($photo->getErrors());
    		} else {
				$transaction->commit();
			    return true;
    		}
    	}

    	$transaction->rollBack();
    }

}

```

samdark:
Covered by PSR-7 in Yii 3.0.

PR #8315, opened 2015-05-07
Performance of `yii\base\Model` methods `attributes()` and `formName()`

klimov-paul:
Extracted from #8247

> The proposed optimization approach is problematic. See this gist: https://gist.github.com/qiangxue/9c6635ffeacef3eff8a9
> You will get "BB" rather than "BA".

Fixed

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -239,9 +239,15 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      */
     public function formName()
     {
-        $reflector = new ReflectionClass($this);
+        static $formNames = [];
 
-        return $reflector->getShortName();
+        $className = get_class($this);
+        if (!isset($formNames[$className])) {
+            $reflector = new ReflectionClass($this);
+            $formNames[$className] = $reflector->getShortName();
+        }
+
+        return $formNames[$className];
     }
 
     /**
@@ -252,15 +258,21 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      */
     public function attributes()
     {
-        $class = new ReflectionClass($this);
-        $names = [];
-        foreach ($class->getProperties(\ReflectionProperty::IS_PUBLIC) as $property) {
-            if (!$property->isStatic()) {
-                $names[] = $property->getName();
+        static $attributes = [];
+
+        $className = get_class($this);
+        if (!isset($attributes[$className])) {
+            $class = new ReflectionClass($className);
+            $names = [];
+            foreach ($class->getProperties(\ReflectionProperty::IS_PUBLIC) as $property) {
+                if (!$property->isStatic()) {
+                    $names[] = $property->getName();
+                }
             }
+            $attributes[$className] = $names;
         }
 
-        return $names;
+        return $attributes[$className];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cebe:
have you run any script to test how much this brings?

klimov-paul:
I am afraid I don't have a pure environment for this.
Still, you can refer to begining of this at #8189

laszlovl:
I have created a benchmark script to go with https://github.com/laszlovl/yii2-staticactiverecord. It works on `ActiveRecord` instead of `Model`, so it can't test `Model::attributes()`, but you could adapt it fairly easily.

qiangxue:
>  We should mainly do this to deal with performance bottlenecks on the critical path. If we take a look at the three places being optimized in this PR, we may find they are not really called frequently (formName() is only used when you take form inputs, which usually happens once; tableName() is usually overridden with a plain string; attributes() is mainly used when setting/getting attributes in batch mode, which again usually happens once at most.)

This is my comment about such optimization. I think in this situation, it's better we keep simple code rather than micro-optimization.

klimov-paul:
`Model::attributes()` is involved at `setAttributes()` and `getAttributes()`, thus it is involved at `Model::popuplate()` and at any processing, which comes in case of successful validation.

> formName() is only used when you take form inputs, which usually happens once

This is not true. `Model::formName()` is invoked by `Html::getInputName()`, which is invoked per each(!) created form control. Also `Html::getInputId()` invokes `Html::getInputName()`.
If form have 10 inputs `formName()` will be executed 20 times.

Also do not forget about possible tabular input.

cebe:
why did you close it?

klimov-paul:
I have found somehing interesting:
https://github.com/yiisoft/yii/pull/2739#discussion_r5756308

Although I still think `attributes()` and `formName()` may be improved, unless there is a benchmark, which proofs such optimization makes actual sense, I will not proceed with it.

Unfortunally, I don't have a clear environment to perfrom such benchmark.
If anyone is willing to investigate issue further, feel free to reopen this.

cebe:
lets keep this open, I will run some benchmarks, when I find time.

klimov-paul:
Relates #7280

Faryshta:
I don't like that the `static` variable declaration is inside the function. I know its allowed on php 5.4 but it will make it harder to debug since all the other functions and methods are class properties and not static function properties.

And I also think you need to update the `ActiveRecord::refresh()` to unset the formName and attribute list after refreshing.

And finally... how much this would affect the usage of scenarios?

klimov-paul:
> I don't like that the static variable declaration is inside the function. I know its allowed on php 5.4 but it will make it harder to debug since all the other functions and methods are class properties and not static function properties.

Using class properties in this case will produce redundant meta data, if you will you may override  `attributes()` or `formName()` creating your own implementation.

> I also think you need to update the ActiveRecord::refresh() to unset the formName and attribute list after refreshing.

Why? `ActiveRecord::refresh()` reloads attribute values(!) from DB, it does not relates to `attribute()` method. And most of all this PR does NOT affect `ActiveRecord` at all.

> how much this would affect the usage of scenarios?

It will not affect it anyhow.

cebe:
closing as this is minor. If someone finds the time feel free to do a benchmark.

Issue #8152, opened 2015-04-21
Сreate view via migration


unclead:
Cannot find similar issues.

Sometimes i need to create view. Nowadays i have to use `execute()` in this case.
Is it possible to imlement method `createView`?
Or it is special case and the best way is use `execute` method

samdark:
That's a good idea. Let's do it.

samdark:
Want to work on a pull request?

unclead:
@samdark ok i'll try

unclead:
@samdark i started work on a pull request and i have a question.
is it better create new methods for set/drop not null or better extend method `alterColumn` for each dbms?

klimov-paul:
Relates #3335

samdark:
New methods are better.

unclead:
As a part of this issue i implemented methods `dropNotNull` and `setNotNull`. See PR #8319
To implement functionality of working with `VIEW` i've created a separate issue #8347

sergeymakinen:
This issue is really related to #8163 and I think a `NULL`/`DEFAULT` constraint management should be shipped in 2.0.13 as well.
Hopefully I'm back and #8152 / #8319 both describe the whole process, so it should not go a long way.

Issue #8212, opened 2015-04-28
Failed unit test tests for DB section


unclead:
During implementation feature #8152 i found some failed tests for Cubrid DB
I use cubrid version 9.3

```
vendor/bin/phpunit --verbose tests/framework/db/cubrid/
PHPUnit 4.5.0-20-g10f6685 by Sebastian Bergmann and contributors.

Configuration read from /var/www/yii2/phpunit.xml.dist

................FF...................................F.........  63 / 195 ( 32%)
............................................................... 126 / 195 ( 64%)
............................................................S.. 189 / 195 ( 96%)
......

Time: 23 seconds, Memory: 14.75Mb

There were 3 failures:

1) yiiunit\framework\db\cubrid\CubridActiveRecordTest::testStoreNull
Failed asserting that false is true.

/var/www/yii2/tests/framework/db/ActiveRecordTest.php:231
/var/www/yii2/vendor/phpunit/phpunit/phpunit:36

2) yiiunit\framework\db\cubrid\CubridActiveRecordTest::testStoreEmpty
Failed asserting that false is true.

/var/www/yii2/tests/framework/db/ActiveRecordTest.php:277
/var/www/yii2/vendor/phpunit/phpunit/phpunit:36

3) yiiunit\framework\db\cubrid\CubridActiveRecordTest::testExplicitPkOnAutoIncrement
Failed asserting that 3 matches expected 1337.

/var/www/yii2/tests/framework/ar/ActiveRecordTestTrait.php:891
/var/www/yii2/vendor/phpunit/phpunit/phpunit:36
```

cebe:
thanks for reporting this. currently our test setup on travis and cubrid is broken, I am working on a new setup on own infrastructure.

unclead:
I set up the following configutration:
* postgreSQL 9.3
* Cubrid 9.3.2
* mysql 5.5.43
* oracle 11g

and found the following failures:
```
.............................................................   61 / 1174 (  5%)
.............................................................  122 / 1174 ( 10%)
...........................................F.................  183 / 1174 ( 15%)
.......................FF...................................F  244 / 1174 ( 20%)
.............................................................  305 / 1174 ( 25%)
.......................................................E.....  366 / 1174 ( 31%)
..........S........SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS  427 / 1174 ( 36%)
[...]

PR #8319, opened 2015-05-07
[Enh] Implementation of methods for managing NOT NULL constraint

unclead:
PR for #8152

Testes on:
- mysql
- cubrid
- oracle 11g
- postgresql 9.3

unfortunately I was not able to set up the environment to work with MSSQL that why i skipped this feature for this DB


unclead:
Also i added into an issue #8212 errors that i found during running unit test

unclead:
i will post PR for each feature separetely

PR #8338, opened 2015-05-09
Enh #8329: Extract messages for specified language

### LINKED ISSUE #8329: Enh: extract messages for specified language

```
```

Patch code:
```php
      */
     {
         if (!is_file($configFile)) {
             'format' => 'php',
             'ignoreCategories' => [],
         ], require($configFile));
+        }
 
```

vchenin:
close #8329


framework/console/controllers/MessageController.php
```diff
@@ -76,9 +76,12 @@ public function actionConfig($filePath)
      * @param string $configFile the path or alias of the configuration file.
      * You may use the "yii message/config" command to generate
      * this file and then customize it for your needs.
+     * @param string $specifiedLanguage the language code.
+     * This parameter override 'languages' array content
+     * from $configFile with one specified language code.
      * @throws Exception on failure.
      */
-    public function actionExtract($configFile)
+    public function actionExtract($configFile, $specifiedLanguage = NULL)
```
samdark: `specifiedLanguage` could be renamed to `language`.

samdark: We're using `null`, not `NULL`.

framework/console/controllers/MessageController.php
```diff
@@ -93,6 +96,9 @@ public function actionExtract($configFile)
             'format' => 'php',
             'ignoreCategories' => [],
         ], require($configFile));
+        if (!empty($specifiedLanguage)) {
+            $config['languages'] = [ $specifiedLanguage ];
```
samdark: No need for space after `[` and before `]`.

samdark: Can I specify multiple languages?

vchenin: We can use comma-separated list of language's codes and then explode it in array:
```php
$config['languages'] = explode(",", $languages);
```

framework/console/controllers/MessageController.php
```diff
@@ -76,9 +76,12 @@ public function actionConfig($filePath)
      * @param string $configFile the path or alias of the configuration file.
      * You may use the "yii message/config" command to generate
      * this file and then customize it for your needs.
+     * @param string $languages comma-separated list of language's codes.
+     * This parameter override 'languages' array content
+     * from $configFile with specified language's codes.
      * @throws Exception on failure.
      */
-    public function actionExtract($configFile)
+    public function actionExtract($configFile, $languages = null)
```
cebe: you can use an `array` typehint here to make the explode be done automatically by yii.
Here is an example of how this works: https://github.com/yiisoft/yii2-apidoc/blob/master/commands/ApiController.php#L42

vchenin: @cebe all done

PR #8339, opened 2015-05-09
add script for composer installation

vchenin:


zelenin:
composer should be a global tool, not project specific

vchenin:
not always
for example translators don't need globally accessible composer, but need it for producing reports with `php build translation`

zelenin:
@vchenin it's not argument for it.
Are you place PHPStorm or your other IDE/editor to project dir? Git? Other tools for project developing?

cebe:
still this script does not belong to yii in this form. Add it to your project if you need it.
For translators we could enhance the build command so that it ensures it can run out of the box.

vchenin:
I already advance functionality of script, now it's more useful.
Now is possible to install composer globally:
`sudo sh getcomposer.sh -g`
or locally:
`./getcomposer.sh -l`
or with specified options:
`./getcomposer.sh -o --filename=composer --version=1.0.0-alpha8`

PR #8364, opened 2015-05-11
Correcting beforeCopy example

YasserHassan:
The `$from` parameter takes values like `/folder1/folder2/folder3/css`. `dirname($from)` returns `/folder1/folder2/folder3` and therefore `basename(dirname($from))` returns `folder3` instead of the intended `css` in the given example to make it work right.


cebe:
isn't this called per file?

YasserHassan:
What do you mean @cebe?

YasserHassan:
This PR makes the example code create folders but the files are not published. It's not fully correct.

PR #8365, opened 2015-05-12
Changing the phrase 'the first chain in the process'.

elysdir:
I'd like to suggest some changes to the docs; figured I'd start with something small. I think the phrase "the first chain in the process" doesn't quite make sense, so I'm suggesting changing "chain" to "step."


klimov-paul:
Thank you.

PR #8373, opened 2015-05-12
FileValidator->getSizeLimit: check also php's post_max_size

maxxer:


samdark:
Good idea. Can you add a changelog? Also what do you think about reserving a few bytes more for it since there will form data as well.

maxxer:
I don't know if it's worth adding a fixed value to the size. The form may be very different, from a single input field to a big textarea....

samdark:
I think it worth doing since it's quite mysterious fail when post_max_size === upload_max_filesize and there's a file uploaded that matches max size.

framework/validators/FileValidator.php
```diff
@@ -258,7 +259,12 @@ protected function validateValue($file)
      */
     public function getSizeLimit()
     {
+        // Max POST size minus some bytes for form data 
+        $post_limit = $this->sizeToBytes(ini_get('post_max_size')) - 1024;
```
cebe: how do you decide for the constant of 1024 here?

samdark: Randomly, I guess :) You can't really predict it.

maxxer: Yep, that's a *possible estimate* of a form data. We cannot really know for sure

klimov-paul: Do not forget, that form may actually contain only single file input, so reserving any value here is inconsistent.

maxxer: That was my initial though, but @samdark correctly pointed out that it's likely that you have some form data along with the file, so it's correct to warn the user.
Anyway I think it's a border line situation, if the file size is SO close to the php limit it will almost certainly need to be raised.

samdark: @klimov-paul even in case of one input you need extra bytes for mime boundaries and input name.

cebe: In general one should configure POST size to be bigger than upload file size in PHP config. maybe we should just do a check and log a warning instead of using this value for guessing the max size?

klimov-paul: Method `getSizeLimit()` used only for descriptive purposes, while composing error message.
I suppose it is fine to use plain `ini_get('post_max_size')` for this case.

samdark: @cebe yep, that's even better.

maxxer: since the purpose of the check is to give the user a feedback I don't think it's useful to just log a message. 
I think using the pure value is ok, it's really a corner case if the user is trying to upload a file which is *bytes* greater than the limit... Also, considering people usually refer to sizes in MB they won't even see this difference.

cebe: Well, if `$post_limit < $limit` is true, I would consider this as a configuration error of the system. It still may use the value though. So both, show the value and log a warning.

maxxer: I've created commit 994929a which should address the above suggestions, please let me know if you have more

cebe:
Merged, thanks!

PR #8375, opened 2015-05-12
Fix #5219 Ordering Appoarch of Shutdown Functions

### LINKED ISSUE #5219: Suggestion for New Ordering Appoarch of Shutdown Functions
It is known that nested `register_shutdown_function` calls can cause problem under HHVM and there were no fix for this currently. The official PHP document does not guarantee that any sub-sequence calls of `register_shutdown_function` can be made during shutdown.

I suggest another approach to arrange the order of executions of shutdown functions by making use of the standard SplPriorityQueue class. Here is a simple example illustrating how this method works.
```php
<?php
function register_shutdown_callback(callable $c, $weight = 0)
{
    static $queue = null;
    if ($queue === null) {
        /* The queue will not be created and register_shutdown_function will not be called */
        /* if the function is not used. */
        $queue = new SplPriorityQueue();
        /*we need the weight as the only parameter for the shutdown callback, so extracting both.*/
        $queue->setExtractFlags(SplPriorityQueue::EXTR_BOTH); 
        register_shutdown_function(function() use ($queue)
        {
            while ($queue->valid()) {
                $arr  = $queue->current();
                $callback = $arr['data'];
                $priority = $arr['priority'];
                call_user_func($callback, $priority);
                $queue->next();
            }
        });
    }
    $queue->insert($c, $weight);
}

/* Some examples. */
register_shutdown_callback(function($weight)
{
    echo 'Hello, world!' . PHP_EOL;
    register_shutdown_callback(function($weight)
    {
        echo 'Hello again.' . PHP_EOL;
    }, $weight - 1);
});

/* This function is ensured to be called at last. */
register_shutdown_callback(function($weight)
{
    echo 'FINAL callback.' . PHP_EOL;
}, -INF);
```
... and the output
```
Hello, world!
Hello again.
FINAL callback.
```

klimov-paul:
Fix #5219 Ordering Appoarch of Shutdown Functions

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -533,4 +533,42 @@ class BaseYii
     {
         return get_object_vars($object);
     }
+
+    private static $shutdownHandlers;
+
+    /**
+     * Register a function for execution on PHP script shutdown.
+     * This method should be preferred to the plain `register_shutdown_function()` function.
+     * @param callable $callback the shutdown callback to register.
+     * @param array $parameters additional parameters for the the shutdown callback.
+     * @param integer $priority handler priority: the higher priority - the later callback will be executed,
+     * e.g. if 'callback1' has priority 8 and 'callback2' has priority 7, 'callback2' will be executed first,
+     * callbacks with same priority will be executed in order they been registered.
+     */
+    public static function registerShutdownFunction($callback, array $parameters = [], $priority = 0)
+    {
+        if (!is_array(self::$shutdownHandlers)) {
+            self::$shutdownHandlers = [];
+            register_shutdown_function([__CLASS__, 'shutdown']);
+        }
+        self::$shutdownHandlers[$priority][] = [$callback, $parameters];
+    }
+
+    /**
+     * This function will be automatically invoked on script shutdown in case at least one callback is
+     * registered via [[registerShutdownFunction()]].
+     */
+    public static function shutdown()
+    {
+        if (empty(self::$shutdownHandlers)) {
+            return;
+        }
+        ksort(self::$shutdownHandlers, SORT_NUMERIC);
+        foreach (self::$shutdownHandlers as $shutdownHandlers) {
+            foreach ($shutdownHandlers as $shutdownHandler) {
+                list($callback, $parameters) = $shutdownHandler;
+                call_user_func_array($callback, $parameters);
+            }
+        }
+    }
 }
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -534,7 +534,10 @@ class BaseYii
         return get_object_vars($object);
     }
 
-    private static $shutdownHandlers;
+    /**
+     * @var array|null list of registered shutdown functions in format: priority => [[callback, args], ...]
+     */
+    private static $shutdownFunctions;
 
     /**
      * Register a function for execution on PHP script shutdown.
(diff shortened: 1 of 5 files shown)


framework/log/Logger.php
```diff
@@ -115,13 +115,11 @@ class Logger extends Component
...
     {
         parent::init();
-        register_shutdown_function(function () {
-            // make regular flush before other shutdown functions, which allows session data collection and so on
-            $this->flush();
-            // make sure log entries written by shutdown functions are also flushed
-            // ensure "flush()" is called last when there are multiple shutdown functions
-            register_shutdown_function([$this, 'flush'], true);
-        });
+        // make regular flush before other shutdown functions, which allows session data collection and so on
```
cebe: maybe the session shutdown should be aware of this and register on a later priority.

In general we may introduce constants e.g. `Session::SHUTDOWN_PRIORITY = 50` , which can be used to register shutdown functions before some others, e.g. 

```php
// do something before session is closed
Yii::registerShutdownFunction($callback, [], Session::SHUTDOWN_PRIORITY - 1);
```

klimov-paul: Make sense.
Stil for the logger flush should be executed twice one way or another.

cebe: >  Stil for the logger flush should be executed twice one way or another.

yes thats true.

framework/BaseYii.php
```diff
@@ -533,4 +533,42 @@ public static function getObjectVars($object)
...
+    /**
+     * Register a function for execution on PHP script shutdown.
+     * This method should be preferred to the plain `register_shutdown_function()` function.
+     * @param callable $callback the shutdown callback to register.
+     * @param array $parameters additional parameters for the the shutdown callback.
+     * @param integer $priority handler priority: the higher priority - the later callback will be executed,
+     * e.g. if 'callback1' has priority 8 and 'callback2' has priority 7, 'callback2' will be executed first,
+     * callbacks with same priority will be executed in order they been registered.
+     */
+    public static function registerShutdownFunction($callback, array $parameters = [], $priority = 0)
```
samdark: `$shutdownHandlers` but `registerShutdownFunction()`. Could be either `$shutdownFunctions` or `registerShutdownHandler()`.

framework/BaseYii.php
```diff
@@ -533,4 +533,42 @@ public static function getObjectVars($object)
     {
         return get_object_vars($object);
     }
+
+    private static $shutdownHandlers;
```
samdark: Could be `[]` by default. Won't require you to initialize it later.

klimov-paul: No it could not.
Check `registerShutdownFunction()` it registers `shutdown()` as PHP shutdown function only on first demand, which `null` value indicates.

framework/BaseYii.php
klimov-paul: I ma unsure about proper name. The structure of this array is following:
```
[
    priority1 => [
        ['callback1', [args]],
        ['callback2', [args]],
    ],
    priority2 => [
        ['callback3', [args]],
        ['callback4', [args]],
    ],
]
```
So 'shutdownFunctions' also does not sound right.

klimov-paul: Although, perhaps `$shutdownFunctions` is better.

klimov-paul: Renamed.

framework/log/Logger.php
klimov-paul: Perhaps it is actully better to have a public property for this instead of constants?
```php
// do something before session is closed
Yii::registerShutdownFunction($callback, [], Yii::$app->session->shutdownPriority - 1);
```
This will also allow manipulation on priorities via application configuration array.

framework/BaseYii.php
samdark: Right. OK.

samdark:
@klimov-paul any more things to fix in this pull request?

klimov-paul:
The main question here: is it really needed?

samdark:
Ah, I have no idea. It seems to be HHVM-specific thing that was fixed.

samdark:
Do you have a way to verify it?

klimov-paul:
I don't have HHVM environment, so I can not tell if there is a problem in it.
If there is no restriction on amount of shutdown functions registrations, this change does not make much sense.

samdark:
It doesn't seem that there's any restriction nowadays.

PR #8393, opened 2015-05-13
Guide structure applications

elysdir:
Lightweight editing pass on Applications page, cleaning up grammar, punctuation, etc.


samdark:
Good changes. Thanks!

docs/guide/structure-applications.md
```diff
@@ -330,15 +330,15 @@ image size as a parameter like the following:
 ]
 ```
 
-Then in your code where you need to use the size value, you can simply use the code like the following:
```
vchenin: "like the following" -> "as follows"

docs/guide/structure-applications.md
```diff
@@ -330,15 +330,15 @@ image size as a parameter like the following:
...
+Then in your code where you need to use the size value, you can simply use code like the following:
```
vchenin: "like the following" -> "as follows"

cebe: whats wrong with "like the following"?

vchenin: nothing special
purpose of this remark to save same style in document 
all other word-combinations "like the following" was replaced with "as follows" except this one
this is not critical so may be ignored

cebe: I think in this case it is fine because it reads "use code like the following". In the other places it was a separated statement. Thanks for noting though.

elysdir: Yeah, I waffled about this instance--I was initially going to change it to match the others, but it felt to me like it read better this way.

PR #8403, opened 2015-05-13
Fix client-side trim validator.

### LINKED ISSUE #7707: FilterValidator|trim works incorrectly on client side
Here is my Model::rules:
```
public function rules()
{
    return [
        [['email', 'password'], 'required'],
        ['rememberMe', 'boolean'],

        [['email'], 'trim'],
        ['email', 'email'],
        ...
    ]
}
```

When I insert "     test@mail.com" trim works, but I still get error:
d32f0832-c995-11e4-93b2-cd9385334871

nkovacs:
Subsequent validators now use the trimmed value instead of the original value.

fixes #7707

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.5 under development
 -----------------------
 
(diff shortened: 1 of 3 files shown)


cebe:
looks good to me.

qiangxue:
Thanks!

PR #8404, opened 2015-05-13
Fix #8349: Properties `fields` and `expand` added to `yii\rest\Serializer`

### LINKED ISSUE #8349: Should class yii\rest\Serializer support  a public property $params like Pagination and Sort?
This will be useful when we want to use an array parameters (name => value) to specify `fields` and `expand` parameters instead of using values come from `Yii::$app->getRequest()`.

The implementing logic is just like the `params` property of `yii\data\Pagination` or  `yii\data\Sort`

klimov-paul:
Fix #8349
Properties `fields` and `expand` added to `yii\rest\Serializer`


mdmunir:
:+1: 
but litle question. How to inject `expand` properti from controller action? I have code like this
```php
public function actionView($id)
{
    $model = MyAr::findOne($id);
    return $model; // <- how to add Serializer::expand to this. IYKWIM
}
```

Faryshta:
i think this is a good proposal for default value for `fields` and `expand` which are allowed to get rewritten by GET parameters.

or example when visiting `sales` it would be good to have by default `$expand = ['products', 'customer', 'store']` but if someone request only the products its also valid. `sales?expand=products

PR #8415, opened 2015-05-14
implemented conditional comments for new IE versions and other modern browsers

salaros:
usage scenario: load both versions of jQuery (2.0 for IE9+ and other browsers and 1.x for IE8)

```
'yii\web\JqueryAsset' => [
    'depends' => [ 'app\assets\JqueryAssetIe8' ],        // injects JQuery legacy
    'jsOptions' => ['condition' => '(gte IE 9)|(!IE)']   // generates a conditional comment for IE9+
    ...
],
```

the snippet above will generate:

```
<!--[if lt IE 9]>
<script src="/assets/xxxxxxxx/jquery.min.js"></script>
<![endif]-->
<!--[if (gte IE 9)|(!IE)]><!-->
<script src="/assets/yyyyyyyy/jquery.min.js"></script>
<!--<![endif]-->
```


klimov-paul:
Issue is valid

klimov-paul:
Resolved by commit 3a39f8a

PR #8417, opened 2015-05-14
Requirement to install the ctype extension

ARACOOOL:


klimov-paul:
Indeed `ctype` extension can be not enabled.

klimov-paul:
Resolved by commit 219bf07

PR #8418, opened 2015-05-14
update schema.php getLastInsertedId

Sanya1991:
I become ORA00936 errors and ORA-08002: sequence MySequence.CURRVAL is not yet defined in this session.
This is the soulution.

### DIFF
--- a/framework/db/oci/Schema.php
+++ b/framework/db/oci/Schema.php
@@ -195,7 +195,7 @@ SQL;
             // get the last insert id from the master connection
             $sequenceName = $this->quoteSimpleTableName($sequenceName);
             return $this->db->useMaster(function (Connection $db) use ($sequenceName) {
-                return $db->createCommand("SELECT {$sequenceName}.CURRVAL FROM DUAL")->queryScalar();
+                return $db->createCommand("SELECT LAST_NUMBER FROM USER_SEQUENCES WHERE SEQUENCE_NAME=:sequenceName")->bindValue(':sequenceName', $sequenceName)->queryScalar();
             });
         } else {
             throw new InvalidCallException('DB Connection is not active.');


samdark:
@cebe do we have test environment for it?

Sanya1991:
Yes, but not public test environment.

cebe:
not yet, @nineinchnick can you check it?

nineinchnick:
I can't confirm that. I've added unit tests for `Schema::getLastInsertId()` in #8441 but it's working with the current code.

@Sanya1991 you are supposed to run an `INSERT` statement before calling `Schema::getLastInsertId()`, otherwise it won't work and it doesn't make sense anyway. If you just want to check the sequence value, execute a specific query.

Sanya1991:
08002. 00000 -  "sequence %s.CURRVAL is not yet defined in this session"
In a same session the sequence hasn't CURRVAL value if one is unused, you know?

The problem: I click the save button and yii catch the exception: sequence %s.CURRVAL is not yet defined in this session => the record is unsaved. 
After I push F5 on my keyboard and click yes send data again.. Then bumm the record is saved. Attach pictures?

I have a solution: SELECT LAST_NUMBER - 1 FROM USER_SEQUENCES WHERE SEQUENCE_NAME=:sequenceName

The -1 is important, because, the LAST_NUMBER is not the last inserted number, it's the next number..

This problem is old... http://saadnawaz.blogspot.hu/2010/10/using-single-query-determine-current.html

psrustik:
MR is not valid because: 
1) System object USER_SEQUENCES don't contains [all sequenses] (http://docs.oracle.com/cd/B14117_01/server.101/b10755/statviews_2630.htm). Schema for connected user and ActiveRecord schema may be various, for example:
```
//config.php
'db'=>[
    'dsn' => 'oci:dbname=CRMDB',
    'username' => 'web_user',
    'password' => 'mypassword',
];
// Model
class Record extends ActiveRecord {
    /**
     * @inheritdoc
     */
    {
        return 'ANOTHER_SCHEMA.RECORDS';
    }
}
```
2) CURRVAL method for sequence may be called only after inserting data with NEXTVAL.

In your case probably insert query not called.

samdark:
@psrustik you mean `PR` is not valid, not `MR`, right?

psrustik:
Yes, mean pull request

Faryshta:
can we still keep the idea to use parameters instead of php concatenation?

samdark:
No since parameter binding doesn't work for schema names.

PR #8435, opened 2015-05-15
ensure primary key string comparison is made identical

cebe:
this is to avoid cases where string primary keys starting with '0e'.

See http://securityaffairs.co/wordpress/36732/hacking/php-hash-comparison-flaw.html

Other parts of Yii 2, escpecially the security and auth parts, are **not** affected.


samdark:
Looks OK to me.

PR #8437, opened 2015-05-16
Update LinkPager.php

IgorSalevan:
Added "activeLinkable" property. 
It serves for turning off the anchor tag from the LinkPager active item. I set the default value to false and this changed the previous functionality.
The main idea of this revision is to follow google recommendation that the page must avoid links to itself.

### DIFF
--- a/framework/widgets/LinkPager.php
+++ b/framework/widgets/LinkPager.php
@@ -104,6 +104,11 @@ class LinkPager extends Widget
      * @var boolean Hide widget when only one page exist.
      */
     public $hideOnSinglePage = true;
+    /**
+     * @var boolean turns on|off the <a> tag for the active pager item. It is recommended to set it to false
+     * for the sake of Search Engines Optimisation. Defaults to false.
+     */
+    public $activeLinkable = false;
 
 
     /**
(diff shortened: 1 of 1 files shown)


samdark:
Would you please point to that Google recommendation?

cebe:
> The main idea of this revision is to follow google recommendation that the page must avoid links to itself.

This does not really make sense, most of navigation points to itself when you click the link in nav it is still there.

mulat:
**samdark**, **cebe**, there is no direct recommendation from Google. Only indirect. 
As I undersand, internal links serve to provide intuitive interface; create clear navigation; show SE crawlers new pages on sites. So the internal link to itself doesn't make any sense. SERP of Google doesn't have such link in LinkPager functionality, but Yandex has.
There are many threads on SEO forums where webmasters are concerned of getting rid of all excessive internel links and links to itself too. 
So I think that giving such a choice to WM is not a bad idea. At least the the property `activeLinkable` may have default value `true`, like the logic of the current core functionality.

samdark:
I prefer not to include unconfirmed SEO mythical stuff into framework. Closing unless there's a strong evidence of not linking to itself being a good thing.

PR #8440, opened 2015-05-16
Copyediting for Controllers page

elysdir:
General copyediting/cleanup of the Controllers page. Main changes: changing "dash" to "hyphen" throughout (for accuracy and clarity); changing bullet lists for procedures to numbered lists; assorted small grammar/clarity tweaks.


docs/guide/structure-controllers.md
```diff
@@ -420,25 +420,25 @@ to fulfill the request:
...
 
 ## Best Practices <span id="best-practices"></span>
 
-In a well-designed application, controllers are often very thin with each action containing only a few lines of code.
+In a well-designed application, controllers are often very thin, with each action containing only a few lines of code.
 If your controller is rather complicated, it usually indicates that you should refactor it and move some code
 to other classes.
 
-In summary, controllers
+Here are some specific best practices. Controllers
```
samdark: What do you mean by "some specific best practices"?

docs/guide/structure-controllers.md
```diff
@@ -112,32 +112,32 @@ For this reason, controller IDs are often nouns referring to the types of the re
...
 
-* Turn the first letter in each word separated by dashes into upper case. Note that if the controller ID
+1. Turn the first letter in each word separated by hyphens into upper case. Note that if the controller ID
   contains slashes, this rule only applies to the part after the last slash in the ID.
-* Remove dashes and replace any forward slashes with backward slashes.
-* Append the suffix `Controller`.
-* And prepend the [[yii\base\Application::controllerNamespace|controller namespace]].
+1. Remove hyphens and replace any forward slashes with backward slashes.
+1. Append the suffix `Controller`.
+1. Prepend the [[yii\base\Application::controllerNamespace|controller namespace]].
```
cebe: I'd prefer lists to be readable even when not compiled, so please use `1,2,3...` instead of only `1`.

docs/guide/structure-controllers.md
```diff
@@ -236,11 +236,11 @@ created to be used in different controllers or be redistributed as [extensions](
...
 
-The names of the action methods are derived from action IDs according to the following criteria:
+The names of the action methods are derived from action IDs according to the following procedure:
 
-* Turn the first letter in each word of the action ID into upper case;
-* Remove dashes;
-* Prepend the prefix `action`.
+1. Turn the first letter in each word of the action ID into upper case.
+1. Remove hyphens.
+1. Prepend the prefix `action`.
```
cebe: same as above.

docs/guide/structure-controllers.md
elysdir: This list used to be labeled as a summary, but as far as I can tell it's not a summary of anything on the page. Since it's in the Best Practices section, I assumed it was a list of best practices. But I may well have misunderstood. I'd be happy to replace that line with a better description of what this is a list of.

docs/guide/structure-controllers.md
elysdir: OK, done. But I'm fairly new to git, so let me know if I got anything wrong while adding that change. (I've been following the instructions at (https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md), which are super useful, but I may've done something wrong anyway.)

(Side note, just to explain why I did the numbering that way before: I tend to figure that since Markdown ignores the numbers specified, using the "1, 2, 3" format can give a misleading sense of security about what the step numbers will be when rendered. For example, if someone adds a new step in the middle but neglects to renumber the later steps, then hardcoded references like "go to step 4" will be wrong, but that won't necessarily be obvious in the Markdown source. That said, I do recognize the advantages of having readable source, and I'm happy to go with your standard, so I made the change--I'm not arguing against it, just explaining why I did it that way before.)

docs/guide/structure-controllers.md
elysdir: Addendum, since I'm new to this project and I don't know if my tone in my earlier note was clear: I'm happy to make whatever changes you folks want--my suggestions are just suggestions, and if I get something wrong or write something unclear or confusing, then I definitely want to correct that. The Yii docs are mostly really solid; I just want to tweak them a little to make them even better.

docs/guide/structure-controllers.md
cebe: I fully understand your arguments and they are valid in general. In our case, changes are reviewed by many people so we can go with explicit numbering without the fear to introduce errors.

qiangxue:
Good job. Thanks!

docs/guide/structure-controllers.md
cebe: All fine, your contributions are very welcome :)

I agree that summary does not make sense as it does not summarize anything. But "specific best practices" sounds weird as well.
I also find it weird to start with the word "Controller" and then have the list, imo it does not read well.

The purpose of the list is to show best practices when someone wants to decide which logic belongs to the controller and which not.

PR #8446, opened 2015-05-17
Extend the class description

lukos:
Help users understand the concept of massive assignment without relying on the Guide by extending the description for this type.


qiangxue:
Thanks!

PR #8490, opened 2015-05-20
Update yii\db\ActiveRecord code style

grigorieff:


samdark:
Isn't it OK as is?

klimov-paul:
@grigorieff, which code style entry do you think this related?

Issue #6347, opened 2014-12-02
inverseOf not working in most situations


tvdavid:
The mechanism for inverseOf is implemented in ActiveRelationTrait::findFor(). This means that in `$customer->orders` the system works fine, because it's routed through the magic getter: `return $this->_related[$name] = $value->findFor($name, $this)`.

However, if you're customizing the query, like `$customer->getOrders()->with('status')->all()`, the inverseOf system is bypassed entirely.

Worse, because ActiveDataProvider calls `all()` directly (prepareModels: `return $query->all($this->db)`) inverseOf doesn't work at all when a DataProvider is involved.

On first sight, it looks like the inverseOf logic could safely be applied to all calls to `all()`. Is it indeed just a matter of moving some code around or am I missing something?

VojtechH:
Is there any prospect of resolving this bug in the near future or should I rather implements some workaround in my application?

It seems to me that the solution @laszlovl suggested makes sense.

If there is currently no time to implement the fix perhaps it could be useful to mention here some tried-tested workaround.

VojtechH:
Also if the solution is not expected in the near future I would suggest to add some note to the documentation. I spent some time figuring out where the problem is. If the documentation specifies that the inverseOf() is currently supported only for the queries made through magic getter it would be much clearer.

davidhirtz:
@laszlovl fix looks fine to me, why is this not merged since May 20, 2015?

cebe:
@davidhirtz every pull request needs review and testing before merge. We have limited time so we have to prioritize. This issue is tagged for the 2.0.8 release.

cebe:
the fix is merged and will be released with 2.0.9

PR #8497, opened 2015-05-20
Make sure inverseOf relations are also populated during dynamic relational queries

laszlovl:
I also encountered the bug that's described in #6347.

The solution seemed obvious: populate the inverse relationships in `ActiveQuery::populate()` where it'll be applied for all calls to `ActiveQuery::all()` and `ActiveQuery::one()`, instead of in `ActiveRelationTrait::findFor()`, where it will only be applied for "static" relations retrieved via ActiveRecord's __get.

I added test coverage for these scenarios, verified that none of the existing tests break and ran my application's testsuite without any problems.

Extensions that implement `ActiveQueryInterface` in one of their classes, such as yii2-sphinx, will have to be updated as well to call `addInverseRelations()` from within `populate()`. If this PR is accepted, I can submit those changes as well.

cc @qiangxue


tests/framework/db/ActiveRecordTest.php
```diff
@@ -527,6 +527,7 @@ public function testInverseOf()
         $this->assertTrue($customer->orders2[0]->customer2 === $customer);
         $this->assertTrue($customer->orders2[1]->customer2 === $customer);
         // ad-hoc lazy loading
+        // NB: this batch of asserts probably does not do what was intended ($customer->orders2[0] vs $orders[0])
```
cebe: good catch, if this test would be written correctly, it would have already cought the bug. :)

cebe:
Merged with some adjustments. Thank you!

Issue #2022, opened 2014-01-17
Comparing operators like >,< >=,<= to search in GridView


indicalabs:
Yii1 Gridview has a feature to search with comparing operators like >,< >=,<= in GridView,

I can't see they are working now with gii generated code?

mklemarczyk:
:+1: What happen with `compare` function from `CDbCriteria` class form Yii 1.x?
I need this feature because a lot of my users use it.

Notice that:
I see some potential problem with this feature in Yii 2.0, because search attributes must be valid.
For example we can have `quantity` attribute with rule `integer` and expression `>10` will be invalid.

ilyapoz:
Hey, guys. I wonder how do you envision implementing this? I will need this for rest api in my project and can dig and make a contribution, will just need a few pointers.
Just improving the code generated by gii to replace simple `andFilterWhere` with more elaborate ones?
Where do you suggest putting the parsing code? @qiangxue @samdark

cebe:
@ilyapoz we have no concrete ideas or plans on this. you could make a proposal for an implementation and we can discuss it.

lennartvdd:
Something like that? ^^  I'd find this feature very helpful too. Yii 1's CDbCriteria::compare() was very user friendly.

ilyapoz:
@lennartvdd I guess functions with the property "Note that when the value is empty, no comparison expression will be added to the search condition." have "filter" in their name in yii2 by convention.

lennartvdd:
ah alright! Makes sense!  how about `andFilterCompare()` ?

ilyapoz:
I guess it's time to open a PR :)

lennartvdd:
any updates on this issue?

ArekX:
@ilyapoz 
I would actually just extend `andFilterWhere`, `orFilterWhere`, `andWhere`, `orWhere`, etc...

To use currently available operator format `[operator, operand1, operand2, ...]` where `operator` would be an instance of an implementation of an abstract class `QueryComparator`, let's say `NumericQueryComparator`  which implements comparator functions like `compare($value, $withValue)`, in this case the comparing with operators which are in `$withValue`. Of course `QueryComparator` would at the very least extend `yii\base\Object` to allow configuration.

This way we would keep `QueryInterface` pure, and also allow for custom implementations of `QueryComparator` which people can use. And also it will allow for easier implementation for different DBMS which Yii uses.

[...]

PR #8505, opened 2015-05-21
Query helper method for filter values containing operators

lennartvdd:
Basic implementation proposal for yiisoft/yii2#2022

implementation in Gii CRUD generator in PR yiisoft/yii2-gii#43
implementation in MongoDB extensions in PR yiisoft/yii2-mongodb#24

Shortcomings/todo's:
- filtering on non-scalars (where IN statements) not supported
- Numeric properties with a corresponding validation rule will yield
  errors when they contain an alphanumeric operator. These properties'
  rules should be defined as 'string' or 'safe'.


framework/db/QueryTrait.php
```diff
@@ -375,4 +375,35 @@ public function offset($offset)
...
+    public function andFilterCompare($name, $value, $defaultOperator = '=')
+    {
+        $matches=[];
+        if (preg_match("/^(<>|>=|>|<=|<|=)/", $value, $matches)) {
+            $op = $matches[1];
+            $value = substr($value, strlen($op));
+        } else {
+            $op = $defaultOperator;
+        }
+        $this->andFilterWhere([$op, $name, $value]);
```
klimov-paul: This will fail for NOSQL queries such as `yii\elasticsearhc\Query` or `yii\mongodb\Query`

lennartvdd: True.. it's RDBMS specific. Would moving this method out of the generic QueryTrait to yii\db\Query suffice?

cebe: I think we could try to port it to the noSQL implementations.

lennartvdd:
@cebe Porting to the NoSQL implementations is a good idea. I've made an attempt in yiisoft/yii2-mongodb#24.
Can you review it? It seems to work, but I've had to add support for arbitrary operators in the conditionBuilders. My experience with MongoDB is limited, so i'm not sure if this might break anything or cause security risks.

githubjeka:
Please see my some enhancement https://github.com/githubjeka/yii2/commit/cf8b0ae07e2757b5c41ecc65938c8125ba981dea?diff=split

Probably you need a method `orFilterCompare()` also.

lennartvdd:
I think `orFilterCompare()` is not very usefull for the use case in GridView filtering for which this PR is intended, because results would be counterintuitive. Perhaps there is another use case for which it's more usefull?

acorncom:
@cebe Any updates on getting this merged in?  I can use an override for now (and use the code in the PR), but would love to see it in the framework! :+1:

cebe:
Sorry, I do not remember to have seen this one actually :) Set for 2.0.8 milestone, will try to review it for next release.

acorncom:
@cebe :+1: Seems like you must have seen it at one point, as it was `self-assigned`, but that was almost a year ago and you guys have been busy! :smile:

Faryshta:
can you add this to the guide? it will be easier to review this way.

Faryshta:
@lennartvdd can you add this to the guide? it will be easier to review that way

cebe:
merged, thank you! Keeping yiisoft/yii2#2022 open for documentation.

sr1234:
Does using numeric comparison operators work with decimal data types (mysql decimals)?  Would love to be able to filter for instance on employees with hourly rates greater than some number (e.g. >= 50.00).   Thanks, S.

lennartvdd:
Should. Did you try it?

sr1234:
I should probably clarify that I am using the Gii generated GridView.  Yes, tried it.  Gives error: "Hourly Rate Base must be a number."  I'm entering '>100.00' in the input for the field in the GridView search.

Faryshta:
@sr1234 you should check #14014 since thats  rules issue

sr1234:
@Faryshta Thank you :)

Issue #8496, opened 2015-05-20
onCondition added to where part of child query


yomexzo:
I shared the same problem on stackoverflow - http://stackoverflow.com/questions/30349913/oncondition-added-to-where-part-of-child-query

I noticed in v2.1, the onCondition is being added to the where condition -https://github.com/yiisoft/yii2/blob/2.1/framework/db/ActiveQuery.php#L202

I do not know why, but this is affecting me from being able to use my relation as below.

I am working on the RBAC module for a new project. I am not using Yii2's default because it is a SaaS system, roles have to be defined per `account`, we would rather use `id` instead of name as unique keys and we do not like the naming conventions.

I have tables `permissions`, `privileges` and `roles`

`privileges - [id, name]` - things you can do in the system
`roles - [account_id, id, name]` - different roles per account
`permissions - [parent_type, parent_id, child_type, child_id]` - mapping of roles to privileges, roles to roles, and privileges to privileges.

Problem now is i need to `findAll` `permissions` and get the child / parent having the appropriate objects.

For parent, I have 

    public function getParentPrivilege($type) {
        return $this->hasOne(Privilege::className(), ['id' => 'parent_id'])->onCondition('parent_type = "privilege"');
    }
    
    public function getParentRole() {
        return $this->hasOne(Role::className(), ['id' => 'parent_id'])->onCondition('parent_type = "role"');
    }

    public function getParent(){
        if($this->parentPrivilege)
            return $this->parentPrivilege;
        else
            return $this->parentRole;
    }

Then i do a find by:

    $permissions = \app\models\Permission::find()
        ->joinWith(['parentRole','parentPrivilege'])
        ->all();

Problem is, `onCondition` is used in the SQL on condition since I used `joinWith` BUT, the `onCondition` is still being added to the `where` clause of the parent queries as seen below.

	SELECT `permissions`.* FROM `permissions` 
	LEFT JOIN `roles` ON (`permissions`.`parent_id` = `roles`.`id`) AND (parent_type = "role") 
	LEFT JOIN `privileges` ON (`permissions`.`parent_id` = `privileges`.`id`) AND (parent_type = "privilege")

	SELECT * FROM `roles` WHERE (`id` IN ('3', '1', '4', '2', '25')) AND (parent_type = "role")
	SELECT * FROM `privileges` WHERE (`id` IN ('5')) AND (parent_type = "privilege")

WHY?!!!

[...]

PR #8507, opened 2015-05-21
Fixed issue with adding "onCondition" to "where".

yomexzo:
Developers should know when the want to add a `onCondition` different from when they want to add a `where` clause

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -201,10 +201,6 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             $this->where = $where;
         }
 
-        if (!empty($this->on)) {
-            $query->andWhere($this->on);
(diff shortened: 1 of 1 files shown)


cebe:
this is correct according to the documentation, `onCondition` is meant to be used in the relation definiotion.

> The condition will be used in the ON part when yii\db\ActiveQuery::joinWith() is called. Otherwise, the condition will be used in the WHERE part of a query.
> http://www.yiiframework.com/doc-2.0/yii-db-activequery.html#onCondition%28%29-detail

yomexzo:
@cebe, you probably should have tried with the `joinWith` before closing this pull request.

Of course I did with `joinWith` and still had it in the `where` condition.

See https://github.com/yiisoft/yii2/issues/8496

PR #8515, opened 2015-05-21
resolves #8483: properly quote sequence name in Schema::getLastInsertId()

### LINKED ISSUE #8483: yii\db\Schema::getLastInsertID fails with schema.sequence name in postgresql
Hi all, after a framework update I receive this error from postgresql:

PDOException: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "admin.user_id_seq" does not exist    // see the way the sequence name is quoted

Since this fix https://github.com/yiisoft/yii2/commit/4f49feee8d2b0f0973ae1d0f29d66dccf9f06605 the sequence name is quoted with yii\db\Schema::quoteSimpleTableName.

Problem can be solved using yii\db\Schema::quoteTableName but I don't know what would happen with other dbms. Maybe this could be done in a new yii\db\pgsql\Schema::getLastInsertID method?

Note: the yii\db\pgsql\Schema::quoteSimpleTableName() override its parent's method but the code is the same. Maybe by choice.

Thanks.

nineinchnick:
Resolves #8483: properly quote sequence name in Schema::getLastInsertId().

### DIFF
--- a/framework/db/Schema.php
+++ b/framework/db/Schema.php
@@ -348,7 +348,7 @@ abstract class Schema extends Object
     public function getLastInsertID($sequenceName = '')
     {
         if ($this->db->isActive) {
-            return $this->db->pdo->lastInsertId($sequenceName === '' ? null : $this->quoteSimpleTableName($sequenceName));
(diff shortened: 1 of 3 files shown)


cebe:
Thank you!

Issue #8508, opened 2015-05-21
"Attribute name must contain word characters only" when using non-english characters in GridView


amlopezalonso:
MySQL allows the use of non-english characters in column names but attributes in GridView don't. For instance, using *año* (spanish for *year*) makes Yii 2.0.4 to spit an *Attribute name must contain word characters only* error.

cebe:
whats the point in using non-ASCII characters in column names? This makes many places in the implementation very difficult. Not sure we really want to support this...

amlopezalonso:
I agree this is not quite a critical issue but as long as MySQL supports it I just designed my schema to take advantage of it. Moreover, gii had no trouble with it and when I thought everything was about to be up and running I faced this issue which made me go back and change my db schema.

Fair enough for me If you don't want to support it but IMHO gii should  warn the developer at scaffolding stage instead of going further as if everything was ok.

amlopezalonso:
Using non-ASCII characters may be useful in the case of some words which cannot be comfortably replaced. Most prominent example in spanish is "año" which is often replaced by "anyo", "anno" or "ano" (the latter being the spanish translation of "anus", so...).

yurii-github:
just for fun. it is possible though.

	class test
	{
		private $props=[];
		
		function __get($n)
		{
			if (array_key_exists($n, $this->props)) {
				return $n . ' : ' . $this->props[$n] . ' : get'. '<br/>';
			}
			//error
		}
		
		function __set($n,$v)
		{
			$this->props[$n] = $v . ' : set ';
		}
	}
	
	
	$t = new test();
	$t->p1 = 'val1';
	echo $t->p1;
	
	$t->año = 'val2'; // spanish
	echo $t->año;
	
	$t->{'año año'} = 'val2'; // spanish . we can do even like this
	echo $t->{'año año'};
	
	$t->рік = 'val3'; // ukrainian
	echo $t->рік;
	
	// ukrainian vars
	$тест = new test();
	$тест->рік = '2015'; 
	echo $тест->рік;

yurii-github:
this code

		$ar = Test::find()->select(['*'])->one();
		$f = ActiveForm::begin();
		echo $f->field($ar, 'рік')->label('Your Name');
		ActiveForm::end();

will produce

	<form id="w0" action="/mylibrary-yii2/public/" method="post"><div class="form-group field-test-���">
	<label class="control-label" for="test-���">Your Name</label>
[...]

PR #4446, opened 2014-07-25
[WIP] #4408 unicode characters in attribute names

kmindi:
This PR is WorkInProgress for the issue: #4408
The request is to support different names for attributes if the columns in the database contain special characters like german umlauts.
Those umlauts lead to an exception in activeTextInput and DetailView::widget "Attribute name must contain word characters only in vendor\yiisoft\yii2\helpers\BaseHtml.php"

---
- [ ] Does yii2 want to support non ASCII-characters in column names/attributes?
  - @samdark: "It's not the best practice to use umlauts in code"
  - @cebe: "we already accept them in gridview":
    There the regular expression accepts any character except ':' for the attribute name. That means "..." is a valid attribute name.
- [ ] Discuss valid attribute names (column names of database can contain Unicode characters?):
  - [MySQL](http://dev.mysql.com/doc/refman/5.0/en/identifiers.html) states that "Permitted characters in quoted identifiers include the full Unicode Basic Multilingual Plane (BMP)"
- [ ] Discuss impact of attribute names in other classes than BaseHtml and DetailView
- [ ] Derive regular expression for valid attribute names
- [ ] Implement regular expression
- [ ] Implement tests

PR #8529, opened 2015-05-23
unicode in parameter names

yurii-github:
for https://github.com/yiisoft/yii2/issues/8508
just to show what i mean

---

<b>similar</b>
#4446 (will fail if source is not in utf8)

---

<b>notes</b>
- htmlspecialchars hit a limit of available charsets for code files
- code becomes very heavy on check if other charset than utf-8  is supported
- gives ability to use url with non ascii chars like  index.php?єжыъэ=эъхж , in fact, it's a must in some cases when non ascii attribute is translated into query string via GET

---

<b>current</b>

```
// Yii::$app->charset = 'cp1251', project files are encoded in cp1251 accordingly     
$f = ActiveForm::begin();
echo $f->field($ar, 'рік');
ActiveForm::end();
```

output looks like

```
<form id="w0" action=".." method="post"><div class="form-group field-test-рік">
<label class="control-label" for="test-рік">Рік</label>
<input type="text" id="test-рік" class="form-control" name="Test[рік]" value="2010">

<div class="help-block"></div>
</div></form>
```

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -96,7 +96,7 @@ class BaseHtml
      * @param unknown $utfString
      * @return unknown|string
      */
-    protected static function convertToEncoding($utfString)
+    public static function convertToEncoding($utfString)
     {
     	$charset = strtoupper(Yii::$app->charset);
     	if ($charset === 'UTF-8') { //nothing to do
@@ -111,7 +111,7 @@ class BaseHtml
      * @param unknown $encString
      * @return unknown|string
      */
-    protected static function convertToUTF($encString)
+    public static function convertToUTF($encString)
     {
         $charset = strtoupper(Yii::$app->charset);
         if ($charset === 'UTF-8') { //nothing to do
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1940,7 +1940,7 @@ class BaseHtml
      */
     public static function getAttributeName($attribute)
     {
-        if (preg_match('/(^|.*\])([\w\.]+)(\[.*|$)/', $attribute, $matches)) {
+        if (preg_match('/(^|.*\])([\w\.]+)(\[.*|$)/u', $attribute, $matches)) {
             return $matches[2];
         } else {
             throw new InvalidParamException('Attribute name must contain word characters only.');
@@ -1963,7 +1963,7 @@ class BaseHtml
      */
     public static function getAttributeValue($model, $attribute)
     {
-        if (!preg_match('/(^|.*\])([\w\.]+)(\[.*|$)/', $attribute, $matches)) {
+        if (!preg_match('/(^|.*\])([\w\.]+)(\[.*|$)/u', $attribute, $matches)) {
             throw new InvalidParamException('Attribute name must contain word characters only.');
         }
         $attribute = $matches[2];
@@ -2013,7 +2013,7 @@ class BaseHtml
     public static function getInputName($model, $attribute)
     {
         $formName = $model->formName();
-        if (!preg_match('/(^|.*\])([\w\.]+)(\[.*|$)/', $attribute, $matches)) {
+        if (!preg_match('/(^|.*\])([\w\.]+)(\[.*|$)/u', $attribute, $matches)) {
             throw new InvalidParamException('Attribute name must contain word characters only.');
         }
         $prefix = $matches[1];
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -89,6 +89,37 @@ class BaseHtml
      */
     public static $dataAttributes = ['data', 'data-ng', 'ng'];
 
+    
+    /**
+     * converts UTF-8 encoded string to app charset
+     * if string is UTF8 function returns it w/o modification
+     * @param unknown $utfString
+     * @return unknown|string
+     */
+    protected static function convertToEncoding($utfString)
+    {
+    	$charset = strtoupper(Yii::$app->charset);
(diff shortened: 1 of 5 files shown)


framework/helpers/BaseHtml.php
```diff
@@ -2040,7 +2040,7 @@ public static function getInputName($model, $attribute)
      */
     public static function getInputId($model, $attribute)
     {
-        $name = strtolower(static::getInputName($model, $attribute));
+        $name = mb_strtolower(static::getInputName($model, $attribute));
```
samdark: Encoding should be passed when using `mb_`  i.e. `mb_strtolower($string, Yii::$app->charset);`.

framework/helpers/BaseInflector.php
```diff
@@ -317,7 +317,7 @@ public static function camelize($word)
      */
     public static function camel2words($name, $ucwords = true)
     {
-        $label = trim(strtolower(str_replace([
+        $label = trim(mb_strtolower(str_replace([
```
samdark: Same here.

yurii-github: yes. you are right. but tests have not failed

framework/helpers/BaseInflector.php
```diff
@@ -317,13 +317,12 @@ public static function camelize($word)
...
     public static function camel2words($name, $ucwords = true)
     {
-        $label = trim(strtolower(str_replace([
-            '-',
-            '_',
-            '.'
-        ], ' ', preg_replace('/(?<![A-Z])[A-Z]/', ' \0', $name))));
-
-        return $ucwords ? ucwords($label) : $label;
+        $label = preg_replace('/(?<![A-Z])[A-Z]/', ' \0', $name); // add spaces TODO: unicode?
```
samdark: Yes, this one is not unicode-aware...

yurii-github:
i managed to get it going, but i'm not sure it worth the candles.
such approach requires dependency on app object which is not good for static method.

yurii-github:
i looked a bit at code more. does this 'Yii::charset" do anything at all except sending useless contentType header?

imagine example: my code base is in cp1251, yii2 in utf8. i set Yii::app->charset = 'cp1251', db connection to cp1251 too. now yii2 - dies or lives?

i tried it recently on cp1251 code files. in config i put charset to cp1251.

     echo yii\helpers\Html::encode('фівїхфжівї');
     echo '<br>';
     echo 'фівїхфжівї';

output was with right cp1251 encoding, but Html encode failed

     пїЅпїЅпїЅпїЅпїЅ
     фівїхфжівї

how this charset option is supposed to work?

regards

yurii-github:
it has huge chain.. it looks like better to abandon this

samdark:
Hmm... indeed it looks too complicated. Let's leave this pull request open for a while. Need to think about it.

yurii-github:
alright. whenever you please. i'll be pushing some from time to time with a study purpose.

yurii-github:
i have a question about code -  $this->_queryParams is never reached in Request:: getQueryParams() because this var is never set. also  this class is 0% covered with tests.

what's the purpose of method Request::setQueryParams() ?

klimov-paul:
`Request::setQueryParams()` can be used in unit tests.

framework/helpers/BaseHtml.php
```diff
@@ -89,6 +89,35 @@ class BaseHtml
...
+     * @return unknown|string
+     */
+    public static function convertToUTF($encString)
+    {
+        if (strcasecmp(Yii::$app->charset, 'UTF-8') === 0) { //nothing to do
+            return $encString;
+        }
+        return mb_convert_encoding($encString, 'UTF-8', Yii::$app->charset);
+    }
+
```
cebe: why do we need these encoding conversions? shouldn't the whole applicaion shere the same encoding?

yurii-github: if file encode is different from utf8
1 - preg_match with /u expects unicode
2 - $_GET parses data from url as unicode because url is in utf8

unfortunately, i don't know how to preg_match non unicode and unicode without conversion

yurii-github:
klimov-paul,
why do so? PHPUnit resets super globals on each test

framework/helpers/BaseHtml.php
cebe: every file in the project should be utf8 encoded. if not we would go crazy try to catch all the problems this causes.

yurii-github: hm
what's the point in Yii::app->charset? it must be forced to utf8 then. or i don't get something

cebe: if you change the charset, you should change it overall. So there still should be no need to convert anything.

yurii-github: i don't understand what "if you change the charset, you should change it overall" means

here's sample code which will fail (and that's right): 
yii charset in config = windows-1251 (Cyrillic)  / 'charset' => 'windows-1251'
code files are encoded in utf-8
sample layout outputs meta as <meta charset="<?= Yii::$app->charset ?>">
we want output 
echo 'year : рік'; 

when you open in IE you will see 
year : СЂС–Рє 

regards

cebe: why would you set application charset to windows-1251 and have the files in utf-8?

yurii-github: to show you that setting yii charset other than utf-8 fails unless we convert project files into set encoding (windows-1251 etc). and doing so will fail my modification above if i remove string conversions to utf-8 and vice versa.

yurii-github:
this pull doesnt make any sense so far. closing

regards

PR #8569, opened 2015-05-27
Trigger garbage collection in FileCache

kidol:
gc() not triggered at all for now.


cebe:
good find! thanks!

klimov-paul:
Are you sure `gc` should be invoked at `setValue()`?
It has a probability. I assume it should be triggered at `init()`.

kidol:
Yes I first had it at init, but DbCache has it at set as well and it makes more sense because when you get, you don't care about gc at all.
In yii1 there was property $_gced that made sure automatic gc is guaranteed to only run once. Maybe we can add in future. This pr is basically just to fix the bug.

cebe:
Had the same thought about `init()` but this change complies with the docs:

> ... when storing a piece of data in the cache.
https://github.com/yiisoft/yii2/blob/88371fb147dc68e8eaa728fbe8fe262006ab64a8/framework/caching/FileCache.php#L53

PR #8580, opened 2015-05-27
Fixes #8537 - Cookie reuse attack

### LINKED ISSUE #8537: Cookie Reuse attack vulnerability
Hello,
Apparently yii2's session manager is vulnerable to Cookie Reuse attacks (type of session hijacking).

Steps to repeat:
- Login as a user  
- Copy the content of all cookies (can be done easily with the 'edit this cookie' chrome extension, export cookies)
- Logout
- Import the previously exported cookies
- Refresh the page -- Now you are logged in without password.

I tried this with pasting the cookies on a different computer in the same network and it worked (didn't try for another ip though)

Desired effect: When I logout the cookies saved should not work

Note: I only tried this with the default PHP session manager. Didn't try the db or cache ones.

kidol:
Fixes #8537

Added integer support to $autoRenewCookie as well (since we can now check by timestamp). Yii by default sends the cookie on **every** request when $autoRenewCookie is set to true. No need for that. See doc comments for details.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -20,9 +20,11 @@ Yii Framework 2 Change Log
 - Enh #8070: `yii\console\controllers\MessageController` now sorts created messages, even if there is no new one, while saving to PHP file (klimov-paul)
 - Enh #8286: `yii\console\controllers\MessageController` improved allowing extraction of nested translator calls (klimov-paul)
 - Enh #8415: `yii\helpers\Html` allows correct rendering of conditional comments containing `!IE` (salaros, klimov-paul)
-- Enh #8444:  Added `yii\widgets\LinkPager::$linkOptions` to allow configuring HTML attributes of the `a` tags (zinzinday)
-- Enh #8486:  Added support to automatically set the `maxlength` attribute for `Html::activeTextArea()` and `Html::activePassword()` (klimov-paul)
-- Enh #8566:  Added support for 'only' and 'except' options for `yii\web\AssetManager::publish()` (klimov-paul)
+- Enh #8444: Added `yii\widgets\LinkPager::$linkOptions` to allow configuring HTML attributes of the `a` tags (zinzinday)
+- Enh #8486: Added support to automatically set the `maxlength` attribute for `Html::activeTextArea()` and `Html::activePassword()` (klimov-paul)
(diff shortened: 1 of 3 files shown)


samdark:
Won't the change give attacker some time before cookie is renewed?

ali-alamine:
Since the timestamp is saved in the cookie, this solution won't work because cookies are saved as plain text, hense they can be easily tampered with.

klimov-paul:
Timestamp expiration can be used. But it requires extra protection like OAuth HMAC signature does.

In particular:

__While saving auto loging cookie:__
1) We have `$id`, `$authKey`, `$duration`, `$timestamp`.
2) We contact them via separator like '&' or '.'
3) We add some secret key to the concated string, so it value becomes:
```php
$baseString = "$id&$authKey&$duration&$timestamp&$secretKey"
```
4) We calculate hash from this `$baseString` using `sha1()` for instance:
$hash = sha1($baseString)
5) We place calculated hash into source array for cookie JSON value.
6) Save the cookie

__While auto login:__
1) Retrieve the cookie
2) JSON decode
3) We have: `$id`, `$authKey`, `$duration`, `$timestamp` and `$hash`(!)
4) Calculate new `$hash` based on `$id`, `$authKey`, `$duration`, `$timestamp` we got like we did while savin cookie
5) Compare calculated hash with the one we got: if they are equal - login is allowed, if not - login fails.

kidol:
@klimov-paul But cookies are already hmac-hashed. Look this is cookie from basic app template with this PR applied 
cookie

klimov-paul:
That is so only if `Request::enableCookieValidation` is enabled.

klimov-paul:
We can of course rely on `Request::enableCookieValidation` is enabled in this case. But it is better to be ckecked, with exception is thrown on failure perhaps. At least docs should be updated warning the developer he need to enabled `Request::enableCookieValidation` in order to make autoLogin secure.

kidol:
I think throwing exception is best. If users have problems because they need cleartext cookies for some reason, then there is always the option to add a property to Request which holds a list of cookie names that should NOT be modified by the app.
In Yii1 however, I don't remember anyone complaining that cookies are hashed with hmac
Well I will make PR.

kidol:
Most likely outdated, also no interest in working further on this.

PR #8594, opened 2015-05-28
 Auto convert AR arrays to Postgres arrays

YevSent:
Added encoders/decoders from php arrays to postgres arrays.
For now works with `integer[]`, `text[]` and `numeric[]` arrays.
Related to #7911.


framework/db/pgsql/EncoderTrait.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Convert php string array to postgres text array
+     *
+     * @access public
+     * @param mixed $value
+     * @return string
+     */
+    public function arrayTextEncode($value)
+    {
+        return '{'.implode(',', (array)$value).'}';
```
klimov-paul: Why there is no value escaping?
What if some element contains '}' or other special chars?

joni-jones: Need to add more units tests and recheck.

klimov-paul: Also what about `null` value?

klimov-paul:
At the first gralce this does not seem right.
I assume array values should be saved using param binding - not harcoded strings.
I have implemented similar functionality for the Sphinx extension to cover MVA (multi value attribute):
https://github.com/yiisoft/yii2-sphinx/blob/master/QueryBuilder.php#L186

I suppose this feature should be implemented in more or less similar way.

framework/db/pgsql/EncoderTrait.php
joni-jones: if `value` is not it will not be converted.

klimov-paul: So Postgres array type columns does NOT allow storing `null` value at all?

nineinchnick:
~~~sql
SELECT ARRAY[1,2,null,3]
~~~
gives: array integer[] `{1,2,NULL,3}`

joni-jones: `null` is allowed is value. But I'm not see anywhere `null` as default value. I've see only `{}` as default value.

klimov-paul: As I see from DDL attached to this PR:
```sql
related integer[] NOT NULL DEFAULT '{}'::integer[],
```
This indicates that entire column value can be `null`, otherwise  there is not reason to say it is 'NOT NULL'.
This means there IS a difference between `null` value and `{}` value.
So while inserting or updating there should be an ability to set `null` value to column.

nineinchnick: You could try to decode that using [str_fgetcsv](http://php.net/manual/en/function.str-getcsv.php) after stripping the brackets.

joni-jones: @klimov-paul. I've removed default value and `not null` from `tags` column. I re-run tests with empty and `null` values for `tags` attributes. All tests are passed.
If not specify `tags` attribute - column will be empty.
If set `null` - column will be contain empty postgres array.

klimov-paul: Are you listen or not?
Current impelemtation does NOT allow explict set of `null` value to the column, which developer may require.
```php
$command->update('mytable', ['array_field' => null], 'id=1')->execute();
```
produces:
```sql
UPDATE ... SET (array_field, {})
```
while it should produce:
```sql
UPDATE ... SET (array_field, NULL)
```

joni-jones:
Moved to new PR #8823.

PR #8596, opened 2015-05-28
added missing use statement to fix #8572

### LINKED ISSUE #8572: Strange issue with TimestampBehavior
UpdatedAt works well after any moficiation in row. But, I can't say the same for row creation. 
After new row creation `createdAt` still 0000-...
      
    public function behaviors()
      {
        return [
            [
                'class' => TimestampBehavior::className(),
                'createdAtAttribute' => 'createdAt',
                'updatedAtAttribute' => 'updatedAt',
                'value' => new Expression('NOW()')
            ],
          ];
      }

Here is database table: http://prntscr.com/79xt5n

And here is model: https://gist.github.com/turalsprofile/85eca5d307074b42a215

What am I doing wrong?

cebe:
- added unit tests for timestamp behavior and insert with expression
- fixed issue #8572 

tests may not pass on postgres, waiting for travis so I can fix it.

problem has been introduced by 6e48ed0 (#8034, 2.0.5) so no changelog is needed.

/cc @nineinchnick

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -9,6 +9,7 @@ namespace yii\db;
 
 use yii\base\InvalidParamException;
 use yii\base\NotSupportedException;
+use yii\db\Expression;
 
 /**
  * QueryBuilder builds a SELECT SQL statement based on the specification given as a [[Query]] object.
--- a/framework/db/mysql/QueryBuilder.php
+++ b/framework/db/mysql/QueryBuilder.php
@@ -7,8 +7,9 @@
 
 namespace yii\db\mysql;
 
-use yii\db\Exception;
 use yii\base\InvalidParamException;
+use yii\db\Exception;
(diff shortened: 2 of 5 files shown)


cebe:
memo to myself:

- [ ] add tests for batchInsert too.

framework/db/QueryBuilder.php
```diff
@@ -9,6 +9,7 @@
 
 use yii\base\InvalidParamException;
 use yii\base\NotSupportedException;
+use yii\db\Expression;
```
klimov-paul: Why?
`yii\db\QueryBuilder` and `yii\db\Expression` are under the same namespace.

cebe: good point, for some reson my PHPStorm complained here...

cebe: fixed and merged. aa5887f

PR #8604, opened 2015-05-29
added strict option for camel2words

schmunk42:
see https://github.com/yiisoft/yii2/pull/4166

### DIFF
--- a/framework/helpers/BaseInflector.php
+++ b/framework/helpers/BaseInflector.php
@@ -313,15 +313,17 @@ class BaseInflector
      * For example, 'PostTag' will be converted to 'Post Tag'.
      * @param string $name the string to be converted
      * @param boolean $ucwords whether to capitalize the first letter in each word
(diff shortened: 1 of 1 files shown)


framework/helpers/BaseInflector.php
```diff
@@ -313,15 +313,17 @@ public static function camelize($word)
      * For example, 'PostTag' will be converted to 'Post Tag'.
      * @param string $name the string to be converted
      * @param boolean $ucwords whether to capitalize the first letter in each word
+     * @param boolean|string $strict whether to create a space between two consecutive uppercase chars, defaults to false
      * @return string the resulting words
      */
-    public static function camel2words($name, $ucwords = true)
+    public static function camel2words($name, $ucwords = true, $strict = false)
```
samdark: That's backwards compatibility break thus it can be merged into 2.1.x only.

Issue #8049, opened 2015-04-12
how to pass variables to view dynamically?


tlanyan:
Hi, 
I want to pass some specific parameters(such as static base url) to all the views. After searching apis, I find that only the method yii\web\View::beforeRender can meet my requirement. I use $this->params[key] = value in the beforeRender to pass values to all view. But the passed variables can only be accessed in the form of $this->params[key].

I want the passed variables can be available in the direct form: $key. I notice that the parameters of beforeRender is $viewFile and $params.  The second parameter $params is the local variables available among view files. However $params is declared as passed by value, so I can't use $params[key]=value to add variables dynamically.

Is there another api which can meet my need? Many thanks!

mdmunir:
you can not

samdark:
Correct. There's no such feature. What's your use case?

kmergen:
If you need a static variable in all your views you can extend ```yii\web\view```

cebe:
> But the passed variables can only be accessed in the form of $this->params[key].
> 
> I want the passed variables can be available in the direct form: $key.

why do you need it this way? You can use http://php.net/manual/en/function.extract.php to turn params into variables.

tlanyan:
For example, the base url of static files should passed to all the view. The view files are written in Smarty. I just want all the passed parameters are the same for the front end. I know there is an api called View::setVar in the Phalcon framework. You can always use it to pass variables, and the passed variables are the same with call render

cebe:
I still do not see why `$this->params` does not work, you can even configure it in the application config...

tlanyan:
The problem is not $this->params does not work(actually it works well in my project). The problem is that front end must use {$this->params.key} to get the value in the smarty syntax and it's weird for the front end. 

What I want is that is there an approach to pass variables to the view just like passed in the View::render method?

kmergen:
How I can understand you have a few variables that you need in all your views. And therefore I would extend the the yiew class with all your needed variables and than you can set them in your controller or view and use them in your view with ```$this->variable```

tlanyan:
I wish these variables can be added or removed dynamically in the configure file, rather than they are property of some class...

While it's ok that view class does not support this feature...

samdark:
I don't think we're going to support it. That's not really common request and I doubt it will be used often.

eacevedof:
To solve this i have extended base/Controller:  

**namespace common\masters;
use Yii;
use yii\web\Controller;
class CommonController extends Controller**
{
[...]

PR #8611, opened 2015-05-31
add setVar function for the View object.

tlanyan:
User can call `$this->setVar($key, $value)` within the view context or
`$view->setVar($key, $value)` anywhere with the view object, and the `$key` and `$value` can be accessed in the view file by `$key`.
This feature is very useful in the view::beforeRender function, such as
pass the site url to all views:

```
public function beforeRender($viewFile, $params)
{
    if (parent::beforeRender($viewFile, $params)) {
        $this->setVar('host', Yii::$app->params['host']);
        // other codes
        return true;
    }
    return false;
}
```

With the setVar function, you don't need to pass the url in the action
using `$this->render('view files', ['host'=>'xxx.com'])` any more.

### DIFF
--- a/framework/base/View.php
+++ b/framework/base/View.php
@@ -52,6 +52,10 @@ class View extends Component
      */
     public $params = [];
     /**
+     * @var array variables that can be put into the views context dynamically
+     */
+    private $_vars = [];
+    /**
      * @var array a list of available renderers indexed by their corresponding supported file extensions.
      * Each renderer may be a view renderer object or the configuration for creating the renderer object.
      * For example, the following configuration enables both Smarty and Twig view renderers:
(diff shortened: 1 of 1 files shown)


samdark:

I think it's OK to add it.

klimov-paul:
Why don't use `View::params`?

klimov-paul:
We have already discussed this at #8049

samdark:
Yes. This time there's valid use case.

cebe:
> $this->setVar('host', Yii::$app->params['host']);

why not access `Yii::$app->params['host']` where you need it? use case does not look different from #8049.

I am closing this PR. Thank you for your contribution!

PR #8614, opened 2015-05-31
Fixes #8613: FragmentCache storing empty content

### LINKED ISSUE #8613: FragmentCache storing empty content causes problem with PageCache
Assuming PageCache is enabled for an action:

- In case of an exception, `ob_get_clean()` in `FragmentCache::run()` returns `false` (since the error handler cleared it), but it will still be cached. Usually not a problem because `getCachedContent()` returns `false` if content is `false`. But if we had dynamic content in the view, `getCachedContent()` will return an empty string instead (because of call to `updateDynamicPlaceholders()`). This means on next page load we get a white page.

Fix: Check whether `false` is returned and just `return;` in this case: https://github.com/yiisoft/yii2/blob/master/framework/widgets/FragmentCache.php#L106

- When doing `return $this->redirect(...);`, `ob_get_clean()` in `FragmentCache::run()` will return an empty string which will be stored in cache. If `PageCache::$cacheHeaders` is enabled, this is not a problem, because we get a redirect instead of white page. If `$cacheHeaders` is disabled though, we get a white page again.

Fix A: Same as fix on top, but instead also check for empty string.
Fix B: Clear output buffer in `Response::redirect()` so that `ob_get_clean()` in `FragmentCache` returns `false` instead of empty string.

A or B? I tend to A, but it would mean caching empty strings with fragment cache is not possible anymore. Also caching redirects would not be possible anymore in either case, but it's not expected behaviour anyway imo. Let me know so I can make PR.

Another issue is that page cache will store headers even if fragment cache did not store anything. This is part of bigger problem (page cache does not know about fragment cache). Not harmful though, just unnecessary. Related: https://github.com/yiisoft/yii2/issues/545#issuecomment-19526209

kidol:
Fixes #8613

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -29,6 +29,7 @@ Yii Framework 2 Change Log
 - Enh #8486: Added support to automatically set the `maxlength` attribute for `Html::activeTextArea()` and `Html::activePassword()` (klimov-paul)
 - Enh #8566: Added support for 'only' and 'except' options for `yii\web\AssetManager::publish()` (klimov-paul)
 - Enh #8574: Added `yii\console\controllers\MessageController` support .pot file creation (pgaultier)
+- Enh #8613: `yii\widgets\FragmentCache` will not store empty content anymore which fixes some problems related to `yii\filters\PageCache` (kidol)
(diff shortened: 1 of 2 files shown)


samdark:
Thank you!

PR #8626, opened 2015-06-01
Fixes #6254: CORS response for Preflight Request

### LINKED ISSUE #6254: CORS Filter and authenticator s bug
## the authenticator will broken the CORS Filter when did the ajax request with rest api.

the code:
```

    public function behaviors()
    {
        $behaviors = ArrayHelper::merge(parent::behaviors(), [
            'corsFilter' => [
                'class' => \yii\filters\Cors::className(),
            ]
        ]);

            'class' => CompositeAuth::className(),
            'authMethods' => 
                  HttpBasicAuth::className(),
                  HttpBearerAuth::className(),
                  QueryParamAuth::className(),
                ]
            ];
    
        return $behaviors;
    }
```

and then do the ajax call with Bearer auth, if the access token match, everything is working nice.
if the access token is wrong, the browser like Chrome will show error and return a canceled status (it means the ajax can not get the response json data):
XMLHttpRequest cannot load Origin “xxx” is not allowed by Access-Control-Allow-Origin.

the bugs was in the code 'beforeAction' of the classes Http****Auth, it will return false or null when the access token was wrong. but the return false or null also stop the CORS Filter class, the header 'Access-Control-Allow-Origin' will not sent. (maybe the invoked order issue of the external actions).

the bug was shown under yii2 RC.

aheart:
This is an enhancement for yii\filters\Cors that fixes #6254 

For Some CORS requests (e.g. Content-Type: application/json) the browser first sends a [Preflight Request](http://www.html5rocks.com/en/tutorials/cors/) (HTTP method: OPTIONS) and after that the actual request.
If an authenticator action filter is used the filter throws an exception because the Preflight Request doesn't contain an access token or any data for that matter.
without preflight response
The proposed fix handles the Preflight Requests and stops the execution of the Action resulting in normal operation.
preflight requests handled
For this to work the authenticator filter must come after the corsFilter otherwise the authenticator will throw an exception.

``` PHP
class ApiController extends yii\rest\Controller
{
    /**
     * @inheritdoc
     */
    public function behaviors()
    {
        return [
            'corsFilter' => [
                'class' => \yii\filters\Cors::className(),
                'cors' => [
                    'Origin' => ['*'],
                    'Access-Control-Request-Method' => ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
                    'Access-Control-Request-Headers' => ['Origin', 'X-Requested-With', 'Content-Type', 'Accept', 'Authorization'],
                    'Access-Control-Allow-Credentials' => true,
                    'Access-Control-Max-Age' => 86400,
                    'Access-Control-Expose-Headers' => [],
                ],
            ],
            'authenticator' => [
                'class' => HttpBearerAuth::className(),
                'only' => ['dashboard'],
            ],
//...
```

This is useful for all Preflight Requests. Even those that are not blocked by the authenticator because these requests don't need a response from the action.

What do you think?

### DIFF
--- a/framework/filters/Cors.php
+++ b/framework/filters/Cors.php
@@ -76,6 +76,10 @@ class Cors extends ActionFilter
      */
     public $actions = [];
     /**
+     * @var boolean enable response for Preflight Request
+     */
+    public $preflightResponse = true;
+    /**
      * @var array Basic headers handled for the CORS requests.
      */
     public $cors = [
@@ -101,6 +105,10 @@ class Cors extends ActionFilter
         $requestCorsHeaders = $this->extractHeaders();
         $responseCorsHeaders = $this->prepareHeaders($requestCorsHeaders);
         $this->addCorsHeaders($this->response, $responseCorsHeaders);
+
+        if ($this->preflightResponse && $this->request->method === 'OPTIONS') {
+            return false;
+        }
 
         return true;
     }


AlanBreck:
I have my CORS configuration defined prior to my authentication, but I'm still getting blocked at pre-flight. Here's my `behaviors` method:

```php
public function behaviors() {
	$behaviors = parent::behaviors();

	$content_negotiator = $behaviors['contentNegotiator'];
	unset($behaviors['contentNegotiator']);
	$content_negotiator['formats'] = Yii::$app->params['formats'];

	return ArrayHelper::merge(
			[
					'corsFilter' => [
							'class' => Cors::className(),
							'cors' => [
									'Origin' => ['*'],
									'Access-Control-Expose-Headers' => [
											'X-Pagination-Per-Page',
											'X-Pagination-Total-Count',
											'X-Pagination-Current-Page',
											'X-Pagination-Page-Count',
									],
							],
					],
					'authenticator' => [
							'class' => CompositeAuth::className(),
							'authMethods' => [
									['class' => HttpBearerAuth::className()],
									['class' => QueryParamAuth::className(), 'tokenParam' => 'access_token'],
							]
					],
					'exceptionFilter' => [
							'class' => ErrorToExceptionFilter::className()
					],
					'access' => [
							'class' => AccessControl::className(),
							'rules' => $this->accessRules(),
							'ruleConfig' => ['class' => 'app\components\AccessRule'],
					],
			],
			$behaviors
	);
}
```

aheart:
@AlanBreck Please verify that the corsFilter is placed before the authenticator in your $behaviors by dumping it. 
If I'm right it isn't because you're using $behaviors = parent::behaviors();
If the corsFilter is actually placed before the authenticator it should work.

Try defining all the behaviors the way I showed in the first post. Tell me if still won't behave as expected.

AlanBreck:
I've confirmed that the corsFilter precedes the authenticator, and tried using your exact corsFilter configuration, but it's still saying that the Access-Control-Allow-Origin header is missing.

aheart:
@AlanBreck I received an email notification about your reply, but I can't see you comment for some reason. I decided to investigate the issue and found out that I have the same problem. The code that was working in June stopped working. 
If I understand correctly it was broken by PHP 5.6
This part in particular http://php.net/manual/en/ini.core.php#ini.always-populate-raw-post-data

When I set `always_populate_raw_post_data = -1` in php.ini it started working again.

When I created the PR I was still using PHP 5.5

AlanBreck:
@aheart, thank you so much for both of your responses. I deleted my initial reply because I realized that I had failed to try something and didn't want to waste your time. :-) That said, though, I now have tried all the solutions I know to try and am still getting a preflight error.

I'm running PHP 5.4, so the `always_populate_raw_post_data` was commented out, but just to be sure, un-commented it, set it to `-1` and restarted my php service. Still nothing.

Any thoughts?

aheart:
@AlanBreck You're welcome. I'm glad someone has taken an interest in this PR.
I got carried away with my own problem yesterday and completely missed the fact that your problem relates to the preflight request and not the request that comes after that (which was my problem).

I wanted to ask you just in case: Are you using my branch of Yii?

If you are, then the only thing I can do is post the code that I was using to test this PR along with instructions.

samdark:
@aheart is it possible to convert that code into unit test somehow?

AlanBreck:
@aheart Ahh, I see. While I didn't actually use your branch, I did try manually inserting your changes into my copy of Yii, but that still didn't work.

aheart:
@samdark It should be trivial. All we need is to simulate an OPTIONS request.
All requests (GET, POST etc.) except OPTIONS should pass trough the CorsFilter.
post
Whereas an OPTIONS request should be stopped by CorsFilter from propagating further resulting in an empty response with appropriate headers.
options1

@AlanBreck I've uploaded an app that I used to test my PR.
For everything to work you need to run the commands as described in README.md

As a result you should have something along the lines of:
1) Test app running on http://localhost/frontend/web (I used apache) or any other url lik http://localhost/testapp/frontend/web
2) Api on http://localhost:8000 (built in PHP server) in this case the address is important, but you can change it if you want to in https://github.com/aheart/angularjs-yii2-part-2-authentication/blob/master/frontend/web/js/controllers.js

Go to http://localhost/frontend/web/#/login and try to log in (demo:demo)
If you'll have the chance to experiment with it tell me about your results.

Best of luck

tunecino:
@aheart I think this commit is not required. in your code just change this :
```php
'authenticator' => [
    'class' => HttpBearerAuth::className(),
    'only' => ['dashboard'],
],
```
to this : 
```php
1 => [
    'class' => HttpBearerAuth::className(),
    'only' => ['dashboard'],
],
```
or this : 
```php
[
    'class' => HttpBearerAuth::className(),
    'only' => ['dashboard'],
],
```

and it should work fine as OPTIONS verbs is excepted from the `authenticator`. I think it is an ordering issue relating to merging with `parent::behaviors()`. 

See [my comment here](https://github.com/yiisoft/yii2/issues/6254#issuecomment-163759832) for more details.

aheart:
@tunecino 
I've looked at the link you provided. This indeed should solve the problem
```
[
    'class' => \yii\filters\auth\HttpBearerAuth::className(),
    'except' => ['options'],
],
```
Thank you for taking the time to comment.

tunecino:
@aheart. I'm glad it helped and thanks for the feedback. I'll try to do a PR for it tonight. I think `corsFilter` should be added right before `authenticator` inside `yii\rest\Controller`([here](https://github.com/yiisoft/yii2/blob/4a3a94346d76e2aaac577065b159735c5b8db870/framework/rest/Controller.php#L46)) with its defaults configs defined [here](https://github.com/yiisoft/yii2/blob/4a3a94346d76e2aaac577065b159735c5b8db870/framework/filters/Cors.php#L81). this should solve the ordering issue when merging as described in docs (with named keys). Also adding `'except' => ['options']` to the `authenticator` should be mentioned in REST Authentication docs as most navigators are doing [preflighted requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests).

giuliogatto:
None of the solutions in here worked for me. Even if the CORS filter is activated no header is sent from the server. If I debug this::behavior() from the controller the filter is there, but the browser gets stuck at the OPTIONS preflight call because no header is sent back.

The only way I had CORS working in YII2 is by manually intercepting OPTIONS requests and ignoring them, then manually send back the correct headers to the browser.

    // ignore the preflight CORS request in the controller
    if(Yii::$app->getRequest()->getMethod() == 'OPTIONS')return;

And adding headers manually.

    $headers = Yii::$app->response->headers;
    $headers->add('Access-Control-Allow-Methods', 'GET');
    $headers->add('Access-Control-Allow-Origin', '*');
    $headers->add('Access-Control-Allow-Headers', 'Content-Type, Access-Control-Allow-Headers, Authorization, X-Requested-With');

jhebb:
So is the recommended solution to add

`'except' => ['options']`

to all Auth behaviors? Seems like this should be set as the default.

rockccf:
@giuliogatto 
I was having the same issue as yours and stuck for quite some time, and I figured out that because I'm calling a custom action (actionLogin), in the application component (web.php),
you have to explicitly set the OPTIONS method for your custom action to point to the default OptionsAction.

`'urlManager' => [
            'enablePrettyUrl' => true,
            'enableStrictParsing' => true,
            'showScriptName' => false,
            'rules' => [
                [
                    'controller' => 'user',
                    'extraPatterns' => [
                        'POST login' => 'login',
                        'OPTIONS login' => 'options'
                    ],
                ],
            ],
        ],`

giuliogatto:
@rockccf 
thank you for pointing that out. I'll give it a try!

doublehops:
@jhebb I have seen your example and many like it to configure your application to ignore auth on OPTIONS requests. I see this as a massive security issue. When doing this you open up the application to trust any requests that would otherwise require auth when made simply by making the request as OPTIONS rather than (for example) POST.

What's stopping someone from doing (another example) `curl -X OPTIONS --data "{ 'id': '1', 'password': 'newpassword' } http://domain.api/user/change-password` or adding/modifying/deleting any other records?

jhebb:
@doublehops This isn't my area of expertise, but I was under the impression yii\rest\UrlRule handled the routing of an OPTIONS request and you couldn't actually change any data with a request like that.

If there's a better way of doing this, I'm all for it! This was the only way I could get it working myself.

tunecino:
it is safe. any OPTIONS request will be simply redirected by `yii\rest\UrlRule` to `OptionsAction ` defined [here](https://github.com/yiisoft/yii2/blob/master/framework/rest/OptionsAction.php). it will only return the suppoted list of verbs.

doublehops:
@tunecino. You're right. I think this was an issue in the past but it must have been resolved in later version.

kiuka:
it seems I have the same issue.

I tried to change the order of behaviors (cors before authenticator), added OPTIONS rule to UrlRule's rules, also tried to change yii\rest\Controller by add cors before authenticator, but still not working.

```
curl -X OPTIONS --data "{ 'id': '1', 'password': 'newpassword' }" http://localhost/v1/user/login
[{"field":"username","message":"Username cannot be blank."},{"field":"password","message":"Password cannot be blank."}]
```
```
curl -X POST --data "{ 'id': '1', 'password': 'newpassword' }" http://localhost/v1/user/login
[{"field":"username","message":"Username cannot be blank."},{"field":"password","message":"Password cannot be blank."}]
```

any idea what should I do? here is my whole UserController.php:

`<?php
namespace api\modules\v1\controllers;

use common\models\User;
use yii\filters\AccessControl;
use yii\filters\auth\CompositeAuth;
use yii\filters\auth\HttpBearerAuth;
use yii\filters\VerbFilter;
use yii\helpers\ArrayHelper;
use yii\rest\Controller;
use common\models\LoginForm;
use yii\web\NotFoundHttpException;

class UserController extends Controller
{
	/**
	 * @inheritdoc
	 */
	public function behaviors()
	{
		$behaviors = parent::behaviors();
		unset($behaviors['authenticator']);

		return ArrayHelper::merge(parent::behaviors(), [
			'verbs'         => [
				'class'   => VerbFilter::className(),
				'actions' => [
				],
			],
			'corsFilter' => [
				'class' => \yii\filters\Cors::className(),
				'cors'  => [
					'Origin'                           => ['*'],
					'Access-Control-Request-Method'    => ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
					// Allow only POST and PUT methods
					'Access-Control-Request-Headers'   => ['Origin', 'X-Requested-With', 'Content-Type', 'Accept', 'Authorization'],
					// Allow only headers 'X-Wsse'
					'Access-Control-Allow-Credentials' => true,
					// Allow OPTIONS caching
					'Access-Control-Max-Age'           => 3600,
					// Allow the X-Pagination-Current-Page header to be exposed to the browser.
					'Access-Control-Expose-Headers'    => ['X-Pagination-Current-Page'],
				],
			],
			'authenticator' => [
				'class'       => CompositeAuth::className(),
				'except'      => ['login'],
				'authMethods' => [
					HttpBearerAuth::className(),
				],
			],
		]);
	}

	/**
	 * @return array
	 * @throws \yii\base\InvalidConfigException
	 */
	public function actionLogin()
	{
		$params = \Yii::$app->request->getBodyParams();
		$model = new LoginForm();

		if ($model->load(['LoginForm' => $params])) {
			$user = $model->login();

			if ($user) {
				return [
					'jwt' => $user->getJWT(),
				];
			}
		}

		return $model;
	}
}
`

and my urlManagers config:

`		'urlManager'   => [
			'enablePrettyUrl' => true,
			'showScriptName'  => false,
			'rules'           => [
				[
					'class'      => 'yii\rest\UrlRule',
					'controller' => 'v1/user',
					'extraPatterns' => [
						'POST login' => 'login',
						'OPTIONS login' => 'options'
					]
				],
			],
		],
`

tunecino:
@kiuka 

1- it should be `'except' => ['login', 'options']` otherwise `authenticator` rejects the preflight.
2- you are extending `yii\rest\Controller`. the options action is defined in its child `ActiveController` so you'll need to add it manually:
```php
private $_verbs = ['POST', 'OPTIONS'];

public function actions() 
{
    return [
        'options' => [
            'class' => 'yii\rest\OptionsAction',
            'collectionOptions' =>  $this->_verbs,
            'resourceOptions' =>  $this->_verbs
        ]
    ];
}
```
or this may be enough:

```php
public function actionOptions()
{
    Yii::$app->getResponse()->getHeaders()->set('Allow', implode(', ', $this->_verbs));
}
```

kiuka:
@tunecino thanks for your reply! I modified the code, but it still not working.

the code looks like the following now:

```
		'urlManager'   => [
			'enablePrettyUrl' => true,
			'showScriptName'  => false,
			'rules'           => [
				[
					'class'      => 'yii\rest\UrlRule',
					'controller' => 'v1/user',
					'patterns' => [
						'POST login' => 'login',
						'OPTIONS login' => 'options'
					]
				],
			],
		],

```

```
namespace api\modules\v1\controllers;

use yii\filters\auth\CompositeAuth;
use yii\filters\auth\HttpBearerAuth;
use yii\filters\VerbFilter;
use yii\helpers\ArrayHelper;
use yii\rest\Controller;
use common\models\LoginForm;

class UserController extends Controller
{
	private $_verbs = ['POST', 'OPTIONS'];

	public function behaviors()
	{
		$behaviors = parent::behaviors();
		unset($behaviors['authenticator']);

		return ArrayHelper::merge(parent::behaviors(), [
			'verbs'         => [
				'class'   => VerbFilter::className(),
				'actions' => [
				],
			],
			'corsFilter'    => [
				'class' => \yii\filters\Cors::className(),
				'cors'  => [
					'Origin'                           => ['*'],
					'Access-Control-Request-Method'    => $this->_verbs,
					'Access-Control-Request-Headers'   => ['Origin', 'X-Requested-With', 'Content-Type', 'Accept', 'Authorization'],
					'Access-Control-Allow-Credentials' => true,
					'Access-Control-Max-Age'           => 3600,
					'Access-Control-Expose-Headers'    => ['X-Pagination-Current-Page'],
				],
			],
			'authenticator' => [
				'class'       => CompositeAuth::className(),
				'except'      => ['login', 'options'],
				'authMethods' => [
					HttpBearerAuth::className(),
				],
			],
		]);
	}

	public function actions()
	{
		return [
			'options' => [
				'class'             => 'yii\rest\OptionsAction',
				'collectionOptions' => $this->_verbs,
				'resourceOptions'   => $this->_verbs,
			],
		];
	}

	public function actionOptions()
	{
		\Yii::$app->getResponse()->getHeaders()->set('Allow', implode(', ', $this->_verbs));
	}

	public function actionLogin()
	{
		$params = \Yii::$app->request->getBodyParams();
		$model = new LoginForm();

		if ($model->load(['LoginForm' => $params])) {
			$user = $model->login();

			if ($user) {
				return [
					'jwt' => $user->getJWT(),
				];
			}
		}

		return $model;
	}
}

```

any other advice? thanks in advance!

tunecino:
@kiuka by manually adding the options action I meant choose one don't implement both. keep `actions()` and remove `actionOptions()`.

is that all your list of `rules` ? 

If I'm not wrong according to your urlManager configs `'controller' => 'v1/user'` should be same as doing `'controller' => ['v1/users' => 'v1/user']` as it will be pluralized by UrlRule so what you did:

> curl -X OPTIONS --data "{ 'id': '1', 'password': 'newpassword' }" http://localhost/v1/user/login
> [{"field":"username","message":"Username cannot be blank."},{"field":"password","message":"Password cannot be blank."}]

should throw a 404 NotfoundException instead of redirecting to `actionLogin`. So or you have more not well-configured rules or it is a different problem to which you should open a different issue page.

kiuka:
@tunecino Thanks for your help.

I've debugged, and it was because the reason that the patterns of the urlManager only apply if request the pluralized form of the controller. So I changed the urlManager config, now it works like a charm:

```
				[
					'class'      => 'yii\rest\UrlRule',
					'controller' => [
						'v1/user' => 'v1/user',
					],
					'patterns'   => [
						'OPTIONS <action>' => 'options',
					],
				],

```

here is the source of my controller as well, it could be useful for someone:

```
namespace api\modules\v1\controllers;

use yii\filters\auth\CompositeAuth;
use yii\filters\auth\HttpBearerAuth;
use yii\helpers\ArrayHelper;
use yii\rest\Controller;
use common\models\LoginForm;

class UserController extends Controller
{
	public function behaviors()
	{
		return ArrayHelper::merge(parent::behaviors(), [
			'corsFilter'    => [
				'class' => \yii\filters\Cors::className(),
				'cors'  => [
					'Origin'                           => ['*'],
					'Access-Control-Request-Method'    => ['GET', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
					'Access-Control-Request-Headers'   => ['Origin', 'X-Requested-With', 'Content-Type', 'Accept', 'Authorization'],
					'Access-Control-Allow-Credentials' => true,
					'Access-Control-Max-Age'           => 3600,
					'Access-Control-Expose-Headers'    => ['X-Pagination-Current-Page'],
				],
			],
			'authenticator' => [
				'class'       => CompositeAuth::className(),
				'except'      => ['login', 'options'],
				'authMethods' => [
					HttpBearerAuth::className(),
				],
			],
		]);
	}

	public function actions()
	{
		return [
			'options' => [
				'class' => 'yii\rest\OptionsAction',
			],
		];
	}

	protected function verbs()
	{
		return [
			'login'  => ['POST'],
		];
	}

	public function actionLogin()
	{
		$params = \Yii::$app->request->getBodyParams();
		$model = new LoginForm();

		if ($model->load(['LoginForm' => $params])) {
			$user = $model->login();

			if ($user) {
				return [
					'jwt' => $user->getJWT(),
				];
			}
		}

		return $model;
	}
}

```

jaminm:
I was stuck on this one too and this little bit fixed it for me:
```
[
                    'class'      => 'yii\rest\UrlRule',
                    'controller' => [
                        'v1/user' => 'v1/user',
                    ],
                    'patterns'   => [
                        'OPTIONS <action>' => 'options',
                    ],
                ],

```
You have to do everything else mentioned here, but if you aren't using the plural versions of the controller in your endpoint call you have to put this in your urlManager settings or none of the other code works.

zjien:
@kiuka Thx,your answer are so helpful

radalin:
For those who was not able to solve the problem from above or require some other things then the above a simple but a dirty way to solve this could be defining the behaviors such as this: 

>     public function behaviors()
    {
        $behaviors = []; // Or merge it from the parent
        $behaviors[] = [
            'class' => \yii\filters\Cors::className(),
            'cors' => [
                'Origin' => ['*'],
                'Access-Control-Request-Method' => ['GET', 'POST', 'PUT', 'PATCH', 'DELETE', 'HEAD', 'OPTIONS'],
                'Access-Control-Request-Headers' => ['*'],
                // This part is necessary and important
                'Access-Control-Allow-Credentials' => true,
                'Access-Control-Max-Age' => 86400,
                'Access-Control-Expose-Headers' => [],
            ]
        ];
        if (\Yii::$app->request->method !== "OPTIONS") {
            $behaviors[] = [
                'class' => HttpBasicAuth::className(),
                'except' => ["options"],
            ];
        }
        return $behaviors;
    }

This is not an elegant solution, I agree but it actually works. The example is working on Basic Auth not on Bearer Auth but logic is the same on preflight operation

tunecino:
@radlin there is documentation about it now: 
http://www.yiiframework.com/doc-2.0/guide-rest-controllers.html#cors

radalin:
@tunecino probably there is a configuration which I'm missing so that the "except" keyword works. Because if I delete that check about "OPTIONS", preflight requests fail again. Probably because I'm missing something. Thanks for letting me know though.

arboshiki:
I have authentication enabled for every action except `options`. I redirect `OPTIONS` method request on any action into options action but still I get Unauthorized error. Do you have any idea?

Here is what I have in url manager.

```
[
    'controller' => [
        'v1/user' => 'v1/user',
    ],
    'patterns' => [
        'OPTIONS <action>' => 'options'
    ]
]
```

Here is my controller

```
public function behaviors()
{
    $behaviors = parent::behaviors();

    // add CORS filter
    $behaviors['corsFilter'] = [
        'class' => \yii\filters\Cors::className(),
    ];

        'class' => CompositeAuth::className(),
        'except' => ['options'],
        'authMethods' => [
            [
                'class' => HttpBasicAuth::className(),
                'auth' => function ($username, $password) {
                    $user = User::findByLogin($username);
                    return $user->validatePassword($password)
                        ? $user
                        : null;
                }
            ],
            HttpBearerAuth::className(),
            QueryParamAuth::className()
        ]
    ];

    return $behaviors;
}
```

arboshiki:
I have disabled pluralization specifying `'pluralize'     => false,` and now it works fine.

PR #8632, opened 2015-06-02
i18n MessageController now supports config parameter 'markUnused' that turns adding @@s either on or off

marius7383:
This commit addresses issue #8625 and adds the config parameter `markUnused` to the i18n MessageController. Being set to false it prevents adding `@@` strings to potentially unused messages.

Issue #8625, opened 2015-06-01
MessageController adds "@@" to messages, that aren't directly used via Yii::t()


marius7383:
In our Yii2 application we use the awesome i18n mechanisms that really save a lot of work by creating the message files automatically. We also have a lot of message keys that aren't passed directly to the `Yii::t(...)` method and therefore always get `@@` appended and prepended. At one point this starts to become really annyoing, since they have to be removed again. It is a nice feature, but we feel it should be possible to turn it off. 

We enhanced the `MessageController` by an additional parameter named `markUnused` that allows to turn the `@@` off ([Gist](https://gist.github.com/marius7383/bfe59084495f1e429498)). Maybe it is possible to introduce this feature into future Yii2 releases.

cebe:
@marius7383 would you mind sending a pull request?

samdark:
Yep, that's OK to add it.

marius7383:
I just sent the pull request, hope it helps.

PR #8633, opened 2015-06-02
i18n MessageController now supports config parameter 'markUnused' that turns adding @@s either on or off

marius7383:
This is a rework of pull request #8632 since it was not setup properly and addresses issue #8625 as well. 

@klimov-paul  About the newly added content in the upgrade-log I am not fully sure, because I had to add the `Upgrade from Yii 2.0.4` section and the `MessageController` doesn't seem to be a class that is supposed to be used in application development. Hope it's still correct.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -29,6 +29,7 @@ Yii Framework 2 Change Log
 - Enh #8486: Added support to automatically set the `maxlength` attribute for `Html::activeTextArea()` and `Html::activePassword()` (klimov-paul)
 - Enh #8566: Added support for 'only' and 'except' options for `yii\web\AssetManager::publish()` (klimov-paul)
 - Enh #8574: Added `yii\console\controllers\MessageController` support .pot file creation (pgaultier)
+- Enh #8625: MessageController adds "@@" to messages, that aren't directly used via Yii::t() (marius7383)
 - Chg #6354: `ErrorHandler::logException()` will now log the whole exception object instead of only its string representation (cebe)
 - Chg #8556: Extracted `yii\web\User::getAuthManager()` method (samdark)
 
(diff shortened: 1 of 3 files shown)


framework/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ Yii Framework 2 Change Log
 - Enh #8486: Added support to automatically set the `maxlength` attribute for `Html::activeTextArea()` and `Html::activePassword()` (klimov-paul)
 - Enh #8566: Added support for 'only' and 'except' options for `yii\web\AssetManager::publish()` (klimov-paul)
 - Enh #8574: Added `yii\console\controllers\MessageController` support .pot file creation (pgaultier)
+- Enh #8625: MessageController adds "@@" to messages, that aren't directly used via Yii::t() (marius7383)
```
klimov-paul: Should be:
```
Added 'markUnused' option to `yii\console\controllers\MessageController`
```

framework/UPGRADE.md
```diff
@@ -8,6 +8,16 @@ if you want to upgrade from version A to version C and there is
 version B between A and C, you need to following the instructions
 for both A and B.
 
+
+Upgrade from Yii 2.0.4
+----------------------
+  
+* The signature of several `saveMessagesTo...` methods in `yii\console\controllers\MessageController` is changed. They now have an additional parameter named `markUnused`:
```
klimov-paul: Should be:
```
The signature of the following methods in `yii\console\controllers\MessageController` has changed:
```
Better will be following `yii\base\View::beforeRender()` and `yii\base\View::afterRender()` note at:
https://github.com/yiisoft/yii2/blob/master/framework/UPGRADE.md#upgrade-from-yii-20-rc

klimov-paul:
Overall PR looks fine. Still there is a minor BC break.
@yiisoft/core-developers, what is your opinion: should we include this to 2.0.5 ?

framework/console/controllers/MessageController.php
```diff
@@ -227,10 +230,14 @@ protected function saveMessagesToDb($messages, $db, $sourceMessageTable, $messag
...
+                if ($markUnused) {
+                    $expression = "CONCAT('@@',message,'@@')";
+                }
                 $db->createCommand()
                    ->update(
                        $sourceMessageTable,
-                       ['message' => new \yii\db\Expression("CONCAT('@@',message,'@@')")],
+                       ['message' => new \yii\db\Expression($expression)],
                        ['in', 'id', $obsolete]
                    )->execute();
```
cebe: the whole update statement is not needed if we should not mark unused.

cebe:
If the `$markUnused` parameter was added at the end, the BC break would be okay for me as the param would be simply ignored. This way it will result in weird behavior.

Issue #8573, opened 2015-05-27
oci schema map wrongly NUMBER dbType to either type INTEGER or DECIMAL


bobonov:
In /yii2/db/oci/Schema.php
extractColumnType method wrongly map NUMBERS to INTEGER or DECIMAL.

According to Oracle documentation
http://docs.oracle.com/cd/B28359_01/server.111/b28318/datatype.htm#CNCPT1832

When scale is 0 the number should be treated as an integer.
But if "If a precision is not specified, the column stores values as given."
As I tested, even is is not clear from the docs, this override the scale.

So if the NUMBER column has scale=0 and precision=0 a double value is stored correctly.

cebe:
thanks for reporting this. can you send a pull request that fixes it?

vbelogai:
I have solution for this:
https://github.com/vbelogai/yii2/blob/oci/framework/db/oci/Schema.php#L398

but my project's tree contains another pending pull request (https://github.com/yiisoft/yii2/pull/8368).

How i can make new pull request without pending changes (only framework/db/oci/Schema.php)?

cebe:
> How i can make new pull request without pending changes (only framework/db/oci/Schema.php)?

you should always work on a different branch to separate the changes.

vbelogai:
OK, #8643 pull request is ready

SilverFire:
Fixed https://github.com/yiisoft/yii2/commit/1f8aafd36ba6b7f8bb33e8f06067b7f1ff2a3989

PR #8643, opened 2015-06-03
#8573 fix

vbelogai:
fixing [#8573](https://github.com/yiisoft/yii2/issues/8573)
**NUMBER, NUMBER(x,y)** - (data_scale is null or > 0) should be mapped to **decimal**,
**NUMBER(x,0), NUMBER(x)** - (data_scale = 0) should be mapped to **integer**,
**INTEGER** always should be mapped to **integer**

### DIFF
--- a/framework/db/oci/Schema.php
+++ b/framework/db/oci/Schema.php
@@ -395,12 +395,14 @@ SQL;
 
         if (strpos($dbType, 'FLOAT') !== false || strpos($dbType, 'DOUBLE') !== false) {
             $column->type = 'double';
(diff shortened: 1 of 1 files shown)


bobonov:
You misunderstood the issue.
The key is the precision, your proposal does not fix the problem.

vbelogai:
Did you mean case when 10^precision can be  > 2^PHP_INT_SIZE ?

чт, 4 июня 2015 г. в 11:38, bobonov <notifications@github.com>:

Issue #8627, opened 2015-06-01
DAO layer is broken


creocoder:
```php
$db = Yii::$app->db;
$db->createCommand()->createTable('foo', ['id' => Schema::TYPE_PK, 'bar' => Schema::TYPE_INTEGER])->execute();
$db->createCommand()->insert('foo', ['bar' => 1])->execute();
$db->createCommand()->alterColumn('foo', 'bar', Schema::TYPE_STRING)->execute();
$db->createCommand()->insert('foo', ['bar' => 'hello'])->execute();
```

Result:

```
id | bar
 1 | 1
 2 | 0
```

Should be:

```
id | bar
 1 | 1
 2 | hello
```

Same thing using pure PDO show that everything is fine. So this is fw DAO layer bug.

creocoder:
Tested on 2.0.4 release version.

cebe:
how to you get the result? It is not part of the commands above...

creocoder:
@cebe Before report i made investigations around 6 hours. This trouble happens on production server when apply migrations. This bug was reproduced on 3 different machines with different OS, php versions and mysql server version. Also its cleanly Yii 2 DAO layer bug. Because when using pure PDO everything is fine.

> how to you get the result?

Just run code on your machine.

cebe:
have you tried turning off the schema cache?

creocoder:
@cebe Ofcourse ;-)

cebe:
okay :)

creocoder:
@cebe I assume this trouble caused by wrong PDO usage and prepared statement. Most interesting thing is if you kill first insert everything will be fine with second insert. So bug is very funny. On other side its extremally dangerous especially for production environment. For example consider you have 4 migrations: 1) create table, 2) insert into that table, 3) alter table, 4) insert into that table. On dev environment you'll run migrations step by step and you'll get one result. When you run `yii migrate` on production and all migrations in one step you'll be very surprised.

creocoder:
@cebe Look PDO example:

```php
$db = Yii::$app->db;
$db->createCommand()->createTable('foo', ['id' => Schema::TYPE_PK, 'bar' => Schema::TYPE_INTEGER])->execute();

$sth = $this->db->pdo->prepare('INSERT INTO foo (bar) VALUES (:bar);');
$sth->bindValue(':bar', 1, PDO::PARAM_INT);
$sth->execute();

[...]

PR #8646, opened 2015-06-03
connection schema check per action

yurii-github:
sample answer to issue topic #8627 

```
function actionTest()
{
    ob_start();

    $m = new \yii\db\Migration();
    //clean
    $m->dropTable('test');
    $m->createTable('test', ['a' => 'int', 'b' => 'int']);
    // 1
    $m->insert('test', ['b' => 'b']); // 0 -- expected (but strange why no error)
    // 2
    $m->alterColumn('test', 'b', 'text');
    $m->insert('test', ['b' => 'b']); // 0 -- silent fail, per request "cache" as schema object is never updated
    // 3
    $m->db->enableSchemaCache = false; // after added check
    $m->insert('test', ['b' => 'b']); // 'b' - OK

    echo str_replace("\n", '<br/>', ob_get_clean());
}
```

### DIFF
--- a/framework/db/Connection.php
+++ b/framework/db/Connection.php
@@ -182,6 +182,7 @@ class Connection extends Component
      * @var boolean whether to enable schema caching.
      * Note that in order to enable truly schema caching, a valid cache component as specified
      * by [[schemaCache]] must be enabled and [[enableSchemaCache]] must be set true.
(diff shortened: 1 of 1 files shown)


klimov-paul:
We hav ealready concluded such fix to be unacceptable.

yurii-github:
then just use 
$m->db->schema->getTableSchema('test',true);

'cause that's not a bug then if you accept this way

regards

PR #8654, opened 2015-06-04
docs/guide/helper-html.md small fixes [ci skip]

softark:


docs/guide/helper-html.md
```diff
@@ -92,21 +93,23 @@ if ($type === 'success') {
     Html::addCssClass($options, ['btn-success', 'btn-lg']);
 }
 
-echo Html::tag('div', 'Save', $options); // renders `class="btn btn-success btn-lg"`
+echo Html::tag('div', 'Save', $options);
+// renders '<div class="btn btn-success btn-lg">Save</div>'
 ```
 
-`Html::addCssClass()` prevents duplicating classes, so you don't need to worry the same class may appear twice:
+`Html::addCssClass()` prevents duplicating classes, so you don't need to worry if the same class may appear twice:
```
klimov-paul: Should not coma be here: before 'if'?

softark: 'worry if ...' is almost equivalent to 'worry about the possibility that ...'.
So you don't need a comma before 'if'.

softark: Um, wait a moment, please.
I might be wrong.

klimov-paul: There are 2 senetnces: "you don't need..." and "class may appear" with conjunction. As far as I know it should be a comma between them.

klimov-paul:
Thank you

Issue #8671, opened 2015-06-05
Add JsHelper


SilverFire:
There is a `clientValidateAttribute` method in the `RegularExpressionValidator`.
In that code we have some magic with the regular expression, before it will be passed to the client-side.
I need the same code in the [`IpValidator`](https://github.com/yiisoft/yii2/issues/3506). 

Propose to extract it to a separate method of a new `JsHelper`
Pull request below.

PR #8672, opened 2015-06-05
Enh. #8671 - added JsHelper

SilverFire:
For details please open #8671


samdark:
Refactoring looks reasonable to me. @yiisoft/core-developers thoughts?

klimov-paul:
I don't think we should introduce extra helper `JsHelper`, at least not for the single method.
Regex converting method can be moved to `helpers\Html` or to `JsExpression`, or even to the `Validator`.

framework/helpers/BaseJsHelper.php
```diff
@@ -0,0 +1,43 @@
...
+        if ($deliminator !== '/') {
+            $pattern = '/' . str_replace('/', '\\/', substr($pattern, 1, $pos - 1)) . '/';
+        } else {
+            $pattern = substr($pattern, 0, $pos + 1);
+        }
+        if (!empty($flag)) {
+            $pattern .= preg_replace('/[^igm]/', '', $flag);
+        }
+
+        return new JsExpression($pattern);
```
klimov-paul: Instantiating of the `JsExpression` here places unnecessary limitation: escaped pattern can be used as a part(!) of other expression, which such approach does not allow.

framework/helpers/BaseJsHelper.php
```diff
@@ -0,0 +1,43 @@
...
+ * @since 2.0.5
+ */
+class BaseJsHelper
+{
+    /**
+     * Escapes regular expression to use in JavaScript (client-side)
+     * @param string $regexp
+     * @return JsExpression
+     */
+    public static function escapeRegexp ($regexp) {
```
klimov-paul: Name should be full: `escapeRegularExpression` just like `RegularExpressionValidator`

framework/helpers/BaseJsHelper.php
SilverFire: @klimov-paul `JsExpression` has `__toString` method, so almost all thins you can do with strings, will also work with JsExpression. http://3v4l.org/mc8Q3

framework/helpers/BaseJsHelper.php
SilverFire: @klimov-paul Agree

SilverFire:
@klimov-paul Thank you for ideas.
Moving to `Validator` is not a good idea, as for me. `Validator::escapeRegularExpression` looks very strange. `JsExpression` is used only as a wrapper now, I'm not sure, that adding helping method into it is realy resonable.
Moving to `helpers/Html` may be better, than creating `JsHelper`. Now I can't see any other helping methods, that can be moved to `JsHelper`. Do you?

framework/helpers/BaseJsHelper.php
klimov-paul: This does not change anything. Why create an object from a string if it will probably converted to the string next?

samdark:
It definitely should not go into `Validator`. `Html` is OK but the method name would be not so cool. `escapeRegularExpressionForJavaScript`?

klimov-paul:
Could be `escapeJsRegularExpression()`.

SilverFire:
@klimov-paul `JsHelper::escapeRegularExpression` look better, as for me

samdark:
@yiisoft/core-developers need help with naming it right.

qiangxue:
Are there any other helper methods related with JS manipulation?

samdark:
Not yet.

qiangxue:
If we only have a single method, I would prefer we put it in an existing helper class. 
Otherwise, introducing a new helper class would be better.

samdark:
What about name of the method if we put it into existing helper?

qiangxue:
`escapeJsRegularExpression` is fine

klimov-paul:
> Are there any other helper methods related with JS manipulation?

We can consider moving `Json::htmlEncode()` there.

samdark:
No since that one can't work with JavaScript.

SilverFire:
So what's the decision of core team?

samdark:
The method name could be `ecodeJsRegularExpression()` to be consistent with `Html::encode()`. Let's put it into Html.

samdark:
https://github.com/yiisoft/yii2/commit/b16d734c29396b794d7ac37cdb331bac8b2f3acd

PR #8675, opened 2015-06-06
Fixed \yii\db\BaseActiveRecord::findOne return type

creocoder:
Method `one()` used inside this method never return anything like `null`.


lynicidn:
in theme:
 * https://github.com/yiisoft/yii2/blob/master/framework/db/QueryInterface.php#L37
 * https://github.com/yiisoft/yii2/blob/master/framework/db/Command.php#L360
only look `gii` and how it work) ? 
 - https://github.com/yiisoft/yii2-gii/blob/master/generators/crud/default/controller.php#L167

PR #8684, opened 2015-06-07
Local number options must override global options

leandrogehlen:


cebe:
can you explain a bit more?

leandrogehlen:
Eg: into `config/web.php` i have the following:

```php
components' => [
     'formatter' => [
          'class' => 'app\components\Formatter',
               'dateFormat' => 'medium',
               'nullDisplay' => '',
               'numberFormatterOptions' => [
                   NumberFormatter::FRACTION_DIGITS => 2,
               ]
          ]
     ]
     ...
```
where i use the `Yii::$app->formatter->asDecimal(10)` the result will be '10,00'
but if i use `Yii::$app->formatter->asDecimal(10, 4)` the result continue being '10,00' but should be '10,0000'

cebe:
fixed it by keeping BC for the symbols. Thanks for reporting!

Issue #7193, opened 2015-02-06
ArrayHelper::getValue() has a bug inside


aprokopenko:
### Issue: 

https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseArrayHelper.php#L190

This line of code has bug. If you try to access the object with no key - you will get a notice instead of default value. So code breaks on this line.

if we have such object:

```php
stdClass Object
(
    [id] => 54d39c0adc7a1dc7fbdb24e8
    [data] => stdClass Object
    (
        [attachment] => stdClass Object
        (
            [id] => 54d39c021b3a91687acc5d17
            [name] => filename
        )
....
```

and then try to run:

```php
$url = ArrayHelper::getValue($data, 'data.attachment.url', 'n/a');
```

we will get error message:

```php
 PHP Notice – yii\base\ErrorException
Undefined property: stdClass::$url
```

### Resolution

change this line of code to this one:

```php
return isset($array->$key)? $array->$key : $default;
```

### Working environment:

- php 5.5, E_STRICT AND E_NOTICES error level enabled
- apache 2.4.7
- ubuntu 14.04

qiangxue:
This is by design. If you are trying to access a nonexisting property of an object, it's a syntax error which should be reported rather than hidden. This is different from the array case.

aprokopenko:
Then you need to update at least the documentation for this function. Code contain such description for this function:

```
If `$array['x']` or `$array->x` is neither an array nor an object, the default value will be returned.
```

So the description is wrong.

qiangxue:
This description is correct. It says if `$array->x` is neither an array nor an object, which doesn't mean the property doesn't exist.

PR #8685, opened 2015-06-07
Object key check error fix

serhatozles:
it's returning error to getting key if the array haven't that key. 
i think this is corret
return property_exists($array,$key) ? $array->$key : $default;
but it doesn't work with ActiveRecord. 
What can i do for this?


serhatozles:
i think this solved with following code;
return isset($array->$key) ? $array->$key : $default;

cebe:
accessing a non existing property of an object is considered to be an error, the original code should stay as is. https://github.com/yiisoft/yii2/issues/7193#issuecomment-73287156

Issue #7772, opened 2015-03-18
optimization method className()


LAV45:
```php
$start = microtime(true);
for($i=0; $i < 1000000; $i++) {
//    User::className(); // 0.81278204917907715
    User::class; // 0.065196990966796875
}
$stop = microtime(true);

echo($stop - $start);
```
This option is 12 times faster
But the method class available since at php 5.5
When do you plan to abandon support for php 5.4?

samdark:
Not in 2.0.x.

LAV45:
Then php 5.4 you can delete from here? https://github.com/yiisoft/yii2/blob/2.1/.travis.yml

samdark:
It wasn't decided yet about 2.1 PHP version.

Faryshta:
I think this one got solved already

SilverFire:
Yeah, it is decided to use `::class`. Thank you @Faryshta

PR #8693, opened 2015-06-08
Use PHP 5.5 ::class property to replace Object::className()

neatlife:
rt
http://php.net/manual/tr/migration55.new-features.php#migration55.new-features.class-name


fernandezekiel:
but isn't the minimum supported version is 5.4? this won't work on that version right?

su-xiaolin:
@fernandezekiel yes, at least php 5.5

klimov-paul:
Relates to #7772.
Closing this, because require PHP 5.5 as minumum is not decided and will not be concidered before 2.1 milestone.

samdark:
Reopened and set milestone to 2.1.

framework/base/Application.php
```diff
@@ -616,14 +616,14 @@ public function getSecurity()
...
-            'assetManager' => ['class' => 'yii\web\AssetManager'],
-            'security' => ['class' => 'yii\base\Security'],
+            'log' => ['class' => \yii\log\Dispatcher::class],
+            'view' => ['class' => \yii\web\View::class],
+            'formatter' => ['class' => \yii\i18n\Formatter::class],
+            'i18n' => ['class' => \yii\i18n\I18N::class],
+            'mailer' => ['class' => \yii\swiftmailer\Mailer::class],
+            'urlManager' => ['class' => \yii\web\UrlManager::class],
+            'assetManager' => ['class' => \yii\web\AssetManager::class],
+            'security' => ['class' => \yii\base\Security::class],
```
klimov-paul: This is incorrect: it will force all these classes source code to be loaded even if the components will not used during script execution.

samdark: Agree. We should not do that because of autoload.

Faryshta: Actually they won't be loaded, `::class` is determined at compilation time, not execution.

You can even do `Fake\Namespace\UnexistentClass::class` and it will return the string even if the class doesn't exists.

samdark: Wow, that's really really nice.

mikehaertl: What is the point of this change anyway? Aren't the classnames configered here already fully qualified names? I mean, what is the point in writing `\yii\base\Security::class` if that gives you `\yii\base\Security`?

klimov-paul: It is mainly for the IDE support.

samdark: Yeah, just nicer autocomplete and checks.

yurii-github: it doesnt worth to set minimum php from 5.4 ->  5.5
if you have php 5.5 nothing holds you from using this ::clas stuff.

in other words, http://php.net/manual/en/migration55.new-features.php
it ridiculous addition to set higher php version

Faryshta:
Related to #8870 i can't find a way to solve this issue without the `::class` contant in php5.5

Currently `static::className()`,  `self::className()` and  `parent::className()` return the exact same values

Vovan-VE:
The are some problems like <https://bugs.php.net/bug.php?id=66811> with `::class` in PHP < 7.

framework/console/Application.php
```diff
@@ -176,12 +176,12 @@ public function runAction($route, $params = [])
     public function coreCommands()
     {
         return [
-            'message' => 'yii\console\controllers\MessageController',
-            'help' => 'yii\console\controllers\HelpController',
-            'migrate' => 'yii\console\controllers\MigrateController',
-            'cache' => 'yii\console\controllers\CacheController',
-            'asset' => 'yii\console\controllers\AssetController',
-            'fixture' => 'yii\console\controllers\FixtureController',
+            'message' => \yii\console\controllers\MessageController::class,
```
Faryshta: why not use

```php
'message' => controllers\MessageController::class,
```

since you are already on the `yii\console` namespace?

Faryshta:
@Vovan-VE the same happens if we use the current solution `static::className()` doesn't work outside the class. Actually `static` doesn't work outside the class at all.

you can test it on views for example.

```php
/**
 * @this yii\web\View
 */
$this->registerJs('alert("hello");', static::POS_END);
```

framework/db/ActiveRecord.php
```diff
@@ -268,7 +268,7 @@ public static function deleteAll($condition = '', $params = [])
      */
     public static function find()
     {
-        return Yii::createObject(ActiveQuery::className(), [get_called_class()]);
+        return Yii::createObject(ActiveQuery::class, [get_called_class()]);
```
Faryshta: We should change it to

```php
new ActiveQuery(static::class);
```

framework/filters/AccessControl.php
```diff
@@ -31,7 +31,7 @@
  * {
  *     return [
  *         'access' => [
- *             'class' => \yii\filters\AccessControl::className(),
+ *             'class' => \yii\filters\AccessControl::class,
```
Faryshta: we should put it on the `use` section

framework/grid/GridView.php
```diff
@@ -526,7 +526,7 @@ protected function initColumns()
                 $column = $this->createDataColumn($column);
             } else {
                 $column = Yii::createObject(array_merge([
-                    'class' => $this->dataColumnClass ? : DataColumn::className(),
+                    'class' => $this->dataColumnClass ? : DataColumn::class,
```
Faryshta: why not set

```php
public $dataColumnClass = DataColumn::class;
```

and then change this line to 

```php
'class' => $this->dataColumnClass,
```

framework/db/ActiveRecord.php
samdark: It won't go through DI then,

Faryshta: then

```php
 return Yii::createObject(ActiveQuery::class, [static::class]);
```

samdark: `get_called_class()` is necessary as well since it could be inherited from `ActiveQuery`.

Faryshta: > since it could be inherited from ActiveQuery.

I think you meant `ActiveRecord`. And if thats the case then `static::class` will work perfectly.

if you mean `ActiveQuery` when I must ask for an example on what that is.

samdark: I mean `ActiveQuery`. In this case `find()` is called on AR instance and returns AQ instance. The difference is how it's created.

Faryshta: since `find` is called from the AR then `get_called_class()` and `static::class` should behave similary.
```php
User extends ActiveRecord
{
}

User::find()->modelClass // will return a query created using the `User` class.
```

https://3v4l.org/gZFEV here is a live example.

framework/grid/GridView.php
serieznyi: +1

Faryshta:
I think this one got solved already

SilverFire:
Closing this PR because changes for 2.1 will be accepted to the separate branch.
Continue in https://github.com/yiisoft/yii2/pull/11477

PR #8698, opened 2015-06-08
"Move Identity Cookie code into separate functions in web\User.php"

ghost:
Pull Request for feature request "Move Identity Cookie code into separate functions in web\User.php"

Thank you!

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -682,7 +682,7 @@ class User extends Component
      */
     public function can($permissionName, $params = [], $allowCaching = true)
     {
-        $auth = Yii::$app->getAuthManager();
+        $auth = $this->getAuthManager();
         if ($allowCaching && empty($params) && isset($this->_access[$permissionName])) {
             return $this->_access[$permissionName];
         }
@@ -692,5 +692,18 @@ class User extends Component
         }
 
         return $access;
+    }
+
+    /**
+     * Returns auth manager associated with the user component.
+     *
+     * By default this is the `authManager` application component.
+     * You may override this method to return a different auth manager instance if needed.
+     * @return \yii\rbac\ManagerInterface
+     * @since 2.0.5
+     */
+    protected function getAuthManager()
+    {
+        return Yii::$app->getAuthManager();
     }
 }
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -112,21 +112,17 @@ class User extends Component
     /**
      * @var boolean whether to automatically renew the identity cookie.
      * This property is effective only when [[enableSession]] and [[enableAutoLogin]] are true.
-     * When the is true, the identity cookie will expire based on [[autoRenewEachPage]] setting.
-     * When this is false, the identity cookie will expire [[identityCookieDuration]] seconds after the first login.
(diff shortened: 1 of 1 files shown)


framework/web/User.php
```diff
@@ -437,17 +426,17 @@ public function loginRequired($checkAjax = true)
...
      * so that the event is triggered.
      * @param IdentityInterface $identity the user identity information
-     * @param boolean $cookieBased whether the login is cookie-based
      * @param integer $duration number of seconds that the user can remain in logged-in status.
      * If 0, it means login till the user closes the browser or the session is manually destroyed.
+     * @param boolean $cookieBased whether the login is cookie-based
      * @return boolean whether the user should continue to be logged in
      */
-    protected function beforeLogin($identity, $cookieBased, $duration)
+    protected function beforeLogin($identity, $duration, $cookieBased)
```
samdark: You can't change signature since it's part of API already and it will break backwards compatibility.

framework/web/User.php
```diff
@@ -460,16 +449,16 @@ protected function beforeLogin($identity, $cookieBased, $duration)
      * If you override this method, make sure you call the parent implementation
      * so that the event is triggered.
      * @param IdentityInterface $identity the user identity information
-     * @param boolean $cookieBased whether the login is cookie-based
      * @param integer $duration number of seconds that the user can remain in logged-in status.
      * If 0, it means login till the user closes the browser or the session is manually destroyed.
+     * @param boolean $cookieBased whether the login is cookie-based
      */
-    protected function afterLogin($identity, $cookieBased, $duration)
+    protected function afterLogin($identity, $duration, $cookieBased)
```
samdark: Same here.

framework/web/User.php
```diff
@@ -532,9 +521,10 @@ protected function renewIdentityCookie()
      * information in the cookie.
      * @param IdentityInterface $identity
      * @param integer $duration number of seconds that the user can remain in logged-in status.
+     * @param boolean $cookieBased whether the login is cookie-based
      * @see loginByCookie()
      */
-    protected function sendIdentityCookie($identity, $duration)
+    protected function sendIdentityCookie($identity, $duration, $cookieBased)
```
samdark: And here.

framework/web/User.php
```diff
@@ -559,9 +593,13 @@ protected function sendIdentityCookie($identity, $duration)
      * If null, it means switching the current user to be a guest.
      * @param integer $duration number of seconds that the user can remain in logged-in status.
      * This parameter is used only when `$identity` is not null.
+     * @param boolean $cookieBased whether the login is cookie-based
      */
-    public function switchIdentity($identity, $duration = 0)
+    public function switchIdentity($identity, $duration = 0, $cookieBased = false)
```
samdark: And here.

maine-mike: So wouldn’t this be OK since there is a default if the third parameter is omitted?

maine-mike: Hi,  sorry for the mess.  I think I tried to do too much at the same time.  I created a new pull request just moving the cookie code into two new protected functions.

PR #8707, opened 2015-06-08
Update BlameableBehavior.php

oleh-ozimok:
Added ability to specify default user in BlameableBehavior

### DIFF
--- a/framework/behaviors/BlameableBehavior.php
+++ b/framework/behaviors/BlameableBehavior.php
@@ -63,6 +63,10 @@ class BlameableBehavior extends AttributeBehavior
      */
     public $updatedByAttribute = 'updated_by';
     /**
+     * @var integer the value that will be used as a default user.
+     */
+    public $defaultUserId = null;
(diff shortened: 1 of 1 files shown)


cebe:
what is the use case for this?

oleg-ozimok:
When the model changed by application (e.g. cron job) both attributes = null. By means of current property can be avoided of this (Sorry for my english).

cebe:
why don't you use `$value` property for this?

oleg-ozimok:
too much code

cebe:
It is not more code:
```
'value' => function() { return 123; },
vs.
'defaultUserId' => 123,
```

oleg-ozimok:
if ($this->value === null) {
  $user = Yii::$app->get('user', false);
  return $user && !$user->isGuest ? $user->id : null;
} else {
  
 // there is no verification of the user
 // ---------------------------------------------
          
  return call_user_func($this->value, $event);
}

PR #8714, opened 2015-06-09
Just move Identity Cookie code into separate functions in web\User.php

ghost:
OK.  This just creates two new protected functions that separate out the cookie functions.

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -279,35 +279,17 @@ class User extends Component
      */
     protected function loginByCookie()
     {
-        $value = Yii::$app->getRequest()->getCookies()->getValue($this->identityCookie['name']);
-        if ($value === null) {
-            return;
-        }
-
-        $data = json_decode($value, true);
-        if (count($data) !== 3 || !isset($data[0], $data[1], $data[2])) {
-            return;
-        }
-
-        list ($id, $authKey, $duration) = $data;
-        /* @var $class IdentityInterface */
-        $class = $this->identityClass;
-        $identity = $class::findIdentity($id);
-        if ($identity === null) {
-            return;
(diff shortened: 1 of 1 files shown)


samdark:
It seems to be changing behavior. Not time to verify it for 2.0.6.

maine-mike:
OK.

samdark:
Doesn't look merge-able to me.

maine-mike:
Can you tell me why?

samdark:
1. It changes behavior.
2. Extracted method name doesn't match return value.
3. The scope of extraction is questionable.

maine-mike:
Would you please explain those three things?

What behavior is changed?

samdark:
This: https://github.com/yiisoft/yii2/pull/8714/files#diff-12395c67a35c91aa5fc9ac3f6554375fR591

There are more issues. For example, https://github.com/yiisoft/yii2/pull/8714/files#diff-12395c67a35c91aa5fc9ac3f6554375fR570. Parameter isn't used anywhere.

maine-mike:
OK.  Thanks for letting me know.

The first thing I considered a bug.  If a call is made to switchIdentity with duration > 0, then a cookie is created.  If the next call to switchIdentity has duration = 0 then the cookie isn’t removed.  So I’ll report this as a bug separately.

The $identity variable is passed to sendIdentityCookie, so I thought it would make sense to pass it to removeIdentityCookie.  My implementation of cookie logins uses information in the $identity variable in both the sendIdentityCookie and removeIdentityCookie procedures.  $identity is obtainable though other means, so I can remove that.

Thanks for explaining this to me.  I’ll report that bug and see how that goes.

Mike.

samdark:
OK.

Issue #8723, opened 2015-06-09
Using VarDumper::dumpAsString() instead var_export() in \yii\helpers\BaseVarDumper


XzAeRo:
When handling an undefined variable error, the debugger shows: `Warning: var_export does not handle circular references in ~/vendor/yiisoft/yii2/helpers/BaseVarDumper.php on line 187` at the bottom of the debugger page.

Looking for an answer I found an issue (#3327), where adresses a similar problem but for another file, and it was fixed properly by changing `var_export()` to `VarDumper::dumpAsString()`.

I tested the change and the error is gone.

Environment used:
* nginx/1.4.6
* Yii Framework/2.0.5-dev
* PHP 5.5.9-1ubuntu4.9 (php5-fpm)

klimov-paul:
Confirmed. Unit test to reproduce:
```php
public function testExportRecursive()
    {
        $master = new \StdClass();
        $slave = new \StdClass();
        $master->slave = $slave;
        $slave->master = $master;
        $master->function = function() {return true;};

        $exportResult = VarDumper::export($master);
        $this->assertNotEmpty($exportResult);
    }
```

samdark:
The fix is not correct since `dumpAsString` produces output that can't be evaluated by PHP.

samdark:
The issue confirmed and test is OK.

XzAeRo:
[This implementation may be useful](https://github.com/doctrine/common/blob/master/lib/Doctrine/Common/Util/Debug.php)

klimov-paul:
Resolved by commit 16e1e8f

klimov-paul:
d9eb660

PR #8724, opened 2015-06-09
#8723 BaseVarDumper var_export() fix

XzAeRo:
Changed var_export to VarDumper::dumpAsString()


klimov-paul:
Fix is unacceptable, because `dumpAsString` does not maitain the PHP syntax for the complex data structure.

PR #8739, opened 2015-06-10
Enh #8329: add console options for MessageController

### LINKED ISSUE #8329: Enh: extract messages for specified language

vchenin:
close #8329 
This changes-set enables:
- generate config file for 'message/extract' with specified parameters via CLI;
- change parameters via CLI when extract messages, even possible don't use config file.


samdark:
Hmm, it changes behavior of existing `message/config` so isn't backwards compatible. Correct?

vchenin:
@samdark these changes don't break BC-compatibility, just inject additional features.
Existing message/config (after these changes will be message/config-template) create config file by copying early prepared template to the specified destination path.
New implementation of message/config also create config file, but by dynamical generation with specified options to the specified destination path.

samdark:
That's what I mean. Previously it was `message/config`, now it's `message/config-template`.

vchenin:
@samdark new message/config and the old one each does the same: creates config file.
Both implementations require one parameter: destination file path. Added options are optional in new implementation. So there are no breaking changes.

Yes produced files are little different. But both contain configuration array for 'message/extract' command.
Old implementation always require edit config file manually to adjust it to suit personal needs because that file is just a template.
New implementation allows to change parameters dynamically via CLI, but options usage is not required. It's possible use values by default.

In new config it's possible to use aliases for `sourcePath` and `messagePath` parameters, but it still support normal paths as well.
Another difference that new config not contain any languages in list by default. But considering manual edit requirement I think that don't break BC-compatibility.

New implementation of 'message/config'
```shell
./yii message/config config.php
```

produces next content
```php
<?php
/**
 * Configuration file for 'yii message/extract' command.
 *
 * This file is automatically generated by 'yii message/config' command.
 * It contains the parameters for messages extraction from source code.
 * You may modify this file to suit your needs.
 *
 * You can use 'yii message/config-template' command to create
 * template configuration file with detaild description for each parameter.
 */
return [
    'sourcePath' => '@yii',
    'messagePath' => '@yii/messages',
    'languages' => [],
    'translator' => 'Yii::t',
    'sort' => false,
    'overwrite' => true,
    'removeUnused' => false,
    'markUnused' => true,
    'except' => [
        '.svn',
        '.git',
        '.gitignore',
        '.gitkeep',
        '.hgignore',
        '.hgkeep',
        '/messages',
        '/BaseYii.php',
    ],
    'only' => [
        '*.php',
    ],
    'format' => 'php',
    'db' => 'db',
    'ignoreCategories' => [],
];
```

Old implementation of 'message/config'
```shell
./yii message/config-template config-template.php
```

produces next content
```php
<?php

return [
    // string, required, root directory of all source files
    'sourcePath' => __DIR__ . DIRECTORY_SEPARATOR . '..',
    // array, required, list of language codes that the extracted messages
    // should be translated to. For example, ['zh-CN', 'de'].
    'languages' => ['de'],
    // string, the name of the function for translating messages.
    // Defaults to 'Yii::t'. This is used as a mark to find the messages to be
    // translated. You may use a string for single function name or an array for
    // multiple function names.
    'translator' => 'Yii::t',
    // boolean, whether to sort messages by keys when merging new messages
    // with the existing ones. Defaults to false, which means the new (untranslated)
    // messages will be separated from the old (translated) ones.
    'sort' => false,
    // boolean, whether to remove messages that no longer appear in the source code.
    // Defaults to false, which means each of these messages will be enclosed with a pair of '@@' marks.
    'removeUnused' => false,
    // array, list of patterns that specify which files (not directories) should be processed.
    // If empty or not set, all files will be processed.
    // Please refer to "except" for details about the patterns.
    'only' => ['*.php'],
    // array, list of patterns that specify which files/directories should NOT be processed.
    // If empty or not set, all files/directories will be processed.
    // A path matches a pattern if it contains the pattern string at its end. For example,
    // '/a/b' will match all files and directories ending with '/a/b';
    // the '*.svn' will match all files and directories whose name ends with '.svn'.
    // and the '.svn' will match all files and directories named exactly '.svn'.
    // Note, the '/' characters in a pattern matches both '/' and '\'.
    // See helpers/FileHelper::findFiles() description for more details on pattern matching rules.
    // If a file/directory matches both a pattern in "only" and "except", it will NOT be processed.
    'except' => [
        '.svn',
        '.git',
        '.gitignore',
        '.gitkeep',
        '.hgignore',
        '.hgkeep',
        '/messages',
    ],

    // 'php' output format is for saving messages to php files.
    'format' => 'php',
    // Root directory containing message translations.
    'messagePath' => __DIR__,
    // boolean, whether the message file should be overwritten with the merged messages
    'overwrite' => true,

    /*
    // Message categories to ignore
    'ignoreCategories' => [
        'yii',
    ],
     */

    /*
    // 'db' output format is for saving messages to database.
    'format' => 'db',
    // Connection component to use. Optional.
    'db' => 'db',
    // Custom source message table. Optional.
    // 'sourceMessageTable' => '{{%source_message}}',
    // Custom name for translation message table. Optional.
    // 'messageTable' => '{{%message}}',
    */

    /*
    // 'po' output format is for saving messages to gettext po files.
    'format' => 'po',
    // Root directory containing message translations.
    'messagePath' => __DIR__ . DIRECTORY_SEPARATOR . 'messages',
    // Name of the file that will be used for translations.
    'catalog' => 'messages',
    // boolean, whether the message file should be overwritten with the merged messages
    'overwrite' => true,
    */
];
```

I found that I missed several parameters `sourceMessageTable`, `messageTable`, `catalog`.
And it is necessary to add new parameter `markUnused` with description to template.
So it needs to adjust code.

vchenin:
@samdark already done

samdark:
OK. Thanks for explanation.

PR #8788, opened 2015-06-14
add join with alias

nainoon:


Faryshta:
how about that the same alias work for the relation name too?

would that be wise?

cebe:
closing this in favor of #10817

PR #8790, opened 2015-06-14
abort removing duplicate records when pk is not in result set

### LINKED ISSUE #8772: ActiveRecord groupBy()->joinWith('relation',false)->asArray() should be supported
Currently if you try to do a combination of groupBy()->joinWith('relation', false)->asArray() you get caught by the requirement of having a primary key as part of the select statement. I think that this use case should be supported as long as there are no other eager loading or lazy loading joins present in the query builder.

I think this would just be a matter of tweaking the ActiveQuery `populate` function to check to make sure that there are no with statements, no joins that use eager loading, and that asArray is true and if they are simply return the `$models` variable without doing anything else.

cebe:
in this cases it does not make sense to remove duplicates as the result is not on record level anymore.
This could be the case after GROUP BY has been applied.

fixes #8772


samdark:
Looks OK.

PR #8802, opened 2015-06-15
Fix Logger memory performance

mohorev:
Hi everyone. I was trying to write a php daemon on yiiiiiiiiii and came across one unpleasant thing.
This is my example on pseudocode:

``` php

// Worker.php

class Worker
{
    public function work()
    {
        while (true) {
            try {
                $this->run();
            } catch (\Exception $e) {
                // do nothing
            }
        }
    }

    protected function run()
    {
        // some code

        throw new \yii\base\InvalidConfigException();
    }
}


// test.php

$worker = new Worker();
$worker->work();

```

I run the code in an infinite loop and measure the memory that has been allocated to this PHP script (logging is enabled). 

This is my simple memory test (php::memory_get_usage).

```
      yii2                        after patch
```

|-------------------------|------------------------|
| 3.0283966064453  |2.9009246826172|
| 3.041259765625    |2.9009246826172|
| 3.0538177490234  |2.9009246826172|
| .............................. |2.9009246826172|
| .............................. |2.9009246826172|
| .............................. |2.9009246826172|
| 3.977897644043    |2.9009246826172|
| 3.9904556274414  |2.9009246826172|
| 4.0030136108398  |2.9009246826172|
| 3.977897644011    |2.9009246826172|
| 3.8009246826172  |2.9009246826172|

### DIFF
--- a/framework/log/Logger.php
+++ b/framework/log/Logger.php
@@ -124,6 +124,8 @@ class Logger extends Component
         });
     }
 
+    private $_messagesCount = 0;
+
     /**
      * Logs a message with the given type and category.
      * If [[traceLevel]] is greater than 0, additional call stack information about
@@ -153,8 +155,11 @@ class Logger extends Component
                 }
             }
         }
+
         $this->messages[] = [$message, $level, $category, $time, $traces];
-        if ($this->flushInterval > 0 && count($this->messages) >= $this->flushInterval) {
+        $this->_messagesCount++;
+
+        if ($this->flushInterval > 0 && $this->_messagesCount >= $this->flushInterval) {
             $this->flush();
         }
     }


klimov-paul:
How does this fix anything? You should set `flushInterval` to some low value, if you wish to run long-time script.

mongosoft:
I think it, but this patch allows me to work at the default settings. 
In my opinion it is an optimization of this method? Or am I wrong?

cebe:
you somehow try to avoid calling count() on the messages array, how does this relate to memory usage?

mongosoft:
Sorry) I was wrong, I admit it. My code does not improve performance, moreover, my PR contains a mistake, due to which the all logs will be immediately flushed) That's why my test has "worked".
Once more, please forgive me for having taken up your time.

cebe:
> Once more, please forgive me for having taken up your time.

no problem, contribution is always welcome :)

PR #8811, opened 2015-06-16
Adding I18N to deny message on VerbFilter

alemedinam:
I think it should allow Internationalization on deny message


samdark:
The message makes no sense for end user even if it's translated. And, in many cases, translated message won't make any sense if passed to developer.

PR #8812, opened 2015-06-16
check callable

pine3ree:


framework/widgets/ActiveField.php
```diff
@@ -193,7 +193,7 @@ public function render($content = null)
                 $this->hint(null);
             }
             $content = strtr($this->template, $this->parts);
-        } elseif (!is_string($content)) {
+        } elseif (!is_string($content) && is_callable($content)) {
```
klimov-paul: What a sense of extra check?

pine3ree:
@klimov-paul 
Not everything that is not a string a valid parameter for call_user_func. If you use it as a callback you should chek that is callable.

klimov-paul:
So what? At the original code you will get a PHP error, indicating you have passed wrong value. With your solution code will just fail silently. What is the benefit of it?

pine3ree:
Hello @klimov-paul 
i see your point.
In those cases we can either: 1) skip the $content part, 2) return null|empty string 3) throws a customized exception.

I added a custom exception message to help indentify where the wrong input parameter

klimov-paul:
I still don't think an extra check is required here.

PR #8825, opened 2015-06-17
Update QueryBuilder.php

sitawit:
Implement resetSequence function using DBCC CHECKIDENT command

### DIFF
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -71,10 +71,33 @@ class QueryBuilder extends \yii\db\QueryBuilder
         }
     }
 
-//	public function resetSequence($table, $value = null)
-//	{
-//		return '';
-//	}
+    /**
+     * Creates a SQL statement for resetting the sequence value of a table's primary key.
+     * The sequence will be reset such that the primary key of the next new row inserted
+     * will have the specified value or 1.
+     * @param string $tableName the name of the table whose primary key sequence will be reset
+     * @param array|string $value the value for the primary key of the next new row inserted. If this is not set,
+     * the next new row's primary key will have a value 1.
+     * @return string the SQL statement for resetting sequence
+     * @throws NotSupportedException if this is not supported by the underlying DBMS
+     */
+    public function resetSequence($tableName, $value = null)
+    {
+        $table = $this->db->getTableSchema($tableName);
+        if ($table !== null) {
+            $tableName = $this->db->quoteTableName($tableName);
+            if ($value === null) {
+                $key = reset($table->primaryKey);
+                $value = $this->db->createCommand("SELECT MAX(`$key`) FROM $tableName")->queryScalar() + 1;
+            } else {
+                $value = (int) $value;
+            }
+
+            return "DBCC CHECKIDENT ($tableName, RESEED, $value)";
+        } else {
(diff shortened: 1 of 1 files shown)


sitawit:
It seems conflict happens because my forked isn't sync with master. Should I resubmit or something?

samdark:
No, that's OK.

framework/db/mssql/QueryBuilder.php
```diff
@@ -71,10 +71,33 @@ public function buildLimit($limit, $offset = 0)
...
-//	}
+    /**
+     * Creates a SQL statement for resetting the sequence value of a table's primary key.
+     * The sequence will be reset such that the primary key of the next new row inserted
+     * will have the specified value or 1.
+     * @param string $tableName the name of the table whose primary key sequence will be reset
+     * @param array|string $value the value for the primary key of the next new row inserted. If this is not set,
+     * the next new row's primary key will have a value 1.
+     * @return string the SQL statement for resetting sequence
+     * @throws NotSupportedException if this is not supported by the underlying DBMS
```
cebe: this line can be removed because this feature now is supported.

framework/db/mssql/QueryBuilder.php
```diff
@@ -71,10 +71,33 @@ public function buildLimit($limit, $offset = 0)
...
-//	{
-//		return '';
-//	}
+    /**
+     * Creates a SQL statement for resetting the sequence value of a table's primary key.
+     * The sequence will be reset such that the primary key of the next new row inserted
+     * will have the specified value or 1.
+     * @param string $tableName the name of the table whose primary key sequence will be reset
+     * @param array|string $value the value for the primary key of the next new row inserted. If this is not set,
+     * the next new row's primary key will have a value 1.
```
cebe: this comment does not seem to be correct. The implementation finds the maximum + 1 and not just 1.

framework/db/mssql/QueryBuilder.php
```diff
@@ -71,10 +71,33 @@ public function buildLimit($limit, $offset = 0)
...
+     * @throws NotSupportedException if this is not supported by the underlying DBMS
+     */
+    public function resetSequence($tableName, $value = null)
+    {
+        $table = $this->db->getTableSchema($tableName);
+        if ($table !== null) {
+            $tableName = $this->db->quoteTableName($tableName);
+            if ($value === null) {
+                $key = reset($table->primaryKey);
+                $value = $this->db->createCommand("SELECT MAX(`$key`) FROM $tableName")->queryScalar() + 1;
```
cebe: according to the docs at https://msdn.microsoft.com/en-us/library/ms176057.aspx you can even skip this part and just return `DBCC CHECKIDENT ($tableName, RESEED)`, right?

cebe:
@sitawit any feedback to my comments?

sitawit:
@cebe depends on what you would like it to happen when `value` is null. I copied the same idea from the same function of **MySQL** so it has the same behavior. I'm not sure what is the best practice here so I guess removing it is also alright.

And yes, we can use `DBCC CHECKIDENT ($tableName, RESEED)` solely to change current identity value. 

And you're right about the comment, I didn't read it carefully.

samdark:
Need tests to verify it.

Issue #3335, opened 2014-05-03
yii2 migration code style


Asetss:
In yii2 a great migration,on there lame coding style.

```php
$tableOptions = 'CHARACTER SET utf8 COLLATE utf8_general_ci ENGINE=InnoDB';
        // Users table
		$this->createTable('{{%users}}', [
			'id' => Schema::TYPE_PK,
			'username' => Schema::TYPE_STRING . '(30) NOT NULL',  
			'email' => Schema::TYPE_STRING . '(100) NOT NULL',
			'password_hash' => Schema::TYPE_STRING . ' NOT NULL',
			'auth_key' => Schema::TYPE_STRING . '(32) NOT NULL',
			'name' => Schema::TYPE_STRING . '(50) NOT NULL',
			'surname' => Schema::TYPE_STRING . '(50) NOT NULL',
			'avatar_url' => Schema::TYPE_STRING . '(64) NOT NULL',
			'role_id' => 'tinyint NOT NULL DEFAULT 0',
			'status_id' => 'tinyint(4) NOT NULL DEFAULT 0',
			'create_time' => Schema::TYPE_INTEGER . ' NOT NULL',
			'update_time' => Schema::TYPE_INTEGER . ' NOT NULL'
		], $tableOptions);
```
And an example of a new style of coding
```php

			$this->createTable('{{%users}}', [
			'id' => Schema::primaryKey(),
			'username' => Schema::string(30)->notNull(),  
			'email' => Schema::string(100)->notNull(),
			'password_hash' => Schema::string(30)->notNull(),
			'auth_key' => Schema::string(32)->notNull(),
			'name' => Schema::string(50)->notNull(),
			'surname' => Schema::string(50)->notNull(),
			'avatar_url' => Schema::string(64)->notNull(),
			'role_id' => 'tinyint NOT NULL DEFAULT 0',      // thought so tinyint()->notNull()->default(0)
			'status_id' => 'tinyint(4) NOT NULL DEFAULT 0', // thought so tinyint(4)->notNull()->default(0)
			'create_time' => Schema::integer()->notNull(),
			'update_time' => Schema::integer()->notNull()
		], $tableOptions);
```

Is an example of what I mean. What do you think about this?

ifocus22:
I found this to be useful with Rails. 

$ bin/rails generate scaffold HighScore game:string score:integer
    invoke  active_record
    create    db/migrate/20130717151933_create_high_scores.rb

$ bin/rails generate model
[...]

PR #8833, opened 2015-06-18
New migration code style issue #3335

vaseninm:
Hi. It new migration code style from issue #3335. 
I wrote and successfully launched tests for mysql, postgres and sqllite. 
I wrote test for cubrid and oracle, but i write on macos and can't install it.
I don't wrote php docs, because my english is bad. If my code is good for yii2, i write docs or somebody help me.
Review my code, plz.


framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\db;
+
+use Yii;
+use yii\base\Object;
+
```
pana1990: Add @author and @since

framework/db/SchemaBuilderTrait.php
```diff
@@ -0,0 +1,128 @@
...
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\db;
+
+use Yii;
+use yii\base\Object;
+use yii\di\Instance;
+
```
pana1990: Add @author and @since  again

framework/db/cubrid/SchemaBuilder.php
```diff
@@ -0,0 +1,16 @@
...
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\db\cubrid;
+
+
+use yii\db\SchemaBuilder as AbstractSchemaBuilder;
+
```
pana1990: Add @author and @since  again

pana1990:
it is very nice, thanks for your job :)

Note : Add @author and @since  in files class.

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+    {
+        return static::createDefault(Schema::TYPE_STRING, $length);
+    }
+
+    public static function text($length = null)
+    {
+        return static::createDefault(Schema::TYPE_TEXT, $length);
+    }
+
+    public static function smallint($length = null)
```
pana1990: it should be camelCase `smallInt`

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+    {
+        return static::createNumeric(Schema::TYPE_DOUBLE, $precision, $scale);
+    }
+
+    public static function decimal($precision = null, $scale = null)
+    {
+        return static::createNumeric(Schema::TYPE_DECIMAL, $precision, $scale);
+    }
+
+    public static function datetime($length = null)
```
pana1990: it should be camelCase `dateTime`

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    public function notNull()
+    {
+        $this->isNull = false;
+
+        return $this;
+    }
+
+    public function null()
```
pana1990: Columns are nullable by default, why do you create this method?

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    public function setDefault($default = null)
+    {
+        $this->default = $default;
+
+        return $this;
+    }
+
+    public function check($check)
```
pana1990: it should be `null` by default

framework/db/SchemaBuilderTrait.php
```diff
@@ -0,0 +1,128 @@
...
+trait SchemaBuilderTrait
+{
+
+    private static $dbName = 'db';
```
pana1990: The property names must start with an initial underscore if they are private

framework/db/oci/SchemaBuilder.php
```diff
@@ -0,0 +1,24 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\db\ocl;
```
pana1990: `yii\db\oci` instead `yii\db\ocl`

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,204 @@
...
+    {
+        return static::createDefault(Schema::TYPE_TEXT, $length);
+    }
+
+    public static function smallint($length = null)
+    {
+        return static::createDefault(Schema::TYPE_SMALLINT, $length);
+    }
+
+    public static function integer($length = null)
```
vaseninm: maybe int or smallInteger and bigInteger?

framework/db/SchemaBuilder.php
vaseninm: @pana1990 check already null by default
``
    protected $check = null;
``
and i think, it wrong to use check without params.

pana1990: Of course ;)

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    public function notNull()
+    {
+        $this->isNotNull = true;
+
+        return $this;
+    }
+
+    public function setDefault($default = null)
```
pana1990: Can you change `setDefault` to `default`?

vaseninm: http://stackoverflow.com/questions/10395930/cant-have-function-named-default-in-php
+ @method in phpdoc?

framework/db/SchemaBuilder.php
samdark: Nah, just integer is OK.

samdark: Or it could be int, yes. A bit shorter but recognizable.

framework/db/SchemaBuilder.php
samdark: Yes. I think it worth doing via `__call()`. Performance isn't concern in this case.

tests/framework/db/QueryBuilderTest.php
```diff
@@ -25,19 +26,23 @@ class QueryBuilderTest extends DatabaseTestCase
      */
     protected function getQueryBuilder()
     {
+        $connection = $this->getConnection(true, false);
+
+        \Yii::$container->set('db', $connection);
```
samdark: Why do we need to replace it in container?

samdark:
Also phpdoc is needed for each method added. Tests are passing btw.

Great job! I'll try to get this one in 2.0.5.

tests/framework/db/QueryBuilderTest.php
vaseninm: Сontainer does not have 'db' on this test case.

samdark:
I've meant descriptive phpdoc, not just for the sake of presence of some doc :) But if you don't have time for it, I can handle it.

vaseninm:
@samdark i understand you. It was intermediate commit. I have time for writing phpdocs, but my english is my problem and it takes some time.

samdark:
Don't worry, we'll review it.

pana1990:
@samdark  i have some questions.

$_dbName in SchemaBuilderTrait should be public?

in SchemaBuilderTrait should be refactor with __callStatic method magic and we use @method tag for It must support autocomplete in IDE

here my propositional: https://github.com/vaseninm/yii2/pull/1

pana1990:
i think to it should be config db component in migration.php, for example : 

```php
// framework/console/controllers/MigrateController.php
    {
        if (parent::beforeAction($action)) {
            if ($action->id !== 'create') {
                $this->db = Instance::ensure($this->db, Connection::className());
                Schema::setDb($this->db); // Here set up db component
            }
            return true;
        } else {
            return false;
        }
    }
```

@samdark what do you thing about it?

vaseninm:
I fix tomorrow. Please, review php docs and magic methods in SchemaBuilderTrait.php.
Thanks @pana1990 for phpdocs.

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,409 @@
...
+ *
+ * @method SchemaBuilder default($default = null) see [[SchemaBuilder::_default()]] for more info
+ *
+ * @author Vasenin Matvey <vaseninm@gmail.com>
+ * @since 2.0.5
+ */
+abstract class SchemaBuilder extends Object
+{
+    /**
+     * @var string schema of column
```
samdark: schema of **a** column

samdark: or "column schema"

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,409 @@
...
+     */
+    protected $schema = null;
+    /**
+     * @var integer size of column
```
samdark: column size

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,409 @@
...
+     */
+    protected $length = null;
+    /**
+     * @var boolean whether may column value be not null
```
samdark: whether value could be null

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,409 @@
...
+     * @var string check value of column
+     */
+    protected $check = null;
+    /**
+     * @var mixed default value of column
+     */
+    protected $default = null;
+
+    /**
+     * Specify type of field as primary key
```
samdark: Makes column a primary key

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,409 @@
...
+     * will be implicitly called when an unknown method is being invoked.
+     * @param string $name the method name
+     * @param array $params method parameters
+     * @return mixed the method return value
+     */
+    public function __call($name, $arguments)
+    {
+        if ($name === 'default') {
+            return call_user_func_array([$this, '_default'], $arguments);
+        }
```
samdark: You should call parent class method afterwards.

pana1990:
i created pull request [here](https://github.com/vaseninm/yii2/pull/2) that fix phpdocs.

@vaseninm to fix the error must be added to the timestamp method the $length param

vaseninm:
done

samdark:
Looks good to me now. @yiisoft/core-developers any comments?

pana1990:
I think that we should assign $_dbName when developer set $db in migration.

For example, i execute the following conmands console : 
```
yii migrate/to m150101_185401_create_news_table --db=db1
yii migrate/to m150101_185401_create_news_table --db=db2
```
How you set up db component in case use?

i think to it should be config in migration.php, for example : 

```php
// framework/console/controllers/MigrateController.php
    {
        if (parent::beforeAction($action)) {
            if ($action->id !== 'create') {
                $this->db = Instance::ensure($this->db, Connection::className());
                Schema::setDb($this->db); // Here set up db component
            }
            return true;
        } else {
            return false;
        }
    }
```

@samdark  What do you think about it?

framework/db/SchemaBuilderTrait.php
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * @param Connection|array|string $dbName the DB connection object or the application component ID of the DB connection
+     */
+    public static function setDb($dbName)
+    {
+        self::$_dbName = $dbName;
+    }
+
+    /**
+     * Calls ethe named static method which is not a class method.
```
cebe: typo `ethe`

tests/framework/db/QueryBuilderTest.php
```diff
@@ -62,77 +67,81 @@ protected function replaceQuotes($sql)
...
+            [Schema::TYPE_DATE . " CHECK (value BETWEEN '2011-01-01' AND '2013-01-01')", Schema::date()->check("value BETWEEN '2011-01-01' AND '2013-01-01'"), "date CHECK (value BETWEEN '2011-01-01' AND '2013-01-01')"],
+            [Schema::TYPE_DATE . ' NOT NULL', Schema::date()->notNull(), 'date NOT NULL'],
+            [Schema::TYPE_BINARY, Schema::binary(), 'blob'],
+            [Schema::TYPE_BOOLEAN, Schema::boolean(), 'tinyint(1)'],
+            [Schema::TYPE_BOOLEAN . ' NOT NULL DEFAULT 1', Schema::boolean()->notNull()->default(1), 'tinyint(1) NOT NULL DEFAULT 1'],
+            [Schema::TYPE_MONEY, Schema::money(), 'decimal(19,4)'],
+            [Schema::TYPE_MONEY . '(16,2)', Schema::money(16, 2), 'decimal(16,2)'],
+            [Schema::TYPE_MONEY . ' CHECK (value > 0.0)', Schema::money()->check('value > 0.0'), 'decimal(19,4) CHECK (value > 0.0)'],
+            [Schema::TYPE_MONEY . '(16,2) CHECK (value > 0.0)', Schema::money(16, 2)->check('value > 0.0'), 'decimal(16,2) CHECK (value > 0.0)'],
+            [Schema::TYPE_MONEY . ' NOT NULL', Schema::money()->notNull(), 'decimal(19,4) NOT NULL'],
```
cebe: why do you replace explicit tests with return value of functions? this way the test is useless as it can not detect issues which happen to occur on left and right side code.
original tests should stay, you may add others additionally.

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,412 @@
...
+     * Do not call this method directly as it is a PHP magic method that
+     * will be implicitly called when an unknown method is being invoked.
+     * @param string $name the method name
+     * @param array $arguments method parameters
+     * @return mixed the method return value
+     */
+    public function __call($name, $arguments)
+    {
+        if ($name === 'default') {
+            return call_user_func_array([$this, '_default'], $arguments);
```
cebe: what is this used for?

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,412 @@
...
+ *
+ * @author Vasenin Matvey <vaseninm@gmail.com>
+ * @since 2.0.5
+ */
+abstract class SchemaBuilder extends Object
+{
+    /**
+     * @var string column schema
+     */
+    protected $schema = null;
```
cebe: better name: `$type`, this is not the schema but the column data type.

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,412 @@
...
+    /**
+     * @var integer column size
+     */
+    protected $length = null;
+    /**
+     * @var boolean whether value could be null
+     */
+    protected $isNotNull = null;
```
cebe: what is the semantic of `null` here? a column is either nullable or not. The `is` prefix is also confusing. a column is not null itself, it is nullable.

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,412 @@
...
+
+        return $string;
+    }
+
+    /**
+     * Returns check value of column
+     *
+     * @return string
+     */
+    protected function getCheckString()
```
cebe: for these methods the nameing would be better `build...` i.e. `buildCheckString()` as this is what they do. analog as it is done in the querybuilder.

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    /**
+     * Create schema builder for types with length
+     *
+     * @param string $type schema of column
+     * @param integer $length length of column
+     * @return SchemaBuilder
+     */
+    protected static function createDefault($type, $length = null)
```
cebe: maybe just `create`? "default" is already the default value of the colum so a bit confusing.

framework/db/SchemaBuilder.php
```diff
@@ -0,0 +1,412 @@
...
+use Yii;
+use yii\base\Object;
+
+/**
+ * SchemaBuilder is the class help to define DB's schema types.
+ *
+ * For example you may use the following code inside your migration files:
+ *
+ * ```php
+ * $this->createTable('{{table}}', [
```
cebe: quoting of table name is not needed here.

cebe:
In general a good solution but I do not like the idea of setting the database schema statically. It would be better to get the schema builder from the db schema itself.

I see no reason why it is bound to the Schema class with static methods and a trait, we could just attach the trait to the migraiton and do something like:

```php
$this->createTable('table_name', [
    'name' => $this->string(64)->notNull(),
    'type' => $this->integer()->notNull()->default(10),
]);
// or maybe with a prefix
$this->createTable('table_name', [
    'name' => $this->columnString(64)->notNull(),
    'type' => $this->columnInteger()->notNull()->default(10),
    'description' => $this->columnText(),
]);
```

this would ensure it is always the correct schema syntax for the migration.

framework/db/SchemaBuilder.php
vaseninm: We discussed about it here. We want write method with name 'default', but it reserved in PHP.

vaseninm:
@cebe thanks, it's good remarks in code.
About Schema::string() vs $this->string(). I think, that Schema:: is regular for yii coders and also allows to use outside the migration.

framework/db/SchemaBuilder.php
samdark: It's fine as it currently is in code. I've verified there's no other way to do it.

cebe: what about using `defaultValue` then? I think it is better to avoid reserved names instead of building workarounds for them.

vaseninm:
@samdark I must do something else?

samdark:
I think no. I have to verify everything...

samdark:
Merged with a bit of cleanup: https://github.com/yiisoft/yii2/commit/8e1d982da4d145590c429f1fe4aef562821415f7

samdark:
@vaseninm, @pana1990 thank you very much for such a great addition!

samdark:
https://github.com/yiisoft/yii2-app-advanced/commit/d6ae6d472f421f8ef279e19241890f0e410d512e

kartik-v:
:+1: Well done and good addition to yii...

cebe:
Hm... my concern about static setting of db connection is still valid (https://github.com/yiisoft/yii2/pull/8833#issuecomment-114667691)

@samdark any comments?

samdark:
@cebe how would you use it outside of migrations?

pana1990:
we should update https://github.com/yiisoft/yii2/blob/master/docs/guide/db-migrations.md imo

samdark:
Absolutely.

samdark:
https://github.com/yiisoft/yii2/commit/3c9e6be413df2b1e74748ac902137d35861d74a0

githubjeka:
Hi!
How to create `unsigned`?

```php
$this->integer(10)->defaultValue(13)
```

samdark:
There's no such ability now.

githubjeka:
Ok, thanks.

PR #8834, opened 2015-06-18
Update QueryBuilder.php

sitawit:
### Background Problem

In case of SQL Server. The function checkIntegrity function of [InitDbFixture](https://github.com/yiisoft/yii2/blob/master/framework/test/InitDbFixture.php) will always result in 

```
Table not found: []
```
### Cause

The function won't specify a table which needs to be checked.
### Solution

Update checkIntegrity function to enable or disable integrity checking to all tables in case of no table specify. This update makes aforementioned function can enable or disable integrity check on all tables when initializing fixtures. Moreover, it will have the same behavior with other database types. MySQL and SQLite will always perform on all tables. PgSQL will perform on all tables in case of no table specify. The rest of database types don't support.
### Note

The description checkIntegrity of [QueryBuilder](https://github.com/yiisoft/yii2/blob/master/framework/db/QueryBuilder.php) say:

```
@param string $table the table name. Defaults to empty string, meaning that no table will be changed.
```

Which is no longer relevant because all overrides functions of each database types likely to perform integrity check in case of no table specify.

### DIFF
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -122,12 +122,16 @@ class QueryBuilder extends \yii\db\QueryBuilder
      * Builds a SQL statement for enabling or disabling integrity check.
      * @param boolean $check whether to turn on or off the integrity check.
      * @param string $schema the schema of the tables. Defaults to empty string, meaning the current or default schema.
-     * @param string $table the table name. Defaults to empty string, meaning that no table will be changed.
+     * @param string $table the table name. Defaults to empty string, meaning that all tables will be changed.
      * @return string the SQL statement for checking integrity
      * @throws InvalidParamException if the table does not exist or there is no sequence associated with the table.
      */
     public function checkIntegrity($check = true, $schema = '', $table = '')
     {
+        $enable = $check ? 'CHECK' : 'NOCHECK';
+        if ($table === '') {
+    		return "EXEC sp_MSforeachtable 'ALTER TABLE ? {$enable} CONSTRAINT ALL'";
+    	}
         if ($schema !== '') {
             $table = "{$schema}.{$table}";
         }
@@ -135,8 +139,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
         if ($this->db->getTableSchema($table) === null) {
             throw new InvalidParamException("Table not found: $table");
         }
-        $enable = $check ? 'CHECK' : 'NOCHECK';
-
+        
         return "ALTER TABLE {$table} {$enable} CONSTRAINT ALL";
     }
 


framework/db/mssql/QueryBuilder.php
```diff
@@ -122,12 +122,16 @@ public function alterColumn($table, $column, $type)
...
-     * @param string $table the table name. Defaults to empty string, meaning that no table will be changed.
+     * @param string $table the table name. Defaults to empty string, meaning that all tables will be changed.
      * @return string the SQL statement for checking integrity
      * @throws InvalidParamException if the table does not exist or there is no sequence associated with the table.
      */
     public function checkIntegrity($check = true, $schema = '', $table = '')
     {
+        $enable = $check ? 'CHECK' : 'NOCHECK';
+        if ($table === '') {
+    		return "EXEC sp_MSforeachtable 'ALTER TABLE ? {$enable} CONSTRAINT ALL'";
```
cebe: this seems to be an unsupported API, is there a way to do it in an officially supported way?

>  It is our general policy not to provide documentation to internal unsupported API like this one.
> https://connect.microsoft.com/SQLServer/feedback/details/264676/sp-msforeachtable-provide-supported-documented-version

sitawit:
@cebe We can use [sys.tables](https://msdn.microsoft.com/en-us/library/ms187406.aspx) to get all tables' name and then apply *ALTER* to each table instead. What do you think?

cebe:
how would that look like? (I have never worked with MSSQL)

If sys.tables is too complicated I am fine with the current solution. Seems to be quite common according to stackoverflow: http://stackoverflow.com/questions/159038/how-can-foreign-key-constraints-be-temporarily-disabled-using-t-sql/161410#161410

sitawit:
After review the code, we can use existing code to get all tables' name instead of using other sql command. And then we can recursive this function to get multiple sql for these tables.

However, using `getTableNames` function I'm not so sure if we need to refresh the tables' name or not. So I leave it using default value.

And also `WITH CHECK` is added.

```php
    public function checkIntegrity($check = true, $schema = '', $table = '')
    {
    	if ($table === '') {
    		$tableNames = $this->db->getSchema()->getTableNames($schema);
    		$sql = '';
    		foreach ($tableNames as $tableName) {
    			$sql .= $this->checkIntegrity($check, $schema, $tableName) . ';';
    		}
    		return $sql;
    	}
        if ($schema !== '') {
            $table = "{$schema}.{$table}";
        }
        $table = $this->db->quoteTableName($table);
        if ($this->db->getTableSchema($table) === null) {
            throw new InvalidParamException("Table not found: $table");
        }
        $enable = $check ? 'WITH CHECK CHECK' : 'NOCHECK';
        return "ALTER TABLE {$table} {$enable} CONSTRAINT ALL";
    }
```

cebe:
is the mssql PDO driver able to handle multiple queries in one execute call? If so the solution looks good to me.

samdark:
That's a very good question because [for MySQL it's really tricky](http://stackoverflow.com/questions/6346674/pdo-support-for-multiple-queries-pdo-mysql-pdo-mysqlnd). It should be checked in both emulation mode and normal mode.

sitawit:
I did some read so I came up with something.
To execute multiple queries at the same time. In MSSQL, three are 2 approaches.

#### Using : Go

[Go](https://msdn.microsoft.com/en-us/library/ms188037.aspx) is the MSSQL specific command. It can separate each [*batch*](https://msdn.microsoft.com/en-us/library/ms712553(v=vs.85).aspx) of SQL statement. One batch could look like this.
```
INSERT INTO Orders (OrderID, CustID, OpenDate, SalesPerson, Status)
   VALUES (2002, 1001, {fn CURDATE()}, 'Garcia', 'OPEN');
INSERT INTO Lines (OrderID, Line, PartID, Quantity)
   VALUES (2002, 1, 1234, 10);
INSERT INTO Lines (OrderID, Line, PartID, Quantity)
   VALUES (2002, 2, 987, 8);
```

And using `Go` is like this
```
CREATE TABLE dbo.t3(a int) ;
INSERT INTO dbo.t3 VALUES (1) ;
INSERT INTO dbo.t3 VALUES (1,1) ;
INSERT INTO dbo.t3 VALUES (3) ;
GO

SELECT * FROM dbo.t3 ;
```

The pros of doing this, if one of the batch fails, the rest of batch can still execute.

#### Using : Semi-colon (;)

Semi-colon is a statement separator. Just like any other T-SQL script. But if one of the statement fails, the rest will stop executing.

#### Different
There are 2 good explanation here. And also more info about these 2 approaches.
- http://stackoverflow.com/questions/3701147/in-sql-server-when-should-you-use-go-and-when-should-you-use-semi-colon
- http://stackoverflow.com/questions/1517527/what-is-the-difference-between-and-go-in-tsql-sql-2008

#### Conclusion
If we want the `ALTER` command to continue in case of one is fail, `Go` will be used to separate the command. But if we want the whole to stop using `;` is enough.

So what do you think.

samdark:
Maybe let's try sending a lot of separate SQL statements at first and check if performance is acceptable?

sitawit:
It's been a while... so here is an update.

**Q**: Is mssql PDO can handle multiple query?
**A**: Yes

**Q**: How about the performance?
**A**: I haven't wrote any testing script. However, I did try to use multiple queries in mssql and try to use its analyze tool. The cost of multiple alter queries are really low. 

**Q**: Should we use So Semi-colon (;) or Go
**A**: Use semi-colon is required in the near future according to this:- 
https://technet.microsoft.com/en-us/library/ms143729.aspx.
`Not ending Transact-SQL statements with a semicolon` will be deprecated. 
Currently, MSSQL is version 2014 but soon it will be 2016. So we have to use semi-colon from now on. And it has full backward competible.

samdark:
Need tests to verify it.

dynasource:
we need unit tests to continue this PR.

PR #8842, opened 2015-06-18
Fixed #8779: Automatically set enctype form option when using file in…

### LINKED ISSUE #8779: Automatically set enctype form option when using file input field
What if when using file input like that:

```php
<?= $form->field($model, 'imgFile')->fileInput() ?>
```

the `enctype` will be set to `multipart/form-data` automatically?

pana1990:
https://github.com/yiisoft/yii2/issues/8779


klimov-paul:
Fix is not correct, because if form have only file inputs there is no need to set `enctype` for it.

pana1990:
@klimov-paul  what is the problem?

For example, Ruby on Rails works well .

view in rails : 

screenshot_1

browser : 
screenshot_2

thanks in advance for your answer.

klimov-paul:
It will work, but following your logic we can just add 'enctype' to all forms by default, which by the way can be done via DI.

pana1990:
But when i use fileinput  helper i always want to add enctype.

I think it is expected behavior in my humble opinion.

samdark:
I think the change is OK.

githubjeka:
Up.

It's the normal fix.

mdmunir:
:+1:

pana1990:
@samdark @klimov-paul @cebe i think to ready to merge

samdark:
https://github.com/yiisoft/yii2/commit/7dc984d35928128b3612fb6aa4ed19b03ee963fd merged. Thank you!

PR #8844, opened 2015-06-19
Fixed #8843: 修正 Oracle 11g 11.2.0.1.0 ORA-03113: end-of-file on communication channel 錯誤

### LINKED ISSUE #8843: 使用 Oracle ActiveRecord Model 顯示錯誤 ORA-03113: end-of-file on communication channel
PHP 5.6.10
Yii2 2.0.4
Oracle Database 11g Enterprise Edition Release 11.2.0.1.0 - 64bit Production

無論是使用 Gii 或是自己寫 Oracle ActiveRecord Model

執行時都會顯示錯誤 ORA-03113: end-of-file on communication channel

nidgetgod:
Fixed #8843: 修正 Oracle 11g 11.2.0.1.0 ORA-03113: end-of-file on communication channel 錯誤


cebe:
can you try to explain the problem you are fixing in english?

nidgetgod:
When using ActiveRecord Model created by manual or using Gii to create ActiveRecord Model, both will cause "ORA-03113: end-of-file on communication channel" error.

This error seems only happened in Oracle 11.2.0.1.0.

nineinchnick:
How does it help to fetch more data and filter it in PHP instead of directly in Oracle?

nidgetgod:
In [commit 273b6da](https://github.com/nidgetgod/yii2/commit/273b6da2dac946e1d71c6ad0bfdb0507a3f1af1c) I added D.CONSTRAINT_TYPE select column.

So [in the next part of foreach loop](https://github.com/nidgetgod/yii2/commit/1c7c57681a398f3b469c85a18200c45595c4316c), if CONSTRAINT_TYPE value isn't 'R' then continue filter next element.

nineinchnick:
@nidgetgod please don't describe the diff, because we can see it. Explain why is it necessary.

nidgetgod:
Because without  "AND D.CONSTRAINT_TYPE = 'R'" the foreache loop result will totally different.

nineinchnick:
I don't understand how is the SQL condition different than the PHP one. It seems they would produce exactly same results. And when you reported the issue you wrote that Oracle throws an error, not that it produces different results.

nidgetgod:
That's the tricky part. If I using sqlplus and execute the original sql it will cause "ORA-03113: end-of-file on communication channel" with no reason, but when I remove "AND D.CONSTRAINT_TYPE = 'R'" and it works, no more error.

nidgetgod:
FYI

Here's what I did reseach before and found [Issue 1779:	Work around for oracle 11g bug](https://code.google.com/p/yii/issues/detail?id=1779).

This issue happened in Yii(not Yii2) too.

nineinchnick:
Thanks. Can you add comments in the code explaining this is an Oracle bug? Otherwise someone could revert those changes in the future.

cebe:
Merged, thanks!

PR #2378, opened 2014-02-09
WIP, a message queue interface and three implementations

nineinchnick:
Resolves #492. **WIP - do not merge, request for comments.**

Introducing a generic message queue interface with following features:
- A single message queue as an application component
- Three implementations: System V, Redis and DB (SQL)
- Publish/Subscribe (delivery to multiple queues) with filtering by categories
- Peeking and reserving/locking messages with a timeout
- Blocking mode to allow running a consumer/worker that waits for messages
- Yii events

The design is suppose to allow creating job queues as well as messaging, notifications, logging, audits etc.

This PR is a request for comments. It means there are a few design issues to discuss:
1. Blocking
   Blocking is required to be able to implement an efficient consumer fired up as a deamon that waits for arriving messages. This is   implemented in the Redis and SysV queues and could be done for Db queue but requires a db engine specific sql expressions (locking,     notifications in PostgreSQL and queus in MsSQL).
   Currently it's a property of the queue component. Should it be an argument to the send/receive methods?
2. Db implementation
   Should AR models be used? They do simplify the queue interface implementation by removing raw SQL from it. Can the namespace hierarchy be improved? 
3. Separating PubSub interface
   Should the PubSub methods be extracted into a separate interface? A good example is the SysV implementation, which provide only a   single queue. It could implement the PubSub interface by storing subscriptions somewhere else, ie a database.
4. Should a console controller be included?
5. Are there any other implementations that should be available?

Note: db implementation is important to be able to use in a local dev enviroment (like mail with file delivery) and in enviroments where services like Redis cannot be installed (shared hosting). It CAN be acceptable in low traffic/congestion systems.

Note: the code here may not actually run. It has been ported into the framework from the [yii2-nfy module](https://github.com/nineinchnick/yii2-nfy) for discussion of the basic design.

Tests will be added after current design will be accepted and/or changes will be made.

Note: the pubsub in Redis requires resolving the #1932 issue first.

Issue #492, opened 2013-06-04
Provide queue component


samdark:
Useful for background jobs and all projects processing relatively large amounts of data.

Implementation can be designed the same way as cache i.e. base API + drivers such as:

- "immediate execution" (for local development)
- db
- http://aws.amazon.com/sqs/

Proposed base API:

```php
abstract class Queue {
  // add job to the queue. $data will be passed as an argument to $callback and should be serializable
  public function push($callback, $data)
  {
  }

  // get a job and lock it queue for further processing
  public function get()
  {
  }
}
```

sergebezborodov:
http://kr.github.io/beanstalkd/ is good solution for small-medium size projects. Has full control on tubes and tasks, really fast.
Here is my old implementation https://github.com/sergebezborodov/beanstalk-yii2

lubosdz:
+1 if this can support concurrent/parallel jobs

Just a hint on possible implementation details:
* [Cooperative multitasking via generators by Nikita Popov](https://nikic.github.io/2012/12/22/Cooperative-multitasking-using-coroutines-in-PHP.html)
* [PHP multithreading](http://stackoverflow.com/questions/70855/how-can-one-use-multi-threading-in-php-applications)

samdark:
It's now managed in https://github.com/yiisoft/yii2-queue

nineinchnick:
@samdark I guess I should start splitting and moving my #2378 PR there too

samdark:
@nineinchnick check what's there first.

PR #8864, opened 2015-06-22
[WIP] Queue components

creocoder:
Usage from application side:

``` php
// use plain payload
Yii::$app->queue->push('Some message');

// use job payload
SendEmailJob::push();
```

Concrette job classes:

``` php
class SendEmailJob extends \yii\queue\Job
{
    /**
     * @var common\models\User
     */
    protected $user;

    /**
     * @inheritdoc
     */
    public function run()
    {
        // ...
    }
}
```

Workers:

``` sh
./yii queue/listen // listen default back-end, default pool
./yii queue/listen rabbitmq // listen rabbitmq back-end, default pool
./yii queue/listen rabbitmq --queue=emails //listen rabbitmq back-end, emals pool
```

### DIFF
--- a/framework/console/controllers/QueueController.php
+++ b/framework/console/controllers/QueueController.php
@@ -1,0 +1,21 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\console\controllers;
+
+use Yii;
+use yii\console\Controller;
+
+/**
+ * Runs the queue listener.
+ *
+ * @author Alexander Kochetov <creocoder@gmail.com>
+ * @since 2.0.5
+ */
+class QueueController extends Controller
+{
+}
--- a/framework/queue/Job.php
+++ b/framework/queue/Job.php
@@ -1,0 +1,20 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\queue;
+
+use yii\base\Object;
+
+/**
+ * Job is the base class for queue job classes.
+ *
+ * @author Alexander Kochetov <creocoder@gmail.com>
+ * @since 2.0.5
+ */
+abstract class Job extends Object
+{
+}
--- a/framework/queue/Job.php
+++ b/framework/queue/Job.php
@@ -7,6 +7,7 @@
 
 namespace yii\queue;
 
+use Yii;
 use yii\base\Object;
 
 /**
@@ -17,4 +18,25 @@ use yii\base\Object;
  */
 abstract class Job extends Object
 {
+    /**
+     * Returns the queue connection used by this Job class.
+     * By default, the "queue" application component is used as the queue connection.
+     * You may override this method if you want to use a different queue connection.
+     * @return Queue the queue connection used by this Job class.
+     */
+    public static function getQueue()
+    {
+        return Yii::$app->getQueue();
+    }
+
+    /**
+     * Push this job onto the queue.
+     *
+     * @param string|null $queue
+     * @param integer $delay
+     */
+    public static function push($queue = null, $delay = 0)
+    {
+        // implementation
+    }
 }
--- a/framework/queue/Queue.php
+++ b/framework/queue/Queue.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\queue;
+
+use yii\base\Component;
+
+/**
+ * Queue is the base class for queue classes supporting different queue back-ends.
+ *
+ * @author Alexander Kochetov <creocoder@gmail.com>
+ * @since 2.0.5
+ */
+abstract class Queue extends Component
+{
+    /**
+     * Push a new job to the queue.
+     *
+     * @param string $job
+     * @param mixed|null $data
+     * @param string|null $queue
+     * @param integer $delay
+     */
+    abstract public function push($job, $data = null, $queue = null, $delay = 0);
+
+    /**
+     * Pop the next job off of the queue.
+     *
+     * @param string|null $queue
+     *
+     * @return Job
+     */
+    abstract public function pop($queue = null);
+
+    /**
+     * Return the size (number of pending jobs) of the specified queue.
+     *
+     * @param string|null $queue
+     *
+     * @return integer
+     */
+    abstract public function size($queue = null);
+}
--- a/framework/queue/Queue.php
+++ b/framework/queue/Queue.php
@@ -18,30 +18,20 @@ use yii\base\Component;
 abstract class Queue extends Component
 {
     /**
-     * Push a new job to the queue.
+     * Push a new message to the queue.
      *
-     * @param string $job
-     * @param mixed|null $data
+     * @param string $message
      * @param string|null $queue
      * @param integer $delay
      */
-    abstract public function push($job, $data = null, $queue = null, $delay = 0);
+    abstract public function push($message, $queue = null, $delay = 0);
 
     /**
-     * Pop the next job off of the queue.
+     * Pop the next message off of the queue.
      *
      * @param string|null $queue
      *
-     * @return Job
+     * @return string
      */
     abstract public function pop($queue = null);
-
-    /**
-     * Return the size (number of pending jobs) of the specified queue.
-     *
-     * @param string|null $queue
-     *
-     * @return integer
-     */
-    abstract public function size($queue = null);
 }
--- a/framework/base/Application.php
+++ b/framework/base/Application.php
@@ -19,6 +19,8 @@ use Yii;
  * @property string $basePath The root directory of the application.
  * @property \yii\caching\Cache $cache The cache application component. Null if the component is not enabled.
  * This property is read-only.
+ * @property \yii\queue\Queue $queue The queue application component. Null if the component is not enabled.
+ * This property is read-only.
  * @property \yii\db\Connection $db The database connection. This property is read-only.
  * @property \yii\web\ErrorHandler|\yii\console\ErrorHandler $errorHandler The error handler application
  * component. This property is read-only.
@@ -516,6 +518,15 @@ abstract class Application extends Module
     public function getCache()
     {
         return $this->get('cache', false);
+    }
+
+    /**
+     * Returns the queue component.
+     * @return \yii\queue\Queue the queue application component. Null if the component is not enabled.
+     */
+    public function getQueue()
+    {
+        return $this->get('queue', false);
     }
 
     /**


framework/console/controllers/QueueController.php
```diff
@@ -0,0 +1,21 @@
...
+use Yii;
+use yii\console\Controller;
+
+/**
+ * Runs the queue listener.
+ *
+ * @author Alexander Kochetov <creocoder@gmail.com>
+ * @since 2.0.5
+ */
+class QueueController extends Controller
```
klimov-paul: Such class seems to be redundant.

framework/queue/Job.php
```diff
@@ -0,0 +1,20 @@
...
+
+use yii\base\Object;
+
+/**
+ * Job is the base class for queue job classes.
+ *
+ * @author Alexander Kochetov <creocoder@gmail.com>
+ * @since 2.0.5
+ */
+abstract class Job extends Object
```
klimov-paul: Why this is needed?
Job itself should not be exposed at public interface. If particular queue implemetation requires extra class for the Job, it can add it and use internally.

framework/queue/Queue.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * Push a new job to the queue.
+     *
+     * @param string $job
+     * @param mixed|null $data
+     * @param string|null $queue
+     * @param integer $delay
+     */
+    abstract public function push($job, $data = null, $queue = null, $delay = 0);
```
klimov-paul: Why pass `queue` as a parameter to the method of the `Queue` object?

framework/console/controllers/QueueController.php
creocoder: @klimov-paul Why you think so? To be honest its main class with 90% of functionality. Example:

```
./yii queue/listen
./yii queue/listen --queue=emails
./yii queue/listen --queue=avatars
```

klimov-paul: At present state this controller is empty: in such state it does not make any sense.
Usually queues are processed by deamons - not console commands. Still it could be implemented by console command running by cron with mutex.
Either way - it is better to be decided by developer.

framework/queue/Job.php
creocoder: @klimov-paul Let ask that question after see implementation. Btw do you saw any modern queue components implementation? Please check one.

framework/queue/Queue.php
creocoder: Because `Yii::$app->queue` is queue back-end, while `queue` param is concrette pool. This term is used in around 9 implementations which i checked and in every AMQ documentation. So renaming it to something else === complettely confuse people.

framework/console/controllers/QueueController.php
creocoder: > Usually queues are processed by deamons - not console commands.

Ok, so you think daemon on php is impossible. Ok, maybe its time to start checking other queue implementations on php before provice such strange feedback?

framework/queue/Queue.php
samdark: I confirm that it's common AMQP naming. Better not to change it.

creocoder:
@klimov-paul I updated first post description with use case examples.

framework/console/controllers/QueueController.php
klimov-paul: > you think daemon on php is impossible

Just where I have told you daemon on PHP is impossible?
All I say: there is no need for predefined controller for the queue processing. Developer should decide, if he want it inside console command or inside daemon or anywhere else.
That is why I said at the begining:

> Such class seems to be redundant.

framework/queue/Queue.php
klimov-paul: Then why not name `Yii::$app->queue` as `QueueManager` or something else to avoid confusion?

klimov-paul:
> Usage from application side...

Well, perhaps I am indeed missing modern queue approach. But for me this is overcomplecate things.
Why bind queue job to the intrenal application classes?
I still remember the project, where I was using Amazon SQS: back then I passed queue messages (jobs) via SQS to .NET (!) application, which is obviously don't know anything about PHP classes or PHP `serialize()`.
Binding with `Job` classes is acceptable as high-level implementation, but low level (using simple strings) should be exposed as well.

> class Job extends Object

If you wish `Job` be some active queue processor, perhaps it is better to inherited from `yii\base\Model` allowing automatic validation of its parameters before send to the queue or/and before actual execution.

klimov-paul:
Comparing this solution to `yii\db\*`, I would say it provides `ActiveRecord` usage level, while DAO is missing.

creocoder:
@klimov-paul 

> But for me this is overcomplecate things.

So you suggest push abstract **plain** messages to queues and maybe write workers on C/C++? Every abstraction around queue back-end i was investigated for those weeks (i watch lot of them) has high-level Job class. No one suggest push abstract plan messages to queues because its absoluttely does not make sence for practice usage, while using Job approach is lot about practice. Php workers is also about practice usage.

But probably we should leave feature when user can make plain payload to queue backend.

klimov-paul:
One more thing just in last thread: why there is a need of such complex construction?

> Yii::$app->queue->push(SendEmailJob::className(), ['user' => $user]);

Why don't use following:
```php
$job = new SendEmailJob(['user' => $user]);
$job->push();
```
or even:
```php
(new SendEmailJob(['user' => $user]))->push();
```

creocoder:
@klimov-paul Because instantiating Job does not make any sence in place where you put it to queue. Note that it will be serialized and will be used in different place.

klimov-paul:
Fine, still why not following:
```php
SendEmailJob::push(['user' => $user]);
```

creocoder:
@klimov-paul What if you want to use non-default backend? What if you want yo use non default pool?

klimov-paul:
> probably we should leave feature when user can make plain payload to queue backend.

This is from what we should start, then we can thinl about high-level approach.

creocoder:
@klimov-paul Anyway its syntax sugar, Queue component obviously must have push method too...

klimov-paul:
> What if you want to use non-default backend? What if you want yo use non default pool?

You can add extra parameters to `Job::push()`.

creocoder:
@klimov-paul Ok, i like that we start from some constructive points :) I suggest start from realize that plain payload practiceless while it still should be possible. Now about instantiate Jobs or not. Not instantiate Job classes allow to make payload like:

```json
{
    "job" => "app\jobs\SendEmail",
    "data" => // serialized data, note that serializer can be configured
}
```

> You can add extra parameters to Job::push()

Its syntax sugar we can think, do you agree that its obvious that Queue component must have push method too?

klimov-paul:
> instantiating Job does not make any sence in place where you put it to queue. Note that it will be serialized and will be used in different place.

I don't get it... As far as I know in order to actualy serialize an object, it should be intantiated first, should not it? It is also better to be validated first and throw an exception if its parameters are not valid or something.
That is why it seems logical to allow job push itself, like we allow ActiveRecord to save itself.

creocoder:
@klimov-paul Check my above comment about proposed payload format.

klimov-paul:
> Its syntax sugar we can think, do you agree that its obvious that Queue component must have push method too?

For me `Queue::push()` should accept plain string, while all high-level logic about serialize and so on should lay on `Job` level.

creocoder:
@klimov-paul I'm fine with that, good suggestion.

creocoder:
@klimov-paul On other side how worker will get job from particular queue in that case?

creocoder:
@klimov-paul Look at `Queue::pop()` method, it should return prepared Job to perform its operations. So `Queue` class is still high level.

creocoder:
We can do something like `Queue::findAllPending()` or similar. Not that jobs has statuses, can be reserved, failed, etc. I just do not show this yet in suggested approach.

framework/console/controllers/QueueController.php
tom--: i have been using daemons i wrote in php (not yii) for many years. they have been perfectly stable. i also use yii console commands started by cron to process certain queues. both models make sense

klimov-paul:
> On other side how worker will get job from particular queue in that case?

Plain queue:
```php
while ($message = Yii::$app->queue->pop()) {
    // do something
}
```

`Job` level:
```php
while (MyJob::pop()) {
    // do something
}
```
or 
```php
MyJob::listen(); // internal loop with `Job::run()` invocation.
```

framework/console/controllers/QueueController.php
creocoder: @klimov-paul 

> there is no need for predefined controller for the queue processing

Since we talk about high-level approach in addition to low level approach ofcource we need such controller with default implementation. Ninjas who want non-PHP workers may write it even on Erlang if they want. Main idea is we should provide php workers from box for typical user.

creocoder:
@klimov-paul Ok, i updated PR with new approach. Lets than decide

```php
MyJob::push($data, $queue, $delay)
```

VS

```
(new MyJob($data))->push($queue, $delay)
```

I think that 2nd variation is better since i see no drawbacks on creating payloads from instantiated classes.

framework/console/controllers/QueueController.php
klimov-paul: > Main idea is we should provide php workers from box for typical user.

For me such worker should be written as:
```php
MyJob::listen();
```
Then I can place it wherever I want.

klimov-paul:
I still think `Job` should extend `Model` and run `validate()` before been put into a queue.
Also particular Job should define its queue name via method like `queueName()`, so there should not be necessary to pass queue name or other parameters to `Job::push()`.

creocoder:
> I still think Job should extend Model and run validate() before been put into a queue.

Any use case for that? Job is not model and is not form model and its data already validated. I think extends Model is monstrous overcoding and does not make sence at all.

creocoder:
@klimov-paul 

> Also particular Job should define its queue name via method like queueName(), so there should not be necessary to pass queue name or other parameters to Job::push().

Yes, agree, great suggestion!

framework/console/controllers/QueueController.php
creocoder:
```php
MyJob::listen();
```

MyJob listen what? It does not make sence. Queue usually contain **different** jobs (*Job classes). Worker run it step by step depends on queue (pool). Also note that job may fail, worker can create forks, worker can be daemon, etc. So worker technically is very complex component and we need provide it implementation from box.

nineinchnick:
I've made the #2378 PR prosing a queue component interface a long time ago. I'm interested in more generic message queues with optional publish-subscribe capabilities. Do you consider to have such base on top of which a job queue can be implemented?

creocoder:
@nineinchnick I checked your implementation before propose mine. I did not get how this can be practically used or interested without 1) background workers, 2) job classes. My implementation is classic queue abstraction implementation inspired mainly by RoR, Laravel.

creocoder:
@klimov-paul With your proposals now there is trouble. For example realize we have queue:

```
1 | SendEmailJob | default
2 | CreateStripeAccountJob | default
3 | SendEmailJob | default
4 | RemoveOldImagesJob | default
```

Usually worker does not care about job type. This is controlled by pools (queues). This is why your idea about `Job::listen()` or `Job::pop()` does not make sence at all. Worker run jobs from particular or all pools regardless of it type. Now after we make Queue component low level `Queue::pop()` can't return Job and its neccessary to create Job classes from plain messages. So i'm doubt about that a little.

creocoder:
@klimov-paul If only `Job::pop()` will be fabric for concrette Job classes. Like:

```php
$job = Job::pop(); // can return SendEmailJob, CreateStripeAccountJob, etc
```

But in that case we'll have trouble with choose connection. What if you want pop jobs from different connection or/and pool (queue).

creocoder:
@klimov-paul After some thinking i realize that your suggestions complettely break original idea and add several drawbacks without any benefits, moreover its just can't be implemented this way. First of all features planned:

- send raw message to queue
- send anonymous function with serialized data to queue
- send Job to queue

Its impossible to implement with having silly Queue component which can only send raw messages to queue. Also Job instanse should not be serialized from message size point of view.

Examples:

```php
Yii::$app->queue->push(function ($data) {
    Yii::$app->mailer->send(...);
}, ['data' => $data]);

Yii::$app->queue->push(SendEmailJob::className(), ['data' => $data]);
```

It was designed first way for such pushes because raw message push does not make sence from practical reasons and even if its needed it can be implemented through `pushMessage()` method for example. Second with your suggestions its impossible to pop different Job classes and serialized anonymous functions from queue.

> MyJob::listen();

Will not work because queue servers work different way FYI. You do not know what kind of Job will come next. This is why listening concrette Job is just naive. If you suggest put one kind of job per pool it would be serious restrictions and its absoluttely unacceptable. You can push into one pool 2 class based Jobs, 3 anonymous function jobs and 5 plain messages. When you pop from queue you do not know what you'll get to process (check Amazon SQS documentation about non garantee orders for example).

Because all of this i'll rollback my original idea.

creocoder:
@klimov-paul If short i come to conclusion that you can't have low level Queue components and high level architecture around that. Queue should be low/high level from start.

acorncom:
> I checked your implementation before propose mine. I did not get how this can be practically used or interested without 1) background workers, 2) job classes. My implementation is classic queue abstraction implementation inspired mainly by RoR, Laravel.

@creocoder Regarding the above, how will you be handling queues if you don't have background workers?  Seems like that's a core part of all of these queue systems (RoR, Laravel, Resque, etc) ...

framework/queue/Queue.php
```diff
@@ -0,0 +1,37 @@
...
+
+use yii\base\Component;
+
+/**
+ * Queue is the base class for queue classes supporting different queue back-ends.
+ *
+ * @author Alexander Kochetov <creocoder@gmail.com>
+ * @since 2.0.5
+ */
+abstract class Queue extends Component
```
cebe: this looks more like an interface, what are the arguments for using abstract class?

creocoder: It will have common implementation like `\yii\caching\Cache`

mikehaertl:
Guys, are you sure that you want to rush out a quick queue implementation here? The discussion here seems to imply that the authors are not really experts on queues themselves. I've not used many queues myself over the years. But from what I have seen, creating a message queue from scratch is a very difficult task that requires a lot of experience and careful planning.

There's a reason, why there are so many queue implementations out there. There's even a famous implementation for PHP: [phpresque](https://github.com/chrisboulton/php-resque), a port of a Ruby implementation that's based on Redis, but even while being actively maintained it hasen't seen a new release for some years.

Many queues are built on top of Redis. And even the author of redis is working on a new queue system as he found some limitations in the Redis-Queue approach: http://antirez.com/news/88. This article alone show, that things are not so easy and that there's a lot to consider for a reliable queue system.

All I want to say here is: Can we please plan this a bit more carefully? The discussion here makes me nervous, that we end up with a half-baked Queue solution for Yii2 and have to stick with a not so good API for a long time.

nineinchnick:
@mikehaertl I had the same feeling here. I did put a lot of work in my PR. While doing it I investigated Redis and some online services to come up with an universal API. Unfortunately, the discussion died there.

creocoder:
@mikehaertl Why you think it will be done from scratch? As i said before you can look at this as straight port of http://laravel.com/docs/5.1/queues and https://github.com/chrisboulton/php-resque.

@nineinchnick As i said your PR is dead long time ago. Because... i do not know... maybe its not something what people waiting from queue abstraction? Compare your PR with for example http://laravel.com/docs/5.1/queues by features and practice usage.

creocoder:
> The discussion here makes me nervous, that we end up with a half-baked Queue solution for Yii2 and have to stick with a not so good API for a long time.

Seems you misunderstood idea. Everybody realize that develop queue abstraction is very hard task. I just made PR with simple interface with only 2 methods to check core dev reaction and opinion on this. So it was just checking is Laravel/Resque ideas compatible with our fw ideas. As you see from discussion its not. Maybe that even mean that yii 2 queue component will never be developed. I do not know. I cleanly see now that we need much more low level solution. So we do not rush out a quick queue implementation here, just checking approach compatibility, nothing more.

nineinchnick:
@creocoder you're talking about job queues. I'm talking about more generic message queues. I did read Laravel docs BEFORE making my PR and obviously it only supports job queues.

If you have a generic message queue interface it's easy to add a job queue interface on top of it. But you can't use a job queue for messaging. That's my point.

I would really like this discussion to help create a queue interface in the core framework. Somebody should establish some base assumptions and direct the discussion.

creocoder:
@nineinchnick You right that Laravel implementation is highly specialized. You wrong with

> If you have a generic message queue interface it's easy to add a job queue interface on top of it.

mikehaertl:
> I just made PR with simple interface with only 2 methods to check core dev reaction and opinion on this

I understand that. But given the mass of low-level queue implementations, each with slightly different features, I think it's a very demanding task to get this queue interface right. Implementations take different approaches to queueing and thus come with different features and APIs. A generic API must "rule them all" and be feature complete. It's not easy to define such a complete set of features.

A concrecte example: I'm missing features for job control and monitoring. I want to get information about all jobs in a queue, which jobs where recently completed, etc. The information here can vary a lot among different queue implementations. The API should have a generic approach to this.

That's just one example. I'm sure there's much more to miss.

nineinchnick:
@creocoder  why can't I make a job queue on top of a message queue? Resque does that with Redis.

nineinchnick:
@mikehaertl I tried doing that in my PR #2378, supporting 3 completely different backends. And I think it could handle job control as there is peeking into the queue and referencing messages by id. As for monitoring, you can put finished jobs in another queue.

@creocoder I would be really happy to try to build an example of a job queue on my interfaces.

creocoder:
@nineinchnick Ok lets try think. For example we have:

```
\yii\queue\Queue <- abstract class
\yii\queue\DbQueue extends \yii\queue\Queue <- work with generic messages
\yii\queue\RedisQueue extends \yii\queue\Queue <- work with generic messages
\yii\queue\AmazonSqsQueue extends \yii\queue\Queue <- work with generic messages
```

Now you want implement jobs on top of this. Can you show suggestion on at least class level?

Note: For example for jobs you need more fields than you have for DbQueue.
Note: Same situation may happen with any other adapter.

creocoder:
I just want to say that or we support job-like payloads from start or it will lead to separate set of same components.

nineinchnick:
Not necessary, you already discussed that a Job class should use a queue component just like AR uses the db connection component. I'm going to write up an example later.

creocoder:
Ok, after some thinking i realize that mistake was in decision:

```php
MyJob::push($data);
```

VS

```php
(new MyJob($data))->push();
``

Now its obvious that 2nd variation can't be used because when pop we do not know what serializer for data was used. This is proof that special format of payloads should be used. For example Laravel use next format:

```
{
    "job": "simple string, for example class name",
    "data": ..., // some serializer used, for example php serialize(),
}
```

By going that way we can make data serialization method optional and part of concrette Job class. Now how to pop jobs:

```php
// only 
$job = Job::pop();
// and NOT
$job = SendEmailJob::pop();
```

Second call does not make sence at all. When you do pop you do not know what kind of job you will receive. First call result may leat to `$job instance of SendEmailJob` or `$job instance of RemoveOldImagesJob`, so its some sort of fabric.

But there i have some doubts that concrette job classes will inherit method which they can't use at all. Maybe its smart to separate `pop` method to different class?

mongosoft:
Not necessarily the Job object should have a pop method.
For example, extracting and processing the jobs should be able a worker only.

```php
class SendEmail extends ActiveJob
{
    public function queue()
    {
         return 'send_email';
    }

     public function run()
     {
           // do something...
     }
}

class Worker
{
    public function process($tube)
    {
         while (true) {
             $job = Yii::$app->queue->pop($tube);
             $job->run();
         }
    }
}


// usage

$job = EmailJob();
$job->delay = 60;
$job->push(); // or save()
```

The Job may be stored for example in a format:

```php
{
    "type": "job",
    "payload": '...'
}
```

creocoder:
@mongosoft If you read original discussion

```php
Yii::$app->queue->pop($tube);
```

will return message and @klimov-paul idea was Queue is set of low level components. It does not know anything about `jobs`.

mongosoft:
I agree that the base queue component must be able to only push-pop a plain text messages, because job queue it's a particular case of the implementation of the message queue. Implementation of job queue should deal with a component of higher abstraction, ie queue should be used only as a storage of messages which are associated with some tasks. Yes, it will be difficult to implement (and perhaps impossible or senselessly), but we are thinking and proposing ideas.

Why it does not know anything about the ```jobs```? If I run a special workers and he was signed to read messages from the queue - I made it clear that there is a list of jobs. In laravel, for example, there is always some work - therefore it's easy.

tom--:
I think we're talking about APIs and implementations before there's any sort of agreement on requirements. @nineinchnick PR may be "dead" but it has a much more useful discussion of requirements than the above.

Why are we discussing PRs on core anyway? Doesn't this kind of thing work as an extension? I tend to agree with @mikehaertl that doing generic, flexible queues is hard. As such, wouldn't it be better to have one or more [50% solutions](http://blog.ircmaxell.com/2014/10/an-open-letter-to-php-fig.html) prove themselves in the community than to put into core something that *collectively* we don't appear to properly understand?

creocoder:
@tom-- 

> Doesn't this kind of thing work as an extension?

Just for example look at this http://laravel.com/docs/5.1/events#queued-event-listeners . Still think it should be extension?

creocoder:
@tom-- 

> I tend to agree with @mikehaertl that doing generic, flexible queues is hard.

Does that mean that Yii 2 should never have queue components? Its hard and what? Not harder than any other abstraction. Why do you set so pessimistic? This PR only contains **1 interface with 2 methods**.just to listen to different developers opinions and find our way of implementing queue abstraction.

usualdesigner:
Yii 2 should contain queue component definitely!

samdark:
It's nothing wrong with discussing it at least. If the product of discussion is the right thing, it's very welcome to the core at least if form of interfaces.

samdark:
#492

iJackUA:
I want to show another one example ActiveJob from rails http://edgeguides.rubyonrails.org/active_job_basics.html
I feel it like a very good solution.

First of all - we need clearly separate Generic Queue implementation and Jobs/Workers.
Generic Queue should just  push/get payload from some tube.
And ActiveJob is higly specialized thing made ON TOP of Generic Queue.
Rails workflow is very explicit and does not say anything about Queue (because when you use Job, you should not think about Queue etc.) just call `MyJob.perform` for immediate exectuion or `MyJob.perform_later` for delayed.
In PHP it could look like
`MakeLongRunTask::job(['param1'=>'value1'])->perform()` or `MakeLongRunTask::job(['param1'=>'value1'])->performLater()`

I have prototyped it in Yii2 like that

```
abstract class ActiveJob extends Object
{
    /**
     * Main job code contained here
     * Each job should implement it
     *
     * @return mixed
     */
    abstract function perform();
    /**
     * TBD
     * Should delay job execution in queue
     *
     * @return mixed
     */
    public function performLater()
    {
       // ADD TO QUEUE HERE 
        throw new Exception('Delayed execution is not implemented yed');
    }
    /**
     * Initialisation helper, to call a job in semi-static way
     * Instantiates job object and passes array of configuration in it
     *
     * @return static
     */
    public static function job($config)
    {
        $class = get_called_class();
        return new $class($config);
    }
}
```

When ActiJob adds itself (all public attributes as a payload) to the queue it automatically adds some `_mate_classname` property to serialized payload.
ActiveJob can have a lot of different configs - like queue/tube name, timeouts, serializers etc. 
And due with two init ways  via explicit `$job = new AbcJob(); ...` or via shortcut ` AbcJob::job()->perform()` you can configure anything before execution or run it very quickly with defaults.

Then via console command (special for ActiveJobs execution) payload is being taken from tube, unserialized, Object restored via `_mate_classname` and `perform` called on than restored object.

So first of all  - I propose to think of Queue and Jobs implementation like two separated things.

petrabarus:
Really eager to see this gets implemented.
I already implemented my version of queue. You might want to check this https://github.com/urbanindo/yii2-queue
It's already in production using AWS SQS.

creocoder:
Guys, just want to inform than new totally reworked concept is near ready and will be pushed in few days.

creocoder:
Moved to https://github.com/yiisoft/yii2-queue/pull/1

PR #8880, opened 2015-06-24
toArray can be string

cornernote:
update docs to reflect the available type


cebe:
seems to be valid, I am wondering whether there is a reasonable use case for this?

cornernote:
scrutinizer-ci was giving issues because this was missing.  It saw that we allowed string|array, and it said `ArrayHelper::toArray()` didn't allow a string.

samdark:
Probably not. But I see no harm in it. The change makes sense because it describes current behavior.

PR #8883, opened 2015-06-24
Fixed some words in Indonesian language

muhammadcahya:
Fixed some words to make more humane in indonesian language
Consistent `invalid` words to `tidak valid`


samdark:
Thanks!

Issue #8901, opened 2015-06-25
AssetManager can not get the correct published url or path when only partial published


hightman:
For example, my AssetBundle specified `@bower` as `$sourcePath`, after publishing only the realpath of `@bower` is saved to `AssetManager::$_published`.

When I try to pass "@bower/admin-lte/dist" as parameter into `Yii::$app->assetManager->getPublishedUrl()`, it can not return the correct result. The same issue also occurred in `AssetManager::getPublishedPath()`.

Faryshta:
1) there is already and adminlte asset, you can google it

2) can you show us your composer.json?

hightman:
Please look at the problem, `Admin-LTE` is only an example.

klimov-paul:
Rejected: this is missusage of the AssetBundle publish mechanism.
If there are several bundles which uses files under same root directory, they should specify `sourcePath` to be this root directory, while particular JS or CSS files should be specified as related path.

This how it is done at `yii2-botstrap`:
https://github.com/yiisoft/yii2-bootstrap/blob/master/BootstrapAsset.php#L20-L23
https://github.com/yiisoft/yii2-bootstrap/blob/master/BootstrapPluginAsset.php#L20-L23
https://github.com/yiisoft/yii2-bootstrap/blob/master/BootstrapThemeAsset.php#L20-L23

Overcomplicating the code is not justified here.

PR #8902, opened 2015-06-25
Bug #8901: Fixed AssetManager to get correct published path or URL

hightman:
see issue #8901


hightman:
Hi, why not consider to accept fix this issue, is it cause other problems? @samdark @qiangxue

klimov-paul:
See https://github.com/yiisoft/yii2/issues/8901#issuecomment-243693006

PR #8903, opened 2015-06-25
\yii\db\pgsql\QueryBuilder add createIndex() for the creating specific indexes

lav45:
In the current version it is not possible to create a btree, rtree, hash and other indices since the method createIndex() is limited to the creation of unique indexes.

### DIFF
--- a/framework/db/pgsql/QueryBuilder.php
+++ b/framework/db/pgsql/QueryBuilder.php
@@ -17,6 +17,12 @@ use yii\base\InvalidParamException;
  */
 class QueryBuilder extends \yii\db\QueryBuilder
 {
+    const INDEX_UNIQUE = 'unique';
+    const INDEX_B_TREE = 'btree';
+    const INDEX_R_TREE = 'rtree';
+    const INDEX_HASH = 'hash';
+    const INDEX_GIST = 'gist';
+
     /**
      * @var array mapping from abstract column types (keys) to physical column types (values).
      */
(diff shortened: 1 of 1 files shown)


samdark:
Looks OK to me. I'm going to merge it. @yiisoft/core-developers any comments?

framework/db/pgsql/QueryBuilder.php
```diff
@@ -64,6 +70,27 @@ class QueryBuilder extends \yii\db\QueryBuilder
...
     ];
 
+    /**
+     * Builds a SQL statement for creating a new index.
+     * @param string $name the name of the index. The name will be properly quoted by the method.
+     * @param string $table the table that the new index will be created for. The table name will be properly quoted by the method.
+     * @param string|array $columns the column(s) that should be included in the index. If there are multiple columns,
+     * separate them with commas or use an array to represent them. Each column name will be properly quoted
+     * by the method, unless a parenthesis is found in the name.
+     * @param boolean|string $index whether to add UNIQUE constraint or create another index.
```
cebe: in the parent class implementation the name of this param is `$unique`, why did you change it to `$index`?

cebe: Oh, now I got it, this can be a string now to indicate the index type. Better rename it to indexType and add explaination how this parameter can be used.

framework/db/pgsql/QueryBuilder.php
```diff
@@ -17,6 +17,12 @@
  */
 class QueryBuilder extends \yii\db\QueryBuilder
 {
+    const INDEX_UNIQUE = 'unique';
+    const INDEX_B_TREE = 'btree';
+    const INDEX_R_TREE = 'rtree';
+    const INDEX_HASH = 'hash';
+    const INDEX_GIST = 'gist';
```
cebe: these constants should have a phpdoc description. Not sure if they are needed at all, we could allso passing raw string to `createIndex()`.

LAV45:
@cebe Will any more wishes?

LAV45:
Thanks! ))

cebe:
Thank you ;)

Issue #7627, opened 2015-03-08
ActiveForm inputs with custom id are not validated


SDKiller:
If you pass your custom `id` like this:

```php
$form->field($model, 'comment')->textarea(['id' => 'myCustomId'])
```

I suppose the reason is in `\yii\widgets\ActiveField::getClientOptions`:

```php
        $inputID = Html::getInputId($this->model, $this->attribute);
        $options['id'] = $inputID;
```

It does not care if `id` was passed to field options.

I cannot use autogenerated `id` as there are several forms of the same model on the page - `ids` will not be unique.

SDKiller:
UPD

I see this in docs
```
* If you set a custom `id` for the container element, you may need to adjust the [[$selectors]] accordingly.
```

But what is the sense to autogenerate `id` for input if this `id` does not exist in `DOM` as custom id was set explicitly?

cebe:
is it clear now, or is the question still open?

SDKiller:
As I could understand - `id` is used internally and not always represents `id` of html element `.
While it's confusing and  inconvenient to specify in addition to `id` of html element 

```php
$form->field($model, 'comment', ['selectors' => ['input' => '#myCustomId']])
            ->textarea(['id' => 'myCustomId']);?>
```

cebe:
as far as I see, yii should be able to automatically update the selector in most cases. Need to check that.

SDKiller:
In my example `options` passed to `textarea()` have no chance to be fetched by `getClientOptions()` as they are merged with `inputOptions` only for `$this->parts['{input}']` rendering.

They are not merged into `inputOptions` property of `ActiveField` object itself.

SDKiller:
I mean in `\yii\widgets\ActiveField::getClientOptions` it could be done like this

```php
$inputID = isset($this->inputOptions['id']) ?  $this->inputOptions['id'] : Html::getInputId($this->model, $this->attribute);
```

but it will work only if you create field in this way:

```php
$form->field($model, 'comment', ['inputOptions' => ['id' => 'myCustomId']])->textarea();?>
```

dynasource:
I can confirm this issue and the solution of SDKiller to be working for my scenario. Can it be implemented? thnx

cebe:
seems there is no good way to solve it.

dynasource:
I can confirm this issue and the solution provided by SDKiller to be working for me.
Two lines should be replaced:

[...]

PR #8912, opened 2015-06-26
solves #7627 https://github.com/yiisoft/yii2/issues/7627

dynasource:
enables Ajax validation when customizating the ID for an input

solves #7627


dynasource:
@cebe: can you pass this one through plz? Thanks!

cebe:
found a way which may cover more cases: #9294 have not tested it yet.

cebe:
merged #9294, thanks for your proposal, @dynasource!

Issue #8933, opened 2015-06-29
Fatal errors show blank page in HHVM Yii2


dieend:
Similar issue with https://github.com/laravel/framework/issues/8744.

I have make sure to set up hhvm options correctly from https://github.com/facebook/hhvm/issues/4818
```
 hhvm.server.implicit_flush = true
 hhvm.error_handling.call_user_handler_on_fatals = true
```

samdark:
@dieend can you check if changes in #8935 are enough?

dieend:
I don't think adding the error code is enough, I've tried it previously but let me check again later, because I wasn't adding the getName() parts.

Typed in mobile, sorry for any typo.

dieend:
@samdark I confirms those changes in #8935 doesn't enough. I'm able to show error exception by modifying https://github.com/yiisoft/yii2/blob/master/framework/web/ErrorHandler.php#L217 to 
```
    public function renderFile($_file_, $_params_)
    {
        $_params_['handler'] = $this;
        if ($this->exception instanceof ErrorException || !Yii::$app->has('view')) {
//            ob_start();
//            ob_implicit_flush(false);
            extract($_params_, EXTR_OVERWRITE);
            return require(Yii::getAlias($_file_));
        } else {
            return Yii::$app->getView()->renderFile($_file_, $_params_, $this);
        }
    }
```
I don't know why but `ob_start()` makes empty result for all output after that. `ob_get_clean()` returns empty string.
Looks like after throwing error, output buffer doesn't works in hhvm. I don't know whether it's bug/php-incompability in hhvm, or there is another thing that's missing.

PR #8935, opened 2015-06-29
Added HHVM fatals to error handler

samdark:
#8933

### DIFF
--- a/framework/base/ErrorException.php
+++ b/framework/base/ErrorException.php
@@ -17,6 +17,8 @@ use Yii;
  */
 class ErrorException extends \ErrorException
 {
+    const E_HHVM_FATAL_ERROR = 16777217;
+
     /**
      * Constructs the exception.
      * @link http://php.net/manual/en/errorexception.construct.php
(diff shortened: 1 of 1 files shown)


framework/base/ErrorException.php
```diff
@@ -17,6 +17,8 @@
  */
 class ErrorException extends \ErrorException
 {
+    const E_HHVM_FATAL_ERROR = 16777217;
```
klimov-paul: Is not such constant defined at HHVM environment already?

klimov-paul: Also, is this the only one extra error added at HHVM?

cebe: where is this constant from, it does not seem to be documented: http://docs.hhvm.com/manual/en/errorfunc.constants.php

dieend: 16777217 comes from here: https://github.com/facebook/hhvm/blob/master/hphp/runtime/base/runtime-error.h#L62 More info is in this issue https://github.com/facebook/hhvm/issues/4818 in comment from iansltx.

petrabarus:
As I reported in #8960, this fix does show the error. But it doesn't show the stack trace.

PR #8946, opened 2015-06-29
Spelling and sc fixes

mohorev:


framework/base/ArrayableTrait.php
```diff
@@ -141,7 +141,7 @@ protected function resolveFields(array $fields, array $expand)
         $result = [];
 
         foreach ($this->fields() as $field => $definition) {
-            if (is_integer($field)) {
```
samdark: It's OK to use is_integer: http://php.net/manual/en/function.is-integer.php

mongosoft: Yes, i know it. There is no preferred name among using alias or main function name. 
http://php.net/manual/en/aliases.php

But, for example:
using ```is_int```:
https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseArrayHelper.php#L77
and using ```is_integer``` in the same file:
https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseArrayHelper.php#L122


I just thought it would be better to use something one.

samdark: OK. Makes sense.

samdark:
Thanks!

PR #8956, opened 2015-06-30
clear code, yii\base\Event is not used

lav45:


cebe:
it is used: https://github.com/yiisoft/yii2/blob/master/framework/db/BaseActiveRecord.php#L46

LAV45:
PhpStorm recognizes the Event as plain text.

cebe:
looka like a bug in your IDE

PR #8976, opened 2015-07-02
Hhvm fatal error

### LINKED ISSUE #8933: Fatal errors show blank page in HHVM Yii2

dieend:
fixes #8935 
closes #8933 

@samdark I'm able to show hhvm error this way. I adapted this mostly from https://github.com/symfony/symfony/pull/14999

### DIFF
--- a/framework/base/ErrorException.php
+++ b/framework/base/ErrorException.php
@@ -17,6 +17,8 @@ use Yii;
  */
 class ErrorException extends \ErrorException
 {
+    const E_HHVM_FATAL_ERROR = 16777217;
+
     /**
      * Constructs the exception.
      * @link http://php.net/manual/en/errorexception.construct.php
@@ -65,7 +67,7 @@ class ErrorException extends \ErrorException
      */
     public static function isFatalError($error)
     {
-        return isset($error['type']) && in_array($error['type'], [E_ERROR, E_PARSE, E_CORE_ERROR, E_CORE_WARNING, E_COMPILE_ERROR, E_COMPILE_WARNING]);
+        return isset($error['type']) && in_array($error['type'], [E_ERROR, E_PARSE, E_CORE_ERROR, E_CORE_WARNING, E_COMPILE_ERROR, E_COMPILE_WARNING, self::E_HHVM_FATAL_ERROR]);
     }
 
     /**
@@ -89,6 +91,7 @@ class ErrorException extends \ErrorException
             E_USER_NOTICE => 'PHP User Notice',
             E_USER_WARNING => 'PHP User Warning',
             E_WARNING => 'PHP Warning',
+            self::E_HHVM_FATAL_ERROR => 'HHVM Fatal Error',
         ];
 
         return isset($names[$this->getCode()]) ? $names[$this->getCode()] : 'Error';
--- a/framework/base/ErrorException.php
+++ b/framework/base/ErrorException.php
@@ -17,7 +17,13 @@ use Yii;
  */
 class ErrorException extends \ErrorException
 {
-    const E_HHVM_FATAL_ERROR = 16777217;
+
+    /**
+     * Zend runtime won't call the error handler on fatals, HHVM will, with an error code of 16777217
+     * We will handle fatal error a bit different on HHVM.
+     * @see https://github.com/facebook/hhvm/blob/master/hphp/runtime/base/runtime-error.h#L62
+     */
+    const E_HHVM_FATAL_ERROR = 16777217; // E_ERROR | (1 << 24)
 
     /**
      * Constructs the exception.
--- a/framework/base/ErrorHandler.php
+++ b/framework/base/ErrorHandler.php
@@ -100,6 +100,10 @@ abstract class ErrorHandler extends Component
             }
             $this->renderException($exception);
             if (!YII_ENV_TEST) {
+                \Yii::getLogger()->flush(true);
+                if (defined('HHVM_VERSION')) {
+                    flush();
+                }
                 exit(1);
             }
         } catch (\Exception $e) {
@@ -119,6 +123,9 @@ abstract class ErrorHandler extends Component
             }
             $msg .= "\n\$_SERVER = " . VarDumper::export($_SERVER);
             error_log($msg);
+            if (defined('HHVM_VERSION')) {
+                flush();
+            }
             exit(1);
         }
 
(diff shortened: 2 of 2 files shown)


klimov-paul:
This PR breaks the BC: it changes public methods API.

framework/web/ErrorHandler.php
```diff
@@ -219,7 +219,10 @@ public function renderFile($_file_, $_params_)
         $_params_['handler'] = $this;
         if ($this->exception instanceof ErrorException || !Yii::$app->has('view')) {
             ob_start();
-            ob_implicit_flush(false);
+            // somehow implicit flush doesn't work for hhvm when handling error
+            if (!defined('HHVM_VERSION')) {
+                ob_implicit_flush(false);
+            }
```
klimov-paul: I assume only the aboce block is necessary to actually solve the issue.

dieend: Well, it solves the issue, but it doesn't have any stacktraces. The output with only above block give me this (more or less):
```
HHVM Fatal Error – yii\base\ErrorException

syntax error, unexpected '}'
2. yii\base\ErrorHandler::handleFatalError()

// context variable
$_SERVER = [
...
```

But that output only shows up after commenting out https://github.com/yiisoft/yii2/blob/master/framework/base/View.php#L323 and https://github.com/yiisoft/yii2/blob/master/framework/base/View.php#L484

Meanwhile the whole changes give me this output (more or less):
```
HHVM Fatal Error – yii\base\ErrorException

syntax error, unexpected '}'
1. in /opt/application/apps/desktop/controllers/HomeController.php at line 28 {
    public $layout = 'basic';
 
    /**
     * Display the index page.
     */
    public function actionIndex()
    {
        return $this->render('index')
    }
}
2. in /opt/application/vendor/yiisoft/yii2/BaseYii.php – include(...) at line 288
3. yii\BaseYii::autoload('desktop\controllers\HomeControll...')
4. in /opt/application/vendor/yiisoft/yii2/base/Module.php – class_exists('desktop\controllers\HomeControll...') at line 563
5. in /opt/application/vendor/yiisoft/yii2/base/Module.php – yii\base\Module::createControllerByID('home') at line 521
6. in /opt/application/vendor/yiisoft/yii2/base/Module.php – yii\base\Module::createController('index') at line 449
7. in /opt/application/vendor/yiisoft/yii2/web/Application.php – yii\base\Module::runAction('home/index', []) at line 84
8. in /opt/application/vendor/yiisoft/yii2/base/Application.php – yii\web\Application::handleRequest(yii\web\Request) at line 375
9. in /opt/application/apps/desktop/web/index.php – yii\base\Application::run()

// context variable
$_SERVER = [
...
```

dieend:
Which public methods API that I changed? I'm willing to find workaround for that.

samdark:
@dieend every change of public or protected method signature results in potential breaks in user classes which extend from the framework class and override a method.

framework/web/ErrorHandler.php
dieend: @klimov-paul apparently you are right. I've updated the pull request.

dieend:
@samdark @klimov-paul apparently the issue is related to output buffer doesn't get flushed when `exit` being called in hhvm. The relevant issue in hhvm is https://github.com/facebook/hhvm/issues/4055. Does this PR looks good now?

dieend:
Ah, btw this PR only fix error not showing. 

But I can't get stacktrace for fatal error (other error works ok), without changing public api for `public function handleError($code, $message, $file, $line)`. The only workaround I can suggest is by setting different error handler for hhvm. something like:
```
if (defined('HHVM_VERSION')) {
    set_error_handler([$this, 'handleHhvmError']);
} else {
    set_error_handler([$this, 'handleError']);
}
```
At least showing error without backtraces is better than having no error.

samdark:
Separate handler for HHVM sounds OK actually.

dieend:
Ok, added hhvm error handler in the latest commit. I am using a new private variable called `$this->_hhvmException`. But I wonder should I just directly uses `$this->exception` instead?

framework/base/ErrorException.php
```diff
@@ -17,6 +17,8 @@
  */
 class ErrorException extends \ErrorException
 {
+    const E_HHVM_FATAL_ERROR = 16777217;
```
cebe: need a comment that explains where this comes from. should contain info like given in https://github.com/yiisoft/yii2/pull/8935#discussion_r33483774

framework/base/ErrorHandler.php
```diff
@@ -45,6 +45,10 @@
      */
     private $_memoryReserve;
 
+    /**
+     * @var \Exception from hhvm error that store backtrace
```
samdark: hhvm → HHVM

framework/base/ErrorHandler.php
```diff
@@ -119,11 +131,44 @@ public function handleException($exception)
...
+            }
             exit(1);
         }
 
         $this->exception = null;
     }
+    /**
+     * Handles HHVM execution errors such as warnings and notices.
+     *
+     * This method is used as a Hhvm error handler. It will store exception that will
```
samdark: Hhvm → HHVM

framework/base/ErrorHandler.php
```diff
@@ -119,11 +131,44 @@ public function handleException($exception)
...
+     * be used in fatal error
```
samdark: in fatal error handler?

dieend:
@cebe @samdark Addressed reviews.

framework/base/ErrorHandler.php
```diff
@@ -100,6 +108,10 @@ public function handleException($exception)
             }
             $this->renderException($exception);
             if (!YII_ENV_TEST) {
+                \Yii::getLogger()->flush(true);
```
samdark: It affects non HHVM as well.

samdark: Overall flushing log makes sense on exception. Any reason why it's done for `if (!YII_ENV_TEST) {`?

dieend: Yes it's affect non HHVM, but it's make sense (I follows [this line](https://github.com/yiisoft/yii2/pull/8976/files#diff-f000680354101249e9f60e71e6278e3bL193) patterns where it's calling flush before exit).
It's done for `if (!YII_ENV_TEST) {` because the exit only called inside that branch.

samdark:
Merged with some minor adjustments https://github.com/yiisoft/yii2/commit/8ced2ba96a6ddf9bd4db9198850371e7b546162e

PR #8983, opened 2015-07-02
only truncate the original log file for rotation

matthewyang:
other rotation log files will be overwrite, only the original log file need to truncate after copy.

### DIFF
--- a/framework/log/FileTarget.php
+++ b/framework/log/FileTarget.php
@@ -144,7 +144,7 @@ class FileTarget extends Target
                 } else {
                     if ($this->rotateByCopy) {
                         @copy($rotateFile, $file . '.' . ($i + 1));
-                        if ($fp = @fopen($rotateFile, 'a')) {
+                        if ($i === 0 && $fp = @fopen($rotateFile, 'a')) {
                             @ftruncate($fp, 0);
                             @fclose($fp);
                         }


samdark:
Can you describe the issue it's fixing? How to reproduce it?

matthewyang:
It avoid the unnecessary truncate when do log file rotation.
Existing rotation code work like this, let's say the max log files is 5:
1. delete app.log.5
2. copy app.log.4 to app.log.5, truncate app.log.4
3. copy app.log.3 to app.log.4, truncate app.log.3
4. copy app.log.2 to app.log.3, truncate app.log.2
5. copy app.log.1 to app.log.2, truncate app.log.1
6. copy app.log to app.log.1, truncate app.log

After this fix, it will only do truncate for the 6th step, only truncate the original log file:
1. delete app.log.5
2. copy app.log.4 to app.log.5
3. copy app.log.3 to app.log.4
4. copy app.log.2 to app.log.3
5. copy app.log.1 to app.log.2
6. copy app.log to app.log.1, truncate app.log

developeruz:
It seems we already have [tests for that](https://github.com/yiisoft/yii2/blob/75dbdb008363b90d4ffc8f502022763aedd6998a/tests/framework/log/FileTargetTest.php#L102). It doesn't make sense to truncate rotated files because they will be [overwritten](https://github.com/yiisoft/yii2/blob/75dbdb008363b90d4ffc8f502022763aedd6998a/framework/log/FileTarget.php#L146) anyway. Meanwhile, we need to truncate the original log file in case that some other process has a handle to this file and is writing something.

developeruz:
@samdark, I can't edit files in this PR, should I create a new PR to refactor current rotateFiles() method to make codeclimate happy? This PR doesn't make the method worse. I can make PR with a refactored code after this PR was merged.

samdark:
@developeruz yes. Please do.

developeruz:
Replaced by #15173

PR #8988, opened 2015-07-02
[Feature request] Fixture data classes

a-kolomensky:
The problem:

transaction.php fixture file

``` php
<?php

return [
    [
        'id' => 1,
        'currency_id' => 1,
        'type_id' => 2,
        'state_enum' => 'success',
    ],
]
```

What is id 1 in currency field? id2? In project that contains many relations and many tests that hardcored ids make trouble work with fixtures. And heavy troubles when need refactor tests. In refactoring I will replace all _sucess_ strings and remember what means 1 and 2.

Look at this TransactionFixtureData.php

``` php
<?php

class TransactionFixtureData extends BaseActiveFixtureData
{
    public function getData()
    {
        $rubCurrency = Currency::findByName(Currency::RUB);
        $refundType = TransactionType::findByName(TransactionType::REFUND);
        $successState = TransactionState::findByName(TransactionState::IS_SUCCESS);

        return [
            [
                'id' => 1,
                'currency_id' => $rubCurrency->id,
                'type_id' => $refundType->id,
                'state_enum' => $successState->name,
            ]
        ];
    }
}
```

Problems:
- It's class, not static file
- Class makes requests to database (need right depends)
- Impossible use templates

Benefit: it's readable.
Yii communities need it?

### DIFF
--- a/framework/test/BaseActiveFixture.php
+++ b/framework/test/BaseActiveFixture.php
@@ -104,7 +104,7 @@ abstract class BaseActiveFixture extends DbFixture implements \IteratorAggregate
 	    if ($this->dataClass !== null) {
 		    $activeFixtureData = Yii::createObject($this->dataClass);
 		    if (!is_subclass_of($activeFixtureData, BaseActiveFixtureData::className())) {
-			    throw new InvalidConfigException("{$this->dataClass} is not instance of ActiveFixtureData");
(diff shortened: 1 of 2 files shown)


klimov-paul:
Feature makes no sense.
The obvious solution for the described usecase will be creating own fixture class, overriding its `getData()` method. The code will be the same without extra base class involved.

```php
class TransactionFixture extends \yii\test\ActiveFixture
{
    protected function getData()
    {
       //return array;
    }
}
```

PR #8990, opened 2015-07-03
IDE type hinting fix

briedis:
IDE does not understand `object` type, and highlights an error.

Error

With mixed type it works.


cebe:
looks more like a bug of the IDE...

samdark:
Indeed, it's IDE bug: http://www.phpdoc.org/docs/latest/guides/types.html

briedis:
Huh, alrighty then, sorry for this.

samdark:
Report it to JetBrains, they have a tracker here: http://youtrack.jetbrains.net/issues/WI

briedis:
Yup, that's what I did.. https://youtrack.jetbrains.com/issue/WI-27921

PR #8996, opened 2015-07-03
Update FileValidator.php

### LINKED ISSUE #8995: Add ability to NOT limit number of files that can be uploaded in FileValidator
Currently, if I want to upload multiple files I HAVE to set the maxFiles value to be >1 or the validator will assume that only a single file can be uploaded. But what if I do not want to limit amount of files that can be uploaded? Not setting maxFiles or setting it to 0 will make FileValidator assume that only a single file can be uploaded:
```
    {
        if ($this->maxFiles > 1) {
            // ...
        } else {
            $result = $this->validateValue($model->$attribute);
            if (!empty($result)) {
                $this->addError($model, $attribute, $result[0], $result[1]);
            }
        }
    }
```
To work around this I have to set a very big maxFiles value, but it would be better to let the user set it to 0 (telling the validator to expect multiple files upload with unlimited amount of files) and change the validator code accordingly.

P.S. Despite of this statement https://github.com/yiisoft/yii2/issues/4847#issuecomment-53720338 setting maxFiles to 0 is currently neither documented nor implemented in the validator code.

PowerGamer1:
Implementation of https://github.com/yiisoft/yii2/issues/8995

Closes #8995

### DIFF
--- a/framework/validators/FileValidator.php
+++ b/framework/validators/FileValidator.php
@@ -63,7 +63,8 @@ class FileValidator extends Validator
     /**
      * @var integer the maximum file count the given attribute can hold.
      * It defaults to 1, meaning single file upload. By defining a higher number,
-     * multiple uploads become possible.
+     * multiple uploads become possible. Setting it to 0 means there is no limit on
+     * the amount of files that can be uploaded.
      * @see tooMany for the customized message when too many files are uploaded.
      */
     public $maxFiles = 1;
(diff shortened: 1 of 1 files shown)


akaNightmare:
what about ```max_file_uploads``` in php.ini?

PowerGamer1:
> what about  max_file_uploads  in php.ini?

Nothing you can do about it, other than mentioning in the comment for `maxFiles` that the amount of files that can be uploaded is still a subject to `max_file_uploads` PHP setting.

SilverFire:
Merged, thank you!
Changelog, unit test in https://github.com/yiisoft/yii2/commit/2be8a716b1d3284e645efbfb3f711e268e30a64f

PR #9000, opened 2015-07-04
RU edited for tutorial shared hosting [skip ci]

HawkingEtha:


samdark:
Thanks!

HawkingEtha:
My pleasure, I'll continue to observe russian invariants and will do some tryouts to translate by myself. The essential problem still goes on - glossary ignored in many translation docs, in instance, helpers docs. In my opinion it should be reinvented on a different level (linguistic explanations and invariants to each tearm according to usage or else communicating properrties).

PR #9011, opened 2015-07-05
Allow an input tag to have a customisable type

TriAnMan:
The jquery.inputmask allows to use it with `<input type='tel' />` see https://github.com/RobinHerbots/jquery.inputmask#allowed-html-elements


samdark:
Seems valid. Can you add changelog entry?

TriAnMan:
@samdark done.

samdark:
Thanks!

Issue #8708, opened 2015-06-08
Correction for multiple values in filters with select2 jquery


ebisbe:
I've found that the filter is not working for multiple values with a select. I'm currently using the kartik/gridview that has implemented select2 jquery plugin. 

The issue is with the file yii.gridView.js. The way it handles the name:value inputs make that selects with ending _[]_ for the name like _MessageSearch[language][]=fi-FI&MessageSearch[language][]=sv-SE_ are not working. 
I'm not sure if this issue corresponds to here but I think that the modification doesn't affect other behaviors so it's better here.
My change is :
```javascript
var cont = 0;
$.each($(settings.filterSelector).serializeArray(), function () {
    if(this.name.indexOf('[]') !== false) {
        this.name = this.name.replace('[]', '[' + cont++ + ']');
    }
    data[this.name] = this.value;
});

$.each(yii.getQueryParams(settings.filterUrl), function (name, value) {
    if(name.indexOf('[]') !== false) {
        name = name.replace('[]', '[' + cont++ + ']');
    }
    if (data[name] === undefined) {
        data[name] = value;
    }
});
```
from :
```javascript
$.each($(settings.filterSelector).serializeArray(), function () {
    data[this.name] = this.value;
});

$.each(yii.getQueryParams(settings.filterUrl), function (name, value) {
    if (data[name] === undefined) {
        data[name] = value;
    }
});
```

ebisbe:
Somebody fix it. 

Thanks!

samdark:
@ebisbe is there an issue that's reproduceable on Yii w/o kartik/gridview?

ebisbe:
Yes I have the issue in my project. I can set up something for share. How can I do it?

samdark:
Take basic application template, use Yii's own widgets with minimum amount of customizations. Then zip it or put it into github repo.

ebisbe:
@samdark You can checkout this repo https://github.com/ebisbe/yii2-issue-8708 . 
Install composer and run migrations. 
Then you'll need to go to /index.php?r=message%2Findex. In the language filter column you'll see that you can choose one language but whenever you try to choose another one ( after reloading page ) there is always one.

samdark:
@ebisbe I mean w/o kartik extension, not with them so it's clear it's Yii problem and not kartik's.

ebisbe:
The problem is in how the post is made and that is made by Yii2. I update without Kartik GridView

ebisbe:
[...]

PR #9016, opened 2015-07-06
#8708 Fix multiselect error for table

ebisbe:
[ci skip]


ebisbe:
First contribution, hope is all fine.

framework/assets/yii.gridView.js
```diff
@@ -85,25 +85,40 @@
...
-            $.each($(settings.filterSelector).serializeArray(), function () {
-                data[this.name] = this.value;
-            });
+            var dataForm = [];
 
+            // get the query params from the request url
             $.each(yii.getQueryParams(settings.filterUrl), function (name, value) {
-                if (data[name] === undefined) {
-                    data[name] = value;
+                data[name.replace(/\[[0-9]+\]/i, '[]')] = value;
```
samdark: That's very hacky...

ebisbe: Yes I know that. The problem is that we cannot be sure that the values from `name` will have numbers or not. For example the paginaton adds them automatically. I think that we don't have to modify that part because we don't have any problem with the pagination. So we remove all numbers to make the comparison. 
I don't like the solution either but I haven't think of a better one

nineinchnick: Maybe UrlManager::createUrl() should use a custom implementation of http_build_query that doesn't add numeric indices?

framework/assets/yii.gridView.js
```diff
@@ -85,25 +85,40 @@
...
+            });
+
+            $.each($(settings.filterSelector).serializeArray(), function () {
+                if (data[this.name] !== undefined) {
+                    //unset all the values that appear on the url to avoid setting them twice.
+                    delete data[this.name];
+                }
+                if (this.value !== '') { // only add to the url values that are not empty strings
+                    dataForm.push([this.name, this.value]);
```
nineinchnick: I'd push objects instead of arrays:
~~~js
dataForm.push({name: this.name, value: this.value});
~~~

It's better to later reference name and value by keys instead of numeric indices.

framework/assets/yii.gridView.js
```diff
@@ -155,6 +170,17 @@
         data: function () {
             var id = $(this).attr('id');
             return gridData[id];
+        },
+
+        setValue: function (data, name, value) {
```
nineinchnick: Where is it used?

framework/helpers/BaseHtml.php
```diff
@@ -828,10 +828,6 @@ public static function listBox($name, $selection = null, $items = [], $options =
         }
         $options['name'] = $name;
         if (isset($options['unselect'])) {
-            // add a hidden field so that if the list box has no option being selected, it still submits a value
```
nineinchnick: Why this is removed?

framework/helpers/BaseHtml.php
```diff
@@ -828,10 +828,6 @@ public static function listBox($name, $selection = null, $items = [], $options =
...
-            if (!empty($name) && substr_compare($name, '[]', -2, 2) === 0) {
```
nineinchnick: I'd keep this change but it might break BC. Maybe instead of removing this we should right trim brackets from query params when comparing them in _applyFilter_?

framework/assets/yii.gridView.js
ebisbe: It could be a good solution but I think it's better to have that hacky code in the JS than implement a custom core function from the PHP. We could remove them in the response of the http_build_query() so it will be easier to maintain.

framework/web/UrlManager.php
```diff
@@ -488,4 +488,16 @@ public function setHostInfo($value)
...
+
+    /**
+     * Remove numeric indexes from the response
+     * @param $params
+     *
+     * @return mixed|string
+     */
+    protected function custom_http_build_query($params) {
+        $request = http_build_query($params);
+        $request = preg_replace('#%5B[0-9]%5D#', '%5B%5D', $request);
```
nineinchnick: What about indexes with more than one digit?

framework/web/UrlRule.php
```diff
@@ -352,4 +352,16 @@ protected function getParamRules()
...
         return $this->_paramRules;
     }
+
+    /**
+     * Remove numeric indexes from the response
+     * @param $params
+     *
+     * @return mixed|string
+     */
+    protected function custom_http_build_query($params) {
```
nineinchnick: Avoid code duplicates. Make this public and static, possibly move to some helper class.

framework/web/UrlManager.php
ebisbe: @nineinchnick I forgot the + from the [0-9]+. What do you mean with break the param values? I'll make unit test if I understand what could be broken

nineinchnick: For example, `http_build_query(['one' => 'two[1]'])` returns _one=two%5B1%5D_ where the value would be incorrectly modified. That's why I opt for a custom implementation of `http_build_query` where this can be easily avoided.

ebisbe: Ok. Got it. I've moved the function to yii\helpers\Url is fine there?

nineinchnick: Sounds good. Also I hope some core devs will comment on this.

ebisbe:
@nineinchnick I've updated the code. And made a test. I'm not sure if more are needed.

framework/web/UrlManager.php
```diff
@@ -488,4 +489,5 @@ public function setHostInfo($value)
     {
         $this->_hostInfo = rtrim($value, '/');
     }
+
```
nineinchnick: Remove this empty line.

framework/helpers/Url.php
```diff
@@ -15,4 +15,31 @@
  */
 class Url extends BaseUrl
 {
+    /**
+     * @param        $params  Array of values to convert into http query
```
nineinchnick: Is it always an array? Maybe add a type hint here.

framework/helpers/Url.php
```diff
@@ -15,4 +15,31 @@
...
+     * @param string $keyfather Inheritance of keys
+     * @param string $response
+     * @param bool   $ltrim
+     *
+     * @return string
+     */
+    public static function custom_http_build_query($params, $keyfather = '', $response = '', $ltrim = true) {
```
nineinchnick: Did you wrote this yourself or is there a source to referenced? Could you describe the arguments a bit more? It's not clear how they're used. Shouldn't they params be compatible with built-in `http_build_query`?

framework/helpers/Url.php
```diff
@@ -15,4 +15,31 @@
...
+                if(is_numeric($key)) {
+                    $key = '';
+                }
+                $key = "{$keyfather}[{$key}]";
+            }
+            if(is_array($value)) {
+                $response = Url::custom_http_build_query($value, $key, $response, false);
+            } else {
+                if(!is_null($value)) {
+                    $response .= "&".urlencode($key)."=".urlencode($value);
```
nineinchnick: Wouldn't it be better to build the `$response` as an array and implode it in the return statement?

framework/assets/yii.gridView.js
```diff
@@ -155,6 +172,6 @@
         data: function () {
             var id = $(this).attr('id');
             return gridData[id];
-        }
+        },
```
nineinchnick: In JS there shouldn't be a trailing comma after last object item.

tests/framework/helpers/UrlHelperTest.php
```diff
@@ -0,0 +1,37 @@
...
+ * StringHelperTest
+ *
+ * @group helpers
+ */
+class UrlHelperTest extends TestCase
+{
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->mockApplication();
```
nineinchnick: You only test one static method, maybe you don't need the whole application mock for it?

framework/helpers/Url.php
ebisbe: Yes. I've already thought on that but he implode needs to be made on the last return. I'll change it.

framework/helpers/Url.php
ebisbe: I wrote this myself. I'll describe the arguments more, if thought that they are mostly self explainable by the name but more explanation wouldn't hurt anybody. 
Yii is only using array values as parameteres when calling http_build_query. If needed I'll implement the object aswell but I'm not really sure is needed.

nineinchnick:
Overall, this is going very well. Good job @ebisbe! I'm really looking forward to getting this merged.

framework/helpers/Url.php
nineinchnick: This could be named just `http_build_query` without the _custom_ prefix and differences against the built-in function could be explained in phpdoc.

samdark:
@nineinchnick have you tested it?

nineinchnick:
@samdark I've used the JS file only and it works. I'll test the whole branch after the custom `http_build_query` is refactored a bit more.

ebisbe:
@nineinchnick Updated

framework/helpers/Url.php
```diff
@@ -15,4 +15,38 @@
...
+     * @param array  $params Array of values to convert into http query string
+     * @param string $inheritKey String used when array has more than one dimesion.
+     *                           We pass the our actual key to our child array.
+     * @param array  $response Array of urlencoded values
+     * @param bool $implode Used to return the response as http_build_query would return by default.
+     *                      If set false array is given in response
+     *
+     * @return array|string
+     */
+    public static function custom_http_build_query(Array $params, $inheritKey = '', $response = [], $implode = true) {
```
nineinchnick: Using the code below you could remove the `$response` and `$implode` arguments.

~~~php
<?php
public static function custom_http_build_query(array $params, $inheritKey = '') {
    $response = [];
	foreach($params as $key => $value) {
		if(!empty($inheritKey)) {
			if(is_numeric($key)) {
				$key = '';
			}
			$key = "{$inheritKey}[{$key}]";
		}
		if(is_array($value) || is_object($value)) {
			$response[] = Url::custom_http_build_query($value, $key);
		} else {
			if(!is_null($value)) {
				$response[] = urlencode($key)."=".urlencode($value);
			}
		}
	}

	return implode('&', $response);
}
~~~

ebisbe:
@nineinchnick done. I'm not sure if every time I PR you receive and email so I'm adding the comments

framework/helpers/Url.php
```diff
@@ -15,4 +15,35 @@
...
+    public static function custom_http_build_query(Array $params, $inheritKey = '') {
+        $response = [];
+        foreach($params as $key => $value) {
+            if(!empty($inheritKey)) {
+                if(is_numeric($key)) {
+                    $key = '';
+                }
+                $key = "{$inheritKey}[{$key}]";
+            }
+            if(is_array($value)) {
```
nineinchnick: You could also check for objects using `is_object($value)`. Some could be traversable.

ebisbe: But adding the is_object inside the foreach would meant that the params is still an array. We could check if params is an object or an array and also if the array has objects o more arrays. Something like:
```
public static function custom_http_build_query($obj, $inheritKey = '')
    {
        $response = [];
        $params = $obj;
        if (is_object($obj)) {
            $params = get_object_vars($obj);
        }

        foreach ($params as $key => $value) {
            if (!empty($inheritKey)) {
                if (is_numeric($key)) {
                    $key = '';
                }
                $key = "{$inheritKey}[{$key}]";
            }
            if (is_array($value)) {
                $response[] = Url::custom_http_build_query($value, $key);
            } elseif (is_object($value)) {
                $response[] = Url::custom_http_build_query(get_object_vars($value), $key);
            } else {
                if (!is_null($value)) {
                    $response[] = urlencode($key) . "=" . urlencode($value);
                }
            }
        }

        return implode('&', $response);
    }
```
If it's fine I'll need to think about different combination of unit tests.

nineinchnick: I wouldn't add `get_object_vars`, let's just assume the object is iterable, that is it would work in `foreach`. It's just an idea, we need to keep it simple.

ebisbe: Then this should work ( not tested yet ).
```
public static function custom_http_build_query($params, $inheritKey = '')
    {
        $response = [];

        foreach ($params as $key => $value) {
            if (!empty($inheritKey)) {
                if (is_numeric($key)) {
                    $key = '';
                }
                $key = "{$inheritKey}[{$key}]";
            }
            if (is_array($value) || is_object($value)) {
                $response[] = Url::custom_http_build_query($value, $key);
            } else {
                if (!is_null($value)) {
                    $response[] = urlencode($key) . "=" . urlencode($value);
                }
            }
        }

        return implode('&', $response);
    }
```
About this the point is that is not used and I'm not sure if will be ever be used so I don't see the point on doing it. Why do you want to implement it anyway?

nineinchnick: Yeah, sorry for the confusion. Maybe lets just drop the subject here.

samdark:
@nineinchnick any update?

framework/helpers/Url.php
ebisbe: @nineinchnick my part is done right now, right?

nineinchnick: I think so. This just needs more reviews now, as the url building is a critical part of the framework. I'm thinking about checking out if there are some unit tests for `http_build_query` in php sources and comparing their results with this implementation.

framework/helpers/Url.php
```diff
@@ -15,4 +15,35 @@
...
+     * For example:
+     * - http_build_query(['language' => ['en', 'ca']])           -> 'language%5B0%5D=en&language%5B1%5D=en'
+     * - custom_http_build_query(['language' => ['en', 'ca']])    -> 'language%5B%5D=en&language%5B%5D=ca'
+     *
+     * @param array  $params Array of values to convert into http query string
+     * @param string $inheritKey String used when array has more than one dimesion.
+     *                           We pass the our actual key to our child array.
+     * @return array
+     */
+    public static function custom_http_build_query(Array $params, $inheritKey = '') {
```
samdark: Should be moved to `BaseUrl`. `Url` is meant to be empty.

samdark: Method should be named `customHttpBuildQuery` as per naming guidelines.

SilverFire:
Seems to be resolved by https://github.com/yiisoft/yii2/commit/d3cfb87ba67a328d197d1a9e52dbb12a7bf8476b, #10302

I'm not happy with a `customHttpBuildQuery` crutch, because it breaks nested arrays:

```php
$silverfire = ['id' => 1, 'user' => 'silverfire'];
$samdark = ['id' => 2, 'user' => 'samdark'];
$url = customHttpBuildQuery(['users' => [$silverfire, $samdark]]); 
// users[][id]=1&users[][user]=silverfire&users[][id]=2&users[][user]=samdark
parse_str($url, $array);
// results in
array (size=1)
  'users' => 
    array (size=4)
      0 => 
        array (size=1)
          'id' => string '1' (length=1)
      1 => 
        array (size=1)
          'user' => string 'silverfire' (length=10)
      2 => 
        array (size=1)
          'id' => string '2' (length=1)
      3 => 
        array (size=1)
          'user' => string 'samdark' (length=7)
```

I'm closing this PR because it's generally overlapped.

Issue #9017, opened 2015-07-06
Add indexes on expressions to migrations


YasserHassan:
A use case would be using createIndex() to create case-insensitive unique indexes in PostgreSQL as in
```
CREATE UNIQUE INDEX test1_lower_email_idx ON test1 (lower(email));
```
as the above statement without `lower()` would allow the emails `test@example.com` and `Test@example.com` to coexist. Another example is mentioned at [Indexes on Expressions](http://www.postgresql.org/docs/9.4/interactive/indexes-expressional.html)

My suggested approach is adding a boolean option called `quoteColumnNames` to `createIndex()` that goes all the way down to `QueryBuilder::buildColumns()` to allow escaping column name quoting.

nineinchnick:
What if it would allow passing \yii\db\Expression as columns instead of adding a new argument?

YasserHassan:
`yii\db\Expression` does not work but by following the code it turned out that quoting the column name is escaped if it includes parenthesis, an opening "(" to be specific. It needs to be added to the documentation of `Migration::createIndex()`.

By going further, I found that auto-escaping is mentioned in the documentation of `QueryBuilder::createIndex()` which is internally used by `Migration::createIndex()` but I'd prefer if it is clarified more.

PR #9021, opened 2015-07-06
Clarify escaping column name quoting in createIndex()

YasserHassan:
Added information about when column names are not quoted when using createIndex() as discussed in #9017. "left parenthesis" was used rather than just "parenthesis" to reflect the exact code working.


samdark:
Thanks!

YasserHassan:
Thank you for the prompt action @samdark.

PR #9033, opened 2015-07-07
Fixes #9032: Improve scenarios usage docs [skip ci]

### LINKED ISSUE #9032: Improve scenarios usage docs
The default scenario is implemented via constant `SCENARIO_DEFAULT`. However, in multiple docs pages I saw scenarios are used as raw strings, for example: `'default'`, `'update'`. This is not right and can lead to errors. I saw many developers use just raw strings.

arogachev:
Sorry, didn't create separate branch. Is it required for docs changes?

Also I have some questions:
- Should I change translated docs as well? Or it's synchronized by the people who do translation?
- Should we add some additional notes about constants usage?

### DIFF
--- a/docs/guide/input-multiple-models.md
+++ b/docs/guide/input-multiple-models.md
@@ -33,8 +33,8 @@ class UserController extends Controller
             throw new NotFoundHttpException("The user was not found.");
         }
         
-        $user->scenario = 'update';
-        $profile->scenario = 'update';
+        $user->scenario = User::SCENARIO_UPDATE;
+        $profile->scenario = Profile::SCENARIO_UPDATE;
         
         if (Model::loadMultiple([$user, $profile], Yii::$app->request->post())) {
             if ($user->validate() && $profile->validate()) {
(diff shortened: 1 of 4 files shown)


samdark:
Separate branch is needed in order to submit multiple pull requests at once.

arogachev:
What should I do in order to correct this? Close this and create new one with separate branch?

arogachev:
Maybe it will be better if the docs will contail both ways of doing this. Also we need to mention that we need to create constants before using it, some new users can misunderstand it.

samdark:
It's OK as is if you aren't going to open another pull request.

That's why I asked about constants. These should be defined in classes.

arogachev:
So what's better to do now?

samdark:
I'll take care of it.

samdark:
Changed to constants in "Models" guide section where it makes sense because full classes are provided.

samdark:
Thanks for idea. It could result in better code in apps.

PR #9037, opened 2015-07-07
Update Validator.php

thiagotalma:


samdark:
What's the problem it's trying to fix?

thiagotalma:
It's jQuery.

`$('#country').value` = undefined (always)
`$('#country').val()` = input value

cebe:
Thanks!

PR #9038, opened 2015-07-07
Added warning log to FileCache when the 'file_put_contents' method returned false

lattn:


framework/caching/FileCache.php
```diff
@@ -139,6 +139,8 @@ protected function setValue($key, $value, $duration)
 
             return @touch($cacheFile, $duration + time());
         } else {
+            $error = error_get_last();
+            Yii::warning("Unable to write file '{$cacheFile}': {$error['message']}", __METHOD__);
```
cebe: better: `"Unable to write cache file '{$cacheFile}': {$error['message']}"`

cebe:
also need a changelog line

samdark:
Thanks!

PR #9051, opened 2015-07-09
Add note about composer requirements

kirkhansen:
Add a note about the `fxp/composer-asset-plugin:1.0.0` dependency.


samdark:
Thanks!

PR #9057, opened 2015-07-09
 Adds `.required` class if method `when` is true

### LINKED ISSUE #5696: Problem with "Validator::when" and .required class for label
My validation rule in AR:

``` php
......
[
                'reason_for_cancel',
                'required',
                'when' => function (Job $model) {
                    return $model->status ===  'cancelled';
                },
                'whenClient' => function () {
                    return "function (attribute, value) {
                           return $('#job-status').val() == 'cancelled';
                     }";
                },
            ],
]
....
```

But ActiveForm does not generate .required class for input when i change status to "cancelled"

### LINKED ISSUE #6744: How to force validate when using "when" in conditional validation?
my Model code:

```
//...
public function rules()
    {
        return [
            ['companyId', 'required', 'when' => function($model) {
                return $model->Grade == '1';
            }, 'whenClient' => "function (attribute, value) {
                return $('#user-grade').val() == '1';
            }"],
        ];
    }
//...
```

I want validate companyId as required when Grade =="1"

but  it seems not work when user-grade changed  in browser,

even $('#user-grade,#user-companyid').trigger("change"),javascript validation dos not work.

so I checked yii.activeForm.js:

```
//...
   var validateAttribute = function ($form, attribute, forceValidate, validationDelay) {
        var data = $form.data('yiiActiveForm');

        if (forceValidate) {
            attribute.status = 2;
        }
        $.each(data.attributes, function () {
            if (this.value !== getValue($form, this)) { //  value not changed
                this.status = 2;
                forceValidate = true;
            }
        });
        if (!forceValidate) { // value not changed , do not needs to validate
            return;
        }
//...

```

so , I wonder how to force validate field value when using conditional validation?

githubjeka:
Fixes #5696
Fixes #6744

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -455,8 +455,7 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      *
      * Note that when the validator has a conditional validation applied using
      * [[\yii\validators\RequiredValidator::$when|$when]] this method will return
-     * `false` regardless of the `when` condition because it may be called be
-     * before the model is loaded with data.
+     * `false` or `true` of depending on the result of closure `$when`.
      *
      * @param string $attribute attribute name
      * @return boolean whether the attribute is required
@@ -464,8 +463,11 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
     public function isAttributeRequired($attribute)
     {
         foreach ($this->getActiveValidators($attribute) as $validator) {
-            if ($validator instanceof RequiredValidator && $validator->when === null) {
-                return true;
+            if ($validator instanceof RequiredValidator) {
+                if ($validator->when === null) {
+                    return true;
+                }
+                return call_user_func($validator->when, $this, $attribute);
             }
         }
         return false;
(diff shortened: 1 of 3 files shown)


cebe:
that works of course but what if condition changes on client side? It should either react to that or stay as is.

githubjeka:
@cebe  https://github.com/yiisoft/yii2/pull/9057/files#diff-c817e4a599a24a005e9a74901522642eR113

First change is work correctly, second change can not removed css class because not possible. 
https://github.com/yiisoft/yii2/blob/master/framework/widgets/ActiveField.php#L701
```php
if ($validator->whenClient !== null) {
    $js = "if (({$validator->whenClient})(attribute, value)) { $js }";
}
```

Will be BC.

cebe:
doesn't this break BC in behavior of the form? You even change the test return value.
To properly fix this we need to define different situations and write more tests for it.

SilverFire:
Risky change, no tests, no public demand. Closing this PR.

githubjeka:
Mersi

PR #9061, opened 2015-07-09
Added force validation for `whenClient`

### LINKED ISSUE #8382: Conditional statement - does not always trigger a new validation
Hello,

I noticed that the conditional statement is not always working correctly on the client side with Yii2.

For example: create a statement where the attribute year in a model should be required when a name is not empty. Make a form and then add the two fields. Now, fill the name and go to the year. Leave the field without entering data. The validator says the field is required according to the rule. That's good.
Now empty the name field and go to the year field. Leave the field again without entering data. The validator is now not triggered and so the message that it is required remains there, when it is actually now not required anymore.

Perhaps it would be better to trigger the validator on "focusout" instead of on a change.

Best regards.

githubjeka:
Fixes #8382


SilverFire:
@githubjeka Sorry for the late reply. Your patch definitely fixes the problem, but as a side-effect, it causes redundant AJAX queries on `TAB` navigating from field to field when `enableAjaxValidation` is true.

I still think that it may be merged because `validateOnBlur` may be turned off.

@yiisoft/core-developers nobody complains?

samdark:
`validateOnBlur` default is on.

samdark:
Is it possible to issue a query only if it's conditional statement somehow?

samdark:
If not, solution is OK to me.

SilverFire:
Maybe passing `hasWhen` flag to the client side, but I don't think that it's necessary.

SilverFire:
@githubjeka could you add a CHANGELOG line, please

SilverFire:
Merged https://github.com/yiisoft/yii2/commit/88241f24f37498ffcd99d25248edebf2f12ed9e5! Thank you!

PR #9068, opened 2015-07-09
unused imports removed

schojniak:


framework/db/BaseActiveRecord.php
```diff
@@ -8,7 +8,6 @@
 namespace yii\db;
 
 use yii\base\InvalidConfigException;
-use yii\base\Event;
```
cebe: this one is used in the docs e.g. in line 46

samdark:
Thanks!

schojniak:
You're right. I restored this import.

PR #9071, opened 2015-07-10
Translated the word 'Update'

squio:
In general the word 'Update' can be better translated into 'Bewerk' in Dutch.


samdark:
Thanks!

squio:
Cool that was fast; thanks!

PR #9076, opened 2015-07-10
Fixes PageCache not setting any duration when caching response data

kidol:


cebe:
same issue with dependency as far as I see.

kidol:
@cebe Yes missed that one. `$variations` are handled already.

cebe:
Merged, thanks!

PR #9089, opened 2015-07-13
Add comment method to SchemaBuilder

CTOlet:
Schema::integer()->comment('Definition');

### DIFF
--- a/framework/db/SchemaBuilder.php
+++ b/framework/db/SchemaBuilder.php
@@ -28,6 +28,7 @@ use yii\base\Object;
  * ```
  *
  * @method SchemaBuilder default($default = null) see [[SchemaBuilder::_default()]] for more info
+ * @method SchemaBuilder comment($comment = null) see [[SchemaBuilder::_comment()]] for more info
  *
  * @author Vasenin Matvey <vaseninm@gmail.com>
  * @since 2.0.5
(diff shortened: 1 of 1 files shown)


samdark:
`comment()` is not reserved keyword. Why adding it into `__call()`?

Nemmo:
I got it just now, should I fix that by implementing ``` comment()``` method?

samdark:
Yes.

vaseninm:
It works on postgress?

Nemmo:
I have not tested it on postgre, but as I know postgre has another writing rules about comments on columns

vaseninm:
@Nemmo yes. It does not work on postgre. But your method must be a cross platform.

valonsoft:
Explain

cebe:
http://www.postgresql.org/docs/current/static/sql-comment.html

framework/db/SchemaBuilder.php
```diff
@@ -418,6 +441,16 @@ protected function buildDefaultString()
     }
 
     /**
+     * Returns string with comment string of column
+     *
+     * @return string
+     */
+    protected function buildCommentString()
+    {
+        return $this->comment === null ? '' : " COMMENT '{$this->comment}'";
```
Faryshta: this needs sanitization

samdark:
PR can't be merged because it's MySQL only and has issues.

SilverFire:
Implemented in fde8aa6bcbb902aaea10f02f07fa87e11e5d2d17

PR #9127, opened 2015-07-16
Fix rename table function for MSSQL

sitawit:
Current sql will generate SQL that looks like this
`
EXEC sp_rename '[old_table]', '[new_table]'
`
This makes the new table has a new `[new_table]` instead of `new_table`. It could be a flaw within `sp_rename` stored procedure.

However, by removing a single quote to make SQL becomes this
`
EXEC sp_rename [old_table], [new_table]
`
The new table has a name `new_table` as expected.


samdark:
Thanks!

PR #9128, opened 2015-07-16
Fix rename column function for MSSQL

sitawit:
Current function will generate sql that looks like this.
`
EXEC sp_rename '[table].[old_col]', '[new_col]', 'COLUMN';
`
This will result in having `[new_col]` instead of `new_col`.

So it needs to be fixed into
`
EXEC sp_rename '[table].[old_col]', [new_col], 'COLUMN';
`
Removing the single quote from `[new_col]` will result in having `new_col` as expected. The single quote for `[table].[old_col]` has to stay. Without it an error from server will occur.


samdark:
Thanks!

PR #9136, opened 2015-07-16
Update yii.php for arabic language

abdullah-dev11:
Urgent Fix for arabic language , This will stop yii2 from working if arabic language is chosen.

there is strange space making this problem


samdark:
Thanks!

Issue #8595, opened 2015-05-28
Cache DbManager checkAccessFromCache calls checkAccessRecursive


achretien:
In https://github.com/yiisoft/yii2/blob/master/framework/rbac/DbManager.php

The checkAccessFromCache method call the checkAccessRecursive one :

```php
 protected function checkAccessFromCache($user, $itemName, $params, $assignments)
    {
 
...
        if (!empty($this->parents[$itemName])) {
            foreach ($this->parents[$itemName] as $parent) {
                if ($this->checkAccessRecursive($user, $parent, $params, $assignments)) {
                    return true;
                }
            }
        }
...
    }
```

Shouldn't it call `checkAccessFromCache` instead ?

qiangxue:
Thanks!

Issue #7743, opened 2015-03-16
Cache RBAC assignments


RomeroMsk:
Save `SELECT * FROM auth_assignment WHERE user_id='...'` results in `User` component property to avoid duplicate DB queries on one page.
[See comment here](https://github.com/yiisoft/yii2/issues/3168#issuecomment-76709786).

RomeroMsk:
Any thoughts?

williamyiu:
I ran into the same issue as well.

juban:
:+1: 
I also think this is a major performance issue in some scenarios.
I'm maybe off topic here (sorry), but here is a very simple workaround :

```php
class DbManagerEx extends DbManager
{
    private $_assignments = [];
    
    public function getAssignments($userId)
    {
        // Avoid multiple queries per request
        if(!isset($this->_assignments[$userId]))
            $this->_assignments[$userId] = parent::getAssignments($userId);
        return $this->_assignments[$userId];
    }

}
```

acuom:
@samdark did this ever get done? 

as I am getting hundreds of 

`SELECT `b`.* FROM `auth_assignment` `a`, `auth_item` `b` WHERE ((`a`.`item_name`=`b`.`name`) AND (`a`.`user_id`='91')) AND (`b`.`type`=1)`

On some pages 👎

samdark:
Not yet. Scheduled for next release though.

acuom:
@samdark ahh ok, any tips how the above problem could be solved in the meantime?

samdark:
Try using file storage instead.

PR #9138, opened 2015-07-16
Cache rbac assignments

nineinchnick:
Based on #8810, makes caching assignments optional and disabled by default.


framework/rbac/DbManager.php
```diff
@@ -885,8 +921,8 @@ public function loadFromCache()
         }
 
         $data = $this->cache->get($this->cacheKey);
-        if (is_array($data) && isset($data[0], $data[1], $data[2])) {
-            list ($this->items, $this->rules, $this->parents) = $data;
+        if (is_array($data) && isset($data[0], $data[1], $data[2], $data[3])) {
```
nineinchnick: This will fail if caching assignments is disabled because the fourth item is null. Should `array_keys_exists` or just `count()` be used here? Or should the `assignments` property have an empty array as the default value?

nineinchnick:
Actually it might be better to cache assignments for specific user in `getAssignments()` and `getAssignment()`, going to update the PR.

klimov-paul:
@nineinchnick, the referred issue has been closed and there is a resolved issue #8595.
Can we close this PR?

nineinchnick:
This is a bit different. It's about to avoid fetching assignments in each request.

nineinchnick:
I know there's a valid argument that sometimes there's a large number of assignments and caching them all is not worth it, that's why its optional. But there are other cases where are few or none assignments because default roles are used and then this could help avoid unecessary queries.

klimov-paul:
Relates #9659

nineinchnick:
This could be enhanced to cache for single request only by default and optionally allow using a cache component.

samdark:
It makes no sense to use cache component to cache for a single request.

nineinchnick:
I ment the PR can be enhanced to cache using a local var and _optionally_ using a cache component.

nineinchnick:
I agree with your arguments about the dangers of caching assignments but still some applications could benefit from it. I'm using only web based user managment and take care to only use auth manager for auth related tasks - never directly modify the db.

samdark:
Well, it could be introduced as a separate option that has a **huge** warning in phpdoc about dangers of using it...

samdark:
Can't merge it as is. See #7743 for discussion.

framework/rbac/DbManager.php
```diff
@@ -80,6 +80,10 @@ class DbManager extends BaseManager
      */
     public $cache;
     /**
+     * @var bool should assignments be cached
+     */
+    public $cacheAssignments = false;
```
samdark: No need making it optional for per-request caching, I think...

framework/rbac/DbManager.php
```diff
@@ -118,8 +126,8 @@ public function init()
      */
     public function checkAccess($userId, $permissionName, $params = [])
     {
-        $assignments = $this->getAssignments($userId);
         $this->loadFromCache();
+        $assignments = $this->getAssignments($userId);
```
samdark: What was it moved below `loadFromCache`?

framework/rbac/DbManager.php
```diff
@@ -600,6 +608,10 @@ public function getAssignment($roleName, $userId)
         if (empty($userId)) {
             return null;
         }
+        if ($this->cacheAssignments) {
```
samdark: Isn't it better to query `$this->assignments` here?

samdark: Current code would result in extra SQL query from `getAssignments` in case these aren't in the cache yet.

samdark:
Found another problem with it. If RBAC is used in a long running process (via CLI) there could be a need to refresh PHP cache from time to time. There's a way to do it for current PHP cache in `master` via `invalidateCache()` but there's no way to reset cache for assignments in this pull request.

framework/rbac/DbManager.php
nineinchnick: We've settled that this is dangerous for long running processes and is questinable when there's a large number of assignments (users). It might require to be documented elsewere and possibly the base applications config should be updated.

framework/rbac/DbManager.php
nineinchnick: Currently this PR does not use the cache component for assignments but I has before, so this is a leftover. It should not have any side effects.

framework/rbac/DbManager.php
nineinchnick: It does query assignments, just through `getAssignments()`. There's no extra query executed, `getAssigmnents()` would populate the query and the result would be returned immediately.

nineinchnick:
I've added clearing the `assignments` property in `invalidateCache()`.

framework/rbac/DbManager.php
samdark: If there's nothing in the cache? Nope. It will get to this query and execute it: https://github.com/yiisoft/yii2/pull/9138/files#diff-85a8556277e468a30afc79beb7aa25dfR681. This code which makes more optimal query https://github.com/yiisoft/yii2/pull/9138/files#diff-85a8556277e468a30afc79beb7aa25dfR653 would be never reached.

nineinchnick: It will populate the cache and return a single result from it. If the number of assignments for a single user is not big there won't be any significant difference in the query speed as the execution itself has larger overhead than fetching data. Obviously I'd need a benchmark to back it up but I've been making such tests in Yii 1.

samdark: Well, https://github.com/yiisoft/yii2/pull/9138/files#diff-85a8556277e468a30afc79beb7aa25dfR653 is never reached and should be removed anyway, right?

nineinchnick: No, it's reached when caching is disabled.

jafaripur:
When this pull request will be merge?

samdark:
@yiisoft/core-developers need additional review for it.

PR #9140, opened 2015-07-17
remove redefined parent property

matthewyang:
Property $controller already defined in parent class \yii\base\Application, both console\Application and web\Application already inherit it.

No need to redefine it in both classes.

### DIFF
--- a/framework/console/Application.php
+++ b/framework/console/Application.php
@@ -64,11 +64,6 @@ class Application extends \yii\base\Application
      * Defaults to true.
      */
     public $enableCoreCommands = true;
-    /**
-     * @var Controller the currently active controller instance
-     */
-    public $controller;
-
 
     /**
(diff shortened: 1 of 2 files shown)


samdark:
Thanks!

cebe:
this is intentional to have better IDE type completing for web and console Controller...

samdark:
Fixed via phpdoc.

matthewyang:
Which IDE?
There are many other property inherited, why only need to redefine controller?
Can you show me some doc to read?

samdark:
@matthewyang based on which app it is, IDE returned different Controller classes (either console or web).

matthewyang:
@samdark , I see. So $controller in \yii\console\Application is actually the object of class \yii\console\Controller, right?

cebe:
yes

PR #9146, opened 2015-07-17
Update Request.php

ptheofan:
Using `isset()` in getBodyParams treats `null` as `$defaultValue` which is wrong since `$defaultValue` should be returned only when parameter is missing!


cebe:
well, `null` could be interpreted as "not set". we can not change this in 2.0.x as it breaks BC.

cebe:
@ptheofan could you explain more about why you think this is needed. Show a real case where current behavior causes a problem?

PR #9147, opened 2015-07-17
Variable was being enclosed in an array in comparison example

usecomposable:
Haven't noticed this behaviour to be the case after following the examples.


klimov-paul:
Thank you

cebe:
As far as I see, the original code is correct: http://www.yiiframework.com/doc-2.0/yii-di-container.html#get%28%29-detail

klimov-paul:
Yes, fix is incorrect - my mistake.

PR #9148, opened 2015-07-18
Revert "Variable was being enclosed in an array in comparison example"

cebe:
Reverts yiisoft/yii2#9147

> As far as I see, the original code is correct: http://www.yiiframework.com/doc-2.0/yii-di-container.html#get%28%29-detail


klimov-paul:
Perhaps the example should be adjusted to 
```php
$engine = $container->get('app\components\SearchEngine', [$apiKey, ['type' => 1]]);
```

klimov-paul:
Resolved by commit 77c1fcf. I think this example will not produce any confusion now.

PR #9178, opened 2015-07-22
Introduced Command interface

samdark:
https://sourcemaking.com/design_patterns/command

My intent:
- Prepare for queues.
- Would allow to create standard command executor components.

### DIFF
--- a/framework/base/Command.php
+++ b/framework/base/Command.php
@@ -1,0 +1,22 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\base;
+
+/**
+ * Command incapsulates parameters and logic for performing an action and is useful for:
+ *
+ * - Decoupling the action itself from when it's executed i.e. scheduling for later execution.
+ * - Moving context out of the action so it can be reused in many different contexts.
+ *
+ * @author Alexander Makarov <sam@rmcreative.ru>
+ * @since 2.0.6
+ */
+interface Command
+{
+    public function execute();
+}


nineinchnick:
Looks a bit like an auth rule. They're also serialized and stored in the db.

klimov-paul:
I am unsure I understand the intent.
Can you provide some code snippet for usage example?

samdark:
```php
class FetchFacebookInfo implements Command
{
    private $userID;

   public function __construct($userID)
   {
        $this->userID = $userID;
   }

    public function execute()
    {
        $user = User::findOne($this->userID);
        // ... fetch data from facebook ... (it takes time)
        $user->email = $facebookData['email'];
        $user->username = $facebookData['username'];
        $user->save();
    }
}
```

Web controller:

```php
class SiteController extends Controller
{
    public function actionSignup()
    {
        $user = new User();
        // usual signup stuff
        if ($user->save()) {
            // fetching facebook data isn't critical, could do it later
            $command = new FetchFacebookInfo($user->id);
            Yii::$app->queue->push($command);
        }
    }
}
```

klimov-paul:
Such abstraction is untypical for Yii. It is too... abstract.
Such addition may be useful in queue implementation, but it does not seem to be general unless there is any check for this interface inside the framework.

cherifGsoul:
Very interesting, in some complicated business logic this can be very useful, for example we can use it for spliting ActiveRecord as Command and Query, the logic in commands can be used with any UI or delivery system (Web or console)

```php
class AddTodoCommand implements Command
{
   private $toDoListId;
   private $priority;
   private $description;

  public function __construct($todoListId,$priority,$description){
   $this->todoListId=$todoListId;
   $this->description = $description;
   $this->priority = $priority;
  }

 public function execute() {
  $list = Yii::$app->listFinder->findById($this->todoListId);
  $list->addTodo($priority,$description);
 }
```

samdark:
It's quite similar to `BootstrapInterface` and `Configurable` in its spirit. There will be `instanceof` checks in at least queue component.

samdark:
@nineinchnick serialization itself isn't part of this interface.

samdark:
@klimov-paul probably you're right and the interface itself makes sense only for domain part of the app which isn't covered by framework in any way.

cherifGsoul:
@samdark for the same I like having command bus component instead

samdark:
Bus component isn't really useful. For anything that's currently in Yii and isn't queue replacement in any way. I guess I'll close this pull request.

Issue #8803, opened 2015-06-15
base\Model::load() return documentation is misleading


tom--:
The [docs for yii\base\Model::load()](https://github.com/yiisoft/yii2/blob/master/framework/base/Model.php#L791) says it returns:

>       boolean whether the model is successfully populated with some data

when it also returns true if it populates nothing at all, for example if the data contains no safe attributes.

I don't have a suggestion for better text. I think `load()` returns false if and only if the data are not structured in a way it understands. But that's not a very useful text for the docs.

ajayvembu90:
The 'load()' method in Model class is mostly overloaded either with 'Yii::$app->request->post()' or 'Yii::$app->request->get()', so it returns false when there is no request and true if there is a request and some of the model fields are present.

If you try to use 'load()' without a request it will always return true as some includes none.

tom--:
In normal English, "some" and "none" are mutually exclusive. For example, "Some apples are on the table," is a false statement if there are in fact "none".

I wasted time looking in the wrong places for why a `assertFalse($model->load([test data]))` was failing. Eventually I looked at source of `load()` and saw that the test is wrong because it was designed according to the described behavior instead of the actual behavior.

PowerGamer1:
I think the behavior of `Model::load()` should be changed to actually be as currently documented, i.e.: `load()` should return `true` only if at least ONE of the model fields was populated with value from provided array.

Currently, load() basically tests if provided array was empty or not, caller of `load()` method could perform such test easily using PHP's built-in functionality. On the other hand, a test to see if at least a single form field was provided in $_GET or $_POST would be quite cumbersome to implement outside of Model::load() method (which, in contrast could do such test rather easily).

Besides, when model formName() returns an empty string (makes URL prettier when submitting form by GET) and $_GET contains only some additional params (which are not related to the HTML input form at all) and not a single param from the corresponding form model class, Model::load() will return `true` which is very bad, because it does not allow to distinguish between the first time opening of the page with form and form submission.

PR #9182, opened 2015-07-22
Fixed returned value of Model::load() to be as documented

PowerGamer1:
Changed the behavior of Model::load() to be as documented, i.e.: return true if at least one of the model fields was populated with a value from provided $values array.

For a rationale see https://github.com/yiisoft/yii2/issues/8803#issuecomment-123657742.

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -682,21 +682,25 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      * @param array $values attribute values (name => value) to be assigned to the model.
      * @param boolean $safeOnly whether the assignments should only be done to the safe attributes.
      * A safe attribute is one that is associated with a validation rule in the current [[scenario]].
+     * @return boolean true if at least one of the model fields was assigned a value from provided $values array.
      * @see safeAttributes()
      * @see attributes()
      */
     public function setAttributes($values, $safeOnly = true)
     {
+        $populated = false;
         if (is_array($values)) {
             $attributes = array_flip($safeOnly ? $this->safeAttributes() : $this->attributes());
             foreach ($values as $name => $value) {
                 if (isset($attributes[$name])) {
                     $this->$name = $value;
+                    $populated = true;
                 } elseif ($safeOnly) {
                     $this->onUnsafeAttribute($name, $value);
                 }
             }
         }
+        return $populated;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


cebe:
such change will break BC, so it can not be included in 2.0.x. The original purpose of load() is to have a shortcut for the following code:

```php
if (!empty($_POST['Model'])) {
    $model->attributes = $_POST['Model'];
    if ($model->save()) {
        // something
    }
}
// ...
```
to be shortended as:

```php
if ($model->load($_POST) && $model->save()) {
        // something
}
// ...
```

your use case is a bit different, using load() does not really make sense when the formName is empty, you can just assign the attributes directly in that case.

samdark:
I think docs should be adjusted in 2.0.x since if we'll merge this change, it won't be possible to submit empty form and get validation error.

PowerGamer1:
> using load() does not really make sense when the formName is empty

It makes the exact same sense as in case when the from name is specified (see explanation below).

>  you can just assign the attributes directly in that case

It won't help at all, its about determining if form was submitted or not.

> ```php
if (!empty($_POST['Model']))
```

How would you express the above check from "original purpose" with words? Like this: "if we got in a subset of $_POST which stores submitted form field values AT LEAST a SINGLE form field - meaning the form was submitted". If so we should proceed with handling submission accordingly. Such check is exactly what my version of load() does - no changes in functionality (when form name is set) whatsoever.

Now, in case form name is not set. Isn't it natural to ask the same question too: "if we got in a subset of $_POST which stores submitted form field values AT LEAST a SINGLE form field - meaning the form was submitted". The only difference from the case when form has name is that you can't easily isolate "a subset of $_POST which stores submitted form field values" when form has no name. Of course such check can still be performed with a bit more work than a simple `!empty($_POST['Model'])` - by seeing if ANY of form field names are present in $_POST - exactly what my version of load() does.

Now, what does the current version of load() do when form has no name? Does the return value of current version of load() correspond with the "original purpose" - i.e. answers the question:  "if we got in a subset of $_POST which stores submitted form field values AT LEAST a SINGLE form field - meaning the form was submitted"? NO, for example, when $_POST has values not related to form the load() returns true - and that is a BUG and not BC break, because the return value of load() in this case does not conform to "original purpose".

> it won't be possible to submit empty form and get validation error.

If you imply that would happen with my version of load() I don't see how - when you left some input unfilled before pressing submit you will get $_POST['formName']['inputName'] == '' - both load()s will produce the same result (return value).

cebe:
closing as #8803 is solved. And we also decided not to return anything in setAttributes() in another issue (can not find it right now).

PR #9184, opened 2015-07-22
Fixes #9183: ErrorHandler not log some errors

### LINKED ISSUE #9183: ErrorHandler::handleError() should log errors not included into error_reporting()
Yii by default uses own errorHandler, that handle all errors, but ignore them by error_reporting level setting. In my oppinion it's comon situation you wont stop execution of entire application when notice, warning or strict error raised. But you may want to catch and log this errors in a common way by yii log.
And you dont want to miss these errors, instead you should see them in log files and deside how fix them or suppress by @ operator.

mrsombre:
fixes #9183


mrsombre:
@cebe 9 months over and still be verified, 5 lines of code.

Issue #9102, opened 2015-07-14
SchemaBuilder TYPE_TEXT bugs


nkovacs:
[SchemaBuilder::text](https://github.com/yiisoft/yii2/blob/master/framework/db/SchemaBuilder.php#L110) allows you to specify a length. This is wrong, you cannot use length for text fields: http://dev.mysql.com/doc/refman/5.7/en/create-table.html

Instead, you have to use TINYTEXT,  TEXT, MEDIUMTEXT or LONGTEXT. SchemaBuilder does not support these, it always uses TEXT (see also #8950)

nkovacs:
MySQL doesn't fail (it might throw a warning depending on server configuration, I didn't check).

But on postgres you get this error:

```
create table test (foo text(255));

ERROR:  type modifier is not allowed for type "text"
```

samdark:
OK. What's your proposal?

nkovacs:
The length parameter should be removed. I don't think it's used by any of the databases Yii supports. They either ignore it or error out like postgre.

Alternatively, the length parameter could select between TINYTEXT, MEDIUMTEXT etc. on mysql, but like I said in #8950, I think that's a bad idea, because it's a mysql-specific workaround (at first I thought the length parameter did that, before I checked the code).

nkovacs:
2.0.5's release date is almost here. If you release it like this, you won't be able to remove the parameter without breaking BC.

I've checked the documentation of all the supported database systems:

* Cubrid: [text is converted to varchar] (https://github.com/yiisoft/yii2/blob/master/framework/db/cubrid/QueryBuilder.php#L27), which can accept a length parameter. Omitting it defaults to the maximum length.
* MSSQL: does not accept length according to [documentation](https://msdn.microsoft.com/en-us/library/ms187993.aspx).
* MySQL: length is ignored, may cause warnings or errors depending on server settings.
* Oracle: [text is converted to clob](https://github.com/yiisoft/yii2/blob/master/framework/db/oci/QueryBuilder.php#L30). Does not accept length according to [documentation] (http://docs.oracle.com/cd/B28359_01/server.111/b28286/sql_elements001.htm#i54330)
* PostgreSQL: throws an error if you add a length parameter.
* SQLite: lengths are [ignored](https://www.sqlite.org/datatype3.html). I tested this with sqlite 3:

```
sqlite> create table foo (bar varchar(2), baz text(2));
sqlite> .schema
[...]

PR #9191, opened 2015-07-23
Portability fixes for SchemaBuilder.

nkovacs:
Removed length from integer and boolean types. This is not supported on PostgreSQL,
and is only used for display in MySQL, not input validation.

Removed length from text and binary type. This is not supported on most RDBMS's and ignored by others.

Changed binary mapping to `varbinary(max)` on MSSQL instead of `binary(1)` (which requires a length and is fixed length) to better match others.

Changed timestamp mapping to datetime on MSSQL. Timestamp is a unique number on MSSQL, not a timestamp.

Removed precision and scale parameters from floating point types. This is not supported by PostgreSQL, among others.

Renamed timestamp length parameter to precision. The precision parameter is not supported on all RDBMS's, and may throw an error. This can be fixed either by removing the precision parameter, or overriding the SchemaBuilder of those RDBMS's to ignore the precision parameter.
**This still needs to be fixed**, though  it's not a problem on MySQL and PostgresQL.

Since the release date of 2.0.6 is close, I've opted to remove these parameters. Removing them later would require a BC break. Adding them in later (though I don't think that's a good idea) would not.
I think it would be a better idea to add a `length`, `precision` etc. function, so that if you know you're using a DB that supports these, you can do:

```
    'foo' => Schema::integer()->length(10)
    'bar' => Schema::float()->precision(4)->scale(4)
```

See #9102


framework/db/mssql/QueryBuilder.php
```diff
@@ -34,10 +34,10 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_DOUBLE => 'float',
         Schema::TYPE_DECIMAL => 'decimal',
         Schema::TYPE_DATETIME => 'datetime',
-        Schema::TYPE_TIMESTAMP => 'timestamp',
+        Schema::TYPE_TIMESTAMP => 'datetime',
         Schema::TYPE_TIME => 'time',
         Schema::TYPE_DATE => 'date',
-        Schema::TYPE_BINARY => 'binary(1)',
```
samdark: https://msdn.microsoft.com/en-us/library/ms188362.aspx

Fixed length makes more sense, doesn't it?

nkovacs: TYPE_BINARY is blob elsewhere, which is variable length.

nkovacs:
Ok, so QueryBuilder::getColumnType will remove the lengths if the value in typemap does not have a length, so `text(255) NOT NULL` is converted to `text NOT NULL`. This means the DB won't throw errors,  so there's no need to remove the parameter where some RDBMS's support it.

samdark:
@nkovacs yep. Can you adjust pull request accordingly?

nkovacs:
Ok, I'll put the length back where it makes sense.

samdark:
btw., changing public or protected method signature is always a backwards compatibility break. It doesn't matter if it's removing parameters or adding parameters.

nkovacs:
I still have to check how float works, since the typemap doesn't have a length.

> btw., changing public or protected method signature is always a backwards compatibility break. It doesn't matter if it's removing parameters or adding parameters.

Yeah, but if it has a default value, only subclasses are affected, and you've done that before in point releases. It's still a smaller break.

nkovacs:
Ok, so the problem with float and double is that the only typemap that includes a length is cubrid's, but cubrid only accepts a single parameter, not two. So there's no point in having precision and scale for those two.

nkovacs:
I had to comment out two tests on Cubrid, because they won't work without the float parameter.

framework/db/mssql/QueryBuilder.php
samdark: OK.

samdark:
@nkovacs you mean Cubrid doesn't work w/o float parameter?

nkovacs:
I removed the precision and scale parameters from the float function, because only cubrid used it, but cubrid only accepts one parameter, and it might fail if you specify both (e.g. `float(4,4)`), and I think getColumnType will allow that.
Two of the tests tested that the float parameter works, i.e. that `Schema::float(16)` creates a column that's `float(16)`, not the default in typemap (`float(7)`), so those tests cannot pass now. For the other RDBMS's, these tests pass, because the parameters to `Schema::float` are ignored.

samdark:
Then it makes sense to remove these tests instead of commenting them out.

nkovacs:
Those tests also verify that `Schema::TYPE_FLOAT . '(16)'` is converted to `float(16)`, which works on cubrid because the typemap allows it.

cebe:
why did you remove the precision from float and double then?

nkovacs:
Another possibility is to keep the precision and scale parameters, and extend CubridSchemaBuilder so that it ignores scale. Then you can add support for the parameters to other db's later.

framework/db/mssql/QueryBuilder.php
cebe: these changes are BC breaking, separated them into another PR and tagged it for 2.1: #9305

nkovacs: Yeah, but this is totally broken.

cebe: sure but if someone was using these in MSSQL we can not just break their code.

nkovacs: I know. I guess MSSQL is not used that often, so it's not that critical.

PR #9193, opened 2015-07-23
[Fixes #9177] Password Hash Cost setting for Security component

### LINKED ISSUE #9177: Change cost for generatePasswordHash when testing
With the current default cost of 13, every generation of password goes at cost of performance of the testing process of at least 600ms per occurence.

Because tests dont need to be delayed by this cost, it would be a nice idea to lower the cost for generatePasswordHash during YII_ENV_TEST. Setting the value to 4 will work great, almost instantly. Anythin below is giving issues (dont know why).

This would cut my total testing length with many seconds.

freezysko:
Fixes #9177


samdark:
Thanks!

cebe:
:+1:

PR #9199, opened 2015-07-24
using \yii\base\Behavior::events and passing variables to event handlers

smsteel:
For using `\yii\base\Behavior::events` and passing variables to event handlers:

```
public function events()
{
    return [
        ActiveRecord::EVENT_BEFORE_INSERT => [[$this, 'myMethod'], 'parameter']
    ];
}
```

### DIFF
--- a/framework/base/Behavior.php
+++ b/framework/base/Behavior.php
@@ -69,8 +69,15 @@ class Behavior extends Object
     public function attach($owner)
     {
         $this->owner = $owner;
-        foreach ($this->events() as $event => $handler) {
-            $owner->on($event, is_string($handler) ? [$this, $handler] : $handler);
+        foreach ($this->events() as $event => $handler_data) {
+            if ((is_array($handler_data) && is_array($handler_data[0]) && $handler_data[1])) {
+                $handler = $handler_data[0];
+                $data = $handler_data[1];
+            } else {
+                $handler = $handler_data;
+                $data = null;
+            }
+            $owner->on($event, is_string($handler) ? [$this, $handler] : $handler, $data);
         }
     }
 


samdark:
What's your use case?

smsteel:
I have Behavior attached to ActiveRecord and it must do specific calls to external API when events triggered. I can write 3 methods containing the exact same calls, but different parameters in current Yii2:

    public function events()
    {
        return [
            ActiveRecord::EVENT_BEFORE_INSERT => 'myMethodInsert',
            ActiveRecord::EVENT_BEFORE_UPDATE => 'myMethodUpdate',
            ActiveRecord::EVENT_BEFORE_DELETE => 'myMethodDelete',
        ];
    }
   
    public function myMethodInsert()
    {
        $this->beforeAPICall();
        Api::call(['parameter1', 'parameter2']);
        $this->afterAPICall();
    }

    // etc...

Or i can use the same method, but check for $event->name:

    public function events()
    {
        return [
            ActiveRecord::EVENT_BEFORE_INSERT => 'myMethod',
            ActiveRecord::EVENT_BEFORE_UPDATE => 'myMethod',
            ActiveRecord::EVENT_BEFORE_DELETE => 'myMethod',
        ];
    }

    public function myMethod($event)
    {
        $this->beforeAPICall();
        switch($event->name) {
            case ActiveRecord::EVENT_BEFORE_INSERT:
               Api::call(['parameter2', 'parameter3']);
            // etc...
        }
        $this->afterAPICall();
    }

With my changes, code becomes cleaner:

    public function events()
    {
        return [
            ActiveRecord::EVENT_BEFORE_INSERT => [[$this, 'myMethod'], ['param1', 'param2']],
            ActiveRecord::EVENT_BEFORE_UPDATE => [[$this, 'myMethod'], ['param1']],
            ActiveRecord::EVENT_BEFORE_DELETE => [[$this, 'myMethod'], ['param3', 'param4']]
        ];
    }

    public function myMethod($event)
    {
        $this->beforeAPICall();
        Api::call($event->data);
        $this->afterAPICall();
    }

samdark:
OK. Would be great to have unit tests for new behavior.

klimov-paul:
I am against this change. It overcomplecates things.
It is better to use separated methods or `Closure`.

smsteel:
Actually, it is true, it can be done this way, didn't consider (But PHP 5.4+ due to $this usage):

    public function events()
    {
        return [
            ActiveRecord::EVENT_BEFORE_INSERT => function () { $this->myMethod(['param1']) }
            ActiveRecord::EVENT_BEFORE_UPDATE => function () { $this->myMethod(['param2', 'param3']) }
            ActiveRecord::EVENT_BEFORE_DELETE => function () { $this->myMethod(['param4']) }
        ];
    }

    public function myMethod($data)
    {
        $this->beforeAPICall();
        Api::call($event->data);
        $this->afterAPICall();
    }

But i still think it's expected behaviour, since `\yii\base\Component::on` accepts `$data`. At least, i expected it to be parsed from array. Maybe the structure of array must be changed to be more clear?

klimov-paul:
OK, I see your point - it make sense to consider at least.

klimov-paul:
It whould be better to use named array keys for thsi feature:
```php
public function events()
{
    return [
        ActiveRecord::EVENT_BEFORE_INSERT => ['handler' => 'myMethod', 'data' => ['param1', 'param2']],
    ];
}
```

This will also allow to provide an 'append' option, which is availbale at `Component::on()` as well.

creocoder:
Vote against. Looks like violates of standart workflow and smells like wrong refactoring wishes.

cebe:
> But PHP 5.4+ due to $this usage

Yii requires php 5.4 anyway.

I am also against adding this as it does not really add something useful, code does not look cleaner imo.
I am closing this PR now. Thanks for your contribution anyway!

Issue #9067, opened 2015-07-09
Email validator error on valid email.


PommeVerte:
Hey guys. 
I've been testing our app against possible email submissions (and some case insensitive checks but that's beyond the scope of this issue) and I noticed that some valid emails were being rejected by the email validator. The email set I used for testing come from the examples here: https://en.wikipedia.org/wiki/Email_address

Here are the emails that fail:

```
"much.more unusual"@example.com
admin@mailserver1
"very.unusual.@.unusual.com"@example.com
"very.(),:;<>[]\".VERY.\"very@\\ \"very\".unusual"@strange.example.com
"()<>[]:,;@\\\"!#$%&\'*+-/=?^_`{}| ~.a"@example.org
" "@example.org
üñîçøðé@example.com
üñîçøðé@üñîçøðé.com
```

cebe:
> admin@mailserver1
> "very.unusual.@.unusual.com"@example.com

these are okay to reject because the first is a local address, you do not want your webserver try to resolve an internal name and try to deliver mails there.
second is delivering via a different server @ in the mailbox part is hardly allowed by any MTA.
allowing these is a security issue.

about the `"..."` syntax, while they may be valid, a normal user would not even understand what "" means in his email address so I am quite sure there is no mail provider that allows email addresses like that.

> üñîçøðé@example.com
> üñîçøðé@üñîçøðé.com

these are the only two I would consider to be important. They should work.

Kolyunya:
> these are okay to reject because...

@cebe the point of the validator is to verify if the string meets the generic requirements to an abstract email address. It has nothing to do with host resolution, mail delivery or MTAs. What if we wand to develop an intranet web application and we want to validate local mailboxes? What is we use our custom MTA?

dynasource:
> to develop an intranet web application and we want to validate local mailboxes

this would show a usecase for adding an extra flag to EmailValidator in which local emails will also be accepted

SilverFire:
It's a popular thing to discuss :) 

> The point of the validator is to verify if the string meets the generic requirements to an abstract email address defined by RFC

The problem is that the most public email services does not allow to create fully RFC-compatible email address.

In case you validate email addresses according to RFC and have working website with at least 10k users, could you share statistics on how many users use emails that are rejected by EmailValidator?

SilverFire:
I've added tests to confirm that üñîçøðé addresses work fine.
[...]

PR #9201, opened 2015-07-24
fix for #9067 - EmailValidator error on valid email

andrewnester:
#9067: Email validator error on valid email

Regex in EmailValidator changed to accept unicode letter

### DIFF
--- a/framework/validators/EmailValidator.php
+++ b/framework/validators/EmailValidator.php
@@ -24,13 +24,30 @@ class EmailValidator extends Validator
      * @var string the regular expression used to validate the attribute value.
      * @see http://www.regular-expressions.info/email.html
      */
-    public $pattern = '/(*UTF8)^[\pL0-9!#$%&\'*+\\/=?^_`{|}~-]+(?:\.[\pL0-9!#$%&\'*+\\/=?^_`{|}~-]+)*@(?:[\pL0-9](?:[\pL0-9-]*[\pL0-9])?\.)+[\pL0-9](?:[\pL0-9-]*[\pL0-9])?$/';
+    public $pattern = '/^[a-zA-Z0-9!#$%&\'*+\\/=?^_`{|}~-]+(?:\.[a-zA-Z0-9!#$%&\'*+\\/=?^_`{|}~-]+)*@(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\.)+[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$/';
+
     /**
      * @var string the regular expression used to validate email addresses with the name part.
      * This property is used only when [[allowName]] is true.
      * @see allowName
      */
-    public $fullPattern = '/(*UTF8)^[^@]*<[\pL0-9!#$%&\'*+\\/=?^_`{|}~-]+(?:\.[\pL0-9!#$%&\'*+\\/=?^_`{|}~-]+)*@(?:[\pL0-9](?:[\pL0-9-]*[\pL0-9])?\.)+[\pL0-9](?:[\pL0-9-]*[\pL0-9])?>$/';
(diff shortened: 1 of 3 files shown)


framework/validators/EmailValidator.php
```diff
@@ -24,13 +24,13 @@ class EmailValidator extends Validator
      * @var string the regular expression used to validate the attribute value.
      * @see http://www.regular-expressions.info/email.html
      */
-    public $pattern = '/^[a-zA-Z0-9!#$%&\'*+\\/=?^_`{|}~-]+(?:\.[a-zA-Z0-9!#$%&\'*+\\/=?^_`{|}~-]+)*@(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\.)+[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?$/';
+    public $pattern = '/(*UTF8)^[\pL0-9!#$%&\'*+\\/=?^_`{|}~-]+(?:\.[\pL0-9!#$%&\'*+\\/=?^_`{|}~-]+)*@(?:[\pL0-9](?:[\pL0-9-]*[\pL0-9])?\.)+[\pL0-9](?:[\pL0-9-]*[\pL0-9])?$/';
```
samdark: Isn't it better to use `u` modifier at the end of the pattern instead of inline `(*UTF8)`??

tests/framework/validators/EmailValidatorTest.php
```diff
@@ -26,8 +26,10 @@ public function testValidateValue()
         $this->assertFalse($validator->validate('Carsten Brandt <mail@cebe.cc>'));
         $this->assertFalse($validator->validate('"Carsten Brandt" <mail@cebe.cc>'));
         $this->assertFalse($validator->validate('<mail@cebe.cc>'));
-        $this->assertFalse($validator->validate('info@örtliches.de'));
-        $this->assertFalse($validator->validate('sam@рмкреатиф.ru'));
+        $this->assertTrue($validator->validate('info@örtliches.de'));
+        $this->assertTrue($validator->validate('sam@рмкреатиф.ru'));
```
samdark: It's not a valid email according to http://sphinx.mythic-beasts.com/~pdw/cgi-bin/emailvalidate which is claimed to be RFC-compatible.

samdark: I guess that this is IDN stuff and should be validated with `enableIDN` set to `true`.

framework/validators/EmailValidator.php
andrewnester: @samdark yeah, you're right, we can use /u instead, (*UTF8) is just advice from PCRE http://www.pcre.org/pcre.txt
will fix it

tests/framework/validators/EmailValidatorTest.php
andrewnester: @samdark yeah, you're right, but info@örtliches.de according to validator is valid email. but anyway will move it to IDN stuff only, thanks

samdark: It's better to re-check RFC itself. It could be validator bug.

tests/framework/validators/EmailValidatorTest.php
```diff
@@ -22,36 +22,37 @@ public function testValidateValue()
...
         $this->assertTrue($validator->validate('sam@rmcreative.ru'));
         $this->assertTrue($validator->validate('5011@gmail.com'));
+
         $this->assertFalse($validator->validate('rmcreative.ru'));
         $this->assertFalse($validator->validate('Carsten Brandt <mail@cebe.cc>'));
         $this->assertFalse($validator->validate('"Carsten Brandt" <mail@cebe.cc>'));
         $this->assertFalse($validator->validate('<mail@cebe.cc>'));
         $this->assertFalse($validator->validate('info@örtliches.de'));
         $this->assertFalse($validator->validate('sam@рмкреатиф.ru'));
+        $this->assertFalse($validator->validate('üñîçøðé@example.com'));
```
samdark: It's valid email. Should be in `assertTrue`.

Issue #9059, opened 2015-07-09
ErrorHandling


dynasource:
I get a notice when displaying an error:

Notice: Undefined index: args in E:\xampp\htdocs\dmn\wc\6\dev\vendor\yiisoft\yii2\views\errorHandler\exception.php on line 373

The reason:
* when no $args are provided in the stacktrace, $arg is NULL instead of a array and it will show errors notices.

Solution:
Make $args=[] in https://github.com/yiisoft/yii2/blob/master/framework/web/ErrorHandler.php#L262

cebe:
can you send a PR with a fix?

dynasource:
I will

PR #9202, opened 2015-07-24
fix for #9059 - processed case when $args is not provided in stacktrace

andrewnester:
fix for #9059 
Processed case when $args is not provided in stacktrace
Processing done in the same way it's done for other ones parameters of exception handling


samdark:
Merged with some adjustments: https://github.com/yiisoft/yii2/commit/716057ed39f2ce3e28f84337ec09c70033d64d58. Thanks!

Issue #9027, opened 2015-07-07
An UploadedFile - early instantiation problem


daydiff:
If we use a few (more than one) inheritors of the ``UploadedFile`` in a controller than a ``getInstance()`` method called on any of inheritors returns the instance of inheritor that was used first in current php session.

Here an example:
```php
class ProductController extends \yii\web\Controller {
    public function actionUpdate($id)
    {
        $productImage = \app\components\ProductImage::getInstance($this, 'prod_image'); //returns an instance of ProductImage as expected
        $vendorImage = \app\components\VendorImage::getInstance($this, 'vendor_image'); //returns an instance of ProductImage - expected an instance of VendorImage!
    }
}
```

The problem is that ``yii\web\UploadedFile`` loads all files and instantiate its when fisrt call of ``getInstance()`` occurs using late static binding. So, when next calls of ``getInstance()`` occurs we getting a wrong instance from ``$_files`` static property.

mdmunir:
why you need to extend `yii\web\UploadedFile`?

daydiff:
Why not? It's not about my business logic. Btw, I need to implement some custom naming and manipulation for uploaded files and I don't want to put it into main model. I suppose that inheritor of ``yii\web\UploadedFile`` the best solution to do that.

Anyway, if ``yii\web\UploadedFile`` isn't declared as a final class, it's ``getInstance()`` method must return an instance of a class on which we call that method. It's logical and expected behavior.

klimov-paul:
Resolved by commit  2fe0a40

PR #9203, opened 2015-07-24
fix for #9027 - UploadedFile early instantiation problem

andrewnester:
fix for #9027 - UploadedFile early instantiation problem
To solve issue now class instance is created on demand when getInstance or getInstances called, not when loadFiles called

### DIFF
--- a/tests/framework/web/UploadedFileTest.php
+++ b/tests/framework/web/UploadedFileTest.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace yiiunit\framework\web;
+
+use yii\base\Model;
+use yiiunit\TestCase;
+use yii\web\UploadedFile;
+use yiiunit\framework\web\stubs\ProductImage;
+use yiiunit\framework\web\stubs\VendorImage;
+use yiiunit\framework\web\stubs\ImageModel;
+
+/**
+ * @group web
+ */
+class UploadedFileTest extends TestCase
+{
+
+    protected function setUp()
+    {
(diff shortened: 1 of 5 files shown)


tests/framework/web/UploadedFileTest.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+namespace yiiunit\framework\web;
+
+use yii\base\Model;
+use yiiunit\TestCase;
+use yii\web\UploadedFile;
+use yiiunit\framework\web\stubs\ProductImage;
+use yiiunit\framework\web\stubs\VendorImage;
+use yiiunit\framework\web\stubs\ImageModel;
```
klimov-paul: Why 2 stub classes?
Was not single one enough?

tests/framework/web/stubs/ImageModel.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace yiiunit\framework\web\stubs;
+
+use yii\base\Model;
+
+class ImageModel extends Model
```
klimov-paul: There is `yiiunit\framework\web\stubs\ModelStub` already.
There is no reason to add extra class here.

klimov-paul:
Overall, looks fine.
@andrewnester, please remove redundant code from the unit tests, pull from upstream add CHANGELOG file.

klimov-paul:
Migrated to #12439

PR #9205, opened 2015-07-24
fixes #9006

### LINKED ISSUE #9006: MySQL BIT column always returned as true
Table structure:
```sql
--
-- Table structure for table `demo`
--

CREATE TABLE IF NOT EXISTS `demo` (
`id` bigint(20) unsigned NOT NULL,
  `test` bit(1) NOT NULL
) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8;

--
-- Dumping data for table `demo`
--

INSERT INTO `demo` (`id`, `test`) VALUES
(1, b'0'),
(2, b'1');
```

Test code using auto-generated model:
```php
$all = Demo::find()->all();
foreach ($all as $current) {
    var_dump($current->test);
}
```
Results are TRUE 2 times.

Fetching as array:
```php
$all = Demo::find()->asArray()->all();
foreach($all as $current) {
    if ($current['test'] === chr(0x00))
        echo 0 . PHP_EOL;
    elseif ($current['test'] === chr(0x01))
        echo 1 . PHP_EOL;
}
```
returns 0 and 1.

stratoss:
bypassing PHP's (bool) which sees 0x01 and 0x00 both as true.

fixes #9006

### DIFF
--- a/framework/db/ColumnSchema.php
+++ b/framework/db/ColumnSchema.php
@@ -133,7 +133,8 @@ class ColumnSchema extends Object
             case 'integer':
                 return (int) $value;
             case 'boolean':
-                return (bool) $value;
+		// bypassing PHP's (bool) which considers both 0x00 and 0x01 as true
+                return ($value === chr(0x01) || $value === chr(0x31));
             case 'double':
                 return (double) $value;
         }


samdark:
Would it work for non-MySQL?

stratoss:
I can't see a reason why it wouldn't.

samdark:
https://github.com/yiisoft/yii2/pull/9264

PR #9206, opened 2015-07-24
Enhancement #9198 - Add param buttonsVisible in ActionColumn

### LINKED ISSUE #6162: Please add a way to specify visibility of buttons in ActionColumn
A very common use case of overriding the `buttons` property is to hide buttons under certain conditions (eg, the user does not have permission to do that). Overriding the buttons property is very annoying because I then need to know how to build that button, including markup.

It would be much easier to just specify that the `delete` button should not be shown on this request. On Yii1 we could do that by passing a 'visible' key in the button specification. Please implement something similar for Yii2.

Thanks for developing Yii!

### LINKED ISSUE #9198: [Feature] Visible of buttons in ActionColumn
Hi. My english is very bad, i'm sorry.

I want to manage visible of buttons in ActionColumn (GridView). In current version we have the next solution: https://github.com/yiisoft/yii2/blob/5c79ab202cce74048957dce586d7ef2796908138/framework/grid/ActionColumn.php#L70

But i want set only boolean value for this. Example:

```php
'buttonsVisible' => [
      'update' => \Yii::$app->user->can('update'),
      'delete' => function($model){
            return \Yii::$app->user->can('delete') && $model->status != Status::DELETED
       },
 ]
```

That is, I want to separate setting html-content of buttons in ActionColumn and setting visibility of this buttons.
To do this, we can add the parameter buttonsVisible to yii\grid\ActionColumn and modify of method renderDataCellContent:

```php
namespace yii\grid;
class ActionColumn 
{
    public $buttonsVisible = [];


    protected function renderDataCellContent($model, $key, $index)
    {
        return preg_replace_callback('/\\{([\w\-\/]+)\\}/', function ($matches) use ($model, $key, $index) {
            $name = $matches[1];

            if (isset($this->buttonsVisible[$name])) {
                $visible = is_callable($this->buttonsVisible[$name])
                    ? call_user_func($this->buttonsVisible[$name], $model)
                    : $this->buttonsVisible[$name];
            } else {
                $visible = true;
            }

            if ($visible && isset($this->buttons[$name])) {
                $url = $this->createUrl($name, $model, $key, $index);
                return call_user_func($this->buttons[$name], $url, $model, $key);
            } else {
                return '';
            }
        }, $this->template);
    }   
}
```

I can prepare a pull request.
What do you think?

fornit1917:
Closes #9198
Closes #6162


framework/grid/ActionColumn.php
```diff
@@ -174,9 +196,17 @@ protected function renderDataCellContent($model, $key, $index)
     {
         return preg_replace_callback('/\\{([\w\-\/]+)\\}/', function ($matches) use ($model, $key, $index) {
             $name = $matches[1];
-            if (isset($this->buttons[$name])) {
-                $url = $this->createUrl($name, $model, $key, $index);
 
+            if (isset($this->buttonsVisible[$name])) {
+                $visible = is_callable($this->buttonsVisible[$name])
```
SilverFire: Should be replaced to `$this->buttonsVisible[$name] instanceof \Closure`

framework/grid/ActionColumn.php
```diff
@@ -174,9 +196,17 @@ protected function renderDataCellContent($model, $key, $index)
...
+                    ? call_user_func($this->buttonsVisible[$name], $model)
```
SilverFire: I'm suggesting to pass `$key` and `$index` too.

framework/grid/ActionColumn.php
```diff
@@ -85,6 +85,28 @@ class ActionColumn extends Column
      * ```
      */
     public $buttons = [];
+    /** @var array visibility conditions for each button. The array keys are the button names (without curly brackets),
+     * and the values are the conditions whether this button is visible.Can be boolean value or callback. Defaults to true.
+     * The callbacks should use the following signature:
```
SilverFire: should -> must

framework/grid/ActionColumn.php
```diff
@@ -85,6 +85,28 @@ class ActionColumn extends Column
      * ```
      */
     public $buttons = [];
+    /** @var array visibility conditions for each button. The array keys are the button names (without curly brackets),
+     * and the values are the conditions whether this button is visible.Can be boolean value or callback. Defaults to true.
```
SilverFire: I'm suggesting the following wording. Do you like it?

`are the conditions whether this button is visible. Can be boolean value or callback. Defaults to true.` -> `are the boolean true/false or the anonymous function. The button will be shown, when its name is not specified in this array`

framework/grid/ActionColumn.php
```diff
@@ -85,6 +85,28 @@ class ActionColumn extends Column
...
+     *
+     * Or you can pass a boolean expression:
+     *
+     * ```php
+     * [
+     *     'update' => \Yii::$app->user->can('update'),
+     * ],
+     * ```
+     */
+    public $buttonsVisible = [];
```
SilverFire: I don't really like this name because it breaks rules about words order in English. Let's rename this property to `visibleButtons` instead

framework/grid/ActionColumn.php
SilverFire: `$visible` -> `$isVisible`

SilverFire:
Moved from https://github.com/yiisoft/yii2/issues/9198
Originally authored by @fornit1917 

___

Hello. Thank you for review. I agree with your comments. I'll fix all in the near future.

framework/grid/ActionColumn.php
```diff
@@ -85,6 +85,28 @@ class ActionColumn extends Column
      * ```
      */
     public $buttons = [];
+    /** @var array visibility conditions for each button. The array keys are the button names (without curly brackets),
+     * and the values are are the boolean true/false or the anonymous function. The button will be shown,
```
SilverFire: `are are` -> `are`

PR #9207, opened 2015-07-24
Added 301 redirect for URLs not matching suffix

### LINKED ISSUE #7670: UrlRule Problem with trailing slash (/)
```
```

If i then call a url like 
```

```

```
```

Regards Horizons

samdark:
Should fix #7670. Not sure about calling `response` directly.

### DIFF
--- a/framework/web/UrlRule.php
+++ b/framework/web/UrlRule.php
@@ -217,6 +217,21 @@ class UrlRule extends Object implements UrlRuleInterface
 
         $pathInfo = $request->getPathInfo();
         $suffix = (string) ($this->suffix === null ? $manager->suffix : $this->suffix);
+
+        if (!empty($pathInfo)) {
+            if (substr_compare($pathInfo, '/', -1, 1) === 0) {
+                if ($suffix === '') {
+                    Yii::$app->response->redirect('/' . rtrim($pathInfo, '/'), 301);
+                    return false;
+                }
+            } else {
+                if ($suffix === '/') {
+                    Yii::$app->response->redirect('/' . $pathInfo . '/', 301);
+                    return false;
+                }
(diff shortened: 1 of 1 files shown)


klimov-paul:
Should not this be done in `UrlManager` instead of `UrlRule`?

samdark:
Probably yes.

samdark:
But then implementation becomes a bit more complicated.

framework/web/UrlRule.php
```diff
@@ -217,6 +217,21 @@ public function parseRequest($manager, $request)
 
         $pathInfo = $request->getPathInfo();
         $suffix = (string) ($this->suffix === null ? $manager->suffix : $this->suffix);
+
+        if (!empty($pathInfo)) {
+            if (substr_compare($pathInfo, '/', -1, 1) === 0) {
+                if ($suffix === '') {
```
klimov-paul: I am afraid this is not aceeptable: `$suffx` may change from one URL rule to another, but this code will do redirect using suffix from the first rule in the list no matter if it matches URL or not.
This may as well cause the infinite redirection loop if 2 rules in the list use different suffixes.

klimov-paul:
May be this functionality should be even moved to `yii\web\Request::resolve()` and made optional.

framework/web/UrlRule.php
samdark: Does it make sense to have separate URL rules for `http://exampe.com/contact` and `http://exampe.com/contact/`?

samdark:
I'm against making it optional (or at least making it off by default) since the issue is common and quite hurts website's SEO.

samdark:
Closed in favor of #7670

PR #9210, opened 2015-07-25
[fix-tests-config-infinite-loop] Ensuring file config.local.php is no…

felixmaier1989:
I'm trying to set up my yii2 dev environment, following get started guidelines. I went through an infinite loop in tests/data/config.local.php which my patch would solve. Additionally I replaced '/' with DIRECTORY_SEPARATOR to be cross-OS.

### DIFF
--- a/tests/data/config.php
+++ b/tests/data/config.php
@@ -46,8 +46,9 @@ $config = [
     ],
 ];
 
-if (is_file(__DIR__ . '/config.local.php')) {
-    include(__DIR__ . '/config.local.php');
+if (is_file(__DIR__ . DIRECTORY_SEPARATOR  . 'config.local.php') &&
+    __FILE__ !== __DIR__ . DIRECTORY_SEPARATOR  . 'config.local.php') {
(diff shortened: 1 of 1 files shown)


cebe:
`/` is already cross OS, DIRECTORY_SEPARATOR is only needed if you do comparison.

I do not see how you could run into an infinite loop here, did you copy the complete code of the config.php file? this is not how it works, you should only use that file to override things in the already existing `$config` variable.

felixmaier1989:
Indeed I copied the whole config.php file (as it is) to config.local.php.
Then config.php checks for config.local.php and includes it
config.local.php checks for config.local.php and includes it ==> infinitely 

I agree that leaving the lines where config.local.php is included is not very smart. But with the condition I added to the if, you dont even have to removed these lines. You just have to copy the file as said in the get started documentation.

cebe:
the instructions on how to use this file are very clear: https://github.com/yiisoft/yii2/blob/708c0d22e515cfc118cd5b861541d233c445f632/tests/data/config.php#L7-L13 your change is not needed.

felixmaier1989:
Indeed, I didnt pay enough attention.

PR #9212, opened 2015-07-25
No need to echo db exception

usecomposable:


samdark:
Why?

df2:
$exception contains a raw exception message from the database server, and that exception also seems to be handled fine like any other exception: 

Extra info in error page
See the text printed here underneath the standard error screen (in prod. mode).

cebe:
how did you get this output? looks a bit like yii 1.1 error page...

df2:
When no errorAction is set in the errorHandler config, the following view file is rendered: yiisoft/yii2/views/errorHandler/error.php. That's the one in the pic. 

BTW, a full stacktrace is shown in debug mode, with a longer version of the exception being echoed at the end of the page (because of echo $exception).

samdark:
echo was added in https://github.com/yiisoft/yii2/issues/1946

samdark:
I've verified that in current code it seems to work just fine w/o echo. @cebe any comments?

kidol:
The echo part is intended for `writeSession()` only. If some database exception happens before `writeSession()` (and the exception is handled correctly), then `writeSession()` will be called anyhow because of `register_shutdown_function([$this, 'close']);` in `Session::init()`.

This works correctly though:

```php
Yii::$app->session->get('test');
sleep(5); // shutdown db server here
Yii::$app->session->set('test', 123);
```

samdark:
If there's no DB table it works correcty as well i.e. exception is handled properly.

kidol:
@samdark Yes it's handled, but not in `writeSession()`. You must test db exception that's thrown in `writeSession()` to see the need for `echo()`.

Use this:

```php
Yii::$app->session->get('test');
sleep(5); // shutdown db server here
Yii::$app->session->set('test', 123); // will cause exception in writeSession
exit;
```
This outputs the exception as string without rendering any html.

And if you do

```php
// Shutdown db server here
Yii::$app->session->open();
exit;
```
You get rendered html + the string at bottom.

Dirty fix could be:

```php
// Only echo and log if we haven't processed an exception yet
if (Yii::$app->errorHandler->exception === null) {
   $exception = ErrorHandler::convertExceptionToString($e);
   // its too late to use Yii logging here
   error_log($exception);
   echo $exception;
}
```

@cebe Is there really no way to properly handle exceptions thrown in `writeSession()`? For example, looking at symfony they seem to throw: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/PdoSessionHandler.php#L377

cebe:
> @cebe Is there really no way to properly handle exceptions thrown in writeSession()? For example, looking at symfony they seem to throw: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpFoundation/Session/Storage/Handler/PdoSessionHandler.php#L377

see http://us3.php.net/manual/en/function.session-set-save-handler.php#refsect1-function.session-set-save-handler-notes

however this issue has been worked around by https://github.com/yiisoft/yii2/commit/fa845ee481650a26d304017065ea2fa985f59ccd already. Closing this PR.

cebe:
related to  #11726

PR #9216, opened 2015-07-25
improved documentation of Model::load() return value

### LINKED ISSUE #8803: base\Model::load() return documentation is misleading

cebe:
tried to make it more clear. @tom-- are you happy with this version?

fixes #8803

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -785,10 +785,12 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      * If [[formName()]] is empty, the whole `$data` array will be used to populate the model.
      * The data being populated is subject to the safety check by [[setAttributes()]].
      * @param array $data the data array. This is usually `$_POST` or `$_GET`, but can also be any valid array
-     * supplied by end user.
+     * supplied by the end user.
      * @param string $formName the form name to be used for loading the data into the model.
      * If not set, [[formName()]] will be used.
-     * @return boolean whether the model is successfully populated with some data.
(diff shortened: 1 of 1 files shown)


framework/base/Model.php
```diff
@@ -785,10 +785,12 @@ public function activeAttributes()
...
      * The data being populated is subject to the safety check by [[setAttributes()]].
      * @param array $data the data array. This is usually `$_POST` or `$_GET`, but can also be any valid array
-     * supplied by end user.
+     * supplied by the end user.
      * @param string $formName the form name to be used for loading the data into the model.
      * If not set, [[formName()]] will be used.
-     * @return boolean whether the model is successfully populated with some data.
+     * @return boolean whether `$data` contained some data to populate the models attributes.
+     * This does only mean that there was some data provided. It does not necessarily mean that any field
+     * was populated, which may be the case if none of the provided fields is safe in the current [[scenario]].
```
tom--: It's good. But to make it more typical English, change "This does only mean that" to "This only means that"

cebe: merged with adjustment. thanks!

PR #9224, opened 2015-07-27
Update start-gii.md

et-soft:
Update example for config Gii module.


samdark:
Duplicate of #9223

et-soft:
This is not duplicate. In #9223 version for Polish version. In this commit, for Russian. Or may be sync for other version create automatic?

samdark:
Thanks!

PR #9239, opened 2015-07-28
Handle json_encode errors

grzegorzkurtyka:
When encoding response to JSON format, if it contains malformed UTF-8 strings error is unhandled and results in empty response.


samdark:
Merged with some enhancements. Thanks!

PR #9246, opened 2015-07-29
Change private member variables to protected in class UrlRule

usecomposable:
By changing the private member variables of class UrlRule to protected, they may be accessed by sub-classes of UrlRule (which in turn are specified by 'ruleConfig' => ['class' => 'app\src\components\UrlRule'], in UrlManager config).

$_paramRules is especially important to be able to determine expected parameters to a route.

### DIFF
--- a/framework/web/UrlRule.php
+++ b/framework/web/UrlRule.php
@@ -91,19 +91,19 @@ class UrlRule extends Object implements UrlRuleInterface
     /**
      * @var string the template for generating a new URL. This is derived from [[pattern]] and is used in generating URL.
      */
-    protected $_template;
+    private $_template;
     /**
      * @var string the regex for matching the route part. This is used in generating URL.
      */
-    protected $_routeRule;
+    private $_routeRule;
     /**
      * @var array list of regex for matching parameters. This is used in generating URL.
      */
-    protected $_paramRules = [];
+    private $_paramRules = [];
     /**
      * @var array list of parameters used in the route.
      */
-    protected $_routeParams = [];
+    private $_routeParams = [];
 
 
     /**
@@ -340,5 +340,14 @@ class UrlRule extends Object implements UrlRuleInterface
         }
 
         return $url;
+    }
+
+    /**
+     * Allows read access to this rule's parameter keys and regexp rules.
+     * @return array parameter keys and regexp rules.
+     */
+    public function getParamRules() 
+    {
+        return $this->_paramRules;
     }
 }
--- a/framework/web/UrlRule.php
+++ b/framework/web/UrlRule.php
@@ -91,19 +91,19 @@ class UrlRule extends Object implements UrlRuleInterface
     /**
      * @var string the template for generating a new URL. This is derived from [[pattern]] and is used in generating URL.
      */
-    private $_template;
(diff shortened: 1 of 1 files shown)


samdark:
Can you show an example of when/where you want to use it?

df2:
When I do like this:

```php
<?php

namespace app\src\components;

use yii\web\UrlRule as YiiUrlRule;

class UrlRule extends YiiUrlRule {

    public function getParamRules() {

        return $this->_paramRules;
    }
}
```

and combination with $controller->actionParams I can then override Url::to() to inject parameters automatically that are in the current URL so I don't have to type them each time.

lynicidn:
it bad idea - template and pattern do init https://github.com/df2/yii2/commit/07ebdab7ed53faf364b884eea28f0d28a37f7fcc#diff-74b1030c748d9da559378bf10fba0078R138

df2:
I'm not sure I understand what you mean about template and pattern doing init... how else would you suggest determining the required url parameters for a given route?

lynicidn:
u get unexpected value after init

df2:
Bud, we're talking about changing some private member variables to protected so that sub-classes of UrlRule can access them.

samdark:
Is it something global like language in URL? If yes, it should not be put into rule but rather implemented on the manager level like https://github.com/codemix/yii2-localeurls/blob/master/UrlManager.php

df2:
Imagine more like hierarchical data structures. A car has a wheel has a tire.

```
<carId> => car/index
<carId>/<wheelId> => wheel/index
<carId>/<wheelId>/<tireId> => tire/index
```

Imagine you're at 11/22/33, so tire controller, index action. Say you render a view and make a link like so <?= Url::to(['car/index']); ?>. Without a carId specified, 11 will be used automatically.

samdark:
Why not to use Url helper then?

```php
<?= Url::current(['tireId' => null]);
```

df2:
That won't help linking to other controllers. Also, the fix is about enabling determination of required params to a specified route. We do this by cycling through each compiled url rule until we match on route. Then we cannot go to more depths of knowledge because the member variable $_paramRules is private instead of protected.

df2:
Also, then we're back to specifying parameters by hand, which we like to avoid :-)

df2:
As far as I can make out, the alternatives would be: 
* to add a public getParamRules method to yii\web\UrlRule, 
* or to copy the entire yii\web\UrlRule class and use that instead of sub-classing it.

Seems a bit restrictive to not be able to sub-class and still access the rule's route parameters, doesn't it?

samdark:
Yes.

df2:
Great, then it needs to be decided if the proposed changes can be merged.

samdark:
Personally I prefer getters for this case.

One more question. Have you tried implementing your use case already? Works?

df2:
Sounds fine. Should there be a getter for each private variable or just $_paramRules? Should it go above or below the existing methods. Will this affect test coverage? 

Yes, use case confirmed working.

samdark:
Merged with adjustments. Thanks!

df2:
You bet! Thanks!

PR #7260, opened 2015-02-12
Added `updateMessages()` to `yii.activeForm.js` to support manually upda...

nkovacs:
...ting ActiveForm messages.

Fixes #5991, #5420, #3017

PR #9249, opened 2015-07-29
Adding callback for custom hash generation in AssetManager.

### LINKED ISSUE #9245: Refactoring `AssetManager` hash function.
In my cases, I deployed a lot of identical Yii2 app on instances behind a load balancer. And I usually set the load balancer to not use session stickiness, meaning the following requests can be served by different instances.

The problem is in the hashed directory generated by `AssetManager` that uses `filemtime`. Since the deployment in instances can be in different seconds, the hashed directory for an asset can be different for each instances.

So I extended my own `AssetManager`, which means I have to modify all `getPublishedUrl`, `publishFile`, `publishDirectory`, `getPublishedPath`, which pretty much have a lot of similar codes like `$this->hash(dirname($path) . filemtime($path))`.

- https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L443
- https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L488
- https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L536
```
 if (is_file($path)) {
    return $base . $this->hash(dirname($path) . filemtime($path)) . DIRECTORY_SEPARATOR . basename($path);
} else {
    return $base . $this->hash($path . filemtime($path));
}
```
- https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L561
```
 if (is_file($path)) {
    return $this->baseUrl . '/' . $this->hash(dirname($path) . filemtime($path)) . '/' . basename($path);
} else {
    return $this->baseUrl . '/' . $this->hash($path . filemtime($path));
}
```

I'm just thinking that maybe the `hash` method and the methods using it can be refactored and the `filemtime` can be put into the refactored method where the developers can opt to use their own hash generation strategy. The strategy can maybe change the length of the hash string or maybe a certain salt that is generated every time the app is built.

petrabarus:
Fixes #9245

### DIFF
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -168,6 +168,31 @@ class AssetManager extends Component
      * @since 2.0.3
      */
     public $appendTimestamp = false;
+    /**
+     * @var callable a callback that will be called to produce hash for asset directory generation.
+     * The signature of the callback should be as follows:
+     *
+     * ```
+     * function ($path)
+     * ```
+     *
+     * where `$path` is the asset path. Note that the `$path` can be either directory where the asset
+     * files reside or a single file. For a CSS file that uses relative path in `url()`, the hash
+     * implementation should use the directory path of the file instead of the file path to include 
+     * the relative asset files in the copying.
+     *
+     * If this is not set, the asset manager will use the default CRC32 and filemtime in the `hash`
+     * method.
+     * 
+     * Example of an implementation using MD4 hash:
+     *
+     * ```php
+     * function ($path) {
+     *     return hash('md4', $path);
+     * }
+     * ```
+     */
+    public $hashCallback;
 
     private $_dummyBundles = [];
 
@@ -440,7 +465,7 @@ class AssetManager extends Component
      */
     protected function publishFile($src)
     {
-        $dir = $this->hash(dirname($src) . filemtime($src));
+        $dir = $this->hash($src);
         $fileName = basename($src);
         $dstDir = $this->basePath . DIRECTORY_SEPARATOR . $dir;
         $dstFile = $dstDir . DIRECTORY_SEPARATOR . $fileName;
@@ -485,7 +510,7 @@ class AssetManager extends Component
      */
     protected function publishDirectory($src, $options)
     {
-        $dir = $this->hash($src . filemtime($src));
+        $dir = $this->hash($src);
         $dstDir = $this->basePath . DIRECTORY_SEPARATOR . $dir;
         if ($this->linkAssets) {
             if (!is_dir($dstDir)) {
@@ -532,12 +557,7 @@ class AssetManager extends Component
             return $this->_published[$path][0];
         }
         if (is_string($path) && ($path = realpath($path)) !== false) {
-            $base = $this->basePath . DIRECTORY_SEPARATOR;
-            if (is_file($path)) {
-                return $base . $this->hash(dirname($path) . filemtime($path)) . DIRECTORY_SEPARATOR . basename($path);
-            } else {
-                return $base . $this->hash($path . filemtime($path));
-            }
+            return $this->basePath . DIRECTORY_SEPARATOR . $this->hash($path) . (is_file($path) ? DIRECTORY_SEPARATOR . basename($path) : '');
         } else {
             return false;
         }
@@ -558,11 +578,7 @@ class AssetManager extends Component
             return $this->_published[$path][1];
         }
         if (is_string($path) && ($path = realpath($path)) !== false) {
-            if (is_file($path)) {
-                return $this->baseUrl . '/' . $this->hash(dirname($path) . filemtime($path)) . '/' . basename($path);
-            } else {
-                return $this->baseUrl . '/' . $this->hash($path . filemtime($path));
-            }
+            return $this->baseUrl . '/' . $this->hash($path) . (is_file($path) ? '/' . basename($path) : '');
         } else {
             return false;
         }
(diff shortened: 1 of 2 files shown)


framework/web/AssetManager.php
```diff
@@ -534,9 +556,9 @@ public function getPublishedPath($path)
         if (is_string($path) && ($path = realpath($path)) !== false) {
             $base = $this->basePath . DIRECTORY_SEPARATOR;
             if (is_file($path)) {
```
petrabarus: Wondering if there is a coding guideline regarding ternary operator.

```
return $base . (is_file($path) ? 
                $this->hash($path) . DIRECTORY_SEPARATOR . basename($path) : 
                $this->hash($path, false));
```

framework/web/AssetManager.php
```diff
@@ -572,10 +594,15 @@ public function getPublishedUrl($path)
      * Generate a CRC32 hash for the directory path. Collisions are higher
      * than MD5 but generates a much smaller hash string.
      * @param string $path string to be hashed.
+     * @param boolean $isFile whether the path is file.
      * @return string hashed string.
      */
-    protected function hash($path)
+    protected function hash($path, $isFile = true)
```
samdark: That's a problem. Because of change of the signature we're breaking backwards compatibility.

As far as I understand, we can determine is the path is a file by the presence of file extension, can't we?

framework/web/AssetManager.php
samdark: Probably not. At least it's not in any of PSRs.

framework/web/AssetManager.php
petrabarus: Agree, at first I was going to do make

```
    protected function hash($path)
    {
        if (!empty($this->hashCallback)) {
            return call_user_func($this->hashCallback, $path, $isFile);
        }
        $path = (is_file($path) ? dirname($path) : $path) . filemtime($path);
        return sprintf('%x', crc32($path . Yii::getVersion()));
    }
```

But then, there will be a redundant `is_file($path)` calls. If you want to trade that off with function signature, I'm totally okay.

samdark: `is_file` is a stat operation. I'd prefer something like `pathinfo($path, PATHINFO_EXTENSION) === ''` which won't touch FS.

petrabarus: Yes, in case of asset files which mostly have extensions like `css`, `jpg`, `js` this makes sense. Going to update this.

samdark:
@yiisoft/core-developers need additional review.

framework/web/AssetManager.php
```diff
@@ -169,6 +169,27 @@ class AssetManager extends Component
      */
     public $appendTimestamp = false;
```
klimov-paul: There should be no extra line: both fields are public.

framework/web/AssetManager.php
```diff
@@ -532,12 +553,8 @@ public function getPublishedPath($path)
             return $this->_published[$path][0];
         }
         if (is_string($path) && ($path = realpath($path)) !== false) {
-            $base = $this->basePath . DIRECTORY_SEPARATOR;
-            if (is_file($path)) {
-                return $base . $this->hash(dirname($path) . filemtime($path)) . DIRECTORY_SEPARATOR . basename($path);
-            } else {
-                return $base . $this->hash($path . filemtime($path));
-            }
+            return $this->basePath . DIRECTORY_SEPARATOR . $this->hash($path) .
```
klimov-paul: I am usure splitting in 2 lines makes sense here: it would be better to write in single one - it whould be long then former line number 537.

klimov-paul:
CHANGELOG line should be added.

petrabarus:
@klimov-paul done. please check.

framework/web/AssetManager.php
```diff
@@ -576,6 +587,10 @@ public function getPublishedUrl($path)
      */
     protected function hash($path)
     {
+        if (!empty($this->hashCallback)) {
+            return call_user_func($this->hashCallback, $path);
+        }
+        $path = (pathinfo($path, PATHINFO_EXTENSION) === '' ? $path : dirname($path)) . filemtime($path);
```
petrabarus: @samdark @klimov-paul 

I'm wondering if we should just pass this `$path = (pathinfo($path, PATHINFO_EXTENSION) === '' ? $path : dirname($path))` to the `hashCallback` or let the users recheck the `$path`?

klimov-paul: I suppose current implementation is fine: particular hash callback may not need such logic. Still perhaps note about `$path` can lead to directory or file should be added to the doc commets at `hashCallback` declaration.

klimov-paul:
This one looks good.

samdark:
Merged. Thanks!

nkovacs:
Re what I said on #7260:
My extension was passing just a filemtime to hash, so now it breaks (though the time is not a valid path, so filemtime and is_file in hash errors out). Arguably, I was using it wrong, since the timestamp returned by filemtime was not a hash.
So I don't think this should be reverted or anything like that. But hash is a protected function, and now it gets a different parameter (just `$path` instead of `$path . filemtime($path)` or `dirname($path).filemtime($path)`), and it's stricter about what it accepts.

framework/web/AssetManager.php
```diff
@@ -576,6 +592,10 @@ public function getPublishedUrl($path)
      */
     protected function hash($path)
     {
+        if (!empty($this->hashCallback)) {
```
creocoder: @klimov-paul , @samdark 

This should be `$this->hashCallback !== null` since `empty()` make sense with arrays and in some very rare situations for other types. Whole framework at least parts which written by @qiangxue used this convention. Why on review stage other core developers did not see such things?

samdark: https://github.com/yiisoft/yii2/commit/372211a865b3832c4d27b9377ebe27039d4d13e2

samdark: We can't see everything. Help is welcome.

petrabarus: Thanks for the correction! I'll keep that in mind for future PRs.

PR #9253, opened 2015-07-29
PageCache&FragmentCache property $variations forced to be a array

schojniak:
Before, if somebody made a mistake, for example:

`'variations' => Yii::$app->language` instead of `'variations' => [Yii::$app->language]`

then variations wouldn't work without any warning.


developeruz:
Replaced with #15385

PR #9256, opened 2015-07-29
simplified foreach in Model::scenarios()

schojniak:
it's ok to call `array_keys([])`


samdark:
I guess `empty()` is a much cheaper call.

schojniak:
```if``` + ```empty``` vs unlikely empty scenario

schojniak:
auto-generated scenario is empty only if every attribute is on "except" list

you're right ```empty([])``` is about 8x faster than ```array_keys([])```

PR #9258, opened 2015-07-29
fixed message with missing keyword

dima-gusyatiner:
Fixed error:

Formatting message for language 'he-IL' failed with error: Message pattern is invalid: msgfmt_create: message formatter creation failed: U_DEFAULT_KEYWORD_MISSING. The message being formatted was: {attribute} אמור לא לכלול יותר מ {max, number} {max, plural, one{תו} few{תוים} many{תוים}}..


samdark:
Thanks!

PR #9259, opened 2015-07-29
Added validation of scopes in HttpBearerAuth

shcherbanich:
The use of Scope in an OAuth2 application is often key to proper permissioning. Scope is used to limit the authorization granted to the client by the resource owner.

For example:

``` php
    'class' => CompositeAuth::className(),
    'authMethods' => [
        ['class' => HttpBearerAuth::className(), 'scopes' => 'wallet,music,video']
    ]
];
```

This code adds validation for scopes of accessToken

``` php
/**
 * Finds user by access token
 *
 * @param string $token
 * @param mixed $type
 * @param mixed $scopes
 *
 * @return static|null
 */
public static function findIdentityByAccessToken($token, $type = null, $scopes = null);
```

&#10071;**This code breaks backward compatibility, because changes the interface. Interface change yet can be canceled**&#10071;

### DIFF
--- a/framework/filters/auth/HttpBearerAuth.php
+++ b/framework/filters/auth/HttpBearerAuth.php
@@ -33,6 +33,10 @@ class HttpBearerAuth extends AuthMethod
      */
     public $realm = 'api';
 
+    /**
+     * @var string|null available scopes
+     */
+    public $scopes = null;
 
     /**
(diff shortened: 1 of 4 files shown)


samdark:
Won't be applied to 2.0.

PR #9263, opened 2015-07-30
Increase rbac security: Prevent getting the first item when name is empty.

rhertogh:
When an empty string or null value is passed by accident to the DbManager unintentional access might be granted if the user has permission on the first item in the table.


samdark:
The check makes sense in order to avoid useless SQL but I don't think current code has any issue besides that.  It generats either `SELECT * FROM auth_item WHERE name IS NULL` or `SELECT * FROM auth_item WHERE name = ''`. Both will return `null` because there are no unnamed items.

PR #9264, opened 2015-07-30
Fix boolean check

mj4444ru:
Correction of checking the value of the type of bit received from mysql

### DIFF
--- a/framework/db/ColumnSchema.php
+++ b/framework/db/ColumnSchema.php
@@ -133,7 +133,7 @@ class ColumnSchema extends Object
             case 'integer':
                 return (int) $value;
             case 'boolean':
-                return (bool) $value;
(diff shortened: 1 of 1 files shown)


samdark:
@mj4444ru are you trying to fix #9006?

mj4444ru:
@samdark Yes, there is described the same problem

samdark:
Do you have reliable way to reproduce it? A unit test maybe?

mj4444ru:
@samdark I have installed OS Ubuntu and this problem is present

Faryshta:
I think you should check http://www.yiiframework.com/doc-2.0/yii-db-columnschema.html#phpTypecast()-detail and make the patch there. This will work for the format check, anywhere else the problem is still there.

PR #9283, opened 2015-08-02
do not use reserved php words as function in schemabuilder

cebe:
there is a reason for them to be reserved, we should not add such workarounds to still use them.
- [x] adjust tests

### DIFF
--- a/framework/db/SchemaBuilder.php
+++ b/framework/db/SchemaBuilder.php
@@ -11,14 +11,14 @@ use Yii;
 use yii\base\Object;
 
 /**
- * SchemaBuilder is the class help to define DB's schema types.
+ * SchemaBuilder helps to define database schema types using a PHP interface.
  *
  * For example you may use the following code inside your migration files:
  *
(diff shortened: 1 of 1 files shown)


samdark:
I disagree. It's very convenient.

cebe:
what is convenient? using reserved words as function names should be avoided, even if it is possible using some hack with __call().

nineinchnick:
It matches the SQL keyword, making all methods naming more consistent. But I agree with CeBe, you could memorise this as an exception quickly and IDE autocompletion would help too.

samdark:
Well, it's not a big issue for me...

cebe:
I agree that `default` is the best name if we were free to use it as a normal method name. As it is a reserved word in php I still would not use it and choose the next best name which is `defaultValue` imo.

samdark:
OK. Let's merge it.

cebe:
k, will do it in a separate commit by also adjusting the tests.

Issue #9282, opened 2015-08-02
PHP 5.5 adds new JSON error codes


cebe:
- JSON_ERROR_RECURSION 	One or more recursive references in the value to be encoded
- JSON_ERROR_INF_OR_NAN 	One or more NAN or INF values in the value to be encoded
- JSON_ERROR_UNSUPPORTED_TYPE 	A value of a type that cannot be encoded was given

http://php.net/manual/en/function.json-last-error.php

should be handled in https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseJson.php#L89

freezy-sk:
I don't like hardcoded constant values in #9289 so I was trying to make `handleJsonError` more compatible with differences between **PHP 5.4** and **5.5** and I've found this:
**PHP 5.4** is throwing `E_WARNING json_encode(): type is unsupported, encoded as null` when you are trying to encode Resource like this
```php
$fp = fopen('php://stdin', 'r');
Json::encode(['a' => $fp]);
```
and result is `{"a":null}` with `json_last_error() JSON_ERROR_NONE`

**PHP 5.5** is result `false` with `json_last_error() JSON_ERROR_UNSUPPORTED_TYPE`

This can be solved with custom error handler in `encode()`, but I think it's breaking compatibility
```php
set_error_handler(function () {
    static::handleJsonError(JSON_ERROR_SYNTAX);
}, E_WARNING);
$json = json_encode($value, $options);
restore_error_handler();
```

freezy-sk:
This is how I handled it https://github.com/freezy-sk/yii2/commit/f0e9078cc61feaa25cc396625f928dbf90bdbb79

SilverFire:
@freezy-sk solution looks good as for me

PR #9289, opened 2015-08-03
Adds cases for newer JSON error codes.

mcgrogan91:
Adds handling for PHP 5.5 JSON error codes JSON_ERROR_RECURSION, JSON_ERROR_INF_OR_NAN, and JSON_ERROR_UNSUPPORTED_TYPE in BaseJson::handleJsonError() related to #9282

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -68,6 +68,7 @@ Yii Framework 2 Change Log
 - Enh #9249: Added `hashCallback` in `yii\web\AssetManager` to allow custom hash generation for asset directory (petrabarus)
 - Enh #9263: Avoid extra DB query in RBAC DbManager in case auth item name is empty (samdark)
 - Enh #9268: Improved display of boolean parameters in logged SQL queries (arkhamvm, samdark)
+- Enh #9282: Adding PHP 5.5 JSON error codes to JSON error handler (mcgrogan91)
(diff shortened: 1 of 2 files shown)


cebe:
this would break on php 5.4 because the constants are not defined. we have to check for the actual constant value here, e.g. `case 1234: // CONSTANT_NAME`.

freezy-sk:
@cebe please don't merge this PR, there are some related issues around encoding/decoding json which should be discussed first.

RusAlex:
Why do not use pre-defined PHP constants http://php.net/manual/en/reserved.constants.php to determine specific PHP version and use constants only for that version ?

freezy-sk:
@RusAlex see https://github.com/yiisoft/yii2/issues/9282

framework/CHANGELOG.md
```diff
@@ -68,6 +68,7 @@ Yii Framework 2 Change Log
 - Enh #9249: Added `hashCallback` in `yii\web\AssetManager` to allow custom hash generation for asset directory (petrabarus)
 - Enh #9263: Avoid extra DB query in RBAC DbManager in case auth item name is empty (samdark)
```
thevataman: л

SilverFire:
Closed by https://github.com/yiisoft/yii2/commit/7bea7b65fdf21ae3e58339cc04376123416fcf81

PR #9290, opened 2015-08-03
fixes https://github.com/yiisoft/yii2/issues/9288

### LINKED ISSUE #9288: Mkdir errors parallel testing
In my parallel testing scenario (which could also potentially occur in production mode) 

```php
yii\base\Exception: Failed to create directory 'T:\wc6dev_test\assets\e87a0c9b\css': mkdir(): File exists
```
FileHelper::createDirectory can throw an error when it tries to create a folder which already exists.

https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseFileHelper.php#L465

What happens:
- it checks for a directory
- if not exist, it will try to create
- but when it is created in parallel already, it would throw an error.

The solution for this, is to re-check folder existance when an exception has occurred. If the folder is already created, throwing an error is not required anymore.

dynasource:
fixes #9288

### DIFF
--- a/framework/helpers/BaseFileHelper.php
+++ b/framework/helpers/BaseFileHelper.php
@@ -462,9 +462,10 @@ class BaseFileHelper
             $result = mkdir($path, $mode);
             chmod($path, $mode);
         } catch (\Exception $e) {
-            if(!is_dir($path)){
-                throw new \yii\base\Exception("Failed to create directory '$path': " . $e->getMessage(), $e->getCode(), $e);
+            if(is_dir($path)){ // #9288
+                return true;
             }
+            throw new \yii\base\Exception("Failed to create directory '$path': " . $e->getMessage(), $e->getCode(), $e);
         }
 
         return $result;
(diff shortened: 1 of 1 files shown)


framework/helpers/BaseFileHelper.php
```diff
@@ -462,7 +462,9 @@ public static function createDirectory($path, $mode = 0775, $recursive = true)
             $result = mkdir($path, $mode);
             chmod($path, $mode);
         } catch (\Exception $e) {
-            throw new \yii\base\Exception("Failed to create directory '$path': " . $e->getMessage(), $e->getCode(), $e);
+            if(!is_dir($path)){
+                throw new \yii\base\Exception("Failed to create directory '$path': " . $e->getMessage(), $e->getCode(), $e);
+            }
```
SDKiller: And what about return value when the second check found that directory exists?

dynasource: I was under the conception that it was returning the file path, but now I see it is returning the $result. I will upload the fix.

samdark:
Merged: https://github.com/yiisoft/yii2/commit/4d21c48bc42a3ada4ec40e645de9f58a42f598f7

Thanks!

samdark:
https://github.com/yiisoft/yii2/commit/6b2748a0b3dafaad4350d92cd50e24da4b4075ec

PR #9294, opened 2015-08-03
alternative solution for issue #7627

### LINKED ISSUE #7627: ActiveForm inputs with custom id are not validated

cebe:
should cover more cases then #8912. Have not tested it yet.

see #7627

Closes #8912
Closes #7627

### DIFF
--- a/framework/widgets/ActiveField.php
+++ b/framework/widgets/ActiveField.php
@@ -143,6 +143,12 @@ class ActiveField extends Component
      */
     public $parts = [];
 
+    /**
+     * @var string this property holds a custom input id if it was set using [[inputOptions]] or in one of the
+     * `$options` parameters of the `input*` methods.
+     */
+    private $_inputId;
+
 
     /**
      * PHP magic method that returns the string representation of this object.
@@ -169,11 +175,11 @@ class ActiveField extends Component
      * and use them as the content.
      * If a callable, it will be called to generate the content. The signature of the callable should be:
      *
-     * ~~~
+     * ```php
      * function ($field) {
      *     return $html;
      * }
-     * ~~~
+     * ```
      *
      * @return string the rendering result
      */
@@ -213,7 +219,7 @@ class ActiveField extends Component
             }
         }
 
-        $inputID = Html::getInputId($this->model, $this->attribute);
+        $inputID = $this->getInputId();
         $attribute = Html::getAttributeName($this->attribute);
         $options = $this->options;
         $class = isset($options['class']) ? [$options['class']] : [];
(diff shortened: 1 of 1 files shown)


dynasource:
Im trying to figure out how one can set _input in your commit, but I cannot find it.

cebe:
you should not set it, it is an internal variable to keep track of whether an explicit ID has been configured or not.

dynasource:
ahh, its done in adjustLabelFor(). There it is.

framework/widgets/ActiveField.php
```diff
@@ -143,6 +143,12 @@ class ActiveField extends Component
      */
     public $parts = [];
 
+    /**
+     * @var string this property holds a custom input id if it was set using [[inputOptions]] or in one of the
+     * `$options` parameters of the `input*` methods.
+     */
+    private $_inputId;
```
SilverFire: I think it should be protected, because `adjustLabelFor` and `getInputId` are protected and it'll be hard to override them when `$_inputId` is private

SilverFire:
Looks better than #8912 and works. :+1:

framework/widgets/ActiveField.php
cebe: I can not think of a case where you would need to override them and if so, you could just override both.

framework/widgets/ActiveField.php
```diff
@@ -748,4 +758,14 @@ protected function getClientOptions()
             'error' => '.help-block',
         ]);
     }
+
+    /**
+     * Returns the HTML `id` of the input element of this form field.
+     * @return string the input id.
+     * @since 2.0.6
```
SDKiller: @cebe 

I guess shoul be `@since 2.0.7`

SilverFire: Thank you! https://github.com/yiisoft/yii2/commit/723e3a9850d6175cffcc5013e1a5f6b5baca4c61

SilverFire:
Added CHANGELOG: https://github.com/yiisoft/yii2/commit/723e3a9850d6175cffcc5013e1a5f6b5baca4c61

PR #9297, opened 2015-08-03
Modified setAttributes()

pBazsi:
Added condition in if statement. Now it does not update the unchanged attributes, so they don't become 'dirty'.

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -690,7 +690,7 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
         if (is_array($values)) {
             $attributes = array_flip($safeOnly ? $this->safeAttributes() : $this->attributes());
             foreach ($values as $name => $value) {
-                if (isset($attributes[$name])) {
+                if (isset($attributes[$name]) && $this->$name !== $value) {
                     $this->$name = $value;
                 } elseif ($safeOnly) {
                     $this->onUnsafeAttribute($name, $value);


cebe:
unchanged attributes aren't marked as dirty anyway, why did you add this?

samdark:
@pBazsi please provide a case where current code fails.

pBazsi:
I can't easily provide exact code sample, because this needs AR to work, but see the following.

If you create a new instance of an AR class, all attributes are set to `null`.
Assume, class `Test` has an attribute named `city`. If I set anywhere the `city` attribute to the same `null` value, I don't want it to get dirty.

@cebe unchanged attributes aren't marked as dirty, if you load an existing record of the database, and it has old attributes, but they do get dirty, when you create an empty new model and set some attributes to `null`

```php
$model = new Test(); // AR class, has attribute `city`
var_dump($model->city); // This dumps `null`
var_dump($model->getDirtyAttributes()); // This dumps `[]`
$model->city = null; // Setting to the same value
var_dump($model->getDirtyAttributes()); // This dumps `array(1) ['city'=>null]`
```

My other idea is to fix `BaseActiveRecord::getDirtyAttributes()`, because it's mainly an AR problem, and my fix was in `Model::setAttributes()`.

In `getDirtyAttributes()`, from line 552 it does not check the `null` value.

```php
if ($this->_oldAttributes === null) {
    foreach ($this->_attributes as $name => $value) {
        if (isset($names[$name])) {
            $attributes[$name] = $value;
        }
    }
}
```

I would change to the following:
```php
if ($this->_oldAttributes === null) {
    foreach ($this->_attributes as $name => $value) {
        if (isset($names[$name]) && $value != null) {
            $attributes[$name] = $value;
        }
    }
}
```

samdark:
Making attribute dirty when setting it to null is expected. Null is a valid value.

tibee:
I don't understand why, because `null` is a valid value, but thinking of a new, empty model, replacing `null` with `null` why should make it dirty? On an old record, replacing an attribute with same value does not make it dirty.

@cebe wrote that 'unchanged attributes aren't marked as dirty anyway', why not on new, empty models?

mj4444ru:
If the getter will be overridden, the result is unpredictable.

mj4444ru:
Perhaps this lack may fix elsewhere.

mj4444ru:
If `$safeOnly` is `true` and the value has not changed, is called `onUnsafeAttribute`.

mj4444ru:
This pull request adds an error in the code. I have previously described how. Why is this pull request is still open?

samdark:
Right. Closing it because it doesn't work in all cases and there are now typecasting capabilities in the core.

PR #9299, opened 2015-08-03
refactored SchemaBuilder

cebe:
The current implementation of SchemaBuilder does not fit the style of how the other schema related classes are organized. It also has the problem that it is created using static methods while still depending on a non-global state. This state is made global by setting a DB instance with static method. I have refactored it to avoid these issues while keeping a similar syntax.

This is what has been done:
- rename class to ColumnSchemaBuilder as this is more appropriate
- changed internal organisation to match how the rest of schema related classes work
  - do not duplicate schema map but use existing one
  - the ColumnSchemaBuilder is now created the same way as QueryBuilder is
- removed static call magic and method annotations, now real methods are called as they are
- the whole code works on objects in a db context now instead of setting database connection in global state
- trait is now used by Migration by default but can be used in other contexts as well

Migration usage is now as follows:

``` php
$this->createTable('example_table', [
  'id' => $this->primaryKey(),
  'name' => $this->string(64)->notNull(),
  'type' => $this->integer()->notNull()->defaultValue(10),
  'rule_name' => $this->string(64),
  'data' => $this->text(),
  'created_at' => $this->datetime()->notNull(),
  'updated_at' => $this->datetime(),
]);
```

related to #8833


framework/db/SchemaBuilderTrait.php
```diff
@@ -37,62 +35,245 @@
...
+     * @return Connection the database connection to be used for schema building.
+     */
+    protected abstract function getDb();
+
+    /**
+     * Creates a primary key column.
+     * @param integer $length column size or precision definition.
+     * This parameter will be ignored if not supported by the DBMS.
+     * @return ColumnSchemaBuilder the column instance which can be further customized.
+     * @since 2.0.6
```
samdark: No need to mention 2.0.6 for each method if the whole class is new.

samdark:
Looks good overall. Advanced app migrations should be adjusted.

framework/db/SchemaBuilderTrait.php
cebe: did this because it is a trait and the method will show up in the Migration class. This way it is clear which are new from the docs.

cebe:
@vaseninm, @pana1990 do you have anything to add here?

vaseninm:
It looks right

vaseninm:
@cebe sorry. i did not understand problem of current implementation in first pull request conversion.

framework/db/SchemaBuilderTrait.php
samdark: OK/

cebe:
no problem. Just wanted to make sure I did not overlook something.

pana1990:
nice job @cebe :+1:

PR #9301, opened 2015-08-04
Test find with params

thiagotalma:
Find with params fails:

``` php
$customer = $customerClass::find()->where(['name' => ':user'])->params([':user' => 'user2'])->one();
```

`yii\db\Exception: SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens`

Possible fix:

db\QueryBuilder.php

``` php
public function buildHashCondition($condition, &$params)
{
    $parts = [];
    foreach ($condition as $column => $value) {
        if (is_array($value) || $value instanceof Query) {
            // IN condition
            $parts[] = $this->buildInCondition('IN', [$column, $value], $params);
        } else {
            if (strpos($column, '(') === false) {
                $column = $this->db->quoteColumnName($column);
            }
            if ($value === null) {
                $parts[] = "$column IS NULL";
            } elseif ($value instanceof Expression) {
                $parts[] = "$column=" . $value->expression;
                foreach ($value->params as $n => $v) {
                    $params[$n] = $v;
                }
            } elseif (isset($params[$value])) {
                $parts[] = "$column=$value";
            } else {
                $phName = self::PARAM_PREFIX . count($params);
                $parts[] = "$column=$phName";
                $params[$phName] = $value;
            }
        }
    }
    return count($parts) === 1 ? $parts[0] : '(' . implode(') AND (', $parts) . ')';
}
```

### DIFF
--- a/tests/framework/ar/ActiveRecordTestTrait.php
+++ b/tests/framework/ar/ActiveRecordTestTrait.php
@@ -100,6 +100,11 @@ trait ActiveRecordTestTrait
         $this->assertTrue($customer instanceof $customerClass);
         $this->assertEquals(2, $customer->id);
 
+        // find with params
+        $customer = $customerClass::find()->where(['name' => ':user'])->params([':user' => 'user2'])->one();
+        $this->assertTrue($customer instanceof $customerClass);
+        $this->assertEquals(2, $customer->id);
+
         // scope
         $this->assertEquals(2, count($customerClass::find()->active()->all()));
         $this->assertEquals(2, $customerClass::find()->active()->count());


klimov-paul:
This is incorrect usage: if you use array syntax for the `where` value will be bound as parameter automatically. You should use `yii\db\Expression` if you wish to pass raw SQL, which includes parameter name.

thiagotalma:
@klimov-paul Sorry, I forgot to explain the use case.

Obviously, the query is not made that way.

In a model with composite primary key, I override the `find()` and add the `where()`, with the intention of forcing that that model is always seeking to foreign primary key.

Items.php
```php
/**
 * @inheritDoc
 */
public static function find()
{
    return parent::find()->joinWith('order')->andWhere(['{{%Item}}.id' => ':item_id', '{{%Item}}.order_id' => ':order_id']);
}
```

Controlle.php
```php
$item = Item::find()->params([':item_id' => $item_id, ':order_id' => $order_id])->one();
$item->delete();
```

The intention here is not to risk inform only one of the keys and delete the wrong registry.

Or you suggest me another way to ensure this safety?

Thanks!

klimov-paul:
This is not correct:
```php
->andWhere(['{{%Item}}.id' => ':item_id', '{{%Item}}.order_id' => ':order_id']);
```
Here `:item_id` will be added as named param value.
You should either pass pass values for columns directly or use string syntax:
```php
->andWhere('{{%Item}}.id =:item_id AND {{%Item}}.order_id = :order_id']);
```

thiagotalma:
Yes. It was just a didactical example.  :wink:

PR #9302, opened 2015-08-04
Do not add param when it was already set

thiagotalma:
Fix #9301


klimov-paul:
See explanation at #9301

PR #9305, opened 2015-08-04
Portability fixes for SchemaBuilder on MSSQL

cebe:
Changed binary mapping to `varbinary(max)` on MSSQL instead of `binary(1)` (which requires a length and is fixed length)
to better match other dbms.

Changed timestamp mapping to datetime on MSSQL. Timestamp is a unique number on MSSQL, not a timestamp.

separated from #9191

### DIFF
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -34,10 +34,10 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_DOUBLE => 'float',
         Schema::TYPE_DECIMAL => 'decimal',
         Schema::TYPE_DATETIME => 'datetime',
-        Schema::TYPE_TIMESTAMP => 'timestamp',
+        Schema::TYPE_TIMESTAMP => 'datetime',
         Schema::TYPE_TIME => 'time',
         Schema::TYPE_DATE => 'date',
-        Schema::TYPE_BINARY => 'binary(1)',
+        Schema::TYPE_BINARY => 'varbinary(max)',
         Schema::TYPE_BOOLEAN => 'bit',
         Schema::TYPE_MONEY => 'decimal(19,4)',
     ];


nkovacs:
Docs for timestamp: https://technet.microsoft.com/en-us/library/aa260631(v=sql.80).aspx

samdark:
Timestamp doesn't look right according to these docs while datetime does.

nkovacs:
You can work around both problems by using `datetime` and `varbinary(max)` manually, but you lose portability.

samdark:
Looks OK to merge.

cebe:
what about BC break?

samdark:
Umm, yeah. 2.1 then...

nkovacs:
It would also make sense to remove the `$length` parameter of `SchemaBuilderTrait::binary`, for the same reasons as text after this change.

framework/db/mssql/QueryBuilder.php
```diff
@@ -34,10 +34,10 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_DOUBLE => 'float',
         Schema::TYPE_DECIMAL => 'decimal',
         Schema::TYPE_DATETIME => 'datetime',
-        Schema::TYPE_TIMESTAMP => 'timestamp',
+        Schema::TYPE_TIMESTAMP => 'datetime',
         Schema::TYPE_TIME => 'time',
         Schema::TYPE_DATE => 'date',
-        Schema::TYPE_BINARY => 'binary(1)',
+        Schema::TYPE_BINARY => 'varbinary(max)',
```
cebe: this has already been applied on master: #12599

cebe:
Is this still valid, if so, should we merge it to 2.1?

nkovacs:
There's no point in keeping backwards compatibility, since the current code is completely wrong and doesn't work, which is why binary has already been fixed in master.

cebe:
I see, thanks!

cebe:
Merged, thanks!

PR #9307, opened 2015-08-04
Remove length from SchemaBuilderTrait::text.

nkovacs:
I did not want to remove the tests, because they also checked that `"text(255)"` is converted to `"text"` by getColumnType, so I fixed them the same way I did in #9191


cebe:
Thank you!

PR #9309, opened 2015-08-04
Improve docs for bindParam() $value

mikehaertl:
I think, the original comment here is incorrect. If it was a name, it would be of type `string`. But it's the variable itself. This is also wrong in the PHP manual IMO.


samdark:
Correct. Merged.

samdark:
Thanks!

Issue #9317, opened 2015-08-05
Fatal error when using $user->can


dynasource:
A fatal error is thrown when using $user->can, when authManager is not yet defined. This breaks the stacktrace. A pull request will follow.

samdark:
What do you mean by "breaks the stacktrace"?

dynasource:
the following line renders a fatal error when authmanager is not defined:
https://github.com/yiisoft/yii2/blob/master/framework/web/User.php#L661

cannot access property on object which isnt defined.

samdark:
Ah, right. If you'll install XDebug it will show you full stacktrace. That's PHP limitation.

But overall, that's probably a good check.

dynasource:
I guess this check better suits the method description also: https://github.com/yiisoft/yii2/blob/master/framework/web/User.php#L642

creocoder:
@dynasource Why you close it?!

dynasource:
I prefer the solution in application . If it has getters over there, I prefer errorhandling over there like in your solution.

PR #9318, opened 2015-08-05
fixes fatal error when authManager is not yet defined

dynasource:
#9317

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -658,7 +658,10 @@ class User extends Component
         if ($allowCaching && empty($params) && isset($this->_access[$permissionName])) {
             return $this->_access[$permissionName];
         }
-        $access = $this->getAuthManager()->checkAccess($this->getId(), $permissionName, $params);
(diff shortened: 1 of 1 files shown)


cebe:
if we do this, we should do it inside of `getAuthManager()` but I am not sure it is really needed.

dynasource:
its a choice. Either catch a fatal error or not.

creocoder:
Suggest to fix different way by default config to `access manager` component.

dynasource:
good suggestion

should I update this commit to fix this in https://github.com/yiisoft/yii2/blob/master/framework/base/Application.php#L591

Currently:
```php
    public function getAuthManager()
    {
        return $this->get('authManager', false);
    }
```

To:
```php
    public function getAuthManager()
    {
        return $this->get('authManager');
    }
```

creocoder:
@dynasource https://github.com/yiisoft/yii2/pull/9368 (fix for asset manager and cache component)

Issue #7331, opened 2015-02-18
Proposal for Model, setter method for attributes


adamasantares:
Hi all.

I've create and tested this method for class /vendor/yiisoft/yii2/base/Model.php

```php
    /**
     * Sets the attribute value by calling it as setter method
     * By default the attributes work as this:  $model->attributeName = value; $mode->save();
     * But this overload allow to use like this:  $model->setAttributeName(value)->setAnotherAttr(value2)->save();
     * @param string $name
     * @param array $attributes
     * @return mixed
     */
    public function __call($name, $attributes)
    {
        if (!empty($attributes) && (is_numeric($attributes[0]) || is_string($attributes[0]))) {
            foreach ($this->attributes() as $attr) {
                $setter = 'set' . ucfirst($attr);
                if ($name == $setter) {
                    $this->$attr = $attributes[0];
                    return $this;
                }
            }
        }
        // default overload from Component class
        return parent::__call($name, $attributes);
    }
```

I think this method is good solution, with it you can set values for different attributes in one line.
But I really  have not time for make a PR.

samdark:
What's the benefit exactly?

adamasantares:
Now I must write:

```php
$user->email = 'aa@bb.cc';
$user->phone = '8888888888';
$user->save();
```

totally 3 time uses of $user.

With this method I can write

```php
$user->setEmail('aa@bb.cc')->setPhone('8888888888')->save();
```

I'd like that style more.

It can be without prefix 'set', just ```$user->email('value');``` but in this case may be problem with methods which has same names.

klimov-paul:
Whay don't simply use `setAttributes()` then?

adamasantares:
@klimov-paul good question... :)
Try to compare

```php
$model->setAttributes(['email' => 'aa@bb.cc', 'phone' => '8888888888']);
$model->save();
```

or

```php
$user->setEmail('aa@bb.cc')->setPhone('8888888888')->save();
```

well, last style more compact and nothing more...

lynicidn:
solve:
```php
function setPhone($phone)
{
$this->setAttribute('phone', $phone);
return $this;
}
```

dynasource:
I like the idea of saving a value in 1 line, but I dont think you have thought about the huge performance loss. 

__call is one of the most called functions in Yii2. If its not optimal, it shouldnt be altered

samdark:
I don't think we want it in the framework:

1. Performance issue.
[...]

PR #9321, opened 2015-08-05
enable one-line $model update

dynasource:
fix for #7331 and enable shorter code


samdark:
Wasn't it decided not to implement it?

cebe:
yes

dynasource:
have you read my post https://github.com/yiisoft/yii2/issues/7331#issuecomment-127970269

dynasource:
I cannot think of a good reason why the framework should not provide a way of setting & saving a model in 1 line.

* its useful
* its something many people want to use (and finally come to the conclusion its not supported)
* this solution does not affect performance (no magic and not using __call)
* arguments about similar proposals (https://github.com/yiisoft/yii2/issues/7331#issuecomment-74869942) is not an argument

PR #9324, opened 2015-08-05
Fix mysql QueryBuilder boolean type

mj4444ru:
The problem is that yii type boolean type is considered mysql bit (1), and in QueryBuilder it is incorrect. When we create a migration using Schema :: TYPE_BOOLEAN, database type we get Schema :: TYPE_SMALLINT, and the model of type integer.

/framework/db/mysql/Schema.php

if ($column->size === 1 && $type === 'bit') {
    $column->type = 'boolean';
}

/framework/db/mysql/QueryBuilder.php

public $typeMap = [
...
Schema::TYPE_BOOLEAN => 'tinyint(1)',
...
];


samdark:
What is the problem currently?

mj4444ru:
Проблема в том, что для yii типом boolean считается тип mysql bit(1), а в QueryBuilder он некорректен. Когда мы создаём миграцию с использованием Schema::TYPE_BOOLEAN, в базе данных мы получаем тип Schema::TYPE_SMALLINT, а модель с типом integer.

Простите что сразу не написал. Не успел оформить.

mj4444ru:
The problem is that yii type boolean type is considered mysql bit (1), and in QueryBuilder it is incorrect. When we create a migration using Schema :: TYPE_BOOLEAN, database type we get Schema :: TYPE_SMALLINT, and the model of type integer.

Sorry that just has not. I do not have time to issue.

samdark:
What do you mean by "in QueryBuilder it is incorrect"?

mj4444ru:
public $typeMap = [
...
Schema::TYPE_BOOLEAN => 'tinyint(1)',
...
];

samdark:
Why `tinyint(1)` is incorrect?

Faryshta:
@samdark its not wrong to use `tinyint(1)` simply 'bit(1)` requires less space in disk and can get lesser possible values. For example using tinyint you can insert a value equal 2.

The problem I think would be if all SQL engines support it to avoid generating problems when migrating from one db engine to another

cebe:
the mysql synonym for boolean is tinyint(1):

> BOOL, BOOLEAN
> 
> These types are synonyms for TINYINT(1). A value of zero is considered false. Nonzero values are considered true: 

https://dev.mysql.com/doc/refman/5.0/en/numeric-type-overview.html

mj4444ru:
Why the alias MySQL? yii type boolean considers the type bit (1).

cebe:
it considers bit(1) and tinyint(1), what is the problem?

mj4444ru:
The problem is that yii in accordance with the "mysql Schema" said type boolean type mysql bit (1).

Code file "/framework/db/mysql/Schema.php"
if ($ column-> size === 1 && $ type === 'bit') {
     $ column-> type = 'boolean';
}

At this time, "/framework/db/mysql/QueryBuilder.php" said boolean type tinyint (1).

Thus "QueryBuilder.php" contradicts "Schema.php".

It does not matter what's in the mysql alias type "boolean".
It is important to consider that type of "boolean" the core yii.

cebe:
this line should cover the tinyint(1) case too then. https://github.com/yiisoft/yii2/blob/master/framework/db/mysql/Schema.php#L160 I thought it was already doing this...

mj4444ru:
@cebe So can not be done
`tinyint(1)` - a number from -127 to 127

cebe:
closing this in favor of #9339

PR #9332, opened 2015-08-06
Use minify assets

muhammadcahya:


cebe:
This is not valid PHP code...

PR #9339, opened 2015-08-06
typecast tinyint(1) as boolean

Faryshta:
i don't know where to make an unit test for this.

### DIFF
--- a/framework/db/mysql/Schema.php
+++ b/framework/db/mysql/Schema.php
@@ -157,7 +157,7 @@ class Schema extends \yii\db\Schema
                     if (isset($values[1])) {
                         $column->scale = (int) $values[1];
                     }
-                    if ($column->size === 1 && $type === 'bit') {
+                    if ($column->size === 1 && in_array($type, ['bit', 'tinyint'], true)) {
                         $column->type = 'boolean';
                     } elseif ($type === 'bit') {
                         if ($column->size > 32) {
--- a/tests/framework/db/SchemaTest.php
+++ b/tests/framework/db/SchemaTest.php
@@ -269,7 +269,7 @@ class SchemaTest extends DatabaseTestCase
                 'defaultValue' => '2002-01-01 00:00:00',
             ],
             'bool_col' => [
-                'type' => 'smallint',
+                'type' => 'boolean',
                 'dbType' => 'tinyint(1)',
                 'phpType' => 'boolean',
                 'allowNull' => false,
@@ -281,7 +281,7 @@ class SchemaTest extends DatabaseTestCase
                 'defaultValue' => null,
             ],
             'bool_col2' => [
-                'type' => 'smallint',
+                'type' => 'boolean',
                 'dbType' => 'tinyint(1)',
                 'phpType' => 'boolean',
                 'allowNull' => true,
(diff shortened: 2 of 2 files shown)


cebe:
there is already a test for that and it is broken by your PR: https://travis-ci.org/yiisoft/yii2/jobs/74439667#L206 ;)
However this change is breaking BC so we have to wait for 2.1.

related to #9324

Faryshta:
finally it passes all tests :smile:

mj4444ru:
`tinyint(1)` - a number from -127 to 127 (in practice)
`tinyint(1)` - a number from 0 to 9 (in theory)

possible conflict with user-defined types

mg-code:
I agree with mj4444ru, in some tables I make type column as tinyint(1).

Faryshta:
@mj4444ru you mean things like when a user have an `stage` column like

0 => Inactive
1 => In Process
2 => Completed
3 => Cancelled

mj4444ru:
@Faryshta yes
Some people prefer to use instead of `enum` `smallint (1)`.

Faryshta:
@mj4444ru I agree with you, I prefer to use `bit(1)` for boolean but right now mysql bool means tinyint(1)

a way to 'fix' this `tinyint` as enum would be to use `smallint(1)` instead which means one extra byte

mj4444ru:
@Faryshta In mysql `boolean` alias `tinyint(1)` for compatibility.
You do not need to pay attention to it.
You are all too complicated.

dynasource:
I dont see the value of introducing this magic. It looks wrong.

cebe:
@dynasource https://github.com/yiisoft/yii2/pull/9324#issuecomment-128127967 mysql considers tinyint(1) as boolean and so should yii. Note that tinyint(1) is different from tinyints that have more that one digit, e.g. tinyint(2) or tinyint(3), which are not considered to be boolean.

mg-code:
@cebe actually mysql considers boolean as tinyint(1), not visa versa. 
And by the way the only one difference between tinyint(1) and tinyint(3) is column width in mysql clients.
So you can save up to 255 in tinyint(1) column.

cebe:
@mg-code I know. My point here is that if you create a table with column e.g. `'active' => 'boolean NOT NULL'` and then get back the table schema from yii to find a `tinyint(1)` is confusing.
If you intend to store a 127 (note that 255 is unsigned) in a tinyint you would not choose a column width of 1. So a `tinyint(1)` should be boolean in column schema, `tinyint(2 or higher)` should be normal tinyints.

mg-code:
@cebe I know about what you are talking. Personally I use smallint(2) for statuses and types, but tinyint(1) works great for values up to 9 too.
It does not look weird that someone who were using tinyint(1) for statuses will get boolean instead of integer?
What will happen when we will use asArray() function or query builder? We will get different data type for same column?
Synonym of boolean -> tinyint(1) was introduced before MySQL 5.0.3, because there was no bit data type. bit(1) also was alias of tinyint(1). Bit(1) data type fits better for boolean.

I never saw that MySQL said that tinyint(1) should be treated as boolean. 
If they do not recommend it, why Yii framework should? Especially if it was wrong decision to make such synonym (otherwise we would not discussing this).

I don't like idea that Yii framework determine rules for database scheme.

Faryshta:
https://github.com/yiisoft/yii2/blob/2.1/framework/db/mysql/QueryBuilder.php#L44 the migrations created by yii2 generate a `tinyint(1)` when using the `Migration::boolean()` method to define a column.

samdark:
Won't be merged into 2.0.

PR #9356, opened 2015-08-08
Session's get&has optimisations

requilence:
In the case when user didn't sent Session ID and we don't yet opened the session, variable is 100% not exist so we can return $defaultValue. This fix also protects from auto-opening the session when not necessary.

I'm digging in this issue because of flooding my db with empty sessions. This happens because of  this line in my controller `$session=Yii::$app->session;` without actually performing the `->get()` (not reached within if condition) + Yii Debug panel that call  `get->($session->flashParam)` after content already sent to user. 
In Debug panel's **RequestPanel**:

```
$session = Yii::$app->has('session', true) ? Yii::$app->get('session') : null;
        if ($session === null) {
            return [];
        }
```

So this check passes, `session->get->('__flash')` called and empty **__flash** object stored in Session storage but because of content already sent to user this happens again and again on every request.

I don't think saving fresh session to storage without actually sending PHPSESSID to user is good way. How we can fix this?

### DIFF
--- a/framework/web/Session.php
+++ b/framework/web/Session.php
@@ -555,6 +555,7 @@ class Session extends Component implements \IteratorAggregate, \ArrayAccess, \Co
      */
     public function get($key, $defaultValue = null)
     {
+        if(!$this->getHasSessionId() && !$this->getIsActive()) return $defaultValue;
         $this->open();
         return isset($_SESSION[$key]) ? $_SESSION[$key] : $defaultValue;
     }
@@ -606,6 +607,7 @@ class Session extends Component implements \IteratorAggregate, \ArrayAccess, \Co
      */
     public function has($key)
     {
+        if(!$this->getHasSessionId() && !$this->getIsActive()) return false;
         $this->open();
         return isset($_SESSION[$key]);
     }


dynasource:
this doubles the call getIsActive as its also used in ->open
https://github.com/yiisoft/yii2/blob/master/framework/web/Session.php#L119

```php
      if ($this->getIsActive()) {
            return;
        }
```

Requilence:
@dynasource You are right. ` && !$this->getIsActive() ` is unnecessarily in my commit but I think it's a reasonable to leave this full condition here to avoid possible logical error if something in the ->open method will change.

In other words. Here we check in context of method (getting the variable). When the session variable is 100% inaccessible? When request doesn't contains session indification info and we wasn't create session object for the response yet.

kidol:
Wouldn't it be better to change signature of `open()` instead? Something like:
```php
/**
@param boolean $force whether to forcefully open the session even if no session id is present.
@return boolean whether the session has been opened.
*/
public function open($force = true)
```

Then do:

```php
return ($this->open(false) && isset($_SESSION[$key])) ? $_SESSION[$key] : $defaultValue;
```

Requilence:
@kidol I think that additional param is not clear in the context of open function. **->open**  method just need to start the PHP session if it is not already been started. This atomicity is also important for Unit testing.

dynasource:
apart from ->open, what you are also pointing out with your commit, is the need for a central function covering both '$session->getHasSessionId() || $session->getIsActive()'

You are using it in your commit (so it means something) and it is also applied in 
https://github.com/yiisoft/yii2/blob/master/framework/web/User.php#L607

Vatruha:
I'm not sure that I'm in time here.

My task #9694 was closed as duplicat, but I'm not sure about it.

So I'm going to write here :)

As I can see in commit there you want to change only 2 methods. I think that we need to change more:
getIterator
getCount
get
remove
removeAll
has
offsetExists
offsetGet
offsetUnset

The main idea: we will create session (session->open()) ONLY when we need to put some data in it or when it was created previosly.

So, also we need to add $this->open() for:
setFlash
addFlash

Can you look at #9694 task for more infomation?

Thanks

dynasource:
@Vatruha: I like your fundamental view on session by separating these in 'getters' and 'setter'. From this approach, perhabs its possible to create solution, in which the checking on session is done only once, instead off in >10 functions.

Vatruha:
@dynasource, yeah, it make sense.

```php
    /**
     * Initializes the application component.
     * This method is required by IApplicationComponent and is invoked by application.
     */
    public function init()
    {
        parent::init();
        startSessionIfItWasCreatedBefore();
        register_shutdown_function([$this, 'close']);
    }

    /**
     * Starts the session which was created in previous requests
     */
    protected function startSessionIfItWasCreatedBefore()
    {
        // Session is already started
        if ($this->getIsActive()) {
            return;
        }

        // Session was started in previous requests. Now we just open it
        // If we have no session at all we just create $_SESSION array to make our life easier
        if ($this->getHasSessionId()) $this->open();
        elseif (!isset($_SESSION)) $_SESSION = [];
    }

    /**
     * Starts the session.
     */
    public function open()
    {
        if ($this->getIsActive()) {
            return;
        }
        $this->registerSessionHandler();
        $this->setCookieParamsInternal();

        // If someone somewhere has put some data into $_SESSION directly, we have to pass this data into real $_SESSION array 
        $tmp = $_SESSION;
        @session_start();
        $_SESSION = array_merge_recursive($tmp, $_SESSION);

        if ($this->getIsActive()) {
            Yii::info('Session started', __METHOD__);
            $this->updateFlashCounters();
        } else {
            $error = error_get_last();
            $message = isset($error['message']) ? $error['message'] : 'Failed to start session.';
            Yii::error($message, __METHOD__);
        }
    }
```

And we need delete $this->open() in these methods:
getIterator
getCount
get
remove
removeAll
has
offsetExists
offsetGet
offsetUnset

And we need add $this->open() in the beginning of:
setFlash
addFlash

What do you think?
I didn't check this but it should work ;-)

samdark:
Won't make it into 2.0.

Issue #9353, opened 2015-08-08
Using `AccessFilter` may lead to php fatal error in some cases


creocoder:
Example:

```php
class SiteController extends Controller
{
    /**
     * @inheritdoc
     */
    public function behaviors()
    {
        return [
            'access' => [
                'class' => AccessControl::className(),
                'rules' => [
                    [
                        'actions' => ['index'],
                        'roles' => ['*'],
                        'allow' => true,
                    ],
                ],
            ],
        ];
    }
```

Will lead to:

```
PHP Fatal Error – yii\base\ErrorException
Call to a member function checkAccess() on null
```

cebe:
stacktrace?

creocoder:
@cebe Stacktrace:

1. in C:\Users\creocoder\Projects\test\vendor\yiisoft\yii2\web\User.php at line 661
2. yii\base\ErrorHandler::handleFatalError()

Nothing more.

cebe:
do you have an error action configured and not provided access to it in AccessControl filter?

creocoder:
@cebe Please check code. This is pure AccessFilter (under layer components) coding mistakes. Any non existent role will lead to that error. Also not that trouble happen only when user is guest.

creocoder:
> do you have an error action configured and not provided access to it in AccessControl filter?

@cebe `'only' => ['index']` does not affect anything. Fatal error is happen in any case.

dynasource:
this is a reason for applying https://github.com/yiisoft/yii2/pull/9318

You are using $user->can when authManager is not yet defined.
With a full stacktrace you could find out where.

creocoder:
> this is a reason for applying #9318

This is NOT RELATED issue.

creocoder:
Dublicate of #9317

PR #9368, opened 2015-08-10
Fix for #9317, #9353

creocoder:
Removed two Yii 2 disservices which may lead to fatal errors in several components.

### DIFF
--- a/framework/base/Application.php
+++ b/framework/base/Application.php
@@ -514,7 +514,7 @@ abstract class Application extends Module
      */
     public function getCache()
     {
-        return $this->get('cache', true);
+        return $this->get('cache');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


creocoder:
Rethink my mind. It may be interested to know that components not defined.

dynasource:
its quite confusing that some application's components do return an error and some dont. I would choose for consistency and let the enduser decide if an error is to be thrown

creocoder:
@dynasource This PR is closed 8 minutes before you write comment ;-) Correct fix is https://github.com/yiisoft/yii2/pull/9369

dynasource:
I know. I would like to know why you choose to skip this commit and prefer inconsistency.

creocoder:
@dynasource #9369 is much more consistent and agile.

PR #9369, opened 2015-08-10
Fix for #9317, #9353 version 2

creocoder:


creocoder:
Re #9317, #9353

samdark:
Not sure it's a good idea. How could I figure out that I haven't configured manager instead of not having access?

creocoder:
There is lot of situations when you do not want configure `authManager`. I'm sure fix is 100% correct.

creocoder:
There is same situation like with `cache` component. Not configured === no cache.

samdark:
OK. Makes sense since defaults are strict.

dynasource:
shouldnt $application->get('i18n') do the same then? Not configured, no translation?

klimov-paul:
Fix seems to be correct, since 'authManager' component is not mandatory for the application:
https://github.com/yiisoft/yii2/blob/master/framework/base/Application.php#L591
Compare with 'urlManager' for instance:
https://github.com/yiisoft/yii2/blob/master/framework/base/Application.php#L563

creocoder:
@dynasource 

> shouldnt $application->get('i18n') do the same then? Not configured, no translation?

Its already works like that.

@samdark , @klimov-paul If everything ok, i suggest to merge.

dynasource:
To sum up my confusion:

I was trying to see the logic in these components https://github.com/yiisoft/yii2/blob/master/framework/base/Application.php.

I saw $app->i18n in corecomponents:
```php
    public function coreComponents()
    {
        return [
            'log' => ['class' => 'yii\log\Dispatcher'],
            'view' => ['class' => 'yii\web\View'],
            'formatter' => ['class' => 'yii\i18n\Formatter'],
            'i18n' => ['class' => 'yii\i18n\I18N'],
            'mailer' => ['class' => 'yii\swiftmailer\Mailer'],
            'urlManager' => ['class' => 'yii\web\UrlManager'],
            'assetManager' => ['class' => 'yii\web\AssetManager'],
            'security' => ['class' => 'yii\base\Security'],
        ];
    }
```

 but I didnt see $app->db over there. Still, both have the same syntax:
```php
  public function getDb()
    {
        return $this->get('db');
    }
    public function getI18n()
    {
        return $this->get('i18n');
    }
```

Therefore I thought the logic was not inside the corecomponents/mandotory part. But now you say it is, I dont (yet) understand why a DB is mandatory.

creocoder:
@dynasource `Db` is not mandatory, but it obvious that it should throw error when you try access it and its not configured. There is 3 situations:

1) component which you try to acess throw error when not configured (e.g. db)
2) component which have default configuration
3) component which will be null (**FROM LOT OF PRACTICAL REASONS**) when not configured (e,g. cache, authManager)

dynasource:
I see. Thanx

samdark:
Thanks!

Issue #5004, opened 2014-09-13
viaTable issue


user57376:
When i use the following code:

``` php
A::find()->with(['b','b.c'])->where(['a.id'=>1,'a.v'=>1])->one();
```

The result gives A and B filled but C as an empty array. When I use asArray() C is filled as expected.

All tables have composite keys and the relation 'b.c' uses viaTable().

I found out that the variables in the serialized keys used for mapping have different datatypes and therefore could not be matched.

A clean workaround is using via() instead of viaTable().

nineinchnick:
@user57376 can you point to the code with the datatype checks you mention?

user57376:
The datatype checks are within the trait ActiveRelationTrait in several places.

In my case, when the activerecord class gets populated the key values are converted to integers. The viaTable stays as an array with datatype 'string' values. In the method ActiveRelationTrait::getModelKey($model, $attributes) both composite keys become serialized and later compared to each other. but 'a:3:{i:0;i:1;i:1;i:1;i:2;i:5;}' (from AR) and 'a:3:{i:0;s:1:"1";i:1;s:1:"1";i:2;s:1:"5";}' (from viaTable) wont match.
 
This issue will not occur when using asArray() because then the values aren't converted.

Maybe type casting to string in getModelKey() will solve this issue.

user57376:
Didn't realize i could propose a solution too. What about this:


```php
trait ActiveRelationTrait
{
    
    ...
    
    public function populateRelation($name, &$primaryModels)
    {
        if (!is_array($this->link)) {
            throw new InvalidConfigException('Invalid link: it must be an array of key-value pairs.');
        }

        if ($this->via instanceof self) {
            // via junction table
            /* @var $viaQuery ActiveRelationTrait */
            $viaQuery = $this->via;
            $viaModels = $viaQuery->findJunctionRows($primaryModels, $this->modelClass, $this->link, $this->asArray);
            $this->filterByModels($viaModels);
        } elseif (is_array($this->via)) {
            // via relation
            
        ...
        
    }
    
    ...
    
    /**
     * @param array $primaryModels either array of AR instances or arrays
     * @return array
     */
    private function findJunctionRows($primaryModels, $childClass, $childLink, $childAsArray)
    {
        if (empty($primaryModels)) {
            return [];
        }
        $this->filterByModels($primaryModels);
        /* @var $primaryModel ActiveRecord */
        $primaryModel = reset($primaryModels);
[...]

PR #9371, opened 2015-08-10
Fixed eager loading with composite viatable relation #5004

user57376:
#5004


cebe:
Thanks for your effort providing this PR and the fix. However I am not able to check this PR as there are conflicts with current test files. Also @dynasource does not seem to be able to reproduce the problem in #5004. I am closing this PR now as it is outdated. If the issue is still valid, please let us know, I am happy to reopen this and discuss the fix further. Thanks!

PR #9374, opened 2015-08-10
Prevent AR in migrations by default

TriAnMan:
Prohibit usage of an ActiveRecord in migrations by default.
AR in migrations leads in future to broken migrations, indulges a hard-to-predict behavior of a migration in case of `ActiveRecord::beforeSave()` events.

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -8,8 +8,11 @@
 namespace yii\console\controllers;
 
 use Yii;
+use yii\base\Event;
 use yii\console\Exception;
 use yii\console\Controller;
+use yii\db\ActiveQuery;
+use yii\db\BaseActiveRecord;
 use yii\helpers\Console;
 use yii\helpers\FileHelper;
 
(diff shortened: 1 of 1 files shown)


klimov-paul:
Although ActiveRecord usage inside the mogation is a bad practice - it is up to developer to choose if he/she wants to use it or not.
There should not be any explicit restrictions on this.

TriAnMan:
Ok, but this patch allows developer to choose his strategy. He can enable AR for single migration.
But as I see in the real world everyone is using the most straightforward way to do something. So AR is very straightforward to be used in the migrations but its long time effects are not so obvious but are deadly. And you can face them when you can't change anything. Then you have to deal with SQL dumps. Migrations become nonworking at all.

I think the framework should lead developers to right ways even this ways are not so comfortable at the beginning.

cebe:
this is a documentation issue then. We can add a note to the docs about it.

TriAnMan:
@cebe You are right from some point of view. But in a "real world" everybody ignores restrictions of the documentation. Only the code that will fail or can't be used in a wrong way can prevent somebody from "shooting himself".

cebe:
We are not here to tell users what to do and force our opinion on them by adding code that is only ther to prevent some usage. Yii is a flexible tool that can be used in different ways even in some that we do not see yet or did not make it to be used in. If someone wants to "shoot himself" he is free to do so. The only thing we can do is warn about possible problems in the docs. If you don't read the docs that is not our problem.

TriAnMan:
We can do this patch less restrictive by setting `$prohibitActiveRecord = false` by default and providing a comprehensive documentation note about importance of changing it to `true` and set it to `true` in example apps configs.

cebe:
If in a dev team "everybody ignores restrictions of the documentation" that is a problem of that team and should not be fixed by code in yii.

dynasource:
I think this is overcomplicating things and making Yii less flexible. If I want to access ActiveRecord in a migration, I might well have a very good reason for it. Why letting Yii2 decide not to do it.

TriAnMan:
"Everybody lies" (c) House MD
"Everybody ignores the doc" (c) proverb

TriAnMan:
@cebe @klimov-paul what do you think about less restrictive patch I've proposed:
> We can do this patch less restrictive by setting `$prohibitActiveRecord = false` by default and providing a comprehensive documentation note about importance of changing it to `true` and set it to `true` in example apps configs.

cebe:
@TriAnMan you can apply this in your project if you feel the need for it. Such restrictions do not fit with yii philosophy.

cebe:
added a note to the docs,.

TriAnMan:
@cebe
> you can apply this in your project
 have already done that. (:

Nice note you wrote. Can't formulate this idea better. (:

cebe:
thanks :)

dynasource:
definately. This sentence makes me doubt you're actually German:
>  who's nature is to stay constant forever, application logic is subject to change

PR #9375, opened 2015-08-10
Refresh a DB schema cache after any db manipulations

TriAnMan:
This patch will help with DB cache invalidation after any migrations were applied.


klimov-paul:
Migration already resets schema before being applied.
Why such change is needed?

TriAnMan:
Ok, sorry.
I'm using 2.0.2 and there is no schema reset into `Migration::init()`.

PR #9380, opened 2015-08-11
use binary save string comparison

cornernote:


cebe:
can you explain what this change is for? What problem are you trying to solve?

cornernote:
I want to check for dirty attributes and some of my attributes are gzcompressed.  It tells me the data is changed, although its the same.

cebe:
can you provide a unit test for this? Or some example PHP code to reproduce the issue?

cornernote:
I just realised it causes issues.  It also needs to check that the attribute is in fact a string.

I'll make some tweaks, and provide a unit test.

dynasource:
there are some caveats as you can see at php.net
http://php.net/manual/en/function.strcmp.php#108563

cornernote:
While creating the test to prove my point I found that the `!==` operator works fine and the issue was somewhere else in my code.

Sorry for being hasty about this and not providing tests to begin with.

PR #9389, opened 2015-08-12
Fix problem, that Application::getUser can return null

dkushnikov:


samdark:
How to reproduce the problem?

dkushnikov:
You may override getUser or coreComponents methods in your own Application (based on yii\web\Application).

samdark:
Thanks!

PR #9391, opened 2015-08-12
improved Chg #9369: 

schojniak:
if authManager wasn't configured then caching wouldn't be used

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -658,10 +658,8 @@ class User extends Component
         if ($allowCaching && empty($params) && isset($this->_access[$permissionName])) {
             return $this->_access[$permissionName];
         }
-        if (($manager = $this->getAuthManager()) === null) {
-            return false;
-        }
-        $access = $manager->checkAccess($this->getId(), $permissionName, $params);
+        $manager = $this->getAuthManager();
+        $access = $manager ? $manager->checkAccess($this->getId(), $permissionName, $params) : false;
         if ($allowCaching && empty($params)) {
             $this->_access[$permissionName] = $access;
         }


samdark:
It's the same as it was. Why changing it?

schojniak:
not the same, I think that return false between two ```if ($allowCaching &&...``` is bad idea

samdark:
Why?

schojniak:
1. ```$this->_access[$permissionName] = $access``` won't be executed
2. each time it will check empty cache then check if authManager is configured instead of return false from cache
3. I know that calling ```getAuthManager``` has low cost, so it's only cosmetic...

samdark:
1. It won't be executed in the code you've changed as well.
2. Why it's better?
3. There's no change in this regard. In both cases it's called.

schojniak:
1. In my code it WILL be executed, because:
default parameters are ```$params = [], $allowCaching = true```, so a condition in ```if ($allowCaching && empty($params)) {``` is true, and inner block will be executed; 
in previous code there is ```return false``` before this ```if```
2. It's description of behavior without my fix
3. It's not true. If you call again ```can(...)``` then ```$this->_access[$permissionName]``` has value ```false``` so ```isset($this->_access[$permissionName])``` is true, then only  ```return $this->_access[$permissionName]``` will be executed and ```getAuthManager``` won't be

schojniak:
I think we don't understand each other and my english is not enough to explain.

Can you look at code again?

samdark:
I did. I understand the change but don't understand why it's better than current code.

cebe:
@schojniak I understand your change but I think it should not be done. Caching is used to avoid multiple requests to the authmanager. If the authmanager is not configured it does not make sense to cache anything because what we would cache is our own decision for returning false and not the value we received from auth manager.
If can() gets called before an auth manager is configured which may be configured later based on some criteria we would have cached a false for a check that would now become true once there is an auth manager instance available.

schojniak:
You're right. Sorry for taking your time.

cebe:
no problem, thanks for your contribution

Issue #9398, opened 2015-08-13
Add mssql\Schema character_maximum_length support


thejahweh:
We use mainly MSSQL and do have some trouble with the max-length of some fields. We have to add manually the string `max` rules, to avoid ugly DB errors.

So I thought it would be handy to just add this to the `yii\db\mssql\Schema`, so it would be automatically used by Gii to create the models.

thejahweh:
I've tried to do it on my own, but I'm stuck at this point `php build/build dev/app basic` from:
https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md#3-prepare-the-testing-environment

I'm on a windows machine, which leads to trouble with ssh/git things `Permission denied (publickey)`.

PR #9400, opened 2015-08-13
Added MSSQL `character_maximum_length` support #9398

aliechti:
#9398


framework/db/mssql/Schema.php
```diff
@@ -189,6 +189,7 @@ protected function loadColumnSchema($info)
         $column->comment = $info['comment'] === null ? '' : $info['comment'];
 
         $column->type = self::TYPE_STRING;
+        $column->size = $info['character_maximum_length'];
```
SilverFire: According to the [MSDN article](https://msdn.microsoft.com/en-us/library/ms188348.aspx), `character_maximum_length` is maximum length, in characters, for binary data, character data, or text and image data.

Returned values:
-1 for xml and large-value type data. Otherwise, NULL is returned. For more information, see Data Types (Transact-SQL).

We need to handle `-1` properly

thejahweh: Something like?:
```php
if ($column->size == -1) $column->size = null;
```

SilverFire: `$column->size = $info['character_maximum_length'] < 0 ? null : $info['character_maximum_length'];`

SilverFire:
Looks good.

@samdark could you test it on MSSQL?

samdark:
Yes and no. I have MSSQL 2015 but it fails to clean up initial DB in our `master` tests. Gonna fix it first.

samdark:
Fixing done: https://github.com/yiisoft/yii2/commit/619c50d93340305f758430c2c07a2bfb9a7d7f4b

samdark:
Umm... nope. Fixing more :)

samdark:
https://github.com/yiisoft/yii2/commit/615a29f04e92cbcb7214ab36f58816676f98ba5d

samdark:
Still not done...

nineinchnick:
I've got lots of mssql tests fixes in #9773

samdark:
That one changes method signatures...

nineinchnick:
I'm just talking about tests, maybe I should extract those changes into a separate PR.

samdark:
Would be great.

samdark:
Can't verify it w/o tests. Moving to 2.0.8.

thejahweh:
What has to be done here? I've lost track...

samdark:
Tests.

Issue #9074, opened 2015-07-10
With DataGrid option $showHeader=false, "selectedRows" not work.


afernandes465:
In this doc http://www.yiiframework.com/doc-2.0/yii-grid-checkboxcolumn.html we have: 
"Users may click on the checkboxes to select rows of the grid. The selected rows may be obtained by calling the following JavaScript code:
var keys = $('#grid').yiiGridView('getSelectedRows');"

With option "$showHeader=false", this JavaScript always result in null.

Someone cound check please.
Thanks.

Faryshta:
Also I would like to add that such method must work from among the columns not the entire grid in case there are two Checkbox Columns.

Sadly that won't be backwards compatible.

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/3ee89ac42ca103544f9ef65a9532c6e667c768ca

PR #9418, opened 2015-08-14
DataGrid selectedRows fix for $showHeader=false #9074

NekitoSP:
I just fixed issue #9074. 
Problem was due to column not initialized if `$showHeader=false`, cause column initialization code was placed in `renderHeaderCellContent`. And i moved `GridView`  JS initializer to `init();` instead of `run()` because `GridView` should be initialized  before its columns.

### DIFF
--- a/framework/grid/CheckboxColumn.php
+++ b/framework/grid/CheckboxColumn.php
@@ -81,6 +81,15 @@ class CheckboxColumn extends Column
         if (substr_compare($this->name, '[]', -2, 2)) {
             $this->name .= '[]';
         }
+
+        $name = $this->grid->showHeader ? $this->getHeaderCheckBoxName() : NULL;
+        $id = $this->grid->options['id'];
+        $options = json_encode([
+            'name' => $this->name,
+            'multiple' => $this->multiple,
+            'checkAll' => $name,
+        ], JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
+        $this->grid->getView()->registerJs("jQuery('#$id').yiiGridView('setSelectionColumn', $options);");
     }
 
     /**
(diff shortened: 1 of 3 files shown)


framework/grid/GridView.php
```diff
@@ -272,6 +272,11 @@ public function init()
             $this->filterRowOptions['id'] = $this->options['id'] . '-filters';
         }
 
+        $id = $this->options['id'];
+        $options = Json::htmlEncode($this->getClientOptions());
+        $view = $this->getView();
+        $view->registerJs("jQuery('#$id').yiiGridView($options);");
+
```
SilverFire: Could you elaborate, please, why did you move this code from `run()` to init?
I'm not sure that it's a good idea.

framework/grid/CheckboxColumn.php
```diff
@@ -81,6 +81,15 @@ public function init()
         if (substr_compare($this->name, '[]', -2, 2)) {
             $this->name .= '[]';
         }
+
+        $name = $this->grid->showHeader ? $this->getHeaderCheckBoxName() : NULL;
```
SilverFire: `NULL` -> `null`

SilverFire:
@NekitoSP would you like to continue work on this PR?

framework/grid/CheckboxColumn.php
```diff
@@ -81,6 +81,15 @@ public function init()
...
+        $id = $this->grid->options['id'];
+        $options = json_encode([
```
samdark: `Json::encode` helper should be used.

SilverFire:
Moving to the milestone 2.0.8 because the PR is unfinished and @NekitoSP  does not respond

Issue #8293, opened 2015-05-06
Add INSERT INTO SELECT support


arogachev:
Description of syntax can be found [here](http://www.w3schools.com/sql/sql_insert_into_select.asp).

Example (taken from [here](http://stackoverflow.com/questions/30068624/how-to-write-insert-query-in-yii2)):

``` sql
INSERT INTO Customers (CustomerName, Country)
SELECT SupplierName, Country FROM Suppliers;
```

Can't find it in core. `QueryBuilder`'s `insert` method returns concatenated string with `VALUES`:

``` php
return 'INSERT INTO ' . $schema->quoteTableName($table)
    . ' (' . implode(', ', $names) . ') VALUES ('
    . implode(', ', $placeholders) . ')';
```

cebe:
Well, w3schools is not a good resource to refer to for a specific syntax.

Need to check the manuals of all Yii supported DBMS, here are some:

- MySQL: https://dev.mysql.com/doc/refman/5.5/en/insert-select.html
- Postgres: http://www.postgresql.org/docs/current/static/sql-insert.html
- ...

mdmunir:
How if this feature added to `batchInsert`?
```php
     * @param string $table the table that new rows will be inserted into.
     * @param array $columns the column names
     * @param array|Query $rows the rows to be batch inserted into the table
     * @return $this the command object itself
     */
    public function batchInsert($table, $columns, $rows)
    {
```

xiongxinag:
if (is_array($info)) {..........}
    if (is_array($result) && isset($result[0])) {........}


    if (isset($info[0])) {..........}
    if (isset($result[0])) {........}

cebe:
what do you want to say?

JQueryBoy14:
there is an issue regarding this added function.

I tried it like **insert into ... select ... from ... where ...**

An error says **[Microsoft][ODBC Driver 11 for SQL Server]COUNT field incorrect or syntax error**

I tried removing **where** condition and it works

samdark:
Please create a separate issue and describe everything in detail: how to reproduce the issue, what SQL is generate, what SQL is expected. Thanks.

JQueryBoy14:
Please see #13807

PR #9420, opened 2015-08-14
Basic syntax of 'INSERT INTO ... SELECT' statement was added (#8293).

voroks:
Method "insert()" accepts instance of \yii\db\Query for second parameter as well as array.
Column names (or aliases, if exists) of SELECT statement became column names for INSERT statement. So, support of basic usage of "INSERT INTO ... SELECT" statement (as in issue #8293) was added.


samdark:
How about non-MySQL?

voroks:
PostgreSQL - OK
SQLite - OK
Unfortunately, I can't perform test with using other DBs.

framework/db/Command.php
```diff
@@ -418,7 +418,8 @@ public function queryColumn()
      * Note that the created command is not executed until [[execute()]] is called.
      *
      * @param string $table the table that new rows will be inserted into.
-     * @param array $columns the column data (name => value) to be inserted into the table.
+     * @param array|\yii\db\Query $columns the column data (name => value) to be inserted into the table or instance
+     * of \yii\db\Query to perform INSERT INTO ... SELECT SQL statement.
```
cebe: Need a note that the `Query` can be passed since version 2.0.11.

framework/db/QueryBuilder.php
```diff
@@ -123,7 +123,8 @@ public function build($query, $params = [])
      * The method will properly escape the table and column names.
      *
      * @param string $table the table that new rows will be inserted into.
-     * @param array $columns the column data (name => value) to be inserted into the table.
+     * @param array|\yii\db\Query $columns the column data (name => value) to be inserted into the table or instance
+     * of \yii\db\Query to perform INSERT INTO ... SELECT SQL statement.
```
cebe: Need a note that the Query can be passed since version 2.0.11.

cebe (changes requested):
test seems to pass on travis so this looks good to me. A CHANGELOG is needed and it should be rebased on current master.

cebe:
@voroks are you willing to work on this further?

framework/db/mysql/QueryBuilder.php
```diff
@@ -195,30 +195,42 @@ public function insert($table, $columns, &$params)
         }
         $names = [];
         $placeholders = [];
-        foreach ($columns as $name => $value) {
-            $names[] = $schema->quoteColumnName($name);
```
cebe: no need to check `=== true`. Change this to:

```php
if ($columns instanceof \yii\db\Query) {
```

framework/db/oci/QueryBuilder.php
```diff
@@ -178,30 +178,42 @@ public function insert($table, $columns, &$params)
         }
         $names = [];
         $placeholders = [];
-        foreach ($columns as $name => $value) {
-            $names[] = $schema->quoteColumnName($name);
```
cebe: no need to check `=== true`. Change this to:

```php
if ($columns instanceof \yii\db\Query) {
```

framework/db/pgsql/QueryBuilder.php
```diff
@@ -245,17 +245,21 @@ public function update($table, $columns, $condition, &$params)
...
-     * @param array $columns the column data (name => value) to be saved into the table.
+     * @param array|\yii\db\Query $columns the column data (name => value) to be saved into the table or instance
+     * of \yii\db\Query to perform INSERT INTO ... SELECT SQL statement.
+     * Passing of \yii\db\Query is available since version 2.0.11.
      * @return array normalized columns
      * @since 2.0.9
      */
     private function normalizeTableRowData($table, $columns)
     {
-        if (($tableSchema = $this->db->getSchema()->getTableSchema($table)) !== null) {
```
cebe: no need to check `=== false`. Also better not wrap the whole function in the if but return early, so change this to:

```php
if ($columns instanceof \yii\db\Query) {
    return $columns;
}
```

cebe (changes requested):

tests/framework/db/CommandTest.php
```diff
@@ -321,6 +321,51 @@ public function testInsert()
...
+                'name' => 'test',
+                'address' => 'test address',
+            ]
+        )->execute();
+
+	$query = new \yii\db\Query();
+	$query->select([
+                '{{customer}}.email as name',
+                'name as email',
+                'address',
```
cebe: could you add the alias syntax `'alias' => 'column'`. Seems this is not supported by your implementation right now.

> When the columns are specified as an array, you may also use array keys as the column aliases (if a column
>  does not need alias, do not use a string key).
> http://www.yiiframework.com/doc-2.0/yii-db-query.html#select%28%29-detail

sorry, have not noticed this earlier.

cebe (changes requested):

framework/db/QueryBuilder.php
```diff
@@ -157,23 +159,56 @@ public function insert($table, $columns, &$params)
...
-            . (!empty($placeholders) ? ' VALUES (' . implode(', ', $placeholders) . ')' : ' DEFAULT VALUES');
+            . (!empty($placeholders) ? ' VALUES (' . implode(', ', $placeholders) . ')' : $values);
+    }
+
+    /**
+     * Prepare select-subquery and field names for INSERT INTO ... SELECT SQL statement.
+     *
+     * @param \yii\db\Query $columns Object, which represents select query
+     * @param \yii\db\Schema $schema Schema object to qoute column name
+     * @return array
```
samdark: Need `@since 2.0.11`

samdark (approved):

framework/db/QueryBuilder.php
```diff
@@ -157,23 +159,58 @@ public function insert($table, $columns, &$params)
...
-            $names[] = $schema->quoteColumnName($name);
-            if ($value instanceof Expression) {
-                $placeholders[] = $value->expression;
-                foreach ($value->params as $n => $v) {
-                    $params[$n] = $v;
+        $values = ' DEFAULT VALUES';
+        if ($columns instanceof \yii\db\Query) {
+            list($names, $values) = $this->prepareInsertSelectSubQuery($columns, $schema);
+        } else {
+            foreach ($columns as $name => $value) {
```
SilverFire: @yiisoft/core-developers how about moving this part into a separate method to make `insert` simpler to read?

samdark: I'm OK with that.

dynasource: agree. Placeholders are not always defined. 
I would expect a separate method for the creation of an "insert" statement with all the relevant arguments.

framework/db/QueryBuilder.php
```diff
@@ -157,23 +159,58 @@ public function insert($table, $columns, &$params)
...
-                $placeholders[] = $phName;
-                $params[$phName] = !is_array($value) && isset($columnSchemas[$name]) ? $columnSchemas[$name]->dbTypecast($value) : $value;
             }
         }
 
         return 'INSERT INTO ' . $schema->quoteTableName($table)
             . (!empty($names) ? ' (' . implode(', ', $names) . ')' : '')
-            . (!empty($placeholders) ? ' VALUES (' . implode(', ', $placeholders) . ')' : ' DEFAULT VALUES');
+            . (!empty($placeholders) ? ' VALUES (' . implode(', ', $placeholders) . ')' : $values);
+    }
```
dynasource: the creation of the insert statement could be done a lot cleaner in a separate method returning a <<<SQL block

framework/db/oci/QueryBuilder.php
```diff
@@ -178,30 +178,35 @@ public function insert($table, $columns, &$params)
...
+                    $names[] = $schema->quoteColumnName($name);
+                    $placeholders[] = 'DEFAULT';
+                }
             }
         }
 
         return 'INSERT INTO ' . $schema->quoteTableName($table)
             . (!empty($names) ? ' (' . implode(', ', $names) . ')' : '')
-            . (!empty($placeholders) ? ' VALUES (' . implode(', ', $placeholders) . ')' : ' DEFAULT VALUES');
+            . (!empty($placeholders) ? ' VALUES (' . implode(', ', $placeholders) . ')' : $values);
```
dynasource: this should be refactored to a solution in which with https://github.com/yiisoft/yii2/pull/9420/files#diff-e2271b5ca7ffbb33d4f2bfa5ed2e94f4R181 is reused

tests/framework/db/CommandTest.php
```diff
@@ -321,6 +321,136 @@ public function testInsert()
         ], $record);
     }
 
+    /**
+     * Test INSERT INTO ... SELECT SQL statement
+     */
+    public function testInsertSelect()
+    {
```
dynasource: please split these tests into separate selfcontained methods

dynasource (changes requested):

samdark:
Merged https://github.com/yiisoft/yii2/commit/25f08afc96035b517d6a7be13b28bb09514bfc39. Thank you!

PR #9422, opened 2015-08-15
Fixed link in REST doc to point to correct page

sblaut:
Filtering out fields is not described in the Response Formatting section but in the Resources section.


samdark:
Thanks!

PR #9430, opened 2015-08-15
According to: Composer installation guide

ravshanium:
According to: Composer installation guide: https://github.com/huebs/composer-asset-plugin/blob/master/Resources/doc/index.md#installation


samdark:
Thanks!

Issue #8582, opened 2015-05-27
data-confirm and data-method with pjax


xskif:
Is there any way to do this?
In rails `remote` option work together with `method` and `confirm` options, in yii they not.
Maybe if we change `handleAction` behavior when `confirm` specified but `method`not:
```js
//--------------------------- handleAction --------------------------
            if (method === undefined) {
                if (action && action != '#') {
                    // window.location = action; // <- delete this
                    return false; // <- this one

//--------------------------- initDataMethods -----------------------

            if (message !== undefined) {
                pub.confirm(message, function () {
                    pub.handleAction($this);
                });
            } else {
                pub.handleAction($this);
            }

            if (method !== undefined) {
                event.stopImmediatePropagation();
                return false;
            }
            return true;
```

And when `method` specified we can add data-pjax attribute like that:
```js
if (data.pjax !== undefined) {
    $form.attr('data-pjax', data.pjax);
}
```

armpogart:
I have the same issue. We have anchor with `data-confirm` attribute to show prompt, and `data-confirm` breaks pjax request.

xskif:
@armpogart , I was found temporary solution:
You can wrap links with `data-method` by `form` with `data-pjax`. `data-confirm` also works fine if you have `data-method` on the link

nkovacs:
Aside from the form missing the data-pjax attribute (so the default pjax form selector doesn't pick it up), the other problem is that it's appended to body.
The default form selector in the Pjax widget is `'#' . $id . ' form[data-pjax]'`, so it won't find the form, since it's outside the widget's container.

nkovacs:
I have not tested this thoroughly: https://github.com/nkovacs/yii2/commit/c636e506bae6198727fb3359a2d13a8c08d2d63c
You have to add data-form-pjax to the link, so it's opt-in. Since `data-pjax="0"` is used to disable pjax on links, I didn't want to overload it with this, though there wouldn't really be a conflict (`data-pjax` would mean enable pjax on the generated form, `data-pjax="0"` would mean disable pjax).
It also inserts the invisible form after the element instead of append it to the body. I don't know what the side-effects of this change are.

xskif:
@nkovacs , i think this:
```php
if (data.pjax !== undefined) {
    $form.attr('data-pjax', data.pjax); // there are dynamic value
}
```

better than this:

```php
if ($e.data('form-pjax') !== undefined) {
[...]

Issue #8301, opened 2015-05-06
GridView with PJAX don't work AJAX delete


katzz0:
I have seen some many posts about this problem,  but all of they don't introduce clarity.  I am use solution from box. GridView + Pjax and delete action.  When I will try to delete item the Pjax reload all page,  when I do redirect in my delete acton.  What I do wrong? 

Yii 1.1 send ajax request and then update GridView by update js method. I don't see similar behavior in Yii 2.

cebe:
So you expect Pjax to follow the redirect of the delete action (assuming it redirects to the index page that contains the gridview) and reload the grid only instead of following the redirect for the whole page?

katzz0:
I think this is simplest and straight way. But the PJAX headlines will be lost after the browser will be redirected from delete action to the index action. The index action thinking that redirect request is usually not PJAX (or AJAX) request and return full web page. Sure, PJAX will reload full web page.

And other question: what about filter and page params? Because they will be lost after redirect.

P.S. I think true way use AJAX request to delete action and then refresh current page by pjax update.

What is the philosophy of Yii2 about buttons with AJAX action in GridView?

katzz0:
I am using this enhancement of yii.handleAction method in yii.js:

if (isAjax) {
        $.ajax({
            url: $form.attr('action'),
            method: $form.attr('method'),
            data: $form.serialize(),
            success: function (data) {
                if (data) {
                    var $pjaxContainer = $e.closest('.pjax');
                    if ($pjaxContainer.length && $.pjax.reload) {
                        $.pjax.reload('#' + $pjaxContainer.attr('id'), {
                            url: window.location.href
                        });
                    } else {
                        window.location.reload();
                    }
                }
            },
            error: function (jqXHR, textStatus, errorThrown) {
                alert('Error was thrown while AJAX request: ' + textStatus);
            }
        });
    } else {
        $form.trigger('submit');
    }

It's requires that pjax container has .pjax class and button or link has data-ajax="1" attribute.

yurii-github:
this?
use yii\widgets\Pjax;
Pjax::begin();
echo GridView::widget([....]
Pjax::end();?>

Issue #6363, opened 2014-12-03
PJAX form submit and yii confirm don't mix.


pcfreak30:
Right now if you use ActionColumn and set data-pjax to 1, then it doesn't do an ajax submit like in v1.1.

The yii.js file needs to be modified so that ```$form = $('<form method="' + method + '" action="' + action + '"></form>');``` has ```data-pjax```.

The form needs to be inserted as well after the called element, not the body, so change ```$form.hide().appendTo('body');``` to ```$e.after($form.hide());```

This is so the jquery selector defined in Pjax.php on line 175 get matched.

There also should be a better way to handle this as you must define pjax history to be off, or the replaceState js function does a window.location.reload().

In summary, the ActionColumn should support ajax actions via some means. This was a mess to figure out...

germansokolov13:
Same problem happens to me. `data-confirm` attribute on a link doesn't let pjax-call to fire in a gridview. After you click OK in a browser-provided confirm dialog, full page reload occurs. If you remove `data-confirm` it starts to work via pjax.

PR #9432, opened 2015-08-16
Fix to get data methods and pjax working together

derekisbusy:
This is a fix for the issue of pjax links/buttons not working when using data methods. This fix also requires a change in the Pjax widget that adds the attribute data-pjax-container to the container element so that it can be found via JavaScript. The pjax container can be defined in two ways, either by setting the data-pjax-container="#someSelector" in the link/button attributes, or else it will find the closest element with data-pjax-container="". So the HTML attribute 'data-pjax-container' is used in two ways: if the value is blank than that means the element is a pjax container, if the value is set then that means it's a selector pointing to a pjax container. If neither are set and the container is not found it will default to the body tag.

### DIFF
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -143,16 +143,37 @@ yii = (function ($) {
          *
          * @param $e the jQuery representation of the element
          */
-        handleAction: function ($e) {
+        handleAction: function ($e,event) {
             var method = $e.data('method'),
                 $form = $e.closest('form'),
                 action = $e.attr('href'),
-                params = $e.data('params');
+                params = $e.data('params'),
+                pjax = $e.data('pjax'),
+                pjaxContainer;
 
+            if(pjax!==undefined && $.support.pjax) {
+                if($e.data('pjax-container'))
+                    pjaxContainer = $e.data('pjax-container');
+                else
+                    pjaxContainer = $e.closest('[data-pjax-container=""]');
+                // default to body if pjax container not found
+                if(!pjaxContainer.length)
+                    pjaxContainer = $('body');
+            }
+            
             if (method === undefined) {
                 if (action && action != '#') {
-                    window.location = action;
+                    if(pjax!==undefined && $.support.pjax) {
+                        $.pjax.click(event, {container: pjaxContainer});
+                    } else {
+                        window.location = action;
+                    }
                 } else if ($e.is(':submit') && $form.length) {
+                    if(pjax!==undefined && $.support.pjax) {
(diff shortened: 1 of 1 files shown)


derekisbusy:
Fixes these issues:

https://github.com/yiisoft/yii2/issues/8582
https://github.com/yiisoft/yii2/issues/8301
https://github.com/yiisoft/yii2/issues/6363

derekisbusy:
I also forgot to mention that when using pjax with data method you have to set data-pjax attribute in the link element. This could be removed if we change it to check for $e.parents('[data-pjax-container=""]').length instead. Since the Pjax widget would now add the data-pjax-container attribute to the container element you could use that to check if the link is contained within a pjax container. However depending on the pjax settings not all links may be pjax enabled so might be best to leave it the way it is. For example right now you would set a link like this...

```php
Html::a('My Link','/controller/action',[
            'data-confirm'=>Yii::t('common','Are you sure?'),
            'data-method'=>'post',
            'data-pjax'=>'', // define a pjax link
        ]);
```

derekisbusy:
Another issue I found with this is that when pjax is submitted it uses to default pjax settings. This is because there's no way to get the settings from a pjax container after it is created. There are two ways to fix this, either add the pjax settings to the container as data attributes, or update pjax so that the options for each pjax container can be referenced via something like $('.container').pjax('settings').

elastic-od:
I have just tested your fix on delete button in GridView and I faced an issue of pjax pushs new state even if I set  'enablePushState' to false. Can you give an example how to handle this issue? Here is my code:
```
\yii\widgets\Pjax::begin([
    'id' => 'subject-channels-grid',
    'timeout' => false,
    'enablePushState' => false,
]);
```
```
'buttons' => [
    'delete' => function ($url, $model, $key) {
        $options = [
            'title' => Yii::t('yii', 'Delete'),
            'aria-label' => Yii::t('yii', 'Delete'),
            'data-confirm' => Yii::t('yii', 'Are you sure you want to delete this item?'),
            'data-method' => 'post',
            'data-pjax' => '1',
            'data-pjax-container' => '#subject-channels-grid',
        ];
        return Html::a('<span class="glyphicon glyphicon-trash"></span>', $url, $options);
    },
]
```

derekisbusy:
@elastic-od Please use the new patch here #9952. They never merged this patch and ran into problems with git so had to re-apply them. 
@samdark Please disregard this pull request and merge the new one #9952. It fixes most the issues people are having with Pjax and data methods. Thanks!

PR #9437, opened 2015-08-16
fix #9108 ( Negative number results in no formatting when using asSize or asShortSize )

### LINKED ISSUE #9108: Negative number results in no formatting when using asSize or asShortSize
When using formatter to format byte values on a negative value, it does not provide a value with the correct suffix. Instead it returns the full number in bytes - without the suffix.

I believe it has to do with this call in **yii\i18n\Formatter @ line 1137**:

```php
            if ($value < $this->sizeFormatBase) {
                break;
            }
```

Maybe throwing an `abs()` call on the `$value` variable to satisfy this check can solve the issue?

nxnx:
fix #9108

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -1134,7 +1134,7 @@ class Formatter extends Component
 
         $position = 0;
         do {
-            if ($value < $this->sizeFormatBase) {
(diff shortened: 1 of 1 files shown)


cebe:
Merged, thanks!

PR #9440, opened 2015-08-17
Removed yii\db\Schema from use section

arogachev:
Declaration of yii\db\Schema is not needed by default. Migration can be related with changing data or renaming column, so user should explicitly add it when it's needed.

And by the way use classes were not it alphabetical order.


arogachev:
And by the way it also is not needed because of new way of handling fields.

samdark:
Thanks!

Issue #9337, opened 2015-08-06
Add ability to use \yii\db\Expression for SchemaBuilder::defaultValue()


alxkolm:
Add ability to use `\yii\db\Expression` for `SchemaBuilder::defaultValue()`.
Particularly this need for assign default value to TIMESTAMP column:
`$this->timestamp()->defaultValue(new \yii\db\Expression('CURRENT_TIMESTAMP'))`

klimov-paul:
Perhaps we should introduce another method `defaultExpression()` for expressions to simplify syntax:
```php
$this->timestamp()->defaultExpression('CURRENT_TIMESTAMP')
```

cebe:
:+1:

yupe:
+1

bajabob:
+1

PR #9447, opened 2015-08-17
Added defaultExpression() to ColumnSchemaBuilder

### LINKED ISSUE #9445: New migration style (defaultValue('NULL')
As far as I can see in new migration style, there is no ability to define `NULL` default.

* `defaultValue('NULL')` will create `DEFAULT 'NULL'` which is not right
* `defaultValue(null)` will ignore the specified null

chuprik:
Fix for https://github.com/yiisoft/yii2/issues/9337 and... I'm not sure that this is correct but can be used for fix #9445:

``` php
$this->defaultExpression(new Expression('NULL'))
```

### DIFF
--- a/framework/db/ColumnSchemaBuilder.php
+++ b/framework/db/ColumnSchemaBuilder.php
@@ -105,7 +105,7 @@ class ColumnSchemaBuilder extends Object
     }
 
     /**
-     * Specify the default DB Expression for the column.
+     * Specify the default value for the column in terms of a [[Expression|SQL expression]].
      * @param Expression $default the default value.
      * @return $this
      */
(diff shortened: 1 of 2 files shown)


framework/db/ColumnSchemaBuilder.php
```diff
@@ -105,6 +105,17 @@ public function defaultValue($default)
     }
 
     /**
+     * Specify the default DB Expression for the column.
```
cebe: better: `Specify the default value for the column in terms of a [[Expression|SQL expression]].`

kotchuprik:
thanks! and sorry for the typo.

mikk150:
what is the status on this one?

samdark:
Merged https://github.com/yiisoft/yii2/commit/e94810551c4ce2c7c5d59dbdcbd46a653ad601bb. Thanks!

Issue #9455, opened 2015-08-18
After column for migration command


hulyman:
In ColumnSchemaBuilder not enough method for set AFTER column command.

bryglen:
please add a parameter such as `$this->string(255)->custom('UNSIGNED')->notNull()->custom('AFTER status')` so it will generate by order `VARCHAR(255) UNSIGNED NOT NULL AFTER status`

arogachev:
@samdark, @cebe What do you think about it? I think it's better to create separate MysqlColumnSchemaBuilder as it's done with Oracle and add custom `after()` method. Also I thinks it will be useful to have possibility to use `FIRST` statement for inserting column to the very beginning, before second column. Or define `custom()` method for insertion custom code?

samdark:
I don't like the idea of having different API for different database types.

bryglen:
we can create `custom()` sort of like that to handle different methods for each database

arogachev:
Yes, creating additional property and method for each statement is indeed not good idea.

Faryshta:
wasn't this closed by #10494 ?

samdark:
Yes, it was. Thanks!

PR #9456, opened 2015-08-18
Add first() and after($column) methods for ColumnSchemaBuilder

dudareviv:
Add methods for ColumnSchemaBuilder:
- first -- adds a `FIRST` constraint to the column.
- after  -- adds a `AFTER` constraint for the column and the column name.

issue #9455

### DIFF
--- a/framework/db/ColumnSchemaBuilder.php
+++ b/framework/db/ColumnSchemaBuilder.php
@@ -31,6 +31,11 @@ class ColumnSchemaBuilder extends Object
      */
     protected $length;
     /**
+     * @var boolean|string the column insert position (column name). If this is `true`, a `FIRST` constraint will be
+     * added. If this is a string, a `AFTER [[position]]` constraint will be added.
+     */
+    protected $position;
+    /**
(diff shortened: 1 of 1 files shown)


samdark:
Is it supported by all databases we have in framework?

hulyman:
It seems like a support exists only in mysql.
PostgreSQL doesn't support yet: https://wiki.postgresql.org/wiki/Alter_column_position
Sqlite doesn't support: http://stackoverflow.com/questions/4253804/insert-new-column-into-table-in-sqlite
For ms sql it support in mssql server 2016: https://msdn.microsoft.com/en-us/library/aa337556.aspx
Oracle: http://www.dba-oracle.com/t_change_column_order_within_oracle_table.htm

hulyman:
Add `AFTER` constraint need sometimes. What to do?

samdark:
That would be MySQL-specific things. Not sure if we're going to support that.

samdark:
Closing since this pull request can't be merged because of not being compatible with all DBs supported.

PR #9475, opened 2015-08-19
Allow Expression as parameter in where case for QueryBuilder (bug #9452)

### LINKED ISSUE #9452: QueryBuilder does not add params for Expression in where case
It does not work if I pass params to new Expression()
```php
AR::find()->where(new Expression(':item = ANY (array)', ['item' => 'value']))->all();
```
but it works if I pass params to where()
```php
AR::find()->where(new Expression(':item = ANY (array)'), ['item' => 'value'])->all();
```
sometimes it is not comfortable when I pass where expression as parameter of function or method
```php
function getAll($where) {
    return AR::find()->where($where)->all();
}

// it does not work
$models = getAll(new Expression(':item = ANY (array)', ['item' => 'value']));
```
This can be fixed if to add one more condition in ```QueryBuilder::buildCondition()```
https://github.com/yiisoft/yii2/blob/master/framework/db/QueryBuilder.php#L910
```php
if ($condition instanceof Expression) {
    foreach ($condition->params as $n => $v) {
        $params[$n] = $v;
    }
    return (string) $condition;
}
```

Also it requires for ```QueryBuilder::buildAndCondition()``` and ```QueryBuilder::buildNotCondition()```
https://github.com/yiisoft/yii2/blob/master/framework/db/QueryBuilder.php#L975
https://github.com/yiisoft/yii2/blob/master/framework/db/QueryBuilder.php#L1004
to replace ```if (is_array($operand)) {``` on ```if (is_array($operand) || $operand instanceof Expression) {```

mnvx:
Fixed #9452. Allowed Expression as parameter in where case for QueryBuilder

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.7 under development
 -----------------------
 
+- Bug #9452: Fixed QueryBuilder does not add params for Expression in where case (iriscrm)
 - Bug #8723: Fixed `yii\helpers\VarDumper::export()` unable to export circle referenced objects with `Closure` (klimov-paul)
 - Bug #9108: Negative number resulted in no formatting when using `Formatter::asSize()` or `Formatter::asShortSize` (nxnx, cebe)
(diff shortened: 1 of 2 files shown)


SilverFire:
Fixed in https://github.com/yiisoft/yii2/commit/bea90e122837cc374dd1ce0199b0a146bfb05e48
Thank you!

PR #9484, opened 2015-08-20
fixes #9314, rbac\BaseManager::getPermissionsByUser also returns directly assigned permissions

### LINKED ISSUE #9314: RBAC DbManager getPermissionsByUser cannot get all permissions
I allot permission and role to user （insert into auth_assignment）
such as 

```
item                user_id
permission1    1
permission2    1
role1               1                           
role2               1                           
```
 
`role1` include `permission3` and `role2` include `permission4` `permission5`

when i try to get all permissions by the method named `getPermissionsByUser` of `DbManager`
it just return the `permission3` `permission4` `permission5`

i viewed the source,in my opinion,This is because the code below：

```php
protected function getChildrenRecursive($name, $childrenList, &$result)
    {
        if (isset($childrenList[$name])) {
            foreach ($childrenList[$name] as $child) {
                $result[$child] = true;
                $this->getChildrenRecursive($child, $childrenList, $result);
            }
        }
      // when i add the below code,i get all permissions exactly
       else {
            $result[$name] = true;
        }
    }
```

hesna:
fixes #9314 
It's my first contribution, please let me know if there are any mistakes or missteps. tests has been adjusted to reflect the changes and are all green.

### DIFF
--- a/framework/rbac/DbManager.php
+++ b/framework/rbac/DbManager.php
@@ -501,6 +501,39 @@ class DbManager extends BaseManager
             return [];
         }
 
+        $directPermission = $this->getDirectPermissionsByUser($userId);
+        $inheritedPermission = $this->getInheritedPermissionsByUser($userId);
+
+        return array_merge($directPermission, $inheritedPermission);
+    }
+
+    /**
+     * Returns all permissions that are directly assigned to user.
+     * @param string|integer $userId the user ID (see [[\yii\web\User::id]])
+     * @return Permission[] all direct permissions that the user has. The array is indexed by the permission names.
+     */
+    protected function getDirectPermissionsByUser($userId)
+    {
+        $query = (new Query)->select('b.*')
+            ->from(['a' => $this->assignmentTable, 'b' => $this->itemTable])
+            ->where('{{a}}.[[item_name]]={{b}}.[[name]]')
+            ->andWhere(['a.user_id' => (string) $userId])
+            ->andWhere(['b.type' => Item::TYPE_PERMISSION]);
+
+        $permissions = [];
+        foreach ($query->all($this->db) as $row) {
+            $permissions[$row['name']] = $this->populateItem($row);
+        }
+        return $permissions;
+    }
+
+    /**
+     * Returns all permissions that the user inherits from the roles assigned to him.
+     * @param string|integer $userId the user ID (see [[\yii\web\User::id]])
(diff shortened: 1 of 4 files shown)


samdark:
Merged with some minor adjustments. Well done. Thanks!

https://github.com/yiisoft/yii2/commit/5569eaae061a8adbb222489c532c320baa673e87

hesna:
Thank you! :wink:

framework/rbac/DbManager.php
```diff
@@ -501,6 +501,39 @@ public function getPermissionsByUser($userId)
...
+
+        return array_merge($directPermission, $inheritedPermission);
+    }
+
+    /**
+     * Returns all permissions that are directly assigned to user.
+     * @param string|integer $userId the user ID (see [[\yii\web\User::id]])
+     * @return Permission[] all direct permissions that the user has. The array is indexed by the permission names.
+     */
+    protected function getDirectPermissionsByUser($userId)
```
razvanphp: any reason why this method is protected? It would've been easier to implement a form for changing the permissions too.

hesna: Well it's been three years and I don't recall the details, but the problem was that the `getPermissionsByUser()` method did not return directly assigned permissions, only returning permissions connected to a user via assigned roles.

Suggested solution aimed to resolve that issue, having `getDirectPermissionsByUser()` as public method would've been a new feature (maybe a useful one) that I'm not sure if maintainers would agree with at the time.

And I don't get the form suggestion in this context. can you elaborate please?

razvanphp: Thanks for the explanation. We have implemented a profile page for users, where admins can assign roles to users. We also added the possibility to assign extra permissions to users, for a more granular access schema, for example giving access to team members from another team just to view items, but not edit them.

Problem is, if we use `getPermissionsByUser()` we fill the extra permissions input with the inherited ones too, so on second form submit, all those will be directly assigned to user.

Right now I extended the DbManager class to make `getDirectPermissionsByUser()` public so we can only display and edit those ones in the form.

I can open a PR if maintainers have nothing against.

hesna: We had the exact same problem! and made a workaround just like you. didn't cross my mind to make these methods public in framework's core though. 
please let me know when you created the new issue, it'll have a thumb up from me :)

PR #9485, opened 2015-08-20
\yii\web\Response::sendHeaders() does now send status header last

### LINKED ISSUE #9412: Response->sendHeaders() patch to allow sending a Location header with a 202 status
The current sendHeaders method in the Response class does not handle a php bug.

The fix is very simple and the explanation is in the comments below:

    /**
     * Sends the response headers to the client
     */
    protected function sendHeaders()
    {
        if (headers_sent()) {
            return;
        }
        $statusCode = $this->getStatusCode();
        header("HTTP/{$this->version} $statusCode {$this->statusText}");
        if ($this->_headers) {
            $headers = $this->getHeaders();
            foreach ($headers as $name => $values) {
                $name = str_replace(' ', '-', ucwords(str_replace('-', ' ', $name)));
                // set replace for first occurrence of header but false afterwards to allow multiple
                $replace = true;
                foreach ($values as $value) {
                    header("$name: $value", $replace);
                    $replace = false;
                }
            }
        }
        
        /*
         * Patch to allow sending Location header with 202 status
         * 
         * If the Location header is set
         * And the status code is not 201 or 3xx
         * PHP will automatically set the status code to 302  
         * 
         * However, the status code 202 would also typically pass a Location header.
         * 
         * From: http://www.php.net/manual/en/function.header.php
         * 		...
         * 		The second special case is the "Location:" header. 
         * 		Not only does it send this header back to the browser, 
         * 		but it also returns a REDIRECT (302) status code to the browser 
         * 		unless the 201 or a 3xx status code has already been set.
         * 		...
         * 
         * To resolve this problem PHP does allow forcing the status:
         * 
         * From: http://www.php.net/manual/en/function.header.php
         * 		void header ( string $string [, bool $replace = true [, int $http_response_code ]] )
         * 		...
         * 		http_response_code
         * 			Forces the HTTP response code to the specified value. 
         * 			Note that this parameter only has an effect if the string is not empty.
         * 
         * So we need to force the status AFTER all the headers are sent.
         * 
         * However, we do NOT want to force it if it is not the 202 status.  
         * This is to allow for the default PHP behavior
         * 
         */
    	if ($statusCode == 202) {
			header("HTTP/{$this->version} $statusCode {$this->statusText}", true, $statusCode);
		}
                
        $this->sendCookies();
    }

kidol:
Proposal
Fixes #9412

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -8,6 +8,7 @@ Yii Framework 2 Change Log
 - Bug #9108: Negative number resulted in no formatting when using `Formatter::asSize()` or `Formatter::asShortSize` (nxnx, cebe)
 - Bug #9415: Fixed regression in 2.0.6 where on Oracle DB `PDO::ATTR_CASE = PDO::CASE_LOWER` did not work anymore (cebe)
 - Bug #9442: Fixed `yii\db\Migration::renameTable()` caused fatal error when using SQLite driver (fetus-hina)
+- Enh #9412: `yii\web\Response::sendHeaders()` does now set the status header last which negates certain magic PHP behavior regarding the `header()` function (nd4c, kidol)
 - Chg #9369: `Yii::$app->user->can()` now returns `false` instead of erroring in case `authManager` component is not configured (creocoder)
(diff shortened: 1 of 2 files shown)


cebe:
Thank you!

PR #9487, opened 2015-08-20
Update ActiveDataProvider.php

fcaldarelli:
Added formatModelOutput member callable function and overwritten getModels() to allow customization when output single model. This change allow to customize each model output, in particular in rest api.

### DIFF
--- a/framework/data/ActiveDataProvider.php
+++ b/framework/data/ActiveDataProvider.php
@@ -78,6 +78,13 @@ class ActiveDataProvider extends BaseDataProvider
      */
     public $db;
 
+    /**
+     * @var callable customize output of each single model
+     * 
+     * Callable is in the form function($id, $model), where id is the counter of the item
+     * 
+     */ 
+    public $formatModelOutput = null;
 
     /**
      * Initializes the DB connection component.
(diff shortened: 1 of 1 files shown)


FabrizioCaldarelli:
Other example, that it would be difficult using fields() and extraFields() in rest api using ActiveDataProvider to return data:

```
$dataProvider = new \yii\data\ActiveDataProvider([
	'query' => $query,
	'formatModelOutput' => function($id, $model) {

		return array_merge(
			$model->attributes, 
			[ 
				'relation1' => array_merge(
					$model->relation1->attributes, 
					[
						'innerRel' => $model->relation1->relation2->attributes
					]
				)
			]
		);
	}
]);		
```

cebe:
can you explain in more detail why the use of fields() and Serializer do not work in your case?

FabrizioCaldarelli:
Sure. The fields() and extraFields() work well, but I want to avoid that third parties call my rest api url passing the 'expand' extra parameter, such as:

http://hostname/api/news/index?expand=userProfile

I want instead simply call 

http://hostname/api/news/index

and inside controller's action get userProfile relation for each news.

Think at this structure:
'News' -> 'User' -> 'Profile image', where all these three element are object (relation).

If I want that the news/index return this structure for each news, I've not possibility to use ActiveDataProvider directly, because I don't know how render this structure. I can also made this returning a custom array of object, but then i have to implement pagination myself.

cebe:
There is no need to use expand. you can add the relation to the fields() method and they should be included by default.

FabrizioCaldarelli:
But I need the ability to change this behavior based on action of the Controller.

In some case I need to send extra data, in other no. I know that i can check controller action also in fields() method, but I don't want to control this inside the model.

cebe:
Can you explain your use case in a new issue? I am sure this PR is not the right fix so I'd close it. This should be solved on the level of fields() and Serializer.

FabrizioCaldarelli:
Example. This structure: 'User' -> 'Track' -> 'PointOfTrack' 
An user can have many tracks.
A track can have many PointOfTracks.

In user/index action I want to get 'User' -> 'Track' structure, missing 'PointOfTrack' because there could be thousand of records.
But in user/view action I want to get the complete structure, with 'PointOfTrack' data.

If use use fields() methods of the model, I have to handle the output or not of the PointOfTrack struture based on action of controller that I'm calling. But I think that make this check at model level is not right and it should be made at controller level.

Issue #9419, opened 2015-08-14
ActiveRecord updateCounters fails when target field is Null


dbd5:
Hello,

It seems that updateCounters fails when the target field is null.

To reproduce, create a counter field like 'view_counter' in any table, then try to update field when form is viewed with code like this;

public function actionView($id)
{
    $model = $this->findModel($id);
    $model->updateCounters(['view_count' => 1]);

    return $this->render('view', [
        'model' => $model,
    ]);
}

No update occurs and updateCounters returns false.

My theory is that Yii attempts UPDATE post SET view_count = view_count + 1 WHERE id = 100 which evaluates as SET view_count = Null + 1 WHERE id = 100 thus failing

Not sure if this requires a fix or clearer error message but maybe updating AR documentation with a note suggesting initializing counter fields to zero before any call to updateCounters will suffice?

Regards,

Adam

samdark:
Closed since current behavior is valid. See comments in a pull request.

PR #9488, opened 2015-08-20
9419 fix null column cant be incremented

RusAlex:
#9419 fix


klimov-paul:
Using `null` field for the counter seems to be incorrect usage.
The fix make functionality too complicated for the simple case.

cebe:
I agree that a counter field should be defined as `int NOT NULL DEFAULT 0` instead of allowing NULL. What would be the semantics of a NULL value in the DB?

RusAlex:
In this way core should throw exception or warn user about updating
counters for NULL column.

nineinchnick:
`null` in databases is an unknown values and can't be incremented, that's why `select null+1` will return null.

As a side note, `COALESCE(column, 0)` should be used instead of a `CASE` expression.

nineinchnick:
> In this way core should throw exception or warn user about updating counters for NULL column.

The database does not throw an error, so why would the framework do it?

nineinchnick:
Trying to increment a `null` value is a valid expression, just the result is still `null` because the value is stull unknown.

RusAlex:
What about COALESCE @cebe, @klimov-paul ?

samdark:
Closing for reasons @nineinchnick provided.

PR #9491, opened 2015-08-21
Cookies - base64 encoding

matteotrubini:
Add base64 encoding/decoding for cookies value to avoid OWASP
ModSecurity Core Rule Set (CRS) false-positive detection

In our tests "_identity" cookie offends rules 981245 and 981246 as it contains a serialized data that it's detected as a suspect SQL injection.
See https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/master/base_rules/modsecurity_crs_41_sql_injection_attacks.conf

It's a first review, maybe the encoding support can be enabled with a flag like enableCookieValidation.


samdark:
Hmm. That doesn't seem to be real issue but false positive of security scanner. Not going to change working code just in order for it to be silent.

madolo:
I agree that it's not a bug, but an enhancement to make the cookie management more "robust": IMHO base64 encoded strings are preferable instead of serialized values.
I have suggested to use the same technique adopted for CSRF token as in that way it would be used a well known method.

I had made tests on a cPanel 11.50 host where I've enabled ModSecurity's "OWASP ModSecurity Core Rule Set" that's shipped by default.

samdark:
base64 encoded strings are significantly longer. Thus may violate cookies size limit of IEs easily.

samdark:
I don't see how base64 making it more robust or why it's preferrable.

madolo:
It's just my point of view, I prefer OWASP compatibility paying a 33% overhead on string length.
I understand your position, I will simply maintain my fork adding a flag to enable explicitly the encoding when needed.
Thanks for your review!

PR #9498, opened 2015-08-21
Update Query doc return type

Panoptik:
When used ActiveQuery chain call after call any of join type IDE cannot resolve ActiveQuery methods such an asArray
Example:

``` PHP
SomeActiveRecordModel::find()
            ->select('...') // return ActiveQuery
            ->leftJoin('...', '...') // return Query
            ->asArray() // it's become undefined
            ->column();
```


samdark:
Which IDE?

Panoptik:
PhpStorm 9.0.2

samdark:
That's weird. It shoud support `$this` as type well.

Panoptik:
I used to use "static" keyword instead. But I saw in the same code $this and consider it can be used.

samdark:
Got it. Thanks!

PR #9517, opened 2015-08-25
BaseArrayHelper treats objects properties as array keys

vambus:
If object property is private or not exists, it returns default value.


cebe:
This has been discussed many times before, the current behavior is exptected. see https://github.com/yiisoft/yii2/issues/7193#issuecomment-73287156

Issue #9520, opened 2015-08-26
Docblock "@return" into \yii\db\ActiveRecord::findOne and others failed


antivir88:
findOne return static or null , IDE PhpStorm thinking that is Controller , not Model ((((

samdark:
How can I reproduce it?

antivir88:
public function actionDeltrans()
    {
        /** @var Transaction $trans */
        $trans = Transaction:



Remove doc-block and add ":" into PHPStorm

antivir88:
https://github.com/yiisoft/yii2/pull/9521

cebe:
closed. see https://github.com/yiisoft/yii2/pull/9521#issuecomment-135395423 and #7404 and https://youtrack.jetbrains.com/issue/WI-19953 .

PR #9521, opened 2015-08-26
DocBlock errors into @return

antivir88:
https://github.com/yiisoft/yii2/issues/9520


cebe:
this is not correct, the method returns the class type which it is called on. `static` is correct. If it does not work your IDE, that is an IDE bug and should be fixed there.

PR #9528, opened 2015-08-27
Add supporting correct output formating for SPL classes and Generators.

MaXL-ru:
Supporting correct output formatting for SPL classes and Generators.

Current implementation `\yii\web\XmlResponseFormatter` does not support SPL objects and Generators (PHP 5.5+).

E.g.:

``` php
public function actionTest()
{
  \Yii::$app->response->format = \yii\web\Response::FORMAT_XML;

  $generator = function ($start, $end) {
    for ($i = $start; $i <= $end; ++$i) {
      yield $i;
    }
  };

  return $generator(1, 3);
}
```

will output:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<response>
  <Generator>
    <item>1</item>
    <item>2</item>
    <item>3</item>
  </Generator>
</response>
```

but `<Generator`> is extra element and output must be:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<response>
  <item>1</item>
  <item>2</item>
  <item>3</item>
</response>
```

For SPL objects we have same situation, for example:

``` php
public function actionSitemap()
{
  \Yii::$app->response->format = \yii\web\Response::FORMAT_XML;
  \Yii::$app->response->formatters[\yii\web\Response::FORMAT_XML] = [
    'class'   => '\yii\web\XmlResponseFormatter',
    'rootTag' => 'urlset',
    'itemTag' => 'url'
  ];

  return \Yii::$app->contentTree->getItemsHeapForSitemap();
}
```

`getItemsHeapForSitemap` method return instance of `\SplQueue`.

will output:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset>
  <Heap>
    <url>
      <loc>https://domain/path.html</loc>
      <lastmod>2015-06-17</lastmod>
      <priority>0.99</priority>
    </url>
    <!-- next urls -->
  </Heap>
</urlset>
```

but preffered is:

``` xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset>
  <url>
    <loc>https://domain/path.html</loc>
    <lastmod>2015-06-17</lastmod>
    <priority>0.99</priority>
  </url>
  <!-- next urls -->
</urlset>
```

Because Generator and SPL classes implements `\Traversable` interface, I think it is possible to use it as arrays in `buildXml` method of `\yii\web\XmlResponseFormatter`.

I add `$useTraversableAsArray` (true by default) member to `\yii\web\XmlResponseFormatter`, if someone wants to use old style.

What do you think about this ?

### DIFF
--- a/framework/web/XmlResponseFormatter.php
+++ b/framework/web/XmlResponseFormatter.php
@@ -44,6 +44,10 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      * @var string the name of the elements that represent the array elements with numeric keys.
      */
     public $itemTag = 'item';
+    /**
+     * @var boolean whether to enable to interpret implemented \Traversable interface objects as array
+     */
+    public $useTraversableAsArray = true;
 
 
     /**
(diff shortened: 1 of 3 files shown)


samdark:
Looks good to me overall. @yiisoft/core-developers comments?

klimov-paul:
What will happen with `Traversable` in JSON format?
Does it also need special treatment?

MaXL-ru:
With `\yii\web\Response::FORMAT_JSON` all OK.

For 1st example output is:
```
[1,2,3]
```

For 2nd:
```
[
  {
    "loc":"https://domain/path.html",
    "lastmod":"2015-06-17",
    "priority":0.99
  }
  ...
]
```

samdark:
Thanks!

PR #9532, opened 2015-08-27
Fix documentation of DbMessageSource.

nkovacs:
Setting cachingDuration to 0 doesn't disable caching,
enableCaching controls that.

[ci skip]

### DIFF
--- a/framework/i18n/DbMessageSource.php
+++ b/framework/i18n/DbMessageSource.php
@@ -62,7 +62,7 @@ class DbMessageSource extends MessageSource
     /**
      * @var Cache|array|string the cache object or the application component ID of the cache object.
      * The messages data will be cached using this cache object. Note, this property has meaning only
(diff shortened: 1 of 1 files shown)


samdark:
Thanks!

nkovacs:
But setting cachingDuration to 0 does not disable caching.

cebe:
@samdark PR is correct, cachingDuration=0 does not disable, it will cause the cache to never expire.

samdark:
There's not a single word about anything being disabled:

> The messages data will be cached using this cache object. Note, this property has meaning only
> in case [[cachingDuration]] set to non-zero value and [[enableCaching]] is true.

nkovacs:
why would the property not have meaning if cachingDuration is set to zero? The cache is still used, but cached data never expires.

cebe:
fixed by rewording the comment.

samdark:
:+1:

PR #9537, opened 2015-08-28
The value of ['PHP_AUTH_USER'] and ['PHP_AUTH_PW'] variables may be s…

tanlanxing:
The value of ['PHP_AUTH_USER'] and ['PHP_AUTH_PW'] variables may be string 'null', but not null.fix it.

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -876,7 +876,7 @@ class Request extends \yii\base\Request
      */
     public function getAuthUser()
     {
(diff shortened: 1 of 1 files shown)


tanlanxing:
The value of ['PHP_AUTH_USER'] and ['PHP_AUTH_PW'] variables may be string 'null', but not null.fix it.

klimov-paul:
> The value of ['PHP_AUTH_USER'] and ['PHP_AUTH_PW'] variables may be string 'null'

In which case it could be?
This does not sounds right. If such things are normal it means word 'null' can not be used as username or password.

samdark:
Closing since no info was provided to make a decision.

PR #9562, opened 2015-09-01
Add char type to db schema.

usecomposable:
https://github.com/yiisoft/yii2/blob/master/framework/db/SchemaBuilderTrait.php is also missing a char() method.

### DIFF
--- a/framework/db/Schema.php
+++ b/framework/db/Schema.php
@@ -42,6 +42,7 @@ abstract class Schema extends Object
      */
     const TYPE_PK = 'pk';
     const TYPE_BIGPK = 'bigpk';
+    const TYPE_CHAR = 'char';
     const TYPE_STRING = 'string';
     const TYPE_TEXT = 'text';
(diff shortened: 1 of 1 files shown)


df2:
Can I tackle this properly? I created some notes: https://gist.github.com/df2/c35fff85909a277b9f40

Let me know if that is the right way to go about this and I will see what I can do over the weekend.

samdark:
The plan is good.

df2:
OK.

Ziggizag:
Hi all,

In case - as I understood - 'char' datatype is already added to ColumType, please update Yii2 documentation appropriately here:

http://www.yiiframework.com/doc-2.0/yii-db-columnschema.html#$type-detail

and here:

http://www.yiiframework.com/doc-2.0/yii-db-querybuilder.html#getColumnType%28%29-detail

SilverFire:
@Ziggizag not yet. Thanks for notice, we will update docs

Issue #9476, opened 2015-08-20
DI injection via method signature


samdark:
While constructor injection is enough in general, controllers are a special case because there are too many different contexts (actions) and each needs a separate set of class instances.

Would be cool to support doing it like this instead of checking for action name in constructor:

```php
class PostController extends Controller
{
    public function actionView($id, OpenGraphInterface $og)
    {
        $post = Post::findOne($id);
        $meta = $og->getMeta($post);
        // ...
    }
}
```

samdark:
Preferrably to be implemented via trait so method-injection could be added to any user class.

samdark:
```php
class MyClass
{
    use MethodInjection;

    public function doIt(CoolInterface $coolClass)
    {
        // ...
    }
}
```

briedis:
Oh yes, at last. Really enjoyed this feature in Laravel.

I would like to suggest to cleanup the controller constructor, because it's kinda stupid, that I have to inject dependencies between second and third parameter. I still don't understand why would some kind of `$config` be needed there.

I like how cleanly Laravel does it - no default arguments for constructors and actions. 

And it works awesome when you can request Request object to be injected in the method - and it just injects the current request.

```
    public function actionNewOrder()
    {
        $storeId = (int)Yii::$app->request->post('storeId');
        $isSampleOrder = (bool)Yii::$app->request->post('isSampleOrder');
```

would nicely become, much much cleaner:

```
    public function actionNewOrder(Request $request)
    {
        $storeId = (int)$request->post('storeId');
        $isSampleOrder = (bool)$request->post('isSampleOrder');
```

df2:
> I still don't understand why would some kind of $config be needed there.
Have you tried it without?

> would nicely become, much much cleaner:
Would conflict with passing url parameters as action parameters. And I really think you could add one line to store the request object locally to get the same effect.

df2:
@samdark 

When would the services be binded? After beforeAction event?

briedis:
I bind services in a separate file. Sadly, there isn't a documented place where you should put your bindings. Service provider class would be useful.

df2:
Gotcha. I have a particular use case where I don't know the implementation of a particular interface until beforeAction(), so I have to set it there. Other than that I am using the constructor method and filling the container in the entry script before running the application.

samdark:
@df2 not sure about it yet. I guess, yes. Either before or after `beforeAction`.

df2:
[...]

PR #9565, opened 2015-09-01
DI injection via method signature

mdmunir:
#9476 I hope i am doing right.


df2:
:+1: Looks good. Could you add the following to the tests?

```php
$result = $controller->runAction('aksi5', $params);
$this->assertEquals(['d426', 'independent', 'other_qux'], $result);
...
public function actionAksi5($q, Bar $bar, QuxInterface $quxApp)
{
    return [$q, $bar->foo, $quxApp->quxMethod()];
}
```

samdark:
In #9476 I've meant general purpose method injection, not necessary controller but probably it's OK.

tests/framework/web/ControllerTest.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->assertTrue($quxApp instanceof OtherQux);
+        $this->assertEquals($quxApp, Yii::$app->quxApp);
+        $this->assertEquals('belong_app', $quxApp->b);
+
+        $result = $controller->runAction('aksi4', $params);
+        $this->assertEquals(['independent', 'other_qux', 'd426'], $result);
+    }
+}
+
+class FakeController extends Controller
```
samdark: It's better to move these into separate files.

mdmunir:
@samdark ya i know. But how to do that?

```php
class MyObj extends Object
{
    public function test(Cache $cache)
    {
        $cache->set();
    }
}

$obj = new MyObj();
$obj->test(); // <- will always error
```

Other scenario
```php

class MyObj extends Object
{
    protected function test(Cache $cache) // as protected
    {
        $cache->set();
    }

    public __call($name,$args)
    {
        $ref = getReflectionMethod($name);
        $args = injectArgument($ref,$args);
        return call_function_array([$this,$name],$args);
    }
}

$obj = new MyObj();
$obj->test(); 
```
I am not sure its a good scenario

df2:
I think one-step at a time. First get it working well with controller actions and the modified tests passing. Then it can be put through its paces after merging. Once that is stable, some thought can be put in to a more generic method version. Just my 2¢.

mdmunir:
@df2 do you need another test before i add new commit?

samdark:
Agree.

samdark:
Overall it works well. One thing is, for example, if I define method signature like the following

```php
public function actionIndex(BooleanValidator $validator)
```

and then pass `?validator=123` via request, Yii raises an exception. I think if an argument is type-hinted, it should not receive any value.

framework/web/Controller.php
```diff
@@ -84,6 +84,13 @@ public function bindActionParams($action, $params)
                 unset($params[$name]);
             } elseif ($param->isDefaultValueAvailable()) {
                 $args[] = $actionParams[$name] = $param->getDefaultValue();
+            } elseif (($c = $param->getClass()) !== null) {
+                $type = $c->getName();
```
samdark: Better to use `$class` instead of `$c`.

samdark: Also `$type` could be renamed to `$className`.

mdmunir:
@samdark I think the same. Prefer change the order of checking. Check type at first before $params avaliable.
```php
            $name = $param->getName();
            if (($class = $param->getClass()) !== null) {
                $className = $class->getName();
                if (Yii::$app->has($name) && ($obj = Yii::$app->get($name)) instanceof $className) {
                    $args[] = $actionParams[$name] = $obj;
                }else{
                    $args[] = $actionParams[$name] = Yii::$container->get($className);
                }
            }elseif (array_key_exists($name, $params)) {
                if ($param->isArray()) {
                    $args[] = $actionParams[$name] = (array) $params[$name];
                } elseif (!is_array($params[$name])) {
                    $args[] = $actionParams[$name] = $params[$name];
                } else {
                    throw new BadRequestHttpException(Yii::t('yii', 'Invalid data received for parameter "{param}".', [
                        'param' => $name,
                    ]));
                }
                unset($params[$name]);

```

samdark:
Yep. The case I've described should be tested.

mdmunir:
```php
    // FakeController
    public function actionAksi6($q, EmailValidator $validator)
    {
        return [$q, $validator->validate($q), $validator->validate('misbahuldmunir@gmail.com')];
    }

    // Test
    $params = ['fromGet'=>'from query params','q'=>'d426','validator'=>'avaliable'];
    $result = $controller->runAction('aksi6', $params);
    $this->assertEquals(['d426', false, true], $result);
```
?

samdark:
Yes.

tests/framework/web/FakeController.php
```diff
@@ -0,0 +1,49 @@
...
+    public function actionAksi5($q, Bar $bar, QuxInterface $quxApp)
+    {
+        return [$q, $bar->foo, $quxApp->quxMethod()];
+    }
+
+    public function actionAksi6($q, EmailValidator $validator)
+    {
+        return [$q, $validator->validate($q), $validator->validate('misbahuldmunir@gmail.com')];
+    }
+}
```
Andre-487: Missing `\n`?

tests/framework/web/stubs/Bar.php
```diff
@@ -0,0 +1,19 @@
...
+use yii\base\Object;
+
+/**
+ * @author Misbahul D Munir <misbahuldmunir@gmail.com>
+ * @since 2.0
+ */
+class Bar extends Object
+{
+    public $foo;
+}
```
Andre-487: And this place have missing `\n` too

samdark:
The code looks OK to me now (except really minor formatting issues). @mdmunir do you want to attempt extending DI docs?

mdmunir:
You do @samdark, i am not good in english :grinning:

samdark:
OK.

samdark:
Found another thing. While web controllers are good we haven't covered console controllers.

mdmunir:
Ya, i miss it. Should i take it in this PR or create new PR?

samdark:
This one is OK.

samdark:
Merged https://github.com/yiisoft/yii2/commit/b7020065c8bdea22878589a57ba4513ff53ea15e. Thank you!

PR #9577, opened 2015-09-02
Update installation docs to latest version of composer-asset-plugin

usecomposable:
[ci skip]

Anyone stuck on 1.0.0 will be riddled with deprecation notices.

### DIFF
--- a/docs/guide/start-installation.md
+++ b/docs/guide/start-installation.md
@@ -37,7 +37,7 @@ by running `composer self-update`.
 
 With Composer installed, you can install Yii by running the following commands under a Web-accessible folder:
 
-    composer global require "fxp/composer-asset-plugin:~1.0.0"
+    composer global require "fxp/composer-asset-plugin:~1.0.3"
     composer create-project --prefer-dist yiisoft/yii2-app-basic basic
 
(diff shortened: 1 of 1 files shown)


docs/guide/start-installation.md
```diff
@@ -37,7 +37,7 @@ by running `composer self-update`.
 
 With Composer installed, you can install Yii by running the following commands under a Web-accessible folder:
 
-    composer global require "fxp/composer-asset-plugin:~1.0.0"
+    composer global require "fxp/composer-asset-plugin:~1.0.3"
```
klimov-paul: Shouldn't it be `~1.0` then?

df2: Depends on what they do with 1.1 ;-) But yeah, that also would pick up 1.0.3 as the latest.

samdark: `~1.0.0` works fine i.e. it will install 1.0.3 and 1.0.4 when it will be released.

klimov-paul:
Adjusted b107d61aac1684e940749abaa1a8aa1241734c76

samdark:
:+1:

Issue #9443, opened 2015-08-17
Add support for unsigned integer to migration schema builder 


lisi4ok:
Exception 'yii\base\UnknownMethodException' with message 'Calling unknown method: yii\db\ColumnSchemaBuilder::unsigned()'

in /var/www/ivm/vendor/yiisoft/yii2/base/Object.php:220

My code:

	public function up() {
		$this->createTable('auditLog', [
			'id' => $this->primaryKey(),
			'model' => $this->string(255)->notNull(),
			'action' => $this->string(255)->notNull(),
			'oldAttributes' => $this->text(),
			'newAttributes' => $this->text(),
			'IPAddress' => $this->string(255),
			'createdAt' => $this->datetime(),
			'updatedAt' => $this->datetime(),
			'createdBy' => $this->integer(10)->unsigned(),
			'updatedBy' => $this->integer(10)->unsigned(),
		]);
	}

samdark:
There's no support for it.

samdark:
Yet.

bryglen:
Bump for this. we need this

cebe:
you can just use old syntax to specify and custom type you like. ColumnSchemaBuilder is mostly useful to create schema independed defintions.

lisi4ok:
@cebe  I don`t like old syntax. So i report this error.
Please release the new version more faster. I can`t w8 for the new syntax. Its soo easy :P

rezident:
You can use on version < 2.0.7
```php
$this->createTable('{{%ipgeo}}', [
      'from' => $this->integer() . ' UNSIGNED NOT NULL',
      'to' => $this->integer() . ' UNSIGNED NOT NULL'
]);
```

lisi4ok:
I know but it looks ugly to me.
Better to use only 1 way. Only OLD WAY or Only NEW WAY.
When will be released -v 2.0.7 ?

rezident:
My answer not only for you. I had this problem too. I agree about one way but I admit use OLD WAY temporary. Still I use this code (before -v 2.0.7):
~~~php
$this->createTable('{{%ipgeo}}', [
      'from' => $this->integer() . ' UNSIGNED NOT NULL', //@todo ->notNull()->unsigned(),
      'to' => $this->integer() . ' UNSIGNED NOT NULL', //@todo ->notNull()->unsigned(),
      'vocabulary_id' => $this->integer()->notNull()
]);
~~~

PR #9617, opened 2015-09-06
Adds char datatype [Fixes #9562]

usecomposable:
Notes:

Why does db/oci/Schema not have a typemap?
There were 4 "rows" commented out in OracleQueryBuilderTest. They are back in now.
All columnTypes() methods are reformatted to be human readable.

Fixes #9562

Need to verify on DBMS:
- [x] MySQL
- [x] PostgreSQL
- [ ] MSSQL
- [x] SQLite
- [ ] Curbid
- [ ] Oracle


tests/framework/db/SchemaTest.php
```diff
@@ -161,7 +161,7 @@ public function getExpectedColumns()
                 'defaultValue' => 1,
             ],
             'char_col' => [
-                'type' => 'string',
+                'type' => 'char',
```
df2: This is causing the test to fail. Shouldn't the type be recognized as `'char'` now? If so, then something else is wrong.

df2:
This should be merged before someone tackles #9443 as they will be using the same test arrays that have been reformatted. I can give it a shot btw as it should be fairly similar to this one.

SilverFire:
Looks good. What do we need to merge this?

samdark:
Some real world testing on various DBs we support.

SilverFire:
Added checklist to the header. I will test on MySQL, PgSQL, SQLite.
If anybody can help with others - I will be thankful

framework/db/mssql/QueryBuilder.php
```diff
@@ -25,6 +25,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
     public $typeMap = [
         Schema::TYPE_PK => 'int IDENTITY PRIMARY KEY',
         Schema::TYPE_BIGPK => 'bigint IDENTITY PRIMARY KEY',
+        Schema::TYPE_CHAR => 'char(1)',
```
samdark: I thing this could be wrong. `char` in MSSQL stands for non-unicode only same as `varchar`. There are unicode aware `nchar` and `nvarchar`.

df2: Noted. Think I may redo this branch and leave the tests in the horizontal format.

SilverFire:
Is this PR solved by #10921?

df2:
@SilverFire yes

PR #9628, opened 2015-09-07
remove unnecessary gettimeofday system call

nkovacs:
`new DateTime()` initializes the datetime object with the current time, which uses the `gettimeofday` syscall. This can be pretty expensive on a system without vsyscall or vdso, and we had some serious performance problems partly because of this.
This is completely unnecessary, since the timestamp is then overwritten.


cebe:
I'm quite sure there was a reason why the code was written this way, need to check it in detail.

nkovacs:
negative timestamps, but it works with @

cebe:
Merged, thanks!

PR #9634, opened 2015-09-08
Doc block confuse in ManagerInterface::getRolesByUser()

Vovan-VE:
Confisung documentation block for `ManagerInterface::getRolesByUser()`. The method DOES NOT return roles assigned indirectly.


samdark:
Thanks!

PR #9636, opened 2015-09-08
Show errors from all models

icron:
`$user->validate() && $profile->validate()`  It shows errors only the first model.


cebe:
Thanks!

Issue #8061, opened 2015-04-13
Nested fields and expand in REST API


ghost:
Facebook Graph API also uses fields and expand query parameters in REST APIs. Now Yii2 supports same but upto 1 level. Can we do nested fields like Facebook, for example:

GET graph.facebook.com/me?fields=albums.limit(5){name, photos.limit(2){name, picture, tags.limit(2)}},posts.limit(5)

In above example, my albums are limits to 5 only and again in each album only name and photos fields will be fetched with max 2 photos per album with again photos nested to give max 2 tags per photo. Along with albums, it will also give max 5 posts of the user.

Now limit(n) would be advance feature but I think nested expand is something easy to implement and really necessary feature to have.

PS: On separate note if you plan to support nested limit, you can also look into offset and sort on nested elements.

fernandezekiel:
as @qiangxue  said, this is more of a rare case
https://github.com/yiisoft/yii2/issues/303#issuecomment-36846427
i am however looking forward on how to implement this

lichunqiang:
Hi all

What's going on about this question.

akaNightmare:
hi, maybe it make sense implement simple functionality like:
```php
 expand="user.contacts"
```
(nested relations)

mdmunir:
My solusion https://github.com/yiisoft/yii2/issues/6844#issuecomment-131482508

Yiivgeny:
+1

josegustavo:
How I can convert this:

`id,name,items{id,title,categories{id,title,image},relates{id,title,image},image},image`

in this:

`{
	'id' : {},
	'name' : {},
	'items' : {
		'id' : {},
		'title' : {},
		'categories' : {
			'id' : {},
			'title' : {},
			'image' : {}
		},
		'relates' : {
			'id' : {},
			'title' : {},
			'image' : {}
		},
		'image' : {}
	},
	image : {}
}`

tunecino:
> GET graph.facebook.com/me?fields=albums.limit(5){name, photos.limit(2){name, picture, tags.limit(2)}},posts.limit(5)

This looks more like a query language than REST. Maybe a complete implementation of [GraphQL](http://graphql.org/) as extension would be better than adding it as feature to current REST.

some resources:
https://facebook.github.io/graphql/
https://philsturgeon.uk/api/2017/01/24/graphql-vs-rest-overview/

Faryshta:
to me it looks like a proposal for aggregated reporting in SOA.

[...]

Issue #6844, opened 2015-01-13
Serialize deeply nested relations


HriBB:
How can we serialize deeply nested relations?

I have a REST endpoint which you can call like this: 

```
/api/locations?expand=types,files,locations,locations.types,locations.files 
```

Note: in my case Location model hasMany child Locations, hence the expand 'locations' on itself.

This is my controller actions() function:

```
public function actions()
{
    $actions = parent::actions();

    $actions['view']['findModel'] = function($id, $action) {
        $modelClass = $this->modelClass;

        $expand = $this->module->request->get('expand');
        $relations = preg_split('/\s*,\s*/', $expand, -1, PREG_SPLIT_NO_EMPTY);

        if (empty($relations)) {
            $model = $modelClass::findOne($id);
        } else {
            $model = $modelClass::find()->with($relations)->where(['id' => $id])->one();
        }

        if (isset($model)) {
            return $model;
        } else {
            throw new NotFoundHttpException("Object not found: $id");
        }
    };
}
```

If I set the breakpoint on 'return $model' I can see that all required relations and nested relations are populated.
I want to serialize everything into one big json, but the problem is that serializer does not understand nested relations beyond first level.
Location types, files and locations are successfully serialized, but locations.types and locations.files are not.

Model->toArray() calls ArrayHelper::toArray(), which runs a recursive loop and finally calls $model->toArray() again, but this time without the $expand parameter, which is needed to serialize relations.

At the moment I am overriding toArray() method on Location model:

```
public function toArray(array $fields = [], array $expand = [], $recursive = true)
{
    $data = [];

    foreach ($this->resolveFields($fields, $expand) as $field => $definition) {
        $data[$field] = is_string($definition) ? $this->$definition : call_user_func($definition, $this, $field);
    }

    if ($this instanceof Linkable) {
        $data['_links'] = Link::serialize($this->getLinks());
    }

    // start override
    $relations = [];

    // construct relation graph
    foreach ($expand as $relation) {
        if (strstr($relation, '.')) {
            list($relation,$child) = explode('.', $relation);
            $relations[$relation][] = $child;
        } else if (!isset($relations[$relation])) {
            $relations[$relation] = [];
        }
    }

    // serialize relations
    foreach ($relations as $relation => $children) {
        if (isset($data[$relation])) {
            $rel = $data[$relation];
            if (is_array($rel)) {
                foreach ($rel as $k => $v) {
                    $data[$relation][$k] = $v->toArray([], $children);
                }
[...]

Issue #7098, opened 2015-01-30
REST: limit fields list for index and view actions


RomeroMsk:
I'm using `yii\rest\ActiveController` and want to limit list of resource fields returned by `index` and `view` actions. I can use `yii\rest\ViewAction::$findModel` to set `scenario` of found model - it can help to restrict fields for `view`, but for `index` it is not so easy.
Why not add something like `defaultFields` (and `deafultExtraFields`) property to these actions and pass it to `yii\rest\Serializer::getRequestedFields()` to intersect with arrays from $_GET? Name `defaultFields` is not good for restriction, but you can suggest another names.

fernandezekiel:
what do you mean by *restrict*?
can it not be accomplished by configuring the `fields()` or `extraFields`?

RomeroMsk:
Model has `attr1`, `attr2`, `attr3`, `attr4`. I need to show only `attr1`, `attr2` and `attr3` by default in `index` and `view`. If consumer will add `?fields=attr1,attr2` to URL, he will get only those. But if he will ask for `?fields=attr2,attr3,attr4`, he will get an intersection (`attr2` and `attr3`), because `attr4` is not allowed.

fernandezekiel:
i probably would just deal with the model to do this, override the 
`toArray()`

```php
public function toArray(array $fields = [], array $expand = [], $recursive = true)
{
    if (empty($fields)) {
        $fields = ['username', 'id'];
    }

    return parent::toArray($fields, $expand, $recursive);
}
```

this would work for me because i have models for my `api` that extends my `common` models, i don't know if this would be ok to you

fernandezekiel:
if not, i suppose your solution about `defaultFields` in the Serializer would work better

RomeroMsk:
My idea is to make it easier than extending each model for API. BTW, you can do the same by overriding model's `fields()` method.

fernandezekiel:
yeah i guess

```php
class UserController extends Controller
{
    public $serializer = [
        'class' => 'api\\Serializer',
        'defaultFields' => ['id', 'name']
    ];

```

RomeroMsk:
Also these new action properties can accept a callback to set different fields arrays according to API user. Then you can do it like this:
```php
    public function actions()
    {
        $actions = parent::actions();
        $actions['view']['defaultFields'] = function () {
            if (Yii::$app->user->identity->username == 'testuser') {
                return ['attr1', 'attr2', 'attr3'];
            }
            return null;
[...]

PR #9642, opened 2015-09-09
Nested expand field

mdmunir:
#8061 #6844

### DIFF
--- a/framework/helpers/BaseArrayHelper.php
+++ b/framework/helpers/BaseArrayHelper.php
@@ -56,13 +56,20 @@ class BaseArrayHelper
      * @param boolean $recursive whether to recursively converts properties which are objects into arrays.
      * @return array the array representation of the object
      */
-    public static function toArray($object, $properties = [], $recursive = true)
+    public static function toArray($object, $properties = [], $recursive = true, $expands = [])
     {
         if (is_array($object)) {
             if ($recursive) {
                 foreach ($object as $key => $value) {
                     if (is_array($value) || is_object($value)) {
-                        $object[$key] = static::toArray($value, $properties, true);
+                        if (is_int($key)){
+                            $expand = $expands;
+                        } elseif (isset($expands[$key])) {
+                            $expand = $expands[$key];
+                        } else {
+                            $expand = [];
+                        }
+                        $object[$key] = static::toArray($value, $properties, true, $expand);
                     }
                 }
             }
@@ -85,7 +92,7 @@ class BaseArrayHelper
                 }
             }
             if ($object instanceof Arrayable) {
-                $result = $object->toArray([], [], $recursive);
+                $result = $object->toArray([], $expands, $recursive);
             } else {
                 $result = [];
                 foreach ($object as $key => $value) {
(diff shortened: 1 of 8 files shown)


bazaglia:
Tried locally and it works. Also, looking good! Thanks @mdmunir. Hope it gets merged soon :)

framework/helpers/BaseArrayHelper.php
```diff
@@ -56,13 +56,20 @@ class BaseArrayHelper
      * @param boolean $recursive whether to recursively converts properties which are objects into arrays.
      * @return array the array representation of the object
      */
-    public static function toArray($object, $properties = [], $recursive = true)
+    public static function toArray($object, $properties = [], $recursive = true, $expands = [])
```
samdark: Method signature changed. Can't merge it in 2.0.x. 2.1 only.

tunecino:
nice feature @mdmunir but is there a way to control authorization ? I mean when used to navigate where we are not authorized to like: `/item?expand=cart.owner.account`

tunecino:
otherwise I think it should be disabled by default to be implemented after unsetting related fields within the checkAccess method.

fernandezekiel:
@tunecino the problem with this kind of design is that the fields are declared inside a model instance...
i would like to see somebody implement another layer for this one to implement in controller level wherein it can filter by user's credentials which expand it can see or not...

it can still be done via the model to some degree...
i just have a subclass of my main model that does the check

tunecino:
I think you are right @fernandezekiel.

Other thing that may help solving this could be by first solving Nested Urls. then we can also have something that works like `scenarios` in `model::rules` to control fields expanding. like when doing for example `GET /carts/2/orders/4` we can easily have pre-setted conditions on fields to auto expand some related relations like the data in the juction_table between both models.

I think I have few ideas on implementing nested urls via a custom UrlRule class. I'll start by that. maybe an extension. if it works we can implement similar thing to the framework. if it dosent, who knows, maybe we help influencing others on figuring out a good solution.

fernandezekiel:
the issue is related here though
https://github.com/yiisoft/yii2/issues/7098

fernandezekiel:
i have some implementation of this one already, although it kind of makes me re-declare the mapped return fields in the controller... 
i can make the usage also that if it's not declared in the expandDependency that it will not be in the response...
i also use this for eager loading or callbacks for the query object if a field was requested in the expand
```php

    /**
     * usages:
     * 1. declare the expand field itself as the array value without a string pair, example: location
     * 2. declare the expand field by key value pair , the key will be the expand field requested and the
     * value can be a string or an array of the needed relations to be eager loaded, example: 'created_user' => 'createdUser',
     * 'is_deletable' => ['productAchPv', 'productGateway']
     * 3. declare the expand field wherein the other value is a callable that accepts query as a parameter and will return
     * the query object
     * ~~~
     * return [
     *      'location', 'created_user' => 'createdUser',
     *      'is_deletable' => ['recurrings', 'projects'],
     *      'special_field' => function ($query) {
     *          return $query;
     *      }
     * ];
     * ~~~
     * @return array
     */
    public function expandDependency()
    {
        return [];
    }
```

samdark:
Won't make it into 2.0.

PR #9643, opened 2015-09-09
Add migration for db cache

mdmunir:


framework/caching/migrations/m150909_153426_cache_init.php
```diff
@@ -0,0 +1,72 @@
...
+                break;
+            default:
+                $blob = 'BLOB';
+                break;
+        }
+
+        $this->createTable($cache->cacheTable, [
+            'id' => $this->string(128)->notNull(),
+            'expire' => $this->integer(),
+            'data' => $blob,
```
LAV45: why not use 
```php 
$this->createTable($cache->cacheTable, [
    'id' => $this->string(128)->notNull(),
    'expire' => $this->integer(),
    'data' => $this->binary(),
], $tableOptions);

$tableName = $this->db->quoteTableName($cache->cacheTable);
$this->addPrimaryKey($tableName . '_pk', $cache->cacheTable, 'id')
```

mdmunir: `binary()`, ya.. i miss it :grinning: 
about primary key, i just copied from this https://github.com/yiisoft/yii2/blob/master/framework/rbac/migrations/m140506_102106_rbac_init.php#L49

samdark:
Thanks!

PR #9645, opened 2015-09-09
Add new event  \yii\base\Model::EVENT_AFTER_CREATE_VALIDATORS

lav45:


klimov-paul:
What is the sense of it?
Looks useless at the first glance.

LAV45:
This event will give the opportunity for the Behavior to add your own validators to the moment when they are initialized in the owner class.

klimov-paul:
And what is the benefit? We already have 'beforeValidate'.

fernandezekiel:
if that's the case then yes, why not just use beforeValidate() ?

PR #9655, opened 2015-09-10
Closed #9635: Added default CSS class for ActionColumn header

### LINKED ISSUE #9635: Supply ActionColumn with corresponding header class
currently, the ActionColumn is rendered without column class. it would be logical to have this one set as default because of i.e. width declarations in CSS:
* $headerOptions = ['class'=>'action-column']

Can make a PR if requested

arogachev:


samdark:
Thanks!

Issue #9661, opened 2015-09-10
Refactoring BaseActiveRecord


dynasource:
After some investigation in BaseActiveRecord, I am surprised by the number of refactorings that still have to be done. Many lines with logic do the same thing. I will do some PR's for this.

PR #9662, opened 2015-09-10
9661 refactoring baseactiverecord

dynasource:
I have posted 3 commits in which functions within BaseActiverecord are reused and cleaned up. Much more can be done on refactoring. 

Please take a look at these and let me know if its desired to take it steps further.
#9661

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -1228,8 +1228,8 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
 
         // update lazily loaded related objects
         if (!$relation->multiple) {
-            $this->_related[$name] = $model;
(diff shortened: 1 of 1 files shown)


cebe:
most of these changes are replacing code that is there for performance reasons. Can you elaborate more on why you think these changes are necessary?

dynasource:
https://github.com/dynasource/yii2/commit/1719cc4541fb31cb06c2ff79347a5a1ec89564fd#commitcomment-13169603

dynasource:
To elaborate on your argument. I dont see my current commits to affect performance. Using central functions would not affect performance. On the contrary, it would enable performance improvements by:
* less memory through smaller files (because of code reuse, less lines is better)
* performance gain through reusing functions (php or yii caching)

PR #9701, opened 2015-09-16
Session is started ONLY when we need put some data in it or when session was created previosly.

vatruha:
Right now, if we execute command \Yii::$app->session->getAllFlashes() Yii will create session for us. Even if we didn't have it before.

Every 'get' method in Session creates session. As result we will flood DB (or another session mechanism) with a lot of empty sessions.

@Requilence has created pull request #9356 about it.

But his pull request fixes only 2 methods.

I'm proposing absolutely different approach for this.

The main idea: we will create session (session->open()) ONLY when we need to put some data in it or when it was created previosly.

As result, only some methods are really needed to execute $this->open(), and a lot of methods are not needed to execute $this->open()


dynasource:
This solution looks a lot more logical than the previous one.

samdark:
Won't make it into 2.0.

Issue #7534, opened 2015-03-02
Item envelope in \yii\rest\Serializer


FabrizioCaldarelli:
Hi all,
this is a possible improvement (not a bug).

I think that if I use collection evelope "items" in serializer, it could be useful to have single item envelope, as "item". In this way i'll have same structure of response.

Now i make this functionality extending ActiveController's actions() in this way:

```
    $actions['view'] = [
        'class' => 'api\components\CustomViewAction',
        'modelClass' => $this->modelClass,
        'checkAccess' => [$this, 'checkAccess'],
    ];
```

So in CustomViewAction i have

```
public function run($id) {
    $model = $this->findModel($id);
    if ($this->checkAccess) {
        call_user_func($this->checkAccess, $this->id, $model);
    }
    return ['item' => $model];
}
```

In this way i have single item or many items response in same way, using an envelope.

At the end, in $serializer's controller variable should be:

```
public $serializer = [
    'class' => 'yii\rest\Serializer',
    'collectionEnvelope' => 'items',
   'singleEnvelope' => 'item'
];
```

cebe:
why would you need an envelope for a single item?

FabrizioCaldarelli:
Mainly for two reasons: first to have a normalized answer between multiple and single response, first with "items" and last with "item"; second reason is to have a container for all fields of item, so if i had extra field to send as answer i can add it to response without mixing with item's field.

EagleMoor:
Single item it's a single item. We gat him from same url and we know what type object we ask.
Maybe it's need for socket, because in socket we have asynchronous requests and don't have url.

I developed rest api 1 year for android & iOS. I try use "item" for Single item, but this and problem on mobile device. For parsing "root" element (as in xml) whey need use same class.

If we have collection with pagination 
```json
{
  "users": [],
  "_links": {},
  "_meta": {}
}
```
Whey need create 4 class: UserCollection (root), User (item of user), Links, Meta.
UserCollection have rules for parsing "users" as Array of User. It's justified because we have a lot of objects in root object.
If we have one item with key
```json
{
  "users": {}
}
```
Whey need use UserCollection (root) and User (user) object too.

But if we write Single object without key

```json
{
 "id": 124,
 "login": "User1"
}
```

Whey need use only one class User (like a root).

I hope that answered your question. =)

FabrizioCaldarelli:
[...]

PR #9702, opened 2015-09-16
#7534 Add `$singleEnvelope` property to `\yii\rest\Serializer`

tunecino:


tunecino:
This is too old. it was related to https://github.com/yiisoft/yii2/issues/7534. But it doesn't seem to be highly requested. not even used by me as I prefer parsing headers where the same output is present. 

Maybe it is better to keep things as they are for now and rethink about all those tiny changes for REST when releasing 2.1 as there is a talk about a separate repo for it now. 

The same applies for https://github.com/yiisoft/yii2/issues/9705 ans its PR https://github.com/yiisoft/yii2/pull/9706. Feel free to close both unless requested by others or you think it is needed. In that case let me know and I'll add all the required unit tests.

samdark:
OK. Thanks for letting us know. Closing related issues and PRs for now.

PR #9709, opened 2015-09-17
[WIP] Length and weight formatters

### LINKED ISSUE #9703: Provide a Formatter for distance
Please provide a formatter for distance, e.g. given a value in meters:
891 -> 891 m
2305 -> 2.3 km
86397 -> 86.4 km

(km: kilometers)

nineinchnick:
Resolves #9703.

Still missing:
- unit tests
- changelog entry

I'm waiting for comments.

Current implementation:
- adds length and weight units
- supports imperial and metric systems
- allows to set default system of units
- allows to change system of units and base unit when calling the formatters
- allows to override default units configuration to add new systems and unit formatters


framework/i18n/Formatter.php
```diff
@@ -1112,18 +1112,128 @@ public function asSize($value, $decimals = null, $options = [], $textOptions = [
...
+
+        switch ($position) {
+            case 0:  return Yii::t('yii', '{nFormatted} mm', $params, $this->locale);
+            case 1:  return Yii::t('yii', '{nFormatted} m', $params, $this->locale);
+            default: return Yii::t('yii', '{nFormatted} km', $params, $this->locale);
+        }
+    }
+
+    /**
+     * Formats the value in grams as a weight in human readable form for example `12 kilograms`.
```
klimov-paul: Why so? In SI the canonical unit of weihgt if 'kligram' - not a gram.

framework/i18n/Formatter.php
```diff
@@ -1112,18 +1112,128 @@ public function asSize($value, $decimals = null, $options = [], $textOptions = [
         }
     }
 
+    /**
+     * Formats the value in millimeters as a length in human readable form for example `12 meters`.
```
klimov-paul: Why in 'millimeters'?
In 'SI' canonical unit of length is 'metre'.

nineinchnick:
If this gets rejected as too specific maybe at least the `formatNumber` method can be made protected (with proposed changes) to allow using it when extending the Formatter class.

framework/i18n/Formatter.php
```diff
@@ -1112,18 +1112,128 @@ public function asSize($value, $decimals = null, $options = [], $textOptions = [
...
+        if ($value === null) {
+            return $this->nullDisplay;
+        }
+
+        list($params, $position) = $this->formatNumber($value, $decimals, 2, 1000, $options, $textOptions);
+
+        switch ($position) {
+            case 0:  return Yii::t('yii', '{nFormatted} {n, plural, =1{millimeter} other{millimeters}}', $params, $this->locale);
+            case 1:  return Yii::t('yii', '{nFormatted} {n, plural, =1{meters} meters}}', $params, $this->locale);
+            default: return Yii::t('yii', '{nFormatted} {n, plural, =1{kilometers} kilometers}}', $params, $this->locale);
```
cebe: what about centimeters and decimeters?

cebe:
this approach is too specific and limited in usage, we should think of making it more general.

nineinchnick:
There could be a new property `siUnits` holding steps for each unit formatter, like:
~~~php
[
    'length' => [
        'imperial' => [
            1 => '{n, plural, =1{inch} other{inches}',
            12 => '{n, plural, =1{foot} other{feet}',
            36 => '{n, plural, =1{yard} other{yards}',
            // ...
        ],
        'metric' => [
            1 => '{n, plural, =1{millimeter} other{millimeters}',
            100 => '{n, plural, =1{centimeter} other{centimeters}',
            1000 => '{n, plural, =1{meter} other{meters}'
            // ...
        ],
    ],
]
~~~

Using that, the keys can be passed instead of a single constant as`$formatBase` in `formatNumber()`.

This would allow to:

* switch the formatter between different systems of units,
* configure your own
* set the base unit
* extend with new units

nineinchnick:
The base unit should actually be a formatter argument with a default taken from a property.

cebe:
have you checked whether this kind of formatting information is available in CLDR or intl extension?

nineinchnick:
@cebe I've looked around and unit translations are available in CLDR. There are some classes calles `MeasureUnits` in ICU but I haven't found anything in the intl library to access that data.

I've updated the PR to include the imperial system. It starts to look messy, mainly because of the `Yii::t()` usage. I'm all for using intl for translations here but it's not possible at the moment.

So we either close this PR and optionally move it to an extension to point people to if they need such formatters or keep it open and wait for changes in intl.

I'm also updating this PR description to sum up the changes.

framework/i18n/Formatter.php
```diff
@@ -1167,6 +1289,133 @@ private function formatSizeNumber($value, $decimals, $options, $textOptions)
...
+                switch ($system) {
+                    case self::UNIT_SYSTEM_IMPERIAL:
+                        return [1, 27.34375, 437.5, 7000, 98000, 196000, 784000, 15680000];
+                    case self::UNIT_SYSTEM_METRIC:
+                        return [1, 1000, 1000000];
+                }
+        }
+    }
+
+    protected function getMeasureUnit($unit, $system, $type, $baseUnit, $params)
```
nineinchnick: This method was added to group all unit translations for easier extending. Unfortunately, because of the `Yii::t()` usage, that doesn't make much sense - it can't be just an array. So all translations can be moved back to specific formatters.

samdark:
@nineinchnick check http://slides.rmcreative.ru/2015/i18n-l10n/#/27

nineinchnick:
@samdark thanks! It looks like it's all there. I'm going to update the PR soon to use that data.

nineinchnick:
@samdark @cebe @klimov-paul looks like all current issues has been resolved. Remaining questions are:

* How to ease adding new systems and changing base unit via configuration?
* Should the `format()` method detect units that exist in configuration but do not have their own methods? Should specific unit method be dropped in favor of this?
* Should the byte formatter be converted to this implementation?

samdark:
Is there data for byte formatter as well?

nineinchnick:
@samdark yes, but not for the 1024 base (kibibyte). And it only goes up to a terabyte. So I've probably answered myself.

samdark:
Also the con of using intl stuff is that it is less reliable. It could be missing, ICU could be outdated, intl itself has bugs i.e. if it could be done without it and is OK for all the languages and variations than it's great. If not, as in this case, intl is OK to use.

As an alternative we can work with CLDR data the way we did in Yii 1.1 but I don't think it's a good idea because of amount of raw data.

nineinchnick:
Yeah even the basic unit test I added fails because of intl issues.

On the other hand, using our own translations instead of intl equals to using CLDR. Such translations can be generated from CLDR instead of created manually, just for a subset for used units.

samdark:
For 2.1 we may consider re-inventing intl using PHP + CLDR data... all those issues.

nineinchnick:
I hope as a library separate from the framework, just like (first google hit) [commerceguys/intl](https://github.com/commerceguys/intl).

samdark:
Yes. I don't like too coupled general-purpose libraries anymore...

framework/i18n/Formatter.php
```diff
@@ -219,11 +226,74 @@ class Formatter extends Component
...
+                'foot' => 12,
+                'yard' => 36,
+                'chain' => 792,
+                'furlong' => 7920,
+                'mile' => 63360,
+            ],
+            self::UNIT_SYSTEM_METRIC => [
+                'millimeter' => 1,
+                'centimeter' => 10,
+                'meter' => 1000,
```
cebe: shouldn't meter be the base unit = 1?

framework/i18n/Formatter.php
```diff
@@ -219,11 +226,74 @@ class Formatter extends Component
      * Defaults to 1024.
      */
     public $sizeFormatBase = 1024;
+    /**
+     * @var string default system of measure units.
+     */
```
cebe: `@since` PHPdoc annotation is needed here.

framework/i18n/Formatter.php
```diff
@@ -219,11 +226,74 @@ class Formatter extends Component
...
+    public $systemOfUnits = self::UNIT_SYSTEM_METRIC;
+    /**
+     * @var array configuration of measure units
+     */
```
cebe: `@since` PHPdoc annotation is needed here.

framework/i18n/Formatter.php
```diff
@@ -219,11 +226,74 @@ class Formatter extends Component
...
+            self::UNIT_SYSTEM_METRIC => [
+                'gram' => 1,
+                'kilogram' => 1000,
+                'ton' => 1000000,
+            ],
+        ],
+    ];
+    /**
+     * @var array base units as multipliers for smallest possible unit from $measureUnits
+     */
```
cebe: `@since` PHPdoc annotation is needed here.

framework/i18n/Formatter.php
```diff
@@ -1174,18 +1244,157 @@ public function asSize($value, $decimals = null, $options = [], $textOptions = [
...
+     * @param double $baseUnit unit of value as the multiplier of the smallest unit
+     * @param string $system either self::UNIT_SYSTEM_METRIC or self::UNIT_SYSTEM_IMPERIAL, if null defaults to metric
+     * @param integer $decimals the number of digits after the decimal point.
+     * @param array $options optional configuration for the number formatter. This parameter will be merged with [[numberFormatterOptions]].
+     * @param array $textOptions optional configuration for the number formatter. This parameter will be merged with [[numberFormatterTextOptions]].
+     * @return string the formatted result.
+     * @throws InvalidParamException if the input value is not numeric or the formatting failed.
+     * @see asWeight
+     */
+    public function asShortWeight($value, $baseUnit = null, $system = null, $decimals = null, $options = [], $textOptions = [])
```
cebe: these methods have a bit too many options, which would override global formatter settings. Should only allow overriding those that are commonly expected to differ from global settings.

cebe (changes requested):
Overall a great addition, needs a bit of work to be mergeable. @nineinchnick are you willing to work on this further? If not I'll take over and fix it.

nineinchnick:
@cebe go ahead and take over.

framework/i18n/Formatter.php
rob006: @cebe Do we really want to involve floats in this? 
https://3v4l.org/BPJ32

cebe: probably not, but a millimeter is not the smallest unit...

samdark: "atto" is...

SilverFire:
@nineinchnick could you allow to change your branch?

nineinchnick:
@SilverFire how do I do that? Is there a new Github feature I don't know about? You don't really need to ask me for permission, you can just fork my fork and continue the work.

nineinchnick:
Oh just saw the checkbox on the right.

SilverFire:
Thank you very much for the contribution @nineinchnick

PR #5704, opened 2014-10-23
Added asDuration() function to yii\i18n\Formatter.php

lindelius:
Closes #5146 

With the weird outputs for certain locales, and the lack of other units than hours, minutes, and seconds in the output no matter how big the input is, I gotta say that I don't see any point in using the PHP intl extension. At least not for durations (haven't checked the other functions yet).

I only managed to get two different outputs from the intl extension no matter what locale and how big the input number was, and for `123456789.12345` that was either `123 456 789` or `34,294:33:09`. My proposed code uses the "correct" intl extension format (`h:mm:ss`) as default but it also supports days, weeks, and years, in the customizable output format.

I'm not too familiar with Yii2 yet, so I'm not sure if there are any classes or methods that could be used to improve the code or make it more customizable, but it works and might be something we can use as base and then improve upon.

PR #9710, opened 2015-09-17
[WIP] Interval formatter

### LINKED ISSUE #5146: Add yii\i18n\Formatter::asDuration()
` NumberFormatter::DURATION (integer) `

http://php.net/manual/en/class.numberformatter.php

nineinchnick:
Resolves #5146.

Still missing:
- messages
- unit tests
- changelog entry

In contrast to #5704 it's more similar to the relativeTime formatter.

Waiting for comments.

Interesting fact: PostgreSQL has an interval format and can be configured to display it using ISO8601 format, same as PHP's DateInterval.


framework/i18n/Formatter.php
```diff
@@ -787,6 +787,50 @@ public function asRelativeTime($value, $referenceTime = null)
...
+            $parts[] = Yii::t('yii', '{delta, plural, =1{an hour} other{# hours}}', ['delta' => $interval->h], $this->locale);
+        }
+        if ($interval->i > 1) {
+            $parts[] = Yii::t('yii', '{delta, plural, =1{a minute} other{# minutes}}', ['delta' => $interval->i], $this->locale);
+        }
+        if ($interval->s > 0) {
+            $parts[] = Yii::t('yii', '{delta, plural, =1{a second} other{# seconds}}', ['delta' => $interval->s], $this->locale);
+        }
+
+        return empty($parts) ? $this->nullDisplay : (($negative ? '-' : '') . implode(', ', $parts));
```
SilverFire: `$parts` concatenation string must be configurable

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,50 @@ public function asRelativeTime($value, $referenceTime = null)
...
+            if (strpos($value, '-') !== false) {
+                $negative = true;
+                $interval = new \DateInterval(str_replace('-', '', $value));
+            } else {
+                $negative = false;
+                $interval = new \DateInterval($value);
+            }
+        }
+        if ($interval->y > 1) {
+            $parts[] = Yii::t('yii', '{delta, plural, =1{a year} other{# years}}', ['delta' => $interval->y], $this->locale);
```
SilverFire: I think you've copied `a year`, `a month`, `an hour`, etc from `Formatter::asRelevantTime`. This verbal form is used, when you are talking about a relevant time - `in an hour`, `a second ago`. In `asInterval` it's a wrong verbal form.
Replace all `a`, `an` with `#` and you will get correct `1 hour, 1 minute` or `1 day, 1 hour, 35 seconds` instead of `an hour, a minute, 31 seconds`

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,50 @@ public function asRelativeTime($value, $referenceTime = null)
         }
     }
 
+    /**
+     * Formats the value as a time interval.
+     * @param mixed $value the value to be formatted, in ISO8601 format.
```
SilverFire: `mixed` - in not clear enough. `\DateInterval|integer` is better

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,50 @@ public function asRelativeTime($value, $referenceTime = null)
...
+    public function asInterval($value)
+    {
+        if ($value === null) {
+            return $this->nullDisplay;
+        }
+        if ($value instanceof \DateInterval) {
+            $negative = $value->invert;
+            $interval = $value;
+        } else {
+            if (strpos($value, '-') !== false) {
```
SilverFire: I'd prefer to treat `$value` as an `interger` and check whether `$value < 0`

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,50 @@ public function asRelativeTime($value, $referenceTime = null)
...
+                $negative = true;
+                $interval = new \DateInterval(str_replace('-', '', $value));
```
SilverFire: If we are treating `$value` as an `integer`, we can use `abs($value)`

framework/i18n/Formatter.php
SilverFire: As for me, ISO8601 is not convenient to format intervals.
I'd prefer simple format - in seconds.

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,50 @@ public function asRelativeTime($value, $referenceTime = null)
         }
     }
 
+    /**
+     * Formats the value as a time interval.
+     * @param mixed $value the value to be formatted, in ISO8601 format.
+     * @return string the formatted result.
+     */
+    public function asInterval($value)
```
SilverFire: Maybe, it's better to rename method to `asDuration`? Interval is more about relative time

SilverFire:
@cebe Could you, please, review my comments?

framework/i18n/Formatter.php
nineinchnick: How?

framework/i18n/Formatter.php
nineinchnick: Isn't _an hour_ the same as _1 hour_?

framework/i18n/Formatter.php
SilverFire: Method can get additional argument

framework/i18n/Formatter.php
nineinchnick: It could be detected if it's a valid ISO format or a number. But for numbers there should be an optional argument with some kind of denominator if somebody uses a base unit different than second (for example, millisecond).

framework/i18n/Formatter.php
nineinchnick: If it's an ISO format it can't be treated as an integer, see my previous comment.

framework/i18n/Formatter.php
SilverFire: It is, but the formatted result will look very inconsistently.
Compare:
`1 day, 1 hour, 15 minutes, 1 second`
and
`A day, an hour, 15 minutes, a second`

framework/i18n/Formatter.php
SilverFire: For example we have a stored phone call. We have 2 use cases, how we can show it's duration.

* We store a real duration of call in seconds.
* We have start and end datetime and every time calculate the difference between 2 dates.

For 2'nd case, you have to pass 2 dates to calculate the duration OR you have to prepare the duration in seconds and use it.

I think, that seconds is more correct way to operate with intervals

framework/i18n/Formatter.php
nineinchnick: Can you give an example? I don't get it.

SilverFire: `public function asDuration($value, $implodeString = ', ')`
I'm not sure, that this argument name is very good

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,51 @@ public function asRelativeTime($value, $referenceTime = null)
...
+    {
+        if ($value === null) {
+            return $this->nullDisplay;
+        }
+        if ($value instanceof \DateInterval) {
+            $negative = $value->invert;
+            $interval = $value;
+        } elseif (is_numeric($value)) {
+            $negative = $value < 0;
+            $interval = new \DateInterval('PT' . abs($value) . 'S');
```
SilverFire: Just found, that this will not work properly. You will have only count of seconds instead of splitting to days, hours, minutes etc.

```php
$interval = new \DateInterval('PT300S');
echo $interval->s; // 300
echo $interval->m; // 0
```

nineinchnick:
Ah yes, but number of days in a month depends on the date, so spreading number of seconds over higher units depends on a start date. I'm not shure how to deal with it.

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,51 @@ public function asRelativeTime($value, $referenceTime = null)
...
+        } elseif (strpos($value, '-') !== false) {
```
SilverFire: This will result in a problem with ISO8601 date. `2016-08-09T18:31:42` has a minus symbol, but the date is not negative.

nineinchnick: You can't pass dates here at all.

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,51 @@ public function asRelativeTime($value, $referenceTime = null)
...
+            $negative = true;
+            $interval = new \DateInterval(str_replace('-', '', $value));
+        } else {
+            $negative = false;
+            $interval = new \DateInterval($value);
+        }
+        if ($interval->y > 1) {
+            $parts[] = Yii::t('yii', '{delta, plural, =1{1 year} other{# years}}', ['delta' => $interval->y], $this->locale);
```
SilverFire: Why did you changed it to `1 year` instead of `# year`?

framework/i18n/Formatter.php
SilverFire: `2016-08-09T18:31:42` is also a ISO8601-formatted time. So we need to fix PHPDoc

framework/i18n/Formatter.php
SilverFire: Try this

```php
$zeroDateTime = (new DateTime())->setTimestamp(0);
$valueDateTime = (new DateTime())->setTimestamp($value);
$interval = $valueDateTime->diff($zeroDateTime);
```

framework/i18n/Formatter.php
nineinchnick: Because it can only be the value of 1?

framework/i18n/Formatter.php
nineinchnick: I know you can do that but it's up to the user what start date he chooses. You can't force current time here.

nineinchnick: At least it would result in constant results and if anybody wants to use another start date he can create the DateInterval object himself.

framework/i18n/Formatter.php
SilverFire: Yes :fearful:

framework/i18n/Formatter.php
SilverFire: I think that a use case with another start date is more rare, than a use case with seconds.
When you have seconds - why do you have to copy-paste 3 lines of code every time, when you want to format duration?

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,53 @@ public function asRelativeTime($value, $referenceTime = null)
...
+        }
+        if ($value instanceof \DateInterval) {
+            $negative = $value->invert;
+            $interval = $value;
+        } elseif (is_numeric($value)) {
+            $negative = $value < 0;
+            $zeroDateTime = (new DateTime())->setTimestamp(0);
+            $valueDateTime = (new DateTime())->setTimestamp(abs($value));
+            $interval = $valueDateTime->diff($zeroDateTime);
+        } elseif (strpos($value, '-') !== false) {
```
SilverFire: Previous condition covers this one. Maybe, we can get red of it?

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,53 @@ public function asRelativeTime($value, $referenceTime = null)
         }
     }
 
+    /**
+     * Formats the value as a time interval.
+     * @param \DateInterval|string|integer $value the value to be formatted, in ISO8601 duration format or number of seconds.
```
SilverFire: There is a `use DateInterval` at the top of formatter. `\` is not needed.

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,53 @@ public function asRelativeTime($value, $referenceTime = null)
...
+     * @return string the formatted result.
+     */
+    public function asDuration($value)
+    {
+        if ($value === null) {
+            return $this->nullDisplay;
+        }
+        if ($value instanceof \DateInterval) {
```
SilverFire: There is a `use DateInterval` at the top of formatter. `\` is not needed.

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,53 @@ public function asRelativeTime($value, $referenceTime = null)
...
+            $negative = $value->invert;
+            $interval = $value;
+        } elseif (is_numeric($value)) {
+            $negative = $value < 0;
+            $zeroDateTime = (new DateTime())->setTimestamp(0);
+            $valueDateTime = (new DateTime())->setTimestamp(abs($value));
+            $interval = $valueDateTime->diff($zeroDateTime);
+        } elseif (strpos($value, '-') !== false) {
+            $negative = true;
+            $interval = new \DateInterval(str_replace('-', '', $value));
```
SilverFire: There is a `use DateInterval` at the top of formatter. `\` is not needed.

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,53 @@ public function asRelativeTime($value, $referenceTime = null)
...
+        } else {
+            $negative = false;
+            $interval = new \DateInterval($value);
```
SilverFire: There is a `use DateInterval` at the top of formatter. `\` is not needed.

framework/i18n/Formatter.php
nineinchnick: No, it doesn't. Previous conditions checks if this is a digit. This condition checks if this is a ISO duration format starting with a dash.

framework/i18n/Formatter.php
nineinchnick: Yeah you dont't have to repeat it 4 times.

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,53 @@ public function asRelativeTime($value, $referenceTime = null)
         }
     }
 
+    /**
+     * Formats the value as a time interval.
+     * @param \DateInterval|string|integer $value the value to be formatted, in ISO8601 duration format or number of seconds.
+     * @return string the formatted result.
```
SilverFire: Add `@since 2.0.7`

framework/i18n/Formatter.php
SilverFire: Agree

framework/i18n/Formatter.php
SilverFire: Got it, sorry

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,54 @@ public function asRelativeTime($value, $referenceTime = null)
         }
     }
 
+    /**
+     * Formats the value as a time interval.
+     * @param DateInterval|string|integer $value the value to be formatted, in ISO8601 duration format or number of seconds.
+     * @return string the formatted result.
+     * @since 2.0.7
+     */
+    public function asDuration($value)
```
SilverFire: Let's add `$implodeString = ', '`, `$negativePrefix = '-'` method arguments

nineinchnick: Why `$negativePrefix`? When can it be anything else than a dash?

SilverFire: Maybe some HTML `span` with icon, etc

nineinchnick: Oh, you should comment it on line 835 containing the return statement. I thought you mean the dash used in `str_replace()` calls.

SilverFire: Okay, that's my fault, sorry

SilverFire:
@nineinchnick Let's finish it

nineinchnick:
@SilverFire what do you mean? Can you write a summary what's left? I'm waiting for approval from core devs.

SilverFire:
- `$implodeString, $negativePrefix` in this discussion: https://github.com/yiisoft/yii2/pull/9710#discussion_r41011367
 - Unit tests
 - Changelog

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,54 @@ public function asRelativeTime($value, $referenceTime = null)
...
+    {
+        if ($value === null) {
+            return $this->nullDisplay;
+        }
+        if ($value instanceof DateInterval) {
+            $negative = $value->invert;
+            $interval = $value;
+        } elseif (is_numeric($value)) {
+            $negative = $value < 0;
```
samdark: negative → isNegative

framework/i18n/Formatter.php
samdark: Shouldn't it check for `=== 0`?

framework/i18n/Formatter.php
```diff
@@ -787,6 +787,54 @@ public function asRelativeTime($value, $referenceTime = null)
...
+            $zeroDateTime = (new DateTime())->setTimestamp(0);
+            $valueDateTime = (new DateTime())->setTimestamp(abs($value));
+            $interval = $valueDateTime->diff($zeroDateTime);
+        } elseif (strpos($value, '-') !== false) {
+            $negative = true;
+            $interval = new DateInterval(str_replace('-', '', $value));
```
samdark: Are we replacing only the first char?

samdark:
Overall approach is good.

SilverFire:
@nineinchnick also we need to add the description of fornatter to the guide https://github.com/yiisoft/yii2/blob/master/docs/guide/output-formatting.md

SilverFire:
@nineinchnick OK, I'll continue.

PR #9717, opened 2015-09-17
update comments in \yii\base\Controller

ghost:
I had problems with understanding of why AccessControll filter didn't work, as i wish, with PageCache. Then i remembered the quote "The validation rules are evaluated in the order they are listed." from guide http://www.yiiframework.com/doc-2.0/guide-input-validation.html#declaring-rules, but when i tried to redirect some certain user using beforeAction(), as you advised, it didn't work.


cebe:
Thanks!

PR #9728, opened 2015-09-19
Added clearErrors parameter in validateMultiple()

hector-del-rio:


hector-del-rio:
@cebe I think this is not a BC braking issue... right?

samdark:
It is. Public method signature was changed.

hector-del-rio:
But the default value of the new parameter $clearErrors is the same as in ```yii\base\model::validate()``` method. The behaviour is the same than previously if you do not add that parameter, so no BC is broken.

samdark:
It is broken since if you have any class that extends from `Model` and overrides `validateMultiple` method, it will break.

hector-del-rio:
I am sorry, but I still can't see how. Could you spare few minutes to write an example?

samdark:
```php
class MyForm extends Model
{
    public function validateMultiple($models, $attributeNames = null)
    {
        // whatever
    }
}
```

would cause the following error:

```
Strict standards: Declaration of MyForm::validateMultiple() should be compatible with Model::validateMultiple($models, $attributeNames = null, $clearErrors = true)
```

hector-del-rio:
Ok, I see. Thank you for the clarification @samdark. Learning something new everyday. :)

PR #9733, opened 2015-09-19
Add Unprocessable Entity Exception as in #2106

janjitsu:
Continue the line of tought in #2106

I want to be able to send 422 Http Error to inform form validation errors on my RESTful Api

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.7 under development
 -----------------------
 
+- Enh #2106: Added Unprocessable Entity Http Exception (janfrs)
 - Bug #8723: Fixed `yii\helpers\VarDumper::export()` unable to export circle referenced objects with `Closure` (klimov-paul)
 - Bug #9108: Negative number resulted in no formatting when using `Formatter::asSize()` or `Formatter::asShortSize` (nxnx, cebe)
 - Bug #9288: Fixed `FileHelper::createDirectory` directory creation to be concurrency friendly (dynasource)
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,6 @@ Yii Framework 2 Change Log
 2.0.7 under development
 -----------------------
 
-- Enh #2106: Added Unprocessable Entity Http Exception (janfrs)
 - Bug #8723: Fixed `yii\helpers\VarDumper::export()` unable to export circle referenced objects with `Closure` (klimov-paul)
 - Bug #9108: Negative number resulted in no formatting when using `Formatter::asSize()` or `Formatter::asShortSize` (nxnx, cebe)
 - Bug #9288: Fixed `FileHelper::createDirectory` directory creation to be concurrency friendly (dynasource)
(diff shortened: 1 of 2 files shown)


samdark:
Are you sure it's correct code for form validation errors?

framework/web/UnprocessableEntityHttpException.php
```diff
@@ -0,0 +1,35 @@
...
+ * exception with status code 422.
+ *
+ * Use this exception to inform that the server understands the content type of
+ * the request entity and the syntax of that request entity is correct but was 
+ * unable to process the contained instructions. For example, to return form 
+ * validation errors. 
+ *
+ * @link http://www.webdav.org/specs/rfc2518.html#STATUS_422
+ * @author Jan Silva <janfrs3@gmail.com>
+ * @since 2.0
```
Renkas: I'm pretty surethis Exception is not available in version 2.0 of Yii. Should change to the real release version - 2.0.7 probably

Renkas:
AFAIK there is no correct "Validation failed" HTTP response code defined and this 422 is commonly used for that.

samdark:
I saw 400 used for the purpose.

Renkas:
400 would be fine if the data sent to server is syntactically wrong eg. malformed XML/json etc. But if the syntax of sent data is right then the proper response would still be 422.

janfrs:
I agree with @Renkas, but I also agree with @samdark, AFAIK there's an ongoing debate to what is the more appropriate error code to send on form validation errors, as you can see [here](http://stackoverflow.com/questions/16133923/400-vs-422-response-to-post-of-data), [here](http://www.bennadel.com/blog/2434-http-status-codes-for-invalid-data-400-vs-422.htm) and [here](http://stackoverflow.com/questions/1959947/whats-an-appropriate-http-status-code-to-return-by-a-rest-api-service-for-a-val) 

I, personally, want my API to respond differently when the client is sending all expected fields, but some information on them isn't right. For that I find 422 (UnprocessableEntity) a more suitable response.

framework/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.7 under development
 -----------------------
 
+- Enh #2106: Added Unprocessable Entity Http Exception (janfrs)
```
samdark: Should be grouped with other `Enh` and ordered according to its number.

framework/web/UnprocessableEntityHttpException.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+namespace yii\web;
+
+/**
+ * UnprocessableEntityHttpException represents an "Unprocessable Entity" HTTP 
+ * exception with status code 422.
+ *
+ * Use this exception to inform that the server understands the content type of
+ * the request entity and the syntax of that request entity is correct but was
```
samdark: correct but was → correct but **server** was

samdark:
I'm OK adding it.

samdark:
Thanks!

cebe:
FYI: https://github.com/yiisoft/yii2/commit/73d6f32234057cab4d097773f303e0fd8c0388b4

PR #9746, opened 2015-09-21
Validation of authKey for existing session

### LINKED ISSUE #9718: User stays authorized despite auth key is changed
While renewAuthStatus() method of yii\web\user is getting user id from session variable, it's not checking auth key. If auth key changes, the user still will be authorized while his id is stored in session.

Steps to reproduce this issue with Advanced application template:
1. Add `$user->generateAuthKey();` in resetPassword method of frontend\models\ResetPasswordForm directly after `$user->setPassword($this->password);`
2. Login as any user in browser
3. Open new inconito tab
4. Login as same user in incognito tab
5. Change password in incognito tab
6. Check that you are still authorized in all tabs

kidol:
WIP
Fixes #9718

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -27,6 +27,7 @@ Yii Framework 2 Change Log
 - Enh #9635: Added default CSS class for `\yii\grid\ActionColumn` header (arogachev, dynasource)
 - Enh #9643: Added migrations for DB cache (mdmunir)
 - Enh #9711: Added `yii\widgets\LinkPager::$pageCssClass` that allows to set default page class (ShNURoK42)
+- Enh #9718: The authKey-feature of `yii\web\IdentityInterface` is now also used to validate active sessions (kidol)
 - Enh #9733: Added Unprocessable Entity HTTP Exception (janfrs)
 - Chg #9369: `Yii::$app->user->can()` now returns `false` instead of erroring in case `authManager` component is not configured (creocoder)
 - Chg #9411: `DetailView` now automatically sets container tag ID in case it's not specified (samdark)
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -8,6 +8,15 @@ if you want to upgrade from version A to version C and there is
 version B between A and C, you need to follow the instructions
 for both A and B.
 
+Upgrade from Yii 2.0.6
+----------------------
+
+* The methods `getAuthKey()` and `validateAuthKey()` of `yii\web\IdentityInterface` are now also used to validate active sessions
+  (previously these methods were only used for cookie-based login). If your identity class does not properly implement these methods yet,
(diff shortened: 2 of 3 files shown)


samdark:
Looks OK to me.

@yiisoft/core-developers any comments?

@kidol is it still WIP? Do you have something in mind to improve?

klimov-paul:
Instead of using `authKey` for this purpose - a new Identity parameter should be added.
Most obvious and quick solution for the original issue can be storing and comparing of the 'password_hash'. Once password has been changed by any mean - user will have to re-login.

However introduction of the new Identity property requires changing of the `IdentityInteface` signature and BC break.

samdark:
Why authkey couldn't be used for the purpose?

kidol:
Well currently docs all say authKey is only needed for cookie.
Also not sure whether session and/or identityCookie should be destroyed when wrong authKey is detected (currently log gets filled with warnings on every pageview).
There is probably more.

Will look into it soon.

klimov-paul:
> Why authkey couldn't be used for the purpose?

It could. However, the original issue sounds: "after user changes his password all other sessions should be dropped". The most obvious solution will be compare `password_hash` sinse it changes once user change his password, while in current state you should regenerate `authKey` on new password setup.
I am saying that we should introduce another method `IdentityInterface::getRestoreValidationKey()`. It can return `authKey` or anything else - `password_hash` is most obvious here.

Also remember that original purpose of the `authKey` is different. Even more it can be left unimplemented for the particular identity, if application does not provides 'auto login' feature.
These should not be mixed up.

kidol:
Original issue title is:
>User stays authorized despite auth key is changed

So if he changes authKey because of new password or some other reason, it doesn't matter. He wants any active session to be invalid and that's what authKey can be used for (at least for cookie-login currently).

>Also remember that original purpose of the authKey is different.

Yes, but current implementation is not logical: The identity gets fetched every time you access session, but only cookie-login uses authKey.

>Even more it can be left unimplemented for the particular identity, if application does not provides 'auto login' feature.

If there is such app, then break BC for them and note in upgrade file: If they don't want to implement authKey feature, they can simply return `null` in `getAuthKey()`. This will keep current behavior.
Also currently they already have to provide the authKey related methods because it's an interface.

kidol:
One problem is if application changes underlying authKey for an active session. For example, if app has button "invalidate all other sessions" which will update authKey in db, then current session will still have old authKey in next pageload. However, user can call `yii\web\User::switchIdentity()` to recreate session with updated authKey. This is noted in upgrade file.

@laendyl Were you able to test this patch yet?

kidol:
@klimov-paul If you look at my explanation and current diff, you still have doubts?

@undestroyer, @dynasource Could you help test this patch?

dynasource:
I will when I find time. thx

kidol:
Most likely outdated, also no interest in working further on this.

PR #9749, opened 2015-09-22
canonical url needs to ignore specified params

sinaru:
Canonical URL should not capture all the parameters provided in the URL. So we need to specify what parameters to ignore when generating the canonical URL. For example, on a product listing page I would need to ignore a "sort"parameter and point all the sorting links to the one specific product page. 
Ex: 
1. www.mysite.com/mens/dresses?sort=by_price
2. www.mysite.com/mens/dresses?sort=by_date

Canonical URL for above two samples should be.
www.mysite.com/mens/dresses

So I need to tell to ignore 'sort' param when generating the canonical url. That is why this change is proposed.

### DIFF
--- a/framework/helpers/BaseUrl.php
+++ b/framework/helpers/BaseUrl.php
@@ -296,13 +296,20 @@ class BaseUrl
      * ```php
      * $this->registerLinkTag(['rel' => 'canonical', 'href' => Url::canonical()]);
      * ```
-     *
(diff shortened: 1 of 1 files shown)


klimov-paul:
This does not make any sense.
`Url::canonical()` composes URL from route and action params, which are declared as action method arguments.
You should either remove 'sort' from action arguments or compose URL manually instead of usage of `Url::canonical()`.

Issue #9689, opened 2015-09-15
Allow setting name attribute in hidden input file upload


graphcon:
Hi Yii2 team.

Currently, one can rename input fields in ActiveForm by passing the name key in $options array. This does work (as I've seen it) in every field but not for the hidden field in a file upload.

I would suggest as fix in this code line:
https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1317

From:
`return static::activeHiddenInput($model, $attribute, ['id' => null, 'value' => ''])`
To:
```php
$hiddenInputOptions = ['id' => null, 'value' => ''];
if(isset($options['name'])) {
    $hiddenInputOptions['name'] = $options['name'];
}
return static::activeHiddenInput($model, $attribute, $hiddenInputOptions)
```

As this would be the first contribution to this project, I want to first ask whether that is the supposed way and if it suits the code style.

moicirk:
The options array must be used for activeInput only I guess.
Why don`t you just use hiidenInput contatinated with fileInput manually in a template?

graphcon:
@moicirk Well, that is because I'd prefer it when the framework does this conveniently for me :) Concatenating hiddenInput and fileInput manually is a workaround for an issue in the framework, I would say...

cebe:
Fixed, thanks for reporting!

PR #9771, opened 2015-09-26
This is an enhancement in Html::activeFileInput()

HanafiAhmat:
This is an enhancement in Html::activeFileInput() for assigning htmlOptions to the hiddenInput field so that client script could interact directly with it.

In a case where a form uses image upload through ajax like BlueImp, the file input field would be empty since the file was already uploaded before the form was submitted. Therefore, the uploaded file's name could not be retrieved for back end processing after the form is submitted.

Assigning the uploaded file's name to the file input through client script is not possible as it is an insecure operation as shown in this snapshot.
screen shot 2015-09-27 at 2 27 41 am

One of the possible solution is to assign the uploaded file's name to the hidden field which has a specific id or class. 

On top of this, if the form attribute was not specified with an enctype specifically to enable file uploads, there will be two inputs having the same input name in the post request. Since the file input is being rendered after the hidden field, the value of the hidden input will be overridden by the file input value.

By making the htmlOptions of the hidden field configurable, a use case such as described above can be achieved without the need to add an additional hidden field to achieve the requirement.


framework/helpers/BaseHtml.php
```diff
@@ -1302,19 +1302,34 @@ public static function activePasswordInput($model, $attribute, $options = [])
...
+        $value = ArrayHelper::getValue($hiddenInputOptions, 'value');
+        if (empty($value)) {
+            $hiddenInputOptions['value'] = '';
+        }
         // add a hidden field so that if a model only has a file field, we can
-        // still use isset($_POST[$modelClass]) to detect if the input is submitted
-        return static::activeHiddenInput($model, $attribute, ['id' => null, 'value' => ''])
+        // still use isset($_POST[$modelClass]) to detect if the input is submitted.
+        // The hidden input will be assigned its own set of html options via `$hiddenInputOptions`.
+        // This is to provide the possibility to interact with the hidden field via client script.
```
SilverFire: This provides the possibility

framework/helpers/BaseHtml.php
```diff
@@ -1302,19 +1302,34 @@ public static function activePasswordInput($model, $attribute, $options = [])
...
     {
+        $hiddenInputOptions = ArrayHelper::remove($options, 'hiddenInputOptions');
+        $id = ArrayHelper::getValue($hiddenInputOptions, 'id');
+        if (empty($id)) {
+            $hiddenInputOptions['id'] = null;
+        }
+        $value = ArrayHelper::getValue($hiddenInputOptions, 'value');
+        if (empty($value)) {
+            $hiddenInputOptions['value'] = '';
+        }
```
SilverFire: All the previous part can be simplified with the code bellow:

```php
$hiddenInputOptions = ArrayHelper::merge([
    'id' => null,
    'value' => ''
], ArrayHelper::remove($options, 'hiddenInputOptions'));
```

HanafiAhmat:
Thanks @SilverFire. I applied the recommended changes with an extra check. If $hiddenInputOptions was not included in the $options array, an empty array will be passed to ArrayHelper::merge(). This will avoid the foreach loop from throwing a php error.

framework/helpers/BaseHtml.php
```diff
@@ -1302,19 +1302,30 @@ public static function activePasswordInput($model, $attribute, $options = [])
...
+     * from `$options` to be used for the hidden input.
      * @return string the generated input tag
      */
     public static function activeFileInput($model, $attribute, $options = [])
     {
+        $hiddenInputOptions = ArrayHelper::remove($options, 'hiddenInputOptions');
+        $hiddenInputOptions = ArrayHelper::merge([
+            'id' => null,
+            'value' => ''
+        ], (!empty($hiddenInputOptions) ? $hiddenInputOptions : []));
```
SilverFire: May be replaced with `ArrayHelper::remove($options, 'hiddenInputOptions', []);`

cebe:
can you show how you would use this option? I do not really get why it is needed. The hidden input has just one purpose as described in the comment. for all other use cases you should use separate fields.

note also that your changes conflict with https://github.com/yiisoft/yii2/commit/c3bfa5c77ec271d92ffb056583313ad058e6fd29

HanafiAhmat:
Hi @cebe, 

Sorry I just noticed about issue #9689.
I need to assign the hidden input a specific id or class so that client script could interact directly with it. The elaborated scenario was described in my first post.

cebe:
why do you need to use the existing hidden input for this and not an other extra field?

HanafiAhmat:
i just feel that it is more convenient.

without configurable options:
```
<?= $form->field($model, 'imageFile')->fileInput() ?>
<?= $form->field($model, 'imageFile')->hiddenInput(['id'=>'specific-id'])->label(false) ?>
```

with configurable options:
```
<?= $form->field($model, 'imageFile')->fileInput(['hiddenInputOptions'=>['id'=>'specific-id']]) ?>
```

i tried to merge the conflict and came out with this.
```
$hiddenOptions = ['id' => null, 'value' => ''];
if (isset($options['name'])) {
    $hiddenOptions['name'] = $options['name'];
}
$hiddenOptions = ArrayHelper::merge($hiddenOptions, ArrayHelper::remove($options, 'hiddenOptions', []));
```
if no hiddenOptions was defined but option['name'] is defined, it will work as needed by #9689.
if 'name' is defined in hiddenOptions then it will overwrite option['name'].

Is this ok?

cebe:
code looks okay, will check the use case in detail later.

dmirogin (approved):

framework/helpers/BaseHtml.php
```diff
@@ -1430,22 +1430,29 @@ public static function activePasswordInput($model, $attribute, $options = [])
...
      * unless they are explicitly specified in `$options`.
+     * Additionally, if a separate set of html options array is defined inside `$options` with a key named `hiddenOptions`,
+     * it will be passed to the `activeHiddenInput` field as its own `$options` parameter.
      * @param Model $model the model object
      * @param string $attribute the attribute name or expression. See [[getAttributeName()]] for the format
      * about attribute expression.
      * @param array $options the tag options in terms of name-value pairs. These will be rendered as
      * the attributes of the resulting tag. The values will be HTML-encoded using [[encode()]].
      * See [[renderTagAttributes()]] for details on how attributes are being rendered.
+     * If `hiddenOptions` parameter which is another set of html options array is defined, it will be extracted
```
samdark: html → HTML

framework/helpers/BaseHtml.php
```diff
@@ -1430,22 +1430,29 @@ public static function activePasswordInput($model, $attribute, $options = [])
      * Generates a file input tag for the given model attribute.
      * This method will generate the "name" and "value" tag attributes automatically for the model attribute
      * unless they are explicitly specified in `$options`.
+     * Additionally, if a separate set of html options array is defined inside `$options` with a key named `hiddenOptions`,
```
samdark: html → HTML

samdark (approved):

samdark:
@HanafiAhmat would you please allow changes in your pull request?

HanafiAhmat:
@samdark permission changed.

samdark:
Merged. Thank you!

PR #9780, opened 2015-09-28
Allow callback as range in RangeValidator

### LINKED ISSUE #7581: RangeValidator: allow callback as range
In some situations will be useful to pass a callback to `range` param of `RangeValidator`.
Rule configuration array can be set like this:
```php
['city_id', 'in', 'range' => function ($model) use ($cities) { return $cities[$model->country_id]; }]
```
Now to pass value of another attribute of model (via `$this`) we need to use InlineValidator:
```php
['city_id', function ($attribute, $params) use ($cities) {
    if (!in_array($this->$attribute, $cities[$this->country_id])) {
        $this->addError($attribute, Yii::t('app', '{attribute} is not a valid city for specified country.', [
            'attribute' => $this->getAttributeLabel($attribute),
        ]));
    }
}],
```

RomeroMsk:
Fixes #7581

### DIFF
--- a/framework/validators/RangeValidator.php
+++ b/framework/validators/RangeValidator.php
@@ -23,7 +23,15 @@ use yii\base\InvalidConfigException;
 class RangeValidator extends Validator
 {
     /**
-     * @var array list of valid values that the attribute value should be among
+     * @var mixed list of valid values that the attribute value should be among or a PHP callable that returns
+     * such list. The signature of the PHP callable should be as follows,
+     *
+     * ```php
+     * function foo($model, $attribute) {
+     *     // compute range
+     *     return $range;
+     * }
+     * ```
      */
     public $range;
     /**
@@ -47,7 +55,7 @@ class RangeValidator extends Validator
     public function init()
     {
         parent::init();
-        if (!is_array($this->range)) {
+        if (!is_array($this->range) && !($this->range instanceof \Closure)) {
             throw new InvalidConfigException('The "range" property must be set.');
         }
         if ($this->message === null) {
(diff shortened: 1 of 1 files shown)


samdark:
What's your use case?

framework/validators/RangeValidator.php
```diff
@@ -23,7 +23,15 @@
 class RangeValidator extends Validator
 {
     /**
-     * @var array list of valid values that the attribute value should be among
+     * @var mixed list of valid values that the attribute value should be among or a PHP callable that returns
```
samdark: PHP callable !== anonymous function

samdark: Either fix description or fix method itself.

samdark:
Looks good to me except closure !== valid PHP callback... @RomeroMsk was it meant to support all callback types or just closures?

framework/validators/RangeValidator.php
RomeroMsk: It was just copied from here:
https://github.com/yiisoft/yii2/blob/master/framework/validators/DefaultValueValidator.php#L22

samdark: Then it's wrong in `DefaultValueValidator` as well.

RomeroMsk:
Is it ok now?

samdark:
Absolutely.

samdark:
Merged https://github.com/yiisoft/yii2/commit/e93b797098338b349c5da76fd917fe7fd615943b. Thank you.

RomeroMsk:
Thanks!

PR #9801, opened 2015-09-30
Fix links in documentation

nkovacs:
The links on http://www.yiiframework.com/doc-2.0/yii-widgets-inputwidget.html are broken, the generator prefixes them with `\yii\widgets\`, so I think this will fix them. (so would removing `yii\widgets`, but I think it's better to be explicit)


samdark:
Thanks!

Issue #9814, opened 2015-10-02
Events `EVENT_AFTER_INSERT` and `EVENT_AFTER_UPDATE` documented as `Event`, but they are `AfterSaveEvent` in fact


Vovan-VE:
Is `AfterSaveEvent` an internal stuff (so, we cannot rely on it) or is there a documentation bug?

Declaration https://github.com/yiisoft/yii2/blob/master/framework/db/BaseActiveRecord.php#L58 :

```php
/**
 * @event Event an event that is triggered after a record is inserted.
 */
const EVENT_AFTER_INSERT = 'afterInsert';
...
/**
 * @event Event an event that is triggered after a record is updated.
 */
const EVENT_AFTER_UPDATE = 'afterUpdate';
```

Triggering https://github.com/yiisoft/yii2/blob/master/framework/db/BaseActiveRecord.php#L907 :

```php
public function afterSave($insert, $changedAttributes)
{
    $this->trigger($insert ? self::EVENT_AFTER_INSERT : self::EVENT_AFTER_UPDATE, new AfterSaveEvent([
        'changedAttributes' => $changedAttributes
    ]));
}
```

Vovan-VE:
Pull request #9822.

PR #9822, opened 2015-10-03
Fix: Issue #9814: event type EVENT_AFTER_INSERT and EVENT_AFTER_UPDATE

Vovan-VE:
Events `EVENT_AFTER_INSERT` and `EVENT_AFTER_UPDATE` are documented
as `Event`, but infact they are `AfterSaveEvent`.


samdark:
Thanks!

PR #9824, opened 2015-10-03
fix #9823

### LINKED ISSUE #9823: Regression caused by DI controller action injection
As reported at https://github.com/yiisoft/yii2/commit/b7020065c8bdea22878589a57ba4513ff53ea15e:

After you changes i got this mistakes:

```
class TestController extends Controller
{
    public function actionIndex($arg1, $arg2)
    {
        var_dump(func_get_args());
    }
}
```

output:
```
./yii test arg1 arg2 arg3
array(2) {
  [0] =>
  string(4) "arg1"
  [1] =>
  string(4) "arg2"
}
```

Previously there was arg3.

mdmunir:


mdmunir:
#9823

samdark:
What about web controller?

framework/console/Controller.php
```diff
@@ -140,6 +140,10 @@ public function bindActionParams($action, $params)
             throw new Exception(Yii::t('yii', 'Missing required arguments: {params}', ['params' => implode(', ', $missing)]));
         }
 
+        // #9823
```
samdark: No need to reference it since the change wasn't released yet and this is regression fix.

mdmunir:
> What about web controller?

Web controller has no this issue. As default, not all of query params passed to controller action

samdark:
Right. Thank you!

PR #9826, opened 2015-10-04
Fixes #9708

### LINKED ISSUE #9708: Deleting original js/css files from web accessible folder after compression
Would be nice if `asset/compress` command deleted original js/css files that are published to web accessible folder, once compression is performed. They may contain some sensitive information (e.g. extensive comments), and not exposing them is one of the reasons for compression in the first place.

Even though these files are not served to the client browser directly, they are accessible if URL is known. And all you need to know is an 8-symbol hash, which is pretty easy to brute-force, and a relative path/filename, which quite often is easy to guess (e.g. I bet most projects have `/css/main.css`, and sometimes JS file names are similar to function names that you can find in a compressed script).

Are there any reasons not to delete original js/css files that have been compressed? Should be pretty easy to do.

pana1990:
Fixes #9708


klimov-paul:
Resolved by commit 56cc97ae420073ff4086db4119faacfbb610263d

Issue #9775, opened 2015-09-27
Error if role name is 0


Claud:
I have default roles

``` php
const ROLE_SUPER_ADMIN = 0;
const ROLE_ADMIN = 1;
....
```

This check not work

https://github.com/yiisoft/yii2/blob/master/framework/rbac/DbManager.php#L219

samdark:
That's quite weird name for a role... Usualy names are meaningful strings like "role_super_admin".

dynasource:
the whole RBAC architecture is built around strings. Examples:
* the empty() functions
* variable names ($itemName)
* several method arguments
* method docs

A lot has to be changed in the architecture in order to accept integers. 

Its easier for you to implement strings for your constants. This which would not lead to many problems if these are applied consistently. An example:
```php
const ROLE_SUPER_ADMIN = 'super-admin';
const ROLE_ADMIN = 'admin';
```

Please elaborate if integers are of vital importance  in order for this issue to be opened again.

dynasource:
more opinions about this https://github.com/yiisoft/yii2/pull/9832#issuecomment-145531425

PR #9832, opened 2015-10-04
fix rbac issue #9775

korvin0:
Very small fix, change from "empty" to "is_null"

### DIFF
--- a/framework/rbac/DbManager.php
+++ b/framework/rbac/DbManager.php
@@ -216,7 +216,7 @@ class DbManager extends BaseManager
      */
     protected function getItem($name)
     {
-        if (empty($name)) {
+        if (is_null($name)) {
             return null;
         }
 


SilverFire:
What's the purpose of your fix?

cebe:
> correct check - is_string && !empty

why? please elaborate about the problem, what are you doing and what is not working as you expect?

korvin0:
Here https://github.com/yiisoft/yii2/issues/9775 the user complained that he has the role name === 0 and the function getItem doesn't work. By the way, it won't be work also with any permissions with name === 0.

samdark:
Ah, I see. `0` is weird role name.

@cebe do you think it should be fixed?

cebe:
if we want to allow 0 as a role name I am quite sure this is not the only place that needs to be changed. not sure if we should allow it.

samdark:
Personally I see no point allowing it.

PR #9869, opened 2015-10-09
Allow the use of Yii aliases in a SQLite dsn

ASlatius:
Making it possible to use a SQLite DSN like;

```
'db' => [
  'dsn' => 'sqlite:@app/db/database.sqlite3',
]
```

That way it'll always work out of the box and absolute paths aren't
needed.

### DIFF
--- a/framework/db/Connection.php
+++ b/framework/db/Connection.php
@@ -521,6 +521,8 @@ class Connection extends Component
 
         if (empty($this->dsn)) {
             throw new InvalidConfigException('Connection::dsn cannot be empty.');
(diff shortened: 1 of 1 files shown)


SilverFire:
Good idea, I like it. We need a PUPUnit test of the feature.

cebe:

SilverFire:
@cebe May I add a test, or it's not required?

cebe:
good point, had not seen your comment before merge. added a test: 3826b6f

SilverFire:
@cebe thank you

PR #9870, opened 2015-10-09
Changed `View::findViewFile` method visibility

leandrogehlen:
It's necessary for controller customizations

### DIFF
--- a/framework/base/View.php
+++ b/framework/base/View.php
@@ -160,7 +160,7 @@ class View extends Component
      * @throws InvalidCallException if a relative view name is given while there is no active context to
      * determine the corresponding view file.
      */
-    protected function findViewFile($view, $context = null)
+    public function findViewFile($view, $context = null)
     {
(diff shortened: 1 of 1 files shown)


samdark:
Could you elaborate about the use case?

leandrogehlen:
In my case i want to modify `Controller::renderAjax` behavior.
I don't need that `view` renders asset files, but i need of `js` widgets in a specifc case.
For this i did:

```php
public function renderAjax($view, $params = [])
{
    $vw = $this->getView();
    $method = new \ReflectionMethod(View::className(), 'findViewFile');
    $method->setAccessible(true);
    $viewFile = $method->invoke($vw, $view, $this);

    ob_start();
    ob_implicit_flush(false);

    $vw->beginBody();
    echo $vw->renderFile($viewFile, $params, $this);
    echo View::PH_BODY_END;
    $vw->endPage(false);

    return ob_get_clean();
}
```

samdark:
You can use custom View component instead.

PR #9874, opened 2015-10-09
Fix raw response show exception stack trace when YII_DEBUG=false

nainoon:


samdark:
How to reproduce the issue?

nainoon:
```php
define('YII_DEBUG', false);

Yii::$app->response->format = Response::FORMAT_RAW;

throw new \Exception;
```

samdark:
Thanks!

Issue #9879, opened 2015-10-10
Stop using openssl_random_pseudo_bytes()


tom--:
I am uncomfortable with \base\Security::generateRandomKey()'s use of openssl_random_pseudo_bytes(). It's complicated, sorry, but worth discussing.

**Why I don't like OpenSSL's RNG**

TL;DR: I don't properly understand how OpenSSL's RNG works nor how it is configured in PHP but I do know that are [ways that it can be configured](https://wiki.openssl.org/index.php/Random_Numbers) so that it has properties that a CSPRNG must not have.

For example, the only way to be sure that OpenSSL's RNG is properly seeded is to do it yourself from a known-good source, i.e. the kernel or CryptGenRandom. But generateRandomKey() uses openssl_random_pseudo_bytes() as its last-chance source after trying and failing to read from those. So I think it likely in practice that generateRandomKey() uses OpenSSL only when it seeded itself using its fall-back, undocumented, user-space methods that I don't trust.

For another example, OpenSSL can be configured to use, when available, Intel's RDRAND, now [known to be unsafe](https://twitter.com/DefuseSec/status/408975222163795969/photo/1).

**What to do?**

1. Simply removing it will break everything on Windows and some non-Windows deployments.
1. We removed mcrypt_create_iv() when we replaced Mcrypt with OpenSSL. But mcrypt_create_iv() does not use libmcrypt and it is the only good way to access CryptGenRandom in PHP < 7. I think it's ok to use it where available so we could restore it.
1. We could leave openssl_random_pseudo_bytes() in but disabled by default, with a configuration member variable, `Security::$allowOpenSslRng = false` with new documentation to educate the user about the concerns.

Perhaps doing both 2. and 3. is reasonable.

Opinions?

samdark:
Hmm... `mcrypt_create_iv` requires mcrypt dependency so there will be both dependencies on mcrypt and openssl.

cebe:
we could check if function exists and only then use it.

tom--:
I don't want to bring the Mcrypt PHP extension back as a Yii requirement.

But `if (function_exists('mcrypt_create_iv')) {}` is, I think, a better way to get random numbers on Windows that OpenSSL.

SamMousa:
How about not implementing a solution yourselves, but using an existing one:

https://github.com/paragonie/random_compat

That way you can use random_bytes in Yii code and keep it clean and simple.

[...]

Issue #9878, opened 2015-10-10
Use PHP 7 random_bytes() in \base\Security


tom--:
\base\Security::generateRandomKey() should use [random_bytes()](https://secure.php.net/manual/en/function.random-bytes.php) as a first priority when it is available.

samdark:
Agree. Could you work on a pull request?

tom--:
Yes, this one is easy. The next two require some consideration.

tunecino:
a quick question. I don't know how those libs works, so when doing this:
```php
$authKey = \yii\base\Security::generateRandomString();
```
which is built on top of `generateRandomKey()` and maybe `random_byte()` in a future. Is there any probability that 2 users may have the same authKey ?  or is it safer to do something like this instead :
```php
$authKey = \yii\base\Security::generateRandomString() . '_' . time();
```

tom--:
@tunecino `generateRandomString()` is just plain random, like rolling dice, so repetitions are possible. But your proposed solution is risky, see below.

You can make the probability of repetitions very small by making the random string long enough. For the full story, read about the [Birthday Problem](https://en.wikipedia.org/wiki/Birthday_problem). I can't be bothered with that so I ask for a 12-byte string (which has 64^12 ~ 4.7E21 different values) and assume the chance of repetitions is so small I can ignore it.

If you append `'_' . time()` then you are still not safe. One second is a long time on modern processors. And the algorithm is so obvious that it makes it easy to guess things. If you want to do that you might want to run the result through something like sha256 to hide the algorithm, although that reintroduces the very small chance of repetitions, although the number of different outputs of sha256 is much bigger than the Shannon number.

samdark:
@tom-- so it this issue still here for 2.0.7?

tunecino:
@tom-- I think you are right. tagging each generated token with its manufacturer date is just making it easier to reorder & compare a couple of them. at least that. and I'm not sure if running that result thought sha256 will not just look like throwing dices one more time and just getting a different random number.

I'm trying to apply OAuth 2.0 with a long living refresh_token (maybe 30 days) + a short living access_token (maybe 1 hour) so it won't be a good idea to check DB for existence each time I generate an access_token (maybe if using cache to hold them instead of DB. I'm not sure about). 

[...]

PR #9882, opened 2015-10-10
9878 use php random bytes()

tom--:


samdark:
#9878

framework/base/Security.php
```diff
@@ -455,6 +459,15 @@ public function generateRandomKey($length = 32)
 
         $bytes = '';
 
+        if (function_exists('random_bytes')) {
+            try {
+                $bytes .= random_bytes($length);
```
cebe: on success, we should return the bytes, otherwise you will run into `/dev/urandom` in both cases.

cebe: Also I find the existing implementation a bit confusing that we append bytes to empty string instead of initializing it inside the ifs.

tom--: @cebe 

1st point - agreed.

2nd point. It's probably historical. In the [original version](http://www.yiiframework.com/extension/randomness/) of this method, lots of sources were tried. Nesting everything was a mess, iirc, and appending was tidier. It was also trying to accommodate the fact that some sources could return a value that wasn't as long as requested.

Your comment makes me think I should reconsider that structure. But first it would be good to get some more opinion on #9879, in particular, is `if (function_exists('mcrypt_create_iv')) {...}` in our out?

cebe: > It was also trying to accommodate the fact that some sources could return a value that wasn't as long as requested.

good point, have not thought about that, its probably fine.

samdark:
@tom-- what's the status of this pull request?

tom--:
Good question. I want to revise the 2.0.6 code but different from this PR.

samdark:
Closing it then.

Issue #9047, opened 2015-07-08
innerJoin / innerJoinWith in AR relation


joe-meyer:
I have a use case where I want to filter my relation based on that relation's related records such that the order of hierachy is `Dispatch` => `OrdMthdResList` => `Resource`

```
public function getOrdMthdResList()
    {
        return $this->hasMany(
            OrdMthdResList::className(),
            [
                'ord_id' => 'ord_id',
                'opr_seq_no' => 'opr_seq_no'
            ]
        )

            ->andOnCondition('is_actv = 1')
            ->andOnCondition(['>', 'ord_mthd_res_list.user_dec2', 0]);
    }

Results in the following SQL:
SELECT * 
FROM ord_mthd_res_list 
WHERE ((("ord_id" = 'S05691628' AND "opr_seq_no" = 100) OR ("ord_id" = 'S05542558' AND "opr_seq_no" = 100) OR ("ord_id" = 'S05542576' AND "opr_seq_no" = 100) OR ("ord_id" = 'S03797047' AND "opr_seq_no" = 100) OR ("ord_id" = 'S05773189' AND "opr_seq_no" = 100) OR ("ord_id" = 'S05773190' AND "opr_seq_no" = 100) OR ("ord_id" = 'S05773191' AND "opr_seq_no" = 100) OR ("ord_id" = 'S05773192' AND "opr_seq_no" = 100) OR ("ord_id" = 'S05773193' AND "opr_seq_no" = 100) OR ("ord_id" = 'S05773194' AND "opr_seq_no" = 100))) 
AND ((is_actv = 1)
AND (ord_mthd_res_list."user_dec2" > 0))
```

However, I actually need to filter these results down based on the related resource table like so, the problem is that the `Dispatch` attributes for my `ord_id` and `opr_seq_no` are all coming across as null as soon as I try to add my `innerJoinWith()`

```
    public function getOrdMthdResList()
    {
        return $this->hasMany(
            OrdMthdResList::className(),
            [
                'ord_id' => 'ord_id',
                'opr_seq_no' => 'opr_seq_no'
            ]
        )
            ->andOnCondition('is_actv = 1')
            ->andOnCondition(['>', 'ord_mthd_res_list.user_dec2', 0])
            ->innerJoinWith('resource')
            ->andOnCondition(
                'res.res_type = :materialType',
                [':materialType' => 'MATERIAL']
            );
    }
 
 
 
//resulting sql
SELECT ord_mthd_res_list.*
[...]

PR #9895, opened 2015-10-12
fixes composite key multi join

joe-meyer:
Fixes issue #9047

### DIFF
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -466,7 +466,7 @@ trait ActiveRelationTrait
             foreach ($models as $model) {
                 $v = [];
                 foreach ($this->link as $attribute => $link) {
-                    $v[$attribute] = $model[$link];
+                    $v[($this->prefixKeyColumns([$attribute])[0])] = $model[$link];
                 }
                 $values[] = $v;
             }


samdark:
Would be cool to get a unit test. It's not trivial thing so it's easy to break.

joe-meyer:
If you guys don't take care of this I'll try and look at building a unit test for it. Really something to make sure of is that we test both single condition relations as well as composite relations. I ihaven't done anything to attempt to correct relations with single conditions in the joins but I fear they may also be subject to this same bug.

samdark:
Yeah, that would be great.

nineinchnick:
@joe-meyer what db are you using? I've set up some composite fks for unit tests for pgsql.

joe-meyer:
@nineinchnick I'm using a custom odbc database driver for a [Progress](http://www.progress.com) database, but it's 99% the same as sqllite. 

If you take a look back at the original issue I feel like I outlined when the problem occurs fairly well (a `hasMany()` with an `innerJoinWith` as a part of the relation)

nineinchnick:
I understand the problem, I just wanted to point out it should be easy to add an unit test for that because there are fixtures ready.

PowerGamer1:
This fixes the problem I am currently having (need to perform order by the field in inner joined table inside active relation), can this be merged into `master` branch ASAP please?

cebe:
@PowerGamer1 if you provide a unit test I can merge it immediately. otherwise I would have to write one myself ;)

PowerGamer1:
> if you provide a unit test I can merge it immediately. otherwise I would have to write one myself ;)

I don't know how you do the unit testing but if you need an example here is a simple one. Lets say there are tables in the database: `user (id, name, ...)`, `project (id, ...)` and `project_participant (project_id, user_id, ...)`. One project can have many participants.

```PHP
class Project extends \yii\db\ActiveRecord
{
  public function getParticipants()
  {
    return $this->hasMany(ProjectParticipant::className(), ['user_id' => 'id'])->innerJoinWith('user')->orderBy(['user.name' => SORT_ASC])->indexBy('user_id');
  }
}

$project->findOne(123);
var_dump($project->participants);
```

Relation `Proejct::getParticipants()` returns no records without the fix in this pull request.

cebe:
> $this->hasMany(ProjectParticipant::className(), ['user_id' => 'id'])

This does not make sense to me. Why is a ProjectParticipant joined with Project on the user_id?

cebe:
Also I am not really able to reproduce the issue so it would be great if someone could create a test case or at least some info on how to reproduce it with AR classes defined in unit test directory of the framework. https://github.com/yiisoft/yii2/issues/9047#issuecomment-213431506

PowerGamer1:
> $this->hasMany(ProjectParticipant::className(), ['user_id' => 'id'])
This does not make sense to me. Why is a ProjectParticipant joined with Project on the user_id?

Just a typo, should be `['project_id' => 'id']`.

cebe:
@PowerGamer1 the case with no composite key involved does not fail for me.

PR #9905, opened 2015-10-13
StringHelper::pregExplode implemented

SilverFire:
Propose to add pregExplode method to the `StringHelper`.
The difference with `StringHelper::explode` is that this one is for splitting using regular expression.

### DIFF
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -268,4 +268,37 @@ class BaseStringHelper
         }
         return $result;
     }
+
+    /**
+     * Explodes string into array using regular expression, optionally trims values and skips empty ones
+     *
+     * @param string $string String to be exploded.
+     * @param string $pattern Delimiter regular expression. Default is `/[\s,;]+/`.
+     * @param mixed $trim Whether to trim each element. Can be:
+     *   - boolean - to trim normally;
+     *   - string - custom characters to trim. Will be passed as a second argument to `trim()` function.
+     *   - callable - will be called for each value instead of trim. Takes the only argument - value.
+     * @param boolean $skipEmpty Whether to skip empty strings between delimiters. Default is false.
+     * @return array
+     * @since 2.0.7
+     */
+    public static function pregExplode($string, $pattern = '/[\s,;]+/', $trim = true, $skipEmpty = false)
+    {
+        $result = preg_split($pattern, $string);
+        if ($trim) {
+            if ($trim === true) {
+                $trim = 'trim';
+            } elseif (!is_callable($trim)) {
+                $trim = function ($v) use ($trim) {
+                    return trim($v, $trim);
+                };
+            }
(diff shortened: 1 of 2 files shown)


framework/helpers/BaseStringHelper.php
```diff
@@ -268,4 +268,37 @@ public static function explode($string, $delimiter = ',', $trim = true, $skipEmp
...
+            } elseif (!is_callable($trim)) {
+                $trim = function ($v) use ($trim) {
+                    return trim($v, $trim);
+                };
+            }
+            $result = array_map($trim, $result);
+        }
+        if ($skipEmpty) {
+            // Wrapped with array_values to make array keys sequential after empty values removing
+            $result = array_values(array_filter($result));
```
kidol: array_filter will also remove string `0`, I doubt that's expected? https://3v4l.org/dXpPr

SilverFire: Aw, snap! It really does! Thank you for your attention!

SilverFire: Fixed. Thank you very much

framework/helpers/BaseStringHelper.php
```diff
@@ -270,4 +270,39 @@ public static function explode($string, $delimiter = ',', $trim = true, $skipEmp
...
+     * @return array
+     * @since 2.0.7
+     */
+    public static function pregExplode($string, $pattern = '/[\s,;]+/', $trim = true, $skipEmpty = false)
+    {
+        $result = preg_split($pattern, $string);
+        if ($trim) {
+            if ($trim === true) {
+                $trim = 'trim';
+            } elseif (!is_callable($trim)) {
```
samdark: It's not necessary. There could be no situation where `trim()` doesn't exist.

framework/helpers/BaseStringHelper.php
```diff
@@ -270,4 +270,39 @@ public static function explode($string, $delimiter = ',', $trim = true, $skipEmp
...
+                $trim = function ($v) use ($trim) {
+                    return trim($v, $trim);
+                };
+            }
+            $result = array_map($trim, $result);
```
samdark: Condition at line 290 could be moved here.

framework/helpers/BaseStringHelper.php
```diff
@@ -270,4 +270,39 @@ public static function explode($string, $delimiter = ',', $trim = true, $skipEmp
...
+        }
+        if ($skipEmpty) {
+            // Wrapped with array_values to make array keys sequential after empty values removing
+            $result = array_values(array_filter($result, function ($value) {
```
samdark: There's `PREG_SPLIT_NO_EMPTY` for that.

samdark:
After all adjustments isn't becoming just `preg_split` with defaults?

SilverFire:
You're right.

PR #9912, opened 2015-10-13
Fixed #9911 - StringHelper::explode with skip_empty option removed item eq to 0

### LINKED ISSUE #9911: StringHelper::explode with skip_empty option removes item with value eq 0
As @kidol mentioned in #9905, `StringHelper::explode` with skip_empty option really removes items with value equal to 0.
Going to create a PR.

SilverFire:
Closes #9911


samdark:
Thanks!

PR #9915, opened 2015-10-14
Update BaseArrayHelper.php

totaldev:
Terrible behavior. What happen when we have \ArrayObject without some properties?

### DIFF
--- a/framework/helpers/BaseArrayHelper.php
+++ b/framework/helpers/BaseArrayHelper.php
@@ -198,7 +198,7 @@ class BaseArrayHelper
             $key = substr($key, $pos + 1);
         }
 
-        if (is_object($array)) {
+        if (is_object($array) && isset($array->$key)) {
             return $array->$key;
         } elseif (is_array($array)) {
             return array_key_exists($key, $array) ? $array[$key] : $default;


samdark:
Thanks!

alchimik:
@samdark https://github.com/yiisoft/yii2/issues/7193#issuecomment-73287156 ?

samdark:
@alchimik I don't like error in this case. Much more logical is to return `null`.

azureru:
Hi @samdark. 

But if the property is exists and the value is `null` it will return `default` value instead of `null`

isn't it better to use `property_exists` in this case?

samdark:
Nope. Why?

samdark:
Yes. Probably. Can you send a pull request fixing it?

cebe:
> @alchimik I don't like error in this case. Much more logical is to return null.

isn't this a BC break?

samdark:
It probably is.

samdark:
Yes. Current behavior could be used. @azureru let's leave it as is.

cebe:
@samdark I mean the original PR is a BC break.

cebe:
Sorry, guess I quoted incorrectly.

samdark:
@cebe umm, I don't think anyone used erroring there...

samdark:
Ah, and since it's for 2.0.7 then there's no BC break yet.

samdark:
https://github.com/yiisoft/yii2/commit/4b511fb9bcbb1fc5838ad8bc65f3852b09287acb

SilverFire:
@samdark commit reverted https://github.com/yiisoft/yii2/issues/10588
f72e9d5

cebe:
That is only one problem that could arise from this. adding this may break code that overrides __get() but not __isset(). Imo the whole check should be reverted.

cebe:
I have opended a PR to revert this, lets discuss the topic there: #10590

PR #9919, opened 2015-10-14
Added "IF EXISTS" option for "DROP TABLE"

edgardmessias:


samdark:
Can't be merged.

1. It breaks backwards compatibiltiy.
2. `IF EXISTS` isn't part of SQL standard and has different syntax in some DBs we support. The patch will work for MySQL and probably PosgreSQL but not MSSQL or Oracle.

edgardmessias:
I tested it with MySQL, PosgreSQL, Cubrid and Sqlite.
I wrote to MSSQL and OCI, but without testing. See [here](https://github.com/edgardmessias/yii2/commit/2c87a971134c8b98e6b7070b1872c3c624fcd72a#diff-de218df12e89df5f7e70b466c16dc137R132)

samdark:
OK. That's something to consider for 2.1 then. Could you re-submit a pull request?

edgardmessias:
Ok, I wait by 2.1

PR #9926, opened 2015-10-15
Fixed #9924 - yii.js handleAction corrupted parameters with quote char

### LINKED ISSUE #9924: yii.js: yii.handleAction is not possible to use parameter value with quote "
```php
echo Html::a('submit', ['site/foobar'], [
    'data' => [
        'method' => 'post',
        'params' => [
            'name1' => 'val"ue1',
        ],
    ],
]);
```

On click will send parameter name1 with value 'val' instead of 'val"ue1'

It can be fixed in input generation at
https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.js#L194

SilverFire:
Closes #9924


samdark:
Thanks!

PR #9932, opened 2015-10-15
review translater, correct terms, sintax e outhers

lukBarros:
review translater, correct terms, sintax e outhers for ptBR.


samdark:
Thanks!

PR #9951, opened 2015-10-17
Update tip 'style required fields with asterisks'

JohnyDNU:
Update tip 'style required fields with asterisks' with solution to remove the red start after each radiolist item


samdark:
Thanks!

PR #9954, opened 2015-10-18
`TimestampBehavior::getValue()` value processing is consistent with `AttributeBehavior`

### LINKED ISSUE #9953: TimestampBehavior::value is inconsistent with other behaviors
`TimestampBehavior::getValue()` overrides `AttributeBehavior::getValue()`.
`AttributeBehavior::getValue()` accepts one special type in `$this->value`: if it's not a Closure, the value will be used directly.
`TimestampBehavior::getValue()` accepts one special type in `$this->value`: if it's not an Expression, the value will be used as a callback, unless it's null.

So if one wants to set the timestamp to a fixed value like `$_SERVER['REQUEST_TIME']`, they have to use a closure.
This behavior is not intuitive, leads to unexpressive errors, and is inconsistent with its parent. TimestampBehavior, like its parent, should detect Closure instances and accept constant values.

SilverFire:
Closes #9953


samdark:
Thanks!

SilverFire:
The PR is BC-breaking according to the [report](https://github.com/yiisoft/yii2-framework/commit/ebc8f6aa6ccd1978e5e6483c90f845998f5d7ddd#commitcomment-16013367)

PR #9956, opened 2015-10-18
Add basePath and baseUrl to compressed assets bundle

NikolayMurha:
If you use construction like this:

```
$bundle = AppAsset::register($this);
<img src="<?= $bundle->baseUrl ?>/images/some-image.gif">" title="My cool image">
```

After compression it will be broken, because baseUrl will be empty.
This pull request fix it.

### DIFF
--- a/framework/console/controllers/AssetController.php
+++ b/framework/console/controllers/AssetController.php
@@ -438,6 +438,7 @@ class AssetController extends Controller
     protected function saveTargets($targets, $bundleFile)
     {
         $array = [];
+        $bundles = $this->loadBundles($this->bundles);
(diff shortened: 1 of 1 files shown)


klimov-paul:
> If you use construction like this

It is incorect usage. not every AssetBundle has `baseUrl`, compressed one included.

MurgaNikolay:
Ok, how I can get url to image url ore another asset?
So far I have found only this solution. Or I should put images to document root directly?

rugabarbo:
> After compression it will be broken, because baseUrl will be empty.

@NikolayMurha, how did you solve this problem for **compressed** assets?

Issue #9967, opened 2015-10-20
How to set a PDO statement attribute?


thejahweh:
I have to use the each method to select in batches. The problem is, that I have to delete some columns in this foreach, which generates a table lock. Now I have to set the cursor `PDO::ATTR_CURSOR` to forward only `PDO::CURSOR_FWDONLY`. Is this even possible by now?

```php
$simpleQuery = SimpleObject::find();
// need something like: $simpleQuery->setPdoStmtAttr(\PDO::ATTR_CURSOR, \PDO::CURSOR_FWDONLY);
foreach ($simpleQuery->each() as $simpleObject) {
    // fancy things...
}
```

samdark:
Yes. You can access PDO instance via connection class.

thejahweh:
Yes but the `PDO::ATTR_CURSOR` attribute must be set on the pdo statement object.

samdark:
`Command::$pdoStatement` is public.

thejahweh:
And how do I use it in my snipped?

samdark:
Same as you'd use it with PDO.

thejahweh:
So it is not possible with the each statement?

samdark:
Probably not. I know nothing about your use case.

samdark:
Currently not.

thejahweh:
Hmm I just looked into it, but I don't know where I should put the change in.

1.Loop the Attributes through the Query
```php
$query->setPdoAttributes([\PDO::ATTR_CURSOR => \PDO::CURSOR_FWDONLY])->each();
```
2.Make the method `Query::createCommand` return the same Command for every call, save it in a private/protected variable and just return it if called twice.
```php
$query->command->setPdoAttributes([\PDO::ATTR_CURSOR => \PDO::CURSOR_FWDONLY]);
$query->each();
```

samdark:
`#1` looks better.

PR #9973, opened 2015-10-20
9967 set pdo statement attributes

aliechti:
#9967


samdark:
Assigned to 2.1 since it breaks backwards compatibility.

thejahweh:
Where is the BC? Its just a parameter more, which has a default value.

samdark:
Public method signature change. When overridden it will throw an exception if original method signature is changed.

Issue #7435, opened 2015-02-25
Behaviors and Event handling for Widget.


petrabarus:
I have a lot of fragment cache for widgets in the view level.

``` php
if ($this->beginCache($id)){
    echo app\MyWidget::widget([]);
    $this->endCache();
}
```

I'm wondering if we can have `beforeRender` and `afterRender` widget, so that the output caching can be done in the Widget level instead of View level.

Something similar like [PageCache](https://github.com/yiisoft/yii2/blob/master/framework/filters/PageCache.php) but for widget level.

I know I can always use something like this

``` php
class BookList extends Widget {
     public $authorId;
     public function run(){
        if ($this->getView()->beginCache($authorId)){
            $this->getView()->endCache();
        }
     }
}
```

But I think something like this is much more convenient

``` php
class MyWidget extends Widget {
    public function behaviors(){
         return [
            ['class' => WidgetCache::class]
         ]
    }
}
```

This way if I want to change a cache strategy, (for example appending `Yii::$app->language` before key generation), I can just replace the `WidgetCache` to something other instead of updating all lines before the `$this->getView()->beginCache()`. Or I can easily change the `cache` object (memcached, filecache, etc).

The only use case I can find is only widget-level output caching, but I'm sure this can be convenient for a lot of stuffs.

petrabarus:
Some of the use cases I can find for this.

#### 1. Widget-level Cache Behavior

As mentioned above, a lot of times we need to cache widget in the views. I think it will be great if we can do something like this.

```php
MyWidget::widget([
    'id' => 'widget1',
    'as cacheable' => WidgetCache::class,
]);
```
instead of
```
if ($this->beginCache()) {
   MyWidget::widget([
      'id' => 'widget1',
  ]);
}
```

If the cache has a lot of options that are repeatable, we can probably wrap in a class extended from `WidgetCache`.

#### 2. Obtaining WidgetId For Javascript

Let's say I have a Javascript that manipulate every links in a segment. Something like

```js
$.fn.changeUrl = function() {
    $('a', this).attr('href', 'http://www.example.com');
}
```

Then I put this in a widget.

```php
class MyUrlWidget extends \yii\base\Widget {
     public $widgetId;
     public function run() {
         $this->getView()->registerJs(<<<JS
$.fn.changeUrl = function() {
    $('a', this).attr('href', 'http://www.example.com');
}
JS
         , View::POS_END, __CLASS__);
         $this->getView()->registerJs(<<<JS
[...]

PR #9981, opened 2015-10-21
[WIP] Support behavior and event handling for Widget (#7435)

petrabarus:
Idea for supporting behavior and event handling for `\yii\base\Widget`. More description on #7435.

### DIFF
--- a/framework/base/Widget.php
+++ b/framework/base/Widget.php
@@ -25,6 +25,16 @@ use ReflectionClass;
 class Widget extends Component implements ViewContextInterface
 {
     /**
+     * @event WidgetEvent an event raised right before executing a widget.
+     * You may set [[WidgetEvent::isValid]] to be false to cancel the widget execution.
+     */
+    const EVENT_BEFORE_RUN = 'beforeRun';
+    /**
+     * @event WidgetEvent an event raised right after executing a widget.
+     */
+    const EVENT_AFTER_RUN = 'afterRun';
+    
+    /**
      * @var integer a counter used to generate [[id]] for widgets.
      * @internal
      */
@@ -70,7 +80,12 @@ class Widget extends Component implements ViewContextInterface
         if (!empty(static::$stack)) {
             $widget = array_pop(static::$stack);
             if (get_class($widget) === get_called_class()) {
-                echo $widget->run();
+                /* @var $widget Widget */
+                if ($widget->beforeRun()) {
+                    $result = $widget->run();
+                    $result = $widget->afterRun($result);
+                    echo $result;
+                }
                 return $widget;
             } else {
                 throw new InvalidCallException('Expecting end() of ' . get_class($widget) . ', found ' . get_called_class());
@@ -95,7 +110,11 @@ class Widget extends Component implements ViewContextInterface
             /* @var $widget Widget */
             $config['class'] = get_called_class();
             $widget = Yii::createObject($config);
-            $out = $widget->run();
+            $out = '';
+            if ($widget->beforeRun()) {
+                $result = $widget->run();
+                $out = $widget->afterRun($result);
+            }
         } catch (\Exception $e) {
             // close the output buffer opened above if it has not been closed already
             if (ob_get_level() > 0) {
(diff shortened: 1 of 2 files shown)


petrabarus:
@samdark hi alex, this is what i told you about on the messenger. what do you think?

samdark:
I'm worrying about performance a bit. That's two events per widget and there coud be many widgets at the same page.

samdark:
The code itself looks OK though.

petrabarus:
Yeah, I totally agree with that. Haven't tested how big the overhead will be.

My main goal with the event is so that I can leverage the trade-off by using something like `WidgetCache` below. That way, widgets like `GridView` and `ListView` that loads heavy data can be cached.

```php
<?= 
yii\widgets\ListView::widget([
    'id' => 'my-list',
    'as cached' => WidgetCache::className(),
]);
?>
```

```php
<?php
/**
 * @author Petra Barus <petra.barus@gmail.com>
 */

use yii\caching\Cache;
use yii\di\Instance;

/**
 * @author Petra Barus <petra.barus@gmail.com>
 * @property-read \yii\base\Widget $owner
 */
class WidgetCache extends \yii\base\Behavior {
    
    /**
     * @var Cache|array|string the cache object or the application component ID of the cache object.
     */
    public $cache = 'cache';
    
    /**
     * @var integer number of seconds that the data can remain valid in cache.
     * Use 0 to indicate that the cached data will never expire.
     */
    public $duration = 60;
    
    /**
     * @var array|Dependency the dependency that the cached content depends on.
     * This can be either a [[Dependency]] object or a configuration array for creating the dependency object.
     * For example,
     *
     * ```php
     * [
     *     'class' => 'yii\caching\DbDependency',
     *     'sql' => 'SELECT MAX(updated_at) FROM post',
     * ]
     * ```
     *
     * would make the output cache depend on the last modified time of all posts.
     * If any post has its modification time changed, the cached content would be invalidated.
     *
     * If [[cacheCookies]] or [[cacheHeaders]] is enabled, then [[\yii\caching\Dependency::reusable]] should be enabled as well to save performance.
     * This is because the cookies and headers are currently stored separately from the actual page content, causing the dependency to be evaluated twice.
     */
    public $dependency;
    
    /**
     * @var array list of factors that would cause the variation of the content being cached.
     * Each factor is a string representing a variation (e.g. the language, a GET parameter).
     * The following variation setting will cause the content to be cached in different versions
     * according to the current application language:
     *
     * ~~~
     * [
     *     Yii::$app->language,
     * ]
     * ~~~
     */
    public $variations;
    
    /**
     * @inheritdoc
     */
    public function init() {
        parent::init();
        $this->cache = Instance::ensure($this->cache, Cache::className());
    }
    
    /**
     * @inheritdoc
     */
    public function attach($owner) {
        $this->owner = $owner;
        $this->owner->on(WidgetEventInterface::EVENT_BEFORE_RUN, [$this, 'beforeRun']);
    }
    
    /**
     * Handle cache before run.
     * @param WidgetEvent $event the event.
     */
    public function beforeRun(WidgetEvent $event) {
        foreach (['cache', 'duration', 'dependency', 'variations'] as $name) {
            $properties[$name] = $this->{$name};
        }
        
        if ($this->owner->getView()->beginCache($this->owner->getId(), $properties)) {
            $this->owner->on(WidgetEventInterface::EVENT_AFTER_RUN, [$this, 'afterRun']);
            $event->isValid = true;
        } else {
            $event->isValid = false;
        }
    }
    
    /**
     * Handle cache after run.
     * @param WidgetEvent $event the event.
     */
    public function afterRun(WidgetEvent $event) {
        echo $event->result;
        $event->result = '';
        $this->owner->getView()->endCache();
    }
}

```

petrabarus:
Alternatively, when people want to use the event, for example, to do the caching. They can always extend the `\yii\base\Widget` and put the `WidgetEventTrait` like below.

```

use Yii;
/**
 * WidgetEvent represent event by Widget.
 * @author Petra Barus <petra.barus@gmail.com>
 * @since 2015.04.21
 */
trait WidgetEventTrait
{
    
    /**
     * This method is invoked right before an widget is executed.
     *
     * The method will trigger the [[EVENT_BEFORE_RUN]] event. The return value of the method
     * will determine whether the widget should continue to run.
     *
     * @return boolean Description
     */
    public function beforeRun() {
        $event = new WidgetEvent($this);
        $this->trigger(WidgetEventInterface::EVENT_BEFORE_RUN, $event);
        return $event->isValid;
    }
    /**
     * This method is invoked right after an widget is executed.
     *
     * The method will trigger the [[EVENT_AFTER_ACTION]] event. The return value of the method
     * will be used as the widget return value.
     * ```
     *
     * @param mixed $result the widget return result.
     * @return mixed the processed widget result.
     */
    public function afterRun($result) {
        $event = new WidgetEvent($this);
        $event->result = $result;
        $this->trigger(WidgetEventInterface::EVENT_AFTER_RUN, $event);
        return $event->result;
    }
    /**
     * Begins a widget.
     * This method creates an instance of the calling class. It will apply the configuration
     * to the created instance. A matching [[end()]] call should be called later.
     * @param array $config name-value pairs that will be used to initialize the object properties
     * @return static the newly created widget instance
     */
    public static function begin($config = [])
    {
        $config['class'] = get_called_class();
        /* @var $widget Widget */
        $widget = Yii::createObject($config);
        static::$stack[] = $widget;
        return $widget;
    }
    /**
     * Ends a widget.
     * Note that the rendering result of the widget is directly echoed out.
     * @return static the widget instance that is ended.
     * @throws InvalidCallException if [[begin()]] and [[end()]] calls are not properly nested
     */
    public static function end()
    {
        if (!empty(static::$stack)) {
            $widget = array_pop(static::$stack);
            /* @var $widget \yii\base\Widget */
            if (get_class($widget) === get_called_class()) {
                if ($widget->beforeRun()) {
                    $result = $widget->run();
                    $result = $widget->afterRun($result);
                    echo $result;
                }
                return $widget;
            } else {
                throw new InvalidCallException('Expecting end() of ' . get_class($widget) . ', found ' . get_called_class());
            }
        } else {
            throw new InvalidCallException('Unexpected ' . get_called_class() . '::end() call. A matching begin() is not found.');
        }
    }
    /**
     * Creates a widget instance and runs it.
     * The widget rendering result is returned by this method.
     * @param array $config name-value pairs that will be used to initialize the object properties
     * @return string the rendering result of the widget.
     * @throws \Exception
     */
    public static function widget($config = [])
    {
        ob_start();
        ob_implicit_flush(false);
        try {
            /* @var $widget Widget */
            $config['class'] = get_called_class();
            $widget = Yii::createObject($config);
            /* @var $widget \yii\base\Widget */
            $out = '';
            if ($widget->beforeRun()) {
                $result = $widget->run();
                $out = $widget->afterRun($result);       
            }
        } catch (\Exception $e) {
            // close the output buffer opened above if it has not been closed already
            if (ob_get_level() > 0) {
                ob_end_clean();
            }
            throw $e;
        }
        return ob_get_clean() . $out;
    }
}
```

samdark:
Would you perform some performance tests?

petrabarus:
I am gonna try.

--
Petra Novandi Barus, MT
http://www.linkedin.com/in/petrabarus

Sent from Outlook




On Fri, Oct 23, 2015 at 7:25 AM -0700, "Alexander Makarov" <notifications@github.com> wrote:












Would you perform some performance tests?

samdark:
Closed in favor of https://github.com/yiisoft/yii2/pull/13457 which includes some polishing...

PR #9983, opened 2015-10-21
Do not render client validation when `Validator::with` is set and `withClient` is not

SilverFire:
There is a common problem with model attribute validation: when developer defines `when` condition in the model rule, but forgots about `whenClient`.
This leads to inconsistend client-side validation.

This PR fixes behaviour of `ActiveFieleld::getClientOptions()` so JS will not be rendered, when `when` is defined, and `withClient` is not.


SilverFire:
Set `'enableClientValidation' => false` will do the same without this PR.

Considering this PR will change important default behaviour, i'm not sure that it can be merged.
So when there is something this in the code:
```php
['price', 'required', 'when' => function ($model) {
    return Yii::$app->user->can('setPrice');
}
```
The form will be busted.

samdark:
Then I don't think it's a good idea to change it.

PR #9996, opened 2015-10-23
Update rest-quick-start.md

free6k:
The method of the guide does not work

### DIFF
--- a/docs/guide/rest-quick-start.md
+++ b/docs/guide/rest-quick-start.md
@@ -52,7 +52,7 @@ Then, modify the configuration about the `urlManager` component in your applicat
     'enableStrictParsing' => true,
     'showScriptName' => false,
     'rules' => [
-        ['class' => 'yii\rest\UrlRule', 'controller' => 'user'],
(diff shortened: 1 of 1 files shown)


cebe:
we have a test for this, what exactly is not working?
https://github.com/yiisoft/yii2/blob/master/tests/framework/rest/UrlRuleTest.php#L71

cebe:
note that singular names will be pluralized automatically when not turned off.

free6k:
so why do I have problems with the process of documentation? this solution I found in Google.

cebe:
which problem are you facing exactly?

free6k:
>>>> note that singular names will be pluralized automatically when not turned off.
pluralized! Problem here. In documentation haven't information about it. Reject this pull request

cebe:
> In documentation haven't information about it.

https://github.com/free6k/yii2/blob/ff0e46524eaf2b7a0335523f7363dd5a8dc46985/docs/guide/rest-quick-start.md#trying-it-out-

it is in the info box under the list of HTTP methods.

PR #10001, opened 2015-10-24
Added IpValidator description to `core-validators` ru,en guides. IpValidator PHPDoc enhanced [ci skip]

SilverFire:


samdark:
Thanks!

PR #10015, opened 2015-10-26
mysql\Schema: Reintroduced older (SQL CREATE) style fk detection

Blizzke:
As discussed in 191185f5e301dbea93783870d3b650916c304719 I've reintroduced the previous way of determining FKs for a MySQL table. Took the liberty of implementing a patch for bug #6351 as well in the old style.

### DIFF
--- a/framework/db/mysql/Schema.php
+++ b/framework/db/mysql/Schema.php
@@ -277,25 +277,24 @@ SQL;
             }
         } catch (\Exception $e) {
             $previous = $e->getPrevious();
-            if ($previous instanceof \PDOException && strpos($previous->getMessage(), 'SQLSTATE[42S02') !== false) {
-                // table does not exist, try to determine the foreign keys using the table creation sql
-                $sql = $this->getCreateTableSql($table);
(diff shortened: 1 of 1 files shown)


framework/db/mysql/Schema.php
```diff
@@ -261,18 +261,42 @@ protected function findConstraints($table)
...
+            $table->foreignKeys = [];
+            foreach ($constraints as $constraint) {
+                $table->foreignKeys[] = array_merge(
+                    [$constraint['referenced_table_name']],
+                    $constraint['columns']
+                );
+            }
+        } catch (\Exception $e) {
+            $previous = $e->getPrevious();
+            if ($previous instanceof \PDOException && strpos($previous->getMessage(), 'SQLSTATE[42S02') !== false) {
```
nineinchnick: Maybe the condition can be reversed? This saves one level of indentation:
~~~php
if (!$previous instanceof \PDOException || strpos($previous->getMessage(), 'SQLSTATE[42S02') === false) {
    throw $e;
}
~~~

samdark:
Thanks! https://github.com/yiisoft/yii2/commit/d40435e9311427d8f012c682802931702d506e7a

Issue #8649, opened 2015-06-04
final report about applied migrations


dizews:
Hello.

I think it will be a good if we add how many migrations was applied in a last time

mdmunir:
```
yii migrate/history
```

dizews:
@mdmunir I talk about migrations applied in a last time. It will be useful for release engineer to rollback last release

samdark:
You mean getting it with a separate command or the output of the command just applied them?

dizews:
@samdark I mean the output of the command.
for example: ```2 from 5 migrations was applied```

samdark:
What do you mean by `from 5`?

dizews:
In the begin the script show us how many migrations will be applied. At the end, I expect to see how many migrations was applied from the first one.

samdark:
OK.

PR #10017, opened 2015-10-26
Added total applied migrations to final report

nikashitsa:
This Pull Request solves #8649.
Looking forward for any feedback :blush:

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -124,14 +124,19 @@ abstract class BaseMigrateController extends Controller
         }
         $this->stdout("\n");
 
+        $applied = 0;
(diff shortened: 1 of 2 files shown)


framework/console/controllers/BaseMigrateController.php
```diff
@@ -124,14 +124,19 @@ public function actionUp($limit = 0)
         }
         $this->stdout("\n");
 
+        $applied = 0;
         if ($this->confirm('Apply the above ' . ($n === 1 ? 'migration' : 'migrations') . '?')) {
             foreach ($migrations as $migration) {
                 if (!$this->migrateUp($migration)) {
+                    $this->stdout("\n$applied from $n " . ($applied === 1 ? 'migration was' : 'migrations were') ." applied.\n", Console::FG_RED);
```
samdark: `migration` is still common part so only `was` and `were` could be in condition.

samdark:
Thanks!

Issue #9803, opened 2015-09-30
Option to register asset with scheme


leandrogehlen:
it's necessary create option to register asset with `scheme`.

Currently `AssetBundle` generate resources with relative path like: 
```html
<link href="/yii2-myapp/web/assets/ca59520f/toolbar.css" rel="stylesheet"></head>
```
in some cases it's necessary that will be generate:
```html
<link href="http://localhost/yii2-myapp/web/assets/ca59520f/toolbar.css" rel="stylesheet"></head>
```
i need this feature, when i was using [wkhtmltopdf](http://wkhtmltopdf.org/) to generate pdf from html

samdark:
Yeah. Could be added as an option.

nikashitsa:
@samdark @leandrogehlen I've implemented it. Please have a look at Pull Request #10031.

klimov-paul:
The most obvious way to achieve the described effect will be usage of an special alias at `AssetBundle::baseUrl` like `@absoluteweb`, which will be resolved as absolute URL with schema part.
Other option will be override existing `@web` alias with another value.
If you may also change `AssetManager::baseUrl` to the absolute URL to have same effect for the published assets.

PR #10031, opened 2015-10-27
Added option to register asset with scheme

nikashitsa:
This Pull Request solves #9803.
Covered by AssetBundleTest `tests\framework\web\AssetBundleTest::testRegister`.
Looking forward for feedback  :eyes:

### DIFF
--- a/framework/web/AssetBundle.php
+++ b/framework/web/AssetBundle.php
@@ -59,6 +59,16 @@ class AssetBundle extends Object
      */
     public $baseUrl;
     /**
+     * @var boolean|string the URI scheme to use. If null, it uses [[AssetManager::defaultScheme]].
+     * Possible values:
+     *
+     * - `false`: assets URL without host info.
+     * - `true`: assets URL with an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::hostInfo]].
+     * - string: assets URL with an absolute base URL with the specified scheme (either `http` or `https`).
+     * @since 2.0.7
+     */
+    public $scheme;
+    /**
      * @var array list of bundle class names that this bundle depends on.
      *
      * For example:
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -72,6 +72,15 @@ class AssetManager extends Component
      */
     public $baseUrl = '@web/assets';
     /**
+     * @var boolean|string the URI scheme to use. Possible values:
+     *
+     * - `false` (default): assets URL without host info.
+     * - `true`: assets URL with an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::hostInfo]].
+     * - string: assets URL with an absolute base URL with the specified scheme (either `http` or `https`).
+     * @since 2.0.7
+     */
+    public $defaultScheme = false;
+    /**
      * @var array mapping from source asset files (keys) to target asset files (values).
      *
      * This property is provided to support fixing incorrect asset file paths in some asset bundles.
@@ -290,6 +299,7 @@ class AssetManager extends Component
     /**
      * Returns the actual URL for the specified asset.
      * The actual URL is obtained by prepending either [[baseUrl]] or [[AssetManager::baseUrl]] to the given asset path.
+     * [[scheme]] or [[AssetManager::defaultScheme]] is used, if specified.
      * @param AssetBundle $bundle the asset bundle which the asset file belongs to
      * @param string $asset the asset path. This should be one of the assets listed in [[js]] or [[css]].
      * @return string the actual URL for the specified asset.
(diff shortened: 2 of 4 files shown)


framework/web/AssetBundle.php
```diff
@@ -59,6 +59,14 @@ class AssetBundle extends Object
      */
     public $baseUrl;
     /**
+     * @var boolean|string the URI scheme to use. Possible values:
+     *
+     * - `false` (default): assets URL without host info.
+     * - `true`: assets URL with an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::hostInfo]].
+     * - string: assets URL with an absolute base URL with the specified scheme (either `http` or `https`).
+     */
+    public $scheme = false;
```
samdark: Need `@since 2.0.7` tag.

framework/web/AssetManager.php
```diff
@@ -72,6 +72,14 @@ class AssetManager extends Component
      */
     public $baseUrl = '@web/assets';
     /**
+     * @var boolean|string the URI scheme to use. Possible values:
+     *
+     * - `false` (default): assets URL without host info.
+     * - `true`: assets URL with an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::hostInfo]].
+     * - string: assets URL with an absolute base URL with the specified scheme (either `http` or `https`).
+     */
+    public $scheme = false;
```
samdark: Need `@since 2.0.7` tag.

framework/web/AssetBundle.php
samdark: Also it seems description is wrong i.e. `false` means "use scheme as specified in a asset manager".

framework/web/AssetManager.php
samdark: Could be `defaultScheme` to reflect the purpose better.

framework/web/AssetBundle.php
nikashitsa: Thank you for quick feedback!

Tag has been added.
> Also it seems description is wrong i.e. false means "use scheme as specified in a asset manager".

Nope, it will not use scheme value from asset manager as default scheme. I've updated description in AssetManager: scheme in AssetManager is for assets from assetMap. Only if `resolveAsset()` found $actualAsset, AssetManager scheme will be used.

framework/web/AssetManager.php
nikashitsa: Tag has been added.
> Could be defaultScheme to reflect the purpose better.

It's not a defaultScheme.
> it will not use scheme value from asset manager as default scheme. I've updated description in AssetManager: scheme in AssetManager is for assets from assetMap. Only if resolveAsset() found $actualAsset, AssetManager scheme will be used.

nikashitsa:
@samdark Thank you for quick feedback! I've added tags and changed scheme description in AssetManager.

framework/web/AssetManager.php
```diff
@@ -290,6 +299,7 @@ protected function loadDummyBundle($name)
     /**
      * Returns the actual URL for the specified asset.
      * The actual URL is obtained by prepending either [[baseUrl]] or [[AssetManager::baseUrl]] to the given asset path.
+     * [[scheme]] or [[AssetManager::scheme]] is used, if specified.
```
samdark: What if both specified?

nikashitsa: If both specified:
1. and asset is missing from AssetManager::assetMap, then bundle scheme will be used.
2. and asset matched in assetMap, then AssetManager::scheme will be used.
So, AssetManager::scheme overwrites AssetBundle::schema, if asset exists in AssetManager::assetMap.

Perhaps, to keep it simple it's better to remove AssetsManager::scheme and leave only AssetBundle::scheme.
What do you think? It will be enough to solve issue.

samdark: Shouldn't it be the other way around?

nikashitsa: Yes, currently it's not obvious.
I think, I've got your idea: if AssetBundle::scheme is not set, use AssetManager::defaultScheme.
It looks good for me too. Should I do this way?

samdark: Yes.

nikashitsa: I agree with you. I've fixed it :blush:

samdark:
Merged https://github.com/yiisoft/yii2/commit/57dc8540fd0842a76196b92e6c86e1b3441d9c46. Thanks!

klimov-paul:
I don't agree with the fix. There is already an option for the asset base URL, so you are able to setup it with absolute URL if needed. Having `schema` parameter seems to be redundant.

samdark:
@klimov-paul how would you configure a bundle to be registered via absolute URL if it's used via endless number of sub-domains such as `username.example.com` using base URL?

klimov-paul:
First of all, this could be solved at `AssetManager` level.
Second there is no big deal with it: all you need as an extra path alias. While `@web` contains realtive base URL, you can easily register `@absweb` (or something), which can hold absolute path for the current domain. Most probably, you will need such alias anyway, if things came to described problem with asset bundles.

samdark:
Right.

samdark:
https://github.com/yiisoft/yii2/commit/fe4cbd30d61ececa805d5532c830d1594496d673 reverted.

Issue #9458, opened 2015-08-18
Log profiling duration to file


mikehaertl:
Currently the profiling result is not logged to file (or at least I couldn't figure out how to do that). All I get in my logfiles is something like:

```
2015-08-18 12:07:06 [-][-][-][profile begin][app\export\shop\BaseUpdate::update] Update prices
2015-08-18 12:07:06 [-][-][-][profile end][app\export\shop\BaseUpdate::update] Update prices
```

So it's missing the actual profiling information about duration.

cronfy:
What about imlementing this feature?

Indeed, currently there is no way of profiling database queries and other stuff. There is just no enough information. E. g. for database queries, if we enable profile logging, we can see number of queries and queries itself, but we can't find out how much time it was taken to execute a query.

What prevents #10045 from merging?

samdark:
@cronfy seems I've somehow missed it. Will review for next release.

cronfy:
@samdark What if use another approach here?

1. Create separate `yii\base\Profiler` class with `begin()` and `end()` methods (copy `beginProfile()` and `endProfile()` from `BaseYii`).
2. Make profiler a core component.
3. Replace `BaseYii::beginProfile()` contents with `Yii::$app->profiler->begin($token, $category)`, and same with `BaseYii::endProfile()`.

Then everyone will be able to replace profiler with his own implementation. This gives a starting point to solve all "profiler does not work as I need" issues (#8112, #7617, #9458, http://stackoverflow.com/questions/30711195/how-to-profile-a-yii2-based-api/30717856).

samdark:
@cronfy sounds like a good plan.

klimov-paul:
> Then everyone will be able to replace profiler with his own implementation.

It is already possible to replace standard `Logger` with your own implementation. What the sense of introduction of another abstraction layer here?

klimov-paul:
Resolved via #14611

PR #10045, opened 2015-10-28
Added profiling duration to log formatted message

nikashitsa:
This Pull Request solves #9458.
P.S. I hope my Japanese translation is correct :)


cebe:
it should not be logged by default. in most cases it is not relevant information.

nikashitsa:
By default it will produce `[-]` for duration value. I think, log format should be the same for profiling messages and for other log messages. @samdark could you please bring your expertise here?

framework/log/Logger.php
```diff
@@ -311,4 +332,22 @@ public static function getLevelName($level)
...
+
+    /**
+     * Calculates the elapsed time for the given profile end token.
+     * @param string $endToken token for the code block.
+     * @param float $time the profile end timestamp.
+     * @return float the elapsed time in seconds.
+     */
+    private function calculateProfileDuration($endToken, $time)
+    {
+        foreach ($this->_profileBeginnings as $i => $beginning) {
```
cebe: as far as I see you could index the `$this->_profileBeginnings` array by `$token` and then select the necessary token to calculate timing instaead of iterating over the array. Or am I missing something?

samdark:
Solved in 2.1 https://github.com/yiisoft/yii2/pull/14611

samdark:
Won't fix in 2.0.

PR #10056, opened 2015-10-29
Allow all callables instead of only Closures in ActionColumn

freezysko:
Allow all callables instead of only Closures when defining custom urlCreator in ActionColumn


samdark:
Thanks!

PR #10058, opened 2015-10-30
Fix #9133: optimitic lock doesn't work when using load()

### LINKED ISSUE #9133: Optimistic Locking doesn't work if the lock value is not sent
I'm working on a REST API.
If I send an outdated lock value (e. g. `{"some_field":"some new value","version":0}`), I get StaleObjectException, which is expected.
If I do not send the optimistic lock field (e. g. if I send only `{"some_field":"some new value"}`), the record gets updated and everything is fine, which is not expected.
I expect to get StaleObjectException if I do not send the optimistic lock field.

EvgenyOrekhov:
Optimitic lock didn't work when the lock field was missing
in the input data passed to load()

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -1571,4 +1571,20 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
             unset($this->$offset);
         }
     }
+
+    /**
+     * @inheritdoc
+     *
+     * This method sets the optimistic lock value to null before calling
+     * the parent implementation.
+     */
+    public function load($data, $formName = null)
+    {
+        $lock = $this->optimisticLock();
+        if ($lock !== null) {
+            $this->$lock = null;
+        }
+
+        return parent::load($data, $formName);
+    }
 }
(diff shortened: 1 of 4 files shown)


framework/db/BaseActiveRecord.php
```diff
@@ -1571,4 +1571,20 @@ public function offsetUnset($offset)
...
+    /**
+     * @inheritdoc
+     *
+     * This method sets the optimistic lock value to null before calling
+     * the parent implementation.
+     */
+    public function load($data, $formName = null)
+    {
+        $lock = $this->optimisticLock();
+        if ($lock !== null) {
```
EvgenyOrekhov: @samdark 
The documentation of BaseActiveRecord::optimisticLock() says:
> Add a required validation rule for the version column to ensure the version value is submitted.

So maybe this line should be
```php
if ($lock !== null && $this->isAttributeRequired($lock)`) {
```
And looks like I should have added the `['version', 'required']` rule to the Document class. Am I right?

samdark: Well, it expects the value to be submitted...

samdark: So this should be by design. Not sure why though...

samdark: @yiisoft/core-developers any idea?

cebe:
if the data sent does not contain a version, how could we check if data was changed? I do not see a problem in current code.

PR #10065, opened 2015-10-31
Update AssetManager.php

nikitindv:
Add ability to define and override all assetsBundle options without necessarily enumeration of all used vendor bundles.

### DIFF
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -47,7 +47,7 @@ class AssetManager extends Component
      * The array keys are the asset bundle names, which typically are asset bundle class names without leading backslash.
      * The array values are the corresponding configurations. If a value is false, it means the corresponding asset
      * bundle is disabled and [[getBundle()]] should return null.
-     * You can use '*' to assign following configuration to all included asset bundles.
+     * You can use '*' key to assign following configuration to all included asset bundles.
      *
      * If this property is false, it means the whole asset bundle feature is disabled and [[getBundle()]]
      * will always return null.
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -260,6 +260,11 @@ class AssetManager extends Component
         if (!isset($config['class'])) {
             $config['class'] = $name;
         }
+        
+        if (isset($this->bundles['*'])) {
+            $config = array_merge($config, $this->bundles['*']);
+        }
+        
         /* @var $bundle AssetBundle */
         $bundle = Yii::createObject($config);
         if ($publish) {
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -47,6 +47,7 @@ class AssetManager extends Component
      * The array keys are the asset bundle names, which typically are asset bundle class names without leading backslash.
      * The array values are the corresponding configurations. If a value is false, it means the corresponding asset
      * bundle is disabled and [[getBundle()]] should return null.
+     * You can use '*' to assign following configuration to all included asset bundles.
      *
      * If this property is false, it means the whole asset bundle feature is disabled and [[getBundle()]]
      * will always return null.


samdark:
Could you elaborate what is it doing?

nikitindv:
@samdark It allows me to override AssetBundle publish options for all included asset bundle, for example except by wildcard all *_spec.js files from BootstrapAsset.

samdark:
Should be added to phpdoc / docs.

nikitindv:
@samdark can i add pull request for documentation?
And if i do, in which language i shall do this?

samdark:
Absolutely. Commit to the same `patch-1` branch. Commits will end up in this pull requests. Docs should be in English.

nikitindv:
@samdark phpdoc updated.

framework/web/AssetManager.php
```diff
@@ -260,6 +261,11 @@ protected function loadBundle($name, $config = [], $publish = true)
         if (!isset($config['class'])) {
             $config['class'] = $name;
         }
+        
+        if (isset($this->bundles['*'])) {
+            $config = array_merge($config, $this->bundles['*']);
```
samdark: Should it be the other way around? i.e. what's specified in `$this->bundles['*']` is default values and what's passed to the method are more specific values.

nikitindv: That argument requires same values as bundle specific value.
31 окт. 2015 г. 18:42 пользователь "Alexander Makarov" <
notifications@github.com> написал:

samdark:
```
```

ntemocentr:
Yes, there we merge bundle specific config with common config. So, if we define some bundle specific parameters, they will be applied as is, but if no - there will be used explicitly defined parameters from * key

ntemocentr:
Offtopic: can we discuss my pull request in native language, and then translate for our foreign collegues, because my english is not good.

samdark:
Yep. Catch me at https://gitter.im/yiisoft/yii2/rus

klimov-paul:
I don't get it: why don't use a DI for this?

samdark:
@klimov-paul example?

klimov-paul:
`AssetBundle` instances are created via `Yii::createObject()`:

https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L264

Thus you should be able to configure thier properties via DI just the same way as we do this for widgets.
I can't see any obstacle for doing this at the first glance.

samdark:
@ntemocentr ?

klimov-paul:
In order to apply certain options for all Asset Bundles, you should be able to do something like following:

```php
$container->set('yii\web\AssetBundle', [
    'publishOptions' => [/* my options*/],
]);
```

ntemocentr:
Imho it is not obvious for junior developers.
So many projects does not use assetic bundles and simpliy puts assets into webroot dir and uses plain links in templates.
DI is cool concept, but there must be simple way to override all default is arameters within any assetBundle, because assetsBundle is point extension.

klimov-paul:
This does not sound like a convincing argument for me.
Problem of overring certain paramters for the several class of the same hierarchy or same functionality is solved long ago with the 'factory' OOP pattern, while "DI" pattern is an eveoled version of it.
I don't think the framework should introduce extra way to do such a thing, while DI alreay present.

ntemocentr:
Thanks, using di a can introduce some syntax sugar within my projects)
I reached enlightenment, this pull request can be closed)

ntemocentr:
@klimov-paul Hello, di doesn't allows me to override properties for inherited bundles, because of https://github.com/yiisoft/yii2/blob/master/framework/BaseYii.php#L341
\yii\BaseYii::createObject, does not look through class parents when applying properties.
So, i can't override BootstrapAsset props using 
$container->set('yii\web\AssetBundle', [
    'publishOptions' => [/* my options*/],
]);

ntemocentr:
lets discuss this question in https://gitter.im/yiisoft/yii2/rus

beowulfenator:
@klimov-paul, your argument in favor of DI pattern is valid. However, right now `AssetManager` has `bundles` property that allows to selectively override settings of certain asset bundles. This patch does not really introduce a new mechanism that has functionality similar to DI, it just extends existing functionality. After all, if we already allow to override settings of certain asset bundles in `AssetManager` config, what sense does it make to prohibit overriding them all?

cebe:
This PR does not look like the right solution to the problem. Please open an issue to discuss possible solutions.

Issue #10076, opened 2015-11-03
TimestampBehavior trying to insert invalid values on MySQL


c-schmitz:
I am using the latest 'release' Yii version 2.0.6.

When trying to use the TimestampBehaviour it tries to insert invalid values instead of  NOW().

Query in trace actually executed is:

    INSERT INTO `responses_day` (`response_count`, `user_id`, `time_period`, `created`, `modified`) VALUES (3, 11, '2015-11-03', '1446550646', '1446550646')

Code of the model:

    <?php

    namespace app\models;
    use yii\db\ActiveRecord;
    use yii\behaviors\TimestampBehavior;

    class ResponseDay extends ActiveRecord
    {
        /**
         * @return string the name of the table associated with this ActiveRecord class.
         */
        public static function tableName()
        {
            return 'responses_day';
        }

        public static function getDb()
        {
            return \Yii::$app->mytestdb;
        }
        public function behaviors()
        {
            return [
                [
                    'class' => TimestampBehavior::className(),
                    'createdAtAttribute' => 'created',
                    'updatedAtAttribute' => 'modified'
                ],
            ];
        }
    }

SQL for table: 

    CREATE TABLE IF NOT EXISTS `responses_day` (
      `id` int(11) NOT NULL AUTO_INCREMENT,
      `user_id` int(11) DEFAULT NULL,
      `response_count` float NOT NULL DEFAULT '0',
      `time_period` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
      `modified` datetime DEFAULT NULL,
      `created` datetime DEFAULT NULL,
      PRIMARY KEY (`id`),
      UNIQUE KEY `Index_2` (`user_id`,`time_period`)
    ) ENGINE=MyISAM DEFAULT CHARSET=utf8;

Insertion code:

            $ResponseDay=new ResponseDay;
            $ResponseDay->response_count=3;
            $ResponseDay->user_id=11;
            $ResponseDay->time_period='2015-11-03';
             $ResponseDay->save();

Expected behaviour:
Created/modified should be filled with the actual time stamp.

Instead:
Entries which contain '0000-00-00 00:00:00' are created.

samdark:
Can you try removing `DEFAULT '0000-00-00 00:00:00',`?

c-schmitz:
[...]

PR #10081, opened 2015-11-03
Fixed issue #11076: TimestampBehavior trying to insert invalid default values

c-schmitz:
Uses now as default the DB ANSI compliant method for current datetime stamp which is compatible across DB types.


c-schmitz:
#10076

samdark:
Nope. It won't be changed.

1. People are using UNIX timestamp (int) as well as datetime. These two methods are equally popular.
2. It breaks backwards compatibility.

PR #10083, opened 2015-11-03
Wrapper for PHP webserver

samdark:
In order to try it checkout basic app template and run `./yii serve` in the app root.

Here's a help:

```
$ ./yii help serve

DESCRIPTION

Runs PHP built-in web server


USAGE

yii serve [address] [...options...]

- address: string (defaults to 'localhost')
  address to serve on


OPTIONS

--appconfig: string
  custom application configuration file path.
  If not set, default application configuration is used.

--color: boolean, 0 or 1
  whether to enable ANSI color in the output.
  If not set, ANSI color will only be enabled for terminals that support it.

--docroot: string (defaults to 'web')
  path to directory to serve

--interactive: boolean, 0 or 1 (defaults to 1)
  whether to run the command interactively.

--port: int (defaults to 8080)
  port to serve on. Either "host" or "host:port".

--router: string
  path to router script.
  See https://secure.php.net/manual/en/features.commandline.webserver.php
```


framework/console/controllers/ServeController.php
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * @param string $address server address
+     * @return boolean if another server is running at address specified
+     */
+    protected function isOtherServerProcessRunning($address)
+    {
+        return file_exists($this->getLockFile($address));
+    }
+}
```
zelenin: no new line in the end of file )

framework/console/controllers/ServeController.php
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Runs PHP built-in web server
+ *
+ * In order to access server from remote machines use 0.0.0.0:8000. That is especially useful when running server in
+ * a virtual machine.
+ *
+ * @author Alexander Makarov <sam@rmcreative.ru>
+ * @since 2.0.7
+ */
+class ServeController extends Controller
```
zelenin: Serve? Why not Server?

framework/console/controllers/ServeController.php
```diff
@@ -0,0 +1,141 @@
...
+    public $router;
+
+    /**
+     * Runs PHP built-in web server
+     *
+     * @param string $address address to serve on
+     *
+     * @return int
+     */
+    public function actionIndex($address = 'localhost')
```
zelenin: where is start, stop, restart? It seems like unfinished.

xicond: I think it's not a daemon background service that should have start, stop, restart

framework/console/controllers/ServeController.php
pana1990: Laravel use serve too.

pana1990:
:+1:

framework/console/controllers/ServeController.php
samdark: Sounds better:

```
./yii serve localhost

vs 

./yii server localhost
```

framework/console/controllers/ServeController.php
samdark: @xicond is right. It's not daemon.

samdark:
@pana1990 no since there's no such concept at Yii app level.

framework/console/controllers/ServeController.php
samdark: Added.

framework/console/controllers/ServeController.php
samdark: Also it's consistent with `./yii migrate`.

framework/console/controllers/ServeController.php
```diff
@@ -0,0 +1,141 @@
...
+            $this->stdout("http://$address is taken by another process.\n", Console::FG_RED);
+            return self::EXIT_CODE_ADDRESS_TAKEN_BY_ANOTHER_PROCESS;
+        }
+
+        if ($this->router !== null && !file_exists($this->router)) {
+            $this->stdout("Routing file \"$this->router\" does not exist.\n", Console::FG_RED);
+            return self::EXIT_CODE_NO_ROUTING_FILE;
+        }
+
+        $this->stdout("Server started on http://{$address}/\n");
```
pana1990: I prefer "Yii {current version} starting in {enviroment} on ..."

what do you think about it?

samdark: 1. There's no `{enviroment}`.
2. Why do we need version here?

framework/console/controllers/ServeController.php
zelenin: but how will I stop server? it' not for some production uses, just for test.
It's daemon because i cannot stop it )

framework/console/controllers/ServeController.php
pana1990: 1. use YII_ENV constant?
2. info additional, laravel&rails&other frameworks use this one.

framework/console/controllers/ServeController.php
samdark: CTRL+C or COMMAND+C as written right after starting the server.

samdark: https://github.com/yiisoft/yii2/pull/10083/files#diff-1360e0aa03b2bc341ea828f4593560c4R87

zelenin: ok

framework/console/controllers/ServeController.php
samdark: 1. How?
2. Well, the fact that others are using it doesn't mean it's actually useful. What's the value in the framework version in this particular case?

pana1990:
```php
$env = YII_ENV;
$version = Yii::getVersion();
$this->stdout("Yii $version server started in $env on http://{$address}/\n");
```

samdark: 1. OK. We can pass that env to the PHP server. How would it be applied to the application?

pana1990: basic and advaced templates define YII_ENV, what's the problem?

samdark: How would you pass a constant from webserver to the app?

pana1990: sorry, i was understanding... i only want to include YII_ENV in ouput but it would pass to use [this] (http://stackoverflow.com/questions/13784116/setting-environment-variables-with-the-built-in-php-web-server), but it is bit tricky imo.

samdark: Yii isn't reading environment variables by default so it doesn't make much sense.

pana1990: i agree but i only like show YII_ENV value (in script entry) in output.

PR #10087, opened 2015-11-04
Allow set params from actions

arturf:
use case

``` php
class UsersViewAction extends \yii\rest\ViewAction
{
    /**
     * @inheritdoc
     */
    public function run($id = null, $expand = 'profile')
    {
        if (is_null($id)) {
            $id = \Yii::$app->user->id;
        }

        return parent::run($id);
    }
}
```

and we have clean url like this `/api/v1/user/`

### DIFF
--- a/framework/rest/Serializer.php
+++ b/framework/rest/Serializer.php
@@ -155,8 +155,11 @@ class Serializer extends Component
      */
     protected function getRequestedFields()
     {
-        $fields = $this->request->get($this->fieldsParam);
-        $expand = $this->request->get($this->expandParam);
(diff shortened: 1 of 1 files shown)


cebe:
I don't really get it. what is the code you would write now and how does it change with this feature?

arturf:
@cebe 
to include default values for actions. With this changes we can get extra fields without changing url:
`GET /api/users/` 
instead of 
`GET /api/users/?expand=profile`

cebe:
default params for URLs should be added in the UrlManager, not at action level like this. Also adding this to Serializer and not making it available in other parts of the framework is inconsistent. I can not accept this PR in this way. If you want to discuss the problem further, please open an issue. Thanks for your contribution!

PR #10091, opened 2015-11-04
Fix arabic language

hableel:
One of the translations has a wrong variable.
variable "compareValue" is written instead of  "max"


samdark:
Thanks!

PR #10093, opened 2015-11-04
BaseInflector::transliterate() is now public. Introduced different levels of transliteration strictness

### LINKED ISSUE #10061: Change Inflector::transliterate to public
In order to support transliteration, I have to install another composer module (yii2-transliteration) to do so.

Just found that simple conversion can be made with our transliterate.

Why don't we change `protected` to `public static transliterate` instead? It quite convenient some time.

Thanks,

SilverFire:
Closes #10061 
- [x] Code
- [x] Tests
- [x] Changelog
- [x] Docs
- [x] Code review

:exclamation: Changes:
- Default `$transliterator` changed from `Any-Latin; NFKD` to `Any-Latin; Latin-ASCII; [\u0080-\uffff] remove`, so the slug will be generated more correctly. Leaded to changing a test-case on old line 180

### DIFF
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -14,6 +14,10 @@ Make sure you have global install of latest version of composer asset plugin:
 php composer.phar global require "fxp/composer-asset-plugin"
 ```
 
+Upgrade from Yii 2.0.6
+----------------------
+  
+* The signature of `yii\helpers\BaseInflector::transliterate()` was changed. The method is now public and has an extra optional parameter `$transliterator`.
 
 Upgrade from Yii 2.0.5
 ----------------------
--- a/framework/helpers/BaseInflector.php
+++ b/framework/helpers/BaseInflector.php
@@ -217,7 +217,7 @@ class BaseInflector
         'Yengeese' => 'Yengeese',
     ];
     /**
-     * @var array fallback map for transliteration used by [[slug()]] when intl isn't available.
+     * @var array fallback map for transliteration used by [[transliterate()]] when intl isn't available.
      */
     public static $transliteration = [
         'À' => 'A', 'Á' => 'A', 'Â' => 'A', 'Ã' => 'A', 'Ä' => 'A', 'Å' => 'A', 'Æ' => 'AE', 'Ç' => 'C',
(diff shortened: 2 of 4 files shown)


framework/helpers/BaseInflector.php
```diff
@@ -231,13 +231,46 @@ class BaseInflector
...
+     * transliterated to characters of ASCII table. For example:
+     * `获取到 どちら Українська: ґ,є, Српска: ђ, њ, џ! ¿Español?` will be transliterated to
+     * `huo qu dao dochira Ukrainska: g,e, Srpska: d, n, d! Espanol?`
+     *
+     * You can set your own aliases.
+     * For detailed information see [unicode normalization forms](http://unicode.org/reports/tr15/#Normalization_Forms_Table)
+     * @see http://unicode.org/reports/tr15/#Normalization_Forms_Table
+     * @see transliterate()
+     */
+    public static $transliteratorAliases = [
```
samdark: Instead of array constants could be used:

```php
const TRANSLITERATOR_STRICT = 'Any-Latin; NFKD';
const TRANSLITERATOR_MEDIUM = 'Any-Latin; Latin-ASCII';
const TRANSLITERATOR_LOOSE = 'Any-Latin; Latin-ASCII; [\u0080-\uffff] remove';
```

framework/helpers/BaseInflector.php
```diff
@@ -438,12 +471,24 @@ public static function slug($string, $replacement = '-', $lowercase = true)
...
      *
      * @param string $string input string
+     * @param string|\Transliterator $transliterator either a [[Transliterator]] or a string
+     * from which a [[Transliterator]] can be built. When $transliterator is a string, the method will
+     * try to treat is as an alias name and use transliterator value from [[transliteratorAliases]] array.
      * @return string
+     * @see transliteratorAliases
      */
-    protected static function transliterate($string)
+    public static function transliterate($string, $transliterator = null)
```
samdark: This is potential BC break. Not a serious one but should be mentioned in UPGRADE guide.

framework/UPGRADE.md
```diff
@@ -14,6 +14,10 @@ Make sure you have global install of latest version of composer asset plugin:
 php composer.phar global require "fxp/composer-asset-plugin"
 ```
 
+Upgrade from Yii 2.0.6
+----------------------
+  
+* The signature of `yii\helpers\BaseInflector::transliterate()`` was changed. The method is now public and has an extra optional parameter `$transliterator`.
```
samdark: Remove extra `

samdark:
Looks good to me.

samdark:
You have "docs" not ticked. Do you plan to do some extra polishing?

SilverFire:
Squashed and ready to be merged

samdark:
Perfect. Thanks!

framework/helpers/BaseInflector.php
```diff
@@ -231,13 +231,53 @@ class BaseInflector
...
+     */
+    const TRANSLITERATE_LOOSE = 'Any-Latin; Latin-ASCII; [\u0080-\uffff] remove';
+
+
+    /**
+     * @var mixed Either a [[\Transliterator]], or a string from which a [[\Transliterator]] can be built
+     * for transliteration. Used by [[transliterate()]] when intl is available. Defaults to [[TRANSLITERATE_LOOSE]]
+     * @see http://php.net/manual/en/transliterator.transliterate.php
+     */
+    public static $transliterator = self::TRANSLITERATE_LOOSE;
```
cebe: what are the arguments for changing the default value? as far as I see, this is a BC break that directly affects how application code works. no?

SilverFire: Yes, it is and I mentioned about it. Also there is a one [changed test](https://github.com/yiisoft/yii2/pull/10093/files#diff-2fbefc87fa0366130221976f84b37861R288). With new default, transliteration is more precise than before.
Maybe we should mention in in UPGRADE guide.

SilverFire: Nobody can guarantees that transliteration or formatting using i18n will be steady release-to-release because each ICU update changes the rules of formatting or transliteration for languages and nobody complains on it.
My change does not break the transliteration, as you see. It just gives a little more precise result for some cases.

samdark: @SilverFire is correct. Transliteration already varies depending on ICU version. There was never guarantee that result will be the same.

cebe: okay, thanks.

PR #10095, opened 2015-11-05
Use the ActiveQuery syntax for examples in doc

Faryshta:


SilverFire:
Looks good

samdark:
It's OK but `{{table}}` syntax becomes kinda hidden. It's not easy to figure out you can do it.

Faryshta:
Maybe for that we would need a different section, I think thats more for the ´ActiveQuery´ class than ´ActiveRecord´

samdark:
Overall I don't think it should be merged as is.

Faryshta:
at least we should change lines like

```
 'COUNT({{order}}.id) AS ordersCount'
// to
'ordersCount' => 'COUNT({{order}}.id)'
```

Faryshta:
@samdark what do you say about changing the alias usage on the select?

samdark:
I see no profit in changing it.

Faryshta:
its more adhoc with the documentation and the way its adviced to write the alias in yii2.

samdark:
Where is it advised to write alias like that?

PR #10096, opened 2015-11-05
Examples for extra field values from internal data

Faryshta:


docs/guide/db-active-record.md
```diff
@@ -1422,3 +1422,84 @@ $customers = Customer::find()
     ->groupBy('{{customer}}.id') // group the result to ensure aggregation function works
     ->all();
 ```
+
+A disadvantage of using this method would be that if the information isn't loaded on the sql query it has to be calculated separatedly, which also means that newly saved records won't contain the information from any extra field
```
SilverFire: sql -> SQL
separatedly -> separately
`field` -> `field.`

docs/guide/db-active-record.md
```diff
@@ -1422,3 +1422,84 @@ $customers = Customer::find()
...
+```
+
+Using the magic methods [[yii\db\BaseActiveRecord::__get()|__get()]] and [[yii\db\BaseActiveRecord::__set()|__set()]] we can emulate the behavior of a property
+
+```php
+class Room extends \yii\db\ActiveRecord
+{
+    private $_volume;
+    
+    public function setVolumen($volume)
```
SilverFire: `setVolumen` -> `setVolume`

docs/guide/db-active-record.md
```diff
@@ -1422,3 +1422,84 @@ $customers = Customer::find()
...
+        }
+        
+        return $this->_volume;
+    }
+
+    // ...
+}
+```
+
+With this if the select query doesn't provide the volume the record will obtain it from its internal value.
```
SilverFire: When the select query doesn't provide the volume, model will be able to calculate it manually.

docs/guide/db-active-record.md
```diff
@@ -1422,3 +1422,84 @@ $customers = Customer::find()
...
+
+Similary it can be used on extra fields depending on relational data
```
SilverFire: I'm not sure, that we need one more example of the same feature.

docs/guide/db-active-record.md
Faryshta: It will be automatic, not manually since the user won't need to declare it. I think we can change it to.

> When the select query doesn't provide the volume, model will be able to calculate it automatically from the attributes in the model.

docs/guide/db-active-record.md
Faryshta: This one is meant for relational data, The logic is a little different because of that.

SilverFire:
It's fine now

samdark:
Merged with minor adjustments.

samdark:
Thank you!

samdark:
@SilverFire thanks for review.

githubjeka:
:point_up: [November 5, 2015 9:43 PM](https://gitter.im/yiisoft/yii2/rus?at=563ba34c0379c7e74069534f)

Faryshta:
@githubjeka agree but I am following https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md#42-properties

githubjeka:
@Faryshta https://github.com/yiisoft/yii2/pull/10105 OK

Faryshta:
how about the line I shared? will that be changed too?

SilverFire:
@Faryshta what line are you talking about?

Faryshta:
@samdark @githubjeka @SilverFire 

Private variables should be named like $_varName.

 https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md#42-properties

PR #10102, opened 2015-11-05
Added method call `saveAssignments` to `\yii\rbac\PhpManager::removeItem`

rezident:
…tem`

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -53,6 +53,7 @@ Yii Framework 2 Change Log
 - Chg #9369: `Yii::$app->user->can()` now returns `false` instead of erroring in case `authManager` component is not configured (creocoder)
 - Chg #9411: `DetailView` now automatically sets container tag ID in case it's not specified (samdark)
 - Chg #9953: `TimestampBehavior::getValue()` changed to make value processing consistent with `AttributeBehavior::getValue()` (silverfire)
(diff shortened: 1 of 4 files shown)


framework/CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ Yii Framework 2 Change Log
 - Chg #9369: `Yii::$app->user->can()` now returns `false` instead of erroring in case `authManager` component is not configured (creocoder)
 - Chg #9411: `DetailView` now automatically sets container tag ID in case it's not specified (samdark)
 - Chg #9953: `TimestampBehavior::getValue()` changed to make value processing consistent with `AttributeBehavior::getValue()` (silverfire)
+- Bug #10101: Added method call `saveAssignments` to `\yii\rbac\PhpManager::removeItem` (rezident1307)
```
SilverFire: I think it's better to describe the fixed bug instead of changed line of code.
How about `Fixed assignments saving on role removing`?

SilverFire:
It happened because we do not have a proper test.
@rezident1307 could you add a test to [PhpManagerTest.php:143](https://github.com/yiisoft/yii2/blob/master/tests/framework/rbac/PhpManagerTest.php#L143)?

rezident1307:
Ok. I will add a new commit later

rezident1307:
Look a new pool request #10109

PR #10105, opened 2015-11-05
Update db-active-record.md

githubjeka:
PSR-2 4.1: 
"Property names SHOULD NOT be prefixed with a single underscore to indicate protected or private visibility."

### DIFF
--- a/docs/guide/db-active-record.md
+++ b/docs/guide/db-active-record.md
@@ -1441,11 +1441,11 @@ we can emulate the behavior of a property
 ```php
 class Room extends \yii\db\ActiveRecord
 {
-    private $_volume;
+    private $volume;
     
     public function setVolume($volume)
     {
-        $this->_volume = (float) $volume;
+        $this->volume = (float) $volume;
     }
     
     public function getVolume()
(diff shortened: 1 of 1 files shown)


samdark:
Thanks!

cebe:
should not, but it is not forbidden. also yii code style is so that private is with underscore...

samdark:
This is not core but example docs. As for the core, I'd like to change it removing underscore...

klimov-paul:
I agreed with @cebe.
There is a reason why we use underscore for private and protected prperties: `yii\base\Object` introduces virtual properties based on setters and getters. So method `getVolume()` produces property `volume`. If no underscore is present at the field itself there is no way to use virtual property inside the class:

```
class Room
{
    // ...
    public function foo()
    {
        echo $this->volume; // expected `getVolume()` call, while private property accessed directly instead.
    }
}
```

Underscore prefix is meant to resolve naming conflict - it not just a "pretty convension" as someone may think.

cebe:
> If no underscore is present at the field itself there is no way to use virtual property inside the class:

:+1:

cebe:
therefor this code style has a good justification as it prevents confusion and errors. This is also more important in example and application code than in the core.

samdark:
OK.

samdark:
https://github.com/yiisoft/yii2/commit/709414554712afb3e3c518347a8e24e97f2f52dc

githubjeka:
`__get()` is utilized for reading data from inaccessible properties.

**Because:**

```php
class Room
{
    // ...
    public function foo()
    {
        echo $this->volume; // expected ` $this->volume`!!! not _get()
    }
}
```


But I think samdark  was tired and so everything will stay as it is. :(

PR #10106, opened 2015-11-05
Migration update

dkulyk:
Fix quote default value(use scheme);
Add unsigned contains in ColumnSchemaBuilder(type control by scheme);

### DIFF
--- a/framework/db/ColumnSchemaBuilder.php
+++ b/framework/db/ColumnSchemaBuilder.php
@@ -39,6 +39,10 @@ class ColumnSchemaBuilder extends Object
      */
     protected $isUnique = false;
     /**
+     * @var boolean whether the column values should be unsigned. If this is `true`, a `UNSIGNED` constraint will be added.
+     */
+    protected $isUnsigned = false;
+    /**
      * @var string the `CHECK` constraint for the column.
      */
     protected $check;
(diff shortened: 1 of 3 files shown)


framework/db/ColumnSchemaBuilder.php
```diff
@@ -83,6 +93,16 @@ public function unique()
     }
 
     /**
+     * Adds a `UNSIGNED` constraint to the column.
+     * @return $this
+     */
+    public function unsigned()
+    {
+        $this->isUnique = true;
```
Renkas: Shouldn't this be:
$this->isUnsigned = true;

samdark: Definitely.

framework/db/ColumnSchemaBuilder.php
```diff
@@ -46,17 +50,23 @@ class ColumnSchemaBuilder extends Object
      * @var mixed default value of the column.
      */
     protected $default;
+    /**
+     * @var Schema $schema Used schema
+     */
+    protected $schema;
```
cebe: schema should not be coupled to this class. we already have DBMS specific implementations of ColumnSchemaBuilder.

SilverFire:
I don't think that this PR can be merged as is.

Issue #9535, opened 2015-08-27
$myForm.yiiActiveForm('validate') does NOT run form validation


tekord:
I have the form:

```
<?php $form = ActiveForm::begin([
                'enableClientValidation' => false,
                'enableAjaxValidation' => true,
                'validationUrl' => ['validate-form-step-2'],
                'validateOnBlur' => false,
                'validateOnChange' => false
            ]) ?>
```

This form validates only on button click. Button click handler:

```
// Getting my form
var $form = $('#<?= $form->id ?>');

// Run validation
$form.yiiActiveForm('validate');
```

This does nothing. yiiActiveForm has method 'validate', nothing happens.

Validation runs only if I call this:

```
$form.yiiActiveForm('submitForm')
```

I tried to subscribe on event 'beforeValidate' on my form. Event triggered only on call 'submutForm'.

How to run form validation programmatically? Is it a bug?

samdark:
You have it turned off:

> 'enableClientValidation' => false,

In order to validate via AJAX you have to submit form.

tekord:
No, even if I set enableClientValidation = true it still does not work.

Chrome console:
```
$('#w0').yiiActiveForm('validate')
undefined
```

samdark:
Yes, it returns nothing but it triggers validation.

tekord:
If it triggers validation then handler of 'beforeValidate' event should be triggered. It does not happened.

```
$form.on('beforeValidate', function (event, messages, deferreds) {
        console.log('IT WORKS');
    });
```
Also there is no any validation result messages (red color highlighting and error summary updating).

tekord:
Any progress here?

samdark:
No.

bluezed:
Just ran into the same issue.

I need to trigger the form validation without actually submitting the form and 
``` $('#my-form').yiiActiveForm('validate'); ```
does not seem to do any validation.

I looked into the JS-code and it looks like the issue is the ``` this.status ``` that gets checked in the validate function:
```
// client-side validation
$.each(data.attributes, function () {
    this.cancelled = false;
    // perform validation only if the form is being submitted or if an attribute is pending validation
    if (data.submitting || this.status === 2 || this.status === 3) {
```
and the updateInputs function:
```
} else {
    $.each(data.attributes, function () {
        if (!this.cancelled && (this.status === 2 || this.status === 3)) {
            updateInput($form, this, messages);
        }
    });
}
```

[...]

PR #10110, opened 2015-11-06
Add validation for attributes that are untouched

bluezed:
Proposed fix for #9535

### DIFF
--- a/framework/assets/yii.activeForm.js
+++ b/framework/assets/yii.activeForm.js
@@ -303,8 +303,8 @@
             $.each(data.attributes, function () {
                 if (!$(this.input).is(":disabled")) {
                     this.cancelled = false;
-                    // perform validation only if the form is being submitted or if an attribute is pending validation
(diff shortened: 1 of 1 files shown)


SilverFire:
It leads to a red form full of validation errors after you touched at least one field.

I open a form:
screenshot from 2015-11-06 14-23-50

Then I touch the first field. BOOM!
screenshot from 2015-11-06 14-26-15


I'm voting against this idea. Maybe we can create an option for input like `validate-always`.

bluezed:
I see, didn't notice that in my use-case, but I agree that the solution cannot be accepted like that.

I was wondering if it would be feasible to add a new function just for triggering validation manually? That'd not affect any existing code.

SilverFire:
You can already do it manually:

```php
echo $form->field($model, "type");
echo $form->field($model, "value");
```

```js
$('#model-type').on('change', function () {
    $(this).closest('form').yiiActiveForm('validateAttribute', 'model-value');
});
```

Issue #10108, opened 2015-11-06
Add support for events declared in the interfaces


omnilight:
Currently using `Event` object we can subscribe only on events, that are declared in some class or it's parent. But sometimes I want to declare event in the interface, implement it by some class, and subscribe on event in this interface, not in the class, example:

```php
interface SomeInterface
{
    const EVENT_SOME_EVENT = 'someEvent';
}

class SomeClass extends Component implements SomeInterface
{
   public function domSomething() {
       $this->trigger(self::EVENT_SOME_EVENT);
   }
}

Event::on(SomeInterface::class, SomeInterface::EVENT_SOME_EVENT, function() {
   // My handler
});
```

But in current implementation of the `Event::trigger` method only events for parent classes are checked, but not from interfaces. May be we should add checking for events in the interfaces too?

samdark:
Any idea how implementation could look like?

omnilight:
In the pull-request

samdark:
OK. Looks quite interesting but also makes events not as lightweight as they used to be. Worth benchmarking...

SilverFire:
https://github.com/yiisoft/yii2/commit/b68613b7469a7e5df42e0f532cc5370964b07899 https://github.com/yiisoft/yii2/commit/e0f4114d77bef0e3f1847801b3a41f5a9f6dc62e

PR #10113, opened 2015-11-06
#10108 add support for events in interfaces

omnilight:
Adds support for events in interfaces #10108


SilverFire:
I like this idea! @samdark said in the [comment](https://github.com/yiisoft/yii2/issues/10108#issuecomment-154513513) that PR can drop speed of execution, but the code looks simple. Can you make a benchmark to check the difference?

omnilight:
Ok, I've tested my code with the following script:

```php
namespace app\commands;
use yii\base\Component;
use yii\base\Event;
use yii\console\Controller;


/**
 * Class SpeedTestController
 */
class SpeedTestController extends Controller
{
    public $counter = 0;

    public function actionIndex()
    {
        Event::on(SomeClass::class, SomeInterface::EVENT_FROM_INTERFACE, function (Event $e) {
            $this->counter++;
        });

        $class = new SomeClass();

        $start = microtime(true);
        for ($i = 0; $i < 10000; $i++) {
            $class->emitEvent();
        }
        $end = microtime(true);

        echo $end - $start;
    }
}

interface SomeInterface
{
    const EVENT_FROM_INTERFACE = 'fromInterface';
}

class SomeClass extends Component implements SomeInterface
{
    public function emitEvent()
    {
        $this->trigger(self::EVENT_FROM_INTERFACE);
    }
}
```

Results:

Code | Result
-------  | ---------
Original | 0.4721
With pull-request | 1.45

I noticed, that `call_user_func` takes most of the time, so I've refactored code, after that:

Code | Result
-------  | ---------
Refactoring | 0.51

A bit slower, but not to much

omnilight:
It is also possible to cache results of the `class_implements`, but it will increase amount of memory used

samdark:
0.4721 to 0.51 for 10000 events isn't much but... how it behaves if there are multiple handlers?

omnilight:
```php
namespace app\commands;
use yii\base\Component;
use yii\base\Event;
use yii\console\Controller;


/**
 * Class SpeedTestController
 */
class SpeedTestController extends Controller
{
    public $counter = 0;

    public function actionIndex()
    {
        for ($i = 0; $i < 100; $i++) {
            Event::on(SomeClass::class, 'event_'.$i, function (Event $e) {
                $this->counter++;
            });
        }

        $class = new SomeClass();

        $start = microtime(true);
        for ($i = 0; $i < 10000; $i++) {
            $class->emitEvent();
        }
        $end = microtime(true);

        echo $end - $start;
    }
}

interface SomeInterface
{

}

class SomeClass extends Component implements SomeInterface
{
    public function emitEvent()
    {
        for ($i = 0; $i < 100; $i++) {
            $this->trigger('event_'.$i);
        }
    }
}
```

Results:

Code | Result
-------- | ---------
Original | 45.301
Pull-request | 49.247

But if some class implements a lot of interfaces, it could be worth

cebe:
> But if some class implements a lot of interfaces, it could be worth

that is not a common case so that should be fine.

omnilight:
Oh, I've noticed, that @samdark meant "multiple handlers", ok, this code:

```php
<?php

namespace app\commands;

use yii\base\Component;
use yii\base\Event;
use yii\console\Controller;


interface SomeInterface
{

}

/**
 * Class SpeedTestController
 */
class SpeedTestController extends Controller
{
    public $counter = 0;

    public function actionIndex()
    {
        for ($i = 0; $i < 100; $i++) {
            Event::on(SomeClass::class, 'event', function (Event $e) {
                $this->counter++;
            });
        }

        $class = new SomeClass();

        $start = microtime(true);
        for ($i = 0; $i < 10000; $i++) {
            $class->emitEvent();
        }
        $end = microtime(true);

        echo $end - $start;
    }
}

class SomeClass extends Component implements SomeInterface
{
    public function emitEvent()
    {
        $this->trigger('event');
    }
}
```

will give:

Code | Result
-------- | --------
Original | 19.31
Pull | 19.50

samdark:
Looks acceptable. @cebe, @klimov-paul what do you think?

framework/base/Event.php
```diff
@@ -192,5 +203,17 @@ public static function trigger($class, $name, $event = null)
                 }
             }
         } while (($class = get_parent_class($class)) !== false);
```
klimov-paul: Why don't use `class_parents()` here then?
Both cycles can be unified this way.

klimov-paul:
Perhaps it could be better to use `ReflectionClass` for both parent class and interface checks.

framework/base/Event.php
omnilight: In original algorithm check for event listeners is performed in the following way:

1. Listeners for $class
2. Listeners for $class's parents
3. Listeners for $class's parents's parentents
4. So on...

If we will use `class_parents()` an order will be unpredictable

omnilight:
Will check for perfomance in that case...

SilverFire:
@omnilight any news?

omnilight:
I'll work on this current weekend, and will update current issue

omnilight:
Ok, I've tested reflection implementation of the event class (when we simply replaces calls to class_parent and class_implements to their reflection equivalent):

Version | Speed
---------- | ---------
Original | 18.72
Reflection | 19.28

So reflection is slower, and there is no reason to use it (and it does not give us any improvements on the algorithm of searching for class parents and it's interfaces)

I've updated the pull-request to use class_parents, as suggested by @klimov-paul - as I can understand this method parents in the correct order

samdark:
Now it looks good to me. Need docs though.

omnilight:
Ok, what kind of docs I should provide?

framework/base/Event.php
```diff
@@ -24,6 +24,7 @@
  */
 class Event extends Object
 {
+    private static $_events = [];
```
SilverFire: We need a PHPDoc here

framework/base/Event.php
```diff
@@ -145,11 +143,18 @@ public static function hasHandlers($class, $name)
         } else {
             $class = ltrim($class, '\\');
         }
-        do {
+
+        $classes = array_merge(
```
SilverFire: We need to add update PHPDoc for `hasHandlers` method and mention new feature there

SilverFire:
@omnilight there is a guide for [Events](http://www.yiiframework.com/doc-2.0/guide-concept-events.html) and you feature should be mentioned there as well.

omnilight:
I'll try to write docs, but I'll need help with english :)

SilverFire:
@omnilight можешь написать доку на русском, а я переведу на английский

omnilight:
@SilverFire Спасибо, напишу :)

omnilight:
Added docs in russian

docs/guide-ru/concept-events.md
```diff
@@ -222,6 +222,65 @@ Event::off(Foo::className(), Foo::EVENT_HELLO, $handler);
 Event::off(Foo::className(), Foo::EVENT_HELLO);
 ```
 
+Обработчики событий на уровне интерфейсов <span id="interface-level-event-handlers"></span>
+-------------
+
+Еще более общим случаем, чем обработка событий на уровне класса, является обработка на уровнне *интерфейсов*. Для этого по аналогии с обработчиками событий на уровне классов, необходимо вызвать статический метод [[yii\base\Event::on()]], передав ему в качестве первого параметра имя интерфейса. В этом случае инициализация события для любого класса, реализующего указанный интерфейс, будет приводить к срабатыванию соответствующего обработчика.
```
SilverFire: `Для этого` -> `Для этого, `

SilverFire:
How about https://gist.github.com/SilverFire/467b2def3115ecc38c43? :)

omnilight:
Greate!, I'll include englush version and will change russian variant

SilverFire:
@omnilight are you willing to continue work on this PR?

omnilight:
Yes, shure, I'll work on this PR during current weekend, sorry for delay

SilverFire:
Thank you very much for you effort! :+1:

SilverFire:
@omnilight if you have no time to do it - i'll continue your work

omnilight:
@SilverFire Yes, I do not have time for this now, so it would be greate if you work on this PR, thank you!

SilverFire:
Merged! Thank you! https://github.com/yiisoft/yii2/commit/b68613b7469a7e5df42e0f532cc5370964b07899
https://github.com/yiisoft/yii2/commit/e0f4114d77bef0e3f1847801b3a41f5a9f6dc62e

PR #10119, opened 2015-11-07
numerous refactors for SluggabeBehavior class, fixes #10118

### LINKED ISSUE #10118: extending SluggableBehavior
I need a slug field for my model and it's going to be shown in url address. current `yii\helpers\Inflector::slug()` method does not produce what I need (it removes characters of my language) so I easily extended it and made needed changes.
The problem is there is no easy way to make SluggableBehavior use the new `slug()` method instead of default one! there is a 50-line-length method named [`getValue()`](https://github.com/yiisoft/yii2/blob/master/framework/behaviors/SluggableBehavior.php#L132) and among so many other things, it produces the real slug value to be saved. Am I supposed to extend the behavior and copy the whole method to edit just [1 line](https://github.com/yiisoft/yii2/blob/master/framework/behaviors/SluggableBehavior.php#L157) of it?
I think it should have a `protected generateSlug()` method which makes it easy to extend and alter slug creation logic. or is there another way I'm not aware of?

hesna:
it was made to fix #10118 but @cebe was way faster :+1: 
it still applies some refactors which I find useful.

### DIFF
--- a/framework/behaviors/SluggableBehavior.php
+++ b/framework/behaviors/SluggableBehavior.php
@@ -131,49 +131,49 @@ class SluggableBehavior extends AttributeBehavior
      */
     protected function getValue($event)
     {
-        $isNewSlug = true;
-
         if ($this->attribute !== null) {
-            $attributes = (array) $this->attribute;
-            /* @var $owner BaseActiveRecord */
-            $owner = $this->owner;
-            if (!empty($owner->{$this->slugAttribute})) {
-                $isNewSlug = false;
-                if (!$this->immutable) {
-                    foreach ($attributes as $attribute) {
-                        if ($owner->isAttributeChanged($attribute)) {
-                            $isNewSlug = true;
-                            break;
-                        }
-                    }
-                }
+            if ($this->isNewSlugNeeded()) {
+                $slug = $this->generateSlug($this->getSlugParts());
+            }else{
+                return $this->owner->{$this->slugAttribute};
             }
-
-            if ($isNewSlug) {
-                $slugParts = [];
-                foreach ($attributes as $attribute) {
-                    $slugParts[] = $owner->{$attribute};
-                }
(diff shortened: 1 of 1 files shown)


cebe:
Sorry for the late reply, could you explain why you think these refactorings are needed?
I currently do not really see the value of splitting out a lot of methods.

hesna:
@cebe no problem. actually I wouldn't touch it if it wasn't for #10118. while fixing that issue, I find it better to refactor the whole class. 
It's just about readability and thus easier maintenance. I know code quality is a relative issue and one may not agree with me, but I find current implementation better than the old one. just having a 50-line method bothers me actually :) (btw it seems [Scrutinizer](https://scrutinizer-ci.com/g/yiisoft/yii2/inspections/fa83399a-d252-461e-b7c6-26d7d6bc721b) likes the new version better)

framework/behaviors/SluggableBehavior.php
```diff
@@ -131,49 +131,49 @@ public function init()
...
-        return $slug;
+
+        return false;
     }
 
     /**
+     * checks if owner model already has a slug value or not.
+     * @return boolean
+     */
+    private function ownersSlugValueIsEmpty()
```
SilverFire: I think this method is redundant.

SilverFire:
I go for this approach because is is more OOP-style, the responsibilities are separated to dedicated methods.

We need to improve PHPDocs to these methods and fix code style, then it's ready to merge.
@cebe may I reassign this PR to myself and finish it?

hesna:
@SilverFire Hi, I tried to fix some coding style problems, please let me know if there are more. sorry about PHPDocs, I'm not good at that.

SilverFire:
Merged with minor adjustments, thank you!
https://github.com/yiisoft/yii2/commit/6936c41cf1a047a2e35b50175d9bf83619f3ab2a

PR #10121, opened 2015-11-08
Catalogue support for models

Faryshta:
I don't know if this is done in other frameworks with other names.

The case usage is that on the database sometimes the columns contain an index or abbreviaton of the value to be shown to the user. Other times the data needs to be translated to be shown and it can be time consuming to type `Yii::t('app', $model->attribute)` everytime its used or create internal methods to do it.

With this there is an introduction to a new functionality named `Model::catalogues()` which handles static data to be used inside the class. The data can be accesed using the `Model::getTerminology()` static method or `Model::getAttributeTerminology` instanced method.

The method `Model::catalogues contain documentation with usage examples.`
- [x] functional methods.
- [x] phpDoc
- [x] Unit tests
- [ ] include it on the model guide.
- [ ] squash commits on another branch

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -1076,11 +1076,12 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      */
     public static function ensureCatalogues()
     {
-        if (self::$_catalogues === null) {
-            self::$_catalogues= self::catalogues();
+        $class = static::className();
+        if (!isset(self::$_catalogues[$class])) {
+            self::$_catalogues[$class] = self::catalogues();
         }
 
-        return self::$_catalogues;
+        return self::$_catalogues[$class];
     }
 
     /**
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -1090,9 +1090,9 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      * @param string $name catalogue to find
      * @return array|null the indexed terms or null if there is no catalogue
      */
-    public static function getCatalogue($name)
+    public static function getCatalogue($catalogue)
     {
-        return ArrayHelper::getValue(self::ensureCatalogues(), $name);
+        return ArrayHelper::getValue(self::ensureCatalogues(), $catalogue);
     }
 
     /**
@@ -1103,7 +1103,7 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      */
     public static function getTerminology($catalogue, $index)
     {
-        return ArrayHelper::getValue(self::getCatalogue($name), $index);
+        return ArrayHelper::getValue(self::getCatalogue($catalogue), $index);
     }
 
     /**
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -14,6 +14,7 @@ use ArrayIterator;
 use ReflectionClass;
 use IteratorAggregate;
 use yii\helpers\Inflector;
+use yii\helpers\ArrayHelper;
 use yii\validators\RequiredValidator;
 use yii\validators\Validator;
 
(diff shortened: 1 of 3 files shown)


Faryshta:
I don't know why it fails on php7

Faryshta:
If this gets accepted I will make an `InCatalogueValidator` to simplify rules and an `CatalogueInput` widget to be used with `ActiveForm`

samdark:
Won't merge it into framework since it's more application layer thing than framework one. Implementation strategy depends very much on the project. If it's a CMS, the approach is OK. Personally, since I'm usually working on non-CMS projects, I'm handling it differently using int field values and constants.

Faryshta:
@samdark this is very related to frameworks, even the yii app advanced and yii app basic templates uses it on the User model.

The most common ussage is for status, there are several status, each one of them with an static meaning. On the database status = 2 means nothing, a person needs to check what it means. So this was originally though as a method to have a way to handle those columns more easily.

Please read the changes on the Model.php file. This is not about CMS.

Faryshta:
@qiangxue @cebe any input on this?

cebe:
@Faryshta generating human readable output for data is not part of the mode layer but more of the view layer. This should be implmented in a formatter class so it can be used in views like gridview and detailview.

Faryshta:
@cebe then why there are methods like `attributeLabels` or `attributeHints` if both are for human readable data.

> This should be implmented in a formatter class so it can be used in views like gridview and detailview.

How about things like `status` for a row status = 1 might mean active, for another might mean suspended, pending or anything else.

nineinchnick:
I agree with @Faryshta, models currently hold presentation layer data anyway.

But I do it outside of a model anyway:

* creating an enum component in my app, 
* registering new enums there (for example in on a module bootstrap) 
* adding a `asEnum` formatter method.

In models I only added a `attributeFormat` method, but it's another topic.

Faryshta:
maybe my mistake was naming it 'catalogues' i should have named it 'attributeFormats' or 'attributeTaxonomies'

samdark:
Personally I'm creating a separate method for each attribute i.e.:

```php
public function getStatuses()
{
   return [
      self::STATUS_NEW => 'New',
      self::STATUS_CLOSED => 'Closed';
   ];
}
```

Faryshta:
I am doing your same approach at the moment @samdark but you have to agree having a method to handle all the catalogues/dictionaries/taxonomies is very useful when you want to add one.

and as i said, i want to build validators and generators for this patch so it can be used more easily.

I think I should change the name of the method since `catalogues` is not very self explanatory.

samdark:
I don't see how having all data available via single method is useful for the case.

Faryshta:
@samdark using your statuses example.

If the method `getStatuses` has functions like translations they will be called everytime the method is invoked causing a performance issue.

and if you want to get the status of a certain model in human readable format you have to do something like

`ArrayHelper::getValue($model->getStatuses, $model->status)`

everytime you use it or create a separate method `getStatusText` which is what i do.

And now repeat for all the enumerable columns you have on all the models you have. With this you don't need to define any other method and it fixes the performance.

klimov-paul:
My opinion: feature is too specific. There are many ways for handling such things as verbose status names.
i personally use static file ActiveRecord:
https://github.com/yii2tech/filedb

Proposed solution can be provided as extension, but should not belong to the core.

samdark:
I tend to agree.

Faryshta:
Should I put it on a behavior or similar? @samdark @klimov-paul @cebe

samdark:
I think either behavior or trait could be a good match.

Faryshta:
I guess the name of 'catalogue' was not very appreciated. Any other name which is more adecuate for this?

TaxonomyTrait, OptionsTrait ?

samdark:
fieldValues?

Faryshta:
AttributeEnumTrait?

samdark:
EnumTrait?

nineinchnick:
I like EnumTrait.

unclead:
+1 for EnumTrait

SilverFire:
+1 EnumTrait

Faryshta:
Ok then I will create the `yii\base\EnumTrait`, `yii\validators\EnumValidator`, `yii\widgets\EnumWidget` so that this can be automatically used for validations and formularies.

samdark:
I've meant that it could be released as an extension. That means namespace should not be `yii` but something your own.

Faryshta:
https://github.com/Faryshta/yii2-enum here is the extension. Can someone help me configure travis and unit testing?

samdark:
Check travis docs and look at https://github.com/yiisoft/yii2-redis/blob/master/.travis.yml and https://github.com/yiisoft/yii2-redis/blob/master/phpunit.xml.dist.

Faryshta:
@samdark https://github.com/Faryshta/yii2-enum i added the files and followed the travis doc. where can i see if its executing?

samdark:
You need to get to repo settings and enable travis.

Faryshta:
@samdark I think I already one it. I added you as collaborator, can you help?

samdark:
Add me as admin. I'll check things, then remove me, OK?

SilverFire:
@Faryshta i left my comments under commits

Faryshta:
I published the new repo here http://faryshta.github.io/yii2-enum

Any feedback and help is welcome.

Right now I am stuck using the yii2-apidoc (or any other phpDoc tool) to generate the documentation for classes and traits.

PR #10123, opened 2015-11-09
Update tutorial-yii-integration.md

asmythe:
I found the existing language to be quite confusing. It took me several attempts of several hours each to get it working on my computer. I have several years of Yii 1 development behind me and have now done four or five Yii 2 apps, so I don't think I'm being dense. I think my language is a lot clearer and more specific.

However, note that I could not get it to work without commenting out a line of vendor/yiisoft/yii2/base/Object.php, which can't be a good thing. Hopefully someone will know how to avoid this problem and will be able to further correct my instructions.

---

After thinking about this overnight, I would amend my directions to recommend that the user copy all the main Yii folders, e.g. controllers, models, etc. into the @app directory, but I can't figure out how to change my commit or add another commit to it.

### DIFF
--- a/docs/guide/tutorial-yii-integration.md
+++ b/docs/guide/tutorial-yii-integration.md
@@ -127,25 +127,11 @@ This can be achieved as described below.
 > Note: Yii 2 requires PHP 5.4 or above. You should make sure that both your server and the existing application
 > support this.
 
-First, install Yii 2 in your existing application by following the instructions given in the [last subsection](#using-yii-in-others).
+Yii 1 typically puts its models, controllers, etc. in a folder called `protected` inside the webroot. Yii 2 typically puts these folders one level above the webroot (@webroot), which it calls the application root (or @app, for short). Depending on how you have your Yii 1 application configured, you may need to set up a parent folder to contain your existing Yii 1 application. We will call that parent folder @app. You'll then want to rename your existing Yii 1 folder `public_html` or `web` or the like. This will be your @webroot.
 
-Second, modify the entry script of the application as follows,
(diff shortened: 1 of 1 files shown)


samdark:
Checked changes in detail. Won't accept these since there are recommendations which should not be recommended:

1. Not using Composer.
2. Using basic template instead of advanced (actually it depends on a project very much).
3. Storing 1.1 app under webroot.
4. Commenting out something in the core framework files in the `vendor` directory.

asmythe:
Thanks Alexander … here are some further thoughts.

On using Composer … the instructions give you the choice of downloading or using composer. I’m used to using composer on my yii2 projects, but not yii1. Maybe this could be clearer? If you are moving from Yii1 to Yii2, here’s how you would use composer in Yii1?

On the basic template versus advanced … again not that clear in the existing docs.

On storing 1.1 under webroot … that is the way the current documentation proposes it be done, is it not? Protected lives inside the webroot. I was just trying to clarify where everything went.

On commenting out some code in the core framework … agreed, that is a non starter. But I could not get the solution proposed in the existing docs to work any other way. Probably just me, but can you confirm that the documented method in fact works?

I’m now doing all my new development with Yii2, which is fabulous. Increasingly I want to get my old Yii1 sites upgrade to Yii2, but it ain’t easy. I’ve decided for now, I’m going to install Yii1 and Yii2 side by side, both working off the same tables. Then I’ll slowly re-create my Y1 classes in Y2, moving more and more of the functionality from Y1 to Y2. That way I can avoid the difficulties of getting the two integrated and make my transition gradually.

I feel there must be many like me struggling with this problem and the documentation could be stronger in explaining the different strategies that one can employ and exactly how to get them to work.

Finally, I just want to express my thanks to the team that wrote Yii1 and Yii2. Using Yii1, and now Yii2, has enabled me to do things that I just could not have done before. They have made me a much better developer, not to mention a lot of money developing really cool projects for clients. I am truly grateful to you and the team for giving us this amazing product.

alec

PR #10124, opened 2015-11-09
UrlRule possible BC break notice

SilverFire:
As a resolve of messaging under 289b051c4a396d3aa0d2078aa24314118bb7c833

Ready to be merged

### DIFF
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -18,6 +18,11 @@ Upgrade from Yii 2.0.6
 ----------------------
   
 * The signature of `yii\helpers\BaseInflector::transliterate()` was changed. The method is now public and has an extra optional parameter `$transliterator`.
+* If you override `yii\web\UrlRule` you have to know that the `pattern` matching group names are being replaced with the
+placeholders on class initialization to support wider range of allowed characters. If you implement `parseRequest()` or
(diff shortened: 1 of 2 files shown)


samdark:
Thanks!

petrabarus:
Just some comment.

This update caused failure in parsing message in my website. Because the URL manager cache the URL rule objects, my site reuse the cache after deployment. It causes `placeholders` field becomes empty array. The `parseRequest` in `yii\web\UrlRule` will return the rule instead of the route.

Simple solution is to invalidate the cache or set `cache` value as `false` in the `URL Manager`.

SilverFire:
@petrabarus it's a good tim! Thank you very much.

@samdark any ideas how can we deal with it automatically? I have one: add a post-update composer action, that will clear the cache forcefully.

samdark:
I don't think invalidation is required on each composer update so I'm for extending upgrade guide instead.

petrabarus:
My site uses zip file for deployment. The composer is after build process. Since it reads from a constant cache directory, it will read old cache data.

My guess, put some upgrade notice to invalidate the url rule cache right before deployment. More safer is to use two deployments. First deploy with `UrlManager`'s `cache` field as `false`, invalidate the cache, and then deploy again with `cache` field as the name of the cache.

PR #10130, opened 2015-11-10
Fix #6780: add findAll without conditions

### LINKED ISSUE #6780: BaseActiveRecord::findAll() without $condition
Sometimes you want to get all records from table (for example if it is a dictionary).

Each model has findAll($condition) method, but you can call it without condition.

I suggest **modify Model::findAll($condition = null) to call it without $condition** if you want. 
Just `Model::findAll();`

For example:
```
$records = Dictionaty::findAll(['field' => 'value']);

// find all records the same way
$records = Dictionary::findAll();
```

For now it looks like:
```
$records = Dictionaty::findAll(['field' => 'value']);

// for now it is a different way
$records = Dictionary::find()->all();
```

pana1990:
#6780

Task : 
- [x] Update [activeRecord guide](https://github.com/yiisoft/yii2/blob/master/docs/guide/db-active-record.md)
- [x] Update change log


samdark:
It's by design.

pana1990:
@samdark Have you seen issue 6780? @cebe said that's  ok https://github.com/yiisoft/yii2/issues/6780#issuecomment-69351727

samdark:
OK. Then I'll let @cebe decide it.

pana1990:
@samdark thanks :)

samdark:
No worries. I've closed it because the issue was raised multiple times during Yii 2.0 beta and the decision was to keep API as we see it now. Next time it's better to reference issues in a pull request to get proper context.

btw., it changes method signature so it's a possible BC break.

pana1990:
@cebe can you check my pr?

it's ready to merge

framework/db/ActiveRecordInterface.php
```diff
@@ -203,6 +204,9 @@ public static function findOne($condition);
      * instances. For example,
      *
      * ```php
+     * // find all the customers
```
pana1990: maybe returns all the customers is better

cebe: It's fine.

cebe:
This change is a breaking change:

> Fatal error: Declaration of MyARClass::findAll() must be compatible with ActiveRecordInterface::findAll($condition = NULL) in /in/7M8Zl on line 16
https://3v4l.org/7M8Zl

so I would prefer to add it to 2.1 instead as it is not a critical change.

pana1990:
@cebe maybe include [yii\db\ActiveRecord::all()] method like laravel or rails.

what do you think about it?

cebe:
how is it different to findAll()?

pana1990:
nevermind. i will wait to 2.1.x version

pana1990:
@cebe can you set this milesto to 2.1.0?

PR #10135, opened 2015-11-11
Update composer-asset-plugin version

bluezed:
The composer-asset-plugin is now on version 1.1.0 and unless there's a reason why Yii requires ~1.0.3 I propose this update to the documentation.


samdark:
Thanks!

PR #10155, opened 2015-11-13
Fixed EmailValidator to check email length properly

### LINKED ISSUE #10142: Validator email Yii2
class User extends \yii\base\Model 
{
...
 public function rules() 
    {
        return [
        ['email', 'email', 'message' => 'Invalid email'],
];
....
}

$model->validate() do not work with email max length. It may be 256 symbols. But when i input more then 256 symbols, it return, then email is valid. 
P.S. Sorry, i am russian =)

SilverFire:
Closes #10142
- [x] PHP & JS
- [x] Tests
- [x] Changelog
- [x] Code review

### DIFF
--- a/framework/validators/EmailValidator.php
+++ b/framework/validators/EmailValidator.php
@@ -70,19 +70,34 @@ class EmailValidator extends Validator
      */
     protected function validateValue($value)
     {
-        // make sure string length is limited to avoid DOS attacks
-        if (!is_string($value) || strlen($value) >= 320) {
+        if (!is_string($value)) {
             $valid = false;
-        } elseif (!preg_match('/^(.*<?)(.*)@(.*?)(>?)$/', $value, $matches)) {
+        } elseif (!preg_match('/^(?P<name>(?:"?([^"]*)"?\s)?)(?:\s+)?(?:(?P<open><?)((?P<local>.+)@(?P<domain>[^>]+))(?P<close>>?))$/i', $value, $matches)) {
             $valid = false;
         } else {
-            $domain = $matches[3];
             if ($this->enableIDN) {
-                $value = $matches[1] . idn_to_ascii($matches[2]) . '@' . idn_to_ascii($domain) . $matches[4];
+                $matches['local'] = idn_to_ascii($matches['local']);
+                $matches['domain'] = idn_to_ascii($matches['domain']);
+                $value = $matches['name'] . $matches['open'] . $matches['local'] . '@' . $matches['domain'] . $matches['close'];
             }
-            $valid = preg_match($this->pattern, $value) || $this->allowName && preg_match($this->fullPattern, $value);
-            if ($valid && $this->checkDNS) {
-                $valid = checkdnsrr($domain, 'MX') || checkdnsrr($domain, 'A');
+
+            if (strlen($matches['local']) > 64 || mb_strlen($matches['name']) > 64) {
+                // The maximum total length of a user name or other local-part is 64 octets. RFC 5322 section 4.5.3.1.1
+                // http://tools.ietf.org/html/rfc5321#section-4.5.3.1.1
+                $valid = false;
+            } elseif (strlen($matches['local'] . '@' . $matches['domain']) > 254) {
+                // There is a restriction in RFC 2821 on the length of an address in MAIL and RCPT commands
+                // of 254 characters. Since addresses that do not fit in those fields are not normally useful, the
+                // upper limit on address lengths should normally be considered to be 254.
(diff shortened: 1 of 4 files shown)


framework/validators/EmailValidator.php
```diff
@@ -70,19 +70,34 @@ public function init()
...
-                $value = $matches[1] . idn_to_ascii($matches[2]) . '@' . idn_to_ascii($domain) . $matches[4];
+                $matches['local'] = idn_to_ascii($matches['local']);
+                $matches['domain'] = idn_to_ascii($matches['domain']);
+                $value = $matches['name'] . $matches['open'] . $matches['local'] . '@' . $matches['domain'] . $matches['close'];
             }
-            $valid = preg_match($this->pattern, $value) || $this->allowName && preg_match($this->fullPattern, $value);
-            if ($valid && $this->checkDNS) {
-                $valid = checkdnsrr($domain, 'MX') || checkdnsrr($domain, 'A');
+
+            if (strlen($matches['local']) > 64 || mb_strlen($matches['name']) > 64) {
```
samdark: The order of checks here looks a bit weird. Also using default encoding for mb_strlen is not a good idea.

SilverFire: First we check the length of a local part of email address, then - the length of name. I don't think it's weird.

Could you suggest an idea for `mb_strlen`?

samdark: OK. I was wrong about that.

samdark:
Thank you!

nkovacs:
Why is the display name (if allowName is true) not allowed to be more than 64 characters?

The linked rfc (http://tools.ietf.org/html/rfc5321#section-4.5.3.1.1) does not say anything about the display name.

The display-name is specified here: http://tools.ietf.org/html/rfc5322#section-3.4, but there's nothing about lengths in there.

samdark:
https://stackoverflow.com/questions/386294/what-is-the-maximum-length-of-a-valid-email-address/574698#574698

nkovacs:
`John Doe <johndoe@example.org>` is not a valid email address at all according to this, which is linked from that answer: http://isemail.info/John%20Doe%20%3Cjohndoe%40example.org%3E

That's because `John Doe <johndoe@example.org>` is the address format used in the message, in the from field, specified in RFC5322. For SMTP (RFC5321), you have to strip the display name, and only use `johndoe@example.org`. At least that's how I understand it. The php mail function probably does this for you automatically.

SilverFire:
> John Doe <johndoe@example.org> is not a valid email address at all according to this, which is linked from that answer: http://isemail.info/John%20Doe%20%3Cjohndoe%40example.org%3E

This test is not designed to validate the name of recipient, as far as I know.


@nkovacs I didn't quite catch your thought.

> Why is the display name (if allowName is true) not allowed to be more than 64 characters?

Do you think that the local part of the email address must not be limited?

nkovacs:
You're confusing the display name with the local part.

In `John Doe <johndoe@example.org>`, `John Doe` is the display name, `johndoe` is the local part. The local part must not be longer than 64 characters. The display name can be any length, because it's not used in SMTP, it's not even allowed in the path.
It's used in the message body itself, which is governed by RFC5322, and that doesn't specify a length for the display name.

nkovacs:
Try this:

```php
<?php

$email = 'your@email.here'

mail('Test this is a really long display name that is longer than 400 characters Lorem ipsum dolor sit amet consectetur adipiscing elit. Donec magna neque imperdiet et orci a fringilla viverra massa. Aenean ullamcorper fringilla odio eu consequat. Donec euismod semper nunc non venenatis. Sed ac placerat nisi. Quisque tempus purus eu nisl vestibulum tempus facilisis velit molestie. Quisque et varius ligula. Sed ullamcorper diam quis fermentum venenatis ante est cras amet <' . $email  . '>', 'Test', 'test message php');
```

It works for me.

SilverFire:
> You're confusing the display name with the local part.

That was a logical typo, sorry :)

I tend to agree, it seems that you're right.

What about http://tools.ietf.org/html/rfc5321#section-4.5.3.1.1? What do they mean in words "total length of a user name or other local-part"? What kind of "other local-part"?

nkovacs:
Yeah, that wording is a bit confusing.
It says "user name or other local part", so to me that means that the local part can be a user name, but it can also be something else (also supported by http://tools.ietf.org/html/rfc5321#section-2.3.11).

SilverFire:
Fixed in 127eeb5
Thank you @nkovacs

PR #10166, opened 2015-11-16
Allow setting css options per CSS file in AssetBundle::$css, fixes #10158

### LINKED ISSUE #10158: Allow setting css options per CSS file in AssetBundle::$css
Situation : I need to register two CSS files in my asset bundle, `screen.css` for media="screen" and `print.css` for media="print". 

As far as I can see, the only way to do this for now is to call $view->registerCssFile() by hand when registering my assets, for my print.css file and pass the options in the second parameter.

I would find it better and simpler if I could specify the options in the AssetBundle::$css property like this:

```php
public $css = [
    'screen.css',
    ['print.css', 'media' => 'print'],
];
```

No BC break would be introduced by accepting an array as a $css entry type..

Any blockers?

machour:
Closes #10158


framework/web/AssetBundle.php
```diff
@@ -147,7 +152,10 @@ public function registerAssetFiles($view)
             $view->registerJsFile($manager->getAssetUrl($this, $js), $this->jsOptions);
         }
         foreach ($this->css as $css) {
-            $view->registerCssFile($manager->getAssetUrl($this, $css), $this->cssOptions);
+            $css = ArrayHelper::toArray($css);
```
SilverFire: Do we really need to use `ArrayHelper::toArray`? I think `$css = (array) $css` will be enough for this purpose.

machour: I'm fine with both. Is there some Yii core guidelines way of doing it?

SilverFire: There is a [core code style guide](https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md), but my notice comes from optimization thoughts and simplicity. When it can be done with PHP type-casting, we have no need to use something else.

SilverFire:
Looks good as for me

framework/web/AssetBundle.php
```diff
@@ -147,7 +152,10 @@ public function registerAssetFiles($view)
             $view->registerJsFile($manager->getAssetUrl($this, $js), $this->jsOptions);
         }
         foreach ($this->css as $css) {
-            $view->registerCssFile($manager->getAssetUrl($this, $css), $this->cssOptions);
+            $css = (array) $css;
```
samdark: I'd better check if `$css` is an array and do merge/shift only in that case.

framework/web/AssetBundle.php
```diff
@@ -170,8 +178,12 @@ public function publish($am)
                 }
             }
             foreach ($this->css as $i => $css) {
-                if (Url::isRelative($css)) {
-                    $this->css[$i] = $converter->convert($css, $this->basePath);
+                $css = (array) $css;
```
samdark: Same as https://github.com/yiisoft/yii2/pull/10166/files#r44996406

samdark:
I'd like to see the same to be possible for JS as well. Having it for CSS only is inconsistent.

tests/framework/web/AssetBundleTest.php
```diff
@@ -271,3 +287,20 @@ class TestAssetCircleB extends AssetBundle
...
+    public $basePath = '@testWebRoot';
+    public $baseUrl = '@testWeb';
+    public $css = [
+        'default_options.css',
+        ['tv.css', 'media' => 'tv'],
+        ['screen_and_print.css', 'media' => 'screen, print']
+    ];
+    public $js = [
+        'normal.js',
+        ['defered.js', 'defer' => 'defer'],
```
machour: @SilverFire @samdark 
How can I get 'defer' to be rendered as an attribute without a value ? 'defer' => null or 'defer' => false doesn't not work.
Is that supported in Yii?

SilverFire: `['defer' => '']`

samdark: `'defer' => true`?

machour: 'defer' => true is the way to to that, thanks for letting me know :)

SilverFire: Will you change it?

machour: Actually I was just asking for my own knowledge :)
I just made the change into the testcase and fixed my CHANGELOG entry.

On Thu, Nov 19, 2015 at 7:48 AM, Dmitry Naumenko <notifications@github.com>
wrote:

> In tests/framework/web/AssetBundleTest.php
> <https://github.com/yiisoft/yii2/pull/10166#discussion_r45307257>:
>
> > @@ -271,3 +287,20 @@ class TestAssetCircleB extends AssetBundle
> >          'yiiunit\\framework\\web\\TestAssetCircleA'
> >      ];
> >  }
> > +
> > +class TestAssetPerFileOptions extends AssetBundle
> > +{
> > +    public $basePath = '@testWebRoot';
> > +    public $baseUrl = '@testWeb';
> > +    public $css = [
> > +        'default_options.css',
> > +        ['tv.css', 'media' => 'tv'],
> > +        ['screen_and_print.css', 'media' => 'screen, print']
> > +    ];
> > +    public $js = [
> > +        'normal.js',
> > +        ['defered.js', 'defer' => 'defer'],
>
> Will you change it?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/yiisoft/yii2/pull/10166/files#r45307257>.
>



-- 
Mehdi Achour

SilverFire:
Thank you!

PR #10169, opened 2015-11-16
Normalized mb_* function calls to use Yii::$app->charset

SilverFire:
There are calls of mb_\* functions without specified charset or with wrong charset (`utf-8`, 'utf8').

This PR normalizes mb_\* functions calls to use `UTF-8` or `Yii::$app->charset`.
- [x] code
- [x] tests
- [x] changelog
- [x] code review

### DIFF
--- a/framework/i18n/MessageFormatter.php
+++ b/framework/i18n/MessageFormatter.php
@@ -7,6 +7,7 @@
 
 namespace yii\i18n;
 
+use Yii;
 use yii\base\Component;
 use yii\base\NotSupportedException;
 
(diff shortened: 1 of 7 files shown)


samdark:
Looks good to me.

samdark:
Changelog line could say it's a bug since it is what it is.

framework/i18n/MessageFormatter.php
```diff
@@ -274,19 +275,20 @@ protected function fallbackFormat($pattern, $args, $locale)
      */
     private static function tokenizePattern($pattern)
     {
+        $charset = Yii::$app ? Yii::$app->charset : 'UTF-8';
```
klimov-paul: I am unsure about this: i18n feature requires usage of 'UTF-8' encoding no matter of application encoding.

klimov-paul: Although, since by default `mb_strpos()` uses `mb_internal_encoding()`. Perhaps it is fine.

framework/web/ErrorHandler.php
```diff
@@ -391,8 +391,8 @@ public function argumentsToString($args)
                 $args[$key] = '<span class="keyword">' . ($value ? 'true' : 'false') . '</span>';
             } elseif (is_string($value)) {
                 $fullValue = $this->htmlEncode($value);
-                if (mb_strlen($value, 'utf8') > 32) {
-                    $displayValue = $this->htmlEncode(mb_substr($value, 0, 32, 'utf8')) . '...';
+                if (mb_strlen($value, Yii::$app->charset) > 32) {
+                    $displayValue = $this->htmlEncode(mb_substr($value, 0, 32, Yii::$app->charset)) . '...';
```
cebe: The error handler should not rely on an application to be existing. I think in this case it is fine to assume utf8

samdark: There's no `utf8`, only `UTF-8`.

SilverFire: @cebe it makes sense. Maybe it's better to leave `UTF-8` hardcoded, as the ErrorHandler may be fired before the application creating.

SilverFire: Do we need to hardcode UTF-8 as well on [line 165](https://github.com/yiisoft/yii2/blob/master/framework/web/ErrorHandler.php#L165)? It was there since @cebe commit https://github.com/yiisoft/yii2/commit/d4ad533d48773f90f3b89c550860a65986a10eca#diff-7a2fbae94c890a0d999b38f8a82fe81eR141

cebe: > Do we need to hardcode UTF-8 as well on line 165?

yes, should be done.

SilverFire: ac7b0ef

framework/i18n/MessageFormatter.php
cebe: I thinks this is fine, if you choose another encoding for whatever reason, all strings should have the same encoding.

PR #10186, opened 2015-11-18
Use native hash_equals in Security::compareString

aotd1:
Using native timing attack safe function http://php.net/manual/ru/function.hash-equals.php

### DIFF
--- a/framework/base/Security.php
+++ b/framework/base/Security.php
@@ -648,6 +648,9 @@ class Security extends Component
      */
     public function compareString($expected, $actual)
     {
+        if (function_exists('hash_equals')) {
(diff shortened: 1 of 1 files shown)


SilverFire:
PHPUnit tests failed :(

samdark:
Won't accept it as is considering that it fails tests. Current solutuion was tested and verified.

TimWolla:
@samdark The test fails, because `hash_equals` requires the parameters to be strings, while the test input are non-strings. As this is a security critical function non dynamic type casting should happen anyway, like it currently does.

This pull request is correct and I ask you to re-consider.

Issue #10171, opened 2015-11-17
Segmentation fault in Yii2.0.7-dev (works in 2.0.6)


it3rmit:
I am getting segmentation fault on CentOS using yii2 v2.0.7-dev. This code used to work recently. 
It still works on my localhost on Ubuntu 14.04
My Centos Version
```
cat /etc/centos-release
CentOS release 6.6 (Final)
```
After I found the issue #10152  I have change ```composer.json```:
```
"yiisoft/yii2": "*"
```
to 
```
"yiisoft/yii2": "2.0.6"
```
Everything has started to work again.
I am getting a segfault on save() method of ActiveRecord instance.
The info about my environment
```
php -v
```
```
PHP 5.5.23 (cli) (built: Mar 21 2015 22:04:59) 
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2015 Zend Technologies
    with Zend OPcache v7.0.4-dev, Copyright (c) 1999-2015, by Zend Technologies
    with Xdebug v2.2.6, Copyright (c) 2002-2014, by Derick Rethans

```
```
php -m
```
```
[PHP Modules]
bz2
calendar
Core
ctype
curl
date
dom
ereg
exif
fileinfo
filter
ftp
gd
gettext
gmp
hash
iconv
intl
json
libxml
mbstring
mcrypt
memcache
mhash
mysql
mysqli
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
readline
Reflection
session
shmop
SimpleXML
sockets
SPL
sqlite3
standard
tokenizer
wddx
xdebug
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Xdebug
Zend OPcache

```
```
apachectl -V
```
```
Server version: Apache/2.2.15 (Unix)
Server built:   Oct 16 2014 14:48:21
Server's Module Magic Number: 20051115:25
Server loaded:  APR 1.3.9, APR-Util 1.3.9
Compiled using: APR 1.3.9, APR-Util 1.3.9
Architecture:   64-bit
Server MPM:     Prefork
  threaded:     no
    forked:     yes (variable process count)
Server compiled with....
 -D APACHE_MPM_DIR="server/mpm/prefork"
 -D APR_HAS_SENDFILE
 -D APR_HAS_MMAP
[...]

PR #10192, opened 2015-11-19
Added requirement for ICU Data

SilverFire:
As a result of segfaults #10152, #10171

### DIFF
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -17,6 +17,17 @@ php composer.phar global require "fxp/composer-asset-plugin:~1.1.0"
 Upgrade from Yii 2.0.6
 ----------------------
 
+* Added new requirement: ICU Data version >= 49.1. Please, ensure that your environment has ICU data installed and
+up to date to prevent unexpected behavior or crashes.
+
+ > Tip: The easiet way to check: `php --info | grep ICU`
+
+ > Normal output example (version shoud be >= 49.1):
+   ```
+   $> php --info | grep ICU
+   ICU version => 55.1
+   ICU Data version => 55.1
+   ```
 * The signature of `yii\helpers\BaseInflector::transliterate()` was changed. The method is now public and has an
 extra optional parameter `$transliterator`.
 * In `yii\web\UrlRule` the `pattern` matching group names are being replaced with the placeholders on class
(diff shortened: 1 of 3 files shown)


framework/requirements/requirements.php
```diff
@@ -73,6 +73,17 @@
         INTL_ICU_VERSION . '.'
     ),
     array(
+        'name' => 'ICU Data version',
+        'mandatory' => false,
```
SilverFire: I think we can switch `mandatory` to `true` in both ICU and ICU Data requirements

framework/UPGRADE.md
```diff
@@ -17,6 +17,17 @@ php composer.phar global require "fxp/composer-asset-plugin:~1.1.0"
 Upgrade from Yii 2.0.6
 ----------------------
 
+* Added new requirement: ICU Data version >= 49.1. Please, ensure that your environment has ICU data installed and
+up to date to prevent unexpected behavior or crashes.
+
+ > Tip: The easiet way to check: `php --info | grep ICU`
```
samdark: It's not:

1. We have requirements script.
2. The common mistake is to use different `php.ini` in CLI / fpm / apache module.

framework/requirements/requirements.php
samdark: I don't think so. If you're developing English-only website you're fine w/o intl at all.

framework/requirements/requirements.php
```diff
@@ -73,6 +73,17 @@
...
+        'condition' => defined('INTL_ICU_DATA_VERSION') && version_compare(INTL_ICU_DATA_VERSION, '49.1', '>='),
+        'by' => '<a href="http://www.php.net/manual/en/book.intl.php">Internationalization</a> support',
+        'memo' => 'ICU Data 49.1 or higher is required when you want to use <code>#</code> placeholder in plural rules
+        (for example, plural in
+        <a href=\"http://www.yiiframework.com/doc-2.0/yii-i18n-formatter.html#asRelativeTime%28%29-detail\">
+        Formatter::asRelativeTime()</a>) in the <code>yii\i18n\Formatter</code> class. Your current ICU Data version is ' .
+        INTL_ICU_DATA_VERSION . '.'
```
cebe: defined check needed for this too.

framework/requirements/requirements.php
```diff
@@ -73,6 +73,17 @@
         INTL_ICU_VERSION . '.'
```
cebe: same here btw.

SilverFire:
https://github.com/yiisoft/yii2/commit/9d79038485ea4ca79a6137b3d09bb271bc97c9d4

SilverFire:
Thank you!

samdark:
:+1:

PR #10199, opened 2015-11-20
Added tests for dropTable, truncateTable and renameTable

edgardmessias:


tests/framework/db/CommandTest.php
```diff
@@ -347,24 +347,51 @@ public function testAlterTable()
...
     }
 
     public function testRenameTable()
     {
+        $db = $this->getConnection();
+
+        $fromTableName = 'type';
+        $toTableName = 'new_type';
+
+        $db->createCommand('DROP TABLE IF EXISTS {{' . $toTableName . '}};')->execute();
```
samdark: There's no `IF EXISTS` in DBs other than MySQL.

samdark: Better to do it via checking schema and doing drop table via API.

tests/framework/db/mssql/MssqlCommandTest.php
```diff
@@ -127,4 +127,27 @@ public function testAlterTable()
...
+
+    public function testRenameTable()
+    {
+        $db = $this->getConnection();
+
+        $fromTableName = 'type';
+        $toTableName = 'new_type';
+
+        // on the first run the 'new_type' table does not exist
+        try {
```
samdark: This approach would work in MySQL well. Let's use it instead of duplicating tests.

edgardmessias:
Fixed @samdark

samdark:
https://github.com/yiisoft/yii2/commit/7c9b7c77a7800ccd9d836f21c542b9828c2aa52c

samdark:
Thanks!

PR #10200, opened 2015-11-20
Controller->viewPath can be set, fixes #10086

### LINKED ISSUE #10086: The view file not found when a controller is mapped
Since there is no corresponding file under mapped-to controller' view folder, application throws exception. When a controller is mapped, the views must be tried to find under mapped-from controller' view folder, not mapped-to controller.

Sibilino:
Added setViewPath() that allows setting the base path for the
controller's views. Changed getViewPath() to return the value that was
explicitly set, before defaulting to the usual module-based path.

Thus, #10086 can be fixed using:

``` php
public $controllerMap = [
    'account' => [
        'class' => 'app\controllers\UserController',
        'viewPath' => '/home/[user]/public_html/yii_example/views/user'
    ],
    // ...
]
```

### DIFF
--- a/framework/base/Controller.php
+++ b/framework/base/Controller.php
@@ -19,8 +19,7 @@ use Yii;
  * @property string $uniqueId The controller ID that is prefixed with the module ID (if any). This property is
  * read-only.
  * @property View|\yii\web\View $view The view object that can be used to render views or view files.
- * @property string $viewPath The directory containing the view files for this controller. This property is
- * read-only.
+ * @property string $viewPath The directory containing the view files for this controller.
  *
  * @author Qiang Xue <qiang.xue@gmail.com>
  * @since 2.0
@@ -67,6 +66,10 @@ class Controller extends Component implements ViewContextInterface
      * @var View the view object that can be used to render views or view files.
      */
     private $_view;
+    /**
+     * @var string the root directory that contains view files for this controller.
+     */
+    private $_viewPath;
 
 
     /**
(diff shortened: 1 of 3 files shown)


framework/base/Controller.php
```diff
@@ -67,6 +66,10 @@ class Controller extends Component implements ViewContextInterface
      * @var View the view object that can be used to render views or view files.
      */
     private $_view;
+    /**
+     * @var string the root directory that contains view files for this controller.
+     */
+    private $_viewPath;
```
SilverFire: `@since 2.0.7`

framework/base/Controller.php
```diff
@@ -449,7 +452,21 @@ public function setView($view)
      */
     public function getViewPath()
     {
-        return $this->module->getViewPath() . DIRECTORY_SEPARATOR . $this->id;
+        if ($this->_viewPath !== null) {
+            return $this->_viewPath;
+        } else {
+            return $this->_viewPath = $this->module->getViewPath() . DIRECTORY_SEPARATOR . $this->id;
+        }
```
SilverFire: It's better to rewrite it to:

```php
if (!$this->_viewPath === null) {
    $this->_viewPath = $this->module->getViewPath() . DIRECTORY_SEPARATOR . $this->id;
}

return $this->_viewPath;
```

framework/base/Controller.php
cebe: `@since` is not really useful on private properties as they do not belong to the public api, it does not matter when they were introduced.

framework/base/Controller.php
Sibilino: I agree, but I copied the code block from `yii\base\Module` to make sure I was keeping consistent style. Should I change?

framework/base/Controller.php
Sibilino: Added @since 2.0.7 to setViewPath()

framework/base/Controller.php
cebe: module should be adjusted to this style too.

samdark: Yes.

SilverFire:
Merged, thank you! https://github.com/yiisoft/yii2/commit/d08ef83e2ad2d9b9813ec558d1adabfe7a90fc28

framework/base/Controller.php
SilverFire: Fixed https://github.com/yiisoft/yii2/commit/2d68840898ce56457d45c8b27ba739d630461af5

SilverFire:
@coveralls oh really? :D

PR #10214, opened 2015-11-21
fix issues in guide-zh-CN

uwang:


samdark:
Why is it closed?

hobartwang:
Network problem.It works in other request.

samdark:
OK.

PR #10220, opened 2015-11-22
add runtime-logging.md to guide-zh-CN

uwang:
copy english version into guide-zh-CN


samdark:
This doesn't make sense.

Issue #10217, opened 2015-11-21
GridView widget problem with array query parameters


omnilight:
I want to use filter in grid view that allows me to select multiple values (for example `<select multiple=multiple>`). But with current implementation of the `yii.gridView::applyFilter`  Js method I can't do this, because, it does not support mulpile parameters with the same name here:
https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.gridView.js#L89

I think it would be greate to rewrite this part of code to support multiple parameters with the same name

SilverFire:
@omnilight yeah, it's a problem. Can you make a PR?

omnilight:
#10223

SilverFire:
https://github.com/yiisoft/yii2/commit/d3cfb87ba67a328d197d1a9e52dbb12a7bf8476b

omnilight:
I've found small, but important bug.

I the case when you have standard `Html::activeListView` as a filter input, it really adds two inputs: one is select with the name `name[]` and second is hidden input with name `name`. Second one is passed, when no value is selected.

In the case we have selected some value, query string will be: `name=&name[]=1&name[]=2`. That's correct. Now, if you deselect all values from select, `applyFilter` function will find only hidden's input value (because of `$(settings.filterSelector).serializeArray()`). So, when later it will try to override query string values, it will not override `&name[]=1&name[]=2` part because no key `name[]` will be in the `keysInFilter` variable. So my new pull-request is fixing with behavior

PR #10223, opened 2015-11-22
#10217 Support for multiple params with the same name

omnilight:
Fixes for #10217

### DIFF
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -99,7 +99,7 @@
                     if (data[name] === undefined) {
                         data[this.name] = [];
                     }
-                    data[name] = value;
+                    data[name].push(value);
                 }
             });
 
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -49,7 +49,7 @@
          */
         afterFilter: 'afterFilter'
     };
-    
+
     var methods = {
         init: function (options) {
             return this.each(function () {
@@ -85,12 +85,20 @@
             var $grid = $(this), event;
             var settings = gridData[$grid.attr('id')].settings;
             var data = {};
+            var keysInFilter = [];
             $.each($(settings.filterSelector).serializeArray(), function () {
-                data[this.name] = this.value;
+                if (data[this.name] === undefined) {
+                    data[this.name] = [];
+                    keysInFilter.push(this.name);
+                }
+                data[this.name].push(this.value);
             });
 
             $.each(yii.getQueryParams(settings.filterUrl), function (name, value) {
-                if (data[name] === undefined) {
+                if (keysInFilter.indexOf(name) < 0) {
+                    if (data[name] === undefined) {
+                        data[this.name] = [];
+                    }
                     data[name] = value;
                 }
             });
@@ -100,10 +108,12 @@
 
             $grid.find('form.gridview-filter-form').remove();
             var $form = $('<form action="' + url + '" method="get" class="gridview-filter-form" style="display:none" data-pjax></form>').appendTo($grid);
-            $.each(data, function (name, value) {
-                $form.append($('<input type="hidden" name="t" value="" />').attr('name', name).val(value));
+            $.each(data, function (name, values) {
+                $.each(values, function (index, value) {
+                    $form.append($('<input type="hidden" name="t" value="" />').attr('name', name).val(value));
+                });
             });
-            
+
             event = $.Event(gridEvents.beforeFilter);
             $grid.trigger(event);
             if (event.result === false) {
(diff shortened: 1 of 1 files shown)


framework/assets/yii.gridView.js
```diff
@@ -100,10 +108,12 @@
 
             $grid.find('form.gridview-filter-form').remove();
             var $form = $('<form action="' + url + '" method="get" class="gridview-filter-form" style="display:none" data-pjax></form>').appendTo($grid);
-            $.each(data, function (name, value) {
-                $form.append($('<input type="hidden" name="t" value="" />').attr('name', name).val(value));
+            $.each(data, function (name, values) {
+                $.each(values, function (index, value) {
+                    $form.append($('<input type="hidden" name="t" value="" />').attr('name', name).val(value));
```
SilverFire: Did you test it? If you submit `<input name="test">` more than once, you will get the last `input` value.

omnilight: Well, yes, I'll check it better...

omnilight: Ok, I've tested.

http://plnkr.co/edit/0L2tBgNhy8ZbNUuif119

All form elements will be sent together.

SilverFire: They will be sent, but PHP will use only the last one because `$_POST` is a key-value associative array.

When there are more than 1 selected option, you need to submit `name="login[]"` instead of `name="login"`.

omnilight: That's true. But this change is only required when you use some kind of input, that allows you to pass multiple values into it. For example you can use select with `multiple` set to true.
Of couse setting correct name for input is developer's task, but if you use for example `Html::listBox` method it will add `[]` to the name automatically, same think will happen if you use some widget based on the list box (for example one of select2 widgets). If you use checkboxes in the filter, then `[]` will be added too.

omnilight:
Any news on this pull request? It is required for me to implement filters in my app

SilverFire:
@omnilight I'll verify it ASAP

SilverFire:
The patch produces JS errors:

Grid column definition:

```php
[
    'attribute' => 'state[]'
    'filter' => function ($column, $model, $attribute) {
        return Html::activeDropDownList($model, $attribute, ['ok' => 'ok', 'incoming' => 'incoming', 'other' => 'other'], ['multiple' => true, 'class' => 'form-control']);
     },
     'value' => 'state'
],
```

Chrome, JS debug:
screenshot from 2015-11-25 00-13-10

Could you please test it once more, fix bugs and ensure that all works really good?

omnilight:
Fixed variable assignment

SilverFire:
screenshot from 2015-11-26 21-31-47

framework/assets/yii.gridView.js
```diff
@@ -85,13 +85,21 @@
             var $grid = $(this), event;
             var settings = gridData[$grid.attr('id')].settings;
             var data = {};
+            var keysInFilter = [];
             $.each($(settings.filterSelector).serializeArray(), function () {
-                data[this.name] = this.value;
+                if (data[this.name] === undefined) {
+                    data[this.name] = [];
+                    keysInFilter.push(this.name);
```
SilverFire: It leads to duplication of the same key in array more than once

framework/assets/yii.gridView.js
```diff
@@ -85,13 +85,21 @@
...
             $.each(yii.getQueryParams(settings.filterUrl), function (name, value) {
-                if (data[name] === undefined) {
-                    data[name] = value;
+                if (keysInFilter.indexOf(name) < 0) {
+                    if (data[name] === undefined) {
+                        data[this.name] = [];
+                    }
+                    data[name].push(value);
                 }
             });
```
SilverFire: All above may be simplified to: 

```js
            var data = {};
            $.each($(settings.filterSelector).serializeArray(), function () {
                if (!(this.name in data)) {
                    data[this.name] = [];
                }
                data[this.name].push(this.value);
            });

            var namesInFilter = Object.keys(data);

            $.each(yii.getQueryParams(settings.filterUrl), function (name, value) {
                if (namesInFilter.indexOf(name) === -1) {                                                                                                                                                                                      
                    if (!(name in data)) {
                        data[name] = [];
                    }
                    data[name].push(value);
                }
            });
```

framework/assets/yii.gridView.js
SilverFire: I'd prefer to rewrite this line to:

```js
$form.append($('<input>').attr({'type': 'hidden', 'name': name, 'value': value}));
```

SilverFire:
Merged with adjustments https://github.com/yiisoft/yii2/commit/d3cfb87ba67a328d197d1a9e52dbb12a7bf8476b

SilverFire:
Thank you!

omnilight:
Greate! Thank you!

PR #10225, opened 2015-11-23
Returns object from methods for chaining purpose

arturf:


samdark:
I'm not sure it's a good idea.

samdark:

klimov-paul:
Method `setAttributes()` produces a virtual property `attributes`, so usually its return value is never used.
Still we can return a self reference as a result. However, there are many similar 'setters' around the framework. There should be a consistency among them.

SilverFire:
Agree with @klimov-paul

samdark:
Since some setters [return true/false depending on the result](https://github.com/yiisoft/yii2/blob/master/framework/caching/Cache.php#L188) I don't think it's a good idea to return object instance for other cases.

PR #10235, opened 2015-11-24
Fixed `yii\console\Application::runAction` to not to corrupt response

hiqsol:
… exit code when action returns Response object

Original code forced casting to integer which resulted in changing Response object with exitStatus = 0 to integer 1.


cebe:
Merged. Thank you!

PR #10241, opened 2015-11-24
Fix phpdocs and small bug

pana1990:
Fix small bug, fields need be array by default beacuse yii parse string to array only if option is array by default (https://github.com/yiisoft/yii2/blob/master/framework/console/Controller.php#L84).

I fix test too, sorry for my mistake

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -45,10 +45,10 @@ abstract class BaseMigrateController extends Controller
      * @var array a set of template paths for generating migration code automatically.
      *
      * The key is the template type, the value is a path or the alias. Supported types are:
-     * - `create`: table creating template
-     * - `drop`: table dropping template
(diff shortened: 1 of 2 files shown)


SilverFire:
Merged. Thank you!

pana1990:
@SilverFire why do you  know test fail with hhvm? i dont understand.

SilverFire:
Because of 1-second lag and missing `body` https://travis-ci.org/yiisoft/yii2/jobs/93006055#L220

When I checked the HHVM test fail I didn't see the missing `body` option. I'm not sure why this happens in HHVM only :thought_balloon:

pana1990:
It is a strange behavior. i will check with hhvm in local.

pana1990:
@SilverFire  i think regular expression was wrong. check my pr https://github.com/yiisoft/yii2/pull/10248

SilverFire:
Thank you, merged

PR #10244, opened 2015-11-25
Fixes #8329: Added support of options for `message` console command

### LINKED ISSUE #8329: Enh: extract messages for specified language

samdark:
Replaces https://github.com/yiisoft/yii2/pull/8739


samdark:
@yiisoft/core-developers need review. Especially:

- `framework/console/Controller.php`.
- `framework/console/controllers/MessageController.php`.

framework/console/controllers/MessageController.php
```diff
@@ -78,22 +245,25 @@ public function actionConfig($filePath)
      * this file and then customize it for your needs.
      * @throws Exception on failure.
      */
-    public function actionExtract($configFile)
+    public function actionExtract($configFile = null)
```
SilverFire: Signature changed. Should be mentioned in UPGRADE.md

SilverFire: Also we need to update to the PHPDoc and describe behavior in case when `$configFile` is null

SilverFire:
Looks great.
I'm not sure whether we need to add `@since` everywhere :)

framework/console/controllers/MessageController.php
samdark: No need to do it since console output would tell you what's wrong in this case.

PR #10248, opened 2015-11-25
Fix test for migration

pana1990:
@SilverFire it pass all tests now, the error was in regular expression and olds migrations (now i call tearDownMigrationPath in parseName method to avoid this one).


SilverFire:
Merged! Thank you!

PR #10249, opened 2015-11-25
Fix methods references

arturf:
And one small question.  
`encryptByKey` returns value that not ready to store in db. I use base64 encode/decode, but maybe exist some other way? Thanks.


SilverFire:
Thank you!

Issue #8743, opened 2015-06-10
defaultScope, relation alias


JamesBarnsley:
Hi,

I am having to use relationship alias when referencing my join because I have two foreign keys that reference the same table.

The problem is this breaks my default scope ...

```php
public static function find() {

	return parent::find()->where([static::tableName() . ".deleted" => 0]);
							
}
```

The reason why it breaks is because it is looking for tableName() but the relationship is needing to use the alias so the tableName() will not work.

cebe:
can you show the query you are trying to run? also include the names of the models so it is clear which is realted to which one.

JamesBarnsley:
Hi,

Here is the query, I have a PrivateMessageTopic model which is related to the User model.

As you can see in the joinWith() part of the query, I have used the "from" to get the second relation as "userTo".

```php
$query = PrivateMessageTopic::find();
			
$query->joinWith(["user", "userTo" => function($q) { $q->from("user userTo"); }]);
						
$query->orWhere(["fkIDWithUserID" => Yii::$app->user->id])->orWhere(["fkIDWithUserIDTo" => Yii::$app->user->id]);
							
$activeDataProvider = new ActiveDataProvider([
				
	"query" => $query,
	"pagination" => [
					
		"pageSize" => 20
					
	]
				
]);
```
But this causes problems because the find() is looking for the tableName()

cebe:
closing this as duplicate of #7263

Issue #7263, opened 2015-02-13
Disambiguing column's table on ActiveQuery select/addSelect


petrabarus:
I have several tables that have columns with same name. Let's say `shelf`, `book`, `chapter`, have the same field `name`.

I have a page that will show list of `chapter` that will also show the name of the `book` as well as the `shelf`.

Naturally this is the query I used for the `ActiveDataProvider`

```php
Chapter::find()
  ->joinWith([
      'book' => function($query) {
          $query->joinWith([
              'shelf'
          ])
      }
  ])
```

But I only wanted to show the name in the `GridView`, so this is what I do to avoid the `SELECT *`.

```php
Chapter::find()
  ->addSelect(['name', 'bookId'])
  ->joinWith([
          'book' => function($query) {
              $query->addSelect(['name', 'shelfId'])
                         ->joinWith([
                                'shelf' => function($query){
                                    $query->addSelect(['name']);
                                }
                         ]);
          }
  ])
```

On my Yii2 (c21895d4dd4c18d5bab37e0b7b359668f8aa8b67) this will output error `Column 'name' in field list is ambiguous`. So I have to put something like this

```php
Chapter::find()
  ->addSelect(['Chapters.name', 'bookId'])
  ->joinWith([
          'book' => function($query) {
              $query->addSelect(['Books.name', 'shelfId'])
                         ->joinWith([
                                'shelf' => function($query){
                                    $query->addSelect(['Shelves.name']);
                                }
                         ]);
          }
  ])
```

Do I really have to do this for every query like this? Or is there any simpler way I don't know?

I'm thinking that if I can disambiguate the table name right from the `addSelect` method, it would be much easier. I extend the `ActiveQuery` and do something like this.

```php
    private $_tableName;
    private function getTableName() {
        //since the `from` will be resolved in the `prepare` method.
        if (!isset($this->_tableName)) {
            $class = $this->modelClass;
            $this->_tableName = $class::tableName();
        }
        return $this->_tableName;
    }
    public function addSelect($columns) {
        if (is_array($columns)) {
            $columns = array_map(function($column) {
                return (strpos($column, ".") == false) ? $this->getTableName() . ".{$column}" : $column;
            }, $columns);
        } else {
            $columns = (strpos($column, ".") == false) ? $this->getTableName() . ".{$columns}" : $columns;
        }
        return parent::addSelect($columns);
    }
```

mdmunir:
mungkin sama dengan ini #2377

[...]

Issue #5137, opened 2014-09-23
Using a joinwith to two different relations on the same table causes alias error


Sammaye:
I have an order object with two relations like so:

``` php
    public function getUser()
    {
        return $this->hasOne('common\models\User', ['id' => 'user_id']);
    }

    public function getPartner()
    {
        return $this->hasOne('common\models\User', ['ns_partner_id' => 'partner_id']);
    }
```

when I do:

```
<?php foreach($query->joinWith(['partner', 'user'])->each() as $order){ ?>
```

I get:

```
SQLSTATE[42000]: Syntax error or access violation: 1066 Not unique table/alias: 'tbl_user'
```

Is this expected?

CedricYii:
With the new alias possibilities introduced in yii2.0.7, you could use the following simple syntax:
````
<?php 
foreach( $query->joinWith(['partner p', 'user u'])->each() as $order ) {
...
}
?>
````
This should work well with the simple relations described in your first post.

(But maybe you are expected Yii to detect this situation, and handle it automatically for you? Which may be done by assigning the relation's name as default alias instead of keeping tableName.)

cebe:
related to #7263

leandrogehlen:
@CedricYii, really this works, but i think that query must works without to force explict alias.
It's necessary,by default, join always use relation name as alias

bizley:
Closing due to inactivity. If this should be reopen please write here.

PR #10253, opened 2015-11-25
Advanced table alias handling for Query and ActiveQuery

### LINKED ISSUE #2377: Table alias for joinWith needed

### LINKED ISSUE #9326: ActiveQuery getAlias
I am looking for a function which returns the alias for a 'from' in ActiveQuery. It doesnt seem available.
Looking forward into ActiveQuery, it seems that some refactoring is still left to do:
* https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveQuery.php#L161
* https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveQuery.php#L541

Is this something on the roadmap or should someone do a PR?

cebe:
Allows to disambiguate colums without knowing the alias that has been
defined for a table elsewhere (e.g. in a relation definition).
- add methods: getAlias(), applyAlias() to Query for retrieving the
  alias for a table or disambiguate a column name.
- add methods: getRelationAlias() and applyRelationAlias() for
  retrieving the alias for a joined relation table and to disambiguate
  column names of these.
- Add alias syntax to joinWith(), e.g. joinWith('author a').
  No need to know the table name for defining an alias for the relation.

includes tests, docs are following.

Addresses the following issues:
- fixes #2377 (finally) :)
- fixes #9326
- fixes alias issues with default scope: #8743
- introduces new syntax for table alias resolution: `{{@user}}` similar to table prefix and table name escape syntax.
- picked test from https://github.com/yiisoft/yii2/pull/8788, thanks @nainoon

Open todos:
- [ ] add more tests
- [ ] complete documentation (cover #5137, adjust [gridview relation guide](https://github.com/yiisoft/yii2/blob/master/docs/guide/output-data-widgets.md#working-with-model-relations))
- [ ] make getAlias() work for tables from relations
- [x] add alias() method to ActiveQuery to fix #4972
- [ ] check how this relates to #7263

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -403,9 +403,9 @@ class ActiveQuery extends Query implements ActiveQueryInterface
                 $callback = null;
             }
 
-            if (preg_match('/^(.*?)(\s+AS\s+|\s+)(\w+)$/i', $name, $matches)) {
+            if (preg_match('/^(.*?)(?:\s+AS\s+|\s+)(\w+)$/i', $name, $matches)) {
                 // relation is defined with an alias, adjust callback to apply alias
-                list(, $relation, , $alias) = $matches;
+                list(, $relation, $alias) = $matches;
                 $this->_relationAliases[$relation] = $alias;
                 $name = $relation;
                 $callback = function($query) use ($callback, $alias) {
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -148,7 +148,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             $this->joinWith = null;    // clean it up to avoid issue https://github.com/yiisoft/yii2/issues/2687
         }
 
-        if (empty($this->from)) { // TODO provide alias() method for convenience
+        if (empty($this->from)) {
             /* @var $modelClass ActiveRecord */
             $modelClass = $this->modelClass;
             $tableName = $modelClass::tableName();
@@ -768,6 +768,37 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             call_user_func($callable, $relation);
         }
 
+        return $this;
+    }
+
+    /**
+     * Define an alias for the table defined in [[modelClass]].
+     *
+     * This method will adjust [[from]] so that an already defined alias will be overwritten.
+     * If none was defined, [[from]] will be populated with the given alias.
+     *
+     * @param string $alias the table alias.
+     * @return $this the query object itself
+     * @since 2.0.7
+     */
+    public function alias($alias)
+    {
+        if (empty($this->from) || count($this->from) < 2) {
+            list($tableName, ) = $this->getQueryTableName($this);
+            $this->from = [$alias => $tableName];
+        } else {
+            /* @var $modelClass ActiveRecord */
+            $modelClass = $this->modelClass;
+            $tableName = $modelClass::tableName();
+
+            foreach($this->from as $key => $table) {
+                if ($table === $tableName) {
+                    unset($this->from[$key]);
+                    $this->from[$alias] = $tableName;
+                }
+            }
+        }
+        $this->populateAliases($this->from);
         return $this;
     }
 
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -352,7 +352,8 @@ class ActiveQuery extends Query implements ActiveQueryInterface
      * which is equivalent to calling [[with()]] using the specified relations.
      *
      * Note that because a JOIN query will be performed, you are responsible to disambiguate column names.
-     * You may specify the table names manually or use the methods [[getRelationAlias()]] or [[applyRelationAlias()]] for this.
+     * You may specify the table names manually or use the methods described in the [Guide about alias handling](db-querybuilder#aliases)
+     * for this.
      *
      * This method differs from [[with()]] in that it will build up and execute a JOIN SQL statement
      * for the primary table. And when `$eagerLoading` is true, it will call [[with()]] in addition with the specified relations.
@@ -388,7 +389,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
      *
      * @param boolean|array $eagerLoading whether to eager load the relations specified in `$with`.
      * When this is a boolean, it applies to all relations specified in `$with`. Use an array
-     * to explicitly list which relations in `$with` need to be eagerly loaded.
+     * to explicitly list which relations in `$with` need to be eagerly loaded. Defaults to `true`.
      * @param string|array $joinType the join type of the relations specified in `$with`.
      * When this is a string, it applies to all relations specified in `$with`. Use an array
      * in the format of `relationName => joinType` to specify different join types for different relations.
@@ -403,9 +404,9 @@ class ActiveQuery extends Query implements ActiveQueryInterface
                 $callback = null;
             }
 
-            if (preg_match('/^(.*?)(?:\s+AS\s+|\s+)(\w+)$/i', $name, $matches)) {
+            if (preg_match('/^.*?(?:\s+AS\s+|\s+)(\w+)$/i', $name, $matches)) {
                 // relation is defined with an alias, adjust callback to apply alias
-                list(, $relation, $alias) = $matches;
+                list($relation, $alias) = $matches;
                 $this->_relationAliases[$relation] = $alias;
                 $name = $relation;
                 $callback = function($query) use ($callback, $alias) {
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -8,6 +8,7 @@
 namespace yii\db;
 
 use yii\base\InvalidConfigException;
+use yii\base\InvalidParamException;
 
 /**
  * ActiveQuery represents a DB query associated with an Active Record class.
@@ -147,7 +148,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             $this->joinWith = null;    // clean it up to avoid issue https://github.com/yiisoft/yii2/issues/2687
         }
 
-        if (empty($this->from)) {
+        if (empty($this->from)) { // TODO provide alias() method for convenience
             /* @var $modelClass ActiveRecord */
             $modelClass = $this->modelClass;
             $tableName = $modelClass::tableName();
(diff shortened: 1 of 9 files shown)


Faryshta:
which would be the proposed syntaxis?

cebe:
```php
 // find all orders that contain books of the category 'Science fiction', using the alias "b" for the books table
 Order::find()->joinWith(['books b'], true, 'INNER JOIN')
              ->where(['b.category' => 'Science fiction'])
              ->all();
```

cebe:
And you can also do this:

```php
$query = Order::find()->joinWith(['books b'], true, 'INNER JOIN');
// somewhere later in the code:
$query->andWhere([$query->applyRelationAlias('books', 'category') => 'Science fiction']);
// which will use the alias defined above.
```

Same works for aliases set via from(), join(), etc....
`{get|apply}RelationAlias()` take relation names, 
`{get|apply}Alias()` take table names.

Faryshta:
> Order::find()->joinWith(['books b'], true, 'INNER JOIN')
>              ->where(['b.category' => 'Science fiction'])
>              ->all();

I still think

```php
Order::find()->joinWith(['b' => 'books'], true, 'INNER JOIN')
              ->where(['b.category' => 'Science fiction'])
              ->all();
```

is more related to the query syntax already being used for the select alias.

cebe:
> is more related to the query syntax already being used for the select alias.

yes, it is, but in this case it does not match how the current syntax works. The array keys are always the relation name so it would not work if we add a case where the key is the alias. Adding a syntax that works the otherway round, i.e. `'relationName' => 'alias'` would cause even more confusion about in which case which syntax would work. I think using this syntax, which is the original proposal of #2377 is the best solution as it is compatible with how from() and join() work when a string is passed.

framework/db/ActiveQuery.php
```diff
@@ -382,8 +396,42 @@ protected function queryScalar($selectExpression, $db)
      */
     public function joinWith($with, $eagerLoading = true, $joinType = 'LEFT JOIN')
     {
-        $this->joinWith[] = [(array) $with, $eagerLoading, $joinType];
+        $relations = [];
+        foreach((array) $with as $name => $callback) {
```
samdark: Won't it cast at each iteration?

framework/db/ActiveQuery.php
```diff
@@ -382,8 +396,42 @@ protected function queryScalar($selectExpression, $db)
...
+            if (is_int($name)) {
+                $name = $callback;
+                $callback = null;
+            }
 
+            if (preg_match('/^(.*?)(\s+AS\s+|\s+)(\w+)$/i', $name, $matches)) {
+                // relation is defined with an alias, adjust callback to apply alias
+                list(, $relation, , $alias) = $matches;
```
samdark: In the regex above why doing second match if it's not used? Could be non-match group instead: `(?:\s+AS\s+|\s+)`

Faryshta:
> $query->andWhere([$query->applyRelationAlias('books', 'category') => 'Science fiction']);

I also see this counter intuitive. I have shown it to everyone at work and everyone had a different interpretation of what it means.

framework/db/ActiveQuery.php
cebe: no? same as if I would do `foreach($this->getMeSomeArray() as $k => $v)`. This would call the method once.

cebe:
> I also see this counter intuitive. I have shown it to everyone at work and everyone had a different interpretation of what it means.

I am open for choosing a different naming for this, proposals?

framework/db/ActiveQuery.php
cebe: done, thanks. c574155

Faryshta:
I honestly don't know what it means. I have my own theory but I would like you to explain me please.

framework/db/ActiveQuery.php
samdark: Right. I was sleepy :)

cebe:
I'm going to write documentation for this soon.

framework/db/ActiveQuery.php
```diff
@@ -716,4 +765,79 @@ public function viaTable($tableName, $link, callable $callable = null)
...
+    private $_relationAliases = [];
+    /**
+     * @var array array of aliases or table names from relations joined by joinWith() for which
+     * we do not have an explicit table alias. Array keys are
+     * the relation names, a value of `false` indicates, that the relation was added by joinWith() but
+     * the alias has not been used before. It will be populated upon a call to [[getRelationAlias()]].
+     */
+    private $_relationAliasesLookup = [];
+    /**
+     * @var array
```
SilverFire: Need to add PHPDoc

SilverFire:
looks great

Faryshta:
What I would expect to see.

```php

$order = Order::find()->joinWith(['books b'], true, 'INNER JOIN')
    ->where(['b.category' => 'Science fiction'])
    ->one();

echo $order->b->category // prints 'Science fiction'
```

framework/db/ActiveQuery.php
```diff
@@ -716,4 +765,79 @@ public function viaTable($tableName, $link, callable $callable = null)
...
+     */
+    private $_relationAliasesCallback = [];
+
+    // so tablename of relation does not need to be known
```
samdark: What's this comment for?

framework/db/Query.php
```diff
@@ -824,6 +836,65 @@ public function addParams($params)
...
+     *    numeric key, string value == table without alias
+     *    string key, string value == table with alias
+     *    string key, obect value == subquery with alias. This alias will not be populated
+     * @see from()
+     * @see join()
+     */
+    protected function populateAliases($tables)
+    {
+        if (is_string($tables) && preg_match('/^(.*?)(\s+AS\s+|\s+)({{\w+}}|\w+)$/i', $tables, $matches)) {
+            $this->aliases[$matches[1]] = $matches[2];
```
samdark: Is it OK? There are 3 catching groups in regex but only 2 matches are used.

framework/db/Query.php
```diff
@@ -824,6 +836,65 @@ public function addParams($params)
...
+        } elseif (is_array($tables)) {
+            foreach ($tables as $alias => $tableName) {
+                if (is_object($tableName)) {
+                    continue;
+                }
+                if (is_string($alias)) {
+                    $this->aliases[$tableName] =  $alias;
+                } elseif (preg_match('/^(.*?)( as |\s+)({{\w+}}|\w+)$/i', $tableName, $matches)) {
```
samdark: Is it OK? There are 3 catching groups in regex but only 2 matches are used.

tests/framework/db/ActiveRecordTest.php
```diff
@@ -409,6 +410,33 @@ public function testJoinWith()
...
+        $this->assertEquals(3, $orders[1]->id);
+        $this->assertEquals(1, $orders[2]->id);
+        $this->assertTrue($orders[0]->isRelationPopulated('customer'));
+        $this->assertTrue($orders[1]->isRelationPopulated('customer'));
+        $this->assertTrue($orders[2]->isRelationPopulated('customer'));
+
+        // join with table alias sub-relation
+        $orders = Order::find()->innerJoinWith([
+            'items as t' => function ($q) {
+//                $q->from(['t' => 'item']);
```
samdark: Remove commented out code.

framework/db/Query.php
cebe: first one catches the table name which is not used here so it may be removed.

cebe:
@Faryshta here are the docs, hope it is more clear now: https://github.com/yiisoft/yii2/commit/9557c2be653ec8dc70ab4e2227d983bc7372fae1

Faryshta:
I still think its very un-intuitive, I guess I understand what it does now and its not easy to read, besides it undermines the purpose of the alias if you still have to use the full relation name to use it. And actually

`where([$this->applyRelationAlias('affiliate', 'registration_date') => '2015-01-01']);`

is much more code than

`where([''affiliate.registration_date' => '2015-01-01']);`

and much less intuitive. What you want with an alias is to reduce the amount of code and improve the legibility, this achieves the exact opposite.

Also it doesn't fix one of the feature requests which is define an alias for the relation itself. instead of accesing `$client->orders` being able to define an alias to use `$client->ods` or `$client->bigOds` dinamically on the fly.

cebe:
> Also it doesn't fix one of the feature requests which is define an alias for the relation itself. instead of accesing $client->orders being able to define an alias to use $client->ods or $client->bigOds dinamically on the fly.

I have not seen this request in any of the related issues and it also does not make any sense to me. What problem do you solve with this?

@Faryshta the point here is that you do not need to know the **table name** of the relation and still define an alias for it to be able to disambiguate the column names. This is what was originally reported in the issue https://github.com/yiisoft/yii2/issues/2377#issue-27218719.

Note that this pull request solves a lot more problems then just #2377 as it introduces aliases handling in Query and ActiveQuery as a complete set of methods. `applyRelationAlias()` has nothing to do with issue #2377. It is an additional method for solving issues where the alias is defined in the relation definition, e.g. #5137, #8743 and #9326.

Faryshta:
I requested that feature here https://github.com/yiisoft/yii2/issues/6951

and it works to be able to reuse the same relation for different executions of the same query. Example lets say you want to show the top ten customers and show their shoppings this month, week, day and over all

```php
Customer::find()->joinWith([
    'orders',
    'orders dayOrders' => function ($query) {
        $query->andWhere('date(created_at) = date(now())');
    },
    'orders weekOrders' => function ($query) {
        $query->andWhere('week(created_at) = week(now()) and year(created_at) = year(now())');
    },
    'orders monthOrders' => function ($query) {
        $query->andWhere('month(created_at) = month(now()) and year(created_at) = year(now())');
    },
])->limit(10);
```

Also it would allow simplier `joinWith()`

`Customer::find()->joinWith(['orders o', 'o.products p', 'o.p.description d'])`

to give an example.

>  has nothing to do with issue #2377. It is an additional method for solving issues where the alias is defined in the relation definition, e.g. #5137, #8743 and #9326.

I still don't like it but I understand why its better at the moment so I am not debating that at all anymore.

> the point here is that you do not need to know the table name of the relation and still define an alias for it to be able to disambiguate the column names. This is what was originally reported in the issue

how about a simplier notation then? Something like

`where('{{@affiliate.registration_date}}' => $date])`

it will also be useful when you use nested relations

`where('{{@affiliate.profile.name}}' => $name])`

cebe:
I have reopened #6951 as it seems unrelated to this one.

I like the idea of introducing a replacement syntax for the table name, need to check whether it is easily possible to implement that.

Faryshta:
thanks. I am glad I could provide a useful idea.

what about the other idea I gave you? the one about the aliased relation name?

cebe:
> what about the other idea I gave you? the one about the aliased relation name?

I think we should discuss that in #6951 as it does not seem related to this and would need a totally different approach for implementing.

cebe:
splitting this PR into multiple ones for easier review:

first one comes here: https://github.com/yiisoft/yii2/pull/10813

cebe:
second PR: https://github.com/yiisoft/yii2/pull/10817

PR #10258, opened 2015-11-26
Change ServiceLocator::_components and _definitions to protected

derekisbusy:


cebe:
please elaborat in detail why you think this change is needed.

derekisbusy:
So that I can override the get() method in my module without also having to override all the other methods that use _components and _definitions. I guess there's only a few other methods so not too important.

cebe:
how would you override get()?

derekisbusy:
I'm extending the Module class and implementing ArrayAccess so that the module will return different components depending on the offset accessed. In this case I'm using it for a user module where the components differ depending on the user type.

```php
$module[User::TYPE_VENDOR]->layouts;
// would return a different component configuration
$module[User::TYPE_CUSTOMER]->layouts;
```

```php

    public function get($id, $throwException = true, $key = null)
    {
        if ($key !== null) {
            if (is_object($this->{$this->_containerName}[$key][$id]) && !$this->{$this->_containerName}[$key][$id] instanceof Closure) {
                return $this->{$this->_containerName}[$key][$id];
            } elseif ($key == $this->getCurrent() && isset($this->_components[$id])) {
                return isset($this->_components[$id]);
            } else {
                $config = [];
                // default definitions
                if (isset($this->_defaults[$id])) {
                    $config = $this->_defaults[$id];
                }
                // config components
                if (isset($this->_config[$id])) {
                    $config = array_replace_recursive($config, $this->_config[$id]);
                }
                // module types
                if (isset($this->{$this->_containerName}[$key][$id])) {
                    $config = array_replace_recursive($config, $this->{$this->_containerName}[$key][$id]);
                }
                if (!empty($config)) {
                    $this->set($id, $config);
                    return $this->{$this->_containerName}[$key][$id] = Yii::createObject($config);
                }
            }
        }
        return parent::get($id, $throwException);
    }
```

cebe:
You could make `$module[User::TYPE_CUSTOMER]` return a different service locator than `$module[User::TYPE_VENDOR]`. That would be a better solution imo.

PR #10263, opened 2015-11-26
Add testValidateTargetClass for reproduce the bug.

githubjeka:
When IDs targetClass & model(not new)  is same then unique validator passed

### DIFF
--- a/tests/framework/validators/UniqueValidatorTest.php
+++ b/tests/framework/validators/UniqueValidatorTest.php
@@ -5,6 +5,8 @@ namespace yiiunit\framework\validators;
 use yii\validators\UniqueValidator;
 use Yii;
 use yiiunit\data\ar\ActiveRecord;
+use yiiunit\data\ar\Category;
+use yiiunit\data\ar\Item;
 use yiiunit\data\ar\Order;
 use yiiunit\data\ar\OrderItem;
 use yiiunit\data\validators\models\FakedValidationModel;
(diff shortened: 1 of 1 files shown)


githubjeka:
This pull for reproduce on travic

LAV45:
OK. You only check for the insert of linked data, but what about the updates?

meshzp:
Hi, mates! We've noticed that unique validator logic was changed. So,
a. What we must to do now if we need to update non-unique fields in model that has unique attributes?
b. What we must to do if we need to update single value in unique attribute in model, which has several unique attributes?

For example. We have User model which has unique attributes like nickname and email. We need to allow to change that attributes to any other values which are not in use by other users. 

Right now uniqueness is checking by query like
SELECT 1 FROM `users` WHERE `email`='someuser@somedomain.com'
Can we add to this query current id of model which we are going to update?
SELECT 1 FROM `users` WHERE `email`='someuser@somedomain.com' AND id != {:id}

githubjeka:
@meshzp via [$filter](http://www.yiiframework.com/doc-2.0/yii-validators-uniquevalidator.html#$filter-detail)

This PR not changed logic work with an model. 
See https://github.com/yiisoft/yii2/pull/10265/files#diff-b7eeda2fac6ce8131377ce36953a329aR108 . Changed only `$model::className() !== $targetClass` , It's logic work with diff models.

SilverFire:
@meshzp could you check, whether your case works in another way at 2.0.6?

meshzp:
Yes. Update worked for me and fails now. On the first look, i can see that month ago before my last YII2 update (not sure that it was exactly 2.0.6) uniquiness was checking by query like

SELECT count(*) WHERE email='someuser@somedomain.com' LIMIT 2 

and failed when the count equals two.
Will look deeper through the code and let you know if found some more details. Thanks.

meshzp:
@githubjeka using the filter is the solution, however, we should apply it to all unique attributes in a project and consider on scenarios we are using. Too much unexpected work to do.

meshzp:
@githubjeka @SilverFire 
May be this is my bad but i have stoped \vendor\yiisoft\yii2\validators\UniqueValidator.php on line 109 with xDebug and found that it walks into condition $model->className() !== $targetClass , because
$model->className() = "common\models\User"
$targetClass = "\common\models\User" 

difference in leading slash in some reason.

meshzp:
@samdark
The problem was found, thanks for your attention. This is funny but the problem related with YII2 advanced application template SignUpForm.php. I think that it will not work correctly from now. 
 https://github.com/yiisoft/yii2-app-advanced/commit/cba4e7711d57632a1f5306218bbc471961951a3d line 30.

githubjeka:
Different scope of namespace. Funny ^_^... need a fix

samdark:
@githubjeka can you handle it?

githubjeka:
Done

SilverFire:
@meshzp thank your for your attention, the bug is fixed https://github.com/yiisoft/yii2/commit/c9580b5ba406bdf3652b91004d18e6c59cdb6b05

Could you check now, please?

PR #10265, opened 2015-11-26
Fix bug: When IDs targetClass & model(not new) is same then unique validator passed

githubjeka:
Fix #10263

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.7 under development
 -----------------------
-
+- Bug #10263: Unique validator passed for update-scenario when IDs targetClass and model is same (bupy7, githubjeka)
 - Bug #6351: Find MySQL FK constraints from `information_schema` tables instead of `SHOW CREATE TABLE` to improve reliability (nineinchnick)
 - Bug #6363, #8301, #8582, #9566: Fixed data methods and PJAX issues when used together (derekisbusy)
 - Bug #6876: Fixed RBAC migration MSSQL cascade problem (thejahweh)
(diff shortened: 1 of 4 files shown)


tests/framework/validators/UniqueValidatorTest.php
```diff
@@ -136,4 +138,28 @@ public function testValidateCompositeKeys()
         $val->validateAttribute($m, 'id');
         $this->assertFalse($m->hasErrors('id'));
     }
+
+    public function testValidateTargetClass()
+    {
+        $val = new UniqueValidator([
+            'targetClass' => Category::className(),
+            'targetAttribute' => ['name'],
```
cebe: could you explain the semantics behind this validation? why would I validate uniqueness of item name against categories?

githubjeka: This is done so only for reproduction

githubjeka: I need two models with same attributes for unique

samdark: UUID?

samdark: Seems to be valid case.

cebe:
needs a changelog

framework/CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.7 under development
 -----------------------
-
+- Bug #10263: Unique validator passed for update-scenario when IDs targetClass and model is same (bupy7, githubjeka)
```
SilverFire: We order our changelog by type (bug, enh, chg, new), then by issue ID.

githubjeka:
@cebe  if need semantic I can change test. Because attribute can be different.

SilverFire:
Merged with minor adjustments: https://github.com/yiisoft/yii2/commit/a1ff095f91f1da59305fd108f1bf0c063a554cb1, https://github.com/yiisoft/yii2/commit/6eb9830d645f9a00ac5568e695e209a761b226d2. Thank you!

PR #9952, opened 2015-10-18
Fix to get data methods and pjax working together

derekisbusy:
I added this commit before #9432 but never got merged and ended up having to reset my branch because the branch was both ahead and behind in commits and wasn't sure how to resolve that with git. Anyway here are the updates applied again.. I also added support to fix @elastic-od issue in previous patch. I have tested this code and it does work. Please merge with master branch. Thanks!

PR #10267, opened 2015-11-27
Unable to get data passed via "data-params" in pjax events

psolom:
Hey, I have found another issue related to #9952.
I am not able to reach extra params passed to pjax via "data-params" attribute.

My button link options are:

``` php
[
    'data-method' => 'post',
    'data-pjax' => 1,
    'data-pjax-container' => '#data-grid',
    'data-pjax-push-state' => 0,
    'data-pjax-timeout' => 0,
    'data-params' => ['foo' => 'bar'],
];
```

But I was not able to get ['foo' => 'bar'] object in pjax events ('pjax:success' e.g.)

I have discovered that "data-params" is not passed to pjax options in yii.js
To solve this issue "pjaxOptions" var in yii.js should be extended with "params" option, and shall look like:

``` js
pjaxOptions = {
    container: pjaxContainer,
    push:pjaxPushState,
    replace:pjaxReplaceState,
    scrollTo:pjaxScrollTo,
    timeout:pjaxTimeout,
    params:params
}
```

After the change I made I have managed to reach passed params as:

``` js
$('#data-grid').on('pjax:success', function (event, data, status, xhr, options) {
    options.params // contains {foo: "bar"} object
});
```

CHANGELOG: 

Option `params` replaced with `originalEvent` which is more flexible as you can reach triggered DOM object. For example, you are able to reach passed params in PJAX event as:

``` js
$('#data-grid').on('pjax:success', function (event, data, status, xhr, options) {
    $(options.originalEvent.currentTarget).data('params'); // contains {foo: "bar"} object
});
```

### DIFF
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -172,7 +172,7 @@ yii = (function ($) {
                     replace: pjaxReplaceState,
                     scrollTo: pjaxScrollTo,
                     timeout: pjaxTimeout,
-                    params: params
+                    originalEvent: event
                 }
             }
 
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -171,7 +171,8 @@ yii = (function ($) {
                     push: pjaxPushState,
                     replace: pjaxReplaceState,
                     scrollTo: pjaxScrollTo,
-                    timeout: pjaxTimeout
+                    timeout: pjaxTimeout,
+                    params: params
                 }
             }
 


SilverFire:
Well, I'm not sure whether we can accept the PR "as is" because `params` is not a documented property of PJAX plugin, so it will be unavailable unless `pjax.submit` was called by yii.js
However, the use case is valid and I can't find any better options to solve it.

When we are going to cheat a little, maybe it's better to store `$e` element (that triggered pjax form submit)? Having the element, we have much more control over it.

> Tip. `event` variable in `pjax:success` has the `pjax-container` as a target.

@samdark, @klimov-paul do you have any ideas?

samdark:
Having the element is better.

servocoder:
This not documented, you are right, but it is a common solution. In fact one is able to pass any "extra params" to PJAX plugin and it will be possible to get it in the plugin events. I have created this fix after reading this post:
http://stackoverflow.com/questions/21679655/jquery-pjax-ajax-success-function

Anyway I believe there is should be an abbility to pass such params. In my case this helped me to define which button in grid was clicked. I don't mind if somebody have any better solution for implementation.

SilverFire:
@servocoder I'm not standing against your idea, but I'm trying to make it more flexible.

Let's choose a name:
 - `triggeredBy`
 - `submitedBy`
 - propose your own :)

servocoder:
@SilverFire It is ok, I agree with you.

Both names are good for me :)

samdark:
@SilverFire in plain JS it's usually event `currentTarget` or `target`. The element often available via `this` as well.

SilverFire:
@samdark 

> ```js
$('#data-grid').on('pjax:success', function (event, data, status, xhr, options) {
    options.params // contains {foo: "bar"} object
});
```
> > Tip: `event` variable in `pjax:success` has the `pjax-container` as a target.

There is an `event` as the first attribute, but it points to the `pjax-container`, and we need to get the button, which triggered the event.

Although maybe it's a good idea to store event, instead of element.

What about `originalEvent`?

**Comment by samdark** [comment]
Yep. That sounds better than introducing another option.

**Comment by SilverFire** [comment]
@samdark it will be a new option :)

**Comment by SilverFire** [comment]
@servocoder could you update your PR and add a CHANGELOG line

**Comment by servocoder** [comment]
@SilverFire 
Yes, I am gonna do it now.
I can pass to PJAX both event and $e from yii.js
In this case you can access initial triggered object in two different ways:
```
$('#data-grid').on('pjax:success', function (event, data, status, xhr, options) {
        console.log($(options.originalEvent.currentTarget));
        console.log(options.originalTarget); // originalTarget is jQuery object
});
```
The second one is more obvious. Is this ok?

SilverFire:
`options.originalTarget.target` is the same as `options.originalTarget` in your example. I can't see any reasons for duplicating.

servocoder:
Not `options.originalEvent.target` but `options.originalEvent.currentTarget`.
`options.originalEvent.target` represent `<span>`, not `<a>` DOM object.
Anyway both can be reached via `options.originalEvent`, so I have commited changes and added changelog as you asked. Check it, please.

SilverFire:
Merged! https://github.com/yiisoft/yii2/commit/a301d7154a1c34508bbf3b9252ea52e7cd90c439

SilverFire:
Thank you!

servocoder:
@SilverFire 
We have to reopen this task.
I have found that `options.originalEvent` is undefined after deleting row in grid.
But `options.originalTarget` - `$e` jQuery object passed in yii.js - is ok.
So we have to pass `options.originalTarget` as well, to get target DOM object even if row was deleted.
Don't you mind if I add it to `options` ?

SilverFire:
Well, than the problem might be a little more complex. Passing an element by reference might be a wrong way, because it can be removed any moment.
Since it's a PR, I would kindly ask you ti create a new issue and we'll decide how to deal with the problem. Thank you

PR #10284, opened 2015-11-30
Fix #10217 for the case when there is hidden input combined with list view

### LINKED ISSUE #10217: GridView widget problem with array query parameters

omnilight:
I've found small, but important bug.

I the case when you have standard `Html::activeListView` as a filter input, it really adds two inputs: one is select with the name `name[]` and second is hidden input with name `name`. Second one is passed, when no value is selected.

In the case we have selected some value, query string will be: `name=&name[]=1&name[]=2`. That's correct. Now, if you deselect all values from select, `applyFilter` function will find only hidden's input value (because of `$(settings.filterSelector).serializeArray()`). So, when later it will try to override query string values, it will not override `&name[]=1&name[]=2` part because no key `name[]` will be in the `keysInFilter` variable. So my new pull-request is fixing with behavior
#10217

### DIFF
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -95,7 +95,7 @@
             var namesInFilter = Object.keys(data);
 
             $.each(yii.getQueryParams(settings.filterUrl), function (name, value) {
-                if (namesInFilter.indexOf(name) === -1) {
+                if (namesInFilter.indexOf(name) === -1 && namesInFilter.indexOf(name.replace(/\[\]$/, '')) === -1) {
                     if (!(name in data)) {
                         data[name] = [];
                     }


SilverFire:
I've checked this case during the validation the your previous PR and decided that this behavior is ok, because usually we add an option `<option name="state" value="">Any</option>`

But when you use select2 with multiple tags instead of default HTML select, it may not work.

I'll verify this PR a little bit later

omnilight:
Thank you

servocoder:
Hey,
I have faced an important issue, related to this one I believe.
I hope you guys, help clarify it to me and I will create pull request for it then.

My url:
`http://domain.com/product/index?ProductSearch%5Bname%5D=test&page=1&per-page=2&sel%5B%5D=8&sel%5B%5D=9&sel%5B%5D=10`

After it filtered with `applyFilter` method of `yii.gridView.js` it lost the second `sel` var and becomes like:
`http://domain.com/product/index?ProductSearch%5Bname%5D=test&page=1&per-page=2&sel%5B%5D=10`

So filtering cuts all vars with the same name and left only the last in order.
This happens due to `getQueryParams` method in `yii.js`. Here is the result of `console.log(yii.getQueryParams(settings.filterUrl));` in `applyFilter` method:

```
{
PromoteProduct[name]: "test",
page: "1",
per-page: "2",
sel[]: "10"
}
```

As you can see `yii.getQueryParams` doesn't respect multiple veriables with the same name. Expected behavior for me is it returns object like:

```
{
PromoteProduct[name]: "test",
page: "1",
per-page: "2",
sel: [
    "10",
}
```

omnilight:
As I found by the quick search of the framework files, `getQueryParams` method is only used in the `yii.gridView.js` file. 

Behaviour of the `getQueryParams` could be easily fixed, but wouldn't it break BC?

If so, we can simply implement special version for `yii.gridView.js`, or for example add special parameter for `getQueryParams` to respect `name[]` notation of variables

servocoder:
@omnilight 
I agree, now `getQueryParams` parses url very poorly.
It doesn't cover many cases of url params, so I have suggested to use either third-party solution instead of `getQueryParams` or... the code of some third-party solution inside `getQueryParams` :)
But I don't like the idea to reinvent the wheel.

SilverFire:
Merged, thank you!

SilverFire:
@servocoder you're right, this feature needs support by `getQueryParams`. If you can make a PR - go ahead, it will be great, I'll take care to fit it in upcoming 2.0.7 release.

SilverFire:
https://github.com/AceMetrix/jquery-deparam is a good plugin, but it's a little over-complicated and I don't like the idea of adding new JS dependency. I think we can make own simpler implementation

servocoder:
@SilverFire
I have created #10302, check please

PR #10290, opened 2015-11-30
Change createModles method from private to protected

speeeechless:
Inheritance problem when I extend yii\db\ActiveQuery (need to overwrite the populate method to add some processing functions)

### DIFF
--- a/framework/db/ActiveQueryTrait.php
+++ b/framework/db/ActiveQueryTrait.php
@@ -108,7 +108,7 @@ trait ActiveQueryTrait
      * @param array $rows
      * @return array|ActiveRecord[]
      */
-    private function createModels($rows)
+    protected function createModels($rows)
     {
         $models = [];
         if ($this->asArray) {


samdark:
What's your use case exactly?

speeeechless:
I’ve a table with translations in it and I want to get the translations based on the current culture transparently thru the ActiveRecord class when I use find(). Maybe I’m not going on the right way ?


De : Alexander Makarov [mailto:notifications@github.com]
Envoyé : lundi 30 novembre 2015 17:22
À : yiisoft/yii2
Cc : speeeechless
Objet : Re: [yii2] Change createModles method from private to protected (#10290)


What's your use case exactly?

samdark:
Are these message translations?

speeeechless:
Yes, kind of.

For exemple, I’ve a table News with the fields : id, title, summary, date and I’m storing the translations for the field title and summary in a specific table. I want to load those translations after getting the records from the table News. I’d like to use the event afterFind but this doesn’t work combinated with the asArray option.

samdark:
Please make sure it can't be implemented w/o the change. Check https://github.com/samdark/yii2-cookbook/blob/master/book/ar-single-table-inheritance.md for ideas.

speeeechless:
Hmm nice idea, it might work for a one to one relation but in my case this is a little bit different (one item, many translations, one for each field in each language). I’m setting up a translation model as describe here http://stackoverflow.com/questions/28118054/better-way-of-storing-translations-in-mysql (second option)

samdark:
I still think it can be done with the current accessible methods. Gonna try implementing it.

PR #10292, opened 2015-12-01
use ArrayHelper::remove() for the html helper methods

Faryshta:


framework/helpers/BaseHtml.php
```diff
@@ -911,10 +914,8 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
         } else {
             $hidden = '';
         }
-        $separator = isset($options['separator']) ? $options['separator'] : "\n";
 
-        $tag = isset($options['tag']) ? $options['tag'] : 'div';
-        unset($options['tag'], $options['unselect'], $options['encode'], $options['separator'], $options['item'], $options['itemOptions']);
+        unset($options['unselect']);
```
SilverFire: It can be moved to the `if` condition above.

framework/helpers/BaseHtml.php
```diff
@@ -884,9 +884,12 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
             $name .= '[]';
         }
 
-        $formatter = isset($options['item']) ? $options['item'] : null;
-        $itemOptions = isset($options['itemOptions']) ? $options['itemOptions'] : [];
-        $encode = !isset($options['encode']) || $options['encode'];
+        $formatter = ArrayHelper::remove($options, 'item', null);
```
SilverFire: no need to pass `null`, because 3rd argument of `remove` defaults to null

framework/helpers/BaseHtml.php
```diff
@@ -953,9 +954,12 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
      */
     public static function radioList($name, $selection = null, $items = [], $options = [])
     {
-        $encode = !isset($options['encode']) || $options['encode'];
-        $formatter = isset($options['item']) ? $options['item'] : null;
-        $itemOptions = isset($options['itemOptions']) ? $options['itemOptions'] : [];
+        $formatter = ArrayHelper::remove($options, 'item', null);
```
SilverFire: no need to pass `null`, because 3rd argument of `ArrayHelper::remove` defaults to null

framework/helpers/BaseHtml.php
```diff
@@ -981,8 +984,7 @@ public static function radioList($name, $selection = null, $items = [], $options
             $hidden = '';
         }
 
-        $tag = isset($options['tag']) ? $options['tag'] : 'div';
-        unset($options['tag'], $options['unselect'], $options['encode'], $options['separator'], $options['item'], $options['itemOptions']);
+        unset($options['unselect']);
```
SilverFire: It can be moved to the `if` condition above.

framework/helpers/BaseHtml.php
```diff
@@ -1012,11 +1014,10 @@ public static function radioList($name, $selection = null, $items = [], $options
...
     public static function ul($items, $options = [])
     {
-        $tag = isset($options['tag']) ? $options['tag'] : 'ul';
-        $encode = !isset($options['encode']) || $options['encode'];
-        $formatter = isset($options['item']) ? $options['item'] : null;
-        $itemOptions = isset($options['itemOptions']) ? $options['itemOptions'] : [];
-        unset($options['tag'], $options['encode'], $options['item'], $options['itemOptions']);
+        $tag = ArrayHelper::remove($options, 'tag', 'ul');
+        $encode = ArrayHelper::remove($options, 'encode', true);
+        $formatter = ArrayHelper::remove($options, 'item', null);
```
SilverFire: the same as before

framework/helpers/BaseHtml.php
```diff
@@ -1085,8 +1086,8 @@ public static function activeLabel($model, $attribute, $options = [])
     {
         $for = array_key_exists('for', $options) ? $options['for'] : static::getInputId($model, $attribute);
```
SilverFire: Also may be redone with `ArrayHelper`

Faryshta:
@SilverFire I noticed that `checkBoxList()` has an special option named 'separator' to be used on the `implode()` at the end but `radioList()` doesn't have the same feature. Should I add it?

SilverFire:
@Faryshta sure

samdark:
I'd measure performance after applying the change.

Faryshta:
I think its ready. And I don't see how this should impact too much in performance.

Plus the advantage of this PR is that the code looks much cleaner now.

framework/helpers/BaseHtml.php
```diff
@@ -884,9 +884,12 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
             $name .= '[]';
         }
 
-        $formatter = isset($options['item']) ? $options['item'] : null;
-        $itemOptions = isset($options['itemOptions']) ? $options['itemOptions'] : [];
-        $encode = !isset($options['encode']) || $options['encode'];
+        $formatter = ArrayHelper::remove($options, 'item');
+        $itemOptions = ArrayHelper::remove($options, 'itemOptions', []);
+        $encode = ArrayHelper::remove($options, 'encode', true);
+        $separator = ArrayHelper::remove($options, 'separator', PHP_EOL);
```
samdark: I think it should use `\n` as it was before.

Faryshta: PHP_EOL is specially useful when you want your code to be multiplataform and you will save the output to a file for example a cache file.

I can change it if you want http://stackoverflow.com/questions/4975411/when-do-i-use-php-eol-instead-of-n-and-vice-versa-ajax-jquery-client-problem

samdark: Yes but it's changing current behavior.

Faryshta: is that causing any bug or bc break? none of the tests resulted in failure.

samdark: We have no tests for it.

Faryshta: We have no reason to believe this causes any bc break or bug then.

samdark: We have HTML output. Newline there is to format HTML nicely. `PHP_EOL` gives different output on Windows (`\r\n`) and Linux (`\n`) but HTML is viewed on many different machines. In theory it could break weird platform-specific text editors.

In practice nothing will seriously break: browsers are OK with any form of line endings. To me it's still kinda clueless to use `PHP_EOL` in this case but I'm not strictly against it anymore.

samdark: Can be merged as is.

Faryshta: that is describing the opposite of what want to imply. the solution described there is to use "\r\n" instead of PHP_EOL or "\n"

The issue is that a file is being generated on linux and then its downloaded and got open on windows. the line separator on that file is "\n" so windows see it as a continuos line.

so the solution was to use "\r\n" instead of PHP_EOL or "\n" (which causes the same problem).

thats the only case in which it would happen and if you wanna fix it we should use "\r\n" by default instead of the other two options.

This change WILL make the tests fail for sure.

SilverFire: Agree with @samdark. We should leave `\n`

SilverFire:
Thank you!

PR #10297, opened 2015-12-01
Fix #9101 resolve appendTimestamp to registerJsFile and registerCssFile

### LINKED ISSUE #9101: AssetManager appendTimestamp not working if you use registerJsFile()
Timestamp is appended only if you create your own bundle. This is too complicated if you just want to register on JS file on certain page.

`registerJsFile()` creates bundle inside, but does not fill the `basePath`. I guess this is the reason timestamp is not appended.

saada:
I couldn't find a way to call assetManager->getAssetUrl to get the appendedTimestamp back. Is there a better approach to solve this problem?

### DIFF
--- a/framework/web/View.php
+++ b/framework/web/View.php
@@ -371,6 +371,7 @@ class View extends \yii\base\View
      */
     public function registerCssFile($url, $options = [], $key = null)
     {
+        $url = $this->appendTimestamp($url);
         $url = Yii::getAlias($url);
         $key = $key ?: $url;
         $depends = ArrayHelper::remove($options, 'depends', []);
@@ -435,6 +436,7 @@ class View extends \yii\base\View
      */
     public function registerJsFile($url, $options = [], $key = null)
     {
+        $url = $this->appendTimestamp($url);
         $url = Yii::getAlias($url);
         $key = $key ?: $url;
         $depends = ArrayHelper::remove($options, 'depends', []);
@@ -547,5 +549,21 @@ class View extends \yii\base\View
         }
 
         return empty($lines) ? '' : implode("\n", $lines);
+    }
+
+    /**
+     * Append timestamp if the url starts with the alias 'web' and the configuration is turned on
+     *
+     * @param $url
+     * @return string
+     */
+    public function appendTimestamp($url)
+    {
+        if (Yii::$app->assetManager->appendTimestamp && strncmp($url, '@web/', 5) === 0) {
+            $fileToCacheBust = Yii::getAlias(str_replace('@web', '@webroot', $url));
+            if (($timestamp = @filemtime("$fileToCacheBust")) > 0)
+                $url .= "?v={$timestamp}";
+        }
+        return $url;
     }
 }


framework/web/View.php
```diff
@@ -548,4 +550,20 @@ protected function renderBodyEndHtml($ajaxMode)
...
         return empty($lines) ? '' : implode("\n", $lines);
     }
+
+    /**
+     * Append timestamp if the url starts with the alias 'web' and the configuration is turned on
+     *
+     * @param $url
+     * @return string
+     */
+    public function appendTimestamp($url)
```
githubjeka: public ?

samdark:
#9101

Faryshta:
what if i don't want a timestamp? I don't see a way to choose if i want it or not.

what about generated files like .less or .ts ? will the generated file will know when there is a change in all the depending files?

what about personalized alias? like `Yii::setAlias('@web/ui/', '@web/userinterface')`

githubjeka:
> what if i don't want a timestamp? I don't see a way to choose if i want it or not.

`if (Yii::$app->assetManager->appendTimestamp`

> what about generated files like .less or .ts ? will the generated file will know when there is a change in all the depending files?

here `registerCssFile()` & `registerJsFile()` only

framework/web/View.php
```diff
@@ -548,4 +550,20 @@ protected function renderBodyEndHtml($ajaxMode)
...
+    {
+        if (Yii::$app->assetManager->appendTimestamp && strncmp($url, '@web/', 5) === 0) {
```
Faryshta: I think the first part of this if should be outside this function and be called on the `registerCssFile` and `registerJsFile` instead. As @githubjeka I might want to use this method for some file and not for others. `appendTimestamp` is the default behavior but it should be possible to use it file by file

saada:
@Faryshta Most of the logic related to the '@web' alias that I've implemented is derived and in line with the existing implentation for bundles: `AssetManager::getAssetUrl($bundle, $asset)` [(source)](https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L297-L323)
If you think there's a better approach then perhaps we should incorporate that into the existing logic as well. Or better yet, we could make one appendTimestamp method for both?

@githubjeka should I not keep the method public?

githubjeka:
>@githubjeka should I not keep the method public?

@saada yes - better is `protected`. But generally I think the behavior for adding timestamp should be execute from `assetManager` component, not from `View`. See my [Alternative way](https://github.com/yiisoft/yii2/pull/10429)

saada:
@githubjeka , I prefer your approach then. closing this in favor of your cleaner solution

githubjeka:
@samdark add milestone to my PR please .

cebe:
@githubjeka done.

SilverFire:
Resolved by dc23a6d3d8a9cd937cbbc64652db903ee4ff5598

githubjeka:
Please reopen. Because https://github.com/yiisoft/yii2/pull/10429#issuecomment-249145689

DarthLegiON:
@SilverFire I have a problem with this encancement. Here is my register JS code:
```php
$this->registerJsFile('@web/js/result-index.js', ['depends' => [AppAsset::className()]]);
```

I have an `appendTimestamp` property is set on `true` but timestamp doesn't appears. I started digging into your code and found that you don't use `basePath` property while creating new AssetBundle for file. And function `@filemtime("$basePath/$asset")` returns false because path `/js/result-index.js` doesn't exist.

Am I doing something wrong? Is this a bug or indefinite behavior?

githubjeka:
@DarthLegiON Can you send PR with a test for clarity like as [ dc23a6d](https://github.com/yiisoft/yii2/commit/dc23a6d3d8a9cd937cbbc64652db903ee4ff5598#diff-85bfe4d6bdfa423baec549f1a9bd3e0dR383) ?

DarthLegiON:
@githubjeka Oh I've just understood the reason of my behavior. @SilverFire reverted default values for `basePath` and `baseUrl` in https://github.com/yiisoft/yii2/commit/3ec3ce931216a44ec0b850c429c9d8c7f9a17e6d because of breaking BC. Is there a solution for my situation? Unfortunately, I have no time for creating a test right now but I want at least temporary solution for my problem. Maybe I should create an issue or post to existing one?

githubjeka:
Please create new issue.

PR #10302, opened 2015-12-02
applyFilter method in yii.gridView.js cuts off all vars with the same name and left only the last in order

psolom:
As I described in #10284 I have faced an important issue that brokes url upon filtering grid view. This PR brings fix to this issue. Description of the bug:

My url:
`http://domain.com/product/index?ProductSearch%5Bname%5D=test&page=1&per-page=2&sel%5B%5D=8&sel%5B%5D=9&sel%5B%5D=10`

After it filtered with `applyFilter` method of `yii.gridView.js` it lost all `sel` vars except last one and becomes like:
`http://domain.com/product/index?ProductSearch%5Bname%5D=test&page=1&per-page=2&sel%5B%5D=10`

So filtering cuts all vars with the same name and left only the last in order.
This happens due to `getQueryParams` method in `yii.js`. Here is the result of `console.log(yii.getQueryParams(settings.filterUrl));` in `applyFilter` method:

```
{
    ProductSearch[name]: "test",
    page: "1",
    per-page: "2",
    sel[]: "10"
}
```

As you can see `yii.getQueryParams` didn't respect multiple veriables with the same name. With the PR `yii.getQueryParams` returns now object like:

```
{
    ProductSearch[name]: "test",
    page: "1",
    per-page: "2",
    sel[]: [
        "8",
        "9",
        "10"
    ]
}
```


framework/assets/yii.js
```diff
@@ -276,12 +276,29 @@ yii = (function ($) {
...
+                params = {},
+                pair,
+                d = decodeURIComponent,
+                i;
+
+            for (i = pairs.length; i > 0;) {
+                pair = pairs[--i].split('=');
+                var name = d(pair[0]);
+                var value = d(pair[1]);
+                if(name.length) {
```
SilverFire: `if(` -> `if (`

framework/assets/yii.js
```diff
@@ -276,12 +276,29 @@ yii = (function ($) {
...
             }
-            var qs = url.substring(pos + 1).split('&');
-            for (var i = 0, result = {}; i < qs.length; i++) {
-                qs[i] = qs[i].split('=');
-                result[decodeURIComponent(qs[i][0])] = decodeURIComponent(qs[i][1]);
+
+            var pairs = url.substring(pos + 1).split('&'),
+                params = {},
+                pair,
+                d = decodeURIComponent,
```
SilverFire: This alias is useless because it brings nothing except confusion.

framework/assets/yii.js
```diff
@@ -276,12 +276,29 @@ yii = (function ($) {
...
+                i;
+
+            for (i = pairs.length; i > 0;) {
+                pair = pairs[--i].split('=');
+                var name = d(pair[0]);
+                var value = d(pair[1]);
+                if(name.length) {
+                    if (params[name] !== undefined) {
+                        if (!params[name].push) {
```
SilverFire: Is it a check for existence of `name` in `params`? Then it should be replaced with `!(name in params)`

SilverFire: Oh, i guess it'a check whether `params[name]` is an array.
Then I think it should be replaced with `!Array.isArray(params[name])`

framework/assets/yii.js
```diff
@@ -276,12 +276,29 @@ yii = (function ($) {
...
+                            params[name] = [params[name]];
+                        }
+                        params[name].push(value || '');
```
SilverFire: It will break parsing of integer 0

framework/assets/yii.js
```diff
@@ -276,12 +276,29 @@ yii = (function ($) {
...
-                qs[i] = qs[i].split('=');
-                result[decodeURIComponent(qs[i][0])] = decodeURIComponent(qs[i][1]);
+
+            var pairs = url.substring(pos + 1).split('&'),
+                params = {},
+                pair,
+                d = decodeURIComponent,
+                i;
+
+            for (i = pairs.length; i > 0;) {
```
SilverFire: Why are we iterating form backwards?

framework/assets/yii.js
servocoder: It won't after decodeURIComponent(0), it is ok

framework/assets/yii.js
servocoder: Fixed

framework/assets/yii.js
servocoder: Fixed

framework/assets/yii.js
servocoder: Fixed

framework/assets/yii.js
```diff
@@ -276,12 +276,28 @@ yii = (function ($) {
...
-            for (var i = 0, result = {}; i < qs.length; i++) {
-                qs[i] = qs[i].split('=');
-                result[decodeURIComponent(qs[i][0])] = decodeURIComponent(qs[i][1]);
+
+            var pairs = url.substring(pos + 1).split('&'),
+                params = {},
+                pair,
+                i;
+
+            for (i = pairs.length; i > 0;) {
```
SilverFire: Why are we iterating backwards?

servocoder: There was no real reason. Fixed

SilverFire:
`?name=alex&name=jack` => `{ name: ['alex', 'jack'] }`
It's not a valid behavior, since it will be encoded back to `?name[]=alex&name[]=jack`, which is inconsistent.

servocoder:
Which case do you mean?
If you are talking about something like `$.param({name: ['alex', 'jack'] })` - yes, it encodes back to `name%5B%5D=alex&name%5B%5D=jack`. But this is not a problem of implementation of `yii.getQueryParams`, you will get the same result using $.deparam plugin as I suggested in #10284, or any other. I am not sure is this a problem in fact, one should pass correct query params to get correct result, isn't it?

SilverFire:
Well, agree.

SilverFire:
Merged https://github.com/yiisoft/yii2/commit/b6b2befb00145e693935b4a523c854168d53c976

SilverFire:
Thank you!

PR #10309, opened 2015-12-02
Changed UrlManagers rule cacheKey to be a protected property

lordthorzonus:
I ran into an issue with route caching while extending the UrlManager in my project to load completely different sets of url rules per app locale. 

This enhancement makes the UrlManager and its descendants able to change the cache key with which the rules are stored and therefore allows it to be more easily extendable.


samdark:

PR #10310, opened 2015-12-02
Fetch data for refresh as array

arturf:
I think that object don't needed, because it used only for fetch new values from db


samdark:
Looks safe in general but I think it may break if there are custom getters or custom record population.

cebe:
This breaks type casting in `yii\db\ActiveRecord::populateRecord()`.

PR #10320, opened 2015-12-03
Fixed client-side IP validator

tafid:
I found a problem with client-side IP validation.
When the validated value does not have a CIDR, the `matches[4]` that represents CIDR part is `undefined` instead of expected empty string.
It leads to a wrong comparison in the if statement below, causing invalid behavior.
My patch fixes the default value for the described case and restores expected behavior.


SilverFire:
Merged, thank you!

PR #10335, opened 2015-12-06
Update db-active-record.md

agamayaservers:
добавил Содержимое раздела


SilverFire:
Merged! Thank you!

cebe:
@SilverFire I do not think  this is correct. TOC is generated on the framework page but should not be added manually like this.

SilverFire:
Yeah, sorry, it's my fault

cebe:
no problem.

PR #10343, opened 2015-12-07
Remove block comment [skip ci]

pana1990:


cebe:
Initially kept these to remember the syntax but I'm fine with removing it.

cebe:
Thanks!

PR #10347, opened 2015-12-07
Add support of verbs in actions naming

uaoleg:
Pretty often we have huge if-else blocks in our actions, e.g. 

``` php
if (\yii::$app->request->isPost)  { 
    ... 
} elseif (\yii::$app->request->isDelete) {
    ... 
} else {
    ... 
}
```

I propose to implement support on verbs in actions naming, e.g. `postActionPage()`, `deleteActionPage()`, etc. Simple `actionPage()` will be called of no verb-related action is found.


SilverFire:
Thank you for your idea. Actually I don't think, that the PR may be accepted as is, because it breaks the actions naming logic and is not enough extensible.

@yiisoft/core-developers when you don't share my opinion, reopen the PR, please.

cebe:
the idea is interesing in general but this is more a job of the router to decide which action to choose.

PR #10364, opened 2015-12-09
FIX rule for camelCase controller name in REST example

ruslanbat:
controller:w+s
if you  use this version of rule you cant get controller with camelCase name as  camel-case


samdark:
Thanks!

PR #10374, opened 2015-12-11
Moved AssetManager::$basePath verification to setter method

tomtomsen:
i started an application from scratch. at first i wasn't using assets, therefor i didn't need an assets folder.
later i added a stylesheet by  `$this->registerCss('body { padding: 100px }');` which results in a `InvalidConfigException("The directory does not exist: {$this->basePath}");`.

I think, the exception should only be thrown if the asset directory is really necessary.

### DIFF
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -66,7 +66,7 @@ class AssetManager extends Component
     /**
      * @var string the root directory storing the published asset files.
      */
-    public $basePath = '@webroot/assets';
+    public $_basePath;
     /**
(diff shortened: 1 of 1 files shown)


SilverFire:
Thank you for your PR, but unfortunately we can not accept it. Assets folder is required to use AssetManager and exists in basic and advanced application templates by default. Why are you trying to avoid using assets?

tomtomsen:
in my case i just didnt needed that folder, but i see your point. i will set the basePath to ```@web```. then i didnt need to create a folder which i do not use and i will not get an exception. (i thinks thats even better then my proposed change)

thanks.

PR #10377, opened 2015-12-12
Added support of doubles to fallback message formatter

### LINKED ISSUE #10363: Double formatting in i18n fallback
When intl extension is not available, fallback formatting works wrong for double values:

```php
echo Yii::t('app', 'Here is a big number: {n, number}', ['n' => 2e+8]);

// Intl: Here is a big number 200,000,000
// No intl: Exception(Message format 'number' is only supported for integer values. You have to install PHP intl extension to use this feature
```

Expected: float should be formatted as with intl

[MessageFormatter.php:344](https://github.com/yiisoft/yii2/blob/master/framework/i18n/MessageFormatter.php#L344)

AnatolyRugalev:
Fixes #10363 

@SilverFire, I'm stuck on this:

```
1) yiiunit\framework\i18n\FallbackMessageFormatterTest::testNamedArguments with data set #3 ('Here is a big number: {f, number}', 'Here is a big number: 200,000,000', array(200000000.0))
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Here is a big number: 200,000,000'
+'Here is a big number: 200000000'
/home/travis/build/yiisoft/yii2/tests/framework/i18n/FallbackMessageFormatterTest.php:151
/home/travis/build/yiisoft/yii2/vendor/phpunit/phpunit/phpunit:36
```

Is it really needs to split thousands as with `intl`?

I don't see that current implementation supports any number formatting, we are getting just plain number:
[MessageFormatter.php:345](https://github.com/yiisoft/yii2/blob/master/framework/i18n/MessageFormatter.php#L345)


SilverFire:
I've expected this problem :)

I thing that fallback formatting should be same as intl on EN locale. We can use [number_format](http://php.net/manual/en/function.number-format.php) to achieve it.

framework/i18n/MessageFormatter.php
```diff
@@ -341,7 +341,7 @@ private function parseToken($token, $args, $locale)
             case 'selectordinal':
                 throw new NotSupportedException("Message format '$type' is not supported. You have to install PHP intl extension to use this feature.");
             case 'number':
-                if (is_int($arg) && (!isset($token[2]) || trim($token[2]) === 'integer')) {
+                if (is_numeric($arg) && (!isset($token[2]) || trim($token[2]) === 'integer')) {
```
SilverFire: I think the condition should be a little more smart.

Possible values of `$token[2]` are `null`, `integer`, `currency`, or `percent`. Currency and percent are not supported, but we should implement a proper fallback for `integer` and `null`.

The following lines should return the same for both intl and fallback:
```php
echo Yii::t('app', 'Here is a big number: {n, number}', ['n' => 2e+8]); // 200,000,000
echo Yii::t('app', 'Here is a big number: {n, number, integer}', ['n' => 2e+8]); // 200,000,000

echo Yii::t('app', 'Here is a big number: {n, number}', ['n' => 200000000.101]); // 200,000,000.101
echo Yii::t('app', 'Here is a big number: {n, number, iinteger}', ['n' => 200000000.101]); // 200,000,000
```

AnatolyRugalev: Ok, got it. I will add tests for cases that you provided

AnatolyRugalev:
Ready for review

framework/i18n/MessageFormatter.php
```diff
@@ -341,8 +341,14 @@ private function parseToken($token, $args, $locale)
             case 'selectordinal':
                 throw new NotSupportedException("Message format '$type' is not supported. You have to install PHP intl extension to use this feature.");
             case 'number':
-                if (is_int($arg) && (!isset($token[2]) || trim($token[2]) === 'integer')) {
-                    return $arg;
+                $format = isset($token[2]) ? trim($token[2]) : null;
+                if (is_numeric($arg) && $format === null || $format === 'integer') {
```
SilverFire: Seems, you've removed parentheses and accidentally changed logic of condition

tests/framework/i18n/FallbackMessageFormatterTest.php
```diff
@@ -52,6 +59,38 @@ public function patterns()
...
+                ]
+            ],
+
+            [
+                'Here is a big number: {'.self::D.', number, integer}', // pattern
+                'Here is a big number: '.self::D_VALUE_FORMATTED_INTEGER, // expected
+                [ // params
+                    self::D => self::D_VALUE
+                ]
+            ],
```
SilverFire: also we need negative tests for fallback

`Here is a percent: {p, number, percent}` throws expected exception

framework/i18n/MessageFormatter.php
AnatolyRugalev: Ah, my bad, thanks

tests/framework/i18n/FallbackMessageFormatterTest.php
AnatolyRugalev: added

SilverFire:
Merged https://github.com/yiisoft/yii2/commit/3b31aa771a4f59cbf4bc01202a3f29e994f14cdd

SilverFire:
Thank you

PR #10378, opened 2015-12-12
add ArrayHelper::unique()

lav45:
This implemented often faster than native array_unique()

[benchmark array_unique() vs \yii\helpers\ArrayHelper::unique()](https://gist.github.com/LAV45/efed1e93948cd8369699)

### DIFF
--- a/framework/helpers/BaseArrayHelper.php
+++ b/framework/helpers/BaseArrayHelper.php
@@ -466,6 +466,21 @@ class BaseArrayHelper
     }
 
     /**
+     * Removes duplicate values from an array
+     * It's often faster than native array_unique
+     *
+     * @param string[]|int[] $array Can be only STRING or INTEGER values!
+     * @param bool $keepKeys Whether to maintain the array keys. If false, the resulting array
+     * @return array
+     */
+    public static function unique($array, $keepKeys = true)
+    {
+        return $keepKeys === true ?
+            array_flip(array_flip($array)) :
+            array_keys(array_flip($array));
+    }
+
+    /**
      * Checks if the given array contains the specified key.
      * This method enhances the `array_key_exists()` function by supporting case-insensitive
      * key comparison.
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -194,6 +194,8 @@ Yii Framework 2 Change Log
 - Enh #10545: `yii\web\XMLResponseFormatter` changed to format models in a proper way (andrewnester)
 - Enh #10783: Added migration and unit-tests for `yii\i18n\DbMessageSource` (silverfire)
 - Enh #10797: Cleaned up requirements checker CSS (muhammadcahya)
+- Enh #10840: Added `yii\console\Controller::optionAliases()` method to support aliases for commands (pana1990)
+- Enh #10378: Add `yii\helpers\ArrayHelper::unique()` it's often faster than native array_unique (LAV45)
 - Enh: Added last resort measure for `FileHelper::removeDirectory()` fail to unlink symlinks under Windows (samdark)
 - Enh: `AttributeBehavior::getValue()` now respects the callable in array format (silverfire)
 - Chg #6419: Added `yii\web\ErrorHandler::displayVars` make list of displayed vars customizable. `$_ENV` and `$_SERVER` are not displayed by default anymore (silverfire)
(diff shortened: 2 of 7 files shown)


AnatolyRugalev:
My benchmark results are:
```
array_unique($arr)              to 666667 in 5.1696400642395
foreach + array_keys            to 666667 in 0.42175698280334
ArrayHelper::unique($arr)       to 666667 in 0.25078105926514
ArrayHelper::unique($arr, true) to 666667 in 0.29126000404358
```

I think this is a good idea, but I think case with non-scalar data should be handled with exception

AnatolyRugalev:
Probably, you should add a note to changelog

samdark:
Verified that indeed this method is faster. Any idea why?

framework/helpers/BaseArrayHelper.php
```diff
@@ -382,6 +382,24 @@ public static function map($array, $from, $to, $group = null)
...
+     * It's often faster than native array_unique
+     *
+     * @param string[]|int[] $array Can be only STRING or INTEGER values!
+     * @param bool $keepKeys Whether to maintain the array keys. If false, the resulting array
+     * @return array
+     */
+    public static function unique($array, $keepKeys = true)
+    {
+        if ($keepKeys === true) {
+            $result = array_flip(array_flip($array));
```
Faryshta: why not put the return here already? it would save a bit of memory since you don't create a new variable.

it can also be done in line 397

cebe:
> Verified that indeed this method is faster. Any idea why?

because `array_unique` works not this way. it does sorting on the array and dependend on the sorting different strings are considered equal or not. http://php.net/manual/en/function.array-unique.php
We should carefully check if this is the expected behaviors in all places where the new option is used.

Faryshta:
another issue that comes to mind is that `ArrayHelper::unique([0, false, "", 1, "1", true])` will return `[2 => 0, 5 => 1]`

LAV45:
ArrayHelper::map() this isn't the same as array_map()
I use both depending on the situation.

ArrayHelper::unique() is not intended to completely replace the native array_unique(). ArrayHelper::unique() - accelerates the most frequent use case.

cebe:
the documentation should be more clear about this and explain the difference to php array_map in detail.

framework/helpers/BaseArrayHelper.php
```diff
@@ -382,6 +382,21 @@ public static function map($array, $from, $to, $group = null)
     }
 
     /**
+     * Removes duplicate values from an array
+     * It's often faster than native array_unique
+     *
+     * @param string[]|int[] $array Can be only STRING or INTEGER values!
+     * @param bool $keepKeys Whether to maintain the array keys. If false, the resulting array
```
samdark: I think there should be more description after `the resulting array`

samdark: - true - keeps keys
- false - remove keys

samdark: If true, keeps keys as is. If false, removes keys.

framework/helpers/BaseArrayHelper.php
```diff
@@ -466,6 +466,21 @@ public static function map($array, $from, $to, $group = null)
     }
 
     /**
+     * Removes duplicate values from an array
+     * It's often faster than native array_unique
+     *
+     * @param string[]|int[] $array Can be only STRING or INTEGER values!
```
samdark: Need to explain why.

framework/helpers/BaseArrayHelper.php
```diff
@@ -466,6 +466,21 @@ public static function map($array, $from, $to, $group = null)
     }
 
     /**
+     * Removes duplicate values from an array
+     * It's often faster than native array_unique
```
samdark: Need to mention limitations.

framework/helpers/BaseArrayHelper.php
```diff
@@ -466,6 +466,21 @@ public static function map($array, $from, $to, $group = null)
...
+     *
+     * @param string[]|int[] $array Can be only STRING or INTEGER values!
+     * @param bool $keepKeys Whether to maintain the array keys. If false, the resulting array
+     * @return array
```
samdark: Need `@since` tag.

sgabler:
So apart from those missing annotations, this pull request would be good to go? Or can I help in any way by providing some additional tests/functionality?

samdark:
It's good apart from what I've commented on.

PR #10379, opened 2015-12-13
Optimize the code

TianJinRong:
Missing the return process when send content.


cebe:
There is no need for `return` in a `void` method.

PR #10393, opened 2015-12-15
file validator and mimeTypes by mask #5469

### LINKED ISSUE #5469: file validator and mimeTypes by mask
Hello.
I need to upload images and pdf files therefore I can't use image validator and I need to use file validator.
My proposal is to add checking ```mimeTypes``` by mask similar on html5 client validator of input tag.

example: ```'accept' => 'image/*,.pdf'```

kirsenn:
Closes #5469

Added feature, now you can use

``` php
new FileValidator([
    'mimeTypes' => ['image/*', 'text/*']
]);
```

Client validation also available.
I've tried to add some unit tests, but I couldn't mock the FileHelper which method getMimeType is trying to reach real file in filesystem, so I removed them, you can check it at https://github.com/yiisoft/yii2/commit/39f737df206893079ac008434377dc19a18f6e9f

How it works.
When mimeTypes are presented the `FileValidator::validateMimeType` will be called, if any mime type has '*' symbol then it calls `FileValidator::validateMimeTypeMask` method which creates a regular expression for validating by mask. The same at client side.


SilverFire:
We need PHPUnit tests of the feature

kirsenn:
@SilverFire Yes, but there is a problem with FileHelper::getMimeType it requires a real file to exists in filesystem which is not possible for unit test. I couldn't find out how to mock it.
I got an idea if new method FileValidator::validateMimeType would accept only 1 argument - uploaded file mime type, then, making this method public we could write some tests. What do you think?

samdark:
I think the test should put the file into FS for real.

SilverFire:
Checkout for the [FileValidatorTest::createTestFiles()](https://github.com/yiisoft/yii2/blob/master/tests/framework/validators/FileValidatorTest.php#L214) method, I guess that's what wee need

kirsenn:
How I see it
```php
/**
     * @param string $fileMimeType
     * @return bool
     * @throws \yii\base\InvalidConfigException
     */
    public function validateMimeType($fileMimeType)
    {
        foreach ($this->mimeTypes as $mimeType) {
            if (strpos($mimeType, '*') !== false && $this->validateMimeTypeMask($fileMimeType, $mimeType)) {
                return true;
            }

            if ($mimeType == $fileMimeType) {
                return true;
            }
        }

        return false;
    }
```
then in `protected function validateValue($file)` do
`
elseif (!empty($this->mimeTypes) &&  !$this->validateMimeType(FileHelper::getMimeType($file->tempName))) {
`

The disadvantage is in making method public

kirsenn:
@samdark if we put in real FS it becomes not a unit test, we need to generate files with different mimetypes
@SilverFire It doesn't work as this method creates only instance of UploadedFile, but FileHelper requires a real file in FS

samdark:
@kirsenn we don't care how it's called and how unit it is if it does the job.

kirsenn:
More problem is that FileValidator is tightly coupled with FileHelper implementation and it doesn't use DI. Having DI we could put mock FileHelper implementation.

kirsenn:
@samdark So decision is to create some files in test directory with different mime types and test them? Do I get you right?

SilverFire:
Yes, it will be OK

kirsenn:
@SilverFire What if it's new Star Wars movie eh? :-) All right I'll create some empty text and image files. What directory would your recommend to place these test files?

SilverFire:
@kirsenn Do you already have Star Wars in 1080p? :open_mouth: 
I guess `tests/framework/validators/data/mimeType/image_jpeg.jpg` will be OK.

kirsenn:
@SilverFire sadly, no
Added tests, please review

SilverFire:
I like it, thank you!

SilverFire:
I'll check JS part a little bit later

framework/assets/yii.validation.js
```diff
@@ -436,5 +436,32 @@ yii.validation = (function ($) {
         }
     }
 
+    function validateMimeType(mimeTypes, fileType)
+    {
+        for (var i=0; i<mimeTypes.length; i++) {
```
samdark: `for (var i=0, len = mimeTypes.length; i<len; i++) {` is a bit faster way to iterate since check is executed each time.

framework/assets/yii.validation.js
```diff
@@ -436,5 +436,32 @@ yii.validation = (function ($) {
...
+            if (mimeTypes[i].indexOf('*') !== -1 && validateMimeTypeMask(mimeTypes[i], fileType)) {
+                return true;
+            }
+
+            if (mimeTypes[i] == fileType) {
```
samdark: `===` could be safely used. Also this check is simpler than the one above. Thus should be put on top.

samdark: This check is simpler than the one above and should come first.

samdark: Or could be combined:

```javascript
if (mimeTypes[i] === fileType || mimeTypes[i].indexOf('*') !== -1 && validateMimeTypeMask(mimeTypes[i], fileType)) {
    return true;
}
```

framework/assets/yii.validation.js
```diff
@@ -436,5 +436,32 @@ yii.validation = (function ($) {
...
+    function validateMimeTypeMask(mimeTypeMask, fileType)
+    {
+        var escapedString = escapeRegExp(mimeTypeMask);
+        var regexp = new RegExp('^' + escapedString.replace('\\*', '.*') +  '$');
+
+        return regexp.test(fileType);
+    }
+
+    function escapeRegExp(str) {
+        return str.replace(/[-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
```
samdark: There's no need to escape most of the chars in this regex. Could be:

```javascript
return (str+'').replace(/[.?*+^$[\]\\(){}|-]/g, "\\$&");
```

samdark: I guess replace at https://github.com/yiisoft/yii2/pull/10393/files#diff-56c3babe983e0fddaebd160c0bf620bdR457 is needed because of extra slashes...

framework/validators/FileValidator.php
```diff
@@ -420,4 +420,37 @@ protected function getClientOptions($model, $attribute)
 
         return $options;
     }
+
+    /**
+     * @param $mimeType
```
samdark: Need docs.

framework/validators/FileValidator.php
```diff
@@ -420,4 +420,37 @@ protected function getClientOptions($model, $attribute)
...
+     * @return bool
+     */
+    protected function validateMimeTypeMask($mimeType, $mask)
+    {
+        $regexp = "/^" . str_replace('\*', '.*', preg_quote($mask, "/")) . "$/";
+        return (bool)preg_match($regexp, $mimeType);
+    }
+
+    /**
+     * @param UploadedFile $file
```
samdark: Need docs.

framework/validators/FileValidator.php
```diff
@@ -420,4 +420,37 @@ protected function getClientOptions($model, $attribute)
...
+    protected function validateMimeType($file)
+    {
+        $fileMimeType = FileHelper::getMimeType($file->tempName);
+
+        foreach ($this->mimeTypes as $mimeType) {
+            if (strpos($mimeType, '*') !== false && $this->validateMimeTypeMask($fileMimeType, $mimeType)) {
+                return true;
+            }
+
+            if ($mimeType == $fileMimeType) {
```
samdark: Strict comparison could be safely used.

samdark: Since the condition is simpler than the one above, this one could be checked first.

samdark: Both conditions could be combined.

tests/framework/validators/FileValidatorTest.php
```diff
@@ -263,6 +264,23 @@ protected function createTestFiles($params = [])
         return $files;
     }
 
+    /**
+     * @param $fileName
```
samdark: Need docs.

framework/assets/yii.validation.js
kirsenn: Seems to be micro optimization as of access to `length` property complexity is O(1) and it's value stored in hashtable. If you still sure to replace with variable I do it

framework/assets/yii.validation.js
kirsenn: I think combining them could make worse code readability. I agree with second check is simpler, I'll replace then.

framework/assets/yii.validation.js
kirsenn: Okay

framework/validators/FileValidator.php
kirsenn: Okay

framework/validators/FileValidator.php
kirsenn: Okay

framework/validators/FileValidator.php
kirsenn: I'll replace with strict and change order

kirsenn:
@samdark Fixed, added docs

framework/assets/yii.validation.js
```diff
@@ -436,5 +436,32 @@ yii.validation = (function ($) {
...
+    function validateMimeTypeMask(mimeTypeMask, fileType)
+    {
+        var escapedString = escapeRegExp(mimeTypeMask);
+        var regexp = new RegExp('^' + escapedString.replace('\\*', '.*') +  '$');
+
+        return regexp.test(fileType);
+    }
+
+    function escapeRegExp(str) {
+        return (str+'').replace(/[.?*+^$[\]\\(){}|-]/g, "\\$&");
```
SilverFire: I'd prefer to escape the regexp on PHP side with `Html::escapeJsRegularExpression`

kirsenn: @SilverFire didn't get that, Client side doesn't receive regex from PHP side, it's generating it's own using mimeType mask

SilverFire: Sorry, forget it. I missed.

framework/assets/yii.validation.js
samdark: I see no reason not to do it.

framework/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.7 under development
 -----------------------
 
+- Enh #5469: Add mimetype validation by mask in FileValidator
```
samdark: Need you handle after the changelog line i.e. `(kirsenn)`. Also changelog line should be ordered by type and number.

kirsenn: >changelog line should be ordered by type and number.

didn't get that

kirsenn: >Enh #5469

samdark: We're soring changelog lines for each release by type and then by issue number.

kirsenn: Oh, sorry got it

samdark: i.e. it should be moved to all other lines with `Enh` type and should be right after issue with the number that's less than its own.

samdark:
Looks good. @SilverFire client side needs a bit more testing.

SilverFire:
Merged with adjustments: https://github.com/yiisoft/yii2/commit/7aef67187285ba43d8a07707d623c0bf8aef992b

SilverFire:
Thank you!

PR #10395, opened 2015-12-15
Added warning to session component initialization

### LINKED ISSUE #10255: Session component initialization over active session
Of course,  the best practice is use `Yii::$app->session` or cookie instead of handle with `$_COOKIE $_SESSION ...` directly.

but (yes, here is a but ) , if by any chance,  one call `session_start` before that called by `Yii`. 

for example, call `session_start` in the entry file or somewhere else ,

this will prevent the session component  to  `registerSessionHandler` `updateFlashCounters` .... 

then the session component no longer works well,

and in this case,  the bug is difficult to pull out.

so maybe we can give a detect at first `session_start` in `\yii\web\Session` ?

AnatolyRugalev:
Fixes #10255

### DIFF
--- a/framework/web/Session.php
+++ b/framework/web/Session.php
@@ -96,6 +96,7 @@ class Session extends Component implements \IteratorAggregate, \ArrayAccess, \Co
     public function init()
     {
         parent::init();
+        register_shutdown_function([$this, 'close']);
         if ($this->getIsActive()) {
             Yii::warning("Session is already started", __METHOD__);
         }
@@ -131,7 +132,6 @@ class Session extends Component implements \IteratorAggregate, \ArrayAccess, \Co
         if ($this->getIsActive()) {
             Yii::info('Session started', __METHOD__);
             $this->updateFlashCounters();
-            register_shutdown_function([$this, 'close']);
         } else {
             $error = error_get_last();
             $message = isset($error['message']) ? $error['message'] : 'Failed to start session.';
--- a/framework/web/Session.php
+++ b/framework/web/Session.php
@@ -96,7 +96,9 @@ class Session extends Component implements \IteratorAggregate, \ArrayAccess, \Co
     public function init()
     {
         parent::init();
-        register_shutdown_function([$this, 'close']);
+        if ($this->getIsActive()) {
+            Yii::warning("Session is already started", __METHOD__);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


framework/web/Session.php
```diff
@@ -129,6 +131,7 @@ public function open()
         if ($this->getIsActive()) {
             Yii::info('Session started', __METHOD__);
             $this->updateFlashCounters();
+            register_shutdown_function([$this, 'close']);
```
samdark: What if we call open multiple times?

AnatolyRugalev: In beginning of `open()`:
```php
if ($this->getIsActive()) {
    return;
}
```
So, nothing will happen

samdark: Right.

SilverFire:
Looks good

framework/web/Session.php
klimov-paul: It is not very good. In case you are opening and closing session multiple times during the single HTTP request process, such code will produce numerous shutdown functions registration.

SilverFire: Is it OK to open/close session multiple times per request?

klimov-paul: Yes it is OK. Such approach usually used if you are dealing with standard PHP session based on files to avoid blocking for the cuncurrent web requests.

samdark: But that means in this case there will be tons of shutdown functions registered.

klimov-paul: With this change - yes.
Original code was aware of it.

samdark: @AnatolyRugalev could you adjust code to take this case into account?

SilverFire: @klimov-paul you're right. I do the same :laughing:

AnatolyRugalev: Can we delete this handler on `close()`?

AnatolyRugalev: Ah, there is no unregister function in PHP, got it.
Should I return original code? Or use some kind of flag to check handler registration?

klimov-paul: Unlikely. There is no `unregister_shutdown_function()` PHP function to do so.

samdark: Moving it to init as it was, I think, is better.

cebe: yep, should be moved back to `init()`. `close()` will be called only once and will only take action if a session is still open.

SilverFire:
Merged, thank you

PR #10396, opened 2015-12-15
Support getters for virtual attributes

uaoleg:
In case I have virtual attributes (e.g. "photos" for album AR) I'd like to have validation for them. But I can't because getters won't work.

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -239,7 +239,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
         if (isset($this->_attributes[$name]) || array_key_exists($name, $this->_attributes)) {
             return $this->_attributes[$name];
         } elseif ($this->hasAttribute($name)) {
-            return null;
+            return parent::__get($name);
         } else {
             if (isset($this->_related[$name]) || array_key_exists($name, $this->_related)) {
                 return $this->_related[$name];


klimov-paul:
http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#data-transformation

uaoleg:
It's not what I mean. I'll try to explain it more detailed. So I have Albums AR. This AR has one-to-many relation named Photos. And I'd like to check it Album has enough number of photos. And I'd like to have a possibility to check it on separate basis. So I've extended `attributes` method and added virtual attribute `photos` and added custom validation on this attribute. But this breaks `photos` magic property, now it returns `NULL` instead of Photos collection.

SilverFire:
It's absolutely wrong idea. You can not have attribute `photos` and relation `photos`.

> And I'd like to check it Album has enough number of photos. And I'd like to have a possibility to check it on separate basis

```php
$album = Album::find()->with('photos')->where(['id' => 1])->one();
if (count($album->photos)) {
    // ...
}
```

Anyway, the PR will not be merged.

uaoleg:
Lol, I know how to count photos. I'm wondering how to reuse this code.

uaoleg:
Actually I come up with pretty acceptable solution without PR. But still I believe it make sense that getter is called if attribute wasn't mapped from DB.

```
    public function validate($attributeNames = null, $clearErrors = true)
    {
        parent::validate($attributeNames, $clearErrors);

        // Photos
        if (!$attributeNames || (is_array($attributeNames) && in_array('photos', $attributeNames))) {
            if ($this->getPhotos()->count() < static::PHOTOS_MIN_NUMBER) {
                $this->addError('photos', \yii::t('app', 'Please, upload more photos.'));
            }
        }

        return !$this->hasErrors();
    }
```

PR #10412, opened 2015-12-17
added prettyPrint option to JsonResponseFormatter

### LINKED ISSUE #9762: Json response formatting
When creating Restful apis, it would be great if we could have a good presentation, which can be read and understood easily. But the actual response has no formatting (indentation etc.) which makes it hard to read when we have lots of data.

It would be great if we could have a pretty print for json or simply use the "JSON_PRETTY_PRINT" in Json::encode().

Json::encode($response, JSON_PRETTY_PRINT);

Thanks

cebe:
fixes #9762


SilverFire:
:+1:

samdark:
:+1:

Issue #9880, opened 2015-10-10
\base\Security should use PHP's password_hash() by default


tom--:
password_hash() was introduced in PHP 5.5 and today I think the way Security::generatePasswordHash() and validatePassword() choose between that and using crypt() is silly. Better to simply prefer password_hash() when available and avoid the confusing $passwordHashStrategy config variable. Moreover, using crypt() is the default strategy at present.

Options

1. Change algorithms to use password_hash() when available and remove $passwordHashStrategy → Breaks BC.
2. Change algos, leave $passwordHashStrategy alone but ignore it → works, is BC, but unsavory, perhaps even dishonest.
3. Like option 2 but throw an exception with a specific message when an app sets $passwordHashStrategy → is BC [**update** actually it isn't BC but any dev that tests before deployment can easily correct config and be done] but how exactly to implement?
4. Introduce a third value $passwordHashStrategy = 'default', as new default, and this strategy prefers password_hash() when available over crypt() → BC but adds complexity rather than simplifying and making it easier for devs to understand

I think option 1 is a possibility for Yii 2.1. Options 3 and 4 could be done now.

samdark:
`3.` Could be done via check in init.

tom--:
I just updated option 3 because it isn't really BC. It's just not a big problem to upgrade.

@samdark if $passwordHashStrategy can't be changed after init() then I like this option rather a lot.

samdark:
Right. It can be changed..

tom--:
Another option

5. Keep current algorithms and change the default `$passwordHashStrategy = 'password_hash'` → Not BC for anyone on PHP 5.4 but easily fixed in config. Disadvantage: doesn't automatically get apps to use password_hash() when the upgrade from 5.4 happens.

samdark:
Password hashes produced by our current code (w/ any strategy) are compatible with PHP `password_hash`, right?

tom--:
yes, they are. afaik, they are two ways to access the same underlying Bcrypt lib.

samdark:
How about marking `$passwordHashStrategy` w/ `@deprecated` tag and not using it anymore? That's quite prominent in IDEs, is compatible in terms of resulting hash and won't break existing apps.

tom--:
I'm unfamiliar with the behavior of @deprecated in either IDEs or in the Yii online docs. So long as it is obvious then the objection to option 2 is overcome and I like it.

tom--:
[...]

PR #10421, opened 2015-12-17
\yii\base\Security updates

tom--:
This PR attempts to implement everything discussed in
- Use PHP 7 random_bytes() in \base\Security #9878 
- Stop using openssl_random_pseudo_bytes() #9879 
-  \base\Security should use PHP's password_hash() by default #9880 

It also adds LibreSSL as a safe alternative to OpenSSL on any platform and, for Windows only on PHP >= 5.4.0, keeps `openssl_random_pseudo_bytes()` after all because it doesn't actually use OpenSSL and goes to CryptGenRandom instead.

I also added type and range checks to `generateRandomString()`. We don't generally do a lot of input checking in Yii methods but in a security API like this I think it's wise. Otherwise invisible app bugs are unmasked by these checks.

Testing the new `generateRandomKey()` isn't simple. I temporarily instrumented `::$_randomSource` so I could see that the different branches were operating correctly and I tested with
-  5.4.45 
-  5.5.30 with and without mcrypt ext
-  5.6.16 with and without LibreSSL 2.3.1
-  7.0.0

But I don't know how to implement those tests permanently.

**I haven't tested on Windows.**

I could add tests for `generateRandomString()`'s type and range checks. But it's difficult for `generateRandomKey()` because in PHP 7, different exceptions are thrown. I haven't found a solution for that.


samdark:
Looks good to me. @yiisoft/core-developers need additional review.

samdark:
btw., run tests on Windows 10 with PHP 5.4.41, PHP 5.6.9, PHP 7.0.1. All passed.

tom--:
I'm pretty sure it's safe to ignore the second parameter of 

[`string openssl_random_pseudo_bytes ( int $length [, bool &$crypto_strong ] )`](https://secure.php.net/manual/en/function.openssl-random-pseudo-bytes.php) 

given the other tests we do, but I'd welcome any other opinion. Here's the function itself:

https://github.com/php/php-src/blob/master/ext/openssl/openssl.c#L5466-L5518

Can the function return a string *and* set `$crypto_strong` false.

tom--:
Is Scrutinizer right? Or is this alleged bug actually a Scrutinizer bug?

Scrutinizer

framework/base/Security.php
```diff
@@ -77,8 +77,10 @@ class Security extends Component
      * - 'password_hash' - use of PHP `password_hash()` function with PASSWORD_DEFAULT algorithm.
      *   This option is recommended, but it requires PHP version >= 5.5.0
      * - 'crypt' - use PHP `crypt()` function.
+     * @deprecated Since version 2.0.7, @generatePasswordHash() ignores @passwordHashStrategy and
```
SilverFire: I'm not sure, whether `@generatePasswordHash()` is a valid APIDOC reference. Shouldn't it be `[[generatePasswordHash()]]`?

SilverFire:
:+1:

samdark:
> Can the function return a string and set $crypto_strong false?

Nope based on current code. But you never know what would be implemented in the future so it's better to rely on method intended signature and check for it.

framework/base/Security.php
samdark: Yes, it should be like @SilverFire suggested.

samdark:
`$matches` thing would work as expected and Scritinizer basically failed but we may introduce nested ifs to make it more readable. That's not really critical though.

framework/CHANGELOG.md
```diff
@@ -90,6 +90,7 @@ Yii Framework 2 Change Log
 - Chg #9953: `TimestampBehavior::getValue()` changed to make value processing consistent with `AttributeBehavior::getValue()` (silverfire)
 - Chg #6419: Added `yii\web\ErrorHandler::displayVars` make list of displayed vars customizable. `$_ENV` and `$_SERVER` are not displayed by default anymore (silverfire)
 - Chg #9528: Traversable objects are now formatted as arrays in `yii\web\XmlResponseFormatter` to support SPL objects and Generators (MaXL-ru)
+- Chg #9878,9879,9880: Make `\base\Security` use `random_bytes()`, LibreSSL, mcrypt, limit OpenSSL to Windows, and to prefer `password_hash()` over `crypt()`
```
samdark: Need your handle at the end of the line i.e. `(tom--)`

framework/base/Security.php
```diff
@@ -507,6 +565,14 @@ public function generateRandomKey($length = 32)
      */
     public function generateRandomString($length = 32)
     {
+        if (!is_integer($length)) {
```
samdark: Better to use `is_int` since `is_integer` is an alias.

framework/base/Security.php
tom--: OK. There's another of these somewhere

framework/base/Security.php
tom--: OK

samdark:
Merged https://github.com/yiisoft/yii2/commit/1f41a2c9af106e70076b72991139d29f07a60a37. Huge thanks for taking care of it.

tom--:
Glad to be of help.

Although I am a bit nervous because it was a big change and there were few line comments. I'll see if I can poke anyone else into reviewing it.

samdark:
Yeah. The more people the better.

githubjeka:
@samdark somehow marked `@deprecated` in https://github.com/yiisoft/yii2/wiki/Ideas-for-2.1 or is checked by searching afterwards?

samdark:
,

githubjeka:
I asked about:
```
@deprecated Since version 2.0.7, [[generatePasswordHash()]] ignores [[passwordHashStrategy]] and
uses `password_hash()` when available or `crypt()` when not.

 -    public $passwordHashStrategy = 'crypt';
 +    public $passwordHashStrategy;
```

samdark:
It's marked in IDEs and would be removed afterwards.

githubjeka:
Ok.

P.S. @tom-- thanks for your contribute

tom--:
> This pull request is closed, but the `tom--:9878-9879-9880-base/Security-updates` branch has unmerged commits.

How to see what these changes are?

samdark:
That's composer.lock.

PR #10425, opened 2015-12-18
Enhancement to phpDoc of AssetManager

githubjeka:


samdark:
Thanks!

framework/web/AssetManager.php
```diff
@@ -549,7 +549,7 @@ protected function publishDirectory($src, $options)
      * This method does not perform any publishing. It merely tells you
      * if the file or directory is published, where it will go.
      * @param string $path directory or file path being published
-     * @return string the published file path. False if the file or directory does not exist
+     * @return string|false string the published file path. False if the file or directory does not exist
```
SilverFire: Shouldn't it be `boolean`?

githubjeka: In this case - no. 

Generally see [last example](http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.return.pkg.html) and [valid keywords](http://www.phpdoc.org/docs/latest/references/phpdoc/types.html#abnf).

PR #10429, opened 2015-12-18
[Alternative] Fix #9101 resolve appendTimestamp to registerJsFile and registerCssFile 

### LINKED ISSUE #9101: AssetManager appendTimestamp not working if you use registerJsFile()

githubjeka:
This just sketch, for [the discussion](https://github.com/yiisoft/yii2/pull/10297).


saada:
@githubjeka, this approach is a lot better than where I was heading IMO. Thank you!

githubjeka:
Thank you too.

githubjeka:
I cant merge with master because https://github.com/yiisoft/yii2/commit/01a6235059027c6ab692b86523118498f77055d7#commitcomment-15105103

SilverFire:
It seems, that the PR was overseen, sorry. Will try to verify it

SilverFire:
Sorry for a long delay. Merged: dc23a6d3d8a9cd937cbbc64652db903ee4ff5598 Thank you!

SilverFire:
Reverted 3ec3ce931216a44ec0b850c429c9d8c7f9a17e6d

sazo:
@SilverFire Its stands in the change log for 2.0.10 that #9101 is fixed but it is not. What happens now? is a new PR the way to go or do we fix this one?

cebe:
> @SilverFire Its stands in the change log for 2.0.10 that #9101 is fixed but it is not. What happens now? is a new PR the way to go or do we fix this one?

sorry, I have removed the changelog entry: 40394b8abf18e808253bed809a1df361cd7e3f76

It is questionable if View should depend on AssetManager in this way. If you want functionality of AssetManager it may be better to use an AssetBundle instead of registering the file directly.

PR #10433, opened 2015-12-19
Fixed #10390

### LINKED ISSUE #10390: Disable tag options in radiolist
Hi,
how can i do disable tag option in radiolist?
[radiolist](https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L987)
in my form i have status field and i want show this with toggle bottun and my code is:
```php
echo $form->field($model, 'status',
		['template'=>'{label}<div class="col-md-6 col-xs-9"><div class="btn-group" data-toggle="buttons">{input}</div> {hint} {error}</div>']
	)
	->radioList(
		[1=>'active', 0=>'deactive'],
		[
			'item' => function($index, $label, $name, $checked, $value) {
				return '<label class="btn btn-white btn-info btn-round">' . Html::radio($name, $checked, ['value'  => $value, 'autocomplete'=>'off']) . $label . '</label>';
			}
		]
	)
	->label(NULL, ['class'=>$labelClass]);
```
but in html show this:
```html
<div data-toggle="buttons" class="btn-group">
	<input type="hidden" value="" name="model[status]">
	<div id="model-status">
		<label class="btn btn-white btn-info btn-round">
			<input type="radio" autocomplete="off" value="1" name="model[status]">active
		</label>
		<label class="btn btn-white btn-info btn-round">
			<input type="radio" autocomplete="off" value="0" name="model[status]">deactive
		</label>
	</div>
</div>
```
how can fix it and don't show ```<div id="model-status"> ```?

TianJinRong:
Closes #10390


framework/helpers/BaseHtml.php
```diff
@@ -939,9 +940,9 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
      * - item: callable, a callback that can be used to customize the generation of the HTML code
      *   corresponding to a single item in $items. The signature of this callback must be:
      *
-     *   ```php
+     *   ~~~
```
SilverFire:
```
```php
```
 was correct

framework/helpers/BaseHtml.php
```diff
@@ -930,6 +930,7 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
      * The following options are specially handled:
      *
      * - tag: string, the tag name of the container element.
+     * - disableTag: boolean, whether to disable the container of the radio buttons. Defaults to false.
```
SilverFire: We don't need this option. It's better to pass `false` to `tag` option

framework/helpers/BaseHtml.php
```diff
@@ -981,10 +982,20 @@ public static function radioList($name, $selection = null, $items = [], $options
             $hidden = '';
         }
 
-        $tag = isset($options['tag']) ? $options['tag'] : 'div';
-        unset($options['tag'], $options['unselect'], $options['encode'], $options['separator'], $options['item'], $options['itemOptions']);
+        if (isset($options['disableTag']) && $options['disableTag'] == true) {
+            $tag = '';
+        } else {
+            $tag = isset($options['tag']) ? $options['tag'] : 'div';
+            $tag = json_encode($options['disableTag']);
```
SilverFire: Could you explain the idea of this line?

framework/helpers/BaseHtml.php
TianJinRong: Fixing.

SilverFire:
We also need to update PHPUnit tests

SilverFire:
Set your email in your local git settings https://help.github.com/articles/setting-your-email-in-git/

SilverFire:
We need a CHANGELOG line

TianJinRong:
Ok~

githubjeka:
@TianJinRong but you did not do it in the order as is required :) See https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md#5-update-the-changelog

framework/helpers/BaseHtml.php
```diff
@@ -929,7 +929,8 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
      * @param array $options options (name => config) for the radio button list container tag.
      * The following options are specially handled:
      *
-     * - tag: string, the tag name of the container element.
+     * - tag: string|boolean, the tag name of the container element. Disable the container of the
```
githubjeka: `tag: string|false` only `false` - not `true`. Otherwise misleading.

framework/helpers/BaseHtml.php
```diff
@@ -981,10 +982,19 @@ public static function radioList($name, $selection = null, $items = [], $options
             $hidden = '';
         }
 
-        $tag = isset($options['tag']) ? $options['tag'] : 'div';
+        if (isset($options['tag']) && $options['tag'] == false) {
```
githubjeka: Better `$options['tag'] === false`

githubjeka:
I sent PR with more clearer the code. https://github.com/TianJinRong/yii2/pull/1/files

githubjeka:
https://github.com/yiisoft/yii2/issues/10390

TianJinRong:
@githubjeka 
1. false === $tag is great;
2. For tag, it might be confused if we accept one type and a specified
vlaue instead of two types, so there we accept true and see it as
setting default value.

cebe:
> For tag, it might be confused if we accept one type and a specified
> value instead of two types, so there we accept true and see it as
> setting default value.

it is a general convention used in Yii to accept some values or bool false to disable it. Introducing true here will be inconsistent.

TianJinRong:
@cebe optimize for the hidden.

framework/helpers/BaseHtml.php
```diff
@@ -975,16 +980,14 @@ public static function radioList($name, $selection = null, $items = [], $options
...
         }
+        $visibleContent = implode($separator, $lines);
 
-        if (isset($options['unselect'])) {
-            // add a hidden field so that if the list box has no option being selected, it still submits a value
-            $hidden = static::hiddenInput($name, $options['unselect']);
-            unset($options['unselect']);
-        } else {
-            $hidden = '';
+        if (false === $tag || 'false' === $tag) {
```
SilverFire: We usually write the variable first: `$tag === false`

Also I don't think that `fasle` as a string should do the same as a boolean `false`.

framework/helpers/BaseHtml.php
```diff
@@ -975,16 +980,14 @@ public static function radioList($name, $selection = null, $items = [], $options
...
+            return $hidden . $visibleContent;
         }
 
-        return $hidden . static::tag($tag, implode($separator, $lines), $options);
+        // returned the result with visible content contained by container.
```
SilverFire: I think that the line below is obvious and does not need any comments

framework/helpers/BaseHtml.php
```diff
@@ -953,11 +954,15 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
      */
     public static function radioList($name, $selection = null, $items = [], $options = [])
     {
+        // read options
```
SilverFire: I think that the lines below is obvious and does not need any comments

framework/helpers/BaseHtml.php
```diff
@@ -929,7 +929,8 @@ public static function checkboxList($name, $selection = null, $items = [], $opti
      * @param array $options options (name => config) for the radio button list container tag.
      * The following options are specially handled:
      *
-     * - tag: string, the tag name of the container element.
+     * - tag: string|false, the tag name of the container element. Disable the container of the
```
githubjeka: - tag: string|false, string is the tag name of the container element. False to render radio buttons without container.

githubjeka:
:+1:

TianJinRong:
@githubjeka  :)

framework/helpers/BaseHtml.php
```diff
@@ -975,16 +979,13 @@ public static function radioList($name, $selection = null, $items = [], $options
...
         }
+        $visibleContent = implode($separator, $lines);
 
-        if (isset($options['unselect'])) {
-            // add a hidden field so that if the list box has no option being selected, it still submits a value
-            $hidden = static::hiddenInput($name, $options['unselect']);
-            unset($options['unselect']);
-        } else {
-            $hidden = '';
+        if (false === $tag) {
```
SilverFire: And i'm still voting for `$tag === false`

TianJinRong: Well, i think it doesn`t matter between two scripts if there are no more reasons for deciding which one is better :)

samdark: Yep. There's no sense in doing reverse check in PHP. It's a thing from Java and this way people are avoiding null pointer exceptions.

rugabarbo: This way some developers are avoiding erroneous use `=` instead of `==`|`===`, but I am against this trick.

samdark: It doesn't look natural to me.

TianJinRong: Well, it worth to modify if the script looks wierd.

SilverFire:
Merged. I've also added the same feature to `checkboxList()` method. https://github.com/yiisoft/yii2/commit/85c6a903dced72cce0ce95f8fd424fd492ceb4e6

SilverFire:
Thank you!

TianJinRong:
@SilverFire :-)

PR #10434, opened 2015-12-19
#8466: FileValidator - fixed tooBig and tooSmall error messages

### LINKED ISSUE #8466: Translate messages with byte size
I've struggled into some ugly things with the ICU translation.

The original error message for uploaded file which are too big is:
```
The file "{file}" is too big. Its size cannot exceed {limit, number} {limit, plural, one{byte} other{bytes}}.
```
Which results at 20MB in:
```
The file "xyz" is too big. Its size cannot exceed 20971520 bytes.
```
This message is not that user friendly, because nobody knows how much 20971520 bytes really is.

My current workaround is this:
```php
public function rules()
{
	return [
		['attachments', 'file', 'maxFiles' => 10, 'maxSize' => 20 * 1024 * 1024,
			'extensions' => ['png', 'jpg', 'gif', 'tif', 'doc', 'docx', 'pdf'],
			'tooBig' => Yii::t('app', 'The file "{file}" is too big. Its size cannot exceed {limitSize}', [
				'limitSize' => Yii::$app->formatter->asShortSize(20 * 1000 * 1000)])
		]
	];
}
```

SilverFire:
Closes #8466 

I've added additional option `formattedLimit` for `tooBig` and `tooSmall` error messages.
Then I translated almost all languages, where it was obviously easy.

I found inconsistency in `maxSize` check condition, also fixed.

Need a review.


framework/validators/FileValidator.php
```diff
@@ -98,6 +100,8 @@ class FileValidator extends Validator
      * - {attribute}: the attribute name
      * - {file}: the uploaded file name
      * - {limit}: the value of [[minSize]]
```
samdark: `{limit}` isn't used, right?

samdark: Ah, it kept for BC. OK.

samdark:
:+1:

cebe:
looks great.

PR #10435, opened 2015-12-20
fixed reading /dev/urandom loop in yii\base\Security

cebe:
see https://github.com/yiisoft/yii2/commit/1f41a2c9af106e70076b72991139d29f07a60a37#commitcomment-15083428

@tom-- @yiisoft/core-developers need review.

### DIFF
--- a/framework/base/Security.php
+++ b/framework/base/Security.php
@@ -521,14 +521,16 @@ class Security extends Component
             if ($urandomFile !== null) {
                 // $length could be large so read using a loop.
                 $key = '';
+                $bytesLeft = $length;
                 do {
-                    $buffer = fread($urandomFile, $length);
+                    $buffer = fread($urandomFile, min($bytesLeft, 8092));
                     if (!$buffer) {
                         $key = null;
                         break;
                     }
                     $key .= $buffer;
-                } while (StringHelper::byteLength($key) < $length);
+                    $bytesLeft = $length - StringHelper::byteLength($key);
+                } while ($bytesLeft > 0);
 
                 if ($key !== null && StringHelper::byteLength($key) === $length) {
                     $this->_randomSource = $urandomFile;
--- a/tests/framework/base/SecurityTest.php
+++ b/tests/framework/base/SecurityTest.php
@@ -810,6 +810,28 @@ TEXT;
         $this->assertTrue($key1 != $key2);
     }
 
+    public function testGenerateRandomKeyURandom()
+    {
+        if (function_exists('random_bytes')) {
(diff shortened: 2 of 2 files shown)


framework/base/Security.php
```diff
@@ -521,14 +521,16 @@ public function generateRandomKey($length = 32)
             if ($urandomFile !== null) {
                 // $length could be large so read using a loop.
                 $key = '';
+                $bytesLeft = $length;
                 do {
-                    $buffer = fread($urandomFile, $length);
+                    $buffer = fread($urandomFile, min($bytesLeft, 8092));
```
samdark: Why 8092?

cebe: see "chunk size" in http://php.net/manual/en/function.fread.php
but we could also keep `$bytesLeft` here only.

tom--: @cebe Yes, you're right. Perhaps this is why I had the nagging feeling that I had forgotten something.

samdark:
:+1:

framework/base/Security.php
tom--: actually it should be  8192 not  8092

tests/framework/base/SecurityTest.php
```diff
@@ -810,6 +810,28 @@ public function testGenerateRandomKey()
         $this->assertTrue($key1 != $key2);
     }
 
+    public function testGenerateRandomKeyURandom()
+    {
+        if (function_exists('random_bytes')) {
+            $this->markTestSkipped('This test can only work on platforms where random_bytes() function does not exist. You may disable it in php.ini for testing. http://php.net/manual/en/ini.core.php#ini.disable-functions');
```
tom--: thanks for pointing out ini.disable-functions. it actually provides a way to automatically test more of the branches. i will try it out in a separate branch

cebe: automated testing still does not work, you can not use ini_set for it.

framework/base/Security.php
cebe: right, thanks @tom--.

How do we proceed here?

This will read in chunks and is covered by the test:
```php
$buffer = fread($urandomFile, min($bytesLeft, 8192));
```

This will read in nearly all cases in on chunk but there is no test to verify it can handle the case where multiple chunks are read:
```php
$buffer = fread($urandomFile, $bytesLeft);
```
Still we have now reviewed it and know it works. which one is better?

performance is no difference as far as I see (tested with reading 50MB).

tom--:
@cebe how do we go from here? first, did you review [my alternative implementation](https://github.com/yiisoft/yii2/pull/10440). i prefer it a bit although i'm not sure about the pros and cons.

cebe:
@tom-- I just saw it, there was no link from here so I found it later. tend to prefer it.

Issue #9578, opened 2015-09-02
Add limit 1 to improve performance of BaseActiveRecord::findOne()


hightman:
I have noticed manual annotation, but in `BaseActiveRecord::findOne()` there is no chance to add `limit (1)` unless override this method.

Lots of time we can not construct an unqiue condition (such as just get the first record after sorting), it cause performance problem when granted call `findOne()`.

There is no reason to refuse such modifications.
```
    public static function findOne($condition)
    {    
        return static::findByCondition($condition)->limit(1)->one();
    } 
```

Further, we can optimize `Query::one()` as following.
```
    public function one($db = null)
    {
        $oldLimit = $this->limit;
        $result = $this->limit(1)->createCommand($db)->queryOne();
        $this->limit = $oldLimit;
        return $result;
    } 
```

klimov-paul:
Duplicates #4348

hightman:
Hello, my main point is to add `limit(1)` within `BaseActiveRecord::findOne()`, rather than the `Query::one()`.

Could you look into my content, completely unable to add a `limit (1)` to `findOne()`, it is a static method, unless overridden method.  This is entirely a trick.

@qiangxue @klimov-paul

cebe:
findOne() is a shortcut for the very simple case. If you need anything more, you should use ::find()->limit(1)->one()

hightman:
@cebe I certainly know that we can do so, I accept the implementation of `Query::one()` implementation, but it is difficult to accept not add limit(1) to `ActiveRecord::findOne()`.

hightman:
And, I really do not know what harm with `limit(1)` in `findOne()`, internal `Query` object will not be reused forever.

PR #10438, opened 2015-12-20
Add limit 1 to hasOne

ameen-sarsour:
We notify there is no limit on hasOne function, so We solved it by call limit(1) in our relation function. 
But I think it's better to add it in core framewoek


cebe:
duplicate of #9578

PR #10440, opened 2015-12-20
fix #10435 with file get contents(/dev/urandom)

tom--:
let's consider this alternative too (compare to #10435).

Using `fopen()` and `fread()` allows us to cache file pointer. But that's not necessarily useful.

`file_get_contents()` has two advantages:
- simpler code, easier to understand, less likely to have bugs—**as already demonstrated**
- "file_get_contents() is the preferred way to read the contents of a file into a string. It will use memory mapping techniques if supported by your OS to enhance performance."

I'm not really comfortable with the `fread()` code.


framework/base/Security.php
```diff
@@ -30,6 +30,11 @@
  */
 class Security extends Component
 {
+    const DEV_URANDOM = '/dev/urandom';
```
samdark: These are internal things useless to outside code.

tom--: i know. but i have been chided often for using "magic numbers". they are considered bad practice.

http://refactoring.com/catalog/replaceMagicNumberWithSymbolicConstant.html

for example, in this file we '/dev/urandom' four times. with a constant, the static inspections ensure we don't have a typo in any of them.

samdark: Yes, magic numbers are bad because both they're error prone and not descriptive. Named constants solve both issue. Too bad PHP has no private constants.

tom--: [coming in 7.1](https://wiki.php.net/rfc/class_const_visibility)

Iiuc, an undocumented class const doesn't appear in the Yii web docs, which is close enough for now. Probably even better if they are not located at the top of the class, where PhpStorm automatically put these.

cebe: > if they are not located at the top of the class, where PhpStorm automatically put these.

lets put them above `generateRandomKey` as we do with private properties.

framework/base/Security.php
```diff
@@ -493,48 +498,21 @@ public function generateRandomKey($length = 32)
...
-                    }
-                } else {
-                    $urandomFile = null;
-                }
-            }
+        // If not on Windows, test for a /dev/urandom device.
+        if ($this->_randomSource === null && DIRECTORY_SEPARATOR === '/') {
+            // Check it for speacial character device protection mode.
+            $lstat = @lstat(self::DEV_URANDOM);
+            $urandomDevice = $lstat !== false && ($lstat['mode'] & 0170000) === 020000;
```
cebe: have you evaluated [stat](https://secure.php.net/manual/en/function.stat.php) vs. [lstat](https://secure.php.net/manual/en/function.lstat.php)?

>  lstat() is identical to stat() except it would instead be based off the symlinks status.

cebe: Am I right, that if `/dev/urandom` is a symlink, we will not accept it regardsless of where it points when using `lstat`?

tom--: That was the idea, yes.

tom--: Should probably add a comment to explain it.

cebe: > Should probably add a comment to explain it.

:+1:

PR #10453, opened 2015-12-21
Dieharder randomness tests for Security::generateRandomKey()

tom--:
No issue for this PR. 

This adds one unit tests running ten of the [dieharder](http://www.phy.duke.edu/~rgb/General/dieharder.php) randomness tests. It is skipped if the dieharder program isn't on the test machine. There are Homebrew and Ubuntu dieharder packages.

Works on OS X with:
- PHP 7.0.0  → should use `random_bytes()`
- PHP 5.6.16 LibreSSL  + mcrypt → should use the first `openssl_random_pseudo_bytes()`
- PHP 5.4.45 OpenSSL + mcrypt → should use `mycrypt_create_iv()`
- PHP 5.4.45 OpenSSL - mcrypt → should use `file_get_contents('/dev/urandom')`

The second `openssl_random_pseudo_bytes()` only runs on Windows.


samdark:
No Windows binaries? :)

tom--:
Marsaglia's original diehard came with binaries for windows and dos. Try compiling from sources. Testing on Windows is really valuable. 

I'll keep looking.

samdark:
There are more tests like that such as NIST one: https://www3.cs.stonybrook.edu/~algorith/implement/rng/implement.shtml

tom--:
I believe the STS tests are included in diehard, and the STS Monobit test is `-d 100` in this unit test.

tom--:
@samdark To test the Windows generator paths (`random_bytes()`, mcrypt and any openssl) you just need to generate the temp random file on Windows. You could then move it to a Unixish system and run diehard on it.

cebe:
We should add a script that installs dieharder to travis.yml.

cebe:
Hm... it depends on `libgsl0-dev` so if we want it on travis we need to get that installed somehow.

tom--:
Bear in mind that it is intensive – about 1 min of core time. It's not the kind of thing that needs to be run after every commit but it would be good to run it before every release.

These are the configurations to test on all OSs

- PHP 7
- PHP 5 LibreSSL ~~OpenSSL~~ mcrypt
- PHP 5 ~~LibreSSL~~ OpenSSL mcrypt
- PHP 5 ~~LibreSSL~~ OpenSSL ~~mcrypt~~

dieharder examines RNG about as closely as anyone knows how. And different OSs have different RNGs. I don't think it's our job to test across Linux, Solaris, FreeBSD, NetBSD, OpenBSD and so forth but bear it in mind that there could be surprises.

PR #9466, opened 2015-08-19
add unsigned to the new migrations

berezuev:
resolves #9443

PR #9441, opened 2015-08-17
Comment methods for schema builder

vaseninm:
Hi. I start work with issue #8148. 
I add 4 methods for Migration:
- addCommentOnColumn($table, $column, $comment)
- dropCommentFromColumn($table, $column)
- addCommentOnTable($table, $comment)
- dropCommentFromTable($table)

and one method to ColumnSchemaBuilder:
- comment($comment)

Now i add support for 
- mysql (all methods)
- postgresql (Migration methods and empty for column string, but if we create method in migrate controller - we add comments on each column)
- sqllite (NotSupportedException for all)

I hope for your code review... 

Tested with DBMS:
- [x] MySQL
- [x] MSSQL
- [x] PostgreSQL
- [x] SQLite
- [x] CUBRID
- [x] Oracle

PR #10463, opened 2015-12-23
Migration - added comment(), unsigned() methods. Supports only [MySQL]

skhdev:
1. All integer types can have an optional (nonstandard) attribute UNSIGNED. Unsigned type can be used to permit only nonnegative numbers in a column or when you need a larger upper numeric range for the column. For example, if an INT column is UNSIGNED, the size of the column's range is the same but its endpoints shift from -2147483648 and 2147483647 up to 0 and 4294967295.
2. Also added comment() method, more useful.

How to use:

$this->createTable('table', [
            'id' => $this->primaryKey(),
            'some_column' => $this->integer()->unsigned()->comment('Comment will be here.'),
            ...
        ]);

### DIFF
--- a/framework/db/ColumnSchemaBuilder.php
+++ b/framework/db/ColumnSchemaBuilder.php
@@ -18,6 +18,11 @@ use yii\base\Object;
  * @author Vasenin Matvey <vaseninm@gmail.com>
  * @since 2.0.6
  */
+
+/**
+ * @method \yii\db\mysql\ColumnSchemaBuilder unsigned()  supports only [MySQL]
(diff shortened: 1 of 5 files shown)


SilverFire:
Thank you for your effort, but we can't accept it as is because it does not cover all DBMSs.

We already have PRs for [comments](#9441) and [unsigned type](#9466).
Comments PR is almost ready to be merged and covers all the supported DBMSs. Now we need volunteers to verify it work on MSSQL, Curbid and Oracle.

Unsigned is not so good yet, but we still need it. If you want take care of this problem and implement support for all the DBMSs - we will be thankful.

P.S.: 1 feature = 1 branch = 1 PR. Don't try to put all at once.

Issue #10467, opened 2015-12-23
[\yii\di\Instance] Something went wrong


LAV45:
It seems there is a problem with the documentation and method implementation.
```php
Instance::ensure('db');
```
![oops](https://cloud.githubusercontent.com/assets/675367/11985734/ec63a44c-a9d8-11e5-9f7e-a1d6e889d3ce.png)
- https://github.com/yiisoft/yii2/blob/master/framework/di/Instance.php#L110
- https://github.com/yiisoft/yii2/blob/master/framework/di/Instance.php#L129

In [unit tests](https://github.com/yiisoft/yii2/blob/master/tests/framework/di/InstanceTest.php) there is no validation on the use case

SilverFire:
It makes no sense to call `ensure` without the second argument.
You MUST pass an object or a class name to ensure that `db` in `Yii::$app` is instance of second argument.

LAV45:
Can I something not so understood
```php
/*
* ...
* @param string $type the class/interface name to be checked. If null, type check will not be performed.
* ...
*/
public static function ensure($reference, $type = null, $container = null)
```

SilverFire:
Well, it seems I was wrong, sorry.
Let's move to the PR

SilverFire:
Thank you for digging in the problem and a good fix! https://github.com/yiisoft/yii2/commit/fcf25e8fa1e8070abe9df21c89a35f245a338cd4

Happy NY :)

PR #10468, opened 2015-12-23
Fix \yii\di\Instance::ensure()

lav45:
Bug is described here https://github.com/yiisoft/yii2/issues/10467

### DIFF
--- a/framework/di/Instance.php
+++ b/framework/di/Instance.php
@@ -107,7 +107,9 @@ class Instance
      */
     public static function ensure($reference, $type = null, $container = null)
     {
-        if ($reference instanceof $type) {
+        if (is_object($reference) && $type === null) {
+            return $reference;
+        } elseif (is_object($reference) && $reference instanceof $type) {
             return $reference;
         } elseif (is_array($reference)) {
             $class = isset($reference['class']) ? $reference['class'] : $type;
@@ -126,7 +128,9 @@ class Instance
 
         if ($reference instanceof self) {
             $component = $reference->get($container);
-            if ($component instanceof $type || $type === null) {
+            if ($type === null) {
+                return $component;
+            } elseif ($component instanceof $type) {
                 return $component;
             } else {
                 throw new InvalidConfigException('"' . $reference->id . '" refers to a ' . get_class($component) . " component. $type is expected.");
--- a/framework/di/Instance.php
+++ b/framework/di/Instance.php
@@ -107,11 +107,7 @@ class Instance
      */
     public static function ensure($reference, $type = null, $container = null)
     {
-        if (is_object($reference) && $type === null) {
-            return $reference;
-        } elseif (is_object($reference) && $reference instanceof $type) {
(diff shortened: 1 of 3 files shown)


tests/framework/di/InstanceTest.php
```diff
@@ -41,9 +43,81 @@ public function testEnsure()
...
+
+    public function testCheckException()
+    {
+        $container = new Container;
+        $container->set('db', [
+            'class' => 'yii\db\Connection',
+            'dsn' => 'test',
+        ]);
+
+        try {
```
SilverFire: It should be done with separated test cases and `$this->setExpectedException`, because current test will pass successfully when no exception was thrown.

LAV45:
@SilverFire something else?

framework/di/Instance.php
```diff
@@ -107,7 +107,9 @@ public static function of($id)
      */
     public static function ensure($reference, $type = null, $container = null)
     {
-        if ($reference instanceof $type) {
+        if (is_object($reference) && $type === null) {
+            return $reference;
+        } elseif (is_object($reference) && $reference instanceof $type) {
```
samdark: There's no need for `is_object($reference)`. `$reference instanceof $type` is enough.

SilverFire:
It's a very global change so I'd like to see thoughts of @yiisoft/core-developers

framework/di/Instance.php
```diff
@@ -107,7 +107,9 @@ public static function of($id)
      */
     public static function ensure($reference, $type = null, $container = null)
     {
-        if ($reference instanceof $type) {
+        if (is_object($reference) && $type === null) {
```
samdark: These could be switched around because null check is cheaper.

```php
if ($type === null && is_object($reference)) {
```

framework/di/Instance.php
```diff
@@ -126,7 +128,9 @@ public static function ensure($reference, $type = null, $container = null)
 
         if ($reference instanceof self) {
             $component = $reference->get($container);
-            if ($component instanceof $type || $type === null) {
```
samdark: No need to change it. It's the same as it was.

LAV45: If `$type = null;` then instanceof return Error 
I wrote about this in https://github.com/yiisoft/yii2/issues/10467

framework/di/Instance.php
LAV45: if you remove the is_object leads to error `Error: Class name must be a valid object or a string`
so we checked is_string below, then here you need to check is_object

framework/di/Instance.php
samdark: OK.

cebe:
looks good to me :+1:

SilverFire:
Merged with squash https://github.com/yiisoft/yii2/commit/fcf25e8fa1e8070abe9df21c89a35f245a338cd4

SilverFire:
Thank you!

PR #10473, opened 2015-12-24
fix BaseArrayHelper::getValue() to work with '' (empty string) keys

sharpensteel:
fix `BaseArrayHelper::getValue()` to work with '' (empty string) keys after "."

text example:
    `\yii\helpers\BaseArrayHelper::getValue( [ 'key1' => [ '' => 'val2' ] ], 'key1.' )`
should return 'val2'


samdark:
Would be good to add a unit test for it.

SilverFire:
The proposed behavior is inconsistent with `ArrayHelper::getValue('')` and can not be accepted

PR #10475, opened 2015-12-24
Making internal UrlManager cache protected(instead of private) to all…

dmdark:
There is an rare issue with UrlManager internal private cache. For example we have this rule configuration:

``` php
 'urlManager' => [
    'class' => 'yii\web\UrlManager',
    'enablePrettyUrl' => true,
    'showScriptName' => false,
    'hostInfo' => 'http://www.test.dev',
    'baseUrl' => '/',
    'suffix' => '/',
    'rules' => [
        [
            'pattern' => '/special-cities/<city:(moscow|kazan|barcelona)+>',
            'route' => 'company/index',
        ],
        [
            'pattern' => '<city:[\w-]+>/cities',
            'route' => 'company/index'
        ],

    ]
```

We have special rule for 3 cities and the second universal rule. If we start generating our urls with URL that matched by second rule (for example "London"), than ALL other cities, including cities from the first rule we by matched by second rule (NOT the first one). That is how internal cache work - it just saves the first matched rule in private property and than use it for same routes.

For example:

``` php
  echo Url::to(['company/index', 'city' => 'kazan'], true) . "\n";
  echo Url::to(['company/index', 'city' => 'moscow'], true) . "\n";
  echo Url::to(['company/index', 'city' => 'london'], true) . "\n";
```

will be: 
http://www.test.dev/special-cities/kazan/
http://www.test.dev/special-cities/moscow/
http://www.test.dev/london/cities/

**Its correct!** But if we use:

``` php
echo Url::to(['company/index', 'city' => 'london'], true) . "\n";
echo Url::to(['company/index', 'city' => 'kazan'], true) . "\n";
echo Url::to(['company/index', 'city' => 'moscow'], true) . "\n";
```

the result will be:
_http://www.test.dev/london/cities/
http://www.test.dev/kazan/cities/
http://www.test.dev/moscow/cities/_

Its **incorrect** because of internal cache.
### Fix

We can implement some public settings. Like `$useCache` in UrlRule or `$useInternalCache` in UrlManager. But the case is pretty rare and I think it is not it worth it. Better just to make cache _overridable_ (protected instead of private) so the developers could override it or turn it off. 

So I made it in this pull request.

### DIFF
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -312,28 +312,19 @@ class UrlManager extends Component
         if ($this->enablePrettyUrl) {
             $cacheKey = $route . '?' . implode('&', array_keys($params));
 
-            /* @var $rule UrlRule */
-            $url = false;
-            if (isset($this->_ruleCache[$cacheKey])) {
-                foreach ($this->_ruleCache[$cacheKey] as $rule) {
-                    if (($url = $rule->createUrl($this, $route, $params)) !== false) {
-                        break;
-                    }
-                }
-            } else {
-                $this->_ruleCache[$cacheKey] = [];
-            }
+            $url = $this->getUrlFromInternalCache($cacheKey, $route, $params);
 
             if ($url === false) {
                 $cacheable = true;
                 foreach ($this->rules as $rule) {
+                    /* @var $rule UrlRule */
                     if (!empty($rule->defaults) && $rule->mode !== UrlRule::PARSING_ONLY) {
                         // if there is a rule with default values involved, the matching result may not be cached
                         $cacheable = false;
                     }
                     if (($url = $rule->createUrl($this, $route, $params)) !== false) {
                         if ($cacheable) {
-                            $this->_ruleCache[$cacheKey][] = $rule;
+                            $this->setRuleToInternalCache($cacheKey, $rule);
                         }
                         break;
                     }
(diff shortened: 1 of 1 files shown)


samdark:
I agree since I see no straightforward way to implement it without causing drawbacks.

framework/web/UrlManager.php
```diff
@@ -126,11 +126,14 @@ class UrlManager extends Component
      */
     public $ruleConfig = ['class' => 'yii\web\UrlRule'];
 
+    /**
+     * @var array internal cache to store matched rules
+     */
+    protected $ruleCache;
```
cebe: why change this to protected if we now have methods to access it?

dmdark: I think we want to access $ruleCache inside subclasses. For example if we want to use parent cache logic but with some improvements (for example calculating cache hits and misses):

```php
class UrlManagerWithCacheCount extends UrlManager
{
    private $cacheHit = 0;
    private $cacheMiss = 0;
    
    protected function getUrlFromInternalCache($cacheKey, $route, $params)
    {
        if (!empty($this->ruleCache[$cacheKey])) {
            foreach ($this->ruleCache[$cacheKey] as $rule) {
                /* @var $rule UrlRule */
                if (($url = $rule->createUrl($this, $route, $params)) !== false) {
                    $this->cacheHit++;
                    return $url;
                }
            }
        } else {
            $this->ruleCache[$cacheKey] = [];
        }
        $this->cacheMiss++;
        
        return false;
    }
}
```

Otherwise we would have to make another private property. That why I see $ruleCache protected. Dont you agree?

kidol: In this case you can simply override the method:
```php
$url = parent::getUrlFromInternalCache(...);
if ($url === false)  {
   // miss
} else {
   // hit
}
return $url;
```

dmdark: You are right, my example not helping to prove my opinion. But what if we want to count hits for specific rule. 
OR
we want to store only one rule, instead of array. 
 
```$this->ruleCache[$cacheKey] = $rule;```
instead of
```$this->ruleCache[$cacheKey][] = $rule;```

Well, maybe that cases are rare but can happen. I am not really against making it private but I think it can become sometimes a little pain for developers.

Comment (kidol) [inline framework/web/UrlManager.php]: Check out this article: http://fabien.potencier.org/pragmatism-over-theory-protected-vs-private.html
>Having a few well defined extension points force the developer to extend your library the right way instead of hacking your code.

In general, refactoring to allow access to previously private properties has been done many times in the past for Yii. If it makes sense and is not a very rare use case it will be implemented if requested.

Comment (dmdark) [inline framework/web/UrlManager.php]: You are right. Thanks for article. Made it private.

File: framework/web/UrlManager.php [hunk 2]
```diff
@@ -371,6 +362,39 @@ public function createUrl($params)
     }
 
     /**
+     * Geting url from internal cache if exists
+     * @param string $cacheKey generated cache key to store data.
+     * @param string $route the route (e.g. `site/index`).
+     * @param array $params rule params.
+     * @return bool|string the created URL
+     * @see createUrl()
+     */
```
cebe: need `@since 2.0.8` annotation here and also on the other method.

dmdark: done

PR #10476, opened 2015-12-24
ArrayDataProvider add options `sliceModels`

lav45:
This option allows you to build proper paging when getting data from API

My example used

``` php
$pageSize = 20;
$page = Yii::$app->request->get('page', 1);

// request to the Yii2 rest api server
$response = Yii::$app->curl->get($this->api_url, [
    'per-page' => $pageSize,
    'page' => $page,
]);

$dataProvider = new ArrayDataProvider([
    'sliceModels' => false,
    'allModels' => Json::decode($response->body),
    'totalCount' => $response->headers['X-Pagination-Total-Count'],
    'pagination' => [
        'pageSize' => $pageSize,
    ],
]);

    'dataProvider' => $dataProvider
]);
```

### DIFF
--- a/framework/data/ArrayDataProvider.php
+++ b/framework/data/ArrayDataProvider.php
@@ -64,8 +64,10 @@ class ArrayDataProvider extends BaseDataProvider
      */
     public $allModels;
     /**
-     * @var bool If the $allModels need is already part of the data set the value to false
-     * then [[prepareModels]] return all data from the $allModels
+     * @var boolean whether we need to slice [[allModels]] for pagination. Set to `false`,
+     * if [[allModels]] are already paginated. Works only when [[pagination]] is set. Defaults to `true`.
+     * @see pagination
+     * @since 2.0.7
      */
     public $sliceModels = true;
 
--- a/framework/data/ArrayDataProvider.php
+++ b/framework/data/ArrayDataProvider.php
@@ -63,6 +63,11 @@ class ArrayDataProvider extends BaseDataProvider
      * The array elements must use zero-based integer keys.
      */
     public $allModels;
+    /**
+     * @var bool If the $allModels need is already part of the data set the value to false
+     * then [[prepareModels]] return all data from the $allModels
+     */
+    public $sliceModels = true;
 
 
     /**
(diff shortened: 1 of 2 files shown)


framework/data/ArrayDataProvider.php
```diff
@@ -63,6 +63,11 @@ class ArrayDataProvider extends BaseDataProvider
      * The array elements must use zero-based integer keys.
      */
     public $allModels;
+    /**
+     * @var bool If the $allModels need is already part of the data set the value to false
```
SilverFire: It took much time for me to understand, what does it do. I suggest to change description to

```
@var boolean whether we need to slice [[allModels]] for pagination. Set to `false`, if [[allModels]] are already paginated. Works only when [[pagination]] is set. Defaults to `true`.
```

SilverFire:
The use case is ok. We need unit-tests for the feature

klimov-paul:
> The use case is valid.

Disagreed. For me this is missuse of the `ArrayDataProvider`. If you wish to handle REST API pagination, you should craete your own data provider class, which should encapsultate all related logic, including resposnse parsing.

LAV45:
@klimov-paul, I will be glad to see your solution to this problem, and while this is the easiest and fastest way to correctly display the list of data received from an external source.

klimov-paul:
> But if in place RestAPI will use the set of files then I need another own data provider class?

Of course. Thus we have `ArrayDataProvider`, `SqlDataProvider` and `ActiveDataProvier`.
Usage of the `ArrayDataProvider` in your example is not correct twice: it performs sorting data via PHP array functions, while sorting of the REST API respnse data requires request adjusting.

The actual purpose of the  data provider is providing end up data processing like pagination or sorting. If pagination preformed elsewhere - there is no need for data provider at all!

I assume you need this 'hack' to be able working with widgets like `GridView`.
As an alternative solution, we can create yet another data provider called something like `DummyDataProvider` or `StaticDataProvider`, which will not perform actual pagination or sorting at all.

> This is the easiest and fastest way to correctly display the list of data received from an external source.

This is lazy one's way, which leads to inconsistent code. You have to write REST-pagination code anyway - why you dislike the idea to write it inside dedicated data provider class so much, and prefer to leave it somewhere else?
Look at `SqlDataProvider` or `ActiveDataProvier`. It also would be easier to use your approach in those cases, but whould it be better?

For the ispiration you can see following:
https://github.com/yiisoft/yii2-sphinx/blob/master/ActiveDataProvider.php
https://github.com/yiisoft/yii2-elasticsearch/pull/16

LAV45:
@klimov-paul, you are right, this is a fairly hardcore option that came to solve a specific problem, but is hardly suitable for everyone. Thank you for your detailed reply.

PR #10478, opened 2015-12-25
safe log migration

vitalyzhakov:
I have error in my config, after applying log migration DB structure changes, but migration not applying.
Next migration attempt crushed.

I think, log migration must be safe.


SilverFire:
Not all the DBMS support DDL in transaction, so it will not work as desired.

PR #10480, opened 2015-12-25
Update User.php to cleanup identity cookie

ghost:
If "login" is called in "User.php" with a new identity and $duration = 0 while a user is already logged in with $duration > 0, then the current user's identity cookie is not cleaned up.

I would like to implement an "impersonate user" function what allows an admin to be logged in as any user without having to enter a password.  If the admin is logged in with "remember me" checked, when login is called without "remember me", the admin's identity cookie is not removed.

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -568,6 +568,10 @@ class User extends Component
             return;
         }
 
+        /* Ensure any existing identity cookies are removed. */  
+        if ($this->enableAutoLogin)  
+            Yii::$app->getResponse()->getCookies()->remove(new Cookie($this->identityCookie));  
+
         $session = Yii::$app->getSession();
         if (!YII_ENV_TEST) {
             $session->regenerateID(true);
@@ -586,8 +590,6 @@ class User extends Component
             if ($duration > 0 && $this->enableAutoLogin) {
                 $this->sendIdentityCookie($identity, $duration);
             }
-        } elseif ($this->enableAutoLogin) {
-            Yii::$app->getResponse()->getCookies()->remove(new Cookie($this->identityCookie));
         }
     }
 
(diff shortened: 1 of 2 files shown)


samdark:
Is it possible to unit-test it?

cebe:
should be possible: https://github.com/yiisoft/yii2/blob/master/tests/framework/web/UserTest.php

samdark:
@maine-mike would you take care of unit tests?

maine-mike:
Yes.

framework/web/User.php
```diff
@@ -568,6 +568,10 @@ public function switchIdentity($identity, $duration = 0)
             return;
         }
 
+        /* Ensure any existing identity cookies are removed. */  
+        if ($this->enableAutoLogin)  
+            Yii::$app->getResponse()->getCookies()->remove(new Cookie($this->identityCookie));
```
Renkas: Always use braces for single line statements.

https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md#54-control-statements

maine-mike:
I wrote a unit test that fails without this fix and passes with this fix.  I added it to the existing UserTest.php unit test as part of this pull request.  Was that the right thing to do?

I couldn't find a clear example of how to mock cookies for the unit test so I made my own.

Thank you.

samdark:
Great. Thank you.

samdark:
Merged. Thanks!

ajoudeh:
Dears,

Kindly look at this scenario:

1. enableAutoLogin ==> true
2. autoRenewCookie ==> false
3. session ==> Database 
3. logged in to the system.
4. new session row added.
5. after some idle time (no action on the webpage) the database session row is garbage OR if i delete the session row directly from table 
6. refresh webpage. 
7. still logged in. because the cookie is exist.
8. after refresh, the identityCookie is not exist
9. try refresh again. ==> logout (no cookie to login)

this happen because you are add this change


Thanks,

samdark:
Please report as separate issue. Thanks.

cebe:
@ajoudeh please open a new issue for that.

PR #10482, opened 2015-12-26
added notes about parameter binding to the guide and classes

### LINKED ISSUE #10462: Error when querying using LIKE with params
```php
...
class Suite extends \yii\db\ActiveRecord
{
   ...
    public static function findByLocation(string $location) : array
    {
        return self::find()
            ->innerJoin('hotel', 'suite.hotel_id=hotel.id')
            ->innerJoin('pt_hotel_address', 'hotel.id=pt_hotel_address.hotel_id')
            ->innerJoin('address', 'pt_hotel_address.address_id=address.id')
            ->where(['like', 'address.district', ':location'], [':location' => $location])
//            ->params([':location' => $location]) <--- it does not work too, same error. Same for addParams()
            ->all();
    }
}
```
Error:
```
Database Exception – yii\db\Exception

SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens
The SQL being executed was: SELECT `suite`.* FROM `suite` INNER JOIN `hotel` ON suite.hotel_id=hotel.id INNER JOIN `pt_hotel_address` ON hotel.id=pt_hotel_address.hotel_id INNER JOIN `address` ON pt_hotel_address.address_id=address.id WHERE `address`.`district` LIKE '%:location%'
Error Info: Array
(
    [0] => HY093
    [1] => 0
)
↵
Caused by: PDOException

SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens

in /home/sidney/www/bitbucket/..../vendor/yiisoft/yii2/db/Command.php at line 837
```

Should not any of those tries do work?

Yii 2.0.6
PHP 7.0

cebe:
fixes #10462 

not sure if it could be more prominent or added to other places too?

### DIFF
--- a/docs/guide/db-query-builder.md
+++ b/docs/guide/db-query-builder.md
@@ -215,6 +215,9 @@ $userQuery = (new Query())->select('id')->from('user');
 $query->where(['id' => $userQuery]);
 ```
 
+Using the Hash Format, Yii internally uses parameter binding so in contrast to the [string format](#string-format), here
+you do not have to add parameters manually.
+
 
 #### Operator Format <span id="operator-format"></span>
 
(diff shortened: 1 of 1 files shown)


SilverFire:
:+1:

PR #10483, opened 2015-12-26
Bug #9670: Fixed PJAX redirect in IE

### LINKED ISSUE #9670: Redirect problems in Internet Explorer 11 (IE11)
When I use the included IE debugger and view the response headers, the header is there and the url is correct, but IE11 isn't reading it.

The response error:
pjax-response-error

The response headers:
pjax-response-headers

One comment on StackOverflow suggested using the status code `200`, and after testing, this does appear to work.  I know the default redirect status code is 302, but this just doesn't work.

## Update

I just respond with redirect to ajax request. Page redirects in all browsers except IE. I'm testting in IE11. I'm getting the following error in console: XMLHttpRequest: Network Error 0x2f76, Could not complete the operation due to error 00002f76..

URL for redirect sent in X-Redirect header. yii.js script handles this header, see line 265. var url = xhr.getResponseHeader('X-Redirect');. In case of IE11 getResponseHeader() returns null instead of URL. As I mentioned before I just respond with 200 status at the moment instead of 302.

SilverFire:
Closes #9670
Works in pair with https://github.com/yiisoft/jquery-pjax/pull/38

### DIFF
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -745,6 +745,7 @@ class Response extends \yii\base\Response
      * meaning if the current request is an AJAX or PJAX request, then calling this method will cause the browser
      * to redirect to the given URL. If this is false, a `Location` header will be sent, which when received as
      * an AJAX/PJAX response, may NOT cause browser redirection.
+     * Takes effect only when request header `X-Ie-Redirect-Compatibility` is absent.
      * @return $this the response object itself
      */
     public function redirect($url, $statusCode = 302, $checkAjax = true)
@@ -758,7 +759,7 @@ class Response extends \yii\base\Response
             $url = Yii::$app->getRequest()->getHostInfo() . $url;
         }
 
-        if ($checkAjax) {
+        if ($checkAjax && Yii::$app->getRequest()->getHeaders()->get('X-Ie-Redirect-Compatibility') !== null) {
             if (Yii::$app->getRequest()->getIsPjax()) {
                 $this->getHeaders()->set('X-Pjax-Url', $url);
             } elseif (Yii::$app->getRequest()->getIsAjax()) {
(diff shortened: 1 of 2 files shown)


framework/web/Response.php
```diff
@@ -758,7 +759,7 @@ public function redirect($url, $statusCode = 302, $checkAjax = true)
             $url = Yii::$app->getRequest()->getHostInfo() . $url;
         }
 
-        if ($checkAjax) {
+        if ($checkAjax && Yii::$app->getRequest()->getHeaders()->get('X-Ie-Redirect-Compatibility') !== null) {
```
klimov-paul: This missmatches the description:

> Takes effect only when request header `X-Ie-Redirect-Compatibility` is absent.

It should be:

```php
Yii::$app->getRequest()->getHeaders()->get('X-Ie-Redirect-Compatibility') === null
```

then

klimov-paul:
This PR breaks regular AJAX redirection.
Perhaps the condition is incorrect, see
https://github.com/yiisoft/yii2/pull/10483/files#r49183315

framework/web/Response.php
SilverFire: Thank you, you're right. Had a TODO to check this PR after holidays, but you came first :)
https://github.com/yiisoft/yii2/commit/019587e544f6909e7f6d2c83546fb2e063b801b9

SilverFire:
Fixed

klimov-paul:
Still fix seems to be incorrect.
With the fix AJAX will follow location __before__ response being processed by JS.
The better fix will be usage of the 200 status code in case of `X-Ie-Redirect-Compatibility` is sent.

```php
if ($checkAjax && ) {
    if (Yii::$app->getRequest()->getHeaders()->get('X-Ie-Redirect-Compatibility') === null && func_num_args() < 2) {
        $statusCode = 200;
    }
            if (Yii::$app->getRequest()->getIsPjax()) {
                $this->getHeaders()->set('X-Pjax-Url', $url);
            } elseif (Yii::$app->getRequest()->getIsAjax()) {
                $this->getHeaders()->set('X-Redirect', $url);
            } else {
                $this->getHeaders()->set('Location', $url);
            }
        } else {
            $this->getHeaders()->set('Location', $url);
        }
```

klimov-paul:
See #10534

PR #10488, opened 2015-12-28
Fix incorrect behaviour of number validator in some locales

vlad-khramov:
In some locales (ru for example) decimal point is comma, so number validator didn't work properly in this locales, because string representation of floats in this locale contains comma instead of dot.

``` php
$val = 0.5;
print "$val"; // "0,5"
```

Build of commit with failing tests: https://travis-ci.org/quantum13/yii2/builds/99038650

### DIFF
--- a/framework/validators/NumberValidator.php
+++ b/framework/validators/NumberValidator.php
@@ -85,7 +85,8 @@ class NumberValidator extends Validator
             return;
         }
         $pattern = $this->integerOnly ? $this->integerPattern : $this->numberPattern;
-        if (!preg_match($pattern, "$value")) {
+
+        if (!preg_match($pattern, $this->getStringValue($value))) {
             $this->addError($model, $attribute, $this->message);
         }
         if ($this->min !== null && $value < $this->min) {
@@ -105,7 +106,7 @@ class NumberValidator extends Validator
             return [Yii::t('yii', '{attribute} is invalid.'), []];
         }
         $pattern = $this->integerOnly ? $this->integerPattern : $this->numberPattern;
-        if (!preg_match($pattern, "$value")) {
+        if (!preg_match($pattern, $this->getStringValue($value))) {
             return [$this->message, []];
         } elseif ($this->min !== null && $value < $this->min) {
             return [$this->tooSmall, ['min' => $this->min]];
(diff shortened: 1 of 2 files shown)


tests/framework/validators/NumberValidatorTest.php
```diff
@@ -37,7 +37,14 @@ public function testValidateValueSimple()
         $this->assertTrue($val->validate(-20));
         $this->assertTrue($val->validate('20'));
         $this->assertTrue($val->validate(25.45));
+
+        $oldLocale = setlocale(LC_ALL, "0");
+        setlocale(LC_ALL, "en_US.UTF-8");
```
samdark: It won't run on Windows. Should be:

```php
setlocale(LC_ALL, 'en_US.UTF-8', 'English_United States.1252');
```

tests/framework/validators/NumberValidatorTest.php
```diff
@@ -37,7 +37,14 @@ public function testValidateValueSimple()
...
         $this->assertFalse($val->validate('25,45'));
+        setlocale(LC_ALL, "en_DK.UTF-8"); //decimal point is comma
```
samdark: It won't run on Windows. Should be:

```php
setlocale(LC_ALL, 'da_DK.UTF-8', 'Danish_Denmark.1252'); //decimal point is comma
```

framework/validators/NumberValidator.php
```diff
@@ -117,6 +118,22 @@ protected function validateValue($value)
...
 
     /**
+     * Returns string represenation of number value with replaced commas to dots, if decimal point
+     * of current locale is comma
+     * @param $value
+     * @return string
+     */
+    private function getStringValue($value)
+    {
+        $localeInfo = localeconv();
```
samdark: Could be something like:

```php
$value = (string)$value;

$locale = localeconv();
$decimalPointSeparator = isset($locale['decimal_point']) ? $locale['decimal_point'] : null;

if ($decimalPointSeparator !== null && $decimalPointSeparator !== '.') {
    $value = str_replace($decimalPointSeparator, '.', $value);
}

return $value;
```

samdark:
It doesn't take locale's thousands separator into account.

samdark:
@quantum13 do you have time to work further on it?

thiagotalma:
This problem is VERY ANNOYING!

It would be great if this were solved ...

samdark:
It is but also it's hard to solve properly one...

samdark:
Closed in favor of https://github.com/yiisoft/yii2/issues/13114

Issue #9340, opened 2015-08-06
Add after('some_column') in ColumnSchemaBuilder


ViZH:
It's better to use new syntax:

```php
$this->addColumn(Model::tableName(), 'new_column', $this->boolean()->after('some_column'));
```

instead of 

```php
$this->addColumn(Model::tableName(), 'new_column', Schema::TYPE_BOOLEAN.' AFTER some_column');
```

cebe:
can you do some research about the syntax for this in different dbms? Links to the corresponding docs would be helpful.

ViZH:
It's not possible to add a column between two existing columns with an ALTER TABLE statement in SQLite, PostgreSQL and MSSQL. But it's important to Oracle and MySQL. Fixed length fields in some storage engines must be aligned before varchars, texts and blobs.

Oracle syntax (http://docs.oracle.com/cd/E17952_01/refman-5.0-en/alter-table.html)
ALTER TABLE table CHANGE COLUMN column AFTER column

MySQL syntax (https://dev.mysql.com/doc/refman/5.1/en/alter-table.html)
ALTER TABLE table CHANGE COLUMN column AFTER column

Cubrid syntax (http://www.cubrid.org/manual/841/en/CHANGE%7CMODIFY%20Clause)
ALTER TABLE table CHANGE COLUMN column AFTER column

df2:
This covers it I think: https://github.com/df2/yii2/commit/8501bb2e644f24b43742de923f550412558fcc1e

Can someone please take that and plz make it into a pull request.

Regarding the tests: I made some modifications to the formatting of the test data structures in another pull request (https://github.com/yiisoft/yii2/pull/9617).

I think someone has made changes to the test data in the mean time so the tests aren't mergable anymore.

It might be worth deciding if you want to stick with the old or new test data format before we write the tests for this. We can get the `char` tests up to shape at the same time then.

samdark:
@df2 if it's MySQL and Oracle, it should not add anything when used with other DBs.

df2:
@samdark well I would have thought in that case they would just not use the `after()` method when writing their migrations.

df2:
@samdark I suppose switching from one database that supports it to another that doesn't and you want to run the migrations against the new one... So something like this then? Where should that list of compatible databases be stored? Thx!

```php
protected function buildAfterString()
{
    if ($this->after !== null && in_array(Yii::$app->db->..., [ORACLE,MYSQL,CUBRID]) {
[...]

PR #10494, opened 2015-12-28
Add `after()` option to migrations schema builder #9340

usecomposable:


samdark:
#9340

samdark:
What do you think about https://github.com/hulyman/yii2/commit/8f0277687b76280cdfecb84cc0d4cd2ce313b782?

df2:
`FIRST` could be added too, yeah. `custom()` in not so in to.

PR #10517, opened 2016-01-02
added support for non-gregorian calendar to formatter

### LINKED ISSUE #8687: jalali date in formatter
if it's possible for  you add an ablility to framework, to have jalali date format and  when we set some property in config, like (dateType => 'jalali',) or (dateType => 'gregorian ')
by depending on this property in config, formatter works.

hear is the best jalalidate vendor in php that tested in all iranian app and it's working currectly:
http://jdf.scr.ir/

cebe:
fixes #8687 

Need to verify this actually works.

@hooman-pro @z-avanes could you provide information about calendars and expected formatting?

### DIFF
--- a/tests/framework/i18n/FormatterDateTest.php
+++ b/tests/framework/i18n/FormatterDateTest.php
@@ -88,6 +88,43 @@ class FormatterDateTest extends TestCase
         $this->assertSame($this->formatter->nullDisplay, $this->formatter->asDate(null));
     }
 
+    public function testIntlAsDateOtherCalendars()
+    {
+        // Persian calendar
+        $this->formatter->locale = 'fa_IR@calendar=persian';
+        $this->formatter->calendar = \IntlDateFormatter::TRADITIONAL;
+        $this->formatter->timeZone = 'UTC';
+
+        $value = 1451606400; // Fri, 01 Jan 2016 00:00:00 (UTC)
+        $this->assertSame('۱۳۹۴', $this->formatter->asDate($value, 'php:Y'));
+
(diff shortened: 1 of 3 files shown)


cebe:
Seems there are a lot of different calendars out there. https://en.wikipedia.org/wiki/Template:Calendars would be good if we could cover the most relevant here.

z-avanes:
yes a lot of calendars exist and i think ICU project cover all ( http://www.icu-project.org/ )
i use new PR code in my local machine and its ok and verfiy for persian calender ( probably its ok for other calender supported by intl )

tests/framework/i18n/FormatterDateTest.php
```diff
@@ -88,6 +88,32 @@ public function testAsDate()
...
 
+    public function testIntlAsDateOtherCalendars()
+    {
+        // Persian calendar
+        $this->formatter->locale = 'fa_IR@calendar=persian';
+        $this->formatter->calendar = \IntlDateFormatter::TRADITIONAL;
+        $this->formatter->timeZone = 'UTC';
+
+        $value = 1451606400; // Fri, 01 Jan 2016 00:00:00 (UTC)
+        $this->assertSame('۱۳۹۴/۱۰/۱۱ ه‍.ش.', $this->formatter->asDate($value, 'short'));
```
cebe: this fails on travis but works on my computer, seems icu version makes this untestable...

cebe:
merged. @z-avanes thanks for checking!

z-avanes:
thanks too @cebe

PR #10521, opened 2016-01-04
Security exception cleanup

tom--:
In 2.0.7 these 4 methods don't throw these exceptions any more.

This doesn't require an issue or changelog entry, does it?


samdark:
Nope. It's fine. Thanks!

PR #10525, opened 2016-01-05
Code of conduct

samdark:
Yii community is a good place and in order for it to stay so here are the rules. We hope we'll never have to force these but, well, it happens.

Comments/suggestions are welcome.

### DIFF
--- a/code-of-conduct.md
+++ b/code-of-conduct.md
@@ -1,0 +1,45 @@
+Yii Contributor Code of Conduct
+=======================
+
+As contributors and maintainers of this project, and in order to keep Yii community open and welcoming, we ask to respect all community members.
+
+We are committed to making participation in this project a good experience for everyone.
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery
+* Personal attacks
+* Trolling or insulting/derogatory comments
+* Public or private harassment
+* Publishing other's private information, such as physical or electronic
+  addresses, without explicit permission
+* Other unethical or unprofessional conduct
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+By adopting this Code of Conduct, project maintainers commit themselves to
+fairly and consistently applying these principles to every aspect of managing
+this project. Project maintainers who do not follow or enforce the Code of
+Conduct may be permanently removed from the project team.
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community.
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting core team members. All
+complaints will be reviewed and investigated and will result in a response that
+is deemed necessary and appropriate to the circumstances. Maintainers are
+obligated to maintain confidentiality with regard to the reporter of an
+incident.
+
+
+This Code of Conduct is adapted from the [Contributor Covenant][homepage],
+version 1.3.0, available at
+[http://contributor-covenant.org/version/1/3/0/][version]
+
+[homepage]: http://contributor-covenant.org
+[version]: http://contributor-covenant.org/version/1/3/0/


githubjeka:
I swear :hand:

SilverFire:
:+1:

cebe:
A link to it in https://github.com/yiisoft/yii2/blob/master/CONTRIBUTING.md should be added too.

cebe:
:+1: btw

kidol:
> Yii community is a good place

Yes.

> and in order for it to stay so

It worked for years without this PR.

>Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by contacting core team members.

That's common sense.

> All complaints will be reviewed and investigated and will result in a response that is deemed necessary and appropriate to the circumstances.

This is already the case I guess.

:-1:

cebe:
@kidol if you agree with everything why is it a problem to have a public statement that makes it clear?

samdark:
@kidol yes, it works and I'm very glad it is common sense for our community. But there were rare exceptions as well which likely won't happen if we would've pointed to this common sense written in stone.

I see nothing bad in having it despite that I'd love to never force people to behave according to common sense.

mdomba:
:+1:

klimov-paul:
:+1:

kidol:
@samdark 
I remember only one instance of really bad behavior which was dealt with after the affected user contacted admin. Some guy apparently downvoted all his forum posts: Code of conduct certainly won't help for such type of persons.

@cebe 
I simply see no issue to be fixed. Putting something in place which is de-facto already in place makes no sense. New users reading that are just wasting time. If a new user comes he certainly won't look for a code of conduct himself, instead it will be forced on him for no valid reason.

Well just my opinion. It works good without.

cebe:
> I simply see no issue to be fixed.

@kidol we felt the need for some official statement to point to, so this is why we add it here. It is not fixing an existing issue but is useful if we need to deal with possible issues in the future. Having this will not prevent bad behavior by itself but it will help us when dealing with it, having something written down that we agreed on beforehand.

kidol:
@cebe 
> we felt the need

When? Years it worked without. I doubt the related PHP RFC is a coincidence.

>but is useful if we need to deal with possible issues in the future

If something serious happens, you don't need the code of conduct to notice it nor ask it for a decision, but instead communicate with others to form it, just like you do now.

Also the notion of using the code of conduct as a reference... I mean would you link a user to the code of conduct to remind him when he acts a little over-agressive, instead of talking to him trying to cool the situation down? That would be rather anti-social. I wouldn't like that as a user nor as an outside observer.

cebe:
> When? Years it worked without. I doubt the related PHP RFC is a coincidence.

We have had some small situations in the past where we discussed about introducing a CoC afterwards but we never actually started working on it. Today it popped up again because of the PHP RFC and we decided to finally do it.

> If something serious happens, you don't need the code of conduct to notice it nor ask it for a decision, but instead communicate with others to form it, just like you do now.
> ...
> I mean would you link a user to the code of conduct to remind him when he acts a little over-agressive, instead of talking to him trying to cool the situation down?

No, I would not do anything different as I would do without the CoC. The main purpose of it is to write down what we agree on as "common sense" for the rules in the community because there are people for which the sense is not as common as it should be.

What's wrong with having a public statement that makes it clear for everyone?

kidol:
>I would not do anything different as I would do without the CoC

I know that you or any core member for that matter wouldn't do the example I gave, but trust me, some people in PHP project will do it. What I want to say: It can have it's downsides in social-interaction aspect, for in my opinion, no real benefit.

The people you talk about may not read the code of conduct at all until they overstep the line. After all it's not a binding agreement that you must accept but a voluntary read. If they overstep the line you must interact with them in any case. Whether they read something doesn't matter. They get proper punishment.

> What's wrong with having a public statement that makes it clear for everyone?

I think it's totally unnecessary (waste of time) and potentially cause more harm than good. Maybe not in Yii project because mostly we get well along here anyway it seems, but in the end it's about principles.

On the other hand, it doesn't hurt me if implemented. Just voicing opinion. People seem to like the idea, so go ahead.

qiangxue:
:+1:

DaSourcerer:
:-1: This is atrocious.

DaSourcerer:
May I ask why the second paragraph differs from the version found over at contributor-covenant.org?

@kidol:
> it doesn't hurt me if implemented

You wish. I have a very sinister apprehension about this.

samdark:
> May I ask why the second paragraph differs from the version found over at contributor-covenant.org?

Because original version emphasizes on a particular list of differences which doesn't matter. If we'll get contribution from non-human life form and it would behave well, it's OK.

DaSourcerer:
@samdark I see, thanks for the clarification.

thathumanthere:
You have a code of conduct. Ima let that sink in a while. :+1:

Issue #7546, opened 2015-03-03
Client validation should be disabled if 'when' is used


mikehaertl:
I suggest to skip clientside validation for an input field, when the following conditions are met:
- A `when` callback is set on a rule AND
- no `whenClient` is configured

Because if you use `when` you obviously only want to validate under certain conditions - otherwhise you'd not use it. But (without a `whenClient` function) the validation is still always performed on the clientside. So `when` practically has no effect at all.

As a workaround I now always have to set `['enableClientValidation' => false]` in the field options, which is quite inconvenient.

samdark:
What are use cases for specifying `when` but not specifying `whenClient`?

mikehaertl:
Your question implies, that you can't think of a valid use case. Even if that was the case, then the above change wouldn't hurt you, as the condition would never be met for you.

The use case for me is:

I have a dependend validation: A field is only required if certain form values match a condition that I can only check against the database. Yes, you could add some js code that does an AJAX request - but that's tedious and you may simply want to rely on server validation as a fallback. Still I want all other fields to do clientside validation where possible.

I should not have to touch any view files, just because I added some more validation logic.

SilverFire:
For the future: `enableClientValidation` may differ from the `whenClient` in ajax validation.

whenClient | enableClientValidation | Result
------------ | ------------- | ------------
`null|true`  | `true` | JS+AJAX client validation
`false` | `true` | AJAX-only client validation
`null|true` | `false` | Neither JS nor AJAX client validation

mikehaertl:
Coming back to this rather old issue: I still thinks it's valid.

Let me try to phrase it better:

If you define `when` but not `whenClient`  this means: Whether the validation rule should be applied is decided on the server side on form submission. So on the client side you don't know, whether the rule should be applied or not. In that case there must not be any client validation as this would assume, that the rule **is always** applied - which is wrong.

samdark:
Won't be implemented in 2.0.

PR #10526, opened 2016-01-06
allow disabilng client validators by setting 'whenClient' = false

k-timoshenko:
Based on proposal https://github.com/yiisoft/yii2/issues/7546

`whenClient` enchancement. Allow disable client validators by seting `'whenClient' => false` in AR's rules.
Pull request little differs from the original issue: `whenClient` **MUST** be set to `false`. Its because of backward compatibility.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.7 under development
 -----------------------
 
+- Enh #7546: Add possibility to disable `whenClient` option on input
 - Bug #6351: Find MySQL FK constraints from `information_schema` tables instead of `SHOW CREATE TABLE` to improve reliability (nineinchnick)
 - Bug #6363, #8301, #8582, #9566: Fixed data methods and PJAX issues when used together (derekisbusy)
 - Bug #6876: Fixed RBAC migration MSSQL cascade problem (thejahweh)
--- a/framework/validators/Validator.php
+++ b/framework/validators/Validator.php
@@ -156,13 +156,14 @@ class Validator extends Component
      */
     public $when;
     /**
-     * @var string a JavaScript function name whose return value determines whether this validator should be applied
+     * @var string|false a JavaScript function name whose return value determines whether this validator should be applied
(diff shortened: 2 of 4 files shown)


Faryshta:
how is it different from `enableClientValidation`?

SilverFire:
@t-kanstantsin thank you for your effort, but I'm closing this PR, because the original idea in #7546 was about adding a default behavior, but we can not do it because of BC. I agree with @Faryshta, the client validation may be disabled manually with `'enableClientValidation' => false`.

If you don't agree or have any additional explanation - feel free to post it. Thank you

t-kanstantsin:
@Faryshta @SilverFire I agree, that with `enableClientValidation` we can disable client validation, but it will disable __all__ validators. But sometimes it is needed to disabled only __few__ validations and it is uncomfortably each time write code like: `function() {return false;}`.
In the description I emphasized difference with issue #7546 - in pull request `whenClient` must be defined as `false` to disable client validator. Now (without it) this would lead to an error, so this enchancement will no break BC.

Example:
```php
[['foo'], 'required', 'when' => function ($model, $attribute) {
    return {some magick};
},],
```
In this case I might not want to check `$bar` on client side. It might be very complex or  Or I don't want to write js code right here in AR's rules.
But, generally, I want client validation to be enable for other validators. So I can't just set `'enableClientValidation' => false,`.

t-kanstantsin:
Sorry, but I don't knew about `enableClientValidation` in Validator class too.
This is what I search to use. 

Although I think, that allow to disable validation by setting `when` or `whenClient` to `false` is a better way, than special field like `enableClientValidation`.
But it may break BC or make it less clear how really disable client validation.

Faryshta:
you can propose it for 2.1, i think it might be a good feature to use only 'whenClient'

SilverFire:
For the future: `enableClientValidation` may differ from the `whenClient` in ajax validation.

whenClient | enableClientValidation | Result
------------ | ------------- | ------------
`null|true`  | `true` | JS+AJAX client validation
`false` | `true` | AJAX-only client validation
`null|true` | `false` | Neither JS nor AJAX client validation

PR #10534, opened 2016-01-08
#10483 Response::redirect() uses 200 status code for IE 11

klimov-paul:
Adjustments for #10483
`Response::redirect()` uses 200 status code if 'X-Ie-Redirect-Compatibility' header sent

### DIFF
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -760,7 +760,7 @@ class Response extends \yii\base\Response
         }
 
         if ($checkAjax) {
-            if (Yii::$app->getRequest()->getIsPjax()) {
+            if (Yii::$app->getRequest()->getIsAjax()) {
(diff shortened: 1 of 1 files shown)


SilverFire:
https://github.com/yiisoft/yii2/commit/57ab949be543471a7f59aa3c18c2d5199e17de91

michaelarnauts:
Wait, shouldn't the check be if the header is absent, as it is specified in the phpdoc above the function? This MR changed the `===` back to a `!==`. The documentation contradicts the code now.

How is it supposed to work? Should the header be specified for the redirect to work on IE? That would require to change every ajax call to include the `X-Ie-Redirect-Compatibility` header for the redirect to work in IE. That doesn't seem right?

samdark:
@michaelarnauts please create a separate issue since comment is likely to be lost...

PR #10536, opened 2016-01-08
Updated phpDoc to ManagerInterface::addChild()

githubjeka:
P.S. PhpManager and DbManager returns `true` only. Not false.


samdark:
Thanks!

PR #10538, opened 2016-01-09
Prefixes for yii.activeForm.js event names

aleskiontherun:
The yii.activeForm.js script triggers `ajaxComplete` event named the same as jQuery's `ajaxComplete` which causes duplicate `$(document).ajaxComplete()` handlers execution.

For example the built in ajax redirect (`initRedirectHandler()` in yii.js) is triggered two times, which makes `Yii::$app->session->setFlash()` useless at least with Firefox 43.0.4 browser because it makes two requests but aborts the first one so the flash data is lost. In Chrome 47 the redirect is OK but still duplicate handlers execution.

Using prefixes in custom event names is a good practice.

screenshot

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -5,6 +5,7 @@ Yii Framework 2 Change Log
 2.0.9 under development
 -----------------------
 
+- Bug #11032: Fixed double `ajaxComplete` event in `yii.activeForm.js` by namespacing event names; `yii.gridView.js` events also namespaced (dizeee)
 - Enh #11725: Added indexes on message tables (OndrejVasicek)
 - Enh #10422: Added `null` method on `yii\db\ColumnSchemaBuilder` to explicitly set column nullability (nevermnd)
 - Enh #9574: Implicit run `ColumnSchemaBuilder::null()` when default value is set to `null`. (rob006)
--- a/framework/assets/yii.activeForm.js
+++ b/framework/assets/yii.activeForm.js
@@ -24,7 +24,7 @@
 
     var events = {
         /**
-         * beforeValidate event is triggered before validating the whole form.
+         * beforeValidate.yiiActiveForm event is triggered before validating the whole form.
          * The signature of the event handler should be:
          *     function (event, messages, deferreds)
          * where
@@ -36,9 +36,9 @@
          * If the handler returns a boolean false, it will stop further form validation after this event. And as
          * a result, afterValidate event will not be triggered.
          */
-        beforeValidate: 'beforeValidate',
+        beforeValidate: 'beforeValidate.yiiActiveForm',
         /**
-         * afterValidate event is triggered after validating the whole form.
+         * afterValidate.yiiActiveForm event is triggered after validating the whole form.
          * The signature of the event handler should be:
          *     function (event, messages, errorAttributes)
          * where
@@ -47,9 +47,9 @@
          *    for the corresponding attributes.
          *  - errorAttributes: an array of attributes that have validation errors. Please refer to attributeDefaults for the structure of this parameter.
          */
-        afterValidate: 'afterValidate',
+        afterValidate: 'afterValidate.yiiActiveForm',
         /**
-         * beforeValidateAttribute event is triggered before validating an attribute.
+         * beforeValidateAttribute.yiiActiveForm event is triggered before validating an attribute.
          * The signature of the event handler should be:
          *     function (event, attribute, messages, deferreds)
          * where
@@ -61,9 +61,9 @@
          * If the handler returns a boolean false, it will stop further validation of the specified attribute.
          * And as a result, afterValidateAttribute event will not be triggered.
          */
-        beforeValidateAttribute: 'beforeValidateAttribute',
+        beforeValidateAttribute: 'beforeValidateAttribute.yiiActiveForm',
         /**
-         * afterValidateAttribute event is triggered after validating the whole form and each attribute.
+         * afterValidateAttribute.yiiActiveForm event is triggered after validating the whole form and each attribute.
          * The signature of the event handler should be:
          *     function (event, attribute, messages)
          * where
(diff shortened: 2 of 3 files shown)


Faryshta:
@dizeee @SilverFire is there an standar or something about it? we can follow the bootstrap notation

http://bootstrapdocs.com/v3.3.6/docs/javascript/#js-events

`afterValidate.yii.activeForm`

SilverFire:
Related to #11032, seem to be a smarter fix than I've suggested in https://github.com/yiisoft/yii2/issues/11032#issuecomment-193140267

dizeee:
@Faryshta You are right. Didn't find anything straight right away but this example shows how jQuery treats event namespaces: https://api.jquery.com/event.namespace/. I'll change my patch accordingly.

Faryshta:
@dizeee I still think that the namespace should be `yii.activeForm` specially since grid view handles js events too

https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.gridView.js

Also this is a bc-break for someone using the existings events since they won't be triggered anymore.

dizeee:
According to jQuery [documentation](http://api.jquery.com/on/):

> Namespaces are similar to CSS classes in that they are not hierarchical;

So I've left just a single `yiiActiveForm` namespace and placed it after event names.

dizeee:
Added `yiiGridView` namespace in yii.gridView.js. I also noticed that it is already used for `click.yiiGridView` event so notation is preserved.

dizeee:
Tinkered around with this code a bit. Looks like it's not so simple. Here's how events with different names are triggered:
```
// ajaxComplete
ajaxComplete on form by yii
ajaxComplete on document by yii
ajaxComplete on document by jQuery

// ajaxComplete.yiiActiveForm
ajaxComplete.yiiActiveForm on form by yii
ajaxComplete.yiiActiveForm on document by yii
ajaxComplete.yiiActiveForm on document by jQuery
ajaxComplete on document by jQuery

// yiiActiveForm.ajaxComplete
yiiActiveForm.ajaxComplete on form by yii
yiiActiveForm.ajaxComplete on document by yii
ajaxComplete on document by jQuery
```
In the current fix global `ajaxComplete.yiiActiveForm` event is still triggered twice somehow. In my initial fix `yiiActiveForm.ajaxComplete` is triggered once as expected, but this notation is not correct (or is it?).

Any way it needs some more research at this point. I'll try to clear things out.

dizeee:
> Namespaces are similar to CSS classes in that they are not hierarchical;

Looks like it's a tricky one. As I see it, triggering events is like applying CSS rules while event handlers are like DOM elements. Quick example:

```js
var $d = $(document),
    $b = $(document.body),
    l = function(e, lvl){
      console.log(
        'EVENT: ', e.type, e.namespace, ';',
        'HANDLE:', e.handleObj.type, e.handleObj.namespace, ';',
        'LEVEL:', lvl
      );
    };

$b.html('<div id="myElement"></div>');

var $el = $('#myElement');

$d.on('myEvent', function(e){l(e, 'document')});
$d.on('myEvent.myNamespace', function(e){l(e, 'document')});
$d.on('myEvent', 'div', function(e){l(e, 'document div')});
$d.on('myEvent.myNamespace', 'div', function(e){l(e, 'document div')});

console.log('=== Triggering myEvent on docunent ===');
$d.trigger('myEvent');

console.log('=== Triggering myEvent.myNamespace on docunent ===');
$d.trigger('myEvent.myNamespace');

console.log('=== Triggering myEvent on element ===');
$el.trigger('myEvent');

console.log('=== Triggering myEvent.myNamespace on element ===');
$el.trigger('myEvent.myNamespace');
```
Output will be:
```
=== Triggering myEvent on docunent ===
EVENT: myEvent (empty string) ; HANDLE: myEvent (empty string) ; LEVEL: document
EVENT: myEvent (empty string) ; HANDLE: myEvent myNamespace ; LEVEL: document

=== Triggering myEvent.myNamespace on docunent ===
EVENT: myEvent myNamespace ; HANDLE: myEvent myNamespace ; LEVEL: document

=== Triggering myEvent on element ===
EVENT: myEvent (empty string) ; HANDLE: myEvent (empty string) ; LEVEL: document div
EVENT: myEvent (empty string) ; HANDLE: myEvent myNamespace ; LEVEL: document div
EVENT: myEvent (empty string) ; HANDLE: myEvent (empty string) ; LEVEL: document
EVENT: myEvent (empty string) ; HANDLE: myEvent myNamespace ; LEVEL: document

=== Triggering myEvent.myNamespace on element ===
EVENT: myEvent myNamespace ; HANDLE: myEvent myNamespace ; LEVEL: document div
EVENT: myEvent myNamespace ; HANDLE: myEvent myNamespace ; LEVEL: document
```

In our case yii.activeForm.js triggers `ajaxComplete.yiiActiveForm` event that bubbles up the DOM hierarchy from the form element level to the document level and calls all handlers for `ajaxComplete.yiiActiveForm` event (now there's no double redirect because `ajaxComplete` is not triggered at this point). Than jQuery triggers its `ajaxComplete` event and all document level handlers with or without namespaces are called including those for `ajaxComplete.yiiActiveForm` event.

So I think it should be explicitly documented that all yiiActiveForm event handlers should be attached to the form elemtnt(s):
```js
// Good
$('#formId').on('eventName.yiiActiveForm', handler);

// Even better
$(document).on('eventName.yiiActiveForm', '#formId', handler);

// Globally
$(document).on('eventName.yiiActiveForm', 'form', handler)

// Bad
$(document).on('eventName.yiiActiveForm', handler);

// Possible but not recommended
$(document).on('eventName', function(e) {
    if (e.namespace !== 'yiiActiveForm') {
        return;
   }
});
```

SilverFire:
@dizeee @Faryshta sorry for a long delay with this PR.

As far as I see, current PR is BC-breaking. If someone has code like the following:

```javascript
$('#contact-form').on('beforeValidate', (e) => console_log(e.type, e.namespace));
```

Adding namespace to the `beforeValidate` event will make this listener useless. Am I right?

dizeee:
@SilverFire No, you are not. Namespacing won't brake listeners with not namespaced event names.

dizeee:
Sorry for the delay too. I haven't been working with Yii for a while, though it's still my favorite framework. I'll try to review the regarding framework changes, resolve conflicts and provide some clear examples on this issue as soon as I have time.

samdark:
Won't make it into 2.0.

PR #10541, opened 2016-01-10
Divide by zero prevention.

youmad:
There can be a situation when `BaseConsole::$_progressEtaLastDone` is equal to `$done`, so it can throw a "Division by zero" exception.


cebe:
Merged, thanks!

PR #10553, opened 2016-01-12
phpDoc ActiveQuery:with/joinWith updated

githubjeka:


samdark:
You're using both `\yii\db\ActiveQuery` and `\yii\db\ActiveQueryInterface` for the same hint. Let's be consistent.

githubjeka:
- The `ActiveQueryInterface` **interface** doesn't know about the `\yii\db\ActiveQuery` class. Because phpDoc of `ActiveQueryInterface` uses the self class. 

- The `ActiveQueryTrait` **trait** doesn't know about the `\yii\db\ActiveQuery` class. Because phpDoc of `ActiveQueryTrait` uses known `ActiveQueryInterface` interface. 

- The `ActiveQuery` class is a specific implementation of the `ActiveQueryInterface` interface and have other methods. For this reason and for the autocompleate phpDoc of `ActiveQuery::joinWith` uses the self class.

githubjeka:
@samdark do you mean the other side - considering as an example of the finished implementation `\yii\db\ActiveQuery` only ?

githubjeka:
Right you are.

samdark:
Thanks!

Issue #9935, opened 2015-10-16
Unable to use EachValidator with CompareValidator to check by $compareAttribute


Vovan-VE:
When using `EachValidator` with `CompareValidator` inside, then `$compareAttribute` is ignore, t.i. only `$compareValue` used.

```php
public function rules() {
    return [
        // each element in $this->foo[] is not equal to $this->bar
        ['foo', 'each', 'rule' => ['compare', 'compareAttribute' => 'bar', 'operator' => '!=']],
    ];
}
```
Got `InvalidConfigException` from there https://github.com/yiisoft/yii2/blob/master/framework/validators/CompareValidator.php#L152 .

Yii/2.0.6

Vovan-VE:
I used ugly temporary workaround:

```php
public function rules() {
    return [
        ['bar', '...'],
        // each element in $this->foo[] is not equal to $this->bar
        ['foo', 'each', 'rule' => [
            'compare',
            // FIXME: HACK https://github.com/yiisoft/yii2/issues/9935
            //'compareAttribute' => 'bar',
            'compareValue' => &$this->bar,
            'operator' => '!=',
        ]],
    ];
}
```

SilverFire:
Confirmed

Vovan-VE:
Look's like line 128 does nothing, rather then extected, due to line 118: https://github.com/yiisoft/yii2/blob/master/framework/validators/EachValidator.php#L115

```php
public function validateAttribute($model, $attribute)
{
    ...
    $validator = $this->getValidator();
    if (...) {
        ...
    } else {
        $this->getValidator($model); // ensure model context while validator creation
        ...
    }
}
```
Is logic correct?

This does not cause the bug itself, but probably may cause another problems.

andrewnester:
Hi guys!
I've just added pull request with fix for this issue + unit test to prove it.

As solution for issue it was decided to check inside `EachValidator` if we use `CompareValidator`.

If we do and `compareValue` property is not set for this validator, but only `compareAttribute` is set, then we are manually set `compareValue` to corresponding value of model `compareAttribute`

klimov-paul:
Resolved by commit 205f16811a6071ad5d1dbea4cba9fd8a3282f6c5

PR #10565, opened 2016-01-13
Fix for using CompareValidator inside EachValidator #9935

andrewnester:
As solution for issue it was decided to check inside `EachValidator` if we use `CompareValidator`.

If we do and `compareValue` property is not set for this validator, but only `compareAttribute` is set, then we are manually set `compareValue` to corresponding value of model `compareAttribute`


samdark:
#9935

klimov-paul:
Should be resolved by commit 205f16811a6071ad5d1dbea4cba9fd8a3282f6c5

Issue #10218, opened 2015-11-22
Flash messages not showing after logging out a user.


drew1two:
With a clean install of advanced app.  Open frontend (and backend) siteController.
Add 
Yii::$app->session->setFlash('success', 'You have been successfully logged out!');
to line 107 of frontend/siteController
and Add
Yii::$app->session->setFlash('success', 'You have been successfully logged in!');
to line 91 before the return $this->goBack(); call.

Login and see the successfully logged in message.
Logout and notice there is no successfully logged out message.

Happens the same in the backend as well.

I have a working version I downloaded / updated at:-
"time": "2015-09-24 23:06:03".
(working) meaning I get my flash messages displayed as expected.
Just for some reason my new apps don't display them.

I can see from the logs that the flash messages are being set in the POST logout.
Just not being displayed in the GET index page.

I tried using $this->redirect(['site/login']); as well with the same results.

unclead:
Sessio is destroyed after calling logout method tahts why you dont see any flash messages after redirect

You can call `Yii::$app->user->logout(false);` if you dont want to session will be destroyed
But i am unsure is it rght way or not

drew1two:
I thought that as soon as you call:-
Yii::$app->session->setFlash(...
It automatically creates a new session?
So logout kills the 'user' session, but setting flash message immediately afterwards will create a new session for basic application functionality.

If I use Yii::$app->user->logout(false);. Does it still retain private information about the user (like username, email etc.)?  Or just their publically available info (like ip, browser etc.)?

unclead:
This code works correct for me
```
    public function actionLogout()
    {
        Yii::$app->user->logout();
        Yii::$app->session->setFlash('success', 'Success');

        $this->redirect(Yii::$app->getUser()->loginUrl);
    }
```

But if i call `setFlash` before `logout` it doesnt work

unclead:
If it doesnt work for you please write which version of yii do you use and show us full version of your action

drew1two:
Your example is exactly the same I've implemented.  Because it wasn't working, I tested it out with a clean install of the advanced app.
I did exactly as you described above (which falls on line 107 of frontend/siteController when clean install).
example:-

    /**
     * Logs out the current user.
     *
     * @return mixed
     */
    public function actionLogout()
    {
        Yii::$app->user->logout();

        Yii::$app->session->setFlash('success', 'You have been successfully logged out!');
        return $this->redirect(['site/login']);
        // return $this->goHome();
    }

[...]

PR #10573, opened 2016-01-14
#10218 - Fix for session regenerate id issue after user logged out

andrewnester:
This is fix for #10218 
It is based on solution @klimov-paul offered in conversation for this ticket

### DIFF
--- a/framework/web/Session.php
+++ b/framework/web/Session.php
@@ -182,7 +182,9 @@ class Session extends Component implements \IteratorAggregate, \ArrayAccess, \Co
     {
         if ($this->getIsActive()) {
             @session_unset();
+            $sessionId = session_id();
             @session_destroy();
+            @session_id($sessionId);
         }
     }
 
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -43,6 +43,7 @@ Yii Framework 2 Change Log
 - Bug #10052: Fixed `yii\i18n\Formatter` to work with huge numbers on 32-bit arch (necrox87, silverfire)
 - Bug #10101: Fixed assignments saving on role removing in `\yii\rbac\PhpManager` (rezident1307)
 - Bug #10142: Fixed `yii\validators\EmailValidator` to check the length of email properly (silverfire)
+- Bug #10218: Fixed Flash messages not showing after logging out a user (andrewnester)
 - Bug #10263: Fixed `yii\validators\UniqueValidator` to work properly when model is not instance of `targetClass` (bupy7, githubjeka, silverfire)
 - Bug #10278: Fixed `yii\helpers\BaseJson` support \SimpleXMLElement data (SilverFire, LAV45)
 - Bug #10302: Fixed JS function `yii.getQueryParams`, which parsed array variables incorrectly (servocoder, silverfire)
--- a/tests/framework/web/SessionTest.php
+++ b/tests/framework/web/SessionTest.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace yiiunit\framework\web;
+
+use yii\web\Session;
+use yiiunit\TestCase;
+
+/**
+ * @group web
+ */
+class SessionTest extends TestCase
+{
+    /**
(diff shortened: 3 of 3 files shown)


SilverFire:
This fix does not work on my env `PHP 5.6.11-1ubuntu3.1`

```php
    public function actionLogout()
    {
        $id0 = Yii::$app->session->getId(); // g0c21rtignr91r9353asddg370
        Yii::$app->user->logout();
        $id1 = Yii::$app->session->getId(); // '' (empty string)
        Yii::$app->session->setFlash('success', 'You have been successfully logged out!');
        $id2 = Yii::$app->session->getId(); // '' (empty string)

        return $this->goHome();
    }
```

And then session data will be never saved correctly. I don't have a clue, at least for now.

SilverFire:
@klimov-paul how did it work for you?

andrewnester:
Hi @SilverFire 
You were absolutely right, it didn't work as expected.
I've just implemented new solution, tested it on basic application - looks like it works fine.
Now after `Session::destroy` called, new session with new session id is started

SilverFire:
Travis failed on PHP7

andrewnester:
Fixed, sorry. All checks are passed now. Looks like session_start behavior was changed in PHP7.

framework/web/Session.php
```diff
@@ -183,6 +183,8 @@ public function destroy()
         if ($this->getIsActive()) {
             @session_unset();
             @session_destroy();
+            @session_start();
```
klimov-paul: This is obviously incorrect: `session_start()` will start new session, thus there is no way to close session anymore.

klimov-paul:
> It is based on solution @klimov-paul offered in conversation for this ticket

Well, I don't see my proposal here. Back in #10218 I suggested setting session ID to `null` manually on destroy: 

```php
@session_id(null);
```

Using of `session_regenerate_id()` or especially `session_start()` inside `destroy()` is not acceptable.
If my original solution is not correct by some reason, we should leave the code as it is and update the documentation, warning the developers they may need to invoke `Session::regenerateId()` manually.

SilverFire:
> Well, I don't see my proposal here

It was implemented initially, but then @andrewnester did a force push with update, I guess

SilverFire:
> Using of session_regenerate_id() or especially session_start() inside destroy() is not acceptable.

Agree on this.

andrewnester:
@klimov-paul yes, @SilverFire is right, previously it was using `session_id(null)`, sorry for forced update. it was changed because `session_id(null)` doesn't solve the problem with flash messages, at least for PHP 5.6 and PHP 5.5 (tested on couple of versions).

I will try to investigate further how it could be solved without starting new session

andrewnester:
I just found and implemented solution based on this comment http://php.net/manual/en/function.session-destroy.php#114709
What do you think? PR has been updated too

klimov-paul:
I am afraid this relates to the different matter and unlikely can solve the problem.

andrewnester:
@klimov-paul I've tested it on basic app with PHP 5.6.16 and PHP 5.5.30 - looks like works fine and flash messages successfully displayed after logging out

klimov-paul:
Well, update PR - lets have a look on it.

SilverFire:
It's already updated

klimov-paul:
Looks fine.

SilverFire:
```php
    public function actionLogout()
    {
        $id0 = Yii::$app->session->getId(); // fqv0q23h6870fdc9esmvdcpri2
        Yii::$app->user->logout();
        $id1 = Yii::$app->session->getId(); // v3apcm77672sbrof45onm0m9q4
        Yii::$app->session->setFlash('success', 'You have been successfully logged out!');

        return $this->goHome();
    }
```

Session id changes, but the bug seems to be fixed

andrewnester:
Yes, it is expected, because call to `User::logout` that calls `User::switchIdentity` actually creates new session. The most important thing for fix is that after `Session::destroy`, session id must remain the same.

SilverFire:
Merged, thank you!

PR #10581, opened 2016-01-15
Add note about overriding get/set methods.

### LINKED ISSUE #10576: Behaviour magic get and set not working.
I am creating a behaviour to easily create extra fields in my ActiveRecord when needed. I have some ActiveRecords where i have many extra fields to add like mentioned here: http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#selecting-extra-fields

So to avoid creating each extra fields manually in every ActiveRecord, i am trying to create a behavior to achieve this, but the __get and __set are not working.

```php
class MyBehavior extends Behavior
{
    private $_dynamicAttributes = [];

    public function __get($name)
    {
        try {
            return parent::__get($name);
        } catch (UnknownPropertyException $exception) {
            return $this->getDynamicAttribute($name);
        }
    }

    public function __set($name, $value)
    {
        try {
            parent::__set($name, $value);
        } catch (UnknownPropertyException $exception) {
            $this->setDynamicAttribute($name, $value);
        }
    }

    protected function getDynamicAttribute($name)
    {
        return ArrayHelper::getValue($this->_dynamicAttributes, $name);
    }

    protected function setDynamicAttribute($name, $value)
    {
        $this->_dynamicAttributes[$name] = $value;
    }
}
```

Here is my try on this. If i add a `die()` at the top of either the `__get()` or `__set()`, none of them triggers. But let's assume i am trying to get an attribute named '`average`', the `__get()` will trigger if i have  the function `getAverage()` defined in the behavior, else it will just throw `UnknownPropertyException` without triggering the `__get()` or `__set()` of the behavior.

githubjeka:
Fix #10576


SilverFire:
Thank you. Could you update Russian guide, please?

samdark:
Going to merge it with some editorial.

cebe:
A note to `concept-properties.md` should be added to e.g.

> because of the properties introduced by yii, a normal call to `property_exists()` does not work to determine magic properties, you should call `canGetProperty()` or `canSetProperty()` respectively.

Issue #10574, opened 2016-01-14
BaseArrayHelper::htmlEncode() change


DrDeath72:
BaseArrayHelper::htmlEncode() must be equal to BaseHtml::encode()
```php
    public static function htmlEncode($data, $valuesOnly = true, $doubleEncode = true)
    {
        $d = [];
        foreach ($data as $key => $value) {
            if (!$valuesOnly && is_string($key)) {
                $key = htmlspecialchars($key, ENT_QUOTES | ENT_SUBSTITUTE, Yii::$app ? Yii::$app->charset : 'UTF-8', $doubleEncode);
            }
            if (is_string($value)) {
                $d[$key] = htmlspecialchars($value, ENT_QUOTES | ENT_SUBSTITUTE, Yii::$app ? Yii::$app->charset : 'UTF-8', $doubleEncode);
            } elseif (is_array($value)) {
                $d[$key] = static::htmlEncode($value, $valuesOnly, $doubleEncode);
            } else {
                $d[$key] = $value;
            }
        }

        return $d;
    }
```

samdark:
Verified.

cebe:
not necessarily the signature must be the same but the options passed to htmlspecialchars should be the same. This should be done in 2.0.x already. Would be a good idea to just call Html::encode() to have only one place that may change.

PR #10584, opened 2016-01-15
#10574 - Changed signature for BaseArrayHelper::htmlEncode and added doubleEncode parameter

andrewnester:
Implementing #10574

Signature for `BaseArrayHelper::htmlEncode` was changed (made similar to `BaseHtml::encode`), we should be careful with it.

In addition now `ENT_SUBSTITUTE` is used for `htmlspecialchars` .
Added new tests for `BaseArrayHelper::htmlEncode` .

### DIFF
--- a/framework/helpers/BaseArrayHelper.php
+++ b/framework/helpers/BaseArrayHelper.php
@@ -457,25 +457,23 @@ class BaseArrayHelper
      * @param array $data data to be encoded
      * @param boolean $valuesOnly whether to encode array values only. If false,
      * both the array keys and array values will be encoded.
-     * @param string $charset the charset that the data is using. If not set,
-     * [[\yii\base\Application::charset]] will be used.
+     * @param boolean $doubleEncode whether to encode HTML entities in `$data`. If false,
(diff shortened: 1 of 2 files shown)


samdark:
Signature change can't be merged into 2.0.x.

samdark:
@andrewnester can you adjust it so there's no signature change?

cebe:
I'm fine with changing signature in 2.1. So a separate PR would be good.

andrewnester:
what about if we split it on 2 PRs: one without breaking BC for version 2.0.x, and second one non-compatible for 2.1.0?
What do you think?

samdark:
That's what @cebe just suggested.

andrewnester:
@samdark oh, sorry, missed his reply. So it's better to adjust this one for 2.0.8 not to break BC and new one PR for 2.1.0, correct?

samdark:
Leave this as is for 2.1.0. Create another one for 2.0.8.

andrewnester:
Just added new PR for 2.0.8 #10592

Issue #10545, opened 2016-01-10
yii\rest\Controller - json vs xml output for an ActiveRecord model


amnah:
I noticed that the output from json and xml were different. It seems like json takes ```$user->toArray()``` into account while xml doesn't.

**Is this intentional?**

In my controller I have:

```php
public function actionIndex()
{
    // ... stuff
    /** @var /app/models/User $user */
    return ["success" => ["user" => $user, "userToken" => $userToken, "hasPassword" => $hasPassword]];
}
```

Here's the output. Notice that xml lists all attributes instead of the ones listed in ```User::fields()```

Json:

```json
{"success":{"user":{"id":2,"email":"neo2@neo.com","username":null},"userToken":null,"hasPassword":true}}
```

Xml: 

```xml
<response>
<success>
<user>
<id>2</id>
<role_id>2</role_id>
<status>1</status>
<email>neo2@neo.com</email>
<username/>
<password>
$2y$13$a6WLkpqd1vVOAarRyATaS.SAMTFLMJwnIgMJTxdFI.Io4EOA9NXly
</password>
<auth_key>0NOGxIG0qs5_RDIXBTv-0LcoCDlEThBC</auth_key>
<access_token>L7hI8R92ef93QbrEazWw_Y28KPme76j2</access_token>
<logged_in_ip/>
<logged_in_at/>
<created_ip>10.0.2.2</created_ip>
<created_at>2016-01-10 22:30:10</created_at>
<updated_at>2016-01-10 22:30:10</updated_at>
<banned_at/>
<banned_reason/>
</user>
<userToken/>
<hasPassword>1</hasPassword>
</success>
</response>
```

amnah:
The corresponding code is here:

https://github.com/yiisoft/yii2/blob/master/framework/web/XmlResponseFormatter.php#L80

It looks like it checks for ```Traversable``` first before ```is_object()``` (the latter would call ```toArray()```

PR #10589, opened 2016-01-15
#10545 - Fix for XMLResponseFormatter to format models in a proper way

andrewnester:
Fix for #10545

Now if data passed to formatter is `Arrayable`, then it is using `toArray()` method to get data for formatting.

Actually, it is kind of security fix because in corresponding issue we can see that in XML response return password hashed that is supposed to be hidden

### DIFF
--- a/framework/web/XmlResponseFormatter.php
+++ b/framework/web/XmlResponseFormatter.php
@@ -77,7 +77,9 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      */
     protected function buildXml($element, $data)
     {
-        if (is_array($data) || $data instanceof \Traversable && $this->useTraversableAsArray) {
+        if (is_array($data) ||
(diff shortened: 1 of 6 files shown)


framework/web/XmlResponseFormatter.php
```diff
@@ -77,6 +77,10 @@ public function format($response)
      */
     protected function buildXml($element, $data)
     {
+        if ($data instanceof Arrayable) {
```
SilverFire: What if we modify the condition below like the following:

```php
         if (is_array($data) || $data instanceof \Traversable && $this->useTraversableAsArray && !$data instanceof Arrayable) {
```

andrewnester: @SilverFire yeah, you are absolutely right, it will be better. Will send changes in a moment

andrewnester: @SilverFire changes has been done

SilverFire:
:+1: looks goo

SilverFire:
Thank you!

PR #10592, opened 2016-01-16
#10574 - (2.0.x compatible)  Fix for BaseArrayHelper::encode to work with invalid characters in proper way

andrewnester:
Implementing #10574 

Compatible with version 2.0.x (non-BC breaking)
BC breaking version is here #10584


SilverFire:
Looks good, I'm voting for merge in 2.0.7

samdark:
OK.

Issue #10533, opened 2016-01-08
Url validator passes incorrect values


sosnovski:
К примеру url вида http://ya.1ru.!@.!$%^ считается валидным. Это нормальное поведение?

For example URL like `http://ya.1ru.!@.!$%^` passes the validation. Is it a normal behavior?

SilverFire:
Verified. The url passes the validation, but it's inforrect.

SilverFire:
The regular expression captures only `http://ya.1ru`, leaving `.!@.!$%^` unmatched.

```pcre
/^(http|https):\/\/(([A-Z0-9][A-Z0-9_-]*)(\.[A-Z0-9][A-Z0-9_-]*)+)/i
```

Maybe, we should improve the regexp to check the domain name from the biggining till the first `/` or `?`

samdark:
Yep.

Shadowmad:
Good evening,

I have fixed this bug and did PR. https://github.com/yiisoft/yii2/pull/10600 Please review the code

PR #10600, opened 2016-01-18
Changed pattern of URL to match appropriate URL's

Shadowmad:
I have changed pattern of the url validator to match correct URLs so that error #10533 would not occure again. As explained in 10533, URLs are passing the tests, but URLs are not correct, this issue was encountered dut to the pattern issues. I ran PHPUnit tests and now everything is passing as well as checked URL in variety of programs to check if URL is correct or not. Please review and let me know if I should re do it or accepted

Thank you!

### DIFF
--- a/framework/validators/UrlValidator.php
+++ b/framework/validators/UrlValidator.php
@@ -28,7 +28,7 @@ class UrlValidator extends Validator
      * The pattern may contain a `{schemes}` token that will be replaced
      * by a regular expression which represents the [[validSchemes]].
      */
-    public $pattern = '/^{schemes}:\/\/(([A-Z0-9][A-Z0-9_-]*)(\.[A-Z0-9][A-Z0-9_-]*)+)/i';
+    public $pattern = '/^({schemes}:\\/\\/)(www\\.)?([0-9u0000-uFFFF]*\\.\\-)?([0-9u0000-uFFFF\\-]*)\\.([a-zA-Z]{2,6})?([u0000-uFFFF]{2,6})\\/?([u0000-uFFFF0-9_\\-\\?=\\.\\&])*?/i';
     /**
      * @var array list of URI schemes which should be considered valid. By default, http and https
      * are considered to be valid schemes.
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.7 under development
 -----------------------
-
+- Bug #10533: URL validator passed the tests but allowed incorrect values. Changed Patter to more reliable one with URI as options(Shadowmad; Max)
 - Bug #6351: Find MySQL FK constraints from `information_schema` tables instead of `SHOW CREATE TABLE` to improve reliability (nineinchnick)
 - Bug #6363, #8301, #8582, #9566: Fixed data methods and PJAX issues when used together (derekisbusy)
 - Bug #6876: Fixed RBAC migration MSSQL cascade problem (thejahweh)


SilverFire:
Thank you for your effort.
Could you add a test case for original problem? (http://ya.1ru.!@.!$%^)

framework/validators/UrlValidator.php
```diff
@@ -28,7 +28,7 @@ class UrlValidator extends Validator
      * The pattern may contain a `{schemes}` token that will be replaced
      * by a regular expression which represents the [[validSchemes]].
      */
-    public $pattern = '/^{schemes}:\/\/(([A-Z0-9][A-Z0-9_-]*)(\.[A-Z0-9][A-Z0-9_-]*)+)/i';
+    public $pattern = '/^({schemes}:\\/\\/)(www\\.)?([0-9u0000-uFFFF]*\\.\\-)?([0-9u0000-uFFFF\\-]*)\\.([a-zA-Z]{2,6})?([u0000-uFFFF]{2,6})\\/?([u0000-uFFFF0-9_\\-\\?=\\.\\&])*?/i';
```
SilverFire: Why did you add additional `\` everywhere?

SilverFire: Why did you add `(www\\.)?` as a separated case?

SilverFire: `0-9u0000-uFFFF` what is it? Very strange condition

SilverFire:
The Regexp is overcomplicated and must be simplified.

framework/validators/UrlValidator.php
Shadowmad: Good evening,

I added www as URL might have www in their name. It is optional, I can
remove that, just thought that would be good addition.
Also, double slash to worked for me in PHP as ultimate regular expression
work. Previously I had troubles with no double slash.
I can create test case for original problem. That is not a problem.

Please let me know if that is good and I do the fixes asap.

Thank you!

Best Regards,
Maksud Aghayev

*Maksud Aghayev *| Full LAMP stack web application developer/PHP Web
Developer
Linkedin: *https://www.linkedin.com/in/maksudaghayev
<https://www.linkedin.com/in/maksudaghayev>*

SilverFire: > I added www as URL might have www in their name. It is optional, I can
> remove that, just thought that would be good addition.

The regular expression MUST allow `www.domain.com` as a regular case, 'cause it's the same as `subdomain.domain.com`

SilverFire: > Also, double slash to worked for me in PHP as ultimate regular expression
> work. Previously I had troubles with no double slash.

Previously we had no troubles without double backslash. It's not necessary to add it, because it does not fix or enhance anything

Shadowmad: I can fix double back-slash issue, that not a problem. Regarding www and subdomain: what if the url goes as follows www.subdomain.domain.com? Will regular expression still work? With the one I wrote it works by checking www and subdomain. Could you clarify that please

Shadowmad: Also, u0000-uFFFF is all possible Unicode. I can change that to 0-9a-zA-Z to limit regular expression to only English words.

SilverFire: > www.subdomain.domain.com? Will regular expression still work

yes, it's a valid URL

SilverFire: We need to respect `enableIDN` option

samdark: Unicode ranges could be replaced with `\pL`

cebe:
The regex itself should not allow UTF8 characters, utf8 is handled by converting the IDN name first and then checking against the pattern.

samdark:
@cebe is right about it.

samdark:
I propose fixing it by adding `(:?$|/)` to the end of existing regex.

Shadowmad:
@samdark That might help with escaping issues of not recognized symbols, byt what about patterns like https://www.google.de/search?q=yii+framework&ie=utf-8&oe=utf-8.&rls=org.mozilla:de:official&client=firefox-a&gws_rd=cr

How do you work them around? Trough regexp or trough some system in the code? I apologize in advance, I am still trying to figure out the code and dependencies in the whole system

I came up with something like this, taking in cosideration what you had and suggestions of @samdark 
^(http|https):\/\/(([A-Z0-9][A-Z0-9_-]*)(\.[A-Z0-9][A-Z0-9_-]*)+)([\.\/\?\w\d=\_\-&\%\+\:])*

Could you please provide feedback on that and I would update my PR as well as add test cases as requested above. 

Thank you

SilverFire:
> [\.\/\?\w\d=\_\-&\%\+\:]

I don't like this part. All we need to do - to stop matching _domain name_ part of the string. We mustn't take care anything else in the URL.

The original problem was a domain `http://ya.1ru.!@.!$%^`, that is not a valid domain, for sure, but `http://ya.1ru/!@.!$%^` might be a valid one.

I'm suggesting the following regexp:

`/^(http|https):\/\/((?:[A-Z0-9][A-Z0-9_-]*)(?:\.[A-Z0-9][A-Z0-9_-]*)+)(?::?|\/|\?)/i`

`(?::?|\/|\?)` means stop matching domain name on `:`, `/` or `?` character.

Shadowmad:
I apologize for inconvenience. I totally misunderstood the problem. If I
need to cut out URL as only to domain name part then proposed fix is better
one, I thought that I need to check for possible folder names, variables or
"pretty" URLs. I will take that into consideration tomorrow and apply all
the fixed if that is still applicable.

Thank you for guidance!

Best Regards,
Maksud Aghayev

*Maksud Aghayev *| Full LAMP stack web application developer/PHP Web
Developer
Linkedin: *https://www.linkedin.com/in/maksudaghayev
<https://www.linkedin.com/in/maksudaghayev>*

SilverFire:
Closed in favor or #10753

Issue #8348, opened 2015-05-11
PHP Fatal Error: Nesting level too deep - recursive dependency?


callmez:
```
\Yii::$app->getDb()->createCommand('CREATE TABLE IF NOT EXISTS `test` (type INT(11))')->execute();
$models = [
    new \app\models\Test(['type' => 0]),
    new \app\models\Test(['type' => 0]),
];
\yii\helpers\ArrayHelper::multisort($models, 'type', SORT_ASC);
```
Test  model file
```
<?php
namespace app\models;

use yii\db\ActiveRecord;
use yii\behaviors\TimestampBehavior;

class Test extends ActiveRecord
{
    public static function tableName()
    {
        return 'test';
    }

    public function behaviors()
    {
        return [
            'timestamp' => TimestampBehavior::className()
        ];
    }
}
```

Get error

```
PHP Fatal Error – yii\base\ErrorException

Nesting level too deep - recursive dependency?
```

callmez:
related https://github.com/yiisoft/yii/issues/2082

cebe:
Reproduced. Only happens if timestamp behavior is added. The error happens inside of PHP so it is hard to figure out the cause...

cebe:
Reproduced with plain PHP:

```php
<?php

class Related
{
	public $owner;
}

class Model
{
	public $related;
	public $type = 0;

	public function __construct($type)
	{
		$this->related = new Related();
		$this->related->owner = $this;

		$this->type = $type;
	}
}

$models = array(
	0 => new Model(1),
	1 => new Model(1),
);

$args = [
	[0 => 1, 1 => 1],
	SORT_ASC,
	SORT_REGULAR,
];
$args[] = &$models;

var_dump(call_user_func_array('array_multisort', $args));
print_r($models);
```

As far as I see the error results from the fact that the given ordering does not result in an ordering so PHP will compare the objects directly and this results in the error:

```php
var_dump($models[0] < $models[1]);
```

Seems that the fix is to force a default ordering if given columns do not specify an ordering:
https://github.com/yiisoft/yii/issues/2082#issuecomment-38397101
need to verify this.

Bhoft:
the same also happens also with the AttributeBehavior.
are they any news how to fix this issue or when it will be resolved?

zebratuk:
[...]

PR #10607, opened 2016-01-18
8348 - Fixed processing nested objects multisort in BaseArrayHelper 

andrewnester:
Fix for #8348 

When we have equal values for main sorted columns it will take in consideration added column of consecutive values and that's why objects themselves will not be compared between each other


SilverFire:
It looks weird, but it works... Did you find this approach somewhere or it your's own?

andrewnester:
@SilverFire I've seen this trick couple of years ago in some basic algorithm on C++ but can't remember what it was. If I will remember will let you know.

cebe:
@andrewnester Looks interesting. And on first sight makes sense. Thanks! Will check it in detail later.

cebe:
Merged, thank you!

PR #10610, opened 2016-01-19
允许在使用Url::to和Url::toRoute生成Url的时候指定urlManager。

cdcchen:
在使用Advanced Template 模板架构的时候会有很多应用，有时候会需要生成非当前应用的url，添加$urlManager参数可以很方便的达到此目的。

### DIFF
--- a/framework/helpers/BaseUrl.php
+++ b/framework/helpers/BaseUrl.php
@@ -85,15 +85,16 @@ class BaseUrl
      * @return string the generated URL
      * @throws InvalidParamException a relative route is given while there is no active controller
      */
-    public static function toRoute($route, $scheme = false)
+    public static function toRoute($route, $scheme = false, $urlManager = null)
     {
         $route = (array) $route;
         $route[0] = static::normalizeRoute($route[0]);
 
+        $urlManager = $urlManager ?: Yii::$app->getUrlManager();
         if ($scheme) {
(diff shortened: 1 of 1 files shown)


samdark:
Can't apply the change in 2.0 for sure since it breaks backwards compatibility.

samdark:
How about introducing a class var $urlManager that could be set manually?

cebe:
> How about introducing a class var $urlManager that could be set manually?

:+1:

githubjeka:
An update is required for the phpDoc.

samdark:
@githubjeka we aren't going to do it the way proposed.

vistart:
> How about introducing a class var $urlManager that could be set manually?

@samdark @cebe 
How about this:[MultiDomainsManager](https://github.com/vistart/yii2-models/blob/master/components/MultiDomainsManager.php)

The `$subDomains` property contains all url manager configuration array and its scheme.

samdark:
@vistart it's good for that specific use case but since use case isn't common, we'll better introduce a way to solve it rather than provide a full featured solution.

vistart:
@samdark 
Yeah, the [MultiDomainsManager] is designed for web application which has a common base domain, like `a.example.com` and `b.example.com` contain the base domain `example.com`, not suitable for variable domains.

Issue #10555, opened 2016-01-12
ActiveForm multiple submit buttons and enableClientScript enabled, when output outside browser


tsdogs:
Generating an ActiveForm with multiple submit buttons and rendering the output based on the button value will keep last pressed button value active if output is not HTML (es. PDF)

To reproduce:
```php
// view
    <?php $form = ActiveForm::begin([ 
        'type' => ActiveForm::TYPE_INLINE, 
        'method' => 'GET',
        'action' => Url::to(['model/action']), 
        'enableClientScript' => true, // default
    ]); ?>
    <?=$form->field($model, 'date')->textInput(); ?>
    <?= Html::submitButton('View',[ 'name'=>'view', 'value' => 'view', 'class' => 'btn btn-default','data-pjax' => 0]) ?>
    <?= Html::submitButton('Print',[ 'name'=>'print', 'value' => 'print', 'class' => 'btn btn-default','data-pjax' => 0]) ?>
    <?php ActiveForm::end(); ?>

// controller action

public function actionAction()
{
    if (isset($_GET['print')) {
       // output something that is not HTML. i.e PDF or XLS
       \Yii::$app->response->format = \yii\web\Response::FORMAT_RAW;
       return $pdfcontent;
    } else if (isset($_GET['view'])) {
       $this->render('view', [ 'model' => $model ]);
    }
}
```
If you press the view button everything is fine. But if you press the Print button and the PDF file opens outside the browser (not with the internal pdf viewer), then after this every time you click on the View button the Print button action gets triggered.
This is because the client javascript is generating an hidden input with the "print" value before submitting, and the page does not get refreshed (as it opens outside the broswer)

the incriminated code is in the javascript which creates the hidden input.

andrewnester:
Reproduced in Chrome.
You need to disable Chrome PDF plugin first like this http://chris.dziemborowicz.com/blog/2010/12/12/how-to-disable-chrome's-built-in-pdf-viewer/

After this when you click Print it will download pdf file without page reloading

SilverFire:
@andrewnester that's not a solution.

I the problem is in submit button name. We have a code in [yii.activeForm.js:551](https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.activeForm.js#L581), which searches for a hidden input with name eq to submit button name and creates one, if it does not exist yet. But if the input exists, the script sets its value to the button value.

[...]

Issue #10322, opened 2015-12-03
ActiveForm prevents button to open in new window (formtarget="_blank")


twinni:
The **formtarget** directive is not working when the button is part of an ActiveFormorm.

```php
// This example does not work.  It submits the form data to the current window.
<?php $form = ActiveForm::begin(['method' => 'get']); ?>
        First name: <input type="text" name="fname"><br>
        Last name: <input type="text" name="lname"><br>
        <button type="submit" >Submit</button>
        <button type="submit" formtarget="_blank">Submit to a new window</button>
    <?php ActiveForm::end(); ?> 
```

```php
// This example works well. It submits the form data to a new window.
 <form id="w0"  method="get"> 
        First name: <input type="text" name="fname"><br>
        Last name: <input type="text" name="lname"><br>
        <button type="submit" >Submit</button>
        <button type="submit" formtarget="_blank">Submit to a new window</button>
  </form> 
```

samdark:
Correct. `formtarget` isn't handled now in any way.

twinni:
@samdark , isn't it a bug instead of an enhancement? The generated code for the button is ok. I assume that the activeForm.js handles the sumbit function without considering ``formtarget``.
Anyway, when you have a solution it would be great if you could solve it in 2.0.7.

SilverFire:
I think it's an enhancement for JS ActiveForm plugin to support the `formtarget` attribute.
We are preparing the release, so 2.0.7 milestone is closed for new non-critical issues and PR.

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/fbedb8e750dbb752e43af718f4692c7d5d5e43d7

PR #10625, opened 2016-01-20
10555 - Fix for ActiveForm - When previous action doesn't reload page, this action will be triggered again when another submit button clicked

andrewnester:
Updating hidden field that represents submit button was moved to separate method `updateHiddenButton`

This method now called in `submitForm` method after data is validated and valid.

It is used for cases when we have multiple submit buttons on one active form.
It relates to #10555

Also this fix is important for #10322 because otherwise after we click on Submit to a new window button it will be opened in new window but the page with form will not be reloaded. If after this we click on Submit button, form will be loaded in new window anyway.

This fix solves these problems by updating hidden field every time submit button clicked and form is sent

### DIFF
--- a/framework/assets/yii.activeForm.js
+++ b/framework/assets/yii.activeForm.js
@@ -393,6 +393,7 @@
                     submitFinalize($form);
                     return false;
                 }
+                updateHiddenButton($form);
                 return true;   // continue submitting the form since validation passes
             } else {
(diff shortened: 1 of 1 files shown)


SilverFire:
Thank you!

SilverFire:
Added CHANGELOG 969469b55aae1a16b889cb7e9b97af8954280b14

Issue #10629, opened 2016-01-20
BaseStringHelper::truncateHtml adds suffix regardless of the string length


mattheobjornson:
Hi,

I am using StringHelper::truncateWords (Yii2 v2.0.6) and found out that $suffix always gets added to returned string if $asHtml is set to true, regardless of fact that the string was not actually truncated. This differs from the behaviour of the same function when $asHtml is set to false. 
I would expect it to behave consistently and return the same string if it was not truncated (either with suffix, or without).
```php
$str='lorem ipsum';

StringHelper::truncateWords($str, 3, '...', true);
// returns "lorem ipsum..."

StringHelper::truncateWords($str, 3, '...', false)
// returns "lorem ipsum"
```
For now I have overridden BaseStringHelper::truncateHtml(...)
```php
return $generator->generateFromTokens($truncated) . $suffix;
```
to
```php
return $generator->generateFromTokens($truncated) . ($totalCount >= $count ? $suffix : '');
```
which seems to do the job.
Maybe should be considered in upstream.

andrewnester:
Hi @mattheobjornson 
I've created PR with fix provided by you. Thanks!

PR #10633, opened 2016-01-21
10629 - BaseStringHelper::truncateHtml fix - should not add suffix if words count less than needed

andrewnester:
Fix for #10629


SilverFire:
Thank you, merged

Issue #10535, opened 2016-01-08
Restriction in order By


llfm:
How can I add a "custom" order by?
I can't write this:
```php
$query->orderBy('CASE WHEN x = y THEN 0 ELSE 1 END');
```

Shouldn't this [method](http://www.yiiframework.com/doc-2.0/yii-db-querytrait.html#orderBy%28%29-detail) read the string as it is? Like the `where` method.

samdark:
Have you tried wrapping a string with `\yii\db\Expression`?

llfm:
This?
```php
$query->orderBy(new \yii\db\Expression('CASE WHEN x = y THEN 0 ELSE 1 END'));
```
 It doesn't work. The final guilty is [guy](https://github.com/yiisoft/yii2/blob/master/framework/db/QueryBuilder.php#L808).

samdark:
Right. It works with columns only now...

cebe:
It should work when using array syntax: `$query->orderBy(['CASE WHEN x = y THEN 0 ELSE 1 END' => SORT_ASC]);`

But we may allow expression too.

andrewnester:
@cebe I've checked, actually it already allows using `Expression` for order by.
following test is passed successfully
```
$query->orderBy(new Expression('CASE WHEN x = y THEN 0 ELSE 1 END'));
$this->assertEquals(['CASE WHEN x = y THEN 0 ELSE 1 END' => SORT_ASC], $query->orderBy);
```

Expressions are processing here https://github.com/yiisoft/yii2/blob/master/framework/db/QueryBuilder.php#L801

So I guess there is no need to change anything

samdark:
Yes, a test would not hurt.

cebe:
related to #8824

PR #10637, opened 2016-01-21
10535 -  \yii\db\Expression with complex expression (aggregate functions and etc) for Order By clause

andrewnester:
Added unit test for #10535

### DIFF
--- a/framework/db/QueryTrait.php
+++ b/framework/db/QueryTrait.php
@@ -341,7 +341,7 @@ trait QueryTrait
         if (is_array($columns)) {
             return $columns;
         } else {
-            $columns = preg_split('/\s*,\s*/', trim($columns), -1, PREG_SPLIT_NO_EMPTY);
+            $columns = preg_split('/\s*,(?![^\(]*\))\s*/', trim($columns), -1, PREG_SPLIT_NO_EMPTY);
             $result = [];
             foreach ($columns as $column) {
                 if (preg_match('/^(.*?)\s+(asc|desc)$/i', $column, $matches)) {
--- a/framework/db/QueryTrait.php
+++ b/framework/db/QueryTrait.php
@@ -305,7 +305,11 @@ trait QueryTrait
      */
     public function orderBy($columns)
     {
-        $this->orderBy = $this->normalizeOrderBy($columns);
+        if ($columns instanceof Expression) {
+            $this->orderBy = [$columns];
+        } else {
+            $this->orderBy = $this->normalizeOrderBy($columns);
+        }
         return $this;
     }
 
@@ -321,7 +325,12 @@ trait QueryTrait
      */
     public function addOrderBy($columns)
     {
-        $columns = $this->normalizeOrderBy($columns);
+        if ($columns instanceof Expression) {
+            $columns = [$columns];
+        } else {
+            $columns = $this->normalizeOrderBy($columns);
+        }
+
         if ($this->orderBy === null) {
             $this->orderBy = $columns;
         } else {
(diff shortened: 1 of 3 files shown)


tests/framework/db/QueryTest.php
```diff
@@ -166,6 +167,9 @@ public function testOrder()
 
         $query->addOrderBy('age ASC, company DESC');
         $this->assertEquals(['team' => SORT_ASC, 'company' => SORT_DESC, 'age' => SORT_ASC], $query->orderBy);
+
+        $query->orderBy(new Expression('CASE WHEN x = y THEN 0 ELSE 1 END DESC'));
```
cebe: this will internally be cased to a string and not fail as it does not contain a `,` if you add an `Expression` that contains a comma, it should fail.

anestor-sugar: I tried to run following test on my local machine and it passed successfully
```
$query->orderBy(new Expression('CASE WHEN x = y THEN 0 ELSE 1 END DESC, x ASC'));
$this->assertEquals(['CASE WHEN x = y THEN 0 ELSE 1 END' => SORT_DESC, 'x' => SORT_ASC], $query->orderBy);
```

andrewnester: I tried to run following test on my local machine and it passed successfully
```
$query->orderBy(new Expression('CASE WHEN x = y THEN 0 ELSE 1 END DESC, x ASC'));
$this->assertEquals(['CASE WHEN x = y THEN 0 ELSE 1 END' => SORT_DESC, 'x' => SORT_ASC], $query->orderBy);
```

I can update my PR to include this test instead previous one

samdark: Yep. Would be great.

cebe:
Even this case works as expected, what does not work is:

```php
$query->orderBy(new Expression('SUBSTR(name, 3, 4) DESC'));
$this->assertEquals(['SUBSTR(name, 3, 4)' => SORT_DESC], $query->orderBy);
```
and
```php
$query->orderBy(new Expression('SUBSTR(name, 3, 4) DESC, x ASC'));
$this->assertEquals(['SUBSTR(name, 3, 4)' => SORT_DESC, 'x' => SORT_ASC], $query->orderBy);
```

andrewnester:
@cebe You are absolutely right! I've added new commit with changes to regular expression used in `normalizeOrderBy` and updated tests. Now aggregate functions in `ORDER BY` clause are processed fine

framework/db/QueryTrait.php
```diff
@@ -341,7 +341,7 @@ protected function normalizeOrderBy($columns)
         if (is_array($columns)) {
             return $columns;
         } else {
-            $columns = preg_split('/\s*,\s*/', trim($columns), -1, PREG_SPLIT_NO_EMPTY);
+            $columns = preg_split('/\s*,(?![^\(]*\))\s*/', trim($columns), -1, PREG_SPLIT_NO_EMPTY);
```
samdark: Why do we need grouping here?

andrewnester: Actually, it's not grouping. it's `if-else` syntax for regular expression
http://www.regular-expressions.info/conditional.html

samdark: Ah, that's negative lookahead. Haven't used it for a while. Verified that solution works well.

samdark:
This one works well. We can merge it either for 2.0.7 or 2.0.8. @cebe ?

cebe:
When an Expression object is passed I'd prefer it to be kept as is (not casting to string) until it reaches query builder. This would be constistent with how Expressions are handled when passed in an array:

https://github.com/yiisoft/yii2/blob/c79ebf49f4e9ae28a570ac44e0d6b3015b00dc74/framework/db/QueryBuilder.php#L801

A proper fix would be to check for `$columns instanceof Expression` and then `return [$columns];`

andrewnester:
@cebe I've just added new commit with fix you offered

cebe:
Merged with some adjustments. Thank you!

PR #10638, opened 2016-01-21
 Bug: `yii\validators\UniqueValidator` doesn't work when `targetClass` is specified as class from global scope of namespace .

githubjeka:
See https://github.com/yiisoft/yii2/pull/10263#issuecomment-173562767

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -106,7 +106,7 @@ class UniqueValidator extends Validator
             $query->andWhere($this->filter);
         }
 
-        if (!$model instanceof ActiveRecordInterface || $model->getIsNewRecord() || $model->className() !== $targetClass) {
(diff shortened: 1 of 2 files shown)


framework/validators/UniqueValidator.php
```diff
@@ -106,7 +106,7 @@ public function validateAttribute($model, $attribute)
             $query->andWhere($this->filter);
         }
 
-        if (!$model instanceof ActiveRecordInterface || $model->getIsNewRecord() || $model->className() !== $targetClass) {
+        if (!$model instanceof ActiveRecordInterface || $model->getIsNewRecord() || $model->className() !== $targetClass::className()) {
```
cebe: isn't calling `ltrim(..., '\\')` enough here?

SilverFire: `::className()` is a the most reliable way, as for me. Since we load the class on line 100, extra `className()` does not impact the performance.

SilverFire:
Thank you!

PR #10642, opened 2016-01-23
Fixed a wrong method calling.

drodata:
If using `all()`, the value of `$customer` will be an array of
`Customer` instances, not a `Customer` instance.  Consequently,

```
$customer->unlink('orders', $customer->orders[0]);
```

will throw a PHP Fatal Error: Call to a member function unlink() on array.


SilverFire:
Thank you!

SilverFire:
I've also fixed the line by adding a where condition and copy-pasted it to all the translations of this guide page. See https://github.com/yiisoft/yii2/commit/89f677ba2abb6783d1059da83728b77b0004c3db

Issue #10649, opened 2016-01-24
i18n strange behaviour


turalsprofile:
# Here is actual problem:

i18n parser completely ignores example below:
```
  <?
            $html = "";
            if ($vacancies) {
                $html .= Html::tag("h2", Yii::t("vacancies", "We have following vacancies:"));

            } else {
                $html .= Html::tag("h2", Yii::t("vacancies", "Currently we have no active vacancies. But you can send your CV to our email address:"));
            }
            echo $html;
            ?>
```

I'm using command below to generate messages:

```
 ./yii message/extract messages-config.php
```

And here is `messages-config.php`'s contents

```
<?php

return [
    'sourcePath' => __DIR__,
    'languages' => ['az', 'en', 'ru', 'tr'],
    'translator' => 'Yii::t',
    'sort' => true,
    'removeUnused' => false,
    'only' => ['*.php'],
    'except' => [
        '.svn',
        '.git',
        '.gitignore',
        '.gitkeep',
        '.hgignore',
        '.hgkeep',
        '/messages',
        '/vendor',
        '/backend',
        '/environments',
        '/tests'
    ],

    'format' => 'db',
    'overwrite' => true,
    'sourceMessageTable' => 'TranslationSourceMessage',
    'messageTable' => 'TranslationMessage',
];
```

I'm trying to figure out why this problem happen, when I'm doing everything based on Yii documentation?

samdark:
`@@` means outdated.

turalsprofile:
@samdark so AFAIK it's because of removeUnused turned off?

samdark:
Yes.

samdark:
So the issue is about ignoring the piece of code you've provided. It looks OK so I think parse has a bug.

turalsprofile:
@samdark oh god, don't make us wait for 2.0.8, man. :) it's for project which is in production currently. Please fix it ASAP. Hope for understanding

samdark:
You can speed it up by fixing it yourself and providing a pull request. Or at least by providing extra unit test with your example that fails.

andrewnester:
Hi guys! I've found why the issue is happened, looks like @turalsprofile is using short php tags `<?` instead of `<?php`. That's why parser does not work. I am gonna fix this

turalsprofile:
@andrewnester yes you're right, changing <? to <?php fixed problem. @samdark please fix based on Andrew's suggestion. Thank you all, guys.

andrewnester:
[...]

PR #10654, opened 2016-01-25
10649 - Fix for parsing short tags when extracting messages #10649

andrewnester:
Fix for #10649 

Solution is to parse file content, find short tags and replace them with full one.

### DIFF
--- a/framework/console/controllers/MessageController.php
+++ b/framework/console/controllers/MessageController.php
@@ -429,6 +429,8 @@ EOD;
         $this->stdout("Extracting messages from $coloredFileName...\n");
 
         $subject = file_get_contents($fileName);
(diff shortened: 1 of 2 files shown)


framework/console/controllers/MessageController.php
```diff
@@ -429,6 +429,8 @@ protected function extractMessages($fileName, $translator, $ignoreCategories = [
         $this->stdout("Extracting messages from $coloredFileName...\n");
 
         $subject = file_get_contents($fileName);
+        $subject = preg_replace("/<\?[^a-zA-Z0-9\=]*/", '<?php ', $subject);
```
klimov-paul: This is risk prone: you replaces any appearance of `<?` to `<?php`.
This means if there is a translation message like:
```php
Yii::t('app', 'You can use short tag like "<?"');
```
it will be broken.

Issue #10667, opened 2016-01-26
Caption for DetailView Widget


gbespyatykh:
Missed caption for table rendered by DetaiView widget

SilverFire:
Could you described your problem in detail?

gbespyatykh:
The caption and captionOptions widget options isn`t presented in DetailView

klimov-paul:
> caption for table

This does not make any sense: table does not have caption

gbespyatykh:
Caption tag presented in HTML specification: http://www.w3schools.com/tags/tag_caption.asp
Also yii/grid/GridView has the caption and captionOptions in widget properties

gbespyatykh:
Okay, then I should use 'foreach' instead of widget...

klimov-paul:
```php
<?= DetailView::widget([
        'model' => $model,
        'attributes' => [
            [
                'label' => 'My Caption Label',
                'value' => 'My Caption Value',
            ],
            'id',
            'name',
        ],
    ]) ?>
```

gbespyatykh:
Yes, of course. But this is bad solution, because you must also specify colspan attribute at row to merge columns. CSS styles for columns and rows will affect on this caption made by tr and td tags. Caption is easer for styling.

Why caption is used in the GridView widget, but not used in DetailView?

PR #10668, opened 2016-01-26
Added caption for yii\widgets\DetailView

gbespyatykh:
#10667


klimov-paul:
This does not make any sense: you can add any extra rows to the grid view using `attributes` already.

PR #10685, opened 2016-01-28
Throw an exception once http headers alreay been sent by other script while Yii2 headers is not empty

aiddroid:
throw an exception once http headers alreay been sent by other script while Yii2 headers is not empty

### DIFF
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -336,10 +336,10 @@ class Response extends \yii\base\Response
      */
     protected function sendHeaders()
     {
-        if (headers_sent()) {
-            return;
-        }
         if ($this->_headers) {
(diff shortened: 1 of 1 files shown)


SilverFire:
Could you please add a unit test?

klimov-paul:
Does not make much sense for me: `header()` function will trigger PHP error if headers have already been sent anyway.
Current change simply destroying any ability to use `Response` in unit tests, which is proved by broken Travis built.

aiddroid:
Sorry,I am new here,I noticed that if some default headers sent by server(such as php session_cache_limiter's Cache-Control),then the headers set by Yii2 will disappear without any notice,I just want to fix it.

PR #10694, opened 2016-01-28
Fixes yiisoft/yii2#10692

### LINKED ISSUE #10692: Problem with $model->loadDefaultValues() and DEFAULT NULL in PostgreSQL
We have a bug here with Yii2 2.0.6 and our PostgreSQL database. When a column is set with `DEFAULT NULL`, Yii2 puts a string `NULL` (NULL written as a text) in the attribute after we call `$model->loadDefaultValues()`.

Example:

```SQL
CREATE TABLE "test"
(
	"id" SERIAL PRIMARY KEY,
	"name" varchar(255) DEFAULT NULL
);
```

and

```php
$model = new Test;
$model->loadDefaultValues();
var_dump($model->name);
```

The output is

`string(4) "NULL"`

The expected output would be

`NULL`

<br>
I believe the problem is in yii\db\pgsql\Schema:

```php
protected function findColumns($table)
{
    ...
    foreach ($columns as $column) {
        ...
        if ($column->isPrimaryKey) {
            ...
        } elseif ($column->defaultValue) {
            if ($column->type === 'timestamp' && $column->defaultValue === 'now()') {
                ...
            } elseif ($column->type === 'boolean') {
                ...
            } elseif (stripos($column->dbType, 'bit') === 0 || stripos($column->dbType, 'varbit') === 0) {
                ...
            } elseif (preg_match("/^'(.*?)'::/", $column->defaultValue, $matches)) {
                ...
            } elseif (preg_match("/^(.*?)::/", $column->defaultValue, $matches)) {
                $column->defaultValue = $column->phpTypecast($matches[1]);
            } else {
                ...
            }
        }
    }

    return true;
}
```

In our case, the value of `$column->defaultValue` comes as a string `NULL::character varying`.

wallysalami:
Fixes #10692

### DIFF
--- a/framework/db/pgsql/Schema.php
+++ b/framework/db/pgsql/Schema.php
@@ -412,7 +412,11 @@ SQL;
                 } elseif (preg_match("/^'(.*?)'::/", $column->defaultValue, $matches)) {
                     $column->defaultValue = $matches[1];
                 } elseif (preg_match('/^(.*?)::/', $column->defaultValue, $matches)) {
-                    $column->defaultValue = $column->phpTypecast($matches[1]);
+                    if ($matches[1] === 'NULL') {
+                        $column->defaultValue = null;
+                    } else {
+                        $column->defaultValue = $column->phpTypecast($matches[1]);
+                    }
                 } else {
                     $column->defaultValue = $column->phpTypecast($column->defaultValue);
                 }


cebe:
This conflicts with #9773, more specifically https://github.com/yiisoft/yii2/pull/9773/files#diff-03b64c04d37c13a3118b4b8119f5ae8dR420 so it should be considered when merging.

wallysalami:
I think #9773 has already fixes my problem.

```php
private function getColumnDefaultValue($column, $value)
{
     if (!strcasecmp($value, 'null') || $value === '' || $value === null) {
          return null;
     }
    ...
}
```

So, are you going to accept that merge instead of mine? That would be fine by me as well. But it looks like there is a long conversation there since September.

cebe:
yes, that PR is quite complex. I will leave both open until we decided which to merge. If  #9773 will not be included in 2.0.7 we can merge your fix imo.

wallysalami:
All right! I'll follow that PR as well.

wallysalami:
Oh, I've forgot to give @gabrielhomsi some credit for finding the problem.

samdark:
#9773 isn't going to 2.0.7

cebe:
Merged, thank you!

PR #10698, opened 2016-01-29
Fixed typo in docs [skip ci]

arogachev:


arogachev:
Actually according to dictionary form without `h` is also acceptable but less widespread.

SilverFire:
Thank you

PR #10707, opened 2016-01-29
add types,commands,migrations for work with postgresql DBMS

kruatech:
Add types and functions for work with postgresql DBMS.

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -529,6 +529,42 @@ class QueryBuilder extends \yii\base\Object
     }
 
     /**
+     * Creates a SQL command for creating a new postgresql index.
+     * @param string $name the name of the index. The name will be properly quoted by the method.
+     * @param string $table the table that the new index will be created for. The table name will be properly quoted by the method.
+     * @param string|array $columns the column(s) that should be included in the index. If there are multiple columns, please separate them
+     * by commas. The column names will be properly quoted by the method.
+     * @param string $method to set access method can be:btree,hash,gist,gin,spgist
+     * @return $this the command object itself
+     * This function can be used only postgresql.
+     */  
+    public function createPgIndex($name, $table, $columns, $method = 'btree')
+    {
+        return ('CREATE INDEX ')
+            . $this->db->quoteTableName($name) . ' ON '
+            . $this->db->quoteTableName($table). ' USING '
+            . $this->db->quoteTableName($method)   
+            . ' (' . $this->buildColumns($columns) . ')';
+    }
+    
+    /**
+     * Creates a SQL command for creating a new unique constraint index.
+     * @param string $name the name of the index. The name will be properly quoted by the method.
+     * @param string $table the table that the new index will be created for. The table name will be properly quoted by the method.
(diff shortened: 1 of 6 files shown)


framework/db/QueryBuilder.php
```diff
@@ -529,6 +529,42 @@ public function createIndex($name, $table, $columns, $unique = false)
...
+     * Creates a SQL command for creating a new postgresql index.
+     * @param string $name the name of the index. The name will be properly quoted by the method.
+     * @param string $table the table that the new index will be created for. The table name will be properly quoted by the method.
+     * @param string|array $columns the column(s) that should be included in the index. If there are multiple columns, please separate them
+     * by commas. The column names will be properly quoted by the method.
+     * @param string $method to set access method can be:btree,hash,gist,gin,spgist
+     * @return $this the command object itself
+     * This function can be used only postgresql.
+     */  
+    public function createPgIndex($name, $table, $columns, $method = 'btree')
```
klimov-paul: How this is different from `yii\db\pgsql\QueryBuilder::createIndex()` ?

framework/db/QueryBuilder.php
```diff
@@ -529,6 +529,42 @@ public function createIndex($name, $table, $columns, $unique = false)
...
+     * by commas. The column names will be properly quoted by the method.
+     * @return $this the command object itself
+     * This function can be used only postgresql.
+     */  
+    public function createPgUnique($name, $table, $columns)
+    {
+        return ('ALTER TABLE ')
+            . $this->db->quoteTableName($table). ' ADD CONSTRAINT '
+            . $this->db->quoteTableName($name) . ' UNIQUE '            
+            . ' (' . $this->buildColumns($columns) . ')';
```
klimov-paul: Does result of such DDL will be different from `CREATE UNIQUE INDEX `?

framework/db/QueryBuilder.php
diselop: method access, in postgresql can be btree,hash,gist,gin,spgist.

framework/db/Schema.php
```diff
@@ -57,6 +57,45 @@
     const TYPE_BINARY = 'binary';
     const TYPE_BOOLEAN = 'boolean';
     const TYPE_MONEY = 'money';
+    //Postgresql types
```
klimov-paul: Such addition unacceptable: `Schema` can provide only general types.

framework/db/SchemaBuilderTrait.php
```diff
@@ -258,4 +258,355 @@ public function money($precision = null, $scale = null)
         }
         return $this->getDb()->getSchema()->createColumnSchemaBuilder(Schema::TYPE_MONEY, $length);
     }
+
```
klimov-paul: Same as above.

cebe:
> Add types and functions for work with postgresql DBMS.

The purpose of the columnschema builder is to create dbms independent schema. If you want to use postgres only, you can just define the column type as string using postgres types directly.

Also, we have subclasses for each DBMS for the querybuilder so dbms specific things should go there.

I'm sorry but we can not take this PR. I am open to discuss parts of this code to be applied in other places but I'll close this PR for now.

PR #10708, opened 2016-01-29
fix dropdown list filtering in gridview, when gridview inserted with …

vologab:
When GridView ([/some_grid_view]  route) (surrounded by Pjax widget) inserted in page (for example in Tab widget [/tab_with_gridview route]), then, when using dropdown filter sending request without pjax, and page redirecting to [ /some_grid_view route].

It's happen because, Pjax generate listeners for element $('#some_grid_id form[data-pjax]'), but yii.gridView.js create form in "root" DOM, so event listener not catch form.submit event.


samdark:
Won't accept this pull request because it will catch things that are outside PJAX block which isn't desired. You can achieve the effect you need by setting `formSelector` when calling the widget.

PR #10711, opened 2016-01-29
Fixes yiisoft/yii2#10710

wallysalami:
Fixes yiisoft/yii2#10710

### DIFF
--- a/framework/helpers/BaseFileHelper.php
+++ b/framework/helpers/BaseFileHelper.php
@@ -253,6 +253,9 @@ class BaseFileHelper
      */
     public static function copyDirectory($src, $dst, $options = [])
     {
+        if ($src === $dst || strpos($dst, $src) !== false){
+            throw new InvalidParamException( "Trying to copy a directory to itself or a subdirectory." );
+        }
         if (!is_dir($dst)) {
             static::createDirectory($dst, isset($options['dirMode']) ? $options['dirMode'] : 0775, true);
         }


framework/helpers/BaseFileHelper.php
```diff
@@ -253,6 +253,9 @@ protected static function loadMimeTypes($magicFile)
      */
     public static function copyDirectory($src, $dst, $options = [])
     {
+        if ($src === $dst || strpos($dst, $src) !== false){
```
cebe: strpos should only match if its result is 0, e.g. copying from `/data` to `/backup/data` should work.
Would be good to also add tests to https://github.com/yiisoft/yii2/blob/master/tests/framework/helpers/FileHelperTest.php

SilverFire: `){` -> `) {`

framework/helpers/BaseFileHelper.php
```diff
@@ -253,6 +253,9 @@ protected static function loadMimeTypes($magicFile)
...
+            throw new InvalidParamException( "Trying to copy a directory to itself or a subdirectory." );
```
SilverFire: Could you please remove redundant spaces before after `(` and before `)`

SilverFire:
Merged https://github.com/yiisoft/yii2/commit/4f4fa6fdf0a45921ef48ffcd2941352c33a8dd28

PR #10715, opened 2016-01-30
Fixed a typo

drodata:


SilverFire:
Thank you!

klimov-paul:
Original text was correct: 
> that takes the input from

meaning taking the input **from** model

SilverFire:
@klimov-paul yes, you're right. Reverted the commit: https://github.com/yiisoft/yii2/commit/9691d3dcc117e37085cfb5e7701a334f04b9680c

samdark:
I think the sentence is still incorrect.

samdark:
I think it should be:

```
For filtering data the GridView needs a [model](structure-models.md) to take the input from and the filtering form that adjusts the query of the dataProvider to respect the search criteria.
```

klimov-paul:
Sounds good

PR #10716, opened 2016-01-30
Allow extract attribute value from path property

leandrogehlen:
Example:

``` php
/* @var $model app\models\Person */

<?php $form = ActiveForm::begin() ?>
    <?= $form->field($model, 'city.name')->textInput() ?>
<?php ActiveForm::end() ?>
```


cebe:
This does not work. How do you validate the input of the form? You need a model for the relation and pass that one to `field()`

PR #10726, opened 2016-02-01
Additional functions for build user friendly role management interface

dkhlystov:


samdark:
@dkhlystov could you add these methods to interface and implement PHP manager as well as tests?

dkhlystov:
When testing Php Manager a bug was found with determine users by role. Php automaticaly converts string array key to integer when I'm assigning role to user. So === checking is false and getUserIdsByRole returns wrong result.

dkhlystov:
Is this realy bug?

dkhlystov:
```
$role = $auth->getRole('page');
$auth->assign($role, 1);
var_dump($auth->getUserIdsByRole('page')); //array(1) { [0]=> string(1) "1" }
$auth->revoke($role, 1);
$auth->assign($role, '1');
var_dump($auth->getUserIdsByRole('page')); //array(0) { }
$auth->revoke($role, 1);

```

dkhlystov:
It sems that is bug after some research how php working with arrays.

framework/rbac/ManagerInterface.php
```diff
@@ -129,6 +129,14 @@ public function getRule($name);
     public function getRules();
 
     /**
+     * Checks the possibility of adding a child to parent
+     * @param Item $parent the parent item
+     * @param Item $child the child item to be added to the hierarchy
+     * @return boolean possibility of adding
+     */
+    public function canAddChild($parent, $child);
```
cebe: @samdark Isn't adding methods to interfaces a BC break?

samdark: It is but in this case I'd make an exception.

samdark:
Thanks!

PR #10733, opened 2016-02-02
Redirect 301 for unresolved url matching with suffix

### LINKED ISSUE #7670: UrlRule Problem with trailing slash (/)

dynasource:
fixes #7670

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -7,6 +7,7 @@ Yii Framework 2 Change Log
 - Bug #6363, #8301, #8582, #9566: Fixed data methods and PJAX issues when used together (derekisbusy)
 - Bug #6876: Fixed RBAC migration MSSQL cascade problem (thejahweh)
 - Bug #7627: Fixed `yii\widgets\ActiveField` to handle inputs validation with changed ID properly (dynasource, cebe)
+- Bug #7670: Fixed `yii\web\Request`: forces redirect when unresolved url matches an url with suffix (dynasource)
 - Bug #7806: Fixed `yii\grid\CheckboxColumn` fixed `_all` checkbox column name generation (cebe, silverfire)
 - Bug #8466: Fixed `yii\validators\FileValidator` to display error for `tooBig` and `tooSmall` with formatted unit (silverfire)
 - Bug #8573: Fixed incorrect data type mapping for NUMBER to INTEGER or DECIMAL in Oracle (vbelogai)
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -181,7 +181,7 @@ class Request extends \yii\base\Request
         if($url = $this->resolveRequest($this,$currentQueryParams)){
             return $url;
         }else if(isset(Yii::$app->getUrlManager()->suffix)){
-            $request = $this;
+            $request = clone $this;
             $request->setPathInfo($request->getPathInfo() . Yii::$app->getUrlManager()->suffix);                
             if($url = $this->resolveRequest($request,$currentQueryParams)){
                 return Yii::$app->response->redirect($url, 301);
@@ -197,7 +197,7 @@ class Request extends \yii\base\Request
      * @return array the first element is the route, and the second is the associated parameters.
      * @return false if the request cannot be resolved.
      */
-    public function resolveRequest($request,$currentQueryParams)
+    private function resolveRequest($request,$currentQueryParams)
     {
         $result = Yii::$app->getUrlManager()->parseRequest($request);       
         if ($result !== false) {
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -169,26 +169,47 @@ class Request extends \yii\base\Request
      */
     private $_headers;
 
-
     /**
      * Resolves the current request into a route and the associated parameters.
      * @return array the first element is the route, and the second is the associated parameters.
+     * @redirect when an unresolved url matches an url with suffix 
      * @throws NotFoundHttpException if the request cannot be resolved.
      */
     public function resolve()
     {
-        $result = Yii::$app->getUrlManager()->parseRequest($this);
+        $currentQueryParams = $this->_queryParams;
+        if($url = $this->resolveRequest($this,$currentQueryParams)){
+            return $url;
+        }else if(isset(Yii::$app->getUrlManager()->suffix)){
+            $request = $this;
+            $request->setPathInfo($request->getPathInfo() . Yii::$app->getUrlManager()->suffix);                
+            if($url = $this->resolveRequest($request,$currentQueryParams)){
+                return Yii::$app->response->redirect($url, 301);
+            }              
+        }        
+        throw new NotFoundHttpException(Yii::t('yii', 'Page not found.'));
+    }
+    
+    /**
+     * Resolves a given $request at its $queryParams into a route and the associated parameters.
+     * @param Request $request, current or custom request
+     * @param array $currentQueryParams of current request
+     * @return array the first element is the route, and the second is the associated parameters.
+     * @return false if the request cannot be resolved.
+     */
+    public function resolveRequest($request,$currentQueryParams)
+    {
+        $result = Yii::$app->getUrlManager()->parseRequest($request);       
         if ($result !== false) {
-            list ($route, $params) = $result;
-            if ($this->_queryParams === null) {
+            list ($route, $params) = $result;            
+            if ($currentQueryParams === null) {
                 $_GET = $params + $_GET; // preserve numeric keys
             } else {
-                $this->_queryParams = $params + $this->_queryParams;
+                $request->setQueryParams($params + $currentQueryParams);
             }
             return [$route, $this->getQueryParams()];
-        } else {
-            throw new NotFoundHttpException(Yii::t('yii', 'Page not found.'));
         }
+        return false;        
     }
 
     /**
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -172,44 +172,43 @@ class Request extends \yii\base\Request
     /**
      * Resolves the current request into a route and the associated parameters.
      * @return array the first element is the route, and the second is the associated parameters.
-     * @redirect when an unresolved url matches an url with suffix 
+     * @return \yii\web\Response when an unresolved url matches an url with suffix.
      * @throws NotFoundHttpException if the request cannot be resolved.
      */
-    public function resolve()
+    public function resolve() 
     {
         $currentQueryParams = $this->_queryParams;
-        if($url = $this->resolveRequest($this,$currentQueryParams)){
+        if ($url = $this->resolveRequest($currentQueryParams)) {
             return $url;
-        }else if(isset(Yii::$app->getUrlManager()->suffix)){
+        } else if (isset(Yii::$app->getUrlManager()->suffix)) {
             $request = clone $this;
-            $request->setPathInfo($request->getPathInfo() . Yii::$app->getUrlManager()->suffix);                
-            if($url = $this->resolveRequest($request,$currentQueryParams)){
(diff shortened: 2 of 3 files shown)


framework/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Yii Framework 2 Change Log
 - Bug #6363, #8301, #8582, #9566: Fixed data methods and PJAX issues when used together (derekisbusy)
 - Bug #6876: Fixed RBAC migration MSSQL cascade problem (thejahweh)
 - Bug #7627: Fixed `yii\widgets\ActiveField` to handle inputs validation with changed ID properly (dynasource, cebe)
+- Bug #7670: Fixed `yii\web\Request`: forces redirect when unresolved url matches an url with suffix (dynasource)
```
samdark: It should work the other way around as well.

framework/web/Request.php
```diff
@@ -169,26 +169,47 @@ class Request extends \yii\base\Request
...
+     * @return array the first element is the route, and the second is the associated parameters.
+     * @return false if the request cannot be resolved.
+     */
+    private function resolveRequest($request,$currentQueryParams)
+    {
+        $result = Yii::$app->getUrlManager()->parseRequest($request);       
         if ($result !== false) {
-            list ($route, $params) = $result;
-            if ($this->_queryParams === null) {
+            list ($route, $params) = $result;
```
SilverFire: Redundant spaces at the end of the line

framework/web/Request.php
```diff
@@ -169,26 +169,47 @@ class Request extends \yii\base\Request
      */
     private $_headers;
 
-
     /**
      * Resolves the current request into a route and the associated parameters.
      * @return array the first element is the route, and the second is the associated parameters.
+     * @redirect when an unresolved url matches an url with suffix
```
SilverFire: `@redirect` is not a valid PHPDoc tag. Move this line above the `@return`, please.

framework/web/Request.php
```diff
@@ -169,26 +169,47 @@ class Request extends \yii\base\Request
...
      * Resolves the current request into a route and the associated parameters.
      * @return array the first element is the route, and the second is the associated parameters.
+     * @redirect when an unresolved url matches an url with suffix 
      * @throws NotFoundHttpException if the request cannot be resolved.
      */
     public function resolve()
     {
-        $result = Yii::$app->getUrlManager()->parseRequest($this);
+        $currentQueryParams = $this->_queryParams;
+        if($url = $this->resolveRequest($this,$currentQueryParams)){
```
SilverFire: Here and bellow the code style must be fixed according to the PSR

`if(` -> `if (`
`){` -> `) {`

framework/web/Request.php
```diff
@@ -169,26 +169,47 @@ class Request extends \yii\base\Request
...
+            return $url;
+        }else if(isset(Yii::$app->getUrlManager()->suffix)){
+            $request = clone $this;
```
klimov-paul: Odd.
It does not feel right to clone `Request` instance just to check if there is a redundant/missing suffix.

framework/web/Request.php
```diff
@@ -169,26 +169,47 @@ class Request extends \yii\base\Request
...
+    }
+    
+    /**
+     * Resolves a given $request at its $queryParams into a route and the associated parameters.
+     * @param Request $request, current or custom request
+     * @param array $currentQueryParams of current request
+     * @return array the first element is the route, and the second is the associated parameters.
+     * @return false if the request cannot be resolved.
+     */
+    private function resolveRequest($request,$currentQueryParams)
```
klimov-paul: Why passing `Request` instance as parameter to non static `Request` method?

framework/web/Request.php
```diff
@@ -169,26 +169,47 @@ class Request extends \yii\base\Request
...
     public function resolve()
     {
-        $result = Yii::$app->getUrlManager()->parseRequest($this);
+        $currentQueryParams = $this->_queryParams;
+        if($url = $this->resolveRequest($this,$currentQueryParams)){
+            return $url;
+        }else if(isset(Yii::$app->getUrlManager()->suffix)){
+            $request = clone $this;
+            $request->setPathInfo($request->getPathInfo() . Yii::$app->getUrlManager()->suffix);                
+            if($url = $this->resolveRequest($request,$currentQueryParams)){
```
klimov-paul: This will produce significant performance impact, since URL rules will be parsed twice per each request processing.
Thus `preg_match()` calls will be doubled.

klimov-paul:
Solution does not seem right to me. There must be a simplier way to achieve the goal.

framework/web/Request.php
dynasource: thats not correct. It will only be parsed twice if the request is NOT resolved in the first processing.

framework/web/Request.php
dynasource: this method is to:
- at first process the current Request 
- if not resolved, reprocess a modified Request with an altered URL (with suffix)

klimov-paul: Why you can not invoke it twice like following:
```php
$this->resolveRequest();
$request->resolveRequest();
```
???

framework/web/Request.php
dynasource: well, it has been done to keep the current Request clean. When the 2nd resolve does not succeed, an error will still be thrown with the original Request (url)

dynasource:
with the current state of the UrlManager + UrlRules + Request, this is a simple way of achieving this goal.

When you desire a cleaner and even more simple way, than more refactoring and cleanups have to be done. I would say, this is something for later releases.

framework/web/Request.php
dynasource: originally to create a processing method, but now I would prefer your suggestion

SilverFire:
Nothing is so permanent as a temporary solution.

framework/CHANGELOG.md
dynasource: there are 2 problems with this other way around:

1) The current UrlManager + UrlRules (2.0.6) accept infinit '/' suffixes in the url (try it yourself with a route like /index.php/test/route//////). This means that a suffix like '/' is always accepted in the URL, despite a suffix not being set (perhabs this is another issue for the tracker)

2) Taken this into account this means that, if you still want to check the other way around, you have to check on a '/' suffix for every request that is correctly parsed (because an given url with '////' are always parsed correctly). This goes at cost of performance.

dynasource:
nothing is so permanent as a perminant 301 redirect ;)

framework/CHANGELOG.md
samdark: 1. rtrim `/`.
2. URLs with `////` parsed correctly? What do you mean?

samdark:
@klimov-paul do you have an idea for better approach to the problem?

df2:
Just to throw it out there, we have things setup like below. This way, the default is no trailing slash but it is optional.

```php
<?php
namespace yii\helpers;

use Yii;
use yii\helpers\BaseUrl;

class Url extends BaseUrl {

    public static function to($url = '', $scheme = false)
    {
        if (is_array($url)) {

            if (!isset($url['tbs']) && $url[0] !== 'home/index') {

                $url['tbs'] = '';
            }

            return static::toRoute($url, $scheme);
        }
```

and at the end of urls.php:

```php
$newUrls = [];

foreach ($urls as $url => $route) {

    if ($url !== '/') {

        $url .= '<tbs:\/?>';
    }

    $newUrls[$url] = $route;
}

return $newUrls;
```

framework/CHANGELOG.md
dynasource: Dived into it. 
I started with a fresh new Advanced project with the following setting:

        'urlManager'=>[
            'class'=>'yii\web\UrlManager',
            'enablePrettyUrl' => true,            
        ]

Both following urls render correctly:
- http://localhost/dmn/yii-application/frontend/web/index.php/site/about
- http://localhost/dmn/yii-application/frontend/web/index.php/site/about/////

Seems like unwanted behavior.

dynasource:
@df2, this patch is about automatic redirection

klimov-paul:
> @klimov-paul do you have an idea for better approach to the problem?

This is hard to tell for sure. But at first glance just analyzing of `$_SERVER['REQUEST_URI']` and `$_SERVER['QUERY_STRING']` should be enough to find out if suffix is present or not.

samdark:
Correct but we need to get the suffix we're checking for from somewhere. It could be defined either in UrlManager or in UrlRule.

klimov-paul:
We may anvance `UrlManaer::parseRequest()` so it can return URL suffix as a third element of the result array.
This line:
https://github.com/yiisoft/yii2/blob/master/framework/web/UrlManager.php#L268

can be following:
```php
return [(string) $route, [], $urlSuffix];
```

This will not cause a BC break.

klimov-paul:
Inside `Request::resolve()`, we parse 3 elements instead of 2 and so something if `$ulrSuffix` is present:

```php
public function resolve()
    {
        $result = Yii::$app->getUrlManager()->parseRequest($this);
        if ($result !== false) {
            list ($route, $params, $urlSuffix) = $result;
            
            if (isset($urlSuffix)) {
                 // check and redirect if necessary
            }

            //...
    }
```

dynasource:
well, it looks like your excluding the (rare) usecase of https://github.com/yiisoft/yii2/pull/9207#discussion_r36042232

Based on your example:
- you are matching by route & params first
- to get a 100% match (whether its with or without suffix)
- then you do decide to redirect

If you do this, than someone cannot have both Urls as UrlRules:
- http://exampe.com/contact/ -> pageX
- http://exampe.com/contact -> pageY

Because pageX is matched first, pageY is ignored

klimov-paul:
> it looks like your excluding the (rare) usecase

Why? Url suffix is used for URL rule parsing already. Does not it?
Parsing `http://exampe.com/contact/` will return result:
```php
[
    'site/contact',
    [],
    '/'
]
```
And everything will be fine.

Parsing `http://exampe.com/contact` will return
```php
[
    'site/contact',
    [],
    ''
]
```
Also should be fine.

Then we compare returned 3rd parameter with the request URI, checking if it is present or not.

dynasource:
the issue is about hard & softmatching and the correct order which means, first hard matching, then softmatching.

If you are doing a redirect, than it means, the URL is not right by hard matching, but with softmatching on a suffix, it should be executed.

If you want are using the current implementation of UrlManager::parseRequest, you are looping on the UrlRules. You are returning [(string) $route, [], $urlSuffix] if a match of the UrlRule has been found.

With your example, this means a softmatching on suffix is done in an early stage in the UrlRule itself, otherwise it could never return the array with the correct suffix.

This also mean that a hard match on a following UrlRule is ignored.

klimov-paul:
> If you are doing a redirect, than it means, the URL is not right by hard matching, but with softmatching on a suffix, it should be executed.

Are you sure about that?
UrlRule matchign includes the suffix:
https://github.com/yiisoft/yii2/blob/master/framework/web/UrlRule.php#L238

If suffix is missmatch the Rule will return `false` instead of array and thus will be skipped.

The goal is enforce some trainling suffix ("/") in case NO suffix is explicitly macthed by `Rule` and it is present in UrlManager. Thus is should be fine.

The only problem I can see if you want enforce some suffix like "/", but still keep a route without any suffix. For this use case some magic value should be reserved like `false`, indicating `Rule` should not have a suffix.

klimov-paul:
I have no garantee my idea will work. But at least it worth trying.

dynasource:
true. Your solution is cleaner & faster. If you get it more robust, fixing the problem you mentioned, I would prefer yours.

klimov-paul:
See #10745

PR #10745, opened 2016-02-03
Redirect 301 on missing URL suffix

### LINKED ISSUE #7670: UrlRule Problem with trailing slash (/)

klimov-paul:
Fixes #7670 :`yii\web\Request`: forces redirect when unresolved URL matches an URL with suffix

### DIFF
--- a/framework/web/Application.php
+++ b/framework/web/Application.php
@@ -74,8 +74,8 @@ class Application extends \yii\base\Application
         if (empty($this->catchAll)) {
             try {
                 list ($route, $params) = $request->resolve();
-            } catch (RedirectException $exception) {
-                return $this->getResponse()->redirect($exception->url, $exception->statusCode);
+            } catch (RedirectException $e) {
+                return $this->getResponse()->redirect($e->url, $e->statusCode);
             }
         } else {
             $route = $this->catchAll[0];
@@ -96,8 +96,8 @@ class Application extends \yii\base\Application
 
                 return $response;
             }
-        } catch (InvalidRouteException $exception) {
-            throw new NotFoundHttpException(Yii::t('yii', 'Page not found.'), $exception->getCode(), $exception);
+        } catch (InvalidRouteException $e) {
+            throw new NotFoundHttpException(Yii::t('yii', 'Page not found.'), $e->getCode(), $e);
         }
     }
 
--- a/framework/web/Application.php
+++ b/framework/web/Application.php
@@ -72,7 +72,11 @@ class Application extends \yii\base\Application
     public function handleRequest($request)
     {
         if (empty($this->catchAll)) {
-            list ($route, $params) = $request->resolve();
+            try {
+                list ($route, $params) = $request->resolve();
+            } catch (RedirectException $exception) {
+                return $this->getResponse()->redirect($exception->url, $exception->statusCode);
+            }
         } else {
             $route = $this->catchAll[0];
             $params = $this->catchAll;
@@ -92,8 +96,8 @@ class Application extends \yii\base\Application
 
                 return $response;
             }
-        } catch (InvalidRouteException $e) {
-            throw new NotFoundHttpException(Yii::t('yii', 'Page not found.'), $e->getCode(), $e);
+        } catch (InvalidRouteException $exception) {
+            throw new NotFoundHttpException(Yii::t('yii', 'Page not found.'), $exception->getCode(), $exception);
         }
     }
 
--- a/framework/web/RedirectException.php
+++ b/framework/web/RedirectException.php
@@ -13,7 +13,7 @@ use yii\base\Exception;
  * RedirectException represents an information for the HTTP location redirect.
  *
  * @author Paul Klimov <klimov.paul@gmail.com>
- * @since 2.0
+ * @since 2.0.7
  */
 class RedirectException extends Exception
 {
--- a/framework/web/RedirectException.php
+++ b/framework/web/RedirectException.php
@@ -1,0 +1,43 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\web;
+
+use yii\base\Exception;
+
+/**
+ * RedirectException represents an information for the HTTP location redirect.
+ *
+ * @author Paul Klimov <klimov.paul@gmail.com>
+ * @since 2.0
+ */
+class RedirectException extends Exception
+{
+    /**
+     * @var string  the URL to be redirected to.
+     */
+    public $url;
+    /**
+     * @var integer the HTTP status code.
+     */
+    public $statusCode;
+
+    /**
+     * Constructor.
+     * @param string $url the URL to be redirected to
+     * @param integer $statusCode the HTTP status code. Defaults to 302.
+     * @param string $message error message
+     * @param integer $code error code
+     * @param \Exception $previous The previous exception used for the exception chaining.
+     */
+    public function __construct($url, $statusCode = 302, $message = null, $code = 0, \Exception $previous = null)
+    {
+        $this->url = $url;
+        $this->statusCode = $statusCode;
+        parent::__construct($message, $code, $previous);
+    }
+}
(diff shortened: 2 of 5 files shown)


framework/web/Application.php
```diff
@@ -92,8 +96,8 @@ public function handleRequest($request)
 
                 return $response;
             }
-        } catch (InvalidRouteException $e) {
-            throw new NotFoundHttpException(Yii::t('yii', 'Page not found.'), $e->getCode(), $e);
+        } catch (InvalidRouteException $exception) {
```
samdark: Why renaming it? I think it's a common practice to use `$e` for `catch`.

framework/web/RedirectException.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Constructor.
+     * @param string $url the URL to be redirected to
+     * @param integer $statusCode the HTTP status code. Defaults to 302.
+     * @param string $message error message
+     * @param integer $code error code
+     * @param \Exception $previous The previous exception used for the exception chaining.
+     */
+    public function __construct($url, $statusCode = 302, $message = null, $code = 0, \Exception $previous = null)
```
samdark: Should be 301, not 302.

samdark:
@dynasource let's test it.

dynasource:
this is a different version than we've discussed about? I will test it locally

framework/web/Request.php
```diff
@@ -187,6 +188,24 @@ public function resolve()
...
         } else {
+            if (Yii::$app->getUrlManager()->enablePrettyUrl && !empty(Yii::$app->getUrlManager()->suffix)) {
+                $pathInfo = $this->getPathInfo();
+                $suffix = Yii::$app->getUrlManager()->suffix;
+                $n = strlen($suffix);
+                if (substr_compare($pathInfo, $suffix, -$n, $n) !== 0) {
+                    $this->setPathInfo($pathInfo . $suffix);
+                    $result = Yii::$app->getUrlManager()->parseRequest($this);
+                    if ($result !== false) {
+                        $url = $this->getBaseUrl() . '/' . $this->getPathInfo();
```
dynasource: this should be getScriptUrl instead of baseUrl:
$url = $this->getScriptUrl() . '/' . $this->getPathInfo();

Otherwise:
- http://localhost/dmn/yii-application/frontend/web/index.php/site/about will redirect to:
- http://localhost/dmn/yii-application/frontend/web/site/about

framework/web/RedirectException.php
```diff
@@ -0,0 +1,43 @@
...
+
+namespace yii\web;
+
+use yii\base\Exception;
+
+/**
+ * RedirectException represents an information for the HTTP location redirect.
+ *
+ * @author Paul Klimov <klimov.paul@gmail.com>
+ * @since 2.0
```
SilverFire: `2.0.7`?

tests/framework/web/RequestTest.php
```diff
@@ -146,4 +147,35 @@ public function testResolve()
...
+        $request->pathInfo = 'posts/list';
+        $request->baseUrl = '';
+
+        try {
+            $request->resolve();
+        } catch (RedirectException $exception) {
+            // catch redirect exception
+        }
+        $this->assertTrue(isset($exception));
+        $this->assertEquals('/posts/list/', $exception->url);
```
dynasource: Change to include scriptUrl:
        $this->assertEquals($request->scriptUrl.'/posts/list/', $exception->url);

tests/framework/web/RequestTest.php
```diff
@@ -146,4 +147,35 @@ public function testResolve()
...
         $this->assertEquals($_GET, ['id' => 63]);
     }
+
+    public function testResolveSuffixRedirect()
+    {
+        $this->mockWebApplication([
+            'components' => [
+                'urlManager' => [
+                    'enablePrettyUrl' => true,
+                    'showScriptName' => false,
```
dynasource: remove $showScriptName => false and keep it on true (which is default)

tests/framework/web/RequestTest.php
```diff
@@ -146,4 +147,35 @@ public function testResolve()
...
+                        'posts' => 'post/list',
+                    ],
+                ]
+            ]
+        ]);
+
+        $request = new Request();
+        $request->hostInfo = 'http://example.com/';
+        $request->pathInfo = 'posts/list';
+        $request->baseUrl = '';
```
dynasource: add:
        $request->scriptUrl = '/path/index.php';

framework/web/Request.php
dynasource: See comments beneath for the updated test, to show this issue.

framework/web/Request.php
```diff
@@ -187,6 +188,24 @@ public function resolve()
             }
             return [$route, $this->getQueryParams()];
         } else {
+            if (Yii::$app->getUrlManager()->enablePrettyUrl && !empty(Yii::$app->getUrlManager()->suffix)) {
+                $pathInfo = $this->getPathInfo();
+                $suffix = Yii::$app->getUrlManager()->suffix;
+                $n = strlen($suffix);
+                if (substr_compare($pathInfo, $suffix, -$n, $n) !== 0) {
+                    $this->setPathInfo($pathInfo . $suffix);
```
dynasource: you are changing the current Path to include the $suffix. When it still is not resolving, the NotFoundHttpException will have an updated Request and not the original. This would be confusing.

tests/framework/web/RequestTest.php
samdark: I think we need to test both.

framework/web/Request.php
```diff
@@ -187,6 +188,24 @@ public function resolve()
             }
             return [$route, $this->getQueryParams()];
         } else {
+            if (Yii::$app->getUrlManager()->enablePrettyUrl && !empty(Yii::$app->getUrlManager()->suffix)) {
+                $pathInfo = $this->getPathInfo();
+                $suffix = Yii::$app->getUrlManager()->suffix;
```
dynasource: I see now that both of our solutions, miss out the opportunity that an Url rule can have its own suffix, which is ignored now.

tests/framework/web/RequestTest.php
dynasource: Overal set of tests to add
1) test redirection of:
- example.com/path/controller/action ->  /path/controller/action/
- example.com/path/controller/action ->  /path/controller/action/
- example.com/path/index.php/controller/action ->  /path/index.php/controller/action/
- example.com/path/index.php/module/controller/action ->  /path/index.php/module/controller/action/

2) the other way around
- example.com/path/controller/action/ ->  /path/controller/action
- etc

3) UrlRules
- a hard rules setup (like accepting both /contact and /contact/ )
- a UrlRule suffix scenario

tests/framework/web/RequestTest.php
dynasource: I agree, the comment was to show the issue when $showScriptName is at true.

framework/web/Application.php
klimov-paul: reverted

framework/web/RedirectException.php
klimov-paul: Sure, fixed

framework/web/Request.php
klimov-paul: fixed

tests/framework/web/RequestTest.php
klimov-paul: test added

framework/web/Request.php
klimov-paul: fixed

samdark:
@klimov-paul what's left on this one?

klimov-paul:
Need to think about it a little more. I'll move it to 2.0.8 milestone.

samdark:
OK.

samdark:
@klimov-paul any good ideas since last time?

klimov-paul:
The basic idea is extraction of separated method inside `UrlManager`, which will allow parsing URLs without suffix being involved.
E.g. convert current `parseUrl()` to something like `rawParseUrl`, which accepts additional parameter indicating whether suffix should be parsed or not. Then `parseUrl()` will at first invoke this new method with suffix parsing enabled and if it fails invoke it second time without suffix parsing. If second run success, we can perform the 301 redirect.

rob006:
see #12240

samdark:
Closed in favor of #12240

Issue #3581, opened 2014-05-26
Enhance CompareValidator for correct comparison based on data type


kartik-v:
Currently the `CompareValidator` assumes the comparison type is mostly numeric for comparison operators like (> and <). This may not work for all data types. Especially DATE strings misbehave (e.g. consider 2 attributes for comparison with a format like `24-Mar-2014` and `23-Apr-2014`, I am sure the latter will wrongly be treated lesser than the former.

Could we add a data type parameter e.g. `date`, `char`, or `number`. 

Probably take in an additional `format` string if needed and set default value if not set from the `Yii:$app->formatter`.  Now with all of these, we can do the following based on `dataType` parameter:
- for `date` (includes `time` as well): convert the string to Unix timestamp by reading the `format` and then do the comparison
- for `number`: convert the string to FLOAT and do the comparison.
- for `char`: do the comparison as is without any conversion.
- add Closure method to return value for complex comparisons

I did go through a related PR #3368 (but it handles and converts everything to float - which will not work for `date` strings).

cebe:
number and string are supported now, date is more complex and should be done in combination with date validator, see http://www.yiiframework.com/doc-2.0/guide-tutorial-core-validators.html#comparing-date-values

PR #10748, opened 2016-02-04
Fix compare validator

hustnaive:
as issue #8884 said, when I use `CompareValidator` to compare numberic string, I would get the wrong result sometimes. As if I compare the input string '00' and '000', and use compare type 'string', then the result would be true, but I don't think the string made of two zeroes is equal to that of three.

so, I think it is a bug, when the compare type is string, we must use `strcmp` instead of `==` .


cebe:
we should ensure that the client validation implementation has the same results as the server side validation.

dynasource:
closed in favor of a multi usecase solution for https://github.com/yiisoft/yii2/issues/3581

PR #10751, opened 2016-02-04
Update UrlValidator.php

hrom252:
Added sign of string's end in regular expression for the correctly validating urls with banned characters like ',' in "http://domain.com,".


klimov-paul:
Can not accept this as with this change any URL with query string like `?param=value` will be rejected.

hrom252:
Ok, maybe later I will try to fix it only between zone and params start. Because now the string "http://domain.com,?param=value" will validate as correct.

cebe:
@hrom252 should be fixed by #10753 now.

PR #10753, opened 2016-02-04
Fixed UrlValidator pattern to improve matching

SilverFire:
Instead of #10751
- [x] Tests
- [x] Review

### DIFF
--- a/framework/validators/UrlValidator.php
+++ b/framework/validators/UrlValidator.php
@@ -9,6 +9,7 @@ namespace yii\validators;
 
 use Yii;
 use yii\base\InvalidConfigException;
+use yii\helpers\Html;
 use yii\web\JsExpression;
 use yii\helpers\Json;
 
@@ -28,7 +29,7 @@ class UrlValidator extends Validator
      * The pattern may contain a `{schemes}` token that will be replaced
      * by a regular expression which represents the [[validSchemes]].
      */
-    public $pattern = '/^{schemes}:\/\/(([A-Z0-9][A-Z0-9_-]*)(\.[A-Z0-9][A-Z0-9_-]*)+)/i';
+    public $pattern = '/^{schemes}:\/\/(([A-Z0-9][A-Z0-9_-]*)(\.[A-Z0-9][A-Z0-9_-]*)+)(?::\d{1,5})?(?:$|[?\/#])/i';
     /**
      * @var array list of URI schemes which should be considered valid. By default, http and https
      * are considered to be valid schemes.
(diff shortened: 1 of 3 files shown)


tests/framework/validators/UrlValidatorTest.php
```diff
@@ -28,6 +28,14 @@ public function testValidateValue()
...
         $this->assertFalse($val->validate('ftp://ftp.ruhr-uni-bochum.de/'));
         $this->assertFalse($val->validate('http://invalid,domain'));
+        $this->assertFalse($val->validate('http://example.com,'));
+        $this->assertFalse($val->validate('http://example.com*12'));
+        $this->assertTrue($val->validate('http://example.com/*12'));
+        $this->assertTrue($val->validate('http://example.com#test'));
+        $this->assertTrue($val->validate('http://example.com/?test'));
+        $this->assertTrue($val->validate('http://example.com?test'));
+        $this->assertTrue($val->validate('http://example.com/#test'));
+        $this->assertFalse($val->validate('http://example.com,?test'));
```
klimov-paul: At least following tests are missing:
```
http://example.com/?param1=test1&param2=test2
http://example.com/?param1=FormName[name1]=value1&FormName[name2]=value2
```

framework/validators/UrlValidator.php
```diff
@@ -28,7 +29,7 @@ class UrlValidator extends Validator
      * The pattern may contain a `{schemes}` token that will be replaced
      * by a regular expression which represents the [[validSchemes]].
      */
-    public $pattern = '/^{schemes}:\/\/(([A-Z0-9][A-Z0-9_-]*)(\.[A-Z0-9][A-Z0-9_-]*)+)/i';
+    public $pattern = '/^{schemes}:\/\/(([A-Z0-9][A-Z0-9_-]*)(\.[A-Z0-9][A-Z0-9_-]*)+)($|[?\/#])/i';
```
klimov-paul: It seems ampersand (`&`) is not covered

SilverFire: `&` cat not be at the beginning of the parameter part of URL

`http://example.com&param=1` is not a valid URL

tests/framework/validators/UrlValidatorTest.php
SilverFire: These test are redundant because only the first char after the domain name has limitations

framework/validators/UrlValidator.php
klimov-paul: OK. Fine then.

cebe:
:+1:

Issue #10739, opened 2016-02-03
PrettyUrl accepts unlimited slashes allowing /route//////


dynasource:
based on https://github.com/yiisoft/yii2/pull/10733#discussion_r51703531

It seems that slashes can be appended in the URL, without issues.
This means that the following urls render correctly:

http://localhost/dmn/yii-application/frontend/web/index.php/site/about
http://localhost/dmn/yii-application/frontend/web/index.php/site/about/////
or
http://localhost/dmn/yii-application/frontend/web/index.php/site/about/////////////////

All returning the same page.

This can be reproduced by installing i.e. the Advanced App with setting:
```php
   'urlManager'=>[
        'class'=>'yii\web\UrlManager',
        'enablePrettyUrl' => true,            
    ]
```

andrewnester:
Verified. It accepts multiple slashes at the end of URL

andrewnester:
Just added PR that fixes issue - if we have multiple slashes at the end of URL it should be handled as invalid URL

SilverFire:
https://github.com/yiisoft/yii2/commit/e8a7a8a0f4e86e3a374cf273fc6cbbbbfd214dfe

PR #10757, opened 2016-02-05
10739 - Multiple slashes at the end of route should not be accepted as valid url

andrewnester:
Fix for #10739

### DIFF
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -242,6 +242,11 @@ class UrlManager extends Component
 
             Yii::trace('No matching URL rules. Using default URL parsing logic.', __METHOD__);
 
+            // Ensure, that $pathInfo does not end with more than one slash.
+            if (strlen($pathInfo) > 1 && substr_compare($pathInfo, '//', -2, 2) === 0) {
+                return false;
+            }
+
             $suffix = (string) $this->suffix;
             if ($suffix !== '' && $pathInfo !== '') {
                 $n = strlen($this->suffix);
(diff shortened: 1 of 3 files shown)


framework/web/UrlManager.php
```diff
@@ -242,6 +242,12 @@ public function parseRequest($request)
 
             Yii::trace('No matching URL rules. Using default URL parsing logic.', __METHOD__);
 
+            // before processing pathInfo and suffix there should not be end multiple slashes in $pathInfo
+            $endSlashPattern = '/\/\/$/';
+            if (preg_match($endSlashPattern, $pathInfo)) {
```
SilverFire: `substr_compare` should be faster

dynasource: agreed. Wherever possible, preg_ functions should be avoided

andrewnester: @dynasource @SilverFire fixed by using substr_compare instead

framework/web/UrlManager.php
```diff
@@ -242,6 +242,12 @@ public function parseRequest($request)
 
             Yii::trace('No matching URL rules. Using default URL parsing logic.', __METHOD__);
 
+            // before processing pathInfo and suffix there should not be end multiple slashes in $pathInfo
+            $endSlashPattern = '//';
```
SilverFire: Why do we need a separate variable for this? Comment above is enough to describe, what's going here

framework/web/UrlManager.php
```diff
@@ -242,6 +242,12 @@ public function parseRequest($request)
 
             Yii::trace('No matching URL rules. Using default URL parsing logic.', __METHOD__);
 
+            // before processing pathInfo and suffix there should not be end multiple slashes in $pathInfo
```
SilverFire: I'd like the following wording:

`Ensure, that $pathInfo does not end with more than one slash`

framework/web/UrlManager.php
andrewnester: Sounds reasonable to me

framework/web/UrlManager.php
andrewnester: yeah, this one sounds more clear

SilverFire:
Thanx

dynasource:
thanx

PR #10760, opened 2016-02-05
Fixed for arrayable models

boehsermoe:
The toArray() function return [attribute => value] but there is need a list of attribute names.


SilverFire:
Could you add a test for this?

cebe:
Also need a changelog line.

boehsermoe:
Unit test and changelog are done.

cebe:
Awesome, thanks!

Issue #10764, opened 2016-02-05
ListView options=>tag=>false doesn't work correctly


pkirill99:
Hi all

I need accordion, and use ListView (no Collapse), but ListView insert '<>' berore summary:

recipe:

view.php :
```
<div class="panel-group" id="accordion">
		<?= ListView::widget([
			'dataProvider' => $dataProvider,
			'itemView'     => '_item',
			'id'           => false,
			'options'      => [
				'tag' => false,
			],
			'itemOptions'  => [
				'tag' => false,
			],
		]); ?>
</div>
```

_item.php
```
<div class="panel panel-default">
	<div class="panel-heading">
		<h4 class="panel-title">
.........................................
```
and see in html:

```
<div class="panel-group" id="accordion">
 		<><div class="summary">............ <b>1-3</b> ..... <b>3</b>.</div>
 <div class="panel panel-default">
 	<div class="panel-heading">
 		<h4 class="panel-title">
```

but I expect to see

```
<div class="panel-group" id="accordion">
 		<div class="summary">............ <b>1-3</b> ..... <b>3</b>.</div>
 <div class="panel panel-default">
 	<div class="panel-heading">
 		<h4 class="panel-title">
```

SilverFire:
ListView is not designed to render no outer tag. You can contribute by providing a PR or wait until someone will pick up this feature request

pana1990:
this issue affect to BreadCrumbs and another widgets, is it posible change to logical in  Html::tag to render content without tags when $tag is false?

For example : 

```php
public static function tag($name, $content = '', $options = [])
{
        if ($name === false) {
              return $content;
        }
        $html = "<$name" . static::renderTagAttributes($options) . '>';
        return isset(static::$voidElements[strtolower($name)]) ? $html : "$html$content</$name>";
}

```

What do you think about it? @SilverFire @pkirill99

pkirill99:
good idea, it can minimize a code in the future, but I am not sure ... I don't know what ideas initially were put in a code

because this patch (solution) - copy from ListView->renderItem(....), that it was uniform

it is perhaps correct to do so

```
public static function tag($name, $content = '', $options = [])
{
        if ($name === false || $name === null) {
             return $content;
        }
        $html = "<$name" . static::renderTagAttributes($options) . '>';
        return isset(static::$voidElements[strtolower($name)]) ? $html : "$html$content</$name>";
}
```

or

```
public static function tag($name, $content = '', $options = [])
{
[...]

PR #10768, opened 2016-02-06
fix issue #10764

pkirill99:
fix issue https://github.com/yiisoft/yii2/issues/10764

### DIFF
--- a/framework/widgets/BaseListView.php
+++ b/framework/widgets/BaseListView.php
@@ -133,7 +133,11 @@ abstract class BaseListView extends Widget
 
         $options = $this->options;
         $tag = ArrayHelper::remove($options, 'tag', 'div');
-        echo Html::tag($tag, $content, $options);
+        if($tag !== false) {
+                return Html::tag($tag, $content, $options);
+        } else {
+                return $content;
+        }
     }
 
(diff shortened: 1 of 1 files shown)


framework/widgets/BaseListView.php
```diff
@@ -133,7 +133,11 @@ public function run()
 
         $options = $this->options;
         $tag = ArrayHelper::remove($options, 'tag', 'div');
-        echo Html::tag($tag, $content, $options);
+        if($tag !== false) {
```
SilverFire: Might be simplified to one-line with a ternary operator.

SilverFire:
Changelog and tests are required

SilverFire:
I'm closing this PR in favor of changing `Html::tag` behavior as described in https://github.com/yiisoft/yii2/issues/10764#issuecomment-180781200

pkirill99:
but also in BaseListView:

` echo Html::tag($tag, $content, $options);`

maybe

` return Html::tag($tag, $content, $options);`

?

cebe:
there is no real difference https://github.com/yiisoft/yii2/blob/master/framework/base/Widget.php#L107

PR #10771, opened 2016-02-06
change echo to return in BaseListView

pkirill99:
'echo' is not true ...

### DIFF
--- a/framework/widgets/BaseListView.php
+++ b/framework/widgets/BaseListView.php
@@ -133,7 +133,7 @@ abstract class BaseListView extends Widget
 
         $options = $this->options;
         $tag = ArrayHelper::remove($options, 'tag', 'div');
-        echo Html::tag($tag, $content, $options);
+        return Html::tag($tag, $content, $options);
     }
 
     /**


cebe:
for widgets that have no call to begin() and end() it does not matter:
https://github.com/pkirill99/yii2/blob/ab9d13834df5406672638072e0921833d66973a6/framework/base/Widget.php#L107

should be consistent about all widgets, @pkirill99 could you go through and make sure it is consistent?

pkirill99:
I am not sure that I understood you correctly, sorry for my bad english

if we look in https://github.com/yiisoft/yii2/blob/master/framework/base/Widget.php#L98 

`$out = $widget->run();`

and in https://github.com/yiisoft/yii2/blob/master/framework/base/Widget.php#L107

`return ob_get_clean() . $out;`

now - _echo_ send data in ob_ , and if use _return_ - data was in $out

pkirill99:
GridView also need fix if apply this patch

https://github.com/yiisoft/yii2/blob/master/framework/grid/GridView.php#L288

to 

`return parent::run();`

cebe:
moved this to 2.1, as it might break something.

cebe:
merged, thank you!

PR #10188, opened 2015-11-18
Enh #7405: Added `yii\filters\auth\AuthMetod::optional` property

SilverFire:
Closes #7405
- [x] Code
- [x] Tests
- [x] Changelog
- [x] Code review

Issue #7405, opened 2015-02-23
Optional authentication in REST


a-komarev:
I've faced situation when single action should be available to guest users and to authenticated users.
What is a best way to solve it?
First idea - implement new attribute `optional`:
```
'authenticator' => [
    'class' => CompositeAuth::className(),
    'optional' => ['feed', 'feed-item'],
    'authMethods' => [
        HttpBearerAuth::className(),
    ],
],
```
Methods `actionFeed` and `actionFeedItem` will perform Bearer authentication if header is passed. Otherwise authentication should be ignored and no errors should be raised.
Thats needed because application's pages for guest & authenticated users are nearly same, excluding a bit of additional data for authenticated guys.

fernandezekiel:
@a-komarev  does your specific action handle authenticated and guest users differently?

a-komarev:
@fernandezekiel authenticated users gets additional data in results.
```
[
    {
        "message": {
            "body": "First message",
            "favorite": true,
            "created_at": "2015-02-23 00:23:52"
        }
    },
]
```

Guest users are getting `favorite` attribute 'false' all the time, and authenticated users will get this value accordingly state in database.

fernandezekiel:
i see, i think though we can upgrade that to callable that has `$action` in the argument
```
'optional' => function ($action) {

}
```

a-komarev:
@fernandezekiel how will it help us?

fernandezekiel:
@a-komarev  sorry just throwing ideas out there,

jpodpro:
same situation here - it's a very simple scenario: a public page shows a list accessible to anyone. however the user is considered a guest even though they sent the bearer token. shouldn't yii process the token if it is sent? how else does a user remain "logged in" through out public and private pages on the site?

SilverFire:
https://github.com/yiisoft/yii2/commit/035a9ce0a8642dc1218f4b5d587dfde4daa382f2

PR #10776, opened 2016-02-07
fixed behavior of AuthMethod::isActive

cebe:
this is an adjustment of #10188 fixing  #7405
isActive() should not be affected in any way by specifying "optional" property.
isActive() should be determined from only and except.
If it is active it can then be checked if optional or not.
Currently this changes the behavior that except will always have precedence.

@SilverFire please merge if you agree.


tests/framework/filters/auth/AuthTest.php
```diff
@@ -140,6 +142,65 @@ public function testHttpBearerAuth($token, $login) {
...
+        $this->assertEquals(true, $method->invokeArgs($filter, [new Action('index', $controller)]));
+        $this->assertEquals(false, $method->invokeArgs($filter, [new Action('view', $controller)]));
+
+        $filter->only = ['index', 'view'];
+        $filter->except = ['view'];
+        $filter->optional = ['view'];
+        $this->assertEquals(true, $method->invokeArgs($filter, [new Action('index', $controller)]));
+        $this->assertEquals(false, $method->invokeArgs($filter, [new Action('view', $controller)]));
+
+        $filter->only;
```
SilverFire: What does this line do? Maybe, it must be `$filter->only = [];`?

SilverFire:
Thank you for a a smart notice, @cebe 
Agree to merge it after a conversation resolve

tests/framework/filters/auth/AuthTest.php
cebe: yes, should be the empty array case.

Issue #10165, opened 2015-11-16
DateValidator doesn't correctly validate DateTime values


VirtualRJ:
I have a model with the following rule:

```PHP
[['start'], 'date', 'timestampAttribute' => 'start'],
````

and the following code in my controller:

```PHP
$curTime = Yii::$app->formatter->asDateTime(new \DateTime('NOW'));
$model->start = $curTime;
```

When I run $model->validate it returns false. If I do the same, but change asDateTime to asDate it validates correctly.

I have even tried setting the formatter to the formatter defined by making the rule in the model like this:
```PHP
[['start'], 'date', 'format' => Yii::$app->formatter->datetimeFormat, 'timestampAttribute' => 'start'],
```

I presume this isn't expected behavior since as far as I can tell the DateValidator also validates DateTime and there is no specific seperate validator for DateTime.

SilverFire:
```php
$curTime = Yii::$app->formatter->asDateTime(new \DateTime('NOW')); // gives Nov 16, 2015, 5:18:59 PM
```
As it gives you string formatted in you local settings, it is really hard to guess, what are we expecting.

Create the `$curTime` in specific format and check for the same specific format:

```php
$curTime = Yii::$app->formatter->asDateTime(new \DateTime('NOW'), 'yyyy-MM-dd HH:mm:ss');
$curTime = Yii::$app->formatter->asDateTime(new \DateTime('NOW'), 'php:Y-m-d H:i:s'); // the same as previous, but date format is in PHP date() format
```

Then specify the expected format in the validator:

```php
[['start'], 'date', 'format' => 'php:Y-m-d H:i:s'],
```

VirtualRJ:
Yes, that does work, but it is still strange (and unexpected) behavior. Even if I say:

```PHP
$curTime = Yii::$app->formatter->asDateTime('now', Yii::$app->formatter->datetimeFormat);
```
and make the rule like:
```PHP
[['start'], 'date', 'format' => Yii::$app->formatter->datetimeFormat, 'timestampAttribute' => 'start'],
```
It returns false. In that case I don't think there should be any unexpected formatting issue even with localisation.

VirtualRJ:
I did some digging. The reason why it does work in the example is in the way parseDateValueIntl($value, $format) works in the DateValidator. I'm not completely sure how to describe it, but if something is unclear please let me know.

```PHP
$format = Yii::$app->formatter->datetimeFormat; //$format = 'medium'
```
[...]

Issue #7177, opened 2015-02-05
DateValidator add support for time values


nkovacs:
DateValidator sets the time part to 00:00:00 when inserting the timestamp into timestampAttribute, because [here](https://github.com/yiisoft/yii2/blob/58becaceab5cf3f7896d7934de715cecfd95c210/framework/validators/DateValidator.php#L133) the `$timetype` argument is set to `IntlDateFormatter::None`.
It doesn't help if you use `Yii::$app->formatter->datetimeFormat` as a format, because both `Yii::$app->formatter->datetimeFormat` and `Yii::$app->formatter->dateFormat` are "medium", so DateValidator doesn't know if you want to keep the time or not.

The other problem is that even though it cuts off the time part, it still accepts input that contains a time part.

cebe:
DateValidator is not designed to work with time values currently.

nkovacs:
I did some more testing. It works fine if you specify a format in php syntax.
If you use the ICU syntax and specify a time component, it won't cut off the time. But if the format does not contain a time component, it will still accept values that contain a time.

So as a workaround, you can either use `IntlDateFormatter::getPattern` with an appropriately constructed formatter to fix the first bug, or `\yii\helpers\FormatConverter::convertDateIcuToPhp(Yii::$app->formatter->datetimeFormat)` to fix both, but the latter can't convert some special formats.

cebe:
> But if the format does not contain a time component, it will still accept values that contain a time.

this should now be fixed by #6553

nkovacs:
I made an extension that fixes this: https://github.com/nkovacs/yii2-datetimepicker/blob/master/DateTimeValidator.php

Of course there would be no need to convert the short format to a full pattern if implemented in Yii's validator, but I had to do it so I wouldn't have to reimplement all the code in it.

cebe:
should be fixed by #5053 (d60445c066003cab78fb16458e7234fae5db4765)

cebe:
hm okay, not really.

nkovacs:
Not entirely. If you use an intl short format (which is the default), the validator still uses IntlDateFormatter::NONE for the time value: https://github.com/yiisoft/yii2/commit/d60445c066003cab78fb16458e7234fae5db4765#diff-32236b8409f422c4a80339d1ef6466e0R167

cebe:
yep, just seen that while reading the related issue.

cebe:
resolved by  2ac17dd...00d506c

PR #10778, opened 2016-02-07
Enhancement for DateValidator (#7177, #10165):

VirtualRJ:
- Added parameter to define if checking for 'date' or 'dateTime' and updated parseDateValueIntl($value, $format) to facilitate this validation based on defined type.
- Added unit tests to for validating dateTime values when using short format validation.


VirtualRJ:
@cebe I have not added a line to changelog since there was no section for 2.0.8 yet and not sure which version this will be included (if at all).

Just so this pull request will be referenced in the issues: (#7177, #10165)

framework/validators/DateValidator.php
```diff
@@ -143,7 +143,15 @@ class DateValidator extends Validator
...
     public $minString;
-
+    /**
+     * @var string type used for defining if short format validation should be on just
+     * date or on datetime. Defaults to 'date'. Possible values are:
+     * - `date`
+     * - `dateTime`
+     * @since 2.0.8
+     */
+    public $type = 'date';
```
SilverFire: I'm suggesting to introduce constants for types

SilverFire:
We need a changelog line

SilverFire:
What about JS?

cebe:
> What about JS client-side validation?

DateValidator does not have client validation as it would be too complex to duplicate the functionality in JS.

SilverFire:
Ok

VirtualRJ:
Added the suggested constants.

@SilverFire, @cebe: Where should I add the line for the changelog? As mentioned earlier there is no section for 2.0.8 (according to the issue this is added to the 2.0.8 milestone) and I believe 2.0.7 is frozen for release, right?

samdark:
@VirtualRJ you're right. You may post changelog line right here in comments, we'll handle it during merge.

VirtualRJ:
Ok! Changelog line would be:

```
- Enh #7177, #10165: Added support for validating datetime using short format (VirtualRJ)
```

cebe:
merged, with a few adjustments: 2ac17dd..00d506c Thank you!

PR #10782, opened 2016-02-07
IpValidator: use message property for default error

cebe:
its better to use already existing `message` property for the default error message.

breaking change so it should be decided before 2.0.7 release.


SilverFire:
Agree on this. Will resolve the conflict and merge soon

SilverFire:
Updated from master, added examples to PHPDocs.
@cebe please, check out this [comment](https://github.com/yiisoft/yii2/commit/e41c2e39469992a67b9d2c2e839d61cd57cf8e3e#commitcomment-15971059)

PR #10786, opened 2016-02-08
Parameter "$file" can be a URL in method "getMimeType"

shcherbanich:
If `$file` is url, finfo_file return error: "finfo_file(): Failed identify data 0:no magic files loaded". In this case, I propose to use this construction: `finfo_buffer($info, file_get_contents($file))`.


samdark:
File helper works with local files and I don't think it's a good idea to mix in remote ones as well.

cebe:
if you want mime type of remote file I'd rather rely on `Content-Type` header than guessing it from the content.

PR #10813, opened 2016-02-11
Advanced table alias handling for Query

### LINKED ISSUE #7263: Disambiguing column's table on ActiveQuery select/addSelect

### LINKED ISSUE #8743: defaultScope, relation alias

### LINKED ISSUE #9326: ActiveQuery getAlias

cebe:
This is the basis to implement a solution for #2377 on top of it, which is #10817.

Allows to disambiguate colums without knowing the alias that has been
defined for a table elsewhere (e.g. in a relation definition).
- fixes #8743 and fixes #7263 by introducing alias syntax on querybuilder level `{{@table}}` to resolve table aliases.
  similar to table prefix and table name escape syntax.
- fixes #9326 by adding `getAlias()` and `applyAlias()` to `Query`
- fixes #4972 by adding and `alias()` to `ActiveQuery`

TODO:
- [x] add CHANGELOG lines.


docs/guide/db-query-builder.md
```diff
@@ -620,6 +620,53 @@ $query = (new \yii\db\Query())
 The anonymous function takes a parameter `$row` which contains the current row data and should return a scalar
 value which will be used as the index value for the current row.
 
+## Dealing with table aliases <span id="aliases"></span>
+
+Some use cases of query builder are to create an initial [[yii\db\Query]] object in one place and adjust
+it in other places of the application by adding further conditions or to change the sorting options.
```
samdark: `or to change options such as sorting`?

docs/guide/db-query-builder.md
```diff
@@ -620,6 +620,53 @@ $query = (new \yii\db\Query())
...
+When more than one table is used in the query, you may need to disambiguate column names to be explicit about
+which column a condition refers to, e.g. `user.id` vs. `post.id`.
```
samdark: don't need `.` after `vs`

docs/guide/db-query-builder.md
```diff
@@ -620,6 +620,53 @@ $query = (new \yii\db\Query())
...
+
+Since version 2.0.7 Yii provides methods that help you with disambiguating column names without the need
+to rely on a specific alias being defined in the original query. You may call the [[yii\db\Query::getAlias()]] method
+to get the alias name of a table in the query context, or [[yii\db\Query::applyAlias()]] to disambiguate a column.
+
+The following example shows how these methods are used:
+
+Consider this method to be defined in the `User` class:
```
samdark: this method → the following method

docs/guide/db-query-builder.md
```diff
@@ -620,6 +620,53 @@ $query = (new \yii\db\Query())
...
+
+```php
+public static function getActiveUserQuery()
+{
+    return (new \yii\db\Query)->from(['u' => 'user'])->where(['active' => 1]);
+}
+```
+
+The following code, which is located elsewhere in the application, calls this methods and needs to adjust the query.
```
samdark: calls this methods → calls this method

docs/guide/db-query-builder.md
```diff
@@ -620,6 +620,53 @@ $query = (new \yii\db\Query())
...
+
+By using [[yii\db\Query::applyAlias()|applyAlias()]] here, this code will still work, even when the alias is changed or removed
+inside of the `getActiveUserQuery()` method implementation.
+
+In some situations the alias may not be known at the time the query is created because we allow to change the alias later in the code.
+Therefor a special syntax for referring to a table alias
+can be used which is simlar to the syntax used for [quoting table names](db-dao.md#quoting-table-and-column-names) and
+the [table prefix feature](db-dao.md#using-table-prefix).
+
+To refer to a table alias, the following syntax may be used:
```
samdark: This sentence could be removed. `:` could be moved to sentence ↑

framework/db/ActiveQuery.php
```diff
@@ -721,4 +724,35 @@ public function viaTable($tableName, $link, callable $callable = null)
...
+    {
+        if (empty($this->from) || count($this->from) < 2) {
+            list($tableName, ) = $this->getQueryTableName($this);
+            $this->from = [$alias => $tableName];
+        } else {
+            /* @var $modelClass ActiveRecord */
+            $modelClass = $this->modelClass;
+            $tableName = $modelClass::tableName();
+
+            foreach($this->from as $key => $table) {
```
samdark: Space missing after `foreach`

framework/db/Query.php
```diff
@@ -844,6 +849,119 @@ public function addParams($params)
...
+
+    /**
+     * Populate table aliases used in this query.
+     *
+     * This function is used by [[from()]] and [[join()]] to make information about
+     * tables aliases available for [[getAlias()]].
+     *
+     * @param array $tables array of tables in the format allowed for [[from()]]:
+     *
+     * - numeric key, string value == table without alias, or table with alias specified after the name, separated by white space.
```
samdark: It's better to use `-` rather than `==` for descriptions.

framework/db/Query.php
```diff
@@ -844,6 +849,119 @@ public function addParams($params)
...
+     */
+    protected function populateAliases($tables)
+    {
+        foreach ($tables as $alias => $tableName) {
+            if (is_object($tableName)) {
+                continue;
+            }
+            if (is_string($alias)) {
+                $this->_aliases[$tableName] = $alias;
+            } elseif (preg_match('/^(\\{\\{(\w+)\\}\\}|(.*?))(\s+AS\s+|\s+)(\\{\\{\w+\\}\\}|\w+)$/i', $tableName, $matches)) {
```
samdark: It desn't look like all group results are used. Consider using `(?: something )`.

framework/db/Query.php
```diff
@@ -844,6 +849,119 @@ public function addParams($params)
...
+    }
+
+    /**
+     * Disambiguate a column name by looking up the table alias.
+     *
+     * This method will look up the alias for a table using [[getAlias()]] and
+     * return the disambiguated column name. e.g. for `applyAlias('user', 'name')`
+     * it will return `u.name` if the table alias was `u`.
+     *
+     * This function can be used to disambiguate column names in situations
```
samdark: This function can be used → It can be used

Because you're calling it a method above.

framework/db/ActiveQuery.php
```diff
@@ -721,4 +724,35 @@ public function viaTable($tableName, $link, callable $callable = null)
...
+     * Define an alias for the table defined in [[modelClass]].
+     *
+     * This method will adjust [[from]] so that an already defined alias will be overwritten.
+     * If none was defined, [[from]] will be populated with the given alias.
+     *
+     * @param string $alias the table alias.
+     * @return $this the query object itself
+     * @since 2.0.7
+     */
+    public function alias($alias)
```
klimov-paul: Looks weird:

```php
(new Query())->select(...)
    ->from()
    ->joinWith()
    ->alias() // ???
    ->all();
```

It simply don't fit the DSL. While writing normal SQL you don't deal with 'alias' section.

framework/db/QueryBuilder.php
```diff
@@ -122,6 +122,15 @@ public function build($query, $params = [])
...
         }
 
+        // replace aliases with table names
+        $sql = preg_replace_callback(
+            '/\\{\\{@([\w\-\. ]+)\\}\\}/',
+            function ($matches) use ($query) {
+                return '{{' . $query->getAlias($matches[1]) . '}}';
+            },
+            $sql
+        );
```
klimov-paul: This looks weird. Why do we perform alias post-processing in SQL?
Can't we deal with aliases during SQL composition, instead of after it is complete?

klimov-paul: Ah, I see now.

klimov-paul: It make sense.

framework/db/ActiveQuery.php
klimov-paul: It is unlikely main model table will need an alias.
Someone, who is willing to add alias for it anyway may always use `from` if needed.
I don't think there is a necessity for `alias()` method here.

framework/db/sqlite/QueryBuilder.php
```diff
@@ -394,6 +394,15 @@ public function build($query, $params = [])
...
         }
 
+        // replace aliases with table names
+        $sql = preg_replace_callback(
+            '/\\{\\{@([\w\-\. ]+)\\}\\}/',
+            function ($matches) use ($query) {
+                return '{{' . $query->getAlias($matches[1]) . '}}';
+            },
+            $sql
+        );
```
klimov-paul: Does this somehow different from the one at `QueryBuilder`?

klimov-paul: Perhaps a separated method should be extracted.
Perhaps it should be somewhere at `yii\db\Schema` allong with `quoteTableName()` and `quoteColumnName()`.

framework/db/Query.php
```diff
@@ -844,6 +849,119 @@ public function addParams($params)
     }
 
     /**
+     * @var array aliases defined for tables used in this query. `tablename => alias`.
```
klimov-paul: > `tablename => alias`

I am unsure: does this means single table name can have only one alias?
Was not original issue about using same table twice inside single query?

For example, I may need to compose following SQL:
```
SELECT u.*, COUNT(sent_m) AS sentCount, COUNT(received_m) AS receiveCount
FROM User AS u
LEFT JOIN PrivateMessage AS sent_m ON sent_m.senderId = u.id
LEFT JOIN PrivateMessage AS received_m ON received_m.receiverId = u.id
GROUP BY u.id
```

What should I do then?

framework/db/ActiveQuery.php
cebe: note, that this method is added for ActiveQuery only. The use case is e.g. the following:

```php
// in Post AR class
public function getAuthor()
{
     return $this->hasOne(User::classname(), ['id', 'author_id'])->alias('author');
}
```

framework/db/sqlite/QueryBuilder.php
cebe: it needs to be within Query and QueryBuilder layer because the lower layer classes like Schema and connection do not have anything to do with this feature.
About duplicated code problem in this case I have opened: https://github.com/yiisoft/yii2/issues/10815

framework/db/ActiveQuery.php
klimov-paul: May be rename it to `as()`?

cebe: Not sure, `alias` in context of a query is more clear than `as`, it defines an alias for the main table in the query and not uses the query "as" something.

framework/db/Query.php
cebe: getAlias is not suited for this case, see docs: https://github.com/yiisoft/yii2/pull/10813/files#diff-586d87573b57677fb773d9ca73cf20b3R893

cebe: >  Was not original issue about using same table twice inside single query?

that was the issue with joinWith() where you can specify the alias with the changes in the other PR.
If a table is used twice you have to use the alias yourself to refer to the correct table. This method can not help you with that.

framework/db/sqlite/QueryBuilder.php
klimov-paul: OK, fine.

framework/db/Query.php
klimov-paul: If we can not deal with multiple aliases, why we can not limit ourselves with applying alias to `ON` at `JOIN`, and leave everything else to developer?

cebe: multiple aliases for the same table is quite a rare case and not covered here. This syntax is to allow writing query conditions without the need to care about which alias was defined for a given table. The use cases are queries that are composed in different places over the application, e.g. starting in the relation definition of AR or the default scope, which are later modified in other places of the application.

cebe:
@klimov-paul thanks for your comments! Is this ready for merge now?
@yiisoft/core-developers anything to adjust here?

docs/guide/db-query-builder.md
```diff
@@ -620,6 +620,51 @@ $query = (new \yii\db\Query())
...
 value which will be used as the index value for the current row.
 
+## Dealing with table aliases <span id="aliases"></span>
+
+Some use cases of query builder are to create an initial [[yii\db\Query]] object in one place and adjust
+it in other places of the application by adding further conditions or to change options such as sorting.
+When more than one table is used in the query, you may need to disambiguate column names to be explicit about
+which column a condition refers to, e.g. `user.id` vs `post.id`.
+
+Since version 2.0.7 Yii provides methods that help you with disambiguating column names without the need
```
SilverFire: help -> helps

framework/db/Query.php
SilverFire: EAV is a quite popular thing and it works on one table with different aliases.
Agree with @klimov-paul

klimov-paul:
I am against such slution: it overcomplicates the `Query` interface without significant benefit.

In my opinion method `getAlias()` has no purpose. If I join particular table once I usually have no need in alias. Even if I specify alias for the table, why do I need this method to rememeber it? Does developer forgets the alias he gave to the table right away?
If I join same table several times, which is a main case for the table alias usage, all these additions can not help me anyhow!

> The use cases are queries that are composed in different places over the application, e.g. starting in the relation definition of AR or the default scope, which are later modified in other places of the application.

How does this help with the original issue? If I join same relation twice all my preparations for the aliases (`{{@tableName}}`) will not help me - they will just produce an SQL error.
Thus I am still unable to join same relation several time - the issue is not resolved.

docs/guide/db-query-builder.md
cebe: no, it's plural. "multiple methods that help you" vs. "one method that helps you"

cebe:
moved to next milestone for further discussion. some parts of this will go into 2.0.7 via #10828.

vercotux:
Any chance this will get into 2.0.8? It would be really great to have this excellent feature.

framework/CHANGELOG.md
```diff
@@ -64,9 +64,11 @@ Yii Framework 2 Change Log
 - Bug: Fixed generation of canonical URLs for `ViewAction` pages (samdark)
 - Bug: Fixed `mb_*` functions calls to use `UTF-8` or `Yii::$app->charset` (silverfire)
 - Enh #3506: Added `yii\validators\IpValidator` to perform validation of IP addresses and subnets (SilverFire, samdark)
+- Enh #4972: Added `yii\db\ActiveQuery::alias()` to allow specifying a table alias for the model table without having to know the name (cebe, stepanselyuk)
 - Enh #5146: Added `yii\i18n\Formatter::asDuration()` method (nineinchnick, SilverFire)
 - Enh #5902: `yii\widgets\Pjax::options` now support special option `tag` to specify tag of container (Alex-Code)
 - Enh #6162, #9198: Added parameter visibleButtons for `yii\grid\ActionColumn` (fornit1917, silverfire)
+- Enh #7263, #8743: Added a new syntax for resolving table aliases to `yii\db\Query`, `{{@tablename}}` will now be replaced by the table alias (cebe)
```
Faryshta: about this. can it be anidated? `{{@user.profile}}.relationship_status`

cebe: what do you mean with "anidated"?

Faryshta: Sorry I meant nested.

CedricYii:
@klimov-paul The issue of joining a same relation several times is made possible with a simple syntax by #11646. Does it answer your concern?

klimov-paul:
In order to fix the issue we should return the solution similar to the one we used at Yii 1.
back then we used special classes for the join tree building:
https://github.com/yiisoft/yii/blob/master/framework/db/ar/CActiveFinder.php#L315
https://github.com/yiisoft/yii/blob/master/framework/db/ar/CActiveFinder.php#L1228
And we have more or less fine aliases processing there.

CedricYii:
However as far as I remember in Yii1, it was static aliasing.
But the idea to have a property dedicating for the alias was a good point.

PR #10817, opened 2016-02-11
Added Alias Syntax for joinWith()

### LINKED ISSUE #2377: Table alias for joinWith needed

cebe:
Add alias syntax to joinWith(), e.g. joinWith('author a').
No need to know the table name for defining an alias for the relation.

fixes #2377, alternative to #8788, which allows later implementation of
getting alias and column name ambiguation.

depends on #10813 to be merged first.

TODO:
- [x] fix tests

### DIFF
--- a/docs/guide/db-active-record.md
+++ b/docs/guide/db-active-record.md
@@ -1089,7 +1089,28 @@ Note that this differs from our earlier example which only brings back customers
 > Info: When [[yii\db\ActiveQuery]] is specified with a condition via [[yii\db\ActiveQuery::onCondition()|onCondition()]],
   the condition will be put in the `ON` part if the query involves a JOIN query. If the query does not involve
   JOIN, the on-condition will be automatically appended to the `WHERE` part of the query.
+  Thus it may only contain conditions including columns of the related table.
 
+#### Relation table aliases <span id="relation-table-aliases"></span>
+
+As noted before, when using JOIN in a query, we need to disambiguate column names. Therefor often an alias is
+defined for a table. Setting an alias for the relational query would be possible by customizing the relation query in the following way:
+
+```php
+$query->joinWith([
+    'orders' => function ($q) {
+        $q->from(['o' => Order::tableName()]);
+    },
+])
+```
+
+This however looks very complicated and involves either hardcoding the related objects table name or calling `Order::tableName()`.
+Since version 2.0.7, Yii provides a shortcut for this. You may now define and use the alias for the relation table like the following:
+
+```php
+// join the orders relation and sort the result by orders.id
+$query->joinWith(['orders o'])->orderBy('o.id');
+```
 
 ### Inverse Relations <span id="inverse-relations"></span>
 
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -357,11 +357,19 @@ class ActiveQuery extends Query implements ActiveQueryInterface
      * This method differs from [[with()]] in that it will build up and execute a JOIN SQL statement
      * for the primary table. And when `$eagerLoading` is true, it will call [[with()]] in addition with the specified relations.
      *
-     * @param string|array $with the relations to be joined. Each array element represents a single relation.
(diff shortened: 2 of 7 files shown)


docs/guide/db-active-record.md
```diff
@@ -1089,7 +1089,28 @@ Note that this differs from our earlier example which only brings back customers
...
+
+```php
+$query->joinWith([
+    'orders' => function ($q) {
+        $q->from(['o' => Order::tableName()]);
+    },
+])
+```
+
+This however looks very complicated and involes either hardcoding the related objects tablename or calling `Order::tableName()`.
```
samdark: involes → invol**v**es

framework/db/ActiveQuery.php
```diff
@@ -384,8 +396,33 @@ protected function queryScalar($selectExpression, $db)
...
     {
-        $this->joinWith[] = [(array) $with, $eagerLoading, $joinType];
+        $relations = [];
+        foreach((array) $with as $name => $callback) {
+            if (is_int($name)) {
+                $name = $callback;
+                $callback = null;
+            }
+
+            if (preg_match('/^(.*?)(?:\s+AS\s+|\s+)(\w+)$/i', $name, $matches)) {
```
samdark: No matches are used. Consider using `(?: something )`

tests/framework/db/ActiveRecordTest.php
```diff
@@ -525,6 +551,120 @@ public function testJoinWithVia()
...
+        $this->assertEquals(2, $orders[0]->id);
+        $this->assertEquals(3, count($orders[0]->items));
+        $this->assertTrue($orders[0]->items[0]->isRelationPopulated('category'));
+        $this->assertEquals(2, $orders[0]->items[0]->category->id);
+
+    }
+
+
+
+    // TODO join a relation twice, once with
```
samdark: ?

framework/db/ActiveQuery.php
```diff
@@ -384,8 +396,33 @@ protected function queryScalar($selectExpression, $db)
...
+                    }
+                };
+            }
 
+            if ($callback === null) {
+                $relations[] = $name;
+            } else {
+                $relations[$name] = $callback;
+            }
+        }
```
klimov-paul: It whould be better to perform all this processing at `buildJoinWith()` method, leaving this one intact.

cebe: one benefit of doing it here would be, that one could later add methods like `getAlias()` for relation names e.g. `getRelationAlias()`. That is the main reason why I put it here.

klimov-paul:
At the first glance, all we have to do is composition of the valid `ON` condition for the join. Anything else - select, order by and so on - should be left to developer to deal with.
The goal is to achieve the ability to call `joinWith()` several times per **same** relation with different aliases without SQL error.

klimov-paul:
The root of the problem is while `joinWith()` composes `ON` condition it does not take an alias into account.
It seems the fix does not have to be so complex.

cebe:
This PR is not for fixing a bug but for introducing a nicer way of defining table name for relations.

For `join()` you could do:

```php
$query->join('orders o')->orderBy('o.id');
```

for joinWith() this currently looks like:

```php
$query->joinWith([
    'orders' => function ($q) {
        $q->from(['o' => Order::tableName()]);
    },
])
```

what is added here is simlified syntax that allows to do the same as in join():

```php
$query->joinWith(['orders o'])->orderBy('o.id');
```

I have not verified what happens when the same relation is joined twice. Will write a test for it, but that is not the main purpose of this PR.

cebe:
@klimov-paul added a test to verify that ON is built correctly.

cebe:
tests are passing. @yiisoft/core-developers please let me know if you have further comments or if you think it is ready to merge.

SilverFire:
:+1:

klimov-paul:
Seems to be all fine.
:+1:

cebe:
closed in favor of #10828 which has no dependency on the Query changes but still adds `alias()` method to ActiveQuery.

Issue #10673, opened 2016-01-26
Add Date range formatter


cebe:
format date ranges like `1.-5. January 2016`, `Oct 4-6, 2016`.

```php
asDateRange($fromDate, $toDate)
```

similar implementation in ruby: https://github.com/darkleaf/date_range_formatter

cebe:
https://github.com/yiisoft/yii2/pull/10820#r58304388

> There is formatting information about this in intl ICU data. I think we should take advantage of that:
>
> e.g. http://intl.rmcreative.ru/tables?locale=de search for `intervalFormats` on that page.
>
--

> Implementing this properly is not easy, we might take advantage of existing implementations such as https://developer.android.com/reference/android/icu/text/DateIntervalFormat.html
source code: https://github.com/android/platform_external_icu4c/blob/e5311394ca22b280da41cd17059288dab3fb1ea6/i18n/dtitvfmt.cpp

PR #10820, opened 2016-02-11
Date range formatter issue #10673

vorontsov-pavel:
Hello. This is my first revision Yii2 framework. I added a feature formatter for easy reading of the date range. I have also documented this function, and add a test for it. I hope you all want, and you make this functionality in the next release Yii2.

issue #10673

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -19,6 +19,7 @@ use yii\base\InvalidConfigException;
 use yii\base\InvalidParamException;
 use yii\helpers\FormatConverter;
 use yii\helpers\HtmlPurifier;
+use yii\helpers\ArrayHelper;
 use yii\helpers\Html;
 
 /**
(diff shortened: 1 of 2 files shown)


cebe:
cool, thank you for this! Will review it in detail for next release.

framework/i18n/Formatter.php
```diff
@@ -168,6 +169,21 @@ class Formatter extends Component
...
+                'day' => 'j',
+                'time' => 'H:i',
+                'templates' => [
+                    'notDiff' => '{from_time} {from_day} {from_month} {from_year}',
+                    'diffYear' => '{from_time} {from_day} {from_month} {from_year} - {to_time} {to_day} {to_month} {to_year}',
+                    'diffMonth' => '{from_time} {from_day} {from_month} - {to_time} {to_day} {to_month} {to_year}',
+                    'diffDay' => '{from_time} {from_day} - {to_time} {to_day} {to_month} {to_year}',
+                    'diffTime' => '{from_time} - {to_time} {to_day} {to_month} {to_year}'
+                ]
+            ];
```
cebe: There is formatting information about this in intl ICU data. I think we should take advantage of that:

e.g. http://intl.rmcreative.ru/tables?locale=de search for `intervalFormats` on that page.

cebe: Implementing this properly is not easy, we might take advantage of existing implementations such as https://developer.android.com/reference/android/icu/text/DateIntervalFormat.html
source code: https://github.com/android/platform_external_icu4c/blob/e5311394ca22b280da41cd17059288dab3fb1ea6/i18n/dtitvfmt.cpp

PR #10826, opened 2016-02-12
Fix defaultRoute for console application

109th:
if configuration file contains a defaultRoute parameter, it will overwrites a variable defaultRoute.


klimov-paul:
This is exected behavior. With such patch there is no way to change default route.

109th:
Thank you for answer, i didn't known about this feature :+1:

PR #10828, opened 2016-02-12
Activequery::joinWith() alias

### LINKED ISSUE #2377: Table alias for joinWith needed

cebe:
this is the final solution to fix #2377 and #4972.

replaces #10817.
- [x] TODO: agree on a name for `alias()` method. A short summary about discussion so far:

`alias()` is short and quite clear about what it does. it defines an alias for something.

`tableAlias()` adds no further information as we are deailing with tables only anyway here.
"alias" should be part of the name as that is what it does.
not sure if we can add a word that makes it clear that it is the alias for the base table and not for joined ones.

`modelTableAlias()` as proposed by @klimov-paul is very specific about what it does but a bit long.
we already have `modelClass` inside an `ActiveQuery`
thus he supposed it should be `modelTableAlias`

`baseAlias()` as proposed by @mdomba is shorter but maybe not clear enough about what `base` means.

@qiangxue @samdark @SilverFire can we have your input on the name?


qiangxue:
Looks great to me (so is the name `alias()`)?

PR #10835, opened 2016-02-14
Update input-file-upload.md

dr1v3:
An important note about saveAs method

### DIFF
--- a/docs/guide/input-file-upload.md
+++ b/docs/guide/input-file-upload.md
@@ -46,7 +46,7 @@ class UploadForm extends Model
 
 In the code above, the `imageFile` attribute is used to keep the uploaded file instance. It is associated with
 a `file` validation rule which uses [[yii\validators\FileValidator]] to ensure a file with extension name `png` or `jpg`
-is uploaded. The `upload()` method will perform the validation and save the uploaded file on the server.
+is uploaded. The `upload()` method will perform the validation and save the uploaded file on the server. 
 
 The `file` validator allows you to check file extensions, size, MIME type, etc. Please refer to
 the [Core Validators](tutorial-core-validators.md#file) section for more details.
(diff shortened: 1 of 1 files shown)


docs/guide/input-file-upload.md
```diff
@@ -55,6 +55,12 @@ the [Core Validators](tutorial-core-validators.md#file) section for more details
   implemented via [[yii\validators\ImageValidator]] which verifies if an attribute has received a valid image 
   that can be then either saved or processed using the [Imagine Extension](https://github.com/yiisoft/yii2-imagine).
 
+If you need to validate your model after use of `saveAs` method, especially with the `image` validator, you need to add `false` paramether to `saveAs`, because this method deletes an uploaded file during saving by default and validator will give out an error.
```
cebe: > If you need to validate your model after use of `saveAs` method

why would I do that?

dr1v3: If $uploadedfile is a property of ActiveRecord model, every save() method calls validate() method too. Usually I process uploaded file through separate method before saving the db record or in a beforeSave method. I know, that processing forms in AR model isn't the best practice, but it's really often usecase.

githubjeka: The bad and too vague note.
In your case: you should save your new $filename, if you need.

dr1v3: > The bad and too vague note.
May be. But if I found out this option earlier in this guide, it would have saved me from hours of debug and invent of a bicycles.

githubjeka: http://www.yiiframework.com/doc-2.0/yii-web-uploadedfile.html#saveAs()-detail

> Whether to delete the temporary file after saving. If true, you will not be able to save the uploaded file again in the current request.

dr1v3: Yep. But i talking about The Guide, not a phpdoc comment. Newbies reads Guide first.

cebe:
it is not clear what the prupose of this change is. it looks like wrong usage to me, so I am closing this PR. Thanks for trying to improve the docs anyway!

Issue #9457, opened 2015-08-18
Case-sensitivity in OCI still does not work


dryu:
I have OCI driver declaration to put all results to the lower case:

```php
		'attributes' => [
			PDO::ATTR_CASE => PDO::CASE_LOWER,
		],
```

This means that I want ALL query results coming from OCI driver should have column names in the lower case. However, with AR classes I still have those populated incorrectly or not populated at all, because OCI driver in Yii2 reads the schema in the upper case and populates AR ```attributes``` with upper case column names values, resulting in failures when trying to access properties of AR class.

So far I'm unable to work with OCI AR classes at all.

Yii 2.0.4

samdark:
Can you check 2.0.6?

dryu:
Checked with 2.0.6. Initially it hard failed in OCI Schema class:

```
2015-08-18 09:04:25 [-][-][-][error][yii\base\ErrorException:8] exception 'yii\base\ErrorException' with message 'Undefined index: CONSTRAINT_TYPE' in C:\inetpub\sites\qtool\vendor\yiisoft\yii2\db\oci\Schema.php:283
Stack trace:
#0 C:\inetpub\sites\qtool\vendor\yiisoft\yii2\db\oci\Schema.php(283): yii\base\ErrorHandler->handleError(8, 'Undefined index...', 'C:\\inetpub\\site...', 283, Array)
#1 C:\inetpub\sites\qtool\vendor\yiisoft\yii2\db\oci\Schema.php(88): yii\db\oci\Schema->findConstraints(Object(yii\db\TableSchema))
```

The underlying code obviously has an issue - it should correct the row case first, then work with it:

```php
        foreach ($command->queryAll() as $row) {
            if ($row['CONSTRAINT_TYPE'] !== 'R') {
                // this condition is not checked in SQL WHERE because of an Oracle Bug:
                // see https://github.com/yiisoft/yii2/pull/8844
                continue;
            }
            if ($this->db->slavePdo->getAttribute(\PDO::ATTR_CASE) === \PDO::CASE_LOWER) {
                $row = array_change_key_case($row, CASE_UPPER);
            }
...
}
```

But even when I moved CONSTRAINT_TYPE check, this is still fails with the same original exception:

```
2015-08-18 09:02:49 [-][-][-][info][yii\db\Command::query] SELECT ..., "CREATION_DATE", ... FROM "X"."Y" WHERE "ID"=Z ORDER BY "CREATION_DATE"
[...]

PR #10861, opened 2016-02-16
Add case sensitivity support for Oracle

phelan:
Issue #9457 
but only fix mysql part


SilverFire:
Thank you for your contribution.
Could you add a unit-test for this change and try to do the same for other DBMSs, please?

samdark:
Closing since it cannot be merged in this state.

Issue #10869, opened 2016-02-16
Add Hint Index in queries


fdezmc:
I need force index in my project, so I've add the method addHintIndex to set the hints indexes associated to a table.
The param is an assoc array of the hintIndex(s) to be used on the query.
The key is the name of the table to be hinted.
The value is the hint settings or an array of hints settings to apply. 
Each hint setting is compose by one to three string values defining the hint and one array defining the index(s) to hint.

Here are some examples:

```
$query = (new \yii\db\Query())->from(['t' => 'user'])->addHintIndex(['user' => [
    ['force', 'index', ['primary']],
    ['ignore', 'index', 'order by', ['i1']],
]])->leftJoin('profile as p', 'user.id = profile.user_id')->addHintIndex(['profile' => [
    'use', 'index', ['i2']
]]);

// sql -> SELECT * FROM `user` `t` FORCE INDEX (primary) IGNORE INDEX FOR ORDER BY (i1)
 LEFT JOIN `profile` `p` ON user.id = profile.user_id USE INDEX (i2)
```

```
$users = User::find()->addHintIndex(['user' => [['use', 'index', ['primary']]]])->one();

// sql -> SELECT * FROM {{%user}} USE INDEX (primary)
```

I have no experience enought with other drivers than mysql, so just write functionality for mysql and sqlite drivers (the former tested the last pending for testing).

Driver overwritting can be done on:
- _/yii2/framework/db/mysql/QueryBuilder.php_
- _/yii2/framework/db/sqlite/QueryBuilder.php_ (require testing)*
- _/yii2/framework/db/cubrid/QueryBuilder.php_ (trace a message and ignore addHintIndex)*
- _/yii2/framework/db/mssql/QueryBuilder.php_ (trace a message and ignore addHintIndex)*
- _/yii2/framework/db/oci/QueryBuilder.php_ (trace a message and ignore addHintIndex)*
- _/yii2/framework/db/pgsql/QueryBuilder.php_ (trace a message and ignore addHintIndex)*

cebe:
As far as I see this is already possible with existing methods. no need to add anything:

```php
// your first example:
$query = (new Query)->from([new Expression('{{%user}} USE INDEX (primary)')]);
// your second example
$query = (new Query)
       ->from([new Expression('`user` `t` FORCE INDEX (primary) IGNORE INDEX FOR ORDER BY (i1)')])
[...]

PR #10872, opened 2016-02-16
10869 query hint index

fdezmc:
Code and unit tests for query hint index #10869


samdark:
Tests seem to error: https://travis-ci.org/yiisoft/yii2/jobs/109650849

cebe:
Thank you for your contribution. The code looks good and also the documentation you added.
But as I wrote in the issue https://github.com/yiisoft/yii2/issues/10869#issuecomment-184856367 this is already possible with the existing API and I prefer to not add further methods if they add further abstraction.

nsanden:
seems this would do what i need

@cebe - can you reconsider merging this in light of the fact that your comment in #10869 only works for basic queries, not joins?

nsanden:
Workaround for this - use innerjoin and leftjoin instead of innerjoinwith and leftjoinwith. its not DRY but it works - mostly

Issue #10877, opened 2016-02-17
FR: Create multiple model from post request


mdmunir:
Acording this doc http://www.yiiframework.com/doc-2.0/guide-input-tabular-input.html#creating-a-dynamic-set-of-new-records prefer we add method to encapsulation that thing.

``` php
public function actionCreate()
{
    $settings = Setting::createMultiple(Yii::$app->request->post());
    foreach($settings as $setting){
        ...
        $setting->save();
    }
}
```

Assume in the view we have javascript that can add or delete lines dynamically. We can use this to create, update, delete multiple.

``` php
public function actionCreate()
{
    $oldSettings = Setting::find()->indexBy('id')->all();
    $newSettings = Setting::createMultiple(Yii::$app->request->post(), null, $oldSettings);

    // delete unused old data
    foreach($oldSettings as $setting){
        ...
        $setting->delete();
    }

    // create new or update 
    foreach($newSettings as $setting){
        ...
        $setting->save();
    }
}
```

I can submit PR for this.

arisk:
You could use `deleteAll` instead of deleting in a loop. But in any case I don't see why you would delete and recreate most, if not all the settings. Imagine the case where you have 100 settings but only change 1 one of them (not uncommon).

mdmunir:
@arisk 
Thats it that i want to propose. Its only delete unused old data. If we have 100 record and just one that not posting again, we just delete one record.

samdark:
Yes, that's a good thing to fix but it should not get too complicated.

mdmunir:
It should be more simple than `Model::loadMultiple()`. The problem is, how to check the two model are equal. I mean, the one are model but the other are array.

samdark:
Probably by getting an array from model via `$model->attributes`?

mdmunir:
@samdark 
I can submit PR for this. You can review it.

samdark:
OK. Please do.

mdmunir:
#11271

SilverFire:
Duplicates https://github.com/yiisoft/yii2/issues/10299

Issue #10299, opened 2015-12-01
Collecting tabular input guide - fix example actionUpdate


SilverFire:
In the guide article [Collecting Tabular Input](http://www.yiiframework.com/doc-2.0/guide-input-tabular-input.html) there is [an example `actionUpdate`](http://www.yiiframework.com/doc-2.0/guide-input-tabular-input.html#updating-a-fixed-set-of-records).

I don't like this line of code:
```php
        $settings = Setting::find()->indexBy('id')->all();
```

This line load all rows from the table for each update, even when only 1 row is intended to change.

Instead of this expensive operation, we should filter rows by ID, that are mentioned in the POST request. Need to redo this example and make it thrifty.

samdark:
Agree. That should be fixed.

cebe:
I think the example should focus on the point of how to do it and not on how to be optimized. code should be easy to understand. We can add a note about the fact that it is not efficient but I'd leave the example as it is.

samdark:
It would still be copy-pasted :)

cebe:
that is not our problem, if someone copy pastes it, without checking related comments...

SilverFire:
Joke time. Let's write examples like
```php
Yii::$app->db->createCommand('SELECT * FROM user WHERE id = ' . $_GET['id'])->execute();
```

I agree with @samdark, we tend to copy-paste examples from guides, because we usually trust them. This example is only about the example action for handling tabular inputs, it's not about SQL queries or anything else, so this LOC is a part of point of focus in this case. That's why I think it should be changed to something wiser.

samdark:
Let's try implementing it properly. If it will be too big piece for a guide, then we can put a note there. If not — we'll include it into the guide. OK?

cebe:
okay

Aradiv:
How about just the following ?

```php
$affected_ids = ArrayHelper::getColumn(Yii::$app->request->post(), 'id', false);
$settings = Setting::find()->where(['id' =>$affected_ids])->indexBy('id')->all();
```

this will only load the rows present in the post request.

PR #10886, opened 2016-02-17
Create multiple model from data array supplied by end user

mdmunir:
#10877
## Usage

``` php
public function actionCreate()
{
    if(($settings=Setting::createMultiple(Yii::$app->request->post())) !== false){
        foreach($settings as $setting){
            ...
        }
    }
    return $this->render('create',['settings' => $settings]);
}

public function actionUpdate()
{
    $olds = Setting::find()->orderBy('id')->all();
    if(($settings=Setting::createMultiple(Yii::$app->request->post(),null,$olds)) !== false){
        foreach($settings as $setting){
            ...
        }
        ....
        return $this->render('update',['settings' => $settings]);
    } else{
        return $this->render('update',['settings' => $olds]);
    }
}
```

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -876,6 +876,84 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
     }
 
     /**
+     * Populates a set of models with the data from end user.
+     * This method is mainly used to collect tabular data input.
+     * The data to be loaded for each model is `$data[formName][index]`, where `formName`
+     * refers to the sort name of model class, and `index` the index of the model in the `$data` array.
+     * If `$formName` is empty, `$data[index]` will be used to populate each model.
+     * The data being populated to each model is subject to the safety check by [[setAttributes()]].
+     * @param array $data the data array. This is usually `$_POST` or `$_GET`, but can also be any valid array
+     * supplied by end user.
+     * @param string $formName the form name to be used for loading the data into the models.
+     * If not set, it will use the sort name of called class.
+     * @param static[] $origin original models to be populated. It will be check using `$keys` with supplied data.
+     * If same then will be used for result model.
+     * @param string|array $keys list of attribute to check is two model are equals. If `$keys` is `null`
+     * then it will use array index to check. If `$keys` is empty then always create new model.
+     * @param string $scenario scenario for model.
+     * @return boolean|static[] whether at least one of the models is successfully populated.
+     */
+    public static function createMultiple($data, $formName = null, &$origin = [], $keys = null, $scenario = null)
+    {
+        if ($formName === null) {
+            $reflector = new ReflectionClass(get_called_class());
+            $formName = $reflector->getShortName();
+        }
+        if ($formName != '') {
+            $data = isset($data[$formName]) ? $data[$formName] : null;
+        }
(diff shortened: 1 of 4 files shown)


mdmunir:
Use for `create`, `update`, `delete` in one place.

* If posted data not same with old data then create new.
* If posted data same with old data then update.
* If old data not same with posted data then delete.

It can be use for header detail schema.
```php
$oldItems = $order->items;

$items = Item::createMultiple($post, null, $oldItems);

foreach($oldItems as $item){
    ...
    $item->delete();
}
foreach($items as $item){
    ...
    $model->link('items', $item);
}

```

mdmunir:
related PR #11271

koxu1996:
@mdmunir thanks for awesome work! Do you reccomend use this code or your tabularInput widget?
@cebe, @SilverFire `createMultiple()` is neccessary for tabular input CRUD in one page. Why not merge this crucial PR in 2.0.10 milestone?

framework/base/Model.php
```diff
@@ -876,6 +876,93 @@ public static function loadMultiple($models, $data, $formName = null)
...
+        $args = isset($options['arguments']) ? $options['arguments'] : [];
+        if ($formName === null) {
+            /* @var $model static */
+            $model = empty($args) ? new static() : $reflector->newInstanceArgs($args);
+            $formName = $model->formName();
+        }
+        if ($formName != '') {
+            $data = isset($data[$formName]) ? $data[$formName] : null;
+        }
+        if ($data === null) {
```
koxu1996: I think createMultiple instead of false should return empty array or array with one new model, depending it is request or first page load. What about slight change like this? 
```
$oldData = $data;
if ($data === null) {
    if ($oldData === []) {
        return [$model]; 
	} else {
		return [];
	}
}
```
This will allow simplification controller:
```
public function actionCreate()
{
    $count = count(Yii::$app->request->post('Setting', []));
    $settings = [new Setting()];
    for($i = 1; $i < $count; $i++) {
        $settings[] = new Setting();
    }

    // ...
}
```
to this one:
```
public function actionCreate()
{
    $settings = Setting::createMultiple(Yii::$app->request->post());

    // ...
}
```

framework/base/Model.php
```diff
@@ -876,6 +876,93 @@ public static function loadMultiple($models, $data, $formName = null)
...
+            return false;
+        }
+
```
koxu1996: @mdmunir I think createMultiple instead of false should return empty array or array with one new model, depending it is request or first page load. What about slight change like this? 
```
$oldData = $data;
if ($data === null) {
    if ($oldData === []) {
        return [$model]; 
	} else {
		return [];
	}
}
```
This will allow simplification controller:
```
public function actionCreate()
{
    $count = count(Yii::$app->request->post('Setting', []));
    $settings = [new Setting()];
    for($i = 1; $i < $count; $i++) {
        $settings[] = new Setting();
    }

    // ...
}
```
to this one:
```
public function actionCreate()
{
    $settings = Setting::createMultiple(Yii::$app->request->post());

    // ...
}
```

SilverFire:
Should solve #10299

samdark:
Won't make it into 2.0.

Issue #10789, opened 2016-02-08
Guide Test Fixtures - broken link


duynguyen0511:
Go to http://www.yiiframework.com/doc-2.0/guide-test-fixtures.html#fixtures-format 

the "documentation" hyperlink is broken. 

![2016-02-08_1352](https://cloud.githubusercontent.com/assets/436950/12898709/446f4a86-ce6b-11e5-8104-b17179a90c46.png)

samdark:
That document refers to itself. It was 2 separate documents before. These are to be merged.

lukBarros:
Samdark believe he's speaking What OS 2 links are broken , lead paragraph 404 um , like myself checked .

WHY NOT exists This file EVEN IF HE AND linked ?

lukBarros:
Look, in the current master version this file exists: https://github.com/yiisoft/yii2/blob/master/docs/guide/test-fixtures.md

Who problem? Named erro!

test-fixtures.md e test-fixture.md

lukBarros:
See #10890 PU

SilverFire:
PR is merged https://github.com/yiisoft/yii2/commit/9b9d198f2482b0b47115762837e19d33fb994402
Thank you!

lukBarros:
@duynguyen0511 done! @SilverFire  lets rock!

PR #10890, opened 2016-02-18
Change link broken by outher file of doc

lukBarros:
Related in #10789 .... real fail in link url, broken, result in 404 page...

This Changeset resolve it!


SilverFire:
Thank you

PR #10913, opened 2016-02-19
AccessRule should do more than just string comparison to match IPs

cebe:
See for example on IPv6, the same address may be written in different formats:
- `2a01:04f8:0120:7202:0000:0000:0000:0002`
- `2a01:04f8:120:7202:0:0:0:2`
- `2a01:4f8:120:7202::2`

See also http://v6decode.com/#address=2a01%3A4f8%3A120%3A7202%3A%3A2

This only adds a failing test, no implementation, so do not merge. If someone wants to work on this, feel free to do so.

### DIFF
--- a/tests/framework/filters/AccessRuleTest.php
+++ b/tests/framework/filters/AccessRuleTest.php
@@ -202,4 +202,29 @@ class AccessRuleTest extends \yiiunit\TestCase
         $this->assertNull($rule->allows($action, $user, $request));
     }
 
+    public function testMatchIPExpandIpv6()
+    {
+        /** @var $action Action */
+        /** @var $user User */
+        /** @var $request Request */
+        list($action, $user, $request) = $this->mockObjects();
+
+        $rule = new AccessRule();
+
+        // match, IPv6
+        $_SERVER['REMOTE_ADDR'] = '2a01:04f8:0120:7202:0000:0000:0000:0002';
+        $rule->ips = ['2a01:4f8:120:7202::2'];
+        $rule->allow = true;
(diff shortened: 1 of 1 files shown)


SilverFire:
We can use IpValidator in AccessRule

Vovan-VE:
It's might be usefull to implement IP range configuration like `1.2.3.10-1.2.3.20` (`1.2.10.200-1.2.20.199`) which cannot be covered by single `IP/mask`. IpValidator currently does not support this.

cebe:
> It's might be usefull to implement IP range configuration like 1.2.3.10-1.2.3.20 (1.2.10.200-1.2.20.199) which cannot be covered by single IP/mask.

is there a standard for this?

Vovan-VE:
> is there a standard for this?

In case of IpValidator itself it not so usefull, but in terms of AccessRule it is. My current use case is a controller to handle callbacks from the remote third-party service. The only way to check if a request is trusted is to check if IP is within specific range (which is given from the service documentation). That range cannot be covered by single IP/mask. So, I need to pass to AccessRule expanded list of Nteen neighbor IPs.

SilverFire:
1. It is not a standard way of IP ranges specifying
2. It can be easily done outside of IpValidator with a simple function (for example [link](http://tutorialspots.com/php-convert-ip-range-to-cidr-353.html))

IP addresses range can be easily converted to CIDR notation, but I would not like to complicate IpValidator with this feature, at least until much more people request it.

samdark:
Won't make it into 2.0.

PR #10920, opened 2016-02-20
Update phpDoc MailerInterface::compose()

githubjeka:


klimov-paul:
PhpDoc for `BaseMailer::compose` should be update as well in the same way.

githubjeka:
Done

Issue #9903, opened 2015-10-13
Migration. Alter table, set default and get syntax error. postgresql


johnluxor:
Hello.
I have migration. Yii 2.0.6. Postgres 9.4
```
public function up()
{
        $this->alterColumn('client', 'banned', $this->smallInteger(1)->defaultValue(0));
}
```

And got error:

> alter column banned in table client to smallint(1) DEFAULT 0 ...Exception 'yii\db\Exception' with message 'SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "DEFAULT"
LINE 1: ...TABLE "client" ALTER COLUMN "banned" TYPE smallint DEFAULT 0
                                                              ^
The SQL being executed was: ALTER TABLE "client" ALTER COLUMN "banned" TYPE smallint DEFAULT 0'

```
in /Users/yevgeniy/sites/nefm/vendor/yiisoft/yii2/db/Schema.php:628

Error Info:
Array
(
    [0] => 42601
    [1] => 7
    [2] => ERROR:  syntax error at or near "DEFAULT"
LINE 1: ...TABLE "client" ALTER COLUMN "banned" TYPE smallint DEFAULT 0
                                                              ^
)

Stack trace:
#0 /Users/yevgeniy/sites/nefm/vendor/yiisoft/yii2/db/Command.php(791): yii\db\Schema->convertException(Object(PDOException), 'ALTER TABLE "cl...')
#1 /Users/yevgeniy/sites/nefm/vendor/yiisoft/yii2/db/Migration.php(356): yii\db\Command->execute()
#2 /Users/yevgeniy/sites/nefm/console/migrations/m151013_101357_setDefaultBannedValueForUser.php(10): yii\db\Migration->alterColumn('client', 'banned', Object(yii\db\ColumnSchemaBuilder))
#3 /Users/yevgeniy/sites/nefm/vendor/yiisoft/yii2/console/controllers/BaseMigrateController.php(492): m151013_101357_setDefaultBannedValueForUser->up()
#4 /Users/yevgeniy/sites/nefm/vendor/yiisoft/yii2/console/controllers/BaseMigrateController.php(129): yii\console\controllers\BaseMigrateController->migrateUp('m151013_101357_...')
#5 [internal function]: yii\console\controllers\BaseMigrateController->actionUp(0)
#6 /Users/yevgeniy/sites/nefm/vendor/yiisoft/yii2/base/InlineAction.php(55): call_user_func_array(Array, Array)
[...]

PR #10921, opened 2016-02-20
Brings migrations up to parity

usecomposable:
## Schema Builder Bonanza

MySQL, Cubrid and Oracle support AFTER, FIRST and UNSIGNED. SQLite supports UNSIGNED but not AFTER or FIRST. MSSQL and PostgreSQL support neither. 

✔ New: Unsigned primary keys: `primaryKey()->unsigned()` and `bigPrimaryKey()->unsigned()`
✔ New: Constraint methods: `first()` and `after()`
✔ New: Added `char` data type to framework
✔ New: Pattern template in `ColumnSchemaBuilder::__toString()`
✔ Enh: Reorganized `QueryBuilderTest::columnTypes()` data structures
### Pt. 1

Need to verify on DBMS:
- [x] MySQL
- [x] PostgreSQL
- [x] SQLite
### Pt. 2
- [x] Tests for `after()`, `first()` and `unsigned()`
### Pt. 3 _complete_

Need to verify on DBMS:
- [x] MySQL
- [x] PostgreSQL
- [x] SQLite


samdark:
That's quite big set of changes... Could you please do separate pull requests for each fix so it's easier to review?

df2:
That's not a problem Alex. The reason it has to be as one is the dependency issues that would have to be resolved sequentially. There really isn't much rocket science going on... the bulk of the changes are from the now human-readable test case data structures. Additional tests will follow the `char()` tests that are already in.

If someone could give an SP on MSSQL that would be very much appreciated. I'll confirm the other 5.

framework/db/ColumnSchemaBuilder.php
```diff
@@ -21,6 +21,16 @@
 class ColumnSchemaBuilder extends Object
 {
     /**
+     * The following are the categories of column data types.
+     * @since 2.0.8
+     */
+    const CAT_PK = 'pk';
```
samdark: Better to name it `CATEGORY` instead of `CAT`.

samdark: PhpDoc above will stick to `CAT_PK` in both IDEs and API docs. Regular `//` comment should be used.

samdark: Also it's good to leave a not that these are internal framework data type representations.

framework/db/mysql/ColumnSchemaBuilder.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        return $this->isUnsigned ? ' UNSIGNED' : '';
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function buildAfterString()
+    {
+        return $this->after !== null ? " AFTER ('{$this->after}')" : '';
```
samdark: Should `$this->after` be escaped somehow?

framework/db/mysql/Schema.php
```diff
@@ -52,6 +52,9 @@ class Schema extends \yii\db\Schema
         'enum' => self::TYPE_STRING,
     ];
 
+    /**
+     * @var array mapping from type groups to
```
samdark: What this comment is for?

framework/db/oci/ColumnSchemaBuilder.php
```diff
@@ -20,14 +20,42 @@ class ColumnSchemaBuilder extends AbstractColumnSchemaBuilder
...
+    {
+        return $this->isUnsigned ? ' UNSIGNED' : '';
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function buildAfterString()
+    {
+        return $this->after !== null ? " AFTER ('{$this->after}')" : '';
```
samdark: Should `$this->after` be escaped?

samdark:
Indeed, this pull request isn't that big. Reviewed.

samdark:
@yiisoft/core-developers I'd like this one in. Need additional review.

framework/db/oci/ColumnSchemaBuilder.php
df2: I wondered. What about this... except how could it be written so it doesn't depend on fixed "db" component? Thanks for the swift review btw :+1: 
```php
return ($this->after !== null) ? 
		' AFTER (' . Yii::$app->db->quoteValue($this->after) . ')' : 
		'';

File: framework/db/mssql/QueryBuilder.php [hunk 5]
```diff
@@ -24,8 +24,10 @@ class QueryBuilder extends \yii\db\QueryBuilder
     public $typeMap = [
         Schema::TYPE_PK => 'int IDENTITY PRIMARY KEY',
         Schema::TYPE_BIGPK => 'bigint IDENTITY PRIMARY KEY',
-        Schema::TYPE_STRING => 'varchar(255)',
-        Schema::TYPE_TEXT => 'text',
+        Schema::TYPE_UNSIGNEDPK => 'int IDENTITY PRIMARY KEY',
+        Schema::TYPE_CHAR => 'nchar(1)',
+        Schema::TYPE_STRING => 'nvarchar(255)',
+        Schema::TYPE_TEXT => 'ntext',
```
df2: Please note changes to `nvarchar` and `ntext` are contentious. Test data will now be invalid for MSSQL.

framework/db/oci/ColumnSchemaBuilder.php
samdark: Should be `quoteColumnName()`, not `quoteValue()`.

framework/db/mssql/QueryBuilder.php
samdark: Do you have any idea on how to fix that?

samdark:
Yes, that's correct way of passing `$db`. A note should be added to `UPGRADE` about possible breaks caused by signature change.

framework/db/ColumnSchemaBuilder.php
```diff
@@ -223,12 +299,65 @@ protected function buildCheckString()
...
+    {
+        $placeholderValues = [
+            '{type}' => $this->type,
+            '{length}' => $this->buildLengthString(),
+            '{unsigned}' => $this->buildUnsignedString(),
+            '{notnull}' => $this->buildNotNullString(),
+            '{unique}' => $this->buildUniqueString(),
+            '{default}' => $this->buildDefaultString(),
+            '{check}' => $this->buildCheckString(),
+            '{pos}' => ($this->isFirst) ?
```
samdark: Does that mean that either after or first would be always added to the query even if no `first()` or `after()` present in column definition?

framework/db/mysql/QueryBuilder.php
```diff
@@ -25,6 +25,8 @@ class QueryBuilder extends \yii\db\QueryBuilder
     public $typeMap = [
         Schema::TYPE_PK => 'int(11) NOT NULL AUTO_INCREMENT PRIMARY KEY',
         Schema::TYPE_BIGPK => 'bigint(20) NOT NULL AUTO_INCREMENT PRIMARY KEY',
+        Schema::TYPE_UNSIGNEDPK => 'int(11) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY',
+        Schema::TYPE_CHAR => 'char(1)',
```
samdark: Why default length is 1?

framework/db/mssql/QueryBuilder.php
samdark: Also it should be mentioned about how to approach upgrade with such change considering existing project with live DB and migrations already present.

framework/db/ColumnSchemaBuilder.php
df2: Nope, in that case, if `{pos}` was present in the template it would be replaced by the the result of `buildAfterString()`, which will be an empty string.

framework/db/mysql/QueryBuilder.php
df2: Hired a committee at first but then went with the day of the month I was born on ;)

Seriously though, it makes sense as the shortest length that saves space compared to varchar.

samdark: OK.

cebe: default value should be equal to the value the DBMS uses if you do not specify a length. This is because yii always uses a value if you do not specify one.

df2: Goes to 1 with MySQL variant MariaDB.

df2:
When can this be ready to get in to mainline?

df2:
We could complete tests as a separate issue.

df2:
The main thing is the repetition on columns one and two... that doesn't make any sense really.

framework/db/ColumnSchemaBuilder.php
```diff
@@ -51,19 +60,61 @@ class ColumnSchemaBuilder extends Object
...
     /**
      * Create a column schema builder instance giving the type and value precision.
      *
      * @param string $type type of the column. See [[$type]].
      * @param integer|string|array $length length or precision of the column. See [[$length]].
+     * @param \yii\db\Connection $db the current database connection. See [[$db]].
      * @param array $config name-value pairs that will be used to initialize the object properties
      */
-    public function __construct($type, $length = null, $config = [])
+    public function __construct($type, $length = null, $db = null, $config = [])
```
SilverFire: Method signature change must be mentioned in UPGRADE.md

SilverFire:
This PR will probably fix #9903, @df2 right?

framework/db/ColumnSchemaBuilder.php
samdark: Yes. Absolutely.

samdark:
> We could complete tests as a separate issue.
I don't like the idea.

df2:
@SilverFire will have to look in to that particular case.

@samdark then it should be a main array and a DBMS specific array. The main array as:

```php
[
    Schema::Type . x,
    $this->type()->x,
    [
        'mysql' => 'type x',
        'pgsql' => 'type "x"',
        ...
    ]
],
...
```
There are reasons that there need to be DBMS specific tests too, so these will get merged (added) and an array in the old format can be returned for each DBMS.

cebe:
Sorry to say that but to me this PR looks like overcomplicating things a lot.
What are the categories used for?
Also introduction of `TYPE_UNSIGNEDPK` is inconsistent, what about `TYPE_UNSIGNEDBIGPK`? As these make sense in MySQL I do not think they should be added for dbms like pgsql. Current approach to PK was a bit wrong and gets more complex with these changes. I think we should go another way and simplify primary key creation and also make it more flexible by allowing other types than integer for example. Currently do not have time to write down my proposal but will try to do so soon.
basically API would allow things like this:

```php
$this->string(16)->primaryKey()->notNull(),
```

df2:
@cebe read through it again, it does not complicate things it makes the schema builder substantially smarter and cleaner.

The categories are used to decide which modifiers make sense for that field type. It's cleaner. It's like going down a street and knocking on every door vs only knocking on doors 3, 9 and 24 where you need to know if there is someone home.

I agree that there is no TYPE_UNSIGNEDBIGPK. The reason being that you can specify the size to `unsignedPrimaryKey(size_here)`. 

I also like the idea of using `primaryKey()` as a modifier. That is a breaking change for 2.1. Right now we are being pragmatic and coming up with a solution to not being able to use unsigned primary keys.

df2:
@samdark please can you do some digging in to why this commit has to be made https://github.com/df2/yii2/commit/553cca642cef7b2d415f9dd2e8035158def2b138

samdark:
Because you're referencing constants as `self` while it should be `Schema`?

samdark:
Btw., `unsigned()` PK with length looks very specific to MySQL and probably Oracle. PostgreSQL doesn't have size for PKs as well as SQLite and MSSQL.

df2:
@samdark thanks, that was it :)

df2:
Taking a few days off... will be back to finish the tests... not much left to do really.

Will need help testing Oracle, MSSQL and Cubrid. Welcome to suggestions :)

Have a great weekend :beers:

framework/UPGRADE.md
```diff
@@ -20,6 +20,10 @@ ______________________
 * The signature of `yii\helpers\BaseArrayHelper::index()` was changed. The method has got an extra optional parameter
   `$groupBy`.
 
+* The signature of `yii\db\ColumnSchemaBuilder::__construct()` was changed. The method has got an extra optional parameter `$db`.
+
+* String types in the MySQL column schema map were upgraded to unicode storage types.
```
SilverFire: UPGRADE.md must answer the question: `What must I do to upgrade from x.x.x to x.x.y`.
What must I do with this change? :)

df2:
@samdark @SilverFire please check https://github.com/df2/yii2/blob/9562-add-char-datatype/docs/schema-builder-patterns.xlsx?raw=true

That has the pattern mappings for each DBMS and column category. Will complete tests if you can confirm the mappings are all correct.

What do you want to do with Cubrid, MSSQL, Oracle?

SilverFire:
@df2 sorry for a late reply

Your table looks great! I didn't check each cell, but in general it seems to be correct.

Before 2.0.7 release me and @samdark tried to fix Oracle and MsSQL tests. Both of us are stucked.

I think we can accept OCI and MsSQL changes as is, since test are broken any way. Then, when we will come to test fixing again, we will check this too. @samdark ?

samdark:
Yes. The only thing is that I'm not sure if it's going to go in 2.0.8 or 2.0.9...

SilverFire:
@df2 could you check https://github.com/yiisoft/yii2/issues/11158, please?

df2:
Well I think this can be merged and then any new code can build on top of this. What would be the reason to delay it @samdark? The tests still need some TLC down the road, esp. in terms of the outstanding DBMSes, but that is easier now we have a clearer picture of what tests are in place currently. I cannot see any reason not to merge this; it improves the schema builder to be used in migrations and fixes a significant number of open issues.

samdark:
@df2 we have 2.0.8 planned for April 17th. I'm not sure we'll be able to test/fix enough regressions (there could be some) for that date. That's about 3.5 weeks. If you're willing to push it forward no matter what and make it 100% solid before that date, we'll, of course, merge.

df2:
Sounds good @samdark. I will do one more systematic review of the tests then and then hand it over to you. I will be on standby from there on in case any new issues arise between now and April 17th.

samdark:
OK. Ping me when you think it's ready to be merged.

df2:
OK, all good to go @samdark

samdark:
@df2 can you help closing issues which are now solved?

Issue #8639, opened 2015-06-03
Problem while injecting dependencies in ActiveRecord model


akireikin:
Given ActiveRecord class.

```php
class Place extends ActiveRecord
{
    /**
     * @var \app\core\helpers\GeoTransformerInterface
     */
    private $geoTransformer;

    /**
     * @param \app\core\helpers\GeoTransformerInterface $geoTransformer
     * @param array                                     $config
     */
    public function __construct(GeoTransformerInterface $geoTransformer, $config = [])
    {
        $this->geoTransformer = $geoTransformer;

        parent::__construct($config);
    }

    /**
     * @inheritdoc
     */
    public static function tableName()
    {
        return 'place';
    }

    /**
     * @return array
     */
    public function getLocation()
    {
        return $this->geoTransformer->decode($this->location_polygon);
    }

    /**
     * @param array $value
     */
    public function setLocation(array $value)
    {
        $this->location_polygon = $this->geoTransformer->encode($value);
    }

    /**
     * @return \yii\db\ActiveQuery
     */
    public function getCategories()
    {
        return $this->hasMany(Category::className(), ['id' => 'category_id'])->viaTable('place_has_category', ['place_id' => 'id']);
    }
}
```

# Problem 1

```php
$place = Place::find()->one();
```

This code raises an Exception.

`Argument 1 passed to app\\modules\\places\\ar\\Place::__construct() must be an instance of app\\core\\helpers\\GeoTransformerInterface, none given`

## Possible solution

Overwrite `BaseActiveRecord::instantiate()` with:

```php
    public static function instantiate($row)
    {
        return \Yii::$container->get(static::class);
    }
```

# Problem 2

```
$place = Place::find()->with('categories')->asArray()->all();
```

This code also raises an Exception.

`Argument 1 passed to app\\modules\\places\\ar\\Place::__construct() must be an instance of app\\core\\helpers\\GeoTransformerInterface,  none given`

The problem is located in `ActiveQueryTrait::findWith()`, line 162.

```php
    public function findWith($with, &$models)
    {
        $primaryModel = new $this->modelClass;
        $relations = $this->normalizeRelations($primaryModel, $with);
        /* @var $relation ActiveQuery */
        foreach ($relations as $name => $relation) {
            if ($relation->asArray === null) {
                // inherit asArray from primary query
                $relation->asArray($this->asArray);
            }
            $relation->populateRelation($name, $models);
        }
    }
```

## Possible solution

[...]

PR #10939, opened 2016-02-23
Avoid some calls to `new $this->modelClass` in ActiveQuery

cebe:
this fixes some cases of the problem described in #8639 and is fully BC.
Fixing other places may need BC breaks.

With this `inverseOf()` relation and work `viaTable()` should work.


klimov-paul:
Looks good

PR #10957, opened 2016-02-24
Set limit = 1 in one()

DrDeath72:


cebe:
this has been discussed already, see #9578 and https://github.com/yiisoft/yii2/issues/4348#issuecomment-49419664

PR #10971, opened 2016-02-26
replace all occurrences of migration class

vlad-khramov:
Class was replaced only in definition of migration class.
In testGenerateDefaultMigration classname used in error message (https://github.com/yiisoft/yii2/blob/31cf978f385b1ffb16aaf3ff04175d5b65922948/tests/framework/console/controllers/MigrateControllerTestTrait.php#L188), so sometimes this test failed: https://travis-ci.org/quantum13/yii2/jobs/111687879


SilverFire:
Thank you!

PR #10989, opened 2016-03-01
Phpdoc read/write only

githubjeka:
The proposal to enhancement **all** `"This property is read-only."` and `"This property is write-only"` to [`@property-read`](https://www.phpdoc.org/docs/latest/references/phpdoc/tags/property-read.html) and [`@property-write`](https://www.phpdoc.org/docs/latest/references/phpdoc/tags/property-write.html) respectively.

image


SilverFire:
Looks good to me

githubjeka:
Do I need to retain the text "This property is read-only." for [doc-api-site](http://www.yiiframework.com/doc-2.0/yii-base-application.html#$authManager-detail) or your change it by tags `@property-write`?

klimov-paul:
This will work for PhpStorm, but what about other IDE, like NetBeans or Eclipse?

githubjeka:
Need feedback. I haven't it

cebe:
We had this already. The reason we changed it was missing IDE support. So we need to find out if situation has changed.

githubjeka:
To feedback:
```php
<?php
/**
 *  @property-read string $id
 */
class FakeClass {

}

$class = new FakeClass();
$class->id = 123;
```

on phpStorm


### People please check this by  Eclipse

NetBeans <= 8.0.2 doesn't :sob:

samdark:
It's a no go then. NetBeans is half of the userbase...

githubjeka:
@samdark need to find out whether the autocompleate works at the set value @property-read.

youmad:
Autocomplete works on NetBeans IDE 8.1
2016-03-01-153540_1366x768_scrot

cebe:
FYI: if we change it, change has to be made here and tags will be auto generated afterwards: https://github.com/yiisoft/yii2/blob/master/build/controllers/PhpDocController.php#L587

dyar74:
Autocomplete  on NetBeans IDE 8.0.2 works too.

Faryshta:
works on atom too.

cebe:
lets keep this for 2.0.x and reconsider for 2.1

Dzhuneyt (approved):

Dzhuneyt (approved):
`@property-read` works fine on the latest PHPStorm version - 2018.3.2.

image

githubjeka:
Why closed?

samdark:
@githubjeka:

1. It was initially planned to have it in 2.1 but since 3.0 won't have properties it made no sense anymore.
2. Document fixer should be adjusted in order to generate these properties else it will be quite hard to release anything. See https://github.com/yiisoft/yii2/blob/master/build/controllers/PhpDocController.php

samdark:
@githubjeka do you plan to update https://github.com/yiisoft/yii2/blob/master/build/controllers/PhpDocController.php? Current PR can not be merged because it will be immediately overwritten by generator.

samdark:
Docs are going to be generated with build tool.

PR #10990, opened 2016-03-01
Use theme applyTo in View findViewFile

jacmoe:
In View::findViewFile use `$this->theme->applyTo($path)` before checking if $path is a file.

If this is not done, then Yii will find the view file in the active module's view path instead of looking in the theme path map. This is a problem only when using a view renderer with a different file extension than php. Like .twig and .jade.

Without this small fix, we have to drop dummy empty files in the module's view directories :-1:


samdark:
Is it possible to cover it with a test?

jacmoe:
I will look into creating a test for it.
This is only an issue when the default extension is different from php, and when using modules.

jacmoe:
I am going to update the PR with a slightly different approach where the
theme apply is only used when the default view extension is not php and the
view is the same otherwise. I think my fix now will mean that Yii would
accept a view that does not exist in the module with the php extension..

jacmoe:
I will add a test in yii2-twig first - as that is probably going to be easier to do. :)

I have almost zero experience with writing tests..

Perhaps I will manage to add one for this PR, let's see.

jacmoe:
Something went wrong - so I am going to open up a new PR later :)

With a test.

jacmoe:
I don't need this bug to be fixed anymore because I went with overriding the View component:

```php
class View extends \yii\web\View {

    /**
     * Override to always pass the current context to render
     * https://github.com/yiisoft/yii2/issues/4382
	 * @inheritDoc
	 */
    public function render($view, $params = array(), $context = null)
    {
        if ($context === null) {
            $context = $this->context;
        }

        return parent::render($view, $params, $context);
    }

    /**
     * If the view is not found by normal means
     * then use the theme pathmap to find it.
	 * @inheritDoc
	 */
    protected function findViewFile($view, $context = null)
    {
        $path = $view . '.' . $this->defaultExtension;
        if ($this->theme !== null) {
            $path = $this->theme->applyTo($path);
        }
        $viewfile = parent::findViewFile($path, $context);

        return $viewfile;
    }

}

```

cebe:
okay, but is there a bug in the framework or not? If there is, we should have an issue to fix it later.

jacmoe:
Yes, there is a bug - but it will only affect users who use a view renderer
with a different view file extension than the default (php) and where the
defaultExtension option has been set.

And perhaps only where a pathmap has been set in the view component's theme
setting.

Anyway, I will open a ticket for it later as I would like to implement a
test case for it :)

cebe:
okay, thanks.

PR #10999, opened 2016-03-01
Fix protected/private calls

DrDeath72:


SilverFire:
It does not look acceptable to me. Can you elaborate the idea of these changes?

DrDeath72:
method_exists return true on private/protected methods and after call it got error

DrDeath72:
Or maybe i wrong because __call allways present and is_callable allways give true so maybe need 
```
    public static function getObjectMethods($object)
    {
        return get_class_methods($object);
    }
```
And replace method_exists with it

DrDeath72:
or replace getter/setter method_exist check with
```
public function isCallable($object, $name) 
{
   return in_array($name, get_class_methods($object));
}
```

samdark:
Doesn't look good to me. First of all, it degrades performance for everything. Second, I don't see a point. Could you elaborate on why it's needed?

DrDeath72:
maybe you right it's don't need for getters/setters just don't make it not public, but 
```php
    public function hasMethod($name)
    {
        return method_exists($this, $name);
    }
```
return true for protected methods and can't call it
```php
    public function __call($name, $params)
    {
        $this->ensureBehaviors();
        foreach ($this->_behaviors as $object) {
            if ($object->hasMethod($name)) {
                return call_user_func_array([$object, $name], $params);
            }
        }
        throw new UnknownMethodException('Calling unknown method: ' . get_class($this) . "::$name()");
    }
```

DrDeath72:
And canGet/Set return true but they can'tGet/Set on real :D

klimov-paul:
Indeed magic setters, getters and `__call` do not aware of corresponding method to be protected or private.
But I don't see much harm in it. Why it is so crucil for you?
Currently it is just the same as for the normal object:

```php
class Foo
{
    protected function some() {...}
}

$obj = new Foo();
$obj->some(); // no one prevents you from attempt to call protected method
// however it will end up with PHP error
```

Of course we can improve methods detection to filter private and protected methods, but it will cause significant performance impact.

Anyway such refactoring should be performed via reflection library not `get_class_methods()` calls:

```php
$reflectionClass = new ReflectionClass($this); 
if ($reflectionClass->hasMethod($method)) {
    return $reflectionClass->getMethod($method)->isPublic();
}
return false;
```

DrDeath72:
when you call by hand some protected methods it's your call, when you call in behavior, 1 has protected. 2 has same public, yii try get 1 protected. maybe it's all waste of time, but if yii make checks and decide doing something after it, it must be without errors :D

samdark:
@DrDeath72 can you measure performance impact?

DrDeath72:
nope :(

SilverFire:
Fully agreed with @klimov-paul

PR #11005, opened 2016-03-02
	Add behavior "global switch"

DrDeath72:
If $global = false we can access behavior only by $model->getBehavior('name')->method()

### DIFF
--- a/framework/base/Behavior.php
+++ b/framework/base/Behavior.php
@@ -25,7 +25,11 @@ class Behavior extends Object
      */
     public $owner;
 
-
+    /**
+     * @var bool Allow access to behavior methods and properties from owner
+     */
+    public $global = true;
+    
     /**
      * Declares event handlers for the [[owner]]'s events.
      *
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -132,7 +132,7 @@ class Component extends Object
             // behavior property
             $this->ensureBehaviors();
             foreach ($this->_behaviors as $behavior) {
-                if ($behavior->canGetProperty($name)) {
+                if ($behavior->global && $behavior->canGetProperty($name)) {
                     return $behavior->$name;
                 }
             }
@@ -184,7 +184,7 @@ class Component extends Object
             // behavior property
             $this->ensureBehaviors();
             foreach ($this->_behaviors as $behavior) {
-                if ($behavior->canSetProperty($name)) {
+                if ($behavior->global && $behavior->canSetProperty($name)) {
                     $behavior->$name = $value;
 
                     return;
@@ -221,7 +221,7 @@ class Component extends Object
             // behavior property
             $this->ensureBehaviors();
             foreach ($this->_behaviors as $behavior) {
-                if ($behavior->canGetProperty($name)) {
+                if ($behavior->global && $behavior->canGetProperty($name)) {
                     return $behavior->$name !== null;
                 }
             }
@@ -252,7 +252,7 @@ class Component extends Object
             // behavior property
             $this->ensureBehaviors();
             foreach ($this->_behaviors as $behavior) {
-                if ($behavior->canSetProperty($name)) {
+                if ($behavior->global && $behavior->canSetProperty($name)) {
                     $behavior->$name = null;
                     return;
                 }
@@ -278,7 +278,7 @@ class Component extends Object
     {
         $this->ensureBehaviors();
         foreach ($this->_behaviors as $object) {
-            if ($object->hasMethod($name)) {
+            if ($object->global && $object->hasMethod($name)) {
                 return call_user_func_array([$object, $name], $params);
             }
         }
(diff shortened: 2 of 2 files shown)


samdark:
Why?

DrDeath72:
because every behavior spam in owner properties and method even if it don't need, if one behavior attach twice it's useless too and don't workable...

samdark:
OK. Got it. I think it could be named better such as `$exposeMethods`. @yiisoft/core-developers thoughts?

DrDeath72:
> $exposeMethods

properties too :)

SilverFire:
We are going to get rid of behaviors in favor of traits in 2.1. Setting `global` (`exposeMethods`, etc) to `true` forces to use `$model->getBehavior('name')->method()`, which is not compatible with the aforementioned idea.

I don't see use cases, that become simpler with this PR

DrDeath72:
no useless spam to owner

klimov-paul:
> because every behavior spam in owner properties and method even if it don't need

Since when it becomes a problem?
Please, provide a particular use case, which indicates this actually worth fixing.

We use 'convension over configuration' principle. You are asking of overcomplecating functionality without practicle purpose. Just because 'it seems wrong to you'. This is not an argument.

What will come next? You will demand that 'setter/getter' should actually introduce magic propreties only if they were explicilty configured?

DrDeath72:
> Since when it becomes a problem?

ok, i add my behavior, than yours, than 3, 4, 5. My have properties|methods needed only for internal work or just setup, yours or 3, 4, 5 have it with same names to. when i call it, it return my first useless one

so we must look for all names conflicts instead just disable global behavior and use it all like relations. or create unique long names...

for my behavior to hide it from "global" scope i just code this sux and use only protected _name vars/methods... and can access it only by $model->behavior->method, but it can be done by one behavior "global" param without this sux...

```php
<?
namespace app\components\base;

use ReflectionClass;

class Behavior extends \yii\base\Behavior {

	public $params = [];

    public function __get($sName) {
        $oReflectionClass = new ReflectionClass($this);
        if($oReflectionClass->hasProperty('_' . $sName) and $oReflectionClass->getProperty('_' . $sName)->isProtected()) {
            return $this->{'_' . $sName};
        }
        return parent::__get($sName);
    }

    public function __set($sName, $sValue) {
        $oReflectionClass = new ReflectionClass($this);
        if($oReflectionClass->hasProperty('_' . $sName) and $oReflectionClass->getProperty('_' . $sName)->isProtected()) {
            $this->{'_' . $sName} = $sValue;
        }
        else {
            parent::__set($sName, $sValue);
        }
    }

	public function __call($sName, $aArguments) {
        $oReflectionClass = new ReflectionClass($this);
        if($oReflectionClass->hasMethod('_' . $sName) and $oReflectionClass->getMethod('_' . $sName)->isProtected()) {
            return call_user_func_array([$this, '_' . $sName], $aArguments);
        }
        return parent::__call($sName, $aArguments);
	}

}
```

> We use 'convension over configuration' principle. You are asking of overcomplecating functionality without practicle purpose. Just because 'it seems wrong to you'. This is not an argument.

when function called canGet... return true and than can't get it, it's fail :)

> What will come next?

i will use many extended/replaced method because you don't add small flexibility params :D

klimov-paul:
> so we must look for all names conflicts instead just disable global behavior and use it all like relations. or create unique long names...

Behaviors emulates mixings (traits). Every behavior should be considered as a mixing, taking in account possible naming collision. Developer should take care of it. I do so, for example:
https://github.com/yii2tech/ar-variation/blob/master/VariationBehavior.php#L65

Here 'variation' prefix is widely used for the behavior properies and methods, which can be accessed from owner scope.

> i add my behavior, than yours, than 3, 4, 5

You can always change the order of the behaviors been attached. It is fully under your control.
You can also disable any behavior at will if make a trouble in particular program flow.

> i will use many extended/replaced method because you don't add small flexibility params

If we add 'flexibily' to every such place it will be no end to them.

klimov-paul:
Behaviors are "mixings" emulation, analog of "traits". Thus they behave as a traits.
If it is not acceptable for you, you should use another mechanism instead of them.

DrDeath72:
Ok, i think i create better solution for me...
```php
<?
namespace app\components\base;

class IsolatedBehavior extends Behavior {

    public function canSetProperty($name, $checkVars = true) {
        return method_exists(get_parent_class($this), 'set' . $name) || $checkVars && property_exists(get_parent_class($this), $name) ?: false;
    }

    public function canGetProperty($name, $checkVars = true) {
        return method_exists(get_parent_class($this), 'get' . $name) || $checkVars && property_exists(get_parent_class($this), $name) ?: false;
    }

    public function hasMethod($name) {
        return method_exists(get_parent_class($this), $name) ?: false;
    }

}
```
i'm right?

DrDeath72:
thanks all, add more extended methods...
```php
<?
namespace app\components\base;

use ReflectionClass;

class Behavior extends \yii\base\Behavior {

	public $params = [];
    public $isolated = false;

    public function canSetProperty($name, $checkVars = true) {
        return $this->isolated ? false : parent::canSetProperty($name, $checkVars);
    }

    public function canGetProperty($name, $checkVars = true) {
        return $this->isolated ? false : parent::canGetProperty($name, $checkVars);
    }

    public function hasMethod($name) {
        return $this->isolated ? false : parent::hasMethod($name);
    }

}
```

Issue #7498, opened 2015-02-28
Add REPL support


callmez:
is there possible to extension psysh as  feature or offical extension。its really useful for developers。
like Rails Console or Laravel tinker.  thanks

samdark:
Could you elaborate? I'm not using any of the tools mentioned...

samdark:
Ah, interactive shell.

samdark:
What Yii specific features do you want?

callmez:
yes，for example i can use it test some function from command line without code file ， watch the data changed and more

cebe:
@callmez can you give some links that describe how these work in other frameworks?

callmez:
http://blog.enge.me/post/tinkering-with-tinker-like-an-artisan 
is this ok?

callmez:
and you can get the psysh repository from https://github.com/bobthecow/psysh

bobthecow:
I think this is a great idea :)

For a few examples of projects that already do this:

 * [artisan tinker for Laravel](https://github.com/laravel/framework/blob/5.0/src/Illuminate/Foundation/Console/TinkerCommand.php)
 * [eZ Debug Shell for eZ Publish](https://github.com/lolautruche/ezsh)
 * [drush-psysh for Drupal](https://github.com/grota/drush-psysh)
 * [magerun console command for Magento](https://github.com/netz98/n98-magerun/blob/develop/src/N98/Magento/Command/Developer/ConsoleCommand.php)
 * And tons for Symfony: [sf1-psysh-bootstrap](https://github.com/varas/sf1-psysh-bootstrap), [sf2-psysh-bootstrap](https://github.com/varas/sf2-psysh-bootstrap), [symfony-repl](https://github.com/luxifer/symfony-repl), [psymf](https://github.com/navitronic/psymf)

FWIW, [PsySH is working on a plugin architecture](https://github.com/bobthecow/psysh/issues/166), and it would prolly be ideal for projects to use that in addition to a shell wrapper (like Laravel's `artisan tinker` or `ezsh` or whatever). We're also interested in ways we can make creating a framework- or library-specific bespoke REPL a better experience, so if you have thoughts, let us know!

cornernote:
Seems there is a way to do this in Yii1, havent tried it yet but will do so soon and report back as to how it works with Yii2:

https://gist.github.com/mrdziuban/7c3cffc52c41f0bb7577

cornernote:
Yep, works a treat on Yii2:
[...]

PR #11008, opened 2016-03-02
WIP #7498: add php interactive shell command

pana1990:
WIP : https://github.com/yiisoft/yii2/issues/7498

Task : 
- [ ] Add phpdocs
- [ ] Add @since
- [ ]  Update changelog
- [ ] Add another casters (Suggestion?)


pana1990:
Example : 

screenshot_3

framework/console/controllers/TinkerController.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public function actionIndex()
+    {
+        $config = new Configuration;
+
+        $config->getPresenter()->addCasters(
+            $this->getCasters()
+        );
+        $config->setTabCompletion(true);
```
bobthecow: This line is unnecessary. Tab completion is enabled by default, and this line will prevent a user from disabling it locally via psysh config.

pana1990: Yeah, my bad.

pana1990: @bobthecow  done! thanks for your help :)

bobthecow: No problem!

framework/console/Application.php
```diff
@@ -184,6 +184,7 @@ public function coreCommands()
             'message' => 'yii\console\controllers\MessageController',
             'migrate' => 'yii\console\controllers\MigrateController',
             'serve' => 'yii\console\controllers\ServeController',
+            'tinker' => 'yii\console\controllers\TinkerController',
```
samdark: What a name :) That's a bit weird.

pana1990: copy/paste from laravel :) 

http://laravel-recipes.com/recipes/280/interacting-with-your-application

Suggestion?

samdark: `shell` as it was in Yii 1.1?

samdark:
Found https://github.com/svkurowski/yii2-shell while googling...

pana1990:
@samdark done! can you include this feature in 2.0.8 milestone?

cebe:
I am usure if we should add a dependency to psysh to the framework. also you now have added it to require-dev which will cause the dependency to not be installed by default.
If may be useful make an extension of this and add this to the application templates by default instead of adding a framework dependency.

pana1990:
@yiisoft/core-developers opinios?

klimov-paul:
I am agree with @cebe.
There should not be dependencies on "psysh" in the core.

We already have a lot of complains about extra dependencies to the external libraries and assets of the core. In 2.1 we are going to split the main repository into several ones in order to solve that. This change just make things worse at this aspect.

This can be released as a separated extension or should be reworked to be more independent.

pana1990:
i am agree, then, can you create repo for this feature?

samdark:
https://github.com/yiisoft/yii2-shell

pana1990:
@samdark thanks.

pana1990:
@samdark i can't fork repo...

SilverFire:
Fixed

pana1990:
i close this pr, new pr in https://github.com/yiisoft/yii2-shell/pulls

PR #11013, opened 2016-03-03
Fixed #11012:  getBaseName return null if filename is chinese string

### LINKED ISSUE #11012: yii\web\UploadedFile getBaseName return null in 'phpStudy' tookit
in [phpStudy](http://www.phpstudy.net) toolkit, use 'Apache + PHP5.5', if upload file is chinese, will return null.

hiscaler:


samdark:
#11012

framework/web/UploadedFile.php
```diff
@@ -171,7 +171,7 @@ public function saveAs($file, $deleteTempFile = true)
      */
     public function getBaseName()
     {
-        return pathinfo($this->name, PATHINFO_FILENAME);
+        return substr(pathinfo("_$this->name", PATHINFO_FILENAME), 1);
```
LAV45: please wrap it in parentheses `"_{$this->name}"`

hiscaler: Thanks

SilverFire:
It's a very specific problem and the fix is not obvious, so it's easy to crash it in future. Could you add a unit test, please?

SilverFire:
Merged: https://github.com/yiisoft/yii2/commit/75db69f90aadf2b4c2953065d3866f2f7bb4bcb0 https://github.com/yiisoft/yii2/commit/a0f9bf289353c4fc4aebc46fe263c20ec39d5976

SilverFire:
Thank you!

PR #11014, opened 2016-03-03
Fixed formatting issues in docs for ArrayHelper

MartijnHols:
The docs formatter at http://www.yiiframework.com/doc-2.0/ does not like code tags starting on the first line after plain text.

See code blocks in http://www.yiiframework.com/doc-2.0/guide-helper-array.html#reindexing-arrays


samdark:
Thanks!

PR #11015, opened 2016-03-03
Unsafe attribute event

RangelReale:
This enables behaviours to override unsafe attribute setting, allowing for example setting values in a nested model.

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -69,6 +69,10 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      * @event Event an event raised at the end of [[validate()]]
      */
     const EVENT_AFTER_VALIDATE = 'afterValidate';
+    /**
+     * @event ModelUnsafeAttributeEvent an event raised when setting and unsafe attribute
+     */
+    const EVENT_UNSAFE_ATTRIBUTE = 'unsafeAttribute';
 
     /**
      * @var array validation errors (attribute name => array of errors)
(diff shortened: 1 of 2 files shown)


samdark:
Can you elaborate about your use case?

RangelReale:
My use case is a model where I want to set nested (hasMany) model attributes at the same time as the main model.
For example, I have a Campaign Model, which have a hasMany relation with CampaignSchedule.
In my HTML both are entered on the same screen, and created dinamically via JS.
The POST from this screen is like this:

```
[Campaign => 
    [
        title => Campaign name
        schedules => [
            [ 'start_date'=> 10, 'end_date'=> 20 ],
            [ 'start_date'=> 30, 'end_date'=> 40 ],
        ]
    ]
]
```

I wanted that this POST to automatically set the 'schedules' item in some way using a behavior, but  there was no way for a behavior to receive this non-existant attribute directly from setAttributes().

I made an extension to allow this, but it depends on being able to receive the post data, I put it here:
https://gist.github.com/RangelReale/f2a47f34cd8aeabad2dd

mdmunir:
https://github.com/mdmsoft/yii2-ar-behaviors#relationtrait

RangelReale:
Hmm I did this some time ago, I will evaluate @mdmunir code to see if there was any edge case in my use case and get back to this.

samdark:
Any updates?

RangelReale:
I think maybe this approach could work, but my Yii2 project was postponed, so I think this can be closed for now.

PR #11023, opened 2016-03-04
db/oci/Schema.php implementing PDO::CASE_LOWER

CedricYii:
When you want to use Pdo lowercase in Yii, you have to set db component in config/main-local.php with:

```
<?php
return [
    'components' => [
        'db' => [
            'attributes' => [PDO::ATTR_CASE => PDO::CASE_LOWER,],
...
```

Then with this schema version, you will get all tables and columns names in lowercase in Yii.
All tables and columns MUST be left in the default upper case in the database Oracle.

With this configuration, you can now use the advance template out of the box.

### DIFF
--- a/framework/db/oci/Schema.php
+++ b/framework/db/oci/Schema.php
@@ -12,6 +12,7 @@ use yii\db\Connection;
 use yii\db\Expression;
 use yii\db\TableSchema;
 use yii\db\ColumnSchema;
+use \PDO;
 
 /**
  * Schema is the class for retrieving metadata from an Oracle database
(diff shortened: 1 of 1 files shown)


framework/db/oci/Schema.php
```diff
@@ -57,7 +58,25 @@ public function releaseSavepoint($name)
      */
     public function quoteSimpleTableName($name)
     {
-        return strpos($name, '"') !== false ? $name : '"' . $name . '"';
+        if (strpos($name, '"') !== false) {
+            return $name;
+        } elseif ($this->db->slavePdo->getAttribute(PDO::ATTR_CASE) === PDO::CASE_LOWER) {
+            $name = strtoupper($name);
```
klimov-paul: Why?
This whould NOT work, if tabled named in "CamelCase", for example: 'UserComment'.

klimov-paul:
Sorry, but we can not take this PR.
It provides solution only for single specifc use case.

framework/db/oci/Schema.php
CedricYii: It works as usual unless you define PDO::CASE_LOWER.
If you are using case-sensitive tables, you just can't use PDO::CASE_LOWER or you need to adress the table name with " as '"UserComment"' (just like in Oracle).
This is because, you can't expect the lowercase pdo behavior to work if there is a table "UserComment" and another table "USERCOMMENT".

PR #11025, opened 2016-03-04
fix name method

pana1990:


SilverFire:
Hm.. it seems that we do not have a convention on this.
Both options are OK to me.

pana1990:
which one? optionAliases or optionsAliases? both is ok for me.

SilverFire:
For me too. Why do you want to rename it then? :)

pana1990:
Currently method name is "optionAliases" (see https://github.com/yiisoft/yii2/blob/master/framework/console/Controller.php#L319). it cause error in HelpController (see https://github.com/yiisoft/yii2/blob/master/framework/console/controllers/HelpController.php#L424). 

Can you understand me?

SilverFire:
Sorry, I've overseen it.

pana1990:
thanks :)

PR #11043, opened 2016-03-07
fixes #11040: make copyDirectory method work correctly with parameter 'recursive'

### LINKED ISSUE #11040: Parameter 'recursive' not work in function FileHelper::copyDirectory
Documentation has description: 
```doc
recursive: boolean, whether the files under the subdirectories should also be copied. Defaults to true.
```
But, it not working.

pchapl:
Closes #11040

### DIFF
--- a/tests/framework/helpers/FileHelperTest.php
+++ b/tests/framework/helpers/FileHelperTest.php
@@ -138,6 +138,85 @@ class FileHelperTest extends TestCase
         }
     }
 
+    public function testCopyDirectoryRecursive()
+    {
+        $srcDirName = 'test_src_dir_rec';
+        $structure = [
+            'directory1' => [
+                'file1.txt' => 'file 1 content',
+                'file2.txt' => 'file 2 content',
+            ],
+            'directory2' => [
+                'file3.txt' => 'file 3 content',
+                'file4.txt' => 'file 4 content',
+            ],
+            'file5.txt' => 'file 5 content',
+        ];
+        $this->createFileStructure([
+            $srcDirName => $structure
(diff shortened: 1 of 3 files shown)


tests/framework/helpers/FileHelperTest.php
```diff
@@ -138,6 +138,85 @@ public function testCopyDirectory()
...
+
+        $basePath = $this->testFilePath;
+        $srcDirName = $basePath . DIRECTORY_SEPARATOR . $srcDirName;
+        $dstDirName = $basePath . DIRECTORY_SEPARATOR . 'test_dst_dir';
+
+        FileHelper::copyDirectory($srcDirName, $dstDirName);
+
+        $this->assertFileExists($dstDirName, 'Destination directory does not exist!');
+
+        $checker = function ($structure, $dstDirName) use (&$checker) {
```
SilverFire: Why did you put this algorithm in the closure?

Ni-san: I think there is no need to create a separate method that used and can be used only in one place

SilverFire: Why it can't be done without putting it in a closure? It's not a problem or principle question, I'm just asking to know, what is the idea of this approach.

kidol: It's calling itself recursively.

SilverFire: Oh, sorry, I've missed it

SilverFire:
Merged, thank you!

PR #11046, opened 2016-03-07
Update BaseHtml.php

Skinka:
If the transfer of Html::activeCheckbox in the options 'label' => false all the same Tag <label> is created but empty. and in theory it should not be at

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -691,7 +691,7 @@ class BaseHtml
         } else {
             $hidden = '';
         }
-        if (isset($options['label'])) {
+        if (isset($options['label']) && $options['label']) {
             $label = $options['label'];
             $labelOptions = isset($options['labelOptions']) ? $options['labelOptions'] : [];
             unset($options['label'], $options['labelOptions']);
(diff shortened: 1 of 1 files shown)


SilverFire:
Looks reasonable. Could you create a unit test for this case?

Skinka:
unfortunately I do not know how to write a unit test

klimov-paul:
> If the transfer of Html::activeCheckbox in the options 'label' => false all the same Tag is created but empty. and in theory it should not be at

Why? You should pass `label` to be `null` if you do not want label to be rendered.

SilverFire:
I'm reopening this PR as a documentation issue in order to update method's PHPDoc and mention explicitly that we can pass `null`

Skinka:
but it is logical to use the same non-null in the class and mainly use the widget and then false
http://www.yiiframework.com/doc-2.0/yii-bootstrap-activefield.html#label()-detail

SilverFire:
No need to update PHPDoc, it is already correct:

> label: string, a label displayed next to the radio button. It will NOT be HTML-encoded. Therefore you can pass in HTML code such as an image tag. If this is is coming from end users, you should encode() it to prevent XSS attacks. __When this option is specified__, the radio button will be enclosed by a label tag.

PR #11055, opened 2016-03-08
Added Yii logo

philipgura:


SilverFire:
The idea is good, but image loading via HTTP on HTTP**S** page usually leads to resource blocking because of insecure content.

philipgura:
you are right, missed that, found a logo with https
https://en.wikipedia.org/wiki/File:Yii-logo-transparent.png
#11067

PR #11056, opened 2016-03-09
Support customer logger objects via config arrays

bionoren:
For reasons I don't fully understand, if you define in your web.php config ['log' => 'logger' => ['class' => 'app\custom\logger'], ...], the value of $config['logger'] at this point is ['class' => 'app\custom\logger'], which setLogger will try to treat as an option on line 127 (which obviously causes an error). This is my attempt to support the standard config array.

Note that you can't instantiate app\custom\logger and provided it in the logger config value, since the autoloader doesn't know about your app namespace yet.

### DIFF
--- a/framework/log/Dispatcher.php
+++ b/framework/log/Dispatcher.php
@@ -80,7 +80,7 @@ class Dispatcher extends Component
     {
         // ensure logger gets set before any other config option
         if (isset($config['logger'])) {
-            $this->setLogger($config['logger']);
+            $this->setLogger(Yii::createObject($config['logger']));
             unset($config['logger']);
         }
         // connect logger and dispatcher
--- a/framework/log/Dispatcher.php
+++ b/framework/log/Dispatcher.php
@@ -80,7 +80,7 @@ class Dispatcher extends Component
     {
         // ensure logger gets set before any other config option
         if (isset($config['logger'])) {
-            $this->setLogger(Yii::createObject($config['logger']));
+            $this->setLogger(is_array($config['logger']) ? Yii::createObject($config['logger']) : $config['logger']);
             unset($config['logger']);
         }
         // connect logger and dispatcher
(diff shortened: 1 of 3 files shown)


framework/log/Dispatcher.php
```diff
@@ -80,7 +80,7 @@ public function __construct($config = [])
     {
         // ensure logger gets set before any other config option
         if (isset($config['logger'])) {
-            $this->setLogger($config['logger']);
+            $this->setLogger(Yii::createObject($config['logger']));
```
kidol: Looks like you have to check if `$config['logger']` is actually an array, otherwise `createObject()` will complain (see failing tests). This might work:
`$this->setLogger(is_array($config['logger']) ? Yii::createObject($config['logger']) : $config['logger']);`

bionoren: I updated the tests to use the config semantics before I saw this comment. Would you prefer I back those changes out and add the is_array check?

kidol: Well with your fix you can set object instance via `setLogger()`, but not via constructor config. I don't think this is the way to go. Better do is_array check and don't touch the tests.

Though you may wait until Yii team member makes a comment. :)

cebe:
Fixed by 22503eb, thanks for your contribution!

framework/log/Dispatcher.php
```diff
@@ -80,7 +80,7 @@ public function __construct($config = [])
     {
         // ensure logger gets set before any other config option
         if (isset($config['logger'])) {
-            $this->setLogger($config['logger']);
+            $this->setLogger(is_array($config['logger']) ? Yii::createObject($config['logger']) : $config['logger']);
```
klimov-paul: Should not it be `Instance::ensure()` instead?

klimov-paul:
CHANGELOG entry is missing.

cebe:
@klimov-paul see this commit for the changes in master: https://github.com/yiisoft/yii2/commit/22503ebe97c3280e516e323105f51e368d017959 I did not merge this PR.

framework/log/Dispatcher.php
cebe: I think that could be a BC break as we allowed the logger to be any class before this. `Instance::ensure` will fail if someone passes a custom implementation.

klimov-paul: OK, fine then.

PR #11066, opened 2016-03-09
Update yii.js

DrDeath72:
add empty url and fix hash

### DIFF
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -272,12 +272,13 @@ yii = (function ($) {
         },
 
         getQueryParams: function (url) {
+            url = url || document.location.href;
             var pos = url.indexOf('?');
             if (pos < 0) {
                 return {};
             }
 
-            var pairs = url.substring(pos + 1).split('&'),
+            var pairs = url.substring(pos + 1).split('#')[0].split('&'),
                 params = {},
                 pair,
                 i;


SilverFire:
Could you elaborate in details, what this PR enhances/fixes, please?

DrDeath72:
1. yii.getQueryParams() use page url by default, not error
2. fix hash error {page: "4#list"}

framework/assets/yii.js
```diff
@@ -272,12 +272,13 @@ yii = (function ($) {
         },
 
         getQueryParams: function (url) {
+            url = url || document.location.href;
```
klimov-paul: Are we allow calling `getQueryParams` without arguments?
This change will just hide possible developer mistake.
Does not seem to be right.

framework/assets/yii.js
```diff
@@ -272,12 +272,13 @@ yii = (function ($) {
...
             var pos = url.indexOf('?');
             if (pos < 0) {
                 return {};
             }
 
-            var pairs = url.substring(pos + 1).split('&'),
+            var pairs = url.substring(pos + 1).split('#')[0].split('&'),
```
klimov-paul: Make sense. Still need to test it out.

SilverFire: Works as expected :+1:

framework/assets/yii.js
SilverFire: Agree on this, should be removed.

SilverFire:
We need a CHANGELOG line

framework/assets/yii.js
DrDeath72: noooooooooooooooo

SilverFire: [No!](http://www.nooooooooooooooo.com/) :)
Share your thoughts, please. What is your use case?

DrDeath72: don't write every time document.location.href to work with current url

SilverFire:
Merged https://github.com/yiisoft/yii2/commit/08b9f12afa39ed55c0f35e67502bc23a1d4ff1f2

SilverFire:
Thank you!

PR #11087, opened 2016-03-11
fixes #9561 canGetProperty and canSetProperty on ActiveRecord checks attributes

### LINKED ISSUE #9561: `canGetProperty()` and `canSetProperty()` does not take in account `ActiveRecord::attributes()`
Methods `Object::canGetProperty()` and `Object::canSetProperty()` does not take in account `ActiveRecord::attributes()`, always returning `false` for them:

```php
/**
 * Properties, which are attributes from source table:
 *
 * @property integer $id
 * @property string $username
 */
class User extends \yii\db\ActiveRecord
{
    ...
}

$user = new User();
$this->assertTrue($user->canGetProperty('id')); // FAILS !!!
$this->assertTrue($user->canSetProperty('id')); // FAILS !!!
```

Inside `yii\db\BaseActiveRecord` we override `__set()` and `__get()` methods, but there is no redeclaration of `canGetProperty()` or `canSetProperty()`.

pchapl:
Maybe need to add more tests for readonly/writeonly properties and "related"?

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -423,6 +423,42 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
     }
 
     /**
+     * Returns a value indicating whether a property can be read.
+     * This method is overridden as the attributes can be accessed like properties.
+     *
+     * @param string $name the property name
+     * @param boolean $checkVars whether to treat member variables as properties
+     * @param boolean $checkBehaviors whether to treat behaviors' properties as properties of this component
+     * @return boolean whether the property can be written
+     */
+    public function canGetProperty($name, $checkVars = true, $checkBehaviors = true)
+    {
+        if ($this->hasAttribute($name) || isset($this->_related[$name]) || array_key_exists($name, $this->_related)) {
+            return true;
+        } else {
+            return parent::canGetProperty($name, $checkVars, $checkBehaviors);
+        }
+    }
+
+    /**
+     * Returns a value indicating whether a property can be set.
+     * This method is overridden as the attributes can be accessed like properties.
+     *
+     * @param string $name the property name
+     * @param boolean $checkVars whether to treat member variables as properties
+     * @param boolean $checkBehaviors whether to treat behaviors' properties as properties of this component
+     * @return boolean whether the property can be written
+     */
+    public function canSetProperty($name, $checkVars = true, $checkBehaviors = true)
+    {
+        if ($this->hasAttribute($name)) {
+            return true;
+        } else {
+            return parent::canSetProperty($name, $checkVars, $checkBehaviors);
+        }
+    }
+
+    /**
      * Returns the named attribute value.
      * If this record is the result of a query and the attribute is not loaded,
      * null will be returned.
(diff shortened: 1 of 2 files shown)


framework/base/Component.php
```diff
@@ -333,7 +333,7 @@ public function hasProperty($name, $checkVars = true, $checkBehaviors = true)
      */
     public function canGetProperty($name, $checkVars = true, $checkBehaviors = true)
     {
-        if (method_exists($this, 'get' . $name) || $checkVars && property_exists($this, $name)) {
+        if (parent::canGetProperty($name, $checkVars)) {
```
cebe: I think these should stay as is for performance reasons.

framework/base/Component.php
```diff
@@ -363,7 +363,7 @@ public function canGetProperty($name, $checkVars = true, $checkBehaviors = true)
      */
     public function canSetProperty($name, $checkVars = true, $checkBehaviors = true)
     {
-        if (method_exists($this, 'set' . $name) || $checkVars && property_exists($this, $name)) {
+        if (parent::canSetProperty($name, $checkVars)) {
```
cebe: same here.

cebe:
Thanks, would be great if you could add more tests.

framework/db/BaseActiveRecord.php
```diff
@@ -423,6 +423,42 @@ public function hasAttribute($name)
...
+     * This method is overridden as the attributes can be accessed like properties.
+     *
+     * @param string $name the property name
+     * @param boolean $checkVars whether to treat member variables as properties
+     * @param boolean $checkBehaviors whether to treat behaviors' properties as properties of this component
+     * @return boolean whether the property can be written
+     */
+    public function canSetProperty($name, $checkVars = true, $checkBehaviors = true)
+    {
+        if ($this->hasAttribute($name)) {
```
Faryshta: Would be faster and cleaner to do something like

```php
return $this->hasAttribute($name)
    || parent::canSetProperty($name, $checkVars, $checkBehaviors);
```

framework/db/BaseActiveRecord.php
```diff
@@ -423,6 +423,42 @@ public function hasAttribute($name)
...
+     * This method is overridden as the attributes can be accessed like properties.
+     *
+     * @param string $name the property name
+     * @param boolean $checkVars whether to treat member variables as properties
+     * @param boolean $checkBehaviors whether to treat behaviors' properties as properties of this component
+     * @return boolean whether the property can be written
+     */
+    public function canGetProperty($name, $checkVars = true, $checkBehaviors = true)
+    {
+        if ($this->hasAttribute($name) || isset($this->_related[$name]) || array_key_exists($name, $this->_related)) {
```
Faryshta: Would be faster and cleaner to do something like

```php
return $this->hasAttribute($name)
    || isset($this->_related[$name])
    || array_key_exists($name, $this->_related)
    || parent::canGetProperty($name, $checkVars, $checkBehaviors);
```
If any of those is true, then we should return true.

cebe:
This has been fixed by https://github.com/yiisoft/yii2/commit/75439d367d2d1270bb61642f236df6a552acf29e alredy, but I have merged your test, thank you!

Issue #11092, opened 2016-03-11
Bug with Expression parameters binding in a query


Mirocow:
Such dont work

```php
                    $models = CarModel::find()->select([
                            '*',
                            new Expression('MAX( IF(car_generation.year_end IS NULL, 9999, car_generation.year_end) ) as max_year')
                    ])->joinWith('generation')->where(['and',
                            new Expression('car_generation.year_begin IS NOT NULL'),
                            new Expression('cars_model.mark_id=:mark_id AND car_generation.year_begin IS NOT NULL',  [':mark_id' => $this->context->route['models']['MARK']['id']]),
                    ])->groupBy(['cars_model.model_id'])->asArray()->all();
```
but that work right

```php
                    $models = CarModel::find()->select([
                            '*',
                            new Expression('MAX( IF(car_generation.year_end IS NULL, 9999, car_generation.year_end) ) as max_year')
                    ])->joinWith('generation')->where(['and',
                            new Expression('car_generation.year_begin IS NOT NULL'),
                            new Expression('cars_model.mark_id=:mark_id AND car_generation.year_begin IS NOT NULL'),
                    ], [':mark_id' => $this->context->route['models']['MARK']['id']])->groupBy(['cars_model.model_id'])->asArray()->all();
```

So it is a bug or intended?

samdark:
Looks like parameters passed via Expression constructor either aren't used or are overridden instead of being merged.

samdark:
@Mirocow can you check pull request ↑

klimov-paul:
> Such dont work

How exactly it does not work? What is exact error message?

samdark:
@klimov-paul about the parameters aren't bound/

SilverFire:
Fixed in https://github.com/yiisoft/yii2/commit/b182586cd3ded908cb89801e2d7cdf6fd21d810e

Thank you for your report

CedricYii:
Btw, I think that this feature should be dropped (in Expression and other where clauses as well) as I couldn't see its added value.
The code is getting less clear when using this whereas it's always possible to add an ->addParams() after the query definition. So the framework is encouraging poor coding style, and it's getting more complicated and slower without significant benefits imho.

cebe:
@CedricYii I am planning to review and simplify Querybuilder in this regard for 2.1.x

CedricYii:
@cebe Thanks you for the feedback. So Yii will stay on the good way to keep it simple, light with useful features only, that's a great news!

PR #11093, opened 2016-03-11
Fix Expression parameter binding in 'And' clause

CedricYii:
This is done to fix issue [https://github.com/yiisoft/yii2/issues/11092]


SilverFire:
Could you add add a test case for this bug?

CedricYii:
I tried to do a test case... But don't know how to test the test...
Can you give it a try?

PR #11101, opened 2016-03-12
Added BaseArrayHelper::copy

leandrogehlen:


hiscaler:
Why add this function?
```php
$a = ['a', 1 => 111];
$b = $a;
```
What's different full with copy function?

leandrogehlen:
You are right. Really it does not make sense

PR #11109, opened 2016-03-14
Adding replacement property

brussens:
Adding properties $replacement, to be able to override the delimiter


SilverFire:
Hello. Thank you for the contribution. Since 2.0.7 `SluggableBehavior` has [`generateSlug()`](http://www.yiiframework.com/doc-2.0/yii-behaviors-sluggablebehavior.html#generateSlug()-detail) method, that designed especially for such cases.

Just create your own `SluggableBehavior`, the extents from `yii\behaviors\SluggableBehavior` and override `generateSlug()` method as you wish.

PR #11110, opened 2016-03-14
Add session migration

mdmunir:


framework/web/migrations/m160313_153426_session_init.php
```diff
@@ -0,0 +1,66 @@
...
+            // http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci
+            $tableOptions = 'CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE=InnoDB';
+        }
+
+        $this->createTable($session->sessionTable, [
+            'id' => $this->string(40)->notNull(),
+            'expire' => $this->integer(),
+            'data' => $this->binary(),
+            'PRIMARY KEY ([[id]])',
+            ], $tableOptions);
```
SilverFire: One redundant tab

SilverFire:
Perfect! Could you add a CHANGELOG line?

tests/framework/console/controllers/MigrateControllerTest.php
```diff
@@ -50,4 +50,34 @@ protected function getMigrationHistory()
         $query = new Query();
         return $query->from('migration')->all();
     }
+
+    public function testSessionMigration()
```
klimov-paul: This should be removed.
The test itself may exist, but it does not belong ot the `MigrateConrollerTest`

framework/web/migrations/m160313_153426_session_init.php
```diff
@@ -0,0 +1,66 @@
...
+ * @since 2.0.8
+ */
+class m160313_153426_session_init extends Migration
+{
+
+    /**
+     * @throws yii\base\InvalidConfigException
+     * @return DbSession
+     */
+    protected function getSession()
```
klimov-paul: This is bad idea: in this way migration depends on application configuration, which may change in the future.
If you switch to the memcache based session in the future, reverting this migration will produce PHP fatal error.

tests/framework/console/controllers/MigrateControllerTest.php
mdmunir: remove or move to another section?

klimov-paul: That is hard to tell.
We do not have any unit tests for the migrations provided by the core.
I am unsure if they are necessary or not.
But if adding it - it should be a separated test case.

framework/web/migrations/m160313_153426_session_init.php
SilverFire: You are suggesting to leave only `$db`, right?

tests/framework/console/controllers/MigrateControllerTest.php
SilverFire: Well, tend to agree

framework/web/migrations/m160313_153426_session_init.php
mdmunir: Agree. But rbac migration use the same way. https://github.com/yiisoft/yii2/blob/master/framework/rbac/migrations/m140506_102106_rbac_init.php#L23-L30
For me. I prefer create table automatically when table does not exists like in Yii1. https://github.com/yiisoft/yii/blob/master/framework/web/CDbHttpSession.php#L158

SilverFire: > For me. I prefer create table automatically when table does not exists like in Yii1.

I'm strongly against this idea. I don't want Yii to touch my DB without my explicit command to do it.

janisto:
varchar(40) isn't always enough. https://github.com/yiisoft/yii2/issues/5395

SilverFire:
@janisto thank you for attention.

We can modify migration to check `session.hash_function` options's value:

```php
public function safeUp() {
    if (ini_get('session.hash_function') === 1) {
        throw new \yii\base\InvalidConfigException('PHP option "session.hash_function" is equal to 1. It means, that SHA1 hash function is used to generate the session\'s ID. Please, modify this migration and increase length of "id" column');
    }
}
```

mdmunir:
??
```php
        $length = ini_get('session.hash_function') === 1 ? 256 : 40;
        $this->createTable($session->sessionTable, [
            'id' => $this->string($length)->notNull(),
            'expire' => $this->integer(),
            'data' => $this->binary(),
            'PRIMARY KEY ([[id]])',
        ], $tableOptions);
```

df2:
Why don't you just make it 256?

SilverFire:
@mdmunir it can lead to different tables in result of same migration on different envs.

@df2 it makes row allocation 216 bytes bigger even if we don't need it.

@yiisoft/core-developers

df2:
@SilverFire less than that because string will get converted to varchar. So 40+1 or 256+1 is what it would end up as in terms of storage.

samdark:
Looks OK to me as is. I'd leave ID length to 256 since it's varchar and it doesn't really affect storage space or performance if lower length strings are stored. `char(40)` would make a difference...

framework/web/migrations/m160313_153426_session_init.php
SilverFire: I tend to leave code as is.

samdark: Session is never configured for console app so I'd remove this check.

samdark: RBAC makes sense in console.

framework/web/migrations/m160313_153426_session_init.php
```diff
@@ -0,0 +1,66 @@
...
+        $session = $this->getSession();
+        $this->db = $session->db;
+
+        $tableOptions = null;
+        if ($this->db->driverName === 'mysql') {
+            // http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci
+            $tableOptions = 'CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE=InnoDB';
+        }
+
+        $this->createTable($session->sessionTable, [
```
samdark: Since session is never used in console I'm not sure it's a good idea to require it. Since there's unlikely session component will be in console, session table can't be obtained so probably should be hardcoded...

mdmunir: how about prompt with default value, isit make sense?
```php
$tableName = Yii::$app->controller->prompt('Session table name', ['default'=>'session']);
```

samdark: Would it work well?

cebe: I do not like prompts in migrations, it means the same migration run different times will apply different schema, this is not good practise.

mdmunir:
Without interaction
screenshot from 2016-03-23 18 01 29

Use interaction
screenshot from 2016-03-23 18 01 57

framework/web/migrations/m160313_153426_session_init.php
SilverFire: I'd prefer to hardcode session table name in migration. If you want to change table name - copy migration to your project scope, change table name and apply it.

samdark: OK. Let's hardcode it then...

SilverFire:
Merged: https://github.com/yiisoft/yii2/commit/d437c4bd3755109b786131195d5e4e0e6985768b

SilverFire:
Thank you!

PR #11113, opened 2016-03-14
phpDoc \yii\widgets\ActiveField::error

githubjeka:
Now phpDoc is wrong. `Null` causes an error! :ambulance:


SilverFire:
Thank you!

Issue #11111, opened 2016-03-14
generateRandomKey in freebsd


quantum13:
Hello.

I have Yii 2.0.7 on FreeBSD 10.2-RELEASE
output of diagnostic script from #10519:
```
1  function_exists('random_bytes')
    false

 2  defined('OPENSSL_VERSION_TEXT') ? OPENSSL_VERSION_TEXT : null
    'OpenSSL 1.0.1p-freebsd 9 Jul 2015'

 3  PHP_VERSION_ID
    50618

 4  function_exists('mcrypt_create_iv') ? bin2hex(mcrypt_create_iv(8, MCRYPT_DEV_URANDOM)) : null
    NULL

 5  DIRECTORY_SEPARATOR
    '/'

 6  sprintf('%o', lstat('/dev/urandom')['mode'])
    '120755'

 7  sprintf('%o', lstat('/dev/urandom')['mode'] & 0170000)
    '120000'

 8  bin2hex(file_get_contents('/dev/urandom', false, null, 0, 8))
    '9918d772ab5fd4cb'

 9  ini_get('open_basedir')
    ''
```
I know, that I can install mcrypt, and generateRandomKey will work fine. 

In comment in generateRandomKey in branch with /dev/urandom written `Do not follow symbolic link at '/dev/urandom', as such would be suspicious`. But in my system /dev/urandom is link to /dev/random. I checked freebsd 9.0, /dev/urandom is also link to /dev/random. 

https://en.wikipedia.org/wiki//dev/random#FreeBSD :

> Unlike the Linux /dev/random, the FreeBSD /dev/random device never blocks. Its behavior is similar to the Linux /dev/urandom, and /dev/urandom on FreeBSD is linked to /dev/random.

Maybe for freebsd addtitional checks can be added?

SilverFire:
FreeBSD 9.3:

```
 1  function_exists('random_bytes')
    false

 2  defined('OPENSSL_VERSION_TEXT') ? OPENSSL_VERSION_TEXT : null
    'OpenSSL 1.0.1i 6 Aug 2014'

 3  PHP_VERSION_ID
    50431

 4  function_exists('mcrypt_create_iv') ? bin2hex(mcrypt_create_iv(8, MCRYPT_DEV_URANDOM)) : null
    '32399989674a4be0'

 5  DIRECTORY_SEPARATOR
    '/'

 6  sprintf('%o', lstat('/dev/urandom')['mode'])
    '120755'

[...]

PR #11118, opened 2016-03-15
use /dev/random in not linux systems

vlad-khramov:
#11111

### DIFF
--- a/framework/base/Security.php
+++ b/framework/base/Security.php
@@ -423,10 +423,12 @@ class Security extends Component
         return false;
     }
 
+    const DEV_RANDOM = '/dev/random';
     const DEV_URANDOM = '/dev/urandom';
     const SOURCE_LIBRE_SSL = 'LibreSSL';
     const SOURCE_MCRYPT = 'mcrypt';
     const SOURCE_OPEN_SSL = 'OpenSSL';
+    const SOURCE_RANDOM = 'random';
     const SOURCE_URANDOM = 'urandom';
     /**
      * @var string|null Identifies the random source of the last successful call of [[generateRandomKey]].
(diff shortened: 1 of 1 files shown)


SilverFire:
:+1: 

@tom-- @yiisoft/core-developers

tom--:
This looks very good.

tom--:
On second thoughts (sorry) I think the branching logic should be a bit different. The purpose of the `$_randomSource` property is to "cache" the branch decision for the next call (and all subsequent calls) to `generateRandomKey()`. So if, on the second call, we don't test `$_randomSource` *first* and immediately use the cached value then the property isn't doing any useful work for us--in fact setting and checking it is just added complexity.

tom--:
Let's take this opportunity to consider the design.

The only argument *pro* `$_randomSource` (i.e. "caching" the branch decision for subsequent calls) is performance when the application makes multiple calls during the life of one instance of the Security component.

The *contra* arguments are

1. it makes performance a little worse when the app makes only one call
2. it adds complexity

The performance argument is not really persuasive one way or the other given that the Yii framework supports many kinds of apps and processing. It's an optimization that helps in some use cases.

Otoh, code complexity is never desirable. Get rid of this optimization and the method simplifies considerably.

samdark:
Unlike password hashing, random numbers may be used in performance-critical app parts. I think `$_randomSource` makes sense in this case.

quantum13:
>I think the branching logic should be a bit different

Yes, I understand it. But overhead in my code is only of two conditions and two variables assignment before checking _randomSource

In another way it could be done like this https://github.com/quantum13/yii2/commit/4ee2f474ed5b77477e16138e7488bd691aa9a8d2 , but I thought, that three almost identical ternary operators is uglier than pre assignment.

And, I think, full copypaste of this branch with condition about Linux and replacing urandom to random isn't good choice.

samdark:
Closed in favor of https://github.com/yiisoft/yii2/issues/11146

PR #11125, opened 2016-03-15
JSON_ERROR_SYNTAX fix for json_decode(null) in PHP 7

fps01:
json_decode produce error when decoding `null` or empty string in PHP 7.

Example.

``` php
var_dump(json_decode(null));
var_dump(json_last_error());
```

PHP 5.4.0 - 5.6.19, hhvm-3.6.1 - 3.12.0 returns:

```
NULL
int(0)
```

PHP 7.0.0 - 7.0.4 returns:

```
NULL
int(4)
```

I think we must exclude `null` and epmty string from a decode function becouse json_decode returns a right value.
What do you think about it?

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -28,6 +28,7 @@ Yii Framework 2 Change Log
 - Bug #11040: Check parameter 'recursive' and disable recursive copying with option 'recursive' => false in method BaseFileHelper::copyDirectory (Ni-san)
 - Chg: HTMLPurifier dependency updated to `~4.6` (samdark)
 - Chg #11071: `yii\helpers\BaseArrayHelper::isIn()` and `isTraversable()` since now throw `\yii\base\InvalidParamException` instead of `\InvalidArgumentException` (nukkumatti)
+- Bug #11125: Fixed JSON_ERROR_SYNTAX for json_decode(null) in PHP 7
 
 2.0.7 February 14, 2016
 -----------------------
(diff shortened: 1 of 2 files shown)


framework/helpers/BaseJson.php
```diff
@@ -90,6 +90,8 @@ public static function decode($json, $asArray = true)
     {
         if (is_array($json)) {
             throw new InvalidParamException('Invalid JSON data.');
+        } elseif ((string) $json === '') {
```
samdark: It's better to check it explicitly with comparing to both null and empty string.

fps01:
@samdark I split it

SilverFire:
Could you add a CHANGELOG line, please?

fps01:
@SilverFire Done

samdark:
Thanks!

framework/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ Yii Framework 2 Change Log
 - Bug #11040: Check parameter 'recursive' and disable recursive copying with option 'recursive' => false in method BaseFileHelper::copyDirectory (Ni-san)
 - Chg: HTMLPurifier dependency updated to `~4.6` (samdark)
 - Chg #11071: `yii\helpers\BaseArrayHelper::isIn()` and `isTraversable()` since now throw `\yii\base\InvalidParamException` instead of `\InvalidArgumentException` (nukkumatti)
+- Bug #11125: Fixed JSON_ERROR_SYNTAX for json_decode(null) in PHP 7
```
SilverFire: nickname is missing

samdark: Not in https://github.com/yiisoft/yii2/commit/258175149d1dceb44c382ea34339dafff3eed107

fps01:
@SilverFire @samdark thanks!

PR #11140, opened 2016-03-18
Implement weak Etag setting for HttpCache filter, fixes #11137

### LINKED ISSUE #11137: HttpCache should be able to generate weak etags
When using `yii\filters\HttpCache` for etag generation, it should be configurable to allow generation of weak etags due to following problem:

1. Yii generates _standard_ etag
2. nginx modifies it to be a _weak_ etag
3. the client sends `If-None-Match` with the weak etag
4. Yii2 compares a weak to a non-weak etag, fails and re-sends response

This happens only when nginx has gzip compression enabled and with nginx >= 1.7 ( http://hg.nginx.org/nginx/rev/e491b26fa5a1 )

I think a configurable option to generate _weak_ etags would be the best solution to that problem.

particleflux:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11137 |

### DIFF
--- a/framework/filters/HttpCache.php
+++ b/framework/filters/HttpCache.php
@@ -71,6 +71,12 @@ class HttpCache extends ActionFilter
      */
     public $etagSeed;
     /**
+     * @var bool whether to generate weak Etags
+     *
+     * Weak Etags should be used if the content should not be considered byte-equal, but semantically equivalent.
+     */
+    public $weakEtag = true;
+    /**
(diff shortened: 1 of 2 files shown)


samdark:
Merged https://github.com/yiisoft/yii2/commit/1dfab19b77d2ea719ccd0c59a0bf68f52adcd996 with some adjustmentes. Thanks!

PR #11144, opened 2016-03-18
Add fail test 

ibra-x347:
Error upload file .svg : $mimeType == Null (\yii\validators\FileValidator::validateExtension)
FileValidator: [extensions] => ['svg']
Error after upload: "Only files with these extensions are allowed: svg"

``` sh
~/www/yii2$ vendor/bin/phpunit tests/framework/validators/FileValidatorTest.php
PHPUnit 4.8-dev by Sebastian Bergmann and contributors.

.......F..............

Time: 75 ms, Memory: 6.75Mb

There was 1 failure:

1) yiiunit\framework\validators\FileValidatorTest::testValidateMimeTypeMaskValid with data set #0 ('test.svg', 'image/*')

/home/GURTAM/zibr/www/yii2/tests/framework/validators/FileValidatorTest.php:358
/home/GURTAM/zibr/www/yii2/vendor/phpunit/phpunit/phpunit:35

FAILURES!
Tests: 22, Assertions: 71, Failures: 1.

```


SilverFire:
That happens because file's mime type is `text-plain`.

```
# > file --mime-type test.svg
test.svg: text/plain
```

When I try to upload your test file via web form, browser sends the following data:

```
------WebKitFormBoundaryBMShRGnGSB5cL0nu
Content-Disposition: form-data; name="TestForm[file]"; filename="test.svg"
Content-Type: image/svg+xml
```

So we can say the Chrome makes it possible to handle SVG uploads correctly. But it does not work for saved file. Unfortunately, the problem is out of Yii framework scope.

Thank you for your report!

PR #11150, opened 2016-03-20
EachValidator injects specific attribute value in error message

### LINKED ISSUE #11139: Each Validator message formatting

```php
public function rules() {
    return [
			['mmm', 'each', 'rule' => [
				'exist', 'targetClass' => zzz::className(), 'targetAttribute' => 'xxx',
				'message' => 'User "{value}" was not found.'
			]],
    ];
}
```
```User xxx was not found```
```User yyy was not found```
### What do you get instead?
```User "array()" was not found.```
### Additional info

| Q                | A
| ---------------- | ---
|   Yii    | 2.0.7-dev
|   Php    | 5.5
| Windows | 7

### PR DESCRIPTION
**Pull request by SilverFire**
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11139 |

Closes #11139
- [x] Review

Could anybody review this PR, please?

### CODE REVIEW

File: framework/validators/EachValidator.php [hunk 1]
```diff
@@ -146,7 +146,12 @@ protected function validateValue($value)
             }
             $result = $validator->validateValue($v);
             if ($result !== null) {
-                return $this->allowMessageFromRule ? $result : [$this->message, []];
+                if ($this->allowMessageFromRule) {
+                    $result[1]['value'] = $v;
```
samdark: Looks a bit weird. Why index is 1?

SilverFire: `['This is {value}', ['value' => 'test]]`
0 - message
1 - parameters

samdark: Ah. Then PR looks fine.

PR #11166, opened 2016-03-22
`yii migrate/create` command generate migrate file support table pref…

hiscaler:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues |  |

Improve migrate command.

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -61,7 +61,12 @@ abstract class BaseMigrateController extends Controller
      * @since 2.0.7
      */
     public $fields = [];
-
+    /**
+     * @var Connection|array|string the DB connection object or the application component ID of the DB connection to use
+     * when applying migrations. Starting from version 2.0.3, this can also be a configuration array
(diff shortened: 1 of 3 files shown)


framework/console/controllers/BaseMigrateController.php
```diff
@@ -61,7 +61,12 @@
      * @since 2.0.7
      */
     public $fields = [];
-
+    /**
+     * @var Connection|array|string the DB connection object or the application component ID of the DB connection to use
+     * when applying migrations. Starting from version 2.0.3, this can also be a configuration array
+     * for creating the object.
+     */
+    public $db = 'db';
```
samdark: That is wrong because `BaseMigrateController` is used, for example, for MongoDB which does not have prefixes.

samdark: Should be done in `MigrateController` class instead.

samdark:
Closed in favor of https://github.com/yiisoft/yii2/issues/11287

PR #11171, opened 2016-03-23
Fixed yii\widgets\ActiveField to handle AJAX validation for inputs with changed IDs

aleskiontherun:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #7627 |

If `ActiveForm::$enableClientValidation` is set to `false` and `ActiveForm::$enableAjaxValidation` is set to `true`, than validation doesn't work no more because errors returned from the ajax validation are still formatted as `{"model-attribute": ["Error"]}`, but the attribute id in clientOptions is now set to `customid`, not `model-attribute`, so the script doesn't see the attribute errors in the response. I think the proper code in \yii\widgets\ActiveField::getClientOptions() should be:

```
$inputID = $this->getInputId();
$options['id'] = Html::getInputId($this->model, $this->attribute);
// not $options['id'] = $inputID
```

I'll tinker a bit with it and make a pull request if I'll find this working for both client and ajax validation.
Fix for the #7627 was already made by @dynasource and @cebe and the issue was closed, but it fixed only client validation while AJAX validation was left broken.

### DIFF
--- a/framework/widgets/ActiveField.php
+++ b/framework/widgets/ActiveField.php
@@ -725,7 +725,7 @@ class ActiveField extends Component
         $options = [];
 
         $inputID = $this->getInputId();
-        $options['id'] = $inputID;
+        $options['id'] = Html::getInputId($this->model, $this->attribute);
         $options['name'] = $this->attribute;
 
         $options['container'] = isset($this->selectors['container']) ? $this->selectors['container'] : ".field-$inputID";
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -5,6 +5,7 @@ Yii Framework 2 Change Log
 -----------------------
 
 - Bug #11196: Fixed VarDumper throws PHP Fatal when dumping __PHP_Incomplete_Class (DamianZ)
+- Bug #7627: Fixed `yii\widgets\ActiveField` to handle inputs AJAX validation with changed ID properly (dizeee)
 - Bug #9851: Fixed partial commit / rollback in nested transactions (sammousa)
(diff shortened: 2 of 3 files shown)


SilverFire:
Looks good, but it's a tricky code and it will be good to cover it with tests. Could you do it, please?

dizeee:
Yes, sure.

framework/widgets/ActiveField.php
```diff
@@ -721,7 +721,7 @@ protected function getClientOptions()
         $options = [];
 
         $inputID = $this->getInputId();
-        $options['id'] = $inputID;
+        $options['id'] = Html::getInputId($this->model, $this->attribute);
```
cebe: doesn't this break client validation again?

dizeee: @cebe How exactly? As far as I can tell the issue was in css selectors which were fixed earlier but the `id` was fine and after the fix it's not. Any way this one fixes AJAX validation while client validation remains working. I checked it manually but I'll make some automatic tests this week to clear things out.

dizeee:
Added tests and looked through yii.activeForm.js and yii.validation.js once again. It's all right now: `id` is used to identify error messages and is not changed when custom input id is set; `input` and `container` are used to find elements and are changed when custom input id is set.

dizeee:
Rebased on top of master.

SilverFire:
Merged https://github.com/yiisoft/yii2/commit/8e856bbb5a13599835e8d8e3cc4d5fd9e6287ee1

SilverFire:
Thank you!

shushenghong:
this really breaks the client validation, if couple same models in one <form>, i will post a bug

Issue #11172, opened 2016-03-23
Cache Component Enhancement, shortcut to retrieve cache data


ryusoft:
##### Suggestion to add shortcut function "retrieve" to "\yii\caching\Cache" Component as below:
```php
/**
  * Retrieve data from cache or generate the content if cache not found or expired
  * a short cut for get and set the value from cache
  *
  * @param mixed $key a key identifying the value to be cached. This can be a simple string or
  * a complex data structure consisting of factors representing the key.
  * @param mixed $callback the callback function that return data to be cache
  * @param integer $duration the number of seconds in which the cached value will expire. 0 means never expire.
  * @param \yii\caching\Dependency $dependency dependency of the cached item. If the dependency changes,
  * the corresponding value in the cache will be invalidated when it is fetched via [[get()]].
  * This parameter is ignored if [[serializer]] is false.
  * @return mixed the data cached
  */
public function retrieve($key, $callback, $duration = 0, $dependency = null)
{
     if (($data = $this->get($key)) === false) {
         $data = call_user_func($callback);
         $this->set($key, $data, $duration, $dependency);
     }
     return $data;
}
```

So instead of every time when trying to retrieve data from cache e.g.
```php
$key = 'data-cache-key';
$duration = 3600;
$dependency = null;//or some cache dependency

if (($data = Yii::$app->cache->get($key)) === false) {
    $data = [
        ['name'=>'RYU', 'Ranking': 9],
    ];//heavy computation db query or calculation to generate this data
    Yii::$app->cache->set($key, $data, $duration, $dependency);
}

//continue using $data to do something else
```

We could write the code as below:
```php
$key = 'data-cache-key';
$duration = 3600;
$dependency = null;//or some cache dependency

$data = Yii::$app->cache->retrieve($key, function() {
    return [
        ['name'=>'RYU', 'Ranking': 9],
    ];//heavy computation db query or calculation to generate this data
}, $duration, $dependency);

//continue using $data to do something else
```

SilverFire:
No way :laughing: I discussed this idea with @samdark in PM month ago and planned to implement it soon :)

Currently I have a project-scope implementation, that works for me ([code](https://github.com/hiqdev/hipanel-core/blob/master/src/base/Cache.php)) but it can not be accepted in core because it is not flexible enough. This thing is trickier than it appears.

The following use cases must be covered:

1. Static key, closure depends on some variables:
```php
[...]

PR #11173, opened 2016-03-23
enhancement as stated in #11172, a function to simplify cache data re…

ryusoft:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? |  |
| Fixed issues | #11172 |


SilverFire:
No way :laughing: I discussed this idea with @samdark in PM month ago and planned to implement it soon :)

Currently I have a project-scope implementation, that works for me ([code](https://github.com/hiqdev/hipanel-core/blob/master/src/base/Cache.php)) but it can not be accepted in core as is because it is not flexible enough.

One of the main idea was closure with key based on some variables:
```php
$cache->retrieve(['foobarcache', $user->images, $user->id], function () use ($user) {
      return Foo::longApiCall('getBar', [$user->images, $user->id]);
}, 3600);
```

We need docs, tests and CHANGELOG to merge this PR

ryusoft:
I implemented this to my code base, and thinking really save time on repeating if condition, just the IDE can't auto suggest this function.

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/aed3a222a20edd14859053a45b6b62889ca41a59

PR #11176, opened 2016-03-23
Issue 11088 field not quoted in joinwith condition

pchapl:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no? |
| Tests pass? | yes (tested only with pgsql) |
| Fixed issues | #11088; #11135 |

The only solution I found: it looks ugly, but must not break BC. Should be tested with other db-drivers and, probably, in other cases.


cebe:
Cool, thanks for finding this. Will take a more detailed look later.

cebe:
Thanks for the fix, however the problem is on the lower layer in Schema and Connection and should be fixed there.

closing in favor of #11403

PR #11179, opened 2016-03-23
Fix documentation of VerbFilter

micro-maureen:
Delimited asterisks in documentation that were being interpreted as Markdown emphasis.


SilverFire:
I don't think this PR is not acceptable because it confuses developers when the read this PHPDoc in IDE.

@cebe can we handle it in APIDoc somehow?
http://www.yiiframework.com/doc-2.0/yii-filters-verbfilter.html#$actions-detail

cebe:
better wrap them into a code block instead of excaping.

cebe:
fixed, thanks for reporting!

Issue #8351, opened 2015-05-11
Array value in Request::getQueryParam


githubjeka:
Select2 send `search['term']=123123`:

```
/index.php?r=site%2Fnode-list&search%5Bterm%5D=123123
```

My code is bad:

```php
        if (Yii::$app->request->getQueryParam('search') === null) {
            throw new BadRequestHttpException('Не корректный запрос');
        }

        $searchModel = new ElasticSearch();
        $dataProvider = $searchModel->search(
            [
                'address' => Yii::$app->request->getQueryParam('search')["term"],
                'number' => Yii::$app->request->getQueryParam('search')["term"],
            ]
        );
```

Three problems:
- `getQueryParam('search["term"]')` return null
- `getQueryParam('search%5Bterm%5D')` return null
- For future getQueryParam('search')["term"] may return PHP Notice. Need check of `isset(Yii::$app->request->getQueryParam('search')["term"])`

:(

cebe:
```
        if (($search = Yii::$app->request->getQueryParam('search')) === null) {
            throw new BadRequestHttpException('Не корректный запрос');
        }

        $searchModel = new ElasticSearch();
        $dataProvider = $searchModel->search(
            [
                'address' => ArrayHelper::getValue($search, 'term'),
                'number' => ArrayHelper::getValue($search, 'term'),
            ]
        );
```

githubjeka:
Ok it is fix for my third problem. And first or second?

githubjeka:
```php
if ($search = ArrayHelper::getValue(Yii::$app->request->getQueryParam('search'),  'term') === null) {
    throw new BadRequestHttpException('Не корректный запрос');
}
//...
 $dataProvider = $searchModel->search(['address' => $search,'number' => $search]);
}
```
Hmm instead  :)
```php
isset($_GET['search']['term'])
```

TerraSkye:
it would be nice to have the ability to acces nested params by doing  request->getQueryParam('search.term')

Pixel and tonic do it in theur HttpRequestService.
<code>
private function _getParam($name, $defaultValue, $data)
	{
		// Do they just want the whole array?
		if (!$name)
		{
			return $this->_utf8AllTheThings($data);
		}

		// Looking for a specific value?
		if (isset($data[$name]))
		{
			return $this->_utf8AllTheThings($data[$name]);
		}

		// Maybe they're looking for a nested param?
		if (strpos($name, '.') !== false)
		{
			$path = explode('.', $name);
[...]

Issue #8100, opened 2015-04-16
Using ArrayHelper::getValue() for Yii::$app->request->get() and Yii::$app->request->post()


arogachev:
There is built-in way to access nested array properties via `ArrayHelper::getValue()` using dot notation, for example:

```php
use yii\helpers\ArrayHelper;

$array = [
    'property1' => [
        'property2' => [
            'property3' => 'value',
        ],
    ],
];

$value = ArrayHelper::getValue($array, 'property1.property2.property3');
```
Maybe add shortcut for `Yii::$app->request->get()` and `Yii::$app->request->post()`?

Example:

```php
$value = Yii::$app->request->post('property1.property2.property3');
```

instead of:

```php
$value = ArrayHelper::getValue(Yii::$app->request->post(), 'property1.property2.property3');
```

Maybe it's better to check second parameter with `strpos` for dot existence first and then apply `ArrayHelper::getValue()` call. What do you think?

bethrezen:
:+1:

qiangxue:
I'm against this. This is abusing the dot notation because you can say the same thing for every place like `post()`.

ScottHuangZL:
Agree Qiang's point.

PR #11185, opened 2016-03-24
Multidimensional Superglobals in request

E-conomy:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes (also used for months in large live environment) |
| Fixed issues | Multidimensional Superglobals can now be written/read |

Currently, users can only read one dimensional $_GET $_POST etc.. SUPERGLOBALS. Such as 

`Yii::$app->request->post('item');`

With this change users can also read multidimensional GET/POST etc..

An array can be provided as parameter now:

`Yii::$app->request->post(['item', 'description'])`

or

`Yii::$app->request->get(['user', 'profile', 'name'])`

There is no limit (outside HTML form HTTP header limitations) on the depth of dimensions that can be accessed.

**Most importantly: The current functionality is fully maintained, this addition only extends with new functionality.**

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -440,7 +440,18 @@ class Request extends \yii\base\Request
     {
         $params = $this->getBodyParams();
 
-        return isset($params[$name]) ? $params[$name] : $defaultValue;
+        if (is_array($name)) {
+            foreach ($name as $n) {
+                if (isset($params[$n])) {
+                    $params = $params[$n];
+                } else {
+                    return $defaultValue;
+                }
+            }
+            return $params;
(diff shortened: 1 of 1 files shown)


cebe:
related discussions: #8351 #8100

cebe:
> Tests pass? 	yes (also used for months in large live environment)

there should be tests covering the new feature too. However this is only worth the effort if we decide to add this feature.

If we add it, it should be available for both, post and get.

cebe:
why did you close it?

PR #11187, opened 2016-03-25
documentation snippet for createTableMigration.php

Faryshta:
Small documentation snippet

### DIFF
--- a/framework/views/createTableMigration.php
+++ b/framework/views/createTableMigration.php
@@ -12,8 +12,14 @@ echo "<?php\n";
 
 use yii\db\Migration;
 
+/**
+ * Handles the creation and droping for table `<?= $table ?>` in the database.
+ */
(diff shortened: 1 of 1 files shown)


framework/views/createTableMigration.php
```diff
@@ -12,8 +12,14 @@
...
 use yii\db\Migration;
 
+/**
+ * Handles the creation and droping for table `<?= $table ?>` in the database.
+ */
 class <?= $className ?> extends Migration
 {
+    /**
+     * @inheritdoc
+     */
```
cebe: these are not needed imo. `@inheritdoc` is too much noise in migration files.

Faryshta: The same doc blocks are used for assets, models and modules generated by gii.

samdark: It won't hurt...

PR #11188, opened 2016-03-25
Correct rectangle size for Captcha background

alsopub:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues |  |

### DIFF
--- a/framework/captcha/CaptchaAction.php
+++ b/framework/captcha/CaptchaAction.php
@@ -275,7 +275,7 @@ class CaptchaAction extends Action
             (int) ($this->backColor % 0x10000 / 0x100),
             $this->backColor % 0x100
         );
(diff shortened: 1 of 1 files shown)


cebe:
can you briefly describe the problem? also we need a CHANGELOG entry to merge it.

alsopub:
This is not a bug.
Function "imagefilledrectangle" gets rectangle as "x1, y1, x2, y2", so to fill rectangle on entire image we need to pass 0, 0, width-1, height-1. Now passed 0, 0, width, height, that 1 pixel more than needed on both axis.

samdark:
OK. `imagefilledrectangle` uses coordinates and these are 0-indexed. The fix is valid.

samdark:
Merged. Thank you.

PR #11216, opened 2016-03-28
Addition argument for generate absolute url

zlakomanoff:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |


cebe:
You can already do it this way: `Html::a('text', Url::to($url, true))`

cebe:
updated the docs about it

PR #11233, opened 2016-03-30
Fix LV translation in compare validator

DMGPage:
Compare validator error message translations are incorrect in Latvian.
{compareValue} is used instead of {compareValueOrAttribute}


samdark:
Thanks!

PR #11234, opened 2016-03-30
Update UrlRule.php

michalkniotekappdate:
| Q | A |
| --- | --- |
| Is bugfix? | yes/no |
| New feature? | yes/no |
| Breaks BC? | yes/no |
| Tests pass? | yes/no |

Probably this is an error in all id tokens regexps. Why there is "," char?

### DIFF
--- a/framework/rest/UrlRule.php
+++ b/framework/rest/UrlRule.php
@@ -102,7 +102,7 @@ class UrlRule extends CompositeUrlRule
      * @see patterns
      */
     public $tokens = [
-        '{id}' => '<id:\\d[\\d,]*>',
+        '{id}' => '<id:\\d[\\d]*>',
     ];
     /**
      * @var array list of possible patterns and the corresponding actions for creating the URL rules.


samdark:
Why shouldn't it be there?

michalkniotekappdate:
GET api-user.awesome.dev/v1/users/1,2
such request pass with id=1 and 2 is neglected and return response for id=1. Here is my question why? 

When there is no "," character such request will return 404 response.
In my opinion, should return badrequest excepcion, but even 404 is better than correct response for id=1.

SilverFire:
I'm closing this PR because target branch is yiisoft:2.1-date-format-consistency insetead of yiisoft:master. Please, recreate your PR in master branch.

tunecino:
@michalkniotekappdate that is needed for resources mapping models with composite primary keys. otherwise how would you load them ? check [this](https://github.com/yiisoft/yii2/blob/master/framework/rest/Action.php#L88-L93)

Issue #11245, opened 2016-03-31
RBAC is missing getRolesByRole()


michaelarnauts:
Rbac makes it possible to specify a role that has a parent role. There is however no function `getRolesByRole($roleName)` to get it's children.

This function is almost identical to the function `getPermissionsByRole($roleName)`.

in `yiisoft/yii2/rbac/DbManager.php`:
```php
    /**
     * @inheritdoc
     */
    public function getRolesByRole($roleName)
    {
        $childrenList = $this->getChildrenList();
        $result = [];
        $this->getChildrenRecursive($roleName, $childrenList, $result);
        if (empty($result)) {
            return [];
        }
        $query = (new Query)->from($this->itemTable)->where([
            'type' => Item::TYPE_ROLE,
            'name' => array_keys($result),
        ]);
        $roles = [];
        foreach ($query->all($this->db) as $row) {
            $roles[$row['name']] = $this->populateItem($row);
        }
        return $roles;
    }
```
in `yiisoft/yii2/rbac/ManagerInterface.php`:
```php
    /**
     * Returns all child roles that the specified role represents.
     * @param string $roleName the role name
     * @return Role[] all child roles that the role represents. The array is indexed by the role names.
     */
    public function getRolesByRole($roleName);

```

samdark:
I think it should something like `getChildRoles()`.

samdark:
What do you need it for?

michaelarnauts:
I have a admin-interface to manage the Users, Roles and Permissions where I would like to be able to get the assigned child Roles of a specified Role. This already exists for Permissions (`getPermissionsByRole`), but it doesn't for Roles.

samdark:
OK. Makes sense.

samdark:
Do you have time for a pull request?

michaelarnauts:
I've added a PR, but you should check to be sure. I haven't tested the code of `PhpManager`.

PR #11033, opened 2016-03-05
Fix #10993 yii\rbac\DbManager::removeAllRules()

lav45:
Fix #10993

PR #11250, opened 2016-03-31
Add getRolesByRole() to RBAC to fetch the child Roles of a Role

michaelarnauts:
See #11245.

I haven't tested the `getRolesByRole` of `PhpManager`, but it is almost identical to the `getPermissionsByRole`, so I guess it should work.

### DIFF
--- a/framework/rbac/DbManager.php
+++ b/framework/rbac/DbManager.php
@@ -495,6 +495,28 @@ class DbManager extends BaseManager
     /**
      * @inheritdoc
      */
+    public function getRolesByRole($roleName)
+    {
+        $childrenList = $this->getChildrenList();
+        $result = [];
+        $this->getChildrenRecursive($roleName, $childrenList, $result);
+        if (empty($result)) {
+            return [];
+        }
+        $query = (new Query)->from($this->itemTable)->where([
+            'type' => Item::TYPE_ROLE,
+            'name' => array_keys($result),
+        ]);
+        $roles = [];
+        foreach ($query->all($this->db) as $row) {
+            $roles[$row['name']] = $this->populateItem($row);
+        }
+        return $roles;
+    }
+
+    /**
+     * @inheritdoc
+     */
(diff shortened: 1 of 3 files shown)


samdark:
Could you create a unit test for it?

mikk150:
```php
/**
     * [getChildRoles description]
     * @param  string $parentRole parent role
     * @return yii\rbac\Role[] roles
     */
    public function getChildRoles($parentRole)
    {
        $roles = [];

        $childRoles = array_keys(array_filter($this->childrenList, function ($childrens) use ($parentRole) {
            return in_array($parentRole, $childrens);
        }));
        return array_filter($this->roles, function ($role) use ($childRoles) {
            return in_array($role->name, $childRoles);
        });

        return $roles;
    }
```
should be the same but without query?

michaelarnauts:
@mikk150 `getRolesByRole` is recursive, `getChildRoles` is not. The function is working the same as the `getPermissionsByRole`.

samdark:
@michaelarnauts so how about a unit test?

LAV45:
Maybe we should add phpunit tests?

samdark:
We definitely should.

LAV45:
That did not work then as there => https://github.com/yiisoft/yii2/pull/11033/files

samdark:
?

mikk150:
@michaelarnauts maybe getting it not recursive is a good thing(I actually need it to be not recursive)
If it is not recursive you could make admin panel out of it and show exactly where particular role comes

michaelarnauts:
@mikk150 indeed. I've forgot about this PR. We've used the non-recursive approach (by just using `getChildRoles()` for the admin panel. Then I then wanted to use this to check if a user has a specific Role, but aparently, you can also use the `checkAccess()` for that, so the code in this requrst is not needed anymore.

LAV45:
Походу психанул.

mikk150:
Maybe I'll open PR for getChildRoles()

PR #11256, opened 2016-04-01
Fix #11252 use strict comparison on hasAttribute()

### LINKED ISSUE #11252: ActiveRecord::hasAttribute(0) return true
```php
class Project extends \yii\db\ActiveRecord 
{
    {
        return 'projects';
    }
}

$Project = Project::findOne(1);
var_dump($Project->hasAttribute(0));
```
it should be false
### What do you get instead?
it returned true

### Additional info
The reason because `in_array($name, $this->attributes())` in `hasAttribute()` has returned true
[http://stackoverflow.com/questions/13846769/php-in-array-0-value](http://stackoverflow.com/questions/13846769/php-in-array-0-value)

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.7
| PHP version      | 7.0.2
| Operating system | MacOSX 10.11 El Capitan

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11252 |

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -419,7 +419,8 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      */
     public function hasAttribute($name)
     {
-        return isset($this->_attributes[$name]) || in_array($name, $this->attributes());
+        return isset($this->_attributes[$name])
+            || in_array($name, $this->attributes(), true);
     }
 
     /**


SilverFire:
It is not the only place, where `is_array` is used for similar purpose and may need strict flag. I think it will be good to fix this in other parts of code.

@yiisoft/core-developers should we?

samdark:
Yes.

SilverFire:
@Faryshta would you like to expand? :)

samdark:
Be careful with string/int/boolean though.

Faryshta:
@SilverFire @samdark its quite the opposite, the strict flag is used in every other part of the models code but here.

https://github.com/yiisoft/yii2/blob/da82ff3fb1625597706d9da3c76ca5a0a1ee7ad5/framework/base/Model.php here `in_array` is used 5 times, 4 use the strict flag and the last one uses an string conversion by concatenation.

cebe:
quite unlikely in general but could there be column names like '1', '2', '3', '4' ... ?

cebe:
okay, given http://stackoverflow.com/questions/13846769/php-in-array-0-value this should totally be fixed...

cebe:
Merged, thank you!

PR #11262, opened 2016-04-02
Enable use of yii2 inside of PHAR packaged console applications

hiqsol:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

This fix doesn't solve the PHAR question completely but fixes the main stopper.
Unfortunately `realpath` is not suitable for `phar://` pathes [PHPBUG#52769](https://bugs.php.net/bug.php?id=52769)

### DIFF
--- a/framework/base/Module.php
+++ b/framework/base/Module.php
@@ -217,7 +217,7 @@ class Module extends ServiceLocator
     public function setBasePath($path)
     {
         $path = Yii::getAlias($path);
-        $p = realpath($path);
+        $p = strncmp($path, 'phar://', 7) === 0 ? $path : realpath($path);
         if ($p !== false && is_dir($p)) {
             $this->_basePath = $p;
         } else {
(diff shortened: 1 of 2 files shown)


framework/base/Module.php
```diff
@@ -217,7 +217,7 @@ public function getBasePath()
     public function setBasePath($path)
     {
         $path = Yii::getAlias($path);
-        $p = realpath($path);
+        $p = substr($path, 0, 7) === 'phar://' ? $path : realpath($path);
```
samdark: What if the path has national symbols?

samdark: Better to use `strncmp($path, 'phar://', 7) === 0`. Otherwise everything's OK.

hiqsol:
Updated PR according to @samdark's notice.

cebe:
Looks good to me, need a changelog line.

SilverFire:
Thank you!

PR #11285, opened 2016-04-06
Security enhancements

samdark:
Based on @tom-- #11146 with `fread` commit applied on top. I've made some changes:
- Removed PHP version checks since both are met by default because Yii requires 5.4.0+.
- Limit PHP fread buffer in order to prevent entropy wasting.
- Fixed incorrect bytes to read calculation.
- Added more notes explaining decisions.

| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues | #11111, #11146 |

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -51,6 +51,10 @@ Yii Framework 2 Change Log
 - Enh #11139: `yii\validators\EachValidator` injects specific attribute value in error message parameters (silverfire)
 - Enh #11187: migrate command now generates phpdoc for table migrations (Faryshta)
 - Enh #11254: Added ability to attach RBAC rule using class name (mdmunir)
+- Enh #11285: `yii\base\Security` enhancements (tom--, samdark)
+  - Avoid reading too much from `/dev/urandom` and `/dev/random` not to waste entropy.
+  - Pefer `/dev/random` to `/dev/urandom` when running on FreeBSD.
+  - Better RNG performance.
 - Enh: Added `StringHelper::countWords()` that given a string returns number of words in it (samdark)
 - Enh #11207: migrate command can create foreign keys. (faryshta)
 - Chg: HTMLPurifier dependency updated to `~4.6` (samdark)
--- a/framework/base/Security.php
+++ b/framework/base/Security.php
@@ -427,7 +427,7 @@ class Security extends Component
         return false;
     }
 
-    private $_libreSSL;
+    private $_useLibreSSL;
     private $_randomFile;
 
     /**
--- a/framework/base/Security.php
+++ b/framework/base/Security.php
@@ -31,6 +31,10 @@ use Yii;
 class Security extends Component
 {
     /**
+     * Buffer size for reading from /dev/random or /dev/urandom
+     */
+    const RANDOM_FILE_BUFFER = 8;
+    /**
      * @var string The cipher to use for encryption and decryption.
      */
     public $cipher = 'AES-128-CBC';
@@ -433,6 +437,7 @@ class Security extends Component
      *
      * @param integer $length the number of bytes to generate
      * @return string the generated random bytes
+     * @throws InvalidParamException if wrong length is specified
      * @throws Exception on failure.
      */
     public function generateRandomKey($length = 32)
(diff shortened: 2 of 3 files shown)


tests/framework/base/SecurityTest.php
```diff
@@ -810,6 +811,55 @@ public function testGenerateRandomKey()
...
+            "DIRECTORY_SEPARATOR",
+            "ini_get('open_basedir')",
+        ];
+        if (DIRECTORY_SEPARATOR === '/') {
+            $tests[] = "sprintf('%o', lstat(PHP_OS === 'FreeBSD' ? '/dev/random' : '/dev/urandom')['mode'] & 0170000)";
+            $tests[] = "bin2hex(file_get_contents(PHP_OS === 'FreeBSD' ? '/dev/random' : '/dev/urandom', false, null, 0, 8))";
+        }
+        foreach ($tests as $i => $test) {
+            $result = eval('return ' . $test . ';');
+            fwrite(STDERR, sprintf("%2d %s ==> %s\n", $i + 1, $test, var_export($result, true)));
```
klimov-paul: Why this is performed?
Currently it outputs parse info direct into the shell while running the unit test.

samdark: It's performed to check performance of RNG. We may comment.

klimov-paul: That whoould be nice, since following output is annoying:

urandom

samdark: https://github.com/yiisoft/yii2/commit/5f7cd1cbc49ea6e9f147119d6919787e2f383fc8

tom--:
Easily annoyed :þ  

What I find annoying is how the **portability** requirement makes this method so hard to test. When you multiply these lists as vectors: operating systems, OS versions, PHP versions, configuration of PHP extensions, and extension versions there are a lot of cases in which this method will return a value that we claim is random enough for security work.

In each of those cases we should be confident that we produce actually random numbers, with sufficient performance, without disturbing system resources. Our tests do not accomplish this.

I considered different ways to implement this test, e.g. in a console command. But how would I commit that in yiisoft/yii2? It seems yii2 doesn't have a convention or framework for this kind of thing.

We also need empirical testing of its output with things like Fourmilab ent, dieharder and TestU01. For this you need a command that writes a byte stream to stdout.
Travis isn't the answer. Manual testing is acceptable since the hard part is setting up the test environments.

What do you suggest?

framework/CHANGELOG.md
```diff
@@ -51,6 +51,10 @@ Yii Framework 2 Change Log
 - Enh #11139: `yii\validators\EachValidator` injects specific attribute value in error message parameters (silverfire)
 - Enh #11187: migrate command now generates phpdoc for table migrations (Faryshta)
 - Enh #11254: Added ability to attach RBAC rule using class name (mdmunir)
+- Enh #11285: `yii\base\Security` enhancements (tom--, samdark)
+  - Avoid reading too much from `/dev/urandom` and `/dev/random` not to waste entropy.
```
tom--: This sentence should not be mine :þ

I don't really know what it means to waste entropy and I certainly can't defend the proposition that this change does wastes less entropy than anything else.

framework/base/Security.php
```diff
@@ -31,6 +31,10 @@
 class Security extends Component
 {
     /**
+     * Buffer size for reading from /dev/random or /dev/urandom
+     */
+    const RANDOM_FILE_BUFFER = 8;
```
tom--: If you docblock like this then it becomes part of the published API for programmers. I don't think that's wise. I would put it at line ~430 and put the comment in the method where you use the const.

framework/CHANGELOG.md
samdark: Umm, I'll re-phrase it to `reading more bytes than needed`.

samdark: https://github.com/yiisoft/yii2/commit/2c611b410e6ac85256c8ccf434497d635fd3338d

framework/base/Security.php
samdark: https://github.com/yiisoft/yii2/commit/6a80a132e2e86181620aecacc28d07d051e5454a

samdark:
As for testing, we may extract different RNG sources into separate protected methods so these could be tested separately.

framework/base/Security.php
```diff
@@ -472,71 +479,61 @@ public function generateRandomKey($length = 32)
...
+                    // By default PHP buffer size is 8192 bytes which causes wasting
+                    // more entropy that we're actually using. Therefore setting it to
+                    // lower value.
+                    if (function_exists('stream_set_read_buffer')) {
+                        stream_set_read_buffer($this->_randomFile, self::RANDOM_FILE_BUFFER);
+                    }
+                    // stream_set_read_buffer() isn't implemented on HHVM
+                    if (function_exists('stream_set_chunk_size')) {
+                        stream_set_chunk_size($this->_randomFile, self::RANDOM_FILE_BUFFER);
+                    }
```
tom--: Is it the case that if one does not call `stream_set_read_buffer($stream, ...)` then  `fread($stream, 1)` causes PHP to actually transfer 8 KiB of data from the stream into a PHP buffer?

If so, for what kinds of streams is this the case? Surely not all?

samdark: Yes. That's for file streams at least. The function sets buffer size for a file pointer specified.

tom--: Again I'm not keen on "wasting entropy" language. Suggest instead for the comment:

> Reduce PHP stream buffer from default 8192 bytes to optimize data transfer from the random device for smaller values of $length. This also helps to keep future randoms out of user memory space.

samdark: https://github.com/yiisoft/yii2/commit/0c6c1eebd1b46ecdc2bfc13d5cb370c2c7ed413e

tom--:
@cebe Does the method fail safe if used statically? Or is it not possible?

samdark:
It should not be used statically.

PR #11287, opened 2016-04-06
Enh: Support for useTablePrefix parameter in migrate command

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11166 |


framework/CHANGELOG.md
```diff
@@ -52,7 +52,8 @@ Yii Framework 2 Change Log
 - Enh #11187: migrate command now generates phpdoc for table migrations (Faryshta)
 - Enh #11254: Added ability to attach RBAC rule using class name (mdmunir)
 - Enh: Added `StringHelper::countWords()` that given a string returns number of words in it (samdark)
-- Enh #11207: migrate command can create foreign keys. (faryshta)
+- Enh #11207: migrate command can create foreign keys. (Faryshta)
```
Faryshta: this was a minimal change, just one letter from lower case to uppercase, it can be undone without issue.

samdark: That's OK.

framework/console/controllers/MigrateController.php
```diff
@@ -281,6 +313,7 @@ protected function generateMigrationSourceCode($params)
             'table' => $table,
             'fields' => $this->fields,
             'foreignKeys' => $this->foreignKeys,
+            'tName' => [$this, 'generatedTableName'],
```
samdark: Instead of passing around function you can execute it and pass result around. That would be much cleaner.

Faryshta:
I added in total 3 more tests, 1 for `create_test` with foreign keys and `useTablePrefix`, one for `add_columns_to_test` with foreign keys and one for `add_columns_to_test` with foreign keys and `useTablePrefix`

samdark:
Thanks!

Issue #9091, opened 2015-07-13
Url caching has error


genichyar:
If I use following url rules
```php
[
    'pattern' => '',
    'route' => 'paracms/structure/frontend/page/view',
    'defaults' => ['slug' => 'index'],
],
'page/<slug>' => 'paracms/structure/frontend/page/view',
```
and then run code
```php
echo Url::to('/paracms/structure/frontend/page/view', ['slug' => 'services']);
echo '<br>';
echo Url::to('/paracms/structure/frontend/page/view', ['slug' => 'index']);
```
I have got
```
/page/services/
/page/index/
```
But I expeсted
```
/page/services/
/
```

This problem rises due to https://github.com/yiisoft/yii2/commit/ebc089e6c26ba85642e5b32eaae017c6b637be79

When I call ```Url::to('/paracms/structure/frontend/page/view', ['slug' => 'services']);``` rule № 2 was cached and then it was used when I call ```Url::to('/paracms/structure/frontend/page/view', ['slug' => 'index']);``` instead of rule № 1.

klimov-paul:
See #8205
It seems 'create URL cache' feature should be reverted (which whould be a pity) or made optional.
In present state you should be able to walk around the issue exluding 'index' from second rule:
```php
[
    'pattern' => '',
    'route' => 'paracms/structure/frontend/page/view',
    'defaults' => ['slug' => 'index'],
],
'page/<slug:((?!index).)+>' => 'paracms/structure/frontend/page/view',
```

genichyar:
@klimov-paul, thank you!

I use same regexp as temporary solution: `page/<slug:((?!index).)*>`. But it doesn't work properly if slug has word "index". For example: `index1`, `indexation`, etc.

> It seems 'create URL cache' feature should be reverted (which whould be a pity) or made optional.

I think 'create URL cache' was good idea. If there is no issue with this feature except my issue I propose to not revert this feature and add an ability to make inversed regexp in url.

For example:
`page/<slug:!index>` — All except "index"
`page/<slug:!(index|sitemap|contacts)>` — All except "index", "sitemap", "contacts"

cebe:
there is no need for complex inverse regex, we can simply skip the cache if we detect situations which we can not cache.

checked in a fix for this.

PR #11288, opened 2016-04-06
(fix) UrlManager not cached rules correctly.

zalatov:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues |  |

It's appears when UrlManager processing rule with defaults or parsingOnly attributes.

### DIFF
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -335,14 +335,9 @@ class UrlManager extends Component
             }
 
             if ($url === false) {
-                $cacheable = true;
                 foreach ($this->rules as $rule) {
-                    if (!empty($rule->defaults) && $rule->mode !== UrlRule::PARSING_ONLY) {
-                        // if there is a rule with default values involved, the matching result may not be cached
-                        $cacheable = false;
-                    }
                     if (($url = $rule->createUrl($this, $route, $params)) !== false) {
-                        if ($cacheable) {
+                        if (false === (!empty($rule->defaults) && $rule->mode !== UrlRule::PARSING_ONLY)) {
                             $this->_ruleCache[$cacheKey][] = $rule;
                         }
                         break;


zalatov:
I don't undestand, why test is failed:
https://github.com/yiisoft/yii2/blob/master/tests/framework/web/UrlManagerTest.php#L169

Why it's asserting "/test/"?

Faryshta:
lets go to the previous test.

`/test/page/services` by the rules of the UrlManager will be converted too the route `['frontend/page/view', 'slug' => 'services']` because it follows the pattern  `<baseUrl>/page/<slug>` where `$baseUrl = 'test'` and `<slug>='services'` So the pattern is `page/<slug>`

Now there is another pattern defined on the rules which is `'pattern' => ''` which means if you go to `<baseUrl>` it will direct you to `'frontend/page/view'` action with parameters `['slug' => 'index']`.

That means that the opposite is true, if I want to build a route to  `'frontend/page/view'` action with parameters `['slug' => 'index']` i don't need to add anything more than the `<baseUrl>`

zalatov:
I found cache problem in UrlManager.

For example. list of rules:
1. ['page/view']
2. ['page/index'],
3. ['/user/login', {defaults=>[item]}, {parsingOnly=true}]
4. ['page/update'],

In step 3 flag "cachable" is set from true to false.
But rule don't using for generation.

Thereby, all rules after this case are not caching.

In fact, this list are caching never:
1. ['routepath', 'defaults' => ['slug' => 'index', 'mode' => PARSING_ONLY]
2. ['user/user/login'],
3. ['page/page/view'],
4. ['catalog/product/view'],
5. and more...

And when UrlManager::createUrl() calls many times (100, 200, 300), it's hard slowing page generation.

Faryshta:
@zalatov that behavior was added because the bug #9091

zalatov:
What I need to do next?

Faryshta:
@zalatov i honestly don't know, i can't figure out a good way to solve it, you should ask @cebe about it.

cebe:
@zalatov could you write a unit test to demonstrat the issue?

zalatov:
@cebe, I can write test, but that it's broke current logiс when default value is set.
@Faryshta wrote "that behavior was added because the bug #9091"

rob006:
see #13350

PR #11289, opened 2016-04-06
Multidimensional Request Superglobals

E-conomy:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues |  |
- Ability to read out multidimensional GET POST Superglobals
- Keeps current functionality by extending it
- Works by passing an array parameter instead of the usual string

Usage:

$_GET['item']['description'][0] now becomes:

Yii::$app->request->get(['item', 'description', '0']);

$_POST['test'] Stays:

Yii::$app->request->post('test');

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -440,7 +440,18 @@ class Request extends \yii\base\Request
     {
         $params = $this->getBodyParams();
 
-        return isset($params[$name]) ? $params[$name] : $defaultValue;
+        if (is_array($name)) {
+            foreach ($name as $n) {
+                if (isset($params[$n])) {
+                    $params = $params[$n];
+                } else {
+                    return $defaultValue;
+                }
+            }
+            return $params;
+        } else {
+            return isset($params[$name]) ? $params[$name] : $defaultValue;
+        }
     }
 
     /**
@@ -516,7 +527,18 @@ class Request extends \yii\base\Request
     {
         $params = $this->getQueryParams();
 
-        return isset($params[$name]) ? $params[$name] : $defaultValue;
+        if (is_array($name)) {
+            foreach ($name AS $n) {
+                if (isset($params[$n])) {
+                    $params = $params[$n];
+                } else {
+                    return $defaultValue;
+                }
+            }
+            return $params;
+        } else {
+            return isset($params[$name]) ? $params[$name] : $defaultValue;
+        }
     }
 
     private $_hostInfo;


SilverFire:
Thank your for your effort and contribution.

This API confuses me. Looking on it intuitively I think that it must return array of `$_POST['item']`, `$_POST['description']` and `$_POST[0]`.

Is this enhancement is better than the following code:
```php
Yii::$app->request->get('item')['description'][0];
```

MercadoVentures:
Yes, I would consider it's better. If say:

`Yii::$app->request->get('item')`

was null, then it would lead to a PHP warning when trying to reach Yii::$app->request->get('item')['description'].

More code would be needed to make this work, for example isset(Yii::$app->request->get('item')['description']) before trying to read out the multidimensional values.

Now we can simply echo it, because by default it returns NULL, without applying isset() first for each value we want to read out. Saving a whole lot of code.

tunecino:
agree with @SilverFire because in those 2 we can see the **hierarchy** :
```php
$_GET['item']['description'][0]
Yii::$app->request->get('item')['description'][0];
```

while doing this way:
```php
Yii::$app->request->get(['item', 'description', '0']);
```
I see it more like:
```php
Yii::$app->request->get(['item1', 'item2', 'item3']);
```
and I would expect items to be at same level which may be confusing.

samdark:
I don't like the syntax in this proposal but idea itself is OK because of warnings and default value.

SilverFire:
```php
ArrayHelper::getValue(Yii::$app->request->get('item'), ['description', '0']);
// or
ArrayHelper::getValue(Yii::$app->request->get('item'), 'description.0');
```

SilverFire:
I can see almost no profit in this kind of enhancement because it makes `Request` object more complex, maybe adds dependency on `ArrayHelper` and does not bring more convenience.

Usually, you do not need to extract some deeeeeply nested value from the request variables, because you work in terms of models. If this use case is popular is your project - 90% that you are doing something wrong.

MercadoVentures:
@SilverFire 
Thanks for your response. We currently use Cassandra as a database system, which has no activerecord implementation, therefore no model implementation. For us, if I'm not mistaken, there really was no other choice then to directly call on input.

tunecino:
what about a new different method ? like:
```php
Yii::$app->request->parse('get', 'item.description.0')
```

klimov-paul:
Name like 'item.description' is a totally valid for the input name. You'll be unable to get such input from request with this syntax.

samdark:
I agree with @SilverFire and @klimov-paul. It's not common enough to be in the core framework and could be implemented in many other ways such as using `item.description` as names (good for your case since custom form is used).

PR #11291, opened 2016-04-06
Fixes #11283: `ActiveRecord::unlink()` is not setting FK to `null` before deleting itself anymore

### LINKED ISSUE #11283: ActiveRecord::unlink() is unnecessarily nulling foreign key property
Hello,
I have problem with foreach on line 1307 in BaseActiveRecord.php, which is nulling foreign key property of parent model.
I'm using custom delete behavior, where I cancel "real" model deletion and I just set the "deleted" attribute to "true" instead.

Important part of my custom delete behavior code:
```php
public function beforeDelete($event) {
    ...
    $owner->{$this->deletedAttribute} = true;
    $owner->save();
    $event->isValid = false;
}
```

So imho it would be correct to keep foreign key value, but only when "$delete" parameter is set to "true", so it won't break the other functionality.

**Original code:**
```php
foreach ($relation->link as $a => $b) {
    $model->$a = null;
}
$delete ? $model->delete() : $model->save(false);
```

**Altered code:**
```php
if (!$delete) {
    foreach ($relation->link as $a => $b) {
        $model->$a = null;
    }
}
$delete ? $model->delete() : $model->save(false);
```

When programmer want to unlink without delete, it will do the nullification of foreign key, which is OK. But when he wants to delete foreign model, it will keep the foreign key value. This doesn't matter when you do "real" delete (foreign model will be deleted anyway), but when you override deletion in behavior by updating "deleted" attribute, you keep data in database and when the non-nullable foreign key is nullified, the SQL update crashes.

Cheers and thanks.

~PMan

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.6
| PHP version      | 5.5.30
| Operating system | Win 8.1

samdark:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? |  |
| Fixed issues | #11283 |

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -1304,10 +1304,14 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
             $p1 = $model->isPrimaryKey(array_keys($relation->link));
             $p2 = static::isPrimaryKey(array_values($relation->link));
             if ($p2) {
-                foreach ($relation->link as $a => $b) {
-                    $model->$a = null;
+                if ($delete) {
+                    $model->delete();
+                } else {
+                    foreach ($relation->link as $a => $b) {
+                        $model->$a = null;
+                    }
+                    $model->save(false);
                 }
-                $delete ? $model->delete() : $model->save(false);
             } elseif ($p1) {
                 foreach ($relation->link as $a => $b) {
                     if (is_array($this->$b)) { // relation via array valued attribute
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -53,6 +53,7 @@ Yii Framework 2 Change Log
 - Enh #11254: Added ability to attach RBAC rule using class name (mdmunir)
 - Enh: Added `StringHelper::countWords()` that given a string returns number of words in it (samdark)
 - Enh #11207: migrate command can create foreign keys. (faryshta)
+- Chg #11283: `ActiveRecord::unlink()` is not setting FK to `null` before deleting itself anymore (samdark)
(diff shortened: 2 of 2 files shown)


framework/db/BaseActiveRecord.php
```diff
@@ -1304,10 +1304,14 @@ public function unlink($name, $model, $delete = false)
             $p1 = $model->isPrimaryKey(array_keys($relation->link));
             $p2 = static::isPrimaryKey(array_values($relation->link));
             if ($p2) {
-                foreach ($relation->link as $a => $b) {
-                    $model->$a = null;
+                if ($delete) {
+                    $model->delete();
```
klimov-paul: We should use return value of `delete()` - it will return `false` in case delete is canceled.

samdark: Probably but it wasn't like that and it is not part of the issue.

klimov-paul: This change produce a possible bug for the DBMS without foreigh keys support, such as SQLite.
The link fields will remain pointing to the unexisting record!

klimov-paul: The issue #11283 was about use case, when `delete()` method is canceled at `beforeDelete()` stage and no actual record removal was performed.

samdark: Why? The only change I've introduced is avoiding changing fields of the model if exactly this model is deleted right away.

klimov-paul: Yes, you are right, since `$relation` is not affected at this program flow it is correct.

PR #11292, opened 2016-04-06
`MigrateController` refactored avoiding internal state changing

klimov-paul:
| Q | A |
| --- | --- |
| Is bugfix? | depends |
| New feature? | no |
| Breaks BC? | yes: changes signature for the new methods intoduced at 2.0.7 |
| Tests pass? | yes |

Recent changes at `MigrateController`, which allows autogenerating migration source code, are done in the way internal controller state is changed during the process.

Thus in case you invoke `actionCreate()` twice at the same program flow - it will end with an error.

This refactoring allows `fields` paramter to be left intact and removes redundant internal field `foreignKeys`.

### DIFF
--- a/framework/console/controllers/MigrateController.php
+++ b/framework/console/controllers/MigrateController.php
@@ -96,12 +96,6 @@ class MigrateController extends BaseMigrateController
      */
     public $db = 'db';
 
-    /**
-     * @var array columns which have a foreign key and their related table.
-     * @since 2.0.8
-     */
-    protected $foreignKeys = [];
-
 
     /**
      * @inheritdoc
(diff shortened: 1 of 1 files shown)


Faryshta:
Can you use #11287 as your starting point?

framework/console/controllers/MigrateController.php
```diff
@@ -268,28 +264,36 @@ protected function generateMigrationSourceCode($params)
...
     /**
      * Parse the command line migration fields
+     * @return array parse result with following fields:
+     *
+     * - fields: array, parsed fields
+     * - foreignKeys: array, detected foreign keys
+     *
      * @since 2.0.7
      */
     protected function parseFields()
```
Faryshta: Related to #11287  since this method no longer needs to get called during `beforeAction` can we make it take a parameter `$table` to behave similar to https://github.com/yiisoft/yii2/pull/11287/files#diff-952bc7045bade5b9e9c44621c5f93c19R335 ?

samdark:
@klimov-paul can you merge from master?

klimov-paul:
Done.

samdark:
Merged.

klimov-paul:
Build fixed by 8e50598515b9137f172eb02fa601e2b06db68a62

PR #11293, opened 2016-04-06
Update phpdoc.

wirwolf:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

If i extend this class

```
class UploadedFile extends \yii\web\UploadedFile

```

and i use my function, IDE not see my method

### DIFF
--- a/framework/web/UploadedFile.php
+++ b/framework/web/UploadedFile.php
@@ -74,7 +74,7 @@ class UploadedFile extends Object
      * @param \yii\base\Model $model the data model
      * @param string $attribute the attribute name. The attribute name may contain array indexes.
      * For example, '[1]file' for tabular file uploading; and 'file[1]' for an element in a file array.
-     * @return null|UploadedFile the instance of the uploaded file.
+     * @return null|$this the instance of the uploaded file.
      * Null is returned if no file is uploaded for the specified model attribute.
      * @see getInstanceByName()
      */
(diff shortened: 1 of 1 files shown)


klimov-paul:
`$this` can not be used as return type for static methods.

cebe:
should be `static` instead.

samdark:
Hmm. Seems PhpStorm isn't good with `static[]`.

cebe:
there are a few bugs related to this. not sure if this specific case is covered there though https://youtrack.jetbrains.com/issues/WI?q=return+static+%23Open+

mdmunir:
`@return $this` does not work in netbeans if method defined in trait.
screenshot from 2016-04-10 10 54 36
after `limit()`
screenshot from 2016-04-10 10 55 35
`limit()` is defined in `QueryTrait` https://github.com/yiisoft/yii2/blob/master/framework/db/QueryTrait.php#L373

Should i create new issue?

cebe:
report it to Netbeans.

mdmunir:
@cebe 
> report it to Netbeans.

can solve using `@return $this|static`

wirwolf:
I use inetlij idea

samdark:
Closed since IDEs support isn't good for now.

Issue #10998, opened 2016-03-01
Getter function for View::$title so that Pjax requests don't change the page title.


Bhoft:
As already mentioned yesterday [in this old issue](https://github.com/yiisoft/yii2/issues/662) I would prefer to have a View::getTitle() function which returns the Views $title attribute.

And that this function is used where the title of the View is accessed e.g. in class Pjax.

Easy example:
in my layout/main.php I have this code to add the application name to the page title:

```
<title><?= $this->title ? Html::encode(Yii::$app->name." | $this->title") : (Yii::$app->name.' | Home') ?></title>
```

And in the views something like:

```
$this->title = Yii::t('app', MyModel::label(2)); 
// or something else
$this->title = 'Some title';
```

With the code in the layout view file the application name is added to the title => Fine.

BUT on every Pjax request only the normal view title is used (so without the application name) and the page title is changed incorrectly. This is because only the `$view->title` is used.
(code from class Pjax)

```
if ($view->title !== null) {
     echo Html::tag('title', Html::encode($view->title));
}
```

Do you have another suggestion how this issue can be fixed?
I mean other then extend the `yii\web\View` and `yii\widget\Pjax` classes as overwrite/extend them with functions?

Which would be cumbersome because in the pjax class the title is directly echoed in `pjax::init` so an setting of the title before `parent::init()` call would work. I would have to copy and paste the complete init function and change one line `$view->title` to `$view->getTitle()` and have to check for changes on every yii update to also re-do the changes to my extended pjax class.

Of course I could change every view where i set the title and add the application name directly.
But in my opinion a getter (maybe also a setter) function for the title would be the easiest. 

If the View Class has a getTitle() function (and this function would be used instead of $view->title) we could just overwrite the Class via components and extend it with e.g. 

```
public function getTitle()
{
    if ($this->title === null) {
        return Yii::$app->name;
    }
   return  Yii::$app->name .' | '.$this->title;
}   
```

Then the Application Name (or something else) could be easily added to page and wouldn't change after pjax requests.

This would also help if the title comes from the db (e.g. multi language) or would be different for e.g. mobile devices.

Faryshta:
i can't understand your example or how is it changing the title.

can you make an example with copies of your layout and pjax widget?

SilverFire:
@Faryshta pjax is designed to find `<title>` tag inside a response and update document's title [Source](https://github.com/yiisoft/jquery-pjax/blob/master/jquery.pjax.js#L751)

[...]

PR #11294, opened 2016-04-06
adds getter / setter for the title to View to fix issues with the page title after pjax requests

Bhoft:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #10998 |


SilverFire:
I meant to create `getTitle()` in `Pjax` class, not in `View`.

Bhoft:
It wouldn't solve the issue. If a getTitle function will be added to the Pjax Class it also must be added to the View class as well.

SilverFire:
Why? Then you can redeclare `getTitle()` in your own Pjax class and all works as you want

Bhoft:
As i mentioned also on the issue page. 
Of course if i add a getTitle() function to the Pjax and use it it would solve `this` issue. But for external modules which maybe also use the $this->title in their views the page title would still be not correctly set. 

And as I mentioned in the [issue ](https://github.com/yiisoft/yii2/issues/10998#issuecomment-206352347) if I have as an example an dynamic page title in e.g. multi-language and the title prefix differs I would still have multiple places where I have to extend the code to get the correct result.

Issue #11002, opened 2016-03-02
TimestampBehavior and BlameableBehavior cause non-dirty records updated


lacek:
`ActiveRecord::update` is supposed to save only changed attributes.
Suppose we have an ActiveRecord with `TimestampBehavior` attached and there's no dirty attributes.
When `save` or `update` method is called on this record, `beforeSave` method is called and triggers the behavior to update the `updated_at` attribute. Thus update is always performed.

Similarly, `BlameableBehavior` updates `updated_by` attribute if `update` is called on a clean record by a user with ID different from original `updated_by` attribute.

lacek:
A fix that works for me is to override the `evaludateAttributes` method and skips calling parent one when it is beforeUpdate event with no dirty attributes:

```php
    public function evaluateAttributes($event)
    {
        if ($event->name === BaseActiveRecord::EVENT_BEFORE_UPDATE &&
                empty($this->owner->getDirtyAttributes())) {
            return;
        }
        parent::evaluateAttributes($event);
    }
```

samdark:
It changes current behavior a bit but seems to be a good change. @yiisoft/core-developers opinions?

mdmunir:
:+1:

SilverFire:
:+1:

tonydspaniard:
That was actually to be my new suggestion

Faryshta:
I am working on a patch. is the property name `skipUpdateOnCleanAttributes` too long? any other suggestion?

PR #11302, opened 2016-04-07
Enh new property `skipUpdateOnClean` for `AttributeBehavior`

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | kind of |
| New feature? | yes |
| Breaks BC? | yes |
| Fixed issues | #11002 |

Break BC since before this patch `$model->save()` would always execute an update query even when no  attributes were edited.

### DIFF
--- a/framework/behaviors/AttributeBehavior.php
+++ b/framework/behaviors/AttributeBehavior.php
@@ -11,6 +11,7 @@ use Yii;
 use Closure;
 use yii\base\Behavior;
 use yii\base\Event;
+use yii\db\ActiveRecord;
 
 /**
  * AttributeBehavior automatically assigns a specified value to one or multiple attributes of an ActiveRecord
@@ -61,6 +62,7 @@ class AttributeBehavior extends Behavior
      * ```
      */
     public $attributes = [];
+
     /**
      * @var mixed the value that will be assigned to the current attributes. This can be an anonymous function,
      * callable in array format (e.g. `[$this, 'methodName']`), an [[Expression]] object representing a DB expression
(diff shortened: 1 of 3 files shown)


samdark:
I don't think that option is needed.

Faryshta:
@samdark it in case the user wants to configure it or another behavior requires to always trigger an update on the attributes. Example

```php
class Store
{
     public function behaviors()
     {
         return [
             'blame' => [
                    'createdByAttribute' => 'seller_id',
                    'updatedByAttribute' => 'manager_id',
             ],
             'blameSuperVisor' => [
                    'skipUpdateOnClean' => false,
                    'updatedByAttribute' => 'supervisor_id',
             ],
         ];
     }
}
```

Also it helps to keep BC compatibility since now it only requires to change the option to false to have the exact same behavior as before.

If those arguments are still not enough, then just delete the property and simplify the `if` statement

samdark:
OK. It makes sense.

samdark:
@yiisoft/core-developers need a brief review.

framework/behaviors/AttributeBehavior.php
```diff
@@ -77,13 +79,21 @@ class AttributeBehavior extends Behavior
      */
     public $value;
 
+    /**
+     * @var boolean wheter to skip this behavior when the `$owner` has not been
+     * modified
```
SilverFire: `@since 2.0.8`

SilverFire:
We need a CHANGELOG line and a note in UPGRADE.md

Looks good to me

Faryshta:
@samdark @SilverFire done

samdark:
Merged! Thank you!

Issue #7540, opened 2015-03-03
Could Validator::when add Validator object parameter?


callmez:
``` php
// Validator::validateAttributes()
public function validateAttributes($model, $attributes = null)
    {
        if (is_array($attributes)) {
            $attributes = array_intersect($this->attributes, $attributes);
        } else {
            $attributes = $this->attributes;
        }
        foreach ($attributes as $attribute) {
            $skip = $this->skipOnError && $model->hasErrors($attribute)
                || $this->skipOnEmpty && $this->isEmpty($model->$attribute);
            if (!$skip) {
                if ($this->when === null || call_user_func($this->when, $this, $model, $attribute)) { // add validator object parameter
                    $this->validateAttribute($model, $attribute);
                }
            }
        }
    }

// ActiveRecord::rules()
// so we can control validator in function
public function rules()
{
      return [
          [['count'], 'integer', 'when' => function($validator, $model, $attribute) {
              $validator->max = 10;
              $validator->min = 5;
              return true;
          }]
       ];
}
```

qiangxue:
Can you show a use case? What about `whenClient`?

callmez:
actually this scenario can't use validate at client side. 
```php 
class Order{

    public function rules()
	{
		return [
			[['product_id', 'required']],
			// this line can't be use when run NumberValidator::clientValidateAttribute()
			//[['order_num'], 'integer', 'max' => $this->product->max_order_num, 'min' => $this->product->min_order_num],
			[['order_num'], 'integer', 'when' => function($validator, $model, $attribute) {
				$validator->max = $this->product->max_order_num;
				$validator->min = $this->product->min_order_num;
				return $true; // then we can validate the order_num on server side
			}]
		];
	}

	public function getProduct()
	{
		return $this->hasOne(Product::className(), ['id' => 'product_id']);
	}

}

```

dynasource:
related https://github.com/yiisoft/yii2/issues/6242

SilverFire:
Closing as a duplicate

PR #11307, opened 2016-04-07
Generate the javascript to validate the client only when really necessary

thiagotalma:
When we use the `RequiredValidator->when` and `RequiredValidator->whenClient`, we want to decide whether the field is required or not.

However, the generation of validation javascript code does not respect the rule when we have only the `RequiredValidator->when` set. Then the client will always validate the field as required.

In this PR, the validation on the client is only done when there is no `RequiredValidator->whenClient` rule and, having the `RequiredValidator->when` rule, the field is mandatory.

| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #7540 #7546 |

### DIFF
--- a/framework/validators/RequiredValidator.php
+++ b/framework/validators/RequiredValidator.php
@@ -88,6 +88,10 @@ class RequiredValidator extends Validator
      */
     public function clientValidateAttribute($model, $attribute, $view)
     {
+        if ($this->whenClient === null && $this->when !== null && !call_user_func($this->when, $model, $attribute)) {
+            return null;
+        }
+
         $options = [];
         if ($this->requiredValue !== null) {
             $options['message'] = Yii::$app->getI18n()->format($this->message, [


framework/validators/RequiredValidator.php
```diff
@@ -88,6 +88,10 @@ protected function validateValue($value)
      */
     public function clientValidateAttribute($model, $attribute, $view)
     {
+        if ($this->whenClient === null && $this->when !== null && !call_user_func($this->when, $model, $attribute)) {
```
samdark: Doesn't look right to me. It could depend on the value of the model attribute which we don't have till client script validation passes.

SilverFire:
I've tried to apply similar patch in #9983, but it was rejected. Than I've found, that it's not a good idea becuase:

1. Often `when` property is used without `whenClient` just because there is no way to determine on the client side whether this rule is applicable. But the attribute must be validated normally on the client side. 
2. During development, it's always better to see an error in a form and think about it, rather than do not see it and think that validation works as expected.
3. Indeed it breaks BC by changing client-side behavior that is not covered by tests usually and can hurt.

Thank  you for your effort

thiagotalma:
@samdark @SilverFire 

Use case: A field is required only when a condition, regardless of the model, is true.

For example, the field is required if there are already 30 records in the database.

Expected implementation: use the `when` to test if there are already 30 records. No matter what the user input. The field is required.

The way the code is now, the solution ends up making a ugly hack, creating a hidden field for example.

What makes no sense is set by default the field required on the client, regardless of the `when` condition.

It is worth asking the opinion of other members?

Faryshta:
@thiagotalma why not use `whenClient` ?

```php
[['attribute'], 'required', 'whenClient' => self::find()->count() > 30 ? 'return true' : 'return false'],
```

thiagotalma:
@Faryshta 
Thanks for the suggestion, but my problem is more complex than the example I cited.

My rule is something like this:

```PHP
public function rules()
{
    return [
        [
            [
                'contactName',
                'contactEmail',
                'contactPhone',
                'contactMobile',
            ],
            'required',
            'when' => function ($model, $attribute) {
                $form = $model->profile->form;
                if ($form) {
                    switch ($attribute) {
                        case 'contactName':
                            $required = $form->isNnameRequired;
                            break;
                        case 'contactEmail':
                            $required = $form->isEmailRequired;
                            break;
                        case 'contactPhone':
                            $required = $form->isPhoneRequired;
                            break;
                        case 'contactMobile':
                            $required = $form->isMobileRequired;
                            break;
                        default:
                            throw new InvalidParamException('Attributte unknow');
                    }

                    return $required;
                }

                return false;
            }
        ]
    ];
}
```

thiagotalma:
Uggly solution:

```PHP
public function rules()
{
    return [
    
        [
            'contactName',
            'required',
            'when' => $checkRequired,
            'whenClient' => 'function() { return ' . ($checkRequired($this, 'contactName') ? 'true' : 'false') . '; }'
        ],
        [
            'contactEmail',
            'required',
            'when' => $checkRequired,
            'whenClient' => 'function() { return ' . ($checkRequired($this, 'contactEmail') ? 'true' : 'false') . '; }'
        ],
        [
            'contactPhone',
            'required',
            'when' => $checkRequired,
            'whenClient' => 'function() { return ' . ($checkRequired($this, 'contactPhone') ? 'true' : 'false') . '; }'
        ],
        [
            'contactMobile',
            'required',
            'when' => $checkRequired,
            'whenClient' => 'function() { return ' . ($checkRequired($this, 'contactMobile') ? 'true' : 'false') . '; }'
        ]
    ];
}
```

Faryshta:
I think the first solution is just as ugly as the second one.

I would change the database schema since it doesn't follow normalization rules.

Anyway a way to solve it

```php
public function rules()
{
    $rules = [];
    $form = $this->profile->form;
    foreach(['contactName', 'contacEmail', 'contactPhone', 'contactMobile'] as $attribute) {
        $attributeRequired = 'is' . str_replace('contact', '', $attribute) . 'Required';
        if ($form->$attributeRequired) {
            $rules[] = [$attribute, 'required'];
        }
    }

    return $rules;
}
```

Faryshta:
@thiagotalma i simplified my example. check it please.

Issue #10566, opened 2016-01-13
EachValidator && ExistValidator


pkirill99:
Hi all

Yii 2.0.6

in model

```php
public function rules() {
        return [
            [['question', 'answers'], 'required'],
            [['question',], 'integer'],
            [['answers',], 'integer', 'min' => 1, 'on' => self::SCENARIO_SINGLE],
            [
                ['answers'],
                'exist',
                'targetClass'     => Answer::className(),
                'targetAttribute' => ['answers' => 'id', 'question' => 'question_id'],
                'on'              => self::SCENARIO_SINGLE,
            ],
            [['answers'], 'each', 'rule' => ['integer', 'min' => 1], 'on' => self::SCENARIO_MULTI],
            [
                ['answers'],
                'each',
                'rule' => [
                    'exist',
                    'targetClass'     => Answer::className(),
                    'targetAttribute' => ['answers' => 'id', 'question' => 'question_id'],
                ],
                'on'   => self::SCENARIO_MULTI,
            ],
        ];
    }
```

set scenario=SCENARIO_MULTI, run and see:

> The "targetAttribute" property must be configured as a string.

change rule for SCENARIO_MULTI, 
```php

            [
                ['answers'],
                'exist',
                'allowArray'      => true,
                'targetClass'     => Answer::className(),
                'targetAttribute' => ['answers' => 'id', 'question' => 'question_id'],
                'on'              => self::SCENARIO_MULTI,
            ],
```

run and see:
> The "targetAttribute" property must be configured as a string.

but in first time error in https://github.com/yiisoft/yii2/blob/master/framework/validators/ExistValidator.php#L131

and in second time error in https://github.com/yiisoft/yii2/blob/master/framework/validators/ExistValidator.php#L90

Vovan-VE:
May be related to #9935.

andrewnester:
I am not sure if it relates to #9935

Seems like behaviour of ExistValidator doesn't depend on context it's used (directly used or inside EachValidator).
For both cases we mention that string targetAttribute must be used

So if it will be confirmed as bug or improvement I suppose it will be connected only with ExistValidator

pkirill99:
yes, ExistValidator has 'allowArray' as param, but it not work in 
https://github.com/yiisoft/yii2/blob/master/framework/validators/ExistValidator.php#L90

andrewnester:
@pkirill99 you are absolutely right, I just fixed it and sent PR

cebe:
[...]

PR #11318, opened 2016-04-10
Bug: EachValidator invokes the rule validator

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Fixed issues | #10566, #10565, #9935 |

The logic for `validateAttribute()` got simplified. The validator is created with the model, then each value is validated individually using the model, the attribute and the validator without skipping any step.


framework/validators/EachValidator.php
```diff
@@ -114,20 +114,34 @@ private function createEmbeddedValidator($model)
...
+                if ($cloneModel->hasErrors($attribute)) {
+                    $this->addError(
+                        $model,
+                        $attribute,
+                        $this->message,
+                        ['value' => $v]
+                    );
+                    $cloneModel->clearErrors($attribute);
+                }
+                $values[$i] = $model->$attribute;
```
Faryshta: note to self: fix this line

klimov-paul:
Resolved by commit 205f16811a6071ad5d1dbea4cba9fd8a3282f6c5

Issue #11314, opened 2016-04-09
ActionColumn buttons (by default) reset filters


vologab:
### What steps will reproduce the problem?

1. Create GridView with ActionColumn and filters/sorting features.
2. Select some filters and sort some columns
3. Click button in ActionColumn
4. As result - all filters reseted

### What is the expected result?

Filters/Sorters should be saved/restored after clicking buttons in action column.

### What do you get instead?

As temporary solution, it's possible use closure in action column

```php
[
'class' => '\yii\grid\ActionColumn',
'urlCreator' => function ($action, $model, $key, $index) {

 $request = \Yii::$app->getRequest();
 $paramsFromRequest = $request instanceof \yii\web\Request ? $request->getQueryParams() : [];

$params = is_array($key) ? $key : ['id' => (string) $key];
$params[0] = \Yii::$app->controller->id . '/' . $action;

return \Yii::$app->urlManager->createUrl(array_merge($paramsFromRequest, $params));
]
```

### Additional info

Method `createUrl()` also used in `Pagination.php`, `Sort.php`, but in these classes for creating Url is used
current parameters from request, so we don't have problems with resetting filters/sorting.

But for action column we should use slightly different solution because, creteUrl will be executed for each row/column, so we need params 'caching'   (may be in GridView widget) from request for avoiding ` $paramsFromRequest = $request instanceof \yii\web\Request ? $request->getQueryParams() : [];` calculating  in each row.
 
| Q                | A
| ---------------- | ---
| Yii version      | 2.0.7-372-g5921865
| PHP version      |  PHP 5.6.11
| Operating system |  Ubuntu

PR #11320, opened 2016-04-10
11314 action column reset filters

vologab:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11314 |


samdark:
It probably breaks BC behavior-wise i.e. previously no parameters were submitted to action triggered by a button but not it is.

Issue #11322, opened 2016-04-11
UniqueValidator misleading error message for multiple targetAttribute's


PowerGamer1:
When `UniqueValidator` is setup like this:
```PHP
 // a1 and a2 need to be unique together, and they both will receive error message
 [['a1', 'a2'], 'unique', 'targetAttribute' => ['a1', 'a2']]
```
and the pair of fields fails unique check the validator produces two separate error messages that are completely wrong and misleading:
```
a1 "123" has already been taken.
a2 "456" has already been taken.
```

The actual error messages should have been along the lines:
```
The combination ("123", "456") of (a1, a2) has already been taken.
The combination ("123", "456") of (a1, a2) has already been taken.
```

| Yii version      |  2.0.8-dev
| PHP version      |  7.0.1
| Operating system |  Windows

SilverFire:
Thank you for the report. Would you like to submit a PR?

PowerGamer1:
@SilverFire
> Would you like to submit a PR?

Here is one: https://github.com/yiisoft/yii2/pull/11323. Code is fixed but since a new error message was introduced it will need translation to many languages which I cannot do.

lynicidn:
it bad idea, your code execute 2 equals query, for generate two errors

PowerGamer1:
> it bad idea, your code execute 2 equals query, for generate two errors

After the fix you will use the `unique` validator only on a single field (out of all the fields constituting the combo) - the field you want the error message to show on. For example:

```PHP
 // a1 and a2 need to be unique together, but only a1 will receive error message
 ['a1', 'unique', 'targetAttribute' => ['a1', 'a2']]
```

This way you get single query and single error message.

enigmatix:
I'm not sure if this has found it's way into the dev branch yet, but I have just rolled back to 2.0.8 due to an introduced bug in 2.0.9-dev.  It seems similar to what you are discussing here.  The issue i had was for this configuration:

```
 // a1 and a2 need to be unique together, and they both will receive error message
 [['a1', 'a2'], 'unique', 'targetAttribute' => ['a1', 'a2']]
```

The behavior was the error was only showing under a2, not both a1 and a2.

Likewise, with this configuration

` [['a2'], 'unique', 'targetAttribute' => ['a1', 'a2']]`

The error message was still displayed under a1.

PowerGamer1:
@enigmatix 
[...]

PR #11323, opened 2016-04-11
Update UniqueValidator.php

PowerGamer1:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11322 |
- [x] Code
- [x] Reviewed
- [x] Changelog
- [x] Tests (@SilverFire)

Fix for https://github.com/yiisoft/yii2/issues/11322
Needs many language translations for new error message: 'Combination [{valueCombo}] of [{attributeCombo}] has already been taken.'

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -157,6 +157,7 @@ class UniqueValidator extends Validator
                         $valueCombo[] = $model->$k;
                     }
                 }
+                array_walk($attributeCombo, function (&$attribute) use($model) { $attribute = $model->getAttributeLabel($attribute); });
                 array_walk($valueCombo, function (&$value) { $value = '"' . $value . '"'; });
                 $this->addError($model, $attribute, $this->comboNotUnique, ['attributeCombo' => implode(', ', $attributeCombo), 'valueCombo' => implode(', ', $valueCombo)]);
             } else {
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -59,8 +59,7 @@ class UniqueValidator extends Validator
      */
     public $filter;
 	/**
-     * @var string the user-defined error message used when `targetAttribute` is an array. It may contain the following placeholders which
-     * will be replaced accordingly by the validator:
+     * @var string the user-defined error message used when [[targetAttribute]] is an array. It may contain the following placeholders:
      *
      * - `{attributeCombo}`: the labels of the attributes being validated
      * - `{valueCombo}`: the values of the attributes being validated
@@ -148,17 +147,15 @@ class UniqueValidator extends Validator
             if (is_array($targetAttribute)) {
                 $attributeCombo = [];
                 $valueCombo = [];
-                foreach ($targetAttribute as $k => $v) {
-                    if(is_int($k)) {
-                        $attributeCombo[] = $v;
-                        $valueCombo[] = $model->$v;
+                foreach ($targetAttribute as $key => $value) {
+                    if(is_int($key)) {
+                        $attributeCombo[] = $model->getAttributeLabel($value);
+                        $valueCombo[] = '"' . $model->$value . '"';
                     } else {
-                        $attributeCombo[] = $k;
-                        $valueCombo[] = $model->$k;
+                        $attributeCombo[] = $model->getAttributeLabel($key);
+                        $valueCombo[] = '"' . $model->$key . '"';
                     }
                 }
-                array_walk($attributeCombo, function (&$attribute) use($model) { $attribute = $model->getAttributeLabel($attribute); });
-                array_walk($valueCombo, function (&$value) { $value = '"' . $value . '"'; });
                 $this->addError($model, $attribute, $this->comboNotUnique, ['attributeCombo' => implode(', ', $attributeCombo), 'valueCombo' => implode(', ', $valueCombo)]);
             } else {
                 $this->addError($model, $attribute, $this->message);
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -145,18 +145,23 @@ class UniqueValidator extends Validator
 
         if ($exists) {
             if (is_array($targetAttribute)) {
-                $this->addComboNotUniqueError();
+                $this->addComboNotUniqueError($model, $attribute);
             } else {
                 $this->addError($model, $attribute, $this->message);
             }
         }
     }
     
-    private function addComboNotUniqueError()
+    /**
+     * Builds and adds [[comboNotUnique]] error message to the specified model attribute.
+     * @param \yii\base\Model $model the data model.
+     * @param string $attribute the name of the attribute.
+    */
+    private function addComboNotUniqueError($model, $attribute)
     {
         $attributeCombo = [];
         $valueCombo = [];
-        foreach ($targetAttribute as $key => $value) {
+        foreach ($this->targetAttribute as $key => $value) {
             if(is_int($key)) {
                 $attributeCombo[] = $model->getAttributeLabel($value);
                 $valueCombo[] = '"' . $model->$value . '"';
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -145,21 +145,26 @@ class UniqueValidator extends Validator
 
         if ($exists) {
             if (is_array($targetAttribute)) {
-                $attributeCombo = [];
-                $valueCombo = [];
-                foreach ($targetAttribute as $key => $value) {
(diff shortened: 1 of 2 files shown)


Faryshta:
Can you add that line to the translations?

framework/validators/UniqueValidator.php
```diff
@@ -69,6 +78,9 @@ public function init()
         if ($this->message === null) {
             $this->message = Yii::t('yii', '{attribute} "{value}" has already been taken.');
         }
+	    if ($this->comboNotUnique === null) {
+		    $this->comboNotUnique = Yii::t('yii', 'Combination [{valueCombo}] of [{attributeCombo}] has already been taken.');
```
Faryshta: Can you add this to translations messages?

PowerGamer1: @Faryshta
> Can you add this to translations messages?

I can't translate this from English into other languages, so I cannot add it.

Faryshta: just add it to the english translation, then the translators (including myself) will use that to translate it to other languages.

PowerGamer1: "messages" folder of the framework does not have a subfolder for English language. I think it is sufficient to specify English strings directly in the code using `Yii::t()` and then "messages/XX/yii.php" files for each language XX are generated automatically using `yii message` command line command.

framework/validators/UniqueValidator.php
```diff
@@ -58,6 +58,15 @@ class UniqueValidator extends Validator
      * is the [[\yii\db\Query|Query]] object that you can modify in the function.
      */
     public $filter;
+	/**
+     * @var string the user-defined error message used when `targetAttribute` is an array. It may contain the following placeholders which
+     * will be replaced accordingly by the validator:
+     *
```
SilverFire: This line is redundant

framework/validators/UniqueValidator.php
```diff
@@ -58,6 +58,15 @@ class UniqueValidator extends Validator
      * is the [[\yii\db\Query|Query]] object that you can modify in the function.
      */
     public $filter;
+	/**
+     * @var string the user-defined error message used when `targetAttribute` is an array. It may contain the following placeholders which
```
SilverFire: <code>\`targetAttribute\`</code> -> `[[$targetAttribute]]`

framework/validators/UniqueValidator.php
```diff
@@ -133,7 +145,24 @@ public function validateAttribute($model, $attribute)
         }
 
         if ($exists) {
-            $this->addError($model, $attribute, $this->message);
+            if (is_array($targetAttribute)) {
```
SilverFire: The code bellow should be moved into a separate protected method to make its modification easier. And current method is already too long

framework/validators/UniqueValidator.php
```diff
@@ -133,7 +145,24 @@ public function validateAttribute($model, $attribute)
...
+                $attributeCombo = [];
```
SilverFire: The code bellow should be moved into a separate protected method to make its modification easier. And current method is already too long

framework/validators/UniqueValidator.php
```diff
@@ -133,7 +145,24 @@ public function validateAttribute($model, $attribute)
...
+                $valueCombo = [];
+                foreach ($targetAttribute as $k => $v) {
```
SilverFire: According to [core code style](https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md#42-properties) we prefer to avoid short variable names such as `$k`, `$v`. Could you rename it, please?

framework/validators/UniqueValidator.php
```diff
@@ -133,7 +145,24 @@ public function validateAttribute($model, $attribute)
...
+                    if(is_int($k)) {
+                        $attributeCombo[] = $v;
+                        $valueCombo[] = $model->$v;
+                    } else {
+                        $attributeCombo[] = $k;
+                        $valueCombo[] = $model->$k;
+                    }
+                }
+                array_walk($attributeCombo, function (&$attribute) use($model) { $attribute = $model->getAttributeLabel($attribute); });
```
SilverFire: I'd prefer to put this logic in a previous loop

framework/validators/UniqueValidator.php
```diff
@@ -133,7 +145,24 @@ public function validateAttribute($model, $attribute)
...
+                array_walk($valueCombo, function (&$value) { $value = '"' . $value . '"'; });
```
SilverFire: same here

SilverFire:
We also need a CHANGELOG line and a unit test

framework/validators/UniqueValidator.php
PowerGamer1: Makes sense, but just so you know I copy-pasted this from `public $message` in `\yii\validators\Validator` so you may want to fix it there (and probably many other validator classes) too.

framework/validators/UniqueValidator.php
PowerGamer1: I can agree if you can come up with a good name for this separate method (`addError` cannot be used since its already defined in base class) and IMO this method should be made private so that potential refactorings of `UniqueValidator` class can have more freedom in splitting/merging class methods, etc., without worrying of breaking backward compatibility if someone used this `protected` method in their code.

framework/validators/UniqueValidator.php
PowerGamer1: No problem, but it would be totally inconsistent to have here full names and just a few lines above in the same method in the similar loop I copy pasted to have short names breaking your core code style.

framework/validators/UniqueValidator.php
PowerGamer1: The reason I coded it outside a loop was to prevent duplication of quoting char (") for value but for performance reasons I can agree that it may be better to avoid `array_walk` and closure and put it inside a loop.

klimov-paul:
Probably, same changes should be applied to `ExistValidator`.

framework/validators/UniqueValidator.php
SilverFire: How about `addCompositeError`? Then it makes sense to rename `comboNotUnique` -> `compositeMessage`

PowerGamer1:
@SilverFire
> We also need a CHANGELOG line and a unit test

Changelog can read along the lines "Fixed incorrect error message in UniqueValidator class when multiple model attributes must constitute a unique combination" but I am not sure were to put it - feel free to rephrase and put it where needed.

As for the unit tests - no new tests should be required since no new ways of using `UniqueValidator` class were added and only its error output in some existing use cases was fixed (In other words existing unit tests need to be changed to expect a different error message added on a model from `UniqueValidator`). As I never dealt with unit testing before I leave it up to you to deal with them in whatever way you deem necessary.

Also, I fixed most lines you commented on in my pull request (you may still want to read my comments on the fixed lines though). If it makes things faster/easier feel free to copy over all the code from my pull request and make your own with whatever changes you want.

SilverFire:
I'd suggest the following wording: <code>Bug #11322: Fixed incorrect error message in \`yii\validators\UniqueValidator\` for composite target attribute (PowerGamer1, silverfire)</code>. It must be added to https://github.com/yiisoft/yii2/blob/master/framework/CHANGELOG.md


> As I never dealt with unit testing before I leave it up to you to deal with them in whatever way you deem necessary.

Tests are required because we are introducing new logic. I'll add them before merge.

framework/validators/UniqueValidator.php
PowerGamer1: Hm, can't get rid of word "combo" everywhere in the same way: there are also message placeholders {attributeCombo} and {valueCombo}. Saying {attributeComposition} and {valueComposition} sounds wrong. Besides, `comboNotUnique` is consistent with other validators names like `tooBig`, `wrongMimeType`, `ipv4NotAllowed`, etc. And `compositeMessage` would break the consistency. So I will name the method `addComboNotUniqueError`.

cebe:
> Probably, same changes should be applied to ExistValidator.

this does not apply to ExistValidator as that one simply talks about an attribute being invalid if validation fails.

cebe:
Merged with small adjustments. Thank you!

rob006:
This breaks BC - if someone has custom message specified in `message` property, it will be ignored and default message for `comboNotUnique` will be used.

chathurasudarsha:
this will not be used since 2.1.  need a promised fix here. something like message =>['attribute1'=>'message1', 'attribute2'=>'message2'];

cebe:
@chathurasudarsha it is fixed in #12305

Issue #10883, opened 2016-02-17
ActiveRecord: alias in relation


CedricYii:
Dynamic aliases in relation seem to be missing. Consider this example:
````php
class Customer {
...
  public function getOrders()
  {
      return $this->hasMany( Order::className(), ['customerid' => 'customerid'] )
                  ->andOnCondition( [ 'statusid' => 1 ] );
  }
...
}
````
Then
````php
$customer = Customer::find()
                   ->alias( 'c' )
                   ->innerJoinWith('orders o');
````
Build the query:
````
select customer.* from customer c
join order o on o.customerid = c.customerid 
where statusid = 1
````
So the alias has been applied on customerid but not on statusid in the getOrders relation.
Which is likely to result in a ``column ambiguously defined`` error (as statusid is in every tables).

Is there a way to define 
````php
public function getOrders()
{
      return $this->hasMany( Order::className(), ['customerid' => 'customerid'] )
                 //add dynamic alias here with $this->getCurrentRelationAlias()
                  ->andOnCondition( [ $this->getCurrentRelationAlias() . '.statusid' => 1 ] ); 
}

````
so that all the conditions can be aliased dynamically (a same relation may be used several times with different aliases)?
As Yii is able to know the alias of order in the hasMany() part for this relation, I guess it should be able to recall it in the OnCondition part?

(Issue [https://github.com/yiisoft/yii2/issues/7263] doesn't seem to deal with conditions added inside relations; but I may misunderstand).

cebe:
> (Issue [#7263] doesn't seem to deal with conditions added inside relations; but I may misunderstand).

it does. I do not see how this is different from #7263

CedricYii:
Examples in #7263 are putting emphasis on columns' aliases. Here my main concern is about using the same table in different relations with different aliases. Consider this example:
````php
class Order {
...
  //first relation with table COMPANY
  public function getCustomer()
  {
      return $this->hasOne( Company::className(), ['id' => 'customerid'] ) //table COMPANY
                  ->andOnCondition( [ 'type' => 'customer' ] );
  }

  //second relation with table COMPANY
  public function getSupplier()
  {
      return $this->hasOne( Company::className(), ['id' => 'supplierid'] ) //table COMPANY
                  ->andOnCondition( [ 'type' => 'supplier' ] );
  }
...
}
````
Then we could write
````php
$orders = Order::find()
                   ->alias( 'o' )
                   ->innerJoinWith('customer c')
[...]

PR #11326, opened 2016-04-11
Handles relation's alias in relation declaration

CedricYii:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | not tested through yii unit tests |
| Fixed issues | #10883 |

Using the syntax `==relationAlias==`, the query inside a relation can now adress the alias of the relation without knowing it in advance.

Example:

``` php
class Customer {
...
  public function getOrders()
  {
      return $this->hasMany( Order::className(), ['customerid' => 'customerid'] )
                  ->andOnCondition( [ '==relationAlias==.statusid' => 1 ] );
  }
...
}
```

Then

``` php
$customer = Customer::find()
                   ->alias( 'c' )
                   ->innerJoinWith('orders o');
```

Build the query:

```
select customer.* from customer c
join order o on o.customerid = c.customerid 
where o.statusid = 1
```

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -577,12 +577,28 @@ class ActiveQuery extends Query implements ActiveQueryInterface
 
     /**
      * Replace the key string ==relationAlias== by aliasName into given conditions
-     * @param string|array $condition the given conditions.
+     * @param string|array $condition the given conditions
      * @param string $aliasName
      */
     private function replaceRelationAlias( $condition, $aliasName )
     {
-        return str_replace('==relationAlias==', $aliasName, $condition);
+        if (is_int($condition)) {
+            return $condition;
+        } elseif (is_string($condition)) {
+            return str_replace('==relationAlias==', $aliasName, $condition);
+        } elseif (is_array($condition)) {
+            $conditionNew = [];
+            foreach ($condition as $key => $value) {
+                unset($condition[$key]);
+                $key    = self::replaceRelationAlias($key,   $aliasName);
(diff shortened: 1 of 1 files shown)


cebe:
I'd prefer a complete solution to  #7263 insated of this.

CedricYii:
@cebe So would I. However the alias feature is not really usable without this, and it is unsure if a perfect solution will be out with 2.0.8. I believe Yii can't afford to wait too long for it.
Especially since you planed to refactor all those things in 2.1 (this feature could then be dropped for a better one).

Anyhow, would you have the ability to check if this commit pass yii unit tests?

cebe:
as this adds new functionaility, there should be additional tests added to verfy that it works. However I do not think it is worth adding tests for this particular implenetation as I am quite sure I will not merge it this way. I will try to find some time next week to check it out in detail.

CedricYii:
Thanks for your feedback.
Well I was willing to make sure it doesn't break anything at least... And it sounds to be the case since I finally managed to run the tests.

cebe:
there is also the green checkmark on the commit in this PR, which is from travis-ci tests.

CedricYii:
Ah! Good to know! I was wondering what it was about...

CedricYii:
The last commit fixed alias replacement, it should be usable now.

Renkas:
Any progress on better solution? I need this functionality for my project ASAP - I would rather use framework functionality than make some temporary (which will stay there untill end of time like with all "temporay" things) custom solution.

CedricYii:
Closed in favor to a more complete solution in #11646

PR #11335, opened 2016-04-12
fixes #11333 avoid serializing php7 errors

### LINKED ISSUE #11333: Php7 errors may be not serializable
Php7 introduced \Throwable interface that covers old \Exception and new \Error types. When \Error is thrown \yii\log\Target in formatMessage() method tries to serialize it and fails. There is already workaround to avoid serializing \Exceptions and it should be expanded to cover all \Throwable.

zuozp8:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | i didn't have db prepared, so i run only `phpunit --group=base,log` and it passed |
| Fixed issues | #11333 |


samdark:
:+1:

cebe:
Perfect, thank you!

cebe:
same fix applied also for mongodb logger: https://github.com/yiisoft/yii2-mongodb/commit/4c072ba7cb92580d515be9725027534d95daf68c

PR #11337, opened 2016-04-12
11334 add LogStashTarget

qihuajun:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | no |

Add a LogStashTarget to record logs in JSON format and can be easily imported into logstash.


samdark:
I think it's more suitable to have it as a separate package.

PR #11347, opened 2016-04-13
Fix pjax container option

h0ckeypl4yer:
If we set client option "container", it will be ignored, because param "container" of function always set.


SilverFire:
Thank you!

PR #11348, opened 2016-04-13
Remove (parentheses)

thiagotalma:
As it is, this method causes an error in the database.

According to the [reference](http://dev.mysql.com/doc/refman/5.7/en/alter-table.html), you should not use parentheses.

| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

### DIFF
--- a/framework/db/mysql/ColumnSchemaBuilder.php
+++ b/framework/db/mysql/ColumnSchemaBuilder.php
@@ -31,7 +31,7 @@ class ColumnSchemaBuilder extends AbstractColumnSchemaBuilder
     protected function buildAfterString()
     {
         return $this->after !== null ?
-            ' AFTER (' . $this->db->quoteColumnName($this->after) . ')' :
+            ' AFTER ' . $this->db->quoteColumnName($this->after) :
             '';
     }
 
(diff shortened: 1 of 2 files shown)


SilverFire:
Thank you for your contribution. @thiagotalma could you also update wrong test, please?

Summoning @df2

thiagotalma:
Done!

df2:
Confirmed that parentheses cause error... good catch! Oci and Cubrid ColumnSchemaBuilder classes also should have the parentheses removed.

thiagotalma:
@SilverFire I must correct for other DBMS cited by @df2?
Or open another PR?

SilverFire:
It is OK to update this one

df2:
We should also add `after` tests to non-pk attributes. Dumb error there in the tests on my part--the pk attributes are not included in the database create table simulation.

SilverFire:
@thiagotalma thank you for a good catch!

I'm closing your PR in favor of #11362 that covers all supported DBMSs. Thank you!

thiagotalma:
@SilverFire Excuse me for the delay in answering it.

SilverFire:
@thiagotalma all is fine. Thank you for finding and pointing the problem.

PR #11353, opened 2016-04-14
fix unique validator support out active record model

lynicidn:
This pr allow use unique validator in base model for update other record model, example

``` php
constructor(ActiveRecordInterface $record):
$this->_record = $record;

rules:
['id', 'unique', 'targetClass' => $this->_record, 'targetAttribute' => 'id'],
```

it case check unique relatively record pk (not targetAttributes)

may be best way add new property `target` for it, but here not break BC

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -37,7 +37,7 @@ use yii\db\ActiveRecordInterface;
 class UniqueValidator extends Validator
 {
     /**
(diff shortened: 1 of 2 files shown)


cebe:
Thank you for your contribution, however it seems to me this PR is outdated and in a large conflict with the current implementation of the UniqueValidator, which has recently been refactored. As it is unmergeable I am closing this PR. Feel free to open a new one based on the current implementation.

PR #11378, opened 2016-04-18
Update yii.js

bperek:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

Fixes handling middle or right click on element with data-method (usually POST link). Opening this link in new tab often caused errors (404, method not allowed - when switched to GET on new tab). This solution swap href into data-href disabling other click on element. If element doesn't have a data-method defined nothing happens.

### DIFF
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -146,7 +146,7 @@ yii = (function ($) {
         handleAction: function ($e, event) {
             var $form = $e.attr('data-form') ? $('#' + $e.attr('data-form')) : $e.closest('form'),
                 method = !$e.data('method') && $form ? $form.attr('method') : $e.data('method'),
-                action = $e.attr('href'),
+                action = $e.attr('href') || $e.data('href'),
                 params = $e.data('params'),
                 pjax = $e.data('pjax'),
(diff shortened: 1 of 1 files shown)


framework/assets/yii.js
```diff
@@ -146,7 +146,7 @@ yii = (function ($) {
         handleAction: function ($e, event) {
             var $form = $e.attr('data-form') ? $('#' + $e.attr('data-form')) : $e.closest('form'),
                 method = !$e.data('method') && $form ? $form.attr('method') : $e.data('method'),
-                action = $e.attr('href'),
+                action = $e.attr('href') || $e.data('href'),
```
lynicidn: href is require attribute for link

bperek: Ok, I'll try to do some other fixes with preserving href and solve problem of this bug

lynicidn: try it https://github.com/defunkt/jquery-pjax/issues/106#issuecomment-5250347

zuozp8:
I discussed it with @bperek . I'll provide fix in separate pull request. This can be dropped.

PR #11379, opened 2016-04-18
Validation in rest delete action

kikimor:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

Some records shouldn't be deleted in some business cases. The current implementation yii\rest\DeleteAction does not handle validation errors.

So, currently when I implement validation:

``` php
public function beforeDelete()
{
    if (!$this->validate()) {
        return false;
    }
    return parent::beforeDelete();
}
```

 I get 500 internal server error.

### DIFF
--- a/framework/rest/DeleteAction.php
+++ b/framework/rest/DeleteAction.php
@@ -8,6 +8,7 @@
 namespace yii\rest;
 
 use Yii;
+use yii\db\ActiveRecord;
 use yii\web\ServerErrorHttpException;
 
 /**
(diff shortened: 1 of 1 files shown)


Faryshta:
http://www.yiiframework.com/doc-2.0/yii-db-activerecord.html#delete()-detail delete doesn't trigger any validation.

SilverFire:
@Faryshta is right

PR #11380, opened 2016-04-18
doc typo

tunecino:
not defined in parent. it is [yii\rest\ActiveController::checkAccess()](http://www.yiiframework.com/doc-2.0/yii-rest-activecontroller.html#checkAccess%28%29-detail)


SilverFire:
Thank you

PR #11381, opened 2016-04-18
Fixes #7670 Added normalizer to yii\web\UrlRule

### LINKED ISSUE #7670: UrlRule Problem with trailing slash (/)

cronfy:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | yes/no |
| Fixed issues | #7670 |

This pull request introduces canonizer option in `yii\web\UrlRule`. With it one is able to choose strategy to use when dealing with trailing slashes: redirect, throw 404 or just route.

This is PREVIEW-only version (with some debug code). I would like to hear any feedback and correct the code.

I created this enhancement to avoid routing logic to spread between yii config, `.htaccess` and `on beforeRequest` events. I want to keep this logic in single place and as close to routing rules itself as possible.

Implementation also helps when dealing with legacy code, where different behavior is required at different sections of a site and it is required to keep existing behavior while switching to yii2. Say, `site.org/news/` would require trailing slash with redirect if no slash found, but `site.org/catalog` would require no trailing slashes with 404 if one is given.

Example usage for yii2-basic (I wrap it into GroupUrlRule to demonstrate how it can be isolated from other routing rules, but standalone rules still working of course):

```
'rules' => [
    [
        'class' => 'yii\web\GroupUrlRule',
        'ruleConfig' => ['class' => 'yii\web\UrlRule', 'canonize' => 'trailing-slash'],
        'rules' => [
            'test/home' => 'site/index',
            // ... other rules
        ]
    ]
]
```

With this config 
- `http://site.com/test/home/` shows `site/index` page (yii2 default behavior for `$suffix = '/'`)
- `http://site.com/test/home` will redirect to `http://site.com/test/home/` (introduced behavior).
- `http://site.com/////test///home////` will redirect to  `http://site.com/test/home/` too (introduced behavior).
- `Url::toRoute('site/index')` will produce `/test/home/` (yii2 default behavior for `$suffix = '/'`).

More complex example: 

```
'rules' => [
    [
        'class' => 'yii\web\GroupUrlRule',
        'ruleConfig' => [
            'class' => 'yii\web\UrlRule',
            'canonize' => [
                'trailing-slash' => null, // ignore whether trailing slash exists or not
                'collapse-slashes' => true, // collapse multiple slashes to one
                'action' => '404' // throw NotFound exception for urls with multiple shashes
            ]
        ],
        'rules' => [
            'test/home' => 'site/index',
            // ... other rules
        ]
    ]
]
```

With this config 
- `http://site.com/test/home` shows the page `site/index` (yii2 default behavior for `$suffix = ''`)
- `http://site.com/test/home/` shows 404 page from yii2`(yii2 default behavior for``$suffix = ''``)
- `http://site.com/////test///home////` shows 404 from canonizer (introduced bhavior).
- `Url::toRoute('site/index')` will produce `/test/home` (no trailing slash, yii2 default behavior for `$suffix = ''`).


samdark:
@klimov-paul thoughts?

framework/web/UrlRule.php
```diff
@@ -256,6 +355,25 @@ public function parseRequest($manager, $request)
...
         $matches = $this->substitutePlaceholderNames($matches);
 
+        // if routing succeeded (some rule matched), check whether pathInfo
+        //  was changed during canonization. If so, take the required action.
+        if ($pathInfoCanonized != $pathInfoOrig) {
+            switch ($this->canonize['action']) {
+                case 'redirect':
+                    $url = '/' . $pathInfoCanonized
+                        . (Yii::$app->request->queryString ? '?' . Yii::$app->request->queryString : '');
+                    die('redirect ' . $url);
```
klimov-paul: ???

cronfy: This is for demo only. If required, I can submit another PR without debug code.

framework/web/UrlRule.php
```diff
@@ -213,6 +218,95 @@ public function init()
...
+            // true: yes
+            // false: no
+            // null: leave as is
+            'trailing-slash' => null,
+            // what action to perform if pathInfo was changed during canonization
+            // redirect: redirect to canonical path
+            // 404: throw NotFound exception
+            // none: no action, just route by canonized path
+            'action' => 'none'
+        ];
```
klimov-paul: Does not seem right: it should be either set of separater options of the `UrlRule`|`UrlManager` or a seprated class.

framework/web/UrlRule.php
```diff
@@ -256,6 +355,25 @@ public function parseRequest($manager, $request)
...
+        // if routing succeeded (some rule matched), check whether pathInfo
+        //  was changed during canonization. If so, take the required action.
+        if ($pathInfoCanonized != $pathInfoOrig) {
+            switch ($this->canonize['action']) {
+                case 'redirect':
+                    $url = '/' . $pathInfoCanonized
+                        . (Yii::$app->request->queryString ? '?' . Yii::$app->request->queryString : '');
+                    die('redirect ' . $url);
+                    Yii::$app->getResponse()->redirect($url)->send();
+                    break;
```
klimov-paul: This is definetely incorrect, since program continues after response been sent.

framework/web/UrlRule.php
samdark: Options should be OK, I think.

framework/web/UrlRule.php
samdark: You can commit into the same branch. Commits will be added to PR automatically.

framework/web/UrlRule.php
samdark: Do you mean to termintate it with `Yii::$app->end()`?

klimov-paul:
I can't see how this fixes issue with URL suffixes like `.html`.

samdark:
@klimov-paul I think we need to collect all the cases and write tests. Personally I'm missing something each time :(

framework/web/UrlRule.php
cronfy: It can be set via separate options in rule config, i. e.:

```
        'ruleConfig' => [
            'class' => 'yii\web\UrlRule',
            'canonize' => [
                'trailing-slash' => null, // ignore whether trailing slash exists or not
                'collapse-slashes' => false, // do not warry about multiple slashes
                'action' => 'none' // do nothing
            ]
        ],
```

Or it can be set via shortcut:

```
        'ruleConfig' => [
            'class' => 'yii\web\UrlRule',
            'canonize' => 'trailing-slash' // expands to { trailing-slash: true, collapse-slashes: true, action: redirect }
        ],
```

Is it OK? If not, could you pease give an example of how it should look like in configuration?

Currently supported shortcuts: 'trailing-shash', 'no-trailing-shash', 'disabled'.

By default if  ```canonize``` not set, behavior is not changed.

framework/web/UrlRule.php
klimov-paul: Perhaps.
However, there is another problem: with this logic rule will terminate parsing process on its own with redirect, while there could be another rule in the list, which will match as it is.

cronfy:
> I can't see how this fixes issue with URL suffixes like .html.

@klimov-paul , I can't find .html suffixes mentioned in #7670, could you please give a link to issue about it?

I tested canonizer with .html suffixes, it works ok (just do not changes behavior). Problem can occur only if we set ```suffix => '.html'``` and ```canonize => 'traiing-slash'``` for the same rule, but this is obvious configuration conflict. ```suffix => '.html'``` and ``` canonize => 'no-trailing-slash'``` should work just fine.

framework/web/UrlRule.php
cronfy: Redirection code injected to the point where rule already matched. Current yii2 logic (without canonizer feature) will just process this rule anyway - either action will be run, or NotFound exception will be thrown.

But I just thought - maybe we should return a route instead of explicit redirect? Something like

```
        'ruleConfig' => [
            'class' => 'yii\web\UrlRule',
            'canonize' => [
                'trailing-slash' => true,
                'collapse-slashes' => true,
                'action' => 'redirect',
                'redirect-route' => 'site/redirect'
            ]
        ],
```

It then would return ```[ 0 => 'site/redirect, 1 => ['redirect-url' => $pathInfoCanonized ]]```

samdark: Wouldn't that require redirect action in site controller?

cronfy: Of course, but since canonization is configured manually, developer should take the responsibility of adding redirect action to the controller.

We can add an option (i.e. 'force-redirect' => true) to just redirect without returning a redirection route (as canonizer currently behaves). But as forced redirect can break existing code (if someone uses parseRequest() to just get route info without routing indeed), this should not be the default, I think.

arisk:
Could you change the name to something else, maybe `canonicalize`? Canonize means something completely different.

cronfy:
> Could you change the name to something else, maybe canonicalize? Canonize means something completely different.

:) 
maybe `normalize` then?

arisk:
That works 👍

framework/web/UrlRule.php
samdark: That sounds not much better than current solution in terms of developer actions required: https://github.com/samdark/yii2-cookbook/blob/master/book/handling-trailing-slash-in-urls.md

cronfy: Do you mean yii2 needs to provide some redirection by default, without additional configuration?

cronfy: Cookbook examples require to put routing logic outside of rules configuration, opposite to this solution, which gives a way to set it right in configuration rules. And we can even set different logic to different rule sets in a simple way.

samdark: Well, the initial idea was about doing redirection by default because, for example, Wordpress, which is very common, adds slashes to all URLs in the content therefore all links to Yii websites are becoming 404.

samdark: Of course, I like the fact that it could be set per rule.

cronfy: Ok, I'll try to handle that Wordpress issue in the next commit.

cronfy:
I added some changes.

First, I moved redirect handler to `web\Application->handleRequest()`, because in `UrlRule` there is no `$response` object. Normalizer now throws an exception when instant redirect is required. This breakes BC in case one uses UrlRule by itself, i. e. for checking how a request would be routed. `UrlManagerTest->testParseRequest()` failed with this change, so I updated `UrlManagerTest` to be aware of the exception thrown.

Second, I added strategy `safe-remove-trailing-slash` for that issue with Wordpress back links. This is default strategy. It redirects 'posts.html/' to 'posts.html', 'posts/' to 'posts' etc, but only if suffix is not '/'. This should be safe enough and passed all `UrlRuleTest`'s tests.

Third, I added ability to specify redirect route. With it one can handle redirect in desired controller action. Redirect route is returned instead of original. Original route is kept in `origRoute` param. Example rule:

```
[
    'pattern' => 'post/<id>/<title>',
    'route' => 'post/view',
    'normalize' => [
        'collapse-slashes' => true,
        'add-trailing-slash' => false,
        'remove-trailing-slash' => true,
        'action' => 'route',
        'route' => 'site/redirect'
    ],
]
```

If normalization occurs, returns route:

```
// request: post/123/this+is+sample/
[
    'site/redirect',
    [
        'origPathInfo' => 'post/123/this+is+sample/',
        'redirectUrl' => 'post/123/this+is+sample',
        'origRoute' => ['post/view', ['id' => 123, 'title' => 'this+is+sample']],
    ]
]
```

@klimov-paul, @samdark thoughts?

cronfy:
The Travis CI build could not complete... why? I merged with master, no luck.

samdark:
That's our issues. Will be fixed soon.

tests/framework/web/UrlManagerTest.php
```diff
@@ -310,8 +311,14 @@ public function testParseRequest()
...
         $request->pathInfo = 'post/123/this+is+sample/';
-        $result = $manager->parseRequest($request);
-        $this->assertEquals(['post/123/this+is+sample/', []], $result);
+        try {
+            $resultExceptionClass = null;
+            $result = $manager->parseRequest($request);
+        } catch (NormalizerActionException $e) {
+            $resultExceptionClass = get_class($e);
+        }
+        $this->assertEquals('yii\web\NormalizerActionException', $resultExceptionClass);
```
samdark: Could we use existing HTTP exception instead?

samdark:
Fixed. Please merge from master once again.

framework/web/UrlRule.php
```diff
@@ -216,6 +221,145 @@ public function init()
...
+    protected function initNormalizer($manager) {
+        $normalize = $this->normalize;
+
+        $configNoAction = [
+            // whether to collapse multiple slashes to single
+            'collapse-slashes' => false,
+            // remove trailing slash if it exists
+            'remove-trailing-slash' => false,
+            // add trailing slash if there is no one
+            'add-trailing-slash' => false,
```
rob006: This two settings can be changed to one: `'trailing-slash' => true|false|null`.
* `true` - add trailing slash
* `false` - remove trailing slash
* `null` - do nothing

framework/web/UrlRule.php
```diff
@@ -216,6 +221,145 @@ public function init()
...
+            // none: no action, just route by normalized path
+            'action' => 'none',
+            // string: route for $action == 'route', i.e. 'site/redirect'
+            'route' => null
+        ];
+
+        if (is_string($normalize) || is_null($normalize)) {
+            switch ($normalize) {
+                case '':
+                case 'safe-remove-trailing-slash':
```
rob006: It is better to declare such keys as constants - then you do not have to remember them, because the IDE will tell you possible values.

cronfy: Do you mean that instead

```
        $configNoAction = [
            'collapse-slashes' => false,
            'trailing-slash' => false,
            'action' => 'none',
            'route' => null
        ];
```

is better to write something like:

```
        $configNoAction = [
            self::NORMALIZE_COLLAPSE_SLASHES => false,
            self::NORMALIZE_TRAILING_SLASH => false,
            self::NORMALIZE_ACTION => self::NORMALIZE_ACTION_NONE,
            self::NORALIZE_ROUTE => null
        ];
```

like this?

Second example looks ugly to me.. I've never seen such practice in ```yii-config.php```.

rob006: This is more reliable approach (no typos) and much more convenient (you can use code completion from IDE).

samdark: You haven't seen it because Yii components aren't typo-tolerant i.e. make a typo and it will error.

cronfy: @samdark yes, now I see. 

Ok, let;s imagine this config:

```
'ruleConfig' => [
    'class' => 'yii\web\UrlRule',
    'suffix' => '',
    'normalize' => [
        UrlRule::NORMALIZE_TRAILING_SLASH => false,
        UrlRule::NORMALIZE_COLLAPSE_SLASHES => true
    ]
],
```

does not look good for me.

Maybe Normalizer should be a separate class then? Then it can be configured as usual Yii object. And all settings become class properties (e. g. ```$this->trailingSlash```), then it will be typo-tolerant and IDE-friendly.

cronfy:
I extracted normalizer logic into it's own class ```UrlNormalizer```. It is now created and configured via ```Yii::createObject()```. All configurable options (e.g. ```trailingSlash```, ```collapseSlashes``` are now class properties, thus protected against typos and are IDE-friendly.

I also added ```normalizer``` property to ```UrlManager```allowing normalizer to be configured globally. I could use ```ruleConfig```, but other rule classes (e.g. ```GroupUrlRle```) may be not aware of normalizer, so I decided to create separate property.

Example configuration:

```
'urlManager' => [
    'enablePrettyUrl' => true,
    'showScriptName' => false,
    'normalizer' => ['class' => \yii\web\UrlNormalizer::className(), 'strategy' => 'safe-remove-trailing-slash'],
    'rules' => [
        // uses global normalizer configuration from UrlManager
        'test/home' => 'site/index',
        // ... other rules
    ]
]
```

By-option configuration example:

```
'rules' => [
    [
        'class' => 'yii\web\GroupUrlRule',
        'ruleConfig' => [
            'class' => 'yii\web\UrlRule',
            'normalizer' => [
                'trailingSlash' => false, // remove trailing slash
                'collapseSlashes' => true, // collapse multiple slashes to one
                'action' => 'redirect' // redirect if url was changed during normalization
            ],
            'suffix' => '.html'
        ],
        'rules' => [
            'blog/<id:\d+>' => 'post/show',
            // ... other rules
        ]
    ]
]
```

Default UrlManager configuration for normalizer is:

```
    public $normalizer = ['class' => 'yii\web\UrlNormalizer', 'strategy' => 'safe-remove-trailing-slash'];
```

which fixes #6498 (redirect to url without slash if ```$suffix``` is not ```/```).

rob006:
You still have magic values  for `strategy` and `action` that are not constants. 

You also create Normalizer object for each rule, and they are not cached, so this can degrade performance. Why not create one normalizer for `urlManager` and use it for all rules which don't have own normalizer settings? 

```
            'normalizer' => [
                'trailingSlash' => false, // remove trailing slash
                'collapseSlashes' => true, // collapse multiple slashes to one
                'action' => 'redirect' // redirect if url was changed during normalization
            ],
```
It looks like this will not work - right now `strategy` is required and it will always override this settings.

cronfy:
Now explicit configuration options will override strategy settings. E. g.:

```
'normalizer' => [
    'strategy' => UrlNormalizer::STRATEGY_REMOVE_TRAILING_SLASH,
    'collapseSlashes' => false,
],
```

will use strategy to remove trailing slashes, but will not collapse multiple slashes.

'Magic strings' replaced with constants.

But I have no good idea how to implement ```UrlNormalizer``` caching right now. I think that just using ```UrlNormalizer``` from ```UrlManager```, if ```normalizer``` was not specified for a rule is not enough. If one sets ```normalizer``` for ```ruleConfig```, a new object will be created for each rule each time it is accessed anyway. I would like to know about a good way how to solve this in Yii.

rob006:
see #12240

samdark:
Closing in favor of #12240

Issue #11384, opened 2016-04-19
middle/right clicking links that should go through handleAction()


zuozp8:
### What steps will reproduce the problem?
Create link like
```php
echo Html::a('submit', ['site/foobar'], [
     'data' => [
        'method' => 'post',
        'params' => [
            'name1' => 'value1',
            'name2' => 'value2',
        ],
    ],
]);
```
Click that link with middle mouse button or click with right button → open in new tab
### What is the expected result?
Nothing should happen.
### What do you get instead?
this link opens in new tab with GET request without extra params, this often caused errors like 404 or 405

SilverFire:
See #11385

PR #11385, opened 2016-04-19
middle/right clicking links that should go through handleAction()

zuozp8:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | I found no tests for js |
| Fixed issues | #11384 |


SilverFire:
I'm not feeling comfortable with creating a global-scope handler for a `click` and `contextmenu` event. 

Please, check out this [fiddle](https://jsfiddle.net/dbfd5qwv/2/). It seems to work as expected, isn't it?

zuozp8:
It works only because the alert is called and it prevents from fully performing the click.
I think handling this with alert would rather nasty for end user.

As I investigated that issue I haven't found nicer solution working on all major browsers. https://jsfiddle.net/dbfd5qwv/3/

SilverFire:
There is a bunch of ways how you can open a link in new tab:

1. Right Mouse Button (MB) -> Open in new tab
2. Ctrl + Left MB
3. Middle MB
4. Tab (repeat X times) -> Ctrl + Enter (or Shift+Enter for new window)
5. Tab (repeat X times) -> Menu key -> Open in new tab

Once we will fix middle button, issue to fix another way will appear.
As a result - over complicated code and not all for sure not all ways closed.

There only reliable way I see: do not use `href` attribute and use `data-href` instead, but it's a BC break.

I'm marking this issue as wontfix.

PR #11152, opened 2016-03-20
Batch query test

cebe:
This adds a test for batch() and each() query with large table as some issues have been reported in the past. The default table will be created with 1000 records to not stress travis too much. You can try it locally by setting the table size to 100 000 or even 10 000 000. 

I was not able to reproduce the problem with this test on my machine, maybe someone else can run it and see what happens.

| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | lets wait for travis ;) |
| Related issues | #8420, #9269 |

/cc @particleflux can you try this test on your machine?

Issue #9269, opened 2015-07-31
Query::batch - memory limit


yujin1st:
Hello, I need to iterate 200k record table, but when i use  `foreach (Item::find()->each() as $item)` it fails with memory error.  Loop doesn't execute at all.
I found only [one topic](http://www.yiiframework.com/forum/index.php/topic/32502-120-million-rows-iterator-needed/) about this with no solution.

It works only on development environment (mac, mamp 3.0.6, 128 mb memory limit), but on deployment servers  it throws an error. 
Could you suggest where should i search the difference?

samdark:
1. PHP version.
2. PDO version.
3. PDO settings about emulating prepared statements.

samdark:
http://yiiframework.ru/forum/viewtopic.php?f=19&t=31157

samdark:
https://github.com/yiisoft/yii2/issues/8420

mikehaertl:
I have the exact same problem and already tried the suggestion in #8420 (`PDO::MYSQL_ATTR_USE_BUFFERED_QUERY => true`) but it didn't help. My code is basically this:

```
$query = (new \yii\db\Query())
    ->select('*')
    ->from('bigtable')
    ->where('active=1');

foreach ($query->batch(200) as $batch) {
    // this line is never excueted due to a memory exhaustion exception
}
```

> 'yii\base\ErrorException' with message 'Allowed memory size of 134217728 bytes exhausted (tried to allocate 32 bytes)' in /var/www/vendor/yiisoft/yii2/db/Command.php:837

@samdark Your above link goes to a russian site, so not sure, if I missed some essential information. Anything else we could try? It seems like `batch()` doesn't really behave as promised by default.

samdark:
Is debug turned off?

mikehaertl:
No, should it be?

samdark:
Yes beause Yii keeps debug stuff in memory and it grows.

nineinchnick:
I usually disable all log targets when processing large amounts of data in dev environment. Logs get spammed with query profiling info and I couldn't figure out how to force them to flush.

mikehaertl:
I've tried with `YII_DEBUG==false` and `YII_ENV = 'prod'`. Still got the same error. I also tried both, console and web application. It happens in both.

@samdark BTW your tipp should still be documented somewhere. It's not really obvious.
@nineinchnick Also tried that, didn't help.

samdark:
I'm out of ideas.

[...]

PR #11393, opened 2016-04-20
Batch Query improve for large mysql tables #9269 #11152

aivanouski:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #9269, #11152 |

### DIFF
--- a/framework/db/BatchQueryResult.php
+++ b/framework/db/BatchQueryResult.php
@@ -49,7 +49,10 @@ class BatchQueryResult extends Object implements \Iterator
      * If false, a whole batch of rows will be returned in each iteration.
      */
     public $each = false;
-
+    /**
+     * @var boolean $split split query to many queries with "offset, limit"
+     */
+    public $split = false;
     /**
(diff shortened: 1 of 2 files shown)


ihipop:
`each()` is `batch()` with `internel iteration`
i ditn't see  any `$split` parameters in `each()`

cebe:
right, this change is not correct.

ihipop:
@cebe I think the only thing to make it  right is just add `$split` parameters in `each()` function, and pass it to  ` Yii::createObject` of `BatchQueryResult::className()` .The others a fine.
@aivanouski

ihipop:
anyone ?
#9269

Sammaye:
This PR is not correct, it is too simple for one, it does not take pre-defined offset or limit into account

andrey-denisenko:
@Sammaye 
> This PR is not correct, it is too simple for one, it does not take pre-defined offset or limit into account
Your solution does that?

PR #11404, opened 2016-04-22
loadMultiple() returns true even if load() returns false

xaerobiont:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | may be |
| Tests pass? | yes |
| Fixed issues | yes |

The problem is `loadMultiple()` method in `base/Model` doesn't checks result of `load()`.
I.e. you don't specify some property as safe or any other situation when `load()` returns `false` - you get `true` from loadMultiple. In other words method anyway returns true even is something failed. I think it's bug.

Futhermore: I beleive this method should return false if one of the models doesn't loaded, but now it returns true whether at least one of them populated. I think it's not good, but I didn't change this logic.


cebe:
The purpose of return value of `load()` is to determine whether some data was sent or not.
Return value of `loadMultiple()` indicates the same purpose:

> whether at least one of the models is successfully populated.

if a record exists in POST, even though it is empty there is still a post that has been sent so we should start triggering validation. As far as I see, current implementation is correct in all parts.

zvook:
> The purpose of return value of load() is to determine whether some data was sent or not.
> Return value of loadMultiple() indicates the same purpose

Ok, but `load()` returns false if data was not loaded into the model (it's obviously, I think) and it means that `loadMultiple()` should return the same indicator, isn't it?

lynicidn:
without formName set $success = true when model not loaded is bug

mastikhinanton:
`loadMultiple()` always returns true if `$data[$formName][$i]` or `$data[$i]` not empty (depends on `$formName`). In fact, result depends only on `$data` content, ignoring is `load() `successfull or not.

Example from [Collecting tabular input](http://www.yiiframework.com/doc-2.0/guide-input-tabular-input.html):
`if (Model::loadMultiple($settings, Yii::$app->request->post()) && Model::validateMultiple($settings)) {`
If $settings are already exist, this condition will return true even if data wasn't loaded.

samdark (approved):
The change makes sense to me.

SilverFire:
This change breaks logical BC, isn't it?

samdark:
@SilverFire not according to docs. New behavior matches em.

SilverFire (approved):

samdark:
Merged. Thanks!

cebe:
There should be a test for this.

samdark:
@zvook would you like to work on a test?

zvook:
@samdark yes sure

zvook:
@samdark here it is - https://github.com/yiisoft/yii2/pull/13585

Issue #10878, opened 2016-02-17
ActiveRecord: Alias with nested relations


CedricYii:
I am starting to use the new alias possibility in 2.0.7 which sounds great.
However, I can't figure out how to use it with nested relations.

The expected sql would be:

```
select * from customer c
left join order o on o.customerid = c.customerid and o.amount=100
left join product p on p.productid = o.productid
```

But building the query:

``` php
$customer = Customer::find()
                   ->alias('c')
                   ->joinWith('order o')
                   ->joinWith('order.product p') 
                   ->addOnCondition('o.amount' => 100)
```

Gives something like:

```
select * from customer c
left join order o on o.customerid = c.customerid and o.amount=100
left join order on order.customerid = c.customerid
left join product p on p.productid = order.productid

```

Maybe using `joinWith('order o.product p')` or `joinWith('order.product o.p')`  would make sense but none are allowed.
Am I missing something?

Alex-Code:
You need to do the second relation as an anonymous function.
```php
$customer = Customer::find()
    ->alias('c')
    ->joinWith(['order o' => function($q) {
        $q->joinWith('product p') 
            ->addOnCondition('o.amount' => 100);
    }]);
```

CedricYii:
Thanks, you are right, it does the trick.
I will use it as a workaround. However it breaks the new magic of easily writing aliases.
I believe the syntax ``joinWith('order o.product p')`` would be clearer and should be implemented.

Alex-Code:
Each call to `joinWith` will be treated as a new join, it doesn't check to see if the relation has previously been specified.

CedricYii:
I mean I would expect to be able to write:
````
$customer = Customer::find()
                   ->alias('c')
                   ->innerJoinWith('order o.product p') 
                   ->addOnCondition('o.amount' => 100)
````
(ie only one joinWith).

CedricYii:
I was hoping that the label for this would rather be ``feature/db`` or even ``type:enhancement`` to have this writing syntax enabled.

dynasource:
> Thanks, you are right, it does the trick.
> I will use it as a workaround.

Lets keep it simple for now by following 1 way to do it (suggested by @Alex-Code  https://github.com/yiisoft/yii2/issues/10878#issuecomment-185116346)

PR #11421, opened 2016-04-25
Handles alias in nested relations

CedricYii:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | not written yet |
| Fixed issues | #10878 |

Make it possible to write simple alias syntax with nested relations like:

``` php
$customer = Customer::find()
                   ->alias('c')
                   ->innerJoinWith('order o.product p') 
                   ->addOnCondition('o.amount' => 100)
```

As proposed in issue #10878

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -502,9 +502,18 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             while (($pos = strpos($name, '.')) !== false) {
                 $childName = substr($name, $pos + 1);
                 $name = substr($name, 0, $pos);
+             
+                $alias = null;
+                if (preg_match('/^(.*?)(?:\s+AS\s+|\s+)(\w+)$/i', $name, $matches)) {
+                    // relation is defined with an alias, adjust $name and extract alias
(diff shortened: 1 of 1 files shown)


CedricYii:
It doesn't work with eagerLoading relations.

cebe:
doesn't this already work like this?

```php
$customer = Customer::find()
                   ->alias('c')
                   ->innerJoinWith('order o') 
                   ->innerJoinWith('order.product p') 
                   ->addOnCondition('o.amount' => 100)
```

CedricYii:
Not really. Eventhough in the simple example you provided, it might work but looks hacky and unintuitive. 

As far as I understand, it would not work with a second alias on the same relation. For example:
````php
$customer = Customer::find()
                   ->alias('c')
                   ->innerJoinWith('order o.product p') 
                   ->innerJoinWith('order v.service s') 
                   ->addOnCondition('o.amount' => 100)
                   ->addOnCondition('v.amount' => 500)
````

samdark:
Won't make it into 2.0.

PR #11438, opened 2016-04-28
Markdown set default flavor

mdmunir:
| Q | A |
| --- | --- |
| Is bugfix? | yes/no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |


mdmunir:
property `Markdown::$defaultFlavor` is introduced since the begining. But cannot be used until now :D

cebe:
good catch :-D

SilverFire:
How about a changelog line?

mdmunir:
> How about a changelog line?

That always be my missing. Because i am not good to describe something :D

SilverFire:
Thank you, @mdmunir!
Looks good to me

cebe:
> Because i am not good to describe something :D

you can only get better when you practise ;)

cebe:
Merged. thank you!

PR #11439, opened 2016-04-28
better unit test for generateRandomBytes()

cebe:
This is a better test for `generateRandomBytes()` which tries to test as much branches of the method as possible using some tricks with overriding `function_exists()`.

tests are passing locally for me, want to check what travis says and it would be nice if someone could check which cases are not testable on windows.

There is also one line of code in the new implementatoin which I am unsure about:
https://github.com/yiisoft/yii2/blob/master/framework/base/Security.php#L527
I could not come up with a failing test but I believe it must be 

``` php
                $stillNeed -= StringHelper::byteLength($someBytes);
```

and not `$buffer`.

@tom-- @yiisoft/core-developers could you please verify this?

Postponing the 2.0.8 release until this is sorted out.

### DIFF
--- a/framework/base/Security.php
+++ b/framework/base/Security.php
@@ -438,16 +438,17 @@ class Security extends Component
      */
     public function generateRandomKey($length = 32)
     {
-        if (function_exists('random_bytes')) {
-            return random_bytes($length);
-        }
-
         if (!is_int($length)) {
             throw new InvalidParamException('First parameter ($length) must be an integer');
         }
 
         if ($length < 1) {
             throw new InvalidParamException('First parameter ($length) must be greater than 0');
+        }
+
+        // always use random_bytes() if it is available
+        if (function_exists('random_bytes')) {
+            return random_bytes($length);
         }
 
         // The recent LibreSSL RNGs are faster and likely better than /dev/urandom.
--- a/framework/base/Security.php
+++ b/framework/base/Security.php
@@ -525,7 +525,7 @@ class Security extends Component
                     break;
                 }
                 $buffer .= $someBytes;
-                $stillNeed -= StringHelper::byteLength($buffer);
+                $stillNeed -= StringHelper::byteLength($someBytes);
                 if ($stillNeed === 0) {
                     // Leaving file pointer open in order to make next generation faster by reusing it.
                     return $buffer;
--- a/tests/framework/base/SecurityTest.php
+++ b/tests/framework/base/SecurityTest.php
@@ -7,6 +7,12 @@
 
 namespace yii\base;
 
+/**
+ * emulate availability of functions, to test different branches of Security class
+ * where different execution paths are chosen based on calling function_exists.
+ *
+ * This function overrides function_exists from the root namespace in yii\base.
+ */
 function function_exists($name) {
 
     if (isset(\yiiunit\framework\base\SecurityTest::$functions[$name])) {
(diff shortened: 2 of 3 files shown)


framework/base/Security.php
```diff
@@ -450,6 +446,11 @@ public function generateRandomKey($length = 32)
             throw new InvalidParamException('First parameter ($length) must be greater than 0');
         }
 
+        // always use random_bytes() if it is available
```
samdark: Should not be moved.

samdark: Actually this file should not be changed if that's about tests...

cebe: I moved it behind the checks to allow consistency in exception throwing. PHP7 will cause `\Error` while php5 will cause `yii\base\InvalidParamException`.

samdark: Ah, OK. That's fine.

cebe: see the `testRandomKeyInvalidInput` test.

samdark:
Ran tests on Windows 10:

```
There was 1 error:

1) yiiunit\framework\base\SecurityTest::testGenerateRandomKey with data set #7 (array(false, false, false))
yii\base\Exception: Unable to generate a random key

C:\dev\yii2\framework\base\Security.php:538
C:\dev\yii2\tests\framework\base\SecurityTest.php:879
C:\dev\yii2\vendor\phpunit\phpunit\phpunit:47
```

tests/framework/base/SecurityTest.php
```diff
@@ -5,6 +5,18 @@
  * @license http://www.yiiframework.com/license/
  */
 
+namespace yii\base;
+
+function function_exists($name) {
```
klimov-paul: Would be nice to have an explanation comment here. It is hard to understand the meaning of this trick at the first glance.

tests/framework/base/SecurityTest.php
```diff
@@ -799,9 +823,70 @@ public function testEncryptByPasswordCompat($password, $data, $encrypted)
...
+            [ ['random_bytes' => false, 'openssl_random_pseudo_bytes' => false, 'mcrypt_create_iv' => false ] ],
+        ];
+    }
+
+    /**
+     * @dataProvider randomKeyVariants
+     */
+    public function testGenerateRandomKey($functions)
+    {
+        foreach($functions as $fun => $available) {
```
klimov-paul: Code style: missing space after `foreach`

tests/framework/base/SecurityTest.php
```diff
@@ -799,9 +823,70 @@ public function testEncryptByPasswordCompat($password, $data, $encrypted)
...
+            if ($available && !\function_exists($fun)) {
+                $this->markTestSkipped("Can not test generateRandomKey() branch that includes $fun, because it is not available on your system.");
+            }
+        }
+        static::$functions = $functions;
+
+        // test various string lengths
+        for($length = 1; $length < 64; $length++) {
```
klimov-paul: Code style: missing space after `for`

samdark:
PHP 7.0.5, mcrypt enabled

klimov-paul:
Overall looks good.

samdark:
This failure is expected on Windows. Should be reflected in tests.

samdark:
Excellent change overall.

cebe:
I'm glad you like my tests but I'd love to have a comment on the actual problem pointed out above about  https://github.com/yiisoft/yii2/blob/master/framework/base/Security.php#L527  :)

samdark:
Yes. Seems you're right about that line. Should be `$buffer`.

cebe:
> Yes. Seems you're right about that line. Should be $buffer.

found a way to break it :) same hack for fopen and fread.

Now we have complete test coverage for /dev/urandom case. I also fixed the issue with buffer.

cebe:
@samdark can you re-run test on windows and see if all passes?

samdark:
Now it runs fine on Windows.

tom--:
Yes, you're right about that bug. When I fixed that line in the `fread-random-device` of https://gist.github.com/tom--/35f8a029d77467559cf2 I did

                $stillNeed = $length - StringHelper::byteLength($someBytes);

which I think is easir to understand than with `-=`

tom--:
The unit tests look great. Smart testing idea. I knew we kept you around for something :p

framework/base/Security.php
tom--: I struggled with this problem too but decided it was better to do what PHP 7 does in PHP 7. But I guess we cannot honor to our API spec otherwise.

cebe:
> $stillNeed = $length - StringHelper::byteLength($someBytes);

here it should be $buffer again because that is permanent.

tom--:
Gah! yes.

Issue #11387, opened 2016-04-20
Oracle (oci) charset not respected 


SSiwek:
Hi,

please be lenient with my English. Thank you.

I connect to an Oralce Instance wich is not in charset utf8.

I dont get the correct chars for the German special chars ÄÖÜ ...

I tried to give the "charset Attribut" but it is not respected in connecton.php .

I inspected the source and found a good place to add that 
With my modification I get the German Umlaute.

connection.php in function createPdoInstance()

I added the part with oci.

[...]
```php
            if (isset($driver)) {
                if ($driver === 'mssql' || $driver === 'dblib') {
                    $pdoClass = 'yii\db\mssql\PDO';
                } elseif ($driver === 'sqlsrv') {
                    $pdoClass = 'yii\db\mssql\SqlsrvPDO';
                }
                if ($driver==='oci' && $this->charset !== null ){
                	$this->dsn .= ";charset=".$this->charset;
                }
            }
```

I think, that it can be added to the source...

thank you
Sven

samdark:
@SSiwek do you have time to work on a pull request?

SSiwek:
I havent done that before.
I have read some docu now and I am unsure how to create a branch to put my modification in. I installed yii2 via composer only....

Sven

samdark:
Well, here's the guide: https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md If it's too much for you at this stage, let us know so others would be able to work on it.

SSiwek:
Hi,
I have done the "Prepare your development enviroment" Part of git-workflow.md
Now I am on "Working on bugs and featrures" 
Points 1 to 4 are done. 
Now I am stuck on "5. Update Changelog".  I dont have a number "enh #999:"  and my version of the file CHANGELOG is missing the Section "Work in progress"
Maybe I have done something wrong? I work with Egit/Eclipse

samdark:
1. You need a fresh code from upstream.
2. The number for changelog is a number of issue. In this case it's `#11387`.

SSiwek:
Hi,
I deleted all lokal files and starteted all over. I have done all steps from git-workflow.md.
I still dont have "Work in progress" in CHANGELOG. I am giving up.
Sven

cebe:
it's actually `2.0.8 under development`, no need to be so strict about the changelog, concentrate on the code, we can fix the rest ourselfs :)

cebe:
[...]

PR #11443, opened 2016-04-28
11387 oracle charset not respected

SSiwek:
#11387


SilverFire:
Well... I believe we can found a better place for this change.

SSiwek:
There is no other place to tell Oracle the charset. I found the only solution in the connectstring,
And because of that, I put it there.

cebe:
better to be documented in $charset and added to $dsn in config. https://github.com/yiisoft/yii2/issues/11387#issuecomment-230742133

PR #11462, opened 2016-04-29
 Added support of filtering rules to yii\log\Target::$logVar

HaruAtari:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

Target will exclude this keys from logVars.
For example if we need to exclude user's password on login error. Or other confidential data which shouldn't be in logs.


framework/log/Target.php
```diff
@@ -60,6 +60,12 @@
      */
     public $logVars = ['_GET', '_POST', '_FILES', '_COOKIE', '_SESSION', '_SERVER'];
     /**
+     * @var array list of keys which will be excluded from [[logVars]]
+     * Default to []
+     * @see lofVars
```
samdark: lofVars?

samdark:
I feel it could be done otherwise. For example, we could allow specifying `globalname.key`, `!globalname.key` and `globalname.*`.

framework/log/Target.php
HaruAtari: I fexed this. https://github.com/yiisoft/yii2/pull/11462/commits/7bc5ae4881e7cea80285daf28fea9cab09db8142

HaruAtari:
I don't fully understand. What you mean?
If we specify `public $logVarsHiddenKeys = ['_POST.hidden_key'];` - `$_POST['hidden_key']` will be excluded.
If we specify `public $logVarsHiddenKeys = ['_POST.*'];` - all `$_POST` variables will be excluded.
If we specify `public $logVarsHiddenKeys = ['!_POST.hidden_key'];` - all `$_POST` variables will be excluded except `hidden_key`.
DId I understand you correctly?

HaruAtari:
If I am right why should we use `public $logVarsHiddenKeys = ['_POST.*'];`? We can exclude `_POST` from `logVars` instead this.

samdark:
I mean instead of introducing `$logVarsHiddenKeys` we may extend `$logVars` syntax.

HaruAtari:
Now I understand. But may be we should use `globalname` instead of `globalname.*` for BC?

`globalname` - all `globalname` variables.
`globalname.key` - only `key`
`!globalname.key` - all `globalname` variables except `key`.

samdark:
I'd process `!` differently i.e. in order to get all `globalname` except `key`:

```php
['globalname' , '!globalname.key']
```

HaruAtari:
Should i do anything else?

samdark:
I think no. @yiisoft/core-developers need review/suggestions.

framework/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Yii Framework 2 Change Log
 - Enh #11432: Added HTTP status 421 "Misdirected Request" to list of statuses in `yii\web\Response` (dasmfm)
 - Enh #11438: Configurable `yii\helpers\Markdown` default flavor (mdmunir)
 - Bug #11459: Fixed flash messages not destroyed when `session.auto_start = 1` set in php.ini (cartmanchen)
+- Enh: Elements of `yii\log\Target::$logVar` can be in different forms. (Viktor Pikaev)
```
cebe: this message is quite unclear. should be more like PR title.

cebe:
how about integrating this functionality into VarDumper or ArrayHelper?

framework/CHANGELOG.md
HaruAtari: @cebe I don't know english well. How can it be written?

HaruAtari:
We can add method `BaseArrayHelper::filter($array, $fields)`. Move filter logic to it. And calling this method from `Target::getContextMessage()` by `ArrayHelper::filter($GLOBALS, $this->logVars)`.

HaruAtari:
I did this.

samdark:
Merged and then adjusted a bit in https://github.com/yiisoft/yii2/commit/4fd4ac04b1e3c367e585eedd373b7e4938fe9297. Thanks!

PR #11470, opened 2016-04-30
#4479 REST filtering support by involving a search class

tunecino:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | not covered but should pass |
| Fixed issues | none. ENH related to #4479 |

**Usage:**

``` php
namespace app\controllers;

class UserController extends \yii\rest\ActiveController
{
    public $modelClass = 'app\models\User';
    public $searchClass = 'app\models\UserSearch'
}
```

Then being sure that the search class is using `$this->load($params, '' )` instead of `$this->load($params)` (as generated by _gii_) **or** by alternatively overriding `formName()` in model:

``` php
public function formName()
{
    return '';
}
```

**Results:**
`yii\rest\IndexAction` should use that search class to filter results (if is set) when doing `/resources?attribute1=value&attribute2=value2&...` as defined and handled in the search class where also validations, scenarios and custom rules can be maintained.

**Note:**
The search class existed and has been widely used since Yii 1.x but as a simple example generated by _gii_ and as first was designed by @qiangxue. I think it is time to make it implement an interface as we should expect the class to have a search method in this case or in future cases.


framework/data/SearchInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\data;
+
+/**
+ * The SearchInterface must represent the model behind search forms.
+ *
+ * @author Qiang Xue <qiang.xue@gmail.com>
+ * @since 2.0
```
cebe: since 2.0.9

framework/data/SearchInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+interface SearchInterface
+{
+    /**
+     * Creates data provider instance with search query applied
+     *
+     * @param array $params
```
cebe: need description that these are the params from search query.

framework/rest/ActiveController.php
```diff
@@ -42,6 +42,11 @@ class ActiveController extends Controller
      */
     public $modelClass;
     /**
+     * @var string the search class related to the model. This property is not required.
+     * @see \yii\rest\IndexAction::$searchClass
+     */
```
cebe: `@since 2.0.9`

framework/rest/IndexAction.php
```diff
@@ -30,6 +30,28 @@ class IndexAction extends Action
...
+     * ```php
+     * public function formName()
+     * {
+     *    return '';
+     * }
+     * ```
+     *
+     * @see \yii\base\Model::formName()
+     */
+    public $searchClass;
```
cebe: `@since 2.0.9`

framework/rest/IndexAction.php
```diff
@@ -30,6 +30,28 @@ class IndexAction extends Action
...
     public $prepareDataProvider;
+    /**
+     * @var string class name of the search class related to this model. It can be optionally involved but not required.
+     * The called search class should implement [[\yii\data\SearchInterface]] to prepare an instance of [[ActiveDataProvider]].
+     * If not set, the default data provider returned by [[prepareDataProvider()]] will be used instead.
+     *
+     * Note that [[\yii\base\Model::formName()]] is used by default in a search class generated by **gii** when retrieving the attributes names from the URL query request.
+     * If the URL endpoint should look like `/resources?attribute=value` instead of `/resources?FormName[attribute]=value`
+     * then be sure that your search class is using `load($params, '')` instead of `load($params)`.
+     * @see \yii\base\Model::load()
```
cebe: `@see` does not work in the middle of a description. Should be a sentence `See [[\yii\base\Model::load()]] for more details.` instead.

cebe:
This looks really interesting, good job :)

tunecino:
thanks @cebe. I'll what commented and re-push

tunecino:
done. didn't mention

> params from search query.

in the interface because it is also used within forms and maybe not used only with user inputs.

There is also another idea of making it:
```php
public function search($params, $formName = null)
```
if `$formName` is null it will use model's `formName()` otherwise will do `$this->load($params,$formName)`. what do you think ?

framework/data/SearchInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\data;
+
+/**
+ * The SearchInterface must represent the model behind search forms.
+ *
+ * @author Qiang Xue <qiang.xue@gmail.com>
```
Faryshta: the author is not @qiangxue it would be @tunecino

Faryshta: also I don't see this interface being used anywhere.

Faryshta:
For rest I see little use to the `FormName[paramName]` notation since you will be handling just one model at the time.

Another thing. When handling validation errors you need to return the search model, not the data provider.

```php
        if ($this->searchClass !== null) {
            // won't return errors, instead it will dump all results or none at all
            // and the validations will be silently ignored
            return (new $this->searchClass)->search(Yii::$app->request->queryParams);
        }

        // in my personal project I use
        if ($this->searchClass !== null) {
            $searchModel = new $this->searchClass;
            $searchModel->load(Yii::$app->request->queryParams, ''); // we can debate the last param
            if ($searchModel->validate()) {
                // return data
                return $searchModel->search([]);
            } else {
                // the `yii\rest\Serialize` class will show the validation errors and change the http status code
                return $searchModel;
             }
        }
```

fernandezekiel:
good job so far...
are we going to have some default searching features so that the developer wouldn't have to implement the search() by itself?
something like
```
trait Searching
{
	/**
     * @param array $params
     * @return \yii\data\ActiveDataProvider
     */
    public function search($params = [])
    {
    	$query = static::find();
        $this->unsetAttributes();
        $this->setAttributes($params);

        $attributes = array_keys($this->getAttributes());

            $query->andFilterWhere(['like', attribute, $this->{$attribute}]);
        }

        return new ActiveDataProvider(['query' => $query]);
    }
}
```

Faryshta:
@fernandezekiel that sounds like something to discuss on another issue. I think it wasn't possible before due to the numeric fields but now it can be done using `filterCompareAttribute()` method.

tunecino:
Thanks for feedback :+1: 

@fernandezekiel 

I agree with @Faryshta it is a different feature than this one. but I personally prefer flexible solutions by design so I can extend or redesign over features doing the work instead of me. and that's why I prefer Yii over other frameworks. I mean if I can't redesign it or if it is doing other classes job like what should ActiveQuery do or what does the the SearchClass (even if it doesn't really exist and is just a proposed code by gii) then I think it should be better implemented as an extension.

@Faryshta 

I kept @qiangxue name in the interface because nowhere in the framework it says who made the search class (at least as idea) and I will add my name next to his as soon as I add or change a single line of code different from what already generated by gii.



> I don't see this interface being used anywhere.

yes not yet used. I think we should make the gii extension implement it to the generated search classes if this get merged. plus as we are expecting a `search()` method, it is better to tell in docs that your class should implement that interface instead of telling: "please make it look exactly like the example gii did generate".



> For rest I see little use to the FormName[paramName] notation since you will be handling just one model at the time.

exactly. but because the generated search class is using `$this->load($params)` by default which makes urls looking like `/resource?formName[paramName]=value` and because the same search class may be used for rest and for forms I first thought it make sense to make it more flexible so we can do this in rest:
```php
return (new $this->searchClass)->search(Yii::$app->request->queryParams, '' );
```
But now after seeing your code I think it make more sense to do it like you've done instead:
```php
if ($this->searchClass !== null) {
    $searchModel = new $this->searchClass;
    $searchModel->load(Yii::$app->request->queryParams, '');
    return $searchModel->search([]);
}
```



> When handling validation errors you need to return the search model, not the data provider.

The problem here is that not everyone want to throw errors if validation fails. by default the generated search class has those lines:

```php
if (!$this->validate()) {
    // uncomment the following line if you do not want to return any records when validation fails
    // $query->where('0=1');
    return $dataProvider;
}
```
and yes people may want to just tell 'no results found' when input has a string instead of  an integer.
so I think it is better left that way and in case you need to throw errors it would be as simple as overriding the previous lines in the search class by this:

```php
if ($this->validate() === false) return $this;
```
that will throw the `422` same way. it just need to be documented.

Faryshta:
>by default which makes urls looking like 

How about a new param for the `RestController` and `IndexAction`? `searchFormName` which by default will be `null` (means it will trigger default functionality` and you can simply set it to `''`

tunecino:
but isn't a rare case that someone wants his REST urls to look like `?FormName[paramName]=value` ?
I'd prefer the way how you did it:
```php
$searchModel->load(Yii::$app->request->queryParams, '');
return $searchModel->search([]);
```
and if someone need to use formNames in urls he can always override the IndexAction. loading with empty string as formName is how it is also done in the other Rest actions.

tunecino:
unless you think it is wired to do:
```php
return $searchModel->search([]);
```
in that case we should change the search class so we can do `$searchModel->search($params, '' )`. see https://github.com/yiisoft/yii2/pull/11470#issuecomment-215986989

Faryshta:
@tunecino i agree with what you wrote more but it would be a BC break, don't know if the core devs would agree

tunecino:
the argument is optional. if *gii* implements the new interface it will be for new code only. the only BC I can think about so far is when people are already using a `$searchClass` in their controllers without the second argument. so maybe it can be considered as BC. I'm not really sure about. we can change the search class and leave it to millstone 2.1 or just do it the way you did and work with the search class as already generated by *gii*.

Faryshta:
@cebe @samdark @SilverFire what do you guys think?

niksamokhvalov:
Another model isn't always a convenient option for filtering the collections. It would be convenient to have the basic functionality of filtering.

framework/rest/IndexAction.php
```diff
@@ -54,6 +77,10 @@ protected function prepareDataProvider()
             return call_user_func($this->prepareDataProvider, $this);
         }
 
+        if ($this->searchClass !== null) {
```
samdark: Should be `$this->searchClass instanceof SearchInterface`.

framework/data/SearchInterface.php
samdark: All the search classes should implement this interface.

tunecino:
@niektenhoopen I don't know if there is better alternatives but this is why I think using a search model class is a good idea:

1. It has been used with web controllers since yii1 so I guess most yii developers are familiar with it. and I personally found it a good practice to maintain such logic in a separate model same as the ActiveQuery models.
2. It allows validating user inputs so only attributes defined in rules() will be used when filtering and it make it easier to solve advanced validation cases like validating each arrays elements, implementing custom validation rules when none of [core validators](http://www.yiiframework.com/doc-2.0/guide-tutorial-core-validators.html) can help or using different scenarios, ...
3. it is easier to throw errors in some cases instead if returning 0 results like @Faryshta's example here: https://github.com/yiisoft/yii2/issues/4479#issuecomment-207914197 which can be done by simply returning the model itself: see https://github.com/yiisoft/yii2/pull/11470#issuecomment-216038368.
4. Filtration may not always be tied to the model itself but may also involve related models which may require advanced SQL queries like maybe in [this example](https://github.com/tunecino/Yii2_foundation-apps/blob/master/backend/models/searches/ImageSearch.php#L66-L78).

samdark:
#12251

framework/rest/IndexAction.php
```diff
@@ -54,6 +78,10 @@ protected function prepareDataProvider()
             return call_user_func($this->prepareDataProvider, $this);
         }
 
+        if ($this->searchClass instanceof SearchInterface) {
+            return (new $this->searchClass)->search(Yii::$app->request->queryParams);
+        }
```
klimov-paul: To think: `IndexAction::prepareDataProvider` can be used for this.
Thus, it does not seem this entire PR make much sense.

tunecino: if you mean https://github.com/yiisoft/yii2/issues/4479#issuecomment-248318709 then I agree. It is much simpler. I think it just need to be documented somewhere.

tunecino:
https://github.com/yiisoft/yii2/issues/4479#issuecomment-248318709

PR #11476, opened 2016-05-01
Object to array with nested expand

mdmunir:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | yes |
| Fixed issues | #9642, #8061, #6844 |

### DIFF
--- a/framework/base/Arrayable.php
+++ b/framework/base/Arrayable.php
@@ -88,5 +88,5 @@ interface Arrayable
      * @param boolean $recursive whether to recursively return array representation of embedded objects.
      * @return array the array representation of the object
      */
-    public function toArray(array $fields = [], array $expand = [], $recursive = true);
+    public function toArray(array $fields = [], array $expand = [], array $except = [], $recursive = true);
 }
--- a/framework/base/ArrayableTrait.php
+++ b/framework/base/ArrayableTrait.php
@@ -113,10 +113,10 @@ trait ArrayableTrait
      * @param boolean $recursive whether to recursively return array representation of embedded objects.
      * @return array the array representation of the object
      */
-    public function toArray(array $fields = [], array $expand = [], $recursive = true)
+    public function toArray(array $fields = [], array $expand = [], array $except = [], $recursive = true)
     {
         $data = [];
-        foreach ($this->resolveFields($fields, $expand) as $field => $definition) {
+        foreach ($this->resolveFields($fields, $expand, $except) as $field => $definition) {
             $data[$field] = is_string($definition) ? $this->$definition : call_user_func($definition, $this, $field);
         }
 
@@ -124,7 +124,7 @@ trait ArrayableTrait
             $data['_links'] = Link::serialize($this->getLinks());
         }
 
-        return $recursive ? ArrayHelper::toArray($data) : $data;
+        return $recursive ? ArrayHelper::toArray($data, [], true, $expand, $except) : $data;
     }
 
     /**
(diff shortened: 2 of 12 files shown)


Faryshta:
I think we need to change the `$expand` logic to make it more end user friendly. For example how do I tell the interface that I want the `customer.fullName` but not `customer.password_hash` on a rest request?

I think we should merge the usage of the parameter `$fields` and `$expand` into just `$fields` and use it like this.

```php
Order::findOne(1)->toArray([
    'id',
    'product_id',
    // loads only the `name` field from the `product`relation in order.
    'product.name',
    // loads all the fields in the `category` relation in products.
    'product.category',
    'customer_id',
    // loads only `email` and `fullName` from `customer` relation
    'customer.email',
    'customer.fullName',
]);
```

its much simplier to understand.

@mdmunir idea for an `$except` idea is one of those "so obvious no one though about it before" ideas and can be used with this change too.

Order::findOne(1)->toArray([
    'id',
    'product_id',
    'customer_id',
    'product.category',
    'customer',
], [
    'product.description',
    'customer.password_hash',
    'customer.auth_key',
]);
```

This is even simplier to understand than the previous example.

**Comment by mdmunir** [comment]
> I think we should merge the usage of the parameter $fields and $expand into just $fields and use it like this.

Aggre. To handle except field, IMO we should remove `expands()` from `Arrayable` and add `excepts()`. I mean, all property (defined by getter and setter) should be avaliable to expand with out define it to `expands()`. Then all field that defined in `except()` should not be showed for all condition.

But, it also give another problem. In case `tests/framework/base/ArrayableTest`. To show all order field and expand `customer`, we should write all as parameter value. May be we should introduce `*` to handle all field defined in `fields()`.
```php
Order::findOne(1)->toArray([
    // represent to all Order::fields() definition
    '*', 
    // loads only the `name` field from the `product`relation in order.
    'product.name',
    // loads all the fields in the `category` relation in products.
    'product.category',
    // load all customer::fields() and profile relation in customer
    'customer.*', // or just 'customer'
    'customer.profile', // profile defined by getter
]);
```

Faryshta:
@mdmunir wanna work with me on a different PR to see which one gets more attention and has easier code to implement?

Faryshta:
Also just ocurred to me that we can avoid the `$except` parameter if we introduce a new syntax to avoid adding new parameters from the `$fields` property, for example `-password_hash` or `!password_hash` means to NOT bring you that attribute.

mdmunir:
@Faryshta 
need more core developer feedback. But, go ahead if you want to create another PR

mdmunir:
I also suggest to reorder `ArrayHelper::toArray()` parameters.
```php
public static function toArray($object, $expand = [], $except = [], $properties = [], $recursive = true)
{

```

SilverFire:
Agree with @Faryshta thoughts. Did not dig in too deeply yet, so have nothing to say about the approach

mdmunir:
@SilverFire 
I also agree. More simple is more better. But, if i want to add extra field, i must write all field plus extra field. Is not good to me. It can be solve with introduce `*` to represet all standart field.
I will change my PR

dynasource:
because of these many arguments, this approach is too BC prone:

```php
public static function toArray($object, $expand = [], $except = [], $properties = [], $recursive = true)
```

I would rather see a config variable as the second argument. It would also easier to extend.

samdark:
Won't make it into 2.0.

PR #11477, opened 2016-05-01
Use PHP 5.5+ `::class` instead of full class namespace

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |


Faryshta:
notice that ::class is determined on compilation time, not execution time, that means the classes won't be autoloaded using this change.

SilverFire:
Aha, we will accept this PR. Actually it's the only places where it should be changed:

```bash
grep -F -ir --exclude="classes.php" "'yii\\" framework
```

cebe:
> Use PHP 5.6+ `::class` instead of full class namespace

it's a 5.5 feature ;)

@SilverFire you mean it's **not** the only places, right? :)

SilverFire:
Oops :)
Yes, __not__ the only places

Faryshta:
I made the changes suggested. Also can we reconsider #11487. I think its consistent to have a `->class` property on an object since we have now an static way to call the class

cebe:
Thank you!

PR #11487, opened 2016-05-03
[Enh] getClass() method

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |

instead of `get_class($model)` we can obtain the class of a model by `$model->class` if this is approved i will make another pr to modify all the `get_class($object)` calls

### DIFF
--- a/framework/base/Object.php
+++ b/framework/base/Object.php
@@ -77,6 +77,15 @@ use Yii;
 class Object implements Configurable
 {
     /**
+     * Returns the fully qualified name of this class.
+     * @return string the fully qualified name of this class.
+     */
+    public static function getClass()
+    {
+        return static::class;
+    }
+
+    /**
      * Constructor.
      * The default implementation does two things:
      *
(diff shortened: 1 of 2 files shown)


SilverFire:
This change means introducing of one more reserved attribute name in models. I don't think it's a good idea.

I'm closing the PR but if anybody have something to add - feel free to express your thoughts

cebe:
calling `get_class($model)` is just fine, why would we need an extra methods that does the same?

Faryshta:
@cebe there is nothing wrong with using tha tcall, i just don't like to see it since it feels as if it stop being OOP.

@SilverFire I unit tested it and it doesn't produce any issue or error. I am not introducing an attribute, we are just using the `__get()` class to emulate it. So its not breaking any rule for reserved words.

Faryshta:
there is a commit here https://github.com/Faryshta/yii2/commits/getClass that is not being parsed on this PR. Any idea why?

PR #11493, opened 2016-05-03
Object to array with nested expand

mdmunir:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | - |
| Fixed issues | #11476 |

merge `$fields` and `$expand` into just `$fields`. I split from another PR

### DIFF
--- a/framework/base/Arrayable.php
+++ b/framework/base/Arrayable.php
@@ -88,5 +88,5 @@ interface Arrayable
      * @param boolean $recursive whether to recursively return array representation of embedded objects.
      * @return array the array representation of the object
      */
-    public function toArray(array $fields = [], array $expand = [], $recursive = true);
+    public function toArray(array $fields = [], array $expand = [], array $except = [], $recursive = true);
 }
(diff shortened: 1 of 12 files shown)


framework/base/Arrayable.php
```diff
@@ -82,11 +82,11 @@ public function extraFields();
      *
      * @param array $fields the fields that the output array should contain. Fields not specified
      * in [[fields()]] will be ignored. If this parameter is empty, all fields as specified in [[fields()]] will be returned.
-     * @param array $expand the additional fields that the output array should contain.
+     * @param array $except the additional fields that the output array should contain.
```
samdark: I don't think this name change is needed. These are still list of fields, not exceptions.

mdmunir:
@samdark 
How to define exception fields?

samdark:
As I can see, logic of this field wasn't changed much while name was changed and I wonder if it should be changed. That's it.

mdmunir:
No. parameter `$fields` and `$expand` are merge to just `$fields` then introduce new parameter `$except`.
```php
// previous version
$model->toArray([],['customer', 'product']);

// current version
$model->toArray(['*', 'customer', 'product']);
```

Faryshta:
@samdark the problem with the `$expand` parameter is that it is not intuitive now to call certain fields inside an extra field... and if we can call fields inside the `$expand` parameter then we don't need the `$fields` parameter at all since we can just use `$expand`.

We can just use the `fields()` and `extraFields()` in the model to convert to determine when a parameter can be expanded or is a final field

Issue #11505, opened 2016-05-05
ActiveQuery create model not through dependency injection


Liv1020:
### What steps will reproduce the problem?
```
\Yii::$container->set('app\models\Order', 'app\resources\Order');

$product = Product::findOne(1);
$orders = $product->orders;

//$order not app\resources\Order
```
### What is the expected result?
```
\Yii::$container->set('app\models\Order', 'app\resources\Order');

$product = Product::findOne(1);
$orders = $product->orders;

//$order == app\resources\Order
```
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.8
| PHP version      | 5.6
| Operating system | centos7

Liv1020:
yii core should call any class through dependency injection(Yii::createObject())

cebe:
no, it should not. If you want to replace an existing class in the framework, you can set it in `Yii::$classMap`. No idea why you would do that in your application...

PR #11506, opened 2016-05-05
ActiveRecord relations support dependency injection

pavlelee:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes/no |
| Tests pass? | yes |
| Fixed issues | https://github.com/yiisoft/yii2/issues/11505 |

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -1089,7 +1089,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      */
     public static function instantiate($row)
     {
-        return new static;
+        return \Yii::createObject(static::className());
     }
 
     /**


Faryshta:
whats the use case?

lynicidn:
fix it also for active query
echo User::find()->model_class

Liv1020:
@Faryshta if you injection class, but ActiveRecord relations not support.
example:
https://github.com/yiisoft/yii2/issues/11505

samdark:
Description in #11505 is a bit vague. I understand technical part of it but why do you need it?

Faryshta:
I mean why not use `\app\resources\Order::findOne(1)` instead of `\app\models\Order::findOne(1)` ?

Liv1020:
@samdark   
I inject a class(app\resources\Order) to container,a product model has related attribute named getOrders():common\models\Order. then I called $product->orders, it should return common\models\Order[], Ideally this would be app\resources\Order. 
because I have changed common\models\Order logically,it's like usage :\Yii::$container->set('yii\data\Sort', 'app\modules\xg\components\Sort');

Liv1020:
@Faryshta  because I called $product->orders, not Order::find()

samdark:
@Liv1020 I perfectly understand technical side of it. I'm asking for the reasons doing so.

Liv1020:
@samdark Reasons?Since the use of the injection , of course, every place using an injection classes , rather than in some places .

Liv1020:
@samdark laravel use injection anywhere,I do not know my bad English, can you understand?

klimov-paul:
Dependency injection at ActiveRecord layer will produce huge performance impact.

> I inject a class(app\resources\Order) to container,a product model has related attribute named getOrders():common\models\Order. then I called $product->orders, it should return common\models\Order[], Ideally this would be app\resources\Order. 

Why do not declare relation to `app\resources\Order` at the first place? What is a reason for such twisted approach?

samdark:
I don't like the idea of using DI everywhere for the sake of using it w/o any real need.

lynicidn:
ar isolated from all, only service locator
http://programmers.stackexchange.com/questions/119352/does-the-activerecord-pattern-follow-encourage-the-solid-design-principles

is pattern problem

Liv1020:
@samdark @klimov-paul 
Your views I carefully thought for a moment, and finally I use a rewrite every relation attributes to solve rest expand returned

Faryshta:
> Your views I carefully thought for a moment, and finally I use a rewrite every relation attributes to solve rest expand returned

how i solved it.

I have `common\models\Orders` which are used in `backend` and `frontend` applications `api\models\Orders` for my `api` application.

```php
   public static $orderClass = 'common\\models\\Orders';

   public function getOrders()
   {
       return $this->hasMany(static::$orderClass, ['id' => 'order_id']):
   }

   // on api

   Client::$orderClass = 'api\\models\\Order';
```

PR #11522, opened 2016-05-07
Skype and phone links generation

matiosfree:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |


samdark:
These aren't as common as mailto and could be achieved easily. Sorry, but won't merge it.

PR #11530, opened 2016-05-10
calling getValue() for each attribute in AttributeBehavior

h0lg:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | I hope not |
| Tests pass? | haven't checked |
| Fixed issues | none |

calling getValue() for each attribute (instead of once for all attributes) and passing attribute name along as a parameter to value function so that value function can return different values depending on current attribute

### DIFF
--- a/framework/behaviors/AttributeBehavior.php
+++ b/framework/behaviors/AttributeBehavior.php
@@ -34,8 +34,8 @@ use yii\db\ActiveRecord;
  *                 ActiveRecord::EVENT_BEFORE_INSERT => 'attribute1',
  *                 ActiveRecord::EVENT_BEFORE_UPDATE => 'attribute2',
  *             ],
- *             'value' => function ($event) {
- *                 return 'some value';
+ *             'value' => function ($event, $attributeName) {
+ *                 return $attributeName === 'attribute1' ? 'some value' : 'some other value';
  *             },
  *         ],
  *     ];
@@ -73,7 +73,7 @@ class AttributeBehavior extends Behavior
      * The signature of the function should be as follows,
      *
      * ```php
-     * function ($event)
+     * function ($event, $attributeName)
      * {
      *     // return value will be assigned to the attribute
      * }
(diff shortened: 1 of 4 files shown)


cebe:
what is the use case for this? wouldn't it be better to add multiple behaviors in this case?

forger:
Just seeing that this breaks the build. I'll fix this.
Anyway - use case: I want to return a value for an attribute depending on the current value of the attribute. That is, I want to a type conversion. I want to convert a DateTime to a time stamp or string on beforeUpdate/beforeInsert and the timestamp/string back into a DateTime on afterFind.

forger:
@cebe you asked:
> wouldn't it be better to add multiple behaviors in this case?

The behavior is actually the same for each attribute. I just want to be able to change an attribute's value depending on the original value of the attribute. So running the value function once for all attributes won't do because that would set the same new value on all attributes. So I want to have the attribute's name in the value function to identify the attribute`to get its value. Does this make sense to you?
 I can of course do that with a custom attribute which is what I'm doing right now. I just thought the existing AttributeBehavior could benefit from this. Any thoughts?

cebe:
it could maybe, even though that is out of the original scope of it. Need to check it in more detail.

forger:
I basically suggested the change to AttributeBehavior so I can do what I've done in the following subclass. Maybe that makes it easier to understand than my previous comment does.

`final class DateTimeConversionBehavior extends AttributeBehavior
{
    const DATETIME_FORMAT = "Y-m-d H:i:s";
    public $dateTimeAttributes;

    function init()
    {
        parent::init();

        if (empty($this->attributes)) {
            $this->attributes = [
                BaseActiveRecord::EVENT_BEFORE_INSERT => $this->dateTimeAttributes,
                BaseActiveRecord::EVENT_BEFORE_UPDATE => $this->dateTimeAttributes,
                BaseActiveRecord::EVENT_AFTER_FIND => $this->dateTimeAttributes,
            ];
        }

        if ($this->value === null) {
            $this->value = [$this, 'getAttributeValue'];
        }
    }

    function getAttributeValue($event, $attributeName)
    {
        switch ($event->name) {
            case BaseActiveRecord::EVENT_BEFORE_INSERT:
            case BaseActiveRecord::EVENT_BEFORE_UPDATE:
                return $this->convertToMySqlDateTime($event->sender->$attributeName);
            case BaseActiveRecord::EVENT_AFTER_FIND:
                return $this->convertToPhpDateTime($event->sender->$attributeName);
        }
    }

    function convertToMySqlDateTime($phpDateTime)
    {
        return $phpDateTime === null ?
            null :
            $phpDateTime->format(static::DATETIME_FORMAT);
    }

    function convertToPhpDateTime($mySqlDateTime)
    {
        return $mySqlDateTime === null ?
            null :
            date_create_from_format(static::DATETIME_FORMAT, $mySqlDateTime);
    }
}
`

samdark:
Won't make it into 2.0.

Issue #11429, opened 2016-04-27
The message file for category 'yii' does not exist


MaXL-ru:
I have many errors in log file:
```
The message file for category 'yii' does not exist: ...vendor/yiisoft/yii2/messages/en-GB/yii.php Fallback file does not exist as well: ...vendor/yiisoft/yii2/messages/en/yii.php
```

Yii (2.0.7) `$language` is `en-GB`.

I have already seen that @SilverFire has been already fixed (#7964).

cebe:
do you get this error with yii 2.0.7 ?

MaXL-ru:
Yes.

klimov-paul:
Well, we do not have translations for `en-GB` as well as for `en`:
https://github.com/yiisoft/yii2/tree/master/framework/messages

They are considred as not necessary ones.
If you wish to avoid error messages in logs, you should configure `sourceLanguage` to be `en-GB`.

MaXL-ru:
I already set this in config:

``
'language'  => 'en-GB',
'sourceLanguage' => 'en-GB'
```
but errors happened.

**Comment by cebe** [comment]
what does your message source config look like, i.e. i18n component config.

**Comment by MaXL-ru** [comment]
```
'i18n' => [
      'translations' => [
        'app*' => [
          'class' => 'yii\i18n\PhpMessageSource',
          'basePath' => '@app/messages',
          'sourceLanguage' => 'en-GB'
        ]
      ]
    ]
```

I already tried to comment this config.

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/c3c0d04f5bc58c5ad08da13797e27d956ebec043

PR #11547, opened 2016-05-11
Fixed `yii\i18n\PhpMessageSource::loadFallbackMessages()`

SilverFire:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11429 |

Fixed `yii\i18n\PhpMessageSource::loadFallbackMessages()` error when source and language is same, but locales are different

### DIFF
--- a/framework/i18n/GettextMessageSource.php
+++ b/framework/i18n/GettextMessageSource.php
@@ -103,7 +103,11 @@ class GettextMessageSource extends MessageSource
         $fallbackMessageFile = $this->getMessageFilePath($fallbackLanguage);
         $fallbackMessages = $this->loadMessagesFromFile($fallbackMessageFile, $category);
 
-        if ($messages === null && $fallbackMessages === null && $fallbackLanguage !== $this->sourceLanguage) {
+        if (
+            $messages === null && $fallbackMessages === null
+            && $fallbackLanguage !== $this->sourceLanguage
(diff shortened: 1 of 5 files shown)


framework/i18n/GettextMessageSource.php
```diff
@@ -103,7 +103,11 @@ protected function loadFallbackMessages($category, $fallbackLanguage, $messages,
         $fallbackMessageFile = $this->getMessageFilePath($fallbackLanguage);
         $fallbackMessages = $this->loadMessagesFromFile($fallbackMessageFile, $category);
 
-        if ($messages === null && $fallbackMessages === null && $fallbackLanguage !== $this->sourceLanguage) {
+        if (
+            $messages === null && $fallbackMessages === null
+            && $fallbackLanguage !== $this->sourceLanguage
+            && $fallbackLanguage !== substr($this->sourceLanguage, 0, 2)
```
samdark: Not sure it's correct. See https://github.com/yiisoft/yii2/pull/8057/files#diff-97dd491a4c8c30f7d804fbea50cd0b40R293

SilverFire: We do not support something like `zh-Hant-CN` fully. I think current implementation is OK.
Since you did not update MessageSources in your PR yet, no conflicts should occur.

PR #11549, opened 2016-05-12
array_key_exists problem

timestee:
Double/Float key will be automatically translated to an int. So when
you do $array[1.0] = 2;, $array[1] will be defined.But for
array_key_exists(), this conversion is not done automatically. I
encountered this problem with key 660019  under a weird  environment.

To take the speed advantage of isset() while keeping the reliable
result from array_key_exists(), Perhaps the below code works the best:
isset($array[$key]) || array_key_exists($key, $array)

### DIFF
--- a/framework/helpers/BaseArrayHelper.php
+++ b/framework/helpers/BaseArrayHelper.php
@@ -189,7 +189,7 @@ class BaseArrayHelper
             $key = $lastKey;
         }
 
-        if (is_array($array) && array_key_exists($key, $array)) {
+        if (is_array($array) && (isset($array[$key]) || array_key_exists($key, $array)) ) {
             return $array[$key];
         }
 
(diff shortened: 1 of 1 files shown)


SilverFire:
Oh wow. Interesting thing :)
Could you compose a test that failes before and passes after your fix?

samdark:
I confirm the issue.

samdark:
The test should be setting value with key = `1.0` and then getting it by key `1.0`.

zelenin:
https://github.com/yiisoft/yii2/issues/11739
is it related?

cebe:
could be related. would be good if someone could create a unit test case for this.

samdark:
Merged after @AnikanovD created a pull request with a test. Thank you @AnikanovD and @zsounder.

Issue #8795, opened 2015-06-15
Move Identity Cookie code into separate functions in web\User.php


maine-mike:
Hi,

I tried this last week, but thought it might be better to start over.  Here is a pull request to separate out the cookie-related functions in web\User.php.  This will allow someone to write different cookie handling functions to override the base remember-me cookie functions.

https://github.com/yiisoft/yii2/pull/8714

Thank you!

SilverFire:
Movin to 2.0.8 because the original PR was declined.

maine-mike:
My original enhancement included a bug fix.  It appears that it's a bad idea to mix a bug fix with an enhancement.  So I created this pull request for the bug fix:

https://github.com/yiisoft/yii2/pull/10480

Once that is complete, I'll create a new pull request for this enhancement.  Thank you.

samdark:
@maine-mike #10480 is now merged. Sorry that it took so long.

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/7249a6c99e96cafae522da379fc7c609ce910bc3 with minor adjustments in https://github.com/yiisoft/yii2/commit/38be74446e943f238f13be02ac93f1880451548c

PR #11558, opened 2016-05-12
Move Identity Cookie code into separate functions and cleanup invalid identity cookies

ghost:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #8795 |


llfm:
I've this code in my application: https://ideone.com/oygl4K
It was necessary to me to separate the parse cookie logic, because you can have a console application that uses cookies (e.g. websocket).

framework/web/User.php
```diff
@@ -562,6 +544,49 @@ protected function sendIdentityCookie($identity, $duration)
...
+            }
+        }
+        $this->removeIdentityCookie();
+        return null;
+    }
+     
+    /**
+     * Removes the identity cookie.
+     * This method is used when [[enableAutoLogin]] is true.
+     * @param IdentityInterface $identity
```
slinstj: Here it seems that this `@param` was forgotten since your method has no params.

maine-mike: Thanks!  I had originally intended to pass $identity in case it was needed, but someone told me that was a bad idea.  So I took it out and neglected to update the comment.

framework/web/User.php
```diff
@@ -562,6 +544,48 @@ protected function sendIdentityCookie($identity, $duration)
...
+                }
+            }
+        }
+        $this->removeIdentityCookie();
+        return null;
+    }
+     
+    /**
+     * Removes the identity cookie.
+     * This method is used when [[enableAutoLogin]] is true.
```
SilverFire: `@since 2.0.9`

framework/web/User.php
```diff
@@ -562,6 +544,48 @@ protected function sendIdentityCookie($identity, $duration)
     }
 
     /**
+     * Determines if an identity cookie format is valid and has a valid auth key.
+     * This method is used when [[enableAutoLogin]] is true.
+     * This method attempts to authenticate a user using the information in the identity cookie.
+     * @return array|null Returns an array of 'identity' and 'duration' if valid, otherwise null.
+     * @see loginByCookie()
```
SilverFire: `@since 2.0.9`

framework/web/User.php
```diff
@@ -562,6 +544,48 @@ protected function sendIdentityCookie($identity, $duration)
     }
 
     /**
+     * Determines if an identity cookie format is valid and has a valid auth key.
```
SilverFire: `and has a valid auth key`

Which objective do you mean? Who has an auth key?

framework/web/User.php
```diff
@@ -562,6 +544,48 @@ protected function sendIdentityCookie($identity, $duration)
...
+     * @return array|null Returns an array of 'identity' and 'duration' if valid, otherwise null.
+     * @see loginByCookie()
+     */
+    protected function validateIdentityCookie()
+    {
+        $value = Yii::$app->getRequest()->getCookies()->getValue($this->identityCookie['name']);
+        if ($value === null) {
+            return null;
+        }
+        $data = json_decode($value, true);
```
SilverFire: Might be replaced with `Json::decode()`

framework/web/User.php
```diff
@@ -562,6 +544,48 @@ protected function sendIdentityCookie($identity, $duration)
...
+            /* @var $class IdentityInterface */
+            $class = $this->identityClass;
+            $identity = $class::findIdentity($id);
+            if ($identity !== null) {
+                if (!$identity instanceof IdentityInterface) {
+                    throw new InvalidValueException("$class::findIdentity() must return an object implementing IdentityInterface.");
+                } elseif (!$identity->validateAuthKey($authKey)) {
+                    Yii::warning("Invalid auth key attempted for user '$id': $authKey", __METHOD__);
+                } else {
+                    return [ 'identity' => $identity, 'duration' => $duration ];
```
SilverFire: Spaces after and before square brackets are redundant

framework/web/User.php
```diff
@@ -562,6 +544,48 @@ protected function sendIdentityCookie($identity, $duration)
...
+                }
+            }
+        }
+        $this->removeIdentityCookie();
+        return null;
```
SilverFire: might be simplified to `return;`

framework/web/User.php
```diff
@@ -284,35 +284,17 @@ public function loginByAccessToken($token, $type = null)
...
-        $identity = $class::findIdentity($id);
-        if ($identity === null) {
-            return;
-        } elseif (!$identity instanceof IdentityInterface) {
-            throw new InvalidValueException("$class::findIdentity() must return an object implementing IdentityInterface.");
-        }
-
-        if ($identity->validateAuthKey($authKey)) {
+        $data = $this->validateIdentityCookie();
+        if ($data !== null && isset( $data['identity'], $data['duration'])) {
```
SilverFire: `isset( $data...`

space after parenthesis is redundant

SilverFire:
Generally looks good. Could you add a CHNAGELOG line, please?

framework/web/User.php
samdark: Cookie, I guess.

framework/web/User.php
samdark: There's no need to check `$data` at all. `isset()` for keys is enough.

framework/web/User.php
samdark: `Determines if an identity cookie has a valid format and a valid auth key.`

maine-mike:
OK.  I made those changes.  Thank you!

framework/web/User.php
```diff
@@ -562,6 +545,48 @@ protected function sendIdentityCookie($identity, $duration)
...
+     * @see loginByCookie()
+     */
+    protected function validateIdentityCookie()
+    {
+        $value = Yii::$app->getRequest()->getCookies()->getValue($this->identityCookie['name']);
+        if ($value === null) {
+            return;
+        }
+        $data = json_decode($value, true);
+        if (count($data) === 3 && isset($data[0], $data[1], $data[2])) {
```
samdark: I'm not sure `count($data)` check is needed. Also `count()` won't work with non-arrays which may appear after `json_decode`.

samdark: Additionally, it's better to use Json::decode because of enhanced error handling.

framework/web/User.php
```diff
@@ -562,6 +545,48 @@ protected function sendIdentityCookie($identity, $duration)
...
+            list ($id, $authKey, $duration) = $data;
+            /* @var $class IdentityInterface */
+            $class = $this->identityClass;
+            $identity = $class::findIdentity($id);
+            if ($identity !== null) {
```
samdark: No need to check for `null` since `$identity instanceof IdentityInterface` will return `false` if `$identity` is `null`.

framework/web/User.php
```diff
@@ -562,6 +545,48 @@ protected function sendIdentityCookie($identity, $duration)
     }
 
     /**
+     * Determines if an identity cookie has a valid format and contains a valid auth key. 
+     * This method is used when [[enableAutoLogin]] is true.
+     * This method attempts to authenticate a user using the information in the identity cookie.
+     * @return array|null Returns an array of 'identity' and 'duration' if valid, otherwise null.
```
samdark: Based on what it returns I'd name a method something like `getIdentityAndDurationFromCookie()`. Rather long but probably reflects what's done inside better.

framework/web/User.php
maine-mike: If we can't find an Identity with this $id, then it should drop through and not throw an exception.
An exception should only be thrown if what is returned in not the correct class.

framework/web/User.php
maine-mike: I changed my code to use Json::decode and did some testing.  If the contents of the cookie are malformed in some way and Json::decode is used, then an exception will be thrown.  When json_decode is used, my code recognizes that the contents are not usable, no cookie login is performed, and a standard login screen is presented.  I think the behavior using json_decode is preferable, so I change it back.

framework/web/User.php
samdark: Makes sense. OK.

SilverFire:
Merged https://github.com/yiisoft/yii2/commit/7249a6c99e96cafae522da379fc7c609ce910bc3 with minor adjustments https://github.com/yiisoft/yii2/commit/38be74446e943f238f13be02ac93f1880451548c

SilverFire:
Thank you @maine-mike!

PR #11563, opened 2016-05-13
translate guide to bahasa

mdmunir:
[ASK]
How to access translated guide in yii website?


samdark:
1. It should be added https://github.com/yiisoft-contrib/yiiframework.com/blob/master/config/params.php#L41.
2. Then it will appear at http://yiiframework.domain-na.me/ when it will be re-generated.

It won't show up at current website yet but we're going to replace it with ↑ when it's done.

PR #11577, opened 2016-05-17
trim pathInfo slashes

DrDeath72:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |

Not both removed as described
- The starting and ending slashes are both removed.

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -736,9 +736,7 @@ class Request extends \yii\base\Request
             throw new InvalidConfigException('Unable to determine the path info of the current request.');
         }
 
-        if (substr($pathInfo, 0, 1) === '/') {
-            $pathInfo = substr($pathInfo, 1);
-        }
+        $pathInfo = trim($pathInfo, '/');
 
         return (string) $pathInfo;
     }


DrDeath72:
hm... wtf?
https://github.com/yiisoft/yii2/blob/master/framework/web/Request.php#L669
https://github.com/yiisoft/yii2/blob/master/framework/web/Request.php#L696

SilverFire:
Why do you suggest to make this change?

DrDeath72:
See prev comment with code comments

mikehaertl:
Instead of changing code which could break BC I suggest to only fix the docs for both methods to:

> If there's a starting slash, it gets removed. End slashes will be kept.

DrDeath72:
php pathinfo, basename etc remove ending slash

cebe:
trailing slash is relevant as `/example/` and `/example` are two different urls.

zhou19901104:
我想找一个Yii项目 练习

shuangjie:
试试将webuploader整合进yii2?哈哈

SilverFire:
Agree with @cebe 

@DrDeath72 would you like to update this PR?

samdark:
Closing since 3.0 implementation is different.

PR #11580, opened 2016-05-18
YUI Compressor fix

DrDeath72:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

### DIFF
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -123,7 +123,7 @@
             var $form = $('<form/>', {
                 action: url,
                 method: 'get',
-                class: 'gridview-filter-form',
+                'class': 'gridview-filter-form',
                 style: 'display:none',
                 'data-pjax': ''
             }).appendTo($grid);


framework/assets/yii.gridView.js
```diff
@@ -123,7 +123,7 @@
             var $form = $('<form/>', {
                 action: url,
                 method: 'get',
-                class: 'gridview-filter-form',
+                'class': 'gridview-filter-form',
```
klimov-paul: Why?

DrDeath72: error on compression

klimov-paul: Sounds like YUI Compressor issue

DrDeath72: "class" is keyword, phpstorm highlight it different too

klimov-paul: which keyword?
JavaScript does not have `class` keyword.

klimov-paul: > phpstorm highlight it different too

I can't see any extra highlight at PhpStorm at this line.

rob006: @klimov-paul 
http://www.w3schools.com/js/js_reserved.asp
http://www.ecma-international.org/ecma-262/6.0/#sec-keywords

DrDeath72: https://www.dropbox.com/s/6uxukanl7jy4sn0/%D0%A1%D0%BA%D1%80%D0%B8%D0%BD%D1%88%D0%BE%D1%82%202016-05-18%2014.43.58.png?dl=0

klimov-paul: I see. So this is a recent addition to ECMA.

I suppose all option keys should be quoted then for the consistency, not just `class`. And same fix should be applied to other places too.

rob006: @klimov-paul I think that quoting only `class` is enough, there is no point for quoting all keys.

klimov-paul: Quoting single key, while leaving other as they are looks weird, unpretty.
Besides, do we realy need to wait until more of those keys become reserved?

rob006: Quoting single key, while leaving other as they are looks intentional. Quoting all keys looks like chaotic and unnecessary increase file size. 

> Besides, do we realy need to wait until more of those keys become reserved?

I don't think we should worry about it. `class` was reserved in ECMAScrip 1, which has almost 20 years. Reserved words do not appear surprisingly and still can affect variables names, so quoting everything will not solve anything.

DrDeath72: > Quoting single key, while leaving other as they are looks weird, unpretty.

```
 style: 'display:none',
 'data-pjax': ''
``` 
ahah... strong argument

SilverFire:
Let it be

SilverFire:
@DrDeath72 thank you

PR #11594, opened 2016-05-19
Fix #11591 `ActionFilter` wildcard

### LINKED ISSUE #11591: Wildcard support for `ActionFilter` `only` and `except` properties
It would be nice to have support for wildcards at `yii\base\ActionFilter` `only` and `except` properties.

It make a little sense in the scope of the controller, but if `ActionFilter` is attached to the module or application itself - the feature is exceptionally useful.
In case of `Module` (`Application`) the actions are processed with full id, e.g. 'site/index' or 'module/default/index' and operating them is not practical. However, wildcard usage allows to cover set of actions: 

```php
return [
    'as filter' => [
        'class' => 'app\filters\SomeFilter',
        'only' => [
            'particular/*', // all actions at controller 'particular'
            '*/captcha', // all 'captcha' actions through all controllers
        ],
    ],
    // ...
];
```

klimov-paul:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11591 |


df2:
Merge please! :-)

df2:
@klimov-paul thanks!

df2:
@klimov-paul so in trying this I think I have found a bug. It is this line:

https://github.com/yiisoft/yii2/blob/e49bc29aa80a382a7de22835a76693bc7d25790e/framework/base/ActionFilter.php#L129

I think it should be `$id = $action->getUniqueId();`

Otherwise, say a controller that all other controllers extend from has `behaviours()` as:

```php
        return [
            'guests' => [
                'class' => 'app\filters\GuestsFilter',
            ],
            'suspended' => [
                'class' => 'app\filters\SuspendedFilter',
                'except' => [
                    'closed/suspended/*',
                ],
            ],
        ];
```

The except clause, when this behaviour is activated from `app/closed/SuspendedController::indexAction()`, will run `fnmatch('closed/suspended/*', 'index')` instead of `fnmatch('closed/suspended/*', 'closed/suspended/index')`.

klimov-paul:
This is expected behavior.

df2:
@klimov-paul then the tests are disingenuous... the environment is virtual and the action id is set to something that can never apply if the action is a concrete action of a controller. (`actionTest\index()` ?)

klimov-paul:
Tests are simpified: they emulates the situation, when filter attached to the module not to a controller.
Still, if you have an idea for the test refactoring - you are welcome to open a PR.

PR #11597, opened 2016-05-19
Doc translation es

lucianobaraglia:
I don't know what my two previous commits appears here in the pull request even if they are already merged.
New ones are only 187255e and 8c817c0.
Please tell me if I made something wrong.


samdark:
You've used the same branch. That's why.

lucianobaraglia:
@samdark will it lead to any problem? do you prefer a new branch each pull request?

Issue #10157, opened 2015-11-14
cannot use @web or @webroot in config.


jpodpro:
i want to setup some aliases in the app config such as an uploads directory:
'aliases' => [
        '@uploadsurl' => '@web/uploads'
    ],

however this fails because @web is unavailable in the config stage which sucks. @web and @webroot shouldn't need to wait until Request to be defined. are they not extremely independent values?

samdark:
Config is the first thing included even before the app is started. In order to have these aliases available in config these should be defined explicitly either at the very start of `index.php` or on top of the config itself. I'm not sure it's a good idea though.

cebe:
We can not change this. To set up `@web` and `@webroot` aliases we need a request component of the application. This is done after configuring the application, in the bootstrap phase. If you want to define aliases based on `@web` or `@webroot` you have to implement a method for application bootstrap and register the alias there.

jpodpro:
i do not agree with "cannot change this"

software allows anything to be possible with the proper design. defining aliases from @web are a very common and essential requirement for many web apps so insisting on a design that prevents this is, in my opinion, bad practice. a request component should not be required to define basic app-wide constants.

for now i am forced to hack a solution. but i'm disappointed that this would not be up for discussion as i've seen it reported before.

samdark:
https://github.com/yiisoft/yii2/blob/master/framework/web/Application.php#L60 that's what @cebe is referring to.

@jpodpro how do you propose to solve that?

jpodpro:
well the obvious answer to me is to move the setting of these aliases into the main config:

'aliases' => [
        '@web' => stripos($_SERVER['SERVER_PROTOCOL'],'https') === true ? 'https://' : 'http://' . $_SERVER['SERVER_NAME']
    ],

there might require fancier logic to determine if the site isn't installed into the web root. but this doesn't seem very complicated to me.

jpodpro:
here's another related problem: if i'm doing an ajax query on a relative path ( ie: /track/owner ) then the @web alias is empty. this is terrible as i would again expect @web to be universally defined as an app constant and not dependent on the current request. obviously i need to define my own version of @web (which seems silly). if you guys are stuck on this way of doing things then your documentation should be much more clear somewhere that @web and @webroot are not absolute values that can be depended on.

mikehaertl:
[...]

PR #11620, opened 2016-05-24
Issue #10157 Resolve aliases in getAlias() instead of setAlias()

mikehaertl:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | yes/no |
| Fixed issues | #10157 |

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -138,11 +138,13 @@ class BaseYii
 
         if (isset(static::$aliases[$root])) {
             if (is_string(static::$aliases[$root])) {
-                return $pos === false ? static::$aliases[$root] : static::$aliases[$root] . substr($alias, $pos);
+                $path = $pos === false ? static::$aliases[$root] : static::$aliases[$root] . substr($alias, $pos);
+                return strncmp($path, '@', 1) ? $path : static::getAlias($path);
             } else {
                 foreach (static::$aliases[$root] as $name => $path) {
                     if (strpos($alias . '/', $name . '/') === 0) {
-                        return $path . substr($alias, strlen($name));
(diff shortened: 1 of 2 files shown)


Faryshta:
can you add a unit test for the scenario i described on #10157 ?

mikehaertl:
Added some unit tests. It should cover most cases.

framework/BaseYii.php
```diff
@@ -138,11 +138,13 @@ public static function getAlias($alias, $throwException = true)
 
         if (isset(static::$aliases[$root])) {
             if (is_string(static::$aliases[$root])) {
-                return $pos === false ? static::$aliases[$root] : static::$aliases[$root] . substr($alias, $pos);
+                $path = $pos === false ? static::$aliases[$root] : static::$aliases[$root] . substr($alias, $pos);
+                return strncmp($path, '@', 1) ? $path : static::getAlias($path);
```
samdark: Result of `strncmp` should be compared with `=== 0`.

Sxderp:
Should the Alias be re-fetched every time a call is made to getAlias? I don't think it should be, I think it adds unnecessary complexity. I can't think of a valid use case for wanting this functionality. Actually, I can't really think of a use case for changing an alias while the application is running (that is, after the configs have been loaded).

framework/BaseYii.php
mikehaertl: It's already used without `===0` on other places. That's why I've used the same pattern. If we change this we should change it everywhere.

samdark:
Won't make it into 2.0. There's separate alias component in 3.0 now.

Issue #11528, opened 2016-05-09
[php 7.0.*]IntlException  Constructor failed in vendor/yiisoft/yii2/i18n/MessageFormatter.php


hezll:
### What steps will reproduce the problem?
vendor/yiisoft/yii2/i18n/MessageFormatter.php

about line 101

        $formatter = new \MessageFormatter($language, $pattern); //the problem is here.
        if ($formatter === null) {
            $this->_errorCode = intl_get_error_code();
            $this->_errorMessage = 'Message pattern is invalid: ' . intl_get_error_message();

when MessageFormatter has some problems with construct
this will call a IntlException immediately due to the new exception feature under php7 
### What do you get instead?
I just add up 

    try{
        $formatter = new \MessageFormatter($language, $pattern);
    }cache(\IntlException $e){
        $formatter = null;
    }

This helps me fix it.
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.8
| PHP version      | 7.0.*
| Operating system | centos 6.5

cebe:
in previous php versions this resulted in the constructor to create a NULL formatter. Needs to be fixed.

tom--:
Judging from these patterns https://3v4l.org/94MvA I would say

- This constructor should not throw in php7 but return an empty object
- Yii should test with `empty($formatter)` instead of `$formatter === null`

cebe:
Looks like HHVM also has a different implementation... :-/

PR #11644, opened 2016-05-27
Fix MessageFormatter for PHP7

subdee:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11528 |

### DIFF
--- a/framework/i18n/MessageFormatter.php
+++ b/framework/i18n/MessageFormatter.php
@@ -99,10 +99,14 @@ class MessageFormatter extends Component
         $pattern = $this->replaceNamedArguments($pattern, $params, $newParams);
         $params = $newParams;
 
-        $formatter = new \MessageFormatter($language, $pattern);
-        if ($formatter === null) {
-            $this->_errorCode = intl_get_error_code();
-            $this->_errorMessage = 'Message pattern is invalid: ' . intl_get_error_message();
+        try {
+            $formatter = new \MessageFormatter($language, $pattern);
+            if ($formatter === null) {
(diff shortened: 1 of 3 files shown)


SilverFire:
We need a changelog line and a test case to verify this behavior. Could you do it, please?

framework/i18n/MessageFormatter.php
```diff
@@ -99,10 +99,14 @@ public function format($pattern, $params, $language)
...
-        $formatter = new \MessageFormatter($language, $pattern);
-        if ($formatter === null) {
-            $this->_errorCode = intl_get_error_code();
-            $this->_errorMessage = 'Message pattern is invalid: ' . intl_get_error_message();
+        try {
+            $formatter = new \MessageFormatter($language, $pattern);
+            if ($formatter === null) {
+                throw new \Exception(intl_get_error_message(), intl_get_error_code());
+            }
+        } catch (\Exception $e) {
```
samdark: Is it throwing exactly `\Exception` or more specific one?

subdee: @samdark Hm, it catches \IntlException so maybe that is indeed better to use, I will amend this.
@SilverFire I'm on it.

cebe:
IntlException does not exist in PHP5.4 and the fix also does not seem to work this way on HHVM.

samdark:
OK. Then general exception should be used.

cebe:
@SilverFire @samdark I could fix it, are you fine with assigning it to me?

cebe:
Applied a slightly different fix, thanks!

PR #11646, opened 2016-05-27
Added dynamic alias '@alias' in ActiveQuery

CedricYii:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #7263, #10883 |

The `@alias` placeholder enables to address dynamicaly the alias of the current table name in an active query or in a relation (without knowing it in advance).

Example:

``` php
class Order {
...
  public static function find()
  {
    return parent::find()->orderBy( '@alias.id' );
  }

  //relation with table COMPANY
  public function getCustomer()
  {
    return $this->hasOne( Company::className(), ['id' => 'customerid'] )
                ->andOnCondition( [ '@alias.type' => 'customer' ] );
  }
...
}
```

Then

``` php
$orders = Order::find()->alias('o');
```

Build the query:

``` sql
select * from order o order by o.id
```

And

``` php
$orders = Order::find()->alias('o')->joinWith('customer c');
```

Build the query:

``` sql
select o.* from order o 
left join company c on c.id = o.customerid and c.type = 'customer'
order by o.id
```

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -79,7 +79,6 @@ class ActiveQuery extends Query implements ActiveQueryInterface
      * @event Event an event that is triggered when the query is initialized via [[init()]].
      */
     const EVENT_INIT = 'init';
-
     /**
      * string placeholder to be replaced by the table alias of a query.
      */
@@ -160,8 +159,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             if ($this->getIsAliasDynamicEnabled()) {
                 $this->select = ['@alias.*'];
             } else {
-                $alias = $this->getTableAlias();
-                $this->select = ["$alias.*"];
+                $this->select = [$this->getTableAlias() . '.*'];
             }
         }
 
@@ -897,7 +895,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
      */
     protected function normalizeAliasInQuery()
     {
-        //Skip normalisation if dynamic alias feature is not enabled (to save speed)
+        //Skip normalisation if dynamic alias feature is not enabled
         if (!$this->getIsAliasDynamicEnabled()) {
             return $this;
         }
@@ -911,8 +909,8 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         $this->orderBy = self::replaceAliasPlaceholder($this->orderBy, $alias);
         $this->groupBy = self::replaceAliasPlaceholder($this->groupBy, $alias);
         if (!empty($this->join)) {
-            foreach ($this->join as &$join) {
-                $join = self::replaceAliasPlaceholder($join, $alias);
+            foreach ($this->join as $key => $join) {
+                $this->join[$key] = self::replaceAliasPlaceholder($join, $alias);
             }
         }
 
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -870,7 +870,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
                 $value = self::replaceAliasPlaceholder($value, $aliasName);
                 $conditionNew[$key] = $value;
             }
-            $condition = $conditionNew;
+            return $conditionNew;
         } elseif ($condition instanceof \yii\db\Expression) {
             $condition->expression = self::replaceAliasPlaceholder($condition->expression, $aliasName);
         }
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -81,6 +81,11 @@ class ActiveQuery extends Query implements ActiveQueryInterface
     const EVENT_INIT = 'init';
 
     /**
+     * string placeholder to be replaced by the table alias of a query.
+     */
+    const ALIAS_PLACEHOLDER = '@alias';
+
+    /**
      * @var string the SQL statement to be executed for retrieving AR records.
      * This is set by [[ActiveRecord::findBySql()]].
      */
@@ -152,9 +157,16 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         }
 
         if (empty($this->select) && !empty($this->join)) {
-            list(, $alias) = $this->getTableNameAndAlias();
-            $this->select = ["$alias.*"];
+            if ($this->getIsAliasDynamicEnabled()) {
+                $this->select = ['@alias.*'];
+            } else {
+                $alias = $this->getTableAlias();
+                $this->select = ["$alias.*"];
+            }
         }
+
+        //replace dynamic alias placeholder in the query by the current aliasTableName
+        $this->normalizeAliasInQuery();
 
         if ($this->primaryModel === null) {
             // eager loading
@@ -588,6 +600,15 @@ class ActiveQuery extends Query implements ActiveQueryInterface
     }
 
     /**
+     * Returns the table alias (or the table name if empty) of the current ActiveQuery
+     * @return string table alias or table name of the current activeQuery
+     */
+    private function getTableAlias()
+    {
+        return $this->getTableNameAndAlias()[1];
+    }
+
+    /**
      * Joins a parent query with a child query.
      * The current query object will be modified accordingly.
      * @param ActiveQuery $parent
@@ -613,7 +634,11 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         list($parentTable, $parentAlias) = $parent->getTableNameAndAlias();
         list($childTable, $childAlias) = $child->getTableNameAndAlias();
 
+        //replace dynamic alias placeholder in the child query by the current aliasTableName
+        $child->normalizeAliasInQuery();
+
         if (!empty($child->link)) {
+            //automatically quote table aliasName only for the link part
             if (strpos($parentAlias, '{{') === false) {
                 $parentAlias = '{{' . $parentAlias . '}}';
             }
@@ -824,5 +849,73 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         /* @var $modelClass ActiveRecord */
         $modelClass = $this->modelClass;
         return $modelClass::tableName();
+    }
+
+    /**
+     * Replace occurence of self::ALIAS_PLACEHOLDER by $aliasName into given conditions
+     * @param string|array|Expression $condition the given conditions
+     * @param string $aliasName
+     * @return string|array|Expression $condition with alias replaced
+     */
+    protected static function replaceAliasPlaceholder($condition, $aliasName)
+    {
+        if (empty($condition) or is_int($condition)) {
+            return $condition;
+        } elseif (is_string($condition)) {
+            return str_replace(self::ALIAS_PLACEHOLDER, $aliasName, $condition);
+        } elseif (is_array($condition)) {
+            $conditionNew = [];
+            foreach ($condition as $key => $value) {
+                $key = self::replaceAliasPlaceholder($key, $aliasName);
+                $value = self::replaceAliasPlaceholder($value, $aliasName);
+                $conditionNew[$key] = $value;
+            }
+            $condition = $conditionNew;
+        } elseif ($condition instanceof \yii\db\Expression) {
+            $condition->expression = self::replaceAliasPlaceholder($condition->expression, $aliasName);
+        }
+
+        return $condition;
+    }
+
+    /**
+     * Returns a value indicating whether the dynamic alias feature is enable
+     * @return bool the value indicating whether the dynamic alias feature is enable
+     */
+    public static function getIsAliasDynamicEnabled()
+    {
+        if (isset(\Yii::$app->enableAliasDynamic)) {
+            return \Yii::$app->enableAliasDynamic;
+        }
+
+        return false;
+    }
+
+    /**
+     * Replace the dynamic alias placeholders by the alias table name of the current ActiveQuery
+     * @return $this the current ActiveQuery with relation alias replaced
+     */
+    protected function normalizeAliasInQuery()
+    {
+        //Skip normalisation if dynamic alias feature is not enabled (to save speed)
+        if (!$this->getIsAliasDynamicEnabled()) {
+            return $this;
+        }
+
+        $alias = $this->getTableAlias();
+
+        $this->select = self::replaceAliasPlaceholder($this->select, $alias);
+        $this->on = self::replaceAliasPlaceholder($this->on, $alias);
+        $this->where = self::replaceAliasPlaceholder($this->where, $alias);
+        $this->having = self::replaceAliasPlaceholder($this->having, $alias);
+        $this->orderBy = self::replaceAliasPlaceholder($this->orderBy, $alias);
+        $this->groupBy = self::replaceAliasPlaceholder($this->groupBy, $alias);
+        if (!empty($this->join)) {
+            foreach ($this->join as &$join) {
+                $join = self::replaceAliasPlaceholder($join, $alias);
+            }
+        }
+
+        return $this;
     }
 }
--- a/tests/data/ar/Item.php
+++ b/tests/data/ar/Item.php
@@ -25,4 +25,11 @@ class Item extends ActiveRecord
     {
         return $this->hasOne(Category::className(), ['id' => 'category_id']);
     }
+
+    //relation with dynamic alias
+    public function getCategoryWithAlias()
+    {
+        return $this->hasOne(Category::className(), ['id' => 'category_id'])
+              ->andOncondition(['like', '@alias.name', 'o']);
+    }
 }
--- a/tests/data/ar/Order.php
+++ b/tests/data/ar/Order.php
@@ -8,7 +8,7 @@
 namespace yiiunit\data\ar;
 
 /**
- * Class Order
+ * Class Order.
  *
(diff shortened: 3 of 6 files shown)


lynicidn:
return parent::find()->alias('ali')->orderBy( 'ali.id' );

lynicidn:
where you want configure your aliases ?

framework/db/ActiveQuery.php
```diff
@@ -81,6 +81,11 @@ class ActiveQuery extends Query implements ActiveQueryInterface
     const EVENT_INIT = 'init';
 
     /**
+     * string placeholder to be replaced by the table alias of a query.
+     */
+    const ALIAS_PLACEHOLDER = '@alias';
```
lynicidn: what? once alias for all queries ?
try it:
A::find()->joinWth(['b' => function($q){$q->andWhere(['@alias.id' => 1])}])->andWhere(['@alias.id' => 2])

framework/db/ActiveQuery.php
```diff
@@ -624,16 +635,16 @@ private function joinWithRelation($parent, $child, $joinType)
...
-            $this->andHaving($child->having);
+            $this->andHaving( $child->having );
         }
         if (!empty($child->orderBy)) {
-            $this->addOrderBy($child->orderBy);
+            $this->addOrderBy( $child->orderBy );
         }
         if (!empty($child->groupBy)) {
-            $this->addGroupBy($child->groupBy);
+            $this->addGroupBy( $child->groupBy );
```
lynicidn: read code style concept

framework/db/ActiveQuery.php
```diff
@@ -789,4 +800,69 @@ public function alias($alias)
         }
         return $this;
     }
+
+    /**
+     * Replace occurence of self::ALIAS_PLACEHOLDER by $aliasName into given conditions
+     * @param string|array $condition the given conditions
+     * @param string $aliasName
+     */
+    private function replaceAliasPlaceholder($condition, $aliasName)
```
lynicidn: why is private? you work with private var else where profit of private method ?

framework/db/ActiveQuery.php
```diff
@@ -789,4 +800,69 @@ public function alias($alias)
...
+     */
+    private function normalizeAliasInQuery()
+    {
+        $alias = $this->getAliasTableName();
+
+        $this->on       = self::replaceAliasPlaceholder( $this->on, $alias );
+        $this->where    = self::replaceAliasPlaceholder( $this->where, $alias );
+        $this->having   = self::replaceAliasPlaceholder( $this->having, $alias );
+        $this->orderBy  = self::replaceAliasPlaceholder( $this->orderBy, $alias );
+        $this->groupBy  = self::replaceAliasPlaceholder( $this->groupBy, $alias );
```
lynicidn: select?

tests/data/ar/Item.php
```diff
@@ -20,4 +20,11 @@ public function getCategory()
     {
         return $this->hasOne(Category::className(), ['id' => 'category_id']);
     }
+
+    //relation with dynamic alias
+    public function getCategoryWithAlias()
```
lynicidn: has one relation exists, need write test for exists relations, then all test cover you suggestion

tests/data/ar/Order.php
```diff
@@ -196,4 +218,14 @@ public function attributeLabels()
             'total' => 'Invoice Total',
         ];
     }
+
+
+    public static function findWithAlias()
```
lynicidn: what what what? for what it?

tests/framework/db/ActiveRecordTest.php
```diff
@@ -1250,4 +1250,77 @@ public function testLinkWhenRelationIsIndexed3()
...
+        $orders = $query->all();
+
+        $this->assertEquals(3, $orders[2]->id);
+        $this->assertEquals(3, count($orders));
+    }
+
+    /**
+     * Check if the dynamic alias @alias is replaced in a query a related joinWith
+     */
+    public function testAliasDynamicInJoin()
```
lynicidn: you think it case need test in active record? - this test for query

framework/db/ActiveQuery.php
CedricYii: Thanks for your feedback.
However, I have a hard time to understand what you are trying to point out. 
Could you be more explicit?

tests/framework/db/ActiveRecordTest.php
CedricYii: How would you test it through Query (my commit is on ActiveQuery only)?

tests/data/ar/Item.php
CedricYii: I am testing a relation with a dynamic alias (that doesn't exist), not a "regular" relation.

framework/db/ActiveQuery.php
CedricYii: Ok, I'll change it

framework/db/ActiveQuery.php
CedricYii: Would be nice indeed. I had an unexpected behavior with select during development, I need to check if it's now ok with this last version.

petrabarus:
This is really great! Awesome work @CedricYii!

CedricYii:
@petrabarus Thank you for your comment! You are welcome.
Now that I've used this feature in a project, it sounds like it's a must have as it makes building complex ActiveQueries really easier.

lynicidn:
what is real use case?
(new UserQuery())->select(['@alias.username, @alias.contact_phone'])->from('@alias')->with('profile') 
???

petrabarus:
@lynicidn you can read one of the use case from me here #7263

framework/db/ActiveQuery.php
```diff
@@ -806,7 +806,7 @@ public function alias($alias)
      * @param string|array $condition the given conditions
      * @param string $aliasName
      */
-    private function replaceAliasPlaceholder($condition, $aliasName)
+    public function replaceAliasPlaceholder($condition, $aliasName)
```
petrabarus: why do you need this to be public?

CedricYii: It's not needed but wouldn't hurt as @lynicidn suggested.
For some reasons I don't know, methods (removeDuplicatedModels, buildJoinWith, getQueryTableName, ...) are in private so I sticked to it for normalizeAliasInQuery. However I would have put it in protected.

CedricYii:
@lynicidn It doesn't make sense to use this dynamic aliasing in a single query nor in directly a single activeQuery itself.
An simple example is provided with this PR.
Let's try to extend this example a little:
````php
class Order {
...
  public static function find()
  {
    return parent::find()->orderBy( '@alias.id' );
  }
...
}
````
So you can write,
````php
$orders = Order::find()->alias('o');
````
But you can also write in the same application using the same model
````php
$orders = Order::find()->alias('otherAlias');
````
Or even
````php
$orders = Order::find();
````
All three statements will give the same result (list of orders ordered by id) without any need to worry about what's in the model Order, nor to modify it nor to extend it not to overload it.
Does this make things clearer?

**Comment by lynicidn** [comment]
@CedricYii 
```php
class Order {
...
  public static function find()
  {
    return parent::find()->orderBy( '@alias.id' );
  }
...
}

class User {
...
  public static function find()
  {
    return parent::find()->orderBy( '@alias.id' );
  }
    public function getOrders() {
        $this->hasMany('Order', ['user_id' => 'id']);
    }
...
}
```
what is expected result where ? User::find()->with('orders') ?

CedricYii:
@lynicidn: You will get
// SELECT * FROM user ORDER BY user.id
// SELECT * FROM order WHERE `user_id` IN (...) ORDER BY order.id

To change condition for relation, modify getOrders:
````php
class User {
...
    public function getOrders() {
        return $this
            ->hasMany('Order', ['user_id' => 'id'])
            ->andOnCondition('@alias.customer_id' => 15)
            ->orderBy(['@alias.id' => 'DESC']);
    }
...
}
````
Note that there are no conflict names in these queries, so `@alias` remains useless here.

**Comment by CedricYii** [comment]
@cebe Would it be possible to plan this in 2.0.9 as it provides a complete solution as we talked about?

**Comment by CedricYii** [comment]
Note that another practical usage is when you want to get extra columns for an ActiveRecord model.

For example, let's say you want to get the customer name with the order:
Before you needed to write:
````php
class Order {
...
  public static function find()
  {
    return parent::find()
            ->select([static::tableName().'*', 'c.name'])
            ->joinWith(['customer c'], false);
  }
...
}
````
Which was not really explicit and not portable (you couldn't define a new alias on this model).

Now you can just write:
````php
class Order {
...
  public static function find()
  {
    return parent::find()
            ->select(['@alias.*', 'c.name'])
            ->joinWith(['customer c'], false);
  }
...
}
````
Which looks more intuitive and straightforward, and you can reuse it with dynamic aliasing.

**Comment by CedricYii** [comment]
@cebe I have rebased the commits, could you give it a try to have it included in 2.0.10?

**Comment by CedricYii** [comment]
@cebe Let me know if I can help you to have this one merged?

**Comment by dynasource** [comment]
although I appreciate all the effort that is put into this solution, IMHO, it not yet ready for integration as:
* it bloats ActiveQuery and adding too much complexity to ActiveQuery
* it kills ActiveQuery performance

**Comment by CedricYii** [comment]
@dynasource Most of the time adding a feature means adding code, which one shouldn't be afraid of if the feature is adding value to the framework.
In this case, we are talking about a 9% inflation (72 lines out of 797 lines of code). And it does not change the way you are using ActiveQuery if you don't want to use it.
And if you end up using it one day, you will find that you couldn't go back coding without `@alias` as it greatly simplifies development. 
Or maybe I missed the point you are thinking about in your first sentence?

As for performance issue, have you done a benchmark or tried it and noticed consequent lags? 
Note that I am using str_replace only (not regex like for already merged table alias feature) to keep it fast.
However ActiveQuery is not the fastest way to build queries for sure, but it is the most flexible, maintainable and scalable; and `@alias` is going this way.

**Comment by CedricYii** [comment]
I did a benchmark on a page overloaded with 200 queries using dynamic `@alias` (and ActiveQuery with aliasDynamic) and the same one with hard-coded aliases (and original ActiveQuery), and it has been slowed down by only 3%.
Estimation trying to isolate the impact on ActiveQuery part itself is 15%.
So I would not say it kills performance in regard of the feature offered.

However it remains quite simple to desactivate this by overriding `ActiveQuery::normalizeAliasInQuery` or add an option to do so.

**Comment by dynasource** [comment]
> I did a benchmark on a page overloaded with 200 queries using dynamic @alias (and ActiveQuery with aliasDynamic) and the same one with hard-coded aliases (and original ActiveQuery), and it has been slowed down by only 3%.

thanks for your replies and doing a benchmark. 
Your benchmark unfortunately proves what I was afraid of:
* 3% performance loss on a new feature which will cover ~1% of usecases 

There it must be recoded in such way that if somebody want to use it, he/she can, but for default scenario's it must be left out (unless performance is in no way killed).

The reason why I care about performanceloss, even if its only 3% in your opinion, comes from 2 reasons: local testing + production. Both are extremely important for me to be optimal.

**Comment by antonmarin** [comment]
> will cover ~1% of usecases
@dynasource, why you think it's only about 1%? As i understand it as almoust every ActiveQuery method, that can be used to detail query by relative record.

**Comment by dynasource** [comment]
True. I misformulated that. It can indeed be used in every usecase, however, what I tried to point out, is that you will not always need this, but I does go at cost of 3% at all times.

**Comment by antonmarin** [comment]
I see now. @CedricYii do you have ideas how to implement placeholder without perfomance regression?

**Comment by Renkas** [comment]
Any progress on this? This feature is long overdue in framework :(

**Comment by samdark** [comment]
@Renkas no. Performance will be degraded if we'll merge it as is.

File: framework/db/ActiveQuery.php [hunk 9]
```diff
@@ -794,4 +805,64 @@ public function alias($alias)
...
+        return $condition;
+    }
+
+    /**
+     * Returns the table alias (or the table name if empty) of the current ActiveQuery
+     * @return string table alias or table name of the current activeQuery
+     */
+    public function getAliasTableName()
+    {
+        list (, $alias) = self::getQueryTableName($this);
```
Faryshta: retun self::getQueryTableName($this)[1];

?

framework/db/ActiveQuery.php
```diff
@@ -794,4 +805,64 @@ public function alias($alias)
...
+    public function replaceAliasPlaceholder($condition, $aliasName)
+    {
+        if ((empty($condition)) or (is_int($condition))) {
+            return $condition;
+        } elseif (is_string($condition)) {
+            return str_replace(self::ALIAS_PLACEHOLDER, $aliasName, $condition);
+        } elseif (is_array($condition)) {
+            $conditionNew = [];
+            foreach ($condition as $key => $value) {
+                unset($condition[$key]);
```
Faryshta: redundant since line 828 rewrites `$condition` anyway

Faryshta:
@samdark the method `normalizeAliasInQuery()` only gets called when building the actual query.

the travis build doesn' seem wildly affected either. although i can't see individual tests performance.

Faryshta:
@CedricYii i was thinking to use another name for this alias such as `@main` to be able to use other automatic aliases later based on relation names. example

```php
User::find()->innerJoinWith('tickets')->andWhere([
     '@main.status' => User::STATUS_ACTIVE,
     '@tickets.status' => Ticket::STATUS_OPEN,
]);
```

all the inner joins would cause an automatic alias with the name of the relation being turned into the related table name and would allow nesting.

```php
Store::find()->innerJoinWith('sales.articles')->select([
    'brute_sales' => 'sum(@sales.articles.price)' 
]);
```

for this pr i think you just need to edit:

`const ALIAS_PLACEHOLDER = '@main';`

Faryshta:
as for the performance, how about a method `ActiveQuery::dynamicAlias(boolean $active)` by default it will be unactivated and `normalizeAliasInQuery()` won't be executed. saving the query builder a whooping 3% performance.

rob006:
Why we event care about this 3%? This is essential feature. Right now I don't even consider porting some apps from Yii 1.1 to Yii 2 because of poorly aliases support.

Faryshta:
i am working on a pr which will be based on this but wont use `@alias` instead it will use `@main` for the main table and the relation name prefixed with `@` symbol will be tthe automatic alias for the a related record supporting nesting.

example

```php
class Order {
  public static function find()
  {
    return parent::find()->orderBy( '@main.id' );
  }

  //relation with table COMPANY
  public function getCustomer()
  {
    return $this->hasOne( Company::className(), ['id' => 'customerid'] )
                ->andOnCondition( [ '@customer.type' => 'customer' ] );
  }
}
```

nested example

```php
Store::find()->joinWith('sales.articles')
    ->where(['@main.id' => 1])
    ->sum('@sales.articles.amount')
```

njasm:
any decisions on this one? /cc @CedricYii  @Faryshta @cebe @rob006 @samdark

cebe:
this needs some serious amount of dedicated time to be handled correctly, which I currently do not have. Sorry. Its not going to make it into 2.0.13

samdark:
https://travis-ci.org/yiisoft/yii2/jobs/338904472

CedricYii:
@samdark I know but QueryBuilder in master seem to be broken anyway...  Not related to my commit.

samdark:
It is not: https://travis-ci.org/yiisoft/yii2/builds/338703011

CedricYii:
I see now. I thought 2.0.14 wasn't supposed to break BC, but it did in the QueryBuilder commit (d165863) 4 days ago... I will adapt accordingly.

CedricYii:
@samdark Is it ok for you now?

samdark:
@CedricYii I'm not into this issue so can't say about the whole concept w/o help from @cebe and @SilverFire.

CedricYii:
@samdark Thank you. As the main concern about performance for those that may not use this feature is now addressed, would you mind letting it a chance in the 2.0 milestones (instead of the far far away 2.1.1 you set 9 days ago)?

framework/db/ActiveQuery.php
CedricYii: Thanks in removed it.

framework/db/ActiveQuery.php
CedricYii: Ok I changed it this way, but it's not the way it was done in the master.

samdark:
@cebe said he'll check it soon.

CedricYii:
@Faryshta Concerning your proposal about using relation name as the alias placeholder, I am not sure to see the advantages. As far as I understand, it would not work when the same relation is used twice, it would be harder to maintain (if we change the name of the relation, the placeholder must be changed too) and it would be slower. I feel like it means using an alias of the alias of a relation.
Your nested example could be handled by alias in nested relation #11421.

samdark:
@CedricYii discussed within code team and decided to move it to 2.1.

CedricYii:
@samdark Too bad, thanks for the feedback anyway.

MattiCerio:
Thanks for the work, this would be useful!
At the moment there is a temporary solution to retrieve in the relation definition the alias used ?

samdark:
Won't be merged into 2.x.

Issue #11647, opened 2016-05-27
Wrong pageCount in Pagination


Nex-Otaku:
ActiveDataProvider pagination returns wrong pageCount.

### Reproducing

// Total 8 items.
// We have 5 items on page.
// It makes 2 pages.

$dataProvider = new ActiveDataProvider(['query' => Notes::find()]);

$dataProvider->setPagination(['pageSize' => 5]);

$pageCount = $dataProvider->pagination->pageCount;

### Expected result

$pageCount is 2

### What do you get instead

$pageCount is 0

### Additional info

Console test is attached.
[TestPageCountController.zip](https://github.com/yiisoft/yii2/files/286274/TestPageCountController.zip)

Put controller class in console/controllers and run
 yii test-page-count

Pull request with bugfix is provided.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0
| PHP version      | 5.6
| Operating system | Windows

Alex-Code:
I don't think setting the `pagination` component should fetch the models, It might cause confusion.

```php
$query = Notes::find();
$dataProvider = new ActiveDataProvider(['query' => $query]);
$dataProvider->setPagination(['pageSize' => 5]);
$query->where(['foo' => 'bar']); //Will have no effect as the models have already been fetched.
```

Nex-Otaku:
@Alex-Code it is not.

It is calling prepareTotalCount which queries "COUNT(*)" on table. No models are fetched.

And every time query object is used in ActiveDataProvider, it is cloned. So there should be no problems.

Alex-Code:
Ah yes, It only performs a `COUNT(*)` but I think my issue above is still valid.
```php
$query = Notes::find();
$dataProvider = new ActiveDataProvider(['query' => $query]);
$dataProvider->setPagination(['pageSize' => 5]);
$pageCount = $dataProvider->pagination->pageCount; //May return 5
$query->where(['foo' => 'bar']);
$pageCount = $dataProvider->pagination->pageCount; //Would still return 5 even though the query has changed.
```
I wouldn't expect setting the options of `Pagination` to initiate a query.
Maybe passing the total count to the `Pagination` object itself would be better.

```php
$dataProvider->pagination->getPageCount($dataProvider->totalCount);
```

Nex-Otaku:
I assume that developer should not change query after fetching data. 

Asking "what total page count" is fetching data.

If he needs to change query after fetching data, he must reset models with DataProvider methods, and fetch new data.

[...]

PR #11648, opened 2016-05-27
Fixed pageCount in ActiveDataProvider pagination #11647

Nex-Otaku:
BaseDataProvider should set pagination totalCount, otherwise pagination will return wrong pageCount.

We set it in setPagination. 

| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | not checked |
| Tests pass? | not checked |
| Fixed issues | #11647 |

### DIFF
--- a/framework/data/BaseDataProvider.php
+++ b/framework/data/BaseDataProvider.php
@@ -200,6 +200,9 @@ abstract class BaseDataProvider extends Component implements DataProviderInterfa
         } else {
             throw new InvalidParamException('Only Pagination instance, configuration array or false is allowed.');
         }
+        if ($this->_pagination !== false) {
+            $this->_pagination->totalCount = $this->getTotalCount();
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


framework/data/BaseDataProvider.php
```diff
@@ -200,6 +200,9 @@ public function setPagination($value)
         } else {
             throw new InvalidParamException('Only Pagination instance, configuration array or false is allowed.');
         }
+        if ($this->_pagination !== false) {
```
lynicidn: for clean code use $value and need check implement interface - not `false`

Faryshta:
i don't think thats a valid use case. Example

```php
{
     $query = Model::find();
         'query' => $query,
         'pagination' => ['pageSize' => 5], // with your patch, the query count will be called here.
     ]);

     if (!$this->validate()) {
         $query->where('0=1'); // this WONT affect the pagination count.
         return $dataProvider;
     }

     $query->andFilterWhere([
          // lots of filters here, none will affect the page count since it was already calculated.
     ]);

     return $dataProvider;
}
```

Issue #11656, opened 2016-05-29
Can we have some of yii\base\application properties keep as protected?


magicaner:
I am developing CMS and find my self that I need retrieve application preterites like `name`, `charset`, `sourceLanguage`, `language` from another sources, eg from custom configuration component, which keeps all this information at database. 

I was planing to keep those properties as protected and add getter|setter functions to avoid standard behavior. Unfortunately, base abstract class have them as `public`, and php do not allows to change property visibility in descendant classes. To solve my issue I have to copy core class and replace it by my needs. 

I would be much appreciated if you will add ability for developers to set getters and setters for app level properties like `name`, `charset`, `sourceLanguage`, `language`. Not big issue actually. 

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.8

lynicidn:
add test for it - all methods need cover

magicaner:
@lynicidn tests are added

samdark:
It looks OK but I want to ask for @yiisoft/core-developers opinions first.

cebe:
this is a bit of overkill, why do you need this for all of the mentioned properties?

changing `sourceLanguage` means you need to change the code so there is no sense in putting this into the database, same for charset.

magicaner:
@cebe I am having custom configuration source, and all application options are saved here. And want add getter to application which allows me read those options transparently from my configuration, just like they are application properties, this is actually very flexible. Why not allow developers such feature? I tried to explain carefully at issue above, please take a look.

cebe:
@magicaner you did not give reasoning for the single properties. Imo it does not make any sense to configure sourceLanguage and charset dynamically. it may be useful to language and name though.

magicaner:
@cebe okie, I will try again :)
I suppose they are stored at configuration for some reasonable reasons. And if configuration moved to separate component than it is reasonable to read them directly instead of setting application properties on each initialization? also setting them at application will create duplications in such case.

cebe:
the question is, what is the point of setting sourceLanguage and charset dynamically from database? it does not make any sense to me as these depend on how the code is written.

magicaner:
@cebe if you need real example here is one:
Lets purpose CMS allows to setup multiple websites per one application instance. Each website can have own configuration, can setup configuration from admin panel without touching static config files,  can have own themes and translation files and so on. They can have different sourceLanguage and charset. 

Do you think something wrong with that?

cebe:
[...]

PR #11657, opened 2016-05-29
change visibility of application public properties to protected

magicaner:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11656 |
- following properties set as protected, which allows to override getters and setters at descendant classes:
  name, charset, langugage, sourceLangugage


cebe:
I'm sorry to waste your effort here but this adds a lot of unnecessary code and for most of the properties there is no benefit as explained in https://github.com/yiisoft/yii2/issues/11656#issuecomment-237953087.

PR #11673, opened 2016-06-01
More control over check access for create action

uaoleg:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |

For example we create a review and passing 3 parameters in our request: `productId`, `reviewerId`, `rating`
Currently there is no ability to check that `reviewerId` is equal to ID of authenticated user, or perform any other check access based on received data.

### DIFF
--- a/framework/rest/CreateAction.php
+++ b/framework/rest/CreateAction.php
@@ -37,16 +37,17 @@ class CreateAction extends Action
      */
     public function run()
     {
-        if ($this->checkAccess) {
-            call_user_func($this->checkAccess, $this->id);
-        }
-
         /* @var $model \yii\db\ActiveRecord */
         $model = new $this->modelClass([
             'scenario' => $this->scenario,
         ]);
 
         $model->load(Yii::$app->getRequest()->getBodyParams(), '');
+
+        if ($this->checkAccess) {
+            call_user_func($this->checkAccess, $this->id, $model);
+        }
+
         if ($model->save()) {
             $response = Yii::$app->getResponse();
             $response->setStatusCode(201);


tunecino:
I think this will take the `checkAccess` out of its context. The model doesn't exist yet in this case as it is not yet saved so it make no sense to me to check access to a non existant model. 

I would use model validation rules to control the model attributes before saving them. and for attributes like `reviewerId` I won't let user set its value via any input in the first place and maybe using a `beforeSave` instead:

```php
public function beforeSave($insert)
{
    if ($insert) $this->reviewerId = Yii::$app->user->id;
    return parent::beforeSave($insert);
}
```
Then when the model is successfully saved it will make sense to check access to it in the other actions as already implemented:
```php
call_user_func($this->checkAccess, $this->id, $model);
```

uaoleg:
Not sure if it's a good practice to use WebUser (I mean \yii::$app->user) inside data model, as far as data models are used not only in web application.

tunecino:
is creating that model require user to be authenticated ? 

If answer is YES then `Yii::$app->user` should hold what is handled by the User Identity Class, the one usually defined in the config file and the one that should handle authentication for web or non web applications:
```php
'user' => [
    'identityClass' => 'app\models\User',
],
```
If answer is NO then model's validation rules steal making more sense as it is designed to validate a model before creating it. I don't know about your use case but if you have to accept from a list of allowed `reviewerId` as example you can use the range validator for it:

```php
['reviewerId', 'in', 'range' => [1,2,5,7], 'strict' => true],
// or
['reviewerId', 'in', 'range' => ArrayHelper::getColumn(User::find()->where(...)->all(), 'id'), 'strict' => true],
```
You can even build a custom validator and throw a `yii\web\ForbiddenHttpException`  or any other exception if required.

tunecino:
If the same model should be used by a web user and no web user then I think the proper way is to use different scenarios to your validation rules or building 2 different models or handling validation dynamically (I've seen that somewhere in docs but i don't remember where)

uaoleg:
@tunecino did I get you right that you propose to validate access for update action using `checkAccess` method, but for create action you propose to do the same stuff inside AR `beforeSave` method? Don't you think it's kind a logic duplication.

tunecino:
well a validator that throws a ForbiddenHttpException wasn't a good idea. you are right on that @uaoleg.

I just see it this way and I may be wrong:
**adding a new record to an app ->** should be a model job : Model::rules() , Model::validate() should handle it before the record getting to exist in the first place.
**check access to an existing record ->** checkAccess or equivalent

cebe:
> Currently there is no ability to check that reviewerId is equal to ID of authenticated user, or perform any other check access based on received data.

this step belongs to validation of the model, i.e. model rules() and not into access check. access check is for checking if the logged in user is allowed to access the action in general. if he provides wrong data that is part of input validation.

PR #11686, opened 2016-06-03
fix BaseArrayHelper::isIn() in strict mode

taobig:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | yes? |
| Tests pass? | yes |
| Fixed issues | none |


samdark:
What's the problem exactly?

cebe:
@samdark needle should be compared with value, just as it is the case for non-strict comparsion. seems the tests do not cover this case.

cebe:
Merged, thank you!

cebe:
also added a test case:  	b211e6b

PR #11697, opened 2016-06-05
add functions for filtering

nicdnepr:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | no |

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -706,6 +706,24 @@ class Query extends Component implements QueryInterface
         return $this;
     }
 
+    public function andFilterHaving(array $condition)
+    {
+        $condition = $this->filterCondition($condition);
+        if ($condition !== []) {
+            $this->andHaving($condition);
+        }
+        return $this;
+    }
+
+    public function orFilterHaving(array $condition)
+    {
+        $condition = $this->filterCondition($condition);
+        if ($condition !== []) {
+            $this->orHaving($condition);
+        }
+        return $this;
+    }
+
     /**
      * Adds an additional HAVING condition to the existing one.
      * The new condition and the existing one will be joined using the 'OR' operator.


framework/db/Query.php
```diff
@@ -706,6 +706,24 @@ public function andHaving($condition, $params = [])
         return $this;
     }
 
+    public function andFilterHaving(array $condition)
```
samdark: Need `@since` and phpdoc.

framework/db/Query.php
```diff
@@ -706,6 +706,24 @@ public function andHaving($condition, $params = [])
...
+    {
+        $condition = $this->filterCondition($condition);
+        if ($condition !== []) {
+            $this->andHaving($condition);
+        }
+        return $this;
+    }
+
+    public function orFilterHaving(array $condition)
```
samdark: Need `@since` and phpdoc.

cebe:
@nicdnepr could you explain a use case for these methods?

nicdnepr:
It's the same as function andFilterWhere, but only for having. I had a task to add GridView filtering for having results, and was surprised that there is no analog of function andFilterWhere. So I decided to add this code

samdark (changes requested):
It looks good to me except `@since` and CHANGELOG.

SilverFire (changes requested):
Tests are required

samdark:
Not sure if it should be in QueryInterface...

samdark:
> Not sure if it should be in QueryInterface...

No, it should not. It's in correct class.

samdark:
@nicdnepr do you want to work on it further?

nicdnepr:
@samdark Hi, I don't have much time at this moment

samdark:
Closed in favor of https://github.com/yiisoft/yii2/pull/13210

PR #11705, opened 2016-06-06
typo in PHP_SELF matching?

pine3ree:


samdark:
Nope. It's correct.

pine3ree:
@samdark 
i think we should at least add `isset($_SERVER['SCRIPT_NAME'])` to that `elsif` clause.
SCRIPT_NAME is not always set by web-servers even if is a cgi standard.

Nevertheless $_SERVER['PHP_SELF'] should contain $_SERVER['SCRIPT_NAME'] when they are both set:
```
$scriptName             index.php
$_SERVER['PHP_SELF']    /sub/dir/index.php/path/to/resource
$_SERVER['SCRIPT_NAME'] /sub/dir/index.php
```

position of $scriptName in both server parameters would be the same, but in the condition i was referring to, we are not checking for the existence of SCRIPT_NAME

or am i missing some other case? could you please provide an example?

kind regards

samdark:
Theoretically you're right about `SCRIPT_NAME` check but there were no reports about its unavailability for more than 5 years so I guess it's safe to rely on the code. If you know how to break it, it would be cool to see.

pine3ree:
I believe we should ask @qiangxue to really know if it really was a typo. :-)
I checked zend-http, zf2 http, and zf-1 up to v 1.0. The code used in determining the baseUrl stops at the 3rd conditions used by yii (1 & 2), then it tries to match the basename of the entry script path SCRIPT_FILENAME up backwards. I believe the remaining 2 conditions were added specifically to yii.

Anyway, i was writing a little middleware to set the base-url in a zend-expressive app and i came back to yii to see how the problem was solved and i found the reference to zf1. So it was mere curiosity about that condition. I won't bother you anymore about this, I promise. :-)

thanks and kind regards

PR #11715, opened 2016-06-07
Fix scope for multiple forms file inputs

uaoleg:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

If there are multiple forms on a page (of the same model and with the same `formName`) and these forms contain file input, than JS validation will affect only file input of the first form and will ignore all the other.

### DIFF
--- a/framework/assets/yii.activeForm.js
+++ b/framework/assets/yii.activeForm.js
@@ -302,6 +302,7 @@
 
             // client-side validation
             $.each(data.attributes, function () {
+                this.$form = $form;
                 if (!$(this.input).is(":disabled")) {
                     this.cancelled = false;
                     // perform validation only if the form is being submitted or if an attribute is pending validation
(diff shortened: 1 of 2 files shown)


Alex-Code:
This could be a wider issue.
If you have two forms with the same model the input fields will have the same `id` which is against the `HTML` spec.

SilverFire:
@Alex-Code you're right. This problem is listed in our [plan for 2.1](https://github.com/yiisoft/yii2/wiki/Plan-for-next-major-versions#models). Currently you have to take care of input IDs by yourself.

@uaoleg thank you for the PR

SilverFire:
Changelog: https://github.com/yiisoft/yii2/commit/c9f500b3ed198812fc75d6278640481bdac0b751

PR #11729, opened 2016-06-08
CheckboxColumn - Use CSS class to find the elements

thiagotalma:
Adds the ability to use CSS class as a filter for the function to check/uncheck all the rows.

By default, nothing is changed and the current operation is maintained.

| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |


samdark:
What's the reason for this change?

thiagotalma:
@samdark 
The field name, which is used in javascript, is set in the `init` method.

However, in some cases we need to define the name of the field after the `init` and before the `renderDataCell`.
Use case: https://github.com/kartik-v/yii2-grid/pull/498

This makes the checkAll not work, since it is based on the name that has changed.

The way I propose in PR, this problem can be solved without any loss in the standard operation of the widget.

thiagotalma:
@samdark how about this?
It can be merged? It would help a lot ...

samdark:
Add a line for the changelog and we'll merge it.

thiagotalma:
@samdark done!

samdark:
Thanks!

PR #11736, opened 2016-06-10
204 response for empty result

arturf:


tunecino:
that will be a BC.  **200** status code with an empty array `[]` means to my frontend code that:
- no data was found.
- no change has been made to database.
- results are caming straight from database and not from cache (otherwise I would expect **304**).

I personally prefer receiving a **204** (*[rfc7231](https://tools.ietf.org/html/rfc7231#section-6.3.5)*) only when something change into my database which means when performing POST, PUT or DELETE verbs so I can tell my client app that cached data is stale.

samdark:
Won't change that.

cebe:
204 means that there is no HTTP body, in this case there is one that contains an empty array. fine detail but important difference.

Issue #10422, opened 2015-12-17
ColumnSchemaBuilder - Not null column


nevermnd:
With the new migration syntax, in MSSQL all columns are created as NOT NULL whether they have been defined so or not. I think the implementation of the class ColumnSchemaBuilder for MSSQL should add a explicit NULL for nullable columns.

Ex:
```php
protected function buildNotNullString()
{
    return $this->isNotNull ? ' NOT NULL' : ' NULL ';
}
```

samdark:
How's that? I've just tried MSSQL 2015:

```sql
create table A (
	id int
);
```

Creates null-able column as expected.

nevermnd:
In MSSQL 2008

```sql
create table B (
   name varchar(10)
);
```
name is created as NOT NULL by default

nevermnd:
Apparently there are two scenarios in which the columns are created as not null by default:
* When SET ANSI_NULL_DFLT_ON is Set On
* When ANSI_NULL_DEFAULT setting of the database is Set to OFF and BOTH SET and SET ANSI_NULL_DFLT_ON ANSI_NULL_DFLT_ON are Set to OFF

http://sqlstudies.com/2014/03/31/default-nullability-of-a-new-column/

samdark:
I see. Yes, I agree that it should be adjusted. Need to check how all supported DBs behave with `NULL`.

samdark:
MySQL is OK with it.

samdark:
PostgreSQL is OK.

nevermnd:
Oracle, SQLite and Cubrid are ok with it too.

samdark:
Merged https://github.com/yiisoft/yii2/commit/2498a9b3bb8b1b535b0dfc451218217a63610eb7. Thank you!

samdark:
Revered proposed fix because of tests break. It doesn't behave well with primary keys.

nevermnd:
Since we can't change `notNull` behavior, i think adding a method `null` or `nullable` to 
explicitly set the column nullability is the best solution.

Issue #9574, opened 2015-09-02
Migration bug Timestamp


ventayol:
There's a related bug that someone has already closed (https://github.com/yiisoft/yii2/issues/8080) but I think it's a yii bug.

I've the following migration:

        $this->createTable('sector', [
            'id' => Schema::TYPE_PK,
            'name_i18' => Schema::TYPE_INTEGER. ' NOT NULL',
            'created_at' => Schema::TYPE_TIMESTAMP,
            'updated_at' => Schema::TYPE_TIMESTAMP
        ], $tableOptions);

And on the mysql server it's created like this:

![screen shot 2015-09-02 at 15 51 51](https://cloud.githubusercontent.com/assets/1505998/9632976/929b9c2c-518a-11e5-9a11-09e18be541ad.png)
which is completely wrong as the `created_at` field is updated every time the row is updated (I already have the TimeStampBehaivour for this).

I've done a test running the following SQL directly on MySQL:

    CREATE TABLE test (created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, updated_at TIMESTAMP NOT NULL DEFAULT '0000-00-00');

and here you have the result:

<img width="894" alt="screen shot 2015-09-02 at 15 55 45" src="https://cloud.githubusercontent.com/assets/1505998/9633072/1e4105f0-518b-11e5-8796-1e30c911a8b2.png">
So it really doesn't look like a MySQL error.

yaKostin:
I have this bug too.

alchimik:
https://github.com/yiisoft/yii2/issues/8080#issuecomment-92953426

directly on MySQL:
``` sql 
CREATE TABLE test (
    created_at TIMESTAMP, 
    updated_at TIMESTAMP
);

show create table test;
```

```
CREATE TABLE `test` (
  `created_at` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  `updated_at` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00'
)
```

ventayol:
@alchimik 
Any idea on how to avoid this problem?

alchimik:
set default value explicitly:

``` sql
CREATE TABLE test (
    created_at TIMESTAMP NULL DEFAULT NULL, 
    updated_at TIMESTAMP NULL DEFAULT NULL
);

show create table test;
```
```
CREATE TABLE `test` (
[...]

PR #11745, opened 2016-06-10
New proposal fix for ColumnSchemaBuilder not null column

sugiruu:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #10422, #9574 |


cebe:
it's a bit confusing to have two properties for the same method.
It could be possible to have semantics like: 

- `isNotNull = true` - explicit `NOT NULL` is added
- `isNotNull = null` - implicit null
- `isNotNull = false` - explicit `NULL`

nevermnd:
Makes sense. I've changed it.

thiagotalma:
Solve this too: #9574

samdark:
Thanks!

Issue #11777, opened 2016-06-18
RBAC docs: Be more clear about how to add access check to controller


nikhiltri:
I've found the doc that [explains RBAC](http://www.yiiframework.com/doc-2.0/guide-security-authorization.html) use very helpful. It's straightforward and easy to follow for a concept that's a little cloudy for me. One gap that I had a hard time with though is the [Access Check](http://www.yiiframework.com/doc-2.0/guide-security-authorization.html#access-check) section. It provides an example condition for checking access before executing a block of code:

```php
if (\Yii::$app->user->can('createPost')) {
    // create post
}
```

But if you have a series of CRUD permissions as described in the doc, is it best practice to add this condition to each of the action* methods? Or is it better to use the Access Control Filter to centralize the permissions in a controller? Something like this:

```php
public function behaviors()
    {
        return [
            'access' => [
              'class' => AccessControl::className(),
              'rules' => [
                [
                  'allow' => true,
                  'actions' => ['index'],
                  'roles' => ['manageOrder'],
                ],
                [
                  'allow' => true,
                  'actions' => ['view'],
                  'roles' => ['viewOrder'],
                ],
                [
                  'allow' => true,
                  'actions' => ['create'],
                  'roles' => ['createOrder'],
                ],
                [
                  'allow' => true,
                  'actions' => ['update'],
                  'roles' => ['updateOrder'],
                ],
                [
                  'allow' => true,
                  'actions' => ['delete'],
                  'roles' => ['deleteOrder'],
                ],
              ],
            ],
        ];
    }
```

Access Control Filter is described as an _alternative_ to RBAC, so if this is a good practice it's unclear that it is, and how exactly to use it.

samdark:
It depends on if you want granular permissions or not. If you want to separate adding and deleting then you need to check it in each action. If all the CRUD is managed together then it's a good idea to use a single permission `manageX` and check it in `beforeAction()`.

samdark:
https://github.com/yiisoft/yii2/pull/11791

PR #11791, opened 2016-06-20
Add instructions for adding authorization to your controllers #11777

nikhiltri:
Do these changes to the doc make it more clear how authorization should be implemented in controllers?


samdark:
Thanks!

Issue #10876, opened 2016-02-17
Ambiguity when delete table with migration tool yii v2.0.7


s-o-f:
I have table "products_from_store" in database.
For deleting this table with Yii Migration Tool in Yii v.2.0.7 i use console command:

`yii migrate/create drop_products_from_store`

but Yii generates migration with empty methods up and down.

There's ambiguity with commands for dropping table or column. 
Yii tries to delete column "products" from table "store".

zelenin:
```migrate/create drop_products_from_store``` creates an empty migration with name ```drop_products_from_store```. that's all.

**Comment by s-o-f** [comment]
In yii version 2.0.7 it must add dropTable method in up();
http://www.yiiframework.com/doc-2.0/guide-db-migrations.html#drop-table

**Comment by zelenin** [comment]
Oh, you're right.

**Comment by VirtualRJ** [comment]
I guess it is because the regex that creates the migration checks first for a combination of `drop....from` and your table name includes 'from'. Which actually tries to create a drop column which it probably cannot create.

https://github.com/yiisoft/yii2/blob/master/framework/console/controllers/BaseMigrateController.php#L519

The code responsible for it:

```php
elseif (preg_match('/^drop_(.+)_from_(.+)$/', $name, $matches)) {
                $content = $this->renderFile(Yii::getAlias($this->generatorTemplateFiles['drop_column']), [
                    'className' => $className,
                    'table' => mb_strtolower($matches[2], Yii::$app->charset),
                    'fields' => $this->fields
                ]);
```

samdark:
@VirtualRJ is correct.

samdark:
@s-o-f any proposals on how to fix it?

s-o-f:
Yeah, i think the name of new migration must be: "remove_column_from_table" (replace "drop" to "remove").

samdark:
It doesn't matter for the case. `products_from_store` would still match it.

VirtualRJ:
@samdark but it wouldn't match the 'remove' and then move to next if where it would match the drop regex

samdark:
Right. I don't like such change though because SQL used is `drop` in both cases and Migration methods are `dropColumn()` and `dropTable()`. That would be inconsistent.

VirtualRJ:
If it is allowed to change the command then you could do:

- `drop_..._from_table_...` (but it has the same issue as current one, but makes it even less likely to happen) or
[...]

PR #11808, opened 2016-06-23
Handle `_table` and `_column` suffixes on migration creation.

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | - |

This supports more explicit migration names, like:
- `create_user_table`,
- `add_name_column_to_user_table`,
- `create_junction_table_for_post_and_tag_tables`.

### DIFF
--- a/tests/framework/console/controllers/MigrateControllerTestTrait.php
+++ b/tests/framework/console/controllers/MigrateControllerTestTrait.php
@@ -194,93 +194,121 @@ CODE;
 
     public function testGenerateCreateMigration()
     {
-        $migrationName = 'create_test';
+        $migrationNames = [
+            'create_test',
+            'create_test_table',
+        ];
+        foreach ($migrationNames as $migrationName) {
+            $this->assertCommandCreatedFile('create_test', $migrationName);
 
-        $this->assertCommandCreatedFile('create_test', $migrationName);
+            $this->assertCommandCreatedFile('create_fields', $migrationName, [
+                'fields' => 'title:string(10):notNull:unique:defaultValue("test"),
+                    body:text:notNull,
+                    price:money(11,2):notNull'
+            ]);
 
-        $this->assertCommandCreatedFile('create_fields', $migrationName, [
-            'fields' => 'title:string(10):notNull:unique:defaultValue("test"),
-                body:text:notNull,
-                price:money(11,2):notNull'
-        ]);
+            $this->assertCommandCreatedFile('create_title_pk', $migrationName, [
+                'fields' => 'title:primaryKey,body:text:notNull,price:money(11,2)',
+            ]);
 
-        $this->assertCommandCreatedFile('create_title_pk', $migrationName, [
-            'fields' => 'title:primaryKey,body:text:notNull,price:money(11,2)',
-        ]);
+            $this->assertCommandCreatedFile('create_id_pk', $migrationName, [
+                'fields' => 'id:primaryKey,
+                    address:string,
+                    address2:string,
+                    email:string',
+            ]);
 
-        $this->assertCommandCreatedFile('create_id_pk', $migrationName, [
-            'fields' => 'id:primaryKey,
-                address:string,
-                address2:string,
-                email:string',
-        ]);
+            $this->assertCommandCreatedFile('create_foreign_key', $migrationName, [
+                'fields' => 'user_id:integer:foreignKey,
+                    product_id:foreignKey:integer:unsigned:notNull,
+                    order_id:integer:foreignKey(user_order):notNull,
+                    created_at:dateTime:notNull',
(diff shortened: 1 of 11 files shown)


pana1990:
this pr is a bugfix, it fix https://github.com/yiisoft/yii2/issues/10876

It neccesary update docs and it changes to drop tables and drop columns syntax.

i suggest the following syntax :

- drop_xxx_from_yyy to dropcolumn_xxx_from_yyy or dropColumn_xxx_yyy

- drop_xxx

 to droptable_xxx or dropTable_xxx

What do you about it?

i can help  you with docs and check code

rob006:
This PR still support old short syntax, so if #10876 is a bug, this PR will not fix it. My intention was only to remove annoying suffixes from generated migrations, for example `./yii migrate/create create_user_table` generate:

```php
<?php

use yii\db\Migration;

/**
 * Handles the creation for table `user_table`.
 */
class m160623_141619_create_user_table extends Migration
{
    /**
     * @inheritdoc
     */
    public function up()
    {
        $this->createTable('user_table', [
            'id' => $this->primaryKey(),
        ]);
    }

    /**
     * @inheritdoc
     */
    public function down()
    {
        $this->dropTable('user_table');
    }
}
```

Your solution does not really solve this issue. You can still have column like `name_from_api` and migration like `dropcolumn_name_from_api_from_user`. IMO `_column` and `_table` suffixes should be mandatory, but @yiisoft/core-developers should decide.

pana1990:
ups yeah, you are right

tests/framework/console/controllers/MigrateControllerTestTrait.php
```diff
@@ -194,93 +194,121 @@ public function testGenerateDefaultMigration()
...
     {
-        $migrationName = 'create_junction_post_and_tag';
-        $this->assertCommandCreatedFile('junction_test', $migrationName);
+        $migrationNames = [
+            'create_junction_post_and_tag',
+            'create_junction_table_for_post_and_tag_tables',
+            'create_junction_table_for_post_and_tag_table',
+            'create_junction_post_and_tag_tables',
+        ];
+        foreach($migrationNames as $migrationName) {
```
samdark: `foreach (`

tests/framework/console/controllers/MigrateControllerTestTrait.php
```diff
@@ -194,93 +194,121 @@ public function testGenerateDefaultMigration()
...
-            'fields' => 'title:string(10):notNull,body:text:notNull,
-                price:money(11,2):notNull,
-                created_at:dateTime'
-        ]);
+        $migrationNames = [
+            'drop_columns_from_test',
+            'drop_columns_from_test_table',
+            'drop_columns_column_from_test_table',
+        ];
+        foreach($migrationNames as $migrationName) {
```
samdark: `foreach (`

tests/framework/console/controllers/MigrateControllerTestTrait.php
```diff
@@ -194,93 +194,121 @@ public function testGenerateDefaultMigration()
...
-        $this->assertCommandCreatedFile('drop_test', $migrationName);
-
-        $this->assertCommandCreatedFile('drop_fields', $migrationName, [
-            'fields' => 'body:text:notNull,price:money(11,2)'
-        ]);
+        $migrationNames = [
+            'drop_test',
+            'drop_test_table',
+        ];
+        foreach($migrationNames as $migrationName) {
```
samdark: `foreach (`

tests/framework/console/controllers/MigrateControllerTestTrait.php
```diff
@@ -194,93 +194,121 @@ public function testGenerateDefaultMigration()
...
-            'fields' => 'user_id:integer:foreignKey,
-                product_id:foreignKey:integer:unsigned:notNull,
-                order_id:integer:foreignKey(user_order):notNull,
-                created_at:dateTime:notNull',
-        ]);
+        $migrationNames = [
+            'create_test',
+            'create_test_table',
+        ];
+        foreach($migrationNames as $migrationName) {
```
samdark: `foreach (`

samdark:
Need a line for changelog. Looks OK to me otherwise.

rob006:
@samdark What about #10876? We could close it if we remove support for short syntax without `_table` and `_column` suffixes. Or just close #10876 as it is edge case which is not supported and user should fix migration himself (it's just a template and does not need to be 100% bulletproof).

samdark:
I think we should remove short syntax.

samdark:
Merged. Thank you!

PR #11811, opened 2016-06-23
Fix migration junction template file name

dmytroraschepkin:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? |  |
| Fixed issues |  |

Fix name of the junction template file name in migration console controller.

### DIFF
--- a/framework/console/controllers/MigrateController.php
+++ b/framework/console/controllers/MigrateController.php
@@ -80,7 +80,7 @@ class MigrateController extends BaseMigrateController
         'drop_table' => '@yii/views/dropTableMigration.php',
         'add_column' => '@yii/views/addColumnMigration.php',
         'drop_column' => '@yii/views/dropColumnMigration.php',
-        'create_junction' => '@yii/views/createTableMigration.php',
+        'create_junction' => '@yii/views/createJunctionMigration.php',
     ];
     /**
      * @var boolean indicates whether the table names generated should consider


samdark:
https://travis-ci.org/yiisoft/yii2/jobs/139799227

dmitriy-luch:
@samdark I'm new to all these things. Sorry for stupid PR.
Current one can be closed, since it is not needed. There is no error in the core. Junction migrations runs well.

But in case you have a minute:
I tried to run tests locally using this instruction https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md#3-prepare-the-testing-environment
But the `php build/build dev/app basic` command did not work. The error is 

> Unknown command "dev/app"

I can not understand how that should be executed since I can not see "dev" controller in build/controller folder.
Is there any more detailed manual to run everything locally?

p.s. Do I really need to run `composer update` instead of `composer install` ?

As for the problem itself, I did not found where exactly "testGenerateCreateJunctionMigration" is stored to understand the issue better. Is there any documentation regarding core tests that I can check?

Also, I thought there was an error since I found `yiisoft/yii2/views/createJunctionMigration.php` file but as I can see it is not used anywhere (is that file is for future changes or was used somewhere earlier?).
Thanks in advance.

samdark:
I'm executing it as `./build/build dev/app basic`.

> p.s. Do I really need to run composer update instead of composer install ?

Not really.

> As for the problem itself, I did not found where exactly "testGenerateCreateJunctionMigration" is stored to understand the issue better.

https://github.com/yiisoft/yii2/blob/master/tests/framework/console/controllers/MigrateControllerTestTrait.php#L280

> Also, I thought there was an error since I found yiisoft/yii2/views/createJunctionMigration.php file but as I can see it is not used anywhere (is that file is for future changes or was used somewhere earlier?).

It was used but now is marked as deprecated (see comments on top of the file).

PR #11813, opened 2016-06-23
Implicit run ColumnSchemaBuilder::null() when default value is set  to null.

### LINKED ISSUE #9574: Migration bug Timestamp

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #9574 |

fix #9574
see https://github.com/yiisoft/yii2/issues/9574#issuecomment-199385521 and https://github.com/yiisoft/yii2/issues/9574#issuecomment-199411265

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -6,6 +6,7 @@ Yii Framework 2 Change Log
 -----------------------
 
 - Enh #10422: Added `null` method on `yii\db\ColumnSchemaBuilder` to explicitly set column nullability (nevermnd)
+- Enh #9574: Implicit run `ColumnSchemaBuilder::null()` when default value is set to `null`. (rob006)
 - Enh #8795: Refactored `yii\web\User::loginByCookie()` in order to make it easier to override (maine-mike, silverfire)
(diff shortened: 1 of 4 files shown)


samdark:
Thank you!

PR #11818, opened 2016-06-24
Adds indexes on message tables which partly fixes #11725

### LINKED ISSUE #11725: Selecting data for translation
* using the 'yii\i18n\DbMessageSource' for translations
* calling Yii:t() with sourceLanguage different  from language

Some clean, performance-considering query. Something like this:

``` SQL
SELECT
	"t1"."message" AS "message",
	"t2"."translation" AS "translation"
FROM
	sp_source_message t1
JOIN sp_message t2 ON ("t1"."id" = "t2"."id")
WHERE
	"t1"."category" = 'app'
AND (
	"t2"."language" = 'cs-CZ'
	OR "t2"."language" = 'cs'
)
```

### What do you get instead?
``` SQL
(
	SELECT
		"t1"."message" AS "message",
		"t2"."translation" AS "translation"
	FROM
		"sp_source_message" "t1",
		"sp_message" "t2"
	WHERE
		("t1"."id" = "t2"."id")
	AND ("t1"."category" = 'app')
	AND ("t2"."language" = 'cs-CZ')
)
UNION ALL
	(
		SELECT
			"t1"."message" AS "message",
			"t2"."translation" AS "translation"
		FROM
			"sp_source_message" "t1",
			"sp_message" "t2"
		WHERE
			(
				("t1"."id" = "t2"."id")
				AND ("t1"."category" = 'app')
				AND ("t2"."language" = 'cs')
			)
		AND (
			"t2"."id" NOT IN (
				SELECT
					"id"
				FROM
					"sp_message"
				WHERE
					"language" = 'cs-CZ'
			)
		)
	)
```

### Additional info
Do we really need to seek for the ‘cs-CZ’ form of language? In the message table is only ‘cs’. Probably because my translation setting for console.
But even though – maybe I’m missing something, but the original query with subqueries seems to be unnecessary and highly ineffective. There is no need to explain what it does. It’s quite clear.

The other thing are indexes. There aren’t any. For small web, it’s not a problem, but I think this is a great example, where inserting and updating are so rare, that the ratio for read/insert(update) is extremely high.
So I wouldn’t hesitate to put indexes on language and category columns. I know how indexes works and I know that in many cases, they will be useless, because the number of same content in these columns will be same. But it’s still far better situation to have useless index on table, where reading is much higher, then situation we have large message table with loads of categories and no indexes. At least the category column is important. But again – from the principal perspective… :)
I know there is a caching option, but it just covers the problem.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.8
| PHP version      | 5.6.15
| Operating system | Win10

OndrejVasicek:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | enhancement |
| Breaks BC? | no |
| Tests pass? | - |
| Fixed issues | #11725 |

I added message indexes into the migration script and tested it.

I also added indexes into the SQL file for PostgreSQL. I used “clustered” (see https://www.postgresql.org/docs/9.1/static/sql-cluster.html), which isn’t possible via migration method, but it’s welcome here. Also tested.

I’m afraid I don’t have enough knowledge to complete other languages. Is it generated or made manually? Do you think you could complete it?

DBs support:
- [x] PostgreSQL
- [x] MySQL
- [x] MSSQL
- [x] SQLite
- [x] Oracle

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 
 2.0.9 under development
 -----------------------
-
+- Enh #11725: Added indexes on message tables (OndrejVasicek)
 - Enh #10422: Added `null` method on `yii\db\ColumnSchemaBuilder` to explicitly set column nullability (nevermnd)
 - Enh #8795: Refactored `yii\web\User::loginByCookie()` in order to make it easier to override (maine-mike, silverfire)
 - Enh #9948: `yii\rbac\PhpManager` now invalidates script file cache performed by 'OPCache' or 'APC' on file saving (klimov-paul)
--- a/framework/i18n/migrations/m150207_210500_i18n_init.php
+++ b/framework/i18n/migrations/m150207_210500_i18n_init.php
@@ -39,6 +39,8 @@ class m150207_210500_i18n_init extends Migration
 
         $this->addPrimaryKey('pk_message_id_language', '{{%message}}', ['id', 'language']);
         $this->addForeignKey('fk_message_source_message', '{{%message}}', 'id', '{{%source_message}}', 'id', 'CASCADE', 'RESTRICT');
+        $this->createIndex('idx_source_message_category', '{{%source_message}}', 'category');
+        $this->createIndex('idx_message_language', '{{%message}}', 'language');
     }
 
     public function down()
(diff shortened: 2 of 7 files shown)


samdark:
Merged. Thank you!

Issue #11807, opened 2016-06-23
Add Hint Index in queries with join relations


fdezmc:
Hi, 
the use of Expression don't work to hint index with activeRecord in joins with relations.

```
// works!!!
$query = (new Query)->from([new Expression('{{%user}} USE INDEX (primary)')]);

// joinWith relation fails
$query = (new Query)->from([new Expression('{{%user}} USE INDEX (primary)')])
->joinWith('userProfile');
```

Note: userProfile is a relation between user and profile ON user.id = profile.user_id

ActiveQuery class method joinWithRelation takes as $parentAlias an expression resulting in:

SELECT \* FROM `user` USE INDEX (primary) LEFT JOIN `profile` ON `user` USE INDEX (primary).`id` = `profile`.`user_id`

instead of 

SELECT \* FROM `user` USE INDEX (primary) LEFT JOIN `profile` ON `user`.`id` = `profile`.`user_id`

Related issue: [Add Hint Index in queries](https://github.com/yiisoft/yii2/issues/10869)

pstddc:
try:
```php
// joinWith relation fails
$query = (new Query)->from(['userAlias' => new Expression('{{%user}} USE INDEX (primary)')])
->joinWith('userProfile');
```

pstddc:
allow set from as 
- array [alias => expression|tablename]
- string 'tablename'

this incorrect ['tablename']

fdezmc:
Hi @pstddc, 

```
// joinWith relation fails
$query = (new Query)->from(['userAlias' => new Expression('{{%user}} USE INDEX (primary)')])
->joinWith('userProfile');
```

expected:
SELECT `userAlias`.* FROM `user` `userAlias` USE INDEX (primary) LEFT JOIN `profile` ON `userAlias`.`id` = `profile`.`user_id`

get error:
SELECT `userAlias`.* FROM `user` USE INDEX (primary) `userAlias` LEFT JOIN `profile` ON `userAlias`.`id` = `profile`.`user_id`

fdezmc:
Long time ago I open a issue #10869 in order to Add Hint Index in queries. I did a pull request to solve this. 
@cebe point out a way to do this with existing methods using Expression: 

```
// your first example:
$query = (new Query)->from([new Expression('{{%user}} USE INDEX (primary)')]);
// your second example
$query = (new Query)
[...]

PR #11838, opened 2016-06-27
Added hint index support in queries with join relations #11807

fdezmc:
This pull add feature for using expressions in activeRecord from clause with joinWith relations, allowing hint index support in queries with join relations. See #11807


fdezmc:
Pull request removed because of I think this require a discussion and a better approach to face with hint index. While I deploy a better approach I will discuss it at #11807

PR #11845, opened 2016-06-28
Model->getFirstErrors() - Ensure that the first error is a string

thiagotalma:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

Use case:

``` PHP
$model->addErrors(['relation' => $relation->getErrors()]);
```


klimov-paul:
How does error messsage become an array?

klimov-paul:
This is incorrect usage:
https://github.com/yiisoft/yii2/blob/master/framework/base/Model.php#L615

> The array values should be error messages. If an attribute has multiple errors, these errors must be given in terms of an array.

PR #11850, opened 2016-06-29
Possbible pjax form event selector to be changed

ghost:
I have added a new variable that makes it possible to change the selector on the .on() function that used to call the submit for a form with pJax.

### DIFF
--- a/framework/widgets/Pjax.php
+++ b/framework/widgets/Pjax.php
@@ -68,6 +68,11 @@ class Pjax extends Widget
      */
     public $formSelector;
     /**
+     * @var string The jQuery event selector that will be used to trigger the handler code on the form
+     * This defaults to submit.
+     */
+    public $eventSelector = 'submit';
+    /**
      * @var boolean whether to enable push state.
      */
     public $enablePushState = true;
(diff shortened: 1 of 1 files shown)


samdark:
What's use case for it?

Bvanleeuwen:
Yesterday I was developing a search box that is in a pjax container, and we would like the event selector to have a different selector then submit, since we have no submit button visible. We would like it to be keyup or change (we are still discussing that internal).

samdark:
OK. Makes sense.

samdark:
Merged: https://github.com/yiisoft/yii2/commit/1dc688e35f4b918a44e2159bb4fe6e41017aea28. Note that the name of the property was changed.

nevermnd:
This commit broke pjax form submit. The following javascript is being registered:
```js
jQuery(document).on('"submit"', "#form-historico", function(event) {
    jQuery.pjax.submit(event, '#w1', {
        "push": true,
        "replace": false,
        "timeout": 2000,
        "scrollTo": false
    });
});
```

Bvanleeuwen:
Strange, it did work for me when I created it, I see what is wrong, and I will fix it tomorrow.

PR #11856, opened 2016-07-01
Fix reset populated records after fetch other relations with via()

affka:
Scenario to reproduce bug:

Class with two relations. Second relation used first relation as junction.

``` php
class Act extends ActiveRecord {

    /**
     * @return \yii\db\ActiveQuery
     */
    public function getItems()
    {
        return $this->hasMany(ActItem::className(), ['actId' => 'id']);
    }

    /**
     * @return \yii\db\ActiveQuery
     */
    public function getWeightings() {
        return $this->hasMany(ActItemWeighting::className(), ['id' => 'weightingId'])
            ->via('items');

   ...
}

$model = Act::findOne(...);
$firstItem = $model->items[0]; // items - is relation
$model->weightings; // fetch other relation, which used via('items')
var_dump($firstItem === $model->items[0]); // false

```

| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | yes |
| Tests pass? | yes |
| Fixed issues | without issue |

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -174,13 +174,17 @@ class ActiveQuery extends Query implements ActiveQueryInterface
                 // via relation
                 /* @var $viaQuery ActiveQuery */
                 list($viaName, $viaQuery) = $this->via;
-                if ($viaQuery->multiple) {
-                    $viaModels = $viaQuery->all();
-                    $this->primaryModel->populateRelation($viaName, $viaModels);
+                if ($this->primaryModel->isRelationPopulated($viaName)) {
+                    $viaModels = $this->primaryModel->$viaName;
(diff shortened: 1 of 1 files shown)


klimov-paul:
Fix is unacceptable, since there is no garantee `ActItemWeighting` declares it srelation to the `ActItem` using same relation name `item` as the `Act` model.

klimov-paul:
You should use inverse relations in your case:
https://github.com/yiisoft/yii2/blob/master/docs/guide/db-active-record.md#inverse-relations-

affka:
If i will use inverseOf for relation "weightings", then relation items all the same will be reset.
Why relation fetch affect to other relations?

PR #11864, opened 2016-07-01
Fix incorrect formatting in documentation

nkovacs:
The asterisks and underscores were interpreted as markdown formatting, which wasn't intended.

Also fix invalid link to FG_COLOR.

This is how it looks after this patch:

screenshot from 2016-07-01 18-08-44

screenshot from 2016-07-01 18-08-50


samdark:
Thanks!

PR #11874, opened 2016-07-03
Very often in the code, we have to create a method that returns a new…

Razzwan:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | not found |

Very often in the code, we have to create a method that returns a new object, if not found in the database. I propose to add this method in the base class.


Razzwan:
When I want to save more dependent models and I do not know whether there are already those links. Suppose it is a draft of a document. Looks like that:
~~~php
    public function saveWithRelations($runValidation = true, $attributeNames = null)
    {
        if ($this->save($runValidation, $attributeNames) || !$this->getIsNewRecord()) {
            $isValid = true;
            foreach ([1,2,3,4,5] as $order) {
                  $relatedModel = SomeModelClassName::findOrNew(['owner_id' => $this->id, 'order' => $order]);
                  if (!$relatedModel->save) {
                       $this->addError('relationName', $relatedModel->getFirstErrors());
                       $isValid = false;
                  }
            }
        }

        return !$this->getIsNewRecord();
    }
~~~

It saved all valid relations and return true if main model is saved.

pstddc:
@Razzwan is support transaction? or where profit ?

Razzwan:
> is support transaction? 

No! I want to save all valid relations in my draft.

Razzwan:
@samdark 
Or do you mean that you want to add the most simple example in the phpdoc?

Razzwan:
Have you never wrote code like this?:
~~~php
$model = ModelClass::findOne($condition);
if ($model === null) {
    $model = new ModelClass($someData);
}
~~~
In my opinion, it is very frequent usecase. 

Instead, I would like to write:
~~~php
$model = ModelClass::findOrNew($condition);
~~~

But in this usecase, the use is not seen much profit. All good is revealed, if we need to look for a lot of different, but similar patterns for similar conditions. It is this situation, I tried to recreate in the previous example.

rob006:
> ```php
$model = ModelClass::findOne($condition);
if ($model === null) {
    $model = new ModelClass($someData);
}
```

This PR does not work in this way, current implementation creates an empty model, which is useless in most cases.

**Comment by samdark** [comment]
> Have you never wrote code like this?

I don't remember. It should be extremely rare... That's why I've asked for real non-hypothetical case.

**Comment by klimov-paul** [comment]
The proposed method does not make much sense: it reduces 2 lines of code for some rare use case.

> $model = ModelClass::findOrNew($condition);

This is risk prone, since condition for the `ActiveRecord::findOne()` is not always an array of attributes.

> As example: When I want to save more dependent models and I do not know whether there are already those links.

This is a bad practice. In your example you run may SQL queries at `findOne`, while it is better to find all existing models at once and then check if some of them missing.

> In my opinion, it is very frequent usecase.

Disagreed - I have never use such construction.

**Comment by samdark** [comment]
I don't think we'll accept this method in the core framework. @Razzwan if you have good use cases and ides to address issues mentioned feel free to comment, I'll re-open PR.

**Comment by Razzwan** [comment]
Very simple and often usecase. Login form. User set email and password. If exist - login, if not exist - new user.
~~~php
$user = User::findOrNew([
    'email' => $model->email,
    'password' => $model->password,
]);
Yii::$app->user->login($user);
~~~

**Comment by klimov-paul** [comment]
> Very simple and often usecase. Login form. User set email and password. If exist - login, if not exist - new user.

So always save user record after login? Or you create single-use login account?

**Comment by samdark** [comment]
That would create new user each time I'm making a typo...

**Comment by Razzwan** [comment]
@klimov-paul 
> So always save user record after login? Or you create single-use login account?

I mean Save rocord after login. It's quite fashionable simplification.

@samdark 
> That would create new user each time I'm making a typo...

Password can be replaced by sending an email to reset your password. It is important that the user be registered as quickly as possible. Until then, he will leave the site.

In this case, a slight simplification, but it is. And I must improve the mechanism for creating a new object. It is necessary to transmit the $condition, but I consider that it may be entirely in the form of key-value

**Comment by samdark** [comment]
I mean typos in email which happen often. Let's say I already have an account `sam@rmcreative.ru` and trying to login but instead typing `sam@rmcerative.ru`. Instead of an error which I could fix right away I'm getting new account created which is both bad for project's user base, accidental email receivers and is very irritating for the person trying to log in.

**Comment by DrDeath72** [comment]
```
$model = ModelClass::findOne($condition) ?: new ModelClass($someData);
```
no? :D

samdark:
Yes.

Razzwan:
Even a small convenience can be a decisive factor. This method nuzhyn because he is a little more convenient. And it is a typical problem is solved. Yes, you can a million other ways to solve the same problem. You can, for example, to shift the responsibility for the creation of the object on the database.

I know that would use this method, as well as many of those who participates in the discussion also be used, if it was. That is why it is necessary.

DrDeath72:
extend AR and use it, i use oneOrDie  :D

Razzwan:
@DrDeath72 
> extend AR and use it, i use oneOrDie :D

Good name ;) 
I'm doing. But I want to share my best in my opinion the decisions of the others.

samdark:
@Razzwan it's very good that you're doing that and we appreciate it. Still, if it's not covering 80% of cases (in this case it's not) we're trying not to include it into framework.

PR #11878, opened 2016-07-03
Fix i18n gettext fallback language message loading

stevekr:
Fallback language messages didn't get assigned to `$messages` variable in `yii\i18n\GettextMessageSource::loadMessages()`.


samdark:
Merged. Thanks!

Issue #9079, opened 2015-07-10
Add Html::tel();


skworden:
With all of the other types of HTML helpers Yii2 has I think it's only fitting to have a tel: hyperlink type as well. 

It is almost an exact copy of the mailto function.

``` /**
     * Generates a tel hyperlink.
     * @param string $text link body. It will NOT be HTML-encoded. Therefore you can pass in HTML code
     * such as an image tag. If this is coming from end users, you should consider [[encode()]]
     * it to prevent XSS attacks.
     * @param string $tel telephone number. If this is null, the first parameter (telephone body) will be treated
     * as the telephone number and used.
     * @param array $options the tag options in terms of name-value pairs. These will be rendered as
     * the attributes of the resulting tag. The values will be HTML-encoded using [[encode()]].
     * If a value is null, the corresponding attribute will not be rendered.
     * See [[renderTagAttributes()]] for details on how attributes are being rendered.
     * @return string the generated mailto link
     */
    public static function tel($text, $tel = null, $options = []) {
        $options['href'] = 'tel:' . ($tel === null ? $text : $tel);
        return static::tag('a', $text, $options);
    }
```

lynicidn:
use Html::tel(012312313, 'tel') ?

samdark:
https://github.com/yiisoft/yii2/issues/8075

SilverFire:
Rejected. See [this comment](https://github.com/yiisoft/yii2/pull/11882#issuecomment-260145718) for details. Thank you

Issue #8075, opened 2015-04-14
Phone number formatter 


dbsparkle:
Currently the **Data Formatter** has `asEmail` or `asUrl` hyperlink specific formats... 
So it would be nice if it would also have a support for other hyperlink types as  `tel:` (and maybe even for `callto:` as well).

(Sorry, if I am duplicating  an issue - I tried to search if someone else submitted this proposal, but couldn't find anything relevant to it.)

samdark:
What do you expect to pass for it as input?

dbsparkle:
Phone number as a string, so it would render in a similar manner as `asUrl` does:

`<a href="tel:[PHONE-NUMBER]">[PHONE-NUMBER]</a>`

Errata: `callto` has been discarded by [rfc3966](http://tools.ietf.org/html/rfc3966), _because URI schemes locate a resource and do not specify an action to be taken_ .

Also, there is an internationally accepted [E.164 ](http://en.wikipedia.org/wiki/E.164) recommendation by [ITU-T](http://www.itu.int/en/ITU-T/publications/Pages/recs.aspx), but I guess it would be "kinda too much" for a **Data Formatter**. 

What do you think?

samdark:
That's not really a formatter than. Let's keep it as is. Creating a link with callto is easy with anonymous function + raw format.

dbsparkle:
Then what about adding it to HTML helper? http://www.yiiframework.com/doc-2.0/guide-helper-html.html#hyperlinks

Considering that we all are shifting to a mobile web domination, and because the hierarchical part `tel:` is as standard as `mailto:` - it was accepted 10 years ago in [rfc3986](http://tools.ietf.org/html/rfc3986#section-1.1.2) and registered in [IANA](http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml).

Another reason is to preserve consistency and to provide users with a convenient method
for generating links like <?= Html::tel('Call Now', 'some-phone-number') ?>

samdark:
<?php Html::a('call now', 'tel:some-phone-number') ?> isn't any worse.

dbsparkle:
Haha, nice. Try it and then check what you'll get :)

samdark:
I've meant `<?=`, of course, not `<?php`.

dbsparkle:
I got it, I mean you'll have to do extra work before you are able to dial that number.
[...]

PR #11882, opened 2016-07-04
Add HTML::tel() helper

CyberPunkCodes:
Pull request for adding `HTML::tel()` helper function. Feel free to make any recommendations/fixes and I will update this pull.

@samdark  

| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #9079 , #8075 |

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -405,6 +405,27 @@ class BaseHtml
     }
 
     /**
+     * Generates a tel hyperlink.
+     * @param string $text link body. It will NOT be HTML-encoded. Therefore you can pass in HTML code
+     * such as an image tag. If this is coming from end users, you should consider [[encode()]]
+     * it to prevent XSS attacks.
+     * @param string $phone the phone number. If this is null, the first parameter (link body) will
+     * be treated as the phone number and used.
+     * @param array $options the tag options in terms of name-value pairs. These will be rendered as
+     * the attributes of the resulting tag. The values will be HTML-encoded using [[encode()]].
+     * If a value is null, the corresponding attribute will not be rendered.
+     * See [[renderTagAttributes()]] for details on how attributes are being rendered.
+     * @return string the generated tel link
+     */
+    public static function tel($text, $phone = null, $options = [])
+    {
+        $phone = ($phone === null) ? $text : $phone;
+        $phone = preg_replace('~[^+0-9]+~', '', $phone);
+        $options['href'] = 'tel:' . $phone;
+        return static::tag('a', $text, $options);
+    }
+
+    /**
      * Generates a mailto hyperlink.
      * @param string $text link body. It will NOT be HTML-encoded. Therefore you can pass in HTML code
      * such as an image tag. If this is coming from end users, you should consider [[encode()]]
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -410,7 +410,8 @@ class BaseHtml
      * such as an image tag. If this is coming from end users, you should consider [[encode()]]
      * it to prevent XSS attacks.
      * @param string $phone the phone number. If this is null, the first parameter (link body) will
-     * be treated as the phone number and used.
+     * be treated as the phone number and used. This will be passed through 'mb_convert_kana' for half-width
+     * characters.
      * @param array $options the tag options in terms of name-value pairs. These will be rendered as
      * the attributes of the resulting tag. The values will be HTML-encoded using [[encode()]].
      * If a value is null, the corresponding attribute will not be rendered.
@@ -420,6 +421,7 @@ class BaseHtml
     public static function tel($text, $phone = null, $options = [])
     {
         $phone = ($phone === null) ? $text : $phone;
+        $phone = mb_convert_kana($phone, 'a', 'UTF-8');
         $phone = preg_replace('~[^+0-9]+~', '', $phone);
         $options['href'] = 'tel:' . $phone;
         return static::tag('a', $text, $options);


pstddc:
https://github.com/yiisoft/yii2/issues/1022

pstddc:
https://github.com/yiisoft/yii2/issues/169

WadeShuler:
Are you referencing them for letting others now we have re-opened the issue? Or are you trying to agree with them and these should not be added?

No disrespect, but I believe that it should, and many other members do too. Even kartrik-v and others who have contributed to the community, think so.

If we have one for mail to, we should have one for tel!

samdark:
How about https://github.com/yiisoft/yii2/issues/8075#issuecomment-92920262?

WadeShuler:
It takes care of all his cases, except the half-width char. It removes all spaces and extra chars (dashes, dots). It only accepts a `+` sign at the beginning and 0-9 digits. So by passing any phone number through it, it should generate a useable `tel` link. If I recall, it should work with skype, chrome, safari, firefox, on your mac, on your phone.. 

I don't know much about the half-width chars. Should we accommodate for it in the framework? I am open to suggestions and I will update the code.

samdark:
Yes, half-width characters should be handled somehow as well...

samdark:
That's why it's not there yet...

WadeShuler:
I just tested this, with my normal phone and it works for me.

samdark:
Good. Now we need some unit tests and a line for changelog.

SilverFire:
@WadeShuler would you like to continue work in this PR?

WadeShuler:
Yes I just never got back to making a unit test

SilverFire:
Then could you like to add a unit test to this PR to get it merged?

SilverFire:
I've thought better about this case and decided that all this char converting and normalization should be done *before* saving data, rather than on output. 

Kicking out 

```php
$phone = mb_convert_kana($phone, 'a', 'UTF-8');
$phone = preg_replace('~[^+0-9]+~', '', $phone);
```

these lines make method useless.

SilverFire:
Thank you for your effort, but IMO we should not accept this PR to the `HtmlHelper`

rob006:
@SilverFire 

There is lots of nuances on this matter (see https://github.com/yiisoft/yii2/issues/8075#issuecomment-92920262) such helper will be much more useful than `mailto()` (which is in the core).

SilverFire:
Thank you, I've seen it. Case 2 looks the most correct (am I right?) and the most normalized simultaneously. My point is that as we should just normalize phone to digits only, there will no work left for `Html::tel()` method.

rob006:
> Case 2 looks the most correct (am I right?)

You need ask? If so, this means that this is not obvious how such link should look like and such helper could save lots of hours on searching "why my tel links does not work correct on browser X on system Y". I can't say the same about `mailto()` although it is in core.

WadeShuler:
It should remove everything except for numbers. I kept the + sign in front, because it is for international numbers.

SilverFire:
> You need ask? If so, this means that this is not obvious how such link should look like

For me is obvious that phone must be normalized to digits only. Otherwise you will have 1001 different forms of phone format in your DB and it will blow up your mind in one moment.

> why my tel links does not work correct on browser X on system Y

You will find many interesting things in this case. For example [this](http://stackoverflow.com/questions/27164715/tel-href-click-to-call-link-not-working-in-android), [this](https://bugs.chromium.org/p/chromium/issues/detail?id=364522) and most probably [this](http://stackoverflow.com/questions/226131/how-to-disable-phone-number-linking-in-mobile-safari). However, there is even more advanced and SEO-friendly way, like [this](http://microformats.org/wiki/hcard).

I don't want to encourage developers to store mobile phones without normalization so IMO this helper should not be included in core.

rob006:
> For me is obvious that phone must be normalized to digits only.

And it was obvious for you, that you should convert full-width digits to half-width version?

> I don't want to encourage developers to store mobile phones without normalization

Why? This is developer decision how he want store this data. In some cases keeping formating of user input is a feature without any drawback.

Anyway, if you expect that phone numbers should be normalized before saving to database, there should be validator for this.

samdark:
Yes. Anyway, the pull request can't be merged without tests. @WadeShuler if you'll add tests, feel free to ping us to reopen it. Thanks!

WadeShuler:
This has nothing to do with normalization. This is to generate an HTML link to "tel". This is exactly like "mailto". You can write/store emails in incorrect ways. You can enter "MYEMAIL@MAIL.COM", but should convert and store it as lowercase in the db.

When storing phone numbers, you should normalize them so all your data in the db is the same.

However, this feature has nothing to do with any of that. If we have a mailto function, we should have a tel one. Your always going to have people who don't know what the hell they are doing. You can't hold back on things just because of the small percent of people who can't read the docs or follow standards.

SilverFire:
> You can't hold back on things just because of the small percent of people who can't read the docs or follow standards

:+1: 

> This has nothing to do with normalization. This is to generate an HTML link to "tel".

```php
        $phone = mb_convert_kana($phone, 'a', 'UTF-8'); // normalization
        $phone = preg_replace('~[^+0-9]+~', '', $phone); // normalization

        // does not makes much sense without normalization
        // because you can simply call just call Html::a('tel:' . $phone, $phone)
        $options['href'] = 'tel:' . $phone; 
        return static::tag('a', $text, $options);
```

So all this method is about normalization (which should be done before saving IMO) and adding `tel:` (which could be added manually with no drawbacks).

WadeShuler:
That is not why we want this function.

Your exact same argument can be used to justify removing "mailto" from Yii2. People can write "mailto:" in the link tag just the same.

SilverFire:
And I'd remove it, but it will break BC

rob006:
@SilverFire You probably could remove half of methods from `Html` helper - most of them are only shortcuts for 1-2 lines of code...

samdark:
I disagree about removing these ones and even about not including `tel:` but I agree that it's tricky because of normalization and that it's not a good idea to accept pull request without tests.

SilverFire:
@WadeShuler, @rob006 thank you for discussing this PR and sharing your opinion.

I've said all I wanted to say on this matter. Let the majority of @yiisoft/core-developers decide about necessity of `Html::tel()` helper with built-in normalization.

PR #11883, opened 2016-07-04
2.1

nighon:
I found the Validator is coupled with Model class, it is not easy to use Validators for other classes. For example, I want to validate request parameters in Controller which is irrelevant with any Models, is there any recommended way to do this?

And the rules() for Models could not be edit, adding a method addRules() will be fine.


samdark:
Please reopen as an issue.

PR #7625, opened 2015-03-08
test for eager loading 

fernandezekiel:
when there are 3 tables between the relation using via() to link it

Issue #5341, opened 2014-10-03
HasMany via two relations


angelcoding:
I have the following tables and fields (simplified a bit here) and associated AR models

```
tbl_account
id, plan_id, name

tbl_user
id, name

tbl_plan
id, name

tbl_account_user
id, account_id, user_id, type
```

The FK's are obvious.

Basically ...
An account has one plan.
User and account is many to many.
A plan can belong to many accounts and therefore be related to many users.

I want to get all users that are related to a plan so I have setup the following relations in the Plan model ...

```
public function getAccounts()
{
   return $this->hasMany(Account::className(), ['plan_id' => 'id']);
}
public function getAccountUsers()
{
    return $this->hasMany(AccountUser::className(), ['account_id' => 'id'])->via('accounts');
}
public function getUsers()
{
    return $this->hasMany(User::className(), ['id' => 'user_id'])->via('accountUsers');
}
```

If I call $planModel->accountUsers I get the expected models but calling $planModel->users returns no results.

I have done some digging and have found the problem is that the user models are correctly fetched from the db but they are indexed by the account_id rather than the user_id so are not found when populating the relation.

Is this expected behavior?
Should I be doing something differently in this scenario?

I know this is probably a rare case and I'm most likely doing something stupid but I couldn't find any docs or examples anywhere for setting up these types of multi-level relations so it would be very useful to know where I'm going wrong.

cebe:
> @qiangxue if what I have just written in the test did not pass I'd consider this a major issue with AR relations that prevents release, as the test passes I am fine with postponing it.

this test passes because all tables share the same name of the primary key `id` and accidentally the tested pks match the expected result :-/

I have changed the primary keys in the test data and now that test fails. See also #7625 for another failing test.

cebe:
too much work to fix this in 2.0.9, set for 2.0.10

whuhacker:
After 2+ years I just run into the same situation in my project. so when will this problem be fixed?

cebe:
As the label says: "complexity:hard" ;) so it is not likely to be fixed very soon. This needs a rewrite of the bucket matching algorithm in eager loading implementation of AR which is quite complex already and might get even more complex when fixing this.

SamMousa:
If this is complex and no one is working on it yet, should the milestone be removed?

cebe:
I am planning to work on this soon.

Erkebulan69:
I've met with this problem a while ago, and I hope that you'll fix it asap

[...]

PR #11898, opened 2016-07-05
Deeply nesting relation using `via` does not work

cebe:
This PR adds a failing test from #7625 by @fernandezekiel and adjusted test data that fails even one of the existing relation tests.

 these test failures cover the issue described in #5341

### DIFF
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -56,7 +56,7 @@ trait ActiveRelationTrait
      * is the "orders", and the inverse of the "orders" relation is the "customer".
      * If this property is set, the primary record(s) will be referenced through the specified relation.
      * For example, `$customer->orders[0]->customer` and `$customer` will be the same object,
-     * and accessing the customer of an order will not trigger new DB query.
+     * and accessing the customer of an order will not trigger a new DB query.
      * This property is only used in relational context.
      * @see inverseOf()
      */
@@ -139,8 +139,8 @@ trait ActiveRelationTrait
     /**
      * Finds the related records for the specified primary record.
      * This method is invoked when a relation of an ActiveRecord is being accessed in a lazy fashion.
-     * @param string $name the relation name
-     * @param ActiveRecordInterface|BaseActiveRecord $model the primary model
+     * @param string $name the relation name.
+     * @param ActiveRecordInterface|BaseActiveRecord $model the primary model.
      * @return mixed the related record(s)
      * @throws InvalidParamException if the relation is invalid
      */
@@ -441,6 +441,7 @@ trait ActiveRelationTrait
     }
 
     /**
+     * Adjust this query's condition to match related models primary keys only, i.e. to limit query to related records only.
      * @param array $models
      */
     private function filterByModels($models)
@@ -456,6 +457,7 @@ trait ActiveRelationTrait
             foreach ($models as $model) {
                 if (($value = $model[$attribute]) !== null) {
                     if (is_array($value)) {
+                        // relation via multi value column
                         $values = array_merge($values, $value);
                     } else {
                         $values[] = $value;
--- a/tests/data/ar/Category.php
+++ b/tests/data/ar/Category.php
@@ -28,7 +28,7 @@ class Category extends ActiveRecord
     public function getLimitedItems()
     {
         return $this->hasMany(Item::className(), ['category_id' => 'id'])
-            ->onCondition(['item.id' => [1, 2, 3]]);
+            ->onCondition(['item.id' => [21, 22, 23]]);
     }
 
     public function getOrderItems()
--- a/tests/data/ar/Customer.php
+++ b/tests/data/ar/Customer.php
@@ -44,6 +44,21 @@ class Customer extends ActiveRecord
         return $this->hasMany(Order::className(), ['customer_id' => 'id'])->orderBy('id');
     }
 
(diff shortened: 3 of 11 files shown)


cebe:
The problem lies in ActiveRelationTrait bucket building among these lines: https://github.com/yiisoft/yii2/blob/1c200626287588b53b51c3abec4ba653c0962f31/framework/db/ActiveRelationTrait.php#L240-L251

GlyphDataServices:
What's the status of deeply nesting relations with via? Is there an AR solution for this in Yii2?  I'm getting an error when trying to chain two relations. It's looking for the property in the primary model instead of the next relation in the chain.  I was able to chain as many as I needed in Yii1.  I hope we didn't lose this functionality because it was extremely useful for my relational model describing biological samples, in which I frequently need to traverse 4 or more tables.

If this is not possible, is there some documentaton on writing sql joins within relations?

cebe:
It is currently not possible. I am planning to try fixing this for 2.0.13

ztdan4ik:
Hi there!
Thanks for your great work!
Need this fix
Hope soon we will see it :)

GlyphDataServices:
@ztdan4ik this works. Just use via instead of  via_table for chaining relations.

ztdan4ik:
@GlyphDataServices 
I've problem exactly with via() and joinWith()

My problem is same like in tests https://github.com/yiisoft/yii2/commit/7ec045bfa4c7690ca9942ed94b1b4f2f78876a88#diff-dca0f800c4dd790fb5d5914365547549R1194

See the screenshots 
1. Model relations http://i.de-v.net/1528213816.png
2. ModelSearch http://i.de-v.net/1528213884.png
3. Result http://i.de-v.net/1528213917.png

In [this](http://i.de-v.net/1528213981.png) case work fine  But this way looks not good.

GlyphDataServices:
@ztdan4ik two possible issues I've had to work around-
1. try creating in Orders model public $digitId; as a dummy param, even if it's not an actual model param.
2. I can't be sure without seeing your db schema, but your ProductItem relation looks wrong.  Does it really have an ID column?  Looks like a pivot table that should have a product_id and item_id category?

ztdan4ik:
@GlyphDataServices 
Thanks for your reply
See please my [schema](http://i.de-v.net/1528217638.png)

Dummy param not solving the problem. First it's asking for $digit_id, then it's asking for $item_id, then $product_id... And in result I see _Trying to get property of non-object_ in my view because actually relations is not chaining :(

GlyphDataServices:
@ztdan4ik your schema is a bit confusing. Why are you using product_digit_items if they are all hasOne relations?  This sounds like a m2m pivot table but all your relations are hasOne.  I would expect to see digit_id and item_id columns in this table.
It's hard for me to make sense of this schema.  
But regardless, I can confirm that I've been able to chain my relations using both hasMany and hasOne.  

Also note that I set an alias for all my relations, even if it's the same as the function name.

```

	/**
	 * @return \yii\db\ActiveQuery
	 */
	public function getLibrariesIsolates()
	{
		return $this->hasMany(LibrariesIsolates::className(), ['library_id' => 'id'])->alias('librariesIsolates');
	}

	/**
	 * @return \yii\db\ActiveQuery
	 */
	public function getIsolates()
	{
		return $this->hasMany(Isolates::className(), ['id' => 'isolate_id'])->via('librariesIsolates')->alias('isolates');
	}

	/**
	 * @return \yii\db\ActiveQuery
	 */
	public function getSamplesIsolates()
	{
		return $this->hasMany(SamplesIsolates::className(), ['isolate_id' => 'isolate_id'])->via('librariesIsolates')->alias('samplesIsolates');
	}

	/**
	 * @return \yii\db\ActiveQuery
	 */
	public function getSamples()
	{
		return $this->hasMany(Samples::className(), ['id' => 'sample_id'])->via('samplesIsolates')->alias('samples');
	}

	/**
	 * @return \yii\db\ActiveQuery
	 */
	public function getOrganisms()
	{
		return $this->hasMany(Organisms::className(), ['id' => 'organism_id'])->via('samples')->alias('organisms');
	}
	

	/**
	 * @return \yii\db\ActiveQuery
	 */
	public function getOrganismsPrograms()
	{
		return $this->hasMany(OrganismsPrograms::className(), ['organism_id' => 'id'])->via('organisms')->alias('organismsPrograms');
	}

	/**
	 * @return \yii\db\ActiveQuery
	 */
	public function getPrograms()
	{
		return $this->hasMany(Programs::className(), ['id' => 'program_id'])->via('organismsPrograms')->alias('programs');		
	}
```

ztdan4ik:
@GlyphDataServices 

**Shop**
has **Product[]** Windows 8 and Windows 10
Windows 10 has **ProductItem[]** Windows 10 _(two copies per order)_ and Windows 10 _(one copy per order)_
Windows 10 _(two copies)_ has **ProductDigit[]** №1 _(created by manager 1)_ and №2 _(created by manager 2)_
№1 has ProductDigitItem[] license code 1, license code 2, license code 3
№2 has ProductDigitItem[] license code 1, license code 2, license code 3

So,

Shops **hasMany** Product
each Product **hasMany** ProductItem
each ProductItem **hasMany** ProductDigit
each ProductItemDigit **hasMany** ProductDigitItem
each ProductDigitItem **hasMany** Orders

In back way it's always o2m, like you seen in my schema

I'll try wit aliases and give feedback here later

PR #11899, opened 2016-07-05
Add possibility of unset or force replace former value in `ArrayHelper::merge()`

### LINKED ISSUE #11275: FR: Allow replace and unset in ArrayHelper::merge()
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | #11253, #9982, #4021

Use Case
------------
Currently `ArrayHelper::merge()` only support for add or merge value. But, sometimes we need to replace or unset original array (set to null doesnt mean unset it). Eg, in `common/config/main.php` we have config like this
```php
'components' => [
    'cache' => [
        'class' => 'yii\caching\FileCache',
        'cachePath' => '@common/runtime/cache',
    ]
]
```
Then in `backend` we have config
```php
'components' => [
    'cache' => [
        'class' => 'yii\caching\DbCache',
        'db' => ['dsn' => 'sqlite:@common/runtime/cache.db'],
    ]
]
```
It will getting error because merged array still contain `cachePath` but `DbCache` has no `cachePath` property.
Another case. For non asosiated array, `ArrayHelper::merge()` always add new value to previous value.
```php
'as access' => [
    'class' => 'mdm\admin\components\AccessControl',
    'allowActions' => ['site/*', ],
],
// merge with
'as access' => [
    'class' => 'mdm\admin\components\AccessControl',
    'allowActions' => ['admin/*', ],
],
```
We cannot remove `'site/*'` value from array.

Solution
----------
I propose we introduce new class `None` and `Replace` (or another name class)
```php
class None
{

}

class Replace
{
    public $value;
    public function __construct($value)
    {
        $this->value = $value;
    }
}
```
Then in `ArrayHelper::merge()`
```php
    public static function merge($a, $b)
    {
        $args = func_get_args();
        $res = array_shift($args);
        while (!empty($args)) {
            $next = array_shift($args);
            foreach ($next as $k => $v) {
                if ($v instanceof None) {
                    unset($res[$k]);
                    continue;
                }
                if (is_int($k)) {
                    if (isset($res[$k])) {
                        $res[] = $v;
                    } else {
                        $res[$k] = $v;
                    }
                } elseif (is_array($v) && isset($res[$k]) && is_array($res[$k])) {
                    $res[$k] = self::merge($res[$k], $v);
                } else {
                    $res[$k] = $v instanceof Replace ? $v->value : $v;
                }
            }
        }

        return $res;
    }
```

Usage
--------
```php
'components' => [
    'cache' => [
        'class' => 'yii\caching\DbCache',
        'db' => ['dsn' => 'sqlite:@common/runtime/cache.db'],
        'cachePath' => new None(),
    ]
],
'as access' => [
    'class' => 'mdm\admin\components\AccessControl',
    'allowActions' => new Replace(['admin/*', ]),
],
```

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11275 |

closes  #11275
- [x] Implementation
- [x] Tests
- [x] Update guide

### DIFF
--- a/docs/guide/helper-array.md
+++ b/docs/guide/helper-array.md
@@ -303,21 +303,68 @@ Encoding will use application charset and could be changed via third argument.
 
 ## Merging Arrays <span id="merging-arrays"></span>
 
+You can use [[yii\helpers\ArrayHelper::merge()|ArrayHelper::merge()]] to merge two or more arrays into one recursively.
+If each array has an element with the same string key value, the latter will overwrite the former
+(different from [array_merge_recursive()](http://php.net/manual/en/function.array-merge-recursive.php)).
+Recursive merging will be conducted if both arrays have an element of array type and are having the same key.
+For integer-keyed elements, the elements from the latter array will be appended to the former array.
+You can use [[yii\helpers\UnsetArrayValue]] object to unset value from previous array or
+[[yii\helpers\ReplaceArrayValue]] to force replace former value instead of recursive merging.
+
+For example:
+
 ```php
-  /**
-    * Merges two or more arrays into one recursively.
-    * If each array has an element with the same string key value, the latter
-    * will overwrite the former (different from array_merge_recursive).
-    * Recursive merging will be conducted if both arrays have an element of array
-    * type and are having the same key.
-    * For integer-keyed elements, the elements from the latter array will
-    * be appended to the former array.
-    * @param array $a array to be merged to
-    * @param array $b array to be merged from. You can specify additional
-    * arrays via third argument, fourth argument etc.
-    * @return array the merged array (the original arrays are not changed.)
-    */
-    public static function merge($a, $b)
+$array1 = [
+    'name' => 'Yii',
+    'version' => '1.1',
+    'ids' => [
+        1,
+    ],
+    'validDomains' => [
+        'example.com',
+        'www.example.com',
+    ],
+    'emails' => [
+        'admin' => 'admin@example.com',
+        'dev' => 'dev@example.com',
+    ],
+];
+
+$array2 = [
+    'version' => '2.0',
+    'ids' => [
+        2,
+    ],
+    'validDomains' => new \yii\helpers\ReplaceArrayValue([
+        'yiiframework.com',
+        'www.yiiframework.com',
+    ]),
+    'emails' => [
+        'dev' => new \yii\helpers\UnsetArrayValue(),
+    ],
+];
+
+$result = ArrayHelper::merge($array1, $array2);
+```
+
+The result will be:
+
+```php
+[
+    'name' => 'Yii',
+    'version' => '2.0',
+    'ids' => [
+        1,
+        2,
+    ],
+    'validDomains' => [
+        'yiiframework.com',
+        'www.yiiframework.com',
+    ],
+    'emails' => [
+        'admin' => 'admin@example.com',
+    ],
+]
 ```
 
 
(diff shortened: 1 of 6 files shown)


rob006:
Just for record: this is ready to merge.

samdark:
Thank you! It's really good to check pull requests which are polished.

klimov-paul:
This fix does not seem right: what if I want to operate array containing `ReplaceArrayValue` objects?
This may be a general use case for composition of the argument for replacement of array.

Besides why do we need several extra classes for such trivial thing?

rob006:
> This fix does not seem right: what if I want to operate array containing ReplaceArrayValue objects?

You can use nested objects or put `ReplaceArrayValue` objects in first array. But generally you should not use `ArrayHelper::merge()` to prepare arguments for `ArrayHelper::merge()`...

> Besides why do we need several extra classes for such trivial thing?

IMO this is cleanest and safest way to do that. But I'm open to suggestion how to improve that.

klimov-paul:
> I'm open to suggestion how to improve that.

My opinion: the entire issue is wrong.

What is the noncense about unseting value via `ArrayHelper::merge()`?
Why anyone need to write following:

```php
ArrayHelper::merge($array, ['config' => ['options' => new \yii\helpers\UnsetArrayValue()]]);
```

Instead is simple:

```php
unset($array['config']['options']);
```

If you wish to replace value you can always unset it before merging:

```
unset($array['config']['options']);
ArrayHelper::merge($array, ['config' => ['options' => [...]]]);
```

Yii follows concept 'simplicity over comlexity', which seems to be forgotten here.

klimov-paul:
Creating 2 classes and introduction of complex syntax in order to get rid of sinle simpliest PHP operator call!
This does not feel right.

rob006:
Actually it is pretty useful when you have different files with config, like:

```php
$config = ArrayHelper::merge(
    require(__DIR__ . '/params.php'),
    require(__DIR__ . '/params-local.php')
);
```
You can directly override some values from `params.php` in `params-local.php`.

rob006:
Also, we already have similar objects:
https://github.com/yiisoft/yii2/blob/master/framework/db/Expression.php
https://github.com/yiisoft/yii2/blob/master/framework/web/JsExpression.php

Not mention exceptions...

klimov-paul:
> Also, we already have similar objects:

These are different cases.
`JsExpression` is handled by `Json::encode()`, which converts objects and arrays into the strings.
`JsExpression` just provide an algorithm of this conversion. It resturs the value, which should be used instead of it.
Same happens for `db\Expression`.
Both of these classes logic affects only thier instances and thier values.

In your case you create objects, which define processing for other data structure, not for them.

Also what will happen in following case:

```php
ArrayHelper::merge(
    $array,
    [
        'foo' => new \yii\helpers\ReplaceArrayValue([
            'some' => new \yii\helpers\ReplaceArrayValue(['anything']);
        ]);
    ],
);
```

rob006:
> These are different cases.

This is **exactly** the same case - this are extra dummy object that do only 2 things:
* Store some value (in most cases - simply string).
* Reserve namespace, that is used in further processing.

> Also what will happen in following case:

```php
    [
        ...
        'foo' => [
            'some' => new \yii\helpers\ReplaceArrayValue(['anything']);
        ],
    ]
```

PR #11907, opened 2016-07-06
entangled parameters Lines and Columns on Windows

Spell6inder:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | may be |
| Tests pass? | no |
| Fixed issues | yes |

entangled parameters Lines and Columns on Windows
'Lines' on line 3
'Columns' on line 4 

http://www.computerhope.com/modehlp.htm#examples

```
Status for device CON:
----------------------
Lines: 300
Columns: 80
Keyboard rate: 31
Keyboard delay: 1
Code page: 437
```

### DIFF
--- a/framework/helpers/BaseConsole.php
+++ b/framework/helpers/BaseConsole.php
@@ -614,7 +614,7 @@ class BaseConsole
             $output = [];
             exec('mode con', $output);
             if (isset($output, $output[1]) && strpos($output[1], 'CON') !== false) {
-                return $size = [(int) preg_replace('~\D~', '', $output[3]), (int) preg_replace('~\D~', '', $output[4])];
+                return $size = [(int) preg_replace('~\D~', '', $output[4]), (int) preg_replace('~\D~', '', $output[3])];
             }
         } else {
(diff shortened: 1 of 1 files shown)


cebe:
it would be better to search for the words instead of trusting the output to be always in order.

samdark:
I'd not rely on it either. It's as fragile as order.

samdark:
Hmm... probably you're right. Parsing names may be better...

samdark:
Fixed. Thanks!

kids-return:
`C:\Users\root>mode con

设备状态 CON:
---------
    行:　       9001
    列:　　     120
    键盘速度:   31
    键盘延迟:　 1
    代码页:     936


C:\Users\root>mode con cp select=437

Status for device CON:
----------------------
    Lines:          9001
    Columns:        120
    Keyboard rate:  31
    Keyboard delay: 1
    Code page:      437


C:\Users\root>`

samdark:
@kids-return seems to be OK, isn't it?

PR #11908, opened 2016-07-06
Add find or new method to base active record. last try

Razzwan:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | not found |

Very often in the code, we have to create a method that returns a new object, if not found in the database. I propose to add this method in the base class.


samdark:
As it was discussed before, the use cases for the method are rare so we won't merge it into core framework. Thanks for trying though (it's better to start a discussion first not to waste too much time on code).

Razzwan:
Is not it better to talk to a real code example?

samdark:
Of course, but that's more workload for you. We don't want to waste too much of your time.

PR #11910, opened 2016-07-06
automatic generete keys in migration

Razzwan:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | not found |

### DIFF
--- a/framework/db/Migration.php
+++ b/framework/db/Migration.php
@@ -8,6 +8,7 @@
 namespace yii\db;
 
 use yii\base\Component;
+use yii\base\InvalidParamException;
 use yii\di\Instance;
 
 /**
@@ -78,6 +79,32 @@ class Migration extends Component implements MigrationInterface
     protected function getDb()
     {
         return $this->db;
+    }
+
+    /**
+     * @param string $type type of key
+     * @param string $table table name
+     * @param string|array $columns the name of the column to that the constraint will be added on.
+     * If there are multiple columns, separate them with commas or use an array.
+     * @return string
+     */
+    protected function getKeyName($type, $table, $columns)
+    {
+        if ($columns === null) {
+            throw new InvalidParamException('Set $columns param for automatic generate key name');
+        }
+        $columnsString = is_array($columns) ? implode('_', $columns) : preg_replace("/\,\s*/", '_', $columns);
+
+        return implode('_', [$type, $this->getTableName($table), $columnsString]);
+    }
+
+    /**
+     * @param string $table table name, for example '{{%users}}'
+     * @return string
+     */
+    public function getTableName($table)
+    {
+        return str_replace(['"', "'"], '', $this->getDb()->quoteSql($table));
     }
 
     /**
@@ -371,12 +398,15 @@ class Migration extends Component implements MigrationInterface
     /**
      * Builds and executes a SQL statement for creating a primary key.
      * The method will properly quote the table and column names.
-     * @param string $name the name of the primary key constraint.
+     * @param string|null $name the name of the primary key constraint.
      * @param string $table the table that the primary key constraint will be added to.
      * @param string|array $columns comma separated string or array of columns that the primary key will consist of.
      */
     public function addPrimaryKey($name, $table, $columns)
     {
+        if ($name === null) {
+            $name = $this->getKeyName('pk', $table, $columns);
+        }
         echo "    > add primary key $name on $table (" . (is_array($columns) ? implode(',', $columns) : $columns) . ') ...';
         $time = microtime(true);
         $this->db->createCommand()->addPrimaryKey($name, $table, $columns)->execute();
@@ -387,9 +417,14 @@ class Migration extends Component implements MigrationInterface
      * Builds and executes a SQL statement for dropping a primary key.
      * @param string $name the name of the primary key constraint to be removed.
      * @param string $table the table that the primary key constraint will be removed from.
+     * @param null|string|array $columns the name of the column to that the constraint will be added on.
+     * If there are multiple columns, separate them with commas or use an array.
      */
-    public function dropPrimaryKey($name, $table)
+    public function dropPrimaryKey($name, $table, $columns = null)
     {
+        if ($name === null) {
+            $name = $this->getKeyName('pk', $table, $columns);
+        }
         echo "    > drop primary key $name ...";
         $time = microtime(true);
         $this->db->createCommand()->dropPrimaryKey($name, $table)->execute();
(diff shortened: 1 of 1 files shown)


rob006:
This is very risky behavior. The algorithm for generating key names can never change, because you will not be able to delete the existing keys. The same if the name of a table or column will be changed.

samdark:
I agree with @rob006. It is better to name keys yourself in this case.

samdark:
Thanks for the pull request anyway. I appreciate it.

Razzwan:
I use it in real project and all right. 
If you whant to generete key himself - you can. 
If you create migration with revert - you can use revert methods support this methenism.

Even if the algorithm for generating the name change, it will change the same everywhere. Otherwise, it will not be programming. So the problems can not arise.

A very short discussion of such a significant simplification, which is absolutely not hurt the existing mechanism. (

rob006:
> Even if the algorithm for generating the name change, it will change the same everywhere.

But it will not change existing keys names and new algorithm will generate a different name than when key was created.

samdark:
Well, probably algorithm won't change but the thing is that you can't have 2 keys with the same name and in this case it could go bad... Need more time to check all cons and pros.

Razzwan:
@rob006 
> But it will not change existing keys names and new algorithm will generate a different name than when key was created.

Is the existing scheme allows you to do this?

@samdark 
> Well, probably algorithm won't change but the thing is that you can't have 2 keys with the same name and in this case it could go bad... Need more time to check all cons and pros.

A khow this. And The proposed mechanism solves this problem. The key name is always generated by the following rule: `"$type_$tablePrefix_$tableName_" . implode('_', $columns)`

samdark:
One more thing. There's a limitation on FK name length.

rob006:
> Is the existing scheme allows you to do this?

I do not understand your question. Currently I, as a developer, decide about the name of the key, so I know it and I can always use it. In your solution I do not know the name of the key that I create - it dependents on the automagic that generating key name. So if I create a key, and after a year I want to delete it, then if in the meantime the algorithm that generates the key will change, I will not be able to do that because the migration will generate a different key name than a year ago. The same thing happens if in the meantime I change the name of the table - existing key name will not change and will use previous table name.

> I know. And The proposed mechanism solves this problem.

No, it doesn't. You can still have combinations of tables and columns names that will create the same key name.

```php
Migration::addForeignKey(null, 'test', ['a_b', 'c'], ...); 
Migration::addForeignKey(null, 'test', ['a', 'b_c'], ...); 
```

Razzwan:
@samdark 
> One more thing. There's a limitation on FK name length.

With automatic generation if FK length will be wrong - there will be the same as now, if you manually write the same length.

@rob006 
> Currently I, as a developer, decide about the name of the key, so I know it and I can always use it.

Automatic generate is option. If you whant to decide about the name of the key - you still can do this. Moreover, you can get the current key name from the method `getKeyName($type, $table, $columns)`

> So if I create a key, and after a year I want to delete it, then if in the meantime the algorithm that generates the key will change, I will not be able to do that because the migration will generate a different key name than a year ago.

And what are you doing now to remove the old foreign key? You will find it in the database. And you can delete it with old mechanism if you can not do it with the help of automatic generation.

> The same thing happens if in the meantime I change the name of the table - existing key name will not change and will use previous table name.

If the table name changed - you need to hand in any case take care of it. To do this, there are 2 ways: you can regenerate all foreign keys in this table in accordance with the automatic generation, but you can just use the old mechanism.

> No, it doesn't. You can still have combinations of tables and columns names that will create the same key name.

If you use such names of the columns in a table, then, sorry, but there's something wrong with you. Consider this - the same as to take into account the fall of a meteorite on your computer.

I suggested this pull request because very successfully using this path itself. And this saves time for the creation of indexes and keys.

samdark:
> With automatic generation if FK length will be wrong - there will be the same as now, if you manually write the same length.

Nope. Limitations are known so generation could be adjusted to fit names into the length.

Razzwan:
@samdark 
> Nope. Limitations are known so generation could be adjusted to fit names into the length.

I do not understand what you mean. The same problem existed before my code. I did not decide it. I did not create it - it just is.

rob006:
> And what are you doing now to remove the old foreign key? You will find it in the database.

No. I will find key name in specified **migration** code - in many cases I don not have access to database of all installations. If application has 5 instances, each one could run specified migration in different time and have different key names. This behavior breaks main migration purpose - represents specified state of database. If result of migration depends on time when it is run, it means that something is wrong. This is the same type of solution like `magic_quotes` in PHP - it was intended to simplify common operations, but turned out to be completely unreliable and using this was a bad practice.

> If you use such names of the columns in the One table, then, sorry, but there's something wrong with you.

It was only a example, use your imagination. :) Assuming that your solution is bulletproof and no one ever will create a database that will break it, is simply naive. There already were similar problems with migration names (#10876), so I'm sure that same issues will return also in this case. But this time you can not ever fix this, because any change in the algorithm will break existing migrations that rely on this feature.

There is also existing and more reliable solution for this: http://www.yiiframework.com/doc-2.0/guide-db-migrations.html#foreign-keys

Razzwan:
@rob006 
> No. I will find key name in specified migration code - in many cases I don not have access to database of all installations. If application has 5 instances, each one could run specified migration in different time and have different key names. This behavior breaks main migration purpose - represents specified state of database. If result of migration depends on time when it is run, it means that something is wrong. This is the same type of solution like magic_quotes in PHP - it was intended to simplify common operations, but turned out to be completely unreliable and using this was a bad practice.

OK. In your case, you can use the old mechanism. My pull request is - it is an option for more localization projects.

> Assuming that your solution is bulletproof and no one ever will create a database that will break it, is simply naive

My wife manages to kill a Linux system in a couple of months. For talented people, nothing is impossible - then I agree with you.

> There already were similar problems with migration names (#10876), so I'm sure that same issues will return also in this case. But this time you can not ever fix this, because any change in the algorithm will break existing migrations that rely on this feature.
There is also existing and more reliable solution for this: http://www.yiiframework.com/doc-2.0/guide-db-migrations.html#foreign-keys

I never touched any of the existing mechanisms. Only the added option of automatically generating, if key name set null - that's all.

samdark:
I've meant that if we'd ever adopt this generation it should not generate names which are out of maximum key name length database limit.

Razzwan:
> I've meant that if we'd ever adopt this generation it should not generate names which are out of maximum key name length database limit.

OK. I will fix it.

SilverFire:
The PR seems to be abandoned.

Issue #11895, opened 2016-07-05
Get all users having a RBAC role assignment


Kolyunya:
Current implementation of RBAC managers can find users having only direct assignments to a specific role.

If role `Foo` is a parent of role `Bar` and role `Foo` is assigned to some user then `getUserIdsByRole()` won't list that user as having role `Bar`.

I think we should implement a method which will find all users having all (even indirect) assignments to a specific role.

pstddc:
Колюня, ты не понимаешь что такое рбак и как им пользоваться, рбак это доступ на основе ролей, вообще если утебя более 1 роли привязано к пользователю, то гдето ты свернул не на ту тропу и начал велосипедить, к тому же, то что ты предлагаешь довольно таки накладная операция... Да и вообще вот получишь ты что админ имеет 3 роли - юзер, модератор, админ, попробуй отвязать юзера - ведь по твоей логике он же принадлежит роли юзер

Kolyunya:
@pstddc thank you for your comment but the [official guide](http://www.yiiframework.com/doc-2.0/guide-security-authorization.html#basic-concepts) states that organizing roles in hierarchies is the way the RBAC framework is designed to be used.
> Both roles and permissions can be organized in a hierarchy. In particular, a role may consist of other roles or permissions;

Also I understand that this is an expensive operation but it is required in particular scenarios.

githubjeka:
RbacManager has all the means to solve this task:
```php
$this->addIds($manager->getUserIdsByRole($roleName));

foreach($manager->getChildren($roleName) as $child) {
    if ($child->type === Item::TYPE_ROLE) {
         $this->addIds($manager->getUserIdsByRole($child->name));
    }
}
```

P.S. 
@pstddc к пользователю можно привязать сколько угодно ролей. Это не противоречит спецификации RBAC [см. pdf](http://csrc.nist.gov/rbac/sandhu-ferraiolo-kuhn-00.pdf)

Kolyunya:
@githubjeka thank you for your response. Of-course we can solve the task manually. My point is that there should be a built-in method in auth-manager since we have `getUserIdsByRole` there.

At the moment the manager's interface is incomplete. It has the method to get direct role users (which also can be solved manually) but it does not have a method to get all (including indirect) role users.

pstddc:
@githubjeka роль должна быть одна, а остальное это пермишены и рулесы

Kolyunya:
[...]

PR #11927, opened 2016-07-09
Enhance `yii\rbac\ManagerInterface::getUserIdsByRole()`

Kolyunya:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11895 |

It is now possible to get users by role recursively.

### DIFF
--- a/framework/rbac/DbManager.php
+++ b/framework/rbac/DbManager.php
@@ -15,6 +15,7 @@ use yii\db\Expression;
 use yii\base\InvalidCallException;
 use yii\base\InvalidParamException;
 use yii\di\Instance;
+use yii\helpers\ArrayHelper;
 
 /**
  * DbManager represents an authorization manager that stores authorization information in database.
@@ -753,6 +754,24 @@ class DbManager extends BaseManager
     /**
      * @inheritdoc
      */
+    public function getParents($name)
+    {
+        $query = (new Query)
+            ->select(['name', 'type', 'description', 'rule_name', 'data', 'created_at', 'updated_at'])
+            ->from([$this->itemTable, $this->itemChildTable])
+            ->where(['child' => $name, 'name' => new Expression('[[parent]]')]);
+
+        $parents = [];
+        foreach ($query->all($this->db) as $row) {
+            $parents[$row['name']] = $this->populateItem($row);
+        }
+
+        return $parents;
+    }
+
+    /**
+     * @inheritdoc
+     */
     public function getChildren($name)
     {
         $query = (new Query)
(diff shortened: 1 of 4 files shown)


framework/rbac/DbManager.php
```diff
@@ -753,6 +754,24 @@ public function hasChild($parent, $child)
     /**
      * @inheritdoc
      */
+    public function getParents($name)
```
pstddc: may be add second argument `getParents($name, $type = Item::TYPE_ROLE)` and add condition for select?

Kolyunya:
@samdark I would be glad to make any improvements to this patch if you provide me with information about it's issues.

framework/rbac/DbManager.php
samdark: Yes, that could be useful.

framework/rbac/ManagerInterface.php
```diff
@@ -161,6 +161,13 @@ public function removeChildren($parent);
     public function hasChild($parent, $child);
 
     /**
+     * Returns the parent permissions and/or roles.
+     * @param string $name the child name
+     * @return Item[] the parent permissions and/or roles
+     */
```
samdark: Need `@since 2.0.10`

framework/rbac/DbManager.php
```diff
@@ -966,18 +985,36 @@ public function loadFromCache()
     /**
      * Returns all role assignment information for the specified role.
      * @param string $roleName
+     * @param string $recursive
```
samdark: Need to elaborate what is recursive and how it works.

framework/rbac/ManagerInterface.php
```diff
@@ -212,10 +219,11 @@ public function getAssignments($userId);
     /**
      * Returns all user IDs assigned to the role specified.
      * @param string $roleName
+     * @param string $recursive
```
samdark: Need description.

framework/rbac/DbManager.php
Kolyunya: This would be inconsistent with `getChildren()` and require its refactoring too.

samdark: OK.

Kolyunya:
@samdark this feature doesn't break the BC. Why can't we have it in `2.0.x`?

dynasource:
@Kolyunya, thats because you are changing an interface signature at https://github.com/yiisoft/yii2/pull/11927/files#diff-72766818d5fa80f81ef61a9d0dde4b63L218

Kolyunya:
@dynasource I do change the signature but it's compatible with the previous version and can not break the backward compatibility.

samdark:
The interface change is actually backwards compatible.

Kolyunya:
@samdark `ManagerInterface` changed. Custom implementations will break. How is it backward compatible?

samdark:
https://3v4l.org/VSVdn — default value is a key.

Kolyunya:
@samdark **custom** managers will break after **interface** change, won't they?

samdark:
They won't if `$recursive` value has a default.

Kolyunya:
@samdark could you provide an example? It seems, [it's not going to work out](https://3v4l.org/A48Rh).

samdark:
Reduction of parameters doesn't work like that. Adding more parameters with default values does.

samdark:
So yeah. You're right.

Issue #11912, opened 2016-07-06
Postgresql negative default values support for integer/float/decimal columns


nsknewbie:
Hello!
Currently yii doesn't handle negative default values and converts them to zero, because postgresql returns them in parenthesis, e.g. `(-1.234)`.

samdark:
How to reproduce the issue?

nsknewbie:
Test:
```php
$connection = new \yii\db\Connection([...]);
$connection->createCommand('CREATE TABLE issue11912 (col INTEGER DEFAULT -1)')->execute();
$schema = new \yii\db\pgsql\Schema(['db' => $connection]);
$tableSchema = $schema->loadTableSchema('issue11912');
$connection->createCommand('DROP TABLE issue11912');

$this->assertEquals('-1', $tableSchema->getColumn('col')->defaultValue);
```

Output:
```
Failed asserting that 0 matches expected '-1'.
Expected :-1
Actual   :0
```

samdark:
Thank you! If you have extra time you may speed things up by pull requesting a fix.

PR #11933, opened 2016-07-11
#11912: Postgresql negative default values support for integer/float/decimal columns

nsknewbie:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | possibly yes, if users expected `0` instead of correct default value |
| Tests pass? | yes |
| Fixed issues | #11912 |

New regexp looks bulky, it can be improved for sure, but i tried to make it stricter


cebe:
thanks! the tests could possibly be added for other dbms too so we can check whether it needs to be fixed in other places too.

nsknewbie:
@cebe added more tests, seems all's fine

framework/db/pgsql/Schema.php
```diff
@@ -468,7 +468,7 @@ protected function findColumns($table)
                     $column->defaultValue = bindec(trim($column->defaultValue, 'B\''));
                 } elseif (preg_match("/^'(.*?)'::/", $column->defaultValue, $matches)) {
                     $column->defaultValue = $matches[1];
-                } elseif (preg_match('/^(.*?)::/', $column->defaultValue, $matches)) {
+                } elseif (preg_match('/^(?:\()?(.*?)(?(1)\))(?:::.+)?$/', $column->defaultValue, $matches)) {
```
cebe: just for me to fully understand this: why did you add the part that requires text after `::`? are you sure that can not be empty?

cebe:
please also add a line to the CHANGLOG.

framework/db/pgsql/Schema.php
nsknewbie: there should be a data type after `::`

cebe:
Merged, thank you!

PR #11950, opened 2016-07-14
Improve BaseArrayHelper::keyExists speed a little bit

egorio:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

Somebody can use BaseArrayHelper::keyExists in loops. 
I think we can improve BaseArrayHelper::keyExists with added isset(...) before array_key_exists because it works a little bit faster.

We also can do it in some other places, but I'm not sure it's necessary.

Ref: http://php.net/manual/en/function.array-key-exists.php#107786


framework/helpers/BaseArrayHelper.php
```diff
@@ -486,7 +486,7 @@ public static function map($array, $from, $to, $group = null)
     public static function keyExists($key, $array, $caseSensitive = true)
     {
         if ($caseSensitive) {
-            return array_key_exists($key, $array);
+            return isset($array[$key]) || array_key_exists($key, $array);
```
samdark: This changes behavior:

```php
$array = ['test' => null];
echo array_key_exists('test', $array); // true
echo isset($array[$key]); // false
```

samdark:
Makes sense. Please add inline comment explaining the code since it's not clear right away + changelog.

egorio:
@samdark Done ;)

samdark:
Thanks!

PR #11954, opened 2016-07-14
Fix double space after array key

nkovacs:
I'm pretty sure one space after the key was the originally intended style.


samdark:
Thanks!

PR #11959, opened 2016-07-14
Added default value for `get` method in the cache class

siamskoi:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.10 under development
 ------------------------
-
+- Enh: Added default value for `get` method in the cache class
 - Bug #11949: Fixed `ActiveField::end` generates close tag when it's `option['tag']` is null (egorio)
 - Bug #11912: Fixed PostgreSQL Schema to support negative default values for integer/float/decimal columns (nsknewbie)
 - Bug #11947: Fixed `gridData` initialization in `yii.gridView.js` (pavlm)
--- a/framework/caching/Cache.php
+++ b/framework/caching/Cache.php
@@ -94,11 +94,12 @@ abstract class Cache extends Component implements \ArrayAccess
     /**
      * Retrieves a value from cache with a specified key.
      * @param mixed $key a key identifying the cached value. This can be a simple string or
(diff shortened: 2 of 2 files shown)


alexndr-novikov:
Isn't it better not to pass the default value, and return it later, but to pass value or closure  and store it in cache if it wasn't found there?
I know it's not a good idea to let one function to set and get values simultaneously, but it's really often used operation by me: check if value is in cache, return it if it is or set it and return the result if not

samdark:
@alexndr-novikov you mean something like the following?

```php
$value = $cache->get('key', function() {
    return 'value';
}, ...);
```

alexndr-novikov:
@samdark yep, exactly
It would be rather convenient to declare this function before calling $cache->get, and call it only if cache value doesn't exist. 

```
$newVal = function() {
    return Blablaclass::calculateValue(); // or whatever
}
$value = $cache->get('key', $newVal);
```

and in framework/caching/Cache.php:
```
public function get($key, $closure = false)
{
  $key = $this->buildKey($key);
  $value = $this->getValue($key);
  if (is_array($value) && !($value[1] instanceof Dependency && $value[1]->getHasChanged($this))) {
      return $value[0];
  } else {
  	return ($closure instanceof \Closure) ? $closure() : $closure;
  }
}
```

Not sure about the syntax and arguments  pass, but idea is like this

rob006:
@alexndr-novikov You'll probably want to save the generated value. Maybe we should create new method for that.

```php
public function getOrGenerate('key', $callback, $duration = 0, $dependency = null);
```

If cache for `key` does not exist, method will generate new value from `$callback`, save it in cache and return it.

cebe:
we had a similar discussion just recently and decided not to add it. #11172 + #11173

samdark:
The issue was raised again independently. It seems it could be a handy method despite doing the same thing differently. Personally I'm not sure if I'd prefer it or the current explicit method. Both are fine to me.

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/aed3a222a20edd14859053a45b6b62889ca41a59

PR #11963, opened 2016-07-15
if configuration cache server but yii_debug is true and yii_env is dev so urlrule isn't update in main.php

aslijiasheng:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | yes |
| Tests pass? | yes |

```
     if configuration cache server but yii_debug is true and yii_env is dev so urlrule isn't update in main.php
     unless cache LRU expired
```

…v so urlrule is not update in main.php


cebe:
you should disable caching generally in dev environment. You can use `DummyCache` class to test with enabled caching.

Issue #11923, opened 2016-07-08
AttributeBehavior should be able to respect current attributes values


Kolyunya:
I would like to propose an improvement to an `AttributeBehavior` which will allow it to respect current attributes values. The behavior should update only non-empty attributes when a special flag like `skipUpdateOnFilled` is set to `true`. This won't break the BC if we set this flag to `false` by default.

Use-case: `Page` entity has three attributes: `meta_title`, `meta_description`, `meta_keywords`. We need to copy `title` to each of the attributes _only if_ they are empty. This will allow users to set some of them manually and to let the behavior fill the others.

The behavior should be defined like this:

``` php
[
    'class' => AttributeBehavior::className(),
    'skipUpdateOnFilled' => true,
    'attributes' => [
        self::EVENT_AFTER_VALIDATE => [
            'meta_title',
            'meta_description',
            'meta_keywords',
        ],
    ],
    'value' => function ($event) {
        return $event->sender->title;
    },
],
```

At the moment the `AttributeBehavior` won't allow you to achieve this behavior. The only workaround is to add three behaviors for each attribute with custom `value` which will check if the corresponding attribute is empty or not.

The problem can't be solved even by extending the base behavior without copy-pasting the `evaluateAttributes()` method.

samdark:
@yiisoft/core-developers thoughts?

I think it could be useful.

Kolyunya:
By the way the `SluggableBehavior::immutable` illustrates that the option I propose is *really* needed.

SilverFire:
Good idea, could be implemented.

@Kolyunya, would you like to submit a PR?

Kolyunya:
@SilverFire thank you! I've already did a pull request: #11969. It needs a review.

PR #11969, opened 2016-07-17
Enhance `yii\behaviors\AttributeBehavior`

Kolyunya:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11923 |

`AttributeBehavior` can preserve current attribute values now.
Basic tests coverage is included.

### DIFF
--- a/framework/behaviors/AttributeBehavior.php
+++ b/framework/behaviors/AttributeBehavior.php
@@ -86,6 +86,11 @@ class AttributeBehavior extends Behavior
      * @since 2.0.8
      */
     public $skipUpdateOnClean = true;
+    /**
+     * @var boolean whether to skip attribute update when it is not empty
+     * @since 2.0.10
+     */
+    public $skipUpdateOnFilled = false;
 
 
     /**
@@ -118,6 +123,10 @@ class AttributeBehavior extends Behavior
             foreach ($attributes as $attribute) {
                 // ignore attribute names which are not string (e.g. when set by TimestampBehavior::updatedAtAttribute)
                 if (is_string($attribute)) {
+                    // skip attribute update when it is not empty and the corresponding mode is enabled
+                    if ($this->skipUpdateOnFilled && !empty($this->owner->$attribute)) {
+                        continue;
+                    }
                     $this->owner->$attribute = $value;
                 }
             }
--- a/tests/framework/behaviors/AttributeBehaviorTest.php
+++ b/tests/framework/behaviors/AttributeBehaviorTest.php
@@ -1,0 +1,139 @@
+<?php
+
+namespace yiiunit\framework\behaviors;
+
+use Yii;
+use yiiunit\TestCase;
+use yii\db\Connection;
+use yii\db\ActiveRecord;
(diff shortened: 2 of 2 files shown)


samdark:
Hmm... this makes https://github.com/yiisoft/yii2/blob/master/framework/behaviors/SluggableBehavior.php#L88 pointless so either `immutable` should be moved into base `AttributeBehavior` or `immutable` should be deprecated.

@Kolyunya what do you think would be the best?

Kolyunya:
@samdark I'm not sure but I think it is important to maintain backward compatibility. If it is, we should choose the first option.

samdark:
Both my suggestions are backwards compatible, aren't they?

Kolyunya:
@samdark yes, they are. I just thought that deprecated properties and methods are subject for removal.

samdark:
In 2.1 — yes.

samdark:
Thinking about it more, condition for not touching a field could be anything: only certain existing value, value of another field etc.

samdark:
Since closures are supported for values and we have dirty attributes support, we can achieve it using existing code:

```php
'value' => function($event) {
    if (empty($this->create_time)) {
        return time();
    }

    return $this->create_time;
}
```

samdark:
So I'd decline this pull request and document ↑

Kolyunya:
@samdark I'm sorry but we can't achieve this using existing code in a general scenario when we are populating *multiple* attributes.

Use-case: Page entity has three attributes: meta_title, meta_description, meta_keywords. We need to copy title to each of the attributes only if they are empty. This will allow users to set some of them manually and to let the behavior fill the others.

The only workaround is to add three behaviors for each attribute with custom value which will check if the corresponding attribute is empty or not which is quite inelegant.

samdark:
`$skipUpdateOnFilled` is per-behavior switch. Code above is the same.

samdark:
Right, it's not convenient, at least...

DrDeath72:
just want to write about it, i want use BlameableBehavior with beforeValidate without rewrite filled user_id

Faryshta:
how about a `$when` callable that can return `false` if you don't want the attribute to update?

Kolyunya:
@samdark @cebe what is the status of this PR? Is it under discussion or does it need some modifications?

framework/behaviors/AttributeBehavior.php
```diff
@@ -86,6 +86,11 @@ class AttributeBehavior extends Behavior
      * @since 2.0.8
      */
     public $skipUpdateOnClean = true;
+    /**
+     * @var boolean whether to skip attribute update when it is not empty
+     * @since 2.0.10
```
samdark: 2.0.12

samdark (changes requested):
Changelog line needed.

Kolyunya:
@samdark I wasn't able to push to this PR (since I already deleted that branch and repo) and I made a [new one](https://github.com/yiisoft/yii2/pull/13586).

Issue #11804, opened 2016-06-23
Strict attribute types


NaLLiFFuNT:
### The problem
I work on api. I use RedisActiveRecord.
I do create request 
{
 stringValue : '123'
 intValue : 20,
 boolValue : true
}

I load AR model & save it.
So, model have integer validator for intValue & bool Validator for boolValue 

when i do get request
I find model by Id from RedisDB & return $model->toArray().
the ploblem that after loading from DB intValue & boolValue become string '20' & '1'.  

### Suggestion
Maybe that is not a problem for PHP7. But at this momen i don't have ability to upgrade a php version. I guess, you can face the challenge with other DB & AR.
I'd suggest following solution.
Add method attributeTypes to BaseActiveRecord which return array in following format
[
 'attributeName1'=> 'int'
 'attributeName2'=> 'bool'
 'attributeName3'=> 'string'
]

in __set method we have to verify type of value for attribute & bring it, if it's required.

### Additional info
I saw similar solution at Laravel and i find it very usefull.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.5
| PHP version      | 5.6
| Operating system | Ubuntu mate

klimov-paul:
Resolved by #12067

PR #11974, opened 2016-07-18
11804 active record attributes typecasting

NaLLiFFuNT:
| New feature?  | yes
| Fixed issues  | https://github.com/yiisoft/yii2/issues/11804

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -62,7 +62,7 @@ Yii Framework 2 Change Log
 - Chg #11364: Updated jQuery dependency to include versions `1.12.*` (cebe)
 - Bug #11252: Use strict comparison on ActiveRecord::hasAttribute() to avoid cases where it returns true when a number is passed to it (Faryshta)
 - Chg #11683: Fixed fixture command to work with short syntax. `yii fixture "*, -User"` should be used instead of `yii fixture "*" -User` (Faryshta, samdark)
-
+- Enh #11804: Added Typecasting for ActiveRecord attributes. (nalliffunt)
 
 2.0.8 April 28, 2016
 --------------------
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -8,7 +8,6 @@
 namespace yii\db;
 
 use yii\base\InvalidConfigException;
-use yii\base\Event;
 use yii\base\Model;
 use yii\base\InvalidParamException;
 use yii\base\ModelEvent;
@@ -16,6 +15,7 @@ use yii\base\NotSupportedException;
 use yii\base\UnknownMethodException;
 use yii\base\InvalidCallException;
 use yii\helpers\ArrayHelper;
+use yii\helpers\TypecastingHelper;
 
 /**
  * ActiveRecord is the base class for classes representing relational data in terms of objects.
@@ -97,6 +97,10 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      */
     private $_related = [];
 
+    /**
+     * @var array|null array of cast rules for attributes
+     */
+    private $_castsRules = null;
 
     /**
      * @inheritdoc
@@ -266,6 +270,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      */
     public function __set($name, $value)
     {
+        $value = $this->castAttribute($name, $value);
         if ($this->hasAttribute($name)) {
             $this->_attributes[$name] = $value;
         } else {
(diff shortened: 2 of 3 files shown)


samdark:
Seems to be BC compatible.

samdark:

klimov-paul:
I am against the change.

Issue #11804 is a Redis specific trouble. Attributes typecasting for relational database works fine, in its scope this change is absolutely redundant.
If it should be fixed - it should be fixed at `yii2-redis` repository.

Besides solution based on static helper `TypecastingHelper` does not looks appropriate.

NaLLiFFuNT:
You can face this problem with MySQL too. For instance, boolean values are strored as TINYINT. When you load model from MySQL, model's filed have value 1 instead true value, which you saved before. Anyway if you won't use this feature, AR's behavior won't change.

cebe:
does the issue occur in any other place than tinyint in mysql? we should just get the handling of tinyint(1) to be fixed then.

klimov-paul:
> does the issue occur in any other place than tinyint in mysql? we should just get the handling of tinyint(1) to be fixed then.

In case you create field using migration mechanism and use `boolean` type for it - there will be no problem whatever DB driver you are using.
`tinyint` DB type can strore more then 2 values even if it length set to 1. In particular possible values will be `1`, `0`, `-1`. Only `unsigned tinyint` allows 2 values only.
Although it can be fixed, there is no complain on such problem so far.

NaLLiFFuNT:
No, the problem is only with TINYINT and boolean in MySQL. Do you suggest to apply this or similar solution for Redis-AR only? Or, may be other suggestion?
----
Other example.  You load data from the POST request. All data come as string. You load data into model and validate it. At this moment you can work with your model, but data fields have string values. You have to save model and load model from DB to make typecasting of their fields.

klimov-paul:
> Other example. You load data from the POST request. All data come as string. You load data into model and validate it. At this moment you can work with your model, but data fields have string values. You have to save model and load model from DB to make typecasting of their fields.

This is already covered by filters validators such as `NumberValidator` and `BooleanValidator`.

klimov-paul:
> Do you suggest to apply this or similar solution for Redis-AR only? Or, may be other suggestion?

Attributes typecasting for relational database is already covered by other means.
MongoDB and ElasticSearch do not have such problem at all: they operate via JSON and its ability of data-type storage.
So, yes I consider this is a Redis only problem.

NaLLiFFuNT:
NumberValidator and BooleanValidator only verify that your string match with number pattern. This validators don't make typecasting of your value.

klimov-paul:
> NumberValidator and BooleanValidator only verify that your match with number pattern. This validators don't make typecasting of your value.

Indeed, I admin I was wrong about that.

Speaking of validation. Imagine we have a model with attribute 'price' which is considered as `integer`.
Thus this attribute has a validator `NumberVallidator` with `integerOnly` set to `true`.
We have a web form (or REST API) for our model, which has an imput for 'price'.

With this fix in case user enters value `abc` it will be automatically and silently typecasted to `0` and model validation will pass, without notifying user about his mistake.
Thus I suppose this fix is unacceptable, since it could ruin the model validation workflow.

The typecasting could be done before saving record, after restoring record from DB or after successsful validation. But it should NOT be performed on attribute setter level.

klimov-paul:
Alternative solution proposed at #12067

NaLLiFFuNT:
> The typecasting could be done before saving record, after restoring record from DB or after successsful validation. But it should NOT be performed on attribute setter level.

I have thought about it. My realisation should be used in pair with validation rules .

For instance:

We have AR with fields intValue, boolValue, stringValue & etc. We add attibuteCasts() (typecast rules)  method & rules() (validation rules) method .
Following code is part of model:
```
public function rules() {
    return [
        ['stringValue', 'string'],
        ['boolValue', 'boolean'],
        ['boolValue', 'default', 'value' => true],
        ['intValue', 'integer'],
        [['intValue1','intValue2'], 'integer']
    ];
}

public function attributeCasts() {
    return [
        ['someString', 'string'],
        ['boolValue', 'boolean'],
        ['intValue', 'integer'],
        [['intValue1','intValue2'], 'integer']
    ];
}
```

How we can use it. By default typecasting will be in soft mode. If it's impossible to convert some value into scalar value of defined type, attribute will have raw value. There are some examples:

```
$model->intValue = '123'; // attribute will get 123 value
$model->load(['intValue'=>'200'], '') // attribute will get 200 value
$model->intValue = '124a3'; // attribute will get'124a3' 
//(because, it isn't an integer value and it cannot be coverted to int), 
//after that you will use validator before saving model and it will say: you have some error.
```


My realization solves following issues:
1) issue with Redis-AR ( it made me find some solution)
2) issue with typecasting tinyint to bool
3) you can use strict comparison of model fields before you save model & load model from DB

So, it doesn't matter how you set attributes values, you can load it from DB, from any POST\GET request, you can set set some attibute's value by one, if it's possible, value converts into scalar value of defined type.

I believe, this solution is very helpful.

Now i have described my suggestion completely. What do you think about it?

PS: Alternative solution proposed at #12067, also looks good.

klimov-paul:
> My realisation should be used in pair with validation rules

I have added auto dection for the typecast attribute map from validation rules:
https://github.com/yiisoft/yii2/pull/12067/commits/608b886a4fc15027f3a41b7c326fad246f2ca566

> Now i have described my suggestion completely. What do you think about it?

I still think there should not be any typecast at setter level. 
In most cases typecasting is not needed at all. Lets take your 'tinyint' example. Any value, which passes `BooleanValidator` will be saved into 'tinyint' column correctly without extra typecasting to `boolean`.
Maintainance of strict types is requried only in relatively rare cases. Thus typecast should be triggered manually only if needed.

NaLLiFFuNT:
In that case, i guess your solution is the best. So I guess we can decline & close this pull request.

PR #11981, opened 2016-07-19
11979 oracle mutex

zlakomanoff:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |


framework/mutex/OracleMutex.php
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * @author Alexander Zlakomanov <zlakomanoff@gmail.com>
+ * @since 1.0
+ */
+class OracleMutex extends DbMutex
+{
+    /**
+     * @var string driver name.
+     */
+    public $driverName = 'oci';
```
klimov-paul: Does not make sense: it is better to check driver name at `init()` method against all options:
- oci
- odbc

framework/mutex/OracleMutex.php
```diff
@@ -0,0 +1,131 @@
...
+     * @var array available lock modes
+     */
+    private $availableLockModes = [
+        'X_MODE',
+        'NL_MODE',
+        'S_MODE',
+        'SX_MODE',
+        'SS_MODE',
+        'SSX_MODE'
+    ];
```
klimov-paul: Better to be constants.

framework/mutex/OracleMutex.php
```diff
@@ -0,0 +1,131 @@
...
+    /**
+     * Acquires lock by given name.
+     * @param string $name of the lock to be acquired.
+     * @param integer $timeout to wait for lock to become released.
+     * @param string $lockMode lock mode.
+     * @param boolean $releaseOnCommit release on commit.
+     * @return boolean acquiring result.
+     * @see http://docs.oracle.com/cd/B19306_01/appdev.102/b14258/d_lock.htm
+     */
+    protected function acquireLock($name, $timeout = 0, $lockMode = 'X_MODE', $releaseOnCommit = false)
```
klimov-paul: It make absolutely no sense to add extra parameters for the protected method: no one will be able to use them unless he extends this class.
It is better to move extra options to the component public properties.

framework/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Yii Framework 2 Change Log
 
 - Bug #11912: Fixed PostgreSQL Schema to support negative default values for integer/float/decimal columns (nsknewbie)
 - Bug #11947: Fixed `gridData` initialization in `yii.gridView.js` (pavlm)
+- Enh #11979: Oracle mutex (zlakomanoff)
+
```
klimov-paul: Redundant blank line.

framework/mutex/OracleMutex.php
```diff
@@ -0,0 +1,131 @@
...
+        'NL_MODE',
+        'S_MODE',
+        'SX_MODE',
+        'SS_MODE',
+        'SSX_MODE'
+    ];
+
+    /**
+     * Initializes Oracle specific mutex component implementation.
+     * @throws InvalidConfigException if [[db]] is not MySQL connection.
```
klimov-paul: 'is not **MySQL** connection' => 'is not **Oracle** connection'

framework/mutex/OracleMutex.php
zlakomanoff: Not sure, im currently use custom driver name like oci8, we need this option for setup.

klimov-paul: Option make absolutely no sense here: you just make develper to setup extra field in case he uses not `oci` PDO driver.

Either the check should be dropped completely or make more flexible.
For example:

```php
if (strpos($this->db->driverName, 'oci') !== 0 || strpos($this->db->driverName, 'odbc') !== 0) {
    // exception
}
```

framework/mutex/OracleMutex.php
zlakomanoff: This is setup for current lock, in one session we can create to many locks with different behaviors.
If we move this setup to component setup, we lose a lot of goods.

framework/mutex/OracleMutex.php
zlakomanoff: Done

framework/CHANGELOG.md
zlakomanoff: done

framework/mutex/OracleMutex.php
zlakomanoff: done

framework/mutex/OracleMutex.php
klimov-paul: It seems you do not listen.
Method `acquireLock()` is **protected**, thus it can NOT be called from outside the mutex component:

```php
Yii::$app->mutex->acquireLock('mylock'); // PHP fatal error!
```

zlakomanoff: I think this is readon for extend yii\mutex\Mutex->acquire method

klimov-paul: Signature of the `acquire` method can not be changed.
Also keep in mind that `Mutex` class declares universal interface: it may lack some specific features, but following it you can switch from one mutex to another without code adjustments - this is important.

zlakomanoff: Updated previos comment

klimov-paul: > Mutex class declares universal interface: it may lack some specific features, but following it you can switch from one mutex to another without code adjustments - this is important.

klimov-paul: > Another solution - move this options to setup component options

This is what I am talking to you from the begining!

zlakomanoff: Relax man, fixed

framework/mutex/OracleMutex.php
zlakomanoff: fixed

framework/mutex/OracleMutex.php
rob006: This is invalid for PHP 5.4 and 5.5 - prior to 5.6 constants can be only scalars.

framework/mutex/OracleMutex.php
```diff
@@ -0,0 +1,137 @@
...
+ *              ...
+ *         ],
+ *     ],
+ * ]
+ * ```
+ *
+ * @see Mutex
+ *
+ * @author Alexander Zlakomanov <zlakomanoff@gmail.com>
+ * @since 1.0
```
rob006: `@since 2.0.10`?

zlakomanoff: fixed

framework/mutex/OracleMutex.php
zlakomanoff: revert to array

klimov-paul:
`composer.lock` should NOT be commited.

framework/mutex/OracleMutex.php
```diff
@@ -0,0 +1,137 @@
...
+        'SX_MODE',
+        'SS_MODE',
+        'SSX_MODE'
+    ];
+
+    /**
+     * @var string lock mode
+     */
+    public $lockMode = 'X_MODE';
+
```
klimov-paul: Redundant line

framework/mutex/OracleMutex.php
```diff
@@ -0,0 +1,137 @@
...
+     */
+    protected function acquireLock($name, $timeout = 0)
+    {
+        $lockStatus = null;
+
+        /** clean vars before using */
+        $releaseOnCommit = ($this->releaseOnCommit === true) ? 'TRUE' : 'FALSE';
+        $timeout = abs((int)$timeout);
+
+        if(!in_array($this->lockMode, $this->availableLockModes)){
```
klimov-paul: Code style is invalid

framework/mutex/OracleMutex.php
```diff
@@ -0,0 +1,137 @@
...
+ * ```
+ *
+ * @see Mutex
+ *
+ * @author Alexander Zlakomanov <zlakomanoff@gmail.com>
+ * @since 2.0.10
+ */
+class OracleMutex extends DbMutex
+{
+
```
klimov-paul: redundant line

framework/mutex/OracleMutex.php
klimov-paul: In this case restriction should be removed completely: there is no sense to keep this array if its value can not be used anyhow.
It is better to list all available options at PhpDoc for `lockMode` instead.

framework/mutex/OracleMutex.php
samdark: Also 3rd argument should be specified as `true`.

framework/mutex/OracleMutex.php
rob006: It is better to put this in constants like:
```php
const MODE_X = 'X_MODE';
const MODE_NL = 'NL_MODE';
...
```

framework/mutex/OracleMutex.php
zlakomanoff: If you about releaseOnCommit parameter - this parameter false by default
http://docs.oracle.com/cd/B19306_01/appdev.102/b14258/d_lock.htm#i1002556

klimov-paul: Should be:
```php
if (!in_array($this->lockMode, $this->availableLockModes, true)) {
```

framework/mutex/OracleMutex.php
zlakomanoff: fixed

framework/mutex/OracleMutex.php
zlakomanoff: fixed

framework/mutex/OracleMutex.php
zlakomanoff: fixed

framework/mutex/OracleMutex.php
zlakomanoff: done

klimov-paul:
Merged manually 9d5bf60a8927c1ccfd9c5c7ab02dd24166b318dc f1af3133c9488d73aaa28f922bc34eb61886d806

PR #11984, opened 2016-07-19
Japanese framework messages and docs updated [ci skip]

softark:
framework/messages/ja/yii.php updated
docs/guide-ja updated
docs/internals-ja updated


SilverFire:
Thank you!

Issue #11977, opened 2016-07-19
yii\rest\Serializer serialize ArrayDataProvider error


dcb9:
#### Creating a Controller
First, create a controller class app\controllers\UserController as follows,

```
namespace app\controllers;

use yii\data\ArrayDataProvider;
use yii\rest\Controller;

class UserController extends Controller
{
    public function actionIndex()
    {
        $users = [
            ['id' => 1, 'nickname' => 'Bob'],
            ['id' => 2, 'nickname' => 'Tom'],
        ];

        return new ArrayDataProvider([
            'allModels' => $users,
            'pagination' => [
                'pageSize' => 1,
            ]
        ]);
    }
}
```

#### Configuring URL Rules
```
'urlManager' => [
    'enablePrettyUrl' => true,
    'enableStrictParsing' => true,
    'showScriptName' => false,
    'rules' => [
        ['class' => 'yii\rest\UrlRule', 'controller' => 'user'],
    ],
]
```

#### Serve

`$ php yii serve`

in the other terminal:

`$ curl http://localhost:8080/users?page=2`

### What is the expected result?

```
[{"id":2,"nickname":"Tom"}]
```

### What do you get instead?

```
{"1":{"id":2,"nickname":"Tom"}}
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 5.5.34
| Operating system | Mac OS X

klimov-paul:
This is expected behavior.

dcb9:
And what do you want to do with this?

klimov-paul:
Nothing: you are requesting the list of models via REST (second page to be precise):

> $ curl http://localhost:8080/users?page=2

So you get the list of models in return (second page to be precise).

Everything is correct.

Forin:
The question is why "1:" is inserted when the ?page=2 parameter is added.
And why does the response-format differ:
?page=1:` [{"id":1,"nickname":"Bob"}]`
?page=2:` {"1":{"id":2,"nickname":"Tom"}}`

klimov-paul:
I see now.
It is caused by `ArrayDataProvider` pagination mechanism, which splits original array preserving its keys.

dcb9:
➜  ~ cat a.php
```
<?php

	$arr = [ ['id' => 1, 'nickname' => 'Bob'] ];

	echo json_encode($arr), PHP_EOL;
[...]

PR #11987, opened 2016-07-19
Fix yii\rest\Serializer serialize ArrayDataProvider with pagination bug

dcb9:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | yes/no |
| Tests pass? | yes |
| Fixed issues | #11977 |

```
➜  yii2 git:(fixRestSerializeArrayDataProvider) ./vendor/bin/phpunit --group=rest
PHPUnit 4.8.24-15-g3c4becb by Sebastian Bergmann and contributors.

....

Time: 895 ms, Memory: 46.50Mb

OK (4 tests, 39 assertions)
```

### DIFF
--- a/framework/rest/Serializer.php
+++ b/framework/rest/Serializer.php
@@ -12,6 +12,7 @@ use yii\base\Arrayable;
 use yii\base\Component;
 use yii\base\Model;
 use yii\data\DataProviderInterface;
+use yii\data\ArrayDataProvider;
 use yii\data\Pagination;
 use yii\helpers\ArrayHelper;
 use yii\web\Link;
@@ -171,7 +172,11 @@ class Serializer extends Component
      */
     protected function serializeDataProvider($dataProvider)
     {
-        $models = $this->serializeModels($dataProvider->getModels());
+        $models = $dataProvider->getModels();
+        if ($dataProvider instanceof ArrayDataProvider) {
+            $models = array_values($models);
+        }
+        $models = $this->serializeModels($models);
 
         if (($pagination = $dataProvider->getPagination()) !== false) {
             $this->addPaginationHeaders($pagination);
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.10 under development
 ------------------------
 
+- Bug #11977: Fixed `yii\rest\Serializer::serialize` serialize ArrayDataProvider with pagination error (dcb9)
 - Bug #11912: Fixed PostgreSQL Schema to support negative default values for integer/float/decimal columns (nsknewbie)
 - Bug #11947: Fixed `gridData` initialization in `yii.gridView.js` (pavlm)
 - Bug #11949: Fixed `ActiveField::end` generates close tag when it's `option['tag']` is null (egorio)
--- a/tests/framework/rest/SerializerTest.php
+++ b/tests/framework/rest/SerializerTest.php
@@ -1,0 +1,57 @@
+<?php
+
+namespace yiiunit\framework\rest;
+
+use Yii;
+use yii\data\ArrayDataProvider;
+use yiiunit\TestCase;
+
+/**
+ * @group rest
+ */
+class SerializerTest extends TestCase
+{
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->mockWebApplication();
+    }
+
+    /**
+     * @dataProvider serializeDataProvider
+     */
+    public function testSerializeDataProvider($allModels, $page, $expectModels)
+    {
+        $dataProvider = new ArrayDataProvider([
+            'allModels' => $allModels,
+            'pagination' => [
+                'route' => '/',
+            ],
+        ]);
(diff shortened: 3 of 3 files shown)


framework/rest/Serializer.php
```diff
@@ -171,7 +172,11 @@ protected function getRequestedFields()
      */
     protected function serializeDataProvider($dataProvider)
     {
-        $models = $this->serializeModels($dataProvider->getModels());
+        $models = $dataProvider->getModels();
+        if ($dataProvider instanceof ArrayDataProvider) {
```
AnatolyRugalev: Do we really need this check? I think this bug may appear in any kind of `DataProvider`, not only `ArrayDataProvider`

klimov-paul:
Resolved by commit cd8c01f555e425d4654775002ae5f44152e0c4b1

dcb9:
thanks @klimov-paul

LAV45:
@klimov-paul , `array_values()` removes the key which has been added using `->indexBy('tag')`

```php
class ManufacturerController extends \yii\rest\Controller
{
    public $serializer = [
        'class' => 'yii\rest\Serializer',
        'collectionEnvelope' => 'manufacturers',
    ];

    public function actionIndex()
    {
        $query = Manufacturer::find()
            ->indexBy('tag');

        return new ActiveDataProvider([
            'query' => $query,
        ]);
    }
}
```

klimov-paul:
> removes the key which has been added using ->indexBy('tag')

I can't see a problem here: data provider serialization is meant to return array of objects in JSON format, so it does it now.
If this behavior is not desiarable you should extend `Serializer` and implement `serializeDataProvider()` in your own way.

cebe:
still this change is a BC break.

klimov-paul:
> still this change is a BC break.

Why? Using of named keys is not something documented for the DataProvider serialization. Serializing of the list of rows expects production of the **list** of equal entities, not a data structure, where each row has its key and thus is logic meaning.

LAV45:
This API design was given to me as inheritance)
before change
```js
{
    code: 0,
    text: "ok",
    result: {
        manufacturers: {
            its_soft: {
                id: 1,
                name: '...',
                tag: 'its_soft'
            },
            tehnokom: {
                id: 2,
                name: '...',
                tag: 'tehnokom'
            }
        }
    }
}
```
after change
```js
{
    code: 0,
    text: "ok",
    result: {
        manufacturers: [
            {
                id: 1,
                name: '...',
                tag: 'its_soft'
            },
            {
                id: 2,
                name: '...',
                tag: 'tehnokom'
            }
        ]
    }
}
```

klimov-paul:
As I said this is not meant to be working in this way. If it should - the whole change should be reverted and this issue closed with resolution: 'not bug but feature'.

You can still return the list of models without data provider wrapped around them - in this case keys will be preserved.

AnatolyRugalev:
@LAV45 you may implement your own `Serializer` in this case. I'm agree with @klimov-paul, this is not documented feature (not a feature at all) and `DataProvider`  is not designed to provide structures except **list**.

Also I agree with @cebe, this fix breaks BC.

Original issue was about REST pagination problems of ArrayDataProvider, which may be resolved by changing 4th parameter of `array_slice` to `false` [here](https://github.com/yiisoft/yii2/blob/217d50e4ff690d651d910e61cb3601044031e185/framework/data/ArrayDataProvider.php#L91)

cebe:
@klimov-paul even though it was not documented, the current behavior with models that have non-integer keys was to return a JSON object indexed by the $key property of the dataprovider.
DataProvider itself was designed to work with keys, we even have separation of data and key values in it. It is not correct to assume the expectation of array in this case but a fix like proposed in #12683 should be appliced to make it configurable. It is not possible to autodetect whether indexed output is wanted or not, so a config option should be added.

klimov-paul:
> even though it was not documented, the current behavior with models that have non-integer keys was to return a JSON object indexed by the $key property of the dataprovider.

This behavior is inconsistent - this entire issue was about this.
While `ActiveDataProvider` iterates over pages with native keys per each page, `ArrayDataProvider` preserves orginal keys. Thus, in general case, for `ActiveDataProvider` each page will be converted to JSON array, while for `ArrayDataProvider` first page will be an array, while second and so on becomes object!

For me `DataProviver::getModels()` should return the list of records, e.g. array in JSON representation - not the data strucure - object in JSON representation.

> DataProvider itself was designed to work with keys, we even have separation of data and key values in it

It does work with keys but in different manner: there is a method `DataProvider::getKeys()`, which return  s keys for the data. `ActiveDataProvider` uses model primaky key for it, while `ArrayDataProvider` relies on native source array keys. That is why `ArrayDataProvider` can not strip them while splitting data into pages.

As I already said: in order to return a data-structure you can return array of models instead of `DataProvider`. Being indexed by cutom keys it will be serialized correctly.

> fix like proposed in #12683 should be appliced to make it configurable. It is not possible to autodetect whether indexed output is wanted or not, so a config option should be added.

Let it be. However such option should be disabled by default.

cebe:
fixed by  	04848a6

PR #11993, opened 2016-07-20
Activerecord where exists relation

urusaich:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | ? |

Allows to use relations in exists where.

``` php
$document->andWhere(['exists', 'comments']);
```

result:

``` sql
SELECT * FROM {{%document}} WHERE EXISTS (SELECT * FROM {{%comment}} WHERE {{%document}}.[[id]] = {{%comment}}.[[doc_id]])
```

``` php
$document->andWhere(['and', ['exists', 'comments'], ['exists', 'opens']]);
```

result:

``` sql
SELECT * FROM {{%document}} WHERE (EXISTS (SELECT * FROM {{%comment}} WHERE {{%document}}.[[id]] = {{%comment}}.[[doc_id]])) AND (EXISTS (SELECT * FROM {{%open}} WHERE {{%document}}.[[id]] = {{%open}}.[[doc_id]]))
```

``` php
$document->andWhere(['exists', 'comments' => function(ActiveQuery $q){
    $q->alias('alias');
}]);
```

result:

``` sql
SELECT * FROM {{%document}} WHERE EXISTS (SELECT * FROM {{%comment}} `alias` WHERE {{%document}}.[[id]] = {{alias}}.[[doc_id]])
```

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -159,6 +159,10 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             $this->select = ["$alias.*"];
         }
 
+        if(!empty($this->where)){
+            $this->where = $this->buildExistsRelations($this->where);
+        }
+
         if ($this->primaryModel === null) {
             // eager loading
             $query = Query::create($this);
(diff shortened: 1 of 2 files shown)


samdark:
@yiisoft/core-developers thoughts?

SilverFire:
Looks interesting and useful

framework/db/ActiveQuery.php
```diff
@@ -159,6 +159,10 @@ public function prepare($builder)
             $this->select = ["$alias.*"];
         }
 
+        if (!empty($this->where)) {
+            $this->where = $this->buildExistsRelations($this->where);
```
SilverFire: Is is really necessary to call `buildExistsRelations()` in each `prepare()`?

framework/db/ActiveQuery.php
```diff
@@ -728,6 +732,78 @@ public function orOnCondition($condition, $params = [])
...
+     */
+    private function buildExistsRelations($where)
+    {
+        if (!isset($where[0]) || !is_array($where)) {
+            return $where;
+        }
+
+        $operator = array_shift($where);
+        /** @var ActiveRecord $model */
+        $model = new $this->modelClass;
```
SilverFire: It is not OK to create new `ActiveRecord` object on each call even when method has nothing to do

nanodesu88: True. Whats about replace `new $this->modelClass` with `Yii::createObject($this->modelClass)`?

nanodesu88: True. Modified.

framework/db/ActiveQuery.php
```diff
@@ -728,6 +732,88 @@ public function orOnCondition($condition, $params = [])
     }
 
     /**
+     * Starts a transform relations in exists condition
+     */
+    private function buildExistsRelations()
```
cebe: this  should be implemented in QueryBuilder instead.

nanodesu88: Why? Is QueryBuilder knows something about Active db components?
`buildJoinWith`, `joinWithRelations` and `joinWithRelation` builds join from relations. 
`buildExistsRelations`, `existsRelations` and `existsRelation` does the same, but builds exists where instead.

samdark:
Won't make it into 2.0.

Issue #12000, opened 2016-07-21
Widget INIT event


user57376:
In order to enhance widgets with behaviors, we could simply use of the attach method.

```
class MyBehavior extends Behavior
{
   
    public function attach($owner)
    {
        parent::attach($owner);
        
        // do stuff
    }
    
}
```

Usually that's enough, but in this case I want to manipulate the configuration values too. 

For example:

```
echo Tabs::widget([
    'as remember' => ['class' => TabsRememberBehavior::className()],
    'items' => [
        ...
    ],
]);

```
The TabsRememberBehavior has to read the last tab position from a cookie and then activate that one in the items. But in this configuration order the behavior can't do his job because the items aren't set yet.

So, to be sure all configuration values are set I would like to have a Widget::EVENT_INIT.

unclead:
i think it is similar with what was discussed in scope of #8133

samdark:
@unclead yes but scope is narrower.

unclead:
If you decided to append this feature to the core it would be great, because i also think that such feature can be useful in some cases

cebe:
adding events to Widgets can help with making them more extensible but I would not limit this to EVENT_INIT. there should also be something like before and after render/run. not sure about which exactly yet.

samdark:
We can start with init and then proceed as requested.

klimov-paul:
Duplicates #7435

PR #12001, opened 2016-07-21
Added EVENT_INIT to widget

user57376:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
#12000

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -9,6 +9,7 @@ Yii Framework 2 Change Log
 - Bug #11949: Fixed `ActiveField::end` generates close tag when it's `option['tag']` is null (egorio)
 - Enh #11950: Improve BaseArrayHelper::keyExists speed (egorio)
 - Bug #11972: Fixed active form `afterValidate` wasn't triggered in some cases (lynicidn)
+- Enh #12000: Added EVENT_INIT to widget (user57376)
 
 2.0.9 July 11, 2016
 -------------------
--- a/framework/base/Widget.php
+++ b/framework/base/Widget.php
@@ -25,6 +25,11 @@ use ReflectionClass;
 class Widget extends Component implements ViewContextInterface
 {
     /**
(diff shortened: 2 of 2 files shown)


unclead:
i think it is similar with what was discussed in scope of #8133

SilverFire (approved):
Changelog line should be moved to the active milestone before merge

SilverFire:
Sorry, the PR seems to be overseen.
Waiting for more reviews to mergse

samdark (approved):
CHANGELOG should be modified so the line is in current release. Otherwise looks OK to me.

dynasource:
this is a useful addon. This would fix part 1 of https://github.com/yiisoft/yii2/issues/7435. Shouldnt we add another trigger after the run to fix part 2?

cebe:
> Shouldnt we add another trigger after the run to fix part 2?

yes, as far as I remember the main reason why this has not been merged is that it is incomplete. If we add INIT event, we should maybe also add others.

dynasource:
> yes, as far as I remember the main reason why this has not been merged is that it is incomplete. If we add INIT event, we should maybe also add others.

so what action is there to take

cebe:
look at how widget could be extended and decide which events to add.

samdark:
Can't we start with just init?

dynasource:
> Can't we start with just init?

IMO, we can, just as we can start with a closing trigger to fix https://github.com/yiisoft/yii2/issues/7435 which is a good example of a usecase. Having a starting & ending trigger enables wrapping by Events and extends the frameworks flexibility significantly.

samdark:
Closed as incomplete. Will work on solution with two events.

cebe:
See https://github.com/yiisoft/yii2/pull/13457#issuecomment-275772363

samdark:
Right. https://github.com/yiisoft/yii2/commit/7c1693479fbc8441d1565958942aeb7490872cb6

Merged. Thanks!

PR #12035, opened 2016-07-28
Fix scroll length

springimport:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | fix scroll length on main page |

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,5 +77,5 @@ If you are using Yii 2 as part of an OpenSource project, a way to acknowledge it
 If your code is hosted at GitHub, you can place the following in your README.md file to get the badge:
 
 ```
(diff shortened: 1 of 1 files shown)


springimport:
@samdark Go to the main repository and look readme.

Then, open the developer console and remove the slash after `http://www.yiiframework.com/`.
```
[![Yii2](https://img.shields.io/badge/Powered_by-Yii_Framework-green.svg?style=flat)](http://www.yiiframework.com/)
```

Scrolls will be lost. It looks better.

samdark:
Yes but that would mean extra redirect when anyone's hitting the server from such link.

springimport:
Yes, I see. Ok.

PR #12039, opened 2016-07-29
make console directory support sub-directory

922:
| Q | A |
| --- | --- |
| Is bugfix? | yes/no |
| New feature? | yes/no |
| Breaks BC? | yes/no |
| Tests pass? | yes/no |

### DIFF
--- a/framework/console/controllers/HelpController.php
+++ b/framework/console/controllers/HelpController.php
@@ -146,18 +146,35 @@ class HelpController extends Controller
         }
 
         $controllerPath = $module->getControllerPath();
-        if (is_dir($controllerPath)) {
-            $files = scandir($controllerPath);
-            foreach ($files as $file) {
-                if (!empty($file) && substr_compare($file, 'Controller.php', -14, 14) === 0) {
-                    $controllerClass = $module->controllerNamespace . '\\' . substr(basename($file), 0, -4);
-                    if ($this->validateControllerClass($controllerClass)) {
-                        $commands[] = $prefix . Inflector::camel2id(substr(basename($file), 0, -14));
-                    }
+        $commandsRes = $this->readDir($controllerPath,$module,$prefix);
+        if($commandsRes)$commands=array_merge($commands,$commandsRes);
+        return $commands;
+    }
+
+    /**
+     * @param string $file  the controllerPath
+     * @param $module \yii\base\Module $module the module instance
+     * @param string $prefix   namespace as prefix
+     * @return array the available command names
+     */
+    private function readDir($file,$module,$prefix){
+        $commands=[];
+        if(is_dir($file) ){
+            $files = scandir($file);
+            foreach($files as $v){
+                if( '.' === $v || '..'===$v)continue;
+                $commandsRes = $this->readDir($file.'/'.$v,$module,$prefix);
+                if($commandsRes)$commands=array_merge($commands,$commandsRes);
+            }
+        }else{
+            if (!empty($file) && substr_compare($file, 'Controller.php', -14, 14) === 0){
+                $controllerClass = $module->controllerNamespace . '\\' .substr(substr($file,strlen($module->getControllerPath())+1), 0, -4);
+                $controllerClass =  preg_replace('/\//','\\',$controllerClass);
+                if ($this->validateControllerClass($controllerClass)){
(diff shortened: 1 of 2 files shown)


922:
when we use yii command ,there is some problems with us . There has too many files in the  controllers directory .for make console directory support sub-directory, we update two files:
1、yii2\framework\console\controllers\HelpController.php
2、yii2\framework\helpers\BaseInflector.php

VisualRun:
Great work ！！！

AIfenfen:
建国，你真牛，大家都为你骄傲！！！

wu-wen-feng:
建国，棒棒的

chamuyaye:
建国，萌萌哒

KnightVan:
You are so Diao！

SilverFire:
Code style must be adjusted according to [Yii 2 Core Framework Code Style](https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md). Unit tests are required

cebe:
anyone willing to work on this further?

Ximich:
What happens with this pull request? It's really necessary feature.

framework/helpers/BaseInflector.php
```diff
@@ -383,7 +384,7 @@ public static function camel2id($name, $separator = '-', $strict = false)
         if ($separator === '_') {
             return trim(strtolower(preg_replace($regex, '_\0', $name)), '_');
         } else {
-            return trim(strtolower(str_replace('_', $separator, preg_replace($regex, $separator . '\0', $name))), $separator);
+            return trim(strtolower(preg_replace('/\/'.$separator.'/','/',str_replace('_', $separator, preg_replace($regex, $separator . '\0', $name)))), $separator);
```
cebe: this function should not be changed. all code necessary to support subfolders in console should be made in yii\console\* classes.

framework/console/controllers/HelpController.php
```diff
@@ -146,18 +146,35 @@ protected function getModuleCommands($module)
...
+                if( '.' === $v || '..'===$v)continue;
+                $commandsRes = $this->readDir($file.'/'.$v,$module,$prefix);
+                if($commandsRes)$commands=array_merge($commands,$commandsRes);
+            }
+        }else{
+            if (!empty($file) && substr_compare($file, 'Controller.php', -14, 14) === 0){
+                $controllerClass = $module->controllerNamespace . '\\' .substr(substr($file,strlen($module->getControllerPath())+1), 0, -4);
+                $controllerClass =  preg_replace('/\//','\\',$controllerClass);
+                if ($this->validateControllerClass($controllerClass)){
+                    $commands[] = $prefix . Inflector::camel2id(substr(substr($file,strlen($module->getControllerPath())+1), 0, -14));
```
cebe: this does not look complete to me, how does changing the help-controller enable subdirectories in the application?

cebe (changes requested):
@Ximich see my comments. Also a unit test would be great that verifies behavior.

framework/console/controllers/HelpController.php
Ximich: Subdirectories is already works in console application. But they doesn't appear in the list of all commands. This pull request fix that.

framework/helpers/BaseInflector.php
Ximich: Unfortunately there is no special place for only console helpers. Of course this can be done, but making such exception for only one console controller doesn't seem reasonable.

Ximich:
@cebe thanks for fast response. With surprise I found that subfolders are already working for the console application. But it's not documented anywhere and prompts for controllers from subfolders don't appear.

samdark:
So instead of changing inflector docs should be improved and help command adjusted, right?

framework/helpers/BaseInflector.php
cebe: handling of path elements does not belong to Inflector, if its needed it needs to be done outside. There is no need to create a helper from it, a private method in the class that is using it is fine.

cebe:
> So instead of changing inflector docs should be improved and help command should be adjusted, right?

yes, plus a test should verify behavior so we mark it as officially supported from then on.

samdark:
@Ximich do you have time to handle it?

developeruz:
It seems the problem is solved by #15249 and [tested](https://github.com/yiisoft/yii2/blob/f4c5c4931bb937f43ec8ce9338e0208406fdbe06/tests/framework/console/controllers/HelpControllerTest.php#L74)

Issue #11953, opened 2016-07-14
When download utf8 filename via  Response::sendfile()  , filename is not correct in IE11


pigochu:
Example Code

~~~php
return Yii::$app->getResponse()->sendFile($filepath, $attachment->filename);
~~~

if $attachment->filename is "中文.txt"  <== this is chinese filename with utf8 encoded.

When I download 
- Chome & Firefox save correct filename
- IE 11 is not correct

In http://php.net/manual/en/function.header.php , yjf_victor wrote :
![image](https://cloud.githubusercontent.com/assets/1013744/16832701/0de4414e-49de-11e6-89c4-ac25a1107317.png)

So I change my code looks like:

~~~php
$response = Yii::$app->getResponse()->sendFile($filepath, $attachment->filename);
$response->headers['Content-Disposition'] = 'attachment; '
        . sprintf('filename="%s"; ', rawurlencode($attachment->filename))
        . sprintf("filename*=utf-8''%s", rawurlencode($attachment->filename));
        return $response;
~~~

IE11 & Chrome & Filefox can download file with correct filename.

cebe:
I remember a similar issue report but I could not find it.

samdark:
https://github.com/yiisoft/yii2/issues/10563 ?

cebe:
Yes, thanks.

samdark:
Closed as duplicate of #10563.

PR #12040, opened 2016-07-29
fix #10563, #11953 current implementation of file download does not encode UTF8 filenames correctly

### LINKED ISSUE #10563: Current implementation of file download does not encode UTF8 filenames correctly
see https://github.com/yiisoft/yii/issues/2620#issuecomment-20663991

https://github.com/yiisoft/yii2/blob/8f1245553f1b3dcb5b4022780eb83a69f5065f33/framework/web/Response.php#L565

shirase:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues | #10563, #11953 |

### DIFF
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -438,6 +438,7 @@ class Response extends \yii\base\Response
         }
         if ($attachmentName === null) {
             $attachmentName = basename($filePath);
+            $this->headers->setDefault('Content-Disposition', "attachment; filename=\"$attachmentName\"");
         }
         $handle = fopen($filePath, 'rb');
         $this->sendStreamAsFile($handle, $attachmentName, $options);
@@ -563,7 +564,7 @@ class Response extends \yii\base\Response
             ->setDefault('Accept-Ranges', 'bytes')
             ->setDefault('Expires', '0')
             ->setDefault('Cache-Control', 'must-revalidate, post-check=0, pre-check=0')
-            ->setDefault('Content-Disposition', "$disposition; filename=\"$attachmentName\"");
+            ->setDefault('Content-Disposition', "$disposition; filename=\"$attachmentName\"; filename*=utf-8''" . rawurlencode($attachmentName));
 
         if ($mimeType !== null) {
             $headers->setDefault('Content-Type', $mimeType);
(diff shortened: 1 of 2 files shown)


PowerGamer1:
So now just because `mb_detect_encoding` is not installed or failed to detect encoding UTF-8 filename will be absent in HTTP header?

I would not use `mb_detect_encoding` at all and instead either:
a) require user to always pass `$attachmentName` in 'UTF-8'
or
b) allow to specify the encoding of `$attachmentName` as additional option in `Response::sendFile()` and then specify that encoding in generated HTTP header.

Also, it may be beneficial for compatibility with older browsers to also always include "filename=" (without asterisk , see the last example and explanation here: https://tools.ietf.org/html/rfc6266#section-5) and also not perform any encoding at all if every character in UFT-8 encoded `$attachmentName` is already US-ASCII.

shirase:
a) breaks backward compatibility, because current master works in all modern browsers except IE. 
b) i vote for it and make changes

PowerGamer1:
@shirase 
Sorry, but I don't think you really realize the extent of the problem with regards of what is broken and how to fix it. The following is already wrong (HTTP headers cannot have non US-ASCII characters in them):

```PHP
->setDefault('Content-Disposition', "$disposition; filename=\"$attachmentName\"" .
```

For that same reason when the `$attachmentName` is `null` you cannot just grab `basename($filePath)` disregarding the underlying encoding of `$filePath` (which is completely different on different OSes) and shove it directly into HTTP header.

shirase:
If $attachmentName is null, you must to set attachmentFileNameEncoding to your filesystem encoding, But it is better to set $attachmentName.

shirase:
IE supports only UTF-8, I should think.

shirase:
How about this one?

PowerGamer1:
> How about this one?

Exactly the same problem as before - invalid characters in HTTP header. See my version: https://github.com/yiisoft/yii2/pull/12062.

shirase:
$attachmentName must be ASCII only. If you need UTF-8 file name, use utf8Name option. Your solution can be used in 2.1.x.

shirase:
I made some changes. Not by RFC but it works in modern and old browsers and it have BC.

samdark:
Which old browsers have you tested?

samdark:
The solution technically works in majority of cases but isn't complete.

Issue #12025, opened 2016-07-27
FR, ENH: Database View Emulator (AR for query statement)


mdmunir:
It same as `ActiveRecord` except the source not from `table` (or existing view) but from query statement.

```php
class InvoicePaid extends QueryRecord
{
    public static function query()
    {
        return 'select invoice_id, sum(value) as value from payment group by invoice_id';
        // or return query object
    }
}

class Invoice extends ActiveRecord
{
    ...
    
    public function getPaid()
    {
        return $this->hasOne(InvoicePaid::class, ['invoice_id' => 'id']);
    }
}

// ***********
InvoicePaid::find()->where(['>', 'value', 100])->all();

foreach(Invoice::find()->joinWith('paid')->all() as $model) {
    echo $model->paid->value;
}
```

Why?
-------
Of course it can be more simple by create `view` for that statement then create AR for that `view`. 
But, i want in my application, my database is used only as storage. All rules (validation, integrity, relation, trigger, storage procedure and also view) should be done in aplication.

What do you think?

samdark:
The concept is interesting.

1. I assume it's read only. Correct?
2. Could it use eager loading somehow in case of `with`?

mdmunir:
> 1. I assume it's read only. Correct?
> 2. Could it use eager loading somehow in case of with?

1. Yes. Or, to make it writable, we should override method `updateAll()` and `insert()`. 
But, read only is enough.
2. I wish.

The problem is, how to resolve attributes? IMO, hardcode is more simple.

klimov-paul:
What is wrong with exending ActiveRecord overriding `find()` method to add extra query condition?

```php
class Payment extends \yii\db\ActiveRecord
{
     ...
}

class InvoicePaid extends Payment
{
    public $value;
 
    public static function find()
    {
        return parent::find()
            ->select(invoice_id, sum(value) as value)
            ->groupBy('invoice_id');
    }
}
```

Any relation, which uses `InvoicePaid` will use adjusted query.

mdmunir:
> What is wrong with exending ActiveRecord overriding find() method to add extra query condition?

Not sure.Need test.

mdmunir:
It work for me
```php
class InvoicePaid extends \yii\db\ActiveRecord
{

    public function attributes()
    {
        return[
            'invoice_id', 'paid'
        ];
    }

    public static function find()
    {
        return parent::find()
                ->select(['invoice_id', 'paid' => 'sum([[value]])'])
[...]

PR #12050, opened 2016-07-30
Database view emulator. ActiveRecord for query statement.

mdmunir:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12025 |

### DIFF
--- a/framework/db/QueryRecord.php
+++ b/framework/db/QueryRecord.php
@@ -1,0 +1,126 @@
+<?php
+
+namespace yii\db;
+
+use Yii;
+use yii\base\NotSupportedException;
+use yii\base\InvalidConfigException;
+use yii\helpers\Inflector;
+use yii\helpers\StringHelper;
+
+/**
+ * QuerRecord is the base class for classes representing query statement data in terms of objects.
+ *
+ * To declare an QueryRecord class you need to extend [[\yii\db\QueryRecord]] and
+ * implement the `query` method:
+ *
+ * ```php
+ * <?php
+ *
+ * class InvoicePaid extends \yii\db\QueryRecord
+ * {
+ *     public function attributes()
+ *     {
+ *         return ['invoice_id', 'total'];
+ *     }
+ *
+ *     public static function query()
+ *     {
+ *         return (new Query())
+ *             ->select(['invoice_id', 'total' => 'sum(value)'])
+ *             ->from('payment')
+ *             ->groupBy(['invoice_id']);
+ *     }
+ * }
+ * ```
+ *
+ * The `query` method only has to return the [[\yii\db\Query]] representation of the class.
+ *
+ * Below is an example showing some typical usage of QueryRecord:
+ *
+ * ```php
+ * $paids = InvoicePaid::find()
+ *              ->where(['>', 'total', 3426])
+ *              ->all();
+ * ```
+ *
+ * Use as relation
+ * ```php
+ * class Invoive extends ActiveRecord
+ * {
+ *     ...
+ *     public function getPaid()
+ *     {
+ *         return $this->hasOne(InvoicePaid::className(), ['invoice_id' => 'id']);
+ *     }
+ * }
+ * ```
+ *
+ * Then access it as
+ * ```php
+ * $invoices = Invoice::find()
+ *                 ->alias('i')
+ *                 ->joinWith(['paid p'])
+ *                 ->where('[[i.value]] > [[p.total]]')
+ *                 ->all();
+ *
+ * foreach (Invoice::find()->with('paid')->all() as $model) {
+ *     echo $model->paid->total;
+ * }
+ * ```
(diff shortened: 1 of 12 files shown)


framework/db/QueryRecord.php
```diff
@@ -0,0 +1,126 @@
...
+ *     echo $model->paid->total;
+ * }
+ * ```
+ * For more details and usage information on ActiveRecord, see the [guide article on ActiveRecord](guide:db-active-record).
+ *
+ * @method ActiveQuery hasMany($class, array $link) see [[BaseActiveRecord::hasMany()]] for more info
+ * @method ActiveQuery hasOne($class, array $link) see [[BaseActiveRecord::hasOne()]] for more info
+ *
+ * @author Misbahul D Munir <misbahuldmunir@gmail.com>
+ * @since 2.10
```
samdark: 2.0.10

framework/db/QueryRecord.php
```diff
@@ -0,0 +1,126 @@
...
+ */
+class QueryRecord extends BaseActiveRecord
```
samdark: Should it be abstract?

framework/db/QueryRecord.php
```diff
@@ -0,0 +1,126 @@
...
+     */
+    public static function getDb()
+    {
+        return Yii::$app->getDb();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public static function primaryKey()
```
samdark: Could be left w/o implementation.

tests/framework/db/QueryRecordTest.php
```diff
@@ -0,0 +1,52 @@
...
+use yiiunit\data\ar\ActiveRecord;
+use yiiunit\data\ar\QueryRecord;
+use yiiunit\data\ar\TotalOrder;
+use yiiunit\data\ar\Item;
+
+/**
+ * Description of QueryRecordTest
+ *
+ * @author Misbahul D Munir <misbahuldmunir@gmail.com>
+ * @since 1.0
```
samdark: ?

framework/db/QueryRecord.php
mdmunir: ya, agree.
but when create unit test, raise error `should not be abstract`.

tests/framework/db/QueryRecordTest.php
mdmunir: autogonerate from netbeans :D.

framework/db/QueryRecord.php
mdmunir: > PHP Fatal error:  Class yiiunit\data\ar\TotalOrder contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (yii\db\ActiveRecordInterface::primaryKey) in /home/mdmunir/projects/php/mdmunir/yii2/tests/data/ar/TotalOrder.php on line 25

framework/db/QueryRecord.php
mdmunir: disagree.
This class should can be implemented with minimal code. I mean, just implement `query()` and `attributes()`.

framework/db/QueryRecord.php
samdark: Well, yes. `TotalOrder` needs to implement all required methods.

samdark:
@klimov-paul what do you think about it?

klimov-paul:
> @klimov-paul what do you think about it?

This should not belong to the core.

As I already mentioned at th issue, there is no necessity of such thing as it can be easily achieved via default scopes feature:
https://github.com/yiisoft/yii2/issues/12025#issuecomment-235575887

Of course, there could be a situation, when such trick will not work, however I consider this to be a rare case.

I suggest this solution should be composed into an extension and maintained out of 'yiisoft' scope.

cebe:
I agree with @klimov-paul. @mdmunir feel free to publish it as an extension.

mdmunir:
As i said, this is View Emulator. I mean, everithing that can be do using view, can be done using this class. Some people prefer use `view` another people prefer using sql directly. If i have ability to create view in aplication, i chose create it in aplication.

@klimov-paul code not work for situation like this
```php
Invoice::find()
    ->alias('i')
    ->joinWith(['paid p'])
    ->where('[[i.value]] > [[p.total]]') // must use having instead of where
```
^ that code not a rare case (for me). That the reason why i need this class.

Simple way to create the same class without `ViewRecord` is directly inherit from `BaseActiveRecord`. But, if that so, i must override methods `getDb()`, `insert()`, `primaryKey()` many many times because that methods are abstract. 

> feel free to publish it as an extension.

I guess my own extension. Ya, i use this in my project (not as extension, but as a single class in my project). I submit it to core yii just to share my code to other yii user.

PR #12057, opened 2016-07-31
Added new token {attribute_name} in DetailView template

mohdqasim98:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues |  |

Added support for new token `{attribute_name}` in the template property of `yii\widgets\DetailView`.
The occurrence of  `{attribute_name}` in template property  will be replaced by name of the corresponding attribute when rendering.It can be used to add class to the DetailView row.
Eg. 

``` php
   $template = '<tr class="{attribute_name}"><th>{label}</th><td>{value}</td></tr>';
```

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,21 +4,21 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "d2005b487c5ff761d806b9a94bfe4cac",
-    "content-hash": "de99885237d7d9364d74fb5f93389801",
+    "hash": "b2a85a27f93f171bd5e078e087bef6cf",
+    "content-hash": "c4a9bb575b714fb7901c0c49f9012211",
     "packages": [
         {
             "name": "bower-asset/jquery",
-            "version": "2.2.3",
+            "version": "2.2.4",
             "source": {
                 "type": "git",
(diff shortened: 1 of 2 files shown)


mohdqasim98:
Hi Guys,
This is my first pull request,
Please correct me if I am wrong some whrere

unclead:
You can use an anonimous function as value of a template property for your needs

```php
'template' => function ($attribute, $index, $widget) {
    return '<tr class="' . $attribute['attribute']. '"><th>{label}</th><td>{value}</td></tr>';
}
```

don't you?

mohdqasim98:
@unclead 
Yes, I did that.
But wouldn't it be nice, if this functionality is added ?.
Beside that , can you please tell me why "-integration/travis-ci/pr" failed ?

samdark:
Since attribute name is rarely to be displayed to end user and it could be achieved as @unclead mentioned, I'm rejecting the pull request. Thanks for contribution attempt though. We appreciate it.

What failed in Travis could be found by clicking "Details". In this case it is HHVM and if you'll check output of it, it will be clear that it's not connected in any way with the code itself. So your code is fine.

mohdqasim98:
Thanks for the responses.
Will try some other thing to get involved in this project :)

PR #12059, opened 2016-08-01
Make simple url rule with default parameter

mdmunir:
Currently, to make url rule with default parameter, we must do this

``` php
'rules' => [
    ...
    [
        'pattern' => 'about',
        'route' => 'site/page',
        'defaults' => ['view' => 'about']
    ]
]
```

IMO, its more simple if we just do this

``` php
'rules' => [
    'about' => ['site/page', 'view' => 'about']
]
```

| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |


cebe:
I am against introducing another way of expressing the same thing. There are already two different syntaxes for Url rules. A third one just makes it even more complicated.

samdark:
Agree. Let's not introduce it.

Issue #12047, opened 2016-07-29
Response::sendFile Content-Disposition header with non-ASCII filenames


PowerGamer1:
When `$attachmentName` of `Response::sendFile` method contains non-ASCII characters the method creates invalid `Content-Disposition` HTTP header (according to specification HTTP headers cannot contain non-ASCII characters). In such a case the filename must be encoded according to https://tools.ietf.org/html/rfc5987 (also see https://tools.ietf.org/html/rfc6266#section-5).

Example:
```PHP
return \Yii::$app->getResponse()->sendFile('somepath', 'руссо туристо.txt');
```
The file "Save As..." window prompted by IE11 contains invalid file name.

shirase:
#12040

samdark:
Duplicate of #10563

PR #12062, opened 2016-08-01
Fix for non US-ASCII characters in Content-Disposition header

PowerGamer1:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | yes |
| Tests pass? | yes |
| Fixed issues | https://github.com/yiisoft/yii2/issues/12047, https://github.com/yiisoft/yii2/issues/10563, https://github.com/yiisoft/yii2/issues/11953 |

Fix for non US-ASCII characters in Content-Disposition header.

### DIFF
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -422,7 +422,8 @@ class Response extends \yii\base\Response
      * until [[send()]] is called explicitly or implicitly. The latter is done after you return from a controller action.
      *
      * @param string $filePath the path of the file to be sent.
-     * @param string $attachmentName the file name shown to the user. If null, it will be determined from `$filePath`.
+	 * @param string $attachmentName the file name shown to the user. Must be specified in UTF-8 encoding. If null,
+	 * the browser will try to determine the filename from the URL.
      * @param array $options additional options for sending the file. The following options are supported:
      *
      *  - `mimeType`: the MIME type of the content. If not set, it will be guessed based on `$filePath`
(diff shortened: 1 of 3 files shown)


PowerGamer1:
Fixing this without breaking BC would not be possible. The old code would put `basename($filePath)` into the HTTP header. Since `$filePath` can have arbitrary encoding (for ex., UTF-16 under Windows) it would produce the invalid HTTP header. Another problem of the old code - it would not allow to create the `Content-Disposition` header without specifying filename. Doing so would allow the browser to infer the filename from the URL. And the final problem fixed - now the UTF-8 `$attachmentName` would be properly encoded in the HTTP header.

shirase:
Many people use ascii only. Your solution have problem for them and it breaks BC.

PowerGamer1:
> Many people use ascii only. Your solution have problem for them 

No problems with ASCII at all.

> and it breaks BC.

If that is the price to fix broken HTTP headers then BC must be broken.

framework/web/Response.php
```diff
@@ -547,23 +547,29 @@ public function sendStreamAsFile($handle, $attachmentName, $options = [])
...
         $disposition = $inline ? 'inline' : 'attachment';
         $headers->setDefault('Pragma', 'public')
             ->setDefault('Accept-Ranges', 'bytes')
             ->setDefault('Expires', '0')
-            ->setDefault('Cache-Control', 'must-revalidate, post-check=0, pre-check=0')
-            ->setDefault('Content-Disposition', "$disposition; filename=\"$attachmentName\"");
+            ->setDefault('Cache-Control', 'must-revalidate, post-check=0, pre-check=0');
+
+		if($attachmentName !== null) {
+			$headers->setDefault('Content-Disposition', $disposition . '; filename*=UTF-8\'\'' . rawurlencode($attachmentName));
```
samdark: Older browsers aren't recognizing `filename*` at all so `filename` should be always provided.

samdark:
This solution won't work in older browsers.

PR #12066, opened 2016-08-02
fix mistake in dependency injection section of documentation

DumpOfTheVar:
Fix called method name in dependency injection docs


samdark:
What was wrong?

DumpOfTheVar:
@samdark In class MyClass method name is 'doSomething'. Invoke was called with method name 'doSomethingWithHeavyDependency'

cebe:
Thanks!

PR #12067, opened 2016-08-02
Added `yii\behaviors\AttributeTypecastBehavior`

klimov-paul:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11804 |

### DIFF
--- a/framework/behaviors/AttributeTypecastBehavior.php
+++ b/framework/behaviors/AttributeTypecastBehavior.php
@@ -9,18 +9,19 @@ namespace yii\behaviors;
 
 use yii\base\Behavior;
 use yii\base\InvalidParamException;
+use yii\base\Model;
 use yii\db\BaseActiveRecord;
 use yii\validators\BooleanValidator;
 use yii\validators\NumberValidator;
 use yii\validators\StringValidator;
 
 /**
- * AttributeTypecastBehavior provides an ability of automatic ActiveRecord attribute typecasting.
- * This behavior is very useful in case of usage of the schema-less databases like MongoDB or Redis.
- * It may also come in handy for regular [[\yii\db\ActiveRecord]], allowing to maintain strict
- * attribute types after model validation.
+ * AttributeTypecastBehavior provides an ability of automatic model attribute typecasting.
+ * This behavior is very useful in case of usage of ActiveRecord for the schema-less databases like MongoDB or Redis.
+ * It may also come in handy for regular [[\yii\db\ActiveRecord]] or even [[\yii\base\Model]], allowing to maintain
+ * strict attribute types after model validation.
  *
- * This behavior should be attached to [[\yii\db\BaseActiveRecord]] descendant.
+ * This behavior should be attached to [[\yii\base\Model]] or [[\yii\db\BaseActiveRecord]] descendant.
  *
  * You should specify exact attribute types via [[attributeTypes]].
  *
@@ -85,14 +86,15 @@ use yii\validators\StringValidator;
  * }
  * ```
  *
- * Attribute typecasting will be automatically performed at following cases:
+ * This behavior allows automatic attribute typecasting at following cases:
  *
  * - after successful model validation
  * - before model save (insert or update)
  * - after model find (found by query or refreshed)
  *
- * You may disable automatic typecasting for particular case using fields [[typecastAfterValidate]],
+ * You may control automatic typecasting for particular case using fields [[typecastAfterValidate]],
  * [[typecastBeforeSave]] and [[typecastAfterFind]].
+ * By default typecasting will be performed only after model validation.
  *
  * Note: you can manually trigger attribute typecasting anytime invoking [[typecastAttributes()]] method:
  *
@@ -103,7 +105,7 @@ use yii\validators\StringValidator;
  * $model->typecastAttributes();
  * ```
  *
- * @property BaseActiveRecord $owner the owner of this behavior
+ * @property Model|BaseActiveRecord $owner the owner of this behavior.
  *
  * @author Paul Klimov <klimov.paul@gmail.com>
  * @since 2.0.10
@@ -145,6 +147,7 @@ class AttributeTypecastBehavior extends Behavior
      * @var boolean whether to perform typecasting after owner model validation.
      * Note that typecasting will be performed only if validation was successful, e.g.
      * owner model has no errors.
+     * Note that changing this option value will have no effect after this behavior has been attached to the model.
      */
     public $typecastAfterValidate = true;
     /**
@@ -152,16 +155,18 @@ class AttributeTypecastBehavior extends Behavior
      * This option may be disabled in order to achieve better performance.
      * For example, in case of [[\yii\db\ActiveRecord]] usage, typecasting before save
      * will grant no benefit an thus can be disabled.
+     * Note that changing this option value will have no effect after this behavior has been attached to the model.
      */
-    public $typecastBeforeSave = true;
+    public $typecastBeforeSave = false;
     /**
      * @var boolean whether to perform typecasting after retrieving owner model data from
      * the database (after find or refresh).
      * This option may be disabled in order to achieve better performance.
      * For example, in case of [[\yii\db\ActiveRecord]] usage, typecasting after find
      * will grant no benefit in most cases an thus can be disabled.
+     * Note that changing this option value will have no effect after this behavior has been attached to the model.
      */
-    public $typecastAfterFind = true;
+    public $typecastAfterFind = false;
 
     /**
      * @var array internal static cache for auto detected [[attributeTypes]] values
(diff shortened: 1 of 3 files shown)


framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+
+use yii\base\Behavior;
+use yii\base\InvalidParamException;
+use yii\db\BaseActiveRecord;
+use yii\validators\BooleanValidator;
+use yii\validators\NumberValidator;
+use yii\validators\StringValidator;
+
+/**
+ * AttributeTypecastBehavior provides ability of automatic ActiveRecord attribute typecasting.
```
samdark: an ability

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+ * This behavior is very useful in case of usage schema-less databases like MongoDB or Redis.
```
samdark: is very useful in case of usage of schema-less databases

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+ * Tip: you may left [[attributeTypes]] blank - in this case its value will be detected
+ * automatically based on owner validation rules.
+ *
+ * Attribute typecasting will be automatically performed at following cases:
+ *
+ * - after successful model validation
+ * - before model save (insert or update)
+ * - after model find (found by query or refreshed)
+ *
+ * You may disable automatic typecasting at particular case using fields [[typecastAfterValidate]],
```
samdark: at particular case → for particular case

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+ * [[typecastBeforeSave]] and [[typecastAfterFind]].
+ *
+ * Note: you can manually trigger attribute typecasting at anytime invoking [[typecastAttributes]] method:
```
samdark: at anytime → anytime

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+     *     'amount' => 'integer',
+     *     'price' => 'float',
+     *     'is_active' => 'boolean',
+     *     'date' => function ($value) {
+     *         return ($value instanceof \DateTime) ? $value->getTimestamp(): (int)$value;
+     *     },
+     * ]
+     * ```
+     *
+     * If not set attribute type map will be composed automatically from the owner validation rules.
```
samdark: If not set**,**

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+     */
+    public $attributeTypes;
+    /**
+     * @var boolean whether to skip typecasting on `null` values.
+     * If enabled any attribute, which value equals to `null` will not be processed.
```
samdark: Sentence isn't clear. What was meant here?

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+     * For example, in case of [[\yii\db\ActiveRecord]] usage, typecasting after find
+     * will grant no benefit in most cases an thus can be disabled.
+     */
+    public $typecastAfterFind = true;
+
+    /**
+     * @var array internal static cache for auto detected [[attributeTypes]] values
+     * in format: ownerClassName => attributeTypes
+     */
+    private static $autoDetectedAttributeTypes = [];
```
samdark: I'm not sure making it static is a good idea. What if there are multiple models w/ such behaviors are used at the same time?

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+            if (!isset(self::$autoDetectedAttributeTypes[$ownerClass])) {
+                self::$autoDetectedAttributeTypes[$ownerClass] = $this->detectAttributeTypes();
+            }
+            $this->attributeTypes = self::$autoDetectedAttributeTypes[$ownerClass];
+        }
+    }
+
+    /**
+     * Typecast owner attributes according to [[attributeTypes]].
+     * @param null $attributeNames list of attribute names that should be type-casted.
```
samdark: Should type be `array`?

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,298 @@
...
+                case self::TYPE_BOOLEAN:
+                    return (boolean)$value;
+                case self::TYPE_STRING:
+                    return (string)$value;
+                default:
+                    throw new InvalidParamException("Unsupported type '{$type}'");
+            }
+        }
+
+        return call_user_func($type,  $value);
```
samdark: Remove extra space after `,`

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: This is exacty the case: autodecting of the attribute types take an extra time. If you selecting 100 AR instances it will run 100 times. Thus there is a cache for its result.
Cache is per owner class. Thus several different model classes will not share same value.

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: fixed

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: fixed

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: fixed

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: fixed

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: fixed

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: Description rephrased

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: fixed

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: fixed

framework/behaviors/AttributeTypecastBehavior.php
samdark: Seems I haven't noticed you have class as the key. It's fine.

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,329 @@
...
+     */
+    public $typecastBeforeSave = true;
+    /**
+     * @var boolean whether to perform typecasting after retrieving owner model data from
+     * the database (after find or refresh).
+     * This option may be disabled in order to achieve better performance.
+     * For example, in case of [[\yii\db\ActiveRecord]] usage, typecasting after find
+     * will grant no benefit in most cases an thus can be disabled.
+     */
+    public $typecastAfterFind = true;
```
thejahweh: Does this really make that much sense? This is done in the `yii\db\ColumnSchema::typecast` method.

framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -0,0 +1,329 @@
...
+                    return (int)$value;
+                case self::TYPE_FLOAT:
+                    return (float)$value;
+                case self::TYPE_BOOLEAN:
+                    return (boolean)$value;
+                case self::TYPE_STRING:
+                    return (string)$value;
+                default:
+                    throw new InvalidParamException("Unsupported type '{$type}'");
+            }
```
thejahweh: Would it be possible to support `DateTime` objects per default?
This is a bit tricky, because the format depends on the DB in wich it comes after find and if it comes from a form its based on a rule.

I would anyways recommend to give datetime fields from DBs more attention, because it is often used and always a pain.

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: It make sense for Redis

framework/behaviors/AttributeTypecastBehavior.php
klimov-paul: You should use a PHP callback for the object handling.

svit-ws:
This behavior must be disabled in nested models wich not perform save() operations.
For example, in data provider model if I put 'integer' rules for filter fields - they becomes 0 if empty

... or makes rules 'safe'

svit-ws:
and it not working for 'each' rule
```php [['array_field'], 'each', 'rule' => ['integer']]```

Issue #11992, opened 2016-07-20
yiiGridView applyFilter submit another form


izac1:
Have some problem whith  GridView in ActiveForm tag . I try do this in js but i have error 

> Uncaught RangeError: Maximum call stack size exceeded

my js code  in views 

`<?php
 
$this->registerJs(

$("#w0").on("beforeValidate", function (event, messages, deferreds) {
    $("#gdb").yiiGridView("applyFilter");
    return false;
});

    });'
);
?>`

if delete beforeValidate listner `$("#gdb").yiiGridView("applyFilter"); ` submit w0 form too.
I find this [solution](https://github.com/yiisoft/yii2/issues/5958#issuecomment-99103889 )
but its works only for pagination not for filter. 

my code  in views 

`    <?php $form = ActiveForm::begin(); ?>

    <?= $form->field($model, 'delivery_name')->textInput(['maxlength' => true]) ?>

    <?= $form->field($model, 'title')->textInput(['maxlength' => true]) ?>

    <?= $form->field($model, 'template_id')->dropDownList($templates) ?>

    <div class="form-group">
        <?= Html::submitButton('Create', ['class' =>'btn btn-success']) ?>
    </div>

	<?php yii\widgets\Pjax::begin(['id' => 'users','timeout' => false,'enablePushState' => false,]) ?>
    <?= GridView::widget([
        'dataProvider' => $dataProvider,
        'filterModel' => $searchModel,
        'id'=>"gdb",
        'columns' => [
            ['class' => 'yii\grid\SerialColumn'],
            ['class' => 'yii\grid\CheckboxColumn'],
           ......
        ],
    ]); ?>
    

        <?php yii\widgets\Pjax::end() ?>
    <?php ActiveForm::end(); ?>`

Alex-Code:
https://github.com/yiisoft/yii2/issues/8621

I had the same issue.
I intercepted the `submit` event so I could check the target.

```javascript
(function() {
    var form = $('#form-id').tabValidation()
        .off('submit.yiiActiveForm')
        .on('submit.yiiActiveForm', function(e) {
            if (this !== e.target) {
                return;
            }
            form.yiiActiveForm('submitForm', e);
        });
})();
```

izac1:
its works , but have another  bug , client validator show error 1 second and reload page

izac1:
i have dirty way like @Alex-Code  

`$("document").ready(function(){ 
		var form = $("#w0")
[...]

Issue #11972, opened 2016-07-18
yiiActiveForm beforeValidate and afterValidate incorrect triggering


githkim:
### What steps will reproduce the problem?
After fix issue #10681, event `beforeValidate` will called after blur on any field with enabled client validation. But `afterValidate` will not called. For example, this code does not work now:
```javascript
    // Form submition with enabling/disabling submit-button to prevent request multisend
    $(document).on('beforeValidate', 'form', function (event, messages, deferreds) {
        console.log('beforeValidate');
        $(this).find('[type=submit]').attr('disabled','disabled');
    }).on('afterValidate', 'form', function (event, messages, errorAttributes) {
        console.log('afterValidate');
        if (errorAttributes.length) {
            alert('Check form errors!');
            $(this).find('[type=submit]').removeAttr('disabled');
        }
    }).on('beforeSubmit', 'form', function (event) {
        console.log('beforeSubmit');
        $(this).find('[type=submit]').attr('disabled','disabled');
    });
```
In result, button will disabled after blur on first field with clientValidationRule.

| Q                | A
| ---------------- | ---
| Yii version      |  2.0.9
| PHP version      |  5.5.30
| Operating system | Windows 7, Ubuntu 14.04

lynicidn:
https://github.com/yiisoft/yii2/pull/11994

check this code

githkim:
@lynicidn thanks! I've checked this pull request. It works on right way!

samdark:
PR merged.

gmalpeli:
Sorry my english. Problems whit ajax css validate:  In a form, when a field is filled with a valid value, the div style of this field , change to "form-group..... has-success",  With this change, all fields in the form, change the div style, to "form-group..... has-success",  So all are green...

pstddc:
@gmalpeli it's not related issue?

это же не связано с этим багом? дивы никто не трогал, меняли клиентскую валидацию в яваскрипте

SilverFire:
The issue is reopened since the accepted PR caused wrong behavior

SilverFire:
See #12069

SilverFire:
Fixed by https://github.com/yiisoft/yii2/commit/418b27326c8753e1224080a958930550ff4d11ac

PR #12069, opened 2016-08-03
Reverted fix of beforeValidate event calling in `yii.activeForm.js`

SilverFire:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | yes (fixes) |
| Fixed issues | #10681, #11972, probably #11992 |

I'm reverting @lynicidn commit e8bcb93 because it introduces buggy behavior. The problem that was reported in #10681 is not a bug indeed.
1. From the docs: 
   > beforeValidate event is triggered before validating the whole form
2. AJAX validation is being executed each time for all available inputs, but the result of AJAX validation will be used for each particular input only when it was touched at least once. It means, that each AJAX validation call does not mean full form validation.
3. For Ajax validation we have `ajaxBeforeSend` and `ajaxComplete` events

@nkovacs, @samdark, @githkim, @lynicidn - if you have something to add - feel free to text your ideas. Otherwise, the PR will be merged in 3 days.

Thank you

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -14,6 +14,7 @@ Yii Framework 2 Change Log
 - Bug #12037: Fixed 2.0.7 regression in memcahe/memcached cache backend (samdark)
 - Bug #12043: Fixed `yii\helpers\Json::encode()` encodes empty array returned by `JsonSerializable::jsonSerialize()` as object (klimov-paul)
 - Bug #12045: Added missing `LEVEL_PROFILE` to `yii\log\Logger::getLevelName()` map (Mak-Di)
+- Bug #10681: Reverted fix of beforeValidate event calling in `yii.activeForm.js` (silverfire)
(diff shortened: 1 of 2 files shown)


nkovacs:
I'm confused. That commit was made by @lynicidn

samdark:
Either we should add comments in the JS file itself or document it. It raises too many questions.

nkovacs:
It's documented in the js file: https://github.com/yiisoft/yii2/blob/e8bcb93507a7859f2a5115ec2179182353e9cc04/framework/assets/yii.activeForm.js#L27
(well, sort of, it says "before validating the whole form", so it might not be clear that means before submitting the form)
Unfortunately the documentation in the js files is not picked up by the documentation generator.

samdark:
Yes. That's why we need to document it as a guide topic. There aren't many files so should be maintainable.

PR #12072, opened 2016-08-03
Fixed #11541: Simply changed numbers for MySQL tables PK generation

### LINKED ISSUE #11541: MySQL integer pk overflow vector
`yii\db\mysql\QueryBuilder`

```php
public $typeMap = [
    Schema::TYPE_PK => 'int(11) NOT NULL AUTO_INCREMENT PRIMARY KEY',
    Schema::TYPE_UPK => 'int(11) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY',
];
```

Max value of `int` column is 4,294,967,295 (unsigned) (http://dev.mysql.com/doc/refman/5.7/en/integer-types.html).

That's only 10 digits, so the type map should be changed to `int(10)`.

h311ion:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11541 |

### DIFF
--- a/framework/db/mysql/QueryBuilder.php
+++ b/framework/db/mysql/QueryBuilder.php
@@ -23,8 +23,8 @@ class QueryBuilder extends \yii\db\QueryBuilder
      * @var array mapping from abstract column types (keys) to physical column types (values).
      */
     public $typeMap = [
-        Schema::TYPE_PK => 'int(11) NOT NULL AUTO_INCREMENT PRIMARY KEY',
-        Schema::TYPE_UPK => 'int(11) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY',
+        Schema::TYPE_PK => 'int(10) NOT NULL AUTO_INCREMENT PRIMARY KEY',
+        Schema::TYPE_UPK => 'int(10) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY',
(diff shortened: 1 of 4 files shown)


framework/db/mysql/QueryBuilder.php
```diff
@@ -23,8 +23,8 @@ class QueryBuilder extends \yii\db\QueryBuilder
      * @var array mapping from abstract column types (keys) to physical column types (values).
      */
     public $typeMap = [
-        Schema::TYPE_PK => 'int(11) NOT NULL AUTO_INCREMENT PRIMARY KEY',
-        Schema::TYPE_UPK => 'int(11) UNSIGNED NOT NULL AUTO_INCREMENT PRIMARY KEY',
+        Schema::TYPE_PK => 'int(10) NOT NULL AUTO_INCREMENT PRIMARY KEY',
```
rob006: This is incorrect.

samdark: Why?

bizley: In case of signed int doesn't minus take one space? So it's eleven in total for -2147483648.

cebe:
the numbers are correct as they are the default value set by mysql if no number is specified.

Issue #12073, opened 2016-08-03
ActiveField::hint() must behave analogous to ActiveField::label()


PowerGamer1:
Current implementation of `ActiveField::hint()` does not allow to suppress the generation of the hint for the input in the form if the hint is defined through `Model::attributeHints()`. The very similar method pair `ActiveField::label()` and `Model::attributeLabels()` provides such possibility.

PR #12074, opened 2016-08-03
Patch 10

PowerGamer1:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | yes |
| Fixed issues | https://github.com/yiisoft/yii2/issues/12073 |

### DIFF
--- a/framework/widgets/ActiveField.php
+++ b/framework/widgets/ActiveField.php
@@ -199,7 +199,7 @@ class ActiveField extends Component
                 $this->error();
             }
             if (!isset($this->parts['{hint}'])) {
-                $this->hint(null);
+                $this->hint();
             }
             $content = strtr($this->template, $this->parts);
         } elseif (!is_string($content)) {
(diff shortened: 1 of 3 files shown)


framework/widgets/ActiveField.php
```diff
@@ -315,10 +317,17 @@ public function error($options = [])
      *
      * @return $this the field object itself
      */
-    public function hint($content, $options = [])
+    public function hint($content = null, $options = [])
```
samdark: In order for it to be merged in 2.0 it should not change public method signature.

PowerGamer1: What is the problem with this particular change? It brings the method signature in line with `label()` method and does not break BC in any way.

rob006: @PowerGamer1 It will break any class that override `hint()` method using old signature: https://3v4l.org/IoSqL

PowerGamer1: Alright, that would break BC indeed. I added upgrade instructions to UPGRADE.md.

rob006: But why you change this signature? There is no big benefits.

PowerGamer1: @rob006 
As I said, for consistency with `public function label($label = null, $options = [])`. IMO you either do something properly or don't do it at all. Besides, affected code upgrade for this change is super easy.

samdark: I think I'll merge it w/o signature change in 2.0 and w/ signature change in 2.1.

tunecino:
@PowerGamer1 please check the [Yii Versioning Policy](https://github.com/yiisoft/yii2/blob/master/docs/internals/versions.md) section. someone somewhere will upgrade a production app to 2.0.10 and his online code may stop working properly. Yii should guarantee that no changes is required for minor releases upgrades.

samdark:
Merged differently to both 2.0 and 2.1. Thank you for contributing!

PR #12089, opened 2016-08-05
Fixing issue with filterInputValidation variable in DataColumn

ghost:
| Q | A |
| --- | --- |
| Is bugfix? | Yes |
| New feature? | No |
| Breaks BC? | No |
| Tests pass? | Yes |

Currently when the $filterInputOptions in yii\grid\DataColumn is overruled with only a custom prompt in the new array, it will loose its bootstrap class and make it look like a default select box.

This PR fixes to problem, that way when only the prompt is changed, the class will stay it is default, and does need to be specific in the custom array.

### DIFF
--- a/framework/grid/DataColumn.php
+++ b/framework/grid/DataColumn.php
@@ -106,7 +106,7 @@ class DataColumn extends Column
      * render the HTML attributes for the generated filter input fields.
      * @see \yii\helpers\Html::renderTagAttributes() for details on how attributes are being rendered.
      */
-    public $filterInputOptions = ['class' => 'form-control', 'id' => null];
+    public $filterInputOptions = [];
 
 
     /**
@@ -177,6 +177,9 @@ class DataColumn extends Column
 
         $model = $this->grid->filterModel;
 
+        // The default options the filterCell has
+        $defaultFilterOptions = ['class' => 'form-control', 'id' => null];
+
         if ($this->filter !== false && $model instanceof Model && $this->attribute !== null && $model->isAttributeActive($this->attribute)) {
             if ($model->hasErrors($this->attribute)) {
                 Html::addCssClass($this->filterOptions, 'has-error');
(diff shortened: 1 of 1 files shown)


cebe:
as far as I see, with these changes there is no way to remove the default bootstrap class...

bvanleeuwen:
I did need to do that infact at one point, and I just put in 'class' => '', and it went back to having no class set. This does leave a empty class element behind.

bvanleeuwen:
Almost a week has passed since the last activity in this PR, if you have any improvements that can be made, please tell me, but I have just re-tested your reply, and I can remove the default Bootstrap class with no problems.

cebe:
Would be nice to cover this with a unit test. Would you have time to write one?

bvanleeuwen1995:
If you can tell me what you want the unit test to do specifically I can write a test for you, no problem.

samdark:
As you described in the 1st post: "when only the prompt is changed, the class will stay it is default".

bvanleeuwen1995:
Thank you, I have a idea which tests I will create to validate this PR, I don't have that much experience with unit testing, so it might take a little longer then usual.

framework/grid/DataColumn.php
```diff
@@ -177,6 +177,9 @@ protected function renderFilterCellContent()
 
         $model = $this->grid->filterModel;
 
+        // The default options the filterCell has
+        $defaultFilterOptions = ['class' => 'form-control', 'id' => null];
+
```
dynasource: * this $variable can be moved closer to the please where it is used
* it could also directly be merged with $filterInputOptions
* the prompt can be merged below

dynasource (changes requested):
please enter the wonderful world of unit testing. Once you know how it works, you will be surprised by it's power.

cebe:
This change may break BC when JS code relies on the fact that the ID is generated when a custom filter intput option is supplied.

cebe:
merged into 2.1 with the adjustments requested by @dynasource and a unit test. Thank you!

Issue #12009, opened 2016-07-24
Yii2: RadioList and CheckboxList - The for attribute of the label element must refer to a non-hidden form control.


adampblack:
### What steps will reproduce the problem?
$form->field($model, 'attribute')->radioList($model->arrAttributeOptions)
$form->field($model, 'attribute2')->checkboxList($model->arrAttribute2Options)

### What is the expected result?
For example, the checkbox list active field produces this (note label and hidden input):
```
<div class="form-group field-modelname-attributename">
   <label class="control-label" for="modelname-attributename">Attribute Label</label>
   <input type="hidden" name="Modelname[attributename]" value="">
   <div id="modelname-attributename">
       <div class="radio"><label><input type="radio" name="Modelname[attributename]" value="attributevalue1" checked> attributevaluelabel1</label></div>
       <div class="radio"><label><input type="radio" name="Modelname[attributename]" value="attributevalue2"> attributevaluelabel2</label></div>
       <div class="radio"><label><input type="radio" name="Modelname[attributename]" value="attributevalue3"> attributevaluelabel3</label></div>
   </div>
</div>
```

### What do you get instead?
Go to a w3C validator: "Error: The for attribute of the label element must refer to a non-hidden form control."

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | yii2.05
| PHP version      | PHP 5.5.9
| Operating system | Ubuntu 14.04.4 LTS

shevchik87:
I tried and repeated problem
Can I solve this problem?

cebe:
If you have a way to solve it, you can send a pull request :)

nagyt234:
This issue still exists.

PR #12106, opened 2016-08-08
12009 label for

shevchik87:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | yes |
| Tests pass? | no |
| Fixed issues | #12009 |

RadioList and CheckboxList - The for attribute of the label element must refer to a non-hidden form control.

This is my first commit not judge strictly

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.10 under development
 ------------------------
-
+- Bug #12009: RadioList and CheckboxList delete attribute 'for' in label (shevchik87)
 - Bug #12068: Added missing 'LEVEL_PROFILE' for the syslog target (Mak-Di)
 - Bug #11461: Fixed migration tool error when create migrate with comma in defaultValue (pana1990, s-o-f)
 - Bug #11912: Fixed PostgreSQL Schema to support negative default values for integer/float/decimal columns (nsknewbie)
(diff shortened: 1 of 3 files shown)


samdark:
This pull request is backwards compatibility breaking. Can't be applied to 2.0.

shevchik87:
what i have to do?

samdark:
You can either change method signature to be compatible or wait till 2.1 when we can technically break method signatures.

shevchik87:
I am sorry but 
I did not get anything

samdark:
BC = backwards compatibility break i.e. inability to just update via Composer w/o modifying existing application code. If method signature (number or order or parameters in a method) changes, it breaks compatibility with previous versions because of inheritance i.e.:

```php
class X
{
    public function hello($name)
    {
        return 'Hello, ' . $name;
    }
}

// application defines Y:

class Y extends X
{
    public function hello($name)
    {
        return 'Welcome,' . $name;
    }
}
```

If we'll change `X::hello()` method to be

```php
public function hello($name, $word = 'Hello, ')
{
    return $word . $name;
}
```

then Y will fail with PHP error.

shevchik87:
Thak you 
Now I all understand

samdark:
Looks like wrong solution to me.

samdark:
Merged with adjustments: https://github.com/yiisoft/yii2/commit/b8c07f6839ef5facf67ff1482e3189abf1145978

Issue #12099, opened 2016-08-07
HttpCache triggers 304 HTTP code when callbacks return null


sergeymakinen:
### What steps will reproduce the problem?

Use `HttpCache` filter with `$lastModified` and/or `$etagSeed` defined and return null.

### What is the expected result?

`HttpCache` won't do anything.

### What do you get instead?

`HttpCache` will trigger `304 Not Modified`. 

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.9
| PHP version      | 5.6.23
| Operating system | Debian

When `Last Modified` and/or `ETag` both can't be determined (but callbacks exist) `HttpCache` should just `return true` in `beforeAction`. It allows to do some complex logic in an action leaving the behavior simple (without need to enable/disable the behavior in a controller's `beforeAction`).

Kolyunya:
@sergeymakinen according to the docs you are triggering an undefined behavior since those callbacks must return UNIX timestamp, not `null`.

> The callback should return a UNIX timestamp.

sergeymakinen:
@Kolyunya Sure, but I guess it's not a normal behavior in this case. It should throw `UnexpectedValueException` or treat it like "no data" - and when we have no data, we don't trigger `Not Modified` at least.

cebe:
This already works with `lastModified` but not with etag. Want to send a pull request to fix it?

sergeymakinen:
@cebe Basically it should `return false` on the [line 171](https://github.com/yiisoft/yii2/blob/master/framework/filters/HttpCache.php#L171). Yes I'd like to try. May be it's a good idea to follow 4.1 of [RFC 7232](https://tools.ietf.org/html/rfc7232#section-4.1) in `validateCache` & `beforeAction` (like RFC tells to send `Last-Modified` on 304 if `Etag` is not available).

cebe:
[generateEtag()](https://github.com/yiisoft/yii2/blob/552ad141b66d63e1fbe245941e702ff7f5a1ec78/framework/filters/HttpCache.php#L204) should return null if the seed is null.

And yes, the httpcache class should correspond to RFC defined behavior.

sergeymakinen:
Yes, I see.

PR #12115, opened 2016-08-08
HttpCache no longer returns 304 HTTP code when callbacks return null (#12099)

sergeymakinen:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12099 |

As a bonus, it follows [4.1 of RFC 7232](https://tools.ietf.org/html/rfc7232#section-4.1) more strict.


samdark:
Looks OK to me. @cebe ?

cebe:
Merged, thank you!

PR #12117, opened 2016-08-09
Check if callable

Mak-Di:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

Verify that the contents of a variable $this->prefix can be called as a function.

### DIFF
--- a/framework/log/Target.php
+++ b/framework/log/Target.php
@@ -276,7 +276,7 @@ abstract class Target extends Component
      */
     public function getMessagePrefix($message)
     {
-        if ($this->prefix !== null) {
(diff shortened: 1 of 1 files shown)


samdark:
What's the use case for it?

Mak-Di:
If accidentally specify not a function, the application will be fail. I think it could be better if the application to continue working with the default prefix than fail.

samdark:
I think it's better to fail in this case. It's a configuration issue and it should be revealed as soon as possible.

PR #12176, opened 2016-08-16
Fix error in docs

DumpOfTheVar:
I found error in formating in guide http://www.yiiframework.com/doc-2.0/guide-runtime-handling-errors.html#using-error-actions. In note "If you need to redirect in an error handler, do ..." code block displays wrong. In github this block displays correctly https://github.com/yiisoft/yii2/blob/master/docs/guide/runtime-handling-errors.md#using-error-actions-.

I found similar block in http://www.yiiframework.com/doc-2.0/guide-runtime-sessions-cookies.html#flash-data which displays correctly and add one line before code block by analogy. I hope this will fix the error, hoverer I am not sure.

### DIFF
--- a/docs/guide/runtime-handling-errors.md
+++ b/docs/guide/runtime-handling-errors.md
@@ -148,6 +148,7 @@ the following variables if the error action is defined as [[yii\web\ErrorAction]
 the error action and the error view are already defined for you.
 
 > Note: If you need to redirect in an error handler, do it the following way:
+>
 > ```php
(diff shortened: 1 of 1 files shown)


samdark:
Cherry-picked first commit https://github.com/yiisoft/yii2/commit/8290f9ac9a9bbe69b03348fcb806c00b5e286d17. Thanks!

Second one isn't necessary. It's OK as is.

Issue #2365, opened 2014-02-08
Debug Panel should have information about cache


dilip-vishwa:
If the cache is used in page, then we should get some informative information about the cache component used in particular view, query, fragment, or page, etc. It should also show the content of the particular cache.

kidol:
Anyone in favor for the event based approach? In yii 1 one had to parse though the log to figure out cache hits/misses. With events, the debug toolbar can listen for class-level events I think:

```php
Event::on(Cache::className(), Cache::EVENT_SET, function ($event) {
    // Collect all results for later
    // $event->sender can be used to figure out cache backend
    // $event might provide: $cacheKey, $success (whether get, set, etc. succeeded)
});
```

Normal logging like in yii 1 should still be added I guess?

Also a simplified version of the idea
> expiration settings and it's duration till now

is possible if cache does store expire timestamp within the cache data (serialize, like with dependency). Then we could see when cache item expires. It could be an optional feature, disabled by default.

PR #12181, opened 2016-08-17
Added PageCache info of in log 

githubjeka:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
### not found:

image
### found:

image

### DIFF
--- a/framework/filters/PageCache.php
+++ b/framework/filters/PageCache.php
@@ -164,10 +164,7 @@ class PageCache extends ActionFilter
         ob_implicit_flush(false);
         if ($this->view->beginCache($id, $properties)) {
             $response->on(Response::EVENT_AFTER_SEND, [$this, 'cacheResponse']);
-            Yii::info(
-                'Valid page content is not found in the cache.' . print_r($properties,true),
-                __METHOD__
-            );
+            Yii::trace('Valid page content is not found in the cache.', __METHOD__);
             return true;
         } else {
             $data = $this->cache->get($this->calculateCacheKey());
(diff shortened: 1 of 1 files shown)


cebe:
shouldn't the log level be trace instead of info?

SilverFire:
I don't think that `print_r` is suitable in application logs at all

githubjeka:
May be ArrayHelper:toArray ?

SilverFire:
It is enough to trace whether the valid cache was found

githubjeka:
IMHO  'Valid page content is not found in the cache.' is worthless without details.

SilverFire:
What information is the most valuable in this `print_r`?

samdark:
Looks fine to me. @SilverFire how about merging it into 2.0.10?

SilverFire:
I'm fine with it. @githubjeka could you add a CHANGELOG line please?

kidol:
I don't like this. Logging should be implemented in cache component, which would make this PR obsolete.

SilverFire:
@kidol do you suggest to trace all `cache->get()` and `cache->set()` calls?

kidol:
Yes. Here you can see how it was done in Yii1: https://github.com/yiisoft/yii/blob/master/framework/caching/CCache.php#L102

Also DebugToolbar could make use of these logs to show cache hits/misses (#2365).

Though, I'm not interested in working on this in any way. So, see it as my 2 cents on this topic...

SilverFire:
I don't know the original reasons why this trace has been removed, by I guess because of performance reasons. @cebe @samdark

samdark:
Could be added. Trace is rarely used in production.

cebe:
guess it was performance, not sure.

SilverFire:
Merges as is, thank you

githubjeka:
@kidol if you want you could send your PR for all cache by deleting this code. 

Thanks.

PR #12187, opened 2016-08-18
Added BaseYii test for the setLogger() method

Mak-Di:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

### DIFF
--- a/tests/framework/BaseYiiTest.php
+++ b/tests/framework/BaseYiiTest.php
@@ -91,7 +91,7 @@ class BaseYiiTest extends TestCase
      * @covers yii\BaseYii::setLogger()
      * @covers yii\BaseYii::getLogger()
      */
-    public function testSetAndGetLoggerTogether()
+    public function testSetupLogger()
     {
         $logger = new Logger();
         BaseYii::setLogger($logger);
--- a/tests/framework/BaseYiiTest.php
+++ b/tests/framework/BaseYiiTest.php
@@ -89,16 +89,13 @@ class BaseYiiTest extends TestCase
 
     /**
      * @covers yii\BaseYii::setLogger()
+     * @covers yii\BaseYii::getLogger()
      */
-    public function testSetLogger()
+    public function testSetAndGetLoggerTogether()
     {
         $logger = new Logger();
         BaseYii::setLogger($logger);
 
-        $reflection = new \ReflectionClass('yii\BaseYii');
-        $log = $reflection->getProperty('_logger');
-        $log->setAccessible(true);
-
-        $this->assertEquals($log->getValue('yii\BaseYii'), $logger);
+        $this->assertSame($logger, BaseYii::getLogger());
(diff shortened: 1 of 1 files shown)


tests/framework/BaseYiiTest.php
```diff
@@ -84,4 +86,19 @@ public function testCreateObjectCallable()
...
+
+    /**
+     * @covers yii\BaseYii::setLogger()
+     */
+    public function testSetLogger()
+    {
+        $logger = new Logger();
+        BaseYii::setLogger($logger);
+
+        $reflection = new \ReflectionClass('yii\BaseYii');
```
klimov-paul: ???

What is wrong about:

```php
$this->assertSame($logger, BaseYii::getLogger());
```
?

Mak-Di: The test will pass, but it will depend on the getLogger method. If the method getLogger change the logic, the test with name "setLogger" will fail.
Is it okay?

klimov-paul: Now your test depends on the implementation of the `setLogger()` method, which is contradicts the unit test paradigma.
If the internal implemenation of the method changes - the test will fail.

tests/framework/BaseYiiTest.php
```diff
@@ -89,16 +89,13 @@ public function testCreateObjectCallable()
 
     /**
      * @covers yii\BaseYii::setLogger()
+     * @covers yii\BaseYii::getLogger()
      */
-    public function testSetLogger()
+    public function testSetAndGetLoggerTogether()
```
klimov-paul: Better name will be `testSetupLogger`

Mak-Di: Do not review, please. I would like to add one more case.

klimov-paul:
Thank you.

tests/framework/BaseYiiTest.php
```diff
@@ -84,4 +86,16 @@ public function testCreateObjectCallable()
             return true;
         }));
     }
+
+    /**
+     * @covers yii\BaseYii::setLogger()
+     * @covers yii\BaseYii::getLogger()
+     */
+    public function testSetupLogger()
+    {
```
Mak-Di: @klimov-paul, that do you think about this assertion:
$this->assertInstanceOf('yii\log\Logger', BaseYii::getLogger());

for this part:
```php
// yii2/framework/BaseYii.php:363
} else {
    return self::$_logger = static::createObject('yii\log\Logger');
}
```

klimov-paul: 3e3c4b12ef8c422452a0c0803523c1d5979809cc

Mak-Di: Thank you.

PR #12198, opened 2016-08-18
Add short validator names `time` and `datetime`.

nkovacs:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes, but highly unlikely in practice (see below) |
| Tests pass? | yes |
| Fixed issues |  |

These map to `DateValidator` with `$type` set to
TYPE_TIME and TYPE_DATETIME.

Looking at [Validator::createValidator](https://github.com/yiisoft/yii2/blob/2.0.9/framework/validators/Validator.php#L190), if the model has a method called `time` or `datetime`, this won't override it, but if there's a class called `time` or `datetime`, it will override the class. This is extremely unlikely, since classes are usually uppercased and namespaced.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -30,6 +30,7 @@ Yii Framework 2 Change Log
 - Enh #12082: Used `jQuery.on(` instead of event method to ensure forwards compatibility (newerton)
 - Enh #12028: Add -h|--help option to console command to display help information (pana1990)
 - Enh #12099: HttpCache no longer returns 304 HTTP code when callbacks return null (sergeymakinen)
+- Enh #12198: Added `time` and `datetime` validator short names (nkovacs)
 - Bug #12053: `./yii migrate/create` was generating wrong code when using `bigPrimaryKey` (VojtechH, samdark)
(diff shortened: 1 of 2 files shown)


SilverFire:
Looks useful to me

cebe:
:+1: need a CHANGELOG entry and the guide needs to be updated.

nkovacs:
I've added a changelog entry. The guide is missing information about the `$type` attribute completely.

SilverFire:
> The guide is missing information about the $type attribute completely

http://www.yiiframework.com/doc-2.0/yii-validators-datevalidator.html#$type-detail

?

nkovacs:
https://github.com/nkovacs/yii2/blob/datetime-shorthands/docs/guide/tutorial-core-validators.md#yiivalidatorsdatevalidatordate-

SilverFire:
Ah :) It could be added

cebe (approved):

cebe:
Merged, thank you!

cebe:
Merged, thank you!

Issue #12191, opened 2016-08-18
UniqueValidator on combinations problems


PowerGamer1:
Let say a model has three attributes a1, a2, a3 which can be empty. Lets say all three must constitute a unique combination. Lets say model validators are configured as follows:

```PHP
['a1', 'makeSureIsValidIfNotEmpty'],
['a2', 'makeSureIsValidIfNotEmpty'],
['a3', 'makeSureIsValidIfNotEmpty'],
[['a1', 'a2', 'a3'], 'unique', 'targetAttribute' => ['a1', 'a2', 'a3']],
```

Problem 1. When a1, a2, a3 have valid non-empty values `UniqueValidator` will perform the same validation THREE times. This is completely unnecessary and undesirable considering the validation involves quering the database.

Problem 2. When a1 is valid and non-empty, but a2 is empty and a3 is invalid `UniqueValidator` will perform validation once (on a1) but using EMPTY value of a2 and INVALID value of a3. This is simply pointless and wrong.

I am not sure how this can be fixed considering the overall design of Yii2 data validation, but the problems are bad. In its current state `UniqueValidator` is good enough for single attribute validation only.

PowerGamer1:
I think I figured an acceptable (and backwards compatible) solution for most of the problems:

https://github.com/yiisoft/yii2/compare/master...PowerGamer1:patch-12

The validation will still be performed when some of the attributes are empty (which may be needed in some cases).

terabytesoftw:
Thanks for the report. I re-checked this on current Yii2 master on March 2026.

It looks like this issue is **partially resolved**:

- **Problem 2** (running `UniqueValidator` with invalid target attributes) appears fixed by #19407 (released in `2.0.46`), where `skipOnError` is respected across `targetAttribute`.
- **Problem 1** (duplicate validation/query execution for composite `targetAttribute`) still reproduces: with 3 attributes in one composite rule, validation is executed 3 times.

So this issue still seems relevant for the performance/behavior part of composite unique validation.

PR #12200, opened 2016-08-18
Update UniqueValidator.php

PowerGamer1:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | https://github.com/yiisoft/yii2/issues/12191 |

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -103,7 +103,7 @@ class UniqueValidator extends Validator
             if (is_int($modelAttribute)) {
                 $modelAttribute = $targetAttribute;
             }
-            if ($this->skipOnError && $model->hasError($modelAttribute))
+            if ($this->skipOnError && $model->hasErrors($modelAttribute))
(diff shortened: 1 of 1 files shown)


cebe:
@PowerGamer1 are you willing to work on this further?

PowerGamer1:
@cebe
Sorry, I don't remember now how exactly this code works and what it solves. But I remember deciding not to use this fix and making my own custom validator instead (no problems with BC that way).

Ideally, validation of multiple attributes in Yii2 (like in this case) should be redesigned from scratch (in Yii2 2.0 or whatever, personally I don't care much for breaking BC). Also, the second problem I described in https://github.com/yiisoft/yii2/issues/12191 can be solved in backwards compatible way as I suggested here: https://github.com/yiisoft/yii2/issues/12636#issuecomment-305740699.

cebe:
Thanks, will check that.

PR #12240, opened 2016-08-21
Add UrlNormalizer for UrlRule and UrlManager

### LINKED ISSUE #7670: UrlRule Problem with trailing slash (/)

### LINKED ISSUE #8281: GroupUrlRule with prefix has unexpected behavior for empty ('') rules
Hello,

I've encountered a "strange" (at least for me it is strange) behavior while using a GroupUrlRule.

Using the following rule:

```php
new \yii\web\GroupUrlRule(
        [
            'prefix'      => 'ideas',
            'routePrefix' => 'keywordIdeas',
            'rules'       => [
                ''    => 'index/index',
            ]
        ]
    )
```

Matches for the given URL:

```
/ideas
```

But NOT for:

```
/ideas/
```

The same applies if I change the rule to:

```
'/'    => 'index/index',
```

But if I add a rule:

```
'/'    => 'index/index',
'/?' => 'index/index'
```

/ideas is matched by the first rule and /ideas/ by the second. I think this is kind of unexpected.

I would assume that an empty rule would always be matched by /ideas and /ideas/.
And given the case I mentioned here, I would further assume, that the rule '/' only matches for /ideas/ but it only matches /ideas.

Edit: Any rule starting with '/' does not work.

Is this a bug or expected behaviour?

### LINKED ISSUE #8370: Default url parsing logic does not enforce trailing slash or not
If you have actionIndex in SiteController, `site/index` and `site/index/` both work. But if you create a url rule, e.g.:

`'foo/<action:\w+>' => 'site/<action>'`

then only `foo/index` will work, because of #1807.

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | yes |
| Breaks BC? | ? |
| Tests pass? | yes |
| Fixed issues | #7670, #8281, #8370 |

fixes #7670, fixes #8281, fixes #8370
closes #10745, closes #11381

Another attempt to #7670. :) Mostly based on #11381, but with many simplifications and improved performance. By default, there is only one normalizer assigned for UrlMananger and UrlRules reuse it. Each UrlRule can have their own normalizer, which will be cached with rule. Config for normalization of trailing slashes was simplified - right now this behavior depends on suffix: 
- if `suffix` has trailing slash, trailing slash will be forced during normalization
- if `suffix` does not have trailing slash, trailing slashes will be removed during normalization.

TODO:
- [x] Implementation
- [x] Tests
- [x] Docs


framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,112 @@
...
+            return $pathInfo;
+        }
+
+        $sourcePathInfo = $pathInfo;
+        if ($this->collapseSlashes) {
+            $pathInfo = ltrim(preg_replace('#/{2,}#', '/', $pathInfo), '/');
+        }
+
+        if ($this->normalizeTrailingSlash === true) {
+            if (substr($suffix, -1) === '/' && substr($pathInfo, -1) !== '/') {
```
samdark: I'd not expect URL is necessary in ASCII.

cronfy:
Just tested it with my live site - it is working (configured for three different rule groups: ```suffix => '/'```, ```suffix => '.html'``` and ```'normalizer' => false```).

Simplifying normalizer config via ```$suffix``` is good, config looks cleaner.

The only thing I've found confusing was that normalizer can be disabled in rules with ```'normalizer' => false```, but the same was not working for ```UrlManager```.

rob006:
Ready for review.

I accidentally removed all trailing whitespaces in `docs/guide/runtime-routing.md` - please let me know if I need to revert this or can remain as it is now.

docs/guide/runtime-routing.md
```diff
@@ -512,6 +512,50 @@ contains a customized URL rule which uses `.json` as its suffix instead of the g
 ```
 
 
+### URL normalization <span id="url-normalization"></span>
+
+Since version 2.0.10 [[yii\web\UrlManager|UrlManager]] uses [[yii\web\UrlNormalizer|UrlNormalizer]] for dealing
+with variations of the same URL with and without trailing slash. Because technically `http://example.com/path`
+and `http://example.com/path/` are different URLs, serving the same content for both of them can degrade SEO.
+By default normalizer collapses consecutive slashed, adds or removes trailing slashes depending on whether the
```
samdark: collapses consecutive slashed → collapses consecutive slashes?

docs/guide/runtime-routing.md
```diff
@@ -512,6 +512,50 @@ contains a customized URL rule which uses `.json` as its suffix instead of the g
...
+
+```php
+[
+    'components' => [
+        'urlManager' => [
+            'enablePrettyUrl' => true,
+            'showScriptName' => false,
+            'enableStrictParsing' => true,
+            'suffix' => '.html',
+            'normalizer' => [
```
samdark: How about naming the property `normalize` instead of `normalizer`?

framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * UrlNormalizer normalizes URLs for UrlManager and UrlRule.
+ *
+ * @author Robert Korulczyk <robert@korulczyk.pl>
+ * @author Cronfy <cronfy@gmail.com>
+ * @since 2.0.10
+ */
+class UrlNormalizer extends Object
+{
+    const ACTION_REDIRECT_PERMANENT = 301;
```
samdark: Worth adding phpdoc for these constants.

docs/guide/runtime-routing.md
rob006: It contains an object - naming it using the verb is IMO really weird.

samdark: OK.

docs/guide/runtime-routing.md
rob006: Fixed.

framework/web/UrlNormalizer.php
rob006: Done.

samdark:
Nice! I looks good to me now code-wise. @yiisoft/core-developers any issues?

framework/web/UrlManager.php
```diff
@@ -125,6 +125,12 @@ class UrlManager extends Component
      * specified via [[rules]] will take precedence when the same property of the rule is configured.
      */
     public $ruleConfig = ['class' => 'yii\web\UrlRule'];
+    /**
+     * @var UrlNormalizer|array|boolean the configuration for [[UrlNormalizer]] used by this UrlManager.
+     * If `false` normalization will be skipped. If true, normalizer will use default settings.
```
SilverFire: `true` -> <code>\`true\`</code>

framework/web/UrlRule.php
```diff
@@ -232,8 +270,11 @@ public function parseRequest($manager, $request)
             return false;
         }
 
-        $pathInfo = $request->getPathInfo();
         $suffix = (string)($this->suffix === null ? $manager->suffix : $this->suffix);
+        $pathInfo = $request->getPathInfo();
+        if ($this->hasNormalizer($manager)) {
+            $pathInfo = $this->getNormalizer($manager)->normalizePathInfo($pathInfo, $suffix, $normalized);
```
SilverFire: Maybe, it's better to initialize `$normalize` as `false`, before passing it somewhere?

framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+namespace yii\web;
+
+use Yii;
+use yii\base\Object;
+use yii\base\InvalidConfigException;
+
+/**
+ * UrlNormalizer normalizes URLs for UrlManager and UrlRule.
```
SilverFire: UrlManager -> `[[\yii\web\UrlManager|UrlManager]]`
UrlRule -> `[[\yii\web\UrlRule|UrlRule]]`

framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,125 @@
...
+     *       $route[1]['oldRoute'] = $route[0];
+     *       $route[0] = 'site/redirect';
+     *       return $route;
+     *   }
+     *   ```
+     */
+    public $action = self::ACTION_REDIRECT_PERMANENT;
+
+    /**
+     * Performs normalization action for specified route.
```
SilverFire: `for specified route` -> `for the specified $route`

framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,125 @@
...
+            return call_user_func($this->action, $route, $this);
+        } else {
+            throw new InvalidConfigException('Invalid normalizer action.');
+        }
+    }
+
+    /**
+     * Normalizes specified pathInfo.
+     * @param $pathInfo string pathInfo for normalization
+     * @param $normalized boolean if specified, this variable will be set to true if pathInfo
```
SilverFire: true -> <code>\`true\`</code>

SilverFire: pathInfo -> $pathInfo

framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public function normalizePathInfo($pathInfo, $suffix, &$normalized = false)
+    {
+        if (empty($pathInfo)) {
+            return $pathInfo;
+        }
+
+        $sourcePathInfo = $pathInfo;
+        if ($this->collapseSlashes) {
+            $pathInfo = ltrim(preg_replace('#/{2,}#', '/', $pathInfo), '/');
```
SilverFire: Worth moving to a separate private method `$this->collapseSlashes($pathInfo)`

framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        if ($this->normalizeTrailingSlash === true) {
+            if (substr($suffix, -1) === '/' && substr($pathInfo, -1) !== '/') {
```
SilverFire: Worth moving to a separate private method `$this->normalizeTrailingSlash($pathInfo, $suffix)`

framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,125 @@
...
+                $pathInfo .= '/';
+            } elseif (substr($suffix, -1) !== '/' && substr($pathInfo, -1) === '/') {
+                $pathInfo = rtrim($pathInfo, '/');
+            }
+        }
+
+        $normalized = $sourcePathInfo !== $pathInfo;
```
SilverFire: I'd prefer `strcmp` here

framework/web/UrlNormalizerRedirectException.php
```diff
@@ -0,0 +1,48 @@
...
+     * @var array params used for generate URL for redirection
+     */
+    public $params;
+    /**
+     * @var integer the HTTP status code
+     */
+    public $statusCode;
+
+    /**
+     * @param string $route route used for generate URL for redirection
```
SilverFire: for -> to

framework/web/UrlNormalizerRedirectException.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param string $params params used for generate URL for redirection
```
SilverFire: same here

framework/web/UrlNormalizer.php
samdark: Why?

framework/web/UrlNormalizer.php
rob006: What are the benefits of this? Moving this to private method only make it harder to extend.

SilverFire: Oh, I've meant protected

framework/web/UrlNormalizer.php
SilverFire: Nevermind :)

rob006:
Done.

SilverFire:
Have nothing more to add. Good work!

rob006:
Great. I will squash and rebase tonight.

framework/web/UrlManager.php
```diff
@@ -117,14 +117,20 @@ class UrlManager extends Component
...
      * @var array the default configuration of URL rules. Individual rule configurations
      * specified via [[rules]] will take precedence when the same property of the rule is configured.
      */
     public $ruleConfig = ['class' => 'yii\web\UrlRule'];
+    /**
+     * @var UrlNormalizer|array|boolean the configuration for [[UrlNormalizer]] used by this UrlManager.
+     * If `false` normalization will be skipped. If `true`, normalizer will use default settings.
+     * @since 2.0.10
+     */
+    public $normalizer = true;
```
cebe: the usual convention for default config is to use `null`, not `true`

cebe: or even better use `['class' => 'yii\web\UrlNormalizer']`.

framework/web/UrlNormalizer.php
```diff
@@ -0,0 +1,151 @@
...
+     *   ```php
+     *   function ($route, $normalizer) {
+     *       // use custom action for redirections
+     *       $route[1]['oldRoute'] = $route[0];
+     *       $route[0] = 'site/redirect';
+     *       return $route;
+     *   }
+     *   ```
+     */
+    public $action = self::ACTION_REDIRECT_PERMANENT;
```
cebe: permanent redirection can be a real pain when developing as they are cached in the browser. Maybe we should make this value depend on YII_DEBUG?

framework/web/UrlRule.php
```diff
@@ -89,6 +89,13 @@ class UrlRule extends Object implements UrlRuleInterface
      * @var boolean a value indicating if parameters should be url encoded.
      */
     public $encodeParams = true;
+    /**
+     * @var UrlNormalizer|array|boolean the configuration for [[UrlNormalizer]] used by this rule.
+     * If `true`, [[UrlManager::normalizer]] will be used, if `false`, normalization will be skipped
+     * for this rule.
+     * @since 2.0.10
+     */
+    public $normalizer = true;
```
cebe: should also default to `null` instead of `true`.

framework/web/UrlManager.php
rob006: `['class' => 'yii\web\UrlNormalizer']` will not work for UrlRule - each rule will get then own instance of normalizer. I'd like to avoid different syntax for configuring normalizer for UrlRule and UrlManager. 
Also `true|false` is IMO more intuitive, because you can read it as "normalizer is enabled/disabled". `$normalizer = null` looks like "there is no normalizer, so no normalization should be done".

framework/web/UrlNormalizer.php
rob006: This should be configured in app config. IMO changing `YII_DEBUG` should not implicitly change app behavior.

framework/web/UrlManager.php
cebe: If this property was named `normalize`, setting it to `true` would make sense. As it contains the `normalizer` object true is not really fitting here. 

Having a difference between the options for UrlRule and for UrlManager is no problem because UrlManager has a default config, while UrlRule has a default fallback to what is configured in UrlManager.

My main point here is consistency with other places where similar object dependencies exist.

For UrlRule it should be null like e.g. `yii\web\User::$accessChecker`: https://github.com/yiisoft/yii2/blob/09b1fd47e071acdc9ec0c1988684d367fb067e02/framework/web/User.php#L107-L112

- null: use the default instance from UrlManager
- false: disable it (similar to Sort and Pagination in DataProviders)
- array: a config to create custom normalizer.

For UrlManager:

- `['class' => 'yii\web\UrlNormalizer']` - default value
- false to disable
- array for other config.

rob006: What if `UrlManager::$normalizer` is null?

cebe: that should not be a possible option.

rob006: Fixed.

framework/web/UrlRule.php
rob006: Fixed.

rob006:
Squashed and rebased. :ship:

samdark:
Merged. Thanks everyone for working on it. It was a hard one but you did it.

cronfy:
Excellent! Thank you all.

stovbur:
What about lettercase normalization? (in UrlNormalizer.php)

samdark:
@stovbur place create separate issue for that. This one is closed.

PR #12241, opened 2016-08-22
Use capital letter

Mak-Di:


samdark:
Since it's a type, we should not capitalize in this case.

Issue #11582, opened 2016-05-18
Pagination on page validate must redirect to right page not replace inside


DrDeath72:
wrong pages became copy of real last page

samdark:
How to reproduce the issue?

DrDeath72:
?page=99999999

DrDeath72:
first version same behavior http://yiiframework.ru/news/page/100000

samdark:
I see. So you want this to throw 404 instead, right?

DrDeath72:
i think it must give you choice, 404, redirect or like now :) seo etc...

tunecino:
isn't that a good thing ? I mean the last page is auto calculated by default while `yii\data\Pagination::getPageCount()` can be used if a different logic is required. same happens with REST: you get an `X-Pagination-Page-Count` value in headers or `_meta` to control it if needed. changing it now won't be a BC ?

rob006:
> isn't that a good thing ?

This is definitely a bad thing if it is the default behavior. If page 123 returns data from page 99 it is simply a bug.

githubjeka:
To fix this behavior set `\yii\data\Pagination::$validatePage` to `false`
By default it's `true`

>  * When this property is true, the value of [[page]] will always be between 0 and ([[pageCount]]-1).

DrDeath72:
validatePage only link/alias wrong page to last

githubjeka:
Yes it's by default and documented. To fix it you can set `validatePage` to `false`.

![image](https://cloud.githubusercontent.com/assets/874234/19142004/ca64773e-8ba2-11e6-951f-22d042997cb9.png)

DrDeath72:
it's not equal 404 error, just empty data with wrong limit/offset

githubjeka:
404 - No sure that it's correct when `$validatePage = false`. But when you set page set to a negative value or a word then the `page` param **always** return first page. This is bug sure.

githubjeka:
And I don't understand responsibility of class Pagination. Is it should be to return NotFoundException eg...
Probably not. You should validate `page` param in controller and retrun NotFoundException youself.

githubjeka:
@samdark Here it's necessary to clarify exactly what you need to improve.

samdark:
Documentation at least.

rob006:
> And I don't understand responsibility of class Pagination. Is it should be to return NotFoundException eg...
[...]

PR #12251, opened 2016-08-22
\yii\data\Filter

niksamokhvalov:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #4479 |

`\yii\data\Filter` for filtering in the RESTful services and forms (GridView). I suggest used parameter `?filter` in format JSON for filtering the collections:

```
/api/v1/users?filter={"field": "value"}
```

and array from the query parameters for forms:

`/news?NewsSearch[year]=2016&NewsSearch[active]=1`

**Operators**
- !=
- >
- <
- <=
- >= 
- % (set in the value)

**Examples**

```
/api/v1/users?filter={"id": [1,6,46,8], "active": 1)
/api/v1/users?filter={"!=id": 13, "age": null)
/api/v1/users?filter={"email": "nik@samokhvalov.info)
/api/v1/users?filter={">=age": 18, "<age": 30, "gender": "w", "name": "%Natalie")
```

**Work in progress**
- [x] Basic logic
- [ ] Data providers
- [ ] tests
- [ ] docs


framework/data/Filter.php
```diff
@@ -0,0 +1,122 @@
...
+                                
+                if ($operator) {
+                    if (is_array($value) && !empty($value) && $operator === '!=') {
+                        $operator = 'NOT IN';
+                    }
+                    $filter[] = [$operator, $field, $value];
+                } elseif (substr($value, 0, 1) === '%' || substr($value, -1, 1) === '%') {
+                    $filter[] = ['like', $field, $value, false];
+                } else {
+                    $filter[] = [$field => $value];
```
SilverFire: We already have [Query::andFilterCompare()](https://github.com/yiisoft/yii2/blob/master/framework/db/Query.php#L584-L619). Can we reuse it and prevent code duplication.

framework/data/Filter.php
```diff
@@ -0,0 +1,122 @@
...
+            $model = new $modelClass;
+
+            foreach ($requestFilter as $field => $value) {
+                list($operator, $field) = $this->prepareField($field);
+
+                if (!$this->isAcceptableField($field) || !$model->hasAttribute($field)) {
+                    throw new NotAcceptableHttpException('Filter by field "' . $field . '" unsupported.');
+                }
+                                
+                if ($operator) {
```
SilverFire: I'd suggest to list in `$operators` property all operators and corresponding methods:

```php
public $operators = [
   '>=' => 'buildGteCondition',
   '<=' => 'buildLteCondition', 
   '>' => 'buildGtCondition',
   '<' => 'buildLtCondition',
   '!=' => 'buildNeqCondition'
];
```

```php
if ($operator) {
    $this->buildCondition($operator, $field, $value);
}
```

```php
public function buildCondition($operator, $field, $value)
{
    if (isset($this->opeartors[$operator])) {
        $method = $this->conditionBuilders[$operator];
    } else {
        $method = 'buildSimpleCondition';
    }

    return $this->$method($operator, $field, $value);
}

protected function buildSimpleCondition($operator, $field, $value)
{
    return [$operator, $field, $value];
}

protected function buildNeqCondition($operator, $field, $value) 
{
     if (is_array($value) && !empty($value)) {
         $operator = 'NOT IN';
     }

     return $this->buildSimpleCondition($operator, $field, $value);
}
```

vologab:
must have feature, good way to describe filters

framework/data/Filter.php
niksamokhvalov: In `andFilterCompare()` operators is passed via values. While in the REST of the operators are passed via the field names. `andFilterCompare()` does not support `!=` operator.

samdark:
Build seems to fail: https://travis-ci.org/yiisoft/yii2/builds/156590029

niksamokhvalov:
Yeah, I know. First make the base, then will fix everything and test.

samdark:
OK.

tests/framework/data/FilterTest.php
```diff
@@ -0,0 +1,35 @@
...
+    public function testBuildCondition()
+    {
+        $field_name   = 'some_field';
+        $value_string = 'some_string';
+        $value_int    = 8;
+        $value_array  = [15, 16];
+        $filter       = new Filter();
+
+        $condition = $filter->buildCondition('>=', $field_name, $value_int);
+        $this->assertEquals($condition, ['>=', $field_name, $value_int]);
```
SilverFire: Could be done in one line:

```php
$this->assertEquals(['>=', $field_name, $value_int], $filter->buildCondition('>=', $field_name, $value_int));
```

tests/framework/data/FilterTest.php
```diff
@@ -15,21 +18,25 @@ public function testBuildCondition()
...
-        $this->assertEquals($condition, ['<', $field_name, $value_int]);
-        $condition = $filter->buildCondition('!=', $field_name, $value_int);
-        $this->assertEquals($condition, ['NOT', [$field_name => $value_int]]);
-        $condition = $filter->buildCondition('!=', $field_name, $value_array);
-        $this->assertEquals($condition, ['NOT IN', $field_name, $value_array]);
-        $condition = $filter->buildCondition(null, $field_name, $value_int);
-        $this->assertEquals($condition, [$field_name => $value_int]);
-        $condition = $filter->buildCondition('like', $field_name, $value_string);
-        $this->assertEquals($condition, ['like', $field_name, $value_string, false]);
+        $this->assertEquals($filter->buildCondition('>=', $field_name, $value_int), ['>=', $field_name, $value_int]);
```
SilverFire: `assertEquals` signature is `assertEquals($expected, $actual)`. Flip the arguments please

SilverFire (changes requested):

klimov-paul:
I am against such change: it looks like re-enventing our current model layer.
`yii\base'Model` alreay handles all features added here.

Filters are usualy specified via web forms, which are supported by models. The `DataProvider::filter` introduced here can be used to compose a filter web form.

Here we redeclare entire `Query::where` specification at the `Filter` class, for what?
`Filter::fields` is a just another version of `Model::getSafeAttributes()`.

niksamokhvalov:
> Filter::fields is a just another version of Model::getSafeAttributes().

@klimov-paul it is not. `\yii\data\Filter::$fields` gives the flexibility to manage filtering rules. For example, you can disable filter by big attribute of the model.

klimov-paul:
> \yii\data\Filter::$fields gives the flexibility to manage filtering rules. For example, you can disable filter by big attribute of the model.

In the `Model` we have a `scenario` for this task, so may disable some attributes from being populated and processed at particular use case. At the Gii CRUD generator we create a special model for the search, which declares its own validation rules and filtering logic.

Instead of using the instuments already provided, you decide to duplicate them all into a separated entity called `Filter`.

klimov-paul:
See https://github.com/yiisoft/yii2/issues/4479#issuecomment-247289250

niksamokhvalov:
@klimov-paul I have not done much work with forms, but very much with RESTful API. Problem with RESTful APIs is that there is no filtering. Filtering in the API should be implemented on a single standard.

In my opinion…

---

**The format of the filter**
Regardless of how you pass parameters to a filter (`$_POST['FormName']['field']` or `$_GET['filter']`), model must work with the one data format.

**Operators**
The client must manage the filtering rules, specifying in the request the operators (`>`, `<`, `!=`, etc.).

**Errors**
Invalid or unsupported filter should lead to the exception and display an error to the client. It is unacceptable that the client received the 200 OK response and believed that a collection filtered. This is especially critical for RESTful services, with which work bots.

---

So I think that the DataProvider should deal with the filtering model. Because it's just a constructing of query to the model, is the same as sorting (Sort class). DataProvider configuration still can be  SearchModel->search(). Just will not need to set up simple rules about which framework can guess itself (the default behavior).

> yii\base'Model alreay handles all features added here.

The filter specified on the client is different from what is required in ActiveRecord. So need the adapter.

I want to add filtering in RESTful Yii core, but if you are not interested, just tell me about it, please ;-)

klimov-paul:
> Filtering in the API should be implemented on a single standard.

It is weird to here this from you. When I told you all your code is a duplication of already existing features, and general approach of using "search" models like it is done in normal CRUD should be used for REST as well - you do not listen. However, instead you are talking about single filtering approach, while you are unabel to answer how **your** single approach can work with regular forms in the CRUD.

> The format of the filter
Regardless of how you pass parameters to a filter ($_POST['FormName']['field'] or $_GET['filter']), model must work with the one data format.

So, what is the problem? The following code works just fine:

```php
$model = new SearchModel();
$model->setAttributes($_POST['filter']); // or $model->setAttributes($_GET['filter']);
```

As well as you can create filter model with form name `filter`:

```php
class SearchModel extends \yii\base\Model
{
    public function formName()
    {
        return 'filter';
    }
}
```
and then use regular `load()` for it:

```php
$model = new SearchModel();
$model->load(Yii::$app->request->post());
```

> Operators
The client must manage the filtering rules, specifying in the request the operators (>, <, !=, etc.).

It is already supported by `yii\db\Query`:
https://github.com/yiisoft/yii2/blob/master/framework/db/Query.php#L584-L610
What is the reason of duplicating somewhere else?

> Errors
Invalid or unsupported filter should lead to the exception and display an error to the client. It is unacceptable that the client received the 200 OK response and believed that a collection filtered. This is especially critical for RESTful services, with which work bots.

Do I need to teach you how to use `Model::validate()` method and process its result? Very well, here you go:

```php
$model = new SearchModel();
$model->load(Yii::$app->request->post());
if ($model->validate()) {
    $dataProvider = $model->search();
    // ...
} else {
    throw new BadRequestHttpException('Error: ' . implode("\n", $model->getFirstErrors()));
}
```

> So need the adapter.

The search model is an adapter already. It incapsulates processing of the filter specification. For example:

```php
class SearchModel extends \yii\base\Model
{
    public $title;

    public function search()
    {
        $query = Item::find();        
            'query' => $query,
        ]);
  
        // here search model decides how it will apply search attributes to the actual `Query`
        $query->andFilterWhere(['like', 'title', $this->title]); // could be partial match
        // or exact match:
        //$query->andFilterWhere([''title' => $this->title]); 

        return $dataProvider;
    }
}
```

framework/helpers/BaseArrayFilterHelper.php
```diff
@@ -0,0 +1,249 @@
...
+use yii\base\InvalidParamException;
+
+/**
+ * BaseArrayFilterHelper provides concrete implementation for [[ArrayFilterHelper]].
+ *
+ * Do not use BaseArrayFilterHelper. Use [[ArrayFilterHelper]] instead.
+ *
+ * @author Daniil Razorenov <daniltmb@mail.ru>
+ */
+class BaseArrayFilterHelper
```
SilverFire: This class contains a single public method, thus this is not a helper. It must be implemented not as a class of static method.

SilverFire (changes requested):

SilverFire:
> The search model is an adapter already. It incapsulates processing of the filter specification. For example:

Yes it does, but what happens, when you want to create all possible filters for some filed? I mean do think that creating of 6 properties and 6 calls of `$query->andFilterWhere` is worth than having some agreement on how we can prefix an attribute to get the same work done without copy-pasting?



> Do I need to teach you how to use Model::validate() method and process its result? Very well, here you go:

It's a question of preferring imperative over declarative error handling.


> https://github.com/yiisoft/yii2/blob/master/framework/db/Query.php#L584-L610
> What is the reason of duplicating somewhere else?

We duplicate thing to isolate them. I don't think that coupling between `Form` and `Query` is the way to solve this problem.


@niksamokhvalov don't rush to apply my suggestions, we should wait for more opinisons from @yiisoft/core-developers

niksamokhvalov:
@klimov-paul I know how to do a search and validation :-) My PR is about a different issue.

I did a lot of RESTful services to many projects on Yii2. Sometimes the models contain a large set of attributes. If I hadn't used my custom dataProvider, and every time doing copy-paste as you suggest:

* I would have to always write the same code in the `IndexAction` and `SearchModel`.
* The client does not have the opportunity to flexibly control filtering (filtering on the same field using different operators).

Your approach: the developer must write filter logic (search) for each model.
My approach: collection filtering should default to work as sorting and pagination.

I'm just trying to say: "Hey, guys, let's add this feature in the Yii core and stop doing the copy-paste"? :-)

--

> However, instead you are talking about single filtering approach, while you are unabel to answer how your single approach can work with regular forms in the CRUD.

As I already said, I have not done much work with forms, but very much with RESTful API. Let's think together about how to combine form and REST?

> It is already supported by yii\db\Query:
> https://github.com/yiisoft/yii2/blob/master/framework/db/Query.php#L584-L610
> What is the reason of duplicating somewhere else?

1. Operators must be passed with field names, not in the values. Otherwise, you will not be able to do such filtering: : `filter={"field":!null}` — `!null` is not valid.
1.  Encapsulation.

klimov-paul:
> As I already said, I have not done much work with forms, but very much with RESTful API. Let's think together about how to combine form and REST?

klimov-paul:
> I know how to do a search and validation

Your are ignoring validation of the filter data. What will happen if incoming filter has a following value:

```json
{
    "price": "pretty"
}
```
?
Who will check that filter value is valid, e.g. client does not attempt to compare integer field with string? Who will check if operators like `>`, `<` and so on will not be used against string fields?

> As I already said, I have not done much work with forms, but very much with RESTful API. Let's think together about how to combine form and REST?

Then why do you put `filter` inside the `DataProvider`?
If your solution covers REST - it should be bound to the REST only. In particular all(!) your changes shuld be made inside `yii\rest\*` namespace. Why do they affect `DataProvider`?

> My approach: collection filtering should default to work as sorting and pagination.

If your solution is general enough to be actually put inside `DataProvider` as you claim it, then provide the answer for the question I have asked you twice already: 

HOW IT CAN BE USED TO COMPOSE A REGUALR FILTER WEB FORM?

>  Let's think together about how to combine form and REST?

There is NO such combination. because they are different things!
While you use JSON format you can create conditions like:

```json
{
    "filter": {"price": {">": 10}}
}
```

But such input can NOT be composed from the regular `<input type="text"/>`!

klimov-paul:
Indeed we can have some layer/helpers for the process of incoming filter. 
But actual application of the filter can not be made universal. It is sad - but it can not.

1) Actual data filtering depends on the data source abilities: relational database, Redis and MongoDB compose search conditions in the different ways.

2) Search can not be limited by the list of attributes with same operands avalable. Think about Full-Text search: what if I want to allow filtering by full-text match? I will create a special filter condition named `search` or `match`, which should accept only string(!) and should NOT be used in operators like `<` or `>`. And it processing depends on the application internals: it may require usage of the Sphinx or ElasticSearch or compose non-standard conditions for MySQL or MongoDB.

Thus there must be a separated layer, allowing to control filter specification and application for particular resource.

klimov-paul:
Competitor pull request opened at #12641

niksamokhvalov:
> Who will check that filter value is valid, e.g. client does not attempt to compare integer field with string? Who will check if operators like >, < and so on will not be used against string fields?

Yes, some validations must be done by the model.

> Search can not be limited by the list of attributes with same operands avalable. Think about Full-Text search: what if I want to allow filtering by full-text match? I will create a special filter condition named search or match, which should accept only string(!) and should NOT be used in operators like < or >. 

But it's not filtering, that's exactly search. A full-text search query is passed as a separate parameter, for example: `/news?q=blabla-bla-bla`. Of course, for this query you need to write custom logic in the model.

I looked at the work of the forms (GridView::widget), and corrected the Filter for working with REST and forms.

## Concept

### Configuration

`Filter` is configured via `ActiveDataProvider`. You can specify the parameter name in the request and the expected format:

```php
new ActiveDataProvider([
      'query' => $model::find(),
      'filter' => [
           'format' => Filter::FORMAT_ARRAY, // by default, or `FORMAT_JSON` for REST
           'filterParam' => $model->formName() // or `filter` for REST
       ]
]);
```

### RESTful

`GET /api/v1/users?filter={"id": [1,6,46,8], "active": 1)`

You need set the search model to `public $searchModelClass;` in REST controller class (if not set, it will use the basic model). 

Maybe we should add the ability to specify the search scenario for model... (?)

### Form

For example, models `News` and `NewsSearch`.

`GET /news?NewsSearch[year]=2016&NewsSearch[active]=1`

Everything works the same, as already generates CRUD Generator. Just the usual filtering rules do not need to register manually:

```php
class NewsSearch extends News
{
    {
            'query' => static::find(),
            'filter' => [
                'filterParam' => $this->formName()
            ]
        ]);

        if ($dataProvider->getFilter()) {
            $this->load($dataProvider->getFilter()->getData(), '');

            if (!$this->validate()) {
                // uncomment the following line if you do not want to return any records when validation fails
                // $query->where('0=1');
                // and comment out the following line
                $dataProvider->setFilter(false);
                return $dataProvider;
            }
        }

        return $dataProvider;
   }
}
```

Profit:

1. Use the same Filter.
1. The GridView will be possible in the future to learn to use operators in field names, for example: `<input name="NewsSearch[>year]">`. In addition, will be able to substitute the operators dynamically using JS.
1. Now all the filter options from the form will be transmitted with an accurate comparison, but you can add the ability to modify the filter like this (it has not yet done):
```php
new ActiveDataProvider([
      'query' => $model::find(),
      'filter' => [
           'filterParam' => $model->formName(),
           'conditions' => [
           		'date_from' => function($query, $operator, $operand, $value) {
           			return $query->andWhere(['>=', 'date', $value]);
           		}
           ]
       ]
]);
```

Now it is only a draft. If the concept suits all, I'm ready to continue.

niksamokhvalov:
@klimov-paul @SilverFire @samdark What do you think about my last comments?

klimov-paul:
> What do you think about my last comments?

First of all, the build is failing, see: https://travis-ci.org/yiisoft/yii2/jobs/164109508

Secondly the entire concept is risk prone: hacking prone - to be more specfific. You allow any(!) listing built up via `DataProvider` to be magically filtered by custom condition specified via `filter` GET parameter!

> I looked at the work of the forms (GridView::widget), and corrected the Filter for working with REST and forms.

Solution does not seem to be suitable for the REST as it does not support conjuction conditions like `OR` and `AND`. It can not put 2 conditions on the same attribute, like `$price >2 && $price < 10`.

>             'filter' => [
                'filterParam' => $this->formName()
            ]

This is simply dirty hack, easy to be forgotten, by the way. Instead of leave data population at the model layer - you still drag it to the filter incapsulating it somewhere deep inside, instead of putting it to the controller layer where it belongs!

Compare code generated by Gii CRUD:

```php
class ItemController extends \yii\web\Controller
{
public function actionIndex()
    {
        $searchModel = new ItemSearch();
        $dataProvider = $searchModel->search(Yii::$app->request->queryParams); // filter source detemined by controller

            'searchModel' => $searchModel,
        ]);
    }
...
class ItemSearch extends Model
{
    /**
     * Creates data provider instance with search query applied
     * @param array $params
     * @return ActiveDataProvider
     */
    {
         ...
        $this->load($params);
```

with your approach:

```php
class ItemSearch extends Model
{
    public function search()
    {
        if ($dataProvider->getFilter()) {
            $this->load($dataProvider->getFilter()->getData(), ''); // filter source for data provider comes from within data provider
             ...
```

Just look at data provider creation code you have made: it is twisted like a hell: model controls data-provider creating it, but data-provider controls model, providing filter data for it, but model should apply it manually. And, most of all, in case model will not load filter manually and check if it is valid - filter will be applied automatically behind the scene!

> You can specify the parameter name in the request and the expected format:

Similar problem: format should be independent: URL-encoded, JSON or XML - this is a matter for controller layer.

My opinion: this PR is just getting worse.

niksamokhvalov:
> First of all, the build is failing

Of course. As I have already said that we first need to approve the concept, then I'm going to finish this PR.

> Secondly the entire concept is risk prone: hacking prone - to be more specfific. You allow any(!) listing built up via DataProvider to be magically filtered by custom condition specified via filter GET parameter!

Guide of Yii recommended to create a SearchModel. Accordingly, it will specify only the attributes that you can use when filtering.

> Solution does not seem to be suitable for the REST as it does not support conjuction conditions like OR and AND. It can not put 2 conditions on the same attribute, like $price >2 && $price < 10.

Why not? You can: `/api/catalog?filter={">price": 2, "<price": 10}`.

> model controls data-provider creating it, but data-provider controls model

Yes, exactly the same as working sorting and pagination in ActiveDataProvider. `yii\data\Sort` and `yii\data\Pagination` works with query parameters. And it does not bother anyone. Why suddenly `yii/data/Filter` can't do the same?

> And, most of all, in case model will not load filter manually and check if it is valid - filter will be applied automatically behind the scene!

Yes, but be protected from the carelessness of the developer impossible. In addition, we can update Gii, that he would have correctly generated the SearchModel.

> Compare code generated by Gii CRUD:

Unfair example, you missed the setting conditions in the old implementation:

```php
// Your current approach
{
     ...
    $this->load($params);
    ...
    $query->andFilterWhere([
       'id' => $this->id,
       'category_id' => $this->category_id,
       'author_id' => $this->author_id,
       'created_by' => $this->created_by,
       'updated_by' => $this->updated_by,
       'country_id' => $this->country_id,
       'count' => $this->count,
    ]);
    $query->andFilterWhere(['>=', 'created_at', $this->created_at_from]);
    $query->andFilterWhere(['<=', 'created_at', $this->created_at_to]);
    $query->andFilterWhere(['>=', 'updated_at', $this->updated_at_from]);
    $query->andFilterWhere(['<=', 'updated_at', $this->updated_at_to]);
    $query->andFilterWhere(['>=', 'date', $this->date_from]);
    $query->andFilterWhere(['<=', 'date', $this->date_to]);
    $query->andFilterWhere(['like', 'title', $this->title]);
    $query->andFilterWhere(['like', 'description', $this->description]);
    ...
}

// My suggestion
{
    ...
    $this->load($params);
    ...
    if ($dataProvider->getFilter()) {
            $this->load($dataProvider->getFilter()->getData(), '');
    ...
}
```

Most often there is no need to describe the filtering rules, ActiveDataProvider must be able to control the model, just as it does with sorting and pagination. I'm not saying the example download the model attribute is the best solution. If you have smart ideas, I'd love to hear them.

> Similar problem: format should be independent: URL-encoded, JSON or XML - this is a matter for controller layer.

Don't consider it a problem, because this format is only set for the filter. But I can offer to do forms and REST to the general format: `filter[>id]=10&filter[%title%]=ololo`.

klimov-paul:
> Why not? You can: /api/catalog?filter={">price": 2, "<price": 10}.

What about following condition:

```php
($price < 10 && $name == 'cheap') || ($price > 10 && $name == 'expensive')
```

> Yes, exactly the same as working sorting and pagination in ActiveDataProvider. yii\data\Sort and yii\data\Pagination works with query parameters. And it does not bother anyone. Why suddenly yii/data/Filter can't do the same?

This is ignorance speaking: they are different. Sorting and pagination are considered as view representation without much logic put on them. It does not significant if list is ordered in some way or not: retrieving entire data set end user can manually compose resorted list - there is nothing to prevent here.
Sort and pagination are more or less universal - that is why they are put into the DataProvider. However, still they have a sophisticated validation built on them.
In particular `Sort` will never automatically compose ordering attributes unless you are using `ActiveDataProvider`. `ArrayDataProvider` and `SqlDataProvider` requires manual setup of sort attributes.
Sort always checked to be in valid format and contain valid attributes.
Pagination also verifies passed page number ensuring it is valid.

Manipulating of the SQL query conditions in automatic way may cause list composition logic change. It may allow end-user to expose the data he should not be allowed to see. Sort and Pagination do NOT bear business logic, while search condition does.
For example: if we have a shopping cart list - it should display records for the currently authenticated user, so for it there is a condition like `$userId == Yii::$app->user->id`. There are search conditions for statuses, groups and so on. Search is business logic - that is why it is always packed inside the Model.

> If you have smart ideas, I'd love to hear them.

See #12641

> Don't consider it a problem, because this format is only set for the filter. But I can offer to do forms and REST to the general format

It seems we should drop such thing as `yii\web\JsonParser` as well and put them inside data provider, should not we?

I am out of this p[ointless disscussion, I will let the rest of the core team to consider this. All I can tell that this PR is breaking a BC changing `DataProviderInterface` and data provider logic, thus can be considered only for 2.x

niksamokhvalov:
> severity:BC breaking

This is ridiculous. The code is not yet completed, and you're talking about breaking BC.

I can see that filtering can be made uniform, but you go the other way. I don't want to argue, in any case, the decision is yours.

samdark:
@niksamokhvalov sorry if it offended you. All the labels applied and all the attacking of the pull request aren't to hurt your feelings but to get the best solution possible in the very end. You definitely did a great job proposing a good concept (I still think it may be applied instead of Paul's proposal). Thank you for it.

If you don't mind, I'll re-open it so we won't lose track of the possible solution when discussing it more. OK?

niksamokhvalov:
OK, no problem. If the concept is interesting, I'm willing to continue discussion and development.

samdark:
@niksamokhvalov I'd love to get your review/feedback for #12641. Initially I thought it's very different to end user compared to your approach but it's not except a bit weird syntax with `$` which I can't say I like much. Probably we'll be able to take best from two proposals.

klimov-paul:
> Yes, exactly the same as working sorting and pagination in ActiveDataProvider. yii\data\Sort and yii\data\Pagination works with query parameters. And it does not bother anyone. 

Actually there are matters around sort and pagination, which DO bother someone: see #11582 for example.

Sort and pagination are just the same user input as and a filter. The `Model` has been designed to deal with user input: pick it up from request, validate and show a corresponding error messages to the end user in case something was specified in wrong way.
For REST especially, sort and pagination should be also passed through the model and validated. And a proper error messages should be returned if they have been specified in wrong way.

The actual reason, why `Pagination` and `Sort` process thier data on thier own is the fact, that in most cases there is no direct input for them. Page number and sort direction are usually picked up from pre-created URL, which are placed in links tag or buttons, and will be triggered by user via mouse pointer. Thus it is less sugnificant for them to be passed via `Model` as they unlikely to be specified wrong. However, they still can - see #11582. And thier incorrect value still need to be handled, and may be handled in different ways.

klimov-paul:
> the developer must write filter logic (search) for each model.
I'm just trying to say: "Hey, guys, let's add this feature in the Yii core and stop doing the copy-paste"?

If you concern about code duplication at the search models you may try to use following extensions:
https://github.com/yii2tech/ar-search
https://github.com/yii2tech/admin

samdark:
Closed in favor of #12641.

samdark:
@niksamokhvalov thank your for working on it.

PR #12259, opened 2016-08-23
Fix "Call to a member function getRoute() on null"

iskander-g:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

When you try to use Url::current([], true)]) on page without controller (404 error) it cause " E_ERROR: Call to a member function getRoute() on null "

### DIFF
--- a/framework/helpers/BaseUrl.php
+++ b/framework/helpers/BaseUrl.php
@@ -385,7 +385,9 @@ class BaseUrl
     public static function current(array $params = [], $scheme = false)
     {
         $currentParams = Yii::$app->getRequest()->getQueryParams();
-        $currentParams[0] = '/' . Yii::$app->controller->getRoute();
+        if(Yii::$app->controller) {
+            $currentParams[0] = '/' . Yii::$app->controller->getRoute();
+        }
         $route = ArrayHelper::merge($currentParams, $params);
         return static::toRoute($route, $scheme);
     }


cebe:
What is the point of calling current() when the current route is invalid?

iskander-g:
Of course it was mistake in logic of our app, but i think anyway fatal error it's not correct result in this situation.

You can return null or false, for example.

cebe:
the functions that generate an url always return a string or fail. I do not see why current() should be an exception here. We could possibly make the error message nicer but it should still result in an exception imo.

samdark:
Agree. Silencing it isn't a good option.

Mak-Di:
>When you try to use Url::current([], true)]) on page without controller

@iskander-g, how to reproduce it?

iskander-g:
@Mak-Di I display a "404 page", by using my main app layout with my global site header widget.

In this header widget I have a link to the sign in form, with one get parameter - redirect URL after login.

So I used this code to create link:
`Url::to(['/site/signin', Users::REDIRECT_GET_PARAM => Url::current([], true)])`

I know it might not seem logical to redirect a user back to the “404 page” after login, however, I wouldn’t consider a fatal error as a better option

Mak-Di:
@iskander-g, Thank you for information.
And in this case the error controller exists.
By default:
```php
/**
     * @inheritdoc
     */
    public function actions()
    {
        return [
            'error' => [
                'class' => 'yii\web\ErrorAction',
            ],
           ...
        ];
    }
```
Method `Url::current([], true)])` returns `http://example.com/site/error`

PR #12267, opened 2016-08-23
Added missing hyphen

Mak-Di:
Added missing hyphen for server/client-side and removed one on the ending of 'ajax'


samdark:
Merged. Thanks!

PR #12273, opened 2016-08-23
Fix code sample indentation in Polish guide [skip ci]

rob006:
This probably should fix missing highliting in PDF:

aa6e640c

vs

3843b591


samdark:
Thanks!

PR #12288, opened 2016-08-24
PL start-installation

bizley:
Start Installation updated with fixed links, better translation and current Nginx configuration.


cebe:
Thanks!

PR #12293, opened 2016-08-24
resolveTableNames() update for MSSQL to add case of linked database t…

hAppywAy:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | no |

With Mircrosoft SQL Server 2008+, [you may use linked servers](https://technet.microsoft.com/en-us/library/ms188279%28v=sql.110%29.aspx) to work with data from multiple sources. In your statement, to request new server and its db from the old one server, you have to prefix remote dB table by server name you declared into MSSQL.

`return 'remoteServerName.dbName.schemaName.tableName';`

To make it possible, it's necessary to modify Schema.php from mssql classes to include remote server name into resolveTableNames().

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,21 +4,21 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "d2005b487c5ff761d806b9a94bfe4cac",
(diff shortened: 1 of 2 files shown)


cebe:
can you elaborate on why and in which cases this is needed?

cebe:
looks good to me, could you add a entry in the CHANGELOG. Also composer.lock should not be committed.

PR #12294, opened 2016-08-24
Added Target test of the following methods:

Mak-Di:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
- getLevels()
- setLevels()

It may be necessary to add one more check [here](https://github.com/yiisoft/yii2/blob/master/framework/log/Target.php#L192) and throws exception, if level is incorrect. What do you think about that?

### DIFF
--- a/tests/framework/log/TargetTest.php
+++ b/tests/framework/log/TargetTest.php
@@ -123,6 +123,24 @@ class TargetTest extends TestCase
         $this->assertNotContains('E_b', $context);
         $this->assertNotContains('E_c', $context);
     }
+
+    /**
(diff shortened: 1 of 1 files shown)


samdark:
Thanks!

Mak-Di:
@samdark , It may be necessary to add one more check here and throws exception, if level is incorrect. What do you think about that?

samdark:
Yes. Definitely.

Mak-Di:
Ok, I'll add it.

Issue #10806, opened 2016-02-11
Active record collection object


dhiman252:
Hi Team,

I cannot find any collection object in the documentation for the active records.

Do we have any feature like the laravel framework offers : https://laravel.com/docs/5.1/collections ?

If not please refer to the link and if possible please add this feature in the next version release !!

Thanks

SilverFire:
Yes, currently Yii has not collection object, we use arrays instead to be lightweight.

In my own development practice it's usually enough except some special cases, that can be solved locally without affecting overall application performance.

Could you describe your use case?

adiramardiani:
Interesting :)
Is it possible to reference the link above (collections) as an enhancement for ArrayHelper ?

SilverFire:
Could you elaborate your idea, please

cebe:
@dhiman252 part of the functionality of the laravel collection is provided by DataProviders in Yii. Another part is directly part of the QueryInterface already. Please describe in detail what you need and why.

dhiman252:
For the ease of use I would like this functionality incorporated in the framework, lets say i have table for user reviews and it contains the message and the points, now if i had to display the user rating average and all the reviews on a single page then i need to fire 2 SQL Queries one getting the reviews average and the other all reviews in an array
 so if we have a collection object then 1 Query and ease of use
like
`/*
$reviewsAvg = $user->reviews->avg('points');
$reviewsCount = $user->reviews->count();
froreach($user->reviews as $review){
    //display review
}
*`
There are a lot of functions associated with the collections which makes the development its easy.

cebe:
```php
$reviewsAvg = $user->getReviews()->avg('points');
$reviewsCount = count($user->reviews);
froreach($user->reviews as $review){ 
    //display review
}
```

This already works in yii.

Alex-Bond:
@cebe here is much more better example:
For example i need to query 1 time and show data in 2 columns, but not more that 10 records per column. First column contains only active elements (`active=1`) and second one showing not active (`active=0`).
In case of not using collections i need to filter it with if's and than show. If i use collections i can use `filter() ` and using closure filter data as i want.
Or lets imagine that i need to show items 3 per row in bootstrap. According to best practice you need to create new row element for each 12 columns, so i need to create some $i and increase it every iteration and based on value print new row element.
In case of collections i can just use `chunk()` method and create loop inside loop.

This is only few of use cases. There are of different ways to use it.

Razzwan:
[...]

PR #12304, opened 2016-08-24
Basic support for asCollection

samdark:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #10806 |

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -225,6 +225,10 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             }
         }
 
+        if ($this->collectionClass !== null) {
+            return new $this->collectionClass($models);
+        }
+
         return $models;
     }
 
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -241,7 +241,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
      */
     protected function createCollection($models)
     {
-        return new $this->collectionClass($models);
+        return \Yii::createObject($this->collectionClass, $models);
     }
 
     /**
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -226,10 +226,23 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         }
 
         if ($this->collectionClass !== null) {
-            return new $this->collectionClass($models);
+            return $this->createCollection($this->collectionClass, $models);
         }
 
         return $models;
+    }
+
+    /**
+     * Instantiates collection
+     *
+     * @param string $class collection class to be instantiated and filled with results.
+     * @param array $models array of models to fill collection with.
+     * @return object collection instance.
+     * @since 2.0.10
+     */
+    protected function createCollection($class, $models)
+    {
+        return new $class($models);
     }
 
     /**
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -226,7 +226,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         }
 
         if ($this->collectionClass !== null) {
-            return $this->createCollection($this->collectionClass, $models);
+            return $this->createCollection($models);
         }
 
         return $models;
@@ -235,14 +235,13 @@ class ActiveQuery extends Query implements ActiveQueryInterface
     /**
      * Instantiates collection
      *
-     * @param string $class collection class to be instantiated and filled with results.
      * @param array $models array of models to fill collection with.
      * @return object collection instance.
      * @since 2.0.10
      */
-    protected function createCollection($class, $models)
+    protected function createCollection($models)
     {
-        return new $class($models);
+        return new $this->collectionClass($models);
     }
 
     /**
--- a/framework/db/ActiveQueryTrait.php
+++ b/framework/db/ActiveQueryTrait.php
@@ -30,6 +30,11 @@ trait ActiveQueryTrait
      */
     public $asArray;
 
+    /**
+     * @var string collection class to be instantiated and filled with results.
+     * @since 2.0.10
+     */
+    public $collectionClass;
 
     /**
      * Sets the [[asArray]] property.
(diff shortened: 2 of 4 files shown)


framework/db/ActiveQuery.php
```diff
@@ -225,6 +225,10 @@ public function populate($rows)
             }
         }
 
+        if ($this->collectionClass !== null) {
+            return new $this->collectionClass($models);
```
SilverFire: I'd suggest to move this line into a separate protected method. Then it could be overridden with custom logic.
Use case: I want to pass original `Query` object in my collection as a 2nd argument of the constructor.

SilverFire:
Are we going to implement `CollectionInterface`?

framework/db/ActiveQuery.php
samdark: Makes sense.

framework/db/ActiveQuery.php
```diff
@@ -225,10 +225,27 @@ public function populate($rows)
...
 
     /**
+     * Instantiates collection
+     *
+     * @param string $class collection class to be instantiated and filled with results.
+     * @param array $models array of models to fill collection with.
+     * @return object collection instance.
+     * @since 2.0.10
+     */
+    protected function createCollection($class, $models)
```
SilverFire: IMO `$models` is enough. Method can access `$this->collectionClass` by itself

klimov-paul:
> Are we going to implement CollectionInterface?
No since we aren't using collections.

I suppose we should create a `CollectionInterface` and at least a one basic implementation for it.
This is necessary thing to do in order to call this feture 'complete'. Otherwise the entire change should be dropped.

framework/db/ActiveQueryTrait.php
```diff
@@ -43,6 +48,18 @@ public function asArray($value = true)
     }
 
     /**
+     * Sets the [[collectionClass]] property.
+     * @param string $className collection class to be instantiated and filled with results.
+     * @return $this the query object itself
+     * @since 2.0.10
+     */
+    public function asCollection($className)
```
klimov-paul: See https://github.com/yiisoft/yii2/pull/12304#issuecomment-242328482

$className should be optional here, while default collection class is used.

framework/db/ActiveQuery.php
```diff
@@ -225,10 +225,26 @@ public function populate($rows)
...
     /**
+     * Instantiates collection
+     *
+     * @param array $models array of models to fill collection with.
+     * @return object collection instance.
+     * @since 2.0.10
+     */
+    protected function createCollection($models)
+    {
+        return new $this->collectionClass($models);
```
klimov-paul: DI should be used here:

```
Yii::createObject($this->collectionClass, $models);
```

framework/db/ActiveQueryTrait.php
samdark: Default is no collection at all. It doesn't make sense to introduce any default collection since the only thing you can achieve w/ collection that you can't achieve otherwise is type safety.

samdark:
> I suppose we should create a `CollectionInterface` and at least a one basic implementation for it.
This is necessary thing to do in order to call this feture 'complete'. Otherwise the entire change should be dropped.

Nope. First of all, it will prevent using external collection implementations such as Laravel's. Second, there's no requirements except that first argument of the constructor is an array of data.

klimov-paul:
> Default is no collection at all. It doesn't make sense to introduce any default collection since the only thing you can achieve w/ collection that you can't achieve otherwise is type safety.

As I already said then there is no sense to introduce the entire feature. What we are saving here?
Ability to write following code:

```php
$collection = User::find()->asCollection('UserCollection')->all();
```

insead of:
```php
$collection = new UserCollection(User::find()->all());
```

Make absolutely no sense to me.

Also what do you mean by `type safety`?

One more thing how this feature should behave for the `Query::one()`?

klimov-paul:
Also collections should be introduced at the level of the `Query` instead of `ActiveQuery`.

samdark:
The implementation here is incomplete proof of concept. You're absolutely correct that primary collection could be filled outside of AR and it's preferred to do so. The difference should be in handling relations...

By type safety I mean in an interface you'll be able to do:

```php
interface X
{
    public function doit(PostCollection $posts);
}

// instead of

interface X
{
    public function doit(array $posts);
}
```

samdark:
> One more thing how this feature should behave for the Query::one()?

`Query::one()` should behave exactly the same way as before.

klimov-paul:
> The difference should be in handling relations

In which way it will be different?

> By type safety I mean in an interface you'll be able to do:

So how the default collection class existance contradicts this ability? It is like telling the existance of the `BaseActiveRecord` class prevents you from specify type restiction like following:

```php
interface X
{
    public function doit(PostActiveRecord $post);
}
```

while it provides more flexisbility to write following type restiction:

```php
interface X
{
    public function doit(BaseActiveRecord $model);
}
```

> Query::one() should behave exactly the same way as before.

In this case `asCollection()` method makes absolutely no sense, since it affects only single method `all()`. Another 'finish' method should be introduced instead like `collection()`:

```
$postCollection = Post::find()->collection();
```

Note that `asArray()` method, which was used as prototype here, affects both `all()` and `one()` methods. - thus is is a separated method. For collection it make no sense.

samdark:
> In which way it will be different?

Well, relations could be automatically wrapped into collections.

klimov-paul:
By the way, what about `BatchQueryResult`?
All thus fuss about collection reminds the concept of data readers. This is actually the only use case, which make sense for me.

I agree with @cebe here: https://github.com/yiisoft/yii2/issues/10806#issuecomment-182892802

The collection functionality in Yii is split into different classes, providing all necessay support for most common features.

klimov-paul:
> relations could be automatically wrapped into collections.

For what purpose? Just to 'eat' more memory?

klimov-paul:
For the linking improvement we should use `ActiveQuery` as I said at https://github.com/yiisoft/yii2/issues/10806#issuecomment-242331322

```
$object->getRelationName()->link([$id1, $id2]);
```

samdark:
OK. I agree with most of your arguments. Seems it's better to approach it differently via enhancing `ActiveQuery`  and probably writing a recipe in Yii 2.0 cookbook about how to implement simple collections.

samdark:
https://github.com/samdark/yii2-cookbook/blob/master/book/structure-collections.md

cebe:
https://github.com/yiisoft/yii2-model-collection/issues/6

PR #12320, opened 2016-08-25
internals PL translation status

bizley:
Docs PL


SilverFire:
Is the work in progress?

bizley:
It's direct translation from `internals` folder (the same file). If it should stay it's already translated - but if not you should remove it from the folder here and in `internals` (together with README sections _Documentation_ > _Translation Status_).

PR #12323, opened 2016-08-25
Delete obsolete file

bizley:
In favour of translators.json. See #12321

PR #12321, opened 2016-08-25
internals PL translation teams

bizley:
BTW what's the status of this file? Is it up to date? How does it relate to translators.json for each language?

PR #12324, opened 2016-08-25
Obsolete translation teams removed

bizley:
In favour of translators.json. See #12321 #12323


cebe:
Imo translation-teams.md and translators.json have different purpose. The translation-team page has contact information which you can use if you want to contact translators or get involved, the translators.json has only the names for attribution in the rendered version.

bizley:
I agree.
Should the translation-team page be updated with current translators.json lists merged?

samdark:
Yes.

PR #12325, opened 2016-08-25
Obsolete translation teams removed

bizley:
Like in #12324


bizley:
Closing since it's one liner and thing is under discussion #12324

PR #12366, opened 2016-08-28
Cosmetic changes and typo,.. and issue

w3lifer:
In preview of the commit some part  of text is bold ...
I guess because of `__get()`/`__set()` ...
I don't know if it will be fine parsed ...

bold

### DIFF
--- a/docs/guide/db-active-record.md
+++ b/docs/guide/db-active-record.md
@@ -1424,7 +1424,7 @@ $rooms = Room::find()
     ->all();
 
 foreach ($rooms as $room) {
-    echo $room->volume, '<br>'; // contains value calculated by SQL
+    echo $room->volume; // contains value calculated by SQL
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guide/db-active-record.md
```diff
@@ -1418,19 +1418,19 @@ Then you need to compose a query, which calculates volume of the room and perfor
...
         '{{room}}.*', // select all columns
-        '([[length]] * [[width]].* [[height]]) AS volume', // calculate a volume
+        '([[length]] * [[width]] * [[height]]) AS volume', // calculate a volume
     ])
     ->orderBy('volume DESC') // apply sort
     ->all();
 
 foreach ($rooms as $room) {
-    echo $room->volume; // contains value calculated by SQL
+    echo $room->volume, '<br>'; // contains value calculated by SQL
```
SilverFire: I don't think this example is correct in terms of responsibility.

If this is code from view - then `Room::find()` is wrong, because views must not be responsible for data retentive from the database.
If this code is controller - then `<br>` is wrong, because controllers must not render HTML.

SilverFire:
Yes, double underscore makes text bold. That's ok

samdark:
Thanks!

PR #12369, opened 2016-08-28
The ability to add a parameters to a method of obtaining relation

Razzwan:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | no |
| Fixed issues |  |

Sometimes we use relation with params. As example:

``` php
class Document
{
    publick function getUserDocument($user_id)
    {
        return hasOne(UserDocument::class, ['document_id' => 'id'])
            ->andOnCondition(['user_id' => $user_id]);
    }
}
```

This relationship does not make sense without the parameter. But we can not specify this param when use `joinWith` method and instead use:

``` php
class Document
{
    publick function getUserDocuments()
    {
        return hasMany(UserDocument::class, ['document_id' => 'id']);
    }
}

class PlantController
{
    publick function actionGetUserDocuments()
    {
        $user_id = Yii::$app->getRequest()->getQueryParam('user_id');
        return Document::find()->joinWith(['userDocuments' => function ($q) {
            $q->andWhere(['user_id' => $user_id])
        }]);
    }
}
```

In the latter example, there are two problems: 
1. relationship has no useful effect, and will always be used only with the user parameter. 
2. the setting of the recording is quite complicated. 

The changes I propose to solve this problem:

``` php
class Document
{
    publick function getUserDocument($user_id)
    {
        return hasMany(UserDocument::class, ['document_id' => 'id'])
            ->andOnCondition(['user_id' => $user_id]);
    }
}

class PlantController
{
    publick function actionGetUserDocuments()
    {
        $user_id = Yii::$app->getRequest()->getQueryParam('user_id');
        return Document::find()
            ->joinWith('userDocuments')
            ->bindRelationParams(['user_id' => $user_id]);
    }
}
```

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -97,6 +97,10 @@ class ActiveQuery extends Query implements ActiveQueryInterface
      * @var array a list of relations that this query should be joined with
      */
     public $joinWith;
+    /**
+     * @var array
+     */
+    public $relationParams = [];
 
 
     /**
@@ -429,6 +433,12 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         return $this;
     }
 
+    public function bindRelationParams($params)
+    {
+        $this->relationParams = $params;
+        return $this;
+    }
+
     private function buildJoinWith()
     {
         $join = $this->join;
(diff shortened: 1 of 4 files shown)


cebe:
> publick function getUserDocument($user_id)

it is not possible to use this as a relation because it does not define something that is possible to express with SQL. It is not possible to load this relation in eager loading using 'with'.

I get the idea but your current approach is the correct approach that should be used instead of added this feature.

Razzwan:
> it is not possible to use this as a relation because it does not define something that is possible to express with SQL

Do you mean that the current implementation of the framework does not use the expression `adnOnCondition` in formatting of SQL? I do not understand.

cebe:
andOnCondition is used but it is not possible to implement eager loading with a dynamic onCondition. It must be fix per relation definition for that implementation to be able to generate an SQL query for retrieving records for eager loading.

cebe:
can not accept this for the reasons given, sorry.

PR #12381, opened 2016-08-29
enhanced BaseListView summary display with no result

taobig:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | yes |
| Tests pass? | yes |

if the dataProvider has no result, the summary div will be hidden. the outer list view div will move up.

### DIFF
--- a/framework/widgets/BaseListView.php
+++ b/framework/widgets/BaseListView.php
@@ -177,7 +177,7 @@ abstract class BaseListView extends Widget
     {
         $count = $this->dataProvider->getCount();
         if ($count <= 0) {
-            return '';
+            return Yii::t('yii', 'No results found.');;
         }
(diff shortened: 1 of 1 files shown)


cebe:
this text is already provided by the `$emptyText` property. With your change the text would appear twice.

taobig:
the $emptyText is not effective on summary div

cebe:
and it should not as the text is already displayed below the summary, why do you want to display it twice?

taobig:
As we konw, the summary-div is above of list-view div. If the dataProvider has no result, the summary-div will be hidden. the list-view div will up to the position of summary-div. To avoid going up, we can fill in the several words in the summary-div.

PR #12383, opened 2016-08-29
enhanced BaseListView summary display with no result

taobig:
#12381


cebe:
https://github.com/yiisoft/yii2/pull/12381#issuecomment-243376874

PR #12395, opened 2016-08-30
typos in en-US concept-di-container.md [skip ci]

jaaf:
| Q | A |
| --- | --- |
| Is bugfix? | yes/no |
| New feature? | yes/no |
| Breaks BC? | yes/no |
| Tests pass? | yes/no |


samdark:
Both are correct.

samdark:
Dependent class is the class that depends on something.

Though means something like "still" in this context.

jaaf:
Hi.
I agree on the idea that a dependent class is a class that depends on something (in this case on the dependency).
You are probably more clever than me in OOP but I thought that in the context on this sentence:
> The container will first inspect the class constructor to find out dependent class or interface names
> and then automatically resolve those dependencies recursively
what is passed to the constructor is the dependency. I also thought that the dependent class was the class that was being constructed.  I am probably confusing things. 

.

samdark:
"dependent class or interface names" means name of either dependent class or dependent interface. When you're getting something from container you're giving it this name. Dependencies of these classes are determined from dependent class constructor signature.

I know the language for dependency injection is rather vague but that's what's commonly used in this field.

PR #12398, opened 2016-08-30
fixed serializer to not convert objects

### LINKED ISSUE #12394: How to return empty object in json response?
override function fields() in my Model like this
```
public function fields()
{
    return [
        'test' => function () {
              return someFailedQuery() ?? new \stdClass();
        }
    ]
}
```

and return the Model in Controller but get json response like this
```
{
    "test": []
}
```

if I don't use Model fields but define a array by myself
```
public function actionTest()
{
   return  [
            'a' => new \stdClass(),
        ];
}
```
got response 
```
{
    "a": {}
}
```
this response really I need.

### What do you get instead?

i found the problem in ArrayableTrait::toArray()
ArrayHelper::toArray will convert stdClass to empty array

so how can i get the json empty object in response?

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.8
| PHP version      | 7.0.8
| Operating system | CentOS7

cebe:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | yes |
| Tests pass? | yes |
| Fixed issues | #12394 |

currently it will recursively convert objects to arrays.
This is the more expected way imo and would fix #12394.

It breaks BC however.

### DIFF
--- a/framework/rest/Serializer.php
+++ b/framework/rest/Serializer.php
@@ -242,7 +242,7 @@ class Serializer extends Component
             return null;
         } else {
             list ($fields, $expand) = $this->getRequestedFields();
-            return $model->toArray($fields, $expand);
+            return $model->toArray($fields, $expand, false);
         }
     }
 
--- a/tests/framework/rest/SerializerTest.php
+++ b/tests/framework/rest/SerializerTest.php
@@ -4,6 +4,7 @@ namespace yiiunit\framework\rest;
 
 use yii\base\Model;
 use yii\data\ArrayDataProvider;
+use yii\helpers\Json;
 use yii\rest\Serializer;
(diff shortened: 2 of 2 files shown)


samdark (approved):

SilverFire (approved):

klimov-paul:
Since `Json::encode()` converts `Arrayable` to array, model's processing will not change.
Thus it is more or less safe to apply this patch at 2.0.x. However UPGRADE note required for it anyway.

madsnedergaard:
What is status on this request? I am having the same problem, and this change fixes the issue.

samdark:
Won't be merged into 2.0.

PR #12402, opened 2016-08-30
tip to Url::current

githubjeka:
About merge current parameter with parameter of Search Model.

### DIFF
--- a/framework/helpers/BaseUrl.php
+++ b/framework/helpers/BaseUrl.php
@@ -369,6 +369,11 @@ class BaseUrl
      *
      * // /index.php?r=post%2Fview&id=100&src=google
      * echo Url::current(['id' => 100]);
+     * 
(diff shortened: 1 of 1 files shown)


samdark:
Umm... what's the goal of this example?

githubjeka:
It's my code:
```php
 \yii\helpers\Url::current(
        [
            Html::getInputName($filterModel, 'month') => $prevMonth->format('m'),
            Html::getInputName($filterModel, 'year') => $prevMonth->format('Y'),
         ]
)
```

This example doesn't corrected. To fix you need to delve into the source.(result of https://github.com/githubjeka/yii2/blob/e05a30c0e32da5b0bc2362ac229aa8aa73e02a7f/framework/helpers/BaseUrl.php#L394)

samdark:
OK so the issue is that array-parameters from GET are presented as arrays thus should be passed as arrays as well in order to be processed properly.

PR #12409, opened 2016-08-31
Cosmetic changes 2

w3lifer:


samdark:
Thanks!

framework/base/Module.php
```diff
@@ -239,7 +239,7 @@ public function getControllerPath()
 
     /**
      * Returns the directory that contains the view files for this module.
-     * @return string the root directory of view files. Defaults to "[[basePath]]/views".
+     * @return string the root directory of view files. Defaults to `[[basePath]]/views`.
```
cebe: this would break the link to `basePath` property.

framework/base/Module.php
```diff
@@ -261,7 +261,7 @@ public function setViewPath($path)
 
     /**
      * Returns the directory that contains layout view files for this module.
-     * @return string the root directory of layout files. Defaults to "[[viewPath]]/layouts".
+     * @return string the root directory of layout files. Defaults to `[[viewPath]]/layouts`.
```
cebe: this would break the link to `viewPath` property.

framework/base/Module.php
samdark: Why?

cebe: because with `"` the link defined by [[ will be parsed and transformed into a link. with this change [[ is within a code block and will not be parsed anymore.

samdark: Oh. I'll change these back then...

samdark: https://github.com/yiisoft/yii2/commit/dbe284c618a91550993a69004913cce2e7b2db38

w3lifer: Real example: http://www.yiiframework.com/doc-2.0/yii-i18n-phpmessagesource.html

`[[basePath]]/LanguageID/CategoryName.php`

cebe: 3159582 thanks!

PR #12418, opened 2016-08-31
Fix phpdoc for constants in db\Schema [skip ci]

rob006:
Phpdoc affects only one element - it should not be used for documenting group of constants.

244cb24d


samdark:
Indeed. Thanks!

PR #12419, opened 2016-09-01
Capability to hide root tag and model tag in xml responser

mhthnz:
### Usage:

``` php
        Yii::$app->response->formatters['xml'] = [
            'class' => \yii\web\XmlResponseFormatter::className(),
            'rootTag' => null,
            'objectTags' => false,
        ];        
        Yii::$app->response->format = \yii\web\Response::FORMAT_XML;
```

Very helpful for working with strict xml api.

### DIFF
--- a/framework/web/XmlResponseFormatter.php
+++ b/framework/web/XmlResponseFormatter.php
@@ -37,7 +37,7 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      */
     public $encoding;
     /**
-     * @var string the name of the root element.
+     * @var string the name of the root element or null.
      */
     public $rootTag = 'response';
     /**
@@ -50,7 +50,10 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      * @since 2.0.7
      */
     public $useTraversableAsArray = true;
-
+    /**
+     *  @var boolean mandatory addition of the tag of model name
+     */
+    public $modelTag = true;
 
     /**
      * Formats the specified response.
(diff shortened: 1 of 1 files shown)


samdark:
Overall it makes sense but I have no idea what "strict xml api" is. Could you elaborate?

mhthnz:
When working with XML API external service, API requires strict format answer. But forced insertion root tag, and model tag slightly hampered for this. And we have to use in controller:
```php
return $this->renderPartial('xml_response', ['model' => $model]);
```
Not more convenient than working with properties of model across Model:toArray()

framework/web/XmlResponseFormatter.php
```diff
@@ -50,7 +50,10 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      * @since 2.0.7
      */
     public $useTraversableAsArray = true;
-
+    /**
+     *  @var boolean mandatory addition of the tag of model name
+     */
+    public $modelTag = true;
```
samdark: Should be renamed to `objectTags` since objects aren't necessary models in this case.

framework/web/XmlResponseFormatter.php
```diff
@@ -50,7 +50,10 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      * @since 2.0.7
      */
     public $useTraversableAsArray = true;
-
+    /**
+     *  @var boolean mandatory addition of the tag of object name
```
samdark: Need `@since 2.0.11`.

samdark (changes requested):

samdark:
Closed in favor of #13214

PR #12428, opened 2016-09-01
Fixed bug in yii\db\mysql\QueryBuilder

DrmagicE:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |

In yii\db\mysql\QueryBuilder::insert() function,the $columns should be an array for the foreach.
Here is a simple fix that just turns the param into an array.

An example for reproducing the bug:

-- (without a primary key)
create table yiitest（
test int(11) default 1
）
//The code below will throw a warning

Yii::$db->createCommand()->insert('yiitest',[]);


klimov-paul:
CHANGELOG entry missing

DrmagicE:
@klimov-paul  well..seem i mess up again..I think i should try later after learning more about the workflow..

framework/CHANGELOG.md
```diff
@@ -3,7 +3,8 @@ Yii Framework 2 Change Log
 
 2.0.10 under development
 ------------------------
-
+- Bug #12428: Fixed `yii\db\mysql\QueryBuilder` causes warning when insert default rows into a table without primary key (DrmagicE)
+- Bug #12401: Fixed `yii\db\mysql\QueryBuilder` causes warning when insert default rows into a table without primary key (DrmagicE)
```
klimov-paul: Why 2 lines? And those with the 2 different issue numbers.

klimov-paul:
> well..seem i mess up again..I think i should try later after learning more about the workflow..

It almost fine: you just need one more commit to fix CHANGELOG file.

DrmagicE:
@kimov-paul I have fixed it.Hope it is right.Thanks for your patient!

klimov-paul:
Merged. Thank you.

PR #12434, opened 2016-09-02
phpDoc: some updates by null value

githubjeka:


samdark:
I'm not sure IDEs are OK about it. There were some problems with autocomplete before with such multi-typed hints.

rob006:
@samdark You don't get any autocomplete for string or array anyway, so there is nothing to lose here.

samdark:
I've meant inspections overall.

Issue #3047, opened 2014-04-10
Https on selected action.


bryglen:
Would it be great if we support on the core the https on selected action

Just like qiang written in the wiki. This is very useful when we have this on the core. 
http://www.yiiframework.com/wiki/407/url-management-for-websites-with-secure-and-nonsecure-pages/

klimov-paul:
https://github.com/yii2tech/https

samdark:
That's general enough to be in the core... @klimov-paul what do you think?

klimov-paul:
That is hard to tell. 
In general I am fine to move any `yii2tech` extension under `yiisoft` or to `yii2` itself.
However, speaking about this particular feature: what about `UrlNormalier`, which was added recently? Should not this feature be bound to it somehow?

samdark:
Yes. It may be a good fit.

dynasource:
closed in favor of #3047

samdark:
It is #3047...

dynasource:
closed in favor of #12691

rob006:
@dynasource How #12691 solve this?

dynasource:
The reason for closing it was to prevent us from having 2 similar discussions around https & UrlManager. As you are declaring it be a separate issue, I will reopen it again. Thanks!

PR #12451, opened 2016-09-03
Improvements for handling redirections produced by UrlNormalizer

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no (if merged before 2.0.10 release) |
| Tests pass? | yes |
| Fixed issues | - |

replaces #12450

I was wondering how to solve https://github.com/yiisoft/yii2/issues/3047, and I realized that the current implementation of redirections produced by UrlNormalizera is very limited. It should be improved before 2.0.10 will be released, to avoid BC trap.

### DIFF
--- a/framework/web/Application.php
+++ b/framework/web/Application.php
@@ -8,6 +8,7 @@
 namespace yii\web;
 
 use Yii;
+use yii\helpers\Url;
 use yii\base\InvalidRouteException;
 
 /**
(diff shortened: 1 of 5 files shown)


samdark:
What's the idea behind this change?

rob006:
Without this there is no simple way to force https on redirection or redirect to custom URL. These limitations can be very problematic when extending `UrlNormalizer` or using `UrlNormalizer::$action` with callback with custom redirection.

samdark:
OK. Makes sense.

samdark:
Thanks!

Issue #12462, opened 2016-09-05
Docs: Customizing Query Classes could be improved


CedricYii:
There is a small topic on Customizing Query Classes in the guide:
http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#customizing-query-classes

But some points remain unclear to me.

For example, it looks like the guide is advising to write the class Comment and the class CommentQuery in the same file Comment.php. Is it intentional? Then it should be explained and written explicitly.

Another good issue is how to have a model with a default custom query (can be done with init() in the custom query class).
And the most complex and useful information is how to have a model extending another model with a default custom query.
These issues have been explained by @xVismutx in https://github.com/yiisoft/yii2/issues/7869

But it would be really better to find in the guide directly (rather than in a "hard to find" closed issue)!

cebe:
> For example, it looks like the guide is advising to write the class Comment and the class CommentQuery in the same file Comment.php. Is it intentional? Then it should be explained and written explicitly.

no that should not be in the same file, because autoloading would not work anymore. fixed: 5bdca129a4a59ea6074118139983c642a347aaf7

About the other issues, the problem is very specific so I am not sure how to document the solution in a generic way in the guide. maybe a wiki entry is better suited for this?

CedricYii:
Thank you for clarifying this.
It is better to have all information needed in one place (so maybe adding a detail section would then be preferable).
Working with a common model uses by several frontends extending this model is quite common.
I have proposed a new version from which I think I could have guessed the way to do it.

PR #12464, opened 2016-09-05
Guide: Customizing Query Classes

CedricYii:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues | #12462 |

Clearly separate classes' definition.
Add information for default query condition with init().
Suggest using andOnCondition() instead of andWhere() as it didn't work
with the left join example provided.

### DIFF
--- a/docs/guide/db-active-record.md
+++ b/docs/guide/db-active-record.md
@@ -1349,13 +1349,15 @@ class Comment extends ActiveRecord
         return new CommentQuery(get_called_class());
     }
 }
-````
+```
+
 Now whenever you are performing a query (e.g. `find()`, `findOne()`) or defining a relation (e.g. `hasOne()`)
 with `Comment`, you will be calling an instance of `CommentQuery` instead of `ActiveQuery`.
 
 You can customize the `CommentQuery` class in many creative ways to improve your query building experience. For example,
 you can define new query building methods:
-````php
+
+```php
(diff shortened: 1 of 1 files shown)


CedricYii:
I just realized that I missread `$customers = Customer::find()->with('activeComments')->all();` is ok with andWhere().
But it wouldn't with `$customers = Customer::find()->joinWith('activeComments')->all();`

docs/guide/db-active-record.md
```diff
@@ -1350,41 +1349,44 @@ class Comment extends ActiveRecord
         return new CommentQuery(get_called_class());
     }
 }
+````
+Now whenever you are performing a query (e.g. `find()`, `findOne()`) or defining a relation (e.g. `hasOne()`)
+with `Comment`, you will be calling an instance of `CommentQuery` instead of `ActiveQuery`.
 
+You can customize the `CommentQuery` class in many creative ways to improve your query building experience. For example,
+you can define new query building methods:
+````php
```
samdark: Need 3 ticks, not 4 for code blocks. Also there should be empty line before code.

CedricYii: Thank you, I didn't notice. Fixed.

samdark:
Thanks!

Issue #6996, opened 2015-01-24
[REST API] PUT method and UploadedFile


Claud:
Hello.
If used for data update "PUT" method, then UploadedFile does not work.

fernandezekiel:
what Content-type are you using for your request?

Claud:
multipart/form-data

fernandezekiel:
multipart/form-data only works with POST

fernandezekiel:
https://bugs.php.net/bug.php?id=55815

klimov-paul:
Depends on #12463

klimov-paul:
Solution proposed at #12466

klimov-paul:
Resolved by commit  4e59607

PR #12466, opened 2016-09-05
`MultipartFormDataParser` added

klimov-paul:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #6996 |


klimov-paul:
Successfully tested agains following request:

```
curl "http://dev54.quartsoft.com/pklimov/yii2test/site/test" -X PUT -F "photo=@composer.json;type=text/plain" -F "foo=bar"
```

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+ *                 'multipart/form-data' => 'yii\web\MultipartFormDataParser'
+ *             ],
+ *         ],
+ *         // ...
+ *     ],
+ *     // ...
+ * ];
+ * ```
+ *
+ * Method `parse()` of this parser automatically populates `$_FILES` with the files parsed from raw body.
```
SilverFire: <code>\`parse()\`</code> -> [[parse()]]

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+ *
+ * Note: since this is a request parser, it will initialize `$_FILES` values on [[Request::getBodyParams()]].
```
SilverFire: `Note: ...`
-> 

```
> Note: 
```

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+ * Until this method is invoked `$_FILES` array will remain empty event if there are submitted files in the
```
SilverFire: Comma after `invoked` is missing

SilverFire: event -> even

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+ *
+ * $restRequestData = Yii::$app->request->getBodyParams();
+ * $uploadedFile = UploadedFile::getInstancesByName('photo');
+ *
+ * $model = new Item();
+ * $model->populate($restRequestData);
+ * copy($uploadedFile->tempName, '/path/to/file/storage/photo.jpg');
+ * ```
+ *
+ * Note: although this parser fully emulates regular structure of the `$_FILES`, related temporary
```
SilverFire: `Note: ...`
-> 

```
> Note: ...
```

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+ * files, which are available via 'tmp_name' key, will not be recognized by PHP as uploaded ones.
```
SilverFire: `'tmp_name'` -> <code>\`tmp_key\`</code>

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+            $headers = $this->parseHeaders($headers);
+            
+            if (!isset($headers['content-disposition']['name'])) {
+                continue;
+            }
+
+            if (isset($headers['content-disposition']['filename'])) {
+                // file upload:
+                if ($filesCount >= $this->getUploadFileMaxCount()) {
+                    continue;
```
SilverFire: As for me, this case should not silenced. Exception must be thrown

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+namespace yii\web;
+
+use yii\base\Object;
+use yii\helpers\ArrayHelper;
+use yii\helpers\StringHelper;
+
+/**
+ * MultipartFormDataParser parses content encoded as 'multipart/form-data'.
+ * This parser provides the fallback for the 'multipart/form-data' processing on non POST requests,
+ * for example: the one with 'PUT' request method.
```
SilverFire: `'PUT'` -> <code>\`PUT\`</code>

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+            return [];
+        }
+
+        if (!preg_match('/boundary=(.*)$/is', $contentType, $matches)) {
+            return [];
+        }
+        $boundary = $matches[1];
+
+        $bodyParts = preg_split('/-+' . preg_quote($boundary) . '/s', $rawBody);
+        array_pop($bodyParts); // last block always has no data
```
samdark: You can pass `PREG_SPLIT_NO_EMPTY` as a flag to `preg_split` so it won't return empty elements.

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+
+        $bodyParams = [];
+        $filesCount = 0;
+        foreach ($bodyParts as $bodyPart) {
+            if (empty($bodyPart)) {
+                continue;
+            }
+            list($headers, $value) = explode("\r\n\r\n", $bodyPart, 2);
```
samdark: Is it always `\r\n\r\n`?

framework/web/MultipartFormDataParser.php
samdark: Currently it's on par with PHP behavior but I agree that exception is a bit more developer-friendly.

samdark: Should it be `>` instead of `>=`?

framework/web/MultipartFormDataParser.php
```diff
@@ -0,0 +1,357 @@
...
+
+    /**
+     * Parses content part headers.
+     * @param string $headerContent headers source content
+     * @return array parsed headers.
+     */
+    private function parseHeaders($headerContent)
+    {
+        $headers = [];
+        $headerParts = preg_split("/[\n|\r]/s", $headerContent);
```
samdark: You can pass `PREG_SPLIT_NO_EMPTY` as a flag to `preg_split` so it won't return empty elements.

samdark: There's `\R` meta-character which is better to use in this case.

tests/framework/web/MultipartFormDataParserTest.php
```diff
@@ -0,0 +1,87 @@
...
+
+use yii\web\MultipartFormDataParser;
+use yiiunit\TestCase;
+
+class MultipartFormDataParserTest extends TestCase
+{
+    public function testParse()
+    {
+        if (defined('HHVM_VERSION')) {
+            static::markTestSkipped('Can not test on HHVM because it does not support proper handling of the temporary files.');
```
samdark: Any details? There should be a workaround.

framework/web/MultipartFormDataParser.php
klimov-paul: No it is correct: remember how enty in `$_FILES` looks like.

framework/web/MultipartFormDataParser.php
klimov-paul: Yes, it is always so: a single skipped line.

framework/web/MultipartFormDataParser.php
klimov-paul: > As for me, this case should not silenced. Exception must be thrown
> Currently it's on par with PHP behavior but I agree that exception is a bit more developer-friendly.

This is a same way PHP processes such request. Throwning exception here will make behavior incosistent with regular POST.

klimov-paul: > Should it be > instead of >=?

No, condition is correct, see `max_file_uploads` PHP ini directive.

tests/framework/web/MultipartFormDataParserTest.php
klimov-paul: I do not have an HHVM environment and thus I am unable to track the problem or solve it.
All I can tell it fails on Travis.

framework/web/MultipartFormDataParser.php
klimov-paul: Why? It is a simple string constant it is not some PHP type or expression.

framework/web/MultipartFormDataParser.php
klimov-paul: No, it does not work at this case. I am not sure why, but it fails with `PREG_SPLIT_NO_EMPTY`.

framework/web/MultipartFormDataParser.php
klimov-paul: fixed

framework/web/MultipartFormDataParser.php
klimov-paul: fixed

framework/web/MultipartFormDataParser.php
klimov-paul: fixed

framework/web/MultipartFormDataParser.php
klimov-paul: fixed

framework/web/MultipartFormDataParser.php
klimov-paul: Fixed

framework/web/MultipartFormDataParser.php
SilverFire: The only difference is grave instead of apostrophe

framework/web/MultipartFormDataParser.php
SilverFire: @klimov-paul agreed with all

framework/web/MultipartFormDataParser.php
SilverFire: Just to be consistent

https://github.com/yiisoft/yii2/blob/master/framework/widgets/ActiveForm.php#L33
https://github.com/yiisoft/yii2/blob/master/framework/filters/AccessRule.php#L60

klimov-paul: At `ActiveForm` values `post`, `get` and so on are PHP expressions describing the possible field value, while here is a general text related to HTTP request raw source.

framework/web/MultipartFormDataParser.php
klimov-paul: I see. Fixed

framework/web/MultipartFormDataParser.php
samdark: I'm still not sure it's correct. It will parse no more than 3 files if `max_file_uploads` is 4, right?

framework/web/MultipartFormDataParser.php
samdark: I've meant that a single skipped line could be `\n\n` as well. Are you sure it's always `\r\n\r\n` in the case?

samdark: Anyway, it's better to use `\R` and a `preg_split`.

framework/web/MultipartFormDataParser.php
klimov-paul: I have added unit tests to verify both 'max size' and 'max count' options works correctly.

framework/web/MultipartFormDataParser.php
klimov-paul: I do remember PHP unable to parse content in other format. As I have already worked with it at `yii2-httpclient`.
Still, fine I fix to `preg_split()` with double `\R`.

PR #12472, opened 2016-09-06
message/extract - support for multiple sourcePath

ptheofan:
Added support for multiple paths as `sourcePath` for the `console:message/extract`. Because sometimes it makes more sense to add 2 or 3 sourcePaths instead of adding 25 except cases :)

| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |

### DIFF
--- a/framework/console/controllers/MessageController.php
+++ b/framework/console/controllers/MessageController.php
@@ -282,12 +282,22 @@ EOD;
             $configFileContent,
             $this->getPassedOptionValues()
         );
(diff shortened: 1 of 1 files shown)


klimov-paul:
Feature rejected. We will not support 2 different ways for doing the same thing.

> Because sometimes it makes more sense to add 2 or 3 sourcePaths instead of adding 25 except cases

I am afraid this case is not general as well as risk prone: once you add new folder to the project root with translations you'll need to reedit config, while `except` value is more general and static

PR #12490, opened 2016-09-08
Added missing hyphen

Mak-Di:


SilverFire:
Thank you! It is good to search for the occurrences of the same mistake in other files and fix them as well:

```bash
yiisoft/yii2/docs# grep -irEn "(server|client) side" .                                                                                                                                
./guide/start-forms.md:211:Yes, the data validation is initially done on the client side using JavaScript, and secondarily performed on the server side via PHP.
./guide/start-forms.md:214:JavaScript on your browser, the validation will still be performed on the server side, as shown in
./guide/input-validation.md:462:client and server sides; the other uses the `validatePassword` inline validator which is only supported on the server
./guide/input-validation.md:525:that performs the validation on the client side. Within the JavaScript code, you may use the following
./guide/input-validation.md:534:against the existing status data. The validator supports both server side and client side validation.
./guide/input-validation.md:607:is hit. The following example shows how to validate the dimensions of an uploaded image file on the client side.
./guide/input-validation.md:663:Some validations can only be done on the server side, because only the server has the necessary information.
./guide/input-validation.md:664:For example, to validate if a username is unique or not, it is necessary to check the user table on the server side.
./guide/caching-http.md:8:actions whose rendering result may be cached on the client side. [[yii\filters\HttpCache|HttpCache]]
./guide/caching-http.md:55:the server will not re-generate the page, and the browser will use the cached version on the client side.
./guide/caching-http.md:62:is changed, the hash will be changed as well. By comparing the hash kept on the client side with the hash
./guide/caching-http.md:100:the server will not re-generate the page, and the browser will use the cached version on the client side.
./guide/runtime-sessions-cookies.md:306:the risk of a client side script accessing the protected cookie (if the browser supports it). You may read
./guide/runtime-sessions-cookies.md:314:on the client side. This is achieved by signing each cookie with a hash string, which allows the application to
./guide/runtime-sessions-cookies.md:315:tell if a cookie has been modified on the client side. If so, the cookie will NOT be accessible through the
./guide/caching-overview.md:11:of the most recent articles. On the client side, HTTP caching may be used to keep most recently visited page content in
./guide/runtime-responses.md:192:the redirection URL as its value. On the client side, you may write JavaScript code to read this header value and
./guide/tutorial-core-validators.md:114:Client validation is not enabled as this will only work on the server side because the date validator currently does not
./guide/caching-page.md:4:Page caching refers to caching the content of a whole page on the server side. Later when the same page
```

Fixed in 49448426

Issue #9469, opened 2015-08-19
namespace for migrations


lorinczelod:
It would be useful to add a namespace to migrations. For example:
I want to apply the `vendor/yiisoft/yii2/rbac/migrations/m140506_102106_rbac_init.php` migration.
For that now I need to run: ``yii migrate --migrationPath=@yii/rbac/migrations``
But would it be more convenient, if I made a local 'reference':
My migration in `console/migrations/m150819_075235_rbac_init.php` would be:
``class m150819_075235_rbac_init extends \yii\rbac\migrations\m140506_102106_rbac_init`` { }
In this case the simple `yii migration` would install that 'custom' migration, too.

lorinczelod:
now I am using this working method to workaround the problem:
```
<?php
$alias = Yii::getAlias('@yii/rbac/migrations/m140506_102106_rbac_init.php');
Yii::$classMap['m140506_102106_rbac_init'] = $alias;

class m150819_075235_rbac_init extends m140506_102106_rbac_init { }
```
PS. my issue title is probably wrong, as it gives reference to the resolve, not to the problem. Please feel free to change.

klimov-paul:
Why you don't use old good `require_once` ?

lorinczelod:
[klimov-paul](https://github.com/klimov-paul), you have right!

klimov-paul:
Resolved by commit 8aa0e85

Issue #11096, opened 2016-03-11
Enhancement: PSR-2 style named migrations.


davidjeddy:
**Problem:**
 + Migration classes are named using a different convention than other classes in the framework

**Example:**

    yii migrate/mark 150101_185401                      # using timestamp to specify the migration
    yii migrate/mark "2015-01-01 18:54:01"              # using a string that can be parsed by strtotime()
    yii migrate/mark m150101_185401_create_news_table   # using full name
    yii migrate/mark 1392853618                         # using UNIX timestamp

Suggestion is to allow PSR2 style names:

    yii migration/mark m150101185401CreateNewsTable   # PSR2 compliant file and class name

**Impact:**
 + Linters will not fail when encountering migration classes
 + Global consistency in class naming convention

cesarnicola:
I think I can do this change, PSR2 should be the new default naming format, right?

samdark:
Yes.

davidjeddy:
Thank you both very much!

cesarnicola:
Will use "M" instead of "m" as the prefix to follow the upper-camel convention.

Also, should the create command raise a warning/error if the name does not met the convention?

Thanks!

Faryshta:
I still think its better  to have the migration names to be something like `console\migrations\CreatePost` and have a dependency property to know which migrations should be executed first.

```php
namespace common\migrations;

class CreatePost extends Migration
{
    public function up()
    {
    }

    public function down()
    {
    }

    $depends = [
        'common\\migrations\\Init';
        'common\\migrations\\CreatePostCategory';
    ];
}
```

samdark:
@Faryshta this way last migration out of 100 will have 99 dependencies. That's no good.

samdark:
@cesarnicola yes, a warning would be good. Also don't forget namespaces.

cesarnicola:
Thanks, working on it.

Faryshta:
> this way last migration out of 100 will have 99 dependencies.

Nested migrations like in a fixture or an asset. If migration Z depends on Y and X, then it will check the dependencies of Y and X.

The last migration might depend on a 100 migrations but the code will be much simplier.

klimov-paul:
Solution proposed at #12511

klimov-paul:
Resolved by commit 8aa0e85

Issue #9698, opened 2015-09-16
Migration (--migrationPath)


vhBushi:
hi all, sorry for any inconvenience
when making a migration indicating the migrationPath, for example:

>$./yii migrate --migrationPath=modules/foo/migrations

yii does not save the migration path used in migration. So when I run the command down, I have to specify the **migrationPath** for each migration or grup. but when many migrations, this becomes a problem.

when i run the command without the parameter migrationPath in a migration that is not in the default directory y get the exception **failed to open stream: No such file or directory'** obviously 

with many migrations and modules do this is difficult, I think yii should store the migrationPath used for each migration.

something like change the migration table to :
```
CREATE TABLE migration (
        version varchar(180) PRIMARY KEY,
        path varchar(x),
        apply_time integer,
  )
```

if there is a way to do it, let me note. thk team :D

samdark:
There's currently no way doing it automatically.

mdmunir:
Alternative https://github.com/deesoft/yii2-console
Add your migration paths as application params
```php
// params.php
return [
    'dee.migration.path' => [
        '@yii/rbac/migrations',
        '@mdm/autonumber/migrations',
        '@mdm/upload/migrations',
    ]
];
```

mubushi:
modify the database would be the best way? and change the getMigrationHistory() (in yii\console\controllers\MigrationController) to modify the basePath Attribute based on the value of the database

pana1990:
maybe "db" should be include too. it is useful when i have two or more dbms.

@samdark @mubushi @mdmunir  what do you think about it? i can create pull request for this issue.

samdark:
I think the best way to solve it is to create bash script that runs a couple of commands specifying necessary `--migrationPath`. Are there any pros implementing it at console command leve?

pana1990:
if you want to use " migrate/redo all or migrate/down all"?

samdark:
Why?

pana1990:
how @mubushi  said 

"when i run the command without the parameter migrationPath in a migration that is not in the default directory y get the exception failed to open stream: No such file or directory' obviously "

if you want to run migrate/up you have create bash script, if you want to run migrate/down you have create another bash script again.

when you have many migrations and modules do this is difficult, i think to this feature should be include out of the box.

samdark:
Not really another. It's about adding more lines to existing `migrate_down.sh` or `migrate_up.sh`. Same amount of lines as would end up in config, I guess..

[...]

Issue #384, opened 2013-05-23
improve migrate command


schmunk42:
* running migrations from different locations (modules)

more or less like this one from @cebe https://github.com/yiiext/migrate-command

vladim1:
I'll leave this post here to not create the same one.
It's really pain in the ass to run migrations of the plugged in modules with specifiing their pathes.
This issue is almost 2 years.  I am waiting for this feature from milestone to milestone. What is the problem with solving it? So, we could discuss it...
PS: Is there any tool/extension to automate the process of running migrations of plugged in modules?

schmunk42:
@vladim1 Check out our extension https://github.com/dmstr/yii2-migrate-command

Faryshta:
I think #8202 is a good way to fix this issue for modules, extensions, external libraries and all the other cases discused here.

jacmoe:
Since I include migrations in my test routines - ie migrations should be able to migrate all the way down and all the way up at any time - I am looking forward to a built-in solution. :+1: 
The https://github.com/dmstr/yii2-migrate-command is part of my standard tool-kit already, but it would make Yii even more cool if it was built in. Especially since newcomers wouldn't have to discover it themselves. :)

klimov-paul:
I whould solve this matter using namespaces instead of paths, just I have proposed here:
https://github.com/yiisoft/yii2/issues/9698#issuecomment-147696833

schmunk42:
@klimov-paul We're currently using aliases, which is essentially the same, isn't it? https://github.com/yiisoft/yii2/issues/384#issuecomment-63638791

ItsReddi:
I discovered this issue before two years. Well the eta was RC. Any chance that this feature will be implemented in the next release or should we stick to some extensions that can handle this approach?

klimov-paul:
Solution proposed at #12511

klimov-paul:
Resolved by commit 8aa0e85

PR #12511, opened 2016-09-10
Add migration namespace support

klimov-paul:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #384, #9698, #11096, #9469 |

Added `migrationNamespaces` allowing to specify several namespaces to be used.
For the namespaced migrations PSR-2 naming applied.
Fully backward compatible: if you specify no `migrationNamespaces` command works as before.


framework/console/controllers/BaseMigrateController.php
```diff
@@ -476,23 +523,77 @@ public function actionNew($limit = 10)
...
+     * @since 2.0.10
+     */
+    private function generateClassName($name)
+    {
+        $namespace = null;
+        $name = trim($name, '\\');
+        if (strpos($name, '\\') !== false) {
+            $nameParts = explode('\\', $name);
+            $name = array_pop($nameParts);
+            $namespace = implode('\\', $nameParts);
```
SilverFire: Could be done without arrays

```php
$namespace = substr($name, strrpos($name, '\\') + 1);
$name = substr($name, 0, strrpos($name, '\\'));
```

framework/console/controllers/BaseMigrateController.php
```diff
@@ -624,25 +727,42 @@ protected function migrateToVersion($version)
...
         $applied = [];
-        foreach ($this->getMigrationHistory(null) as $version => $time) {
-            $applied[substr($version, 1, 13)] = true;
+        foreach ($this->getMigrationHistory(null) as $class => $time) {
+            $applied[trim($class, '\\')] = true;
+        }
+
+        $migrationPaths = [];
+        if (!empty($this->migrationPath)) {
+            $migrationPaths[''] = $this->migrationPath;
```
SilverFire: Maybe, it's better to do this operation in `init()` method and mark `migrationPath` property as deprecated to drop it in 2.1.x?

framework/console/controllers/BaseMigrateController.php
klimov-paul: Fixed

framework/console/controllers/BaseMigrateController.php
klimov-paul: At the present state you can use both namespaced and not namespaced migrations at the same time.
It provides the support for the existing old projects.

We could mark `migrationPath` as deprecated, but this should be disscussed separately.

Composing migration paths at `init()` level is not necessary, I prefer to keep it this way at least for now.

framework/console/controllers/BaseMigrateController.php
```diff
@@ -33,9 +34,31 @@
     /**
      * @var string the directory storing the migration classes. This can be either
      * a path alias or a directory.
+     *
+     * You may set this field to `null` in case you have setup [[migrationNamespaces]] in order
```
SilverFire: `setup` -> `set up`

klimov-paul: fixed

framework/console/controllers/BaseMigrateController.php
```diff
@@ -292,7 +326,10 @@ public function actionRedo($limit = 1)
      */
     public function actionTo($version)
     {
-        if (preg_match('/^m?(\d{6}_\d{6})(_.*?)?$/', $version, $matches)) {
+        if (preg_match('/^\\\\?([\w_]+\\\\)+m(\d{6}_?\d{6})(\D.*?)?$/is', $version, $matches)) {
```
samdark: If `\` is trimmed afterwards it makes sense to allow any number of slashes before and after.

samdark: Both regexps are repeated multiple times. Worth moving to class level constants.

framework/console/controllers/BaseMigrateController.php
```diff
@@ -465,8 +505,19 @@ public function actionNew($limit = 10)
      * yii migrate/create create_user_table
      * ```
      *
+     * In order to generate migration with namespace, you should add it before migration self name.
```
samdark: In order to generate namespaced migration you should specify namespace before migration's name.

framework/console/controllers/BaseMigrateController.php
```diff
@@ -476,23 +527,75 @@ public function actionNew($limit = 10)
...
                 'className' => $className,
+                'namespace' => $namespace,
             ]);
             file_put_contents($file, $content);
             $this->stdout("New migration created successfully.\n", Console::FG_GREEN);
         }
     }
 
     /**
+     * Generates class name with possible namespace from migration name.
```
samdark: Description isn't clear.

framework/console/controllers/BaseMigrateController.php
```diff
@@ -476,23 +527,75 @@ public function actionNew($limit = 10)
...
+     * @param string $name migration name from user input.
+     * @return array list of 2 elements: 'namespace' and 'class self name'
```
samdark: What's "class self name"? Name of the class itself>

SilverFire:
Closes #9469

framework/console/controllers/BaseMigrateController.php
klimov-paul: 2 methods extracted

framework/console/controllers/BaseMigrateController.php
klimov-paul: fixed

framework/console/controllers/BaseMigrateController.php
klimov-paul: Fixed

framework/console/controllers/BaseMigrateController.php
klimov-paul: Fixed

rob006:
I there any practical reason for using `M101129185401CreateUserTable` as migration name instead of `m101129_185401_create_user_table`?

klimov-paul:
> I there any practical reason for using M101129185401CreateUserTable as migration name instead of m101129_185401_create_user_table?

See #11096

nkostadinov:
I pulled the last changes and tried to use migrations but it fails.

When I put a migration namespace like this "insight/oauth/migrations" the command finds the migration like this "insight/oauth/migrations\m140501_075311_add_oauth2_server" but when I try to migrate gives me this error: 
````
PHP Fatal error:  Class 'insight/oauth/migrations\m140501_075311_add_oauth2_server' not found
````
If I set it like this "insight\oauth\migrations" e.g. like in the docs I get 
````
Exception 'yii\base\InvalidParamException' with message 'Invalid path alias: @insight\oauth\migrations'
````

BTW this is already applied migration. How exactly is the proper way to use namespaced migration and why it shows already applied migrations ?

**Comment by klimov-paul** [comment]
> When I put a migration namespace like this "insight/oauth/migrations"

Namespaces uses backslash (`\`) as separator not the slash ('/').

> If I set it like this "insight\oauth\migrations" e.g. like in the docs I get

See
https://github.com/yiisoft/yii2/blob/master/framework/console/controllers/BaseMigrateController.php#L45-L47
Obviously you are uning namespace, which has not path alias refers to it.

> How exactly is the proper way to use namespaced migration and why it shows already applied migrations ?

Setup migration namespaces use `yii migrate/create app\migrations\newMigration`, edit generated file and apply it.

**Comment by nkostadinov** [comment]
The alias I use has a path alias which refers to it, I used deesoft/yii2-console till now which uses array of path aliases, I just converted it to namespaces. I know how to write correctly namespaces but when i use it Yii cannot find the migration, so I tried both ways. It seems that it just doesnt convert "\" to "/".

@klimov-paul I tried what you suggested but again I get error 
````
>yii migrate/create app\migrations\newMigration
Yii Migration Tool (based on Yii v2.0.10-dev)

Exception 'yii\base\InvalidParamException' with message 'Invalid path alias: @app\migrations'
````

**Comment by nkostadinov** [comment]
@klimov-paul I`m using windows and debugged the problem. My guess is that the constant DIRECTORY_SEPARATOR which in windows is "\" is properly interpreted as path alias in yii.

**Comment by klimov-paul** [comment]
Resolved by 7acd3b6400b2d7f78600fbb9b4e1c2cf9207e901

File: framework/console/controllers/BaseMigrateController.php [hunk 8]
```diff
@@ -365,6 +404,34 @@ public function actionMark($version)
...
 
     /**
+     * Checks if given migration version specification matches namespaced migration name.
+     * @param string $rawVersion raw version specification received from user input.
+     * @return string|false actual migration version, `false` - if not match.
+     * @since 2.0.10
+     */
+    private function extractNamespaceMigrationVersion($rawVersion)
+    {
+        if (preg_match('/^\\\\?([\w_]+\\\\)+m(\d{6}_?\d{6})(\D.*?)?$/is', $rawVersion, $matches)) {
```
cebe: Why is the last part `(\D.*?)` non-greedy? Should be `(\D.*)`

framework/console/controllers/BaseMigrateController.php
```diff
@@ -365,6 +404,34 @@ public function actionMark($version)
...
+
+    /**
+     * Checks if given migration version specification matches migration base name.
+     * @param string $rawVersion raw version specification received from user input.
+     * @return string|false actual migration version, `false` - if not match.
+     * @since 2.0.10
+     */
+    private function extractMigrationVersion($rawVersion)
+    {
+        if (preg_match('/^m?(\d{6}_?\d{6})(\D.*?)?$/is', $rawVersion, $matches)) {
```
cebe: same here.

framework/console/controllers/BaseMigrateController.php
klimov-paul: It just follows the original implementaton. It is enough to specify version timeline using numbers, instead of writing full class name.
Can not be changed otherwise it will cause BC break.

framework/console/controllers/BaseMigrateController.php
klimov-paul: 239cdcdf8af5bf669f975db81f80320f8373ee4c

PR #12531, opened 2016-09-13
 added input-file-upload.md FR [skip ci]

jaaf:
| Q | A |
| --- | --- |
| Is bugfix? | yes/no |
| New feature? | yes/no |
| Breaks BC? | yes/no |
| Tests pass? | yes/no |

### DIFF
--- a/docs/guide-fr/concept-components.md
+++ b/docs/guide-fr/concept-components.md
@@ -1,7 +1,7 @@
 Composants
 ==========
 
-Les composants sont les blocs de constructions principaux de vos applications Yii. Les composants sont des instances de la classe [[yii\base\Component]],
+Les composants sont les blocs de constructions principaux de vos applications Yii. Les composants sont des instances de la classe [[yii\base\Component]](http://denentzat.fr),
 ou de ses classes filles. Les trois fonctionnalités principales fournies par les composants aux autres classes sont :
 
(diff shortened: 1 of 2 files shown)


docs/guide-fr/concept-components.md
```diff
@@ -1,7 +1,7 @@
 Composants
 ==========
 
-Les composants sont les blocs de constructions principaux de vos applications Yii. Les composants sont des instances de la classe [[yii\base\Component]],
+Les composants sont les blocs de constructions principaux de vos applications Yii. Les composants sont des instances de la classe [[yii\base\Component]](http://denentzat.fr),
```
samdark: Hmm... that's not what it should be linked to.

jaaf:
I apologize. I was just wondering what is [[ ]] in markdown and what is the difference with [ ]( ) so I made some trial to see the result. After that I forgot to put the file back to its original state.
BTW could you tell me how the link mechanism works in this case (not url explicitly given)  and what happens if the file is used on an other site?
Thank you for finding the mistake.

samdark:
We have our own markdown extensions on top of the GitHub-flavored one. `[[className]]` is one of those.

samdark:
It won't work well on other websites. Will be displayed as is.

jaaf:
Ok, I see.  Thanks.

PR #12548, opened 2016-09-16
Schema cache cut out to simple component.

zalatov:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

Schema cache cut out to simple component.
Now it can replace to custom implementation.
Backward compatibility is available (old implementations marked as deprecated).

---

Кэширование схемы таблиц перенесено в отдельный компонент.
Теперь появилась возможность задать свою реализацию кэширования.
Изменения не ломают обратную совместимость (старая реализация помечена как deprecated).


cebe:
Can you explain why you need this and what a custom implementation would do?

zalatov:
We are use custom TagDepencency and extend Cache functions (include statistic and debug-panel), so we are need to cache schema via own components.

For example, we don't invalidate cache when tag is out of date - we only mark it as older.

But we can do it because Yii use hard code dependency for caching schema.

P.S.:
May be it's a old error, but only schema cache use cache key name as array - it's not needed strong.
This "feature" required additional support in code, which is not functional needed.

samdark:
@zalatov tests are failing after these changes. Seems to be incomplete. Overall it may make sense to decouple it a bit more as you propose so logic is a bit simpler in each class.

zalatov:
Ok, I understood that I'll correct tests and commit changes.

zalatov:
Fix tests

zalatov:
I need to add row into CHANGELOG? Create new section with ```2.1.0 under development```?

cebe:
not yet, we are going to review the changes for 2.1 and if we agree on a solution, you can add a changelog. otherwise it will likely cause unnecessary merge conflicts.

zalatov:
Update:
Backward compatibility is available (old implementations marked as deprecated).

samdark:
Build still fails: https://travis-ci.org/yiisoft/yii2/builds/179311073

samdark:
Won't make it into 2.0.

PR #12553, opened 2016-09-16
display the table when dropping index

rhertogh:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | - |

display the table, helping when dropping multiple indices with the same name


SilverFire:
Thank you

PR #12554, opened 2016-09-16
Update UniqueValidator.php

DrDeath72:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |

fix if indexBy('id') for example


samdark:
What does the change do?

DrDeath72:
get first array value if indexed by field

samdark (changes requested):
OK. That makes sense. Can you add a line for changelog and a test for it?

DrDeath72:
nope, i noob :D

SilverFire (approved):

samdark (approved):

samdark:
Merged. Thanks!

PR #12563, opened 2016-09-18
Added first() method to ActiveRecord and Query classes.

zalatov:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

Find one selects all data from DB and then reduce it to one record.
Find first set limit to returned data.

For example, you can execute this:
`
$lastLog = BigLogTable::find()->orderBy(['insert_stamp' => SORT_DESC])->one();
`
And you do not suspect that all entries will be selected before result truncated to one record.

This speciality discussed on yii forums, but not included in repository yet.

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -104,6 +104,15 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
 
     /**
      * @inheritdoc
+     * @return static|null ActiveRecord instance matching the condition, or `null` if nothing matches.
+     */
+    public static function findFirst($condition)
+    {
+        return static::findByCondition($condition)->first();
+    }
+
+    /**
+     * @inheritdoc
      * @return static[] an array of ActiveRecord instances, or an empty array if nothing matches.
      */
     public static function findAll($condition)
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -248,6 +248,19 @@ class Query extends Component implements QueryInterface
     }
 
     /**
+     * Executes the query and returns a first single row of result.
+     * @param Connection $db the database connection used to generate the SQL statement.
+     * If this parameter is not given, the `db` application component will be used.
+     * @return array|boolean the first row (in terms of an array) of the query result. False is returned if the query
+     * results in nothing.
+     */
+    public function first($db = null)
+    {
+        $query = clone $this;
+        return $query->limit(1)->one($db);
+    }
+
+    /**
      * Returns the query result as a scalar value.
      * The value returned will be the first column in the first row of the query results.
      * @param Connection $db the database connection used to generate the SQL statement.
--- a/framework/db/ActiveRecordInterface.php
+++ b/framework/db/ActiveRecordInterface.php
@@ -185,6 +185,40 @@ interface ActiveRecordInterface
     public static function findOne($condition);
 
     /**
+     * Returns a first single active record model instance by a primary key or an array of column values.
+     *
+     * The method accepts:
+     *
+     *  - a scalar value (integer or string): query by a single primary key value and return the
+     *    corresponding record (or null if not found).
+     *  - a non-associative array: query by a list of primary key values and return the
+     *    first record (or null if not found).
+     *  - an associative array of name-value pairs: query by a set of attribute values and return a single record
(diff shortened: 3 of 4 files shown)


framework/db/ActiveQuery.php
```diff
@@ -297,6 +297,19 @@ public function one($db = null)
...
+     * Executes query and returns a first single row of result.
+     * @param Connection $db the DB connection used to create the DB command.
+     * If null, the DB connection returned by [[modelClass]] will be used.
+     * @return ActiveRecord|array|null a single row of query result. Depending on the setting of [[asArray]],
+     * the query result may be either an array or an ActiveRecord object. Null will be returned
+     * if the query results in nothing.
+     */
+    public function first($db = null)
+    {
+        return $this->limit(1)->one($db);
```
mdmunir: How about
```php
$query = clone $this;
return $query->limit(1)->one($db);
```

framework/db/Query.php
```diff
@@ -248,6 +248,18 @@ public function one($db = null)
...
     /**
+     * Executes the query and returns a first single row of result.
+     * @param Connection $db the database connection used to generate the SQL statement.
+     * If this parameter is not given, the `db` application component will be used.
+     * @return array|boolean the first row (in terms of an array) of the query result. False is returned if the query
+     * results in nothing.
+     */
+    public function first($db = null)
+    {
+        return $this->limit(1)->one($db);
```
mdmunir:
```php
$query = clone $this;
return $query->limit(1)->one($db);
```

zalatov: Good idea, I will change to it

framework/db/BaseActiveRecord.php
```diff
@@ -104,6 +104,15 @@ public static function findOne($condition)
 
     /**
      * @inheritdoc
+     * @return static|null ActiveRecord instance matching the condition, or `null` if nothing matches.
+     */
+    public static function findFirst($condition)
```
arkhamvm: Better to add ```$sort``` parameter, cause no guarantees that ```limit 1``` will always return the same record.

klimov-paul:
Feature rejected: such code overcomplication is not justified.

framework/db/Query.php
```diff
@@ -248,6 +248,19 @@ public function one($db = null)
     }
 
     /**
+     * Executes the query and returns a first single row of result.
+     * @param Connection $db the database connection used to generate the SQL statement.
+     * If this parameter is not given, the `db` application component will be used.
+     * @return array|boolean the first row (in terms of an array) of the query result. False is returned if the query
```
SilverFire: `static` will be returned, not array, won't it?

framework/db/Query.php
```diff
@@ -248,6 +248,19 @@ public function one($db = null)
     }
 
     /**
+     * Executes the query and returns a first single row of result.
```
SilverFire: `a` -> `the`

SilverFire: Maybe change `of result` -> `of the result`?

zalatov:
@klimov-paul, in 99% of cases it's necessary to use a limit.
Very-very small cases needs to use one() without limit().

Many peoples don't know that need to use limit() before one() because it's implicitly.

May be change default behavior and append limit everytime?
But when you need no limit for optimization, you can add limit(false) before one()?

framework/db/BaseActiveRecord.php
```diff
@@ -104,6 +104,15 @@ public static function findOne($condition)
 
     /**
      * @inheritdoc
+     * @return static|null ActiveRecord instance matching the condition, or `null` if nothing matches.
```
zalatov: This is copy of phpdoc from one() method.

framework/db/Query.php
zalatov: This is copy of phpdoc from one() method.

zalatov: This is copy of phpdoc from one() method.

framework/db/Query.php
zalatov: Query returns array, not object.

And...
This is copy of phpdoc from one() method.

cebe:
Also `->first()` is less obvious than `->limit(1)->one()` to the reader of the code.

zalatov:
I think that it's conversely.

Set limit(1) before call one() looks strange.
Adding first() method desides this.

PR #12568, opened 2016-09-19
Wrap "null"

Mak-Di:


SilverFire:
I think it's better to fix all occurrences at once instead of fixing word-by-word. @Mak-Di what do you think?

Mak-Di:
@SilverFire , agree with you.

PS: When I have free time I read the docs. If I find some mistake, I fix it. Because it looks like  fixing word-by-word.

SilverFire:
@Mak-Di see #12570 please

PR #12578, opened 2016-09-20
Runtime cache of found view files.

zalatov:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

Runtime cache of found view files.
This reduce count of calls to disk and manipulations with strings when calculation the result path.
Caching increase speed when findViewFile() calls many times.

For example, we calls $this->renderPartial() for one view file more than 500 times.

### DIFF
--- a/framework/base/View.php
+++ b/framework/base/View.php
@@ -102,6 +102,12 @@ class View extends Component
      */
     private $_viewFiles = [];
 
+    /**
+     * @var string[] Runtime cache of found view files.
+     * This reduce count of calls to disk and manipulations with strings when calculation the result path.
+     * Caching increase speed when findViewFile() calls many times.
+     */
+    private $_foundViewFiles = [];
 
     /**
      * Initializes the view component.
@@ -162,6 +168,11 @@ class View extends Component
      */
     protected function findViewFile($view, $context = null)
     {
+        $cacheKey = get_class($context) . '@' . $view;
+        if (isset($this->_foundViewFiles[$cacheKey])) {
+            return $this->_foundViewFiles[$cacheKey];
+        }
+
         if (strncmp($view, '@', 1) === 0) {
             // e.g. "@app/views/main"
             $file = Yii::getAlias($view);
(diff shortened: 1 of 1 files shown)


samdark:
It looks interesting. Have you benchmarked it?

zalatov:
It's a clear yii2-basic application.

```
1 000 iterations
cache:    float 0.0070009231567383
no-cache: float 0.012001037597656

10 000 iterations
cache:    float 0.070007085800171
no-cache: float 0.12501192092896
```

Caching give more performance when using and extends theme component.
For example we have view files of default theme and extends some of files in advance theme.
In this case we have more operations with disk and concatenitions.

Each file checking as this (as example):
```
$files = [
    '/themes/advance/view.php',
    '/themes/default/view.php',
];

foreach ($files as $file) {
    if (file_exists($file)) return $file;
}

return false;
```

framework/base/View.php
```diff
@@ -162,6 +168,11 @@ public function render($view, $params = [], $context = null)
      */
     protected function findViewFile($view, $context = null)
     {
+        $cacheKey = get_class($context) . '@' . $view;
```
SilverFire: 1. `$context` can be `null`
2. `Yii::$app->getViewPath()` on lines [:181](https://github.com/yiisoft/yii2/pull/12578/files#diff-64dcde6a6056dd54d40f3d5a0a79b2efR181), [:190](https://github.com/yiisoft/yii2/pull/12578/files#diff-64dcde6a6056dd54d40f3d5a0a79b2efR190) could be changed during the runtime
3. `Yii::$app->controller->module->getViewPath()` ([:185](https://github.com/yiisoft/yii2/pull/12578/files#diff-64dcde6a6056dd54d40f3d5a0a79b2efR185) will be changed during the runtime for sure
4. Even `$context->getViewPath()` on line [:180](https://github.com/yiisoft/yii2/pull/12578/files#diff-64dcde6a6056dd54d40f3d5a0a79b2efR190) colud be changed as well

IMO, this caching will save some micro-time, but will cause tons of troubles.

SilverFire (changes requested):
Against caching at all

SilverFire:
Sorry, but we can not accept this PR for the reasons I've posted in my review

PR #12579, opened 2016-09-21
Added example from prompt method.

nadar:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | none |

### DIFF
--- a/framework/console/Controller.php
+++ b/framework/console/Controller.php
@@ -260,7 +260,7 @@ class Controller extends \yii\base\Controller
      * ```php
      * $code = $this->prompt('Enter 4-Chars-Pin', ['required' => true, 'validator' => function($input, &$error) {
      *     if (strlen($input) !== 4) {
-     *         $error = 'The Pin must be at least 4 chars!';
+     *         $error = 'The Pin must be exactly 4 chars!';
      *         return false;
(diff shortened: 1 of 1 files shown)


framework/console/Controller.php
```diff
@@ -254,6 +254,19 @@ public function stderr($string)
      *  - validator: a callable function to validate input. The function must accept two parameters:
      *      - $input: the user input to validate
      *      - $error: the error value passed by reference if validation failed.
+     * 
+     * An example of how to use the prompt method with a validator function.
+     * 
+     * ```php
+     * $code = $this->prompt('Enter 4-Chars-Pin', ['required' => true, 'validator' => function($input, &$error) {
+     *     if (strlen($input) !== 4) {
+     *         $error = 'The Pin must be at least 4 chars!';
```
cebe: `at least 4 chars` means 4 chars or more, but the check is whether it is equal to 4. should be changed.

cebe (changes requested):
good addition except the one comment!

samdark (approved):

cebe:
Thanks you!

Issue #12600, opened 2016-09-26
Validator whenClient should also accept a PHP callback


mikehaertl:
With `Validator::when` you receive the currently validated `$model`, so you can make the validation depend on model values.

But with `whenClient` you can only specify a fixed string with a js validation function. So currently it's not possible to create a dynamic client validation function that depends on a model property. I suggest to make 'whenClient' either a string or a PHP callback which returns a string with the js validation function.

Here's an example how this could be used to create dynamic js:

```
'whenClient' => function ($model, $attribute) {
    return 'function (attribute, value) {'.
        ($model->status==='open' ? 'return true;' : 'return value=="foo"').
        '}';
},
```

SilverFire:
It is possible to generate client validation function with code like this:

```php
'whenClient' => $this->clientStatusValidation()
...
private function clientStatusValidation()
{
    return 'function (attribute, value) { ... }'
}
```

Closure will make the thing easier, but it's a BC breaking change and I doubt about it's necessity.

cebe:
The model passed to the closure does not have any difference compared to just using `$this` when defining rules in this case. If the attribute can change, that must be reflected in JS code, if it can not change in the current form you can access the current value using `$this`. I do not see why allowing a closure here is necessary.

mikehaertl:
>The model passed to the closure does not have any difference compared to just using $this

Hmmm, you're probably right. I first thought, it's the same situation as in `when`. There we can't use `$this` as the submitted attributes are not assigned yet. But that makes no sense when talking about client validation.

I just wasn't sure, if `rules()` isn't called on a "naked" instance (`new MyModel`). If you can confirm that this is not the case, and we always get the current model instance in `$this` then I agree that this change here is not neccessary.

mikehaertl:
I've tested with `$this` in my use case and it seems to work as expected. So feel free to close this unless you still see a case where `$this` would not work.

PR #12601, opened 2016-09-26
Issue #12600 Make whenClient also accpet a PHP callback

mikehaertl:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12600 |

### DIFF
--- a/framework/validators/Validator.php
+++ b/framework/validators/Validator.php
@@ -156,10 +156,11 @@ class Validator extends Component
      */
     public $when;
     /**
-     * @var string a JavaScript function name whose return value determines whether this validator should be applied
+     * @var string|callable a JavaScript function name whose return value determines whether this validator should be applied
      * on the client-side. The signature of the function should be `function (attribute, value)`, where
      * `attribute` is an object describing the attribute being validated (see [[clientValidateAttribute()]])
-     * and `value` the current value of the attribute.
+     * and `value` the current value of the attribute. This can also be a PHP callback which must return such JavaScript
+     * function as string. The signature of the callable should be `function ($model, $attribute)`.
      *
      * This property is mainly provided to support conditional validation on the client-side.
      * If this property is not set, this validator will be always applied on the client-side.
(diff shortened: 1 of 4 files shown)


framework/widgets/ActiveField.php
```diff
@@ -738,7 +738,9 @@ protected function getClientOptions()
                 $js = $validator->clientValidateAttribute($this->model, $attribute, $this->form->getView());
                 if ($validator->enableClientValidation && $js != '') {
                     if ($validator->whenClient !== null) {
-                        $js = "if (({$validator->whenClient})(attribute, value)) { $js }";
+                        $exp = is_callable($validator->whenClient) ?
```
SilverFire: We do not use variables name shortening.

SilverFire: The method was already over-complicated with nested `if` conditions, adding one more does not make the method better :(

SilverFire (changes requested):

framework/validators/Validator.php
```diff
@@ -156,10 +156,11 @@ class Validator extends Component
      */
     public $when;
     /**
-     * @var string a JavaScript function name whose return value determines whether this validator should be applied
+     * @var string|callable a JavaScript function name whose return value determines whether this validator should be applied
      * on the client-side. The signature of the function should be `function (attribute, value)`, where
      * `attribute` is an object describing the attribute being validated (see [[clientValidateAttribute()]])
-     * and `value` the current value of the attribute.
+     * and `value` the current value of the attribute. This can also be a PHP callback which must return such JavaScript
```
SilverFire: Closure signature must be described

mikehaertl: Done.

framework/widgets/ActiveField.php
mikehaertl: Hmm, the complexity is because we mix PHP with Js here. I can't see how this could be simplified. Actually I also don't think that it's extremely hard to understand, once you got that basic distinction right.

mikehaertl: Updated.

PR #12608, opened 2016-09-26
makes ModuleTest pass in windows environment

dynasource:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

### DIFF
--- a/tests/framework/base/ModuleTest.php
+++ b/tests/framework/base/ModuleTest.php
@@ -19,7 +19,7 @@ class ModuleTest extends TestCase
     {
         $module = new TestModule('test');
         $this->assertEquals('yiiunit\framework\base\controllers', $module->controllerNamespace);
-        $this->assertEquals(__DIR__ . '/controllers', $module->controllerPath);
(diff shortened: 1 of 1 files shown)


SilverFire (changes requested):
Looks like a hack rather than fix :)

Could you try to wrap `$this->controllerPath` with a `realpath()` function?

dynasource:
well, these paths are quite messed up in the first place. Therefore realpath doesnt work. I choose this solution instead of this one:
```php
        $this->assertEquals(__DIR__ . DIRECTORY_SEPARATOR . 'controllers', str_replace(['/','\\'], DIRECTORY_SEPARATOR , $module->controllerPath));
```

which one do you prefer?

SilverFire:
The second one. It it clear to understand that we do something only with slashes in the path, but do not affect the path itself.

Is this the only place where tests fail on Win env?

dynasource:
it seems so on the filepath subject. I will push the second solution

SilverFire (approved):

PR #12610, opened 2016-09-26
Allow omit optional params on URL creation

### LINKED ISSUE #10970: URL Rule skipped if default parameter in omitted
I'm probably missing something obvious, but default parameter in a rule just doesn't seem to behave as stated in the [routing guide](http://www.yiiframework.com/doc-2.0/guide-runtime-routing.html). It claims that if I specify a default for some parameter, I would be able to create URLs using this rule and omitting the parameter. Well, that doesn't work.

Rule configuration:

``` php
[
     'pattern' => 'post/<id:\d+>/<slug>',
     'route' => 'post/view',
     'defaults' => ['slug' => ''],
]
```

I want to be able to create URLs without the slug. So I expect the following call

``` php
Url::to(['post/view', 'id' => 100]);
```

to generate URL `/post/view/100`. Instead it skips the rule and creates the regular URL `/post/view?id=100`.

Looking at the framework code, there's a part that  puzzles me in `UrlRule` class:

``` php
// match default params
// if a default param is not in the route pattern, its value must also be matched
foreach ($this->defaults as $name => $value) {
// ...
    if (!isset($params[$name])) {
        return false;
    } // ...
}
```

So it clearly reads that if a parameter is mentioned in defaults section (such as `slug` in my case), and the value for this parameter is not provided when creating URL, then skip the rule. That contradicts what I read in the documentation.

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #10970 |

closes #10970


samdark (approved):

SilverFire (approved):

cronfy (changes requested):
I do not think this PR has to be merged.
1. It does not fix #10970 
2. It probably breaks BC.

Please see my comments in #10970

rob006:
I've made some changes. Right now this behavior affects only empty default params. It still fix example from #10970 but should keep BC.

rob006:
This is ready to merge IMO.

framework/web/UrlRule.php
```diff
@@ -367,8 +367,15 @@ public function createUrl($manager, $route, $params)
                 continue;
             }
             if (!isset($params[$name])) {
-                return false;
-            } elseif (strcmp($params[$name], $value) === 0) { // strcmp will do string conversion automatically
+                // allow omit empty optional params
+                // @see https://github.com/yiisoft/yii2/issues/10970
+                if (in_array($name, $this->placeholders) && strcmp($value, '') === 0) {
```
samdark: Isn't it better to do `$value === ''` instead of `strcmp($value, '') === 0`?

samdark: Or empty() is you've meant it to work w/ conversion.

rob006: `$value` could be `0`. `empty()` will treat `0` the same as `''` (empty string) which is incorrect, and `===` will return fail when comparing `0` (integer) and `"0"` (string), which is impractical - finally all params are converted to string, so type is irrelevant. String comparison is the best option here.

samdark (approved):

framework/web/UrlRule.php
liongen: if my default params is empty it works well, but when I tried with non empty default value, it doesn't work

samdark: @liongen could you elaborate? How it doesn't work? Errors? What's the error?

rob006: @liongen It is by design, see https://github.com/yiisoft/yii2/issues/10970#issuecomment-250121267

liongen: @rob006 I get some puzzled, I'll try to explain what I want to do:

I have rules, let's say like this
`
[
           'pattern' => '<lang>/page/<id:\d+>',
           'route' => 'site/view',
           'host' => 'http://yii2.app',
           'defaults' => ['lang' => '']
]
`

First, when I tried to make url with `Url::to(['site/view', 'id' => 123]);`
the output was /site/view?id=123, when I applied your patch (from here), the url was like `http://yii2.app/page/123` and that's ok, but when I set some data to lang (let's say `'lang' => 'en'`) in defaults, my url returned to /site/view?id=123 type, but I expected something like http://yii2.dev/en/page/123

The fact is that value in defaults is global and can be set based on some conditions (`'defaults' => ['lang' => $someValue]`), and I don't want to set this value in all my urls when generate them

rob006: @liongen If you want add default language in this way, you should override default `UrlManager::createUrl()` and inject default language into params before URL creation. Magic injecting the default parameters when creating URLs will create unexpected restults and probably will break BC.

SilverFire (approved):

SilverFire:
Ready to merge IMO

samdark:
@cebe ?

rob006:
@cebe Are there any problems with this PR?

cebe:
I had no time to check it yet, sorry. If someone else wants to do it. Feel free to work on it.

tests/framework/web/UrlRuleTest.php
```diff
@@ -51,8 +51,8 @@ public function testParseRequest()
                 $request->pathInfo = $test[0];
                 $expected = $test[1];
                 $result = $rule->parseRequest($manager, $request);
-                if ($expected === false) {
-                    $this->assertFalse($result, "Test#$i-$j: $name");
+                if (empty($expected)) {
```
dynasource: why are these checks done with empty? I dont see any occurrence in the test execution in which an empty array is checked here? Only `false` or a filled `array`

rob006: This check is for empty strings. Empty string and `false` means completely different thing in this case. See https://github.com/yiisoft/yii2/pull/12610/files#diff-96fb0ef29c09f20871090c41bf5b3839R527

dynasource: I dont see any empty strings going through that $expected variable

rob006: b47fffb8

rob006: Ah, you're right, this should be done only for `createUrl()` tests. I will revert this.

SilverFire (approved):
Good to merge

SilverFire:
Squashed and merged: 97cad130f38766e1eb0a1945d51701091b0ab157

Thank you very much!

PR #12612, opened 2016-09-26
Better Query::andWhere() behavior

brandonkelly:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

This commit slightly improves the behavior of Query::andWhere().

If the query already has conditions on it created with the `['AND', condition1, ...]` array syntax, andWhere() will just append $condition to that array now, rather than creating another nested AND condition.

There’s no tangible difference here, but it makes debugging queries a little easier if andWhere() had been called on it multiple times. Now you're just looking at a single-dimensional array of all the conditions added by andWhere(), rather than several nested conditions.


SilverFire:
This change is completely wrong and will break BC a lot for no reason. It can not be accepted, sorry.

brandonkelly:
@SilverFire can you post an example of something it would break?  I can’t think of any.

SilverFire:
Sure, just check out the tests you've broken https://travis-ci.org/yiisoft/yii2/builds/162868184

brandonkelly:
@SilverFire Those two assertions just needed to be updated. Done here: https://github.com/brandonkelly/yii2/commit/83298c3c0106bef29366c64a2bf1699877551d9f

With all due respect, I really can’t think of any real-world scenarios where this change would break anything. It just makes debugging queries a bit easier.

Take the conditions created by that QueryTest::testAndFilterCompare() test, for example. Here’s what `$query->where` looks like without this change:

screen shot 2016-09-26 at 2 00 01 pm

and here’s what it looks like after the change:

screen shot 2016-09-26 at 1 59 20 pm

And here's how the generated SQL differs:

```sql
# Old
SELECT * WHERE (((`name` = :qp0) AND (`name` LIKE :qp1)) AND (`rating` > :qp2)) AND (`value` <= :qp3)

# New
SELECT * WHERE (`name` = :qp0) AND (`name` LIKE :qp1) AND (`rating` > :qp2) AND (`value` <= :qp3)
```

Both will have the same effect, but I hope you'll agree this is much nicer and clearer :)

dynasource:
although these look nicer, you loose the information about how many andWhere statements were used, which I actually find quite handy.

brandonkelly:
@dynasource If you really need to know that, another way to do it (if you use PhpStorm) would be to set a breakpoint in andWhere() with the “Log message to console” option selected.

screen shot 2016-09-27 at 4 46 52 am

dynasource:
dont you loose valuable information in the following usecase:
```php
$query->andWhere('attrA = "A" || attrB ="B"')
$query->andWhere('attrC = "C" || attrD ="D"')
```

tests/framework/db/QueryTest.php
```diff
@@ -275,11 +275,11 @@ public function testAndFilterCompare()
         $query->andFilterCompare('name', 'Doe', 'like');
         $this->assertEquals($condition, $query->where);
 
-        $condition = ['and', $condition, ['>', 'rating', '9']];
+        $condition[] = ['>', 'rating', '9'];
```
brandonkelly: @githubjeka What problems?

brandonkelly:
@dynasource That code would continue working as expected with this change.

CedricYii:
The parenthesis are useful when mixing 'and' and 'or'.
How would it behave with 
`andWhere(['or', ['and', 'company.type=1', 'order.type=2'],  ['and', 'company.type=3', 'order.type=4']])` or 
`andWhere(['and', ['or', 'company.type=1', 'order.type=2'],  ['or', 'company.type=3', 'order.type=4']])` ?

brandonkelly:
@CedricYii Those would continue working as they did before.

framework/db/Query.php
```diff
@@ -553,6 +553,8 @@ public function andWhere($condition, $params = [])
     {
         if ($this->where === null) {
             $this->where = $condition;
+        } elseif (is_array($this->where) && isset($this->where[0]) && strtoupper($this->where[0]) == 'AND') {
```
samdark: `strncasecmp($this->where[0], 'AND', 3)` could be used instead of `strtoupper` and comparison.

brandonkelly: @samdark Good point, though maybe `strcasecmp()` would be better? `strncasecmp()` would also return `0` if `where[0]` was `'andfoobar'`

samdark: Yes. `strcasecmp()` would be better.

brandonkelly:
@samdark just made the change you suggested: https://github.com/yiisoft/yii2/pull/12612/commits/5603a16f918bc59fc35adfc920b2f6941c5d9fb2

cebe (approved):
looks good to me. needs a CHANGELOG entry.

brandonkelly:
@cebe Done: https://github.com/yiisoft/yii2/pull/12612/commits/c6ea2ea077c3e21b69532d2c61b669c428e2586c

SilverFire (approved):

samdark (approved):

samdark:
Merged. Thank you!

PR #12614, opened 2016-09-26
Fix consistent callable conditions GridView

dynasource:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |

### DIFF
--- a/framework/grid/GridView.php
+++ b/framework/grid/GridView.php
@@ -463,7 +463,7 @@ class GridView extends BaseListView
         $rows = [];
         foreach ($models as $index => $model) {
             $key = $keys[$index];
-            if ($this->beforeRow !== null) {
+            if (is_callable($this->beforeRow)) {
                 $row = call_user_func($this->beforeRow, $model, $key, $index, $this);
                 if (!empty($row)) {
                     $rows[] = $row;
@@ -472,7 +472,7 @@ class GridView extends BaseListView
 
             $rows[] = $this->renderTableRow($model, $key, $index);
 
-            if ($this->afterRow !== null) {
+            if (is_callable($this->afterRow)) {
                 $row = call_user_func($this->afterRow, $model, $key, $index, $this);
                 if (!empty($row)) {
                     $rows[] = $row;
@@ -503,7 +503,7 @@ class GridView extends BaseListView
         foreach ($this->columns as $column) {
             $cells[] = $column->renderDataCell($model, $key, $index);
         }
-        if ($this->rowOptions instanceof Closure) {
+        if (is_callable($this->rowOptions)) {
             $options = call_user_func($this->rowOptions, $model, $key, $index, $this);
         } else {
             $options = $this->rowOptions;


rob006:
This degrades performance without any real benefits.

dynasource:
please elaborate on why the following improvements arent 'real benefits':
* add more flexibility to $rowOptions not limiting it to Closure
* add consistency to conditional statements for callbacks

With respect to performance, I doubt this will be significant for a gridview in an application. To quote @qiangxue, 'prefer readibility over micro-optimization'.

rob006:
1. Everything you can do with `callable` you can do also with `Closure` - there is no additional functionality, only another way to do the same thing (which typically reduces the readability).
2. Passing incorrect callback to `beforeRow` or `afterRow` will be just ignored without any error - this makes debugging much harder.

dynasource:
1) please explain how an callback configuration array $rowOptions could get past 'instanceof Closure'?
2) I agree. No other types are allowed here. Still, for readibility I would prefer isset()

rob006:
> please explain how an callback configuration array $rowOptions could get past 'instanceof Closure'?

You can use:
```php
function ($model) {
    return Helper::generateBeforeRowConfig($model);
}
```

instead of `['Helper', 'generateBeforeRowConfig']` - same result.

dynasource:
still defining $rowOptions as a property from an extending class is not supported, as functions cannot be assigned to PHPs properties. Thus, less flexibility, less usercases.

framework/grid/GridView.php
```diff
@@ -463,7 +463,7 @@ public function renderTableBody()
         $rows = [];
         foreach ($models as $index => $model) {
             $key = $keys[$index];
-            if ($this->beforeRow !== null) {
+            if (is_callable($this->beforeRow)) {
```
klimov-paul: This make absolutely no sense, since `beforeRow` accept only `null` or `callable`.
Such change produces silent fall in case user make a mistake in configuration.

framework/grid/GridView.php
```diff
@@ -472,7 +472,7 @@ public function renderTableBody()
 
             $rows[] = $this->renderTableRow($model, $key, $index);
 
-            if ($this->afterRow !== null) {
+            if (is_callable($this->afterRow)) {
```
klimov-paul: Same here

framework/grid/GridView.php
```diff
@@ -503,7 +503,7 @@ public function renderTableRow($model, $key, $index)
         foreach ($this->columns as $column) {
             $cells[] = $column->renderDataCell($model, $key, $index);
         }
-        if ($this->rowOptions instanceof Closure) {
+        if (is_callable($this->rowOptions)) {
```
klimov-paul: Such change is risk prone.
See https://github.com/yiisoft/yii/issues/3492

klimov-paul:
PR can not be accepted, because it produces potentioal errors.

framework/grid/GridView.php
dynasource: why? Are rowOptions ever to be defined by users?

framework/grid/GridView.php
dynasource: agreed (depending on which hat you put on)

framework/grid/GridView.php
dynasource: @klimov-paul: is it also risk-prone when using is_callable($this->rowOptions, true) in which syntax_only is set to true (If set to TRUE the function only verifies that name might be a function or method, see http://php.net/manual/en/function.is-callable.php#refsect1-function.is-callable-examples)

cebe: usually not, but the change is not needed and its better to stick with the more restrictive approach. is_callable can have a lot of side effects and should be avoided if only annonymous function is expected.

dynasource: please look at https://github.com/yiisoft/yii2/issues/12711 for an integral approach

PR #12615, opened 2016-09-26
enhance tests for DataColumn

dynasource:
this is a first step towards extension of tests for the DataColumn class:
- current functions are made distinguisable
- unnecessary code is removed
- useful @see references are added
- new tests are added


samdark:
Thanks!

PR #12622, opened 2016-09-28
New branch

jaaf:
| Q | A |
| --- | --- |
| Is bugfix? | yes/no |
| New feature? | yes/no |
| Breaks BC? | yes/no |
| Tests pass? | yes/no |

### DIFF
--- a/docs/guide-fr/db-active-record.md
+++ b/docs/guide-fr/db-active-record.md
@@ -1,9 +1,9 @@
-Enregistrement actif
-====================
+Enregistrement actif (*Active Record*)
+===================================== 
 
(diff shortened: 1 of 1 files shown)


jaaf:
@CedricYii 
I changed my mind since yesterday night. I apologize but after trying, with an open mind, to come to you way of translating, I found things more confusing and I could not resolve myself to do that way. It is above my strength.
I admit that class names should no be translated at all, and it is what I did (never translated class name), but when referring to the concept of Active Record, I don't understand why I could not translate it. Being consistent with you point of wiew I could never call a *query* requête, or a *model* modèle and I think things turn out to be a little crazy. 
The only important thing in my opinion is to help the reader localize Active Record if he is used to using the English word when thinking. Thus I keep a reference to Active Record in the title of the section.

CedricYii:
As I said, it is difficult to change one's opinion on translation.

However note that you are not always complying to your own rule as in:
````
1. Créer un nouvel objet *query* (requête) en appelant la méthode [[yii\db\ActiveRecord::find()]] ;
2. Construire l'objet *query* en appelant des [méthodes de construction de requête](db-query-builder.md#building-queries);
````
Or 
````
La seule différence est que, au lieu d'utiliser l'opérateur `new` pour créer un objet *query* (requête), vous appelez la méthode [[yii\db\ActiveRecord::find()]] pour retourner un nouvel objet *query* qui est de la classe [[yii\db\ActiveQuery]].
````
In my opinion these two extracts are offering an excellent useful easy to understand and to follow translation. Unfortunately they are not consistent with the other sections.

Anyhow, I appreciate your work on this translation and I hope it will be useful for newcomers.

jaaf:
I will sleep a while on it and then see. Thank you for your comments

PR #12628, opened 2016-09-28
Fix PHP notice on changelog sort command

rob006:
``` shell
robert@robert-laptop:~/NetBeansProjects/GitHub/yii2-dev$ ./build/build release/sort-changelog framework
PHP Notice 'yii\base\ErrorException' with message 'Only variables should be passed by reference'

in /home/robert/NetBeansProjects/GitHub/yii2-dev/build/controllers/ReleaseController.php:321

```


cebe:
Thanks!

Issue #12629, opened 2016-09-28
A custom input widget won't be validated if you change the ID


coderlex:
Validation of custom widget will not happen if you change it's auto-generated id.
But all works fine when changing the id of 'standard' inputs.

This happens because all input-setting methods (like `ActiveField::textInput()` or `ActiveField::checkboxList()`) call the protected `ActiveField::adjustLabelFor()` function which sets the internal `$_inputId` variable eigther. But the `ActiveField::widget()` method doesn't do this (BTW it also causes incorrect label's `for` value).

I suppose the issue can be solved by adding call of `adjustLabelFor()` in `widget()` method.

Now I'm using the following weird-looking workaround:
```php
echo $form->field($tag, 'category_id')
	// A hack to set internal input id
	->textInput([
		'id' => $form->id . '-category_id'
	])
	->widget(\frontend\widgets\Dropdown::className(), [
		'items' => $categoryIdOptions,
		'prompt' => 'Select a category...',
		'blockOptions' => [
			'class' => 'remodal__dropdown'
		],
		'options' => [
			'id' => $form->id . '-category_id'
		]
	]);
```

cebe:
can you send a pull request with your fix?

PR #12632, opened 2016-09-28
Fixed custom widget validation (#12629)

lexkrstn:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12629 |

### DIFF
--- a/framework/widgets/ActiveField.php
+++ b/framework/widgets/ActiveField.php
@@ -697,6 +697,9 @@ class ActiveField extends Component
         $config['model'] = $this->model;
         $config['attribute'] = $this->attribute;
         $config['view'] = $this->form->getView();
+        if (in_array('yii\widgets\InputWidget', class_parents($class)) && isset($config['options'])) {
+            $this->adjustLabelFor($config['options']);
+        }
         $this->parts['{input}'] = $class::widget($config);
 
         return $this;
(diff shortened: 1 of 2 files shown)


framework/widgets/ActiveField.php
```diff
@@ -697,6 +697,9 @@ public function widget($class, $config = [])
         $config['model'] = $this->model;
         $config['attribute'] = $this->attribute;
         $config['view'] = $this->form->getView();
+        if (in_array('yii\widgets\InputWidget', class_parents($class)) && isset($config['options'])) {
```
samdark: Couldn't it be checked like `$class instanceof yii\widgets\InputWidget`?

samdark (changes requested):

framework/widgets/ActiveField.php
coderlex: Type of `$class` variable is string. The expression will ever return false.

samdark: Right.

samdark (approved):

samdark:
Thanks!

framework/widgets/ActiveField.php
rob006: This should be more efficient:
```php
if (isset($config['options']) && isset(class_parents($class)['yii\widgets\InputWidget'])) {
```

PR #12641, opened 2016-09-30
Fix for #4479 REST filter

klimov-paul:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #4479 |

Implemented REST filter.
Format inspired from the ones used by ElasticSearch and MongoDB.


klimov-paul:
Added support for request via body content at `IndexAction`.
For the test you can use command like following:

```
curl -XGET http://domain.com/api/item?_format=json -H content-type:application/json -d '{"filter":{"id": 1}}'
```

samdark:
Is there a reason to use `$` in keys?

klimov-paul:
> Is there a reason to use $ in keys?

This is a practice borrowed from ElasticSearch and MongoDB. In particular I can provide following reasons:
- keywords with `$` are unlikely to be missplaced for a search attribute or value: `and`, `eq`, `lte` and so on can be names of attribute or actual values.
- keys with `$` are easy to notice while reading the filter in JSON format
- JSON documents can be automatically sorted by some reading tools (firebug console, for example), using `$` ensures keywords will appear at the top.

klimov-paul:
> Is there a reason to use $ in keys?

Also note that all operators are configurable: you can change thier keywords to whatever you like.

samdark:
Yes, I've noticed that. I'm concerned about defaults.

niksamokhvalov:
A good format. It seems to abandon the `$` of the impossible, though the format and complicated. This symbol needs to clearly define the terms `and` and `or`.

Here's what I don't like: applications, simultaneously, using forms and REST API for the same model, will have to make several implementations of the filters (searchModel and dataFilter models). In fact, we want to implement the same behavior, but we still need to write it twice.

klimov-paul:
> will have to make several implementations of the filters (searchModel and dataFilter models). In fact, we want to implement the same behavior, but we still need to write it twice.

You can reuse a search model create for the regular 'frontend' listing in REST even without this fix via `IndexAction::prepareDataProvider`:

```php
class SearchModel extends \yii\base\Model
{
    // ...

    public function search($params, $formName = null)
    {
        $query = Item::find();

        // add conditions that should always apply here

            'query' => $query,
        ]);

        $this->load($params, $formName);

        if (!$this->validate()) {
            $query->where('0=1');
            return $formName === '' ? $this : $dataProvider;
        }
    }

class ItemController extends \yii\rest\ActiveController
{

    // ...

    public function actions()
    {
        return [
            'index' => [
                'class' => 'yii\rest\IndexAction',
                'modelClass' => $this->modelClass,
                'prepareDataProvider' => function () {
                    $searchModel = new SearchModel();
                    return $searchModel->search(Yii::$app->request->queryParams, '');
                },
            ],
        ];
    }
```

With such code: search at the regular 'frontend' will be performed via following URL:

```
domain.com/items?ItemSearch[name]=foo&ItemSearch[status]=1
```

while for the REST - URL will be following:

```
domain.com/api/items?name=foo&status=1
```

There is no need to change something to implement simpliest search in the REST. I have alreay told this here:
https://github.com/yiisoft/yii2/issues/4479#issuecomment-248318709

This PR introduces sophisticated search condition builder, which can be used for creating 'database-like' search interface.
It may be useful in case you are splitting your application into a mirco services, which interact with each other via REST.

tunecino:
@klimov-paul shouldn't that SearchModel replace the old one [here](https://github.com/yiisoft/yii2-gii/blob/master/generators/crud/default/search.php) ? also what do you think about adding an interface for it ? (see related discussion in https://github.com/yiisoft/yii2/pull/11470)

klimov-paul:
> shouldn't that SearchModel replace the old one here ?

Unlikely, because it more complex and thus will confuse newcomers.
You can create your own Gii CRUD template, if you want to use it by default.

> what do you think about adding an interface for it ?

I am unsure if search interface can be made generic enough. It still may need extra options.

ddanielroche:
Hello @klimov-paul

I've been testing its implementation and I like it, but I have some questions:

1. When you plan to merge this feature in the kernel?
2. Which proposes for filters of related models?

On the other hand I think the operators 'like' and 'not like' are missing, in addition the operator `'$eq' => '=='` generate error in PostgreSql

Regards

klimov-paul:
> When you plan to merge this feature in the kernel?

This is not for me to decide. Until the team comes to consensus on the original issue - I can not merge it.
There is another proposal: #12251, which is under consideration.

> Which proposes for filters of related models?

There could be an improvement involving attributes mapping: `search-name => real-name`, but I don't want to spend additional effots on this until we agree on the proposal in general.

> On the other hand I think the operators 'like' and 'not like' are missing, in addition the operator '$eq' => '==' generate error in PostgreSql

Should be fixed now.

framework/rest/ActiveDataFilter.php
```diff
@@ -0,0 +1,165 @@
...
+ * @since 2.0.10
+ */
+class ActiveDataFilter extends DataFilter
+{
+    /**
+     * @var array map of filter condition keywords to build methods.
+     * These methods are used by [[buildCondition]] to build the actual filter conditions.
+     */
+    public $conditionBuilders = [
+        '$and' => 'buildConjunctionCondition',
```
dynasource: are these dollar signs really necessary? They add confusion about whether these are processed as PHP variables. If these are just used for mapping, if would expect a easy to read format like:
* `__OR__`
* `__LTE__`

or even `CONSTANTS`

framework/rest/ActiveDataFilter.php
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Builds conjunction condition, which consist of multiple independent ones.
+     * This covers such operators like `$and` and `$or`.
+     * @param string $operator operator keyword.
+     * @param mixed $condition raw condition.
+     * @return array actual condition.
+     */
+    protected function buildConjunctionCondition($operator, $condition)
+    {
```
dynasource: wouldnt Junction instead of Conjunction be better readable then "`Con`junction`Con`dition"

dynasource (changes requested):

niksamokhvalov:
Can I do something to help, that would rather merge this PR? It is very necessary, I would like to see in the next release.

framework/rest/ActiveDataFilter.php
```diff
@@ -0,0 +1,165 @@
...
+        '$or' => 'buildConjunctionCondition',
+        '$not' => 'buildBlockCondition',
+        '$lt' => 'buildOperatorCondition',
+        '$gt' => 'buildOperatorCondition',
+        '$lte' => 'buildOperatorCondition',
+        '$gte' => 'buildOperatorCondition',
+        '$eq' => 'buildOperatorCondition',
+        '$neq' => 'buildOperatorCondition',
+        '$in' => 'buildOperatorCondition',
+        '$nin' => 'buildOperatorCondition',
```
Faryshta: how about including a '$like'?

framework/rest/ActiveDataFilter.php
Faryshta: is there an RFC for this or similar filtering syntax? yii2 rest follows json hypermedia, hateoas and hal natively very closely so i think we should start searching there for filtering implementation.

Faryshta:
http://json-schema.org/latest/json-schema-hypermedia.html#rfc.section.5.6.2 json hypermedia propose other type of query parameters which can be autodocumented with json.

Example

```php
"self": {
        "encType": "application/x-www-form-urlencoded",
        "method": "get",
        "href": "/Product/",
        "properties": {
            "name": {
                "description": "name of the product",
                "type": "string"
            },
            "categoryId": {
                "description": "Category identifier",
                "type": "integer"
            }
        }
    }
```

usage

`/Product?name=banana&categoryId=1`

Faryshta:
http://jsonapi.org/recommendations/#filtering also found this reference but i don't think it will provide any complex filtering as proposed on this PR.

Faryshta:
overall i couldn't find any rfc with a rest-centered strategy for filtering or any library which already provides some form of filtering like he one proposed on this PR.

I am very in favor of this PR since the functionality is needed and I think it needs a `FilterStrategy` interface to be able to later support other type of complex filters.

```php
interface FilterStrategy
{
    public function filter(
        \yii\base\Model $model,
       array $queryParams, // you can pass the request GET params
       string $formName = '' // form name used in load() method
    ): ?\yii\data\DataProviderInterface; // can return null
}
```

`rules()`, `load()`, `validate()`  methods would be handled by `$model`. if the `FilterStrategy::filter()` method returns null, then the `$model` had validation errors and as such they can be accessed by `$model->getErrors()`

that way it will be easy to change the filter strategy.

framework/rest/ActiveDataFilter.php
klimov-paul: No, 'conjuction' is easier to spell, also uppercase 'J' is easy to mistake for other letter on many fonts.

klimov-paul:
Refactored:

- Extracted `DataFilter::$filterControls` allowing setup of control keywords to be used in filter.
- Added `DataFilter::$attributeMap` supporting table joins
- Added I18N for error messages

samdark (approved):

dmirogin (approved):

framework/rest/DataFilter.php
```diff
@@ -0,0 +1,809 @@
...
+    }
+
+    /**
+     * @param Model|array|string|callable $model model instance or its DI compatible configuration.
+     * @throws InvalidConfigException on invalid configuration.
+     */
+    public function setSearchModel($model)
+    {
+        if (is_object($model)) {
+            if (!$model instanceof Model && !$model instanceof \Closure) {
```
SilverFire: Could be moved to the condition above.

framework/rest/DataFilter.php
```diff
@@ -0,0 +1,809 @@
...
+        $model = $this->getSearchModel();
+
+        $attributeTypes = [];
+        foreach ($model->activeAttributes() as $attribute) {
+            $attributeTypes[$attribute] = self::TYPE_STRING;
+        }
+
+        foreach ($model->getValidators() as $validator) {
+            $type = null;
+            if ($validator instanceof BooleanValidator) {
```
SilverFire: I'd move this part to a separate protected method to make it easier to adjust for custom validators.

framework/rest/DataFilter.php
```diff
@@ -0,0 +1,809 @@
...
+     *     'authorName' => '{{author}}.[[name]]'
+     * ]
+     * ```
+     *
+     * Attribute map will be applied to filter condition at [[normalize()]].
+     */
+    public $attributeMap = [];
+
+    /**
+     * @var array list of error messages responding to invalid filter structure, in format: `[errorKey => message]`.
```
SilverFire: `array|\Closure`

SilverFire (changes requested):

framework/rest/DataFilter.php
klimov-paul: No, it could not: `$model` may be an array or string class name, which is a valid value and should not throw an exception.

framework/rest/DataFilter.php
klimov-paul: Disagreed.
It is meant to override a whole `detectSearchAttributeTypes()` ot provide some sophosticated type detecttion mechanism. Particular implementation may not process validators at all.

framework/rest/DataFilter.php
klimov-paul: fixed

framework/rest/DataFilter.php
SilverFire: > $model may be an array

Then it will not be object and no exception will be thrown

framework/rest/DataFilter.php
SilverFire: Adding custom validator does not lead to any sophisticated type detection logic. It is still a map: class to type. I don't think that overriding the whole method (including logic) is a right way for filter configuration.

Moreover, current approach increases formal cyclomatic complexity.

klimov-paul: Repeat once more: method `detectSearchAttributeTypes()` as a whole determines the search attribute types. The actual algorithm of its detection may be **different**.
For example: you may want to get types from DB table schema, or may want to parse PHPDoc searching for type-hints.
It is only a single particular implementation, which relies on model validators.

If I add extra method for validators processing, it will become useless and never will be invoked in case `detectSearchAttributeTypes()` rewritten to use DB table schema.

> I don't think that overriding the whole method (including logic) is a right way for filter configuration.

How is method override related to filter configuration? You can set `$searchAttributeTypes` via `setSearchAttributeTypes()` to whatever you like. Method `detectSearchAttributeTypes()` invoked only in case there are `$searchAttributeTypes` explicitly configured.

framework/rest/DataFilter.php
klimov-paul: fixed

samdark:
Code and user API looks good to me. @klimov-paul should it be described in the guide?

klimov-paul:
> should it be described in the guide?

Extra docs will not hurt. But I can not afford to waste any more time on this.
If you suppose guide section to be essential - someone else should take care of it.

samdark:
OK.

samdark:
@yiisoft/reviewers need your help to:

1. Ensure overall code is good. I think it is but I could be wrong.
2. Ensure phpdoc is clear.
3. Ensure English in phpdoc is OK.

Thanks.

framework/rest/IndexAction.php
```diff
@@ -26,14 +26,47 @@ class IndexAction extends Action
...
+     *             ->addRule('name', 'trim')
+     *             ->addRule('name', 'string')
+     *             ->addRule('price', 'number');
+     *     },
+     * ]
+     * ```
+     *
+     * @see DataFilter
+     *
+     * @since 2.0.10
```
SilverFire: 2.0.13

SilverFire (approved):
Good otherwise

samdark:
Merged. Thanks for solving it.

samdark:
Docs will be handled separately but for 2.0.13 as well..

PR #12645, opened 2016-10-01
Keeping queries more inexpensive

Abbas-Hashemian:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | no |

In different parts of app repetitive cross joins are expensive. An intermediate inner join will enhance it :
(10 + 10)where... < (10 \* 10)where...
We have similar thing in the ActiveQuery. The child queries could be a sub query for a cleaner join but they are merged into the parent query which means multiplication of the number of rows instead of adding the number of rows.

And actually if we use joinWith , the eager loading won't work properly and it still makes two queries instead of one single join query (tracked up to ActiveQueryTrait::findWith -> ActiveRelationTrait::popuplateRelation ... $this->one() and $this->all() without any condition about joinWith). To check it we can make some passive breakpoints (echo without exit) in the Command::queryInternal to see how much load is on an active query command including information_schema.

### DIFF
--- a/framework/db/mysql/Schema.php
+++ b/framework/db/mysql/Schema.php
@@ -254,6 +254,7 @@ SELECT
     kcu.referenced_table_name,
     kcu.referenced_column_name
 FROM information_schema.referential_constraints AS rc
+INNER JOIN (SELECT 1) AS tmp ON rc.constraint_schema = database() AND rc.table_name = :tableName
 JOIN information_schema.key_column_usage AS kcu ON
     (
         kcu.constraint_catalog = rc.constraint_catalog OR
@@ -261,8 +262,7 @@ JOIN information_schema.key_column_usage AS kcu ON
     ) AND
     kcu.constraint_schema = rc.constraint_schema AND
     kcu.constraint_name = rc.constraint_name
-WHERE rc.constraint_schema = database() AND kcu.table_schema = database()
-AND rc.table_name = :tableName AND kcu.table_name = :tableName1
+WHERE kcu.table_schema = database() AND kcu.table_name = :tableName1
 SQL;
 
         try {


cebe:
The usual way to reduce load of schema queries would be schema caching.

samdark:
Any benchmarks performed?

Abbas-Hashemian:
@samdark Hi there
My late reply, sorry.
About information_schema I'm agree with Cebe however I'm not so familiar with Yii2 schema underneath cache (Yii1 had described it in the doc about caching table info).

But about benchmark of this style of query (tmp inner join) yes when I was working with Yii1 I had a table with about some thousands of records joining with other tables and a temporary fake inner join had a notable improvement. AFAK up to MySQL 5.5 query optimizer can not understand what clause of our where condition is to filter the left side of our joins based on our defined "constant" conditions (by params) however I think it is possible for optimizer to do it as a declarative programming language (I'm not sure about MySQL 5.7).

But this is only a suggestion about query style (not essentially to approve commit to git master). In ActiveQuery if we have a joinWith it will be placed in the first place of joins in the highest priority against all other joins. We can not change the place of the join or the conditions of the join no way. However [MySQL where optimization](http://dev.mysql.com/doc/refman/5.7/en/where-optimizations.html) seems to optimize the sort of columns of a composite index in a where/on clause (index(a,b,c) and where(b,a,c)) however it is not said clearly. But about the sort of joins I think it is a good idea to respect to the sort of joins when we have joins and joinWiths simultaneously. As far as I understood joinWiths are kept separately and will not be converted to $this->joins until the query prepare time then all joinWith will be converted. Thus the sort of joins can not be kept. I copied the ActiveQuery (because of private methods) and ActiveRelationTrait (because it was using the ActiveQuery class type) to make a different way of joinWith that converts the joinWith to join immediately (not at the prepare) and it worked well however by a createCommand I found that whether with the Original Yii2 ActiveQuery(and ActiveRelationTrait) or my customized ActiveQuery(and ActiveRelationTrait) the ActiveRelationTrait::popuplateRelation will make an additional separate simple query to build the relation.

samdark:
Briefly checked. Everything works with this pull request but I haven't found any performance difference on a real project with schema caching turned off and 20 tables in schema. Not sure if it's newer versions of MySQL optimizing better to if it's the fact that I don't have enough tables to notice the difference...

dynasource:
I've tested the SQLs without cache locally in Mysql with 1000 tables. Did not see any indication of performance gain so lets keep it as it is for now.

Issue #12399, opened 2016-08-30
Accessibility: add ARIA attributes (required and invalid) for inputs


tekord:
Suggestion. Add `aria-required` attribute rendering to required inputs. Example:

``` html
<div class="form-group field-supportissueform-subject required">
  <label class="control-label" for="supportissueform-subject">Subject</label>
  <input type="text" id="supportissueform-subject" class="form-control" 
         name="SupportIssueForm[subject]" aria-required> <!-- here! -->
</div>
```

Also if the field has validation errors then add `aria-invalid` attribute to field.

More about these attrubutes here:
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute

Oxyaction:
started work on this feature

tekord:
Thank you :)

PR #12647, opened 2016-10-01
Adds aria-attributes inputs support

Oxyaction:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12399 |

Aria-invalid supports both server, client and ajax validations.


framework/widgets/ActiveForm.php
```diff
@@ -164,6 +164,11 @@ class ActiveForm extends Widget
     public $attributes = [];
 
     /**
+     * @var array the list of model attributes that support aria-invalid and aria-required html attributes
```
SilverFire: `aria-invalid` -> <code>\`aria-invalid\`</code>
`aria-required` -> <code>\`aria-required\`</code>
`html` -> `HTML`

framework/widgets/ActiveField.php
```diff
@@ -800,4 +837,22 @@ protected function getInputId()
     {
         return $this->_inputId ?: Html::getInputId($this->model, $this->attribute);
     }
+
+    /**
+     * adds aria attributes to the input options
```
SilverFire: adds -> Addes

framework/widgets/ActiveField.php
```diff
@@ -800,4 +837,22 @@ protected function getInputId()
...
+     * @param $options array input options
+     */
+    protected function addAriaAttributes(&$options)
+    {
+        if ($this->addAriaAttributes) {
+
+            $options['aria-required'] =
+                (!isset($options['aria-required']) && $this->model->isAttributeRequired($this->attribute))
+                ? 'true' : 'false';
```
SilverFire: This line is redundant. PHP boolean will be converted to `true` or `false` automatically

framework/widgets/ActiveField.php
```diff
@@ -800,4 +837,22 @@ protected function getInputId()
...
+
+            $options['aria-invalid'] =
+                (!isset($options['aria-valid']) && $this->model->hasErrors($this->attribute))
+                ? 'true' : 'false';
```
SilverFire: Same here

framework/widgets/ActiveField.php
```diff
@@ -792,6 +811,24 @@ protected function getClientOptions()
     }
 
     /**
+     * Checks if client validation enabled for field
+     * @return bool
+     */
+    protected function getEnableClientValidation() {
```
SilverFire: `{` must be moved to the new line

framework/widgets/ActiveField.php
```diff
@@ -792,6 +811,24 @@ protected function getClientOptions()
     }
 
     /**
+     * Checks if client validation enabled for field
```
SilverFire: `for field` -> `for the field`

framework/widgets/ActiveField.php
```diff
@@ -800,4 +837,22 @@ protected function getInputId()
...
+    /**
+     * adds aria attributes to the input options
+     * @param $options array input options
+     */
+    protected function addAriaAttributes(&$options)
+    {
+        if ($this->addAriaAttributes) {
+
+            $options['aria-required'] =
+                (!isset($options['aria-required']) && $this->model->isAttributeRequired($this->attribute))
```
SilverFire: `aria-required` option must not be set when `$options['aria-required']` is already set

framework/widgets/ActiveField.php
```diff
@@ -800,4 +837,22 @@ protected function getInputId()
...
+                ? 'true' : 'false';
+
+            $options['aria-invalid'] =
+                (!isset($options['aria-valid']) && $this->model->hasErrors($this->attribute))
```
SilverFire: Same here

framework/widgets/ActiveForm.php
```diff
@@ -418,4 +424,49 @@ public static function validateMultiple($models, $attributes = null)
 
         return $result;
     }
+
+    /**
+     * Registers javascript function that toggles 'aria-invalid' attribute
+     * depending on input validation state
+     */
+    protected function registerClientAriaToggle()
```
SilverFire: The whole method should be removed. The functionality should be moved to `yii.activeForm.js`

SilverFire (changes requested):
`aria` is an abbreviation, should be upped-cased in all the PHPDocs

rob006 (changes requested):
Missing `@since 2.0.11` for all new methods and properties.

framework/widgets/ActiveField.php
Oxyaction: I'm not sure about this particular case, according to specs
[aria-required](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-required_attribute)
[aria-invalid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute)
need to be provided explicitly as `true` or `false`
when to use booleans instead of strings for example
```
$options = [
  'aria-required' => false,
];
```
can be confusing because `aria-required` won't be rendered at all which can be
treated as absence of `aria-required` attribute and not as the input value is actually not required.

framework/widgets/ActiveField.php
```diff
@@ -145,6 +145,10 @@ class ActiveField extends Component
      * it is maintained by various methods of this class.
      */
     public $parts = [];
+    /**
+     * @var bool adds ARIA HTML attributes `aria-required` and `aria-invalid` for inputs
+     */
+    public $addAriaAttributes = false;
```
cebe: how about enabling this by default. as far as I see it can not break anything and will make yii apps accessable by default?

samdark: Good idea.

framework/widgets/ActiveField.php
```diff
@@ -800,4 +845,21 @@ protected function getInputId()
     {
         return $this->_inputId ?: Html::getInputId($this->model, $this->attribute);
     }
+
+    /**
+     * Adds ARIA attributes to the input options
+     * @param $options array input options
```
SilverFire: `@since 2.0.11`

framework/widgets/ActiveField.php
```diff
@@ -788,10 +813,30 @@ protected function getClientOptions()
...
+     */
+    protected function getEnableClientValidation()
+    {
+        return $this->enableClientValidation
+            || $this->enableClientValidation === null && $this->form->enableClientValidation;
+    }
+
+    /**
+     * Checks if ajax validation enabled for the field
+     * @return bool
```
SilverFire: `@since 2.0.11`

framework/widgets/ActiveField.php
```diff
@@ -788,10 +813,30 @@ protected function getClientOptions()
             'validationDelay' => 500,
             'encodeError' => true,
             'error' => '.help-block',
+            'ariaInvalidToggle' => false,
         ]);
     }
 
     /**
+     * Checks if client validation enabled for the field
+     * @return bool
```
SilverFire: `@since 2.0.11`

framework/widgets/ActiveField.php
```diff
@@ -145,6 +145,10 @@ class ActiveField extends Component
      * it is maintained by various methods of this class.
      */
     public $parts = [];
+    /**
+     * @var bool adds ARIA HTML attributes `aria-required` and `aria-invalid` for inputs
```
SilverFire: `@since 2.0.11`

SilverFire (changes requested):

samdark:
Replaced by #13116

Issue #11058, opened 2016-03-09
controller->redirect() not support checkAjax


pigochu:
Hi

yii\web\Respone->redirect() has tree parameters

~~~
public function redirect($url, $statusCode = 302, $checkAjax = true)
~~~

why not add the same parameter in yii\web\Controller->redirect() ?

SilverFire:
It's a minor BC break, however the idea is good.

I think we can do it. @yiisoft/core-developers thoughts?

samdark:
What's the use case when checking AJAX should be turned off?

pigochu:
@samdark 
I test in Yii 2.0.7. and create simple CRUD code via gii.

If I use GridView in Pjax , when I click "delete" button  . it will load full page.

but when I turn off ajaxCheck , it will redirect to  actionIndex() and load pjax content.

~~~
    public function actionDelete($id)
    {
        $this->findModel($id)->delete();

        return Yii::$app->getResponse()->redirect(['category/index'], 302, false);
    }
~~~

samdark:
OK. Looks valid.

almirb:
Hello @pigochu !

I used your approach above and is working fine. The only thing that's a little annoying is the scroll of the page to the top after the delete action.  
Are you experiencing the same?

lynicidn:
```php
if (Yii::$app->request->isAjax) {
    return Yii::$app->response->redirect(['category/index'], 302, false);
}
```

what is behavior if request is ajax and need check on it? return 302 code?

almirb:
I solved It with:
````js
$.pjax.defaults.scrollTo = false;
````

Also tried to put this setting internally into my Kartik gridview with the code below, to avoid global changes i in my pjax configuration:
````php
'pjaxSettings' => ['options' => ['scrollTo' => false]],
````
But It didn't work.

**Comment by pigochu** [comment]
Now I upgrade yii to 2.0.8 , and create a test controller for testing.
I create CRUD from gii and "Enable Pjax" checkbox is checked .

I have not change any code . I got the response when I click "Delete".
![image](https://cloud.githubusercontent.com/assets/1013744/15796553/413aac52-2a32-11e6-949d-ae659aa85ad8.png)

We can see , the page will not load by pjax, other file still loaded.

But I change code in my action.

```
    public function actionDelete($id)
    {
        $this->findModel($id)->delete();
        if (Yii::$app->request->isAjax) {
[...]

PR #12650, opened 2016-10-02
Add checkAjax support in Controller::redirect

ivanovyordan:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | yes |
| Fixed issues | #11058 |

### DIFF
--- a/framework/web/Controller.php
+++ b/framework/web/Controller.php
@@ -140,10 +140,10 @@ class Controller extends \yii\base\Controller
      * See <http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html>
      * for details about HTTP status code
      *
-     * @param boolean $checkAjax whether to specially handle AJAX (and PJAX) requests. Defaults to true,
-     * meaning if the current request is an AJAX or PJAX request, then calling this method will cause the browser
-     * to redirect to the given URL. If this is false, a `Location` header will be sent, which when received as
-     * an AJAX/PJAX response, may NOT cause browser redirection.
+     * @param boolean $checkAjax whether to specially handle AJAX (and PJAX) requests. Defaults to false,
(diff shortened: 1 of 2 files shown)


framework/web/Controller.php
```diff
@@ -139,11 +139,17 @@ public function beforeAction($action)
      * @param integer $statusCode the HTTP status code. Defaults to 302.
      * See <http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html>
      * for details about HTTP status code
+     *
+     * @param boolean $checkAjax whether to specially handle AJAX (and PJAX) requests. Defaults to true,
+     * meaning if the current request is an AJAX or PJAX request, then calling this method will cause the browser
+     * to redirect to the given URL. If this is false, a `Location` header will be sent, which when received as
```
SilverFire: `false` -> <code>\`false\`</code>

SilverFire (changes requested):
CHANGELOG and UPGRADE.md should be updated

framework/web/Controller.php
ivanovyordan: Setting default value to `false` would break BC right?

SilverFire: Yeah

rob006:
This breaks BC anyway, since it changes method signature.

samdark:
Agree. It's a BC break.

framework/web/Controller.php
ivanovyordan: @SilverFire I made the changes.

Thanks for the feedback guys.

SilverFire:
I'm fine with merging into 2.1 right now

samdark:
Merged into 2.1: https://github.com/yiisoft/yii2/commit/0d72e37f3ce759a7286423eb32f19814bc2e0392

PR #12651, opened 2016-10-02
Fixed race condition in yii.js AJAX prefilter

### LINKED ISSUE #10358: Race condition in yii.js's script filter
The script filter in yii.js which prevents loading the same script twice has a race condition.
When a script file is first loaded, the ajax request is allowed to go through, and the script is added to loadedScripts. If the same script file is requested again while the first request is still loading, it will immediately be aborted. However, the script is not yet loaded, so subsequent scripts that might depend on this file will fail.

This does not cause a problem with $.html and friends, because they use synchronous ajax to load the scripts. However, this is deprecated, and will hopefully be fixed in jquery 3.0: https://github.com/jquery/jquery/issues/1895. Pjax also does it asynchronously: https://github.com/yiisoft/jquery-pjax/commit/7a47df7b36d1031fbd576929e040ebab588b207b

So e.g. loading two pjax partials which both have an activeform (and thus need yii.activeform.js)  at the same time will cause the second one to break, because the inline script will execute before yii.activeform.js is loaded (assuming the base page didn't load yii.activeform.js).

SilverFire:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | JS |
| Fixed issues | #10358 |

Closes #10358

The PR fixes race condition in AJAX prefilter.

Waiting for reviews from someone else.

@nkovacs

### DIFF
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -427,6 +427,12 @@ window.yii = (function ($) {
                             xhr.abort();
                         }
                     });
+                    // Or about previous XHR if the current one is loaded faster
+                    xhr.done(function () {
+                        if (item && item.readyState !== 4) {
+                            item.abort();
+                        }
+                    });
                 } else if (!isReloadable(url)) {
                     xhr.abort();
                 }
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -400,10 +400,14 @@ window.yii = (function ($) {
 
     function initScriptFilter() {
         var hostInfo = getHostInfo();
+        var loadedScripts = {};
 
-        var loadedScripts = $('script[src]').map(function () {
+        var scripts = $('script[src]').map(function () {
             return this.src.charAt(0) === '/' ? hostInfo + this.src : this.src;
         }).toArray();
+        for (var i = 0; i < scripts.length; i++) {
+            loadedScripts[scripts[i]] = true;
+        }
 
         $.ajaxPrefilter('script', function (options, originalOptions, xhr) {
             if (options.dataType == 'jsonp') {
@@ -411,12 +415,27 @@ window.yii = (function ($) {
             }
 
             var url = options.url.charAt(0) === '/' ? hostInfo + options.url : options.url;
-            if ($.inArray(url, loadedScripts) === -1) {
-                loadedScripts.push(url);
-            } else {
-                if (!isReloadable(url)) {
+
+            if (url in loadedScripts) {
+                var item = loadedScripts[url];
+
+                // If the concurrent XHR request is running and URL is not reloadable
+                if (item !== true && !isReloadable(url)) {
+                    // Abort the current XHR request when previous finished successfully
+                    item.done(function () {
+                        if (xhr && xhr.readyState !== 4) {
+                            xhr.abort();
+                        }
+                    });
+                } else if (!isReloadable(url)) {
                     xhr.abort();
                 }
+            } else {
+                loadedScripts[url] = xhr.done(function () {
+                    loadedScripts[url] = true;
+                }).fail(function () {
+                    delete loadedScripts[url];
+                });
             }
         });
 
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.10 under development
 ------------------------
 
+- Bug #10358: Fixed race condition in `yii.js` AJAX prefilter (silverfire)
 - Bug #7670: Added `UrlNormalizer` for normalizing requests with and without trailing slashes (rob006, cronfy, klimov-paul)
 - Bug #9027: Fixed descendant class of `yii\web\UploadedFile` returns parent instances in case invoked after it (andrewnester)
 - Bug #9101: Fixed `yii\web\View` to respect `yii\web\AssetManager::appendTimstamp` property (githubjeka, silverfire)


nkovacs:
I'm not sure this will work correctly, I think ajaxPrefilter is synchronous, so if you return from it, jQuery will start to load the script, so you may end up with the same js file loaded twice (another race condition), which this prefilter is supposed to prevent.
I think I tried to solve it like this when I reported the bug, but I'll have to check again.

SilverFire:
> so if you return from it, jQuery will start to load the script

[$.ajaxPrefilter docs](http://api.jquery.com/jquery.ajaxprefilter/)
No, it's async. You can return modify XHR or return a new `dataType`

> so you may end up with the same js file loaded twice

I'll explain the flow. For example, you have a page that has 2 `script` tags: for `a.js` and `b.js`.
You request to load that page twice and have 4 scripts request in queue:

1. a.js
2. b.js
3. a.js
4. b.js

№2 and №4 are set on hold because №1 and №2 are not loaded yet.

№1 and №3 start loading almost simultaneously, but when №1 finishes loading successfully, №3 will be aborted.

> another race condition

Hmm, is possible when №3 gets loaded faster than №1. Will fix it soon

nkovacs:
Yeah, I wanted to solve it by preventing No3's loading, but that's not possible (that's what I meant  by ajaxPrefilter being synchronous: once you return from it, the script will start loading, you can't delay it). I guess aborting the second a.js (regardless of which one finishes second) should be good enough.

SilverFire:
Could you check it now?

framework/assets/yii.js
```diff
@@ -400,23 +400,48 @@ window.yii = (function ($) {
...
+
+                // If the concurrent XHR request is running and URL is not reloadable
+                if (item !== true && !isReloadable(url)) {
+                    // Abort the current XHR request when previous finished successfully
+                    item.done(function () {
+                        if (xhr && xhr.readyState !== 4) {
+                            xhr.abort();
+                        }
+                    });
+                    // Or about previous XHR if the current one is loaded faster
```
nkovacs: about -> abort

nkovacs:
I can still reproduce the same error sometimes: Uncaught TypeError: jQuery(...).yiiActiveForm is not a function
But I'm not sure why, this should work.

I modified the basic template to add pjax to the entire page.
This is what my network tab looks like when the error happens:

contact.php (basic template's contact page)
yii.validation.js
yii.captcha.js
contact.php (second, slightly delayed click on contact link in menu)
yii.activeform.js

I use chrome's throttling to slow down everything, so I can more easily reproduce the bug.

nkovacs:
Wait, shouldn't it be readystate === 4 (DONE), i.e. abort the other one when this one is done?

nkovacs:
Ok, so the there's also a bug in pjax. It's caused by this line: https://github.com/yiisoft/jquery-pjax/blob/master/jquery.pjax.js#L815

The html already contains yii.activeform.js (even though it's actually still being loaded asynchronously), so pjax lets the inline script execute. The problem is that it inserts the script into the head before it is finished loading: https://github.com/yiisoft/jquery-pjax/blob/master/jquery.pjax.js#L821

The reason it does that is to prevent loading the script twice. So essentially pjax has the same bug with a completely different implementation.

SilverFire:
Moving https://github.com/yiisoft/jquery-pjax/blob/master/jquery.pjax.js#L821 into `done()` should solve the problem, right?

nkovacs:
That would create another bug in pjax, it would allow a script to be loaded multiple times as long as it has not finished loading.
Of course, yii.js would stop that, but then there's no point in even having that code in jquery.pjax.js, and you couldn't use your fork of pjax without Yii.

SilverFire:
> Of course, yii.js would stop that, but then there's no point in even having that code in jquery.pjax.js, and you couldn't use your fork of pjax without Yii.

I hope nobody uses fork without Yii :)

SilverFire:
Well, I can adopt code from Yii prefilter to Pjax, but I don't think we should do it

framework/assets/yii.js
```diff
@@ -400,23 +400,48 @@ window.yii = (function ($) {
 
     function initScriptFilter() {
         var hostInfo = getHostInfo();
+        var loadedScripts = {};
 
-        var loadedScripts = $('script[src]').map(function () {
+        var scripts = $('script[src]').map(function () {
             return this.src.charAt(0) === '/' ? hostInfo + this.src : this.src;
         }).toArray();
+        for (var i = 0; i < scripts.length; i++) {
```
samdark: `for (var i = 0, len = scripts.length; i < len; i++) {`

samdark (approved):

SilverFire:
@nkovacs I am fine with moving `document.head.appendChild(this)` to `done()`. Are you?

nkovacs:
If you're going to depend on yii.js's prefilter, then there's no point in even having jquery.pjax.js duplicating its functionality (preventing scripts being loaded twice).

But it's also not ideal that you're using a plugin that appears to be independent from Yii, but actually implicitly depends on it to work correctly.

It's also much easier to properly fix this in jquery.pjax.js, since there you can delay the dependent scripts: https://github.com/nkovacs/jquery-pjax/commit/f5dd009d39368266d2679ee4c2e1fb496cfb0a19

One other thing I realized while doing this: The inline script will execute twice in some cases, and since the html has already been replaced, activeForm will be initialized twice on the same form element, for example. This can only be properly fixed in pjax, e.g. abort any outstanding script executions when another page load is started.

SilverFire:
> One other thing I realized while doing this: The inline script will execute twice in some cases, and since the html has already been replaced, activeForm will be initialized twice on the same form element, for example. This can only be properly fixed in pjax, e.g. abort any outstanding script executions when another page load is started.

It comes more and more complicated. Let's skip this case for now

> It's also much easier to properly fix this in jquery.pjax.js, since there you can delay the dependent scripts: nkovacs/jquery-pjax@f5dd009

But it's not compatible with `reloadableScripts` in `yii.js`, is it?

nkovacs:
> But it's not compatible with reloadableScripts in yii.js, is it?

It wasn't compatible before either. If the script file is found in head, it won't ever be requested again by pjax, so it will skip the prefilter.

SilverFire:
> It wasn't compatible before either. If the script file is found in head, it won't ever be requested again by pjax, so it will skip the prefilter.

It was not, but I thought it could be fixed.

And I see only two ways how we can fix it:

1. Add logical coupling between `yii.js` and `pjax.js`.
2. Remove all extra logic and rely on `yii.js`. Yes, it makes fork useless without Yii, but we can mention it explicitly in fork description and forget about this problem

nkovacs:
You could rip out the duplicate script prevention in pjax and rely on ajaxPrefilter, but it's better to implement it in pjax (because you can delay dependent scripts if the requested script is already loading), and you can cancel script loading/execution (useful for inline scripts) when the user navigates away.

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/ff15100eab5217e0ccf9350e396f2a4aa93620b4

SilverFire:
Changes on pjax side will be applied later

Issue #12022, opened 2016-07-27
Ability to set showing values  asDuration() formatter


sokoji:
### What steps will reproduce the problem?
```<?= Yii::$app->formatter->asDuration($customTime) ?>```

### What is the expected result?
```25 days, 5 hours, 39 minutes, 43 seconds```

### What i want to get?
```25 days, 5 hours```

It will be good to make ability to set intervals wich will be showing in the result: years, months days, hours, minutes, seconds. For example:

```<?= Yii::$app->formatter->asDuration($customTime, ['seconds' => false, 'minutes' => false]) ?>```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      |  2.0.7+
| PHP version      |  any
| Operating system | any

sokoji:
@samdark can you give an advice how to wite params more better:

Now:
```public function asDuration($value, $implodeString = ', ', $negativeSign = '-')```

After:
```public function asDuration($value, $implodeString = ', ', $negativeSign = '-', $settings = [])```

or:

```public function asDuration($value, $settings = [], $implodeString = ', ', $negativeSign = '-')```

The first will be good becouse of backward compatibility. But in this case we need to pass $implodeString and $negativeSign.

I want to write this code and create pull request.
Thanx for your helping!

SilverFire:
Since we are to change the method signature, I prefer to drop `$implodeString` and `$negativeSign` parameters in favor of `$options`.

`public function asDuration($value, $options = [])`

cebe:
https://github.com/yiisoft/yii2/pull/12660#issuecomment-265848608

PR #12660, opened 2016-10-04
Added `\yii\i18n\Formatter::asDurationTime()`

githubjeka:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12022 fixed by the PR |

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -17,6 +17,7 @@ use Yii;
 use yii\base\Component;
 use yii\base\InvalidConfigException;
 use yii\base\InvalidParamException;
+use yii\helpers\ArrayHelper;
 use yii\helpers\FormatConverter;
 use yii\helpers\HtmlPurifier;
 use yii\helpers\Html;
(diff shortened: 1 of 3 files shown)


framework/i18n/Formatter.php
```diff
@@ -828,14 +827,66 @@ public function asRelativeTime($value, $referenceTime = null)
...
+     *    - boolean `showMinutes` will be displayed for minutes data. Default minutes data is displayed.
+     *      If `false` - seconds data isn't displayed also.
+     *    - boolean `showHours` will be displayed for hourly data. Default hourly data is displayed.
+     *      If `false` - minutes and seconds data isn't displayed also.
+     * @return string the formatted duration.
+     * @since 2.0.10
+     */
+    public function asDurationTime($value, $options = [])
+    {
+        if (!isset($options['implodeString'])) {
```
SilverFire: H

SilverFire: How about

```php
$implodeString = ArrayHelper::get($options, 'implodeString', ', ');
```

framework/i18n/Formatter.php
```diff
@@ -849,37 +900,53 @@ public function asDuration($value, $implodeString = ', ', $negativeSign = '-')
...
+            $interval = new DateInterval('P' . substr($value, 2));
             $isNegative = true;
         } else {
             $interval = new DateInterval($value);
             $isNegative = $interval->invert;
         }
 
         if ($interval->y > 0) {
-            $parts[] = Yii::t('yii', '{delta, plural, =1{1 year} other{# years}}', ['delta' => $interval->y], $this->locale);
+            $parts[] = Yii::t('yii', '{delta, plural, =1{1 year} other{# years}}', ['delta' => $interval->y],
```
SilverFire: Splitting for two lines does not look reasonable here

SilverFire (changes requested):
And we still have no way to configure years, months and days

framework/i18n/Formatter.php
```diff
@@ -828,14 +827,66 @@ public function asRelativeTime($value, $referenceTime = null)
      *    `P1D2H30M/2015-01-02T13:30:00Z` - time interval before datetime value
      *    `P1D2H30M` - simply a date interval
      *    `P-1D2H30M` - a negative date interval (`-1 day, 2 hours, 30 minutes`)
-     *
      * @param string $implodeString will be used to concatenate duration parts. Defaults to `, `.
      * @param string $negativeSign will be prefixed to the formatted duration, when it is negative. Defaults to `-`.
      * @return string the formatted duration.
      * @since 2.0.7
+     * @deprecated Since version in 2.0.10. The signature will be changed in 2.1 to [[Formatter::asDurationTime()]]
```
samdark: `in` isn't needed here. Just `Since version 2.0.10`

SilverFire:
Maybe, we should agree three different options for every part:

- `null` (default) - display part only when it's not zero
- `true` - always display a part, even it's zero
- `false` - never display a part

samdark (changes requested):

githubjeka:
is label `2.0.11` ?

SilverFire:
The PR does not look ready to be merged right now, so I've assigned it to the next milestone.
If you can continue work on this PR, we can accept it in 2.0.10 once it's ready

framework/i18n/Formatter.php
githubjeka: fixed thanks

framework/i18n/Formatter.php
githubjeka: Fixed

githubjeka:
Corrected

framework/i18n/Formatter.php
```diff
@@ -849,37 +906,44 @@ public function asDuration($value, $implodeString = ', ', $negativeSign = '-')
...
-            $interval = new DateInterval('P'.substr($value, 2));
+            $interval = new DateInterval('P' . substr($value, 2));
             $isNegative = true;
         } else {
             $interval = new DateInterval($value);
             $isNegative = $interval->invert;
         }
 
-        if ($interval->y > 0) {
+        if ($showYears || (is_null($showYears) &&$interval->y > 0)) {
```
samdark: Formatting! `&&$inteval`. Same in more lines ↓

samdark: `=== null` should be used instead of `is_null`

framework/i18n/Formatter.php
```diff
@@ -828,18 +828,75 @@ public function asRelativeTime($value, $referenceTime = null)
...
+     *    - `null|true|false` `showYears` will be displayed part for years data. Behaves like as `showSeconds`.
+     * Note that if `value` parameter equals zero value, will be displayed `0 seconds`, if `showSeconds` not `false`.
+     * And if `showYears` is `false` that is mean low priority parts(`showMonths`, `showDays`, `showDays`, ...) will be
+     * set to false also.
+     * @return string the formatted duration.
+     * @since 2.0.10
+     */
+    public function asDurationTime($value, $options = [])
+    {
+        if (is_null($value)) {
```
samdark: `=== null` should be used instead of `is_null`

samdark (changes requested):

framework/i18n/Formatter.php
githubjeka: Why? Is requirement to yii2 code?

samdark: It's overall better to use simple checks instead of function calls where possible. More explicit, less opcodes.

SilverFire:
Btw, the implementation is not very smart.

Take a look at implementation in `moment.js` https://github.com/moment/moment/blob/develop/src/lib/duration/humanize.js

githubjeka:
@SilverFire I think about. But it will pull the more changes. And so, I added a few lines :) May be better create new issue...

SilverFire:
It's fine to do it in this pull request.

SilverFire:
Considering that we implement completely different behavior, I suggest to leave `asDuration()` almost changes and implement `asRoudedDuration()` that will create `DateInterval` object and round it so only required parameters are left

githubjeka:
It will called DecoratorDateInterval class, formatter should use it.... #эко_меня_понесло

githubjeka:
>Does it make any sense to show something like 1 year 11 months 3 seconds ago?

Yes you can - via `\yii\i18n\Formatter::asRelativeTime`. It isn't about this issue.

I added more tests. Now, I suggest leave as is or please explain what do you want to exactly?

cebe:
Sorry for the late reply. It seems that the requirements for this method are not very clear so before putting more work into this we should agree on what the different methods should do. Also the interface for specifying which part of the duration to show is not simple. It ususally does not make sense to display e.g. only year and seconds so you display `1 year, 16 seconds` for a timestamp that was actually a year, a month and some hours + 16 seconds. What should instead applied is a parameter that defines the precision. There are also multiple options on how to handle that:

- precision is `year`:
  - we could display `0 years` if duration <= 1 year, `1 year` if <= 2 year etc.
  - alternative: display `x sec` if duration <= 1 min, `x min` if duration <= 1 h, and switch to above display if we reached year
- precision is `month`:
  - `1 year, 2 month` if > duration is over a year or just display the months if it is not.
  - same couting as above could be applied. (or not)

There are a lot of options here that should be discussed to define a method interface (or multiple methods).
See also 
- http://www.icu-project.org/apiref/icu4j/com/ibm/icu/text/MeasureFormat.html
- http://www.icu-project.org/apiref/icu4j/com/ibm/icu/text/RelativeDateTimeFormatter.html
guess intl has something we could use here. Need to find it.

Issue #12659, opened 2016-10-04
FR, ENH: Show alternative when command not found


mdmunir:
![screenshot from 2016-10-04 12 52 42](https://cloud.githubusercontent.com/assets/5828252/19064042/cfb631fe-8a31-11e6-9a3b-b4d542c9469f.png)

samdark:
Excellent idea.

cebe:
Solution is implemented. Thanks for the suggestion and your contribution!

PR #12661, opened 2016-10-04
Show alternative when command not found

mdmunir:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues | #12659 |

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -73,6 +73,7 @@ Yii Framework 2 Change Log
 - Enh #12382: Changed `yii\widgets\MaskedInput` to use `jQuery` instead of `$` to prevent conflicts (samdark)
 - Enh #12440: Added `yii\base\Event::offAll()` method allowing clear all registered class-level event handlers (klimov-paul)
 - Enh #12580: Make `yii.js` comply with strict and non-strict javascript mode to allow concatenation with external code (mikehaertl)
+- Enh #12659: Show alternative when command not found (mdmunir)
 - Enh: Method `yii\console\controllers\AssetController::getAssetManager()` automatically enables `yii\web\AssetManager::forceCopy` in case it is not explicitly specified (pana1990, klimov-paul)
 
 
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -73,7 +73,7 @@ Yii Framework 2 Change Log
 - Enh #12382: Changed `yii\widgets\MaskedInput` to use `jQuery` instead of `$` to prevent conflicts (samdark)
 - Enh #12440: Added `yii\base\Event::offAll()` method allowing clear all registered class-level event handlers (klimov-paul)
 - Enh #12580: Make `yii.js` comply with strict and non-strict javascript mode to allow concatenation with external code (mikehaertl)
-- Enh #12659: Show alternative when command not found (mdmunir)
+- Enh #12659: Suggest alternatives when console command typed was not found (mdmunir)
 - Enh: Method `yii\console\controllers\AssetController::getAssetManager()` automatically enables `yii\web\AssetManager::forceCopy` in case it is not explicitly specified (pana1990, klimov-paul)
 
 
--- a/framework/console/Application.php
+++ b/framework/console/Application.php
@@ -9,6 +9,7 @@ namespace yii\console;
 
 use Yii;
 use yii\base\InvalidRouteException;
+use yii\helpers\Inflector;
 
 // define STDIN, STDOUT and STDERR if the PHP SAPI did not define them (e.g. creating console application in web env)
 // http://php.net/manual/en/features.commandline.io-streams.php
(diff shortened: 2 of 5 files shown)


SilverFire:
Console application class looks overcomplicated now.

Maybe it's better to create c separate class `AlternativesFinder` and instantiate it in the `catch` block?

mdmunir:
The most complicated code are get list of command. The same code also use in `HelpController`. Much better if we separated it to another class.

framework/CHANGELOG.md
```diff
@@ -73,6 +73,7 @@ Yii Framework 2 Change Log
 - Enh #12382: Changed `yii\widgets\MaskedInput` to use `jQuery` instead of `$` to prevent conflicts (samdark)
 - Enh #12440: Added `yii\base\Event::offAll()` method allowing clear all registered class-level event handlers (klimov-paul)
 - Enh #12580: Make `yii.js` comply with strict and non-strict javascript mode to allow concatenation with external code (mikehaertl)
+- Enh #12659: Show alternative when command not found (mdmunir)
```
samdark: Suggest alternatives when console command typed was not found

framework/console/Application.php
```diff
@@ -180,8 +181,156 @@ public function runAction($route, $params = [])
             $res = parent::runAction($route, $params);
             return is_object($res) ? $res : (int)$res;
         } catch (InvalidRouteException $e) {
-            throw new Exception("Unknown command \"$route\".", 0, $e);
+            $message = "Unknown command \"$route\"";
+            if($alternatives = $this->findAlternatives($route)){
```
samdark: Formatting!

framework/console/Application.php
```diff
@@ -180,8 +181,156 @@ public function runAction($route, $params = [])
...
+                $message .= count($alternatives) > 1 ?  "\nDid you mean one of these?" : "\nDid you mean this?";
```
samdark: In case of only one alternative it could be outputted right away as `Did you mean %s?`

framework/console/Application.php
```diff
@@ -180,8 +181,156 @@ public function runAction($route, $params = [])
...
+
+    /**
+     *
+     * @param string $route
+     * @return array
+     */
+    public function findAlternatives($route)
+    {
+        $commands = $this->getCommands();
+        $threshold = 1e3;
```
samdark: What's `1e3`?

samdark: Need a comment explaining why exactly this value is used.

framework/console/Application.php
```diff
@@ -180,8 +181,156 @@ public function runAction($route, $params = [])
...
+
+        foreach ($commands as $command) {
+            $lev = levenshtein($route, $command);
+            if ($lev <= strlen($route) / 3 || strpos($command, $route) !== false) {
+                $alternatives[$command] = isset($alternatives[$command]) ? $alternatives[$command] - $lev : $lev;
+            }
+        }
+        
+        $alternatives = array_filter($alternatives, function ($lev) use ($threshold) { 
+            return $lev < 2 * $threshold;
```
samdark: What's `2`?

samdark (changes requested):

samdark:
I agree that it could be a good idea to move logic into different class.

SilverFire:
Suggest to create `ModuleReflection` class which will know all about modules reflection.

And something like `ActionAlternativesFinder`, which will be responsible for alternatives search:

```php
$alternatives = (new ActionAltertativesFinder([
     'allActions' => (new ModuleReflection($this->module))->getActions(),
]))->findFor($action);
```

framework/console/Application.php
mdmunir: Copied from Sympony https://github.com/symfony/console/blob/master/Application.php#L971.
I am also not understand what actually is :grin:

framework/console/Application.php
```diff
@@ -182,155 +182,20 @@ public function runAction($route, $params = [])
...
+                $commands = RouteHelper::getRoutes(true);
+                if ($alternatives = RouteHelper::find($route, $commands)) {
+                    if (count($alternatives) == 1) {
+                        $message .= sprintf("\nDid you mean \"%s\"?", reset($alternatives));
+                    } else {
+                        $message .= "\nDid you mean one of these?\n    " . implode("\n    ", $alternatives);
                     }
                 }
+            } catch (\Exception $exc) {
+
```
samdark: Looks suspicious...

framework/console/controllers/HelpController.php
```diff
@@ -52,7 +53,21 @@ public function actionIndex($command = null)
...
+                    $commands = RouteHelper::getRoutes(true);
+                    if ($alternatives = RouteHelper::find($command, $commands)) {
+                        if (count($alternatives) == 1) {
+                            $message .= sprintf("\nDid you mean \"%s\"?", reset($alternatives));
+                        } else {
+                            $message .= "\nDid you mean one of these?\n    " . implode("\n    ", $alternatives);
+                        }
+                    }
+                } catch (\Exception $exc) {
+
```
samdark: Same here. Looks suspicious. Do we need to ignore exceptions here?

samdark (commented):
Overall it's much better but I'm not sure @SilverFire meant a static helper when he suggested a separate class...

SilverFire:
No, I didn't mean helper in this case

SilverFire:
## Separate non-static classes

### Pros:

 - Can create class-scope constants to explain Levenshtein magic numbers
 - Low coupling
 - Easier to test - the class has no dependencies
 - Easier to re-use
 - (Generally it's about SOLID)

### Cons:

 - Not so convenient to call as Helpers

## Helper class

### Pros:

 - Easy to call - good for top-level public API

### Cons:

 - Mixed responsibility
 - Higher coupling (`Yii::$app` calls)
 - Harder to test
 - Just a bunch of function, grouped in some namespace. Functional programming masked behind OOP-style

___

Considering it will not be a frequently used API, I don't see profits from making it as a helper.

samdark:
Added link to the summary to https://github.com/yiisoft/yii2/blob/master/docs/internals/design-decisions.md

cebe:
imo it is much easier to implement this by introducing an exception class that handles the case and cover the display in the ErrorHandler. See #12675 for a draft implementation.

samdark:
Agree. Exception solution looks cleaner.

samdark:
@mdmunir what do you think about @cebe proposal?

cebe:
I am closing this PR in favor of  #12675, which is now complete. Thank you for your contribution @mdmunir!

Issue #11835, opened 2016-06-27
NTEXT is being deprecated in MSSQL


Lifesnoozer:
The column type `NTEXT` is being deprecated in MSSQL, it's recommended to use `NVARCHAR(MAX)` instead (https://msdn.microsoft.com/en-us/library/ms187993.aspx), `\yii\db\mssql\QueryBuilder` should probably use `NVARCHAR(MAX)` instead of `NTEXT` for `Schema::TYPE_TEXT` to avoid problems.

SilverFire:
Resolved in #12663

PR #12663, opened 2016-10-04
MSSQL: Fix to NTEXT is being deprecated

githubjeka:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11835 fixed by the PR |

Original issue request to use `NVARCHAR(MAX)` . But MAX keyword is unknown for old MSSQL . 
To avoid error `Incorrect syntax near 'MAX'` I use 4000 that is [max value for nvarchar](https://msdn.microsoft.com/en-us/library/ms186939%28SQL.90%29.aspx)

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -58,6 +58,7 @@ Yii Framework 2 Change Log
 - Enh #11494: `yii.reloadableScripts` now support wildcards with `*` character (silverfire)
 - Enh #11658: Added argument to `yii\grid\ActionColumn::urlCreator` callback, which holds reference to the column instance (klimov-paul)
 - Enh #11804: Added `yii\behaviors\AttributeTypecastBehavior` for maintaining of strict ActiveRecord attribute types (klimov-paul)
+- Enh #11835: Fixed NTEXT in `\yii\db\mssql\QueryBuilder::$typeMap` that is being deprecated in MSSQL(githubjeka)
 - Enh #11950: Improve `yii\helpers\BaseArrayHelper::keyExists()` speed (egorio)
 - Enh #11979: Added `yii\mutex\OracleMutex` which implements mutex "lock" mechanism via Oracle locks (zlakomanoff)
 - Enh #12028: Add -h|--help option to console command to display help information (pana1990)
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -58,7 +58,7 @@ Yii Framework 2 Change Log
 - Enh #11494: `yii.reloadableScripts` now support wildcards with `*` character (silverfire)
 - Enh #11658: Added argument to `yii\grid\ActionColumn::urlCreator` callback, which holds reference to the column instance (klimov-paul)
 - Enh #11804: Added `yii\behaviors\AttributeTypecastBehavior` for maintaining of strict ActiveRecord attribute types (klimov-paul)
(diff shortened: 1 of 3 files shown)


SilverFire:
Should be mentioned in UPGRADE.md

SilverFire (approved):

framework/CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ Yii Framework 2 Change Log
 - Enh #11494: `yii.reloadableScripts` now support wildcards with `*` character (silverfire)
 - Enh #11658: Added argument to `yii\grid\ActionColumn::urlCreator` callback, which holds reference to the column instance (klimov-paul)
 - Enh #11804: Added `yii\behaviors\AttributeTypecastBehavior` for maintaining of strict ActiveRecord attribute types (klimov-paul)
+- Enh #11835: Fixed NTEXT in `\yii\db\mssql\QueryBuilder::$typeMap` that is being deprecated in MSSQL(githubjeka)
```
samdark: Space missing before `(`

samdark (approved):

githubjeka:
Corrected.

framework/UPGRADE.md
```diff
@@ -15,6 +15,14 @@ php composer.phar self-update
 php composer.phar global require "fxp/composer-asset-plugin:^1.2.0"
 ```
 
+Upgrade from Yii 2.0.9
+----------------------
+
+* MSSQL: NTEXT data type marked as deprecated. (https://msdn.microsoft.com/en-us/library/ms187993.aspx)
```
samdark: `NTEXT` data type was marked as deprecated (https://msdn.microsoft.com/en-us/library/ms187993.aspx) so `\yii\db\Schema::TYPE_TEXT` was changed from `ntext` to `nvarchar`. By default `nvarchar` defines a string of length 4000. If you need `Schema::TYPE_TEXT = nvarchar(max)`, you can adjust it in your code. It's not like that by default because `MAX` keyword is new to SQL Server 2005.

samdark (changes requested):

githubjeka:
Corrected.

samdark (approved):

SilverFire:
Thank you

Lifesnoozer:
SQL Server 2005 isn't even supported by Microsoft anymore, the last SQL Server release that didn't support `nvarchar(MAX)` is 13(!) years old. The difference between `nvarchar(4000)` and `nvarchar(MAX)` is pretty significant. Why add confusing behaviour just to support old and unsupported software?

In case you need support for something older than SQL Server 2005, you should have to change the mapping on your own, not the other way around.

samdark:
Probably you're right but I've seen many unsupported versions running for years in huge enterprises.

samdark:
Worth doing as you've said though w/ max as default.

samdark:
@SilverFire can you handle the change?

githubjeka:
Because then Breaks BC = yes. My MSSQL production server was throwed ` Incorrect syntax near 'MAX'`.

githubjeka:
But the current implementation suggests Breaks BC anyway. Because `NTEXT != nvarchar(4000)`. But nvarchar(4000) safer I thought so.

Lifesnoozer:
`nvarchar(MAX)` gives the expected behaviour, so that makes it the sane and safe default, new developers going in will expect the text option to work with "unlimited" text, a limit of 4000 is pretty low for a text field.

githubjeka:
I agree

SilverFire:
Done: e6eed9a

PR #12664, opened 2016-10-04
Added support for wildcards for `optional` at `yii\filters\auth\AuthMethod`

mg-code:
This option now works similar to `only` and `except`.
Logic of `only` and `except` was updated in release 2.0.9 by @klimov-paul, but this option left as it is.

| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |


SilverFire:
What is the use case?

mg-code:
Authenticator can be set in application configuration.
For some controllers or modules it can be optional.

mg-code:
Updated: 
This option now works similar to `only` and `except`.
Logic of `only` and `except` was updated in release 2.0.9 by @klimov-paul, but this option left as it is.

klimov-paul:
Resolved by commit 2a9a9972775d0b6d90dfe69ba8e846e9a9de094a

PR #12676, opened 2016-10-05
break test

githubjeka:
| Tests pass?   | no


tests/framework/db/QueryBuilderTest.php
```diff
@@ -1024,6 +1024,7 @@ public function conditionProvider()
 
             // between
             [ ['between', 'id', 1, 10], '[[id]] BETWEEN :qp0 AND :qp1', [':qp0' => 1, ':qp1' => 10] ],
+            [ ['between', '[[id]]', 1, 10], '[[id]] BETWEEN :qp0 AND :qp1', [':qp0' => 1, ':qp1' => 10] ],
```
cebe: why do you quote id here? it is quoted by yii already.

githubjeka: Yes the construction like '{{tableName}}.[[id]]'` ignored in the tests. 
I had complexities when I make changes in commits like #12678

cebe:
I do not really see a problem here, table and column names are quoted by QueryBuilder if they are raw, but left alone if already quoted. The lower layer (Command) will care for the quoting later.

PR #12678, opened 2016-10-05
Added ability to use column range in BETWEEN operator

### LINKED ISSUE #11611: Query builder between 2 columns
Предлагаю фичу, ну или как минимум дописать документацию.

В документации:
http://www.yiiframework.com/doc-2.0/yii-db-query.html

Написано:

> between: operand 1 should be the column name, and operand 2 and 3 should be the starting and ending values of the range that the column is in. For example, ['between', 'id', 1, 10] will generate id BETWEEN 1 AND 10.

Можно ли сделать поддержку следующего?
$this->where(['between', $value, 'column_from', 'column_to'])->all();

Как я читал SQL подобную конструкцию поддерживает. MySQL поддерживает подобный синтаксис.

``` SQL
SELECT * FROM  `mytable` 
WHERE 4 
BETWEEN  `from` AND  `to`
```

githubjeka:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | Fixes #11611 |

Now you can use `['between', ['from', 'to'], 999]` to generate:

``` sql
SELECT `id` FROM `table` WHERE 999 BETWEEN `from` AND `to`
```


SilverFire:
Replaced by https://github.com/yiisoft/yii2/pull/15641

PR #12682, opened 2016-10-05
Documentation ActiveQueryInterface::one()

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

Currently there is an issue when using `BaseActiveRecord::findOne()` because it uses the return type for `ActiveQueryInterface::one()` which is inherited from `QueryInterface::one()` which returns `array|false`

https://scrutinizer-ci.com/g/yiisoft/yii2/issues/master/files/framework/db/BaseActiveRecord.php?selectedAuthors%5B0%5D=angeldelcaos%40gmail.com&orderField=path&order=asc&honorSelectedPaths=0

### DIFF
--- a/framework/db/ActiveQueryInterface.php
+++ b/framework/db/ActiveQueryInterface.php
@@ -32,9 +32,9 @@ interface ActiveQueryInterface extends QueryInterface
     /**
      * Executes query and returns a single row of result.
      * @param Connection $db the DB connection used to create the DB command.
-     * If null, the DB connection returned by [[modelClass]] will be used.
+     * If `null`, the DB connection returned by [[modelClass]] will be used.
      * @return ActiveRecordInterface|array|null a single row of query result. Depending on the setting of [[asArray]],
-     * the query result may be either an array or an ActiveRecord object. Null will be returned
+     * the query result may be either an array or an ActiveRecord object. `null` will be returned
      * if the query results in nothing.
      */
     public function one($db = null);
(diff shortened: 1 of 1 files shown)


framework/db/ActiveQueryInterface.php
```diff
@@ -28,6 +28,16 @@
      * @return $this the query object itself
      */
     public function asArray($value = true);
+    
+    /**
+     * Executes query and returns a single row of result.
+     * @param Connection $db the DB connection used to create the DB command.
+     * If null, the DB connection returned by [[modelClass]] will be used.
```
SilverFire: null -> <code>\`null\`</code>

framework/db/ActiveQueryInterface.php
```diff
@@ -28,6 +28,16 @@
...
+     * @return ActiveRecordInterface|array|null a single row of query result. Depending on the setting of [[asArray]],
+     * the query result may be either an array or an ActiveRecord object. Null will be returned
```
SilverFire: null -> <code>\`null\`</code>

SilverFire (approved):

framework/db/ActiveQueryInterface.php
Faryshta: thanks, done.

SilverFire (approved):

cebe:
Thank you!

SilverFire:
Should we mention this change in UPGRADE.md?

cebe:
it is only a documentation change.

samdark:
Isn't it adding a method to interface?

cebe:
yes, but it is overriding a method from the parent interface, the only change is the PHPDoc that now includes ActiveRecord and replaces false with null.

Issue #12489, opened 2016-09-07
PHP 7.1 changes JSON_PRETTY_PRINT indentation?


cebe:
### What steps will reproduce the problem?

Failing test, see https://travis-ci.org/yiisoft/yii2/jobs/158310722#L431-L436
### What is the expected result?

see normal indentation as in other PHP versions.
### What do you get instead?

Wrong indentation, however I am unable to reproduce this on 3v4l: https://3v4l.org/m3Xjo
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.10-dev |
| PHP version | 7.1 on travis |

cebe:
looks like it is related to https://bugs.php.net/bug.php?id=66025 but that context does not really apply here.

rob006:
It is PHP bug: https://3v4l.org/reS06

I've made a PR that skips test with failing `Json::encode()` for PHP 7.1: #12684 - now all tests are green.

SilverFire:
As @rob006 mentioned [here](https://github.com/yiisoft/yii2/pull/12684#issuecomment-257151382), the bug is solved but Travis still has outdated PHP version, so 2c3e3f3 can not be reverted right now

rob006:
@SilverFire [This fix](https://github.com/php/php-src/commit/c34de0b61c68bd77e324b869760a53abead4c5ed) is not yet released - this should be fixed when 6th RC of PHP 7.1 comes out (~November 10th).

SilverFire:
Ah, fine. Thank you for clarifying!

PR #12684, opened 2016-10-05
Skip failing test in PHP 7.1

rob006:
Refs https://github.com/yiisoft/yii2/issues/12489

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ php:
 
 matrix:
   allow_failures:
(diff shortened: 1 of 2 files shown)


LAV45:
This exception can be added when php 7.1, will be stable version.

rob006:
This is already fixed in PHP core by https://github.com/php/php-src/pull/2173.

cebe:
Thanks for checking!

PR #12670, opened 2016-10-04
Added Method yii\web\Request::getDomain()

ivanovyordan:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11309 |

Also modified `yii\web\Request::getHostInfo()` to use the new method.
P.S. Not sure what kind of unit test I need to write to cover the new methods.

PR #12686, opened 2016-10-06
Add `yii\web\Request::getHostName()`

### LINKED ISSUE #11309: Request class cann't return ONLY domain (without scheme)
`Yii::$app->request->getHostInfo()` returns 'http://example.com'. There is no method for handling $_SERVER['HTTP_HOST'] and returning ONLY domain, e.g. `Yii::$app->request->getDomain()` which returns only 'example.com'.
I think will be better to add this method in yii\webRequest class for working in same yii-interface with superglobal variable $_SERVER.

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11309 |

fixes #11309
replaces #12670


samdark (approved):

framework/web/Request.php
```diff
@@ -554,13 +555,29 @@ public function getHostInfo()
...
     {
+        $this->_hostName = null;
         $this->_hostInfo = $value === null ? null : rtrim($value, '/');
     }
 
+    /**
+     * Returns the host part of the current request URL.
+     * Value is calculated from current [[getHostInfo()|hostInfo]] property.
+     * @return string|null hostname part of the request URL (e.g. `www.yiiframework.com`)
+     * @see getHostInfo()
```
cebe: need `@since 2.0.10`

cebe (changes requested):
looks great, except the `@since` annotation.

framework/web/Request.php
rob006: Done.

cebe (approved):
ready to be merged for 2.0.10 imo.

samdark:
Thank you!

Issue #7936, opened 2015-03-31
[PHP7] Consider to deprecate yii\base\Object as object might become soft-reserved


suralc:
PHP 7 will [soft reserve](https://wiki.php.net/rfc/reserve_even_more_types_in_php_7) `object` as a possible typehint for a 7.x release (or may never use it at all). Deprecating `yii\base\Object` in favor of `yii\base\BaseObject` (or a better name) early (in 2.1) might reduce some pain once installations are upgraded if php should decide to use the object name.

I'm not sure if the patch below is the best way to go forward as the deprecation notice is only triggered if `parent::__construct()` is called in all child classes. However triggering the deprecation message on autoload makes it harder to locate the class using the old inheritance chain.
The deprecation is based on the fact that php in the future **might** use `object`; causing the class declaration to cause a fatal error.

Possible patch (only targets the framework itself not extensions, no tests for the deprecation notice).:
https://github.com/yiisoft/yii2/compare/master...suralc:deprecate-php7

I did not create this as a pull-request as the vote is not yet formally closed. The vote on soft-reserving `object` sits at 38:10 in favor so it is likely going to pass.

cebe:
Thanks for pointing this out, need to check it.

qiangxue:
Will the name be case  sensitive or not?

samdark:
I guess both will be reserved. We'll see when it will be accepted and someone will create a patch.

cebe:
Class names are case insensitive so it would not matter in which case we write it.

suralc:
@cebe 

> [...] the following typenames are now reserved:
Resource
Object
Mixed
Numeric

http://marc.info/?l=php-internals&m=142833998201071&w=2

cebe:
Thanks for the info, have already seen it. This is causing much trouble. Yii 2.0.x will not run at all on php7 because of this... Not sure how to solve the issue yet. As far as I see, there is no way without breaking BC. Renaming the base class of all things is a huge thing...

ref: https://twitter.com/cebe_cc/status/585573046808879104

cebe:
maybe something like this could work:

```php
<?php

namespace yii\base;

class BaseObject { /* current implementation */ }

if (version_compare(PHP_VERSION, '7.0.0', '<')) {
    // ensure code works on php < 7.0.0 to not break BC
    class_alias('yii\base\BaseObject', 'yii\base\Object', false);
}
```

suralc:
[...]

Issue #12154, opened 2016-08-13
Unbound composer constraints are unsafe


rob006:
This does not affect framework itself, but in extensions and templates in `composer.json` there are many dependencies defined like: 

``` json
"yiisoft/yii2": "*",
```

This setting is dangerous and even composer complaining about it (yii2-app-advanced template):

```
$ composer validate
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
./composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
require.yiisoft/yii2 : unbound version constraints (>=2.0.6) should be avoided
require.yiisoft/yii2-bootstrap : unbound version constraints (*) should be avoided
require.yiisoft/yii2-swiftmailer : unbound version constraints (*) should be avoided
```

When Yii 2.1 will be released (which can break BC), composer will just upgrade Yii to v2.1 even if used extensions and application itself will be incompatible with new version. There are already problems with this: https://github.com/yiisoft/yii2-bootstrap/issues/150

Yii extensions, app templates and extension template generated by Gii should use more safe constraints, like:

``` json
"yiisoft/yii2": "~2.0.0",
```

klimov-paul:
This is debatable matter.

> When Yii 2.1 will be released (which can break BC), composer will just upgrade Yii to v2.1 even if used extensions and application itself will be incompatible with new version.

This is actually is unknown. It could be that particular extension will be not compatible with the next major release, but it could it will. Unless there is no centaintly for such matter I prefer to have flexible version constaint to avoid possible version conflict in the future.

For example.
Assume we have a major release of `yii2-framework` - '2.1'. We have extension `yii2-mongodb` which have dependency specified as `yii2-framework: "~2.0.0"`, while it can work with `yii2-framework: "~2.1.0"`.
And we have extension `yii2-jui` with dependency `yii2-framework: "~2.1.0"` because it relies on some BC changes at main repo.
While  `yii2-mongodb` can function fine with `yii2-framework: "~2.1.0"` you are unable to install in this way due to the version restiction. And thus you are unbale to install both `yii2-jui` and `yii2-mongodb`.

The answer to the question "why there could be no version of `yii2-mongodb`, which refers `yii2-framework 2.1`" is it may be not necessary. Since there is no code changes for such migration - no release has been made.

Thus developer ends with unresolving situation of version conflict.

[...]

PR #12699, opened 2016-10-07
Deprecate `yii\base\Object`

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #7936 |

See https://github.com/yiisoft/yii2/issues/7936#issuecomment-128629578


samdark (approved):
Well done.

klimov-paul:
I can't see how this solves the problem.

The original issue was about `Object` will be a reserved keyword at PHP, thus it can not be used for the name of the class. This PR does not solve this problem: `Object` class still exists and used in Yii heirarchy.
At the moment `object` becomes reserved keyword the entire framework code will crush with this PR as well as without it.

To solve the issue `Object` class should be renamed to `BaseObject` so `object` keyword is excluded from hierarchy.

rob006:
> To solve the issue Object class should be renamed to BaseObject so object keyword is excluded from hierarchy.

We can't do this in 2.0.x, because this breaks BC. This PR is only first step - after this will be merged I open another PR against 2.1.x line, that removes `Object`. 

Merge this into 2.0.x allows you to prepare to BC break - extensions and apps could start extending from `yii\base\BaseObject` instead of `yii\base\Object` and use `instanceof yii\base\BaseObject` instead of `instanceof yii\base\Object` - this will be compatible with 2.1.x line and >2.0.11.

klimov-paul:
> extensions and apps could start extending from yii\base\BaseObject instead of yii\base\Object and use instanceof yii\base\BaseObject instead of instanceof yii\base\Object - this will be compatible with 2.1.x line and >2.0.11.

At what is the sense of it? Back at issue #12154 you were so eager to establish strict version constraints for any Yii extension, despite all my arguments it may be unnecessary. Here you are talking about extensions base on Yii 2.0.x should extend `BaseObject` preparing for `Object` removal, which will comes only with major version 2.1.
For me your words here and at #12154 contradict each other.

If some extension is prepared to use 2.1 branch it can also use `BaseObject` from that branch, for the extension based on 2.0.x using of `BaseObject` provides no benefit.

klimov-paul:
If you are upgrading your extension or project from Yii 2.0.x to Yii 2.1.x, you still need to change version constrait at `composer.json`.
Do you really think it would be challenge to developer to replace all `yii\base\Object` usage to `yii\base\BaseObject` at the same commit?

Thoulah:
@klimov-paul This commit gives us, developers, the chance to move to the new class `BaseObject` already. This will guarantee a more painless transition to 2.1 (for this particular issue) (or whenever Object will be removed definitely, depending on PHP choices). Without this change we will have to change from Object when required by others, instead of now already. I don't think preparing our code is a bad thing.

klimov-paul:
> This will guarantee a more painless transition to 2.1

It is a noble goal, but this particular change is one of the most unsignificant on this path.
There will be a large 'upgrade' instructions about migration from 2.0.x to 2.1.x anyway - a checklist, which should be processed for the package before it can use `2.1`. Renaming `Object` to `BaseObject` will be one of the easiest parts of it.

> after this will be merged I open another PR against 2.1.x line, that removes Object. 

This also does not make much sense. As I already said replacing `Object` with `BaseObject` is a trivial task. However applying such diff to `2.1` brach will make merging changes from `2.0.x` more problematic. While in `2.0.x` classes will extend `Object`, at `2.1` they will use `BaseObject`. This can cause conflicts and code breaks when merging `2.0.x` into `2.1`.
We are already have such issues around `Object::className()` method, which was already removed at `2.1` branch.

This is a one of the final steps, which should be applied to `2.1` before its release. It make no sense to deal with it now.  That is why issue #7936 is kept opened.

I do not want to sound ungrateful, but all these changes produces more troubles then fixes. If you expect this PR and the 'future' one will speed up `2.1` release - you are wrong - they will not.

Anyway, I can not make a decision for the whole core team - if they want this to be merged - I can not stop them.

rob006:
> At what is the sense of it? Back at issue #12154 you were so eager to establish strict version constraints for any Yii extension, despite all my arguments it may be unnecessary.

How can you say that strict version constraints for Yii extensions may be unnecessary? #7936 proves that #12154 is necessary - if 2.1 introduce BC break in most essential Yii class, how can you assume that apps and extensions for Yii 2.0 will work on Yii 2.1?

> For me your words here and at #12154 contradict each other.

I don't see how they would contradict. All code that uses `Object` will be incompatible with Yii 2.1 - that's why changes proposed in #12154 are necessary. 
This PR creates compatibility layer between 2.0 and 2.1, which simplify maintaining extensions (you can have fewer differences between lines for Yii 2.0 and Yii 2.1) and gives you possibility to write new code in way that will be compatible with Yii 2.1, which gives you less work in future - for me these are big benefits.
IMO this PR and #12154 perfectly complement each other.

> I do not want to sound ungrateful, but all these changes produces more troubles then fixes.

Can you specify what exactly problems this particular PR creates?

dynasource:
> This commit gives us, developers, the chance to move to the new class BaseObject already. This will guarantee a more painless transition to 2.1 (for this particular issue) (or whenever Object will be removed definitely, depending on PHP choices). Without this change we will have to change from Object when required by others, instead of now already. I don't think preparing our code is a bad thing.

@Thoulah clearly states why this PR is useful. Its useful for preparing developers that Object will deprecated in the future. Not only will they see it in the @docs, they will also be hinted for deprecation by ie. PhpStorm.

With respect to this PR. Its looks good in my opinion. @rob006, can you update the changelog?

rob006:
I have another idea how to solve this, I will try to propose alternative PR this weekend.

rob006:
Alternative: #12929

samdark:
I think alternative one is better.

rob006:
@samdark ?

samdark:
?

rob006:
Why do you close this PR?

samdark:
Because you proposed alternative solution which looked better?

rob006:
According to discussion in #12929 2.1 will still contain `Component` and `Object` to simplify using of this traits, so #12929 is not really alternative for this PR - we still need to rename `Object` to `BaseObject`.

rob006:
Ready to merge IMO.

dynasource (approved):
well done

klimov-paul:
I still can't see any benefit from this change. I would rather exclude `yii\base\Object` from Yii hierarchy at this stage with UPGRADE note.
There is only one use case, which will be broken if `Object` is replaced by `BaseObject` - the `instanceof Object` checks. We have made more sugnificant BC breaks already.
`Component`, `Event` and other classes should extend `BaseObject` instead of `Object`, while `Object` should remain for BC keeping.

samdark:
The change is to gracefully prepare developers for future changes i.e. your suggested change may go into 2.1 instead.

klimov-paul:
I can not see why it should wait until 2.1.
See the upgrade notes created so far in 2.0.x:
https://github.com/yiisoft/yii2/blob/master/framework/UPGRADE.md

We have changed the methods signatures, broke the interfaces and so on. What hard can bring the change to `BaseObject` in such comparison?

Besides the problem described at #7936 goes nowhere, once new PHP 7.x will be released Yii2 will become incompatible with it. I do not like the idea to put PHP version restiction in following way:

```
5.5.0 < Yii2 < 7.2.0
```

He we are talking about support of future PHP versions or stuck with current one.

samdark:
> We have changed the methods signatures, broke the interfaces and so on. What hard can bring the change to BaseObject in such comparison?

Immediately broken projects? What's bad in having this change done gracefully?

> Besides the problem described at #7936 goes nowhere, once new PHP 7.x will be released Yii2 will become incompatible with it. 

Why? Yii itself will be compatible with it. Projects may be not if they've used Object directly. The change as it is proposed ensures that future projects started with 7.2+ will be compatible and that current projects will get deprecation warnings in IDEs such as PhpStorm. Also it's not breaking anything compared to nowadays state i.e. existing projects work fine if PHP version is less than 7.2. By the time people would upgrade they'll see deprecation warnings and update their code.

klimov-paul:
> Immediately broken projects? What's bad in having this change done gracefully?

Only one rare usecase will be broken: check like this:

```php
if ($foo instance of \yii\base\Object) {}
```

> Why? Yii itself will be compatible with it.

Becase `object` word will be reserved by PHP 7 as syntax keyword and once `yii\base\Component` is loaded PHP will break with error "unexpected object keyword at line 100":
https://github.com/yiisoft/yii2/blob/master/framework/base/Component.php#L100

Lines like following:

```php
class Foo extends Object {}
```

will be equal to following:

```php
class Foo extends array {}
```

dynasource:
let's keep it gracefully & non BC till the time PHP7 is really complaining about it.

klimov-paul:
> till the time PHP7 is really complaining about it.

klimov-paul:
> till the time PHP7 is really complaining about it.

And what will happen when it will?
I can not consider such words as a solid argument - I can as well say that we should not change anything around `yii\base\Object` untill PHP will actually complain on it.

rob006:
> Becase object word will be reserved by PHP 7

There is no strict plans for that so we don't know when it will happen. `object` keyword could be reserved long after Yii 2.0 EOL, there is no point in introducing BC break "just in case".

klimov-paul:
Then I can see absolutely no reason for this PR to be merged. 
It just introduce redundant inheritance layer, while it is unclear if it actually needed or not.
Following your logic - this matter should be left untouched until PHP will actually reserve `object` as a keyword.

dynasource:
the reason & logic behind the PR is to prepare the community to already use BaseObject in both old & new projects

klimov-paul:
> the reason & logic behind the PR is to prepare the community to already use BaseObject in both old & new projects

For me such reason does not justifies overcomplication of the source code and performance degradation.
I can see no problem of possible fix for usages of `yii\base\Object` to `yii\base\BaseObject` in my Yii project.
If such refactoring is a challenge for some developer - it is better for him to cease coding right away.

rob006:
>  It just introduces redundant inheritance layer, while it is unclear if it actually needed or not.

We know that it is needed - sooner or later we will rename `Object` to `BaseObject`, so there will be BC break. This PR make it less painful.

dynasource:
> overcomplication of the source code and performance degradation.

Both are temporary and really negligible. Also, this is how deprecation works.

klimov-paul:
> We know that it is needed - sooner or later we will rename Object to BaseObject

I am aware, so what is the problem to fix this now? 

> This PR make it less painful.

I don't get. What is the problem of search for lines `yii\base\Object` over the code and replace them by `yii\base\BaseObject`? The entire change will take maximum of 1 hour. Why this is so painful? Why do you need several years to prepare for that?

>Both are temporary and really negligible. Also, this is how deprecation works.

It seems you have forgotten the original issue: this change makes NOTHING. It can not prevent the crash once PHP evolves taking `object` as keyword. It exist to support one-hour-consuming fix, which by some unknown reason is so complex at your glance, so you need several years to prepare for it.

We should either prevent the future crash or ignore the possible threat.

Note that I do not propose tp **remove** `yii\base\Object` - I propose to exlude it from standard classes hierarchy, so Yii 2.0.12 will run successfully on new PHP version, while any custom class extending `yii\base\Object` continues working. The only use case, which will be broken is check `instanceof Object`, which is rare.

dynasource:
> It seems you have forgotten the original issue: this change makes NOTHING. It can not prevent the crash once PHP evolves taking object as keyword.

Isn't it you who seems to forget that (new) users will benefit from this PR by already using BaseObject in their new projects.

samdark:
What @klimov-paul explains actually makes sense. Yii itself would break if we'll leave `Component extends Object` as is so changing to `Component extends BaseObject` is a must in order to Yii 2.0 to work when such deprecation would take place.

So to sum up:

- If we want Yii 2.0 not to break when `Object` will stop working we have to actually use `BaseObject` everywhere.
- If we are OK that Yii 2.0 won't work on that PHP version, current deprecation approach is OK.

klimov-paul:
Benefit what exactly?

rob006:
> If we want Yii 2.0 not to break when Object will stop working we have to actually use BaseObject everywhere.

We could do this when it is actually needed, for example if PHP 7.3 introduce `object` keyword. This PR make it really easy, and people will have some time to prepare for this.

samdark:
@rob006 what was mentioned about `instanceof Object` by @klimov-paul should be in upgrade notes.

rob006:
Done.

samdark:
Moved to next milestone since @cebe want to check possible implications more.

rob006:
Bump: https://github.com/yiisoft/yii2/issues/7936#issuecomment-306042456

samdark:
@rob006 will review for next milestone.

cebe:
I'd have adjusted and merged this PR, but it conflicts with master, sorry @rob006.
I took only the docs replacement commit and re-did the renaming in #14478, closing this one.

PR #12701, opened 2016-10-08
Catch `Throwable` in `yii\base\ErrorHandler::handleException()`

### LINKED ISSUE #12619: catch Throwable in ErrorHandler::handleException
If there's an exception during the rendering of the error page, ErrorHandler will catch it and display a simple message: https://github.com/yiisoft/yii2/blob/2.0.9/framework/base/ErrorHandler.php#L117

This should catch `Throwable` in PHP 7 as well.

rob006:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12619 |

Fixes #12619

https://3v4l.org/DlA8p
https://3v4l.org/FL1fa


dynasource:
when doing a CTRL F through the framework, I see in total 47 occurrences of ```php catch (\Exception $e) ```
It would be nice to fix all in 1 commit

rob006:
I don't think anything else needs to be fixed. We should not blindly catch all `Throwable` in framework .

dynasource:
could you elaborate on your decision to do this for ErrorHandler & Widget only?

rob006:
Both are for easier debugging. Not really sure about `Widget`, I'm actually considering reverting it and leave this only for `ErrorHandler`.

Related discussion: https://github.com/symfony/symfony/pull/18765#issuecomment-219031688

samdark:
What's the point doing that in a widget?

rob006:
> What's the point doing that in a widget?

I thought that without this output will be corrupted on PHP error on widget rendering, but it was a wrong assumption. I will remove it.

samdark (approved):

samdark:
Merged. Thank you!

Issue #12390, opened 2016-08-29
Additional sql query with junction table


krebse:
### What steps will reproduce the problem?

My project contain three table : post , category , category_post_assignment

I get ID of the category and search for related posts

```
$category = Category::find()
            ->with('posts')
            ->where(['id' => 2])
            ->one();
```

```
$category = Category::find()
            ->where(['id' => 2])
            ->one();

$posts = $category->posts;
```

In the normal first query get category from relevant table
`SELECT * FROM `tbl_category` WHERE `id`=2`

second query get relevant posts id from assignment table
`SELECT * FROM `tbl_category_post_assignment` WHERE `category_id`=2`

third query get the posts from post table
`SELECT * FROM `tbl_post` WHERE `id` IN (1, 2, ....)`

The issue is after second query if don't find relevant post the third query still execute in additional query.
`SELECT * FROM `tbl_post` WHERE 0=1`
### What is the expected result?

When don't find relevant post in second query the third query shouldn't be execute.
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.9 |
| PHP version | 7.0 |

klimov-paul:
Possible fix: #12708

cebe:
Fixed, thanks for reporting!

Issue #6373, opened 2014-12-04
EmptyDataProvider


leandrogehlen:
there are situations, where it is necessary, to show, initially, empty datagrid . 
After the user to select required filters, then it will execute query.

Currently, it is necessary to execute a query that not return data.

I don't think it is necessary to execute this query,  knowing that any data it will returned?

How can we solve this problem?

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/53e1018648af6b79d429687ce685c5d7eb6fd054

leandrogehlen:
The label problems was solved, but the how to show,  initially, empty datagrid whit any dataprovder?

SilverFire:
Use the approach suggested by @samdark

leandrogehlen:
It's working, but the issue is trigger query to show empty grid.
Currently, is necessary to write something like: 
```php
$query = Movfinan::find()->where(['id' => null]); // this is not necessary, but, currently, it's the unique way

$dataProvider = new ActiveDataProvider([
     'query' => $query     
]);

```

SilverFire:
Well yeah...

Faryshta:
what do you suggest @leandrogehlen ?

leandrogehlen:
@cebe  suggested create a method: 

```php
public function forceEmpty()
{
    $this->setModels([]); 
    $this->setKeys([]); 
    $this->setTotalcount(0);
}
``` 
But this is the solution to ArrayDataProvider.

i think better to add `autoload` property into ```BaseDataProvider```, then all provider can have this feature

Faryshta:
can you make a pr?

leandrogehlen:
Yes, but today, i can't.
Could be tomorow?

Faryshta:
@leandrogehlen yup, its not obligatory :+1:

leandrogehlen:
I was trying to implement this feature, but i think that the implementation made by @SilverFire  resolved the issue.

```php
public function actionIndex()
{
    if (Yii::$app->request->get('empty', false)) {
        $dataProvider = new ArrayDataProvider([
            'modelClass' => Person::className()
        ]); 
    } else {
        $query = Person::find();
        $dataProvider = new ActiveDataProvider([
            'query' => $query
        ])
    }      

    return $this->render('index', [
        'dataProvider' => $dataProvider
    ]);
}

```

Faryshta:
[...]

PR #12708, opened 2016-10-10
Added `QueryInterface::emulateExecution()`

klimov-paul:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | yes |
| Fixed issues | #6373, #12390 |

Added `QueryInterface::emulateExecution()`, which allows preventing of the actual query execution.
This allows to cancel `DataProvider` preventing search query execution in case search model is invalid:

``` php
    {
        $query = Item::find();

            'query' => $query,
        ]);

        $this->load($params);

        if (!$this->validate()) {
            $query->where('0=1');
            $query->emulateExecution(); // No SQL execution will be done
            return $dataProvider;
        }
```

This also fix unecessary query in case of `via()` usage. See #12390.

Note: this PR produces minor BC break as it alters `yii\db\QueryInterface`, however the implementation of new method is present at `yii\db\QueryTrait` - thus it is less likely to produce a trouble.

### DIFF
--- a/framework/db/QueryInterface.php
+++ b/framework/db/QueryInterface.php
@@ -254,14 +254,14 @@ interface QueryInterface
     public function offset($offset);
 
     /**
-     * Cancels the query actual execution, preventing interaction with data storage.
-     * After this method is triggered, methods, returning query results like [[one()]], [[all()]], [[exists()]]
+     * Sets whether to emulate query execution, preventing any interaction with data storage.
+     * After this mode is enabled, methods, returning query results like [[one()]], [[all()]], [[exists()]]
      * and so on, will return empty or false values.
      * You should use this method in case your program logic indicates query should not return any results, like
      * in case you set false where condition like `0=1`.
      * @param boolean $value whether to prevent query execution.
-     * @return $this the query object itself
+     * @return $this the query object itself.
      * @since 2.0.11
      */
-    public function preventExecution($value = true);
+    public function emulateExecution($value = true);
 }
--- a/framework/db/QueryInterface.php
+++ b/framework/db/QueryInterface.php
@@ -252,4 +252,15 @@ interface QueryInterface
      * @return $this the query object itself
      */
     public function offset($offset);
+
+    /**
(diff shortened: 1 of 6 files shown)


framework/db/QueryInterface.php
```diff
@@ -252,4 +252,15 @@ public function limit($limit);
...
+    /**
+     * Cancels the query actual execution, prevention interaction with data storage.
+     * After this method is triggered, methods, returning query results like [[one()]], [[all()]], [[exists()]]
+     * and so on will return empty or false values.
+     * You should use this method in case your program logic indicates query should not return any results, like
+     * in case you set false where condition like `0=1`.
+     * @return $this the query object itself
+     * @since 2.0.11
+     */
+    public function cancel();
```
samdark: I think the name of the method isn't OK since we as users do not clearly separate actual SQL query and getting the result. I propose using another name such as `skipQuery()` (not sure it's the best one).

samdark (changes requested):

framework/db/QueryInterface.php
klimov-paul: Renamed to `preventExecution()`

samdark (approved):

samdark:
Still looks like a hacky solution but I can't suggest anything better...

framework/db/QueryInterface.php
dynasource: another possibility is 'simulate'

SilverFire (approved):

samdark:
Oh! Now I like the naming.

dynasource:
emulate is even better

SilverFire (approved):

dynasource:
I've reviewed it. I don't know with this solution. It works but it seems to be an overkill for realizing the objective 'DataProvider preventing search query execution'. Aren't there any other solutions out there that are less obtrusive?

cebe (approved):
Good changes! Need a CHANGELOG and UPGRADE note. Will add them myself. Merging now.

PR #12717, opened 2016-10-11
clear cached filemtime call, fix #12713

### LINKED ISSUE #12713: FileDependency is prone to fail in certain situations
### Minimal example:

config.php:

``` php
'components' => [
    'cache' => [
        'class' => 'yii\caching\FileCache',
    ],
    'volatileCache' => 'yii\caching\ArrayCache',
    ...
],
```

controller.php:

``` php
// insert appropriate namespace and use statements here
class TestController extends \yii\console\Controller
{
    private $_parsedContent;


    private function loadFromCache()
    {
        // Cache is volatile -> get('isModified') evaluates to false when first executed
        if (!Yii::$app->volatileCache->get('isModified')) {
            var_dump("Invalid Cache!");
            $this->_parsedContent = json_decode(Yii::getAlias('@app/someconfig.json'), true);
            Yii::$app->volatileCache->set('isModified', $value = true,
               $duration = 5, // invalidate after 5 seconds...
               new FileDependency(['fileName' => Yii::getAlias('@app/someconfig.json')])); // ...or when config file changes
        }

        return $this->_parsedContent;
    }

    public function actionIndex()
    {
        $this->loadFromCache();
        sleep(2);  // Modify the file during this sleep **outside** of php (in an editor)
        $this->loadFromCache(); // Expected output: Invalid Cache! (But thats not happening)
        sleep(4); //  Total waittime now 6 seconds -> Cache invalidates after 5
        $this->loadFromCache(); // Output: Invalid Cache! (But because of the regular timeout)
    }
}
```
### Problem Analysis

Interestingly enough the above example does work when using the `FileCache` component in favor of the `ArrayCache` component. Since I could not trace any direct call on `clearstatcache`, I assume that the stat cache is implicitly cleared by PHP after either of the calls to `file_put_contents`, `chmod` and `touch` inside `FileCache::setValue`.

Probably the easiest way to fix this bug is to explicitly call `clearstatcache` inside `FileDependency::generateDependencyData`, i.e. modify it as follows:

``` php
protected function generateDependencyData($cache)
{
    if ($this->fileName === null) {
        throw new InvalidConfigException('FileDependency::fileName must be set');
    }

    clearstatcache(false, Yii::getAlias($this->fileName));
    return @filemtime(Yii::getAlias($this->fileName));
}
```
### Additional info

Yii: 2.0.9
PHP: 7.0.8-0ubuntu0.16.04.2 (cli)

P.S. This issue may qualify for https://yourfirstpr.github.io/

SG5:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes/no |
| Fixed issues | #12713 |

### DIFF
--- a/framework/caching/FileDependency.php
+++ b/framework/caching/FileDependency.php
@@ -41,6 +41,7 @@ class FileDependency extends Dependency
             throw new InvalidConfigException('FileDependency::fileName must be set');
         }
 
+        clearstatcache(false, Yii::getAlias($this->fileName));
         return @filemtime(Yii::getAlias($this->fileName));
     }
 }
--- a/framework/caching/FileDependency.php
+++ b/framework/caching/FileDependency.php
@@ -41,7 +41,8 @@ class FileDependency extends Dependency
             throw new InvalidConfigException('FileDependency::fileName must be set');
         }
 
-        clearstatcache(false, Yii::getAlias($this->fileName));
-        return @filemtime(Yii::getAlias($this->fileName));
+        $fileName = Yii::getAlias($this->fileName);
+        clearstatcache(false, $fileName);
+        return @filemtime($fileName);
     }
 }
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -47,6 +47,7 @@ Yii Framework 2 Change Log
 - Bug #12605: Make 'safe' validator work on write-only properties (arthibald, CeBe)
 - Bug #12629: Fixed `yii\widgets\ActiveField::widget()` to call `adjustLabelFor()` for `InputWidget` descendants (coderlex)
 - Bug #12649: Fixed consistency of `indexBy` handling for `yii\db\Query::column()` (silverfire)
+- Bug #12713: Fixed `yii\caching\FileDependency` to clear stat cache before reading filemtime (SG5)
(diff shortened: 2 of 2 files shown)


samdark:
Have you verified it helps?

SG5:
yes, i've got 3 time `Invalid Cache!` message

Coksnuss:
I can confirm that it works as well (as stated in my issue).
If the proposed fix is accepted, maybe the ticket can be moved to the 2.0.10 milestone?

framework/caching/FileDependency.php
```diff
@@ -41,6 +41,7 @@ protected function generateDependencyData($cache)
             throw new InvalidConfigException('FileDependency::fileName must be set');
         }
 
+        clearstatcache(false, Yii::getAlias($this->fileName));
         return @filemtime(Yii::getAlias($this->fileName));
```
cebe: getAlias() should be called once and the result be stored in a variable.

cebe (changes requested):

samdark (approved):

cebe (approved):

dynasource (approved):

samdark:
Thanks!

PR #12724, opened 2016-10-12
Duplicate hash link [skip ci]

mdmunir:
http://www.yiiframework.com/doc-2.0/guide-start-installation.html#installing-yii


samdark:
Thanks!

Issue #12726, opened 2016-10-12
Provide `version` property for the `Application`


klimov-paul:
Extracted from https://github.com/yiisoft/yii2-debug/issues/143

It may be useful to provide `version` property for the `Application` class, which can be used for logging and debugging.

There should be a posibility to setup version as a PHP callback, which will return its actual value.

It also make sense to inroduce this property at `Module` level.

klimov-paul:
Possible implementation proposed at #12727

klimov-paul:
Actually there is already `yii\base\Application::version` public field, which can be used:
https://github.com/yiisoft/yii2/blob/master/framework/base/Application.php#L100

So the question is if it should be converted to virtual property for more flexibility.

klimov-paul:
Resolved by commit 8d05bd9

dynasource:
thanks!

PR #12727, opened 2016-10-12
Add `Module::version`

klimov-paul:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | yes |
| Tests pass? | yes |
| Fixed issues | #12726 |


framework/base/Module.php
```diff
@@ -283,6 +300,58 @@ public function setLayoutPath($path)
...
+     *
+     * ```php
+     * function (Module $module) {
+     *     //return string|integer
+     * }
+     * ```
+     *
+     * @since 2.0.11
+     */
+    public function setVersion($version)
```
samdark: Why do we need a setter?

framework/base/Module.php
```diff
@@ -283,6 +300,58 @@ public function setLayoutPath($path)
...
+
+    /**
+     * Returns default module [[version]] based last modification time of the [[basePath]].
+     * Child class may override this method to provide more specific version detection.
+     * @return string|integer module version.
+     * @since 2.0.11
+     */
+    protected function defaultVersion()
+    {
+        $versionTimestamp = @filemtime($this->getBasePath());
```
samdark: It would return local file date instead of date of a file from repository, right?

samdark (commented):
It's it better to introduce real version reading or a meta file?

framework/base/Module.php
klimov-paul: It allows you to set version as a PHP callback, which will return the actual value intead of extending class. It is crucial for `Application`, allowing specification of version logic in config file:

```php
return [
    'name' => 'MyApplication',
    'basePath' => dirname(__DIR__),
    'version' => function ($application) {
        return exec("cd {$application->basePath}; git rev-parse HEAD");
    },
    // ...
];
```

framework/base/Module.php
klimov-paul: Although there is a problem here: current implementation will pick up last modification date from `runtime` directory as well, which will cause it to always reference current date time, as `runtime` is modified almost at any program run.

rob006: > current implementation will pick up last modification date from runtime directory as well

Did you test it? Modifications in `runtime` should not change mtime of root directory.

klimov-paul: Improved to use VCS files as a timestamp source.

klimov-paul:
Just noticed: `yii\base\Application` already has a `version` public property:
https://github.com/yiisoft/yii2/blob/master/framework/base/Application.php#L100

klimov-paul:
Legacy `Application::version` removed.
Thus this PR produces a slight BC break, as it will return version based on timestamp instead of `1.0` be default.

rob006:
Timestamp-based versions are useless anyway. mtime is completely unrelated to real application version - changes in application not always change mtime of project root.

klimov-paul:
> Timestamp-based versions are useless anyway. 

Make sense. I have changed default value to `1.0` keeping the BC.

samdark (approved):

framework/base/Module.php
dynasource: there shouldnt be any calculated defaultVersion. Thats to arbitrary. Its either the user who is defining it or no version.

dynasource:
> Just noticed: yii\base\Application already has a version public property

This shows that its an idea to open a topic to further cleanup Application & Module. Bootstrap for instance, is something that is useful too for Modules.

SilverFire (approved):

samdark (approved):

PR #12732, opened 2016-10-13
Add is_dir() validation for BaseFileHelper::findFiles()

zalatov:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

Add is_dir() validation for BaseFileHelper::findFiles().
Sometimes it throws exception when file/directory is deleted or moved while executing script.

### DIFF
--- a/framework/helpers/BaseFileHelper.php
+++ b/framework/helpers/BaseFileHelper.php
@@ -411,7 +411,7 @@ class BaseFileHelper
             if (static::filterPath($path, $options)) {
                 if (is_file($path)) {
                     $list[] = $path;
-                } elseif (!isset($options['recursive']) || $options['recursive']) {
+                } elseif (is_dir($path) && (!isset($options['recursive']) || $options['recursive'])) {
                     $list = array_merge($list, static::findFiles($path, $options));
                 }
             }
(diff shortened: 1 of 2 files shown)


SilverFire (changes requested):
It will be good to have a test of this case.

CHANGELOG is required

framework/helpers/BaseFileHelper.php
```diff
@@ -403,7 +403,7 @@ public static function findFiles($dir, $options = [])
             if (static::filterPath($path, $options)) {
                 if (is_file($path)) {
                     $list[] = $path;
-                } elseif (!isset($options['recursive']) || $options['recursive']) {
+                } elseif (is_dir($path) && (!isset($options['recursive']) || $options['recursive'])) {
```
samdark: Need to make sure it works well with symlinks.

samdark: On Windows as well...

zalatov:
@SilverFire,
I don't know how test it.
We catch this error when two concurrent requests.
While first request already deleting files in dir, second request only start find files.

@samdark,
From the documentation:
`is_dir() Note that this functions follows symbolic links. It will return true if the file is actually a symlink that points to a directory.`

zalatov:
Added row to CHANGELOG

cebe:
> But... in CHANGELOG last version is 2.08, therefore this pull request is 2.0.11.

can you rebase your branch on master?

zalatov:
Pulled last changes from master

SilverFire:
Merged, thank you!

SilverFire:
https://github.com/yiisoft/yii2/commit/50953c2960549a727a9650d0db36cf21f0c542e9

PR #12737, opened 2016-10-13
Skip pjax if data-pjax="0"

arisk:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

This makes the behavior of links with data-pjax="0" consistent regardless of data-method property.
See https://github.com/yiisoft/jquery-pjax/blob/master/jquery.pjax.js#L78

### DIFF
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -164,7 +164,7 @@ window.yii = (function ($) {
                 pjaxContainer,
                 pjaxOptions = {};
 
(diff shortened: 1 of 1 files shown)


SilverFire:
I'd prefer to case `false` to `undefined` before the conditions block and get rid of redundant conditions

SilverFire:
Duplicates #13118, partially merged though. Thank you

Issue #11627, opened 2016-05-25
Some confusions with ListView


Locustv2:
There are some confusions with the ListView attributes and some minor issues as well.

Assume this ListView here:

``` php
<?= ListView::widget([
    'dataProvider' => $dataProvider,
    'summary' => 'My Summary',
    'summaryOptions' => ['class' => 'bold'],
    'options' => ['class' => 'bordered bgwhite'],
    'itemOptions' => ['class' => 'item'],
    'itemView' => '_list-item',
    'showOnEmpty' => false,
]); ?>
```

First of all, when i modify the `summary` attribute, the content is not rendered in a `tag` but it's just text. So `summaryOptions` does not apply the class to the summary text.

And next, according to the docs, `showOnEmpty` is :

> Whether to show the list view if $dataProvider returns no data.

But when the `showOnEmpty` attribute is `false`, i still see the listview with the empty text inside. When `true` i see the listview with nothing inside.

Shouldn't it be:
- when `false` => don't show the list view container at all
- when `true` => show the listview with the empty text?
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.8 |
| PHP version | 7.0.2 |
| Operating system | Windows 7 |

klimov-paul:
Resolved by #12740

PR #12740, opened 2016-10-13
Document and test ListView.showOnEmpty, see #11627

mytskine:
This addresses #11627 by completing the documentation without changing the behavior. Some basic unit tests are added on ListView to ensure this won't change unintentionnally.

That means a ListView widget with an empty provider and `showOnEmpty=true` will still display an empty element (by default, an empty "div.list-view").

### DIFF
--- a/framework/widgets/BaseListView.php
+++ b/framework/widgets/BaseListView.php
@@ -69,7 +69,9 @@ abstract class BaseListView extends Widget
      */
     public $summaryOptions = ['class' => 'summary'];
     /**
-     * @var boolean whether to show the list view if [[dataProvider]] returns no data.
+     * @var boolean whether to show an empty list view if [[dataProvider]] returns no data.
+     * The default value is false which displays an element according to the `emptyText`
+     * and `emptyTextOptions` properties.
      */
     public $showOnEmpty = false;
     /**
--- a/tests/framework/widgets/ListViewTest.php
+++ b/tests/framework/widgets/ListViewTest.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace yiiunit\framework\widgets;
+
+use yii\data\ArrayDataProvider;
+use yii\widgets\ListView;
+
+/**
+ * @group widgets
+ */
+class ListViewTest extends \yiiunit\TestCase
+{
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->mockApplication();
+    }
+
+    public function testEmptyListShown()
+    {
+        $dataProvider = new ArrayDataProvider([
+            'allModels' => [],
+        ]);
+
+        ob_start();
+        echo ListView::widget([
+            'dataProvider' => $dataProvider,
+            'showOnEmpty' => false,
+            'emptyText' => "Nothing at all",
+        ]);
+        $actualHtml = ob_get_clean();
+
+        $this->assertTrue(strpos($actualHtml, "Nothing at all") !== false, "displays the empty message");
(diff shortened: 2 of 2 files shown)


dynasource:
Thanks for putting effort in creating unit tests. Would you be willing to put the tests in smaller units to prove your usecase?

mytskine:
I'm sorry but I don't understand your answer. What do you mean with "to
prove your usecase"? What I want to "prove" is that the documentation of
this property was bad, as can be seen in #11627.

Does "put the tests in smaller units" means that tests should have only
one assertion?

I only put assertions on the output controlled by the `showOnEmpty`
property, so I can't see any smart way to split those units.  If that's
your wish, I'll remove the non-essential assertions. I'd keep
"!showOnEmpty displays the empty message" and "showOnEmpty contains no
text".

dynasource:
if its your goal to have these unit tests in core (to prove/cover your usecase), you should further cut these in blocks like doing assertings on subfunctions instead of the run() function

mytskine:
That makes no sense because the property I want to test is only used in
`ListView.run()` — more exactly, in the abstract parent class
[BaseListView](https://github.com/mytskine/yii2/blob/master/framework/widgets/BaseListView.php#L124)

dynasource:
you're right, didnt know run() was that bad. Perhabs your can also update ListView::run and move the logic for setting $content to its own getter (protected function getContent)

Then its easier to test your usecase in a more contained enviroment.

SilverFire:
@mytskine the tests are failing now. Is it expected?

mytskine:
These errors already occured before I changed anything. I thought it was
because of my environment. Moreover, the error messages involve
ActiveRecordTestTrait which is out of my concern.

The widgets tests, where I added on test case, do not fail.

The official yii2/master has the same failing tests:
<https://travis-ci.org/yiisoft/yii2>

mytskine:
The unrelated unit tests are now fixed on master, see #12745. I've rebased the branch.

tests/framework/widgets/ListViewTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+        ob_start();
+        echo ListView::widget([
+            'dataProvider' => $dataProvider,
+            'showOnEmpty' => false,
+            'emptyText' => "Nothing at all",
+        ]);
+        $actualHtml = ob_get_clean();
+
+        $this->assertTrue(strpos($actualHtml, "Nothing at all") !== false, "displays the empty message");
```
dynasource: these assertions are a but difficult to read. Can you use `assertContains` or `assertNotContains`

framework/widgets/BaseListView.php
```diff
@@ -69,7 +69,9 @@
      */
     public $summaryOptions = ['class' => 'summary'];
     /**
-     * @var boolean whether to show the list view if [[dataProvider]] returns no data.
+     * @var boolean whether to show an empty list view if [[dataProvider]] returns no data.
+     * The default value is false which displays an element according to the `emptyText`
+     * and `emptyTextOptions` properties.
```
dynasource: I agree with that this is confusion. With a `false` for showOnEmpty something is actually "shown". Can we fix this problem in this PR  by renaming the $property (for 2.1 and future releases)

dynasource (changes requested):
thanks for your efforts. Note: There are some conflicts with the master branch.

samdark (approved):

cebe (approved):

SilverFire:
Merged with simplified tests (https://github.com/yiisoft/yii2/commit/0261a5498f1fbbe9e84dc181c6fc8cada94f25bd).

Thank you!

Issue #12748, opened 2016-10-14
Migration Generator - incorrect foreign key field


tsanchev:
### What steps will reproduce the problem?

`yii migrate/create create_products_combine_with_table  --fields="combine_with_id:bigPrimaryKey:unsigned,product_id:bigInteger:unsigned:notNull:foreignKey(products)"`

Some clarification - primary key of `products` table is `product_id`
It is generating `id` but should generate `product_id`
### What is the expected result?

```
public function up()
{
    ...

    // add foreign key for table `products`
    $this->addForeignKey(
        'fk-products_combine_with-product_id',
        'products_combine_with',
        'product_id',
        'products',
        'product_id',
        'CASCADE'
    );
}
```
### What do you get instead?

```
public function up()
{
    ...

    // add foreign key for table `products`
    $this->addForeignKey(
        'fk-products_combine_with-product_id',
        'products_combine_with',
        'product_id',
        'products',
        'id',
        'CASCADE'
    );
}
```
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.10-dev |
| PHP version | 5.6.9 |

SilverFire:
@mdmunir

MKiselev:
@tsanchev everything works as described in the documentation, migration mechanism is unknown about what field in the products table is the primary key.
------
Всё работает как и описано в документации, механизму миграции неизвестно о том, какое поле в таблице products является первичным ключом.

@SilverFire предлагаю поставить label:type:enhancement и в качестве решения могу предложить когда foreignKey прописывается с параметрами, вторым параметром передавать название поля, т.е. вот так: ```foreignKey(products, product_id)```
Если решение устраивает, то готов приступить к реализации.

tsanchev:
@MKiselev i think it would be better if the migration mechanism gets the primary key field directly from the database of the given table instead of an extra parameter to foreignKey

MKiselev:
@tsanchev the mechanism of creating migration nowhere use the database component, so I think members unlikely to support the idea.

samdark:
There are 2 ways of solving the problem:

1. Specify FK column as parameter.
2. Get FK column from DB (that would require querying existing schema).

samdark:
These two could be combined. If key isn't specified, get it from DB.

MKiselev:
@samdark Ok.

I resolve this issue.

MKiselev:
I was made this logic:
[...]

PR #12770, opened 2016-10-18
Issue #12748 fix

MaksimKiselev:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12748 |

### DIFF
--- a/framework/console/controllers/MigrateController.php
+++ b/framework/console/controllers/MigrateController.php
@@ -166,7 +166,9 @@ class MigrateController extends BaseMigrateController
     public function beforeAction($action)
     {
         if (parent::beforeAction($action)) {
-            $this->db = Instance::ensure($this->db, Connection::className());
+            if ($action->id !== 'create') {
+                $this->db = Instance::ensure($this->db, Connection::className());
+            }
             return true;
         } else {
             return false;
@@ -293,8 +295,8 @@ class MigrateController extends BaseMigrateController
 
             $foreignKeys[$firstTable . '_id']['table'] = $firstTable;
             $foreignKeys[$secondTable . '_id']['table'] = $secondTable;
-            $foreignKeys[$firstTable . '_id']['column'] = 'id';
-            $foreignKeys[$secondTable . '_id']['column'] = 'id';
+            $foreignKeys[$firstTable . '_id']['column'] = '';
+            $foreignKeys[$secondTable . '_id']['column'] = '';
             $table = $firstTable . '_' . $secondTable;
         } elseif (preg_match('/^add_(.+)_columns?_to_(.+)_table$/', $name, $matches)) {
             $templateFile = $this->generatorTemplateFiles['add_column'];
@@ -317,16 +319,21 @@ class MigrateController extends BaseMigrateController
             $relatedTable = $foreignKey['table'];
             if (empty($relatedColumn)) {
                 $relatedColumn = 'id';
-                $relatedTableSchema = $this->db->getTableSchema($relatedTable);
-                if ($relatedTableSchema !== null) {
-                    $primaryKeyCount = count($relatedTableSchema->primaryKey);
-                    if ($primaryKeyCount > 1) {
-                        $this->stdout("Related table for field {$column} already exist, but primary key is a composite. Default name will be used for related field\n", Console::FG_YELLOW);
-                    } elseif ($primaryKeyCount < 1) {
-                        $this->stdout("Related table for field {$column} already exist, but do not have primary key. Default name will be used for related field.\n", Console::FG_YELLOW);
-                    } else {
-                        $relatedColumn = $relatedTableSchema->primaryKey[0];
+                try {
+                    $this->db = Instance::ensure($this->db, Connection::className());
+                    $relatedTableSchema = $this->db->getTableSchema($relatedTable);
+                    if ($relatedTableSchema !== null) {
+                        $primaryKeyCount = count($relatedTableSchema->primaryKey);
+                        if ($primaryKeyCount > 1) {
+                            $this->stdout("Related table for field \"{$column}\" exists, but primary key is composite. Default name \"id\" will be generated for related field\n", Console::FG_YELLOW);
+                        } elseif ($primaryKeyCount < 1) {
+                            $this->stdout("Related table for field \"{$column}\" exists, but does not have a primary key. Default name \"id\" will be used for related field.\n", Console::FG_YELLOW);
+                        } else {
+                            $relatedColumn = $relatedTableSchema->primaryKey[0];
+                        }
                     }
+                } catch (\ReflectionException $e) {
+                    $this->stdout("Cannot initialize database component to try read referenced table schema for field \"{$column}\". Default name \"id\" will be generated for related field.\n", Console::FG_YELLOW);
                 }
             }
             $foreignKeys[$column] = [
--- a/framework/console/controllers/MigrateController.php
+++ b/framework/console/controllers/MigrateController.php
@@ -293,6 +293,8 @@ class MigrateController extends BaseMigrateController
 
             $foreignKeys[$firstTable . '_id']['table'] = $firstTable;
             $foreignKeys[$secondTable . '_id']['table'] = $secondTable;
+            $foreignKeys[$firstTable . '_id']['column'] = 'id';
+            $foreignKeys[$secondTable . '_id']['column'] = 'id';
             $table = $firstTable . '_' . $secondTable;
         } elseif (preg_match('/^add_(.+)_columns?_to_(.+)_table$/', $name, $matches)) {
             $templateFile = $this->generatorTemplateFiles['add_column'];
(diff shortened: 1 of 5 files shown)


framework/console/controllers/MigrateController.php
```diff
@@ -166,9 +166,7 @@ public function optionAliases()
     public function beforeAction($action)
     {
         if (parent::beforeAction($action)) {
-            if ($action->id !== 'create') {
-                $this->db = Instance::ensure($this->db, Connection::className());
-            }
+            $this->db = Instance::ensure($this->db, Connection::className());
```
samdark: I'd leave it as is here and duplicate the call in `create` action so it won't trigger errors if it doesn't need database.

framework/console/controllers/MigrateController.php
```diff
@@ -312,11 +312,28 @@ protected function generateMigrationSourceCode($params)
             $table = mb_strtolower($matches[1], Yii::$app->charset);
         }
 
-        foreach ($foreignKeys as $column => $relatedTable) {
+        foreach ($foreignKeys as $column => $foreignKey) {
+            $relatedColumn = $foreignKey['column'];
+            $relatedTable = $foreignKey['table'];
+            if (empty($relatedColumn)) {
+                $relatedColumn = 'id';
+                $relatedTableSchema = $this->db->getTableSchema($relatedTable);
```
samdark: You need DB here so call `$this->db = Instance::ensure($this->db, Connection::className());`.

framework/console/controllers/MigrateController.php
```diff
@@ -312,11 +312,28 @@ protected function generateMigrationSourceCode($params)
...
+                if ($relatedTableSchema !== null) {
+                    $primaryKeyCount = count($relatedTableSchema->primaryKey);
+                    if ($primaryKeyCount > 1) {
+                        $this->stdout("Related table for field {$column} already exist, but primary key is a composite. Default name will be used for related field\n", Console::FG_YELLOW);
```
samdark:
```php
"Related table for field \"{$column}\" exists, but primary key is composite. Default name \"id\" will be generated for related field\n"
```

framework/console/controllers/MigrateController.php
```diff
@@ -312,11 +312,28 @@ protected function generateMigrationSourceCode($params)
...
+                    } elseif ($primaryKeyCount < 1) {
+                        $this->stdout("Related table for field {$column} already exist, but do not have primary key. Default name will be used for related field.\n", Console::FG_YELLOW);
```
samdark:
```php
"Related table for field \"{$column}\" exists, but does not have a primary key. Default name \"id\" will be used for related field.\n"
```

framework/console/controllers/MigrateController.php
```diff
@@ -363,10 +380,15 @@ protected function parseFields()
 
             foreach ($chunks as $i => &$chunk) {
                 if (strpos($chunk, 'foreignKey') === 0) {
-                    preg_match('/foreignKey\((\w*)\)/', $chunk, $matches);
-                    $foreignKeys[$property] = isset($matches[1])
-                        ? $matches[1]
-                        : preg_replace('/_id$/', '', $property);
+                    preg_match('/foreignKey\((\w*)\s?(\w*)\)/', $chunk, $matches);
```
samdark: Why the regexp change?

framework/console/controllers/MigrateController.php
MKiselev: How to check that the component DB may be initialized here?

Я имею в виду, что если в конфигурации приложения db не задан, то DI не может разрешить эту зависимость и получаю исключение.

samdark: https://github.com/yiisoft/yii2/pull/12770#discussion_r83872685

You can probably try-catch exception and fall back to default generation scenario...

framework/console/controllers/MigrateController.php
MKiselev: Regular expression changed for support setting field name for reference key by using whitespace like this: ```foreignKey(table_name reference_key)```

Now FK may by defined in three ways:
```table_id:foreignKey(table_name reference_key)```
```table_id:foreignKey(table_name)```
```table_id:foreignKey```

**Comment by MKiselev** [comment]
Видимо не так CHANGELOG.md надо править, я стянул с мастера и поправил...
Ещё и имя с битбакета в гите настроено.

**Comment by MKiselev** [comment]
@samdark have any comments for last commit?

File: framework/console/controllers/MigrateController.php [hunk 6]
```diff
@@ -293,8 +295,8 @@ protected function generateMigrationSourceCode($params)
 
             $foreignKeys[$firstTable . '_id']['table'] = $firstTable;
             $foreignKeys[$secondTable . '_id']['table'] = $secondTable;
-            $foreignKeys[$firstTable . '_id']['column'] = 'id';
-            $foreignKeys[$secondTable . '_id']['column'] = 'id';
+            $foreignKeys[$firstTable . '_id']['column'] = '';
```
samdark: I'd still fall back to just `id`.

framework/console/controllers/MigrateController.php
```diff
@@ -317,16 +319,21 @@ protected function generateMigrationSourceCode($params)
...
+                try {
+                    $this->db = Instance::ensure($this->db, Connection::className());
+                    $relatedTableSchema = $this->db->getTableSchema($relatedTable);
+                    if ($relatedTableSchema !== null) {
+                        $primaryKeyCount = count($relatedTableSchema->primaryKey);
+                        if ($primaryKeyCount > 1) {
+                            $this->stdout("Related table for field \"{$column}\" exists, but primary key is composite. Default name \"id\" will be generated for related field\n", Console::FG_YELLOW);
+                        } elseif ($primaryKeyCount < 1) {
+                            $this->stdout("Related table for field \"{$column}\" exists, but does not have a primary key. Default name \"id\" will be used for related field.\n", Console::FG_YELLOW);
+                        } else {
```
samdark: This is the most common variant so I'd check for `=== 1` first.

framework/console/controllers/MigrateController.php
```diff
@@ -317,16 +319,21 @@ protected function generateMigrationSourceCode($params)
...
-                    } else {
-                        $relatedColumn = $relatedTableSchema->primaryKey[0];
+                try {
+                    $this->db = Instance::ensure($this->db, Connection::className());
+                    $relatedTableSchema = $this->db->getTableSchema($relatedTable);
+                    if ($relatedTableSchema !== null) {
+                        $primaryKeyCount = count($relatedTableSchema->primaryKey);
+                        if ($primaryKeyCount > 1) {
+                            $this->stdout("Related table for field \"{$column}\" exists, but primary key is composite. Default name \"id\" will be generated for related field\n", Console::FG_YELLOW);
+                        } elseif ($primaryKeyCount < 1) {
```
samdark: It's better to write it as `=== 0` since it can't be negative anyway.

samdark (changes requested):

framework/console/controllers/MigrateController.php
MKiselev: @samdark  if here use just ```id```, then here https://github.com/MKiselev/yii2/blob/4126e46c3889373a54d59a22fe49c5b485d370c1/framework/console/controllers/MigrateController.php#L320 check will be false.

MKiselev:
@samdark update code.

framework/console/controllers/MigrateController.php
samdark: Ah. OK then.

framework/console/controllers/MigrateController.php
```diff
@@ -293,8 +293,10 @@ protected function generateMigrationSourceCode($params)
                 ]
             );
 
-            $foreignKeys[$firstTable . '_id'] = $firstTable;
-            $foreignKeys[$secondTable . '_id'] = $secondTable;
+            $foreignKeys[$firstTable . '_id']['table'] = $firstTable;
+            $foreignKeys[$secondTable . '_id']['table'] = $secondTable;
+            $foreignKeys[$firstTable . '_id']['column'] = '';
```
samdark: Maybe it should be set to `null` and then the check could be ` === null`?

framework/console/controllers/MigrateController.php
samdark: Docs should be updated accordingly.

samdark (changes requested):
Looks much better now. Added a bit more inline comments. Also need a changelog.

MKiselev:
@samdark code was updated. My english is not very good, please check docs. RU docs I will update after accept PR.

framework/console/controllers/MigrateController.php
MKiselev: Was updated.

samdark (approved):
Everything's OK except wording.

MKiselev:
@samdark What exactly is wrong?
I added RU docs.

samdark:
Some wording in English. I can handle that...

samdark:
Merged with some adjustments: https://github.com/yiisoft/yii2/commit/4f89265fff54723b6a2fa624a2f89429cd482d54. Thanks!

PR #12792, opened 2016-10-20
Fix #12791

### LINKED ISSUE #12791: AttributeTypecastBehavior uses its owner before attachment
Inside a behavior `$this->owner` is `NULL` at the moment of the initialization.

[Here](https://github.com/yiisoft/yii2/blob/master/framework/behaviors/AttributeTypecastBehavior.php#L197) `get_class($this->owner)` always evaluates to `yii\behaviors\AttributeTypecastBehavior` since `$this->owner` is `NULL`.

[Here](https://github.com/yiisoft/yii2/blob/master/framework/behaviors/AttributeTypecastBehavior.php#L272) a `Call to a member function getValidators() on null` error occurs for the same reason.

rmrevin:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12791 |

### DIFF
--- a/framework/behaviors/AttributeTypecastBehavior.php
+++ b/framework/behaviors/AttributeTypecastBehavior.php
@@ -199,9 +199,9 @@ class AttributeTypecastBehavior extends Behavior
     }
 
     /**
-     * Reinitialization typecast attributes for model
+     * Reinitialization attribute types for model
      */
-    public function reinitAutoDetectedAttributeTypes()
+    private function reinitAutoDetectedAttributeTypes()
     {
         $ownerClass = get_class($this->owner);
         if (!isset(self::$autoDetectedAttributeTypes[$ownerClass])) {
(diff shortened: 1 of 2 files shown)


framework/behaviors/AttributeTypecastBehavior.php
```diff
@@ -189,17 +189,25 @@ public static function clearAutoDetectedAttributeTypes()
...
-            }
-            $this->attributeTypes = self::$autoDetectedAttributeTypes[$ownerClass];
+            $this->reinitAutoDetectedAttributeTypes();
+        }
+    }
+
+    /**
+     * Reinitialization typecast attributes for model
+     */
+    public function reinitAutoDetectedAttributeTypes()
```
cebe: I don't think there is the need to create a public method for this.

cebe (changes requested):

framework/behaviors/AttributeTypecastBehavior.php
cebe: as far as I see, you can call attach() in the test directly.

rmrevin: @cebe fixed

klimov-paul:
Resolved by commit d4df72a2a4d83ce2f4ae4fef4ca252834d6a319b

PR #12793, opened 2016-10-20
Updated Greek translation of core framework (updated for 2.0.10).

lourdas:
Please merge the request for the update of core messages in Greek.

2.0.10 was just released and I didn't have the chance to check/updated any pending messages for the Greek language. There must be a way for any translators to be notified for an imminent release, maybe 3-4 days before the actual release, so they would have the time to update their language's messages. Maybe an open issue that notifies of the imminent release where everyone could subscribe to?

Thank you.


cebe:
> There must be a way for any translators to be notified for an imminent release, maybe 3-4 days before the actual release, so they would have the time to update their language's messages. Maybe an open issue that notifies of the imminent release where everyone could subscribe to?

Thanks for noting this, we should set up a communication channel for this on the new website. I have created an issue there: https://github.com/yiisoft-contrib/yiiframework.com/issues/172

cebe:
Thank you!

PR #12800, opened 2016-10-21
Update UPGRADE.md

nadar:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

The newest composer version should be part of the upgrade guide.


samdark:
Thanks!

It works either way since there's `~` but with the current version it looks better.

PR #12802, opened 2016-10-21
Documentation: Clarified what a relation name is

winternet-studio:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? |  |
| Fixed issues |  |
| Documentation improvement | yes |

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -424,7 +424,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
     /**
      * Populates the named relation with the related records.
      * Note that this method does not check if the relation exists or not.
-     * @param string $name the relation name (case-sensitive)
+     * @param string $name the relation name (case-sensitive). If relation is defined as `public function getCustomerOrders()` in the model the relation name is `customerOrders`.
      * @param ActiveRecordInterface|array|null $records the related records to be populated into the relation.
      */
     public function populateRelation($name, $records)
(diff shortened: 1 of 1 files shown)


samdark:
It looks a bit too verbose. @yiisoft/core-developers need your opinions.

dynasource:
'If you can't explain it simply, you don't understand it well enough' (Albet Einstein)

winternet-studio:
If you can abbreviate it while keeping it clear it's great. Otherwise I would say: is it better keep it short, or is it better to keep it understandable?? If it's not understandable what's the point of having text there at all?

It could also be a link to a page that explains Yii terms, eg. "relation name". The Definitive Guide doesn't even explain what the argument is comprised of - even though it talks about the link() methods.

bizley:
I'm not sure if it's necessary here. And Guide [does explain the relation name](http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#declaring-relations).

winternet-studio:
Ah I see now, it was just far away from where I needed it! :) I spent quite a bit of time googling and trying to figure out what relation name was but failed. Maybe it was just my bad. But may I ask what the issue of making the documentation very clear is? That is what I LOVE about PHP documentation and why I think it has become so popular...

cebe:
Imo it is fine to give an example in the place where it is needed. relation name itself is not defined explicitly anywhere so adding this makes sense to me. Added with different wording: 1433d8a Thanks!

PR #12815, opened 2016-10-22
Cosmetic changes in DocBlocks

w3lifer:
In addition, it ('\*') has caused an parsing "error": http://www.yiiframework.com/doc-2.0/yii-i18n-i18n.html#$translations-detail


samdark:
Thanks!

PR #12816, opened 2016-10-24
custom db column schema

urusaich:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |

What about CustomColumnSchema? Primary for db typecasting.

Allows to do something like this:

Record.php

``` php
public function behaviors()
{
    return [
        'class' => \yii\behaviors\AttributeTypecastBehavior::class,
        'attributeTypes' => [
            'datetime' => function($value){
                if(!$value instanceof \DateTime){
                    $value = \DateTime::createFromFormat(?, $value);
                }

                return $value;
            }
        ],
        'typecastAfterFind' => true,
        'typecastAfterValidate' => false
    ];
}
```

CustomColumnSchema.php

``` php
class ColumnSchema extends \yii\db\ColumnSchema
{
    public function dbTypecast($value)
    {
        if($value instanceof \DateTime){
            return $value->format(?);
        }

        return parent::dbTypecast($value);
    }
}
```

config-db.php

``` php
'db' => [
    'schemaMap' => [
        'pgsql' => [
            'class' => \yii\db\pgsql\Schema::class,
            'columnSchemaClass' => \common\CustomColumnSchema::class
        ]
    ]
 ]
```

### DIFF
--- a/framework/db/Schema.php
+++ b/framework/db/Schema.php
@@ -76,7 +76,8 @@ abstract class Schema extends Object
     ];
 
     /**
-     * @var string
+     * @var string column schema class
+     * @since 2.0.11
      */
     public $columnSchemaClass = 'yii\db\ColumnSchema';
 
--- a/framework/db/Schema.php
+++ b/framework/db/Schema.php
@@ -76,6 +76,11 @@ abstract class Schema extends Object
     ];
 
     /**
+     * @var string
+     */
+    public $columnSchemaClass = 'yii\db\ColumnSchema';
+
+    /**
      * @var array list of ALL schema names in the database, except system schemas
      */
     private $_schemaNames;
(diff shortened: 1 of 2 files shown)


framework/db/Schema.php
```diff
@@ -76,6 +76,11 @@
     ];
 
     /**
+     * @var string
```
samdark: Need a phpdoc here w/ `@since 2.0.11`.

samdark (changes requested):
Overall looks good to me. Need a changelog.

nanodesu88:
updated

samdark (approved):

samdark:
Perfect. If you'll merge your branch with current master I'll merge pull request right away.

nanodesu88:
Solved?

samdark (approved):

samdark:
Yes. Thank you!

PR #12824, opened 2016-10-24
Enable FileMutex on Window

davidsonalencar:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | no |

Suggestion for enable Mutex for Windows. By my tests with this change this feature works on Windows.

### DIFF
--- a/framework/mutex/FileMutex.php
+++ b/framework/mutex/FileMutex.php
@@ -72,9 +72,6 @@ class FileMutex extends Mutex
      */
     public function init()
     {
-        if (DIRECTORY_SEPARATOR === '\\') {
-            throw new InvalidConfigException('FileMutex does not have MS Windows operating system support.');
-        }
         $this->mutexPath = Yii::getAlias($this->mutexPath);
         if (!is_dir($this->mutexPath)) {
             FileHelper::createDirectory($this->mutexPath, $this->dirMode, true);
@@ -137,6 +134,6 @@ class FileMutex extends Mutex
      */
     protected function getLockFilePath($name)
     {
-        return $this->mutexPath . '/' . md5($name) . '.lock';
+        return $this->mutexPath . DIRECTORY_SEPARATOR . md5($name) . '.lock';
     }
 }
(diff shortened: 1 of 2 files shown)


samdark:
Are you sure it works correctly as a mutex?

samdark:
This is why it's disabled on Windows:

```php
flock($file, LOCK_EX | LOCK_NB)
```

davidsonalencar:
Based on the [PHP documentation](http://php.net/manual/en/function.flock.php), the function `flock` works on Windows with limitation, but meets to the requirement of the `FileMutex`.

davidsonalencar:
I don't see why close this issue...

samdark:
@yiisoft/core-developers do you remember why exactly it's disabled on Windows?

klimov-paul:
`FileMutex` was created at #580 by @resurtm . May be he remebers.

klimov-paul:
> Note: flock() uses mandatory locking instead of advisory locking on Windows.

kidol:
When code was created in 2013, PHP docs were wrong on this.

https://bugs.php.net/bug.php?id=65272
>2015-02-02 18:46 UTC
>The docs are updated to reflect the fact that LOCK_NB does indeed work on Windows systems.

framework/mutex/FileMutex.php
```diff
@@ -137,6 +134,6 @@ protected function releaseLock($name)
      */
     protected function getLockFilePath($name)
     {
-        return $this->mutexPath . '/' . md5($name) . '.lock';
+        return $this->mutexPath . DIRECTORY_SEPARATOR . md5($name) . '.lock';
```
samdark: No need to change that. It works well as is on Windows.

samdark (changes requested):
Need a changelog.

framework/mutex/FileMutex.php
dynasource: Im considering '/' to be a best practice and to use DIRECTORY_SEPARATOR only when there is some matching to do.

davidsonalencar:
Added CHANGELOG change in the pull request,

samdark:
Thank you!

davidsonalencar:
I thank you!

PR #12829, opened 2016-10-25
Closed #12778: Adding common validation error (not related with particular attribute)

### LINKED ISSUE #12778: Adding common validation error (not related with particular attribute)
When writing a custom validation, we need to use `addError` method of `yii\base\Model` (if we write it in model) or `yii\validators\Validator` (if we write separate validator).

Both methods require `$attribute` parameter (attribute name) to be passed, and as we can see from the sources, we can't leave it blank:

`addError` of `yii\base\Model`:

``` php
/**
 * Adds a new error to the specified attribute.
 * @param string $attribute attribute name
 * @param string $error new error message
*/
public function addError($attribute, $error = '')
{
    $this->_errors[$attribute][] = $error;
}
```

`addError` of `yii\validators\Validator`:

``` php
/**
 * Adds an error about the specified attribute to the model object.
 * This is a helper method that performs message selection and internationalization.
 * @param string $attribute the attribute being validated
 * @param string $message the error message
 * @param array $params values for the placeholders in the error message
*/
public function addError($model, $attribute, $message, $params = [])
{
    $params['attribute'] = $model->getAttributeLabel($attribute);
    if (!isset($params['value'])) {
        $value = $model->$attribute;
        if (is_array($value)) {
            $params['value'] = 'array()';
        } elseif (is_object($value) && !method_exists($value, '__toString')) {
            $params['value'] = '(object)';
        } else {
            $params['value'] = $value;
        }
    }
    $model->addError($attribute, Yii::$app->getI18n()->format($message, $params, Yii::$app->language));
}
```

Possible options:

**1)** Select the most important relevant field and add error to it.

**2)** Select multiple important relevant fields and add the same error message to them (we can store and pass message in separate variable before passing to keep code DRY).

**3)** We can use not existing attribute name for adding error, let's say `all`, because attribute existence is not checked at that point.

``` php
class YourForm extends \yii\base\Model
{
    /**
     * @inheritdoc
     */
    public function rules()
    {
        return [
            ['name', 'yourCustomValidationMethod'],
        ];
    }        

    /**
     * @return boolean
     */
    public function yourCustomValidationMethod()
    {
        // Perform custom validation here regardless of "name" attribute value and add error when needed
        if (...) {
            $this->addError('all', 'Your error message');
        }            
    }
}
```

Note that we still need to attach validator to existing attribute(s) (otherwise exception will be thrown). We can use the most relevant attribute.

As a result, we will see error only in error summary. We can display error summary in form like that:

```
<?= $form->errorSummary($model) ?>
```

But in most cases there are always one or multiple attributes related with error so we can use option **1** or **2**. Option **3** is kind of hack but I think it's pretty good workaround for that problem.

Please also see this related [question](http://stackoverflow.com/questions/39732214/yii2-generate-common-error-in-validation-not-related-with-particular-attribut) and my [answer](http://stackoverflow.com/a/39745671/4323648) on StackOverflow, basically the text above was taken from there.

Bottom line, I think we either need:

**1)** Add support for adding common validation errors. For example Django has [support](https://docs.djangoproject.com/el/1.10/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other) for that.
**2)** If you find implementing this ambiguous or not necessary because it goes against some framework philosophy or principles, I think the solution described in option **3** proves that validation errors implementation is a bit inconsistent in terms of passing attribute.
- We can't pass not existing attribute when declaring validation rule.
- But we can't easily add error for non existing attribute.

At least we can get rid of that.

Also according docs section would be helpful.

arogachev:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12778 |

Please review and tell what's wrong / needs to be improved. When everything will be OK, I'll add both english and russian docs.

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -593,6 +593,7 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      * Returns the common errors only
      * @return array array of common error messages
      * @see getErrors()
+     * @since 2.0.11
      */
     public function getCommonErrors()
     {
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -661,7 +661,7 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
     public function addErrors(array $items)
     {
         foreach ($items as $attribute => $errors) {
-            $attribute = ($attribute === '') ? null : $attribute;
+            $attribute = ($attribute === '' || $attribute === self::COMMON_ERRORS_KEY) ? null : $attribute;
             if (is_array($errors)) {
                 foreach ($errors as $error) {
                     $this->addError($attribute, $error);
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -637,11 +637,11 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
 
     /**
      * Adds a new error to the specified attribute or common error.
-     * @param string $attribute attribute name. Leave `null` for common errors (not related with particular attribute).
+     * @param string $attribute attribute name. Pass `null` for common errors (not related with particular attribute).
      * @param string $error new error message
      * @throws InvalidParamException in case of attribute not exist
      */
-    public function addError($attribute = null, $error = '')
+    public function addError($attribute, $error = '')
     {
         if ($attribute === null) {
             $attribute = self::COMMON_ERRORS_KEY;
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -69,9 +69,15 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
      * @event Event an event raised at the end of [[validate()]]
      */
     const EVENT_AFTER_VALIDATE = 'afterValidate';
+    /**
+     * Array key for storing common validation errors (not related with particular attribute)
+     * @since 2.0.11
+     */
+    const COMMON_ERRORS_KEY = '*';
 
     /**
-     * @var array validation errors (attribute name => array of errors)
+     * @var array validation errors. Attribute related errors are indexed by according attribute names (attribute name
+     * => array of errors). Common errors are stored under the special key declared in `COMMON_ERRORS_KEY` constant.
      */
     private $_errors;
     /**
@@ -353,7 +359,7 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
         }
 
         foreach ($this->getActiveValidators() as $validator) {
-            $validator->validateAttributes($this, $attributeNames);
+            $validator->delegateValidation($this, $attributeNames);
         }
         $this->afterValidate();
 
@@ -441,7 +447,7 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
         foreach ($this->rules() as $rule) {
             if ($rule instanceof Validator) {
                 $validators->append($rule);
-            } elseif (is_array($rule) && isset($rule[0], $rule[1])) { // attributes, validator type
+            } elseif (is_array($rule) && array_key_exists(0, $rule) && isset($rule[1])) { // attributes, validator type
                 $validator = Validator::createValidator($rule[1], $this, (array) $rule[0], array_slice($rule, 2));
                 $validators->append($validator);
             } else {
@@ -525,7 +531,7 @@ class Model extends Component implements IteratorAggregate, ArrayAccess, Arrayab
 
     /**
      * Returns a value indicating whether there is any validation error.
-     * @param string|null $attribute attribute name. Use null to check all attributes.
+     * @param string|null $attribute attribute name. Use null to check all attributes and common errors.
      * @return boolean whether there is any error.
      */
     public function hasErrors($attribute = null)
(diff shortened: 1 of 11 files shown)


framework/base/Model.php
```diff
@@ -441,7 +447,7 @@ public function createValidators()
         foreach ($this->rules() as $rule) {
             if ($rule instanceof Validator) {
                 $validators->append($rule);
-            } elseif (is_array($rule) && isset($rule[0], $rule[1])) { // attributes, validator type
+            } elseif (is_array($rule) && array_key_exists(0, $rule) && isset($rule[1])) { // attributes, validator type
```
samdark: Why the change?

framework/base/Model.php
```diff
@@ -566,6 +577,30 @@ public function getErrors($attribute = null)
...
+        return $errors;
+    }
+
+    /**
+     * Returns the common errors only
+     * @return array array of common error messages
+     * @see getErrors()
+     * @since 2.0.11
+     */
+    public function getCommonErrors()
```
samdark: I'd rename this to "general" errors. What do you think?

framework/base/Model.php
```diff
@@ -601,26 +636,32 @@ public function getFirstError($attribute)
...
     /**
-     * Adds a new error to the specified attribute.
-     * @param string $attribute attribute name
+     * Adds a new error to the specified attribute or common error.
+     * @param string $attribute attribute name. Leave `null` for common errors (not related with particular attribute).
      * @param string $error new error message
+     * @throws InvalidParamException in case of attribute not exist
      */
-    public function addError($attribute, $error = '')
+    public function addError($attribute = null, $error = '')
```
samdark: It breaks backwards compatibility. It's better to introduce another method such as `addGeneralError($error)`.

framework/base/Model.php
```diff
@@ -601,26 +636,32 @@ public function getFirstError($attribute)
...
     {
+        if ($attribute === null) {
```
samdark: Here you should check for `*`.

framework/base/Model.php
```diff
@@ -601,26 +636,32 @@ public function getFirstError($attribute)
...
-     * these errors must be given in terms of an array.
+     * @param array $items a list of errors. The array keys must be attribute names. Use `null` key for adding common
+     * errors. The array values should be error messages. Pass array in case of multiple errors.
      * You may use the result of [[getErrors()]] as the value for this parameter.
      * @since 2.0.2
      */
     public function addErrors(array $items)
     {
         foreach ($items as $attribute => $errors) {
+            $attribute = ($attribute === '') ? null : $attribute;
```
samdark: Check for `*`?

framework/validators/InlineValidator.php
```diff
@@ -60,6 +60,18 @@ class InlineValidator extends Validator
     /**
      * @inheritdoc
      */
+    public function validateModel($model)
```
samdark: Is this change related?

framework/validators/Validator.php
```diff
@@ -234,6 +234,34 @@ public function init()
     }
 
     /**
+     * Delegates validation to appropriate method depending on attributes.
+     * @param \yii\base\Model $model the data model being validated.
+     * @param array|null $attributes the list of attributes to be validated.
+     * @since 2.0.11
+     */
+    public function delegateValidation($model, $attributes)
```
samdark: Is this change related?

samdark (changes requested):

framework/base/Model.php
arogachev: To support adding rules like that:

```php
[null, CommonValidator::className()]
```

or

```php
[null, `validateCommon`]
```

We can't leave `isset($rule[0])` check as is because it will always give `false` value for `null`s and as a result exception will be thrown that validation rule needs both attribute(s) and validatior to be specified.

From the other side, omitting first parameter completely is very ambiguous.

With `array_key_exists` we are making sure that it just exists, but it can be `null`.

`null` means that this rule does not directly related with specific attribute(s).

framework/base/Model.php
arogachev: Django for example uses `non_field_errors`, At first I thought to use `getNonAttributeErrors` which is more suitable for Yii2, but then decided to switch to `getCommonErrors`. `getGeneralErrors` is OK too.

framework/base/Model.php
arogachev: Sorry, but I can't see how it breaks BC. And by the way this default value is not necessary, we can remove that.

framework/validators/InlineValidator.php
arogachev: Yes, I am not sure about this implementation and name of this method. Can be `validateCommon` or something like that. Want to hear your opinions and advices.

I separated it because it doesn't fit `validateAttributes` logic and flow. Can't think of better solution with not too much changes at the moment.

framework/base/Model.php
samdark: @yiisoft/core-developers need naming suggestions.

framework/base/Model.php
samdark: Every change of public or protected method signature is BC break since you have to change all the classes overriding it when it changes.

framework/base/Model.php
samdark: Since when you're getting all errors, there's `*` key. It's natural that one would think setting `*` will add errors to that group.

framework/validators/Validator.php
arogachev: Yes, but name can be different. Do you think it's better to leave this in `validateAttributes` method (at the very beginning)?

framework/base/Model.php
arogachev: Got it. We can get rid of this default value because it's not necessary.

framework/base/Model.php
dynasource: if this method is the counterpart of geAttributeErrors(), it would be logical to name it getNonAttributeErrors()

samdark: It sounds too vague.

SilverFire:
Rejected. For the reasons, see [this comment](https://github.com/yiisoft/yii2/issues/12778#issuecomment-260192304), please.

PR #12832, opened 2016-10-25
Update BaseStringHelper.php

nadar:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

Missing docs in endsWith() method added, also provide a more user readable understanding of how case sensitive works (as its maybe not sure what part of the string is case sensitive or not...).


samdark:
Thanks!

PR #12833, opened 2016-10-25
Remove unnecessary comparsion

ghost:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

Removed unnecessary comparsion

### DIFF
--- a/framework/data/Pagination.php
+++ b/framework/data/Pagination.php
@@ -262,7 +262,7 @@ class Pagination extends Object implements Linkable
             $request = Yii::$app->getRequest();
             $params = $request instanceof Request ? $request->getQueryParams() : [];
         }
-        if ($page > 0 || $page >= 0 && $this->forcePageParam) {
+        if ($page >= 0 && $this->forcePageParam) {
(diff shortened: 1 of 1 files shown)


klimov-paul:
Original code was correct: remember that `&&` has higher priority then `||`.

dynasource:
why should there be checked for larger than 0 again? That has already been done?

Lanrik:
Ok, but may condition be rewrited as "if ($page > 0 || $page == 0 && $this->forcePageParam)" ?

dynasource:
what other options does $page have? It isnt really clear from the class (also room for improvement here) . If 0 is the only option, it can be removed completely.

cebe:
It could be changed to `$this->forcePageParam || $page > 0` to be more clear.

dynasource:
agreed. @Lanrik, can you do an update to your PR?

cebe (approved):

samdark (approved):

cebe:
Thank you!

PR #12834, opened 2016-10-25
little ActiveRecord tweak

urusaich:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |

Hey! How about some beauty and refactoring simplification? 

Now

``` php
$thread->getComments()->link($comment);
```

is similar to

``` php
$thread->link('comments', $comment);
```

### DIFF
--- a/docs/guide-pl/start-databases.md
+++ b/docs/guide-pl/start-databases.md
@@ -44,7 +44,7 @@ INSERT INTO `country` VALUES ('RU','Russia',146519759);
 INSERT INTO `country` VALUES ('US','United States',322976000);
 ```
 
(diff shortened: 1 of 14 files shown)


dynasource:
with respect to clean type hinting, we should avoid mixed parameters. It could be nice and useful in the beginning, but after a while it gets in the way. It adds confusion and more time to understand what a function does. Try to think in the reasoning behind php7 http://php.net/manual/en/migration70.new-features.php

nanodesu88:
True. It also breaks bc, my fail. Closing.

PR #12837, opened 2016-10-25
Fixes Gridview JS to respect # in filterUrl

### LINKED ISSUE #12836: Gridview FilterUrl does ignores # link in URL

Create a Gridview Filter Url that should jump to the gridview inside of the page by appending the `#` id of the gridview:

```php
	<?= GridView::widget([
		'id' => 'some-grid',
		'filterModel' => $filterModel,
		'filterUrl' => Url::current(['#' => 'some-grid']),
		'columns' => [
// ...
		],
	]); ?>
```

After filtering the URL should end with `#some-grid`.

### What do you get instead?

No `#` part of the URL.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10

cebe:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | n/a |
| Fixed issues | #12836 |

fixes #12836

### DIFF
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -118,6 +118,10 @@
 
             var pos = settings.filterUrl.indexOf('?');
             var url = pos < 0 ? settings.filterUrl : settings.filterUrl.substring(0, pos);
+            var hashPos = settings.filterUrl.indexOf('#');
+            if (hashPos >= 0) {
+                url += settings.filterUrl.substring(pos);
+            }
 
             $grid.find('form.gridview-filter-form').remove();
             var $form = $('<form/>', {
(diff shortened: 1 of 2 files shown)


samdark (approved):

arogachev:
This fix is incorrect. With these filter urls we will get the following results:

- With filter url `/posts/index#post`: expected - `/posts/index#post`, actual - `/posts/index#post/posts/index#post`.
- With filter url `/posts/index?foo=1&bar=2#post`: expected - `/posts/index#post`, actual - `/posts/index?foo=1&bar=2#post`.
- With filter url `#post`: expected - `#post`, actual - `#post#post`.
- With filter url `?foo=1&bar=2#post`: expected  - `#post`, actual - `?foo=1&bar=2#post`.

I found this during writing of the tests for `yii.gridView.js`. I've already fixed and will send PR today.

framework/assets/yii.gridView.js
```diff
@@ -118,6 +118,10 @@
 
             var pos = settings.filterUrl.indexOf('?');
             var url = pos < 0 ? settings.filterUrl : settings.filterUrl.substring(0, pos);
+            var hashPos = settings.filterUrl.indexOf('#');
+            if (hashPos >= 0) {
+                url += settings.filterUrl.substring(pos);
```
arogachev: `hashPos` must be used instead of a `pos` here. Also the anchor needs to be added only when get parameters was cut off, otherwise url already contains anchor and it will lead to duplication again.

arogachev:
Fully fixed in #13230.

PR #12839, opened 2016-10-26
Fixed #11037: yii.js and yii.validation.js should use Regexp.test instead of String.match

### LINKED ISSUE #11037: yii.validation.js should use Regexp.test instead of String.match
This is just a small nitpick.
yii.validation.js uses [String.match](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/match), but since it only cares whether there is a match or not, it should use [RegExp.test](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test).

On a related note, you don't have any tests for the js validation code, right?

arogachev:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11037 |

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -12,6 +12,7 @@ Yii Framework 2 Change Log
 - Enh #12816: Added `columnSchemaClass` option for `yii\db\Schema` which adds ability to specify custom `\yii\db\ColumnSchema` class (nanodesu88)
 - Enh #12807: Added console controller checks for `yii\console\controllers\HelpController` (schmunk42)
 - Bug #12824: Enabled usage of `yii\mutex\FileMutex` on Windows systems (davidsonalencar)
+- Enh #11037 yii.js and yii.validation.js should use Regexp.test instead of String.match (arogachev, nkovacs)
 
 2.0.10 October 20, 2016
 -----------------------
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -208,7 +208,8 @@ window.yii = (function ($) {
 
             var newForm = !$form.length;
             if (newForm) {
-                if (!action || !action.match(/(^\/|:\/\/)/)) {
+                var regexp = /(^\/|:\/\/)/;
+                if (!action || !regexp.test(action)) {
                     action = window.location.href;
                 }
                 $form = $('<form/>', {method: method, action: action});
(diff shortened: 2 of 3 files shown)


dynasource:
could you post the reason for educational purposes?

arogachev:
@dynasource The reason is shortly desribed in original issue (#11037). [String.match](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/match) returns `array` with matches in case of match and `null` otherwise. [Regexp.test](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) always returns boolean value (`true` whem match has found and `false` otherwise).

There is also [String.search](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/search) method returning an index.

> ... if you only want to know it exists, use the similar `test()` method, which returns a boolean ...

> ... for more information (but **slower execution**) use match() (similar to the regular expression exec() method). ...

In used places we only need to care about is there match or not, so using `Regexp.test` is better.

samdark (approved):

samdark:
Thanks!

SilverFire:
Simplified Regex usage 6f4c52a

It is not required to create a new variable each time when you want to `.test()`

arogachev:
@SilverFire I was thinking about replacing that way. But this is less readable in my opinion, especially combined with negation. Exclamation mark visually merges with regex pattern and it's harder to read, also it increases string length (not crucial in this particular case because regex is not so long).

I think It's better either:

- have separate regex variables like `purposeRegexp`, `anotherPurposeRegexp`;
- leave it as is, because regex is used right after assignment and needed in only this one place;
- wrap it with `new RegExp()`, but combined with negation it will produce pair of extra brackets in case of using without prior assignment - `!(new RegExp())`.String needs to be modified to - we need to remove wrapping slashes and escape remaining backslashes. But I saw `new RegExp()` is used in some other places of Yii's js.

SilverFire:
I don't see readability drawbacks when using inline regex, unless they are monstrous.

> have separate regex variables like purposeRegexp, anotherPurposeRegexp;

Good point for complex Regexps to explain what it stands for.

> wrap it with new RegExp(), but combined with negation it will produce pair of extra brackets in case of using without prior assignment - !(new RegExp()). String needs to be modified too - we need to remove wrapping slashes and escape remaining backslashes. But I saw new RegExp() is used in some other places of Yii's js.

It is mainly used to create `RegExp` object from dynamic string, stored in variable. I see no benefits for our case.

> leave it as is, because regex is used right after assignment and needed in this one place only;

That was my thought when I was writing this change.

arogachev:
> That was my thought when I was writing this change.

I mean with prior assignment to variable, but nevermind.

SilverFire:
Are you strongly disagree with my changes?

samdark:
Changes are OK. It's not really less readable.

arogachev:
@SilverFire No, just shared my opinion.

PR #12854, opened 2016-10-26
Added RangeNotSatisfiableHttpException that uses when the requested range is not satisfiable

zalatov:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

It's need to catch this error.
For example, we can't need to log this error.
Now we can do `try...catch` or add it to exclude category.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -45,7 +45,7 @@ Yii Framework 2 Change Log
 - Enh #13035: Use ArrayHelper::getValue() in SluggableBehavior::getValue() (thyseus)
 - Enh #13020: Added `disabledListItemSubTagOptions` attribute for `yii\widgets\LinkPager` in order to customize the disabled list item sub tag element (nadar)
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
-
+- Enh #12854: Added `RangeNotSatisfiableHttpException` that uses when the requested range is not satisfiable (zalatov)
 
 2.0.10 October 20, 2016
 -----------------------
--- a/framework/web/RangeNotSatisfiableHttpException.php
+++ b/framework/web/RangeNotSatisfiableHttpException.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\web;
+
+/**
+ * RangeNotSatisfiableHttpException represents an exception caused by an improper request of the end-user.
+ * This exception uses when the requested range is not satisfiable.
+ *
+ * Throwing an RangeNotSatisfiableHttpException like in the following example will result in the 416 page to be displayed.
+ *
+ * @author Zalatov Alexander <CaHbKa.Z@gmail.com>
+ *
+ * @since 2.0.11
+ */
+class RangeNotSatisfiableHttpException extends HttpException
+{
+    /**
+     * Constructor.
+     * @param string $message error message
+     * @param integer $code error code
+     * @param \Exception $previous The previous exception used for the exception chaining.
+     */
+    public function __construct($message = null, $code = 0, \Exception $previous = null)
+    {
+        parent::__construct(416, $message, $code, $previous);
+    }
+}
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -482,7 +482,7 @@ class Response extends \yii\base\Response
      *    meaning a download dialog will pop up.
      *
      * @return $this the response object itself
-     * @throws HttpException if the requested range is not satisfiable
+     * @throws RangeNotSatisfiableHttpException if the requested range is not satisfiable
      * @see sendFile() for an example implementation.
      */
     public function sendContentAsFile($content, $attachmentName, $options = [])
@@ -494,7 +494,7 @@ class Response extends \yii\base\Response
 
         if ($range === false) {
             $headers->set('Content-Range', "bytes */$contentLength");
-            throw new HttpException(416, 'Requested range not satisfiable');
+            throw new RangeNotSatisfiableHttpException();
         }
 
         list($begin, $end) = $range;
@@ -533,7 +533,7 @@ class Response extends \yii\base\Response
      *    This option is available since version 2.0.4.
      *
      * @return $this the response object itself
-     * @throws HttpException if the requested range cannot be satisfied.
+     * @throws RangeNotSatisfiableHttpException if the requested range cannot be satisfied.
      * @see sendFile() for an example implementation.
      */
     public function sendStreamAsFile($handle, $attachmentName, $options = [])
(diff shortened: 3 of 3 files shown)


framework/web/ContentRangeHttpException.php
```diff
@@ -0,0 +1,31 @@
...
+class ContentRangeHttpException extends HttpException
+{
+    /**
+     * Constructor.
+     * @param integer $status HTTP status code, default is 416.
+     * @param string $message error message
+     * @param integer $code error code
+     * @param \Exception $previous The previous exception used for the exception chaining.
+     */
+    public function __construct($status = 416, $message = 'Requested range not satisfiable', $code = 0, \Exception $previous = null)
```
cebe: code should be hardcoded as it is the case with other Http exceptions: https://github.com/yiisoft/yii2/blob/af0119a910b9d6edd81d55d0b9395f27192edb2f/framework/web/NotFoundHttpException.php#L24

cebe: message should be `null` by default, default message is set by response if not present.

framework/web/ContentRangeHttpException.php
```diff
@@ -0,0 +1,31 @@
...
+namespace yii\web;
+
+/**
+ * ContentRangeHttpException represents an exception caused by an improper request of the end-user.
+ * This exception uses when the requested range is not satisfiable.
+ *
+ * Throwing an ContentRangeHttpException like in the following example will result in the 416 page to be displayed.
+ *
+ * @author Залатов Александр <CaHbKa.Z@gmail.com>
+ */
```
cebe: `@since 2.0.11`

cebe (changes requested):

framework/CHANGELOG.md
```diff
@@ -58,6 +58,7 @@ Yii Framework 2 Change Log
 - Chg #10921: Inverts responsibility of database specific column schema builder classes (df2)
 - Chg #11071: `yii\helpers\BaseArrayHelper::isIn()` and `isTraversable()` since now throw `\yii\base\InvalidParamException` instead of `\InvalidArgumentException` (nukkumatti)
 - New #8920: Added `yii\mutex\PgsqlMutex` which implements mutex "lock" mechanism via PgSQL locks (nineinchnick, CSharpRU)
+- New: Added ContentRangeHttpException that uses when the requested range is not satisfiable (zalatov)
```
samdark: Need `#12854` here.

framework/web/ContentRangeHttpException.php
```diff
@@ -0,0 +1,31 @@
...
+class ContentRangeHttpException extends HttpException
```
samdark: How about renaming it to `InvalidContentRangeException`?

samdark (changes requested):

framework/web/ContentRangeHttpException.php
zalatov: fixed

framework/web/ContentRangeHttpException.php
zalatov: fixed

framework/CHANGELOG.md
zalatov: added

I'm add a new section - 2.0.11

framework/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ Yii Framework 2 Change Log
 - Bug #12824: Enabled usage of `yii\mutex\FileMutex` on Windows systems (davidsonalencar)
 - Enh #11037 yii.js and yii.validation.js should use Regexp.test instead of String.match (arogachev, nkovacs)
 - Bug #9796: Initialization of not existing `yii\grid\ActionColumn` default buttons (arogachev)
+- Enh #12854: Added `InvalidContentRangeHttpException` that uses when the requested range is not satisfiable (zalatov)
```
samdark: The class name here in changelog and in the actual code is different.

samdark (changes requested):

zalatov:
fixed

dynasource:
where is the unit test? Can you please add it?

framework/web/ContentRangeHttpException.php
```diff
@@ -0,0 +1,32 @@
...
+ * ContentRangeHttpException represents an exception caused by an improper request of the end-user.
+ * This exception uses when the requested range is not satisfiable.
+ *
+ * Throwing an ContentRangeHttpException like in the following example will result in the 416 page to be displayed.
+ *
+ * @author Zalatov Alexander <CaHbKa.Z@gmail.com>
+ *
+ * @since 2.0.11
+ */
+class ContentRangeHttpException extends HttpException
```
rob006: Exception name is inconsistent - all other HttpExceptions uses [status code name](https://github.com/yiisoft/yii2/blob/master/framework/web/Response.php#L163) as exception name, so to this should be named `RangeNotSatisfiableHttpException`.

rob006 (changes requested):

framework/web/ContentRangeHttpException.php
samdark: Agree. Thanks for noticing that. @zalatov could you rename it?

zalatov:
Sorry for so delay...

Fixed - renamed to RangeNotSatisfiableHttpException.

samdark (approved):

framework/CHANGELOG.md
```diff
@@ -45,7 +45,7 @@ Yii Framework 2 Change Log
 - Enh #13035: Use ArrayHelper::getValue() in SluggableBehavior::getValue() (thyseus)
 - Enh #13020: Added `disabledListItemSubTagOptions` attribute for `yii\widgets\LinkPager` in order to customize the disabled list item sub tag element (nadar)
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
-
+- Enh #12854: Added `RangeNotSatisfiableHttpException` that uses when the requested range is not satisfiable (zalatov)
```
dynasource: ->  Added `RangeNotSatisfiableHttpException` to cover 416 file request exceptions

framework/web/RangeNotSatisfiableHttpException.php
```diff
@@ -0,0 +1,32 @@
...
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\web;
+
+/**
+ * RangeNotSatisfiableHttpException represents an exception caused by an improper request of the end-user.
+ * This exception uses when the requested range is not satisfiable.
```
dynasource: Please be more clarifying about this error. Its unclear that this is about files. Wikipedia states:
'The client has asked for a portion of the file (byte serving), but the server cannot supply that portion. For example, if the client asked for a part of the file that lies beyond the end of the file.'

framework/web/Response.php
```diff
@@ -533,7 +533,7 @@ public function sendContentAsFile($content, $attachmentName, $options = [])
      *    This option is available since version 2.0.4.
      *
      * @return $this the response object itself
-     * @throws HttpException if the requested range cannot be satisfied.
+     * @throws RangeNotSatisfiableHttpException if the requested range cannot be satisfied.
```
dynasource: please make this line consistent with the other method at line 485

dynasource (changes requested):

samdark:
Merged with adjustments suggested. Test is modified to check that correct exception is thrown. Thanks for contribution.

PR #12857, opened 2016-10-27
Fixed #12822: Datetime creation from microseconds in i18n\Formatter

### LINKED ISSUE #12822: Exception when using datetime-formatting of log_time
I'm logging errors using Yii/log/DbTarget with the default migration for mysql.
Then I use yii\widgets\DetailView to display the log, and added a datetime-formatter to the attribute log_time, like this:

```
'log_time:datetime',
```

This causes errors occasionally like this...

```
exception 'Exception' with message 'DateTime::__construct(): Failed to parse time string (@1477307739.608) at position 12 (6): Unexpected character' in C:\...\vendor\yiisoft\yii2\i18n\Formatter.php:675
Stack trace:
#0 C:\...\vendor\yiisoft\yii2\i18n\Formatter.php(675): DateTime->__construct('@1477307739.608', Object(DateTimeZone))
#1 C:\...\vendor\yiisoft\yii2\i18n\Formatter.php(594): yii\i18n\Formatter->normalizeDatetimeValue('<span class="nu...')
#2 C:\...\vendor\yiisoft\yii2\i18n\Formatter.php(557): yii\i18n\Formatter->formatDateTimeValue('<span class="nu...', '<span class="st...', '<span class="st...')
#3 [internal function]: yii\i18n\Formatter->asDatetime('<span class="nu...')
 #4 C:\...\vendor\yiisoft\yii2\i18n\Formatter.php(316): call_user_func_array('[<span class="t...', '[<span class="n...')
```

The attribute log_time is defined as a double in the database (defined in default migration) but the yii/i18n/Formatter seems to fail whenever the attribute value does not have exactly 4 decimals.

   $a = new Datetime("@1477307739.6080");  // ok
   $b = new Datetime("@1477307739.608");  // fails

When logging errors, eg using Yii::error(), the timestamp will sometimes end with a zero, which then is truncated when saved as a double in the database. 

This behaviour might be a bug in PHP, but it could be corrected if the Formatter did some zeropadding to reach 4 decimals.

| Q | A |
| --- | --- |
| Yii version | 2.0.9 |
| PHP version | 5.6.27 |
| Operating system | Win10 |

h311ion:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #12822 |

Fixed datetime creation from microseconds, added tests to \i18n\Formatter::asTimestamp

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -114,16 +114,16 @@
         },
         {
             "name": "bower-asset/yii2-pjax",
-            "version": "dev-master",
+            "version": "v2.0.6",
             "source": {
                 "type": "git",
                 "url": "https://github.com/yiisoft/jquery-pjax.git",
-                "reference": "3f20897307cca046fca5323b318475ae9dac0ca0"
+                "reference": "60728da6ade5879e807a49ce59ef9a72039b8978"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/yiisoft/jquery-pjax/zipball/3f20897307cca046fca5323b318475ae9dac0ca0",
-                "reference": "3f20897307cca046fca5323b318475ae9dac0ca0",
+                "url": "https://api.github.com/repos/yiisoft/jquery-pjax/zipball/60728da6ade5879e807a49ce59ef9a72039b8978",
+                "reference": "60728da6ade5879e807a49ce59ef9a72039b8978",
                 "shasum": ""
             },
             "require": {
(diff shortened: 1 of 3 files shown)


SilverFire (changes requested):
composer.lock should be excluded from the PR.

Could you add a line to CHANGLOG.md?

SilverFire:
Good PR otherwise

samdark (approved):
Good one except what @SilverFire suggested. Please fix it.

SilverFire:
Merged a24abb09909e724f058613403061ed6c366f9e1d

SilverFire:
Thank you!

Issue #12860, opened 2016-10-27
FileMutex: unlink(...): No such file or directory


schojniak:
### What steps will reproduce the problem?

doing many parallel locks
### What is the expected result?

no error
### What do you get instead?

No such file or directory in
 `unlink($this->getLockFilePath($name));`
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.10 |
| PHP version | 5.5.9 |
| Operating system | linux Ubuntu 14.04.4 |

schojniak:
first unlink() then fclose() solved it for me

samdark:
This is extremely weird. Any idea why?

schojniak:
``fclose() `` release lock, then other process call ``acquireLock`` and ``releaseLock``; there is no file to ``unlink`` in first process.

schojniak:
you can call ``unlink`` on open file (Unix, I don't know how it works on windows)

kidol:
I tested, does work on unix, but not on windows (permission denies error). So following should be enough to fix:
```php
@unlink($this->getLockFilePath($name));
```

cebe:
@kidol wouldn't that possibly unlink a file that is locked by another process already?

schojniak:
Can you look at #12831? 

"FileMutex does not have MS Windows operating system support." was deleted to enable tests on windows.

kidol:
@cebe Yes that's what's happening already on unix (see error message in issue description). Fix is just to add `@` to suppress that error.
Little strange if script A deletes file while script B already acquired file handle and is now trying to call flock on it. But it works, you can see here: https://3v4l.org/SilHY

kidol:
Some more info regarding unlink on unix: http://stackoverflow.com/questions/23287997/why-is-unlink-successful-on-an-open-file

samdark:
Also verified that Windows gives you `Permission denied` in case you're trying to unlink locked file.

samdark:
@kidol is it possible to solve w/o using `@`?

kidol:
Hmm, looks like symfony does not remove the lock file at all: https://github.com/symfony/symfony/pull/10475#issuecomment-51464260

Problem is the following:
- Script A unlocks and calls unlink
- Script B was waiting (file handle already acquired) and will now acquire a lock on non-existing file
- Script C starts and is simply creating a new file and is able to lock it...

[...]

PR #12861, opened 2016-10-27
FileMutex: unlink(...): No such file or directory #12860

schojniak:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | no |
| Fixed issues | #12860 |

### DIFF
--- a/framework/mutex/FileMutex.php
+++ b/framework/mutex/FileMutex.php
@@ -86,12 +86,13 @@ class FileMutex extends Mutex
      */
     protected function acquireLock($name, $timeout = 0)
     {
-        $file = fopen($this->getLockFilePath($name), 'w+');
+        $fileName = $this->getLockFilePath($name);
+        $file = fopen($fileName, 'w+');
         if ($file === false) {
             return false;
         }
         if ($this->fileMode !== null) {
-            @chmod($this->getLockFilePath($name), $this->fileMode);
+            @chmod($fileName, $this->fileMode);
         }
         $waitTime = 0;
         while (!flock($file, LOCK_EX | LOCK_NB)) {
@@ -118,8 +119,8 @@ class FileMutex extends Mutex
         if (!isset($this->_files[$name]) || !flock($this->_files[$name], LOCK_UN)) {
             return false;
         } else {
-            fclose($this->_files[$name]);
             unlink($this->getLockFilePath($name));
+            fclose($this->_files[$name]);
             unset($this->_files[$name]);
 
             return true;


cebe:
looks good to me in general, is it possible to write a test case for this?

SilverFire:
I've tried to cover this fix with a test, but I failed.

```php
public function testDeleteLockConcurrency()
{
    $mutex = $this->createMutex();
    $concurrentMutex = $this->createMutex();

    $name = self::$mutexName;
    $fileName = $this->invokeMethod($mutex, 'getLockFilePath', [$name]);

    // Acquire lock
    $mutex->acquire($name);
    $this->assertFileExists($fileName);

    // Reproducing step-by-step the `release()` method and
    // make sure that new lock can not be acquired until all steps are finished
    $files = $this->getInaccessibleProperty($mutex, '_files');

    $this->assertFalse($concurrentMutex->acquire($name));

    // Step 1
    $this->assertTrue(flock($files[$name], LOCK_UN));
    $this->assertFalse($concurrentMutex->acquire($name));

    // Step 2
    $this->assertTrue(unlink($fileName));
    $this->assertFalse($concurrentMutex->acquire($name));

    // Step 3
    $this->assertTrue(fclose($files[$name]));
    $this->assertTrue($concurrentMutex->acquire($name));

    $this->assertTrue($concurrentMutex->release($name));
    $this->assertFileNotExists($fileName);
}
```

This test fails. The most interesting thing is that `$concurrentMutex->acquire($name)` will success after any of steps 1-3, the order is not important.

I do not know we can solve this probable race condition

SilverFire:
Btw, anybody knows why do we remove the file after lock is released?

SilverFire:
[NinjaMutex](https://github.com/arvenil/ninja-mutex/blob/master/src/Lock/FlockLock.php) does not remove the file, for example

kidol:
Symfony doesn't remove it either. But it's simply the ideal behavior to get rid of the log file if possible, isn't it?

Please take a look for possible solution: https://github.com/yiisoft/yii2/issues/12860#issuecomment-256901532

Alternative is to just keep the log file...

SilverFire:
Possible solution does not work

SilverFire:
> But it's simply the ideal behavior to get rid of the log file if possible, isn't it?

You've meant "lock" instead of "log", right?

Actually, not always. Sometimes we need to create a lock file outside of the web project dir to share lock with other (even non-PHP) processes. In this case I put lock somewhere in `/tmp/locks/geedyProcesses.lock`. The `/tmp/locks` directory itself is not writable for `www-data`, but the lock files could have `chmod 666` and be writable. 

So I don't think we should delete lock file

kidol:
Yes lock 😆 

>Possible solution does not work

Why?

In your use case you could set `$autoRelease` to `false` for given mutex component.

rob006:
> Btw, does anybody know why do we remove the file after lock is released?

To avoid leaving hundreds of thousands of old lock files. But this could be solved by garbage collector (could be disabled by default).

SilverFire:
> To avoid leaving hundreds of thousands of old lock files

Disagreed. Lock files are not generated randomly, because they are based on name given by the developer. It means that the count of files will be equal the maximum count of unique lock names in your code.

```php
$mutex->acquireLock('billingCalculation');
```

It does not make any sense to acquire locks with random name, because it will not end up in locking something.

kidol:
>Lock files are not generated randomly

But dynamically based on some id for example.

rob006:
Lock name could be generated. For example it could contain ID of record from database - having table with 1kk records is not a big deal for DBMS, having 1kk useless files in single directory is serious problem fore most of the filesystems. I already use mutex for generating thumbnails for images (to avoid doing the same job multiple times) - one unique ID is used only once for 1-5 seconds and never reused (because I already have a thumbnail). End up with 20k old lock files after 2 weeks.

SilverFire:
Yes, makes sense.

Atomicity is one of the most important things in MUTEX. We have the situation when lock releasing is not atomic - there is a micro-time gap between lock releasing and file deletion.

Another interesting thing is that according to the [`flock()` CHANGELOG on PHP.net](http://php.net/manual/en/function.flock.php), since PHP 5.3.2 - "The automatic unlocking when the file's resource handle is closed was removed. Unlocking now always has to be done manually". So `fopen(); flock(); fclose();` should not release lock... Need to check it once more.

> I already use mutex for generating thumbnails for images (to avoid doing the same job multiple times) - one unique ID is used only once for 1-5 seconds and never reused (because I already have a thumbnail). End up with 20k old lock files after 2 weeks.

Wow, that's cool. Good that files are empty :D

cebe:
> Good that files are empty :D

that does not help, you can not enter that directory. You can not run ls or rm on it :)
https://twitter.com/cebe_cc/status/783762785171861509

SilverFire:
Closing in favor https://github.com/yiisoft/yii2/pull/13258

PR #12866, opened 2016-10-28
simplify yii\helpers\Html methods

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

https://scrutinizer-ci.com/g/yiisoft/yii2/code-structure/master/operation/yii%5Chelpers%5CBaseHtml%3A%3AcssFile aiming to rise this metric

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1719,7 +1719,8 @@ class BaseHtml
         );
         $options['encode'] = ArrayHelper::getValue($options, 'encode', $encode);
 
-        if (null !== ($prompt = ArrayHelper::remove($tagOptions, 'prompt'))) {
+        $prompt = ArrayHelper::remove($tagOptions, 'prompt');
+        if (null !== $prompt) {
             $lines[] = static::tag(
                 'option',
                 self::applyEncode($prompt, $encode, $encodeSpaces),
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1641,25 +1641,36 @@ class BaseHtml
 
     /**
      * Generates a list of input fields.
-     * This method is mainly called by [[activeListBox()]], [[activeRadioList()]] and [[activeCheckBoxList()]].
-     * @param string $type the input type. This can be 'listBox', 'radioList', or 'checkBoxList'.
+     * This method is mainly called by [[activeListBox()]],
+     * [[activeRadioList()]] and [[activeCheckBoxList()]].
+     * @param string $type the input type. This can be 'listBox',
+     * 'radioList', or 'checkBoxList'.
      * @param Model $model the model object
-     * @param string $attribute the attribute name or expression. See [[getAttributeName()]] for the format
+     * @param string $attribute the attribute name or expression. See
+     * [[getAttributeName()]] for the format
      * about attribute expression.
      * @param array $items the data item used to generate the input fields.
-     * The array keys are the input values, and the array values are the corresponding labels.
-     * Note that the labels will NOT be HTML-encoded, while the values will.
-     * @param array $options options (name => config) for the input list. The supported special options
-     * depend on the input type specified by `$type`.
+     * The array keys are the input values, and the array values are the
+     * corresponding labels. Note that the labels will NOT be HTML-encoded,
+     * while the values will.
+     * @param array $options options (name => config) for the input list.
+     * The supported special options depend on the input type specified by
+     * `$type`.
      * @return string the generated input list
      */
     protected static function activeListInput($type, $model, $attribute, $items, $options = [])
     {
-        $name = isset($options['name']) ? $options['name'] : static::getInputName($model, $attribute);
-        $selection = isset($options['value']) ? $options['value'] : static::getAttributeValue($model, $attribute);
-        if (!array_key_exists('unselect', $options)) {
-            $options['unselect'] = '';
-        }
+        $name = isset($options['name'])
+            ? $options['name']
+            : static::getInputName($model, $attribute);
+        $selection = isset($options['value'])
+            ? $options['value']
+            : static::getAttributeValue($model, $attribute);
+        $options['unselect'] = ArrayHelper::getValue(
+            $options,
+            'unselect',
+            ''
+        );
         if (!array_key_exists('id', $options)) {
             $options['id'] = static::getInputId($model, $attribute);
         }
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -224,7 +224,7 @@ class BaseHtml
      */
     public static function cssFile($url, $options = [])
     {
-        $options = array_merge(['rel' => 'stylesheet'], $options);
+        $options['rel'] = ArrayHelper::getValue($options, 'rel', 'stylesheet');
         $noscript = ArrayHelper::remove($options, 'noscript');
         $condition = ArrayHelper::remove($options, 'condition');
         $options['href'] = Url::to($url);
@@ -1446,7 +1446,7 @@ class BaseHtml
         $options['uncheck'] = ArrayHelper::getValue(
             $options,
             'uncheck',
-            '1'
+            '0'
         );
 
         if (!array_key_exists('label', $options)) {
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1719,7 +1719,7 @@ class BaseHtml
         );
         $options['encode'] = ArrayHelper::getValue($options, 'encode', $encode);
 
-        if ($prompt = ArrayHelper::remove($tagOptions, 'prompt')) {
+        if (null !== ($prompt = ArrayHelper::remove($tagOptions, 'prompt'))) {
             $lines[] = static::tag(
                 'option',
                 self::applyEncode($prompt, $encode, $encodeSpaces),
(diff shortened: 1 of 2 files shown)


Faryshta:
https://scrutinizer-ci.com/g/yiisoft/yii2/inspections/5c534484-97ba-4fc1-b07e-6895bcea9e62

dynasource:
can you elaborate on the fact if this would be a BC break. Looking at your code it does, as arrays are a result instead of strings.

Faryshta:
@dynasource when `$options['style']` and `$options['class']` are handled by the `Html::renderTagAttributes()` method, they are automatically turned into strings. I don't think this is BC-break because the output html will be the same.

```php
Html::addCssStyle($options, 'width: 100%');
echo Html::a($url, $options);
```

Will work the same with and without this patch.

dynasource:
but then you are making an assumption that people actually use these functions. You are leaving out the opportunity that people have their own implementation in place, like further string concatenation

Faryshta:
@dynasource can you give me a code example? and if they don't use this functions then the string to array conversion never happen.

```php
$options['class'] = $options['class'] . ' row';
echo Html::a('hello', $url, $options);
``` 

will work the same as before, this patch wouldn't even be invoked.

dynasource:
Im looking primarely at the tests you have changed, for instance in this change:
https://github.com/yiisoft/yii2/pull/12866/files#diff-580fa7c1d99b2db708afd866e9d71cb9L674

I can imagine its possible that it could cause a BC for some people, have a complex architecture in which 'cssClass' operations with strings is still expected.

Faryshta:
https://scrutinizer-ci.com/g/yiisoft/yii2/inspections/4ad7a57c-9641-439d-b4b3-e695b1eee9d7

framework/helpers/BaseHtml.php
```diff
@@ -224,21 +224,17 @@ public static function script($content, $options = [])
      */
     public static function cssFile($url, $options = [])
     {
-        if (!isset($options['rel'])) {
-            $options['rel'] = 'stylesheet';
-        }
+        $options = array_merge(['rel' => 'stylesheet'], $options);
```
samdark: `array_merge` is way worse performance-wise than setting a single key.

tests/framework/helpers/HtmlTest.php
```diff
@@ -671,11 +670,11 @@ public function testRemoveCssClass()
     {
         $options = ['class' => 'test test2 test3'];
         Html::removeCssClass($options, 'test2');
-        $this->assertEquals(['class' => 'test test3'], $options);
+        $this->assertEquals(['class' => ['test', 2 => 'test3']], $options);
```
samdark: Doesn't look good to me. Return type should be compatible if you want it to be Yii 2.0.x-compatible.

samdark (changes requested):

framework/helpers/BaseHtml.php
```diff
@@ -1423,29 +1423,36 @@ public static function activeCheckbox($model, $attribute, $options = [])
...
-            $options['value'] = '1';
-        }
-        if (!array_key_exists('uncheck', $options)) {
-            $options['uncheck'] = '0';
-        }
+        $options['value'] = ArrayHelper::getValue($options, 'value', '1');
+        $options['uncheck'] = ArrayHelper::getValue(
+            $options,
+            'uncheck',
+            '1'
```
samdark: Shouldn't it be `0`?

samdark (changes requested):

tests/framework/helpers/HtmlTest.php
Faryshta: `$options['class']` can be an `array` since all the methods that deal with HTML tags call `renderTagAttributes()` which can accept arrays for this key.

```php
$options['class'] = 'hello';
Html::addCssClass($options, 'world');
echo Html::tag('div', '', $options); // will echo '<div class="hello world"></div>'
```

This behavior keeps compatible with 2.0.x using this patch.

framework/helpers/BaseHtml.php
```diff
@@ -1719,7 +1719,7 @@ public static function renderSelectOptions($selection, $items, &$tagOptions = []
         );
         $options['encode'] = ArrayHelper::getValue($options, 'encode', $encode);
 
-        if ($prompt = ArrayHelper::remove($tagOptions, 'prompt')) {
+        if (null !== ($prompt = ArrayHelper::remove($tagOptions, 'prompt'))) {
```
dynasource: 2 lines plz

tests/framework/helpers/HtmlTest.php
samdark: It is a public method which can be used directly.

framework/helpers/BaseHtml.php
dynasource: and null !== is to be avoided. Should match the other way around or ideally use isset() for readibility (if possible)

dynasource: > How would it look like in two lines?

```php
$prompt = ArrayHelper::remove($tagOptions, 'prompt');
if($prompt !== null){
}
```

cebe:
```php
$prompt = ArrayHelper::remove($tagOptions, 'prompt');
if ($prompt !== null) {
```

dynasource: hmmm our messages crossed. Sorry

tests/framework/helpers/HtmlTest.php
Faryshta: what happens with something like this?

```php
$options['class'] = ['btn' => 'btn-red'];
Html::addCssClass($options, ['btn' => 'btn-yellow']); // gets converted to string 'btn-red'
Html::addCssClass($options, ['btn' => 'btn-blue']);
echo Html::tag('div', '', $options);
// will output '<div class="btn-red btn-blue"></div>'
```

framework/helpers/BaseHtml.php
```diff
@@ -1719,7 +1719,8 @@ public static function renderSelectOptions($selection, $items, &$tagOptions = []
         );
         $options['encode'] = ArrayHelper::getValue($options, 'encode', $encode);
 
-        if (null !== ($prompt = ArrayHelper::remove($tagOptions, 'prompt'))) {
+        $prompt = ArrayHelper::remove($tagOptions, 'prompt');
+        if (null !== $prompt) {
```
dynasource: you are validating $prompt against something, not the other way around.
It should be:
```php
if($prompt !== null){    //which reads like: if 'prompt' is not null.....

}

``

**Review by dynasource — CHANGES_REQUESTED** [review]

File: framework/helpers/BaseHtml.php (hunk 5 above)
Comment (dynasource) [inline framework/helpers/BaseHtml.php]: well, have you read the 'Criticism' part on Wikipedia?:
'Critics of Yoda conditions see the lack of readability as a disadvantage that outweighs the benefits described....'

Comment (dynasource) [inline framework/helpers/BaseHtml.php]: I disagree based on two important arguments:
* readability
* consistency

Comment (dynasource) [inline framework/helpers/BaseHtml.php]: and lets not forget the third argument: we're not Yoda

File: framework/helpers/BaseHtml.php [hunk 6]
```diff
@@ -1427,29 +1423,36 @@ public static function activeCheckbox($model, $attribute, $options = [])
...
+     * This method is mainly called by [[activeCheckbox()]] and
+     * [[activeRadio()]].
+     * @param string $type the input type. This can be either `radio` or
+     * `checkbox`.
      * @param Model $model the model object
-     * @param string $attribute the attribute name or expression. See [[getAttributeName()]] for the format
-     * about attribute expression.
+     * @param string $attribute the attribute name or expression. See
+     * [[getAttributeName()]] for the format about attribute expression.
      * @param array $options the tag options in terms of name-value pairs.
```
rob006: Do we really need all these wrapping changes?

rob006 (changes requested):

tests/framework/helpers/HtmlTest.php
Faryshta: @samdark  I was checking the doc and it seems like $options['class'] can be an array or an string depending on the input. http://www.yiiframework.com/doc-2.0/yii-helpers-basehtml.html#addCssClass()-detail basically if it started as a non-empty array, then it will remain as a non-empty array after being modified.

if it started as string then it will remain as string after modifications.

framework/helpers/BaseHtml.php
Faryshta: they cause no harm and i am just applying an 80 characters stop to make it easier to read a long with the code.

dynasource: I agree. We should strive to 1 short line without line breaks.

framework/helpers/BaseHtml.php
samdark: Not Yoda we are.  Oh...  Wait...

Faryshta: The readability part is debatable. Specially when you deal with many `if` statements.

```php
if (nulll === $value) {
   // ...
}
if (0 === $value) {
   // ...
}
if (1 === $value) {
   // ...
}
if ('' === $value) {
   // ...
}
```

samdark: Sure it is. Else there would be no such style. The thing is in Yii we're primarily using normal order and want consistency.

cebe:
I do not think this is mergeable as is, a lot of changes with no good reason. Sorry to waste your effort @Faryshta but I do not see an easy way to handle such PRs. Refactoring may be useful in some places but such huge refactors are really hard to review properly so its quite impossible to merge. If you want to work on such things, either discuss it in detail before, or make multiple PRs with small changes. Thanks!

PR #12881, opened 2016-10-31
Added removeValue method to ArrayHelper

nilsburg:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

New method in ArrayHelper to remove an item by value.

```php
// $array = ['value1', 'value2'];
$key = \yii\helpers\ArrayHelper::removeValue($array, 'value2');
// result:
// $array = ['value1']

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -17,6 +17,7 @@ Yii Framework 2 Change Log
 - Enh #11037 yii.js and yii.validation.js should use Regexp.test instead of String.match (arogachev, nkovacs)
 - Bug #9796: Initialization of not existing `yii\grid\ActionColumn` default buttons (arogachev)
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
+- Enh #12881: Added `removeValue` method to `yii\helpers\BaseArrayHelper` (nilsburg)
 
 
 2.0.10 October 20, 2016
--- a/framework/helpers/BaseArrayHelper.php
+++ b/framework/helpers/BaseArrayHelper.php
@@ -247,6 +247,38 @@ class BaseArrayHelper
     }
 
     /**
+     * Removes items with matching values from the array and returns the removed items.
+     *
+     * Example,
+     *
+     * ```php
+     * $array = ['Bob' => 'Dylan', 'Michael' => 'Jackson', 'Mick' => 'Jagger', 'Janet' => 'Jackson'];
+     * $removed = \yii\helpers\ArrayHelper::removeValue($array, 'Jackson');
+     * // result:
+     * // $array = ['Bob' => 'Dylan', 'Mick' => 'Jagger'];
+     * // $removed = ['Michael' => 'Jackson', 'Janet' => 'Jackson'];
+     * ```
+     *
+     * @param array $array the array where to look the value from
+     * @param string $value the value to remove from the array
+     * @return array the items that were removed from the array
+     * @since 2.0.11
+     */
+    public static function removeValue(&$array, $value)
+    {
+        $result = [];
+        if (is_array($array)) {
+            foreach ($array as $key => $val) {
+                if ($val === $value) {
+                    $result[$key] = $val;
+                    unset($array[$key]);
+                }
+            }
+        }
+        return $result;
+    }
+
+    /**
      * Indexes and/or groups the array according to a specified key.
      * The input should be either multidimensional array or an array of objects.
      *
--- a/tests/framework/helpers/ArrayHelperTest.php
+++ b/tests/framework/helpers/ArrayHelperTest.php
@@ -142,6 +142,47 @@ class ArrayHelperTest extends TestCase
         $this->assertEquals('defaultValue', $default);
     }
 
+    public function testRemoveValueMultiple()
+    {
+        $array = [
+            'Bob' => 'Dylan',
+            'Michael' => 'Jackson',
+            'Mick' => 'Jagger',
+            'Janet' => 'Jackson'
+        ];
+
+        $removed = ArrayHelper::removeValue($array, 'Jackson');
+
+        $this->assertEquals([
+            'Bob' => 'Dylan',
+            'Mick' => 'Jagger'
+        ], $array);
+        $this->assertEquals([
+            'Michael' => 'Jackson',
+            'Janet' => 'Jackson'
+        ], $removed);
+    }
+
+    public function testRemoveValueNotExisting()
+    {
+        $array = [
+            'Bob' => 'Dylan',
+            'Michael' => 'Jackson',
+            'Mick' => 'Jagger',
+            'Janet' => 'Jackson'
+        ];
+
+        $removed = ArrayHelper::removeValue($array, 'Marley');
+
+        $this->assertEquals([
+            'Bob' => 'Dylan',
+            'Michael' => 'Jackson',
+            'Mick' => 'Jagger',
+            'Janet' => 'Jackson'
+        ], $array);
+        $this->assertEquals([], $removed);
+    }
+
     public function testMultisort()
     {
         // single key

### CODE REVIEW

**Comment by dynasource** [comment]
this is a usefull method for ArrayHelper. Can you add the unit tests to your PR to show it works correctly?

**Comment by nilsburg** [comment]
Added the unit test. I don't have much experience with unit tests so let me know if I need to change something.
Thanks!

**Comment by dynasource** [comment]
Looks good for a rookie ;)  Can you update the changelog and reset your commits to 1 commit? Thanks

**Comment by nilsburg** [comment]
Hi,
I've updated the CHANGELOG, but I've no idea how to reset the commits to 1. Would you mind guiding me? thanks!

**Comment by dynasource** [comment]
this is a way:
* soft reset your branch to the last commit from the repo (before your commits)
* commit again
* force push

**Comment by nilsburg** [comment]
Done!
Thank you for the help!

File: framework/CHANGELOG.md [hunk 1]
```diff
@@ -17,7 +17,7 @@ Yii Framework 2 Change Log
 - Enh #11037 yii.js and yii.validation.js should use Regexp.test instead of String.match (arogachev, nkovacs)
 - Bug #9796: Initialization of not existing `yii\grid\ActionColumn` default buttons (arogachev)
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
-
+- Enh: Added `removeValue` method to `yii\yii\helpers\BaseArrayHelper` (nilsburg)
```
dynasource: incorrect namespace yii\yii\helpers\BaseArrayHelper

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,32 @@ public static function remove(&$array, $key, $default = null)
     }
 
     /**
+     * Search and removes an item from an array and returns the key if found. If the value does not exists it will return the default value.
```
dynasource: please use multiple lines

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,32 @@ public static function remove(&$array, $key, $default = null)
...
+     *
+     * Example,
+     *
+     * ```php
+     * // $array = ['value1', 'value2'];
```
dynasource: remove the //

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,32 @@ public static function remove(&$array, $key, $default = null)
...
+     * $key = \yii\helpers\ArrayHelper::removeValue($array, 'value2');
+     * // result:
+     * // $array = ['value1']
```
dynasource: remove the //

dynasource (changes requested):

dynasource:
sorry, still saw some improvements

framework/helpers/BaseArrayHelper.php
nilsburg: I think that I should leave the // on this line since it represents the result. Please let me know if I am wrong.

dynasource: correct. I guess these // dont really add to the readability ;), but it is consistent with the other methods.

nilsburg:
Ok, I've just pushed the changes. Sorry for the mistakes!

dynasource:
thanks! Now lets wait for others @yiisoft/core-developers opinions

rob006:
What is the use case for `$default`?

framework/CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ Yii Framework 2 Change Log
 - Enh #11037 yii.js and yii.validation.js should use Regexp.test instead of String.match (arogachev, nkovacs)
 - Bug #9796: Initialization of not existing `yii\grid\ActionColumn` default buttons (arogachev)
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
-
+- Enh: Added `removeValue` method to `yii\helpers\BaseArrayHelper` (nilsburg)
```
samdark: Need issue number here i.e. `Enh #12881`.

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,33 @@ public static function remove(&$array, $key, $default = null)
...
+     * ```
+     *
+     * @param $array
+     * @param $value
+     * @param null $default
+     * @return mixed|null
+     */
+    public static function removeValue(&$array, $value, $default = null)
```
samdark: I think naming could be improved because `removeValue` sounds like `unset` and nothing more. How about `pullValue`?

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,33 @@ public static function remove(&$array, $key, $default = null)
...
+     * Example,
+     *
+     * ```php
+     * $array = ['value1', 'value2'];
+     * $key = \yii\helpers\ArrayHelper::removeValue($array, 'value2');
+     * // result:
+     * // $array = ['value1']
+     * ```
+     *
+     * @param $array
```
samdark: Need description.

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,33 @@ public static function remove(&$array, $key, $default = null)
...
+     * @param $value
```
samdark: Need description.

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,33 @@ public static function remove(&$array, $key, $default = null)
...
+     * @param null $default
```
samdark: Need description.

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,33 @@ public static function remove(&$array, $key, $default = null)
...
+     * @return mixed|null
+     */
```
samdark: Need `@since 2.0.11` tag here.

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,33 @@ public static function remove(&$array, $key, $default = null)
...
+    public static function removeValue(&$array, $value, $default = null)
+    {
+        if(is_array($array) && ($key = array_search($value, $array)) !== false){
```
samdark: Formatting is wrong. Should be `if (is_array`

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,33 @@ public static function remove(&$array, $key, $default = null)
...
+            unset($array[$key]);
+            return $key;
```
samdark: Shouldn't it return value instead of key?

samdark (changes requested):
`composer.lock` should not be committed.

Overall it looks OK except the fact that it returns key which is kinda useless and not value which is useful.

framework/helpers/BaseArrayHelper.php
Faryshta: whats the use of returning the value you sent to begin with?

nilsburg: I think that returning the key can be useful in case you want to store the removed key somewhere.

dynasource:
thinking more about this method, there is a problem. There can be more similar values in an array. Currently only the first one is removed. Of course all occurrences should be unset.

Regarding the return value. There are usecases thinkable. Lets say you have an array with firstnames => surnames. You want to remove a family from the array, but get all the first names. This implies we would expect the method to return all the first names.

nilsburg:
Oh, sorry, I think I closed the PR by mistake... is there a way to reopen it?

framework/helpers/BaseArrayHelper.php
nilsburg: I think `removeValue` tells you exactly what the function does, I agree that `pullValue` could work, but I think it is confusing because `pull` can be confused by just retrieving some value from the array, like `git pull` which doesn't remove anything.

dynasource:
dont you have a button reopen?

nilsburg:
Yes, but it is greyed out because there are no new commits. Do you think if I reset and commit again it would be active again?
Damm I I feel so dumb!

dynasource:
No problem. I cant either :) so I guess we need new commits.

Regarding the new commits, please take a look at https://github.com/yiisoft/yii2/pull/12881#issuecomment-257509054

nilsburg:
About similar values in the array. I think it is the user who would have to implement that in his code to remove all the similar values, but it is possible to add an option to remove all the occurrences... I am not sure which is the best option.

dynasource:
removing only the first one is rather arbitrary

nilsburg:
The thing is that if I remove all the occurrences and I want to store the removed keys, should I return an array with the removed keys?

dynasource:
correct. Your method should be altered and return an array at all times. This would be consistent, also with respect to future return type hinting of PHP7

framework/helpers/BaseArrayHelper.php
rob006: It should be named `popValue`.

nilsburg: Could it not be confused by the functionality of the `array_pop` function and expect that it will only remove the last occurrence? I've no problem renaming it, but in my opinion removeValue tells you exactly what it does.

rob006: This method is useful only because returns keys. IMO removing elements from array is not the main functionality of this method - if I want only remove something, I just use native `array_diff()`.

nilsburg: the main reason I came up with this method was because I needed to remove only one value from an array, I think this method does the job and is easier to use than `array_diff()`.

arogachev:
@dynasource Now Git has an option of squashing commits with merge so manual reset to 1 commit is not required. Check this [Github blog bost](https://github.com/blog/2141-squash-your-commits). I think it's already configured properly because I saw that @samdark merded some PRs that way.

dynasource:
@arogachev, nice to know. Thanks for sharing!

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,40 @@ public static function remove(&$array, $key, $default = null)
...
+     * $keys = \yii\helpers\ArrayHelper::removeValue($array, 'value2');
+     * // result:
+     * // $array = ['value1', 'value3']
+     * // $keys = [1,3]
+     * ```
+     *
+     * @param array $array the array where to look the value from
+     * @param string $value the value to remove from the array
+     * @param mixed $default the value to return in case the value to remove was not found
+     * @return mixed|null the key of the element if found, default value otherwise
```
samdark: But why returning keys? What's the use case for that?

dynasource: plz see https://github.com/yiisoft/yii2/pull/12881#issuecomment-257509054

SilverFire (changes requested):
Code style should be adjusted

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,40 @@ public static function remove(&$array, $key, $default = null)
...
+     * @since 2.0.11
+     */
+    public static function removeValue(&$array, $value, $default = null)
+    {
+        $result = [];
+        if (is_array($array)) {
```
SilverFire: Will not work for ArrayAccess objects

SilverFire: Does it make any sense to continue execution when the `$array` is not array indeed?

SilverFire (changes requested):

framework/helpers/BaseArrayHelper.php
samdark: That sounds very specific to me...

rob006: We could return values indexed by keys.

tests/framework/helpers/ArrayHelperTest.php
```diff
@@ -142,6 +142,18 @@ public function testRemove()
...
+    public function testRemoveValue()
+    {
+        $array = ['value1', 'value2', 'value3', 'value2'];
+        $keys = ArrayHelper::removeValue($array, 'value2');
+
+        $this->assertEquals($keys, [1,3]);
+        $this->assertEquals($array, [0=>'value1', 2=>'value3']);
+
+        $default = ArrayHelper::removeValue($array, 'nonExisting', 'defaultValue');
+        $this->assertEquals('defaultValue', $default);
```
SilverFire: Wrong cases should be tested as well

SilverFire (changes requested):

nilsburg:
Hi,
Following the suggestions I've made some changes in the method.
I removed the $default parameter, I couldn't find really a good use for it.
It now returns the removed items instead of just the keys.
I've also modified the example following @dynasource example here: https://github.com/yiisoft/yii2/pull/12881#issuecomment-257509054
About the code styling, I applied the PSR1/PSR2 code styling with PhpStorm, hope it's ok now.

So let me know if you can think of more improvements and thanks for the suggestions.

samdark (approved):

samdark:
I think now it's ready. @dynasource, @SilverFire ?

framework/helpers/BaseArrayHelper.php
```diff
@@ -247,6 +247,38 @@ public static function remove(&$array, $key, $default = null)
...
+     * Removes items with matching values from the array and returns the removed items.
+     *
+     * Example,
+     *
+     * ```php
+     * $array = ['Bob'=>'Dylan', 'Michael'=>'Jackson', 'Mick'=>'Jagger', 'Janet'=>'Jackson'];
+     * $removed = \yii\helpers\ArrayHelper::removeValue($array, 'Jackson');
+     * // result:
+     * // $array = ['Bob'=>'Dylan', 'Mick'=>'Jagger'];
+     * // $removed = ['Michael'=>'Jackson', 'Janet'=>'Jackson'];
```
samdark: Yes. Indeed.

nilsburg: Done

tests/framework/helpers/ArrayHelperTest.php
```diff
@@ -142,6 +142,31 @@ public function testRemove()
         $this->assertEquals('defaultValue', $default);
     }
 
+    public function testRemoveValue()
```
samdark: The test method has two cases. Could you split it into test method and data-provider?

samdark (changes requested):

tests/framework/helpers/ArrayHelperTest.php
dynasource:
```php
    public function testRemoveValueMultiple()
    {
        $array = [
            'Bob' => 'Dylan',
            'Michael' => 'Jackson',
            'Mick' => 'Jagger',
            'Janet' => 'Jackson'
        ];

        $removed = ArrayHelper::removeValue($array, 'Jackson');

        $this->assertEquals([
            'Bob' => 'Dylan',
            'Mick' => 'Jagger'
        ], $array);
        $this->assertEquals([
            'Michael' => 'Jackson',
            'Janet' => 'Jackson'
        ], $removed);
    }

    public function testRemoveValueNotExisting()
    {
        $array = [
            'Bob' => 'Dylan',
            'Michael' => 'Jackson',
            'Mick' => 'Jagger',
            'Janet' => 'Jackson'
        ];

        $removed = ArrayHelper::removeValue($array, 'Marley');

        $this->assertEquals([
            'Bob' => 'Dylan',
            'Michael' => 'Jackson',
            'Mick' => 'Jagger',
            'Janet' => 'Jackson'
        ], $array);
        $this->assertEquals([], $removed);
    }
```

dynasource: you can use the code above. Please note the following improvements:
* tests should be clear in 1 second. If you still have to think when you read it, its not ready yet
* 1 usecase per method. This enables you to create a good name for the method  
* try to follow the following 3 step template and split them with a blank line. It will be much easier to distinct.

1. prepare the data
2. execute the code
3. do the assertions

* dont put arrays in 1 line. Split them over multiple lines (otherwise you still dont see what has changed)
* dont reverse 'actual' and 'expected' in assertions. First you have the 'expected' and then 'actual':

Example:
```php public static function assertEquals($expected, $actual) ```

dynasource (changes requested):
see comments about tests

tests/framework/helpers/ArrayHelperTest.php
nilsburg: @dynasource I've added your code for the tests. Thank you very much for the help and explanation.

samdark (approved):

dynasource (approved):

tests/framework/helpers/ArrayHelperTest.php
dynasource: thank you for your perseverance

samdark:
Merged. Thank you!

dynasource:
good work

PR #12886, opened 2016-11-01
doc: GridView::filterModel

### LINKED ISSUE #9220: Suggestion: add to GridView::filterModel doc note that the filter control will not be displayed if an attribute of the model is not active
I spent two hours in reasearch why my filter controls are not displayed on the grid. The reason is filter model's attributes are not active (not declared in screnarios()/rules()).

The filter cell rendering code has line (file vendor/yiisoft/yii2/grid/DataColumn.php, function renderFilterCellContent):

`
if ($this->filter !== false && $model instanceof Model && $this->attribute !== null && $model->isAttributeActive($this->attribute))
`

I thought it was a bug because expecting behavior is just not populate model attributes with user data, but this line prevents rendering of control if attribute is not active. 

I am sure that you need to write a note about this case in GridView documentation.

githubjeka:
Fixes #9220

### DIFF
--- a/docs/guide/output-data-widgets.md
+++ b/docs/guide/output-data-widgets.md
@@ -345,8 +345,8 @@ For filtering data, the GridView needs a [model](structure-models.md) that repre
 usually taken from the filter fields in the GridView table.
 A common practice when using [active records](db-active-record.md) is to create a search Model class
 that provides needed functionality (it can be generated for you by [Gii](start-gii.md)). This class defines the validation
-rules to show filter controls on the GridView table and to provides a `search()` method that will return the data 
-provider with an adjusted query that respects the search criteria.
+rules to show filter controls on the GridView table and to provide a `search()` method that will return the data 
+provider with an adjusted query that processes the search criteria.
 
 To add the search capability for the `Post` model, we can create a `PostSearch` model like the following example:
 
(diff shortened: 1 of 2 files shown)


docs/guide/output-data-widgets.md
```diff
@@ -345,8 +345,8 @@ For filtering data, the GridView needs a [model](structure-models.md) that repre
 usually taken from the filter fields in the GridView table.
 A common practice when using [active records](db-active-record.md) is to create a search Model class
 that provides needed functionality (it can be generated for you by [Gii](start-gii.md)). This class defines the validation
-rules for the search and provides a `search()` method that will return the data provider with an
-adjusted query that respects the search criteria.
+rules to show filter controls on the GridView table and to provides a `search()` method that will return the data 
+provider with an adjusted query that respects the search criteria.
```
dynasource: an adjusted query -> a adjusted query

arogachev: Usage of "an" in phrase "an adjusted query" is correct. Not sure if it's needed at all, we can write "with adjusted query" in this sentence.

dynasource: the subject is 'query'. Adjusted is an 'adjective' and therefore not leading.

arogachev: @dynasource It's ambiguous actually. But seems like "an adjusted query" is more acceptable, check [this link](http://english.stackexchange.com/questions/152/when-should-i-use-a-vs-an): 

> He ate a green apple.

is correct.

> In any case, the rule is that you use "an" if the next word begins with a vowel sound

Think about it from pronouncing point of view, it souds very awkward.

dynasource: I totally agree. Seem that Im following the wrong grammar rule. How did that happen.

samdark:
Thanks!

PR #12889, opened 2016-11-01
Unit tests created for ActiveQuery

dynasource:
basic unit tests for ActiveQuery are created:
- all protected & public methods are covered
- method return types are tested
- some functions are already full tested, others have the @todo tag


samdark (approved):

cebe:
fixed `@group` annotations: 24809c5 this should also solve the OCI test issue.

Issue #12775, opened 2016-10-19
Profile cache level


vitalyzhakov:
### What steps will reproduce the problem?
Process connect to several cache systems. For performance purposes programmers need to know how effectively use cache.

### What is the expected result?
In debug panel: performance, time to do query.

### What do you get instead?
Nothing in debug panel if I use memcache.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.9
| PHP version      | 7.0.12
| Operating system | Debian 8, Docker

vitalyzhakov:
@samdark , why moved to debug? I think, Yii::beginProfile must be in core files (for example, in
https://github.com/yiisoft/yii2/blob/master/framework/caching/Cache.php) such as https://github.com/yiisoft/yii2/blob/master/framework/db/Command.php#L842

samdark:
Probably you're right...

dynasource:
this is a useful indeed.

The question is how to implement this in an optimal & flexible way.
Perhabs we want this to be covered with Events inside the Cache.php. Then developers have more flexibility.

Looking at your current PR https://github.com/yiisoft/yii2/issues/12891, I see that the logging is also triggered on production. Performancewise, this is not preferable. There should at least be a check in place for YII_DEBUG.

If you'd ask me, I would go for the triggering of an Event. Then you can decide in the applications bootstrap.php which Events you want to trigger for your debug-toolbar.

vitalyzhakov:
Can't understand your idea, please, give me an example of using events.
Checking YII_DEBUG I will add, thank you.

dynasource:
of course:

in bootstrap.php

```php
use yii\base\Event;
use yii\base\View;

Event::on(View::class,View::EVENT_BEFORE_RENDER,function(){
    Yii::beginProfile(Yii::$app->params['RENDER'] ?? Yii::$app->params['RENDER'] = 0,'View::render()');
    Yii::$app->params['RENDER']++;
});
Event::on(View::class,View::EVENT_AFTER_RENDER,function(){
    Yii::endProfile(Yii::$app->params['RENDER']-1);
});
```

samdark:
Triggering events should have even more performance impact than calling profiling methods so I'd keep it as profiling ones only and in debug mode only.

dynasource:
I thought about that, but it can be questioned:
Firstly, is triggering and empty array of events really a performance killer?
Secondly, the total number of cache hits on any page does usually not exceed 5 hits.

[...]

PR #12891, opened 2016-11-01
Profile cache time for performance purpose

vitalyzhakov:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12775

### DIFF
--- a/framework/caching/Cache.php
+++ b/framework/caching/Cache.php
@@ -101,8 +101,11 @@ abstract class Cache extends Component implements \ArrayAccess
      */
     public function get($key)
     {
-        $key = $this->buildKey($key);
-        $value = $this->getValue($key);
+        \Yii::beginProfile($key);
+        $buildedKey = $this->buildKey($key);
+        $value = $this->getValue($buildedKey);
+        \Yii::endProfile($key);
+
         if ($value === false || $this->serializer === false) {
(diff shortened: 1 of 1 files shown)


vitalyzhakov:
```* @param mixed $key a key identifying the cached value. This can be a simple string or
     * a complex data structure consisting of factors representing the key.```
@samdark , this function really using mixed argument or strings only?

**Comment by samdark** [comment]
Yes, mixed is correct. See `buildKey()`.

File: framework/caching/Cache.php [hunk 1]
```diff
@@ -101,8 +101,11 @@ public function buildKey($key)
      */
     public function get($key)
     {
-        $key = $this->buildKey($key);
-        $value = $this->getValue($key);
+        \Yii::beginProfile($key);
```
samdark: `__METHOD__` should be used as a second argument.

samdark: Need to check how these calls affect production performance. If much, add a check for `YII_DEBUG`.

samdark (changes requested):

vitalyzhakov:
I have changed code and create new PR #12911

PR #12892, opened 2016-11-01
fixes #12015, request for making createModels() extendable

### LINKED ISSUE #12015: Feature Request: private to protected scope changes in ActiveQueryTrait
I know this might be a bit strange to ask, but I would like if it is possible to change private scope of `createModels` in `ActiveQueryTrait`. 

Generally my opinion is that one should never use private scope inside frameworks since you never know which part of the framework somebody might use and modify by extending that class and overriding that function. It would be best to use only protected and public.

Here is a reason why I would find this change to be really good:
We wanted to make a functionality which extends ActiveRecord query and allows you to do things like:

``` php
ActiveRecord::find()->configure([
   'scenario' => 'some-scenario',
   'attribute1' => '...'
   // ...
])->andWhere()->orWhere()->all();
```

Which would basically return all models already set with attributes and scenarios of my choosing. You could argue that I could do this in `afterFind` of a specific model but this leaves way more customization ability because sometimes I want to avoid doing things (like loading additional data) in `afterFind` when a specific scenario is set and there is no way to do this in current Yii 2. If I could extend `createModels` function. Here is how it would look like in extended ActiveQuery:

``` php
class ActiveQueryExtension extends yii\db\ActiveQuery {
    public $modelConfig = [];

    public function configure($data)
    {
        $this->_modelConfig = $data;
        return $this;
    }

    protected function createModels($rows) 
    {
        $models = parent::createModels($rows);

        foreach($models as $model) {
              \Yii::configure($model, $this->modelConfig);
        }   

        return $models;
    }
}
```

dynasource:
fixes #12015, request for making createModels() extendable

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -18,9 +18,10 @@ Yii Framework 2 Change Log
 - Enh #11037 yii.js and yii.validation.js should use Regexp.test instead of String.match (arogachev, nkovacs)
 - Bug #9796: Initialization of not existing `yii\grid\ActionColumn` default buttons (arogachev)
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
--- Enh #12748: Added Migration tool automatic generation reference column for foreignKey (MKiselev)
+- Enh #12748: Added Migration tool automatic generation reference column for foreignKey (MKiselev)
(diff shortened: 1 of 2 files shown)


samdark (approved):

SilverFire (approved):
@since should be added

samdark:
Merged. Thanks!

PR #12893, opened 2016-11-01
Change the name of method getQueryTableName and remove its $query argument

### LINKED ISSUE #12878: Expose ActiveQuery->getQueryTableName()
Can this method be exposed so it can be used in extended `ActiveQuery` classes.
Would allow you to easily use table aliases.

```php
class ExtraQuery extends ActiveQuery
{
    //...
    public function extraColumn()
    {
        list(, $tableName) = $this->getQueryTableName($this);
        $this->addSelect("$tableName.column");
        return $this;
    }
    //...
}
```

dynasource:
see the descriptions of the different commits.
I also fixed consistency for some other unit tests

### DIFF
--- a/tests/framework/db/ActiveQueryTest.php
+++ b/tests/framework/db/ActiveQueryTest.php
@@ -135,7 +135,7 @@ abstract class ActiveQueryTest extends DatabaseTestCase
     public function testGetQueryTableName_from_not_set()
     {
         $query = new ActiveQuery(Customer::className());
-        $result = $query->getQueryTableName();
+        $result = $this->invokeMethod($query,'getTableNameAndAlias');
         $this->assertEquals(['customer','customer'], $result);
     }
 
@@ -143,7 +143,7 @@ abstract class ActiveQueryTest extends DatabaseTestCase
     {
         $options = ['from' => ['alias'=>'customer']];
         $query = new ActiveQuery(Customer::className(),$options);
-        $result = $query->getQueryTableName();
+        $result = $this->invokeMethod($query,'getTableNameAndAlias');
         $this->assertEquals(['customer','alias'], $result);
     }
 
--- a/tests/framework/db/ActiveQueryTest.php
+++ b/tests/framework/db/ActiveQueryTest.php
@@ -118,7 +118,7 @@ abstract class ActiveQueryTest extends DatabaseTestCase
     }
 
     /**
-     * @todo: tests for internal logic of joinWith()
+     * @todo: tests for internal logic of innerJoinWith()
      */
     public function testInnerJoinWith()
     {
@@ -127,6 +127,24 @@ abstract class ActiveQueryTest extends DatabaseTestCase
         $this->assertEquals([
             [['profile'], true, 'INNER JOIN']
         ], $result->joinWith);
+    }
+
+    /**
+     * @todo: tests for the regex inside getQueryTableName
+     */
+    public function testGetQueryTableName_from_not_set()
+    {
+        $query = new ActiveQuery(Customer::className());
+        $result = $query->getQueryTableName();
+        $this->assertEquals(['customer','customer'], $result);
+    }
+
+    public function testGetQueryTableName_from_set()
+    {
+        $options = ['from' => ['alias'=>'customer']];
+        $query = new ActiveQuery(Customer::className(),$options);
+        $result = $query->getQueryTableName();
+        $this->assertEquals(['customer','alias'], $result);
     }
 
     public function testOnCondition()
@@ -139,7 +157,7 @@ abstract class ActiveQueryTest extends DatabaseTestCase
         $this->assertEquals($params, $result->params);
     }
 
-    public function testAndOnCondition_OnIsNull()
+    public function testAndOnCondition_on_not_set()
     {
         $query = new ActiveQuery(Customer::className());
         $on = ['active' => true];
@@ -149,7 +167,7 @@ abstract class ActiveQueryTest extends DatabaseTestCase
         $this->assertEquals($params, $result->params);
     }
 
-    public function testAndOnCondition_OnIsNotNull()
+    public function testAndOnCondition_on_set()
     {
         $onOld = ['active' => true];
         $query = new ActiveQuery(Customer::className());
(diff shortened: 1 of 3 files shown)


samdark (approved):

tests/framework/db/ActiveQueryTest.php
```diff
@@ -129,6 +129,24 @@ public function testInnerJoinWith()
         ], $result->joinWith);
     }
 
+    /**
+     * @todo: tests for the regex inside getQueryTableName
+     */
+    public function testGetQueryTableName_from_not_set()
```
klimov-paul: Why methods are named not in 'camelCase' ?

klimov-paul (changes requested):

tests/framework/db/ActiveQueryTest.php
samdark: Yes, should be renamed.

klimov-paul:
Actually the entire change does not feels right. In the scope of `ActiveQuery` method `getQueryTableName()` seems to be ambiguous: the table name, which query is acciciated to is already covered by `Query::from`.
Originally method `getQueryTableName()` was an internal helper for the query bulding process and thus its presence was OK. But this PR changes it purpose creating a twisted public interface.

If you wish `ActiveQuery` to be able to return a related table name you should convert `from` into a virtual property with getter, which can prefill it by `ActiveRecord::tableName()` if its value is empty.

tests/framework/db/ActiveQueryTest.php
dynasource: for the simple reason that multiple tests for the same method can be easily distinguished

dynasource: an alternative could be testGetQueryTableName_fromNotSet(), its a matter of taste and the depth in which you are testing. If you have 20 different test methods with similarities, the underscores help significantly with an easy comparison

klimov-paul: Does not sound convincing to me

dynasource:
I totally disagree. The current implementation shows zero reason for having $query as an argument.

klimov-paul:
???
So do you think the argument should be removed entirely then?
What about these places:

https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveQuery.php#L606
https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveQuery.php#L607

samdark:
Hmm... what @klimov-paul says makes sense. At least problem part of it.

dynasource:
I dont see a $this inside getQueryTableName(), then it makes no sense to have $query as an argument. Its confusing.

samdark:
@dynasource see lines @klimov-paul linked to.

samdark (changes requested):
Seems it won't work this way.

samdark (approved):
No, that's OK :)

samdark:
Re-checked. Changes looks OK to me.

tests/framework/db/ActiveQueryTest.php
dynasource: have you every created 20 test methods to test 1 PHP function?

samdark:
@klimov-paul the places you've referred to are covered.

klimov-paul:
I will not accept this change unless @cebe agreed on it.

samdark:
#12878

tests/framework/db/ActiveQueryTest.php
dynasource: for normal framework code, yes. camelCase will do. 
But not if you are testing:
* 1 method:
* on multiple criteria
* in multiple test methods (to keep the smallest contained environment + to have an instant understanding if a test fails)

dynasource: https://books.google.nl/books?id=K5dYi2HfAOYC&pg=PA82&lpg=PA82&dq=phpunit+methods+underscore+distinguish&source=bl&ots=L_aiHseLUp&sig=qk_bWtDkN7yk17oDjgeZC_hZ6KU&hl=nl&sa=X&ved=0ahUKEwiModPdjorQAhXKCywKHReLAwoQ6AEIUDAG#v=onepage&q=phpunit%20methods%20underscore%20distinguish&f=false

cebe (changes requested):
we may turn it into a protected method as it was discussed in #12878 but it should not be public. Public method will add something to the interface that does not belong here. All other methods are for constucting/modifying the Query object. Only this one would be a getter, this creates inconsistency in the interface.

dynasource:
the function name should also be altered. The current PR still shows the old name 'getQueryTableName' and without $query as an argument, it should at least be named getTableName(). But, regarding all of our conversations, this name is also not preferable. @klimov-paul stated that this method is meant for internal processing. It should get a name that clarifies that.

samdark:
Agree.

cebe:
I have reviewed the method now in more detail and I think it should not be added to the public interface, not even protected.

- Its original implementation was to determine table names from `from` part in certain circumstances. This works as expected for the code where it is used now and appropriate checks are made, e.g. in `alias()` ,method. It may not work as expected, when `from` contains multiple tables.
- If we want to provide a new method for retrieving an alias or table name, we get to the same discussion we already have/had in #9326 and #7263.
- there should be a `getAlias()` and/or `getTableName()` method whos implementation may be based on this internal method, but that should be discussed in the issues and addressed separately from the refactoring made here.

additionally to `from`, which is already there

dynasource:
I agree. Do you have a suggestion for the private method to be newly called?

samdark:
`getTableNameAndAlias()`?

dynasource:
'getTableNameAndAlias' is the most logical with respect to the objective of the method. But we should also take the 'temporary processing' aspect into account. Couldnt we add add the '@internal' tag to this method? Based on the PHPdoc specs https://phpdoc.org/docs/latest/references/phpdoc/tags/internal.html, this tag:
* The @internal tag can be used as counterpart of the @api tag, indicating that the associated Structural Elements are used purely for the internal workings of this piece of software.

This would saves us from confusion in the future

cebe:
a private method is "internal" anyway so there is no need for such a label.

dynasource:
of course, but it helps emphasizing that it should only be used internally and never be made public. It would have prevented this whole discussion and the issue at https://github.com/yiisoft/yii2/issues/12878

samdark:
Name should be changed in the rest of the code, it seems...

dynasource:
good to see my added tests in action ;)

cebe (approved):

cebe:
Thank you!

PR #12895, opened 2016-11-02
Use custom error action regardless of response format

erikverheij:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | 

`Yii\web\ErrorHandler` has the ability to set a custom action for rendering errors. Now it uses that action regardless of the response format. Previously the custom action was not triggered when the response format was something other than `FORMAT_HTML` which is unexpected behavior.


SilverFire (approved):
Except minor changes in CHANGELOG

cebe:
this changes existing behavior, so it is a BC break.

erikverheij:
Thank you for reviewing this PR. Is there any action from my side required?

framework/CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.11 under development
 ------------------------
-
+- Bug: Use custom error action when provided, regardless of the response format.
```
samdark: Should go to 2.1 changelog.

samdark:
I am OK changing that in 2.1. If user set error explicitly it is logical to do exactly what was configured.

SilverFire:
Btw it will be good to have unit-tests for ErrorHandler.

samdark:
Merged. Thank you!

PR #12898, opened 2016-11-03
fix #4790

Faryshta:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #4790

### DIFF
--- a/framework/data/Sort.php
+++ b/framework/data/Sort.php
@@ -12,6 +12,7 @@ use yii\base\InvalidConfigException;
 use yii\base\Object;
 use yii\helpers\Html;
 use yii\helpers\Inflector;
+use yii\helpers\ArrayHelper;
 use yii\web\Request;
 
 /**
@@ -133,6 +134,12 @@ class Sort extends Object
      * @see params
      */
     public $sortParam = 'sort';
+    /**
+     * @var string the name of the parameter that specifies the page number in
+     * the url. This will be used to unset said value on the sort link.
+     * @since 2.0.11
+     */
+    public $pageParam = 'page';
     /**
      * @var array the order that should be used when the current request does not specify any order.
      * The array keys are attribute names and the array values are the corresponding sort directions. For example,
@@ -301,9 +308,7 @@ class Sort extends Object
      */
     public function getAttributeOrder($attribute)
     {
-        $orders = $this->getAttributeOrders();
-
-        return isset($orders[$attribute]) ? $orders[$attribute] : null;
+        return ArrayHelper::getValue($this->getAttributeOrders(), $attribute);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


framework/data/Sort.php
```diff
@@ -134,6 +135,12 @@ class Sort extends Object
      */
     public $sortParam = 'sort';
     /**
+     * @var string the name of the parameter that specifies the page number in
+     * the url. This will be used to unset said value on the sort link.
+     * @since 2.0.11
+     */
+    public $pageParam = 'page';
```
cebe: could be `null` by default, and if set it will be used. this way we can introduce it backwards compatible.

framework/data/Sort.php
```diff
@@ -301,9 +308,7 @@ public function setAttributeOrders($attributeOrders, $validate = true)
      */
     public function getAttributeOrder($attribute)
     {
-        $orders = $this->getAttributeOrders();
-
-        return isset($orders[$attribute]) ? $orders[$attribute] : null;
+        return ArrayHelper::getValue($this->getAttributeOrders(), $attribute);
```
cebe: this adds too much overhead, it does not make sense to replace a simple isset() with the complex implementation that `ArrayHelper::getValue()` has.

framework/data/Sort.php
```diff
@@ -322,29 +327,22 @@ public function getAttributeOrder($attribute)
...
-            $class = $direction === SORT_DESC ? 'desc' : 'asc';
-            if (isset($options['class'])) {
-                $options['class'] .= ' ' . $class;
-            } else {
-                $options['class'] = $class;
-            }
+            Html::addCssClass(
+                $options,
+                $direction === SORT_DESC ? 'desc' : 'asc'
+            );
```
cebe: :+1:

framework/data/Sort.php
```diff
@@ -322,29 +327,22 @@ public function getAttributeOrder($attribute)
...
-        return Html::a($label, $url, $options);
+        return Html::a(
+            // Getting label from options, attribute configuration or generating
+            // the label from attribute name
+            ArrayHelper::remove($options, 'label')
+                ?: ArrayHelper::getValue($this->attributes[$attribute], 'label')
+                ?: Inflector::camel2words($attribute),
+            $this->createUrl($attribute),
+            $options
+        );
```
cebe: I'd prefer original code. it is more readable than nested ternary operators.

framework/data/Sort.php
```diff
@@ -394,14 +391,16 @@ public function createSortParam($attribute)
             $direction = $directions[$attribute] === SORT_DESC ? SORT_ASC : SORT_DESC;
             unset($directions[$attribute]);
         } else {
-            $direction = isset($definition['default']) ? $definition['default'] : SORT_ASC;
+            $direction = ArrayHelper::getValue(
+                $definition,
+                'default',
+                SORT_ASC
+            );
```
cebe: same as above, `ArrayHelper::getValue` is overkill.

framework/data/Sort.php
cebe: It also can break code that uses related attributes, `author.name` will evaluate differently in new and old code.

cebe (changes requested):
In general, please address only one issue per pull request. It is hard to review general refactoring together with adding a feature. Small changes are okay, but you did a lot of unrelated things here that makes it hard to accept the PR.

framework/data/Sort.php
cebe: yes, if it is added with default value null, this can go into 2.0.x

Faryshta: on #4790 this is considered a bug so yes this changes previous behavior because it fixes a bu

PR #12901, opened 2016-11-03
Extract default help header of console commands into a separate method.

ww-stas:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

I would like to overite  default help header but copy of all source code of getDefaultHelp method is wrong way.


framework/console/controllers/HelpController.php
```diff
@@ -437,4 +437,13 @@ protected function getScriptName()
     {
         return basename(Yii::$app->request->scriptFile);
     }
+
+    /**
+     * Display a default help header.
+     * @return string default help header.
```
samdark: Need `@since 2.0.11` here.

framework/console/controllers/HelpController.php
```diff
@@ -437,4 +437,13 @@ protected function getScriptName()
     {
         return basename(Yii::$app->request->scriptFile);
     }
+
+    /**
+     * Display a default help header.
```
samdark: It's not displaying, it's returning.

samdark (changes requested):
Makes sense overall but fixes I've mentioned should be done. Additionally we need a CHANGELOG update.

framework/console/controllers/HelpController.php
diezztsk: Fixed.

dynasource (approved):

samdark (approved):

samdark:
Merged. Thank you!

PR #12906, opened 2016-11-03
fixes #12905

### LINKED ISSUE #12905: unnecessary cache misses
advanced yii project (multiple apps with different routes) uses same key that is hardcoded into `yii\\web\\UrlManager`

```
    protected $cacheKey = __CLASS__;
```

this would case hash to always be a miss, each hit will do both `get` and `set`

```
        if ($this->cache instanceof Cache) {
            $cacheKey = $this->cacheKey;
                $this->rules = $data[0];
            } else {
```

the solution is to suffix `cacheKey` with  let's say basepath or app name ...etc.

muayyad-alsadi:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12905

### DIFF
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -183,12 +183,12 @@ class UrlManager extends Component
             $this->rules = $this->buildRules($this->rules);
         }
     }
-    
+
     /**
      * return class name suffixed with 
      **/
-    protected function $getCacheKey() {
-        return __CLASS__.'.'. Yii::getAlias('@app');
+    protected function getCacheKey() {
+        return __CLASS__.'.'.Yii::getAlias('@app');
     }
 
 
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -143,11 +143,6 @@ class UrlManager extends Component
      */
     public $normalizer = false;
 
-    /**
-     * @var string the cache key for cached rules
-     * @since 2.0.8
-     */
-    protected $cacheKey = __CLASS__;
 
     private $_baseUrl;
     private $_scriptUrl;
@@ -188,6 +183,14 @@ class UrlManager extends Component
             $this->rules = $this->buildRules($this->rules);
         }
     }
+    
+    /**
+     * return class name suffixed with 
+     **/
+    protected function $getCacheKey() {
+        return __CLASS__.'.'. Yii::getAlias('@app');
+    }
+
 
     /**
      * Adds additional URL rules.


framework/web/UrlManager.php
```diff
@@ -143,11 +143,6 @@ class UrlManager extends Component
      */
     public $normalizer = false;
 
-    /**
-     * @var string the cache key for cached rules
-     * @since 2.0.8
-     */
-    protected $cacheKey = __CLASS__;
```
dynasource: this will cause a BC

framework/web/UrlManager.php
```diff
@@ -188,6 +183,14 @@ public function init()
             $this->rules = $this->buildRules($this->rules);
         }
     }
+    
+    /**
+     * return class name suffixed with 
+     **/
+    protected function $getCacheKey() {
+        return __CLASS__.'.'. Yii::getAlias('@app');
```
MysteryDragon: White spaces missing

framework/web/UrlManager.php
muayyad-alsadi: @dynasource  first this is protected, second I implemented a getter with same name

samdark: A setter should be there as well.

samdark: The property should become private. Getter should return default value as you've proposed if property it's not set explicitly.

samdark (changes requested):
See comments. Also formatting should be fixed and phpdoc should be in order.

muayyad-alsadi:
@samdark  one of the suggestions is to have a public cacheKeyPrefix that defaults to `__CLASS__` and  no need for any getter as we append `@app` to it. This might not seem backward compatible but it's because it was private/protected.

What do you think?

samdark:
Not being able to remove `@app` from it sounds bad to me.

muayyad-alsadi:
is there a way to distinct different routes in different apps other than `@app`.

with current Yii code it's the key is a protected property that you can't change. In the new approach I allow you to change the prefix.

Another approach is to append `$hash` to the prefix, the downside when one changes the route, a redundant stall key will remain.

so `$hash` and `@app` I can't think of any other thing.

Btw I don't like having a setter because the cache is get/set at the initializing of yii app (loading routes) setting it later have no meaningful purpose.

If someone want to change it, he should extend the class.

samdark:
Yes. A manually set prefix, for example. It's a common case when in backend you're using another instance of URLManager with the frontend rules to link to frontend.

muayyad-alsadi:
here is the use case that you are talking about,

https://github.com/yiisoft/yii2-app-advanced/blob/master/docs/guide/topic-link-backend-frontend.md

I need to know the `@app` (or whatever that identifies it) of the app for which the url rules belong to.
If we are able to do this, then providing custom prefix is meaning less.

Let me read more to see how to do that.

muayyad-alsadi:
```
        'urlManager' => [
            'enablePrettyUrl' => true,
            'showScriptName' => false,
            'rules' => require 'urls.php',
        ],
        'urlManagerApp2' => [
            'enablePrettyUrl' => true,
            'showScriptName' => false,
            'rules' => require 'urls2.php',
            'cachePrefix'=>'app2-urls'
        ],
```

I'm trying to eliminate the need of passing cachePrefix (principle of least astonishment)
I think there is there a way to get the name of the component (`urlManagerApp2`)
So we are left with stressing that cachePrefix should be customized.

samdark:
Name of the component will be different for frontend app and backend app in this case. Rules will be the same.

muayyad-alsadi:
in that case I prefer using the hash, but we should have ttl (public configrable ttl) to avoid stall  cache of unused hashes

samdark:
How about https://github.com/yiisoft/yii2/issues/12905#issuecomment-258259468?

muayyad-alsadi:
I've replaced this with the suggested approach,

* make cacheKey public
* make it expand using `getAlias`
 
https://github.com/yiisoft/yii2/pull/12926

framework/web/UrlManager.php
```diff
@@ -190,6 +185,14 @@ public function init()
     }
 
     /**
+     * return class name suffixed with 
+     **/
+    protected function getCacheKey() {
```
SilverFire: `{` should be move to the new line

framework/web/UrlManager.php
```diff
@@ -190,6 +185,14 @@ public function init()
...
+        return __CLASS__.'.'.Yii::getAlias('@app');
```
SilverFire: `return __CLASS__ . '.' . Yii::getAlias('@app');`

framework/web/UrlManager.php
```diff
@@ -190,6 +185,14 @@ public function init()
...
+    }
+
```
SilverFire: Redundant line

framework/web/UrlManager.php
```diff
@@ -190,6 +185,14 @@ public function init()
     }
 
     /**
+     * return class name suffixed with
```
SilverFire: PHPDoc should be corrected

SilverFire (changes requested):

framework/web/UrlManager.php
muayyad-alsadi: Check my other pr with different approach. There is no need for getters nor setters

samdark:
This approach isn't good in its inflexibility. Since there's better alternative in for of https://github.com/yiisoft/yii2/pull/12926, I'm closing this one.

SilverFire:
Agreed. Didn't see #12926 when commented this PR

PR #12909, opened 2016-11-03
moved & updated test 'testFindScalar' from trait to ActiveRecordTest.…

### LINKED ISSUE #12897: The test is not being used
The test \yiiunit\framework\ar\ActiveRecordTestTrait::testFindScalar() is not being used
There are other tests of a trait are not used.

https://github.com/LAV45/yii2/commit/8201dee0abbd997b109d8a3bc495ace915629020
https://travis-ci.org/LAV45/yii2/builds/172776879

dynasource:
fixes #12897

### DIFF
--- a/tests/framework/ar/ActiveRecordTestTrait.php
+++ b/tests/framework/ar/ActiveRecordTestTrait.php
@@ -161,23 +161,6 @@ trait ActiveRecordTestTrait
         $this->assertFalse($customer->hasAttribute(42));
     }
 
-    public function testFindScalar()
-    {
-        /* @var $customerClass \yii\db\ActiveRecordInterface */
-        $customerClass = $this->getCustomerClass();
-
-        /* @var $this TestCase|ActiveRecordTestTrait */
-        // query scalar
-        $customerName = $customerClass::find()->where(['id' => 2])->scalar('name');
-        $this->assertEquals('user2', $customerName);
-        $customerName = $customerClass::find()->where(['status' => 2])->scalar('name');
-        $this->assertEquals('user3', $customerName);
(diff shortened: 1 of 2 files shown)


samdark (approved):

tests/framework/ar/ActiveRecordTestTrait.php
```diff
@@ -161,23 +161,6 @@ public function testHasAttribute()
...
-        $customerName = $customerClass::find()->where(['id' => 2])->scalar('name');
-        $this->assertEquals('user2', $customerName);
-        $customerName = $customerClass::find()->where(['status' => 2])->scalar('name');
-        $this->assertEquals('user3', $customerName);
-        $customerName = $customerClass::find()->where(['status' => 2])->scalar('noname');
-        $this->assertNull($customerName);
-        $customerId = $customerClass::find()->where(['status' => 2])->scalar('id');
-        $this->assertEquals(3, $customerId);
-    }
-
```
cebe: this is not correct, they may be used in non-db AR.

cebe (changes requested):

tests/framework/ar/ActiveRecordTestTrait.php
dynasource: according to @cebe, 'when the framework and extensions where in one repo that trait was shared between DB AR implementation and other, like Elasticsearch, redis and mongo. When we split the repos out, that was the workaround' more info https://github.com/yiisoft/yii2-redis/blob/master/tests/bootstrap.php#L17-L26

dynasource:
see https://github.com/yiisoft/yii2/pull/12909#discussion_r86779365

PR #12911, opened 2016-11-03
Profile cache time for performance purpose

vitalyzhakov:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12891


dynasource (dismissed):
looks good to me

samdark (approved):
Need a line for CHANGELOG and some actual testing in DEBUG mode.

kidol:
- Just use `$key` as variable name?
- In other cases it's: `use Yii;`

vitalyzhakov:
1. `$key` type is mixed, thats why I use `$buildedKey`
2. I thought about it, but doesn't find any difference between `use Yii;` and `\Yii` and select short notation.

If this project have code style rules, which recommended another behavior, I will replace PR.

kidol:
> $key type is mixed, thats why I use $buildedKey 

So, what exactly is the downside when you would use `$key`? I don't understand.

>I thought about it, but doesn't find any difference between use Yii; and \Yii

It's just about consistency - no need to break it.

vitalyzhakov:
> what exactly is the downside when you would use $key
First parameter to `\Yii::trace()` must be string. If we send object instead of string, we will have an error, I think.

>Need a line for CHANGELOG and some actual testing in DEBUG mode.
How I can help with this?

kidol:
>If we send object instead of string, we will have an error, I think.

But we don't since `$this->buildKey()`  returns a string so `$key` would become a string too.

I made some suggestion for enhancement in #12775.

samdark (changes requested):
Please check if it's possible to adopt @kidol suggestion.

vitalyzhakov:
@kidol , I think, change the value of function actual parameter in function body is bad idea, because you may do mistake. Do you agree with me?

vitalyzhakov:
@Renkas , variable renamed, thank you!

@samdark , 
> Please check if it's possible to adopt @kidol suggestion.
It's possible, but I think this advice is wrong.

samdark:
How about `set` calls?

vitalyzhakov:
@samdark , I have covered `set` call. What are next steps?

samdark:
Well, I guess the rest of the call types...

samdark:
`exists`, `multiGet`, `multiSet`, `multiAdd`, `add`, `delete`, `flush`.

vitalyzhakov:
@samdark , may be this functions will be covered by another pull request?

dynasource:
shouldnt we solve this problem with DI?

samdark:
@dynasource what's your idea?

dynasource:
to replace Cache.php with something like ProfiledCache.php in the debugging environment with DI.

dynasource:
like in https://github.com/yiisoft-contrib/github-bot/pull/8/files#diff-cd881d809f549ba1740f96411777a312R109

samdark:
The possible pro I see is that original class will remain the same w/o complicating logic. If it could be done the way so it's turned on automatically in debug mode, I'm all for it.

vitalyzhakov:
How I can help for accept this pull to release?

SilverFire:
I agree with @dynasource that adding `if (YII_DEBUG)` in all places where we need optional debug/profiling/etc will make out code look like a nasty if-condition spaghetti.

Considering we age going to improve DI in Yii framework, the suggested idea of DI/DIC looks useful.

dynasource:
> How I can help for accept this pull to release?

@vitalyzhakov, in order for DI to be utilized, we have to create a new class overriding Cache, like i.e:
* ProfiledCache.php

This class contains all the logic about profiling.

With DI you are calling this class in the Debug environment instead of Cache.php (in bootstrap.php). You are using the Yii::createObject() mechanism in combination with $container->set like in https://github.com/yiisoft-contrib/github-bot/pull/8/files#diff-cd881d809f549ba1740f96411777a312R109:
```php
        Yii::$container->set('Github\Api\Repo', function ($container, $params) {
             return new \yiiunit\extensions\githubbot\mocks\RepoMock($params[0]);
         });
```

samdark:
That's interesting approach. @vitalyzhakov do you want to try it?

vitalyzhakov:
Now I have code in my web.php
```php
'dataCache' => [
            'class' => 'yii\caching\MemCache',
            'servers' => [
                [
                    'host' => getenv('MEMCACHED_DATA_ADDRESS'),
                    'port' => getenv('MEMCACHED_DATA_PORT'),
                ],
            ],
            'useMemcached' => true,
],
```
And index.php
```php
<?php
if (getenv('STAGE') === 'development') {
    defined('YII_DEBUG') or define('YII_DEBUG', true);
    defined('YII_ENV') or define('YII_ENV', 'dev');
}

require(__DIR__ . '/../vendor/autoload.php');
require(__DIR__ . '/../vendor/yiisoft/yii2/Yii.php');

$config = require(__DIR__ . '/../config/web.php');

(new yii\web\Application($config))->run();
```
How I can change Cache to ProfiledCache in dev environment with DI?

samdark:
Should be something like:

```php
Yii::$container->set('yii\caching\Cache', 'yii\caching\LoggedCache');
```

samdark:
But I'm not sure how it would work because of inheritance chain. @dynasource ?

cebe:
This does not really work here because we have a lot of classes extending `yii\caching\Cache` for the real implementations. We'd need a profiled version for all of these if we'd want to use DI.
If we introduce a "ProfiledCache" class, that could be a wrapper around the actual class instead, e.g.

```php
class ProfiledCache extends Component
{
    public $cache;

    // in init() do Instance::ensure() on $this->cache

    public function get(...)
    {
         // begin profile
         $val = $this->cache->get(...)
         // end profile
         return $val;
    }

...
}
```

klimov-paul:
There should explicit fields-flags: `enableLogging` and `enableProfiling`, like it is done at `yii\mongodb\Connection`:
https://github.com/yiisoft/yii2-mongodb/blob/master/Connection.php#L138
https://github.com/yiisoft/yii2-mongodb/blob/master/Connection.php#L144

dynasource:
> We'd need a profiled version for all of these if we'd want to use DI.

correct, forgot to mention that.

samdark:
@dynasource then it's a bit too much IMO.

samdark:
@vitalyzhakov so:

1. Let's continue with the approach you've followed so far.
2. I think https://github.com/yiisoft/yii2/pull/12911#issuecomment-263526369 makes sense. Let's have it.

dynasource:
> @dynasource then it's a bit too much IMO.

Can't we fix this side effect in the DI layer?

samdark:
Not without significantly adjusting classes design which is definitely not for 2.0.x.

klimov-paul:
Actual implementation:
https://github.com/yiisoft/yii2-mongodb/blob/master/Command.php#L798-L833
https://github.com/yiisoft/yii2-mongodb/blob/master/Command.php#L179-L190

samdark:
So how about the rest type of the calls? `set` etc.?

vitalyzhakov:
`Set` call is covered.
In my project I have only `set` and `get` calls. And test in it. I'm not sure, that if I cover other method, it will work.
May be other call will covered by future PR?

samdark:
Well, it's a bit weird to have part of methods accessing storage covered and part not.

vitalyzhakov:
Ok, I have covered all methods. How we make sure, that all works?

framework/caching/Cache.php
```diff
@@ -462,7 +474,9 @@ protected function getValues($keys)
     {
         $results = [];
         foreach ($keys as $key) {
+            $this->beginProfile($key, __METHOD__);
```
samdark: This one isn't part of public interface. No need to profile it.

framework/caching/Cache.php
```diff
@@ -480,9 +494,11 @@ protected function setValues($data, $duration)
     {
         $failedKeys = [];
         foreach ($data as $key => $value) {
+            $this->beginProfile($key, __METHOD__);
```
samdark: This one isn't part of public interface. No need to profile it.

framework/caching/Cache.php
```diff
@@ -500,9 +516,11 @@ protected function addValues($data, $duration)
...
```
samdark: This one isn't part of public interface. No need to profile it.

samdark (changes requested):

samdark:
@vitalyzhakov do all operations possible with cache in an action then go to debug panel and see if they're all there.

framework/caching/Cache.php
```diff
@@ -78,6 +79,11 @@
      */
     public $defaultDuration = 0;
 
+    /**
+     * @var bool whether to enable profiling the queries
```
cebe: I would say "cache operations" instead of "queries".

framework/caching/Cache.php
```diff
@@ -537,4 +569,29 @@ public function offsetUnset($key)
     {
         $this->delete($key);
     }
+
+    /**
+     * Marks the beginning of a code block for profiling.
+     * @param string $token token for the code block
+     * @param string $category the category of this log message
+     * @see endProfile()
+     */
```
cebe: `@since 2.0.11`

framework/caching/Cache.php
```diff
@@ -537,4 +569,29 @@ public function offsetUnset($key)
...
+    {
+        if ($this->enableProfiling) {
+            Yii::beginProfile($token, $category);
+        }
+    }
+    /**
+     * Marks the end of a code block for profiling.
+     * @param string $token token for the code block
+     * @param string $category the category of this log message
+     * @see beginProfile()
```
cebe: `@since 2.0.11`

cebe (changes requested):
The following methods have no profiling:

- multiGet()
- multiSet()
- multiAdd()

instead you have implemented it inside of setValues(), addValues() and getValues(). this should be moved to the multiGet()/multiSet()/multiAdd() functions, to be consistent with other methods and also because getValues() and setValues() could be overridden in extended classes.

framework/caching/Cache.php
vitalyzhakov: This is part of public `multiGet`. If I move what will be key for `beginProfile`?

cebe: @vitalyzhakov see my comments below: https://github.com/yiisoft/yii2/pull/12911#pullrequestreview-11046723

vitalyzhakov: So, what will be key for `beginProfile` in `multiAdd`?

cebe: Sorry I did not understand the question first. Maybe a combination of the keys via implode() that could get really long however.

vitalyzhakov:
What are the next steps?

framework/caching/Cache.php
```diff
@@ -187,15 +202,15 @@ public function multiGet($keys)
...
-                        : call_user_func($this->serializer[1], $values[$newKey]);
+                    $value = $this->serializer === null ? unserialize($values[$newKey]) : call_user_func($this->serializer[1], $values[$newKey]);
 
                     if (is_array($value) && !($value[1] instanceof Dependency && $value[1]->isChanged($this))) {
                         $results[$key] = $value[0];
                     }
                 }
             }
         }
+        $this->endProfile($keyMap, __METHOD__);
```
cebe: In `get()` we currently only profile the interaction with the cache backend, here we also include deserialisation. This should be consistent, so endProfile call should be moved to line 198.

framework/caching/Cache.php
```diff
@@ -358,9 +389,11 @@ public function add($key, $value, $duration = 0, $dependency = null)
         } elseif ($this->serializer !== false) {
             $value = call_user_func($this->serializer[0], [$value, $dependency]);
         }
-        $key = $this->buildKey($key);
 
-        return $this->addValue($key, $value, $duration);
+        $addResult = $this->addValue($builtKey, $value, $duration);
+        $this->endProfile($builtKey, __METHOD__);
```
cebe: same as above, profiling should only be done for the cache interaction, not the serializaion to be consistent with the other methods.

framework/caching/Cache.php
```diff
@@ -468,6 +505,7 @@ protected function setValues($data, $duration)
             if ($this->setValue($key, $value, $duration) === false) {
                 $failedKeys[] = $key;
             }
+            $this->endProfile($key, __METHOD__);
```
cebe: this line seems to be left over.

cebe: Seems there are no tests that run with enableProfiling = true. These should be added as they are expected to fail at this line.

cebe (changes requested):
looks good to me in general, except the comments.

What is needed:

- unit tests run with enableProfiling=true
- method consistency about whether serializing is profiled or not.

framework/caching/Cache.php
samdark: I agree with overall consistency though.

framework/caching/Cache.php
samdark: See https://github.com/yiisoft/yii2/pull/12911#discussion_r90924063

framework/caching/Cache.php
cebe: Yes, current cache class tests are located in https://github.com/yiisoft/yii2/blob/master/tests/framework/caching/ these should be extended so that they run once with `enableProfiling = true` and once with `enableProfiling = false` to test these variations.
Here is how to run tests: https://github.com/yiisoft/yii2/tree/master/tests#yii-20-unit-tests

Let us know if you need help on this.

vitalyzhakov:
@cebe , can you help me with unit tests idea?
I see CacheTestCase, how add `enableProfiling` field with `true` and `false` values in all child classes?

Next, I try torun test, but it does not work
```php vendor/bin/phpunit tests/framework/caching/CacheTestCase.php 
Class 'tests/framework/caching/CacheTestCase' could not be found in 'XXX/tests/framework/caching/CacheTestCase.php'
```

Test with groups run successfully

framework/caching/Cache.php
```diff
@@ -217,6 +235,8 @@ public function multiGet($keys)
      */
     public function set($key, $value, $duration = null, $dependency = null)
     {
+        $this->beginProfile($builtKey, __METHOD__);
```
bashkarev: Undefined variable: builtKey.
forgot this
```php
 $builtKey = $this->buildKey($key);
```

bashkarev:
Token has no debug information.
Would be good to add trace info
```php
Yii::trace([
    'class' => $this::className(),
    'token' => $token,
    'options' => [
        'key' => $key,
        'value' => $value,
        'result' => $setResult,
        'duration' => $duration,
        'dependency' => $dependency,
    ]
], $category);
``` 
or debug Event

framework/caching/Cache.php
```diff
@@ -537,4 +582,46 @@ public function offsetUnset($key)
...
+            return $srcToken;
+        }
+        return join(', ', $srcToken);
+    }
+
+    /**
+     * Marks the beginning of a code block for profiling.
+     * @param string $token token for the code block
+     * @param string $category the category of this log message
+     * @see endProfile()
```
bashkarev: @see Yii::beginProfile()

framework/caching/Cache.php
```diff
@@ -537,4 +582,46 @@ public function offsetUnset($key)
...
+        if ($this->enableProfiling) {
+            Yii::beginProfile($this->getProfileKey($token), $category);
+        }
+    }
+
+    /**
+     * Marks the end of a code block for profiling.
+     * @param string $token token for the code block
+     * @param string $category the category of this log message
+     * @see beginProfile()
```
bashkarev: @see Yii::beginProfile()

bashkarev:
not unique profile token
```php
$cache = \Yii::$app->getCache();
$cache->get("key.0");    // token profile 'c07358804335293043d7b8e1b11c92b3'
$cache->set("key.0", 0); // token profile 'c07358804335293043d7b8e1b11c92b3'
$cache->delete("key.0"); // token profile 'c07358804335293043d7b8e1b11c92b3'
```

bashkarev:
https://github.com/yiisoft/yii2/pull/12911#issuecomment-271100950 with additional data, can make a debug panel
image

SilverFire:
@vitalyzhakov Would you like to apply the requested changes during this week to get this PR in 2.0.11?

vitalyzhakov:
Hello!
I think, big changes need more testing.
If this PR holds release, may be move to next milestone?

samdark:
Good idea in this case. Let's merge it into 2.0.12...

Issue #12912, opened 2016-11-03
Repeat of past migrations


RichWeber:
### What steps will reproduce the problem?

I have some migrations before v2.0.10

<img width="666" alt="2016-11-03 21 52 47" src="https://cloud.githubusercontent.com/assets/1702252/19982726/e71dc42e-a20f-11e6-9e8b-8e6d72cdc1ce.png">

Versions names are without namespaces.

### What is the expected result?

When I added the new config for `migrationNamespaces` I see that my past migrations were not executed.

<img width="566" alt="2016-11-03 21 59 47" src="https://cloud.githubusercontent.com/assets/1702252/19982986/e5acaa14-a210-11e6-9dc0-317b90036b71.png">

My configuration:

```
'controllerMap' => [
        'migrate' => [
            'class' => 'yii\console\controllers\MigrateController',
            'migrationNamespaces' => [
                'console\migrations',
                'usolutions\sms\migrations',
                //'lajax\translatemanager\migrations', // wait for namespace in module
            ],
            'migrationPath' => null, // allows to disable not namespaced migration completely
        ],
    ],
```

Migrations table:

<img width="697" alt="2016-11-03 22 00 59" src="https://cloud.githubusercontent.com/assets/1702252/19983065/2137ca3c-a211-11e6-9191-7465c6abc951.png">

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 5.6
| Operating system | Linux

klimov-paul:
This is expected behavior. If migration has been applied without namespace in the past - it should remain to be without namespace.

You should manually adjust migration history to contain namespaced names if you wish o refactor migrations to namespace
.
This can be done using `migrate mark` command:
1) mark all migrations down
2) change config to use namespaces
3) add namespaces to migration source codes
4) mark all migrations up

RichWeber:
OK, I understand

But, it turns out all of these actions are to be done by all the team members on the local machines. And then still have it all to do on servers.

Okay, it's not a problem if there are couple dozen migrations. But if there are hundreds or thousands - it will be a very long process, which can lead to errors.

Why can not we verify the version of migration without namespaces and with it?

dynasource:
@klimov-paul, for what reason are namespaces so important for migrations? Arent the base names not unique in 99,99% of the time. Or are we already takings migrations into account without the m000000_000000 formatting?

klimov-paul:
[...]

PR #12913, opened 2016-11-03
#12912 fix repeat of past migrations

RichWeber:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Fixed issues  | #12912


samdark (changes requested):
Need a line for changelog.

PR #12923, opened 2016-11-04
fix line with wrong return statement

### LINKED ISSUE #7182: ActiveRecord extraFields returns null
I've read the documentation and it says that **yii\db\ActiveRecord** returns all the relations as default in the **extraFields**.

I've a model that has some relations but the **extraFields** returns null.

I've debugged some of the Yii code and I've seen that the _related property is null for the class.

dynasource:
fixes #7182


samdark:
Merged.

Issue #11638, opened 2016-05-26
[2.1] Proposal for handling primary keys in migrations.


Faryshta:
Currently the class `yii\db\Migration` has a methods `primaryKey()` and `bigPrimaryKey()` which create a column based on `Schema::TYPE_PK` or `Schema::TYPE_BIGPK` respectively.

I propose to move the `primaryKey()` method to  the `yii\db\ColumnSchemaBuilder` class as a column modifier with the signature.

```php
/**
 * @var boolean|null $autoIncrement if this column use AUTO_INCREMENT.
 * this variable is null then the value will be determined on whether or not this
 * column is integer or big interger.
 * @return $this the column schema builder itself
 */
public function primaryKey($autoIncrement = null);
```

Usage

```php
public function up()
{
    $this->createTable('post', [
          'id' => $this->integer()->primaryKey(), // automatically sets AUTO_INCREMENT
          //...
    ]);

    $this->createTable('tag', [
          'tag' => $this->string()->primaryKey(), // doesn't set AUTO_INCREMENT
          //...
    ]);
}
```

And for mixed primary keys I propose to create a new method. `yii\db\Migration::tablePrimaryKey()`  (the name can be different) with signature.

```php
/**
 * @var string[] $columns the columns to be used.
 * @return string
 */
public function tablePrimary(...$columns):
```

Usage

```php
public function up()
{
    $this->createTable('post', [
          'id' => $this->integer()->primaryKey(),
          //...
    ]);

    $this->createTable('tag', [
          'tag' => $this->string()->primaryKey(),
          //...
    ]);

    $this->createTable('post_tag', [
          'post_id' => $this->integer()->foreignKey(),
          'tag' => $this->string()->foreignKey(),

          $this->tablePrimaryKey('post_id', 'tag');
    ])
}
```

the `...$column` syntax is available since php5.6 but we can use simple arrays instead. If you don't like the method name `tablePrimaryKey()` please propose something else in comments.

samdark:
Yes. Good idea.

mdmunir:
No break BC if only add primaryKey() method.

Faryshta:
@samdark how about we folow the advice of @mdmunir ? and make `Migration::primaryKey()` a shortcut for `Migration::integer()->primaryKey()` ?

The `Schema::TYPE_PK` nad siilar constants wouldn't be useful anymore so I deleted them already on the branch where I am working for this PR.

Faryshta:
[...]

Issue #10889, opened 2016-02-17
Migration $this->primaryKey()->unsigned() don't work for MySQL!


Dromok:
Showed error "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'UNSIGNED'".

Because it generated "`id` int(11) NOT NULL AUTO_INCREMENT PRIMARY KEY UNSIGNED" invalid sql.
Need correct sql like this "`id` int(11) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY".

MysteryDragon:
+1
Oracle MySQL v5.7.11

samdark:
`primaryKey()` uses schema `pk` type which is `int(11) NOT NULL AUTO_INCREMENT PRIMARY KEY`.

samdark:
Same as `'id' => 'pk'` in the old syntax.

SilverFire:
So you think it's a wontfix?

MysteryDragon:
I noticed some mistakes in my solutions:
1. We can't set the AUTO_INCREMENT attribute this ways. Only alterTable or SQL expression can help us here, I guess.
2. "PRIMARY" isn't good name for PRIMARY index. :) It may be blank string or any correct index name but not the "PRIMARY". Or else we'll receive an MySQL exception.
So, to my mind it's quite nice to set this attribute in some other way.

df2:
```php
public function unsignedPrimaryKey();
public function unsignedBigPrimaryKey();
```

samdark:
How about `charPrimaryKey()`, `stringPrimaryKey()`?

MysteryDragon:
Maybe `ColumnSchemaBuilder::autoIncrement()`?
But in this case such method must define some index type.
Or could Yii replace schema parameters, if we extend it by `unsigned()` methods and so on? This is expected behavior of such methods, I guess.

df2:
@samdark It makes sense that ```pk()->unsigned()``` would produce a primary key as defined by Yii, but an unsigned version.

This is an argument about a virtual capacity. It doesn't _really_ matter that we're throwing away half our "pool of resources" in this case. There is perhaps an OCD affection that drives one nuts tho, when you could access that "half" by the UNSIGNED modifier.

mdmunir:
[...]

PR #12925, opened 2016-11-05
Primary key for column schema builder

mdmunir:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | #11638 #10889

* Introduce more relevant name for auto increment column
```php
[
    'id' => $this->serial(), // instead of $this->primaryKey()
]
```
* Add `primaryKey()` to `ColumnSchemaBuilder`
```php
        $this->createTable($cache->cacheTable, [
            'id' => $this->string(128)->primaryKey(),
            'expire' => $this->integer(),
            'data' => $this->binary(),
        ], $tableOptions);
```


framework/db/ColumnSchemaBuilder.php
```diff
@@ -317,12 +332,13 @@ protected function buildNotNullString()
...
-     * Builds the unique constraint for the column.
-     * @return string returns string 'UNIQUE' if [[isUnique]] is true, otherwise it returns an empty string.
+     * Builds the unique or primary constraint for the column.
+     * @return string returns string 'PRIMARY KEY' if [[isPrimary]] is true, else
+     * returns string 'UNIQUE' if [[isUnique]] is true, otherwise it returns an empty string.
      */
     protected function buildUniqueString()
     {
-        return $this->isUnique ? ' UNIQUE' : '';
+        return $this->isPrimary ? ' PRIMARY KEY' : ($this->isUnique ? ' UNIQUE' : '');
```
dynasource: could be improved on readability. Switch statement on true.

dynasource:
to add to the conversion:
https://dev.mysql.com/doc/refman/5.5/en/numeric-type-overview.html
'SERIAL is an alias for BIGINT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE.'

framework/db/SchemaBuilderTrait.php
```diff
@@ -44,9 +44,9 @@
      * @param integer $length column size or precision definition.
      * This parameter will be ignored if not supported by the DBMS.
      * @return ColumnSchemaBuilder the column instance which can be further customized.
-     * @since 2.0.6
+     * @since 2.0.11
      */
-    public function primaryKey($length = null)
+    public function serial($length = null)
     {
         return $this->getDb()->getSchema()->createColumnSchemaBuilder(Schema::TYPE_PK, $length);
```
dynasource: if you look at mysqls definition for SERIAL (SERIAL is an alias for BIGINT UNSIGNED NOT NULL AUTO_INCREMENT UNIQUE.), this is a bit confusing and perhabs error prone, using the Schema::TYPE_PK.

dynasource (changes requested):
I am missing some tests for other DBs like postgresql & mssql. Is there a specific reason no to include?

framework/db/ColumnSchemaBuilder.php
mdmunir: ya.. you are right.

framework/db/SchemaBuilderTrait.php
mdmunir: only mysql right? IMO, `serial()` still more make sense than `primaryKey()`.

dynasource: please elaborate

rob006: Why not `autoIncrement()`?

framework/db/ColumnSchemaBuilder.php
```diff
@@ -154,6 +159,16 @@ public function null()
     }
 
     /**
+     * Adds a `PRIMARY KEY` constraint to the column.
+     * @return $this
```
SilverFire: `@since`

framework/db/SchemaBuilderTrait.php
```diff
@@ -270,4 +270,22 @@ public function money($precision = null, $scale = null)
         }
         return $this->getDb()->getSchema()->createColumnSchemaBuilder(Schema::TYPE_MONEY, $length);
     }
+
+    /**
+     * @deprecated since version 2.0.11
+     * @see [[serial()]]
```
SilverFire: `@since 2.0.6`

framework/db/SchemaBuilderTrait.php
```diff
@@ -270,4 +270,22 @@ public function money($precision = null, $scale = null)
...
+     */
+    public function primaryKey($length = null)
+    {
+        return $this->serial($length);
+    }
+
+    /**
+     * @deprecated since version 2.0.11
+     * @see [[bigSerial()]]
```
SilverFire: `@since 2.0.6`

SilverFire (changes requested):

SilverFire:
I don't like `serial()` much. This name is not self-explanatory and does not reference to the appropriate DBMS concept.

mdmunir:
> I don't like serial() much

postgre use `serial` as type.

framework/db/ColumnSchemaBuilder.php
SilverFire: Resolved

framework/db/SchemaBuilderTrait.php
```diff
@@ -270,4 +270,24 @@ public function money($precision = null, $scale = null)
         }
         return $this->getDb()->getSchema()->createColumnSchemaBuilder(Schema::TYPE_MONEY, $length);
     }
+
+    /**
+     * @deprecated since version 2.0.11
```
SilverFire: Deprecation does not mean that the method description should be dropped. Could you restore it?

SilverFire (changes requested):
DOCS update is required

samdark:
Won't make it into 2.0.

PR #12926, opened 2016-11-05
fixes #12905: expanded using get alias

### LINKED ISSUE #12905: unnecessary cache misses

muayyad-alsadi:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12905

### DIFF
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -144,10 +144,10 @@ class UrlManager extends Component
     public $normalizer = false;
 
     /**
-     * @var string the cache key for cached rules
+     * @var string the cache key for cached rules, this will be expanded using `Yii::getAlias`
      * @since 2.0.8
      */
-    protected $cacheKey = __CLASS__;
+    public $cacheKey = '@app/'.__CLASS__;
 
     private $_baseUrl;
     private $_scriptUrl;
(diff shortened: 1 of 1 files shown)


dynasource:
why a new PR?

muayyad-alsadi:
it's a different approach, which uses a method suggested by @kidol

muayyad-alsadi:
I have php 5.6. It seems this approach did not work with php < 5.5.

muayyad-alsadi:
to implement this in a php 5.4, we need to create a protected property and a getter and a setter.

dynasource:
because of the string concationation, right. 

But looking at your commit, it could also be solved by using:
'public $cacheKey = '@app/yii\web\UrlManager ';

muayyad-alsadi:
@dynasource yes, by making it literal

samdark (changes requested):
Looks good to me. Need unit tests covering it.

cebe (changes requested):
See https://github.com/yiisoft/yii2/issues/12905#issuecomment-259814617

cebe:
closing as this is not the solution to the problem. see https://github.com/yiisoft/yii2/issues/12905#issuecomment-260491558 for more details.

Issue #12021, opened 2016-07-26
Allow to set undefined properties in yii\base\Object


mikehaertl:
For 2.1 I'd like to propose a change in `yii\base\Object` to make it behave like standard PHP objects again, when it comes to setting undefined properties.

This change could also be configurable, if some consider it to be too drastical.

### What is it about?

```php
<?php
class X extends yii\base\Object
{
}

$x = new X;

// This will throw an error:
$x->demo = 'x';
```
You will see an `yii\base\UnknownPropertyException`.

### How does PHP behave usually in this case?

Any class in PHP allows to set undefined properties, just like `stdClass`. So if you take the above example but don't extend from `yii\base\Object`, it will work:

```php
<?php
class X
{
}

$x = new X;
$x->demo = 'x';   // No error

echo $x->wtf;  // Error, because 'wtf' ist not set yet.
```

All Yii2 objects break this standard behavior, so you lose a feature that works for all other PHP classes.

### Why does this matter?

Say you have a complex SOAP service with 100s of SOAP types, some nested into each other. For example there could be a `User` type object somewhere in the SOAP result. By default PHP will return a instance of `stdClass` with all the properties set. Something like this:

```
 stdClass Object
(
    [id] => 123456,
    [firstName] => 'John',
    [lastName] => 'Donson'
    [street] => '...'
...
```

Now you wish to have a `fullName` property. You therefore could define a custom class with a getter method like:

```
namespace app\models\soap;
class User
{
    public function getFullName()
    {
        return $this->firstName.' '.$this->lastName;
    ]
}
```

And configure it in the `classmap` option of your `SoapClient`:

```
'classmap' => [
    'User' => 'app\models\soap\User',
],
```

This works, but you still have to call `$user->getFullName()`. If you like to do it the Yii way and have access to `$user->firstName` you can't, because your custom class can not extend from `yii\base\Object`. You would have to define each and every property of the `User` SOAP type manually - something you don't have to do in pure PHP classes or in `stdClass` in this situation.

### Proposed solution for Yii 2.1

Instead of throwing an error in `Object::__set()` for undefined properties, it should simply set the property:

```
    public function __set($name, $value)
    {
        $setter = 'set' . $name;
        if (method_exists($this, $setter)) {
            $this->$setter($value);
        } elseif (method_exists($this, 'get' . $name)) {
            throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
        } else {
            // Instead of throwing exception:
[...]

PR #12929, opened 2016-11-06
[WIP] Extract `Object` and `Component` into traits

### LINKED ISSUE #9748: Extract events from Component class into trait
Currently, if I want to use events in my class, I have to make it a child fo the `Component`. But `Component` class have a lot of other functions, that I do not need.

Whould not it be more useful, if we have somthing like `EventsTrait` , which we can to use in any class?

``` php
trait EventsTrait
{
   public function on(...) { ... }
   public function off(...) { ... }
   public function trigger(...) { ...}
}
```

``` php
class Component
{
   use EventsTrait;
}
```

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #9748

This is only proof of concept, it still need adjust naming and fix descriptions.

Closes #9748


samdark:
I don't think `Object::className()` is useful since we're raising PHP requirements and there's going to be native `::class` support.

samdark:
Makes sense to me overall. Do you think it worth also separating events and behaviors?

rob006:
This could be merged into 2.0.x IMO. And if we want to prepare to drop `Object` and `Component` and propose an alternative to deprecated `Object` and `Component` it should provide `Object::className()` implementation too.

> Do you think it worth also separating events and behaviors?

Yes, but I'm not sure if there is clean way to do this.

samdark (approved):

samdark:
@yiisoft/core-developers need your opinions.

framework/base/ComponentTrait.php
```diff
@@ -0,0 +1,595 @@
...
+                if ($behavior->hasMethod($name)) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Returns a list of behaviors that this component should behave as.
```
dynasource: I see duplication with the @docs of the ComponentInterface. The docs in the traits can be replaced with @inheritdoc tags.

dynasource (changes requested):
this proof of concept looks good. I approve the split in traits. I agree with @samdark that we should also split the Events & Behaviors. Those two look less entangled as it first seems. Isnt it possible to create BehaviorTrait.php and a EventTrait in which the EventTrait includes BehaviorTrait.

With respect to className(), that one should stay in Component.php with a @deprecated mark. In 2.1 it will be removed.

framework/base/ComponentTrait.php
samdark: How IDEs would react to that?

rob006: Using `@inheritdoc` in trait is incorrect, since traits doesn't handle any inheritance. It even break `yii2-apidoc` as far as I remember.

rob006:
> Isnt it possible to create BehaviorTrait.php and a EventTrait in which the EventTrait includes BehaviorTrait.

Doesn't should be opposite? BehaviorTrait includes EventTrait?

> With respect to className(), that one should stay in Component.php with a @deprecated mark. In 2.1 it will be removed.

I that case we can't say "Don't use deprecated `Object` and `Component`, use new traits" because they're missing one of main features of `Object`.

dynasource:
> Doesn't should be opposite? BehaviorTrait includes EventTrait?

The event trait needs ```php->ensureBehaviors()``` https://github.com/yiisoft/yii2/pull/12929/files#diff-9c14d263e64e7854e2ef80847ef021b5R397, not the other way around

> I that case we can't say "Don't use deprecated Object and Component, use new traits" because they're missing one of main features of Object.

Is that a problem? People can still be advised to use the new traits? And if they want to use the current className(), they still can. Its required for merging in 2.0.x/

framework/base/ComponentTrait.php
rob006: This gives you completely useless IDE hints:

```php
interface FooInterface {
	
	/**
	 * Interface description.
	 */
	public function bar();
}

trait FooTrait {
	
	/**
	 * Implements FooInterface::bar().
	 */
	public function bar() {
		echo 'bar';
	}
}

class Foo implements FooInterface {
	use FooTrait;
}
```
2fb89f21

rob006:
> The event trait needs php->ensureBehaviors()

Behaviors uses Events: https://github.com/yiisoft/yii2/blob/master/framework/base/Behavior.php .

> Is that a problem?

If we are deprecating `Object::className()` without any alternative, this is a problem.

framework/base/ComponentTrait.php
rob006: It doesn't change anything - you still doesn't have any real documentation, only references.

dynasource:
> Behaviors uses Events: https://github.com/yiisoft/yii2/blob/master/framework/base/Behavior.php .
Another entanglement. Lets fix this in another PR and keep this one clean.

> If we are deprecating Object::className() without any alternative, this is a problem.
* it is deprecated in 2.0.x but it still  can be used
* it will be removed in 2.1 and be replaced with ::class (discussed in other topic)

framework/base/ComponentTrait.php
samdark: Yeah. References are not enough...

samdark: Is there a way to re-structure it the way not to duplicate docs?

samdark:
@rob006 isn't PHP 5.6 `MyClass::class` an alternative to `Object::className()` for most cases?

framework/base/MutatorsTrait.php
```diff
@@ -0,0 +1,189 @@
...
+ */
+
+namespace yii\base;
+
+/**
+ *
+ * @author Robert Korulczyk <robert@korulczyk.pl>
+ * @since 2.0.11
+ */
+trait MutatorsTrait
```
klimov-paul: I am unsure about the naming.
Better be `VirtualPropertyTrait`

klimov-paul (changes requested):

klimov-paul:
> Alternative approach for #7936 

Dispite the change itself can be useful, this PR does not solves issue #7936 anyhow. In case `object` becomes a PHP reserved word the Yii core code will still crash.

I am sure we have a disscussion for such improvement somewhere - but I can not find it.
Actually related issue is #9748.

klimov-paul:
Related disscussion:
https://github.com/yiisoft/yii2/issues/1837#issuecomment-31916630

rob006:
@klimov-paul We can't really fix #7936 in 2.0.x, but extracting `Object` and `Component` into traits will make it pretty easy in 2.1. The only reason to leave `Component` and `Object` in this PR is BC.

@samdark You can't use `MyClass::class` in PHP 5.4, so this is not an alternative for Yii 2.0. IMO it is better to create another trait with `className()` implementation than just leave it in `Object`.

samdark:
Right. I agree on another trait for `className()`.

klimov-paul:
> We can't really fix #7936 in 2.0.x, but extracting Object and Component into traits will make it pretty easy in 2.1

Then issue #7936 should not be placed under 'Fixed issues' section as it produces a delusion.

> The only reason to leave Component and Object in this PR is BC.

Disagreed: `Component` and `Object` should not be dropped as constant usage of traits and interface implementation is not practical.
Compare:

```php
class Foo extends \yii\base\Object
{
}
```

with:

```php
class Foo implements \yii\base\Configurable, \yii\base\MutatorsInterface
{
    use \yii\base\ConfigurableTrait;
    use \yii\base\MutatorsTrait;
{
}
```

framework/base/ComponentTrait.php
dynasource: clicking on the reference can be made quite obvious

klimov-paul:
The side effect of this PR is sugnificant increase of entities in class herarchy: additional interfaces and traits. It makes the source code a little bit twisted.

It would be nice to have @qiangxue opinion on this matter - he was the one insisted inheritance tree should be as simple as possible.

Extracting a traits for `Object` and `Component` functionality makes sense only for the '3rd party' library, which may use them to integrate some standalone classes into Yii in easy way.
However, in general this initiative was refused as DI supports object creation and configuration via `Closure` callbacks.

klimov-paul:
One more related disscussion: #12021

dynasource:
I agree with not deprecating Component & Object for clean & easy inheritance (with those traits). 

I do not agree with keep the current situation:
* perhabs its twisted because of the trait & interface combination, but such a combination is something to get used to. Btw., its already getting more and more common in practice.
* this traits approach makes the software more modular and cleaner organized
* this traits approach make the software more logical: currently Component is uselessly overriding methods of Object. This should be done cleaner (I now see in this PR that Component is still overriding Object. I expect this is for BC reasons?)
* like you said, the software will be more flexibile to integrate in 3rd party software

klimov-paul:
> perhabs its twisted because of the trait & interface combination, but such a combination is something to get used to.

Such combination is risk prone. Usage of traints is not the same as extending an class. The main drawback is `private` variables and methods scope. Any method or field decalred inside the trait as `private` will be accesible(!) in the class, which uses it:

```php
class Foo implements ComponentInterface
{
    use \yii\base\ComponentTrait;

    public function hackPrivateField()
    {
         return [
            $this->_events, // declared as `private` at `ComponentTrait`
            $this->_behaviors, // declared as `private` at `ComponentTrait`
        ];
    }
}
```

> this traits approach makes the software more modular and cleaner organized

One of the reasons why multiple inheritance was abandonded not long ago is because it is complex and twisted. It is always easier to follow single `extends` reference, then scouting over multiple traits and combine thier code in mind to create an actual picture.

> this traits approach make the software more logical: currently Component is uselessly overriding methods of Object.

It has no choice. At least without sugnificant performance degradation. I can see no problem `Component` overrides magic methods of the `Object`. It serves the same convension `Object` instroduces even, if it implements it in the other way. Direct inheritance of `Object` inside `Controller` is quite a logical since you can expect setters and getter to be supported as they are at `Object`.

The problem is there could not be an interface, which bounds behavior introducted by `Object` - that is why this class exists and it is at the top of hierarchy.

> the software will be more flexibile to integrate in 3rd party software

It hard to say if it makes a sugnificant benefit. It is hardly to imagine someone will intergrate these traits into `Symfony` or `Laravel` classes. 3rd party libraries are better to be wrapped rather then directly extended.

dynasource:
> Such combination is risk prone. Usage of traits is not the same as extending an class.

I see this as a minor side effect compared to the gain you get by using the flexible trait. In the end, its the developer who wants to make use of this added flexibility. He will have to accept this side effect. If he does not, he can still extend from Component or Object (which should not be deprecated IMO)

> One of the reasons why multiple inheritance was abandonded not long ago is because it is complex and twisted.

The current implementation with all methods in 1 class is also complex. Overriding many methods of Object by Component is also confusing and twisted.

> Currently Component is uselessly overriding methods of Object. It has no choice. At least without sugnificant performance degradation.

Do we have proof of this? Now we have a fatter inheritance tree, which does not seem optimal either.

> It serves the same convension Object instroduces even, if it implements it in the other way.

Which make you expect an interface

> The problem is there could not be an interface, which bounds behavior introducted by Object

which is to be solved with a shared trait

> The software will be more flexibile to integrate in 3rd party software. It hard to say if it makes a sugnificant benefit

well, people can choose if they want to use either the mutators, configurables, behaviors or events to create their third party software. They are not bound by the inheritance restriction of having to inherit from Object or Component. They can practice composition over inheritance and create a cleaner inheritance tree, only using what they want.

klimov-paul:
> I see this as a minor side effect compared to the gain you get by using the flexible trait. In the end, its the developer who wants to make use of this added flexibility.

No this is a framework problem then: normally any private method or field change does not breaks a BC - that is why we use `private` scope widely. But for the trait you create here changing of private scope will cause a BC to the developer uses this trait.

> Which make you expect an interface

Getter/setter usage for the virtual property creation is NOT an interface - it is a convernsion - and it is NOT a strict convension, which may be broken any time if needed by overriding magick methods and `canSetPropety()`/`canGetProperty()`. 
Framework itself breaks this convension at ActiveRecord level for relation methods, which are getters:

```
$model = new User(); // ActiveRecord, which has relation ‘status’

$this->assertEquals($model->getAttributes(), $model->attributes) // pass
$this->assertEquals($model->getStatus(), $model->status) // FAILED
```

See #583 for details

Also `ComponentInterface` and `MutatorInterface` are useless! There is no check for them anywhere in the code and should not ever be one. The behavior and events functionality are created to be completely hidden by the scenes as a replacement for the multiple inheritance.
Behavior can add a virtual property or method to its owner in the way it belongs to the owner itself - that is the cause. We do not use constructions like:

```php
$object->getBehavior('foo')->doSomething();
```

we invoke methods directly:

```php
$object->doSomething();
```

> people can choose if they want to use either the mutators, configurables, behaviors or events to create their third party software. They are not bound by the inheritance restriction of having to inherit from Object or Component. They can practice composition over inheritance and create a cleaner inheritance tree, only using what they want.

Yii follows several credos:

- simplicity over complexity
- convernsion over configuration

These are broken here. Why should ovecomplicate the code to cover some hypothetical use case? We never do so. It is a philosophy of such libraries as `Symfony` and `Zend` to cever every possible usecase, which may ever appear.
If you dream of code, where every step is covered by a separated interface and all functionality is split into a tiny fragments for maximum simplification of replacement - then you should took `Zend` or `Symfony` - they have it all already.
I can see no reason to compete with them in the matter of SOLID.

There are countless places in the Yii, where extra interface or extra trait can be extracted. If start now - there will no end to this. And what cause? If we all agree that strict SOLID is our goal, then we should simply abandon Yii developemnt and say: 'people go to Symfony', because it is already done there and it is unlikely we can do it better.

rob006:
> Getter/setter usage for the virtual property creation is NOT an interface

Same as `Configurable` interface. 

> There are countless places in the Yii, where extra interface or extra trait can be extracted. If start now - there will no end to this.

It has already been started - you have a couple of traits and corresponding interfaces in framework, and some of them could be avoided. But this will make code less flexible. Same as here.

klimov-paul:
> Same as Configurable interface.

No, it is not the same! Configurable interface demands exact method signature for `__construct()`.
It can decalre this method because of PHP language restrictions on it.
Virtual properties and methods can not be covered by interface even in theory.

> It has already been started - you have a couple of traits and corresponding interfaces in framework, and some of them could be avoided. But this will make code less flexible. Same as here.

We have traits only for ActiveRecord related classes, because they are reused at extensions like `yii2-redis`, `yii2-mongodb`, `yii2-elasticsearch`. These traits exist because there is no other option for thier case.
We keep only a minimun amount of interfaces, which covers only the code, which is likely to have custom implementation from developer like 'Data Provider', 'Active Record', 'Url Rule', 'Request parser'.
We do NOT have interfaces for such classes as `UrlManager`, `Request`, `Response`, `DI` and so on.

rob006:
> No, it is not the same! Configurable interface demands exact method signature for __construct().

No. It expects some behavior from `__construct()`, but implementing `Configurable` interface doesn't force any specific implementation. Same situation as `__get()` and `MutatorsInterface`.

klimov-paul:
It seems you do not bother to read the disscussion at #583

`Configurable` insterface creates a specific bound which is checked at DI level. The bound is the last argument should be an configuration array. This bound is used while creating a new object via DI.

Virtual properties and methods existance are the matter of documentation(!) they can not be explicitely declared or expected, there can not be some bound which tells that method `get...` creates a virtual property as it never force to do it. You can declare method `get...` with params:

```php
public function getAvg($data) {...}
```

or getter may exist but not meant to create a virtual propery.
And there is an ActiveRecord which relation getter produces virtual properties of the different type.

rob006:
> It seems you do not bother to read the disscussion at #583

I don't see how #583 is related to `Configurable` example. `Configurable` doesn't give you any warranty how constructor will work. This is valid code:

```php
class Foo implements Configurable {
}
```

So `Configurable` is not really interface, only some hack to express promise how object should work. Same as `MutatorsInterface` - it does not give you warranty that `$object->getSome() === $object->some` but tells you how accessing non-existent properties could work and declare some methods that implementing class must have - in this context it is more valid interface than `Configurable`.

klimov-paul:
`Configurable` is used inside DI, it checks if particular class implements it. It is a part of logic:
https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php#L378-L382

It was created to allow usage of 3rd party components in DI, keeping the simplicity for the Yii classes instatiation at the same time.

What should be a use case for the interfaces you create with this PR? Where you will check if particular class implements `MutatorsInterface` and for what purpose?

rob006:
```php
public function fillObject(MutatorsIterface $object, array $data) {
	foreach ($data as $key => $value) {
		if ($object->canSetProperty($key)) {
			$object->$key = $value;
		} else {
			$this->handleInvalidProperty($key, $value, $object);
		}
	}
}
```

or 

```php
public function fillObject($object, array $data) {
	if ($object instanceof MutatorsInterface) {
		foreach ($data as $key => $value) {
			if ($object->canSetProperty($key)) {
				$object->$key = $value;
			} else {
				$this->handleInvalidProperty($key, $value, $object);
			}
		}
	} else {
		foreach ($data as $key => $value) {
			$object->$key = $value;
		}
	}
}
```

klimov-paul:
Does not look practicle.
Your example suggests you need to fill up and object, which exact type is completely unknown, with array, which structure is completely unknown. I can not refer to any real world task, which may need this.

The closest will be configuring some component from configuration array in the way Yii does it. But in this case skipping not existing properties is not an option as will hide possible developer mistakes like typos.

If source object type is unknown there should not be checks for `canSetProperty()` on it.  Particular object may not follow your `MutatorsInterface` and still provide a virtual properties in its own way. See your own example you have a nice `else` statement, which does not perform any checks while setting properties, while it may be unsafe.

klimov-paul:
Do you have more real-world-close example for `MutatorsInterface` ?

Also waht about use case for `ComponentInterface` ?

klimov-paul:
One more thing about `MutatorInterface`: it does not cover behavior check trigger introduced at `Component` level.

Compare `Object` ( https://github.com/yiisoft/yii2/blob/master/framework/base/Object.php#L273 ):

```php
public function canSetProperty($name, $checkVars = true) {}
```

with `Component` (https://github.com/yiisoft/yii2/blob/master/framework/base/Component.php#L366) :

```php
public function canSetProperty($name, $checkVars = true, $checkBehaviors = true)
```

Although `Component` methods declaration are compatible with the ones from `Object`, following your logic and your example there should be a separated interface `Mutator+Component`, which should redeclare `canSetProperty()` adding 3rd paratemer to it. Otherwise in your example there is no way of knowing if I should pass 3rd paratemer as `false` to disable behavior check or not.

rob006:
> See your own example: you have a nice else statement after if ($object instanceof MutatorsInterface) {

In PHP you can set undeclared property in simple object, but you can't do this on object that implements `MutatorsInterface`. This implementation gives you two possibilities - you can pass simple object and set all `$data` values as object properties or pass `MutatorsInterface` instance and set only allowed attributes.

> Your example suggests you need to fill up and object, which exact type is completely unknown, with array, which structure is completely unknown. I can not refer to any real world task, which may need this.

You may need this when work with external APIs. Component that handle request to API does not need to know API result structure or required object structure - it only make request and fill result object. 

> Also what about use case for ComponentInterface ?

You need real-world example when you need to check whether object handle events/behaviors?

klimov-paul:
> You may need this when work with external APIs. Component that handle request to API does not need to know API result structure or required object structure - it only make request and fill result object.

Then why do you use class with restricted properties list in this case???
In the case you describe you definetely need to use simple `stdClass` or other class without restrictions on the properties list.

> You need real-world example when you need to check whether object handle events/behaviors?

Yes.
Events and behaviors are object internal matters. I can not imagine a use case, where you need to check whether particular class implements `ComponentInterface` and then add event listener for `BaseActiveRecord::afterInsert` event on it.
Behavior usually tend to work with particular class as they handle particular events which exists in particular class - what is the sence of checking `ComponentInterface`  in thier case?

rob006:
> Then why do you use class with restricted properties list in this case???

For performance reasons. If I don't need whole response data, what is the point of wasting memory for it?

> Events and behaviors are object internal matters. I can not imagine a use case, where you need to check whether particular class implements ComponentInterface and then add event listener for BaseActiveRecord::afterInsert event on it.

Well, if you can't imagine a use case for check whether object handles events or behaviors, I probably can't give you any example that you consider as "real-world example". But you should remember, that this is public interface - you can attach event/behavior and use it outside the object. So you don't need any knowledge about the object itself to use events or behaviors, you can just build it by attaching behaviors.

And I think this discussion is going nowhere. Just made some core-developers internal vote and decide. If you choose to create the trait that completely changes the default behavior of PHP object, and gives no way to specify this change - so be it. 15 years ago there were no interfaces at all and people live with it, I'm sure this is still possible. Same for `EventsInterface` and `BehaviorsInterface` (I will try to separate `Component` logic into two traits). Personally I think they will be useful and don't see any problem that they may create, but will respect your decision.

dynasource:
I scout some feeling of fear for change @klimov-paul and a strong protectionism for old choices. However, I do not think that is entirely appropiate for many reasons. The current implementation:
* is not simple (despite that argument on 'simplicity over complexity'). It is too much entangled
* is not structured by objective. Many logic is pushed into 2 classes. 
* is not optimal and DRY: component is overriding logic of Object without reusing logic
* is already twisted: canSetProperty($name, $checkVars = true) {} vs canSetProperty($name, $checkVars = true, $checkBehaviors = true)
* is confusing. Interfaces are expected but not in place (because of the previous point)
* is not flexible. Developers can not practice composition over inheritance

We should also not be afraid to leave old choices because thats what software evolution is. PHP evolves. Yii should evolve.

klimov-paul:
> Well, if you can't imagine a use case for check whether object handles events or behaviors, I probably can't give you any example that you consider as "real-world example". But you should remember, that this is public interface - you can attach event/behavior and use it outside the object. So you don't need any knowledge about the object itself to use events or behaviors, you can just build it by attaching behaviors.

Here is a practical example of event attachment:

```php
$model = Item::find()->one();
$model->on(BaseActiveRecord::EVENT_AFTER_SAVE, fucntion () {...});
```

If I want to attach some particular event for unkown obkect, the code will look like following:

```php
if ($object instanceof BaseActiveRecord) { // check if object is ActiveRecord, and thus have and 'afterSave' event
    $object->on(BaseActiveRecord::EVENT_AFTER_SAVE, fucntion () {...});
}
```

Now, lets see what you suggest:

```php
if ($object instanceof ComponentInterface) { // check if object can handle events
    $object->on(BaseActiveRecord::EVENT_AFTER_SAVE, fucntion () {...}); // attaching event for ActiveRecord???
}
```

Such code is totally rediculus: it suggests some unknown class can handle `afterSave` event, which exists only in ActiveRecord, in case this class just can handle events.

Here is a practicle usage of behavior attachment:

```php
$model = Item::find()->one();
$model->attachBehavior('timestamp', new TimestampBehavior());
$model->save();
```

If I want to attach some particular event for unkown obkect, the code will look like following:

```php
if ($object instanceof BaseActiveRecord) { // check if object is ActiveRecord, and thus `TimestampBehavior` can be in use of it
    $object->attachBehavior('timestamp', new TimestampBehavior());
}
```

Now, lets see what you suggest:

```php
if ($object instanceof ComponentInterface) { // check if object can attach behavior
    $object->attachBehavior('timestamp', new TimestampBehavior()); // attach behavior that modifies internal field before saving ActiveRecord???
}
```

Such code is totally rediculus: it suggests some unknown class can attach `TimestampBehavior` just because it can attach some behavior. What if you will receive `UrlManager` instance here - you will attach `TimestampBehavior` for it as well?

Attaching events or behaviors requires a check for much more particular object type, which makes checking of the `ComponentInteface` completely redundant.

Creating dynamic-functionality object using behavior attachments is not common practice - it makes a little sense. It is hard to imagine some particular behavior, which should be general enough to be attached to any component without knowing of its internal structure.

As I already said there could be some hypothetical use case, which justifies all the changes introduced by this PR. As well as I can found some hypothetical use case, which demands method `canSetProperty()` and `canGetProperty()` should belong to 2 seprate interfaces instead of being united under the single one. Why not? It can be a situation where you need to check only if object can set virtual properties, while at other place - whether it can get them.

At the present state I can see no practical reason, which justifies code overcomplication.

klimov-paul:
> is not simple (despite that argument on 'simplicity over complexity'). It is too much entangled
is not structured by objective. Many logic is pushed into 2 classes.

I can not see how existance of 3 traits and 2 interface in addition(!) to the 2 already esitsing clases simpifies anything.

> is not optimal and DRY: component is overriding logic of Object without reusing logic

This PR does nothing about it: `Component` still extends `Object`, `ComponentInterface` extends `MutatorsInterface`. What has changed?

> is already twisted: canSetProperty($name, $checkVars = true) {} vs canSetProperty($name, $checkVars = true, $checkBehaviors = true)

There is no choice: `Component` should be able to check for virtual properties with or without behavior to be scout. This cause method signature change and code duplication. This PR does nothing to solve it: implementation still varies at `Object` and `Component` level.

> is confusing. Interfaces are expected but not in place (because of the previous point)

I do not understand this part: which interface is expected where exactly?

> is not flexible. Developers can not practice composition over inheritance

Developer has his choice noone forces him. Still until now there was no issues on this matter. And you unable to provide any practical use case - just somewhere someone may need it at somepoint.

If you able to provide common practicle use case, which demands such change - I will reconsider, but so far I can not see any.

> We should also not be afraid to leave old choices because thats what software evolution is. PHP evolves. Yii should evolve.

The entire issue with `Object` and `Component` existence is lack of PHP native functionality.
For example if PHP can provide extra magic method named for example `__canGetproperty()`, we would be able to use it and seize all debates arout `canGetPropery()`, same for `__hasMethod()` and so on.
Or if PHP can provide a native way to declare properties by setter/getter like C# and other languages do. 
When PHP will actully evolve in this matter - we will consider this evolution - but now there is nothing to be considered.

Anyway, as I have spoken several times before: if the rest of the core team suggests this PR should be taken - I can not stop them. I am not a God - I am just a human, and thus I can not be sure my opinion is correct. I have spoken my arguments - if they are not a concern for you - so be it.

samdark:
Thanks for explanation @klimov-paul. I think this is a central point of your critics:

> This PR does nothing about it: Component still extends Object, ComponentInterface extends MutatorsInterface. What has changed?

In 2.0 nothing is changed in the framework itself. Classes are still there because we need backwards compatibility. In the user code though, one will be able to use traits separately. There are two gains in that:

1. It won't require to inherit everything from Object so user may keep inheritance tree he wants while getting properties and events in one of a child classes.
2. Each trait could be used separately if one doesn't need both getters-setters and behaviors.

dynasource:
> I have spoken my arguments - if they are not a concern for you - so be it.

your arguments are important. Your vision & effort is highly appreciated.

> if the rest of the core team suggests this PR should be taken

this PR is a concept. If its fixing the criteria of https://github.com/yiisoft/yii2/pull/12929#issuecomment-259681283, I will be positive, otherwise I won't be.

klimov-paul:
> Classes are still there because we need backwards compatibility. In the user code though, one will be able to use traits separately. 

Are you serious? I have spoken on this here alreay: https://github.com/yiisoft/yii2/pull/12929#issuecomment-259119886

Are you telling we will compose each(!) core internal class in this way:

```
class Foo implements \yii\base\Configurable, \yii\base\MutatorsInterface
{
    use \yii\base\ConfigurableTrait;
    use \yii\base\MutatorsTrait;
{
}
```
???

> It won't require to inherit everything from Object so user may keep inheritance tree he wants while getting properties and events in one of a child classes.

No one forces you to inherit anything from `Object`!
You can use any native class inside Yii and with its DI.

> if one doesn't need both getters-setters and behaviors.

Same here.

dynasource:
> Are you telling we will compose each(!) core internal class in this way:

Nothing will change for core classes. These will still extend from Component / Object

samdark:
@klimov-paul no. We won't compose each core class using traits. That's more for user classes which can't inherit from Object but need properties and behaviors at the same time.

klimov-paul:
> We won't compose each core class using traits. That's more for user classes which can't inherit from Object but need properties and behaviors at the same time.

Then what do you mean by telling:

> Classes are still there because we need backwards compatibility.

?

framework/base/ComponentTrait.php
```diff
@@ -0,0 +1,595 @@
...
+                    $behavior->$name = $value;
+
+                    return;
+                }
+            }
+        }
+        if (method_exists($this, 'get' . $name)) {
+            throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
+        } else {
+            throw new UnknownPropertyException('Setting unknown property: ' . get_class($this) . '::' . $name);
```
rob006: You could override `__set()`  in target object and catch these exceptions  or create behavior with custom `__set()` and `canSetProperty()` which accepts everything.

samdark:
I've meant that that there will be no changes about core classes in 2.0. In 2.1 for some classes it potentially could make sense to remove all properties and either leave constructor only or create setters without defining properties. In many cases behaviors aren't meant to be attached as well. For these cases one may either not use traits at all or use one of them only.

framework/base/ComponentTrait.php
rob006: Overriding 2 methods is still a lot less work than implementing own behaviors/events logic. Current behavior is Yii convention which has some advantages (see https://github.com/yiisoft/yii2/issues/12021#issuecomment-236886911). If you you disagree with this, you can always change this pretty easy by implementing own traits with specified behavior.

rob006:
This is more problematic than I expected. Due some limitations in handling traits and interfaces in PHP 5 (sometimes PHP 5 does not take into account used traits when checking interface implementation, see https://3v4l.org/JWXqX ), we have 3 options:

1. Do no implement `BehaviorsInterface` in `Component` (or don't use interfaces at all, like @klimov-paul requested) 
   https://github.com/rob006/yii2-dev/blob/aba6d8b9e83597565efec4023740eed8e2ad787c/framework/base/Component.php#L101 => https://travis-ci.org/rob006/yii2-dev/builds/175317588

2. Use some ugly workaround in `Component` 
   https://github.com/rob006/yii2-dev/blob/f8ef6ae8c35145f013eb9af03ed8180a78dd84df/framework/base/Component.php#L101 => https://travis-ci.org/rob006/yii2-dev/builds/175317203

3. Do not extend `Component` from `Object` (breaks BC)
   https://github.com/rob006/yii2-dev/blob/955019c90bf95b6a0ff9d6d09260ef1afc7661d5/framework/base/Component.php#L101 => https://travis-ci.org/rob006/yii2-dev/builds/175314851

samdark:
I'd like to have it in 2.1.

dynasource:
me too

cebe:
I don't like splitting the base classes of all classes in Yii into multiple traits, this will blow up the minimum overhead added by the framework, as instead of loading two classes (Object and Component) it would also need to include a set of traits and interface.

sergeymakinen:
I don't think that 2 traits and an interface is a measurable issue in terms of performance. And even 10 traits and interfaces. Especially in a PHP 7 era.

cebe:
Its not only performance its more about general complexity of the base. Yii philosophy is to keep things simple and practical. Separating the base class into multiple traits adds complexity in understanding it.
To understand what is going on in Yii you'd have to look into multiple classes, traits and interfaces and put all that together to get the idea of what you are building on, while currently you have two classes and thats it.

What exactly are the practical benefits of such a separation?

sergeymakinen:
I was about a performance. My thoughts about a PR usefulness are still controversial.

I love Yii's KISS. But I find `VirtualPropertiesTrait` useful. Because it provides things are common without a need to inherit `BaseObject`. `EventsInterface` and its trait are also handy when you need to accept an object implementing this thing without inheriting. TBH I would implement these changes and that's it.

On the other hand, as far as I understand the whole thing is about breaking inheritance chains. From this point I think this PR is a solid step forward.

samdark:
Yes, it's about a step towards removing inheritance chain that's way too long.

rob006:
> What exactly are the practical benefits of such a separation?

1. Extracting this logic into traits will allow to easily use events/behaviors/virtual attributes in third party classes.
2. Separating events and behaviors may increase performance in some cases - if you need only events, you can skip behaviors overhead by using only `EventsTrait`.
3. At least for me splitting `Component` into smaller traits focused on single feature, make it easier to read and understand.

dynasource:
Perfectly summarized

cebe:
> Yes, it's about a step towards removing inheritance chain that's way too long.

in current state its making inheritance chain more complex.

> Extracting this logic into traits will allow to easily use events/behaviors/virtual attributes in third party classes.

what kind of third party are you thinking of here? Which package would require `yiisoft/yii2` just to use these traits?



> Separating events and behaviors may increase performance in some cases - if you need only events, you can skip behaviors overhead by using only EventsTrait.

the overhead is marginal and having to decide every time you create a class does not really safe anything, it takes time to make a decision whether to include which kind of trait, and it may limit usage if you later decide to add some behavior. For extensions that means if an extension excluded behaviors you can not inject your functionality and start a discussion with the author instead of using existing functionality that is always there if needed. Having behaviors available by default makes Yii perfectly extensible in many ways by default. I have experienced this when working on HumHub, which is very flexible when writing extensions for it, behaviors allow you to inject a lot of functionality. I'd keep this in by default, arguing about performance here would result in the question whether you'd want to use a framework at all.

rob006:
> what kind of third party are you thinking of here? Which package would require yiisoft/yii2 just to use these traits?

Any yii2 extension that provides integration with any existing non-yii2 library. Instead just wrapping everything you can extend library class, use these traits and interfaces and you will have yii2-like component. 

> the overhead is marginal and having to decide every time you create a class does not really safe anything, it takes time to make a decision whether to include which kind of trait, and it may limit usage if you later decide to add some behavior.

1. You already need to make decision whether to use `Component`, `BaseObject` or nothing as base for your class.
2. `Component` will still provide behaviors and events, and creating events-only class will be more complicated than just `extend Component`, so it will be probably used only if someone known what is doing and really need it. I don't think that framework should force using behaviors in this way.

rob006:
I'm closing this PR for now since after https://github.com/yiisoft/yii2/pull/14478 it needs to be redone from scratch. Please continue discussion in https://github.com/yiisoft/yii2/issues/9748 and let me know about your decision - I'm interested in to finish this in new PR if the idea gets a green light.

Issue #12840, opened 2016-10-26
Have at least basic tests for js


arogachev:
Extracted from #11037.

TODO list (added by @cebe):

- [x] implement basic test setup with example test suite for `yii.validation.js` (#12936)
- [x] make tests run on travis
- [x] add tests for yii.activeForm.js
- [x] add tests for yii.captcha.js (#13160)
- [x] add tests for yii.gridView.js (#13230)
- [x] add tests for yii.js (#13316)
- [x] add tests for yii.validation.js (#12936)

dynasource:
yes, this should be implemented. A few months ago, I created a local testsuite based on Mocha & Node.js for speed (instead of phantom). Its not easy to setup, but its possible. Perhabs also possible for githubs Jenkins/Travis?

cebe:
If someone wants to work on this, that would be awesome!

dynasource:
it would be nice to discuss how to implement this properly. Questions to be answered:
* should we use nodejs + jsdom or phantom?
* should we use Mocha?
* should we start with ES5  | Typescript | javascript

If you'd ask me:
* Lets use nodejs + jsdom because of high speed & good support & less error prone thant phantom
* Lets use Mocha (seems to be the most popular at the moment)
* Lets ES5 and transpile at every change into source code (required to be done by the developer)

We could make use of the ES5 'class' system, which makes the code way easy to understand and to split in self contained units, making testing also easier. Of course this is tricky. At every change of the code, the end result has to be transpiled into the source code. Also, the mapping files have to be generated so viewing in Chrome is possible.

One can argue that it is too tricky to do, but on the contrary, if you look at how many bugs we have with JS in our issue tracker, a modular design is really demanded.

arogachev:
@dynasource Good news - I started working on this about week or more ago. You can set me as assignee. Currently I wrote full tests suite for `yii.validation.js`. It's based on Mocha, Chai, Sinon and Leche. Will open PR soon.

As for `yii.validation.js` - there are almost no DOM manipulations. Currently the tests are adapted to run only in browser, but I'm going to swap it with Node / Phantom.

To answer your questions:

> should we use nodejs + jsdom or phantom?

Not sure about that at the moment.

> should we use Mocha?

I think we should. I investigated this problem before writing tests, have read some articles. Mocha seems to be good choice, but QUnit and Jasmine are reliable too. Personally I'm for Mocha.

> should we start with ES5 | Typescript | javascript

[...]

PR #12936, opened 2016-11-08
#12840: Added unit tests for yii.validation.js

arogachev:
| Q             | A
| ------------- | ---
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12840

Used @nkovacs draft as starting point(https://github.com/yiisoft/yii2/issues/11037#issuecomment-193356450). I also fixed some bugs in `yii.validation.js` and did some refactoring.

## Implemented tests

- yii.validation.js

## Running tests

- JS tests use jQuery and Punycode.js libraries from `vendor/bower` to match those that are used in assets and prevent installing and synchronizing duplicates via NPM. So make sure you have them installed. If no - run `composer install` to install them via Composer Asset Plugin.
- Install Node.js. I used the latest version available at the moment - `v7.1.0`. Note that used `jsdom` dependent library version requires `Node.js >= 4`.
- Run `npm install` to install dependencies.
- Execute `npm test` for running tests.

## Problems list

- [x] Sinon can not be installed via Composer Asset Plugin. If I add line `"npm-asset/sinon": "^1.17.6"` to `required-dev` section of `composer.json` and run `composer update` I get the following error:

```
[RuntimeException]
Failed to execute git clone --mirror 'git+ssh://git@github.com/sinonjs/sinon.git' '/home/vagrant/.composer/cache/vcs/git-ssh---git-github.com-sinonjs-sinon.git/'
Cloning into bare repository '/home/vagrant/.composer/cache/vcs/git-ssh---git-github.com-sinonjs-sinon.git'...
Permission denied (publickey).
fatal: Could not read from remote repository.
Please make sure you have the correct access rights and the repository exists.
```

Possibly related: https://github.com/fxpio/composer-asset-plugin/issues/187.

I tried `--prefer-dist` option, that didn't help. I also have generated SSH key and it's added it to `ssh-agent` (error is mentioned in [git-workflow](https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md#1-fork-the-yii-repository-on-github-and-clone-your-fork-to-your-development-environment) document which is a part of internal docs).

This is not actual anymore because Node.js and NPM are used for testing.

- [x] Sometimes "image validator with FileReader API is not available" test takes a long time to load and gives warning with time written inside of a red label. This is because of swapping native `FileReader`:

```js
var initialFileReader = FileReader;

beforeEach(function () {
    FileReader = undefined;
});

afterEach(function () {
    FileReader = initialFileReader;
});
```

But:

- It happens pretty seldom and increases time by about half of second only.
- I can't see a better solution for testing this, so this is not a big problem.
- Similar technique is used for swapping native JS components in other places.

So this is not a big deal.

This is not actual anymore because Node.js and NPM are used for testing and for testing that case different approach is used.

- [x] Run tests in command line instead of a browser.

Node.js and NPM are used for testing.

- [x] Error while running `npm test`:

> Error: timeout of 15000ms exceeded. Ensure the done() callback is being called in this test.

Happens very seldom, this is because of jsdom and integrating jQuery. Default timeout is 2000ms. And this is always not enough, so I increased it to be 15000ms.

Solved with increasing timeout to 30s, see #13160.

- [x] In build detailed info the line after "with width more than minimum width" (validateImage method, with successfully read image) is not printed. It's OK locally though.

Can't see it in some builds, maybe it was Travis glitch related with formatting.

It's gone in the latest builds, for example in [this one](https://travis-ci.org/yiisoft/yii2/jobs/182190268).

## Integration with Travis CI

I found this [article](https://ma27.github.io/travis-ci/2016/05/06/travis-for-nodejs-and-php.html) and used it as starting point.

> TravisCI uses the software which is shipped by default on a `precise`: An ancient setup with NodeJS 0.10 and NPM 1.x. The current 3.x series of NPM is much more stable than the ancient NPM 1.x and fixed a lot of misbehavior which has been implemented in the older versions of NPM. Especially code which has been written for the browser will experience issues due to some implementation details of NPM releases prior to the 3.x series.

We can't use Node.js < 4 because of [jsdom](https://github.com/tmpvar/jsdom) dependency. I refused to use [3.x branch](https://github.com/tmpvar/jsdom/tree/3.x) of jsdom because it's considered legacy and no longer maintained.

I tried to change dist to `trusty` and install newer versions of Node.js and NPM. It worked, but another error appeared during creating test MySQL database:

```
- mysql -e 'CREATE DATABASE yiitest;';
```

Error's text:

> ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)

Then I found https://github.com/travis-ci/travis-ci/issues/6842 and fixed it with installation of MySQL 5.6 (default is MySQL 5.5 for trusty dist).

Hovewer new problems appeared.

- [x] On PHP 5.4:

> Warning: PHP Startup: Unable to load dynamic library
> '/home/travis/.phpenv/versions/5.4.45/lib/php/extensions/no-debug-zts-20100525/apc.so' -
> /home/travis/.phpenv/versions/5.4.45/lib/php/extensions/no-debug-zts-20100525/apc.so:
> cannot open shared object file: 
> No such file or directory in Unknown on line 0

Related: https://github.com/travis-ci/travis-ci/issues/6853

I'm trying to fix it by installing PECL package, see this [article](http://olvlvl.com/2016-09-travis-apc). Hovewer no results yet, because it tries to install it for PHP 7 version (that's weird because version in the build is 5.4).

Fixed.

- [x] On HHVM 2 tests related with PostgreSQL are now failing.

Specified 3.6 version for testing (the same as on Precise), created separate issues - #12981, #12982.

- [x] Get rid of oauth token, #12987.


framework/assets/yii.validation.js
```diff
@@ -13,7 +13,7 @@
 yii.validation = (function ($) {
     var pub = {
         isEmpty: function (value) {
-            return value === null || value === undefined || value == [] || value === '';
+            return value === null || value === undefined || ($.isArray(value) && value.length === 0) || value === '';
```
arogachev: Fixed bug. 

In JavaScript array can not be determined just by using this kind of check: `value == []`, `[] == []` is always `false` for example. Since the code is jQuery dependent it's OK to use `$.isArray` (and by the way it's used above in the same file). Another advantage of using `$.isArray` over [Array.isArray()](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) is that 1st version branch has fallback for older browsers, so in case user want that he need just override jquery version in `yii\web\JQueryAsset`.

framework/assets/yii.validation.js
```diff
@@ -58,15 +58,17 @@ yii.validation = (function ($) {
                 return;
             }
 
+            if (options.is !== undefined && value.length != options.is) {
```
arogachev: `is` conflicts with `min` and `max` and it does not make sense to use these options together. We either check for exact value or `min` or `max`. So it's better to skip `min` and `max` checks if `is` option is specified and check it first.

framework/assets/yii.validation.js
```diff
@@ -76,9 +77,8 @@ yii.validation = (function ($) {
             });
         },
 
-        image: function (attribute, messages, options, deferred) {
+        image: function (attribute, messages, options, deferredList) {
             var files = getUploadedFiles(attribute, messages, options);
-
```
arogachev: Inconsistency. There is no newline separator in similar code in file validator.

framework/assets/yii.validation.js
```diff
@@ -76,9 +77,8 @@ yii.validation = (function ($) {
             });
         },
 
-        image: function (attribute, messages, options, deferred) {
+        image: function (attribute, messages, options, deferredList) {
```
arogachev: It's now clear that this is an array.

framework/assets/yii.validation.js
```diff
@@ -87,48 +87,33 @@ yii.validation = (function ($) {
...
-                        messages.push(options.overHeight.replace(/\{file\}/g, file.name));
-                    }
-                    def.resolve();
-                };
-
-                img.onerror = function () {
-                    messages.push(options.notImage.replace(/\{file\}/g, file.name));
-                    def.resolve();
-                };
+                var deferred = $.Deferred();
```
arogachev: Removed reduction. Moved `validateImage` to separate method because it's impossible to test this logic otherwise - local variables exist only in their scope and we can't redefine native event handlers. This for example:

```js
FileReader.prototype.onload = undefined;
```

gives:

```
TypeError: 'set onload' called on an object that does not implement interface FileReader.
```

I tried my best to think a workaround, but seems like moving it to separate method is the best solution to avoid ugly hacks.

framework/assets/yii.validation.js
```diff
@@ -87,48 +87,33 @@ yii.validation = (function ($) {
...
+                pub.validateImage(file, messages, options, deferred, new FileReader(), new Image());
+                deferredList.push(deferred);
+            });
+        },
 
-                fr.onload = function () {
-                    img.src = fr.result;
-                };
+        validateImage: function (file, messages, options, deferred, fileReader, image) {
```
arogachev: Removed variable names reductions.

framework/assets/yii.validation.js
```diff
@@ -87,48 +87,33 @@ yii.validation = (function ($) {
...
-                    def.resolve();
-                };
+            image.onerror = function () {
+                messages.push(options.notImage.replace(/\{file\}/g, file.name));
+                deferred.resolve();
+            };
 
-                fr.readAsDataURL(file);
+            fileReader.onload = function () {
+                image.src = this.result;
```
arogachev: We can use `this,result` instead in the scope of event handler.

framework/assets/yii.validation.js
```diff
@@ -170,6 +155,10 @@ yii.validation = (function ($) {
                 }
             });
 
+            if (options.not === undefined) {
+                options.not = false;
```
arogachev: We expect client validators to behave the same way as according server validators. By default `not` is `false`, but if it was set to undefined, we get wrong result.

framework/assets/yii.validation.js
```diff
@@ -190,25 +179,26 @@ yii.validation = (function ($) {
                 return;
             }
 
-            var valid = true;
-
```
arogachev: Extra newlines, variables can be declared together.

framework/assets/yii.validation.js
```diff
@@ -190,25 +179,26 @@ yii.validation = (function ($) {
...
-
-            var regexp = /^((?:"?([^"]*)"?\s)?)(?:\s+)?(?:(<?)((.+)@([^>]+))(>?))$/,
+            var valid = true,
+                regexp = /^((?:"?([^"]*)"?\s)?)(?:\s+)?(?:(<?)((.+)@([^>]+))(>?))$/,
                 matches = regexp.exec(value);
 
             if (matches === null) {
-                valid = false
+                valid = false;
```
arogachev: Code style. Missed semicolon.

framework/assets/yii.validation.js
```diff
@@ -190,25 +179,26 @@ yii.validation = (function ($) {
...
             } else {
+                var localPart = matches[5],
```
arogachev: It's better to use separate named variables for local-part and domain because they are used more than once.

framework/assets/yii.validation.js
```diff
@@ -262,11 +252,7 @@ yii.validation = (function ($) {
 
             // CAPTCHA may be updated via AJAX and the updated hash is stored in body data
             var hash = $('body').data(options.hashKey);
-            if (hash == null) {
```
arogachev: Code style. Can be refactored to one liner, does not exceed max string length.

framework/assets/yii.validation.js
```diff
@@ -281,7 +267,8 @@ yii.validation = (function ($) {
                 return;
             }
 
-            var compareValue, valid = true;
```
arogachev: Code style. Use the same way of declaring multiple variables at once.

framework/assets/yii.validation.js
```diff
@@ -328,17 +315,13 @@ yii.validation = (function ($) {
         },
 
         ip: function (value, messages, options) {
-            var getIpVersion = function (value) {
```
arogachev: Separate function is useless here because it's used only once and it's just one short line.

framework/assets/yii.validation.js
```diff
@@ -328,17 +315,13 @@ yii.validation = (function ($) {
...
-            };
-
-            var negation = null, cidr = null;
-
             if (options.skipOnEmpty && pub.isEmpty(value)) {
                 return;
             }
 
-            var matches = new RegExp(options.ipParsePattern).exec(value);
+            var negation = null,
```
arogachev: Better grouping of variables.

framework/assets/yii.validation.js
```diff
@@ -358,7 +341,8 @@ yii.validation = (function ($) {
                 return;
             }
 
-            if (getIpVersion(value) == 6) {
+            var ipVersion = value.indexOf(':') === -1 ? 4 : 6;
```
arogachev: Separate function is useless (see above).

framework/assets/yii.validation.js
```diff
@@ -405,15 +389,8 @@ yii.validation = (function ($) {
 
     function validateFile(file, messages, options) {
         if (options.extensions && options.extensions.length > 0) {
-            var index, ext;
-
-            index = file.name.lastIndexOf('.');
```
arogachev: What's the point of declaring `index` as `undefined` initially if then immediately assign some value to it?

framework/assets/yii.validation.js
```diff
@@ -405,15 +389,8 @@ yii.validation = (function ($) {
...
-
-            if (!~index) {
```
arogachev: Good candidate for one liner.

framework/assets/yii.validation.js
dynasource: agreed

framework/assets/yii.validation.js
dynasource: agreed

framework/assets/yii.validation.js
dynasource: agreed

framework/assets/yii.validation.js
dynasource: agreed

dynasource:
your PR is impressive. You've put a lot of effort in. Good job.

A critical remark though ;). Wasnt it possible to create the tests first and refactor yii.validation.js after? This would improve the trust on the refactorings. On the contrary, I can well imagine the requirement to modularize the JS first.

Regarding the other js file refactorings. Lets fix these 1 per PR (easier to review & revert if necessary)

dynasource (approved):
your PR is impressive. You've put a lot of effort in. Good job.

A critical remark though ;). Wasnt it possible to create the tests first and refactor yii.validation.js after? This would improve the trust on the refactorings. On the contrary, I can well imagine the requirement to modularize the JS first.

Regarding the other js file refactorings. Lets fix these 1 per PR (easier to review & revert if necessary)

dynasource:
@arogachev,@cebe, it is already possible to execute the tests in Travis?

arogachev:
@dynasource Thanks. Yes, refactoring can be done separately, but there are few bugs related to tests. Yes, I think covering one js file with tests per PR is better for review.

framework/assets/yii.validation.js
```diff
@@ -36,7 +36,7 @@ yii.validation = (function ($) {
             }
         },
 
-        'boolean': function (value, messages, options) {
+        boolean: function (value, messages, options) {
```
nkovacs: This is because in es3 boolean is a keyword, and could not be used without quotes.
It's ok in es5.

arogachev: @nkovacs Got it. Thanks for explanation. Should we leave it as is in this case to support ES3?

dynasource: @samdark?

nkovacs: Not IE9, IE8 and older.

arogachev: Got it. I'll return it as it was. I'll also add additional explaining comment about it in this case.

tests/js/assets/js/yii.validation.test.js
```diff
@@ -0,0 +1,1598 @@
+describe('yii.validation', function () {
+    var VALIDATOR_SUCCESS_MESSAGE = 'should leave messages as is';
+    var VALIDATOR_ERROR_MESSAGE = 'should extend messages with according error(s)';
```
arogachev: I think my variation is more specific. With this message it's unclear what kind of result we actually expect from given data set.

nkovacs: Why is it unclear?

The validator adds a message to the messages array using [Array.prototype.push](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/push). The documentation for that method uses similar wording ("**adds** one or more elements to the end of an array"), and it's simpler and clearer than "extend", which makes me think of [jQuery.extend](https://api.jquery.com/jquery.extend/).

Secondly, "according error" sounds very unnatural and weird (it might not be grammatically incorrect, since according can be an adjective, but it's not the right word to use here). "Appropriate error" or "relevant error" or "correct error" is much better and clearer.

arogachev: @nkovacs. Sorry, I thought you are suggesting to use one generic message instead of two. Yes, I can modify message to be more precise.

nkovacs: Ah, I see. Thanks!

cebe:
Thanks for your effort!

> Sinon can not be installed via Composer Asset Plugin. If I add line "npm-asset/sinon": "^1.17.6" to required-dev section of composer.json and run composer update I get the following error:

adding it to composer.json worked fine for me. Seems like you have not set up your SSH public key in the github settings.

> @arogachev,@cebe, it is already possible to execute the tests in Travis?

I guess since it is running in a browser right now, we can not easily run this on travis directly.
I am for getting this PR ready first and then tackle the problem of automating the test run on travis.

I'd also recommend to limit the scope of this PR to yii.validation.js and add more tests in separate PRs.

cebe:
I have edited the issue on https://github.com/yiisoft/yii2/issues/12840 to include a list of separete steps that should be made in separate PRs.

cebe:
@arogachev I have sent you a pull request: https://github.com/arogachev/yii2/pull/1 if you merge it that commit should be inluded here.

arogachev:
Implemented execution of tests using Node.js and NPM. PHP test app is not needed anymore.

nkovacs:
There are two reasons I used php to run the tests and not node.
I wanted to avoid requiring node for Yii development, and that setup allowed me to use php code to generate the client options required by the js code (though that does technically make them integration tests).

dynasource:
how do you test JS with php? Dont you still need a JS executioner like Phantom or Selenium to do the magic? If thats the case, it really doesnt matter to use Nodejs as an alternative, right?

About that fixture preparation. This can still be done? You run your PHP file to create the required JS fixtures?

nkovacs:
You're right, that is a problem for Yii, because you don't have a php server in your CI setup.
We already had a running apache server with php in our CI setup (for acceptance tests) and we used [gulp-mocha-phantomjs](https://www.npmjs.com/package/gulp-mocha-phantomjs) (because we also already had a gulp setup).
Also, when you're working on something, you can just open the test page in a browser. That also gives you easy access to the browser's development tools and a nicer output.

arogachev:
I think JS unit tests should not be dependent on PHP in any way.

As for `clientValidateAttribute` - we can test that correct JS code part is generated on PHP side. Also additionally we can have some integration tests with simple app, but this is out of scope of this task in my opinion.

As for Node.js vs PhantomJS for running tests - there is no exact answer which one to use or which one is better. Both have pros and cons. I decided to use Node.js.

Running tests in the browser with existing Node.js / NPM implementation doesn't make much sense. In case of errors you can see the place where it's coming from in stack trace. Also you can perform some debugging using same `console.log()` and `process.exit()` if needed. Full output can be found in detailed Travis CI build info (you push your code to development branch anyway). You can scroll and investigate it in details.

I don't think we should support both ways of executing tests (browser and command line), in case of PhantomJS it probably would be easier, but with Node.js you can't load testable code in usual manner.

.travis.yml
```diff
@@ -32,6 +34,11 @@ cache:
 
 # try running against postgres 9.3
 addons:
+  apt:
+    packages:
+      - mysql-server-5.6
+      - mysql-client-core-5.6
+      - mysql-client-5.6
```
arogachev: @cebe It's related. I updated PR's first post. Please see "Integration with Travis CI" section.

nkovacs:
Using a browser to run the tests gives you access to much better debugging tools than a simple `console.log` and the stack trace.
It also allows you to test and debug the code in different browsers and you can even set up CI with different browsers, e.g. with browserstack.

.travis.yml
```diff
@@ -1,3 +1,5 @@
+dist: trusty
```
arogachev: Switch from `precise` to `trusty` is required mainly because of `jsdom` dependency. It requires Node.js > 4. Precise has very old Node.js and NPM versions and they can not be updated because of default versions of C++ compilers. While it's possible somehow to install Node 6 on Precise, see this [document](https://github.com/nodesource/distributions/blob/master/OLDER_DISTROS.md), I think this is not elegant solution and I'm not 100% sure it will work correctly with Travis and container-based precise dist.

Trusty has v6.3.0 which is enough and way better and newer.

.travis.yml
```diff
@@ -7,7 +9,7 @@ php:
   - 7.0
   - 7.1
   - nightly
-  - hhvm
+  - hhvm-3.6
```
arogachev: Default version of HHVM on Precise is 3.6.6, on Trusty - 3.15.2, see https://github.com/yiisoft/yii2/pull/12971#issuecomment-260134653. That causes failing of 2 tests(#12981, #12982). By specifying exact version we keep the same version as on Precise until these 2 bugs will be fixed. In this case tests are passing without failures.

.travis.yml
```diff
@@ -54,9 +65,13 @@ before_script:
   - php -r "echo INTL_ICU_DATA_VERSION . \"\n\";"
   - mysql --version
   - psql --version
+  - node --version
```
arogachev: Additionally check Node.js and NPM versions.

.travis.yml
```diff
@@ -33,19 +35,28 @@ cache:
...
+  - composer config -g github-oauth.github.com 7cecf3a16747d8cc66e21b459aaa233356e95dde
   - travis_retry composer self-update && composer --version
   - travis_retry composer global require "fxp/composer-asset-plugin:^1.2.0"
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
 # core framework:
   - travis_retry composer update --prefer-dist --no-interaction
   - tests/data/travis/apc-setup.sh
   - tests/data/travis/memcache-setup.sh
+# NPM packages
+  - travis_retry npm install
```
arogachev: Install npm dev dependencies used for testing.

.travis.yml
```diff
@@ -71,6 +86,8 @@ script:
   - cd framework && composer validate --no-check-lock && cd ..
   # run PHPUnit
   - vendor/bin/phpunit --verbose $PHPUNIT_FLAGS --exclude-group mssql,oci,wincache,xcache,zenddata,cubrid
+  # run Mocha
+  - npm test
```
arogachev: Run the actual JS unit tests.

.travis.yml
```diff
@@ -54,9 +65,13 @@ before_script:
   - php -r "echo INTL_ICU_DATA_VERSION . \"\n\";"
   - mysql --version
   - psql --version
+  - node --version
+  - npm --version
 
   # initialize databases
   - mysql -e 'CREATE DATABASE yiitest;';
+  - mysql -e "CREATE USER 'yiitest'@'localhost' IDENTIFIED BY 'yiitest'";
```
arogachev: Using `root` user fails:

```
yii\db\Exception: SQLSTATE[HY000] [1045] Access denied for user 'root'@'localhost' (using password: NO)
```
Using `travis` user fails too:

```
yii\db\Exception: SQLSTATE[HY000] [1045] Access denied for user 'travis'@'localhost' (using password: NO)
```

See also #12971.

So I decided to create separate dedicated user for test.

Also note that trusty by default uses MySQL 5.6, while precise - 5.5.

.travis.yml
```diff
@@ -33,19 +35,28 @@ cache:
...
+      yes '' | pecl install apc
+    fi
+
 install:
   - |
     if [ $TRAVIS_PHP_VERSION != '5.6' ]; then
         # disable xdebug for performance reasons when code coverage is not needed
         phpenv config-rm xdebug.ini || echo "xdebug is not installed"
     fi
+  - composer config -g github-oauth.github.com 7cecf3a16747d8cc66e21b459aaa233356e95dde
```
arogachev: This is a temporary workaround, see #12987.

arogachev:
Please review and give some advices on solving #12987.

.travis.yml
```diff
@@ -7,14 +9,17 @@ php:
...
 matrix:
   allow_failures:
     - php: nightly
 
 env:
-  - CUBRID_VERSION=9.3.0/CUBRID-9.3.0.0206 CUBRID_PDO_VERSION=9.3.0.0001
+  global:
+    - CUBRID_VERSION=9.3.0/CUBRID-9.3.0.0206
+    - CUBRID_PDO_VERSION=9.3.0.0001
+    - secure: Q3dD9eL9yGDbMbdnNeLPSruRmZH515tLOTBam5rnsqX8spE+/HLqskLwUKIX78ansVdhu/OLGO72rBHvUotrQPLjqp6GnOSXw15gaz1LcxzfqbiZSbNvYw+7nl0hUjysLEMfZFEGWqrrk8BMuRAWyxABCrdR/91CyTFJhhcTnpg=
```
arogachev: This is an encrypted token from my account. I guess it's better to replace it with token of someone from core developers.

.travis.yml
```diff
@@ -33,19 +35,28 @@ cache:
 addons:
   postgresql: "9.3"
 
+before_install:
+  - |
+    if [ $TRAVIS_PHP_VERSION = '5.4' ]; then
```
arogachev: This needs to be added:

```yaml
before_install:
  - composer config -g github-oauth.github.com "$GITHUB_TOKEN"
```

See https://github.com/yiisoft/yii2/issues/12987#issuecomment-260254948.

arogachev:
#12987 is currently blocking this from successfully passing tests. Without it overall tests pass without failing.

arogachev:
@nkovacs While I agree with you that testing with browser / different browsers has some advantages, here is my opinion.

> Using a browser to run the tests gives you access to much better debugging tools than a simple console.log and the stack trace.

How often this is actually needed? For existing functionality I think `console.log` and stack trace are enough. For something complex or new functionality we can generate test application. But we can add support for running tests via both command line and browser. Need others opinions.

> It also allows you to test and debug the code in different browsers and you can even set up CI with different browsers, e.g. with browserstack.

I checked Browserstack, It seems to be a good tool, but:

- Yii2 is not a JS framework.
- It has relatively small amount of JS code.
- JS code uses ES5 and some external libraries like JQuery, pjax, Punycode.js which are already tested.
- In unit tests stubs are used for testing some code, for example related with File API, jQuery, AJAX. To simulate real browser behavior we need integration tests which are good to have but unit tests have higher priority in my opinion.
- Browserstack is a commercial service. Even it's free for open source, as I understood, you need to contact them with details about your project before using. We don't know their conditions, maybe he will refuse, because of specifics of Yii2 code - and it's coupled with PHP framework and can't be used outside of it.
- We need to start with something and move from dead point. We can have this setup and consider using this later. I tried to not pollute whole file with Node.js integration.

nkovacs:
I don't think it'll be that difficult to set up both browser and nodejs testing, since mocha supports both. I definitely think you should have a proper in-browser CI setup (not necessarily browserstack). It doesn't matter that you have very little JS code, one small thing like the quoted boolean above would make the entire code break in old IE for example (and you don't need integration tests to catch that).

But yeah, I agree, it's lower priority.

package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "yii2",
+  "version": "2.0.10",
```
cebe: is this "version" key needed? I'd remove it to avoid the need of maintaining it.

cebe:
moved to #13111, see PR description of #13111 for more details.

PR #12949, opened 2016-11-08
Update ServiceLocator.php

Slayer911:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes

Before:

```php
var_dump(isset(Yii::$app->oldDb)); // false
```

But:
```php
\Yii::$app->oldDb;
var_dump(isset(Yii::$app->oldDb));  // true
```

Now:

```php
var_dump(isset(Yii::$app->oldDb)); // true
```

### DIFF
--- a/framework/di/ServiceLocator.php
+++ b/framework/di/ServiceLocator.php
@@ -107,7 +107,13 @@ class ServiceLocator extends Component
      */
     public function has($id, $checkInstance = false)
     {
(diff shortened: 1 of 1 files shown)


SilverFire:
Thank you for you effort.

According to the method description, 

> If `$checkInstance` is true, the method will return a value indicating whether the locator has
instantiated the specified component.

and 

> @param bool $checkInstance whether the method should check if the component is shared and instantiated

Forceful instantiating of `checkInstance` is a BC-breaking change and does not seem to be valid to me.

Issue #12938, opened 2016-11-08
Feature request: Allow additional parameters for renderDynamic


mikehaertl:
I propose that similar to Yii 1, it should be possible to pass additional parameters to `renderDynamic` that get serialized and stored in cache together with the cached content. The parameters should then be supplied to the `$statement` when it is `eval()`ed with `evaluateDynamicContent()`:

```php
public function renderDynamic($statement, $params);
```

```php
public function evaluateDynamicContent($statements, $params = [])
{
    extract($params);
    return eval($statements);
}
```

## How this worked in Yii 1

In Yii 1 we could have a construct like the following:

**`views/layouts/main.php`**
```php
<?php $this->renderDynamic('renderTracking', $this->trackingParams) ?>
```

**`components/BaseController.php`**
```php
public $trackingParams = [];
public function renderTracking($params = [])
{
    $params['user_id'] = Yii::app()->user->id;
    return 'dataLayer = '.json_encode($params);
}
```

**`views/site/some-page.php`** 

```php
$this->trackingParams = [
    'pageType' => 'some-page',
];
```

**`views/site/some-other-page.php`** 

```php
$this->trackingParams = [
    'pageType' => 'some-other-page',
    'other_param' => 'foo',
];
```

If we enable pache caching for these views the following happens:

 * The `pageType` and other data is set per view in `trackingParams`
 * The current `trackingParams` are passed to `renderDynamic` and thus serialized and stored in cache
 * In `renderTracking()` we receive these stored view dependend data in `$params` and enhance it with the current user id

## Situation in Yii 2

The above is no longer possible in Yii 2, at least not without an ugly trick like:

```php
<?= $this->renderDynamic(
    'return $this->renderTracking(unserialize('. serialize($this->trackingParams).'));'
); ?>
```

I would not consider this to be very readable or easy to follow for fellow team members.

dynasource:
can you supply an example in which the parameters are processed in eval()?

mikehaertl:
Not sure if I understand your question. But after more thinking I think, the implementation is actually very simple:

```
/**
 * Renders dynamic content returned by the given PHP statements.
 * This method is mainly used together with content caching (fragment caching and page caching)
 * when some portions of the content (called *dynamic content*) should not be cached.
 * The dynamic content must be returned by some PHP statements.
 * @param string $statements the PHP statements for generating the dynamic content.
 * @param array|null $params parameters that should be made available as variables in $statement.
[...]

PR #12955, opened 2016-11-09
Issue #12938 Add $params to View::renderDynamic()

mikehaertl:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   |  ?
| Fixed issues  | #12938

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -6,6 +6,7 @@ Yii Framework 2 Change Log
 
 - Enh #11058: Add `$checkAjax` parameter to method `yii\web\Controller::redirect()` which controls redirection in AJAX and PJAX requests (ivanovyordan)
 - Enh #12592: Optimized `yii\filters\AccessController` on processing accessrules (dynasource)
+- Enh #12938: Allow to pass additional parameters to `yii\base\View::renderDynamic()` (mikehaertl)
 - Removed methods marked as deprected in 2.0.x (samdark)
 - Chg #10771: Consistent behavior of `run()` method in all framework widgets. All return the result now for better extensibility (pkirill99, cebe)
 - Chg #11397: Minimum required version of PHP is 5.5.0 now (samdark)
--- a/framework/base/View.php
+++ b/framework/base/View.php
@@ -332,13 +332,26 @@ class View extends Component
      * Renders dynamic content returned by the given PHP statements.
      * This method is mainly used together with content caching (fragment caching and page caching)
      * when some portions of the content (called *dynamic content*) should not be cached.
-     * The dynamic content must be returned by some PHP statements.
+     * The dynamic content must be returned by some PHP statements. You can optionally pass
+     * additional parameters that will be available as variables in the PHP statement:
+     *
+     * ```php
+     * <?= $this->renderDynamic('return foo($myVar);', [
+     *     'myVar' => $model->getMyComplexVar(),
+     * ]) ?>
+     * ```
      * @param string $statements the PHP statements for generating the dynamic content.
+     * @param array $params the parameters (name-value pairs) that will be extracted and made
+     * available in the $statement context. The parameters will be stored in the cache and be reused
+     * each time $statement is executed. You should make sure, that these are safely serializable.
      * @return string the placeholder of the dynamic content, or the dynamic content if there is no
      * active content cache currently.
      */
-    public function renderDynamic($statements)
+    public function renderDynamic($statements, $params = null)
     {
+        if ($params !== null) {
+            $statements = 'extract(unserialize(\'' . serialize($params) . '\'));' . $statements;
+        }
         if (!empty($this->cacheStack)) {
             $n = count($this->dynamicPlaceholders);
             $placeholder = "<![CDATA[YII-DYNAMIC-$n]]>";
--- a/framework/base/View.php
+++ b/framework/base/View.php
@@ -344,12 +344,13 @@ class View extends Component
      * @param array $params the parameters (name-value pairs) that will be extracted and made
      * available in the $statement context. The parameters will be stored in the cache and be reused
      * each time $statement is executed. You should make sure, that these are safely serializable.
+     * @throws \yii\base\ErrorException if the statement throws an exception in eval()
      * @return string the placeholder of the dynamic content, or the dynamic content if there is no
      * active content cache currently.
      */
-    public function renderDynamic($statements, $params = null)
+    public function renderDynamic($statements, array $params = [])
     {
-        if ($params !== null) {
+        if (!empty($params)) {
             $statements = 'extract(unserialize(\'' . serialize($params) . '\'));' . $statements;
         }
         if (!empty($this->cacheStack)) {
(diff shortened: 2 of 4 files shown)


framework/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Yii Framework 2 Change Log
 
 - Enh #11058: Add `$checkAjax` parameter to method `yii\web\Controller::redirect()` which controls redirection in AJAX and PJAX requests (ivanovyordan)
 - Enh #12592: Optimized `yii\filters\AccessController` on processing accessrules (dynasource)
+- Enh #12938: Allow to pass additional parameters to `yii\base\View::renderDynamic()`
```
dynasource: please add your name (mikehaertl)

framework/base/View.php
```diff
@@ -334,11 +334,17 @@ public function renderPhpFile($_file_, $_params_ = [])
      * when some portions of the content (called *dynamic content*) should not be cached.
      * The dynamic content must be returned by some PHP statements.
      * @param string $statements the PHP statements for generating the dynamic content.
+     * @param array $params the parameters (name-value pairs) that will be extracted and made
```
dynasource: can you add one of your clear examples in the method doc above?

dynasource: shouldnt this be:
* 'and be made'
* 'and be reused'

framework/base/View.php
```diff
@@ -334,11 +334,17 @@ public function renderPhpFile($_file_, $_params_ = [])
...
+     * available in the $statement context. The parameters will be stored in the cache and
+     * reused each time $statement is executed. You should make sure, that they are safely serializable.
```
dynasource: they = these

dynasource (changes requested):
good addon to the framework. Don't forget the unit tests of course. We want your solution to be perfect now and in the future.

framework/base/View.php
```diff
@@ -334,11 +334,17 @@ public function renderPhpFile($_file_, $_params_ = [])
...
      * @return string the placeholder of the dynamic content, or the dynamic content if there is no
      * active content cache currently.
      */
-    public function renderDynamic($statements)
+    public function renderDynamic($statements, $params = null)
     {
+        if ($params !== null) {
+            $statements = 'extract(unserialize(\'' . serialize($params) . '\'));'.$statements;
```
samdark: Need spaces around `.` when concatenating.

samdark (approved):
Overall all is good.

framework/base/View.php
mikehaertl: @dynasource I've copied the sentence from the other `$params`, e.g. from `render()`. There's no `be` there either.

samdark:
How about a unit test so it's set in stone and never broken?

mikehaertl:
Regarding the unit tests: There are no tests for `renderDynamic()` yet, at least I couldn't find any. I'm not a real expert on writing framework tests, so I will not find time to write a new test from scratch. Even more as this is a non-trivial task here that involves a complex setup including cache etc.

dynasource:
no problem. Ill add these later.

dynasource (approved):

framework/base/View.php
```diff
@@ -332,13 +332,26 @@ public function renderPhpFile($_file_, $_params_ = [])
...
+     * ```
      * @param string $statements the PHP statements for generating the dynamic content.
+     * @param array $params the parameters (name-value pairs) that will be extracted and made
+     * available in the $statement context. The parameters will be stored in the cache and be reused
+     * each time $statement is executed. You should make sure, that these are safely serializable.
      * @return string the placeholder of the dynamic content, or the dynamic content if there is no
      * active content cache currently.
      */
-    public function renderDynamic($statements)
+    public function renderDynamic($statements, $params = null)
```
rob006: Why not typehint `$params` as array?

mikehaertl: I found it easier to check `$params === null` than `count($params) === 0`. But I don't really mind to change this. We should also use `@param array|null ...` in the comment block if we keep `null`.

rob006: With

```php 
public function renderDynamic($statements, array $params = null)
```

you could use `$params !== []` or just `!empty($params)`.

mikehaertl: @dynasource Any preferences? Should I change it?

rob006: I fixed my example.

dynasource: Im working on the tests now (encountering some problems with testing eval()). Will include the array hinting, so you're allowed to take a break ;)

samdark (approved):

rob006 (changes requested):
This change should be noted in `UPGRADE.md`.

dynasource:
thanks to you all!

framework/base/View.php
samdark: @mikehaertl `$array !== []` with array type hinting is preferred.

rob006: @samdark Why? `empty()` is faster than `=== []` and IMO more readable.

dynasource: although the performance differences are minor, empty is indeed more readable

samdark: @rob006 is it faster?

samdark: Indeed it is. Let's use it then.

mikehaertl: @rob006 @samdark Any references? And is this only for empty arrays? For many years now we've seen `===` as the fastest way to compare e.g. for `null` or empty strings. It's still used in many places all over the framework. Did this change?

samdark: I've made a test for `empty($array)` and `$array !== []` for both empty and non-empty arrays at https://3v4l.org/.

dynasource: yeah, did the same for `isset()` vs `===` a week ago. Therefore I prefer isset() from now. Faster and more readable

mikehaertl: @samdark @dynasource Hmm, I'd vote against that. Unless there's a reliable "official" source that `isset` and `empty` are really always faster, why suddenly change what was considered best practice over all the years? It's like a convention in the Yii framework to always use `===` whenever applicable.

I also don't agree that `isset()` is more readable than `===null`. It's not even the same thing. `isset()` also works for undefined variables. Something you don't always want.

So before changing this convention lightheartedly we should really ask, if this is really neccessary. "If it ain't broken - don't fix it!".

dynasource: * https://3v4l.org/ can show you the prove. You can try it yourself. @samdark, where your full url :).
* isset() is more readable. You read from left to right. Isset() shows immediately whats going on, compared to `===` which can be a equation to anything
* 'It's not even the same thing': agree, which shows you, you have to use it when it is appropiate

samdark: @mikehaertl it's not suddenly. Actually [it's in our code style](https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md#--vs-empty) for quite a while...

mikehaertl: @samdark If it's only about arrays I'm fine. It wasn't me who brought up the `isset()` topic ;).

Still the `===null` vs `isset()` convention should also be part of the coding style guide.

dynasource: :)

srnden:
I would like to see this change in 2.0.*

samdark:
It won't be merged since 2.0.x doesn't accept features from last year.

PR #12959, opened 2016-11-09
Added bash completion file for ./yii commands

### LINKED ISSUE #475: yii command bash autocompletion
I created one for yii1.1 yiic for a specific case. Will try to work on a generic solution.
https://gist.github.com/cebe/1141050
Why? because bash is only fun when it helps you typing less as much as possible ;)

cebe:
| Q             | A
| ------------- | ---
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | none included
| Fixed issues  | #475

This allows to auto-complete commands available by the `./yii`
executable on the bash by expanding the first argument with the TAB key.
E.g. Typing `./yii [TAB]` will list all commands. `./yii mig[TAB]` will
auto-complete to `./yii migrate` and tab after that will list all
migration sub-commands(actions).

fixes #475


framework/yii.bash-completion
```diff
@@ -0,0 +1,31 @@
+# This file implements bash completion for the ./yii command file.
```
SilverFire: `#!/bin/bash`

framework/yii.bash-completion
```diff
@@ -0,0 +1,31 @@
...
+# It completes the commands available by the ./yii command.
+# See also:
+# - https://debian-administration.org/article/317/An_introduction_to_bash_completion_part_2 on how this works.
+# - https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion.html
+#
+# Usage:
+# Temporarily you can source this file in you bash by typing: . yii.bash-completion
```
SilverFire: I'd prefer `source yii.bash-completion` instead of `. yii.bash-completion` as it is more readable and less confusing.

framework/yii.bash-completion
```diff
@@ -0,0 +1,31 @@
...
+# For permanent availability, copy or link this file to /etc/bash_completion.d/
```
SilverFire: Maybe it's better to put it to `/usr/share/bash-completion/completions`?

docs/guide/tutorial-console.md
```diff
@@ -107,6 +107,21 @@ You can see an example of this in the advanced project template.
...
 
 
+Bash completion
+---------------
+
+Auto-completion of command arguments is a useful thing when working with the shell. Since version 2.0.11, the `./yii` command provides auto completion
+for the bash out of the box. This is provided by the `yii.bash-completion` file located in the framework base directory.
+For temporary usage you can source that file by typing:
+
+    . yii.bash-completion
```
SilverFire: I'd prefer source yii.bash-completion instead of . yii.bash-completion as it is more readable and less confusing.

docs/guide/tutorial-console.md
```diff
@@ -107,6 +107,21 @@ You can see an example of this in the advanced project template.
...
+
+Note the dot on the beginning of that line. It means you want to include the definitions give in the `yii.bash-completion` file in
```
SilverFire: And then this line could be removed

SilverFire (changes requested):

SilverFire:
Will try to do the same for ZSH

framework/yii.bash-completion
cebe: no, this file should not be executed as a script but only be sourced.

framework/yii.bash-completion
cebe: `/usr/` is for software installed by the package mangager. `/etc/` is the place for custom files like this one.

framework/yii.bash-completion
SilverFire: Ok

framework/yii.bash-completion
SilverFire: Right.

dynasource (approved):

SilverFire (changes requested):
I don't like that completion file is located in the `/framework` directory because it is not a part of framework. I suggest to put in somewhere in `/contrib` or similar

dynasource:
> I don't like that completion file is located in the /framework directory because it is not a part of framework. I suggest to put in somewhere in /contrib or similar

sharply noticed. I would expect a separate folder indeed.

cebe:
Well, it belongs to `./yii` so I but it next to it. and it should be inside of the framework to ship with it. How about putting it into /console folder?

SilverFire:
Take a look how docker-compose solved this task: https://docs.docker.com/compose/completion/

cebe:
You mean to put it somewhere in yiisoft/yii2 repo and use curl in the docs to fetch it?

SilverFire:
@cebe exactly

SilverFire:
ezgif com-gif-maker

SilverFire (approved):

framework/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Yii Framework 2 Change Log
 - Bug #12904: Fixed lowercase table name in migrations (zlakomanoff)
 - Bug #12939: Hard coded table names for MSSQL in RBAC migration (arogachev)
 - Bug #12974: Fixed incorrect order of migrations history in case `yii\console\controllers\MigrateController::$migrationNamespaces` is in use (evgen-d, klimov-paul)
+- Enh #475: Added basic bash completion support for the `./yii` command (cebe)
```
cebe: @SilverFire add your name here and add zsh ;)

cebe (commented):
let me add the options to bash too.

samdark:
It behaves a bit weird to me when there's no `./yii` in current dir and I'm typing `./yii` and then `TAB`. It prints out ` ./yii bash: ./yii: No such file or directory`, then waits for `ENTER` to be pressed, then prints `bash: ./yii: No such file or directory`.

samdark:
Doesn't seem to be working under MINGW64 bash. Script properly calls `help/list` command but then I'm getting nothing back after delay.

cebe:
> It behaves a bit weird to me when there's no ./yii in current dir

good catch, will fix that.

cebe:
added argument completion to bash. also fixed the issue with `./yii` not existing. This PR is ready to review and merge.

PR #12967, opened 2016-11-10
improved error message for Instance::ensure() when component does not exist

### LINKED ISSUE #7333: Not informative ReflectionException
Hello all.
When i run application, i catch not informative error, which took a lot of my time.

When in components.php set session component like this: 

``` php
    'session' => [
        'class' => 'yii\web\CacheSession'
    ]
```

and missing "cache" component, then yii2 throw this error:
https://docs.google.com/file/d/0B93SvIJkBgcwaDlUdFlDVWhSZ1E/edit

Yii2 try to create class "cache", but normaly yii2 should show an error about missing component !!!

cebe:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

fixes #7333 

- Better error message
- and the Instance::ensure() call is on the second line in the stack trace which makes it easier to locate the source of the problem:
    bildschirmfoto von 2016-11-10 22 13 48

### DIFF
--- a/framework/di/Instance.php
+++ b/framework/di/Instance.php
@@ -128,7 +128,11 @@ class Instance
         }
 
         if ($reference instanceof self) {
-            $component = $reference->get($container);
+            try {
+                $component = $reference->get($container);
+            } catch(\ReflectionException $e) {
+                throw new InvalidConfigException('Failed to instantiate component or class "' . $reference->id . '".', 0, $e);
+            }
             if ($type === null || $component instanceof $type) {
                 return $component;
             } else {
--- a/tests/framework/di/InstanceTest.php
+++ b/tests/framework/di/InstanceTest.php
@@ -46,7 +46,27 @@ class InstanceTest extends TestCase
         $this->assertTrue(Instance::ensure(['class' => 'yii\db\Connection', 'dsn' => 'test'], 'yii\db\Connection', $container) instanceof Connection);
     }
 
-    public function testEnsureWithoutType()
+    /**
+     * ensure an InvalidConfigException is thrown when a component does not exist.
+     */
+    public function testEnsure_NonExistingComponentException()
+    {
+        $container = new Container;
+        $this->setExpectedExceptionRegExp('yii\base\InvalidConfigException', '/^Failed to instantiate component or class/i');
+        Instance::ensure('cache', 'yii\cache\Cache', $container);
+    }
+
+    /**
+     * ensure an InvalidConfigException is thrown when a class does not exist.
+     */
+    public function testEnsure_NonExistingClassException()
+    {
+        $container = new Container;
+        $this->setExpectedExceptionRegExp('yii\base\InvalidConfigException', '/^Failed to instantiate component or class/i');
+        Instance::ensure('yii\cache\DoesNotExist', 'yii\cache\Cache', $container);
+    }
+
+    public function testEnsure_WithoutType()
     {
         $container = new Container;
         $container->set('db', [
@@ -59,7 +79,7 @@ class InstanceTest extends TestCase
         $this->assertTrue(Instance::ensure(['class' => 'yii\db\Connection', 'dsn' => 'test'], null, $container) instanceof Connection);
     }
 
-    public function testEnsureMinimalSettings()
+    public function testEnsure_MinimalSettings()
     {
         Yii::$container->set('db', [
             'class' => 'yii\db\Connection',
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -22,6 +22,7 @@ Yii Framework 2 Change Log
 - Bug #12939: Hard coded table names for MSSQL in RBAC migration (arogachev)
 - Bug #12974: Fixed incorrect order of migrations history in case `yii\console\controllers\MigrateController::$migrationNamespaces` is in use (evgen-d, klimov-paul)
 - Enh #6809: Added `\yii\caching\Cache::$defaultDuration` property, allowing to set custom default cache duration (sdkiller)
+- Enh #7333: Improved error message for `yii\di\Instance::ensure()` when a component does not exist (cebe)
 - Enh #7420: Attributes for prompt generated with `renderSelectOptions` of `\yii\helpers\Html` helper (arogachev)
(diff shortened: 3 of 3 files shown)


tests/framework/di/InstanceTest.php
```diff
@@ -46,6 +46,26 @@ public function testEnsure()
         $this->assertTrue(Instance::ensure(['class' => 'yii\db\Connection', 'dsn' => 'test'], 'yii\db\Connection', $container) instanceof Connection);
     }
 
+    /**
+     * ensure an InvalidConfigException is thrown when a component does not exist.
+     */
+    public function testEnsureNonExistingComponent()
```
dynasource: shouldnt the method name be more descriptive about the actual test? In this case testing an exception which matches your method @doc.

cebe: therefore I added the comment, it would be redundant to try to encode the whole description from phpdoc in the method name.

dynasource: why do we need a comment for a test in the first place

cebe: to describe which scenario it tests?

dynasource: it should be the test method to describe that. Thats whats you see in all the exports & test consoles & travis reports, not the doc statement.

cebe: do you have a suggestion for a better name? What you mean would be more like the message on an assertation, no?

samdark (approved):

tests/framework/di/InstanceTest.php
dynasource: for me also a clear case for the requirement of the underscore, as the verb of the original method creates confusion:
* testEnsure_NonExistingClassException
* testEnsure_NonExistingComponentException

dynasource (changes requested):

tests/framework/di/InstanceTest.php
samdark: testNonExistingClassException, testNonExistingComponentException

dynasource: then its unclear that it is about the ensure method

samdark: Tests are always about checking for validity of scenario. Not necessary a single method.

dynasource: agree, but imagine you are testing in your IDE and you get an error:
* testNonExistingClassException -> "Failing test non existing class exception"

Is the developer informed completely enough? Does he instantly know the reason and source of the failing test?

Additionally. When looking at this subject from a 'test management perspective'. Is it clear enough in the 'test file' that tests already have been made for the 'ensure' method.

framework/di/Instance.php
```diff
@@ -128,7 +128,11 @@ public static function ensure($reference, $type = null, $container = null)
         }
 
         if ($reference instanceof self) {
-            $component = $reference->get($container);
+            try {
```
samdark: try-catch may slow down overall execution and since it's a central part of the framework executed hundreds of times per request I think we need to test how it affects performance. @cebe can you do it?

dynasource: good point

cebe: Created this: https://3v4l.org/N4rbA not sure if the results are significant in any way.

samdark: Seems not...

dynasource: there are two ways to evaluate:
1. choose by speed. In this case, the try catch is two times slower
2. choose by threshold: we define a duration threshold in which we evaluate. 

In this case we are talking about a difference of 0.005/100000 = 0.00000005 seconds per occurrence. Even if it is run 100 times, it will be 0.000005 seconds. Neglectable, so to speak. 

Of course, if it can be avoided, it's always good. Imagine 1000 of these usecases. Then it will be 0.005s which is already measurable by the debug toolbar.


image

image

cebe: I do not think the impact is significant. Development speed, i.e. clear error message has more weight here.

cebe:
Ready to review/merge.

tests/framework/di/InstanceTest.php
```diff
@@ -46,7 +46,27 @@ public function testEnsure()
...
+     */
+    public function testEnsure_NonExistingComponent()
+    {
+        $container = new Container;
+        $this->setExpectedExceptionRegExp('yii\base\InvalidConfigException', '/^Failed to instantiate component or class/i');
+        Instance::ensure('cache', 'yii\cache\Cache', $container);
+    }
+
+    /**
+     * ensure an InvalidConfigException is thrown when a class does not exist.
```
dynasource: doc can be removed. The method says it all

tests/framework/di/InstanceTest.php
```diff
@@ -46,7 +46,27 @@ public function testEnsure()
         $this->assertTrue(Instance::ensure(['class' => 'yii\db\Connection', 'dsn' => 'test'], 'yii\db\Connection', $container) instanceof Connection);
     }
 
-    public function testEnsureWithoutType()
+    /**
+     * ensure an InvalidConfigException is thrown when a component does not exist.
```
dynasource: can be removed

tests/framework/di/InstanceTest.php
```diff
@@ -46,7 +46,27 @@ public function testEnsure()
...
+     */
+    public function testEnsure_NonExistingComponent()
```
dynasource: inconsistent with next test method covering the word 'exception'

tests/framework/di/InstanceTest.php
cebe: a readable description is much better than a method name. There is no such thing as self documenting code.

tests/framework/di/InstanceTest.php
cebe: fixed.

tests/framework/di/InstanceTest.php
dynasource: perhabs in normal code, not in a test environment. Tests should be small and self declaring. Your method docs are useless here.

PR #12971, opened 2016-11-11
Test against the current hhvm version

photodude:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes for travis hhvm version
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | comma-separated list of tickets # fixed by the PR, if any fixes version of hhvm test run against

This provides the current HHVM version (3.15.2 as of this PR) and will track with each release (i.e. will be 3.16 when 3.16 is released).

If testing against HHVM LST versions is desired follow this guide. https://docs.travis-ci.com/user/languages/php#HHVM-versions

Should be able to change to container based Trusty after Q1-17 https://blog.travis-ci.com/2016-11-08-trusty-container-public-beta/


.travis.yml
```diff
@@ -7,10 +9,27 @@ php:
...
+            - mysql-server-5.6
+            - mysql-client-core-5.6
+            - mysql-client-5.6
+      services:
+        - mysql
+        - postgresql
+        - redis-server
+      env: CUBRID_VERSION=9.3.0/CUBRID-9.3.0.0206 CUBRID_PDO_VERSION=9.3.0.0001
   allow_failures:
+    - php: hhvm
```
samdark: hhvm failures are not allowed.

photodude: There is a testing setup issue which is why I've set hhvm to  allowed failure.

Testing on Trusty does not have a mysql `travis` user
You can read the reason for root in the travis docs
https://docs.travis-ci.com/user/database-setup/#MySQL-5.6

I'm not sure if there is a way to create a valid mysql `travis` user on Trusty that would have the same permissions and equivalance as the mysql `travis` user on the Precise containers. On Precise, it's part of the build no user creation needed. Unfortunately, yii2 has hard coded the travis user in the test setup. If you have suggestions on working around that detail I'll make the necessary changes

samdark: How about reading credentials from environment and falling back to what it is now in case environment is not defined?

.travis.yml
```diff
@@ -36,7 +52,7 @@ addons:
 
 install:
   - |
-    if [ $TRAVIS_PHP_VERSION != '5.6' ]; then
+    if [[ $TRAVIS_PHP_VERSION != '5.6' || $TRAVIS_PHP_VERSION != hhv* ]]; then
```
cebe: this will disable xdebug always. should be `[ $TRAVIS_PHP_VERSION != '5.6' ] && [ $TRAVIS_PHP_VERSION != hhv* ]`

.travis.yml
```diff
@@ -45,8 +61,14 @@ install:
   - export PATH="$HOME/.composer/vendor/bin:$PATH"
 # core framework:
   - travis_retry composer update --prefer-dist --no-interaction
-  - tests/data/travis/apc-setup.sh
-  - tests/data/travis/memcache-setup.sh
+  - |
+    if [[ $TRAVIS_PHP_VERSION != hhv* ]]; then
```
cebe: this should already be covered inside of the apc-setup script.

.travis.yml
```diff
@@ -45,8 +61,14 @@ install:
...
+      tests/data/travis/apc-setup.sh
+    fi
+  - |
+    if [[ $TRAVIS_PHP_VERSION != hhv* ]]; then
```
cebe: this should already be covered inside of the memcache-setup script.

.travis.yml
```diff
@@ -57,7 +79,7 @@ before_script:
   - psql --version
 
   # initialize databases
-  - mysql -e 'CREATE DATABASE yiitest;';
+  - mysql -u root -e 'CREATE DATABASE yiitest;';
```
cebe: why?

.travis.yml
cebe: we could just add a travis users if none is present.

.travis.yml
photodude: xdebug is not enabled by default on hhvm, additionally phpenv for setting environment variables [has issues with hhvm](https://github.com/travis-ci/travis-ci/issues/3930). The correct way to disable xdebug on travis for hhvm is
```php
# Disable xdebug in hhvm.
if [[ $TRAVIS_PHP_VERSION = hhv* ]]; then echo 'xdebug.enable = 0' >> /etc/hhvm/php.ini; fi
```
but it is unneccesary since xdebug is not enabled by default on hhvm

.travis.yml
photodude: I see, I'll revert the appropriate changes for the two scripts.

.travis.yml
photodude: I think I got the travis user creation working for trusty builds now.

Still need to keep HHVM as allowed failure as there are failing tests.

samdark: Why errors? We have master tested against HHVM currently and it works.

arogachev:
Yes, current HHVM version that is used for testing is [3.6.6](https://github.com/facebook/hhvm/releases/tag/HHVM-3.6.6) which is kind of old, it was released on 31 Aug 2015,

It's a default version on `precise` dist.

`trusty` dist has HHVM 3.15.2 by default and build is failing because of these two tests:

```
There were 2 failures:

1) yii\tests\unit\framework\db\pgsql\CommandTest::testSaveSerializedObject

Failed asserting that 0 matches expected 1.

/home/travis/build/yiisoft/yii2/tests/framework/db/pgsql/CommandTest.php:86

/home/travis/build/yiisoft/yii2/vendor/phpunit/phpunit/phpunit:47

2) yii\tests\unit\framework\db\pgsql\CommandTest::testBindParamValue

/home/travis/build/yiisoft/yii2/tests/framework/db/CommandTest.php:199

/home/travis/build/yiisoft/yii2/vendor/phpunit/phpunit/phpunit:47
```

I faced the same problem after changing dist to `trusty` in PR #12840.

We have two options:

- Set HHVM version to `hhvm-3.6` as I did #12840. Then test will behave the same as for Trusty. See [HHVM versions on Trusty](https://docs.travis-ci.com/user/languages/php#HHVM-versions-on-Trusty) section in Travis docs.
- Allow failures until these two issues will be fixed (as it was with PHP 7.1 recently). I think in this case we can just add `hhvm` to existing `matrix` - `allow_failures` section along with `php: nightly`.

As for `matrix` - `include`. I have read [Explicitly Including Jobs](https://docs.travis-ci.com/user/customizing-the-build#Explicity-Including-Jobs) section in Travis docs, But what's the benefit of it related to HHVM?

Also note that in HHVM related build job you are using PostgreSQL 9.6.1. It should be 9.3 for all jobs according to current config.

.travis.yml
photodude: @samdark the 2 failures on hhvm are related to pgsql. These could be related to the differences between hhvm's pgsql module and the zend pgsql module, or related to a couple of items that are fixed in upcoming releases for hhvm 3.15.3 and hhvm 3.16.0, or could be related to other php vs hhvm differences.

```php
1) yii\tests\unit\framework\db\pgsql\CommandTest::testSaveSerializedObject
Failed asserting that 0 matches expected 1.
/tests/framework/db/pgsql/CommandTest.php:86
/vendor/phpunit/phpunit/phpunit:47
2) yii\tests\unit\framework\db\pgsql\CommandTest::testBindParamValue
/tests/framework/db/CommandTest.php:199
/vendor/phpunit/phpunit/phpunit:47
```

photodude:
@arogachev unless you are using the beta Trusty container for all tests you have to use the matrix include directive to set the hhvm testing to use the sudo required trusty for just hhvm.

In that the failures are specific to pgsql we could try testing on any LTS version before 3.15 when pgsql was moved from a 3rd party add on to the core of hhvm. `hhvm-3.12` would likely be a good option that would exclude testing on pgsql, if that is the direction that is decided on. I personally think it's better to allow failure on hhvm or use other options to skip the pgsql tests and note the incompatibility with running yii2 on hhvm with pgsql.

.travis.yml
photodude: @arogachev I solved it in a similar fassion via creating a travis user
```yml
mysql -u root -e 'CREATE DATABASE yiitest;';
mysql -u root -e 'CREATE USER 'travis'@'localhost' IDENTIFIED WITH mysql_native_password;'
mysql -u root -e 'GRANT ALL PRIVILEGES ON *.* TO 'travis'@'localhost' WITH GRANT OPTION;
```

photodude:
@arogachev the hhvm test is now set for postgresql: "9.3" rather than the default PostgreSQL 9.2 as requested (note: PostgreSQL 9.6 is not currently an option on Travis as it was released in September 2016)

.travis.yml
```diff
@@ -35,8 +52,8 @@ addons:
 
 install:
   - |
-    if [ $TRAVIS_PHP_VERSION != '5.6' ]; then
-        # disable xdebug for performance reasons when code coverage is not needed
+    if [[ $TRAVIS_PHP_VERSION != '5.6' || $TRAVIS_PHP_VERSION != hhv* ]]; then
```
cebe: this needs to be `&&`, not `||`. We want to disable xdebug on all php versions that are not 5.6 **and** not hhvm.

cebe (changes requested):

cebe:
> 2) yii\tests\unit\framework\db\pgsql\CommandTest::testBindParamValue 
>
> Failed asserting that false is true.
> /home/travis/build/yiisoft/yii2/tests/framework/db/CommandTest.php:199
> /home/travis/build/yiisoft/yii2/vendor/phpunit/phpunit/phpunit:47

this can be fixed by applying the following patch to the test:

```diff
diff --git a/tests/framework/db/CommandTest.php b/tests/framework/db/CommandTest.php
index e3a4f6a..8d24723 100644
--- a/tests/framework/db/CommandTest.php
+++ b/tests/framework/db/CommandTest.php
@@ -193,7 +193,7 @@ SQL;
         $this->assertEquals($intCol, $row['int_col']);
         $this->assertEquals($charCol, $row['char_col']);
         $this->assertEquals($floatCol, $row['float_col']);
-        if ($this->driverName === 'mysql' || $this->driverName === 'sqlite' || $this->driverName === 'oci') {
+        if ($this->driverName === 'mysql' || $this->driverName === 'sqlite' || $this->driverName === 'oci' || (defined('HHVM_VERSION') && $this->driverName === 'pgsql')) {
             $this->assertEquals($blobCol, $row['blob_col']);
         } else {
             $this->assertTrue(is_resource($row['blob_col']));
```

cebe:
> 1) yii\tests\unit\framework\db\pgsql\CommandTest::testSaveSerializedObject
> 
> Failed asserting that 0 matches expected 1.
> 
> /home/travis/build/yiisoft/yii2/tests/framework/db/pgsql/CommandTest.php:86
> /home/travis/build/yiisoft/yii2/vendor/phpunit/phpunit/phpunit:47

This would mean that the number of rows affected by the insert query is returned as 0, which I would consider a bug in the pgsql implementation in hhvm. One row is inserted so 1 should be returned...

.travis.yml
photodude: I have made this change

photodude:
>  I would consider a bug in the pgsql implementation in hhvm

@cebe I would agree considering that hhvm's implementation is missing about 20 of the zend pgsql functions. If your underlying driver uses [one of those missing functions](https://github.com/PocketRent/hhvm-pgsql#differences-from-zend) that would be the related issue.

cebe:
> @cebe I would agree considering that hhvm's implementation is missing about 20 of the zend pgsql functions. If your underlying driver uses one of those missing functions that would be the related issue.

no, Yii uses the PDO layer of PHP, not the `pg_*` functions.

photodude:
@cebe I applied your patch and have the following errors

```php
There were 2 failures:
1) yii\tests\unit\framework\db\pgsql\CommandTest::testSaveSerializedObject
Failed asserting that 0 matches expected 1.
/tests/framework/db/pgsql/CommandTest.php:86
/vendor/phpunit/phpunit/phpunit:47
2) yii\tests\unit\framework\db\pgsql\CommandTest::testBindParamValue
Failed asserting that null matches expected Binary String: 0x101112.
/tests/framework/db/CommandTest.php:197
/vendor/phpunit/phpunit/phpunit:47
```

photodude:
> Yii uses the PDO layer of PHP, not the `pg_*` functions.

@cebe I'm not sure how closely coupled hhvm's PDO_pgsql module is to it's full pgsql.so module, They were implemented together in hhvm 3.15.0 as a direct port from the previous 3rd party module.

cebe:
okay, now it looks like two HHVM bugs to me. thanks for investigation. I think we should isolate the issues and report them to HHVM. If they agree that these are bugs we can mark the tests as skipped as long as the issues are unfixed.

photodude:
@cebe sounds like a plan. In the mean time, I would suggest we merge this PR with HHVM as an allowed failure until the tests are skipped or fixed.

.travis.yml
```diff
@@ -7,10 +9,27 @@ php:
   - 7.0
   - 7.1
   - nightly
-  - hhvm
```
cebe: as far as I see, if this stays here, there will be tests run against the old version and the new version so we could keep this to test against two versions of hhvm.

cebe:
merging with "allow failure" will prevent us from spotting hvvm specific problems in other PRs so I'd prefer to skip these specific tests on HHVM for now.

cebe:
I am unable to reproduce the insert count issue with plain PDO: https://travis-ci.org/cebe/hhvm-bugreport/builds/176069208

cebe:
I have reported the two failing tests as a separate issue #13010. I will merge this PR as soon as travis is green.

cebe (approved):

cebe:
Merged, thank you!

photodude:
@cebe You are welcome. 

ps you could [test hhvm via a newer LTS version]( https://docs.travis-ci.com/user/languages/php#HHVM-versions) that didn't have the pgsql module. hhvm 3.12 only recently went EOL use this code in the matrix

```yml
   # test against HHVM 3.12 LTS version by using trusty
    - php: hhvm-3.12
      sudo: true
      dist: trusty
      group: edge # Use edge image until the next travis CI image update
      addons:
        apt:
          packages:
            - mysql-server-5.6
            - mysql-client-core-5.6
            - mysql-client-5.6
      services:
        - mysql
      env: CUBRID_VERSION=9.3.0/CUBRID-9.3.0.0206 CUBRID_PDO_VERSION=9.3.0.0001
```

cebe:
@photodude cool, can you send a PR so that we test against LTS and the latest version?

photodude:
@cebe see PR #13012

Note: hhvm 3.15 is an LTS version but is also current until 3.16 is released. Both 3.15 and 3.16 include the PostgreSQL module that have a problem resulting in 2 tests being skipped.

PR #12977, opened 2016-11-12
fix #12969 pjax not working because of non insync ID

### LINKED ISSUE #12969: Update the guide for the Pjax ID requirement
We're getting many issues about PJAX. I see a red line which is proven by the guide. 
http://www.yiiframework.com/doc-2.0/yii-widgets-pjax.html
There is no Pjax example with an ID.

We should either update the guide with an example with an ID or generate an unique ID by default. This latter is a challenge but can be solved if you generate the ID by the {class}:{line} in which the pjax widget is called.

dynasource:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  |no
| Breaks BC?    |no
| Tests pass?   | yes
| Fixed issues  | #12969

This commit fixes the pjax `ID not set` issues out of the box. It is not required anymore to supply an unique ID to Pjax widgets anymore because the software is creating a truly unique one by itself. The algorithm to do this is based on the line & class (the trace) where the Pjax widget is called. This is unique for most Pjax::begin calls. When developers call this trough a pjax proxy (like i.e. a ->pjaxify), also the second trace is required, which is covered by this commit. MD5 does the rest.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -32,7 +32,7 @@ Yii Framework 2 Change Log
 - Enh #12881: Added `removeValue` method to `yii\helpers\BaseArrayHelper` (nilsburg)
 - Enh #12901: Added `getDefaultHelpHeader` method to the `yii\console\controllers\HelpController` class to be able to override default help header in a class heir (diezztsk)
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
-
+- Bug #12969: Fixed Pjax result when no ID is supplied `yii\widgets\Pjax` (dynasource)
 
 2.0.10 October 20, 2016
 -----------------------
--- a/framework/widgets/Pjax.php
+++ b/framework/widgets/Pjax.php
@@ -101,6 +101,18 @@ class Pjax extends Widget
 
     /**
      * @inheritdoc
+     * @since 2.0.11
+     */
+    public static function begin($config = [])
+    {
+        if (!isset($config['id'])) {
+            $config['id'] = md5(serialize(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2))); //fixes #12969
+        }
+        return parent::begin($config);
+    }
+
+    /**
+     * @inheritdoc
      */
     public function init()
     {
--- a/tests/framework/widgets/PjaxTest.php
+++ b/tests/framework/widgets/PjaxTest.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace yiiunit\framework\widgets;
+
+use Yii;
+use yii\web\Request;
+use yii\widgets\Pjax;
+
+/**
+ * @group widgets
+ */
+class PjaxTest extends \yiiunit\TestCase
+{
+    public function testSetDefaultIdByTrace()
+    {
+        $this->expectOutputRegex('~<div id="[a-zA-Z0-9]{32}".*>~');
+        Pjax::begin();
+    }
+
+    public function testCustomId()
+    {
+        $this->expectOutputRegex('~<div id="custom-id".*>~');
+        Pjax::begin([
+            'id' => 'custom-id'
+        ]);
+    }
+
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->mockApplication();
+        Yii::$app->set('request', new Request());
+    }
+}


dynasource:
please note the fix to be inside `::begin` and not in init() or getId() on purpose, because of a polluted stacktrace.

rob006:
Using `debug_backtrace()` in regular production code looks like really dirty hack...

dynasource:
the application of debug_backtrace in this solution is innocent. It did not add any memory requirement and it is purely used to generate an unique id. But be my guest and come up with a better solution.

rob006:
Dose separate `$autoIdPrefix` and `$counter` for `PjaxWidget` will not solve this problem?

samdark:
It would give non-unique IDs for two PJAX views (same backtrace, different args) rendered on the same page thus would result in conflicts.

dynasource:
> It would give non-unique IDs for two PJAX views (same backtrace, different args) rendered on the same page thus would result in conflicts.

Correct. This usecase is not yet covered. Ive looked into it.

It could be solved by:
* including the arguments in the backtrace, but this would go at cost of performance with potentially rendered objects.
* extending the stacktrace to a point in which the stacktrace shows variety for your usecase. I looked into both the stacktraces for pjax & non-pjax requests and a depth of 12 seems to be enough to cover it. However, the deeper the depth, the more it goes at costs of performance.

I think we have to make a choice here:
* We can create the perfect solution which covers all rare usecases going significantly at cost of performance. 
* We choose for performance and if someone wants to cover his rare usecase like in https://github.com/yiisoft/yii2/pull/12977#issuecomment-260123516, he has to set this ID manually and put this in the guide

IMO, we should go for ultimate performance & choose the solution which covers >95% of the usecases. This PR seems to do that. The other 5% can not be served automatically, this has to be done by the end user theirselves.

rob006:
It still does not solve https://github.com/yiisoft/yii2/issues/12969#issuecomment-260120636

dynasource:
> It still does not solve #12969 (comment)

your statements would be more helpful in combination with arguments

rob006:
Example shows rendering the same widget from different controller actions - the backtraces will be different, so widgets will have different IDs.

dynasource:
Thanks. Now I understand. 

No. You are right. Two different controllers rendering the same PJAX widget is not supported with this PR. Isnt this another rare usecase? @SilverFire?

rob006:
> Two different controllers rendering the same PJAX widget is not supported with this PR. Isnt this another rare usecase?

Then which use cases are supported by this PR but not by my proposal: https://github.com/yiisoft/yii2/pull/12977#issuecomment-260123078?

dynasource:
this PR creates unique ID's for Pjax widgets despite how many (PJAX) widgets are loaded during the AJAX call. 

Your suggestion does not really change the current situation, apart from narrowing the ID creation to pjax widgets only.

rob006:
> this PR creates unique ID's for Pjax widgets despite how many (PJAX) widgets are loaded during the AJAX call. 

Same as my proposal.

> Your suggestion does not really change the current situation, apart from narrowing the ID creation to pjax widgets only.

Which solves https://github.com/yiisoft/yii2/issues/12969#issuecomment-260120636

dynasource:
Your proposal is similar to the current situation in which you are still using a counter, exposed to the same risk of having wrong Pjax ID's when not all Pjax widgets are loaded in the Ajax request.

Therefore, this is not the same as this PR.

rob006:
When you will have the same backtrace and different pjax widgets count?

Alex-Code:
Because of pjax's unique nature would it be simpler to force an ID as a requirement and throw an error otherwise?

SilverFire:
@dynasource and @rob006 battle is a good example of polite arguing :+1: 

@Alex-Code it's a major BC breaking change.

> No. You are right. Two different controllers rendering the same PJAX widget is not supported with this PR. Isnt this another rare usecase? @SilverFire?

Maybe, PJAX placed somewhere in layout. Need to think more

dynasource:
> Because of pjax's unique nature would it be simpler to force an ID as a requirement and throw an error otherwise?

If we can't solve it properly with automation, this would be the best alternative.

dynasource:
> When you will have the same backtrace and different pjax widgets count?

You will get a different widgets count if multiple Pjax widgets reside in both layouts & views:
1. original page: during the load of the non-ajax page, all Pjax widgets are loaded, from both layout & views
2. ajax page: developers choose to load only the view. Not all Pjax widgets are loaded. 

There will be a mismatch between 1. and 2., so it is still error prone, just like the current situation.

dynasource:
What about combining our solutions? By combining it seems that we cover more usecases.

rob006:
> You will get a different widgets count if multiple Pjax widgets reside in both layouts & views:
1. original page: during the load of the non-ajax page, all Pjax widgets are loaded, from both layout & views
2. ajax page: developers choose to load only the view. Not all Pjax widgets are loaded. 

This is covered by my proposal - since view is rendered before layout, it will always have the same ID sequence (started from `p0`) regardless layout.

And I still don't get it how do you want to have different logic (rendering view with and without layout) with the same backtrace. Could you share code of example action which do that?

dynasource:
> This is covered by my proposal - since view is rendered before layout, it will always have the same ID sequence (started from p0) regardless layout.

Correct, but it is still reasoned from the current situation, which has shown error prone because of the counting algorithm. Of course your solution is already way better, it is still not perfect. If somebody skips 1 Pjax widget during the Ajax call (for whatever reason), it will throw an error.

> And I still don't get it how do you want to have different logic (rendering view with and without layout) with the same backtrace. Could you share code of example action which do that?

And I still have a hard time processing your shallow statements, and not for the first time https://github.com/yiisoft/yii2/pull/12977#issuecomment-260128810. Perhabs I could, but then I will have to put a lot of time & effort to interpretate it right, while it should be you to put more effort in supporting your statements with arguments and/or explanations (if you want to be heard of course)

rob006:
> Of course your solution is already way better, it is still not perfect.

I'm not going to prove that my solution is perfect, but that it is at least the same good as your. And since it doesn't use ugly hack with debug function, it is better option here.

> And I still have a hard time processing your shallow statements

You have code like this:

```php
public function actionView() {
    // some code

    if (Yii::$app->request->isAjax) {
        return $this->renderPartial('view');
    } else {
        return $this->render('view');
    }
}
```

If you use pjax widget in `view`, you will have different IDs on regular request and AJAX request, because they will have different backtrace. Am I right?

dynasource:
>And since it doesn't use ugly hack with debug function, it is better option here.

This PR is to approach the issue from a different angle: the 'unique' line where the PJAX widget is called. 
Is it uncommon, yes, but is it ugly? Not in this context as it is the main key of the solution. 

>  I'm not going to prove that my solution is perfect, but that it is at least the same good as your. 

Never said it wasn't. It looks likes it is covering more usecases than the solution in this PR. 

Therefore it is a pity that you did not take the effort to elaborate on it. It would have been helpful if you had shown the strengths of your solution instead of a practicing a negative communication style.

> You have code like this:

This is one of the usecases indeed. This example is covered by both of our solutions.

samdark:
So overall it seems there are two cases:

1. When you want ID to always be the same.
2. When you want ID to always be unique.

Only one of two could be a default and it seems when it's `#2` there's less trouble.

rob006:
> This is one of the usecases indeed. This example is covered by both of our solutions.

I don't get it. How this example is covered by your solution? You will have the same content with different ID. Shouldn't pjax IDs from different request match each other?

samdark:
The ideal solution would be forcing user to specify ID but that would be a BC break...

dynasource:
agree. I assessed your example on the criterium 'uniqueness' but I should have assessed it on 'matching'. Didn't yet notice that this was a multi criteria issue, https://github.com/yiisoft/yii2/pull/12977#issuecomment-260186201

dynasource:
> The ideal solution would be forcing user to specify ID but that would be a BC break...

well, thought that too for while, but then I was thinking about usecases in which people are reusing the same PjaxWidget with different content (ie. pjax layouts). Then they have to supply unique ID's all the time. This is also cumbersome. 

The ideal situation is to have this Pjax widget to be unobtrusive, like in the example in the guide. In the end, this ID doesnt really matter to the developer. He should not be bothered with it.

dynasource:
I've thought it over. The approach in this PR has not proven to cover all usecases of which some common. 

> I'm not going to prove that my solution is perfect, but that it is at least the same good as your.

Perhabs not perfect, but a better approach indeed. It covers more usecases & is less complex & more consistent with current architecture. Therefore this PR is to be closed in favor of that solution. @rob006, are you willing to create a new PR?

rob006:
> @rob006, are you willing to create a new PR?

No, feel free to take it.

PR #12983, opened 2016-11-12
Phpdoc fixes

rob006:
Some minor fixes reported by scrutinizer.

### DIFF
--- a/framework/base/ErrorHandler.php
+++ b/framework/base/ErrorHandler.php
@@ -196,7 +196,7 @@ abstract class ErrorHandler extends Component
      * @param string $message the error message.
      * @param string $file the filename that the error was raised in.
      * @param int $line the line number the error was raised at.
-     * @return bool whether the normal error handler continues.
+     * @return false whether the normal error handler continues.
      *
      * @throws ErrorException
      */
--- a/framework/base/ErrorHandler.php
+++ b/framework/base/ErrorHandler.php
@@ -196,7 +196,7 @@ abstract class ErrorHandler extends Component
      * @param string $message the error message.
      * @param string $file the filename that the error was raised in.
      * @param int $line the line number the error was raised at.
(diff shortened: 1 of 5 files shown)


framework/base/ErrorHandler.php
```diff
@@ -196,7 +196,7 @@ public function handleHhvmError($code, $message, $file, $line, $context, $backtr
      * @param string $message the error message.
      * @param string $file the filename that the error was raised in.
      * @param int $line the line number the error was raised at.
-     * @return bool whether the normal error handler continues.
+     * @return false whether the normal error handler continues.
```
rob006: `handleError()` returns only `false`, which makes this return value useless. Same as this check: https://github.com/rob006/yii2-dev/blob/450eefaa50fc8aae1f9137a6fdf7ed2e5e9fe9e1/framework/base/ErrorHandler.php#L177

Am I missing something?

dynasource: I assume this is done to prevent malfunctioning after future changes. The return value is important for set_error_handler https://github.com/yiisoft/yii2/blob/master/framework/base/ErrorHandler.php#L65. 

According to http://php.net/manual/en/function.set-error-handler.php, `If the function returns FALSE then the normal error handler continues.` When the behavior of this custom errorHandler will change (for whatever reason in the future) and return a true value, the HHVM logic will still function correctly.

However, I would vote to cover this in tests and apply your suggestions.

dynasource (approved):

framework/base/ErrorHandler.php
cebe: `@return bool whether the normal error handler continues.` is the siguature required by php, so we should not change it to `false` just because our implementation returns false only.

cebe (changes requested):

cebe (approved):

SilverFire:
Thank you!

PR #12984, opened 2016-11-12
Disable serialization by default for ArrayCache for increased performance

rhertogh:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | -

I would expect the ArrayCache to be optimized for speed since it's only used during a single requests, but the serialization is turned on by default which has a major performance impact (see test below).
Turning the serialization off might brake backwards compatibilty due to the check in the yii/caching/Cache class: ```if ($dependency !== null && $this->serializer !== false) {```

Test results:
unserialized: 9.3 ms
serialized: 3391.6 ms

Test data: two dimensional array (100x100)

Test run: 100 x 1 write, 10 reads

Test:

```
$data = array_fill(0, 100, array_fill(0, 100, rand(0, 10000)));

Yii::$app->arrayCache->serializer = false;
Yii::beginProfile('cacheTest-unserialized');
for ($i = 1; $i <= 100; $i++) {
    Yii::$app->arrayCache->set("cacheTest-unserialized-$i", $data);
    for ($j = 1; $j <= 10; $j++) {
        Yii::$app->arrayCache->get("cacheTest-unserialized-$i");
    }
}
Yii::endProfile('cacheTest-unserialized');

Yii::$app->arrayCache->serializer = null;
Yii::beginProfile('cacheTest-serialized');
for ($i = 1; $i <= 100; $i++) {
    Yii::$app->arrayCache->set("cacheTest-serialized-$i", $data);
    for ($j = 1; $j <= 10; $j++) {
        Yii::$app->arrayCache->get("cacheTest-serialized-$i");
    }
}
Yii::endProfile('cacheTest-serialized');
```

### DIFF
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -50,7 +50,15 @@ if you want to upgrade from version A to version C and there is
 version B between A and C, you need to follow the instructions
 for both A and B.
 
+Upgrade from Yii 2.0.10
+-----------------------
 
+* The `serializer` property for `yii\caching\ArrayCache` has been changed form `null` to `false`, this disables serialization
+  by default for increased performance. However, this also means that the `dependency` parameter for `yii\caching\Cache::set()`
+  and `yii\caching\Cache::multiSet()` will be ignored by default. If you are using `yii\caching\ArrayCache` in combination with
+  dependencies you should set `yii\caching\ArrayCache::$serializer` to `null`.
+  
+  
 Upgrade from Yii 2.0.9
 ----------------------
 
--- a/framework/caching/ArrayCache.php
+++ b/framework/caching/ArrayCache.php
@@ -23,10 +23,9 @@ namespace yii\caching;
 class ArrayCache extends Cache
 {
     /**
-     * Serialization is disabled by default for ArrayCache for increased performance
+     * Serialization is disabled by default for [[ArrayCache]] for increased performance
      *
-     * @var null|array|false The functions used to serialize and unserialize cached data.
-     * @see Cache::serializer
+     * @inheritdoc
      */
     public $serializer = false;
     
--- a/framework/caching/ArrayCache.php
+++ b/framework/caching/ArrayCache.php
@@ -22,6 +22,14 @@ namespace yii\caching;
  */
 class ArrayCache extends Cache
 {
+    /**
+     * Serialization is disabled by default for ArrayCache for increased performance
+     *
+     * @var null|array|false The functions used to serialize and unserialize cached data.
+     * @see Cache::serializer
+     */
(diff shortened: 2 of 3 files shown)


SilverFire:
Makes sense. Voting for.

framework/caching/ArrayCache.php
```diff
@@ -22,6 +22,14 @@
  */
 class ArrayCache extends Cache
 {
+    /**
+     * Serialization is disabled by default for ArrayCache for increased performance
```
SilverFire: `ArrayCache` -> `[[ArrayCache]]`

framework/caching/ArrayCache.php
```diff
@@ -22,6 +22,14 @@
...
+     *
+     * @var null|array|false The functions used to serialize and unserialize cached data.
```
SilverFire: This line should be replaced with `@inheritdoc`

framework/caching/ArrayCache.php
```diff
@@ -22,6 +22,14 @@
...
+     * @see Cache::serializer
```
SilverFire: Line could be removed

SilverFire (changes requested):
Change should be mentioned in `CHANGELOG.md` and `UPGRADE.md`

dynasource:
agree. Optimal performance should be the default.

framework/UPGRADE.md
```diff
@@ -50,7 +50,15 @@ if you want to upgrade from version A to version C and there is
 version B between A and C, you need to follow the instructions
 for both A and B.
 
+Upgrade from Yii 2.0.10
+-----------------------
 
+* The `serializer` property for `yii\caching\ArrayCache` has been changed form `null` to `false`, this disables serialization
```
samdark: Sounds like total backwards compatibility break to me...

SilverFire: It's only for ArrayCache. All other types of caches will not be affected

samdark: Yes but still ArrayCache won't work in many cases anymore.

cebe:
No need to break anything here. This can be added as a note in the class PHPdoc and can be enabled if needed. ArrayCache has very rare use cases anyway.

samdark:
I'm against merging this change. It makes default behavior very inconsistent. Instead, I'd mention it in either performance guide or the ArrayCache class itself.

cebe:
I have added a note to the class docs. performance improvement is not an acceptable reason for BC breaks. Thanks for your contribution, @rhertogh!

PR #12986, opened 2016-11-13
Fix some phpdoc errors reported by apidoc extension

rob006:
Before:

> 129 errors have been logged to [...]/yii2api/errors.txt

After:

> 81 errors have been logged to [...]/yii2api/errors.txt

### DIFF
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -13,7 +13,7 @@ use Yii;
  * Component is the base class that implements the *property*, *event* and *behavior* features.
  *
  * Component provides the *event* and *behavior* features, in addition to the *property* feature which is implemented in
- * its parent class [[Object]].
+ * its parent class [[\yii\base\Object|Object]].
  *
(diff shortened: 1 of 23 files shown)


framework/base/Component.php
```diff
@@ -13,7 +13,7 @@
  * Component is the base class that implements the *property*, *event* and *behavior* features.
  *
  * Component provides the *event* and *behavior* features, in addition to the *property* feature which is implemented in
- * its parent class [[Object]].
+ * its parent class [[\yii\base\Object|Object]].
```
cebe: is `Object` not resolved in `yii\base` namespace? if not I'd consider that a bug in apidoc generator.

cebe (approved):

framework/base/Component.php
rob006: `object` is reserved keyword in PHP.

dynasource (approved):
useful improvements. Couldn't find any flaws. Well done.

samdark (approved):

cebe:
Thank you!

Issue #12980, opened 2016-11-12
Wrong migration history order


evgen-d:
console config:
```php
'controllerMap' => [
    'migrate' => [
        'class' => 'yii\console\controllers\MigrateController',
        'migrationNamespaces' => [
            'app\migrations',
            'app\modules\mymodule\migrations',
        ],
        'migrationPath' => null,
    ],
],
```
apply some migrations:
```
app\migrations\M140506102106One
app\migrations\M160909083544Two
app\migrations\M160930135248Three
app\modules\mymodule\migrations\M161018124749Four
app\modules\mymodule\migrations\M161025123028Five
app\migrations\M161110133341Six
```
run `./yii migrate/history`
```
Showing the last 6 applied migrations:
    (2016-11-12 16:53:04) app\migrations\M161110133341Six
    (2016-11-12 16:53:04) app\modules\mymodule\migrations\M161025123028Five
    (2016-11-12 16:53:03) app\modules\mymodule\migrations\M161018124749Four
    (2016-11-12 16:52:20) app\migrations\M160930135248Three
    (2016-11-12 16:52:17) app\migrations\M160909083544Two
    (2016-11-12 16:52:16) app\migrations\M140506102106One
```
then run `./yii migrate/history 1`

### What is the expected result?
```
Showing the last 1 applied migration:
    (2016-11-12 16:53:04) app\migrations\M161110133341Six
```
### What do you get instead?
```
Showing the last 1 applied migration:
    (2016-11-12 16:53:04) app\modules\mymodule\migrations\M161025123028Five
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | master
| PHP version      | 7.0
| Operating system | Fedora 24

SilverFire:
@evgen-d would you like to fix this one?

evgen-d:
@SilverFire i think that `version` and `namespace` must be on different columns of DB table, but this can break BC

SilverFire:
@yiisoft/core-developers

samdark:
Seems to be related to https://github.com/yiisoft/yii2/issues/12974?

evgen-d:
It related to, but not the result of this issue

klimov-paul:
Resolved by commit 1157657cab3c3ef867a5bd929b893e44f5f2dbcc

PR #12975, opened 2016-11-11
Migrations order fix

devgen-io:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | #12974

PR #12989, opened 2016-11-13
Fix migration history order

devgen-io:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #12980

### DIFF
--- a/framework/console/controllers/MigrateController.php
+++ b/framework/console/controllers/MigrateController.php
@@ -200,27 +200,14 @@ class MigrateController extends BaseMigrateController
             $this->createMigrationHistoryTable();
         }
         $query = new Query;
-        $rows = $query->select(['version', 'apply_time'])
+        $rows = $query->select(['namespace', 'apply_time'])
             ->from($this->migrationTable)
-            ->orderBy('apply_time DESC, version DESC')
+            ->orderBy('[[apply_time]] DESC, [[version]] DESC, [[namespace]] DESC')
             ->limit($limit)
             ->createCommand($this->db)
             ->queryAll();
-        $history = [];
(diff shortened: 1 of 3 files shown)


SilverFire:
Thank you, this one looks good in general.

How do you suggest to update the `migration` table in case it already exists? Write a new migration? :)

evgen-d:
I don't know, may be.  But now namespaced migrations working incorrect and can break DB data.

evgen-d:
We have to abandon the namespaced migration in our project. I advise others to do the same

SilverFire:
Asking for @yiisoft/core-developers opinions on how to implement migration for migration table.

Suggest to check, whether `namespace` column exists, and if it does not - add it.

rob006:
We could solve this with additional query:

1. Query last X migrations.
2. Query all migrations with `apply_time` equals to results from previous query.
3. Sort them internally ignoring namespace and pick first X results.

samdark:
@klimov-paul need your input.

klimov-paul:
> @klimov-paul need your input.

There is no other option, but removing the limit from the query and use `array_slice()` instead after the sort.
It should be safe in general, since migration history unlikely stores too may records and we use array as result set.

Would be better of course if you have waited for 'my input' before merging #12975, but alright.

I am closing this PR as its solution unacceptable at current state.

samdark:
Should #12975 be reverted?

klimov-paul:
It does not matter: it is too late to do `reset --hard`, which I will override the changes during the fix anyway.

samdark:
OK.

PR #12990, opened 2016-11-13
Default cache ttl (duration) property (fixes #6809)

### LINKED ISSUE #6809: Add application-wide cache duration parameter
I think it could be implemented for convenience.

As of now it falls back to `0 means never expire` if value is not specified when calling `set` - thow this value is not practical.
Setting it each time on each `set` is tedious, while overriding core classes only just for it...

To keep BC it may be set 0 by default in `\yii\caching\Cache`

```
    public function set($key, $value, $duration = null, $dependency = null)
    {
        if ($duration === null) {
             $duration = $this->defaultDuration;
        }
        ...
        return $this->setValue($key, $value, $duration);
    }
```

SDKiller:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #6809 

Some PROS in addition to discussion in #6809

- from the prorposed https://github.com/php-fig/fig-standards/blob/master/proposed/simplecache.md
> If no value is sent and the driver supports TTL then the library may set a default value for it or let the driver take care of that.

Althow it is not obvious from the draft, that default value cannot be hardcoded, IMO everyone expects it should be configurable.

- there are examples of configurable defaults among other frameworks, e.g. https://github.com/zendframework/zend-cache/blob/master/src/Storage/Adapter/AdapterOptions.php#L66

BC: to keep BC default value is 0, proposed change should not affect implementations, which extend `\yii\caching\Cache`.

Note about tests: 
as of now all tests for core cache classes do not suppose setting cache configuration or passing params to `\yiiunit\framework\caching\DbCacheTest::getCacheInstance`

### DIFF
--- a/docs/guide/caching-data.md
+++ b/docs/guide/caching-data.md
@@ -204,6 +204,9 @@ if ($data === false) {
 }
 ```
 
+Since 2.0.11 you may set [[yii\caching\Cache::$ttl|ttl]] value in your cache component configuration if you need default cache duration other than 0.
+This will allow you not to pass custom `duration` parameter to [[yii\caching\Cache::set()|set()]] each time.
+
 
 ### Cache Dependencies <span id="cache-dependencies"></span>
 
--- a/docs/guide/caching-data.md
+++ b/docs/guide/caching-data.md
@@ -204,7 +204,8 @@ if ($data === false) {
 }
 ```
 
-Since 2.0.11 you may set [[yii\caching\Cache::$ttl|ttl]] value in your cache component configuration if you need default cache duration other than 0.
+Since 2.0.11 you may set [[yii\caching\Cache::$ttl|ttl]] value in your cache component configuration if you prefer a custom cache duration
+over the default unlimited duration.
 This will allow you not to pass custom `duration` parameter to [[yii\caching\Cache::set()|set()]] each time.
 
 
--- a/framework/caching/Cache.php
+++ b/framework/caching/Cache.php
@@ -71,6 +71,14 @@ abstract class Cache extends Component implements \ArrayAccess
      * implementations of the cache can not correctly save and retrieve data different from a string type.
      */
     public $serializer;
+    /**
+     * @var integer default ttl of cached value (the number of seconds in which the cached value will expire) -
+     * if it was not defined explicitly in [[set()]].
+     * Default value is 0, depending on implementation which means never expire or maximum ttl supported by cache engine
+     * (e.g. it is 1 year for [[\yii\caching\FileCache]]).
+     * @since 2.0.11
+     */
+    public $ttl = 0;
 
 
     /**
@@ -202,14 +210,19 @@ abstract class Cache extends Component implements \ArrayAccess
      * @param mixed $key a key identifying the value to be cached. This can be a simple string or
      * a complex data structure consisting of factors representing the key.
      * @param mixed $value the value to be cached
-     * @param int $duration the number of seconds in which the cached value will expire. 0 means never expire.
+     * @param int $duration the number of seconds in which the cached value will expire. If not set - the value of [[ttl]] is used,
+     * which defaults to 0 (never expire).
      * @param Dependency $dependency dependency of the cached item. If the dependency changes,
      * the corresponding value in the cache will be invalidated when it is fetched via [[get()]].
      * This parameter is ignored if [[serializer]] is false.
      * @return bool whether the value is successfully stored into cache
      */
-    public function set($key, $value, $duration = 0, $dependency = null)
+    public function set($key, $value, $duration = null, $dependency = null)
     {
+        if ($duration === null) {
+            $duration = $this->ttl;
+        }
+
         if ($dependency !== null && $this->serializer !== false) {
             $dependency->evaluateDependency($this);
         }
--- a/framework/caching/Cache.php
+++ b/framework/caching/Cache.php
@@ -72,10 +72,7 @@ abstract class Cache extends Component implements \ArrayAccess
      */
     public $serializer;
     /**
-     * @var integer default ttl of cached value (the number of seconds in which the cached value will expire) -
-     * if it was not defined explicitly in [[set()]].
-     * Default value is 0, depending on implementation which means never expire or maximum ttl supported by cache engine
-     * (e.g. it is 1 year for [[\yii\caching\FileCache]]).
+     * @var integer default duration in seconds before the cache will expire. Default value is 0, meaning infinity.
      * @since 2.0.11
      */
     public $ttl = 0;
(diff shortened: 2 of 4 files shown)


docs/guide/caching-data.md
```diff
@@ -204,6 +204,9 @@ if ($data === false) {
 }
 ```
 
+Since 2.0.11 you may set [[yii\caching\Cache::$ttl|ttl]] value in your cache component configuration if you need default cache duration other than 0.
```
dynasource: 'if you prefer a custom cache duration over the default unlimited duration'

framework/caching/Cache.php
```diff
@@ -71,6 +71,14 @@
      * implementations of the cache can not correctly save and retrieve data different from a string type.
      */
     public $serializer;
+    /**
+     * @var integer default ttl of cached value (the number of seconds in which the cached value will expire) -
```
dynasource: @var integer default duration in seconds before the cache will expire. Default value is 0, meaning infinity.

framework/caching/Cache.php
```diff
@@ -202,14 +210,19 @@ public function multiGet($keys)
      * @param mixed $key a key identifying the value to be cached. This can be a simple string or
      * a complex data structure consisting of factors representing the key.
      * @param mixed $value the value to be cached
-     * @param int $duration the number of seconds in which the cached value will expire. 0 means never expire.
+     * @param int $duration the number of seconds in which the cached value will expire. If not set - the value of [[ttl]] is used,
```
dynasource: @param int default duration in seconds before the cache will expire. Default value is 0, meaning infinity.

framework/caching/Cache.php
```diff
@@ -202,14 +210,19 @@ public function multiGet($keys)
...
+     * which defaults to 0 (never expire).
      * @param Dependency $dependency dependency of the cached item. If the dependency changes,
      * the corresponding value in the cache will be invalidated when it is fetched via [[get()]].
      * This parameter is ignored if [[serializer]] is false.
      * @return bool whether the value is successfully stored into cache
      */
-    public function set($key, $value, $duration = 0, $dependency = null)
+    public function set($key, $value, $duration = null, $dependency = null)
```
dynasource: this is a BC break. Should be kept 0

framework/caching/Cache.php
```diff
@@ -202,14 +210,19 @@ public function multiGet($keys)
...
     {
+        if ($duration === null) {
```
dynasource: === 0

tests/framework/caching/CacheTestCase.php
```diff
@@ -177,6 +177,13 @@ public function testMget()
         $this->assertEquals(['number_test' => 42, 'non_existent_key' => null], $cache->mget(['number_test', 'non_existent_key']));
     }
 
+    public function testDefaultTtl()
+    {
+        $cache = $this->getCacheInstance();
+
+        $this->assertEquals(0, $cache->ttl);
```
dynasource: assertSame

dynasource (changes requested):
thanks for picking this PR up after your holiday. 

PR is a good start. Have some requests.

framework/caching/Cache.php
SDKiller: Then this will prevent another case: say you have default `ttl = 3600` and for one single item want to set infinite duration explicitly in `set()`.

framework/caching/Cache.php
dynasource: developers may have overrided it

framework/caching/Cache.php
dynasource: is there an alternative solution without that BC https://github.com/yiisoft/yii2/pull/12990#discussion_r87715973

framework/caching/Cache.php
SDKiller: Then we have a problem, see https://github.com/yiisoft/yii2/pull/12990#discussion_r87715812

dynasource: what about func_get_args(). Have you tried that?

SDKiller: Ok, I'll do it. A little bit tricky solution, probably comment should be added to explain it.

framework/caching/Cache.php
```diff
@@ -210,16 +207,17 @@ public function multiGet($keys)
...
      * This parameter is ignored if [[serializer]] is false.
      * @return bool whether the value is successfully stored into cache
      */
-    public function set($key, $value, $duration = null, $dependency = null)
+    public function set($key, $value, $duration = 0, $dependency = null)
     {
-        if ($duration === null) {
+        $args = func_get_args();
+        if (!isset($args[2])) {
+            // https://github.com/yiisoft/yii2/pull/12990#discussion_r87715643
```
rob006: `set('key', 'value')` will give you different result than `set('key', 'value', 0)`? :fearful:

SDKiller: The same if your default `ttl` value is 0.
Different if your default `ttl` value is different.

SDKiller: @rob006 
See discussion from https://github.com/yiisoft/yii2/pull/12990#discussion_r87715643

As for me I'd also prefer `null` as it makes things straightforward and most probably will hurt nobody.

SDKiller: @rob006 
Assuming that `A` is `yii\caching\Cache` and `B` is 3rd-party class - seems correct test should be:

https://3v4l.org/rFqFh

samdark (approved):

framework/caching/Cache.php
```diff
@@ -202,14 +207,20 @@ public function multiGet($keys)
...
+     * @param int $duration default duration in seconds before the cache will expire. Default value is 0, meaning infinity.
      * @param Dependency $dependency dependency of the cached item. If the dependency changes,
      * the corresponding value in the cache will be invalidated when it is fetched via [[get()]].
      * This parameter is ignored if [[serializer]] is false.
      * @return bool whether the value is successfully stored into cache
      */
     public function set($key, $value, $duration = 0, $dependency = null)
     {
+        $args = func_get_args();
+        if (!isset($args[2])) {
```
cebe: this looks really strange. I'd prefer making `$duration = null` the default and if it is `null` use `$this->ttl`.

cebe (changes requested):

framework/caching/Cache.php
samdark: @cebe wouldn't that be a BC break because of signature change?

framework/caching/Cache.php
cebe: it does not break BC as it does not change the signature.

framework/caching/Cache.php
dynasource: this trick was designed to prevent a BC break, see https://github.com/yiisoft/yii2/pull/12990#discussion_r87715643.
But since it isn't, as values dont seem to change a signature, the original code of SDKiller was already fine. Just learned something new again (at your cost @SDKiller, sorry for that ;)

samdark: Thanks for pointing that out. Then `null` as default value is the best choice. @SDKiller

SDKiller: Done https://github.com/yiisoft/yii2/pull/12990/commits/22e9905203f349d184b2389f562aef1a4bb586ce.
Seems github now messed all comments :(

dynasource (approved):

samdark (approved):

cebe (approved):

dynasource:
thanks!

SDKiller:
@dynasource 

Something is wrong with changelog.
Also changes from https://github.com/yiisoft/yii2/commit/1157657cab3c3ef867a5bd929b893e44f5f2dbcc were lost.

cebe:
fixed changelog in 3913e6ebcf0aa750d3248d64e1696411bdfd41b1
@SDKiller I do not see how change from 1157657 are lost. you mean only in CHANGLOG or other changes too?

SDKiller:
I meant changelog, but they actually were not lost, just went up the list.

dynasource:
thanks for notifying. It is fixed now.

kidol:
@SDKiller Why no more `$defaultDuration` like in original issue? :cry:

cebe:
> Why no more $defaultDuration like in original issue? :cry:

I like that name more than just `ttl`. it also fits the overall terminology better because `duration` is used in most caching docs and code instead of ttl. @yiisoft/core-developers opinions?

SDKiller:
@kidol 
It's about 2 years since proposal.
And probably I've read too much php-fig docs :)

@cebe 
As you wish, I'm not stuck to this name.

dynasource:
$defaultDuration is better IMO. It's easier to understand & more consistent with the description and the overall terminology.

cebe:
done. 059a80b

PR #12999, opened 2016-11-14
two primarykey one value default find first value

philpm:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

when table have two primarykey,I can't get model with restful url like this http://www.a.com/api/orders/5
bootstrap set 'GET api/orders/<id>' => 'appapi/orders/view',

### DIFF
--- a/framework/rest/Action.php
+++ b/framework/rest/Action.php
@@ -92,6 +92,8 @@ class Action extends \yii\base\Action
             $values = explode(',', $id);
             if (count($keys) === count($values)) {
                 $model = $modelClass::findOne(array_combine($keys, $values));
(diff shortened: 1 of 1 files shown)


cebe:
> when table have two primarykey,I can't get model with restful url like this http://www.a.com/api/orders/5
> bootstrap set 'GET api/orders/<id>' => 'appapi/orders/view',

of course not, because you need to specify both parts of the key to get a valid result. specifying only one part is ambiguous and may return a wrong record. Can not accept this change.

PR #13002, opened 2016-11-14
Allow resetting dependency registered as singleton

dmitrivereshchagin:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

If `null` passed to `yii\di\Container::setSingleton()` as dependency definition, next call to `yii\di\Container::get()` will generate new instance of dependency.

### DIFF
--- a/framework/di/Container.php
+++ b/framework/di/Container.php
@@ -265,6 +265,7 @@ class Container extends Component
      *
      * @param string $class class name, interface name or alias name
      * @param mixed $definition the definition associated with `$class`. See [[set()]] for more details.
+     * If `null`, next call to [[get()]] behaves in the same way as if the instance was never created.
      * @param array $params the list of constructor parameters. The parameters will be passed to the class
      * constructor when [[get()]] is called.
      * @return $this the container itself
@@ -272,9 +273,13 @@ class Container extends Component
      */
     public function setSingleton($class, $definition = [], array $params = [])
     {
-        $this->_definitions[$class] = $this->normalizeDefinition($class, $definition);
-        $this->_params[$class] = $params;
-        $this->_singletons[$class] = null;
+        if ($definition !== null) {
+            $this->_definitions[$class] = $this->normalizeDefinition($class, $definition);
+            $this->_params[$class] = $params;
+            $this->_singletons[$class] = null;
+        } elseif (isset($this->_singletons[$class])) {
+            $this->_singletons[$class] = null;
+        }
         return $this;
     }
 
--- a/tests/framework/di/ContainerTest.php
+++ b/tests/framework/di/ContainerTest.php
@@ -108,6 +108,39 @@ class ContainerTest extends TestCase
         $this->assertEquals(4, $qux->a);
     }
 
+    public function testSetSingleton()
+    {
+        $namespace = __NAMESPACE__ . '\stubs';
+        $QuxInterface = "$namespace\\QuxInterface";
+        $Qux = Qux::className();
+
+        $container = new Container;
+
+        $container->setSingleton($QuxInterface, null);
+        $this->assertFalse($container->hasSingleton($QuxInterface));
+
+        $container->setSingleton($QuxInterface, $Qux);
+        $this->assertTrue($container->hasSingleton($QuxInterface));
+        $this->assertFalse($container->hasSingleton($QuxInterface, true));
+
+        $qux = $container->get($QuxInterface);
+        $this->assertTrue($qux instanceof $Qux);
+        $this->assertTrue($container->hasSingleton($QuxInterface, true));
+
+        $qux2 = $container->get($QuxInterface);
+        $this->assertTrue($qux === $qux2);
+
+        $container->setSingleton($QuxInterface, null);
+        $this->assertTrue($container->hasSingleton($QuxInterface));
+        $this->assertFalse($container->hasSingleton($QuxInterface, true));
+
+        $qux3 = $container->get($QuxInterface);
+        $this->assertFalse($qux2 === $qux3);
+
+        $qux4 = $container->get($QuxInterface);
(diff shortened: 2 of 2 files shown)


dynasource:
can you elaborate on your usecase?

dmitrivereshchagin:
I simply want to implement `Doctrine\Common\Persistence\ManagerRegistry` interface using `yii\di\Container` to store DBAL connections and entity managers. These changes might be useful to implement `Doctrine\Common\Persistence\ManagerRegistry::resetManager()` for example.

framework/di/Container.php
```diff
@@ -265,16 +265,21 @@ public function set($class, $definition = [], array $params = [])
...
     {
-        $this->_definitions[$class] = $this->normalizeDefinition($class, $definition);
-        $this->_params[$class] = $params;
-        $this->_singletons[$class] = null;
+        if ($definition !== null) {
+            $this->_definitions[$class] = $this->normalizeDefinition($class, $definition);
+            $this->_params[$class] = $params;
+            $this->_singletons[$class] = null;
+        } elseif (array_key_exists($class, $this->_singletons)) {
+            $this->_singletons[$class] = null;
```
cebe: as far as I see there is no need for the elseif and `$this->_singletons[$class] = null;` can be performed in any case.

dmitrivereshchagin: I think there is no need to create `$class` key when there is no dependency registered.

SilverFire:
Could you add a unit test for this PR?

dynasource:
I see that you've replaced a lot of tests. Could you move these improments in a separate PR please.

dmitrivereshchagin:
@SilverFire test for proposed behavior is already here 🙂

dmitrivereshchagin:
@dynasource unnecessary changes in assertions removed.

tests/framework/di/ContainerTest.php
```diff
@@ -108,6 +108,25 @@ public function testDefault()
         $this->assertEquals(4, $qux->a);
     }
 
+    public function testSingleton()
```
dynasource: should be testSetSingleton

tests/framework/di/ContainerTest.php
```diff
@@ -108,6 +108,25 @@ public function testDefault()
...
+    {
+        $namespace = __NAMESPACE__ . '\stubs';
+        $QuxInterface = "$namespace\\QuxInterface";
+        $Qux = Qux::className();
+
+        $container = new Container;
+        $container->setSingleton($QuxInterface, $Qux);
+        $qux = $container->get($QuxInterface);
+        $this->assertTrue($qux instanceof $Qux);
+        $qux2 = $container->get($QuxInterface);
```
dynasource: please separate groups of assertions with line breaks.

framework/di/Container.php
```diff
@@ -265,16 +265,21 @@ public function set($class, $definition = [], array $params = [])
...
     public function setSingleton($class, $definition = [], array $params = [])
     {
-        $this->_definitions[$class] = $this->normalizeDefinition($class, $definition);
-        $this->_params[$class] = $params;
-        $this->_singletons[$class] = null;
+        if ($definition !== null) {
+            $this->_definitions[$class] = $this->normalizeDefinition($class, $definition);
+            $this->_params[$class] = $params;
+            $this->_singletons[$class] = null;
+        } elseif (isset($this->_singletons[$class])) {
```
dynasource: this should be combined with the previous comparison to make it more readable. Are you sure this change of logic is not a BC break?

dynasource (changes requested):

tests/framework/di/ContainerTest.php
dmitrivereshchagin: Fixed

tests/framework/di/ContainerTest.php
dmitrivereshchagin: Fixed

framework/di/Container.php
dmitrivereshchagin: Please see my reply to @cebe. There is no BC break. I have added some checks for `hasSingleton()` to be sure.

dynasource: can you combine the comparisons?

dmitrivereshchagin: No. Null assignment should take place only in two cases:

* normal singleton configuration
* resetting singleton (singleton is configured *and* `null` passed as an argument)

dynasource: so this is not correct?:

```php
if($definition === null && isset($this->_singletons[$class]){

}else{

}

```

dmitrivereshchagin: It will set a key in `$this->_singletons` in any case and break `hasSingleton()`.

dmitrivereshchagin:
It may be better to implement proposed behavior as a separate method. We can call it `resetSingleton()` and put in container class.
```php
public function resetSingleton($class)
{
    if (!array_key_exists($class, $this->_singletons) {
        throw new InvalidConfigException();
    }

    $this->_singletons[$class] = null;

    return $this;
}
```
Besides of that such method can be implemented (slightly less efficiently) in subclass.
```php
public function resetSingleton($class)
{
    if (!$this->hasSingleton($class)) {
        throw new InvalidConfigurationException();
    }

    $definitionitions = $this->getDefinitions();
    $this->setSingleton($class, $definitions[$class]);

    return $this;
}
```

dynasource:
separate method is better. What's the use of the exception? Can you just reset the value if it exists.

dmitrivereshchagin:
It could point out an error in application configuration or usage of obsolete service.

dmitrivereshchagin:
There is no way to introduce this feature without breaking BC. In the case when dependency and its definition are the same object singleton cannot be reset in reliable way.

PR #13003, opened 2016-11-14
Fix incorrect return type in phpdocs [skip ci]

SDKiller:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

Internally used `setValues()`, `addValues()` methods return array


SilverFire:
Thank you!

Issue #12993, opened 2016-11-14
Skip CI or js tests depending on changed files


arogachev:
Even we can include `[skip ci]` in commit message to skip Travis build, many users do not know about it or just forget to add it (I found myself forget to add it multiple times when editing the docs).

Even it's not something crucial, we can have a custom hook to scan changed files, and if it contains only `.md` files - add `[skip ci]` in case if it's missing. But there is also `[ci skip]` alias and skipping build can not be limited to editing only `.md` files. Such changes can involve PHPDoc comments, regular comments, etc.

But even adding it only for `.md` related changes can be useful.

In the same manner we can analyze and automatically add `[skip js]` (#12992) in case there are no `.js` files in changed files list.

**P.S.** However maybe it's not worth implementing and it's just me overcomplicating things.

This is just a raw idea, decided to share it.

samdark:
It's nice to have and we can accept a pull request with it. Any idea about implementation?

arogachev:
The current idea is to use custom git hook and somehow pass this info to Travis, via environment variable maybe. Will try to think of a better solution.

SDKiller:
https://github.com/yiisoft/yii2/issues/12992#issuecomment-260281581

> It will pollute log messages much. Is there a way to do it w/o marking commit messages with it?

For js - do it vice versa (not mark everything with `[skip js]` but to mark only nesessary with `[test js]` for example)

arogachev:
@SDKiller Yes, initially I wanted to suggest this option as alternative, but forgot it.

SDKiller:
Another thing regarding js.
I recall some time ago it was suggested to separate all js stuff from framework in 2.1 (@creocoder suggested, if I'm not mistaken).
I suppose it will also be helpful for testing purposes - as different travis setup could be configured for this separate repo.

arogachev:
Yes, it will be even better and easier for testing and Travis CI setup. We can register bower / npm package and include it along with other dependencies.

dynasource:
@arogachev, what's the status of this topic?

cebe:
I do not see a reliable way to trigger this. There must be some kind of native support from travis to support this as far as I see.

rob006:
Script that prints files modified by commit/PR: https://github.com/melpa/melpa/blob/master/travis-changed-files
Related bug: https://github.com/travis-ci/travis-ci/issues/5133

It does not look so hard to do. We could grep results from this script, filter by file extension and:
[...]

PR #13005, opened 2016-11-15
Closed #12778: Added docs for multiple attributes validation

### LINKED ISSUE #12778: Adding common validation error (not related with particular attribute)

arogachev:
Spent some time to come up with real life example, I think it's better than abstract `attributeA`, etc. 

If everything is OK, I can add russian translation as well.

Will try to extend with more tips.

### DIFF
--- a/docs/guide/input-validation.md
+++ b/docs/guide/input-validation.md
@@ -482,7 +482,7 @@ class MigrationForm extends \yii\base\Model
 Let's say we need to check if salary is enough for children. We can create inline validator `validateChildrenFunds` for
 that which will run only when `childrenCount` is more than 0.
 
-Note that we can't use all validated attributes (`['personalSalary', 'spouseSalary', 'childrenCount']`` when attaching
+Note that we can't use all validated attributes (`['personalSalary', 'spouseSalary', 'childrenCount']`) when attaching
 validator. This is because the same validator will run for all attributes and we only need to run check once.
 
 You can use any of these attributes instead (or use what you think is the most relevant):
--- a/docs/guide/input-validation.md
+++ b/docs/guide/input-validation.md
@@ -664,7 +664,7 @@ Couple more changes are needed.
 First to use our `CustomInlineValidator` instead of default `InlineValidator` we need to override
 [[yii\validators\Validator::createValidator()]] method in `CustomValidator`:
 
-```
+```php
 public static function createValidator($type, $model, $attributes, $params = [])
 {
     $params['attributes'] = $attributes;
--- a/docs/guide/input-validation.md
+++ b/docs/guide/input-validation.md
@@ -481,7 +481,7 @@ class MigrationForm extends \yii\base\Model
 
 ### Creating validator <span id="multiple-attributes-validator"></span>
 
-Let's say we need to check if salary is enough for children. We can create inline validator `validateChildrenFunds` for
+Let's say we need to check if family income is enough for children. We can create inline validator `validateChildrenFunds` for
 that which will run only when `childrenCount` is more than 0.
 
 Note that we can't use all validated attributes (`['personalSalary', 'spouseSalary', 'childrenCount']`) when attaching
--- a/docs/guide/input-validation.md
+++ b/docs/guide/input-validation.md
@@ -463,9 +463,6 @@ class MigrationForm extends \yii\base\Model
     public $childrenCount;
     public $description;
 
-    /**
-     * @inheritdoc
-     */
     public function rules()
     {
         return [
@@ -572,7 +569,7 @@ namespace app\components;
 
 trait BatchValidationTrait
 {
-    /*
+    /**
      * @var bool whether to validate multiple attributes at once
      */
     public $batch = false;
@@ -662,7 +659,7 @@ class CustomInlineValidator extends InlineValidator
 Couple more changes are needed.
 
 First to use our `CustomInlineValidator` instead of default `InlineValidator` we need to override
-[[yii\validators\Validator::createValidator()]] method in `CustomValidator`:
+[[\yii\validators\Validator::createValidator()]] method in `CustomValidator`:
 
 ```php
 public static function createValidator($type, $model, $attributes, $params = [])
@@ -690,7 +687,7 @@ public static function createValidator($type, $model, $attributes, $params = [])
 ```
 
 And finally to support our custom validator in model we can create the trait and override
-[[yii\base\Model::createValidators()]] like this:
+[[\yii\base\Model::createValidators()]] like this:
 
 ```php
 <?php
@@ -804,9 +801,9 @@ public function validateChildrenFunds($attribute, $params)
 
 The advantages of this approach:
 
-- It better reflects all attributes participating in validation (the rules become more readable);
-- It correctly considers [[yii\validators\Validator::skipOnError]] and [[yii\validators\Validator::skipOnEmpty]]
-options for **each** used attribute (not only for that you decided to choose as more relevant).
+- It better reflects all attributes that participate in validation (the rules become more readable);
+- It respects the options [[yii\validators\Validator::skipOnError]] and [[yii\validators\Validator::skipOnEmpty]] for
+**each** used attribute (not only for that you decided to choose as more relevant).
 
 If you have problems with implementing client validation, you can:
 
--- a/docs/guide/input-validation.md
+++ b/docs/guide/input-validation.md
@@ -470,7 +470,7 @@ class MigrationForm extends \yii\base\Model
     {
         return [
             [['personalSalary', 'description'], 'required'],
-            [['personalSalary', 'spouseSalary'], 'integer', 'min' => 0],
+            [['personalSalary', 'spouseSalary'], 'integer', 'min' => self::MIN_ADULT_FUNDS],
             ['childrenCount', 'integer', 'min' => 0, 'max' => 5],
             [['spouseSalary', 'childrenCount'], 'default', 'value' => 0],
             ['description', 'string'],
--- a/docs/guide/input-validation.md
+++ b/docs/guide/input-validation.md
@@ -442,6 +442,106 @@ class EntryForm extends Model
 ```
 
 
+## Multiple Attributes Validation <span id="multiple-attributes-validation"></span>
+
+Sometimes validators involve multiple attributes. Consider the following form:
+
+``` php
+class MigrationForm extends \yii\base\Model
+{
+    /*
+     * Minimal funds amount for one adult person
+     */
+    const MIN_ADULT_FUNDS = 3000;
+    /*
+     * Minimal funds amount for one child
+     */
+    const MIN_CHILD_FUNDS = 1500;
+
+    public $personalSalary;
+    public $spouseSalary;
+    public $childrenCount;
+    public $description;
+
+    /**
+     * @inheritdoc
+     */
+    public function rules()
+    {
+        return [
+            [['personalSalary', 'description'], 'required'],
+            [['personalSalary', 'spouseSalary'], 'integer', 'min' => 0],
+            ['childrenCount', 'integer', 'min' => 0, 'max' => 5],
+            [['spouseSalary', 'childrenCount'], 'default', 'value' => 0],
+            ['description', 'string'],
+        ];
+    }
+}
+```
+
+Let's say we need to check if salary is enough for children. We can create inline validator `validateChildrenFunds` for
+that which will run only when `childrenCount` is more than 0.
+
+Note that we can't use all validated attributes (`['personalSalary', 'spouseSalary', 'childrenCount']`` when attaching
+validator. This is because the same validator will run for all attributes and we only need to run check once.
+
+You can use any of these attributes instead (or use what you think is the most relevant):
+
+```php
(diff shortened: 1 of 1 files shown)


SilverFire (approved):
Great! Thank you

docs/guide/input-validation.md
```diff
@@ -442,6 +442,121 @@ class EntryForm extends Model
...
+            [['spouseSalary', 'childrenCount'], 'default', 'value' => 0],
+            ['description', 'string'],
+        ];
+    }
+}
+```
+
+### Creating validator <span id="multiple-attributes-validator"></span>
+
+Let's say we need to check if salary is enough for children. We can create inline validator `validateChildrenFunds` for
```
samdark: I don't think it's a good example... See https://en.wikipedia.org/wiki/Child_labour.

Can we adjust that so it will be "architects" instead of adults and "builders" instead of children?

samdark (changes requested):

docs/guide/input-validation.md
arogachev: @samdark Sorry, I didn't get it. How it's related? 

The example is about a person who wants to migrate to other country (possibly with family including spouse and children).

This validation checks if their salary will be enough for children in new environment.

arogachev: Used constants are living wage for single adult person and a single child in new country.

samdark: Ah, that's what it does :)

samdark (approved):

arogachev:
@samdark Currently I'm adding another example that @klimov-paul suggested in https://github.com/yiisoft/yii2/issues/12778#issuecomment-256705262. But I found it's pretty complicated for achieving basically the same thing. Maybe it's better to add it to cookbook instead. I think I will add it here, please check, we can always delete it or move to cookbook.

docs/guide/input-validation.md
samdark: I'd still change `if salary is enough for children` to `if family income is enough`.

arogachev:
I added another option with creating custom validator. Hovewer, as I said, I find it pretty complicated (especially for beginners) and I think readability of attributes list and checking `skipOnError` and `skipOnEmpty` simply not worth these changes and overriding core code.

Should we:

- include it?
- delete it fully?
- somehow mention it as and idea in shorter way?
- move it to cookbook?

**P.S.** Beeing keen on writing docs, I forgot to add it `[skip ci]` to some commits. That's why #12993 is useful, at least for `.md` files. :smile:

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+    use BatchValidationTrait;
+}
+```
+
+Couple more changes are needed.
+
+First to use our `CustomInlineValidator` instead of default `InlineValidator` we need to override
+[[yii\validators\Validator::createValidator()]] method in `CustomValidator`:
+
+```
```
githubjeka: \`\`\`php

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
 
 
+## Multiple Attributes Validation <span id="multiple-attributes-validation"></span>
+
+Sometimes validators involve multiple attributes. Consider the following form:
+
+``` php
+class MigrationForm extends \yii\base\Model
+{
+    /*
```
dynasource: /**

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+     * Minimal funds amount for one adult person
+     */
+    const MIN_ADULT_FUNDS = 3000;
+    /*
```
dynasource: /**

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+            [['spouseSalary', 'childrenCount'], 'default', 'value' => 0],
+            ['description', 'string'],
+        ];
+    }
+}
+```
+
+### Creating validator <span id="multiple-attributes-validator"></span>
+
+Let's say we need to check if family income is enough for children. We can create inline validator `validateChildrenFunds` for
```
dynasource: the family income

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+that which will run only when `childrenCount` is more than 0.
+
+Note that we can't use all validated attributes (`['personalSalary', 'spouseSalary', 'childrenCount']`) when attaching
+validator. This is because the same validator will run for each attribute (3 times total) and we only need to run it
```
dynasource: 3 times in total

dynasource: for the whole attribute set

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+```
+
+You can ignore `$attribute` parameter because validation is not related to just one attribute.
+
+
+### Adding errors <span id="multiple-attributes-errors"></span>
+
+Adding error in case of multiple attributes can vary depending on desired form design:
+
+- Select the most important relevant field in your opinion and add error to it:
```
dynasource: to it's attribute:

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+}
+```
+
+You can ignore `$attribute` parameter because validation is not related to just one attribute.
+
+
+### Adding errors <span id="multiple-attributes-errors"></span>
+
+Adding error in case of multiple attributes can vary depending on desired form design:
+
```
dynasource: important relevant -> relevant

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+```
+
+- Add common error (not related to particular attribute). We can use not existing attribute name for adding error, for
+example `*`, because attribute existence is not checked at that point.
+
+```php
+$this->addError('*', 'Your salary is not enough for children.');
+```
+
+As a result, we will not see error message near form fields. To display it, we can include error summary in view:
```
dynasource: we can include the error summary in the view

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+Or use a loop:
+
+```php
+$attributes = ['personalSalary, 'wifeSalary', 'childrenCount'];
+foreach ($attributes as $attribute) {
+    $this->addError($attribute, 'Your salary is not enough for children.');
+}
+```
+
+- Add common error (not related to particular attribute). We can use not existing attribute name for adding error, for
```
dynasource: Add a common

dynasource: We can use the 'not existing attribute name' for adding the error

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+of them. Let's use a separate trait and override [[yii\base\Validator:validateAttributes()]]:
+
+```php
+<?php
+
+namespace app\components;
+
+trait BatchValidationTrait
+{
+    /*
```
dynasource: /**

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+    $childFunds = $totalSalary - $minAdultFunds;
+    if ($childFunds / $this->childrenCount < self::MIN_CHILD_FUNDS) {
+        $this->addError('childrenCount', 'Your salary is not enough for children.');
+    }
+}
+```
+
+The advantages of this approach:
+
+- It better reflects all attributes participating in validation (the rules become more readable);
```
dynasource: that participate in the validation

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+- It correctly considers [[yii\validators\Validator::skipOnError]] and [[yii\validators\Validator::skipOnEmpty]]
```
dynasource: it respects the options ....

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+            $params['class'] = $type;
+        }
+    }
+
+    return Yii::createObject($params);
+}
+```
+
+And finally to support our custom validator in model we can create the trait and override
+[[yii\base\Model::createValidators()]] like this:
```
dynasource: [[\yii\base\Model::createValidators()]]

docs/guide/input-validation.md
```diff
@@ -442,6 +442,381 @@ class EntryForm extends Model
...
+class CustomInlineValidator extends InlineValidator
+{
+    use BatchValidationTrait;
+}
+```
+
+Couple more changes are needed.
+
+First to use our `CustomInlineValidator` instead of default `InlineValidator` we need to override
+[[yii\validators\Validator::createValidator()]] method in `CustomValidator`:
```
dynasource: [[\yii\validators\Validator::createValidator()]]

dynasource (changes requested):
thats quite some book to review ;). I respect your effort to document in such detail

docs/guide/input-validation.md
arogachev: It's participle. This form is acceptable.

dynasource (approved):

samdark:
@arogachev it's is good enough. Do you plan more changes?

docs/guide/input-validation.md
arogachev: In other places of the docs (even in the same file) it's used without leading backslash. Used it as an example.

docs/guide/input-validation.md
arogachev: In other places of the docs (even in the same file) it's used without leading backslash. Used it as an example.

arogachev:
@samdark No. I think it's enough.

dynasource:
thanks @arogachev. Well done.

PR #13006, opened 2016-11-15
Captcha:: have to start with /, otherwise it does not work in modules.

boehsermoe:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  |no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  |


cebe:
Thank you!

PR #13008, opened 2016-11-15
phpDoc: Fix example for ActiveRelationTrait::via()

githubjeka:
Original example:

```php
      public function getOrders() // must be Order or hasMany and $link is bad #ololo
      {
          return $this->hasOne(Order::className(), ['id' => 'order_id']);
      }
     
      public function getOrderItems()
      {
         return $this->hasMany(Item::className(), ['id' => 'item_id'])
                      ->via('orders');
      }
```

I can't imagine structure of tables for this example. ^_^

PR replaces this example to [example from documentation](http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#junction-table)


samdark:
Thanks!

PR #13013, opened 2016-11-16
Closes #9162: Added support of closures in `value` for attributes in `yii\widgets\DetailView`

### LINKED ISSUE #9162: A reason to support DetailView value as Closure - delayed evaluation of value
Currently DataColumn supports setting value as Closure, while DetailView - does not, which is inconsistent. Often exactly the same formatting function is used to format the value in both GridView and DetailView.

Also, allowing setting value as Closure in DetailView delays the evaluation of that value until the actual rendering of that value takes place. And that evaluation may not happen at all if visible is set to false. This allows for the following code to work.

Initial (non-working) code:

```
<?= \yii\widgets\DetailView::widget([
    'model' => $model,
    'attributes' => [
        [
            'label' => 'User Info',
            // When $model->user is null we get PHP error.
            'value' => '<span class="' ($model->user->is_enabled ? 'green' : 'red') . '">' . $model->user->name . '</span>',
            'visible' => $model->user !== null,
        ],
    ],
]); ?>
```

A workaround without support for Closures:

```
'value' => ($model->user !== null ? '<span class="' ($model->user->is_enabled ? 'green' : 'red') . '">' . $model->user->name . '</span>' : ''),
'visible' => $model->user !== null,
```

But why add additional check for $model->user !== null (when it already present in 'visible') and define an empty string value for 'value' when that empty string case will never happen?

Better code with support for Closures:

```
// When $model->user is null Closure will not be called at all (due to visible being false), no PHP error.
'value' => function(){ return '<span class="' ($model->user->is_enabled ? 'green' : 'red') . '">' . $model->user->name . '</span>'; },
'visible' => $model->user !== null,
```

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #9162


framework/CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ Yii Framework 2 Change Log
 - Bug #12904: Fixed lowercase table name in migrations (zlakomanoff)
 - Bug #12939: Hard coded table names for MSSQL in RBAC migration (arogachev)
 - Bug #12974: Fixed incorrect order of migrations history in case `yii\console\controllers\MigrateController::$migrationNamespaces` is in use (evgen-d, klimov-paul)
+- Enh #6809: Added `\yii\caching\Cache::$ttl` property, allowing to set custom default cache duration (sdkiller)
```
SilverFire: Excess line

framework/widgets/DetailView.php
```diff
@@ -231,6 +231,10 @@ protected function normalizeAttributes()
                 throw new InvalidConfigException('The attribute configuration requires the "attribute" element to determine the value and display label.');
             }
 
+            if (is_callable($attribute['value'])) {
```
SilverFire: `$attribute['value'] instanceof \Closure`

docs/guide/output-data-widgets.md
```diff
@@ -37,6 +37,49 @@ echo DetailView::widget([
...
+            'value' => function ($model) {
+                return $model->owner->name;
+            }
+            'visible' => \Yii::$app->user->can('posts.owner.view'),
+        ],
+    ],
+]);
+```
+
+> Note: Closures in `value` for attributes is available only since 2.0.11.
```
SilverFire: is -> are

SilverFire (changes requested):

docs/guide/output-data-widgets.md
```diff
@@ -37,6 +37,90 @@ echo DetailView::widget([
...
+    'model' => $model,
+    'attributes' => [
+        [
+            'attribute' => 'owner',
+            'value' => // Call helper or decorator here instead
+        ],
+    ],
+]);
+```
+
```
arogachev: Not sure about adding reference to https://github.com/drapergem/draper. Can't find something similar for PHP. But at least brief mention of it is useful in my opinion, because I found people often put complex logic there and copy paste it from `GridView`.

docs/guide/output-data-widgets.md
```diff
@@ -37,6 +37,90 @@ echo DetailView::widget([
...
+> Note: Closures in `value` for attributes are available only since 2.0.11.
+
```
arogachev: How many newlines are needed? I see 1 and 2 in different places, it's a bit inconsistent.

docs/guide/output-data-widgets.md
samdark: I don't think referencing it will be of any use.

docs/guide/output-data-widgets.md
samdark: Doesn't matter.

docs/guide/output-data-widgets.md
```diff
@@ -37,6 +37,90 @@ echo DetailView::widget([
 ]);
 ```
 
+Remember that unlike [[yii\widgets\GridView|GridView]] which processes a set of models,
```
samdark: No need for header, I think.

samdark (approved):

docs/guide/output-data-widgets.md
arogachev: @samdark Referencing draper or this advice in common?

samdark: draper.

arogachev: Seems like you mean draper because you already approved changes. 😄

SilverFire (approved):

framework/widgets/DetailView.php
```diff
@@ -231,6 +231,10 @@ protected function normalizeAttributes()
                 throw new InvalidConfigException('The attribute configuration requires the "attribute" element to determine the value and display label.');
             }
 
+            if ($attribute['value'] instanceof \Closure) {
+                $attribute['value'] = call_user_func($attribute['value'], $this->model);
```
klimov-paul: Widget instance should be passed as 3rd argument:

```php
$attribute['value'] = call_user_func($attribute['value'], $this->model, $this);
```

klimov-paul (changes requested):

docs/guide/output-data-widgets.md
```diff
@@ -37,6 +37,90 @@ echo DetailView::widget([
...
+    'attributes' => [
+        [
+            'attribute' => 'owner',
+            'value' => // Call helper or decorator here instead
+        ],
+    ],
+]);
+```
+
+> Note: Closures in `value` for attributes are available only since 2.0.11.
```
klimov-paul: This note should be provided at `DetailView::$attributes` doc comment.

klimov-paul (changes requested):

docs/guide/output-data-widgets.md
```diff
@@ -37,6 +37,90 @@ echo DetailView::widget([
...
+echo DetailView::widget([
+    'model' => $model,
+    'attributes' => [
+        [
+            'attribute' => 'owner',
+            'value' => // Call helper or decorator here instead
+        ],
+    ],
+]);
+```
```
klimov-paul: I do not think all these docs are really necessary: there is no need to convince the developer to use `Closure` as a attribute value.
Docs are better to be simplified: posibility of `Closure` usage should be mentioned, but there is no necessity for such verbose examples.

klimov-paul (changes requested):

framework/widgets/DetailView.php
samdark: Wouldn't that be inconsistent compared to `DataColumn`?

```php
function($model, $key, $index, $column) { ...
```

samdark: Good idea. That would be consistent with `DataColumn` value callback: `function ($model, $key, $index, $column)`.

docs/guide/output-data-widgets.md
samdark: Agree. Closures are useful sometimes in DetailView but generally should be avoided because of performance reasons.

dynasource: I agree. I appreciate the effort, but the more statements you make, the more you have to support, especially when these statements are debatable.

docs/guide/output-data-widgets.md
arogachev: Should I move or copy it there?

docs/guide/output-data-widgets.md
dynasource: try to reason from the least viable example necessary to educate a developer (and remove the rest)

samdark: PHPDoc is enough in this case. Some examples may be moved there if needed.

SilverFire (approved):

klimov-paul (approved):

Issue #13016, opened 2016-11-16
Add `foreignKey` method to `SchemaBuilderTrait` for improve migration syntax


kfreiman:
It is enhancement. Add `foreignKey` method to `SchemaBuilderTrait` for improve migration syntax
### What is the expected result?
```php
$this->createTable('table1', [
    'id' => $this->primaryKey(),
    'table2_id' => $this->integer()->foreignKey('name_of_fk', 'table2', 'id'),
]);
```
### What do you get instead?
```php
$this->createTable('table1', [
    'id' => $this->primaryKey(),
    'table2_id' => $this->integer(),
]);

$this->addForeignKey(
    'name_of_fk', 
    'table',
    'table2_id',
    'table2',
    'id'
);
```

samdark:
Does it work w/ all databases we support?

kfreiman:
It uses `yii\dbCommand::addForeignKey` method, so it should be supported.

Issue #8163, opened 2015-04-23
Better constraints management in migrations


arogachev:
Currently we have to specify foreign key name with both adding and dropping.

```php
$this->addForeignKey(
    'tests_questions_test_id_fkey,
    'tests_questions',
    'test_id',
    'tests',
    'id',
    'CASCADE',
    'CASCADE'
);
```
```php
$this->dropForeignKey('tests_questions_test_id_fkey', 'tests_questions');
```

Some DBMS like MySQL generate it automatically, but the dropping part for me is more frustrating.
You have to look through all previous migrations and find the name of this foreign key.
One solution will be following some convention and ask team members to follow it too, but it will be great if:

1) Foreign key names will be generated automatically so we can just write this:

```php
$this->addForeignKey(
    'tests_questions_test_id_fkey,
    'tests_questions',
    'test_id',
    'tests',
    'id',
    'CASCADE',
    'CASCADE'
);
```

2) And for the dropping part we can specify relations something like that:

```php
$this->dropForeignKey('tests_questions', ['test_id' => 'test.id']);
```

And foreign key will be found and dropped automatically.

That way developer only cares about relations and not names.

I think It can be applied to indices too.

What do you think? Is it possible with DBMS?
Maybe use some convention if DBMS don't support auto generation? In this case we can build constraint name string depending on relations and add / drop it.

arogachev:
Automatic generation can be applied for example if the constraint name is skipped, so user can choose between manual / automatical handling of it.

Example with `users` and `messages` tables.

```php
$this->addForeignKey(
    null,
    'messages',
    'user_id',
    'users',
    'id',
    'CASCADE',
    'CASCADE'
);
```

Generated foreign key name will be for example: `messages_user_id-users-id_fk`.

Dropping: 

```php
$this->dropForeignKey(null, 'messages', ['id' => 'users.id']);
```

We can retrieve this string `messages_user_id-users-id_fk` based on passed relations and drop corresponding foreign key.

arogachev:
Seems like this concept already used in [Phinx](http://docs.phinx.org/en/latest/migrations.html). Not sure about their implementation of it.

arogachev:
And in Phinx relations are not even specified for dropping foreign keys, just the column name.

arogachev:
Also similar concept is used in Laravel.

nineinchnick:
[...]

Issue #9657, opened 2015-09-10
FR: Migration, add encapsulation for primary key, foreign key and etc


mdmunir:
NO, its not same with `Migration::addPrimaryKey()` or `Migration::addForeignKey()`. Its used directly at create table statement.
```php
$this->createTable('{{%auth}}', [
    'user_id' => $this->integer()->notNull(),
    'source' => $this->string(255)->notNull(),
    'source_id' => $this->string(255)->notNull(),
    $this->primaryKey(['user_id','source']), // <-- unsure with this name
    $this->foreignKey('{{%user}}',['user_id'=>'id'],'CASCADE','CASCADE')
], $tableOptions);
```

samdark:
Makes sense to me.

omnilight:
Currently we have `$this->primaryKey()` that returns for mysql `int(11) NOT NULL AUTO_INCREMENT PRIMARY KEY`, so if we will use `$this->primaryKey()` the way suggested by @mdmunir , we will lose backward compatibility.

I think, that if current implementation of migration's `$this->integer()` and other methods return object, why `$this->createTable()` can not do the same?

In this key we will have:

```php
$this
    ->createTable('{{%auth}}', [
        'user_id' => $this->primaryKey(),
        'source' => $this->string(255)->notNull(),
        'source_id' => $this->string(255)->notNull(),
    ], $tableOptions)
    ->index('user_id')
    ->index('source_id')
    ->foreightKey('user_id', '{{%user}}', 'id', 'CASCADE', 'CASCADE');
```

Each of this methods should execute required query and return some object, for example:

```php
public function createTable($table, $columns, $options = null)
{
    echo "    > create table $table ...";
    $time = microtime(true);
    $this->db->createCommand()->createTable($table, $columns, $options)->execute();
    echo " done (time: " . sprintf('%.3f', microtime(true) - $time) . "s)\n";
    return $this->getDb()->getSchema()->createTableSchemaBuilder($table);
}
```

PR #13018, opened 2016-11-16
Added `foreignKey` method for `yii\db\ColumnSchemaBuilder`

kfreiman:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13016, #8163

### DIFF
--- a/framework/db/ColumnSchemaBuilder.php
+++ b/framework/db/ColumnSchemaBuilder.php
@@ -115,6 +115,26 @@ class ColumnSchemaBuilder extends Object
      * @since 2.0.8
      */
     public $comment;
+    /**
+     * @var string the name of the foreign key constraint.
+     */
+    public $foreignKeyName;
+    /**
+     * @var string the table that the foreign key references to.
+     */
+    public $refTable;
+    /**
+     * @var string the name of the column that the foreign key references to.
+     */
+    public $refColumn;
+    /**
+     * @var string the ON DELETE option. Most DBMS support these options: RESTRICT, CASCADE, NO ACTION,
+     */
+    public $delete;
+    /**
+     * @var string the ON UPDATE option. Most DBMS support these options: RESTRICT, CASCADE, NO ACTION,
+     */
+    public $update;
 
     /**
      * Create a column schema builder instance giving the type and value precision.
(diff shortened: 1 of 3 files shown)


arogachev:
`since` tags are missing. What about unit tests?

framework/db/Migration.php
```diff
@@ -265,6 +265,18 @@ public function createTable($table, $columns, $options = null)
             if ($type instanceof ColumnSchemaBuilder && $type->comment !== null) {
                 $this->db->createCommand()->addCommentOnColumn($table, $column, $type->comment)->execute();
             }
+
+            if ($type instanceof ColumnSchemaBuilder && $type->foreignKeyName !== null) {
+                $this->db->createCommand()->addForeignKey(
```
Faryshta: shouldn't also create the index key?

framework/db/ColumnSchemaBuilder.php
```diff
@@ -115,6 +115,26 @@ class ColumnSchemaBuilder extends Object
      * @since 2.0.8
      */
     public $comment;
+    /**
+     * @var string the name of the foreign key constraint.
+     */
+    public $foreignKeyName;
```
Faryshta: is it ok to make them public and not protected since only `foreignKey()` method can modify them

framework/db/ColumnSchemaBuilder.php
```diff
@@ -270,6 +290,25 @@ public function append($sql)
...
     /**
+     * Sets a foreign key for the column
+     * @param string $foreignKeyName the name of the foreign key constraint.
+     * @param string $refTable the table that the foreign key references to.
+     * @param string $refColumn the name of the column that the foreign key references to.
+     * @param string $delete the ON DELETE option. Most DBMS support these options: RESTRICT, CASCADE, NO ACTION, SET DEFAULT, SET NULL
+     * @param string $update the ON UPDATE option. Most DBMS support these options: RESTRICT, CASCADE, NO ACTION, SET DEFAULT, SET NULL
+     * @return $this
+     */
+    public function foreignKey($foreignKeyName, $refTable, $refColumn, $delete=null, $update=null)
```
Faryshta: this name clashes with an existing functionality on the generator http://www.yiiframework.com/doc-2.0/guide-db-migrations.html#foreign-keys

framework/db/ColumnSchemaBuilder.php
samdark: Not really.

Faryshta: not ok or not bad that the properties are public?

samdark: Not OK.

rob006:
Related: 
https://github.com/yiisoft/yii2/issues/9657
https://github.com/yiisoft/yii2/issues/14623

I think this should be rewritten to allow define FK for SQLite - FK definition should be injected into `CREATE TABLE` command.
See https://github.com/yiisoft/yii2/issues/14623#issuecomment-338854045 https://sqlite.org/foreignkeys.html

PR #13020, opened 2016-11-16
Provide options for the span element.

nadar:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no

It should be possibile to provide optional classes for the disable state list item. 

(As we are currently building all widgets for bootstrap 4, the pagination element looks different, there for we need more options to configure)

### DIFF
--- a/framework/widgets/LinkPager.php
+++ b/framework/widgets/LinkPager.php
@@ -74,6 +74,12 @@ class LinkPager extends Widget
      * @var string the CSS class for the disabled page buttons.
      */
     public $disabledPageCssClass = 'disabled';
+    
+    /**
+     * @var array An array with options to pass to the span tag which is generated inside the disabled list element.
+     */
+    public $disabledSpanOptions = [];
+    
     /**
      * @var int maximum number of page buttons that can be displayed. Defaults to 10.
      */
@@ -218,7 +224,7 @@ class LinkPager extends Widget
         if ($disabled) {
             Html::addCssClass($options, $this->disabledPageCssClass);
 
-            return Html::tag('li', Html::tag('span', $label), $options);
+            return Html::tag('li', Html::tag('span', $label, $this->disabledSpanOptions), $options);
         }
         $linkOptions = $this->linkOptions;
         $linkOptions['data-page'] = $page;
--- a/framework/widgets/LinkPager.php
+++ b/framework/widgets/LinkPager.php
@@ -77,8 +77,8 @@ class LinkPager extends Widget
     public $disabledPageCssClass = 'disabled';
     
     /**
-     * @var array An array with options to pass to the disabled tag which is generated inside the disabled list element.
-     * In order to to customize the tag name itself use the array key `tag` and value for its tag name value.
+     * @var array the options for the disabled tag to be generated inside the disabled list element.
+     * In order to customize the html tag, please use the tag key.
      * 
      * ```php
      * $disabledPageElementOptions = ['tag' => 'div', 'class' => 'disabled-div'];
--- a/framework/widgets/LinkPager.php
+++ b/framework/widgets/LinkPager.php
@@ -83,6 +83,7 @@ class LinkPager extends Widget
      * ```php
      * $disabledPageElementOptions = ['tag' => 'div', 'class' => 'disabled-div'];
      * ```
+     * @since 2.0.11
      */
     public $disabledPageElementOptions = [];
     
--- a/framework/widgets/LinkPager.php
+++ b/framework/widgets/LinkPager.php
@@ -12,6 +12,7 @@ use yii\base\InvalidConfigException;
 use yii\helpers\Html;
 use yii\base\Widget;
 use yii\data\Pagination;
+use yii\helpers\ArrayHelper;
 
 /**
  * LinkPager displays a list of hyperlinks that lead to different pages of target.
@@ -76,9 +77,9 @@ class LinkPager extends Widget
     public $disabledPageCssClass = 'disabled';
     
     /**
-     * @var array An array with options to pass to the span tag which is generated inside the disabled list element.
+     * @var array An array with options to pass to the disabled tag which is generated inside the disabled list element.
      */
-    public $disabledSpanOptions = [];
+    public $disabledPageElementOptions = [];
     
     /**
      * @var int maximum number of page buttons that can be displayed. Defaults to 10.
@@ -223,8 +224,8 @@ class LinkPager extends Widget
         }
         if ($disabled) {
             Html::addCssClass($options, $this->disabledPageCssClass);
-
-            return Html::tag('li', Html::tag('span', $label, $this->disabledSpanOptions), $options);
+            $tag = ArrayHelper::remove($this->disabledPageElementOptions, 'tag', 'span');
+            return Html::tag('li', Html::tag($tag, $label, $this->disabledPageElementOptions), $options);
         }
         $linkOptions = $this->linkOptions;
         $linkOptions['data-page'] = $page;
--- a/framework/widgets/LinkPager.php
+++ b/framework/widgets/LinkPager.php
@@ -81,11 +81,11 @@ class LinkPager extends Widget
      * In order to customize the html tag, please use the tag key.
      * 
      * ```php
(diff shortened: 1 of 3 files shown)


framework/widgets/LinkPager.php
```diff
@@ -74,6 +74,12 @@ class LinkPager extends Widget
      * @var string the CSS class for the disabled page buttons.
      */
     public $disabledPageCssClass = 'disabled';
+    
+    /**
+     * @var array An array with options to pass to the span tag which is generated inside the disabled list element.
+     */
+    public $disabledSpanOptions = [];
```
SilverFire: Maybe it's better to rename this property to `disabledPaseElementOptions` to be consistent with `disabledPageCssClass` and `activePageCssClass` property?

framework/widgets/LinkPager.php
```diff
@@ -218,7 +224,7 @@ protected function renderPageButton($label, $page, $class, $disabled, $active)
         if ($disabled) {
             Html::addCssClass($options, $this->disabledPageCssClass);
 
-            return Html::tag('li', Html::tag('span', $label), $options);
+            return Html::tag('li', Html::tag('span', $label, $this->disabledSpanOptions), $options);
```
SilverFire: It makes sense to make tag name customizable as it's done [here](https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1225), for example

SilverFire (changes requested):
Could you add a line to Changelog and update unit tests?

framework/widgets/LinkPager.php
```diff
@@ -74,6 +75,12 @@ class LinkPager extends Widget
      * @var string the CSS class for the disabled page buttons.
      */
     public $disabledPageCssClass = 'disabled';
+    
+    /**
+     * @var array An array with options to pass to the disabled tag which is generated inside the disabled list element.
```
SilverFire: Special option `tag` must be mentioned

framework/widgets/LinkPager.php
```diff
@@ -217,8 +224,8 @@ protected function renderPageButton($label, $page, $class, $disabled, $active)
         }
         if ($disabled) {
             Html::addCssClass($options, $this->disabledPageCssClass);
-
-            return Html::tag('li', Html::tag('span', $label), $options);
+            $tag = ArrayHelper::remove($this->disabledPageElementOptions, 'tag', 'span');
+            return Html::tag('li', Html::tag($tag, $label, $this->disabledPageElementOptions), $options);
```
SilverFire: Empty line before the `return` statement is required according to code style agreement

tests/framework/widgets/LinkPagerTest.php
```diff
@@ -56,4 +56,34 @@ public function testFirstLastPageLabels()
         static::assertNotContains('<li class="first">', $output);
         static::assertNotContains('<li class="last">', $output);
     }
+    
+    public function testDisabledPageElementOptions()
+    {
+    	$pagination = new Pagination();
```
SilverFire: Code align in tests should be fixed

SilverFire (changes requested):

SilverFire (approved):

framework/widgets/LinkPager.php
```diff
@@ -74,6 +75,17 @@ class LinkPager extends Widget
...
+    
+    /**
+     * @var array An array with options to pass to the disabled tag which is generated inside the disabled list element.
+     * In order to to customize the tag name itself use the array key `tag` and value for its tag name value.
+     * 
+     * ```php
+     * $disabledPageElementOptions = ['tag' => 'div', 'class' => 'disabled-div'];
+     * ```
+     */
+    public $disabledPageElementOptions = [];
```
samdark: Need `@since` tag.

samdark (changes requested):

samdark (approved):

framework/widgets/LinkPager.php
dynasource: More consistent:
 @var array An array with  ...->  @var array the options for the disabled tag to be generated inside ...

framework/widgets/LinkPager.php
```diff
@@ -74,6 +75,18 @@ class LinkPager extends Widget
      * @var string the CSS class for the disabled page buttons.
      */
     public $disabledPageCssClass = 'disabled';
+    
+    /**
+     * @var array An array with options to pass to the disabled tag which is generated inside the disabled list element.
+     * In order to to customize the tag name itself use the array key `tag` and value for its tag name value.
```
dynasource: In order to to customize the tag name itself use the array key `tag` and value for its tag name value.
->
In order to customize the html tag, please use the `tag` key.

framework/widgets/LinkPager.php
```diff
@@ -74,6 +75,18 @@ class LinkPager extends Widget
...
+     * 
+     * ```php
+     * $disabledPageElementOptions = ['tag' => 'div', 'class' => 'disabled-div'];
+     * ```
+     * @since 2.0.11
+     */
+    public $disabledPageElementOptions = [];
```
dynasource: pageElement....isn't this rather vague? I have a hard time imagining what a 'disabledPageElement' is...

tests/framework/widgets/LinkPagerTest.php
```diff
@@ -56,4 +56,34 @@ public function testFirstLastPageLabels()
         static::assertNotContains('<li class="first">', $output);
         static::assertNotContains('<li class="last">', $output);
     }
+    
+    public function testDisabledPageElementOptions()
+    {
+        $pagination = new Pagination();
+        $pagination->setPage(0);
```
dynasource: it it possible to remove these methods and to pass these as an options array?

dynasource (changes requested):

framework/widgets/LinkPager.php
nadar: It was suggested by @SilverFire. But its also hard to imagine what the other properties stand for.

tests/framework/widgets/LinkPagerTest.php
nadar: I just copied from the previous existing test.

dynasource (approved):

tests/framework/widgets/LinkPagerTest.php
dynasource: alright, consistent then.

dynasource (changes requested):
I still don't like this $property name:
$disabledPageElementOptions 

It should properly inform me about what it is. Now I'm still wondering.

nadar:
@dynasource Well, do you have a better idea?

samdark:
`$disabledPageButtonOptions`?

dynasource:
> @dynasource Well, do you have a better idea?

Well, lets start by looking at your own intro text:
"It should be possibile to provide optional classes for the disable state list item."

This is understandable.

So I would expect something like a $disabledListItemOptions or $disabledListItemSpanOptions

nadar:
Sounds good to. Infact its not the list element, its the element inside the list element. And if the tag name can be changed it should not contain the name "Span" (as i mentioned first as well).

dynasource:
`$disabledListItemSubTagOptions`

dynasource:
of course this is long and should be prevented, but it is always better than a tag name that nobody understands

SilverFire:
> `$disabledPageButtonOptions`?

No, it's not button options.

> Infact its not the list element, its the element inside the list element. And if the tag name can be changed it should not contain the name "Span" (as i mentioned first as well).

Right.

> `$disabledListItemSubTagOptions`

Looks complex enough

nadar:
So `$disabledListItemSubTagOptions` is suitable for everyone? Then i am going to change it.

samdark:
Yes, it's better than the current name. Not ideal but certainly better.

dynasource (approved):

dynasource:
merged in https://github.com/yiisoft/yii2/commit/aba42fce4dcfdd7c3b9dd97f4d3c9eed9101bdf4, thanks!

PR #13022, opened 2016-11-16
Closes #7420: Attributes for prompt generated with `renderSelectOptions` of `\yii\helpers\Html` helper

### LINKED ISSUE #7420: Attributes for dropDownList prompt
It's possible to add attributes to a dropDownList's option tags using the 'options' element in $options, e.g.:

``` php
$form->field($model, 'foo')->dropDownList(['value1' => 'One', 'value2' => 'Two'], [
    'options' => [
        'value1' => ['disabled' => true],
        'value2' => ['class' => 'bar']
    ],
    'prompt' => 'please choose',
])
```

But it's not possible to add attributes to the option tag generated by 'prompt'. I'd like to do this:

``` php
$form->field($model, 'foo')->dropDownList(['value1' => 'One', 'value2' => 'Two'], [
    'prompt' => 'please choose',
    'promptOptions' => [
        'class' => 'drop-down-prompt'
    ]
])
```

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7420

We have the following options of implementing this:

- Initial suggestion from @nkovacs (https://github.com/yiisoft/yii2/issues/7420#issue-58740067), adding separate additional array element:

```php
echo $form->field($model, 'foo')->dropDownList(['value1' => 'One', 'value2' => 'Two'], [
    'prompt' => 'please choose',
    'promptOptions' => [
        'class' => 'drop-down-prompt',
    ],
]);
```

- Method suggested by @Faryshta (https://github.com/yiisoft/yii2/issues/7420#issuecomment-75860925), using one array, storing prompt text under `label` key and the other attributes under according keys in the same array:

```php
'prompt' => ['label' => 'Please Choose', 'class' => 'drop-down-prompt', 'value' => '-1'];
```

Rendering option without content inside of a tag [will work](http://www.w3schools.com/tags/att_option_label.asp):

```html
<option label="Please Choose", "class"="drop-down-prompt", "value"="-1"></option>
```

But with this method there are some problems with older browsers, also putting text inside of a tag is more traditional way and it's consistent with other options.

Using `label` for both `label` attribute and tag content is not good idea too, because the label attribute value is intended to be shorter form of tag content. Example:

```html
<select>
  <option label="Volvo">Volvo (Latin for "I roll")</option>
  <option label="Saab">Saab (Swedish Aeroplane AB)</option>
  <option label="Mercedes">Mercedes (Mercedes-Benz)</option>
  <option label="Audi">Audi (Auto Union Deutschland Ingolstadt)</option>
</select>
```

So if user decides to specify custom `label` we can't distinguish between them.

As a workaround we can use some other special key like `text`, use it as tag content and use the rest of array as tag attributes:

```php
'prompt' => ['label' => 'Choose', 'text' => 'Please Choose', 'class' => 'drop-down-prompt', 'value' => '-1'];
```

Rendering result:

```html
<option label="Choose", "class"="drop-down-prompt", "value"="-1">Please Choose</option>
```

However such mixing is not clear in my opinion.

- Solution I implemented in this PR. Use `text` for content tag and specify options as subarray.

```php
'prompt' => [
    'text' => 'Please select', 'options' => ['class' => 'prompt', 'value' => '-1', 'label' => 'None'],
],
```

Rendering result:

```html
<option class="prompt" value="-1" label="None">Please select</option>
```

- Also I was thinking about shorter form without associative array at the top level, something like that:

```php
'prompt' => ['Please select', ['class' => 'prompt', 'value' => '-1', 'label' => 'None']],
```

But this adds strict order and inconsistent with config style in other places.

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -754,7 +754,13 @@ class BaseHtml
      * the labels will also be HTML-encoded.
      * @param array $options the tag options in terms of name-value pairs. The following options are specially handled:
      *
-     * - prompt: string, a prompt text to be displayed as the first option;
+     * - prompt: string, a prompt text to be displayed as the first option. Since version 2.0.11 you can use array to
+     * override value and set other tag attributes:
+     *
+     * ```php
+     * ['text' => 'Please select', 'options' => ['value' => 'none', 'class' => 'prompt', 'label' => 'Select']],
+     * ```
+     *
      * - options: array, the attributes for the select option tags. The array keys must be valid option values,
      *   and the array values are the extra attributes for the corresponding option tags. For example,
      *
@@ -803,7 +809,13 @@ class BaseHtml
      * the labels will also be HTML-encoded.
      * @param array $options the tag options in terms of name-value pairs. The following options are specially handled:
      *
-     * - prompt: string, a prompt text to be displayed as the first option;
+     * - prompt: string, a prompt text to be displayed as the first option. Since version 2.0.11 you can use array to
+     * override value and set other tag attributes:
+     *
+     * ```php
+     * ['text' => 'Please select', 'options' => ['value' => 'none', 'class' => 'prompt', 'label' => 'Select']],
+     * ```
+     *
      * - options: array, the attributes for the select option tags. The array keys must be valid option values,
      *   and the array values are the extra attributes for the corresponding option tags. For example,
      *
@@ -1477,7 +1489,13 @@ class BaseHtml
      * the labels will also be HTML-encoded.
      * @param array $options the tag options in terms of name-value pairs. The following options are specially handled:
      *
-     * - prompt: string, a prompt text to be displayed as the first option;
+     * - prompt: string, a prompt text to be displayed as the first option. Since version 2.0.11 you can use array to
+     * override value and set other tag attributes:
+     *
+     * ```php
+     * ['text' => 'Please select', 'options' => ['value' => 'none', 'class' => 'prompt', 'label' => 'Select']],
+     * ```
+     *
      * - options: array, the attributes for the select option tags. The array keys must be valid option values,
      *   and the array values are the extra attributes for the corresponding option tags. For example,
      *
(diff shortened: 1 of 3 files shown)


framework/helpers/BaseHtml.php
```diff
@@ -1693,11 +1693,18 @@ public static function renderSelectOptions($selection, $items, &$tagOptions = []
         $encodeSpaces = ArrayHelper::remove($tagOptions, 'encodeSpaces', false);
         $encode = ArrayHelper::remove($tagOptions, 'encode', true);
         if (isset($tagOptions['prompt'])) {
-            $prompt = $encode ? static::encode($tagOptions['prompt']) : $tagOptions['prompt'];
+            $attrs = ['value' => ''];
```
SilverFire: We prefer not to use shorten variable names

framework/helpers/BaseHtml.php
```diff
@@ -1693,11 +1693,18 @@ public static function renderSelectOptions($selection, $items, &$tagOptions = []
...
+            if (is_string($tagOptions['prompt'])) {
+                $text = $tagOptions['prompt'];
+            } else {
+                $text = $tagOptions['prompt']['text'];
```
SilverFire: Should be mentioned in PHPDocs

SilverFire (changes requested):

SilverFire:
This method already looks so monstrous :(

framework/helpers/BaseHtml.php
arogachev: @SilverFire Me too. I tried to keep the same style that already used in this method above. Is it OK to use full name only in this place?

framework/helpers/BaseHtml.php
arogachev: I haven't added PHPDoc yet because I was not sure about implementation (that's why I listed different options in initial post). If it's OK I will add it.

SilverFire: IMO it's OK. Let's wait for another opinions

arogachev:
We can refactor it by extracting some logic to separate methods, for example `renderSelectPrompt`, `renderSelectOptionGroup`, `renderSelectOption` (for single option).

framework/helpers/BaseHtml.php
samdark: Yes.

SilverFire:
Not now :)

framework/helpers/BaseHtml.php
```diff
@@ -754,7 +754,13 @@ protected static function booleanInput($type, $name, $checked = false, $options
      * the labels will also be HTML-encoded.
      * @param array $options the tag options in terms of name-value pairs. The following options are specially handled:
      *
-     * - prompt: string, a prompt text to be displayed as the first option;
+     * - prompt: string, a prompt text to be displayed as the first option. Since version 2.0.11 you can use array to
```
dynasource: Since version 2.0.11 you can use an array to override the value and to set other tag attributes:

dynasource (changes requested):

dynasource (approved):

samdark (approved):

SilverFire (approved):

SilverFire:
Merged: https://github.com/yiisoft/yii2/commit/c72643b9a35334f5f3753470e69b617b186b2a04
Thank you!

PR #13027, opened 2016-11-17
Closes #6242: Access to validator in inline validation

### LINKED ISSUE #6242: Inline validator improvement
What do you think about passing the validator instance (`$this`) to the `InlineValidator::$method` call? It will provide access to the `Validator`'s common methods like `addError` or `isEmpty`. The first one should be important, because it provide some routine for preparing error message.

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #6242


framework/validators/InlineValidator.php
```diff
@@ -19,6 +19,15 @@
  * where `$attribute` refers to the name of the attribute being validated, while `$params`
  * is an array representing the additional parameters supplied in the validation rule.
  *
+ * In version 2.0.11 additional argument `$validator` containing reference to related `InlineValidator` was added.
```
SilverFire: IMO line 16 must be modified as follows:

```
function foo($attribute, $params, $validator)
```

This line should be changed to:

```
Parameter `$validator` refers to the related [[InlineValidator]] object and is available since version 2.0.11
```

All the following lines of PHP could be removed

SilverFire (changes requested):
Could you update [guide article](http://www.yiiframework.com/doc-2.0/guide-input-validation.html#creating-validators) as well, please?

arogachev:
I should mention that even it seems to be useful, I can't see any benefits from adding `$validator` except extended `addError()` method.

Note that `addError()` in `yii\validators\Validator` does not use `$this` at all, so it can be static.

Also why not put this logic in `addError()` of `yii\base\Model`?

Case mentioned in related issue https://github.com/yiisoft/yii2/issues/7540#issuecomment-77099554 is not clear for me, why not specify these options in validation rule instead?

arogachev:
From the other side, I can't remember when I used `$column` at least once (from personal experience)  in `yii\grid\DataColumn` `$value` closure - `function ($model, $key, $index, $column)`. Maybe it's just less obvious.

tests/framework/validators/ValidatorTest.php
```diff
@@ -63,6 +63,28 @@ public function testCreateValidator()
...
+        $val = TestValidator::createValidator('inlineVal', $model, ['val_attr_a'], ['params' => ['foo' => 'bar']]);
+
+        $val->validateAttribute($model, 'val_attr_a');
+        $args = $model->getInlineValArgs();
+        $this->assertCount(3, $args);
+        $this->assertEquals('val_attr_a', $args[0]);
+        $this->assertEquals(['foo' => 'bar'], $args[1]);
+        $this->assertInstanceOf(InlineValidator::className(), $args[2]);
+
+        $val->clientValidate = 'clientInlineVal';
```
dynasource: put this in a second method with distinctive test method names plz

tests/data/validators/models/FakedValidationModel.php
```diff
@@ -63,4 +71,14 @@ public function getAttributeLabel($attr)
     {
         return $attr;
     }
+
+    /**
+     * Get arguments of inlineVal method in the last call.
```
dynasource: Consistency in framework:
'Returns the arguments of the ...'

framework/validators/InlineValidator.php
```diff
@@ -27,11 +28,12 @@ class InlineValidator extends Validator
     /**
      * @var string|\Closure an anonymous function or the name of a model class method that will be
      * called to perform the actual validation. The signature of the method should be like the following,
-     * where `$attribute` is the name of the attribute to be validated, and `$params` contains the value
-     * of [[params]] that you specify when declaring the inline validation rule:
+     * where `$attribute` is the name of the attribute to be validated, `$params` contains the value
```
dynasource: this would be more readable in list form:
- `$attribute` is the ...
- `$params` contains ...

docs/guide/input-validation.md
```diff
@@ -372,6 +373,14 @@ class MyForm extends Model
 }
 ```
 
+> Note: you can use [[yii\validators\InlineValidator::addError()]] for adding errors instead. That way the error
+> message can be formatted using [[yii\i18n\I18N::format()]] right away. Use `{attribute}` and `{value}` in the error
```
dynasource: [[\yii\....

docs/guide/input-validation.md
```diff
@@ -372,6 +373,14 @@ class MyForm extends Model
...
+> message to refer to attribute label (no need to get it manually) and attribute value accordingly:
```
dynasource: to refer to **an** attribute label (please mind these words before the subject, like `a` `an`, `the` etc...

docs/guide/input-validation.md
```diff
@@ -331,8 +331,9 @@ the method/function is:
 /**
  * @param string $attribute the attribute currently being validated
  * @param mixed $params the value of the "params" given in the rule
+ * @param \yii\validators\InlineValidator Related InlineValidator instance
```
dynasource: related

dynasource (changes requested):
thanks for this pull request. Have some minor requests for improvements

docs/guide/input-validation.md
rob006: @dynasource In guide you don't need leading backslash for namespaces.

arogachev:
Added fixes related with PR review. Please check if it's OK now.

docs/guide/input-validation.md
arogachev: @dynasource I left it as is, without leading backslash, to be the same as in the rest of the docs, also like @rob006 said it's not required.

SilverFire (approved):

samdark (approved):

arogachev:
Fixed problems with CHANGELOG. Is it ready for merge now?

dynasource (approved):

docs/guide/input-validation.md
cebe: Further explanation on this: in the guide the context around is the root namespace so it is like in PHP itself, when you are in a namespace you need `\` to get to the root. In root namespace you do not need it.

cebe (approved):

cebe:
Thank you!

Issue #11758, opened 2016-06-14
Set dependencies in application config


hiqsol:
What do you think about adding function to yii\base\Application:

``` php
public function setDependencies(array $dependencies)
{
    foreach ($dependencies as $class => $definition) {
        Yii::$container->set($class, $definition);
    }
}
```

This will allow setup dependencies in main application config in the same way as aliases are set now:

``` php
$config = [
    ...
    'aliases' => [
         '@foo' => '@runtime/foo',
    ],
    'dependencies' => [
        'yii\mail\MailInterface' => 'yii\swiftmailer\Mailer',
        'yii\widgets\LinkPager' => [
            'maxButtonCount' => 5,
        ],
    ],
    ...
];
```

samdark:
Sounds good to me. I was asked about it multiple times.

samdark:
@yiisoft/core-developers any objections?

SilverFire:
Could be implemented

hiqsol:
Ok, if no objections I will do PR.

cebe:
what does it do? configure Yii::$container?

hiqsol:
It's the same story as with aliases.
What does Application::setAliases() do? configure Yii::$aliases.
And you have it because it allows to setup aliases with **configuration** not code.
According guide:
> This property is provided so that you can define aliases in terms of application configurations instead of by calling the Yii::setAlias() method.

Application config looks good place to put dependencies config. There is no other global config.

lynicidn:
if change
$application = new yii\web\Application($config);
to
$application = Yii::createObject(yii\web\Application::class, $config);
than this solution not work for app

app can not resolve dependencies, example `_bootstrap.php:
```php
Yii::$container->set('yii\web\Application', 'MyApplication');
Yii::$container->set('MyAppInject', 'MyDep');
class MyApplication
{
    public function __construct($config, MyAppInject) {}
}
```
по русски - ваше решение ограничивает использование контейнера для самого приложения

SilverFire:
> what does it do? configure Yii::$container? that is not the job of the application class imo.

It was my first question when we had a discussion with @hiqsol offline.

This suggest might look awkward in terms of design, but since we don't have a better way to manage DI in declarative way, it might be implemented.

@lynicidn Thank you for your idea of a possibly unexpected behavior. The suggested example will not work just because the code that puts your dependencies' config to the DI container is being executed after creation of the target object. The code bellow should work as expected:

```php
[...]

PR #13029, opened 2016-11-17
Added Application::setContainerDefinitions() to configure DI in declarative style

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | not implemented yet
| Fixed issues  | #11758

- [x] Draft implementation & docs
- [x] Tests

This PR is an implementation of idea, suggested by @hiqsol in #11758 

The proposal is to allow dependency injection configuration be: 

- declarative
- located in the standard place
- be consistent with configuring of similar construction (aliases)

Example:

config.php
```php

return [
    'id' => 'my-super-application',
    'basePath' => dirname(__DIR__),
    'bootstrap' => ['log'],
    'containerDefinitions' => [
          yii\web\Request::class => app\components\Request::class,
          yii\web\Response::class => [
              'class' => app\components\Response::class,
              'format' => 'json'
          ],
    ]
];
```


samdark:
I have doubts about naming/structure of the config.

We currently have 2 ways of defining dependency:

1. `Yii::$container->setSingleton()`. In this case `get()` returns the same service object each time. It's not making an object singleton though.
2. `Yii::$container->set()`. In this case `get()` creates new instance of the object each time.

First way is preferred for services so it's mainly used when wiring stuff from container to object constructors. In Yii apps it's not used much because it's partially covered by service locator.  Second way is good for setting defaults for widgets etc.

I have some alternatives for separate `containerDefinitions` and `containerSingletons` keys.

1st (Symfony style):

```php
'services' => [
    yii\web\Request::class => app\components\Request::class, // by default class instances are cached
    yii\web\Connection::class => [
        'class' => app\components\Connection::class,
        'shared' => false, // explicitly declare that we need new instances each time
    ],
],
```

2nd (grouped):

```php
'services' => [
    'create' => [
        yii\web\Request::class => app\components\Request::class, // by default class instances are cached
    ],
    'share' => [
        yii\web\Connection::class => [
           'class' => app\components\Connection::class,
           'shared' => false, // explicitly declare that we need new instances each time
       ],
],    
```

3rd (alternative naming):

```php
'dependencies' => [
    'prototypes' => [
        yii\web\Request::class => app\components\Request::class, // by default class instances are cached
    ],
    'shared' => [
        yii\web\Connection::class => [
           'class' => app\components\Connection::class,
           'shared' => false, // explicitly declare that we need new instances each time
       ],
],    
```

hiqsol:
@samdark 
1st (Symfony style) variant looks most simple and convenient.
But I would vote for `dependencies` not `services`.

Unfortunately 1st variant doesn't work for callback definitions:

```php
'dependencies' => [
    yii\web\Connection::class => function () {
    }, /// no place to put shared = false :(
],
```

So the 4th variant:
```php
'dependencies' => [
    'definitions' => ...
    'singletons' => ...
],
```

These names are consistent to yii\di\Container.

samdark:
I've updated examples.

dynasource:
I don't like the word dependencies. It's too generic and confusing with either assets / composer / NPM dependencies. 

I'm also wondering why we are moving away from the 'main subject': `$container`. I would expect something consistent like:
```php
    'container' => [
          yii\web\Request::class => app\components\Request::class,
          yii\web\Response::class => [
              'class' => app\components\Response::class,
              'format' => 'json'
          ],
    ]
````

**Comment by samdark** [comment]
Because we're not setting container options here but setting dependencies into container.

**Comment by SilverFire** [comment]
> ```php
'dependencies' => [
    'definitions' => ...
    'singletons' => ...
],
```

We try to avoid multi-level configurations on top.

How about:
`dicDefinitions`
`dicSingletons`

dynasource:
> Because we're not setting container options here but setting dependencies into container.

mwah...it's not logical.

SilverFire:
> mwah...it's not logical.

Disagreed.

dynasource:
> Disagreed.

then I would expect something like `Yii::$dependencies` instead of `Yii::$container`

This is not consistent. Its like allowing 'obscure' grouping terminology:
* Yii::$box
* Yii::$groups
* Yii::$pool
* Yii::$collection

hiqsol:
@dynasource 
`dependencies` just because of Dependency Injection :)

But I like container variant:

```php
'container' => [
    'definitions' => ...
    'singletons' => ...
],
```

But it's almost the same as original. And original variant has it's advantages: no second level.

samdark:
> mwah...it's not logical.

It's exactly the same as `components`. According to your line of thought it would've been named `serviceLocator` but it's named according to what's held in there.

SilverFire:
Vote

:+1: for
```php
'diDefinitions' => ...,
'diSingletons' => ...,
```

:heart: for
```php
'dicDefinitions' => ...,
'dicSingletons' => ...,
```

:smile: for
```php
'container' => [
    'definitions' => ...,
    'singletons' => ...,
],
```

:-1: for
```php
`containerDefinitions` => ...,
`containerSingletons` => ...
```

samdark:
`di` has nothing to with container. `dic` is, in my opinion, a very wrong abbreviation since in fact it's just dependency container. I don't like `definitions` since by looking at it you have no idea that multiple instances will be created.

SilverFire:
Ну щас тут и срач будет :)

RSalo:
с dic погорячились. если одну букву добавить, то получится ругательство=)

RSalo:
а вообще логичнее всего вариант с 😄. т.к. у нас уже исходя из того, что он лежит в container'е понятно, что это DI...

samdark:
@RSalo try using English please.

SilverFire:
It's better not to switch to another language, бо якщо я перейду на Українську - буде взагалі круто :)

RSalo:
vote is not fair! i can vote several times!

dynasource:
I'd prefer the Symfony logic with 'container' as key

hiqsol:
@samdark `definitions` is consistent with yii\di\Container naming. Can you propose alternatives?

hiqsol:
@dynasource symfony style doesn't work for callback definitions, as I mentioned earlier.

samdark:
https://github.com/yiisoft/yii2/pull/13029#issuecomment-261245614 is a strong argument.

samdark:
@hiqsol I have no alternatives which are consistent with Container implementation.

hiqsol:
@samdark `$_definitions` is private in yii\di\Container and can be renamed easily. So good alternative name can be helpful in both places.

samdark:
The problem with `definitions` is that the term covers both singletons and prototypes. I'd use something like `prototypes` to indicate the fact that these are for multiple instances creation.

dynasource:
from the examples above, I would vote for:
```php
'container' => [
    'definitions' => ...,
    'singletons' => ...,
],
```

however, must say that I think this is overcomplification. 

I would rather choose for Yii automatically handling the logic:
```php
'container' => [

	// usecase 1
	\yii\web\Connection::class => app\components\Connection::class, 
	
	// usecase 2
	\yii\web\Connection::class => [
		'class' => app\components\Connection::class,
	],
	
	// usecase 3
	yii\web\Connection::class => function () {
	},
	
	// usecase 4
	yii\web\Connection::class => [
		'definition' => '',
		'singleton' => '',
	]
],
```

SilverFire:
Oh god :/

SilverFire:
After a vote, [here](yiisoft#13029 (comment))
I've renamed properties from `containerDefinitions` and `containerDefinitions` to a multi-dimensional structure:

```php
'container' => [
    'definitions' => ...,
    'singletons' => ...,
]
```

samdark:
Now the next thing? Tests...

SilverFire:
Pushed tests

framework/base/Application.php
```diff
@@ -246,6 +246,18 @@ public function preInit(&$config)
             $this->setTimeZone('UTC');
         }
 
+        if (isset($config['container'])) {
```
klimov-paul: Looks inconsistent: it would be better to use single setter name.
With such structure confuguration for DI definitions may come from 2 separated config keys:
1) container -> definitions
2) containerDefinitions

framework/base/Application.php
```diff
@@ -456,6 +468,88 @@ public function setVendorPath($path)
...
+     *     'foo\Bar' => [
+     *          ['class' => 'app\Bar'],
+     *          [Instance::of('baz')]
+     *      ]
+     * ]
+     * ```
+     *
+     * @see \yii\di\Container::set() to know more about possible values of definitions
+     */
+    public function setContainerDefinitions(array $definitions)
```
klimov-paul: Perhaps, `setDiDefinitions` would be better.
Word `container` is too general.

framework/base/Application.php
```diff
@@ -456,6 +468,88 @@ public function setVendorPath($path)
...
+     * This method is provided so that you can configure Dependency Injection Container
+     * when configuring an application.
+     *
+     * @param array $singletons array of singleton definitions. See [[setContainerDefinitions]]
+     * for allowed formats of array.
+     *
+     * @see setContainerDefinitions() for allowed formats of $singletons parameter
+     * @see \yii\di\Container::setSingleton() to know more about possible values of definitions
+     */
+    public function setContainerSingletons(array $singletons)
```
klimov-paul: Same here: better to be `setDiSingletons()` or even `setSingletons()`.

klimov-paul (changes requested):

SilverFire:
There is an unexpected behavior caused by `setContainerDefinitions()` and `setContainerSingletons()` methods.

In case when the application config has `containerDefinitions` or `containerSingletons` options in its root, they will be handled by the representative methods. Considering it will happen during `::configure()`, the order will not be guaranteed.

framework/base/Application.php
klimov-paul: Another option can be introduction of the single setter named `setContainer()`, which will always accept array with `definitions` and `singletons` keys.

framework/base/Application.php
SilverFire: https://github.com/yiisoft/yii2/pull/13029#issuecomment-261243153

See @samdark's comment. I agree with it. Do you?

framework/base/Application.php
SilverFire: https://github.com/yiisoft/yii2/pull/13029#issuecomment-261243153

See @samdark's comment. I agree with it. Do you?

SilverFire: `setSingletons()` is not good as for me, because `singletons` belong to the Dependency Injection Container, not the Application itself

klimov-paul: Then it is better to drop `container` word, using `setDefinitions()` and `setSingletons()`.
It will be much more clear in that way,

klimov-paul: > setSingletons() is not good as for me, because singletons belong to the Dependency Injection Container, not the Application itself

Same can be said about `aliases`, which are global and do NOT belong to the application.
The orignal issue was justified using `aliases` example. So I can see nothing bad here.

Either we should use single words `singletons` and `definitions` or deny the feature completely saying those do not belong to the application.

SilverFire:
There are two options how we can solve the problem from my latest comment. The first one - is to leave as is and provide it as additional way to configure DI: 

```php
Yii::$app->setContainerSingletons([...])
```

But I don't think this is a right way, so I will change `setContainerDefinitions()` and `setContainerSingletons()` methods visibility from public to protected.

klimov-paul:
> I will change setContainerDefinitions() and setContainerSingletons() methods visibility from public to protected.

You may not realize it - but it does not changes anything.
There is no check for setter/getter visibility level during magic property processing.
Even being protected those methods will produce virtual properties.

klimov-paul:
As I said you should either introduce **single** setter for the DI container, which will accept array:

```
public function setDiContainer(array $config) {...}
```

Or use single words `definitions` and `singletons` for 2 separated setters.

SilverFire:
@klimov-paul yes, I was wrong about `protected`. Thank you for noticing.

We've discussed this problem with @klimov-paul in IM and summarized the following options:

1. Create `public function setDiContainer()` (or similar) and drop both `setContainerSingletons()` and `setContainerDefinitions()` method, leaving current configuration interface without changes
2. The same as previous, but `setDiContainer()` can not be called from the outside (after application initialization). Probably by renaming to `protected function _setDiContainer()`.
3. Protect `setContainerSingletons()` and `setContainerDefinitions()` from calling after the application initialization, leave all other as is.
4. Just leave all as it is now, do not change anything.

SilverFire:
@klimov-paul yes, I was wrong about `protected`. Thank you for noticing.

We've discussed this problem with @klimov-paul in IM and summarized the following options:

1. Create `public function setDiContainer()` (or similar) and drop both `setContainerSingletons()` and `setContainerDefinitions()` method, leaving current configuration interface without changes
2. The same as previous, but `setDiContainer()` can not be called from the outside (after application initialization). Probably by renaming to `protected function _setDiContainer()`.
3. Protect `setContainerSingletons()` and `setContainerDefinitions()` from calling after the application initialization, leave all other as is.
4. Remove nested `container` configuration and switch back to the original implementation
5. Just leave all as it is now, do not change anything.

cebe:
What about adding `setDefinitions()` and `setSingletons()` to the Container class, consistent with `ServiceLocator::setComponents()` and make `containerDefinitions` simply contain a config for the Container class that can be applied via Yii::configure()?

hiqsol:
I vote for 1a option: `public function setContainer()` this is most consistent with other setters.
I don't see a reason in protecting container from changing through application cause it is changeable through it's own interface.

Taking @cebe's comment, the function will look simply:

```php
public function setContainer(array $properties)
{
    Yii::configure(Yii::$container, $properties);
}

**Comment by SilverFire** [comment]
@cebe thank you for a great idea!
PR is updated

File: docs/guide/concept-di-container.md [hunk 4]
```diff
@@ -224,11 +224,11 @@ and the container will automatically resolve dependencies by instantiating them
 them into the newly created objects. The dependency resolution is recursive, meaning that
 if a dependency has other dependencies, those dependencies will also be resolved automatically.
 
-You can use [[yii\di\Container::get()]] to create new objects. The method takes a dependency name,
+You can use [[yii\di\Container::get()|get()]] to create new objects. The method takes a dependency name,
```
samdark: I think "to create new objects" is not accurate and a bit confusing since there's `setSingleton()`. I'd say "to either create or get object instance".

docs/guide/concept-di-container.md
```diff
@@ -224,11 +224,11 @@ and the container will automatically resolve dependencies by instantiating them
...
 which can be a class name, an interface name or an alias name. The dependency name may or may
-not be registered via `set()` or `setSingleton()`. You may optionally provide a list of class
-constructor parameters and a [configuration](concept-configurations.md) to configure the newly created object.
-For example,
+not be registered via [[yii\di\Container::set()|set()]] or [[yii\di\Container::setSingleton()|setSingleton()]].
+You may optionally provide a list of class constructor parameters and a [configuration](concept-configurations.md)
+to configure the newly created object. For example,
```
samdark: the newly created → obtained

docs/guide/concept-di-container.md
```diff
@@ -368,6 +368,60 @@ cannot be instantiated. This is because you need to tell the DI container how to
...
+```php
+$container->setSingletons([
+    'foo\Bar' => [
+         ['class' => 'app\Bar'],
+         [Instance::of('baz')]
+     ]
+]);
+```
+
+> Note: These methods are available since version 2.0.11
```
samdark: Dot missing at the end of sentence.

framework/di/Container.php
```diff
@@ -566,4 +566,85 @@ public function resolveCallableDependencies(callable $callback, $params = [])
         }
         return $args;
     }
+
+    /**
+     * Registers a class definitions with this container.
```
samdark: Registers a class definitions with → Registers class definitions within

framework/di/Container.php
```diff
@@ -566,4 +566,85 @@ public function resolveCallableDependencies(callable $callback, $params = [])
...
+     *
+     * @param array $definitions array of definitions. There are two allowed formats of array:
```
samdark: formats of array: → formats of array.

framework/di/Container.php
```diff
@@ -566,4 +566,85 @@ public function resolveCallableDependencies(callable $callback, $params = [])
...
+                $this->set($class, $definition[0], $definition[1]);
+                continue;
+            }
+
+            $this->set($class, $definition);
+        }
+    }
+
+    /**
+     * Registers a class definitions with this container and marks them as singletons
```
samdark: Registers class definitions as singletons within this container by calling [[setSingleton()]].

samdark (changes requested):

docs/guide/concept-configurations.md
```diff
@@ -135,6 +135,28 @@ an [entry script](structure-entry-scripts.md), where the class name is already g
...
+            'yii\widgets\LinkPager' => ['maxButtonCount' => 5]
+        ],
+        'singletons' => [
+            // Dependency Injection Container singletons configuration
+        ]
+    ]
+];
+```
+
+To know more about possible values of `definitions` and `singletons` configuration arrays, please
```
dynasource: the possible values

docs/guide/concept-configurations.md
```diff
@@ -135,6 +135,28 @@ an [entry script](structure-entry-scripts.md), where the class name is already g
...
+read [Setting Multiple Dependencies](concept-di-container.md#multiple-dependencies) subsection of
```
dynasource: of the  ... article

docs/guide/concept-di-container.md
```diff
@@ -224,11 +224,11 @@ and the container will automatically resolve dependencies by instantiating them
 them into the newly created objects. The dependency resolution is recursive, meaning that
 if a dependency has other dependencies, those dependencies will also be resolved automatically.
 
-You can use [[yii\di\Container::get()]] to create new objects. The method takes a dependency name,
+You can use [[yii\di\Container::get()|get()]] to create new objects. The method takes a dependency name,
 which can be a class name, an interface name or an alias name. The dependency name may or may
-not be registered via `set()` or `setSingleton()`. You may optionally provide a list of class
-constructor parameters and a [configuration](concept-configurations.md) to configure the newly created object.
-For example,
+not be registered via [[yii\di\Container::set()|set()]] or [[yii\di\Container::setSingleton()|setSingleton()]].
```
dynasource: The dependency name may or may not be registered -> The dependency name may be registered
("may" already implies "maybe not" ;) )

docs/guide/concept-di-container.md
```diff
@@ -372,6 +368,60 @@ cannot be instantiated. This is because you need to tell the DI container how to
...
+ - `value`: array of two elements. The first element will be passed the [[yii\di\Container::set()|set()]] method as the
+ second argument `$definition`, the second one — as `$params`.
+
+Example:
+
+```php
+$container->setSingletons([
+    'foo\Bar' => [
+         ['class' => 'app\Bar'],
+         [Instance::of('baz')]
```
dynasource: this baz example is a bit unclear. Perhabs you can use something like 'componentId'

dynasource (changes requested):
looks good overall. Could you add a section in which you elaborate on the difference between a definition & singleton and when to choose for which one?

docs/guide/concept-di-container.md
SilverFire: Amazing notice. `Instance::of` is not described in guide at all! :scream:

SilverFire:
Indeed I do not like all these `Foo`, `Bar`, `Baz` and their friends. Redone documentation to make it more practical.

framework/base/Application.php
klimov-paul: This entire block is now redundant.

klimov-paul (changes requested):

framework/base/Application.php
SilverFire: It is done to make possible to set `ErrorHandler` in `container['definitions']` and have an effect.
Without this line, ErrorHandler will be registered after `ErrorHandler` instantiating in `registerErrorHandler()` method

docs/guide/concept-di-container.md
```diff
@@ -368,6 +370,107 @@ cannot be instantiated. This is because you need to tell the DI container how to
...
+
+$reader = $container->get('app\storage\DocumentsReader); 
+// Will create DocumentReader object with its dependencies as described in the config 
+```
+
+> Tip: Container may be configured in declarative style using application configuration since version 2.0.11. 
+Check out the [Application Configurations](concept-service-locator.md#application-configurations) subsection of
+the [Configurations](concept-configurations.md) guide article.
+
+All works, but in case we need to create create `DocumentWriter` class,
```
samdark: All works → Everything works

SilverFire:
It is done to make possible to set ErrorHandler in container['definitions'] and have an effect.
Without this line, ErrorHandler will be registered after ErrorHandler instantiating in registerErrorHandler() method

docs/guide/concept-di-container.md
```diff
@@ -368,6 +370,107 @@ cannot be instantiated. This is because you need to tell the DI container how to
...
+ - `key`: class name, interface name or alias name. The key will be passed to the
+ [[yii\di\Container::set()|set()]] method as a first argument `$class`.
+ - `value`: array of two elements. The first element will be passed the [[yii\di\Container::set()|set()]] method as the
+ second argument `$definition`, the second one — as `$params`.
+
+Let's modify our example:
+
+```php
+$container->setDefinitions([
+    'tempFileStorage' => [ // we've created an alias for convinience
```
samdark: convinience → convenience

samdark (changes requested):

klimov-paul (approved):

dynasource (approved):

samdark (approved):

cebe (approved):

SilverFire:
Thank you very much for your input. Merged

schmunk42:
@SilverFire 

**This is absolutely great work!** I am really looking forward to get my hands on this.

PR #13030, opened 2016-11-17
Closes #10675: Docs for disabling CSRF validation in standalone actions

### LINKED ISSUE #10675: Disable csrf validation in Action::beforeRun()
I have a standalone action.
It is designed to work with csrf validation disabled to avoid error 400 "Unable to verify your data submission." As it is standalone I supposed to disable csrf validation in `Action::beforeRun()` method, as:

```
protected function beforeRun() {
        \Yii::$app->controller->enableCsrfValidation = false;
        return true;
}
```

But it takes no effect and seems the only way to do this is to use `Controller::beforeAction()`, but it wisely to have an ability to manage csrf validation in `Action::beforeRun()` for my opinion, since it is called "standalone" action.

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #10675


framework/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ Yii Framework 2 Change Log
 - Bug #12939: Hard coded table names for MSSQL in RBAC migration (arogachev)
 - Bug #12974: Fixed incorrect order of migrations history in case `yii\console\controllers\MigrateController::$migrationNamespaces` is in use (evgen-d, klimov-paul)
 - Enh #6809: Added `\yii\caching\Cache::$defaultDuration` property, allowing to set custom default cache duration (sdkiller)
+- Enh #10675: Docs for disabling CSRF validation in standalone actions (arogachev)
```
rob006: Changes in docs doesn't need changelog line.

docs/guide/security-best-practices.md
```diff
@@ -185,6 +185,30 @@ class SiteController extends Controller
...
+{
+    public function init()
+    {
+        parent::init();
+        \Yii::$app->controller->enableCsrfValidation = false;
+    }
+
+    public function run()
+    {
+        return "Hello World";
```
samdark: It doesn't make sense to provide `run()` implementation like that. Either it should do something with POST or modify state or do something like that. Also instead of hello world I'd do delete action example and explicitly state that it's disabling CSRF is only when external website should be able to submit forms to your website. Also a warning like "Never do it if you do not plan external websites to ...".

samdark (changes requested):

docs/guide/security-best-practices.md
arogachev: @samdark You will probably laugh but in Django in [similar docs](https://docs.djangoproject.com/en/1.10/ref/csrf/#django.views.decorators.csrf.csrf_exempt) the example contains just `Hello world`.

samdark: @arogachev then they're doing it wrong.

arogachev: @samdark This does not mean we should do the same as in Django. And I found it only after adding, was looking for a better example. I will try to fix it according to your remarks.

> then they're doing it wrong.

Not necessary. Maybe it's just a different philosophy and they think people already understand what CSRF is about when reading this section. Also they have some explanation at the very beginning.

Faryshta: How about

```php

class ContactAction extends Action
{
    public function run()
    {
          $model = new ContactForm();
          $request = Yii::$app->request;
          if ($request->referrer == 'yiipowered.com'
              && $model->load($request->post()
              && $model->validate()
          ) {
              $model->sendEmail();
          }
    }
}
```

samdark:
Merged. Thank you everyone.

PR #13031, opened 2016-11-17
FIX index out of range

chrisschd:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes/no
| Tests pass?   | yes
| Fixed issues  | filter function of GridView is locked 

Sometime the Filter of the GridView is block by issue. (User use the back function of the browser)


samdark:
How to reproduce the issue?

chrisschd:
1.) Setup a search model with session stored filter. 
2.) go into the GridView and enter a filter value (for example: filter on color field e.g. "red")
3.) go to another page
4.) press the "back" button of the browser
5.) you got the old filter value and the correct filtered gridview (filter on color field with value "red")
6.) change the filter to "blue" => nothing happens => 	exasperated JavaScript error :(

But I think checking the range before access to an index of an array should be a good idea?

terabytesoftw:
Thanks for the contribution and for reporting this edge case.

This PR addresses the same `yii.getQueryParams()` problem that was later fixed in #17653  with test coverage.

 Since current master already contains that fix.

PR #13039, opened 2016-11-17
urlName don't contain module name. This is bug, if set prefix and con…

mikebelozorov:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | 

…troller contain module

```
class Module extends \yii\base\Module implements BootstrapInterface
{
    public static $controllerNamespace = 'frontend\modules\api_0_2\controllers';

    /**
     * @inheritdoc
     */
    public function bootstrap($app)
    {
        Yii::$app->urlManager->addRules([
            [
                'controller' => [
                    'api_0_2/test',
                ],
                'prefix' => 'api/0.2',
                'extraPatterns' => [
                    'GET test' => 'test',
                ]
            ]
        ]);
    }
}
```

```
rest/UrlRule:148
        $controllers = [];
        foreach ((array) $this->controller as $urlName => $controller) {
            if (is_int($urlName)) {
                $urlName = $this->pluralize ? Inflector::pluralize($controller) : $controller;
            }
            $controllers[$urlName] = $controller;
        }
        $this->controller = $controllers;
```
As a result, the variable $urlName equals "api_0_2/tests"

```
rest/UrlRule:172
            $prefix = trim($this->prefix . '/' . $urlName, '/');
```
As a result, the variable $prefix equals "api/0.2/api_0_2/tests" - this is bug

### DIFF
--- a/framework/rest/UrlRule.php
+++ b/framework/rest/UrlRule.php
@@ -151,6 +151,8 @@ class UrlRule extends CompositeUrlRule
         foreach ((array) $this->controller as $urlName => $controller) {
             if (is_int($urlName)) {
                 $urlName = $this->pluralize ? Inflector::pluralize($controller) : $controller;
+
+                $urlName = trim(substr($urlName, (int)strpos($urlName, '/')), '/');
             }
             $controllers[$urlName] = $controller;
         }


tunecino:
that is a BC and will break the already working sites using the same class with different rules. 

$urlName is the $controller array key inside the foreach loop and will be set by the framework only if it is an integer. So you just need to set it to a different value. It is documented in the [Rest Routing section](http://www.yiiframework.com/doc-2.0/guide-rest-routing.html):


> In case when the automatic pluralization does not meet your requirement, you may also configure the yii\rest\UrlRule::$controller property to explicitly specify how to map a name used in endpoint URLs to a controller ID. For example, the following code maps the name u to the controller ID user.
> 
> ```php
>  [
>     'class' => 'yii\rest\UrlRule',
>     'controller' => ['u' => 'user'],
> ]
> ```

tunecino:
Not sure what Url Endpoint are you expecting but doesn’t the following what you are trying to get ?

```php
Yii::$app->urlManager->addRules([
    [
        'controller' => [
            '0.2/tests' => 'api_0_2/test',
        ],
        'prefix' => 'api',
        'extraPatterns' => [
            'GET test' => 'test',
        ]
    ]
]);
```

cebe:
@mikebelozorov the change does not look right to me, please open an issue describing the problem in more detail.

PR #13041, opened 2016-11-18
Update ColumnSchema.php

Slayer911:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes

returning as default column value - null when column has option "NOT NULL"

### DIFF
--- a/framework/db/ColumnSchema.php
+++ b/framework/db/ColumnSchema.php
@@ -114,7 +114,11 @@ class ColumnSchema extends Object
     protected function typecast($value)
     {
         if ($value === '' && $this->type !== Schema::TYPE_TEXT && $this->type !== Schema::TYPE_STRING && $this->type !== Schema::TYPE_BINARY && $this->type !== Schema::TYPE_CHAR) {
-            return null;
+            if($this->allowNull){
+                return null;
+            }else{
+                return '';
+            }
         }
         if ($value === null || gettype($value) === $this->phpType || $value instanceof Expression) {
             return $value;


dynasource:
please elaborate on your PR

Slayer911:
When I try insert new row by ActiveRecord

$model = new Model();
$model->event_id = 3652;
$model->date_in = '';
$model->save();

I obtain an error :
"Integrity constraint violation: 1048 Column 'date_in' cannot be null
The SQL being executed was: INSERT INTO `module_event_date` (`date_in`, `event_id`) VALUES (NULL, 3652)'"

created sql "'INSERT INTO `module_event_date` (`date_in`, `event_id`) VALUES (:qp0, :qp1)"

Attributes for request :
[
    ':qp0' => null
    ':qp1' => 3652
]

Value "null" set by function yii\db\ColumnSchema->typecast()

Table 
  event_id int(11) NOT NULL,
  date_in int(11) NOT NULL

SilverFire:
Which rules do you have for `date_in` attribute?

Slayer911:
only 'integer' rule

Slayer911:
I think you skipped, that integer column type in Mysql with "NO NULL" have default value - empty string, same as TEXT and VARCHAR.

ColumnSchema.php on line 116
`if ($value === '' && $this->type !== Schema::TYPE_TEXT && $this->type !== Schema::TYPE_STRING && $this->type !== Schema::TYPE_BINARY && $this->type !== Schema::TYPE_CHAR){
            return null;
}`
you check only for string types, but not the integer

cebe:
inserting  a string into a int column is just wrong and should fail.

Slayer911:
Helper for work with DB must duplicate database rules, instead of coming up with their own.
For example a number of string format - in yii2 passes validation, and insert data into the database
$model->date_in = '123';

cebe:
`'123'` is fine. An empty string should not pass the integer validator though.

Slayer911:
NumberValidator.php does not override variable $skipOnEmpty, whcih have default value "true" in Validator.php . 
That why this column does not validate, and instead generate ActiveRecord error I have MySql error.

cebe:
if the column requirement is "NOT NULL" you should add a `required` validator.

PR #13043, opened 2016-11-18
Fix #13042

### LINKED ISSUE #13042: Menu: isItemActive bug with module route
I believe to add url for module route in the config of Menu widget is just using

```
'items' => [
      [
        'label' => 'Menu for Module',
        'url' => ['module/controller/action'],
      ],
]
```

As i look at function `isItemActive` in the Menu class line 311, it assumes that the defined url is without module id. So, the config above does not activate item on url `localdomain.dev/module/controller/action`. I need to define the url to only `['controller/action']` to activate that menu inside the module. But this will render `/controller/action` in the menu link instead.

I think this is not right. So, i use workaround just remove the line

```
$route = Yii::$app->controller->module->getUniqueId() . '/' . $route;
```

and it works just like i want to using the correct url in config. I will send pull request if it is accepted.

haqqi:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13042

### DIFF
--- a/framework/widgets/Menu.php
+++ b/framework/widgets/Menu.php
@@ -308,7 +308,7 @@ class Menu extends Widget
         if (isset($item['url']) && is_array($item['url']) && isset($item['url'][0])) {
             $route = Yii::getAlias($item['url'][0]);
             if ($route[0] !== '/' && Yii::$app->controller) {
-                $route = Yii::$app->controller->module->getUniqueId() . '/' . $route;
+              
             }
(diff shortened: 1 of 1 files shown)


dynasource:
please elaborate

SilverFire:
See https://github.com/yiisoft/yii2/issues/13042#issuecomment-261536846

PR #13057, opened 2016-11-22
FileTarget::$rotateByCopy: if not specified, automatically set to false when OS is not windows

felix021:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

FileTarget::$rotateByCopy: if not specified, automatically set to false when OS is not windows

### DIFF
--- a/framework/log/FileTarget.php
+++ b/framework/log/FileTarget.php
@@ -70,7 +70,7 @@ class FileTarget extends Target
      * the PHP documentation. By setting rotateByCopy to `true` you can work
      * around this problem.
      */
-    public $rotateByCopy = true;
+    public $rotateByCopy = null;
 
 
     /**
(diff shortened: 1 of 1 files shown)


samdark:
Why?

felix021:
In our application on Yii 2, maxFileSize is set to be 1 GB (with 10 log files, the default value), and by default `$rotateByCopy = true`, which means, when log files need to be rotated, there may be as much as 9 GB of logs to be read into memory, and write to disk, which is meaningless on Linux servers and causes large pressure on disk IO, and thus make the server not responding to requests. Automatic OS detection in this pull request helps improve situation on non-windows servers.

cebe:
you should choose a smaller file size then.

felix021:
I can change it to 100 MB, but 900 MB of disk io will still take my server down for over 1 minute, and as is described in the comments, this flag is set to true because rotation fails *ONLY* on Windows, so why don't we just set it to false on non-windows servers to avoid such condition? And I have to say, that you guys didn't warn developers to use smaller maxFileSize in advance either, which I spent days to find out the reason of downtime. This is not friendly to yii users.

cebe:
Okay now I understand the problem. Thanks for explaining. It makes sense now.

samdark:
Applied to 3.0.

PR #13067, opened 2016-11-24
Add support for creating protocol-relative URLs

### LINKED ISSUE #12738: Url::to("/", "//") or how to generate absolute links without protocol
The UrlManager does not support currently generating absolute links without the protocol part.
This is pretty common to avoid "mixed content" messages in browsers.

I have a domain `http://domain.com` so I need to get from UrlManager `//domain.com`.

Following is current output:

```
Url::to('//', true) ---> "//"
Url::to('/', '//') ---> "//://domain.com/"
Url::to('/', 1) ---> "http://domain.com/" or "https://domain.com/"
Url::to('//', 1) ---> "//"
```

So no proper protocolless absolute link.

Did I miss something or can this feature be added?
Thank you.

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12738

Closes #12738.


samdark:
What do you think about using `//` instead of empty string?

rob006:
I think that empty string is more intuitive. `https` means that schema is `https`, empty string means that there is no schema.

samdark (approved):
Looks good to me except parameter empty value. I have no better proposals than `//` so won't object on having empty string.

SilverFire (approved):

framework/helpers/BaseUrl.php
```diff
@@ -86,8 +86,8 @@ class BaseUrl
      * @param bool|string $scheme the URI scheme to use in the generated URL:
      *
      * - `false` (default): generating a relative URL.
-     * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::hostInfo]].
-     * - string: generating an absolute URL with the specified scheme (either `http` or `https`).
+     * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::$hostInfo]].
+     * - string: generating an absolute URL with the specified scheme (either `http`, `https` or empty string).
```
cebe: maybe add a hint `or empty string, for protocol relative URL`? it is not directly obvious why one would pass an empty string here, also searching for "yii protocol relative url" would find this text better than now.

framework/helpers/BaseUrl.php
```diff
@@ -214,23 +217,46 @@ public static function to($url = '', $scheme = false)
...
+
+        if (substr($url, 0, 2) === '//') {
+            // e.g. //example.com/path/to/resource
+            return $scheme === '' ? $url : "$scheme:$url";
+        }
+
+        if (($pos = strpos($url, '://')) !== false) {
+            if ($scheme === '') {
+                $url = substr($url, $pos + 1);
+            } elseif (is_string($scheme)) {
```
cebe: is_string check is not needed here, it is already checked in line 245

framework/helpers/BaseUrl.php
```diff
@@ -320,20 +345,18 @@ public static function canonical()
      * @param bool|string $scheme the URI scheme to use for the returned URL:
      *
      * - `false` (default): returning a relative URL.
-     * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::hostInfo]].
-     * - string: returning an absolute URL with the specified scheme (either `http` or `https`).
+     * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::$hostInfo]].
+     * - string: returning an absolute URL with the specified scheme (either `http`, `https` or empty string).
```
cebe: same here.

framework/helpers/BaseUrl.php
```diff
@@ -387,8 +410,8 @@ public static function isRelative($url)
      * @param bool|string $scheme the URI scheme to use in the generated URL:
      *
      * - `false` (default): generating a relative URL.
-     * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::hostInfo]].
-     * - string: generating an absolute URL with the specified scheme (either `http` or `https`).
+     * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::$hostInfo]].
+     * - string: generating an absolute URL with the specified scheme (either `http`, `https` or empty string).
```
cebe: and here :)

framework/web/UrlManager.php
```diff
@@ -452,8 +453,8 @@ protected function setRuleToCache($cacheKey, UrlRuleInterface $rule)
      *
      * @param string|array $params use a string to represent a route (e.g. `site/index`),
      * or an array to represent a route with query parameters (e.g. `['site/index', 'param1' => 'value1']`).
-     * @param string $scheme the scheme to use for the url (either `http` or `https`). If not specified
-     * the scheme of the current request will be used.
+     * @param string|null $scheme the scheme to use for the URL (either `http`, `https` or empty string).
```
cebe: and here :)

cebe (changes requested):
good changes! only some minor improvements needed.

cebe:
I like using empty string for no scheme.

cebe (approved):

PR #13070, opened 2016-11-25
Update examples for events using interfaces

sizeg:
Where was a mistake in example for events using interfaces. As events has no method [[className()]] and when called [[Event::trigger()]] we should pass first argument class name implemented this interface, not the interface itself.

### DIFF
--- a/docs/guide/concept-events.md
+++ b/docs/guide/concept-events.md
@@ -296,11 +296,6 @@ pass the interface name as the first argument:
 Event::on('DanceEventInterface', DanceEventInterface::EVENT_DANCE, function ($event) {
     Yii::trace($event->sender->className . ' just danced'); // Will log that Dog or Developer danced
 });
-
-// or with fully qualified interface name
-Event::on(DanceEventInterface::class, DanceEventInterface::EVENT_DANCE, function ($event) {
-    Yii::trace($event->sender->className . ' just danced'); // Will log that Dog or Developer danced
-});
 ```
 
 You can trigger the event of those classes:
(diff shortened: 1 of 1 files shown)


docs/guide/concept-events.md
```diff
@@ -295,13 +295,22 @@ pass the interface name as the first argument:
 ```php
 Event::on('DanceEventInterface', DanceEventInterface::EVENT_DANCE, function ($event) {
     Yii::trace($event->sender->className . ' just danced'); // Will log that Dog or Developer danced
-})
+});
+
+// or with fully qualified interface name
+Event::on(DanceEventInterface::class, DanceEventInterface::EVENT_DANCE, function ($event) {
```
cebe: `::class` does not work on php 5.4 so it should not be used in 2.0.x docs.

cebe (changes requested):

samdark:
It should pass interface name, I think. W/o using `::className()`, of course.

sizeg:
> It should pass interface name

@samdark then it will conflict with the next paragrath
```php
// DOES NOT WORK
Event::trigger('DanceEventInterface', DanceEventInterface::EVENT_DANCE); // error
```

sizeg:
We should pass to [[Event::trigger()]] a fully qualified class name implemented our event interface. By the way I'm using it this way.

According to the previouse examples from doc, were two new objects created from Component, we can use [[className()]] I think.

samdark:
@cebe need your assistance in this one...

dynasource (approved):
these updates are correct

SilverFire:
Resolved by f1f357b

SilverFire:
@sizeg thank you for noticing

Issue #13074, opened 2016-11-25
Unable to change openlog $option attribute for syslogtarget


timbeks:
The SyslogTarget uses openlog for writing to the syslog. The openlog method requires some attributes; one of them $option. This attribute is hardcoded in the SyslogTarget and can't be overwritten.

For my project(s) I don't want to use the LOG_PID, because this changes the syslog filename.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.0-beta - 2.1
| PHP version      | Irrelevant
| Operating system | Irrelevant

PR #13075, opened 2016-11-25
Add attribute to set the openlog $option attribute for openlog

timbeks:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes (Only tested group log, because only logging has been touched)
| Fixed issues  | #13074

### DIFF
--- a/framework/log/SyslogTarget.php
+++ b/framework/log/SyslogTarget.php
@@ -26,6 +26,10 @@ class SyslogTarget extends Target
      * @var int syslog facility.
      */
     public $facility = LOG_USER;
+    /**
+     * @var int openlog option
+     */
+    public $option = LOG_ODELAY | LOG_PID;
 
     /**
      * @var array syslog levels
@@ -40,13 +44,12 @@ class SyslogTarget extends Target
         Logger::LEVEL_ERROR => LOG_ERR,
     ];
 
-
     /**
      * Writes log messages to syslog
      */
     public function export()
     {
-        openlog($this->identity, LOG_ODELAY | LOG_PID, $this->facility);
+        openlog($this->identity, $this->option, $this->facility);
         foreach ($this->messages as $message) {
             syslog($this->_syslogLevels[$message[1]], $this->formatMessage($message));
         }
--- a/framework/log/SyslogTarget.php
+++ b/framework/log/SyslogTarget.php
@@ -51,7 +51,9 @@ class SyslogTarget extends Target
      */
     public function init()
     {
-         $this->options = LOG_ODELAY | LOG_PID;
+        if ($this->options === null) {
+            $this->options = LOG_ODELAY | LOG_PID;
+        }
     }
 
     /**
--- a/framework/log/SyslogTarget.php
+++ b/framework/log/SyslogTarget.php
@@ -27,9 +27,11 @@ class SyslogTarget extends Target
      */
     public $facility = LOG_USER;
     /**
-     * @var int openlog option
+     * @var int openlog options
+     * @see http://php.net/openlog
+     * @since 2.0.11
      */
-    public $option = LOG_ODELAY | LOG_PID;
+    public $options = LOG_ODELAY | LOG_PID;
 
     /**
      * @var array syslog levels
@@ -49,7 +51,7 @@ class SyslogTarget extends Target
      */
     public function export()
     {
-        openlog($this->identity, $this->option, $this->facility);
+        openlog($this->identity, $this->options, $this->facility);
         foreach ($this->messages as $message) {
             syslog($this->_syslogLevels[$message[1]], $this->formatMessage($message));
         }
--- a/framework/log/SyslogTarget.php
+++ b/framework/log/SyslogTarget.php
@@ -49,7 +49,8 @@ class SyslogTarget extends Target
     /**
      * @inheritdoc
      */
-    public function init() {
+    public function init()
+    {
          $this->options = LOG_ODELAY | LOG_PID;
     }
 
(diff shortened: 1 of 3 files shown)


framework/log/SyslogTarget.php
```diff
@@ -26,6 +26,10 @@ class SyslogTarget extends Target
      * @var int syslog facility.
      */
     public $facility = LOG_USER;
+    /**
+     * @var int openlog option
+     */
+    public $option = LOG_ODELAY | LOG_PID;
```
cebe: I would prefer the name `$options` as these are multiple options that can be combined.

framework/log/SyslogTarget.php
```diff
@@ -26,6 +26,10 @@ class SyslogTarget extends Target
      * @var int syslog facility.
      */
     public $facility = LOG_USER;
+    /**
+     * @var int openlog option
```
cebe: documentation should describe the property and list the possible values or refer to the php documentation.
Also `@since 2.0.11` phpdoc annotation is missing.

cebe (changes requested):

framework/log/SyslogTarget.php
timbeks: @cebe, I agree. But PHP5.4 and PHP 5.5 does not allow LOG_ODELAY | LOG_PID. So what do you suggest? Set just one value, or set the default value in the init?

timbeks:
What should I do with the unit tests? I updated the SyslogTargetTest.php with the $options attribute (not pushed yet), but the openlog method is a boolean output that is not caught any where. Plus the testExport() covers openlog method.

samdark:
> not pushed yet

Push it?

samdark:
So test shows there are PHP 5.4 compatibility problems. These should be resolved.

timbeks:
@samdark; I added a comment (https://github.com/yiisoft/yii2/pull/13075#discussion_r89728350) which pointed this out. Unfortunately, this can be resolved in two ways;

1. Set the default value, for example, to LOG_ODELAY. This will revoke BC a bit.
2. Set the default value within the init().
    It does not result in any issues if you set the values in an (configuration) array with bitwise comparison.

How would you suggest in solving this?

samdark:
`#2` since breaking compatibility isn't an option.

framework/log/SyslogTarget.php
```diff
@@ -47,6 +47,13 @@ class SyslogTarget extends Target
     ];
 
     /**
+     * @inheritdoc
+     */
+    public function init() {
```
samdark: Should be formatted properly i.e. `{` on the next line.

samdark (changes requested):

samdark (approved):

framework/log/SyslogTarget.php
```diff
@@ -40,13 +46,20 @@ class SyslogTarget extends Target
         Logger::LEVEL_ERROR => LOG_ERR,
     ];
 
+    /**
+     * @inheritdoc
+     */
+    public function init()
+    {
+         $this->options = LOG_ODELAY | LOG_PID;
```
samdark: Wait, this would prevent setting your own options. Should be:

```php
if ($this->options === null) {
    $this->options = LOG_ODELAY | LOG_PID;
}
```

samdark (changes requested):

framework/log/SyslogTarget.php
timbeks: Whoops, you are totally right.

samdark (approved):

framework/log/SyslogTarget.php
```diff
@@ -26,6 +26,12 @@ class SyslogTarget extends Target
      * @var int syslog facility.
      */
     public $facility = LOG_USER;
+    /**
+     * @var int openlog options
+     * @see http://php.net/openlog
+     * @since 2.0.11
```
cebe: better phpdoc:


```
@var int openlog options. This is a bitfield passed as the `$option` parameter to [openlog()](http://php.net/openlog). Defaults to `null` which means to use the default options `LOG_ODELAY | LOG_PID`.
@see http://php.net/openlog for available options.
```

framework/log/SyslogTarget.php
```diff
@@ -40,13 +46,22 @@ class SyslogTarget extends Target
...
 
+    /**
+     * @inheritdoc
+     */
+    public function init()
+    {
+        if ($this->options === null) {
+            $this->options = LOG_ODELAY | LOG_PID;
+        }
+    }
```
cebe: call to parent::init() is missing. I know it is empty right now but later code changes may add something to it.

cebe (changes requested):
also needs an entry in the CHANGELOG file.

framework/log/SyslogTarget.php
```diff
@@ -68,7 +69,7 @@ public function export()
         closelog();
     }
 
-    /**
+    /**https://www.bol.com/nl/account/orders.html#account_bestellingen
```
cebe: this seems accidentally added here.

timbeks: That's stupid of me! Sorry about that

SilverFire (approved):

framework/log/SyslogTarget.php
cebe: no problem :)

cebe (approved):

cebe:
Merged. Thank you!

PR #13081, opened 2016-11-26
Fix Yii::createObject 

lav45:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | 

If passed the parameters to be passed to the constructor `Yii::createObject` simply ignores them.

### DIFF
--- a/framework/di/Container.php
+++ b/framework/di/Container.php
@@ -430,7 +430,7 @@ class Container extends Component
                     $dependencies[] = $param->getDefaultValue();
                 } else {
                     $c = $param->getClass();
-                    $dependencies[] = Instance::of($c === null ? null : $c->getName());
+                    $dependencies[] = $c === null ? null : Instance::of($c->getName());
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


cebe:
can you please elaborate about the problem?

LAV45:
```php
Yii::createObject([
    'class' => '\yii\db\Expression',
    'expression' => '[[quantity]]+:bp0',
    'params' => [
        ':bp0' => -1
    ]
]);
```
When I try to create a class using `Yii::createObject` error
> yii\base\InvalidConfigException: Missing required parameter "expression" when instantiating "yii\db\Expression".

because the parameter `$expression` is required in the method `yii\db\Expression::__construct()`

maybe I'm wrong trying to solve the problem, then how to do it.

cebe:
Constructor params are the second argument, so you should use it like:

```php
Yii::createObject('\yii\db\Expression', [
    '[[quantity]]+:bp0', // expression
    [':bp0' => -1], // params
]);
```

LAV45:
Thanks!

Issue #13087, opened 2016-11-28
Why yii\base\Model::getActiveValidators($attribute) don`t return validator, if i add '!' as prefix to attribute name in rules.


porshkevich:
if i add next to rules:

```php
[['!coinsImage'], 'file', 'skipOnEmpty' => false, 'extensions' => 'png, jpg'],
```
i will exclude it from massive assignment, but can`t get that validator when use:

```php
$model->getActiveValidators('coinsImage');
```

I think next condition wrong:

```php
if ($validator->isActive($scenario) && ($attribute === null || in_array($attribute, $validator->attributes, true))) {
                $validators[] = $validator;
            }
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10

developeruz:
Workaround: 
```$model->getActiveValidators('!coinsImage');```
but it is not obvious.
You can find a proper fix in my PR #13088

Issue #12762, opened 2016-10-17
Request for making validators accessible by associative key


dynasource:
The more I am working with validators, the more I need to have access to validators that are set. The current implementation is non-associative which makes accessing hard. Additionally, when debugging ActiveRecords with ie. print_r , its hard to tell which validator is which.

Therefore I would like to suggest to make the validators accessible by associative key. This would not have to be BC as the current implementation does not really support such an access.

samdark:
Can't you do it now?

dynasource:
Nope.

The non-associative append function of ArrayObject (http://php.net/manual/en/arrayobject.append.php)  is used:
* https://github.com/yiisoft/yii2/blob/master/framework/base/Model.php#L443

This is consistently processed further in Model.php like i.e. https://github.com/yiisoft/yii2/blob/master/framework/base/Model.php#L424

samdark:
I see. What do you plan to do with validators which are set?

dynasource:
imagine usecases in which you need to know which Validators are active for a Model and to display it to the user. Perhabs you want to do some specific ordering. Perhabs you want to build widgets based on these validators. Then you want to know which one is which.

dynasource:
I already have this functionality in operation for a while now and its very useful for 'reading' purposes too. It enables you to read 'rules' faster. It also helps debugging faster.

Example:
```php
                $rules = [
                    
                    'title-required' => ['title', 'required'],
                    'title-string' => ['title', 'string', 'max' => 100],
                    'title-ucfirst' => ['title', 'filter', 'filter' => 'ucfirst'],
                    'title-purifier' => ['title', 'filter', 'filter' => [Purifier::class, Purifier::ALPHA_NUMERIC_LATIN]],
                    'title-unique' => ['title', 'unique', 'when' => function(){
                        return $this->isTitleUnique();
                    }],
                   
```

samdark:
I've tried to imagine use case and failed to. More readability could be achieved with comments. Having keys in debug info doesn't add much to the overall picture.

dynasource:
Another usecase for associative keys is the overriding of rules() in child ARs. There are cases thinkable in which rules have to be disabled or changed. 

The same applies for attached behaviors who need to check for specific rules.

Last but definately not least. I would foster the appraoch to define unique names for rules. If you have many Active Records, its way easier to compare rules between them. You instantely see if you have set all the right rules for similar attributes and if you are missing any which enhances security & consistency.

[...]

PR #13088, opened 2016-11-28
Fixed getting active validators for safe attribute (#13087)

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13087 

`$model->getActiveValidators('safe_attr')` doesn't return validator if attribute is declared as safe (`['!safe_attr', 'integer']`)

Introduced new `$attributeNames` array that can be used to simplify this repatitive code
```php
if (in_array($attribute, $this->attributes) || in_array('!' . $attribute, $this->attributes))
``` 
into 
```php
if (in_array($attribute, $this->attributeNames))
```


framework/validators/Validator.php
```diff
@@ -102,6 +102,10 @@ class Validator extends Component
      */
     public $attributes = [];
     /**
+     * @var array cleaned attribute names with "!" characters in the beginning removed
+     */
+    public $attributeNames = [];
```
samdark: Should it be public?

samdark (changes requested):

framework/validators/Validator.php
developeruz: No, of course, it should be protected.

framework/validators/Validator.php
```diff
@@ -415,4 +419,26 @@ public function isEmpty($value)
...
+     */
+    public function getAttributeNames()
+    {
+        return $this->attributeNames;
+    }
+
+    /**
+     * @param array $attributeNames
+     */
+    public function setAttributeNames($attributeNames)
```
samdark: Should it be public? It's not used anywhere outside of the class...

samdark (changes requested):

framework/validators/Validator.php
```diff
@@ -102,6 +102,10 @@ class Validator extends Component
      */
     public $attributes = [];
     /**
+     * @var array cleaned attribute names with "!" characters in the beginning removed
```
samdark: Need `@since <version>` tag where `<version>` is Yii version it's merged in.

framework/validators/Validator.php
```diff
@@ -415,4 +419,26 @@ public function isEmpty($value)
             return $value === null || $value === [] || $value === '';
         }
     }
+
+    /**
+     * @return array
```
samdark: Need `@since <version>` tag where `<version>` is Yii version it's merged in.

Also need proper phpdoc.

framework/validators/Validator.php
```diff
@@ -415,4 +419,26 @@ public function isEmpty($value)
...
+     */
+    public function getAttributeNames()
+    {
+        return $this->attributeNames;
+    }
+
+    /**
+     * @param array $attributeNames
```
samdark: Need `@since <version>` tag where `<version>` is Yii version it's merged in.

Also need proper phpdoc.

samdark (changes requested):

framework/validators/Validator.php
```diff
@@ -102,6 +102,10 @@ class Validator extends Component
      */
     public $attributes = [];
     /**
+     * @var array cleaned attribute names with "!" characters in the beginning removed
+     */
+    protected $attributeNames = [];
```
cebe: property should be private as there is a getter and a setter to work with it, there is no need to make it protected.

samdark: Sub-classing setter?

cebe: if that is ever needed you can add a getter too and use private var in subclass.

developeruz:
All is fixed, please, review again

samdark (approved):

framework/validators/Validator.php
```diff
@@ -415,4 +419,30 @@ public function isEmpty($value)
             return $value === null || $value === [] || $value === '';
         }
     }
+
+    /**
+     * Returns cleaned attribute names with "!" characters in the beginning removed
```
SilverFire: How about

<code>Returns cleaned attribute names without the \`!\` character at the beginning</code>

framework/validators/Validator.php
```diff
@@ -415,4 +419,30 @@ public function isEmpty($value)
...
+     * @return array
+     * @since 2.0.12
+     */
+    public function getAttributeNames()
+    {
+        return $this->attributeNames;
+    }
+
+    /**
+     * Saves attribute names with "!" characters in the beginning removed as an array
```
SilverFire: How about

<code>Saves attribute names without \`!\` character at the beginning</code>

framework/validators/Validator.php
```diff
@@ -233,30 +238,29 @@ public function init()
         $this->attributes = (array) $this->attributes;
         $this->on = (array) $this->on;
         $this->except = (array) $this->except;
+        $this->setAttributeNames($this->attributes);
```
SilverFire: `(array)$this->attributes`

framework/validators/Validator.php
```diff
@@ -415,4 +419,30 @@ public function isEmpty($value)
...
+     * @since 2.0.12
+     */
+    public function getAttributeNames()
+    {
+        return $this->attributeNames;
+    }
+
+    /**
+     * Saves attribute names with "!" characters in the beginning removed as an array
+     * @param array|string $attributeNames
```
SilverFire: `|string` is redundant

framework/validators/Validator.php
```diff
@@ -415,4 +419,30 @@ public function isEmpty($value)
...
+     * @since 2.0.12
+     */
+    private function setAttributeNames($attributeNames)
+    {
+        if (is_array($attributeNames)) {
+            $this->attributeNames = array_map(function($attribute){
```
SilverFire: space before `{` is missing

framework/validators/Validator.php
```diff
@@ -415,4 +419,30 @@ public function isEmpty($value)
...
+    }
+
+    /**
+     * Saves attribute names with "!" characters in the beginning removed as an array
+     * @param array|string $attributeNames
+     * @since 2.0.12
+     */
+    private function setAttributeNames($attributeNames)
+    {
+        if (is_array($attributeNames)) {
```
SilverFire: Considering previous comments, this conditions is redundant.

SilverFire (changes requested):

developeruz:
@SilverFire , please, review again. Let me know if anything is missing. Thanks!

SilverFire (approved):

tests/framework/validators/ValidatorTest.php
```diff
@@ -214,4 +214,11 @@ public function testAddError()
         $errors = $m->getErrors('attr_msg_val');
         $this->assertEquals('attr_msg_val::abc::param_value', $errors[0]);
     }
+
+    public function testGetActiveValidatorsForSafeAttributes()
+    {
+        $model = $this->getTestModel();
+        $validators = $model->getActiveValidators('safe_attr');
+        $this->assertInstanceOf('yii\validators\NumberValidator', array_shift($validators));
```
dynasource: will this test still work when someone adds a new rule to FakedValidationModel? Would be made more stable when assigned to a 'key'

dynasource (changes requested):

tests/framework/validators/ValidatorTest.php
developeruz: ``getActiveValidators`` has integer [keys.](https://github.com/yiisoft/yii2/blob/master/framework/base/Model.php#L428) It wouldn't help. One solution that I can come with is a loop of all ActiveValidators and if at least one of them is Instance of NumberValidator, then the test is passed. 
```php
public function testGetActiveValidatorsForSafeAttributes()
    {
        $model = $this->getTestModel();
        $validators = $model->getActiveValidators('safe_attr');
        $is_found = false;
        foreach ($validators as $v) {
            if ($v instanceof NumberValidator) {
                $is_found = true;
                break;
            }
        }
        $this->assertTrue($is_found);
    }
```
Maybe you can suggest a better solution?

dynasource: have you thought about adding a constant in your test classes and add that one to FakedValidationModel as a key?

developeruz: Sorry, I don't understand what you mean. 
Even if I will add  `'some_key' => ['!safe_attr', 'integer']` in `FakedValidationModel ` rules it won't help and I won't get this key from `getActiveValidators`. 
Could you please provide me some example. I will really appreciate it.

dynasource: ahh, you are right. That won't work because the keys are not kept in sync. Related issue topic is https://github.com/yiisoft/yii2/issues/12762

Your new loop is fine.

dynasource (approved):

samdark:
Merged. Thanks!

framework/validators/Validator.php
```diff
@@ -102,6 +102,11 @@ class Validator extends Component
      */
     public $attributes = [];
     /**
+     * @var array cleaned attribute names. Contains attribute names without `!` character at the beginning
+     * @since 2.0.12
+     */
+    private $attributeNames = [];
```
cebe: naming is wrong, private property should start with `_`.

framework/validators/Validator.php
```diff
@@ -233,30 +238,29 @@ public function init()
...
     }
 
     /**
      * Validates the specified object.
      * @param \yii\base\Model $model the data model being validated
      * @param array|null $attributes the list of attributes to be validated.
-     * Note that if an attribute is not associated with the validator, or is is prefixed with `!` char - it will be
+     * Note that if an attribute is not associated with the validator - it will be
      * ignored. If this parameter is null, every attribute listed in [[attributes]] will be validated.
+     * @since 2.0.12
```
cebe: the method existed before 2.0.12.

PR #13089, opened 2016-11-29
Do not replace URL started with # using yii asset

vitalyzhakov:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

Using CSS file with rule
```
clip-path: url('#xxx')
```

yii asset command append path to this url, but it is not correct.

### DIFF
--- a/framework/console/controllers/AssetController.php
+++ b/framework/console/controllers/AssetController.php
@@ -638,7 +638,7 @@ EOD;
             $fullMatch = $matches[0];
             $inputUrl = $matches[1];
 
-            if (strpos($inputUrl, '/') === 0 || preg_match('/^https?:\/\//i', $inputUrl) || preg_match('/^data:/i', $inputUrl)) {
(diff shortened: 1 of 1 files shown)


SilverFire:
According to our  [roadmap](https://github.com/yiisoft/yii2/wiki/Plan-for-next-major-versions), we are willing to drop `AssetCompressor` in our closest major release (v. 2.1), therefore we suggest not to use it for your projects.

@yiisoft/core-developers should we accept this PR?

dynasource:
That AssetController is actually quite useful to me. I would not see it dropped in 2.1. If it is dropped in 2.1, I still would like to use it as an extension. Regarding this, adopting PR's and/or fixes is still useful.

One can ask, why arent you using a Gulp pipeline instead? Well, I have a fully stacked gulp pipeline, but I am dependant on Yii2 for the Assets processing.

samdark:
Well, we're not going to drop it... maybe move it out of the core.

samdark (changes requested):
Change itself looks good. Need a line for changelog at least. Ideally — tests.

klimov-paul:
Resolved by commit 18b0dab360bcbcd41a7bd152ac6c66844746242f

Issue #10583, opened 2016-01-15
the at sign (@) session_regenerate_id


gangbo:
- php7.0.2
- yii2.0.6

`yii2/web/Session.php`
```
public function regenerateID($deleteOldSession = false)
{
    // add @ to inhibit possible warning due to race condition
    // https://github.com/yiisoft/yii2/pull/1812
    @session_regenerate_id($deleteOldSession);
}
```
I know this is a php7's bug https://bugs.php.net/bug.php?id=71187 .
But when I use yii2 debug, there is no any error messages expect http 500.

githubjeka:
+1

See
https://github.com/yiisoft/yii2/issues/10325
https://github.com/yiisoft/yii2/issues/10489
https://github.com/yiisoft/yii2-redis/pull/46

SilverFire:
I can't suggest the ideal solution for the problem.

Maybe, we can remove stfu when `YII_DEBUG` is true...

githubjeka:
SilverFire nice !

SilverFire:
@code

SilverFire:
@yiisoft/core-developers

samdark:
That's a good idea.

georgekaf:
I can confirm this issue with php7.0.2, yii2.0.6 and yii\redis\Session. I had to switch to yii\web\CacheSession for now.

githubjeka:
@georgekaf try update yii2-redis to `"yiisoft/yii2-redis": "dev-master",`
https://github.com/yiisoft/yii2-redis/commit/3c8c6a42c0177a244b5bb878d60c3ad1d2d14f96

georgekaf:
@githubjeka Seems to be working properly, thanks!

tanakahisateru:
My workaround for PHP7's recoverable fatal error:

```php
class Php7SafeSession extends Session
{
    public function regenerateID($deleteOldSession = false)
    {
        // PHP7
        if (interface_exists('\Throwable')) {
            try {
                session_regenerate_id($deleteOldSession);
            } catch (\Throwable $t) { /* Check error content here */ }
        } else {
            @session_regenerate_id($deleteOldSession);
        }
    }
```

adamdry:
I'm using Yii 2.0.6 - \yii\web\Session->regenerateID looks like this:

```    
    public function regenerateID($deleteOldSession = false)
    {
        // add @ to inhibit possible warning due to race condition
[...]

PR #13095, opened 2016-11-29
Update Session.php

only-victor:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | 

If YII_DEBUG is true and cookie identity is set the error has occured after the session timeout: 
PHP Warning – yii\base\ErrorException
session_regenerate_id(): Session object destruction failed. ID: user (path: /tmp/php/sessions)
1. in /srv/http/vic/www/myshop/trunk/dev/src/vendor/yiisoft/yii2/web/Session.php at line 284

If set @session_regenerate_id at line 284 then error will be resolved.

### DIFF
--- a/framework/web/Session.php
+++ b/framework/web/Session.php
@@ -281,7 +281,7 @@ class Session extends Component implements \IteratorAggregate, \ArrayAccess, \Co
             // add @ to inhibit possible warning due to race condition
             // https://github.com/yiisoft/yii2/pull/1812
             if (YII_DEBUG && !headers_sent()) {
-                session_regenerate_id($deleteOldSession);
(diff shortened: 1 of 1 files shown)


Thoulah:
The whole if/else is pretty obsolete now

Thoulah:
According to https://github.com/yiisoft/yii2/issues/10583 this change was made on purpose. After all, if is debug modus

dynasource:
indeed. Please discuss this subject over there https://github.com/yiisoft/yii2/issues/10583

Issue #10896, opened 2016-02-18
yii\validators\UniqueValidator select limited columns


dhoomm:
The current implementation of `yii\validators\UniqueValidator` selects all columns.
I have 3 unique attributes and a table with many columns. So it affects preformance.
At the moment I am using as:
`
[['email'], 'unique', 'filter' => function($query) {
            $query->select(static::primaryKey());
        }]
`
but this becomes verbose and maybe interfere with internal details...

PR #13098, opened 2016-11-30
Select only primary keys in UniqueValidator (#10896)

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #10896

Select only primary key when counting records.
Selected results are used only in these lines
```php 
$n = count($models);
...
reset($models)->getPrimaryKey()
```

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -29,6 +29,7 @@ Yii Framework 2 Change Log
 - Enh #7333: Improved error message for `yii\di\Instance::ensure()` when a component does not exist (cebe)
 - Enh #7420: Attributes for prompt generated with `renderSelectOptions` of `\yii\helpers\Html` helper (arogachev)
 - Enh #9162: Added support of closures in `value` for attributes in `yii\widgets\DetailView` (arogachev)
+- Enh #10896: Select only primary key when counting records in UniqueValidator (developeruz)
 - Enh #11037: `yii.js` and `yii.validation.js` use `Regexp.test()` instead of `String.match()` (arogachev, nkovacs)
 - Enh #11756: Added type mapping for `varbinary` data type in MySQL DBMS (silverfire)
 - Enh #11929: Changed `type` column type from `int` to `smallInt` in RBAC migrations (silverfire)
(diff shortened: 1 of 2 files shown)


samdark (changes requested):
Looks good to me. Need changelog for 2.0.11.

samdark:
@dynasource it's already covered by existing tests.

dynasource:
if that was the case, this PR wouldnt be necessary?

samdark:
Yes. It is about performance optimization only. It is not changing behavior and not fixing any issues.

dynasource:
My fault. Had to expand the code to see it.

developeruz:
Changelog was added. 
Btw, a current version of the validateAttribute function is too long to be fully covered with unit test. May I refactor it and push as other PR with additional tests? Or should I do it in this PR, because of the code climate?

samdark:
Separate PR would be much better. This one is good as is.

framework/CHANGELOG.md
```diff
@@ -49,7 +49,7 @@ Yii Framework 2 Change Log
 - Enh #12988: Changed `textarea` method within the `yii\helpers\BaseHtml` class to allow users to control whether html entities found within `$value` will be double-encoded or not (cyphix333)
 - Enh #13074: Improved `\yii\log\SyslogTarget` with `$options` to be able to change the default `openlog` options. (timbeks)
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
-
+- Enh: #10896 Select only primary key when counting records in UniqueValidator
```
samdark: Usually we write github handles right after i.e.:

```
Enh #10896: Select only primary key when counting records in `yii\validators\UniqueValidator` (developeruz)
```

samdark (changes requested):

framework/CHANGELOG.md
cebe: your name is missing at the end of the line :) Also it would be good to order items in changelog by issue number. This avoids merge problems.

cebe (approved):

framework/CHANGELOG.md
cebe: somehow this github review thing is buggy, did not see alex comment before writing this :-/

samdark (approved):

dynasource (approved):

cebe:
Thank you!

samdark:
Had to revert it: https://github.com/yiisoft/yii2/commit/bede9feba5b39169381d7af3c48ae89753fbe9e5

PR #13063, opened 2016-11-23
Fix for #13050 `HostControl` filter

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13050

PR #13099, opened 2016-11-30
[POC] Add host validation to `yii\web\Request`

### LINKED ISSUE #13050: host header attack problem
Hi,
I cache some menu links in my application (datacaching using filecache), sometimes hostname part of these links change to a random website (usually chinese websites) and I have to flush the cache to make it work again.
after some research and studying webserver logs it turned out to be "Host Header Attack". turning off caching would solve the problem but that's not the solution for me as I have very high traffic websites. reading some articles ( https://www.acunetix.com/vulnerabilities/web/host-header-attack ) suggest not using `$_SERVER['HTTP_HOST']` when creating links and using `$_SERVER['SERVER_NAME']` instead. I tried to simulate the situation using burp suite application and the suggested solution acutally works. but the code responsible for getting host info is located in core framework and is in `\yii\web\Request::getHostInfo` method and the above change should be applied on that method.
although I don't know if it has any other side effect rather than ruined links when caching is enabled.

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13050

Closes #13050

Alternative to #13063 with approach more similar to CSRF validation. It still fails by default if host is invalid, but there are no additional filters -  allowed hosts are set at `Request` level which makes configuration easier and (IMO) more logical. Also validation is performed in `Request::getHostInfo()` so there is no risk that we will have corrupted host in `Request::$hostInfo` - at worst case it will be empty (`false` to be exact).

It still is only POC and mostly copy-paste from #13063. If there will be interest of this approach, I will update phpdoc and add more test cases.

### DIFF
--- a/framework/web/Controller.php
+++ b/framework/web/Controller.php
@@ -27,6 +27,20 @@ class Controller extends \yii\base\Controller
      */
     public $enableCsrfValidation = true;
     /**
+     * @var bool whether to enable host info validation for the actions in this controller, to protect against
+     * ['host header' attacks](https://www.acunetix.com/vulnerabilities/web/host-header-attack).
+     * Host info validation is enabled only when this property is true and [[\yii\web\Request::$allowedHosts]] is not empty.
+     */
+    public $enableHostInfoValidation = true;
+    /**
+     * @var callable|false a callback that will be called if the current host does not match [[\yii\web\Request::$allowedHosts]].
+     * If not set, [[\yii\web\Request::$invalidHostCallback]] will be called.
+     *
+     * > Note: while implementing your own host deny processing, make sure you avoid usage of the current requested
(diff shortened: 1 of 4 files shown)


cebe:
Thank you for this proposal! This approach however has several drawbacks:

- I do not see why this should be configurable on a controller level, you either allow your application to deal with unlimited host names or not.
- Adding this to the Request class will make this class even more complex. It already contains a lot of methods with different purpose. The deny method including validation etc is a lot of code added for a feature that should not be needed on a well configured server. The decision made here is not an easy one. We had a long discussion about this topic and decided that the separation into a behavior is the better solution for it.

Issue #13102, opened 2016-12-01
Fixture load yields "permission denied" error when disabling triggers in PostgreSQL


ricpelo:
### What steps will reproduce the problem?

When we try `./yii fixture/load`, Yii disables triggers before load and enables them after. However, in PostgreSQL, the database user must be superuser so he can disable system triggers like foreign key constraints. If the user is not a superuser, we get an error message like this:

```
ERROR:  permission denied: "RI_ConstraintTrigger_16509" is a system trigger
```

### What is the expected result?

`./yii fixture/load` should works with any database user, but currently fails when the user is not a superuser.

### What do you get instead?

Inability to load fixtures with a non-super user. I think that needing a DB superuser only to load fixtures is too overkill.

### Additional info

According to [1], the best workaround is marking the triggers as deferred as first and mark them as immediate at last, using SET CONSTRAINTS ALL DEFERRED/IMMEDIATE. I've done some tests and seems to work OK, but I would like to see some other comments about that. I can create a pull request if needed.

Interestingly, this change will render PR #11775 unnecessary.

[1] http://kopongo.com/2008/7/25/postgres-ri_constrainttrigger-error

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 7.0.8-0ubuntu0.16.04.3
| Operating system | Ubuntu 16.04.1 LTS 64-bit

samdark:
Sounds good. If you'll do a pull request, start with unit tests.

Dador:
Problem still exists. It doesn't allows to run tests with PostgreSQL without using PostgreSQL superuser (which doesn't sounds very secure).

Suggested solution was rejected, see #13106 (I understand why, however this solution works for tests purpose).

Is there anything we can do about this? (besides using PostgreSQL superuser)

samdark:
I'm not into this part of PostgreSQL much so I'm not sure #13106 is the only way to achieve it... do you have alternative solutions?

Dador:
There is [no way](https://www.postgresql.org/message-id/CALnrrJRmFSdee1p8Km8-2cCucSVvk8kzfEMVNBmgyHVuhCusdg@mail.gmail.com) to exactly "disable constraints checks, insert invalid data, enable constraints checks" without superuser rights. So solutions depends on the goal.

[...]

PR #13106, opened 2016-12-01
Switch from ENABLE/DISABLE to DEFERRED/IMMEDIATE triggers in PostgreSQL

ricpelo:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13102 

Please note that this PR will also render PR #11775 unnecessary.

### DIFF
--- a/framework/db/pgsql/QueryBuilder.php
+++ b/framework/db/pgsql/QueryBuilder.php
@@ -187,20 +187,17 @@ class QueryBuilder extends \yii\db\QueryBuilder
      */
     public function checkIntegrity($check = true, $schema = '', $table = '')
     {
-        if ($table === '') {
+        if (empty($table)) {
             $enable = $check ? 'IMMEDIATE' : 'DEFERRED';
             $command = "SET CONSTRAINTS ALL $enable; ";
         } else {
             $enable = $check ? 'ENABLE' : 'DISABLE';
             $schema = $schema ? $schema : $this->db->getSchema()->defaultSchema;
-            $tableNames = $table ? [$table] : $this->db->getSchema()->getTableNames($schema);
             $viewNames = $this->db->getSchema()->getViewNames($schema);
-            $tableNames = array_diff($tableNames, $viewNames);
             $command = '';
-
-            foreach ($tableNames as $tableName) {
+            if (!in_array($table, $viewNames)) {
                 $tableName = '"' . $schema . '"."' . $tableName . '"';
-                $command .= "ALTER TABLE $tableName $enable TRIGGER ALL; ";
+                $command = "ALTER TABLE $tableName $enable TRIGGER ALL; ";
             }
         }
 
--- a/framework/db/pgsql/QueryBuilder.php
+++ b/framework/db/pgsql/QueryBuilder.php
@@ -196,7 +196,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
             $viewNames = $this->db->getSchema()->getViewNames($schema);
             $command = '';
             if (!in_array($table, $viewNames)) {
-                $tableName = '"' . $schema . '"."' . $tableName . '"';
+                $tableName = '"' . $schema . '"."' . $table . '"';
                 $command = "ALTER TABLE $tableName $enable TRIGGER ALL; ";
             }
         }
--- a/framework/db/pgsql/QueryBuilder.php
+++ b/framework/db/pgsql/QueryBuilder.php
@@ -187,17 +187,8 @@ class QueryBuilder extends \yii\db\QueryBuilder
      */
     public function checkIntegrity($check = true, $schema = '', $table = '')
     {
-        $enable = $check ? 'ENABLE' : 'DISABLE';
-        $schema = $schema ? $schema : $this->db->getSchema()->defaultSchema;
-        $tableNames = $table ? [$table] : $this->db->getSchema()->getTableNames($schema);
-        $viewNames = $this->db->getSchema()->getViewNames($schema);
-        $tableNames = array_diff($tableNames, $viewNames);
-        $command = '';
-
-        foreach ($tableNames as $tableName) {
-            $tableName = '"' . $schema . '"."' . $tableName . '"';
-            $command .= "ALTER TABLE $tableName $enable TRIGGER ALL; ";
-        }
+        $enable = $check ? 'IMMEDIATE' : 'DEFERRED';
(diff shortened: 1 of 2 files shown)


framework/db/pgsql/QueryBuilder.php
```diff
@@ -187,17 +187,8 @@ public function resetSequence($tableName, $value = null)
...
-        $viewNames = $this->db->getSchema()->getViewNames($schema);
-        $tableNames = array_diff($tableNames, $viewNames);
-        $command = '';
-
-        foreach ($tableNames as $tableName) {
-            $tableName = '"' . $schema . '"."' . $tableName . '"';
-            $command .= "ALTER TABLE $tableName $enable TRIGGER ALL; ";
-        }
+        $enable = $check ? 'IMMEDIATE' : 'DEFERRED';
+        $command = "SET CONSTRAINTS ALL $enable; ";
```
klimov-paul: This does not looks to be correct: it can not be table name is not affected by SQL anyhow.

klimov-paul (changes requested):

framework/db/pgsql/QueryBuilder.php
klimov-paul: This produces unexpcted behavior, since triggers are disabled for all tables, while they should be only for the particular one.
We can not accept such change in the core.

ricpelo: I understand that. The point is that a FK constraint is implemented using system triggers which only a superuser can enable/disable. Is seems there's no perfect solution for this.

klimov-paul: This code should be used as fallback only if `$table` argument is empty. Otherwise old implementation should be used.

klimov-paul (changes requested):

framework/db/pgsql/QueryBuilder.php
ricpelo: That sounds good. If you wish, I could modify the PR in that way.

framework/db/pgsql/QueryBuilder.php
```diff
@@ -187,8 +187,22 @@ public function resetSequence($tableName, $value = null)
...
     {
-        $enable = $check ? 'IMMEDIATE' : 'DEFERRED';
-        $command = "SET CONSTRAINTS ALL $enable; ";
+        if ($table === '') {
+            $enable = $check ? 'IMMEDIATE' : 'DEFERRED';
+            $command = "SET CONSTRAINTS ALL $enable; ";
+        } else {
+            $enable = $check ? 'ENABLE' : 'DISABLE';
+            $schema = $schema ? $schema : $this->db->getSchema()->defaultSchema;
+            $tableNames = $table ? [$table] : $this->db->getSchema()->getTableNames($schema);
```
klimov-paul: This line should be reworked now as `$table` can not be empty here now.

framework/db/pgsql/QueryBuilder.php
```diff
@@ -187,8 +187,22 @@ public function resetSequence($tableName, $value = null)
      */
     public function checkIntegrity($check = true, $schema = '', $table = '')
     {
-        $enable = $check ? 'IMMEDIATE' : 'DEFERRED';
-        $command = "SET CONSTRAINTS ALL $enable; ";
+        if ($table === '') {
```
klimov-paul: Better to be `empty($table)`

klimov-paul (changes requested):

framework/db/pgsql/QueryBuilder.php
ricpelo: Nice catch.

framework/db/pgsql/QueryBuilder.php
```diff
@@ -187,16 +187,18 @@ public function resetSequence($tableName, $value = null)
...
-        foreach ($tableNames as $tableName) {
-            $tableName = '"' . $schema . '"."' . $tableName . '"';
-            $command .= "ALTER TABLE $tableName $enable TRIGGER ALL; ";
+        if (empty($table)) {
+            $enable = $check ? 'IMMEDIATE' : 'DEFERRED';
+            $command = "SET CONSTRAINTS ALL $enable; ";
+        } else {
+            $enable = $check ? 'ENABLE' : 'DISABLE';
+            $schema = $schema ? $schema : $this->db->getSchema()->defaultSchema;
+            $viewNames = $this->db->getSchema()->getViewNames($schema);
```
klimov-paul: `$viewNames` now redundandt

klimov-paul (changes requested):

framework/db/pgsql/QueryBuilder.php
ricpelo: I'm using `$viewNames` like previously so the method never tries to disable triggers in a view by accident. Do you think we shouldn't do this test?

klimov-paul:
It seems this fix has no effect.
Foolowing unit test:

```php
public function testCheckIntegrity()
    {
        $db = $this->getConnection();

        $db->createCommand()->checkIntegrity(false)->execute();

        //$sql = 'INSERT INTO {{item}}([[name]], [[category_id]]) VALUES (\'invalid\', 99999)';
        $sql = 'INSERT INTO {{item}}([[name]], [[category_id]]) VALUES (\'invalid\', NULL)';
        $command = $db->createCommand($sql);
        $command->execute();

        $db->createCommand()->checkIntegrity(true)->execute();

        $this->setExpectedException('\yii\db\IntegrityException');
        $command->execute();
    }
```

Fails with message: 

```
 yii\tests\unit\framework\db\pgsql\CommandTest::testCheckIntegrity
yii\db\IntegrityException: SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "category_id" violates not-null constraint
DETAIL:  Failing row contains (6, invalid, null).
The SQL being executed was: INSERT INTO "item"("name", "category_id") VALUES ('invalid', NULL)
```

ricpelo:
That's weird... PostgreSQL documentación says SET CONSTRAINTS ALL command has effect only during a transacción. Could it be the problem? Can you try it inside a transaction?

klimov-paul:
> PostgreSQL documentation says SET CONSTRAINTS ALL command has effect only during a transaction

In this case this fix is unacceptable.

PR #13110, opened 2016-12-01
Custom ID generation for Pjax widgets to solve unstable ajax calls (i…

dynasource:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12969

Implements the solution proposed in the following comment:
https://github.com/yiisoft/yii2/pull/12977#issuecomment-260123078


samdark:
It doesn't solve the issue of conflicting IDs since PJAX contents are reloaded, right?

framework/widgets/Pjax.php
```diff
@@ -97,7 +97,11 @@ class Pjax extends Widget
      * [pjax project page](https://github.com/yiisoft/jquery-pjax) for available options.
      */
     public $clientOptions;
-
+    /**
+     * @var string the prefix to the automatically generated pjax widget IDs.
+     * @see getId()
+     */
+    public static $pjaxAutoIdPrefix = 'w_pjax_';
```
samdark: Why not to override `$autoIdPrefix` instead of creating another property?

samdark: Also I'd use something simple like `p` as prefix.

framework/widgets/Pjax.php
```diff
@@ -133,6 +137,23 @@ public function init()
...
     /**
+     * Custom ID generation for Pjax widgets
+     * This override is to attain a more reliable match between ID's of local and remote Pjax widgets (#12969)
+     * The implementation follows the solution proposed in the following comment:
+     * @link https://github.com/yiisoft/yii2/pull/12977#issuecomment-260123078
+     * @since 2.0.11
+     */
+    private static $pjaxCounter = 0;
+    private $_pjaxId;
+    public function getId($autoGenerate = true)
```
samdark: There's no need to override this method. You just have to declare the following in Pjax class:

```php
 public static $counter = 0;
```

samdark (changes requested):

dynasource:
looks like you are right. The tests failed when reusing the static variables (so I choose for custom ones) but redeclaring seems to do the trick.

samdark (approved):
Other than missing changelog it looks good.

SilverFire:
It's a BC breaking change.

Story:
 - I compose a view with PJAX
 - I know that Pjax is the first widget rendered on this page
 - I attach JS listeners to `#w0`

Story after this PR:
 - `#w0` selector does NOT point on the PJAX widget

However, It's a minor BC break and in case when you bind to something auto-generated, you just prepare a good place for problems in future.

IMO we can accept this PR, but mention possible drawback in UPGRADE.md

SilverFire (changes requested):
Should be mentioned in UPGRADE.md

dynasource:
done

SilverFire (approved):

tests/framework/widgets/PjaxTest.php
```diff
@@ -0,0 +1,35 @@
...
+use yii\data\ArrayDataProvider;
+use yii\widgets\ListView;
+use yii\widgets\Pjax;
+use yiiunit\TestCase;
+
+class PjaxTest extends TestCase
+{
+    public function testGeneratedIdByPjaxWidget()
+    {
+        $nonPjaxWidget1 = new ListView(['dataProvider' => new ArrayDataProvider()]);
```
samdark: Counters should be reset here else test would fail when ran together with other tests. See latest travis build.

samdark (changes requested):

framework/widgets/Pjax.php
```diff
@@ -99,13 +99,13 @@ class Pjax extends Widget
     public $clientOptions;
     /**
      * @inheritdoc
-     * Note: This variable is redeclared to force separate counting for Pjax widgets (issue #12969)
-     * @since 2.0.11
+     * @internal
+     * @since 2.0.11 (#12969)
```
samdark: No need for `@since` tag here. API isn't new and it's not actually `@internal`. It could be set by end user and that's what we're doing in tests.

framework/widgets/Pjax.php
```diff
@@ -99,13 +99,13 @@ class Pjax extends Widget
...
      */
     public static $counter = 0;
     /**
      * @inheritdoc
-     * @since 2.0.11
+     * @since 2.0.11 (#12969)
```
samdark: No need for `@since` here because API isn't new.

samdark (changes requested):

framework/widgets/Pjax.php
dynasource: @since: doesnt it matter that the counter is reset to 0 for Pjax Widgets?
@internal is copied from the Pjax.php to be consistent. IMO this variable should be private in the first place. Why should automatic counters be altered anyway?

samdark: OK, I'm wrong about `@internal`. Seems to be where it should be.

samdark: > doesnt it matter that the counter is reset to 0 for Pjax Widgets?

It matters but it's not API change and `@since` (poor guy, he's mentioned all the time...) is for public API changes and in this case there aren't any.

samdark (approved):

cebe (approved):

cebe (approved):
I have updated the UPGRADE note to be more clear about what is going on here and how to fix it.

samdark:
Merged. Thanks everyone for working on it.

dynasource:
yw. Credits go to @rob006.

PR #13113, opened 2016-12-01
Code cleanup

samdark:
Mainly to test Codeclimate but it makes code a bit more readable as well.

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -151,6 +151,7 @@ class BaseYii
         if ($throwException) {
             throw new InvalidParamException("Invalid path alias: $alias");
         }
+
         return false;
     }
 
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -139,20 +139,19 @@ class BaseYii
         if (isset(static::$aliases[$root])) {
             if (is_string(static::$aliases[$root])) {
                 return $pos === false ? static::$aliases[$root] : static::$aliases[$root] . substr($alias, $pos);
-            } else {
-                foreach (static::$aliases[$root] as $name => $path) {
-                    if (strpos($alias . '/', $name . '/') === 0) {
-                        return $path . substr($alias, strlen($name));
-                    }
+            }
+
+            foreach (static::$aliases[$root] as $name => $path) {
+                if (strpos($alias . '/', $name . '/') === 0) {
+                    return $path . substr($alias, strlen($name));
                 }
             }
         }
 
         if ($throwException) {
             throw new InvalidParamException("Invalid path alias: $alias");
-        } else {
-            return false;
         }
+        return false;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


cebe (approved):
For easier review: https://github.com/yiisoft/yii2/pull/13113/files?w=0 (ignoring whitespace changes)

framework/base/Component.php
```diff
@@ -503,19 +507,19 @@ public function off($name, $handler = null)
         if ($handler === null) {
             unset($this->_events[$name]);
             return true;
-        } else {
-            $removed = false;
-            foreach ($this->_events[$name] as $i => $event) {
-                if ($event[0] === $handler) {
-                    unset($this->_events[$name][$i]);
-                    $removed = true;
-                }
```
SilverFire: This change looks wrong. `return` statement is missing in this condition

framework/base/Component.php
```diff
@@ -503,19 +507,19 @@ public function off($name, $handler = null)
         if ($handler === null) {
```
SilverFire: We leave empty line before return statement or not?

framework/BaseYii.php
```diff
@@ -139,20 +139,19 @@ public static function getAlias($alias, $throwException = true)
...
+            foreach (static::$aliases[$root] as $name => $path) {
+                if (strpos($alias . '/', $name . '/') === 0) {
+                    return $path . substr($alias, strlen($name));
                 }
             }
         }
 
         if ($throwException) {
             throw new InvalidParamException("Invalid path alias: $alias");
-        } else {
```
SilverFire: We leave empty line before return statement or not?

framework/BaseYii.php
```diff
@@ -139,20 +139,19 @@ public static function getAlias($alias, $throwException = true)
...
-            return false;
         }
+        return false;
```
SilverFire: We leave empty line before return statement or not?

framework/base/Controller.php
```diff
@@ -186,9 +186,8 @@ public function run($route, $params = [])
             return $this->runAction($route, $params);
         } elseif ($pos > 0) {
             return $this->module->runAction($route, $params);
-        } else {
-            return Yii::$app->runAction(ltrim($route, '/'), $params);
         }
+        return Yii::$app->runAction(ltrim($route, '/'), $params);
```
SilverFire: We leave empty line before return statement or not?

framework/base/Event.php
```diff
@@ -116,20 +116,20 @@ public static function off($class, $name, $handler = null)
...
+            if ($event[0] === $handler) {
+                unset(self::$_events[$name][$class][$i]);
+                $removed = true;
+            }
         }
+        if ($removed) {
+            self::$_events[$name][$class] = array_values(self::$_events[$name][$class]);
+        }
+
+        return $removed;
```
SilverFire: We leave empty line before return statement or not?

SilverFire (changes requested):

framework/BaseYii.php
samdark: We don't have agreement on that as far as I remember. In these cases it depends on if it's a separate logical return block or not.

framework/base/Event.php
dynasource: rather not. Don't see the use of it as there is already an empty line because of the bracket

SilverFire (approved):

framework/BaseYii.php
cebe: we did not do that but at the point where PSR2 was changing from tabs to spaces, these lines were introduced for some reason. In most cases I do not think adding a blank line makes sense.

dynasource (approved):
the changes look OK. The line breaks are inconsistent though.

PR #13121, opened 2016-12-02
Disable `Controversial/CamelCasePropertyName` codeclimate check

rob006:
This rule is conflicting with properties prefixed by `_`.  See https://codeclimate.com/github/yiisoft/yii2/pull/13117 .

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -14,7 +14,9 @@ engines:
     enabled: true
     checks:
         CleanCode/StaticAccess:
-            enabled: false
+          enabled: false
+        Controversial/CamelCasePropertyName:
+          enabled: false  
 ratings:
   paths:
   - "**.js"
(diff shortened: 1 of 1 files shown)


samdark (approved):

cebe:
while we are at naming things:

> Avoid using short method names like m150909_153426_cache_init::up(). The configured minimum method name length is 3.

also a useless check imo. https://codeclimate.com/github/yiisoft/yii2/framework/caching/migrations/m150909_153426_cache_init.php

cebe:
Thanks!

cebe:
Will add more myself.

cebe:
@rob006 turns out it is possible to configure it to allow underscore in private properties: https://phpmd.org/rules/controversial.html#camelcasepropertyname :)

PR #12145, opened 2016-08-12
Revamps PageCache to store data in one entry (#10506)

sergeymakinen:
| Q | A |
| --- | --- |
| Is bugfix? | yes |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #10506 |

This updated `PageCache` combines together `PageCache` and `FragmentCache` to provide caching in one entry.
### Features
- fully backwards compatible (incl. dynamic content)
- easy to be extended
- able to store/retrieve an additional data in `beforeCacheResponse` and `afterRestoreResponse` (_for example: I store Last-Modified in data and then retrieve it later in `HttpCache`_)
- all data is stored in one cache entry
- able to cancel caching by returning `false` in `beforeCacheResponse`

PR #13124, opened 2016-12-02
[WIP] Added test for PageCache filter

cebe:
started creating a test for PageCache filter. This should be completed before #12145 is merged to reduce risk of unspotted breaks.

If someone wants to complete it, feel free to do so.

### DIFF
--- a/tests/framework/filters/PageCacheTest.php
+++ b/tests/framework/filters/PageCacheTest.php
@@ -1,0 +1,92 @@
+<?php
+
+namespace yiiunit\framework\filters;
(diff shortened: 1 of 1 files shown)


sergeymakinen:
Good job! I have planed tests exactly this way:
- write tests for the current implementation
- than test/adopt them against #12145

I will try to complete tests this weekend.

sergeymakinen:
So I have finished the tests: #13149

samdark:
Closed in favor of #13149

PR #13127, opened 2016-12-03
Fix docs about table prefixes

### LINKED ISSUE #8354: Improve table prefix documentation
The examples in [the ActiveRecord guide](http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#declaring-active-record-classes) and [the ActiveRecord class documentation](http://www.yiiframework.com/doc-2.0/yii-db-activerecord.html) don't use `{{% }}`.
[The documentation for tableName](http://www.yiiframework.com/doc-2.0/yii-db-activerecord.html#tableName%28%29-detail) doesn't mention it. In fact it implies that the default implementation adds the table prefix directly, when actually it just adds the `{{% }}`.
[The documentation for \yii\db\Query](http://www.yiiframework.com/doc-2.0/yii-db-query.html#from%28%29-detail) and the example there also completely omits it.
Only [the documentation for \yii\db\Connection::tablePrefix](http://www.yiiframework.com/doc-2.0/yii-db-connection.html#$tablePrefix-detail) mentions it, but it's not really clear where you can use `{{% }}`, since it just vaguely refers to "a table name". I knew that I can use `{{% }}` in `ActiveRecord::tableName()`, but only because gii has an option to generate it like that, and I was not aware that I could use it in `Query::from()`.
Wouldn't it be better to use `{{% }}` in all these examples?

dynasource:
fixes #8354 by adding docs about prefixes to the mentioned areas

### DIFF
--- a/docs/guide/db-active-record.md
+++ b/docs/guide/db-active-record.md
@@ -53,15 +53,17 @@ However, most content described here are also applicable to Active Record for No
 To get started, declare an Active Record class by extending [[yii\db\ActiveRecord]]. 
 
 ### Setting a table name
+
 By default each Active Record class is associated with its database table.
-The [[yii\db\ActiveRecord::tableName()|tableName()]] method returns the class name as the table name by calling [[yii\helpers\Inflector::camel2id()]].
+The [[yii\db\ActiveRecord::tableName()|tableName()]] method returns the table name by converting the class name via [[yii\helpers\Inflector::camel2id()]].
 You may override this method if the table is not named after this convention.
 
 Also a default [[yii\db\Connection::$tablePrefix|tablePrefix]] can be applied. For example if
  [[yii\db\Connection::$tablePrefix|tablePrefix]] is `tbl_`, `Customer` becomes `tbl_customer` and `OrderItem` becomes `tbl_order_item`. 
 
-If a table name is given as `{{%TableName}}`, then the percentage character `%` will be replaced with this property value. 
-For example, `{{%post}}` becomes `{{tbl_post}}`.
+If a table name is given as `{{%TableName}}`, then the percentage character `%` will be replaced with the table prefix. 
+For example, `{{%post}}` becomes `{{tbl_post}}`. The brackets around the table name are used for
+[quoting in an SQL query](db-dao.md#quoting-table-and-column-names).
 
 In the following example, we declare an Active Record class named `Customer` for the `customer` database table.
 
@@ -80,7 +82,7 @@ class Customer extends ActiveRecord
      */
     public static function tableName()
     {
-        return 'customer';
+        return '{{customer}}';
     }
 }
 ```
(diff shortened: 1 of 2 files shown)


samdark (approved):

docs/guide/db-active-record.md
```diff
@@ -50,9 +50,18 @@ However, most content described here are also applicable to Active Record for No
 
 ## Declaring Active Record Classes <span id="declaring-ar-classes"></span>
 
-To get started, declare an Active Record class by extending [[yii\db\ActiveRecord]]. Because each Active Record
-class is associated with a database table, in this class you should override the [[yii\db\ActiveRecord::tableName()|tableName()]]
-method to specify which table the class is associated with.
+To get started, declare an Active Record class by extending [[yii\db\ActiveRecord]]. 
+
+### Setting a table name
+By default each Active Record class is associated with it's database table,
```
SilverFire: `it's` -> `its`

docs/guide/db-active-record.md
```diff
@@ -50,9 +50,18 @@ However, most content described here are also applicable to Active Record for No
...
 ## Declaring Active Record Classes <span id="declaring-ar-classes"></span>
 
-To get started, declare an Active Record class by extending [[yii\db\ActiveRecord]]. Because each Active Record
-class is associated with a database table, in this class you should override the [[yii\db\ActiveRecord::tableName()|tableName()]]
-method to specify which table the class is associated with.
+To get started, declare an Active Record class by extending [[yii\db\ActiveRecord]]. 
+
+### Setting a table name
+By default each Active Record class is associated with it's database table, 
+[[yii\db\ActiveRecord::tableName()|tableName()]] returns the class name as the table name by calling [[Inflector::camel2id()]].
```
SilverFire: `[[yii\db\ActiveRecord::tableName()|tableName()]] returns` -> `[[yii\db\ActiveRecord::tableName()|tableName()]] method returns`

SilverFire: Namespace for `[[Inflector]]` is missing

SilverFire (changes requested):

SilverFire (approved):

docs/guide/db-query-builder.md
```diff
@@ -153,6 +153,12 @@ $subQuery = (new Query())->select('id')->from('user')->where('status=1');
 $query->from(['u' => $subQuery]);
 ```
 
+#### Prefixes
+Also a default [[yii\db\Connection::$tablePrefix|tablePrefix]] can be applied. For example if
+ [[yii\db\Connection::$tablePrefix|tablePrefix]] is `tbl_`, `Customer` becomes `tbl_customer` and `OrderItem` becomes `tbl_order_item`.
+ 
+If a table name is given as `{{%TableName}}`, then the percentage character `%` will be replaced with this property value. 
+For example, `{{%post}}` becomes `{{tbl_post}}`.
```
cebe: This is already covered in `db-dao.md`. http://www.yiiframework.com/doc-2.0/guide-db-dao.html#quoting-table-and-column-names

cebe:
I have added a commit with some improvements. Also table prefix in query guide seems duplicate, it should link to http://www.yiiframework.com/doc-2.0/guide-db-dao.html#quoting-table-and-column-names instead.

cebe (approved):

Issue #13128, opened 2016-12-03
ColumnSchemaBuilder: {pos} placeholder should always be last


df2:
### What steps will reproduce the problem?

Noticed this by trying to move a `primaryKey` column. It is not possible (at least in MySql), as the `PRIMARY KEY` gets defined twice, resulting in an error.

Workaround would be to use syntax like:

```php
/**
 * This doesn't work
 * $this->alterColumn('bb', 'id', $this->primaryKey()->first());
 */

$this->alterColumn('bb', 'id', $this->integer()->append('AUTO_INCREMENT')->first());
``` 

### What is the expected result?

`ALTER TABLE 'bb' CHANGE 'id' 'id' int(11) AUTO_INCREMENT FIRST`

### What do you get instead?

`ALTER TABLE 'bb' CHANGE 'id' 'id' int(11) FIRST AUTO_INCREMENT`

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 7.0.13
| Operating system | Ubuntu

df2:
MySql docs: http://dev.mysql.com/doc/refman/5.7/en/alter-table.html

```
| CHANGE [COLUMN] old_col_name new_col_name column_definition
      [FIRST|AFTER col_name]
```

PR #13129, opened 2016-12-03
#13128: Fixed incorrect position of {pos} string in ColumnSchemaBuilder __toString

usecomposable:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | unknown
| Fixed issues  | #13128


df2:
Can this please be merged into master?

dynasource:
if the append string is not inserted at last, its not really an append string anymore right ;)

cebe:
Yeah but still it makes sense to change it because DBMS syntax is this way. It may be useful to point that out in the docs though.

dynasource:
does this "append" typology match any mysql typology or is it chosen by us. Tried to find a clarifying reference.

df2:
Basically `append` is used to append to the column definition. The position indicator should always be last in the SQL statement.

MySql: http://dev.mysql.com/doc/refman/5.7/en/alter-table.html

```
| CHANGE [COLUMN] old_col_name new_col_name column_definition
      [FIRST|AFTER col_name]
```

Cubrid: http://www.cubrid.org/manual/91/en/sql/schema/table.html#change-modify-clauses

```
    CHANGE [COLUMN | CLASS ATTRIBUTE ] old_col_name new_col_name column_definition
             [FIRST | AFTER col_name]
  | MODIFY [COLUMN | CLASS ATTRIBUTE] col_name column_definition
             [FIRST | AFTER col_name]
```

Oracle: It seems it was a mistake to allow the position option. I have removed it from the oci `__toString`.

df2:
Not sure why these tests are failing... help much appreciated.

samdark:
Tests are not failing, it's codeclimate, not travis.

tests/framework/db/QueryBuilderTest.php
```diff
@@ -928,6 +928,61 @@ public function columnTypes()
                     'cubrid' => "int NOT NULL AUTO_INCREMENT PRIMARY KEY COMMENT 'test comment'",
                 ],
             ],
+            [
+                Schema::TYPE_PK . " FIRST",
```
samdark: Should it be special for Oracle which doesn't support it?

samdark (changes requested):

tests/framework/db/QueryBuilderTest.php
df2: Hmm, I think then with Oracle, adding a position modifier will have no effect, and thus the tests are correct. It is the same as Postgres in that respect.

samdark: OK.

samdark (approved):

SilverFire (approved):

cebe (approved):

samdark:
Merged. Thank you!

PR #13139, opened 2016-12-05
Code Climate PHPMD settings

bizley:
I have recently started using Code Climate for my project as well and I have noticed that PHPMD settings taken from Yii 2 file are not met. You can verify this by looking at [Code Climate Yii 2 analysis](https://codeclimate.com/github/yiisoft/yii2).

Based on the CC and PHPMD docs I've prepared separate PHPMD ruleset that appears to work.  
This set contains all rules previously listed in `checks` and in addition it excludes `naming.xml\BooleanGetMethodName` which states

> Method which returns a boolean should be named 'is...()' or 'has...()'

since it conflicts with Yii 2 component's getters.


SilverFire (approved):

SilverFire:
Thank you, @bizley 
The changes look good to me.

cebe:
Great, thanks! Could we move that file sowhere into `tests` folder to not clutter the root more than necessary? Maybe `tests/data/codeclimate`?

bizley:
Docs state the path is relative so it should work. Would like me to move it?

cebe:
if you have the time, do it :)

bizley:
Moved, looks like CC is ok with it.

cebe:
> codeclimate — Code Climate found 243 fixed issues.

sounds great :-D

cebe:
Thank you!

Issue #5846, opened 2014-11-01
Rename the rbac migration to a normal class name and for including it in a project


mdmunir:
I have several migration from several extension. How to `migrate/down` just one migration without change/apply other migration. Here is my migration history

```
    (2014-10-28 04:14:30) m141018_105939_create_table_upload
    (2014-10-28 04:14:30) m140624_050135_create_table_accounting
    (2014-10-28 04:14:30) m140624_050114_create_table_sales
    (2014-10-28 04:14:30) m140624_050056_create_table_inventory
    (2014-10-28 04:14:30) m140624_050030_create_table_purchase
    (2014-10-28 04:12:13) m140622_165356_create_table_master
    (2014-09-29 03:11:52) m140602_111327_create_menu_table
    (2014-09-29 03:11:52) m140527_084418_auto_number
    (2014-09-29 03:11:52) m140506_102106_rbac_init
    (2014-09-29 03:11:52) m130524_201442_init
```

How do `migrate/down` or `migrate/redo` only `auto_number` migration.

cebe:
related #384

Faryshta:
#8202 could help

dynasource:
closed in favor of #8202

cebe:
> Thinking about something like renaming the rbac migration to a normal class name and make including it in a project by extending it in the concrete migration.

this is something I'd like to do.

SilverFire:
https://github.com/yiisoft/yii2/issues/13918

Will be namespaced

PR #13141, opened 2016-12-05
Add namespace for RBAC migration

rapita:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    |no
| Tests pass?   | no

### DIFF
--- a/framework/rbac/migrations/m140506_102106_rbac_init.php
+++ b/framework/rbac/migrations/m140506_102106_rbac_init.php
@@ -5,6 +5,8 @@
  * @license http://www.yiiframework.com/license/
  */
 
+namespace yii\rbac\migrations;
+
 use yii\base\InvalidConfigException;
 use yii\rbac\DbManager;
 


cebe:
can you explain the change?

rapita:
This makes it possible to configure MigrateController.
It is much easier to automatically deploy project(using db rbac) in production server.

Example usage:
```
    'controllerMap' => [
        ...
        'migrate' => [
            'class' => 'yii\console\controllers\MigrateController',
            'migrationNamespaces' => [
                'yii\rbac\migrations',
            ],
        ],
        ...
    ],
```

cebe:
this conflicts with another solution to the problem, which is better imo because it will allow integrating the migration with a timestamp that fits application migration order: https://github.com/yiisoft/yii2/issues/5846#issuecomment-258177811 Will check it out later.

resurtm:
@rapita 

> It is much easier to automatically deploy project(using db rbac) in production server.

One line less in your project deploy script would make your life easier? ;-)

klimov-paul:
This is an existing class renaming, e.g. `m140506_102106_rbac_init` to `app\migrations\m140506_102106_rbac_init`. Thus it can not be applied before 2.1

Issue #11122, opened 2016-03-15
count() on Query with orderBy()


zuozp8:
I have relation defined like

``` php
class Company {
    public function getAddresses()
    {
        return $this->hasMany(Address::class, ['company_id' => 'id'])->orderBy('address_id');
    }
}
```

as i always want addresses of companies ordered in the same way

Unfortunately I can't do `$company->getAddresses()->count()` because psql is complaining `[42803] ERROR: column "address.address_id" must appear in the GROUP BY clause or be used in an aggregate function`

Check for orderBy presence could be added here https://github.com/yiisoft/yii2/blob/master/framework/db/Query.php#L392

Alex-Code:
Do you have the debug toolbar enabled?
Could you show what SQL gets generated?

zuozp8:
sure, its:
```sql
SELECT COUNT(*) FROM "address" WHERE "company_id"='17e49540-d3e7-11e4-b3ef-53b449010d08' ORDER BY "address_id";
```

SilverFire:
You can not use order by with `count()`.

```php
$company->getAddresses()->orderBy([])->count();
```

zuozp8:
@SilverFire i thought it should handled like #8068

SilverFire:
Hmm... Yes.

herroffizier:
It seems that check for non-empty ```offset``` and ```limit``` must be also added here https://github.com/yiisoft/yii2/blob/master/framework/db/Query.php#L392. I think that ```count()``` should return actual count of records that may be returned by ```all()``` but as for now it is not working that way.

For example, we expecting to get 200 or less records from this query:
```php
$query = (new \yii\db\Query())->from('offer')->offset(0)->limit(200);
```
But following code will fail when table have more than 200 records in it, because ```$query->count()``` will return total count of records in table.
```php
assert(count($query->all()) == $query->count());
```

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/25e6eb04a5e23203cd6e5da8db427afe5a46f5bb

SilverFire:
Thank you for the report!

PR #13144, opened 2016-12-06
Small refactor to queryScalar

alex-code:
The subquery count in `Query::queryScalar` doesn't need all the code that precedes it to run so moved it inside the `if` block.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -1,6 +1,12 @@
 Yii Framework 2 Change Log
 ==========================
 
+2.0.12 under development
+--------------------------
+
+- Enh #13144: Refactor to `queryScalar` (Alex-Code)
+
+
 2.0.11 under development
 ------------------------
 
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.12 under development
 --------------------------
 
-- Enh #13144: Refactor to `queryScalar` (Alex-Code)
+- Enh #13144: Refactored `queryScalar` (Alex-Code)
 
 
 2.0.11 under development
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -330,14 +330,16 @@ class ActiveQuery extends Query implements ActiveQueryInterface
      */
     protected function queryScalar($selectExpression, $db)
     {
-        if ($this->sql === null) {
-            return parent::queryScalar($selectExpression, $db);
-        }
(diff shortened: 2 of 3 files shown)


samdark (approved):

samdark:
Looks like good optimization.

klimov-paul (approved):

SilverFire (approved):
May be accepted to 2.0.11

dynasource (approved):

samdark:
@Alex-Code could you add a line for changelog for 2.0.12?

SilverFire (approved):

SilverFire:
Merged https://github.com/yiisoft/yii2/commit/bfd0de3e42f79d9d401691c1d85061f025b1417f

SilverFire:
Thank you!

doerler:
This commit (based on #11122) changes the behaviour (which temporary unsets `limit `and `offset`) of  `Query::queryScalar` when `orderBy `is not empty. 

If `limit` is set to a value lower then results count and `orderBy` is not empty the `Query::count()` returns limit instead of the expected count.

Shouldn't  `limit `and `offset` be unset in both blocks, like this?:

```
protected function queryScalar($selectExpression, $db)
{
    if ($this->emulateExecution) {
        return null;
    }

    $limit = $this->limit;
    $offset = $this->offset;
    $this->limit = null;
    $this->offset = null;

    $result = null;
    if (
        !$this->distinct
        && empty($this->groupBy)
        && empty($this->having)
        && empty($this->union)
        && empty($this->orderBy)
    ) {
        $select = $this->select;

        $this->select = [$selectExpression];

        $command = $this->createCommand($db);

        $this->select = $select;

        $result = $command->queryScalar();
    } else {
        $result = (new Query)->select([$selectExpression])
            ->from(['c' => $this])
            ->createCommand($db)
            ->queryScalar();
    }
    $this->limit = $limit;
    $this->offset = $offset;
    return $result;
}
```

Background:
MariaDB for example had no problems with `order by' and aggregate functions like count().
So we use the same query for pagination and count in our REST interface.
A workaround on application site is to empty orderBy or limit before executing the count() query,
but this worked until 2.0.11 - so this is a breaking change IMHO.

 
Example of "wrong" raw sql generated by  `Query::count()` :

```
SELECT COUNT(*) FROM (SELECT * FROM `products` WHERE (`name` LIKE :qp0) AND (`class_id`=:qp1) ORDER BY `parent_id`, `sort` LIMIT 100) `c`
```

samdark:
Please report in a separate issue. Thanks.

samdark:
@silverfire

PR #13145, opened 2016-12-06
Make validator getClientOptions public

### LINKED ISSUE #11163: Add export of validation options without the js string
For example, now the clientValidateAttribute function  in  yii\validators\SomeValidator returns a string

``` php
public function clientValidateAttribute($model, $attribute, $view)
{
    $options = [
        'trueValue' => $this->trueValue,
        'falseValue' => $this->falseValue,
        'message' => Yii::$app->getI18n()->format($this->message, [
        'attribute' => $model->getAttributeLabel($attribute),
        'true' => $this->trueValue,
        'false' => $this->falseValue,
        ], Yii::$app->language),
    ];
    ///...
    ValidationAsset::register($view);

    return 'yii.validation.boolean(value, messages, ' . json_encode($options, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE) . ');';
}
```

This function can't be used without yii.validation.
Please, add new function or change this that it returned an array, like this

``` php
public function clientValidateAttribute($model, $attribute, $view)
 {
    $options = [
        'trueValue' => $this->trueValue,
        'falseValue' => $this->falseValue,
        'message' => Yii::$app->getI18n()->format($this->message, [
        'attribute' => $model->getAttributeLabel($attribute),
        'true' => $this->trueValue,
        'false' => $this->falseValue,
        ], Yii::$app->language),
    ];

    ///...

    return $options;
}
```

It is possible to make it so, for example

``` php
public function clientValidateOptions($model, $attribute)
{
    $options = [
        'trueValue' => $this->trueValue,
        'falseValue' => $this->falseValue,
        'message' => Yii::$app->getI18n()->format($this->message, [
        'attribute' => $model->getAttributeLabel($attribute),
        'true' => $this->trueValue,
        'false' => $this->falseValue,
        ], Yii::$app->language),
    ];
    // ... other setting options
    return $options;
}

public function clientValidateAttribute($model, $attribute, $view)
{
    $options = $this->clientValidateOptions($model, $attribute)
    ///...
    ValidationAsset::register($view);

    return 'yii.validation.boolean(value, messages, ' . json_encode($options, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE) . ');';
}
```

nkovacs:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #11163 

This allows implementing custom client-side validation
without extending every validator.

The backwards compatibility break is caused by the change in ImageValidator. `ImageValidator::getClientOptions` was protected since 2.0.0 (https://github.com/yiisoft/yii2/commit/6dd2203a5c0db168f51369cab206ff07e3c1fd04#diff-504fd084646834cd27fce3dbdce26bbeR350), so if you extended the class and overrode the method with a protected method, the visibility change will cause a fatal error (because you can't make a public method protected in a subclass).

Fixes #11163

### DIFF
--- a/docs/guide/input-validation.md
+++ b/docs/guide/input-validation.md
@@ -895,13 +895,13 @@ the former will take precedence.
 > - if you want to implement your own custom client-side validation but leave the synchronization with server-side
 > validator options;
 > - to extend or customize to fit your specific needs:
-> 
+>
 > ```php
-> protected function getClientOptions($model, $attribute)
+> public function getClientOptions($model, $attribute)
 > {
 >     $options = parent::getClientOptions($model, $attribute);
 >     // Modify $options here
-> 
+>
 >     return $options;
 > }
 > ```
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -29,14 +29,14 @@ The simple way to upgrade Yii, for example to version 2.0.10 (replace this with 
     composer require "yiisoft/yii2:~2.0.10"
 
 This however may fail due to changes in the dependencies of yiisoft/yii2, which may change due to security updates
-in other libraries or by adding support for newer versions. `composer require` will not update any other packages 
+in other libraries or by adding support for newer versions. `composer require` will not update any other packages
 as a safety feature.
 
 The better way to upgrade is to change the `composer.json` file to require the new Yii version and then
 run `composer update` by specifying all packages that are allowed to be updated.
 
     composer update yiisoft/yii2 yiisoft/yii2-composer bower-asset/jquery.inputmask
-    
+
 The above command will only update the specified packages and leave the versions of all other dependencies intact.
 This helps to update packages step by step without causing a lot of package version changes that might break in some way.
 If you feel lucky you can of course update everything to the latest version by running `composer update` without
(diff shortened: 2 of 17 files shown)


cebe:
> The backwards compatibility break is caused by the change in ImageValidator.

could you add that note to the UPGRADE guide please?

nkovacs:
Done.

samdark:
Since it's a backwards compatibilty break and isn't critical to have in 2.0.x, I'd merge it into 2.1.0 instead. @cebe ?

SilverFire:
I insist to merge this PR in 2.0.11 and here are the reasons:

 * We must admit that once we release an API, it will be used by other developers. 
 * Method `getClientOptions()` was introduced on the high-level API in version 2.0.11. Before that only `ImageValidator` class used the same method name. 
 * In case we merge this PR to 2.1 milestone UPGRADE will affect much more code in comparison to `ImageValidator` BC break.

nkovacs:
There's also the option of calling it something else (though even then you're introducing a BC break, since if you call it foobar, and someone has a validator class with a protected function foobar, it'll break their code; the likelihood of this is lower, of course).

samdark (approved):

dynasource (approved):

cebe (approved):

samdark:
Need a changelog. Then ready to merge.

cebe:
no need for a changelog, #11163 does already have one.

cebe:
merged, thank you!

PR #13153, opened 2016-12-07
Polish translations for field validation fix

sztyvny:
In Polish "should" is gender depended verb (masculine). It sounds bad with feminine and neuter attribute names. My change means "must" but it keeps sentence sense, and sounds more correct.


bizley:
I agree.

samdark:
Thanks!

PR #13160, opened 2016-12-08
Added JS tests for yii.captcha.js (#12840), fixes #13159

### LINKED ISSUE #13159: destroy method in yii.captcha.js doesn't work as expected
Let's say we want to use captcha plugin on some image:

```html
<img id="captcha" src="/site/captcha/">
```

jQuery plugin initialization:

```js
$('#captcha').yiiCaptcha({
    refreshUrl: '/site/captcha?refresh=1',
    hashKey: 'yiiCaptcha/site/captcha'
});
```

Then we decide to destroy it:

```js
$('#captcha').yiiCaptcha('destroy');
```

With each repeated initialization:

```js
$('#captcha').yiiCaptcha({
    refreshUrl: '/site/captcha?refresh=1',
    hashKey: 'yiiCaptcha/site/captcha'
});
```

click on the captcha will generate 1 more extra AJAX request. This is because click event handler was not removed correctly. Here is the current implementation:

```js
destroy: function () {
    return this.each(function () {
        $(window).unbind('.yiiCaptcha');
        $(this).removeData('yiiCaptcha');
    });
},
```

Why unbind event handlers from `window` and in the loop which is even more weird? We call destroy on specific element(s), so this will be correct:

```js
destroy: function () {
    return this.each(function () {
        var $e = $(this);
        $e.unbind('.yiiCaptcha');
        $e.removeData('yiiCaptcha');
    });
},
```

Also loop is completely useless here because `this` is jQuery object on which `destroy` called from, we can unbind event handlers and remove data without any loops here:

```js
destroy: function () {
    var $e = $(this);
    $e.unbind('.yiiCaptcha');
    $e.removeData('yiiCaptcha');
    return $e;
},
```

Further improvements:

- `this` is already jQuery object, so there is no need to wrap it with `$(...)`.

- From docs to [unbind](http://api.jquery.com/unbind/):

> As of jQuery 3.0, .unbind() has been deprecated. It was superseded by the .off() method since jQuery 1.7, so its use was already discouraged.

And by the way event handler is initially attached with `.on`:

```js
$e.on('click.yiiCaptcha', function () {
    methods.refresh.apply($e);
    return false;
});
```

so it's better to remove it in similar fashion.

So the final replacement will be:

```js
destroy: function () {
    this.unbind('.yiiCaptcha');
    this.removeData('yiiCaptcha');
    return this;
},
```

I guess this problem did not pop out eariler because the usage of captcha is usually limited to just tuning it on server side without making any JS changes, and lack of tests obviously.

In #12936 I haven't posted problems in separate issues, so do it for `yii.captcha.js` now to be more clear.

The same applies to other JS plugins such as `yii.gridView.js`.

arogachev:
| Q             | A
| ------------- | ---
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12840, #13159

### DIFF
--- a/docs/internals/git-workflow.md
+++ b/docs/internals/git-workflow.md
@@ -39,6 +39,14 @@ The following steps are not necessary if you want to work only on translations o
 
 > Note: If you see errors like `Problem 1 The requested package bower-asset/jquery could not be found in any version, there may be a typo in the package name.`, you will need to run `composer global require "fxp/composer-asset-plugin:^1.2.0"`
 
+If you are going to work with JavaScript:
+
+- run `npm install` to install JavaScript testing tools and dependencies (assuming you have [Node.js and NPM installed]
+(https://nodejs.org/en/download/package-manager/)).
+
+> Note: JavaScript tests depend on [jsdom](https://github.com/tmpvar/jsdom) library which requires Node.js 4 or newer.
+Using of Node.js 6 or 7 is more preferable.
+
 - run `php build/build dev/app basic` to clone the basic app and install composer dependencies for the basic app.
   This command will install foreign composer packages as normal but will link the yii2 repo to
   the currently checked out repo, so you have one instance of all the code installed.
@@ -63,7 +71,13 @@ Some tests require additional databases to be set up and configured. You can cre
 settings that are configured in `tests/data/config.php`.
 
 You may limit the tests to a group of tests you are working on e.g. to run only tests for the validators and redis
-`phpunit --group=validators,redis`. You get the list of available groups by running `phpunit --list-groups`. 
+`phpunit --group=validators,redis`. You get the list of available groups by running `phpunit --list-groups`.
+
+You can execute JavaScript unit tests by running `npm test` in the repo root directory.
+
+> Note: If you get timeout errors like `Error: timeout of 2000ms exceeded. Ensure the done() callback is being called
+in this test.`, you can increase timeout: `npm test -- --timeout 30000` (don't miss `--`, it's needed for passing
+additional arguments).
 
 ### Extensions
 
--- a/framework/assets/yii.captcha.js
+++ b/framework/assets/yii.captcha.js
@@ -16,7 +16,7 @@
         } else if (typeof method === 'object' || !method) {
             return methods.init.apply(this, arguments);
         } else {
-            $.error('Method ' + method + ' does not exist on jQuery.yiiCaptcha');
+            $.error('Method ' + method + ' does not exist in jQuery.yiiCaptcha');
             return false;
         }
     };
@@ -39,7 +39,6 @@
                     methods.refresh.apply($e);
                     return false;
                 });
-
             });
         },
 
@@ -58,10 +57,9 @@
         },
 
         destroy: function () {
-            return this.each(function () {
-                $(window).unbind('.yiiCaptcha');
-                $(this).removeData('yiiCaptcha');
-            });
+            this.off('.yiiCaptcha');
+            this.removeData('yiiCaptcha');
+            return this;
         },
 
         data: function () {
@@ -69,4 +67,3 @@
         }
     };
 })(window.jQuery);
-
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "sinon": "^1.17.6"
   },
   "scripts": {
-    "test": "./node_modules/mocha/bin/mocha tests/js/tests/*.test.js"
+    "test": "./node_modules/mocha/bin/mocha tests/js/tests/*.test.js --timeout 30000"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "sinon": "^1.17.6"
   },
   "scripts": {
-    "test": "./node_modules/mocha/bin/mocha tests/js/tests/*.test.js --timeout 30000"
+    "test": "./node_modules/mocha/bin/mocha tests/js/tests/*.test.js"
   },
   "repository": {
     "type": "git",
--- a/tests/js/tests/yii.captcha.test.js
+++ b/tests/js/tests/yii.captcha.test.js
@@ -51,30 +51,21 @@ describe('yii.captcha', function () {
             hashKey: 'yiiCaptcha/posts/captcha'
         };
 
-        describe('with single element', function () {
-            withData({
-                'no method specified': [function () {
-                    $captcha = $('#captcha').yiiCaptcha(settings);
-                }, settings],
-                'no method specified, custom options': [function () {
-                    $captcha = $('#captcha').yiiCaptcha(customSettings);
-                }, customSettings],
-                'manual method call': [function () {
-                    $captcha = $('#captcha').yiiCaptcha('init', settings);
-                }, settings]
-            }, function (initFunction, expectedSettings) {
-                it('should save settings', function () {
-                    initFunction();
-                    assert.deepEqual($captcha.data('yiiCaptcha'), {settings: expectedSettings});
-                });
-            });
-        });
-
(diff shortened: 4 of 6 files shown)


framework/assets/yii.captcha.js
```diff
@@ -16,7 +16,7 @@
         } else if (typeof method === 'object' || !method) {
             return methods.init.apply(this, arguments);
         } else {
-            $.error('Method ' + method + ' does not exist on jQuery.yiiCaptcha');
+            $.error('Method ' + method + ' does not exist in jQuery.yiiCaptcha');
```
arogachev: Small grammar nitpick.

framework/assets/yii.captcha.js
```diff
@@ -58,15 +57,13 @@
         },
 
         destroy: function () {
-            return this.each(function () {
-                $(window).unbind('.yiiCaptcha');
-                $(this).removeData('yiiCaptcha');
-            });
+            this.off('.yiiCaptcha');
```
arogachev: Fix for #13159.

package.json
```diff
@@ -16,7 +16,7 @@
     "sinon": "^1.17.6"
   },
   "scripts": {
-    "test": "./node_modules/mocha/bin/mocha tests/js/tests/*.test.js"
+    "test": "./node_modules/mocha/bin/mocha tests/js/tests/*.test.js --timeout 30000"
```
arogachev: See #12936:

> Error while running npm test:
>
> > Error: timeout of 15000ms exceeded. Ensure the `done()` callback is being called in this test.
>
> Happens very seldom, this is because of jsdom and integrating jQuery. Default timeout is 2000ms. And this is always not enough, so I increased it to be 15000ms.

I haven't seen this even once with Travis, only locally. I guess maybe it's because I work through Vagrant? Anyways increasing timeout to 30s solves the problem. I moved it here because jQuery is used everywhere (so jsdom needs to be used in every test file) to avoid repeating it in every test file.

arogachev:
For convenience and speed on dev server I included `yii.captcha.js` only in tests suite. Tests were passing, when I added `yii.validation.js` back, some errors appeared. Trying to figure out what can cause them.

framework/assets/yii.captcha.js
```diff
@@ -58,15 +57,13 @@
...
+            this.removeData('yiiCaptcha');
+            return this;
         },
 
         data: function () {
             return this.data('yiiCaptcha');
         }
     };
 })(window.jQuery);
-
```
arogachev: Extra linebreak at the very end of file..

framework/assets/yii.captcha.js
```diff
@@ -39,7 +39,6 @@
                     methods.refresh.apply($e);
                     return false;
                 });
-
```
arogachev: Extra linebreak.

arogachev:
Currently there are some problems with Sinon's fake server and Node.js, see https://github.com/sinonjs/sinon/issues/657. Used this solution as a workaround https://github.com/sinonjs/sinon/issues/657#issuecomment-261739635.

dynasource:
Is it possible to share a PR when it's close to ready? Now everybody is getting notifications of updates during WIP and Im not sure if it's ready for reviewing already  ;)

arogachev:
@dynasource It's ready. Please review. I can't think of a better workaround for testing AJAX now and it's not something ugly. Not sure when it will be fixed in Sinon. We need to test AJAX anyways.

tests/js/tests/yii.captcha.test.js
```diff
@@ -0,0 +1,192 @@
...
+var sinon;
+var withData = require('leche').withData;
+var jsdom = require('mocha-jsdom');
+
+var fs = require('fs');
+var vm = require('vm');
+
+describe('yii.captcha', function () {
+    var yiiCaptchaPath = 'framework/assets/yii.captcha.js';
+    var jQueryPath = 'vendor/bower/jquery/dist/jquery.js';
```
dynasource: shouldnt we use NPM for this instead of Bower?

package.json
dynasource: I guess so. Just tested your "yii.captcha.test.js" locally in 31ms so everything seems to work fine.

tests/js/tests/yii.captcha.test.js
arogachev: @dynasource I used jQuery that ships with Yii2 by default. This prevents from adding the same package to NPM (other packages include `punycode` for example in #12936) and synchronizing the versions on changes. I described it in #12936 and it was approved. With the current implementation of loading framework's JS dependencies I think it's better. When we get rid of Composer Asset Plugin (it will happen in 2.1 I believe) or make js less coupled with PHP, we can remove this as well.

tests/js/tests/yii.captcha.test.js
```diff
@@ -0,0 +1,192 @@
...
+
+                assert.strictEqual(destroyResult, $captcha);
+                assert.isFalse(ajaxStub.called);
+                assert.isUndefined($captcha1.data('yiiCaptcha'));
+                assert.isUndefined($captcha2.data('yiiCaptcha'));
+            });
+        });
+
+        describe('with multiple elements, destroy single element', function () {
+            var message = 'should remove event handlers with saved settings for destroyed element only and return ' +
```
dynasource: are the previous methods really necessary if you have this method already?

dynasource (changes requested):
looks good. Well done.  Note, the tests could be slimmed down because of overlap.

package.json
arogachev: In this case we can remove timeout from here and add a hint in developer's README in "Running JS tests" section about specifying it explicitly (`--timeout 30000`), in case of such error.

However I'm not sure if default `2000ms` will be enough with Travis for entire test suite. But we can try and increase it in case of problems though.

dynasource: can you find out why your environment needs 15 seconds? Does it also occur in an empty test? Is it related to network connections?

arogachev: @dynasource Seems like it's Vagrant problem. This error appeared after adding jsdom, so with empty test or tests without jsdom default timeout (2000ms) is not exceeded.

As for specific value (15000ms) - I tried to increase it gradually till the error is gone. But sometimes even `15000ms` was not enough so I increased it even more. I noticed that error generally happens on the first run of `npm test` then it's gone (I mean the case when it's set 15000ms, 2000ms is always not enough for me on Vagrant).

Let's try to remove timeout completely and see if default will be enough for Travis. We can increase it when it will be necessary. a Also adding note to the docs won't be excess.

docs/internals/git-workflow.md
```diff
@@ -63,7 +71,13 @@ Some tests require additional databases to be set up and configured. You can cre
 settings that are configured in `tests/data/config.php`.
 
 You may limit the tests to a group of tests you are working on e.g. to run only tests for the validators and redis
-`phpunit --group=validators,redis`. You get the list of available groups by running `phpunit --list-groups`. 
+`phpunit --group=validators,redis`. You get the list of available groups by running `phpunit --list-groups`.
```
arogachev: This line was not changed, IDE automatically removed extra whitespace at the end. Github does not show it for some reasons.

tests/js/tests/yii.captcha.test.js
arogachev: @dynasource Simplified it along with `init` method.

dynasource (approved):
thanks for improving it further. Keep up the good work.

samdark (approved):

docs/internals/git-workflow.md
cebe: Tip: you can add `?w=0` to any github ULR that shows a diff to exclude whitespace.

SilverFire:
Merged, thank you very much for your effort!

Issue #13154, opened 2016-12-07
Add comment property to database classes


michaelarnauts:
A database can contain comments on the columns and the table itself. The comment of the columns can be read out trough the `ColumnSchema` class, but the comment on the table itself isn't.

Is this something that is open for inclusion in Yii 2? This is a bit related to #6093, but that PR wasn't accepted, so I want to avoid to work on this if it won't be accepted.

This could be used to generate usefull phpdoc with Gii based on the database comments.

samdark:
I'm OK having it.

dynasource:
me too. Table comments are useful, like i.e. for sorting in Navicat.

privateproperty123:
One offers you a hassle-free way of selling your house fast while the other could make you more money. We put property cash buyers and online estate agents head-to-head to find out who comes out on top.

Online estate agents

Much like online banks, online estate agents now rival traditional high street agents as the best way to sell your home. Some even help with viewings, but be warned: because they usually don’t have local knowledge, they might not be as in tune with the market. To get more detail please visit : - [http://privatepropertybuyers.co.uk/](http://privatepropertybuyers.co.uk/) .

PR #13162, opened 2016-12-08
Load MySQL comments into TableSchema

michaelarnauts:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13154 (partially)

### DIFF
--- a/framework/db/TableSchema.php
+++ b/framework/db/TableSchema.php
@@ -35,6 +35,10 @@ class TableSchema extends Object
      */
     public $fullName;
     /**
+     * @var string the comment that has been assigned to this table.
+     */
+    public $comment;
+    /**
      * @var string[] primary keys of this table.
      */
     public $primaryKey = [];
--- a/framework/db/TableSchema.php
+++ b/framework/db/TableSchema.php
@@ -36,6 +36,7 @@ class TableSchema extends Object
     public $fullName;
     /**
      * @var string the comment that has been assigned to this table.
+     * @since 2.0.12
      */
     public $comment;
     /**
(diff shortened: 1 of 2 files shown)


framework/db/TableSchema.php
```diff
@@ -35,6 +35,10 @@ class TableSchema extends Object
      */
     public $fullName;
     /**
+     * @var string the comment that has been assigned to this table.
```
samdark: Need `@since 2.0.12`.

framework/db/mysql/Schema.php
```diff
@@ -122,6 +123,27 @@ protected function resolveTableNames($table, $name)
     }
 
     /**
+     * Loads the table comment into the [[TableSchema]] object.
+     * @param TableSchema $table the table metadata object
```
samdark: Need `@since 2.0.12`.

samdark: How about other database engines?

samdark (changes requested):

michaelarnauts:
I've added the `@since 2.0.12` to the phpdoc.

samdark:
Great. We can't really release it for MySQL only so will see if we'll have enough time to implement all other DB engines support before 2.0.12...

michaelarnauts:
* `mssql` uses a strange way to store comments, by using "extended properties", not sure if this is considered a standard practice to do so (http://stackoverflow.com/questions/378700/is-it-possible-to-add-a-description-comment-to-a-table-in-microsoft-sql-2000)
* `postgres` seems to provide a decent way to do this (http://stackoverflow.com/questions/5664094/getting-list-of-table-comments-in-postgresql)

I'm not sure about other databases.

samdark:
Oracle: http://stackoverflow.com/questions/16567129/how-to-get-table-comments-via-sql-in-oracle

samdark:
SQLite seems not to support comments at all.

samdark:
Won't make it into 2.0.

samdark:
https://github.com/yiisoft/yii2/pull/13162

PR #13168, opened 2016-12-08
Allow single role in filters/AccessRule.php

thyseus:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | n/a

Allows

'roles' => '@'

in addition to

'roles' => ['@']

in AccessRule.php. Tests are not yet available since 'roles' are not tested.


samdark:
I'm against this change because it makes overall argument type not well defined.

SilverFire (changes requested):
Disagreed. See @samdark comments

SilverFire:
Sorry, but the PR will not be accepted

Issue #7462, opened 2015-02-26
UploadedFile::saveAs doesn't support alias


Faryshta:
The docs doesn't mention it should support it but i think it would make uploading / downloading files more easily

qiangxue:
This was discussed before. We will not support it. Aliases are primarily supported for configurable properties.

Faryshta:
What do you mean?

What i want is to do Something like `$file->saveAs('@pdf/' . $file->baseName . '.pdf')`

lynicidn:
https://github.com/yiisoft/yii2/issues/6289

amlopezalonso:
Before this change I had:

In `params-local.php`:

     `return ['myRoot' => '@web/myroot'];`

...and in `controller`:

`$fullPath = Yii::$app->params[$myRoot] . '/myLast';`
` FileHelper::createDirectory($fullPath);`

and had no trouble with it. Now I got:

`FileHelper::createDirectory(Yii::getAlias($fullPath));`

And it throws: 

Exception – yii\base\Exception
Failed to create directory '/web': mkdir(): Permission denied

Any ideas?
Antonio

SilverFire:
@amlopezalonso please note, that the GitHub Issue Tracker is for bug reports and feature requests only.

We are happy to help you on the [support forum](http://www.yiiframework.com/forum/), on [IRC](http://www.yiiframework.com/chat/) (#yii on freenode), or [Gitter](https://gitter.im/yiisoft/yii2).


> Failed to create directory '/web': mkdir(): Permission denied

The user that runs the script does not have enough permissions to create a directory.

PR #13176, opened 2016-12-09
UploadedFile saveAs with alias

DrDeath72:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes


dynasource (changes requested):
looks useful
- please add a test for it

developeruz:
@dynasource, we don't have a test for `saveAs` method because it is quite difficult to test uploaded files in a unit test without real request. Test for this PR already exists because all you need is to test that `Yii::getAlias()` works as expected.

samdark (approved):

rob006:
This was already rejected: https://github.com/yiisoft/yii2/issues/7462#issuecomment-76290267

I don't think we should merge this. If someone needs to use alias, he could just pass `Yii::getAlias()` result to `saveAs()`.

See also https://github.com/yiisoft/yii2/blob/master/docs/internals/design-decisions.md

Issue #12312, opened 2016-08-25
Proposal: Enhance cache component to cache values by callback


mikehaertl:
When using the cache component I often find myself writing the same repetitive pattern like this:

``` php
<?php
class Article
{
    public function getAvailability()
    {
        $cache = Yii::$app->cache;
        $key = 'article:availabilty:'.$this->number;
        $cachedValue = $cache->get($key);
        if ($cachedValue===false) {
            // ... some expensive operation here e.g. a remote API call ...
            $cache->set($key, $value);
            return $value;
        } else {
            return $cachedValue;
        }
    }
}
```

To make this easier, I propose to add a new very simple method to `yii\caching\Cache`:

``` php
    /**
     * Gets the value stored under $key from cache. If there is no value stored for
     * that key, the value is evaluated from the callback and stored to the cache.
     *
     * @param string $key the cache key
     * @param Callable $callback the callback that evaluates the cache value
     * @param int $duration the caching duration. Default is 3600.
     * @param yii\caching\Dependency|null $dependency the cache dependency
     * or `null` for none.
     * @return mixed the cached value
     */
    public function cached($key, $callback, $duration = 3600, $dependency = null)
    {
        $cachedValue = $this->get($key);
        if ($cachedValue===false) {
            Yii::trace("Evaluating and storing value for $key", __METHOD__);
            $value = call_user_func($callback);
            $this->set($key, $value, $duration, $dependency);
            return $value;
        } else {
            Yii::trace("Serving $key from cache", __METHOD__);
            return $cachedValue;
        }
    }
```

With this method, the above getter can be simplified to:

``` php
    public function getAvailability()
    {
        $key = 'article:availabilty:'.$this->number;
        return Yii::$app->cache->cached($key, function () {
            // ... some expensive operation here e.g. a remote API call ...
            return $value;
        });
    }
```

As an alternative, we could even enhance the implementation of `Cache::get()`. I'm missing a `$default` argument there anyway. This is how it could look:

``` php
/**
 * @param string $key the cache key
 * @param mixed|Callable $default the default value to return if $key does not exist.
 * If this is a Callable, its return value is stored to the cache under the given key.
 * The default is `false`.
 * @param int $duration the caching duration if a callback is used. Default is 3600.
 * @param yii\caching\Dependency|null $dependency the cache dependency if a callback
 * is used or `null` for none.
 * @return mixed the cached or evaluated value
 */
public function get($key, $default = false, $duration = 3600, $dependency = null)
```

[...]

PR #13186, opened 2016-12-11
Added Cache::getOrSet() method

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #11959, #12312, #11173


framework/caching/Cache.php
```diff
@@ -496,7 +498,7 @@ protected function addValues($data, $duration)
     /**
      * Returns whether there is a cache entry with a specified key.
      * This method is required by the interface [[\ArrayAccess]].
-     * @param string $key a key identifying the cached value
+     * @param string $key a key identifying the///// cached value
```
samdark: ?

framework/caching/Cache.php
```diff
@@ -537,4 +539,45 @@ public function offsetUnset($key)
...
+     * @param \Closure $closure the closure that will be used to generate a value to be cached
+     * @param int $duration default duration in seconds before the cache will expire. If not set,
+     * default [[defaultDuration]] value is used.
+     * @param Dependency $dependency dependency of the cached item. If the dependency changes,
+     * the corresponding value in the cache will be invalidated when it is fetched via [[get()]].
+     * This parameter is ignored if [[serializer]] is false.
+     * @return mixed result of $closure execution
+     * @since 2.0.11
+     */
+    public function closure($key, \Closure $closure, $duration = null, $dependency = null)
```
samdark: Method name isn't good. It's not calling what it does in any way. I'd name it `cached`.

samdark (changes requested):

framework/caching/Cache.php
SilverFire: `$cache->closure()` looks more readable than `$cache->cached()` as for me

cebe: `cached` sounds good. Not sure whether or not the active voice `cache` would be better but it may also confuse more as the subject is cache too.

Also the order of the parameters may fit better like this:

```php
$result = $cache->cached(function() {}, $key, $duration);
```

The primary goal of the code is mainly to assign the result of the closure to the variable. With which key it is cached and how long belongs together and is secondary.

samdark: Exactly. It reads as "result = cached function () {} with key=$key, duration=$duration,$dependency=$dependency".

framework/caching/Cache.php
```diff
@@ -537,4 +538,44 @@ public function offsetUnset($key)
...
+    {
+        if (($value = $this->get($key)) !== false) {
+            return $value;
+        }
+
+        $value = call_user_func($closure, $this);
+        if (!$this->set($key, $value, $duration, $dependency)) {
+            Yii::warning("Failed to set cache value with key \"$key\"", __METHOD__);
```
rob006: This will fail when `$key` is not a string.

SilverFire:
@mdomba suggested `getOrSet()`. What do you think?

cebe:
That would be the most explicit/unambiguous name imo.

samdark:
That is, at least, less confusing than alternatives.

klimov-paul:
Naming could be `ensure()`:

```php
$result = $cache->ensure(function() {}, $key, $duration, $dependency);
```

cebe:
What is ensured by it?

klimov-paul:
> What is ensured by it?

Value in cache. `Ensure` means return it if exist, if not - create.

klimov-paul:
Although I would swap first 2 arguments if this name will be used:

```php
$result = $cache->ensure($key, function() {});
$result = $cache->ensure($key, function() {}, $duration, $dependency);
```

rob006:
```php
$value = $cache->remember($key, $duration, function() {}, $dependency);
```

?

resurtm:
@rob006 I disagree. Duration is less important that function. Also making duration as penultimate parameter, and dependency as last parameter would make more sense—AFAIR it's common approach across the framework.

rob006:
@resurtm Duration is much shorter than function, and personally I almost always use custom duration in such situations. And this is much more readable:

```php
$value = $cache->rememeber('my-key', 300, function() {
	// some
	// complex
	// code
});
```

than 

```php
$value = $cache->rememeber('my-key', function() {
	// some
	// complex
	// code
}, 300);
```

But I will not insist on it, this was more method name proposition.

cebe:
@rob006 I like the name. But I agree with @resurtm about the order of parameters.

cebe:
```php
$value = $cache->rememeber(function() {
	// some
	// complex
	// code
}, 'my-key', 300);
```

would read as "remember function result as 'my-key' for 300 seconds"

mikehaertl:
Hmm, "remember"? Isn't that what caches always do? You could even replace `set()` with `remember()`. I can not see, how it makes the situation any clearer than the other suggestions.

cebe:
@mikehaertl good point :)

rob006:
@cebe 
`remember($key, $duration, function() {}, $dependency)` would read as "remember $key for $duration seconds with value $function".

@mikehaertl But "remember" also means "get something from memory". "Do you *remember* what is the value of $key? If not, use function result and *remember* it.". "Set" only means "put something there".

mikehaertl:
Well, I think we can come up with 10-50 more suggestions. Actually I had exactly the same naming problems. That's why I ended up with my [suggestion](https://github.com/yiisoft/yii2/issues/12312#issuecomment-242327056), to simply extend `get()` with further arguments. Because that's what we bascially do here: get a cached value - and use "defaults" (or a callback) if there is no stored value yet.

That comes close to what we do on other similar occasions like `yii\web\Request::get()`. The only difference is really that the default value is not passed as value but is obtained from a callback.

I can't think of any other arguments for `get()` that would make more sense. So I'd still vote for a minor BC breaking interface change here.

cebe:
The signature of `get()` is really simple right now, I am against making get a complicated method. The closest thing to `get()` would be `getOrSet()`:

```php
get($key)
getOrSet($key, function() {}, $duration, $dependency)
```

`getOrSet` extends this nicely here, also for autocompletion, and is the most obvious search for developers searching this functionality imo (same name is used by [Django](https://docs.djangoproject.com/en/1.10/topics/cache/#basic-usage) btw).

`remember` is the name used by [laravel](https://laravel.com/docs/5.3/cache#retrieving-items-from-the-cache) btw.

framework/caching/Cache.php
```diff
@@ -537,4 +538,44 @@ public function offsetUnset($key)
...
+        }
+
+        return $value;
```
mikehaertl: You could eliminate one `return $value` if you rewrite it like:

```php
$value = ...
if ($value === false) {
   $value = ...
   $this->set(...)
}
return $value;
```

dynasource:
the objective is a getter, so like @mikehaertl states, its logical to name it this way. It also 'ensures' that something is actually returned (like @klimov-paul suggests) so I'm expecting a naming like:
* ensuredGet()
* forcedGet()
* setGet()

samdark:
Personally I think what @cebe suggested i.e.

```php
get($key)
getOrSet($key, function() {}, $duration, $dependency)
```

is the best name we've got so far. "remember" sounds more like natural language but it's a bit more harder to comprehend because of the double meaning. It could be OK for native speakers though...

dynasource:
It's not really an "or" scenario, which is a bit confusing. But it's favorable to start with 'get' to be the next related method in line. Perhabs:
```getSet()```

**Comment by cebe** [comment]
> It's not really an "or" scenario, which is a bit confusing.

it is. It either gets the value from the cache or computes it and sets it.

**Comment by dynasource** [comment]
if you say "or" I dont expect any "get" action anymore, so it's illogical with respect to https://github.com/yiisoft/yii2/blob/66ebd797a48172837cd45f88cca3eda664d864d3/framework/caching/Cache.php#L579

**Comment by etki** [comment]
Java has `putIfAbsent` and `computeIfAbsent` methods in standard map, if you're interested in similar experience.

**Comment by etki** [comment]
By the way, i'm not familiar with current Yii internals, but, looking at the code, i have a feeling that `false` can't be cached (and will be recomputed on every hit). If this is true, this is probably not the behavior users will expect.

**Comment by cebe** [comment]
> By the way, i'm not familiar with current Yii internals, but, looking at the code, i have a feeling that false can't be cached (and will be recomputed on every hit). If this is true, this is probably not the behavior users will expect.

This is true and documented http://www.yiiframework.com/doc-2.0/guide-caching-data.html#cache-apis it is the compromise for a simple and fast API, that does not include wrapping of values in a container object like it is done in the cache PSR proposal for example.

**Comment by etki** [comment]
@cebe i understand that, and that it is perfectly valid in other scenarios - i.e. if you have cached false, you'll end up with same one request to cache and will receive false in any case. However, in compute-if-absent case algorithm is not fully dependent on value stored in cache and may cause unwanted computations; this will become obvious and painful as soon as end user will bring side effects in computation (yes, it is completely wrong, but this surely will happen) or just if user tries to cache some hot boolean expression. I think this deserves at least huge capslocked warning in documentation.

**Comment by etki** [comment]
silly me, didn't look at the documentation. disregard previous comment.

**Comment by sergeymakinen** [comment]
Hm. If we need a name that describes the method, maybe let's do it:
```php
getOrSetAndReturn($key, function() {}, $duration, $dependency)
```

?

cebe:
> getOrSetAndReturn

this is what it does and it seems to me that is the reason we can not find a good name. It simply does too much.

sergeymakinen:
> It simply does too much.

I think it does exactly what it should do for user's cases. It just lacks an obvious/well-known name. If `getOrSetAndReturn` sounds too ugly: `cache` or `remember` names fit good IMHO. I just feel the first one (`getOrSetAndReturn`) is a Yii way while the second is a Laravel "expressive" way. I don't know why. :)

kidol:
I like `ensure()` best so far. For another proposal: `load()`.

Also I prefer the closure to be the 2nd argument.

resurtm:
@sergeymakinen 

> getOrSetAndReturn

There is much better variant (and more concise!) mentioned above:

> putIfAbsent
> computeIfAbsent

I'd pick the one proposed by @etki.

mikehaertl:
I agree to what @dynasource said above:

> if you say "or" I dont expect any "get" action anymore,

We *always* get the result of the callback. And it gets cached for later calls. So how about:

```
getCachedResult(function() { }, $key, $duration, $dependency)
```

dynasource:
This is an interesting discussion. 

Whats seems more than logical to me, seems illogical to others and vice versa. Looks to have multiple root causes:
* language interpretations: "set" is an irregular English verb https://www.usingenglish.com/reference/irregular-verbs/set.html that can be interpreted as verb, past simple & past participle
* everyone has its own "naming" style: by method objective, by contents & word preference

framework/caching/Cache.php
```diff
@@ -537,4 +538,44 @@ public function offsetUnset($key)
...
+     *     $cache = $this->cache; // Could be Yii::$app->cache
+     *     return $cache->closure(['top-n-products', 'n' => $count], function ($cache) use ($count) {
+     *         return Products::find()->mostPopular()->limit(10)->all();
+     *     }, 1000);
+     * }
+     * ```
+     *
+     * @param mixed $key a key identifying the value to be cached. This can be a simple string or
+     * a complex data structure consisting of factors representing the key.
+     * @param \Closure $closure the closure that will be used to generate a value to be cached
```
mikehaertl: @SilverFire I think you should add a note like: "Note, that the return value `false` will not get cached.".

SilverFire:
Renamed to `getOrSet()`

samdark (approved):

framework/CHANGELOG.md
```diff
@@ -70,6 +70,7 @@ Yii Framework 2 Change Log
 - Enh: Added constants for specifying `yii\validators\CompareValidator::$type` (cebe)
 - Enh #12854: Added `RangeNotSatisfiableHttpException` to cover HTTP error 416 file request exceptions (zalatov)
 - Enh #13122: Optimized query for information about foreign keys in `yii\db\oci` (zlakomanoff)
+- Enh #11959: Added `yii\caching\Cache::closure()` method to cache closures (silverfire)
```
rob006: `Cache::closure()` -> `Cache::getOrSet()`

framework/caching/Cache.php
```diff
@@ -537,4 +538,45 @@ public function offsetUnset($key)
...
+     * }
+     * ```
+     *
+     * @param mixed $key a key identifying the value to be cached. This can be a simple string or
+     * a complex data structure consisting of factors representing the key.
+     * @param \Closure $closure the closure that will be used to generate a value to be cached.
+     * In case $closure returns `false`, the value will not be cached.
+     * @param int $duration default duration in seconds before the cache will expire. If not set,
+     * default [[defaultDuration]] value is used.
+     * @param Dependency $dependency dependency of the cached item. If the dependency changes,
```
rob006: `default [[defaultDuration]]` => `[[defaultDuration]]`

framework/caching/Cache.php
```diff
@@ -537,4 +538,45 @@ public function offsetUnset($key)
...
+     * the corresponding value in the cache will be invalidated when it is fetched via [[get()]].
+     * This parameter is ignored if [[serializer]] is false.
```
rob006: `false` => `` `false` ``

framework/caching/Cache.php
```diff
@@ -537,4 +538,45 @@ public function offsetUnset($key)
...
+     * ```
+     *
+     * @param mixed $key a key identifying the value to be cached. This can be a simple string or
+     * a complex data structure consisting of factors representing the key.
+     * @param \Closure $closure the closure that will be used to generate a value to be cached.
+     * In case $closure returns `false`, the value will not be cached.
+     * @param int $duration default duration in seconds before the cache will expire. If not set,
+     * default [[defaultDuration]] value is used.
+     * @param Dependency $dependency dependency of the cached item. If the dependency changes,
+     * the corresponding value in the cache will be invalidated when it is fetched via [[get()]].
```
rob006: `when it is fetched via [[get()]]` - this is confusing - it will be invalidated also in `getOrSet()` method, not only `get()`.

rob006 (changes requested):

framework/caching/Cache.php
```diff
@@ -537,4 +538,45 @@ public function offsetUnset($key)
...
+    public function getOrSet($key, \Closure $closure, $duration = null, $dependency = null)
+    {
+        if (($value = $this->get($key)) !== false) {
+            return $value;
+        }
+
+        $value = call_user_func($closure, $this);
+        if (!$this->set($key, $value, $duration, $dependency)) {
+            Yii::warning('Failed to set cache value for key "' . json_encode($value) . '"', __METHOD__);
+        }
```
rob006: This will return `Failed to set cache value for key ""string""'` for strings?

mikehaertl:
"OR" now really? :-1:  Where is there an OR in this operation? This name is plain incorrect and the worst pick from all the suggestions above.

"Get me the value OR set it" does not describe what the code does. The code does really "Get me the value of that method. If it's cached, get me the cached value instead. If not, execute the method an store it"

**There is no OR here!**

mikehaertl:
Again: Why was this discussion ended? Someone now made a decision which is obviously wrong! Please reconsider this decision. We really can do better than this I think.

samdark:
@mikehaertl obviously, it's not possible to have a method with the accurate description in the name so we have what we have. Other names aren't any better.

mikehaertl:
But this name is **wrong**:

If you tell someone: *Get up and fight or sit down and have a drink*

What do you think this person will do? Will he still fight? Or simply sit down and drink?

**OR* implies, there's a choice between two things. But here is no choice. So `getOrSet()` is simply not the correct description what this method does. There were much clearer suggestions. We could at least pick one, which is not wrong!

samdark:
> What do you think this person will do? Will he still fight? Or simply sit down and drink?

There are two choices available. That's up to the person.

kidol:
Since there is no best name for this method, at least pick a short one from the suggestions.

mikehaertl:
> There are two choices available. That's up to the person.

That's my point: If you say "or" it implies there is a choice. In our case there is no choice between "get OR set". It's not just wrong, it's also not telling you anything about what this method does. Even `cached()`, `closure()`, `ensure()` or my suggested `getCachedResult()` are better - none of them is logically wrong. But the current choice definitely is.

mikehaertl:
One more suggestion, which at least is not wrong: `getAndCache()`. It's rather an AND than an OR.

bizley:
`getAndCache()` produces really weird `andCache` property.

I still like `ensure` better. No one said it's bad, word meaning matches the purpose, it's short... What else we need?

samdark:
@mikehaertl 

> If you say "or" it implies there is a choice. In our case there is no choice between "get OR set".

But there is a choice. What action is taken is chosen based on current state. It's similar to `getPostOrThrowException()` i.e. you're not deciding which action is taken, it's chosen for you.

mikehaertl:
> But there is a choice. What action is taken is chosen based on current state.

This is pretty much true for almost any method with medium complexity. Just picking a random example from `yii\base\Request`: 

```
    public function getScriptFile()
    {
        if ($this->_scriptFile === null) {
            if (isset($_SERVER['SCRIPT_FILENAME'])) {
                $this->setScriptFile($_SERVER['SCRIPT_FILENAME']);
            } else {
                throw new InvalidConfigException('Unable to determine the entry script file path.');
            }
        }

        return $this->_scriptFile;
    }
```

Shouldn't we call this method `getOrSetScriptFile()` then?

samdark:
No because it's not accepting any value so clearly it's a simple getter with static cache.

mikehaertl:
But even your example is different: `getPostOrThrowException()`: This will **either** get a value **or** throw an exception. Perfectly fine. It's never both.

Method names should explain what the method does from a logical perspective in plain simple words. It should not go into details about the internal implementation (which is what you said above about "choice": Every method with an `if` inside is about a choice, but that's irrelevant and doesn't enforce an "Or" in the method name).

The description of this method from a logical perspective is:

 * Always return me the result of the evalutated closure
 * Cache this result and return this cached result on later calls

Where exactly is here any **or** in this description?

dynasource:
@mikehaertl, I appreciate your persistence on this matter. It clearly shows that "getOrSet" is too controversial to accept.

SilverFire:
"Get value from cache by the key, or set it, when it does not exist in cache"

SilverFire:
I just want to say that language is flexible. Presence of **OR** depends on how you want to compose your message. In case you want to say something without **OR** - you will do it for sure.

As it was said many times before, there will be no absolute winner, because the method does too much to say it in one word. (We can create a new word for this. How about `peloozoid()` or `safome()`?)

I've choosen `getOrSet()` for the following reasons:

 - It gets
 - It sets
 - It will be shown on IDE autocomplete for `get`, making it easier to find
 - It is simple and short

rob006:
How about `getCached()`?

samdark:
@rob006 how about `executeCached`?

rob006:
> how about executeCached?

It sounds like "execute something from cache".

dynasource:
I vote for `getCached`

SilverFire:
`getCached` has nothing about `set`

alexraputa:
How about
``` php
$result = $cache->caching($key, function() {}, $duration, $dependency);
```
or 
``` php
$result = $cache->store($key, function() {}, $duration, $dependency);
```
?

dynasource:
> getCached has nothing about set

the word `cached` implies that already

rob006:
> getCached has nothing about set

Does it have to?  There are already getters in Yii that have side effects and save some values in cache. Why method name should say everything about how it works?

rob006:
Also, it not really `set` but `add` something to cache.

dynasource:
> Why method name should say everything about how it works

Thats strokes with @mikehaertl arguments about the focus on `what` a method does instead of `how` he does it, favoring a declarative programming style over a imperative style.

mikehaertl:
I think `getCached()`, `cached()`, `executeCached()` - they (and others) are all fine and make sense. I could live with that. If you find any of them in the code and look at the arguments, it should be easy to guess what they do.

But this brings me to my new favourite:

```
$expensive = $cache->execute(function() { }, 'expensive');
```

It's missing the "*please cache this*" part? Maybe, but why the heck would I call an `execute()` method on a **cache** component and pass a long a callable and something that looks like a cache key? I would say, even someone new to Yii could guess what this method does.

SilverFire:
`$cache->execute()` is extremely close to original `$cache->closure()`, isn't it?

`$cache->closure()` reads easy: `Cache this closure`. I can not read `$cache->execute()` as easy.

dynasource:
it's a sister of the `get` method and should have this prefix for easy comparison.

mikehaertl:
@SilverFire `closure` is not a verb but a noun. `execute` like most methods is a command.
@dynasource I'd say this is a little exception. I was thinking about other getter names too. But the discussion here already shows, that something's wrong with this approach. We also have methods like `resolve()` or `evaluate()` that return things and don't have a `get` prefix.

Just look at this list and let it sink in for a minute:  ;)

* `$cache->set()`: Set a value to cache
* `$cache->get()`: Get a cached value
* `$cache->execute()`: Execute a method and cache the value

SilverFire:
> `$cache->execute()`: Execute a method and cache the value

To be more precise, _Get the value from cache and return it. When it's not available, execute function, set it to cache and return value._

This argue will never end.

mikehaertl:
> This argue will never end.

That's my point. We should pick a simple concise mnemonic and move on.

Look at `Request::resolve()` (it's not `getResolved()` or something). Not every method that returns something must contain `get`. It's also not `Cache::getValue()` or even `getCachedValue()` but simply `Cache::get()`. Why I think `execute()` makes sense:

* Calling this on the `cache` component already implies that it's about caching something. Just like with `get()` or `set()`
* `execute()` is correct, because the method is at least called once
* Having a return value doesn't always enforce a `get` in the method name
* `get` should rather be reserved to "pure" getters that define virtual properties

dynasource:
> This argue will never end.

You're underestimating the flexibility of the community ;).
To agree on what we don't want is also a good approach to come to a result.

@SilverFire, whats your opinion about `ensure`?

samdark:
`execute` is a good variant. I have no objections.

rob006:
> get should rather be reserved to "pure" getters that define virtual properties

Definitely not. `get` prefix is common practice for any getters, reserve this only for "pure" getters sound extremely impractical and probably impossible.

Also, when you have something like that:

```php
$cache->getCached('most-popular-news', ...
``` 
You don't need to read rest of the code to know what is going on. Method name and first parameter says what you will get, rest is only implementation detail. 
This:
```php
$cache->execute(function() {
	// some
	// complex
	// code
}, 'most-popular-news');
```
requires much more attention to understand the code. Also `execute()` is far from obvious name for such logic, most people will start with `get` (because "get something" is what they want) and look into IDEs hints.

dynasource:
cannot agree more. 

Getting is done 100% of the time in this function. All the secondary "fallback" logic, which is only occassionally being executed, should logically come second.

mikehaertl:
> Definitely not. get prefix is common practice for any getters, reserve this only for "pure" getters sound extremely impractical and probably impossible.

Well, maybe, but that does not mean, that *anything* that returns a value must use the `get` prefix.

> ... requires much more attention to understand the code.

That's very arguable. Reading the code to me feels very natural: "*Ok cache, execute this method here for me: ...*". That it returns the evaluated value is so obvious that we don't need any getter prefix, because what else would make much sense to return?

> You don't need to read rest of the code to know what is going on

Could not disagree more :). You say, the *key* is more important than the logic which actually fetches the data you want. Let's look at a simple real world example:

```
$cars = $cache->get('cars');
if ($cars === false) {
    $cars = Car::find()->active()->all();
    $cache->set('cars', $cars);
}
```

This would become:

```
$cars = $cache->execute(function() {
    return Car::find()->active()->all();
}, 'cars');
```

Is the cache key `cars` really the important part here? I think not. The interesting part is not *how* is it cached (under which key), but *what* is cached (the expensive code fragment). 

And I'm not sure if this would work, but we could maybe even make the key optional and use the serialized callable as cache key if no key is provided. In this case you could simply call `$cache->execute([$object, 'method'])`.

rob006:
> That it returns the evaluated value is so obvious that we don't need any getter prefix, because what else would make much sense to return?

For example, `yii\db\Command::execute()` does not return inserted/updated records, it returns **number** of updated/inserted records. So no, it is definitely not obvious what `execute()` should return. For me the most obvious thing is that `execute()` should return boolean whether execution is succeed or failed. 

> Is the cache key cars really the important part here?

Your example is wrong. First, you chose wrong name for key (it should be 'active-cars' or 'active-cars-models'). Second, you chose most basic usecase - in many cases cached closure will be much more complicated, valuable key name will save you lots of time spent on reading closure body.

dynasource:
> Could not disagree more :). You say, the key is more important than the logic which actually fetches the data you want. Let's look at a simple real world example:

that's a strange example. Why would I execute something from the cache. I don't want to execute anything from the cache :). 

I want to retrieve something from the cache. That's what I want. And I want to do that by `key`, in your example the `car`. And if it doesn't exist? Well, for that risk (which does not occur often because its cached), I have a fallback closure which re ensures the the cache.

mikehaertl:
Heh, interesting how one can have different views on the same problem.

Ok, another try:

>  I have a fallback closure which re ensures the the cache.

I think this is the root of our disagreement and I think you're wrong here: It's not at all a fallback! The callback **will definitely be executed**. The cache only makes sure, that this happens only once. The key is completely irrelevant - it should even be autogenerated if the cache can do this for me.

Just imagine how the story evolves:

1. You have some code for an operation that returns some data
2. You find out that the operation takes too long or eats too many resources
3. You decide, that you want to cache the result of the operation

Why would you now say, that the implememntation is no longer interesting and the most interesting thing is a cache key?

Nothing has changed, except that you want to cache the result. *How* this is cached should be hidden away as much as possible. The key is not interesting, because apart from caching the result, nothing in your operation logic has changed.

If you think you need a cache label to know what you're doing, you're doing it wrong anyway: You should have moved your operation into a method with a meaningful name. You should not use a cache key to label your code. And you will not use the cache key anywhere else. You would never use `$cache->get('active-cars')` or the like. So why are you so interested in the key?

If you really need a label, assign the result to a meaningful name:

```
$activeCars = $cache->execute([$this, 'queryActiveCarsFromRemoteApi')
```

samdark:
> You should not use a cache key to label your code. And you will not use the cache key anywhere else. You would never use $cache->get('active-cars') or the like. So why are you so interested in the key?

To invalidate cache.

rob006:
> You would never use $cache->get('active-cars') or the like.

I would use `$cache->delete('active-cars')`.

> If you really need a label, assign the result to a meaningful name:

So instead:

```php
public function getActiveCars() {
    return Yii::$app->cache->getCached('active-cars', function () {
        //some code
    });
}
```

I need:

```php
public function getActiveCars() {
    return Yii::$app->cache->execute([$this, 'reallyGetActiveCars']);
}

public function reallyGetActiveCars() {
    //some code
}
```
? And you really think that this is better solution than using cache key as label?

mikehaertl:
> To invalidate cache.

That's what a dependency is for.

One final attempt then I give up.

What's the most easy and simple description what this method does in human language?

Here's the two versions that I see in this discussion:

* *Give me the value you have stored under X. If you don't have one, execute this method and store it under X and give it back to me.*
* *Execute this method, give me the result and remember the result under key X for the next time when I ask you to execute it again*

The first emphasises the key `X`. It's very close to the implementation. It's what the  `getCached($key, $callback)` etc. guys here like to have.

The second is an abstract human description of what this method *really* does for me: It *executes* and *remembers* things.

As the sole purpose of a cache is to remember things, we can leave away that part in the method name and end up with `execute()`.

mikehaertl:
> ? And you really think that this is better solution than using cache key as label?

You are notoriously missing my point. The operation whose result you want to cache has a purpose. It will stay the same, no matter if you cache it or not. You should not need a cache key to describe it. It should speak for itself.

Say, you decide one day, that you no longer cache it. Now suddenly you don't have a cache key anymore. Are you telling me, that you now don't understand your code anymore, because the cache key that labels your code is gone?

In your scenario you're abusing this method to stick a label on it. All I say: If you think your code is so complex that you don't understand it without a label, put it inside another method. But you should do this *anyway* then! It has nothing to do with the method we discuss here.

rob006:
> That's what a dependency is for.

In many cases using `delete()` is easier and faster that using dependencies. 

> All I say: If you think your code is so complex that you don't understand it without a label

I'm not talking about "I don't understand this code, so I need a label". I'm talking about "I don't want waste 5 seconds every time when read this code part - label will save me 4.5 second of this time". :) Cache keys have the same profits than extracting code to separate methods, but it does not force specific implementation.

mikehaertl:
> In many cases using delete() is easier and faster that using dependencies.

You still have a cache key. It's just not the main information.

We even have other places in the framework were you don't have any cache key at all:

```php
$result = $db->cache(function ($db) {
    return $db->createCommand('SELECT * FROM customer WHERE id=1')->queryOne();
});
```

SilverFire:
As we all know, there are only tho hard things in computer since: cache invalidation and naming things [©}(http://martinfowler.com/bliki/TwoHardThings.html). Now I can say for sure: naming cache-related things is the worst :smiley: 

Thank you, community, for sharing your opinions in this PR. As long as there is no general agreement on the method name, @yiisoft/core-developers team decided to vote internally.

The following method names were put to vote:

- `ensure()`
- `remember()`
- `execute()`
- `closure()`
- `cached()`
- `getCached()`
- `getOrSet()` 
- `getSet()`


I will publish a final decision and voters statistics as soon as it possible. 

For now I'm locking the conversation to prevent time additional time spends for this argue.

SilverFire:
Results:

Method name | Voters
------------ | -------------
`cached()` | 1: @samdark 
`getSet()` | 1: @dynasource 
`ensure()` | 1: @dynasource 
`getCached()` | 2: @dynasource, @samdark 
`getOrSet()` | 4: @silverfire, @samdark, @mdomba, @qiangxue 
other | no votes

@klimov-paul and @cebe have abstained from voting.

SilverFire:
Merged: https://github.com/yiisoft/yii2/commit/aed3a222a20edd14859053a45b6b62889ca41a59

Thank you, community, for active participating in this Pull Request.

PR #13189, opened 2016-12-13
Speed up hhvm unit tests - disable jit

photodude:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | speeds up the hhvm unit tests

Turn off the hhvm JIT to speed up the hhvm unit tests.

Source for why the JIT is bad for unit testing: facebook/hhvm#6979

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -117,6 +117,8 @@ install:
 #  - tests/data/travis/cubrid-setup.sh
 
 before_script:
+  # Disable the HHVM JIT for faster Unit Testing
+  - if [[ $TRAVIS_PHP_VERSION = hhv* ]]; then echo 'hhvm.jit = 0' >> /etc/hhvm/php.ini; fi
   # show some versions and env information
   - php -r "echo INTL_ICU_VERSION . \"\n\";"
   - php -r "echo INTL_ICU_DATA_VERSION . \"\n\";"


resurtm:
@photodude Have you done real time measurements for Yii2? Would be very interesting to see the difference.

photodude:
@resurtm this is specific to the unit tests, you would still want the JIT in production.

but the numbers for the hhvm unit tests are simple to look up from the travis logs (note: some variation on each run)

Although there is something going on here (possibly a build cache issue) as the patch didn't speed the tests up. (it's knocked about 2-3 minutes off the hhvm unit tests on other repo's I've applied this to)

before patch

hhvm-3.12 | hhvm (current)
------------ | ------------ 
4 min 51 sec | 5 min 19 sec

photodude:
Like I said, something is going on here (possibly a build cache issue). For example the [Yii repo was faster,](https://travis-ci.org/yiisoft/yii/builds/183481184)

For Yii: before patch 2 min 50 sec, and after patch 2 min 25 sec
which for Yii it's not as big a speed up as in other repos but still 25 seconds is a nice improvement.

cebe:
As far as I understood it, this only applies to the case where PHPUnit has to spawn a separate process. Also it does not really add much in most cases, see https://github.com/facebook/hhvm/issues/6979#issuecomment-210265920

The speedup in Yii could be caused by PHPUnit `@runInSeparateProcess` annotation. This is not used in yii2.

SilverFire:
JIT does not make sense for a short application lifetime cycle. IMHO JIT could be disabled.

cebe:
Makes sense to merge it anyway. Thank you!

photodude:
You are welcome

PR #13190, opened 2016-12-13
updated images gii

githubjeka:
Screen of Gii Doc is out of date.


cebe (approved):

dynasource (approved):

samdark:
Template paths are partially missing which isn't good.

In start-gii-crud.png there's a tooltip leftover.

start-gii-model-preview.png was displaying "diff" thing which was good for showing off features. Also there's something in the model class field (right side of the field).

samdark (changes requested):
Would be good to have full paths and no tip leftovers.

SilverFire:
@githubjeka would you like to apply changes suggested by @samdark, or you have objections?

githubjeka:
It's hard work to do screenshots. Tommorow to get involved in my work I will create new screenshots before.

cebe:
for better review: it would be great to only replace the english images and we copy them ourselfs because github does not display all images in the diff.

PR #13192, opened 2016-12-13
Throw InvalidConfigException when 'roles' isn´t an array

thyseus:
| Q             | A
| ------------- | ---
| Is bugfix?    | /no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Throw InvalidConfigException when 'roles' isn´t an array. Would have saved me a lot of debugging time.
This is another approach to https://github.com/yiisoft/yii2/pull/13168 .

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -57,6 +57,7 @@ Yii Framework 2 Change Log
 - Enh #12881: Added `removeValue` method to `yii\helpers\BaseArrayHelper` (nilsburg)
 - Enh #12901: Added `getDefaultHelpHeader` method to the `yii\console\controllers\HelpController` class to be able to override default help header in a class heir (diezztsk)
 - Enh #13035: Use ArrayHelper::getValue() in SluggableBehavior::getValue() (thyseus)
+- Enh #13192: Throw InvalidConfigException in filters/AccessRule.php when 'roles' isn´t an array (thyseus)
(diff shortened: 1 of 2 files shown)


resurtm:
@thyseus Why the framework needs an additional check considering the documentation clearly states we should only use array of the roles?

```php
    /**
     * @var array list of roles that this rule applies to. Two special roles are recognized, and
     * they are checked via [[User::isGuest]]:
     *
     * - `?`: matches a guest user (not authenticated yet)
     * - `@`: matches an authenticated user
     *
     * If you are using RBAC (Role-Based Access Control), you may also specify role or permission names.
     * In this case, [[User::can()]] will be called to check access.
     *
     * If this property is not set or empty, it means this rule applies to all roles.
     */
    public $roles;
```

> Would have saved me a lot of debugging time.

This is clearly your fault of misusing the framework classes.

cebe:
I am all for improving error messages that are unclear but in this case:

```
 PHP Warning – yii\base\ErrorException
Invalid argument supplied for foreach()

in /home/cebe/dev/yiisoft/yii2/framework/filters/AccessRule.php  at line 142 

...

foreach($this->roles as $role) {

...

```

It points to AccessRule -> $this->roles so you could directly check what you set there.

thyseus:
Ok, i agree to dismiss this change.

PR #13195, opened 2016-12-13
fix js compare validation for type double and integer

egosaw:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/framework/assets/yii.validation.js
+++ b/framework/assets/yii.validation.js
@@ -276,7 +276,7 @@ yii.validation = (function ($) {
                 compareValue = $('#' + options.compareAttribute).val();
             }
 
-            if (options.type === 'number') {
+            if (['double', 'integer', 'number'].indexOf(options.type) >= 0) {
                 value = parseFloat(value);
                 compareValue = parseFloat(compareValue);
             }


cebe:
What should this change go? `type` can only be  string or number: https://github.com/yiisoft/yii2/blob/ae83dd0524cab9841b9632c0c904aaac506a4a57/framework/validators/CompareValidator.php#L64-L70

egosaw:
There are two aliases for type number (double and integer) which are not considered in the code,
so the data are not converted by the parseFloat function and comparison is not correct.

samdark:
@egosaw how to reproduce the issue?

egosaw:
@samdark
```
public $price_old;
public $price_new;

    /**
     * @return array the validation rules.
     */
    public function rules()
    {
        return [
            [ ['price_old', 'price_new'], 'double', 'min' => 1 ],
            [ 'price_new', 'compare', 'compareAttribute' => 'price_old', 'operator' => '<=', 'type' => 'double' ],
        ];
    }
```

image
image

image

https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.validation.js#L279
image

cebe:
> There are two aliases for type number (double and integer)

where did you get these from? they are neither documented nor implemented in the CompareValidator class.

samdark:
Yes. That's incorrect usage. Supported types are `string` and `number` as documented in `CompareValidator` docs.

Thanks for effort but the pull request won't be accepted.

PR #13199, opened 2016-12-13
Added ability to set multiple event handlers

### LINKED ISSUE #11269: Array of event handlers in config
In `common\config\main.php` i'm use this config for :

``` php
   'components' => [ /* ... */ ],
   'on beforeRequest' => function () {
        \Yii::$container->set('yii\widgets\LinkPager', ['maxButtonCount' => 5]);
    },
```

But in `frontend\config\main.php` i'm need use second handler for styling Html:

``` php
   'components' => [ /* ... */ ],
   'on beforeRequest' => function () {
        Yii::$classMap['yii\helpers\Html'] = '@common/helpers/Html.php';
    },
```

Common config does not apply, because it uses `ArrayHelper::merge()` and `on beforeRequest` is overwrite.
## Suggestion for improvement

It would be nice if as values could be to use an array:

``` php
   'components' => [ /* ... */ ],
   'on beforeRequest' => [
      function () {
         \Yii::$container->set('yii\widgets\LinkPager', ['maxButtonCount' => 5]);
      }
   ],
```

| Q | A |
| --- | --- |
| Yii version | 2.0.7 |

hiqsol:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no

This feature is really required to set (through configuration) multiple event handlers for single event.
And ArrayHelper::isCallable can be helpful elsewhere.
If it is basically ok I will add tests.

Closes: #11269

### DIFF
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -451,14 +451,15 @@ class Component extends Object
     /**
      * Attaches an event handler to an event.
      *
-     * The event handler must be a valid PHP callback. The following are
-     * some examples:
+     * The event handler must be a valid PHP callback or array of callbacks.
+     * The following are some examples:
      *
      * ```
      * function ($event) { ... }         // anonymous function
      * [$object, 'handleClick']          // $object->handleClick()
      * ['Page', 'handleClick']           // Page::handleClick()
      * 'handleClick'                     // global function handleClick()
+     * ['handle1', ..., 'handleN']       // array of global functions
      * ```
      *
      * The event handler must be defined with the following signature,
@@ -481,6 +482,21 @@ class Component extends Object
     public function on($name, $handler, $data = null, $append = true)
     {
         $this->ensureBehaviors();
+        if (is_array($handler) && !is_callable($handler)) {
+            foreach ($handler as $single) {
+                $this->_on($name, $single, $data, $append);
+            }
+            return;
+        }
+        $this->_on($name, $handler, $data, $append);
+    }
+
+    /**
+     * Attaches a single event handler to an event.
+     * @see on()
+     */
+    private function _on($name, $handler, $data = null, $append = true)
+    {
         if ($append || empty($this->_events[$name])) {
             $this->_events[$name][] = [$handler, $data];
         } else {


framework/base/Component.php
```diff
@@ -451,14 +452,15 @@ public function hasEventHandlers($name)
...
-     * some examples:
+     * The event handler must be a valid PHP callback or array of callbacks.
+     * The following are some examples:
      *
      * ```
      * function ($event) { ... }         // anonymous function
      * [$object, 'handleClick']          // $object->handleClick()
      * ['Page', 'handleClick']           // Page::handleClick()
      * 'handleClick'                     // global function handleClick()
+     * ['handle1', ..., 'handleN']       // array of global functions
```
samdark: That could be an array of anything above, correct?

samdark:
Conceptually it looks OK to me.

framework/base/Component.php
```diff
@@ -481,6 +483,12 @@ public function hasEventHandlers($name)
     public function on($name, $handler, $data = null, $append = true)
     {
         $this->ensureBehaviors();
+        if (is_array($handler) && !ArrayHelper::isCallable($handler)) {
```
dynasource: ensureBehaviors() is called multiple times?

framework/base/Component.php
```diff
@@ -481,6 +483,12 @@ public function hasEventHandlers($name)
...
+            foreach ($handler as $single) {
+                $this->on($name, $single, $data, $append);
+            }
+            return;
```
dynasource: can't this solution be done without a new call to on(), by putting a handler always inside an array?

framework/helpers/BaseArrayHelper.php
```diff
@@ -770,6 +770,22 @@ public static function isTraversable($var)
...
+     *
+     * Callable array:
+     * - has exactly two elements
+     * - first element is object or string
+     * - second element is string.
+     * @param mixed $var The variable being checked.
+     * @return bool whether $var is callable array
+     * @since 2.0.11
+     */
+    public static function isCallable($var)
```
dynasource: because this is inside the ArrayHelper, Im expecting a $array and not just a $var

dynasource (changes requested):

framework/base/Component.php
SilverFire: Yes

cebe:
How do you distinguish attachment of two global functions from a single static class method handler?

hiqsol:
Rewritten with use of PHP internal `is_callable()` (which I just forget before).

Removed `ArrayHelper::isCallable()` in favour of `is_callable()` which answers to @cebe's question - it distinguishes those arrays properly.

klimov-paul:
Actually this issue was disscussed and rejected long ago: see #959
There is no sense of such improvement: for the regular program it is fine to call `on()` method multiple times, for the configuration - this make no sense.

I am against such change as it is risk prone and may lead to unexpected results as array-callbale is always hard to separate from array of callables.

For example:

```php
var_dump(is_callable(['SomeClass', 'someMethod'])); // outputs `true`
var_dump(is_callable(['UnexistingClass', 'someMethod'])); // outputs `false`
```

In case you have made a typo in class or method name - class name and method will be considered as 2 seprated callbacks, which do not exist either - and developer will see not relevant error message in the end.

rob006:
> for the configuration - this make no sense.

It make sense if you consider merging of arrays with config. For example, you could add one event in `common/config/main.php` and additional in `backend/config/main.php` - using array will save you repeating event config from `common/config/main.php`.

We could at least support arrays of closures.

rob006:
Related FR: #11269

hiqsol:
@klimov-paul configuration makes great sense - that's one of the best part of Yii ;)
Let us put in configuration as much as possible. I am sure it's a good practice.

@rob006 gets the idea - without this feature there is no way to merge configuration to get multiple handlers for an event.

dynasource:
> It make sense if you consider merging of arrays with config.

true. Experienced this problem a few times already.

klimov-paul:
> Let us put in configuration as much as possible. I am sure it's a good practice.

But still you are unable to resolve the example I have given to you https://github.com/yiisoft/yii2/pull/13199#issuecomment-266988010

The possible solution is usage of the another magic key rather then `on eventName` inside the configuration, which will **always** accept list of handlers. Any other change will be inconsistent or BC break.
All we need to do it determine, which key should it be.
It can be 'multi-on eventName' for example.

klimov-paul:
> It make sense if you consider merging of arrays with config. For example, you could add one event in common/config/main.php and additional in backend/config/main.php - using array will save you repeating event config from common/config/main.php.

General note: it is recommended to put event handlers into a separated behaviors. In such way no program logic will be exposed at the configuration level, remaining inclaplulated inside the behaviors.
Different behavior attachements can be easily merged together.

framework/base/Component.php
```diff
@@ -481,6 +482,21 @@ public function hasEventHandlers($name)
     public function on($name, $handler, $data = null, $append = true)
     {
         $this->ensureBehaviors();
+        if (is_array($handler) && !is_callable($handler)) {
```
SilverFire: Do we really need this check and additional `_on` method?

```php
if (!is_array($handler) || is_callable($handler)) {
    $handler = (array) $handler;
}

foreach (...)
```

SilverFire (changes requested):

SilverFire:
> I am against such change as it is risk prone and may lead to unexpected results as array-callbale is always hard to separate from array of callables.

That's true. Array-callable is hard to distinguish.

> For example:
> ```php
> var_dump(is_callable(['SomeClass', 'someMethod'])); // outputs `true`
> var_dump(is_callable(['UnexistingClass', 'someMethod'])); // outputs `false`
> ```

Good example. It shows a risk of unexpected behaviors. 

> It can be `multi-on eventName` for example.

Most probably this is the only safe and efficient solution. 

I'd like to hear @cebe input

hiqsol:
@klimov-paul Thanks for mentioning behaviors - that's suitable solution for my case.

> class name and method will be considered as 2 separated callbacks, which do not exist either - and  developer will see not relevant error message in the end.

that's not quite so - the error message will slightly change, but:

1. non working program will stay non working and will die in the same place
2. the error message will show the same stack trace and is quite readable:

was:
```
call_user_func() expects parameter 1 to be a valid callback, class 'my\namespace\TypoClass' not found
```

will become:
```
call_user_func() expects parameter 1 to be a valid callback, function 'my\namespace\TypoClass' not found or invalid function name
```

Unfortunately both messages are equally cryptic cause the stack trace doesn't show the place of actual typo.

framework/base/Component.php
Faryshta: how about

```php
if ($append || empty($this->_events[$name])) {
    if (is_array($handler) && !is_callable($handler)) {
        $this->_events[$name] = array_merge(
             $this->_events[$name],
             array_map(function ($singleHandler) use ($data) {
                 return [$singleHandler, $data];
             }, $handler)
        );
    } else {
        $this->_events[$name][] = [$handler, $data];
} else {
    // prepend logic, very similar
}
```
not even a `foreach` required

Faryshta:
what if i need different `$data` for each event?

SilverFire:
Considering that behaviors is a working solution and it provides more control, I'm closing this PR.

njasm:
moving/discussing in the PSR direction seems to be a good move, not only to this issues specifically but to the framework as a whole. 

I would like to see more discutions from the core developers regarding the possibility of introduction of interfaces for the app components as well, seeing the framework capable to be used as an infrastruture component instead of the "all app", allowing developers to segregate bits and parts a little bit more from the problem domain if wished..

does this make any sense? hope so..

Issue #10140, opened 2015-11-11
Incorrect ip when user uses proxy


mohammadhosain:
`yii\web\Request::getUserIP` uses `REMOTE_ADDR` to get user ip.The problem with this is, that `$_SERVER['REMOTE_ADDR']` is not always reliable. What if the user is behind a proxy, and the IP is shown by `HTTP_X_FORWARDED_FOR` header?

samdark:
Then you have to override the method in your own class and make sure your proxy isn't forwarding `X-Forwarded-For` HTTP header. Reading IP from `X-Forwarded-For` by default is a security flaw.

Issue #8746, opened 2015-06-11
ip address behind load balancer


nikunjtrapasiya:
function getUserIP() returns local ip address if the servers are behind load balancer (i am using aws)

```
    public function getUserIP()
    {
        return isset($_SERVER['REMOTE_ADDR']) ? $_SERVER['REMOTE_ADDR'] : null;
    }
```

i need to change it to each time i upgrade my framework

```
return (isset($_SERVER['HTTP_X_FORWARDED_FOR']) && $_SERVER['HTTP_X_FORWARDED_FOR']) ? $_SERVER['HTTP_X_FORWARDED_FOR'] : null;
```

is there any way i can override the function in my config file or somewhere so i can still use it for logging purpose.

samdark:
Yes.

1. Create your own class that extends from `\yii\web\Request`.
2. Override `getUserIP()`.
3. In the config, specify new class as `request` component.

cebe:
or overwrite `$_SERVER['REMOTE_ADDR']` in index.php

dizews:
hello, what do you think if we implement this functionality like in [symphony](https://github.com/symfony/symfony/blob/v2.2.0/src/Symfony/Component/HttpFoundation/Request.php#L708) ?

samdark:
Via separate method? Yes, that's OK.

dizews:
@samdark I suggest to reimplement ```getUserIP``` which should use trustedHeaders property of request object and make docs how it influence to security.

cebe:
related to #10854 and #13050

cebe:
closing as duplicate of #10854

PR #13211, opened 2016-12-15
Update Request.php - User IP enhancement

IhabY:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

when the application is deployed behind a load balancer, the load balancer might pass the user IP in 2 different headers
the suggested code will try to find any of these headers and assign it accordingly
please refer to https://en.wikipedia.org/wiki/X-Forwarded-For

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -919,17 +919,17 @@ class Request extends \yii\base\Request
      */
     public function getUserIP()
     {
-        $_potentialReverseProxyIP = null;
+        $potentialReverseProxyIP = null;
 
         if (isset ($_SERVER['HTTP_X_REAL_FORWARDED_FOR']) && $_SERVER['HTTP_X_REAL_FORWARDED_FOR'] != '') {
-            $_potentialReverseProxyIP = $_SERVER['HTTP_X_REAL_FORWARDED_FOR'];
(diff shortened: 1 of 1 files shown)


klimov-paul:
https://github.com/yiisoft/yii2/issues/10140#issuecomment-155831311

ihaby:
@klimov-paul I hope this is still open for discussion;
Keeping the original REMOTE ADDRESS, has multiple implications such as:
- Yii2 debugger will log an incorrect IP value
- all Yii2 logs will stamp an incorrect IP value
- we are currently using Yii2 rate limiter in hand with this change (with other parameters) to block certain sections on our application and I would prefer to use out of the box functionalities whenever possible

klimov-paul:
Such change is a security breech.

ihaby:
I am willing to add it as a separate column and function, would this approach be acceptable?

klimov-paul:
However, it can be made optional: we can add some flag option, which will enable usage of `x-forwarded-for` header.

rob006:
#8746 
https://github.com/yiisoft/yii2/issues/10854#issuecomment-184246966

samdark:
The issue will be solved but not with this pull request. Thanks for contributing and sorry that it's not merged.

PR #13213, opened 2016-12-15
Fixes #11906: Updated `yii\widgets\MaskedInput` inputmask dependency to `~3.3.3`

### LINKED ISSUE #11906: Update inputmask dependency version
`InputMask` is now on minor version 3 with lots of fixes.

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | #11906

### DIFF
--- a/framework/composer.json
+++ b/framework/composer.json
@@ -65,7 +65,7 @@
         "ezyang/htmlpurifier": "~4.6",
         "cebe/markdown": "~1.0.0 | ~1.1.0",
         "bower-asset/jquery": "2.2.*@stable | 2.1.*@stable | 1.11.*@stable | 1.12.*@stable",
-        "bower-asset/jquery.inputmask": "~3.3.3",
+        "bower-asset/jquery.inputmask": "~3.2.2 | ~3.3.3",
         "bower-asset/punycode": "1.3.*",
         "bower-asset/yii2-pjax": "~2.0.1"
     },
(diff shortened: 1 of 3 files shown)


dynasource (approved):

framework/composer.json
```diff
@@ -65,7 +65,7 @@
         "ezyang/htmlpurifier": "~4.6",
         "cebe/markdown": "~1.0.0 | ~1.1.0",
         "bower-asset/jquery": "2.2.*@stable | 2.1.*@stable | 1.11.*@stable | 1.12.*@stable",
-        "bower-asset/jquery.inputmask": "~3.2.2",
+        "bower-asset/jquery.inputmask": "~3.3.3",
```
cebe: should be `~3.3.3 | ~3.2.2` instead to allow upgrading yii without the need to upgrade the inputmask lib.

cebe (changes requested):

framework/composer.json
samdark: Is there a reason not to upgrade inputmask lib?

cebe: I tend to review the changes of my dependencies to make sure nothing breaks.
A `composer update yiisoft/yii2` will fail and not update to 2.0.11 with this requirement because it will only update yiisoft/yii2 and no other packages. It should be possible to upgrade yii without touching any other packges. The only exception I see here are security issues that force the need of bumping a dependency version.

cebe (approved):

PR #13214, opened 2016-12-15
Xml formatter root and object tags

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12419


framework/web/XmlResponseFormatter.php
```diff
@@ -37,7 +37,7 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      */
     public $encoding;
     /**
-     * @var string the name of the root element.
+     * @var string the name of the root element or null if no root tag should be added.
```
cebe: in Html helper we use false for no tag. In general we could also allow it to be empty(). as `<>` would not be a valid xml tag anyway.

framework/web/XmlResponseFormatter.php
```diff
@@ -50,7 +50,11 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      * @since 2.0.7
      */
     public $useTraversableAsArray = true;
-
+    /**
+     * @var bool if object tags should be added
+     * @since 2.0.10
```
cebe: 2.0.11.

cebe (changes requested):

cebe (approved):

SilverFire (approved):

framework/web/XmlResponseFormatter.php
```diff
@@ -50,7 +50,11 @@ class XmlResponseFormatter extends Component implements ResponseFormatterInterfa
      * @since 2.0.7
      */
     public $useTraversableAsArray = true;
-
+    /**
+     * @var bool if object tags should be added
+     * @since 2.0.11
+     */
+    public $objectTags = true;
```
klimov-paul: `objectTags` does not sound appropriate for a boolean field.
Possible names could be:
- `createObjectTags`
- `useObjectTags`
- `addObjectTags`

samdark: Yes. `useObjectTags` sounds better...

PR #13216, opened 2016-12-16
fix ArrayHelper::getValue() to throw exception on invalid input

### LINKED ISSUE #12927: ArrayHelper::getValue does not throw as documented
```PHP
var_dump(\yii\helpers\ArrayHelper::getValue('test', 'someProp', 123));
```
`InvalidParamException` thrown.
### What do you get instead?
No exception, output: 123.

Yii version: 2.0.10

```PHP
    /**
     * Retrieves the value of an array element or object property with the given key or property name.
     * ...
     * @param array|object $array array or object to extract value from
     * ...
     * @throws InvalidParamException if $array is neither an array nor an object.
     */
    public static function getValue($array, $key, $default = null)
```

cebe:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | only if people program against the documentation. result was kind of undefined. See #12927 
| Tests pass?   | yes
| Fixed issues  | #12927

fixes #12927

### DIFF
--- a/framework/helpers/BaseArrayHelper.php
+++ b/framework/helpers/BaseArrayHelper.php
@@ -178,10 +178,15 @@ class BaseArrayHelper
      * The possibility to pass an array of keys is available since version 2.0.4.
      * @param mixed $default the default value to be returned if the specified array key does not exist. Not used when
      * getting value from an object.
-     * @return mixed the value of the element if found, default value otherwise
+     * @return mixed the value of the element if found, default value otherwise.
+     * @throws InvalidParamException if $array is neither an array nor an object.
      */
     public static function getValue($array, $key, $default = null)
     {
+        if (!is_array($array) && !is_object($array)) {
+            throw new InvalidParamException('Argument passed to getValue() must be an array or object.');
+        }
+
         if ($key instanceof \Closure) {
             return $key($array, $default);
         }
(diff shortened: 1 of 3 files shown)


framework/helpers/BaseArrayHelper.php
```diff
@@ -178,10 +178,15 @@ public static function merge($a, $b)
      * The possibility to pass an array of keys is available since version 2.0.4.
      * @param mixed $default the default value to be returned if the specified array key does not exist. Not used when
      * getting value from an object.
-     * @return mixed the value of the element if found, default value otherwise
+     * @return mixed the value of the element if found, default value otherwise.
+     * @throws InvalidParamException if $array is neither an array nor an object.
```
cebe: this line existed in 2.0.10, so that is not new. https://github.com/yiisoft/yii2/blob/2.0.10/framework/helpers/BaseArrayHelper.php#L182

samdark (approved):

dynasource (approved):
well done

SilverFire (approved):

SilverFire:
Reverted: ed82ba4

See https://github.com/yiisoft/yii2/issues/13248 for details

PR #13217, opened 2016-12-16
Catch `\Throwable` in critical places

### LINKED ISSUE #12619: catch Throwable in ErrorHandler::handleException

cebe:
Added catch `\Throwable` to be compatible with PHP7.
Added it in cases where object state needs to be kept consistent.

Mainly on transactions but also some other places where some values are
reset before exiting.

Most of them could probably be refactored by using `finally` in 2.1, as
that requires PHP 5.5.

fixes #12619

### DIFF
--- a/docs/guide/db-dao.md
+++ b/docs/guide/db-dao.md
@@ -336,7 +336,7 @@ try {
     
     $transaction->commit();
     
-} catch(\Exception $e) {
+} catch(\Exception $e) { // replace \Exception with \Throwable when you are using PHP 7
 
     $transaction->rollBack();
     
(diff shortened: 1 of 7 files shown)


samdark (approved):

docs/guide/db-dao.md
```diff
@@ -336,7 +336,7 @@ try {
     
     $transaction->commit();
     
-} catch(\Exception $e) {
+} catch(\Exception $e) { // replace \Exception with \Throwable when you are using PHP 7
```
klimov-paul: Shouldn't this contain example with extra `catch (\Throwable $e) {` in addition to `\Exception` as it is done inside the core code?

```php
} catch (\Exception $e) { // can be ommited when you are using PHP 7
    ...
} catch (\Throwable $e) {
   ...
}
```

klimov-paul (changes requested):

docs/guide/db-dao.md
samdark: No. If your project is PHP 7 only, there's no point in catching `\Exception`.

samdark: Probably yes. For portability reasons.

cebe: done.

klimov-paul (approved):

SilverFire (approved):

Issue #13224, opened 2016-12-16
Unexpected behavior for clearing cache with ApcCache 


alexxxst:
### What steps will reproduce the problem?

Use "yii\caching\ApcCache" for application (web and console) cacher (APC or APCu).
Use your web-app to put something in cache.
Try to flush cache from console-app.
Check your web-app for cache data exists. It exists.

### What is the expected result?

Expected clear cache, like "yii\caching\FileCache", but I can't flush fastcgi APC(u) cache from a php cli script (console-app) since the cli script will run from a different process as fastcgi.

### What do you get instead?

Probably some way to run cache flush via web-app from console-app?
Any ideas?

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 7.0.14 etc.
| Operating system | Linux / MacOS

rob006:
Duplicates #10503

rob006:
@samdark Maybe `CacheController` should display some notice when user tries flush `ApcCache`?

samdark:
https://github.com/yiisoft/yii2/pull/13226

PR #13226, opened 2016-12-16
Improved cache command output in APC case

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13224


rob006:
Is it possible to run console command in webapp process?

samdark:
Yes.

rob006:
So this PR will make it impossible to flush APC cache by wrapping console command in webapp controller. I suggest flush cache anyway and only display notice, that it will not affect different processes, so it is impossible co flush webapp cache from console.

cebe:
Yes, I think flush should still be called but it should note that the APC you wanted to flash may be in a different process.

cebe (approved):

Issue #12528, opened 2016-09-13
Database logging causes bad performance for large queries


kidol:
Problem is `yii\db\Command:execute()` and related methods always call `getRawSql()` to log the query: https://github.com/yiisoft/yii2/blob/master/framework/db/Command.php#L830
Of course, `Yii::info()` is always called, even if there is no related log target.

This is for just 3.000 large rows inserted (in one query), adds a 9 second overhead. :boom: 

Before:

![rawsql1](https://cloud.githubusercontent.com/assets/440902/18464291/665fbf1a-7990-11e6-8af1-7e5637cd614e.png)

After commenting out `$this->getRawSql()` in `Command::execute()`:

![rawsql2](https://cloud.githubusercontent.com/assets/440902/18464324/a43def96-7990-11e6-91d0-024a81fbb802.png)

Possible solution is to add `$logQueries` to `Connection` class. Setting it to `false` would skip all calls to `Yii::info()` as well as `getRawSql()` within `Command` class.

SilverFire:
Most probably disk IO takes that overhead time.

Do you have YII_DEBUG enabled?
Did you try to create a `/dev/null` handler for `yii\db\Command::execute` message category?

kidol:
No it's only about the call to `getRawSql()`. I don't even have a log target for db.

You can see in stats above, 2.370 seconds are spend within `getRawSql()` itself, the rest (~6.4 seconds) are spend in the methods it's calling.

klimov-paul:
Method `Command::getRawSql()` may cause a perfromance impact  since it parses parameters in the SQL and replaces it with thier values. These string processing may take significant time to be performed.
However, it is still unlikely it will take several seconds to do so.

klimov-paul:
In your timeline it seems the method `Connection::getSlavePdo()` causes the time consumption.
It is used inside `Schema::quoteValue()`, which is used at `Command::getRawSql()`.

kidol:
I have looked at `getSlavePdo()` before and noticed nothing unusual in the way it works. It's expected I guess because it's called 24k times - unnecessarily in this case. Also I have no slave configured.

>However, it is still unlikely it will take several seconds to do so.

9,000 milliseconds / 3,000 rows = 3 ms per row. These are rather large rows (causing big sql string and 3000 * 12 params `getRawSql()` has to crunch through).

klimov-paul:
Are you able to compose standalone performance test for this matter?

kidol:
[...]

PR #13236, opened 2016-12-17
Fixed excess escaping in Command::batchInsert()

### LINKED ISSUE #11242: batchInsert incorrectly quotes SQL query with values, containing {{ }} symbols

`$this->db->createCommand()->batchInsert('table', ['column1'], [["input-{{value}}"]])->execute();`

Value inserted into a table
### What do you get instead?

[yii\db\Exception] SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax
The SQL being executed was: `INSERT INTO "table" ("column1") VALUES ('input-"value"')`
### Additional info

The problem is that `Command::setSql` method quotes already prepared batch insert query with `'/(\\{\\{(%?[\w\-\. ]+%?)\\}\\}|\\[\\[([\w\-\. ]+)\\]\\])/'` regexp inside `Connection::quoteSql` which does not know anything about the context 

| Q | A |
| --- | --- |
| Yii version | Yii v2.0.7 |
| PHP version | PHP 5.6.16 (cli) (built: Nov 27 2015 10:28:34) |
| Operating system | Any |

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no. Read bellow
| Tests pass?   | hope so :)
| Fixed issues  | #11242

Fixes #11242

There is a test that contain a comment about a potentially wrong behavior. This behavior was NOT introduced in this PR, I've just covered it with a unit test.


samdark:
Seems to fail important tests.

SilverFire:
Fixed in master

kidol:
There is name clash with `getRawSql()` method.

Also I think it would be better to first re-evaluated whether prepared statements can be used in regards to #12528

samdark:
#12528 is now fixed. @SilverFire do you remember what's missing in this one?

SilverFire:
> I think it would be better to first re-evaluated whether prepared statements can be used

Prepared statements are limited on number of placeholders, for instance MySQL has limit of 65535. Probably, we can calculate number of placeholders, split input data in buckets not to step over the limit and run multiple queries in transaction, but I think this way is neither easier nor better.



> do you remember what's missing in this one?

I think I would like to see @cebe input. I will resolve conflicts and prepare for merge.

framework/db/Command.php
```diff
@@ -137,19 +137,43 @@ public function getSql()
...
     public function setSql($sql)
     {
         if ($sql !== $this->_sql) {
             $this->cancel();
+            $this->reset();
             $this->_sql = $this->db->quoteSql($sql);
-            $this->_pendingParams = [];
-            $this->params = [];
-            $this->_refreshTableName = null;
+        }
```
tom--: `$this->_sql` is the quoted version of `$sql`. So the condition `if ($sql !== $this->_sql)` doesn't test for changes.

Is this test really required?

framework/db/Command.php
```diff
@@ -137,19 +137,43 @@ public function getSql()
...
+     * @see reset()
+     * @see cancel()
+     */
+    public function setRawSql($sql)
+    {
+        if ($sql !== $this->_sql) {
+            $this->cancel();
+            $this->reset();
+            $this->_sql = $sql;
         }
```
tom--: In this case the test is correct but all of the operations in the `if` branch (cancel, reset, assign) are very fast. Why not do these steps unconditionally?

framework/db/Command.php
```diff
@@ -457,9 +481,16 @@ public function insert($table, $columns)
...
     {
+        $table = $this->db->quoteSql($table);
+        $columns = array_map(function ($column) {
+            return $this->db->quoteSql($column);
+        }, $columns);
+
         $sql = $this->db->getQueryBuilder()->batchInsert($table, $columns, $rows);
 
-        return $this->setSql($sql);
+        $this->setRawSql($sql);
```
tom--: The intermediate variable `$sql` does little work. But its value could be a very long string. Unless the PHP compiler does The Right Thing then it's fine. Otherwise it `$sql` could be eliminated.

tom-- (commented):
Did you (or anyone) verify that the new tests fail in .12?

framework/db/Command.php
tom--: I'm pretty sure it makes no difference so whatever style you prefer.

framework/db/Command.php
cebe: it probably isn't as this check has been introduced before `quoteSql` has been created. This does not belong to this PR however. Changing it could probably break code in very rare cases.

framework/db/Command.php
```diff
@@ -137,19 +137,43 @@ public function getSql()
...
+    }
+
+    /**
+     * Specifies the SQL statement to be executed. The SQL statement will not be modified in any way.
+     * The previous SQL execution (if any) will be cancelled, and [[params]] will be cleared as well. See [[reset()]]
+     * for details.
+     *
+     * @param string $sql the SQL statement to be set.
+     * @return $this
+     * @since 2.0.11
```
cebe: 2.0.13

framework/db/Command.php
```diff
@@ -137,19 +137,43 @@ public function getSql()
...
+        return $this;
+    }
+
+    /**
+     * Specifies the SQL statement to be executed. The SQL statement will not be modified in any way.
+     * The previous SQL execution (if any) will be cancelled, and [[params]] will be cleared as well. See [[reset()]]
+     * for details.
+     *
+     * @param string $sql the SQL statement to be set.
+     * @return $this
```
cebe: `@return $this this command instance`

framework/db/Command.php
cebe: there is no point in testing `$sql !== $this->_sql` here, I am for removing it.

framework/db/Command.php
```diff
@@ -945,4 +976,18 @@ protected function refreshTableSchema()
             $this->db->getSchema()->refreshTableSchema($this->_refreshTableName);
         }
     }
+
+    /**
+     * Resets [[sql]] and [[params]] properties.
+     *
+     * @void
+     * @since 2.0.11
```
cebe: remove `@void`
`@since 2.0.13`

framework/CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Yii Framework 2 Change Log
 - Bug #13159: Fixed `destroy` method in `yii.captcha.js` which did not work as expected (arogachev)
 - Bug #13198: Fixed order of checks in `yii\validators\IpValidator` that sometimes caused wrong error message (silverfire)
 - Bug #13200: Creating Urls for routes specified in `yii\rest\UrlRule::$extraPatterns` did not work if no HTTP verb was specified (cebe)
+- Bug #11242: Fixed excess escaping in `yii\db\Command::batchInsert()` (silverfire)
```
cebe: rebase and move the changelog entry.

cebe (changes requested):
minor change required. solution looks good to me.

framework/CHANGELOG.md
```diff
@@ -271,6 +272,8 @@ Yii Framework 2 Change Log
 - Bug #13128: Fixed incorrect position of {pos} string in ColumnSchemaBuilder `__toString` (df2)
 - Bug #13159: Fixed `destroy` method in `yii.captcha.js` which did not work as expected (arogachev)
 - Bug #13198: Fixed order of checks in `yii\validators\IpValidator` that sometimes caused wrong error message (silverfire)
+- Bug #13200: Creating Urls for routes specified in `yii\rest\UrlRule::$extraPatterns` did not work if no HTTP verb was specified (cebe)
```
samdark: I don't think these should be here.

framework/db/Command.php
```diff
@@ -137,19 +137,43 @@ public function getSql()
     }
 
     /**
-     * Specifies the SQL statement to be executed.
-     * The previous SQL execution (if any) will be cancelled, and [[params]] will be cleared as well.
+     * Specifies the SQL statement to be executed. The SQL statement will be quoted using [[Connection::quoteSql()]].
+     * The previous SQL execution (if any) will be cancelled, and [[params]] will be cleared as well. See [[reset()]]
```
samdark: `The previous SQL execution (if any) will be cancelled` → `The previous SQL (if any) will be discarded`

framework/db/Command.php
```diff
@@ -137,19 +137,43 @@ public function getSql()
...
-            $this->params = [];
-            $this->_refreshTableName = null;
+        }
+
+        return $this;
+    }
+
+    /**
+     * Specifies the SQL statement to be executed. The SQL statement will not be modified in any way.
+     * The previous SQL execution (if any) will be cancelled, and [[params]] will be cleared as well. See [[reset()]]
```
samdark: `The previous SQL execution (if any) will be cancelled` → `The previous SQL (if any) will be discarded`

samdark (approved):
Good to merge except messed up changelog and a few phpdoc corrections.

framework/CHANGELOG.md
SilverFire: It's not the first time when `git merge` messes up changelog :(

sokrat:
This changes has broken ```batchInsert()``` with table name contains special characters.
https://github.com/yiisoft/yii2/blob/75dbdb008363b90d4ffc8f502022763aedd6998a/framework/db/Command.php#L488
```
var_dump(Yii::$app->db->quoteSql('{{%test}}'));
""test""
```

cebe:
@sokrat please open a new issue and give more details about the problem.

PR #13238, opened 2016-12-17
Fix tests on travis

rob006:
This should fix travis tests broken after https://github.com/yiisoft/yii2/commit/66e815871609a87f01b3a396fd974a466dafdd5e


samdark:
Thank you!

PR #13246, opened 2016-12-19
Fixed comments in travis.yml [skip ci]

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

- Aligned comments with YAML node items as it's done for other comments (for consistency).
- Removed commented YAML code because we can always restore it from Git history if needed.


SilverFire:
Thank you

PR #13250, opened 2016-12-19
improved HATEOAS example

githubjeka:
IMHO:
Not the best choice for ActiveRecord to represent a REST resource with HATEOAS  links and other meta information. That why AR replaced to DTO. 

And added some others links.

### DIFF
--- a/docs/guide/rest-resources.md
+++ b/docs/guide/rest-resources.md
@@ -146,7 +146,7 @@ contains a single method [[yii\web\Linkable::getLinks()|getLinks()]] which shoul
 Typically, you should return at least the `self` link representing the URL to the resource object itself. For example,
 
 ```php
-use yii\db\ActiveRecord;
+use \yii\base\Model;
 use yii\web\Link; // represents a link object as defined in JSON Hypermedia API Language.
 use yii\web\Linkable;
 use yii\helpers\Url;
--- a/docs/guide/rest-resources.md
+++ b/docs/guide/rest-resources.md
@@ -146,7 +146,7 @@ contains a single method [[yii\web\Linkable::getLinks()|getLinks()]] which shoul
 Typically, you should return at least the `self` link representing the URL to the resource object itself. For example,
 
 ```php
-use \yii\base\Model;
+use yii\base\Model;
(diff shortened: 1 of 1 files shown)


SilverFire (approved):

docs/guide/rest-resources.md
```diff
@@ -147,22 +147,40 @@ Typically, you should return at least the `self` link representing the URL to th
 
 ```php
 use yii\db\ActiveRecord;
-use yii\web\Link;
+use yii\web\Link; // represents a link object as defined in JSON Hypermedia API Language.
 use yii\web\Linkable;
 use yii\helpers\Url;
 
-class User extends ActiveRecord implements Linkable
+class UserResource extends Model implements Linkable
```
cebe: why not use active record? also if not using it, the import above should be changed.

githubjeka: It was night and I wanted to sleep. And today when I wake up I got the idea, "I did not replace import" :)

For me ActiveRecord better use for common case, not for this.

githubjeka: Because AR for work with DB only.
If use Model in your REST-layer application you can work with simple object like DTO, which have only the necessary data: datum, links and other meta information. API for this object have only methods for work REST logic. Will not be here api for DB. All of this will keep the code more healthy.

Those who will read the documentation, must learn to do things right from the beginning. They will have a bunch of places in their project where they screwed up.

This my opinion. 

If you create simple project OK use Active Record. But your documentation should be focused on certain principles: "You write doc for beginner user" or "you write for beautiful word SOLID."

If 1 - Pay more attention to documentation than writing new code. At least the article [Resource](http://www.yiiframework.com/doc-2.0/guide-rest-resources.html#resources) should be rewrite - because in head you write about Model:
> In this section, we will mainly describe how a resource class extending from yii\base\Model (or its child classes) can specify what data may be returned via RESTful APIs. If the resource class does not extend from yii\base\Model, then all its public member variables will be returned.

If 2 - let's follow the principles of SOLID.

Thank anyway for your hard work.

samdark (approved):

samdark:
Thanks!

PR #13258, opened 2016-12-20
WIP Fix race conditions in FileMutex

kidol:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #12860

Not tested under unix yet. Looking for comments/review.

Also noticed a bug: `FileMutex` didn't callled `init()` of parent causing `$autoRelease` to have no effect.


kidol:
I want to remove `$autoRelease` because at least for `FileMutex` and `MysqlMutex` it never made sense because native lock will be auto-released anyway when script/mysql-session ends. I guess other dbms handle it the same.

Any objections?

SilverFire:
> because native lock will be auto-released anyway when script/mysql-session ends

Are you sure about this?

http://stackoverflow.com/questions/13522273/will-flocked-file-be-unlocked-when-the-process-die-unexpectedly

kidol:
Yes I'm sure (just verified once more on win/unix with 5.6.*), not sure what they are talking on stackoverflow.

Also current FileMutex implementation supports local locks only: https://github.com/yiisoft/yii2/blob/master/framework/mutex/FileMutex.php#L118

```php
if (!isset($this->_files[$name]) ...
```

..means before you can release a file lock, you must acquire it first (locally).

So the removal of `$autoRelease` would make behavior consistent, even if some backend may allow persistent locks: It's just not the common use case.

framework/CHANGELOG.md
```diff
@@ -24,9 +24,11 @@ Yii Framework 2 Change Log
 - Bug #12828: Fixed handling of nested arrays, objects in `\yii\grid\GridView::guessColumns` (githubjeka)
 - Bug #12836: Fixed `yii\widgets\GridView::filterUrl` to not ignore `#` part of filter URL (cebe, arogachev)
 - Bug #12856: Fixed `yii\web\XmlResponseFormatter` to use `true` and `false` to represent booleans (samdark)
+- Bug #12860: Fixed possible race conditions in `yii\mutex\FileMutex` (kidol)
 - Bug #12879: Console progress bar was not working properly in Windows terminals (samdark, kids-return)
 - Bug #12880: Fixed `yii\behaviors\AttributeTypecastBehavior` marks attributes with `null` value as 'dirty' (klimov-paul)
 - Bug #12904: Fixed lowercase table name in migrations (zlakomanoff)
+- Bug #12927: `ArrayHelper::getValue()` did not throw exception if the input was neither an object nor an array, even though it was documented (cebe)
```
samdark: Seems unrelated.

framework/CHANGELOG.md
```diff
@@ -37,10 +39,10 @@ Yii Framework 2 Change Log
 - Bug #13159: Fixed `destroy` method in `yii.captcha.js` which did not work as expected (arogachev)
 - Bug #13198: Fixed order of checks in `yii\validators\IpValidator` that sometimes caused wrong error message (silverfire)
 - Bug #13200: Creating URLs for routes specified in `yii\rest\UrlRule::$extraPatterns` did not work if no HTTP verb was specified (cebe)
-- Bug #13229: Fix fetching table schema for `pgsql` when `PDO::ATTR_CASE` is set (klimov-paul)
```
samdark: Why was it removed?

samdark (changes requested):

rob006:
Why are we complicating this so much instead of creating garbage collector which will periodically delete old lock files? We don't need delete lock files immediately, we just need to avoid storing them forever.

kidol:
Well with garbage collector you shift problem somewhere else: GC would have to check whether the file is locked before removing it. On windows it would be enough to just try `@unlink($oldLockFile)`, but on unix this will not work (because it will also remove currently opened/locked files).

samdark:
GC is more complicated IMO.

rob006:
> Well with garbage collector you shift problem somewhere else: GC would have to check whether the file is locked before removing it.

And what is the problem here? We can just try to acquire lock on this file and skip deleting if it is already locked. 

With GC we have simpler (and faster) code for `acquireLock()` and `releaseLock()`. As an end user I have also a possibility to configure GC to be more safe than this implementation (for example by disabling it if I don't generate lock names and use only a couple of hardcoded IDs).

Also, are you sure that checking inode nr is reliable? Inode nr could be reused:

```console
robert@robert-laptop:~/Poligon/inode$ touch test
robert@robert-laptop:~/Poligon/inode$ ls -i
10625010 test
robert@robert-laptop:~/Poligon/inode$ rm test 
robert@robert-laptop:~/Poligon/inode$ ls -i
robert@robert-laptop:~/Poligon/inode$ touch test
robert@robert-laptop:~/Poligon/inode$ ls -i
10625010 test
robert@robert-laptop:~/Poligon/inode$ touch test2
robert@robert-laptop:~/Poligon/inode$ ls -i
10625010 test  10625011 test2
robert@robert-laptop:~/Poligon/inode$ rm test 
robert@robert-laptop:~/Poligon/inode$ rm test2
robert@robert-laptop:~/Poligon/inode$ touch test2
robert@robert-laptop:~/Poligon/inode$ touch test
robert@robert-laptop:~/Poligon/inode$ ls -i
10625011 test  10625010 test2
robert@robert-laptop:~/Poligon/inode$ rm test2
robert@robert-laptop:~/Poligon/inode$ rm test
robert@robert-laptop:~/Poligon/inode$ touch testing
robert@robert-laptop:~/Poligon/inode$ ls -i
10625010 testing
```

kidol:
>With GC we have simpler (and faster) code for acquireLock() and releaseLock()

Well it's mainly about the inode check? I think it's not significant.

>Inode nr could be reused

It doesn't matter. This is relevant line:

```php
fstat($file)['ino'] !== @fileinode($filePath)
```
If both inodes are the same (reused or not), then everything is ok, that is: the acquired file pointer actually belongs to the physical, existing file.

>As an end user I have also a possibility to configure GC to be more safe than this implementation (for example by disabling it if I don't generate lock names and use only a couple of hardcoded IDs).

I don't understand this.

In general I don't really get why you prefer GC here. If there is a way to properly remove any possible race-condition in `release()` then why not do it? Also the code is not overly complex or something.

Will continue on PR soon.

framework/mutex/FileMutex.php
```diff
@@ -115,15 +141,26 @@ protected function acquireLock($name, $timeout = 0)
      */
     protected function releaseLock($name)
     {
-        if (!isset($this->_files[$name]) || !flock($this->_files[$name], LOCK_UN)) {
+        if (!isset($this->_files[$name])) {
             return false;
-        } else {
+        }
+        
+        if (DIRECTORY_SEPARATOR === '\\') {
```
dynasource: This could be improved on readibility. The code itself should make most of the comments obsolete. I also prefer reuse of code to something like:
```php
$isWindows = DIRECTORY_SEPARATOR === '\\';
if($isWindows ){
  @unlink($this->getLockFilePath($name));
}
flock($this->_files[$name], LOCK_UN);
fclose($this->_files[$name]);
if(!$isWindows ){
  unlink($this->getLockFilePath($name));
}
```

framework/mutex/FileMutex.php
```diff
@@ -86,26 +87,51 @@ public function init()
...
+                if (++$waitTime > $timeout) {
+                    return false;
+                }
+                
+                sleep(1);
+                continue;
             }
-            sleep(1);
+            
+            // Under unix we delete the lock file before releasing the related handle. Thus it's possible that we've acquired a lock on
```
dynasource: deserves a separate method to make it more testable

dynasource (changes requested):

rob006:
Just for record: I have been using this solution for about a month without any problems. 👍 from me.

SilverFire:
@rob006 is everything OK for you with this fix?

SilverFire:
Additionally checked logic. Seems to be correct :)
I don't think we can write a good unit test to cover this race condition because race condition simulation is really hard. [And it's not only mine opinion](https://softwareengineering.stackexchange.com/questions/196105/testing-multi-threaded-race-conditions).

Merging as is. Thank you for a good investigation, @kidol!

PR #13260, opened 2016-12-20
Add support \yii\db\Expression attribute to \yii\data\Sort

lav45:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

If you want to sort the data in an unusual way we usually use the class `\yii\db\Expression` as the wrapper of our expression.
```php
Product::find()
    ->orderBy(new Expression("[[price]] ASC NULLS LAST"))
   ->all();
```

But this option cannot be used when sorting settings `\yii\data\ActiveDataProvider`

I propose to extend the functionality of the class `\yii\data\Sort` to preserve the possibility to sort by an arbitrary expression.
```php
    'query' => Product::find(),
]);
$dataProvider->sort->attributes['price'] = [
    'asc' => [new Expression("[[price]] ASC NULLS LAST")],
    'desc' => [new Expression("[[price]] DESC NULLS LAST")],
];
```

### DIFF
--- a/framework/data/Sort.php
+++ b/framework/data/Sort.php
@@ -107,8 +107,8 @@ class Sort extends Object
      * [
      *     'age',
      *     'name' => [
-     *         'asc' => ['[[last_name]] ASC NULLS FIRST'],
-     *         'desc' => ['[[last_name]] DESC NULLS LAST'],
+     *         'asc' => [new \yii\db\Expression('[[last_name]] ASC NULLS FIRST')],
+     *         'desc' => [new \yii\db\Expression('[[last_name]] DESC NULLS LAST')],
      *         'default' => SORT_DESC,
      *     ],
      * ]
--- a/framework/data/Sort.php
+++ b/framework/data/Sort.php
@@ -100,6 +100,20 @@ class Sort extends Object
      * ]
      * ```
      *
+     * If you are using a PostgreSQL database, you may need to specify the sorting arbitrary expression.
+     * This syntax can be described using the following example:
+     *
+     * ```php
+     * [
+     *     'age',
+     *     'name' => [
+     *         'asc' => ['[[last_name]] ASC NULLS FIRST'],
+     *         'desc' => ['[[last_name]] DESC NULLS LAST'],
+     *         'default' => SORT_DESC,
+     *     ],
+     * ]
+     * ```
+     *
      * In the above, two attributes are declared: `age` and `name`. The `age` attribute is
      * a simple attribute which is equivalent to the following:
      *
@@ -216,7 +230,11 @@ class Sort extends Object
             $definition = $this->attributes[$attribute];
             $columns = $definition[$direction === SORT_ASC ? 'asc' : 'desc'];
             foreach ($columns as $name => $dir) {
-                $orders[$name] = $dir;
+                if (is_int($name)) {
+                    $orders[] = $dir;
+                } else {
+                    $orders[$name] = $dir;
+                }
             }
         }
 
--- a/tests/framework/data/SortTest.php
+++ b/tests/framework/data/SortTest.php
@@ -7,6 +7,7 @@
 
 namespace yiiunit\framework\data;
 
+use yii\db\Expression;
 use yii\web\UrlManager;
 use yiiunit\TestCase;
 use yii\data\Sort;
@@ -58,8 +59,8 @@ class SortTest extends TestCase
         $sort = new Sort([
             'attributes' => [
                 'name' => [
-                    'asc' => ['[[last_name]] ASC NULLS FIRST'],
-                    'desc' => ['[[last_name]] DESC NULLS LAST'],
+                    'asc' => [new Expression('[[last_name]] ASC NULLS FIRST')],
+                    'desc' => [new Expression('[[last_name]] DESC NULLS LAST')],
                 ],
             ],
         ]);
--- a/tests/framework/data/SortTest.php
+++ b/tests/framework/data/SortTest.php
@@ -53,6 +53,28 @@ class SortTest extends TestCase
         $this->assertEquals(SORT_ASC, $orders['age']);
     }
 
+    public function testGetExpressionOrders()
+    {
+        $sort = new Sort([
+            'attributes' => [
+                'name' => [
+                    'asc' => ['[[last_name]] ASC NULLS FIRST'],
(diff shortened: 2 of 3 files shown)


SilverFire:
I don't think that `yii\data\Sort` class should know something about DB-related things, such as `yii\db\Expression`. Otherwise we will encourage developers to write plain SQL in views :)

Maybe it makes sense to introduce a mapping method in `yii\db\ActiveQuery`, such as:

```php
public function mapOrders()
{
    return [
        'active_with_value' => new Expression('[[is_active]] = TRUE AND [[value]] IS NOT NULL'),
    ];
}
```

LAV45:
I immediately thought the same thing. 
I would not want to add another addiction to the \yii\data\Sort , but I decided to follow the example of [\yii\helpers\BaseJson::processData()](https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseJson.php#L145)

```php
class BaseJson
{
    protected static function processData($data, &$expressions, $expPrefix)
    {
// ...
            if ($data instanceof JsExpression) {
                $token = "!{[$expPrefix=" . count($expressions) . ']}!';
                $expressions['"' . $token . '"'] = $data->expression;
                return $token;
            } elseif ($data instanceof \JsonSerializable) {
                return static::processData($data->jsonSerialize(), $expressions, $expPrefix);
            } elseif ($data instanceof Arrayable) {
                $data = $data->toArray();
            } elseif ($data instanceof \SimpleXMLElement) {
                $data = (array) $data;
            } else {
// ...
    }
}
```

In my opinion there is nothing wrong with handling different data types.

LAV45:
@SilverFire , I would suggest to use this option
```php
public function mapOrders()
{
    return [
        'is_active' => [
            SORT_ASC => new Expression('[[is_active]] = TRUE AND [[value]] IS NOT NULL'),
            SORT_DESC => new Expression('[[is_active]] = FALSE OR [[value]] IS NULL'),
        ]
    ];
}
```
then it can be used
```php
Product::find()
    ->orderBy(['is_active' => SORT_DESC])
```

In this case, if the key is the same as the attribute in the table then it will use the version of `yii\db\ActiveQuery::mapOrders()`

SilverFire:
Looks good for me. Let's wait for more opinions from @yiisoft/core-developers

klimov-paul:
I can't see any issue here. `Sort` already supports SQL expressions as an attribute.
Here is a working example:

```
$nameSortExpression = 'CASE ([[title]] IS NULL) WHEN true THEN [[name]] ELSE [[title]] END';
            'query' => $query,
            'sort' => [
                'attributes' => [
                    'name' => [
                        'asc' => [
                            $nameSortExpression => SORT_ASC
                        ],
                        'desc' => [
                            $nameSortExpression => SORT_ASC,
                        ],
                    ],
                ],
                'defaultOrder' => [
                    'name' => SORT_ASC,
                ],
            ],
        ]);
```

LAV45:
@klimov-paul ,
If we follow your example then builds the query
```sql
... ORDER BY "manufacturer"."data"->'nntdays_units' DESC NULLS LAST DESC LIMIT 20
```
At the end of the query is added `DESC`

klimov-paul:
So, it is a matter of skipping automatic `ASC`/`DESC` addition, not `Expression` support.

LAV45:
What do you think about the last change?

klimov-paul:
Makes sense.

framework/data/Sort.php
```diff
@@ -30,8 +30,8 @@
  *         'attributes' => [
  *             'age',
  *             'name' => [
- *                 'asc' => ['first_name' => SORT_ASC, 'last_name' => SORT_ASC],
- *                 'desc' => ['first_name' => SORT_DESC, 'last_name' => SORT_DESC],
+ *                 'asc' => ['first_name' => SORT_ASC, '[[last_name]] ASC'],
+ *                 'desc' => ['first_name' => SORT_DESC, '[[last_name]] DESC'],
```
klimov-paul: Original examples should be preserved.
Direct full sort specification should be added as extra line.

klimov-paul: Also it is better to mention PostgreSQL specific example with `NULLS FIRST` involved to make it more clear when it may be needed.

tests/framework/data/SortTest.php
```diff
@@ -53,6 +53,35 @@ public function testGetOrders()
...
+        $this->assertEquals(SORT_ASC, $orders['age']);
+        $this->assertEquals(SORT_DESC, $orders['first_name']);
+        $this->assertEquals('[[last_name]] DESC NULLS LAST', $orders[0]);
+        $this->assertEquals(['age', 'first_name', 0], array_keys($orders));
+
+        $sort->enableMultiSort = false;
+        $orders = $sort->getOrders(true);
+        $this->assertEquals(1, count($orders));
+        $this->assertEquals(SORT_ASC, $orders['age']);
+    }
```
klimov-paul: I would simplify this test: all waht is necessary is check if expression can be passed as sort attribute value.
There is no need to check multisort here as it is already covered.

klimov-paul (changes requested):
CHANGELOG line is missing

samdark (approved):

SilverFire (approved):

framework/data/Sort.php
```diff
@@ -216,7 +230,11 @@ public function getOrders($recalculate = false)
             $definition = $this->attributes[$attribute];
             $columns = $definition[$direction === SORT_ASC ? 'asc' : 'desc'];
             foreach ($columns as $name => $dir) {
-                $orders[$name] = $dir;
+                if (is_int($name)) {
```
dynasource: this deserves a description so it's objective can be easily understood in the future

dynasource (changes requested):

framework/data/Sort.php
```diff
@@ -100,6 +100,20 @@ class Sort extends Object
      * ]
      * ```
      *
+     * If you are using a PostgreSQL database, you may need to specify the sorting arbitrary expression.
+     * This syntax can be described using the following example:
```
LAV45: @dynasource, and nothing fits this description

klimov-paul:
Resolved by commit 473598ac8ccbc3f0782e41d0ad2a01abd4b55fb2

PR #13267, opened 2016-12-22
Protect remember()/previous() in own session Var

studio255:
To make previous() und remember() work better i would suggest to save the vars in an "protected" session array. otherwise it will interferr with your session (if you don't think about how these little helper will work). 

It's a small improvement.

| Q             | A
| ------------- | ---
| Is bugfix?    | maybe
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no

### DIFF
--- a/framework/helpers/BaseUrl.php
+++ b/framework/helpers/BaseUrl.php
@@ -301,7 +301,8 @@ class BaseUrl
         if ($name === null) {
             Yii::$app->getUser()->setReturnUrl($url);
         } else {
-            Yii::$app->getSession()->set($name, $url);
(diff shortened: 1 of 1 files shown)


SilverFire:
Apart from being a BC break, the developer is responsible for choosing unique name for the session variable name, if it's passed manually.

Thank you for putting your effort and trying to make Yii framework better, but we can not accept your pull request.

kasoft:
You're right. But in this case 

1. both functions are not really needed, you could save your urls directly to the session

2. there is no clue in the docs that the functions use the names as session names. Should be added than

PR #13268, opened 2016-12-22
Logging memory usage

bashkarev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | https://github.com/yiisoft/yii2-debug/issues/34, https://github.com/yiisoft/yii2-debug/pull/184,


samdark (approved):

SilverFire (approved):

framework/log/Logger.php
```diff
@@ -263,27 +264,29 @@ public function getDbProfiling()
...
     public function calculateTimings($messages)
     {
         $timings = [];
         $stack = [];
 
         foreach ($messages as $i => $log) {
-            list($token, $level, $category, $timestamp, $traces) = $log;
-            $log[5] = $i;
+            list($token, $level, $category, $timestamp, $traces, $memory) = $log;
+            $log[6] = $i;
```
dynasource: would be nice to fix this hardcoded number

dynasource (approved):

framework/log/Logger.php
samdark: I don't think it can be done without compatibility breaks.

SilverFire:
Could be merged in 2.0.11 :christmas_tree:

samdark:
Need a line for changelog.

cebe (approved):

framework/log/Logger.php
```diff
@@ -263,27 +264,29 @@ public function getDbProfiling()
...
      */
     public function calculateTimings($messages)
     {
         $timings = [];
         $stack = [];
 
         foreach ($messages as $i => $log) {
-            list($token, $level, $category, $timestamp, $traces) = $log;
-            $log[5] = $i;
+            list($token, $level, $category, $timestamp, $traces, $memory) = $log;
```
cebe: this will break on old logging data, when memory is not set. Will fix that while merging.

PR #13294, opened 2016-12-30
Added missing and fixed wrong plural translations for Latvian language.

mg-code:
- Added missing Latvian translations for `yii\i18n\Formatter:asDuration` function.
- Fixed wrong plural translations for Latvian language. 
In Latvian dot after number stands for ordinal numbers. "2. dienām" equals to "2nd days" (not 2 days).
Word "ago" should be before number.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no


samdark:
Thank you!

Issue #13307, opened 2017-01-01
Preventing of race conditions in script filter in yii.js works incorrectly


arogachev:
I found these issues during writing tests for `yii.js`.

The fix (#12651, https://github.com/yiisoft/yii2/commit/1ab204e481c15a5969098363509407694e7965e1) for #10358 is incorrect. Not only it's not intuitive for understanding (mostly because of the way `if` checks are written), contains copy paste, magic numbers, it also causes some bugs.

The content of `initScriptFilter()` after applying this fix:

```js
function initScriptFilter() {
    var hostInfo = getHostInfo();
    var loadedScripts = {};

    var scripts = $('script[src]').map(function () {
        return this.src.charAt(0) === '/' ? hostInfo + this.src : this.src;
    }).toArray();
    for (var i = 0, len = scripts.length; i < len; i++) {
        loadedScripts[scripts[i]] = true;
    }

    $.ajaxPrefilter('script', function (options, originalOptions, xhr) {
        if (options.dataType == 'jsonp') {
            return;
        }

        var url = options.url.charAt(0) === '/' ? hostInfo + options.url : options.url;

        if (url in loadedScripts) {
            var item = loadedScripts[url];

            // If the concurrent XHR request is running and URL is not reloadable
            if (item !== true && !isReloadable(url)) {
                // Abort the current XHR request when previous finished successfully
                item.done(function () {
                    if (xhr && xhr.readyState !== 4) {
                        xhr.abort();
                    }
                });
                // Or abort previous XHR if the current one is loaded faster
                xhr.done(function () {
                    if (item && item.readyState !== 4) {
                        item.abort();
                    }
                });
            } else if (!isReloadable(url)) {
                xhr.abort();
            }
        } else {
            loadedScripts[url] = xhr.done(function () {
                loadedScripts[url] = true;
            }).fail(function () {
                delete loadedScripts[url];
            });
        }
    });
    
    // Filter for CSS is not related with this fix so its code is not included for brevity
}
```

Let's abstract from using pjax and consider simpler examples using [$.getScript()](https://api.jquery.com/jquery.getscript/) jQuery function. Calls of it will trigger `$.ajaxPrefilter`.

## 1st problem - Setting resolving callbacks only for the 1st request

Let's say we have 2 concurrent requests:

```js
$.getScript('/js/concurrent.js');
$.getScript('/js/concurrent.js');
```

and 2nd will succesfully complete earlier.

[...]

Issue #13312, opened 2017-01-02
skipOuterContainers option is incorrectly passed to pjax in handleAction in yii.js


arogachev:
I found this issue during writing tests for `yii.js`.

Here is the current contents of the final formed `pjaxOptions` array to be passed in pjax plugin:

```js
pjaxOptions = {
    container: pjaxContainer,
    push: pjaxPushState,
    replace: pjaxReplaceState,
    scrollTo: pjaxScrollTo,
    pushRedirect: pjaxPushRedirect,
    replaceRedirect: pjaxReplaceRedirect,
    pjaxSkipOuterContainers: pjaxSkipOuterContainers, // This is incorrect
    timeout: pjaxTimeout,
    originalEvent: event,
    originalTarget: $e
}
```

As you can see, `skipOuterContainers` option is passed as `pjaxSkipOuterContainers`. `pjax` prefix is not needed in the pjax plugin itself.

> **Note:** I already fixed it in PR containing tests for `yii.js`. PR will be sent soon.

Issue #13310, opened 2017-01-02
Handle relative and absolute URLs coincidence in CSS filter in yii.js


arogachev:
I found this issue during writing tests for `yii.js`.

Let's take a look at the current contents of the CSS filter in `yii.js` which is a part of `initScriptFilter()` function and used to prevent loading of the same stylesheets.

```js
$(document).ajaxComplete(function (event, xhr, settings) {
    var styleSheets = [];
    $('link[rel=stylesheet]').each(function () {
        if (isReloadable(this.href)) {
            return;
        }
        if ($.inArray(this.href, styleSheets) == -1) {
            styleSheets.push(this.href)
        } else {
            $(this).remove();
        }
    })
});
```

The url is retrieved from `this.href` and used as is.

With JS filter it's converted to absolute to prevent possible coincidence of urls like `/js/script.js` and `http://foo.bar/js/script.js` which are essentially the same if the current site is `http://foo.bar`:

```js
 var url = options.url.charAt(0) === '/' ? hostInfo + options.url : options.url;
```

So I think the same needs to be done with CSS.

Currently if if the current site is `http://foo.bar` and we load `/css/stylesheet.css` first and then `http://foo.bar/css/stylesheet.css`, it will be loaded as a duplicate.

I think it was simply forgotten to be added for CSS.

> **Note:** I already fixed it in PR containing tests for `yii.js`. PR will be sent soon.

samdark:
Yes, should be fixed,

Issue #13300, opened 2016-12-30
Allow pjax with "data-pjax" with no value in yii.js


arogachev:
I found this issue during writing tests for `yii.js`.

After https://github.com/yiisoft/yii2/commit/9807b2e1a1ac1c671c860ab448eab8bf9154b2bf we lose the ability to have pjax links like 

```html
<a data-pjax href="#">
```

or 

```html
<a data-pjax="" href="#">
```

(which is essentially the same).

This is because of:

```js
pjax = $e.data('pjax') || 0,
```

`"" || 0` resolves to `0`.

This condtradicts with confuring pjax described in the docs for example [here](https://github.com/yiisoft/jquery-pjax#fnpjax):

> Or try this selector that matches any `<a data-pjax href=>` links inside a `<div data-pjax>` container.

I think that the older behavior - `pjax !== undefined && pjax !== 0 && $.support.pjax` is more correct, also note that @arisk sent the PR including this check in #12737 (https://github.com/yiisoft/yii2/pull/12737/commits/383aa471c684e011e48862943a52a8b7c22a08b0), then it was ignored.

It will be better if we disable pjax only if `data-pjax` attribute is set to `0` only:

```js
usePjax = pjax !== undefined && pjax !== 0 && $.support.pjax,
```

Note however that this check in [Yii2 pjax adaptation](https://github.com/yiisoft/jquery-pjax):

```js
// Ignore links with data-pjax="0"
if ($(link).data('pjax') == 0) {
    return;
}
```

needs to be modified accordingly too, because `"" == 0` will evaluate to `true`.

Can be:

```js
$(link).data('pjax') === 0)
```

(`data` returns `0` as `number`).

Or in case of a `string` which is returend by `attr` for example:

```js
$(link).attr('data-pjax') === '0'
```

Or maybe even better:

```js
parseInt($(link).data('pjax')) === 0
```

> Note: I already fixed it in `yii.js` in PR containing tests for `yii.js`. PR will be sent soon.

samdark:
Agree that it should be fixed as you've described.

Mirocow:
Substituting this

PR #13316, opened 2017-01-02
Added JS tests for yii.js (#12840)

arogachev:
| Q             | A
| ------------- | ---
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13300, #13307, #13310, #13312 
| Partially fixed issues  | #13299 (only renamed the private function, renaming according public property is a BC break).
| Other tested issues / PRs  | #8014, #11921, #10974, #11494, #10358, #10097 and others.

This one was not easy. 😰 Anyway I'm glad to present full test suite for `yii.js`. As with previous JS files, I performed deep code analysis, have looked through previous issues and history of changes, found some new issues and fixed some of them during writing tests.

Some of the code is pretty hard to test, but I found workarounds.

I will comment "in place" of made changes in `yii.js`, it will generate a bigger amount of notifications but I think it's more visual and intuitive for understanding.

**Note:** I added test for #8014, but this issue is not reproducible in jsdom. I created according issue there - https://github.com/tmpvar/jsdom/issues/1688.


framework/assets/yii.js
```diff
@@ -17,13 +17,13 @@
...
  * ```javascript
- * yii.sample = (function($) {
+ * window.yii.sample = (function($) {
  *     var pub = {
  *         // whether this module is currently active. If false, init() will not be called for this module
  *         // it will also not be called for all its child modules. If this property is undefined, it means true.
  *         isActive: true,
  *         init: function() {
- *             // ... module initialization code go here ...
+ *             // ... module initialization code goes here ...
```
arogachev: Singular form, it go**es**.

framework/assets/yii.js
```diff
@@ -46,17 +46,18 @@ window.yii = (function ($) {
         /**
          * List of JS or CSS URLs that can be loaded multiple times via AJAX requests.
          * Each item may be represented as either an absolute URL or a relative one.
-         * Each item may contain a wildcart matching character `*`, that means one or more
+         * Each item may contain a wildcard matching character `*`, that means one or more
```
arogachev: Typo.

framework/assets/yii.js
```diff
@@ -46,17 +46,18 @@ window.yii = (function ($) {
...
          * any characters on the position. For example:
-         *  - `/css/*.js` will match any file ending with `.js` in the `css` directory of the current web site
+         *  - `/css/*.css` will match any file ending with `.css` in the `css` directory of the current web site
```
arogachev: Having `js` files in `css` folder is kind of weird. If it was written consciously - it's not a good example, otherwise - it's just a typo.

framework/assets/yii.js
```diff
@@ -46,17 +46,18 @@ window.yii = (function ($) {
...
          *  - `http*://cdn.example.com/*` will match any files on domain `cdn.example.com`, loaded with HTTP or HTTPS
          *  - `/js/myCustomScript.js?realm=*` will match file `/js/myCustomScript.js` with defined `realm` parameter
          */
         reloadableScripts: [],
         /**
          * The selector for clickable elements that need to support confirmation and form submission.
          */
-        clickableSelector: 'a, button, input[type="submit"], input[type="button"], input[type="reset"], input[type="image"]',
+        clickableSelector: 'a, button, input[type="submit"], input[type="button"], input[type="reset"], ' +
```
arogachev: Maximum length of 120 chars per line exceeded.

framework/assets/yii.js
```diff
@@ -107,7 +108,7 @@ window.yii = (function ($) {
          * @param cancel a callback to be called when the user cancels the confirmation
          */
         confirm: function (message, ok, cancel) {
-            if (confirm(message)) {
+            if (window.confirm(message)) {
```
arogachev: For easier creation of a stub for `confirm` in Sinon.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
          *             'name2' => 'value2',
          *         ],
          *     ],
-         * ];
+         * ]);
```
arogachev: Missing closing parenthesis in JSDoc.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
          * ```
          *
          * @param $e the jQuery representation of the element
+         * @param event Related event
```
arogachev: Missing parameter in JSDoc.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
          */
         handleAction: function ($e, event) {
             var $form = $e.attr('data-form') ? $('#' + $e.attr('data-form')) : $e.closest('form'),
                 method = !$e.data('method') && $form ? $form.attr('method') : $e.data('method'),
                 action = $e.attr('href'),
+                isValidAction = action && action !== '#',
```
arogachev: Was copy pasted. Converted check to strict.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
-                pjax = $e.data('pjax') || 0,
-                usePjax = pjax !== 0 && $.support.pjax,
-                pjaxPushState = !!$e.data('pjax-push-state'),
-                pjaxReplaceState = !!$e.data('pjax-replace-state'),
-                pjaxTimeout = $e.data('pjax-timeout'),
-                pjaxScrollTo = $e.data('pjax-scrollto'),
-                pjaxPushRedirect = $e.data('pjax-push-redirect'),
-                pjaxReplaceRedirect = $e.data('pjax-replace-redirect'),
-                pjaxSkipOuterContainers = $e.data('pjax-skip-outer-containers'),
+                areValidParams = params && $.isPlainObject(params),
```
arogachev: Was copy pasted.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
          */
         handleAction: function ($e, event) {
             var $form = $e.attr('data-form') ? $('#' + $e.attr('data-form')) : $e.closest('form'),
                 method = !$e.data('method') && $form ? $form.attr('method') : $e.data('method'),
                 action = $e.attr('href'),
+                isValidAction = action && action !== '#',
                 params = $e.data('params'),
-                pjax = $e.data('pjax') || 0,
-                usePjax = pjax !== 0 && $.support.pjax,
-                pjaxPushState = !!$e.data('pjax-push-state'),
```
arogachev: All these variables are actually redundant. No additional manipulations are done with them, so we can do assignment in `pjaxOptions`.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
-                pjaxReplaceState = !!$e.data('pjax-replace-state'),
-                pjaxTimeout = $e.data('pjax-timeout'),
-                pjaxScrollTo = $e.data('pjax-scrollto'),
-                pjaxPushRedirect = $e.data('pjax-push-redirect'),
-                pjaxReplaceRedirect = $e.data('pjax-replace-redirect'),
-                pjaxSkipOuterContainers = $e.data('pjax-skip-outer-containers'),
+                areValidParams = params && $.isPlainObject(params),
+                pjax = $e.data('pjax'),
+                usePjax = pjax !== undefined && pjax !== 0 && $.support.pjax,
```
arogachev: Fix for #13300.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
                 pjaxContainer,
                 pjaxOptions = {};
 
             if (usePjax) {
-                if ($e.data('pjax-container')) {
```
arogachev: Can be shortened for better readability.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
+                    timeout: $e.data('pjax-timeout'),
                     originalEvent: event,
                     originalTarget: $e
-                }
+                };
             }
 
             if (method === undefined) {
-                if (action && action != '#') {
-                    if (usePjax) {
```
arogachev: Can be a one-liner.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
-                        $.pjax.click(event, pjaxOptions);
-                    } else {
-                        window.location = action;
-                    }
+                if (isValidAction) {
+                    usePjax ? $.pjax.click(event, pjaxOptions) : window.location.assign(action);
                 } else if ($e.is(':submit') && $form.length) {
                     if (usePjax) {
-                        $form.on('submit',function(e){
```
arogachev: Code style.

framework/assets/yii.js
cebe: no need to comment on typo changes, it's obvious :)

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
             }
 
-            var newForm = !$form.length;
-            if (newForm) {
-                if (!action || !/(^\/|:\/\/)/.test(action)) {
-                    action = window.location.href;
+            var oldMethod,
+                oldAction,
+                newForm = !$form.length;
+            if (!newForm) {
```
arogachev: - It's better to have shorter part of `if` first so it's easier to read.
- Have put the part when the form is not exist under `if` check, We do not need any `oldMethod` / `oldAction` processing in case of existing form.

framework/assets/yii.js
```diff
@@ -219,9 +216,10 @@ window.yii = (function ($) {
                 }
                 if (!/(get|post)/i.test(method)) {
                     $form.append($('<input/>', {name: '_method', value: method, type: 'hidden'}));
-                    method = 'POST';
+                    method = 'post';
```
arogachev: - Use lower case. 
- With putting part for the existing form in `if`` check we also need to explicitly set `method` attribute.

framework/assets/yii.js
```diff
@@ -219,9 +216,10 @@ window.yii = (function ($) {
...
+                    $form.attr('method', method);
                 }
-                if (!/(get|head|options)/i.test(method)) {
+                if (/post/i.test(method)) {
```
arogachev: Having check against GET, HEAD and OPTIONS methods is redundant here, because `method` will be only `get` or `post` ath this point (in case of other methods, for example HEAD variable `method` will be assigned to `post` in previous check).

framework/assets/yii.js
```diff
@@ -232,49 +230,41 @@ window.yii = (function ($) {
 
             var activeFormData = $form.data('yiiActiveForm');
             if (activeFormData) {
-                // remember who triggers the form submission. This is used by yii.activeForm.js
+                // Remember the element triggered the form submission. This is used by yii.activeForm.js.
```
arogachev: I think `who` means refering to the person, so it's not suitable here.

framework/assets/yii.js
```diff
@@ -232,49 +230,41 @@ window.yii = (function ($) {
...
                 activeFormData.submitObject = $e;
             }
 
-            // temporarily add hidden inputs according to data-params
-            if (params && $.isPlainObject(params)) {
-                $.each(params, function (idx, obj) {
```
arogachev: `idx` and `obj` are not acceptable here. Params are stored in `name` - `value` structure.

framework/assets/yii.js
```diff
@@ -232,49 +230,41 @@ window.yii = (function ($) {
...
-            var oldMethod = $form.attr('method');
-            $form.attr('method', method);
-            var oldAction = null;
-            if (action && action != '#') {
-                oldAction = $form.attr('action');
-                $form.attr('action', action);
-            }
             if (usePjax) {
-                $form.on('submit',function(e){
+                $form.on('submit', function (e) {
```
arogachev: - Code style problems.
- Missing semicolon.

framework/assets/yii.js
```diff
@@ -232,49 +230,41 @@ window.yii = (function ($) {
...
-            if (params && $.isPlainObject(params)) {
-                $.each(params, function (idx, obj) {
-                    $form.append($('<input/>').attr({name: idx, value: obj, type: 'hidden'}));
+            if (areValidParams) {
+                $.each(params, function (name, value) {
+                    $form.append($('<input/>').attr({name: name, value: value, type: 'hidden'}));
                 });
             }
 
-            var oldMethod = $form.attr('method');
```
arogachev: Moved to the `if` check (see above).

framework/assets/yii.js
```diff
@@ -232,49 +230,41 @@ window.yii = (function ($) {
...
-                    $form.attr('method', oldMethod);
 
-                    // remove the temporarily added hidden inputs
-                    if (params && $.isPlainObject(params)) {
-                        $.each(params, function (idx, obj) {
-                            $('input[name="' + idx + '"]', $form).remove();
-                        });
-                    }
+            $.when($form.data('yiiSubmitFinalizePromise')).then(function () {
+                if (newForm) {
```
arogachev: We don't need any additional processing if the form was new and temporary. Because it will be deleted anyways, we can just use premature `return`.

framework/assets/yii.js
```diff
@@ -232,49 +230,41 @@ window.yii = (function ($) {
...
-            $.when($form.data('yiiSubmitFinalizePromise')).then(
-                function () {
-                    if (oldAction != null) {
-                        $form.attr('action', oldAction);
-                    }
-                    $form.attr('method', oldMethod);
 
-                    // remove the temporarily added hidden inputs
-                    if (params && $.isPlainObject(params)) {
-                        $.each(params, function (idx, obj) {
```
arogachev: `idx` and `obj` are not appropriate for params stored in `name` - `value` structure.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
 
             if (method === undefined) {
-                if (action && action != '#') {
-                    if (usePjax) {
-                        $.pjax.click(event, pjaxOptions);
-                    } else {
-                        window.location = action;
-                    }
+                if (isValidAction) {
+                    usePjax ? $.pjax.click(event, pjaxOptions) : window.location.assign(action);
```
arogachev: Use `assign` method for easier stub creation. See http://stackoverflow.com/a/40495796/4323648. It has no difference - https://developer.mozilla.org/ru/docs/Web/API/Location/assign.

framework/assets/yii.js
```diff
@@ -143,74 +144,70 @@ window.yii = (function ($) {
...
+                        });
                     }
                     $form.trigger('submit');
                 }
                 return;
             }
 
-            var newForm = !$form.length;
-            if (newForm) {
-                if (!action || !/(^\/|:\/\/)/.test(action)) {
```
arogachev: This condradicts with other action checks:

```js
action && action != '#';
```

Also it seems useless. What was the purpose of it? To check if the url is valid? Then this check is not enough.

framework/assets/yii.js
arogachev: @cebe Sorry, indeed. I'll keep this in mind.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
             }
 
             var pairs = url.substring(pos + 1).split('#')[0].split('&'),
-                params = {},
-                pair,
```
arogachev: Predeclaring of `pair and `i` variable seems redundant.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
-                i;
+                params = {};
 
-            for (i = 0; i < pairs.length; i++) {
-                pair = pairs[i].split('=');
+            for (var i = 0, len = pairs.length; i < len; i++) {
```
arogachev: Micro-optimization - calculate `len` only once, at the very beginning of the loop.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
                 var value = decodeURIComponent(pair[1].replace(/\+/g, '%20'));
-                if (name.length) {
-                    if (params[name] !== undefined) {
-                        if (!$.isArray(params[name])) {
-                            params[name] = [params[name]];
-                        }
-                        params[name].push(value || '');
-                    } else {
-                        params[name] = value || '';
+                if (!name.length) {
```
arogachev: Got rid of the extra enclosure of `if` checks with premature `return` for better readability.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
-            if (module.isActive === undefined || module.isActive) {
-                if ($.isFunction(module.init)) {
-                    module.init();
-                }
-                $.each(module, function () {
-                    if ($.isPlainObject(this)) {
-                        pub.initModule(this);
-                    }
-                });
+            if (module.isActive !== undefined && !module.isActive) {
```
arogachev: Got rid of the extra enclosure of `if` checks with premature `return` for better readability.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
+                    pub.initModule(this);
+                }
+            });
         },
 
         init: function () {
             initCsrfHandler();
             initRedirectHandler();
-            initScriptFilter();
+            initAssetFilters();
```
arogachev: Partial fix for #13299.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
             initDataMethods();
+        },
+
+        // Utility functions
```
arogachev: - Added 2 utility functions, both of them need to be publicly accesible in order to create the stubs.
- `getHostInfo` function name was not the best name for what it returns. Besides the host info it also returns protocol. `getBaseCurrentUrl` is a much intuitive IMO.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
+         * Returns the URL of the current page. Used for testing, you can always call `window.location.href` manually
+         * instead.
+         * @returns {string}
+         */
+        getCurrentUrl: function () {
+            return window.location.href;
         }
     };
 
+    function initCsrfHandler() {
```
arogachev: - Reordered private functions to match the order of according calls in `init` method.
- Moved private utility functions to the end.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
+            if (!options.crossDomain && pub.getCsrfParam()) {
+                xhr.setRequestHeader('X-CSRF-Token', pub.getCsrfToken());
+            }
+        });
+        pub.refreshCsrfToken();
+    }
+
     function initRedirectHandler() {
         // handle AJAX redirection
-        $(document).ajaxComplete(function (event, xhr, settings) {
```
arogachev: Redundant parameter.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
+            var url = getAbsoluteUrl(this.src);
+            loadedScripts[url] = true;
+        });
+
+        $.ajaxPrefilter('script', function (options, originalOptions, xhr) {
+            if (options.dataType == 'jsonp') {
+                return;
+            }
+
+            var url = getAbsoluteUrl(options.url),
```
arogachev: Rewrote script filter almost fully to fix #13307.

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
+            xhr.yiiIndex = loadedScripts[url]['xhrList'].length;
+            xhr.yiiUrl = url;
+
+            loadedScripts[url]['xhrList'][xhr.yiiIndex] = xhr;
+        });
+
+        $(document).ajaxComplete(function () {
+            var styleSheets = [];
+            $('link[rel=stylesheet]').each(function () {
+                var url = getAbsoluteUrl(this.href);
```
arogachev: - Fix for #13310.
- Converted check to one-liner.

framework/assets/yii.js
```diff
@@ -374,13 +482,11 @@ window.yii = (function ($) {
             .on('change.yii', pub.changeableSelector, handler);
     }
 
-    function isReloadable(url) {
-        var hostInfo = getHostInfo();
+    // Utility functions
 
+    function isReloadableAsset(url) {
```
arogachev: `isReloadable` is kind of broad name for a function.

framework/assets/yii.js
```diff
@@ -395,76 +501,18 @@ window.yii = (function ($) {
...
-                    $(this).remove();
-                }
-            })
-        });
+    /**
+     * Returns absolute url based on the given url
+     * @param {string} url Initial url
+     * @returns {string}
+     */
+    function getAbsoluteUrl(url) {
```
arogachev: This was copy pasted.

framework/assets/yii.js
```diff
@@ -395,76 +501,18 @@ window.yii = (function ($) {
...
+        return url.charAt(0) === '/' ? pub.getBaseCurrentUrl() + url : url;
     }
 
     return pub;
-})(jQuery);
+})(window.jQuery);
```
arogachev: It's better to use `window` here. Plus the consistency with other JS files.

framework/assets/yii.js
```diff
@@ -395,76 +501,18 @@ window.yii = (function ($) {
...
 
-jQuery(function () {
-    yii.initModule(yii);
+window.jQuery(function () {
+    window.yii.initModule(window.yii);
```
arogachev: Variable `yii` was set using `window.yii`, but then accessed with just `yii`. This is also hard to test.

framework/assets/yii.js
```diff
@@ -17,13 +17,13 @@
  * A module may be structured as follows:
  *
  * ```javascript
- * yii.sample = (function($) {
+ * window.yii.sample = (function($) {
```
arogachev: Added `window` to better reflect Yii modules.

framework/web/Response.php
```diff
@@ -782,7 +782,7 @@ protected function getDispositionHeaderValue($disposition, $attachmentName)
      *
      * ```javascript
      * $document.ajaxComplete(function (event, xhr, settings) {
-     *     var url = xhr.getResponseHeader('X-Redirect');
+     *     var url = xhr && xhr.getResponseHeader('X-Redirect');
```
arogachev: Sync with #10974.

arogachev:
Seems like current Code Climate ESLint config does not allow one-liners with expressions?

```js
usePjax ? $.pjax.click(event, pjaxOptions) : window.location.assign(acti
```

Also what's wrong with:

```js
if (window.confirm(message)) {
}
```

cebe:
don't take code climate too serious we made a lot of adjustments to make it fit in PHP. We can adjust it for JS later if needed.

framework/assets/yii.js
samdark: It's there from 2013. Was implemented by Qiang and, indeed, looks like URL validation.

framework/assets/yii.js
```diff
@@ -232,49 +230,41 @@ window.yii = (function ($) {
...
+            $.when($form.data('yiiSubmitFinalizePromise')).then(function () {
+                if (newForm) {
+                    $form.remove();
+                    return;
+                }
 
-                    if (newForm) {
-                        $form.remove();
-                    }
+                if (oldAction !== undefined) {
```
samdark: Can't we check as `if (oldAction)`?

framework/assets/yii.js
```diff
@@ -284,67 +274,185 @@ window.yii = (function ($) {
...
-                        if (!$.isArray(params[name])) {
-                            params[name] = [params[name]];
-                        }
-                        params[name].push(value || '');
-                    } else {
-                        params[name] = value || '';
+                if (!name.length) {
+                    continue;
+                }
+                if (params[name] === undefined) {
```
samdark: Could we check for `if (params[name])` here?

framework/assets/yii.js
samdark: No need for this section comment, I think...

framework/assets/yii.js
```diff
@@ -374,13 +482,11 @@ window.yii = (function ($) {
             .on('change.yii', pub.changeableSelector, handler);
     }
 
-    function isReloadable(url) {
-        var hostInfo = getHostInfo();
+    // Utility functions
```
samdark: No need for such section comments.

framework/assets/yii.js
```diff
@@ -395,76 +501,18 @@ window.yii = (function ($) {
...
-                }
-                if ($.inArray(this.href, styleSheets) == -1) {
-                    styleSheets.push(this.href)
-                } else {
-                    $(this).remove();
-                }
-            })
-        });
+    /**
+     * Returns absolute url based on the given url
```
samdark: url → URL

samdark (changes requested):

framework/assets/yii.js
arogachev: No. See the next line:

```js
params[name] = value || '';
```

If the value was set to an empty string `if (params[name]) { }` check will evaluate to `false` with existing key.

As an alternative we can use `in` or `hasOwnProperty`.

framework/assets/yii.js
arogachev: Technically - yes, because `oldAction` is set only when `isValidAction` is truthy:

```js
if (isValidAction) {
    oldAction = $form.attr('action');
    $form.attr('action', action);
}
```

And `isValidAction` disallows all falsy values and additionally anchor (`#`):

```js
isValidAction = action && action !== '#',
```

But checking for `undefined` is more clear IMO.

samdark (approved):

framework/assets/yii.js
```diff
@@ -284,67 +274,183 @@ window.yii = (function ($) {
...
+            initAssetFilters();
             initDataMethods();
+        },
+
+        /**
+         * Returns the URL of the current page without params and trailing slash. Separated and made public for testing.
+         * @returns {string}
+         */
+        getBaseCurrentUrl: function () {
+            return window.location.protocol + '//' + window.location.host;
```
SilverFire: Will work incorrect when port is defined, e.g. `localhost:8080`
It will be god to cover it with a test

SilverFire (changes requested):

SilverFire:
Incredible job done, we highly appreciate your efforts. Thank you very much, @arogachev, for making Yii better!

framework/assets/yii.js
arogachev: `window.location.host` includes port.

For example for given url `http://localhost:8000/` the result of:

```js
window.location.protocol + '//' + window.location.host
```
is `http://localhost:8000`, it's correct. So this is not a problem.

SilverFire: I was wrong, mixed up `host` and `hostname` properties.

samdark:
Merged. Thank you very much for this wonderful pull request.

PR #13329, opened 2017-01-05
Added small specification to result of conversion of object to array

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

Small remark from @Nur777 in russian Gitter room  - https://gitter.im/yiisoft/yii2/rus?at=586d2142c895451b75559b44.

It's better to explicitly mention that this is a result for single model, or put inside an other array to show the whole result of called method. Otherwise it's a bit confusing.


samdark:
Thank you!

PR #13336, opened 2017-01-06
Fixed doc about supporting inheritance (traits vs behaviors)

masterklavi:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

> Traits, on the other hand, can be considered as language-supported copy and paste. They do not support inheritance.

There is something wrong. Look at http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.multiple
> Multiple Traits can be inserted into a class.
```
class MyHelloWorld {  
    use Hello, World;  
    public function sayExclamationMark() {  
        echo '!';  
    }  
}
```

### DIFF
--- a/docs/guide/concept-behaviors.md
+++ b/docs/guide/concept-behaviors.md
@@ -336,9 +336,6 @@ both have pros and cons. They are more like complements to each other rather tha
 
 ### Reasons to Use Behaviors <span id="pros-for-behaviors"></span>
 
-Behavior classes, like normal classes, support inheritance. Traits, on the other hand,
-can be considered as language-supported copy and paste. They do not support inheritance.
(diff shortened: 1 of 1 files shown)


samdark:
What's supported by traits isn't inheritance.

masterklavi:
But it's not a negative point of traits.

samdark:
In context of behaviors it is. That's what initially prevented us dropping behaviors in favor of traits while Yii 2 was designed.

samdark:
Don't think these changes are appropriate but thanks for proposal anyway.

PR #13340, opened 2017-01-07
Exception within useMaster() completely disables slaves

Vovan-VE:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | ~hope yes~ yes
| Fixed issues  | (not reported yet)


samdark:
Could you elaborate about it?

Vovan-VE:
`Connection::useMaster()` disables slaves by `enableSlaves = false`, than calls callback and restore `enableSlaves` to old value in the end. In case of uncaught exception from a callback `enableSlaves` will not restore.

samdark (approved):
Looks good to me.

SilverFire (changes requested):
Could you add a CHANGELOG and a test case?

SilverFire (approved):

dynasource (approved):
look good. Well done.

PR #13341, opened 2017-01-07
Note about cache key limitation

bizley:
multiSet() and multiAdd() use array to pass key with value which limits key structure to string and integer.


samdark:
Merged. Thanks!

PR #13346, opened 2017-01-08
Add test case for deadlock in MySQL

### LINKED ISSUE #12715: Exception `SAVEPOINT LEVEL1 does not exist` instead of deadlock exception

``` php
\Yii::$app->db->transaction(function () {
    \Yii::$app->db->transaction(function () {
        // here comes queries to cause unexpected deadlock
    });
});
```

See an exception about deadlock.
### What do you get instead?

Exception `SAVEPOINT LEVEL1 does not exist` from https://github.com/yiisoft/yii2/blob/2.0.9/framework/db/Connection.php#L690
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.9 |
| PHP version | 7.0.8 |
| MySQL | 5.7 |
| OS | Ubuntu 16.04.2 |

Vovan-VE:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | initially 'yes', then intended 'no' and after merging fix 'yes' again
| Fixed issues  | confirm and fix #12715

At first here goes test case which do deadlock and catch it as successful result. Don't merge.


Vovan-VE:
Then here comes just one nested transaction to actually fail the test case.

Vovan-VE:
> There was 1 failure:
> 1) yiiunit\framework\db\mysql\ConnectionTest::testDeadlockException
> child exited with error status. Shared children log:
> [2017-01-08 03:16:29.325] child 2: wait signal from child 1
> [2017-01-08 03:16:29.326] child 1: connect
> [2017-01-08 03:16:29.327] child 1: delete
> [2017-01-08 03:16:29.328] child 1: insert
> [2017-01-08 03:16:29.333] child 1: transaction
> [2017-01-08 03:16:29.335] child 1: select
> [2017-01-08 03:16:29.336] child 1: send signal to child 2
> [2017-01-08 03:16:29.336] child 2: connect
> [2017-01-08 03:16:29.337] child 2: transaction
> [2017-01-08 03:16:29.340] child 2: update
> [2017-01-08 03:16:31.336] child 1: update
> [2017-01-08 03:16:31.337] child 1: commit
> [2017-01-08 03:16:31.338] child 2: exit
> [2017-01-08 03:16:31.339] child 2: ! sql error 42000: 1305: SAVEPOINT LEVEL1 does not exist

I hope the issue can be confirmed with this without merging this test case.

dynasource (approved):
thanks for taking time to dive into this issue. The approach with several committing steps is helpful to indicate your solution to be working. 

The test can be made easier to understand though, by dedicating it into pieces in a separate class.

samdark (approved):

framework/db/Connection.php
```diff
@@ -713,12 +713,22 @@ public function transaction(callable $callback, $isolationLevel = null)
             }
         } catch (\Exception $e) {
             if ($transaction->isActive && $transaction->level === $level) {
-                $transaction->rollBack();
+                try {
+                    $transaction->rollBack();
+                } catch (\Exception $e_rollback) {
```
samdark: `$e_rollback` could be named `$rollbackException`.

samdark (changes requested):

framework/db/Connection.php
```diff
@@ -713,12 +713,22 @@ public function transaction(callable $callback, $isolationLevel = null)
             }
         } catch (\Exception $e) {
             if ($transaction->isActive && $transaction->level === $level) {
```
samdark: These two blocks are exactly the same. Is it possible to extract these into a method?

samdark (changes requested):

framework/db/Connection.php
Vovan-VE: Should it be a method in `Connection` with arguments `($transaction, $level)` or method in `Transaction` with arguments `($level)`. Also these two blocks will become one when minimum PHP version will be required 7.

SilverFire (approved):
One of the most complex tests in Yii :)

framework/db/Connection.php
samdark: Connection. And it should not be public.

samdark (approved):

samdark:
Looks good to me. Please add changelog and we'll merge it.

Vovan-VE:
IMHO, this PR should not be merged. Provided test case is not stable (PHP 5.6 segfailt) and platform specific (fork and signals). It's too complex and too ugly.
#13347 contains fix only and should be merged instead.

samdark:
I don't like to have code for such complicated case not covered by a unit test. In the test method we can probably check for pcntl absence or PHP 5.6 and skip them.

dynasource:
agree with @samdark . Lets keep it and see how the unit test performs. We can always remove / simplify later on.

Vovan-VE:
Last too commits shows that SEGFAULT is unstable at least for PHP 5: https://travis-ci.org/yiisoft/yii2/builds/204767330 - all passed; https://travis-ci.org/yiisoft/yii2/builds/204767391 - all 5.* segfault.
How to locate the reason? I need to do some refactoring anyway. Where should I extract some classes? `tests/framework/db/mysql/?/*.php`

dynasource:
That's a pity indeed. Thanks for looking into it. A new separate class dedicated to this issue would be nice like i.e. /tests/framework/db/mysql/connection/DeadLockTest.php

samdark:
Merged. Thank you very much.

framework/db/Connection.php
```diff
@@ -713,21 +713,37 @@ public function transaction(callable $callback, $isolationLevel = null)
...
         return $result;
     }
 
     /**
+     * Rolls back given [[Transaction]] object if it's still active and level match.
+     * In some cases rollback can fail, so this method is fail safe. Exception thrown
+     * from rollback will be caught and just logged with [[\Yii::error()]].
+     * @param Transaction $transaction Transaction object given from [[beginTransaction()]].
+     * @param int $level Transaction level just after [[beginTransaction()]] call.
+     */
```
cebe: `@since 2.0.12` missing.

cebe:
Great job with the unit test!

framework/db/Connection.php
cebe: I think this could even be made private. if there is a need to customize this one can override transaction().

cebe: fixed: 7d4b33a

Vovan-VE:
Thanks. But it seems to me that I did add one more place for segfaulting at least PHP 5.4 on tests.

cebe:
PHP segfauling means there is a bug in PHP as PHP code itself should not be able to segfault.

Vovan-VE:
Yes, but from now whole build will be treated as failed.
Previously before last refactoring segfault was inconsistent for PHP 5.*. Now it became looks consistent for 5.4.

samdark:
Would be a good idea to either fix it or skip it for 5.4...

cebe:
we need to redo our travis setup anyway in some kind as the mysql errors are still occuring.

Vovan-VE:
5.4 did not segfault here https://travis-ci.org/yiisoft/yii2/builds/210844774

samdark:
I've restarted build. Let's see how it goes.

samdark:
Hmm... works :)

PR #13347, opened 2017-01-08
Fix #12715: deadlock exception lost with nested transaction

### LINKED ISSUE #12715: Exception `SAVEPOINT LEVEL1 does not exist` instead of deadlock exception

Vovan-VE:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12715

Confirmed and fixed in PR #13346.


samdark:
@Vovan-VE isn't #13346 include this pull request changes?

Vovan-VE:
I thought #13346 will not be merged because of too complicated and possibly unstable test case. So, here is fix only.

samdark:
I see. Thanks for explanation.

Vovan-VE:
This PR is included by #13346, which has also test case. Closing this.

PR #13350, opened 2017-01-08
Improve UrlManager::createUrl() cache

### LINKED ISSUE #14094: UrlManager::createUrl() cache may degrade performance in some cases
Consider this test: https://github.com/rob006/yii2-dev/blob/2f520f339dc616f2875ebf258cbacdd4e9ea597d/tests/framework/web/UrlManagerCreateUrlTest.php#L707

`UrlManager::createUrl()` was called 2 times, but this results 3 `UrlRule::createUrl()` calls. That's because on second URL creation first check is on [cached rule](https://github.com/rob006/yii2-dev/blob/2f520f339dc616f2875ebf258cbacdd4e9ea597d/framework/web/UrlManager.php#L466), and then again on [regular rules array](https://github.com/rob006/yii2-dev/blob/2f520f339dc616f2875ebf258cbacdd4e9ea597d/framework/web/UrlManager.php#L378).

There is no point to check the same rule 2 times, second `UrlRule::createUrl()` call should be skipped.

refs #13350

rob006:
Similar bug to #9091 - if rule is skipped because of param value does not match pattern, it is also ignored in all further URL creations.

Fixes #11288
Fixes #14094
Replaces #14104


framework/CHANGELOG.md
```diff
@@ -51,6 +52,7 @@ Yii Framework 2 Change Log
 - Bug #14052: Fixed processing parse errors on PHP 7 since these are instances of `\ParseError` (samdark)
 - Bug #14074: Fixed default value of `yii\console\controllers\FixtureController::$globalFixtures` to contain valid class name (lynicidn)
 - Enh #8641: Enhanced `yii\console\Request::resolve()` to prevent passing parameters, that begin from digits (silverfire)
+- Enh #11288: Added support for caching of `yii\web\UrlRule::createUrl()` results in `yii\web\UrlManager` for rules with defaults (rob006)
```
samdark: Are two changelog lines necessary?

framework/web/CompositeUrlRule.php
```diff
@@ -24,6 +24,12 @@
      */
     protected $rules = [];
 
+    /**
+     * @var string|null status of URL creation after last [[createUrl()]] call.
+     * @since 2.0.12
+     */
+    public $createStatus;
```
samdark: Is it necessary for it to be public?

framework/web/UrlManager.php
```diff
@@ -428,6 +429,27 @@ public function createUrl($params)
...
+     *
+     * @param UrlRuleInterface $rule
+     * @return bool|null `true` if result should be cached, `false` if not, `null` if rule does not provide info about
+     * create URL status.
+     * @since 2.0.12
+     * @see getUrlFromCache()
+     * @see setRuleToCache()
+     * @see UrlRule::$createStatus
+     */
+    protected function canBeCached(UrlRuleInterface $rule)
```
samdark: How about moving it to the rule itself?

framework/CHANGELOG.md
rob006: These are two separate things IMO, but I could merge in to one line with some general description.

framework/web/UrlManager.php
rob006: That was my first shot, but I'm not sure if rule should decide about UrlManager internal cache logic. I'm open to suggestions in this case.

framework/CHANGELOG.md
samdark: OK. That's fine as it is.

framework/web/UrlManager.php
```diff
@@ -428,6 +429,27 @@ public function createUrl($params)
...
+    {
+        if (isset($rule->createStatus)) {
+            return in_array($rule->createStatus, [
+                UrlRule::CREATE_STATUS_PARAMS_MISMATCH,
```
samdark: Is it necessary to know exact status?

rob006: If we want keep cache logic in `UrlManager` I have no idea how to solve this differently.

framework/web/UrlManager.php
SilverFire: Agreed with @rob006 
Caching is not a knowledge of Rule itsels

framework/web/UrlRule.php
```diff
@@ -108,6 +113,11 @@ class UrlRule extends Object implements UrlRuleInterface
      * @since 2.0.7
      */
     protected $placeholders = [];
+    /**
+     * @var string|null status of URL creation after last [[createUrl()]] call.
```
SilverFire: `status of the URL creation after the last`

framework/web/UrlRule.php
```diff
@@ -108,6 +113,11 @@ class UrlRule extends Object implements UrlRuleInterface
...
+     * @since 2.0.12
+     */
+    public $createStatus;
```
SilverFire: I don't think this property should be public. I see no reason to modify this property from outside of the class.

`getCreateStatus()` method may be moved here from stub.

SilverFire (changes requested):

framework/rest/UrlRule.php
```diff
@@ -244,17 +246,33 @@ public function parseRequest($manager, $request)
...
             if (strpos($route, $controller) !== false) {
                 foreach ($this->rules[$urlName] as $rule) {
-                    /* @var $rule \yii\web\UrlRule */
+                    /* @var $rule WebUrlRule */
                     if (($url = $rule->createUrl($manager, $route, $params)) !== false) {
+                        $this->createStatus = WebUrlRule::CREATE_STATUS_SUCCESS;
                         return $url;
+                    } elseif (
+                        $this->createStatus === null
+                        || !method_exists($rule, 'getCreateUrlStatus')
```
samdark: Is that for compatibility?

framework/web/CompositeUrlRule.php
```diff
@@ -24,6 +24,12 @@
      */
     protected $rules = [];
 
+    /**
+     * @var int|null status of the URL creation after the last [[createUrl()]] call.
+     * @since 2.0.12
+     */
+    protected $createStatus;
```
samdark: I think the name could be better. How about `createResult` instead of status?

framework/web/UrlManager.php
```diff
@@ -428,6 +429,24 @@ public function createUrl($params)
     }
 
     /**
+     * Returns the value indicating whether result of [[createUrl()]] of rule should be cached in internal cache.
+     *
+     * @param UrlRuleInterface $rule
+     * @return bool|null `true` if result should be cached, `false` if not, `null` if rule does not provide info about
```
samdark: I'd either return false if rule doesn't provide info.

framework/rest/UrlRule.php
rob006: Yes. `UrlRuleInterface` does not contain this method, so you can't be sure that it exist.

framework/web/CompositeUrlRule.php
rob006: Result is returned by `createUrl()`, so I don't think that this is better name.

framework/web/UrlManager.php
rob006: Then this bug will still exists if custom rule does not provide info about `createStatus` - rule may return `false` if some params does not match, and then it will be skipped in all further URLs generation, even if different (matching) params are passed to `UrlManager::createUrl()`.

rob006:
BTW: In 2.1 we could add `getCreateUrlStatus()` to `UrlRuleInterface` and force to return `bool` - this will simplify some conditions.

rob006:
Done.

framework/web/UrlManager.php
samdark: I see... right.

samdark:
@rob006 is it possible to address https://codeclimate.com/github/yiisoft/yii2/pull/13350 ?

framework/UPGRADE.md
```diff
@@ -64,6 +64,10 @@ Upgrade from Yii 2.0.11
 
 * The signature of `yii\cache\Cache::getOrSet()` has been adjusted to also accept a callable and not only `Closure`.
   If you extend this method, make sure to adjust your code.
+  
+* `yii\web\UrlManager` now checks if rules implements `getCreateUrlStatus()` method in order to decide whether to use
```
samdark: Either should be `rule implements` or `rules implement`.

framework/UPGRADE.md
```diff
@@ -64,6 +64,10 @@ Upgrade from Yii 2.0.11
...
+  internal cache for `createUrl()` calls. Ensure that all your custom rules implements this method in order to fully
```
samdark: custom rules implements → custom rules implement

framework/web/CompositeUrlRule.php
```diff
@@ -68,13 +73,62 @@ public function parseRequest($manager, $request)
...
+
+        if ($this->createStatus === UrlRule::CREATE_STATUS_SUCCESS) {
+            // create status was not changed - there is no rules configured
+            $this->createStatus = UrlRule::CREATE_STATUS_PARSING_ONLY;
+        }
+        return false;
+    }
+
+    /**
+     * Iterates through specified rules and call [[createUrl()]] for each of them.
```
samdark: Iterates through specified rules and call → Iterates through specified rules and calls

rob006:
@samdark https://github.com/yiisoft/yii2/pull/13350/commits/fa9760cd465a4a6e2e0c7b6af1c8fc060d19838e ?

framework/web/CompositeUrlRule.php
```diff
@@ -68,13 +73,62 @@ public function parseRequest($manager, $request)
...
         return false;
     }
+
+    /**
+     * Returns status of the URL creation after the last [[createUrl()]] call.
+     *
+     * For multiple rules statuses will be combined by bitwise `or` operator
+     * (e.g. `UrlRule::CREATE_STATUS_PARSING_ONLY | UrlRule::CREATE_STATUS_PARAMS_MISMATCH`).
+     *
+     * @return null|int
```
samdark: It should be stated clearly what null means.

framework/web/UrlRule.php
```diff
@@ -37,6 +37,28 @@ class UrlRule extends Object implements UrlRuleInterface
      * Set [[mode]] with this value to mark that this rule is for URL creation only
      */
     const CREATION_ONLY = 2;
+    /**
+     * Represents the successful URL generation by last [[createUrl()]] call.
+     * @see $createStatus
+     */
+    const CREATE_STATUS_SUCCESS = 0;
```
samdark: I think `since` tags should be added for constants as well.

samdark (changes requested):

rob006:
Done.

framework/web/UrlManager.php
rob006: On second thought we could treat `null` as `true` - such rules will take precedence (as they're always cached) over regular rules, but this still should be more efficient than disabling the whole cache because of one rule which not implements `getCreateUrlStatus()`.

framework/web/CompositeUrlRule.php
```diff
@@ -68,13 +73,64 @@ public function parseRequest($manager, $request)
...
+     * @since 2.0.12
+     */
+    protected function iterateRules($rules, $manager, $route, $params) {
+        /* @var $rule UrlRule */
+        foreach ($rules as $rule) {
+            $url = $rule->createUrl($manager, $route, $params);
+            if ($url !== false) {
+                $this->createStatus = UrlRule::CREATE_STATUS_SUCCESS;
                 return $url;
+            } elseif (
```
samdark: Could be just `if` since previously there's a `return`.

framework/web/CompositeUrlRule.php
```diff
@@ -68,13 +73,64 @@ public function parseRequest($manager, $request)
...
+     * For multiple rules statuses will be combined by bitwise `or` operator
+     * (e.g. `UrlRule::CREATE_STATUS_PARSING_ONLY | UrlRule::CREATE_STATUS_PARAMS_MISMATCH`).
+     *
+     * @return null|int Status of the URL creation after the last [[createUrl()]] call. `null` if rule does not provide
+     * info about create status.
+     * @see $createStatus
+     * @see http://php.net/manual/en/language.operators.bitwise.php
+     * @since 2.0.12
+     */
+    public function getCreateUrlStatus() {
```
samdark: `{` should be on new line.

framework/web/CompositeUrlRule.php
```diff
@@ -68,13 +73,64 @@ public function parseRequest($manager, $request)
...
+     *
+     * @param UrlRuleInterface[] $rules rules to iterate.
+     * @param UrlManager $manager the URL manager
+     * @param string $route the route. It should not have slashes at the beginning or the end.
+     * @param array $params the parameters
+     * @return bool|string the created URL, or `false` if none of specified rules cannot be used for creating this URL.
+     * @see createUrl()
+     * @since 2.0.12
+     */
+    protected function iterateRules($rules, $manager, $route, $params) {
```
samdark: `{` should on new line.

samdark (approved):
Overall looks good. Commented on minor things to fix.

framework/web/UrlRule.php
```diff
@@ -420,6 +451,7 @@ public function parseRequest($manager, $request)
     public function createUrl($manager, $route, $params)
     {
         if ($this->mode === self::PARSING_ONLY) {
+            $this->createStatus = static::CREATE_STATUS_PARSING_ONLY;
```
dynasource: why static:: and not self::

framework/web/CompositeUrlRule.php
dynasource: name is clear for me

framework/web/UrlManager.php
dynasource: methods starting with 'can' should return booleans. Having multiple return types is not optimal

dynasource (changes requested):
quite some work you've done here, well done

framework/web/UrlRule.php
samdark: Indeed, `static::` doesn't make much sense here.

tests/framework/web/UrlRuleTest.php
```diff
@@ -1285,4 +1287,99 @@ protected function getTestsForToString()
...
+                        ['page' => 1, 'tag' => 'a', 'sort' => 'no', 'category' => 'my-category'],
+                        'post/a/no?category=my-category',
+                        UrlRule::CREATE_STATUS_SUCCESS,
+                    ],
+                    ['post/index', ['page' => 1], false, UrlRule::CREATE_STATUS_PARAMS_MISMATCH],
+                    ['post/index', ['page' => '1abc', 'tag' => 'a'], false, UrlRule::CREATE_STATUS_PARAMS_MISMATCH],
+                    ['post/index', ['page' => 1, 'tag' => 'a', 'sort' => 'YES'], false, UrlRule::CREATE_STATUS_PARAMS_MISMATCH],
+                ],
+            ],
+            'paring only' => [
```
Enkey: Small typo: `paring` -> `parsing`

rob006:
Done.

dynasource (approved):

samdark:
Seems everything's fine. Merge?

framework/web/UrlRule.php
```diff
@@ -96,6 +122,11 @@ class UrlRule extends Object implements UrlRuleInterface
      * @since 2.0.10
      */
     public $normalizer;
+    /**
+     * @var int|null status of the URL creation after the last [[createUrl()]] call.
+     * @since 2.0.12
+     */
+    public $createStatus;
```
SilverFire: The following discussion was not resolved https://github.com/yiisoft/yii2/pull/13350#discussion_r114478945

I think this property should not bt public.

SilverFire (changes requested):

framework/web/UrlRule.php
rob006: Fixed.

cebe (approved):
Good job!

samdark:
Merged. Thank you very much!

PR #13354, opened 2017-01-09
Replace (isset) with (!empty) beacuse isset('') always means true, 'on' condition default to empty string.

suyar:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | join condition $on default to '', isset('') always means true, so I replace 'isset()' with '!empty()'
| note | I'm sorry, my english is poor.


samdark:
Why do you have empty string as join condition?

carolkey:
I'm sorry that it does not affect the code execution, I think I should look at the code more closely.

samdark:
Closing the issue for now then.

Issue #13047, opened 2016-11-18
MSSQL tests


cebe:
Just found it, have not checked it in detail.

Would this allow us to run unit tests against MSSQL?

https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-setup-ubuntu

Possibly even on travis?

schmunk42:
Should be able with Docker, works on Travis and there's also a container for it https://hub.docker.com/r/microsoft/mssql-server-linux/

More expensive tests might only be run on specific branches.

Let me know if you need more input.

samdark:
Yes, it seems this way of solving it is good.

schmunk42:
Above PR & example build/test https://git.hrzg.de/schmunk42/yii2/builds/73924

samdark:
Seems to be done.

PR #13360, opened 2017-01-10
added dockerized test setup

schmunk42:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | it's about testing
| Fixed issues  | relates https://github.com/yiisoft/yii2/issues/13047

essential

- [x] base-image (rela
- [x] mssql stack
- [x] cubrid stack
- [ ] host-volumes (yes/no?)
- [ ] reports

optional

- [x] additonal tools/dev-image(?)
- [x] isolation in GitLab CI

related: https://github.com/Codeception/Codeception/pull/3739


tests/docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+
+  postgres:
+    image: postgres
+    environment:
+      - POSTGRES_DB=yiitest
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+
+  #sqlsrv:
+  #  image: microsoft/mssql-server-linux
```
samdark: Are you going to enable it as well?

schmunk42: Yes, I tried, but you need to install some additional packages - just did not work right away. Let me know if you've some input here, I've never worked with MSSQL.

samdark: I have no experience with dockerizing it or even running it under Linux.

schmunk42:
## TL;dr

So, theoretically this is a complete **setup**. You can view an example GitHub build [here](https://git.hrzg.de/schmunk42/yii2/pipelines/17792) - they're not green yet.

I've split the setup into 3 stacks: default, mssql and cubrid. We can add more later. There are 4 test jobs, which are run in parallel. Whenever possible I've used pre-built images.

Currently there are different base-images, I intentionally left it in this state, because it shows a few issues we should sort out, before continuing.

### Choosing a base image

There are several options how to create a Docker image for testing. You can build from different base-images, like `php`, `composer` or `codeception/codeception` or `debian`, `ubuntu`, ....
Currently there's no official Yii image, another option.

Basically you need to install the required PHP-extensions for Yii (that's why some tests fail, because this is just a vanilla PHP) and for some suites also additional extensions.

I'd say it's not really possible to build a Yii image with all extensions (pdo_mssql, pdo_cubrid, xcache etc...) just because some are not available for `debian`, etc... So I'd just choose what works, i.e. `codemix` base-image, custom build for cubrid, pre-build for mssql.

In tests, there are dependencies from i.e. log suite to db, which make tests fail which would actually run on a vanilla PHP. I thought about also putting them into group `db`, WDYT?
These components might also be reviewed for decoupling.

Depending on how you wanna build the image you also need composer and some other tools, like npm or yarn, i.e. for testing asset bundling via webpack. But that's a separate topic.

Tests do not pass on Travis now, this is just because of changed DB credentials and hosts (should be easy to fix)

---

While I could easily write 5x times more :) I'd like stop here and also CC: @mikehaertl and @motin - your experience and feedback about Docker, PHP & Yii is very welcome.

Would be really nice if you could give it a try and just report if the things start to work.

motin:
Love it. Every open source project should offer a way of running database-dependent tests without having to follow extensive complicated instructions on how to set up databases and credentials. @schmunk42, please supply a single shell-script that will run all tests with all database engines in sequence and I'll verify that it works over here :)

schmunk42:
This should be ready to try out locally.
Let me know if you've questions...

schmunk42:
Example build pipelines
- https://git.hrzg.de/schmunk42/yii2/pipelines/18231 (all, except oci & mssql)
- https://git.hrzg.de/schmunk42/yii2/builds/72942 (like travis but twice as fast)

tests/data/config.php
```diff
@@ -16,29 +16,29 @@
 $config = [
     'databases' => [
         'cubrid' => [
-            'dsn' => 'cubrid:dbname=demodb;host=localhost;port=33000',
+            'dsn' => 'cubrid:dbname=demodb;host=cubrid;port=33000',
```
schmunk42: @cebe @samdark How should we handle this? Using a separate file for Docker tests. It's theoretically possible to fake this with Docker, but we would loose a lot of  isolation & stability.

schmunk42:
Example build: https://git.hrzg.de/schmunk42/yii2/pipelines/18450/builds

tests/data/config.php
samdark: What is the problem here?

schmunk42:
You can't really fake localhost in the dockerized setup.
It also makes no sense to build the database servers into the PHP image.

Can we just create another config file, which is loaded via ENV bar?

Von meinem iPhone gesendet

samdark:
Yes, we can. Why not?

schmunk42:
Was just a discussion :)
I'll add it to the PR...

Von meinem iPhone gesendet

schmunk42:
Would be ready from my side.

.gitlab-ci.yml
```diff
@@ -0,0 +1,118 @@
...
+stages:
+  - travis
+  - test
+  - cleanup
+
+test:
+  stage: test
+  script:
+    - docker-compose up --build -d
+    - docker-compose run --rm php vendor/bin/phpunit -v --exclude caching,db,data --log-junit tests/_junit/test.xml
```
SilverFire: `--rm` option is redundant container was started in background (`docker-compose up -d`). Or I am missing something?

SilverFire (changes requested):
Some bash files do not have a newline at the end.

Great job @schmunk42! Thank you!

.gitlab-ci.yml
cebe: starting in the background is `-d`, removing the container after it has been stopped is `--rm`.

SilverFire: The line I've commented has -rm argument, that according to the docs should remove container after the command execution, but ONLY when the container is NOT running in the background.

Line above (docker-compose up --build -d) stars container in background.
Does is make any sense to pass --rm flag for the RUN command in this case?

schmunk42: There are actually two containers running in line 30. The one in the background (webserver) and the actual run process which does the testing.
I'd say you usually (depends on the use-case) run `run` with `--rm`, I does not really matter in the CI, but it should be cleaner in general, see also https://github.com/docker/compose/issues/943

@mikehaertl @motin WDYT?

PS: Cleanup is actually done in `after_script`.

motin: It is good practice and relevant in this case to clean up with "--rm" when using `run` since `run` always starts a new container which otherwise will be left behind on the Docker host.

samdark:
@SilverFire I don't have enough expertise to review it.

.gitlab-ci.yml
SilverFire: Yeah, I have mixed up `run` and `exec`. Thank you @schmunk42

samdark:
@schmunk42 do we have to commit composer.lock?

.gitlab-ci.yml
```diff
@@ -0,0 +1,118 @@
...
+    - docker-compose up --build -d
+    # wait for db (retry X times)
+    - docker-compose run --rm php bash -c 'while [ true ]; do curl mssql:1433; if [ $? == 52 ]; then break; fi; ((c++)) && ((c==15)) && break; sleep 5; done'
+    - sleep 3
+    # Note: Password has to be the last parameter
+    - docker-compose run --rm sqlcmd sh -c 'sqlcmd -S mssql -U sa -Q "CREATE DATABASE yii2test" -P Microsoft-12345'
+    - docker-compose run --rm php vendor/bin/phpunit -v --group mssql
+    - docker-compose down -v --remove-orphans
+
+travis:
```
cebe: what does this have to do with travis?

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM dmstr/php-yii2:7.0-fpm-1.9-beta2-alpine-nginx
```
cebe: is it possible to move the DockerFiles to some dir in tests/ ? if they need to be located in repo root I would prefer copying them from test folder here before starting docker.

cebe (commented):
Looks great to me except the comments below. Good job, thank you!

cebe:
> @schmunk42 do we have to commit composer.lock?

composer.lock changes look good to me.

cebe:
> In tests, there are dependencies from i.e. log suite to db, which make tests fail which would actually 
> run on a vanilla PHP. I thought about also putting them into group db, WDYT?

Sure, all db related tests should be in the group db, that would also allow excluding them.
`phpunit --exclude=db` should run all tests that are not DB related.

Dockerfile
schmunk42: This would imply that we use another way of getting the framework source-code onto the image, because a Dockerfile defines the build context and you can not add something from outside the build context.

An option is to use host-volumes, but they can problematic regarding performance and other things.
We could also clone or download the current commit when building an image, but that would be annoying in development.

I also don't like it, that these files are in the root folder, but it was the best option to me for now.

I am open for discussion.

.gitlab-ci.yml
schmunk42: Removed, thanks.

schmunk42:
About composer.lock - https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file

You should update it from time to time and always for a release. A developer who wants to run `update` can still do so. You also need it for some CI related tasks, here and there.

tests/framework/console/controllers/DbMessageControllerTest.php
```diff
@@ -1,6 +1,10 @@
 <?php
 namespace yiiunit\framework\console\controllers;
 
+/**
+ * @group db
```
cebe: this must be on the class phpdoc, not here.

cebe (changes requested):

schmunk42:
Current build state:
- tests/mysql https://gitlab.com/yiisoft-contrib/yii2/pipelines/6519636/builds
- tests/pgsql https://gitlab.com/yiisoft-contrib/yii2/pipelines/6519650/builds

schmunk42:
Anything more?

samdark (approved):

SilverFire (approved):

samdark:
I think it's OK to merge. @cebe ?

Dockerfile
cebe: I was thinking about moving the Dockerfile in the build script, e.g. `cp tests/docker/Dockerfile-mysql Dockerfile` and then run docker-compose, so that they are not located in the repo root but are moved to repo root when needed.

schmunk42: Found a way :)

schmunk42: Actually, the one in the root folder could become kind of an official image, I am working with @mikehaertl on something similar.

Or should this be moved to tests also?

cebe: I think it is fine like this now.

cebe (approved):

cebe:
Merged, thank you for this great addition! :)

schmunk42:
You're welcome :)

Auto-pull also works already, see: https://gitlab.com/yiisoft-contrib/yii2/pipelines/6851144
Do these tests also fail in Travis? I need to look that up.

I'll look after the tests from time to time. db, cache and mysql are green AFAIR.

You can push i.e. to `tests/pgsql` on gitlab.com to trigger the PostgreTest suite.

The "exotic" databases still need some testing of the testing...

samdark:
@schmunk42 is there a way to integrate these tests with GitHub? Some failures are about differences in intl/ICU data used. gitlab has newer version. Tests should be adjusted.

schmunk42:
bildschirmfoto 2017-03-06 um 20 40 07

It's set up to sync hourly.

samdark:
Yes, I know. I've meant displaying links and results in pull requests etc.

schmunk42:
Hmm, there's build badge, but further integration is limited AFAIR.
Related: https://gitlab.com/gitlab-org/gitlab-ce/issues/2595

schmunk42:
> is there a way to integrate these tests with GitHub? Some failures are about differences in intl/ICU data used. gitlab has newer version. Tests should be adjusted.

@samdark You could run tests also on Travis, but we didn't want to touch the existing Travis setup. 
Looks like there's no support for multiple files: https://github.com/travis-ci/travis-ci/issues/3540

Could work with a `travis/docker` branch though.

samdark:
I see. I'd let travis alone as it is. It is sometimes weird but that gives additional cases to fix which is good.

schmunk42:
OK, we'll be able to see these issues better, when we have the base images. Since we then can test multiple versions easily.

CC: @mikehaertl

PR #13362, opened 2017-01-10
Changed return value in MemCache::setValues()

masterklavi:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

`Memcached::setMulti()` doesn't return an array of failed keys....ok! But it returns a boolean value, not a void value. So we can and must check success of this operation. For example: if it returns `false` then mark all of the keys as failed. What do you think about it?

### DIFF
--- a/framework/caching/MemCache.php
+++ b/framework/caching/MemCache.php
@@ -304,7 +304,7 @@ class MemCache extends Cache
      * Stores multiple key-value pairs in cache.
      * @param array $data array where key corresponds to cache key while value is the value stored
      * @param int $duration the number of seconds in which the cached values will expire. 0 means never expire.
-     * @return array array of failed keys. Always empty in case of using memcached.
+     * @return array array of failed keys.
      */
     protected function setValues($data, $duration)
     {
@@ -313,9 +313,10 @@ class MemCache extends Cache
             // @see http://php.net/manual/en/memcache.set.php
             // @see http://php.net/manual/en/memcached.expiration.php
             $expire = $duration > 0 ? $duration + time() : 0;
-            $this->_cache->setMulti($data, $expire);
 
-            return [];
+            // Memcached::setMulti() returns boolean
+            // @see http://php.net/manual/ru/memcached.setmulti.php
+            return $this->_cache->setMulti($data, $expire) ? [] : array_keys($data);
         } else {
             return parent::setValues($data, $duration);
         }


samdark:
The empty array was returned because PHP extension isn't really using mset API of memcached but [iterating over data and issuing `set` multiple times](https://github.com/php-memcached-dev/php-memcached/blob/master/php_memcached.c#L1792). Of course, there's no transaction and no rollback so some keys may succeed while another keys may fail. In this case returning all keys passed is not correct.

samdark:
I'd keep it as is.

masterklavi:
> In this case returning all keys passed is not correct.

In my branch the function is returning all keys passed when there aren't any errors in `s_memc_write_zval` calls.
`return $this->_cache->setMulti($data, $expire) ? [] : array_keys($data);`

> Of course, there's no transaction and no rollback so some keys may succeed while another keys may fail.

At `yiisoft:master` branch the function always returns an empty array. Maybe it's better to use another returning value or simply replace `$this->_cache->setMulti()` by `parent::setValues();` **?**

samdark:
@yiisoft/core-developers need additional input.

framework/caching/MemCache.php
```diff
@@ -313,9 +313,10 @@ protected function setValues($data, $duration)
             // @see http://php.net/manual/en/memcache.set.php
             // @see http://php.net/manual/en/memcached.expiration.php
             $expire = $duration > 0 ? $duration + time() : 0;
-            $this->_cache->setMulti($data, $expire);
 
-            return [];
+            // Memcached::setMulti() returns boolean
+            // @see http://php.net/manual/ru/memcached.setmulti.php
+            return $this->_cache->setMulti($data, $expire) ? [] : array_keys($data);
```
SilverFire: Should be explicitly checked with `=== true`

SilverFire (changes requested):
Could be accepted to be more consistent with other cache systems.

framework/caching/MemCache.php
masterklavi: It isn't necessarily because of `Memcached::setMulti()` always returns `boolean`. See https://github.com/php-memcached-dev/php-memcached/blob/master/php_memcached.c#L1840 .

samdark (approved):

samdark:
Merged. Thanks!

PR #13369, opened 2017-01-11
Disabled link at current page

aquy:
linkpager disabled link at current page

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | no

### DIFF
--- a/framework/widgets/LinkPager.php
+++ b/framework/widgets/LinkPager.php
@@ -125,9 +125,10 @@ class LinkPager extends Widget
      */
     public $hideOnSinglePage = true;
     /**
-     * @var bool replace a to span the current page
+     * @var bool whether to render current page button as disabled.
+     * @since 2.0.12
      */
-    public $disableCurrentPageLink = false;
+    public $disableCurrentPageButton = false;
 
 
     /**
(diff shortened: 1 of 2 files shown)


samdark:
Why?

aquy:
link page to itself on the site bad for SEO and the user is already on this page

samdark:
OK. Makes sense.

bashkarev:
Google and others, remove the current link.
Maybe replace `a` to `span`?
And to make it optional.

samdark:
Yes. That is a good idea.

aquy:
I've added the option to replace `a` to `span` - `disabledCurrentPageLink`

framework/widgets/LinkPager.php
```diff
@@ -124,6 +124,10 @@ class LinkPager extends Widget
      * @var bool Hide widget when only one page exist.
      */
     public $hideOnSinglePage = true;
+    /**
+     * @var bool replace a to span the current page
+     */
+    public $disabledCurrentPageLink = false;
```
samdark: `disabledCurrentPageLink` → `disableCurrentPageLink`.

samdark (changes requested):

samdark (approved):

SilverFire (changes requested):
Could you update [tests](https://github.com/yiisoft/yii2/blob/master/tests/framework/widgets/LinkPagerTest.php) to reflect the changes?

aquy:
Yes, tests have been updated

thyseus:
How about making this to be true by default?

samdark:
I think it's OK to make it true by default.

bashkarev:
This can break the page layout

samdark:
How?

bashkarev:
if not use bootstrap
```html
<ul>
    <li><a href="/?page=1">url</a></li>
    <li><span>current</span></li>
</ul>
```
```css
ul > li > a {
    float: left;
    display: block; // etc
}
```

samdark:
Right. Then it's not a good idea :)

SilverFire (approved):

samdark:
Merged. Thanks!

PR #13374, opened 2017-01-11
Fixed autoloading inside BaseYii scope

and800:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Currently the autoloader includes files inside `yii\BaseYii::autoload()` method. That means that the included code shares the scope with that method, including `self` pointing to `BaseYii` and `static` pointing to `Yii`. I've added a workaround which Composer uses: execute `include` statement in separate function.

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -285,7 +285,7 @@ class BaseYii
             return;
         }
 
-        include($classFile);
+        cleanAutoload($classFile);
 
         if (YII_DEBUG && !class_exists($className, false) && !interface_exists($className, false) && !trait_exists($className, false)) {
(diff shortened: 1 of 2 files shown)


cebe:
can you show a case where this causes a problem?

and800:
It's just about cleaner scope, not to confuse with those names in global scope. By the way, when I was creating this PR, I didn't notice that BaseYii is completely public static, so maybe that's really not a serious issue. So the decision is up to you.

cebe:
as there is no problem to fix, this adds too much complexity imo. Thanks for the suggestion but I think we should leave it as is.

PR #13375, opened 2017-01-12
add events before and after BaseActiveRecord::updateAttributes()

thyseus:
| Q             | A
| ------------- | ---
| Is bugfix?    |no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | n/a

Add events before and after BaseActiveRecord::updateAttributes(). This will probably be necessary for https://github.com/bedezign/yii2-audit/issues/183 - or are there any other ideas?


thyseus:
I  just found this in the documentation about #updateAll():

> Note that this method will not trigger any events. If you need [[EVENT_BEFORE_UPDATE]] or
> [[EVENT_AFTER_UPDATE]] to be triggered, you need to [[find()|find]] the models first and then
> call [[update()]] on each of them. For example an equivalent of the example above would be:
> 

What is the reason behind this rationale?

klimov-paul:
In case you wish events to be triggered after ActiveRecord update you should use `save()` method.
Method `ActiveRecord::updateAttributes()` has been extracted exactly to be used without triggering any event.

thyseus:
Thank you. I have adjusted my code to use save();

PR #13384, opened 2017-01-13
Fix #13278: `DbQueryDependency` created

### LINKED ISSUE #13278: DbDependency using QueryBuilder

When working with different sql schema the db dependency should be able to use the query builder to generate the sql statements.

```php
new DbDependency(['sql' => 'SELECT MAX(id) FROM admin_storage_folder WHERE is_deleted=0']);
```

The above example will work in mysql but not pgsql.

### Additional info

Our CMS system provides auto of the box caching mechanism, but db depencies are direct sql statements. It should be possible to provide query build with the query builder in order to provide independent sql schemas.

https://github.com/yiisoft/yii2/blob/master/framework/caching/DbDependency.php#L60

If there is another solution its not yet document in the guide nor api, let me know and will send a PR.  

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 7.x
| Operating system | Ubuntu

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13278

Fixes #13278 adding new `DbQueryDependency` cache dependency


framework/caching/DbQueryDependency.php
```diff
@@ -0,0 +1,109 @@
...
+     */
+    protected function generateDependencyData($cache)
+    {
+        $db = $this->db;
+        if ($db !== null) {
+            $db = Instance::ensure($db);
+        }
+
+        if (!$this->query instanceof QueryInterface) {
+            throw new InvalidConfigException('`' . get_class($this) . '::$query` should be an instance of `yii\db\QueryInterface`.');
```
samdark: Since errors aren't outputted with markdown processing it's better to use regular " instead of `.

samdark (approved):

tests/framework/caching/DbQueryDependencyTest.php
```diff
@@ -0,0 +1,56 @@
...
+            ->limit(1);
+        $dependency->reusable = false;
+
+        $dependency->evaluateDependency($cache);
+        $this->assertFalse($dependency->isChanged($cache));
+
+        $db->createCommand()->insert('dependency_item', ['value' => 'new'])->execute();
+
+        $this->assertTrue($dependency->isChanged($cache));
+    }
```
SilverFire: It will be good to add the following tests:
 - `method` set to `string` and `callback`
 - `query` that is not instance of `yii\db\QueryInterface`

SilverFire (changes requested):
Good in general. CHANGELOG line should be added before merge

SilverFire (approved):

PR #13392, opened 2017-01-15
fixed some comment issue

undefinedor:
| Q             | A
| ------------- | ---
| Is bugfix?    |  no
| New feature?  | no
| Breaks BC?    |  no
| Tests pass?   |  no

just some comment issue

### DIFF
--- a/framework/base/Module.php
+++ b/framework/base/Module.php
@@ -356,9 +356,7 @@ class Module extends ServiceLocator
      * Defines path aliases.
      * This method calls [[Yii::setAlias()]] to register the path aliases.
      * This method is provided so that you can define path aliases when configuring a module.
-     * @property array list of path aliases to be defined. The array keys are alias names
-     * (must start with `@`) and the array values are the corresponding paths or aliases.
-     * See [[setAliases()]] for an example.
+     *
      * @param array $aliases list of path aliases to be defined. The array keys are alias names
      * (must start with `@`) and the array values are the corresponding paths or aliases.
      * For example,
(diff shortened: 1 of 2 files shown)


samdark:
Thank you!

framework/base/Module.php
```diff
@@ -356,9 +356,7 @@ protected function defaultVersion()
      * Defines path aliases.
      * This method calls [[Yii::setAlias()]] to register the path aliases.
      * This method is provided so that you can define path aliases when configuring a module.
-     * @property array list of path aliases to be defined. The array keys are alias names
-     * (must start with `@`) and the array values are the corresponding paths or aliases.
-     * See [[setAliases()]] for an example.
```
cebe: why did you remove this? It totally belongs here...

cebe:
reverted: c1c11d0, @yiisoft/core-developers FYI.

framework/base/Module.php
undefinedor: it's duplicated.
```
/**
     * Defines path aliases.
     * This method calls [[Yii::setAlias()]] to register the path aliases.
     * This method is provided so that you can define path aliases when configuring a module.
     * @property array list of path aliases to be defined. The array keys are alias names
     * (must start with `@`) and the array values are the corresponding paths or aliases.
     * See [[setAliases()]] for an example.
     * @param array $aliases list of path aliases to be defined. The array keys are alias names
     * (must start with `@`) and the array values are the corresponding paths or aliases.
     * For example,
     *
     * ```php
     * [
     *     '@models' => '@app/models', // an existing alias
     *     '@backend' => __DIR__ . '/../backend',  // a directory
     * ]
     * ```
     */
```
 Don't you think it's duplicated?

cebe: No, see my comment in the commit that reverts it. This annotation is to generate the `@property ` annotation in class PHPdoc for this setter.

undefinedor: I see, it's used to generate annotation.  You are right .

PR #13402, opened 2017-01-18
allow multiple sql queries in DbDependency

thyseus:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes - PHP5.5 wrongly configured on travis

And also a little bit of refactoring.

I know that \yii\caching\ChainedDependency exists, but this way is much more convinient:

```php
 'dependency' => [
        'class' => 'yii\caching\DbDependency',
        'sql' => [
           'SELECT updated_at from agency where user_id = ' . Yii::$app->user->id,
           'SELECT max(updated_at) from offer_has_agency where agency_id = ' . Yii::$app->user->identity->agency->id,
           'select count(*) from message where message.status = 1 and message.to = ' . Yii::$app->user->id,
        ],
    ],
```


SilverFire:
Thank you for your idea, but I have some objection:

- The change makes `DbDependency` class name irrelevant, because it will represent multiple dependencies at the same time.
- There is already a way to get the same result without code changing and it looks pretty much the same
- The change complicates the code in `DbDependency`

I'd like to hear more @yiisoft/core-developers opinions

samdark:
I agree that the change should not be accepted. Reasons:

1. Complicates implementation.
2. Another way to the same thing.
3. Adds an easy way to kill application performance by executing multiple queries each time cache is accessed.

Issue #13344, opened 2017-01-07
AccessControl behavior "role" could be confusing 


gvasilopulos:
### What steps will reproduce the problem?
there is no problem just a litte confusing implementation 
### Additional info

taken from the Yii2 guide:

```php
public function behaviors()
{
    return [
        'access' => [
            'class' => AccessControl::className(),
            'rules' => [
                [
                    'allow' => true,
                    'actions' => ['index'],
                    'roles' => ['managePost'],
                ],
```

It seems that 'roles'=>['managePost'] really refers to permission
so it may be 'permissions'=>['managePost']
to avoid confusion.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 5.6
| Operating system |

cebe:
Yeah, according to RBAC checks should only be made against permissions so it would make sense to rename that for better user education.

samdark:
Agree that it should be renamed.

samdark:
I'm for supporting both keys in 2.1.x and dropping `roles` in 2.1.

bizley:
Maybe some other name is needed here, something that means (or concerns) "roles+permissions" because in 2.1 expect issue like: 
> It seems that 'permissions'=>['?', 'admin'] really refers to roles

I suggest "**applyTo**" or "**check**".

samdark:
`only`?

cebe:
"only" is too generic, we have "actions", "controller", "ips"... then "only"-what?

SilverFire:
How about `can`? It is alike `Yii::$app->user->can()`

> I'm for supporting both keys in 2.1.x and dropping roles in 2.1.

IMHO it's not the change that requires such a massive BC breaking. I'm OK even with having both permanently.

samdark:
`can` doesn't sound descriptive. The access rule itself has `allow` which is very close to `can`.

bizley:
What about `privileges`? It sounds descriptive for both roles and permissions: "admin privileges", "editing privileges".

thyseus:
Actually, you can pass an array of roles, of permissions or even mixed here.

So i suggest that 'roles' and 'permissions' should be aliases and be both valid even in yii 2.1, as long as only one of them are specified. 

Roles and Permissions are both valid entities here.

This way we also stay backward-compatible.

cebe:
[...]

PR #13403, opened 2017-01-18
allow 'permissions' as alias for 'roles' in filters\AccessRule

thyseus:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes

PR proposal for discussion in https://github.com/yiisoft/yii2/issues/13344


samdark (approved):

framework/filters/AccessRule.php
```diff
@@ -136,6 +142,10 @@ protected function matchController($controller)
      */
     protected function matchRole($user)
     {
+        if (!empty($this->permissions)) {
+            $this->roles = array_merge($this->roles, $this->permissions);
```
SilverFire: I don't think we should affect the class state at all.
In case the rule runs more than once, the `roles` property will have N-as-many values.

```php
if (!empty($this->permissions)) {
    $roles = array_merge($this->roles, $this->permissions);
}
// then use $roles instead of $this->roles
```

samdark: Agree.

samdark: Adjusted code.

samdark (changes requested):
New change looks not so good: two array_merge instead of just one...

thyseus:
The former method made problems when 'permissions' are set but not 'roles'

thyseus:
How about 

```
foreach(['roles', 'permissions'] as $var)
  if(!empty($this->$var)
    array_merge(...
```
just as an idea ?

thyseus:
I also am not sure if the original problem in #13344 is handled with this PR. Some people still want to find one single name that handles every situation.

samdark:
@thyseus feel free to create a poll.

framework/filters/AccessRule.php
```diff
@@ -136,19 +142,30 @@ protected function matchController($controller)
...
+        if (empty($this->roles) && empty($this->permissions)) {
             return true;
         }
-        foreach ($this->roles as $role) {
-            if ($role === '?') {
+
+        $privileges = [];
+
+        if($this->roles) {
+            $privileges = array_merge($privileges, $this->roles);
```
samdark: You can do `$roles = ($this->roles === null) ? [] : $this->roles` and then merge permissions if they are not `null`:


```php
$roles = empty($this->roles) ? [] : $this->roles;
if (!empty($this->permissions)) {
   $roles = array_merge($roles, $this->permissions);
}
```

samdark (changes requested):

SilverFire (approved):

samdark:
https://github.com/yiisoft/yii2/pull/13403#discussion_r97278588 is still valid.

SilverFire:
yes, sure

samdark (approved):

framework/filters/AccessRule.php
```diff
@@ -136,19 +142,26 @@ protected function matchController($controller)
      */
     protected function matchRole($user)
     {
-        if (empty($this->roles)) {
+        $privileges = empty($this->roles) ? [] : $this->roles;
+
```
dynasource: the introduction of the new 'variable' `$privileges` and the fact that it is assembled by a merge of $roles & $permissions is a bit too magical IMO. 

Ideally, this should be replaced with an OO approach

dynasource (changes requested):

framework/filters/AccessRule.php
samdark: Could you elaborate about OO approach?

dynasource: these arrays are quite vague. Using objects will improve one's understanding of $roles & $permissions and their mutual relation.

samdark: Does it belong to current pull request in any way?

framework/filters/AccessRule.php
```diff
@@ -136,19 +142,26 @@ protected function matchController($controller)
      */
     protected function matchRole($user)
     {
-        if (empty($this->roles)) {
+        $privileges = empty($this->roles) ? [] : $this->roles;
```
samdark: `$privileges` could be renamed to `$items` to be consistent with RBAC implementation.

samdark (changes requested):

tom--:
While I understand that "roles" is not the best name here, I do not like this PR because I don't like the idea of an alias property, this one in particular, or how it works.

- It's a bit confusing just on that basis because it begs the question "Why are there two of these?"

- It sets a bad example. Two properties for the exact same thing isn't good API design.

- Having both compounds the existing confusion (["role" could be confusing](https://github.com/yiisoft/yii2/issues/13344)) because a role and a permission are clearly different things. (I am a janitor and my permissions include unlocking doors.) Saying one is an alias of the other in the `@var` contradicts this.

- The logical concept of an alias is, afaik, undefined. All the explanation we get is "alias for $roles". How is the user to understand it? Should I use one, or the other, or both? The code reveals the answer but that's not ideal.

- Do we use aliases like this anywhere else in the framework?

I know that people can have difficulty with the concepts in access control but I don't think this is a smart approach to helping them nor do I think it will help very much.

samdark:
@tom-- I still think this pull request makes sense. When using RBAC to check for access, one should specify permissions, not roles. Therefore, I'd remove specifying `roles` altogether but since there are backwards compatibility constraints we deprecate it instead.

samdark:
Still, you have a point, `@` and `?` aren't exactly permissions...

tom--:
There's another domain besides RBAC and @/?. AccessRule and matchRole() are quite flexible when you extend the class.

tom--:
On a side note, the foreach in matchRule() gives me vertigo. Is it the same as this?

```
        foreach ($privileges as $privilege) {
            if (($privilege === '?' && $user->getIsGuest())
                || ($privilege === '@' && !$user->getIsGuest())
                || $user->can($privilege)
            ) {
                return true;
            }
        }
```

samdark:
I think it makes sense now with docs adjusted. @tom-- ?

SilverFire (approved):

SilverFire:
Could go with fixed tests

samdark (approved):

samdark:
Merged. Thank you!

PR #13411, opened 2017-01-19
Fixed #13407 - Added URL-safe base64 encode/decode methods to StringHelper

### LINKED ISSUE #13407: FR/ENH: Have url safe base64Encode and base64Decode
IMO, should be part of `StringHelper`

```php
public static function base64Encode($input)
{
    return str_replace('=', '', strtr(base64_encode($input), '+/', '-_'));
}
```

andrewnester:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13407

Implemented according to RFC 4648
https://tools.ietf.org/html/rfc4648#page-7


samdark:
I think we have alike code used so worth checking if we can reuse helpers there.

andrewnester:
@samdark I found similar usage here https://github.com/yiisoft/yii2/blob/master/framework/base/Security.php#L564
gonna to change it to use new StringHelper methods

samdark:
The conversion is different there and that's security component so I'd not be that certain about changing it...

andrewnester:
@samdark Security class and certain method is covered wth unit tests. In addition, introduced changes just changes the order of characters replacement: in first case characters has been replaced in substring, in new one - they are replaced in full string.
But it's resulted in absolutely identical strings

samdark:
Looked more into that. First of all, RFC 4648 doesn't state to replace padding `=` with `%`. It says that `=` should be encoded when used in URLs so it's a special case which is typically achieved separately via `urlencode()` and it's not just `%` but `%3D` in the case.

samdark:
Also, security encoded data is definitely not the same.

Security replacements:

```
+ → _
/ → -
```


Helper replacements:

```
+ → -
/ → _
= → %
```

So

```
c3Viam+jdHM/X2Q9MQ — original
c3Viam_jdHM-X2Q9MQ — security
c3Viam-jdHM_X2Q9MQ — helper
```

andrewnester:
@samdark thanks for the comments.
Agree with = replacement, just have 1 question - should we call `urlencode` inside helper method or it will be done explicitly later if needed?

As to security component: yeah, you're right about the order. Just one question left - was different characters replacement done intentionally there? If not, maybe it's mistyping that should be fixed?

samdark:
I'd not call `urlencode()` right away since it's needed if string is used in URLs only. If you're inserting it in an email or HTML link with `Url` or `Html` it's done automatically.

samdark:
@tom-- need your input about chars replacement.

tom--:
`generateRandomString()` is not an encoder. It returns `$length` 5-bit randoms, each encoded using one of the quasi-standard base-64 alphabets, in the form of a PHP string.

The implementation achieves this by truncating a base64 string that's longer than `$length` so that any base64 padding is removed from the longer string. So any base64 encoder that puts its padding at the end will do for this case.

Second, given the uniform and independently distributed property (we have to assume) of each random bit in the initial random string `$bytes`, any 5bit → [A-Za-z0-9_-] encoder using one permutation of that alphabet is equivalent to any other using the same alphabet.

So if I correctly understand how this new string helper method works, it makes no difference if we use it or the current encoder.

To I leave it up to the core Yii peeps to decide for this case. I can think of arguments pro and contra. In my opinion, its wise to adopt the general stance to not change code that works, especially given the common desire to improve things or make them more how we would like, so @samdark's hesitance here is reasonable. On the other hand, Yii's core code is, like it or not, didactic and I think that's a good thing, so showing how to use your own helpers is reasonable.

framework/helpers/BaseStringHelper.php
```diff
@@ -305,4 +305,30 @@ public static function normalizeNumber($value)
 
         return $value;
     }
+
+    /**
+     * Returns base64-encoded string for $input string with url safe alphabet.
+     * It refers to the following RFC chapter https://tools.ietf.org/html/rfc4648#page-7
```
tom--: Suggest instead:

     * Implements the "Base 64 Encoding with URL and Filename Safe Alphabet" of RFC 4648

framework/helpers/BaseStringHelper.php
```diff
@@ -305,4 +305,30 @@ public static function normalizeNumber($value)
...
+     * @since 2.0.11
+     */
+    public static function base64UrlEncode($input)
+    {
+        return strtr(base64_encode($input), '+/=', '-_%');
+    }
+
+    /**
+     * Returns decoded string for base64-encoded string with url safe alphabet.
+     * It refers to the following RFC chapter https://tools.ietf.org/html/rfc4648#page-7
```
tom--: same comment as ⤴︎

tests/framework/helpers/StringHelperTest.php
```diff
@@ -264,4 +264,39 @@ public function testWordCount()
...
+     * @param $input
+     * @param $base64UrlEncoded
+     */
+    public function testBase64UrlEncode($input, $base64UrlEncoded)
+    {
+        $encoded = StringHelper::base64UrlEncode($input);
+        $this->assertEquals($base64UrlEncoded, $encoded);
+
+        $safeUrlStringRegEx = sprintf('/[a-zA-Z0-9\-_%%]{%d}/', strlen($encoded));
+        $this->assertRegExp($safeUrlStringRegEx, $encoded);
```
tom--: we already know that `$base64UrlEncoded` equals `$encoded`. so what does this additional assertion accomplish?

It seems to be testing the test vector, not the unit under test.

samdark:
Thought more about it. I think usage in Security helper is fine. @tom-- thanks for input.

tom--:
OT but fascinating. https://www.youtube.com/watch?v=C6JKQbOvkgI shows using assertion satisfiability testing on real-world RNG code such as OpenSSL. It only tests one way in which an RNG implementation can fail but it's an important one.

The talk also explains how statistical testing of a CPRNG is pointless, which I needed to be told.

tests/framework/helpers/StringHelperTest.php
andrewnester: @tom-- yeah, you're right, this assertion is excess, thank you!

andrewnester:
@samdark @tom-- guys, I updated PR according to our conversation, thanks! if you have any questions - let me know

samdark (approved):

SilverFire (approved):

dynasource (approved):

samdark:
Thank you!

cebe:
Why are we ignoring the convertion of `=` in the string helper? it does not matter in Security class but it is relevant if we use base64 as an encoder/decoder....

framework/helpers/BaseStringHelper.php
```diff
@@ -305,4 +305,30 @@ public static function normalizeNumber($value)
...
         return $value;
     }
+
+    /**
+     * Encodes string into "Base 64 Encoding with URL and Filename Safe Alphabet" (RFC 4648)
+     * @see https://tools.ietf.org/html/rfc4648#page-7
+     *
+     * @param string $input
+     * @return string
+     * @since 2.0.11
```
cebe: its 2.0.12, not 2.0.11...

tom--:
@cebe it's not that padding is *ignored* but that `=` is not transparent in percent encoding. But this is a feature of RFC 4648. If string length is known, padding is optional, otherwise use `=`, as these methods do.

Other than doing something non-standard like Y64, I don't see an alternative.

framework/helpers/BaseStringHelper.php
samdark: https://github.com/yiisoft/yii2/commit/0b36b27caf09caa6ff53de12cd79a08631dcf780

cebe:
While this function works according to RFC it still produces output that can not be safely used in URLs and needs encoding. This should at least be added as a note to the method as I would a function that has `urlEncode` in its name to return something that is fullly urlencoded.

tom--:
It's a fair point, cebe. We should mention in the docs for `base64UrlEncode` 

     * Encodes string into "Base 64 Encoding with URL and Filename Safe Alphabet" (RFC 4648)
     * Note: Base 64 padding characters `=` may be at the end of the returned string. `=` is
     * not transparent to URL encoding.
     * @see https://tools.ietf.org/html/rfc4648#page-7

Issue #13425, opened 2017-01-22
Cache dynamically added rules with UrlManager::addRules()


yuriystovbur:
Compiled URL rules will be cached only on UrlManager::init() stage.
Additional compiled URL rules are not cached.

SilverFire:
Could you explain your use case?

scriptcube:
@SilverFire, URL rules can be dynamically added to the [[yii\web\UrlManager|URL manager]] https://github.com/yiisoft/yii2/blob/master/docs/guide/runtime-routing.md#user-content-adding-rules-dynamically-

But URL rules added by UrlManager::addRules() will be compiled and not will be cached.

```php
    public function addRules($rules, $append = true)
    {
        if (!$this->enablePrettyUrl) {
            return;
        }
        $rules = $this->buildRules($rules);
        if ($append) {
            $this->rules = array_merge($this->rules, $rules);
        } else {
            $this->rules = array_merge($rules, $this->rules);
        }
    }
```

Compiled URL rules will be cached only on UrlManager::init() stage:

```php
    /**
     * Initializes UrlManager.
     */
    public function init()
    {
        //...
        if ($this->cache instanceof Cache) {
            $cacheKey = $this->cacheKey;
            $hash = md5(json_encode($this->rules));
            if (($data = $this->cache->get($cacheKey)) !== false && isset($data[1]) && $data[1] === $hash) {
                $this->rules = $data[0];
            } else {
                $this->rules = $this->buildRules($this->rules);
                $this->cache->set($cacheKey, [$this->rules, $hash]);
            }
        } else {
            $this->rules = $this->buildRules($this->rules);
        }
    }
```

SilverFire:
Rules caching requires building, serialization and setting to the cache:

```php
$hash = md5(json_encode($this->rules));
$this->rules = $this->buildRules($this->rules);
$this->cache->set($cacheKey, [$this->rules, $hash]);
```

In case you add your rules dynamically, this operation will be performed for each rule. Of course it will happen only once and then you will have all intermediate steps stored in your cache, but I doubt it's a right way.

Why can't you pre-configure the URL manager instead of adding rules dynamically?

scriptcube:
I and others developers use a modular application where each module adds its URL rules.
If many loadable modules in the application, use of cache for URL Rules is more productive.

SilverFire:
> each module adds its URL rules

It could be done with configuration merging instead of bootstrapping.

[...]

PR #13426, opened 2017-01-22
Cache dynamically added rules with `yii\web\UrlManager::addRules()`

yuriystovbur:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13425 

#13425 Cache dynamically added rules with `yii\web\UrlManager::addRules()`


SilverFire (changes requested):
Could you create a separate method instead of code copy-pasting?

scriptcube:
@SilverFire , this is not quite copy-pasting). Maybe create a separate method is not entirely suitable?

in UrlManager::init():
```php
$cacheKey = $this->cacheKey;
    $this->rules = $data[0];
} else {
    $this->rules = $this->buildRules($this->rules);
    $this->cache->set($cacheKey, [$this->rules, $hash]);
}
```
in Urlmanager::addRules():
```php
$cacheKey = $this->cacheKey . ':' . md5(json_encode($rules));
if (($data = $this->cache->get($cacheKey)) !== false) {
    $rules = $data;
} else {
    $rules = $this->buildRules($rules);
    $this->cache->set($cacheKey, $rules);
}
```

rob006:
This will store each combination of rules in cache for one year - this is really long time. I think that using 24-hour duration for cache will be more practical and will not degrade performance that much.

scriptcube:
@rob006 , there is no problem in that the cache value will be stored for a long time. whenever you update the URL rules group, will be the new cacheKey:
 `$cacheKey = $this->cacheKey . '#' . md5(json_encode($rules));`

rob006:
> there is no problem in that the cache value will be stored for a long time

This is a problem - if I change rules 100 times, I will get 99 useless entries in cache. For one year they will take place and slow down some operations, for no reason. And this can be even more problematic if rules will be generated dynamically based on database data - this can change hundreds of times per month.

scriptcube:
UrlManager can configure with a different Сache object in which $defaultDuration = 86400

samdark:
@scriptcube that depends on the project.

scriptcube:
@SilverFire , Now caching occurs in `yii\web\UrlManager::buildRules()`. I hope it is more correct option)

dynasource:
> This is a problem - if I change rules 100 times, I will get 99 useless entries in cache. For one year they will take place and slow down some operations, for no reason. And this can be even more problematic if rules will be generated dynamically based on database data - this can change hundreds of times per month.

@rob006, sharply noticed :+1: 

This points to a fundamental problem with UrlManager in the first place. Why should rules be continiously built and not a the moment of their first use?

It would be more logical to keep configuration & instantiation separate and to do the caching 'once' on arrays instead of on instances, which is more clean & stable.

SilverFire:
I've already asked this question:
https://github.com/yiisoft/yii2/issues/13425#issuecomment-274341903

The problem is that we do not have a unified way to use configs from `yii2-extensions` in core, except bootstrapping or global event handling.

SilverFire:
After a year of thinking (😉) decided that in case when there are 100 `addRules()` calls, getting 99 useless cache items should not harm against 99 rules builds.

SilverFire:
@rob006 I see your thumbs down and I've seen your point above.

> if rules will be generated dynamically based on database data - this can change hundreds of times per month

For me this case does not look real. Do you have a good idea why should anybody ever do this when we have route params?

> For one year they will take place and slow down some operations, for no reason

There is a common step in release: drop caches to prevent data structure BC issues and clear garbage, ofc.

rob006:
> For me this case does not look real. Do you have a good idea why should anybody ever do this when we have route params?

Actually, I've done something similar in one app - rules are created dynamically based on the list of services in multisite CMS, as a combination of websites slugs and supported languages. New site = a new set of rules. New rules modify old ones, so with solution from this PR I would keep this old useless cache for one year.

> There is a common step in release: drop caches to prevent data structure BC issues and clear garbage, ofc.

This has nothing to do with the release - I can have a working app for 2 years with no new release, but with lots of rules generated dynamically. You can't rely on releases to clear trash from app.

On the other hand, I see no reason why not set cache duration to 24 hours. Building rule once per 24-hours should not give any noticeable performance impact and will avoid storing outdated cache for a long time.

cebe:
In most cases URL rules are mainly static and do not change often, so a long value for caching these is okay. Setting an arbitrary value (24hours) as the default does not really make sense imo, a good value depends on how much they really change, so this has to be fine-tuned on a per application basis.
Imo we can keep using the `$defaultDuration` setting of the cache and improve the documentation about it. Added a note in ef8f617.

PR #13432, opened 2017-01-23
Fixed PHPDoc

s1lver:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | no

Return value is expected to be 'ActiveQuery', 'object' returned. Return value type is not compatible with declared. And PHPDoc comment doesn't contain all necessary @throws tag(s).


samdark:
It certainly does not return just object...

samdark:
Also `\Throwable` is too general.

Can't accept this pull request.

PR #13434, opened 2017-01-23
fix: module version bug

programer19:
Couldn't reassign version of module by setting $_version property.
defaultVerison was called anyway.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no


klimov-paul:
This is incorrect. If you wish your module to provide specific default version value, you should override `defaultVersion()` method as docs suggets:
https://github.com/yiisoft/yii2/blob/master/framework/base/Module.php#L343

PR #13441, opened 2017-01-24
Removes redunant code and allows override query relation in descendan…

leandrogehlen:
| Q             | A
| ------------- | ---
| Is bugfix?    |  no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Sometimes it's necessary to change default behavior for relation queries.

Example:

I have created a `MultiTenantRecord` where `ActiveRecord::find()` must be overrided, like:

```php

public static function find()
{
    $identity = Yii::$app->user->identity;
    if ($identity instanceof TenantInterface) {
        return parent::find()->where([static::tableName().'.tenant_id' => $identity->getTenantId()]);
    } else {
        throw new NotSupportedException("Identity does not implements TenantInteface");
    }
}
```
but this should be occurs only "master" record not for relation.
Then this allows customize query for relation

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -371,13 +371,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      */
     public function hasOne($class, $link)
     {
-        /* @var $class ActiveRecordInterface */
-        /* @var $query ActiveQuery */
-        $query = $class::find();
-        $query->primaryModel = $this;
-        $query->link = $link;
-        $query->multiple = false;
-        return $query;
+        return $this->newRelationQuery($class, $link, false);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


framework/db/BaseActiveRecord.php
```diff
@@ -412,12 +406,24 @@ public function hasOne($class, $link)
...
+    }
+
+    /**
+     * Creates a query for `has-one` and `has-many` relation.
+     * @param string $class the class name of the related record
+     * @param array $link the primary-foreign key constraint
+     * @param boolean whether this query represents a relation to more than one record.
+     * @return ActiveQueryInterface the relational query object.
+     */
+    protected function newRelationQuery($class, $link, $multiple)
```
dynasource: this should be called `createRelationQuery` and if we want to be consistent even `createLinkQuery`

dynasource (changes requested):
this is useful for further customizing relations within ActiveRecords

framework/db/BaseActiveRecord.php
leandrogehlen: 👍 , @dynasource  i think better `createRelationQuery`

framework/db/BaseActiveRecord.php
```diff
@@ -412,12 +406,24 @@ public function hasOne($class, $link)
...
     {
+        return $this->newRelationQuery($class, $link, true);
+    }
+
+    /**
+     * Creates a query for `has-one` and `has-many` relation.
+     * @param string $class the class name of the related record
+     * @param array $link the primary-foreign key constraint
+     * @param boolean whether this query represents a relation to more than one record.
+     * @return ActiveQueryInterface the relational query object.
```
SilverFire: `@since 2.0.12`

SilverFire (changes requested):
composer.lock should not be comminted

leandrogehlen:
Sorry, i removed my fork accidentally, i will redo the changes in another PR

Issue #11502, opened 2016-05-04
Newly added languages to message config.php are not working when using yii\i18n\DbMessageSource


Tarekajaj:
### What steps will reproduce the problem?

Add i18n to application components

```
'i18n' => [
    'translations' => [
        '*' => [
            'class' => 'yii\i18n\DbMessageSource',
            'sourceLanguage' => 'en'
        ],
    ],
],
```

create **messages-config.php** file using `yii message/config-template messages-config.php`
configure messages-config.php to use **db** and to translate to **'ar'**

```
return [
    'sourcePath' => __DIR__ . DIRECTORY_SEPARATOR . '../..',
    'languages' => ['ar'],
    'translator' => 'Yii::t',
    'sort' => false,
    'removeUnused' => true,
    'markUnused' => false,
    'only' => ['*.php'],
    'except' => [
        '.svn',
        '.git',
        '.gitignore',
        '.gitkeep',
        '.hgignore',
        '.hgkeep',
        '/messages',
        '/vendor'
    ],    
    'format' => 'db',
    'db' => 'db',
    'sourceMessageTable' => '{{%source_message}}',
    'messageTable' => '{{%message}}',
];
```

now run `yii message messages-config.php`

The tables in the database gets populated with no problem **source_message** contains the messages and their categories and **message** contains foreach source message a row containing the language **ar** and an empty translation (all good till now)

now go back to messages-config.php and add **fr** to the list of languages `'languages' => ['ar', 'fr'],` and run `yii message messages-config.php` you get a message at the end `Inserting new messages...Nothing new...skipped.` and after looking at the database no new row in **message** table is added with **fr** as language
### What is the expected result?

I expected that, for each source message, a new row in the **message** table is added with newly added language to the configuration file and/or translated messages, for any removed language, are deleted from the **message** table
### Additional info

If we add both languages  `'languages' => ['ar', 'fr'],` before the first `yii message messages-config.php` then foreach source message, two rows in the message tables are created as expected.

| Q | A |
| --- | --- |
| Yii version | 2.0.8 |
| PHP version | 5.6.3 |
| Operating system | Windows 7 Ultimate 64-bit |

Best Regards,

Tarek K. Ajaj

PR #13446, opened 2017-01-25
Populating missing languages

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #11502

- `MessageController` now properly populates missing languages in case of extraction with "_db_" format and no new messages.  
  Previously with no new messages to extract missing languages were ignored.
- Unit tests for extracting messages with "_db_" format added.
- Test checking case #11502 added in `BaseMessageControllerTest`.
- `testMarkObsoleteMessages()` had to be overridden because in case of DB messages source message is wrapped in `@@` and not translation.

Additionally I've made few slight modifications in `MessageController`:
- Excessive `else` has been removed in explicit cases.
- Failure to create language directory throws exception now in case of "_php_" and "_po_" formats.
- Db component is verified with `Instance::ensure()` to allow tests configuration.
- Single-character variable `$m` renamed to keep Code Climate happier.
- Console output "_Nothing new...skipped_" replaced with "_Nothing to save_" to make it compatible with other message saving methods and `BaseMessageControllerTest`.

`DbMessageControllerTest` is based on SilverFire's code for `DbMessageSourceTest`.


framework/console/controllers/MessageController.php
```diff
@@ -311,24 +315,23 @@ public function actionExtract($configFile = null)
...
         }
+
+        $catalog = isset($config['catalog']) ? $config['catalog'] : 'messages';
+
         if (in_array($config['format'], ['php', 'po'])) {
             foreach ($config['languages'] as $language) {
                 $dir = $config['messagePath'] . DIRECTORY_SEPARATOR . $language;
-                if (!is_dir($dir)) {
-                    @mkdir($dir);
+                if (!is_dir($dir) && !mkdir($dir)) {
```
samdark: `@` should not be removed since `mkdir` not only fails but triggers warnings.

framework/console/controllers/MessageController.php
```diff
@@ -359,11 +361,20 @@ public function actionExtract($configFile = null)
      */
     protected function saveMessagesToDb($messages, $db, $sourceMessageTable, $messageTable, $removeUnused, $languages, $markUnused)
     {
-        $q = new \yii\db\Query;
-        $current = [];
+        $currentMessages = [];
+        $query = (new Query)->select(['id', 'category', 'message'])->from($sourceMessageTable);
+        foreach ($query->each(100, $db) as $row) {
```
samdark: Does that help to improve performance or memory usage?

samdark (changes requested):

framework/console/controllers/MessageController.php
bizley: Right, probably not in here. Old AR habit.

samdark: I guess that in this case it can actually decrease performance...

bizley:
Should I modify CHANGELOG to 2.0.12 as well?

samdark:
Yes, I think so. No more issues are going to me solved in 2.0.11 except the ones assigned to milestone else we won't release it ever :)

samdark (approved):

framework/console/controllers/MessageController.php
```diff
@@ -35,7 +37,7 @@
  * @author Qiang Xue <qiang.xue@gmail.com>
  * @since 2.0
  */
-class MessageController extends Controller
+class MessageController extends \yii\console\Controller
```
andrewnester: maybe it's better to add `use yii\console\Controller` statement?

bizley: It's used only once so I made it like with other similar Yii 2 classes cases.

samdark:
Merged https://github.com/yiisoft/yii2/commit/1749c46c5d7a14a2c45fbad5f723edbbca37db34 Thank you!

PR #13449, opened 2017-01-26
Fixed #13378 - Added skipOnEmpty option to SluggableBehaviour

### LINKED ISSUE #13378: SluggableBehavior and empty $attribute value

```php
    /**
     * @return array
     */
    public function behaviors()
    {
        $behaviors = parent::behaviors();
        $behaviors['slug'] = [
            'class' => SluggableBehavior::className(),
            'attribute' => 'title',
            'slugAttribute' => 'code',
            'ensureUnique' => true,
            'immutable' => true
        ];

        return $behaviors;
    }
```

SluggableBehavior should not runs if model `title` attribute is empty. In example above if value of `title` attribute is not set, value of `code` attribute should not changes.

### What do you get instead?
In example above if value of `title` attribute is not set, value of `code` attribute sets to '-2'.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10?
| PHP version      | 7.0.13
| Operating system | Debian 8.1

andrewnester:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13378


tests/framework/behaviors/SluggableBehaviorTest.php
```diff
@@ -229,4 +245,20 @@ public function behaviors()
             ],
         ];
     }
-}
\ No newline at end of file
+}
+
+class ActiveRecordSluggableSkipOnEmpty extends ActiveRecordSluggable
+{
```
dynasource: the naming should have ActiveRecord at the end (ActiveRecordSluggable actually too IMO). So a better name is `SkipOnEmptySluggableActiveRecord'. Perhabs you can change that too.

dynasource (changes requested):

andrewnester:
@dynasource thanks! I've just updated my PR.

framework/behaviors/SluggableBehavior.php
```diff
@@ -144,7 +150,9 @@ protected function getValue($event)
                 foreach ((array) $this->attribute as $attribute) {
                     $slugParts[] = ArrayHelper::getValue($this->owner, $attribute);
```
samdark: Empty check and return could be done right here so there's no need to iterate array one more time.

samdark:
```php
$part = ArrayHelper::getValue($this->owner, $attribute);
if ($part === null || $part === '') {
    return $this->owner->{$this->slugAttribute};
}
$slugParts[] = $part;
```

samdark (changes requested):

framework/behaviors/SluggableBehavior.php
andrewnester: @samdark thanks! looks like it would be good change for performance. done!

samdark (approved):
Looks good. Please add a changelog entry.

framework/behaviors/SluggableBehavior.php
```diff
@@ -142,9 +148,12 @@ protected function getValue($event)
             if ($this->isNewSlugNeeded()) {
                 $slugParts = [];
                 foreach ((array) $this->attribute as $attribute) {
-                    $slugParts[] = ArrayHelper::getValue($this->owner, $attribute);
+                    $part = ArrayHelper::getValue($this->owner, $attribute);
+                    if ($this->skipOnEmpty && $this->isEmpty($part)) {
+                        return $this->owner->{$this->slugAttribute};
```
klimov-paul: I am unsure if this logic is correct:
- if there is no slug (e.g. for new record) and its source is empty - the slug will be empty.
- if slug is set (existing record) and its source is empty (but changed!) - the slug remains unchanged, which seems odd

samdark: I'd expect that. Your first point is how it behaves now. Usually if you're using a field to generate a slug the field is required.  Second point is what this property was introduced for.

dynasource (approved):

samdark:
Merged. Thank you!

rob006:
Test are broken: https://travis-ci.org/yiisoft/yii2/builds/249564948

Issue #13451, opened 2017-01-26
X_FORWARDED_PROTO with multiple proxy steps not supported


bpicolo:
### What steps will reproduce the problem?
Call Yii::$app->getRequest()->getIsSecureConnection() when $_SERVER['HTTP_X_FORWARDED_PROTO'] = 'https,http'
### What is the expected result?
true

### What do you get instead?
false

### Additional info
X_FORWARDED_PROTO is a de-facto rather than a canonical http field. It seems the common definition is to add, as comma-separated values, the different protocols used when multiple proxies lie in front of your application. (See https://en.wikipedia.org/wiki/X-Forwarded-For#Format)

I think it makes the most sense for the connection to be considered secure if the Edge node (first step in the chain) is secure, because that's the most common use case. An example would be terminating SSL at an external load balancer / cdn, and forwarding over HTTP to an internal load-balancer.

samdark:
Already fixed by #13780

PR #13452, opened 2017-01-26
Support HTTP_X_FORWARDED_PROTO through multiple proxy levels

bpicolo:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes/no   Depends on the behavior people might be relying on? Doesn't break any well-defined behavior.
| Tests pass?   | yes, + new test added
| Fixed issues  | #13451

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -874,6 +874,8 @@ class Request extends \yii\base\Request
     public function getIsSecureConnection()
     {
         return isset($_SERVER['HTTPS']) && (strcasecmp($_SERVER['HTTPS'], 'on') === 0 || $_SERVER['HTTPS'] == 1)
+            // If we've gone through multiple proxies, only check the edge node.
+            // https://github.com/yiisoft/yii2/issues/13451
             || isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && strncasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https', 5) === 0;
     }
 
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -874,7 +874,7 @@ class Request extends \yii\base\Request
     public function getIsSecureConnection()
     {
         return isset($_SERVER['HTTPS']) && (strcasecmp($_SERVER['HTTPS'], 'on') === 0 || $_SERVER['HTTPS'] == 1)
-            || isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && strcasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') === 0;
+            || isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && strncasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https', 5) === 0;
     }
 
     /**
--- a/tests/framework/web/RequestTest.php
+++ b/tests/framework/web/RequestTest.php
@@ -301,6 +301,13 @@ class RequestTest extends TestCase
     {
         $request = new Request();
 
+        $_SERVER['HTTPS'] = 'on';
+        $this->assertTrue($request->getIsSecureConnection());
+
+        // Trust HTTPS header over the X_FORWARDED header
+        $_SERVER['HTTP_X_FORWARDED_PROTO'] = 'http';
+        $this->assertTrue($request->getIsSecureConnection());
+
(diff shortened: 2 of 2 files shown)


framework/web/Request.php
```diff
@@ -874,7 +874,7 @@ public function getQueryString()
     public function getIsSecureConnection()
     {
         return isset($_SERVER['HTTPS']) && (strcasecmp($_SERVER['HTTPS'], 'on') === 0 || $_SERVER['HTTPS'] == 1)
-            || isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && strcasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https') === 0;
+            || isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && strncasecmp($_SERVER['HTTP_X_FORWARDED_PROTO'], 'https', 5) === 0;
```
bpicolo: this is magical, happy to s/5/strlen('https') if preferred

samdark (approved):

SilverFire (approved):

framework/web/Request.php
samdark: It looks OK to me as is.

dynasource (changes requested):
This PR intoduces new unit tests which are not covering all the state possibilities (like for ```_SERVER['HTTPS'], 'on'```). These should be added or at least be commented as 'todo' for future expansion.

Also, it is unclear why strcasecmp is only changed for the second and not for the first use of strcasecmp.

bpicolo:
@dynasource did you read the Issue? The first strcasecmp is irrelevant to the issue at hand.

As for testing, I don't agree that exhaustive testing is or should be a requirement. Especially as test coverage prior is actual 0.

dynasource:
@bpicolo, I am watching your PR from a future code review / read perspective. When someone sees these similar functions a false assumption can be made that a synchronization is forgotten. It would be helpful to understand this from the code with proper naming.

Similar applies for your tests. When somebody is looking at the tests in the future, a false assumption is made about the tests being complete. In this case, it isn't. Also here it would be helpful to easily whats covered and what not. 

We dont want us to burn time in the future to 'interpretate' while this can be prevented.

bpicolo:
I'll change it up, but imo the bar for making OSS contributions convenient here is not being applied at the right height

bpicolo:
At least partly because this was 3 weeks ago and is now much less on mind 😶

bpicolo:
@dynasource  Does that suit you better? I suspect a comment there is more appropriate than revising the former to strncasecmp, and would prefer to avoid temporary variables for naming's sake (yeah, unfortunate that PHP internals naming is poor in general ☹️  )

dynasource:
@bpicolo, thanks for putting effort in the extra work. It will suit if the comment is 1 line and explationary for the use of strncasecmp (optionally by a reference to the issue topic)

bpicolo:
@dynasource  Already put a 4-line comment in to be extra explicit. Seem good?

dynasource:
if you think you cannot shrink this comment any further without affecting interpretation, it is OK by me.

bpicolo:
Wasn't sure if the former was a question or a statement. there ya go

dynasource (approved):

dynasource:
your persistence is appreciated

bpicolo:
<3 Thanks for the CR. I understand where you're coming from, you know how it is coming back to popups from old contexts at 7AM though, hah.

samdark (changes requested):
Reviewed again. It's not safe to rely on `HTTP_` headers if you're not behind a trusted proxy. I think we should introduce a list of trusted IPs and consider `HTTP_` headers only if requested one is one of these.

samdark:
See https://github.com/yiisoft/yii2/issues/10854

bpicolo:
@samdark seems resolveable by LB configuration no? 

IP whitelist doesn't work well in the world of auto scaling cloud groups.

bpicolo:
True that it's not necessarily a good default though. Not sure.

samdark:
@bpicolo yes, it's resolvable by proper configuration when you have a load balancer in place but isn't a safe default in case you don't have load balancer at all i.e. more than half of all cases.

samdark:
Thanks for contributing. The issue will be solved but not with this pull request. Sorry about that.

PR #13461, opened 2017-01-28
Use low-level query in UniqueValidator

### LINKED ISSUE #13453: UniqueValidator now 'Select only primary keys'. Other attributes have null values when BaseActiverecord::afterFind is called.
Set a 'unique' validation rule for an attribute in a model.
Add some code that requires attributes to the afterFind() method:
```php
public function afterFind() {
    $this->emailUsername = $this->extractEmailUsername($this->email);
}
//...
public function extractEmailUsername($email)
{
    $username = substr($email, 0, strpos($email, '@'));
    if (false !== $username) {
        return $username;
    } else {
        throw new InvalidParamException("Failed to extract username from email '$email'.");
    }
 }
```
Run validation **with an existing value** for the expected-unique attribute.

A validation failure with an error message.

### What do you get instead?
An InvalidParamException is thrown when I validate the 'email' attribute with an existing value.

### Additional info
Changes were made in this [commit](https://github.com/yiisoft/yii2/commit/729ddc5b76e8159d051e9c5783f5534eeb212efb)

 Yii version      | 2.0.11

Thanks

developeruz:
Fixes #13453, relates to #13098

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | Fixes #13453, relates to #13098

Use low-level query in UniqueValidator to avoid triggering `afterFind` and other events.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -114,6 +114,7 @@ Yii Framework 2 Change Log
 - Enh #11464: Populate foreign key names from schema (joaoppereira)
 - Bug #13401: Fixed lack of escaping of request dump at exception screens (samdark)
 - Enh #13417: Allow customizing `yii\data\ActiveDataProvider` in `yii\rest\IndexAction` (leandrogehlen)
+- Enh #13453: Select only primary key when counting records in UniqueValidator (developeruz)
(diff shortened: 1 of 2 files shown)


samdark:
Thank you!

lynicidn:
https://github.com/yiisoft/yii2/commit/bede9feba5b39169381d7af3c48ae89753fbe9e5

Issue #11230, opened 2016-03-30
\yii\rbac\DbManager getRolesByUser() ignores defaultRoles


vercotux:
It would make sense for [getRolesByUser()](https://github.com/yiisoft/yii2/blob/049ea37f778fae576d10835c48b476364a457436/framework/rbac/DbManager.php#L454) to also return the roles specified inside `defaultRoles` parameter, because logically every user should have those roles.

samdark:
Agree.

Oxyaction:
I can see the following problem implementing this feature:

Usually defaultRoles are implicitly given to the all users but, in fact common practice is that such a roles are used with some rules and that rules may require some context params that are passed as 3rd param `Rule::execute()`. So simple doing something like in:

`PhpManager::getRolesByUser($userId)`
```php
$deafultRoles = array_filter($this->defaultRoles, function($role) {
  return $this->checkAccess($userId, $role);
})
```

will make no sense due to lack of `$params` that should be passed to `checkAccess`.
any ideas @samdark ?

samdark:
Nope. I don't have any ideas about getting it right in all contexts in this case...

PR #13462, opened 2017-01-28
Include defaultRoles in getRolesByUser() results

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #11230


framework/rbac/BaseManager.php
```diff
@@ -190,6 +190,18 @@ public function getRoles()
     }
 
     /**
+     * Returns defaultRoles as array of Role objects
```
samdark: Need `@since 2.0.12`.

samdark: Missing `@return`

samdark (changes requested):

framework/rbac/BaseManager.php
developeruz: Added

samdark (approved):

SilverFire (dismissed):

framework/rbac/BaseManager.php
```diff
@@ -190,6 +190,20 @@ public function getRoles()
...
+     * @since 2.0.12
+     * @return Role[] default roles. The array is indexed by the role names
+     */
+    public function getDefaultRoles()
+    {
+        $result = [];
+        foreach ($this->defaultRoles as $roleName) {
+            $result[$roleName] = new Role(['name' => $roleName]);
```
SilverFire: We already have method `createRole()` for this purpose.
I think it is better to call it instead

```php
$result[$roleName] = $this->createRole($roleName);
```

SilverFire (changes requested):

framework/rbac/BaseManager.php
developeruz: I changed this line. Thanks.

SilverFire (approved):

samdark:
Merged. Thanks!

hirenbhut93:
@samdark, @developeruz  It's work when define rule with default role? Like http://www.yiiframework.com/doc-2.0/guide-security-authorization.html#using-default-roles

samdark:
If rule is executed before the call — yes.

hirenbhut93:
@samdark What?

samdark:
In that case it's not available at the call time so I think it won't be returned.

framework/rbac/BaseManager.php
```diff
@@ -190,6 +190,20 @@ public function getRoles()
     }
 
     /**
+     * Returns defaultRoles as array of Role objects
+     * @since 2.0.12
+     * @return Role[] default roles. The array is indexed by the role names
+     */
+    public function getDefaultRoles()
```
cebe: method name conflicts with the property. This should be changed.

developeruz: Should I create a new PR because this one is already merged?
Can you, please, suggest a name for this method?

arogachev: @developeruz Yes, you should create new PR. Suggest your variation of naming and we can discuss from there.

PR #13465, opened 2017-01-29
add options to include directories to findFiles in BaseFileHelper

ArsSirek:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  |

Hi all, firstly, thanks for the great work.
 Desired functionality - to get a list of folders, and be able to easily filter it by names. (like in findFiles with filter option for files)
  If I get it right, there is no function in yii2 to help to get it done, you can create, delete but not search directories with BaseFileHelper? I feel like FileHelper is a proper place to have that functionality, as I read it as "file system helper". Could you please give me feedback on the idea to add it either as options in findFiles or as a separate function?

Maybe it is barely needed to anyone, i am not sure, Is it design decision? (at least i am not alone http://stackoverflow.com/questions/39190811/how-to-list-all-directory-in-yii2 ) ) 
 I am sure my code isn't suit to be merged yet, but I didn't find batter way to show the idea and understand should I work further on that direction or not.


dynasource:
Its useful to have a method for recursively searching for folders. However, this should obviously not be implemented in `findFiles` but in its own method.

samdark:
Another way could be to introduce generic "find" method.

ArsSirek:
Hi again, thanks for the attention and advises.
Sorry for this late response, I just wasn't able to come up with anything more worthy  to offer here.

@dynasource, thanks. It seems to me, that separate function will be very similar to the current  findFiles but with opposite file type checks. In my small case I actually needed both files and directories equally, so having option to get them at one request looks more reasonable to me.

I won't be able to work on this feature in near time, should I close it? or leave it be for now?

Thanks again!

samdark:
Leave it.

samdark:
Merged. Thank you!

Issue #12752, opened 2016-10-16
ActiveDataProvider should avoid query data when data count is zero


lichunqiang:
### What steps will reproduce the problem?

> This is not a bug, just a suggestion.

[ActiveDataProvider::prepareModels](https://github.com/yiisoft/yii2/blob/master/framework/data/ActiveDataProvider.php#L98)  because there had queried total count, so when query count is 0, there is not need to query list.
### What is the expected result?

When query is none result, there is should be one query which is query the total of result, no need for query the list.
### What do you get instead?

Two query performed.
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.9 |
| PHP version | 7.0.11 |
| Operating system | Cenots. |

samdark:
I agree that it would be good if it can be done w/o much hacking...

Alex-Code:
Would this work?

```php
if (($pagination = $this->getPagination()) !== false) {
    if (0 === ($pagination->totalCount = $this->getTotalCount())) {
        return [];
    }
    $query->limit($pagination->getLimit())->offset($pagination->getOffset());
}
```

samdark:
Probably. Is it covered by unit tests?

TerraSkye:
i would swap 
`if (0 === ($pagination->totalCount = $this->getTotalCount())) {`

with 
`if (($pagination->totalCount = $this->getTotalCount()) === 0) {`

as we always compare  against something ( this case its 0 ).
or is this just my personal prefference.

cebe:
related to #6373

PR #13467, opened 2017-01-29
ActiveDataProvider should avoid query data when data count is zero

kLkA:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12752

### DIFF
--- a/framework/data/ActiveDataProvider.php
+++ b/framework/data/ActiveDataProvider.php
@@ -103,7 +103,8 @@ class ActiveDataProvider extends BaseDataProvider
         }
         $query = clone $this->query;
         if (($pagination = $this->getPagination()) !== false) {
-            if (($pagination->totalCount = $this->getTotalCount()) === 0) {
+            $pagination->totalCount = $this->getTotalCount();
+            if ($pagination->totalCount === 0) {
                 return [];
             }
             $query->limit($pagination->getLimit())->offset($pagination->getOffset());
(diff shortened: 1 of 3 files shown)


framework/data/ActiveDataProvider.php
```diff
@@ -103,7 +103,9 @@ protected function prepareModels()
         }
         $query = clone $this->query;
         if (($pagination = $this->getPagination()) !== false) {
-            $pagination->totalCount = $this->getTotalCount();
+            if (($pagination->totalCount = $this->getTotalCount()) === 0) {
```
Renkas: why put all the logic in one line? Good practice is that every action goes to a separate line. Here you are doing assignment and then comparison on one ...

I can't see any performance optimization neither on this syntax.

samdark (approved):

framework/data/ActiveDataProvider.php
kLkA: Hi @Renkas , I agree with you. Thank you

SilverFire (changes requested):
The test does not check the query skipping fact. It just proves, that you will get 0 modes when count query returns 0 rows.

SilverFire:
My suggestions is to create a `ZeroCountQueryStub` class, with `count()` method that always returns 0

dynasource:
> The test does not check the query skipping fact. It just proves, that you will get 0 modes when count query returns 0 rows.

agree

PR #13470, opened 2017-01-30
docs/guide fix [ci skip]

softark:
docs/guide typo fix

"true", "false" and "nul"` in code blocks should not be quoted with backquotes.


samdark:
Thanks.

PR #13488, opened 2017-02-01
Improved description of format of $controllers

niekoost:
Improved explanation of the format of the $controllers property. This explains what is meant by "prefixed with the module ID"

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/framework/filters/AccessRule.php
+++ b/framework/filters/AccessRule.php
@@ -32,6 +32,7 @@ class AccessRule extends Component
     public $actions;
     /**
      * @var array list of the controller IDs that this rule applies to. Each controller ID is prefixed with the module ID (if any).
+     * The comparison uses [[Controller::uniqueId]], so the items in $controller look like ['shop/product'].
      * The comparison is case-sensitive. If not set or empty, it means this rule applies to all controllers.
      */
     public $controllers;


framework/filters/AccessRule.php
```diff
@@ -32,6 +32,7 @@ class AccessRule extends Component
     public $actions;
     /**
      * @var array list of the controller IDs that this rule applies to. Each controller ID is prefixed with the module ID (if any).
+     * The comparison uses [[Controller::uniqueId]], so the items in $controller look like ['shop/product'].
```
cebe: This is only the case if the controller is located inside of a module, that should be more clear. How about the following?

> @var array list of the controller IDs that this rule applies to.
> The comparison uses [[\yii\base\Controller::uniqueId]], so each controller ID is prefixed with the module ID (if any). So for a `product` controllers in the application, you would specify this like `['product']` and if that controller is located in a `shop` module, this would be `['shop/product']`.

samdark: Perfect. Let's use it.

cebe:
Merged with adjustments, thanks for the proposal!

PR #13489, opened 2017-02-01
Made ActionColumn Yii::t() tags static again

cebe:
this was changed in 2.0.11 and cause the messages translation command to not
recognize them anymore.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | n/a

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -5,6 +5,7 @@ Yii Framework 2 Change Log
 ------------------------
 
 - Bug #11502: Fixed `yii\console\controllers\MessageController` to properly populate missing languages in case of extraction with "db" format (bizley)
+- Bug #13489: Fixed button names in ActionColumn to contain proper `Yii::t()` tags and restored missing translations for `ja` and `ru` (cebe, softark)
 
 
 2.0.11 February 01, 2017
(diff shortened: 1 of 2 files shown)


rob006:
Can't we just use:

```php
// Yii::t('yii', 'View');
// Yii::t('yii', 'Update');
// Yii::t('yii', 'Delete');
```
for messages extraction?

samdark:
That's way too hacky.

samdark (approved):

rob006:
Why hacky? In many situation you may need only mark some text as translatable without real translating (other frameworks/tools already have methods for that, like `gettext_noop()`), using comment is the most clean and performance-wise solution for me.

samdark:
I don't think `switch` is worse than `ucfirst`.

samdark (changes requested):
Need changelog

cebe:
@samdark done.
@rob006 for the fix I'd merge it like this now but if you think this feature is needed you can open an issue and we can discuss how to implement it. It may be useful but I am not sure it is a good idea to do it with comments.

samdark (approved):

cebe:
There is now a hotfix release 2.0.11.1, which includes this fix.

PR #13497, opened 2017-02-02
messages/ja/yii.php reverted [ci skip]

softark:
The entries of "Delete", "Update" and "View" had been deleted in https://github.com/yiisoft/yii2/commit/e05a883fdfc77fdb7c1491049d78966ac91a46ac
But it was a mistake. They should not be deleted.

https://github.com/yiisoft/yii2/blob/master/framework/grid/ActionColumn.php#L164

The above change is the cause of the accident. It hides "Delete", "Update" and "View" from the message extractor tool.

The message file for Russian language is also damaged in the same way.


cebe:
Thank you!

Issue #11719, opened 2016-06-07
Enabling queryCache on a database that is also the storage for Cache causes a loop


michaelarnauts:
I notice a loop when enabling `queryCache` on a database that is also the storage for the `cache` component. This might be related to #5106.

It seems that the `DbCache` his `setValue` and `addValue` function just write the info to the default `db` component,  that in the end causes a loop since the query to save the cache value will also trigger a new insert for the `queryCache`.

I've took a quick look in `DbCache` but there seems no way to block a new insertion in the DbCache from generating a new cache entry.

This patch encloses those updates in a `noCache()` call, and if I'm right, that should solve this. Am I correct? Did I miss anything?

``` patch
diff --git a/framework/caching/DbCache.php b/framework/caching/DbCache.php
index 8c1fcbb..55819e1 100644
--- a/framework/caching/DbCache.php
+++ b/framework/caching/DbCache.php
@@ -185,13 +185,16 @@ class DbCache extends Cache
      */
     protected function setValue($key, $value, $duration)
     {
-        $command = $this->db->createCommand()
-            ->update($this->cacheTable, [
-                'expire' => $duration > 0 ? $duration + time() : 0,
-                'data' => [$value, \PDO::PARAM_LOB],
-            ], ['id' => $key]);
+        $result = $this->db->noCache(function (Connection $db) {
+            $command = $db->createCommand()
+                ->update($this->cacheTable, [
+                    'expire' => $duration > 0 ? $duration + time() : 0,
+                    'data' => [$value, \PDO::PARAM_LOB],
+                ], ['id' => $key]);
+            return $command->execute();
+        });

-        if ($command->execute()) {
+        if ($result) {
             $this->gc();

             return true;
@@ -214,12 +217,14 @@ class DbCache extends Cache
         $this->gc();

         try {
-            $this->db->createCommand()
-                ->insert($this->cacheTable, [
-                    'id' => $key,
-                    'expire' => $duration > 0 ? $duration + time() : 0,
-                    'data' => [$value, \PDO::PARAM_LOB],
-                ])->execute();
+            $this->db->noCache(function (Connection $db) {
+                $db->createCommand()
+                    ->insert($this->cacheTable, [
+                        'id' => $key,
+                        'expire' => $duration > 0 ? $duration + time() : 0,
+                        'data' => [$value, \PDO::PARAM_LOB],
+                    ])->execute();
+            });

             return true;
         } catch (\Exception $e) {
[...]

PR #13506, opened 2017-02-03
Enabling queryCache on a database that is also the storage for Cache caused a loop.

michaelarnauts:
Enabling queryCache on a database that is also the storage for Cache caused a loop.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #11719

### DIFF
--- a/framework/caching/DbCache.php
+++ b/framework/caching/DbCache.php
@@ -187,13 +187,16 @@ class DbCache extends Cache
      */
     protected function setValue($key, $value, $duration)
     {
-        $command = $this->db->createCommand()
-            ->update($this->cacheTable, [
-                'expire' => $duration > 0 ? $duration + time() : 0,
-                'data' => [$value, \PDO::PARAM_LOB],
-            ], ['id' => $key]);
+        $result = $this->db->noCache(function (Connection $db) use ($key, $value, $duration) {
+            $command = $db->createCommand()
+                ->update($this->cacheTable, [
+                    'expire' => $duration > 0 ? $duration + time() : 0,
+                    'data' => [$value, \PDO::PARAM_LOB],
+                ], ['id' => $key]);
+            return $command->execute();
+        });
 
-        if ($command->execute()) {
+        if ($result) {
             $this->gc();
 
             return true;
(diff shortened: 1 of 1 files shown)


framework/caching/DbCache.php
```diff
@@ -237,9 +242,11 @@ protected function addValue($key, $value, $duration)
      */
     protected function deleteValue($key)
     {
-        $this->db->createCommand()
-            ->delete($this->cacheTable, ['id' => $key])
-            ->execute();
+        $this->db->noCache(function (Connection $db) {
+            $this->db->createCommand()
```
samdark: Should `$db` be used here instead of `$this->db`?

samdark (changes requested):

framework/caching/DbCache.php
michaelarnauts: Hmm, I guess that makes sence!

framework/caching/DbCache.php
```diff
@@ -216,12 +219,14 @@ protected function addValue($key, $value, $duration)
...
-                    'id' => $key,
-                    'expire' => $duration > 0 ? $duration + time() : 0,
-                    'data' => [$value, \PDO::PARAM_LOB],
-                ])->execute();
+            $this->db->noCache(function (Connection $db) {
+                $db->createCommand()
+                    ->insert($this->cacheTable, [
+                        'id' => $key,
+                        'expire' => $duration > 0 ? $duration + time() : 0,
+                        'data' => [$value, \PDO::PARAM_LOB],
```
michaelarnauts: I'm wondering now (it has been a while since I wrote this code), is `$this` or `$value` accessible from inside this callback?

samdark: `$this` is accessible. Other variables aren't. Should be:

```php
function (Connection $db) use ($value) {
```

framework/caching/DbCache.php
```diff
@@ -237,9 +242,11 @@ protected function addValue($key, $value, $duration)
      */
     protected function deleteValue($key)
     {
-        $this->db->createCommand()
-            ->delete($this->cacheTable, ['id' => $key])
-            ->execute();
+        $this->db->noCache(function (Connection $db) {
```
samdark:
```php
function (Connection $db) use ($key) {
```

framework/caching/DbCache.php
```diff
@@ -187,13 +187,16 @@ protected function getValues($keys)
      */
     protected function setValue($key, $value, $duration)
     {
-        $command = $this->db->createCommand()
-            ->update($this->cacheTable, [
-                'expire' => $duration > 0 ? $duration + time() : 0,
-                'data' => [$value, \PDO::PARAM_LOB],
-            ], ['id' => $key]);
+        $result = $this->db->noCache(function (Connection $db) {
```
samdark:
```php
function (Connection $db) use ($key, $value, $duration) {
```

framework/caching/DbCache.php
samdark: I wonder how did that actually worked for you...

michaelarnauts: I wonder that too, I guess it didn't but it also didn't hang my app anymore (see #11719).

samdark: Could you test with proper code?

SilverFire (changes requested):
Looks good in general. Could you add a changelog line and create test?

samdark:
Verified and merged. Thank you.

Issue #13501, opened 2017-02-02
Incorrect behaviour in DbManager


StalkAlex:
### What steps will reproduce the problem?
Trying to create permission
```
$permission = $auth->createPermission('test');
$permission->ruleName = 'TestRule';
$auth->add($permission)
```
data of rule is ```O:24:"app\rules\TestRule":3:{s:4:"name";s:14:"TestOnly";s:9:"createdAt";i:1485582849;s:9:"updatedAt";i:1485582849;}```

### What do you get instead?
Error
```bash
PHP Warning 'yii\base\ErrorException' with message 'unserialize() expects parameter 1 to be string, resource given'

in /vagrant/vendor/yiisoft/yii2/rbac/DbManager.php:634
```
Here's screen. As it seen data variable isn't sting anymore but resource stream.
![test](https://cloud.githubusercontent.com/assets/2943154/22559100/e7085e4e-e991-11e6-9176-02b529ff6636.png)

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11
| PHP version      | 7.1.1
| Operating system | CentOS 7
| Database | PostgreSQL 9.6

StalkAlex:
There is no such bug in 2.0.10, so it's not db related problem.

samdark:
Probably. Can you find out which commit breaks it using `git bisect`?

StalkAlex:
Yep. I check.

StalkAlex:
@samdark  found commit that causes this bug https://github.com/yiisoft/yii2/commit/23790272dc561aacf2070cbbda396f49e44cbb7d

cebe:
To fix it a check should be added that will read all data from the stream. possibly via http://php.net/manual/en/function.stream-get-contents.php

StalkAlex:
Like this? https://github.com/StalkAlex/yii2/commit/5a76ef110e99d377b1f79a2b814656da45a7b610 Checked in own project, it worked.

samdark:
Looks OK.

cebe:
There is now a 2.0.11.2 release including this fix. https://github.com/yiisoft/yii2/blob/2.0.11.2/framework/CHANGELOG.md#20112-february-08-2017

StalkAlex:
Good news, thanks!

simonevallana:
This problem remain also in the 2.0.11.2 version in the method loadFromCache() in DbManager class (line 990). Could you fix it? Thank you

[...]

PR #13507, opened 2017-02-03
fix reading stream came from db

StalkAlex:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13501

### DIFF
--- a/framework/rbac/DbManager.php
+++ b/framework/rbac/DbManager.php
@@ -631,11 +631,14 @@ class DbManager extends BaseManager
             ->from($this->ruleTable)
             ->where(['name' => $name])
             ->one($this->db);
-        $result = $row['data'];
-        if (is_resource($result)) {
-            $result = stream_get_contents($result);
+        if ($row === false) {
+            return null;
         }
-        return $row === false ? null : unserialize($result);
+        $data = $row['data'];
+        if (is_resource($data)) {
+            $data = stream_get_contents($data);
+        }
+        return unserialize($data);
 
     }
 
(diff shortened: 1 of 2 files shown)


samdark (approved):
Looks good. Need a line for changelog.

framework/rbac/DbManager.php
```diff
@@ -631,7 +631,12 @@ public function getRule($name)
             ->from($this->ruleTable)
             ->where(['name' => $name])
             ->one($this->db);
-        return $row === false ? null : unserialize($row['data']);
+        $result = $row['data'];
```
cebe: `$row` can be false, so that check must be done here.

cebe (changes requested):
Also a testcase would be great to cover this.

framework/rbac/DbManager.php
StalkAlex: ok, I'll change it and make test.

StalkAlex:
I fixed both getRule and getRules method. Some tests were failed due to this bug, now there are all ok. Not sure what else testcase could be here, because current tests have covered this already from my point of view.

samdark:
2.0.11 was not failing: https://travis-ci.org/yiisoft/yii2/builds/197358860

StalkAlex:
I ran locally only rbac group for 2.0.11.1. Here's result:
default
I suppose it's correct behaviour because of bug I reported.
Here's run for fixed version:
default

samdark:
Merged. Thanks!

PR #13516, opened 2017-02-04
Update git-workflow.md

derekisbusy:
vendor/bin/phpunit is bash script not php

| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

### DIFF
--- a/docs/internals/git-workflow.md
+++ b/docs/internals/git-workflow.md
@@ -66,7 +66,7 @@ The following steps are optional.
 ### Unit tests
 
 You can execute unit tests by running `phpunit` in the repo root directory. If you do not have phpunit installed globally
-you can run `vendor/bin/phpunit` instead.
+you can run `php vendor/bin/phpunit` or run `vendor/bin/phpunit.bat` (Windows) instead.
 
 Some tests require additional databases to be set up and configured. You can create `tests/data/config.local.php` to override
 settings that are configured in `tests/data/config.php`.
--- a/docs/internals/git-workflow.md
+++ b/docs/internals/git-workflow.md
@@ -66,7 +66,7 @@ The following steps are optional.
 ### Unit tests
 
 You can execute unit tests by running `phpunit` in the repo root directory. If you do not have phpunit installed globally
-you can run `php vendor/bin/phpunit` instead.
+you can run `vendor/bin/phpunit` instead.
 
 Some tests require additional databases to be set up and configured. You can create `tests/data/config.local.php` to override
 settings that are configured in `tests/data/config.php`.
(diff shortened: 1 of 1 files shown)


rob006:
It **is** PHP script, but with shebang.

derekisbusy:
Okay true but if you try to run `php vendor/bin/phpunit` it outputs the file contents.

rob006:
> Okay true but if you try to run php vendor/bin/phpunit it outputs the file contents.

No, it should execute script.

derekisbusy:
Actually no it is a shell script...

```sh
#!/usr/bin/env sh

dir=$(d=${0%[/\\]*}; cd "$d"; cd "../phpunit/phpunit" && pwd)

# See if we are running in Cygwin by checking for cygpath program
if command -v 'cygpath' >/dev/null 2>&1; then
	# Cygwin paths start with /cygdrive/ which will break windows PHP,
	# so we need to translate the dir path to windows format. However
	# we could be using cygwin PHP which does not require this, so we
	# test if the path to PHP starts with /cygdrive/ rather than /usr/bin
	if [[ $(which php) == /cygdrive/* ]]; then
		dir=$(cygpath -m "$dir");
	fi
fi

dir=$(echo $dir | sed 's/ /\ /g')
"${dir}/phpunit" "$@"
```

cebe:
not sure where you go that shell script from but for me it installs a php script. Also the file in the phpunit repo is a php script: https://github.com/sebastianbergmann/phpunit/blob/fc2561af7296fd54ca3d6ca086e0ddedc01425d9/phpunit

rob006:
@cebe It looks like this is some proxy script generated by composer: https://github.com/composer/composer/blob/14c77130e8f717ceb154720e1abefafe99124cb9/src/Composer/Installer/BinaryInstaller.php#L187

derekisbusy:
I guess you must be using Linux? Might want to add note that it's different on Windows.

cebe:
does running it without `php` work on windows?`Then we could accept this PR as `vendor/bin/phpunit` works on linux too.

derekisbusy:
No, in Windows it outputs the contents of the shell script.

samdark:
It works for me under regular Windows 10 CMD. Running `phpunit` from `vendor/bin` executes phpunit as expected.

samdark:
One thing that it doesn't really work well with relative paths so first `cd` to `vendor/bin` then execute `phpunit`.

dynasource (approved):

SilverFire (approved):

samdark (approved):

samdark:
Merged. Thanks!

Issue #13473, opened 2017-01-30
Yii2 does not call actions in controller's behavior


dslpp056193:
I think it will be very good, if behavior's actions in controller were calling like a controller's actions.

derekisbusy:
It might also be worth adding similar functionality to modules the applications as well. One advantage of using this with application behaviors is that it would allow extensions to integrate with other more cleanly. For example an extension could attach a single behavior to the application that can then add modules, controllers, actions and events as needed.

samdark:
What's your use case for it?

samdark:
See discussion in a pull request.

PR #13517, opened 2017-02-05
Controller behavior interface

derekisbusy:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13473

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -6,7 +6,7 @@ Yii Framework 2 Change Log
 
 - Bug #13501: Fixed `yii\rbac\DbManager::getRule()` and `yii\rbac\DbManager::getRules()` to properly handle resource data came from Rule table when using PostgreSQL (StalkAlex)
 - Bug #13508: Fixed duplicate attachment of behavior BC break (cebe)
-- Enh #13473: Add ControllerBehaviorInterface that allows behaviors who implement the interface to add actions to their owner. (derekisbusy)
+- Enh #13473: Add ControllerBehaviorInterface that allows behaviors which implement the interface to add actions to their owner. (derekisbusy)
 
 2.0.11.1 February 02, 2017
 ------------------------
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -6,6 +6,7 @@ Yii Framework 2 Change Log
 
 - Bug #13501: Fixed `yii\rbac\DbManager::getRule()` and `yii\rbac\DbManager::getRules()` to properly handle resource data came from Rule table when using PostgreSQL (StalkAlex)
 - Bug #13508: Fixed duplicate attachment of behavior BC break (cebe)
+- Enh #13473: Add ControllerBehaviorInterface that allows behaviors who implement the interface to add actions to their owner. (derekisbusy)
 
 2.0.11.1 February 02, 2017
 ------------------------
--- a/framework/base/Controller.php
+++ b/framework/base/Controller.php
@@ -221,6 +221,15 @@ class Controller extends Component implements ViewContextInterface
         }
 
         $actionMap = $this->actions();
+        foreach ($this->getBehaviors() as $behavior) {
+            if ($behavior instanceof ControllerBehaviorInterface) {
+                $actions = $behavior->actions();
+                if (!empty($actions)) {
+                    $actionMap = array_merge($actionMap, $actions);
+                }
+            }
+        }
+
         if (isset($actionMap[$id])) {
             return Yii::createObject($actionMap[$id], [$id, $this]);
         } elseif (preg_match('/^[a-z0-9\\-_]+$/', $id) && strpos($id, '--') === false && trim($id, '-') === $id) {
--- a/framework/base/Controller.php
+++ b/framework/base/Controller.php
@@ -221,6 +221,7 @@ class Controller extends Component implements ViewContextInterface
         }
 
         $actionMap = $this->actions();
+        $this->ensureBehaviors();
         foreach ($this->getBehaviors() as $behavior) {
             if ($behavior instanceof ControllerBehaviorInterface) {
                 $actionMap = array_replace_recursive($actionMap, $behavior->actions());
--- a/framework/base/Controller.php
+++ b/framework/base/Controller.php
@@ -224,7 +224,7 @@ class Controller extends Component implements ViewContextInterface
         foreach ($this->getBehaviors() as $behavior) {
             if ($behavior instanceof ControllerBehaviorInterface) {
                 $actions = $behavior->actions();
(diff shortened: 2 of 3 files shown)


framework/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Yii Framework 2 Change Log
 
 - Bug #13501: Fixed `yii\rbac\DbManager::getRule()` and `yii\rbac\DbManager::getRules()` to properly handle resource data came from Rule table when using PostgreSQL (StalkAlex)
 - Bug #13508: Fixed duplicate attachment of behavior BC break (cebe)
+- Enh #13473: Add ControllerBehaviorInterface that allows behaviors who implement the interface to add actions to their owner. (derekisbusy)
```
arogachev: "Who" is never used to refer to things. See for example [this article](http://blog.oxforddictionaries.com/2012/09/that-who-which/) for more details.

framework/base/ControllerBehaviorInterface.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+namespace yii\base;
+/**
+ * ControllerBehaviorInterface is the class that should be implemented by behaviors who want to add actions to the controller.
+ *
+ * @author Qiang Xue <qiang.xue@gmail.com>
```
arogachev: Write your name and email instead or just omit that.

framework/base/ControllerBehaviorInterface.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+namespace yii\base;
+/**
+ * ControllerBehaviorInterface is the class that should be implemented by behaviors who want to add actions to the controller.
```
arogachev: Again, wrong usage of "who".

framework/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Yii Framework 2 Change Log
 
 - Bug #13501: Fixed `yii\rbac\DbManager::getRule()` and `yii\rbac\DbManager::getRules()` to properly handle resource data came from Rule table when using PostgreSQL (StalkAlex)
 - Bug #13508: Fixed duplicate attachment of behavior BC break (cebe)
+- Enh #13473: Add ControllerBehaviorInterface that allows behaviors which implement the interface to add actions to their owner. (derekisbusy)
```
samdark: Should be moved to 2.0.12...

samdark: Last `.` before nickname should be removed.

framework/base/Controller.php
```diff
@@ -221,6 +221,13 @@ public function createAction($id)
         }
 
         $actionMap = $this->actions();
+        $this->ensureBehaviors();
+        foreach ($this->getBehaviors() as $behavior) {
+            if ($behavior instanceof ControllerBehaviorInterface) {
+                $actionMap = array_replace_recursive($actionMap, $behavior->actions());
```
samdark: Not sure if behavior should replace already defined actions.

framework/base/ControllerBehaviorInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+namespace yii\base;
+/**
+ * ControllerBehaviorInterface is the class that should be implemented by behaviors that want to add actions to the controller.
+ *
+ * @author Derek Gifford <derekisbusy@gmail.com>
+ * @since 2.0.11.2
```
samdark: 2.0.12

samdark (changes requested):

framework/base/Controller.php
derekisbusy: Could change it to use `ArrayHelper::merge()` and add a new helper `yii\helpers\ArrayValueIfKeyNotExists` that would only add the key/value pair if the key doesn't exist in the other array.

```php
$array1 = [
        'error' => 'yii\web\ErrorAction',
        'view' =>  'yii\web\ViewAction',
 ];
$array2 = [
        'error' => 'yii\web\ErrorAction',
        'view' => new \yii\helpers\ArrayValueIfKeyNotExists(  'my\custom\ViewAction'),
        'custom' => new \yii\helpers\ArrayValueIfKeyNotExists(  'my\custom\CustomAction'),
 ];

$result = ArrayHelper::merge($array1, $array2);

// result will be...
[
        'error' => 'yii\web\ErrorAction',
        'view' =>  'yii\web\ViewAction',
        'custom' => 'my\custom\CustomAction',
 ];
```

This would provide the most flexibility as it would allow the developer to decide whether or not the values should be replaced.

derekisbusy: Could also add a second parameter to control what happens if the key already exists to make debugging less ambiguous:

```php
$array1 = [
        'error' => 'yii\web\ErrorAction',
        'view' =>  'yii\web\ViewAction',
        'custom' => 'my\custom\CustomAction',
 ];
$array2 = [
        'error' => 'yii\web\ErrorAction',
        'view' => new \yii\helpers\ArrayValueIfKeyNotExists(  'my\custom\ViewAction' , function() { 
                               throw new Exception('View action already defined.'); 
        }),
        'custom' => new \yii\helpers\ArrayValueIfKeyNotExists(  'my\custom\CustomAction'),
 ];

// would throw an exception
$result = ArrayHelper::merge($array1, $array2);

```

rob006:
This makes actions really magically and hard to follow. Right now if you want check what actions have specified controller, you search methods prefixed by `action` and `action()` method. After this PR you need check additionally each behavior. And if you use vendor behavior, each update could implicitly add new actions to controller.

samdark:
@rob006 yes, it becomes as implicit as behaviors are.

rob006:
The same could be achieved with

```php
    public function actions()
    {
        return array_merge(
            $this->getBehavior('myBehavior1')->actions(),
            $this->getBehavior('myBehavior2')->actions(),
            $this->getBehavior('myBehavior3')->actions(),
            [...],
        );
    }
```

in controller class, but with more control over actions and less magic.

derekisbusy:
@rob006 The only advantage would be that you can add actions to a controller without updating the controller, for example controllers in vendor extensions. But this can also be achieved by extending the controller using the controller map. So I guess the proper way to do this would be to extend the controller and add the behavior and actions.

Issue #13522, opened 2017-02-06
Issue with URL::to when the URL only contains a default element


MattDalm:
### What steps will reproduce the problem?

in the config/main.php add the following rule

`                [
                    'pattern' => '<language>',
                    'route' => 'site/index',
                    'defaults' => ['language' => 'en'],
                ], 
`

Then create a URL like that

`Url::to(['site/index', 'language' =>'en'])`

### What is the expected result?
The generated URL should be 

`<a href="/">Home</a>`

### What do you get instead?

`<a href="//">Home</a>`

The slash is duplicated

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.1
| PHP version      | 7.1.0
| Operating system | Windows 10 and Ubuntu 16.04

Issue is not reproducible on Yii 2.0.10

MattDalm:
Hi 

I just tried the fix you committed and it works fine now. 

Thanks.

--Matt

cebe:
There is now a 2.0.11.2 release including this fix. https://github.com/yiisoft/yii2/blob/2.0.11.2/framework/CHANGELOG.md#20112-february-08-2017

Issue #12691, opened 2016-10-06
Host-based routing for http and https requests


erickskrauch:
### What steps will reproduce the problem?

Проблема строится на том, что текущая реализация роутера не предусматривает вариацию, когда запрос к домену идёт через оба протокола. Например, имеется такой конфиг:

``` php
'urlManager' => [
    'enablePrettyUrl' => true,
    'showScriptName' => false,
    'rules' => [
        "http://you-domain.net/some/path" => 'controller/action',
    ],
],
```

На локальном сервере всё работает как ожидается, но на продакшене, где настроен https, данный роут перестаёт распознаваться, т.к. там `getHostInfo()` уже возвращает https://you-domain.net.

Я пробовал указывать значение без протокола (то есть `"you-domain.net/some/path" => 'controller/action'`), но в таком случае роут вовсе не был добавлен в UrlManager.
### What is the expected result?

Если я не укажу протокол, то сверка роута должна происходить только по имени хоста, без проверки протокола.
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.9 |
| PHP version | 7.0.11 |
| Operating system | Docker 1.12.1 (php:7.0.11-fpm) |

rob006:
Probably related to #3047.

samdark:
Nope, it's different. The problem here is that in domain-relative URL rules you have to specify http or https and can't use something which works for both.

erickskrauch:
@samdark, я близок к завершению PR. Только я не привязывался только к http или https, а делаю поддержку любого протокола.

klimov-paul:
Use optional `s` for protocol specification:

```
'rules' => [
        "http(s)?://you-domain.net/some/path" => 'controller/action',
    ],
```

rob006:
IMO `UrlRule` should have property like `$validSchemas` with value like `['https', 'http']`. So if we have rules like:

``` php
['pattern' => 'login', 'validSchemas' => ['https'], ...],
['pattern' => 'http://exaple.com/home', 'validSchemas' => ['https', 'http'], ...]
```

then:
* `https://example.com/login` will match first rule
* `http://example.com/login` will NOT match first rule
* `http://example.com/home` will match second rule
* `https://example.com/home` will match second rule

`UrlNormalizer` could use `UrlRule::$validSchemas` to handle redirection to valid schema.

erickskrauch:
[...]

PR #13525, opened 2017-02-06
added failing test for #13522

cebe:
added failing test for #13522

regression after 2.0.11 release.

~~Working on a fix.~~ - done.

### DIFF
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -403,6 +403,7 @@ class UrlManager extends Component
                         return $url . $baseUrl . $anchor;
                     }
                 } else {
+                    $url = ltrim($url, '/');
                     return "$baseUrl/{$url}{$anchor}";
                 }
             }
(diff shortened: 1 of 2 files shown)


cebe:
@rob006 as you have been involved in UrlManager changes in the past, you might want to take a look at this one.

rob006:
I think that #12691 created this bug and this should be fixed in `UrlRule`.

cebe:
I do not see how exactly it introduced it but I agree that it may be better to fix it in UrlRule.

rob006:
I checked again and I think I'm wrong. This is not regression, this bug exist even in 2.0.5. In that case this fix looks valid to me.

framework/web/UrlManager.php
```diff
@@ -414,6 +415,7 @@ public function createUrl($params)
                 $route .= '?' . $query;
             }
 
+            $route = ltrim($route, '/');
```
rob006: This is not needed, `$route` is already trimmed: https://github.com/cebe/yii2/blob/9a53571fd2cac0155477bef8420a74b0b665f2c7/framework/web/UrlManager.php#L360

cebe: this fixes the case where route is empty and a `/` suffix is used. see the test case added below (line 447).

samdark (approved):

rob006 (approved):

SilverFire (approved):

samdark:
@cebe merge?

Issue #13475, opened 2017-01-30
UniqueValidator with two targetAttribute not work after update


orcoliver:
Hi,
In a model i have

```php
[['reo_orden',` 'reo_articulo'], 'unique', 'targetAttribute' => ['reo_orden', 'reo_articulo'], 'message' => 'El articulo ya está `cargado.'],
```

Before update the debugger show:

```sql
SELECT EXISTS(SELECT * FROM `cop_renord` WHERE (`reo_orden`='14') AND (`reo_articulo`='001401239'))
```

After update:

```sql
SELECT EXISTS(SELECT * FROM `cop_renord` WHERE (`reo_orden`='14') AND (`reo_articulo`='14'))
```

Composer update output:

```
Updating yiisoft/yii2 dev-master (03e0607 => 221603a) Checking out 221603ae2c
```

composer.json after update:

```json
"minimum-stability": "dev",
    "require": {
        "php": ">=5.4.0",
        "yiisoft/yii2": ">=2.0.5",
```

PHP version: 5.6.22

Thks for help.

Alex-Code:
Could be from this change, https://github.com/yiisoft/yii2/commit/b82caa97e82ff8b99389b56958fb1e1310637c4b#diff-b7eeda2fac6ce8131377ce36953a329aR227

developeruz:
targetAttribute related to validate attribute if another isn't set. 
`[['reo_orden', 'reo_articulo'], 'unique', 'targetAttribute' => ['reo_orden', 'reo_articulo' => 'reo_articulo'], 'message' => 'El articulo ya está cargado.'],`

cebe:
> 'reo_articulo' => 'reo_articulo'

this was equal to simply specifying `'reo_articulo'`, is that still the case? If not, we should fix that.

developeruz:
Here the PR for fix BC, but I believe the current simplifying is wrong, it makes people confusing. Please look at #7820 
If I have written:
`
[ 'attr1',  'unique', 'targetAttribute' => ['attr2', 'attr3'] ],
`
current simplifying mean `'attr2' => 'attr2' and 'attr3' => 'attr3'`. The reasonable question is where attr1 is used even if `unique` specified for attr1? I suggest use simplifying like `'attr1' => attr2` if another is not specified. @cebe, if you agree with this I will create PR and mark it as BC breaking.

cebe:
[...]

PR #13318, opened 2017-01-02
Add OR relation for targetAttribute in UniqueValidator

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7820

1) Bug fix:
`['name', 'unique', 'targetClass' => User::className(), 'targetAttribute' => 'realname']`,
compares that in DB doesn't exist row with `realname=$model->name`
`['name', 'unique', 'targetClass' => User::className(), 'targetAttribute' => [‘realname’]]`,
used to compare that in DB doesn’t exist row with `realname=$model->realname` and $model->name doesn’t affect the validation at all. 
After my fix, if a key is not specified in targetAttributes, validator compares with validated attribute values.
`['name', 'unique', 'targetClass' => User::className(), 'targetAttribute' => [‘realname’]]`,
means `realname=$model->name`
`['name', 'unique', 'targetClass' => User::className(), 'targetAttribute' => [‘realname’ => ‘realname’]]`,
means `realname=$model->realname`

2) New feature: 
All targetAttribute values were combined with AND condition
`['name', 'unique', 'targetClass' => User::className(), 'targetAttribute' => ['realname', 'nick', 'altname']]`
checks that in DB exist row where `realname=name AND nick=name AND altname=name`
$combineType allows user to change condition type
`['name', 'unique', 'targetClass' => User::className(), 'targetAttribute' => ['realname', 'nick', 'altname’], ‘combineType’ => ‘or’]`
checks that in DB row exists where `realname=name OR nick=name OR altname=name`

PR #13534, opened 2017-02-07
Fix BC in Exist Validator (#13533)

### LINKED ISSUE #13533: Possible BC break in ExistValidator

https://github.com/yiisoft/yii2/commit/b82caa97e82ff8b99389b56958fb1e1310637c4b#diff-c36c8f68f6255dc8840038b8c921813eL54

> ~If the key and the value are the same, you can just specify the value.~

Upgrade Yii 2.0.10 => 2.0.11.1

```php
/**
 * @property integer $foo_id
 */
class A extends ActiveRecord {
}

/**
 * @property integer $foo_id
 * @property integer $a_id
 */
class B extends ActiveRecord {
    public function rules()
    {
        return [
            [
                ['a_id'],
                'exist',
                'targetClass' => A::class,
                'targetAttribute' => ['a_id' => 'id', 'foo_id'],  // shortcut 'foo_id' as documented before
            ],
        ];
    }
}
```

So, `B::foo_id == A::foo_id` and `B::a_id == A::id`

A:

id | foo_id
-- | --
42 | 997

```php
$b = new B();
$b->foo_id = 997;
$b->a_id = 42;

if (!$b->save()) {
    \Yii::error($b->errors);
    throw ...;
}
```

```mysql
SELECT EXISTS(SELECT * FROM `a` WHERE (`id`=42) AND (`foo_id`=997))
```

Validation passed.

### What do you get instead?

```mysql
SELECT EXISTS(SELECT * FROM `a` WHERE (`id`=42) AND (`foo_id`=42))
```

Validation failed: `Attribute "a_id" is incorrect`.

### Additional info

Fix code as such:

```diff
- [['a_id'], 'exist', 'targetClass' => A::class, 'targetAttribute' => ['a_id' => 'id', 'foo_id']]
+ [['a_id'], 'exist', 'targetClass' => A::class, 'targetAttribute' => ['a_id' => 'id', 'foo_id' => 'foo_id']]
```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.1
| PHP version      | 7.0.13

developeruz:
It was about simplifying targetAttribute (have discussed in #13475). Related changes from #13318 rolled back.

fixes #13533

### DIFF
--- a/framework/validators/ExistValidator.php
+++ b/framework/validators/ExistValidator.php
@@ -125,6 +125,7 @@ class ExistValidator extends Validator
      * of the attribute currently being validated. You may use an array to validate the existence
      * of multiple columns at the same time. The array key is the name of the attribute with the value to validate,
      * the array value is the name of the database field to search.
+     * If the key and the value are the same, you can just specify the value.
      * @param \yii\base\Model $model the data model to be validated
      * @param string $attribute the name of the attribute to be validated in the $model
      * @return array conditions, compatible with [[\yii\db\Query::where()|Query::where()]] key-value format.
@@ -138,7 +139,7 @@ class ExistValidator extends Validator
             }
             $params = [];
             foreach ($targetAttribute as $k => $v) {
-                $params[$v] = is_int($k) ? $model->$attribute : $model->$k;
+                $params[$v] = is_int($k) ? $model->$v : $model->$k;
             }
         } else {
             $params = [$targetAttribute => $model->$attribute];


ddinchev:
Shouldn't we make sure we have tests for new code as a policy so that we run into such situations less often?

cebe:
@ddinchev sure that is the plan but it did not help in this case. Tests were written for the new changes but not for the old behavior. That is the main problem we have with breaking changes. Seen in this release. We need more test coverage for existing features to make sure they do not break.

cebe:
There is a test for this code but it does not catch this change because of bad choice of IDs...
https://github.com/yiisoft/yii2/blob/a941b3536f5d53ab4c791705a7d40ef7ceab8c1b/tests/framework/validators/ExistValidatorTest.php#L136-L141

cebe (approved):

cebe:
failing test added in fba4e434695b7d25629967cbc7281ced7bbe03ab, merged. Thank you!

ddinchev:
@cebe cool - I was asking for the current PR change. Thank you for fixing this so quickly!

PR #13542, opened 2017-02-08
Added MySQL TEXT types

NickvdMeij:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | (none that I know about, and yes I searched the issues ;-) )

Basically a simple addition for MySQL users to use MEDIUMTEXT if they want to (for example to store a large raw json response if nessesary)

### DIFF
--- a/framework/db/mysql/Schema.php
+++ b/framework/db/mysql/Schema.php
@@ -20,6 +20,26 @@ use yii\db\ColumnSchema;
 class Schema extends \yii\db\Schema
 {
     /**
+     * @var string - MYSQL TINYTEXT data type. 255 bytes (characters)
+     */
+    const TYPE_TINYTEXT = 'tinytext';
+
+    /**
+     * @var string - MYSQL TINYTEXT data type. 65,535 bytes (characters) / 64KB
+     */
+    const TYPE_TEXT = 'text';
+
+    /**
+     * @var string - MYSQL TINYTEXT data type. 16,777,215 bytes (characters) / 16MB
+     */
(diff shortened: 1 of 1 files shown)


cebe:
This is about the schema information part, why do you need the distinction between text and mediumtext here? In migrations you can already use it and your PR does not change anything abou that :)

NickvdMeij:
@cebe its not that you cant already use it, its just a simple addition to be able to use `Schema::TYPE_MEDIUMTEXT` if you want to be able to store more characters in the field. At the moment we are using this, but since there is no schema const for it we need to insert a string with `"mediumtext"`. Like i said, its just convinient

cebe:
I know that there is a difference but the purpose of schema constants is to write dbms independent code. If you write migrations for mysql only, writing the column type as a string is the right way to go. Thanks for the proposal, but we are not going to accept this PR.

NickvdMeij:
@cebe oke correct me if I'm wrong, but the dedicated mysql schema class can't have dedicated mysql settings? dbms independent code should be in the abstract Schema class right?... The translation is made using the typeMap as far as I can see...

cebe:
> but the dedicated mysql schema class can't have dedicated mysql settings?

The idea of this is to return a common set of types independent of the dbms, so for example applications like gii can use it to determine which rules to generate for a db column.
We would not need the conversion and constants if we'd use the type from the db directly.

Issue #13493, opened 2017-02-01
Dynamically loading nested pjax


friek108:
### What steps will reproduce the problem?
I have a main pjax container that I use to navigate between pages.  One one of these pages, I dynamically load in a page and the page includes some content and a grid. The grid itself is wrapped in a pjax widget.

When I set a filter on the grid, a request is sent to update the outer pjax container which is immediately cancelled and replaced by a request to update the nested grid pjax. The nested container updates correctly.

When I click on a sort link however, the outer container is used and updated with only the 'grid' (leaving my other page content to be replaced). 

Further Info:
- If I load the grid page through a browser refresh, everything works as expected. It is only when you load the 'grid page' dynamically (in the outer container).
- skipOuterContainers is set to true in the grid pjax wrapper.
- From the controller, I use:
`if (Yii::$app->request->isPjax) return $this->renderAjax('myView'); 
else return $this->render('myView')`
This is because I need to load some dynamically loaded scripts (for eg. yii.Gridview.js and others).
- I'm using a different filterUrl on the grid to return only the grid content (rather than the grid and the other content on the page).

### What is the expected result?
Nested container updates.
### What do you get instead?
Outer container updates and is replaced by the nested container content.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11-dev
| PHP version      | 5.5.12
| Operating system | Windows

samdark:
It sounds rather vague about how to reproduce it. Would you please supply us with more details? If you'll start with basic application template and do as little changes as possible while still getting the issue, it would be perfect. Thanks.

friek108:
I've created an example from basic application template [here](http://pjax.portablehive.com). You can also view the respository [here](https://github.com/friek108/yii2-basic).

To reproduce: 
1. Go to the about page, then refresh it. After refresh, sort the grid, works as expected.
2. Go to home page, then refresh it. Then navigate to 'about' page. Sort the grid. Uses the outer container.

**Modified files**
- /controllers/SiteController.php (about and about-grid actions only)
- /views/layouts/main
- /views/site/about
- /views/site/about-grid
- /config/web.php & .htaccess (for pretty urls only)

friek108:
Perhaps nested Pjax should be considered just plain Ajax and handled separately?

In general, it's hard to know the correct way to design and optimise an application beyond simple use cases. Eg. Loading js assets on separate pages (in a container), using a menu which might be outside a pjax container, nested pjax, etc.

[...]

PR #13545, opened 2017-02-09
Fixes all issues related to nested pjax. 

derekisbusy:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13493

Requires pull request yiisoft/jquery-pjax/pull/48

This fix will make sure the `$(document).pjax(...)` calls are made in the right order which allows for nested pjax widgets.

### DIFF
--- a/framework/web/View.php
+++ b/framework/web/View.php
@@ -88,6 +88,21 @@ class View extends \yii\base\View
      * This is internally used as the placeholder for receiving the content registered for the end of the body section.
      */
     const PH_BODY_END = '<![CDATA[YII-BLOCK-BODY-END]]>';
+    /**
+     * How the registered JavaScript code block should be included in the JavaScript section.
+     * This means the JavaScript code block will replace the current code block if the code block exists.
+     */
+    const MERGE_REPLACE = 1;
+    /**
+     * How the registered JavaScript code block should be included in the JavaScript section.
+     * This means the JavaScript code block will be prepended to the current code block if the code block exists.
+     */
+    const MERGE_PREPEND = 2;
+    /**
+     * How the registered JavaScript code block should be included in the JavaScript section.
+     * This means the JavaScript code block will be appended to the current code block if the code block exists.
+     */
+    const MERGE_APPEND = 3;
 
     /**
      * @var AssetBundle[] list of the registered asset bundles. The keys are the bundle names, and the values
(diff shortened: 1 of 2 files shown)


derekisbusy:
I added the property `isLayout` because there is a caveat when using PJAX  widgets in the layout file when there are nested PJAX widgets in the content view files.  This is because the content views get rendered before the layout therefore the order of nesting is reversed in the layout file.  This means PJAX widgets in the layout files should set the `isLayout` property to true.

This pull request will also fix issues with PJAX not working on subsequent page calls. For example:

```php
Pjax::begin(['id' => 'my-list-view']);
echo ListView::widget(...);
Pjax::end();
```

Right now only the links on page 1 will work. 

Also it should be noted that you should always set an ID and not rely on the automated ID when nesting PJAX widgets.

dynasource:
@derekisbusy, thanks for your effort. 

because of your comment https://github.com/yiisoft/yii2/issues/13493#issuecomment-277645391 and the huge workload it costs to handle all the PJAX issues in the Yii2 forum, we have decided to focus and not to accept any further development on PJAX. As you already stated yourself, it takes too many hours to work with it (and you are not alone in this). We are deprecating in the 2.0.* branche and drop it with the release of 2.1.0.

Of course, you are free to continue your efforts in your repo, perhabs even come up with your own rewrite of Pjax.

derekisbusy:
@dynasource Actually I should update that comment as it's incorrect. I've managed to fix all the bugs related to pjax without rewriting it. 

I would ask to reconsider since the real problem with the pjax widget is it was never fully implemented.  
If you merge all the following pull requests:

yiisoft/jquery-pjax#48
yiisoft/jquery-pjax#50
yiisoft/jquery-pjax#52

And this one #13545 it fixes all the issues. I have tested it with:

- [x] Pjax in modal that works after closing and reopening modal
- [x] Pjax with Activeform that doesn't submit twice (bug related to nested pjax)
- [x] Pjax with Activeform and enableAjaxValidation
- [x] Pjax with subsequest calls, I suspect many of the grid/listview issues had to do with this.
- [x] Pjax with stylesheets
- [x] Pjax within Pjax, I've tested 5 layers including pjax in layout and works fine.

You should at least merge these last pull requests as it fixes all the issues and finally *properly* implements the pjax widget.

Also included in the pull is two extra parameters to the `registerJs()` method that would be needed to get any widget to work properly with other related widgets when needed. For example the `jsAfter` option makes sure Pjax is attached after activeform so that ajax validation happens before pjax submit. With the `mergerType` and `jsAfter` you can make sure that widgets are initialized in the proper order.

dynasource:
It not that we we don't trust the quality of your work, but the problem with PJAX lies deeper than the patches you have created. The problem with PJAX is that it has not been designed for a lot of usecases we've explored in the previous years. Pjax needs a complete redesign in order for it to be stable & usuable enough, which is also explictly stated by the original creator. Whatever patch you may submit, its dwelling with an open tap.

As core team we finally concluded that it was a road with a dead end, asking too much support in our issue tracker https://github.com/yiisoft/yii2/issues?utf8=%E2%9C%93&q=pjax
Therefore we decided to focus on the core software again and to move time consuming software parts to third party extensions.

However, your efforts to improve the Pjax show a valid case for continued support in a separate extension.  We would love to see you continue in your work and are happy to point the Community to your extension from the moment the we are dropping it from the framework. I hope you can understand our reasoning.

derekisbusy:
Can it at least be merged into a 2.0.X release? There are very basic use cases that get fixed here like: 
```php
Pjax::begin(['id' => 'my-list-view']);
echo ListView::widget(...);
Pjax::end();
```

I will add some of the changes to a separate pull because the only way to get this widget working even in separate repo requires updating the `\yii\web\View` class.

dynasource:
> There are very basic use cases that get fixed here like:
> 
> Pjax::begin(['id' => 'my-list-view']);
> echo ListView::widget(...);
> Pjax::end();

that usecase should be working fine already

derekisbusy:
It's not working right now because the client scripts are not being registered on Pjax request..

https://github.com/yiisoft/yii2/blob/master/framework/widgets/Pjax.php#L151

dynasource:
its working fine for a simple refresh. If you want to load `new` clientScripts than you are going into too much depth already with the current state of the architecture. Whatever solution or patch you will provide, this current state will cause developers too burn hours on debugging. 

It should be rebuild from scratch to have a product which is flexible, stable and userfriendly for developers to use.

alessio6181:
Now my view search in modal work fine.
Thank you.

Issue #13155, opened 2016-12-07
widget's CSS is not inserted with pjax


hastalavista-baby:
When I use Pjax, widget's CSS is not inserted in the page

registerCssFile

//page1.php
```php
<?php Pjax::begin(['timeout' => 5000]); ?>
       <a href="/page2.php">Page 2</a>
<?php Pjax::end(); ?>
```

//page2.php
```php
<?php Pjax::begin(['timeout' => 5000]); ?>
       <?php echo MyWidget::widget(); ?>

       <a href="/page1.php">Page 1</a>
<?php Pjax::end(); ?>
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 7.1
| Operating system | Windows

dynasource:
this is expected behavior as the Pjax widget exits directly after execution durign the ajax call. If you want to have the CSS inserted, you have to output it directly between the opening and closing tag of the Pjax widget (and not use registerCssFile).

hastalavista-baby:
how do it?
so that my widget with pjax is not working. right?

dynasource:
echo your CSS between the `Pjax::begin` and  `Pjax::end`

hastalavista-baby:
this is not correct. 
and if a lot of widgets. all styles write?
why registerJsFile working fine with pjax?

dynasource:
the current Pjax widget only supports endPage scripts, as can be seen at https://github.com/yiisoft/yii2/blob/master/framework/widgets/Pjax.php#L148

CSS is not loaded, I guess because it could change the pages representation

hastalavista-baby:
why I think it's a bug

dynasource:
the Pjax widget in Yii is to be used for pjax widgets which are already loaded in the current page. For this usecase, the CSS would also already have been loaded. 

If you are using Pjax for more complex scenario's with many more Pjax widgets, please understand that this not what Pjax is created for. A lot on this subject has already been said in this issue tracker and the Pjax issue tracker. Please do some searches to find more information.

hastalavista-baby:
Thanks

derekisbusy:
@dynasource the pjax widget does include the head...

https://github.com/yiisoft/yii2/blob/master/framework/widgets/Pjax.php#L126

I added a fix for the stylesheets issue.

yiisoft/jquery-pjax/pull/49

[...]

PR #13546, opened 2017-02-09
Fix stylesheets so that they only get applied once unless specified o…

derekisbusy:
…therwise.

Adds two new options:
 - **$reapplyStyles** Reapply existing stylesheets if set to true.
 - **$stylesHead** Moves stylesheets to the HTML head element so that they are not removed on subsequent PJAX calls.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #2310, #13155

Requires: yiisoft/jquery-pjax/pull/49

### DIFF
--- a/framework/widgets/Pjax.php
+++ b/framework/widgets/Pjax.php
@@ -98,6 +98,14 @@ class Pjax extends Widget
      */
     public $clientOptions;
     /**
+     * @var bool Reapply stylesheets if the stylesheet is already loaded. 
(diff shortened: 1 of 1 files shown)


dynasource:
see https://github.com/yiisoft/yii2/pull/13545#issuecomment-280837915

PR #13553, opened 2017-02-10
Issue 13551: loading fixtures in subdirectories

### LINKED ISSUE #13551: fixture/load ignores fixtures in subdirectories

Problem involves fixture search process on `fixture/load` ( `yii\console\controllers\FixtureController::actionLoad()` ).
Let's say we have these fixtures under fixture namespace:
```
fooFixture.php
barFixture.php
sub\bazFixture.php
sub\fooFixture.php
```
The confirmation dialog would find all of them, but *strips everything to basename()*:
```
foo
bar
baz
foo
```
On confirmation, it would try to find & load them *directly* under fixture namespace, ignoring subdirectories.
Instead of stripping found fixtures to basename(), their relative path sould be used (relatively to namespace dir). For aforementioned example, expected result is:
```
foo
bar
sub\baz
sub\foo
```
### What do you get instead?
```
foo
bar
baz
foo
```

### Additional info

Mentioned *expected result*  raises another problem in `FixtureController::getFixtureConfig()`: flag `isNamespaced` is calculated by presence of slashes in fixture name, whereas fixtures in subdirectories would have slashes and have no namespace.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.5
| PHP version      | 7.0.13

d1rtyf1ng3rs:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | -
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13551 

An attempt to fix #13551.
Would like hear some comments / critics first, expecially about `getFixturesConfig()` changes.

### DIFF
--- a/framework/console/controllers/FixtureController.php
+++ b/framework/console/controllers/FixtureController.php
@@ -415,18 +415,33 @@ class FixtureController extends Controller
         $foundFixtures = [];
 
         foreach ($files as $fixture) {
-            // strip $fixturesPath from $fixture's full path
-            $relativeName = str_replace($fixturesPath . "/", "", $fixture);
-            // get fixtures's directory
-            $relativeDir = dirname($relativeName) == '.' ? '' : dirname($relativeName) . '/';
-            // get fixture name relatively to $fixturesPath
-            $relativeName = $relativeDir . basename($fixture, 'Fixture.php');
-            // normalize path to unix format
-            $relativeName = str_replace("\\", "/", $relativeName);
+            $relativeName = $this->getFixtureRelativeName($fixture);
             $foundFixtures[] = $relativeName;
         }
 
         return $foundFixtures;
+    }
+
+    /**
+     * Calculates $fixture's name relatively to $templatePath.
+     * Basically, strips getFixturePath() and 'Fixture.php' prefix from fixture's full path
+     * @see getFixturePath()
+     * @param $fixture string Full fixture path
+     * @return string Relative fixture name
+     *
+    */
+    private function getFixtureRelativeName($fixture) {
+        // $fixturesPath is normalized to unix format in getFixturesPath()
+        $fixturesPath = $this->getFixturePath();
+        // normalize $fixture to unix format
+        $fixture = str_replace("\\", "/", $fixture);
+        // strip $fixturesPath from $fixture's full path
+        $relativeName = str_replace($fixturesPath . "/", "", $fixture);
+        // get fixtures's directory
+        $relativeDir = dirname($relativeName) === '.' ? '' : dirname($relativeName) . '/';
+        // get fixture name relatively to $fixturesPath
+        $relativeName = $relativeDir . basename($fixture, 'Fixture.php');
+        return $relativeName;
     }
 
     /**
--- a/framework/console/controllers/FixtureController.php
+++ b/framework/console/controllers/FixtureController.php
@@ -421,6 +421,8 @@ class FixtureController extends Controller
             $relativeDir = dirname($relativeName) == '.' ? '' : dirname($relativeName) . '/';
             // get fixture name relatively to $fixturesPath
             $relativeName = $relativeDir . basename($fixture, 'Fixture.php');
+            // normalize path to unix format
+            $relativeName = str_replace("\\", "/", $relativeName);
             $foundFixtures[] = $relativeName;
         }
 
@@ -437,18 +439,13 @@ class FixtureController extends Controller
         $config = [];
 
         foreach ($fixtures as $fixture) {
-            // $isNamespaced = (strpos($fixture, '\\') !== false);
-            // this would have incorrect result on windows-based systems in case of fixtures placed in subdirectories
+            $isNamespaced = (strpos($fixture, '\\') !== false);
+            // replace linux' path backslashes to namespace slashes, in case if $fixture is non-namespaced relative path
+            $fixture = str_replace('/', '\\', $fixture);
+            $fullClassName = $isNamespaced ? $fixture . 'Fixture' : $this->namespace . '\\' . $fixture . 'Fixture';
 
-            // istead of guessing if $fixture is namespaced, let's just check it _both_ as non-namespaced and namespaced
-            $className = $fixture . 'Fixture';
-            if (class_exists($className)) {
-                $config[] = $className;
-            } else {
-                $className = $this->namespace . '\\' . str_replace('/','\\',$fixture) . 'Fixture';
-                if (class_exists($className)) {
-                    $config[] = $className;
-                }
+            if (class_exists($fullClassName)) {
+                $config[] = $fullClassName;
             }
         }
 
--- a/framework/console/controllers/FixtureController.php
+++ b/framework/console/controllers/FixtureController.php
@@ -428,9 +428,9 @@ class FixtureController extends Controller
      * @see getFixturePath()
      * @param string $fullFixturePath Full fixture path
      * @return string Relative fixture name
-     *
-    */
-    private function getFixtureRelativeName($fullFixturePath) {
+     */
+    private function getFixtureRelativeName($fullFixturePath)
+    {
         // $fixturesPath is normalized to unix format in getFixturesPath()
         $fixturesPath = $this->getFixturePath();
         // normalize $fixture to unix format
@@ -455,7 +455,7 @@ class FixtureController extends Controller
 
         foreach ($fixtures as $fixture) {
             $isNamespaced = (strpos($fixture, '\\') !== false);
-            // replace linux' path backslashes to namespace slashes, in case if $fixture is non-namespaced relative path
+            // replace linux' path slashes to namespace backslashes, in case if $fixture is non-namespaced relative path
             $fixture = str_replace('/', '\\', $fixture);
             $fullClassName = $isNamespaced ? $fixture . 'Fixture' : $this->namespace . '\\' . $fixture . 'Fixture';
 
--- a/framework/console/controllers/FixtureController.php
+++ b/framework/console/controllers/FixtureController.php
@@ -426,21 +426,21 @@ class FixtureController extends Controller
      * Calculates $fixture's name relatively to $templatePath.
      * Basically, strips getFixturePath() and 'Fixture.php' prefix from fixture's full path
      * @see getFixturePath()
-     * @param $fixture string Full fixture path
+     * @param string $fullFixturePath Full fixture path
      * @return string Relative fixture name
      *
     */
-    private function getFixtureRelativeName($fixture) {
+    private function getFixtureRelativeName($fullFixturePath) {
         // $fixturesPath is normalized to unix format in getFixturesPath()
         $fixturesPath = $this->getFixturePath();
         // normalize $fixture to unix format
-        $fixture = str_replace("\\", "/", $fixture);
+        $fullFixturePath = str_replace("\\", "/", $fullFixturePath);
         // strip $fixturesPath from $fixture's full path
-        $relativeName = str_replace($fixturesPath . "/", "", $fixture);
+        $relativeName = str_replace($fixturesPath . "/", "", $fullFixturePath);
         // get fixtures's directory
         $relativeDir = dirname($relativeName) === '.' ? '' : dirname($relativeName) . '/';
         // get fixture name relatively to $fixturesPath
-        $relativeName = $relativeDir . basename($fixture, 'Fixture.php');
+        $relativeName = $relativeDir . basename($fullFixturePath, 'Fixture.php');
         return $relativeName;
     }
 
--- a/framework/console/controllers/FixtureController.php
+++ b/framework/console/controllers/FixtureController.php
@@ -415,7 +415,13 @@ class FixtureController extends Controller
         $foundFixtures = [];
 
         foreach ($files as $fixture) {
-            $foundFixtures[] = basename($fixture, 'Fixture.php');
+            // strip $fixturesPath from $fixture's full path
+            $relativeName = str_replace($fixturesPath . "/", "", $fixture);
+            // get fixtures's directory
+            $relativeDir = dirname($relativeName) == '.' ? '' : dirname($relativeName) . '/';
+            // get fixture name relatively to $fixturesPath
+            $relativeName = $relativeDir . basename($fixture, 'Fixture.php');
+            $foundFixtures[] = $relativeName;
         }
 
         return $foundFixtures;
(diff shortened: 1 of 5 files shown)


framework/console/controllers/FixtureController.php
```diff
@@ -431,11 +437,18 @@ private function getFixturesConfig($fixtures)
         $config = [];
 
         foreach ($fixtures as $fixture) {
-            $isNamespaced = (strpos($fixture, '\\') !== false);
-            $fullClassName = $isNamespaced ? $fixture . 'Fixture' : $this->namespace . '\\' . $fixture . 'Fixture';
+            // $isNamespaced = (strpos($fixture, '\\') !== false);
```
samdark: I think we can check for both `/` and `\` here or normalize everything to linux paths format before the check and the problem will be solved.

framework/console/controllers/FixtureController.php
```diff
@@ -431,11 +437,18 @@ private function getFixturesConfig($fixtures)
...
+            // this would have incorrect result on windows-based systems in case of fixtures placed in subdirectories
 
-            if (class_exists($fullClassName)) {
-                $config[] = $fullClassName;
+            // istead of guessing if $fixture is namespaced, let's just check it _both_ as non-namespaced and namespaced
+            $className = $fixture . 'Fixture';
+            if (class_exists($className)) {
```
samdark: If both namespaced and non-namespaced fixtures with the same base name exist there will be conflict.

samdark (changes requested):

d1rtyf1ng3rs:
Added fixture's relative path normalization to linux format in `findFixtures()` method;
reverted changes to `getFixturesConfig()` to its original (so `$isNamespaced` is calculated correctly) and added slash-replace from linux-path-backslashes to namespace slashes.

framework/console/controllers/FixtureController.php
d1rtyf1ng3rs: Added one more commit - issue should be solved now.
Added fixture's relative path normalization to linux format in `findFixtures()` method;
reverted changes to `getFixturesConfig()` to its original (so `$isNamespaced` is calculated correctly) and added slash-replace from linux-path-backslashes to namespace slashes.

framework/console/controllers/FixtureController.php
```diff
@@ -415,7 +415,15 @@ private function findFixtures(array $fixtures = [])
...
         foreach ($files as $fixture) {
-            $foundFixtures[] = basename($fixture, 'Fixture.php');
+            // strip $fixturesPath from $fixture's full path
+            $relativeName = str_replace($fixturesPath . "/", "", $fixture);
+            // get fixtures's directory
+            $relativeDir = dirname($relativeName) == '.' ? '' : dirname($relativeName) . '/';
+            // get fixture name relatively to $fixturesPath
+            $relativeName = $relativeDir . basename($fixture, 'Fixture.php');
+            // normalize path to unix format
+            $relativeName = str_replace("\\", "/", $relativeName);
```
samdark: Shouldn't it be done to `$fixture` prior to replacements right before `// strip $fixturesPath from $fixture's full path` line?

framework/console/controllers/FixtureController.php
```diff
@@ -415,7 +415,15 @@ private function findFixtures(array $fixtures = [])
         $foundFixtures = [];
 
         foreach ($files as $fixture) {
-            $foundFixtures[] = basename($fixture, 'Fixture.php');
+            // strip $fixturesPath from $fixture's full path
+            $relativeName = str_replace($fixturesPath . "/", "", $fixture);
+            // get fixtures's directory
+            $relativeDir = dirname($relativeName) == '.' ? '' : dirname($relativeName) . '/';
```
samdark: `===` can be used instead of `==`

samdark (changes requested):

framework/console/controllers/FixtureController.php
```diff
@@ -415,7 +415,15 @@ private function findFixtures(array $fixtures = [])
         $foundFixtures = [];
 
         foreach ($files as $fixture) {
-            $foundFixtures[] = basename($fixture, 'Fixture.php');
+            // strip $fixturesPath from $fixture's full path
```
dynasource: the added logic should be ensapsulated in a private method with the comments in the description

dynasource (changes requested):
I would like to see your usecase examples covered by unit tests

framework/console/controllers/FixtureController.php
d1rtyf1ng3rs: Good point. Yes, there might be a problem - `$fixture` is calculated automatically, whereas `$fixturesPath` is provided or may be provided by user. So it's not guaranteed they're in the same format.
I'll provide corresponding changes soon.

framework/console/controllers/FixtureController.php
d1rtyf1ng3rs: > ... unit tests

Roger that; will provide soon.

> ... should be encapsulated in a private method

Why?

d1rtyf1ng3rs:
> ... unit tests

Roger that; will provide soon.

framework/console/controllers/FixtureController.php
dynasource: to see 'what' is does by reading the code instead of reading 'how' it is done

d1rtyf1ng3rs:
Welp, it wasn't much of a 'soon', but here it is.
- Unit tests added;
- getRelativeName() method extracted;
- minor path normalization changes.

I have, still, left the inline-comments in getRelativeName() method; feel free to remove them, if those are 'too much'.

samdark (approved):

tests/data/console/controllers/fixtures/subdir/SecondFixture.php
```diff
@@ -0,0 +1,20 @@
...
+namespace yiiunit\data\console\controllers\fixtures\subdir;
+
+use yii\test\Fixture;
+use yiiunit\data\console\controllers\fixtures\FixtureStorage;
+
+class SecondFixture extends Fixture
+{
+    public function load()
+    {
+        FixtureStorage::$subdirSecondFixtureData[] = 'some data set for subdir/second fixture';
```
dynasource: I dont see this `string` being used anywhere? Shouldnt it be used in on of the unit tests?

tests/data/console/controllers/fixtures/subdir/FirstFixture.php
```diff
@@ -0,0 +1,20 @@
...
+namespace yiiunit\data\console\controllers\fixtures\subdir;
+
+use yii\test\Fixture;
+use yiiunit\data\console\controllers\fixtures\FixtureStorage;
+
+class FirstFixture extends Fixture
+{
+    public function load()
+    {
+        FixtureStorage::$subdirFirstFixtureData[] = 'some data set for first fixture in subdir';
```
dynasource: I dont see this string being used anywhere? Shouldnt it be used in on of the unit tests?

framework/console/controllers/FixtureController.php
```diff
@@ -415,13 +415,36 @@ private function findFixtures(array $fixtures = [])
...
+        // normalize $fixture to unix format
+        $fixture = str_replace("\\", "/", $fixture);
+        // strip $fixturesPath from $fixture's full path
+        $relativeName = str_replace($fixturesPath . "/", "", $fixture);
+        // get fixtures's directory
+        $relativeDir = dirname($relativeName) === '.' ? '' : dirname($relativeName) . '/';
+        // get fixture name relatively to $fixturesPath
+        $relativeName = $relativeDir . basename($fixture, 'Fixture.php');
+        return $relativeName;
+    }
```
dynasource: it's hard to determine the quality of the logic without a dedicated unittest. It would be nice to have one for a proper understanding.

framework/console/controllers/FixtureController.php
```diff
@@ -415,13 +415,36 @@ private function findFixtures(array $fixtures = [])
...
     /**
+     * Calculates $fixture's name relatively to $templatePath.
+     * Basically, strips getFixturePath() and 'Fixture.php' prefix from fixture's full path
+     * @see getFixturePath()
+     * @param $fixture string Full fixture path
+     * @return string Relative fixture name
+     *
+    */
+    private function getFixtureRelativeName($fixture) {
+        // $fixturesPath is normalized to unix format in getFixturesPath()
```
dynasource: the argument naming does not match the @param doc. Currently I assume an object $fixture being passed.

dynasource (changes requested):

tests/data/console/controllers/fixtures/subdir/SecondFixture.php
d1rtyf1ng3rs: Well, the string (the value) itself *can*, but *doesn't have to* be used in unit tests.
The property, howerer, [is used in unit-tests](https://github.com/yiisoft/yii2/pull/13553/files/719f6aee4856d2e5bbd9b44100955d9e34f94573#diff-410e9261d60564f099dc41e3dd30a538).
P.s. In fact, all the added fixtures and their usage in tests are *the same* as of existing fixtures.

framework/console/controllers/FixtureController.php
d1rtyf1ng3rs: Negative.
`$fixture` is a string; one of the strings that `FileHelper::findFiles()` returns [here](https://github.com/yiisoft/yii2/pull/13553/files/719f6aee4856d2e5bbd9b44100955d9e34f94573#diff-96190cbb10ccc75fed657c73f25ae920L414).

framework/console/controllers/FixtureController.php
d1rtyf1ng3rs: The new function `getFixtureRelativeName()` is triggered on both `load`/`unload` actions, so is involved in all tests that use those actions.
Yet, I agree - its' tests are rather "not obvious" / "non-direct".

If dedicated tests for it are absolutely necessary, I would need some help in form of hints, guidelines or requirements for them. Or any description of what is the "good-enough" form for them.

framework/console/controllers/FixtureController.php
dynasource: The function argument should be more clear, in this case $fullFixturePath if the @param comment is correctly formulated.

framework/console/controllers/FixtureController.php
dynasource: Agree. It's a nice to have as you already covered it.

tests/data/console/controllers/fixtures/subdir/SecondFixture.php
dynasource: if thats the case, it would be nice to have this string formulated empty or with a more understandable description `'dummy string'`

framework/console/controllers/FixtureController.php
```diff
@@ -415,13 +415,36 @@ private function findFixtures(array $fixtures = [])
...
         }
 
         return $foundFixtures;
     }
 
     /**
+     * Calculates $fixture's name relatively to $templatePath.
+     * Basically, strips getFixturePath() and 'Fixture.php' prefix from fixture's full path
+     * @see getFixturePath()
+     * @param $fixture string Full fixture path
```
rob006: This phpdoc is invalid - type should go before param name:
```
     * @param string $fixture Full fixture path
```

rob006 (changes requested):

tests/data/console/controllers/fixtures/subdir/SecondFixture.php
d1rtyf1ng3rs: > more understandable description 'dummy string'

Agree. But still, I can't see any way this should've been guessed from all the other fixture contents.
I mean, check out any of the `tests/data/console/controllers/fixtures/*Fixture.php` - they all have their data in format of `"some data set for <fixture-name-here> fixture"`...
Hence the question: fixture content should be replaced to `dummy string` only in *added* fixtures, or in *all* of the mentioned fixtures?

framework/console/controllers/FixtureController.php
d1rtyf1ng3rs: Should be fixed now, thanks.

framework/console/controllers/FixtureController.php
d1rtyf1ng3rs: Agree. Should be done now.

dynasource (approved):

framework/console/controllers/FixtureController.php
```diff
@@ -415,13 +415,36 @@ private function findFixtures(array $fixtures = [])
         $foundFixtures = [];
 
         foreach ($files as $fixture) {
-            $foundFixtures[] = basename($fixture, 'Fixture.php');
+            $relativeName = $this->getFixtureRelativeName($fixture);
+            $foundFixtures[] = $relativeName;
```
cebe: no need to have a variable here, just do `$foundFixtures[] = $this->getFixtureRelativeName($fixture);`

framework/console/controllers/FixtureController.php
```diff
@@ -415,13 +415,36 @@ private function findFixtures(array $fixtures = [])
...
+     * @see getFixturePath()
+     * @param string $fullFixturePath Full fixture path
+     * @return string Relative fixture name
+     */
+    private function getFixtureRelativeName($fullFixturePath)
+    {
+        // $fixturesPath is normalized to unix format in getFixturesPath()
+        $fixturesPath = $this->getFixturePath();
+        // normalize $fixture to unix format
+        $fullFixturePath = str_replace("\\", "/", $fullFixturePath);
```
cebe: how does this deal with Windows paths like `C:\Somedir`, looks like that does not work to me.

cebe (changes requested):

framework/console/controllers/FixtureController.php
samdark: Windows doesn't really care about which type of slash is used.

cebe: sure, but if the alias resolves to `C:\whatever/fixturepath` it will not match `C:/whatever/fixturepath` as far as I see.

dynasource: sharp. 
We need more unit tests to make ensure this usecase

samdark: @d1rtyf1ng3rs would you please cover these cases?

SilverFire:
Merged f111ab441312272250fcd5a4372875715bdd7c66 with minor adjustments in f7c9e6b8528b22d086ea1fd95e0b99f8a3d24e35

Thank you!

PR #13558, opened 2017-02-11
Update structure-assets.md [skip ci]

faenir:
`It depends.` -> `Это зависит.` - is incorrect translation into Russian language.
It is necessary to translate correctly, or simply delete it.


samdark:
Thanks!

PR #13560, opened 2017-02-12
Refactor fragment cache widget

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none

Five levels of indentation vs one. One `return` statement vs five.
It's highly arguable but I still believe that one level of indentation is more readable.

[Code Complete](http://www.cc2e.com/) - 17.1 return

> Use a return when it enhances readability. In certain routines, once you know the answer, you want to return it to the calling routine immediately. If the routine is defined in such a way that it doesn't require any cleanup, not returning immediately means that you have to write more code.

### DIFF
--- a/framework/widgets/FragmentCache.php
+++ b/framework/widgets/FragmentCache.php
@@ -134,25 +134,33 @@ class FragmentCache extends Widget
      */
     public function getCachedContent()
     {
-        if ($this->_content === null) {
-            $this->_content = false;
-            if ($this->cache instanceof Cache) {
-                $key = $this->calculateKey();
-                $data = $this->cache->get($key);
-                if (is_array($data) && count($data) === 2) {
-                    list ($content, $placeholders) = $data;
-                    if (is_array($placeholders) && count($placeholders) > 0) {
-                        if (empty($this->getView()->cacheStack)) {
-                            // outermost cache: replace placeholder with dynamic content
-                            $content = $this->updateDynamicContent($content, $placeholders);
-                        }
-                        foreach ($placeholders as $name => $statements) {
-                            $this->getView()->addDynamicPlaceholder($name, $statements);
-                        }
-                    }
-                    $this->_content = $content;
-                }
-            }
+        if ($this->_content !== null) {
+            return $this->_content;
+        }
+
+        $this->_content = false;
+
+        if (!($this->cache instanceof Cache)) {
+            return $this->_content;
+        }
+
+        $key = $this->calculateKey();
+        $data = $this->cache->get($key);
+        if (!is_array($data) || count($data) !== 2) {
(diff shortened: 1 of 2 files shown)


samdark (approved):

SilverFire (approved):

samdark:
The part about placeholders and dynamic content isn't covered by unit tests. Would be great to add these.

dynasource:
the current Unit tests indeed seem not be complete enough to safely accept this PR

Kolyunya:
Thanks for the review. I'm gonna try to find time to add a proper coverage.

Kolyunya:
I added test coverage for dynamic content but I'm not sure any more if we should refactor anything or just leave it as it. If it works don't touch it. We can merge just tests though.
Please review carefully -- all that OB stuff is tricky.

SilverFire (approved):

dynasource:
personally, I do not think its more readable by having multiple return statements. This would only make sense if the code was more self describing ('what' instead of 'how' by the use of private methods). The new version is more time consuming to digest as logic + all the different return statements have to be 'searched' while previously this was instantly obvious.

Kolyunya:
I agree, it's highly opinionated. The problem with the original version, though, is that you have to skip a lot of lines while following the execution logic when you should skip the condition which evaluated to `false`.

dynasource:
> However the problem with the original version is that you have to skip a lot of lines while following the execution logic

I agree. You definately spotted room improvement

It would be nice to have the best of both worlds. To keep the identation but without the side-effects you describe. Another thing. With PHP7 return type hinting, you have to manage/monitor more return statements. This can also be seen as a side effect. Therefore I strive to have 1 return statement with a clear identation structure which is instantly understable and easily manageable.

Kolyunya:
Is this one the rare cases when `goto` usage seems legitimate? Each of the 5 `return` statements in refactored version is `return $this->_content;`. This will allow to have a single `return` statement with a single indentation level.

dynasource:
I've never been fond of `goto` but don't have a problem with identations in the first place, because it instantly indicates the size of logic & dept of a code fragment.

dynasource:
another disadvantage of having multiple 'return' statements is that it is harder to debug inline

SilverFire:
> another disadvantage of having multiple 'return' statements is that it is harder to debug inline

Personally for me multiple return statements are easier to debug.

samdark:
Easier to debug for me as well...

samdark:
Merged. Thank you!

Kolyunya:
I'm glad I could help a bit!

PR #13561, opened 2017-02-12
Update BaseMigrateController.php

semyoren:
Trying to find a real path of namespace in composer class map

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -634,6 +634,21 @@ abstract class BaseMigrateController extends Controller
      */
     private function getNamespacePath($namespace)
     {
+        static $composerClassMap;
+
+        // Filling composer class name array
+        if (is_null($composerClassMap)) {
+            $file = Yii::getAlias('@vendor/composer/autoload_classmap.php');
+            $composerClassMap = file_exists($file) ? require $file : [];
+        }
+
+        // Try to find a path for namespace
+        foreach ($composerClassMap as $className => $path) {
+            if (preg_match('/^' . preg_quote($namespace) . '/s', $className)) {
+                return dirname($path);
+            }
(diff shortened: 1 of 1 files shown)


SilverFire:
@semako What are you trying to fix with this PR?

semako:
Without this change, there is no possibility to start of migration, which are arranged in namespaces that do not correspond to the ultimate directory path classes migration.

For example, you have a composer extension, which placed in path:
/home/www/app/vendor/my/ext/src/migrations
with namespace \my\ext\migrations
but without this fix, and in case of using Composer, namespace of migrations should be:
\vendor\my\ext\src\migrations

In my opinion it's incorrect and mislead.

dynasource:
are you able to proof your usecase with a unit test?

klimov-paul:
> Without this change, there is no possibility to start of migration, which are arranged in namespaces that do not correspond to the ultimate directory path classes migration.

There is an explicit warning about this in the docs:
https://github.com/yiisoft/yii2/blob/master/framework/console/controllers/BaseMigrateController.php#L48-L50

Namespace can be used as migration namespace only if it corresponds the path alias. If such path alias is not set by default (e.g. using `yii2-extension` composer package), you should add it manually using `Yii::setAlias()`.

I do not agree for the change proposed here as composer is not the only way to install and use Yii2. Binding migration code to it (especially to its class autoloading internals) is inconsistent.

klimov-paul:
I am closing this, since use case should be fixed using `Yii::setAlias()` like following:

```php
Yii::setAlias('@my\ext\migrations', '/home/www/app/vendor/my/ext/src/migrations');
```

Issue #9592, opened 2015-09-03
Database type and version properties in \yii\db\Conection?


tom--:
In [yii2-dynamic-ar](http://tom--.github.io/yii2-dynamic-ar/) we haven't found a clean way to discover a connection's database type and version. And I think anyone building an extension (or other reusable code) that depends on advanced DBMS features from more recent versions might face a similar problem.

For example, yii2-dynamic-ar can use Dynamic Columns if the DBMS is Maria 10.0+. It can use JSON if the DB is MySQL 5.7+. And it can use jsonb if it has PostgreSQL 9.4+. But it can't even see the difference between Maria and MySQL and it can't check versions at all.

Reading the version string from the database, parsing it and caching the result is possible but that's as appealing as branching on a parsed user agent string. (I don't like the idea, is what I'm saying.) I would prefer an application configuration that the developer is responsible for. But I don't know where it should go. \yii\db\Conection?

Any thoughts on this would be welcome.

samdark:
Connection seems to be natural choice.

klimov-paul:
Note, that there is already `\yii\db\Connection::driverName`.

PR #13570, opened 2017-02-14
Add DBMS name and DBMS version properties to the DB connection

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #9592

https://github.com/yiisoft/yii2/issues/9592#issuecomment-243462861
> Note, that there is already \yii\db\Connection::driverName.

The driver name is not the DBMS name. E.g. `PDO_DBLIB` driver [supports](http://php.net/manual/en/pdo.drivers.php) the following DBMSs: FreeTDS, Microsoft SQL Server, Sybase.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -7,6 +7,7 @@ Yii Framework 2 Change Log
 - Enh #13523: Plural rule for pasta (developeruz)
 - Bug #13538: Fixed `yii\db\BaseActiveRecord::deleteAll()` changes method signature declared by `yii\db\ActiveRecordInterface::deleteAll()` (klimov-paul)
 - Enh #13278: `yii\caching\DbQueryDependency` created allowing specification of the cache dependency via `yii\db\QueryInterface` (klimov-paul)
+- Enh #13570: Added `$dbmsName` and `$dbmsVersion` properties to the `yii\db\Connection` (Kolyunya)
 
 
 2.0.11.2 February 08, 2017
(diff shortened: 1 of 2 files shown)


Kolyunya:
Need to set enhancement number in changelog if approved.

samdark:
How these properties are filled in?

Kolyunya:
@samdark they are filled in manually.

> Reading the version string from the database, parsing it and caching the result is possible but that's as appealing as branching on a parsed user agent string. (I don't like the idea, is what I'm saying.) I would prefer an application configuration that the developer is responsible for.

samdark:
Why would developer configure it?

samdark:
How to keep this version current when DB version is updated on the server?

Kolyunya:
> Why would developer configure it?
To check against particular DBMS names to use DBMS-specific features.

> How to keep this version current when DB version is updated on the server?
Manually update corresponding application configuration parameter.

Kolyunya:
@samdark am I missing something or I this issue really has a `ready-for-adoption` tag on it?

samdark:
Yes. It does.

samdark (approved):

klimov-paul:
Such change makes absolutely no sense to me.
Since DBMS name and version is not determined automatically - this PR improves nothing.
Developer may use custom constants or `Application::$params` to set and use same values.
What is the sense of creating public fields at `Connection`, which are not populated or processed by framework anywhere?

samdark:
Well, yes. Actually you're right. If we're talking about original issue, the solution would be feature flags.

samdark:
@Kolyunya thanks for looking into the issue.

Issue #13571, opened 2017-02-14
MSSQL checkIntegrity not consistent with other QueryBuilder implementations


boboldehampsink:
### What steps will reproduce the problem?

Load fixtures with MSSQL Database.

### What is the expected result?

InitDb gets fired without table name, firing checkIntegrity = false on all table names.

### What do you get instead?

No specific table name is given, so we get the error "Table does not exist: []"

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.0.13
| Operating system | Ubuntu 16.04

PR #13572, opened 2017-02-14
Fix `yii\db\mssql\QueryBuilder::checkIntegrity` for all tables

boboldehampsink:
This fixes using `yii\db\mssql\QueryBuilder::checkIntegrity` for all tables. This is copied from the PostgreSQL QueryBuilder and modified to work with MSSQL.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #13571


samdark (approved):

boboldehampsink:
@samdark I've updated this to exclude views, also based on the pgsql implementation.

dynasource:
because of the fact that you've copied logic from another class, it would be logical to create an interface to keep consistency.

boboldehampsink:
Okay. I don't have any ideas for a name tho. `ViewInterface`?

dynasource:
something closer to its objective would be `ViewFinderInterface`

boboldehampsink:
Only `getViewNames()` is public, should I only implement that in the interface?

dynasource:
right, forget the interface. A trait is more appropiate to cover both methods, perhabs with even code reuse.

boboldehampsink:
Only `getViewNames()` and `$_viewNames` can be reused - put those in a trait then? Then add `findViewNames()` as abstract method?

dynasource:
IMO that would be a good idea indeed.

boboldehampsink:
How about this

framework/db/mssql/QueryBuilder.php
```diff
@@ -168,23 +168,26 @@ public function alterColumn($table, $column, $type)
...
         $table = $this->db->quoteTableName($table);
-        if ($this->db->getTableSchema($table) === null) {
-            throw new InvalidParamException("Table not found: $table");
+        $tableNames = $this->db->getTableSchema($table) ? [$table] : $this->db->getSchema()->getTableNames($schema);
+        $viewNames = $this->db->getSchema()->getViewNames($schema);
+        $tableNames = array_diff($tableNames, $viewNames);
+        $command = '';
+
+        foreach ($tableNames as $tableName) {
+            $tableName = '"' . $schema . '"."' . $tableName . '"';
```
sergeymakinen: I would quote `$schema` and `$tableName` with `$this->db->quoteColumnName()` - while `$table` could be left as is because we usually only need escaping (a value) when we then inject the value into SQL.

boboldehampsink: I don't understand, why use quoteColumnName for schema and table names? Also if this needs fixing, the current PGSQL implementation also does, as its copied from there.

sergeymakinen: `quoteTableName()` of course, my fault. :)

boboldehampsink: Okay. So `$table = $this->db->quoteTableName($table);` on line 179 can be removed then, because we never use `$table` in SQL directly?

sergeymakinen: Yes, exactly.

boboldehampsink: I've added a commit fixing this, both for mssql and pgsql

sergeymakinen: Thank you! (I can’t say if it’s better to separate ms/pg fixes, let Yii devs decide)

dynasource (approved):

boboldehampsink:
Is this OK to merge?

dynasource:
it needs a third approval

framework/db/ViewFinderTrait.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+namespace yii\db;
+
+/**
+ * ViewFinderTrait implements the method getViewNames for finding views in a database.
+ *
+ * @author Qiang Xue <qiang.xue@gmail.com>
+ * @author Bob Olde Hampsink <b.oldehampsink@nerds.company>
+ * @since 2.0
```
SilverFire: 2.0.12

SilverFire (approved):

samdark:
Merged. Thanks!

Issue #10834, opened 2016-02-13
Native support for srcset in yii\helpers\Html::img()


soulhunter1987:
It would be great to implement support for automatic srcset rendering, it would reduce amount of code on modern adaptive websites. My suggestion:

``` php
    public static function img($src, $options = [])
    {
        if (is_array($src) && ArrayHelper::isAssociative($src) && isset($src['1x'])) {
            $options['src'] = Url::to($src['1x']);
            $srcset = [];
            foreach ($src as $ratio => $url) {
                $srcset[] = Url::to($url) . ' ' . $ratio;
            }
            $options['srcset'] = implode(', ', $srcset);
        } else {
            $options['src'] = Url::to($src);
        }

        if (!isset($options['alt'])) {
            $options['alt'] = '';
        }
        return static::tag('img', '', $options);
    }
```

Usage example:

``` php
$sources = [
    '1x' => '/some/regular/image.jpg',
    '2x' => '/some/image/for/retina.jpg',
];
echo Html::img($sources);
```

Of couse it is the basic usage of this technique, but it is also the most commonly used. Other cases with different ratio units/extra "sizes" attributes may be configured via $options param.

dynasource:
@soulhunter1987 , are you willing to create a PR for your request?

PR #13576, opened 2017-02-14
Add support of `srcset` to `yii\helpers\Html::img()`

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #10834 

Not sure if we should distinguish between `src` and `srcset` passed as the first argument using `yii\helpers\ArrayHelper::isAssociative()` or just pass `srcset` in `options` while passing `null` to `$src`.

If we choose the former it will hit the performance of each and every call to `Html::img()` in all existing and future applications. We also won't be able to support non-assiciative `srcsets` like following. Missing descriptor defaults to `1x` according to the [docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img). Not sure if we should support this though.
```php
$srcset = [
    'base-url',
    '2x' => 'other-url',
];
```

If we choose the latter the method invocation will look somewhat _weird_ but I think we can live with it:
```php
<?= Html::img(null, ['srcset' => [
    '1x' => 'foo',
    '2x' => 'bar',
]]); ?>
```

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -435,6 +435,15 @@ class BaseHtml
     public static function img($src, $options = [])
     {
         $options['src'] = Url::to($src);
+
+        if (isset($options['srcset']) && is_array($options['srcset'])) {
+            $srcset = [];
+            foreach ($options['srcset'] as $descriptor => $url) {
+                $srcset[] = Url::to($url) . ' ' . $descriptor;
+            }
+            $options['srcset'] = implode(',', $srcset);
+        }
+
         if (!isset($options['alt'])) {
             $options['alt'] = '';
         }
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -430,6 +430,8 @@ class BaseHtml
      * the attributes of the resulting tag. The values will be HTML-encoded using [[encode()]].
      * If a value is null, the corresponding attribute will not be rendered.
      * See [[renderTagAttributes()]] for details on how attributes are being rendered.
+     * @since 2.0.12 It is possible to pass the "srcset" option as an array which keys are
+     * descriptors and values are URLs. All URLs will be processed by [[Url::to()]].
      * @return string the generated image tag
      */
     public static function img($src, $options = [])
--- a/tests/framework/helpers/HtmlTest.php
+++ b/tests/framework/helpers/HtmlTest.php
@@ -132,35 +132,98 @@ class HtmlTest extends TestCase
         $this->assertEquals('<a href="mailto:test&gt;">test<></a>', Html::mailto('test<>', 'test>'));
     }
 
-    public function testImg()
+    /**
+     * @return array
+     */
+    public function imgDataProvider()
     {
-        $this->assertEquals('<img src="/example" alt="">', Html::img('/example'));
-        $this->assertEquals('<img src="/test" alt="">', Html::img(''));
-        $this->assertEquals('<img src="/example" width="10" alt="something">', Html::img('/example', ['alt' => 'something', 'width' => 10]));
-        $this->assertEquals('<img src="/base-url" alt="" srcset="/example-100w 100w,/example-500w 500w,/example-1500w 1500w">', Html::img('/base-url', [
-            'srcset' => [
-                '100w' => '/example-100w',
-                '500w' => '/example-500w',
-                '1500w' => '/example-1500w',
+        return [
+            [
+                '<img src="/example" alt="">',
+                '/example',
+                [],
             ],
-        ]));
-        $this->assertEquals('<img src="/base-url" alt="" srcset="/example-1x 1x,/example-2x 2x,/example-3x 3x">', Html::img('/base-url', [
-            'srcset' => [
-                '1x' => '/example-1x',
-                '2x' => '/example-2x',
-                '3x' => '/example-3x',
+            [
+                '<img src="/test" alt="">',
+                '',
+                [],
             ],
-        ]));
-        $this->assertEquals('<img src="/base-url" alt="" srcset="/example-1.42x 1.42x,/example-2.0x 2.0x,/example-3.999x 3.999x">', Html::img('/base-url', [
-            'srcset' => [
-                '1.42x' => '/example-1.42x',
-                '2.0x' => '/example-2.0x',
-                '3.999x' => '/example-3.999x',
+            [
+                '<img src="/example" width="10" alt="something">',
+                '/example',
+                [
+                    'alt' => 'something',
+                    'width' => 10,
+                ],
             ],
-        ]));
-        $this->assertEquals('<img src="/base-url" alt="" srcset="/example-1x 1x,/example-2x 2x,/example-3x 3x">', Html::img('/base-url', [
-            'srcset' => '/example-1x 1x,/example-2x 2x,/example-3x 3x',
-        ]));
+            [
+                '<img src="/base-url" alt="" srcset="">',
+                '/base-url',
+                [
+                    'srcset' => [
+                    ],
+                ],
+            ],
+            [
+                '<img src="/base-url" alt="" srcset="/example-9001w 9001w">',
+                '/base-url',
+                [
+                    'srcset' => [
+                        '9001w' => '/example-9001w',
+                    ],
+                ],
+            ],
+            [
+                '<img src="/base-url" alt="" srcset="/example-100w 100w,/example-500w 500w,/example-1500w 1500w">',
+                '/base-url',
+                [
+                    'srcset' => [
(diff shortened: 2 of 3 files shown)


samdark:
There's another implementation option — a separate method.

Kolyunya:
@samdark speaking of semantics this is still and `img` tag but with other attributes. I'd personally stick to the option of passing `srcset` in `options`.

If we decide to implement a separate tag what name should it have? Maybe `imgSet`? Or `imgSrcset`? Or `srcsetImg`?

samdark:
Either one.

## 1\.  `imgSrcset` (separate method)

- Cleaner.
- Less complicated.
- Method name doesn't precisely match tag name.


## 2\. `srcset` in options

- Still perfect tag name-method name match.
- Weird syntax.
- Complicated logic.

rob006:
@samdark What is weird in this syntax:

```php
<?= Html::img('img.png', ['srcset' => [
    '1x' => 'img-1x.png',
    '2x' => 'img-2x.png',
]]); ?>
```
?

Kolyunya:
@samdark it seems like this method must support both `src` and `srcset` attributes at the same time. Updated the PR accordingly.

rob006:
There is also `sizes`attribute for `img` tag, but not sure if it needs special syntax.

framework/helpers/BaseHtml.php
```diff
@@ -434,7 +434,18 @@ public static function mailto($text, $email = null, $options = [])
      */
     public static function img($src, $options = [])
     {
-        $options['src'] = Url::to($src);
+        if (isset($src)) {
+            $options['src'] = Url::to($src);
+        }
```
rob006: AFAIK `src` attribute is required for `img` tag. Why it is optional here?

Kolyunya:
@rob006 seems like there is not much we can do for `sizes`. Example from [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Example_4_Using_the_srcset_and_sizes_attributes).
```
<img src="clock-demo-thumb-200.png" 
      alt="Clock" 
      srcset="clock-demo-thumb-200.png 200w,
          clock-demo-thumb-400.png 400w"
      sizes="(min-width: 600px) 200px, 50vw">
```

framework/helpers/BaseHtml.php
Kolyunya: @rob006 that is [true](https://www.w3.org/TR/html5/embedded-content-0.html#the-img-element).

> The src attribute must be present, and must contain a valid non-empty URL potentially surrounded by spaces referencing a non-interactive, optionally animated, image resource that is neither paged nor scripted.

rob006:
@Kolyunya Using array for `sizes` list could improve readability - when each option is in its own line it is harder to miss something.

framework/helpers/BaseHtml.php
Kolyunya: @rob006 fixed.

Kolyunya:
Added support for passing `srcset` as a `string` for BC.

samdark:
@rob006 it is fine.

samdark (approved):

tests/framework/helpers/HtmlTest.php
```diff
@@ -137,6 +137,30 @@ public function testImg()
         $this->assertEquals('<img src="/example" alt="">', Html::img('/example'));
         $this->assertEquals('<img src="/test" alt="">', Html::img(''));
         $this->assertEquals('<img src="/example" width="10" alt="something">', Html::img('/example', ['alt' => 'something', 'width' => 10]));
+        $this->assertEquals('<img src="/base-url" alt="" srcset="/example-100w 100w,/example-500w 500w,/example-1500w 1500w">', Html::img('/base-url', [
+            'srcset' => [
+                '100w' => '/example-100w',
+                '500w' => '/example-500w',
+                '1500w' => '/example-1500w',
```
dynasource: are three config options really required for this test?

tests/framework/helpers/HtmlTest.php
```diff
@@ -137,6 +137,30 @@ public function testImg()
...
+            ],
+        ]));
+        $this->assertEquals('<img src="/base-url" alt="" srcset="/example-1x 1x,/example-2x 2x,/example-3x 3x">', Html::img('/base-url', [
+            'srcset' => [
+                '1x' => '/example-1x',
+                '2x' => '/example-2x',
+                '3x' => '/example-3x',
```
dynasource: same

tests/framework/helpers/HtmlTest.php
```diff
@@ -137,6 +137,30 @@ public function testImg()
...
+        $this->assertEquals('<img src="/base-url" alt="" srcset="/example-1.42x 1.42x,/example-2.0x 2.0x,/example-3.999x 3.999x">', Html::img('/base-url', [
+            'srcset' => [
+                '1.42x' => '/example-1.42x',
+                '2.0x' => '/example-2.0x',
+                '3.999x' => '/example-3.999x',
+            ],
+        ]));
+        $this->assertEquals('<img src="/base-url" alt="" srcset="/example-1x 1x,/example-2x 2x,/example-3x 3x">', Html::img('/base-url', [
+            'srcset' => '/example-1x 1x,/example-2x 2x,/example-3x 3x',
+        ]));
```
dynasource: the expansion of this unit test shows the opportunity for a PhpUnit dataprovider to provide:
- code reuse
- clear overview on input data
- easier expansion

dynasource (changes requested):

tests/framework/helpers/HtmlTest.php
Kolyunya: Yes, why not? I think we should have at least the tests with 0, 1 and 3 options.

dynasource: aren't 2 options enough to make your point in the assertion?

Kolyunya:
@dynasource I refactored the tests to use data provider as you advised.

dynasource:
looks good. Thanks for adding the dataProvder. I am still trying to grasp why so many 'srcset' options are required for the assertions.

Kolyunya:
@dynasource zero and one are corner-cases; three and five are regular cases with `w`, `int` and `float` descriptors.

dynasource:
I see. My goal is to have these tests as short as possible (for maintainability purposes) without impacting the coverage of potential examples in practice.

dynasource (approved):

Kolyunya:
```php
     * @since 2.0.12 It is possible to pass the "srcset" option as an array which keys are
     * descriptors and values are URLs. All URLs will be processed by [[Url::to()]].
````
Added a note the method documentation.

**Comment by Kolyunya** [comment]
Can we assign this PR to `2.0.12`? Is there anything else that should be done here?

**Comment by samdark** [comment]
It's fine. Merged. Thank you!

File: framework/helpers/BaseHtml.php [hunk 5]
```diff
@@ -430,11 +430,22 @@ public static function mailto($text, $email = null, $options = [])
      * the attributes of the resulting tag. The values will be HTML-encoded using [[encode()]].
      * If a value is null, the corresponding attribute will not be rendered.
      * See [[renderTagAttributes()]] for details on how attributes are being rendered.
+     * @since 2.0.12 It is possible to pass the "srcset" option as an array which keys are
+     * descriptors and values are URLs. All URLs will be processed by [[Url::to()]].
```
cebe: this should not be an `@since` tag as that would apply to the method itself, which is there already longer. fixed by e12dd89

Kolyunya: @cebe wouldn't it be better to add a separate `@since` for when this method was introduced and leave the `@since 2.0.12` for that patch?

Kolyunya: @cebe you can have multiple `@since` tags according to the [docs](https://phpdoc.org/docs/latest/references/phpdoc/tags/since.html) to specify what happened in which patch. Having a single `@since` doesn't mean that this element was introduced in this patch.

cebe: Interesting, but this is not how our apidoc uses it :)
We currently use `@since` only to indicate when a class or method has been introduced.
Seems this is some kind of changelog for the method, which is a nice idea as we could probably use it to generate more detailed changelogs and provide information on which methods have changed in a release. Its not how it works currently though, so if we start doing this we need to adjust it in other places also and apply this approach to the whole framework.

rob006: Created issue for that: https://github.com/yiisoft/yii2-apidoc/issues/140
I think it is worth to have at some point, but it is not a priority right now.

cebe: Thanks!

Issue #13577, opened 2017-02-14
ActiveFixture::resetTable should use TRUNCATE TABLE


boboldehampsink:
### What steps will reproduce the problem?

Run `ActiveFixture::resetTable` on MSSQL

### What is the expected result?

Table contents are deleted and sequence (identity) is reset.

### What do you get instead?

```
Exception 'yii\base\NotSupportedException' with message 'sqlsrv does not support resetting sequence.'

in /var/www/html/vendor/yiisoft/yii2/db/QueryBuilder.php:602
```

### Additional info

If we use `db->createCommand()->truncateTable()` the contents are deleted AND the sequence is reset. 

If I change 
```php
        $this->db->createCommand()->delete($table->fullName)->execute();
        if ($table->sequenceName !== null) {
            $this->db->createCommand()->resetSequence($table->fullName, 1)->execute();
        }
```
to
```php
        $this->db->createCommand()->truncateTable($table->fullName)->execute();
```
ActiveFixture starts working for MSSQL

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.0.13
| Operating system | Ubuntu 16.04

boboldehampsink:
MSSQL (SQL Server) and MySQL reset the sequence by default when using truncate table. PGSql has an option to do so, that is currently not implemented in Yii2. This is in fact inconsistent, and should be added as well.

boboldehampsink:
After some research, I've found that TRUNCATE TABLE does not work with VIEWS, while the current implementation does. To make it all work for MSSQL we should focus on implementing resetSequence() for that driver, as seen here: #13582

PR #13580, opened 2017-02-15
`yii\db\QueryBuilder::truncateTable` should work consistent over all databases

boboldehampsink:
Cubrid, MSSQL, MySQL, and OCI already restart the identity by default, PGSQL doesn't. This PR makes the use of the `yii\db\QueryBuilder::truncateTable` method more consistent.

_Note: RESTART IDENTITY is available from PGSQL 8.4+_

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #13577


samdark:
The requirement of 8.4+ (2009) is OK. Would be good to mention it in docs here: http://www.yiiframework.com/doc-2.0/guide-db-dao.html

samdark (approved):

boboldehampsink:
Updated all docs

SilverFire (approved):

framework/db/pgsql/QueryBuilder.php
```diff
@@ -206,6 +206,16 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
     }
 
     /**
+     * Builds a SQL statement for truncating a DB table.
```
dynasource: can you add a description for the reason for restarting the Identity?

dynasource (changes requested):

framework/db/pgsql/QueryBuilder.php
boboldehampsink: Sure, what about "Restarts identity to be consistent with other databases."?

dynasource: could you add an explanation about this 'consistency' and the exception for PGSQL for future reference?

boboldehampsink: "Explicitly restart identity for PGSQL to be consistent with other databases which all do this by default"

dynasource: perfect

dynasource (approved):

samdark:
Merged. Thanks!

PR #13581, opened 2017-02-15
`yii\test\ActiveFixture::resetTable()` should use `yii\db\QueryBuilder::truncateTable()` instead of manual delete/reset

boboldehampsink:
Because `yii\db\QueryBuilder::resetSequence()` is not implemented for all drivers, using `yii\db\QueryBuilder::truncateTable()` will yield better results over all drivers. For example, it enables the use of ActiveFixture for MSSQL, where it couldn't be used before.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13577


samdark (approved):

boboldehampsink:
Closing this as discussed in #13577

PR #13586, opened 2017-02-15
Enhance `yii\behaviors\AttributeBehavior`

Kolyunya:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues | #11923 |

`AttributeBehavior` can preserve current attribute values now.
Basic tests coverage is included.

### DIFF
--- a/framework/behaviors/AttributeBehavior.php
+++ b/framework/behaviors/AttributeBehavior.php
@@ -85,6 +85,11 @@ class AttributeBehavior extends Behavior
      * @since 2.0.8
      */
     public $skipUpdateOnClean = true;
+    /**
+     * @var bool whether to preserve non-empty attribute values.
+     * @since 2.0.12
+     */
+    public $preserveNonEmptyValues = false;
 
 
     /**
@@ -117,6 +122,9 @@ class AttributeBehavior extends Behavior
             foreach ($attributes as $attribute) {
                 // ignore attribute names which are not string (e.g. when set by TimestampBehavior::updatedAtAttribute)
                 if (is_string($attribute)) {
+                    if ($this->preserveNonEmptyValues && !empty($this->owner->$attribute)) {
+                        continue;
+                    }
                     $this->owner->$attribute = $value;
                 }
             }
--- a/tests/framework/behaviors/AttributeBehaviorTest.php
+++ b/tests/framework/behaviors/AttributeBehaviorTest.php
@@ -1,0 +1,149 @@
+<?php
+
+namespace yiiunit\framework\behaviors;
+
+use Yii;
+use yiiunit\TestCase;
+use yii\db\Connection;
+use yii\db\ActiveRecord;
+use yii\behaviors\AttributeBehavior;
+
+/**
+ * Unit test for [[\yii\behaviors\AttributeBehavior]].
+ * @see AttributeBehavior
+ *
+ * @group behaviors
+ */
+class AttributeBehaviorTest extends TestCase
+{
+    /**
+     * @var Connection test db connection
+     */
+    protected $dbConnection;
+
+    public static function setUpBeforeClass()
+    {
+        if (!extension_loaded('pdo') || !extension_loaded('pdo_sqlite')) {
+            static::markTestSkipped('PDO and SQLite extensions are required.');
+        }
+    }
+
+    public function setUp()
+    {
+        $this->mockApplication([
+            'components' => [
+                'db' => [
+                    'class' => '\yii\db\Connection',
+                    'dsn' => 'sqlite::memory:',
+                ]
+            ]
+        ]);
+
+        $columns = [
+            'id' => 'pk',
+            'name' => 'string',
+            'alias' => 'string',
+        ];
+        Yii::$app->getDb()->createCommand()->createTable('test_attribute', $columns)->execute();
+    }
+
+    public function tearDown()
+    {
+        Yii::$app->getDb()->close();
+        parent::tearDown();
+    }
+
+    // Tests :
+
+    /**
+     * @return array
+     */
+    public function preserveNonEmptyValuesDataProvider()
+    {
+        return [
+            [
+                'John Doe',
+                false,
+                'John Doe',
+                null,
+            ],
+            [
+                'John Doe',
+                false,
+                'John Doe',
+                'Johnny',
+            ],
+            [
+                'John Doe',
+                true,
+                'John Doe',
+                null,
+            ],
+            [
+                'Johnny',
+                true,
+                'John Doe',
+                'Johnny',
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider preserveNonEmptyValuesDataProvider
+     */
+    public function testPreserveNonEmptyValues(
+        $aliasExpected,
+        $preserveNonEmptyValues,
+        $name,
+        $alias
+    )
+    {
+        $model = new ActiveRecordWithAttributeBehavior();
+        $model->attributeBehavior->preserveNonEmptyValues = $preserveNonEmptyValues;
+        $model->name = $name;
+        $model->alias = $alias;
+        $model->validate();
+
+        $this->assertEquals($aliasExpected, $model->alias);
+    }
+}
+
+/**
+ * Test Active Record class with [[AttributeBehavior]] behavior attached.
+ *
+ * @property integer $id
+ * @property string $name
+ * @property string $alias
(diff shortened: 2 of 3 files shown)


framework/behaviors/AttributeBehavior.php
```diff
@@ -118,6 +123,10 @@ public function evaluateAttributes($event)
             foreach ($attributes as $attribute) {
                 // ignore attribute names which are not string (e.g. when set by TimestampBehavior::updatedAtAttribute)
                 if (is_string($attribute)) {
+                    // skip attribute update when it is not empty and the corresponding mode is enabled
+                    if ($this->skipUpdateOnFilled && !empty($this->owner->$attribute)) {
```
dynasource: this statement should be readable without the comment. Can you use a native function as a replacement for !empty?

tests/framework/behaviors/AttributeBehaviorTest.php
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public function tearDown()
+    {
+        Yii::$app->getDb()->close();
+        parent::tearDown();
+    }
+
+    // Tests :
+
```
dynasource: can you wrap the test methods into 1 test method with a dataprovider? This enables code-reuse and its easier to compare the data used for testing.

framework/behaviors/AttributeBehavior.php
dynasource: 1. this statement should be formulated in such way that a comment is not necessary.
2. can you use a native function instead of empty?
3. a check on $event->name is missing, like with $skipUpdateOnClean. This makes it confusing an less readable and error prone.

dynasource (changes requested):
thanks for adding the first tests for this class

tests/framework/behaviors/AttributeBehaviorTest.php
Kolyunya: Done.

framework/behaviors/AttributeBehavior.php
Kolyunya: @dynasource 

> this statement should be formulated in such way that a comment is not necessary.

It's not necessary, removed it.

> can you use a native function instead of empty?

Refactored that sub-condition to the `$this->owner->$attribute !== ''`.

> a check on $event->name is missing, like with $skipUpdateOnClean. This makes it confusing an less readable and error prone.

I can't get it what is the reason for checking the event name while evaluating each attribute? Can you elaborate please?

dynasource: > can you use a native function instead of empty?
> Refactored that sub-condition to the $this->owner->$attribute !== ''.

the logic behind this should be made less complex. The current format is not self explaining enough and takes time to interpretate and has room for improvement.`skipUpdateOnFilled` in combination with a false comparison is less readable than a positive one with a variable name like `forceUpdateOnFilled`. Additionally, `onFilled` is  a bit vague. What is filled in the context? When is it about a change or about being empty? It would expect something more related to your original issue description.

> a check on $event->name is missing, like with $skipUpdateOnClean. This makes it confusing an less readable and error prone.
see https://github.com/Kolyunya/yii2/blob/8cee4ec59b9553b47adad4cee7e3828ee08123cc/framework/behaviors/AttributeBehavior.php#L113. The current code looks to be run in the 'insert' context also which could be confusing.

Kolyunya: @dynasource I used the `skipUpdateOn...` property because it is consistent with the `skipUpdateOnClean`. Do you really insist on naming the property inconsistently even if it serves the similar purpose? Maybe we can use `skipUpdateOnNonEmpty`?

If you insist on naming it differently I can suggest something like `preserveExisting`.

What property name would you suggest?

>  When is it about a change or about being empty?
It's about being empty.

dynasource: Well, I'd prefer clear naming above consistency. However, your suggestion of `skipUpdateOnNonEmpty` is a lot better.

Kolyunya: @dynasource I refactored the code, what would you say about `preserveExistingValues`?

```php
if ($this->preserveExistingValues && !empty($this->owner->$attribute)) {
    continue;
}
$this->owner->$attribute = $value;
```

`empty` is used here since we want to skip the attribute evaluation both when it's an empty string and `null`.

No event name check is needed here since we want to skip non-empty attributes evaluation while insert too.

dynasource: its getting more readable. However, its still confusing with respect to the 2 scenario's 'insert' & 'update' as 'insert' is not having existingValues right?

Kolyunya: >  'insert' is not having existingValues right?
Insert does have `existing` values if user provided them. User provides values (they exist since this time), then the behavior performs the checks and evaluates them if needed.

Kolyunya: @dynasource
 >  'insert' is not having existingValues right?

Insert does have `existing` values if user provided them. User provides values (they exist since this time), then the behavior performs the checks and evaluates them if needed.

Kolyunya: @dynasource do you think that `preserveNonEmptyValues` is better?

dynasource: thx, its getting better

Kolyunya: @dynasource anything else we can improve?

dynasource: lets wait for the reviews from the rest of the team

samdark (approved):

dynasource (approved):

framework/behaviors/AttributeBehavior.php
```diff
@@ -85,6 +85,11 @@ class AttributeBehavior extends Behavior
      * @since 2.0.8
      */
     public $skipUpdateOnClean = true;
+    /**
+     * @var bool whether to preserve non-empty attribute values.
+     * @since 2.0.12
+     */
+    public $preserveNonEmptyValues = false;
```
klimov-paul: Would `skipNotEmpty` be better?

Kolyunya: @klimov-paul wouldn't `skipNonEmpty` be better?

samdark: Yes. `Non` is better.

dynasource: @klimov-paul, @samdark, see https://github.com/yiisoft/yii2/pull/13586#discussion_r101737303 for the motivation for $preserveNonEmptyValues

samdark: Actually yes, `preserveNonEmptyValues` is more clear despite being a bit longer.

SilverFire: From my POV, `preserve` is a bit more complex word than `ommit` or `skip`

dynasource: skip + not is a double negation which is not ideal. Positive formulation is better for readability. What about `keepSetValues`

samdark: Variable may be set to empty value so it's not precise this way.

dynasource: `keepNonEmptyValues`

cebe:
Note that there is `$immutable` property in subclass `SluggableBehavior` which does the same.

cebe (changes requested):
Note that there is `$immutable` property in subclass `SluggableBehavior` which does the same.

samdark:
@cebe Yes. It should be deprecated.

cebe:
how about moving `$immutable` one layer up?

Kolyunya:
@cebe semantics of `$immutable` means that it preserves values regardless of their emptiness. This may be misleading.

samdark:
@cebe I agree with @Kolyunya.

samdark:
Related to #13449

Kolyunya:
Is here anything else I should fix of it was rescheduled to `2.0.13` because we weren't able to come up with an appropriate variable name yet?

samdark:
Merged with minor adjustments https://github.com/yiisoft/yii2/commit/9031a115b9c85d9656e516ca30854d8ec969942d, thank you!

PR #13591, opened 2017-02-16
Update french translation : yii.php

zoolanders49:
add "must be equal to"  translation for french

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  |

### DIFF
--- a/framework/messages/fr/yii.php
+++ b/framework/messages/fr/yii.php
@@ -17,6 +17,7 @@
  * NOTE: this file must be saved in UTF-8 encoding.
  */
 return [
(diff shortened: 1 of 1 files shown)


cebe:
Thank you! There are other strings that are untranslated in French, could you provide translations for these too?

```
    ' and ' => '',
    'The combination {values} of {attributes} has already been taken.' => '',
```

cebe:
Great, thanks!

zoolanders49:
hello,  I add the two others translations.

PR #13592, opened 2017-02-16
Fixes Oracle’s Schema::setTransactionIsolationLevel()

sergeymakinen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | —

`PDO_OCI` doesn’t like a trailing semicolon (many tests are broken ATM) and Oracle only supports `READ COMMITED` and `SERIALIZABLE`.


framework/db/Schema.php
```diff
@@ -438,7 +438,7 @@ public function rollBackSavepoint($name)
      */
     public function setTransactionIsolationLevel($level)
     {
-        $this->db->createCommand("SET TRANSACTION ISOLATION LEVEL $level;")->execute();
+        $this->db->createCommand("SET TRANSACTION ISOLATION LEVEL $level")->execute();
```
cebe: this change should be made in the OCI subclass only.

framework/db/oci/Schema.php
```diff
@@ -523,4 +525,25 @@ public function insert($table, $columns)
...
+     */
+    public function setTransactionIsolationLevel($level)
+    {
+        switch ($level) {
+            case Transaction::SERIALIZABLE:
+            case Transaction::READ_COMMITTED:
+            parent::setTransactionIsolationLevel($level);
+                break;
+            default:
+                throw new NotSupportedException(get_class($this) . ' only supports transaction isolation levels READ COMMITTED and SERIALIZABLE.');
```
cebe: isn't there an error thrown by the dbms already if you specify a different level? There should not be a need to cover this here.

framework/db/oci/Schema.php
```diff
@@ -523,4 +525,25 @@ public function insert($table, $columns)
...
     }
+
+    /**
+     * Sets the isolation level of the current transaction.
+     * @param string $level The transaction isolation level to use for this transaction.
+     * This can be either [[Transaction::READ_COMMITTED]] or [[Transaction::SERIALIZABLE]].
+     * @throws \yii\base\NotSupportedException when unsupported isolation levels are used.
+     * Oracle only supports SERIALIZABLE and READ COMMITTED.
+     * @see http://docs.oracle.com/cd/E11882_01/server.112/e41084/statements_10005.htm#SQLRF01705
+     * @since 2.0.12
```
cebe: `@since` is only needed if the interface changes. this method existed before 2.0.12 in the parent class so it is not new.

cebe (changes requested):

framework/db/Schema.php
sergeymakinen: There's should not be a semicolon. That's the point of PR.

framework/db/oci/Schema.php
sergeymakinen: I think you're right. In SQLite's class it handles a syntax difference but it's not needed here.

dynasource (changes requested):

samdark (approved):

dynasource (approved):

samdark:
Merged. Thanks!

PR #13595, opened 2017-02-16
fix anchor

w3lifer:
http://www.yiiframework.com/doc-2.0/guide-db-migrations.html#creating-migrations


SilverFire:
Fixed in other languages as well: 5e987b0

PR #13596, opened 2017-02-16
Fixtures named dependencies.

ArgosOleg:
**Example**
In fixture class:
```
class FirstFixture extends ActiveFixture
{
    public $depends  = [
        // the old set dependent fixtures, and it remains if you 
        SecondFixture::class,

        // new set
        'second_fixture_alias' => SecondFixture::class
    ];
}
```

In test class:

```
class SomeCest
{
    public function _before(UnitTester $I)
    {
        $I->haveFixtures([
            'first_fixture_alias' => FirstFixture::class,
        ]);
    }

    public function checkFixtures(UnitTester $I)
    {
        // the old grab dependent fixtures behavior, and it remains if you 
        $data = $I->grabFixture(SecondFixture::class);

        // new behavior
        $data = $I->grabFixture('second_fixture_alias');
    }
}
```

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | don't check

### DIFF
--- a/framework/test/FixtureTrait.php
+++ b/framework/test/FixtureTrait.php
@@ -195,9 +195,10 @@ trait FixtureTrait
                 if (!isset($instances[$name])) {
                     $instances[$name] = false;
                     $stack[] = $fixture = Yii::createObject($fixture);
-                    foreach ($fixture->depends as $dep) {
+                    foreach ($fixture->depends as $alias => $dep) {
                         // need to use the configuration provided in test case
                         $stack[] = isset($config[$dep]) ? $config[$dep] : ['class' => $dep];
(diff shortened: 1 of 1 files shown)


samdark:
What's the goal of having alias for it?

ArgosOleg:
flexibility for using fixture

samdark:
I don't think having aliases makes sense in this case. Thanks for contribution but it won't be merged.

dynasource:
@ArgosOleg: try to keep the architecture around tests as small as possible. The more complex you make it, the harder the maintenance & debugging will be.

Ideally you should not even use fixtures and setup test data right before the actual test. This ensures isolation, more stability and easier maintenance.

PR #13597, opened 2017-02-17
Created unit tests for role based access filter

mrbig:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no - extra unit test coverage
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none

I've noticed that there are no unit test for AccessControl filter that checks RBAC roles. I'm working on an access control enhancement, so to check the current behavior was a must.

This patch introduces a mock auth manager for the AccessRuleTest, than mimicks the auth schema from the [definitive guide](http://www.yiiframework.com/doc-2.0/guide-security-authorization.html), and also a test for role based authorization.

### DIFF
--- a/tests/framework/filters/AccessRuleTest.php
+++ b/tests/framework/filters/AccessRuleTest.php
@@ -136,7 +136,7 @@ class AccessRuleTest extends \yiiunit\TestCase
      *           the id of the action
      *           should the action allow (true) or disallow (false)
      *           test user id
-     *           excepted match result (true, false, null)
+     *           expected match result (true, false, null)
      */
     public function matchRoleProvider() {
         return [
@@ -156,11 +156,11 @@ class AccessRuleTest extends \yiiunit\TestCase
      *
      * @dataProvider matchRoleProvider
      * @param string the action id
-     * @param boolean wether the rule hould allow access
+     * @param boolean whether the rule should allow access
      * @param string the userid to check
-     * @param boolean the excepted result or null
+     * @param boolean the expected result or null
      */
-    public function testMatchRole($actionid, $allow, $userid, $excepted) {
+    public function testMatchRole($actionid, $allow, $userid, $expected) {
         $action = $this->mockAction();
         $auth = $this->mockAuthManager();
         $request = $this->mockRequest();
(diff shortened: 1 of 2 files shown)


samdark (approved):

tests/framework/filters/AccessRuleTest.php
```diff
@@ -39,14 +40,19 @@ protected function mockRequest($method = 'GET')
     }
 
     /**
+     * @param string optional user id
      * @return User
      */
-    protected function mockUser()
+    protected function mockUser($userid = null)
```
dynasource: I dont see an example with an empty argument in the unit tests?

tests/framework/filters/AccessRuleTest.php
```diff
@@ -88,7 +129,55 @@ public function testMatchAction()
 
     // TODO test match controller
 
-    // TODO test match roles
+    public function testMatchRole()
+    {
```
dynasource: because of the repetition in your tests, please convert these to a PhpUnit dataProvider. This enables us to determine if all usecases are covered

dynasource (changes requested):
thanks for adding tests!

tests/framework/filters/AccessRuleTest.php
mrbig: It is used by the already existing test cases (testMatchAction, testMatchVerb). I didn't want to break those, neither changes tests I'm not working with.
See lines: 121 and 185

tests/framework/filters/AccessRuleTest.php
mrbig: Current unit test for the ACF are built up like this. I doubt I'll have the time now to rewrite the whole thing now. Please take a look at the complete test:
https://github.com/yiisoft/yii2/blob/master/tests/framework/filters/AccessRuleTest.php

SilverFire (approved):

cebe (approved):

dynasource (approved):
dataprovider looks great. Thanks.
Fixed some types. 'excepted' is not the same as 'expected' ;)

samdark:
Thank you!

Issue #13598, opened 2017-02-17
AccessControl filter does not support RBAC rules


mrbig:
### Problem description

[RBAC rules](http://www.yiiframework.com/doc-2.0/guide-security-authorization.html#using-rules) are very useful to solve complex authorization problems. Unfortunately the AccessControl filter does not support them out of the box because custom parameters can not be fed to the roles. So one must use each time the matchCallback with custom implementation, but it results in such non elegant access rules where the real role check is hidden in the callback:

``` 
[
    'allow' => true,
    'roles' => ['@'],
    'matchCallback' => function($rule, $action) {
        return Yii::$app->user->can('updateOwnPost', ['postId' => 2]);
    }
]
```

### Proposed solution

Here I'm proposing a solution to add an optional ruleParamsCallback which should return any parameters necessary for the User::can check. For example:

```
[
    'allow' => true,
    'roles' => ['updateOwnPost'],
    'ruleParamsCallback' => function($rule, $action, $role) {
        return ['postId' => 12];
    }
]
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | any
| Operating system | any

mrbig:
Created pull request #13599 to demonstrate proposal

arogachev:
Several duplicate issues already exist: #7643, #13273, #13552.

I think I explained well why this should not be implemented in https://github.com/yiisoft/yii2/issues/13273#issuecomment-269185026.

dynasource:
I'm positive against improving the architecture for a better cooperation between RBAC & Accessfilter. Both are very related and this relation deserves a delicate solution in the framework. 

This PR https://github.com/yiisoft/yii2/pull/13599 covers one of the usecases, although its still some kind "monkey patching". It must be further developed into a general  sophisticated architecture.

arogachev:
@dynasource How you see the handling of this problem then:

> Usually you need to find the model and throw according **404 Not Found** expection if it was not found first and only then throw **403 Forbidden** expection in case of lack of needed permissions. Where we need to write **404 Not Found** check in this approach? Putting this inside a RBAC rule is definetely incorrect, it's a controller's action job no doubt. With additional separate callback you lose integrity of the action and it's not absolutely clear what it does.

dynasource:
good question. It's a problem most of us will recognize. The current state of the framework does not solve this issue out of the box. 

[...]

Issue #13273, opened 2016-12-26
ACF using RBAC roles - No automatic way to obtain params for `$user->can()`


slinstj:
Hi,

It would be useful to have an **automatic** way to pass post/get/any parameters to an ACF filter when it is checking rbac roles (please, look the comments):

```php
    public function behaviors()
    {
        return [
            'acf' => [
                'class' => AccessControl::className(),
                'rules' => [
                    [
                        'allow' => true,
                        'actions' => ['view'],
                         // current implementation of roles:
                        'roles' => ['manageOwnOrganization'], 
                        // what I propose:
                        'roles' => [
                               'manageOwnOrganization' => ['id'], // or ['get' => 'id'] if I want a specific verb
                        ],
                    ],
```
And in https://github.com/yiisoft/yii2/blob/master/framework/filters/AccessRule.php#L151 those values could be get and sent to `$user->can()` as its second parameter... something like `$user->can('manageOwnOrganization', ['id' => 9])`. Where `9` is the $_REQUEST['id'] or $_GET['id'] if using the  `['get' => 'id']` option. If nothing is found it would send a `null` value and the programmer would manage it.

### Use case
I have a rbac rule that verify if the logged user organization is the same that he is trying to manage. The user organization I can obtain from `$user->identity`, but the organization id comes from get or post data. In so many actions I have to make my own `Yii::$app->user->can('manageOwnOrg', ['id' => $id])` to perform the correct filtering but it turns acf rules virtually unuseful to those cases.

It seems a very common use case for rbac/acf. I must confess I am not sure if we have another *automatic* way to get those parameters, but I couldn't find anyone.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 7.0.1
| Operating system | Mint 18

arogachev:
I'm against of this change.

Permissions checks with parameters is something more complicated, putting them inside a controller is a perfectly fine solution.

Here is an example.

Controller's action:

```php
public function actionView($id)
{
    $model = $this->findModel($id);
    if (!Yii::$app->user->can('entry.view', ['model' => $model])) {
        throw new ForbiddenHttpException("Sorry, you don't have permission to view this entry.");
    }
    // The rest of the code
}

/**
 * @param integer $id
 * @return Entry
 * @throws NotFoundHttpException
 */
protected function findModel($id)
{
    $model = Entry::findOne($id);
    if ($model !== null) {
        return $model;
    } else {
[...]

PR #13599, opened 2017-02-17
13598 access filter rule support

mrbig:
Proposed solution to Issue #13598 

I've added the rulesParamCallabck, wich is checked and called before the $user->can is called. Any parameters returned are passed the the $user->can() call.

I think it would be very useful to know about the current action in the callback, because mostly you would look from incoming parameters that are dependent on the action. But because of this I had to change the signature of the 'matchRole' method. I admit this is not 100% safe when there are dependent implementations, so I can accept if you rather ignore the 'action' parameter to ensure backward compatibility.

Unit tests were added to demonstrate the usage of the new feature. The tests mimic the example from the [definitive guide](http://www.yiiframework.com/doc-2.0/guide-security-authorization.html#using-rules)

Please let me know your toughts on this change.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | somewhat on a protected call
| Tests pass?   | yes
| Fixed issues  | 13598

### DIFF
--- a/framework/filters/AccessRule.php
+++ b/framework/filters/AccessRule.php
@@ -83,6 +83,20 @@ class AccessRule extends Component
      */
     public $matchCallback;
     /**
+     * Optional callback to supply parameters for rbac rules
+     * The signature of the callback should be as follows:
+     *
+     * ```php
+     * function ($rule, $action, $role)
+     * ```
+     *
+     * where `$rule` is this access rule,
+     * `$action` is the current [[Action|action]] object,
+     * `$role` is the [[Rule]] object
+     * The callback should return an array with the parameters
+     */
+    public $ruleParamsCallback;
+    /**
      * @var callable a callback that will be called if this rule determines the access to
      * the current action should be denied. If not set, the behavior will be determined by
      * [[AccessControl]].
@@ -108,7 +122,7 @@ class AccessRule extends Component
     public function allows($action, $user, $request)
     {
         if ($this->matchAction($action)
-            && $this->matchRole($user)
+            && $this->matchRole($user, $action)
             && $this->matchIP($request->getUserIP())
             && $this->matchVerb($request->getMethod())
             && $this->matchController($action->controller)
@@ -140,9 +154,10 @@ class AccessRule extends Component
 
     /**
      * @param User $user the user object
+     * @param Action $action the action
      * @return bool whether the rule applies to the role
      */
-    protected function matchRole($user)
+    protected function matchRole($user, $action)
     {
         if (empty($this->roles)) {
             return true;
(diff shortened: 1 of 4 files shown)


framework/filters/AccessRule.php
```diff
@@ -108,7 +122,7 @@ class AccessRule extends Component
     public function allows($action, $user, $request)
     {
         if ($this->matchAction($action)
-            && $this->matchRole($user)
+            && $this->matchRole($user, $action)
```
dynasource: although it suits your usecase, this is a code smell. It adds confusion as $action already had a match with matchAction(). IMO, any further complexification of matching should be done in a cleaner OO way. See next comment.

framework/filters/AccessRule.php
```diff
@@ -140,9 +154,10 @@ protected function matchController($controller)
 
     /**
      * @param User $user the user object
+     * @param Action $action the action
      * @return bool whether the rule applies to the role
      */
-    protected function matchRole($user)
+    protected function matchRole($user, $action)
```
dynasource: this causes a BC break

tests/framework/filters/AccessRuleTest.php
```diff
@@ -88,7 +145,107 @@ public function testMatchAction()
 
     // TODO test match controller
 
-    // TODO test match roles
+    public function testMatchRole()
```
dynasource: thanks for adding tests. Can you transform the tests into PhpUnits dataprovider to benefit from its advantages?

framework/filters/AccessRule.php
```diff
@@ -156,8 +171,13 @@ protected function matchRole($user)
                 if (!$user->getIsGuest()) {
                     return true;
                 }
-            } elseif ($user->can($role)) {
-                return true;
+            } else {
+                $params = empty($this->ruleParamsCallback) ? [] :
```
dynasource: I'd prefer a complete redesign of the core code than to add custom pieces to serve certain usecases. When done properly in an OO way, we can have code serving many more usecases (without ending up with fragmented code)

dynasource (changes requested):

arogachev:
I think this should not be implemented in the core. See https://github.com/yiisoft/yii2/issues/13273#issuecomment-269185026.

dynasource:
I agree based on comment https://github.com/yiisoft/yii2/pull/13599#discussion_r101744732

framework/filters/AccessRule.php
mrbig: IMHO the point of any role is to allow or disallow an action.

But I can accept your worries about BC break, so if you like that I can remove this.

framework/filters/AccessRule.php
mrbig: My feedback here: $user->can($role) part could be separated from the guest and 'all' checks, because adding rule checks even with a custom filter is pretty difficult

PR #13605, opened 2017-02-17
Added method to extract options from `dsn` property

leandrogehlen:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |


almoramim:
Hi

dynasource:
this is a useful extension. Can you put a real example in your unit tests?

framework/db/Connection.php
```diff
@@ -843,6 +843,27 @@ function ($matches) {
...
+     * @param string $name the option name
+     * @return string the required value
+     */
+    public function getDsnInfo($name)
+    {
+        if (preg_match('/\w+:(.*)/', $this->dsn, $matches)) {
+            $options = explode(';', $matches[1]);
+
+            foreach ($options as $opt) {
+                list($key, $value) = explode('=', $opt);
```
Kolyunya: This is an error for `PHP5`. See [warning](http://php.net/manual/en/function.list.php).

leandrogehlen:
This is useful to me, beacuse i have created a command to generate backup before execute migrations,
then it's necessary to get this information

SilverFire:
I'd rather make it as a separate class, in order not to mix-up Connection class responsibility.

```php
$parser = new DsnParser($connection->dsn);
$parser->getDriverName();
$parser->getParams();
```

Kolyunya:
@SilverFire why not add the `Dsn` class then?
```php
$dsn = new Dsn($connection->dsn);
$dsn->getDriverName();
$dsn->getParams();
```

leandrogehlen:
Really, is better `DnsParser` @SilverFire

SilverFire:
This can be easily implemented outside of the core. I'm closing this PR

dynasource:
Disagree. Needed this in the past for several usecases

framework/db/Connection.php
```diff
@@ -843,6 +843,27 @@ function ($matches) {
     }
 
     /**
+     * Extracts information from [[dsn]]
+     * @param string $name the option name
+     * @return string the required value
+     */
+    public function getDsnInfo($name)
+    {
+        if (preg_match('/\w+:(.*)/', $this->dsn, $matches)) {
```
samdark: First `\w+` is a driver name which could be useful as well.

dynasource:
* migration tools
* db dump tools
* fixtures
* advanced cross db queries

framework/db/Connection.php
samdark: We can split string by 1st `:` without regex.

framework/db/Connection.php
```diff
@@ -843,6 +843,27 @@ function ($matches) {
     }
 
     /**
+     * Extracts information from [[dsn]]
+     * @param string $name the option name
+     * @return string the required value
+     */
+    public function getDsnInfo($name)
```
samdark: If you need multiple options it would parse the string each time.

samdark: What @SilverFire would be much better:

```php
class DSN
{
    private $string;
    private $driver;
    private $params;

    public function __construct($string)
    {
        $this->string = $string;
    }

    protected function parse()
    {
        // ...
        $this->driver = xxx
        $this->params = yyy
    }

    public function getDriverName()
    {
        if ($this->driver === null) {
           $this->parse();
        }
        return $this->driver;
    }

    public function getParameter($name) {}
    public function getParameters() {}
}
```

samdark:
@dynasource

> migration tools

How exactly DSN is useful?

> db dump tools

Could be useful but no such tools are part of the framework.

> fixtures

Fixture should not be aware of connection it's executed on.

> advanced cross db queries

Why do you need info from DSN for these?

Kolyunya:
Please keep in mind that the generic DSN [can be](http://php.net/manual/en/pdo.construct.php) like `uri:file:///path/to/dsnfile` or like `dsn_name`.

dynasource:
@samdark, in these examples I needed the specific name of the database

In general, as you have to define the database name yourself explicitly, should't it be part of the framework to extract it easily again? Not being able to get the name of the current database actually surprised me, as it seems such a basic feature.

Usecases around getting the 'dbname' and the 'host' are as @leandrogehlen states 'backups', but one can also think of generating temporary databases based on the original 'dbname' or for other logging purposes.

This is an example how Laravel4 extracts this information http://stackoverflow.com/questions/17132206/laravel-4-read-config-files

mikehaertl:
I also would find it useful. Use case: In some of my projekcts I have to rely on a foreign DB that has wrong charset set. So in my init migration I have this code:

```php
        if (preg_match('/dbname=(.*)$/', Yii::$app->db->dsn, $matches)) {
            $name = $matches[1];
            $this->execute("ALTER DATABASE $name CHARACTER SET utf8 COLLATE utf8_general_ci;");
            $this->execute("ALTER DATABASE $name DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_general_ci;");
        }
```

Instead of adding yet another class, couldn't we say that this information is kind of related to the DB schema? So we could move the code to the respective `yii/db/<dbms>/Schema` class.

```
$dbname = Yii::$app->db->schema->getDsnInfo('name')
```

samdark:
I'm against adding it to schema.

dynasource:
IMO, the DSN should even be generated with array configuration in the first place.

klimov-paul:
I supose such change is risk prone and should not be taken under Yii scope.
DSN is a comlex string with complex syntax. I can not tell for sure about PDO DSN, but here is an example of trivial MongoDB DSN string:

```
mongodb://user:password@10.10.10.10:27017/yii2test
```

> IMO, the DSN should even be generated with array configuration in the first place.

I doubt it is possible: if it can be so then PDO should provide the support for it. 
It must be some reason why PDO driver uses exactly a string for the connection specification instead of array.

klimov-paul:
> Needed this in the past for several usecases

If it is so, such issue should be reported to `php.net` to `PDO` extension bug tracking.
The only safe wat to parse DSN string at gain its information is using PDO internals.

Implementing our own parser means the countless of future issues complaining the facts it can not parse  particular complex DSN.

klimov-paul:
Also keep in mind that relying on `DSN` string for some connection information is not practical.

For example: I may need to get information about port number used by database connection. Inside the DSN port number is a option parameter and can be omitted. So in some cases I can extract port number from it, but in others - I can not. MySQL and PostreSQL may use different default ports, so I in the end I have to create complex functionality to find out actual port number anyway.

dynasource:
we can go into a lot of depth here, but isn't it logical to have an easy access to something basic as the name of the current database.

klimov-paul:
Name of current database can also be omitted at `DSN`: you may connect to specific server and use multiple databases or use `select database` at later program stage.
Gaining the actual database name used at present time, should be done via SQL statement.

klimov-paul:
Whatever functionality, which should compose database connection information, you create require complex processing and lot of `if` statements to gain complete picture. E.g.

```
$parsedDsn = $connection->parseDsn();

if (isset($parsedDsn['port'])) {
    $port = $parsedDsn['port'];
} else {
    switch ($parsedDsn['driverName']) {
        case 'mysql' :
            $port = 123;
            break;
        case 'pgsql' :
            $port = 456;
            break;
        // ...
    }
}
```

If you wish to compose verbose DB connection info at some place it is better to parse DSN there instead of using Yii layer for that.

leandrogehlen:
i think that the better solution, like @dynasource said is build dsn from properties something like

```php
//config/db.php
return [
    'dsn' => [
        'driver' => 'mysql'
        'host' => 'localhost'',
       'dbname' => 'yii2db'
    ],
    'username' => 'aplicacao',
    'password' => 'soldbapp15',
    'charset' => 'utf8',
];
```

klimov-paul:
> i think that the better solution, like @dynasource said is build dsn from properties something like

How does this solves your original problem? You will be able to get DSN verbose information **only** it is has been specified as array at `Connection` config, but if developer chooses to use old DSN string - you still can not do anything.

klimov-paul:
As I said this is the issue for `PDO` project. E.g. there should be some `\PDO::getConnectionInfo()` method, which you can use.
Such kind of issues are better to be addressed to `php.net`.

leandrogehlen:
But after this change, will be allowed only the new format. the dsn string should be removed

klimov-paul:
> But after this change, will be allowed only the new format. the dsn string should be removed

This means BC break - it is not acceptable.

leandrogehlen:
if this is not acceptable, then i agree with you @klimov-paul, it's necessar open a issue in PDO project

samdark:
@leandrogehlen would you handle opening an issue there? If yes, please post a link here, we'll vote for it since having it in PDO would be great.

leandrogehlen:
@samdark, sorry, i can't do this, because my english is very bad, and i don't know how to do this.

samdark:
Here it is: https://bugs.php.net/bug.php?id=74155&thanks=4

leandrogehlen:
Thanks @samdark

PR #13606, opened 2017-02-17
Static Code Analysis with Php Inspections (EA Extended)

kalessil:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

- fixed a greedy regex
- fixed preg_quote (/ is not escaped by default)
- dropped dead code (instanceof a trait always return false)
- language level migration fixes

### DIFF
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -197,13 +197,7 @@ trait ActiveRelationTrait
             throw new InvalidConfigException('Invalid link: it must be an array of key-value pairs.');
         }
 
-        if ($this->via instanceof self) {
-            // via junction table
-            /* @var $viaQuery ActiveRelationTrait */
-            $viaQuery = $this->via;
-            $viaModels = $viaQuery->findJunctionRows($primaryModels);
-            $this->filterByModels($viaModels);
-        } elseif (is_array($this->via)) {
+        if (is_array($this->via)) {
             // via relation
             /* @var $viaQuery ActiveRelationTrait|ActiveQueryTrait */
             list($viaName, $viaQuery) = $this->via;
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -197,7 +197,13 @@ trait ActiveRelationTrait
             throw new InvalidConfigException('Invalid link: it must be an array of key-value pairs.');
         }
 
-        if (is_array($this->via)) {
+        if ($this->via instanceof self) {
+            // via junction table
+            /* @var $viaQuery ActiveRelationTrait */
+            $viaQuery = $this->via;
+            $viaModels = $viaQuery->findJunctionRows($primaryModels);
+            $this->filterByModels($viaModels);
+        } elseif (is_array($this->via)) {
             // via relation
             /* @var $viaQuery ActiveRelationTrait|ActiveQueryTrait */
             list($viaName, $viaQuery) = $this->via;
--- a/framework/i18n/I18N.php
+++ b/framework/i18n/I18N.php
@@ -109,7 +109,7 @@ class I18N extends Component
             return $message;
         }
 
-        if (preg_match('~{\s*[\d\w]+\s*,~u', $message)) {
+        if (preg_match('~{\s*[\w]+\s*,~u', $message)) {
             $formatter = $this->getMessageFormatter();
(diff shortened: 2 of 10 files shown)


framework/db/ActiveRelationTrait.php
```diff
@@ -197,13 +197,7 @@ public function populateRelation($name, &$primaryModels)
             throw new InvalidConfigException('Invalid link: it must be an array of key-value pairs.');
         }
 
-        if ($this->via instanceof self) {
```
kalessil: Dead code: instanceof <trait> returns false

kalessil:
Builds are broken ATM, will check if it's due to my changes

framework/db/ActiveRelationTrait.php
samdark: Isn't `self` referencing the class trait is attached to in this case?

framework/i18n/I18N.php
```diff
@@ -109,7 +109,7 @@ public function format($message, $params, $language)
             return $message;
         }
 
-        if (preg_match('~{\s*[\d\w]+\s*,~u', $message)) {
+        if (preg_match('~{\s*[\w]+\s*,~u', $message)) {
```
samdark: Doesn't look the same.

framework/requirements/YiiRequirementChecker.php
```diff
@@ -112,7 +112,7 @@ function check($requirements)
      */
     function checkYii()
     {
-        return $this->check(dirname(__FILE__) . DIRECTORY_SEPARATOR . 'requirements.php');
+        return $this->check(__DIR__ . DIRECTORY_SEPARATOR . 'requirements.php');
```
samdark: Should not be modified like that because requirements checker requirements are lower than framework's. It should be able to say that requirements aren't met.

framework/requirements/YiiRequirementChecker.php
```diff
@@ -155,7 +155,7 @@ function render()
         if (!isset($this->result)) {
             $this->usageError('Nothing to render!');
         }
-        $baseViewFilePath = dirname(__FILE__) . DIRECTORY_SEPARATOR . 'views';
+        $baseViewFilePath = __DIR__ . DIRECTORY_SEPARATOR . 'views';
```
samdark: Should not be modified like that because requirements checker requirements are lower than framework's. It should be able to say that requirements aren't met.

framework/requirements/views/web/index.php
```diff
@@ -7,7 +7,7 @@
 <head>
     <meta charset="utf-8"/>
     <title>Yii Application Requirement Checker</title>
-    <?php $this->renderViewFile(dirname(__FILE__) . '/css.php'); ?>
+    <?php $this->renderViewFile(__DIR__ . '/css.php'); ?>
```
samdark: Should not be modified like that because requirements checker requirements are lower than framework's. It should be able to say that requirements aren't met.

samdark (changes requested):

framework/db/ActiveRelationTrait.php
kalessil: [It is](https://3v4l.org/ZRSNS), reverting.

framework/requirements/views/web/index.php
kalessil: Sounds reasonable, reverting

framework/i18n/I18N.php
kalessil: Can be confusing, but it still [the same](https://3v4l.org/vDpse).
`\d\w` is a greedy matching because `\w ==  [A-Za-z0-9_]` shares `0-9` with `\d`.

framework/requirements/YiiRequirementChecker.php
kalessil: Reverted

framework/i18n/I18N.php
samdark: Oh. Right. Good catch.

samdark (approved):

dynasource (approved):

cebe (approved):
good finds, thanks :)

samdark:
Merged. Thanks!

kalessil:
You are welcome and thanks for feedback - I had to fix some false positives, what is really great

PR #13610, opened 2017-02-18
Closes #13243: Added support for unicode attribute names in `yii\widgets\DetailView`

### LINKED ISSUE #13243: DetailView & cyrillic attribute names
put Cyrillic attribute name(s) in attribute filed
I have model built on 1С oData API, so field names are mixture of russian and english languages. =|
### What do you get instead?
 InvalidConfigException('The attribute must be specified in the format of "attribute", "attribute:format" or "attribute:format:label"')

### Additional info
Seems like this regexp needs /u modifier:
  if (!preg_match('/^([\w\.]+)(:(\w*))?(:(.*))?$/', $attribute, $matches))
                  

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 5.6.27
| Operating system | Debian 7

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13243

Closes #13243 🔒 :shipit: 

This is already implemented for GridView.
 
I think solution suggested by @bsyomov in https://github.com/yiisoft/yii2/issues/13243#issuecomment-268099889 is probably better because it does not allow characters like comma, etc. in attribute name. From the other side, who knows, maybe such special characters can take place in some API's related attributes. Creating black list for them can be a hassle. Anyway I changed regex to be the same as in GridView for consistency.

Ideally, if regex is the same for GridView and DetailView it's better to store it in some common constant. However GridView and DetailView are not directly related with each other, so I left them separate for now.

GridView is pretty poorly tested. Ideally this needs to be tested there too.


dynasource:
shouldnt that ObjectMock extends from Model?

arogachev:
@dynasource Probably. That's how it initially was.

dynasource:
Im expecting that for testing the DetailView widget

arogachev:
@dynasource I changed the base class to `yii\base\Model`.

> That way we can make definition of attributes with unicode names less verbose.

But this does not make that definition less verbose. Actually PHP natively supports unicode in variable names, so why not to use it instead? Applied this change too.

dynasource (approved):

samdark (approved):

framework/widgets/DetailView.php
```diff
@@ -205,7 +205,7 @@ protected function normalizeAttributes()
 
         foreach ($this->attributes as $i => $attribute) {
             if (is_string($attribute)) {
-                if (!preg_match('/^([\w\.]+)(:(\w*))?(:(.*))?$/', $attribute, $matches)) {
+                if (!preg_match('/^([^:]+)(:(\w*))?(:(.*))?$/', $attribute, $matches)) {
```
SilverFire: The updated regular expression will allow attribute names, starting from the digit, e.g. `5nizza'. Is not allowed it PHP.

SilverFire (changes requested):

framework/widgets/DetailView.php
arogachev: @SilverFire As I said, I decided to use the same regex that's currently used for GridView. Seems like the same problem was fixed earlier for GridView only and DetailView's regex was not synchronized with these changes. That means earlier this regex was accepted. What you suggest to do? Fix it to not allow leading digits in both DetailView and GridView or just DetailView?

SilverFire: Ok, let it be

SilverFire (approved):

samdark:
Merged. Thanks!

PR #13611, opened 2017-02-18
Fixes #13306: Wildcard in `reloadableScripts` in `yii.js` allows 0 characters

### LINKED ISSUE #13306: Wildcard in reloadableScripts in yii.js allows 0 characters
I found this issue during writing tests for `yii.js`.

Currently in `yii.js` we have `reloadableScripts` property used in `initScriptFilter()` function. It supports wildcards since https://github.com/yiisoft/yii2/commit/ea0c6d7a1a5e1b8c20c9bbfef495892b1e1b2564:

```js
/**
 * List of JS or CSS URLs that can be loaded multiple times via AJAX requests.
 * Each item may be represented as either an absolute URL or a relative one.
 * Each item may contain a wildcart matching character `*`, that means one or more
 * any characters on the position. For example:
 *  - `/css/*.js` will match any file ending with `.js` in the `css` directory of the current web site
 *  - `http*://cdn.example.com/*` will match any files on domain `cdn.example.com`, loaded with HTTP or HTTPS
 *  - `/js/myCustomScript.js?realm=*` will match file `/js/myCustomScript.js` with defined `realm` parameter
 */
reloadableScripts: [],
```

However, this statement in JSDoc:

> Each item may contain a wildcart matching character `*`, that means **one or more** * any characters on the position.

contradicts with reality.

`0` characters is allowed, so when we set:

```js
yii.reloadableScripts = ['/js/*.js'];
```

`/js/.js` will be a valid combination.

Example for easier reproducing:

```js
function escapeRegExp(str) {
    return str.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
}

var rule = '/js/*.js';
var url = '/js/.js';
var match = new RegExp("^" + escapeRegExp(rule).split('\\*').join('.*') + "$").test(url);
```

`match` will have a value `true`. Looks like a bug for me.

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13306 

Fixes #13306 🐛 🔨

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.12 under development
 --------------------------
 
+- Bug #13306: Wildcard in `reloadableScripts` in `yii.js` allows 0 characters
 - Enh #13523: Plural rule for pasta (developeruz)
 - Bug #13538: Fixed `yii\db\BaseActiveRecord::deleteAll()` changes method signature declared by `yii\db\ActiveRecordInterface::deleteAll()` (klimov-paul)
 - Bug: Fixed `yii\db\Connection::useMaster()` - Exception within callback completely disables slaves (Vovan-VE)
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.12 under development
 --------------------------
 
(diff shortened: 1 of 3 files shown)


dynasource (approved):

samdark (approved):

samdark:
Merged. Thanks!

Issue #8748, opened 2015-06-11
proposal: add method authManager::hasAssignment


dizews:
Hello.

I am writing own module to roles control. To prevent deleting a role I want to check if the role has assignments or not. What do you think if we add this functionality to core?

samdark:
Direct assignments you mean?

dizews:
yes, I talk about direct assignments.

samdark:
Yes, I think it's OK.

dizews:
thanks!

Kolyunya:
We already have `yii\rbac\ManagerInterface::getUserIdsByRole()` implemented. Is this sufficient for the issue?

dynasource:
the should be implemented by yii\rbac\Role through a method `hasAssignments'

Kolyunya:
@dynasource so it sould be a `yii\rbac\Role::hasAssignments()` method, right? Should it accept a manager as a parameter? If no how would we get a reference to the manager to get assignment information?

dynasource:
Manager should ideally be a dependency in Role

Kolyunya:
@dynasource it's currently not, isn't it?

Kolyunya:
@samdark @cebe how should we implement this feature?

dynasource:
not yet indeed. When done in a proper OO way, it should be added as a dependency in the constructor at https://github.com/yiisoft/yii2/blob/master/framework/rbac/PhpManager.php#L726

PR #13612, opened 2017-02-18
Add `yii\rbac\ManagerInterface::hasAssignments()`

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #8748


framework/rbac/ManagerInterface.php
```diff
@@ -230,6 +230,13 @@ public function getAssignments($userId);
     public function getUserIdsByRole($roleName);
 
     /**
+     * Returns a boolean indicating whether the role has assignments.
+     * @param string $roleName the role name
+     * @return bool whether the role has assignments.
```
samdark: Need `@since 2.0.12`

samdark (changes requested):

samdark:
- Need a line for changelog.
- The change should be mentioned in UPGRADE since it may be breaking for custom RBAC manager implementations (necessary to implement additional method in this case).

Kolyunya:
@samdark I used to think that we can not break the BC in minor releases/ If we indeed can then why is #11927 assigned to `2.1.x`? I used to think that it's because of the BC break.

samdark:
Right. Assigned to 2.1.0.

Kolyunya:
@samdark I've just checked the `UPGRADE.md` and noticed that it contains many BC breaks like:

> * RBAC: `getChildRoles()` method was added to `\yii\rbac\ManagerInterface`. If you've implemented your own RBAC manager
  you need to implement new method.

That's very confusing... If we already have breaking changes then why can't we introduce new ones?

samdark:
We're getting more strict on versions.

Issue #13086, opened 2016-11-28
Error with rules which include a defaults 


MattDalm:
### What steps will reproduce the problem?
I only have 1 rule which includes a defaults but it is not working as expected. 
Here is the rule

`                [
                    'pattern' => '<language>/<brand>',
                    'route' => 'site/brand',
                    'defaults' => ['language' => 'en'],
                ] 
`

If I access domain.com/brand1 I get an error "Not Found (#404)"

Note that if I setup this rule instead: 

`                [
                    'pattern' => 'test/<language>/<brand>',
                    'route' => 'site/brand',
                    'defaults' => ['language' => 'en'],
                ] 
`

and I access domain.com/test/brand1 then I get the expected result and I am redirected to site/brand with the language parameter defaulted to 'en'

### What is the expected result?

The rule should be applied and the default language set to 'en' . 

### What do you get instead?

yii\web\HttpException:404

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 7.0.12
| Operating system | Windows 10

samdark:
Could be related to https://github.com/yiisoft/yii2/pull/12610

rob006:
See #13624

PR #13614, opened 2017-02-18
Add a failing test to the `yii\web\UrlRule`

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | none

Added a failing test to reproduce and verify the #13086.
Not sure if the test should pass or not by design. Two options are possible:
* If it should not pass we will change the first test to return `false` and it would serve as a documentation.
* If it indeed should we will need to fix the `UrlRule`.

### DIFF
--- a/tests/framework/web/UrlRuleTest.php
+++ b/tests/framework/web/UrlRuleTest.php
@@ -884,6 +884,18 @@ class UrlRuleTest extends TestCase
                 ],
             ],
             [
+                'optional param at the begining',
+                [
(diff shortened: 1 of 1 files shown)


samdark:
@rob006 you could be interested in this one.

rob006:
See #13624

samdark:
Closing since the commit is now part of @rob006 pull request.

PR #13615, opened 2017-02-18
Add `mergeType`, `beforeKeys` and `afterKeys` to `registerJs()` method to solve issues dealing with order of inline JS includes.

derekisbusy:
Add method `jsLines()` and replace implode statements with `jsLines()` method.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | needs testing
| Fixed issues  | Helps fix issues related to the order in which JS scripts get included. 

**`mergeType`** parameter to define how existing inline JS with the same keys should be merged together.

 For example the Pjax widget would require inner pjax containers to be appended so that the events get binded in the correct order. In this case you would use `View::MERGE_PREPEND` since the `registerJs()` method is in the `run()` method the inner widgets are called first and thus the order is reversed. An alternative would be to use `registerJs()` in the `begin()` method but this would break if nesting widgets between the layout and action views since the layout is rendered after the action's view.  Thus the only way to correctly order the widgets inline scripts when nested in both the layout and action is to use `View::MERGE_PREPEND` in the `run()` method of the widget. 

**`afterKeys`** which enables JS code to be deferred and included after another widgets code if needed. For example with the Pjax widget you want to make sure the ActiveForm events are attached first to prevent the Pjax request from happening before Ajax validation.

**`beforeKeys`** that works the same as `afterKeys` except the JS code get's included before.

**Example usage:**

```php
$view->registerJs($js, View::POS_READY, 'pjax', View::MERGE_PREPEND,null,'activeform');
```

Or could use separate methods to make code cleaner:

```php
$view->registerJs($js, View::POS_READY, 'pjax', View::MERGE_PREPEND)
      ->addAfter('activeform');

$view->registerJs($js, View::POS_READY, 'some-widget', View::MERGE_PREPEND)
     ->addBefore(['activeform','other-widget', '..']);
```

Added support for callables and closures to the `js` paramter of `registerJs()` method so that widgets can handle their own lines.

 ```php

class MyWidget extends \yii\base\Widget
{
    /**
     * @inheritdoc
     */
    public function run()
    {
         // ... 
         $this->registerClientScript();
    }

    /**
     * Registers the needed JavaScript.
     */
    public function registerClientScript()
    {
            self::$js .= ".fadeIn().fadeOut()";
            $view = $this->getView();
            // merge skip makes sure the callback is added once
            $view->registerJs([self::className()], View::POS_READY, 'my-custom-widget', View::MERGE_SKIP);
    }
  public static $js = "";
    
    public static function initJs($view)
    {
        // can use the callback to customize the output 
        return "jQuery(document)".implode("\n", self::$js).";\n";
    }
}

```

### DIFF
--- a/framework/web/View.php
+++ b/framework/web/View.php
@@ -659,7 +659,7 @@ class View extends \yii\base\View
         if (empty($this->_jsAfter[$scriptKey])) {
             return false;
         }
-        foreach ($this->_jsAfter[$scriptKey] as $beforeKey => $null) {
+        foreach ($this->_jsAfter[$scriptKey] as $beforeKey => $isAfter) {
             if (array_key_exists($beforeKey, $this->_jsAfter) && array_key_exists($scriptKey, $this->_jsAfter[$beforeKey])) {
                 throw new \Exception("Circular dependency detected with keys: $beforeKey <-> $scriptKey");
             } else if (array_key_exists($beforeKey, $array) && !array_key_exists($beforeKey, $this->_jsHandled)) {
--- a/framework/web/View.php
+++ b/framework/web/View.php
@@ -669,12 +669,11 @@ class View extends \yii\base\View
         return $this->add($this->_jsAfter, $afterKeys, $scriptKey, $pos);
     }
     
-    
-    public function jsLines($pos)
+    private function mergeAfter($pos, &$array)
     {
-        $array = $this->js[$pos];
-        $lines = '';
-        $this->_jsHandled[$pos] = [];
+        if (!isset($this->_jsBefore[$pos])) {
+            return;
+        }
         foreach ($this->_jsBefore[$pos] as $beforeKey => $afterKeys) {
             foreach ($afterKeys as $afterKey) {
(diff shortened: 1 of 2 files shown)


dynasource:
I've look at your PR and it shows a few problems:
* it add too much complexity for the issue being solved
* yii\web\View is even more violating the "single responsibility principle" of SOLID
* `registerJs` is exploded with arguments and violates another good design practice
* there are too many questionable methods added to the public interface making the framework less stable
* ActiveForm is even more coupled to View

dynasource:
I've look at your PR and it shows a few problems:
* it adds too much complexity for the issue being solved
* yii\web\View is even more violating the "single responsibility principle" of SOLID
* `registerJs` is exploded with arguments and violates another good design practice
* there are too many questionable methods added to the public interface making the framework less stable
* ActiveForm is even more coupled to View

samdark:
Agree with @dynasource. Despite solving the problem it adds too much complexity to public API.

PR #13617, opened 2017-02-19
Optimize `yii\data\ActiveDataProvider`

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12752 

Added tests to the #13467.

### DIFF
--- a/tests/framework/data/ActiveDataProviderTest.php
+++ b/tests/framework/data/ActiveDataProviderTest.php
@@ -7,12 +7,14 @@
 
 namespace yiiunit\framework\data;
 
+use yii\base\InvalidCallException;
 use yii\data\ActiveDataProvider;
 use yii\db\Query;
 use yiiunit\data\ar\ActiveRecord;
 use yiiunit\data\ar\Customer;
 use yiiunit\data\ar\Item;
 use yiiunit\framework\db\DatabaseTestCase;
+use yiiunit\framework\db\UnqueryableQueryMock;
 use yiiunit\data\ar\Order;
 
 /**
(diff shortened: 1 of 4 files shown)


tests/framework/data/ActiveDataProviderTest.php
```diff
@@ -177,4 +178,42 @@ public function testPaginationBeforeModels()
...
+            $this->assertCount(0, $provider->getModels());
+        } catch (InvalidCallException $exception) {
+            $this->fail('An excessive models query was executed.');
+        }
+
+        $this->assertEquals(0, $pagination->getPageCount());
+    }
+}
+
+class UnqueryableQueryMock extends Query
```
samdark: Class should be in its own file.

tests/framework/data/ActiveDataProviderTest.php
```diff
@@ -177,4 +178,42 @@ public function testPaginationBeforeModels()
         $provider->refresh();
         $this->assertEquals(2, count($provider->getModels()));
     }
+
+    public function testZeroModels()
```
samdark: I'd name it `testNoQueryWhenZeroModels` or something like that.

samdark (changes requested):

tests/framework/data/ActiveDataProviderTest.php
Kolyunya: What about `testDoesNotPerformQueryWhenHasNoModels`? This seems to be more like a real English sentence.

tests/framework/data/ActiveDataProviderTest.php
Kolyunya: Is `use yiiunit\framework\db\UnqueryableQueryMock`

Kolyunya:
@samdark fixed mentioned issues.

tests/framework/data/ActiveDataProviderTest.php
dynasource: I dont have a problem with keeping it in the current file as it is unlikely to be reused in other tests. Additionally it should be replaced by an anonymous class in PHP7 in the (hopefully near) future

tests/framework/data/ActiveDataProviderTest.php
dynasource: lets keep it close to the code. I would prefer something like: `testSkipQueryWhenZeroTotalCount()`

tests/framework/data/ActiveDataProviderTest.php
Kolyunya: @dynasource we may want to test the same thing (that a query was not performed)  if some other tests.

samdark (approved):

dynasource (approved):

samdark:
Merged. Thanks!

Issue #13567, opened 2017-02-14
afterSave dont refresh relational model


markmarco16:
### What steps will reproduce the problem?
Relational model data dont refresh afterSave. If you customize the afterSave Model with a relational model, the first model stay last data save for example:

Appointment have realtion by dealer
```php
public function getDealer()
{
      return $this->hasOne(Dealer::className(), ['id' => 'dealer_id']);
}
```

Dealer 1
name = Toyota

Dealer 2
name = Ford

````php

$model = Appointment->findOne($id);
//current dealer_id = 1

//update this dealer_id
$model->dealer_id = 2;
$model->save();

````

Aftersave model on Appointment send a email with $this var

````php
public function afterSave($insert, $changedAttributes)
{
        parent::afterSave($insert, $changedAttributes);
var_dump($this->dealer_id);
//return 2 is ok
var_dump($this->dealer->id);
//return 1 is wrong
var_dump($this->dealer->name);
//return Toyota is wrong
         return true;
}
````

### What is the expected result?
New relational model data of new value

### What do you get instead?
Old relational model data saved.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 5.6.29-0+deb8u1
| Operating system | GNU\Linux - Debian

samdark:
Since you won't need related model immediately most of the time (in forms) AR is avoiding extra query to refresh data. You may trigger refresh by calling same-named method if you need it. This is by design.

rob006:
@samdark You won't need extra query since you don't need related object immediately. The solution could be just unsetting related object when FK field has changed - extra query would be executed only when user access relation again.

samdark:
Ah, right. That's an option.

dynasource:
somewhat related https://github.com/yiisoft/yii2/issues/12749#issuecomment-254182125

hirenbhut93:
@samdark What is status of this issue or How to resolve this issue?

samdark:
Going to be fixed in 2.0.14.

yujin1st:
I have shop with Order and OrderItem models.
With new order in controller I load data from $_POST, create relational OrderItem records and populate relation 'items' for validation and in-events processing purposes  (for example warehouse module can check item quantity before order creating), and saving them in Order::afterSave.

Now in 2.0.14 after relation is flushed i have no data in Order::afterSave as models were new, my items are lost. What should i do?

SilverFire:
[...]

PR #13618, opened 2017-02-19
Reset outdated related models after AR update

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13567 

I'm not sure if it should be called directly in `afterSave` or we should bind an event handler.


framework/db/BaseActiveRecord.php
```diff
@@ -1647,4 +1652,28 @@ public function offsetUnset($offset)
             unset($this->$offset);
         }
     }
+
+    /**
+     * Resets outdated related models checking if corresponding link keys were changed or not.
+     * @param array $changedAttributes The old values of attributes that had changed and were saved.
+     */
+    private function resetOutdatedRelated($changedAttributes)
```
samdark: resetOutdatedRelations?

tests/framework/db/ActiveRecordTest.php
```diff
@@ -894,6 +894,20 @@ public function testRelationWhereParams($orderTableName, $orderItemTableName)
         OrderItem::$tableName = null;
     }
 
+    public function testRelationAfterSave()
```
samdark: Should be descriptive i.e. `testOutdatedRelationsAreResetAfterSave`

samdark (changes requested):

framework/db/BaseActiveRecord.php
Kolyunya: It meant to be consistent with the related attribute name. Gonna rename it.

Kolyunya:
@samdark fixed.

framework/db/BaseActiveRecord.php
```diff
@@ -944,13 +944,18 @@ public function beforeSave($insert)
      * when the password had changed or implement audit trail that tracks all the changes.
      * `$changedAttributes` gives you the old attribute values while the active record (`$this`) has
      * already the new, updated values.
+     * @since 1.0.12 Outdated related models will be reset before triggering the [[EVENT_AFTER_UPDATE]] event.
```
samdark: 2.0.12 :)

samdark (changes requested):

rob006:
Why this behavior is used only after save? The same problem exist without saving record (in validation for example):

```php
$model = new FormModel();
$model->user_id = $id;
if (!isset($model->user)) {
    $model->user_id = $defaultUserId;
}

var_dump($model->user); // null
```

Kolyunya:
@rob006 thanks for pointing this out. Am I missing something or we won't be able to tell if a relation is outdated or not while insert and the only option is to reset all relations?

Kolyunya:
Added a test case for the insert scenario.

rob006:
You should not trigger this on insert/update but on `setAttribute()`. You could create list of "dependencies" of cached relations. For example when I access to `$model->user` this realtion is dependent on current value of `$model->user_id`. You could check in `setAttribute()` if attribute is dependency and if it has changed - then unset all dependent relations.

Kolyunya:
@rob006 seems like this should be triggered both in `setAttribute()` and the `__set()` since one doesn't call the other for some reason.

Kolyunya:
Refactored the code. Please review.

cebe:
While I understand the problem, the presented solution adds a huge overhead to an operation which is performed quite often. Setting one attribute now includes an iteration over all relations and instantiating a relation object for each of them. Quite sure we do not want to introduce this overhead here. Any ideas?

Kolyunya:
@cebe first I proposed to implement the related models reset on `save()` but it was rejected...

rob006:
@cebe I have some ideas how to improve performance of this solution, but I need more time to do some tests. I will try propose solution in next week.

Kolyunya:
@cebe if we reset related models in `afterSave()` it will add no overhead on insert and just one iteration over `$_related` on update? What do you think about it?

Kolyunya:
@cebe @rob006 @samdark I refactored the code to store relation names indexed by link attributes. Now the reset routine has almost no overhead. What do you think?

dynasource:
IMO, these kind of responsibilities should be moved to its own class to keep BaseActiveRecord clean

samdark:
@dynasource not in this case. That's clearly a responsibility or AR if we want such behavior.

dynasource:
> That's clearly a responsibility or AR if we want such behavior.

responsibilities can be cut into pieces (a.k.a. classes)

rob006:
@dynasource This is about handling magic getters and setters specific to AR - how do you want to extract this into separate class?

dynasource:
by instantiating an object in which the logic of this PR is encapsulated. You can argue performance, but I'm positive against refactoring BaseActiveRecord into a smaller class.

samdark:
Such object would violate AR encapsulation and will be really tightly coupled to AR so I don't see how it would be helpful.

dynasource:
> so I don't see how it would be helpful.

 separation of concerns

samdark:
But that's not separate concern. It is internal object behavior to ensure its external consistency. It has nothing to do with external interface. It is an implementation detail and, ideally, should not be leaked.

dynasource:
My point is that I don't think it's a bad idea to break down the class in more peaces. The current BaseActiveRecord class has 1600 lines. IMO, this is an indication that is just does too much.

samdark:
In this regard — yes. AR does too much by design of the pattern itself... I would love to extract some code from it but only the code that is general enough to be reused somewhere else and which would not lead to encapsulation leaks.

This particular case is not a good example of code to be extracted.

dynasource:
perhabs not, but it's good to start looking at BaseActiveRecord to find abstractions in order to make it slimmer, starting from this PR.

samdark:
It's a good intent but is not related to this pull request in any way. Let's keep it out of it.

Kolyunya:
Any comments regarding proposed solution with `private static $_relations = [];`?

samdark:
Making it static could be a bad idea if you'll think about something like single table inheritance or polymorphic inheritance or any other form of dynamic relations.

Kolyunya:
Makes sense. Refactored to non-static `private $_relations;`.

SilverFire (changes requested):
Generally good.
@Kolyunya could you add tests for composite key relations and `via()` relations?

Kolyunya:
@SilverFire I added a test-case for the composite key relation which passes successfully and a test-case for the `via()` relation which fails. I tried to implement the `via()` relation support but it's getting very complex and I wasn't able to work out a solution for that yet. Not sure it it worth effort either.

rob006:
I've made some improvements: 
* better performance
* add missing db settings for tests (oci and mssql needs some review - I have no experience with these DBMS and can't test it properly)
* add `via()` support

See https://github.com/Kolyunya/yii2/pull/1

SilverFire (approved):

SilverFire:
Thank you, @rob006

samdark (approved):

framework/db/BaseActiveRecord.php
```diff
@@ -1647,4 +1670,30 @@ public function offsetUnset($offset)
...
+
+    /**
+     * Resets dependent related models checking if their links contain specific attribute.
+     * @param string $attribute The changed attribute name.
+     */
+    private function resetDependentRelations($attribute)
+    {
+        foreach ($this->_relationsDependencies[$attribute] as $relation) {
+            unset($this->_related[$relation]);
+            unset($this->_relationsDependencies[$attribute]);
```
rob006: This line should be moved below foreach block.

rob006 (changes requested):

framework/db/BaseActiveRecord.php
Kolyunya: Fixed.

cebe:
I think this can cause some unexpected problems:

e.g. the following approach is not uncommon:

```php
$user = new User();
$user->populateRelation('profile', new Profile());

if ($user->load($_POST) && $user->profile->load($_POST) && $user->validate() && $user->profile->validate()) {

    $user->profile->save(false);
    $user->profile_id = $user->profile->id; // this would reset the relation
    $user->save(false);

    $user->profile->handleFileUpload(); // this method would rely on the original profile class being used as that contains the UploadedFile instance

    return ...;
}

```

Current solution also does not respect relations populated via eager loading (`with()`) as no dependency is registered in that case.

I am still unsure whether we should add automatic invalidation of relations. Imo a call to `refresh()` is the better solution if you need to replace existing relations. btw, if you set related model via `link()` you get the refresh already. Example from #13567:

```php
$model = Appointment->findOne($id);
//current dealer_id = 1

//update this dealer_id
$model->dealer_id = 2;
$model->save(false);
```
update the id via link():

```php
$dealer = Dealer::findOne(2)
$model = Appointment::findOne($id);
//current dealer_id = 1
$model->link('dealer', $dealer);
$model->save(false);
```

Imo it is okay that extra work is needed when you do things manually (e.g. `$model->dealer_id = 2;`).
If you want magic, you can use `$model->link('dealer', $dealer);`

Kolyunya:
@cebe yeah these changes are very major and need extremely thoughtful review and testing!

rob006:
> e.g. the following approach is not uncommon:

This code is weird. Why you first save record to database and then handle files? Shouldn't be opposite?
Why you use relation for new `Profile` model instead of create 2 independent objects and link them at the end?

> `    $user->profile_id = $user->profile->id; // this would reset the relation`

Are you sure? Dependency for `profile_id` was newer set in this case, so [this condition](https://github.com/yiisoft/yii2/pull/13618/files#diff-e1e996b8f4fad49a953291c4f5fc0931R308) will newer pass.

> Imo it is okay that extra work is needed when you do things manually

The problem is that it is not always obvious when you should do that. If some validator touches relation, it will cache result and this will affect all further processing. It may create weird bugs - if I add validator which does not modify any attribute, but accessing relation, this may break some controllers logic if they rely on this relation. So to avoid this bug I need check all model validators on writing controller, and check all controllers which uses this model when I want to add new validator.

cebe:
> Why you use relation for new Profile model instead of create 2 independent objects and link them at the end?

I am not saying that is the best approach but I have seen people doing it. This allows to use `$user->profile` in the view also and would fetch the relation automatically, so update action has less code. At least this is a use case which can be broken by these changes.



> Are you sure? Dependency for profile_id was newer set in this case, so this condition will newer pass.

right, not in my example, but in update action the first line would be `$user = User::findOne($id);` 
and profile would be lazily loaded.

> If some validator touches relation

Why would a validator touch a relation? validation should be per model instance and not validating sub-relations.

rob006:
> right, not in my example, but in update action the first line would be $user = User::findOne($id);
and profile would be lazily loaded.

But in this case you're not having this line: `$user->profile_id = $user->profile->id;` since `$user->profile_id` is already set to correct value.

> Why would a validator touch a relation? validation should be per model instance and not validating sub-relations.

Validator may use data from relation to validate current model. For example if we have `Message` and `Conversation` models, `Message` validation may check if conversation is not closed. Also: #13675

cebe:
nothing for 2.0.13

tests/framework/db/ActiveRecordTest.php
```diff
@@ -894,6 +898,99 @@ public function testRelationWhereParams($orderTableName, $orderItemTableName)
         OrderItem::$tableName = null;
     }
 
+    public function testOutdatedRelationsAreResetForNewRecords()
+    {
+        $orderItem = new OrderItem();
+        $orderItem->order_id = 1;
+        $orderItem->item_id = 3;
+        $orderItem->quantity = 5;
+        $orderItem->subtotal = 80.0;
```
dynasource: `quantity` and `subtotal` are not used. This pollutes the test

tests/framework/db/ActiveRecordTest.php
```diff
@@ -894,6 +898,99 @@ public function testRelationWhereParams($orderTableName, $orderItemTableName)
...
+        // Test `__set()`.
+        $orderItem->order_id = 2;
+        $orderItem->item_id = 1;
+        $this->assertEquals(2, $orderItem->order->id);
+        $this->assertEquals(1, $orderItem->item->id);
+
+        // Test `setAttribute()`.
+        $orderItem->setAttribute('order_id', 2);
+        $orderItem->setAttribute('item_id', 2);
+        $this->assertEquals(2, $orderItem->order->id);
```
dynasource: `order_id` was `2` already so this is a confusing test. Perhabs this is on purpose, in that case it would helpful to supply a third argument in the assertion to explain what you are testing (ie. 'check if order_id can be updated with it own value')

tests/framework/db/ActiveRecordTest.php
```diff
@@ -894,6 +898,99 @@ public function testRelationWhereParams($orderTableName, $orderItemTableName)
...
+        $order->id = 2;
+        sort($orderItemIds);
+        $orderItemIds = ArrayHelper::getColumn($order->items, 'id');
+        $this->assertSame([3, 4, 5], $orderItemIds);
+
+        unset($order->id);
+        $this->assertSame([], $order->items);
+
+        $order = new Order();
+        $this->assertSame([], $order->items);
```
dynasource: does this assertion belong to this test methods scope?

framework/db/BaseActiveRecord.php
```diff
@@ -1647,4 +1670,30 @@ public function offsetUnset($offset)
...
+     */
+    private function resetDependentRelations($attribute)
+    {
+        foreach ($this->_relationsDependencies[$attribute] as $relation) {
+            unset($this->_related[$relation]);
+        }
+        unset($this->_relationsDependencies[$attribute]);
+    }
+
+    private function setRelationDependencies($relationName, $relation)
```
dynasource: missing method doc

dynasource (changes requested):
nice work

rob006 (approved):

samdark:
@Kolyunya would you please resolve conflicts?

samdark:
Did that... waiting for tests to run...

framework/db/BaseActiveRecord.php
samdark: @Kolyunya would you please add it?

samdark (approved):
Looks good to me except what's mentioned by @dynasource.

Kolyunya:
@samdark this can not be merged as is IMO. Tests for `via()` relations are missing and I'm sorry but I can not help with this in near future.

rob006:
@Kolyunya There are tests for `via()`: https://github.com/yiisoft/yii2/pull/13618/files#diff-5c2b59d39a2ea1dc5443fb98d1325d11R1040

Anything else should be tested?

framework/db/BaseActiveRecord.php
```diff
@@ -1682,4 +1707,35 @@ public function offsetUnset($offset)
...
+
+    /**
+     * Sets relation dependencies for a property
+     * @param string $name property name
+     * @param ActiveQueryInterface $relation relation instance
+     */
+    private function setRelationDependencies($name, $relation)
+    {
+        if (empty($relation->via)) {
+            foreach ($relation->link as $attribute) {
```
samdark: `link` isn't part of `ActiveQueryInterface`. Probably it's better to hint against `ActiveQuery` in the method signature...

samdark:
Merged. Thank you!

samdark:
@mg-code try master code.

mg-code:
@samdark it did not help.
In most cases those changes will help, but in some cases not.

It will work if property is accessed for the first time. If you will write such code:
```
if ($model->address) {
    $model->populateRelation('address', new Address());
}
```
it still will reset related data on after save. Because you tried to access relation before `populateRelation`.

mg-code:
Yes, it's the same.

zebraf1:
_relationsDependencies has private access. There is no way to override or fix a problem with this functionality unless I make a fork (or maybe use another orm).
I see this alot in Yii - when you make a private property at least make a public/protected function to get its value; or better yet make it protected if it's ok to change it.

PR #13621, opened 2017-02-19
Remove trailing spaces from sources and tests

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none

Github thinks that I removed the trailing empty line from the `tests/data/ar/Type.php` but it seems like it's still there.


samdark:
Merged. Thanks!

Issue #8120, opened 2015-04-18
LIKE escaping for Cubrid/MSSQL/Oracle/SQLite is wrong in QueryBuilder


sergeymakinen:
Hi!

I'm using Yii 2.0.3 and MSSQL server 2008 R2 and `QueryBuilder::buildLikeCondition()` is currently wrong for MSSQL. It works sometimes, but not if a second operand contains `%`, `_`, `[` or `]`. They should be escaped as `[CHAR_TO_ESCAPE]` or you should add `ESCAPE '\'` in the end of a LIKE statement (and you still need to treat `[` and `]` for MSSQL).

See https://msdn.microsoft.com/en-us/library/ms179859.aspx for more information.

**2017-02-20**:
While solving this issue in #13623, I discovered that the following drivers are also affected:

Driver | Problem
--- | ---
Cubrid | May or may not have an escape character depending on settings. See http://www.cubrid.org/manual/93/en/sql/function/condition_fn.html#like
Oracle | Doesn’t have an escape character by default. See http://docs.oracle.com/cd/E11882_01/server.112/e41084/conditions007.htm#SQLRF52141
SQLite | Doesn’t have an escape character by default. See http://sqlite.org/lang_expr.html#like

PR #13623, opened 2017-02-19
Fixes LIKE special characters escaping for Cubrid/MSSQL/Oracle/SQLite in yii\db\QueryBuilder (#8120)

sergeymakinen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #8120

I report - I fix. (tm) :)
I could fix this with no upgrade note but I don’t think that copy-pasting in this case is a smart idea. But it’s up to you. Or I can move `$escapeChars` to `QueryBuilder` as a protected property.


samdark:
We can probably introduce a method like `protected getLikeEscapeCharacters()` and use it to keep backwards compatibility of the existing interface.

sergeymakinen:
I modified tests to test LIKE special characters and discovered that Oracle, Cubrid and SQLite all suffer from this bug (confirmed with a manual testing). They simply all imply `ESCAPE ‘’` by default.

I want to fix them all and also improve tests for this case.

@samdark, should I:
- modify #8120
- open individual issues per driver
- open a common issue for these 3

?

I also stick to

```php
protected $likeCharacterReplacements = [
    ‘la-la’ => ‘bla-bla’,
    //
];

protected $likeEscapeCharacter = null;
```

as it looks like exactly what we need (luckily all dbs have the same syntax for escape characters).

samdark:
Modify #8120 and this pull request.

sergeymakinen:
Did it in the issue. I will update this PR as soon as all tests pass.

sergeymakinen:
Okay, it’s ready.

framework/db/QueryBuilder.php
```diff
@@ -65,7 +65,22 @@ class QueryBuilder extends \yii\base\Object
...
+        '%' => '\%',
+        '_' => '\_',
+        '\\' => '\\\\',
+    ];
+    /**
+     * @var string|null character used to escape special characters in LIKE conditions.
+     * By default it's assumed to be `\`.
+     * @since 2.0.12
+     */
+    protected $likeEscapeCharacter = null;
```
samdark: There's no need to specify `= null` explicitly.

samdark (changes requested):

framework/db/QueryBuilder.php
```diff
@@ -65,7 +65,22 @@ class QueryBuilder extends \yii\base\Object
...
+     */
+    protected $likeEscapingCharacterReplacements = [
+        '%' => '\%',
+        '_' => '\_',
+        '\\' => '\\\\',
+    ];
+    /**
+     * @var string|null character used to escape special characters in LIKE conditions.
+     * By default it's assumed to be `\`.
+     * @since 2.0.12
```
dynasource: shouldnt we set the default to be '\\' as it is set in most DBMSes?

dynasource (changes requested):

framework/db/QueryBuilder.php
sergeymakinen: This will add `ESCAPE '\'` where it is already by default. TBH, I did it exactly like you suggested but then reverted not to generate unneeded SQL. I think both ways are okay and safe and mostly a matter of preference.

sergeymakinen: A good thing about `'\\'` is that I can simplify tests and get rid of regex magic (and I consider this is good :).

dynasource (approved):

framework/db/QueryBuilder.php
dynasource: Ive looked at it again and Im fine with how it is

SilverFire (approved):

framework/db/QueryBuilder.php
```diff
@@ -65,7 +65,22 @@ class QueryBuilder extends \yii\base\Object
         'EXISTS' => 'buildExistsCondition',
         'NOT EXISTS' => 'buildExistsCondition',
     ];
-
+    /**
+     * @var array map of chars to their replacements in LIKE conditions.
+     * By default it's configured to escape `%`, `_` and `\` with `\`.
+     * @since 2.0.12.
+     */
+    protected $likeEscapingCharacterReplacements = [
```
samdark: Could be named shorter as `$likeEscapingReplacements`.

framework/db/mssql/QueryBuilder.php
```diff
@@ -45,6 +45,20 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_MONEY => 'decimal(19,4)',
     ];
 
+    /**
+     * @inheritdoc
+     */
+    protected $likeEscapingCharacterReplacements = [
```
samdark: Could be named shorter as `$likeEscapingReplacements`.

samdark (changes requested):

framework/db/QueryBuilder.php
sergeymakinen: Done.

samdark (approved):

samdark:
Thank you!

sergeymakinen:
Thanks too! The best is yet to come… ;)

PR #13624, opened 2017-02-19
Fix bug with optional params at the beginning of pattern in `yii\web\UrlRule`

### LINKED ISSUE #13086: Error with rules which include a defaults

### LINKED ISSUE #13656: Duplicate slashes if set param 'suffix' of the component UrlManager to '/'
#13522
```php
[
    'suffix'     => '/',
    ['pattern' => '<page:\d+>', 'route' => 'site/index', 'defaults' => ['page' => 1]],
]
```

Then create a URL like that

Url::to(['site/index', 'page' => 1])

The generated URL should be
```text
<a href="/">Home</a>
```
### What do you get instead?
```text
<a href="//">Home</a>
```
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.0.15
| Operating system | Centos 6

Issue is not reproducible on Yii 2.0.10

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13086

fixes #13086
fixes #13656
closes #13614


framework/web/UrlRule.php
```diff
@@ -96,6 +96,14 @@ class UrlRule extends Object implements UrlRuleInterface
...
      */
     public $normalizer;
+    /**
+     * @var bool Value used when pattern contains only optional params. If `true` rule pattern will contain slash at
+     * the end of param pattern (last optional param can not be omitted if all other patterns are not omitted), if `false`
+     * rule pattern will contain slash at the beginning of param pattern (first optional param can not be omitted
+     * if all other patterns are not omitted).
+     * @since 2.0.12
+     */
+    public $trailingSlashForDefaults = false;
```
samdark: Both description and name are quite vague. If you're not into the issue, it's quite hard to get the meaning...

samdark: Should it be configurable at all?

samdark (changes requested):

framework/web/UrlRule.php
rob006: I have no idea how to describe/name it better. I'm open to suggestions.

> Should it be configurable at all?

I decided to make it configurable because it is more flexible, but not sure if it is worth of API complication.

Kolyunya: I was aware of the issue but when I fist saw this property I was totally confused.

samdark: @rob006 why would you want buggy version of this behavior?

rob006: @samdark Why buggy? If I have pattern like `<language>/<category>/<slug>` when all part are optional, I may have want omit params from beginning, not from the end. With `$trailingSlashForDefaults = true` valid URLs will be:
* `en/my-default-category/project-slug`
* `my-default-category/project-slug`
* `project-slug`

With `$trailingSlashForDefaults = false` valid URLs will be:
* `en/my-default-category/project-slug`
* `en/my-default-category`
* `en`

Both cases seems valid to me.

samdark: Ah. Yes. I wonder how to name it... hard case.

Kolyunya: What about `my-default-category` URL in both scenarios?

rob006: > What about the following URLs in both scenarios?

First will never work, second will work on both cases.

dynasource: > If you're not into the issue, it's quite hard to get the meaning...

when the issue is described in 1 sentence, you're already closer to a proper name

SilverFire: Maybe, it worth to add some examples to explain what does this parameter do?

SilverFire:
Related https://github.com/yiisoft/yii2/issues/13656

framework/web/UrlRule.php
SilverFire: How about

```php
const OMIT_FIRST = 0;
const OMIT_LAST = 1;

public $optionalParamParsingMode;
```

samdark: Yes. That is better.

samdark: @rob006 what do you think?

rob006: I'm going to remove this property. It brings unnecessary complications in the API, and the same effect can be achieved with only one additional rule.

rob006:
I removed this property. Let's focus on fixing this bugs, additional configuration options could be added in future.
🚢

rob006:
Bump.

lynicidn:
1. помоему это ломает обратную совместимость
2. defaultParams в тестах не нужны, так как во все тесты передаются аргументы и не один тест не проверят именно default (когда не переданы какие либо или все параметры)
3. не хватает тестов для parseRequest с default

samdark:
@lynicidn @rob006 can't read your commen...

samdark:
@rob006 the comment from @lynicidn was the following:

1. I think it breaks backwards compatibility.
2. `defaultParams` aren't needed in tests since all the tests are supplied with arguments and no test is checking for default (when one or more params are missing).
3. It's missing tests for `parseRequest` with `default`.

rob006:
> I think it breaks backwards compatibility.

Any examples?

> defaultParams aren't needed in tests since all the tests are supplied with arguments and no test is checking for default (when one or more params are missing).

https://github.com/yiisoft/yii2/pull/13624/files#diff-96fb0ef29c09f20871090c41bf5b3839R999 ?

> It's missing tests for parseRequest with default.

https://github.com/yiisoft/yii2/pull/13624/files#diff-96fb0ef29c09f20871090c41bf5b3839R971 ?

framework/web/UrlRule.php
```diff
@@ -179,6 +179,11 @@ public function init()
             $this->name = $this->pattern;
         }
 
+        $this->preparePattern();
+    }
+
+    private function preparePattern()
```
samdark: phpdoc missing

framework/web/UrlRule.php
```diff
@@ -212,6 +217,11 @@ public function init()
             }
         }
 
+        $this->translatePattern(true);
+    }
+
+    private function translatePattern($allowAppendSlash)
```
samdark: phpdoc missing

framework/web/UrlRule.php
```diff
@@ -232,14 +245,34 @@ public function init()
...
                     $offset = $match[0][1];
-                    if ($offset > 1 && $this->pattern[$offset - 1] === '/' && (!isset($this->pattern[$offset + $length]) || $this->pattern[$offset + $length] === '/')) {
+                    $requiredPatternPart = str_replace("/{$match[0][0]}/", '//', $requiredPatternPart);
+                    if (
+                        $allowAppendSlash
+                        && ($appendSlash || $offset === 1)
+                        && (($offset - $oldOffset) === 1)
+                        && isset($this->pattern[$offset + $length])
+                        && $this->pattern[$offset + $length] === '/'
+                        && isset($this->pattern[$offset + $length + 1])
```
samdark: This one could be merged into previous isset:

```php
isset($this->pattern[$offset + $length], $this->pattern[$offset + $length + 1])
```

samdark (changes requested):

framework/web/UrlRule.php
rob006: This is private method - to we need phpdoc for these?

samdark: Yes.

framework/web/UrlRule.php
rob006: I think it will be less readable - it is easy to miss additional params in `isset`.

samdark: Makes sense.

framework/web/UrlRule.php
rob006: Done.

samdark (approved):

SilverFire (approved):

samdark:
Merged. Thank you very much!

PR #13582, opened 2017-02-15
Implemented `yii\db\mssql\QueryBuilder::resetSequence()`

boboldehampsink:
Mentioned in #13577 and #13581 and compatible with SQL Server 2008+.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13577

PR #13630, opened 2017-02-20
Added tests for all `yii\db\QueryBuilder::resetSequence` implementations

boboldehampsink:
In #13582 we added this to the MSSQL driver, and agreed to add these for all drivers.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13582

### DIFF
--- a/framework/db/sqlite/QueryBuilder.php
+++ b/framework/db/sqlite/QueryBuilder.php
@@ -134,13 +134,13 @@ class QueryBuilder extends \yii\db\QueryBuilder
                 $key = reset($table->primaryKey);
                 $tableName = $db->quoteTableName($tableName);
                 $value = $this->db->useMaster(function (Connection $db) use ($key, $tableName) {
-                    return $db->createCommand("SELECT MAX('$key') FROM $tableName")->queryScalar();
+                    return $db->createCommand("SELECT MAX($key) FROM $tableName")->queryScalar();
                 });
             } else {
                 $value = (int) $value - 1;
             }
             try {
-                $db->createCommand("UPDATE sqlite_sequence SET seq='$value' WHERE name='{$table->name}'")->execute();
+                return "UPDATE sqlite_sequence SET seq='$value' WHERE name='{$table->name}'";
             } catch (Exception $e) {
                 // it's possible that sqlite_sequence does not exist
             }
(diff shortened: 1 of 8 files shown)


samdark (approved):

dynasource (approved):
thanks!

samdark:
https://travis-ci.org/yiisoft/yii2/jobs/203429387#L589

boboldehampsink:
I actually found a bug in the SQLite implementation that I have now fixed.

boboldehampsink:
Okay, locally this test runs just fine. As you can see in the sqlite fixture the table "item" does get created. Maybe there is something wrong with the travis setup?

samdark:
Maybe. I'll re-run locally...

samdark:
@dynasource now it works as expected. Thanks!

samdark:
Merged. Thanks!

framework/db/sqlite/QueryBuilder.php
```diff
@@ -134,13 +134,13 @@ public function resetSequence($tableName, $value = null)
                 $key = reset($table->primaryKey);
                 $tableName = $db->quoteTableName($tableName);
                 $value = $this->db->useMaster(function (Connection $db) use ($key, $tableName) {
-                    return $db->createCommand("SELECT MAX('$key') FROM $tableName")->queryScalar();
+                    return $db->createCommand("SELECT MAX($key) FROM $tableName")->queryScalar();
```
cebe: key should be quoted the same way as it was fixed in pgsql #13582

framework/db/sqlite/QueryBuilder.php
```diff
@@ -134,13 +134,13 @@ public function resetSequence($tableName, $value = null)
...
                 });
             } else {
                 $value = (int) $value - 1;
             }
             try {
-                $db->createCommand("UPDATE sqlite_sequence SET seq='$value' WHERE name='{$table->name}'")->execute();
+                return "UPDATE sqlite_sequence SET seq='$value' WHERE name='{$table->name}'";
             } catch (Exception $e) {
```
cebe: this try and catch does not make sense anymore if the command is not executed here.

framework/db/sqlite/QueryBuilder.php
boboldehampsink: As in `{$key}`

framework/db/sqlite/QueryBuilder.php
boboldehampsink: Indeed

boboldehampsink:
@cebe I submitted a new PR to deal with the issues you've found: #13755

PR #13634, opened 2017-02-21
[Guide] Issue in retrieval list from the DB in Input form

hirenbhut93:
## When does the issue occur?

When I have generate `dropDownList` or `radioButtonList` from DB at time return same key and value like

```php
return [
    2 => 2,
    3 => 3
]
``` 

because `column` method is return first column in `select` statement. when i change position in select statement like set `label` field first at time return my expected results like.

## What was the expected result?

```php
return [
    2 => 'My label/title-1',
    3 => 'My label/title-2'
]
``` 

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/docs/guide/input-forms.md
+++ b/docs/guide/input-forms.md
@@ -144,7 +144,7 @@ or by retrieval from the DB:
 
 ```php
 $items = Category::find()
-        ->select(['id', 'label'])
+        ->select(['label'])
         ->indexBy('id')
         ->column();
 ```


hirenbhut93:
@samdark Ok have updated issue descriptions

samdark:
Still don't get how to reproduce the issue...

hirenbhut93:
@samdark Please refer this link 
http://www.yiiframework.com/doc-2.0/yii-db-query.html#column()-detail
http://www.yiiframework.com/doc-2.0/yii-db-command.html#queryColumn()-detail

dynasource:
your issue description is quite cryptic. Can you try to explain it simple in one line?

samdark:
Ah, I think I've got it. Will check in detail later.

SilverFire (approved):

samdark:
Thank you!

cebe:
how would this method be able to index by ID if ID is not selected?

Issue #13637, opened 2017-02-21
Setting `logger` property in Dispatcher doesn't affect Yii::getLogger()


antonmarin:
### What steps will reproduce the problem?
in config set: 
```
components => [
    log => [
        class => yii\log\Dispatcher,
        logger => someLoggerClass,
    ]
]
```
### What is the expected result?
Yii::getLogger() instance of someLoggerClass
### What do you get instead?
Yii::getLogger() instance of yii\log\Logger

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11
| PHP version      | 5.6
| Operating system | MacOS

antonmarin:
I think that Dispatcher->setLogger(logger) should call Yii::setLogger(logger)

dynasource:
thanks for your feedback. Please look at the following topic https://github.com/yiisoft/yii2/issues/12163 for more info about this subject

antonmarin:
Um. Do you mean that i should call \Yii::setLogger(MyLogger::class); in my index.php ?
That's really ugly solution. Is there another way?

dynasource:
right in your bootstrap.php (or before instantiating your Application).  More info at https://github.com/yiisoft/yii2/issues/7555#issuecomment-131794988

antonmarin:
That's better, but doing like this i will config my logger in two places. Is this really necessary? Why don't just call Yii::setLogger()?

dynasource:
setting in once in your bootstrap should be enough. We also have a guide for you with more info on this topic http://www.yiiframework.com/doc-2.0/guide-runtime-logging.html#log-targets

antonmarin:
i mean that config in main.php will still exist and another part of config (setting logger) will be in different place

antonmarin:
@dynasource what do you think about #13641 ?

Issue #12163, opened 2016-08-15
Incomplete logger depedency


Yiivgeny:
### What steps will reproduce the problem?
```php
\Yii::$app->getLog()->setLogger(\yii\log\Logger::class);
\Yii::$app->getLog()->setFlushInterval(123);

var_dump(\Yii::getLogger()->flushInterval); //missing: 123; actual: 1000
var_dump(\Yii::getLogger() === \Yii::$app->getLog()->getLogger()); //missing: true; actual: false
```

### Result
Configured any `Logger` via `$config['components']['log']['logger']` has no effect for `BaseYii` log methods (warning, info, etc).

### Internal
`\yii\log\Dispatcher::setLogger` don't sync `\yii\BaseYii::$_logger` property.

samdark:
Verified. The way it could be configured is via

```php
\Yii::setLogger(\Yii::createObject([
    'class' => \yii\log\Logger::class,
    'flushInterval' => 123,
]));
```

samdark:
@Yiivgeny any suggestions?

Yiivgeny:
`\Yii::$container->set('\yii\log\Logger', ...)` works better because your solution still problem:
```php
var_dump(\Yii::getLogger() === \Yii::$app->getLog()->getLogger()); //missing: true; actual: false
```
@samdark, мне кажется оба варианта неюзабельны, т.к. требуют дополнительных знаний и конфигурирований. К примеру, распространять расширение реализующее компонент log становится труднее, т.к. невозможно его инициализировать посредством конфига, т.к. класс \Yii еще не определен в этот момент.

klimov-paul:
@Yiivgeny, can you explain what are you trying to achive with your code?

Note that `yii\log\Logger` instance can not be configured at application level anyhow: its instantiation and usage predates the application creation itself.
`yii\log\Logger` instance can be configured or overridden properly only via `yii\BaseYii::setLogger()`!

Also note that each instance of `yii\log\Dispatcher` may use different `yii\log\Logger` instance. That is why `yii\log\Dispatcher::log` property exists.

> \yii\log\Dispatcher::setLogger don't sync \yii\BaseYii::$_logger property.

When you call:

```php
\Yii::$app->getLog()->setLogger(\yii\log\Logger::class);
```

You force your `Dispatcher` to work with another instance of `Logger`. So it works as expected.

Yiivgeny:
[...]

Issue #11397, opened 2016-04-21
Require PHP 7.1 for Yii 2.1


Faryshta:
## Decision

Update to PHP 7.1.

## TODO

- [x] Install APCu via travis instead of APC: https://github.com/yiisoft/yii2/pull/14505
- [x] Drop memcache (w/o D) support.
- [x] Use null coalesce: https://github.com/yiisoft/yii2/pull/16270
- [x] Remove old PHPUnit polyfills: https://github.com/yiisoft/yii2/pull/14506
- [x] Remove intl polyfills: https://github.com/yiisoft/yii2/pull/16273#pullrequestreview-119708556

## Original issue

Currently only php 5.6, 7 and 7.1 are supported

http://php.net/supported-versions.php

php 5.5 is in security-only support and will reach end of life very soon, while php5.6 will enter security-only support next year. Probably yii 2.1 will take around a year to develop, by that time the entire php5 branch will be obsolete.

Today ubunt 16.04 will be released which includes php7 as the default version and its an LTS so we can expect a lot of server providers to switch to it an offer php7 effortlessly during the next months.

Requiring php7 for yii2.1 can be very useful
- code generation in gii which can take full advantage of the type hintings.
- views and controllers can be simplified a lot with the `use` groups
- Interally yii2 has many unit tests and has a good dependency injection container, both of them can be benefited by the anonymous classes.
- Yii security class can benefit a lot of the new random generators.

I propose to make a poll using emoji's. I will post 3 options bellow and any reaction to them (:+1: :-1: etc) count as a vote for that option. That way someone can vote for more than one option.

Faryshta:
keep minimal version at php 5.5

Faryshta:
php 5.6 for minimal version required for yii2.1

Faryshta:
php 7.0 for minimal version required for yii2.1

mdmunir:
Its depend on when 2.1 stable version released. I am pessimistic it can be done in early next year :D.
depend on what default php version of `debian` and `CentOS`.

samdark:
Ubuntu LTS this summer is going to ship PHP 7 by default.

Faryshta:
@samdark actually it just got released today http://releases.ubuntu.com/16.04/

samdark:
Oh! Great.

kavitama:
I remember when we discussed the 2.0 minimum and at that time we were "cautious" and stick with 5.4 that became obsolete before YII2 was "spread out".
[...]

PR #13641, opened 2017-02-22
Dispatcher logger is primary and BaseYii logger is secondary

antonmarin:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13637, #12163

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -354,6 +354,8 @@ class BaseYii
     private static $_logger;
 
     /**
+     * @throws InvalidConfigException When call shortcut before initialization
+     *
      * @return Logger message logger
      */
     public static function getLogger()
@@ -362,7 +364,10 @@ class BaseYii
             return self::$_logger;
         }
 
-        return self::$_logger = static::createObject('yii\log\Logger');
+        throw new InvalidConfigException(
+            'Logger component must be initialized before using logging actions. ' .
+            'Add "log" component to the "bootstrap" section of the configuration on the first position.'
+        );
     }
 
     /**
--- a/framework/log/Dispatcher.php
+++ b/framework/log/Dispatcher.php
@@ -113,7 +113,7 @@ class Dispatcher extends Component
     public function getLogger()
     {
         if ($this->_logger === null) {
-            $this->setLogger(Yii::getLogger());
+            $this->setLogger('yii\log\Logger');
         }
         return $this->_logger;
     }
@@ -130,6 +130,7 @@ class Dispatcher extends Component
         }
         $this->_logger = $value;
         $this->_logger->dispatcher = $this;
+        Yii::setLogger($value);
     }
 
     /**
--- a/tests/framework/BaseYiiTest.php
+++ b/tests/framework/BaseYiiTest.php
@@ -97,10 +97,16 @@ class BaseYiiTest extends TestCase
         BaseYii::setLogger($logger);
 
         $this->assertSame($logger, BaseYii::getLogger());
+    }
 
+    /**
+     * @covers \yii\BaseYii::getLogger()
+     * @expectedException \yii\base\InvalidConfigException
+     */
+    public function testShouldThrowExceptionIfNoLoggerSetup()
+    {
         BaseYii::setLogger(null);
-        $defaultLogger = BaseYii::getLogger();
-        $this->assertInstanceOf(Logger::className(), $defaultLogger);
+        BaseYii::getLogger();
     }
 
     /**
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -48,6 +48,7 @@ abstract class TestCase extends \PHPUnit_Framework_TestCase
             'id' => 'testapp',
             'basePath' => __DIR__,
             'vendorPath' => $this->getVendorPath(),
+            'bootstrap' => ['log'],
         ], $config));
     }
 


antonmarin:
Seems like pr is ready now :)

cebe:
This is a breaking change, so if we fix it, this should be done in 2.1.

antonmarin:
@cebe it breaks only if you try logging before application initialization. Is it possible?

samdark:
Yes. It is what's done automatically for `trace` level if `YII_DEBUG` is true. It logs all the app init sequence.

antonmarin:
@samdark isn't log components bootstapped at this time?

samdark:
It should be.

antonmarin:
@samdark so, do i understand right that it will be ok if log component should always bootstrap because it used while application initialization? What about i add log component to BaseApplication->bootstrap?

samdark:
I think so. @cebe, @SilverFire, @dynasource, @klimov-paul what do you think?

dynasource:
IMO, the Application should not bootstrap anything by default at all. It should be a slim as possible, ideally just enough to survive.

Regarding the logging. This should be refactored in the framework to be optional, like is requested is several topics throughout the issue tracker. The problem is performance and hickups during processing of large data. It is hard for people to understand why it is occurring and to eventually understand it to be happing because of a Logger (which they seemed to have disabled).

antonmarin:
@dynasource as @samdark said there is logging all the app init sequence, so is it possible to make logger optional? I think the only way is to create some DummyLogger or remove logging from app init. If you prefer first way, so my PR is actual, and the other way is to remove both logging and bootstrap logger. Have i missed something?

antonmarin:
any updates?

samdark:
Actually, I've re-checked code and I'm not so sure if we're actually logging init sequence prior to log is initialized. A good way to check it is to set logging level to trace, enable debug, run a single request and save logs. Then check if logs are the same with this pull request.

antonmarin:
`yii\base\Controller:130 Yii::trace('Route to run: ' . $action->getUniqueId(), __METHOD__);`

antonmarin:
well what direction to move? set logger optional and check existence (not so pretty)? create dummy? or leave log bootstrapping?

antonmarin:
`\yii\base\Application:322 Yii::trace('Bootstrap with ' . get_class($component), __METHOD__);`

antonmarin:
Another way is to require Logger config as basePath

samdark:
Leave log bootstrapping. Bootstrap sequence is important and needs logging.

antonmarin:
ok. so what about merge? :)

tests/framework/BaseYiiTest.php
```diff
@@ -99,8 +100,7 @@ public function testSetupLogger()
         $this->assertSame($logger, BaseYii::getLogger());
 
         BaseYii::setLogger(null);
-        $defaultLogger = BaseYii::getLogger();
-        $this->assertTrue($defaultLogger instanceof Logger);
+        BaseYii::getLogger();
```
SilverFire: `\yii\base\InvalidConfigException` may be thrown on `Logger` initialization as well. This test should be split

framework/BaseYii.php
```diff
@@ -362,7 +364,7 @@ public static function getLogger()
             return self::$_logger;
         }
 
-        return self::$_logger = static::createObject('yii\log\Logger');
+        throw new InvalidConfigException('No logger without application');
```
SilverFire: I think this exception will be popular so we should explicitly explain what's going on.

How about `Logger component must be initialized before using logging actions. Add "log" component to the "bootstrap" section of the configuration on the first position.`?

SilverFire (changes requested):

dynasource (changes requested):
I would expect to see unit tests that reflect the title & the changes in this PR

antonmarin:
@dynasource, i don't understand. Could you explain in different words?

dynasource:
the title states "Same logger in Yii and Yii::$app->log->logger" but I can not see this by looking at the new unit tests. I can only see new tests about throwing errors which does not go into depth about your actual usecase.

antonmarin:
> DispatcherTest:62 already reflects current changes
$this->assertSame(Yii::getLogger(), $dispatcher->getLogger());

So, do i understang right, that you mean, that that test should be moved to separate test method?

samdark:
I think he means that the test should assert that logger is the same in both cases.

dynasource:
correct, and otherwise the title of this issue should be changed towards the real goal of this PR

antonmarin:
> I think he means that the test should assert that logger is the same in both cases.
> $this->assertSame(Yii::getLogger(), $dispatcher->getLogger());
isn't that do it?

antonmarin:
so i change title as test already exists

antonmarin:
any update?

framework/log/Dispatcher.php
```diff
@@ -113,7 +113,7 @@ public function init()
     public function getLogger()
     {
         if ($this->_logger === null) {
-            $this->setLogger(Yii::getLogger());
+            $this->setLogger('yii\log\Logger');
```
samdark: `\yii\log\Logger::class` could be used.

samdark (changes requested):

framework/log/Dispatcher.php
antonmarin: But Yii requires php 5.4?

sergeymakinen: This PR is aimed for Yii 2.1 which requires PHP 5.5 ATM.

TerraSkye: 2.1 is set for 7.1 if im not mistaken https://github.com/yiisoft/yii2/issues/11397

cebe: @TerraSkye correct, but that decision has been made after @sergeymakinen commented ;)

samdark:
I'm thinking about ditching our logger altogether and using PSR-3 compatible interface with monolog as default implementation.

sergeymakinen:
@samdark I couldn't agree more. Even though I love the Yii logger.

dynasource:
Yeah. Good idea

antonmarin:
great idea, so maybe i should close PR?

samdark:
Yes. Would be cool if you'll create another issue about using PSR-3 and reference this one. Thanks.

samdark:
Won't be merged into 2.0.

PR #13642, opened 2017-02-22
Allows override query relation in descendant class

leandrogehlen:
Moved from #13441


framework/db/BaseActiveRecord.php
```diff
@@ -411,12 +405,25 @@ public function hasOne($class, $link)
...
     public function hasMany($class, $link)
     {
+        return $this->createRelationQuery($class, $link, true);
+    }
+
+    /**
+     * Creates query instance for `has-one` and `has-many` relation.
+     * @param string $class the class name of the related record
+     * @param array $link the primary-foreign key constraint
+     * @param boolean whether this query represents a relation to more than one record.
```
samdark: Variable name `$multiple` missing.

samdark (approved):

samdark:
Need a line for changelog.

dynasource (approved):

SilverFire (approved):
:exclamation:  CHANGELOG should be adjusted before merge

cebe:
what was the use case for this? its not clear to me from #13441

leandrogehlen:
@cebe the relation queries are been created by `find` method, then if  i change the default behavior in a `MultiTenantRecord` this changes will happens in relational query.

Example: 

```php
class MultiTenantRecord extends ActiveRecord
{
    public static function find()
    {
        $identity = Yii::$app->user->identity;
       if ($identity instanceof TenantInterface) {
           return parent::find()->where([static::tableName().'.tenant_id' => $identity->getTenantId()]);
       } else {
           throw new NotSupportedException("Identity does not implements TenantInteface");
       }
    }
}

class Category extends MultiTenantRecord 
{
    ...
}
class Product extends MultiTenantRecord 
{
    public function getCategory()
    {
        return $this->hasOne(Category ::className(), ['id' => 'category_id']);
    }
}

Product::find()->joinWith('category')->all();// this generate something like:
```
```sql
select 
  * 
from product 
join category on product.category_id = category.id 
where product.tenant_id = 1 
  and category.tenant_id = 1 -- this is not necessary
```

cebe (approved):

cebe:
merged, thank you!

PR #13645, opened 2017-02-23
Applying A Callback To Each Model Collection as per the issue #13644

ElijahGM:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13644

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -225,18 +225,10 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             }
         }
         if (is_callable($this->iterator_callback)) { 
-            $temp=$models;
-            $models=[];
-            foreach ($temp as $model) {               
-                if($return=call_user_func($this->iterator_callback,$model)){
-                  $models[]=$return;
-                }else{
-                  throw new \Exception("Iterator callback must return data", 1);  
-                }
-                
-               
-            }
-            unset($temp);
+            array_walk($models,function($model,$index)use(&$models){
+               $modified = call_user_func($this->iterator_callback,$model);
+               $models[$index]=($modified)?$modified:$model;
+            });
         }      
 
         return $models;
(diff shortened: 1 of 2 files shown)


SilverFire:
What is the use case for this ehancement?

ElijahGM:
What is the use case for this enhancement? 
It provide the array_map() functionality on query of models, you can iterate on models modifying or filtering return objects 

instead of 

```
$models = Model::find()->all();
array_map($models,function($model,$index){
  //code
)))
```

It provide a neater way of doing the same 
```
Model::find()->iterate(function($model)
{
    return model->attribute=modify(model->attribute)
})->all();
```
Maybe is I should rename the function to map instead of iterate

samdark:
It's not map but iterate since it leaves objects as is. Also `return` looks odd in your example. Overall it's an equivalent of simple foreach:

```php
$models = Model::find()->->all();
foreach ($models as $model) {
    $model->attribute = modify($model->attribute);
}
```

samdark:
Foreach actually looks better...

ElijahGM:
@samdark I have corrected the return error, am still seeing the iterate being neat way to do it, without a lot of codes especially for those guys coming from Laravel where this is an inbuilt function

samdark:
Well, the `foreach` in this case works much faster and is actually shorter, isn't it?

ElijahGM:
@samdark i don't think so, but maybe am wrong 
`Model::find()->iterate(function)->all();` is in a single line where models returned are only those filtered/modified by the closure function,

ElijahGM:
@samdark if foreach in this case works much faster, i can use it in the design to do the iteration on the function applyClosure() function

samdark:
Overall, conceptually the pull request is not correct. If something like that is to be done, proper collections and populating models into collections should be introduced. Closing pull request, leaving issue open for discussion.

Issue #10700, opened 2016-01-29
custom validation error messages is not hiding after correcting itsinput


vc7deo:
i have problem with custom validation its not hiding even after correcting the input but other regular validation is working fine.Also if we correct the input we can submit.please check my code below

            [['arrival_date','return_date'], 'date', 'format' => 'php:d-m-Y'],
            ['arrival_date','validateDates'],

            public function validateDates(){
              if(strtotime($this->return_date) <= strtotime($this->arrival_date)){
              $this->addError('arrival_date','Please give correct Start and End dates');
              $this->addError('return_date','Please give correct Start and End dates');
              }
           }

SilverFire:
Do you have a problem with dependent field?
If you have error in both `arrival_date` and `return_date`, than change `arrival_date`, the error for this input disappears, but the error for `return_date` is still being displayed?

vc7deo:
yes..i removed Yes i removed `arrival_date ` error but `return_date` error still showing.Also these error will display only when we click on the submit button.so user don't know whether they enter right input till the input button is submitted..

SilverFire:
Partially duplicates #3987

You must trigger the second field validation manually

```js
    $('#arrival_date').update(function () {
        $(this).closest('form').yiiActiveForm('validateAttribute', 'return_date');
    });
```

vc7deo:
still not working.. you can see it here

http://flywingzdestination.com/flywings/frontend/web/service/airticket

vc7deo:
@ SilverFire ... did you check the issue?

SilverFire:
I can't reproduce the error in your form.

Issue #9811, opened 2015-10-01
CompareValidator client validation does not work with prefixed attributes


SilverFire:
In case of using form fields with prefixed attributes, the ID of input will be `model-prefix-attribute` (for example bellow `user-0-password`). During the generation of client validation JS in `CompareValidator`, the `compareAttribute` (which have to contain ID of target input) will be set to `model-password` instead of `model-0-password` (see `CompareValidator.php:219`), so the JS will generate false errors during the validation.

view.php:
```php
use yii\widgets\ActiveForm;

$form = ActiveForm::begin();
foreach ($models as $i => $model) {
    print $form->field($model, "[$i]password");
    print $form->field($model, "[$i]confirm_password");
}
$form->end();
```

model.php:
```php
class User extents Model {
...
    function rules() {
        return [
            [
                    ['password'],
                    'compare',
                    'compareAttribute' => 'confirm_password',
                    'on' => ['create'],
                ],
        ];
    }
...
}
```

Workaround: set `enableClientValidation` to `false` in model and use ajax validation with `ActiveForm::validateMultiple`

And I don't know, how to pass full attribute with prefix form `ActiveForm` to `CompareValidator`

tafid:
+1

Besedin86:
:+1:

bladeroot:
:+1:

SilverFire:
Any ideas, team?

SilverFire:
Verified :)

SilverFire:
Duplicated in https://github.com/yiisoft/yii2/issues/10230

MigueSchpeir:
Thanks @SilverFire for your assistance in #10230. Any idea on how to solve it?

SilverFire:
Any ideas how to fix it?
I thought before, that current client validation design may be enhanced. The main problem is that we are converting attribute names to IDs and get different problems:

 - We can not handle two forms of same model class on the page (DOM element ID is unique)
 - We are experiencing problems for validation, that depends on other attributes, especially for multiple forms with prefixed input names (as described in this issue)
 - It's hard to write custom JS that interacts with inputs in this item of multiple form
 - etc

I think that the most reliable option is to:
 - Add `data-attribute` to each input `<input name="User[0][id]" data-attribute="id" />`
 - Add a new logical container like `model-item` to disambiguate pointing on inputs.

For example:
```html
<form>
   <div class="model-item" data-id="0" data-model="User">
[...]

Issue #9500, opened 2015-08-21
yii.activeform.js integrate function to check if input is validated


BenasPaulikas:
I have form and 8 fields. What I want to do is in Javascript validate only 2 fields and return true or false depending on if they are both valid or no.

samdark:
```javascript
$('#contact-form').data('yiiActiveForm').submitting = true;
$('#contact-form').yiiActiveForm('validate');
if ($('#contact-form').find('.has-error').length) {
    // error
}
```

samdark:
https://github.com/samdark/yii2-cookbook/blob/master/book/forms-activeform-js.md

BenasPaulikas:
Correct me if I'm wrong but I think $('#contact-form').yiiActiveForm('validate') uses ASYNCHRONOUS methods. So when It start validation and I check for $('#contact-form').find('.has-error').length its equal to 0

samdark:
Ah. Yes.

BenasPaulikas:
console.log($('#contact-form').find('.has-error').length);
    setTimeout(function(){
        console.log($('#test-form').find('.has-error').length);
    }, 1000);

Output:

    0
    5

Alex-Code:
You could use the [`afterValidate`](https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.activeForm.js#L50) event.
That will be called after asynchronous validation.

BenasPaulikas:
afterValidate events works perfect :+1: 

My current code is and it works perfectly. I get message that "Form is validated"

    $('#test-form').data('yiiActiveForm').submitting = true;
    $('#test-form').yiiActiveForm('validate');
    $('#test-form').on('afterValidate', function(e){
        console.log('Form validated');
    });

However if I change 
    $('#test-form').yiiActiveForm('validate');
with
    $('#test-form').yiiActiveForm('validateAttribute', 'test-location');

It doesn't validate and I don't get valiadtion error :(

BenasPaulikas:
Thanks. afterValidate works perfect. :+1: 
However I want to validate 2 specific fields and as I see there's no event for that. I woud use

    afterValidateAttribute

but it requires whole form validation 

    afterValidateAttribute event is triggered after validating the whole form and each attribute.

nkovacs:
afterValidateAttribute is called in both cases (whole form and single attributes - the documentation is a bit confusing). afterValidate is called only after whole form validation.

SilverFire:
I've checked the use case.

[...]

PR #13646, opened 2017-02-23
Update yii.activeForm.js

depesr:
Patch for yii activeForm javascript to validate all fields dependless on their status attribute.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

### DIFF
--- a/framework/assets/yii.activeForm.js
+++ b/framework/assets/yii.activeForm.js
@@ -656,7 +656,7 @@
             }
         } else {
             $.each(data.attributes, function () {
-                if (!this.cancelled && (this.status === 2 || this.status === 3)) {
+                if (!this.cancelled) {
                     updateInput($form, this, messages);
                 }
             });


SilverFire:
It is by design and prevents validation of fields that were not changed.
The PR will not be accepted, sorry.

depesr:
@SilverFire Well, the thing is this.
I have one field which is conditionally required when other field has an exact value.
I change the other field to that exact value and the first one won't get error message unless focused and defocused?
Is there any other way to get the error message shown? The validation message from ajax validation is correctly in the response, just the javascript won't show it.

SilverFire:
Your use case is partially related to issue #9811

See 
https://github.com/yiisoft/yii2/issues/10700#issuecomment-176647176
https://github.com/yiisoft/yii2/issues/9500#issuecomment-166013061

depesr:
Didn't help :(
Can't trigger entire form ajax validation using javascript. But that's not the issue. The ajax validation works ok, it returns json with error message. The issue is that the error message won't show since yii.activeForm.js just shows messages for fields that are changed.

I've created an issue: https://github.com/yiisoft/yii2/issues/13648

PR #13650, opened 2017-02-23
Take advantage of hash_hkdf() on PHP >= 7.1.2

charlesportwoodii:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | n/a

PHP 7.1.2 implements a native `hash_hkdf()` implementation, which is faster than a pure PHP implementation. (see php/php-src#1105). This PR improves Yii2 to take advantage of that native function if it is available.


samdark:
We need to make sure implementations are compatible.

samdark:
But overall it looks OK to me. @tom-- do you see any issues?

tom--:
Yes, I expect that's fine. The existing unit tests should demonstrate compatibility.

No great hurry, I think. I didn't know it was [released](http://www.php.net/ChangeLog-7.php#7.1.2) until now you mentioned me here. (The discussion on internals only just finished. It was a nightmare. omg)

I'm going to be away until April. If someone else wants to have a go, that's fine with me. Better to use `function_exists()` than PHP version.

tom--:
Ignore me. It's already done! Thanks @charlesportwoodii 

I need a vacation sooo much.

samdark (approved):

tom--:
Is there evidence that new branch actually executes in the tests on 7.1.2?

samdark:
It isn't. Travis uses PHP 7.1.0.

tom--:
@samdark well, that's important.

But my question would apply if Travis had 7.1.2. It's one thing for the tests to pass. It's another to also know that the new branch does in fact execute.

cebe:
We can write a test that creates the function and asserts it is being called.

dynasource (approved):

SilverFire (approved):

samdark:
Verified locally and merged. Thank you!

PR #13660, opened 2017-02-25
Add a failing test for the URL rule

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | none

Confirms #13656.

### DIFF
--- a/tests/framework/web/UrlRuleTest.php
+++ b/tests/framework/web/UrlRuleTest.php
@@ -513,6 +513,20 @@ class UrlRuleTest extends TestCase
                 ],
             ],
             [
+                'single optional param with suffix',
+                [
+                    'pattern' => '<page>',
+                    'route' => 'page/view',
+                    'defaults' => ['page' => 'index'],
+                    'suffix' => '/',
+                ],
+                [
+                    ['page/view', [], '/'],
+                    ['page/view', ['page' => 'index'], '/'],
+                    ['page/view', ['page' => 'news'], 'news/'],
+                ],
+            ],
+            [
                 'required params',
                 [
                     'pattern' => 'about-me',


tests/framework/web/UrlRuleTest.php
```diff
@@ -513,6 +513,20 @@ protected function getTestsForCreateUrl()
...
             [
+                'single optional param with suffix',
+                [
+                    'pattern' => '<page>',
+                    'route' => 'page/view',
+                    'defaults' => ['page' => 'index'],
+                    'suffix' => '/',
+                ],
+                [
+                    ['page/view', [], '/'],
```
SilverFire: This assertion is incorrect. Only parameters with empty default values may be omitted.

tests/framework/web/UrlRuleTest.php
```diff
@@ -513,6 +513,20 @@ protected function getTestsForCreateUrl()
...
+                    ['page/view', ['page' => 'index'], '/'],
```
SilverFire: This one is incorrect as well. Suffix may be applied only to non-empty URL

tests/framework/web/UrlRuleTest.php
```diff
@@ -513,6 +513,20 @@ protected function getTestsForCreateUrl()
...
+                    ['page/view', ['page' => 'news'], 'news/'],
```
SilverFire: This case will be covered by https://github.com/yiisoft/yii2/pull/13624

SilverFire (changes requested):

tests/framework/web/UrlRuleTest.php
Kolyunya: @SilverFire so the expected result is `//`, right?

SilverFire: The correct one is empty string.
After #13624 merge the result will be correct
See https://github.com/yiisoft/yii2/pull/13624/commits/7206b2ba8a9e2a7648fa5507ce338fc851bc0a6e#diff-96fb0ef29c09f20871090c41bf5b3839R498

PR #13664, opened 2017-02-26
Refactor migration controller

Kolyunya:
Remove an excessive token from allowed migration names regular expression.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -557,7 +557,7 @@ abstract class BaseMigrateController extends Controller
      */
     public function actionCreate($name)
     {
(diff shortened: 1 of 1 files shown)


SilverFire:
Nope, it is not excessive.
`\` in group should be escaped with `\`, but `\` in `'` (single quotes) string should be escaped with `\` as well.
So we get escaped backslash (`\\`) escaped with backslash (`\\\\`)

Kolyunya:
@SilverFire thanks for the explanation!

PR #13667, opened 2017-02-27
added OWASP references to security guide

cebe:
currently security guide is very thrifty with information on what
a topic is about. So for beginners it is not clear why something like
CSRF or XSS protection is even needed.

I added a few reference links to allow reading about more background
behind the topics. 

@yiisoft/core-developers Please add more links if you have good references.


samdark (approved):

samdark:
Perfect.

SilverFire (approved):

Issue #13671, opened 2017-02-27
ReflectionException: Method yii\base\View::unknown() does not exist


yapi68:
Very hard to explain. But it seems there is a bug in yii2 debug info system.

This exception is raised when trying to get undefined variable in nested view. For example in yii2 basic app with YII_DEBUG sets in true you can change `contact.php` view file to render nested view `contact_form.php`. In that file try to get any undefined variable to raise an exception. Standard debug info will be "broken" - no color highlighting, strange rows in the end.

### What steps will reproduce the problem?
1. Install yii2-basic-app
2. Set YII_DEBUG to `true`
3. Change the view @app/views/site/contact.php to
```
<?php

/* @var $this yii\web\View */
/* @var $form yii\bootstrap\ActiveForm */
/* @var $model app\models\ContactForm */

use yii\helpers\Html;
use yii\bootstrap\ActiveForm;
use yii\captcha\Captcha;

$this->title = 'Contact';
$this->params['breadcrumbs'][] = $this->title;
?>

<?= $this->render("contact_form"); ?>
```
4. Create `contact_form.php` file:
```
<?php

/* @var $this yii\web\View */
/* @var $form yii\bootstrap\ActiveForm */
/* @var $model app\models\ContactForm */

use yii\helpers\Html;
use yii\bootstrap\ActiveForm;
use yii\captcha\Captcha;

$this->title = 'Contact';
$this->params['breadcrumbs'][] = $this->title;

$model->text; // <-- Exception will be throw here
?>
<div class="site-contact">
...
</div>
```
5. Navigate to `/index.php?r=site%2Fcontact` and see broken (not fully colored) debug view.

### What is the expected result?
Correct standard debug view.

### What do you get instead?
Broken (not fully colored) debug view.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2?
| PHP version      | 7.0.15
| Operating system | Debian 8.1.

This bug is appearing **ONLY** in nested views. If you try to call `$model->text` in main `contact.php` file nothing obvious happens. The debug view will be draw perfectly.

yapi68:
Screenshots of "broken" debug view:

![screen_1](https://cloud.githubusercontent.com/assets/1772714/23353414/67e6081e-fcdd-11e6-814f-b2379516c5d7.JPG)

[...]

PR #13673, opened 2017-02-27
Fix XDebug error handler

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13671


framework/base/Security.php
```diff
@@ -273,6 +273,14 @@ protected function decrypt($data, $passwordBased, $secret, $info)
      */
     public function hkdf($algo, $inputKey, $salt = null, $info = null, $length = 0)
     {
+        if (function_exists('hash_hkdf')) {
```
SilverFire: Why are these lines in diff?

SilverFire (changes requested):

SilverFire (approved):

dynasource (approved):
shouldnt we move formatting logic to an object for reuse?

samdark:
@dynasource where are we going to reuse it?

dynasource:
no idea ;) but all that logic in the view file does not feel right

samdark:
@dynasource done.

Issue #13530, opened 2017-02-06
Cacheable widget (behavior)


Kolyunya:
# Problem
Render a very heavy widget (with very expensive DB queries, rendering and calculations) in many views/layouts and cache it entirely as a single entity.

# TL;DR
We need a base cacheable widget.

# Currently possible solutions and pitfalls
## Fragment caching
* Business logic in views is always a bad idea:
```php
$dependency = [
    'class' => 'yii\caching\DbDependency',
    'sql' => 'SELECT MAX(updated_at) FROM post',
];

if ($this->beginCache($id, ['dependency' => $dependency])) {

    // ... our widget goes here ...

    $this->endCache();
}
```
* Code duplication & code complication. The code above can be refactored to store cache config as a widget class constant. This will partially remove business logic from view but we will have to copy-paste fragment cache (which is also a business logic itself) all over our project where this widget is rendered. 
```php
if ($this->beginCache($id, MyWidget::CACHE_CONFIG)) {

    // ... our widget goes here ...

    $this->endCache();
}
```
* Code complication. The code above can be further refactored via extracting it to a separate view fragment and including it everywhere which is not so elegant either.

### Data caching
* Code duplication & code complication. Having multiple widgets implementing this behavior we will have to copy-paste the following low-level caching code over and over again.
```php
public function run()
{
    $cache = Yii::$app->cache;
    $cacheKey = sprintf(
        'my-widget-%d-%d-%d-%d',
        $this->foo,
        $this->bar,
        $this->baz,
        $this->quux,
    );

    $widget = $cache->get($cacheKey);

    if ($widget === false) {
        $this->initialize();

        $widget = $this->render($this->template, [
            'foo' => $this->foo,
            'bar' => $this->bar,
            'baz' => $this->baz,
            'quuz' => $this->quux,
        ]);

        $cache->set($cacheKey, $widget, self::CACHE_DURATION);
    }
    return $widget;
 }
```
### Query caching
Not applicable since can't cache entire widget with template rendering and generic calculations.

# Proposed solution (**OUTDATED** see new proposal [here](https://github.com/yiisoft/yii2/pull/13679))
## General idea
We need to implement a base cacheable widget. We already have a `yii\widgets\FragmentCache` but it's impossible to use it conveniently in described scenario. The protected interface of the widget should have the following methods:
* `getCacheComponent()` - allow widget to define which cache component to use or leave empty to use default.
[...]

PR #13679, opened 2017-02-28
WIP: Implement cacheable behavior draft

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13530


Kolyunya:
Refactored behavior to make use the `yii\web\View::beginCache()` and `yii\web\View::endCache()` instead of the fragment cache.

framework/behaviors/CacheableBehavior.php
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+
+    /**
+     * Initializes a reference to owner widget view.
+     *
+     */
+    private function initializeView()
+    {
+        $this->_view = $this->owner->view;
```
SamMousa: Is there a reason to copy the reference?
This would be out of date if the owners' view is replaced.
Why not just take it from the owner whenever you need it?

framework/behaviors/CacheableBehavior.php
```diff
@@ -0,0 +1,188 @@
...
+use yii\web\View;
+
+/**
+ * Cacheable widget behavior.
+ *
+ * @property Widget $owner
+ *
+ * @group behaviors
+ */
+class CacheableBehavior extends Behavior
```
SamMousa: I think `CacheableWidgetBehavior` makes more sense.

framework/behaviors/CacheableBehavior.php
```diff
@@ -0,0 +1,188 @@
...
+        $this->_view->endCache();
+    }
+
+    /**
+     * Initializes cache instance.
+     *
+     */
+    private function initializeCache()
+    {
+        $this->_cache = Instance::ensure($this->cache, Cache::className());
```
SamMousa: The usual way of doing this in `Yii` currently is to just assign it to `$this->cache`.

framework/behaviors/CacheableBehavior.php
```diff
@@ -0,0 +1,188 @@
...
+        $this->initializeCache();
+        $this->initializeCacheKey();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    public function attach($owner)
+    {
+        parent::attach($owner);
```
SamMousa: Since you only support attaching to widgets, should that be checked here?
This is more a question for the yii team than for you; currently core behaviors do not check any for a valid owner.
I'll make a separate issue for that as well.

framework/behaviors/CacheableBehavior.php
```diff
@@ -0,0 +1,188 @@
...
+     * @var Cache|string|array
+     */
+    public $cache = 'cache';
+
+    /**
+     * Cache duration in seconds.
+     *
+     * @var int
+     */
+    public $cacheDuration = 60;
```
arogachev: If this is a cacheable behavior, we do expect that everything inside of this class will be related to cache, so I think it's better to remove "cache" prefixes from properties, for example instead of `$cacheKey` use just `$key` and so on.

framework/behaviors/CacheableBehavior.php
SamMousa: Related #13896.

Might not be important for this PR, can always be changed later together with the other core behaviors.

framework/behaviors/CacheableBehavior.php
```diff
@@ -0,0 +1,188 @@
...
+     */
+    private function initializeCacheKey()
+    {
+        if ($this->cacheKey !== null) {
+            $this->_cacheKey = $this->cacheKey;
+            return;
+        }
+
+        $this->_cacheKey = [
+            get_class($this->owner),
```
arogachev: This can be a one-liner.

framework/behaviors/CacheableBehavior.php
```diff
@@ -0,0 +1,188 @@
...
+    private function initializeCache()
+    {
+        $this->_cache = Instance::ensure($this->cache, Cache::className());
+    }
+
+    /**
+     * Initializes cache key.
+     *
+     */
+    private function initializeCacheKey()
```
SamMousa: I don't think the cache key itself should be "cached" in a private variable.
Why not use a getter and generate it each time it is needed? That way it is never out of date.

A widget might be rendered multiple times with different settings.

tests/framework/behaviors/CacheableBehaviorTest.php
```diff
@@ -0,0 +1,221 @@
...
+        $this->dynamicWidget = $this->getWidgetMock(DynamicCacheableWidget::className());
+    }
+
+    /**
+     * Returns a widget mock.
+     *
+     */
+    private function getWidgetMock($widgetClass)
+    {
+        $widgetMock = $this
```
arogachev: Is this a new style of writing chained method calls? Why not:

```php
$widgetMock = $this->getMockBuilder($widgetClass)
    ->setMethods(['run'])
    ->enableOriginalConstructor()
    ->enableProxyingToOriginalMethods()
    ->getMock();
```

This is more elegant and shorter IMO. Hovewer this is subjective of course, can't find it in [Core code style guide](https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md).

arogachev (changes requested):

framework/behaviors/CacheableBehavior.php
```diff
@@ -0,0 +1,188 @@
...
+    {
+        if ($this->cacheKey !== null) {
+            $this->_cacheKey = $this->cacheKey;
+            return;
+        }
+
+        $this->_cacheKey = [
+            get_class($this->owner),
+        ];
```
SamMousa: All widgets have an id available via `getId()`, I'd suggest using that in the cache key by default if it's set. 
````
[
    get_class($this->owner),
    $this->owner->getId(false)
]
````

File: framework/behaviors/CacheableBehavior.php (hunk 2 above)
Comment (Kolyunya) [inline framework/behaviors/CacheableBehavior.php]: This would be inconsistent with the `yii\behaviors\AttributeBehavior`. It's not `yii\behaviors\ActiveRecordAttributeBehavior`.

**Comment by SamMousa** [comment]
True. Except that most current behaviors are for AR; this one isn't.

File: tests/framework/behaviors/CacheableBehaviorTest.php (hunk 8 above)
Comment (Kolyunya) [inline tests/framework/behaviors/CacheableBehaviorTest.php]: Changed that. Still believe style is not the best possible.
1. It's inconsistent since methods are not aligned. Why do we make difference for the first method?
2. Placing semicolon on the same line with one of the methods is a bad idea: when we'd need to add another method at the end of the call chain, git diff will contain lines which were not effectively changed what will complicate the code-review process.

File: framework/behaviors/CacheableBehavior.php (hunk 6 above)
Comment (Kolyunya) [inline framework/behaviors/CacheableBehavior.php]: Fixed.

File: framework/behaviors/CacheableBehavior.php (hunk 5 above)
Comment (Kolyunya) [inline framework/behaviors/CacheableBehavior.php]: I'm not sure. That are incorrect variable names. `$this->duration` inside the behavior means `The duration of the behavior` which is incorrect since it's the duration of the cache. The `$this->dependency` means `The dependecy of the behavior` which is incorrect since it's the dependency of the cache.

File: framework/behaviors/CacheableBehavior.php (hunk 3 above)
Comment (Kolyunya) [inline framework/behaviors/CacheableBehavior.php]: `$this->cache` is already in use.

File: tests/framework/behaviors/CacheableBehaviorTest.php (hunk 8 above)
Comment (arogachev) [inline tests/framework/behaviors/CacheableBehaviorTest.php]: @Kolyunya As I said, this is very subjective. I expect that to be covered in the docs to prevent disputes and disagreements. Since there is no such info there let's wait for others contributors' opinions. You can leave your version if you think it's better.

File: framework/behaviors/CacheableBehavior.php (hunk 2 above)
Comment (SamMousa) [inline framework/behaviors/CacheableBehavior.php]: That's a fair argument; not sure consistency is more important than understandability.

> We don't put traffic lights on each intersection for the sake of consistency.

File: framework/behaviors/CacheableBehavior.php (hunk 5 above)
Comment (arogachev) [inline framework/behaviors/CacheableBehavior.php]: @Kolyunya Agree. Yes, some names are ambiguous, but maybe consider renaming at least some of them. Otherwise let's leave it with prefixes.

Comment (SamMousa) [inline framework/behaviors/CacheableBehavior.php]: Why change some of them? Which ones specifically would you like changed?

Comment (arogachev) [inline framework/behaviors/CacheableBehavior.php]: Never mind then, just ignore this remark. The first feeling of possible reduction was wrong.

**Review by samdark — CHANGES_REQUESTED** [review]

File: framework/behaviors/CacheableBehavior.php (hunk 2 above)
Comment (Kolyunya) [inline framework/behaviors/CacheableBehavior.php]: Renamed the behavior as suggested.

File: tests/framework/behaviors/CacheableBehaviorTest.php (hunk 8 above)
Comment (Kolyunya) [inline tests/framework/behaviors/CacheableBehaviorTest.php]: > Is this a new style of writing chained method calls?

@arogachev I just found out it's [not new](https://github.com/symfony/symfony-docs/blame/3.2/bundles/configuration.rst#L192). Just FYI.

**Comment by samdark** [comment]

**Review by dmirogin — APPROVED** [review]

**Comment by SamMousa** [comment]
I still don't like the stored references to the view component.

**Review by sergeymakinen — APPROVED** [review]

File: framework/behaviors/CacheableWidgetBehavior.php [hunk 10]
```diff
@@ -0,0 +1,182 @@
...
+        $this->_cache = Instance::ensure($this->cache, Cache::className());
+    }
+
+    /**
+     * Initializes cache key.
+     */
+    private function initializeCacheKey()
+    {
+        if ($this->cacheKey !== null) {
+            $this->_cacheKey = $this->cacheKey;
```
rob006: Why separate private property?

rob006 (changes requested):
The idea is OK, but implementation is really overcomplicated. We can get rid of all private properties and one-line private methods, and this class will lose half of its size.

tests/framework/behaviors/CacheableWidgetBehaviorTest.php
```diff
@@ -0,0 +1,213 @@
...
+
+    public function testDefaultKeyVariations()
+    {
+        $this->assertEquals([], $this->simpleWidget->cacheKeyVariations);
+    }
+
+    public function testDefaultEnabled()
+    {
+        $this->assertTrue($this->simpleWidget->cacheEnabled);
+    }
```
rob006: Do we really need test defaults?

tests/framework/behaviors/CacheableWidgetBehaviorTest.php
```diff
@@ -38,36 +38,6 @@ protected function setUp()
...
-    {
-        $this->assertNull($this->simpleWidget->cacheKey);
-    }
-
-    public function testDefaultKeyVariations()
-    {
-        $this->assertEquals([], $this->simpleWidget->cacheKeyVariations);
-    }
-
-    public function testDefaultEnabled()
```
Kolyunya: @developeruz why would you delete a test which ensures that the `cacheEnabled` property is actually initialized with `true` by default?

developeruz: Because on line 64 `public $cacheEnabled = true;` `cacheEnabled` set with `true` and imho here no needs to check if I haven't passed anything, that default value is still there.

Kolyunya: @developeruz following this logic we can delete the whole test-suite of the framework. Here is the code https://github.com/yiisoft/yii2, it makes the framework work.

This test covers line `64` you are referring to and ensures somebody does not remove the default value in future.

rob006 (approved):

tests/framework/behaviors/CacheableWidgetBehaviorTest.php
samdark: Removing test in this case is OK.

samdark:
I don't think we should merge it w/o any docs. @Kolyunya do you want to add some?

tests/framework/behaviors/CacheableWidgetBehaviorTest.php
developeruz: @Kolyunya, removing default value is violating backward compatibility, that is why no one won't remove the default value in future without being sure (will add a test in that time) that the same default value set in another way.
PS: sorry for intervention in your PR. I thought it is abandoned one and I just want to help this PR to be merged. Anyway, you did a great job 👍

Kolyunya:
@samdark I will write a brief documentation for the class.

samdark:
There's a deadline for 2.0.14: 10th this month. Without docs this PR won't make it into 2.0.

samdark (approved):

samdark:
Merged. Thank you!

Kolyunya:
Glad to help but I was actually refactoring this when you merged the branch. Do you think this should me merged too https://github.com/Kolyunya/yii2/commit/be6d8c300df7013e9f3a35f6e75d2563d2bc3c5c

Kolyunya:
@samdark glad to help but I was actually refactoring this when you merged the branch. Do you think this should me merged too https://github.com/Kolyunya/yii2/commit/be6d8c300df7013e9f3a35f6e75d2563d2bc3c5c

samdark:
@Kolyunya yes. Do another pull request and I'll merge it.

PR #13694, opened 2017-03-03
Send Pjax Header with response to fix redirect

wleona3:
Based on this issue: https://github.com/yiisoft/jquery-pjax/issues/44.
Note that $checkAjax must be set to false when calling Response::redirect()
Seemed to work for me.

### DIFF
--- a/framework/widgets/Pjax.php
+++ b/framework/widgets/Pjax.php
@@ -171,6 +171,11 @@ class Pjax extends Widget
         $response->setStatusCode(200);
         $response->format = Response::FORMAT_HTML;
         $response->content = $content;
+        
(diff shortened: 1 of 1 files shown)


framework/widgets/Pjax.php
```diff
@@ -171,6 +171,11 @@ public function run()
         $response->setStatusCode(200);
         $response->format = Response::FORMAT_HTML;
         $response->content = $content;
+        
+        if (!isset($response->headers['X-Pjax-Url'])) {
```
Faryshta: you can use http://www.yiiframework.com/doc-2.0/yii-web-headercollection.html#setDefault()-detail and make it one line only

PR #13703, opened 2017-03-06
Improve string validator test coverage

### LINKED ISSUE #13701: StringValidator not failing when sent a true boolean type.
Establish a `'string'` rule on an attribute in a Model (ActiveRecord in my case).

POST a JSON object containing a `true` for that attribute.

Similarly to how POSTing a `false` to that attribute will generate an error about the attribute being empty, I would expect the same for a `true`

### What do you get instead?
The attribute received a `'1'` as its new value.

### Additional info
I verified the `getBodyParam('attribute')` is returning the `boolean` however I noticed in `StringValidator.validateAttribute()` that the `$value` is a `'1'` at that point which is why the attribute is not failing the validation.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.0.13
| Operating system | ubuntu 4.9.5-moby

SilverFire:
Closes #13701

### DIFF
--- a/tests/framework/validators/StringValidatorTest.php
+++ b/tests/framework/validators/StringValidatorTest.php
@@ -22,6 +22,8 @@ class StringValidatorTest extends TestCase
         $val = new StringValidator();
         $this->assertFalse($val->validate(['not a string']));
         $this->assertTrue($val->validate('Just some string'));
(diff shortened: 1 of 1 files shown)


SilverFire:
Thank you @Kolyunya

PR #13706, opened 2017-03-06
Extract profiling methods from BaseYii to standalone Profiler class

cronfy:
I extracted `BaseYii::beginProfile()` and `BaseYii::endProfile()` to standalone `yii\base\Profiler` class as was described at https://github.com/yiisoft/yii2/issues/9458#issuecomment-282320423

Now profiler is a core component and can be configured (replaced) via `components` application configuration:

```php
'components' => [
    'profiler' => [
        'class' => `\my\own\Profiler`,
    ],
    ...
]
```

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #9458


cronfy:
This PR directly fixes #9458 (Profiler can be replaced with own implementation to log duration).

Also https://github.com/yiisoft/yii2-debug/issues/81 can be fixed if we introduce third `$context` parameter to `Yii::beginProfile()`:

```
    public static function beginProfile($token, $category = 'application', $context = null)
    {
        if (static::$app !== null) {
            return static::$app->getProfiler()->begin($token, $category, $context);
        }
        ...
    }
```

It will allow to pass arbitrary data to Profiler, allowing Profiler to be more informative.

cronfy:
What do I do with codeclimate?

> The class Application has 45 public methods and attributes. Consider reducing the number of public items to less than 45

I added `Application->getProfiler()` public method, similarly to other `get...()` methods for core components.

samdark:
Ignore it :0

framework/base/Application.php
```diff
@@ -615,6 +615,15 @@ public function getSecurity()
     }
 
     /**
+     * Returns the profiler component.
+     * @return \yii\base\Profiler the profiler application component.
```
samdark: Need `@since` tag.

framework/base/Profiler.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace yii\base;
+
+use Yii;
+use yii\log\Logger;
+
+/**
+ * Profiler is the class for application profiling.
+ *
+ * It is a basic implementation of profiling class. It has [[begin()]] and
```
samdark: This paragraph could be removed since it repeats previous one and describes interface that we already have in the code.

framework/base/Profiler.php
```diff
@@ -0,0 +1,57 @@
...
+ * [[end()]] methods to start and stop profiling of a code block.
+ *
+ * Profiler is used as a core component of Yii application.
+ *
+ * @see [[Yii::beginProfile()]]
```
samdark: Reference to `Yii::endProfile()` should be added as well.

framework/base/Profiler.php
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @author cronfy <cronfy@gmail.com>
+ * @author Qiang Xue <qiang.xue@gmail.com>
+ * @since 2.0
```
samdark: 2.0.12

samdark (changes requested):
As commented. Also need a changelog line.

framework/BaseYii.php
```diff
@@ -441,12 +441,29 @@ public static function info($message, $category = 'application')
...
+     * ```
+     *
      * @param string $token token for the code block
      * @param string $category the category of this log message
      * @see endProfile()
      */
     public static function beginProfile($token, $category = 'application')
     {
+        if (static::$app !== null) {
+            return static::$app->getProfiler()->begin($token, $category);
```
rob006: Please, don't use such return syntax for methods that returns void - it is really confusing.

samdark (approved):

klimov-paul:
I can see no benefit from such change: `Yii::beginProfile()` and `Yii::endProfile()` are already just a wrappers around `Logger` methods.

> Then everyone will be able to replace profiler with his own implementation.

It is already possible to replace standard `Logger` with your own implementation. What the sense of introduction of another abstraction layer here?

cronfy:
@klimov-paul Logger and Profiler are logically different.

While it is possible to replace Logger with own implementation, it is hard to tell in `Logger::log()` whether  it was intended to profile some block of code or just log a line of text. Only hacks depending on `$message`/`$level`/`$category` are available, which look ugly for me.

Profiler can even not use logging at all. It can decide to store profiling data in it's own storage and update it in many other ways differrent from just saving a line of text in a file/db.

If Logger would be used for that, one could be surprised that `->log()` did not save data to a target.

I would like also to introduce third parameter to `beginProfile()`:

```php
    public static function beginProfile($token, $category = 'application', $context = null)
    {
        if (static::$app !== null) {
            static::$app->getProfiler()->begin($token, $category, $context);
        }
        ...
```

to give Profiler even more data, which could solve issues like https://github.com/yiisoft/yii2-debug/issues/81

I just though  it would be better to make a separate PR for that.

klimov-paul:
> it is hard to tell in Logger::log() whether it was intended to profile some block of code or just log a line of text.

???
What is wrong with check:

```php
if ($level === Logger::LEVEL_PROFILE_BEGIN) {
   // begin profile
}
if ($level === Logger::LEVEL_PROFILE_BEGIN) {
   // begin profile
}
if ($level === Logger::LEVEL_PROFILE_END) {
   // end profile
}
```

klimov-paul:
> give Profiler even more data, which could solve issues like yiisoft/yii2-debug#81

Profiling servers the time-measure purpose, saving object state at different program stages is a logger taks.
Following your own words:

> Logger and Profiler are logically different

those can not be mixed up.

cronfy:
> What is wrong with check:
> if ($level === Logger::LEVEL_PROFILE_BEGIN) {
>   // begin profile
> }

... looks like `$logger->log($message, PLEASE_DO_NOT_LOG_CALL_PROFILER_INSTEAD, $category)`.

Wrong, because this is a hack. It introduces Special_Magic_Log_Level_That_Triggers_Profiling. Don't think about `Yii::beginProfile()` as of logging method similar to `::info()` or `::warning()`. Perhaps it was created with this logic in mind, but then it is just THE thing I would like to change with this PR.

As I said, Profiler and Logger are just different things indeed. Profiler has it's own logic and may not use logging at all. Why call `Logger->log()` if it was not intended to log a message?

Oftenly profiling ends up with saving several [thouthands] lines of logs about program execution, but this is not the best way to profile an application. Profiler should gather data and provide statistics, not just timeline of code execution.

Of course, it is behind of what core component should provide from the box, but this PR gives a way to extend Profiler functionality in a Yii way (via installing packages and configuration) and without hacks. Adding standalone Profiler class (and passing it a `$context` to `beginProfile()` call as I told earlier, if will be implemented) gives a lot of opportunities to gather data and measure application execution time/memory/etc in various aspects, which are blocked now, because `beginProfile()` and `endProfile()` are hardcoded to use Logger and it can't be changed with configuration.

Of course, I could use own Logger or extend `Yii` class (that's what I am doing now), but... I think neither one is a good way.

klimov-paul:
> Wrong, because this is a hack. It introduces Special_Magic_Log_Level_That_Triggers_Profiling. Don't think about Yii::beginProfile() as of logging method similar to ::info() or ::warning(). Perhaps it was created with this logic in mind, but then it is just THE thing I would like to change with this PR.

???
Shouldn't we introduce such classes as `Tracer`, `Informer` and so on to get rid of `Logger` constants usage?

klimov-paul:
> As I said, Profiler and Logger are just different things indeed.

I can not agree. Both profiling and logging are the tasks, which saves information about program execution flow, so it can viewed elsewhere. They both require some kind of persistent storage - otherwise they can not function.

I would rather introduce new methods at `Logger`: `Logger::beginProfile()` and `Logger::endProfile()` instead of creating full-scale `Profiler` class, but even this is questionable.

For me creaing extra class instroduced at this PR is redundant, as use case can be easily solved simply by extending `Logger`.

framework/base/Application.php
```diff
@@ -615,6 +616,16 @@ public function getSecurity()
...
+     * @since 2.0.12
+     */
+    public function getProfiler()
+    {
+        return $this->get('profiler');
```
klimov-paul: `profiler` should not belong to `Application` - it should be hold as static field at `Yii` class in the same way as `Logger`.

klimov-paul (changes requested):

cronfy:
> I would rather introduce new methods at Logger: Logger::beginProfile() and Logger::endProfile() instead of creating full-scale Profiler class, but even this is questionable.

Please further explain what you suggest. Should `Logger::beginProfile()` be called from `Yii::beginProfile()`? Or `Yii:;beginProfile()` should be left as is, and `Logger::beginProfile()` will be used for some other cases?

klimov-paul:
> Should Logger::beginProfile() be called from Yii::beginProfile()?

Yes.

cronfy:
> Shouldn't we introduce such classes as Tracer, Informer and so on to get rid of Logger constants usage?

Absolutely not, of course. If it was required to make profiling calls consistent with other logging calls, then it is enough to remove `::beginProfile()` and `::endProfile()` methods and add simple `::profile()` method.

The specificity of profile methods is just that they are required to work in pair, not as other logging methods. That's why they can be extracted to another layer of abstraction, not related to Logger.

> Should Logger::beginProfile() be called from Yii::beginProfile()?
> Yes.

I see your point of view, I can't agree, but I do not know what to add. You tell that Logger and Profiler is the same thing, I'm still sure they are different (because logging is about saving a single message, and profiler is about measuring execution time/memory/etc; profiler **can** use logger, but logger can even do not know that profiler exists).

However your suggestion about extending Logger solves the issue too, but I'd rather file another PR for it.

Anyway I would appreciate other opinions to get an idea of a best way to solve the issue.

cronfy:
Further exploring Logger class I've found following mehtods: [calculateTimings()](http://www.yiiframework.com/doc-2.0/yii-log-logger.html#calculateTimings()-detail), [getProfiling()](http://www.yiiframework.com/doc-2.0/yii-log-logger.html#getProfiling()-detail) and [getDbProfiling()](http://www.yiiframework.com/doc-2.0/yii-log-logger.html#getDbProfiling()-detail).

I can't find usages of these methods in yii2-framework, yii2-app-basic or yii2-app-advanced, except yiisoft/yii2-debug extension. They also aren't mentioned in the docs.

I wonder if these methods specific to yiisoft/yii2-debug extension (and debug panels in extensions like yiisoft/yii2-mongodb)?

bashkarev:
I support moving methods to the debugger panel, but in 2.1 since this will break almost all the panels

samdark:
@bashkarev why? Here methods are aliased, not actually moved.

bashkarev:
I meant the comment https://github.com/yiisoft/yii2/pull/13706#issuecomment-284662335
It would be nice to move the methods: ` calculateTimings`, `getProfiling` `getDbProfiling` to yii2-debug

samdark:
Ah, yes. These are, indeed, worth moving to debug in 2.1.

framework/base/Application.php
cronfy: How then `profiler` can be configured, if it is not an Application component?

samdark:
I dug deeper into logging and while re-checking Monolog and PSR-3 I came to conclusion that profiler indeed should be a separate component which uses logger for logging, not the part of the logging itself.

samdark:
So all the messages of at least profile_begin and profile_end levels should be, in fact, internal Profiler class business...

samdark:
Yes, such syntax sounds good.

btw., here's a post about PSR-3 + Yii: https://samdark.github.io/blog/yii-2.0-logging-and-psr-3/

bashkarev:
Why?
```php
class CommandProfile extends Profile
{

    /**
     * @param \yii\db\Command $options
     * @param string $category
     * @return string
     */
    public static function begin($options = null, $category = 'application')
    {
        return parent::begin($options->getSql(), $category);
    }
}
```

samdark:
OK, this way it would be fine.

samdark:
Anyway, for 2.0 we can't really change things much so at least old API should work the same as it works now. For 2.1 it's possible to change things significantly.

rob006:
@bashkarev But this will require change all `Profile::` calls to `CommandProfile::` in all your code. And you can't really check code of dependencies, so some part of your app will still use standard `Profiles` class.

bashkarev:
```php
interface Profile
{

    /**
     * @param mixed $options
     * @param string $category
     * @return static
     */
    public static function begin($options = null, $category = 'application');

    /**
     * @param mixed $data
     * @return bool
     */
    public function withData($data);

    /**
     * @return bool
     */
    public function end();
}
```
?

cronfy:
1. I think profiling functions in Yii2 core should be really simple and straightforward, allowing to mark the beginning and the end of a block being profiled.

    Yii2 already has API for this functionality: it's `Yii::beginProfile()` and `Yii::endProfile` methods. 

2. It's great that profiling is built in `yii\db\Command` and `yii\db\Connection`, because it allows general profiling from the box.

3. I think in production profiling calls should not affect application at all, until it is explicitly enabled. It is currently **not true** in Yii2, because profiling messages are stored in '$logger->messages` independently of whether profiling is enabled or not (even more - Yii does not have any option to enable/disable profiling). These profiling data takes up memory always.

    In other words, Yii core should not collect and store profiling data by itself.

4. Any extension interested in gathering profiling data (e. g. `yii2-debug`) should be able to listen to profiling calls and recieve profiling `token` and other context, if available. It should be possible for many listeners to exist and not interfere with each other.

With this thoughts in mind, Profiler can be implemented:

 * using Dispatcher -> Targets pattern, similarly to what we currently have in Logger. 
 * using Events. Simpler than Dispatcher, but requires more overhead, since `new \yii\base\Event()` is required for every Profiler call.

What do you think?

bashkarev:
For me, `events` are better, example:
https://gist.github.com/bashkarev/3c74f7a1e966f2329800797d79a8fd18

samdark:
Events are quite logical for the purpose.

cronfy:
I added events to Profiler. Now Logger subscribes to Profiler events in `init()`, no more direct calls to Logger from Profiler. Any other listener can subscribe to Profiler events as follows:

```
Yii::$app->on(Profiler::EVENT_BEGIN, function(ProfilerEvent $event) {
   // $event->token, $event->category and $event->context can be used
});
Yii::$app->on(Profiler::EVENT_END, function(ProfilerEvent $event) {
   // $event->token, $event->category and $event->context can be used here too
});
```

I also added `$context` argument to `Yii::beginProfile`, it allows to pass arbitrary data down to listeners, e.g.:

```
// somwhere in a listener
Yii::$app->on(Profiler::EVENT_BEGIN, function(ProfilerEvent $event) {
    $query = $event->context['query'];
    Yii::info("Query: " . $query->createCommand()->rawSql);
});


// somewhere in a controller
$query = Model::find();
$token = "my heavy query";
$context = ['query' => $query];
Yii::beginProfile($token, __METHOD__, $context);
$data = $query->all();
Yii::endProfile($token, __METHOD__, $context);
```

framework/base/Profiler.php
```diff
@@ -0,0 +1,68 @@
...
+     * The begin- and end- calls must also be properly nested.
+     *
+     * @param string $token token for the code block
+     * @param string $category the category of this profiler message (used for logging)
+     * @param mixed $context arbitrary data that can be used by profiler event listeners
+     * @see end()
+     */
+    public function begin($token, $category = 'application', $context = null)
+    {
+        $event = new ProfilerEvent(compact('token', 'category', 'context'));
```
bashkarev: Why `compact`?
https://3v4l.org/WTrOJ

cronfy: Other changes are ok?

P.s. 0.03s difference for 100k iterations. Personally I can live with that. Are there any Yii2 guidelines about `compact` and other performance optimizations?

bashkarev: Other changes, I like.

samdark:
Would be good to measure execution performance difference on real app in debug mode and normal mode.

rob006:
What is the point of having separate profiler component, if its only job is triggering events?

dynasource:
agree. In its current form, I dont see the added value yet.

cronfy:
Profiler is a simple component.

The point it is separate component is that if one wants, he can replace it with own class, e.g. to disable it's events completely.

rob006:
I don't really like current event-based approach. What is the benefit of using events for this? Attaching multiple handlers for handling profiler event will give you unreliable results. 
For me it looks like overcomplicated and overblotated solution for simple thing.

samdark:
@rob006 Profiling doesn't really belong to Logger. How would you solve it?

rob006:
I'm not against separate component for profiling, but I don't think that using events in this way will add any value for profiling. 

I prefer solution similar to logger - dispatcher with configured handlers for handling profiling events. In this case we can handle db profiling differently than others, and still keep this simple. Each profiling event is handled by one handler object, so we have better encapsulation. If there is no handlers, profiling is practically disabled.

Simple draft:

```php
class Profiler extends Component {

    // handlers config
    public $handlers = [];

    private $_stack = [];

    public function begin($token, $category = 'application') {
         $handler = $this->createHandler($category);
         if ($handler !== null) {
             $this->_stack[$token] = $handler;
             $handler->begin();
         }

         return $handler;
    }

    public function end($token, $category = 'application') {
         if (isset($this->_stack[$token]) {
             $this->_stack[$token]->end();
             unset($this->_stack[$token]);
         }
   }

    protected function createHandler($category) {
         // returns handler object for profiling event
    }
}
```

cronfy:
@rob006 why invent one more dispatching mechanism if there are already Events in the core?

rob006:
To make profiling faster, more convenient and reliable.

cronfy:
Why do you think with events profiling is not fast, convenient and reliable enough?

rob006:
It is not reliable, because you can add 2 handlers for profiling event. In this case it works in this way:

1. Start event A
2. Start event B
3. Do stuff that you want to profile
4. End event A
5. End event B

So you want profile only 3, but result of event A is affected by 2 and event B is affected by 4. You have two events, but none of them shows real results. 

It is not fast because you create 2 objects for every profiling, even if profiling is not needed.

It is not convenient because custom handling of profiling is still PITA. Just try save profiling results to logs, but as one entry (instead two) with information about time elapsed during profiling. 
Right now you can't even disable logging  profiling events without overriding whole component.

bashkarev:
Benchmark (50000) best result:

|current log|one event|two events|
|---|---|---|
|2.3577499389648|0.51886796951294|0.6534309387207|

[Details here](https://github.com/bashkarev/profiler-benchmark)

cronfy:
I am really glad we have this discussion, because it emphasizes that there are different requirements about profiling, and thus abiilty to extend/replace it is in demand indeed.

But the internals of profiler may require longer discussion, which has no meaning if Profiler is not extracted to a component.

Can we please get back to original PR, where Profiler was just converted to a component without any logic changes? https://github.com/yiisoft/yii2/pull/13706/files/029a33b4f6a09cac66f516309389102f97677b2f

After it is merged, we can further discuss proper Profiler implementation in subsequent PRs.

What's blocking this?

samdark:
@cronfy nothing.

cronfy:
I reverted changes about events. Now this PR is only about moving Profiler to a component without logic changes.

framework/base/Profiler.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Marks the beginning of a code block for profiling.
+     * This has to be matched with a call to [[end()]] with the same category name.
+     * The begin- and end- calls must also be properly nested.
+     *
+     * @param string $token token for the code block
+     * @param string $category the category of this profiler message (used for logging)
+     * @see end()
+     */
+    public function begin($token, $category = 'application')
```
rob006: We can't change signature of `Yii::beginProfile()`, but this is new method, we could add `$context` param at least here.

cronfy: I  doubt it worth adding unused variable in this PR...

rob006: Why unused? `$context` should be logged. And this is public method of public component, people may always use `Yii::$app->logger->begin('my-code', 'application', $context)`. Not mention that changing that in Yii 2.1 will create unnecessary BC break. 

And personally I don't really see much sense of releasing this in current state. You're creating API, that can't be changed easily in the future. We should decide how this should work in first place - this is not so urgent issue, there is no need for rush and temporary solution.

bashkarev: I support @rob006, in the current implementation does not bring much benefit, propose to discuss the implementation, and add in version 2.1

samdark: @rob006, @bashkarev any thoughts about how should it work?

rob006: @samdark I'm not sure about what you're asking. I already proposed how profiler could work here: https://github.com/yiisoft/yii2/pull/13706#issuecomment-287567082 .

bashkarev: @samdark 
1. Profiler and logger must be separated completely
2. Work only in debug mode, it's a Profiler and not a logger
3. Generating a unique token must be the responsibility of Profiler

cronfy: @bashkarev 

1. I agree.
2. I don't agree: Profiler should have enable/disable option instead, not related to debug mode. Consider this use case: you have website that by some reason gives huge load on database **sometimes**. This load does not reproduce in dev environment. Solution is to enable profiler on production and attach a collector to it that will analyze database queries and report statistics about it. Same with CPU load, memory usage.
3. I agree.

bashkarev: There is a big problem, people forget to turn off even the debug panel.

```php
public function init()
{
    if($this->enable === null){
        $this->enable = !YII_DEBUG;
    }
}
```

samdark: 1. Agree.
2. `enabled` should be configurable and set to `YII_DEBUG` by default.
3. Yes.

klimov-paul:
Resolved via #14611

PR #13708, opened 2017-03-06
Fixes #13707: Fixed `\yii\web\ErrorHandler` not setting correct respo…

### LINKED ISSUE #13707: In Default ErrorView render by `yii\web\ErrorAction` ，the `Yii::$app->response->statusCode` is not correct
use  `Composer` to initialize  a clean Yii2 advanced app template
in `frontend/views/site/error.php`   add the following line
```php
<?= Yii::$app->response->statusCode?>
```
then, navigate to `a none-existing Url Location` in Browser,such as `http://yaa-frontend.dev/xxx`

the above PHP code added to   `frontend/views/site/error.php`    should output `404`

### What do you get instead?
the  above PHP code added to   `frontend/views/site/error.php`    just output `200`  in page content
 **_this is not correct_**

**Note:** HTTP code in Browser is 404, **_this is correct_**.
image

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | PHP 7.1
| Operating system | Ubuntu 16.04 LTS

samdark:
…nse code to response object before rendering error view

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #13707

### DIFF
--- a/framework/web/ErrorAction.php
+++ b/framework/web/ErrorAction.php
@@ -98,12 +98,7 @@ class ErrorAction extends Action
      */
     public function run()
     {
-        $response = Yii::$app->getResponse();
-        if ($this->exception instanceof HttpException) {
-            $response->setStatusCode($this->exception->statusCode);
-        } else {
-            $response->setStatusCode(500);
-        }
+        Yii::$app->getResponse()->setStatusCodeByException($this->exception);
 
         if (Yii::$app->getRequest()->getIsAjax()) {
             return $this->renderAjaxResponse();
--- a/framework/web/ErrorAction.php
+++ b/framework/web/ErrorAction.php
@@ -98,6 +98,13 @@ class ErrorAction extends Action
      */
     public function run()
     {
+        $response = Yii::$app->getResponse();
+        if ($this->exception instanceof HttpException) {
+            $response->setStatusCode($this->exception->statusCode);
+        } else {
+            $response->setStatusCode(500);
+        }
+
         if (Yii::$app->getRequest()->getIsAjax()) {
             return $this->renderAjaxResponse();
         }
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -42,7 +42,7 @@ Yii Framework 2 Change Log
 - Bug #13670: Fixed alias option from console when it includes `-` or `_` in option name (pana1990)
 - Enh: Added `yii\di\Instance::__set_state()` method to restore object after serialization using `var_export()` function (silvefire)
 - Enh #13695: `\yii\web\Response::setStatusCode()` method now returns the Response object itself (kyle-mccarthy)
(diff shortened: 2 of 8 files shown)


samdark:
Right. Tests...

ihipop:
@samdark  I‘m interested in  how to write this unit test
should this test  run against `ErrorAction` or `ErrorHandler` ?

if run against `ErrorHandler` should it right below?
```php
 public function testRenderException()
    {
        Yii::$app->getErrorHandler()->handleException(new NotFoundHttpException('this is a 404 error'))
        $this->assertEquals(404,Yii::$app->response->statusCode);
    }
```

dynasource:
@ihipop, ideally you should run it as isolated a possible without Yii::$app against just `ErrorHandler`

dynasource (changes requested):
unit test required

ihipop:
@samdark

samdark:
@ihipop in this case test is for `renderException()` method only. You've got that part right. The thing that is tricker is that your test would pass with both old and new code. What we should test here is that code is set before rendering an exception view.

ihipop:
@samdark so,the `test` should be written against `ErrorAction` ?
I've no Idea  ：（

samdark:
@dynasource unit test added.

framework/web/ErrorAction.php
```diff
@@ -98,6 +98,13 @@ public function init()
      */
     public function run()
     {
+        $response = Yii::$app->getResponse();
+        if ($this->exception instanceof HttpException) {
```
samdark: Not sure this is actually needed...

SilverFire: It's fine IMO

SilverFire (approved):

framework/web/ErrorAction.php
dynasource: why is this done twice in this PR?

samdark: Because in case of ErrorAction, there's a way to trigger it w/ exception without actually going through error handler.

dynasource: what about handling this in Response itself by passing the exception

samdark: Which exception?

dynasource: to fix the duplication of logic in:
* https://github.com/yiisoft/yii2/pull/13708/files#diff-f4cd9aa0a2255e5362d10b956fe70974R102
* https://github.com/yiisoft/yii2/pull/13708/files#diff-7a2fbae94c890a0d999b38f8a82fe81eR92

samdark: Yes. I know that. Check tests. In case of action, error handler isn't involved and it's public API so same could be done in real projects. That's why a safeguard.

dynasource (approved):

Issue #13702, opened 2017-03-05
Proposal: Have you thought about using monolog as logger?


antonmarin:
[Monolog](https://github.com/Seldaek/monolog) used by many frameworks. Have you thought about using monolog instead own logger implementation?

P.S. I know about yii2-monolog extension. I mean core usage

samdark:
I'm OK with replacing some implementations with better components. I haven't really compared Monolog to our implementation in detail. What are cons and pros?

antonmarin:
Pros:
- PSR3
- [Many handlers] (https://github.com/Seldaek/monolog/tree/master/src/Monolog/Handler) 
- large community (49M installs)

Cons:
- PSR3 (differ interface). I think it's resolvable using Yii::info() shortcuts
- no alter to categories mechanics
- no exclude

samdark:
Changing interface is possible in 2.1 so PSR-3 could be excluded from the list of cons, I think.

How about performance? Logging is very central application thing so if Monolog is significantly slower, it would be seen immediately.

antonmarin:
Well, monolog logger is just an observer that throws messages to handlers. So it is closer to yii target then yii logger. I think it is suitable to use same dispatcher with some additions doing this we can achieve same performance and increase count of handlers.
Will be like:
- user call Dispatcher->log()
- Dispatcher collects messages
- application call flush
- Dispatcher send messages to logger

samdark:
Sounds interesting to try to me. @yiisoft/core-developers thoughts?

klimov-paul:
Support of `Monolog` is a matter of PSR-3 support. There was an issue about this already - #1676.

klimov-paul:
Also there is related issue #7555.
It actually may happen some extension for Monolog support already exists.

antonmarin:
As i said earlier, there is extension yii2-monolog. We are talking about core usage

klimov-paul:
> We are talking about core usage and profit of many handlers already implemented

Following this path, inside the core there should be support for PSR-3 compatible logger. Then `monolog` can be used as a particular one.
Including `monolog` as a a default log processor will create an external dependency, which contradicts #8452 .

samdark:
I think that would do. Overall Yii logging is implemented well so there's no intent dropping it in favor of anything to save maintenance cost.

samdark:
I'll leave it open since I want to try creating PSR-3 driver for our logger.

samdark:
[...]

PR #13711, opened 2017-03-06
Draft implementation of PSR-3 log target

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #13702

### DIFF
--- a/framework/log/PsrTarget.php
+++ b/framework/log/PsrTarget.php
@@ -1,0 +1,85 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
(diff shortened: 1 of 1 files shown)


rob006 (changes requested):

rob006:
I think this should be distributed as separate extension. You're using `psr/log` package here, but it is not mentioned in composer require section, so basically  you're using non-existent class. In separate extension you could add `psr/log` into composer require, so composer will take care about required dependencies and possible conflicts.

samdark:
Yes. Indeed it's better to be done as separate extension. I'll probably release it under `samdark/` and mention it in the guide.

Issue #13728, opened 2017-03-08
SluggableBehavior immutable option with Anonymous function fails


dbd5:
### What steps will reproduce the problem?
1. Add the SluggableBehavior to a class
2. Set the value attribute to an annonymous function
3. set 'immutable' => true for the slug
4. Create a new record (RESULT Slug is generated correctly by the anonymous function)
5. Update the record created in (4)
 
### What is the expected result?
The old slug should stay unchanged, the anonymous function should not be called. The slug should not be replaced.

### What do you get instead?
A new slug is generated even though immutable is set to true

### Additional info
This is similar to #6430 except that an anonymous function was not used in that issue.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 5.6.30
| Operating system | Ubuntu

Kolyunya:
Confirmed (https://github.com/Kolyunya/yii2/commit/c7c3ef93141db7a190b53db2640e711f1aeba7e9).

PR #13729, opened 2017-03-08
Fix immutable sluggable behavior

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13728

`isNewSlugNeeded()` wasn't actually checked when `attribute` was equal to `null`.


samdark (approved):

SilverFire (approved):

framework/behaviors/SluggableBehavior.php
```diff
@@ -171,6 +171,10 @@ protected function isNewSlugNeeded()
             return false;
         }
 
+        if ($this->attribute === null) {
```
dynasource: this feels a bit magical and is taking too much time to understand. Can you support this with a comment to explain why true is to be returned when attribute is null?

dynasource (changes requested):

framework/behaviors/SluggableBehavior.php
Kolyunya: Wouldn't [this](https://github.com/yiisoft/yii2/pull/13747) be sufficient?

Kolyunya: Sorry I missed the line.

Kolyunya: Seems like we can optimize this condition for the case when `$value` is a string. We can compare it to the current slug value. Do you think it worth implementing?

For the case when `$value` is `callable` the comment would be:
> Callback-based non-immutable slugs should be recalculated each time.

dynasource: well, the current state of SluggableBehavior is too vague IMO, so every step to make it more readable is a step in the right direction.

Kolyunya: Should we add another condition for the case when `$value` is a string or just recalculate all value-based slugs?

dynasource: what would be the usecase for that?

Kolyunya: When the slug is string-based and it's value is equal to the `$value` we can return `false` and slug value won't be recalculated.

dynasource: if this condition had to be covered by a method with a proper self declaring name, how would you call this method?

Kolyunya: Do you think we should move this condition to a separate method?

I added a WIP comment. The problem is that it tells that `callback-based` slugs are always reevaluated but in fact all `value-based` slugs are always reevaluated at the moment.

dynasource: Im trying to get concrete answers to my questions to get a proper understanding ;)

Kolyunya: Could you please ask a *self-contained* concrete question please? What particularly don't you understand?

dynasource: We want to know 'what' the software does and not 'how' he does it. 

I tried to get to this, but I could not get clear answers to my questions. Examples:

* My question "Can you support this with a comment to explain why true is to be returned when attribute is null" in https://github.com/yiisoft/yii2/pull/13729#discussion_r106017513 gets a vague implementation reaction in https://github.com/yiisoft/yii2/pull/13729#discussion_r106027562
* My question "what would be the usecase for that" gets an answer about implemantation details in https://github.com/yiisoft/yii2/pull/13729#discussion_r106033797 but not about the "why" which was my question
* My last question to understand the "why" https://github.com/yiisoft/yii2/pull/13729#discussion_r106035189 was about trying to get you to formulate the usecase into a understandable testMethod, but again I dont get an corresponding answer to my question in https://github.com/yiisoft/yii2/pull/13729#discussion_r106101040

To go back to the code. My problem is to understand the logic behind `$this->attribute === null`. What is the meaning of it. Why is it added? By purely looking at the code, I have to put effort into understand it. This goes against the best practice that code should be self-explanatory and instantly understandable. 

So again. We want to know 'what' the software does and not 'how' he does it.

Kolyunya: > Can you support this with a comment to explain why true is to be returned when attribute is null?
Yes I can. And I did it here: https://github.com/yiisoft/yii2/pull/13729/files#r106027562

> What would be the usecase for that?
String-based slugs.

> What is the meaning of it. Why is it added?
It seems to be obvious for me. It's not possible to tell if new slug is needed or not for callback-based slugs. That is why a new value is always needed for such slugs.

samdark:
Merged. Thank you!

Kolyunya:
Glad to help!

PR #13732, opened 2017-03-08
WIP: Add code-style validator to Travis CI config

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

I'd like to suggest to make use of a code-style validator during `Travis CI` checks to improve the general quality of the framework code and to prevent `PSR-2` violations.

### DIFF
--- a/framework/views/errorHandler/exception.php
+++ b/framework/views/errorHandler/exception.php
@@ -381,8 +381,14 @@ body.mousedown pre {
         <ul>
             <?= $handler->renderCallStackItem($exception->getFile(), $exception->getLine(), null, null, [], 1) ?>
             <?php for ($i = 0, $trace = $exception->getTrace(), $length = count($trace); $i < $length; ++$i): ?>
-                <?= $handler->renderCallStackItem(@$trace[$i]['file'] ?: null, @$trace[$i]['line'] ?: null,
-                    @$trace[$i]['class'] ?: null, @$trace[$i]['function'] ?: null, @$trace[$i]['args'] ?: [], $i + 2) ?>
+                <?= $handler->renderCallStackItem(
+                    @$trace[$i]['file'] ?: null,
+                    @$trace[$i]['line'] ?: null,
+                    @$trace[$i]['class'] ?: null,
+                    @$trace[$i]['function'] ?: null,
+                    @$trace[$i]['args'] ?: [],
+                    $i + 2
+                ) ?>
             <?php endfor; ?>
         </ul>
     </div>
--- a/.travis.yml
+++ b/.travis.yml
@@ -156,6 +156,8 @@ script:
   - cd framework && composer validate --no-check-lock && cd ..
   # run PHPUnit
   - vendor/bin/phpunit --verbose $PHPUNIT_FLAGS --exclude-group mssql,oci,wincache,xcache,zenddata,cubrid
+  # run PHP_CodeSniffer
+  - vendor/bin/phpcs framework
 
 after_script:
   - |
--- a/composer.json
+++ b/composer.json
@@ -82,6 +82,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "~4.4",
+        "squizlabs/php_codesniffer": "~2.8",
         "cebe/indent": "~1.0.2"
     },
     "suggest": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "30a97926034335c40273795ce316f36c",
-    "content-hash": "58ccdf60f6da90c6d1623d9755c0b830",
+    "content-hash": "d8b984a6c12c67e3f388d1da6d9c2683",
     "packages": [
         {
             "name": "bower-asset/jquery",
@@ -207,7 +206,7 @@
                 "markdown",
                 "markdown-extra"
             ],
-            "time": "2016-09-14 20:40:20"
+            "time": "2016-09-14T20:40:20+00:00"
         },
         {
             "name": "ezyang/htmlpurifier",
@@ -251,7 +250,7 @@
             "keywords": [
                 "html"
             ],
-            "time": "2016-07-16 12:58:58"
+            "time": "2016-07-16T12:58:58+00:00"
         },
         {
             "name": "yiisoft/yii2-composer",
@@ -298,7 +297,7 @@
                 "extension installer",
                 "yii2"
             ],
-            "time": "2016-02-06 00:49:24"
+            "time": "2016-02-06T00:49:24+00:00"
         }
     ],
     "packages-dev": [
@@ -333,7 +332,7 @@
                 }
             ],
             "description": "a small tool to convert text file indentation",
-            "time": "2014-05-23 14:40:08"
+            "time": "2014-05-23T14:40:08+00:00"
         },
         {
             "name": "doctrine/instantiator",
@@ -387,7 +386,7 @@
                 "constructor",
                 "instantiate"
             ],
-            "time": "2015-06-14 21:17:01"
+            "time": "2015-06-14T21:17:01+00:00"
         },
         {
             "name": "phpdocumentor/reflection-docblock",
@@ -436,7 +435,7 @@
                     "email": "mike.vanriel@naenius.com"
                 }
             ],
-            "time": "2015-02-03 12:10:50"
+            "time": "2015-02-03T12:10:50+00:00"
         },
         {
             "name": "phpspec/prophecy",
@@ -498,7 +497,7 @@
                 "spy",
                 "stub"
             ],
-            "time": "2016-06-07 08:13:47"
+            "time": "2016-06-07T08:13:47+00:00"
         },
         {
             "name": "phpunit/php-code-coverage",
@@ -560,7 +559,7 @@
                 "testing",
                 "xunit"
             ],
-            "time": "2015-10-06 15:47:00"
+            "time": "2015-10-06T15:47:00+00:00"
         },
         {
             "name": "phpunit/php-file-iterator",
@@ -607,7 +606,7 @@
                 "filesystem",
                 "iterator"
             ],
-            "time": "2015-06-21 13:08:43"
+            "time": "2015-06-21T13:08:43+00:00"
         },
         {
             "name": "phpunit/php-text-template",
@@ -648,7 +647,7 @@
             "keywords": [
                 "template"
             ],
-            "time": "2015-06-21 13:50:34"
+            "time": "2015-06-21T13:50:34+00:00"
         },
         {
             "name": "phpunit/php-timer",
@@ -692,7 +691,7 @@
             "keywords": [
                 "timer"
             ],
-            "time": "2016-05-12 18:03:57"
+            "time": "2016-05-12T18:03:57+00:00"
         },
         {
             "name": "phpunit/php-token-stream",
@@ -741,7 +740,7 @@
             "keywords": [
                 "tokenizer"
             ],
-            "time": "2015-09-15 10:49:45"
+            "time": "2015-09-15T10:49:45+00:00"
         },
         {
             "name": "phpunit/phpunit",
@@ -813,7 +812,7 @@
                 "testing",
                 "xunit"
             ],
-            "time": "2016-07-21 06:48:14"
+            "time": "2016-07-21T06:48:14+00:00"
         },
         {
             "name": "phpunit/phpunit-mock-objects",
@@ -869,7 +868,7 @@
                 "mock",
                 "xunit"
             ],
-            "time": "2015-10-02 06:51:40"
+            "time": "2015-10-02T06:51:40+00:00"
         },
         {
             "name": "sebastian/comparator",
@@ -933,7 +932,7 @@
                 "compare",
                 "equality"
             ],
-            "time": "2015-07-26 15:48:44"
+            "time": "2015-07-26T15:48:44+00:00"
         },
         {
             "name": "sebastian/diff",
@@ -985,7 +984,7 @@
             "keywords": [
                 "diff"
             ],
-            "time": "2015-12-08 07:14:41"
+            "time": "2015-12-08T07:14:41+00:00"
         },
         {
             "name": "sebastian/environment",
@@ -1035,7 +1034,7 @@
                 "environment",
                 "hhvm"
             ],
-            "time": "2016-08-18 05:49:44"
+            "time": "2016-08-18T05:49:44+00:00"
         },
         {
             "name": "sebastian/exporter",
@@ -1102,7 +1101,7 @@
                 "export",
                 "exporter"
             ],
-            "time": "2016-06-17 09:04:28"
+            "time": "2016-06-17T09:04:28+00:00"
         },
         {
             "name": "sebastian/global-state",
@@ -1153,7 +1152,7 @@
             "keywords": [
                 "global state"
             ],
-            "time": "2015-10-12 03:26:01"
+            "time": "2015-10-12T03:26:01+00:00"
         },
         {
             "name": "sebastian/recursion-context",
@@ -1206,7 +1205,7 @@
             ],
             "description": "Provides functionality to recursively process PHP variables",
             "homepage": "http://www.github.com/sebastianbergmann/recursion-context",
-            "time": "2015-11-11 19:50:13"
+            "time": "2015-11-11T19:50:13+00:00"
         },
         {
             "name": "sebastian/version",
@@ -1241,7 +1240,85 @@
             ],
             "description": "Library that helps with managing the version number of Git-hosted PHP projects",
             "homepage": "https://github.com/sebastianbergmann/version",
-            "time": "2015-06-21 13:59:46"
+            "time": "2015-06-21T13:59:46+00:00"
+        },
+        {
+            "name": "squizlabs/php_codesniffer",
+            "version": "2.8.1",
+            "source": {
+                "type": "git",
+                "url": "https://github.com/squizlabs/PHP_CodeSniffer.git",
+                "reference": "d7cf0d894e8aa4c73712ee4a331cc1eaa37cdc7d"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/squizlabs/PHP_CodeSniffer/zipball/d7cf0d894e8aa4c73712ee4a331cc1eaa37cdc7d",
+                "reference": "d7cf0d894e8aa4c73712ee4a331cc1eaa37cdc7d",
+                "shasum": ""
+            },
+            "require": {
+                "ext-simplexml": "*",
+                "ext-tokenizer": "*",
+                "ext-xmlwriter": "*",
+                "php": ">=5.1.2"
+            },
+            "require-dev": {
+                "phpunit/phpunit": "~4.0"
+            },
+            "bin": [
+                "scripts/phpcs",
+                "scripts/phpcbf"
+            ],
+            "type": "library",
+            "extra": {
+                "branch-alias": {
+                    "dev-master": "2.x-dev"
+                }
+            },
+            "autoload": {
+                "classmap": [
+                    "CodeSniffer.php",
+                    "CodeSniffer/CLI.php",
+                    "CodeSniffer/Exception.php",
+                    "CodeSniffer/File.php",
+                    "CodeSniffer/Fixer.php",
+                    "CodeSniffer/Report.php",
+                    "CodeSniffer/Reporting.php",
+                    "CodeSniffer/Sniff.php",
+                    "CodeSniffer/Tokens.php",
+                    "CodeSniffer/Reports/",
+                    "CodeSniffer/Tokenizers/",
+                    "CodeSniffer/DocGenerators/",
+                    "CodeSniffer/Standards/AbstractPatternSniff.php",
+                    "CodeSniffer/Standards/AbstractScopeSniff.php",
+                    "CodeSniffer/Standards/AbstractVariableSniff.php",
+                    "CodeSniffer/Standards/IncorrectPatternException.php",
+                    "CodeSniffer/Standards/Generic/Sniffs/",
+                    "CodeSniffer/Standards/MySource/Sniffs/",
+                    "CodeSniffer/Standards/PEAR/Sniffs/",
+                    "CodeSniffer/Standards/PSR1/Sniffs/",
+                    "CodeSniffer/Standards/PSR2/Sniffs/",
+                    "CodeSniffer/Standards/Squiz/Sniffs/",
+                    "CodeSniffer/Standards/Zend/Sniffs/"
+                ]
+            },
+            "notification-url": "https://packagist.org/downloads/",
+            "license": [
+                "BSD-3-Clause"
+            ],
+            "authors": [
+                {
+                    "name": "Greg Sherwood",
+                    "role": "lead"
+                }
+            ],
+            "description": "PHP_CodeSniffer tokenizes PHP, JavaScript and CSS files and detects violations of a defined set of coding standards.",
+            "homepage": "http://www.squizlabs.com/php-codesniffer",
+            "keywords": [
+                "phpcs",
+                "standards"
+            ],
+            "time": "2017-03-01T22:17:45+00:00"
         },
         {
             "name": "symfony/yaml",
@@ -1290,7 +1367,7 @@
             ],
             "description": "Symfony Yaml Component",
             "homepage": "https://symfony.com",
-            "time": "2016-10-21 20:59:10"
+            "time": "2016-10-21T20:59:10+00:00"
         }
     ],
     "aliases": [],
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -469,8 +469,10 @@ class BaseYii
     public static function powered()
     {
         return \Yii::t('yii', 'Powered by {yii}', [
-            'yii' => '<a href="http://www.yiiframework.com/" rel="external">' . \Yii::t('yii',
-                    'Yii Framework') . '</a>'
+            'yii' => '<a href="http://www.yiiframework.com/" rel="external">' . \Yii::t(
+                'yii',
+                'Yii Framework'
+            ) . '</a>'
         ]);
     }
 
--- a/framework/console/controllers/HelpController.php
+++ b/framework/console/controllers/HelpController.php
@@ -450,11 +450,12 @@ class HelpController extends Controller
         if (!empty($args)) {
             foreach ($args as $name => $arg) {
                 $this->stdout($this->formatOptionHelp(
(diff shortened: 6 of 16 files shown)


framework/assets/yii.gridView.js
```diff
@@ -237,7 +237,8 @@
      * @param {string} selector jQuery selector
      * @param {function} callback The actual function to be executed with this event
      */
-    function initEventHandler($gridView, type, event, selector, callback) {
+    function initEventHandler($gridView, type, event, selector, callback)
```
samdark: I don't think it should be applied to JavaScript.

framework/assets/yii.js
```diff
@@ -334,7 +334,8 @@ window.yii = (function ($) {
         }
     };
 
-    function initCsrfHandler() {
+    function initCsrfHandler()
```
samdark: I don't think it should be applied to JavaScript.

framework/assets/yii.validation.js
```diff
@@ -361,7 +361,8 @@ yii.validation = (function ($) {
         }
     };
 
-    function getUploadedFiles(attribute, messages, options) {
+    function getUploadedFiles(attribute, messages, options)
```
samdark: I don't think it should be applied to JavaScript.

framework/requirements/views/web/index.php
```diff
@@ -31,15 +31,15 @@ functionality may be not available in this case.
         </p>
 
         <h3>Conclusion</h3>
-        <?php if ($summary['errors'] > 0): ?>
+        <?php if ($summary['errors'] > 0) : ?>
```
samdark: Does not look correct to me.

framework/views/_addColumns.php
```diff
@@ -1,4 +1,4 @@
-<?php foreach ($fields as $field): ?>
+<?php foreach ($fields as $field) : ?>
```
samdark: Doesn't look correct to me.

framework/views/_addForeignKeys.php
```diff
@@ -1,4 +1,4 @@
-<?php foreach ($foreignKeys as $column => $fkData): ?>
+<?php foreach ($foreignKeys as $column => $fkData) : ?>
```
samdark: Doesn't look correct to me.

framework/views/_createTable.php
```diff
@@ -8,10 +8,10 @@
 /* @var $foreignKeys array the foreign keys */
 
 ?>        $this->createTable('<?= $table ?>', [
-<?php foreach ($fields as $field):
-    if (empty($field['decorators'])): ?>
+<?php foreach ($fields as $field) :
```
samdark: Doesn't look correct to me.

framework/views/_dropColumns.php
```diff
@@ -5,6 +5,6 @@
     'foreignKeys' => $foreignKeys,
 ]);
 
-foreach ($fields as $field): ?>
+foreach ($fields as $field) : ?>
```
samdark: Doesn't look correct to me.

framework/views/_dropForeignKeys.php
```diff
@@ -1,4 +1,4 @@
-<?php foreach ($foreignKeys as $column => $fkData): ?>
+<?php foreach ($foreignKeys as $column => $fkData) : ?>
```
samdark: Doesn't look correct to me.

samdark (changes requested):

framework/assets/yii.gridView.js
Kolyunya: I thing JS should be validated as well. If not by the `PHPCS` then by what? Don't mind dropping `JS` validation via `PHPSC` altogether though.

framework/requirements/views/web/index.php
Kolyunya: This seems to be correct: https://github.com/squizlabs/PHP_CodeSniffer/issues/366#issuecomment-66937317

> I'm not sure what PSR2 actually intends in this case. They don't mention the alternate syntax anywhere. So I still need to enforce the spaces after opening parentheses error, which means you code will look like this after I fix these issues:

framework/views/_addColumns.php
Kolyunya: This seems to be correct: https://github.com/squizlabs/PHP_CodeSniffer/issues/366#issuecomment-66937317

> I'm not sure what PSR2 actually intends in this case. They don't mention the alternate syntax anywhere. So I still need to enforce the spaces after opening parentheses error, which means you code will look like this after I fix these issues:

framework/views/_createTable.php
Kolyunya: This seems to be correct: https://github.com/squizlabs/PHP_CodeSniffer/issues/366#issuecomment-66937317

> I'm not sure what PSR2 actually intends in this case. They don't mention the alternate syntax anywhere. So I still need to enforce the spaces after opening parentheses error, which means you code will look like this after I fix these issues:

framework/views/_dropColumns.php
Kolyunya: This seems to be correct: https://github.com/squizlabs/PHP_CodeSniffer/issues/366#issuecomment-66937317

> I'm not sure what PSR2 actually intends in this case. They don't mention the alternate syntax anywhere. So I still need to enforce the spaces after opening parentheses error, which means you code will look like this after I fix these issues:

framework/views/_dropForeignKeys.php
Kolyunya: This seems to be correct: https://github.com/squizlabs/PHP_CodeSniffer/issues/366#issuecomment-66937317

> I'm not sure what PSR2 actually intends in this case. They don't mention the alternate syntax anywhere. So I still need to enforce the spaces after opening parentheses error, which means you code will look like this after I fix these issues:

framework/views/_addForeignKeys.php
Kolyunya: This seems to be correct: https://github.com/squizlabs/PHP_CodeSniffer/issues/366#issuecomment-66937317

> I'm not sure what PSR2 actually intends in this case. They don't mention the alternate syntax anywhere. So I still need to enforce the spaces after opening parentheses error...

framework/assets/yii.js
Kolyunya: https://github.com/yiisoft/yii2/pull/13732#discussion_r105060381

framework/assets/yii.validation.js
Kolyunya: https://github.com/yiisoft/yii2/pull/13732#discussion_r105060381

framework/views/_addColumns.php
Kolyunya: https://github.com/yiisoft/yii2/pull/13732#discussion_r105060849

framework/views/_addForeignKeys.php
Kolyunya: https://github.com/yiisoft/yii2/pull/13732#discussion_r105060849

framework/views/_createTable.php
Kolyunya: https://github.com/yiisoft/yii2/pull/13732#discussion_r105060849

framework/views/_dropColumns.php
Kolyunya: https://github.com/yiisoft/yii2/pull/13732#discussion_r105060849

framework/views/_dropForeignKeys.php
Kolyunya: https://github.com/yiisoft/yii2/pull/13732#discussion_r105060849

framework/views/errorHandler/callStackItem.php
```diff
@@ -10,35 +10,49 @@
 /* @var $args array */
 /* @var $handler \yii\web\ErrorHandler */
 ?>
-<li class="<?php if ($index === 1 || !$handler->isCoreFile($file)) echo 'application'; ?> call-stack-item"
```
Kolyunya: This is what `PHPCBF` thinks it should look like. Haven't reviewed this myself yet.

framework/views/errorHandler/callStackItem.php
```diff
@@ -10,35 +10,49 @@
...
     data-line="<?= (int) ($line - $begin) ?>">
     <div class="element-wrap">
         <div class="element">
             <span class="item-number"><?= (int) $index ?>.</span>
-            <span class="text"><?php if ($file !== null) echo 'in ' . $handler->htmlEncode($file); ?></span>
+            <span class="text"><?php if ($file !== null) {
+                echo 'in ' . $handler->htmlEncode($file);
+} ?></span>
             <span class="at">
-                <?php if ($line !== null) echo 'at line'; ?>
```
Kolyunya: This is what `PHPCBF` thinks it should look like. Haven't reviewed this myself yet.

framework/views/errorHandler/exception.php
```diff
@@ -12,16 +14,16 @@
 
     <title><?php
         $name = $handler->getExceptionName($exception);
-        if ($exception instanceof \yii\web\HttpException) {
```
Kolyunya: This is what `PHPCBF` thinks it should look like. Haven't reviewed this myself yet.

framework/views/errorHandler/exception.php
```diff
@@ -345,44 +347,50 @@
...
             <h1><?php
-                if ($exception instanceof \yii\web\HttpException) {
```
Kolyunya: This is what `PHPCBF` thinks it should look like. Haven't reviewed this myself yet.

framework/views/_createTable.php
Kolyunya: https://github.com/yiisoft/yii2/pull/13732#discussion_r105060849

framework/assets/yii.gridView.js
samdark: It's OK to be validated but the standard should not be the same as for PHP. Common placement of `{` in JavaScript is same line.

framework/views/errorHandler/callStackItem.php
samdark: This fix is OK.

framework/requirements/views/web/index.php
samdark: No PSR mentions it including PSR-12 in the works but Yii's internal doc does: https://github.com/yiisoft/yii2/blob/master/docs/internals/view-code-style.md

Accoding to the doc it should be:

```php
<?php if ($summary['errors'] > 0): ?>
```

i.e. no space.

framework/assets/yii.gridView.js
Kolyunya: Removed `*.js` files validation altogether for the time being.

framework/requirements/views/web/index.php
Kolyunya: Excluded that rule.

rob006:
I think that using https://github.com/FriendsOfPHP/PHP-CS-Fixer would be much more useful and flexible.

Kolyunya:
@rob006 what are the particular issues of the current setup which could be solved the `PHP-CS-Fixer`?

rob006:
PHP-CS-Fixer is easier to configure and extend. It has more active development and community IMO. 

I also made a quick look into fixes from this PR and many of this changes looks invalid and just dumb. I never had such problems with `PHP-CS-Fixer` before. 

I could open alternative PR with PHP-CS-Fixer integration, so we could just compare results and choose better tool for this.

Kolyunya:
> many of this changes looks invalid and just dumb.

Could you name them explicitly please (excluding `views`)?

rob006:
```diff
diff --git a/framework/db/Query.php b/framework/db/Query.php
index b20a8b4..c95b734 100644
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -408,8 +408,7 @@ protected function queryScalar($selectExpression, $db)
             return null;
         }
 
-        if (
-            !$this->distinct
+        if (!$this->distinct
             && empty($this->groupBy)
             && empty($this->having)
             && empty($this->union)
```

```diff
diff --git a/framework/mutex/OracleMutex.php b/framework/mutex/OracleMutex.php
index bf78add..b0f1d88 100644
--- a/framework/mutex/OracleMutex.php
+++ b/framework/mutex/OracleMutex.php
@@ -89,14 +89,14 @@ protected function acquireLock($name, $timeout = 0)
 
         /** inside pl/sql scopes pdo binding not working correctly :(  */
         $this->db->createCommand(
-                'DECLARE
+            'DECLARE
     handle VARCHAR2(128);
 BEGIN
     DBMS_LOCK.ALLOCATE_UNIQUE(:name, handle);
     :lockStatus := DBMS_LOCK.REQUEST(handle, DBMS_LOCK.' . $this->lockMode . ', ' . $timeout . ', ' . $releaseOnCommit . ');
 END;',
-                [':name' => $name]
-            )
+            [':name' => $name]
+        )
             ->bindParam(':lockStatus', $lockStatus, PDO::PARAM_INT, 1)
             ->execute();
```

Why "excluding /views/"? Don't you want fix views files too?

Kolyunya:
> PHP-CS-Fixer is easier to configure and extend.

@rob006  what do you mean by `easier`? Configuring each of the rules is a matter of a single (couple) lines in the config.

```php
diff --git a/framework/db/Query.php b/framework/db/Query.php
index b20a8b4..c95b734 100644
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -408,8 +408,7 @@ protected function queryScalar($selectExpression, $db)
             return null;
         }
 
-        if (
-            !$this->distinct
+        if (!$this->distinct
             && empty($this->groupBy)
             && empty($this->having)
             && empty($this->union)
```
This is a matter of the convention. I like the proposed change.

```php
diff --git a/framework/mutex/OracleMutex.php b/framework/mutex/OracleMutex.php
index bf78add..b0f1d88 100644
--- a/framework/mutex/OracleMutex.php
+++ b/framework/mutex/OracleMutex.php
@@ -89,14 +89,14 @@ protected function acquireLock($name, $timeout = 0)
 
         /** inside pl/sql scopes pdo binding not working correctly :(  */
         $this->db->createCommand(
-                'DECLARE
+            'DECLARE
```
Why would you use 8 spaces for indentation instead of 4 here?

Kolyunya:
@rob006 I simplified the config to the maximum extent possible and left *only* `PSR-2` rules while removing all others. Would you mind reviewing reported issues now?

Kolyunya:
```php
diff --git a/framework/db/Query.php b/framework/db/Query.php
index b20a8b4..c95b734 100644
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -408,8 +408,7 @@ protected function queryScalar($selectExpression, $db)
             return null;
         }
 
-        if (
-            !$this->distinct
+        if (!$this->distinct
             && empty($this->groupBy)
             && empty($this->having)
             && empty($this->union)
```

@rob006 this is an obvious violation of the `PSR-2`.
>Opening parentheses for control structures MUST NOT have a space after them, and closing parentheses for control structures MUST NOT have a space before.

rob006:
> what do you mean by easier?

In PHP-CS-Fixer configuration is an object. So you could create own class with configuration and easy share between different projects. You could also easy modify this configuration by extending config class. 

We could create composer package with yii2 specific configuration and fixers - then everybody could use Yii2 CS in his own project and made some modifications by extending yii2 cs config class. With `phpcs.xml` you need copy this file and keep sync manually.

> This is a matter of the convention. I like the proposed change.

It is much less readable and consistent. With long conditions it is easy to miss that indentation block is part of condition and not `if` body.

You also don't use 

```php
if ($something) { $this->doSomethig();
    $this->doSomethingElse();
}
```
so why you want do this in `if` condition block?

> Why would you use 8 spaces for indentation instead of 4 here?

I'm referring to:

```diff
-                [':name' => $name]
-            )
+            [':name' => $name]
+        )
             ->bindParam(':lockStatus', $lockStatus, PDO::PARAM_INT, 1)
             ->execute();
```

Why closing parenthesis is not intended with `->bindParam()`?

Kolyunya:
> It is much less readable and consistent.

The original version violates the standard:

> Opening parentheses for control structures MUST NOT have a space after them...

Regarding
>Why closing parenthesis is not intended with ->bindParam()?

Why should it be? Could you provide a quote from the standard please?

rob006:
> The original version violates the standard:
> >   Opening parentheses for control structures MUST NOT have a space after them...

To be strict, there is no space after opening parenthesis - there is only a single `LF` char.
But there is a space before closing parenthesis (from indentation) and it is not fixed. Do you think that it should? :D

> Why should it be? Could you provide a rule from the standard which is violated please?

There is no rule, you just broke nice indentation. For what?

Kolyunya:
> There is no rule, you just broke nice indentation. For what?

@rob006 would you mind defining `nice`? `->bindParam()` is indented one block to the right of the `createCommand()` and this indentation is used all over the core [e.g.](https://github.com/yiisoft/yii2/blob/master/framework/caching/DbCache.php#L106).

rob006:
Nice:
```php
        $query
            ->select(
                ['COUNT(*)']
            )
            ->from($this->cacheTable)
            ->where('[[id]] = :id AND ([[expire]] = 0 OR [[expire]] >' . time() . ')', [':id' => $key]);
```

Not nice:
```php
        $query->select(
            ['COUNT(*)']
        )
            ->from($this->cacheTable)
            ->where('[[id]] = :id AND ([[expire]] = 0 OR [[expire]] >' . time() . ')', [':id' => $key]);
```

Whole call chain should be intended once, so you can easily see when it starts and when it ends.

Kolyunya:
@rob006 this could become a `nice` proposal to the `PSR`. I'd gonna upvote it :D

samdark (approved):

dynasource:
@Kolyunya, whats your opinion about PHP-CS-Fixer?

Kolyunya:
@dynasource at the time of posting this PR I haven't tried it yet. I tried it a couple of days ago and it seemed to me that it's not a silver bullet either and should be tweaked too.

I think that we should stick to the tool which would provide most accurate reports. I just proposed a tool I was aware of.

dynasource:
I think it is a good suggestion and that it is useful to have one. Now we have 2 different solutions proposed here, we should choose one. Would be nice to have opinions from the @community.

photodude:
If you are just applying PSR2 why not use style-ci see https://styleci.io/

PHPCS and others are more for code styles that have complex requierments beyond things like PSR2

Kolyunya:
We will probably want to enforce custom Yii2 code-style.

photodude:
Do you have an existing list for what you want to enforce?

for example
- https://developer.joomla.org/coding-standards.html
- https://codex.wordpress.org/WordPress_Coding_Standards
- https://www.drupal.org/docs/develop/standards
- http://symfony.com/doc/current/contributing/code/standards.html
- https://laravel.com/docs/5.4/contributions#coding-style
- https://www.codeigniter.com/user_guide/general/styleguide.html

Kolyunya:
[Sure](https://github.com/yiisoft/yii2/blob/master/docs/internals/view-code-style.md).

photodude:
I think you will find some of those items can't be enforced without custom enforcer scripts (for example the way the comments and docblocks are written) as there is no standard available for what your presenting.

photodude:
There is also a big difference between a written code style with examples, and just an example file that shows acceptable coding (the latter is what you presented).

gsherwood:
Hi all. I maintain PHPCS so I thought I'd drop in and clear some things up.

PHPCS is a tool for checking coding standards in your project, whatever they may be. It ships with a few standards (including the old PEAR standard and the much newer PSR-1 and PSR-2 standards) but it is designed for you to pick and choose from the included rules and design your own. You can also write your own checks and auto-fixes if none of the included checks meet your requirements. For projects that have no idea what standard to use, using a built-in one is probably the easiest thing to do, but that's not a requirement. A lot of projects customise their coding standards, even if it is a simple as PSR-2 minus a couple of things they don't like.

PHPCS ships with lots of different report types to help you integrate it into IDEs, CI systems, text editors, and custom code, but you can also define your own custom report classes if you want. One report that might be useful is the [information](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Reporting#printing-an-information-report) report. This one is used to tell you information about your code base and is helpful when deciding on a standard. I use this report to generate my [Analysis of Coding Conventions in PHP projects](https://squizlabs.github.io/PHP_CodeSniffer/analysis/). Note that I have Yii2 in there and I've found that your code is very consistent. You can view the report of how the project's coding conventions have changed over the last 2+ years here: https://squizlabs.github.io/PHP_CodeSniffer/analysis/yiisoft/yii2/index.html

Writing your own standard for PHPCS is a matter of creating an XML file. The format is quite extensive and allows for customisiation of the checks performed, but also the severity of each message, the message content itself, regex based file exclusions for checks (some projects don't want specific checks done on specific files or folders), and overriding a complete set of config variables.

That XML format is documented here: https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-ruleset.xml

Also note that you can commit your standard as a file called `phpcs.xml` (or `phpcs.xml.dist`) in your project's root directory so that developers can just run `phpcs` and the standard tells PHPCS which files and directories to process, and what checks to perform. But you don't have to. You could also just pass around the XML file and use it with PHPCS to ensure people are running the same checks over their code. Or a sub-project could create their own custom standard that imports your XML file and changes a few things.

WordPress does this with [their coding standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards). You can see they have a main WordPress Core standard, but also extensions like VIP and Extra. They do a great job with their coding standards, have a lot of custom sniffs, contribute a lot back to PHPCS, and can fix quite a lot of the found errors automatically. It's worth taking a look at that repo if you want to see PHPCS in use.

[Drupal's Coder](https://www.drupal.org/project/coder) is another example of this, but it is less extensive. I also get a lot of good contributions from the developers of this project though.

The 3.0 version is the best to use right now if you are starting fresh, and that version also allows you to make use of PHPCS directly in code. For example, if you just want to run PHPCS over a snippet of code, you could do it like this: https://gist.github.com/gsherwood/aafd2c16631a8a872f0c4a23916962ac . Or if you'd like to tokenize JS files th same way PHP files are tokenized, you can use this code: https://gist.github.com/gsherwood/f17cfc90f14a9b29eeb6b2e99e6e7f66 . It adds a bit more flexibility to what is typically just a CLI tool.

PHPCS also comes with a tool called PHPCBF (officially, the PHP Code Beautifier and Fixer, but maybe you can figure out the unofficial name) which is used to fix coding standard issues. PHPCS reports errors and PHPCBF fixes them. Quite a lot of errors that PHPCS finds are fixable, and I've spent several years tweaking the fixers (especially the one that does indents) to work with a wide range of code. 

Despite concerns here that PHPCS development and community is not very active, I can tell you that I get a lot of contributions from developers and I've seen code in bug reports that no developer should probably see. But that helps me make PHPCS more robust and able to deal with a range of issues, including graceful handling of parse errors, which is pretty important if you want developers to be constantly checking standards in their IDE.

I hope this clears up a few things about PHPCS and hopefully I haven't taken up too much of your time.

Ultimately, you should choose whichever tool you think works best for your project. But I wanted to make it clear that PHPCS has been actively developed for more than 10 years and still gets 30-40k downloads a day. It's an active project with an active user base, and I think it has a very extensive feature set that makes it flexible enough for most projects. I'd encourage you to give it a go if you haven't already. It's free :)

cebe:
@gsherwood thanks for sharing your thoughts about this!

samdark:
So overall PHPCS looks like a good match and the changes in the pull request do make sense. @cebe how about merging it?

Kolyunya:
@samdark I think we should tweak config a bit. PHP CS Fixer [reported](https://github.com/yiisoft/yii2/pull/13778/files) some issues which PHPCS didn't.

samdark:
OK. Moving to 2.0.13 then.

jrfnl:
Hiya, I've had a quick look at some of the things PHP CS Fixes found which weren't reported by PHPCS so far and would like to make the below suggestions for rules to add to your ruleset to cover these.
You may also want to review the [existing exclusions in your ruleset](https://github.com/yiisoft/yii2/pull/13732/files#diff-6b216d983cf0a62e9c648934c5687456R8-10) carefully as they are partially to blame for the issues which weren't found.

N.B.: If you'd include the below suggestions, you may need/want to tweaks them a bit to exclude extraneous error codes coming from these sniffs which you are not interested in.

```xml
<!-- Covers: https://github.com/yiisoft/yii2/pull/13778/files#diff-ef822e03c117e20fd1d786afc7c5b5e3L772 and similar -->
<rule ref="Squiz.ControlStructures.ControlSignature" />

<!-- Covers: https://github.com/yiisoft/yii2/pull/13778/files#diff-03b64c04d37c13a3118b4b8119f5ae8dL439, https://github.com/yiisoft/yii2/pull/13778/files#diff-ef822e03c117e20fd1d786afc7c5b5e3L219, https://github.com/yiisoft/yii2/pull/13778/files#diff-85a8556277e468a30afc79beb7aa25dfL660 and similar -->
<rule ref="Generic.WhiteSpace.ScopeIndent">
    <properties>
        <property name="exact" value="4" />
    </properties>
</rule>

<!-- Covers: https://github.com/yiisoft/yii2/pull/13778/files#diff-c6dd99ed2184ed7a7f316aee5fe67513L492 and similar -->
<rule ref="Squiz.Commenting.PostStatementComment" />

<!-- Covers: https://github.com/yiisoft/yii2/pull/13778/files#diff-709ff61c838c754d533b87ec33f14908L115 and similar -->
<rule ref="PSR2.ControlStructures.ElseIfDeclaration.NotAllowed">
    <type>error</type>
</rule>

<!-- Covers: https://github.com/yiisoft/yii2/pull/13778/files#diff-f000680354101249e9f60e71e6278e3bL136 and similar -->
<rule ref="Generic.Functions.OpeningFunctionBraceBsdAllman" />

<!-- If you also want to enforce braces on the next line for closures: -->
<rule ref="Generic.Functions.OpeningFunctionBraceBsdAllman">
    <properties>
        <property name="checkClosures" value="true" />
    </properties>
</rule>

<!-- Covers: https://github.com/yiisoft/yii2/pull/13778/files#diff-3f97b6f0e759a8a52920b07abe345579L69, https://github.com/yiisoft/yii2/pull/13778/files#diff-74bb1570acfdb8abc2d6d89e79b0ab54L88 and similar -->
<rule ref="Squiz.WhiteSpace.SuperfluousWhitespace" />
```

I'm pretty sure there's actually also a sniff which will catch the extraneous blank lines, but I can't think of the name of the sniff off-hand.

Hope this helps ;-)

rob006:
see #14100

samdark:
I like @rob006 approach more. Closing this one. @Kolyunya thank you for preparing the pull request. At least it triggered the continuation @rob006 did.

Kolyunya:
@samdark I'm just curious what do you mean by the `approach`? It's possible to extract config and custom rules to a separate repository and it's possible to enable same rules.

samdark:
PHP configs are a bit more flexible than XML IMO but that both approaches (tools) are OK. Also @rob006 covered a lot of cases without false positives so it doesn't make sense to re-do the same using different tool, I think.

PR #13739, opened 2017-03-09
Fixes #13738: getQueryParams() method in yii.js fails on URL with question mark and no query parameters

### LINKED ISSUE #13738: getQueryParams() method in yii.js fails on URL with question mark and no query parameters

1. Make GridView with filter in column
2. Enter some text, press Enter

Filter must work without JS errors

### What do you get instead?

Filter not work.

In Console i see error "Uncaught TypeError: Cannot read property 'replace' of undefined"
image

This error appeared after this commit https://github.com/yiisoft/yii2/commit/17df7577a53ee2ac35d8e53ed87aaf03ce9309ac#diff-f433792b516845f488519deb1a72741aR293

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.?
| PHP version      |
| Operating system |

vlydev:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13738

### DIFF
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -275,7 +275,8 @@ window.yii = (function ($) {
 
             var pairs = $.grep(url.substring(pos + 1).split('#')[0].split('&'), function (value) {
                 return value !== '';
-            }), params = {};
+            });
+            var params = {};
 
             for (var i = 0, len = pairs.length; i < len; i++) {
                 var pair = pairs[i].split('=');
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -279,10 +279,10 @@ window.yii = (function ($) {
             for (var i = 0, len = pairs.length; i < len; i++) {
                 var pair = pairs[i].split('=');
                 var name = decodeURIComponent(pair[0].replace(/\+/g, '%20'));
-                var value = decodeURIComponent(pair[1].replace(/\+/g, '%20'));
                 if (!name.length) {
                     continue;
                 }
+                var value = decodeURIComponent(pair[1].replace(/\+/g, '%20'));
                 if (params[name] === undefined) {
                     params[name] = value || '';
                 } else {
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -274,7 +274,7 @@ window.yii = (function ($) {
             }
 
             var pairs = $.grep(url.substring(pos + 1).split('#')[0].split('&'), function (value) {
-                return value != '';
+                return value !== '';
             }), params = {};
 
             for (var i = 0, len = pairs.length; i < len; i++) {
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -273,16 +273,16 @@ window.yii = (function ($) {
                 return {};
             }
 
-            var pairs = url.substring(pos + 1).split('#')[0].split('&'),
+            var pairs = url.substring(pos + 1).split('#')[0].split('&').filter(String),
                 params = {};
 
             for (var i = 0, len = pairs.length; i < len; i++) {
                 var pair = pairs[i].split('=');
                 var name = decodeURIComponent(pair[0].replace(/\+/g, '%20'));
+                var value = decodeURIComponent(pair[1].replace(/\+/g, '%20'));
                 if (!name.length) {
                     continue;
                 }
-                var value = decodeURIComponent(pair[1].replace(/\+/g, '%20'));
                 if (params[name] === undefined) {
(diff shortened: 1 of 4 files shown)


vladdnepr:
@arogachev tests for JS?

samdark:
Yes: https://github.com/yiisoft/yii2/tree/master/tests/js/tests

vladdnepr:
@samdark Cool. I didnt notice. Already added.

framework/assets/yii.js
```diff
@@ -279,10 +279,10 @@ window.yii = (function ($) {
             for (var i = 0, len = pairs.length; i < len; i++) {
                 var pair = pairs[i].split('=');
                 var name = decodeURIComponent(pair[0].replace(/\+/g, '%20'));
-                var value = decodeURIComponent(pair[1].replace(/\+/g, '%20'));
                 if (!name.length) {
                     continue;
                 }
+                var value = decodeURIComponent(pair[1].replace(/\+/g, '%20'));
```
arogachev: I think this should be fixed differently. In error case we have an array with one element - empty string - `[""]`. It's better to check and remove it from array before the loop (or construct array in a way that prevents it) so it will not be considered as pair at all. We don't have any name - value pairs with url `/posts/index?`, I think it's more correct.

tests/js/tests/yii.test.js
```diff
@@ -739,6 +739,7 @@ describe('yii', function () {
     describe('getQueryParams method', function () {
         withData({
             'no query parameters': ['/posts/index', {}],
+            'no query parameters with question char': ['/posts/index?', {}],
```
arogachev: Ideally this needs to be tested for `yii.gridView.js` too. See https://github.com/yiisoft/yii2/blob/9fa66b23a53e8268b1808977146056b92ac60a8a/tests/js/tests/yii.gridView.test.js#L305.

arogachev (changes requested):

arogachev:
Also according CHANGELOG line is required.

vladdnepr:
@arogachev All requested changes are made

framework/assets/yii.js
```diff
@@ -273,7 +273,7 @@ window.yii = (function ($) {
                 return {};
             }
 
-            var pairs = url.substring(pos + 1).split('#')[0].split('&'),
+            var pairs = url.substring(pos + 1).split('#')[0].split('&').filter(String),
```
arogachev: First of all, thanks for the made improvements. Here are some remarks however:

1) In `pairs` array we already have strings only so I feel that repeated string conversion is redundant. Something more explicit like:

```js
var pairs = url.substring(pos + 1).split('#')[0].split('&').filter(function (value) {
    return value !== '';
})
```

would be better IMO.

2) Used [Array.prototype.filter](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) method is OK, but:

>`filter()` was added to the ECMA-262 standard in the 5th edition; as such it may not be present in all implementations of the standard.

To be more specific, it's not supported in Internet Explorer 8 and I'm not sure that we can just refuse supporting it (there is a small chance that this requirement takes place to be in someone's project based on Yii2). I don't think that polyfill will be a good option neither, because its usage currently will be limited by this code only and it seems like an overhead. Regular loop can solve the problem, or maybe consider an option to try not include such values initially to avoid additional filter loop.

@samdark Please remind, do we support Internet Explorer 8?

arogachev (changes requested):

framework/assets/yii.js
samdark: If it could be easily replaced with something compatible with earlier standards, it's better to use more compatible solution. If there's no easy replacement, let it be as is.

vladdnepr: @arogachev OK. I made filtering by jQuery grep

vladdnepr:
@arogachev OK. I made filtering by jQuery grep

arogachev:
I've made some corrections. You can find out what was wrong from according commits. Now IMO it's ready for merge.

@samdark We should use our Github nicknames in CHANGELOG, shouldn't we?

arogachev (approved):

samdark (approved):

samdark:
> @samdark We should use our Github nicknames in CHANGELOG, shouldn't we?

Yes. It's easier to find a person who did the change this way.

samdark:
Merged. Thanks!

Issue #13733, opened 2017-03-08
Error on compress "yii\widgets\MaskedInputAsset" using asset command


AndersonDan:
This error is displayed in the structure of "jquery.inputmask.bundle.js":

> Parse error. '}' expected
> null !== test.fn && void 0 !== testPos.input ? static && null !== test.fn && void 0 !== testPos.input && (static = !1

samdark:
Which compressing tool are you using?

AndersonDan:
This is the configuration file for the asset compress:

```php
<?php
/**
 * @link https://www.tubaron.com.br
 * @copyright Copyright (c) 2016-2017 Tubaron LTDA
 * @author Anderson Scouto da Silva <and.dan19@gmail.com>
 */

/**
 * Configuration file for the "yii asset" console command.
 */

// In the console environment, some path aliases may not exist. Please define these:
Yii::setAlias('@webroot', __DIR__ . '/../web');
Yii::setAlias('@web', '/');

return [
    'jsCompressor' => 'java -jar bin/compiler/assets/closure-compiler-v20160911.jar --js {from} --js_output_file {to} --jscomp_off=uselessCode',
    'cssCompressor' => 'java -jar bin/compiler/assets/yuicompressor-2.4.8.jar --type css {from} -o {to}',
    'bundles' => [
        'app\themes\material\MaterialThemeAsset',

        'tubaron\assets\TubaronAsset',
        'tubaron\assets\LightGalleryAsset',
        'tubaron\phoneInput\PhoneInputAsset',
        'tubaron\assets\waitme\WaitMeAsset',
        'tubaron\assets\SpinJsAsset',
        'tubaron\assets\LaddaButtonsAsset',
        'tubaron\assets\SweetAlertAsset',
        'tubaron\assets\JcropAsset',
        'tubaron\assets\LightGalleryAsset',
        'tubaron\assets\SpinJsAsset',
        'tubaron\assets\TextareaAutosizeAsset',
        'tubaron\widgets\AvatarImageUploadAsset',
        'tubaron\widgets\DualListboxAsset',
        'tubaron\web\OutDatedBrowserAsset',
        'tubaron\web\JqueryStorageApiAsset',

        'kartik\select2\ThemeBootstrapAsset',
        'kartik\select2\Select2Asset',
        'kartik\date\DatePickerAsset',
        'kartik\datecontrol\DateFormatterAsset',
        'kartik\file\FileInputAsset',
[...]

PR #13741, opened 2017-03-09
Update jquery.inputmask

dan-developer:
| Q             | A
| ------------- | ---
| Is bugfix?        | yes
| New feature?  | no
| Breaks BC?      | no
| Tests pass?      | yes
| Fixed issues     | #13733


samdark:
There's no need to update dependency. ~3.3.3 will install 3.3.4 if it's available.

PR #13743, opened 2017-03-10
Short syntax for arrays

gagatust:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/requirements/YiiRequirementChecker.php
+++ b/framework/requirements/YiiRequirementChecker.php
@@ -74,14 +74,14 @@ class YiiRequirementChecker
             $this->usageError('Requirements must be an array, "' . gettype($requirements) . '" has been given!');
         }
         if (!isset($this->result) || !is_array($this->result)) {
-            $this->result = array(
-                'summary' => array(
+            $this->result = [
+                'summary' => [
                     'total' => 0,
                     'errors' => 0,
                     'warnings' => 0,
-                ),
-                'requirements' => array(),
-            );
+                ],
+                'requirements' => [],
+            ];
         }
         foreach ($requirements as $key => $rawRequirement) {
             $requirement = $this->normalizeRequirement($rawRequirement, $key);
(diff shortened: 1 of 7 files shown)


framework/requirements/requirements.php
```diff
@@ -5,53 +5,53 @@
...
 /* @var $this YiiRequirementChecker */
-return array(
-    array(
+return [
+    [
         'name' => 'PHP version',
         'mandatory' => true,
         'condition' => version_compare(PHP_VERSION, '5.4.0', '>='),
         'by' => '<a href="http://www.yiiframework.com">Yii Framework</a>',
         'memo' => 'PHP 5.4.0 or higher is required.',
```
arogachev: I think this replacement should not be done for this file. There is a small chance of running this on PHP < 5.4.0 where short array syntax is not yet supported. While it's OK for other core framework files because Yii 2 requires PHP >= 5.4.0 and it's mentioned in the docs, requirements checker has different purpose, it executes separately and out of framework's scope, before its installation. So for earlier versions of PHP instead of showing "PHP 5.4.0 or higher is required." message it will simply fail with parsing square brackets error.

framework/requirements/YiiRequirementChecker.php
```diff
@@ -74,14 +74,14 @@ function check($requirements)
             $this->usageError('Requirements must be an array, "' . gettype($requirements) . '" has been given!');
         }
         if (!isset($this->result) || !is_array($this->result)) {
-            $this->result = array(
-                'summary' => array(
+            $this->result = [
```
arogachev: The same applies to this file as well.

arogachev (changes requested):

framework/requirements/requirements.php
takobell: Agree with @arogachev.

gagatust: You are right.

framework/requirements/YiiRequirementChecker.php
gagatust: Ok

samdark (approved):

SilverFire (approved):

samdark:
Merged. Thanks!

arogachev:
@gagatust Did you check all framework related files with global search in IDE?

gagatust:
@arogachev I used http://plugins.netbeans.org/plugin/51791/php-enhancements plugin for yii2 repo only.

arogachev:
@gagatust OK. Thanks for the contribution!

Issue #13750, opened 2017-03-10
Test-suite fails with ICU 57.1


schmunk42:
### What steps will reproduce the problem?

Run tests on a system with ICU 57.1
https://github.com/yiisoft/yii2/pull/13360#issuecomment-284504403

### What is the expected result?

Green tests

### What do you get instead?

Red tests, see https://travis-ci.org/yiisoft/yii2-docker/jobs/209762936#L2178

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | Any
| Operating system | Alpine LInux (Docker)

Running them on i.e. Debian with 52.1 is fine. See https://gitlab.com/yiisoft/yii2-docker/builds/11975020 or https://travis-ci.org/yiisoft/yii2-docker/builds/209762928

*Ignore the green alpine-7.1 build*

samdark:
@schmunk42 I've adjusted tests a bit to tolerate valid ICU differences.

schmunk42:
There are still some issues left: https://gitlab.com/yiisoft/yii2/builds/12180932

schmunk42:
It looks to me like it does not use the new variations in the test data: https://gitlab.com/yiisoft/yii2/builds/12180932 - like it has no effect to check the ICU version.

Or do I get that wrong?

samdark:
Weird but seems so.

samdark:
I can re-do it other way.

samdark:
https://github.com/yiisoft/yii2/pull/13797

samdark:
@schmunk42 is that still the case?

amarox:
Actual ICU version is 60.2

schmunk42:
@amarox With which OS and PHP versions are you testing?

schmunk42:
@samdark Could you review https://github.com/yiisoft/yii2/pull/15559 - it would greatly help for testing issues like this.

samdark:
#15559 merged.

schmunk42:
-  the builds on Travis still use ICU version `52.1` and are passing
- `-debian` builds (7.2) on GitLab use `57.1` and are passing https://gitlab.com/yiisoft/yii2-docker/-/jobs/48925154
- `-alpine` builds (7.2) on GitLab use `58.2` and are failing https://gitlab.com/yiisoft/yii2-docker/-/jobs/51069444 - but failed tests seem not related to ICU

So I think this can be closed.

[...]

PR #13751, opened 2017-03-10
Attempt to fix ICU tests

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #13750

### DIFF
--- a/tests/framework/helpers/InflectorTest.php
+++ b/tests/framework/helpers/InflectorTest.php
@@ -234,6 +234,14 @@ class InflectorTest extends TestCase
         }
     }
 
+    /**
+     * @return boolean if ICU version is >= 57.1
+     */
+    private function hasNewICUData()
+    {
+        return version_compare(INTL_ICU_DATA_VERSION, '57.1', '>=');
+    }
+
     public function testTransliterateMedium()
     {
         if (!extension_loaded('intl')) {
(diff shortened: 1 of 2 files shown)


mikehaertl:
@samdark Should this also go into the requirement checker? We had a discusstion yesterday where @schmunk42 pointed out, that the checker is not reliable when it comes to ICU. He suggested to run the complete yii2 test suite when building the docker images instead of just the requirement checker. I'd say we should rather fix the checker instead.

samdark:
@mikehaertl what exactly should go into the checker?

schmunk42:
> that the checker is not reliable when it comes to ICU

What I meant is that the checker just checks the version of the library, the failing tests go much deeper.

I can't tell if the tests are wrong or a framework function, but in any case, it's good to see that early.

schmunk42:
What's about those, btw?

```
5) yiiunit\framework\validators\DateValidatorTest::testIntlValidationWithTime with data set #0 ('UTC')

/project/tests/framework/validators/DateValidatorTest.php:294
/project/vendor/phpunit/phpunit/phpunit:52
```

Are they also related to ICU problems?

samdark:
That would require PHPUnit. Not a good idea IMO since checker was meant to be lightweight.

samdark:
> Are they also related to ICU problems?
Probably. I haven't checked these ones yet.

mikehaertl:
> That would require PHPUnit. Not a good idea IMO since checker was meant to be lightweight.

@samdark I haven't looked into the details of this issue und if this could happen at runtime. But if there's a potential runtime error due to a problem with some extension, shouldn't the checker inform you about that? 

As a user I would probably trust the checker and would be surprised to see app errors that are caused by a problem with some extension.

samdark:
Yes, as a user I'd be surprised too but also as a user I won't tolerate installing phpunit at the server just for the sake of checking requirements once.

mikehaertl:
>  as a user I won't tolerate installing phpunit at the server just for the sake of checking requirements once.

Of course not. Is there really no way to test this without PHPUnit? What kind of special tests are these that we can't include them as simple PHP commands in the req checker?

samdark:
Ah, right. We can do that.

Tests aren't small in this area so I though you propose reusing these.

PR #13752, opened 2017-03-10
Improve SQL LIKE condition handling and testing (fixes #13745)

### LINKED ISSUE #13745: SQLSTATE[HY093]: Invalid parameter number: parameter was not defined (MSSQL)
After update yii2. Error after commit https://github.com/yiisoft/yii2/commit/9a915ba10c34171d1b8a51ab97fd2eda4b90ef64
Working SQL as before
### Additional info

| Q                           | A
| ----------------- | ---
| Yii version             | 2.0.12-dev
| PHP version          | 7.1.2
| Operating system | macOS 10.12.3 / CentOS 7.3 x64 / Oracle Linux 7.3 x64

I use search model with this code:
```
    {
        $query = PersonnelEmployee::find()
            ->where(['or',
                ['FireDate' => null],
                ['>', 'FireDate', $this->getFilterDate()],
            ])
            ->with([
                'company',
                'mailboxType',
                'hardware',
                'personnelPhoneNumbers',
            ]);
            'query' => $query,
        ]);

        $this->load($params, '');

        if (!empty($this->text)) {

            $textRoll = $this->changeSearchString($this->text);

            $query
                ->joinWith(['domainUser', 'personnelPhoneNumbers'], false)
                ->andFilterWhere(['or',
                    ['like', 'FullName', $this->text],
                    ['like', 'MaidenName', $this->text],
                    ['like', 'sAMAccountName', $this->text],
                    ['like', 'mail', $this->text],
                    ['like', 'PhoneNumber', $this->text],
                    ['like', 'FullName', $textRoll],
                    ['like', 'MaidenName', $textRoll],
                    ['like', 'sAMAccountName', $textRoll],
                    ['like', 'mail', $textRoll],
                ]);

        } else {
            return $dataProvider = null;
        }

        return $dataProvider;
    }
```

After update yii2 and apply commit https://github.com/yiisoft/yii2/commit/9a915ba10c34171d1b8a51ab97fd2eda4b90ef64 broken search SQL.

If use commit https://github.com/yiisoft/yii2-framework/commit/d17aadf1eb1b9c436998c64046f54c8e91c8df20 everything is working.

Any help?

sergeymakinen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13745 

This PR adds real db tests of QueryBuilder’s `LIKE` handling and fixes all bugs I found for now:

- Cubrid doesn’t like `\`, it’s okay with `!`
- Oracle doesn’t have a quote char by default at all, moreover later pro_oci versions (in PHP 7.x) handle quoting so they won’t quote `\` while in PHP 5.x Yii will do it by itself (which leads to another bug, that inserts `\\` instead of `\` but it’s another issue…)
- sqlsrv driver doesn’t like preparing (it keeps telling about placeholders/params count mismatch) when using `ESCAPE`

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -45,6 +45,7 @@ Yii Framework 2 Change Log
 - Enh: Added `yii\di\Instance::__set_state()` method to restore object after serialization using `var_export()` function (silvefire)
 - Enh #13695: `\yii\web\Response::setStatusCode()` method now returns the Response object itself (kyle-mccarthy)
 - Enh #13698: `yii\grid\DataColumn` filter is automatically generated as dropdown list in case of `format` set to `boolean` (bizley)
+- Bug #13745: `SQLSTATE[HY093]: Invalid parameter number: parameter was not defined` in MSSQL and bug fixes in `yii\db\QueryBuilder::buildLikeCondition()` (sergeymakinen)
 
 
 2.0.11.2 February 08, 2017
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -45,7 +45,6 @@ Yii Framework 2 Change Log
 - Enh: Added `yii\di\Instance::__set_state()` method to restore object after serialization using `var_export()` function (silvefire)
 - Enh #13695: `\yii\web\Response::setStatusCode()` method now returns the Response object itself (kyle-mccarthy)
 - Enh #13698: `yii\grid\DataColumn` filter is automatically generated as dropdown list in case of `format` set to `boolean` (bizley)
-- Bug #13745: `SQLSTATE[HY093]: Invalid parameter number: parameter was not defined` in MSSQL and bug fixes in `yii\db\QueryBuilder::buildLikeCondition()` (sergeymakinen)
 
 
 2.0.11.2 February 08, 2017
--- a/framework/db/oci/QueryBuilder.php
+++ b/framework/db/oci/QueryBuilder.php
@@ -49,7 +49,29 @@ class QueryBuilder extends \yii\db\QueryBuilder
     /**
      * @inheritdoc
      */
-    protected $likeEscapeCharacter = '\\';
+    protected $likeEscapeCharacter = '!';
+    /**
+     * @inheritdoc
+     */
+    protected $likeEscapingReplacements = [
+        '%' => '!%',
+        '_' => '!_',
+        '!' => '!!',
+        // \ will be checked later Because yii\db\Schema::quoteValue may quote it
+    ];
+
+    /**
+     * @inheritDoc
+     */
+    public function init()
+    {
+        parent::init();
+        /*
+         * Different pdo_oci8 version may or may not implement PDO::quote(), so
+         * it may or may not fallback to yii\db\Schema::quoteValue() which quotes \.
+         */
+        $this->likeEscapingReplacements['\\'] = substr($this->db->quoteValue('\\'), 1, -1);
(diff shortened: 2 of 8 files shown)


framework/db/oci/QueryBuilder.php
```diff
@@ -49,7 +49,29 @@ class QueryBuilder extends \yii\db\QueryBuilder
...
+     * @inheritDoc
+     */
+    public function init()
+    {
+        parent::init();
+        /*
+         * Different pdo_oci8 version may or may not implement PDO::quote(), so
+         * it may or may not fallback to yii\db\Schema::quoteValue() which quotes \.
+         */
+        $this->likeEscapingReplacements['\\'] = substr($this->db->quoteValue('\\'), 1, -1);
```
samdark: That would trigger connecting to database. Don't see an issue with it but still...

framework/db/oci/QueryBuilder.php
```diff
@@ -49,7 +49,29 @@ class QueryBuilder extends \yii\db\QueryBuilder
...
-    protected $likeEscapeCharacter = '\\';
+    protected $likeEscapeCharacter = '!';
+    /**
+     * @inheritdoc
+     */
+    protected $likeEscapingReplacements = [
+        '%' => '!%',
+        '_' => '!_',
+        '!' => '!!',
+        // \ will be checked later Because yii\db\Schema::quoteValue may quote it
```
samdark: Suggestion: `\ is initialized in init() method since we need to choose replacement value based on [[\yii\db\Schema::quoteValue()]]`.

sergeymakinen: True. Sounds better.

framework/db/oci/QueryBuilder.php
sergeymakinen: Yeah, I thought about this… May be it’s safer to move this to `buildLikeCondition()` to delay connecting as long as possible. What do you think?

samdark: Yes, would be good.

samdark (approved):

SilverFire (approved):

framework/CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ Yii Framework 2 Change Log
 - Enh: Added `yii\di\Instance::__set_state()` method to restore object after serialization using `var_export()` function (silvefire)
 - Enh #13695: `\yii\web\Response::setStatusCode()` method now returns the Response object itself (kyle-mccarthy)
 - Enh #13698: `yii\grid\DataColumn` filter is automatically generated as dropdown list in case of `format` set to `boolean` (bizley)
+- Bug #13745: `SQLSTATE[HY093]: Invalid parameter number: parameter was not defined` in MSSQL and bug fixes in `yii\db\QueryBuilder::buildLikeCondition()` (sergeymakinen)
```
samdark: Isn't it fixing the issue that was introduced by a commit which haven't yet got into release? If so, changelog line is not required.

samdark:
Confirmed: https://github.com/yiisoft/yii2/issues/13745#issuecomment-285888783

framework/CHANGELOG.md
sergeymakinen: I have no idea. :) But it looks like a changelog entry could be removed in this case.

samdark:
Merged. Thanks!

PR #13763, opened 2017-03-13
Add ability to send PSR-7 streams from response

mikk150:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none

#11328

Makes large data dumps possible since now one can use PumpStream or similar


samdark:
Were large dumps impossible before?

mikk150:
Yes, if file you want to send was bigger than memory limit

samdark:
Isn't it already covered by http://www.yiiframework.com/doc-2.0/yii-web-response.html#xSendFile()-detail ?

mikk150:
Not really, it is almost http://www.yiiframework.com/doc-2.0/yii-web-response.html#sendStreamAsFile()-detail but with more functionality to not only send stream resources but rather PSR-7 streams that can be much more than just resources

samdark:
Aha, got it. So there are real benefits. Good.

mikk150:
I don't know yet if I should rewrite http://www.yiiframework.com/doc-2.0/yii-web-response.html#sendStreamAsFile()-detail to convert $handle into PSR-7 object and sending that(simplifies things)

samdark:
@yiisoft/core-developers thoughts?

composer.json
```diff
@@ -78,7 +78,8 @@
         "bower-asset/jquery": "2.2.*@stable | 2.1.*@stable | 1.11.*@stable | 1.12.*@stable",
         "bower-asset/jquery.inputmask": "~3.2.2 | ~3.3.3",
         "bower-asset/punycode": "1.3.*",
-        "bower-asset/yii2-pjax": "~2.0.1"
+        "bower-asset/yii2-pjax": "~2.0.1",
+        "psr/http-message": "^1.0"
```
cebe: as far as I see this would be an optional dependency which could be in require-dev, we only test for an interface, this does not need to exist so not everyone installing yii needs to install `psr/http-message`.

samdark: It's OK to have this dependency since it's very very lightweight one. No sub-dependencies, low number of files.

SamMousa:
Converting to PSR-7 stream by default is not an good option @mikk150, it would require a concrete implementation to be bundled or created.

samdark:
@SamMousa it's not converting to PSR-7 stream by default: 

```php
} elseif ($this->stream instanceof StreamInterface && $this->stream->isReadable()) {
```

samdark (approved):

mikk150:
No, it is not, I was just asking, if I should do it...

SamMousa:
I know!

samdark:
I think it should not in 2.0.

tests/framework/web/stubs/StreamStub.php
```diff
@@ -0,0 +1,245 @@
...
+            return $this->_customMetadata + stream_get_meta_data($this->_stream);
+        } elseif (isset($this->_customMetadata[$key])) {
+            return $this->_customMetadata[$key];
+        }
+
+        $meta = stream_get_meta_data($this->_stream);
+
+        return isset($meta[$key]) ? $meta[$key] : null;
+    }
+}
```
lynicidn: 🔥

samdark: ?

klimov-paul:
Closed in favor #14701

mikk150:
So, now when 2.1 is pretty much deleted, maybe we should reopen it??

mikk150:
@klimov-paul @samdark can we reopen it, because 2.1 is dead

terabytesoftw:
> @samdark can we reopen it, because 2.1 is dead

Will be added to the 2.2 branch.

PR #13766, opened 2017-03-13
Fixes #13704: Unique validator doesn't prefix attribute name with model's database table name

### LINKED ISSUE #13704: Unique validator doesn't prefix attribute name with model's database table name
Hi,
this rule is in `page` model rules :
`[['code'], 'unique']`
and code attribute exist in `post` model.

I join two table in `page::find()`:
```php
$query = new ActiveQuery(get_called_class()); $query->joinWith('post');
```
then ambiguous error accured on same `code` attribute.

```
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'code' in where clause is ambiguous 
The SQL being executed was: 
SELECT EXISTS(SELECT page.* FROM page LEFT JOIN post ON page.postId=post.id AND ((code='12')))
```

it seems in
`yii\validators\UniqueValidator::prepareConditions($targetAttribute, $model, $attribute)`
adding table name as prefix while preparing conditions solve this problem.

vladis84:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13704

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -148,7 +148,7 @@ class UniqueValidator extends Validator
      */
     private function getTargetClass($model)
     {
-        return is_null($this->targetClass) ? get_class($model) : $this->targetClass;
+        return $this->targetClass === null ? get_class($model) : $this->targetClass;
     }
 
     /**
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -144,7 +144,7 @@ class UniqueValidator extends Validator
 
     /**
      * @param Model $model the data model to be validated
-     * @return string Target class name.
+     * @return string Target class name
      */
     private function getTargetClass($model)
     {
@@ -248,7 +248,7 @@ class UniqueValidator extends Validator
             return $conditions;
         }
 
-        // Add table prefix for column.
+        // Add table prefix for column
         $targetClass = $this->getTargetClass($model);
         $tableName = $targetClass::tableName();
         $conditionsWithTableName = [];
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -10,6 +10,7 @@ namespace yii\validators;
 use Yii;
 use yii\base\Model;
 use yii\db\ActiveQuery;
+use yii\db\ActiveRecord;
 use yii\db\ActiveQueryInterface;
 use yii\db\ActiveRecordInterface;
 use yii\db\Query;
@@ -119,7 +120,7 @@ class UniqueValidator extends Validator
     public function validateAttribute($model, $attribute)
     {
         /* @var $targetClass ActiveRecordInterface */
-        $targetClass = $this->targetClass === null ? get_class($model) : $this->targetClass;
+        $targetClass = $this->getTargetClass($model);
         $targetAttribute = $this->targetAttribute === null ? $attribute : $this->targetAttribute;
         $rawConditions = $this->prepareConditions($targetAttribute, $model, $attribute);
         $conditions[] = $this->targetAttributeJunction === 'or' ? 'or' : 'and';
@@ -139,6 +140,15 @@ class UniqueValidator extends Validator
                 $this->addError($model, $attribute, $this->message);
             }
         }
+    }
+
+    /**
+     * @param Model $model the data model to be validated
+     * @return string Target class name.
+     */
+    private function getTargetClass($model)
+    {
+        return is_null($this->targetClass) ? get_class($model) : $this->targetClass;
     }
 
     /**
@@ -234,7 +244,20 @@ class UniqueValidator extends Validator
             $conditions = [$targetAttribute => $model->$attribute];
         }
 
-        return $conditions;
+        if (!$model instanceof ActiveRecord) {
+            return $conditions;
+        }
+
+        // Add table prefix for column.
+        $targetClass = $this->getTargetClass($model);
+        $tableName = $targetClass::tableName();
+        $conditionsWithTableName = [];
+        foreach ($conditions as $columnName => $columnValue) {
+            $columnNameWithPrefix = "{$tableName}.$columnName";
+            $conditionsWithTableName[$columnNameWithPrefix] = $columnValue;
+        }
+
+        return $conditionsWithTableName;
     }
 
     /**
--- a/tests/framework/validators/UniqueValidatorTest.php
+++ b/tests/framework/validators/UniqueValidatorTest.php
@@ -332,7 +332,7 @@ abstract class UniqueValidatorTest extends DatabaseTestCase
         $expected = ['val_attr_b' => 'test value b', 'val_attr_c' => 'test value a'];
         $this->assertEquals($expected, $result);
 
-        // Add table prefix for column name.
+        // Add table prefix for column name
         $model = Profile::findOne(1);
         $attribute = 'id';
         $targetAttribute = 'id';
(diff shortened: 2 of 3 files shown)


vladis84:
Лучше не смог придумать

framework/validators/UniqueValidator.php
```diff
@@ -142,6 +143,15 @@ public function validateAttribute($model, $attribute)
     }
 
     /**
+     * @param Model $model the data model to be validated
+     * @return string Target class name.
+     */
+    private function getTargetClass($model)
+    {
+        return is_null($this->targetClass) ? get_class($model) : $this->targetClass;
```
samdark: `$this->targetClass === null` should be used instead of `is_null`.

samdark (changes requested):

framework/validators/UniqueValidator.php
vladis84: Well I will correct

vladis84: Can it is better empty($this->targetClass)?

samdark: No.

vladis84:
Add tests. Fix remarks.

tests/framework/validators/UniqueValidatorTest.php
```diff
@@ -331,6 +331,31 @@ public function testPrepareParams()
...
+        $result = $this->invokeMethod(new UniqueValidator(), 'prepareConditions', [$targetAttribute, $model, $attribute]);
+        $expected = [Profile::tableName() . '.' . $attribute => $model->id];
+        $this->assertEquals($expected, $result);
+    }
+
+    public function testGetTargetClass()
+    {
+        // Filled property targetClass.
+        $validator = new UniqueValidator();
+        $validator->targetClass = 'target class';
```
dynasource: this is not a real example of a targetClass right ;). Please strive to create a unit test that is instantly understable and also stable for the future. This is likely to break when is check on a real classname.

tests/framework/validators/UniqueValidatorTest.php
```diff
@@ -331,6 +331,31 @@ public function testPrepareParams()
...
+        $model = new FakedValidationModel();
+        $actualTargetClass = $this->invokeMethod($validator, 'getTargetClass', [$model]);
+        $this->assertEquals('target class', $actualTargetClass);
+
+        // Not filled property targetClass.
```
dynasource: you are preparing a new test scenario here. Please move this to a dedicated test method with a corresponding test name

dynasource (changes requested):

samdark (approved):

tests/framework/validators/UniqueValidatorTest.php
vladis84: Fixed

tests/framework/validators/UniqueValidatorTest.php
vladis84: Fixed

dynasource (approved):
looks better. Thanks

framework/validators/UniqueValidator.php
```diff
@@ -234,7 +244,20 @@ private function prepareConditions($targetAttribute, $model, $attribute)
...
 
-        return $conditions;
+        if (!$model instanceof ActiveRecord) {
+            return $conditions;
+        }
+
+        // Add table prefix for column.
+        $targetClass = $this->getTargetClass($model);
+        $tableName = $targetClass::tableName();
+        $conditionsWithTableName = [];
```
arogachev: IMO `$conditionsWithTableName variable` can be named better and probably shorter.

framework/validators/UniqueValidator.php
```diff
@@ -234,7 +244,20 @@ private function prepareConditions($targetAttribute, $model, $attribute)
...
+        foreach ($conditions as $columnName => $columnValue) {
+            $columnNameWithPrefix = "{$tableName}.$columnName";
```
arogachev: The same applies to `$columnNameWithPrefix` variable. Maybe use `$prefixedColumnName` at least or even `$prefixedColumn` instead? Current name feels pretty verbose.

framework/validators/UniqueValidator.php
```diff
@@ -234,7 +244,20 @@ private function prepareConditions($targetAttribute, $model, $attribute)
...
-        return $conditions;
+        if (!$model instanceof ActiveRecord) {
+            return $conditions;
+        }
+
+        // Add table prefix for column.
+        $targetClass = $this->getTargetClass($model);
+        $tableName = $targetClass::tableName();
+        $conditionsWithTableName = [];
+        foreach ($conditions as $columnName => $columnValue) {
```
arogachev: Maybe use just `$column => $value` instead? These are loop's internal variables, also I don't see any coincidence with other similar variables or anything that can confuse understanding of a shorter variation.

framework/validators/UniqueValidator.php
arogachev: Also why not modify existing `$conditions` array instead of creating new one? I don't see any conditions in `foreach` loop.

arogachev (changes requested):

samdark:
@vladis84 was it closed by accident?

vladis84:
Deliberately closed. 
You do not need help.

samdark:
Hm? Why? We do need help and we appreciate contributing.

framework/validators/UniqueValidator.php
SilverFire: `$columnName` should be quoted with `[$coumnName]`, shouldn't it?

SilverFire (changes requested):

samdark:
Merged with some additional adjustments https://github.com/yiisoft/yii2/commit/3c1f3e20cfa260d1484fece29b255103b3bb6f5e. Thanks for contributing and sorry if we did something wrong. All these requests from core team members are to make framework better, not to offend contributors.

samdark:
> $columnName should be quoted with [$coumnName], shouldn't it?

Nope. It's already prepared in this case.

framework/validators/UniqueValidator.php
vladis84: https://habrahabr.ru/post/136835/

framework/validators/UniqueValidator.php
samdark: Nope. It's already prepared in this case.

framework/validators/UniqueValidator.php
samdark: @arogachev how? Keys are modified in this case, not value.

arogachev: @vladis84, @samdark - OK, I missed that. But I still think that `$conditionsWithTableName` variable can be just something like `$prefixedConditions`. In this context it's clear what it means. Also it's shorter and more readable IMO. The same applies to other related variables such as `$columnNameWithPrefix` (can be just `$prefixedColumn`). What do you think?

arogachev:
@vladis84 wrote:

> Deliberately closed.
> You do not need help.

That's a weak argument.

@samdark wrote:

> sorry if we did something wrong

We did nothing wrong.

@samdark wrote:

> All these requests from core team members are to make framework better, not to offend contributors.

Exactly. You should respect framework's rules and adequately accept criticism. PR review is an important part of workflow for pretty much every program nowadays (especially for those that are intended for wide, non-personal use). Also PR review can contain **suggestions** for discussion. This does not mean that you are wrong and must immediately fix something, but for other core members and contributors to discuss. The final goal is to make things better and easier to maintain.

Sorry if it sounds straight - but If you can't deal with it - maybe it's a better option to not create PRs at all or just create issues to designate problem for example. Also I do not see any offence (even implicit) in this PR.

samdark:
@arogachev the matter is already settled.

PR #13776, opened 2017-03-15
Add default precision and scale for decimal column type (mssql)

arturf:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no

At the moment column definition for decimal type did not work properly in some cases. For example in migrations `$this->decimal(8,4)` will generate only `decimal` part.


arturf:
@samdark I fix `QueryBuilderTest` and squash commit

samdark:
> For example in migrations $this->decimal(8,4) will generate only decimal part.

Would that be fixed with these changes?

arturf:
> Would that be fixed with these changes?

Yes

Please, see https://github.com/yiisoft/yii2/blob/master/framework/db/QueryBuilder.php#L738

With current typeMap definition (without default scale) in yii\db\mssql\QueryBuilder, precisions from params wil be replaced. This PR fix this.

samdark:
I see. Are tests currently checking it?

arturf:
> Are tests currently checking it?

for decimal field - yes. But test contain some incorrect expected results for other types that don't equal to typeMap. For example boolean is mapped to bit, but test expexts for tinyint. Add fix in this PR?

samdark:
Separate pull request will be better.

samdark:
For this pull request we need a changelog.

arturf:
> For this pull request we need a changelog.

done

framework/db/mssql/QueryBuilder.php
```diff
@@ -35,7 +35,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_BIGINT => 'bigint',
         Schema::TYPE_FLOAT => 'float',
         Schema::TYPE_DOUBLE => 'float',
-        Schema::TYPE_DECIMAL => 'decimal',
+        Schema::TYPE_DECIMAL => 'decimal(18,0)',
```
dynasource: please use decimal(10,0) for consistency troughout the different DBMSes. See for Mysql https://github.com/yiisoft/yii2/blob/master/framework/db/mysql/QueryBuilder.php#L38

dynasource (changes requested):

framework/db/mssql/QueryBuilder.php
arturf: and setting it as (10,0) will change current behaviour

markmarco16: Maybe (5,2) is a good value, because (10,0) is similar to INT(10), don't have decimal

dynasource: alright. Lets keep it non BC

dynasource (approved):

samdark (approved):

samdark:
Merged. Thanks!

arogachev:
Thanks! One small remark - please create dedicated branches for PRs (do not use master), please see [Point 3](https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md#3-create-a-new-branch-for-your-feature-based-on-the-current-yii-master-branch) in our [Git Worklow](https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md) document.

framework/db/mssql/QueryBuilder.php
cebe: > @dynasource this change for mssql only and (18, 0) is default values for SQL Server.
> Other DBMS have own type mapping

This is correct, entries in these fields should always use the default by the dbms.

samdark:
Not using master allows you to work on multiple pull requests at the same time. Using master you can't do it.

PR #13783, opened 2017-03-15
Added JSON type to MySQL and PostgreSQL schema

undefinedor:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13019 #10436

### DIFF
--- a/framework/db/mysql/QueryBuilder.php
+++ b/framework/db/mysql/QueryBuilder.php
@@ -43,6 +43,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_BINARY => 'blob',
         Schema::TYPE_BOOLEAN => 'tinyint(1)',
         Schema::TYPE_MONEY => 'decimal(19,4)',
+        Schema::TYPE_JSON => 'json'
     ];
 
 
--- a/framework/db/pgsql/QueryBuilder.php
+++ b/framework/db/pgsql/QueryBuilder.php
@@ -67,6 +67,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_BINARY => 'bytea',
         Schema::TYPE_BOOLEAN => 'boolean',
         Schema::TYPE_MONEY => 'numeric(19,4)',
+        Schema::TYPE_JSON => 'json',
     ];
 
     /**
--- a/framework/db/pgsql/Schema.php
+++ b/framework/db/pgsql/Schema.php
@@ -108,7 +108,7 @@ class Schema extends \yii\db\Schema
         'unknown' => self::TYPE_STRING,
 
         'uuid' => self::TYPE_STRING,
-        'json' => self::TYPE_STRING,
+        'json' => self::TYPE_JSON,
         'jsonb' => self::TYPE_STRING,
         'xml' => self::TYPE_STRING,
     ];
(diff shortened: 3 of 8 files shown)


Renkas:
JSON data type is available from MySQL 5.7.8. What happens if I use 5.6?

undefinedor:
@Renkas  throw a exception . Mysql 5.6 does't support JSON.

undefinedor:
@Renkas  throw a exception . Mysql 5.6 does't support JSON.

RSalo:
@Renkas i think Yii should support that, but checking availability this functionality fall on users shoulders

samdark:
Throwing exception sounds OK to me in this case. MySQL 5.6 is 2013.

dynasource (changes requested):
shouldnt the added support for JSON be a little bigger than what this PR contains?

samdark:
@dynasource what's missing?

sergeymakinen:
Even basic JSON support is good but it'd be good to reflect in the title that the PR adds a JSON support to the column schema builder to not get hopes up.

dynasource:
> Even basic JSON support is good but it'd be good to reflect in the title that the PR adds a JSON support to the column schema builder to not get hopes up.

exactly

framework/db/pgsql/Schema.php
```diff
@@ -108,7 +108,7 @@ class Schema extends \yii\db\Schema
         'unknown' => self::TYPE_STRING,
 
         'uuid' => self::TYPE_STRING,
-        'json' => self::TYPE_STRING,
+        'json' => self::TYPE_JSON,
         'jsonb' => self::TYPE_STRING,
```
cebe: same should apply for jsonb, the difference is in the storage of the data, not in the data format. https://www.postgresql.org/docs/9.4/static/datatype-json.html

cebe (changes requested):
needs a changelog, and ideally a test also.

jeesus:
Any estimate when we could see this in stable release?

samdark:
No. Without tests it won't make it to stable release.

framework/db/mysql/QueryBuilder.php
```diff
@@ -43,6 +43,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_BINARY => 'blob',
         Schema::TYPE_BOOLEAN => 'tinyint(1)',
         Schema::TYPE_MONEY => 'decimal(19,4)',
+        Schema::TYPE_JSON => 'json'
```
Tigrov: It's better to use `Schema::TYPE_JSON => 'jsonb'`
https://www.postgresql.org/docs/current/static/datatype-json.html

framework/db/pgsql/QueryBuilder.php
```diff
@@ -67,6 +67,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
         Schema::TYPE_BINARY => 'bytea',
         Schema::TYPE_BOOLEAN => 'boolean',
         Schema::TYPE_MONEY => 'numeric(19,4)',
+        Schema::TYPE_JSON => 'json',
```
Tigrov: It's better to use `Schema::TYPE_JSON => 'jsonb'`
https://www.postgresql.org/docs/current/static/datatype-json.html

acim:
What is going on with this?

samdark:
Nothing. With tests missing there's no chance for merging it.

nguyenbs:
Hi,
I see the test is already passed, how about next plan?

samdark:
Existing tests did but there are no tests for the code added.

samdark:
Closed in favor of https://github.com/yiisoft/yii2/pull/15560

PR #13784, opened 2017-03-16
$defaultOptions in favor of dependency injection

dan-developer:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/widgets/InputWidget.php
+++ b/framework/widgets/InputWidget.php
@@ -12,6 +12,7 @@ use yii\base\Widget;
 use yii\base\Model;
 use yii\base\InvalidConfigException;
 use yii\helpers\Html;
+use yii\helpers\ArrayHelper;
 
 /**
  * InputWidget is the base class for widgets that collect user inputs.
@@ -59,6 +60,11 @@ class InputWidget extends Widget
      */
     public $value;
     /**
+     * @var array the default HTML attributes for the input tag.
+     * @see \yii\helpers\Html::renderTagAttributes() for details on how attributes are being rendered.
+     */
+    public $defaultOptions = [];
+    /**
      * @var array the HTML attributes for the input tag.
      * @see \yii\helpers\Html::renderTagAttributes() for details on how attributes are being rendered.
      */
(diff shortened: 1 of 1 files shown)


samdark:
Could you elaborate about the change?

AndersonDan:
In large systems code reuse can be improved with a standard, so you can change everything with a simple configuration, a hypothetical example:

```php
\Yii::$container->set('yii\widgets\ActiveForm', [
    'defaultOptions' => [
        'autocomplete' => 'off',
    ],
]);

ActiveForm::begin([
    'options' => [
        'id' => 'test-form',
        //'autocomplete' => 'off', // It is not necessary to repeat the configuration again. If at some point resolve to return all form to be 'autocomplete' => 'on', just change it in dependency injection.
    ],
]);
```

samdark:
Technically I understand why it's made like that. When overriding `options` via `::widget()` call, it replaces options defined via DI globally. I can't say I like additional property though.

samdark:
Need @yiisoft/core-developers opinions.

AndersonDan:
Exactly.

Another example would be: https://github.com/kartik-v/yii2-krajee-base/pull/77

dynasource:
> I can't say I like additional property though.

agree. The usecase is so common that I prefer a solution to be handled centrally so that it can be reused everywhere.

samdark:
@dynasource any ideas?

dynasource:
by changing array_merge to ArrayHelper::merge() in https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php#L168

Then both DI & custom config arrays can be used side by side. Of course the side effect is that DI is not overwritten, but since you have the power to configure this in the first place, this could be tackled. Its a tradeoff. IMO, ArrayHelper::merge() is more useful in practice.

samdark:
I agree that such merge would be a better solution but I'm afraid that would change too much. Developers may rely on the fact that it replaces properties now.

dynasource:
agree

samdark:
Would be designed differently in 3.0.

PR #13787, opened 2017-03-16
Allow limiting the length of the sql string that will be printed

thiagotalma:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

When we run a migration with a very large view, for example, it is bad to follow the migration progress on the screen.

### DIFF
--- a/framework/db/Migration.php
+++ b/framework/db/Migration.php
@@ -63,7 +63,8 @@ class Migration extends Component implements MigrationInterface
     public $db = 'db';
 
     /**
-     * @var int length of the string that will be printed on the screen.
+     * @var int length of the printed SQL. Useful for reduction of long statements and making console output more 
+     * compact and readable.
      */
     public $printedSqlLength;
 
--- a/framework/db/Migration.php
+++ b/framework/db/Migration.php
@@ -191,7 +191,6 @@ class Migration extends Component implements MigrationInterface
     public function execute($sql, $params = [])
     {
         $printedSql = $sql;
-
         if ($this->printedSqlLength !== null) {
             $printedSql = StringHelper::truncate($sql, $this->printedSqlLength, '[... hidden]');
         }
--- a/framework/db/Migration.php
+++ b/framework/db/Migration.php
@@ -61,6 +61,11 @@ class Migration extends Component implements MigrationInterface
      */
     public $db = 'db';
 
+    /**
+     * @var int length of the string that will be printed on the screen.
+     */
+    public $printSqlLength = 0;
+
 
     /**
      * Initializes the migration.
@@ -180,12 +185,11 @@ class Migration extends Component implements MigrationInterface
      * This method executes the specified SQL statement using [[db]].
      * @param string $sql the SQL statement to be executed
      * @param array $params input parameters (name => value) for the SQL execution.
-     * @param int $printSqlLength length of the string that will be printed on the screen.
      * See [[Command::execute()]] for more details.
      */
-    public function execute($sql, $params = [], $printSqlLength = 0)
+    public function execute($sql, $params = [])
     {
-        $sqlPrint = ($printSqlLength === 0 || strlen($sql) <= $printSqlLength) ? $sql : substr($sql, 0, $printSqlLength) . "(...hidden)";
+        $sqlPrint = ($this->printSqlLength === 0 || strlen($sql) <= $this->printSqlLength) ? $sql : substr($sql, 0, $this->printSqlLength) . "[...hidden]";
         echo "    > execute SQL: $sqlPrint ...";
         $time = microtime(true);
         $this->db->createCommand($sql)->bindValues($params)->execute();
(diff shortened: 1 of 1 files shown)


samdark:
Could you elaborate about use case?

thiagotalma:
I have a database view that takes up 4 pages of the screen when it is executed in a migration.
It would be very useful to be able to configure to only print the first 200 characters.

samdark:
I suggest making it a global class property instead of a method parameter.

samdark (approved):

framework/db/Migration.php
```diff
@@ -184,7 +189,8 @@ public function safeDown()
      */
     public function execute($sql, $params = [])
     {
-        echo "    > execute SQL: $sql ...";
+        $sqlPrint = ($this->printSqlLength === 0 || strlen($sql) <= $this->printSqlLength) ? $sql : substr($sql, 0, $this->printSqlLength) . "[...hidden]";
```
dynasource: difficult line to read. Deserves a multiple lines.

dynasource (approved):

framework/db/Migration.php
rob006: You probably should use `mb_strlen()` and `mb_substr()` here.

samdark (changes requested):
See comments.

framework/db/Migration.php
```diff
@@ -184,7 +189,13 @@ public function safeDown()
      */
     public function execute($sql, $params = [])
     {
-        echo "    > execute SQL: $sql ...";
+        $sqlPrint = $sql;
+
+        if ($this->printSqlLength !== 0 && mb_strlen($sql) > $this->printSqlLength) {
```
samdark: `mb_strlen` should have encoding specified.

framework/db/Migration.php
```diff
@@ -184,7 +189,13 @@ public function safeDown()
...
+            $sqlPrint = mb_substr($sql, 0, $this->printSqlLength) . "[...hidden]";
```
samdark: Same here. Encoding should be specified.

samdark (changes requested):

framework/db/Migration.php
thiagotalma: Which one should I use?
I searched for a pattern in the framework and did not find it.
In fact, most cases are without encoding.

samdark: mb_strlen($from, Yii::$app->charset)

rob006: `StringHelper::truncate()` may be better option in this case.

samdark: Actually, yes.

framework/db/Migration.php
```diff
@@ -61,6 +61,11 @@ class Migration extends Component implements MigrationInterface
      */
     public $db = 'db';
 
+    /**
+     * @var int length of the string that will be printed on the screen.
+     */
+    public $printSqlLength = 0;
```
bashkarev: - `-1` - show all
- `0` - not display 
?

dynasource: agree that 0 is counter intuitive

rob006:
```php
public $printSqlLengthLimit;
```

If `null`, there is no limit. If `0`, query id not displayed.

framework/db/Migration.php
```diff
@@ -61,6 +62,10 @@ class Migration extends Component implements MigrationInterface
      */
     public $db = 'db';
 
+    /**
+     * @var int length of the string that will be printed on the screen.
+     */
+    public $printSqlLength;
```
arogachev: I suggest using `$printedSqlLength` or `$displayedSqlLength` instead.

framework/db/Migration.php
```diff
@@ -184,7 +189,13 @@ public function safeDown()
      */
     public function execute($sql, $params = [])
     {
-        echo "    > execute SQL: $sql ...";
+        $sqlPrint = $sql;
```
arogachev: `$printedSql` is better, more clear and correct form IMO.

framework/db/Migration.php
```diff
@@ -184,7 +189,13 @@ public function safeDown()
...
+
+        if ($this->printSqlLength !== null) {
+            $sqlPrint = StringHelper::truncate($sql, $this->printSqlLength, '[...hidden]');
```
arogachev: Nitpicking, but maybe add a space after dots: `[... hidden]`.

arogachev (changes requested):

thiagotalma:
@arogachev done!

framework/db/Migration.php
thiagotalma: @samdark done!

samdark (approved):

arogachev:
@thiagotalma Thanks! 👍 Made small corrections. Now I think it's ready for merge. Others,, please check.

arogachev (approved):

samdark:
Merged with adjustments. Thanks!

PR #13793, opened 2017-03-17
Compatibility with PHPUnit 6.x

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -18,4 +18,9 @@ require_once(__DIR__ . '/../framework/Yii.php');
 
 Yii::setAlias('@yiiunit', __DIR__);
 
+if (!class_exists('PHPUnit_Framework_TestCase') && class_exists('PHPUnit\Framework\TestCase')) {
+    // compatibility with PHPUnit 6.x
+    class PHPUnit_Framework_TestCase extends \PHPUnit\Framework\TestCase {}
+}
+
 require_once(__DIR__ . '/TestCase.php');


samdark:
Seems it's not required since PHPUnit 4.8: https://github.com/sebastianbergmann/phpunit/tree/4.8/src/ForwardCompatibility

tests/bootstrap.php
```diff
@@ -18,4 +18,9 @@
 
 Yii::setAlias('@yiiunit', __DIR__);
 
+if (!class_exists('PHPUnit_Framework_TestCase') && class_exists('PHPUnit\Framework\TestCase')) {
```
samdark: How about moving it to `compatibility.php`? I'm currently working on tests and implemented custom assertion which needs another BC hack because of `extends \PHPUnit_Framework_Constraint`...

samdark:
```php
if (!class_exists('PHPUnit_Framework_Constraint') && class_exists('PHPUnit\Framework\Constraint\Constraint')) {
    class PHPUnit_Framework_Constraint extends \PHPUnit\Framework\Constraint\Constraint {}
}
```

samdark:
Merged with adjustments: https://github.com/yiisoft/yii2/commit/01dc29705fa31d05cc177f6724cbccf6b52da91a

samdark:
Thanks for taking care of it.

Issue #13770, opened 2017-03-14
Menu::renderItems() throws exception if active item's class is defined as an array.


Pikseli:
### What steps will reproduce the problem?

Create a menu with an active item having class defined as an array. Exception is also thrown if item is first or last.

```php
echo \yii\widgets\Menu::widget([
    'activeCssClass' => 'item--active',
    'items' => [
        [
            'label' => 'Foo',
            'url' => '/foo',
            'active' => true,
            'options' => [
                'class' => [
                    'item',
                    'item--foo'
                ]
            ]
        ]
    ]
]);
```
### What is the expected result?

```html
<ul>
    <li class="item item--foo item--active"><a href="/foo">Foo</a></li>
</ul>
```

### What do you get instead?

PHP Notice – yii\base\ErrorException

Array to string conversion

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.1.0
| Operating system | macOS Sierra 10.12.3

Kolyunya:
@Pikseli are you sure it's supposed to work as you expected? Could you please provide a reference to the docs?

Kolyunya:
It's currently [no supported](https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1836). Probably worth implementing.

Kolyunya:
@Pikseli seems like it's [not supposed to work](http://www.yiiframework.com/doc-2.0/yii-widgets-menu.html#$itemOptions-detail) as you expect it. Should not throw exception though.

>options: array, optional, the HTML attributes for the menu container tag.

samdark:
Yes. That would be consistent with `Html::tag()` with `class` option. Should be applied consistently through framework.

PR #13796, opened 2017-03-17
Enhance menu widget item class rendering

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13770


framework/widgets/Menu.php
```diff
@@ -210,9 +210,9 @@ protected function renderItems($items)
             }
             if (!empty($class)) {
                 if (empty($options['class'])) {
-                    $options['class'] = implode(' ', $class);
```
Kolyunya: `$class` is going to be imploded by `renderTagAttributes()`.

framework/widgets/Menu.php
```diff
@@ -210,9 +210,9 @@ protected function renderItems($items)
...
+                    $options['class'] = $class;
                 } else {
-                    $options['class'] .= ' ' . implode(' ', $class);
```
Kolyunya: `$options['class']` is an array so it's not concatenated to `$class` but is merged with it.

tests/framework/widgets/MenuTest.php
```diff
@@ -196,6 +196,91 @@ public function testActiveItemClosure()
...
+                    'label'  => 'item2',
+                    'url'    => '#',
+                    'options' => [
+                        'class' => [
+                            'another-class',
+                            'other--class',
+                            'two classes',
+                        ],
+                    ],
+                ],
```
Kolyunya: A test for the case when `$class` is empty.

framework/widgets/Menu.php
```diff
@@ -210,9 +210,12 @@ protected function renderItems($items)
...
+                    if (!is_array($options['class'])) {
+                        $options['class'] = [$options['class']];
```
Kolyunya: BC support. Probably a subject for removal.

framework/widgets/Menu.php
```diff
@@ -210,9 +210,12 @@ protected function renderItems($items)
...
+                    }
+                    $options['class'] = ArrayHelper::merge($class, $options['class']);
```
Kolyunya: `$options['class']` is an array so it's not concatenated to `$class` but is merged with it.

tests/framework/widgets/MenuTest.php
```diff
@@ -196,6 +196,121 @@ public function testActiveItemClosure()
...
+            ]
+        ]);
+
+        $expected = <<<HTML
+<ul><li class="item--active someclass"><a href="#">item1</a></li>
+<li class="another-class other--class two classes"><a href="#">item2</a></li></ul>
+HTML;
+        $this->assertEqualsWithoutLE($expected, $output);
+
+        $output = Menu::widget([
```
Kolyunya: BC test.

framework/widgets/Menu.php
```diff
@@ -210,9 +210,12 @@ protected function renderItems($items)
             }
             if (!empty($class)) {
                 if (empty($options['class'])) {
-                    $options['class'] = implode(' ', $class);
+                    $options['class'] = $class;
                 } else {
-                    $options['class'] .= ' ' . implode(' ', $class);
+                    if (!is_array($options['class'])) {
```
samdark: Could be done as `$options['class'] = (array) $options['class'];`.

samdark: Or even `$options['class'] = ArrayHelper::merge($class, (array) $options['class']);`.

Kolyunya: I feel like the latter is an over complication of the merge statement. I used the former approach and added a comment clarifying why this is needed. What do you think?

samdark (changes requested):
Code itself looks good. I'd break test method into three separate ones named after what they're testing.

Kolyunya:
Refactored tests.

samdark:
https://en.wikipedia.org/wiki/Pivot_table

framework/widgets/Menu.php
samdark: It's OK.

samdark (approved):

framework/widgets/Menu.php
cebe: there is `Html::addCssClass()` which could be used here as far as I see.

dynasource (approved):

framework/widgets/Menu.php
Kolyunya: Thank. That simplifies things up. Also altered resulting class order in tests.

SilverFire (approved):

samdark:
Merged. Thanks!

Kolyunya:
Glad to help!

Issue #8084, opened 2015-04-15
Allow to change layout of mail for one mail (not global).


Alex-Bond:
For example:

```
Yii::$app->mailer->compose( "@mails/coolMail", [ 'model' => $model ], ['htmlLayout'=>'@mails/layout/anotherLayout'] )
***
                         ->send();
```

Because creating new component instance for each layout is not good.

trntv:
I think you should create this issue here - https://github.com/yiisoft/yii2-swiftmailer/issues

ruskid:
Yep. I also need the possibility to change the layout for a single mail. But there is a better work-around then creating new component for each layout.

You can nest layouts like a normal ones:
```php
<?php $this->beginContent('@common/mail/layouts/single.php') ?>

<?php $this->endContent() ?>
```

dynasource:
issue confirmed

gustavovendramini:
The work-around do not work here, so I fix as follows:

```php
$mailer = Yii::$app->getMailer();
$mailer->htmlLayout = 'layouts/anotherLayout';
```

klimov-paul:
Resolved by #13809

PR #13809, opened 2017-03-20
Fix #9260 introduce `yii\mail\Template` class, allowing view rendering in isolation

### LINKED ISSUE #9260: \yii\mailer\BaseMailer can't be extended due to bad design choices
The problem is class can not be extended due to bad design. Example:

``` php
class MyMailer extends \yii\mailer\BaseMailer
{
    // ...
    public function compose($view = null, array $params = [])
    {
        // my own logic, but i want use $this->render() in my implementation too
        $this->render($view['html'], $params, $this->htmlLayout);
    }
}
```

In child classes `$message` in layouts will always be `null` and you can't do anything with that. Look at original class compose() method:

``` php
    public function compose($view = null, array $params = [])
    {
        if (!array_key_exists('message', $params)) {
            $params['message'] = $message;
        }

        // ...
        $this->_message = $message;
        // render call
        $this->_message = null;
        // ...
    }

    public function render($view, $params = [], $layout = false)
    {
        // $this->_message used there and you can't set its value from child classes
    }
```

Its what i call voodoo magic which make this class methods complettely unextensible.

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #9260, #8084

Introduce `yii\mail\Template` class, allowing view rendering in isolation.
This allows access to `yii\mail\MessageInterface` via `$this->context->message` inside the view all the time,
including layouts, and use `message` key for rendering params without any conflicts.


framework/UPGRADE.md
```diff
@@ -57,6 +57,12 @@ Upgrade from Yii 2.0.x
   providing ability to setup custom mail headers. Make sure your provide implementation for those methods, while
   creating your own mailer solution.
 
+* During mail view rendering `$message` variable is no longer set by default to be an instance of `yii\mail\MessageInterface`,
```
samdark: Why? In the original issue it was about layout only, wasn't it?

klimov-paul: There is a related issue #4147 , which is closed.
At 2.0.x there is a problem in views params conflict: if `$params`does NOT hold `message` key it will be set by message instance - otherwise message instance is unavailable during rendering.

Placing `message` to the view context ensures it is always available and does not conflict with any view param.

samdark: It makes sense but syntax isn't as good as plain `$message`.

tests/framework/mail/TemplateTest.php
```diff
@@ -0,0 +1,141 @@
+<?php
+/**
+ * @link http://quartsoft.com/
+ * @copyright Copyright &copy; 2015 QuartSoft ltd.
+ * @license http://www.quartsoft.com/license/
```
samdark: Is it intentional?

samdark (changes requested):

tests/framework/mail/TemplateTest.php
klimov-paul: No, it should be removed

klimov-paul:
Also fixes #8084

samdark:
How about the case when templates are to be stored/edited in database?

klimov-paul:
> How about the case when templates are to be stored/edited in database?

This is the matter of view rendering.
Still you can create your own implementation of the `Template` and setup it via `Mailer::$templateConfig` allowing to use database info during message composition.

samdark:
> Still you can create your own implementation of the Template

The interface of Template is tied to a view file name.

klimov-paul:
> The interface of Template is tied to a view file name.

It is tied to `view name` - it does not mean it should not always be a file.

samdark:
Oh, right.

klimov-paul:
Still `MessageComposer` (or `MessageBuilder`) layer may be extracted for better support.

samdark:
How about extracting an interface for this case since now compose() and render() are logically an interface and public properties are extra config for particular implementation.

klimov-paul:
Method `compose()` is already belong to the interface - `yii\mail\MailerInterface`.
Particular mailer may provide its own implementation for it, which may not use any rendering at all.

samdark:
I'm talking about `compose()` method of Template class:

```php
class Template extends Object implements ViewContextInterface
{
```

klimov-paul:
`yii\mail\Composer` extracted

samdark:
That's better but if you want to replace it, it's still not clear which methods you should implement. As far as I can see, only `Composer::compose($message, $view, $params)` is required so I'd make an interface for it and hint against it.

klimov-paul:
That is debatable. While extra interface can be justified here - it is not common for Yii.
We have many 'servant' classes similar to `yii\mail\Composer`. For example: 

- `yii\db\QueryBuilder`
- `yii\data\Sort`
- `yii\data\Pagination`
 - `yii\grid\Column`

I can not see benefit of intorduction `yii\mail\ComposerInterface` especially, while particular `yii\mail\MailerInterface` implementation may not use it at all.

samdark:
OK.

samdark (approved):

framework/mail/Template.php
```diff
@@ -0,0 +1,131 @@
...
+     */
+    public $view;
+    /**
+     * @var string path to the directory containing view files.
+     */
+    public $viewPath;
+    /**
+     * @var string|array name of the view to use as a template. The value could be:
+     *
+     * - a string that contains either a view name or a path alias for rendering HTML body of the email.
```
arogachev: Maybe use references to similar Composer's properties to avoid descriptions' repetition?

tests/framework/mail/BaseMailerTest.php
```diff
@@ -45,18 +43,19 @@ protected function getTestFilePath()
...
      */
     protected function createTestMailComponent()
     {
-        $component = new Mailer();
-        $component->viewPath = $this->getTestFilePath();
-
-        return $component;
+        return new TestMailer([
+            'composer' => [
+                'viewPath' => $this->getTestFilePath()
```
arogachev: Should we add comma after last array element if array elements are split to separate lines? Even in tests? [According core code style guide section](https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md#associative) is a bit unclear about it.

tests/framework/mail/ComposerTest.php
```diff
@@ -0,0 +1,65 @@
...
+use yii\base\View;
+use yii\mail\Template;
+use yiiunit\TestCase;
+
+/**
+ * @group mail
+ */
+class ComposerTest extends TestCase
+{
+    // Tests :
```
arogachev: This comment seems redundant especially there is nothing except 3 tests.

arogachev (changes requested):

tests/framework/mail/BaseMailerTest.php
samdark: Yes, we should.

framework/mail/Composer.php
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * @param array|\yii\base\View $view view instance or its array configuration that will be used to
+     * render message bodies.
+     * @throws InvalidConfigException on invalid argument.
+     */
+    public function setView($view)
+    {
+        if (!is_array($view) && !is_object($view)) {
+            throw new InvalidConfigException('"' . get_class($this) . '::view" should be either object or configuration array, "' . gettype($view) . '" given.');
```
klimov-paul: Disagreed: I can see no problem in this - it does not violate any code style rules.
Method `setView()` has been just moved from `BaseMailer` as it was before.

framework/mail/Template.php
klimov-paul: Done for `htmlLayout` and `textLayout`, for other properties descriptions vary and can not be just referenced.

tests/framework/mail/BaseMailerTest.php
klimov-paul: done

framework/UPGRADE.md
```diff
@@ -57,6 +57,15 @@ Upgrade from Yii 2.0.x
   providing ability to setup custom mail headers. Make sure your provide implementation for those methods, while
   creating your own mailer solution.
 
+* During mail view rendering the `$message` variable is no longer set by default to be an instance of `yii\mail\MessageInterface`. Instead it is available via `$this->context->message` expression.
+
+* `yii\mail\BaseMailer::render()` method has been removed. Make sure you do not use it anywhere in your program.
+  Mail view rendering is now encapsulated into `yii\mail\Template` class.
+
+* Properties `view`, `viewPath`, `htmlLayout` and `textLayout` have been removed from `yii\mail\BaseMailer`.
```
klimov-paul: done

PR #13815, opened 2017-03-21
Enhance web response

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13813

### DIFF
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -292,13 +292,13 @@ class Response extends \yii\base\Response
 
     /**
      * Sets the response status code based on the exception.
-     * @param \Exception $e
+     * @param \Exception|\Error $e
      * @throws InvalidParamException if the status code is invalid.
      * @return $this the response object itself
      *
      * @since 2.0.12
      */
-    public function setStatusCodeByException(\Exception $e)
+    public function setStatusCodeByException($e)
     {
         if ($e instanceof HttpException) {
             $this->setStatusCode($e->statusCode);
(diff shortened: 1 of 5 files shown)


framework/web/Response.php
```diff
@@ -292,13 +292,13 @@ public function setStatusCode($value, $text = null)
...
      * Sets the response status code based on the exception.
-     * @param \Exception $e
+     * @param \Throwable $e
      * @throws InvalidParamException if the status code is invalid.
      * @return $this the response object itself
      *
      * @since 2.0.12
      */
-    public function setStatusCodeByException(\Exception $e)
+    public function setStatusCodeByException(\Throwable $e)
```
samdark: Type hint needs to be removed since `\Throwable` is PHP 7 only.

samdark (changes requested):

framework/web/Response.php
Kolyunya: Fixed that. Do we revert docs fixes?

Kolyunya: Updated docs.

samdark (approved):

dynasource (approved):

samdark:
Merged. Thanks!

Kolyunya:
Don't you think that the changelog is a bit misleading? This seems like a new feature to me, not a compatibility fix according to the docs.

samdark:
@Kolyunya you mean `UPGRADE`?

Kolyunya:
I mean that according to the docs only `Exception` was supposed to be accepted by that method. Passing `Error` was not even documented. And now we added new functionality. This seems to be a new feature not a compatibility fix. I'm not sure if this is correct though.

samdark:
Right. https://github.com/yiisoft/yii2/commit/da3bb5aa429f3f75af88694b25bb1e5b1ed97878

Kolyunya:
I don't want to be annoying or anything but the signature of the existing method changed (without breaking the BC) and a new type of parameter is now accepted. It seems to me that the original changelog line is quite precise. Never mind anyway.

rob006:
Tests are broken on PHP 5 - `Error` was introduced in PHP 7: https://travis-ci.org/yiisoft/yii2/jobs/214033905#L524

Kolyunya:
Thanks! I made a patch https://github.com/yiisoft/yii2/pull/13834.

Issue #13804, opened 2017-03-20
Allow FindOne to accept complex where conditions


enigmatix:
The FindOne method checks to see if an array is associative, and if so, prepares a where clause based on the associative array.  Otherwise, it will try to fetch a record based on the integer value of the variable.

I would expect that the findOne method instead checks to see if the variable is an array of any type, rather than just associative.  This would allow for all types where conditions to be supported.

### What steps will reproduce the problem?

Have a table with 2 user records, user id 1 and user id 2 and:

`$user = User::findOne(['!=','id',1]);
`
### What is the expected result?

`echo $user->id; // prints 2;
`
### What do you get instead?

`echo $user->id; // prints 1;
`
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11
| PHP version      | 5.6
| Operating system | Ubuntu

samdark:
The current behavior is not correct indeed. It should either support complex conditions or throw an exception.

Kolyunya:
All "*non-associative*" arrays are currently treated as primary key. I can't see a straightforward fix for that...

```php
    protected static function findByCondition($condition)
    {
        $query = static::find();

        if (!ArrayHelper::isAssociative($condition)) {
            // query by primary key
            $primaryKey = static::primaryKey();
            if (isset($primaryKey[0])) {
                $pk = $primaryKey[0];
                if (!empty($query->join) || !empty($query->joinWith)) {
                    $pk = static::tableName() . '.' . $pk;
                }
                $condition = [$pk => $condition];
            } else {
                throw new InvalidConfigException('"' . get_called_class() . '" must have a primary key.');
            }
        }

        return $query->andWhere($condition);
    }
```

cebe:
I do not think there is a reliable way to distinguish between a string primary key value or a string that could have been meant to be a condition. The docs are very clear about what the parameters expect imo.

enigmatix:
I should point out that my request was to support complex 'where' arrays, not a complex condition as a string.  I agree that it should not support complex string conditions.  Couldn't this be achieved simply changing 

`ArrayHelper::isAssociative($condition)` to `is_array($condition)` @ line 174?

As then, the array will be automatically handled at `return $query->andWhere($condition);` whether there are complex conditions or not.

cebe:
The array syntax accepted by where() is significantly different from what an array means for findOne().

enigmatix:
I take it you are referring to this from the docs:

[Docs link
[...]

PR #13816, opened 2017-03-21
Add complex query support for findOne|findByCondition

idsulik:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #13804

Add complex query support for findOne|findByCondition.
Example: `UserModel::findOne(['>=', 'age', 18]);`


samdark:
https://travis-ci.org/yiisoft/yii2/jobs/213490036

Kolyunya:
This doesn't seem correct to me. A primary key may be an array of three elements.

framework/db/ActiveRecord.php
```diff
@@ -171,18 +171,16 @@ protected static function findByCondition($condition)
...
-        if (!ArrayHelper::isAssociative($condition)) {
+        if (!is_array($condition)) {
             // query by primary key
             $primaryKey = static::primaryKey();
             if (isset($primaryKey[0])) {
-                $pk = $primaryKey[0];
-                if (!empty($query->join) || !empty($query->joinWith)) {
-                    $pk = static::tableName() . '.' . $pk;
-                }
-                $condition = [$pk => $condition];
```
MysteryDragon: Very suspicious removing useful functionality. Is it really necessary?

framework/db/ActiveRecord.php
```diff
@@ -171,18 +171,16 @@ protected static function findByCondition($condition)
...
+                $condition = [$primaryKey[0] => $condition];
             } else {
                 throw new InvalidConfigException('"' . get_called_class() . '" must have a primary key.');
             }
+        } elseif (!is_array($condition) || (!ArrayHelper::isAssociative($condition) && count($condition) != 3)) {
```
MysteryDragon: `!is_array` condition is useless, and actually we can use condition like ['and', 'statement1', 'statement2', 'statement3'], so, `count($condition) != 3` isn't correct.

Kolyunya: Also isn't `findOne()` supposed to find by a composite (e.g. three components) primary key?

samdark: Right. I think that's the reason it was not implemented with hash conditions...

framework/db/ActiveRecord.php
```diff
@@ -171,20 +171,16 @@ protected static function findByCondition($condition)
     {
         $query = static::find();
 
-        if (!ArrayHelper::isAssociative($condition)) {
+        $isConditionArray = is_array($condition);
+        if (!$isConditionArray || ($isConditionArray && !ArrayHelper::isAssociative($condition) && count($condition) < 3)) {
```
samdark: First of all, it could be simplified to:

```php
!$is_array($condition) || (!ArrayHelper::isAssociative($condition) && count($condition) < 3)
```

Then `< 3` looks very suspicious. There could be a composite primary key which consists of any number of columns, not necessary two. So the condition is wrong.

samdark (changes requested):

cebe:
I do not think these methods should support more cases than they do now. they are shortcuts for getting a single record quickly. As soon as there is the need for more complex usage, you should use `find()` instead.

samdark:
Agree. There's a clear syntax conflict so even technically it can't be implemented.

IceJOKER:
Thanks for feedback! I got it

Issue #13689, opened 2017-03-02
Broken Error trace due to regression bug


dynasource:
Just traced a regression bug caused by commit https://github.com/yiisoft/yii2/blob/master/framework/web/ErrorHandler.php#L199

It happens when a Closure method is injected into ReflectionMethod causing:
"An Error occurred while handling another error" and breaking the Error page.

The solution is an extra check on $method for type checking on a Closure. Will create a PR for it.

samdark:
How to reproduce it?

dynasource:
this will be part of the unit test

PR #13817, opened 2017-03-21
Fix error during handling of closure errors

mikehaertl:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | https://github.com/yiisoft/yii2/issues/13689

If an error occurs inside a closure method, the error handler will fail. For example:

```php
        $x = array_map(function ($r) {
            return $r->doesnotexist;
        }, [1, 2, 3]);
```

Will lead to:

> An Error occurred while handling another error:
ReflectionException: Method app\controllers\ClientController::app\controllers\{closure}() does not exist in /var/www/vendor/yiisoft/yii2/web/ErrorHandler.php:199

The fix is easy, but I was not sure how to properly test this.

### DIFF
--- a/framework/web/ErrorHandler.php
+++ b/framework/web/ErrorHandler.php
@@ -191,7 +191,7 @@ class ErrorHandler extends \yii\base\ErrorHandler
         $url = null;
 
         $shouldGenerateLink = true;
-        if ($method !== null) {
+        if ($method !== null  && substr_compare($method, '{closure}', -9) !== 0) {
             $reflection = new \ReflectionMethod($class, $method);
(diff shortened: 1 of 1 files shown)


dynasource:
duplicates https://github.com/yiisoft/yii2/issues/13689

samdark:
@dynasource any idea about tests?

samdark:
Fixed in https://github.com/yiisoft/yii2/commit/6792099bcc0b682cc9fcfa9e91157a29e8bfe460

PR #13823, opened 2017-03-22
Refactor migrations template

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

- [x] Add a `safeDown()` placeholder.
- [x] Make `safeUp()` consistent with `up()`.
- [x] Add a missing period in a comment.
- [x] Update tests.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -53,6 +53,7 @@ Yii Framework 2 Change Log
 - Bug #13738: Fixed `getQueryParams()` method in `yii.js` to correctly parse URL with question mark and no query parameters (vladdnepr)
 - Bug #13776: Fixed setting precision and scale for decimal columns in MSSQL (arturf)
 - Bug #13704: Fixed `yii\validators\UniqueValidator` to prefix attribute name with model's database table name (vladis84)
+- Enh #13823: Refactored migrations template (Kolyunya)
 
 2.0.11.2 February 08, 2017
 --------------------------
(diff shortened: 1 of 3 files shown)


Kolyunya:
Ooops missed the tests.

samdark (approved):

samdark:
Please add a line for changelog and I'll merge it.

Kolyunya:
Would
> Enh: Refactored migrations template
be sufficient?

samdark:
If you'll add `#13823` and your username — yes.

Kolyunya:
I used to think that the `#xxxx` is the issue number, isn't it?

samdark:
It is. In this case it's pull request number because there's no issue.

samdark:
Merged. Thanks!

Kolyunya:
Glad to help!

B.t.w. is unsafe template generated by default purposefully? I know that there are DBMS which don't support transactions but it seems like this adds friction for the majority of the developers. I personally use safe migrations only.

samdark:
@Kolyunya that's a good question.

samdark:
Deserves separate issue. I don't remember why unsafe ones are used and it doesn't look good.

Kolyunya:
Shall I open an issue for that?

Kolyunya:
I think the rationale behind that is that *some* DBMSs don't support transactions and safe migrations may cause errors.

Issue #12588, opened 2016-09-23
Should remove the JqueryAsset registration from [yii\web\View::registerJs()]  ?


Maslow:
### What steps will reproduce the problem?

If I have to use JqueryAsset implemented by myself for a specified version , the yii\web\View::registerJs() method will register the built-in yii\web\JqueryAsset automatically !  Then I have two jquery.js in one Page ~~~~

ps: 
1.  I have to use Jquery for a specified version , because  some jquery plugins required.
2.  I always want to use yii\web\View::registerJs() as well.
3.  Now, I HAVE TO configure the assetsManager to solve my problem , it doesn't look smart.
### What is the expected result?

Just remove it ! 
Do not use JqueryAsset in yii\web\View::registerJs() to deal with the yii\web\View::POS_READY things.
Alternatively , just use NATIVE javascript instead. 
### What do you get instead?

 NATIVE javascript
### Additional info

Also consider to remove  [yii\web\Jquery] from yii2 ? 

| Q | A |
| --- | --- |
| Yii version | 2.0.9 |
| PHP version | PHP 7.0 |
| Operating system | OS X |

SilverFire:
Use Dependecy Injection Container and override `yii\web\JqueryAsset` with your own.

`Yii::$container->set('yii\web\JqueryAsset', 'you\own\JqueryAsset`);`

See [Dependeny injection container](http://www.yiiframework.com/doc-2.0/guide-concept-di-container.html#practical-usage) guide page to know more.

SilverFire:
Or even easier - override `bundles` array

http://www.yiiframework.com/doc-2.0/yii-web-assetmanager.html

Maslow:
I know the two ways you said above,  but what I want to say is not that. 
For example , If I use View::registerJs() , but I don't want to use Jquery in my page.  Maybe I can't use it because of my BOSS ,  my Leader etc.  
This is what I want to express. 
Why hard-coding the Jquery in the View class?

Maslow:
@SilverFire

klimov-paul:
`JqueryAsset` is loaded **only** in case you are using `View::POS_READY` or `View::POS_LOAD` position for your JavaScript files. It is loaded because these positions require `JQuery` functions to be actually applied.

If you do not want JQuery o be involved in the process you should use other script positions.

Maslow:
Thank you

klimov-paul:
[...]

Issue #8452, opened 2015-05-17
Requirements free version of Yii 2


hiqsol:
I want to create console only yii2 based application.
And I don't want all those bower assets and other dependencies to be installed.

Don't you mind if I create (and maintain on Packagist) requirements free version of yii2 core code?
It's going to be exact copy of yiisoft/yii2-framework with only composer.json changed.
I've already tried: https://github.com/sol-hiqdev/bare-yii2
It works fine for me and could be useful for others. 

Do you plan to split your framework to smaller parts for similar purposes?

cebe:
we have a similar discussion already, could not find the issue so I will keep this open.
we will consider it for version 2.1.

sol-hiqdev:
Nice to hear it. 
Hope you are about to cut core part out of the whole framework. And it seems to be a very good quality for the core to have no dependencies.

klimov-paul:
See https://github.com/yiisoft/yii2/issues/10557#issuecomment-178571517

I suppose this issue does not actually require stripping `yii\web\*` classes. Moving bower/asset depent widgets would be just enough here.

klimov-paul:
Current `yii2` repo should be split into several ones.
Note, that we should keep a balance regardless future amount of repositories.
I suppose only the code, which has external dependency like `jquery` should be put into a separated repo.

While using small codebase and many packages may look flexibile it has a drawbacks:

1) It is hard to navigate the code, which is split into several packages

2) Yii architecture is relatively simple and thus sometimes has cross references. For example `\yii\db\Connection::$schemaCache` existance makes a dependency to `yii\caching\*` classes, while there is a `yii\caching\DbCache`, which depends on `\yii\db\Connection`.

3) Fetching many packages from `composer` may produce installation performance degradation: the most expensive part of composer instalation is a resolving of the dependency tree, comparing requirements from different packages. The more packages are involved - the more time will be needed.

4) Composer generates a autoload by PSR for each package, remembering its root path. The more packages are installed - the more paths to remember and process. This may slow down the class autoloading.

Of course there should be some balance to maintained here.
I can see no reason for exracting such packages like `web`, `console` or `caching`.
Except those classes, which have external dependencies, I suppose only `yii\rest\*` should be separated.

klimov-paul:
I propose introduction of following new repositories:

1) `yii2-jquery`
See #9525 .
[...]

Issue #11792, opened 2016-06-21
2.1.x and 2.2.x Milestone - 2 cents +1


WadeShuler:
**PHP Version:** I am for PHP 5.6 for 2.1.x milestone, and PHP 7.x (stable) for 2.2 milestone. I feel a large jump could cause unexpected issues. Since others comment that Yii can currently run fine on 7, we may see most of the issues being with 3rd party extensions (who make up the most important part of Yii). Give them time to adapt and catch up.

**init:** - I actually like the `init` option. It saves from ugly "IF CHECKS" to determine if I am on x server or y, and to use separate configs and settings. Maybe separate it as an "add-on"?

**Assets:** - I am fine with doing away with assets - It should be an add-on and not part of the main framework.

I think it should be more straight-forward for using the jQuery and Bootstrap CDN's. Possibly using them out of the box 👍 I would like to see FontAwesome (ex:nice icons in the text inputs on login page) and using the CDN. Of course, a way to override the CDN and use your local files.

I do like AssetBundles though. I like to make a bundle for say, "Revolution Slider", where it loads the appropriate css and js files. So when I want to use it on a page, I load the bundle. **Life Saver!** Though they could be simpler and more dynamic. If I load 5 js files, I should be able to add a weight to them (1-99) so I can control the order.. or each js file can have an option to "load after x". Each js file should be able to be in the header or footer. 1 asset bundle, calling 5 js files, with control over header/footer and which loads before the other. - I think currently, the only way it to control which loads before the other, is to create 2 AssetBundles, splitting them up and depending on the 1st.

**Pull request to be merged:** Fix the hidden input labels to be off by default!!

**Logic Reflow:** Completely separate `frontend` and `backend`. Every site where I have users create accounts, I have admins to manage them! Am I the only one who thinks like this? Frontend would be your users area, and backend would be your admin area... admins should not be stored in the users table... Default migration should create `user` table and `admin` table. Frontend and Backend should be separated from each other (app id, session, cookie) by default. --- I would like to be able to have `Yii::$app->admin` as well as `Yii::$app->user`. It's confusing when looking through lines of code.

Maybe instead of `backend` it should be called `admin`? Maybe `frontend` should be called `user`. Maybe adding a 3rd for the homepage. You could have site.com, site.com/admin, site.com/user.

Pretty URLs by default.

### New Tricks

SEO friendly out of the box

[DataTables](https://datatables.net) out of the box

A Yii2 logger and Chrome extension for debug

[...]

Issue #10557, opened 2016-01-12
Separate widgets from core framework


leandrogehlen:
I think necessary to separate, `widgets`, `assets` and `grid` to an extension.
the name can be yii2-widgtes.

tom--:
So ActiveForm would not be part of the framework? It would be in an ext?

SilverFire:
The rules of separating are not obvious because of coupling.

klimov-paul:
Relates #8452

dynasource:
it depends on how you look at widgets. Recently, I am starting to see Yii's widgets as something to extend more and more often. From this perspective, I am seeing it as a core functionality. When these Widgets keep on behaving as abstracts, it could still be logical to keep it in core.

With respect to grid, as it is a Widget, I would expect his under widgets. The current location in /root/grid is not intuitive.

leandrogehlen:
@dynasource, there are projects that, don't use view layer provides by yii. (like AngularJS).
Because of this is better separate this part of framewor into another extension

dynasource:
well, if thats the criterium, than IMHO all the folders in the Yii root can be put in separate extensions.

But, this would be logical from a fundamental point of view, put not from a optimal point of view. This would mean I have to fill my composer.json with more and more basic Yii functionalities, making my deploys slower and slower.

I definately prefer speed above core-size.

With respect to this, I would redirect this discussion towards the right criterium. What is most important in this matter. I would say the 'mass'. For instance, by using a 80-20 rule. What is used by 80% of all the users, must be put in core. The rest in seperate extensions.

samdark:
Deploys won't be slower. If the version is fixed, packages will be loaded from cache.

leandrogehlen:
@samdark :+1:

dynasource:
if thats the case, I think @leandrogehlen suggestion is practical and detangle wherever possible

tom--:
There are a lot of Yii class files my project doesn't need. But I'm not trying to save disk space on my servers by proposing that Yii remove them into separate repos. If I needed to optimize disk space use for some reason, I'd be looking at **big** things to get rid of or shorten first (Pareto analysis) and not at PHP source code files. And if the time it takes to copy these files to the server is a serious problem for someone, I'd be surprised.

So I don't think Yii-core's overall byte count should be the reason to remove certain files to separate repos.

[...]

PR #13839, opened 2017-03-23
Preparation for #8452: 'jquery' package extraction

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #12588, preparing for #8452

 'jquery' package has been extracted as preparation for `yii2-jquery` repo extraction, which will allow to create requirements free version of Yii2 and make composer-asset-plugin installation to be optional.


rob006:
Can't we just use jQuery from CDN by default?

klimov-paul:
> Can't we just use jQuery from CDN by default?

It makes no difference. PHP framework should not be bound to any JavaScript framework.
It should be developer choice if he wishes to use 'JQuery', 'ReactJS', 'Angular' or anything else.

rob006:
> It makes no difference.

It fixes main issue - you don't need asset plugin anymore and you can install Yii without any frontend dependencies. 

I agree that framework-agnostic approach is generally good thing, but it looks like overkill and unnecessary complication. If I want write extension which extends `GridView`, which class I should use as a base class?

klimov-paul:
> I agree that framework-agnostic approach is generally good thing, but it looks like overkill and unnecessary complication.

Following your arguments, should not we bind Yii2 to Twitter Bootstrap as well? Why keep `yii2-bootstrap` as a separated repo? Does not it produce an 'overkill and unnecessary complication'?

> If I want write extension which extends GridView, which class I should use as a base class?

It depends on what you want to do: you may use `yii\grid\GridView` and provide your own JS layer for its filter - or do not use any JS layer, but a regular HTML form submit. Or you can extend `yii\jquery\GridView` and inherit related JQuery code with it.

I am fixing the issues, which are opened and which coew team has agreed to be fixed. If you wish to debate on necessity of such changes use the related issues for that: #8452, #11792, #10557, #8709, #9525.
When you are able to convince @WadeShuler, @leandrogehlen, @graphicmist, @fabiocarneiro and others that thier arguments are unfounded I will discard this PR.

klimov-paul:
This PR fixes #12588

rob006:
> Following your arguments [...]

Please, be constructive...

> It depends on what you want to do [...]

My extension is not related to JS at all so I may use any of this class. But if I use `yii\grid\GridView` as a base class people who want use jQuery stuff in `GridView` will not be able to use my widget. If I use `yii\jquery\GridView`, I will force users to download all this frontend dependencies. So basically I need create and maintain 2 extensions instead of 1.

> I am fixing the issues, which are opened and which core team has agreed to be fixed. If you wish to debate on necessity of such changes use the related issues for that: #8452, #11792, #10557, #8709, #9525.

Did you at least read these issues before linking? They are mostly about installing frontend dependencies, which may be solved by using CDN fallback. Some of them even proposing using CDN for jQuery.

samdark:
Hmm... @rob006 has a point about having two separate GridViews...

framework/UPGRADE.md
```diff
@@ -57,6 +57,14 @@ Upgrade from Yii 2.0.x
   providing ability to setup custom mail headers. Make sure your provide implementation for those methods, while
   creating your own mailer solution.
 
+* Classes `yii\widgets\ActiveForm`, `yii\widgets\ActiveField`, `yii\grid\GridView`, `yii\captcha\Captcha`, `yii\web\View`
+  have been refactored to be more generic without including any 'JQuery' support and client-side processing (validation, automatic submit etc.).
+  You should use `yii\jquery\*` package classes instead, to make old code function as before. E.g. use `yii\jquery\ActiveForm` instead
+  of `yii\widgets\ActiveForm`, `yii\jquery\GridView` instead of `yii\grid\GridView` and so on.
+
+* Assets `yii\web\JqueryAsset`, `yii\web\YiiAsset`, `yii\validators\ValidationAsset`, `yii\captcha\CaptchaAsset` have been
```
samdark: I was planning extracting Captcha in its own package.

dynasource:
I agree with @rob006 , I'm positive towards disentanglement of parts, but I dont like the overuse of inheritance in the framework. This is not exactly favoring 'composition' over inheritance and limits flexibility.

samdark:
@dynasource any ideas on how to do it better?

dynasource:
like I said, encapsulate the logic in dependencies instead of inheritance. All these JS parts deserve their own classes and should not pollute the inheritance tree.

dynasource:
this is also an example of GridView `has` jQuery instead of GridView `is` jQuery
https://en.wikipedia.org/wiki/Composition_over_inheritance#Benefits

samdark:
I understand the theory. I was asking about its practical application for this particular case.

klimov-paul:
> My extension is not related to JS at all so I may use any of this class. But if I use yii\grid\GridView as a base class people who want use jQuery stuff in GridView will not be able to use my widget.

Same happens for `yii2-bootstrap`: if you extend `yii\widgets\ActiveForm` your extension can not be used for `yii\bootstrap\ActiveForm`. 
If you extend `yii\base\View` it will not be able to substitute `yii\web\View` and so on.
It is rather funny to hear so much concerns about 'polluting the inheritance tree' here.

> Did you at least read these issues before linking? They are mostly about installing frontend dependencies, which may be solved by using CDN fallback. Some of them even proposing using CDN for jQuery.

I can ask you the same question. So:

1) #8452 (Requirements free version of Yii 2) requires `yii2-framework` package be independent of any external library, allowing easy composer installation.

2) #11792 suggets splitting main repo into several ones, proposing packages `Recaptcha`, `Migrations`, `RBAC`, which I consider unfounded. But just as you he likes CDN a lot, ignoring the fact someone may use AnguarJS instead of JQuery.

3) #10557 (Separate widgets from core framework) demands all widgets should be placed under separated repo because of assets usage. See my comment here https://github.com/yiisoft/yii2/issues/10557#issuecomment-178571517

4) #8709 (Yii2 for REST applications) demands exluding 'bower packages' from installation as a redundant.

5) #9525 (Jquery dependency) demands exactly what I say: 

> Do you really think is a good idea to have a asset as dependency of a PHP FRAMEWORK?

All those issues have been merged into single one #8452.

klimov-paul:
> All these JS parts deserve their own classes and should not pollute the inheritance tree.

This PR has a lot of room for improvement.
For example: at the present state there is `ClientValidatorBuilder`, which is a slave object of the `yii\jquery\ActiveFrom`. In theory it can be evolved in something, which will consume all client-side code composition - but it is hard to say. There are many design flaws around form client-side composition - I can not deal with all them on my own.

> GridView has jQuery instead of GridView is jQuery

Client script for `GridView` can be handled using a behavior, which usage for widgets has been introduced at #7435. For the `ActiveForm` it is hard to say.

rob006:
> Same happens for yii2-bootstrap [...]

And this *is* a problem. Having one solution like that does not justify blind follow in this direction.

> I can ask you the same question [...]

So, if all these issues are about **installing** assets, why do you think that using CDN will not solve them?

Note that using jQuery as default  frontend framework does not force you to use jQuery. If we extract registering assets  and JS to separate protected method, changing fronted framework will be pretty easy - you need to override only one method (could be done by trait or behavior). We can even make it configurable. And this will practically solve this issue without create extra non-jQuery classes - you don't need install frontend dependencies, and you don't need use jQuery at all.

klimov-paul:
> Note that using jQuery as default frontend framework does not force you to use jQuery. If we extract registering assets and JS to separate protected method, changing fronted framework will be pretty easy 

What about issues like #12588 then?

> I understand the theory. I was asking about its practical application for this particular case.

The original issue has 2 main tasks:

1) Exclude JQuery from the core classes.
2) Compose standalone JQuery extension, which usage will produce compatibility with the old Yii2-based code.

I do not consider this solution to be a final - it is the obvious way of solving task and should be improved.
But I need constractive ideas for it instead of arguing about the sense of the original issue.

Here I am trying to put things right, while someone may prefer to ignore the original design flaws and hide behind hacking solutions, like using CDN (proposed by @rob006) or put `provide` into `composer.jaon` (proposed by @cebe).

During the internal disscussions I have promised the core team, that I will show them the direction for splitting `yii2` package removing asset dependencies - so I did. But I can not enforce any solution on this matter.

rob006:
> What about issues like #12588 then?

I don't get your question. Where is the problem? You have already solved this: https://github.com/klimov-paul/yii2/blob/ddcef80c72ae0babbf27d221d2b6a9ab65c1baaf/framework/web/View.php#L581-L606

samdark:
This pull request touches many aspects of the framework at once. Let's address these one by one.

# yii\web\View and yii\jquery\View

| Q                      | A
| -------------------| -------------
| Is web only?     | Yes
| Inheritance issue? | Yes

I don't think `yii\jquery\View` is needed. Having  `document.addEventListener()` and  `window.addEventListener()` ways of doing things are enough.

## GridView

| Q                      | A
| -------------------| -------------
| Is web only?     | Yes
| Inheritance issue? | Yes

Looking at the code I see that most of it if not everything could be moved to a class which doesn't inherit from gridview itself:

```php
GridView::widget([]);
(new GridViewJavaScript($options))->register($this);

class GridViewJavaScript
{
    private $_options;

    public function __construct($options = [])
    {
        $this->_options = $options;
    }

    public function register($view)
    {
        GridViewAsset::register($view);
        // do something with options as it's currenly done in jquery version of gridview
    }
}
```

## Captcha

| Q                      | A
| -------------------| -------------
| Is web only?     | Yes

Could be moved into separate package as a whole and I plan doing that after we agree on how to deal with validators overall.
Widget part could be done similar to GridView.


## Validation

| Q                      | A
| -------------------| -------------
| Is web only?     | No (REST, CLI, code)
| Inheritance issue? | No

The builder solution applies to core validators only leaving extension validators as is i.e. containing both a method to validate value with PHP and a method to generate necessary validation JavaScript code.
It allows moving clientside part into jQuery package as planned but there are issues:

- Inconsistency between core and extensions is introduced.
- It's not clear what to do with extensions which are supposed to work both with and without jQuery.
- Validator while not implementing  clientValidateAttribute(), still has the methods defined. In the context of non-web usage it's not looking good.
- Big piece of code in a single method where each part isn't really connected with each other.

There's alternative solution @dynasource proposed in a short peer coding session today:

```php
abstract class AbstractValidator extends \yii\base\Object
{
    // validator could be modified by stacking more validators
    public $chainedValidators = [];

    /**
     * Executes the validation by running the current Validator and its chained Validators
     */
    final public function execute($value)
    {
        $this->validateValue($value);
        $this->runChainedValidators($value);       
    }

    /**
     * Validates a value by the current validator
     */
    abstract protected function validateValue($value) : bool;


    /**
     * Runs chained validators
     */
    private function runChainedValidators($value) : void
    {
        foreach($this->chainedValidators as $validator)
        {
             $validator = $this->createValidator($validator);
             $validator->execute($value);   
        }

    }

    /**
     * Creates Validator instance by array configuration
     */
    private function createValidator(array $config) : AbstractValidator
    {
         return Yii::createObject($config);   
    }
}

// base class for JavaScript part of the validation
abstract class AbstractJsValidator extends Validator
{
    protected function validateValue($value)
    {
        $js = $this->getJs();
        Yii::$app->view->registerJs($js); 
    }   

    abstract protected function getJs($value);

}

// --------------- third party extension --------------------


class CreditCardValidator extends AbstractValidator
{   
    // could be configured via DI container   
    public $chainedValidators = [
        ['class' => CreditCardJavaScriptExtraValidator::class]
    ];

    public function validate($value)
    {
        // actually validate credit card
    }
}


class CreditCardJavaScriptExtraValidator extends AbstractJsValidator
{

    protected function getJs($value)
    {
        // build JS necessary and return it           
    }   
}
```


## ActiveForm and ActiveField

| Q                      | A
| -------------------| -------------
| Is web only?     | Yes
| Inheritance issue? | Yes

Could be solved in the same manner as gridview.

klimov-paul:
> I don't think yii\jquery\View is needed. Having document.addEventListener() and window.addEventListener() ways of doing things are enough.

I am not so sure. For `jQuery(window).on('load')` jQuery expression `window.addEventListener()` seems to be absolutely valid replacement, but for `jQuery(document).ready()` it is hard to say. It could happen `jQuery.ready()` is doing something more - otherwise it should not ever exitst.

> Looking at the code I see that most of it if not everything could be moved to a class which doesn't inherit from gridview itself

As I said [above](https://github.com/yiisoft/yii2/pull/13839#issuecomment-288995538) it could be a behavior. Then the view syntax will be following:

```php
<?= GridView::widget([
    'as jquery' => \yii\jquery\GridViewFileBehaviorr::class
]); ?>
```

Using absolutely standalone class will overcomplicate view code, besides client script composition usually requires data from the widget internal fields.

> Captcha could be moved into separate package. Widget part could be done similar to GridView.

Perhaps, but I am not sure it should.
Yii internal captcha has a poor strength (see #1100). It is fine to have it in the core as quick (mostly demo) solution.
Extracting it into a separated repo will increase developer's expectations from it and will require significant improvement of the CAPTCHA generation algorithm.
But in general I am not against the idea - it is fine to do so.

Anyway both `yii2-captcha` and `yii2-jquery` should be composed inside 2.1 branch before using GIT tree subsplit.
The main question is where JQuery dependent part of `Captcha` widget should be located in the end: at `yii2-captcha` or at `yii2-jquery`?

> Inconsistency between core and extensions is introduced.

At the present state there is no inconsistency: any existing code with custom validator, which `clientValidateAttribute()` returns not empty result,will function as before.
However it is likely to be working with JQuery only.

> Validator while not implementing clientValidateAttribute(), still has the methods defined. In the context of non-web usage it's not looking good.

`Validator::clientValidateAttribute()` can not be just removed, because of `InlineValidator` existance. 
Its client-side validation is determined by `Model::rules()` configuration and it is always developer-defined:
https://github.com/yiisoft/yii2/blob/master/framework/validators/InlineValidator.php#L78-L90

> Big piece of code in a single method where each part isn't really connected with each other.

Client validators can be extracted into a separated classes mapped to the existing server-side validators.
But they will be almost empty containing a 3-4 lines of code. Still I can adjust code in this way if you like.

> ActiveForm and ActiveField Could be solved in the same manner as gridview.

I am not so sure. Form validation is composed by both `ActiveForm` and `ActiveField`, which interacts together doing so.
It is hard to say if it can be just wrapped around.

samdark:
> I am not so sure. For jQuery(window).on('load') jQuery expression window.addEventListener() seems to be absolutely valid replacement, but for jQuery(document).ready() it is hard to say. It could happen jQuery.ready() is doing something more - otherwise it should not ever exitst.

Well, yes. It's doing a bit more:

1. Providing polyfill for IE lower than 9.
2. Making sure callback will be executed even if code is executed after DOM ready.

Here's the code:

https://github.com/jquery/jquery/blob/master/src/core/ready-no-deferred.js

samdark:
But that's not that helpful...

samdark:
> As I said above it could be a behavior.

I like the idea. It could be configured via container as well...

samdark:
> Anyway both yii2-captcha and yii2-jquery should be composed inside 2.1 branch before using GIT tree subsplit.

Why? I've successfully extracted https://github.com/yiisoft/yii2-maskedinput without doing a subsplit...

samdark:
> The main question is where JQuery dependent part of Captcha widget should be located in the end: at yii2-captcha or at yii2-jquery?

At yii2-captcha but it should be a behavior or a separate class. Not a widget inherited from Captcha.

samdark:
> Client validators can be extracted into a separated classes mapped to the existing server-side validators.
But they will be almost empty containing a 3-4 lines of code. Still I can adjust code in this way if you like.

I think it would make perfect sense.

klimov-paul:
> Why? I've successfully extracted https://github.com/yiisoft/yii2-maskedinput without doing a subsplit

Subsplit is better as it will preserve the git history.

> At yii2-captcha but it should be a behavior or a separate class. Not a widget inherited from Captcha.

So it will, but my question still stands:

> where JQuery dependent part of Captcha widget should be located in the end: at `yii2-captcha` or at `yii2-jquery`?

If it will be located at `yii2-captcha` it will mean that this package will be dependent on `yii2-jquery` and can not be used without it. Thus it will be impossible to create a CAPTCHA solution without installing jQuery.
If it will be located `yii2-jquery` it will be dependent on `yii2-captcha`, thus you'll have to install captcha before using jQuery even if you do not need it.
The solid way to resolve it will be creating a 3rd package, which will depend on both `yii2-captcha` and `yii2-jquery`, but this may be an overcomplication.
Another solution can be putting JS-dependent part at one of `yii2-captcha` or `yii2-jquery` without explicit dependency ('suggets' composer section can be used for example), but this can be inconsistent.

samdark:
About jQuery-related part of CAPTCHA, I prefer to rewrite JS part of it without using jQuery. yii.captcha.js isn't too complicated to do it.

leandrogehlen:
My suggestion.

yii2-widgets (base widigets without js or css dependencies)
yii2-jquery (GridViewBehavior, and assets and capcha)

samdark:
@leandrogehlen separating widgets isn't the goal of this pull request.

leandrogehlen:
Ok, then , i think that the folder `grid` should be moved to `widgets` and removed assets dependencies from this classes, this way we can write a grid with `jQuery` and another with `Angular` , thus each extension can write a grid behavior with js library desired

klimov-paul:
Solution has been refactored:

- Client script for `ActiveForm`, `GridView` and `Captcha` extracted into a behaviors.
- Client validation composition chnaged to use class map and special `ClientValidator` class
- `captcha` related code packed under `framework/captcha`

samdark:
Looks good after quick glance on it. Need to check it in depth though.

schmunk42:
I took a look at this and actually, the PR was looking good, at least no `assets` folder anymore in the framework.

I tried to merge against 2.1, but it looked to me like there were a lot of clientXyz functions added again?!

I think also that all the remaining asset-bundles should be removed from the core, but one step after the other...
bildschirmfoto 2017-09-05 um 16 43 19

Can this be made *mergeable* again?

klimov-paul:
> Can this be made mergeable again?

At the present state I do not have time for this. I need to finish #14701 and underlying tasks, which depends on it. Otherwise, it will likely end in conflicts anyway.

Most likely, I will re-create this PR from scratch afterwards.

Since, at the present state, core team has no interest on reviewing this or attend to the underlying issue, it is not in hurry. Even if I *merge* this changeset, it will solves nothing until the `jquery` and `captcha` reositories will be split. While I do not have enough permissions to create a repositories under `yiisoft`, I am unable to to do it myself.
Without major core team suport this PR moves nowhere anyway.

schmunk42:
> it will solves nothing until the jquery and captcha reositories will be split. While I do not have enough permissions to create a repositories under yiisoft, I am unable to to do it myself.

100% agreed! @samdark Could you create those repos for 2.1 (`yii2-jquery` and `yii2-captcha`; also `yii2-pjax`)? Those need to be moved out.

samdark:
- https://github.com/yiisoft/yii2-jquery
- https://github.com/yiisoft/yii2-captcha

PJAX is to be removed so I haven't created a repo for it.

klimov-paul:
Migrated to #14865

PR #13845, opened 2017-03-23
Php Inspections (EA Extended): minor tweaks

kalessil:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | n/a

Substr calls check adjustments: https://3v4l.org/FC1qY
`$_cookies` re-written in the class constructor, so implicit default value is redundant

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1221,7 +1221,7 @@ class BaseHtml
             foreach ($model->getErrors() as $errors) {
                 foreach ($errors as $error) {
                     $line = $encode ? Html::encode($error) : $error;
-                    if (!in_array($line, $lines)) {
+                    if (!in_array($line, $lines, true)) {
                         $lines[] = $line;
                     }
                     if (!$showAllErrors) {
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1221,7 +1221,7 @@ class BaseHtml
             foreach ($model->getErrors() as $errors) {
                 foreach ($errors as $error) {
                     $line = $encode ? Html::encode($error) : $error;
-                    if (array_search($line, $lines) === false) {
+                    if (!in_array($line, $lines)) {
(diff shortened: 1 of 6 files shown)


framework/helpers/BaseHtml.php
```diff
@@ -1221,7 +1221,7 @@ public static function errorSummary($models, $options = [])
             foreach ($model->getErrors() as $errors) {
                 foreach ($errors as $error) {
                     $line = $encode ? Html::encode($error) : $error;
-                    if (array_search($line, $lines) === false) {
+                    if (!in_array($line, $lines)) {
```
samdark: Wouldn't plugin require strict parameter to be set?

framework/web/CookieCollection.php
```diff
@@ -34,7 +34,7 @@ class CookieCollection extends Object implements \IteratorAggregate, \ArrayAcces
     /**
      * @var Cookie[] the cookies in this collection (indexed by the cookie names)
      */
-    private $_cookies = [];
+    private $_cookies;
```
samdark: That's wrong change. See:

```php
public function getCount()
{
    return count($this->_cookies);
}
```

samdark (changes requested):
Overall good changes but there are some things to adjust.

framework/helpers/BaseHtml.php
kalessil: Yes, will add

framework/web/CookieCollection.php
kalessil: The property is immediately overridden when constructing an object:
```php
    public function __construct($cookies = [], $config = [])
    {
        $this->_cookies = $cookies;
        parent::__construct($config);
    }
```

samdark: OK.

samdark (approved):

samdark:
Merged. Thank you!

Issue #13842, opened 2017-03-23
Uses UniqueValidator and alias for table


vladis84:
### What steps will reproduce the problem?
Uses UniqueValidator and alias for table.

I think the resolution of this Issue has caused a problem with query aliases. If someone overrides the find() method of the model and uses it to add and alias to the model table, the validator will execute a query that will cause an error.

Maybe the prefix used on the validator query should be extracted from the Query object and not from the tableName() method.

### What is the expected result?
Add in column alias for table.

### What do you get instead?
 ambiguous error sql while using unique validator

### Additional info
parent issues #13704 
| Q                | A
| ---------------- | ---
| Yii version      |  2.0.13
| PHP version      | 7.0.17
| Operating system |

vladis84:
There is a method getTableNameAndAlias is **private**.
```php
yii\db\ActiveQuery::getTableNameAndAlias
```
Have any other suggestions for solving the error?

jlorente:
I think yii\db\ActiveQuery::getTableNameAndAlias() method could become a public method since it does not modify anything in the ActiveQuery object and is only used to retrieve information about the query. Then we could use the method to patch the UniqueValidator.

vladis84:
I'm not sure about that.
All public methods ActiveQuery modify the query and use fluent interface.
Consistency code bases.

vladis84:
I propose to render the method in helper.
Alexander you do not mind?
@samdark

samdark:
Which helper?

vladis84:
I will transfer this code to the helper.
https://github.com/vladis84/yii2/blob/40cf0dd7da4e23582a9a844af8a30bc43d09f8d7/framework/db/ActiveQuery.php#L559

I am make pull request.

PR #13799, opened 2017-03-18
Refactor Captcha widget and write tests.

vladis84:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | yes/no
| Tests pass?   | yes
| Fixed issues  |

PR #13857, opened 2017-03-25
 Fixed ambiguous error sql while using unique validator

vladis84:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | yes/no
| Tests pass?   | yes
| Fixed issues  | #13842


samdark:
Helper isn't really good idea and I wanted to discuss it first. You're too fast doing a pull request :) Helpers in general contain methods which could be reused by end users. Active Query helper is rather internal thing and it doesn't really make sense to expose it.

vladis84:
I am think this is an ActiveQuery duty.
Don't know how to do it.

samdark:
Then maybe add it to active query itself?

samdark:
There are `getTableNameAndAlias()` and `alias()` methods which could be relevant.

vladis84:
I will propose options before the end of the week.
I am  want #13799 finish.

vladis84:
I have no other ideas.
Sorry.

samdark:
OK. I'll take care about it. Thank you for contributing.

Renkas:
Any update on this getting merged?

samdark:
@Renkas not yet. Too many other things are in the works. Methods should be moved to ActiveQuery. If you have extra time to do it, please send a pull request.

Renkas:
Unfortunately it's not something I feel comfortable doing. Any chance @vladis84 can finish this up?

Other possibility would be to add a flag to UniqueValidator that would switch on/off prefixing column names with table name. As this functionality was added to cover an edge case - but from my point of view always prefixing with table names affects more developers than not prefixing it.

samdark:
Moved to https://github.com/yiisoft/yii2/pull/14049

PR #13862, opened 2017-03-27
Update phpdoc, added @property-read for read-only property and @property-write for write-only property

lav45:
The @property-read tag allows a class to know which ‘magic’ properties are present that are read-only.
https://phpdoc.org/docs/latest/references/phpdoc/tags/property-read.html

The @property-write tag allows a class to know which ‘magic’ properties are present that are write-only.
https://phpdoc.org/docs/latest/references/phpdoc/tags/property-write.html


rob006:
Duplicates #10989

PR #13865, opened 2017-03-28
Fix input ID uniqueness

uaoleg:
In case when on one page there are several forms for the same `ActiveRecord` model, than auto-generated by `ActiveForm` inputs have identical HTML attribute `id`, that obviously causes lots of problems.

I propose:
1. Add simple counter to guarantee uniqueness even for inputs of same attribute of same model
2. Add "salt" for a case of ajax requests, when counter will turn back to "0"

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | possibly, if developers used auto generated IDs in there JS quires
| Tests pass?   | yes

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -96,6 +96,17 @@ class BaseHtml
      */
     public static $dataAttributes = ['data', 'data-ng', 'ng'];
 
+    /**
+     * Counter to make each input with guaranteed unique ID
+     * @var int
+     */
+    protected static $inputIdCounter = 0;
+
+    /**
+     * Salt to guarantee uniqueness for input IDs in cross-ajax requests
+     * @var string
+     */
+    protected static $inputIdSalt;
 
     /**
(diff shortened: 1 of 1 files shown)


MysteryDragon:
We can not relay on such unpredictable id.

Renkas:
I dont get why is the random and salting needed? All you would need is an option to set prefix as far as I understand the underlying problem.

Why is the input ID prefixed: `jsinput-`? How is JS in any way related to input field? It seems like your specific usecase not a framework level solution.

@MysteryDragon 
> We can not relay on such unpredictable id.

Yeah - ID output needs to be same if you call the method with identical parameters.
Usecase: I output input field and label separately - I would need to get input fields ID reliably from `getInputId`

samdark:
Won't be merged. @uaoleg if there's a problem this pull request solves please describe it in a separate issue. Thanks.

uaoleg:
Agree, it's too specific for my application, sorry for disturbing

PR #13877, opened 2017-03-29
Enhancement #203 - using pattern [schemaname].[tablename] when using PostgreSQL

tontm1981:
Not a new feature, but an improvement for an existing feature.
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  |no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #203


samdark:
Seems it's something very wrong with this pull request...

PR #13898, opened 2017-03-31
Enhancement: add $displayList option to BaseConsole::select(...)

ghost:
This adds an extra parameter to BaseConsole::select() and console\Controller::select() to initially always display the available options. This also alters the prompt in that case so as not to display the keys again. The default behaviour has remained unchanged.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no (but none related to this)
| Fixed issues  | -

### DIFF
--- a/framework/console/Controller.php
+++ b/framework/console/Controller.php
@@ -312,12 +312,13 @@ class Controller extends \yii\base\Controller
      *
      * @param string $prompt the prompt message
      * @param array $options Key-value array of options to choose from
+     * @param bool $displayList whether or not to initially display the option list
      *
      * @return string An option character the user chose
      */
-    public function select($prompt, $options = [])
+    public function select($prompt, $options = [], $displayList=false)
     {
-        return Console::select($prompt, $options);
+        return Console::select($prompt, $options, $displayList);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


framework/console/Controller.php
```diff
@@ -312,12 +312,13 @@ public function confirm($message, $default = false)
      *
      * @param string $prompt the prompt message
      * @param array $options Key-value array of options to choose from
+     * @param bool $displayList whether or not to initially display the option list
      *
      * @return string An option character the user chose
      */
-    public function select($prompt, $options = [])
+    public function select($prompt, $options = [], $displayList=false)
```
samdark: It is a backwards compatibility break so can't be merged into 2.0.x like that.

samdark (changes requested):

framework/helpers/BaseConsole.php
```diff
@@ -855,21 +855,39 @@ public static function confirm($message, $default = false)
...
+     * @param bool $displayList whether or not to initially display the option list
      *
      * @return string An option character the user chose
      */
-    public static function select($prompt, $options = [])
+    public static function select($prompt, $options = [], $displayList=false)
     {
+        $errorCount = 0;
+        if ($displayList) {
+            goto displayList;
```
cebe: please avoid the use of goto here and create a new method for displaying the options.

cebe (changes requested):

PR #13900, opened 2017-04-01
Changed description of RBAC to recommend migrations instead of console commands for managing hierarchy

samdark:
Currently have an example with a console command. It's OK for demo purpose but is not very practical. Migration fits both purposes.

### DIFF
--- a/docs/guide/security-authorization.md
+++ b/docs/guide/security-authorization.md
@@ -259,9 +259,22 @@ Building authorization data is all about the following tasks:
 - assigning roles to users.
 
 Depending on authorization flexibility requirements the tasks above could be done in different ways.
+If your permissions hierarchy is meant to be changed by developers only, you can use either migrations
+or a console command. Migration pro is that it could be executed along with other migrations. Console
+command pro is that you have a good overview of the hierarchy in the code rathe than it being scattered
+among multiple migrations.
 
-If your permissions hierarchy is meant to be changed by developers only, you can use [migrations](db-migrations.md)
-to initialize and modify it via APIs offered by `authManager`.
+Either way in the end you'll get the following RBAC hierarchy:
+
+![Simple RBAC hierarchy](images/rbac-hierarchy-1.png "Simple RBAC hierarchy")
+
+In case you need permissions hierarchy to be formed dynamically you need a UI or a console command. API used to
+build the hierarchy itself won't be different.
+
+#### Using migrations
+
+You can use [migrations](db-migrations.md)
+to initialize and modify hierarchy via APIs offered by `authManager`.
 
 Create new migration using `./yii migrate/create init_rbac` then impement creating a hierarchy:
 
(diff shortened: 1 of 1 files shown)


bizley:
I'm not sure how I feel with this usage of migration... There is nothing wrong with it per se but it encourages developer to put business logic there which I think should be done somewhere else in designated place like initialisation scripts.

samdark:
What if I want to adjust RBAC hierarchy a bit in a running project? How to sync these changes with my team and the server?

dynasource (approved):

bizley:
Well, it's a valid concern. I would still use separate script or maybe an old school controller for this.

Anyway - the `down` step of this "migration" seems to be a bit too much. It's like adding a table with one row of data inside `up` and dropping whole database in `down`...

samdark:
@bizley it is initial migration so it's acceptable.

rob006:
From my experience using migrations for RBAC setup and modifications is straight way to nightmare. After some time you will end up with tens of rbac-migrations - reading current permissions hierarchy from this in nearly impossible. Right now I'm using single command to update and cleanup permissions hierarchy, configuration of permissions is stored in modules. This is much cleaner and easier to maintain than migrations.

samdark:
@rob006 how do you deal with roles and assignments when updating and cleaning up permissions hierarchy?

rob006:
I'm not removing user-role assignments (I'm not using Yii 2 RBAC and I don't have FK on this junction table) only roles and relations between them. So if I don't change roles names I can safely remove and recreate roles hierarchy. I have also command for finding and removing orphaned assignments.

samdark:
Makes sense...

etki:
Migrations are not intended to deal with data, just the database structure. Common pattern for application bootstrapping is using seeds that may be not-so-idempotent and not-so-deterministic if app requires that.

bethrezen (approved):

cebe:
How about showing both approaches in the docs and discuss pros and cons?

SilverFire (approved):

MysteryDragon:
Storing RBAC hierarchy like a migrations can be really convenient.
Actually, when we work with not so predictable project, there are quite a lot changes in database scheme and in RBAC hierarchy too. So, it's hard to see current RBAC hierarchy just exactly as database scheme, and it's not so big deal 'cause we can see it in database, for example.
And my experience shoes me, that once upon a time I started with current documentation, and in such unpredictable (and fast) projects I realized that I need some migration-like mechanism for RBAC. And finally I realized that migrations isn't so bad place for storing such changes, for fast development anyway (with godlike activerecords and other funny things :) ).
So if I read this in docs, I would save some time :)

Renkas:
Maybe document that storing RBAC in PHP format is better than in DB. It's then easily tracked in VCS. I have this setup in multiple projects and have had no problems with it.

And ofc I myself (in my usecases) can't see any real need to store RBAC hierarchy inside DB at first place. I think most developers who are storing it in DB are doing it solely because they don't know any better and think that it's the right place for it.

bizley:
If you have got hundreds of users DB is the right place to store RBAC structure and DB storage is flexible which is not the case for PHP files.

I've thought about it and I cannot really give one single case scenario where having RBAC in migrations could fail, only thing keeping me away from it is my guts and this principle I've mentioned in my first comment. Maybe it's like with this old dog and new tricks.

Could someone try to give one solid reason against RBAC in migrations?

Renkas:
@bizley this is true if you give permissions directly to users. I don't think that is a good practice. I like giving user's a role and they get permissions through RBAC heirarchy.

I have one sai with 68k users and giving each of them permissions separately would be extremely stupid implemetation. I only need 7 roles to cover all the usecases on that particular site.

Storing RBAC data in DB is only good if you have a need to dynamically alter the hiearchy in realtime. But I don't have any examples where that would be actually a requirement. But I would happily read about this kind of usecases.

samdark:
> Storing RBAC data in DB is only good if you have a need to dynamically alter the hierachy in realtime. But I don't have any examples where that would be actually a requirement.

Think about flexible CMF like Drupal.

Renkas:
@samdark dont have experience with Drupal. But I think I know what you are referring to.

If you look at Yii context then I don't see that really being a main use case for a framework. I'm not saying having RBAC in DB should be removed or is not necessary at all. Just that most of the time it adds complexity that is really not needed. Solving it in the PHP would be much better and should be addressed in documentation with a note atleast. It's just my observation that I have seen from forums and other platforms that people put RBAC to DB absolutely unnecessarily.

samdark:
It is already addressed:

> Yii provides two types of authorization managers: yii\rbac\PhpManager and yii\rbac\DbManager. The former uses a PHP script file to store authorization data, while the latter stores authorization data in a database. You may consider using the former if your application does not require very dynamic role and permission management.

atrandafir:
I'd like to add a short feedback from my experience with RBAC on Yii 1.

Basically what worked and seemed very comfortable for my case:  

- Save tasks/roles definition in a array in a file such `protected/config/auth_tasks.php`
- Create a command in `protected/commands/AuthRebuildTasksCommand.php` that reads the file and regenerates the tasks/roles in the database (dropping obsolete, updating existing, creating new)
- Finally, the relations(assignmants) between the users and roles/tasks are saved in the database (as usual)

Basically all works as the RBAC original way, nothing changes, it's just that you have a very easy way to "migrate" tasks/roles definition from one database to the other.

samdark:
Updated it. Now doc has two approaches. What do you think?

PR #13903, opened 2017-04-02
Incorrect use case of Datepicker widget

w3lifer:
`dateFormat` property in the `clientOptions` always overridden by `dateFormat` property of the widget: https://github.com/yiisoft/yii2-jui/blob/master/DatePicker.php#L141


samdark:
Good catch. Thank you!

Issue #13902, opened 2017-04-02
Console app code completion doesn't find in folders


Sarke:
Create a console controller in `commands/foo/BarController.php` and with an `actionTest()` and with `namespace app\commands\foo`.

You can access it with `./yii foo/bar/test`, but it does not list in `./yii help/list` and therefore it doesn't get any shell completion.

developeruz:
It seems the problem is solved by #15249

PR #13904, opened 2017-04-02
Enh #13902: Console command completion in folders

Sarke:
| Q             | A
| ------------- | ---
| Is bugfix?    | yno
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no, but I don't think that's my fault...
| Fixed issues  | #13902


developeruz:
It seems the problem is solved by #15249

PR #13905, opened 2017-04-02
Fix phpDocumentor annotations

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Fixed some inaccurate annotations.

### DIFF
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -566,7 +566,7 @@ class Component extends Object
 
     /**
      * Returns all behaviors attached to this component.
-     * @return Behavior[] list of behaviors attached to this component
+     * @return Behavior[]|null list of behaviors attached to this component
      */
     public function getBehaviors()
     {
--- a/framework/base/Module.php
+++ b/framework/base/Module.php
@@ -305,7 +305,7 @@ class Module extends ServiceLocator
     /**
      * Returns current module version.
      * If version is not explicitly set, [[defaultVersion()]] method will be used to determine its value.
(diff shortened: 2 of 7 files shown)


samdark:
Merged. Thanks!

Kolyunya:
Glad to be of help!

framework/base/Module.php
```diff
@@ -305,7 +305,7 @@ public function setLayoutPath($path)
     /**
      * Returns current module version.
      * If version is not explicitly set, [[defaultVersion()]] method will be used to determine its value.
-     * @return string the version of this module.
+     * @return string|callable the version of this module.
```
cebe: the return value can not be callable as the callable is used to determine it.

framework/base/Component.php
```diff
@@ -566,7 +566,7 @@ public function getBehavior($name)
 
     /**
      * Returns all behaviors attached to this component.
-     * @return Behavior[] list of behaviors attached to this component
+     * @return Behavior[]|null list of behaviors attached to this component
```
cebe: this is not correct, `ensureBehaviors()` will make sure it is an array.

cebe:
fixed in 6301dae

Kolyunya:
@cebe thanks! Sorry for inconvenience.

cebe:
No problem, these were not obvious.

PR #13908, opened 2017-04-02
[docs] internals code style headers fix

bizley:
Headers fixed after latest GitHub Markdown modification (matching the one already used from number 5).


samdark:
Would you please elaborate on why this change is needed?

bizley:
Right now header mixed with list number is not displayed the same way it was before. Compare

1. Header
----------

with

## 1. Header

And on top of it header number 5 in core-code-style file has been already written like this so I suggest to change all of them to be the same.

samdark:
Merged. Thanks!

PR #13915, opened 2017-04-04
codeclimate rules update

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13913

### DIFF
--- a/tests/data/codeclimate/phpmd_ruleset.xml
+++ b/tests/data/codeclimate/phpmd_ruleset.xml
@@ -27,6 +27,7 @@
         <exclude name="StaticAccess" />
     </rule>
 
+    <rule ref="rulesets/controversial.xml/Superglobals" />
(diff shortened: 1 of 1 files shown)


cebe:
> codeclimate — 175 new issues

In some classes that are meant to wrap superglobals these are fine to use so without a list of exceptions this rule does not make much sense.

bizley:
It would be good to somehow make an exception for these specific places but to keep this rule everywhere else.

bizley:
I can not find an option in documentation to switch off this rule check for individual files so probably this has to be rejected.

bizley:
I guess there is one way https://docs.codeclimate.com/docs/marking-false-positives 
Unfortunately it requires going through every detected issue and adding their fingerprints to the configuration.

samdark:
Merged. Will update fingerprints as it goes.

PR #13932, opened 2017-04-06
Fix number attributes comparison

uaoleg:
Assume we have 2 attributes: `min` and `max`. Also we have a rule that `max` is equal or greater than `min` and they both are optional.
Than in case `min` is empty and `max` is set we got a validation error. The reason is that `parseFloat()` function converts empty string of `min` to `NaN`.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/assets/yii.validation.js
+++ b/framework/assets/yii.validation.js
@@ -277,8 +277,8 @@ yii.validation = (function ($) {
             }
 
             if (options.type === 'number') {
-                value = parseFloat(value);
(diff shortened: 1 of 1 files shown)


s1lver:
Something went wrong and a new pull request was created instead of adding to the existing one #16805.

samdark:
Thank you for the fix.

PR #13934, opened 2017-04-06
Fix tag method in BaseHtml class to include closing slash for void elememts

naemhd:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #13934

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.12 under development
 --------------------------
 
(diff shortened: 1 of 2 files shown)


Renkas:
But why?

Yii is outputting HTML not XHTML.

samdark:
As @Renkas mentioned, fix isn't needed for HTML5. Thus won't accept the pull request. Thank you for contributing though.

PR #13944, opened 2017-04-08
render is better than renderFile

nurbeknurjanov:
Because render finds the path where is that file.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no

### DIFF
--- a/framework/widgets/ContentDecorator.php
+++ b/framework/widgets/ContentDecorator.php
@@ -74,6 +74,6 @@ class ContentDecorator extends Widget
         $params = $this->params;
         $params['content'] = ob_get_clean();
         // render under the existing context
-        echo $this->view->renderFile($this->viewFile, $params);
+        echo $this->view->render($this->viewFile, $params);
     }
 }


nurbeknurjanov:
Yes. Sure.
I have noticed that everytime when I using nested layouts
I have to use a path alias (e.g. @app/views/layouts/main).
But I can not use absolute or relative pathes. because I working from different module, not inside standart backend and frontend.

It is ok when I use render method. Cause that methods analyzes where I am, and it smartly finds the way of path, and then renders.

renderFile just thinks that file is under the view of controller. But it is not.
Even if main layout file, and nested layout file in under the same directory. It doesn't find main layout file from nested layout file.

nurbeknurjanov:
Example:
I am in my own module.
/root/modules/product/ProductController
/root/modules/product/views/index.php

inside of index.php
$this->context->layout = '/container';
which successfully finds the file in /frontend/views/layouts/container.php

But inside conatiner.php I allways have to write full path from alias. I can not type here just main.php
cause container.php and main.php are in frontend/views/layouts/

```
<?php $this->beginContent("@frontend/views/layouts/main.php"); ?>

    <div class="container">
          <?=$content;?>
    </div>

<?php $this->endContent(); ?>
```
$this->beginContent("main.php"); this doesnt work, I tried many variants, main, main.php, /main.php, //main.php. it works just from full alias path

nurbeknurjanov:
Guys, do u need me any helps ?
I would like to help u on this issue.

cebe:
need to make sure that this is not breaking BC otherwise it looks fine.

cebe:
This changes breaks backwards compatiblilty. `renderFile()` handles strings starting with `/` as absolute paths while `render()` will interpret these as paths relative to the controller or application view path resulting in files not being found in existing applications. Moved to 2.1.

nurbeknurjanov:
Ok. I am writing unit test for this. But I can tell u one thing. I noticed even inside of frontend module, 
the only way to send content into nested layout is this
`<?php $this->beginContent("@frontend/views/layouts/main.php"); ?>`

```
<?php $this->beginContent("/main.php"); ?>
<?php $this->beginContent("/main"); ?>
<?php $this->beginContent("/layouts/main.php"); ?>
<?php $this->beginContent("/layouts/main"); ?>
<?php $this->beginContent("//layouts/main.php"); ?>
<?php $this->beginContent("//layouts/main"); ?>
```
non of these above doesn't work.

cebe:
yes, that is expected and changing it breaks backwards compatibility. We can make this work in version 2.1.

samdark:
Won't be merged into 2.0.

samdark:
Could be adjusted in yiisoft/view package if needed.

Issue #11018, opened 2016-03-03
Ability to override script type in registerJs()


nirvana-msu:
I have a dynamically generated `JSON-LD` markup that needs to be registered in a head section with `"application/ld+json"` script type:

``` php
<script type="application/ld+json">...</script>
```

How can I achieve this? I think `registerJs` method should provide the ability to override script type, much like you can do in `registerJsFile` by passing `options` array with `type` in it .

nirvana-msu:
I guess one workaround would be to pass the content of the script to layout using `View::params`, and then register where I need it with `Html::script`. But I feel like there should be a more straight-forward solution.

SilverFire:
It's not possible to modify the script type with `registerJs()` method, because method adds scripts to `View::js`, which is grouped by position and key. We can not change this behavior because of BC.

Actually, I can suggest you a better workaround:

Register your own `View` component, extended from `yii\web\View`, add property `jsBlocks` and `registerJsBlock()` method with the following signature:
```php
public function registerJsBlock($js, $options, $position)
{
    $this->jsBlocks[$position][] = Html::script($js, $options);
}
```
Then override `renderBodyEndHtml()` and `renderHeadHtml()` methods in order to use `$this->jsBlocks` array.

I don't think that we must include this behavior in core, so I'm closing the issue. If anybody has another ideas or arguments, why we must do it - feel free to post your opinions. Thanks

khalwat:
I would like to echo @nirvana-msu 's comments... being able to pass in an additional parameter to `View::registerJs()` would be ideal, such as: 

```
        $view->registerJs(
            $js,
            View::POS_HEAD,
            $key,
            $options
        );
```

And if `$options` is present, pass those in to `Html::script()` rather than just hardcoding `['type' => 'text/javascript']`

That could still be passed in by default if no `$options` are provided, to preserve backwards compatibility.

JSON-LD can appear anywhere in the document, so it piggybacks on the various positions offered by `View::js` and wouldn't interfere with that in any way (browser just ignore the `ld+json` scripts).

The reason I can't use your solution @SilverFire is I'm operating inside of a framework that already has extended `View`

It seems like you have a perfect solution 99% in place, there just needs to be a way to pass down `$options` to `$view->registerJs()`, and you maintain BC 100% by just having the function be:

```
    public function registerJs($js, $position = self::POS_READY, $key = null, $options=['type' => 'text/javascript'])
```

brandonkelly:
[...]

PR #13949, opened 2017-04-10
* Add registerJsonLd()

khalwat:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #11018

### DIFF
--- a/framework/web/View.php
+++ b/framework/web/View.php
@@ -129,6 +129,11 @@ class View extends \yii\base\View
      * @see registerJsFile()
      */
     public $jsFiles;
+    /**
+     * @var array the registered JsonLd blocks
+     * @see registerJsonLd()
+     */
+    public $jsonLd;
 
     private $_assetManager;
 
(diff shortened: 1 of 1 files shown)


samdark:
Thanks for a pull request.

Sorry but it looks too specific to be included into framework. The issue was reopened not because of JSON-LD specifically but because there's something which isn't possible to achieve in an easy way.

khalwat:
JSON-LD is a pretty important standard, with Google, Apple, and schema.org behind it, but oh well.

https://developers.google.com/search/docs/guides/intro-structured-data

Hopefully you'll find some way to allow for it to be doable in the Yii2 framework, perhaps generalizing the `$view->includeJs()` method.

samdark:
There are many important standards for meta-markups of different kinds: OpenGraph, microdata etc. We can't support all of them.

khalwat:
OpenGraph is already supported by Yii2, by virtue of the `meta` tag support. Microdata is just adding properties to existing HTML, and is largely deprecated in favor of JSON-LD, due to how troublesome it is to implement.

Under the umbrella of "Structured Data", the preferred encoding is JSON-LD, due to its simplicity, and that it's not intertwined with HTML tags the way microdata is (and thus relatively unmanageable).

I'll take a whack at another PR, if you're interested, which would just generalize `$view->registerJs()` to accept `$options` the way the other register calls do, it you'd like.

It'd be nice to have consistency in the APIs, as well as supporting a modern structured data standard like JSON-LD. All it'd need is to be able to accept an `$options` parameter, and then to put each script into it's own tag, rather than coalescing them together.

PR #13950, opened 2017-04-10
ucfirst with support of UTF-8

uaoleg:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -45,6 +45,23 @@ class BaseStringHelper
     {
         return mb_substr($string, $start, $length === null ? mb_strlen($string, '8bit') : $length, '8bit');
     }
+    
+    /**
+     * Returns given string with first character uppercased
+     * @param string $string
+     * @param string $encoding
+     * @return string
+     */
+    public static function ucfirst($string, $encoding = null)
+    {
+        if (!$encoding) {
+            $encoding = \mb_internal_encoding();
+        }
+        $strlen = mb_strlen($string, $encoding);
+        $firstChar = mb_substr($string, 0, 1, $encoding);
+        $then = mb_substr($string, 1, $strlen - 1, $encoding);
+        return mb_strtoupper($firstChar, $encoding) . $then;
+    }
 
     /**
      * Returns the trailing name component of a path.


arogachev:
I'm not sure if this should be added to the core or not. Guys, what are your opinions?

Also seems like there is a simpler one-liner solution - using [mb_convert_case](http://php.net/mb_convert_case):

```php
$str = "у мэри был маленький ягненок и она его очень любила";
$str = mb_convert_case($str, MB_CASE_TITLE, "UTF-8");
echo $str; // Выведет У Мэри Был Маленький Ягненок И Она Его Очень Любила
```

arogachev:
Also I highly recommend to create issue before the actual PR so we can discuss it beforehand and reject in case it won't be fixed / implemented to prevent wasting time.

framework/helpers/BaseStringHelper.php
```diff
@@ -45,6 +45,23 @@ public static function byteSubstr($string, $start, $length = null)
...
+    
+    /**
+     * Returns given string with first character uppercased
+     * @param string $string
+     * @param string $encoding
+     * @return string
+     */
+    public static function ucfirst($string, $encoding = null)
+    {
+        if (!$encoding) {
```
arogachev: You wrote "support of UTF-8" in PR's title. According to implementation, seems like it's not limited to UTF-8 only. Please be more specific.

framework/helpers/BaseStringHelper.php
```diff
@@ -45,6 +45,23 @@ public static function byteSubstr($string, $start, $length = null)
...
+            $encoding = \mb_internal_encoding();
```
arogachev: Why leading backslash is used here? It's also a function from "mb" group, not a class name.

framework/helpers/BaseStringHelper.php
```diff
@@ -45,6 +45,23 @@ public static function byteSubstr($string, $start, $length = null)
...
+        }
+        $strlen = mb_strlen($string, $encoding);
+        $firstChar = mb_substr($string, 0, 1, $encoding);
+        $then = mb_substr($string, 1, $strlen - 1, $encoding);
```
arogachev: 👎 for `$then` variable name. Be more specific. Also using noun is more preferable.

framework/helpers/BaseStringHelper.php
```diff
@@ -45,6 +45,23 @@ public static function byteSubstr($string, $start, $length = null)
...
+        return mb_strtoupper($firstChar, $encoding) . $then;
```
arogachev: There are simpler options, such as using [mb_convert_case](http://ru.stackoverflow.com/questions/105908/%D0%9F%D0%BE%D1%87%D0%B5%D0%BC%D1%83-ucfirst-%D0%BD%D0%B5-%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D0%B0%D0%B5%D1%82-%D0%B4%D0%BB%D1%8F-%D1%80%D1%83%D1%81%D1%81%D0%BA%D0%B8%D1%85-%D1%81%D0%B8%D0%BC%D0%B2%D0%BE%D0%BB%D0%BE%D0%B2). Have you considered using them?

framework/helpers/BaseStringHelper.php
```diff
@@ -45,6 +45,23 @@ public static function byteSubstr($string, $start, $length = null)
...
         return mb_substr($string, $start, $length === null ? mb_strlen($string, '8bit') : $length, '8bit');
     }
+    
+    /**
+     * Returns given string with first character uppercased
+     * @param string $string
+     * @param string $encoding
+     * @return string
+     */
+    public static function ucfirst($string, $encoding = null)
```
arogachev: Not sure if we should take the same name as in PHP. Maybe use something more clear like `capitalize` (as in Ruby for example).

cebe:
Thanks for your contribution, however if we add this, the implementation needs to be adjusted:

- encoding should behave like in other methods and refer to `Yii::$app->charset` by default.
- it should provide ucfirst and lcfirst functionality
- unit tests are needed to cover the functionality
- the functions may fit more for the Inflector class than the StringHelper

samdark:
It was implemented: https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseStringHelper.php#L431

PR #13953, opened 2017-04-11
Added description to load and unload functions

nurbeknurjanov:
This function are not realized as doc says. it is on discussing, it is hard for developers to read on doc this function works. but in fact it doesn't. They think that they did something wrong

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no

### DIFF
--- a/docs/guide/test-fixtures.md
+++ b/docs/guide/test-fixtures.md
@@ -295,7 +295,7 @@ change this behavior with config or command options. You can exclude some fixtur
 
 To load fixture, run the following command:
 
-At the moment this fucntion works this way, it clears the data before inserting fixtures. Anytime it resets the table.
+> Note: Prior to loading data unload sequence is executed. Usually that results in cleaning up all the existing data inserted by previous fixture executions.
 
 ```
 yii fixture/load <fixture_name>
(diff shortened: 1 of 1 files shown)


docs/guide/test-fixtures.md
```diff
@@ -295,6 +295,8 @@ change this behavior with config or command options. You can exclude some fixtur
 
 To load fixture, run the following command:
 
+At the moment this fucntion works this way, it clears the data before inserting fixtures. Anytime it resets the table.
```
samdark: Yes. This part is definitely missing. I'd change text to the following:

```
> Note: Prior to loading data unload sequence is executed. Usually that results in cleaning up all the existing data inserted by previous fixture executions.
```

docs/guide/test-fixtures.md
```diff
@@ -333,6 +335,8 @@ yii fixture User --globalFixtures='some\name\space\Custom'
 Unloading fixtures
 ------------------
 
+This function does not clear the data at the moment. Because this function on discussing by developers.
```
samdark: `unload()` method of fixtures should be triggered properly by this command. Could you provide steps to reproduce the issue?

samdark (changes requested):

samdark:
Merged note about `load()`. Thanks.

PR #13959, opened 2017-04-11
Fix `\yii\db\QueryInterface` type hints

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Not sure if we should specify `Expression` as one of the acceptable types. `\yii\db\QueryTrait` accepts it too. I will adjust the PR if needed.

### DIFF
--- a/framework/db/QueryInterface.php
+++ b/framework/db/QueryInterface.php
@@ -241,14 +241,14 @@ interface QueryInterface
 
     /**
      * Sets the LIMIT part of the query.
-     * @param int $limit the limit. Use null or negative value to disable limit.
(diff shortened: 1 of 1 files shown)


samdark:
Merged. Thanks!

Kolyunya:
Shouldn't we have added the `Expression` as well?

cebe:
Expression is not necessarily supported by non-SQL implementations so I would keep it out of the interface.

Kolyunya:
Thanks for the clarification!

Issue #13976, opened 2017-04-12
[IpValidator]  IPv6 validation can not be used. PHP is compiled without IPv6


mikk150:
### What steps will reproduce the problem?

Start validating some ip address usin IpValidator

### What is the expected result?

Should be fine?

### What do you get instead?

`yii\base\InvalidConfigException`: IPv6 validation can not be used. PHP is compiled without IPv6

### Additional info

PHP should not be compiled without IPv6 support...
My machine ( works )
```
$ php -a
Interactive mode enabled

php > var_dump(defined('AF_INET6'));
php shell code:1:
bool(true)
php > $hex = unpack('H*hex', inet_pton('2001:db8::1'));
php > echo substr(preg_replace('/([a-f0-9]{4})/i', '$1:', $hex['hex']), 0, -1);
2001:0db8:0000:0000:0000:0000:0000:0001
```

Container ( does not work )
```
$ docker run -it php
Interactive shell

php > var_dump(defined('AF_INET6'));
bool(false)
php > $hex = unpack('H*hex', inet_pton('2001:db8::1'));
php > echo substr(preg_replace('/([a-f0-9]{4})/i', '$1:', $hex['hex']), 0, -1);
2001:0db8:0000:0000:0000:0000:0000:0001

```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      |  7.0
| Operating system | Docker container

cebe:
where did you get the container, how was it created?

mikk150:
It's php docker container..
The problem is, that even though `var_dump(defined('AF_INET6'))` says that IPv6 is not installed `inet_*` methods that this validator relies on are working as they should

mikk150:
Yep
IpValidator works perfectly with without even this AF_INET6

```Dockerfile
FROM php:7.0

ENV COMPOSER_ALLOW_SUPERUSER=1

RUN apt-get update && apt-get install -y \
    libcurl4-gnutls-dev libssl-dev \
    zlib1g-dev libicu-dev g++ \
    git \
 && docker-php-ext-configure intl \
 && docker-php-ext-install -j$(nproc) pdo pdo_mysql curl intl zip \
 && git clone https://github.com/mikk150/yii2.git /usr/src/yii2 \
 && cd /usr/src/yii2 \
 && git checkout check-docker \
 && curl -sS https://getcomposer.org/installer | php -- \
 && php composer.phar global require fxp/composer-asset-plugin \
[...]

PR #13984, opened 2017-04-13
[IpValidator] disable ipv6 checks

mikk150:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes (at least IpValidatorTest did)
| Fixed issues  | #13976

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -64,6 +64,7 @@ Yii Framework 2 Change Log
 - Bug #13901: Fixed passing unused parameter to `formatMessage()` call in `\yii\validators\IpValidator` (Kolyunya)
 - Enh #13945: Removed Courier New from error page fonts list since it looks bad on Linux (samdark)
 - Bug #13961: RBAC Rules: PostgreSQL: PHP Warning "unserialize() expects parameter 1 to be string, resource given" was fixed (vsguts)
+- Enh #13976: Disabled IPv6 check on `\yii\validators\IpValidator` as it turns out it is not needed for inet_* methods to work (mikk150)
 
 2.0.11.2 February 08, 2017
 --------------------------
--- a/framework/validators/IpValidator.php
+++ b/framework/validators/IpValidator.php
@@ -221,7 +221,6 @@ class IpValidator extends Validator
         if (!$this->ipv4 && !$this->ipv6) {
             throw new InvalidConfigException('Both IPv4 and IPv6 checks can not be disabled at the same time');
         }
-        
         if ($this->message === null) {
             $this->message = Yii::t('yii', '{attribute} must be a valid IP address.');
         }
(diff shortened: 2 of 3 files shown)


cebe:
do you know why the constant did not exist in your environment.

mikk150:
I actually have no idea, but seems it doesn't matter for inet_* methods

mikk150:
Yep, confirmed.
My container environment has IPv6 support disabled..

SilverFire:
Checked, the fix is valid.

Could you add a line to CHANGELOG.md?

mikk150:
@SilverFire Like this?

SilverFire (approved):

SilverFire:
Yepp. Good to merge in 2.0.12. Any objections, team>

cebe (approved):

cebe (changes requested):
Seems this is relevant if `expandIPv6` is set to true. http://php.net/manual/en/function.inet-pton.php
inet_pton() does not work when PHP is not compiled with IPv6 support.

mikk150:
So we should watch php compile flags every time we need to validate IPv6? because as it turns out "AF_INET6" constant does not work?

mikk150:
Yes, I checked
I compiled PHP7 with --disable-ipv6 option, and
```
$hex = unpack('H*hex', inet_pton('2001:db8::1'));

Warning: inet_pton(): Unrecognized address 2001:db8::1 in php shell code on line 1
```
So AF_INET6 is certanly not the right way

SilverFire:
@cebe good catch!

@mikk150 is it possible to have `AF_INET6` set to true when PHP is compiled with `--disable-ipv6 option`?

mikk150:
Looking into PHP source code, AF_INET6 does not control inet_pton at all, HAVE_INET_PTON and HAVE_IPV6 are the only constants that matter, but these are only compile-time constants, so you cannot read them

SilverFire:
Have not dig too deep in PHP sources, but [2 lines below](https://github.com/php/php-src/blob/master/ext/standard/basic_functions.c#L3949) constant AF_INET6 is used. Not sure, how it will impact on code execution

mikk150:
This
https://github.com/php/php-src/blob/master/ext/standard/basic_functions.c#L3949
Is actually AF_INET6 that is defined in C library, and is not the same as PHP AF_INET6 BUT what is weird is that AF_INET6 should be defined in  https://github.com/php/php-src/blob/master/ext/sockets/sockets.c#L674 BUT I do not see that..

SilverFire (approved):

SilverFire:
Considering that `AF_INET6` constant does not protect from unexpected errors and we have not discovered a better way to check that IPv6 is supported I merge this PR.

SilverFire:
Added check to `requirements.php`
14a245a4135b90b12dd51f4c7f4ce25b1f0b413a

mikk150:
@SilverFire it has nothing to do with it, why did you add it?

SilverFire:
I've added additional commit https://github.com/yiisoft/yii2/commit/620ce9af191d036efe13018f5626487c9875c1c1 with enhanced wording

> Currently PHP constant <code>AF_INET6</code> is not defined and IPv6 is probably unsupported

It says **probably** because we are not sure. And requirement is defined as not mandatory, so it will just warn developer

Issue #13975, opened 2017-04-12
yii\web\Response::redirect() documentation is vague


micro-maureen:
The documentation states

> **Any relative URL** will be converted into an absolute one by prepending it with the host info of the current request.

### What steps will reproduce the problem?

Call `yii\web\Response::redirect('test.html');`

### What is the expected result?
The HTTP response has this header:

`Location: http://www.example.com/test.html`

### What do you get instead?

`Location: test.html`

### Proposed Solutions
1. Change the documentation of `yii\web\Response::redirect()`, `yii\web\Controller::redirect()` and possibly others to say
> **A relative URL that begins with a forward slash** will be converted into an absolute one by prepending it with the host info of the current request

2. Change `Response::redirect()` to use `Url::isRelative()` instead of its current logic.

cebe:
It seems relative Urls are okay, so we should review the behavior as automatically adding a host part may not be necessary for HTTP 1.1.

> An obsolete version of the HTTP 1.1 specifications (IETF RFC 2616) required a complete absolute URI for redirection.[2] The IETF HTTP working group found that the most popular web browsers tolerate the passing of a relative URL[3] and, consequently, the updated HTTP 1.1 specifications (IETF RFC 7231) relaxed the original constraint, allowing the use of relative URLs in Location headers.[4]
> https://en.wikipedia.org/wiki/HTTP_location

samdark:
Looks like a bug to me.

samdark:
That's a possibility. Anyway, current behavior doesn't look right.

PR #13986, opened 2017-04-13
Temporary solution for #13975 - make documentation of current behaviour more clear

micro-maureen:
Workaround for #13975 - document current behaviour more clearly.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no


samdark:
Merged. Thanks!

Issue #13957, opened 2017-04-11
Yii::setAlias


unasoft:
I encountered a problem when adding an alias, I added in bootstrap an alias @app but Yii overwrites it in
**/yii/base/Application**

`    public function setBasePath($path)
    {
        parent::setBasePath($path);
        Yii::setAlias('@app', $this->getBasePath());
    }`

What if you add to the functions, the parameter overwrite, like so `setAlias($alias, $path, $overwrite = true)`
and in the framework where **setAlias** is used to set the parameter `$overwrite` false, like so

`    public function setBasePath($path)
    {
        parent::setBasePath($path);
        Yii::setAlias('@app', $this->getBasePath(), false);
    }`

PS
Sorry. My english is bad.

samdark:
I think for this particular case we can skip setting an alias if it's already there.

cebe:
what is the use case for defining `@app` but not configuring application `basePath`?

ejsoft:
Directory structure: 
```
| app
  | --- backend
  | --- console
  | --- frontend
| public
  | --- backend
  | --- frontend
```

set in bootstrap.php `Yii::setAlias('@app', 'path to app directory')` and use `namespace app\backend or app\console or app\frontend`

I think it's better to make a choice, not to force set it without the ability to change

samdark:
Re-checked. Seems what you need could be achieved as @cebe suggested. Closing.

PR #13988, opened 2017-04-13
Add new param  for Yii::setAlias() enh #13957

unasoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13957

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -201,6 +201,7 @@ class BaseYii
      * alias translation by [[getAlias()]].
      * @param string $path the path corresponding to the alias. If this is null, the alias will
      * be removed. Trailing '/' and '\' characters will be trimmed. This can be
+     * @param boolean $overwrite the overwrite allow to overwrite the alliance if it is set, or skip if the alias is already set
      *
      * - a directory or a file path (e.g. `/tmp`, `/tmp/main.txt`)
      * - a URL (e.g. `http://www.yiiframework.com`)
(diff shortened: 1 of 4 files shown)


framework/BaseYii.php
```diff
@@ -201,6 +201,7 @@ public static function getRootAlias($alias)
      * alias translation by [[getAlias()]].
      * @param string $path the path corresponding to the alias. If this is null, the alias will
      * be removed. Trailing '/' and '\' characters will be trimmed. This can be
+     * @param boolean $overwrite the overwrite allow to overwrite the alliance if it is set, or skip if the alias is already set
```
Kolyunya: > overwrite the alliance

`alias`?

ejsoft: yeap, thanks

framework/BaseYii.php
```diff
@@ -210,11 +211,14 @@ public static function getRootAlias($alias)
      * @throws InvalidParamException if $path is an invalid alias.
      * @see getAlias()
      */
-    public static function setAlias($alias, $path)
+    public static function setAlias($alias, $path, $overwrite = true)
```
samdark: Public method signature change is a backwards compatibility break.

samdark: I don't think we need it as option. It should be default behavior not to overwrite it.

samdark (changes requested):

PR #13996, opened 2017-04-17
registerJsVar - Shorthand to add JS variable

Eseperio:
This method simplifies work when passing variables from php to javascript. If value is an array a json object will be created.
Usage: $this->registerJsVal($name,$value)

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | --


samdark:
Could be a security issue. I remember that JSON encode wasn't encoding for HTML.

klimov-paul:
Since there is a trend for modern web applications to be built as one-page site on such technology as AngularJS ro ReactJS, methods like `registerJs()` or `registerJsFile()` are less common to be used.
I prefer do NOT introduce another functionality to this layer as it is rare to be used.
Beside it does not seem to be a good practice to setup global variables even in JavaScript.

Eseperio:
I don´t think it so. ReactJS and AngularJS are frameworks and Javascript is native. Anyway Yii is a framework, and one of its benefits is make things less trivial and much clear.  On the other hand the risk exposed is the same that in method registerJS.

klimov-paul:
Proposed method is just a shortcut around `registerJs()`, which seems redundant to me.
If we agree to add it, then there should also method `registerJsFunction()`, `registerJsClass` and so on.

cebe:
@Eseperio check out the link I posted above, your code may have a security issue as strings are not properly encoded.

Eseperio:
Yes, htmlEncode is better.

> then there should also method registerJsFunction(), registerJsClass and so on.

That´s not true, because functions and classes can structured in múltiple ways, meanwhile variables always follow the same format.

To compare with other class check the Html helper. It has Html::tag() and Html::a(). Both ends in the same function. However there is no need to create Html::li, Html::div, etc..

framework/web/View.php
```diff
@@ -486,6 +486,49 @@ public function registerJsFile($url, $options = [], $key = null)
...
+        ]);
+
+        if($validator->validate($name))
+            throw new InvalidConfigException('Variable name must be in a valid ECMAscript format ');
+
+        if(is_array($value)){
+            $val=\yii\helpers\Json::htmlEncode($value);
+            $js=sprintf('var %s=%s',$name,$val);
+        }else {
+            $js=sprintf('var %s="%s"',$name,$value);
```
samdark: Should be HTML-encoded as well.

samdark (changes requested):

framework/web/View.php
cebe: there also is no need to differentiate between array and others.

Eseperio: Yes. It has been modified. Also empty name were not controlled. See [https://github.com/yiisoft/yii2/pull/13996/commits/17faff78b7be0908a6280ffec35aeeee91954c8e](url)

framework/web/View.php
```diff
@@ -486,6 +486,45 @@ public function registerJsFile($url, $options = [], $key = null)
...
+        ]);
+
+        if($validator->validate($name) || empty($name))
+            throw new InvalidConfigException('Variable name must be in a valid ECMAscript format ');
+
+        $val = \yii\helpers\Json::htmlEncode($value);
+        $js=sprintf('var %s=%s',$name,$val);
+
+        $this->registerJs($js,$position,strval($name));
+    }
```
MysteryDragon: There are some PSR-2 and core framework code style mistakes.
https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md

framework/web/View.php
```diff
@@ -516,6 +516,45 @@ public function registerJsFile($url, $options = [], $key = null)
...
+     * @param string $key the key that identifies the JS code block. If null, it will use
+     * $js as the key. If two JS code blocks are registered with the same key, the latter
+     * will overwrite the former.
+     *
+     * @throws InvalidConfigException If variable name does not match ECMA requirements.
+     * @since 2.0.14
+     */
+    public function registerJsVar($name, $value, $position = self::POS_HEAD)
+    {
+        $validator = new \yii\validators\RegularExpressionValidator([
```
rob006: Why not use `preg_match()` directly?

framework/web/View.php
```diff
@@ -516,6 +516,45 @@ public function registerJsFile($url, $options = [], $key = null)
...
+            'pattern' => '/^[^a-zA-Z_]+|[^a-zA-Z_0-9]+/',
```
rob006: This regexp is incorrect. See https://mathiasbynens.be/notes/javascript-identifiers#examples

I'm not sure if we should validate it at all. It should be a developer responsibility to use valid name for variable.

framework/web/View.php
```diff
@@ -516,6 +516,45 @@ public function registerJsFile($url, $options = [], $key = null)
...
+     *
+     * - [[POS_HEAD]]: in the head section. This is the default value.
+     * - [[POS_BEGIN]]: at the beginning of the body section.
+     * - [[POS_END]]: at the end of the body section.
+     * - [[POS_LOAD]]: enclosed within jQuery(window).load().
+     *   Note that by using this position, the method will automatically register the jQuery js file.
+     * - [[POS_READY]]: enclosed within jQuery(document).ready().
+     *   Note that by using this position, the method will automatically register the jQuery js file.
+     *
+     * @param string $key the key that identifies the JS code block. If null, it will use
```
rob006: There is no such param.

rob006 (changes requested):

framework/web/View.php
schmunk42: Reminds me of https://stackoverflow.com/questions/48270127/can-a-1-a-2-a-3-ever-evaluate-to-true :)

I'd say it's good to do this kind of validation, even if JavaScript could use other chars in the variable name - it's just crazy and from my understanding it will be in ES6 that this is no longer allowed. See also these numerous issues with browsers.

rob006: Right now this validation:
* allows invalid variable names (like keywords),
* disallows valid names (not only insane `var ಠ_ಠ = eval;` but also pretty common `var $ = jQuery;`.

It simply doesn't work and create unnecessary overhead.

> I'd say it's good to do this kind of validation, even if JavaScript could use other chars in the variable name - it's just crazy.

I'm not sure if Yii as PHP framework should force some name conventions for JavaScript variables.

schmunk42: Here's a regexp for the validation: https://stackoverflow.com/a/9337047/291573 - it's 11kB - 🤣 

When I think about compatibility to PHP or among browsers and various JS-engines, I'd rather add a strict validation - but I do not really care. If someone thinks it's a good idea to name variables in non-7-bit-ASCII, so be it...

schmunk42: Addon: https://mathiasbynens.be/notes/javascript-identifiers-es6

Some headlines from examples from that page:

// Valid in ES5, but invalid in an ES6 module context:
// Valid in ES5 and ES6:
// Invalid in ES5, but valid in ES6:
// Valid in ES5 & Unicode v5.1.0+, but invalid in ES6:
// Invalid in ES5 and ES6:
// Valid in ES5, but only works in some ES5 engines (i.e. those with Unicode data from v3.2.0 or more recent):
// Valid in ES5, but only works in some ES5 engines (i.e. those with Unicode data from v5.1.0 or more recent):
// Invalid in ES5 and ES6, but supported in old ES5 engines:

samdark: Nice list. We'd better remove validation.

samdark (approved):

rob006 (approved):

samdark:
Merged. Thanks for initial proposal and reviews.

PR #14001, opened 2017-04-18
Fixes #5442: Moved `resetTable` to `unload` method

### LINKED ISSUE #5442: ActiveFixture does not delete actual data in unload() method
ActiveFixture unload does not call `resetTable()` in `unload()` and this causes trouble when you have dependant fixtures with foreign key constraints. As ActiveFixture tries to `loadFixtures()`, which calls `resetTable()` it fails as the dependant data is not deleted. The issue description is related with #5441

leandrogehlen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #5442


samdark:
Wouldn't this change make tests stuck till you clear DB manually in case there was a failure?

leandrogehlen:
This was solved with the following change:
https://github.com/Codeception/Codeception/pull/4016

In the next execution the fist step is to execute de `unload`

samdark:
Ah, right. Need a changelog line then,

cebe:
`ActiveFixture` is not only used with codeception so it should call the method in both places or have an option to choose when to call it.

leandrogehlen:
Where more is used?

samdark:
Fixture itself could be used manually.

cebe:
> Where else is used?

with PHPUnit for example.

leandrogehlen:
Can be created `FixtureTrait::initFixtures` method?:

```php
public function initFixtures()
{
    $this->unloadFixtures();
    $this->loadFixtures();
}
```

samdark:
Yes.

leandrogehlen:
I think this issue needs to be in 2.012, because the `yii2-codeception` is deprecated and the yii2 codeception module does not allow to use https://github.com/yiisoft/yii2/blob/master/framework/test/InitDbFixture.php.
This way, currently, it's impossible to use fixtures with db-related

samdark (approved):

SilverFire (approved):

leandrogehlen:
related Codeception/Codeception#4201

samdark:
Merged. Thank you!

Issue #14006, opened 2017-04-18
Suggestion: DI container registers classes as singletons by default 


glagola:
Hi, I suggest to change current logic of class declaration for DI container. I think it should register singletons by default, it will improve performance and reduce memory consumption for projects which uses dependency injection constantly (like mine).

samdark:
Agree.

cebe:
there is an option to specify singletons if you want them, what exactly should be changed?

samdark:
@cebe defaults. i.e. singletons should be preferred (`set` method), non-singletons should require special method (`setNonSingleton` or `setFactory`).

klimov-paul:
> singletons should be preferred (set method), non-singletons should require special method (setNonSingleton or setFactory).

Disagreed: what about widgets registered at DI? It is unlikely they should be registered as singletons.
Also for me it sounds logical that for singletons there is a special method, while regular `set()` do not register singleton. Any regualr class is NOT a singleton, while creating a singleton class require additional explicit efforts.

samdark:
If we're talking about regular container usage (services), not passing every instance creation through it, then singletons are a common use.

klimov-paul:
Still, I prefer naming pair `set()` and `setSingleton()` insetad of `set()` and `setNotSingleton()` - the first one looks better.

cebe:
> If we're talking about regular container usage (services), not passing every instance creation through it, then singletons are a common use.

for services there is the servicelocator at Yii::$app, not the DI container.

samdark:
Well, we may remove it since these are basically the same thing now.

klimov-paul:
> these are basically the same thing now.

Why is that?
For me DI container take precedence over DI ServiceLocator.
While `Container` is a singleton - `ServiceLocator` can appear in multiple instances inside the program.

glagola:
Let me illustrate the problem:
```php
class Foo {
    public function __construct()
    {
        // Foo will be created twice
        echo "Foo", PHP_EOL;
    }
    
}

class Bar {
    public function __construct(Foo $foo)
    {
    
    }
}

class Hoo {
    public function __construct(Foo $foo, Bar $bar)
    {
    
    }
}

// To solve this issue I have to declare Foo as singleton
// \Yii::$container->setSingleton(Foo::class);

\Yii::$container->get(Hoo::class);
```

cebe:
> Well, we may remove it since these are basically the same thing now.

[...]

PR #14018, opened 2017-04-20
Small change which allows to reuse dependencies as singletons if they were not defined explicitly

glagola:
Small change which allows to reuse dependencies as singletons if they were not defined explicitly

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #14006

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -337,11 +337,11 @@ class BaseYii
     public static function createObject($type, array $params = [])
     {
         if (is_string($type)) {
-            return static::$container->get($type, $params);
+            return static::$container->get($type, $params, [], true);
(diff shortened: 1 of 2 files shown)


samdark:
https://travis-ci.org/yiisoft/yii2/builds/224128032

SilverFire:
This change makes singleton from all dependencies, if `$forceCreation` argument is not set to `true`, which defaults to `false` thus requires developers to change all `$container->get()` calls to achieve previous behavior. It means, that the change is backward-incompatible.

This change is completely wrong and will not be accepted. Thank you

Issue #13890, opened 2017-03-30
DbTarget log transaction bug


shirase:
### What steps will reproduce the problem?
Use DbTarget as targets for log config
Open transaction
Generate error

### What is the expected result?
Error is logged in db

### What do you get instead?
Error is not logged in db

### Additional info
Logger execute insert query, but active transaction is rollback.

### Solution
Rollback current transaction before insert log row.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11
| PHP version      | 5.6.2
| Operating system |

bizley:
Have you called `Yii::error()` before `rollBack()`? Call it after. It's up to you to do it properly, not the framework.

shirase:
I do not call Yii::error(), it called when error occur.
Example:
I start transaction, then try to get not existent property. I have exception, but i don'd see it in db.
I think errors must be logged any way. And that error is logged if i use file target for logger.

shirase:
As solution i can use additional db connection for logger, but it no obviously.

samdark:
Most likely it is a bug. If an error occurred, logging should happen after rolling back a transaction.

PR #14020, opened 2017-04-20
Bug #13890: DbTarget log transaction

shirase:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13890


samdark (approved):

samdark:
Merged. Thank you!

Love the cloning solution.

framework/log/DbTarget.php
```diff
@@ -60,6 +60,12 @@ public function init()
      */
     public function export()
     {
+        if ($this->db->getTransaction()) {
+            $this->db = clone $this->db;
+            $this->db->pdo = null;
```
cebe: sorry for the late reply, but does this work in a master-slave setup? The situation is a bit more complex there: https://github.com/yiisoft/yii2/blob/1c07c429cf54e58ccb494f05ed889e6dce284bbe/framework/db/Connection.php#L578-L597

also transaction property of the DB is not reset so for each call to export another connection is opened even though there is no transaction on current connection anymore.

samdark: master-slave should work well. I see no reason for it not to.

Transaction property — yes. Should be reset. Good catch.

cebe: see #14121

samdark: Ah. It could be slave transaction...

PR #14021, opened 2017-04-20
fix #14012 pgsql materializedview detection

### LINKED ISSUE #14012: function findViewNames() in yii\db\pgsql\Schema skip materialized views
I have some materialized views and try to use InitDbFixture in my test and got exception 
` [yii\db\Exception] SQLSTATE[42809]: Wrong object type: 7 ОШИБКА:  "my_materialized_view" - это не таблица и не сторонняя таблица`
that happens on stage 
```php
    public function beforeLoad()
    {
        $this->checkIntegrity(false);
    }
```

My code research  shows that problem in
```php
namespace yii\db\pgsql\Schema;
....
    protected function findViewNames($schema = '')
    {
        if ($schema === '') {
            $schema = $this->defaultSchema;
        }
        $sql = <<<SQL
SELECT c.relname AS table_name
FROM pg_class c
INNER JOIN pg_namespace ns ON ns.oid = c.relnamespace
WHERE ns.nspname = :schemaName AND c.relkind = 'v'
ORDER BY c.relname
SQL;
        return $this->db->createCommand($sql, [':schemaName' => $schema])->queryColumn();
    }
```

It skip only views,  but for materialized views sql must be 
```php
$sql = <<<SQL
SELECT c.relname AS table_name
FROM pg_class c
INNER JOIN pg_namespace ns ON ns.oid = c.relnamespace
WHERE ns.nspname = :schemaName AND (c.relkind = 'v' OR c.relkind = 'm' )
ORDER BY c.relname
SQL;
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | ~2.0.11
| PHP version      |7.1.3 
| Postgres | 9.6.1

Insolita:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no (this situation not covered as i see - tests\data\postgres.sql not contains creation of materialized view , and i'm not sure, should i make special tests for it )
| Fixed issues  | #14012

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.12 under development
 --------------------------
-- Bug #14012: function findViewNames() in yii\db\pgsql\Schema skip materialized views (insolita)
+- Bug #14012: `yii\db\pgsql\Schema::findViewNames()` was skipping materialized views (insolita)
 - Bug #13362: Fixed return value of `yii\caching\MemCache::setValues()`  (masterklavi)
 - Enh #13963: Added tests for yii\behaviors\TimestampBehavior (vladis84)
 - Enh #13820: Add new HTTP status code 451 (yyxx9988)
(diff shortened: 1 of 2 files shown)


samdark:
Merged. Thanks!

PR #14033, opened 2017-04-22
Fix `yii\filters\AccessRule::matchIP()`

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

[Here](https://github.com/yiisoft/yii2/blob/master/framework/filters/AccessRule.php#L112) `matchIP()` may be passed a `null` which will be passed to `strncmp()` [here](https://github.com/yiisoft/yii2/blob/master/framework/filters/AccessRule.php#L177).
`strncmp()` expects to be receive a `string`, not `null` and it [seems to be triggering a warning](https://3v4l.org/sFe7o) on an `HHVM`..

### DIFF
--- a/framework/filters/AccessRule.php
+++ b/framework/filters/AccessRule.php
@@ -165,7 +165,7 @@ class AccessRule extends Component
     }
 
     /**
-     * @param string $ip the IP address
+     * @param string|null $ip the IP address
      * @return bool whether the rule applies to the IP address
      */
     protected function matchIP($ip)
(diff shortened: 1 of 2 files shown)


framework/filters/AccessRule.php
```diff
@@ -174,7 +174,7 @@ protected function matchIP($ip)
             return true;
         }
         foreach ($this->ips as $rule) {
-            if ($rule === '*' || $rule === $ip || (($pos = strpos($rule, '*')) !== false && !strncmp($ip, $rule, $pos))) {
+            if ($rule === '*' || $rule === $ip || (($pos = strpos($rule, '*')) !== false && $ip !== null && !strncmp($ip, $rule, $pos))) {
```
Kolyunya: I know, that condition should be refactored. I'll make an update to a PR if it's going to be adopted.

cebe (approved):

SilverFire (approved):

samdark:
Merged. Thanks!

Kolyunya:
Glad to help!

Issue #13968, opened 2017-04-12
yii\rest\UrlRule doesn't support route params.


Faryshta:
### What steps will reproduce the problem?

on config/main.php

```php
'urlManager' => [
    'rules' => [
        [
            'class' => \yii\rest\UrlRule::class,
            'controller' => ['store/<store_id:\d+>/employee' => 'store-employee'],
        ],
    ],
],
```

### What is the expected result?

use the `StoreEmployeeController` controller when accessing the route store/1/employee 

### What do you get instead?

http status 404 the rest rule is not even parsed.

### Additional info

caused by line https://github.com/yiisoft/yii2/blob/master/framework/rest/UrlRule.php#L221 since it executes

`strpos('store/1/employee', 'store/<store_id:\d+>/employee')`

which obviously will return false. there is no work-around this. when removing this line (and the ending bracket) the url is parsed properly.

cebe:
Your config looks really weird, why are you configuring the class of UrlManager to be yii\rest\UrlRule?
Also, a controller name can not contain variables.

Faryshta:
@cebe my bad i fixed the original example

Faryshta:
@cebe my bad i fixed the original example

cebe:
From the PHPdoc:

> the controller ID (e.g. `user`, `post-comment`) that the rules in this composite rule are dealing with.

what makes you think it could contain parameters?

Faryshta:
if it doesn't support route params then lets add support to route params.

http://www.yiiframework.com/doc-2.0/yii-rest-urlrule.html#$controller-detail

> you want to explicitly specify how the controller ID should appear in the **patterns*, you may use an array with the array key being as the controller ID in the **pattern**, and the **array value the actual controller ID**. For example, ['u' => 'user'].

it make sense to be able to have a pattern like the one i mention on my example.

Faryshta:
@cebe so can i make a pr with the support for this?

cebe:
I do not think this is a good idea to support this inside $controller property. Using $prefix looks like a better solution to me.

tunecino:
@Faryshta see https://github.com/yiisoft/yii2/issues/9474

Faryshta:
@cebe thats suboptimal specially when handling all the **controllers** in an api

```php
    'rules' => [
        [
            'prefix' => 'api/v1/',
            'class' => \yii\rest\UrlRule::class,
            'controller' => [
                'store',
[...]

Issue #9474, opened 2015-08-19
[DOC] Creating nested RESTful resources routes


creocoder:
This is how that can be done:

``` php
'rules' => [
    [
        'class' => 'yii\rest\UrlRule',
        'controller' => ['collaborators' => 'v1/collaborator'],
        'prefix' => 'v1/projects/<projectId:\d+>',
    ],
],
```

After that api will have next endpoints:
- /projects/1/collaborators
- /projects/1/collaborators/1

P.S. All issues which requests such feature after that can be complettely closed. Feature there from start.

samdark:
Would be cool if you'll help with docs. Also it would be great if you'll point to these other issues.

mdmunir:
I hope we can do this
```php
'rules' => [
    [
        'class' => 'yii\rest\UrlRule',
        'controller' => [
            'v1/<controller>' => '<controller>',
            'v2/<controller>' => 'v2/<controller>',
        ],
    ],
],
```

creocoder:
@mdmunir Its not possible, this syntax is already taken.

fernandezekiel:
this is most straightforward implementation we can have at the moment, we probably can create an example application utilizing nested routes

fernandezekiel:
the application will have to demonstrate not just query adjustment based on the nested URL but also modification of the model to be created on something like:|
```
POST /projects/1/collaborators
``

**Comment by fernandezekiel** [comment]
here's my suggested example:
https://github.com/yiisoft/yii2/issues/7184#issuecomment-73396628

**Comment by tunecino** [comment]
@creocoder I think you mean : 
```php
'rules' => [
    [
        'class' => 'yii\rest\UrlRule',
        'controller' => ['collaborators' => 'v1/collaborator'],
        'prefix' => 'projects/<projectId:\d+>',
    ],
],
```
instead of ```'prefix' => 'v1/projects/<projectId:\d+>'``` or am I wrong ?

One more think I'm not getting. If docs is saying : 
```
$prefix: The common prefix string shared by all patterns.
```
Then how to do to get only ``` /projects``` or ``` /projects/1`` or a different not-related collection to preject ?

tunecino:
mmm I finally understood what you've been trying to do. I didn't know that we can chain rules :  
```php
'rules' => [
    [
        'class' => 'yii\rest\UrlRule', 
        'controller' => [
            'projects' => 'v1/project',
            'collaborators' => 'v1/collaborator'                        
        ],
    ],
    [
[...]

Issue #7878, opened 2015-03-25
Feature Request: Nested Resources (Nested Routes)


saada:
In the Rails world they use "Resources" to describe nested routes. Here's a reference for further reading. 
http://guides.rubyonrails.org/routing.html#nested-resources

It would be great to add this feature to Yii2.

```ruby
class Magazine < ActiveRecord::Base
  has_many :ads
end
 
class Ad < ActiveRecord::Base
  belongs_to :magazine
end
```

```ruby
resources :magazines do
  resources :ads
end
```

The above code gives you these routes out of the box
```
GET      /magazines/:magazine_id/ads
GET      /magazines/:magazine_id/ads/new
POST     /magazines/:magazine_id/ads
```

usualdesigner:
+1

usualdesigner:
BTW some REST libraries (like restangular) allow you to do it out the box. So, if this ability will be on the server side - it would be really cool.

mikehaertl:
+1

samdark:
Any suggestions for PHP syntax for the feature? i.e. how you want it to look in Yii terms when using it.

klimov-paul:
This could be a special URL rule class, which accepts 2 config parameters: 'modelClass' and 'relations':
```php
'rules' => [
    [
        'class' => 'yii\rest\NestedModelUrlRule',
        'modelClass' => 'app\models\Magazine',
        'relations' => ['ads'],
    ],
],
```

lynicidn:
@klimov-paul :+1:

saada:
@klimov-paul, I like that! 
I would suggest doing it at the controller level.
Something along the below lines...

```php
MagazineController extends Controller {
    public function resources() {
         return [
              AdsController::className()
         ];
    }
}
```

usualdesigner:
@klimov-paul it looks very good. It would be great to use arrays instead of string if we nee. E.g. 
```php
'rules' => [
    [
        'class' => 'yii\rest\NestedModelUrlRule',
        'modelClass' => 'app\models\Magazine',
        'relations' => [
            'ads' => [
                'relations' => [
                    'comments'
                ]
            ]
        ],
    ],
],
```

saada:
@usualdesigner , would your example allow all these routes to be exposed?
```
/magazine/:magazine_id:/ads
/magazine/:magazine_id:/ad/:ad_id:/comments
/ad/:ad_id:/comments
/ad/:ad_id:/comment/4
```

[...]

PR #14040, opened 2017-04-24
unit tests for yii\rest\UrlRule with prefix params

Faryshta:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

related #13968 #9474  #7878

### DIFF
--- a/tests/framework/rest/UrlRuleTest.php
+++ b/tests/framework/rest/UrlRuleTest.php
@@ -35,7 +35,10 @@ class UrlRuleTest extends TestCase
     public function testParseRequest()
     {
         $manager = new UrlManager(['cache' => null]);
-        $request = new Request(['hostInfo' => 'http://en.example.com', 'methodParam' => '_METHOD',]);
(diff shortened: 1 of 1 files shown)


tests/framework/rest/UrlRuleTest.php
```diff
@@ -340,19 +383,23 @@ public function createUrlDataProvider()
...
     {
+        $this->mockWebApplication();
+        Yii::$app->set('request', new Request([
+            'hostInfo' => 'http://api.example.com',
+            'scriptUrl' => '/index.php'
+        ]));
+        $manager = new UrlManager([
+            'cache' => null,
+        ]);
+        $rule = new UrlRule($rule);
```
bscheshirwork: `$rule = new UrlRule($rule);` is dangerous and will be changed #14998 

and I like this
realization out of foreach is better for me, I do it same

PR #14043, opened 2017-04-24
Expose some useful functions in IpValidator

cebe:
Expose some useful functions in IpValidator:

- `expandIPv6()`
- `inRange()`

these can be useful outside of the context of validating an IP address.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | need to add tests for these as they are public now
| Fixed issues  | none

### DIFF
--- a/framework/validators/IpValidator.php
+++ b/framework/validators/IpValidator.php
@@ -411,7 +411,7 @@ class IpValidator extends Validator
      * @param string $ip the original IPv6
      * @return string the expanded IPv6
      */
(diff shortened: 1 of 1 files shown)


SilverFire:
Maybe it's more natural to create `IpHelper`?

samdark:
Agree with @SilverFire.

SilverFire:
Closing in favor of #15271

Issue #14042, opened 2017-04-24
ambiguous error in unique validator


sepidemahmoodi:
### What steps will reproduce the problem?
Hi,
I've `department` and `college` table with this structure :

```php
$this->createTable('department', [
            'id' => $this->primaryKey(),
            'code' => $this->string()->notNull(),
            'collegeId' => $this->integer()->notNull(),
            'createdAt' => $this->integer(),
            'updatedAt' => $this->integer(),
        ]);

$this->createTable('department_info', [
       'id' => $this->primaryKey(),
       'name' => $this->string(),
       'languageId' => $this->string(),
       'description' => $this->text(),
      'departmentId' => $this->integer()->notNull(),
 ]);

$this->createTable('college', [
            'id' => $this->primaryKey(),
            'code' => $this->string()->notNull(),
            'createdAt' => $this->integer(),
            'updatedAt' => $this->integer(),
        ]);

$this->createTable('college_info', [
            'id' => $this->primaryKey(),
            'name' => $this->string(),
            'languageId' => $this->string(),
            'description' => $this->text(),
            'collegeId' => $this->integer()->notNull(),
            'websiteId' => $this->integer(),
            'url' => $this->text(),
]);
```

in rules of Department model `code` attribute is unique :

```php
 public function rules()
    {
        return [
            [['code', 'collegeId'], 'required'],
            [['code'], 'unique'],
            [['code'], 'integer'],
        ];
    }
```
then this model use uniqueValidator.
In special condition in `Department::find()` method i'm trying to join that tables :

```php
public static function find()
    {
        $query = new ActiveQuery(get_called_class());
        if(Yii::$app->user->isSuperuser()){
            return $query;
        }
        $query->joinWith('college');
        return $query;
    } 
```
while update one record of department this error was occured :

```sql
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in field list is ambiguous
The SQL being executed was:
 SELECT `id` FROM `department` LEFT JOIN `college` ON `department`.`collegeId` = `college`.`id` WHERE (`code`='1') LIMIT 2
```
3 month ago i wrote #13704 issue and some changes in master branch was occured but seems uniqueValidator already has some problems.

in `UniqueValidator::modelExists($targetClass, $conditions, $model)` 
in this part :

```php
if ($query instanceof \yii\db\ActiveQuery) {
                // only select primary key to optimize query
                $query->select($targetClass::primaryKey());
    }
```
[...]

PR #14049, opened 2017-04-25
Unique and Exist validator fixes

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13842, #14042


framework/db/ActiveQuery.php
```diff
@@ -796,4 +789,89 @@ public function alias($alias)
...
         return $this;
     }
+
+    /**
+     * Table names based on from
+     * @return string[] table names
+     * @throws \yii\base\InvalidConfigException
+     * @since 2.0.12
+     */
+    public function getFromTableNames()
```
samdark: It's not used for a fix. Not sure it needs to be added at all.

framework/db/ActiveQuery.php
```diff
@@ -796,4 +789,89 @@ public function alias($alias)
...
+
+        return $tableNames;
+    }
+
+    /**
+     * @return string[] table aliases based on from
+     * @throws \yii\base\InvalidConfigException
+     * @since 2.0.12
+     */
+    public function getFromAliases()
```
samdark: Only first alias is actually used in unique validator. Not sure we need all aliases at all.

framework/db/ActiveQuery.php
```diff
@@ -796,4 +789,89 @@ public function alias($alias)
...
+    {
+        $cleanedUpTableName = preg_replace('/\'|\"|`|as/u', '', trim($tableName));
+        return preg_replace('/^.+\s+(\w+)$/', '$1', $cleanedUpTableName);
+    }
+
+    /**
+     * @return string primary table name
+     * @since 2.0.12
+     */
+    protected function getPrimaryTableName()
```
samdark: This one is done for refactoring only. Could be left as it was before.

samdark:
@vladis84 check the code. That's what I was talking about.

framework/db/ActiveQuery.php
klimov-paul: I would rather remove it

framework/db/ActiveQuery.php
klimov-paul: Method name does not sound good: `from` word sounds like preposition instead of noun here.

klimov-paul:
Same fix should be added to `ExistValidator` as well, as it has the same problem.

cebe:
This implements parts of what is discussed in #7263

framework/db/ActiveQuery.php
Faryshta: i think staying DRY is better and keeping it.

framework/db/ActiveQuery.php
samdark: Any suggestions for a better name?

samdark:
@klimov-paul added fix for `ExistValidator`.

framework/db/ActiveQuery.php
```diff
@@ -796,4 +789,89 @@ public function alias($alias)
...
+    public function getFromAliases()
+    {
+        $tablesAlias = [];
+
+        if (empty($this->from)) {
+            return [$this->getPrimaryTableName()];
+        }
+
+        if (is_string($this->from)) {
+           $tableNames = preg_split('/\s*,\s*/', trim($this->from), -1, PREG_SPLIT_NO_EMPTY);
```
SilverFire: This should not happen in normal conditions. Method `from()` normalizes string to array before butting it to `$this->from`

framework/db/ActiveQuery.php
SilverFire: Agree with @klimov-paul - it's a bad name, especially for the public API.

SilverFire (changes requested):

rob006:
This could be simplified a lot if #11646 will be merged.

samdark:
@cebe can you take a look at #11646?

cebe:
Sure, but its not trivial to handle. I already spent a lot of time on that issue and finding a good solution is quite hard.

samdark:
Then I'm going to merge this one as is.

samdark:
Adjusted method names.

cebe:
no need to hurry with this.

tests/framework/db/ActiveQueryTest.php
```diff
@@ -231,4 +231,82 @@ public function testAlias_yet_set()
         $this->assertInstanceOf('yii\db\ActiveQuery', $result);
         $this->assertEquals(['alias' => 'old'], $result->from);
     }
+
```
dynasource: dataProvider seems logical here?

framework/db/ActiveQuery.php
```diff
@@ -796,4 +789,89 @@ public function alias($alias)
...
     }
+
+    /**
+     * Table names based on [[from]]
+     * @return string[] table names
+     * @throws \yii\base\InvalidConfigException
+     * @since 2.0.12
+     */
+    public function getTableNamesUsedInFrom()
+    {
```
dynasource: this method is the only one about getting tableNames. Shouldnt this get a cleaner name like 'getTableNames' or 'extractTableNames'

framework/db/ActiveQuery.php
```diff
@@ -796,4 +789,89 @@ public function alias($alias)
...
+        return $tableNames;
+    }
+
+    /**
+     * @return string[] table aliases based on [[from]]
+     * @throws \yii\base\InvalidConfigException
+     * @since 2.0.12
+     */
+    public function getAliasesUsedInFrom()
+    {
```
dynasource: same here. The question is how durable such naming is for publishing it in the public api. Does it matter if it is called 'getAliases' to show 'what is does' instead of 'how it does it'?

dynasource (changes requested):

framework/db/ActiveQuery.php
samdark: There are table names used in sub-queries and other cases. The method works only on `from` clause.

tests/framework/db/ActiveQueryTest.php
samdark: Yes and no. These are alike cases but checks vary so I'd keep it as is.

framework/db/ActiveQuery.php
samdark: `usedInFrom` is important part of what it does. It returns aliases for `from` clause and not for `join` or subqueries.

dynasource (approved):

framework/validators/ExistValidator.php
```diff
@@ -137,14 +139,41 @@ private function prepareConditions($targetAttribute, $model, $attribute)
...
+        if (!$model instanceof ActiveRecord) {
+            return $conditions;
+        }
+
+        // Add table prefix for column
+        $targetClass = $this->getTargetClass($model);
+
+        /** @var ActiveRecord $targetClass */
+        $query = $targetClass::find();
+        $tableAliases = $query->getAliasesUsedInFrom();
```
Faryshta: `getAliasesUsedInFrom()` is not part of `ActiveQueryInterface`. should it be added there?

samdark: @Faryshta it belongs to SQL DBs only so I think it's OK.

Faryshta: ok then. so you mean ExistValidator doesn't work with mongo?

samdark: It will work with mongo well. See line 150:

```php
if (!$model instanceof ActiveRecord) {
    return $conditions;
}
```

ActiveRecord in this code is `\yii\db\ActiveRecord`.

framework/db/ActiveQuery.php
cebe: what is the purpose of this method? it returns alias names but does not relate them to the table names as far as I see. Wouldn't it be better to have one method that returns an array of 'alias' => 'tablename' entries. Just having a list of aliases without reference seems not useful to me.

cebe (changes requested):
This one can be done without affecting the solution to #7263 much. But I'd like to evaluate the implementation using one method instead of two, as pointed out in my comment: https://github.com/yiisoft/yii2/pull/14049#discussion_r115429569

samdark:
@cebe done.

cebe (approved):

framework/validators/ExistValidator.php
```diff
@@ -137,14 +139,41 @@ private function prepareConditions($targetAttribute, $model, $attribute)
...
-            $params = [$targetAttribute => $model->$attribute];
+            $conditions = [$targetAttribute => $model->$attribute];
         }
-        return $params;
+
+        if (!$model instanceof ActiveRecord) {
+            return $conditions;
+        }
+
+        // Add table prefix for column
```
dynasource: this comment looks to be bound solely for the next line and is confusing. A private method for the whole next block would improve readability

framework/validators/UniqueValidator.php
```diff
@@ -249,14 +254,18 @@ private function prepareConditions($targetAttribute, $model, $attribute)
...
+
+        /** @var ActiveRecord $targetClass */
+        $query = $targetClass::find();
+        $tableAliases = array_keys($query->getTablesUsedInFrom());
+        $primaryTableAlias = $tableAliases[0];
+        $prefixedConditions = [];
         foreach ($conditions as $columnName => $columnValue) {
-            $prefixedColumnName = "{$tableName}.$columnName";
-            $conditionsWithTableName[$prefixedColumnName] = $columnValue;
+            $prefixedColumn = "{{{$primaryTableAlias}}}.[[{$columnName}]]";
```
dynasource: this is repeated logic, implement DRY?

framework/validators/ExistValidator.php
```diff
@@ -137,14 +139,41 @@ private function prepareConditions($targetAttribute, $model, $attribute)
...
+        $targetClass = $this->getTargetClass($model);
```
dynasource: $targetModelClass would improve readability

dynasource (changes requested):
last critical review

samdark:
@dynasource done.

PR #14051, opened 2017-04-26
Added copyEmptyDirectories option to FileHelper

### LINKED ISSUE #9669: AssetBundle $publishOptions['only'] still publishes empty directories
I currently have the following asset bundle:

``` php
class FontAwesomeAsset extends AssetBundle
{
    public $sourcePath = '@bower/components-font-awesome/';

    public $publishOptions = [
        'only' => [
            'css/*',
            'fonts/*',
        ],
    ];

    public $css = [
        'css/font-awesome.min.css',
    ];
}

```

Works good, but for some reason I still end up with the empty directories that aren't included in the "only" list. (less / scss). I think the "only" should apply to directories too, not just files.

image

cebe:
also set it to false in AssetManager to avoid creating a lot of empty
directories. fixes #9669

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #9669

### DIFF
--- a/tests/framework/helpers/FileHelperTest.php
+++ b/tests/framework/helpers/FileHelperTest.php
@@ -791,13 +791,7 @@ class FileHelperTest extends TestCase
         }
     }
 
-    /**
-     * @see https://github.com/yiisoft/yii2/issues/9669
-     *
-     * @depends testCopyDirectory
-     * @depends testFindFiles
-     */
-    public function testCopyDirectoryEmptyDirectories()
+    private function setupCopyEmptyDirectoriesTest()
     {
         $srcDirName = 'test_empty_src_dir';
         $this->createFileStructure([
@@ -814,8 +808,21 @@ class FileHelperTest extends TestCase
             ],
         ]);
 
-        $basePath = $this->testFilePath;
-        $srcDirName = $basePath . DIRECTORY_SEPARATOR . $srcDirName;
+        return [
+            $this->testFilePath, // basePath
+            $this->testFilePath . DIRECTORY_SEPARATOR . $srcDirName,
+        ];
+    }
+
+    /**
+     * @see https://github.com/yiisoft/yii2/issues/9669
+     *
+     * @depends testCopyDirectory
+     * @depends testFindFiles
+     */
+    public function testCopyDirectoryEmptyDirectories()
+    {
+        list($basePath, $srcDirName) = $this->setupCopyEmptyDirectoriesTest();
 
         // copy with empty directories
         $dstDirName = $basePath . DIRECTORY_SEPARATOR . 'test_empty_dst_dir';
@@ -830,6 +837,17 @@ class FileHelperTest extends TestCase
         $this->assertFileExists($dstDirName . DIRECTORY_SEPARATOR . 'dir1' . DIRECTORY_SEPARATOR . 'file2.txt');
         $this->assertFileExists($dstDirName . DIRECTORY_SEPARATOR . 'dir2');
         $this->assertFileExists($dstDirName . DIRECTORY_SEPARATOR . 'dir3');
+    }
+
+    /**
+     * @see https://github.com/yiisoft/yii2/issues/9669
+     *
+     * @depends testCopyDirectory
+     * @depends testFindFiles
+     */
+    public function testCopyDirectoryNoEmptyDirectories()
+    {
+        list($basePath, $srcDirName) = $this->setupCopyEmptyDirectoriesTest();
 
         // copy without empty directories
         $dstDirName = $basePath . DIRECTORY_SEPARATOR . 'test_empty_dst_dir2';
--- a/tests/framework/helpers/FileHelperTest.php
+++ b/tests/framework/helpers/FileHelperTest.php
@@ -790,4 +790,59 @@ class FileHelperTest extends TestCase
             $this->assertStringEqualsFile($fileName, $content, 'Incorrect file content!');
         }
     }
+
+    /**
+     * @see https://github.com/yiisoft/yii2/issues/9669
+     *
+     * @depends testCopyDirectory
+     * @depends testFindFiles
+     */
+    public function testCopyDirectoryEmptyDirectories()
+    {
+        $srcDirName = 'test_empty_src_dir';
+        $this->createFileStructure([
+            $srcDirName => [
+                'dir1' => [
+                    'file1.txt' => 'file1',
+                    'file2.txt' => 'file2',
+                ],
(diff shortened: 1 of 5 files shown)


tests/framework/web/AssetBundleTest.php
```diff
@@ -27,6 +28,24 @@ protected function setUp()
         Yii::setAlias('@testAssetsPath', '@webroot/assets');
         Yii::setAlias('@testAssetsUrl', '@web/assets');
         Yii::setAlias('@testSourcePath', '@webroot/assetSources');
+
+        // clean up assets directory
+        $handle = opendir($dir = Yii::getAlias('@testAssetsPath'));
+        if ($handle === false) {
+            throw new \Exception("Unable to open directory: $dir");
```
samdark: RuntimeException?

samdark (approved):

tests/framework/web/AssetBundleTest.php
```diff
@@ -126,11 +144,35 @@ public function testSourcesPublish_AssetBeforeCopy()
...
+            ],
+        ]);
+        $bundle->publish($am);
+
+        $notNeededFilesDir = $bundle->basePath . DIRECTORY_SEPARATOR . 'css';
+        $this->assertFileNotExists($notNeededFilesDir);
+
+        foreach ($bundle->js as $filename) {
+            $publishedFile = $bundle->basePath . DIRECTORY_SEPARATOR . $filename;
+            $this->assertTrue(unlink($publishedFile));
```
dynasource: this looks like you are testing the 'unlink' function

tests/framework/web/AssetBundleTest.php
```diff
@@ -126,11 +144,35 @@ public function testSourcesPublish_AssetBeforeCopy()
...
+        }
+        $this->assertTrue(rmdir($bundle->basePath . DIRECTORY_SEPARATOR . 'js'));
```
dynasource: same here for the rmdir method. What exactly do you want to test here? I expect some framework operations here.

dynasource (changes requested):

tests/framework/web/AssetBundleTest.php
samdark: It's testing for file existence.

tests/framework/web/AssetBundleTest.php
samdark: It's testing for directory existence.

tests/framework/web/AssetBundleTest.php
dynasource: is it? Or is it about checking wheter the file can be unlinked (due to certain file permissions?). 
Would be helpful to add an $error message?

cebe: it is cleanup and test in one step. unlink returns true if it was deleted, i.e. it existed, but it will also be removed as it is not needed anymore. That test is copied from similar tests in the same file so I would keep it as is.

tests/framework/web/AssetBundleTest.php
cebe: not sure, what do you mean? this is just to fail tests in case of permission issues.

tests/framework/web/AssetBundleTest.php
dynasource: don't you want tests to be clear and simple?

samdark: Agree with @dynasource. It's better to do it in two steps.

tests/framework/web/AssetBundleTest.php
samdark: OK.

tests/framework/web/AssetBundleTest.php
cebe: adjusted tests.

tests/framework/web/AssetBundleTest.php
```diff
@@ -126,12 +141,36 @@ public function testSourcesPublish_AssetBeforeCopy()
...
+        $bundle = new TestSourceAsset([
+            'publishOptions' => [
+                'only' => [
+                    'js/*'
+                ]
+            ],
+        ]);
+        $bundle->publish($am);
+
+        $notNeededFilesDir = $bundle->basePath . DIRECTORY_SEPARATOR . 'css';
```
dynasource: looking at the code above, this reference to CSS feels a bit magical and error prone. This should be tied to $bundle->css[0]

tests/framework/web/AssetBundleTest.php
```diff
@@ -126,12 +141,36 @@ public function testSourcesPublish_AssetBeforeCopy()
...
+        $this->assertFileNotExists($notNeededFilesDir);
+
+        foreach ($bundle->js as $filename) {
+            $publishedFile = $bundle->basePath . DIRECTORY_SEPARATOR . $filename;
+            $this->assertFileExists($publishedFile);
+        }
+        $this->assertTrue(is_dir($bundle->basePath . DIRECTORY_SEPARATOR . 'js'));
```
dynasource: this could be tied to dirname(@bundle->js[0])

tests/framework/helpers/FileHelperTest.php
```diff
@@ -790,4 +790,59 @@ public function testCopyDirectoryExclude()
...
+        $this->assertFileExists($dstDirName, 'Destination directory does not exist!');
+        $copiedFiles = FileHelper::findFiles($dstDirName);
+        $this->assertCount(2, $copiedFiles, 'wrong files count copied');
+
+        $this->assertFileExists($dstDirName . DIRECTORY_SEPARATOR . 'dir1');
+        $this->assertFileExists($dstDirName . DIRECTORY_SEPARATOR . 'dir1' . DIRECTORY_SEPARATOR . 'file1.txt');
+        $this->assertFileExists($dstDirName . DIRECTORY_SEPARATOR . 'dir1' . DIRECTORY_SEPARATOR . 'file2.txt');
+        $this->assertFileExists($dstDirName . DIRECTORY_SEPARATOR . 'dir2');
+        $this->assertFileExists($dstDirName . DIRECTORY_SEPARATOR . 'dir3');
+
```
dynasource: this test should be split  in more tests with clear 3 parts:
1) test setup
2) execution
3) assertions

This will make the test a lot easier to read & to debug.

dynasource (changes requested):
looked into it with little more depth.

samdark (approved):

PR #14056, opened 2017-04-26
added option to disable query logging in DB command

### LINKED ISSUE #12528: Database logging causes bad performance for large queries

cebe:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | waiting for travis
| Fixed issues  | #12528

fixes #12528

### DIFF
--- a/framework/db/Command.php
+++ b/framework/db/Command.php
@@ -829,9 +829,12 @@ class Command extends Component
     {
         $sql = $this->getSql();
 
-        $rawSql = $this->getRawSql();
-
-        Yii::info($rawSql, __METHOD__);
+        $log = $this->db->enableQueryLog;
+        if ($log) {
+            $rawSql = $this->getRawSql();
+            Yii::info($rawSql, __METHOD__);
+            $token = $rawSql;
+        }
 
         if ($sql == '') {
             return 0;
@@ -839,20 +842,19 @@ class Command extends Component
 
         $this->prepare(false);
 
-        $token = $rawSql;
         try {
-            Yii::beginProfile($token, __METHOD__);
+            $log && Yii::beginProfile($token, __METHOD__);
 
             $this->pdoStatement->execute();
             $n = $this->pdoStatement->rowCount();
 
-            Yii::endProfile($token, __METHOD__);
+            $log && Yii::endProfile($token, __METHOD__);
 
             $this->refreshTableSchema();
 
             return $n;
         } catch (\Exception $e) {
-            Yii::endProfile($token, __METHOD__);
+            $log && Yii::endProfile($token, __METHOD__);
             throw $this->db->getSchema()->convertException($e, $rawSql);
         }
     }
@@ -868,9 +870,12 @@ class Command extends Component
      */
     protected function queryInternal($method, $fetchMode = null)
     {
-        $rawSql = $this->getRawSql();
-
-        Yii::info($rawSql, 'yii\db\Command::query');
+        $log = $this->db->enableQueryLog;
+        if ($log) {
+            $rawSql = $this->getRawSql();
+            Yii::info($rawSql, 'yii\db\Command::query');
+            $token = $rawSql;
+        }
 
         if ($method !== '') {
             $info = $this->db->getQueryCacheInfo($this->queryCacheDuration, $this->queryCacheDependency);
@@ -883,7 +888,7 @@ class Command extends Component
                     $fetchMode,
                     $this->db->dsn,
                     $this->db->username,
-                    $rawSql,
+                    isset($rawSql) ? $rawSql : $this->getRawSql(),
                 ];
                 $result = $cache->get($cacheKey);
                 if (is_array($result) && isset($result[0])) {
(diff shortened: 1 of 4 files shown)


samdark (approved):

framework/db/Connection.php
```diff
@@ -373,6 +373,13 @@ class Connection extends Component
      * @see masters
      */
     public $shuffleMasters = true;
+    /**
+     * @var bool whether to enable logging and profiling of database queries. Defaults to true.
+     * You may want to disable this option in a production environment to gain performance
+     * if you do not need the information being logged.
+     * @since 2.0.12
+     */
+    public $enableQueryLog = true;
```
klimov-paul: Better will be `enableLogging` to be consistent with the fields we used at Yii 1:
https://github.com/yiisoft/yii/blob/master/framework/db/CDbConnection.php#L226
and at `yii2-mongodb`:
https://github.com/yiisoft/yii2-mongodb/blob/master/Connection.php#L138

It is also be good to have separated `enableProfiling`:
https://github.com/yiisoft/yii/blob/master/framework/db/CDbConnection.php#L232

klimov-paul (changes requested):

framework/db/Command.php
```diff
@@ -829,30 +829,32 @@ public function execute()
...
         if ($sql == '') {
             return 0;
         }
 
         $this->prepare(false);
 
-        $token = $rawSql;
         try {
-            Yii::beginProfile($token, __METHOD__);
+            $log && Yii::beginProfile($token, __METHOD__);
```
klimov-paul: I would better introduce separeted methods for log and profile composition like I did at `yii2-mongodb`:
https://github.com/yiisoft/yii2-mongodb/blob/master/Command.php#L179-L195

It will look better then `&&` operator.

framework/db/Command.php
```diff
@@ -868,9 +870,12 @@ public function execute()
...
     {
-        $rawSql = $this->getRawSql();
-
-        Yii::info($rawSql, 'yii\db\Command::query');
+        $log = $this->db->enableQueryLog;
+        if ($log) {
+            $rawSql = $this->getRawSql();
+            Yii::info($rawSql, 'yii\db\Command::query');
+            $token = $rawSql;
+        }
```
klimov-paul: This block duplicates the one from `execute()`, which is one more argument for https://github.com/yiisoft/yii2/pull/14056/files#r113421830

framework/db/Command.php
cebe: it might look better but it would complicate the code even more, e.g. introduce additional method calls that are not needed.

cebe: note that the main point of disabling is to avoid calling getRawSql() which may be heavy. separate methods would not do much except adding an if statement.

SilverFire (approved):
:+1:

framework/db/Command.php
rob006: `&&` -> `and`

http://stackoverflow.com/a/5998351/5812455

samdark: There's no assignment so it makes no difference.

rob006: If we consider this as convention, it does not matter if code works diferentlly or not. Same as 

```php
if (true) {
    $this->doSomething();
}
```
works the same as 
```php
if (true)
    $this->doSomething();
```

but only first version is preferred.

samdark: I'm fine with either way. @cebe what do you think?

framework/db/Connection.php
cebe: done.

framework/db/Command.php
cebe: done.

framework/db/Command.php
cebe: fine with me, changed it to using `and`.

cebe:
Adjusted code according to comments, ready for a second review.

framework/db/Command.php
```diff
@@ -168,7 +168,7 @@ public function getRawSql()
         }
         $params = [];
         foreach ($this->params as $name => $value) {
-            if (is_string($name) && strncmp(':', $name, 1)) {
+            if (is_string($name) && strpos($name, ':') !== 0) {
```
samdark: What's wrong with strncmp?

samdark (approved):

framework/db/Command.php
cebe: it was suggested by IDE inspections. after thinking about it I would revert it, because strpos is only faster for the positive case but not for the negative case, while strncmp is always the same.

cebe:
rebased on master to make travis happy.

dynasource:
thank you very much @cebe. Xdebug profiling  is showing a reduction of 10%!

PR #14061, opened 2017-04-26
Added assignments cache to check access on request

leandrogehlen:
| Q             | A
| ------------- | ---
| Is bugfix?    |  no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

On request the method checkAccess load assingments from database every invocation.
Example: 

```php
$access = [
  'person',
  'category',
  'product'
];

foreach ($access as $permission)  {
  if (Yii:$app->user->can($permission)) { // this will be executed the same SQL 3 times
   ...
  }
}



```


framework/rbac/DbManager.php
```diff
@@ -120,7 +126,12 @@ public function init()
      */
     public function checkAccess($userId, $permissionName, $params = [])
     {
-        $assignments = $this->getAssignments($userId);
+        $assignments = ArrayHelper::getValue($this->assignments, $userId);
```
SilverFire: I think it is better to do caching in `getAssignments()` method

SilverFire (changes requested):

framework/rbac/DbManager.php
leandrogehlen: i think that is not possible, because it generate  conflict with `defaultRules`

MysteryDragon:
Was this idea already rejected, or am I mistaken? #12902

leandrogehlen:
I think so, but in i disagree with @samdark, because his justification is that the assignments could be removed during the request. 

The same could be happens with anoters access resources, that currently has cache

SilverFire:
This PR implements per-request caching, that should not lead to problems, mentioned by @samdark

leandrogehlen:
Sorry @samdark, @SilverFire  i hadn't understood it

MysteryDragon:
https://github.com/yiisoft/yii2/issues/12902#issuecomment-258111421
> Кешировать (даже в пределах запроса) назначенные юзеру роли не безопасно в общем случае.
I doubt that... All right, it's up to you anyway)

SilverFire:
> Кешировать (даже в пределах запроса) назначенные юзеру роли не безопасно в общем случае.

Translation:

> Roles caching (even per-request) is not save in general

Calling for @yiisoft/core-developers thoughts

klimov-paul:
If such enhancement to be applied, it should be made optional. Caching assignments per user will lead to memory overflow in case of checking permissions over several different users during same request (for example inside console command).

leandrogehlen:
@klimov-paul, Could be used cache system per-request, to avoid memory overflow?

klimov-paul:
> Could be used cache system per-request, to avoid memory overflow?

I do not understand this: for console command there is only one request.
If you chck access for several users during same program flow - it will cause memory leak.

For example: I may want to check if user have permission to access 'gifts' section before sending an email with 'new gifts are availbale' content during mass mail cron.

leandrogehlen:
@klimov-paul, sorry my stupid question

Alex-Code:
In my `DbManager` extension I cached them like below.

```php
class DbManager extends \yii\rbac\DbManager
{
    /**
     * Store the result of getAssignments for the current request to negate multiple queries.
     * @var array
     */
    private $_assignments = [];

    /**
     * @inheritdoc
     */
    public function getAssignments($userId)
    {
        if (isset($this->_assignments[$userId])) {
            return $this->_assignments[$userId];
        }
        $this->_assignments = [];
        
        return $this->_assignments[$userId] = parent::getAssignments($userId);
    }
}
```

This way it'll query the db again if the user id changes.

leandrogehlen:
@Alex-Code, It seems that do the same think, but i don't agree to use cache for this method.
I think better to use only to check access

cebe:
merged with adjustments in  #14696, thanks!

Issue #14062, opened 2017-04-26
Relation models number limits


dryu:
Ok, I know those limits were discussed in depth in #10305 and #10371, and I agree with comments that having too much of IDs inside ```IN()``` statement is a bad idea, however, I don't agree with the conclusion to not do anything. 

The concern raised was that building up lots of models takes lots of memory and there are risks of having memory overflow exceptions. But if we don't do that, Yii2 fails miserably when trying to pull models with a total number more than the underlying database allows inside ```IN()``` statement. So it either always fails because of the database limitation not accounted properly in Yii2, or _sometimes_ fails if memory thresholds for PHP are reached. Why was it decided for the former instead of a later?

As of now I just don't see any legal workarounds against that issue. I have a huge relational hierarchy defined in my application. The logic pulls 1 root model (e.g. a blog post), and then it tries to pull lots of other related things, such as comments, authors, etc, via relations. On a very rare occasion, the number of comments goes beyond 1000 at which point the logic fails to pull models related to those comments and linked via a comment primary key, and in the end, the blog post cannot be displayed. I cannot batch the comments, because those are being pulled indirectly (through relations), rather than directly, thus they are always resolved with ```all()```. 

```php
$post = BlogPost::find()->with('comments.time')->one();
```

I don't want to manage comments separately from the blog post, e.g. I don't want that:

```
$comments = $post->getComments()->all();
```

Because in the above case comments are not present from within the parent model $post, and have to be resolved again if they are accessed again.

I want to handle comments and all models related to comments like that:

```
$comments = $post->comments;
```

And I want _all_ comments on the same page, no pagination. This is a design decision, and we can discuss in length whether it is correct or not, but right now, I just don't see how to do that in Yii2 without hacks.

I personally believe it is better to hit potential memory thresholds when trying to construct thousands of models, rather than allow a database exception from unaccounted database limitations.

Oh, and speaking of hacks, I had to change ```ActiveRelationTrait``` class to break IN() statements in batches of  1000 items to workaround the issue (in case anyone still wants to have that functionality):

```php
    private function filterByModels($models)
    {
        $attributes = array_keys($this->link);

        $attributes = $this->prefixKeyColumns($attributes);

        $values = [];
        if (count($attributes) === 1) {
            // single key
            $attribute = reset($this->link);
            foreach ($models as $model) {
                if (($value = $model[$attribute]) !== null) {
                    if (is_array($value)) {
                        $values = array_merge($values, $value);
                    } else {
[...]

Issue #10371, opened 2015-12-11
MSSQL Server supports a maximum of 2100 parameters


thejahweh:
In some cases there is a problem how Yii2 handles the `[id] IN (1,2,3..)` statement.

``` sql
SQLSTATE[IMSSP]: Tried to bind parameter number 2101. SQL Server supports a maximum of 2100 
parameters. The SQL being executed was:
SELECT COUNT(*) 
FROM [tablexy] 
WHERE [id] IN ('107518', '96017', '79543', '96016', '82370', '96638', ...
```

SilverFire:
This problem related not the MSSQL only. See https://github.com/yiisoft/yii2/issues/10305

samdark:
Usually it's not a huge issue. If there are more than 100 ids in a query per page, there's probably an issue in software logic.

o-rey:
Not a bug, just use joinWith() instead.

cebe:
> Not a bug, just use joinWith() instead.

this should make no difference since `with()` is called by `joinWith()` too...

dryu:
If I have AR relation defined that in runtime resolves to more than 2100 records that needs to be looked up in the database, the application will fail.

dryu:
It seems that this error message text is misleading and the error is thrown by Microsoft SQL Server PHP driver and not by MS SQL Server itself . MSSQL by itself has no problems with the size of parameter arrays passed to it.

samdark:
Makes sense to use non-prepared statements in this case.

kevm256:
I have a work around that fixes this issue for active relations that use via(), may also be faster for relations with lots of records. At least in my case using MSSQL.

Basically this just uses the subquery in the where in condition instead of using filterByModels()

https://gist.github.com/kevm256/cb4557ec5f60d8723d45b4e845d60712

I haven't done extensive testing, so not sure if there are other unintended side effects.

samdark:
@kevm256 that's a good workaround but not a permanent solution. That would likely break relationships between multiple database types that work perfectly fine now.

kevm256:
@samdark yeah, I don't think it will work across databases. I did try adding a buildValues() override in the mssql InConditionBuilder class to put int values directly into the condition instead of binding them:
```php
protected function buildValues(ConditionInterface $condition, $values, &$params)
{
    if(count($values) > 2000){
        $intValues = [];
        foreach ($values as $i => $value)
        {
            if(is_int($value)){
                $intValues[] = strval($value);
[...]

Issue #10305, opened 2015-12-02
Oracle relations limit of 1000


jonny7:
When trying to access a large models relations like so 

```
Model::find()->with('relationName')->all();
```

I get the error, `ORA-01795: maximum number of expressions in a list is 1000 error`

I'm assuming that I haven't done anything wrong or missed something. Looking around, there appears to be a workaround for this: http://stackoverflow.com/a/17844383/895169

Is this something we should look to implement within Yii to help Oracle DBs or am I using Yii incorrectly?

SilverFire:
Should be fixed

samdark:
That's because of `WHERE id IN(1,2,3,4,...)`. MySQL has the same limit as well. Usually it's not a problem because getting 1000 releated records often means something's very wrong. @jonny7 what's your use case for 1000 records on the same request?

jonny7:
I have Table A with 10k procedure records and Table B is a pathology table with 7k rows. Not all procedures will have related pathology. I am using this mainly for generating statistics and I was just testing that the relation worked as I have found Oracle not as easy as MySQL to use, saw this and wondered about it being a bug or my mistake. Is my use case bad?

Out of interest then, If you have a large database, is active record not the way to go when accessing the data?

samdark:
Yes. It doesn't really matter if it's AR or not. Getting more than 100 records in a single request is generally not a good idea. It's not fast and memory greedy.

SilverFire:
@samdark @cebe do you think it should be fixed?
As for me - yes, it should.

> Getting more than 1000 records in a single request is generally not a good idea. It's not fast and memory greedy.

Yes, it's greedy, but sometimes a single SQL with 5000 expressions in the `IN` condition is much faster than 5 separated queries.

klimov-paul:
> Yes, it's greedy, but sometimes a single SQL with 5000 expressions in the IN condition is much faster than 5 separated queries.

This is no the point. The point is fetching 1000 records will consume too much memory for PHP script running. What the sense of fixing `maximum number of expressions in a list` error, just to see `PHP memory overflow` instead?

In general this is not a common use case. Processing of the large amount of data requires more care and attention.
We have implemented querying in batches for the large data-sets processing. As for me - this is enough. Creating a hack aroung `IN` SQL opeartion does not sounds good.

thejahweh:
[...]

PR #14065, opened 2017-04-27
Fixed Oracle SQL queries with `IN` condition and more than 1000 params

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #10305, #14062 

<s>@samdark could you check whether this fix is applicable to #10371, or MSSQL limits parameters count per query?</s> - checked. MSSQL problem is about bound parameters, not about number of params in `in`

### DIFF
--- a/framework/db/oci/QueryBuilder.php
+++ b/framework/db/oci/QueryBuilder.php
@@ -332,6 +332,9 @@ EOD;
         return parent::buildLikeCondition($operator, $operands, $params);
     }
 
+    /**
+     * @inheritdoc
+     */
     public function buildInCondition($operator, $operands, &$params)
     {
         $splitCondition = $this->splitInCondition($operator, $operands, $params);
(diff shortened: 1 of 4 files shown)


framework/db/oci/QueryBuilder.php
```diff
@@ -331,4 +331,56 @@ public function buildLikeCondition($operator, $operands, &$params)
         }
         return parent::buildLikeCondition($operator, $operands, $params);
     }
+
+    public function buildInCondition($operator, $operands, &$params)
```
cebe: add `@inheritdoc`.

cebe (approved):

samdark (approved):

SilverFire:
Merged f6a4f3c18f9f4b3f4e6af5bbad0e297f2aa55108

Issue #4793, opened 2014-08-22
Using yii\filters\AccessControl in yii\rest\Controller with no User::identityClass set


RomeroMsk:
I have an API application with REST controller and use AccessControl with IP rules.
If I don't want to use user auth, what is the right configuration of application?
When `user => identityClass` is not set in config, I'm getting this exception: `User::identityClass must be set`.

alez007:
You can't just create that dummy class without implementing IdentityInterface methods.

MKiselev:
@alez007 yep, I just make stub User model like this:

```php
<?php


namespace frontend\models;


use yii\base\Component;
use yii\web\IdentityInterface;

class NullUser extends Component implements IdentityInterface
{
    public static function findIdentity($id)
    {
        return null;
    }

    public static function findIdentityByAccessToken($token, $type = null)
    {
        return null;
    }

    public function getId()
    {
        return null;
    }

    public function getAuthKey()
    {
        return null;
    }

    public function validateAuthKey($authKey)
    {
        return false;
    }
}

```

MKiselev:
But better and clear way it is write your filter like this:

```php
<?php

namespace common\components\filters;


use Yii;
use yii\base\ActionFilter;
use yii\web\BadRequestHttpException;

class AjaxFilter extends ActionFilter
{
    public function beforeAction($action)
    {
        if (Yii::$app->request->getIsAjax()) {
            return true;
        } else {
            throw new BadRequestHttpException('Ajax requests only allowed!');
        }
    }
}

```

and add to behaviors to Application or Controller
```php
public function behaviors()
{
    return [
        'ajax' => AjaxFilter::class,
    ];
}
```

Faryshta:
@RomeroMsk can you edit the title to something like

"IP validation in yii\filters\AccessControl without required identified user2

PR #14066, opened 2017-04-27
Allow AccessControl filter to work without User component

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | a bit
| Tests pass?   | yes
| Fixed issues  | #4793

This allows `AccessControl` filter to work without `user` component since only matching role mechanism depends on it.

- `yii\filters\AccessControl::user` can be set to `false` to explicitly switch this component support off for the filter,
- `yii\filters\AccessControl::denyAccess()` throws `yii\web\ForbiddenHttpException` if the user is already logged in **or in case of detached User component**,
- `yii\filters\AccessRule::matchRole()` throws `yii\base\InvalidConfigException` if User component is detached and `yii\filters\AccessRule::roles` are not empty.

With default configuration behavior is the same as before.

Also `yiiunit\framework\filters\AccessRuleTest` methods don't require mocking User component anymore.


tests/framework/filters/AccessRuleTest.php
```diff
@@ -180,10 +180,28 @@ public function testMatchRole($actionid, $allow, $userid, $expected) {
...
+    public function testMatchRoleWithoutUser() {
+        $action = $this->mockAction();
+        $request = $this->mockRequest();
+
+        $rule = new AccessRule([
+            'allow' => true,
+            'roles' => ['@'],
+        ]);
+
+        $this->setExpectedException('yii\base\InvalidConfigException');
```
SilverFire: Should be replaced with `expectException()`

https://github.com/yiisoft/yii2/blob/master/tests/compatibility.php#L18

SilverFire (changes requested):
Generally good

SilverFire (approved):

samdark (approved):

framework/filters/AccessRule.php
```diff
@@ -141,12 +142,16 @@ protected function matchController($controller)
...
      * @return bool whether the rule applies to the role
+     * @throws InvalidConfigException if User component is detached
      */
     protected function matchRole($user)
     {
         if (empty($this->roles)) {
             return true;
         }
+        if ($user === false) {
+            throw new InvalidConfigException('The required component is not specified.');
```
cebe: this error message could be improved:

> "The user application component must be available to specify roles in AccessRule."

cebe (changes requested):

cebe (approved):

SilverFire (approved):

samdark:
Merged. Thank you!

Issue #14072, opened 2017-04-27
Table schema not getting refreshed for some Command methods


brandonkelly:
(Follow-up issue for https://github.com/yiisoft/yii2/issues/3305#issuecomment-296772327)

`yii\db\Schema::refreshTableSchema()` is called after most schema-altering methods in `yii\db\Command` are called, but not the following ones:

- `createTable()`
- `addForeignKey()`
- `dropForeignKey()`
- `addCommentOnColumn()`
- `dropCommentFromColumn()`

Each of these do have an effect on what gets memoized by `getTableSchema()`, so they should be clearing the schema cache.

### What steps will reproduce the problem?

1. call `yii\db\Schema::getTableSchema()` for a table that doesn’t exist
2. create the the table
3. call `yii\db\Schema::getTableSchema()` again for the same table name 

### What is the expected result?

a `yii\db\TableSchema` object is returned

### What do you get instead?

`null` is returned, because it was cached as `null` in step 1, and that cache was not cleared in step 2.

### Additional info

Similar steps to reproduce for the other functions listed above.

PR is forthcoming…

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11, 2.0.12 (dev) @ https://github.com/yiisoft/yii2/commit/bda091bbb24de594adaae0554421cb81598b00f1

khalwat:
Awesome. I just pushed a change to `generator-craftplugin` that does `Craft::$app->db->schema->refresh()` in the `Install.php` migration just before calling the `insertDefaultData()` method.

PR #14073, opened 2017-04-27
Schema cache fixes

brandonkelly:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14072


SilverFire (approved):

tests/framework/db/CommandTest.php
```diff
@@ -809,17 +809,34 @@ public function testAutoRefreshTableSchema()
...
             'name' => 'string',
         ])->execute();
         $initialSchema = $db->getSchema()->getTableSchema($tableName);
+        $this->assertNotNull($initialSchema);
 
         $db->createCommand()->addColumn($tableName, 'value', 'integer')->execute();
         $newSchema = $db->getSchema()->getTableSchema($tableName);
         $this->assertNotEquals($initialSchema, $newSchema);
 
+        $db->createCommand()->addForeignKey($fkName, $tableName, 'fk', $tableName, 'id')->execute();
```
MysteryDragon: https://travis-ci.org/yiisoft/yii2/jobs/226569561#L325
I think some tests is better to mark skipped for some particular DBMS (like https://github.com/yiisoft/yii2/blob/master/tests/framework/db/CommandTest.php#L561)

brandonkelly: Done: https://github.com/yiisoft/yii2/pull/14073/commits/237af91e8ef3ebbdeedf1881d2820bbb320fea8b

MysteryDragon: Sorry if I mislead you by pointing to `markTestSkipped()` method without futher explanations.
With this method ALL test method will be skipped, and it's not so correct now 'cause `addColumn()`, for instance, works perfect with `sqlite`, but also will not be tested.
`markTestSkipped()` is applicable if we'll split this method to several ones.

brandonkelly: @MysteryDragon What do you recommend? Should I split `testAutoRefreshTableSchema` into multiple tests (which would require re-building the table multiple times), or just skip the FK command + assertion if running SQLite?

MysteryDragon: @brandonkelly *IMO* the first option is more "theoretically correct", but I would prefer the second one due to simplicity and efficiency. I can imagine some other options, but they also have their disadvantages.

brandonkelly: @MysteryDragon Alright, done: 11c0d4bf159572a9f33fdb5936a1b9fa939ece74 (skipped the column comment things too, since sqlite doesn’t support that either).

tests/framework/db/CommandTest.php
```diff
@@ -809,17 +809,36 @@ public function testAutoRefreshTableSchema()
...
             'name' => 'string',
         ])->execute();
         $initialSchema = $db->getSchema()->getTableSchema($tableName);
+        $this->assertNotNull($initialSchema);
 
         $db->createCommand()->addColumn($tableName, 'value', 'integer')->execute();
         $newSchema = $db->getSchema()->getTableSchema($tableName);
         $this->assertNotEquals($initialSchema, $newSchema);
 
+        if ($this->driverName === 'sqlite') {
```
MysteryDragon: @brandonkelly I think you mean `!==` here.

brandonkelly: Doh! Sorry about that.

cebe (approved):

framework/db/Command.php
```diff
@@ -537,7 +537,7 @@ public function createTable($table, $columns, $options = null)
     {
         $sql = $this->db->getQueryBuilder()->createTable($table, $columns, $options);
 
-        return $this->setSql($sql);
+        return $this->setSql($sql)->requireTableSchemaRefresh($table);
```
samdark: If the table is just created, there should be no past cache so it makes no sense to refresh it.

samdark (changes requested):

framework/db/Command.php
cebe: Schema caches also negative results, i.e. a non existing table will cache a `null` value:

https://github.com/yiisoft/yii2/blob/533da6240e067ad260ddca10757d99a6b4b68ecf/framework/db/Schema.php#L152

brandonkelly: Right, as demonstrated in #14072

brandonkelly: @samdark Also this PR comes with a couple new test assertions that will demonstrate the issue. If this change is removed, this will fail:

https://github.com/yiisoft/yii2/pull/14073/files#diff-e48002914d2773120698a96387ecabd2R822

samdark: 👍

samdark (approved):

samdark:
Merged. Thank you!

Issue #13779, opened 2017-03-15
joinWith() fails with relations provided by behaviors attached via DI configuration


vercotux:
### What steps will reproduce the problem?

1. Create `\yii\db\ActiveRecord`, eg. `Post`, and override `instantiate()`:
```php
public static function instantiate($row) {
	return Yii::$container->get(static::class);
}
```
2. Create `\yii\base\Behavior`, eg. `PostBehavior`, which contains a new *user* relation for `Post`:
```php
public function getUser() {
	return $this->owner->hasOne(User::class, ['id' => 'user_id']);
}
```
3. Attach `PostBehavior` to `Post` using DI container definitions, eg.:
```php
'container' => [
	'definitions' => [
		'common\models\Post' => [
			'as myBehavior' => 'common\behaviors\PostBehavior',
		],
	],
],
```

4. Execute `Post::find()->joinWith('user')->all();`

### What is the expected result?

Everything works fine if `PostBehavior` is added directly inside `Post::behaviors()`, but fails when included through application's DI container definitions.

### What do you get instead?

>Invalid Parameter – yii\base\InvalidParamException
>common\models\Post has no relation named "user".

### Additional info

The problem appears to lie in [`\yii\db\ActiveQuery::buildJoinWith()`](https://github.com/yiisoft/yii2/blob/13de76714bd5482d0c2b9a68b7cb5d4977fc7bd0/framework/db/ActiveQuery.php#L439), where our ActiveRecord is being instantiated using `new $this->modelClass` rather than `Yii::createObject($this->modelClass)` thus causing the relations defined via DI-provided behaviors to be disregarded.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11
| PHP version      | 5.6
| Operating system | Ubuntu

samdark:
Looks valid at the first sight. Would you like to fix it with a pull request containing the fix and a unit test? If yes then it has a chance to be merged into 2.0.12.

lynicidn:
need change `AR::instantiate` too

vercotux:
@samdark I'm working on the PR + unit tests as you requested.
@lynicidn Generally you should, yes, but it isn't relevant to this specific case so I omitted it.

lynicidn:
try
$post = new Post();
$user = User::findOne();
$post->link('user', $user);
you also get exception that user relation not found

vercotux:
Read my post again.

lynicidn:
[...]

PR #14078, opened 2017-04-28
Allow to use custom AR constructors

ElisDN:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | potentially
| Tests pass?   | yes
| Fixed issues  | #5786, #8639, #13779, #5771

The PR replaces all `new $modelClass` internal calls to `$modelClass::instantiate()`. It allows to use custom constructors in ActiveRecord-based classes:

~~~php
class User extends ActiveRecord
{
    public static function instantiate($row = [])
    {
        $record = (new \ReflectionClass(static::className()))->newInstanceWithoutConstructor();
        $record->init();
        return $record;
    }
    
    public function __construct($name, $email)
    {
        $this->name = $name;
        $this->email = $email;
        $this->created_at = time();
        parent::__construct();
    }
}
~~~

without internal issues with all types of relations.

Backwards compatibility
------

The `$row` argument can be an empty array in new cases. It may break existing code. For example, for [STI](https://github.com/samdark/yii2-cookbook/blob/master/book/ar-single-table-inheritance.md) you must add `empty($row)` checking before `switch` sentence:

```php
public static function instantiate($row = [])
{
    if (empty($row)) {
        return new self();
    }

    switch ($row['type']) {
        ...
    }
}
```
or use safe `ArrayHelper::getValue($row, 'type')` instead of unsafe `$row['data']` calling.

### DIFF
--- a/framework/db/ActiveRecordInterface.php
+++ b/framework/db/ActiveRecordInterface.php
@@ -427,4 +427,19 @@ interface ActiveRecordInterface
      * @return mixed the database connection used by this AR class.
      */
     public static function getDb();
+
+    /**
+     * Creates an active record instance.
+     *
+     * This method is called internally by framework for instantiating new record object.
+     * It is not meant to be used for creating new records directly.
+     *
+     * You may override this method if the instance being created
+     * depends on the row data to be populated into the record.
+     * For example, by creating a record based on the value of a column,
+     * you may implement the so-called single-table inheritance mapping.
+     * @param array $row row data to be populated into the record (can be empty).
+     * @return ActiveRecordInterface the newly created active record
+     */
+    public static function instantiate($row = []);
 }
--- a/framework/grid/DataColumn.php
+++ b/framework/grid/DataColumn.php
@@ -146,11 +146,21 @@ class DataColumn extends Column
         if ($this->label === null) {
             if ($provider instanceof ActiveDataProvider && $provider->query instanceof ActiveQueryInterface) {
                 /* @var $model Model */
-                $model = new $provider->query->modelClass();
+                if (is_subclass_of($provider->query->modelClass, 'yii\db\ActiveRecordInterface')) {
+                    $modelClass = $provider->query->modelClass;
+                    $model = $modelClass::instantiate();
+                } else {
+                    $model = new $provider->query->modelClass();
+                }
                 $label = $model->getAttributeLabel($this->attribute);
             } elseif ($provider instanceof ArrayDataProvider && $provider->modelClass !== null) {
                 /* @var $model Model */
-                $model = new $provider->modelClass();
+                if (is_subclass_of($provider->modelClass, 'yii\db\ActiveRecordInterface')) {
+                    $modelClass = $provider->modelClass;
+                    $model = $modelClass::instantiate();
+                } else {
+                    $model = new $provider->modelClass();
+                }
                 $label = $model->getAttributeLabel($this->attribute);
             } elseif ($this->grid->filterModel !== null && $this->grid->filterModel instanceof Model) {
                 $label = $this->grid->filterModel->getAttributeLabel($this->attribute);
--- a/tests/data/ar/Animal.php
+++ b/tests/data/ar/Animal.php
@@ -36,10 +36,11 @@ class Animal extends ActiveRecord
     }
 
     /**
-     * @param type $row
+     *
+     * @param array $row
      * @return \yiiunit\data\ar\Animal
      */
-    public static function instantiate($row)
+    public static function instantiate($row = [])
     {
         $class = $row['type'];
         return new $class();
--- a/tests/data/ar/CustomerWithConstructor.php
+++ b/tests/data/ar/CustomerWithConstructor.php
@@ -1,0 +1,51 @@
+<?php
+namespace yiiunit\data\ar;
+
+use yiiunit\framework\db\ActiveRecordTest;
+
+/**
+ * Class Customer
+ *
+ * @property int $id
+ * @property string $name
+ * @property string $email
+ * @property string $address
+ * @property int $status
+ *
+ * @property ProfileWithConstructor $profile
+ */
+class CustomerWithConstructor extends ActiveRecord
+{
+    const STATUS_ACTIVE = 1;
+    const STATUS_INACTIVE = 2;
+
+    public static function tableName()
+    {
+        return 'customer';
+    }
+
+    public function __construct($name, $email, $address)
+    {
+        $this->name = $name;
+        $this->email = $email;
+        $this->address = $address;
+        parent::__construct();
+    }
+
+    public static function instantiate($row = [])
+    {
+        return (new \ReflectionClass(static::className()))->newInstanceWithoutConstructor();
+    }
+
+    public function getProfile()
+    {
+        return $this->hasOne(ProfileWithConstructor::className(), ['id' => 'profile_id']);
+    }
+
+    public function afterSave($insert, $changedAttributes)
+    {
+        ActiveRecordTest::$afterSaveInsert = $insert;
+        ActiveRecordTest::$afterSaveNewRecord = $this->isNewRecord;
+        parent::afterSave($insert, $changedAttributes);
+    }
+}
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -13,6 +13,7 @@ Yii Framework 2 Change Log
 - Enh #4495:  Added closure support in `yii\i18n\Formatter` (developeruz)
 - Enh #13787: Added `yii\db\Migration::$maxSqlOutputLength` that allows limiting number of characters for outputting SQL (thiagotalma)
 - Enh #13824: Support extracting concatenated strings in `yii message` (developeruz)
+- Enh #14078: Allowed to use custom constructors in ActiveRecord-based classes (ElisDN)
 - Enh #14089: Added tests for `yii\base\Theme` (vladis84)
 - Enh #13586: Added `$preserveNonEmptyValues` property to the `yii\behaviors\AttributeBehavior` (Kolyunya)
 - Bug #14192: Fixed wrong default null value for TIMESTAMP when using PostgreSQL (Tigrov)
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -76,6 +76,9 @@ Upgrade from Yii 2.0.11
 * Inputmask package name was changed from `jquery.inputmask` to `inputmask`. If you've configured path to
   assets manually, please adjust it. 
 
+* The usage of `yii\db\BaseActiveRecord::instantiate()` was changed to more general `instantiate($row = [])`. The method is called
+  everywhere in framework internal code instead of `new $modelClass()` and can receive empty `$row` array in this cases.
+
 Upgrade from Yii 2.0.10
 -----------------------
 
--- a/framework/data/ActiveDataProvider.php
+++ b/framework/data/ActiveDataProvider.php
@@ -177,7 +177,12 @@ class ActiveDataProvider extends BaseDataProvider
         parent::setSort($value);
         if (($sort = $this->getSort()) !== false && $this->query instanceof ActiveQueryInterface) {
             /* @var $model Model */
-            $model = new $this->query->modelClass();
+            if (is_subclass_of($this->query->modelClass, 'yii\db\ActiveRecordInterface')) {
+                $modelClass = $this->query->modelClass;
+                $model = $modelClass::instantiate();
+            } else {
+                $model = new $this->query->modelClass();
+            }
             if (empty($sort->attributes)) {
                 foreach ($model->attributes() as $attribute) {
                     $sort->attributes[$attribute] = [
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -435,7 +435,9 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         $join = $this->join;
         $this->join = [];
 
-        $model = new $this->modelClass();
+        $modelClass = $this->modelClass;
+        $model = $modelClass::instantiate();
+
         foreach ($this->joinWith as $config) {
             list($with, $eagerLoading, $joinType) = $config;
             $this->joinWithRelations($model, $with, $joinType);
@@ -515,7 +517,8 @@ class ActiveQuery extends Query implements ActiveQueryInterface
                 } else {
                     $relation = $relations[$fullName];
                 }
-                $primaryModel = new $relation->modelClass();
+                $modelClass = $relation->modelClass;
+                $primaryModel = $modelClass::instantiate();
                 $parent = $relation;
                 $prefix = $fullName;
                 $name = $childName;
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -177,7 +177,8 @@ trait ActiveRelationTrait
                 $relatedModel->populateRelation($this->inverseOf, $inverseRelation->multiple ? [$this->primaryModel] : $this->primaryModel);
             } else {
                 if (!isset($inverseRelation)) {
-                    $inverseRelation = (new $this->modelClass())->getRelation($this->inverseOf);
+                    $modelClass = $this->modelClass;
+                    $inverseRelation = $modelClass::instantiate()->getRelation($this->inverseOf);
                 }
                 $result[$i][$this->inverseOf] = $inverseRelation->multiple ? [$this->primaryModel] : $this->primaryModel;
             }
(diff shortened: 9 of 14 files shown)


framework/db/InstantiateInterface.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\db;
+
+/**
+ * InstantiateInterface allows
```
SilverFire: ...

SilverFire:
The idea is good in general, but we should work on implementation a bit.

I'm not feeling OK with similar `if` condition all around the code:

```php
if (is_subclass_of($this->query->modelClass, 'yii\db\InstantiateInterface')) {
    $modelClass = $this->query->modelClass;
    $model = $modelClass::instantiate([]);
} else {
    $model = new $this->query->modelClass;
}
```

I'd prefer adding `instantiate()` method to the `ActiveRecordInterface` with clean implementation in `BaseActiveRecord`. @ElisDN, do you any disadvantages of this approach?

ElisDN:
@SilverFire Yes. I added `instantiate` in `ActiveRecordInterface`.

Faryshta:
i still think its overly complicated and counter intuitive to force parameters on the constructor, then force `instantiate()` to skip the constructor.

a method `static User::create($name, $email)` after all since `instantiate()` is public  there is nothing preventing anyone from using

```php
$user = User::instantiate([]);
// instead of
$user = new User();
```

or even just making the parameters optional would do the trick `

```php
public function __construct($name = null, $email = null)
{
     if (is_array($name)) { // logic used for `instantiate`
     } else { // regular constructor
     }
}
```php

**Comment by ElisDN** [comment]
> after all since instantiate() is public there is nothing preventing anyone from using...

@Faryshta, The method always was public and was used by framework in `ActiveQueryTrait::createModels()`.

**Comment by Faryshta** [comment]
i know, thats not my point.

**Comment by ElisDN** [comment]
> its overly complicated and counter intuitive to force parameters on the constructor,

@Faryshta, Using of constructors for development and skipping them during retrieving records from DB are normal developing practices.

**Comment by Faryshta** [comment]
i don't agree they are normal tactics or see why they are GOOD PRACTICES, since they are not intuitive.

For the example you provided i think its way easier to do

```php
public static function create($name, $email)
{
      $user = new static();
      $user->name = $name;
      $user->email = $email;
      $user->created_at = time();
      return $user;
}
```

specially since your solution allows you to omit the constructor by doing `$user = User::instantiate([])`.

which would become the defacto constructor since its easier to use than `new User()`

ElisDN:
@Faryshta, See the PR changesets please. The `instantiate` method still contains your `return new static` and I did not modify its code. I only have owerwritten `new $className` to `$className::instantiate()` in all places in framework. It allows me to override this method if I want to use Reflection and constructors. It is all.

tests/data/ar/CustomerWithConstructor.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->name = $name;
+        $this->email = $email;
+        $this->address = $address;
+        parent::__construct();
+    }
+
+    public static function instantiate($row = [])
+    {
+        return (new \ReflectionClass(static::className()))->newInstanceWithoutConstructor();
```
Faryshta: why not use  `return new static('', '');` or use default values?

tests/data/ar/Animal.php
```diff
@@ -38,10 +38,10 @@ public function getDoes()
 
     /**
      *
-     * @param type $row
+     * @param array $row
      * @return \yiiunit\data\ar\Animal
      */
-    public static function instantiate($row)
+    public static function instantiate($row = [])
```
Faryshta: bc break since it changes the method signature for everyone already extending this method

framework/grid/DataColumn.php
```diff
@@ -144,11 +144,21 @@ protected function getHeaderCellLabel()
         if ($this->label === null) {
             if ($provider instanceof ActiveDataProvider && $provider->query instanceof ActiveQueryInterface) {
                 /* @var $model Model */
-                $model = new $provider->query->modelClass;
+                if (is_subclass_of($provider->query->modelClass, 'yii\db\ActiveRecordInterface')) {
```
Faryshta: @SilverFire how about moving `instantiate()` to the `yii\base\Model` class?

tests/data/ar/Animal.php
Kolyunya: @Faryshta could you please provide an example of code which will break after this change?

tests/data/ar/CustomerWithConstructor.php
ElisDN: It is because normal constructor have business logic and restrictions:

```php
public function __construct($name, $email)
{
    if (empty($name) || empty($email)) {
        throw new \InvalidArgumentException(...);
    }
    $this->name = $name;
    $this->email = $email;
    $this->created_at = time();
    $this->recordEvent(new CustomerCreated($this));
    parent::__construct();
}
```

tests/data/ar/Animal.php
ElisDN: Yes. I noted it in PR.

tests/data/ar/CustomerWithConstructor.php
Faryshta: i think thats up for the developer to know how to handle the constructor and the instantiate.

lets think it the other way around. what if the developer assumes some property, behavior or something got initialized by the constructor and start using it after a `findOne()` for example.

framework/grid/DataColumn.php
Faryshta: the property `ActiveQuery::$modelClass` gets defined here http://www.yiiframework.com/doc-2.0/yii-db-activequerytrait.html#$modelClass-detail and the docs mention is an active record class.

@SilverFire so can we assume `$modelClass` is a subclass of `yii\db\ActiveRecordInterface`?

tests/data/ar/CustomerWithConstructor.php
ElisDN: > i think thats up for the developer to know how to handle the constructor and the instantiate.

If developer really wants to use own customized code and knows how to override `instantiate` method for working with his code that the developer can do it. If developer does not need it or does not know about this method existence he do not do it.

> what if the developer assumes some property, behavior or something got initialized by the constructor and start using it after a findOne() for example.

By default `findOne()` calls `instantiate($row)`. Method `instantiate` still calls `return new $class`. Logically that the developer can put this initialization code into `instantiate` method too if he has overwritten the method by his way.

tests/data/ar/Animal.php
ElisDN: As an alternative without bc breaking we can leave `instantiate($row)` method as is and make another method like `instantiateInternal()` without any parameters.

ElisDN:
As an alternative way without bc breaking we can leave `instantiate($row)` method as is and make another method like `InstantiateInterface::instantiateInternal()` without any parameters and use it:

```php
if (is_subclass_of($this->modelClass, 'yii\base\InstantiateInterface')) {
    $modelClass = $this->modelClass;
    $model = $modelClass::instantiateInternal();
} else {
    $model = new $this->modelClass;
}
```

It will allow us to use `Model implements InstantiateInterface` type inheritance.

tests/data/ar/Animal.php
ElisDN: https://github.com/yiisoft/yii2/pull/14078#issuecomment-298153836

framework/grid/DataColumn.php
ElisDN: https://github.com/yiisoft/yii2/pull/14078#issuecomment-298153836

framework/db/ActiveRecordInterface.php
```diff
@@ -427,4 +427,19 @@ public function unlink($name, $model, $delete = false);
...
+     * It is not meant to be used for creating new records directly.
+     *
+     * You may override this method if the instance being created
+     * depends on the row data to be populated into the record.
+     * For example, by creating a record based on the value of a column,
+     * you may implement the so-called single-table inheritance mapping.
+     * @param array $row row data to be populated into the record (can be empty).
+     * @return ActiveRecordInterface the newly created active record
+     */
+    public static function instantiate($row = []);
```
samdark: Need to have this one as is and, additionally, add another method w/o params that is creating an instance but isn't filling it. It would be used for cases when we're getting metadata such as attr. labels or joinWith relation metadata.

In 2.1 we may call the one w/o arguments `instantiate()` and the one w/ `$row` hydrate.

samdark: Need suggestions for method names.

samdark: @yiisoft/core-developers what do you think about naming?

prowwid:
Hi @yiisoft/core-developers,

What is the status of this bugfix?
This bug affects application our team develops and we are interested when it would be fixed?

Pull request being open for more than two months.  If you don't have time to finalize the code I can help - just say what to do

ElisDN:
@prowwid Reply to @samdark review of https://github.com/yiisoft/yii2/pull/14078#issuecomment-298153836 about method naming please.

samdark:
Yeah. Naming is really hard for this one. The rest is more or less decided.

framework/db/ActiveRecordInterface.php
prowwid: Hello @samdark ,

I think having `instantiate` method as is and add a new method that should simply create an empty model for metadata is the best approach right now. 
Right now people use "instantiate" methods to instantiate models and don't expect changes in the method signature or it's meaning.

But these two methods have different meanings and, from my point of view, should not share **"instantiate"** word in names. 
_So I suggest having a method that creates empty models with a name such as:_
- createEmpty
- createNew
- createPlainModel
- create
- newEmptyModel
- newPlainModel

People probably would want to use this method passing instantiation stuff so I don't think **"internal"** or something similar should be used in this method. 

Also, if skip options with **"Model"**, I think, it makes sense to move method to some of the base classes.
For example, `createNew` might be implemented in Object.
`Object::createNew` is pretty natural from my point of view.

So what do you think?

samdark: `create()` actually sounds good.

samdark: @ElisDN what do you think?

rob006: I would prefer `Model::new()` - it is closer to native PHP syntax and IMO better fits for current naming convention.

klimov-paul: It is unlikely posssible since `new` is PHP reserved keyword.

rob006: It works in PHP7, so we could use it if we target this to Yii 2.1.

https://3v4l.org/v8vAv

klimov-paul:
I am against usage of `instantiate()` or any other static method, which should return a model instance for reference purposes.

The reason why new model is instantiated over the framework code is fetching its meta data such as attribute names, labels relations and so on.
Originally related methods were static until issue #1318.
The actual solution will be moving `attributes()`, `attributeLabels()` and `attributeHints()` to `static` scope, while introducing object-level mirrors for them in case someone wishes to provide object-level variations.
So actual AR code will look like:

```php
class Foo extends ActiveRecord
{
    public static function attributeLabels()
    {
         return [
             'id' => 'ID',
             'name' => 'Name',
         ];
    }

    public function instanceAttributeLabels()
    {
         // can be overriden if there is necessity to object-level variation.
         return static::attributeLabels();
    }
}
```

The current problem is ActiveRecord is unable to provide static information about itself, while some code requires such information. Creating a 'dummy' model to retrieve it is a solution for the problem.
I suppose we need to fix the source problem instead of patching solution for this problem.

prowwid:
Hello @klimov-paul ,

I completely agree with you that better leave 'instantiate()' alone and think of a way to get metadata without a need to create AR instance, but moving only  `attributes()`, ` attributeLabels()` and `attributeHints()` to `static` won't solve all of the issues.

You see, we can easily add behavior to model through `Container` configuration to add some relations or other stuff dynamically.  Or you can even switch specific model to some other realization by the same `Container` and in both situations, core won't get the correct code with the actual approach.

Examples of places that creates AR using `new`.
`yii\db\BaseActiveRecord::link` line 1266
`yii\db\ActiveQueryTrait::findWith` line 165
`yii\db\ActiveQuery::buildJoinWith` line 437
`yii\db\ActiveQuery::joinWithRelations` line 517
`yii\data\ActiveDataProvider::setSort` line 117

I'm not suggesting to create everything through container as I know you won't. 
Simply having a single `public/protected` method that creates AR object would be enough for the end user to extend behavior as they want.

And as for metadata, I believe it's better to separate metadata to an independent object - like ActiveQuery for queries stuff. This way `ActiveQuery` and `ActiveDataProvider` would be able to get metadata without a need to touch model(that probably would have a lot of stuff that not needed to build queries). 

Looking forward to your comments.

With best regards,
Dmitry

klimov-paul:
> You see, we can easily add behavior to model through Container configuration to add some relations or other stuff dynamically. Or you can even switch specific model to some other realization by the same Container and in both situations, core won't get the correct code with the actual approach.

I do not suggest object-level attributes, labels and hints it to be removed. At [my comment](https://github.com/yiisoft/yii2/pull/14078#issuecomment-314104781) I clearly suggest there should be 2 levels:
- static (class-level)
- dynamic (object-level)

Each method should come in pair: like:
- `attributes()` <-> `instanceAttributes()`
- `attributeLabels()` <-> `instanceAttributeLabels()`
- `attributeHints()` <-> `instanceAttributeHints()`

So default implementation of `instanceAttributes()` looks like following:

```php
public function instanceAttributes()
{
     return static::attributes();
}
```

which means at particular model you can overrite `instanceAttributes()` instead of `attributes()` to create object-specific runtime attributes list.

Whenever object is available `instance*` methods should be used, like at Validator:
https://github.com/yiisoft/yii2/blob/master/framework/validators/Validator.php#L406

Static methods should be used only in places, where AR instance is not availble, e.g. during GridView column label fetching and so on.

At 2.0.x branch we can extend methods list in opposite direction adding `staticAttributes()`, `staticAttributeLabels()` and so on in following way:

```php
public static function staticAttributeLabels()
{
    return static::staticModel()->attributeLabels();
}

protected static function staticModel()
{
     return new static(); // allow developer to override this method 
}
```

But such approach looks too complicated without much purpose - just to solve 'custom contructor' problem.

We can not now simply deny the issue #1318. However it is actually very uncommon. In most cases developers specifies attributes, labels and hints as static per whole class, while only in some cases they vary per object. having a pair of methods solves both this issue and #1318

klimov-paul:
> And as for metadata, I believe it's better to separate metadata to an independent object - like ActiveQuery for queries stuff. This way ActiveQuery and ActiveDataProvider would be able to get metadata without a need to touch model(that probably would have a lot of stuff that not needed to build queries).

This is not acceptable. We are not talking about `ActiveRecord` here - we are talking about `Model`!
Attributes, labels and hints are the part of `yii\base\Model` - we need solution at model level. ActiveRecord is just a particular instance of it.

klimov-paul:
> You see, we can easily add behavior to model through Container configuration to add some relations or other stuff dynamically. Or you can even switch specific model to some other realization by the same Container and in both situations, core won't get the correct code with the actual approach.

The problem is in struggling between 'static' and 'dynamic' information. At some places information about relations, attributes etc. is requried **before** actual AR object is instantiated - this is a static level.
On the other hand there is a demand of usage object-level customization for the same data via DI or Behavior - this is a dynamic level.
Support of both levels are required and it can not be provided through the same 'door'.

We can return to the approach, we used at Yii 1 with static method `model()`, which should be used to obtain all static information about AR like attributes and relations:

```php
public static function model()
{
     return new static(); // allow developer to override this method 
}
```

This PR is actually doing the same thing using `instantiate()` for it.

But I do not like returning to such approach. I would prefer to put things right at 2.1 then creation of such patch at 2.0.

With paired methods I suggets deveoper may achieve any behavior for particular ActiveRecord class as he is free to override both 'static' and 'non-static' method. Thus if he wants to use global static object with Behavior or DI interference - he can override 'static' methods for it.

samdark:
@ElisDN I'd like to hear your thoughts about @klimov-paul suggestions.

dynasource:
Id prefer 1 method name for instantiation framework wide. Not `createNew`, `create` or whatever other name that seems to be a solution for a problem at hand. 

I simply don't want to `remember` such different approaches for instantiation, whether it's in the Model scope or other scope. It should be 1 approach and easy to remember and perhabs even reside in `\yii\base\Object`

With respect thoses suggest `method pairing`. That is also something adding too much complexity. We should strive to have 1 way of doing things so we're not faced with any choices during development.

klimov-paul:
Alternative #14456

ElisDN:
@samdark I would prefer to leave `BaseActiveRecord::instantiate($row)` method and all dynamic scopes as is and just add `Model::model()` method from @klimov-paul's PR.

samdark:
Closing in favor of #14456

Issue #14035, opened 2017-04-23
batchInsert support ignore


rossoneri:
When I execute batch inserting into tables with unique index, I hope the `batchInsert` method of `yii\db\Command` Class have a argument to specify whether to execute `INSERT IGNORE INTO` or just `INSERT INTO` .

samdark:
INSERT IGNORE itself is MySQL-specific. Need to check if there are similar features in other DBs.

cebe:
related to #5138

Faryshta:
this is something i would like too but as @samdark said it would break with all databases except for one.

PR #14080, opened 2017-04-28
14035 batchinsert support ignore

qihuajun:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14035


framework/db/cubrid/QueryBuilder.php
```diff
@@ -193,4 +193,12 @@ private function getColumnDefinition($table, $column)
         }
         return null;
     }
+
+    /**
+     * @inheritdoc
+     */
+    public function batchInsert($table, $columns, $rows, $ignore = false, $replace = false)
+    {
+        return parent::batchInsert($table, $columns, $rows, false, false);
```
MysteryDragon: Maybe it's better to throw exception in such cases?

rossoneri: add exceptions

framework/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@ Yii Framework 2 Change Log
 
 2.0.12 under development
 --------------------------
+- Enh #14035: Support `INSERT IGNOTE INTO` or `REPLACE INTO` for MySQL while execute batchInsert
+- Bug #14012: `yii\db\pgsql\Schema::findViewNames()` was skipping materialized views (insolita)
```
samdark: Seems unrelated.

framework/db/Command.php
```diff
@@ -455,11 +455,13 @@ public function insert($table, $columns)
      * @param string $table the table that new rows will be inserted into.
      * @param array $columns the column names
      * @param array $rows the rows to be batch inserted into the table
+     * @param boolean $ignore whether to excute insert ignore into, only support MySQL
+     * @param boolean $replace whether to excute `repace into` instead of `insert into` , only support MySQL
      * @return $this the command object itself
      */
-    public function batchInsert($table, $columns, $rows)
+    public function batchInsert($table, $columns, $rows, $ignore = false, $replace = false)
```
samdark: Method signature change breaks backwards compatibility so with it the pull request could be merged into 2.1 only.

framework/CHANGELOG.md
rossoneri: Sorry,the second line was merged from upstream.

framework/db/Command.php
rossoneri: ok!

SilverFire:
`ignore` and `replace` options can not be used together.
I think it's better to create separate methods:

 - `batchInsertIgnore($table, $columns, $rows)`
 - `batchInsertReplace($table, $columns, $rows)`

Then we can deliver this feature without BC breaking and better semantic.

Faryshta:
@SilverFire with that approach it would break bc and can be added to 2.0 branch?

it can also save a lot of code by building a `UnsupportedInsertIgnore` trait to avoid repeating the same exception over and over

rossoneri:
@SilverFire Update My code according to your advice. Actually I thought that way at the beginning.

samdark:
Won't make it into 2.0.

PR #14081, opened 2017-04-28
Extracted CacheInterface from `yii\caching\Cache`

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | let's see :)
| Fixed issues  | Haven't found


samdark (approved):

framework/caching/Cache.php
```diff
@@ -31,7 +31,7 @@
  * }
  * ```
  *
- * Because Cache implements the [[\ArrayAccess]] interface, it can be used like an array. For example,
+ * Because CacheInterface extends the [[\ArrayAccess]] interface, it can be used like an array. For example,
```
SilverFire: Ok

framework/caching/CacheInterface.php
```diff
@@ -0,0 +1,224 @@
...
+     * Builds a normalized cache key from a given key.
+     *
+     * If the given key is a string containing alphanumeric characters only and no more than 32 characters,
+     * then the key will be returned back prefixed with [[keyPrefix]]. Otherwise, a normalized key
+     * is generated by serializing the given key, applying MD5 hashing, and prefixing with [[keyPrefix]].
+     *
+     * @param mixed $key the key to be normalized
+     * @return string the generated cache key
+     */
+    public function buildKey($key);
```
SilverFire: It was protected, but in July 2014 Qiang turned it to public for some reasons 2e71f6e

I do not think this method is vital, but I've added it to the interface because interface API should be compatible with current abstract class Cache. Don't you agree?

framework/caching/CacheInterface.php
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Retrieves multiple values from cache with the specified keys.
+     * Some caches (such as memcache, apc) allow retrieving multiple cached values at the same time,
+     * which may improve the performance. In case a cache does not support this feature natively,
+     * this method will try to simulate it.
+     * @param string[] $keys list of string keys identifying the cached values
+     * @return array list of cached values corresponding to the specified keys. The array
+     * is returned in terms of (key, value) pairs.
+     * If a value is not cached or expired, the corresponding array value will be false.
+     * @since 2.0.7
```
SilverFire: OK for this and other similar

framework/caching/CacheInterface.php
```diff
@@ -0,0 +1,224 @@
...
+     * In case $callable returns `false`, the value will not be cached.
+     * @param int $duration default duration in seconds before the cache will expire. If not set,
+     * [[defaultDuration]] value will be used.
+     * @param Dependency $dependency dependency of the cached item. If the dependency changes,
+     * the corresponding value in the cache will be invalidated when it is fetched via [[get()]].
+     * This parameter is ignored if [[serializer]] is `false`.
+     * @return mixed result of $callable execution
+     * @since 2.0.11
+     */
+    public function getOrSet($key, $callable, $duration = null, $dependency = null);
```
SilverFire: Same as `buildKey()`- I've added it to the interface because the interface API should be compatible with current abstract class Cache.

Otherwise we should write everywhere in PHPDocs something like `CacheInterface|Cache`.
We can move these methods to a separate `ExtendedCacheInterface` or so. @yiisoft/core-developers

dynasource (approved):

samdark:
Merged! That's a good one.

PR #14085, opened 2017-04-29
 Refactored yii\base\Module. Added tests

vladis84:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes/no
| Tests pass?   | yes


framework/base/Module.php
```diff
@@ -705,4 +705,60 @@ public function afterAction($action, $result)
         $this->trigger(self::EVENT_AFTER_ACTION, $event);
         return $event->result;
     }
+
+    /**
+     * Layout find in parent module.
+     * @return array [module, layout]
+     */
+    private function findLayout()
```
anton-pribora: A little bit better I think
```php
private function findLayout()
{
    $module  = $this;

    do {
        if (is_string($module->layout)) {
            return [$module, $module->layout];
        }
    } while (($module = $module->module) !== null);

    return [null, null];
}
```

framework/base/Module.php
```diff
@@ -705,4 +705,60 @@ public function afterAction($action, $result)
...
+        }
+
+        if (!is_string($layout)) {
+            return false;
+        }
+
+        if (Yii::isAlias($layout)) {
+            $file = Yii::getAlias($layout);
+        } else {
+            $file = $module->getLayoutPath() . DIRECTORY_SEPARATOR . ltrim($layout, '/');
```
anton-pribora: I think it should be one of `DIRECTORY_SEPARATOR . ltrim($layout, DIRECTORY_SEPARATOR);` or `'/' . ltrim($layout, '/')`

anton-pribora (approved):

framework/base/Controller.php
```diff
@@ -479,40 +479,10 @@ public function setViewPath($path)
      * Please refer to [[render()]] on how to specify this parameter.
      * @throws InvalidParamException if an invalid path alias is used to specify the layout.
      */
-    public function findLayoutFile($view)
+    public function findLayoutFile(View $view)
```
samdark: Method signature can't be changed in Yii 2.0.x because of backwards compatibility.

vladis84: ok

framework/BaseYii.php
```diff
@@ -128,7 +128,7 @@ public static function getVersion()
      */
     public static function getAlias($alias, $throwException = true)
     {
-        if (strncmp($alias, '@', 1)) {
+        if (!self::isAlias($alias)) {
```
cebe: wrapping a function call in an extra method is unnecessary.

framework/BaseYii.php
```diff
@@ -156,6 +156,15 @@ public static function getAlias($alias, $throwException = true)
     }
 
     /**
+     * @param string $alias
+     * @return bool
+     */
+    public static function isAlias($alias)
```
cebe: what is the use case for this method? There should not be a need to check for this.

cebe (commented):
can you explain a bit why you made these changes? i.e. move the logic from controller to module.

framework/BaseYii.php
vladis84: https://github.com/yiisoft/yii2/pull/14085/files#diff-d24221cf7f251030273dc9f1c436ba75R747
Code reuse

vladis84:
I think it is the duty of the module

framework/BaseYii.php
vladis84: Public api will be broken

cebe: how exactly? This adds unnecessary overhead.

vladis84: This method return 3 value.
First original parameter.
To alias path.
Three value of of alias not found false

framework/base/Module.php
```diff
@@ -705,4 +705,58 @@ public function afterAction($action, $result)
...
+
+    /**
+     * Layout find in module.
+     * @param string $layout    Layout name
+     * @param string $layoutExt View default extension
+     * @return boolean|string
+     */
+    public function findLayoutFile($layout, $layoutExt)
+    {
+        $module = $this;
```
dynasource: $module is confusing here. This should be in the next `else` statement

framework/base/Module.php
```diff
@@ -705,4 +705,58 @@ public function afterAction($action, $result)
...
+            }
+        }
+        while (($module = $module->module) !== null);
+
+        return [null, null];
+    }
+
+    /**
+     * Layout find in module.
+     * @param string $layout    Layout name
```
dynasource: this should be called $controllerLayout or $defaultLayout for a better understanding

framework/base/Module.php
dynasource: this method feels awkward. Why not simply use
```php
private function findLayout()
{
    $module = $this;
    while(module = $module->module){
        if (is_string($module->layout)) {
            return [$module, $module->layout];
        }
    }
    return false;
}
```

dynasource: additionally, it returns the $module + $layout, so the method should match this return result

dynasource (changes requested):

samdark:
@vladis84 are you going to work more on it?

samdark:
Won't be able to merge this into 2.0 and for 2.1 it doesn't make sense. Sorry.

PR #14087, opened 2017-04-29
add method registerCsrfMetaTags

RobinKamps:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Short version:
There is no convenient way to register csrf tags dynamically (with cache enabled).
`$view->registerCsrfMetaTags();` will register csrf meta tags.
`$view->registerCsrfMetaTags(true);` will register csrf meta tags rendered dynamically.

Long version:

To enable csrf in layout/view `echo Html::csrfMetaTags();` could be used.
However `$this->renderDynamic('echo yii\helpers\Html::csrfMetaTags();');` does not work properly if used in head section of layout.
It is placed before all page content - this could be a bug or wanted behavior? Anyway its causing some widgets to crash (like kartik switch input for example).

`$this->renderDynamic('return Yii::$app->view->registerMetaTag([\'name\' => \'csrf-token\', \'content\' => Yii::$app->request->csrfToken]);');` does not work obviously, too.

Using something like `registerMetaTag(['name' => 'csrf-token', 'content' => $this->renderDynamic('return Yii::$app->request->csrfToken;')]);` 
does not work, because renderTagAttributes are html encoded (thus < and > in BLOB are converted to `&lt;![CDATA[YII-DYNAMIC-0]]&gt;`  (should be &lt;![CDATA[YII-DYNAMIC-0]]&gt;))

Of course one can always add 

```php
echo '<meta name="csrf-param" content="' . $this->renderDynamic('return Yii::$app->request->csrfParam;') . '">';
echo '<meta name="csrf-token" content="' . $this->renderDynamic('return Yii::$app->request->csrfToken;') . '">';
```

to the head section of layout.

However a better way instead of putting this direct to layouts would be to use `registerMetaTags` in controller, so that the existence can be action specific. But dynamicRender fails on `registerMetaTag(['name' => 'csrf-token', 'content' => $this->renderDynamic('return Yii::$app->request->csrfToken;')]);` as mentioned above due to encoding.

With this little enhancement `$view->registerCsrfMetaTags();` could be used from controller or layout and if called with param `$view->registerCsrfMetaTags(true);` it will be automatic handled by dynamicRender.


dynasource:
shouldnt we have this rendered 'dynamic' by default? It's no use to have CRSF stuff cached

RobinKamps:
Good point.
I thought about just rendering it dynamic.
But I added the functionality to render it non dynamically for two reasons:
- to save minimal cpu-cycles for calling if there is no cache enabled at all...(saves only around 1 ms)
- more important: i´m not sure if there are any side effects on the renderDynamic function, if the ob_  settings are changed before/after calling renderDynamic (so addCrsfMetaTags could be used without cache and with custom ob_flush)

1. Shall I keep the functionality to render crsf-tags non dynamic - does anyone know about issues with ob_flushing  on renderDynamic or other drawbacks of only using renderDynamic?
2. You´re right: renderDynamic should be the default - i will change this if point 1. is clear.

dynasource:
it's good to put performance into consideration. I assume this loss to way less than <1 ms but more importantly, performance loss at pages without caching isnt a big deal either.

samdark (approved):
Looks good but tests are needed which should verify that cached page is getting different CSRF tags each time.

samdark:

dmirogin (approved):

rob006 (approved):

samdark:
Merged. Thanks for a pull request @RobinKamps and everyone for reviews.

PR #14088, opened 2017-04-29
Update Pjax.php

rymkapro:
When you reload part of a page, which contains pjax element, then links don't send pjax requests. After this change pjax links always work

### DIFF
--- a/framework/widgets/Pjax.php
+++ b/framework/widgets/Pjax.php
@@ -204,7 +204,7 @@ class Pjax extends Widget
         $js = '';
         if ($this->linkSelector !== false) {
             $linkSelector = Json::htmlEncode($this->linkSelector !== null ? $this->linkSelector : '#' . $id . ' a');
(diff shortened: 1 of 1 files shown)


SilverFire:
The fix will lead to another bug, when the same event handler will be attached to the document more than once.

rymka1989:
Don't think so, because it's delegeted event and it will be attached to document once. And then it will bubble to needed link (selector)

dynasource:
lets not implement changes to PJAX because of BC risks, unless more people request this patch

rymka1989:
I'm sorry, but can you explain me in details what BC risks could be? Because nothing was changed globally.

dynasource:
the risks are well explained by the number of issues in our tracker about PJAX
https://github.com/yiisoft/yii2/issues?utf8=%E2%9C%93&q=pjax

please navigate through these topics and you will get a better understanding. Because this is about PJAX , we are closing this issue. Feel free to create your own fork.

rymka1989:
Okay, thanks

PR #14017, opened 2017-04-20
Switched tests to asset-packagist

SilverFire:

PR #14091, opened 2017-05-01
Fix JS test after #14017

rob006:
It looks like JS tests works on travis only by accident - composer does not remove old dependencies in `vendor/bower`. But test fails on clean install: https://travis-ci.org/rob006/yii2-dev/jobs/227400723

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -107,14 +107,6 @@
     "extra": {
         "branch-alias": {
             "dev-master": "2.0.x-dev"
-        },
-        "asset-installer-paths": {
-            "npm-asset-library": "vendor/npm",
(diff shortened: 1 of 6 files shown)


cebe:
Thanks for noting that! I have removed travis cache, waiting for the build now. ref #14017

cebe:
here it fails: https://travis-ci.org/yiisoft/yii2/jobs/226964209#L358

cebe:
fixed by this PR, looks good to me: https://travis-ci.org/yiisoft/yii2/jobs/227547471

cebe:
Thank you!

PR #14098, opened 2017-05-01
Make FileHelper::normalizeOptions() protected

brandonkelly:
We have a need to call `normalizeOptions()` in an extension of FileHelper in Craft (preceding some calls to `filterPath()`). Assume there’s no harm in making it protected, so we don’t have to copy the function over to our class, just to work around the private visibility.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/helpers/BaseFileHelper.php
+++ b/framework/helpers/BaseFileHelper.php
@@ -696,6 +696,7 @@ class BaseFileHelper
     /**
      * @param array $options raw options
      * @return array normalized options
+     * @since 2.0.12
      */
     protected static function normalizeOptions(array $options)
     {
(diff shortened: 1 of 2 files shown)


brandonkelly:
For context, here’s why we need this: https://github.com/craftcms/cms/commit/e66b5a288fc56ba582ecf7ffebee84e942923357

Lines 426 - 518 could be deleted if `normalizeOptions()` were protected.

framework/helpers/BaseFileHelper.php
```diff
@@ -697,7 +697,7 @@ private static function firstWildcardInPattern($pattern)
      * @param array $options raw options
      * @return array normalized options
```
samdark: Need `@since` because it's now a part of accessible interface.

samdark (changes requested):

framework/helpers/BaseFileHelper.php
brandonkelly: @samdark Done

samdark:
Merged. Thank you!

brandonkelly:
Awesome, thanks!

PR #14101, opened 2017-05-02
Update AttributeBehavior.php

bscheshirwork:
Add `$values` feature

Provide difference assigned values by attributes is user friendly.
Array of behaviors:
```php
            [
                'class' => AttributeBehavior::className(),
                'attributes' => [
                    ActiveRecord::EVENT_BEFORE_INSERT => ['number', 'name'],
                ],
                'values' => [
                    'number' => [$this, 'getNextNumber'],
                    'name' => [$this, 'getNewName']
                ]
            ],
```
instead of 
```php
            [
                'class' => AttributeBehavior::className(),
                'attributes' => [
                    ActiveRecord::EVENT_BEFORE_INSERT => 'number',
                ],
                'value' => [$this, 'getNextNumber']
            ],
           [
                'class' => AttributeBehavior::className(),
                'attributes' => [
                    ActiveRecord::EVENT_BEFORE_INSERT => 'name',
                ],
                'value' => [$this, 'getNewName']
            ],
```

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no

### DIFF
--- a/framework/behaviors/AttributeBehavior.php
+++ b/framework/behaviors/AttributeBehavior.php
@@ -95,6 +95,10 @@ class AttributeBehavior extends Behavior
      */
     public $values = [];
     /**
+     * @var string current processed attribute name. Can be use inside callable item of [[values]].
+     */
+    public $currentAttribute;
+    /**
      * @var bool whether to skip this behavior when the `$owner` has not been
      * modified
      * @since 2.0.8
@@ -170,6 +174,7 @@ class AttributeBehavior extends Behavior
         if (array_key_exists($attribute, $this->values)) {
             $value = $this->value;
             $this->value = $this->values[$attribute];
+            $this->currentAttribute = $attribute;
             $result = $this->getValue($event);
             $this->value = $value;
             return $result;
--- a/framework/behaviors/AttributeBehavior.php
+++ b/framework/behaviors/AttributeBehavior.php
@@ -85,10 +85,12 @@ class AttributeBehavior extends Behavior
      * key - attribute name
      * value - corresponding value
      * Example:
+     * ```php
      * 'values' => [
      *   'number' => [$this, 'getNextNumber'],
      *   'status' => 'updated'
      * ]
+     * ```
      * @see value for possible values
      */
     public $values = [];
--- a/framework/behaviors/AttributeBehavior.php
+++ b/framework/behaviors/AttributeBehavior.php
@@ -80,6 +80,12 @@ class AttributeBehavior extends Behavior
      */
     public $value;
     /**
+     * @var array an array of the values indexed by attribute name that will be assigned to the current attributes respectively by attribute name.
+     * Will be assigned if $this->value is NULL
+     * The single value same at @see $value
+     */
+    public $values = [];
+    /**
      * @var bool whether to skip this behavior when the `$owner` has not been
      * modified
      * @since 2.0.8
@@ -112,9 +118,12 @@ class AttributeBehavior extends Behavior
         }
 
         if (!empty($this->attributes[$event->name])) {
-            $attributes = (array) $this->attributes[$event->name];
+            $attributes = (array)$this->attributes[$event->name];
             $value = $this->getValue($event);
             foreach ($attributes as $attribute) {
+                if (is_null($value)) {
+                    $value = $this->getValueByAttribute($event, $attribute);
+                }
                 // ignore attribute names which are not string (e.g. when set by TimestampBehavior::updatedAtAttribute)
                 if (is_string($attribute)) {
                     $this->owner->$attribute = $value;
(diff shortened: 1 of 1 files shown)


Kolyunya:
If I'm not mistaking this was rejected couple of years ago. One attribute behavior calculates one value by design.

bscheshirwork:
Some strange:
Multiply attributes can be change on each event, but all of this is same

Kolyunya:
Indeed multiple attributes can be changed but with the same value.

bscheshirwork:
So... I think this behavior can be more powerful. 
Because already can process multiply attributes.

framework/behaviors/AttributeBehavior.php
```diff
@@ -80,6 +80,12 @@ class AttributeBehavior extends Behavior
      */
     public $value;
     /**
+     * @var array an array of the values indexed by attribute name that will be assigned to the current attributes respectively by attribute name.
+     * Will be assigned if $this->value is NULL
```
SilverFire: <code>This property will be used only when [[value]] is \`null\`</code>

framework/behaviors/AttributeBehavior.php
```diff
@@ -80,6 +80,12 @@ class AttributeBehavior extends Behavior
...
+     * The single value same at @see $value
```
SilverFire: I'd clarify description a bit

key - attribute name
value - corresponding value
Example:
\`\`\`php
'values' => [
    'number' => [$this, 'getNextNumber'],
    'status' => 'updated'
]
\`\`\`
@see value for possible values

SilverFire: Add `@since 2.0.13`

framework/behaviors/AttributeBehavior.php
```diff
@@ -138,4 +147,25 @@ protected function getValue($event)
...
+     * Returns the value for the current attribute.
+     * This method is called by [[evaluateAttributes()]]. Its return value will be assigned
+     * to the attribute corresponding to the triggering event.
+     * @param Event $event the event that triggers the current attribute updating.
+     * @param string $attribute the attribute name.
+     * @return mixed the attribute value
+     */
+    protected function getValueByAttribute($event, $attribute)
+    {
+        if (array_key_exists($attribute, $this->values)) {
```
SilverFire: `array_key_exists` will return `true`, if the value is `null`.
Function `isset()` returns `false` for `null` value, but works faster. Since value `null` should not be handled in a special way, `isset($this->values[$attribute])` should be enough

SilverFire (changes requested):

framework/behaviors/AttributeBehavior.php
bscheshirwork: So... `null` is correct value for `value` (`value` is null in current representation by default)

framework/behaviors/AttributeBehavior.php
bscheshirwork: Add @since 2.0.13

its not `automatic` tag?

framework/behaviors/AttributeBehavior.php
SilverFire: So `isset` is suitable, isn't it?

bscheshirwork: >Since value null should not be handled in a special way
> So isset is suitable, isn't it?

BC: The `getValue()` can be redefine with `null`-expected logic. 
like this
```
     {
       if (is_null($this->value)) {
         return 'HAHAHA';
       } else  {
         return parent::getValue($event);
       }
     }
```

SilverFire: HAHAHA, ok :)

dynasource:
Im against this approach. The adding of a new array dimension makes AttributeBehavior too complex. The current state of that class already makes me have to look twice to understand the inner workings. 

I'd rather see solution in which AttributeBehavior gets added functionality + improvements on simplicity

bscheshirwork:
What about new behavior? Named someone like "AttributesBehavior"... This is avoid problem with BC => may be simplify.

dynasource:
that could be an idea. We could deprecate AttributeBehavior in favor of AttributesBehavior

framework/behaviors/AttributeBehavior.php
```diff
@@ -80,6 +80,25 @@ class AttributeBehavior extends Behavior
...
+     *   'status' => 'updated'
+     * ]
+     * ```
+     * @see value for possible values
+     */
+    public $values = [];
+    /**
+     * @var string current processed attribute name. Can be use inside callable item of [[values]].
+     */
+    public $currentAttribute;
```
Faryshta: why public instead of protected?

framework/behaviors/AttributeBehavior.php
```diff
@@ -112,9 +131,12 @@ public function evaluateAttributes($event)
         }
 
         if (!empty($this->attributes[$event->name])) {
-            $attributes = (array) $this->attributes[$event->name];
+            $attributes = (array)$this->attributes[$event->name];
             $value = $this->getValue($event);
             foreach ($attributes as $attribute) {
+                if (is_null($value)) {
```
Faryshta: `$value === null` is much faster

and what if i want to save a null value?

bscheshirwork: If u want save null see ln183

framework/behaviors/AttributeBehavior.php
bscheshirwork: May be discussed.
Fore external anonymous function.

And... I think new behavior (prev comment) is better idea.

samdark:
Closed in favor of #14151

Issue #13647, opened 2017-02-23
MSSQL Migration enahnchements


tsdogs:
When using migrations with MSSQL the dropColumn, dropTable are often blocked by constraints created automatically by SQL Server (i.e. a DEFAULT VALUE), and doing up and down is a PITA.

I have created the attached Migration class which I use to mitigate this.

N.B. This though does not solve the problem on when doing an alterTable and changing a field which has a constraint defined.
If I add a column link this:

$this->addColumn('{{%table}}','field',$this->integer()->null());
SQL Server would add a contraint for the NULL default value.

Then doing:
$this->alterColumn('{{%table}}','field',$this->string(40)->notNull());
 would FAIL as there is a constraint.
this->dropColumnConstraints('{{%table}}','field'); 
must be called before.

Maybe can be of help for integration in default yii2 Migration ...

tsdogs:
`<?php

namespace app\migrations;

use Yii;

/**
 * @author Alessandro Briosi <ab1@metalit.com>
 */
class Migration extends \yii\db\Migration
{
    /**
     * @var string
     */
    protected $tableOptions;
    protected $restrict = 'RESTRICT';
    protected $cascade = 'CASCADE';
    protected $dbType;
    

    /**
     * @inheritdoc
     */
    public function init()
    {
        parent::init();

        switch ($this->db->driverName) {
            case 'mysql':
                $this->tableOptions = 'CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE=InnoDB';
                $this->dbType = 'mysql';
                break;
            case 'pgsql':
                $this->tableOptions = null;
                $this->dbType = 'pgsql';
                break;
            case 'dblib':
            case 'mssql':
            case 'sqlsrv':
                $this->restrict = 'NO ACTION';
                $this->tableOptions = null;
                $this->dbType = 'sqlsrv';
                break;
            default:
                throw new \RuntimeException('Your database is not supported!');
        }
    }
    
    public function dropTable($table)
    {
        if ($this->dbType = 'sqlsrv') {
            $this->dropTableConstraints($table);
        }
        return parent::dropTable($table);
    }
    
    public function dropColumn($table, $column)
    {
        if ($this->dbType = 'sqlsrv') {
            $this->dropColumnConstraints($table, $column);
        }
        return parent::dropColumn($table, $column);
    }
    
    /*
     *  Drops contratints and Indexes referencind a Table Column
     */
    public function dropColumnConstraints($table, $column)
[...]

Issue #9003, opened 2015-07-05
Add createConstraint() to Migration


YasserHassan:
The believe it's important to have a function for creating constraints similar to `createIndex()`.

samdark:
Primary and foreign key constraints are already there. Do you mean something additional?

nineinchnick:
I'm using [check and exclusion constraints](http://www.postgresql.org/docs/devel/static/ddl-constraints.html) so it would be handy to add them like indexes and fks instead directly using `execute()`.

samdark:
Makes sense.

nineinchnick:
Actually, FK is just a specific constraint so its syntax could be more general.

YasserHassan:
It is important to consider that there are additional options for constraints. For example in PostgreSQL:
```
ALTER TABLE table1
ADD CONSTRAINT table1_col1_col2_key UNIQUE (col1, col2)
DEFERRABLE INITIALLY DEFERRED;
```
So the function signature could be something like this:
```
createConstraint($name, $table, $columns, $unique = false, $options = null)
```
Notice that PostgreSQL does not allow expressions in constraints so no need to check parenthesis and escape on them. I'm not sure about others databases though.

Probably it would be good for maintainability to consider letting `addForeignKey()` and `addPrimaryKey()` use `createConstraint()` internally and the same for dropping them (using `dropConstraint()`) as a future enhancement.

YasserHassan:
A more generic signature could be:
```
createConstraint($name, $table, $columns, $type = null, $value = null, $options = null)
```
where `$type` is CHECK, UNIQUE, ...etc, and value is something like `char_length(zipcode) = 5` either as array or string. Please take a look at the constraints examples at the end of [ALTER TABLE documentation](http://www.postgresql.org/docs/9.4/static/sql-altertable.html).

sergeymakinen:
I think this issue is now implemented in #14105.

buttflattery:
reading the comments looks like we can do it but could not find any notes so sorry for asking how can i create a migration where i have `CHECK(column<12)` constraint like below

```
CREATE TABLE user (
ID int NOT NULL,
Age int,
CHECK (Age>=18)
);
```

PR #14105, opened 2017-05-03
Implement retrieving DBMS constraints

sergeymakinen:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes (thanks to [this](https://github.com/sergeymakinen/docker-yii2-php) and [this](https://github.com/sergeymakinen/docker-test-db))
| Fixed issues  | partially #8163, also see #8152, #9003, #13647

This small PR adds:

- [x] DBMS constraints retrieving methods.
- [x] DBMS constraints low-level manipulation methods.
- [x] DMBS constraints `QueryBuilder` tests.
- [x] DMBS constraints `Command` tests.

Please note: 

- This PR is **not** a complete solution for a constraints management. It's a first part that adds all that low-level stuff. I will implement other parts during a 2.0.13 development timeframe.
- Some existing code is outdated or not needed anymore (or both O:). But it requires a discussion and deserves a separate issue.
- The SQL tokenizer is aimed mostly at db developers (= (99% of time) no one beside Yii db core developers/contributors). It's not an all-purpose solution. It helps to get a required information from the SQL code without getting trapped by regex limitations.
- MySQL's constraint code works with 5.1+. If this is still an issue (tests only pass on 5.0+), I will make it BC with 4.x+/5.0+ (with a help of `SqlTokenizer`).
- `Schema` caching was updated, now it's trivial to add any new information along with the existing without having to think about caching. Tests were intentionally left untouched to catch regressions (if any).


SilverFire:
Wow, amazing!

sergeymakinen:
I have added `Command` tests and rebased the commits.

samdark:
@sergeymakinen is it ready for review?

sergeymakinen:
@samdark yes, it is!

samdark:
@sergeymakinen have you considered https://github.com/yiisoft/yii2/issues/13647#issuecomment-309023565?

sergeymakinen:
@samdark I did. My current plan is to leave all high level logic in a separate class(-ess). I have started writing tests for this. Pros: it works with any constraint aware Schema/Command classes, it's db independent:

```php
$this->table('test')
    ->column('fk_id')->ifExists()->drop()
    ->alter();
```

Funny thing, it's all still simple. As soon as all tests pass, I'll prepare a PR.

sergeymakinen:
Adding a `dropConstraints()`/`withoutConstraints()->save()` in this case is really trivial.

framework/db/Command.php
```diff
@@ -736,7 +736,110 @@ public function dropIndex($name, $table)
...
+     * Creates a SQL command for adding a check constraint to an existing table.
+     * @param string $name the name of the check constraint.
+     * The name will be properly quoted by the method.
+     * @param string $table the table that the check constraint will be added to.
+     * The name will be properly quoted by the method.
+     * @param string $check the SQL of the `CHECK` constraint.
+     * @return $this the command object itself.
+     * @since 2.0.13
+     */
+    public function addCheck($name, $table, $check)
```
samdark: Need to explicitly mention in the docs that MySQL simply ignores CHECK constraints.

framework/db/ConstraintFinderTrait.php
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\db;
+
+/**
+ * ConstraintFinderTrait provides methos for getting a table constraint information.
```
samdark: metho**d**s

framework/db/mysql/Schema.php
```diff
@@ -53,6 +68,128 @@ class Schema extends \yii\db\Schema
         'varbinary' => self::TYPE_BINARY,
     ];
 
+    /**
+     * @inheritDoc
+     */
+    protected function resolveTableName($name)
+    {
+        $resolvedName = new TableSchema();
+        $parts = explode('.', str_replace('`', '', $name));
```
samdark: Would it work in ANSI mode where quotes are `"`?

samdark:
Overall I love the direction of these changes but I'm a bit concerned about backwards compatibility. What would be broken for the end users at this point?

framework/db/mysql/Schema.php
sergeymakinen: It should not but [it doesn't seem](https://github.com/sergeymakinen/yii2/blob/3de6f8e397d336d59a7fd8c501a4132631c29563/framework/db/mysql/Schema.php#L200-L214) to bother anyone yet.
Should I fix all of this if it's possible?

framework/db/Command.php
sergeymakinen: Sure. I suppose the migration docs should mention all of this new possibilities along with an upcoming table auto column/constraint management. That will be a next part of my work.

framework/db/Command.php
```diff
@@ -736,7 +736,110 @@ public function dropIndex($name, $table)
...
+     * The name will be properly quoted by the method.
+     * @param string $table the table that the unique constraint will be added to.
+     * The name will be properly quoted by the method.
+     * @param string|array $columns the name of the column to that the constraint will be added on.
+     * If there are multiple columns, separate them with commas.
+     * The name will be properly quoted by the method.
+     * @return $this the command object itself.
+     * @since 2.0.13
+     */
+    public function addUnique($name, $table, $columns)
```
sergeymakinen: This migration enchantments cover almost all usable constraints.

I think it _may_ be worth to keep uniques as well as it can be supported (I mean a distinction between unique constraints and unique indexes) in next versions or may be handy when extending core classes.

But I'm not sure. I still can't way any (useful in PHP) db which doesn't create an index behind the constraint. (NoSQL?..)

I would like to hear opinions if it should go or stay.

sergeymakinen:
@samdark thanks for the review! It should not affect a current code. I redesigned a table metadata caching but kept old tests and even added a version in a cache bucket to exclude caching issues. Other things are additions and they don't touch any existing code.

Existing methods to retrieve constraint/column info should be replaced/updated to use this new code but, well, 6k lines changed, so I suggest to review and merge this code, then I'll roll out a PR with a new table builder, then I'll start updating docs and prepare another PR to replace/update old methods but only where BC is safe and secure.

framework/db/Command.php
samdark: Isn't PostgreSQL supporting both?

framework/db/Command.php
samdark: This one isn't about migration docs. It's about phpdoc.

framework/db/mysql/Schema.php
samdark: Yes, I think so.

samdark: Maybe in a separate PR. If current PR doesn't support ANSI quotes and noone complains then probably it's fine to have it as is.

framework/db/Command.php
sergeymakinen: > PostgreSQL automatically creates a unique index when a unique constraint or primary key is defined for a table.

It works the [same](https://www.postgresql.org/docs/9.4/static/indexes-unique.html). Em.

framework/db/Command.php
sergeymakinen: Oh. Okay. :)

framework/db/mysql/Schema.php
sergeymakinen: Then I'll do it later to avoid unneeded rebasing and merging.

framework/db/Command.php
samdark: OK. Let's have it.

framework/db/ConstraintFinderTrait.php
```diff
@@ -0,0 +1,217 @@
...
+     * Returns primary keys for all tables in the database.
+     * @param string $schema the schema of the tables. Defaults to empty string, meaning the current or default schema name.
+     * @param bool $refresh whether to fetch the latest available table schemas. If this is `false`,
+     * cached data may be returned if available.
+     * @return Constraint[] primary keys for all tables in the database.
+     * Each array element is an instance of [[Constraint]] or its child class.
+     */
+    public function getSchemaPrimaryKeys($schema = '', $refresh = false)
+    {
+        return $this->getSchemaMetadata($schema, 'primaryKey', $refresh);
```
dynasource: shouldn't we use `constants` for strings linke 'primaryKey'?

framework/db/Schema.php
```diff
@@ -645,4 +594,162 @@ public function isReadQuery($sql)
         $pattern = '/^\s*(SELECT|SHOW|DESCRIBE)\b/i';
         return preg_match($pattern, $sql) > 0;
     }
+
+    /**
+     * Returns the cache key for the specified table name.
+     * @param string $name the table name
+     * @return mixed the cache key
```
dynasource: mixed? Additionally, isnt key a bit confusing as a name for returning an array?

dynasource (changes requested):
impressive effort. For what I've checked, it looks good. With respect to BC, I tested the branch in one my projects and all tests pass in Sqlite and Mysql, so thats a good signal.

framework/db/Command.php
samdark: I think @dynasource suggestion makes perfect sense.

framework/db/ConstraintFinderTrait.php
samdark: Constants could not be private till PHP 7.2. Exposing implementation details isn't a good idea IMO.

framework/db/Command.php
sergeymakinen: It sounds reasonable, but here are thoughts behind `addUnique`:
1. Almost all `Command` / `QueryBuilder` names follow SQL commands.
2. We `CREATE INDEX`es but we `ADD CONSTRAINT`s when we `ALTER TABLE`.
3. `Unique` isn't ideal, but why don't we use `Constraint` suffix elsewhere then (hint: BC)? I prefer to stick with consistency - like @dynasource said.

But I'm not principled in this case.

framework/db/Schema.php
sergeymakinen: It's because a key can be [anything](https://github.com/yiisoft/yii2/blob/master/framework/caching/Cache.php#L93) JSON serializable.
I can't change it because it's `protected`, the return type is wide (`mixed`) and we're in 2.0.*

framework/db/Constraint.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @author Sergey Makinen <sergey@makinen.ru>
+ * @since 2.0.13
+ */
+class Constraint extends Object
+{
+    /**
+     * @var string[]|null list of column names the constraint belongs to.
+     */
+    public $columnNames;
```
samdark: Wouldn't it make sense to default to `[]` instead of `null` here?

framework/db/IndexConstraint.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @author Sergey Makinen <sergey@makinen.ru>
+ * @since 2.0.13
+ */
+class IndexConstraint extends Constraint
+{
+    /**
+     * @var bool whether the index is unique.
+     */
+    public $isUnique;
```
samdark: Shouldn't it be `false` by default?

framework/db/IndexConstraint.php
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * @var bool whether the index was created for a primary key.
+     */
+    public $isPrimary;
```
samdark: Shouldn't it be `false` by default?

framework/db/Schema.php
```diff
@@ -89,119 +92,91 @@
...
-     * @return mixed the cache key
+     * Loads the metadata for the specified table.
+     * @param string $name table name
+     * @return TableSchema|null DBMS-dependent table metadata, `null` if the table does not exist.
      */
-    protected function getCacheKey($name)
+    abstract protected function loadTableSchema($name);
+
+    /**
+     * @return ColumnSchema
```
samdark: Need phpdoc.

framework/db/CheckConstraint.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\db;
+
+/**
+ * CheckConstraint represents the metadata of a table `CHECK` constraint.
```
samdark: Alternatively, MySQL schema may just throw NotSupportedException since, in fact, CHECK is not supported while syntax is accepted.

framework/db/Constraint.php
sergeymakinen: `null` here and in other places means "we got no info".

framework/db/CheckConstraint.php
sergeymakinen: I do agree with `NotSupportedException`. I will check other cases as well.

framework/db/Constraint.php
samdark: OK.

sergeymakinen:
Probably most issues are addressed.

framework/db/SqlToken.php
```diff
@@ -0,0 +1,296 @@
...
+
+        if ($patternToken->children === $token->children) {
+            $firstMatchIndex = $lastMatchIndex = $offset;
+            return true;
+        }
+
+        $firstMatchIndex = $lastMatchIndex = null;
+        $wildcard = false;
+        for ($index = 0, $count = count($patternToken->children); $index < $count; $index++) {
+            if ($patternToken[$index]->content === 'any') {
```
samdark: Looks suspicious. You're incrementing children index but accessing parent with it. Is it intended?

framework/db/SqlToken.php
```diff
@@ -0,0 +1,296 @@
...
+                $wildcard = true;
+                continue;
+            }
+
+            for ($limit = $wildcard ? count($token->children) : $offset + 1; $offset < $limit; $offset++) {
+                if (!$wildcard && !isset($token[$offset])) {
+                    break;
+                }
+
+                if (!$this->tokensMatch($patternToken[$index], $token[$offset])) {
```
samdark: Same here. You're incrementing children index but accessing parent with it. Is it intended?

framework/db/mssql/Schema.php
```diff
@@ -73,6 +81,169 @@ class Schema extends \yii\db\Schema
...
+    protected function loadTableSchema($name)
+    {
+        $table = new TableSchema();
+        $this->resolveTableNames($table, $name);
+        $this->findPrimaryKeys($table);
+        if ($this->findColumns($table)) {
+            $this->findForeignKeys($table);
+
+            return $table;
+        } else {
```
samdark: Else isn't needed.

framework/db/mysql/Schema.php
```diff
@@ -53,6 +69,130 @@ class Schema extends \yii\db\Schema
...
+    protected function loadTableSchema($name)
+    {
+        $table = new TableSchema();
+        $this->resolveTableNames($table, $name);
+
+        if ($this->findColumns($table)) {
+            $this->findConstraints($table);
+
+            return $table;
+        } else {
```
samdark: Else isn't needed.

framework/db/oci/Schema.php
```diff
@@ -45,6 +54,181 @@ public function init()
...
+     * @inheritDoc
+     */
+    protected function loadTableSchema($name)
+    {
+        $table = new TableSchema();
+        $this->resolveTableNames($table, $name);
+        if ($this->findColumns($table)) {
+            $this->findConstraints($table);
+            return $table;
+        } else {
```
samdark: Else isn't needed.

samdark (changes requested):

framework/db/Schema.php
```diff
@@ -645,4 +596,162 @@ public function isReadQuery($sql)
...
+    }
+
+    /**
+     * Returns the cache tag name.
+     * This allows [[refresh()]] to invalidate all cached table schemas.
+     * @return string the cache tag name
+     */
+    protected function getCacheTag()
+    {
+        return md5(serialize([
```
sergeymakinen: Will it affect BC?

samdark: It could. Let's leave it as is for now.

framework/db/SqlToken.php
sergeymakinen: It is: `ArrayAccess` provides access to the `children` property with a handy ability to specify negative offsets which makes the code shorter and easier to understand.
Basically we iterate token by token with an exception of `any` that toggles an iteration until we matched with a next pattern token or EOF.
Is it clear or should I add this description before `foreach`? As 3 months passed by I found this description useful 'cause it took some time to remember…

framework/db/ConstraintFinderTrait.php
sergeymakinen: > Note:
As of PHP 7.1.0 visibility modifiers are allowed for class constants.

[Source](http://php.net/manual/en/language.oop5.constants.php#language.oop5.basic.class.this)

I think it's a good idea to constantify those magic strings in 2.1.

framework/db/SqlToken.php
samdark: Yes, a comment would be helpful in this case, I think. Wasn't easy to get it.

framework/db/ConstraintFinderTrait.php
samdark: Yes.

framework/db/SqlToken.php
```diff
@@ -0,0 +1,296 @@
...
+            return false;
+        }
+
+        $patternToken = $patternToken[0];
+        return $this->tokensMatch($patternToken, $this, $offset, $firstMatchIndex, $lastMatchIndex);
+    }
+
+    /**
+     * Tests the given token to match the specified pattern token.
+     * @param SqlToken $patternToken
```
sergeymakinen: AFAIK there are no phpdoc requirements for `private`s, em?

samdark: These should be described as well.

sergeymakinen: I don't think there's any need for param docs as I covered it in the public method. But anyway take a look at it again after a commit is pushed.

sergeymakinen:
@samdark okay, your turn! :)

samdark (approved):

sergeymakinen:
Just a side note: while working my high-level extension I figured out there's no clean way to check if a schema supports constraints if we don't introduce an interface for them. Extracting public methods from `ConstraintFinderTrait` to `ConstraintAwareInterface` would be a good idea IMO.

Similar case with views but I believe they should be a part of the core `Schema` in 2.1 (and it's another issue).

samdark:
Do you think this pull request should include this interface?

sergeymakinen:
IMO we can include in a next PR which will introduce this table migrator along with some new DMBS properties in `Schema` (like `$maxIdentifierLength`).

samdark:
OK. I'm fine merging it. Waiting for approval from @klimov-paul, @SilverFire, @cebe, @dynasource.

framework/db/cubrid/Schema.php
```diff
@@ -306,4 +389,44 @@ public function createColumnSchemaBuilder($type, $length = null)
...
+     * @param string $returnType return type:
+     * - indexes
+     * - uniques
+     * @return mixed constraints.
+     */
+    private function loadTableConstraints($tableName, $returnType)
+    {
+        $constraints = $this->db->getSlavePdo()->cubrid_schema(\PDO::CUBRID_SCH_CONSTRAINT, $tableName);
+        $constraints = ArrayHelper::index($constraints, null, ['TYPE', 'NAME']);
+        ArrayHelper::multisort($columns, 'KEY_ORDER', SORT_ASC, SORT_NUMERIC);
```
cebe: variable `$columns` is not defined here? please make sure this is covered by a test. A simple one is sufficient for now.

cebe (changes requested):
Looks good to me overall, thanks for the huge effort!

I have some minor remarks, see comments I made. Also, a CHANGELOG entry is missing :)

Also two questions:

- why is default value a constraint?
- why is index a constraint?

cebe:
@sergeymakinen made some minor changes and pushed to your branch.

framework/db/SqlToken.php
```diff
@@ -0,0 +1,301 @@
...
+     * @param string $pattern SQL code. Everything supported in SQLite is supported here as well.
+     * In addition, `any` keyword is supported which is treated as any number of keywords, identifiers, whitespaces.
+     * @param int $offset token children offset to start lookup with.
+     * @param int|null $firstMatchIndex token children offset where a successful match begins.
+     * @param int|null $lastMatchIndex token children offset where a successful match ends.
+     * @return bool whether this token matches the pattern SQL code.
+     */
+    public function matches($pattern, $offset = 0, &$firstMatchIndex = null, &$lastMatchIndex = null)
+    {
+        $patternToken = (new \yii\db\sqlite\SqlTokenizer($pattern))->tokenize();
```
sergeymakinen: SQLite's one is only implemented ATM so it's generic for now. We can change it later.

framework/db/ConstraintFinderTrait.php
```diff
@@ -0,0 +1,217 @@
...
+ * @since 2.0.13
+ */
+trait ConstraintFinderTrait
+{
+    /**
+     * Loads a primary key for the given table.
+     * @param string $tableName table name.
+     * @return Constraint|null primary key for the given table, `null` if the table has no primary key.
+     */
+    abstract protected function loadTablePrimaryKey($tableName);
```
sergeymakinen: It's called (automagically) by `getTableMetadata` & `getSchemaMetadata`. I forgot to mention it in the method descriptions. Will fix it.

framework/db/ConstraintFinderTrait.php
```diff
@@ -0,0 +1,217 @@
...
+ * an array of [[IndexConstraint]] or its child classes. This property is read-only.
+ * @property Constraint[] $schemaPrimaryKeys Primary keys for all tables in the database. Each array element
+ * is an instance of [[Constraint]] or its child class. This property is read-only.
+ * @property IndexConstraint[][] $schemaUniques Unique constraints for all tables in the database.
+ * Each array element is an array of [[IndexConstraint]] or its child classes. This property is read-only.
+ *
+ * @author Sergey Makinen <sergey@makinen.ru>
+ * @since 2.0.13
+ */
+trait ConstraintFinderTrait
```
sergeymakinen: Do you mean by replacing abstracts with `NotSupportedException` (to maintain BC)? It's possible. The only reason behind this trait was to possibly make `Schema` look "lighter" if someone doesn't need a constraint support at all.
Maybe it's just an overcomplication and we simply need to put them all in `Schema`. I even suggested the same for `ViewFinderTrait`… I don't know TBH. :)
cc @samdark

framework/db/ConstraintFinderTrait.php
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Obtains the primary key for the named table.
+     * @param string $name table name. The table name may contain schema name if any. Do not quote the table name.
+     * @param bool $refresh whether to reload the information even if it is found in the cache.
+     * @return Constraint|null table primary key, `null` if the table has no primary key.
+     */
+    public function getTablePrimaryKey($name, $refresh = false)
+    {
+        return $this->getTableMetadata($name, 'primaryKey', $refresh);
```
sergeymakinen: Sounds reasonable but let's decide on merging `ConstraintFinderTrait` with `Schema`. Maybe these problems will be auto solved.

framework/db/cubrid/Schema.php
sergeymakinen: It actually is covered… I just wanted to be sure an array data is really sorted if something changes in next versions and CUBRID will start returning array values in random order. Thanks for spotting! :)

sergeymakinen:
@cebe thank you for the review! I fixed/answered the questions. We need to decide if we really need a trait or not. If not, if makes `Schema` larger but easier to work with. And it will eliminate the need for the additional interface at all.

sergeymakinen:
Oh, how dedicated questions:
1. Because defaults are constraints in some DBMS. Like MSSQL.
2. Because uniques constraints are usually "powered up" by indexes or vice versa. So they're tied together and it's easier to call an index a constraint and work with them the same way.

framework/db/SqlToken.php
cebe: Changing it later might need to break BC so I'd prefer to change it now. how about passing a SqlToken directly so the tokenizing is made outside of the matches() call?

framework/db/ConstraintFinderTrait.php
cebe: its fine to have them abstract in Schema, but I see the reasoning about allowing to implement Schema without the need to implement the trait. It makes sense and avoids BC break with custom DB implementations. Looks good to me. :shipit:

framework/db/SqlToken.php
sergeymakinen: It sounds good to me. :)

cebe:
@sergeymakinen I proposed some changes in https://github.com/sergeymakinen/yii2/pull/1

sergeymakinen:
@cebe merged, thanks! :)

samdark:
@sergeymakinen please add a line for changelog.

cebe (approved):

cebe:
Merged, thank you very much!

sergeymakinen:
Thanks everyone for reviews! And see you in a next PR. ;)

Issue #14103, opened 2017-05-03
Rbac : getUserIdsByRole() returns wrong results when assign role as a nested


hirenbhut93:
### What steps will reproduce the problem?
I have created two role one is **Admin** and another is **Author**.

Then i can add **Author role** in **Admin Role** as a child items.

Then i can get which user where assign **Author** role, but **Admin** role user is not return. 

The `getUserIdsByRole()` is not return which user where assign `Admin` role, It's only return those user where role/permission is directly assign.

```php
$auth = Yii::$app->authManager;

$rule = new \app\rbac\UserGroupRule;
$auth->add($rule);

$author = $auth->createRole('author');
$author->ruleName = $rule->name;
$auth->add($author);
// ... add permissions as children of $author ...

$admin = $auth->createRole('admin');
$admin->ruleName = $rule->name;
$auth->add($admin);
$auth->addChild($admin, $author);

$userId = Yii::$app->authManager->getUserIdsByRole('Author'); //only return those user where assign role directly not return admin role of user.
```
### What is the expected result?
Check nested/inherited permissions for given role/permission then return userid.
### What do you get instead?
Only return those user where directly assign role/permission. not check nested.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12-dev
| PHP version      | 7.1
| Operating system | Ubuntu 16

samdark:
Yes, indeed currently it returns only directly assigned roles.

samdark:
Will dig into it after 2.0.12 release.

Kolyunya:
Isn't this a duplicate of https://github.com/yiisoft/yii2/issues/11895?

samdark:
It is. Thanks!

PR #14114, opened 2017-05-05
Eng #14103: Rbac Manager's `getUserIdsByRole()` can return all userid…

qihuajun:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14103


framework/rbac/DbManager.php
```diff
@@ -1008,18 +1008,59 @@ public function loadFromCache()
...
-     * @return string[] the ids. An empty array will be
+     * @param bool $recursively whether to search userids recursively.
+     * If it is true, all users assigned roleName role and roleName's all parent
+     * roles will be found.
+     * @return \string[] the ids. An empty array will be
      * returned if role is not assigned to any user.
      * @since 2.0.7
      */
-    public function getUserIdsByRole($roleName)
+    public function getUserIdsByRole($roleName, $recursively = false)
```
samdark: Can't merge it into 2.0.x since method signature was changed. It breaks backwards compatibility.

Could be either merged into 2.1 or adjusted to keep method signature.

samdark (changes requested):

Kolyunya:
@rossoneri have you seen https://github.com/yiisoft/yii2/issues/11895 and https://github.com/yiisoft/yii2/pull/11927?

framework/rbac/DbManager.php
samdark: Sorry, I'm wrong. Such method signature change is actually backwards compatible.

Kolyunya: @samdark `ManagerInterface` change still breaks the backward compatibility.

samdark: https://3v4l.org/VSVdn — default value is a key.

samdark: The interface change above still breaks BC.

samdark:
Yes. Looks like duplicate solution. Need to check both and choose one.

PR #14100, opened 2017-05-01
Add php-cs-fixer coding standards validation to Travis CI

rob006:
Alternative PR to #13732 

Right now it contains only some basic rules, but I will try to extend it in this week.

Refs: https://github.com/FriendsOfPHP/PHP-CS-Fixer

PR #14119, opened 2017-05-07
fix code style

unasoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14119

### DIFF
--- a/framework/console/controllers/MessageController.php
+++ b/framework/console/controllers/MessageController.php
@@ -363,13 +363,13 @@ EOD;
     protected function saveMessagesToDb($messages, $db, $sourceMessageTable, $messageTable, $removeUnused, $languages, $markUnused)
     {
         $currentMessages = [];
-        $rows = (new Query)->select(['id', 'category', 'message'])->from($sourceMessageTable)->all($db);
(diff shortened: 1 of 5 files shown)


cebe:
I am against this change, it makes code less readable and does not add any value.
When creating new class on the fly in parenthesis it is better to leave off constructor parenthesis if nothing is passed to the constructor.

ejsoft:
I think it's better to stick to some kind of standard, or use the `new className ()` or `new className` construct, presently it's used so and so.

rob006:
[PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md) requires parenthesis.

samdark:
I think it's better to accept the change.

rob006:
BTW: In #14100 there is [fixer for this](https://github.com/yiisoft/yii2/pull/14100/commits/dddd5c50d67b5e0ac39a91753dcf0a6dfd1988f7) - it will not fix examples in phpdocs, but will force parenthesis in code.

samdark:
Closing in favor of @rob006 pull request.

Issue #7643, opened 2015-03-09
Sending $params to $user->can() in AccessRule


alagodich:
Hi

Should not we have a way to send params to access rules when configuring them as a controller's behaviour?

For example in your [guide] (http://www.yiiframework.com/doc-2.0/guide-security-authorization.html) you are creating `updateOwnPost` rule which is expecting `postId` param. But we don't have a way to send this param when configuring rules this way:
```PHP
    public function behaviors()
    {
        return [
            'access' => [
                'class' => AccessControl::className(),
                'rules' => [
                    [
                        'actions' => ['update'],
                        'allow' => true,
                        'roles' => ['admin', 'updatePost', 'updateOwnPost'],
                    ],
```
In `AccessRule` class we just [call] (https://github.com/yiisoft/yii2/blob/master/framework/filters/AccessRule.php#L152) `can()` method without any params.
So `updatePost` will not resolve with `updateOwnPost` as we cannot send `postId`.

Is it a bad practice and i have to use the following right in the action?
```PHP
if (\Yii::$app->user->can('updatePost', ['post' => $post])) {
    // update post
}
```
I would prefer to keep all access assigning in one place.

cebe:
I am currently doing it like that:

```php
    public function behaviors()
    {
        return [
            'access' => [
                'class' => AccessControl::className(),
                'rules' => [
				[
					'actions' => ['create', 'quick-create'],
					'allow' => true,
					'matchCallback' => function($rule, $action) {
						return Yii::$app->user->can('task:create', ['project' => $action->controller->findProject()]);
					},
				],
...
```

alagodich:
Right, i can use that, thank you.

PR #14123, opened 2017-05-08
Alternative implementation for passing parameters to AccessRule roles

cebe:
This is an alternative to #8426:

- more flexibility by allowing Closure, thanks @Faryshta for the idea
- no special syntax needed to use it
- added tests

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #8426 #7643 #4302

### DIFF
--- a/docs/guide/security-authorization.md
+++ b/docs/guide/security-authorization.md
@@ -491,36 +491,34 @@ public function behaviors()
 If all the CRUD operations are managed together then it's a good idea to use a single permission, like `managePost`, and
 check it in [[yii\web\Controller::beforeAction()]].
 
-You can also use rules from the [[yii\filters\AccessControl|AccessControl]] filter. For that you specify the
-[[yii\filters\AccessRule::roleParams|roleParams]] that you need to pass to the [[yii\filters\AccessRule|AccessRule]]:
-
+In the above example, no parameters are passed with the roles specified for accessing an action, but in case of the
+`updatePost` permission, we need to pass a `post` parameter for it to work properly.
+You can pass parameters to [[yii\web\User::can()]] by specifying [[yii\filters\AccessRule::roleParams|roleParams]] on
+the access rule:
 
 ```php
-use yii\filters\AccessControl;
-
-class PostsController extends Controller
-{
-    public function behaviors()
-    {
-        return [
-            'access' => [
-                'class' => AccessControl::className(),
-                'rules' => [
-                    [
-                        'actions' => ['update'],
-                        'roles' => ['updatePost'],
-                        'roleParams' => function() {
-                            return ['post' => Post::findOne(Yii::$app->request->get('id'))];
-                        },
-                        'allow' => true,
-                    ],
-                ],
-            ],
-        ];
-    }
-}
+[
+    'allow' => true,
+    'actions' => ['update'],
+    'roles' => ['updatePost'],
+    'roleParams' => function() {
+        return ['post' => Post::findOne(Yii::$app->request->get('id'))];
+    },
+],
 ```
 
+In the above example, [[yii\filters\AccessRule::roleParams|roleParams]] is a Closure that will be evaluated when
+the access rule is checked, so the model will only be loaded when needed.
+If the creation of role parameters is a simple operation, you may just specify an array, like so:
+
+```php
+[
+    'allow' => true,
+    'actions' => ['update'],
+    'roles' => ['updatePost'],
+    'roleParams' => ['postId' => Yii::$app->request->get('id')];
+],
+```
 
 ### Using Default Roles <span id="using-default-roles"></span>
 
--- a/docs/guide/security-authorization.md
+++ b/docs/guide/security-authorization.md
@@ -376,14 +376,7 @@ class AuthorRule extends Rule
      */
     public function execute($user, $item, $params)
     {
-        if (isset($params['post'])) {
-            $post = $params['post'];
-            if (is_integer($post)) {
-                $post = Post::findOne($post);
-            }
-            return $post->createdBy == $user;
-        }
-        return false;
+        return isset($params['post']) ? $params['post']->createdBy == $user : false;
     }
 }
 ```
@@ -517,7 +510,7 @@ class PostsController extends Controller
                         'actions' => ['update'],
                         'roles' => ['updatePost'],
                         'roleParams' => function() {
-                            return ['post' => Yii::$app->request->get('id')];
+                            return ['post' => Post::findOne(Yii::$app->request->get('id'))];
                         },
                         'allow' => true,
                     ],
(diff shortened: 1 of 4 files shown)


framework/filters/AccessRule.php
```diff
@@ -161,8 +191,13 @@ protected function matchRole($user)
                 if (!$user->getIsGuest()) {
                     return true;
                 }
-            } elseif ($user->can($role)) {
-                return true;
+            } else {
+                if (!isset($roleParams)) {
```
Faryshta: how about taking this outside the `foreach()` statement?

```php
$roleParams = $this->roleParams instanceof Closure
    ? call_user_func($this->roleParams, $this)
    : $this->roleParams;
foreach ($this->roles as $role) {
```

Faryshta:
the only thing 'missing' are the special parameters '{controller}' and '{action}' which can be passed manually like this:

```php
'accessRules' => [
    'roleParams' => function ($accessRule) {
         return ['{controller}' => $this->id, '{action}' => $this->action->id];
         // or
         return ['{controller}' => $this->route, '{action}' => $this->action->id];
    }
]
```
so i don't think it needs to be added to the core

framework/filters/AccessRule.php
cebe: I moved this into the foreach to evaluate it only if it is needed, e.g. it will not evaluate if '?' or '@' rule will return true first.

Faryshta: lets say i have `images` which any auth user ('@') can upload but only the owner can delete them.

or anyone can upload images anonymously ('?') but only the creator ip can delete them?

samdark (approved):

cebe:
just found this comment: https://github.com/yiisoft/yii2/issues/4302#issuecomment-223272992 and I like that solution a bit more...

framework/filters/AccessRule.php
Faryshta: and a bit off-topic but i can't load the original commit.

can you edit line 148?

`return $this->allow ? true : false;`

`$allow` is bool so the ternary is unneeded.

```php
public $allow = false; // currently is null which gets taken as false
// ...
return $this->allow;
```php

or

`return (bool) $this->allow; // null case remains identical to 'false'`

Comment (cebe) [inline framework/filters/AccessRule.php]: > or anyone can upload images anonymously ('?') but only the creator ip can delete them?

the if (isset(...)) ensures that it is evaluated when needed first, but then not evaluated multiple times.

**Comment by Faryshta** [comment]
@cebe sounds promising. that would drop the support for closures?

i am trying to think on a user case where two roles would require the same callback but that SHOULD be solved by creating a parent role in rbac

```php
'roles' => [
     'seller',
     'goldClient' => function ($rule) {
          return ['client_id' => Yii::$app->user->id];
     },
     'regularClient' => function ($rule) {
          return ['client_id' => Yii::$app->user->id];
     },
]

// should have both 'goldClient' and 'regularClient' having a parent role 'client'

'roles' => [
     'seller',
     'client' => function ($rule) {
          return ['client_id' => Yii::$app->user->id];
     },
]
```

framework/filters/AccessRule.php
Faryshta:
```php
            if ($role === '?') {
                if ($user->getIsGuest()) {
                    return true; // this line always return true even if the creator ip are not the same.
                }
            } elseif ($role === '@') {
                if (!$user->getIsGuest()) {
                    return true; // this line always return true even if the creator id are not the same.
                }
            } else {
                if (!isset($roleParams)) {
                    $roleParams = $this->roleParams instanceof Closure ? call_user_func($this->roleParams, $this) : $this->roleParams;
                }
                if ($user->can($role, $roleParams)) {
                    return true;
                }
            }
```
i checked the DbManager and seems it does the same

https://github.com/yiisoft/yii2/blob/master/framework/rbac/DbManager.php#L125

unauth users or general roles get rejected without using the `$params` property.

Faryshta: the problem seems to run deeper into the rbac implementation itself so after this gets merged I will create an issue about the use cases i mentioned.

dynasource (approved):

cebe:
Possible variants of specifying the params would be:

```php
'roles' => [
     '?',
     'seller',
     'goldClient' => ['client_id' => Yii::$app->user->id],
     'regularClient' => function ($rule) {
          return ['client_id' => Yii::$app->user->id];
     },
]
```

If you have the same param for different roles you can define a variable that hold the closure:
```php
$params = function ($rule) {
     return ['client_id' => Yii::$app->user->id];
};

// ...

'roles' => [
     '?',
     'seller',
     'goldClient' => $params,
     'regularClient' => $params,
]
```

samdark:
Looks good.

cebe:
Actually, thinking more about it I prefer the current implementation state of the PR. It is more descriptive, to say `'roleParams' => ... , 'roles' => ...` than to have an associative array. The use case to pass different parameters to different roles or permissions seems to be very rare and if needed can be solved by defining multiple access rules. in most cases the matching is one RBAC permission per AccessRule.

framework/filters/AccessRule.php
```diff
@@ -55,9 +56,38 @@ class AccessRule extends Component
...
+     * This can be used for example like this:
+     *
+     * ```php
+     * 'rules' => [
+     *     [
+     *         'allow' => true,
+     *         'actions' => ['update'],
+     *         'roles' => ['updatePost'],
+     *         'roleParams' => function($rule) {
+     *             return ['postId' => Yii::$app->request->get('id')];
```
fsateler: This example does not need a closure. Because the `rules` array is built in a method of the controller class, `Yii:: $app->request` should already be populated, so one could set the `roleParams` array directly. I think it would be nice to have an example for each case, with closure and without.

docs/guide/security-authorization.md
```diff
@@ -487,6 +491,37 @@ public function behaviors()
...
+    public function behaviors()
+    {
+        return [
+            'access' => [
+                'class' => AccessControl::className(),
+                'rules' => [
+                    [
+                        'actions' => ['update'],
+                        'roles' => ['updatePost'],
+                        'roleParams' => function() {
```
fsateler: It would be nice to have an example without closure too, to make it clear it is possible.

fsateler (commented):
This is a bit more verbose than my PR but possibly more intuitive. I see only a couple of doc improvements to make. Thanks!

cebe:
@fsateler I have adjusted docs according to your feedback. thanks!

fsateler (changes requested):
LGTM!

fsateler (approved):
Sorry clicked wrong radio box.

Issue #14134, opened 2017-05-10
Multiple validateAttribute() calls when scenarios() returns duplicate attributes


krukru:
Consider the following snippet
```php
    // some model class
    public function scenarios()
    {
        return [
            Model::SCENARIO_DEFAULT => [
                'foo',
                'foo',
            ],
        ];
    }

    public function rules()
    {
        return [
            ['foo', 'required'],
            ['foo', 'customFunction'],
            ['foo', 'filter', 'filter' => 'boolval'],
        ];
    }
```

Then when performing validation, all of the validation rules will be triggered twice, since the same attribute has been returned twice from ```scenarios()```. This does not have to be the default scenario, it could be your custom defined scenario.

Of course this is an obvious client mistake, you should not return foo twice, but when you have a more complex case, for example when merging with ```parent::scenarios()```, you could end up with the same mistake. 

I think that the framework should simply ignore attributes that have already been validated. If some attribute passed a validator once, it should obviously pass it twice as well and the second validation is not needed. Other then the minor performance waste, there is no real issue when the validator is idempotent.
However filters, custom functions and probably other things need not be idempotent, so calling them multiple times might produce an unwanted result.

One example is if you have a filter that converts a unix timestamp to a datetime string.

bizley:
This is the case indeed.

While this can be fixed in several places with simple `array_unique` I think the best place for this is [Model::activeAttributes()](https://github.com/yiisoft/yii2/blob/master/framework/base/Model.php#L776) because both [Model::validate()](https://github.com/yiisoft/yii2/blob/master/framework/base/Model.php#L337) and [Validator::validateAttributes()](https://github.com/yiisoft/yii2/blob/master/framework/validators/Validator.php#L247) should take duplicated attribute names as arguments if set directly and process them as many times as developer wants.

I can prepare PR.

krukru:
@bizley I agree fully, but could you let me have this one? :)

bizley:
Sure thing.

dynasource:
This usecase doesnt feel right. This problem should solved at the source instead of monkey patching afterwards in the core.

bizley:
This is the first place where you can correct the array (the `scenarios()` method can be overriden so fixing this in parent is useless). The only other options are:
- introduce new method (or any other process) that will fix the list (overkill IMHO),
- teach developers not to duplicate attributes in `scenarios()` :trollface:

dynasource:
[...]

PR #14139, opened 2017-05-10
Prevent multiple validateAttribute() calls when scenarios() returns duplicate attributes

krukru:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | probably no
| Tests pass?   | locally yes, travis fails on PHP7 and and PHP5.5 (probably a fluke)
| Fixed issues  | #14134


framework/base/Model.php
```diff
@@ -773,7 +773,7 @@ public function activeAttributes()
         if (!isset($scenarios[$scenario])) {
             return [];
         }
-        $attributes = $scenarios[$scenario];
+        $attributes = array_unique($scenarios[$scenario]);
         foreach ($attributes as $i => $attribute) {
```
Faryshta: since we are already on the subject, how about reducing this line like this? https://3v4l.org/1N2A6

bizley: Or like this already introduced in [2.0.12 Validator](https://github.com/yiisoft/yii2/blob/master/framework/validators/Validator.php#L461).

tests/data/base/Speaker.php
```diff
@@ -40,6 +40,7 @@ public function scenarios()
     {
         return [
             'test' => ['firstName', 'lastName', '!underscore_style'],
+            'duplicates' => ['firstName', 'firstName', '!underscore_style', '!underscore_style'],
```
cebe: why would you do this?

framework/base/Model.php
Faryshta:
```php
        return array_map(function($attribute) {
            return ltrim($attribute, '!');
        }, array_unique($scenarios[$scenario]));
```

or make new static method `Model::attributeNames(array $attributes)` and invoke it here and on validator

tests/data/base/Speaker.php
krukru: I wasn't really sure if by adding the duplicates to the test scenario I would break any existing tests, so I just added a new scenario. Do you think it's better just to add the attributes to the test scenario?

cebe: no, sorry I had not seen the issue before commenting here so I was wondering why you put duplicate names in the scenario.

krukru:
Tests were failing because the array needed to be reindexed. I used `array_values(array_unique($scenarios[$scenario]))`

`array_keys(array_count_values($scenarios[$scenario]))` is faster, although it's intent is maybe not as clear as in the first example. Since the `$scenarios[$scenario])` array is expected to be very small (<100 elements) I am going with readability in favor of a of few ns of performance

samdark (approved):

samdark:
@yiisoft/reviewers need more reviews.

dmirogin (approved):

framework/base/Model.php
```diff
@@ -781,7 +781,7 @@ public function activeAttributes()
         if (!isset($scenarios[$scenario])) {
             return [];
         }
-        $attributes = $scenarios[$scenario];
+        $attributes = array_values(array_unique($scenarios[$scenario]));
```
rob006: `array_unique()` will change order of attributes. This may affect validation and create some unexpected results when order of attributes is important. 

Maybe it is better to use `array_keys(array_flip($scenarios[$scenario]))` instead? It should be faster and preserves order of attributes.

samdark: Indeed. Order is very important.

rob006 (approved):

samdark:
Merged. Thanks!

framework/base/Model.php
krukru: You're right. Not only does `array_unique()` change the order, it is also very slow because of that (it sorts all the remaining items). Guess PHP developers didn't hear about single responsibility. 
On a side note, what do you think about refactoring this line into a method? The intent would be much clearer to read if it said something like `$this->getUniqueScenarioAttributes()`. Maybe even add something to `yii\helpers\ArrayHelper` since php does not have an efficient or easy to understand way to do this operation. Trying to understand what `array_flip(array_flip($array))` does is so much more harder to understand then something like `ArrayHelper::uniqueValues($array)`

rob006: Personally I'm for adding `ArrayHelper::unique()`, but it was already rejected: #10378

Issue #13940, opened 2017-04-07
Forced client validation doesn't show validation error messages


glagola:
### What steps will reproduce the problem?
$('#contact-form').yiiActiveForm('validate', true);
### What is the expected result?
I expected to see validation error messages. According to documentation last parameter in the function is forceValidation.
### What do you get instead?
I saw no errors.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.1.3
| Operating system | Linux

SilverFire:
Resolved by 8f2d9ba

PR #14141, opened 2017-05-11
Fixed yii.activeForm.js

depesr:
In function validate, when force validation is set to true, it actually set's data.submitting to true.
Few rows later, variable submitting was set to data.submitting && !forceValidate, which will always be false in case forceValidation is true.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #13940


SilverFire:
Resolved by 8f2d9ba

Issue #14133, opened 2017-05-10
Bug in calculateTimings


bashkarev:
### What steps will reproduce the problem?
```php
$data = [
    [
        'INSERT INTO `csv` FORMAT CSV `e1e747f9901e67ca121768b36921fbae.csv`',
        80,
        'bashkarev\\clickhouse\\Command::query',
        1494429676.541369,
        [],
        4818424,
    ],
    [
        'INSERT INTO `csv` FORMAT CSV `ebe191dfc36d73aece91e92007d24e3e.csv`',
        80,
        'bashkarev\\clickhouse\\Command::query',
        1494429676.542989,
        [],
        4856304,
    ],
    [
        'INSERT INTO `csv` FORMAT CSV `e1e747f9901e67ca121768b36921fbae.csv`',
        96,
        'bashkarev\\clickhouse\\Command::query',
        1494429676.676885,
        [],
        4883624,
    ],
    [
        'INSERT INTO `csv` FORMAT CSV `ebe191dfc36d73aece91e92007d24e3e.csv`',
        96,
        'bashkarev\\clickhouse\\Command::query',
        1494429676.74703,
        [],
        4869568,
    ],
];
Yii::getLogger()->calculateTimings($data);
```
### What is the expected result?
calculated data

### What do you get instead?
empty data

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12-dev
| PHP version      | 7.1.4
| Operating system | macOS Sierra 10.12.4 (16E195)

bashkarev:
@bizley Thank you so much!

PR #14148, opened 2017-05-12
14133 fix for calculating timings with nested mixed profile begin and end

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14133


samdark:
Merged. Thank you!

Issue #14157, opened 2017-05-14
Incorrect Value When Using loadDefaultValues with DATETIME Field


Knight-Yoshi:
I have a MySQL table that looks like this
<img width="1059" alt="screenshot 2017-05-14 18 06 34" src="https://cloud.githubusercontent.com/assets/6134576/26038185/60f0da46-38d0-11e7-9edc-d91dbef5f4f6.png">

When saving after calling the AR's `loadDefaultValue()` method it a DB exception is thrown.

```
SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect datetime value: 'CURRENT_TIMESTAMP' for column 'created_on' at row 1
The SQL being executed was: INSERT INTO `image_set` (`created_on`, `last_updated_on`, `total`, `title`, `user_id`, `slug`) VALUES ('CURRENT_TIMESTAMP', CURRENT_TIMESTAMP, 1, 'Title', 1, 'title')
```

In this case I can avoid loading default values. However, I have other tables that will require loading the default values that also have DATETIME fields.

rossoneri:
@Knight-Yoshi It seems the default value of `created_on` is  a string value 'CURRENT_TIMESTAMP', is it true?

PR #14159, opened 2017-05-15
#14157 Fix Incorrect Value When Using loadDefaultValues with DATETIME…

qihuajun:
… Field

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14157


framework/db/mysql/Schema.php
```diff
@@ -174,7 +174,7 @@ protected function loadColumnSchema($info)
         $column->phpType = $this->getColumnPhpType($column);
 
         if (!$column->isPrimaryKey) {
-            if ($column->type === 'timestamp' && $info['default'] === 'CURRENT_TIMESTAMP') {
+            if (($column->type === 'timestamp' || $column->type ==='datetime') && $info['default'] === 'CURRENT_TIMESTAMP') {
```
cebe: is this valid in MySQL? afair CURRENT_TIMESTAMP only works with timestamp columns. My knowledge here could be outdated but we should verify that.

rob006: It is supported since MySQL 5.6

https://dev.mysql.com/doc/refman/5.6/en/timestamp-initialization.html#idm139764692637360

rossoneri: The DEFAULT CURRENT_TIMESTAMP support for a DATETIME (datatype) was added in MySQL 5.6. See details here: https://dev.mysql.com/doc/refman/5.6/en/timestamp-initialization.html .

And notice that the code both check column type and the default value, so even in the older version mysql, though column type is `datetiem`, but the second condition can't be true, so the code is compatible.

I have a question about unit test, base on the test database generated from `tests/data/mysql.sql`, I can't cover the case of this situation. How can I write unit test using the test database?

rossoneri:
@cebe I add a unit test function, but the code is a little complicated, do you think is ok?

samdark:
Merged. Thank you!

Issue #14150, opened 2017-05-12
error on exists validator after d811f1e


Bhoft:
Hi i have just updated to the latest version d811f1e
and now my exists validator fails with an sql error.

I have this rule in my Submission Class.

     [['submission_call_id'], 'exist', 'skipOnError' => true, 'targetClass' => SubmissionCall::className(), 'targetAttribute' => ['submission_call_id' => 'id']],

The SubmissionCall Class have this tablename

    public static function tableName()
    {
	return '{{%submission_call}}';
    }

The generated sql is 

    SELECT EXISTS(SELECT * FROM `submission_call` WHERE (`0`.`id`='12'))`

Of course this sql is wrong, but why?
Are there still some changes to the exists validator missing?
Is this a newly introduced bug due to the latest change?

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | d811f1e
| PHP version      | 5.6
| Operating system | Windows

cebe:
> Is this a newly introduced bug due to the latest change?

quite sure it is.

bscheshirwork:
rules like this
```php
            [['unitId'], 'exist',
                'skipOnError' => true,
                'targetClass' => DurationUnit::className(),
                'targetAttribute' => ['unitId' => DurationUnit::tableName() . '.id'],
                'filter' => function (DurationUnitQuery $query) {
                    $query->active();
                }
            ],
```
produced error now :rage: :anger: :fire: :fire_engine: 
```SQL
SELECT EXISTS(SELECT * FROM `duration_unit` WHERE (`duration_unit`.[[`duration_unit`.id]]=1) AND (`duration_unit`.`stateOfFlags`=1))'
```
:smile_cat:

bscheshirwork:
Can create "error on exists validator after c6ab3f8" :1st_place_medal:

samdark:
Ohh... the case for manual prefixing wasn't taken into account :(

bscheshirwork:
https://github.com/yiisoft/yii2/blob/master/framework/validators/ExistValidator.php#L226
https://github.com/yiisoft/yii2/blob/master/framework/validators/UniqueValidator.php#L179
https://github.com/yiisoft/yii2/blob/master/framework/validators/UniqueValidator.php#L299

I think need function named like `applyPrefix($column, $prefix = null)` for this case

bscheshirwork:
#14160

bscheshirwork:
#14161

[...]

PR #14161, opened 2017-05-15
Apply primary table alias; include case manual set alias

bscheshirwork:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Fixed issues  | #14150


framework/db/ActiveQuery.php
```diff
@@ -790,6 +790,29 @@ public function alias($alias)
...
     }
 
+
+    /**
+     * Returns conditions with alias
+     * @param array $conditions array of condition, keys to be modified
+     * @param null|string $alias set empty string for no apply alias. Set null for apply primary table alias
+     * @return array
+     */
+    public function applyTableAlias($conditions, $alias = null)
```
samdark: I don't think it's a good idea to have it as a public method. I'd rather move it to private methods in unique and exist validators even if it duplicates code a bit.

bscheshirwork: But.. Its functional of ActiveQuery model, not Validator. 

Also... Some case we need the alias for build a query with alias inside child of ActiveQuery. 

May be something like this method.

PR #14163, opened 2017-05-15
14150 apply primarytable alias (include case manual set alias) in validators

bscheshirwork:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Fixed issues  | #14161, #14150


cebe:
what is the difference to #14161? you do not need to open a new PR to make changes to an existing one, you can just push new commits to the same branch.

samdark:
Seems to fail: https://travis-ci.org/yiisoft/yii2/jobs/232464830

bscheshirwork:
> what is the difference to #14161?

@samdark is vote for private method in validators.

But i think we need discuss about method to automated assign alias to condition (i.e. in multiply joinWith)

framework/validators/ExistValidator.php
```diff
@@ -222,7 +222,7 @@ private function applyTableAlias($query, $conditions, $alias = null)
         $prefixedConditions = [];
         foreach ($conditions as $columnName => $columnValue) {
             $prefixedColumn = "{$alias}.[[" . preg_replace(
-                    '/^' . $alias . '\.(.*)$/',
+                    '/^' . preg_quote($alias) . '\.(.*)$/',
                     "$1",
```
bscheshirwork: `\ + * ? [ ^ ] $ ( ) { } = ! < > | : -`
Ok, the alias may consist `-`

samdark:
Merged. Thank you!

Issue #14165, opened 2017-05-16
Slave usage disabled after Connection::close


juffin-halli:
If you close connection to DB, and then reopen it, slaves will not be used again.

Connection::close set Connection::_slave to null instead of false.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11

PR #14168, opened 2017-05-17
Bug #14165: Fix slave usage disabled after Connection::close, set _sl…

qihuajun:
…ave to false instead of nulll

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14165

### DIFF
--- a/framework/db/Connection.php
+++ b/framework/db/Connection.php
@@ -608,7 +608,7 @@ class Connection extends Component
 
         if ($this->_slave) {
             $this->_slave->close();
-            $this->_slave = null;
+            $this->_slave = false;
         }
     }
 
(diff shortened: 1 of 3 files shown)


framework/db/Connection.php
```diff
@@ -608,7 +608,7 @@ public function close()
 
         if ($this->_slave) {
             $this->_slave->close();
-            $this->_slave = null;
+            $this->_slave = false;
```
cebe: same issue with `$this->_master` above in line 599.

rossoneri: @cebe Changed line 599 and unit test added! 😁

cebe:
Merged, thank you!

Issue #14166, opened 2017-05-16
ActiveQuery and Aliases


bscheshirwork:
1 The alias **must** previous to all assoc columns if Alias is set; 
2 In the joinWith the alias **must** be set too (default is a key of array)

### What steps will reproduce the problem?
```php
class Duration extends \yii\db\ActiveRecord {
...
    public function getUnit()
    {
        return $this->hasOne(DurationUnit::className(), ['id' => 'unitId'])->inverseOf('durations');
    }
...
}
```
```php
        echo Duration::find()
            ->joinWith([
                'unit as unitAlias' => function (DurationUnitQuery $query) {
                    $query->andWhere(['id' => [5, 3, 2, 1]]);
                }
            ])
            ->andWhere(['id' => [4, 5, 9]])
            ->alias('durationAlias')
            ->createCommand()->getRawSql();
```
### What is the expected result?
```SQL
SELECT `durationAlias`.* FROM `duration` `durationAlias` LEFT JOIN `duration_unit` `unitAlias` ON `durationAlias`.`unitId` = `unitAlias`.`id` WHERE (`durationAlias`.`id` IN (4, 5, 9)) AND (`unitAlias`.`id` IN (5, 3, 2, 1))
```
### What do you get instead?
```SQL
SELECT `durationAlias`.* FROM `duration` `durationAlias` LEFT JOIN `duration_unit` `unitAlias` ON `durationAlias`.`unitId` = `unitAlias`.`id` WHERE (`id` IN (4, 5, 9)) AND (`id` IN (5, 3, 2, 1))

```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 44fa97f1d66ecc8c8851cc3787f4432f1e8eb0ad
| PHP version      | 7.15
| Operating system |Debian/Alpine/other

bscheshirwork:
`Where` `condition` must be processed with this function;

https://github.com/bscheshirwork/yii2/blob/184f6d385bd1a7eb3e5216774a3a9783978b3760/framework/db/ActiveQuery.php#L800-L814

cebe:
duplicate of #7263

PR #14172, opened 2017-05-17
Apply child table alias/name for joinWith case to condition columns (with test)

bscheshirwork:
One step to resolve the problem with ambigous column name and apply alias.

As part of #7263:

For case `joinWith` functional: only one child table inside `joinWithRelation` for each time
i.e. not allowed 
```SQL
SELECT * FROM
a JOIN b,c 
WHERE a.id = b.a_id and b.id = c.b_id
```
i.e. ActiveQuery for the child table is have single `from` and can be use single prefix.

Note:
for #14161
This method can be private or 
`framework/validators/ExistValidator.php`
`framework/validators/UniqueValidator.php`
can use this method too

Example: after path
```php
class Duration extends \yii\db\ActiveRecord {
...
    public function getUnit()
    {
        return $this->hasOne(DurationUnit::className(), ['id' => 'unitId'])->inverseOf('durations');
    }
...
}
```
```php
        echo Duration::find()
            ->joinWith([
                'unit as unitAlias' => function (DurationUnitQuery $query) {
                    $query->andWhere(['id' => [5, 3, 2, 1]]);
                }
            ])
            ->andWhere(['id' => [4, 5, 9]])
            ->alias('durationAlias')
            ->createCommand()->getRawSql();
```
produce
```SQL
SELECT `durationAlias`.* FROM `duration` `durationAlias` LEFT JOIN `duration_unit` `unitAlias` ON `durationAlias`.`unitId` = `unitAlias`.`id` WHERE (`id` IN (4, 5, 9)) AND (`unitAlias`.`id` IN (5, 3, 2, 1))
```
current 
```SQL
SELECT `durationAlias`.* FROM `duration` `durationAlias` LEFT JOIN `duration_unit` `unitAlias` ON `durationAlias`.`unitId` = `unitAlias`.`id` WHERE (`id` IN (4, 5, 9)) AND (`id` IN (5, 3, 2, 1))
```

and "as target" result must be
```SQL
SELECT `durationAlias`.* FROM `duration` `durationAlias` LEFT JOIN `duration_unit` `unitAlias` ON `durationAlias`.`unitId` = `unitAlias`.`id` WHERE (`durationAlias`.`id` IN (4, 5, 9)) AND (`unitAlias`.`id` IN (5, 3, 2, 1))
```

include some tests

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | yes/no
| Tests pass?   | yes/no
| Fixed issues  | #7263, #14166, #14161, #14150


samdark:
@bscheshirwork seems tests aren't passing in this one.

bscheshirwork:
@samdark this PR is use another concept (my first proposition) of `applyTableAlias` and must be changed

samdark:
@bscheshirwork please close/fix ones that aren't actual. Hard to review all these...

bscheshirwork:
I will close it. I don't have a time for fix this now.

PR #14193, opened 2017-05-23
GridView-columns-special-value-to-disable-colunm

bscheshirwork:
For a case non-simple-gii-generated GridView...

Add a simple logic to turn off the column:
```php
    <?= GridView::widget([
        'filterModel' => $searchModel,
        'columns' => [
            ['class' => 'yii\grid\SerialColumn'],
'~5 first like this' => [
                'format' => 'raw',
                'attribute' => 'attributeName',
                'label' => $searchModel->getAttributeLabel('searchAttributeName'),
                'content' => callbackToAttributeName(),
                'filter' => filterToAttributeName(),
            ],

'underDiscussion' => \Yii::$app->user->can('viewUnderDiscussionColumn') ? [
                'format' => 'raw',
                'attribute' => 'attributeName',
                'label' => $searchModel->getAttributeLabel('searchAttributeName'),
                'content' => callbackToAttributeName(),
                'filter' => filterToAttributeName(),
            ] : false,

'~5 last like this' => [
                'format' => 'raw',
                'attribute' => 'attributeName',
                'label' => $searchModel->getAttributeLabel('searchAttributeName'),
                'content' => callbackToAttributeName(),
                'filter' => filterToAttributeName(),
            ]',
        ],
    ]); ?>
```

p.s.: `visible` is request to create the column before disable it. Its terrible.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no


framework/grid/GridView.php
```diff
@@ -525,6 +525,10 @@ protected function initColumns()
             $this->guessColumns();
         }
         foreach ($this->columns as $i => $column) {
+            if ($column === false){
```
samdark: It should be reflected in phpdoc.

bscheshirwork:
assert count($grid->columns) is equal 2 for array with false value

lynicidn:
это не баг, что мешает вместо false делать так 
```php

'underDiscussion' =>  [
                'format' => 'raw',
                'attribute' => 'attributeName',
                'label' => $searchModel->getAttributeLabel('searchAttributeName'),
                'content' => callbackToAttributeName(),
                'filter' => filterToAttributeName(),
'visible' => \Yii::$app->user->can('viewUnderDiscussionColumn'),
            ],

```
in this code look visible - > 'visible' => \Yii::$app->user->can('viewUnderDiscussionColumn'),

bscheshirwork:
@lynicidn, посмотрите код и "p.s." первого сообщения.

cebe:
> p.s.: visible is request to create the column before disable it. Its terrible.

what exactly is wrong with creating the column first?

bscheshirwork:
@cebe 
> what exactly is wrong with creating the column?
fact of creation.

If column already disabled.

bscheshirwork:
i.e. don't need call 

$searchModel->getAttributeLabel('searchAttributeName'),
...
callbackToAttributeName(),
filterToAttributeName(),
...
etc.

cebe:
The simple instantiation of the Column class does not execute any of these as far as I see... Note that `visible` is checked a few lines below your code.

bscheshirwork:
What? 
I can see:
```php
            if (is_string($column)) {
                $column = $this->createDataColumn($column);
            } else {
                $column = Yii::createObject(array_merge([
                    'class' => $this->dataColumnClass ?: DataColumn::className(),
                    'grid' => $this,
                ], $column));
            }
            if (!$column->visible) {
                unset($this->columns[$i]);
                continue;
            }
```


Test of $visible is end of checks list, 
https://github.com/yiisoft/yii2/blob/master/framework/grid/GridView.php#L536-L539

all of previous is called:
https://github.com/yiisoft/yii2/blob/master/framework/grid/GridView.php#L528-L535

 `Yii::createObject` is called

bscheshirwork:
Use `$column->visible` make sense if `visible` depends on the column directly.

I propose another case: 
visible has no dependence of config structure or column constructor and has odepends of GridView or of any other expression

bscheshirwork:
So... May be I say wrong thing... About applying config in `Yii::configure` (I think `'label' => $searchModel->getAttributeLabel('searchAttributeName')` and property like `filter` called in this moment)
But autoloader / DI for any touch class is called...

samdark (approved):

cebe (approved):
still not completely convinced that its needed, but fine with me. CHANGELOG is missing.

dynasource:
having NULL in place in stead of false would make more sense to me

cebe:
> having NULL in place in stead of false would make more sense to me

we use `false` for disabling things in a lot of places so its fine for me.

PowerGamer1:
I am strongly against this functionality. There are already two approaches to solving the problem raised by OP.

1. Use `visible`. OP claims it is less effective than what he suggests because when using `visible` the function `$searchModel->getAttributeLabel('searchAttributeName'),` will be called unnecessarily. OK, but in HIS proposed solution he adds an extra `if(column == false)` check into the loop processing the column definitions - which would have a (probably even worse than calling `getAttributeLabel`) performance impact for all current users of `GridView` that do not require this new functionality.

2. Add columns one by one into array:
```PHP
$columns = [];
$columns[] = [/*definition of col1*/]
if(condition) $columns[] = [/*definition of colX*/]
$columns[] = [/*definition of colN*/]
```

My main issue with the new functionality is the fact that the array of columns would contain something that is NOT a column in any way - value `false`. This is just bad design.

rob006:
You can achieve the same by wrapping columns config with `array_filter()`.

bscheshirwork:
> My main issue with the new functionality is the fact that the array of columns would contain something that is NOT a column in any way - value false.

Example, pls?

PowerGamer1:
> Example, pls? Value "false" - is "not-a-column" marker.

That's exactly the problem - array of _columns_ containing a _"not-a-column"_ something.

bscheshirwork:
> array of columns containing a "not-a-column" something.

as mark to skip it/itself. "special value to disable column". skip-me, do-not-process, i-am-not-a-column-not-look-at-me = 
Sugar for avoid your 2s proposition "step-by-step add" elements

Of course it is equivalent use `array_filter()` without callback. Thanks @rob006 

```php
    <?= GridView::widget([
        'filterModel' => $searchModel,
        'columns' => array_filter([
            ['class' => 'yii\grid\SerialColumn'],
'~5 first like this' => [
                'format' => 'raw',
                'attribute' => 'attributeName',
                'label' => $searchModel->getAttributeLabel('searchAttributeName'),
                'content' => callbackToAttributeName(),
                'filter' => filterToAttributeName(),
            ],

'underDiscussion' => \Yii::$app->user->can('viewUnderDiscussionColumn') ? [
                'format' => 'raw',
                'attribute' => 'attributeName',
                'label' => $searchModel->getAttributeLabel('searchAttributeName'),
                'content' => callbackToAttributeName(),
                'filter' => filterToAttributeName(),
            ] : false,

'~5 last like this' => [
                'format' => 'raw',
                'attribute' => 'attributeName',
                'label' => $searchModel->getAttributeLabel('searchAttributeName'),
                'content' => callbackToAttributeName(),
                'filter' => filterToAttributeName(),
            ]',
        ]),
    ]); ?>
```

So... Any of configuration can use `false` for disable items. GridView is ... peculiar?

samdark:
Since `array_filter()` is looking good, I think it's a bad idea to merge it. Thanks for proposal though and thank you @rob006 for a good solution.

Issue #8420, opened 2015-05-14
MySql, $query->batch() from more millions of record


AstRonin:
I tried run `yii\db\Query::batch()` for big table and got memory leak. This happens because method  \yii\db\Command::queryInternal() run 

``` php
$this->pdoStatement->execute();
```

I changed attribute `PDO::MYSQL_ATTR_USE_BUFFERED_QUERY` -> `false` and it helps me.
I think, we should add description about this feature into documentation.

yujin1st:
Could you specify where did you set the attribute? 
I tried it in config and runtime(Yii::$app->db->pdo->setAttribute(PDO::MYSQL_ATTR_USE_BUFFERED_QUERY,false)), and got error 2014.

AstRonin:
As I understand this problem, you did not use 'fetch' or 'execute' method after creation request.
Also you can try the next solutions before I will see my code in the evening:

* `$command->reset()`
* read [issues 3695](https://github.com/yiisoft/yii2/issues/3695)

Nevertheless, you can attach code and error stack to better understand your problem.

Other way, you can use `setAttribute(PDO::MYSQL_ATTR_USE_BUFFERED_QUERY,false)` for mehod `batch()` and after that return back attribute to `true` for other cases

AstRonin:
Unfortunately, I cannot find my codes with this attribute, maybe it was a previous project. But you may use advice indicated above or show your code and we will try to  help.

mikehaertl:
I also have this problem and the above fix didn't work for me. Any other suggestions?

Yii 2.0.6 PHP 5.6.10 MySQL 5.6.22

cebe:
unable to reproduce the issue, see #11152

cebe:
https://github.com/yiisoft/yii2/issues/9269#issuecomment-250739252

ihipop:
Any Update on this issue?

SamMousa:
I can reproduce the issue, working on a fix.
What is the opinion of the yii team concerning configurability?
One could argue that it never makes sense to use batch in combination with buffered mode and that the default mode should be unbuffered.
Alternatively we could make the default unchanged and just add a parameter that allows explicitly disabling buffered mode (meaning it won't break BC).


Currently testing with unbuffered mode, there are some issues:
1. We need to guarantee the database schema is known (otherwise the AR layer will try to do a `SHOW FULL COLUMNS FROM xx`).
2. Any logic inside the batch processing cannot run queries.

[...]

PR #14215, opened 2017-05-26
Add doc for issue #8420

ihipop:
https://github.com/yiisoft/yii2/issues/8420#issuecomment-303956282

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #8420

### DIFF
--- a/docs/guide/db-query-builder.md
+++ b/docs/guide/db-query-builder.md
@@ -657,7 +657,7 @@ When working with large amounts of data, methods such as [[yii\db\Query::all()]]
 > Warning: Unless whole data set has been retrieved, no other query could be done through the same connection. The table may remains locked by MySQL and cannot be written to by other queries.
 
 Yii provides the so-called batch query support. A batch query makes use of the data cursor and fetches data in batches, but the `$db` parameter is not the  one that have `PDO::MYSQL_ATTR_USE_BUFFERED_QUERY`
-disabled before `version 2.0.11.2` by default, [and maybe won't ever](https://github.com/yiisoft/yii2/issues/8420) (still under discussion for now), because there is no [`Silver Bullet` ](https://github.com/yiisoft/yii2/issues/8420#issuecomment-295679232) for this.
+disabled before `version 2.0.11.2` by default, [and maybe won't ever](https://github.com/yiisoft/yii2/issues/8420) (still under discussion for now), because there is  [`No Silver Bullet`](https://github.com/yiisoft/yii2/issues/8420#issuecomment-295679232) for this.
 Before PHP 5.6, the memory usage of buffered query in drivers does not count towards PHP's memory limit, but it takes the memory actually,  it may  eat all your servers' memory and get your PHP-FPM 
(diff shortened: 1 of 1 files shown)


samdark:
@ihipop, @tom--, @beowulfenator please check docs this pull request.

beowulfenator:
How do I add something to this PR?

samdark:
You can't since you don't have write access to this repo. Either comment inline or, alternatively, access could be given.

ihipop:
@samdark Thank you for your great work. 
But I have some different advice.

> The table may remain locked by MySQL and cannot be written to by other queries. Thus, new connection should be created for this purpose.

This is not the purpose that we establish  a new connection to MySQL server, we establish   a new connection to MySQL server because :

> Unless whole data set has been retrieved, no other query could be done through the same connection

Even we establish  a new connection to MySQL server ,tables may still `remain locked by MySQL and cannot be written to by other queries`  ，[Because MySQL keeps locks on data in the case of MyISAM on the entire table](https://stackoverflow.com/questions/5753948/php-mysqli-use-result-store-result-and-myisam-table-locks/5782924) ,and In `Unbuffered Query` Mode, because the loop  done by PHP is much more slow than `Buffered Query`  done by PDO (Not only because PHP is slower than C, but also the developer usually do some business logic in the loop for their specify reason), the lock time to table will much more longer  than `Buffered Query` , these two reason are too different.

So, The `lock table statement` should be moving to the `warning part` as I've commited.

`Unbuffered Query` is not a silver built indeed, just as what @SamMousa [have said](https://github.com/yiisoft/yii2/issues/8420#issuecomment-296109257):

> As I've said before, no solution is valid for all cases and therefore we should not just implement something. Instead implement it yourself as a custom function or behavior in your query class.

Maybe the total solution should be done by the developer in their business code, for us , a well document is on call.


@beowulfenator

ihipop:
And, also, I think that [the paragraph  about the memory strategy of PHP before 5.6](https://github.com/yiisoft/yii2/pull/14215/commits/c1e6e6da62c6b2dca8b762f2af06bc9e95be1e8f#diff-356211f47ae64830fff2dadc9d65d7acR661) should not be removed
It explains that why `batch()` and `each()`  _seems to work_ (will not hit the memory limit problem,But will eat the server's memory at all) and lead a  meditate to `Bunffered query`

beowulfenator:
Can you guys give me access to the repo?

ihipop:
@beowulfenator https://github.com/ihipop/yii2/invitations invited

beowulfenator:
Does this look better?

samdark (approved):

samdark:
Looks great to me now.

samdark:
@ihipop how do you find it now?

ihipop:
@samdark did not contains[ the opinion I‘ve said](https://github.com/yiisoft/yii2/pull/14215#issuecomment-304768532)
I still want add these, can I have your  agreement ?

ihipop:
@beowulfenator excellent work.
 I would like to tweak a few words

samdark:
Merged. Thank you!

tom--:
This is good. Thanks

PR #14218, opened 2017-05-27
move $params into parameter list on insert()

the-liquid-metal:
in yii\db\Command::insert(), the  `$params = [];`  statement in method body seems to be useless. so it would be useful if it be placed on parameter.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no

### DIFF
--- a/framework/db/Command.php
+++ b/framework/db/Command.php
@@ -424,11 +424,11 @@ class Command extends Component
      * @param array|\yii\db\Query $columns the column data (name => value) to be inserted into the table or instance
      * of [[yii\db\Query|Query]] to perform INSERT INTO ... SELECT SQL statement.
      * Passing of [[yii\db\Query|Query]] is available since version 2.0.11.
+     * @param array $params the parameters to be bound to the command
      * @return $this the command object itself
      */
-    public function insert($table, $columns)
+    public function insert($table, $columns, $params = [])
     {
-        $params = [];
         $sql = $this->db->getQueryBuilder()->insert($table, $columns, $params);
 
         return $this->setSql($sql)->bindValues($params);


SOHELAHMED7 (approved):

SOHELAHMED7 (approved):

rob006:
> in yii\db\Command::insert(), the $params = []; statement in method body seems to be useless.

No, it's not. Just check `QueryBuilder::insert()` docs: http://www.yiiframework.com/doc-2.0/yii-db-querybuilder.html#insert()-detail

cebe:
params are passed to `QueryBuilder::insert()` by reference so after the call it will not be empty anymore. Its all correct here.

the-liquid-metal:
Please compare:
`yii\db\Command::insert()`
with:
`yii\db\Command::update()` and
`yii\db\Command::delete()`

In `update()` and `delete()`, the `$params` is present at parameter list, not the body.

As a comparison, please refer to [Executing Non-SELECT Queries](http://www.yiiframework.com/doc-2.0/guide-db-dao.html#non-select-queries), at 2nd example. With that code, we can execute single query with only 1 method call.
`Yii::$app->db->createCommand()->`**`insert(...)`** or
`Yii::$app->db->createCommand()->`**`update(...)`** or
`Yii::$app->db->createCommand()->`**`delete(...)`**

Unfortunatly that code is for safety data. If we work with unsafety data, we have to use the last param of `update()` or `delete()`, which is `$params = []`.

What about `insert()`?
There is no `$params = []` in parameter list, so we can not write single method call in order to work with unsafety data.

If `insert()` is revised, we will get consistent API.
Example code if we work in unsafety data, (and of course if `insert()` is revised):

```php
// INSERT (table name, column values, parameters)
Yii::$app->db->createCommand()->insert('user', [
    'name' => ':name',
    'age' => ':age',
], [
    ':name' => $_POST['name'],
    ':age' => $_POST['age'],
])->execute();

// UPDATE (table name, column values, condition, parameters)
Yii::$app->db->createCommand()->update(
    'user',
    ['status' => ':status'],
    'age > 30',
    [':status' => $_POST['status']]    // this is the part which i propose to be
)->execute();

// DELETE (table name, condition, parameters)
Yii::$app->db->createCommand()->delete(
    'user',
    'status = :status'
    [':status' => $_POST['status']]
)->execute();
```

cebe:
@the-liquid-metal you only need $params if you write SQL as a string, this is possible in delete() and update() because you specify a condition. On insert you only pass a PHP array, so there is no need to use param binding yourself, yii will do that internally.

the-liquid-metal:
@cebe is it safe to write:
```php
// INSERT (table name, column values, parameters)
Yii::$app->db->createCommand()->insert('user', [
    'name' => $_POST['name'],
    'age' => $_POST['age'],
])->execute();
```
even though $_POST potentially contains sql injection?

cebe:
yes, it's mentioned in the documentation here: https://github.com/yiisoft/yii2/blob/3dc8f426d01ecbf23d2f1faade292bbe7762d999/framework/db/Command.php#L419

I updated the guide to be more clear about this: a3501ed

the-liquid-metal:
I think it is necessary to provide example code for `update()` and `delete()` on guide in order to use `$param` and handling unsafety data.

cebe:
> I think it is necessary to provide example code for update() and delete() on guide in order to use $param and handling unsafety data.

here you are: fddb34b

the-liquid-metal:
Great! thank you.

PR #14219, opened 2017-05-27
Nested fields and expand

bboure:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #6844 

This is my approach of @mdmunir's https://github.com/yiisoft/yii2/pull/11493
I avoided touching ArrayHelper's toArray() which in my opinion should not be involved into this.

Also, I introduced the concept of selecting nested fields as well and introduced the '*' wildcard for selecting all fields.

You can now do: `user/1?fields=*,orders.*,orders.items.name&expand=orders.items`

Which means: 
- get user 1 and all it's fields.
- expand orders and get all fields from orders
- expand orders' items but only get the name.

Note that `*,orders.*` is required to explicitly get all fields from user and orders.
`order.items.name` alone means "only get order.items.name" fields.


Faryshta:
how about an unified syntax for both?

hal draft 8 introduces an optional property `_embebed` https://tools.ietf.org/html/draft-kelly-json-hal-08#section-4.1.2 this was written after yii 2.0.0 release so it was never considered for that release but it essentially the same as `expand`

the url could look like this

`user/1?embed=*,orders.*,orders.items.name`

it would mean


- get user 1 and all it's fields.
- embed orders and get all fields from orders
- embed order's items but only get the name.

it would look like this

```json
{
  "id": 1,
  "name": "admin",
  // all the other fields
  "_embeded": {
    "orders": [
       {
         "id": 123,
         "user_id": 1, // redundant but still a field
         "amount": 500,
         // all the other fields
        "_embeded": {
          "items": [
            {
              "name": "apple"
            },
            {
              "name": "banana"
            }
          ],
        }
      },
      {
         "id": 124,
         "user_id": 1, // redundant but still a field
         "amount": 1000,
         // all the other fields
        "_embeded": {
          "items": [
            {
              "name": "peach"
            },
            {
              "name": "banana"
            }
          ],
        }
      }
    ]
  }
}
```

bboure:
@Faryshta I see. But that would be a breaking change anyway. We could consider to make it optional by adding a `$useEmbedded` attribute to the [Serializer](https://github.com/yiisoft/yii2/blob/master/framework/rest/Serializer.php)

What do you think?

Faryshta:
i see it as adding three parameters.

`bool $useEmbed = false`

`string $embedParam = 'embed'`

`srtring $embedEnvelope = '_embeded'`

but that would also change some methods such as `$model->toArray($fields, $expand);` to `$model->toEmbededArray($embed);`

bboure:
@Faryshta I don't think replacing `fields` and `expand` with just `embed` is a good idea.
For me, these are 2 different concepts:
- `fields` fetches the fields from a Model
- `expand` determines which relations should be included. 

A another additional big argument for keeping them separated is that we could easily use the `fields` param in [ActiveQuery::select()](http://www.yiiframework.com/doc-2.0/yii-db-query.html#select()-detail) and `expand` to eagerload the relations using [ActiveQuery::with()](http://www.yiiframework.com/doc-2.0/yii-db-activequerytrait.html#with()-detail)

I was actually thinking of making another PR changing [ViewAction](http://www.yiiframework.com/doc-2.0/yii-rest-viewaction.html) and [IndexAction](http://www.yiiframework.com/doc-2.0/yii-rest-indexaction.html) for that.

tunecino:
@bboure shouldn't be a way to control what could be nested? Not sure how it works exactly but if I don't properly override the `fields` method in related models and tie on default implementation, won't this URI to user resource as example help me retrieve what I'm not supposed to access?

```
user/1?fields=orders.items.bestBuyer.userProfile.address
```

I mean shouldn't we better pre-define what could be nested within `fields()` and `extraFields()` methods instead of nesting them all and having to manually unset what shouldn't be there?

bboure:
@tunecino 
> won't this URI to user resource as example help me retrieve what I'm not supposed to access?

I did not invent anything here. Currently, "expanded" relations will be shown with all fields that are returned by `fields()`. If you want to hide some fields you MUST override `fields()`. But that did not change.
Here I am just adding a way to choose the wanted nested fields, but you won't be able to retrieve fields that are not returned by `fields()`.
`resolveFields()` still does that filtering.

tunecino:
I like this feature to be added @bboure. Truth is, I can't agree or argue anything as I'm not familiar with how `fields()` and `resolveFields()` are working under the hood. I'm just turning your attention to what looked to me like a possible security issue or BC for already working api's so we may have to figure out a different way to control what to nest. If it is safe then what you are proposing is great ;)

michaelarnauts:
This actually makes toArray() on a model work as expected, and fixes the big issue we had with `toArray()` on objects that have filled in relations.

```
     * @param array $expand the additional fields being requested for exporting. Only fields declared in [[extraFields()]]
     * will be considered.
     * Expand can also be nested, separeted with dots (.). e.g.: item.expand1.expand2
```
The documentation explained that they could be nested, but that never worked without this patch.

I also had to expand the `yii\rest\Serializer` to automatically recursively fill in the `$expand` with the values of `$model->getRelatedRecords()` and it's children.

Now, I can just do this without having to specify `asArray()` on the ActiveQuery.

```php
    public function actionGetData()
    {
        return Service::find()
            ->with(['serviceItems', 'serviceItems.servicePrices'])
            ->all();
    }
```
The yii\rest\Controller will convert this to a JSON, and the relations will be included.

IMO, this should be default behaviour.

leandrogehlen:
related #8061

michaelarnauts:
What's the status of this MR? We currently depend on it in our production code and we have `cweagans/composer-patches` setup to automatically patch yii2 when installing. 

Is there anything I can do to help this merged?

bboure:
@michaelarnauts I'll have to rebase it.
Also, it has been tagged by the Yii team as part of the 2.0.14 release.
I'd like to have this asap as well on the official repo.

if I get an ok from Yii team, I'll resolve conflicts to make it mergeable.

framework/base/ArrayableTrait.php
```diff
@@ -128,8 +154,54 @@ public function toArray(array $fields = [], array $expand = [], $recursive = tru
...
+     *
+     * @param array $fields The fields requested for extraction
+     * @return array root fields extracted from the given nested fields
+     */
+    public function extractRootFields(array $fields)
+    {
+        $result = [];
+
+        foreach ($fields as $field) {
+            $result[] = current(explode(".", $field, 2));
```
samdark: Why limiting to 2?

samdark:
Overall it looks fine. Need opinions/reviews from @yiisoft/reviewers and @yiisoft/core-developers.

framework/base/ArrayableTrait.php
bboure: @sam002 I guess there is no reason for limit here since I only need the first one.

framework/base/ArrayableTrait.php
```diff
@@ -128,8 +154,54 @@ public function toArray(array $fields = [], array $expand = [], $recursive = tru
...
 
     /**
+     * Extracts the root field names from nested fields.
+     * Nested fields are separated with dots (.). e.g: "item.id"
+     * The previous example would extract "item".
+     *
+     * @param array $fields The fields requested for extraction
+     * @return array root fields extracted from the given nested fields
+     */
+    public function extractRootFields(array $fields)
```
cebe: why is this method public?

framework/base/ArrayableTrait.php
```diff
@@ -128,8 +154,54 @@ public function toArray(array $fields = [], array $expand = [], $recursive = tru
...
+    /**
+     * Extract nested fields from a fields collection for a given root field
+     * Nested fields are separated with dots (.). e.g: "item.id"
+     * The previous example would extract "id".
+     *
+     * @param array $fields The fields requested for extraction
+     * @param string $rootField The root field for which we want to extract the nested fields
+     * @return array nested fields extracted for the given field
+     */
+    public function extractFieldsFor(array $fields, $rootField)
```
cebe: why is this method public?

framework/base/ArrayableTrait.php
```diff
@@ -128,8 +154,54 @@ public function toArray(array $fields = [], array $expand = [], $recursive = tru
     }
 
     /**
+     * Extracts the root field names from nested fields.
+     * Nested fields are separated with dots (.). e.g: "item.id"
+     * The previous example would extract "item".
+     *
+     * @param array $fields The fields requested for extraction
+     * @return array root fields extracted from the given nested fields
+     */
```
cebe: `@since 2.0.14`

framework/base/ArrayableTrait.php
```diff
@@ -128,8 +154,54 @@ public function toArray(array $fields = [], array $expand = [], $recursive = tru
...
+    }
+
+    /**
+     * Extract nested fields from a fields collection for a given root field
+     * Nested fields are separated with dots (.). e.g: "item.id"
+     * The previous example would extract "id".
+     *
+     * @param array $fields The fields requested for extraction
+     * @param string $rootField The root field for which we want to extract the nested fields
+     * @return array nested fields extracted for the given field
```
cebe: `@since 2.0.14`

cebe (changes requested):

framework/base/ArrayableTrait.php
cebe: its optimized, explode will stop after first split.

framework/base/ArrayableTrait.php
bboure: @cebe @samdark added the "since" in doc and made methods protected

samdark (approved):

leandrogehlen:
Is it necessary something to accept this PR?

samdark:
More reviews.

dmirogin (approved):

samdark:
Merged. Thank you!

PR #14227, opened 2017-05-29
Add migration file example for DbSession

bakavic:
Creating a `char` primaryKey in a migration is non-obvious. Used solution from this [issue comment](https://github.com/yiisoft/yii2/issues/10889#issuecomment-302995086).

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none - documentation issue

### DIFF
--- a/docs/guide/runtime-sessions-cookies.md
+++ b/docs/guide/runtime-sessions-cookies.md
@@ -178,6 +178,31 @@ where 'BLOB' refers to the BLOB-type of your preferred DBMS. Below are the BLOB 
   the length of the `id` column. For example, if `session.hash_function=sha256`, you should use a
   length 64 instead of 40.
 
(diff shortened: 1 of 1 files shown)


samdark:
Merged. Thanks!

muhammadcahya:
Migration session file already exists in the yii2 framework

Maybe [this](https://github.com/yiisoft/yii2/blob/master/framework/web/migrations/m160313_153426_session_init.php) file can be updated as well

bakavic:
Hmm, true, the existing migration file should provide a fuller example. Should that file be changed to use char(64) as well?

I have made the change to reference this file in https://github.com/bakavic/yii2/commit/ae765f844fcc08bced15bc6cfa006cd87ada5898.

PR #14232, opened 2017-05-31
fix cloning of DB connection for sqlite in-memory db

### LINKED ISSUE #14131: BC cloning DB does not keep DB's state with SQLite
As discussed with @cebe, this topic is opened to address an issue when cloing a Sqlite DB and expecting it to have the original state. Somehow this is broken and I drilled it down to commit: 
https://github.com/cebe/yii2/commit/5ed6910e1cca13c3fea677a564dc13a36533f50c

and in specific because of:
```php
$this->pdo = null;
```

Hiding this works and enables me to clone & reuse data of a SqlLite database for testing purposes

cebe:
fixes #14131

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14131

### DIFF
--- a/framework/db/Connection.php
+++ b/framework/db/Connection.php
@@ -1090,8 +1090,11 @@ class Connection extends Component
 
         $this->_master = false;
         $this->_slave = false;
-        $this->pdo = null;
         $this->_schema = null;
         $this->_transaction = null;
+        if (strncmp($this->dsn, 'sqlite::memory:', 15) !== 0) {
+            // reset PDO connection, unless its sqlite in-memory, which can only have one connection
+            $this->pdo = null;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


cebe:
@dynasource please check if this fixes your issue.

dynasource:
just checked. This patch seems to fix the issue. Thanks!

dynasource (approved):

framework/db/Connection.php
```diff
@@ -1090,8 +1090,11 @@ public function __clone()
 
         $this->_master = false;
         $this->_slave = false;
-        $this->pdo = null;
         $this->_schema = null;
         $this->_transaction = null;
+        if (strpos($this->dsn, 'sqlite::memory:') !== 0) {
```
samdark: I'd use http://php.net/manual/en/function.strncmp.php instead.

samdark (approved):

framework/db/Connection.php
cebe: right, seems to be faster: https://3v4l.org/MckkL

PR #14239, opened 2017-06-01
Fixes #13058 caught exception thrown during view file rendering produces wrong output

### LINKED ISSUE #13058: Catched View exception still messes up page output
Migrated from https://github.com/yiisoft/yii2-swiftmailer/issues/38

1) Create web controller action:

```php
    public function actionRender()
    {
        try {
            return $this->render('exception');
        } catch (\Exception $e) {
            return $this->render('index');
        }
    }
```

2) Create view file `exception.php` which throws some exception inside:

```php
<h1>File Begin</h1>
<?php throw new Exception('test'); ?>
<h1>File End</h1>
```

Fallback view rendered correctly

### What do you get instead?

Line `<h1>File Begin</h1>` from the view, which causes an exception appears at the top of the render result.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 5.6
| Operating system | CentOS

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13058

### DIFF
--- a/framework/base/View.php
+++ b/framework/base/View.php
@@ -336,6 +336,13 @@ class View extends Component
                 }
             }
             throw $e;
+        } catch (\Throwable $e) {
+            while (ob_get_level() > $_obInitialLevel_) {
+                if (!@ob_end_clean()) {
+                    ob_clean();
+                }
+            }
+            throw $e;
         }
     }
 
--- a/framework/base/View.php
+++ b/framework/base/View.php
@@ -322,12 +322,21 @@ class View extends Component
      */
     public function renderPhpFile($_file_, $_params_ = [])
     {
+        $_obInitialLevel_ = ob_get_level();
         ob_start();
         ob_implicit_flush(false);
         extract($_params_, EXTR_OVERWRITE);
-        require($_file_);
-
-        return ob_get_clean();
+        try {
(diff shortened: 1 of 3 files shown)


framework/base/View.php
```diff
@@ -322,12 +322,21 @@ public function afterRender($viewFile, $params, &$output)
...
         ob_start();
         ob_implicit_flush(false);
         extract($_params_, EXTR_OVERWRITE);
-        require($_file_);
-
-        return ob_get_clean();
+        try {
+            require($_file_);
+            return ob_get_clean();
+        } catch (\Exception $e) {
```
cebe: for PHP 7 compatibility this should include catching \Throwable too.

cebe (changes requested):
other than Throwable it looks good to me.

framework/base/View.php
klimov-paul: added

cebe (approved):

samdark (approved):

samdark:
Merged. Thank you!

PR #14241, opened 2017-06-01
Added support for specifying aliases for migrationNamespaces

cebe:
This is used to specify pathes to migrations that do not have
namespaces.

While not directly supported by the migration command provided by the
framework, these migrations exist in a lot of extensions because custom implementations
of migration controllers out of the framework were using this approach
to load multiple migrations from multiple paths.

Even the framework itself currently ships non-namespaced migrations:

- https://github.com/yiisoft/yii2/blob/17a1d91e4a517f4f15dce973bf3c50dd939dce63/framework/rbac/migrations/m140506_102106_rbac_init.php
- https://github.com/yiisoft/yii2/blob/17a1d91e4a517f4f15dce973bf3c50dd939dce63/framework/caching/migrations/m150909_153426_cache_init.php
- https://github.com/yiisoft/yii2/blob/17a1d91e4a517f4f15dce973bf3c50dd939dce63/framework/log/migrations/m141106_185632_log_init.php

This change allows existing applications to adopt the new namespace-based approach
while keeping existing migrations. While it would be possible to add
namespaces to migrations in the application itself, it is not easily possible
to add namespaces to migrations that come from external sources like
extensions.

related discussion in #13359 and #13356.

this replaces #13359

- [x] I will add some tests to verify behavior later.

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -40,6 +40,7 @@ abstract class BaseMigrateController extends Controller
      *
      * If you have set up [[migrationNamespaces]], you may set this field to `null` in order
      * to disable usage of migrations that are not namespaced.
+     * @see $migrationNamespaces
      */
     public $migrationPath = '@app/migrations';
     /**
@@ -58,7 +59,12 @@ abstract class BaseMigrateController extends Controller
      * ]
      * ```
      *
+     * Since version 2.0.12 you may also specify [aliases](guide:concept-aliases) to directories that contain
+     * migrations without namespace. This is mainly useful to support old extensions that provide migrations
+     * without namespace and to adopt the new feature of namespaced migrations while keeping existing migrations.
+     *
      * @since 2.0.10
+     * @see $migrationPath
      */
     public $migrationNamespaces = [];
     /**
@@ -594,6 +600,9 @@ abstract class BaseMigrateController extends Controller
             if ($this->migrationPath === null) {
                 $migrationNamespaces = $this->migrationNamespaces;
                 $namespace = array_shift($migrationNamespaces);
+                if (strncmp($namespace, '@', 1) === 0) {
+                    $namespace = str_replace('/', '\\', substr($namespace, 1));
+                }
             }
         }
 
@@ -619,7 +628,8 @@ abstract class BaseMigrateController extends Controller
             return $this->migrationPath;
         }
 
-        if (!in_array($namespace, $this->migrationNamespaces, true)) {
+        if (!in_array($namespace, $this->migrationNamespaces, true) &&
+            !in_array('@' . str_replace('\\', '/', $namespace), $this->migrationNamespaces, true)) {
             throw new Exception("Namespace '{$namespace}' not found in `migrationNamespaces`");
         }
 
@@ -634,7 +644,11 @@ abstract class BaseMigrateController extends Controller
      */
     private function getNamespacePath($namespace)
     {
-        return str_replace('/', DIRECTORY_SEPARATOR, Yii::getAlias('@' . str_replace('\\', '/', $namespace)));
+        if (strncmp($namespace, '@', 1) === 0) {
+            return Yii::getAlias($namespace);
+        } else {
+            return str_replace('/', DIRECTORY_SEPARATOR, Yii::getAlias('@' . str_replace('\\', '/', $namespace)));
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


samdark (approved):

klimov-paul:
I am against such change as it mixing up different terms 'namespace' and 'path alias'.

cebe:
@klimov-paul I partly agree with that. We could alternatively add another property or make `migrationPath` accept an array.

dynasource:
I agree on having this in a separate property called $migrationPaths

dynasource (changes requested):
I agree on having this in a separate property called $migrationPaths

samdark (approved):

schmunk42:
Works nicely with this rather complex setup:

```
    'controllerMap' => [
        // Yii 2.0 default migrate controller
        'dev:migrate' => [
            'class' => '\yii\console\controllers\MigrateController',
            'migrationPath' => [
                getenv('APP_MIGRATION_LOOKUP'),
                '@yii/rbac/migrations',
                '@yii/web/migrations',
                '@bedezign/yii2/audit/migrations',
                '@dektrium/user/migrations',
                '@hrzg/widget/migrations',
                '@dmstr/modules/contact/migrations',
                '@dmstr/modules/pages/migrations',
                '@dmstr/modules/redirect/migrations',
                '@vendor/lajax/yii2-translate-manager/migrations',
                '@vendor/pheme/yii2-settings/migrations',
                '@vendor/dmstr/yii2-prototype-module/src/migrations',
            ],
        ],
```

End of output:

```
*** applied m170320_133511_auth_items (time: 0.006s)

*** applying m170322_204909_update_timestamp_columns
    > alter column created_at in table dmstr_page to datetime ... done (time: 0.165s)
    > alter column updated_at in table dmstr_page to datetime ... done (time: 0.184s)
*** applied m170322_204909_update_timestamp_columns (time: 0.360s)

*** applying m170327_120427_update_icon_column
    > update dmstr_page ... done (time: 0.004s)
*** applied m170327_120427_update_icon_column (time: 0.010s)

*** applying m170328_111312_init
    > create table {{%dmstr_contact_log}} ... done (time: 0.053s)
*** applied m170328_111312_init (time: 0.057s)


60 migrations were applied.

Migrated up successfully.
```

---

Also work via the command line:

```
$ yii dev:migrate --migrationPath=@app/migrations/demo-data,@app/migrations
Yii Migration Tool (based on Yii v2.0.12-dev)

Total 1 new migration to be applied:
        m150917_193929_rbac

Apply the above migration? (yes|no) [no]:
```

If the default is wrapped in `[]` - will add a comment in the code.

framework/console/controllers/BaseMigrateController.php
```diff
@@ -32,14 +33,24 @@
...
+     * migrations to load. This is mainly useful to support old extensions that provide migrations
+     * without namespace and to adopt the new feature of namespaced migrations while keeping existing migrations.
+     *
+     * In general, to load migrations from different locations, [[migrationNamespaces]] is the preferable solution
+     * as the migration name contains the origin of the migration in the history, which is not the case when
+     * using multiple migration paths.
+     *
+     * @see $migrationNamespaces
      */
     public $migrationPath = '@app/migrations';
```
schmunk42: I had to wrap this in brackets, so the CLI can recognize it as an array

    public $migrationPath = ['@app/migrations'];

schmunk42 (changes requested):

schmunk42:
At the moment there is `migrationPath` and `migrationNamespaces`, which is a bit confusing due to mixing singular and plural.

What's about having a singular and plural form for both - background is that you could use the singular form to inject just a single folder, while having the basic configuration still available in the plural form. So you do not have to override it with all values, plus the one you want to inject.

We use this workflow i.e. for demo-data or test-data.

cebe:
The naming in plural and singluar has historical reason, in oder to keep BC we can not change it.
Adding more properties sounds complicated to me.

schmunk42:
Basically I am fine with this, just a more general thing from my side to add: I'd find it helpful, if I could see in advance, if a property of method accepts an alias or requires a path.

So having `migrationAliases` along with `migrationPath` in 2.0 and deprecate the latter in 2.1.

cebe:
> I know that migrationPath also accepts an alias, but AFAIR not all framework methods/props with ...path do.

a general rule for this is that all parts that receive input via console or configuration accept an alias. Methods that are only called via code, e.g. FileHelper do not.

PR #14242, opened 2017-06-02
Fixes #5108 `DateValidator` resets `$timestampAttribute` value on empty attribute

### LINKED ISSUE #5108: DateValidator and timestampAttribute
`DateValidator` if attribute is empty `timestampAttribute` not changing. It would be good if it becomes null

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #5108


framework/validators/DateValidator.php
```diff
@@ -266,6 +266,13 @@ public function init()
     public function validateAttribute($model, $attribute)
     {
         $value = $model->$attribute;
+        if ($this->isEmpty($value)) {
+            if ($this->timestampAttribute !== null) {
```
samdark: This check could be skipped. Assigning null is cheaper than checking.

samdark (approved):

framework/validators/DateValidator.php
cebe: this is not the same, current code is correct and must not skip the check.

tests/framework/validators/DateValidatorTest.php
```diff
@@ -200,10 +195,6 @@ public function testValidateAttributeICUFormat($timezone)
             $model->attr_timestamp
         );
         $val = new DateValidator(['format' => 'yyyy-MM-dd']);
-        $model = FakedValidationModel::createWithAttributes(['attr_date' => []]);
-        $val->validateAttribute($model, 'attr_date');
-        $this->assertTrue($model->hasErrors('attr_date'));
-        $val = new DateValidator(['format' => 'yyyy-MM-dd']);
```
cebe: why did you delete these tests?

framework/validators/DateValidator.php
samdark: Oh. Right. I haven't noticed `{` :(

tests/framework/validators/DateValidatorTest.php
klimov-paul: These are tests for empty value parsing failure, which contradicts the issue #5108.
This fix can not be applied keeping them.

cebe: I see what you mean now. If I recall correctly, these should cover robustness of the validator against array input.
The test should be kept for that reason and should be changed to test for non-empty array.

klimov-paul: Tests restored.

cebe (approved):

cebe:
Thanks!

PR #14243, opened 2017-06-02
Fixes #4999: Added support for wildcards at `yii\filters\AccessRule::$controllers`

### LINKED ISSUE #4999: Can we add a "module" option to AccessRule?
And a "matchModule" check, to provide a way to apply rule to all controllers in module.
Use case: I`ve created a behavior that is configured in app config and uses a AccessFilter inside. it provides access filtering across the whole app. "module" option is lacking. 
Something like that

```
protected function matchModule($controller)
    {
        return empty($this->modules) 
            || $controller->module instanceof Application
            || in_array($controller->module->uniqueId, $this->modules, true);

    }
```

Thanks

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #4999

### DIFF
--- a/framework/filters/AccessRule.php
+++ b/framework/filters/AccessRule.php
@@ -43,6 +43,8 @@ class AccessRule extends Component
      * The comparison is case-sensitive.
      *
      * If not set or empty, it means this rule applies to all controllers.
+     *
+     * Since version 2.0.12 controller IDs can be specified as wildcards, e.g. `module/*`.
      */
     public $controllers;
     /**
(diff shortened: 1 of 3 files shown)


cebe (approved):

samdark (approved):

samdark:
Merged. Thanks!

dynasource:
@klimov-paul, would be nice to have some documentation on this feature

framework/filters/AccessRule.php
```diff
@@ -172,7 +174,18 @@ protected function matchAction($action)
...
     protected function matchController($controller)
     {
-        return empty($this->controllers) || in_array($controller->uniqueId, $this->controllers, true);
+        if (empty($this->controllers)) {
+            return true;
+        }
+
+        $id = $controller->getUniqueId();
+        foreach ($this->controllers as $pattern) {
+            if (fnmatch($pattern, $id)) {
```
githubjeka: Why not preg_match ?

> fnmatch - For now, this function is not available on non-POSIX compliant systems except Windows. 

> using fnmatch() instead of preg_match() for frontend search expression input may be way more convenient for non-programming users.

In any case, the use of this function is alarming.

samdark: Alarming? Why?

PR #14253, opened 2017-06-05
Improve base migration controller documentation

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

Other methods currently have the `If it is "all", ...` comments. I thought my wording was a bit better, not sure about it though. Currently comments are a bit inconsistent. Which option do we stick to?

- [x] Fix acceptable parameter types.
- [x] Decide on the `When equals...` vs `If it is...` wording.


samdark:
Merged. Thanks!

Kolyunya:
Glad to be of help!

Issue #14169, opened 2017-05-17
Enh: Count applied/reverted migrations in case of error


Vovan-VE:
Currently MigrateController outputs this

> 2 from 5 migrations were applied

only when a migration does `return false;`. But in case of error (uncaught exception for example due to some stupid typos) there is no such output. If I want to undo migrations just applied (2 from 5), I need to count them manually in the output. It's annoying at least or sometimes may be impossible due too flood produced by each real query.

The suggestion is either to show same summary "2 from 5 migrations were applied" in case of error or output `"$index/$count"` on each processed migration.

### What steps will reproduce the problem?

1. Create N migrations.
2. One migration somewhere in the middle (7th for example) is bad due to some typos.
3. `./yii migrate/up`

### What is the expected result?

```
*** applying foo
    ...
*** applyed foo

*** applying bar
6 of 10 migrations were applied
7th one was failed with error
Exception 'Exception' with message 'message'

in ...
```

### What do you get instead?

```
*** applying foo
    ...
*** applyed foo

*** applying bar
Exception 'Exception' with message 'message'

in ...
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2

njasm:
what about wrapping all migrations Up or Down inside a transaction, and make them atomic? would that be an acceptable solution?

Vovan-VE:
@njasm , no, at least until MySQL would run set of DDL statements under transaction.

PR #14262, opened 2017-06-07
Show number of applied/reverted migrations in case of error

Vovan-VE:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14169
| Obsoleted by | PR #15095

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -179,14 +179,22 @@ abstract class BaseMigrateController extends Controller
 
         $applied = 0;
         if ($this->confirm('Apply the above ' . ($n === 1 ? 'migration' : 'migrations') . '?')) {
-            foreach ($migrations as $migration) {
-                if (!$this->migrateUp($migration)) {
-                    $this->stdout("\n$applied from $n " . ($applied === 1 ? 'migration was' : 'migrations were') ." applied.\n", Console::FG_RED);
-                    $this->stdout("\nMigration failed. The rest of the migrations are canceled.\n", Console::FG_RED);
+            try {
+                foreach ($migrations as $migration) {
+                    if (!$this->migrateUp($migration)) {
+                        $this->stdout("\n$applied from $n " . ($applied === 1 ? 'migration was' : 'migrations were') ." applied.\n", Console::FG_RED);
+                        $this->stdout("\nMigration failed. The rest of the migrations are canceled.\n", Console::FG_RED);
 
-                    return self::EXIT_CODE_ERROR;
+                        return self::EXIT_CODE_ERROR;
+                    }
+                    $applied++;
                 }
-                $applied++;
+            } catch (\Exception $e) {
+                $this->showSummaryOnError($applied, $n, 'applied');
+                throw $e;
+            } catch (\Throwable $e) {
+                $this->showSummaryOnError($applied, $n, 'applied');
+                throw $e;
             }
 
             $this->stdout("\n$n " . ($n === 1 ? 'migration was' : 'migrations were') ." applied.\n", Console::FG_GREEN);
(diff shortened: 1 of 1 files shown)


framework/console/controllers/BaseMigrateController.php
```diff
@@ -179,14 +179,22 @@ public function actionUp($limit = 0)
...
         $applied = 0;
         if ($this->confirm('Apply the above ' . ($n === 1 ? 'migration' : 'migrations') . '?')) {
-            foreach ($migrations as $migration) {
-                if (!$this->migrateUp($migration)) {
-                    $this->stdout("\n$applied from $n " . ($applied === 1 ? 'migration was' : 'migrations were') ." applied.\n", Console::FG_RED);
-                    $this->stdout("\nMigration failed. The rest of the migrations are canceled.\n", Console::FG_RED);
+            try {
+                foreach ($migrations as $migration) {
+                    if (!$this->migrateUp($migration)) {
+                        $this->stdout("\n$applied from $n " . ($applied === 1 ? 'migration was' : 'migrations were') ." applied.\n", Console::FG_RED);
```
samdark: Shouldn't it be replaced with `showSummaryOnError()` call?

framework/console/controllers/BaseMigrateController.php
```diff
@@ -240,14 +248,22 @@ public function actionDown($limit = 1)
...
         $reverted = 0;
         if ($this->confirm('Revert the above ' . ($n === 1 ? 'migration' : 'migrations') . '?')) {
-            foreach ($migrations as $migration) {
-                if (!$this->migrateDown($migration)) {
-                    $this->stdout("\n$reverted from $n " . ($reverted === 1 ? 'migration was' : 'migrations were') ." reverted.\n", Console::FG_RED);
-                    $this->stdout("\nMigration failed. The rest of the migrations are canceled.\n", Console::FG_RED);
+            try {
+                foreach ($migrations as $migration) {
+                    if (!$this->migrateDown($migration)) {
+                        $this->stdout("\n$reverted from $n " . ($reverted === 1 ? 'migration was' : 'migrations were') ." reverted.\n", Console::FG_RED);
```
samdark: Shouldn't it be replaced with `showSummaryOnError()` call?

framework/console/controllers/BaseMigrateController.php
```diff
@@ -301,19 +317,40 @@ public function actionRedo($limit = 1)
...
 
         if ($this->confirm('Redo the above ' . ($n === 1 ? 'migration' : 'migrations') . '?')) {
-            foreach ($migrations as $migration) {
-                if (!$this->migrateDown($migration)) {
-                    $this->stdout("\nMigration failed. The rest of the migrations are canceled.\n", Console::FG_RED);
+            $reverted = 0;
+            try {
+                foreach ($migrations as $migration) {
+                    if (!$this->migrateDown($migration)) {
+                        $this->stdout("\nMigration failed. The rest of the migrations are canceled.\n", Console::FG_RED);
```
samdark: Shouldn't it be replaced with `showSummaryOnError()` call?

framework/console/controllers/BaseMigrateController.php
```diff
@@ -301,19 +317,40 @@ public function actionRedo($limit = 1)
...
-            foreach (array_reverse($migrations) as $migration) {
-                if (!$this->migrateUp($migration)) {
-                    $this->stdout("\nMigration failed. The rest of the migrations are canceled.\n", Console::FG_RED);
 
-                    return self::EXIT_CODE_ERROR;
+            $applied = 0;
+            try {
+                foreach (array_reverse($migrations) as $migration) {
+                    if (!$this->migrateUp($migration)) {
+                        $this->stdout("\nMigration failed. The rest of the migrations are canceled.\n", Console::FG_RED);
```
samdark: Shouldn't it be replaced with `showSummaryOnError()` call?

samdark (changes requested):

framework/console/controllers/BaseMigrateController.php
Vovan-VE: Do you mean to extract full `try` to method with which arguments?

Vovan-VE:
This PR in outdated. Now it would be much better to reimplement it in a new PR. What about handling exceptions once inside `migrateUp()` and `migrateDown()` methods to let them just to return `false`? To do it these methods need to show caught exception. Is there another way but `ErrorHandler::convertExceptionToString()` to do it?

PR #14281, opened 2017-06-09
[Work in progress] Update to PHP 7.1

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | ?
| Fixed issues  | #11397


framework/base/ErrorHandler.php
```diff
@@ -116,11 +116,8 @@ public function handleException($exception)
                 }
                 exit(1);
             }
-        } catch (\Exception $e) {
-            // an other exception could be thrown while displaying the exception
-            $this->handleFallbackExceptionMessage($e, $exception);
         } catch (\Throwable $e) {
-            // additional check for \Throwable introduced in PHP 7
+            // additional check for \Throwable
```
Faryshta: Since the minium requirement is 7.1 how about

`catch(Exception|Throwable $e)`

cebe: keep the comment from above instead:

```
// an other exception could be thrown while displaying the exception
```

there is no additional check now.

cebe (changes requested):
requirement in composer.json should be raised.

framework/base/ErrorHandler.php
samdark: Fixed.

composer.json
```diff
@@ -99,14 +99,14 @@
...
-        "platform": {"php": "5.5"}
+        "platform": {"php": "7.1"}
     },
     "bin": [
         "framework/yii"
     ],
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0.x-dev"
+            "dev-master": "2.1.x-dev"
```
rob006: Why this was changed?

samdark: It wasn't https://github.com/yiisoft/yii2/blob/2.1/framework/composer.json#L85

samdark: Accidentally.

.travis.yml
```diff
@@ -138,12 +96,8 @@ before_script:
   - mysql -e "GRANT ALL PRIVILEGES ON *.* TO 'travis'@'localhost' WITH GRANT OPTION;";
   - psql -U postgres -c 'CREATE DATABASE yiitest;';
 
-  # enable code coverage on PHP 7.1, only one PHP version needs to generate coverage data
-  - |
-    if [ $TRAVIS_PHP_VERSION = '7.1' ]; then
-      PHPUNIT_FLAGS="--coverage-clover=coverage.clover"
-    fi
```
cebe: this should be kept as it now applies to 7.1 and nightly and as soon as 7.2 is out we need it again.

samdark: https://github.com/yiisoft/yii2/commit/f41399a6dc4cf395d6cd0c0d07a98ff93106b240

PR #14286, opened 2017-06-12
User primary inputmask alias

samdark:
Since jquery.inputmask was renamed to just inputmask, this change is necessary in order to keep dependencies more stable. There's an alias to old name but it failed recently.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | ?
| Fixed issues  | -

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -76,7 +76,7 @@
         "ezyang/htmlpurifier": "~4.6",
         "cebe/markdown": "~1.0.0 | ~1.1.0",
         "bower-asset/jquery": "2.2.*@stable | 2.1.*@stable | 1.11.*@stable | 1.12.*@stable",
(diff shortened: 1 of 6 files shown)


cebe (approved):
changelog and upgrade note missing

samdark:
Changelog added. I don't think this change makes any difference to end user so see no need to mention it in UPGRADE.

cebe:
> I don't think this change makes any difference to end user so see no need to mention it in UPGRADE.

If someone relies on the path of installed packages it will break, so we should mention it. AssetManager may be configured manually.

samdark:
Indeed. Done.

PR #14298, opened 2017-06-15
Set default response formatter configs using array syntax

brandonkelly:
With the exception of JSONP, if a controller wants to change default response formatter options, currently it needs to respecify the formatter class name:

```php
Yii::$app->response->formatters[Response::FORMAT_JSON] = [
    'class' => 'yii\web\JsonResponseFormatter',
    'prettyPrint' => true,
];
```

With this change, you can just tweak the setting, since the formatter config will already be defined as an array:

```php
Yii::$app->response->formatters[Response::FORMAT_JSON]['prettyPrint'] = true;
```

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes


samdark:
Looks OK to merge. @yiisoft/core-developers do you see any consequences regarding BC breaks?

klimov-paul (approved):

cebe (approved):

samdark:
Merged. Thanks!

PR #14303, opened 2017-06-16
Enh: comment wording in BaseActiveRecord::init()

drodata:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Based on the 3rd convention of extending Component/Object:

> If you override the [[yii\base\Object::init()]] method, make sure you call the parent implementation of init() **at the beginning of** your init() method.
>
> https://github.com/yiisoft/yii2/blob/master/docs/guide/concept-components.md

Suppose I created an AR class named `Customer`, and attached some events by overriding the `init()` method, I think

```php
public function init()
{
    parent::init();
    $this->on(self::EVENT_AFTER_INSERT, [$this, 'doSomething']);
}
```

is better than

```php
public function init()
{
    $this->on(self::EVENT_AFTER_INSERT, [$this, 'doSomething']);
    parent::init();
}
```

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -889,7 +889,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      * Initializes the object.
      * This method is called at the end of the constructor.
      * The default implementation will trigger an [[EVENT_INIT]] event.
-     * If you override this method, make sure you call the parent implementation at the end
+     * If you override this method, make sure you call the parent implementation at the beginning
      * to ensure triggering of the event.
      */
     public function init()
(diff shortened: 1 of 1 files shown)


cebe:
why is it better? it depends on what you want.

drodata:
@cebe I'm sorry about the fact that I didn't express my thought clearly.

The point that confused me is, giving the following two conditions:

1. > If you override the [[yii\base\Object::init()]] method, make sure you call the parent implementation of init() **at the beginning** of your init() method.
2. the `BaseActiveRecord` class is a desendant of `yii\base\Object`;

I got a conclusion (may be wrong): When descendants of `BaseActiveRecord` override `init()`, they should call `parent::init()` **at the beginning** of there `init()` methods, not **at the end**.

cebe:
Good catch, that should be more clear. In most cases it does not matter whether init() is called at the beginning or at the end, but for classes that implement events on init() or set configuration, it may be relevant. So documentation should be more clear about that.

samdark:
I think we should omit specifying where to call the method since both cases are valid.

Issue #14296, opened 2017-06-15
EmailTarget ignores `mail()` failures


rob006:
### What steps will reproduce the problem?

Disable `mail()` function in PHP config (or just ensure that it will return `false`). Use default configuration of `yii\swiftmailer\Mailer` and configure `EmailTarget` for errors. Log error.

### What is the expected result?

Warning that email can not be sent.

### What do you get instead?

Nothing. `swiftmailer` [swallows `mail()` errors and returns `false`](https://github.com/swiftmailer/swiftmailer/blob/v5.4.8/lib/classes/Swift/Transport/SimpleMailInvoker.php#L33-L37), but `EmailTarget` [ignores `send()` result](https://github.com/yiisoft/yii2/blob/ba0ab403b52124c941dbeb46fbd9efdc12252a5d/framework/log/EmailTarget.php#L85). So failures are just ignored.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      |  7.1.4
| Operating system | Linux

bizley:
FileTarget throws InvalidConfigException only on file opening error - no warning for file write errors.
DbTarget also does not check if log is db inserted and similar story with SyslogTarget. 

How this warning message should be delivered?

rob006:
I have no idea, didn't investigate this further. But I changed my configuration to use SMTP, and with invalid config I get nice warning in debug toolbar and log file:
```
[warning][yii\log\Dispatcher::dispatch] Unable to send log via yii\log\EmailTarget: Exception 'Swift_TransportException' with message 'Connection could not be established with host XXX [Connection refused #61]' 
```

So it looks doable.

bizley:
But that's the point - you need another target configured to get the first target's error.

rob006:
So what is the problem? I have another target, why hide this warning in that case?

bizley:
Well, you are right. All targets should inform in case the logs can not be delivered.

PR #14309, opened 2017-06-17
Bug #14296 log targets failure capturing [not fully completed]

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | isolated log tests - yes, not sure about the whole app
| Fixed issues  | #14296

This fix adds condition check for log targets exporting and in case of failure exception is thrown.
Exception is caught by dispatcher and the warning can be passed to the next log target if there are any left.

### EmailTarget
- [x] condition check based on the result of `send()` method implemented through `\yii\mail\MessageInterface`
- [x] tests covered

### FileTarget
- [x] condition check based on the result of `fwrite()` and `file_put_contents()` methods - exception is thrown when methods return `false` or write less bytes than expected
- [ ] tests covered - not sure how to mock `fwrite()` error properly - possible solution is to do it like in the `/tests/framework/log/SyslogTargetTest.php` or to use `vfsStream` library but this is not included in the `composer.json` right now

### DbTarget
- [x] condition check based on whether the result of `execute()` method with `INSERT` statement returned more than 0 rows or not
- [ ] tests covered - not sure how to mock failed INSERT statement

### SyslogTarget
- [x] condition check based on the result of `syslog()` method
- [x] tests covered - overridden methods `openlog()`, `syslog()` and `closelog()` return boolean values now

### DIFF
--- a/framework/log/DbTarget.php
+++ b/framework/log/DbTarget.php
@@ -9,7 +9,6 @@ namespace yii\log;
 
 use Yii;
 use yii\base\InvalidConfigException;
-use yii\base\InvalidValueException;
 use yii\db\Connection;
 use yii\db\Exception;
 use yii\di\Instance;
@@ -59,8 +58,9 @@ class DbTarget extends Target
 
     /**
      * Stores log messages to DB.
+     * Starting from version 2.0.14, this method throws LogRuntimeException in case the log can not be exported.
      * @throws Exception
-     * @throws InvalidValueException
+     * @throws LogRuntimeException
      */
     public function export()
     {
@@ -93,7 +93,7 @@ class DbTarget extends Target
                 ])->execute() > 0) {
                 continue;
             }
-            throw new InvalidValueException('Unable to export log through database!');
+            throw new LogRuntimeException('Unable to export log through database!');
         }
     }
 }
--- a/framework/log/DbTarget.php
+++ b/framework/log/DbTarget.php
@@ -9,7 +9,9 @@ namespace yii\log;
 
 use Yii;
 use yii\base\InvalidConfigException;
+use yii\base\InvalidValueException;
 use yii\db\Connection;
+use yii\db\Exception;
 use yii\di\Instance;
 use yii\helpers\VarDumper;
 
@@ -57,6 +59,8 @@ class DbTarget extends Target
 
     /**
      * Stores log messages to DB.
+     * @throws Exception
+     * @throws InvalidValueException
      */
     public function export()
     {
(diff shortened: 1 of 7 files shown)


framework/log/DbTarget.php
```diff
@@ -80,13 +84,16 @@ public function export()
...
+            if ($command->bindValues([
+                    ':level' => $level,
+                    ':category' => $category,
+                    ':log_time' => $timestamp,
+                    ':prefix' => $this->getMessagePrefix($message),
+                    ':message' => $text,
+                ])->execute() > 0) {
+                continue;
+            }
+            throw new InvalidValueException('Unable to export log through database!');
```
klimov-paul: `InvalidValueException` does not seem to be appropriate here: its description said:

> represents an exception caused by a function returning a value of unexpected type

Both `true` and `false` belong to the same type - `bool`. Why such exception should be here?

klimov-paul (changes requested):

framework/log/DbTarget.php
bizley: I agree. I used this one only because all other exceptions seem to be inappropriate and I did not want to use general `yii\base\Exception`. Should I change it to `yii\base\Exception`?

klimov-paul: It is better to introduce a new exception class under `yii\log\*` namespace or maybe use [\RuntimeException](http://php.net/manual/en/class.runtimeexception.php).

framework/log/FileTarget.php
```diff
@@ -117,9 +120,21 @@ public function export()
             $this->rotateFiles();
             @flock($fp, LOCK_UN);
             @fclose($fp);
-            @file_put_contents($this->logFile, $text, FILE_APPEND | LOCK_EX);
+            $writeResult = @file_put_contents($this->logFile, $text, FILE_APPEND | LOCK_EX);
+            if ($writeResult === false) {
+                throw new LogRuntimeException('Unable to export log through file!');
```
samdark: Does it make sense to use `error_get_last()` here to output more details about failure reasons?

bizley: Neat idea. Maybe add it to `LogRuntimeException` ?

samdark: We have to be careful about it since errors stack is quite fragile and there could be totally unrelated errors. Here I've suggested it because of silencing operator in `@file_put_contents`. It may fail for too many reasons and not knowing exact one would cause debug hell for sure.

bizley: Ok, so just as a part of thrown message then.

samdark:
@bizley do you plan finishing it?

bizley:
@samdark except two missing tests this is (was?) ready to go. I someone could help with these tests it would be great.

samdark:
@bizley resolving conflicts at least...

bizley:
I've lost this branch completely so I've recreated it and rePRed in https://github.com/yiisoft/yii2/pull/15591

Issue #11170, opened 2016-03-22
Masked input hashPlugin options inside jquery ready


YGugnin:
Hi
https://github.com/yiisoft/yii2/blob/master/framework/widgets/MaskedInput.php#L149
you added variable to JQuery ready block, so other scripts/plugins have no access to this variable. Can you change View::POS_READY to View::POS_HEAD please

PR #14321, opened 2017-06-21
Update MaskedInput.php

DaveFerger:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no
| Fixed issues  | #11170

```php
$view->registerJs("var` {$this->_hashVar} = {$encOptions};", View::POS_READY);
```

to

```php
$view->registerJs("var {$this->_hashVar} = {$encOptions};", View::POS_END);
```

## Solution
If you add variable to JQuery ready block other scripts/plugins have no access to this variable.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.13 under development
 ------------------------
 
-- Chg #14321: `yii\widgets\MaskedInput` is now registering its JavaScript `clientOptions` initialization code in head secrtion (DaveFerger)
+- Chg #14321: `yii\widgets\MaskedInput` is now registering its JavaScript `clientOptions` initialization code in head section (DaveFerger)
 - Bug #14248: `yii\console\controllers\MessageController` no longer outputs colorized filenames when console does not support text colorization (PowerGamer1)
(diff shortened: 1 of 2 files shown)


samdark:
It's not a good practice to have script tags in head. What's your use case?

DaveFerger:
yes, i know it, but this is just a variable for store widget options.
i use a lot of widget from [Kartik](https://github.com/kartik-v), he use this solutions too.
with this solution we can reference for options of this widget at later in code.
e.g. in [dynamic-form](https://github.com/wbraganca/yii2-dynamicform/) widget we can duplicate input[type="text"] elements, and after that init maskedInput by this reference variable.

if you pull this solution, the options variable will be in head, but the init script keep in end of the rendered html

rob006:
@samdark Why having inline scripts in head is a bad practice?

samdark:
Good question. It's not worse than it was before indeed. In both cases there's same rendering delay.

@DaveFerger would you please add a line for the CHANGELOG? Thanks.

samdark (approved):

dynasource (approved):

samdark:
Merged. Thanks!

PR #14327, opened 2017-06-21
Allow failures for now

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,6 @@ env:
     - DEFAULT_COMPOSER_FLAGS="--prefer-dist --no-interaction --no-progress --optimize-autoloader"
     - TASK_TESTS_PHP=1
     - TASK_TESTS_JS=0
-    - TASK_TESTS_CS=0
     - TASK_TESTS_COVERAGE=0
 
 
@@ -111,11 +110,6 @@ matrix:
 
   allow_failures:
     - php: nightly
-    # test coding standards
-    - php: 7.1
-      env: TASK_TESTS_PHP=0 TASK_TESTS_CS=1
-      # overwrite services used for PHP tests
-      services:
 
 install:
   - |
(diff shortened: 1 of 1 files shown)


rob006:
@samdark Why did you remove it?

samdark:
Cause it doesn't make sense to run it in Travis. Going to run it before each release.

rob006:
Then people can't just run fixer in their branches when submitting PRs because it will bring lot of unrelated changes and possible conflicts.

samdark:
How should travis solve that?

rob006:
If tests on travis will fail because of invalid CS, people will fix it to keep travis green. At least in theory - test results are often just ignored. :/ But this at least gives a change to have consistent CS in master - we already have a standard for changelog order and we all know how people follow them. :D

samdark:
And exactly because of that we're fixing changelog order automatically before making releases.

rob006:
Changelog is reordered multiple times between releases. It must be reordered, because random order of entries creates invalid merges and duplicated entries, and it is hard to fix them without reordering. But reordering create more conflicts and invalid merges, so we need fix them more often, and we need reorder changelog for that... and so on. It probably will be less problematic to check changelog order as part of the CS test and force people to put changelog lines in the right place.

samdark:
I understand that. Still, CS fixed in travis creates too much noise that is quite hard to read and fix. Applying automatic fixes before release is faster even considering the merge process...

rob006:
What noise? CS test are separated from the rest of the test. And fixing CS issues is a matter of one command.

samdark:
1. CS errors are not descriptive at all in travis output.
2. It makes tests fail w/o pull requester understanding on how to fix it.

samdark:
I haven't played enough with current config btw. It makes sense overall but, for example, on Windows it touches all the files changing whitespaces so it's nearly impossible to review changes. Once it will get stable, I think we can discuss including it into travis once again.

Issue #14334, opened 2017-06-23
\yii\db\QueryBuilder::buildNotCondition loses params when operand is \yii\db\Expression


Ni-san:
### What steps will reproduce the problem?

Trying to build condition like this:

```php
$query = SomeActiveRecordModel::find()
    ->where([
        'not',
        new Expression(
            'ABS(quantity) < :eps',
            [':eps' => self::EPSILON]
        ),
    ])
    ->exists();
```

### What do you get instead?
Throws exception:

```
ERROR:  bind message supplies 0 parameters, but prepared statement "pdo_stmt_00000001" requires 1
The SQL being executed was: SELECT * FROM "schemaname"."tablename" WHERE NOT (ABS(quantity) < :eps)
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.0.18
| Operating system | Ubuntu 16.04
| PostgreSQL version | 9.5.7

PR #14335, opened 2017-06-23
refs #14334 \yii\db\QueryBuilder::buildNotCondition loses params when operand is \yii\db\Expression

pchapl:
Add test case for reproduce issue #14334 

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | no


samdark:
Does that happen with PostgreSQL only?

samdark (approved):

Ni-san:
> Does that happen with PostgreSQL only?
Reproduced for mysql. Moved testcase.

samdark:
https://travis-ci.org/yiisoft/yii2/jobs/246950005

PR #14339, opened 2017-06-23
Add getFile method for $request

243083df:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -1283,10 +1283,10 @@ class Request extends \yii\base\Request
      * Returns \yii\web\UploadedFile instance
      *
      * ```php
-     * $file = $request->getFile('name.txt')
+     * $file = $request->getUploadedFile('avatar')
      * ```
      *
-     * @param string $name Filename
+     * @param string $name the name of the file input field
      * @return null|UploadedFile
      */
     public function getUploadedFile(string $name)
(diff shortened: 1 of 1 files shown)


243083df:
According to https://github.com/yiisoft/yii2/wiki/Plan-for-next-major-versions

samdark:
It could be added to 2.0.13 instead of 2.1.

cebe:
should not be limited to one method but all static methods in `UploadedFile` should be integrated into request in some way.

klimov-paul:
I suppose a new class called `UploadedFileCollection` should be introduced instead of static methods at `UploadedFile`. It can be integrated into `Request` in the same way as `HeaderCollection`.
The final interface will be:

```php
$uploadedFile = Yii::$app->request->uploadedFiles->getByName('foo');
```

samdark:
There's PSR-7 interface for that btw. http://www.php-fig.org/psr/psr-7/

samdark:
```php
function getUploadedFiles()
{
    // returns an array similar to $_FILES where, instead of arrays
    // there are instances of Psr\Http\Message\UploadedFileInterface
}
```

Since it's likely that we'll adopt PSR-7 in the future, I'd name the method in a similar way i.e. `getUploadedFile()`.

dynasource (approved):

samdark (approved):

klimov-paul:
Resolved via #14801

PR #14340, opened 2017-06-23
Fixes #14304: Fixed `yii\validators\UniqueValidator` and `yii\validators\ExistValidator` to skip prefixes in case expressions are used

### LINKED ISSUE #14304: Recent change in UniqueValidator.php breaks code
I've been using the workaround described in #9014 by @Faryshta to validate uniqueness of a username case insensitive. This stopped working recently and now generates an MySQL error like this...

```
Next exception 'yii\db\Exception' with message 'SQLSTATE[42000]:
Syntax error or access violation: 1064 You have an error in your SQL
syntax; check the manual that corresponds to your MySQL server
version for the right syntax to use
near '[[lower(username)]]='myname'))' at line 1
The SQL being executed was:
SELECT EXISTS(SELECT * FROM `user`
WHERE (`user`.[[lower(username)]]='myname'))' in 
```

Seems at a quick glance like the changes introduced to fix  #14161, #14150 is causing it.

Best would be if issue #9014 was solved to skip the workaround.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      |   5.6.30
| Operating system | Debian

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #14304

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -16,6 +16,7 @@ Yii Framework 2 Change Log
 - Enh #14081: Added `yii\caching\CacheInterface` to make custom cache extensions adoption easier (silverfire)
 - Chg #14286: Used primary inputmask package name instead of an alias (samdark)
 - Enh #14298: The default response formatter configs defined by `yii\web\Response::defaultFormatters()` now use the array syntax (brandonkelly)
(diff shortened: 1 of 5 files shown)


samdark:
Issue is solved by it seems something else is now broken. Need to re-check tests...

PR #14343, opened 2017-06-24
Fix active fixture when used in second db connection

margori:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

I found that ActiveFixture don't set db connection for models that require to be created in not main db connection.

### DIFF
--- a/framework/test/ActiveFixture.php
+++ b/framework/test/ActiveFixture.php
@@ -62,6 +62,8 @@ class ActiveFixture extends BaseActiveFixture
         if ($this->modelClass === null && $this->tableName === null) {
             throw new InvalidConfigException('Either "modelClass" or "tableName" must be set.');
         }
+        $modelClass = $this->modelClass;
+        $this->db = $modelClass::getDb();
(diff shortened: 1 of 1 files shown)


framework/test/ActiveFixture.php
```diff
@@ -62,6 +62,8 @@ public function init()
         if ($this->modelClass === null && $this->tableName === null) {
             throw new InvalidConfigException('Either "modelClass" or "tableName" must be set.');
         }
+        $modelClass = $this->modelClass;
```
cebe: modelClass can be null, so a check must be added.

cebe:
> Unfortunately a use case is missing.

@samdark setting `modelClass` which uses a different DB than the default connection would use the correct DB connection now, current behavior would be to look for the table in the default DB, where it does not exist.

bizley:
Closed in favor of #18577

Issue #14246, opened 2017-06-02
Idea: avoid double declare link and type of relation if used via()


bscheshirwork:
What about AR function?
yii\db\BaseActiveRecord

    /**
     * Get a relation from the another class relation definition
     * @param ActiveRecord|array|callable|string $target the model class who have relation
     * @param string $relationName relation name in target (case sensitive, without 'get')
     * @return ActiveQuery
     */
    public function hasRelation($target, $relationName){
        $target instanceof ActiveRecord || $target = Yii::createObject($target);
        /** @var ActiveQuery $relationQuery */
        $relationQuery = $target->getRelation($relationName);
        $relationQuery->primaryModel = $this;
        return $relationQuery;
    }

class Order extends ActiveRecord
{
    public function getItems(){
        return $this->hasRelation(OrderItem::className(), 'item')->via('orderItems');
    }
    public function getOrderItems()
    {
        return $this->hasMany(OrderItem::className(), ['order_id' => 'id']);
    }
}

It's avoid double declare link and type of relation (multiple)

dynasource:
can you elaborate on your usecase and why it is necessary compared to `BaseActiveRecord::createRelationQuery()`

bscheshirwork:
now I propose
```php
    public function viaRelation($target, $relationName, $via){
        $target instanceof ActiveRecord || $target = Yii::createObject($target);
        /** @var ActiveQuery $relationQuery */
        $relationQuery = $target->getRelation($relationName);
        $relationQuery->primaryModel = $this;
        return $relationQuery->via($via);
    }
```
vs
```php
    protected function createRelationQuery($class, $link, $multiple)
    {
        /* @var $class ActiveRecordInterface */
        /* @var $query ActiveQuery */
        $query = $class::find();
        $query->primaryModel = $this;
        $query->link = $link;
        $query->multiple = $multiple;
        return $query;
    }
```
:
 `$multiple` get from parent declaration of exist relation. And this exist relation already has name for use it.

bscheshirwork:
> It's avoid double declare link and type of relation (multiple)

bscheshirwork:
i.e. model3, who use relation of model2 as via to model1 not necessary know about details of relation definition into model2

bscheshirwork:
From docs
[https://github.com/yiisoft/yii2/tree/master/docs/guide](https://github.com/yiisoft/yii2/tree/master/docs/guide)
```php
class Order extends ActiveRecord
{
[...]

PR #14356, opened 2017-06-27
via and hasRelation

bscheshirwork:
Each relation must be set once. Not mater duplicate declaration in another model. 
Each model can store only declaration of the his relations.

| Q                    | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #14246


framework/db/BaseActiveRecord.php
```diff
@@ -409,6 +410,43 @@ public function hasMany($class, $link)
...
+     *     return $this->hasOne(Item::className(), ['id' => 'item_id']);
+     * }
+     * ```
+     *
+     * @param ActiveRecord|array|callable|string $target the model class who have relation
+     * @param string $relationName relation name in target (case sensitive, without 'get')
+     * @return ActiveQuery
+     */
+    public function hasRelation($target, $relationName){
+        $target instanceof ActiveRecord || $target = Yii::createObject($target);
```
dynasource: this looks wrong. I'd expect a hasRelation to return a boolean

tests/framework/data/ActiveDataProviderTest.php
```diff
@@ -101,6 +101,21 @@ public function testActiveRelationVia()
         $this->assertCount(2, $items);
     }
 
+    public function testActiveRelationViaHasRelation()
```
dynasource: a new method is introduced in framework/db/BaseActiveRecord.php but a unit test is supplied in ActiveDataProviderTest.php. I'd expect specific unit tests for BaseActiveRecord?

dynasource (changes requested):

framework/db/BaseActiveRecord.php
bscheshirwork: like a hasOne, hasMany?

So... How about `fromRelation` or `viaRelation` (and directly add via into method)?

bscheshirwork:
```php
class Order extends ActiveRecord
{
    public function getItems(){
        return $this->viaRelation(OrderItem::className(), 'item', 'orderItems');
    }
    public function getOrderItems()
    {
        return $this->hasMany(OrderItem::className(), ['order_id' => 'id']);
    }
}
```

dynasource: ahh, you took that as a reference, understood. Well, IMHO hasOne & hasMany are not chosen correctly. It should describe what it does, so something like:
- `getRelationQuery()` or `getLinkQuery`

bscheshirwork: So... This more like a `viaTable` (Table name -> relation name)

bscheshirwork: i.e. if add to end of method `->via('relationAttributeName')`

bscheshirwork:
```php
    public function viaRelation($target, $relationName, $via){
        $target instanceof ActiveRecord || $target = Yii::createObject($target);
        /** @var ActiveQuery $relationQuery */
        $relationQuery = $target->getRelation($relationName);
        $relationQuery->primaryModel = $this;
        return $relationQuery->via($via);
    }
```

bscheshirwork:
```php
   public function viaRelation($relationName, $via, callable $callable = null)
   {
       $relation = $this->getRelation($via);
       $target = Yii::createObject($relation->modelClass);
       /** @var ActiveQuery $relationQuery */
       $relationQuery = $target->getRelation($relationName);
       $relationQuery->primaryModel = $this;
       return $relationQuery->via($via, $callable);
   }
```

tests/framework/data/ActiveDataProviderTest.php
bscheshirwork: tests/framework/db/ActiveRecordTest.php
```php
    public function testJoinWithViaRelation()
    {
        Order::getDb()->getQueryBuilder()->separator = "\n";
        $rows = Order::find()->joinWith([
            'items2' => function ($q) {
                $q->orderBy('item.id');
            },
        ])->all();
        $this->assertNotEmpty($rows);
    }
```

bscheshirwork:
@dynasource 
>  requested changes on 29 Jul 
all is OK now?

dmirogin:
Hmm, based on your example, what will return getItemsCategory if there are different categories for those items?

framework/db/BaseActiveRecord.php
```diff
@@ -410,6 +411,58 @@ public function hasMany($class, $link)
...
+     * Its signature should be `function($query)`, where `$query` is the query to be customized.
+     * @return ActiveQuery ActiveQuery type of end-side model
+     * @since 2.0.14
+     * @see hasOne()
+     * @see hasMany()
+     */
+    public function viaRelation($relationName, $via, callable $callable = null)
+    {
+        $relation = $this->getRelation($via);
+        $target = Yii::createObject($relation->modelClass);
```
bscheshirwork: I forgot for new feature `::instance()`  
(PR is older)  
In this place is correct to use it?

bscheshirwork:
@dmirogin thsnks, I will expand a example

bscheshirwork:
Additional question:
What is signature is more user-friendly:

```php
    /**
     * @param string $relationName relation name in target (case sensitive, without 'get')
     * @param string $via relation name in called (self) class (case sensitive, without 'get')
     * @param callable $callable a PHP callback for customizing the relation associated with the junction ActiveQuery in `via`.
     * Its signature should be `function($query)`, where `$query` is the query to be customized.
     * @return ActiveQuery ActiveQuery type of end-side model
     * @since 2.0.14
     * @see hasOne()
     * @see hasMany()
     */
    public function viaRelation($relationName, $via, callable $callable = null)

```
or
 
```php
    /**
     * @param string $via relation name in called (self) class (case sensitive, without 'get')
     * @param string $relationName relation name in target (case sensitive, without 'get')
     * @param callable $callable a PHP callback for customizing the relation associated with the junction ActiveQuery in `via`.
     * Its signature should be `function($query)`, where `$query` is the query to be customized.
     * @return ActiveQuery ActiveQuery type of end-side model
     * @since 2.0.14
     * @see hasOne()
     * @see hasMany()
     */
    public function viaRelation($via, $relationName, callable $callable = null)

```

framework/db/BaseActiveRecord.php
samdark: Yes.

samdark:
How about the following?

```php
public function relationVia($relationName, $via, callable $callable = null)
```

bscheshirwork:
@samdark 
>How about the following?
> public function relationVia($relationName, $via, callable $callable = null)

I try it first. 
I try to write some (~10-15) functions in this notation.
I think this is not user-friendly (have more differences of hasOne/hasMany, not natural)

bscheshirwork:
So... I try to use it with reverse
using `createRelationQuery` is better way.

samdark:
Won't make it into 2.0.

PR #14358, opened 2017-06-28
Both offset and limit can set with a Pagination instance or an array …

a761208:
Before:
`$query->offset($pagination->getOffset())->limit($pagination->getLimit());`
After:
`$query->page($pagination);`
Or:
`$query->page([123, 20]);`
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/docs/guide/db-query-builder.md
+++ b/docs/guide/db-query-builder.md
@@ -473,6 +473,23 @@ If you specify an invalid limit or offset (e.g. a negative value), it will be ig
   statement that emulates the `LIMIT`/`OFFSET` behavior.
 
 
+### [[yii\db\Query::page()|page()]] <span id="page"></span>
+
+The [[yii\db\Query::page()|page()]] method specify both the `LIMIT` and `OFFSET` fragments of a SQL query at one time.
+For example,
+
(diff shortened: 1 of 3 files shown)


dynasource:
Im against such shortcuts. The added value does not weigh up against to facts like:
* perhabs code is shorter but harder to read
* framework pollution
* more ways of doing things adds confusion to developers

samdark:
Same here. I don't see why it's needed. That's why I've asked for use case.

a761208:
Sorry to hear this.
I use active query search things like this:
```php
                                  $query = SomeModel::find();
                                  $query->... add some search filters
                                  $pagination = new Pagination(['totalCount'=>$query->count()]);
                                  $model_list = $query->orderBy(...)->offset($pagination->offset)->limit($pagination->limit)->all(); // It's a little long about this line
                               // $model_list = $query->orderBy(...)->page($pagination)->all();
```

Renkas:
I don't get it. Why are you using Pagination in this usecase at all?

PR #14374, opened 2017-07-01
Small fix phpDoc

lav45:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  |

### DIFF
--- a/framework/db/ActiveRecord.php
+++ b/framework/db/ActiveRecord.php
@@ -557,7 +557,7 @@ class ActiveRecord extends BaseActiveRecord
      * or [[beforeSave()]] stops the updating process.
      * @throws StaleObjectException if [[optimisticLock|optimistic locking]] is enabled and the data
      * being updated is outdated.
-     * @throws \Exception in case update failed.
+     * @throws \Exception|\Throwable in case update failed.
      */
     public function update($runValidation = true, $attributeNames = null)
     {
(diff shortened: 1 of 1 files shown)


samdark:
Hmm... it's definitely exception, isn't it?

samdark:
I don't think this change is needed.

LAV45:
@samdark , open this code in PhpStorm and you will see for yourself
```php
     /**
     * ...
     * @throws \Exception|\Throwable in case insert failed.
     */
    public function insert($runValidation = true, $attributes = null)
    {
        try {
            // ...
        } catch (\Exception $e) {
            // ...
        } catch (\Throwable $e) {
            // ...
        }
    }
```

samdark:
Merged. Thanks!

rob006:
IMO the whole `@throws` section should be removed in this case - almost every method could throw `Exception` or `Throwable`, annotating such general exception without any description what it means is pointless and only generates unnecessary noise.

samdark:
I don't agree. One should be aware than exception may occur in particular case.

rob006:
Can you name any method in ActiveRecord (except `transactions()` :)) that you sure that it will never throw any exception?

samdark:
Yes. Why?

- getDb
- find
- tableName
- equals
- isTransactional

rob006:
All of them will throw an exception if  db connection can not be established. And obviously on misconfiguration or if `Yii::$app` was not initialized.

Issue #14379, opened 2017-07-03
pdo exception on having virtual column in orderby clause


gunnrryy:
Fatal error: Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'sla_percentage1' in 'order clause' in

code:
```
$cases = Self::find()
            ->select('support_cases.*')
            ->addSelect([new \yii\db\Expression("ROUND((TIMESTAMPDIFF(SECOND,request_date, NOW())/sla)*100) as sla_percentage")])
    ->orderBy(['sla_percentage1' => SORT_DESC]);
```

`echo $cases->createCommand()->rawSql;`
generates query :

> SELECT `support_cases`.*, ROUND((TIMESTAMPDIFF(SECOND,request_date, NOW())/sla)*100) as sla_percentage1 FROM `support_cases`  ORDER BY `sla_percentage1` DESC

which is working fine in mysql cli and phpMyAdmin.
but when querried from Yii application, getting FATAL Error.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.9
| PHP version      |  PHP 7.0.19-1+deb.sury.org~xenial+2 (cli) (built: May 22 2017 12:48:42) ( NTS )
| Operating system | Linux Mint 18.1

gunnrryy:
on debugging further, 
in the yii\db\Query->queryScalar()

queryScalar() in the condition which checks for aggregation function like groupBy, having is causing this issue.
the query being generated is:
SELECT COUNT(*) FROM `support_cases`  ORDER BY `sla_percentage` DESC
```

if($selectExpression !== "COUNT(*)") {
    $this->select = [$selectExpression];
}
```
seems to solve the issue.

gunnrryy:
issue occurs when passing complex query for  pagination as in the example given in Pagination class.

terabytesoftw:
This was reported against Yii2 `2.0.9`. The underlying `Query::count()` / `queryScalar()` handling for queries with `orderBy` was fixed long ago (see #13846), and the current `master` branch still clears `orderBy`, `limit`, and `offset` before executing scalar aggregate queries. `ActiveDataProvider` also resets ordering before calculating `totalCount`.

I re-checked the pattern from this report on the current Yii2 master (March 2026) with a read-only MySQL reproduction (`select + addSelect(Expression alias) + orderBy(alias) + clone + count()`), and it is no longer reproducible.

Closing as no longer relevant.

PR #14384, opened 2017-07-04
queryScalar correction

gunnrryy:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | -
| Tests pass?   | -
| Fixed issues  | #14379

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -418,8 +418,10 @@ class Query extends Component implements QueryInterface
             $order = $this->orderBy;
             $limit = $this->limit;
             $offset = $this->offset;
-
-            $this->select = [$selectExpression];
+            
+            if ($selectExpression !== "COUNT(*)") {
+                $this->select = [$selectExpression];
+            }
             $this->orderBy = null;
             $this->limit = null;
             $this->offset = null;


gunnrryy:
removed earlier PR as suggested for maintaining code format.

samdark:
Thanks for fixing formatting. Would be cool to add unit tests as mentioned by our @yii-bot.

gunnrryy:
don't know how to unit test. searched alot but didnt find any understandable source.

can you guide me to some source where i can see examples, in reference with Yii2 unit test.
new to this paradigm and eager to learn.
Thanks in advance.

gunnrryy:
till then, ill try to explain the scenario here.

when ActiveQuery()->count() is called without passing parameter, the 1st parameter being passed to queryScalar() is COUNT(*) by default values. 

when using ORDER BY clause on virtual column, virtual column is not found in * and hence FATAL ERROR.
Since any column that we want to apply aggregate function or Orderby, we need to have that column in select mandatory.

if evaluated from $this->select, the virtual column is calculated from select expression and avoids the issue.

solution: 
if any parameter is passed in ActiveQuery()->count('cases.id'), the parameter being passed to queryScalar is COUNT("cases.id") and evaluated corresponding. if nothing is passed and virtual column is used, current code breaks.

can be handled by checking the 1st parameter passed to queryScalar being default or no to whether evaluate from $this->select or directly by *, the issue seems to  resolve.

gunnrryy:
my bad, 
this doesn't solve it completely. just the evaluation of query part worked by adding the mentioned condition.

but this creates  issue with pagination. coming round from where we started.

for now, a work around that i applied is, 

```
$countQuery = clone $support_case_model;
$countQuery->orderBy = '';
$pages = new Pagination(['totalCount' => $countQuery->count()]);
```

by this, i dont get the error of virtual column and pagination also works correctly.
NOTE: this is not a fix. just a work around.
ISSUE IS STILL THERE FOR HAVING  VIRTUAL COLUMN IN ORDER BY CLAUSE OF QUERY.
issue is STILL there for having **VIRTUAL COLUMN in ORDER BY CLAUSE**

PR #14396, opened 2017-07-05
Removed call to deleteAll from deleteInternal

markux:
in this manner we removed the hidden relation between delete and deleteAll. Example: if we have to override deleteAll and call delete for single model (and trigger delete event)

### DIFF
--- a/framework/db/ActiveRecord.php
+++ b/framework/db/ActiveRecord.php
@@ -607,7 +607,7 @@ class ActiveRecord extends BaseActiveRecord
         if ($lock !== null) {
             $condition[$lock] = $this->$lock;
         }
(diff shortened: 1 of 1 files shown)


cebe:
Pretty sure this will break the AR implementation of non-SQL dbms. If you need to trigger events on delete you should not use `deleteAll()` but use another method for that.

markux:
Ok, but
https://github.com/yiisoft/yii2-mongodb/blob/master/ActiveRecord.php#L324
It not use deleteAll into deleteInternal, and it's correct. 
Only yii2-redis override deleteAll without change deleteInternal. 
I think it's a dangerous relation. IMHO. bye bye

Issue #13979, opened 2017-04-13
Runtime change of `Validator::$attributes` has no effect


klimov-paul:
### What steps will reproduce the problem?

```php
$model = new SomeModel();
$validator = new EmailValidator();
$validator->attributes = ['email'];
$model->getValidators()->append($validator);

$model->email = 'invalid';
$this->assertFalse($model->validate()); // faliure
```

### What is the expected result?

Dynamically created validator with `attributes` set being applied

### What do you get instead?

Validator is ignored

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12 (master)
| PHP version      | 5.6
| Operating system | CentOS

This behavior has been introduced by PR #13088, which makes Validator apply fltering via `attributeNames` private field, which can not be adjusted after Validator creation.

PR #14413, opened 2017-07-08
fix multiple calls getAttributeNames() for same object

nicdnepr:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

It's fix for this code call https://github.com/yiisoft/yii2/blob/master/framework/widgets/ActiveField.php#L241
For each attribute we call this part of code https://github.com/yiisoft/yii2/blob/master/framework/base/Model.php#L427

### DIFF
--- a/framework/validators/Validator.php
+++ b/framework/validators/Validator.php
@@ -190,6 +190,8 @@ class Validator extends Component
      */
     public $whenClient;
 
+    private $_attributeNames;
+
 
     /**
      * Creates a validator object.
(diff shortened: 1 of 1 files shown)


dynasource:
whats the usecase for this PR? When you are doing a validation twice?

nicdnepr:
For each attribute calls the full list of validators and each time calls getAttributeNames

dynasource:
it does not seem to be very relevant with respect to performance gain. It does seem to have some use when you execute validate() twice on the same model, but that is a rare occasion, and mainly in the testing environment.

klimov-paul:
This can not be accepted as runtime change of `Validator::$attributes` will have no effect.
See #13979

cebe:
We should probably add a test that breaks in case such change is made...

PR #14419, opened 2017-07-09
Enable `heredoc_to_nowdoc` rule in php-cs-fixer

rob006:
Continuation of #14100
See https://github.com/FriendsOfPHP/PHP-CS-Fixer

http://php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc


samdark:
I think it should be other way around...

rob006:
@samdark Could you elaborate? Changing heredoc to nowdoc is the same kind of change as changing quoting from `"` to `'` (which was already accepted in #14100) - it is a little bit faster and more readable/safe (less risk of accidental variable parsing).

samdark:
Ah. Right. Then it's good.

samdark:
Merged. Thanks!

PR #14420, opened 2017-07-09
Enable `no_useless_else` rule in php-cs-fixer

rob006:
Continuation of #14100
See https://github.com/FriendsOfPHP/PHP-CS-Fixer

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -328,9 +328,9 @@ class Formatter extends Component
         $method = 'as' . $format;
         if ($this->hasMethod($method)) {
             return call_user_func_array([$this, $method], $params);
-        } else {
-            throw new InvalidParamException("Unknown format type: $format");
         }
+
+        throw new InvalidParamException("Unknown format type: $format");
     }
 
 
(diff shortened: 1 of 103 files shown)


samdark:
Merged. Thanks.

framework/i18n/Formatter.php
```diff
@@ -826,28 +827,30 @@ public function asRelativeTime($value, $referenceTime = null)
             if ($interval->s == 0) {
                 return Yii::t('yii', 'just now', [], $this->locale);
             }
+
             return Yii::t('yii', 'in {delta, plural, =1{a second} other{# seconds}}', ['delta' => $interval->s], $this->locale);
-        } else {
```
cebe: in cases like this, else would be much easier to understand... :-(

cebe:
In general I am fine with improving code style and also thanks for working on this @rob006 but at this scale its massivly breaking git blame, not sure its worth the pain... Maybe its enough for now?

framework/i18n/Formatter.php
rob006: It would be easier to understand if we extract these blocks to separate methods.

rob006:
@cebe I want to add few more fixers, but they should result less changes than this one. 

But if we want to force some code standards, we should do it now or never. The later we do this, the more changes it will require, give a greater impact on git blame and will be harder to merge changes between branches.
And I'm not sure if git blame is i worth of dropping automated CS fixing. Influence on git blame is temporary, with bad CS we will live forever :).

samdark:
I'm for consistent code style.

Issue #14254, opened 2017-06-05
ExistValidator may not work properly in master-slave replication environment


yanhuixie:
### What steps will reproduce the problem?
1) works with a master-slave replication database cluster
2) for some reason disable the Transaction 
3) AR Foo has an auto increment primary key "id"
4) AR Bar has a foreign key related to Foo->id as "foo_id" and guarded by exists validator (generated by gii)
5) Save Foo first, then new a Bar bar and set bar->foo_id  = foo->id, and bar->save()
At that time my be an error occurs (OR NOT) : "Foo ID is invalid." And foo is saved but bar not.

### What is the expected result?
No exists validator error occurs, foo and bar both saved normally.

### A temporary solution
disable the exists validator of Bar.

### Additional info
I think it's not means a bug, but an announcement for the developer. Because it's unavoidable that there is a  lag between the replication process from the master node to the slave nodes. Without the guarantee of high level database transaction, the EXISTS VALIDATOR may query a "NEW" foo from an unsynchronized slave node.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.0.17
| Operating system | OSX or Linux

rossoneri:
We can add an option for ExistValidator to spcify using the master db, how about that? @yanhuixie @samdark

yanhuixie:
Thanks for your reply, @rossoneri ! I think it's a good idea. 
It's reliable that write and read the "foo" from the same db node.
What we should remind (in documentation?) is that there are some premises to do so.
1) config correctly in Yii2 'db' component (master and slaves).
2) the db nodes may NOT behind a proxy middle-ware/load balancer or config it works as read/write splitting mode, but not multi-master.

samdark:
@rossoneri yes, that should be OK.

rossoneri:
@samdark I will do it!

PR #14424, opened 2017-07-10
Enh #14254:add an option to specify whether validator can run on slav…

qihuajun:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #14254


cebe:
can't we just always do it on the master?

rossoneri:
@cebe I don't know, maybe someone want to use slave connections if they have no lag problems.

samdark:
There are always lag problems. They're unavoidable.

rossoneri:
@samdark even with lag, the slave db still can be used for validation, that depends the application's concrete situation.

samdark:
Sure.

framework/validators/ExistValidator.php
```diff
@@ -73,6 +73,11 @@ class ExistValidator extends Validator
      */
     public $targetAttributeJunction = 'and';
 
+    /**
+     * @var bool whether this validator can run on salve connections
+     */
+    public $canValidateOnSlaveDb =  true;
```
samdark: "can" is ambiguous. It means validator can run on slave but it doesn't specify it would run on slave. That's confusing.

rossoneri: > It means validator can run on slave but it doesn't specify it would run on slave. 

Yes, that's what I mean. 

First of all, the purpose of this option is to stop using salve db when validating.  If you set this option to false, you can do it. 

And considering that in the case there is no slave db config, so you can't specify to run a validator on slave db.

rob006: `$useMasterDb`?

cebe: Also I think it should use master connection by default, as that is where the insert will be made it will be safe by default and can be changed if querying slave is okay.

samdark: Agree. Using master is a good default.

rossoneri: Good Point！

rossoneri: @samdark the default value now is set to false, that mans validation runs on master by default

samdark:
@rossoneri seems tests are broken with the changes in PR. Would you please take a look?

rossoneri:
@samdark solved!

samdark (approved):

framework/validators/ExistValidator.php
```diff
@@ -110,13 +116,24 @@ public function validateAttribute($model, $attribute)
         $targetClass = $this->targetClass === null ? get_class($model) : $this->targetClass;
         $query = $this->createQuery($targetClass, $conditions);
 
+        $db = $targetClass::getDb();
+        $disabledSlaves = false;
+        if (!$this->canValidateOnSlaveDb && $db->enableSlaves){
+            $db->enableSlaves = false;
+            $disabledSlaves = true;
+        }
```
rob006: Why not use `Connection::useMaster()`?

SilverFire: The code looks messy now

rossoneri: @rob006 I did think of using `Connection::useMaster()`, but if you consider the situation when `canValidateOnSlaveDb` is true, you'll find it's not a good choice.

rob006: > but if you consider the situation when canValidateOnSlaveDb is true, you'll find it's not a good choice.

Why? Obviously you need some conditions here (use `Connection::useMaster()` only when `canValidateOnSlaveDb` is `false`), but you can move it to separate method.

framework/validators/ExistValidator.php
```diff
@@ -73,6 +73,12 @@ class ExistValidator extends Validator
      */
     public $targetAttributeJunction = 'and';
 
+    /**
+     * @var bool whether this validator can run on salve connections
+     * @since 2.0.14
+     */
+    public $canValidateOnSlaveDb =  false;
```
rob006: `$useMasterDb`? https://github.com/yiisoft/yii2/pull/14424#discussion_r127182082

framework/validators/ExistValidator.php
rossoneri: Yes, that would make it too, but I my code used same logic within `Connection::useMaster()`, and it's simpler.

framework/validators/ExistValidator.php
samdark: I'd say `$forceMasterDb`.

framework/validators/ExistValidator.php
rob006: @rossoneri It does not handle exceptions correctly and at the best case it is just code duplication.

framework/validators/ExistValidator.php
rossoneri: changed to $forceMasterDb and take useMaster to run on master db.

SilverFire (changes requested):
I see copy-pasted code in `ExistValidator::validateAttribute()` and `::validateValue()` methods. @rossoneri we could move it to a separate method

framework/validators/ExistValidator.php
```diff
@@ -73,6 +74,11 @@ class ExistValidator extends Validator
      */
     public $targetAttributeJunction = 'and';
 
+    /**
+     * @var bool whether this validator are forced to  run on master db
```
samdark: is forced to always use master DB

framework/validators/ExistValidator.php
```diff
@@ -110,11 +116,10 @@ public function validateAttribute($model, $attribute)
         $targetClass = $this->targetClass === null ? get_class($model) : $this->targetClass;
         $query = $this->createQuery($targetClass, $conditions);
 
-        if (is_array($model->$attribute)) {
-            if ($query->count("DISTINCT [[$targetAttribute]]") != count($model->$attribute)) {
-                $this->addError($model, $attribute, $this->message);
-            }
-        } elseif (!$query->exists()) {
+        $value = $model->$attribute;
+        $existed = $this->checkExisted($targetClass,$query,$value);
```
samdark: No need for extra variable, could be checked right in the `if`.

samdark: Formatting isn't right, please add spaces after `,`.

framework/validators/ExistValidator.php
```diff
@@ -178,17 +183,56 @@ protected function validateValue($value)
...
+
+
+    /**
+     * Run query to check value existed
+     *
+     * @param $query QueryInterface
+     * @param $value mixed the value want to be checked
+     * @param $existed bool
+     */
+    private function checkValueExisted($query, $value,&$existed){
```
samdark: It doesn't return anything...

samdark: queryValueExists

framework/validators/ExistValidator.php
```diff
@@ -178,17 +183,56 @@ protected function validateValue($value)
...
+
+    /**
+     * Check whether value exist in target table
+     *
+     * @param $targetClass
+     * @param $query QueryInterface
+     * @param $value mixed the value want to be checked
+     * @return boolean
+     */
+    private function checkExisted($targetClass, $query,$value ){
```
samdark: valueExists

samdark (changes requested):

samdark:
Seems to fail tests now: https://travis-ci.org/yiisoft/yii2/jobs/338387216

rossoneri:
@samdark tests passed now!

framework/validators/UniqueValidator.php
```diff
@@ -131,7 +138,19 @@ public function validateAttribute($model, $attribute)
             $conditions[] = [$key => $value];
         }
 
-        if ($this->modelExists($targetClass, $conditions, $model)) {
+        $db = $targetClass::getDb();
+
+        $modelExisted = false;
```
samdark: Existed → Exists

samdark (approved):

samdark:
Yes, additional tests that check if master is preferred won't hurt. Any idea on how to test that efficiently, @rob006 ?

rob006:
@samdark Use invalid config for slave connection and run validation? If it pass, it means that master was used.

samdark:
Sounds good.

samdark:
Finished and merged. Thank you for contribution.

Issue #13854, opened 2017-03-24
Compress Log File Rollover


Knight-Yoshi:
Yii2 should have the option to compress rollover files.

samdark:
Rollover files are the ones produced by file log target when primary file exceeds maximum size.

Knight-Yoshi:
Yeah, and there should be an option to compress the files when they rollover to take up less disk space.

samdark:
I think it makes sense to have it.

letsjump:
As many operating systems has their own configurable _log file rotator_, we could write just write a widget that do the work, so Yii core doesn't grows too much.
If you are on a Debian SO with shell access, for example, you can attach a logrotate by editing `/etc/logrotate.conf`:
`/path/to/my/yiiApp/runtime/logs/*.log {
        weekly
        rotate 52
        delaycompress
        notifempty
        compress
        create 640 owner group
}
`
for more info on logrotate, [read its documentation](http://www.linuxcommand.org/man_pages/logrotate8.html)

PR #14433, opened 2017-07-11
Logfile compression

cebe:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13854

best to be reviewed in two steps: 7b89162 + 3fa3ba6


klimov-paul:
Why `*.gz` only?
if you wish to support archiving of the files there should be option to use any archive format `*.tar`, `*.zip`, `*.rar` etc.
Hardcoding single `*.gz` can not satisfy all needs.

klimov-paul:
Would't it be better to simply introduce an event or callback, which should be triggered after particular file has been rotated, so developer may choose, what to do with it?
For example, besides compression, some people tranfer log file to separated file storage server.

cebe:
> Why *.gz only?

because it is directly supported by PHP, other methods would need more complex implementation or external tools.



> Hardcoding single *.gz can not satisfy all needs.

this is a start for supporting compression, `.gz` is the most common one for logs. it can be extended later.



> Would't it be better to simply introduce an event or callback, which should be triggered after particular file has been rotated, so developer may choose, what to do with it?

you can override `compressFile()` method if you want to customize behavior



> For example, besides compression, some people tranfer log file to separated file storage server.

There are tons of different ways to do this. and dependent on the case a spearate log target should be created for that.

klimov-paul:
> There are tons of different ways to do this. and dependent on the case a spearate log target should be created for that.

I can say just the same thing for compression feature.

klimov-paul:
> this is a start for supporting compression, .gz is the most common one for logs. it can be extended later.

That is the point: it is a start, but where it will lead? Starting support of compression log files we can end in countless request for its expansion for many purposes.
Framework either should provide sophisticated level for file archiving or do not provide it at all.

dynasource:
based on the fact that the framework already covers log rotation, it makes sense to include compression.  It also makes sense to provide such standard GZ compression as a default.

rob006:
We could extract compressor to separate component - basic implementation (gzip) will be provided by framework, other options may be provided by extensions.

framework/log/FileTarget.php
```diff
@@ -137,25 +142,70 @@ protected function rotateFiles()
...
+            } else {
+                $this->rotateFile($rotateFile, $rotateTarget);
+            }
+        }
+    }
+
+    /**
+     * Rotate a single file
+     * @param string $file the file to rotate.
+     * @param string $target the target file name.
```
samdark: Need `@since`.

framework/log/FileTarget.php
```diff
@@ -137,25 +142,70 @@ protected function rotateFiles()
...
+     * @since 2.0.13
+     */
+    protected function compressFile($filename)
+    {
+        if (substr($filename, -3, 3) === '.gz') {
+            return $filename;
+        }
+        $gzFilename = $filename . '.gz';
+        $f = fopen($filename, 'rb');
+        $gz = gzopen($gzFilename, 'wb9');
```
samdark: How about making compression rate and filtration flags configurable?

framework/log/FileTarget.php
```diff
@@ -137,25 +142,70 @@ protected function rotateFiles()
...
+        while(!feof($f)) {
```
samdark: Formatting...

framework/log/FileTarget.php
```diff
@@ -137,25 +142,70 @@ protected function rotateFiles()
...
+            gzwrite($gz, fread($f, 8192));
```
samdark: What is 8192?

framework/log/FileTarget.php
```diff
@@ -137,25 +142,70 @@ protected function rotateFiles()
...
+    }
+
+
+    /**
+     * Compresses a file with gzip and renames it by appending `.gz` to the filename
+     * @param string $filename the file to compress.
+     * @return string the name of the compressed file.
+     * @since 2.0.13
+     */
+    protected function compressFile($filename)
```
samdark: Is it possible that multiple processes call `compressFile` for the same file concurrently?

samdark (changes requested):

framework/log/FileTarget.php
cebe: default chunk size from the fread docs.

framework/log/FileTarget.php
```diff
@@ -137,25 +142,70 @@ protected function rotateFiles()
         for ($i = $this->maxLogFiles; $i >= 0; --$i) {
             // $i == 0 is the original log file
             $rotateFile = $file . ($i === 0 ? '' : '.' . $i);
-            if (is_file($rotateFile)) {
-                // suppress errors because it's possible multiple processes enter into this section
-                if ($i === $this->maxLogFiles) {
-                    @unlink($rotateFile);
+            $rotateTarget = $file . '.' . ($i + 1);
+            if (!is_file($rotateFile)) {
```
cebe: there is no else on this one :) only if the file does not exist, something needs to be done.

framework/log/FileTarget.php
cebe: yes, same as `rotateFile()`. As far as I see they would do the compression twice. needs to be checked.

framework/log/FileTarget.php
samdark: OK. I've missed extra condition inside looking at `continue` statement. Everything's fine.

samdark:
Does not seem we can fix it in time to get into 2.0.

PR #14441, opened 2017-07-12
Add `InputWidget::createInputField()`

### LINKED ISSUE #14294: An error in InputWidget's phpdoc?
Added as a part of the [MaskedInput widget implementation](https://github.com/yiisoft/yii2/commit/249f90f5ee78cee47f6fcb990ae10d2a56d24e1a) the phpdoc of the InputWidget says that if it is associated with a data model and an attribute, the name and the value will be generated automatically. But this is only true for the MaskedInput widget itself, as far as I can see (MaskedInput::run). If it really is an error this currently misleading info should be moved into the MaskedInput's phpdoc (which curiously doesn't say anything about this feature).
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | master
| PHP version      | 
| Operating system |

cebe:
to implement behavior described in class docs directly in the class
itself instead of relying on subclasses to respect the description.

fixes #14294

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14294


samdark (approved):

framework/widgets/InputWidget.php
```diff
@@ -87,4 +87,25 @@ protected function hasModel()
...
+     * This will call [[Html::activeInput()]] if the input widget is [[hasModel()|tied to a model]],
+     * or [[Html::input()]] if not.
+     *
+     * @param string $type the type of the input to create.
+     * @return string the HTML of the input field.
+     * @since 2.0.13
+     * @see Html::activeInput()
+     * @see Html::input()
+     */
+    protected function createInputField($type)
```
dynasource: isnt this ambigious as it renders a HTML tag and not actually creating a Field? Perhabs a better name is `renderInput()`

dynasource (changes requested):

framework/widgets/InputWidget.php
samdark: Good suggestion.

cebe: done.

dynasource (approved):

samdark:
Merged. Good one.

SilverFire:
Introducing `InputWidget::renderInput()` with `$type` as a required parameter causes declaration compatibility exceptions in PHP 7+

In my projects this change affected at least three widgets:

https://github.com/hiqdev/yii2-combo/commit/af68b1577b830c52048903fe7d02abc4215b0677
https://github.com/kartik-v/yii2-widget-datepicker/blob/master/DatePicker.php#L297
https://github.com/omnilight/yii2-bootstrap-daterangepicker/blob/master/DateRangePicker.php#L114

cebe:
Ooops ... @SilverFire can you open an issue?

cebe:
reopened https://github.com/yiisoft/yii2/issues/14294#issuecomment-315144541

PR #14443, opened 2017-07-12
Added sysexits codes in ExitCode class for console

### LINKED ISSUE #14188: Add constants and function for sysexits(3) to ConsoleHelper

cebe:
follow up to #14189
fixes #14188

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14188

/cc @tom--


tom--:
Very nice.

Thanks @cebe for doing this.

Thanks @samdark for suggesting a dictionary class.

samdark (approved):

Kolyunya:
@cebe @samdark what is the rationale behind introducing platform specific exit codes in a platform-independent framework which are not encapsulated in something like a `Platform` class?

tom--:
The codes are not platform-specific. They can be used on any platform and have been adopted by other popular platform-independent frameworks. See comments to #14189 for explanation.

Kolyunya:
@tom-- could you please provide any references to say official Windows docs listing those exit codes?

tom--:
@Kolyunya no, because they are platform-independent. It makes no difference what operating system you're on. They work equally well on all systems that run PHP.

Kolyunya:
@tom-- I can indeed provide a reference to official Windows docs listing standard exit codes and that proves that exit code **definitions** are actually platform-specific.

```diff
+     * The command was used incorrectly, e.g., with the wrong number of
+     * arguments, a bad flag, a bad syntax in a parameter, or whatever.
+     */
+    const USAGE = 64;
```

Exit code `64` [means](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx) 

> The specified network name is no longer available.

on Windows.

tom--:
@Kolyunya That's a code from an unrelated namespace used [in response to OS system calls](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681382(v=vs.85).aspx). The exit codes here describe how a user-space process exits, on which Windows uses the same standard as Linux and Unix, i.e. zero for success and everything else means an error.

Kolyunya:
@tom-- would you be so kind to provide any references proving that those **_exit code definitions_** are standard, platform-independent and have the same meaning across all supported platforms?

tom--:
@Kolyunya It's a *de facto* standard, meaning it's a convention as old as the hills that everyone uses including Yii.

I don't know if this is written down in Windows documentation and it wouldn't make any difference if it were.

I am not trying to convince you, which I don't believe is possible. My answers are for the record and for anyone else who may be interested.

Kolyunya:
@tom-- I see, so you can not provide any references to written documents proving that those exit code **_definitions_** are standard and adopted by all supported platforms.

dynasource (approved):

samdark:
@Kolyunya these are not standard. These are suggestions. The only written standard is 0 for success, the rest are errors. The rest is decided by CLI utility author. Yii as framework aids in this providing error codes framework.

Kolyunya:
@samdark well at least we finally came to a conclusion that those code definitions actually are platform specific. My original question was: why are they a part of a `ExitCode` instead of a `BsdExitCode`?

sergeymakinen:
I think it's because there's basically no error code standard for Windows but we can't throw `null` because there's no Windows' code equivalent. And BSD's one looks okay for this purpose.

samdark:
These are conventions and are OK to use either in BSD, Linux or Windows. According to specs, noone should rely on concrete exit codes without actually reading utility docs.

framework/console/ExitCode.php
```diff
@@ -0,0 +1,154 @@
+<?php
```
klimov-paul: File header is missing.

framework/console/controllers/BaseMigrateController.php
```diff
@@ -287,7 +288,7 @@ public function actionRedo($limit = 1)
         if (empty($migrations)) {
             $this->stdout("No migration has been done before.\n", Console::FG_YELLOW);
 
-            return self::EXIT_CODE_NORMAL;
+            return ExitCode::OK;
```
klimov-paul: Maybe it is actual better to use a predefined method in `console\Controller` for `OK` and `Error`:

```php
public function actionFoo()
{
    // ...
    return $this->exitOk();
}
```

To hide constants behind the scenes.

This would be consistent with recent addition of `web\Controller::asJson()` method.

klimov-paul (changes requested):

framework/console/controllers/BaseMigrateController.php
cebe: asJson() is actually doing something, `exitOk()` would only return a fixed value. Also considering the amount of different values to choose from for an exit code I do not think a method makes sense. You can define your own exit code constants easily for returning them, defining a method for each of them does not make sense imo.

samdark: I agree with @cebe on this regard.

klimov-paul (approved):

samdark:
Merged. Thanks for taking care of it, @cebe.

tom--:
These codes are FreeBSD-specific in the same way that git is Linux-specific because it *originated* in the kernel project. Also, git and sysexits both concern themselves exclusively with user-space software and the interaction between user-space processes.

But I think this description is confusing because "platform-specific" suggests interoperability or portability difficulties.

PR #14444, opened 2017-07-12
Fixes #14423, wrong checking existing keys in ArrayHelper::merge with…

### LINKED ISSUE #14423: ArrayHelper::merge() replaces NULL with value from latter array (or should it?)
        $array1 = [
            1,
            null,
            3,
            null
        ];

        $array2 = [
            'a',
            'b'
        ];

        $array3 = [
            null,
            null,
            'c',
            'd'
        ];

        print_r(ArrayHelper::merge($array1, $array2, $array3));

This is a result produces from array_merge() PHP native function
>Array
>(
>    [0] => 1
>    [1] => 
>    [2] => 3
>    [3] => 
>    [4] => a
>    [5] => b
>    [6] => 
>    [7] => 
>    [8] => c
>    [9] => d
>)

### What do you get instead?
The function replaces `NULL` with with value from latter array. So `$array1[1]` gets replaced by `$array2[1]`.
>Array
>(
>    [0] => 1
>    [1] => b
>    [2] => 3
>    [3] => d
>    [4] => a
>    [5] => 
>    [6] => 
>    [7] => c
>)

### Possible cause ###
In `BaseArrayHelper::merge()`, on line 128 - 134.

                } elseif (is_int($k)) {
                    if (isset($res[$k])) {
                        $res[] = $v;
                    } else {
                        $res[$k] = $v;
                    }
                }

`isset()` will return `FALSE` on `NULL`. So if the former array has a `NULL` value, it will replace it with the value from the latter array.

However, from the description of the function
> For integer-keyed elements, the elements from the latter array will be appended to the former array

So these means the integer-keyed elements will always get appending. Unless it is not yet defined.

The code can be changed code into something like:

                } elseif (is_int($k)) {
                    if (array_key_exists($k, $res)) {
                        $res[] = $v;
                    } else {
                        $res[$k] = $v;
                    }
                }

I'm not fully testing it yet, so I'm not sure about the side effect of this code. However, is it actually alright to be like this?

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11-dev (commit 7fcf75f to be exact)
| PHP version      | 5.6.30
| Operating system |Windows 10

dmirogin:
Wrong checking existing keys in ArrayHelper::merge with null values for integer-keyed elements

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14423


dynasource (changes requested):
thanks for improving the merge function. Could you limit the size of the test in which your usecase is still covered correctly?

samdark (approved):

dmirogin:
Yeah, sure, made shorter, but still covered.

dynasource (approved):

samdark:
Thanks!

PR #14447, opened 2017-07-13
Fix phpdoc for $_master and $_slave properties in yii\db\Connection

rob006:
This fixes issues reported by [Scrutinizer](https://scrutinizer-ci.com/g/yiisoft/yii2/inspections/45e312ad-cd29-42b7-8314-78a9930394d3/issues/files/framework/db/Connection.php?status=new&orderField=path&order=asc&honorSelectedPaths=0).

### DIFF
--- a/framework/db/Connection.php
+++ b/framework/db/Connection.php
@@ -403,11 +403,11 @@ class Connection extends Component
      */
     private $_driverName;
     /**
-     * @var Connection the currently active master connection
+     * @var Connection|false the currently active master connection
      */
     private $_master = false;
     /**
-     * @var Connection the currently active slave connection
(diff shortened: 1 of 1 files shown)


rob006:
Also [this condition](https://github.com/rob006/yii2-dev/blob/25f1f57b1c9e0b06837385d8e7b33569c4020887/framework/db/Connection.php#L949) is probably unnecessary - right now `$_slave` is never set to `null`.

cebe:
> Also this condition is probably unnecessary - right now $_slave is never set to null.

`openFromPool` can return `null`, in that case it is set to `null`.

cebe:
Thanks!

Issue #14421, opened 2017-07-10
Rendering module forms outside of module


vercotux:
1. Take any module with a `'_form'` view containing an ActiveForm.
2. Render module's `'_form'` **outside** of module (eg. in main app `site/index` view). For example: `$this->render('@somevendor/somemodule/frontend/views/post/_form', ['model'=>$model]);`

The rendered ActiveForm `'action'` URL in our module's '_form' will be incorrect no matter what we do:

 * If the form uses a relative route (eg. `['post/create']`) then the URL will be pointing to the wrong controller (app's instead of module's).
 * If the form uses a fixed absolute route (eg. `['//module/post/create']`) then the URL will be pointing to the *potentially* wrong module ID (it could be called something other than "`module`").
 * If we try to dynamically grab the module's ID (eg. by doing `['//'.Module::getInstance()->id.'/post/create']`) we will end up with a *"Trying to get property of non-object"* exception.

The mechanism to do this seems to be missing and/or is not documented.

vercotux:
This isn't a question. It's a request to fix Yii's sad module architecture.

samdark:
Sorry. Hit the wrong label...

vercotux:
No problem.

To clarify a bit more. There is a fundamental problem with the module architecture.

Look at the workarounds module developers are forced to resort to just to get their code to work:
https://github.com/yii2mod/yii2-comments/blob/aae211a6e93b8e06ced1218441a6dc845efec464/traits/ModuleTrait.php

samdark:
I think it would be something like:

```php
$this->controller->module->id . '/controller/action'
```

vercotux:
Sadly, that only works *inside* the modules. Modules also come with widgets and other components which may need to be used externally.

samdark:
Then these aren't part of modules.

vercotux:
Well they still rely on the modules. But they cannot find get to the modules they were shipped along with without knowing in advance what the ID will be. Do you not see the problem?

samdark:
In your original example you're rendering a view by providing a path to it. You're not giving any context to what this view belongs to or anything so you have to supply module's ID explicitly. For example, like this:

```php
$this->render('@somevendor/somemodule/frontend/views/post/_form', [
    'model'=>$model,
    'action' => ['/moduleID/controllerID/actionID'],
]);
```

samdark:
Not sure how you'd like to make it better...

rob006:
[...]

PR #14456, opened 2017-07-14
Fix #5786 via static model

### LINKED ISSUE #5786: Allow using custom constructor with non-optional arguments in AR-based models

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues   | #5786, #13779, #14078

Alternative approach for #14078


rob006:
Maybe rename `model()` to `instance()`?

samdark:
Agree with @rob006 suggestion.

samdark:
If it fixed #5786 there should be a test case validating that constructor with custom arguments works well.

ElisDN:
Take this test cases from my #14078 please.

rob006:
If we agree with using `instance()` as a method name, we could extract it into trait and interface - `Model` is not the only class that can benefit from static instance of itself. We already have `Module::getInstance()` that work similar, but with [some issues](https://github.com/yiisoft/yii2/issues/14421). Personally I created something similar at least 2 times in my projects. It could be more general convention, not only model-specific method.

samdark:
Good idea, @rob006.

klimov-paul:
- `StaticInstanceInterface` extracted
- unit tests for ActiveRecord with constructor added

samdark (approved):

samdark:
@ElisDN any comments on the pull request?

ElisDN:
The `Module::getInstance()` singleton method is actual for module objects only (which are singletons), but it is not actual for instantiating of non-singleton classes.

I did not assume non-model usage of the method. I originally meant internal instantiating like `createInternalModelInstance()` instead of using it with cache like `getSingletonInstance()`. But in nice current implementation with caching for getting meta information only it will more descriptive to name the method as `getMetaInstance` instead of dry `instance` name.

ElisDN:
And I would replace:

```php
self::$_instances[$className] = Yii::createObject($className);
// to
self::$_instances[$className] = new $className;
```

and:

```php
$record = Yii::createObject($viaClass);
// to
$record = $viaClass::instance(false);
// or
$record = $viaClass::instantiate($columns);
```

for avoiding of DIC usage for model classes.

klimov-paul:
In that case the issue #13779 will not be solved.

samdark:
@ElisDN I don't get what is "meta" in "getMetaInstance" name.

samdark:
Overall it looks good to me. Need additional reviews.

ElisDN:
@samdark It is on base of @klimov-paul's description about `meta information`:

```php
interface StaticInstanceInterface
{
    /**
     * Returns static class instance, which can be used to obtain meta information.
     * ...
     */
    public static function instance($refresh = false);
}
```

samdark:
Well, it *can* be used for obtaining meta information but isn't limited to that purpose.

ElisDN:
@samdark, Right now there are two internal purposes for instantiating:

- Obtaining meta information. We can use cache.
- Instantiating `$viaClass` pivot model in `link()` method. We cannot use cache.

For working with both purposes we can:

- use current `instance(false)` method with caching for getting meta information and `instance(true)` one without caching for instantiating of new pivot models. It miss responsibilities and has caching side effects.
- separate `metaInstance()` method for *getting cached instance* with meta information and `instance()` method with plain `new static()` code without caching for *internal instantiating* of pivot models. The second purpose is actual for AR only.
- leave one general `instance()` method without caching for creating new empty instance for both purposes.

samdark:
Personally I'd leave one method w/o caching and without any arguments. That's something I'd like to hear other @yiisoft/core-developers opinions about first.

rob006:
This interface does not have much sense without caching. The whole point is to have static instance, not new instance every time. How this suppose to work if we want to replace `Module::getInstance()` with `instance()` (#14421)?

Also, in my projects I'm using something like `Language::instance('en')` to get static instance of `Language` entity with some information about specified language. I would rather see two methods: `instance(mixed $data = null)` and `newInstance()`. `instance()` will cache instances, but also allow to store different instances of the same class.

ElisDN:
@rob006, It is not about replacing `Module::getInstance()` method. It is about internal Model instantiating method like 'createNewInstance` from #14078.

rob006:
@ElisDN It is about this two (and more) things. You don't need a separate interface if you want to fix only #14078.

klimov-paul:
> Instantiating $viaClass pivot model in link() method.

This will not be supported. When you put mandatory parameters into your ActiveRecord constructor, you introduce a rule, which say new instance of this class can not be created without these parameters specified. This means new model inside `link()` method can NOT be instantiated - if it will - it breaks the restriction you created.
If you say 'some parameters are required to create an instance', this means they are *always required*.
You'll have to implement custom `link()` on your own, or instantiate and save pivot model manually.

dynasource (approved):

samdark:
Merged! Thank you for handling it.

ElisDN:
@klimov-paul Thanks for PR!

> You'll have to implement custom link() on your own, or instantiate and save pivot model manually.

Okay. I agree. It is very nice.

Can you replace `Yii::createObject($viaClass)` back to `new $viaClass` and `Yii::createObject($className)` to `new $className`? For avoiding of using DIC for entities.

klimov-paul:
> Can you replace Yii::createObject($viaClass) back to new $viaClass and Yii::createObject($className) to new $className? For avoiding of using DIC for entities.

No, we can not, Since there is an issue #13779 as well fixed by this PR.

ElisDN:
@klimov-paul The #13779 and same issues have appeared on base of incorrect assumption about using DIC for creating or configuring of entities. It is very specific case, actual for making of reusable extensions only. But it is a bad practice by default.

Right now we can leave `new $className` code and advice the author of #13779 to override `instance` and `instantiate` methods to `Yii::createObject` call if needle.

klimov-paul:
I prefer to leave it as it is. There is no sugnificant performance issue of having DI involved here, which is the main reason it is not used all over ActiveRecord, while leave the flexibility.
Using DI over ActiveRecord is an insane idea to me too, but many people wish to do so in pursuit of Ruby-style coding.
Right now this is matter is up to developer choice. Also while creating ActiveRecord with custom constructor you may benefit of DI usage here providing default constructor arguments if needed.

framework/base/StaticInstanceTrait.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    private static $_instances = [];
+
+
+    /**
+     * Returns static class instance, which can be used to obtain meta information.
+     * @param bool $refresh whether to re-create static instance even, if it is already cached.
+     * @return static class instance.
+     */
+    public static function instance($refresh = false)
```
Ragazzo: I know almost nobody will review it, but really it is a singleton guys in 2017. I mean ok, but just rename it to actually match some naming convention? or rename interface? `singletoneInstance` or whatever. As for me this interface just mixing two things - some simple identity map (from how it is used in AR) and bad naming convention.

samdark: It is not exactly a singleton. Name matches what's there well: trait enables getting a static instance but, unlike singleton, doesn't prevent using constructor.

Map is needed because else a single instance will be shared by classes using this trait no matter what the class is.

bscheshirwork: I think we need a separate chapter in a docs 
witch details about this new functional:
- difference between StaticInstance and Singelton pattern
- Usage case (with examples i.e.)
- Best practics

samdark: Yes, could be useful. If you have ideas on where to mention all that, feel free to create a pull request. Perfect language isn't necessary, we'll edit it together.

devantoine:
How do you handle versions?
This PR clearly breaks BC but still it'll be included in 2.0.13.

Are you aware that when you break a BC you're suppose to at least increase the second version number, the first at best but not the third one?

You're not supposed to break BC in minor versions.

This PR will break humhub again: https://github.com/humhub/humhub/blob/master/protected/humhub/modules/user/models/GroupPermission.php#L26

bscheshirwork:
So... BC broken after insert new feature? Nice joke. 
I can see same name in children. It's pain.

cebe:
> This PR clearly breaks BC but still it'll be included in 2.0.13.

BC break is not obvious. Please open a new issue describing the problem, we'll check how to solve it then.

devantoine:
@bscheshirwork The author does say it breaks BC.

devantoine:
@cebe Just look at the link I posted. You could have classes having an `instance()` method and extending a class which implements `ActiveRecordInterface`  like the one I posted.

cebe:
@devantoine I see the problem, but a comment under a closed PR will be forgotten, if you open an issue, I can add it as a blocker to 2.0.13 milestone.

Issue #14467, opened 2017-07-17
yii\filters\ContentNegotiator violates HTTP 1.1 spec


PowerGamer1:
Upon a failed negotiation `ContentNegotiator` generates `415 Unsupported Media Type` response. According to HTTP 1.1 standard the `406 Not Acceptable` response should be generated instead:

https://tools.ietf.org/html/rfc7231#section-5.3.2
https://tools.ietf.org/html/rfc7231#section-5.3.5
https://tools.ietf.org/html/rfc7231#section-6.5.6
https://tools.ietf.org/html/rfc7231#section-6.5.13

Also, the description for `406 Not Acceptable` HTTP status code generated by `ContentNegotiator` is incorrect:

`None of your requested content types is supported.`

should be 

`None of your requested media types is supported.`

instead (there is no such thing as `content type`, there is only a `media type` and a `Content-Type` header that specifies the media type of response).

Alex-Code:
Should the message be `None of your requested media types are supported.` for correct grammar too?

PowerGamer1:
> Should the message be  None of your requested media types are supported.  for correct grammar too?

I think the "is" here corresponds to a singular "None" rather than plural "media types" so it should be correct.

dynasource:
duplicates #https://github.com/yiisoft/yii2/issues/12539

PowerGamer1:
> duplicates ##12539

My issue is indeed a duplicate of https://github.com/yiisoft/yii2/issues/12539 (sorry about that, haven't searched) but my issue have a PR attached and also explains why this must NOT be treated as BC breaking (see PR) and as such implemented in 2.0 (rather than 2.1) as soon as possible. So please make sure you don't close this one and keep the other one open with a 2.1 milestone attached.

klimov-paul:
Resolved by commit 2e01076abc2c5d4db9edbc2043e8e9442bdfe294

PR #14468, opened 2017-07-17
ContentNegotiator generates 406 HTTP status code

PowerGamer1:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | I think currently tests for Yii2 are broken elsewhere anyway 
| Fixed issues  | https://github.com/yiisoft/yii2/issues/14467

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,6 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.13 under development
 ------------------------
+- Bug #14467: In case of a failed negotiation `yii\filters\ContentNegotiator` now generates "406 Not Acceptable" HTTP status code in compliance with HTTP 1.1 specification (PowerGamer1)
 - Bug #14449: Fix PHP 7.2 compatibility bugs and add explicit closure support in `yii\base\Application` (dynasource)
 - Bug #7890: Allow `migrate/mark` to mark history at the point of the base migration (cebe)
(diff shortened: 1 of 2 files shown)


Kolyunya:
@PowerGamer1 are you sure this patch does not break the backward compatibility? Legacy code which was catching `UnsupportedMediaTypeHttpException` will break now since the filter now throws `NotAcceptableHttpException`.

PowerGamer1:
> are you sure this patch does not break the backward compatibility? Legacy code which was catching  UnsupportedMediaTypeHttpException  will break now since the filter now throws  NotAcceptableHttpException .

To catch exception thrown by `ContentNegotiator` you will have to override a method like `Application::run()` or `Application::handleRequest()`. I highly doubt any application at all both override these methods AND catch `UnsupportedMediaTypeHttpException`.

Regardless, even IF such an app exists it MUST be fixed. The "415 Unsupported Media Type" code has completely different meaning in HTTP so you might as well start sending random codes in response to HTTP requests.

sergeymakinen:
Technically it breaks BC. But I think a BC impact is low while misusing HTTP codes is a pretty bad thing.

PowerGamer1:
I ma not sure what and how you'd want to unit test in this case and even if any tests are needed at all. Regardless, I don't do unit tests, take it from here please.

sergeymakinen:
I could take care of this as soon as I finish with a migration builder PR.

klimov-paul:
Resolved by commit 2e01076abc2c5d4db9edbc2043e8e9442bdfe294

Issue #13853, opened 2017-03-24
Option to hide individual database operations in migration


francislavoie:
I have some large migrations which have many individual `insert` statements to load the database with a good initial state. These SQL statements contain some large base64 encoded data blobs. When I run `yii migrate --interactive=0`, every `insert` is printed to `stdout` in full.

I would like an option to make all the individual SQL operations (those written in white between 
each migration file) to be hidden, so I'm only left with messages about each migration file being run. Could this be added to the framework?

https://github.com/yiisoft/yii2/issues/13040 this is a similar issue, but I more specifically care about only reducing the amount of output rather than getting rid of all of it (I could do `>/dev/null` for that if I wanted to).

Thanks

cebe:
If you do not want your inserts to be printed you can do so by calling 

```php
$this->db->createCommand()->insert($table, $columns)->execute();
```

instead of

```php
$this->insert($table, $columns);
```

rob006:
See https://github.com/yiisoft/yii2/pull/13787

francislavoie:
@cebe that's not at all a solution. That would require me to change every migration to use that kind of execution scheme. What I want is to conditionally show or hide those logs, with some commandline option. For example, Laravel only shows the name of the migration file being run, not every operation being done. An option to get that kind of output would be preferable.

cebe:
> That would require me to change every migration to use that kind of execution scheme.

as far as I understood the issue you were talking about one specific migration that inserts large blobs.

cebe:
Could be implemented by changing `echo` in `yii\db\Migration` to a `output()` method which checks a property `$quiet` that would not print the information when set to true.
This could be conrtolled by a flag in MigrateController.

samdark:
There's `yii\db\Migration::$maxSqlOutputLength` which could possibly solve the issue...

cebe:
its only applied to raw SQL currently. This issue is about removing all output that is generated by Migration.

francislavoie:
@cebe sorry, I wrote the issue a few months ago. I could've been clearer. Poor wording by me.

One another idea is instead of an option, you could quiet the output of debug is not on, leaving SQL logging only when debug is on. I think that would technically be better. (IIRC there is a debug mode for CLI, right? I'm not at my PC right now, can't check). Maybe that's too much of a breaking change for everyone else. If we a implement a `$quiet` property, that is configurable via the application config for the migrations, right? I'll have to dig around in the source later probably. I think I can probably make a PR.

[...]

Issue #12862, opened 2016-10-27
Separate Migrate controller functionallity into separate component


Deele:
This is not the first time I encounter the need to use migration functionality without access of console and to do that programmatically, so I decided to do something about it. I don't like idea of bridges that allow use of console commands from web application interface, so this should be handled in more effective way.

To do that, migration functionality should be separated and migrate controller converted in such way that it is mere user of mentioned migration component, to avoid duplicating code and contribute in code reuse.

I have already started to work on that, I have created `Migrator` component that is fully compatible with requirements of migrate controller using yii component event system. Only issue with such move is that this change will hinder any class that is extending `BaseMigrateController` or `MigrateController` methods, except actions itself.

I think all the functionality that is currently accessible only via console interface, should be converted in similar way, so that Yii framework could be used as a base for complex applications that handle their own package management or custom installers and would like to give extended control of application via web interface, without _hacking_ the framework.

See my progress so far in [migrator](https://github.com/Deele/yii2/tree/migrator) branch.

I would like to see your comments on my use of events (and other parts of code) and idea (separation of console controller functionality into components) overall. I'm trying to make as little changes as possible, to avoid breaking anything and creating new issues. The end result should make migrate controller work the same as it is already working, with the difference of its code sitting inside component.

So far, I have received positive feedback from @cebe , so I will continue my work on it until migrate controller has regained all of its functions via Migrate component.

samdark:
Make a pull request. It will be much easier to review.

Deele:
@samdark I am guessing that this component needs proper testing and better documentation in comments before I make pull request. I just [finished most of implementation](https://github.com/yiisoft/yii2/compare/master...Deele:migrator) that was in my head and will try to do simple testing as soon as possible. What about unit tests (are they mandatory?) and what are recommendations regarding file headers (copyright and stuff) and code style (is anybody checking it)? This is my first contribution that introduces new files into repository, that is why I have such questions.

cebe:
> needs proper testing and better documentation in comments before I make pull request

A pull request does not need to be perfect when submitted, we will comment on the current state and it will evolve until it is good to be merged.

> What about unit tests (are they mandatory?)

Not mandatory, but good to have. There are already tests for [migrate controller](https://github.com/yiisoft/yii2/blob/master/tests/framework/console/controllers/MigrateControllerTestTrait.php).

> what are recommendations regarding file headers (copyright and stuff) and code style (is anybody checking it)?

[...]

PR #14474, opened 2017-07-17
Add compact option to migrations

francislavoie:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13853

Also refactors the `echo`s in `db/Migration.php` into helper functions.


framework/db/Migration.php
```diff
@@ -552,9 +538,36 @@ public function dropCommentFromColumn($table, $column)
...
+    /**
+     * Prepares for a command to be executed, and outputs to the console
+     * 
+     * @param string $description the description for the command, to be output to the console.
+     * @return float the time before the command is executed, for the time elapsed to be calculated.
+     * @since 2.0.13
+     */
+    protected function beginCommand($description)
+    {
+        if($this->quiet === false) {
```
cebe: better `!$this->quiet` as it could be set to 0 instead of false via command line.

framework/db/Migration.php
```diff
@@ -552,9 +538,36 @@ public function dropCommentFromColumn($table, $column)
...
+
+    /**
+     * Finalizes after the command has been executed, and outputs to the console the time elapsed
+     * 
+     * @param float the time before the command was executed.
+     * @since 2.0.13
+     */
+    protected function endCommand($time)
+    {
+        if($this->quiet === false) {
```
cebe: better `!$this->quiet` as it could be set to 0 instead of false via command line.

cebe (changes requested):
Minor thing about checking `$quiet`, looks great otherwise :)

cebe (approved):
great, thanks!

rob006:
This `quiet` switch still will result quite a lot of output from migrations...

francislavoie:
The result will be 3 lines per migration file. My goal is 1 line per file, which I proposed in the linked issue but cebe suggested I omit that change for now. I can make that change if there's some agreement :)

rob006:
> The result will be 3 lines per migration file. My goal is 1 line per file

It does not change main issue - `quiet` switch does not make command quiet. We should either suppress all output or introduce `verbose` switch with some levels.

francislavoie:
Well, technically, this does make `Migration` quiet, it's the controller that's still making all the noise :P

But yeah I get what you're saying, but I'm not too sure what benefit there is to hiding all the output. It's pretty important to see what's going on with migrations I think. There's always the ability to pipe the output to /dev/null if you really want to see nothing.

My particular use-case is that the initial migration I have in my project inserts a lot of binary data into the db, so there's just a massive stream of garbage flying past the console when I run it. Plus, there's just a lot of commands being run, so the new maxSqlOutput option isn't enough for me. My project is installed on-prem by my customers, so I just want a nicer experience for them.

Anyways, I think the mode I implemented could be quiet = 1, and 2 could be trimming the migration controller output to one line per migration, 3 could be just removing all of the per-file output... something like that.

samdark (approved):

framework/db/Migration.php
```diff
@@ -552,9 +538,36 @@ public function dropCommentFromColumn($table, $column)
...
+    }
+
+    /**
+     * Prepares for a command to be executed, and outputs to the console
+     * 
+     * @param string $description the description for the command, to be output to the console.
+     * @return float the time before the command is executed, for the time elapsed to be calculated.
+     * @since 2.0.13
+     */
+    protected function beginCommand($description)
```
klimov-paul: I would rather add a single method `output()` or `log()` instead of 2 separated.
Calculating of the execution time should remain outside it.

framework/db/Migration.php
```diff
@@ -552,9 +538,36 @@ public function dropCommentFromColumn($table, $column)
...
+    {
+        if(!$this->quiet) {
```
klimov-paul: Space is missing after `if`

framework/db/Migration.php
```diff
@@ -552,9 +538,36 @@ public function dropCommentFromColumn($table, $column)
...
+
+    /**
+     * Finalizes after the command has been executed, and outputs to the console the time elapsed
+     * 
+     * @param float the time before the command was executed.
+     * @since 2.0.13
+     */
+    protected function endCommand($time)
+    {
+        if(!$this->quiet) {
```
klimov-paul: Space is missing after `if`

klimov-paul (changes requested):

klimov-paul:
I would keep in mind issue #12862 here as well.
It is crucial to have an ability not just disable migration output, but actually redirecting it to another source.
For example storing output lines in array, so they can be displayed later at web page or something.

klimov-paul:
It would be better to have migration output actuall come through the `View` component.
It make sense to introduce base `print()` method to `yii\base\View` so any view (`yii\console\View` and `yii\web\View`) should have it.
Migration should pass its output via this method through the View related to migration runner (at the present state - `MigrateController`).
In order supress output a dummy view can be used, or there could be a View, which calls `Yii::$logeer->log()` instead of output and so on.

francislavoie:
Hmm, those suggestions sound more complicated than I have time to work on. Feel free to take over if you want to do those things.

Is there any specific reasons for not merging this now and making those changes in a new PR? They seem a bit larger scope than what I did here.

framework/db/Migration.php
cebe: Imo its a good thing to move the duplicated code for calculation to a separate method.

francislavoie: Same. Code deduplication is always good IMO. There was a significant amount of it in this class, my changes clean up the duplication quite a lot.

samdark: I think this part is fine.

klimov-paul: > Imo its a good thing to move the duplicated code for calculation to a separate method.

This is funny, you had quite an opposite opinion back there:
https://github.com/yiisoft/yii2/pull/14056#discussion_r113421830

klimov-paul: Here I am against having 2 separated methods, which practically do the same thing - string output.
What it I with to change the way string is printed? I'll have to override 2 methods instead single one.

francislavoie: It's not just for printing though. Think of it like beforeSave and afterSave events for example. They're fundamentally different events. It does allow for more flexibility to do any other kind of prep work before actually running the DB operation that whoever extends it might want to do. If we made it just one function you'd probably still need some parameter to differentiate if it's a begin or end to get the same effect which is awkward design I think. I doubt there will be many examples of people extending Migration for changing logging though, so I don't think it's too big a deal to need to extend both if you're one of the few who might want to change the way it's logged.

cebe: there is no calculation in #14056 and it would be only moving an if statement to a separate method. In that case it was about reducing overhead. In this case its a calculation, which is subject to change so the risk of duplicate code is much higher.

SilverFire (approved):

cebe:
> It is crucial to have an ability not just disable migration output, but actually redirecting it to another source.

this is out of scope in this PR, and can be added later if needed.

cebe:
> This quiet switch still will result quite a lot of output from migrations...

@rob006 @yiisoft/core-developers how about naming it `quieter` or something similar to keep the possibilty to add a more general `quiet` option to console controllers later? Migration::$quiet is okay imo, as it supresses all migration output.

rob006:
I'm for `verbose` option for command, like `--verbose=3`. Using `--verbose=0` will suppress all output of command. 

> Migration::$quiet is okay imo, as it supresses all migration output.

I agree.

rob006:
After reconsidering: `---muteSql` or `--quietSql` switch will be simpler to use and more descriptive than `verbose` option and less magical than `quieter`.

klimov-paul:
Migrate command has an implementations for NOSQL databases, so not `sql` part can be used for option name.

cebe:
brain storm: `--lessDetail` `--fewerDetails` `--compact` `--compactInfo` `--compactOutput`

samdark:
`--compact` looks like a nice option name.

francislavoie:
`compact` sounds good to me, should I make the change?

samdark:
Yes.

francislavoie:
I had `-q` for `--quiet`. Should we use `-c` for `--compact` or just not have an alias?

cebe:
> Also, anything I should do to make codeclimate not fail?

no, its okay.

cebe (approved):
I'd have kept `$quiet` on the migration and `$compact` on the controller but its fine like this too.

samdark:
Merged. Thank you!

framework/console/controllers/BaseMigrateController.php
```diff
@@ -752,7 +759,7 @@ protected function migrateDown($class)
     protected function createMigration($class)
     {
         $this->includeMigrationFile($class);
-        return new $class();
+        return new $class(['compact' => $this->compact]);
```
klimov-paul: This is unacceptable, it wil break migrations of 3rd party libraries such as `yii2-mongodb` as thier migrations have no `compact` property!

samdark: Oh... right.

samdark: Would the following be better?

```php
...
$migration = new $class();
if ($migration->canSetProperty('compact')) {
    $migration->compact = $this->compact;
}
return $migration;
```

klimov-paul: No.
Who can be sure Migration class is a descendant of `BaseObject`?
`BaseMigrationController::createMigration()` should be left as it was before. `compact` optnios should be applied only at `MigrationController::createMigration()`.

PR #14478, opened 2017-07-18
PHP 7.2: Rename `yii\base\Object` to `yii\base\BaseObject`

### LINKED ISSUE #7936: [PHP7] Consider to deprecate yii\base\Object as object might become soft-reserved

cebe:
- Replaces all occurences of `yii\base\Object` in code and docs with `yii\base\BaseObject`, thanks @rob006 for preparing in #12699
- Add `yii\base\Object` as backwards compatibility fill-in. 5e15b94
- Add detailed UPGRADE instructions: 9548a21
- Updated PHPUnit to run fine on PHP 7.2: 8ed4fac

fixes #7936


samdark (approved):

dynasource (approved):
looks good. 2 replacements around yii\base\Object::init() are still missing in vendor/yiisoft/yii2-dev/docs/guide-fr/concept-components.md

cebe:
> looks good. 2 replacements around yii\base\Object::init() are still missing in vendor/yiisoft/yii2-dev/docs/guide-fr/concept-components.md

good catch, fixed.

PR #14484, opened 2017-07-18
Add testcase demonstrating UniqueValidator bug

laszlovl:
Commit 0df8020dd0e2990a9bc39bd84056aa2c88d50df7 broke UniqueValidator for models with a "default scope".

What happens is that UniqueValidator queries the target class only `select()`'ing the primary key, while the default scoped relation's foreign key is necessary in `filterByModels`. Because of the `asArray()` in UniqueValidator, the missing column isn't handled gracefully by ActiveRecord but triggers a key error.

There are a few possible solutions, not sure which one is preferred.

1. Remove `asArray()` in https://github.com/yiisoft/yii2-framework/blob/master/validators/UniqueValidator.php#L178
2. Remove if block in https://github.com/yiisoft/yii2-framework/blob/master/validators/UniqueValidator.php#L174 so every column is selected
3. Gracefully handle non-existing keys in `filterByModels`

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -174,6 +174,9 @@ class UniqueValidator extends Validator
                 // only select primary key to optimize query
                 $columnsCondition = array_flip($targetClass::primaryKey());
                 $query->select(array_flip($this->applyTableAlias($query, $columnsCondition)));
+                
+                // since we only select() the primary key, a potential default scope can't be loaded
+                $query->with = null;
             }
             $models = $query->limit(2)->asArray()->all();
             $n = count($models);
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -175,7 +175,7 @@ class UniqueValidator extends Validator
                 $columnsCondition = array_flip($targetClass::primaryKey());
                 $query->select(array_flip($this->applyTableAlias($query, $columnsCondition)));
                 
-                // since we only select() the primary key, a potential default scope can't be loaded
+                // any with relation can't be loaded because related fields are not selected
                 $query->with = null;
             }
             $models = $query->limit(2)->asArray()->all();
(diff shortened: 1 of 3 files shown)


cebe:
we could probably just reset `with` to be empty as there is no need to load any related data for the validation.

laszlovl:
Yeah, that makes sense. Pushed a fix :)

samdark:
Hmm... are you sure it's a good idea? Couldn't be there be a case where another table is always loaded together with primary one and there's a condition for filtering based on this joined table?

laszlovl:
Hm, doesn't look like Travis wants to re-run :disappointed:

samdark:
Travis dislikes conflicting branches. Should be fine now.

cebe:
> Hmm... are you sure it's a good idea? Couldn't be there a case where another table is always loaded together with primary one and there's a condition for filtering based on this joined table?

`with` does not do any joins, only `joinWith` does :)

samdark (approved):

tests/framework/validators/UniqueValidatorTest.php
```diff
@@ -434,4 +434,30 @@ public function testExpresionInAttributeColumnName()
...
+    public function testFindModelWith()
+    {
+        $validator = new UniqueValidator([
+            'targetAttribute' => ['status', 'profile_id']
+        ]);
+
+       
+        $model = WithCustomer::find()->one();
+
+        $validator->validateAttribute($model, 'dummy');
```
dynasource: don't we need any assertion + can it be made more descriptive for an easier understanding?

framework/validators/UniqueValidator.php
```diff
@@ -174,6 +174,9 @@ private function modelExists($targetClass, $conditions, $model)
                 // only select primary key to optimize query
                 $columnsCondition = array_flip($targetClass::primaryKey());
                 $query->select(array_flip($this->applyTableAlias($query, $columnsCondition)));
+                
+                // since we only select() the primary key, a potential default scope can't be loaded
```
dynasource: this comment is still quite magical to me. Can it be made a bit more dummy proof?

dynasource (changes requested):

tests/framework/validators/UniqueValidatorTest.php
developeruz: @dynasource please, review again

samdark:
Merged. Thank you!

PR #14492, opened 2017-07-19
Fixed error handler not escaping error info in debug mode

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | -

Current code won't actually output any error info. `print_r()` with `true` passed as a second argument doesn't print itself but returns a string that we aren't printing afterwards.

Seems it's like that since 2014: https://github.com/yiisoft/yii2/commit/7511de6cab2f57a599638063c2823cf83828f0e8

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -38,7 +38,7 @@ Yii Framework 2 Change Log
 - Bug #14165: Set `_slave` of `Connection` to `false` instead of `null` in `close` method (rossoneri)
 - Bug #14423: Fixed `ArrayHelper::merge` behavior with null values for integer-keyed elements (dmirogin)
 - Chg #7936: Deprecate `yii\base\Object` in favor of `yii\base\BaseObject` for compatibility with PHP 7.2 (rob006, cebe, klimov-paul)
-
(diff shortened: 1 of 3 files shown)


cebe (approved):

cebe:
> Seems it's like that since 2014: 7511de6

that code is fine, seems someone changed it to use php tag later and did not see the second parameter.

samdark:
Right. I've pointed out wrong commit. It's the last time it was right...

cebe:
Seems to be a bug in PHP-CS-Fixer :-/ https://github.com/yiisoft/yii2/commit/ba0ab403b52124c941dbeb46fbd9efdc12252a5d#diff-23a5ef85354dc67ddeec3bb5782a94a2L19

cc @rob006 

@samdark same issue exists in `framework/views/errorHandler/previousException.php`

cebe (changes requested):
@samdark same issue exists in `framework/views/errorHandler/previousException.php`

samdark:
Updated.

PR #14498, opened 2017-07-20
Fixes #14493: Fixed getting permissions in  by user with id equals 0

### LINKED ISSUE #14493: rbac permission error when user id is 0
create a user with id=0, then call $auth->getPermissionsByUser(0), the result will be [];
and all permission assigned to this user will not go into effect.

In yii\rbac\Dbmanger.php, line 526,
 if (empty($userId)) {  // here user.id=0 is traded as a empty. 
            return [];
  }

user with id=0 should be run as a normal user.
or we should document the problem to remind the developer of YII2 to avoid using a user with id=0;

### What do you get instead?
I set the user 0 as the admin of my system, and follows the yii2 Authorization programming Tutorial. but can not get the role admin works. So i trace the code and find this. 
Now, I directly edit my DB to avoid all the data associated with user 0. and set a new user with id>0 as the admin Role

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version   | 5.6.3
| Operating system | WIN 10

dmirogin:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14493

### DIFF
--- a/framework/rbac/DbManager.php
+++ b/framework/rbac/DbManager.php
@@ -457,7 +457,7 @@ class DbManager extends BaseManager
      */
     public function getRolesByUser($userId)
     {
-        if (!isset($userId) || $userId === '') {
+        if ($this->isEmptyUserId($userId)) {
             return [];
         }
 
@@ -524,7 +524,7 @@ class DbManager extends BaseManager
      */
     public function getPermissionsByUser($userId)
     {
-        if (!isset($userId) || $userId === '') {
+        if ($this->isEmptyUserId($userId)) {
             return [];
         }
 
@@ -670,7 +670,7 @@ class DbManager extends BaseManager
      */
     public function getAssignment($roleName, $userId)
     {
-        if (empty($userId)) {
+        if ($this->isEmptyUserId($userId)) {
             return null;
         }
 
@@ -694,7 +694,7 @@ class DbManager extends BaseManager
      */
     public function getAssignments($userId)
     {
-        if (empty($userId)) {
+        if ($this->isEmptyUserId($userId)) {
             return [];
         }
 
@@ -851,7 +851,7 @@ class DbManager extends BaseManager
      */
     public function revoke($role, $userId)
     {
-        if (empty($userId)) {
+        if ($this->isEmptyUserId($userId)) {
             return false;
         }
 
@@ -865,7 +865,7 @@ class DbManager extends BaseManager
      */
     public function revokeAll($userId)
     {
-        if (empty($userId)) {
+        if ($this->isEmptyUserId($userId)) {
             return false;
         }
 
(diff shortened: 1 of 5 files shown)


framework/rbac/DbManager.php
```diff
@@ -524,7 +524,7 @@ public function getPermissionsByRole($roleName)
      */
     public function getPermissionsByUser($userId)
     {
-        if (empty($userId)) {
+        if (!is_numeric($userId) && empty($userId)) {
```
jblac: Pretty sure this would conflict with mongodb, since their id's are not numeric.   This PR negates the ability to utilize mongo with yii2 and utilize the O.O.B RBAC system.   I may be completely wrong in that ( it's happened before ), but pretty sure i'm not.

dmirogin: @jblac Before this PR any non-empty (as numeric and the other) data could be passed to function. Now i added possibility to pass any  numeric data (including empty values that is equal to zero).

SilverFire: The condition is:

> When userId is NOT numeric and is empty - return empty array.

I don't think this change will cause problems with non-integer IDs, but @dmirogin, just to make sure that the change is OK, could add a test against strings and Value Objects like this one:

```
class UserID
{
    private $id;

    public function __construct($id)
    {
        $this->id = $id;
    }

    public function __toString()
    {
        return $this->id;
    }
}
````

**Review by SilverFire — CHANGES_REQUESTED** [review]
see https://github.com/yiisoft/yii2/pull/14498/files#r128893051

File: framework/rbac/DbManager.php (hunk 1 above)
Comment (dmirogin) [inline framework/rbac/DbManager.php]: @SilverFire,  did i understand right ?
Just add tests like 
```
$this->auth->assign($createPost, '');
$this->assertEmpty($this->auth->getPermissionsByUser(''));
```

Comment (SilverFire) [inline framework/rbac/DbManager.php]: And test when object is passed to `getPermissionsByUser()` method

**Comment by jblac** [comment]
So if non-integer Id return empty array - doesn't break the function but with mongo it would always return an empty array meaning it is rendered useless for mongo users

**Comment by jblac** [comment]
I missed the most important part- the && -  for some reason I guess I read it as or, or I'm just a moron.   This shouldn't break for mongo since it's not numeric but isn't empty...

**Comment by dmirogin** [comment]
Found same problem for other methods in yii\rbac\DbManager (revokeAll, revoke, getAssignments, getAssignment, getRolesByUser). Complete later.

**Comment by cebe** [comment]
`getRolesByUser()` has a proper check for this: https://github.com/yiisoft/yii2/blob/a55de109aaf16729d19cccba4939ba6a514b02a7/framework/rbac/DbManager.php#L460

```php
        if (!isset($userId) || $userId === '') {
            return [];
        }
```

should be applied to all methods that receive a userId and check for empty. probably a private method should be created to have this logic in one place.

SilverFire (approved):

cebe (changes requested):
Same fix should be applied to:

- DbManager::getAssignment()
- DbManager::getAssignments()
- DbManager::revoke()
- DbManager::revokeAll()

ideally wrap the check in a method `isEmptyUserId()` or similar, to ensure if changes are made, they apply to all methods at once.

cebe (approved):

samdark (approved):

samdark:
Merged. Thank you!

Issue #14370, opened 2017-06-30
Method name same as Validator's


svit-ws:
### What steps will reproduce the problem?
In my search model I have rule with core validator as Closure:
```php
    public function rules()
    {
        return [
            ...
            [['phone'], 'filter', 'skipOnEmpty' => true, 'filter' => function ($value) {
                return preg_replace('/[^0-9]+/', '', $value);
            }],
        ];
    }
```
and method with the same name:
```php
    private function filter()
    {
...
    }
```

### What do you get instead?
Setting unknown property: yii\validators\InlineValidator::filter

### Additional info
If I remove rule or method - all is ok

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.0.18
| Operating system | Ubuntu 16.04

dynasource:
This is caused by the `hasMethod` logic https://github.com/yiisoft/yii2/blob/master/framework/validators/Validator.php#L211

Because this is a default validator, it's unexpected to interfere with inlineValidation, so I would expect another check on `static::$builtInValidators[$type]`

PR #14499, opened 2017-07-20
Fixed creating built-in validator in model with same function name

dmirogin:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14370

### DIFF
--- a/tests/data/validators/models/FakedValidationModel.php
+++ b/tests/data/validators/models/FakedValidationModel.php
@@ -88,12 +88,4 @@ class FakedValidationModel extends Model
     {
         return $this->inlineValArgs;
     }
-
-    /**
-     * @see https://github.com/yiisoft/yii2/issues/14370
-     */
-    private function number()
-    {
-        return 42;
-    }
 }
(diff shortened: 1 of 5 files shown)


tests/data/validators/models/FakedValidationModel.php
```diff
@@ -88,4 +88,12 @@ public function getInlineValArgs()
     {
         return $this->inlineValArgs;
     }
+
+    /**
+     * @see https://github.com/yiisoft/yii2/issues/14370
+     */
+    private function number()
```
dynasource: shouldnt this be mapped to a configured rule `Rules()`?

dynasource (changes requested):

tests/data/validators/models/FakedValidationModel.php
dmirogin: This bug happen when you have function in model class with name that same to one of built-in validators. And Validator::createValidator create InlineValidator instead NumberValidator in this case. I can make separate model class and test for getting this more clear.

dynasource: true. It's a bit magical in the current form

dynasource (approved):
looks more verbose now

cebe (approved):

samdark (approved):

samdark:
Merged. Thanks!

PR #14504, opened 2017-07-21
Change pjax binding on links.

kamynina:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

Very useful fix that bind pjax handler on all links, including links come via pjax.

### DIFF
--- a/framework/widgets/Pjax.php
+++ b/framework/widgets/Pjax.php
@@ -204,7 +204,7 @@ class Pjax extends Widget
         $js = '';
         if ($this->linkSelector !== false) {
             $linkSelector = Json::htmlEncode($this->linkSelector !== null ? $this->linkSelector : '#' . $id . ' a');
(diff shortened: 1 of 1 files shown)


dynasource:
I'm sorry to inform you that the core team has decided to stop with the support of PJAX. This implies that we are not accepting PR's for updating PJAX in the 2.0.x branch anymore. In the 2.1 branch it will be removed from the framework. 

A lot has been said on this subject. Our conclusion is that the current state of PJAX does not meet our standards to properly support in the Yii framework. For this reason, we are also not moving it into a Yii2 extension. Of course the Community is free to fill this gap, but with a warning. It is very time consuming to support and in our humble opinion, it should be completely rewritten to properly meet the many usecases developers have with it.

PR #14516, opened 2017-07-22
Fixes #7823: Added possibility to limit request by ajax in yii\filters\AccessRule

### LINKED ISSUE #7823: behavior access only ajax
hi, i have a question

how can block in the behavior for give access to the action ONLY by ajax ?

any idea or i need create a custom filter for this case

for example
public function behaviors()
    {
        return [  
            'ajax' => [
                'class' => AjaxFilter::className(),
                'only' => ['action1','action2'],  
            ]
        ];
    }
 and if you try enter with URL like (http://localhost/controller1/action1 you get 403

dmirogin:
Added possibility to limit request by ajax in yii\filters\AccessRule

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7823

### DIFF
--- a/framework/filters/AjaxFilter.php
+++ b/framework/filters/AjaxFilter.php
@@ -35,7 +35,7 @@ class AjaxFilter extends ActionFilter
     /**
      * @var string the message to be displayed when request isn't ajax
      */
-    public $errorMessage = 'Request must be XMLHttpRequest only.';
+    public $errorMessage = 'Request must be XMLHttpRequest.';
     /**
      * @var Request the current request. If not set, the `request` application component will be used.
      */
--- a/framework/filters/AjaxFilter.php
+++ b/framework/filters/AjaxFilter.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\filters;
+
+use Yii;
+use yii\base\ActionFilter;
+use yii\web\BadRequestHttpException;
+use yii\web\Request;
+
+/**
+ * AjaxFilter allow to limit access only for ajax requests.
+ *
+ * ```php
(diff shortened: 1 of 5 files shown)


cebe:
this is not really the job of AccessFilter but something like VerbFilter. Its not "Forbidden" but "Invalid Request".

dmirogin:
Yes, really better make separate filter class.
@cebe Do you have other sight about utility this proposal ?

cebe:
> @cebe Do you have other sight about utility this proposal ?

sorry, I do not understand that question, please rephrase :)

dmirogin:
@cebe Guess i didn't understand you correctly from this sentence
> Except from that your code looks great to me, just not whats needed imo

I thought you want other changes except making separate filter class. Never mind so =).

cebe:
Nope. just wanted to point out that I like your style of contribution :)

SilverFire:
I agree with @cebe that it's not a responsibility of AccessRule class but I don't think we should add this feature to VerbFilter because it's not about HTTP request verb.

How about a separate AjaxFilter?

```php
[

**Comment by cebe** [comment]
yep, additional filter was also discussed in the Issue and is the best option imo.

File: framework/filters/AjaxFilter.php [hunk 1]
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @author Dmitry Dorogin <dmirogin@ya.ru>
+ * @since 2.0.13
+ */
+class AjaxFilter extends ActionFilter
+{
+    /**
+     * @var string the message to be displayed when request isn't ajax
+     */
+    public $errorMessage = 'Request must be XMLHttpRequest only.';
```
samdark: Request must be XMLHttpRequest.

samdark (approved):

dynasource (approved):

cebe (approved):

samdark:
Merged. Thank you!

PR #14518, opened 2017-07-23
Fix #14322: Check the result of pdo-commit() in Transaction.Commit

### LINKED ISSUE #14322: Should Transaction.Commit check the return value of pdo->commit()?
I found that in Tranaction.php, the return value of commit is ignored. But according to the document of pdo, commit() will return true or false. So is it right to ignore the return value?

In http://php.net/manual/en/pdo.commit.php ：
> Returns TRUE on success or FALSE on failure.

In https://github.com/yiisoft/yii2/blob/master/framework/db/Transaction.php#L157 the return value of commit() is ignored.
>  $this->db->pdo->commit();

Please let me know if I do not make it clear. Thanks!

shenli:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14322

Please correct me if I am not using the right way to fix it.
I am trying to figure out where and how to add an unit test.

### DIFF
--- a/framework/db/Transaction.php
+++ b/framework/db/Transaction.php
@@ -154,7 +154,9 @@ class Transaction extends \yii\base\BaseObject
         $this->_level--;
         if ($this->_level === 0) {
             Yii::trace('Commit transaction', __METHOD__);
-            $this->db->pdo->commit();
+	    if (!$this->db->pdo->commit()) {
+		throw new Exception('Failed to commit transaction: commit return false.');
+	    }
             $this->db->trigger(Connection::EVENT_COMMIT_TRANSACTION);
             return;
         }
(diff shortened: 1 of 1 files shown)


cebe:
The change looks okay to me, but I'd like to dig further into PDO to figure out in which cases this could actually happen. See my comments in the issue for more detail: https://github.com/yiisoft/yii2/issues/14322#issuecomment-317240802

dynasource (changes requested):
shouldn't we introduce \yii\base\PDOException or use \yii\base\Exception.

PHP advises not to use \PDOException http://php.net/manual/en/class.pdoexception.php, although I see Doctrine extending from it directly

samdark:
It's not advising not to use it. It advices not to throw it yourself.

samdark:
Seems it's not a good idea to check it.

PR #14520, opened 2017-07-23
Fix for invalid example in Request.php

PowerGamer1:
"application/json" media type does not have any parameters (including "charset"):
https://tools.ietf.org/html/rfc7159#section-11
https://www.iana.org/assignments/media-types/application/json

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -


samdark:
Thanks.

Issue #13103, opened 2016-12-01
Improve logging in Yii 2.1


nkovacs:
The problems with the current `\yii\log\Target` and co.:

* `logVars` is not flexible enough. You can't access `Yii::$app->user->id`, for example.
* `getContextMessage` only supports returning a single string. This is inflexible for something like sentry, which allows you to pass structured data.
* The context message is logged as a separate message. This is not very useful for anything more complex than `FileTarget`. It's completely useless for [yii2-sentry](https://github.com/hellowearemito/yii2-sentry), so we had to disable `getContextMessage`.
* The first parameter to `Logger::log` can be an array, but that isn't really documented. You can't easily log extra data, you have to format it all into a string.

My proposal:

* `\yii\log\Target` allows you to attach extra data (an associative array) to each message. This can be configured as a callback or a method you can extend.
* `FileTarget` and `SyslogTarget` can use `VarDumper::dumpAsString` and append it to the message, so you get similar output as in Yii 2.0 (except it can be structured better, e.g. you could indent all the extra data so it's easier to read the log file).
* `DbTarget` can json encode it into a new field (or you can extend DbTarget and add new fields, allowing you to use db filtering, for example).
* `EmailTarget` can also use `VarDumper::dumpAsString`, but format it in a prettier way. Or you can extend it and do something fancier with the data.
* support for `logVars` and `getContextMessage`  can be kept for partial backwards compatibility, if `logVars` is configured, or `getContextMessage` is overridden in a child class, add their output to the extra data (I'm not sure if it's possible to be completely backwards compatible with Yii 2.0, probably not)
* `Logger::log` and `Yii::info` etc. allow you to pass extra data, which will be merged with the global extra data.

samdark:
Yes, there are limitations which could be addressed in 2.1 branch. Do you want to work on it?

nkovacs:
Yeah, I'll try to find some time to implement this.

dynasource:
this issue should be taken into account https://github.com/yiisoft/yii2/issues/6472#issuecomment-288159529

samdark:
#13702

PR #14524, opened 2017-07-23
Implementing Migration::change() [WIP]

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7928, #593

This is the first scratch implementation of Migration::change().
The PR contains changes in Connection logging in order to reduce direct dependencies on `Yii::info()`, `Yii::error()` etc in support of #13103

TODO:
 - [x] Basic implementation for `createTable`
 - [ ] Review and adjustments
 - [ ] Tests
 - [ ] Docs

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -12,6 +12,10 @@ use yii\base\InvalidConfigException;
 use yii\console\Controller;
 use yii\console\Exception;
 use yii\console\ExitCode;
+use yii\db\fake\FakeConnection;
+use yii\db\Migration;
+use yii\db\MigrationReverser;
+use yii\db\ReversibleMigrationInterface;
 use yii\helpers\Console;
 use yii\helpers\FileHelper;
 
@@ -672,7 +676,12 @@ abstract class BaseMigrateController extends Controller
         $this->stdout("*** applying $class\n", Console::FG_YELLOW);
         $start = microtime(true);
         $migration = $this->createMigration($class);
-        if ($migration->up() !== false) {
+
+        $success = $migration instanceof ReversibleMigrationInterface
+            ? $this->applyChangesFrom($migration)
+            : $migration->up();
+
+        if ($success !== false) {
             $this->addMigrationHistory($class);
             $time = microtime(true) - $start;
             $this->stdout("*** applied $class (time: " . sprintf('%.3f', $time) . "s)\n\n", Console::FG_GREEN);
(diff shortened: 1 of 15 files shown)


framework/db/Connection.php
```diff
@@ -1098,4 +1101,32 @@ public function __clone()
             $this->pdo = null;
         }
     }
+
+    /**
+     * @var ConnectionLoggerInterface
+     * @since 2.0.13
+     */
+    protected $_logger;
```
dmirogin: I think you meant private instead protected. And this should be before declaring any functions in this very class.

framework/console/controllers/BaseMigrateController.php
```diff
@@ -895,4 +909,22 @@ protected function generateMigrationSourceCode($params)
      * @param string $version migration version name.
      */
     abstract protected function removeMigrationHistory($version);
+
+    /**
+     * @param ReversibleMigrationInterface|Migration $migration
+     */
+    protected function applyChangesFrom(ReversibleMigrationInterface $migration)
```
dynasource: the method naming could be more verbose. It would be more clear if the naming is about 'reversibility'. 
Perhabs something like:
*  `applyByReversibleMigration()` and the other one `revertByReversibleMigration`
* or `upByReversibleMigration()` and `downBy....`, which is even more consistent

dynasource (changes requested):

framework/console/controllers/BaseMigrateController.php
rob006: @dynasource This will only duplicate info from argument type hint.

cebe:
> The PR contains changes in Connection logging in order to reduce direct dependencies on Yii::info(), Yii::error() etc in support of #13103

is this necessary to implement change migration? If not there should be a separate PR, also because change migration would target 2.0.x while refactoring logger would target 2.1.

SilverFire:
> is this necessary to implement change migration?

It's the shortest path to achieve result, as for me.

framework/console/controllers/BaseMigrateController.php
dynasource: `applyChangesFrom` is too general with respect to that argument type hint. Its makes code less readable

samdark:
Overall approach looks good to me.

SilverFire:
Logging is enhanced in 2.1, so I'm moving this PR to 2.1 and will reorganize it to use enhanced logging.

PR #14526, opened 2017-07-24
Don't add Fixture to the Class if it exists. Fixes #14525

### LINKED ISSUE #14525: InitDbFixture isn't loaded anymore since 2.0.12
It seems that `InitDbFixture` isn't loaded anymore when loading fixtures trough the CLI. I was looking at #5442, but that doesn't seem to be the issue.

It seems that the function `getFixturesConfig()` in `console/controllers/FixtureController.php` always adds a `Fixture` to the namespace. Therefore, if you feed it `yii\test\InitDbFixture`, you will end up with `yii\test\InitDbFixtureFixture`. That class doesn't exists, so it won't load it.

Before 2.0.12, the `$globalFixtures` was `yii\test\InitDb`, and after the `getFixturesConfig()` function, it would be the expected `yii\test\InitDbFixture`.

The commit that broke this is 96dae1cebc1835a712be572fff47505085d035ea, but there is no reference to an issue or PR.

michaelarnauts:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no, it fixes it
| Tests pass?   | yes
| Fixed issues  | #14525

Since 96dae1cebc1835a712be572fff47505085d035ea, the `$globalFixtures` was changed to `yii\test\InitDbFixture`, causing it to become `yii\test\InitDbFixtureFixture` inside `getFixturesConfig()`.

This change checks if the original Fixture name exists before trying to add Fixture.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.13 under development
 ------------------------
 
+- Bug #14525: Fixed 2.0.12 regression of loading of global fixtures trough `yii fixture/load` (michaelarnauts)
 - Bug #14523: Added `yii\web\MultipartFormDataParser::$force` option allowing to enforce parsing even on 'POST' request (klimov-paul)
 - Bug #14533: Fixed `yii\validators\ExistValidator` and `yii\validators\UniqueValidator` throw exception in case they are set for `yii\db\ActiveRecord` with `$targetClass` pointing to NOSQL ActiveRecord (klimov-paul)
 - Bug #14449: Fix PHP 7.2 compatibility bugs and add explicit closure support in `yii\base\Application` (dynasource)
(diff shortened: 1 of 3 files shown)


samdark (commented):
Need tests so it won't break in the future and a line to add to changelog.

michaelarnauts:
@samdark I've added `\yiiunit\data\console\controllers\fixtures\GlobalFixture` to the tests so both `Global` and `GlobalFixture` are tested.

samdark (approved):

dynasource (approved):

michaelarnauts:
Actually, thinking about it. I think I will need to have two global fixtures to test to, and with -Fixture and one without. This will now return true if one of the two succeeds.

michaelarnauts:
I've added a 2nd test to make sure that a fixture can also be loaded when Fixture is appended. This wasn't properly tested in my previous commit since the test would succeed when one of the two Fixtures was loaded.

samdark:
Merged. Thank you!

PR #14531, opened 2017-07-24
Fix #14527: MySQL with ANSI_QUOTES support

### LINKED ISSUE #14527: Bug: Adding a comment on DB column fails when using MySQL with ANSI_QUOTES enabled
Use MySQL database with sql_mode set to ANSI_QUOTES

``QueryBuilder::addCommentOnColumn`` works as expected

### What do you get instead?
``QueryBuilder::addCommentOnColumn`` crashes with a MySQL syntax error, because ``QueryBuilder::getColumnDefinition`` returns ``null``, because its regular expression pattern suddenly does not work.

I am already preparing a fix (a pull request)

neumannjan:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14527

### DIFF
--- a/framework/db/mysql/QueryBuilder.php
+++ b/framework/db/mysql/QueryBuilder.php
@@ -340,7 +340,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
             $row = array_values($row);
             $sql = $row[1];
         }
-        if (preg_match_all('/^\s*`(.*?)`\s+(.*?),?$/m', $sql, $matches)) {
+        if (preg_match_all('/^\s*[`"](.*?)[`"]\s+(.*?),?$/m', $sql, $matches)) {
             foreach ($matches[1] as $i => $c) {
                 if ($c === $column) {
                     return $matches[2][$i];


kogli:
@samdark AFAIK, unit tests for addCommentOnColumn() already exist. See [here](https://github.com/yiisoft/yii2/blob/237e7eda3a1ef7bec3928f2cbc1a0cef240ed82c/tests/framework/db/QueryBuilderTest.php#L1826-L1841). The thing is that these tests would have to be run with two MySQL server variants: one default and one with ``sql_mode`` set to ``ANSI_QUOTES``.

I don't know what to do about this. All I know is that the ``ANSI_QUOTES`` mode modifies the syntax of MySQL SHOW command to use double quotes (``"``) instead of backticks (`` ` ``). That's why my modification of the regular expression is necessary - if the MySQL server that the unit tests use had ``sql_mode`` set to ``ANSI_QUOTES``, the above existing unit test would fail without my bugfix in place.

It is possible that a similar bug is present elsewhere in Yii code.

samdark:
That's why I'm asking to enable testing against MySQL in ANSI mode as well somehow.

samdark:
Doesn't make sense to fix it without ensuring it won't break in the future.

PR #14534, opened 2017-07-25
Restored part of original Yii::trace documentation

rhertogh:
Restored original warning in the Yii::trace documentation that this method will only log a message when the application is in debug mode.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -


samdark:
Merged. Thanks!

PR #14539, opened 2017-07-26
Fixes #14273: yii\log\Target::$enabled now supports callable value

### LINKED ISSUE #14273: Logging by url route
Try to create logging logic of the web-application based on url hardcoded routes or based on regex. Let's say I have `/admin` (would log everything within backend side) or `/admin/bank/*` (would log everything within `bank` controller)  route and I would like to get all `_GET`, `_POST` requests happening in it. 

I would like to be able to set up this kind of structure: 

```php 
'log' => [
    'targets' => [
        [
            'class' => 'yii\log\FileTarget',
            // 'categories' => ['backend'], 
            'urlMatch' => ['/admin/*'], // <---
            'logVars' => ['_GET', '_POST'],
            'logFile' => '@app/runtime/backend.log',
        ],
    ],
],
```
So that covers all of the requests on the `admin/*`. 

> With categories, I will need to go to each controller or model and place this category to get the loggin information, which is not very comfortable. 

If there is anything like this already, please provide a link, this will be useful. 

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11
| PHP version      | 5.5.9
| Operating system | Ubuntu 14.04
| Advanced app or basic     | Advanced

dmirogin:
Added posibility to use callable in yii\log\Target::$enabled

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14273


samdark (approved):

framework/log/Target.php
```diff
@@ -37,7 +37,7 @@
 abstract class Target extends Component
 {
     /**
-     * @var bool whether to enable this log target. Defaults to true.
+     * @var bool|callable whether to enable this log target. Defaults to true.
      */
     public $enabled = true;
```
cebe: It would be better to convert `enabled` to getters and setters, where the getter would evaluate the closure, so that you can still use `if ($target->enabled)` even if it is a closure. Otherwise such check would always return true instead of evaluating the closure.

cebe (changes requested):

cebe (approved):

cebe:
Merged, thank you!

PR #14546, opened 2017-07-27
Added `dataDirectory` property into `BaseActiveFixture`

leandrogehlen:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

With this property it's possible to define folder to data file fixtures.
This allow  to use `tests/_data` as default folder in Yii2 module in Codepection


framework/test/BaseActiveFixture.php
```diff
@@ -115,4 +109,32 @@ public function unload()
...
+     * @throws InvalidConfigException if the specified data file does not exist.
+     */
+    protected function loadDataFile($dataFile)
+    {
+        if ($dataFile === false || $dataFile === null) {
+            return [];
+        }
+
+        if ($this->dataFolder !== null) {
+            $dataFile = $this->dataFolder . '/' . basename($dataFile);
```
samdark: This way file path is ignored so it's not possible to use subdirs.

framework/CHANGELOG.md
```diff
@@ -49,6 +49,7 @@ Yii Framework 2 Change Log
 - Bug #14492: Fixed error handler not escaping error info debug mode (samdark)
 - Chg #14487: Changed i18n message error to warning (dmirogin)
 - Enh #7823: Added `yii\filters\AjaxFilter` filter (dmirogin)
+- Enh #14546: Added `dataFolder` property into `BaseActiveFixture` (leandrogehlen)
```
samdark: What's the use case?

framework/test/BaseActiveFixture.php
```diff
@@ -32,11 +32,15 @@
      */
     public $data = [];
     /**
+     * @var string the directory path or [path alias](guide:concept-aliases) that contains the fixture data
+     * @since 2.0.13
+     */
+    public $dataFolder;
```
samdark: dataDirectory?

samdark (changes requested):

framework/test/BaseActiveFixture.php
leandrogehlen: it's possible, but it's necessary to set subdir in `dataFolder`. Ex:

```php
$this->dataFolder = '@app/subdir1/subdir2
```

framework/CHANGELOG.md
leandrogehlen: Currently, the fixtures directory is `@tests/fixtures/data`. The default directory for it in codeception is `@tests/_data` with this implementation it's possitble to change Yii2 module to use `@tests/_data` as default directory

framework/test/BaseActiveFixture.php
samdark: Why not using datafile w/o `basename`?

framework/CHANGELOG.md
pana1990: @samdark maybe we could change path by default here as well : 

https://github.com/yiisoft/yii2/blob/master/framework/test/ActiveFixture.php#L99

framework/test/BaseActiveFixture.php
leandrogehlen: To avoid conflicts like:

```php
$this->dataDirectory = '@app\tests\_data';

$this->dataFile = '@app\tests\fixutres\data\cutomer.php'
```

samdark: If data file is specified with absolute path then it's meant to be loaded from where it's specified, isn't it?

leandrogehlen: Could be, but currently it's different, the `dataDirectory` is mandatory

samdark: Yes.

leandrogehlen:
@samdark, Probably it's necessary docs improve

samdark:
Merged. Thank you!

mikk150:
I think there is a regression here..
Before when you did not have file it `yii\test\ActiveFixture::getData()` returned `yii\test\ActiveFixture::$data` which was empty array, now since it always tries to load a file, it always throws `yii\base\InvalidConfigException`

samdark:
@mikk150 please create separate issue. Will be lost here.

leandrogehlen:
@mikk150, you need to set 

```php
class MyFixture extends ActiveFixtrue {
   public $dataFile = false;
}
```

mikk150:
leandrogehlen

Breaks BC? | no
-- | --

tells me that I should not change any code

mikk150:
https://github.com/yiisoft/yii2/blob/master/framework/test/BaseActiveFixture.php#L33

Issue #14471, opened 2017-07-17
yii\filters\ContentNegotiator should set response format even when content negotiation has failed


PowerGamer1:
`yii\filters\ContentNegotiator` should set response format to a one defined in `formats` even when content negotiation has failed. Otherwise, in a situation when a client sends request with `Accept: text/html` and ContentNegotiator is configured with `formats = [Response::FORMAT_JSON];` it looks like this: "hey client, you requested the response in text/html but I don't support text/html since I am configured to support only application/json for this request, so here take this error response in - surprise, surprise - text/html format!"

samdark:
Agree.

PR #14548, opened 2017-07-27
Fix for #14471

PowerGamer1:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | practically not
| Tests pass?   | yes
| Fixed issues  | #14471

### DIFF
--- a/framework/filters/ContentNegotiator.php
+++ b/framework/filters/ContentNegotiator.php
@@ -195,14 +195,15 @@ class ContentNegotiator extends ActionFilter implements BootstrapInterface
             }
         }
 
+        foreach ($this->formats as $type => $format) {
+            $response->format = $this->formats[$type];
+            $response->acceptMimeType = $type;
+            $response->acceptParams = [];
+            break;
+        }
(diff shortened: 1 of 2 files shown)


dynasource (approved):

framework/filters/ContentNegotiator.php
```diff
@@ -195,14 +195,15 @@ protected function negotiateContentType($request, $response)
             }
         }
 
+        foreach ($this->formats as $type => $format) {
+            $response->format = $this->formats[$type];
```
SilverFire: Why not to use variable `$format`?

SilverFire (changes requested):

framework/filters/ContentNegotiator.php
PowerGamer1: Good catch, fixed.

SilverFire (approved):

samdark:
Merged. Thanks!

Issue #13105, opened 2016-12-01
javascript form validation trigger the form submition


massimans:
trying to validate a form with javascript with
```
$('#order-form').data('yiiActiveForm').submitting = true;
$('#order-form').yiiActiveForm('validate');
```
when form is 
```
<?php $form = ActiveForm::begin(['id' => 'order-form', 'layout' => 'horizontal']); ?>
// somme fields
<?php ActiveForm::end() ?>
```
i expect the form to just validate the fields show error messages if any, but instead the form is submited
| Yii version      | 2.0.10
| PHP version      | 7.0.4
| Operating system | Windows

samdark:
@massimans can you try code from current master branch?

massimans:
same behavior in master for me, i will try to see whats going on in the javascript code

massimans:
apparently `yiiActiveForm.validate()` calls `updateInputs()` witch will submit the form if you you set submitting to true an no error found in the form

samdark:
What if you'll set it to false or omit it?

massimans:
both cases no validation at all, input dont turn red or green an no event triggred
checked with
```
$('#order-form').on('afterValidate', function(e) {
console.log('afterValidate')
})
$('#order-form').on('beforeValidate', function(e) {
console.log('beforeValidate')
})
$('#order-form').on('beforeValidateAttribute', function(e) {
console.log('beforeValidateAttribute')
})
```

SilverFire:
Thank you for the report, fixed

SilverFire:
The issue is reopened now because the fix contained a regerssion

DarthLegiON:
The bug appears again, I use dev-master version. In 2.0.12, 'validate' method not work instead. Anyway, correct using of this method is impossible right now, isn't it?

Shaeldon:
Mabye soon™ ?

samdark:
@Shaeldon do you know how to fix it properly?

ptolomaues:
Fixed with the crunch, since without
`$(this).data('yiiActiveForm').submitting = true;`
the validation does not work


`diff --git a/~yii.activeForm.js b/yii.activeForm.js
index ab44e0c..5bb8a11 100755
--- a/~yii.activeForm.js
+++ b/yii.activeForm.js
@@ -757,7 +757,9 @@
                 if (data.submitObject) {
                     applyButtonOptions($form, data.submitObject);
                 }
-                $form.submit();
+                if (!data.validate_only) {
[...]

PR #14554, opened 2017-07-29
Fixed #14186, fixed #14510 - forced validation regression (caused by #13105 fix)

### LINKED ISSUE #14186: Forced validation in yiiActiveForm do not trigger afterValidate event
Hi there, so for me YiiActiveForm is not performing validation on field when is set manually so i need your help to check if its my problem or Yii.
From the [Yii cookbook](https://github.com/samdark/yii2-cookbook/blob/master/book/forms-activeform-js.md#trigger-validation-for-the-whole-form) if we want to perform validation on form we just need to call 
```javascript
$('#contact-form').yiiActiveForm('validate', true);
```
we assume that it should validate the whole form and send the event afterValidate because the form was validated. But that doesn't happen and its because of this lines in [yii.activeForm.js](https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.activeForm.js#L619-L627)
that clearly say that the event afterValidate can only be launched if we are trying to submit the form which we don't because we only want the validation itself.
Currently i have a workaround by playing with the events that looks like this
```javascript
var validationHandler = function () {
     var form = $('#criar-tarefa-form');
     var data = form.data('yiiActiveForm');
     data.submitting = true;
     data.shouldSubmit = false;
     form.on('afterValidate', handler);
     form.on('beforeSubmit', function () {
               var data = $('#criar-tarefa-form').data('yiiActiveForm');
               if (!data.shouldSubmit) {
                     data.shouldSubmit = true;
                     return false;
               }
     });
     $form.yiiActiveForm('validate', false);
};
```
but its dirty and unnecessary if the event could be fired after the validation ends.
To solve that problem it could be as simple as
```javascript
 var updateInputs = function ($form, messages, submitting) {
        var data = $form.data('yiiActiveForm');

        if (data === undefined) {
            return false;
        }

        var errorAttributes = [];
            $.each(data.attributes, function () {
                if (!$(this.input).is(":disabled") && !this.cancelled && updateInput($form, this, messages)) {
                    errorAttributes.push(this);
                }
            });

            $form.trigger(events.afterValidate, [messages, errorAttributes]);

        if (submitting) {
            updateSummary($form, messages);

            if (errorAttributes.length) {
                if (data.settings.scrollToError) {
                    var top = $form.find($.map(errorAttributes, function(attribute) {
                        return attribute.input;
                    }).join(',')).first().closest(':visible').offset().top - data.settings.scrollToErrorOffset;
                    if (top < 0) {
                        top = 0;
                    } else if (top > $(document).height()) {
                        top = $(document).height();
                    }
                    var wtop = $(window).scrollTop();
                    if (top < wtop || top > wtop + $(window).height()) {
                        $(window).scrollTop(top);
                    }
                }
                data.submitting = false;
            } else {
                data.validated = true;
                if (data.submitObject) {
                    applyButtonOptions($form, data.submitObject);
                }
                $form.submit();
                if (data.submitObject) {
                    restoreButtonOptions($form);
                }
            }
        } else {
            $.each(data.attributes, function () {
                if (!this.cancelled && (this.status === 2 || this.status === 3)) {
                    updateInput($form, this, messages);
                }
            });
        }
        submitFinalize($form);
    };
```

we just pass the errorAttributes setting and the form event before checking if we are submitting. because we want validation only.´

### Yii Version 2.0.11.2

Thanks in advance

### LINKED ISSUE #14510: The state of a form is always "not validated" when using forced validation in yiiActiveForm
After `$('#form-id').yiiActiveForm('validate',true);`

`$('#form-id').data('yiiActiveForm').validated`

returns `false` always

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14186, #14510
| Related issues | #13105

There is a regression in forced validation caused by @SilverFire's commit https://github.com/yiisoft/yii2/commit/7d494c1915f4c1d4be1d2487d504b53dc2c538b6 (which is a fix for #13105). I reverted the changes, these two issues (#14186, #14510) should be fixed now. I'll reopen #13105 for better fix with tests.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -112,6 +112,8 @@ Yii Framework 2 Change Log
 - Bug #14074: Fixed default value of `yii\console\controllers\FixtureController::$globalFixtures` to contain valid class name (lynicidn)
 - Bug #14094: Fixed bug when single `yii\web\UrlManager::createUrl()` call my result multiple calls of `yii\web\UrlRule::createUrl()` for the same rule (rossoneri)
 - Bug #14133: Fixed bug when calculating timings with mixed nested profile begin and end in `yii\log\Logger::calculateTimings()` (bizley)
+- Bug #14186: Forced validation in `yiiActiveForm` do not trigger `afterValidate` event (arogachev)
+- Bug #14510: The state of a form is always "not validated" when using forced validation in `yiiActiveForm` (arogachev)
 - Enh #4793: `yii\filters\AccessControl` now can be used without `user` component (bizley)
 - Enh #4999: Added support for wildcards at `yii\filters\AccessRule::$controllers` (klimov-paul)
 - Enh #5108: `yii\validators\DateValidator` now resets `$timestampAttribute` value on empty validated attribute value (klimov-paul)
--- a/framework/assets/yii.activeForm.js
+++ b/framework/assets/yii.activeForm.js
@@ -306,9 +306,9 @@
                 needAjaxValidation = false,
                 messages = {},
                 deferreds = deferredArray(),
-                submitting = data.submitting && !forceValidate;
+                submitting = data.submitting;
 
-            if (data.submitting) {
+            if (submitting) {
                 var event = $.Event(events.beforeValidate);
                 $form.trigger(event, [messages, deferreds]);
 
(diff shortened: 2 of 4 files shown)


arogachev:
I configured GPG key, but forgot to set up signing commits using it. Sorry for that. :disappointed: Will do for future commits. :smile:

dynasource (approved):

samdark (approved):

samdark:
Merged. Thank you!

PR #14575, opened 2017-08-02
How to use service locator as shortcut

bscheshirwork:
Little example
`\Yii::$app->tz` represent a `defaultTimeZone` and can be used in `\DateTime()` constructor

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no


samdark:
Example is similar to `search`. What's special about it so it's included?

bscheshirwork:
How to create shortcut to another service. And how to use timezone from formatter

May be can mark it as `tip`

dynasource:
IMHO, an useful example of another usecase

samdark:
OK. Let it be. Merged. Thank you.

Issue #13969, opened 2017-04-12
CacheController "No cache components"


RNACode:
### What steps will reproduce the problem?
I've got the following configuration of cache driver:
```php
return [
    'id' => 'app',
    'basePath' => dirname(__DIR__),
    'bootstrap' => ['log'],
    'language' => 'en-US',
    'components' => [
        'cache' => function () {
            $cfg = ArrayHelper::getValue(App::getConfig(), 'app.cache');
            if (!is_array($cfg)) {
                throw new InvalidConfigException('Invalid cache configuration.');
            }
            $cfg['keyPrefix'] = 'it_does_not_matter';
            return Yii::createObject($cfg);
        }
       // other modules
    ]
];
```
Because definition is Closure [CacheController](https://github.com/yiisoft/yii2/blob/master/framework/console/controllers/CacheController.php) this component doesn't find a driver.
I run the following command:
`./yii cache`
### What is the expected result?
```text
The following caches were found in the system:

	* cache (yii\caching\FileCache)
```

### What do you get instead?
```text
No cache components were found in the system.
```
### Additional info
I propose change this method [findCaches](https://github.com/yiisoft/yii2/blob/master/framework/console/controllers/CacheController.php#L252) to following:
```php
/**
     * Returns array of caches in the system, keys are cache components names, values are class names.
     * @param array $cachesNames caches to be found
     * @return array
     */
    private function findCaches(array $cachesNames = [])
    {
        $caches = [];
        $components = Yii::$app->getComponents();
        $findAll = ($cachesNames === []);

        foreach ($components as $name => $component) {
            if (!$findAll && !in_array($name, $cachesNames)) {
                continue;
            }

            if ($component instanceof Cache) {
                $caches[$name] = get_class($component);
            } elseif (is_array($component) && isset($component['class']) && $this->isCacheClass($component['class'])) {
                $caches[$name] = $component['class'];
            } elseif (is_string($component) && $this->isCacheClass($component)) {
                $caches[$name] = $component;
            } elseif (is_callable($component) && $this->isCacheClass(Yii::$app->get($name, false))) {
                $caches[$name] = get_class(Yii::$app->get($name));
            }
        }

        return $caches;
    }
```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 5.4.45
| Operating system | Ubuntu

rossoneri:
[...]

PR #14586, opened 2017-08-03
Fix console cache controller

Kolyunya:
Fixed a bug in a `yii\console\controllers\CacheController` when caches
defined via a closure were not detected.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13969

### DIFF
--- a/framework/console/controllers/CacheController.php
+++ b/framework/console/controllers/CacheController.php
@@ -270,6 +270,12 @@ class CacheController extends Controller
                 $caches[$name] = $component['class'];
             } elseif (is_string($component) && $this->isCacheClass($component)) {
                 $caches[$name] = $component;
+            } elseif ($component instanceof \Closure) {
+                $cache = Yii::$app->get($name);
+                if ($this->isCacheClass($cache)) {
+                    $cacheClass = get_class($cache);
+                    $caches[$name] = $cacheClass;
+                }
             }
         }
 
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -53,6 +53,7 @@ Yii Framework 2 Change Log
 - Enh #7823: Added `yii\filters\AjaxFilter` filter (dmirogin)
 - Enh #14363: Added `yii\widgets\LinkPager::$linkContainerOptions` and possibility to override tag in `yii\widgets\LinkPager::$options` (dmirogin)
 - Bug #14202: Fixed current time in (UTC) `\Yii::$app->formatter` if time not set (bscheshirwork)
+- Bug #13969: Fixed a bug in a `yii\console\controllers\CacheController` when caches defined via a closure were not detected (Kolyunya)
 
 
 2.0.12 June 05, 2017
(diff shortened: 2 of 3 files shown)


samdark (approved):

framework/console/controllers/CacheController.php
```diff
@@ -270,6 +270,12 @@ private function findCaches(array $cachesNames = [])
                 $caches[$name] = $component['class'];
             } elseif (is_string($component) && $this->isCacheClass($component)) {
                 $caches[$name] = $component;
+            } elseif (is_callable($component)) {
```
cebe: better check for `instanceof Closure`, callable can also be true for arrays and strings.

Kolyunya: @cebe what do you think about removing the whole giant condition altogether and leaving a single call to the `Yii::$app->get($name)`. This *will* add some overhead in certain scenarios but the code will be significantly cleaner. After all this is just a maintenance console command.

Kolyunya: @cebe `BaseYii::createObject()` [accepts](https://github.com/yiisoft/yii2/blob/master/framework/BaseYii.php#L350) not only instances of `Closure` but callable strings and arrays too. Are you sure we want to add this extra restriction?

Kolyunya:
Some design decisions should be taken. See [thread](https://github.com/yiisoft/yii2/pull/14586#pullrequestreview-56053278).

samdark:
Overhead is acceptable in `CacheController` since it's unlikely to be used often enough in production to make a difference.

samdark:
But since defining a component as either string callable or array callable is quite pointless, current code is the best approach, I think.

Kolyunya:
@samdark I'm not quite sure but it seems to me that if we support callable strings and arrays in `createObject()` we should support them everywhere. I can imagine someone trying to refactor a config with inline lambdas and extract them into something like `createMyCustomComponent()`.

samdark:
Is it possible?

Kolyunya:
@samdark I used to think it was for some reason. Seems like it's not.

samdark:
Then current code is fine.

framework/console/controllers/CacheController.php
```diff
@@ -270,6 +270,12 @@ private function findCaches(array $cachesNames = [])
                 $caches[$name] = $component['class'];
             } elseif (is_string($component) && $this->isCacheClass($component)) {
                 $caches[$name] = $component;
+            } elseif ($component instanceof \Closure) {
+                $cache = Yii::$app->get($name);
```
dynasource: this is weird. $component does not seem to be used? Can this be supplemented with a `comment`?

dynasource (changes requested):

framework/console/controllers/CacheController.php
Kolyunya: Why should it be used? It's type is checked then it's being instantiated by name.

cebe (approved):

cebe:
merged, thank you!

Issue #14302, opened 2017-06-16
CSRF fails if mbstring.encoding_translation set to On


si3u:
### What steps will reproduce the problem?
I config php.ini with information bellow:
```ini
[mbstring]
mbstring.language = Japanese
mbstring.internal_encoding = UTF-8
mbstring.encoding_translation = On
```
### What is the expected result?
When I submit a form on website, I seen error

```
Error Bad Request (#400) Unable to verify your data submission
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      |  7.0.20
| Operating system | Windows

cebe:
looks like an error with CSRF instead of PHP encoding configuration. what exactly is the problem?

Ghiya:
I have the similar problem after updating to version 2.0.12.
In my case, as i find out by log traces, the issue is in CSRF cookie value validation.
Although the cookie is presented in `$_COOKIES` array as it has to be, [its value fails validation](https://github.com/yiisoft/yii2/blob/8ae207c3a186d7b4845a0e022d784ce0002dba8c/framework/web/Request.php#L1285) in `loadCookies` method of `yii\web\Request` class.
And in that case, the required cookie wont be presented in `$_cookies` array property and [new token will be generated](https://github.com/yiisoft/yii2/blob/8ae207c3a186d7b4845a0e022d784ce0002dba8c/framework/web/Request.php#L1322) in `getCsrfToken` method [according to condition](https://github.com/yiisoft/yii2/blob/8ae207c3a186d7b4845a0e022d784ce0002dba8c/framework/web/Request.php#L1325), [in its turn](https://github.com/yiisoft/yii2/blob/8ae207c3a186d7b4845a0e022d784ce0002dba8c/framework/web/Request.php#L1409) method `validateCsrfToken` will get the different `$trueToken` value and  [validation will fail](https://github.com/yiisoft/yii2/blob/8ae207c3a186d7b4845a0e022d784ce0002dba8c/framework/web/Request.php#L1414).

Ghiya:
More specificly, hashed CSRF cookie data value `$calculatedHash` [differs form its stored hash value](https://github.com/yiisoft/yii2/blob/master/framework/base/Security.php#L429) `$hash`, so validation fails in `yii\base\Security`.
As i checked, in my case it happens in updated version, in the previous version CSRF cookie data value differs from the updated one and data validation works as it should.

In log traces it looks like:

```
[...]

PR #14590, opened 2017-08-03
Fixes #14542: Ensured only ASCII characters are in CSRF cookie value

### LINKED ISSUE #14542: _csrf cookie sometimes generates invalid values which triggers ModSecurity
activate ModSecurity (always active on shared hostings providers like OVH)
include this in the controller:

public function beforeAction($action) {
        if($action->id == 'index'){
            $this->enableCsrfValidation = false;
            $cookies = Yii::$app->response->cookies;
            $cookies->remove('_csrf');
            unset($cookies['_csrf']);
        }
        return parent::beforeAction($action);
    }

and refresh the index action a few times to generate new _csrf cookies

after some refreshes (randomly) it creates a "bad cookie"

standard index page

### What do you get instead?

403 Forbidden
You don't have permission to access / on this server. 

nor any other page, ModSecurity stops it before reaching the actual index.php file

(you get the bad cookie about once every 10 refreshes, and need to delete the cookie to start it behave normally again)

example of "bad" cookie content: "964315fe9b4ea193c9706742f641945b5e64db97e407423a84ca465195f684b1a%3A2%3A%7Bi%3A0%3Bs%3A5%3A%22_csrf%22%3Bi%3A1%3Bs%3A32%3A%22%EE3%3C%0B%02%9D%5C%91%5D%E4_Z%96%F9%BB%3F%CE%00%EB%B8%A5n%D6%DE%9D%98%8F%9A%1FJ%E5%60%22%3B%7D"

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      |  7.0.17
| Operating system | Linux 4.4.67-ovh-vps-grsec-zfs-classid

samdark:
Since binary data causes issues with ModSecurity and some browsers...

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14542

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.13 under development
 ------------------------
 
+- Bug #14542: Ensured only ASCII characters are in CSRF cookie value since binary data causes issues with ModSecurity and some browsers (samdark)
 - Enh #14022: `yii\web\UrlManager::setBaseUrl()` now supports aliases (dmirogin)
 - Bug #14471: `ContentNegotiator` will always set one of the configured server response formats even if the client does not accept any of them (PowerGamer1)
 - Bug #14525: Fixed 2.0.12 regression of loading of global fixtures trough `yii fixture/load` (michaelarnauts)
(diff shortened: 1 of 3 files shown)


dynasource (approved):

rob006:
This may also fix #14302.

PR #14602, opened 2017-08-05
Added `dataExtra` to `ActiveFixture`

leandrogehlen:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

With this property it's possible to load fixture for relation `hasMany viaTable`

### DIFF
--- a/docs/guide/test-fixtures.md
+++ b/docs/guide/test-fixtures.md
@@ -104,7 +104,7 @@ In the above, we have shown how to define a fixture about a DB table. To define 
 (e.g. a fixture about certain files and directories), you may extend from the more general base class
 [[yii\test\Fixture]] and override the [[yii\test\Fixture::load()|load()]] and [[yii\test\Fixture::unload()|unload()]] methods.
 
-It is possible to load extra fixtures into any tables using the `dataExtra` property. It is expecially
+It is possible to load extra fixtures into any tables using the `dataExtra` property. It is especially
 useful if the table in question has relations. In the following example a table for many to many relation is being filled:
 
 ```php
--- a/docs/guide/test-fixtures.md
+++ b/docs/guide/test-fixtures.md
@@ -104,8 +104,8 @@ In the above, we have shown how to define a fixture about a DB table. To define 
 (e.g. a fixture about certain files and directories), you may extend from the more general base class
 [[yii\test\Fixture]] and override the [[yii\test\Fixture::load()|load()]] and [[yii\test\Fixture::unload()|unload()]] methods.
 
-In case of many to many relations it is possible to load fixtures for related models
-using the `dataExtra` property, like the following:
+It is possible to load extra fixtures into any tables using the `dataExtra` property. It is expecially
+useful if the table in question has relations. In the following example a table for many to many relation is being filled:
 
 ```php
 namespace app\tests\fixtures;
@@ -121,6 +121,7 @@ class UserFixture extends ActiveFixture
     ];
 }
 ```
+
 The data file name must be equal to the table name and should be formatted like the following:
 
 ```php
--- a/docs/guide/test-fixtures.md
+++ b/docs/guide/test-fixtures.md
@@ -104,7 +104,7 @@ In the above, we have shown how to define a fixture about a DB table. To define 
 (e.g. a fixture about certain files and directories), you may extend from the more general base class
 [[yii\test\Fixture]] and override the [[yii\test\Fixture::load()|load()]] and [[yii\test\Fixture::unload()|unload()]] methods.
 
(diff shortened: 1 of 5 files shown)


samdark (changes requested):
Overall looks good but documentation is mandatory in this case.

leandrogehlen:
I have tried to write something, but sorry my english, i hope to solve the request

docs/guide/test-fixtures.md
```diff
@@ -104,6 +104,37 @@ In the above, we have shown how to define a fixture about a DB table. To define
...
+```php
+namespace app\tests\fixtures;
+
+use yii\test\ActiveFixture;
+
+class UserFixture extends ActiveFixture
+{
+    public $modelClass = 'app\models\User';
+
+    public $dataExtra = [
```
samdark: Since we're defining fixtures for relations here, it's better to name it `$relationFixtures` instead of general `$dataExtra`. Do you agree?

samdark (changes requested):

docs/guide/test-fixtures.md
leandrogehlen: i'm not sure, because it's possible to load any data

samdark: Indeed. Not **any** data but data for any database table...

samdark (approved):

githubjeka:
Hmmm, Why you don't use depends?

samdark:
@githubjeka that's actually a good question :) @leandrogehlen what do you think?

leandrogehlen:
Because is not same thing. This classes are not mapped

githubjeka:
Controversially. 

```php
class OrderItemFixture extends ActiveFixture 
{   
    public $tableName = 'order_item';

      public $depends = [
         'app\tests\fixtures\OrderFixture'
    ];
}
```

Not problem for me. IMHO more flexible

And what about user depend by user_id ? 

```
class OrderItemFixture extends ActiveFixture 
{   
    public $tableName = 'order_item';

      public $depends = [
         'app\tests\fixtures\OrderFixture',
         'app\tests\fixtures\UserFixture',
    ];
}
```

How it should look via `dataExtra` and how order for cleanup in this case?

leandrogehlen:
@githubjeka, how can you get model fixture for `User` or `Order` in your case?

```php
class OrderItemFixture extends ActiveFixture 
{   
    public $tableName = 'order_item';

      public $depends = [
         'app\tests\fixtures\OrderFixture',
         'app\tests\fixtures\UserFixture',
    ];
}

class OrderTest extends \Codeception\Test\Unit
{

   public function _fixtures()
   {
         return [
            'items' => OrderItemFixture::className(), 
         ];
   }

   public function testSomething()
   {
       // how to get order fixture
        $order1 = $this->grabFixtures('what use here', 'order1' );
   }
}
```

Can you explain me?

githubjeka:
```php
$order1 = $this->grabRecord(Order::class, ['name'=>'order1'] );
```

or I expect that this should work if you dont like `grabRecord`
```php
class OrderItemFixture extends ActiveFixture 
{   
    public $tableName = 'order_item';

      public $depends = [
         'app\tests\fixtures\OrderFixture',
         'app\tests\fixtures\UserFixture',
    ];
}

class OrderTest extends \Codeception\Test\Unit
{

   public function _fixtures()
   {
         return [
            'order'=> OrderFixture::className(),
            'items' => OrderItemFixture::className(),            
         ];
   }

   public function testSomething()
   {
        $order1 = $this->grabFixtures('order')[0];
   }
}
```

leandrogehlen:
> or I expect that this should work if you dont like grabRecord

Thanks, for your answer. 
I think my PR it`s easier, but the decision is from yii team

docs/guide/test-fixtures.md
```diff
@@ -104,6 +104,38 @@ In the above, we have shown how to define a fixture about a DB table. To define
...
+}
+```
+
+The data file name must be equal to the table name and should be formatted like the following:
+
+```php
+<?php
+return [
+    [
+        'user_id' => 1,
```
githubjeka: It requires check off for foreign key.
If UserFixture calls through implicit dependence on another fixture with same `dateExtra`? How does it merge?

samdark:
Looks like dependencies are much cleaner and better way to achieve the goal rather than filling up tables directly. Thanks for the pull request but won't merge it.

githubjeka:
Anycase you can use it code in your projects ;)

leandrogehlen:
@samdark, no problem

PR #14605, opened 2017-08-06
fix #14604 CompareValidator compareAttribute does not work if `compar…

### LINKED ISSUE #14604: CompareAttribute does not work when you change input ID to compare against
Create simple form model with attributes;
newPassword
newPasswordRepeat

make Compare attribute on top of them, so that newPasswordRepeat checks if it is the same as newPassword

on view change newPassword ID
newPasswordRepeat should check against forms newPassword
### What do you get instead?
since it assumes that no-one changes input ID s it will never be true

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      |  5.6.30
| Operating system | Ubuntu 16.10

mikk150:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14604


framework/assets/yii.validation.js
```diff
@@ -273,7 +273,12 @@ yii.validation = (function ($) {
             if (options.compareAttribute === undefined) {
                 compareValue = options.compareValue;
             } else {
-                compareValue = $('#' + options.compareAttribute).val();
+                var attributes = $form.data('yiiActiveForm').attributes
+                for (var i = attributes.length - 1; i >= 0; i--) {
+                    if (attributes[i].id == options.compareAttribute) {
```
samdark: `===` could be used.

samdark (changes requested):
Overall makes sense.

framework/assets/yii.validation.js
mikk150: yep, that is true

framework/assets/yii.validation.js
```diff
@@ -273,7 +273,12 @@ yii.validation = (function ($) {
             if (options.compareAttribute === undefined) {
                 compareValue = options.compareValue;
             } else {
-                compareValue = $('#' + options.compareAttribute).val();
+                var attributes = $form.data('yiiActiveForm').attributes
+                for (var i = attributes.length - 1; i >= 0; i--) {
```
mikk150: 1st speed (do not need to count every time, just simple substraction)
2nd compatible with current version, which takes first element's value

samdark (approved):

samdark:
Merged. Thank you!

Issue #8527, opened 2015-05-22
Add get currency code by iso code via Yii::$app->formatter


bupy7:
Add get currency code by iso code via Yii::$app->formatter as in Symphony library 

```
echo Symfony\Component\Intl\Intl::getCurrencyBundle()->getCurrencySymbol('RUB');
```

```
.руб
```

bupy7:
For those looking for. Temp solution:

```php
<?php

namespace common\components;

/**
 * @inheritdoc
 * 
 * @author Belosludcev Vasilij http://mihaly4.ru
 */
class Formatter extends \yii\i18n\Formatter
{
    
    /**
     * Return symbol currency by iso code in $currency parameter.
     * 
     * @param string $currency the 3-letter ISO 4217 currency code indicating the currency to use.
     * @return mixed
     * @example RUR => .руб
     */
    public function getCurrencySymbol($currency = null)
    {
        $string = $this->asCurrency(0, $currency);
        $symbol = mb_ereg_replace('^[\d]*.[\d]*.', '', $string);
        return $symbol;
    }
    
}
```

cebe:
what do you need this for? you can just instantiate an IntlFormatter and get it from there...

bupy7:
@cebe I would like to it to be initially to Formatter component. Very often need get symbol of currency without formatting currency. It yes, then what you suggest - this option, but what for so? Why not from initially in component?

cebe:
> Very often need get symbol of currency

I am interested in these cases. Where/when do you need it, please name concrete examples.

bupy7:
@cebe To UI, when can enter few option price of (example) USD, UER, RUB, UAH.

samdark:
You mean these?

![Currencies](https://cloud.githubusercontent.com/assets/47294/7822403/f0704be8-03fd-11e5-93ff-af113c99a0d5.png)

bupy7:
@samdark hz =)  I not known what you do =) 

As example:

![screenshot at 27 01-26-08](https://cloud.githubusercontent.com/assets/5145037/7822520/7f8f6a32-040f-11e5-96ce-091f3505706f.png)

symbol as suffix of value to textInput with mask.

samdark:
So basically you're displaying these. OK.

nkovacs:
This is based on Symfony's solution, it's a bit more robust and returns more info:

```php
/**
 * Return currency info
 * @return array:
 *  - string, currency symbol
[...]

PR #14607, opened 2017-08-07
add getCurrencyCode method to formatter

antgubarev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #8527

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -1127,6 +1127,26 @@ class Formatter extends Component
     }
 
     /**
+     * Return a currency symbol
+     *
+     * @param null $currencyCode
+     * @return string
+     * @throws InvalidConfigException
+     */
+    public function getCurrencySymbol($currencyCode=null)
+    {
+        if ($currencyCode === null) {
+            $currencyCode = $this->currencyCode;
+        }
+        if (strlen($currencyCode) == 0) {
+            throw new InvalidConfigException("The default currency code for the formatter is not defined.");
+        }
+        $locale = $this->locale . '@currency=' . $currencyCode;
+        $formatter = new \NumberFormatter($locale, \NumberFormatter::CURRENCY);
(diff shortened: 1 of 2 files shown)


samdark (approved):

cebe:
quite sure this should not belong to formatter. Formatter is meant to be used for formatting input values and not for retrieving general locale data. This should be part of i18n component or a separate helper.

amarox:
Why? Formatter return format of currency.

cebe:
Formatter implements methods that convert some input value into a human readable format based on the locale, it is not used to provide generic information about locale. That should be a separate component. What is your use case for this method?

amarox:
For example, show label with currency in table head. 

```html
<th>Price (currency)</th>
```
and in rows

```html
<td>3 567</td>
```

amarox:
Currency locale config in Formatter but not i18n. And i mean that Formatter should return human readable currency code.

amarox:
unit tests added

samdark (approved):

rob006:
Fixes #8527

cebe:
I still think it should be part of a separate component, there are many other locale dependent information that could be retrieved, formatter is not the right place for it.

dynasource (approved):

samdark:
@cebe any suggestions on how to name the component?

bupy7:
@samdark `Sign`.

samdark:
`yii\i18n\info\Currency`?

rob006:
`yii\i18n\Locale`?

samdark:
Sounds good, @rob006. @amarox do you have some time to adjust placement of the code?

amarox:
@samdark Sure. I will fix it

samdark:
https://github.com/yiisoft/yii2/commit/63fdd26b7eb5e567e48795b91aa1af534a9c1fa2

samdark:
Thanks for starting it.

PR #14611, opened 2017-08-07
Fix #13702 PSR Logger

### LINKED ISSUE #13702: Proposal: Have you thought about using monolog as logger?

klimov-paul:
Refactoring Yii Logging to be PSR compatible.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #13702, #13706, #9458


klimov-paul:
Status: work in progress.

framework/BaseYii.php
```diff
@@ -451,7 +469,7 @@ public static function info($message, $category = 'application')
      */
     public static function beginProfile($token, $category = 'application')
```
samdark: Profiling doesn't belong to logging. Ideally it should be moved into separate component.

framework/BaseYii.php
```diff
@@ -389,7 +407,7 @@ public static function setLogger($logger)
     public static function trace($message, $category = 'application')
```
samdark: I think the method should be renamed to match `LogLevel::DEBUG` i.e. `debug()`.

samdark: Also missing log levels from `LogLevel` should be added.

framework/log/Logger.php
```diff
@@ -41,27 +46,33 @@
...
 {
+    use LoggerTrait;
+
     /**
      * Error message level. An error message is one that indicates the abnormal termination of the
      * application and may require developer's handling.
+     * @deprecated since 2.1.0 Use [[LogLevel::ERROR]] instead.
      */
-    const LEVEL_ERROR = 0x01;
+    const LEVEL_ERROR = LogLevel::ERROR;
```
samdark: It's 2.1 so I think we should drop these local constants and use `LogLevel` directly.

framework/log/PsrTarget.php
```diff
@@ -0,0 +1,77 @@
...
+        return $this->_logger;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function export()
+    {
+        foreach ($this->messages as $message) {
+            $text = $message[1];
```
samdark: Worth doing `[$level, $text, $context] = $message` (is it correct order?) and then using named variables for clarity.

framework/log/PsrTarget.php
```diff
@@ -0,0 +1,77 @@
...
+
+            if (!is_string($text)) {
```
samdark: Could be bad idea since there are log aggregation targets (Rollbar, Sentry etc.) in PSR-3 compatible loggers that should get original exception in order to work properly.

samdark (changes requested):
Overall, it's good direction.

bashkarev:
How about the `dump` method?  https://github.com/yiisoft/yii2-debug/issues/247

```php
interface ProfilerInterface
{
    /**
     * @param mixed $var variable to be dumped
     */
    public function dump($var);

}
```

framework/profile/Profiler.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    const LEVEL_PROFILE_END = 0x60;
+
+
+    /**
+     * {@inheritdoc}
+     */
+    public function begin($token, $category)
+    {
+        Yii::getLogger()->log(self::LEVEL_PROFILE_BEGIN, $token, ['category' => $category]);
```
bashkarev: `Monolog` will not work
https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php#L361

samdark: Ideally profiler should not rely on logger to write its data.

bashkarev: Events?
https://github.com/bashkarev/profiler-benchmark/blob/master/src/one/Profiler.php

samdark: Yes. Should be enough for debug panel.

klimov-paul:
Status: complete.

samdark:
Updated guide and phpdoc. @klimov-paul please check if everything is correct.

klimov-paul:
Everything looks good. Nice work.

samdark (approved):

framework/log/Target.php
```diff
@@ -252,19 +203,13 @@ public static function filterMessages($messages, $levels = 0, $categories = [],
      */
     public function formatMessage($message)
     {
-        [$text, $level, $category, $timestamp] = $message;
+        [$level, $text, $context] = $message;
```
dynasource: is it time already to add array typing to the $message parameter (since 2.1 is about PHP 7?)

tests/framework/BaseYiiTest.php
```diff
@@ -107,15 +111,64 @@ public function testSetupLogger()
...
+
+        BaseYii::setLogger(null);
+        $defaultLogger = BaseYii::getLogger();
+        BaseYii::setLogger([
+            'class' => Logger::class,
+            'flushInterval' => 987,
+        ]);
+        $logger = BaseYii::getLogger();
+        $this->assertNotSame($defaultLogger, $logger);
+        $this->assertEquals(987, $logger->flushInterval);
```
dynasource: the above tests are a bit hard to match with the test scope (`testAlias`)

dynasource (approved):
quite some work you pulled off. Nicely done.

framework/log/Target.php
samdark: Yes. Could be done.

framework/base/ErrorHandler.php
```diff
@@ -282,4 +284,19 @@ public static function convertExceptionToString($exception)
...
+     * Attempts to flush logger messages.
+     * @since 2.1
+     */
+    protected function flushLogger()
+    {
+        $logger = Yii::getLogger();
+        if ($logger instanceof \yii\log\Logger) {
+            $logger->flush(true);
+        }
+        // attempt to invoke logger destructor:
```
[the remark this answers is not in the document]
samdark: Most probably it will trigger destructor but we can't be certain.

Issue #14032, opened 2017-04-22
Yii2 problem with the authentication due to CORS 


harshitchaudhary:
After looking at various questions and issues related to CORS, I still facing the same problem I am not sure what is going wrong with the CORS and authentication. I also tried to post a question on StackOverflow which is unanswered as of now. So this is my problem

I am trying to use ajax request in reactJs to fetch a resource that require authentication from _Bearer Auth in Yii2_.

**Request Headers** 

**Accept:** */*  
**Accept-Encoding:** gzip, deflate, sdch, br  
**Accept-Language:** en-GB,en-US;q=0.8,en;q=0.6  
**Access-Control-Request-Headers:** authorization  
**Access-Control-Request-Method:** GET  
**Connection:** keep-alive  
**Host:** localhost  
**Origin:** http://localhost:8110  
**Referer:** http://localhost:8110/  
**User-Agent:** Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36  

**Response Headers**  

**Access-Control-Allow-Credentials:** true  
**Access-Control-Allow-Headers:** Authorization  
**Access-Control-Allow-Methods:** GET, POST, PUT, OPTIONS  
**Access-Control-Allow-Origin:** http://localhost:8110  
**Connection:** Keep-Alive  
**Content-Length:** 150  
**Content-Type:** application/json; charset=UTF-8  
**Date:** Fri, 21 Apr 2017 10:36:27 GMT  
**Keep-Alive:** timeout=5, max=99  
**Server:** Apache/2.4.25 (Ubuntu)  
**Www-Authenticate:** Bearer realm="api"  

This is my behavior method in yii2 framework

    public function behaviors()
      {
            $behaviors = parent::behaviors();
            unset($behaviors['authenticator']);
            return [
                    [
                        'class' => 'yii\filters\ContentNegotiator',
                        'only' => ['view', 'index', 'filters', 'slug', 'shortlist'],  // in a controller
                        // if in a module, use the following IDs for user actions
                        'formats' => [
                            'application/json' => Response::FORMAT_JSON,
                        ],
                    ],
                    'corsFilter' => [
                        'class' => \yii\filters\Cors::className(),
                        'cors' => [
                            'Origin' => ['*'],
                            'Access-Control-Request-Method' => ['GET', 'POST', 'PUT', 'OPTIONS'],
[...]

PR #14618, opened 2017-08-09
Add handleOptions to Cors filter so OPTIONS requests are handled for the preflight check.

michaelarnauts:
The handling of the CORS preflight check in yii2 is broken at the moment and workarounds are being applied all the time.

Most posts or comments handle the case of the `yii\rest\ActiveRestController` where you have to add a `'except' => ['options']` to the Authentication filter, because an OPTIONS call will be routed to the `actionOptions` function. If you don't add this, the request will succeed trough the Cors filter, and arrive at the Authentication filter, where it will be rejected since the preflight check has no Authentication.

This recommendation is potentially dangerous, since code in the `actionOptions` can be executed without authentication. Also, when you extend from a `yii\rest\Controller`, the routing isn't done to `actionOptions`, and you can't use `'except' => ['options']` to force the request to go trough. A better approach is needed. 

This MR adds a `handleOptions` to the `Cors` filter. If you add this filter in your behaviour, you obviously expect those requests to arrive at your controller so you want the preflight check to pass. This change makes sure that the preflight checks are handled correctly by accepting them with a 200 OK, but not executing the code of the actual request (since the beforeAction returns `false`).

Since CORS requests don't have authentication, you still need to have the Cors filter before the Authentication filter. By returning false in the CORS filter, the handling is stopped and the Authentication filter isn't executed, so the request isn't rejected with an 401 Unauthorised.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | yes/no
| Tests pass?   | yes/no
| Fixed issues  | #14032, but a lot more. See also https://github.com/yiisoft/yii2/issues?q=cors%20OPTIONS

### DIFF
--- a/framework/filters/Cors.php
+++ b/framework/filters/Cors.php
@@ -106,6 +106,12 @@ class Cors extends ActionFilter
         $responseCorsHeaders = $this->prepareHeaders($requestCorsHeaders);
         $this->addCorsHeaders($this->response, $responseCorsHeaders);
 
+        if ($this->request->isOptions && $this->request->headers->has('Access-Control-Request-Method')) {
+            // it is CORS preflight request, respond with 200 OK without further processing
+            $this->response->setStatusCode(200);
+            return false;
+        }
+
         return true;
     }
 
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -6,6 +6,7 @@ Yii Framework 2 Change Log
 
 - New #14151: Added `AttributesBehavior` that assigns values specified to one or multiple attributes of an AR object when certain events happen (bscheshirwork)
 - Bug #6526: Fixed `yii\db\Command::batchInsert()` casting of double values correctly independent of the locale (cebe, leammas)
+- Chg #14618: Handle OPTIONS request in `yii\filter\Cors` so the preflight check isn't passed trough Authentication filters. (michaelarnauts)
 - Bug #14542: Ensured only ASCII characters are in CSRF cookie value since binary data causes issues with ModSecurity and some browsers (samdark)
 - Enh #14022: `yii\web\UrlManager::setBaseUrl()` now supports aliases (dmirogin)
 - Bug #14471: `ContentNegotiator` will always set one of the configured server response formats even if the client does not accept any of them (PowerGamer1)


framework/filters/Cors.php
```diff
@@ -80,6 +80,10 @@ class Cors extends ActionFilter
      */
     public $actions = [];
     /**
+     * @var bool handle the preflight check and don't proceed with the execution of the request.
+     */
+    public $handleOptions = true;
```
samdark: Are there cases when it's a good idea to set it to `false`?

michaelarnauts: I guess if you have a setup where you want code to be executed on an OPTIONS request, you can do this, but you can't use authentication then. I would not use it, but that could be a use case.

samdark: OK. It makes sense. I don't like the name of the property though. "Handle" is too general. I'd name it like "skipOptions", "allowOptions" or alike. Do you have any naming ideas?

michaelarnauts: Hmm, not sure. allowOptions could indicate that OPTIONS are allowed, and possible to act on with a function. skipOptions doesn't sound right.

IMO handleOptions or processOptions covers what it does, handling the OPTIONS internally instead of passing it trough.

samdark: Well, that's not exactly handling or processing. It's not passing it further...

michaelarnauts: It also does make sure a 200 OK is sent, that's for the handling part.

What about `acceptOptions`? That would suggest that if you disable it, the function won't accept the options, what it does.

michaelarnauts:
Documentation should also be modified when this gets merged. All the docs currently indicate to use the `'except' => ['options']` configuration, what seems dangerous in my opinion. Maybe a small note in the release notes can also be useful to explain the change.

framework/filters/Cors.php
samdark: Thought about it more. We should not introduce this property. Instead, we may handle preflight request specifically:

```php
if ($this->request->isOptions && $this->request->headers->has('Access-Control-Request-Method')) {
    // it is CORS preflight request, respond with 200 OK without further processing
    $this->response->setStatusCode(200);
    return false;
}
```

michaelarnauts:
Rebased and dropped `handleOptions` variable. Checks for *Access-Control-Request-Method* header as suggested by @samdark

leandrogehlen:
@samdark, What is necessary to accept this PR? It's only the tests?

samdark:
Yes.

leandrogehlen:
Ok, I will do it.

samdark:
Closed in favor of #15420

PR #14622, opened 2017-08-09
Fix #879: Add PSR-16 simple cache

klimov-paul:
Add PSR-16 simple cache
https://github.com/php-fig/simple-cache

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #879


klimov-paul:
Complete.

However, there are some questions, which should be decided:

1) Should we keep suport for non-string keys at the `yii\caching\CacheInterface`?
They are widely used in framework code, while technically contradict PSR-16. Also note that methods like `getMultiple()` and `setMultiple()` does not support non-scalar keys even now.

2) Should be keep `add*()` methods?
These methods were added to support race-condition-avoid operations provided by some cache storages. However making solution PSR-compatible it is no longer possible to use low-level cache storage methods for that.

samdark (approved):

dynasource (approved):
It seems ok already. Quite some work you pulled off. Well done.

samdark:
I need to fix the build first :)

klimov-paul:
Build fixed

samdark:
Merged. Thank you for taking care of it.

PR #14625, opened 2017-08-10
#14582 Add Closure to min\max DateValidator

swuppio:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14582


swods:
codeclimate — 2 issues to fix, but when i open page - All good! No new issues were found.

samdark:
https://codeclimate.com/github/yiisoft/yii2/pull/14625 ?

swods:
lol, i just moved to method old code... well ok - updated, hope now it will be ok

framework/validators/DateValidator.php
```diff
@@ -265,6 +272,12 @@ public function init()
      */
     public function validateAttribute($model, $attribute)
     {
+        foreach (['min', 'max'] as $val) {
+            if ($this->$val instanceof \Closure) {
+                $this->$val = call_user_func($this->$val, $model);
+                $this->setMinMax($val);
+            }
+        }
```
rob006: This doesn't look right. Only first validation will work correctly, second will just use `min` and `max` set by previous validation.

rob006 (changes requested):

framework/validators/DateValidator.php
swods: didn't get it, example?

rob006:
```php
if (!$mode->validate()) {
    $model->minClosureDependency = 123;
    $model->validate();
}
```

`$minClosureDependency` change will be ignored by `$min` Closure in second validation, because it is no longer Closure - it was changed to fixed value in first validation. If Closure depends on `$minClosureDependency`, it will give you unpredictable results.

PR #14642, opened 2017-08-14
added possibility to minify the cached HTML response

Lugat:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | ---


cebe:
Also note that there is a Spaceless widget which has the following comment: https://github.com/yiisoft/yii2/blob/3dbc07f8dd97799be8c2188755b92b68e920d076/framework/widgets/Spaceless.php#L38-L40

samdark:
Won't include it in the framework since it doesn't make any sense considering wide availability of gzip.

Issue #14643, opened 2017-08-14
Easy change layout for error action - ErrorAction


swods:
Sometimes when we use ErrorAction, we need use different layout, but now it's little bit difficult to do it easily, so this small new feature help to solve problem

```
    public function actions()
    {
        return [
            'error' => [
                'class' => 'yii\web\ErrorAction',
                'layout' => 'error'
            ],
        ];
    }
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      |  7.1
| Operating system | Winux

klimov-paul:
Makes absolutely no sense. Changing layout inside the view file is easy:

```php
<?php
// File @app/views/site/error.php
$this->context->layout = 'overall';
```

swods:
© The removal of the tonsils through the anus - best

cebe:
@klimov-paul It can be configured in application config, leaving the original view and only changing the layout.

klimov-paul:
> It can be configured in application config, leaving the original view and only changing the layout.

What the sense of it?
How many other paramters we should add to error config to be enough?
May be we should add 'page title' for example, because setting lines line:

```
$this->title = 'Error page';
```

is too complicated for some individuals, or because it will allow to set page title ' leaving the original view' intact?

View file is able to handle all necessary conficugration relate to its rendering already, that is why there are no more options regardless to them.

swods:
why you use `$this->title = 'Error page';` as exp. it's View property, layout - Controller property

When you need to turn off `Controller::$enableCsrfValidation` for some action you also use `$this->context->enableCsrfValidation = false;`?

Your solution is very difficult for intuitive understanding. When somebody start to read the code, first attention there set the layout is in the controller NOT in where view

klimov-paul:
> Your solution is very difficult for intuitive understanding. When somebody start to read the code, first attention there set the layout is in the controller NOT in where view

View is rendered in the context of the controller. This is explainded in the guide:

https://github.com/yiisoft/yii2/blob/master/docs/guide/structure-views.md#named-views-
https://github.com/yiisoft/yii2/blob/master/docs/guide/structure-views.md#accessing-data-in-views-

It is normal to access or adjust context from the view scope - that is why `context` is exposed to it.

klimov-paul:
>why you use $this->title = 'Error page'; as exp. it's View property, layout - Controller property
[...]

PR #14644, opened 2017-08-14
#14643 Easy change layout for error action

swuppio:
Sometimes when we use ErrorAction, we need use different layout, but now it's little bit difficult to do it easily, so this small new feature help to solve problem

```
    public function actions()
    {
        return [
            'error' => [
                'class' => 'yii\web\ErrorAction',
                'layout' => 'error'
            ],
        ];
    }
```

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14643

### DIFF
--- a/framework/web/ErrorAction.php
+++ b/framework/web/ErrorAction.php
@@ -66,6 +66,13 @@ class ErrorAction extends Action
      * Defaults to "An internal server error occurred.".
      */
     public $defaultMessage;
+    /**
+     * @var string|false|null the name of the layout to be applied to this error action view.
+     * If not set, the layout configured in the controller will be used.
+     * @see \yii\base\Controller::$layout
+     * @since 2.0.15
+     */
+    public $layout;
 
     /**
      * @var \Exception the exception object, normally is filled on [[init()]] method call.
@@ -73,13 +80,6 @@ class ErrorAction extends Action
      * @since 2.0.11
      */
     protected $exception;
-
-    /**
-     * @var string|false the name of the layout to be applied to this error action view
-     * @see [[yii\base\Controller::$layout]]
-     * @since 2.0.15
-     */
-    public $layout;
 
 
     /**
@@ -105,7 +105,7 @@ class ErrorAction extends Action
      */
     public function run()
     {
-        if ($this->layout) {
+        if ($this->layout !== null) {
             $this->controller->layout = $this->layout;
         }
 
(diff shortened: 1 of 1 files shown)


framework/web/ErrorAction.php
```diff
@@ -98,6 +105,8 @@ public function init()
      */
     public function run()
     {
+        if ($this->layout) $this->controller->layout = $this->layout;
```
arogachev: Code style must be PSR-2 compatible:

- [Yii 2 Core Framework Code Style](https://github.com/yiisoft/yii2/blob/master/docs/internals/core-code-style.md)
- [PSR-2 - if, elseif, else](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#51-if-elseif-else)

framework/web/ErrorAction.php
```diff
@@ -74,6 +74,13 @@ class ErrorAction extends Action
      */
     protected $exception;
 
+    /**
+     * @var string|false the name of the layout to be applied to this error action view
+     * @see [[yii\base\Component::$layout]]
+     * @since 2.0.~
```
arogachev: It's better to specify exact version.

arogachev (changes requested):

framework/web/ErrorAction.php
swods: when i pull requisting it, i didn't know version

cebe (changes requested):
the behavior of `$layout` property of the action should be:

- if `null`, do not change the value in controller
- if anything else (including `false`) set the value on controller.

framework/web/ErrorAction.php
```diff
@@ -74,6 +74,13 @@ class ErrorAction extends Action
      */
     protected $exception;
 
+    /**
+     * @var string|false the name of the layout to be applied to this error action view
+     * @see [[yii\base\Component::$layout]]
```
cebe: should refer to controller, not component ;) also brackets are not needed.

```
@see \yii\web\Controller::$layout
```

swods: ohh miss copy

cebe:
I've applied some changes in formatting and changed the check to compare with `null` explicitly.

swods:
heh u was faster, but i wanted to do it like this - !is_null($this->layout), bad?

cebe:
` === null` matches the style we use throughout the framework.

klimov-paul:
https://github.com/yiisoft/yii2/issues/14643#issuecomment-322167870

swods:
© The removal of the tonsils through the anus - best

samdark:
https://github.com/yiisoft/yii2/commit/61b47014a2f68ccb9ed000265a8c413dc4d41af3

Merged. Thank you!

Issue #14086, opened 2017-04-29
Different length of a string in js ActiveFrom validation and php StringValidation


MatthewPattell:
### What steps will reproduce the problem?
Create active form with textarea input.
Add string validator to input with max length 20.
Enter in textarea text (19 characters) with line break. 
### What is the expected result?
Js not counting line breaks, but php takes them.
### What do you get instead?
Js validation true and php validation false. If in form exist fileinput, this behavior is bad for validator.

![screenshot_1](https://cloud.githubusercontent.com/assets/10459911/25553914/d3a854e2-2cc8-11e7-9f2f-3a6a2262792e.png)
![screenshot_2](https://cloud.githubusercontent.com/assets/10459911/25553913/d3a74804-2cc8-11e7-8b11-d02b19745631.png)
![screenshot_3](https://cloud.githubusercontent.com/assets/10459911/25553912/d3a6dd7e-2cc8-11e7-9074-c430a1dc7954.png)

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.1
| Operating system | Windows 7

TerraSkye:
I cant find the duplicate, but its been reported already, JS handles newline with 1 charachter while php takes 2

samdark:
It was for Yii 1.1, I guess...

PR #14653, opened 2017-08-16
fix mb_strlen \r\n bug

RobinKamps:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no (server side)- may affect count characters left scripts (client side)
| Tests pass?   | yes
| Fixed issues  | #14086

This PR solves the issue by counting line breaks in texareas, etc. the same way as mb_strlen in php will do - so this does not break BC server side.
However it could have some side effects to client side scripts:
Client side scripts wich count the characters left may show a different count.
(Users may also wonder if they actually count their typed in characters.)

### DIFF
--- a/framework/assets/yii.validation.js
+++ b/framework/assets/yii.validation.js
@@ -59,14 +59,17 @@ yii.validation = (function ($) {
                 return;
             }
 
-            if (options.is !== undefined && value.length != options.is) {
+            //line breaks will be counted as 2 characters (\r\n) in php/mb_strlen
+            var mb_strlen = value.replace(/\r(?!\n)|\n(?!\r)/g, "\r\n").length;
+
+            if (options.is !== undefined && mb_strlen !== options.is) {
                 pub.addMessage(messages, options.notEqual, value);
                 return;
             }
-            if (options.min !== undefined && value.length < options.min) {
+            if (options.min !== undefined && mb_strlen < options.min) {
                 pub.addMessage(messages, options.tooShort, value);
             }
-            if (options.max !== undefined && value.length > options.max) {
+            if (options.max !== undefined && mb_strlen > options.max) {
                 pub.addMessage(messages, options.tooLong, value);
             }
         },
--- a/framework/assets/yii.validation.js
+++ b/framework/assets/yii.validation.js
@@ -59,17 +59,14 @@ yii.validation = (function ($) {
                 return;
             }
 
-            //line breaks will be counted as 2 characters (\r\n) in php/mb_strlen
-            var mb_strlen = value.replace(/\r(?!\n)|\n(?!\r)/g, "\r\n").length;
-
-            if (options.is !== undefined && mb_strlen !== options.is) {
+            if (options.is !== undefined && value.length != options.is) {
                 pub.addMessage(messages, options.notEqual, value);
                 return;
             }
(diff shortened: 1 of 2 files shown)


Renkas:
Better convert it from \r\n to \n in server side and leave the old JS implementation.

There's no reason to save \r to database.

RobinKamps:
Thanks.
I guess you are right - it is much more convenient for client side scripts and it is probably better to have only one format in the DB.
Now this PR will replace Win \r\n (and Mac \r) with Linux \n if string-validator is used.
This solution works on my projects (Windows and Linux Servers).

Notes:
- I have not tested this solution with scripts where CR-LF is really neccessary (perhaps terminal/console related stuff?)
-  Really platform tailored solutions (with PHP_EOL e.g.) will be much more complex as estimated (if CR-LF or LF, etc. is really needed, it has to be custom implemented now as well anyway)
- the docs for string-validator should note this new behaviour

RobinKamps:
after writing some unit tests and doing some more research, i believe that this is a much more complex problem and there is no easy solution.
- This problem does only occur on textareas not on inputs (no line breaks possible on inputs).
- If the replacement of \r\n with just \n is placed inside the string validator the server side count by the string validator will be the same as the javascript count - this is the only good thing.
- If this replacement should be applied to the model and be saved to the DB the PR is missing something like
´´´$model->$attribute = str_replace(["\r\n", "\r"], " ", $model->$attribute);
- on windows server/apache/mysql/mariadb this replacement will be automatically converted back to \r\n. With db fields set to a length (191 for utf-8 mb... e.g.) this will cut off the last characters (quantity of \n)
- altering the input string in a validator that is not supposed to change the input value is a bad thing (adhoc validation with validateValue will not work by design, because there is no way to alter the models attribute with validateValue method - what if a string from a model is validated with validateValue and after that the model is saved to DB with save(false))

More thoughts on this issue:
- convert \r\n to \n just for the count - last characters will be cut off if field length is exceeded
- convert \r\n to blank_space and apply it to the attribute - will work in all situations where line breaks are not really needed
- convert \r\n to <br> etc. and apply it to the attribute  - last characters will be cut off if field length is exceeded
- if used with mysql text fields it is unlikely that the count error happens (text is 64 Kibibytes long)
- in case trim, purifier or other filters to delete multiple blanks etc. is called serverside before string validator the client side string validator script will show different values for length as well

Suggestion for a solution:
- Mention the probems (line breaks and trim/filters/purifier before string validator) in the docs (StringValidator and Textarea?)
- suggested working solution: place a custom rule/filter before string validator rule (e.g. for replace line breaks with blank_space):
´´´ ['string_attribute_name', 'filter', 'filter' => function ($value) {
                return str_replace(["\r\n", "\r", "\n"], " ", $value);
            }],

To solve the issue that js counts the \r\n as one char this custom filter is only neccessary for string attributes, that will be wired up to textareas.
The filter should be applied to all string attributes if no line breaks should occur in the db at all - \r\n could still be posted to input attributes via scripts, postman etc. - but that has nothing to do with the issue.
With this solution there is no "magic" attribute alteration hidden in the framework - it just takes one look to the model rules to realize that the attribute is beeing proccessed and line breaks are replaced.

samdark:
Thanks for research.

PR #14658, opened 2017-08-17
Update core-code-style.md

des1roer:
http://symfony.com/doc/current/contributing/code/standards.html

Add a blank line before return statements, unless the return is alone inside a statement-group (like an if statement);

| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

### DIFF
--- a/docs/internals/core-code-style.md
+++ b/docs/internals/core-code-style.md
@@ -347,6 +347,7 @@ $n = 100;
 $sum = array_reduce($numbers, function ($r, $x) use ($n) {
     $this->doMagic();
     $r += $x * $n;
+    
     return $r;
 });
 


samdark:
Yes. That's fine to add such rule. It is in line with what we're actually doing.

Rules in https://github.com/yiisoft/yii2/tree/master/cs should be updated.

rob006:
@samdark I already tried it, but as a hard rule it generates too many dumb empty lines (IMO).

samdark:
@rob006 was it with "unless" part?

rob006:
Yes, but still sometimes I prefer some code stick to return statement and such empty line  breaks code grouping and creates many "lonely lines". 

Anyway, I could open PR with such rule enable and you can decide about it. In this week I want to move this CS stuff forward anyway.

samdark:
OK. Sounds perfect.

rob006:
@samdark #14666

PR #14661, opened 2017-08-17
Coding Standard fixes

TomasVotruba:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

Few CS fixes

### DIFF
--- a/tests/framework/filters/AjaxFilterTest.php
+++ b/tests/framework/filters/AjaxFilterTest.php
@@ -10,7 +10,6 @@ namespace yiiunit\framework\filters;
 use Yii;
 use yii\base\Action;
 use yii\filters\AjaxFilter;
-use yii\web\BadRequestHttpException;
 use yii\web\Controller;
 use yii\web\Request;
 use yiiunit\TestCase;
--- a/tests/framework/i18n/FormatterTest.php
+++ b/tests/framework/i18n/FormatterTest.php
@@ -8,7 +8,6 @@
 namespace yiiunit\framework\i18n;
 
 use Yii;
-use yii\base\InvalidConfigException;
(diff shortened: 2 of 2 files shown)


samdark:
Requirements script is meant to work with PHP 5.3 and inform that 5.4 is required for the framework. Thus, short arrays syntax should not be applied there.

TomasVotruba:
I see. I thought there was bump to min PHP 7.1, that's why I sent this. Is there some other branch this would be useful in?

Isn't this handled already by PHP requirement in composer.json?

samdark:
Yii could be obtained through zip archive so composer constraints won't be triggered.

Part of your pull request (removed use statements) could be merged so if you'll remove requirements part, I'll make it part of master.

TomasVotruba:
I see. I will fix that later when not on mobile.

Thanks for explaining to me.

TomasVotruba:
Wrong button, ups

TomasVotruba:
Done!

TomasVotruba:
Btw, do you run Coding Standard checks on Travis?

I see you have "no_unused_imports" added, but this still somehow passed.

samdark:
Thanks.

It was removed from travis temporarily till we stabilize code and rules.

TomasVotruba:
You're welcome.

> It was removed from travis temporarily till we stabilize code and rules.

What happend if you took it the other way? Adding only fixers that pass (like this one).

samdark:
We'll do soon. There are other things going on so it was too distractive at the moment.

TomasVotruba:
Sure. Let me know if you need any help with cs. I love them

PR #14665, opened 2017-08-18
Fix CS

rob006:
Fix new issues reported by php-cs-fixer. I'd like to merge this before further work on CS rules, since this changes will be in every PR and will result in many conflicts.


cebe (approved):

cebe:
Thanks!

PR #14666, opened 2017-08-18
[POC] CS - force blank line before return

rob006:
See #14658

### DIFF
--- a/build/controllers/DevController.php
+++ b/build/controllers/DevController.php
@@ -149,6 +149,7 @@ class DevController extends Controller
                     }
                 } else {
                     $this->stderr("Repo argument is required for app '$app'.\n", Console::FG_RED);
+
                     return 1;
                 }
             }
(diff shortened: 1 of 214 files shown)


cebe:
> force blank line before return

why? I never understood the reasoning behind this. Makes code look weird in a lot of cases.

cebe:
seems you agree with me: https://github.com/yiisoft/yii2/pull/14658#issuecomment-323025765 ? :)

rob006:
Yes. Many changes from this PR are valid and could be merged, but as a hard rule it does not have any sense to me.

cebe:
I'd vote to not apply it, it does not really help and makes a lot of code look weird if its forced to be applied everywhere.

rob006:
I could reduce changes to "add empty line before `return` if it is preceded by `}`" - this will separate `if` blocks and `return` statements (which IMO always increase readability).

rob006:
See #14682

PR #14632, opened 2017-08-11
Refactor CAPTCHA extracting Driver

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #13799

PR #14671, opened 2017-08-19
[2.1] use `random_int()` instead of `mt_rand()` to generate cryptographically secure pseudo-random integers

yyxx9988:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | no

like `random_bytes()`, PHP 7.0 introduces `random_int()` as a core function, consider migrating.

refs:
- offical site: https://secure.php.net/manual/en/function.random-int.php
- a userland implementation for php5.2 to 5.6 https://github.com/paragonie/random_compat
- benchmark (random_int vs mt_rand) https://gist.github.com/masakielastic/0f71d119911225b708bc


FabrizioCaldarelli:
Why don't create a method  (for example, randomInt) in yii\base\Security that wraps both implementations ( below php 7 and over php 7 versions)? So this method could be used in other part of framework and applications.

Renkas:
@FabrizioCaldarelli Yii 2.1 has minimum requirenment of php 7.1 so there is no need for that.

FabrizioCaldarelli:
It could be useful also to prevent other future changes.

samdark (approved):

dynasource (approved):

framework/caching/DbCache.php
```diff
@@ -256,7 +256,7 @@ protected function deleteValue($key)
      */
     public function gc($force = false)
     {
-        if ($force || mt_rand(0, 1000000) < $this->gcProbability) {
+        if ($force || random_int(0, 1000000) < $this->gcProbability) {
```
klimov-paul: I am unsure it make sense to change `gc`. It has nothing to do with security or cryptographics.
Replacing `mt_rand` with `random_int` will degrade performance without purpose here.

framework/caching/FileCache.php
```diff
@@ -232,7 +232,7 @@ protected function flushValues()
      */
     public function gc($force = false, $expiredOnly = true)
     {
-        if ($force || mt_rand(0, 1000000) < $this->gcProbability) {
+        if ($force || random_int(0, 1000000) < $this->gcProbability) {
```
klimov-paul: Same here.

framework/mail/BaseMailer.php
```diff
@@ -249,7 +249,7 @@ public function generateMessageFileName()
     {
         $time = microtime(true);
 
-        return date('Ymd-His-', $time) . sprintf('%04d', (int) (($time - (int) $time) * 10000)) . '-' . sprintf('%04d', mt_rand(0, 10000)) . '.eml';
+        return date('Ymd-His-', $time) . sprintf('%04d', (int) (($time - (int) $time) * 10000)) . '-' . sprintf('%04d', random_int(0, 10000)) . '.eml';
```
klimov-paul: Here change also make a little sense. Email message file is meant to be created for debugging purposes and does not affect security anyhow.

klimov-paul (changes requested):

framework/caching/DbCache.php
yyxx9988: agreed.

samdark: @yyxx9988 could you change this one back then? And the one below.

yyxx9988: @samdark updated and changed one more file `UploadedFileTest.php`
```php
    private function generateFakeFileData()
    {
        return [
            'name' => md5(random_int()),
            'tmp_name' => md5(random_int()),
            'type' => 'image/jpeg',
            'size' => random_int(1000, 10000),
            'error' => 0,
        ];
    }
```

klimov-paul (approved):

yyxx9988:
closed after #14632

klimov-paul:
If you are unable to perform a merge - open a new PR.

samdark:
Merged. Thanks!

PR #14672, opened 2017-08-19
[WIP] Prefer `assertSame()` over `assertEquals()` in tests

rob006:
For testing I've enabled the `php_unit_strict` rule for php-cs-fixer, which blindly changes `assertEquals()` to `assertSame()`. Obviously this results some failed tests, but results are still quite surprising to me:

- Over 2500 assertions was changed, and this results only ~270 failed tests. 
- Many of failures relates to comparing objects where `assertSame()` does not make sense.
- Most of failures is result of ambiguous or simply invalid tests. For example:

https://travis-ci.org/rob006/yii2-dev/jobs/266138271
```
6) yiiunit\framework\caching\ArrayCacheTest::testMultiGet
Failed asserting that Array &0 (
    'number_test' => 42
    'non_existent_key' => false
) is identical to Array &0 (
    'number_test' => 42
    'non_existent_key' => null
).

/home/travis/build/rob006/yii2-dev/tests/framework/caching/CacheTestCase.php:183
/home/travis/build/rob006/yii2-dev/vendor/phpunit/phpunit/phpunit:52

7) yiiunit\framework\caching\ArrayCacheTest::testGetOrSet
Failed asserting that false is identical to null.

/home/travis/build/rob006/yii2-dev/tests/framework/caching/CacheTestCase.php:267
/home/travis/build/rob006/yii2-dev/vendor/phpunit/phpunit/phpunit:52
```

This two tests are invalid. For cache type has meaning - tests expects that misses will return `null`, but (correctly) `false` is returned. This tests passes only because invalid assertion is used. 

At first I thought that using `php_unit_strict` as hard rule does not make much sense, but now I'm not sure. Such rule really helps catch invalid assertions, and type should be testes in most (all?) of the cases. Especially since PHP 7 and `strict_types`. We just need to write some assertion for testing objects (which will work as `assertEquals()` for objects and as `assertSame()` for everything else).


tests/framework/behaviors/AttributeTypecastBehaviorTest.php
```diff
@@ -181,7 +181,7 @@ public function testSkipNotSelectedAttribute()
         $model->save(false);
 
         $model->refresh();
-        $this->assertEquals('58', $model->amount);
+        $this->assertSame(58, $model->amount);
```
rob006: Another test that looks invalid. The job of `AttributeTypecastBehavior` is ensuring that `amount` is `int`, but tests expects `string`. Am I missing something?

samdark: Looks weird.

tests/framework/data/SqlDataProviderTest.php
```diff
@@ -32,7 +32,7 @@ public function testTotalCount()
             'sql' => 'select * from `customer`',
             'db' => $this->getConnection(),
         ]);
-        $this->assertSame(3, $dataProvider->getTotalCount());
+        $this->assertSame(3, (int) $dataProvider->getTotalCount());
```
rob006: This is extremely confusing that `getTotalCount()` may return string in this case. :/

tests/framework/test/ActiveFixtureTest.php
```diff
@@ -77,11 +77,11 @@ public function testGetData()
 
         $this->assertSame(CustomerFixture::className(), get_class($fixture));
         $this->assertCount(2, $fixture);
-        $this->assertSame(1, $fixture['customer1']['id']);
+        $this->assertSame('1', $fixture['customer1']['id']);
```
rob006: I have no idea why `id` is `string` here. `profile_id` is fine bellow.

tests/framework/data/SqlDataProviderTest.php
samdark: I think we should fix the method instead of fixing test in this case...

tests/framework/test/ActiveFixtureTest.php
samdark: This is fine. ID may be string.

tests/framework/data/SqlDataProviderTest.php
rob006: #14704

klimov-paul: My opinion: your obsession of usage of `assetSame()` is go too far.
Not every comparison can be performed using it.

rob006: @klimov-paul Are you want to say that it is logical that `getTotalCount()` or `count()` should resturn string instead of integer?

klimov-paul: Absolutely: it is the natural behavior of the PDO query processing.

cebe: as long as PDO does not care about types we can not do anything useful on Yii side. Once the type is gone that information is lost and we can not guess the original type to cast it.

rob006: @cebe Any comments on #14704?

cebe:
> At first I thought that using php_unit_strict as hard rule does not make much sense, but now I'm not sure.

PHP has a type system which is not strict so we rely on that in many places, it would be wrong to try using strict comparison in all places. In general I like the idea of being more strict about asserts in tests to be really sure what is going on, but a hard rule does not make sense. PHP has two comparison operators, which should be used in different situations, so there is a place where assertEquals is appropriate.

rob006:
@cebe PHP has strict typing with `declare(strict_types=1);` and many people using it. Framework should be predictable, we should not test against integer if we know that string should be returned.

cebe:
> PHP has strict typing with declare(strict_types=1); and many people using it.

the version of Yii we are talking about here does not use that, so in our case it does not make sense to be strict about types while the code is written in a different way. In places like interaction with form input via HTTP and database via PDO we simply can not use strict types because of the limitations of these methods. In other places I am fine with testing more strictly.

rob006:
> the version of Yii we are talking about here does not use that

But people using it with Yii. If someone will use `declare(strict_types=1);` in his Yii-based project, then type of `Query::count()` result is important for him (and it is important to everyone who prefer `===` over `==`). Type of returned value should be predictable, and therefore should be tested. If you decide that `count()` should return string, then test against string, because testing against integer make this even more confusing (tests are basically documentation).

rob006:
Closing: If framework doesn't care about return types, then it does not make sense to test it. Maybe this will change some day.

I will fix these 3 invalid tests I found in another PR. Let's hope there are not too many similar problems.

PR #14679, opened 2017-08-20
Enable `ordered_class_elements` rule in php-cs-fixer config

rob006:
https://github.com/FriendsOfPHP/PHP-CS-Fixer

> **ordered_class_elements**
> Orders the elements of classes/interfaces/traits.

Configured order:
1. used traits,
2. constants,
3. properties (public first, private last),
4. constructor,
5. destructor,
6. magic methods,
7. rest of class body.

Forcing this order may be required if we want to write fixer for empty lines between class elements which follows Yii convention (no empty line between properties with the same visibility, two lines between properties and methods).


framework/base/Application.php
```diff
@@ -411,8 +414,6 @@ public function run()
      */
     abstract public function handleRequest($request);
 
-    private $_runtimePath;
-
```
cebe: for private cache variables I prefer to keep them near the methods that use them. We are following this rule quite consistently.

cebe (changes requested):

framework/base/Application.php
rob006: > Can we apply the fixer rule for everything that is not a private property?

Configured elements are positioned from the top of the class - if we don't want to change position of private properties, we won't be able to configure position of elements, that should be bellow private properties (constructor, destructor and magic methods).

I don't like current convention for private cache variables. Scattering them across the class makes it difficult to estimate what information a class instance holds. And I'm not sure what are the benefits of this approach.

boomtow:
rob006:php-cs-fixer-ordered_class_elements

rob006:
Closing, since it is outdated anyway.

PR #14681, opened 2017-08-20
Enable `phpdoc_add_missing_param_annotation` rule in php-cs-fixer config

rob006:
https://github.com/FriendsOfPHP/PHP-CS-Fixer

> **phpdoc_add_missing_param_annotation**
> Phpdoc should contain `@param` for all params.


samdark:
Merged. Thanks!

Issue #14662, opened 2017-08-17
[Enh] add properties `$contentType` and `$encoding` to `web/JsonResponseFormatter`


Faryshta:
XmlResponseFormatter has those properties so its easy to change the `Content-Type` header to

`application/hal+xml; charset=UTF-8`

But the same doesn't apply to the `json` counter part `JsonResponseFormatter` which force you to create a new class to change the `Content-Type` header to

`application/hal+json; charset=UTF-8`

as described by HAL https://tools.ietf.org/html/draft-kelly-json-hal-08#section-3

samdark:
Agree about content type. Do not agree about encoding since JSON should be UTF-8.

PR #14692, opened 2017-08-21
Add content type property to JSON response formatter

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #14662

To do:
- [x] Decide if we need to add a note to the `UPGRADE.md` since backward compatibility is technically broken in case when user extended the base formatter and added a protected or private `$contentType` property.


framework/web/JsonResponseFormatter.php
```diff
@@ -38,6 +38,11 @@
 class JsonResponseFormatter extends Component implements ResponseFormatterInterface
 {
     /**
+     * @var string|null the MIME type used in the `Content-Type` header of the response.
+     * When equals `null` default MIME type will be used based on the `useJsonp` property.
+     */
+    public $mimeType = null;
```
samdark: I'd name it the same way as in `XmlResponseFormatter`. Naming it differently would cause huge usability issue while the name `contentType` is OK.

samdark: There's no need to declare `null` as default value explicitly.

framework/web/JsonResponseFormatter.php
```diff
@@ -96,12 +101,37 @@ protected function formatJson($response)
...
+    private function getContentType($defaultMimeType)
+    {
+        $mimeType = $this->mimeType ?: $defaultMimeType;
+        $contentType = sprintf('%s; charset=UTF-8', $mimeType);
+
+        return $contentType;
+    }
+
+    /**
+     * Aquires and sets a `Content-Type` header value of a `response`.
```
samdark: a**c**quires

samdark (changes requested):

samdark:
No need for UPGRADE.md note.

framework/web/JsonResponseFormatter.php
Kolyunya: @samdark I don't like the inconsistency with the `XmlResponseFormatter` but the `contentType` is incorrect and misleading since we do not allow to set a custom `Content-Type` but just a MIME type instead.

Kolyunya: Someone may try to set it to the application/javascript; charset=UTF-8 which will result in error.

samdark: Maybe we should not automatically add `charset=UTF-8` then and leave it on developer instead?

Kolyunya: @samdark I like the idea. Refactored the code.

samdark (approved):

dynasource (approved):

SilverFire (approved):

Faryshta (approved):

framework/web/JsonResponseFormatter.php
```diff
@@ -65,13 +81,21 @@ class JsonResponseFormatter extends Component implements ResponseFormatterInterf
      */
     public $prettyPrint = false;
 
-
```
Kolyunya: Why would we delete this newline?

framework/web/JsonResponseFormatter.php
```diff
@@ -101,24 +124,20 @@ protected function formatJson($response)
...
+            && isset($response->data['data'], $response->data['callback'])
+        ) {
+            $response->content = sprintf(
+                '%s(%s);',
+                $response->data['callback'],
+                Json::htmlEncode($response->data['data'])
+            );
         } elseif ($response->data !== null) {
             $response->content = '';
-            Yii::warning("The 'jsonp' response requires that the data be an array consisting of both 'data' and 'callback' elements.", __METHOD__);
```
Kolyunya: Is this code style refactoring a subject of this patch?

framework/web/JsonResponseFormatter.php
```diff
@@ -96,12 +124,20 @@ protected function formatJson($response)
      */
     protected function formatJsonp($response)
     {
-        $response->getHeaders()->set('Content-Type', 'application/javascript; charset=UTF-8');
-        if (is_array($response->data) && isset($response->data['data'], $response->data['callback'])) {
-            $response->content = sprintf('%s(%s);', $response->data['callback'], Json::htmlEncode($response->data['data']));
+        if (is_array($response->data)
```
Kolyunya: Is this code style refactoring a subject of this patch?

Kolyunya:
I don't think this can be merged at the moment after the latest patches.

framework/web/JsonResponseFormatter.php
samdark: You mean formatting change? It's fine.

framework/web/JsonResponseFormatter.php
samdark: Yes. It is OK.

samdark:
@Kolyunya do you mean conflicts or something else?

framework/web/JsonResponseFormatter.php
Kolyunya: I mean that the subject of this patch is to add the content type property not to refactor stuff. Never mind.

Kolyunya:
@samdark I mean that there are now some changes (refactoring) in the patch which are not a subject of this patch (add content type property). If you are OK with it it's fine.

samdark:
Merged. Thanks!

PR #14694, opened 2017-08-22
Add checking on HTTP_X_FORWARDED_PROTO in getInSecureConnection method

isudakoff:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -989,7 +989,7 @@ class Request extends \yii\base\Request
      */
     public function getIsSecureConnection()
     {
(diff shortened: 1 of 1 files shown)


samdark:
It isn't there on purpose. `HTTP_` headers could be passed by client thus aren't secure and should be used if you're behind proxy only. Since https://github.com/yiisoft/yii2/pull/13780 was merged recently, you'll be able to use configure it.

PR #12101, opened 2016-08-07
added basic support for database views

igbanam:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | yes |
| Breaks BC? | no |
| Tests pass? | yes |
| Fixed issues |  |

This support only handles creating the basic view and dropping the
view. This is in response to issue #8347

An enhancement to this may be to create views with specific keywords
for specific databases.

Syntax:
- create: `$command->createView(string $name, \yii\db\Query $subquery)`
- drop: `$command->dropView(string $name)`

PR #14699, opened 2017-08-22
WIP: added basic support for database views

### LINKED ISSUE #8347: [Enh] implement functionality to work with database view
Based on issue #8152
Separate issue for implementation functionality to work with database `VIEW`.

Reference about _CREATRE VIEW_ functionality
**MySQL**
https://dev.mysql.com/doc/refman/5.0/en/create-view.html

**Cubrid**
http://www.cubrid.org/manual/93/en/sql/schema/view.html

**PostgreSQL**
http://www.postgresql.org/docs/9.3/static/sql-createview.html http://www.postgresql.org/docs/9.3/static/sql-creatematerializedview.html

**Oracle**
https://docs.oracle.com/javadb/10.8.3.0/ref/rrefsqlj15446.html
http://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_6002.htm#SQLRF01302
http://docs.oracle.com/cd/B28359_01/server.111/b28286/statements_8004.htm#SQLRF01504

Probably we have to implement not only method createView but also
- dropView
- alterView
- renameView

cebe:
This PR is a cleanup of #12101, tests are failing because params are not bound to the CREATE VIEW query. This was missing in the original PR too. Needs to addressed.

----

This support only handles creating the basic view and dropping the
view.

An enhancement to this may be to create views with specific keywords
for specific databases.

Syntax:

  - create: `$command->createView(string $name, \yii\db\Query $subquery)`
  - drop: `$command->dropView(string $name)`

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | **no**
| Fixed issues  | fixes #8347

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -802,6 +802,34 @@ class QueryBuilder extends \yii\base\BaseObject
     }
 
     /**
+     * Creates an SQL View.
+     *
+     * @param string $viewName the name of the view to be created.
+     * @param string|Query $subquery the select statement which defines the view.
+     * This can be either a string or a [[Query]] object.
+     * @return string the `CREATE VIEW` SQL statement.
+     */
+    public function createView($viewName, $subquery)
+    {
+        if ($subquery instanceof Query) {
+            list($subquery, $params) = $this->build($subquery);
+            // TODO how to deal with params here?
+        }
+        return 'CREATE VIEW ' . $this->db->quoteTableName($viewName) . ' AS ' . $subquery;
+    }
+
+    /**
+     * Drops an SQL View.
+     *
+     * @param string $viewName the name of the view to be dropped.
+     * @return string the `DROP VIEW` SQL statement.
+     */
+    public function dropView($viewName)
+    {
+        return 'DROP VIEW ' . $this->db->quoteTableName($viewName);
+    }
+
+    /**
      * Converts an abstract column type into a physical column type.
      *
      * The conversion is done using the type map specified in [[typeMap]].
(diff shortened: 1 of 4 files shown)


framework/db/QueryBuilder.php
```diff
@@ -802,6 +802,34 @@ public function dropCommentFromTable($table)
...
+     * @param string $viewName the name of the view to be created.
+     * @param string|Query $subquery the select statement which defines the view.
+     * This can be either a string or a [[Query]] object.
+     * @return string the `CREATE VIEW` SQL statement.
+     */
+    public function createView($viewName, $subquery)
+    {
+        if ($subquery instanceof Query) {
+            list($subquery, $params) = $this->build($subquery);
+            // TODO how to deal with params here?
```
samdark: TODO should be addressed.

samdark (changes requested):

framework/db/QueryBuilder.php
Faryshta: I created a class which invokes `Query::createCommand()`.

https://github.com/tecnocen-com/yii2-migrate/blob/master/src/CreateViewMigration.php#L22

It works perfectly so far.

cebe: `getRawSql` is only meant to display query for debug reasons, it is not secure in all cases.

Faryshta: in this case its code specifically written by the developer and specifically executed by the developer without anyone else adding anything.

why is it unsafe to use on such conditions?

samdark:
Closed in favor of https://github.com/yiisoft/yii2/issues/15426

adiramardiani:
Thanks @cebe, @samdark  for this feature, very usefull for me
But this feature not able in migration not like in db/Command.php and db/QueryBuilder.php

samdark:
Please create separate issue about that so we'll not forget to handle it. Thanks.

PR #14701, opened 2017-08-22
Fix #11328 PSR-7 HTTP Message Integration

### LINKED ISSUE #11328: Add support for PSR-7

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #11328, #13763


klimov-paul:
Status: under development

klimov-paul:
Status: complete

framework/UPGRADE.md
```diff
@@ -69,6 +69,17 @@ Upgrade from Yii 2.0.x
   Mail view rendering is now encapsulated into `yii\mail\Template` class.
 * Properties `view`, `viewPath`, `htmlLayout` and `textLayout` have been moved from `yii\mail\BaseMailer` to `yii\mail\Composer` class,
   which now encapsulates message composition.
+* Classes `yii\web\Request` and `yii\web\Response` have been updated to match interfaces `Psr\Http\Message\RequestInterface`
+  and `Psr\Http\Message\ResponseInterface` accordingly. Make sure you use their methods and properties correctly.
+  In particular: method `getHeaders()` and corresponding virtual property `$headers` are no longer return `HeaderCollection`
+  instance, you can use `getHeaderCollection()` in order to use old headers setup syntax; `Request|Response::$version` renamed
```
samdark: Do we still need headers collection?

framework/http/FileStream.php
```diff
@@ -0,0 +1,263 @@
...
+ * @since 2.1.0
+ */
+class FileStream extends BaseObject implements StreamInterface
+{
+    /**
+     * @var string file or stream name.
+     * Path alias can be used here, for example: '@app/runtime/items.csv'.
+     * This field can also be PHP stream name, e.g. anything which can be passed to `fopen()`, for example: 'php://input'.
+     */
+    public $filename;
```
samdark: File name seems to be mandatory. Have you thought about making it a constructor argument?

framework/http/MemoryStream.php
```diff
@@ -0,0 +1,205 @@
...
+ * $stream->write('some content...');
+ * // ...
+ * $stream->rewind();
+ * echo $stream->getContents();
+ * ```
+ *
+ * @author Paul Klimov <klimov.paul@gmail.com>
+ * @since 2.1.0
+ */
+class MemoryStream extends BaseObject implements StreamInterface
```
samdark: Is there any benefit extending from BaseObject in this case?

framework/http/MessageTrait.php
```diff
@@ -0,0 +1,301 @@
...
+    /**
+     * Return an instance with the specified HTTP protocol version.
+     *
+     * @param string $version HTTP protocol version
+     * @return static
+     */
+    public function withProtocolVersion($version)
+    {
+        $this->setProtocolVersion($version);
+        return $this;
```
samdark: According to PSR, request and response methods are immutable i.e. it should be something like the following:

```php
public function withProtocolVersion($version)
{
    $newThis = clone $this;
    $newThis->setProtocolVersion($version);
    return $newThis;
}
```

framework/UPGRADE.md
klimov-paul: Yes.
The class itself is useful helper to maintain the PSR headers setup.
Also with 'immutability' of the `MessageInterface`, there is no actual interface for headers setup without instantiating new message instance, while it is required for the web filters.

framework/http/FileStream.php
klimov-paul: I would prefer to keep setup via `$config` array.
In this way instance will be configurable via regular application config if necessary.

framework/http/MemoryStream.php
klimov-paul: The benefit is having consistent restrictions on setting/getting unexisting properties and calling of unexisting methods.

framework/http/MessageTrait.php
klimov-paul: Fixed.

framework/UPGRADE.md
samdark: Why should web filters work with the objects in a mutable way differently than other objects?

framework/http/MemoryStream.php
samdark: I think current PHP have these for basic objects not extended from anything as well...

framework/UPGRADE.md
klimov-paul: They are implemented in this way. Unless you provide the idea how thier interface to be changed allowing override `Application::$response`, feel free to propose it.
See `HttpCache` for example: it needs to setup 'Cache-Control', 'Etag' and 'Last-Modified' headers to the application response.

framework/http/MemoryStream.php
klimov-paul: Check following code:

```php
class Foo {}

class FooTest extends TestCase
{
    public function testClassFields()
    {
        $object = new Foo();
        $object->name = 'test';
        $this->assertSame('test', $object->name);
    }
}
```

Such unit test passes correctly even on PHP 7.1

samdark: OK.

framework/UPGRADE.md
samdark: How about making a `ActionFilter`'s `ActionEvent` to have response instance? Then it would be possible to use regular immutable methods.

klimov-paul: I suppose this matter is out of the scope of this PR.
It should be considered in the scope of #13922 afterwards.

samdark: OK.

samdark:
Looks like some tests aren't OK.

klimov-paul:
Build fixed.

framework/http/MemoryStream.php
```diff
@@ -0,0 +1,205 @@
...
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function write($string)
+    {
+        $size = $this->getSize();
+        $writeSize = strlen($string);
```
dmirogin: This is going to work correctly with multibyte encoding ?

framework/http/ResourceStream.php
```diff
@@ -0,0 +1,237 @@
...
+use yii\base\BaseObject;
+use yii\base\ErrorHandler;
+
+/**
+ * ResourceStream wraps existing PHP stream resource, e.g. one opened by `fopen()`.
+ *
+ * Example:
+ *
+ * ```php
+ * $stream = new FileSteam([
```
dmirogin: ResourceStream you meant :smiley:

framework/http/MemoryStream.php
bashkarev: `strlen` returns the number of bytes, here everything is correct

dmirogin: Yeah, indeed. Sorry.

framework/http/FileStream.php
samdark: I don't think that it's a good reason to affect class design. If application can't configure pretty much regular PHP class with constructor then we need to address it in the application.

samdark: Mode is, as well, mandatory so it should be constructor argument.

framework/http/FileStream.php
```diff
@@ -0,0 +1,263 @@
...
+ * ]);
+ *
+ * $stream->write('some content...');
+ * $stream->close();
+ * ```
+ *
+ * @author Paul Klimov <klimov.paul@gmail.com>
+ * @since 2.1.0
+ */
+class FileStream extends BaseObject implements StreamInterface
```
samdark: There's no benefit in extending from `BaseObject` in this case.

framework/http/FileStream.php
```diff
@@ -0,0 +1,263 @@
...
+    private $_resource;
+    /**
+     * @var array a resource metadata.
+     */
+    private $_metadata;
+
+
+    /**
+     * Destructor.
+     * Closes the stream resource when the destructed.
```
samdark: When destroyed.

framework/http/FileStream.php
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isWritable()
+    {
+        return in_array(
+            $this->getMetadata('mode'),
+            ['w', 'w+', 'rw', 'r+', 'x+', 'c+', 'wb', 'w+b', 'r+b', 'x+b', 'c+b', 'w+t', 'r+t', 'x+t', 'c+t', 'a', 'a+'],
```
samdark: How about `c`? There are also extra flags such as `e`. 

Either modes should be limited and checked for validity in constructor or checks should include all possible configurations. It worth searching for sub-strings then: `w`, `r+`, `x+`, `c`, `a`.

framework/http/FileStream.php
samdark: Another argument for making these constructor arguments and private is that changing these values after opening a stream makes no sense.

framework/filters/PageCache.php
```diff
@@ -208,10 +208,15 @@ public function afterRestoreResponse($data)
...
+        foreach (['format', 'protocolVersion', 'statusCode', 'reasonPhrase', 'content'] as $name) {
             $response->{$name} = $data[$name];
         }
-        foreach (['headers', 'cookies'] as $name) {
+
+        if (isset($data['headers'])) {
+            $response->setHeaders($data['headers']);
+        }
+
+        foreach (['cookies'] as $name) {
```
samdark: What `foreach` is for here?

samdark (approved):

framework/http/FileStream.php
klimov-paul: > I don't think that it's a good reason to affect class design. If application can't configure pretty much regular PHP class with constructor then we need to address it in the application.

I would like to keep a consistentcy comparing other Yii classes, which in most cases do not introduce constructor parameters.
Also it is actually unlikely developer will create these stream objects anywhere manually.
For the file response sending he will likely use `Response::sendFile()`.
I do not think it is a crucial.

> argument for making these constructor arguments and private is that changing these values after 
opening a stream makes no sense.

Same can be said for example for [yii\db\Connection::$dsn](https://github.com/yiisoft/yii2/blob/master/framework/db/Connection.php#L164): after connection opened changing of the `$dsn`, `$username` or `$password` make no effect and no sense.

framework/http/FileStream.php
klimov-paul: Same as [here](https://github.com/yiisoft/yii2/pull/14701#discussion_r135021442)

samdark: We should finally extract it into to traits :( Both aren't good enough reasons to keep inheritance tree as big as it is now.

framework/http/FileStream.php
samdark: > I do not think it is a crucial.

It is not but:

1. It's not good to develop this way with mandatory dependencies set via public properties. Makes too much room for errors.
2. It serves as an example and it's a bad example. I know we have that basically everywhere but we have to start with something.
3. It's a good test for our DI container capabilities.

> Same can be said for example for yii\db\Connection::$dsn: after connection opened changing of the $dsn, $username or $password make no effect and no sense.

Yes. *Ideally* it should be changed as well...

framework/http/FileStream.php
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isReadable()
+    {
+        return in_array(
+            $this->getMetadata('mode'),
+            ['r', 'w+', 'r+', 'x+', 'c+', 'rb', 'w+b', 'r+b', 'x+b', 'c+b', 'rt', 'w+t', 'r+t', 'x+t', 'c+t', 'a+'],
```
samdark: Same checking issues as above.

framework/http/MessageTrait.php
```diff
@@ -0,0 +1,347 @@
...
+
+namespace yii\http;
+
+use Psr\Http\Message\StreamInterface;
+use yii\di\Instance;
+
+/**
+ * MessageTrait provides set of methods to satisfy [[\Psr\Http\Message\MessageInterface]].
+ *
+ * This trait should be applied to descendant of [[\yii\base\BaseObject]] implementing [[\Psr\Http\Message\MessageInterface]].
```
samdark: Why `descendant of [[\yii\base\BaseObject]]` is a requirement here?

framework/http/ResourceStream.php
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isWritable()
+    {
+        return in_array(
+            $this->getMetadata('mode'),
+            ['w', 'w+', 'rw', 'r+', 'x+', 'c+', 'wb', 'w+b', 'r+b', 'x+b', 'c+b', 'w+t', 'r+t', 'x+t', 'c+t', 'a', 'a+'],
```
samdark: Same issue with mode checking as above.

framework/http/ResourceStream.php
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isReadable()
+    {
+        return in_array(
+            $this->getMetadata('mode'),
+            ['r', 'w+', 'r+', 'x+', 'c+', 'rb', 'w+b', 'r+b', 'x+b', 'c+b', 'rt', 'w+t', 'r+t', 'x+t', 'c+t', 'a+'],
```
samdark: Same issue with mode checking as above.

framework/http/MessageTrait.php
klimov-paul: Otherwise virtual properties declared in trait's PHPDoc will not appear.

PR #14703, opened 2017-08-22
Fix \yii\data\Sort

lav45:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #13260

Tested this functionality on a production project

### DIFF
--- a/framework/data/Sort.php
+++ b/framework/data/Sort.php
@@ -229,7 +229,7 @@ class Sort extends BaseObject
                     $orders[$name] = $dir;
                 }
             } else {
-                $orders[] = $columns;
+                $orders = $columns;
             }
         }
 
--- a/framework/data/Sort.php
+++ b/framework/data/Sort.php
@@ -214,7 +214,8 @@ class Sort extends BaseObject
     /**
      * Returns the columns and their corresponding sort directions.
      * @param bool $recalculate whether to recalculate the sort directions
-     * @return array the columns (keys) and their corresponding sort directions (values).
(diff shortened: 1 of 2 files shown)


samdark:
What issue is it fixing? The one referenced was taken care of by @klimov-paul already, wasn't it?

framework/data/Sort.php
```diff
@@ -229,7 +229,7 @@ public function getOrders($recalculate = false)
                     $orders[$name] = $dir;
                 }
             } else {
-                $orders[] = $columns;
+                $orders = $columns;
```
cebe: in this case `columns` is not an array. The return type of this method is `array`, so assigning columns to orders does not make sense.

cebe (changes requested):

framework/data/Sort.php
LAV45: @cebe, updated

cebe: changing the return type may break code that relies on current behavior. SqlDataProvider does expect an array for example: https://github.com/yiisoft/yii2/blob/4081a4b20349974ea9f37957b166fa4398cf3a8b/framework/data/SqlDataProvider.php#L121-L126

please elaborate on why you think this change should be made. It looks wrong to me.

samdark: Same here. I don't see the reason for the change...

PR #14704, opened 2017-08-22
[POC] Return int where possible in statistical methods in `yii\db\Query`

rob006:
See https://github.com/yiisoft/yii2/pull/14672#discussion_r134482371

This should ensure that methods like `Query::count()` will return integer if result from DB can be safely converted to integer.

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -327,7 +327,7 @@ class Query extends Component implements QueryInterface
             return 0;
         }
 
-        return $this->queryScalar("COUNT($q)", $db);
+        return $this->toInt($this->queryScalar("COUNT($q)", $db));
     }
 
     /**
@@ -344,7 +344,7 @@ class Query extends Component implements QueryInterface
             return 0;
         }
 
-        return $this->queryScalar("SUM($q)", $db);
+        return $this->toInt($this->queryScalar("SUM($q)", $db));
     }
 
     /**
@@ -361,7 +361,7 @@ class Query extends Component implements QueryInterface
             return 0;
         }
 
-        return $this->queryScalar("AVG($q)", $db);
+        return $this->toInt($this->queryScalar("AVG($q)", $db));
     }
 
     /**
@@ -374,7 +374,7 @@ class Query extends Component implements QueryInterface
      */
     public function min($q, $db = null)
     {
-        return $this->queryScalar("MIN($q)", $db);
+        return $this->toInt($this->queryScalar("MIN($q)", $db));
     }
 
     /**
@@ -387,7 +387,19 @@ class Query extends Component implements QueryInterface
      */
     public function max($q, $db = null)
     {
-        return $this->queryScalar("MAX($q)", $db);
+        return $this->toInt($this->queryScalar("MAX($q)", $db));
+    }
+
+    /**
+     * Converts string value to int if possible.
+     *
+     * @param string $value
+     * @return int
+     */
+    private function toInt($value)
+    {
+        $int = (int) $value;
+        return (string) $int === $value ? $int : $value;
     }
 
     /**


framework/db/Query.php
```diff
@@ -344,7 +344,7 @@ public function sum($q, $db = null)
             return 0;
         }
 
-        return $this->queryScalar("SUM($q)", $db);
+        return $this->toInt($this->queryScalar("SUM($q)", $db));
```
klimov-paul: `SUM`, `AVG`, `MIN` and `MAX` are availblr for the float values just as well.
Thier result can not be converted to integer.

framework/db/Query.php
```diff
@@ -387,7 +387,19 @@ public function min($q, $db = null)
...
+    /**
+     * Converts string value to int if possible.
+     *
+     * @param string $value
+     * @return int
+     */
+    private function toInt($value)
+    {
+        $int = (int) $value;
+        return (string) $int === $value ? $int : $value;
```
klimov-paul: Integer typecast may be unsafe as it may overflow PHP max int restiction.
Comparing `(string) $int === $value` is a bad idea. How it will handle values like `5E91`?

klimov-paul (changes requested):

rob006:
@klimov-paul Just read this code before commenting. It will not touch values that are not direct representation of integers (in decimal notation). `(string) (int) $value === $value` will not pass for floats or `5E91`.

klimov-paul:
> will not pass for floats or 5E91.

So what does i solves then? This means return type of `count()` is undetermined: it could be `int` and could be not `int` - it all depends of how big the result number is.

klimov-paul:
This does not look like a fix - more likely like a hack.

While calling `MIN`, `MAX` or `AVG` devloper should avare of which return type to expect and perform typecast on his own if needed. This fix simply make behavior of the methods inconsistent as they return value of random type instead of single stable one.

rob006:
It will be always int if it stores value that can be represented as int. Conditions like `<some int> === $query->count()` will always work as expected.

cebe:
This does not make sense, PDO returns strings, it is error prone to try to guess a type from the value itself and apply a cast. PHP has `==` for a reason. The same thing exists with data sent via HTML forms, all data is string there, also `==` works fine in that case.

klimov-paul:
Also note that some DBMS allows operators `MIN` and `MAX` over the string column. The particular value stored in such column can be `12`, which is stored as a string, but will be automatically typecasted to `int`.

cebe:
closing, as this really is not a good idea.

rob006:
Can we do this only for `count()`? I can't imagine scenario when this could be unsafe.

cebe:
overflow case, where the result of count() is bigger than PHP int?

rob006:
Then `(string) (int) $value === $value` will not pass and string will be returned.

rob006:
https://3v4l.org/Ca9DV

cebe:
> and string will be returned.

so what do we win here? we get inconsistent behavior and you still can not strictily test against integer.

klimov-paul:
> this "inconsistent behavior" works exactly the same for bigint in ActiveRecord.

No, it does not.
For the 'bigint' typecast is performed for 64-bit OS, while for 32 it is not. Typecast is performed no matter of the actual size of particular value, e.g. bigint `1` will be string anyway.

rob006:
This is even worse and more inconsistent. In one environment I will get `1` as integer and on another as a string...

klimov-paul:
> This is even worse and more inconsistent. In one environment I will get 1 as integer and on another as a string...

I can not agree. This is done in order to suport old systems, which have no 64-bit support.
If your environments varies thus far, you will get much more inconstistencies anyway.

rob006:
`1` can be represented as integer in 32-bit systems, so leaving string for them is clearly overkill...

klimov-paul:
We do not perform any datadase return value typecasting anywhere but for ActiveRecord attributes.
I can see no reason this policy should be be changed.

rob006:
> I can see no reason this policy should be be changed.

As always...

Anyway, if string is obvious type of `count()` result, then it should return `0` as string when `emulateExecution` is enabled. And docs of `ActiveDataProvider` needs to be adjusted - it suggest integer everywhere for count-like methods.

rob006:
> Guessing a type from a raw string value is error prone.

Can you give just one example when result of `count()` can give some incorrect results is we use `(string) (int) $value === $value` check before casting?

klimov-paul:
> Can you give just one example when result of count() can give some incorrect results is we use (string) (int) $value === $value check before casting?

This can be rephrased in other way: how you can prove this code works in **absolutely all** cases without nay risk?

rob006:
@klimov-paul Please define "absolutely all cases" and I will prove that for you with unit tests. :)

Related topics: 
https://en.wikipedia.org/wiki/Occam%27s_razor
https://en.wikipedia.org/wiki/Russell%27s_teapot

cebe:
> Can you give just one example when result of count() can give some incorrect results is we use (string) (int) $value === $value check before casting?

The code alone might probably work, the problem I see is more general. If the underlying layer does not support proper types and we try to guess them in some places, we add complexity to the code without gaining anything. If your asInt method can still return a string it does not make sense to add it in the first place. In the world of PHP 7 strict types such an addition might be useful to be able to work with the database results more easily. In the current state it does not make sense to me.

rob006:
> I do not think it is an improvement if one method behaves differently.

You already have one method that works differently: `exists()` typecasts results of `queryScalar()` and returns boolean instead of string.
Returning string as result of `count()` also creates inconsistency in higher-level classes. `getTotalCount()` in data providers should return integer, but `ActiveDataProvider::getTotalCount()` returns string.

cebe:
I just tried the following:

```
        $count = (new Query())->from('customer')->count('*', $db);
        var_dump($count);
```

result: 

```
mysql: string(1) "3"
pgsql: int(3)
sqlite: string(1) "3"
```

there are PDO drivers that work correctly and there are PDO drivers that do not. Its not even a Yii issue.

rob006:
> Its not even a Yii issue.

I thought this was one of the main tasks of the `Query` class - creating an abstraction layer that could be used to handle different database engines in the same way.

And obviously Laravel is more pragmatic in this case: 
https://github.com/illuminate/database/blob/1265e6d207188abc197aedc2570d889a7a8e32a3/Query/Builder.php#L1942-L1951

samdark:
It's more than possible to store more than 2147483647 (max int on 32 bit system) rows in SQL table. In MySQL it's not so performant, MS SQL Server is much better. So it seems Laravel is doing it wrong and a very big project one day will run into serious issues.

rob006:
That's why I'm not asking to casting this blindly like Laravel, but only for values which can be safely represented as integer. Then we have more natural behavior in 99.9999999% cases and still handle correctly edge cases with big ints on 32-bit systems.

samdark:
It won't allow us to specify return type as int or rely on it in any way...

Issue #14707, opened 2017-08-23
Documentation error in  yii\i18n\Formatter


therendstudio:
In the description **asDatetime()** function about the **$format** parameter it says that the default value is taken from **$dateFormat** while the correct is **$datetimeFormat**

[http://www.yiiframework.com/doc-2.0/yii-i18n-formatter.html#asDatetime()-detail](url)

Yii Version: **2.0**

cebe:
fixed, thanks for reporting!

PR #14708, opened 2017-08-23
Formatter asDatetime() documentation $format fix

karataserkan:
In the description asDatetime() function about the $format parameter it says that the default value is taken from $dateFormat while the correct is $datetimeFormat

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14707


cebe:
Thank you!

PR #14713, opened 2017-08-24
Fix \yii\filters\RateLimiter

tymoteusz-be:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

If `loadAllowance` returns current time, e.g.:
```
function() {
    sleep(10);
    return [100, time()];
}
```
Calculation `($current - $timestamp)` is always < 0, resulting in adding negative number to `$allowance`.

### DIFF
--- a/framework/filters/RateLimiter.php
+++ b/framework/filters/RateLimiter.php
@@ -106,10 +106,10 @@ class RateLimiter extends ActionFilter
      */
     public function checkRateLimit($user, $request, $response, $action)
     {
-        $current = time();
-
         list($limit, $window) = $user->getRateLimit($request, $action);
         list($allowance, $timestamp) = $user->loadAllowance($request, $action);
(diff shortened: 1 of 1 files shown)


samdark:
Is that the same as #15432?

samdark:
The fix looks weird and it's not clear how to reproduce the issue. @tymoteusz-be would you please clarify it?

Closing the issue for now... (doesn't mean it can't be re-opened).

bizley:
@samdark It's not the same.

I don't think this is reasonable concern. You can basically cast this problem in literally any place in the code so what is the point of just "fixing" this little case? If one of the components adds some non-negligible delay different approach is required.

But to be fair this proposed change can help a bit and brakes nothing.

PR #14714, opened 2017-08-24
Adjust QueryInterface and QueryTrait types on where()

### LINKED ISSUE #14496: ActiveRecordInterface
SpellChecker in IDE mark to problems lines like this:
```php
     * @param string|array $condition the conditions that will be put in the WHERE part of the UPDATE SQL.
    public static function updateAll($attributes, $condition = '', $params = [])
    {
```

ActiveRecordInterface can except $condition as a string

### What do you get instead?
only null|array acceptable
```php
     * @param array $condition the condition that matches the records that should get updated.
    public static function updateAll($attributes, $condition = null);
```

https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveRecordInterface.php#L261

https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveRecordInterface.php#L278

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | dev

cebe:
these did never allow a string condition. Only SQL Query and AR implementation do.

fixes #14496
reverts https://github.com/yiisoft/yii2/commit/79ad74f763229e84c298cf5f5d3c0f59f758aaf1

### DIFF
--- a/framework/db/ActiveRecordInterface.php
+++ b/framework/db/ActiveRecordInterface.php
@@ -254,7 +254,7 @@ interface ActiveRecordInterface
      *
      * @param array $attributes attribute values (name-value pairs) to be saved for the record.
      * Unlike [[update()]] these are not going to be validated.
-     * @param string|array $condition the condition that matches the records that should get updated.
+     * @param array $condition the condition that matches the records that should get updated.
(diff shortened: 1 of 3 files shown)


SilverFire (approved):
I'm fine with it

PR #14719, opened 2017-08-24
Add phpdoc throws

bscheshirwork:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no


samdark:
Merged. Thanks!

rob006:
I'm against such changes. These annotations are not only useless (see https://github.com/yiisoft/yii2/pull/14374#issuecomment-312507991), but actually gives more harm than good. I'm using unhandled exceptions inspection in PHPStorm, and methods with annotations like these are really annoying. You can't do anything practical with it - these exceptions are to general to catch and handle it. The only thing you can do with them is just ignoring. So what is the point of annotations that should be ignored?

bscheshirwork:
Your own PHPStorm make this annotation after find \Exception and \Trowable inside a function body. 

If not include it the PHPDoc is mark as wrong.

PHPdoc tell about function directly. You see function PHPDoc (not body!) in many cases. Without throws you not get any info about fact of exception from function

You use one of PHPStorm tools and you get annoying with it. May be can make this tools better? Config it? Not code?

bscheshirwork:
If you against of common \Exception and \Trowable - may be need concrete class?

rob006:
> If you against of common \Exception and \Trowable - may be need concrete class?

I'm open to suggestions, but I'm pretty sure that you don't know what these exceptions means. Nobody knows - they're caught and rethrown "just in case", they don't have any meaning. Meaningless exceptions should not be annotated.

bscheshirwork:
Yep, this is "just in case".

I think for nested ob_start in third-part `$_file_` for example. For get buffer / get down to current level /before throw error next.

bscheshirwork:
So... this functional must be ignored?

PR #14720, opened 2017-08-24
Remove unnecessary parentheses

thiagotalma:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no

The generated sql queries have an excess of parentheses that bothers us when we need to debug.

What do you think about reducing the excess of unnecessary parentheses in sql queries a bit?

This PR is just a suggestion, other parentheses can also be removed.

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -1258,7 +1258,7 @@ class QueryBuilder extends \yii\base\BaseObject
             }
         }
         if (!empty($parts)) {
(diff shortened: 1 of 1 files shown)


samdark:
It is necessary. For example, operator could be AND and parts could be `A OR B` and `C OR D`. Without parentheses it will be:

```
A OR B AND C OR D
```

with parentheses:

```
(A OR B) AND (C OR D)
```

Issue #14663, opened 2017-08-18
Convert int to string when execute insert query if the type of mysql column is 'bigint unsigned'


bittenApple:
### What steps will reproduce the problem?
1. Create a MySQL table and mapped ActiveRecord.
```
create table foo(id bigint unsigned);
```
Save a model
```
    $f = new Foo;
    $f->id = 1;
    $f->save();
```
### What is the expected result?
The  query should be
```
INSERT INTO `foo` (`id`) VALUES (1)
```
### What do you get instead?

The executed query is
```
INSERT INTO `foo` (`id`) VALUES ('1')
```
The range of MySQL unsigned bigint is larger than PHP_INT_SIZE, but it's not reasonable to convert int attribute manually set to string when execute sql, which may slow the query performance.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.1.3
| Operating system | CentOS 7
| MySQL | 5.7.16

FabrizioCaldarelli:
Have you checked that it is not related to PDO ?

yyxx9988:
This is not a bug.

The maximum int value of php depends on the system. 32 bit systems have a maximum signed integer range of -2147483648 to 2147483647. So for example on such a system, intval('1000000000000') will return 2147483647.

see http://php.net/manual/en/language.types.integer.php for more details

bittenApple:
> This is not a bug.
> 
> The maximum int value of php depends on the system. 32 bit systems have a maximum signed integer range of -2147483648 to 2147483647. So for example on such a system, intval('1000000000000') will return 2147483647.
> 
> see http://php.net/manual/en/language.types.integer.php for more details


I think it's the programmer's responsibility to take care of the risk of int overflow.
If the framework user set the int value of a attribute explicitly,  I don't think it's appropriate to cast string to int implicitly by framework

cebe:
> The executed query is

how did you verify the type of the value bound on query execution time? These are bound as parameter values and are converted inside of PDO, Yii has no logic to convert int to string by default in that case. I can not reproduce the issue. Please provide the CREATE TABLE statement of your table.

bittenApple:
> how did you verify the type of the value bound on query execution time

I find some slow query in my production MySQL slow log which the type of value is not as expected. Maybe the performance of type casting is not the key reason for mysql, but I still think it's better not to cast the type of value if user set it explicitly.


[...]

PR #14722, opened 2017-08-25
Don't cast int to string if the value type is set as int, #14663

bittenApple:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14663

### DIFF
--- a/framework/db/ColumnSchema.php
+++ b/framework/db/ColumnSchema.php
@@ -107,6 +107,7 @@ class ColumnSchema extends BaseObject
     /**
      * Converts the input value according to [[phpType]] after retrieval from the database.
      * If the value is null or an [[Expression]], it will not be converted.
+     * If the value is int and the dbType is int, bigint or smallint, it will not be converted
      * @param mixed $value input value
      * @return mixed converted value
      * @since 2.0.3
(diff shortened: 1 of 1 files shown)


Faryshta:
this is done when `$phpType` is an string. Which use cases happen to have `$phpType` equals 'string' and `$dbType` equals 'integer'?

bittenApple:
> this is done when $phpType is an string. Which use cases happen to have $phpType equals 'string' and $dbType equals 'integer'?

Yes, I think it is necessary to cast int to string in this case

Faryshta:
Your patch is done inside a `switch` statement  for the case when `$phpType` equals 'string'.

bittenApple:
> Your patch is done inside a switch statement for the case when $phpType equals 'string'.

Sry, I don't get your point, do you mean it's better not to set $phpType to 'string' when the db field is 'unsigned bigint'?

Faryshta:
currently there is no bigint support in php https://wiki.php.net/rfc/bigint http://php.net/manual/es/language.types.integer.php

basically they get turned into `float` with uncertain behavior. using bigint would require you to use external libraries such as http://php.net/manual/es/function.gmp-intval.php

```php
public function getIntValBigNumber()
{
       return gmp_intval($this->big_number);
}
```

i am closing this pr since its outside yii2 scope.

rob006:
@Faryshta Pleas read #14663. This is about unnecessary casting integers to string on query generation. PHP bigint support has nothing to do with it.

rob006:
I think that `ColumnSchema::phpTypecast()` and `ColumnSchema::dbTypecast()` should use different implementation. Right now both uses `typecast()` method and changes intended for casting values for generating db query affects casting query results of this query. This may result some weird side effects.

PR #14729, opened 2017-08-28
Update Query.php

harloprillar:
Its need to return static object in queryScalar() method last return statement because inherited createCommand() invoke QueryBuilder::build() method, which may be reloaded in inherited QueryBuilder class.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  |

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -446,7 +446,7 @@ class Query extends Component implements QueryInterface
             return $command->queryScalar();
         }
 
-        return (new self())
+        return (new static())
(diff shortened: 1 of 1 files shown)


eastern1984:
How did you find this bug?

cebe:
this was created to create a plain Query even when called from ActiveQuery. The change is not correct. If there is an issue, report it with details on how to reproduce.

PR #14745, opened 2017-09-01
ensure backwards compatibiliy with migration without $compact

cebe:
See https://github.com/yiisoft/yii2/pull/14474#discussion_r135237472

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | https://github.com/yiisoft/yii2/pull/14474#discussion_r135237472

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -8,6 +8,7 @@
 namespace yii\console\controllers;
 
 use Yii;
+use yii\base\BaseObject;
 use yii\base\InvalidConfigException;
 use yii\base\NotSupportedException;
 use yii\console\Controller;
@@ -759,7 +760,11 @@ abstract class BaseMigrateController extends Controller
     protected function createMigration($class)
     {
         $this->includeMigrationFile($class);
-        return new $class(['compact' => $this->compact]);
+        $migration = new $class();
+        if ($migration instanceof BaseObject && $migration->canSetProperty('compact')) {
+            $migration->compact = $this->compact;
+        }
+        return $migration;
     }
 
     /**


framework/console/controllers/BaseMigrateController.php
```diff
@@ -759,7 +760,11 @@ protected function migrateDown($class)
     protected function createMigration($class)
     {
         $this->includeMigrationFile($class);
-        return new $class(['compact' => $this->compact]);
+        $migration = new $class();
+        if ($migration instanceof BaseObject && $migration->canSetProperty('compact')) {
+            $migration->compact = $this->compact;
+        }
+        return $migration;
```
klimov-paul: Such fix makes a little sense. It is better to simply avoid passing `compact` to constructor here.
In case particular migration implementation supports `compact` it can overridde `createMigration` just in the way `MigrateController` for regular DB does.

klimov-paul (changes requested):

framework/console/controllers/BaseMigrateController.php
cebe: if it does not hurt, why not try to set it?

klimov-paul: In case there is no such field in target class - it will cause `UnknownPropertyException`.

klimov-paul: We can not rely that **any** custom implementation overrides the `createMigration()` method.
It may do so, and it may do not.

samdark: @klimov-paul if we won't pass it then it won't be settable from command line.

cebe: My fix will only set it, if `canSetProperty` returns true.

samdark (approved):

framework/console/controllers/BaseMigrateController.php
klimov-paul: > if we won't pass it then it won't be settable from command line.

If migration class does not suport - it will not be settiable from command line anyway.
That was a mistake at the first place - this option should not belong to migration object, but should belong to some 'view' layer which performs the output. I have told you about this [here](https://github.com/yiisoft/yii2/pull/14474#issuecomment-317697935).

In current state this option may be inconsistent at some cases, but there is nothing we can do.

klimov-paul: > My fix will only set it, if canSetProperty returns true.

How does it changes anything? If particular `Migration` class have no property `$compact` you can not set it from command line. Unless you add it somehow to `MigrationInterface` you can not rely on it.
At the present state only `yii\db\Migration` can accept this property. Any externla library is not aware of it.

klimov-paul: Also what will happen in case particualr Migration class does not extend `BaseObject` but has `$compact` field?

cebe: > At the present state only yii\db\Migration can accept this property. Any externla library is not aware of it.

but is can be added by just adding the property, instead of the need to extend this method.



> Also what will happen in case particualr Migration class does not extend BaseObject but has $compact field?

nothing.

klimov-paul: > nothing

So it is inconsistent anyway.

klimov-paul: > but is can be added by just adding the property ... nothing

So, in case I have NOT extended my migration class from `BaseObject` and add a property `$compact` as you suggets - it will not applied.

cebe: in the Yii eco system it is very uncommon to not extend from BaseObject, so I do not think this is a problem. This addition is made for convenience when implementing Migration, its not meant to cover any extra special use case that may not even exist.

klimov-paul: >  I do not think this is a problem

I can not see any problem to have `BaseMigrateController::createMigration()` without `$compact` either. So?
Why do we need to exploit usage of `canSetProperty()` at this method, creating a bad example for the developers? It only justifies solutions like #12929.

At the present state `canSetProperty()` method used only around the behavior mechanism and can be considered as internal one. Its usage is unnatural for the PHP as uses `isset()` or `property_exists()` for such cases. I would prefer do not expose it anywhere besides setter/getter realted to behaviors.

SilverFire (approved):

schmunk42 (approved):
LGTM

PR #14750, opened 2017-09-01
Fix #14749: Ensured garbage collections for the widgets with behaviors attached to them

### LINKED ISSUE #14749: Widget garbage collection

Relates to #14102

Since now widgets may have event handles and behaviors attached to them.
This cause a memory leaks while widget usage as widget instance and behavior introduce a cycle reference.
Thus any usage of the Widget with attached behavior causes memory leak.

```php
AnyWidget::widget([
    'id' => 'test',
    'as test' => [
        'class' => TestWidgetBehavior::className()
    ],
]);
```

Widget instance to be destroyed.
Memory usage stays the same after `Widget::widget()` or `Widget::end()`.

### What do you get instead?

Widget instance remains in memory consuming space without purpose.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      |  *
| Operating system | *

Unit test to reproduce:

```php
class TestWidget extends Widget
{
    public static $instanceCount = 0;

    public function __construct($config = [])
    {
        parent::__construct($config);
        static::$instanceCount++;
    }

    public function __destruct()
    {
        static::$instanceCount--;
    }
}

class WidgetTest extends TestCase
{
    public function testGc()
    {
        TestWidget::widget([
            'id' => 'test',
            'as test' => [
                'class' => TestWidgetBehavior::className()
            ],
        ]);
        $this->assertSame(0, TestWidget::$instanceCount);
    }
}
```

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14749

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.13 under development
 ------------------------
 
+- Enh #14749: Ensured garbage collections for the widgets with behaviors attached to them (klimov-paul)
 - Bug #14723: Fixed serialization of `yii\db\Connection` instance closes database connection (klimov-paul)
 - Bug #14697: Fixed `console\widgets\Table` rendering when there's no data supplied (bscheshirwork)
 - Enh #13853: Added `yii\db\Migration::$compact` as well as `yii\console\controllers\BaseMigrateController::$compact` to allow making the migration console output more compact (francislavoie)
--- a/framework/base/Widget.php
+++ b/framework/base/Widget.php
@@ -110,6 +110,7 @@ class Widget extends Component implements ViewContextInterface
                     $result = $widget->afterRun($result);
                     echo $result;
                 }
+                $widget->detachBehaviors(); // ensure garbage collection
 
                 return $widget;
             }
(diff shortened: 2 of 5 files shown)


samdark (approved):

rob006:
This adds some magic to the widgets. Since `end()` methods returns widgets instance, it is completely unsuspected that this instance will have all behaviors detached.

samdark:
Oh... you're right. @klimov-paul how about doing the same in `__destruct()`?

rob006:
> how about doing the same in __destruct()?

It will not work - `__destruct()` is called after GC finds and removes object from memory.

samdark:
Right. Any suggestions then?

rob006:
I suggest do nothing. This is not a real memory leak - GC will finally clean these objects, and detaching behaviors like this may create only unnecessary overhead. Such optimizations have some pros and cons, and I think it should be done at application level. Yii could just rely on PHP GC.

samdark:
Yeah. In this case implications are quite serious so I don't think this pull request should be merged.

PR #14756, opened 2017-09-03
Allow get non-exists attributes of DynamicModel

MuVO:
Currently I have an issue when used DynamicModel in case of ad-hoc validation, when creating request-based «virtual» model: if I passed request data which must be validated but request does not contains some required attributes, my application will throws an exception because validator tried to validate non-existent attributes.
Use case:
```php
    public function actionIndex()
    {
        $rules = [
            [['id', 'email'], 'required'],
            ['id', 'integer'],
            ['name', 'string'],
            ['email', 'email'],
        ];

        $preModel = DynamicModel::validateData(\Yii::$app->request->bodyParams, $rules);

        /*
         * some code 
         */
    }
```

If HTTP form wouldn't contain any of attributes, which enumerated in rules this cause fatal exception in application instead of fills of "errors" of dynamic model.

I think this is not right behavior of this class. Dynamic model is a «virtual» model, therefore its logic could have some differs from basic model. For example, primary purpose of dynamic model — ad-hoc data validation. But actually we can't correctly verify source object, if this object hasn't some attributes, if even "required" validators were created.

I offer following solution of this issue: DynamicModel must always "correctly" handling a getting of *any* attributes. If getting of non-exsting attribute called, it just return an empty null-value. If we would need, we could check this attribute by relevant validator — e.g., "required".

| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

### DIFF
--- a/framework/base/DynamicModel.php
+++ b/framework/base/DynamicModel.php
@@ -84,7 +84,7 @@ class DynamicModel extends Model
             return $this->_attributes[$name];
         }
 
-        return parent::__get($name);
+        return null;
     }
 
     /**


samdark:
Could be solved by defining defaults:

```php
 public function actionIndex()
{
    $rules = [
        [['id', 'email'], 'required'],
        ['id', 'integer'],
        ['name', 'string'],
        ['email', 'email'],
    ];

    $data = array_merge([
        'id' => null,
        'email' => null,
        'name' => null,
    ], \Yii::$app->request->bodyParams);

    $preModel = DynamicModel::validateData($data, $rules);
    // ...
}
```

I'd like to hear @yiisoft/core-developers opinions about if it makes sense to do it by default.

MuVO:
Yeah, this should help to solve issue, but this is workaround, not a solution…
I will wait for core-developers opinion too :)
tnx

klimov-paul:
I am against such change as it is unnecessary.
In case you wish to use `DynamicModel` for arbitrary data, you should define the list of its atributes explicitly:

```php
$model = new DynamicModel(compact('name', 'email'));
 $model->addRule(['name', 'email'], 'string', ['max' => 128])
     ->addRule('email', 'email')
     ->validate();
```

This is mentioned quite clear at the [docs](https://github.com/yiisoft/yii2/blob/master/framework/base/DynamicModel.php#L42-L47).

Automatic addition of the attributes based on validation rules is risk prone as rules definition may contains typos, e.g. `emial` instead of `email`.

dynasource:
Not validating `required` attributes that are missing in data seems quite logical to me. You should remove these rules if you want to allow these to be empty.

With respect to that fatal error. That should be fixed.

samdark:
@dynasource how should it behave instead of fatal-erroring?

dynasource:
@MuVO, where does the fatal error occur?

MuVO:
@dynasource an error occured after model created via constructor and tried `validate()`
Validator tries to access attributes which not actually created in model, then causes fatal error like «attribute does not exists in model»

Sorry, can't check *right now* if you need exactly error message, can quote it later

MuVO:
As I wrote above, I _think_ validating of non-existent attributes (e.g. by "required") should not thow a fatal error. It should populate `errors` in model…

framework/base/DynamicModel.php
```diff
@@ -84,7 +84,7 @@ public function __get($name)
             return $this->_attributes[$name];
         }
 
-        return parent::__get($name);
+        return null;
```
cebe: this will break getters and behaviors.

cebe (changes requested):

cebe:
Please open an issue describing the problem so we can discuss how a fix can look like. The current change in this PR can not be accepted like this.

PR #14761, opened 2017-09-05
Remove Yii autoloader

samdark:
When original Yii 2.0 autoloader was implemented there was no PSR-4 so it was a requirement to support aliases for extensions to load their classes. Also we wanted to have something like Yii 1.1 `Yii::import()` which proved to be quite useless when using autoloading.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | ?
| Fixed issues  | -


cebe:
huh? I'd still want to have those features...

samdark:
What for?

cebe:
- `@app` will not work without it
- I do not want `@tests` to be loadable in production, i.e. in a non-test app
- changing something requires composer autoloader to be rebuilt, which is a source of possible errors

samdark:
> @app will not work without it

Alias itself will work just fine. Namespace would require declaration in `composer.json` but that's acceptable since it's 2.1.

> I do not want @tests to be loadable in production, i.e. in a non-test app

Use `--no-dev` when deploying to production.

> changing something requires composer autoloader to be rebuilt, which is a source of possible errors

Only if you change namespace roots. Not often the case. Also autoloader is re-built on any composer operation.

rob006:
> I do not want `@tests` to be loadable in production, i.e. in a non-test app

Tests should be in `autoload-dev` section of `composer.json`: https://getcomposer.org/doc/04-schema.md#autoload-dev

There is also `exclude-from-classmap` option: https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps

docs/guide/concept-autoloading.md
```diff
@@ -2,93 +2,68 @@ Class Autoloading
...
+Since Yii uses Composer-generated autoloader, in order to configure it, `composer.json` should be used:
+
+
+```json
+{
+    "name": "myorg/myapp",
+    "autoload": {
+        "psr-4": {
+          "app\\": "src",
+          "tests\\": "tests"
```
dynasource: `autoload-dev` for `tests`, see https://github.com/yiisoft/yii2/pull/14761#issuecomment-327290560

dynasource (changes requested):
I am missing a clear example for people looking for the alternative for the original Yii::$classMap['namespace'] = pathToFile

dynasource (approved):
thx

docs/guide/helper-overview.md
```diff
@@ -67,12 +67,16 @@ class ArrayHelper extends BaseArrayHelper
...
+Next, in your application's [entry script](structure-entry-scripts.md), add the class to the class map
+to tell the [Composer class autoloader](concept-autoloading.md) to load your custom
 class instead of the original helper class from the framework:
 
 ```php
-Yii::$classMap['yii\helpers\ArrayHelper'] = '@app/components/ArrayHelper.php';
+/** @var \Composer\Autoload\ClassLoader $loader */
+$loader = require __DIR__ . '/vendor/autoload.php';
+$loader->addClassMap([
+    'yii\helpers\ArrayHelper' => 'components/ArrayHelper.php',
```
rob006: Path to file should be absolute.

I'm also against such approach, since it kills some composer autoloader optimizations. For PHP 5.6+ composer uses opcache to store static array with classmap: https://getcomposer.org/doc/articles/autoloader-optimization.md#what-does-it-do- . You will lose this if you modify classmap. Overrides should be defined in `composer.json` - then it will be included in composer classmap and no additional processing will be needed.

samdark: `composer.json` doesn't support per-class mappings.

rob006: Yes it does, in multiple ways, for example:

```json
    "autoload": {
        "psr-4": {
          "app\\": ""
        },
        "classmap": ["overrides/yii/helpers/Html.php"],
        "exclude-from-classmap": ["vendor/yiisoft/yii2/helpers/Html.php"]
    },
```

`exclude-from-classmap` section is required to prevent errors with `-o` option used in `dump-autoload`.

samdark: Oh... I'll check it.

samdark:
@rob thanks for the hint. Adjusted.

docs/guide/tutorial-yii-integration.md
```diff
@@ -62,12 +62,18 @@ In the worst case when the library requires explicitly including every class fil
...
+* List the classes and the corresponding file paths in the application `composer.json`:
+  ```json
+  "autoload": {
+      "psr-4": {
+          "app\\": ""
+      },
+      "classmap": [
+          "path/to/Class1.php",
+          "path/to/Class2.php",
+      ]
```
rob006: `classmap` supports also directories, so sometimes this could be simplified to 

```json
    "classmap": [
        "path/to/library/",
    ]
```

Composer will automatically find all classes in directory and include them into classmap.

rob006:
`tutorial-yii-integration.md` still have instruction how to use (removed) Yii autoloader.

docs/guide/concept-autoloading.md
```diff
@@ -2,93 +2,75 @@ Class Autoloading
...
+<?php
 
-[Aliases](concept-aliases.md) can be used to specify class file paths. You should set the class map in the
-[bootstrapping](runtime-bootstrapping.md) process so that the map is ready before your classes are used.
+// comment out the following two lines when deployed to production
+defined('YII_DEBUG') or define('YII_DEBUG', true);
+defined('YII_ENV') or define('YII_ENV', 'dev');
 
+require(__DIR__ . '/../vendor/autoload.php');
+require(__DIR__ . '/../vendor/yiisoft/yii2/Yii.php');
```
sergeymakinen:
```php
require(__DIR__ . '/../vendor/autoload.php');
```
should do the trick, no? Otherwise, a default `composer.json` should be able to load `Yii` as well.

docs/guide/tutorial-yii-integration.md
```diff
@@ -37,22 +37,26 @@ If a library is not released as a Composer package, you should follow its instal
 In most cases, you will need to download a release file manually and unpack it in the `BasePath/vendor` directory,
 where `BasePath` represents the [base path](structure-applications.md#basePath) of your application.
 
-If a library carries its own class autoloader, you may install it in the [entry script](structure-entry-scripts.md)
-of your application. It is recommended the installation is done before you include the `Yii.php` file so that
-the Yii class autoloader can take precedence in autoloading classes.
+If a library carries its own class autoloader, you may require it in the [entry script](structure-entry-scripts.md)
+of your application. It is recommended require statement is placed before Composer autoloader file so that
```
sergeymakinen:
```diff
-It is recommended require statement is placed
+It is recommended to put `require()` statements
```
It sounds less complex IMO.

sergeymakinen (changes requested):
Except for documentation notes, LGTM.

samdark:
@sergeymakinen updated.

sergeymakinen (approved):

framework/UPGRADE.md
```diff
@@ -105,6 +105,8 @@ Upgrade from Yii 2.0.x
   extend `yii\caching\SimpleCache` abstract class. Use `yii\caching\CacheInterface` only if you wish to replace `yii\caching\Cache`
   component providing your own solution for cache dependency handling.
 * Console command used to clear cache now calls related actions "clear" instead of "flush".
+* Yii autoloader was removed in favor of Composer-generated one.
```
SilverFire: Let's write explicitly what does user should do for this change. How about:

<code>
Yii autoloader was removed in favor of Composer-generated one. You should no longer write `require __DIR__ . '/../vendor/yiisoft/yii2/Yii.php';` in your application entrypoint. Removed method `Yii::autoload()` and property `Yii::$classMap`.
</code>

SilverFire (approved):
Good job. One minor comment

framework/UPGRADE.md
samdark: Done.

cebe:
pretty sure this breaks migrations: https://github.com/yiisoft/yii2/blob/5522d1909cf157395b49086145abed3ba9c96b88/framework/console/controllers/BaseMigrateController.php#L696

rob006:
We could use PSR-4 namespaces configuration from composer settings.

https://github.com/yiisoft/yii2/issues/14850#issuecomment-330792359

PR #14765, opened 2017-09-06
performance issue on rbac schema

ivan-redooc:
This commit solve a performance issue when a medium-tiny prermission/role hierarchy is used with a lot of users adding an index on `auth_assignment`.`user_id`.
See the queries in `yii\rbac\DbManager::getInheritedPermissionsByUser` and `yii\rbac\DbManager::getAssignments`.


samdark (changes requested):
https://github.com/yiisoft/yii2/blob/master/framework/rbac/migrations/m140506_102106_rbac_init.php should be adjusted as well.

Renkas:
I think that https://github.com/yiisoft/yii2/blob/master/framework/rbac/migrations/m140506_102106_rbac_init.php should not be changed.

There should be a new migration for the change.

samdark:
Yes. Correct.

dmirogin (approved):

rob006 (changes requested):
New migration is in wrong place - it should be in `rbac/migrations`.

samdark (approved):
Looks good to me now.

samdark:
Except what @rob006 mentioned, of course. Migration should be move.

rob006 (approved):
Changelog entry could be shorter, "add index on `auth_assignment.user_id` for performance reasons" would be just enough.

framework/rbac/migrations/m170907_052038_rbac_add_index_on_auth_assignment_user_id.php
```diff
@@ -10,7 +10,7 @@
 class m170907_052038_rbac_add_index_on_auth_assignment_user_id extends Migration
 {
     public $column = 'user_id';
-    public $index = 'idx-auth_assignment-user_id';
```
samdark: I'd rather use `idx-auth_assignment-user_id` because `idx_user_id` is too general and naming conflicts are likely to occur.

rob006: Then index name should be adjusted also in `.sql` files.

samdark: Yes.

rob006 (approved):

samdark:
Thanks for working on it. Merged.

PR #14772, opened 2017-09-06
Fix #14770: Correct `Html::addCssClass` array to string conversion bug

### LINKED ISSUE #14770: Avoid `Html::addCssClass` array to string conversion error
When `options` is an empty array and a set of CSS classes are added as an array - the second condition in [Line 1907](https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1907) is executed which wrongly in this case assigns an array instead of converting it to a string.

CSS class setting must always be a string.

### What do you get instead?
An array instead of a string.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13 (dev)
| PHP version      | Any 
| Operating system | Any

Will submit a PR separately to correct this.

kartik-v:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14770

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.13 under development
 ------------------------
 
+- Bug #14770: Avoid `Html::addCssClass` array to string conversion error (kartik-v)
 - Enh #14273: `yii\log\Target::$enabled` now supports callable value (dmirogin)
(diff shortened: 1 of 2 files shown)


kartik-v:
Yes there is a mistake in the PR... it should be:
`
``php
$options['class'] = is_array($class) ? implode(' ', $class) : $class;
```

anyway the issue is different and have created #14773.

PR #14783, opened 2017-09-08
Update Controller.php

manyopensource:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  |


Renkas:
But why?

It only breaks the framework ...

manyopensource:
Answer: It is very logical: SomeController, someAction

manyopensource:
And Phalcon use that...

dmirogin:
I think, do something, cause it works another way in other frameworks is a bad way. Moreover there is no any benefit. Not to mention the fact that it breaks all solutions based on yii2.

PR #14784, opened 2017-09-08
Change `RequestParserInterface::parse()` to accept whole `Request` instance

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/framework/web/JsonParser.php
+++ b/framework/web/JsonParser.php
@@ -39,16 +39,13 @@ class JsonParser implements RequestParserInterface
 
 
     /**
-     * Parses a HTTP request body.
-     * @param string $rawBody the raw HTTP request body.
-     * @param string $contentType the content type specified for the request body.
-     * @return array parameters parsed from the request body
+     * {@inheritdoc}
      * @throws BadRequestHttpException if the body contains invalid json and [[throwException]] is `true`.
      */
-    public function parse($rawBody, $contentType)
+    public function parse($request)
     {
         try {
-            $parameters = Json::decode($rawBody, $this->asArray);
+            $parameters = Json::decode($request->getBody()->__toString(), $this->asArray);
             return $parameters === null ? [] : $parameters;
         } catch (InvalidArgumentException $e) {
             if ($this->throwException) {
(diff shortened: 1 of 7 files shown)


framework/web/RequestParserInterface.php
```diff
@@ -11,15 +11,18 @@
...
-     * @return array parameters parsed from the request body
+     * Parses a given request instance determining its body parameters.
+     * This method MUST return the array of body parameters detected from [[$request]] data.
+     * However, this method MAY adjust the given [[Request]] instance directly for extra configuration.
+     * @param Request $request the HTTP request instance to be parsed.
+     * @return array parameters parsed from the request body.
+     * @throws BadRequestHttpException in case request body format is invalid.
      */
-    public function parse($rawBody, $contentType);
+    public function parse($request);
```
dmirogin: Maybe worth adding Request class type hint ?

framework/web/JsonParser.php
```diff
@@ -39,16 +39,13 @@ class JsonParser implements RequestParserInterface
...
-     * @return array parameters parsed from the request body
+     * {@inheritdoc}
      * @throws BadRequestHttpException if the body contains invalid json and [[throwException]] is `true`.
      */
-    public function parse($rawBody, $contentType)
+    public function parse($request)
     {
         try {
-            $parameters = Json::decode($rawBody, $this->asArray);
+            $parameters = Json::decode($request->getBody()->__toString(), $this->asArray);
```
dmirogin: > $request->getBody()->__toString()

Request has method getRawBody()

framework/web/RequestParserInterface.php
klimov-paul: No, it will make it inconsistent with `ResponseFormatterInterface`:
https://github.com/yiisoft/yii2/blob/master/framework/web/ResponseFormatterInterface.php#L22

framework/web/JsonParser.php
klimov-paul: This method usage should be avoided.
Since PSR-7 accepted its interface should be used as wide as possible.

samdark (approved):

dmirogin (approved):

PR #14797, opened 2017-09-11
Change access type of method

thiagotalma:
BaseMigrateController::generateClassName() as protected.
If there is no impediment, I suggest changing the method of visibility to allow us to overwrite it.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -641,7 +641,7 @@ abstract class BaseMigrateController extends Controller
      * @return array list of 2 elements: 'namespace' and 'class base name'
      * @since 2.0.10
      */
-    private function generateClassName($name)
+    protected function generateClassName($name)
     {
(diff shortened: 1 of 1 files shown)


thiagotalma:
Use case: I want to be able to set the prefix of the migrations name.

cebe:
changing the pattern of migration name also affects a lot of other places in migration controller so just making the method protected does not really help.

thiagotalma:
I have no doubt that changing the name of a migration already applied can cause problems.

But I see no reason to prevent the developer from choosing their template names to create a migration file.

samdark:
If it's to make customizable we need to make sure it's customizable everywhere related.

samdark:
Closing since PR doesn't make sense in its current form: not enough for the purpose.

PR #14802, opened 2017-09-12
Update BaseHtml.php

wcoc:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1506,7 +1506,11 @@ class BaseHtml
             unset($options['label']);
         }
 
-        $checked = "$value" === "{$options['value']}";
+        if (is_array($value)) {
+            $checked = ArrayHelper::isIn($options['value'], $value);
+        } else {
+            $checked = "$value" === "{$options['value']}";
+        }
 
         if (!array_key_exists('id', $options)) {
             $options['id'] = static::getInputId($model, $attribute);


wcoc:
Usage example:

View
```
<table class="table table-bordered table-condensed">
        <thead>
               <tr>
                     <th>ID Item</th>
                     <th>Description</th>
        </thead>

        <tbody>
         <?php foreach($items as $item): ?>
            <tr>
                <td><?= Html::activeCheckbox($modelForm, 'array_id_item', [
                                        'label' => false, 
                                        'value' => $item->id_item,
                       ]) ?></td>
                <td>Description....</td>
           </tr>
         <?php endforeach; ?>
        </tbody>
</table>
```

Modelform
```
class ItemModelForm extends Model {
        /* this attribute is used to save all ids selecteds in view array checkbox. */
        public $array_id_item;

        function rules(){
               return [['array_id_item'], 'safe'];
        }
}
```

dmirogin:
try to use https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1665

wcoc:
@dmirogin this not working for me, because i need show several informations in columns for each line.

Example: [https://imgur.com/KoTxSxs](https://imgur.com/KoTxSxs)

rob006:
This will result multiple checkboxes with the same name - it will not work correctly, since on form submit last checkbox will replace all previous checkboxes.

wcoc:
@rob006 Actually, I just let this happen.

I solved my problem by passing in the option of activeCheckbox `'uncheck' => false`  so that it does not send unselected values, and attribute `'array_id_item['.$item->id_item.']'`.

Thanks.

PR #14814, opened 2017-09-14
Clarify how joinWith()/innerJoinWith() deals with eager loading

mikehaertl:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Intuitively I'd have expected, that the data for the related records would come from the same JOIN query.  

More than once now I was hit by problems caused by this misunderstanding. So I'd say it's worth an extra note in the docs. A note like this would have saved me quite some time.


samdark:
Merged. Thanks!

Issue #14821, opened 2017-09-14
Merging container definitions with any class configs


3dmaxpayne:
When I am trying to setup default configs for some class I cannot do it for double nesting configs. 
example:
```
'container' => [
        'definitions' => [
            zxbodya\yii2\tinymce\TinyMce::class => [
                'settings' => [
                    'height' => '500',
                ],
                'someConfig' => 'value1',
            ],
        ],
    ],
```

value for **someConfig** will apply successfully, but value for **settings** will not apply.
it because you use array_merge instead of array_merge_recursive [here ](https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php#L168)

So, my propose is replace this function to array_merge_recursive 

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.0.22
| Operating system | Ubuntu 16.04

rob006:
Using `array_merge_recursive()` will give your wrong results when you try to override some values: https://3v4l.org/Hg6mD . 
People may also rely on this behavior, so using `ArrayHelper::merge()` will break BC.

klimov-paul:
Duplicates #4154

rob006:
This can be fixed in 2.1, since after #11899 we have a way to fully replace array value.

SilverFire:
@rob006 so would you like to close #14821 or point it to 2.1?

rob006:
@SilverFire You mean #14822? I'm fine with closing it, since it does not contain any valuable changes. Unless @3dmaxpayne wants to rewrite it with `ArrayHelper::merge()`, docs, tests and upgrade notes.

3dmaxpayne:
yes, I still want. for me it is actual, but I didn't get what you guys think about it?

samdark:
Closing since new DI container doesn't seem to have it anymore.

DrDeath72:
2.0.x still need this fix

samdark:
If you know a way to fix it without any backwards compatibility breaks, you're very welcome to suggest it.

DrDeath72:
maybe add some key to definition array that enable ArrayHelper::merge feature, if key not exists use old array_merge by default?

samdark:
That could work. Want to try implementing it? If the change would not be big I think it's possible to merge it.

DrDeath72:
https://github.com/yiisoft/yii2/pull/17436

PR #14822, opened 2017-09-14
#14821 Merging container definitions with any class configs

3dmaxpayne:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #14821

### DIFF
--- a/framework/di/Container.php
+++ b/framework/di/Container.php
@@ -165,7 +165,7 @@ class Container extends Component
             $concrete = $definition['class'];
             unset($definition['class']);
 
-            $config = array_merge($definition, $config);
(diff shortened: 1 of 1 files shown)


SilverFire (changes requested):
- Use `ArrayHelper::merge()` instead `array_merge_recursive`
 - Add a CHANGELOG note
 - Reflect changes in [tests](https://github.com/yiisoft/yii2/blob/master/tests/framework/di/ContainerTest.php)

samdark:
Closing since it's not backwards compatible thus can't be applied to 2.0.

Issue #14827, opened 2017-09-15
No shortcut for createAbsoluteUrl()?


bscheshirwork:
So, `\yii\helpers\Url` not have any way to render full link?
Is it done on purpose?

### What is the expected result?
```php
\yii\helpers\Url::absoluteUrl(['/user/view', 'id' => $id]);
```
### What do you get instead?
```php
\Yii::$app->urlManager->createAbsoluteUrl(['/user/view', 'id' => $id]);
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13@dev

dungphanxuan:
https://github.com/yiisoft/yii2/blob/master/framework/web/UrlManager.php#L503
about?

klimov-paul:
https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseUrl.php#L204-L205

bscheshirwork:
Thank you! 
i.e. 
```
Url::toRoute(['/user/view', 'id' => $id], '');
```
some similar function as a
```
\Yii::$app->urlManager->createAbsoluteUrl(['/user/view', 'id' => $id]);
```

But I can see error into PHPDoc now:
In this doc accept pass `null`
https://github.com/yiisoft/yii2/blob/a86c67ddb3fe32c4269b5ed90f8837bc3daf0cff/framework/web/UrlManager.php#L497-L499

In this doc
https://github.com/yiisoft/yii2/blob/a86c67ddb3fe32c4269b5ed90f8837bc3daf0cff/framework/helpers/BaseUrl.php#L204-L206

`null` not acceptable, but here:
https://github.com/yiisoft/yii2/blob/a86c67ddb3fe32c4269b5ed90f8837bc3daf0cff/framework/helpers/BaseUrl.php#L101-L103
any value is `!== true && !is_string` rise send `null` 

i.e. in `toRoute`
```
     * - `false` (default): generating a relative URL.
     * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::$hostInfo]].
     * - string: generating an absolute URL with the specified scheme (either `http`, `https` or empty string
     *   for protocol-relative URL).
     * - `null`:  generating an absolute URL with the current request scheme
```

PR #14828, opened 2017-09-15
add null to valid value of scheme

bscheshirwork:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Fixed issues  | https://github.com/yiisoft/yii2/issues/14827#issuecomment-329726602

### DIFF
--- a/framework/helpers/BaseUrl.php
+++ b/framework/helpers/BaseUrl.php
@@ -83,12 +83,13 @@ class BaseUrl
      *
      * @param string|array $route use a string to represent a route (e.g. `index`, `site/index`),
      * or an array to represent a route with query parameters (e.g. `['site/index', 'param1' => 'value1']`).
-     * @param bool|string $scheme the URI scheme to use in the generated URL:
+     * @param null|bool|string $scheme the URI scheme to use in the generated URL:
      *
      * - `false` (default): generating a relative URL.
      * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::$hostInfo]].
      * - string: generating an absolute URL with the specified scheme (either `http`, `https` or empty string
      *   for protocol-relative URL).
+     * - `null`: generating an absolute URL with the current request scheme.
      *
      * @return string the generated URL
      * @throws InvalidParamException a relative route is given while there is no active controller
--- a/framework/helpers/BaseUrl.php
+++ b/framework/helpers/BaseUrl.php
@@ -86,10 +86,9 @@ class BaseUrl
      * @param null|bool|string $scheme the URI scheme to use in the generated URL:
      *
      * - `false` (default): generating a relative URL.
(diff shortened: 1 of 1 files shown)


framework/helpers/BaseUrl.php
```diff
@@ -83,12 +83,13 @@ class BaseUrl
...
      * @param string|array $route use a string to represent a route (e.g. `index`, `site/index`),
      * or an array to represent a route with query parameters (e.g. `['site/index', 'param1' => 'value1']`).
-     * @param bool|string $scheme the URI scheme to use in the generated URL:
+     * @param null|bool|string $scheme the URI scheme to use in the generated URL:
      *
      * - `false` (default): generating a relative URL.
      * - `true`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::$hostInfo]].
      * - string: generating an absolute URL with the specified scheme (either `http`, `https` or empty string
      *   for protocol-relative URL).
+     * - `null`: generating an absolute URL with the current request scheme.
```
rob006: How this differ from `returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::$hostInfo]].`?

bscheshirwork: @rob006 can you give an example please?

rob006: AFAIK `null` works the same as `true`, so what is the point of documenting this? If you pass an object or array it will also work as `true`, but there is no sense to document every possible variation for this param.

bscheshirwork: So... Of course, you right.

But `true` antagonist of `null` => we have flip logic. Some strange.
I expect to use same value in each way to create full link:

```php
foreach (['http', 'https', '', null] as $scheme){
  \yii\helpers\Url::toRoute(['/user/view', 'id' => $id], $scheme);
  \Yii::$app->urlManager->createAbsoluteUrl(['/user/view', 'id' => $id], $scheme);
}
```

bscheshirwork: additional, avoid of doc in createAbsoluteUrl is same with true
```php
foreach (['http', 'https', '', null, true] as $scheme){
  \yii\helpers\Url::toRoute(['/user/view', 'id' => $id], $scheme);
  \Yii::$app->urlManager->createAbsoluteUrl(['/user/view', 'id' => $id], $scheme);
}
```

bscheshirwork: Can we change it like this:
```php
     * - `true` | `null`: returning an absolute base URL whose scheme is the same as that in [[\yii\web\UrlManager::$hostInfo]].
```

rob006: >  I expect to use same value in each way to create full link:

Why? These are different components, they works differently, why do you expect similar behavior? Documenting `null` as possible value will only complicate API of this method.

bscheshirwork: Just because I expected easy for understand clearly design.

Anyway, current realization is correct.

PR #14830, opened 2017-09-15
Fix #7551 Path Info Array Format

### LINKED ISSUE #7551: Pathinfo should not be decoded
In `yii\web\Request::resolvePathInfo()`, pathinfo has been decoded using `urldecode()`, which is similar in yii 1.x.

That would be an issue if the url path contains %2f.
For example a url path like `/foo/GNU%2fLinux/bar`, the pathinfo would become `/foo/GNU/Linux/bar`, which used for matching url rule. 
This may cause some problem when routing.

The code for resolving PathInfo is coming from Zend Framework, but the urldecode behavior has been added later, described in https://code.google.com/p/yii/issues/detail?id=1309
Zend Framework also fixed the similar issue in http://framework.zend.com/issues/browse/ZF-11017

Here are some test cases in Zend Framework for reference:
https://github.com/zendframework/zf1/blob/master/tests/Zend/Controller/Request/HttpTest.php#L988

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #7551


rob006:
Why not leave `$pathInfo` as string (but encoded) and implement only `UrlRule::$parseUrlencoded` part?

klimov-paul:
`Request::getPathInfo()` does more then just url-decode: it attempts to convert it into 'utf-8'.
See http://w3.org/International/questions/qa-forms-utf-8.html

Being a url-decoded array it can be used for custom URL rules in simple way:

```php
public function parseRequest($manager, $request)
{
     $pathInfo = $request->getPathInfo();
     if (count($pathInfo) == 2 && $pathInfo[0] === 'page') {
         return [$this->route, 'page' => $pathInfo[1]];
     }
     return false;
}
```

klimov-paul:
Note that using url-encoded path for the route parsing will make particular `UrlRule` pattern to be more complex as you'll have match `%XX` for single character.
This fix allows usage of the old patterns with ability to bypass problem of url-encoded slashes, see [unti test for value with slash](https://github.com/yiisoft/yii2/pull/14830/files#diff-96fb0ef29c09f20871090c41bf5b3839R1226).

rob006:
> Request::getPathInfo() does more then just url-decode

We don't need to change this. In fact we could leave current behavior almost unchanged - `$pathInfo` as string which stores decoded path, except `/` which are still encoded. 
`/my%2fpath%26some+spaces/` will be represented as `/my%2fpath&some spaces/`.

This should fix #7551 but it is more intuitive and should be fully BC.

klimov-paul:
> in fact we could leave current behavior almost unchanged - $pathInfo as string which stores decoded path, except / which are still encoded.

This would not work as particular path fragment may contain `%2f`, which is match urlencoded '/'. Thus having `%2f` inside the path you wil be unable to determine whether it just `%2f` or encoded '/'.

samdark (approved):

samdark:
Won't make it into 2.0.

Issue #14829, opened 2017-09-15
GridView columns title depends of $provider->query->modelClass only, not use both / only $this->grid->filterModel in common case of usage


bscheshirwork:
Order of applied `ifelse` is not ideal for default-like case

https://github.com/yiisoft/yii2/blob/a86c67ddb3fe32c4269b5ed90f8837bc3daf0cff/framework/grid/DataColumn.php#L138-L173

### What steps will reproduce the problem?
1. Gii-generated CRUD: Search model + GridView::widget
2. Expand Search model with addition field - for example `name` of relation model.
order rule + search rule + join in query
3. Expand Search model attributeLabels
```php
    /**
     * @inheritDoc
     */
    public function attributeLabels()
    {
        return ArrayHelper::merge(parent::attributeLabels(), [
            'relationName' => Yii::t('relation', 'Name'),
        ]);
    }
```
4. Use array-definition for column with `attribute`
```php
        'columns' => [
            [
                'format' => 'raw',
                'attribute' => 'relationName',
                'content' => \backend\views\ViewHelper::relatedName(),//function with view
            ],
        ],
```
### What is the expected result?
Table header fill from `searchModel` `attributeLabels()`

### What do you get instead?
Table header fill from `dataProvider`s model  `attributeLabels()`
case of line 147
```php
if ($provider instanceof ActiveDataProvider && $provider->query instanceof ActiveQueryInterface)...
```

and I must put unexpected additional `label`
```php
        'columns' => [
            [
                'format' => 'raw',
                'attribute' => 'relationName',
                'label' => $searchModel->getAttributeLabel('relationName'),
                'content' => \backend\views\ViewHelper::relatedName(),//function with view
            ],
        ],
```
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13@dev

klimov-paul:
Duplicates #4504

bscheshirwork:
@klimov-paul please, see changes in PR. 


I read origin thread... So... Same situation year by year. 


may be this is mark for review the opinion?

bscheshirwork:
Data in grid not for only one model, can be from many case, generated in cascade view, etc
Search model is representation of header. Same for header. (In default case)

PR #14831, opened 2017-09-15
Reorder pick attribute label

bscheshirwork:
Add additional check for filter model before get label from dataProvider's model

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes: same named attribute generate label from the filter model, not from the base model
| Tests pass?   | yes/no
| Fixed issues  | #14829

### DIFF
--- a/framework/grid/DataColumn.php
+++ b/framework/grid/DataColumn.php
@@ -12,6 +12,7 @@ use yii\base\Model;
 use yii\data\ActiveDataProvider;
 use yii\data\ArrayDataProvider;
 use yii\db\ActiveQueryInterface;
+use yii\db\ActiveRecord;
 use yii\helpers\ArrayHelper;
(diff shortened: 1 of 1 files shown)


klimov-paul:
Duplicates #4504

bscheshirwork:
@klimov-paul, please see different:

get from `filterModel` only if `filterModel` is a child model 
```php
            /* @var $modelClass Model */
            $modelClass = $provider->query->modelClass;
            if ($this->grid->filterModel !== null && $this->grid->filterModel instanceof $modelClass) {
                return $this->grid->filterModel->getAttributeLabel($this->attribute);
            }
            /* @var $model ActiveRecord */
            $model = $modelClass::instance();

            return $model->getAttributeLabel($this->attribute);

```

PR #14819, opened 2017-09-14
Support for PgSQL array conditions

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7640

Issue #7640, opened 2015-03-09
Ability to work with postgres arrays.


joni-jones:
Postgres can store arrays in columns. I think it would be cool if yii will be support postgres array operators, like `@>` in `where()` method.

cebe:
can you give a link to the documentation that describes these?

joni-jones:
http://www.postgresql.org/docs/9.3/static/functions-array.html. Now I have few implementations of `@>` operator for `integer` and `text` arrays, but I need time to write tests.

cebe:
you can already use them: `where('<@', 'col1', 'col2')`

joni-jones:
Sorry, but `where()` receive only one parameter, i can create this query similar to this:
```php
->where('id @> ARRAY[2,3,4]')
```
But, in this case need to manually escape column and values and also implode array elements.

cebe:
oh, sorry. should be `where(['<@', 'col1', 'col2'])`

nineinchnick:
Possibly related to #7481

joni-jones:
@cebe, your example not working, if executte this code:
```php
where(['<@', 'id', 'ARRAY[2,3]'])
```
we will get database error: `	exception 'PDOException' with message 'SQLSTATE[22P02]: Invalid text representation: 7 ERROR: array value must start with "{" or dimension information' in /var/www/yii2/framework/db/Command.php:822`

Because expression `ARRAY[2,3]` will be quoted to `"ARRAY[2,3]"` and result query looks like this:
```sql
SELECT COUNT(*) FROM "events" WHERE "id" <@ 'ARRAY[2,3]'
```

cebe:
you have to use a `yii\db\Expression` object to wrap the argument

joni-jones:
Yes, but values escaping not working as expected:
```php
$expression = new \yii\db\Expression('ARRAY[:id]', [':id' => [2, 3]]);
...->where(['<@', 'id', $expression]);
```
Generate query similar to:
```sql
SELECT COUNT(*) FROM "events" WHERE "id" <@ ARRAY[2]
```

cebe:
`->where(['<@', 'id', new \yii\db\Expression('ARRAY[2,3]')])`

joni-jones:
But in this case, also needed manual escaping for values.

joni-jones:
This PR allow to use filter similar to `in` operator. Few examples:
```php
where(['in integer array', 'id', [2, 3]])
```
[...]

Issue #8027, opened 2015-04-09
accept sub-queries in where clause with operator format


twinni:
To use a sub-query in a where clause you can use only the hash format which creates always
an 'in' operator.
To create an **=**  (or any other) operator the operand2 of the operator format should also accept sub-queries.

Example:
```php
$subQuery = (new \yii\db\Query())
                ->select('max(ablesedatum)')
                ->from(self::tableName())
                ->where(['zaehlerID' => $this->zaehlerID])
                ->andWhere(['<', 'ablesedatum', $this->ablesedatum]);

// this creates ... ablesedatum in .... 
$model = Zaehlerstand::find()
            ->where ([
                'zaehlerID' => $this->zaehlerID,
                'ablesedatum' => $subQuery])
            ->one();

// this should be possible to create ... ablesedatum = ...
  $model1 = Zaehlerstand::find()
            ->where (['zaehlerID' => $this->zaehlerID])
            ->andWhere(['=','ablesedatum', $subQuery])
            ->one();
```

PR #14836, opened 2017-09-16
[Alternative] Support for PgSQL array conditions

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7640

This is the alternative implementation of PR #14819


hiqsol:
This one looks better then https://github.com/yiisoft/yii2/pull/14819

framework/db/pgsql/ArrayType.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace yii\db\pgsql;
+
+use yii\db\ExpressionInterface;
+use yii\db\Query;
+use yii\db\QueryBuilder;
+
+class ArrayType implements ExpressionInterface
```
rob006: I would name it `ArrayExpression`.

tests/framework/db/DatabaseTestCase.php
```diff
@@ -120,7 +120,7 @@ protected function replaceQuotes($sql)
             case 'cubrid':
             case 'pgsql':
             case 'oci':
-                return str_replace(['[[', ']]'], '"', $sql);
+                return str_replace(['\\[', '\\]'], ['[', ']'], preg_replace('/(\[\[)|((?<!(\[))\]\])/', '"', $sql));
```
rob006: Why changes for pgsql affects cubrid and oci?

framework/db/ExpressionInterface.php
```diff
@@ -0,0 +1,27 @@
...
+interface ExpressionInterface
+{
+    /**
+     * Method builds the raw SQL expression
+     *
+     * @param QueryBuilder $queryBuilder the QueryBuilder that intends to use this expression
+     * @param array $params the binding parameters
+     * @return string the raw SQL that will not be additionally escaped or quoted
+     */
+    public function buildUsing(QueryBuilder $queryBuilder, &$params = []);
```
sergeymakinen: I would call it `build()` since QB here is a part of the implementation and `using` in `yii\db` usually means it's somehow related to an `using` keyword/expression.

framework/db/pgsql/ArrayType.php
```diff
@@ -0,0 +1,97 @@
...
+
+        return $result;
+    }
+
+    /**
+     * @param QueryBuilder $queryBuilder
+     * @param array $params
+     * @return mixed
+     */
+    public function buildUsing(QueryBuilder $queryBuilder, &$params = [])
```
sergeymakinen: `@inheritdoc` should be sufficient here.

framework/db/pgsql/ArrayType.php
```diff
@@ -0,0 +1,97 @@
...
+     * @param $value
+     * @param string|null $type
+     */
+    public function __construct($value, $type = null)
+    {
+        $this->value = $value;
+        $this->type = $type;
+    }
+
+    protected function getTypecast()
```
sergeymakinen: Missing PHPDoc.

framework/db/pgsql/ArrayType.php
```diff
@@ -0,0 +1,97 @@
...
+
+use yii\db\ExpressionInterface;
+use yii\db\Query;
+use yii\db\QueryBuilder;
+
+class ArrayType implements ExpressionInterface
+{
+    const PARAM_PREFIX = ':qp';
+
+    protected $type;
```
sergeymakinen: Missing PHPDoc.

framework/db/pgsql/ArrayType.php
```diff
@@ -0,0 +1,97 @@
...
+
+    protected $value;
```
sergeymakinen: Missing PHPDoc.

framework/db/pgsql/ArrayType.php
sergeymakinen: Also missing PHPDoc.

framework/db/pgsql/ArrayType.php
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        if (empty($placeholders)) {
+            return "'{}'";
+        }
+
+        return 'ARRAY[' . implode(', ', $placeholders) . ']' . $this->getTypecast();
+    }
+
+    protected function buildSubqueryArray($sql)
```
sergeymakinen: Missing PHPDoc.

sergeymakinen (changes requested):
It looks solid to me and I like introducing `ExpressionInterface`. Some changes are welcome though. :)

framework/db/ExpressionInterface.php
SilverFire: When you see `$expression->build($queryBuilder)`, what is the first meaning you guess?
For me it is something inconsistent like "Build query builder in expression?? Wtf?"
And compare it with `$expression->buildUsing($queryBuilder)`. For me that `Using` improves understanding a lot.

sergeymakinen: Fair enough.

cebe (commented):
Not really happy with introducing the Expression interface here, other than that I like the approach of using Expression for this use case more.

Can you give some usage examples, for the case where the array value needs to be a Query object?

SilverFire:
> Can you give some usage examples, for the case where the array value needs to be a Query object?

For example, it makes possible to show users that belong to disabled roles:

```sql
SELECT      *
FROM        user
WHERE       roles && array(
    SELECT  name
    FROM    role
    WHERE   active = false 
);
```

Generally, the use case is the same as for scalars – see #8027.

Why don't you like interface?

hiqsol:
I like the `ExpressionInterface` idea and I think it can be used even wider to make code shorter and cleaner!
For example `ExpressionInterface` can be added to `Query` and it will simplify code like this:

```php
    if ($value instanceof ExpressionInterface) {
	$placeholders[] = $value->buildUsing($this, $params);
    } elseif ($value instanceof \yii\db\Query) {
	list($sql, $params) = $this->build($value, $params);
	$placeholders[] = "($sql)";
    } 
```

Into this:

```php
    if ($value instanceof ExpressionInterface) {
	$placeholders[] = $value->buildUsing($this, $params);
    } 
```

samdark:
Looks cleaner than #14819.

SilverFire:
Any objections? Otherwise will add PHPDocs and merge today evening

cebe:
I'd like to have a second look. Trying to do that today.

SilverFire:
Ok, will wait for you review

dmirogin (approved):

framework/db/ExpressionInterface.php
cebe: Still not happy with the naming, it shoulds like we are building a "using" - whatever that should mean. How about `buildSql()`?

cebe:
suggested some improvements in https://github.com/SilverFire/yii2/pull/1

TODO:

- [ ] find a better name for `ExpressionInterface::buildUsing()`, it is not building a "using" but either an "expression" or some "sql" so I'd rename it to `buildExpression` or `buildSql`.
- [ ] write docs for the guide. This feature should be documented.

tests/framework/db/DatabaseTestCase.php
cebe: fixed.

cebe:
Ideally a few tests should be added that run queries against postgres to see syntax is valid in these cases.

framework/db/ExpressionInterface.php
klimov-paul: First of all I would not pass `QueryBuilder` to this method. In order to be consistent with other DB-relarted code parts, it should accept `yii\db\Connection` instance:

```php
public function build(Connection $db) {...}
```

But the whole interface is inconsistent. If you wish to exrtact an interface for `yii\db\Expression` it should have 2 methods: `getSql()` and `getParams()` instead being built by `QueryBuilder`.
Also DB Expression should have `__toString()` method, allowing it to be typecasted into string for debug purposes. So interface should look like:

```php
interface ExpressionInterface
{
    public function getSql() : string;

    public function getParams() : array;

    public function __toString() : string;
}
```

If expression should be bound to particular DB connection or query builder (while I am unsure why) it shoud be done via constructor parameter or setter.

klimov-paul (changes requested):

framework/db/pgsql/ArrayExpression.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    public function buildUsing(QueryBuilder $queryBuilder, &$params = [])
+    {
+        $value = $this->values;
+
+        if ($value instanceof Query) {
+            list ($sql, $params) = $queryBuilder->build($value, $params);
```
klimov-paul: It is better to introduce check for `instanceof Query` somewhere inside `QueryBuilder`.
Having `Query` instance as one of the argument affects ay condtion composition, for example `IN` SQL operator.

klimov-paul (changes requested):

framework/db/pgsql/ArrayExpression.php
SilverFire: Than it is impossible to build condition like this:

```
column && array[(SELECT id FROM table), (SELECT id FROM another_table)]
```

framework/db/ExpressionInterface.php
SilverFire: > In order to be consistent with other DB-relarted code parts, it should accept `yii\db\Connection` instance

Expression is related to Query, that is about to build. Not to the connection itself.

> If expression should be bound to particular DB connection or query builder (while I am unsure why) it shoud be done via constructor parameter or setter.

It is not bound, but it can be built within `QueryBuilder`, that belongs to a particular DB connection.
You do not specify neither DB nor `QueryBuilder` when you create a Query, why should we do it for expression. Providing DB connection or `QueryBuilder` via constructor will make the API horrible.

> If you wish to exrtact an interface for yii\db\Expression it should have 2 methods: `getSql()` and `getParams()` instead being built by `QueryBuilder`

Providing **static** SQL and params is only one of the use cases. More general one is providing **some** SQL and params, that shall be used in the concrete query.

> Also DB Expression should have `__toString()` method, allowing it to be typecasted into string for debug purposes. So interface should look like:

I can agree on necessity of Expression dumping for debug purposes, but I am not sure `__toString()` is the best option. Will think more about it.

framework/db/pgsql/ArrayExpression.php
klimov-paul: I would solve this using some kind of placeholder mechanism:

```php
interface ExpressionInterface
{
    public function getSubQueries();
}

class ArrayExpression implements ExpressionInterface
{
   public function getSql()
   {
       return column && array[({[q1]}), ({[q1]})];
   {

    public function getSubQueries()
    {
        return [
            '{[q1]}' => new Query()->setSql('SELECT id FROM table'),
            '{[q2]}' => new Query()->setSql('SELECT id FROM another_table'),
        ];
    }
}
```

klimov-paul: Or there should setter and getter for `yii\db\Connection` instance inside expression, so by default it will pick `Yii::$app->db`, get query builder from it and solve the issue.

framework/db/ExpressionInterface.php
SilverFire: > Still not happy with the naming, it shoulds like we are building a "using" - whatever that should mean. How about `buildSql()`?

It was important for me to indicate causal relationship in this naming: 

I want to build this expression &rarr; I need QueryBuilder and params to do it &rarr; I am building the Expression using QueryBuilder and params &rarr; `$expression->buildUsing($queryBuiler, $params)`

Could you compose something similar with `buildSql()`?

klimov-paul: > Expression is related to Query, that is about to build. Not to the connection itself.

`Query` accepts `Connection` instance as an argument for final execution methods, e.g. methods `one()`, `all()` and so on accepts `$db` as a connection instance. `QueryBuilder` is just a part of the `Connection`, and actually it is unnecessary part. Particular developer may re-implement Db layer in the way `QueryBuilder` will not be in use.
`QueryBuilder` can always be determined from `Connection` thus using `Connection` is more flexible here.

> You do not specify neither DB nor QueryBuilder when you create a Query, why should we do it for expression. Providing DB connection or QueryBuilder via constructor will make the API horrible.

Expression itself should not depend on Connection or be somehow bound ot it anywhere.
At the present `yii\db\Expression` is used to specify some raw SQL, which should bypass the QueryBuilder. If there is some construction, which should be build depending on particular DB schema - it should be incapsulated inside `QueryBuilder` - not in `Expression`.

> Providing static SQL and params is only one of the use cases. More general one is providing some SQL and params, that shall be used in the concrete query.

I do not understand this part. SQL Expression consists of 2 parts: raw SQL text and binding parameters - nothing else. Thus `ExpressionInterface` should return its raw SQL and binding params related to it, so it should have 2 methods.
It is a taks of `QueryBuilder` to place given SQL text on its place in the one of entire Query and merge binding params. It was always doing this before this PR. Why it should be changed?

> I can agree on necessity of Expression dumping for debug purposes, but I am not sure __toString() is the best option. Will think more about it.

PSR standards using `__toString()` in thier interfaces all the time. Since we accepted several PSR for 2.1, we should pick this approach as well.

framework/db/pgsql/ArrayExpression.php
cebe: > Or there should setter and getter for yii\db\Connection instance inside expression, so by default it will pick Yii::$app->db, get query builder from it and solve the issue.

this does not make sense, an expression is always part of some sql query, which should use the connection and query builder in that context and not try to fallback to global db connection.

klimov-paul: The problem is this PR twist the original idea of `yii\db\Expression`. Expression was created to bypass parameters binding, allowing to paste SQL statements instead of string param at particular place.
This PR considers `Expression` is a light-weight version of `yii\db\Query`, which should be built by `QueryBuilder` inside its owner `Query` object.
I would say that `yii\db\Query` should extend `yii\db\Expression` creating a 'Composite' pattern, allowing building of nested queries with unlimited nesting level.

I would rather allow array condition to be specified via `yii\db\Query` instance (or its descendant) rather then `Expression`.

klimov-paul: Instead of introducing `build()` method inside `Expression` and gess a good name for it - it is better to adjsut `QueryBuilder::build()` in the way it can process both `Query` and `Expression` and use it as universal interface.

framework/db/ExpressionInterface.php
SilverFire: > `Query` accepts `Connection` instance as an argument for final execution methods

Exactly – for a final methods. In case we add `Connection` as a constructor parameter for `Expression`, we shall pass it much earlier, than calling `one()` or `all()`

> `QueryBuilder` can always be determined from `Connection` thus using `Connection` is more flexible here.

Connection can always be determined from a `QueryBuilder` unambiguously. However, schema can provide different `QueryBuilders` depending on some conditions, so having, `Connection` we can not determine `QueryBuilder` with 100% guarantee.

> Particular developer may re-implement Db layer in the way `QueryBuilder` will not be in use.

I don't think we should consider corner-cases like this.

> Expression itself should not depend on Connection or be somehow bound ot it anywhere.
> ...
> If there is some construction, which should be build depending on particular DB schema - it should be incapsulated inside `QueryBuilder` - not in `Expression`.

I was thinking about it, but have not found intersections with other DBMSs to go this way: MySQL does not support arrays, [Oracle has collections](https://docs.oracle.com/cd/B28359_01/appdev.111/b28370/collections.htm#LNPLS00507) that must be typed (e.g. `user_ids_typ(123,456,789)`) thus `Expression` will have different constructor, No-SQL is a completely different story. 

> PSR standards using `__toString()` in thier interfaces all the time. Since we accepted several PSR for 2.1, we should pick this approach as well.

Ok, I'm fine with it, but what should this method provide in case when `Expression` is not static SQL and `Expression` is not bound to `Connection` or `QueryBuilder`?

klimov-paul: > MySQL does not support arrays

MySQL has `SET` data type - it is something similar to arrays.

klimov-paul: See #3796

framework/db/pgsql/ArrayExpression.php
cebe: > Instead of introducing build() method inside Expression and gess a good name for it - it is 
> better to adjsut QueryBuilder::build() in the way it can process both Query and Expression 
> and use it as universal interface.

I was thinking about this too but we don't have a Query or expression here, but a nesting, where a Query object resides inside of Expression. 

```php
$query->where(['@>', 'somecolumn', new ArrayExpression([(new Query())->...], 'integer')]);
```

So we need some way to add subqueries to the Expression so that Querybuilder can find it.
One way to allow this would be to allow Query() objects inside parameters, so that paramters that are a Query, would be replaced by the Query SQL string. But we need to check the consequences of supporting this first.

zuozp8: We still have issues with subqueries not being useable everywhere, like
`['in', 'id', [(new Query())->select(1), (new Query())->select(2)]]`
or even worse
`['in', ['id', 'id2'], [['id' => (new Query())->select(1), 'id2' => (new Query())->select(2)]]]`

Implementing it everywhere will be pain, solution to this is not obvious.

If we stick to this approach we'll have more complex code in most methods of Query builder.

We could pass this PR without subquery features, make
```php
interface ExpressionInterface
{
    …
    /**
     * @internal currently not used, introduced to avoid backward-incompatible changes, may be dropped
     */
    public function getSubQueries();
}
```

On the other I like the proposal by @cebe to keep subqueries in $params, as we can easily handle those at the end. Then I would make `Query` implements `ExpressionInterface`, as it should make QueryBuilder simpler and robust.

zuozp8: > The problem is this PR twist the original idea of yii\db\Expression.

Maybe `QueryPartInterface` would be more suitable then `ExpressionInterface` if we decide to put there features like subqueries.

SilverFire:
There is an alternate idea, suggested by @hiqsol in offline discussions.

1. Define `ExpressionInterface::getValue()` and `ExpressionInterface::getParams()`
2. Create DBMS-agnostic `ArrayExpression`, `JsonExpression` that implement `ExpressionInterface`
3. Define `ExpressionBuilderInterface::build($expression)`
4. Create `ArrayExpressionBuilder implements ExpressionBuilderInterface` that holds `QueryBuilder` reference and contains expression building logic.
5. Add `QueryBuilder::expressionBuilders` property that will map expression class on builder class like this:

```php
public $expressionBuilders = [
    \yii\db\ArrayExpression::class => \yii\db\pgsql\ArrayExpressionBuilder::class
]
```

This makes possible to create custom or DBMS- or project-specific expressions, for example `HstoreExpression`.

SilverFire:
Closing in favor of #15348

PR #14865, opened 2017-09-21
Fix #8452 Jquery Extraction (attempt 2)

### LINKED ISSUE #8452: Requirements free version of Yii 2

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #12588, preparing for #8452

Migrated from #13839


schmunk42:
Looks neat. 

Is this planned as a first step to keep the whole system intact, but should `yii\jquery` be moved to https://github.com/yiisoft/yii2-jquery in a next step?

samdark:
@schmunk42 yes, that's the plan.

framework/jquery/ActiveFormClientScript.php
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @author Paul Klimov <klimov.paul@gmail.com>
+ * @since 2.1.0
+ */
+class ActiveFormClientScript extends \yii\widgets\ActiveFormClientScript
+{
+    /**
+     * @inheritdoc
+     */
+    public $clientValidatorMap = [
```
SilverFire: I suggest to move this array to a separate method `getClientValidatorMap()` to make it easier to add custom validators to this map, or override existing.

SilverFire: Or, maybe, create a `ClientValidatorFactory` and inject it in `ActiveFormClientScript`.
It will be responsible for creating `ClientValidator` based on `Validator`, than it should be easier to add new mappings

SilverFire (changes requested):

framework/jquery/ActiveFormClientScript.php
klimov-paul: `defaultClientValidatorMap()` added

framework/jquery/ActiveFormClientScript.php
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * ActiveFormClientScript is a behavior for [[\yii\widgets\ActiveForm]], which allows composition
+ * of the client-side and AJAX form validation via underlying JQuery plugin.
+ *
+ * Usage example:
+ *
+ * ```php
+ * <?php $form = \yii\widgets\ActiveForm::begin([
+ *     'id' => 'example-form',
+ *     'as clientScript' => \yii\jquery\ActiveFormClientScript::class,
```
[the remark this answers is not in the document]
samdark: OK. That was 1st approach. Cons are described here: https://github.com/yiisoft/yii2/pull/13839 That's why behaviors.

framework/jquery/ActiveFormClientScript.php
```diff
@@ -0,0 +1,113 @@
...
+ * <?php \yii\widgets\ActiveForm::end(); ?>
+ * ```
+ *
+ * @see \yii\widgets\ActiveForm
+ * @see \yii\widgets\ActiveFormClientScriptBehavior
+ *
+ * @author Paul Klimov <klimov.paul@gmail.com>
+ * @since 2.1.0
+ */
+class ActiveFormClientScript extends \yii\widgets\ActiveFormClientScript
```
samdark: Should it mention explicitly that it's jQuery client script?

framework/captcha/CaptchaAsset.php
```diff
@@ -8,7 +8,7 @@
 namespace yii\captcha;
```
samdark: Captcha should be moved to separate repository. Script should be adjusted not to use jQuery.

Could be done in separate pull request after this one.

framework/jquery/validators/client/BooleanValidator.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @author Paul Klimov <klimov.paul@gmail.com>
+ * @since 2.1.0
+ */
+class BooleanValidator extends ClientValidator
+{
+    /**
+     * @inheritdoc
+     */
+    public function build($validator, $model, $attribute, $view)
```
samdark: It could not be build with any other validator than `BooleanValidator` from core framework so it makes sense to type-hint here and in similar cases.

samdark:
The pull request is big so I did a drawing while reviewing:

yii21_clientscript

1) `GridView` at core side is stripped down. `GridViewCleintScript` at jQuery extension side is a behavior that now contains all the stripped clientside-specific code.
2) Validator classes at jQuery extension side are mapped to core validators via `ActiveFormClientScript` behavior for `ActiveForm` (similar to point 1). Each validator at jQuery extension is a class that, accepting a corresponding validator instance from the core, builds JavaScript necessary for client validation.
3) Captcha is going to be extracted into its own extension. There isn't much jQuery so it can be rewritten using pure JavaScript so there are no jQuery dependency when it is required in a project.

Overall, approach makes perfect sense.

Sorry that it took that long to review it properly.

samdark (approved):

SilverFire (approved):

samdark:
@klimov-paul, it is possible to resolve conflicts? Other than that, the approach is good as well as code itself. The only minor thing to consider is https://github.com/yiisoft/yii2/pull/14865#discussion_r157327135. Ready to merge when conflicts will be resolved. If you're busy, I can allocate time for it and do it myself, just tell me.

dmirogin (approved):

klimov-paul:
Conflicts resolved.

samdark:
Merged. That was big one. Thank you for taking care of it.

PR #14888, opened 2017-09-29
Fixes #14129: Fixed console help to properly work with tricky camelcased controller names

### LINKED ISSUE #14129: Reflection Issue on Console Commands

I have the following console Command:

```php
use yii\console\Controller;
class EPagesController extends Controller
{
    [...]
    public function actionListShops()
    {
        [...]
    }
}
```

Note the second character of the Controller name: uppercase P. I think this is correct "Camelcase".

```

$ ./yii
- epages
    epages/fetch-orders-from-all-shops          Fetches all orders from all shops that uses the EPages REST API.
    epages/fetch-orders-from-shop               Fetches order from specific shop.
    epages/list-shops                           List all Shops that have the Epages REST-Api configured
```

### What do you get instead?

```
$./yii
- epages
- fixture 
```
- every action in epages is /missing/. 

Renaming the controller to EpagesController instead EPagesController fixes the issue. I think the problem is here: https://github.com/yiisoft/yii2/blob/master/framework/console/Controller.php#L566
The Reflection does not handle one-char subparts of an camelcase string properly.

There is more, tough:

Linux:
```
./yii epages/list-shops
Unknown command: epages/list-shops

Did you mean "epages/list-shops"?
[error]           Array content is
{
    "command": "epages/list-shops"
}
```
he does not find the command at all, regardless of lower or uppercase P !

Same on Windows 10 (cygwin bash) works properly but only with lowercase P:
```
./yii epages/list-shops
- works !
```

- Update: It works when i also rename the file to lowercase p. Win10 seems to have case-insensitive files thats why it works under win10... :) - but camelcase issue is still valid in my opinion.

Any Ideas ? Did i spot a tiny bug or am i understanding CamelCase wrong? Thanks a lot!
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | dev-master
| PHP version      | 7
| Operating system | observed on Win10 and Linux

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14129

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.13 under development
 ------------------------
 
+- Bug #14129: Fixed console help to properly work with tricky camelcased controller names (samdark)
 - Enh #14087: Added `yii\web\View::registerCsrfMetaTags()` method that registers CSRF tags dynamically ensuring that caching doesn't interfere (RobinKamps)
 - Bug #6526: Fixed `yii\db\Command::batchInsert()` casting of double values correctly independent of the locale (cebe, leammas)
 - Bug #7890: Allow `migrate/mark` to mark history at the point of the base migration (cebe)
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.13 under development
 ------------------------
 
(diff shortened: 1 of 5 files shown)


samdark:
@yiisoft/reviewers need your checks for this one.

SilverFire (approved):

samdark:
Thanks for tests, @SilverFire.

Issue #13486, opened 2017-02-01
how to set security flags (httpOnly and secure) for ALL cookies


leetas:
### What steps will reproduce the problem?
set custom cookies with Yii2

### What is the expected result?
All cookies has httpOnly and secure

### Additional info
My code:
```
'components' => [
        'session' => [
            'cookieParams' => [
                'httpOnly' => true,
                'secure' => true
            ]
        ],
        'cookies' => [
            'class' => 'yii\web\Cookie',
            'httpOnly' => true,
            'secure' => true
        ],
        'request' => [
            ....
            'csrfCookie' => [
                'httpOnly' => true,
                'secure' => true
            ]
        ],
        'user' => [
            ....
            'identityCookie' => [
                'name' => '_identity',
                'httpOnly' => true,
                'secure' => true,
            ],
        ],
```
1. I get session, csrf and identity cookies with required flags, but maybe possible set it with one config line?
2. All my custom cookies are without security flags. How to config Yii2 that all my cookies which I created with `Yii::$app->request->cookies;` will have security flags by default?

samdark:
There is currently no way to set defaults application-wide. Instantiation could go through dependency container. Then it would be possible.

PR #14891, opened 2017-09-29
Use DI container to instantiate cookies in order to be able to set defaults

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    |no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | 
| Fixed issues  | #13486

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -1499,7 +1499,8 @@ class Request extends \yii\base\Request
                 }
                 $data = @unserialize($data);
                 if (is_array($data) && isset($data[0], $data[1]) && $data[0] === $name) {
-                    $cookies[$name] = Yii::createObject('yii\web\Cookie', [
+                    $cookies[$name] = Yii::createObject([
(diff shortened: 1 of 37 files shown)


rob006 (approved):
Changelog line is missing.

dmirogin (approved):

framework/web/Request.php
```diff
@@ -1508,7 +1509,8 @@ protected function loadCookies()
             }
         } else {
             foreach ($_COOKIE as $name => $value) {
-                $cookies[$name] = new Cookie([
+                $cookies[$name] = Yii::createObject([
+                    'class' => 'yii\web\Cookie',
```
cebe: in the two cases above it is created from data, should we really allow modifying these?

samdark: Modifying data isn't the only case. You may set your own implementation as well. Not sure it's a good idea to use container in one place but not use it in another.

PR #14892, opened 2017-09-29
Fixes #7316: Documentation for ActiveRelationTrait::inverseOf()

### LINKED ISSUE #7316: HasMany to HasOne relation with inverseOf and indexBy
There is a problem in the one to many relation, with inverseOf and indexBy. The problem is relatively simple: $object->article->objects (returns array but only with the related object).

I have the following two Tables
- Article
  - articleID
- Object
  - objectID
  - articleID

The relations are like this:

``` php
// Object Model
public function getArticle() {
    return $this->hasOne(Article::className(), ['articleID' => 'articleID'])
        ->inverseOf('objects');
}
// Article Model
public function getObjects()
{
    return $this->hasMany(Object::className(), ['articleID' => 'articleID'])
        ->indexBy('objectID')
        ->inverseOf('article');
}
```

And now the problem:

``` php
$object = Object::findOne(['objectID' => 1]);
// $object has one Article and the article of the object has many objects.
var_dump($object->article->objects);
// Returns a array with only the one related object indexed by an integer 
// and not every related object of the article.
```

dmitry-kulikov:
Documentation.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7316

### DIFF
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -114,13 +114,12 @@ trait ActiveRelationTrait
 
     /**
      * Sets the name of the relation that is the inverse of this relation.
-     * For example, an order has a customer, which means the inverse of the "customer" relation
-     * is the "orders", and the inverse of the "orders" relation is the "customer".
+     * For example, a customer has orders, which means the inverse of the "orders" relation is the "customer".
      * If this property is set, the primary record(s) will be referenced through the specified relation.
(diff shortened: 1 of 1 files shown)


samdark (approved):

rob006 (approved):

samdark:
Merged. Thank you!

PR #14893, opened 2017-09-29
Fixes #7316: inverseOf should be ignored for "multiple" relation ("many" in 1-to-many or many-to-many).

### LINKED ISSUE #7316: HasMany to HasOne relation with inverseOf and indexBy

dmitry-kulikov:
Currently inverseOf works by design, but behavior rather tricky, this feature is very close to bug, please see related issue. Unfortunately this PR breaks BC, especially for people who used inverseOf with asArray.

| Q             | A
| ------------- | ---
| Is bugfix?    | somewhat
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #7316

### DIFF
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -114,11 +114,14 @@ trait ActiveRelationTrait
 
     /**
      * Sets the name of the relation that is the inverse of this relation.
-     * For example, an order has a customer, which means the inverse of the "customer" relation
-     * is the "orders", and the inverse of the "orders" relation is the "customer".
-     * If this property is set, the primary record(s) will be referenced through the specified relation.
+     * For example, a customer has orders, which means the inverse of the "orders" relation
+     * is the "customer". Note that "customer" relation has not inverse relation "orders",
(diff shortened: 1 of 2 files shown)


rob006:
See https://github.com/yiisoft/yii2/issues/7316#issuecomment-334929080

PR #14894, opened 2017-09-30
Add pagination method

oldy777:
Added method paginate in ActiveQuery class for quick get pagination data

How to use:
```
$result = Article::find()->where(['status' => 1])->paginate(15);
                 'items' => $result['items'],
                 'pages' => $result['pages'],
           ]);
```

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -840,7 +840,8 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         $query = $this;
         $countQuery = clone $query;
 
-        $paginator = \Yii::$container->get('yii\data\Pagination',['totalCount' => $countQuery->count()]);
+        $paginator = \Yii::$container->get('yii\data\Pagination');
+        $paginator->totalCount = $countQuery->count();
         $paginator->pageParam = $pageName;
 
         if($page){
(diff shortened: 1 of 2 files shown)


dmirogin:
You can use ActiveDataProvider with configured paginator instead this.

oldpunk:
@dmirogin I think it's not so easy and obviously. Especially when you are newbie in framework. With this construction you can do it in 1 line, and will have code autocomplete

samdark:
It's not the job of the query to do pagination. Existing solution is OK regarding complexity and is described well in the guide.

oldpunk:
@samdark can I dynamically add this method to the class?

samdark:
You can use your own query classes extended from your own base class that contains the method: http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#customizing-query-classes

oldpunk:
@samdark but what if already have a bunch of model files? Maybe it is a good idea add ability to add custom methods like my pagination?

samdark:
You have to do it once for all your models where you're using pagination for. If these don't need any other custom query classes, they can instantiate the same query class.

PR #14895, opened 2017-09-30
Simplifying access to ActiveQuery from ActiveRecord

oldy777:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Use case:
Old
```
$customer = Customer::find()->select(['*', '([[status]]*2) AS [[status2]]'])->where(['name' => 'user3'])->one();
```

New
```
$customer = Customer::select(['*', '([[status]]*2) AS [[status2]]'])->where(['name' => 'user3'])->one();
```

Old
```
Customer::find()->count()
```

New
```
Customer::count()
```

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -816,54 +816,6 @@ class ActiveQuery extends Query implements ActiveQueryInterface
     }
 
     /**
-     * Paginate the given query.
-     * ```php
-     * public function getItems()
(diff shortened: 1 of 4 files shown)


samdark:
Sorry but won't accept this change. While syntax is getting shorter, these calls are too implicit.

Issue #6226, opened 2014-11-24
symlink(): File exists


stepanselyuk:
Hi,
I got that error sometimes after I've deleted symlinks from `web/assets` manual or in the code. I can run `clearstatcache( true )` but it worked per thread, so other theads will produce the error until `realpath_cache` will be reloaded. How to fix it? I'm talking about `\yii\web\AssetManager::publishFile` and `\yii\web\AssetManager::publishDirectory`. Maybe try to catch that error and make `clearstatcache( true )`, then try again? I've googled that problem, and it's common problem for PHP.

dynasource:
not able to reproduce

Renkas:
I'm getting the same error from time to time. And I can reproduce it easily when creating a bunch of request to servers. I have an admin tool that opens a URL for each product in a iframe on same page (10+ per page) and about 1/3 or 1/4 of these requests give me said error.

dynasource:
thanks for sharing your reproduction

cebe:
could be solved in a similar way as #9288

dynasource:
fixed by https://github.com/yiisoft/yii2/pull/14905

PR #14905, opened 2017-10-03
fix for #6226, symlinking files/directories during multithreading

dynasource:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes (manually)
| Fixed issues  | https://github.com/yiisoft/yii2/issues/6226

this patch solves the symlink issues in multi threaded environments. It can occur during parallel testing the Yii application in which symlinks are created by multiple threads, causing collisions.

### DIFF
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -482,7 +482,13 @@ class AssetManager extends Component
 
         if ($this->linkAssets) {
             if (!is_file($dstFile)) {
-                symlink($src, $dstFile);
+                try { // fix #6226 symlinking multi threaded
+                    symlink($src, $dstFile);
+                } catch (\Exception $e) {
+                    if (!is_file($dstFile)) {
+                        throw $e;
+                    }
+                }
             }
         } elseif (@filemtime($dstFile) < @filemtime($src)) {
             copy($src, $dstFile);
@@ -521,7 +527,13 @@ class AssetManager extends Component
         if ($this->linkAssets) {
             if (!is_dir($dstDir)) {
                 FileHelper::createDirectory(dirname($dstDir), $this->dirMode, true);
-                symlink($src, $dstDir);
+                try { // fix #6226 symlinking multi threaded
+                    symlink($src, $dstDir);
+                } catch (\Exception $e) {
+                    if (!is_dir($dstDir)) {
+                        throw $e;
+                    }
+                }
             }
         } elseif (!empty($options['forceCopy']) || ($this->forceCopy && !isset($options['forceCopy'])) || !is_dir($dstDir)) {
             $opts = array_merge(
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.13 under development
 ------------------------
-
+- Bug #6226: Fix fatal symlink error when publishing in multi threaded environments (dynasource)
 - Bug #14129: Fixed console help to properly work with tricky camelcased controller names (samdark, silverfire)
 - Enh #14126: Added variadic parameters support to DI container (SamMousa)
 - Enh #14087: Added `yii\web\View::registerCsrfMetaTags()` method that registers CSRF tags dynamically ensuring that caching doesn't interfere (RobinKamps)


framework/web/AssetManager.php
```diff
@@ -482,7 +482,13 @@ protected function publishFile($src)
 
         if ($this->linkAssets) {
             if (!is_file($dstFile)) {
-                symlink($src, $dstFile);
+                try { // fix #6226 symlinking multi threaded
+                    symlink($src, $dstFile);
+                } catch (\Exception $e) {
+                    if (!is_file($dstFile)) {
```
samdark: What if there was a file with the same name you tried creating symlink with?

framework/web/AssetManager.php
```diff
@@ -521,7 +527,13 @@ protected function publishDirectory($src, $options)
         if ($this->linkAssets) {
             if (!is_dir($dstDir)) {
                 FileHelper::createDirectory(dirname($dstDir), $this->dirMode, true);
-                symlink($src, $dstDir);
+                try { // fix #6226 symlinking multi threaded
+                    symlink($src, $dstDir);
+                } catch (\Exception $e) {
+                    if (!is_dir($dstDir)) {
```
samdark: What if there was a directory with the same name you tried creating symlink with?

samdark (changes requested):

framework/web/AssetManager.php
dynasource: this already is a very rare usecase. It highly likely that the symlink is already correct because it did not exist a fraction of time ago https://github.com/yiisoft/yii2/pull/14905/files/54ac1820734bc85dd10b31492c16751b7582b94e#diff-8b967e883b8604768a6f7251d1f1285cR484

Therefore, if an error occurs and it is because a file already exist, it is because of a collision of the same symlink action (happening in a multi threading env)

samdark: OK. Makes sense.

samdark:
Have you verified that it works well on Linux and Windows (symlinks work a bit different there about being considered a file/directory).

samdark (approved):

samdark:
@yiisoft/reviewers need your input.

schmunk42:
I think it's good if the framework handles such edge cases, but wouldn't it be better to detect these cases with (an additionally) [`is_link()`](http://php.net/manual/en/function.is-link.php)?

dynasource:
> Have you verified that it works well on Linux and Windows (symlinks work a bit different there about being considered a file/directory).

@samdark, I have tested it in my Windows environment and it solves these kind of issues I am having when testing multithreaded. I haven't tested it in Linux, but it does not seem necessary with respect to the nature of this patch.

> I think it's good if the framework handles such edge cases, but wouldn't it be better to detect these cases with (an additional) is_link()?

I am positive to put this kind of repetitive logic into helpers. We can open another topic to proceed to such next step.

samdark:
@schmunk42 no. `is_link()` is totally unreliable on some FS.

samdark:
@dynasource OK. Then the patch looks good to me. Need another approval to merge it.

rob006:
Shouldn't we apply similar fix for copying assets?

https://github.com/yiisoft/yii2/blob/54ac1820734bc85dd10b31492c16751b7582b94e/framework/web/AssetManager.php#L493-L498
https://github.com/yiisoft/yii2/blob/54ac1820734bc85dd10b31492c16751b7582b94e/framework/web/AssetManager.php#L538-L560

dynasource:
@rob006, I don't seem to have any issues in case of a `$linkAssets===false` scenario

schmunk42:
@dynasource It might be not directly related to your change, but I was also thinking about the case that you switch from `linkAssets = false` to `true`.

Let's say you have already published a *real file/dir* under a certain hash and now you try to `linkAssets` - wouldn't the current code silently omit the symlink creation and therefore assets would act weird, like not reflecting changes made to a source file, because it is already published (and not linked)?

I haven't tried or verified this, it just came to my mind from reading the code. If so the application would have to throw an exception saying that there's a file, but it's not a link, if `linkAssets = true`.

Sorry if this is too much off-topic.

schmunk42:
> @schmunk42 no. is_link() is totally unreliable on some FS.

@samdark OK, did know that. I had my answer open for some hours and did not see your answer.

dynasource:
> Let's say you have already published a real file/dir under a certain hash and now you try to linkAssets - wouldn't the current code silently omit the symlink creation and therefore assets would act weird, like not reflecting changes made to a source file, because it is already published (and not linked)?

@schmunk42, that is indeed as usecase which is not covered by the core yet. Currently, people have to manually reset their assets to fix that.

rob006:
> It might be not directly related to your change, but I was also thinking about the case that you switch from linkAssets = false to true.

I don't think that we should care about this. If you change component settings, you should also clear its cache. 

@dynasource I don't see a reason why race condition should happen only for symlinks.

dynasource:
theoretically yes, practically over here not yet

SilverFire (approved):
Good to merge

rob006:
@dynasource 
> theoretically yes, practically over here not yet

If we know that race condition is possible why not just fix it instead of waiting for bug report?

dynasource:
because I'm not getting fatal errors in the copy scenario

PR #14915, opened 2017-10-04
add value of linkAssets to string which is to be hashed, fixes #14913

### LINKED ISSUE #14913: Improvements when switching to linked assets in development
Continued from https://github.com/yiisoft/yii2/pull/14905#issuecomment-334139197

(during development)
- publish assets with `linkAssets = false`
- publish assets with `linkAssets = true`
- update source files

- changed assets, since they should be linked

### What do you get instead?

- old published (copied) assets from first request

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.x

From the thread linked above...

> I don't think that we should care about this. If you change component settings, you should also clear its cache.

So this should either be noted [in the guide](http://www.yiiframework.com/doc-2.0/guide-structure-assets.html#asset-publishing) or we also take `linkAssets` into account when generating the hash.

schmunk42:
| Q             | A
| ------------- | ---
| Is bugfix?    | in edge-cases
| New feature?  | no
| Breaks BC?    | no (hash would be different with new version anyway)
| Tests pass?   | ?
| Fixed issues  | #14913

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -81,7 +81,7 @@ Yii Framework 2 Change Log
 - Chg #14321: `yii\widgets\MaskedInput` is now registering its JavaScript `clientOptions` initialization code in head section (DaveFerger)
 - Chg #14487: Changed i18n message error to warning (dmirogin)
 - Enh #14864: Ability to use dependencies in constructor of migrations (vtvz)
-- Enh #14913: Improvements when hashing linked assets (schmunk42)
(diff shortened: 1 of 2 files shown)


samdark (approved):

samdark:
@yiisoft/reviewers need your input.

rob006 (approved):

dmirogin (approved):

samdark:
Merged. Thank you!

PR #14923, opened 2017-10-05
Reformat

243083df:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no(for 2.1)
| Tests pass?   | yes
| Fixed issues  | -

### DIFF
--- a/framework/helpers/BaseConsole.php
+++ b/framework/helpers/BaseConsole.php
@@ -359,32 +359,32 @@ class BaseConsole
     {
         $styleMap = [
                 // http://www.w3.org/TR/CSS2/syndata.html#value-def-color
-                self::FG_BLACK    => ['color' => 'black'],
-                self::FG_BLUE     => ['color' => 'blue'],
-                self::FG_CYAN     => ['color' => 'aqua'],
-                self::FG_GREEN    => ['color' => 'lime'],
-                self::FG_GREY     => ['color' => 'silver'],
+                self::FG_BLACK => ['color' => 'black'],
+                self::FG_BLUE => ['color' => 'blue'],
+                self::FG_CYAN => ['color' => 'aqua'],
+                self::FG_GREEN => ['color' => 'lime'],
+                self::FG_GREY => ['color' => 'silver'],
                 // http://meyerweb.com/eric/thoughts/2014/06/19/rebeccapurple/
                 // http://dev.w3.org/csswg/css-color/#valuedef-rebeccapurple
-                self::FG_PURPLE   => ['color' => 'rebeccapurple'],
-                self::FG_RED      => ['color' => 'red'],
-                self::FG_YELLOW   => ['color' => 'yellow'],
-                self::BG_BLACK    => ['background-color' => 'black'],
-                self::BG_BLUE     => ['background-color' => 'blue'],
-                self::BG_CYAN     => ['background-color' => 'aqua'],
-                self::BG_GREEN    => ['background-color' => 'lime'],
-                self::BG_GREY     => ['background-color' => 'silver'],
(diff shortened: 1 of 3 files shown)


framework/helpers/BaseConsole.php
```diff
@@ -358,32 +358,32 @@ public static function ansiStrlen($string)
     public static function ansiToHtml($string, $styleMap = [])
     {
         $styleMap = [
-            // http://www.w3.org/TR/CSS2/syndata.html#value-def-color
-            self::FG_BLACK => ['color' => 'black'],
```
schmunk42: Array lists should not be indented, since all lines will change, if just one longer entry is added.

framework/helpers/BaseHtml.php
```diff
@@ -147,7 +147,7 @@ public static function tag($name, $content = '', $options = [])
         if ($name === null || $name === false) {
             return $content;
         }
-        $html = "<$name" . static::renderTagAttributes($options) . '>';
+        $html = '<' .$name . static::renderTagAttributes($options) . '>';
```
rob006: Concatenation is less readable and slower. Why change that?

243083df: In this case concatenation is more readable i think. There is no mix of concat and quotes, and seems more clear that brackets surround content.
And concatenation faster and use less memory.

framework/helpers/BaseConsole.php
243083df: Thank, forget about this.

framework/helpers/BaseHtml.php
rob006: >  And concatenation faster and use less memory.

Nope.
https://jonczyk.me/2016/10/23/phpyths-buster-great-string-performance-test/

framework/helpers/BaseConsole.php
```diff
@@ -358,33 +358,33 @@ public static function ansiStrlen($string)
...
-            self::BG_YELLOW => ['background-color' => 'yellow'],
-            self::BOLD => ['font-weight' => 'bold'],
-            self::ITALIC => ['font-style' => 'italic'],
-            self::UNDERLINE => ['text-decoration' => ['underline']],
-            self::OVERLINED => ['text-decoration' => ['overline']],
-            self::CROSSED_OUT => ['text-decoration' => ['line-through']],
-            self::BLINK => ['text-decoration' => ['blink']],
-            self::CONCEALED => ['visibility' => 'hidden'],
-        ] + $styleMap;
+                // http://www.w3.org/TR/CSS2/syndata.html#value-def-color
```
schmunk42: Also the tab is too much.

framework/helpers/BaseHtml.php
schmunk42: @rob006 It's really not that important, but

> Conclusion
> My tests prove that it doesn’t matter if you use single or double quotes both in PHP 5.6 and 7. In concatenation test usage of dots and variable inside string were very close to each other, so I call it even.

rob006: @schmunk42 Yes, but if it is less readable (at least for me) and slower, why change that? Performance would be only valid reason for such change, but this is not the case.

schmunk42: I also thought it's more readable and no mix of quotes and there's no difference in slower/faster (but I really don't wanna debate this :) in this particular case) - it's fine. But mostly I don't care about this change.

When using double-quotes I usually enclose variables in `{}`, so what's about? :)

    "<{$name} " . static::renderTagAttributes($options)} . ">"

samdark (approved):

samdark:
Merged. Thanks!

PR #14930, opened 2017-10-05
Fixes #14929: Add support of semicolon-less JS to asset/compress comm…

### LINKED ISSUE #14929: asset/compress needs semicolon-less JavaScripts support
New released yii-pjax 2.0.7 breaks my compressed JavaScript which generated by `asset/compress` command. JS files are joined with no separators after `jquery.pjax.js`.

Create an Yii project, then simply use `yii asset/compress`.

At least a semicolon should be inserted. I prefer wrapping each scripts by `function(){ ... }();`.

### What do you get instead?
no-semicolon

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.0.15, 7.1.8
| Operating system |Ubuntu, MacOS

tanakahisateru:
My idea to fix #14929 by wrapping with function scope block.

…and.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14929

### DIFF
--- a/framework/console/controllers/AssetController.php
+++ b/framework/console/controllers/AssetController.php
@@ -568,8 +568,12 @@ EOD;
     {
         $content = '';
         foreach ($inputFiles as $file) {
+            $fileContent = rtrim(file_get_contents($file));
+            if (substr($fileContent, -1) !== ';') {
+                $fileContent .= ';';
+            }
             $content .= "/*** BEGIN FILE: $file ***/\n"
-                . file_get_contents($file)
+                . $fileContent . "\n"
                 . "/*** END FILE: $file ***/\n";
         }
         if (!file_put_contents($outputFile, $content)) {
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.13 under development
 ------------------------
 
+- Enh #14929: Ensure trailing `;` on combining files with `asset` command to fix compiler failures (tanakahisateru)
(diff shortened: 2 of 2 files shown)


shirase:
Global variable will be undefined beyond wrapper function context.
```
var test =""; // is global, but
(function() {
var test =""; // is not global
})();
```

tanakahisateru:
To insert a ";" doesn't break BC. Should I change?

tanakahisateru:
I'm sure that wrapping was too aggressive. I will change this PR to avoid function.

samdark:
Isn't https://github.com/yiisoft/jquery-pjax/commit/0d4cc3f479d4b7b800aefca1275a852c8d3dd3c6 fixing the problem?

framework/console/controllers/AssetController.php
```diff
@@ -569,7 +569,9 @@ public function combineJsFiles($inputFiles, $outputFile)
         $content = '';
         foreach ($inputFiles as $file) {
             $content .= "/*** BEGIN FILE: $file ***/\n"
+                . "(function(){\n"
                 . file_get_contents($file)
```
samdark: It's better to check if trimmed contents end with `;` and add it if it's not there.

samdark (changes requested):

tanakahisateru:
@samdark I've modified the commit.

rob006 (approved):
Changelog entry is missing.

samdark (approved):

PR #14941, opened 2017-10-07
Attempt to fix tests for HHVM

rob006:
See https://github.com/yiisoft/yii2/commit/f5591bb70eb9f4622ffa8b435e7d88543bcdb44a#commitcomment-24827510

This is probably only caching issue. Each test should use unique category in order to bypass HHVM cache from previous tests.

/cc @samdark


samdark:
Thank you for taking care of it!

PR #14952, opened 2017-10-11
phpDoc to BaseManager::defaultRoles

### LINKED ISSUE #13640: Rbac  validate permission for guest user when roles are in defaultRoles
Setting defaultRoles
then accessing a page which need one of the roles to as guest user

I expect that only authenticated user can access that page.
According to the docs:
$defaultRoles A list of role names that are assigned to every user automatically without calling assign().
In my view user should be intended as authenticated user since if the page is accessible to guest user I won't implement access checking in it and therefore I do not need to give defaultRoles to guest.

### What do you get instead?
Guest user can load the page

Since the line
if (isset($assignments[$itemName]) || in_array($itemName, $this->defaultRoles)) 
which cause the validation is present in more than one files I think this is the intended behaviour.

So at least is needed a clarification in the docs because such misunderstanding can lead to security problem.
Then I propose to add something like
defaultRolesLoggedUser

and eventually rename the actual config param to something more clear
defaultRolesGuestUser

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | PHP 7.0.13-0
| Operating system | ubuntu0.16.10.1

githubjeka:
fixes #13640

### DIFF
--- a/framework/rbac/BaseManager.php
+++ b/framework/rbac/BaseManager.php
@@ -26,6 +26,7 @@ abstract class BaseManager extends Component implements ManagerInterface
 {
     /**
      * @var array a list of role names that are assigned to every user automatically without calling [[assign()]].
+     * Note that these roles are applied to users, regardless of their state of authentication.
      */
     public $defaultRoles = [];
 


samdark:
Thanks!

Issue #14489, opened 2017-07-19
PHP 7.2: count(): Parameter must be an array or an object that implements Countable


cebe:
### What steps will reproduce the problem?

A lot of tests on PHP 7.2 are failing with

```
count(): Parameter must be an array or an object that implements Countable
```

https://travis-ci.org/yiisoft/yii2/jobs/255175595#L670

`count()` throws a warning since PHP 7.2 in case a scalar or non-countable object is passed. We rely on this behavior in many places so code needs to be adjusted.

### More info

- https://wiki.php.net/rfc/counting_non_countables
- http://php.net/manual/en/function.count.php

tmli3b3rm4n:
same

samdark:
https://travis-ci.org/yiisoft/yii2/jobs/283264298

Issue #14811, opened 2017-09-13
PHP 7.2: session_cache_limiter(): Cannot change cache limiter when session is active


Thoulah:
### What steps will reproduce the problem?
Hitting a page with `HttpCache::className()` defined in `behaviours` in the controller

### What is the expected result?
A working page

### What do you get instead?
```php
PHP Warning – yii\base\ErrorException

session_cache_limiter(): Cannot change cache limiter when session is active
1. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/filters/HttpCache.php at line 192
2. yii\base\ErrorHandler::handleError(2, 'session_cache_limiter(): Cannot ...', '/homepages/user/htdocs/...', 192, ...)
3. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/filters/HttpCache.php at line 192 – session_cache_limiter('')
4. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/filters/HttpCache.php at line 138 – yii\filters\HttpCache::sendCacheControlHeader()
5. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/base/ActionFilter.php at line 75 – yii\filters\HttpCache::beforeAction(yii\base\InlineAction)
6. yii\base\ActionFilter::beforeFilter(yii\base\ActionEvent)
7. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/base/Component.php at line 557 – call_user_func([yii\filters\HttpCache, 'beforeFilter'], yii\base\ActionEvent)
8. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/base/Controller.php at line 274 – yii\base\Component::trigger('beforeAction', yii\base\ActionEvent)
9. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/web/Controller.php at line 164 – yii\base\Controller::beforeAction(yii\base\InlineAction)
10. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/base/Controller.php at line 155 – yii\web\Controller::beforeAction(yii\base\InlineAction)
11. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/base/Module.php at line 528 – yii\base\Controller::runAction('index', [])
12. in /homepages/user/htdocs/yii/vendor/yiisoft/yii2/web/Application.php at line 103 – yii\base\Module::runAction('lyrics/index', [])
[...]

PR #14959, opened 2017-10-13
Fixes for PHP 7.2 compatibility

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no, but will
| Fixed issues  | #14489

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -179,14 +179,16 @@ before_script:
     fi
 
   # disable DEPRECATE messages for PHP 7.2. We need to update PHPUnit to get rid of them.
-  - if [[ $TRAVIS_PHP_VERSION = 7.2.0* ]]; then echo 'error_reporting = E_ALL & ~E_DEPRECATED' >> /etc/php/php.ini; fi
+  - if [[ $TRAVIS_PHP_VERSION = 7.2.0* ]]; then
+      PHP_FLAGS="-d error_reporting='E_ALL & ~E_DEPRECATED'"
+    fi
 
 
 script:
   # PHP tests
   - |
     if [ $TASK_TESTS_PHP == 1 ]; then
-      vendor/bin/phpunit --verbose $PHPUNIT_FLAGS --exclude-group mssql,oci,wincache,xcache,zenddata,cubrid
+      php $PHP_FLAGS vendor/bin/phpunit --verbose $PHPUNIT_FLAGS --exclude-group mssql,oci,wincache,xcache,zenddata,cubrid
     fi
 
   # JS tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -180,10 +180,10 @@ before_script:
 
   # disable DEPRECATE messages for PHP 7.2. We need to update PHPUnit to get rid of them.
   - |
-    if [[ $TRAVIS_PHP_VERSION = 7.2.0* ]]; then
-      echo 'Disabled DEPRECATED notifications for PHP 7.2.0'
-      echo 'error_reporting = E_ALL & ~E_DEPRECATED' >> /tmp/php-config.ini
-      phpenv config-add /tmp/php-config.ini
+    if [ $TRAVIS_PHP_VERSION == 7.2.0* ]; then
+      echo 'Disabled DEPRECATED notifications for PHP 7.2.0';
+      echo 'error_reporting = E_ALL & ~E_DEPRECATED' >> /tmp/php-config.ini;
+      phpenv config-add /tmp/php-config.ini;
     fi
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -178,12 +178,10 @@ before_script:
       PHPUNIT_FLAGS="--coverage-clover=coverage.clover"
     fi
 
-  - echo $TRAVIS_PHP_VERSION
-
   # disable DEPRECATE messages for PHP 7.2. We need to update PHPUnit to get rid of them.
   - |
-    if [ $TRAVIS_PHP_VERSION == 7.2.0 ]; then
-      echo 'Disabled DEPRECATED notifications for PHP 7.2.0';
+    if [ $TRAVIS_PHP_VERSION == 7.2 ]; then
+      echo 'Disabled DEPRECATED notifications for PHP 7.2';
       echo 'error_reporting = E_ALL & ~E_DEPRECATED' >> /tmp/php-config.ini;
       phpenv config-add /tmp/php-config.ini;
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -148,6 +148,7 @@ before_script:
 
   # show some versions and env information
   - php --version
+  - php --ini
   - composer --version
   - |
     if [ $TASK_TESTS_PHP == 1 ]; then
@@ -176,6 +177,9 @@ before_script:
     if [ $TASK_TESTS_COVERAGE == 1 ]; then
       PHPUNIT_FLAGS="--coverage-clover=coverage.clover"
     fi
+
+  # disable DEPRECATE messages for PHP 7.2. We need to update PHPUnit to get rid of them.
+  - if [[ $TRAVIS_PHP_VERSION = 7.2.0* ]]; then echo 'error_reporting = E_ALL & ~E_DEPRECATED' >> /etc/php/php.ini; fi
 
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -178,7 +178,8 @@ before_script:
       PHPUNIT_FLAGS="--coverage-clover=coverage.clover"
     fi
 
-  # disable DEPRECATE messages for PHP 7.2. We need to update PHPUnit to get rid of them.
+  # Disable DEPRECATE messages during PHPUnit initialization on PHP 7.2. To fix them, PHPUnit should be updated to 6.*
+  # For Yii2 tests, messages will be enabled by tests/bootstrap.php
   - |
     if [ $TRAVIS_PHP_VERSION == 7.2 ]; then
       echo 'Disabled DEPRECATED notifications for PHP 7.2';
--- a/.travis.yml
+++ b/.travis.yml
@@ -179,7 +179,8 @@ before_script:
     fi
 
   # disable DEPRECATE messages for PHP 7.2. We need to update PHPUnit to get rid of them.
-  - if [[ $TRAVIS_PHP_VERSION = 7.2.0* ]]; then
+  - |
+    if [[ $TRAVIS_PHP_VERSION = 7.2.0* ]]; then
       PHP_FLAGS="-d error_reporting='E_ALL & ~E_DEPRECATED'"
     fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -148,7 +148,6 @@ before_script:
 
   # show some versions and env information
   - php --version
-  - php --ini
   - composer --version
   - |
     if [ $TASK_TESTS_PHP == 1 ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,9 +55,9 @@ addons:
 matrix:
   fast_finish: true
   include:
-    # run tests coverage on PHP 7.1
     - php: 7.2
 
+    # run tests coverage on PHP 7.1
     - php: 7.1
       env: TASK_TESTS_COVERAGE=1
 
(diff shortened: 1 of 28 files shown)


SilverFire:
There are still issues with `session_cache_limiter()`
https://github.com/yiisoft/yii2/blob/5a8c3d537ba14dbff02b2adee9072d381b9eb6b7/framework/filters/HttpCache.php#L192

And with `session_set_cookie_params()`
https://github.com/yiisoft/yii2/blob/5a8c3d537ba14dbff02b2adee9072d381b9eb6b7/framework/web/Session.php#L381

They work correctly on PHP 7.1, but not on 7.2. Is seems that PHPUnit does not restore environment after tests correctly. Maybe it is fixed in PHPUnit 6.*, but it requires PHP 7 or higher. 

I would not like to skip these tests on PHP 7.2, but I don't have a clue how to fix it. I will be happy to hear some suggestions @yiisoft/reviewers @yiisoft/core-developers

SilverFire:
Checked with PHPUnit 6.4.1 on PHP 7.2 – the problem is still present.

rob006:
There is a issue about `session_cache_limiter()`: https://github.com/yiisoft/yii2/issues/14811

SilverFire:
Assuming our rare releases schedule, I'd prefer to put 7.2 compatibility fixes in 2.0.13

SilverFire:
> When headers are already sent and try to set new INI values, session_name(),
      session_module_name(), session_save_path(), session_cache_limiter() and
      session_cache_expire() are no longer works. Older PHPs accepts new values even
      if new values will not be effective.
      This new corrected behavior may affect command line mode CLI scripts that manage
      sessions. Use output buffer just like web applications to resolve problems on
      CLI scripts.

From PHP 7.2 [upgrade notes](https://github.com/php/php-src/blob/php-7.2.0RC4/UPGRADING)

bscheshirwork:
> This new corrected behavior may affect command line mode CLI scripts that manage
sessions. Use output buffer just like web applications to resolve problems on
CLI scripts.

This is only sessions, not headers?
We need change it?
https://github.com/yiisoft/yii2/blob/277468a7aa7c182ca7a686d3d82a2e38d897a980/framework/base/ErrorHandler.php#L100-L104

SilverFire:
Partially fixed, but there is a mess with date parsing. Don't have enough time to check it now. If anybody can - it will be good

tests/framework/i18n/FormatterTest.php
```diff
@@ -293,18 +294,26 @@ public function lengthDataProvider()
     /**
      * @dataProvider lengthDataProvider
      */
-    public function testIntlAsLength($message, $arguments, $expected)
+    public function testIntlAsLength($message, $arguments, $expected, $_, $expectedException = [])
```
klimov-paul: Why variable named `$_`?

framework/filters/HttpCache.php
```diff
@@ -182,6 +182,10 @@ protected function validateCache($lastModified, $etag)
      */
     protected function sendCacheControlHeader()
     {
+        if (headers_sent($file, $line)) {
```
klimov-paul: Is this check really necessary? Would not PHP throw its own error saying 'headers have already been sent'?

klimov-paul (changes requested):

tests/ResultPrinter.php
```diff
@@ -0,0 +1,29 @@
...
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yiiunit;
+
+/**
+ * IsOneOfAssert asserts that the value is one of the expected values.
+ */
+class ResultPrinter extends \PHPUnit\TextUI\ResultPrinter
```
klimov-paul: Using custom `ResultPrinter` as a solution, does not seem right.
What if some developer creates component or extension, which operates session and wishes to write a unit tests for it? He will have to create his own `ResultPrinter` as well, which does not feel right.

We should ensure framework testability.

klimov-paul (changes requested):

tests/framework/i18n/FormatterTest.php
SilverFire: It contains shotLength expectation and is unused in this test.

klimov-paul: Fact that it is unused does not justifies such variable naming.

tests/ResultPrinter.php
bscheshirwork: +1
I must been redefine PHPStorm printer - his realization send headers too.
```php
class PhpStorm_Codeception_ReportPrinter_Redefine extends PHPUnit_TextUI_ResultPrinter
{
    /**
     * @inheritDoc
     */
    public function __construct(
        $out = null,
        $verbose = false,
        $colors = self::COLOR_DEFAULT,
        $debug = false,
        $numberOfColumns = 80,
        $reverse = false
    ) {
        parent::__construct(STDOUT, $verbose, $colors, $debug, $numberOfColumns, $reverse);
    }

    /**
     * Flush buffer and close output if it's not to a STDOUT stream
     */
    public function flush()
    {
        if ($this->out != STDOUT) {
            parent::flush();
        }
    }
}

class PhpStorm_Codeception_ReportPrinter extends PhpStorm_Codeception_ReportPrinter_Redefine
...
```

SilverFire: It's not a final solution, I'm just trying to make it work somehow :)
There is an issue in PHPUnit repo https://github.com/sebastianbergmann/phpunit/issues/1193

bscheshirwork: @SilverFire yes, I tell about this problem. p.s.: Don't forgot `flush()`

SilverFire:
Reported 2 regressions in PHP core:

https://bugs.php.net/bug.php?id=75378
https://bugs.php.net/bug.php?id=75380

All tests except DB are passing now.

tests/ResultPrinter.php
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yiiunit;
+
+/**
+ * IsOneOfAssert asserts that the value is one of the expected values.
```
samdark: phpdoc looks wrong.

framework/web/Session.php
```diff
@@ -376,6 +376,9 @@ public function setCookieParams(array $value)
      */
     private function setCookieParamsInternal()
     {
+        if (headers_sent($file, $line)) {
```
samdark: Won't `session_set_cookie_params()` trigger the same exception natively?

tests/ResultPrinter.php
samdark: phpunit issue should be referenced. It's not clear about the purpose of the printer from the code itself.

tests/framework/helpers/HtmlTest.php
```diff
@@ -140,12 +140,17 @@ public function testCsrfMetaTagsEnableCsrfValidation()
...
+            'components' => [
+                'request' => [
+                    'class' => 'yii\web\Request',
+                    'enableCsrfValidation' => true,
+                ]
+            ],
+        ]);
+        $this->expectException('yii\base\InvalidConfigException');
+        $this->expectExceptionMessage('yii\web\Request::cookieValidationKey must be configured with a secret key.');
+        Html::csrfMetaTags();
```
samdark: Isn't assertion with a pattern missing here?

tests/framework/i18n/IntlTestHelper.php
```diff
@@ -24,10 +24,18 @@ public static function setIntlStatus($test)
         {
             static::$enableIntl = null;
             if (strncmp($test->getName(false), 'testIntl', 8) === 0) {
+                static::$enableIntl = true;
+
+                if (strpos(PHP_VERSION, '7.2.0RC') !== false && version_compare(PHP_VERSION, '7.2.0.RC.3', '<=')) {
```
samdark: Why using both strpos and version_compare?

tests/framework/validators/FileValidatorTest.php
```diff
@@ -388,15 +388,19 @@ public function testValidateMimeTypeMaskInvalid($fileName, $mask)
...
+        return array_filter([
             ['test.svg', 'image/*', 'svg'],
             ['test.jpg', 'image/*', 'jpg'],
             ['test.png', 'image/*', 'png'],
             ['test.png', 'IMAGE/*', 'png'],
             ['test.txt', 'text/*', 'txt'],
-            ['test.xml', '*/xml', 'xml'],
+            // Disabled for PHP 7.2 RC because of regression:
+            // https://bugs.php.net/bug.php?id=75380
+            version_compare(PHP_VERSION, '7.2.0', '>=') && strpos(PHP_VERSION, 'RC')
```
samdark: Wouldn't it skip all RCs for 7.2, 7.3 etc?

samdark (changes requested):

tests/framework/helpers/HtmlTest.php
SilverFire: No pattern will be generated - exception will be thrown instead.

tests/ResultPrinter.php
SilverFire: Actually, it's not a PHPUnit issue.
PHP have changed `session_cache_limiter()` so it does not work when headers are already sent.
https://github.com/php/php-src/blob/php-7.2.0RC4/UPGRADING#L89

So our tests are broken. Maybe it will be better to mock `session_cache_limiter()` call instead of changing printing output

tests/ResultPrinter.php
SilverFire: Done

SilverFire:
Fixed all remaining tests. Ready for review and merge. @yiisoft/reviewers

cebe (changes requested):

.travis.yml
```diff
@@ -149,6 +148,7 @@ before_script:
 
   # show some versions and env information
   - php --version
+  - php --ini
```
cebe: looks like this breaks HHVM tests, it starts the binary which is waiting for input.

SilverFire:
Thanks, Carsten. Fixed.

> how and why was this working before?

I was surprised as well.

tests/framework/helpers/HtmlTest.php
cebe: I see now. The mock generates a class that does not work as it normally should.

cebe (approved):

framework/validators/UrlValidator.php
```diff
@@ -108,6 +108,16 @@ protected function validateValue($value)
         return [$this->message, []];
     }
 
+    private function idnToAscii($idn)
+    {
+        if (PHP_VERSION_ID < 50600) {
+            // TODO: drop old PHP versions support
```
samdark: 2.1?

SilverFire: Yes

samdark (approved):

tests/framework/helpers/ArrayHelperTest.php
```diff
@@ -810,20 +810,22 @@ public function testGetValueObjects()
 
     /**
      * This is expected to result in a PHP error.
-     * @expectedException \PHPUnit_Framework_Error
+     * @requires PHPUnit 6.0
      */
     public function testGetValueNonexistingProperties1()
     {
+        $this->expectException('PHPUnit\Framework\Error\Notice');
```
rob006: We could use `class_exists()` to detect expected exception and make this test work on phpunit <6.0

SilverFire: I think it's not so important and could be skipped

samdark:
That was big! Thank you for patience. Now we're 7.2 ready! Yay!

bscheshirwork:
We have fix in `php-7.2.0RC5`
[75378](https://github.com/php/php-src/blob/fcbf63d8087db46f2ecd9248687ef2a463013a14/NEWS#L31-L33)

SilverFire:
Already encountered in https://github.com/yiisoft/yii2/commit/336404e0b992406732d80bb239ecf3780dda4359

PR #14963, opened 2017-10-14
Added StringHelper::ensurePrefix

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none

### DIFF
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -366,7 +366,7 @@ class BaseStringHelper
      */
     public static function ensurePrefix($string, $prefix, $caseSensitive = true)
     {
-        if (!self::startsWith($string, $prefix, $caseSensitive)) {
+        if (!static::startsWith($string, $prefix, $caseSensitive)) {
             $string = $prefix . $string;
         }
         return $string;
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -350,4 +350,25 @@ class BaseStringHelper
     {
         return base64_decode(strtr($input, '-_', '+/'));
     }
+
+    /**
+     * Ensures string is prefixed with the prefix specified. Avoids double prefixing.
+     *
+     * Useful for adding "http://" to user entered website addresses, twitter handles etc.
+     *
+     * @param string $string Input string
+     * @param string $prefix Prefix to be ensured in the string
+     * @param bool $caseSensitive Case sensitive comparison. Default is true. When case sensitive is enabled, $prefix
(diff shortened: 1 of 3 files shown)


framework/helpers/BaseStringHelper.php
```diff
@@ -350,4 +350,25 @@ public static function base64UrlDecode($input)
...
+     * @param string $prefix Prefix to be ensured in the string
+     * @param bool $caseSensitive Case sensitive comparison. Default is true. When case sensitive is enabled, $prefix
+     * must exactly match the starting of the string.
+     * @return string
+     *
+     * @since 2.0.13
+     */
+    public static function ensurePrefix($string, $prefix, $caseSensitive = true)
+    {
+        if (!self::startsWith($string, $prefix, $caseSensitive)) {
```
rob006: `self` -> `static` ?

samdark: Yes.

dmirogin (approved):

framework/helpers/BaseStringHelper.php
```diff
@@ -350,4 +350,25 @@ public static function base64UrlDecode($input)
     {
         return base64_decode(strtr($input, '-_', '+/'));
     }
+
+    /**
+     * Ensures string is prefixed with the prefix specified. Avoids double prefixing.
+     *
+     * Useful for adding "http://" to user entered website addresses, twitter handles etc.
```
cebe: how about http vs. https?

framework/helpers/BaseStringHelper.php
```diff
@@ -350,4 +350,25 @@ public static function base64UrlDecode($input)
...
+     * @param bool $caseSensitive Case sensitive comparison. Default is true. When case sensitive is enabled, $prefix
+     * must exactly match the starting of the string.
+     * @return string
+     *
+     * @since 2.0.13
+     */
+    public static function ensurePrefix($string, $prefix, $caseSensitive = true)
+    {
+        if (!static::startsWith($string, $prefix, $caseSensitive)) {
+            $string = $prefix . $string;
```
cebe: no need to modify the string, you can just return the result.

framework/helpers/BaseStringHelper.php
cebe: so its not really useful for user content, what are other use cases?

cebe (changes requested):

samdark:
Closing for the reason @cebe described. Proper handling is more complex than that...

Issue #14877, opened 2017-09-25
Disabled profiling doesn't disable the profiling for connection opening


kashukov:
### What steps will reproduce the problem?

Set 'enableProfiling' to false in DB config.

### What is the expected result?

Profiling is completely disabled.

### What do you get instead?

'Opening DB connection' profiling message in log.

### Additional info

Problem is in ./vendor/yiisoft/yii2/db/Connection.php, lines 578 and 581.
Yii::beginProfile($token, __METHOD__);
Yii::endProfile($token, __METHOD__);

They don't use $profile as in ./vendor/yiisoft/yii2/db/Command.php, line 913.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.1.9
| Operating system | Centos 7

njasm:
something like this would be a solution for this issue?
in vendor/yiisoft/yii2/db/Connection.php
```php
        // ...
        $token = 'Opening DB connection: ' . $this->dsn;
        try {
            Yii::info($token, __METHOD__);
            $this->enableProfiling and Yii::beginProfile($token, __METHOD__);
            $this->pdo = $this->createPdoInstance();
            $this->initConnection();
            $this->enableProfiling and Yii::endProfile($token, __METHOD__);
        } catch (\PDOException $e) {
            $this->enableProfiling and Yii::endProfile($token, __METHOD__);
            throw new Exception($e->getMessage(), $e->errorInfo, (int) $e->getCode(), $e);
        }
```

samdark:
Yes. I'd not use logical `and` side effect for condition checking. It's better to use good old `if`.

cebe:
`enableLogging` is documented as

> whether to enable logging of database queries.

`enableProfiling` is documented as

> whether to enable profiling of database queries.

neither the profile, nor the logging of opening the db connection are database queries. Also these do not have any relevant performance impact so I do not see why we should skip these calls.

njasm:
according to http://www.yiiframework.com/doc-2.0/guide-runtime-logging.html#performance-profiling

> Performance profiling is a special type of message logging that is used to measure the time taken by certain code blocks and find out what are the performance bottlenecks. For example, the yii\db\Command class uses performance profiling to find out the time taken by each DB query.

> To use performance profiling, first identify the code blocks that need to be profiled. Then enclose each code block like the following

[...]

PR #14970, opened 2017-10-14
Disabled profiling on connection opening when profiling is disabled

njasm:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #14877

### DIFF
--- a/framework/db/Connection.php
+++ b/framework/db/Connection.php
@@ -574,21 +574,23 @@ class Connection extends Component
         if (empty($this->dsn)) {
             throw new InvalidConfigException('Connection::dsn cannot be empty.');
         }
+
         $token = 'Opening DB connection: ' . $this->dsn;
+        $enableProfiling = $this->enableProfiling;
         try {
             Yii::info($token, __METHOD__);
-            if ($this->enableProfiling) {
+            if ($enableProfiling) {
                 Yii::beginProfile($token, __METHOD__);
             }
 
             $this->pdo = $this->createPdoInstance();
             $this->initConnection();
 
-            if ($this->enableProfiling) {
+            if ($enableProfiling) {
                 Yii::endProfile($token, __METHOD__);
             }
         } catch (\PDOException $e) {
-            if ($this->enableProfiling) {
+            if ($enableProfiling) {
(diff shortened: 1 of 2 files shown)


samdark (approved):

samdark:

rob006 (approved):
Docs of `enableProfiling` should be updated. See https://github.com/yiisoft/yii2/issues/14877#issuecomment-336674029

samdark:
Yes. @njasm would you please handle docs as well?

njasm:
@samdark  sure but what docs should be updated and with what? are we talking about the DocBock of `yii\db\Connection` ?
```php
    /**
     * @var bool whether to enable profiling of database queries. Defaults to true.
     * You may want to disable this option in a production environment to gain performance
     * if you do not need the information being logged.
     * @since 2.0.12
     * @see enableLogging
     */
```

samdark:
Something like

> whether to enable profiling of opening database connection and database queries

framework/db/Connection.php
```diff
@@ -577,12 +577,21 @@ public function open()
         $token = 'Opening DB connection: ' . $this->dsn;
         try {
             Yii::info($token, __METHOD__);
-            Yii::beginProfile($token, __METHOD__);
+            if ($this->enableProfiling) {
+                Yii::beginProfile($token, __METHOD__);
+            }
+
             $this->pdo = $this->createPdoInstance();
             $this->initConnection();
```
cebe: there is a small chance that someone is using `initConnection` to configure the db connection, i.e. setting `enableProfiling` to another value. To make the code more robust I would get the value `$enableProfiling = $this->enableProfiling` and use the local variable afterwards to keep the nesting of profiling calls.

cebe (changes requested):

framework/db/Connection.php
njasm: sure.. I've missed that. 
done! 
/cc @cebe @samdark @rob006

cebe (approved):

samdark:
Merged. Thanks!

Issue #11401, opened 2016-04-22
DbSession concurrency issue


vikram0207:
### What steps will reproduce the problem?

It worked well with http but with https its making multiple call of writeSession
### What is the expected result?

It should work with https as well
### What do you get instead?

Integrity constraint violation: SQLSTATE[23000]: 
Integrity constraint violation: 1062 Duplicate entry 'h6g8mp45arddisfppio96jj5akkjg9alg8k9kk5f' for key 'PRIMARY'
The SQL being executed was: 
INSERT INTO `session` (`data`, `id`, `expire`) VALUES ('__flash|a:0:{}loginDetailPri|a:1:{s:12:\"permanentKey\";s:64:\"cefbf813b2cf35e9203c53bd24e71b250c74201e5844338b97b7d69af5b127d0\";}loginDetail|a:9:{s:12:\"permanentKey\";s:64:\"0cda4174274e6ca13ec37dd86013dac70779a48210fadce0f40350a4b03b1b1c\";s:7:\"is_bind\";i:1;s:6:\"mobile\";s:12:\"919899195914\";s:5:\"email\";s:0:\"\";s:4:\"name\";s:6:\"Vikram\";s:9:\"last_name\";s:5:\"Singh\";s:3:\"dob\";s:10:\"1981-06-02\";s:7:\"balance\";s:7:\"2575.59\";s:14:\"balance_status\";s:5:\"false\";}otpVerified|s:1:\"0\";otpDetail|a:2:{s:6:\"mobile\";s:10:\"9899195914\";s:8:\"otpValue\";s:1:\"2\";}', 'h6g8mp45arddisfppio96jj5akkjg9alg8k9kk5f1fv9v2qqoef1', 1461308275)

Integrity constraint violation: SQLSTATE[23000]: 
Integrity constraint violation: 1062 Duplicate entry 'h6g8mp45arddisfppio96jj5akkjg9alg8k9kk5f' for key 'PRIMARY'
The SQL being executed was: 

INSERT INTO `session` (`data`, `id`, `expire`) VALUES ('__flash|a:0:{}', 'h6g8mp45arddisfppio96jj5akkjg9alg8k9kk5f1fv9v2qqoef1', 1461308275)
### Additional info

| Q | A |
| --- | --- |
| Yii 2 |  |
| php 5.6.10 |  |
| ubuntu |  |

klimov-paul:
`DbSession` always check for the existance of the session row:
[...]

PR #14972, opened 2017-10-15
Fix for #11401, concurrency when inserting new session ID

andreasanta:
Fix for #11401, concurrency when inserting new session ID

Using master db connection when checking if session exists

### DIFF
--- a/framework/web/DbSession.php
+++ b/framework/web/DbSession.php
@@ -166,12 +166,11 @@ class DbSession extends MultiFieldSession
         // http://us.php.net/manual/en/function.session-set-save-handler.php#refsect1-function.session-set-save-handler-notes
         try {
             // Fix for #11401, concurrency when inserting new session ID            
-            $exists = $this->db->useMaster(function($db) use ($id) {
+            $exists = $this->db->useMaster(function() use ($id) {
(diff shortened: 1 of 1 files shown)


framework/web/DbSession.php
```diff
@@ -165,12 +165,15 @@ public function writeSession($id, $data)
...
-                ->where(['id' => $id])
-                ->createCommand($this->db)
-                ->queryScalar();
+            // Fix for #11401, concurrency when inserting new session ID            
+            $exists = $this->db->useMaster(function($db) use ($id) {
+                return (new Query())->select(['id'])
+                    ->from($this->sessionTable)
+                    ->where(['id' => $id])
+                    ->createCommand($db)
+                    ->queryScalar();
```
rob006: Shouldn't we replace these two lines with `->exists($id);`?

cebe: it should be `->exists($this->db)`

cebe (changes requested):
Same change should apply to `regenerateID()`.

samdark:
Closed in favor of https://github.com/yiisoft/yii2/pull/15624

PR #14980, opened 2017-10-16
Fix looping for Message Formatter tokenize pattern

uaoleg:
If translation string contains open bracket and no close one, than Message Formatter gets into loop.
E.g. `\yii::t('app', 'Hello, {name')`

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


samdark:
Merged. Thank you!

PR #14982, opened 2017-10-17
phpDoc for AccessRule::denyCallback

### LINKED ISSUE #12626: denyCallback in yii\filters\AccessRule is not working as expected

My rule definition

```
                    [
                        'actions' => ['signup'],
                        'allow' => true,
                        'roles' => ['?'],
                        'denyCallback' => function($rule, $action) {
                            return Yii::$app->response->redirect(['user/edit-plan']);
                        }
                    ]
```

I expect that for logged in users the `denyCallback` would be executed.
### What do you get instead?

I get generic HTTP 403 Forbidden error page
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.10-dev |
| PHP version |  |
| Operating system |  |

githubjeka:
Closes #12626

### DIFF
--- a/framework/filters/AccessRule.php
+++ b/framework/filters/AccessRule.php
@@ -56,7 +56,7 @@ class AccessRule extends Component
      *
      * If you are using RBAC (Role-Based Access Control), you may also specify role names.
      * In this case, [[User::can()]] will be called to check access.
-     * 
+     *
      * Note that it is preferred to check for permissions instead.
      *
      * If this property is not set or empty, it means this rule applies regardless of roles.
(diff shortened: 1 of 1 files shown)


framework/filters/AccessRule.php
```diff
@@ -136,8 +136,8 @@ class AccessRule extends Component
     public $matchCallback;
     /**
      * @var callable a callback that will be called if this rule determines the access to
-     * the current action should be denied. If not set, the behavior will be determined by
-     * [[AccessControl]].
+     * the current action should be denied. It's meant then [[$allow]] should be false.
```
samdark: This is less clear than it was...

cebe:
Added more detailed explaination. Thanks!

PR #14983, opened 2017-10-17
avoid use common class like another code

bscheshirwork:
Like 
https://github.com/yiisoft/yii2/blob/0ee17dcb7d763b00819ad764a60e8d582f62f662/framework/base/ArrayAccessTrait.php#L23-L32

Same style is better.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  |

### DIFF
--- a/framework/web/HeaderCollection.php
+++ b/framework/web/HeaderCollection.php
@@ -7,7 +7,6 @@
 
 namespace yii\web;
 
-use ArrayIterator;
 use Yii;
 use yii\base\BaseObject;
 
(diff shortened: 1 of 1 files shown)


samdark:
?

bscheshirwork:
This minor changes
- for equal code
- use global class with \ (1 times new 2 times PHPDoc)

bscheshirwork:
So.. yii2-core-team 
what is the official position
of standard class?
```
use ArrayIterator;
```
or
```
\ArrayIterator
```

samdark:
Merged. Thanks!

cebe:
why not add it in `use`?

rob006:
I'm also for leaving it in `use` block.

samdark:
Using root namespace IMO is better in this case since it hints that it's built-in class.

rob006:
See https://youtrack.jetbrains.com/issue/WI-26767#comment=27-944565

samdark:
Yes. It's about preferences... If you have a strong opinion about it, let's create a vote issue and, when it's done, fix decision in the code style.

rob006:
In my projects I always import everything. It is easier to automate, more DRY and  summary with list of used classes can be really useful. So yes, I have a strong opinion on this, but not strong enough to force this on every project I contribute to. :D

Issue #14633, opened 2017-08-11
Add miliseconds to log time


bologer:
### What steps will reproduce the problem?
This is mostly the feature request. 

When logs are logged now, they are show in this format: 
```
2017-08-11 17:26:42 [][][][] ... 
```

### What is the expected result?

Show logs with miliseconds, such as: 
```
2017-08-11 17:26:42.612315 [][][][] ... 
```

### Maybe this would be useful
```php 
$t = microtime(true);
$micro = sprintf("%06d", ($t - floor($t)) * 1000000);
$d = new \DateTime(date('Y-m-d H:i:s.' . $micro, $t));
$miliseconds = $d->format("u");
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 5.5.6
| Operating system | Ubuntu 14.04

PR #14984, opened 2017-10-17
Add miliseconds to log time

pchapl:
Add property `microtime` to class `\yii\log\Target` for logging time with microseconds

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14633


framework/log/Target.php
```diff
@@ -90,6 +90,12 @@
      */
     public $messages = [];
 
+    /**
+     * @var bool whether to log time with microseconds.
+     * Defaults to false.
```
samdark: Need `@since` tag.

framework/log/Target.php
```diff
@@ -342,4 +348,17 @@ public function getEnabled()
 
         return $this->_enabled;
     }
+
+    /**
+     * Returns formatted ('Y-m-d H:i:s') timestamp for message.
+     * If [[microtime]] is configured to true it will return format 'Y-m-d H:i:s.u'
+     * @param float $timestamp
+     * @return string
```
samdark: Need `@since` tag.

framework/log/Target.php
```diff
@@ -342,4 +348,17 @@ public function getEnabled()
...
+     */
+    protected function getTime($timestamp)
+    {
+        list ($timestamp, $usec) = explode('.', (string)$timestamp);
```
samdark: What's `usec`?

samdark: I think it should be named millisecond.

samdark (approved):
Looks good overall. Need a changelog entry and since tags.

dmirogin (approved):

framework/log/Target.php
```diff
@@ -342,4 +348,17 @@ public function getEnabled()
...
+
+        return date('Y-m-d H:i:s', $timestamp) . ($this->microtime ? ('.' . $usec) : '');
```
dmirogin: LGTM, but i'd write 
```php
$format = $this->microtime ? 'Y-m-d H:i:s.u' : 'Y-m-d H:i:s';
return date($format);
```

Ni-san: Me too, but `date` will not work as expected:

"u - Note that date() will always generate 000000 since it takes an integer parameter"

http://php.net/manual/en/function.date.php

framework/log/Target.php
Ni-san: `usec` is `μsec`. For example, function `gettimeofday` returns array with keys "sec" and "usec".

samdark:
Is the failure in PHP7 travis build related to these changes? https://travis-ci.org/yiisoft/yii2/jobs/289319551

rob006 (approved):

rob006:
@samdark It seems to be unrelated - I restarted tests and they're pass.

samdark:
Merged. Thank you!

PR #14988, opened 2017-10-19
fix guide-ru rest-resources

bscheshirwork:
Some changes - text is compared to code at example below
| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | no
| Breaks BC?    | no


samdark:
Is it an issue in Russian version only or English is wrong as well?

bscheshirwork:
@samdark I think this problem is only a translation problem. 
"
The array keys are the field names, and the array values are the corresponding field definitions which can be either property/attribute names or anonymous functions returning the corresponding field values. In the special case when a field name is the same as its defining attribute name, you can omit the array key.
"
It seems to me that it directly write that the "key" is omitted in the context of the "array key"

samdark:
Merged. Thank you!

PR #14996, opened 2017-10-20
Add rest\GroupUrlRule; add rest\UrlRule param controllerPrefix

bscheshirwork:
For more "versioned" of config of `RESTful` `api` application

All of rules of major version is one rest\GroupUrlRule;
In nested rules `module` prefix is omitted;

Current
```php
return [
    'modules' => [
        'v1' => [
            'class' => 'app\modules\v1\Module',
        ],
        'v2' => [
            'class' => 'app\modules\v2\Module',
        ],
    ],
    'components' => [
        'urlManager' => [
            'enablePrettyUrl' => true,
            'enableStrictParsing' => true,
            'showScriptName' => false,
            'rules' => [
                ['class' => 'yii\rest\UrlRule', 'controller' => ['v1/user', 'v1/post']],
                ['class' => 'yii\rest\UrlRule', 'controller' => ['v2/user', 'v2/post']],
            ],
        ],
    ],
];
```
propose
```php
return [
    'modules' => [
        'v1' => [
            'class' => 'app\modules\v1\Module',
        ],
        'v2' => [
            'class' => 'app\modules\v2\Module',
        ],
    ],
    'components' => [
        'urlManager' => [
            'enablePrettyUrl' => true,
            'enableStrictParsing' => true,
            'showScriptName' => false,
            'rules' => [
                // v1
                [
                    'class' => 'yii\rest\GroupUrlRule',
                    'prefix' => 'v1',
                    'rules' => [
                        [
                            'controller' => 'post',
                            'pluralize' => false,
                            'only' => ['index', 'view', 'options'],
                        ],
                       'user',
                        //another yii\rest\UrlRule of v1 module with differences in params
                    ],
                ],
                // v2
                [
                    'class' => 'yii\rest\GroupUrlRule',
                    'prefix' => 'v2',
                    'rules' => [
                        [
                            'controller' => 'post',
                            'pluralize' => false,
                            'only' => ['index', 'view', 'options'],
                        ],
                        'u' => 'user',
                        //another yii\rest\UrlRule of v2 module with differences in params
                    ],
                ],
                //...
            ],
        ],
    ],
];
```

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/framework/rest/GroupUrlRule.php
+++ b/framework/rest/GroupUrlRule.php
@@ -1,0 +1,164 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\rest;
+
+use Yii;
+use yii\base\InvalidConfigException;
+use yii\web\CompositeUrlRule;
+use yii\web\UrlRule as WebUrlRule;
+use yii\web\UrlRuleInterface;
+
+/**
+ * GroupUrlRule represents a collection of URL rules sharing the same prefix in their patterns and routes.
+ *
+ * GroupUrlRule is best used by a module which often uses module ID as the prefix for the URL rules.
+ * For example, the following code creates a rule for the `admin` module:
+ *
+ * ```php
+ * new GroupUrlRule([
+ *     'prefix' => 'v1',
+ *     'rules' => [
+ *         'user',
+ *         [
+ *             'controller' => ['customer', 'post'],
+ *             'pluralize' => false,
+ *             'only' => ['view'],
+ *         ],
+ *     ],
+ * ]);
+ * ```
+ *
+ * The above code will create a whole set of URL rules supporting the following RESTful API endpoints:
+ *
+ * - `'PUT,PATCH v1/users/<id>' => 'v1/user/update'`: update a user
+ * - `'DELETE v1/users/<id>' => 'v1/user/delete'`: delete a user
+ * - `'GET,HEAD v1/users/<id>' => 'v1/user/view'`: return the details/overview/options of a user
+ * - `'POST v1/users' => 'v1/user/create'`: create a new user
+ * - `'GET,HEAD v1/users' => 'v1/user/index'`: return a list/overview/options of users
+ * - `'v1/users/<id>' => 'v1/user/options'`: process all unhandled verbs of a user
+ * - `'v1/users' => 'v1/user/options'`: process all unhandled verbs of user collection
+ * - `'GET,HEAD v1/customer/<id>' => 'v1/customer/view'`: return the details/overview/options of a customer
+ * - `'GET,HEAD v1/post/<id>' => 'v1/post/view'`: return the details/overview/options of a post
+ *
+ * The above example assumes the prefix for patterns and routes are the same. They can be made different
+ * by configuring [[prefix]] and [[routePrefix]] separately.
+ * Note: [[prefix]] can not be resolved to [[yii\rest\UrlRule::prefix]];
+ * [[routePrefix]] can be resolved to [[yii\rest\UrlRule::controllerPrefix]]
+ *
+ * Using a GroupUrlRule is more efficient than directly declaring the individual rules it contains.
+ * This is because GroupUrlRule can quickly determine if it should process a URL parsing or creation request
+ * by simply checking if the prefix matches.
+ *
+ * @author Qiang Xue <qiang.xue@gmail.com>
+ * @author BSCheshir <bscheshir.work@gmail.com>
+ * @since 2.0.13
+ */
+class GroupUrlRule extends CompositeUrlRule
+{
+    /**
+     * @var array the rules contained within this composite rule. Please refer to [[UrlManager::rules]]
+     * for the format of this property.
+     *
+     * A special shortcut format can be used if a rule only specifies [[\yii\rest\UrlRule::controller]]
+     * That is, instead of using a configuration array, one can use an array with the array key
+     * being as the controller ID in the pattern, and the array value the actual controller ID.
+     * For example, `'rules' => ['u' => 'user']`. You can omit the array key `'rules' => ['user']`
+     * To use [[\yii\rest\UrlRule::controller]] as default, the controller ID will be pluralized
+     * automatically when it is put in the patterns of the generated rules.
+     *
+     * @see prefix
+     * @see controllerPrefix
+     * @see \yii\rest\UrlRule::controller
+     */
+    public $rules = [];
+    /**
+     * @var string the prefix for the pattern part of every rule declared in [[rules]].
+     * The prefix and the pattern will be separated with a slash.
+     */
+    public $prefix;
+    /**
+     * @var string the prefix for the controller part of every rule declared in [[rules]].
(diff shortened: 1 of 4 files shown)


bscheshirwork:
huh, I get some error in tests with dataProvider config %)

rob006:
Is it really necessary to create another group rule class? We could adjust `yii\web\GroupUrlRule` to work properly with rest rules. #9122

bscheshirwork:
@rob006 as you can see in code this class is a sugar for module name. 

`yii\web\GroupUrlRule` not implement `module prefix` for **controller** property of nested `yii\rest\UrlRule`;

rob006:
> yii\web\GroupUrlRule not implement module prefix for controller property of nested yii\rest\UrlRule;

It has `$prefix` and `$routePrefix` which is basically the same. You just changed name of these properties. Why not use `$routePrefix` and `$prefix` directly in `$ruleConfig`?

bscheshirwork:
@rob006 yes of course.

2 differences:

- first  
  `$routePrefix` property is not implemented in `yii\rbac\UrlRule` and in `yii\web\CompositeUrlRule`

- second  
  additional "short form" can't be use.


For first:
I propose add the `$controllerPrefix` property to `yii\rbac\UrlRule`. This property is same at `routePrefix` in another rule, but is named according to meaning.

This property unbind of `$prefix` by BC reason (unlike $routePrefix = $routePrefix ?: $route).  

For this reason the `yii\rbac\GroupUrlRule` process `prefix` and `controllerPrefix` and pass it both.

bscheshirwork:
And, before you can ask...

If we will add only `contrllerPrefix` property in `yii\rest\UrlRule`

```php
        'urlManager' => [
            'enablePrettyUrl' => true,
            'enableStrictParsing' => true,
            'showScriptName' => false,
            'rules' => [
                // v1 use 'yii\web\GroupUrlRule' and property `controllerPrefix`
                [
                    'class' => 'yii\web\GroupUrlRule',
                    'prefix' => 'v1',
                    'ruleConfig' =>  [
                        'prefix' => 'v1',
                        'controllerPrefix' => 'v1',
                    ],
                    'rules' => [
                        [
                            'controller' => 'post',
                            'pluralize' => false,
                            'only' => ['index', 'view', 'options'],
                        ],
                       ['controller' => 'user'],
                        //another yii\rest\UrlRule of v1 module with differences in params
                    ],
                ],
                // v2 use 'yii\rest\GroupUrlRule' use both property `controllerPrefix` and class `yii\rest\GroupUrlRule`
                [
                    'class' => 'yii\rest\GroupUrlRule',
                    'prefix' => 'v2',                    
                    'rules' => [
                        [
                            'controller' => 'post',
                            'pluralize' => false,
                            'only' => ['index', 'view', 'options'],
                        ],
                        'u' => 'user', // example of short syntax of nested rule controller property
                        //another yii\rest\UrlRule of v2 module with differences in params
                    ],
                ],
                //...
            ],

```

bscheshirwork:
related #13968

rob006:
> $routePrefix property is not implemented in yii\rbac\UrlRule and in yii\web\CompositeUrlRule

You could implement it in the same way as `$controllerPrefix`. You will get unified API and less duplicated code and redundant classes.

bscheshirwork:
@rob006 `$controllerPrefix` uses instead of `$routePrefix`
because logic of applied `$prefix` to empty `$controllerPrefix` in `yii\rest\UrlRule` and to empty `$routePrefix` in `yii\web\GroupUrlRule` is difference. 
```php
        $this->prefix = trim($this->prefix, '/');
        $this->routePrefix = $this->routePrefix === null ? $this->prefix : trim($this->routePrefix, '/');
```
```php
        $this->prefix = trim($this->prefix, '/');
        $this->controllerPrefix = trim($this->controllerPrefix, '/');
```

You propose use `$routePrefix` in  `yii\rest\UrlRule` instead of `$controllerPrefix` **anyway**?

rob006:
> because logic of applied $prefix to empty $controllerPrefix in yii\rest\UrlRule and to empty $routePrefix in yii\web\GroupUrlRule is difference.

But why? Sorry, I really don't get what is the problem here and why `$controllerPrefix` must be different from `$routePrefix`.

bscheshirwork:
@rob006 just BC compatibility

I am a developer.
I use yii2 for RESTful api
I have a worked `api`
I have `config`
```php
return [
    'modules' => [
        'v1' => [
            'class' => 'app\modules\v1\Module',
        ],
        'v2' => [
            'class' => 'app\modules\v2\Module',
        ],
    ],
    'components' => [
        'urlManager' => [
            'enablePrettyUrl' => true,
            'enableStrictParsing' => true,
            'showScriptName' => false,
            'rules' => [
                ['class' => 'yii\rest\UrlRule', 'prefix' => 'api/v1', 'controller' => ['v1/user', 'v1/post']],
                ['class' => 'yii\rest\UrlRule', 'prefix' => 'api/v2', 'controller' => ['v2/user', 'v2/post']],
                // ... or any else - see tests case below
                ['class' => 'yii\rest\UrlRule', 'prefix' => 'admin', 'controller' => 'post'], 
            ],
        ],
    ],
];
```
https://github.com/yiisoft/yii2/blob/77ad3dca3f67069efd32c9f3b29b0f10a83c0c8c/tests/framework/rest/UrlRuleTest.php#L84-L90
```
'api/v1/posts', 'v1/post/index'
...
'admin/posts', 'admin/post/index'
```

I (we, contributors of framework) add property to `yii\rest\UrlRule`, property is prefix of controller;

1. With relation of `prefix` (like `yii\web\GroupUrlRule`)
I have `init()`
```php
        $this->prefix = trim($this->prefix, '/');
        $this->routePrefix = $this->routePrefix === null ? $this->prefix : trim($this->routePrefix, '/');
```
I have wrong resolve 
```
'api/v1/posts', 'api/v1/v1/post/index'
...
'admin/posts', 'admin/post/index'
```

2. Unbind / independent properties

I have `init()` 
```php
        $this->prefix = trim($this->prefix, '/');
        $this->controllerPrefix = trim($this->controllerPrefix, '/');
```
I have my code, who still work

rob006:
In your example you're not using `GroupUrlRule`, so this will not change anything. AFAIK `GroupUrlRule` doesn't work well with `yii\rest\UrlRule` so you can't really break BC for thing that doesn't work. :D

But event if I'm wrong and this will break BC, I would rather move this to 2.1 than introduce `yii\rest\GroupUrlRule` which do exactly the same thing as `yii\web\GroupUrlRule`, but in different way (for no reason) and with different names of properties. We should keep API simple and introducing two different interfaces for the same job is confusing and harder to maintain.

bscheshirwork:
> which do exactly the same thing as

Not same. 
`yii\web\GroupUrlRule` set `prefix` and `routePrefix` for self

`yii\rest\GroupUrlRule` set prefix and `controllerPrefix` (aka `routePrefix`) for self
AND propose sugar for set `controller` property
AND set another default class for nested rules.
AND set `controllerPrefix` (aka `routePrefix`) for nested rules


I will change  `controllerPrefix` to `routePrefix` without `null-fill` logic and add todo for BC broken changes in 2.1

Also I will split changes into 2 PR.

But it is not same.

samdark:
Won't be merged into 2.0.

PR #14998, opened 2017-10-20
fix same variable name

bscheshirwork:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | binding to  #14996

### DIFF
--- a/tests/framework/rest/UrlRuleTest.php
+++ b/tests/framework/rest/UrlRuleTest.php
@@ -347,10 +347,10 @@ class UrlRuleTest extends TestCase
 
     /**
      * @dataProvider createUrlDataProvider
-     * @param array $rule
+     * @param array $config
      * @param array $tests
      */
-    public function testCreateUrl($rule, $tests)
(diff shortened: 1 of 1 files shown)


samdark:
What's the problem it is fixing?

bscheshirwork:
@samdark test is broken: attemp to config each non-first object `$rule` from previous object `$rule` (foreach), not from input array `$rule`

cebe (approved):
ideally the name should be `$ruleConfig`

rob006 (approved):

bscheshirwork:
@cebe yes, IMHO you right.  
In my opinion `ruleConfig` is better  
But all other equivalent variables in this file has named "config"

samdark:
Adjusted and merged https://github.com/yiisoft/yii2/commit/eac6a5d6af894606b8b7bdf351594dc451e4b246. Thanks.

PR #15001, opened 2017-10-21
Fixes #5515

### LINKED ISSUE #5515: blameable behavior only for non-guests
Hi,
I've noticed that BlameableBehavior updates the updated_by field when a user is requesting a password reset.  As a result, updated_by becomes NULL as this is an un-recognised user (isGuest).
I'm thinking that BlameableBehavior should only execute if user is !isGuest() .
Not sure if there other instances where BlameableBehavior can be influenced by isGuest, I'm new to yii in general, so starting with yii2.

dmirogin:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #5515


cebe:
ideally the behavior should be disabled when user is guest to not override a value set from a real user.

framework/behaviors/BlameableBehavior.php
```diff
@@ -73,6 +73,15 @@ class BlameableBehavior extends AttributeBehavior
...
 
+    /**
+     * @var mixed Default value for cases when user is guest
+     */
+    public $defaultValue;
+
+    /**
+     * @var bool Whether is [[defaultValue]] enabled
+     */
+    public $useDefaultValue = false;
```
rob006: This property is redundant. `$defaultValue` is `null` by default, same as `Yii::$app->get('user')->id` for guests. We don't need separate property to decide which `null` should we use.

framework/behaviors/BlameableBehavior.php
```diff
@@ -97,7 +106,12 @@ public function init()
     protected function getValue($event)
     {
         if ($this->value === null && Yii::$app->has('user')) {
-            return Yii::$app->get('user')->id;
+            $userId = Yii::$app->get('user')->id;
+            if (is_null($userId) && $this->useDefaultValue) {
```
rob006: `is_null($userId)` => `$userId !== null`

framework/behaviors/BlameableBehavior.php
```diff
@@ -73,6 +73,15 @@ class BlameableBehavior extends AttributeBehavior
      */
     public $value;
 
+    /**
+     * @var mixed Default value for cases when user is guest
+     */
+    public $defaultValue;
```
rob006: Missing `@since`.

rob006 (changes requested):

framework/behaviors/BlameableBehavior.php
dmirogin: @rob006 http://php.net/manual/ru/function.is-null.php#87355

cebe: that comment does not include `===`. behaviour equal and `===` is consistent with other yii code,

dmirogin:
I understood completely this case, i'll finish in this week.

samdark:
@dmirogin any update?

dmirogin:
Sorry, was busy, going to finish in this weekend, maybe earlier.

dmirogin:
i've also noticed a lot of similar constructions for determining closures, like
~~~php
if ($value instanceof \Closure || (is_array($value) && is_callable($value))) {
     return call_user_func($value);
}
~~~

Would be great to add this function inside some helper

dmirogin:
@samdark Could you take a look?

samdark (approved):

rob006 (approved):

sergeymakinen (approved):

samdark:
Merged. Thank you!

PR #15002, opened 2017-10-21
Fixes #14543

### LINKED ISSUE #14543: default generated migration table's 'version' column data length not big enough
the default generated migration table with its "version" column set to varchar(180), but when we create a long alter table column migration,  it will drop some characters, such as the following one:
m170726_070258_add_reqm_source_column_temp_theme_column_baselib_config_column_deploy_date_column_pass_review_column_pass_test_column_rollback_time_column_regular_check_column_to_task
right migration apply result
### What do you get instead?
migration want to re-apply old ones

workaround:
 issue SQL "alter table migration modify version varchar(360);"

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 5.6.30
| Operating system |Mac os 10.9.5

dmirogin:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14543

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -622,8 +622,8 @@ abstract class BaseMigrateController extends Controller
         }
 
         list($namespace, $className) = $this->generateClassName($name);
-        // Check that class name less then 251 (255 - strlen('.php'))
-        if (strlen($className) > 251) {
+        // Abort if name is too long
+        if (strlen($className) > 180) {
             throw new Exception('The migration name is too long.');
         }
 
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -32,6 +32,11 @@ abstract class BaseMigrateController extends Controller
     const BASE_MIGRATION = 'm000000_000000_base';
 
     /**
+     * Maximum length of migration name
+     */
+    const MAX_NAME_LENGTH = 180;
+
+    /**
      * @var string the default command action.
      */
     public $defaultAction = 'up';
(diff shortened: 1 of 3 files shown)


framework/console/controllers/BaseMigrateController.php
```diff
@@ -622,6 +622,12 @@ public function actionCreate($name)
         }
 
         list($namespace, $className) = $this->generateClassName($name);
+        // Check that class name less then 251 (255 - strlen('.php'))
+        if (strlen($className) > 251) {
+            $className = substr($className, 0, 251);
+            $this->stdout('Length of the migration\'s class name was truncated', Console::FG_YELLOW);
```
cebe: Its probably better to error and let the user specify a new name. Using a truncated name does not have much value imo.

cebe (changes requested):

framework/console/controllers/BaseMigrateController.php
samdark: Agree.

framework/console/controllers/BaseMigrateController.php
```diff
@@ -622,6 +622,11 @@ public function actionCreate($name)
         }
 
         list($namespace, $className) = $this->generateClassName($name);
+        // Check that class name less then 251 (255 - strlen('.php'))
+        if (strlen($className) > 251) {
```
rob006: It should be 180:

https://github.com/yiisoft/yii2/blob/980cfde0f1720a9f8516f76d37afa5a5528882ca/framework/console/controllers/MigrateController.php#L260-L263

rob006 (changes requested):

rob006:
I don't like using hardcoded limit - this should be configurable. Or better - length should be dynamically detected from migration table schema. 

This should be also fixed for `migrate/up` - I can manually create migration with too long name.

dmirogin:
Cause base migration independent on database, i created constant.

dmirogin:
Cause base migration independent on database, i created constant.

rob006:
You could create method and override it in `MigrationController`.

samdark (approved):

SilverFire (approved):

SilverFire:
Could be merged in 2.0.13

samdark:
Merged. Thanks!

PR #15003, opened 2017-10-22
Added setting forwardUrlParameters to UrlManager

uldisn:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | no

### DIFF
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -150,6 +150,15 @@ class UrlManager extends Component
     public $normalizer = false;
 
     /**
+     * define URL forward parameters
+     *
+     * if new URL params no forward parameter, get it from request url and add  to new URL.
+     *
+     * @var array
+     */
+    public $forwardGetParams = [];	
+	
+    /**
      * @var string the cache key for cached rules
      * @since 2.0.8
      */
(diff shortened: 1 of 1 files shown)


uldisn:
Use case:
For example, when open Fancybox, for navigation in it, add to url fancybox=1. 
This parameter use for determining correct layout.
When open any link in Fnacybox popup, parameter is forwarder and can use correct layout.

rob006:
I don't think that this use case is common enough to be in core. You can achieve the same by overriding `createUrl()` in custom `UrlManager`.

uldisn:
Use case 2:
For providing return URL through more one page.

samdark:
> For providing return URL through more one page.

Haven't got it. Would you please elaborate?

Renkas:
I use similar parameter forwarding for iframe content (same content/view can be in iframe or in normal layout).

This PR would help me clean up some ugly code from controllers

samdark:
@Renkas I still don't get the use case. Would you please help us by elaborating it?

Renkas:
@samdark I have many (10+) admin views that work in normal ways or in a iframe (with autoresize for height). As there is no way (AFAIK) to know if link is opened in iframe or in normal admin layout I use url parameter to differentiate the layout (iframe=1).

Actually this PR would not allow me to clean up anything from controllers - I remembered it incorrectly.

I have overriden `UrlManager` myself for this.

```
namespace common\components;

class UrlManager extends \yii\web\UrlManager
{
    /**
     * @var string the GET parameter name for iframe layout request.
     */
    public $iframeParam = 'iframe';
    /**
     * @var boolean if the request is done from within iframe
     */
    private $iframe = false;

    public function parseRequest($request)
    {
        if ($request->getQueryParam($this->iframeParam)) {
            $this->iframe = true;
        }

        return parent::parseRequest($request);
    }

    public function createUrl($params)
    {
        if ($this->iframe) {
            $params[$this->iframeParam] = 1;
        }

        return parent::createUrl($params);
    }
}
```

IMHO would be a nice to have thing in core. But it is also easy to override `UrlManager` yourself

samdark:
@Renkas so your case isn't related. OK.

Renkas:
@samdark AFAIK it is exactly the use case this PR resolves.

Just I have my own implementation for it already.

samdark:
Won't be merged because it's not that common. Sorry.

PR #15009, opened 2017-10-23
Added exceptSelect function for Query

ghost:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -597,6 +597,30 @@ PATTERN;
     }
 
     /**
+     * Add columns of $realTable table to the SELECT part of the query, except $exceptColumns
+     *
+     * @param string $asTable
+     * @param string $realTable
+     * @param string|array $exceptColumns
+     * @return $this the query object itself
+     * @throws InvalidConfigException
+     */
+    public function exceptSelect($asTable, $realTable, $exceptColumns)
+    {
+        $tableSchema = Yii::$app->db->getSchema()->getTableSchema($realTable);
+        if ($tableSchema === null) {
+            throw new InvalidConfigException('The table does not exist: ' . $realTable);
+        }
+        $col = $tableSchema->getColumnNames();
+        $exceptColumns = is_array($exceptColumns) ? $exceptColumns : [$exceptColumns];
+        foreach($exceptColumns as $v) {
+            unset($col[array_keys($col, $v)[0]]);
+        }
+        $this->select = array_merge($this->select, array_map(function($v) use($asTable) {return $asTable.'.'.$v;}, $col));
+        return $this;
+    }
+
+    /**
      * Sets the value indicating whether to SELECT DISTINCT or not.
      * @param bool $value whether to SELECT DISTINCT or not.
      * @return $this the query object itself
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -599,9 +599,13 @@ PATTERN;
     /**
      * Add columns of $realTable table to the SELECT part of the query, except $exceptColumns
      *
-     * @param string $asTable
-     * @param string $realTable
-     * @param string|array $exceptColumns
+     * ```php
+     * $query->exceptSelect(["crowd", "crowdfunding", ["title", "content"])->one();
+     * ```
+     *
+     * @param string $asTable a table name will be rename use "as" in SQL
+     * @param string $realTable a real table name
+     * @param string|array $exceptColumns columns will be remove from SELECT
      * @return $this the query object itself
      * @throws InvalidConfigException
      */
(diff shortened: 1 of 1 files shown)


samdark:
I've got it. The case is a table with 100s of columns and the need to select everything except "status", "is_deleted".

samdark:
Personally I both like and don't like this addition. I've been in situation like that and wished there was a method like that but that was only once in like 5 years.

samdark:
@yiisoft/core-developers, @yiisoft/reviewers opinions?

klimov-paul:
I do not consider use case to be common enough. If it would SQL native syntax suported it long ago.
Having a database table with so much columns, so you need `exceptSelect` feature, usually indicates a poor database structure design.
Also such functionality can be easily composed into an `ActiveQuery` behavior or trait and be applied at will. Also building a list of fields can be moved into a separated helper.

Inocence:
I thought about you said, if I build a list , i must update it when i changed sql structure,then, every table just have 20 columns,but i meet this problem in many projects,every time i must write same code to resolve it

framework/db/Query.php
```diff
@@ -597,6 +597,35 @@ public function addSelect($columns)
...
+        $tableSchema = Yii::$app->db->getSchema()->getTableSchema($realTable);
+        if ($tableSchema === null) {
+            throw new InvalidConfigException('The table does not exist: ' . $realTable);
+        }
+        $col = $tableSchema->getColumnNames();
+        $exceptColumns = is_array($exceptColumns) ? $exceptColumns : preg_split('/\s*,\s*/', trim($exceptColumns), -1, PREG_SPLIT_NO_EMPTY);
+        foreach($exceptColumns as $v) {
+            unset($col[array_keys($col, $v)[0]]);
+        }
+        $this->select = array_merge($this->select, array_map(function($v) use($asTable) {return $asTable.'.'.$v;}, $col));
```
bscheshirwork: $v is tooo short (codeclimate min length is 3) and not relate to `column name`

framework/db/Query.php
```diff
@@ -597,6 +597,35 @@ public function addSelect($columns)
...
+     */
+    public function exceptSelect($asTable, $realTable, $exceptColumns)
+    {
+        $tableSchema = Yii::$app->db->getSchema()->getTableSchema($realTable);
+        if ($tableSchema === null) {
+            throw new InvalidConfigException('The table does not exist: ' . $realTable);
+        }
+        $col = $tableSchema->getColumnNames();
+        $exceptColumns = is_array($exceptColumns) ? $exceptColumns : preg_split('/\s*,\s*/', trim($exceptColumns), -1, PREG_SPLIT_NO_EMPTY);
+        foreach($exceptColumns as $v) {
```
bscheshirwork: `$v` is too short (codeclimate), we can use `$value` instead

Inocence:
And,maybe it is not very common,but I think it is necessary. Anyway I really hope you can accept this feature ,so I can never write same code again,thanks

Inocence:
Ok,thank you to review the code,i will rename variables tomorrow

framework/db/Query.php
Inocence: Ok,thank you to review the code,i will rename variables tomorrow

bscheshirwork:
@Inocence 
> I thought about you said, if I build a list , i must update it when i changed sql structure,then, every table just have 20 columns,but i meet this problem in many projects,every time i must write same code to resolve it

> And,maybe it is not very common,but I think it is necessary. Anyway I really hope you can accept this feature ,so I can never write same code again,thanks

I think your can use your own repo with this behaviour just now and use it in all projects. 
For example: I have repo [bscheshirwork/yii2-cubs](https://github.com/bscheshirwork/yii2-cubs)

Inocence:
Ok, I got it,thanks.

cebe:
We currently have `addSelect()` to add columns to the select clause, maybe we can have a `removeSelect()` that works complementary to that.

bscheshirwork:
https://github.com/yiisoft/yii2/blob/376006a2d5048764ebad70a3fd1077ceb4989352/framework/db/Query.php#L568-L597

Have only one parameter `$columns` and independent of table structure or database driver.  
Revere of this will expect same signature. 

i.e. something like 
```php
public function removeSelect($columns, ?array $alias = null)
```

Can I will use `removeSelect` for query of `\yii\redis\ActiveRecord` model?

bscheshirwork:
Additional expects 
```php
$query
->addSelect(["*", "CONCAT(first_name, ' ', last_name) AS full_name"])
->removeSelect(["*", "CONCAT(first_name, ' ', last_name) AS full_name"])
```
still `$query` is unchanged

Inocence:
i think i should change the function to make it works like `addSelect()`

samdark:
I like what @cebe proposed.

samdark:
Won't make it into 2.0.

PR #15010, opened 2017-10-23
Docs: example for concrete file alias

bscheshirwork:
Add little bit concrete example to define non-simple (but root) alias, who resolve path to concrete file.

This will be helpful to redefine autoload path of non-package class or something else :angel:


SilverFire (approved):

samdark:
Merged. Thanks!

Issue #14934, opened 2017-10-05
Document REST filtering


samdark:
https://github.com/yiisoft/yii2/pull/12641 should be documented.

PR #15013, opened 2017-10-24
Added filter docs

samdark:
Added some docs. Please check if these are enough to get started with filters. While putting it together some questions surfaced:

1. Data filters overall aren't bound to REST so probably it's a good idea to move them under `data` namespace. @klimov-paul have you considered it?
2. Not sure it's a good idea to enable filters in REST `IndexAction` by default. My concern is that it could be used for quite complicated conditions that could be exploited to hurt API performance.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12641, #14934


samdark:
@yiisoft/reviewers would be good if you'll take a look. Thanks.

dynasource (approved):

docs/guide/rest-resources.md
```diff
@@ -244,4 +244,41 @@ will also include the pagination information by the following HTTP headers:
...
+
+```php
+$filter = new ActiveDataFilter([
+    'searchModel' => 'app\models\PostSearch'
+]);
+
+$filterCondition = null;
+
+// if you prefer JSON in request body, use Yii::$app->request->getBodyParams() below:
+if ($filter->load(\Yii::$app->request->get())) {
```
bscheshirwork: With `yii\base\Model` and `yii\web\Request` we can use `post(null, null)` and `load(... , '')`
```php
$filter->load(\Yii::$app->request->post(), '')
```
for `json` OR form `post` data...

docs/guide/rest-quick-start.md
```diff
@@ -185,7 +186,11 @@ For example, the URL `http://localhost/users?fields=id,email` will only return t
 
 > Info: You may have noticed that the result of `http://localhost/users` includes some sensitive fields,
 > such as `password_hash`, `auth_key`. You certainly do not want these to appear in your API result.
-> You can and should filter out these fields as described in the [Resources](rest-resources.md) section.
+> You can and should remove these fields from result as described in the [Resources](rest-resources.md) section.
+
+Addionally, you can filter collections like `http://localhost/users?filter[id]=10` or
+`http://localhost/users?filter[email][like]=gmail.com` and sort collections like
+`http://localhost/users?sort=email` or `http://localhost/users?sort=-email`.
```
cebe: should have a link to further reading here, link to `rest-resources.md#filtering-collections`

docs/guide/rest-resources.md
samdark: Of course. Data source could be anything.

samdark: Do you think it should be mentioned specifically?

docs/guide/rest-resources.md
```diff
@@ -244,4 +244,41 @@ will also include the pagination information by the following HTTP headers:
...
+
+### Filtering collections <span id="filtering-collections"></span>
+
+Since version 2.0.13 Yii provides a facility to filter collections. In case you are using `ActiveController` you can enjoy it
+right away. An example can be found in the [Quick Start](rest-quick-start.md#trying-it-out) section. In case you're
+implementing an endpoint yourself, filtering could be done like the following:
+
+```php
+$filter = new ActiveDataFilter([
+    'searchModel' => 'app\models\PostSearch'
```
cebe: how does this model look like, there is no standard for this as far as I see.

cebe (changes requested):

cebe:
> Not sure it's a good idea to enable filters in REST IndexAction by default.

does this enable filtering in existing APIs? should not be done imo.

docs/guide/rest-resources.md
bscheshirwork: > // if you prefer JSON in request body, use Yii::$app->request->getBodyParams() below:
This phrase is so concrete. I think we need weaker mention. 
Or more examples...

docs/guide/rest-resources.md
samdark: It's a regular form model. Just public properties and rules. I'll add an example.

samdark:
> does this enable filtering in existing APIs? should not be done imo.

Yes, it does if the API is built around `ActiveController`.

samdark:
@cebe, @dynasource what's your opinion on points 1 and 2 from the PR description?

docs/guide/rest-resources.md
bscheshirwork: in base PR I can see DynamicModel and anonymous function... 
```php
    public function actions()
    {
        return [
            'index' => [
                'class' => 'yii\rest\IndexAction',
                'modelClass' => $this->modelClass,
                'dataFilter' => [
                    'class' => 'yii\rest\ActiveDataFilter',
                    'searchModel' => function () {
                        return (new \yii\base\DynamicModel(['id' => null, 'name' => null]))
                            ->addRule('id', 'integer')
                            ->addRule('name', 'trim')
                            ->addRule('name', 'string');
                    },
                ],
            ],
...
        ];
    }

```
Can process anonymous class `new class()` instead?

dynasource:
@samdark, I tend to agree:
1) it seems generic enough to be put in the data namespace
2) it's already off by default?

docs/guide/rest-resources.md
bscheshirwork: Hm, Interesting realization of `yii\rest\DataFilter::load()`
```php
    /**
     * @inheritdoc
     */
    public function attributes()
    {
        return [
            $this->filterAttributeName
        ];
    }

    /**
     * @inheritdoc
     */
    public function formName()
    {
        return '';
    }
```
@samdark My be we need add link to this in comment?

@see yii\rest\DataFilter::load()
@see yii\rest\DataFilter::attributes()
@see yii\rest\DataFilter::filterAttributeName

cebe:
> Data filters overall aren't bound to REST so probably it's a good idea to move them under data namespace. @klimov-paul have you considered it?

from a quick look into the implementation it seems to me putting it into `yii\data` makes sense.

> Not sure it's a good idea to enable filters in REST IndexAction by default. My concern is that it could be used for quite complicated conditions that could be exploited to hurt API performance.

its not enabled by default as far as I see:

> https://github.com/yiisoft/yii2/blob/ce65e24d1988f00e57ceaa7493632ee4f3aca776/framework/rest/IndexAction.php#L49
> You must setup this field explicitly in order to enable filter processing.

docs/guide/rest-resources.md
```diff
@@ -244,4 +244,64 @@ will also include the pagination information by the following HTTP headers:
...
+]);
+
+$filterCondition = null;
+
+// You may load filters from any source. For example,
+// if you prefer JSON in request body,
+// use Yii::$app->request->getBodyParams() below:
+if ($filter->load(\Yii::$app->request->get())) {
```
bscheshirwork: Hm, Interesting realization of `yii\rest\DataFilter::load()`
```php
    /**
     * @inheritdoc
     */
    public function attributes()
    {
        return [
            $this->filterAttributeName
        ];
    }

    /**
     * @inheritdoc
     */
    public function formName()
    {
        return '';
    }
```
@samdark My be we need add link to this in comment?
```php
/**
 * You may load filters from any source. For example
 * if you prefer JSON in request body,
 * use Yii::$app->request->getBodyParams() below:
 * @see \yii\rest\DataFilter::load()
 * @see \yii\rest\DataFilter::attributes()
 * @see \yii\rest\DataFilter::filterAttributeName
 */
```

cebe (approved):

docs/guide/rest-resources.md
```diff
@@ -244,4 +244,64 @@ will also include the pagination information by the following HTTP headers:
...
 An example may be found in the [Quick Start](rest-quick-start.md#trying-it-out) section.
+
+### Filtering collections <span id="filtering-collections"></span>
+
+Since version 2.0.13 Yii provides a facility to filter collections. In case you are using `ActiveController` you can enjoy it
+right away. An example can be found in the [Quick Start](rest-quick-start.md#trying-it-out) section. In case you're
+implementing an endpoint yourself, filtering could be done like the following:
+
+```php
+$filter = new ActiveDataFilter([
```
bscheshirwork: We need tell more about `DataFilter::attributeMap`. 

This property `attributeMap` (not `searchModel`) can be use (additional) to resolve alias, who we use in resources definition
```php
                    'attributeMap' => [
                        'aliasId' => 'realFieldId',
                    ],
```
```php
class \api\Post extends \common\Post {
    public function fields(){
        $fields = parent::fields();
        unset($fields['realFieldId']);
        $fields['aliasId'] = function () {
            return $this->realFieldId;
        };
        return $fields;
    }
}
```

bscheshirwork: So... I think we need move $attributeMap to from controller (definition of DataFilter) to the model `searchModel`
```php
class \api\models\PostSearch extends app\models\Post {
    public static function attributeMap(){
         return ['aliasId' => 'realFieldId'];
    }
    public function fields(){
        $fields = parent::fields();
        unset($fields['realFieldId']);
        $fields['aliasId'] = function () {
            return $this->realFieldId;
        };
        return $fields;
    }
}
```
```php
$filter = new ActiveDataFilter([ 
'searchModel' => \api\models\PostSearch::class,
'attributeMap' => [ \api\models\PostSearch::class, 'attributeMap'],
]);
```

i.e. we need closure support of `attributeMap` property too

samdark:
Data filters moved to `data` namespace. Docs adjusted.

framework/data/DataFilter.php
```diff
@@ -388,12 +388,12 @@ public function setErrorMessages($errorMessages)
...
-            'invalidAttributeValueFormat' => Yii::t('yii-rest', "Condition for '{attribute}' should be either a value or valid operator specification."),
-            'operatorRequireAttribute' => Yii::t('yii-rest', "Operator '{operator}' must be used with a search attribute."),
-            'unsupportedOperatorType' => Yii::t('yii-rest', "'{attribute}' does not support operator '{operator}'."),
+            'invalidFilter' => Yii::t('yii-data', 'The format of {filter} is invalid.'),
+            'operatorRequireMultipleOperands' => Yii::t('yii-data', "Operator '{operator}' requires multiple operands."),
+            'unknownAttribute' => Yii::t('yii-data', "Unknown filter attribute '{attribute}'"),
+            'invalidAttributeValueFormat' => Yii::t('yii-data', "Condition for '{attribute}' should be either a value or valid operator specification."),
+            'operatorRequireAttribute' => Yii::t('yii-data', "Operator '{operator}' must be used with a search attribute."),
+            'unsupportedOperatorType' => Yii::t('yii-data', "'{attribute}' does not support operator '{operator}'."),
```
cebe: there are translation files to be renamed:

```
$ find messages/ |grep rest
messages/de/yii-rest.php
messages/hy/yii-rest.php
```

cebe: Also I don't see why we have a separate category for these. why not use `yii` here too? what is different?

samdark: I don't think anything is...

cebe (approved):

PR #15017, opened 2017-10-24
Adjusted implementation of migration name length limit

cebe:
- warn about the problem before applying migrations
- make it specific to SQL dbms because other implementations do not have a limit

see also:
- commit 614fb52c45b06077682114f2ba246399af0e9a9c
- issue #14543
- PR #15002

### DIFF
--- a/framework/console/controllers/MigrateController.php
+++ b/framework/console/controllers/MigrateController.php
@@ -74,6 +74,13 @@ use yii\helpers\Console;
 class MigrateController extends BaseMigrateController
 {
     /**
+     * Maximum length of a migration name.
+     * @since 2.0.13
+     */
+    const MAX_NAME_LENGTH = 180;
+
+
+    /**
      * @var string the name of the table for keeping applied migration information.
      */
     public $migrationTable = '{{%migration}}';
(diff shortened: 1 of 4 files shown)


framework/console/controllers/MigrateController.php
```diff
@@ -319,6 +326,15 @@ protected function removeMigrationHistory($version)
 
     /**
      * @inheritdoc
+     * @since 2.0.13
+     */
+    protected function getMigrationNameLimit()
+    {
+        return static::MAX_NAME_LENGTH;
```
rob006: What about detecting real limit from DB schema? People may want to use longer names and change length of this column in DB.

samdark: Migrations are meant to be applied to DB connection defined so it's a bad idea. We don't know the environment beforehand.

samdark (approved):

framework/console/controllers/MigrateController.php
cebe: I'm generally okay with detecting it, but in `actionCreate` the migration table may not yet exist.

rob006: @samdark Sorry don't get it. There is already check whether migration table exist:
https://github.com/cebe/yii2/blob/37204ebcb49f5ba96bf671687da50a5e27a7213b/framework/console/controllers/MigrateController.php#L208-L210

What is the problem in checking column length in the same way?

Anyway, this is BC break without it - if I changed column limit to 255 and have have migration with 230 chars, this change will break my environment. And it can't be easily fixed by configuration.

cebe: good point.

rob006: > I'm generally okay with detecting it, but in actionCreate the migration table may not yet exist.

Then we can safely use default limit.

cebe: fixed. however create action needs database access now, not sure how much this change could break anything.

rob006 (approved):

dmirogin (approved):

PR #15018, opened 2017-10-24
\rest\DataFilter $searchModel in error message instead $model 

bscheshirwork:
I think we say about $searchModel property in setter/getter $_searchModel, not about `$model`

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    |  no
| Tests pass?   | yes/no
| Fixed issues  | related #15013


samdark:
Merged. Thanks!

PR #15027, opened 2017-10-26
added StringHelper::floatToString() as locale independent cast

### LINKED ISSUE #15015: PHP notice error with different locale in log/Target.php
Throw an exception and have logging enabled

No php error

### What do you get instead?
PHP error:
```
Notice:  Undefined offset: 1 in /var/www/myproject/vendor/yiisoft/yii2/log/Target.php on line 362
Stack trace:
  1. yii\base\ErrorHandler->handleException() /var/www/myproject/vendor/yiisoft/yii2/base/ErrorHandler.php:0
  2. yii\log\Logger->flush() /var/www/myproject/vendor/yiisoft/yii2/base/ErrorHandler.php:113
  3. yii\log\Dispatcher->dispatch() /var/www/myproject/vendor/yiisoft/yii2/log/Logger.php:177
  4. yii\log\Target->collect() /var/www/myproject/vendor/yiisoft/yii2/log/Dispatcher.php:189
  5. yii\log\EmailTarget->export() /var/www/myproject/vendor/yiisoft/yii2/log/Target.php:128
  6. array_map() /var/www/myproject/vendor/yiisoft/yii2/log/EmailTarget.php:83
  7. yii\log\Target->formatMessage() /var/www/myproject/vendor/yiisoft/yii2/log/EmailTarget.php:83
  8. yii\log\Target->getTime() /var/www/myproject/vendor/yiisoft/yii2/log/Target.php:279
```

### Additional info
The bug happens because you split the timestamp by ".".
In my locale (german) it is a ",". Maybe you can fix it with a simple strpos first or maybe you can use some yii-locale functions.
https://github.com/yiisoft/yii2/blob/614fb52c45b06077682114f2ba246399af0e9a9c/framework/log/Target.php#L362

| Q                | A
| ---------------- | ---
| Yii version      | dev-master (614fb52c45b06077682114f2ba246399af0e9a9c)
| PHP version      |  7.0.22
| Operating system | Debian 3.16.7-ckt20-1+deb8u4

cebe:
This problem popped up in a lot of places when casting a float to
string and the locale sets a decimal separator as `,` unexpected
results occur. This adds a helper which is now called in various places
where this problem exists.

fixes #15015

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15015

### DIFF
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -352,7 +352,7 @@ class BaseStringHelper
     }
 
     /**
-     * Savely casts a float to string independent of the current locale.
+     * Safely casts a float to string independent of the current locale.
      *
      * The decimal separator will always be `.`.
      * @param float|int $number a floating point number or integer.
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -350,4 +350,19 @@ class BaseStringHelper
     {
         return base64_decode(strtr($input, '-_', '+/'));
     }
+
+    /**
+     * Savely casts a float to string independent of the current locale.
+     *
+     * The decimal separator will always be `.`.
+     * @param float|int $number a floating point number or integer.
+     * @return string the string representation of the number.
+     * @since 2.0.13
+     */
+    public static function floatToString($number)
+    {
+        // . and , are the only decimal separators known in ICU data,
+        // so its save to call str_replace here
+        return str_replace(',', '.', (string) $number);
+    }
 }
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -362,7 +362,7 @@ class BaseStringHelper
     public static function floatToString($number)
     {
         // . and , are the only decimal separators known in ICU data,
-        // so its save to call str_replace here
+        // so its safe to call str_replace here
         return str_replace(',', '.', (string) $number);
     }
 }
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -97,6 +97,7 @@ Yii Framework 2 Change Log
 - Chg #14487: Changed i18n message error to warning (dmirogin)
 - Enh #14864: Ability to use dependencies in constructor of migrations (vtvz)
 - Enh #14913: Assset hashing now takes asset linking into account to improve cache busting (schmunk42)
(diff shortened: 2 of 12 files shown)


framework/helpers/BaseStringHelper.php
```diff
@@ -350,4 +350,19 @@ public static function base64UrlDecode($input)
...
+
+    /**
+     * Savely casts a float to string independent of the current locale.
+     *
+     * The decimal separator will always be `.`.
+     * @param float|int $number a floating point number or integer.
+     * @return string the string representation of the number.
+     * @since 2.0.13
+     */
+    public static function floatToString($number)
```
samdark: Name doesn't reflect what it does. It's easy to think about formatting taking locale into account. How about something like `floatToDotString()`.

cebe: well, you should read at least the short description to see what it does. Won't make the name to complex. Imo its not clear what a "dot string" is.

samdark: How about `floatToUnifiedString()`?

dmirogin (approved):

framework/helpers/BaseStringHelper.php
cebe: that also brings no improvement imo. The method creates a string from a float, more information is not needed from the name. The fact that this can depend on the locale and must be normalized is an implementation and documentation detail that does not need to influence the name imo. Keep it simple.

samdark: OK.

samdark (approved):

samdark:
Merged! We're even closer to release ;)

rob006:
How about `StringHelper::stringify()` which will also handle other types (like array or object)?

samdark:
Too universal. Better to create separate methods.

rob006:
There are situation where you don't know type of variable, but you must convert it to string. With separate methods you need to add big switch block - one method will be more convenient.

Anyway, this is really weird that we have special helper for replacing commas by dots, but we refuse to add `ArrayHelper::unique()` which will hide some unintuitive constructions to avoid `array_unique()` performance hit and unwanted sorting. Or `Html::tel()` helper with tricky fullwidth chars support. :D

samdark:
One thing at a time.

cebe:
> There are situation where you don't know type of variable, but you must convert it to string. With separate methods you need to add big switch block - one method will be more convenient.

this is more a job for VarDumper, there is no generic way to convert arbitrary things to a string.

PR #15047, opened 2017-10-30
Fix duplicated columns in Query::addSelect

wapmorgan:
Fix for case when `addSelect` called few times in a chain like this and finally it looks like `SELECT *, abc, *, (INNER QUERY)` by this code:
```php
$this->addSelect(['*', new Expression('...')]);
// ...
$this->addSelect(['*', new Expression('...')]);
```

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | ---

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -590,7 +590,20 @@ PATTERN;
         if ($this->select === null) {
             $this->select = $columns;
         } else {
-            $this->select = array_unique(array_merge($this->select, $columns), SORT_STRING);
+            foreach ($columns as $columnName => $columnDefinition) {
+                if (($columnDefinition instanceof Query) || !in_array($columnDefinition, $this->select, true)) {
+                    if (is_string($columnName))
+                        $this->select[$columnName] = $columnDefinition;
+                    else
+                        $this->select[] = $columnDefinition;
+                } else {
+                    $column_alias = array_search($columnDefinition, $this->select, true);
+                    if (is_string($columnName) && $column_alias != $columnName)
+                        $this->select[$columnName] = $columnDefinition;
+                    else if (is_string($column_alias) && !is_string($columnName))
+                        $this->select[] = $columnDefinition;
+                }
+            }
         }
 
         return $this;
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -605,15 +605,13 @@ PATTERN;
      */
     protected function removeDuplicatedColumns($columns)
     {
-        // delete simple columns duplicates
-        $simpleColumns = array_filter($this->select, 'is_integer', ARRAY_FILTER_USE_KEY);
 
         foreach ($columns as $columnName => $columnDefinition) {
             if ($columnDefinition instanceof Query)
                 continue;
 
             if ((is_string($columnName) && isset($this->select[$columnName]) && $this->select[$columnName] === $columnDefinition) ||
-                (is_integer($columnName) && in_array($columnDefinition, $simpleColumns)))
+                (is_integer($columnName) && in_array($columnDefinition, array_filter($this->select, 'is_integer', ARRAY_FILTER_USE_KEY))))
                 unset($columns[$columnName]);
         }
         return $columns;
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -590,7 +590,7 @@ PATTERN;
         if ($this->select === null) {
             $this->select = $columns;
         } else {
-            $this->select = array_unique(array_merge($this->select, $columns), SORT_REGULAR);
+            $this->select = array_unique(array_merge($this->select, $columns), SORT_STRING);
         }
 
         return $this;
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -590,38 +590,37 @@ PATTERN;
         if ($this->select === null) {
             $this->select = $columns;
         } else {
-            $this->mergeSelect($columns);
+            $this->select = array_merge($this->select, $this->removeDuplicatedColumns($columns));
         }
 
         return $this;
     }
 
     /**
-     * Merges given columns definitions into current query.
+     * Removes duplicated columns from given columns definitions.
      *
      * @param array $columns the columns to be merged to the select.
      */
-    protected function mergeSelect($columns)
+    protected function removeDuplicatedColumns($columns)
     {
         foreach ($columns as $columnName => $columnDefinition) {
-            if (($columnDefinition instanceof Query) || !in_array($columnDefinition, $this->select, true)) {
-                if (is_string($columnName))
-                    $this->select[$columnName] = $columnDefinition;
-                else
-                    $this->select[] = $columnDefinition;
-            } else {
-                if (is_string($columnName) && (!isset($this->select[$columnName]) || $this->select[$columnName] !== $columnDefinition))
-                    $this->select[$columnName] = $columnDefinition;
-                else if (!is_string($columnName)) {
-                    foreach ($this->select as $selectAlias => $selectDefinition) {
-                        if ($selectDefinition === $columnDefinition && !is_string($selectAlias))
-                            continue(2);
-                    }
-                    $this->select[] = $columnDefinition;
-                }
+            if ($columnDefinition instanceof Query)
+                continue;
 
+            if (is_string($columnName) && isset($this->select[$columnName]) && $this->select[$columnName] == $columnDefinition)
+                unset($columns[$columnName]);
+
+
+            if (!is_string($columnName) && in_array($columnDefinition, $this->select)) {
+                foreach ($this->select as $selectAlias => $selectDefinition) {
+                    if (!is_string($selectAlias) && $selectDefinition === $columnDefinition) {
+                        unset($columns[$columnName]);
+                        break;
+                    }
+                }
             }
         }
+        return $columns;
     }
 
     /**
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -590,7 +590,7 @@ PATTERN;
         if ($this->select === null) {
             $this->select = $columns;
         } else {
-            $this->select = array_unique(array_merge($this->select, $columns));
+            $this->select = array_unique(array_merge($this->select, $columns), SORT_REGULAR);
         }
 
         return $this;
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -597,11 +597,16 @@ PATTERN;
                     else
                         $this->select[] = $columnDefinition;
                 } else {
-                    $column_alias = array_search($columnDefinition, $this->select, true);
-                    if (is_string($columnName) && $column_alias != $columnName)
+                    if (is_string($columnName) && (!isset($this->select[$columnName]) || $this->select[$columnName] !== $columnDefinition))
                         $this->select[$columnName] = $columnDefinition;
-                    else if (is_string($column_alias) && !is_string($columnName))
+                    else if (!is_string($columnName)) {
+                        foreach ($this->select as $selectAlias => $selectDefinition) {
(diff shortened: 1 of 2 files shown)


wapmorgan:
Also, what about prepending select part with `*`  when `addSelect` used with empty `Query::$select`?
I mean:
```diff
    public function addSelect($columns)
    {
        if ($columns instanceof Expression) {
            $columns = [$columns];
        } elseif (!is_array($columns)) {
            $columns = preg_split('/\s*,\s*/', trim($columns), -1, PREG_SPLIT_NO_EMPTY);
        }
        if ($this->select === null) {
+           if (!in_array('*', $columns)) array_unshift($columns, '*');
            $this->select = $columns;
        } else {
            $this->select = array_merge($this->select, $columns);
        }
        return $this;
    }
```

wapmorgan:
Added unit test for new code.

dmirogin (approved):

tests/framework/db/QueryTest.php
```diff
@@ -37,6 +37,16 @@ public function testSelect()
...
         $query->addSelect('email');
         $this->assertEquals(['id', 'name', 'email'], $query->select);
+        
+        $query = new Query();
+        $query->select('name, lastname');
+        $query->addSelect('name');
+        $this->assertEquals(['name', 'lastname'], $query->select);
+        
+        $query = new Query();
+        $query->addSelect(['*', 'abc']);
```
samdark: I'd test that selecting same column with two different aliases works well i.e.

```php
$query->addSelect(['field1 as a', 'field 1 as b']);
```

wapmorgan: Added

framework/db/Query.php
```diff
@@ -590,7 +590,7 @@ public function addSelect($columns)
         if ($this->select === null) {
             $this->select = $columns;
         } else {
-            $this->select = array_merge($this->select, $columns);
+            $this->select = array_unique(array_merge($this->select, $columns), SORT_REGULAR);
```
rob006: `array_unique()` changes order - this will probably break BC.

rob006 (changes requested):

framework/db/Query.php
samdark: Yes. Order should be kept and it should be in the test.

wapmorgan: As I can see, array_unique does not changes order of elements. It just deletes duplicates. I'm not good at C++, but I think [this code block](https://github.com/php/php-src/blob/a8a17a72b06e6ea39a70eaf45ad2c3d6668e29ba/ext/standard/array.c#L4527) creates internal array of pointers and sorts it for comparation. After that it deletes elements from the original array so it's not being sorted at all. This snippets tests functionality: https://3v4l.org/2v8UV It saved all unique elements from first array and added unique elements from the second one.

rob006: @wapmorgan You're right. But you should probably leave `SORT_STRING` as default `$sort_flags` - `SORT_REGULAR` is unreliable with when you compare different types: https://3v4l.org/coIXF

samdark:
Seems travis isn't happy with the pull request: https://travis-ci.org/yiisoft/yii2/builds/297224586

wapmorgan:
Updated `addSelect`. Now it does clever duplicate detection:
- Checks aliases of column / expression and does not mix columns w/wo aliases.
- Handles `Query` objects properly
- Removes duplicates if they have the same definiton, but not aliases (if present)
Test case shows all that.

wapmorgan:
Should I simplify method `addSelect` (as codeclimate says)?

samdark:
If it is possible.

wapmorgan:
What is the minimal PHP version Yii2 requires?

samdark:
5.4 for 2.0, 7.1 for 2.1.

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
 
         return $this;
     }
 
+    /**
+     * Removes duplicated columns from given columns definitions.
+     * Columns to be removed:
+     * - if column definition already present in SELECT part with same alias
+     * - if column definition without alias already present in SELECT part without alias too
+     * @param array $columns the columns to be merged to the select.
```
samdark: Need `@since 2.0.14`.

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
         } else {
-            $this->select = array_merge($this->select, $columns);
+            $this->select = array_merge($this->select, $this->removeDuplicatedColumns($columns));
         }
 
         return $this;
     }
 
+    /**
+     * Removes duplicated columns from given columns definitions.
```
samdark: Returns unique column names excluding duplicates.

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
+     * Columns to be removed:
+     * - if column definition already present in SELECT part with same alias
+     * - if column definition without alias already present in SELECT part without alias too
+     * @param array $columns the columns to be merged to the select.
+     */
+    protected function removeDuplicatedColumns($columns)
```
samdark: Since the function returns and doesn't affect `$columns`, I'd name it `getUniqueColumns()`.

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
+    {
+        $simple_columns = $this->filterSimpleColumnsFromSelect();
```
samdark: Should be renamed to `$simpleColumns`.

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
+
+        foreach ($columns as $columnName => $columnDefinition) {
+            if ($columnDefinition instanceof Query)
```
samdark: Missing `{` and `}`.

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
         if ($this->select === null) {
             $this->select = $columns;
         } else {
-            $this->select = array_merge($this->select, $columns);
+            $this->select = array_merge($this->select, $this->removeDuplicatedColumns($columns));
```
samdark: Shouldn't it be applied to `select()` as well?

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
+     * Columns to be removed:
+     * - if column definition already present in SELECT part with same alias
+     * - if column definition without alias already present in SELECT part without alias too
+     * @param array $columns the columns to be merged to the select.
+     */
+    protected function removeDuplicatedColumns($columns)
+    {
+        $simple_columns = $this->filterSimpleColumnsFromSelect();
+
+        foreach ($columns as $columnName => $columnDefinition) {
```
samdark: `$columnName` → `$columnAlias`

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
+
+            if ((is_string($columnName) && isset($this->select[$columnName]) && $this->select[$columnName] === $columnDefinition) ||
+                (is_integer($columnName) && in_array($columnDefinition, $simple_columns)))
+                unset($columns[$columnName]);
+        }
+        return $columns;
+    }
+
+    /**
+     * @return array List of simple columns from query SELECT
```
samdark: Need `@since 2.0.14`

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
+     */
+    protected function filterSimpleColumnsFromSelect()
```
samdark: Since it's getter, I'd rename it to `getUnaliasedColumnsFromSelect()`.

framework/db/Query.php
samdark: List of columns without aliases from SELECT statement.

framework/db/Query.php
```diff
@@ -590,12 +590,47 @@ public function addSelect($columns)
...
+    {
+        $result = [];
+        foreach ($this->select as $name => $value) {
+            if (is_integer($name) && !in_array($value, $result))
```
samdark: Have you checked if not doing `in_array` here and instead calling `array_unique` right before return will be better for performance?

samdark: Missing `{` and `}`.

tests/framework/db/QueryTest.php
```diff
@@ -37,6 +37,26 @@ public function testSelect()
         $query->select('id, name');
         $query->addSelect('email');
         $this->assertEquals(['id', 'name', 'email'], $query->select);
+
+        $query = new Query();
+        $query->select('name, lastname');
```
samdark: Need a test for `$query->select('name, name, name as X, name as X')`

samdark (changes requested):

tests/framework/db/QueryTest.php
wapmorgan: Is it real case when programmer duplicates select parts for purpose?

samdark: Typos do happen.

framework/db/Query.php
```diff
@@ -590,45 +590,51 @@ public function addSelect($columns)
...
+        foreach ($columns as $columnAlias => $columnDefinition) {
+            if ($columnDefinition instanceof Query) {
                 continue;
+            }
 
-            if ((is_string($columnName) && isset($this->select[$columnName]) && $this->select[$columnName] === $columnDefinition) ||
-                (is_integer($columnName) && in_array($columnDefinition, $simple_columns)))
-                unset($columns[$columnName]);
+            if ((is_string($columnAlias) && isset($this->select[$columnAlias]) && $this->select[$columnAlias] === $columnDefinition) ||
+                (is_integer($columnAlias) && in_array($columnDefinition, $simpleColumns))) {
```
samdark: in_array is still there and array_unique was added. Doesn't make sense to me.

framework/db/Query.php
```diff
@@ -590,45 +590,51 @@ public function addSelect($columns)
...
      * - if column definition already present in SELECT part with same alias
      * - if column definition without alias already present in SELECT part without alias too
      * @param array $columns the columns to be merged to the select.
+     * @since 2.0.14
      */
-    protected function removeDuplicatedColumns($columns)
+    protected function getUniqueColumns($columns)
     {
-        $simple_columns = $this->filterSimpleColumnsFromSelect();
+        $simpleColumns = $this->getUnaliasedColumnsFromSelect();
```
samdark: Variable should be renamed to unaliasedColumns since method name was adjusted.

samdark (changes requested):

framework/db/Query.php
wapmorgan: I've removed `in_array` in `getUnaliasedColumnsFromSelect()`. This `in_array` is inteded to check if an unalised column is already present.

samdark (approved):
OK. The only thing missing is a line for changelog. Good to merge, as for me.

samdark:
One more thing are HHVM tests...

wapmorgan:
Working with HHVM compatability...

wapmorgan:
But there is no `CHANGELOG` file

wapmorgan:
Oh, I see it's a lot of commits already added. Should I rebase my branch onto master or just merge all commits into my branch?

samdark:
No. Code looks good now. Thanks.

wapmorgan:
But I can't add it to changelog now

rob006 (approved):

samdark:
Merged. Thank you!

PR #15049, opened 2017-10-30
Fixed DbSession for Microsoft SQL Server, enhanced tests

### LINKED ISSUE #13436: MSSQL DB session doesn't work
See https://github.com/yiisoft/yii2/issues/12599#issuecomment-274663294

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11
| PHP version      | any
| Operating system | any

SilverFire:
Fixes #13436

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13436

screen shot 2017-10-30 at 15 20 24
screen shot 2017-10-30 at 15 20 56
screen shot 2017-10-30 at 15 21 09
screen shot 2017-10-30 at 15 22 29


samdark (approved):

tests/framework/web/session/DbSessionTest.php
```diff
@@ -16,42 +16,74 @@
 
 /**
  * @group db
+ * @group sqlite
  */
 class DbSessionTest extends TestCase
```
cebe: better make this class abstract and create an sqlite implementation of it.

framework/web/migrations/m160313_153426_session_init.php
```diff
@@ -20,16 +20,23 @@ class m160313_153426_session_init extends Migration
...
-            // http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci
-            $tableOptions = 'CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE=InnoDB';
+
+        switch ($this->db->driverName) {
+            case 'mysql':
+                // http://stackoverflow.com/questions/766809/whats-the-difference-between-utf8-general-ci-and-utf8-unicode-ci
+                $tableOptions = 'CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE=InnoDB';
+                break;
+            case 'sqlsrv':
+                $dataType = $this->text();
```
cebe: we had issues with non-utf8 data stored in session, which caused text to fail so we used binary for that. Is this covered by a test?

SilverFire: I remembered it, but was not sure about characters that cause that behavior, so typed something random: https://github.com/yiisoft/yii2/pull/15049/files#diff-722a201958cf84ca29390026ac2a10eaR138

Hope it covers. If you remember something more exact – share a link please

samdark:
Hmmm... fails.

SilverFire:
Fixed `rbac/DbManagerTestCase` to revert all migrations, instead of the latest only

dmirogin (approved):

framework/web/migrations/m160313_153426_session_init.php
cebe: The issue is about raw binary data that does not follow utf8 encoding, e.g. https://github.com/yiisoft/yii2/issues/933

cebe (changes requested):
Need a test to verify raw binary data does not cause issues. See e.g. #933, which is about raw binary data created by encryption that does not follow utf8 encoding.

SilverFire:
Yeah, it did not work. I've tried to fix it, but was able to fix only work with control chars (null-byte, etc).
Inserting binary data leads to the error that occurs either with `nvarchar(max)` or `varbinnary(max)`:

```
An error occurred translating string for input param 1 to UCS-2: Error code 0x7
```

I've adjusted test to make sure other DBMSs handle binary session data correctly, but I'm not a Microsoft SQL Server expert and I doubt I can fix any more.

Symfony [PdoSessionHandler test](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Tests/Session/Storage/Handler/PdoSessionHandlerTest.php) does not check possibility of inserting binary data and I think I know why ;)

cebe:
if we can't fix it its okay to merge it, but maybe @sergeymakinen has an idea how to make it work?

samdark:
He's extremely busy so waiting may hold release significantly.

cebe (approved):

Issue #15053, opened 2017-10-31
Use BaseYii::configure in Container::build or vice versa


vtvz:
These two places in framework do the same thing -- configure the object:

https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php#L385-L387

https://github.com/yiisoft/yii2/blob/master/framework/BaseYii.php#L529-L536

Why not to use `BaseYii::configure` (Or `Yii::configure`) method in `Container::build`, or in container make function like `configureObject` and use it like this:

```php
public static function configure($object, $properties)
{
    return self::$container->configureObject($object, $properties);
}
```

I can create PR if it's necessary.

// Sorry for my bad English

samdark:
Makes sense from DRY standpoint.

PR #15054, opened 2017-10-31
 Configure objects through DI container

vtvz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes/no
| Breaks BC?    | no
| Tests pass?   | tests of di group are pass
| Fixed issues  | #15053

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -528,11 +528,7 @@ class BaseYii
      */
     public static function configure($object, $properties)
     {
-        foreach ($properties as $name => $value) {
-            $object->$name = $value;
-        }
-
-        return $object;
+        return self::$container->configureObject($object, $properties);
     }
 
     /**
--- a/framework/di/Container.php
+++ b/framework/di/Container.php
@@ -349,6 +349,21 @@ class Container extends Component
     }
 
     /**
+     * Configures an object with the initial property values.
+     * @param object $object the object to be configured
+     * @param array $properties the property initial values given in terms of name-value pairs.
+     * @return object the object itself
+     */
+    public function configureObject($object, $properties)
+    {
+        foreach ($properties as $name => $value) {
+            $object->$name = $value;
+        }
+
+        return $object;
+    }
+
+    /**
      * Creates an instance of the specified class.
      * This method will resolve dependencies of the specified class, instantiate them, and inject
      * them into the new instance of the specified class.
@@ -382,11 +397,8 @@ class Container extends Component
         }
 
         $object = $reflection->newInstanceArgs($dependencies);
-        foreach ($config as $name => $value) {
-            $object->$name = $value;
-        }
 
-        return $object;
+        return $this->configureObject($object, $config);
     }
 
     /**
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -76,7 +76,7 @@ Yii Framework 2 Change Log
 - Chg #14286: Used primary inputmask package name instead of an alias (samdark)
 - Chg #14321: `yii\widgets\MaskedInput` is now registering its JavaScript `clientOptions` initialization code in head section (DaveFerger)
 - Chg #14487: Changed i18n message error to warning (dmirogin)
-
+- Enh #14864: Ability to use dependencies in constructor of migrations (vtvz)
 
 2.0.12 June 05, 2017
 --------------------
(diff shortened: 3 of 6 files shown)


vtvz:
My tests are fine:

```bash
$ vendor/bin/phpunit -v --group di

Runtime:        PHP 7.0.14
Configuration:  /project/phpunit.xml.dist

............................

Time: 505 ms, Memory: 36.00MB

OK (28 tests, 96 assertions)

```

samdark:
DI migration commits should not be in the pull request.

samdark:
Since container is used a lot I wonder how bad replacing foreach with function call is performance-wise.

framework/BaseYii.php
```diff
@@ -528,11 +528,7 @@ public static function t($category, $message, $params = [], $language = null)
      */
     public static function configure($object, $properties)
     {
-        foreach ($properties as $name => $value) {
-            $object->$name = $value;
-        }
-
-        return $object;
+        return self::$container->configureObject($object, $properties);
```
cebe: container could be unavailable if someone is not using it.

vtvz:
@samdark I'll fix it tomorrow.

framework/BaseYii.php
vtvz: Okey. It can be in opposite way. You can configure object using BaseYii or Yii method.
But I think it should be in one place (DRY)

bscheshirwork: We can configure any/all of object 
or only `implements \yii\base\Configurable`?

samdark: Any object.

vtvz: @bscheshirwork it configures any object.

framework/di/Container.php
```diff
@@ -349,6 +349,21 @@ public function getDefinitions()
     }
 
     /**
+     * Configures an object with the initial property values.
+     * @param object $object the object to be configured
+     * @param array $properties the property initial values given in terms of name-value pairs.
+     * @return object the object itself
```
bscheshirwork: "the object itself" is correct in this context?

framework/di/Container.php
```diff
@@ -349,6 +349,21 @@ public function getDefinitions()
...
+     */
+    public function configureObject($object, $properties)
+    {
+        foreach ($properties as $name => $value) {
```
bscheshirwork: What happens if `$object` has no `$name` property? 

This case is coverage in test?

vtvz: Throws expected exception

framework/di/Container.php
samdark: Yes. You're passing an object and the same object is returned with properties set.

framework/di/Container.php
bscheshirwork: And what happens in this case:

```php
...
private $b;
public function setB(array $value): void{
   $this $b;
   return;
}

...
...configureObject($TestObject, ['b' => (string) 100]);
```

cebe: the setter is called.

cebe:
In general I do not really see the benefit of doing this. Its a simple foreach loop, there is no need to extract that to another component and create a fixed dependency.

samdark:
I tend to agree. @vtvz are there any benefits we don't see?

framework/di/Container.php
bscheshirwork: How to process mismatch of type in setter witch strict types?

samdark: It will throw an exception. That's native PHP behavior and is correct in the case.

vtvz:
Configuration of object should be in one place for easier improving and refactoring of this functionality. I have some ideas for future. It doesn't matter where it will be (Yii, BaseYii, Container or Helper)

samdark:
https://github.com/yiisoft/yii2/pull/15054#issuecomment-340917475 should be checked.

cebe:
> I have some ideas for future.

could you elaborate on what these ideas are about?

vtvz:
1/2. Configure non `Configurable` objects through setter and/or call method. For example:
```php
[
    '!someMethod' => ['arg1', 'arg2'],
    'prop' => 'arg3',
]
```
It will call someMethod with these two argument when configure. And calls setProp('arg3') (may be)

3. Resolve dependencies using `yii\di\Instance` object. (Now it works only with params)

I think it will make Yii's DI container really powerful tool.

cebe:
Imo we should first discuss such behavior and its implementation. If we agree on how to do it we can find a place for it. It does not make sense to move the foreach loop now and add such functionality later, which might be useful in one place but not in the other.

framework/di/Container.php
bscheshirwork: @samdark yes, that's what I mean.

This is expected behaviors AND this behavior is not reflecting in tests?

I can't see It 
```php
$this->expectException('\TypeError');
```
 in `tests/framework/BaseYiiTest.php` and in `tests/framework/di/ContainerTest.php`

https://3v4l.org/LitK2 and https://3v4l.org/KQFRH

samdark: 👍 would be great if you'll add it. Thanks.

PR #15055, opened 2017-10-31
Named params is DI container

vtvz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | tests of DI group is pass

Example:

```php
class Foo
{
    public $bar;
    public $bar2;

    public function __construct($bar, $bar2)
    {
        $this->bar = $bar;
        $this->bar2 = $bar2;
    }
}

//.....

$container->set(Foo::class, [], [
    'bar2' => 'value_two',
    'bar' => 'value_one',
]);

//.....

$container->get(Foo::class);
// which is equivalent to:
new Foo('value_one', 'value_two');
```

Tests:
```bash
docker-compose build; docker-compose run --rm php vendor/bin/phpunit -v --group di

Runtime:        PHP 7.0.14
Configuration:  /project/phpunit.xml.dist

.............................

Time: 839 ms, Memory: 36.00MB

OK (29 tests, 98 assertions)
```

### DIFF
--- a/framework/di/Container.php
+++ b/framework/di/Container.php
@@ -349,6 +349,40 @@ class Container extends Component
     }
 
     /**
+     * Returns the dependencies merged with input params
+     * @param array $dependencies array of dependencies
+     * @param array $params array of params
+     * @param ReflectionClass $reflection the class reflection associated with the dependencies
+     * @return array
+     */
+    protected function mergeDependenciesWithParams($dependencies, $params, ReflectionClass $reflection)
+    {
+        $mapNames = [];
+
+        $constructor = $reflection->getConstructor();
+
+        $constructorParams = [];
+
+        if ($constructor !== null) {
+            $constructorParams = $constructor->getParameters();
+        }
+
+        foreach ($constructorParams as $index => $constructorParam) {
+            $mapNames[$constructorParam->getName()] = $index;
+        }
+
+        foreach ($params as $index => $param) {
+            if (is_string($index)) {
+                $dependencies[$mapNames[$index]] = $param;
+            } else {
+                $dependencies[$index] = $param;
+            }
+        }
+
+        return $dependencies;
+    }
+
+    /**
      * Creates an instance of the specified class.
      * This method will resolve dependencies of the specified class, instantiate them, and inject
      * them into the new instance of the specified class.
@@ -363,9 +397,7 @@ class Container extends Component
         /* @var $reflection ReflectionClass */
         list($reflection, $dependencies) = $this->getDependencies($class);
 
-        foreach ($params as $index => $param) {
-            $dependencies[$index] = $param;
-        }
+        $dependencies = $this->mergeDependenciesWithParams($dependencies, $params, $reflection);
 
         $dependencies = $this->resolveDependencies($dependencies, $reflection);
         if (!$reflection->isInstantiable()) {
--- a/tests/framework/di/ContainerTest.php
+++ b/tests/framework/di/ContainerTest.php
@@ -310,4 +310,20 @@ class ContainerTest extends TestCase
 
         require __DIR__ . '/testContainerWithVariadicCallable.php';
     }
+
+    /**
+     * @requires PHP 5.6
+     */
+    public function testNamedParams()
+    {
+        $container = new Container();
+        $container->set(Qux::className(), [], [
+            'a' => 'value_one'
+        ]);
+
+        $object = $container->get(Qux::className());
+
+        $this->assertInstanceOf(Qux::className(), $object);
+        $this->assertSame($object->a, 'value_one');
+    }
 }


samdark:
Some commits in this pull request should not be there. When making a pull request it's important to make a separate branch based on master that has no changes from upstream.

vtvz:
@samdark done. What else I can do?
I wrote unit test for this case. Can't find any code coverage information on the Travis CI. May be will try to generate it by hand.

About use case. You can use associated array for params on getting object from container, where key is the name of parameter. This is useful, because string is more informative than integer, you can pass parameters in any order, and you can inject dependencies partially (for example when you set definition you can set just first and third params, but when you get instance you define only second parameter)

samdark:
Unit tests and some usage examples.

samdark:
A few examples of newly added syntax would be perfect.

vtvz:
@samdark I leave one example it the description of pull request. Isn't enough?

vtvz:
Something similar made for resolve callable dependencies:
https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php#L533-L541

SilverFire:
As for me, it's not correct. Let me explain why.

Compare two pieces of code:

```php
function format($a, $b, $c) {
    return $a . ': ' . $b . ' ≠ ' . $c;
}

echo format('Unreal', 2, 2); // Unreal: 2 ≠ 2
```

and

```php
function format($statement, $numberOne, $numberTwo) {
    return $statement . ': ' . $numberOne . ' ≠ ' . $numberOne;
}

echo format('Unreal', 2, 2); // Unreal: 2 ≠ 2
```

I've changed parameter names, but the API is still valid and the result is equal.
It does NOT matter how will I name the parameter because it is not a part of API – it's internals. Parameters order changing WILL break backward compatibility.

At the moment we start to rely on internals (parameter names) we make refactoring even harder. Want to rename parameter? Bump a major release because it breaks API. It's ridiculous as for me.

vtvz:
@SilverFire it's partially true. But if you care a lot about BC you should remove these lines of code:

https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php#L533
https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php#L536-L538
https://github.com/yiisoft/yii2/blob/master/framework/di/Container.php#L541-L542
... and some others in `Container::resolveCallableDependencies` method.

It does almost the same what I talk about, but with callables  instead of classes.
You can write in docs do not to use this feature with framework and other libs, but **only** for internal usage (in project). It can be very useful for some purposes.

// Why not to use `Container::resolveCallableDependencies` method to resolve `__construct` dependencies? `__construct` is also callable. An I wrong?

bscheshirwork:
what about this usercase:
`base/Foo.php`
```php
class Foo
{
    public $bar;
    public $bar2;

    public function __construct($bar, $bar2)
    {
        $this->bar = $bar;
        $this->bar2 = $bar2;
    }
}
```
`redefine/Foo.php`
```php
class Foo
{
    public $bar;
    public $bar2;

    public function __construct($a, $b)
    {
        $this->bar = $b;
        $this->bar2 = $a;
    }
}
```

`autoload map ['Foo' => 'base/Foo.php']`
```php
$container->set(Foo::class, [], [
    'bar2' => 'value_two',
    'bar' => 'value_one',
]);

//.....

$container->get(Foo::class);
// which is equivalent to:
new Foo('value_one', 'value_two');
```

Just replace source...

`autoload map ['Foo' => 'redefine/Foo.php']`

```php
$container->set(Foo::class, [], [
    'bar2' => 'value_two',
    'bar' => 'value_one',
]);
```

bscheshirwork:
i.e. I think we have seem of differences between redefine class and redefine callable...

vtvz:
@bscheshirwork This feature very powerful for some cases. It can be (must be) as a feature. But you can (must) warn users do not to use it for external components as libs or framework because it's not a part of API.

bscheshirwork:
@vtvz my example applied for base framework classes too. 

About this case: For example I have some redefine classes who fix a error (before PR will be merged at this official repo)

samdark:
Here's when named arguments support was added for invoke: https://github.com/yiisoft/yii2/pull/10811

samdark:
IMO this one goes in line with the #10811.

samdark:
@yiisoft/core-developers, @yiisoft/reviewers ?

cebe:
@SilverFire has a good point here. What are the benefits of allowing named parameters? Why is it needed. Did not see much about "why" here.

SilverFire:
> You can write in docs do not to use this feature with framework and other libs, but only for internal usage (in project). It can be very useful for some purposes.

This is the truth as well.

Sometimes "incorrect" things are extremely useful. @vtvz what is the biggest profit for you in therms of this PR?

samdark:
Won't be merged in 2.0. If you want it in 2.1, please create an issue at https://github.com/yiisoft/di

PR #15056, opened 2017-10-31
Updated Greek translation.

lourdas:
I know 2.0.13 is probably a few hours away, but maybe you could catch this last pull request that updates the Greek language messages file. Thank you!!!


samdark:
Merged. Thanks!

PR #15061, opened 2017-11-01
Resolve Instances on configuration of objects

vtvz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Example:
```php
[
    'class' => SomeService::class,
    'logger' => Instance::of('logger'),
    'transactionManager'  => Instance::of(TransactionManager::class),
],
```

It allows resolve dependencies by key using existing `\yii\di\Instance` class.

It is what I really need in my project.

### DIFF
--- a/framework/di/Container.php
+++ b/framework/di/Container.php
@@ -382,11 +382,8 @@ class Container extends Component
         }
 
         $object = $reflection->newInstanceArgs($dependencies);
-        foreach ($config as $name => $value) {
-            $object->$name = $value;
-        }
 
-        return $object;
+        return Yii::configure($object, $config);
     }
 
     /**
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -11,6 +11,7 @@ use yii\base\InvalidConfigException;
 use yii\base\InvalidParamException;
 use yii\base\UnknownClassException;
 use yii\di\Container;
+use yii\di\Instance;
 use yii\log\Logger;
 
 /**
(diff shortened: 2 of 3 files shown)


samdark:
Looks good to me.

vtvz:
Added some tests. Hope that helps

framework/di/Container.php
```diff
@@ -382,11 +382,8 @@ protected function build($class, $params, $config)
         }
 
         $object = $reflection->newInstanceArgs($dependencies);
-        foreach ($config as $name => $value) {
-            $object->$name = $value;
-        }
 
-        return $object;
+        return Yii::configure($object, $config);
```
samdark: I think container tests are necessary as well with the case as seen in the pull request description.

vtvz: Okey. I'll try

SilverFire:
I wanted to add this possibility in 2.0.14 for my projects. Thank you for reading my thoughts :)

framework/di/Container.php
SilverFire: I don't agree with this change. Container should not depend on Yii::configure to create object.

SilverFire (changes requested):
I think this behavior should be implemented

framework/di/Container.php
SilverFire: As for me it's better to leave foreach and handle `Instance` in a special way, as is is done in `\Yii::configure()`

samdark:
Seems to be resolved by #15347.

PR #15062, opened 2017-11-01
Documented CSS/JS asset options per file

marc-pelteret:
As per issues #10158 and #10166. Order of details on CSS and JS properties swapped because of the order they're listed in the template.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | n/a


samdark:
Merged. Thank you!

Issue #15044, opened 2017-10-30
document error about the AssetBundle::$publishOptions usage


JeanWolf:
Document  [http://www.yiiframework.com/doc-2.0/guide-structure-assets.html](http://www.yiiframework.com/doc-2.0/guide-structure-assets.html)  error
```
class FontAwesomeAsset extends AssetBundle 
{
    public $sourcePath = '@bower/font-awesome'; 
    public $css = [ 
        'css/font-awesome.min.css', 
    ];
    public $publishOptions = [
        'only' => [
            'fonts/',
            'css/',
        ]
    ];
}  
//The above example defines an asset bundle for the "fontawesome" package. 
//By specifying the only publishing option, only the fonts and css subdirectories will be published.
```

### What steps will reproduce the problem?

set `$publishOptions` in class `AppAsset` like
```
    public $publishOptions = [
        'only' => [
            'fonts/',
            'css/',
        ]
    ];
```

### What is the expected result?

All asset files in  `fonts/` and `css/`  should be published.  

### What do you get instead?

No asset files  published ...

### Right Usage

set `$publishOptions` like this
```
    public $publishOptions = [
        'only' => [
            'fonts/*',
            'css/*',
        ]
    ];
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.0.23
| Operating system |MacOS 10.12

PR #15067, opened 2017-11-02
Fix the wrong way

xutl:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15044


samdark:
Merged. Thanks!

Issue #15052, opened 2017-10-31
GridView - can't set css class on tbody when empty rows


uaoleg:
### `emptyTextOptions` should be assigned to the `tbody` instead of `div` inside it

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.?
| PHP version      | 7.0
| Operating system | Windows

bscheshirwork:
@samdark 
https://github.com/yiisoft/yii2/blob/46b9d607a25769c8554773291698c4f6ec654187/framework/grid/GridView.php#L481-L485

we have 
BC broken

lubosdz:
You can use any stronger CSS selector for overriding default styles. E.g. you may use 

* outer wrapper's ID (`#js-tourists-table tbody tr`), 
* attribute selector e.g. `table td[colspan=2]`, 
* pseudoselectors like `:first-child, tbody > tr:nth-of-type(odd)`  
* combined tags & classes etc ... 

Bootstrap is designed with low specificity to enable quite easy overriding. 
See how CSS power/priority/hierarchy/specifity is calculated - https://stackoverflow.com/questions/4072365/css-understanding-the-selectors-priority-specificity.

PR #15070, opened 2017-11-02
add renderEmptyTableRow to GridView

bscheshirwork:
BC broken:

All of empty tables will be changes.

User functions must correct process null-value.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | YES
| Tests pass?   | yes/no
| Fixed issues  | #15052

### DIFF
--- a/framework/grid/GridView.php
+++ b/framework/grid/GridView.php
@@ -479,9 +479,7 @@ class GridView extends BaseListView
         }
 
         if (empty($rows) && $this->emptyText !== false) {
-            $colspan = count($this->columns);
-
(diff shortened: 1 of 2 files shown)


samdark:
Couldn't it be done without breaking compatibility?

bscheshirwork:
> Couldn't it be done without breaking compatibility?

We have differences between produced output and expected output now.

Of course fix it is change output.

...about closure: this is expected behavior, I think we need change this is same part of changes.

samdark:
Won't make it into 2.0.

Issue #15073, opened 2017-11-02
Slovak translations


pistej:
Add missing Slovak translations

cebe:
@pistej could you provide translations?

pistej:
Im am working on it.

PR #15074, opened 2017-11-02
Added slovak framework messages [skip ci] #15073

pistej:
sk translation


samdark:
Just in time. Thanks!

PR #15077, opened 2017-11-03
Fixes #15076: Perfect `yii\db\QueryBuilder` buildColumns function

### LINKED ISSUE #15076: yii\db\QueryBuilder buildColumns function
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.?
| PHP version      | PHP7.1.0
| Operating system | Win10

```
public function buildColumns($columns)
    {
        if (!is_array($columns)) {
            if (strpos($columns, '(') !== false) {
                return $columns;
            } else {
                $columns = preg_split('/\s*,\s*/', $columns, -1, PREG_SPLIT_NO_EMPTY);
            }
        }
        foreach ($columns as $i => $column) {
            if ($column instanceof Expression) {
                $columns[$i] = $column->expression;
            } elseif (strpos($column, '(') === false) {
                $columns[$i] = $this->db->quoteColumnName($column);
            }
        }

        return is_array($columns) ? implode(', ', $columns) : $columns;
    }
```

$columns is a array, so not required use is_array check $columns type, direct use implode(', ', $columns) return.

hiscaler:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15076

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -1151,6 +1151,9 @@ class QueryBuilder extends \yii\base\BaseObject
             }
 
             $columns = preg_split('/\s*,\s*/', $columns, -1, PREG_SPLIT_NO_EMPTY);
+            if ($columns === false) {
+                return '';
+            }
         }
         foreach ($columns as $i => $column) {
             if ($column instanceof Expression) {
@@ -1160,7 +1163,7 @@ class QueryBuilder extends \yii\base\BaseObject
             }
         }
 
-        return is_array($columns) ? implode(', ', $columns) : $columns;
+        return implode(', ', $columns);
     }
 
     /**
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -1150,9 +1150,10 @@ class QueryBuilder extends \yii\base\BaseObject
                 return $columns;
             }
 
+            $rawColumns = $columns;
             $columns = preg_split('/\s*,\s*/', $columns, -1, PREG_SPLIT_NO_EMPTY);
             if ($columns === false) {
-                throw new InvalidParamException("$columns is not valid columns.");
+                throw new InvalidParamException("$rawColumns is not valid columns.");
             }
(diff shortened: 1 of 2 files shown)


framework/db/QueryBuilder.php
```diff
@@ -1151,6 +1151,9 @@ public function buildColumns($columns)
             }
 
             $columns = preg_split('/\s*,\s*/', $columns, -1, PREG_SPLIT_NO_EMPTY);
+            if ($columns === false) {
+                return '';
```
cebe: what is the error that could occur here? I'm quite sure returning an empty string here will result in an SQL syntax error. It should throw an exception instead.

hiscaler: Yes, you are right.here sql is has syntax error if user use invalid fields or string. system will can't continue and hint use check code. Nothing or throw an exception?

cebe: throwing a meaningful exception is better than creating an sql error.

hiscaler: yes, Thank you. i change it .throw a InvalidParamException exception if `columns` param error.

framework/db/QueryBuilder.php
```diff
@@ -1151,6 +1151,9 @@ public function buildColumns($columns)
             }
 
             $columns = preg_split('/\s*,\s*/', $columns, -1, PREG_SPLIT_NO_EMPTY);
+            if ($columns === false) {
+                throw new InvalidParamException("$columns is not valid columns.");
```
samdark: We know it's `false`. Why building a string from a variable?

samdark (approved):

framework/db/QueryBuilder.php
hiscaler: I am sorry, is fixed now.

samdark:
Merged. Thanks!

Issue #15081, opened 2017-11-03
Random crash: Undefined offset: 1 in log Target


ischenko:
### What steps will reproduce the problem?
Fastest way is to turn on profiling.

A value from `microtime(true)` call does not always contain usec part. So the expression `list($timestamp, $usec) = explode('.', StringHelper::floatToString($timestamp));` will trigger PHP error in such cases.

### What is the expected result?
Application should not crash.

### What do you get instead?
```
PHP Error[8]: Undefined offset: 1
    in file /opt/lpp/vendor/yiisoft/yii2/log/Target.php at line 363
#0 /opt/lpp/vendor/yiisoft/yii2/log/Target.php(280): yii\log\FileTarget->getTime()
#1 unknown(0): yii\log\FileTarget->formatMessage()
#2 /opt/lpp/vendor/yiisoft/yii2/log/FileTarget.php(106): array_map()
#3 /opt/lpp/vendor/yiisoft/yii2/log/Target.php(129): yii\log\FileTarget->export()
#4 /opt/lpp/vendor/yiisoft/yii2/log/Dispatcher.php(189): yii\log\FileTarget->collect()
#5 /opt/lpp/vendor/yiisoft/yii2/log/Logger.php(177): yii\log\Dispatcher->dispatch()
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13
| PHP version      | 7.0.25
| Operating system | Ubuntu

cebe:
in which case does microtime does not contain mircoseconds?

ischenko:
I suppose when microseconds equal to zero then value returned without floating part.

ischenko:
```php
while (true) list($a, $b) = explode('.', (string)microtime(true));
```
This will trigger the error for sure :)

cebe:
@ischenko thanks, would you like to send a pull request to fix this?

PR #15083, opened 2017-11-03
Fixes issue #15081

ischenko:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15081

### DIFF
--- a/framework/log/Target.php
+++ b/framework/log/Target.php
@@ -361,8 +361,8 @@ abstract class Target extends Component
      */
     protected function getTime($timestamp)
     {
-        @list($timestamp, $usec) = explode('.', StringHelper::floatToString($timestamp));
+        $parts = explode('.', StringHelper::floatToString($timestamp));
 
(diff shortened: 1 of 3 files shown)


framework/log/Target.php
```diff
@@ -361,8 +361,8 @@ public function getEnabled()
      */
     protected function getTime($timestamp)
     {
-        list($timestamp, $usec) = explode('.', StringHelper::floatToString($timestamp));
+        @list($timestamp, $usec) = explode('.', StringHelper::floatToString($timestamp));
```
samdark: It's better to use array and isset rather than `@`.

samdark (approved):

cebe (approved):
Needs a CHANGELOG entry

samdark:
Merged. Thanks!

Issue #14620, opened 2017-08-09
Why still using deprecated jQuery(document).ready()?


alexantr:
`yii\web\View` uses not recommended way to include scripts in POS_READY section.

jQuery recommends to use
```js
jQuery(function () { /* code */ });
```
instead
```js
jQuery(document).ready(function () { /* code */ });
```

Also see in [jQuery docs](https://api.jquery.com/ready/):

> There is also `$(document).on( "ready", handler )`, *deprecated as of jQuery 1.8 and removed in jQuery 3.0*. Note that if the DOM becomes ready before this event is attached, the handler *will not be executed*.

One more thing. You can pass jQuery object to handler with `$`:
```js
jQuery(function ($) {
    // using $ here will be safely even jQuery.noConflict() will be enabled
});
```

And one more thing. To use `$` safely in POS_LOAD section construction `jQuery(window).on('load', function () { });` can be placed into `jQuery(function ($) { });`:

```js
jQuery(function ($) {
    $(window).on('load', function () {
        // POS_LOAD scripts. Can use $
    });
    // POS_READY scripts
});
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.*
| PHP version      | 5.4-7.x
| Operating system | all

klimov-paul:
Relates #13839

SilverFire:
Related to #14338

SilverFire:
Resolved by https://github.com/yiisoft/yii2/commit/3a8feb17a8213362d79bfbb32dc89621c1de42f1

klimov-paul:
Complete removal of the `JQuery.ready()` looks pretty odd: this method still exists at jQuery library and I do not see any note about its been deprecated.
Would not its usage removal break particular JS code?

alexantr:
My mistake. Only `.on("ready", handler)` deprecated and removed in 3.0, not `.ready(handler)`. 
But anyway short syntax `$(handler)` is equivalent:
> jQuery offers several ways to attach a function that will run when the DOM is ready. All of the following syntaxes are equivalent:
> `$( handler )`
> `$( document ).ready( handler )`
> `$( "document" ).ready( handler )`
> `$( "img" ).ready( handler )`
> `$().ready( handler )`

gpcaretti:
@alexantr 
> My mistake. Only .on("ready", handler) deprecated and removed in 3.0, 
> not .ready(handler). But anyway short syntax $(handler) is equivalent:
> $( handler )
[...]

PR #15087, opened 2017-11-03
Fixes #15086 - window load inside ready state will not be triggered

### LINKED ISSUE #15086: jQuery 3 - window load inside ready state will not be triggered
In #14620 I proposed to put code from POS_LOAD section to jQuery ready state. But this not working in in jQuery 3. See https://github.com/jquery/jquery/issues/3194

Need to replace `jQuery(function ($) { $(window).on('load', function () { /*code*/ }); });` with `jQuery(window).on('load', function () { /*code*/ });`

Sorry for this.

Register some code in POS_LOAD section.

Code must be executed when page content fully loaded.

### What do you get instead?

Code doesn't run at all.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13
| PHP version      | *
| Operating system | *

alexantr:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no (for 7.1 stopped)
| Fixed issues  | #15086

### DIFF
--- a/framework/web/View.php
+++ b/framework/web/View.php
@@ -604,7 +604,7 @@ class View extends \yii\base\View
                 $lines[] = Html::script($js, ['type' => 'text/javascript']);
             }
             if (!empty($this->js[self::POS_LOAD])) {
-                $js = "jQuery(function ($) {\n$(window).on('load', function () {\n" . implode("\n", $this->js[self::POS_LOAD]) . "\n});\n});";
+                $js = "jQuery(window).on('load', function () {\n" . implode("\n", $this->js[self::POS_LOAD]) . "\n});";
                 $lines[] = Html::script($js, ['type' => 'text/javascript']);
             }
         }


framework/web/View.php
```diff
@@ -604,7 +604,7 @@ protected function renderBodyEndHtml($ajaxMode)
                 $lines[] = Html::script($js, ['type' => 'text/javascript']);
             }
             if (!empty($this->js[self::POS_LOAD])) {
-                $js = "jQuery(function ($) {\n$(window).on('load', function () {\n" . implode("\n", $this->js[self::POS_LOAD]) . "\n});\n});";
+                $js = "jQuery(window).on('load', function () {\n" . implode("\n", $this->js[self::POS_LOAD]) . "\n});";
```
samdark: Hmm... Shouldn't it be just `"jQuery(function($) {\n" . implode("\n", $this->js[self::POS_LOAD]) . "\n});"`?

alexantr: jQuery(function($) { }); - when DOM is ready
jQuery(window).on('load', function() { }); - when content is fully loaded

samdark: Ah, OK.

samdark (approved):
Merge master in, add changelog and it's done.

framework/web/View.php
cebe: I see that we are nesting the calls unnecessarily here but as far as I understand DOM ready event comes before window load event. if we register the handler inside of the DOM ready event, why does it have no effect?

alexantr: In version 3 the ready handler fires asynchronously.

You can check code:

```javascript
jQuery(function ($) {
  console.log('ready');
  $(window).on('load', function () {
    console.log('load in ready');
  });
});
jQuery(window).on('load', function () {
  console.log('load');
});
```

With jQuery 2.2.4 you will see:

```
ready
load
load in ready
```

With jQuery 3.2.1 you will see:

```
load
ready
```

No "load in ready" anymore.

https://github.com/jquery/jquery/issues/3194#issuecomment-228556922
https://github.com/jquery/jquery/issues/3197

alexantr: I just wanted to make `$` working even jQuery.noConflict() is enabled. But missed this great change in jQuery API. The best way to fix this bug - rollback my proposal in #14620

SilverFire (approved):

SilverFire:
Merged: https://github.com/yiisoft/yii2/commit/22467864834b78236469882e6bd169144995161e

SilverFire:
Thank you!

Issue #15092, opened 2017-11-04
[Enh] Yii::createObject()


lav45:
Invite I need to connect the component
```php
class CDN extends \yii\base\Component
{
    /** @var array */
    public $options;
    /** @var StorageInterface */
    private $storage;

    public function __construct(StorageInterface $storage)
    {
        $this->storage = $storage;
    }
}

class Storage implements StorageInterface
{
    /** @var string */
    public $basePath;
    /** @var TransportInterface */
    private $transport;

    public function __construct(TransportInterface $transport)
    {
        $this->transport = $transport;
    }
}
```

The first problem that will face is that you cannot pass the required parameters to the constructor when the component description.
```php
return [
    'componenets' => [
        'cdn' => [
            'class' => CDN::class,
            'options' => [],
            'storage' => [
                'class' => Storage::class,
                'basePath' => '/var/data',
                'transport' => Transport::class,
            ],
        ]
    ]
];
```
To this option is earned, you will implement a recursive nested classes. 
This functionality is already implemented in the https://github.com/lav45/yii2-flysystem
https://github.com/lav45/yii2-flysystem/blob/master/src/MountManager.php#L48

vtvz:
PR #15061 solves this problem like this:
```php
[
        'cdn' => [
            'class' => CDN::class,
            'options' => [],
            'storage' => Instance::of([
                'class' => Storage::class,
                'basePath' => '/var/data',
                'transport' => Transport::class,
            ]),
        ]
    ]
```

lav45:
@vtvz you have a very interesting project but it won't work if the parameters should be passed in the constructor [@see](https://github.com/vtvz/yii2/blob/master/framework/base/BaseObject.php#L106)

my implementation allows you to add classes as a component of third-party developers (not yii2 projects)
https://github.com/lav45/yii2-flysystem#configuring
```php
return [
    'componenets' => [
        'flysystem' => [
            'class' => 'airani\flysystem\MountManager',
            's3' => [ // https://flysystem.thephpleague.com/adapter/aws-s3-v3/
                'class' => 'League\Flysystem\AwsS3v3\AwsS3Adapter',
                'client' => [
                    'class' => 'Aws\S3\S3Client',
                    'args' => [
                        'credentials' => [
                            'key' => 'chvw2NSdzDV2aKbsJg8ypX',
                            'secret' => 'a4aeZBK5ZRTsDTCqBVFGZpL9MEeBMYjDHS3Cx55YazDD'
[...]

PR #15093, opened 2017-11-04
Recursive object creation in Yii::createObject()

lav45:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no
| issues  | #15092

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -344,7 +344,15 @@ class BaseYii
         if (is_string($type)) {
             return static::$container->get($type, $params);
         } elseif (is_array($type) && isset($type['class'])) {
-            return self::recursiveCreateObject($type, $params);
+            $class = $type['class'];
+            unset($type['class']);
+
+            if (isset($type['$'])) {
+                $params = $type['$'];
+                unset($type['$']);
+            }
+
+            return static::$container->get($class, $params, $type);
         } elseif (is_callable($type, true)) {
             return static::$container->invoke($type, $params);
         } elseif (is_array($type)) {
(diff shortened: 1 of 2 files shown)


SilverFire:
Thank you for the pull request.

I do not think mixing class properties and constructor parameters is a good design idea.

1. It makes harder to distinguish property from constructor argument when you look on `Yii::createObject()` call
2. It adds more logic to object creation flow, making it (a bit) slower.
3. In gives another one more way to do the same thing, so may bring some confusions.

© [comment](https://github.com/yiisoft/yii2/issues/15092#issuecomment-341961110) @lav45:
> I love the components for the fact that you can reuse the same component with different settings. In this case Aws\S3\S3Client when you connect can use a different username and password.

But it is possible without making any changes. And saying about one more way, I mean this one, for instance:

```php
'components' => [
    'cdn' => [
        'class' => yiiunit\framework\CDN::class,
        'options' => ['key' => 'open', 'secret' => 'sesame']
    ],
    'anotherCdn' => [
        'class' => yiiunit\framework\CDN::class,
        'options' => ['key' => 'foo', 'secret' => 'bar']
    ]
],
'container' => [
    'definitions' => [
        yiiunit\framework\CDN::class => [
            [/* class parameters config */],
            [
                Instance::of(remote\Storage::class) 
            ]
        ],
        remote\Storage::class => [
            [/* class parameters config */],
            [
                Instance::of(aws\Transport::class)
            ]
        ]
    ]
]
```

lav45:
@SilverFire 
> I do not think mixing class properties and constructor parameters is a good design idea.
> 1. It makes harder to distinguish property from constructor argument when you look on Yii::createObject() call

We can use a wildcard symbol to highlight properties that should be passed to the constructor. 
For example,"$"

> 2. It adds more logic to object creation flow, making it (a bit) slower.

Fixed in https://github.com/yiisoft/yii2/pull/15093/commits/83d0f2c890f4d3c0e87d3741703d8a5318eb07ce

> 3. In gives another one more way to do the same thing, so may bring some confusions.

Here I agree with you. Only one method of passing parameters using an array is more often used. And it is possible to allocate as a priority, and the old version removed in the next version.

The final design will look like this
```php
'components' => [
    'cdn' => [
        'class' => 'yiiunit\framework\CDN',
        'options' => ['key' => 'val'],
        '$' => [
            Yii::createObject([
                'class' => 'yiiunit\framework\Storage',
                'basePath' => '/var/storage',
                '$' => [
                    Instance::of('yiiunit\framework\Transport'),
                    'user',
                    'pass',
                ],
            ])
        ]
    ]
]
```

SilverFire:
?

PR #15095, opened 2017-11-05
Catch and show exceptions when applying/reventing migrations

Vovan-VE:
Catch all `Throwable`s when applying or reverting migration, output it to STDOUT and return `false` as when migration itself returns `false`.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes?
| Breaks BC?    | no
| Tests pass?   | yes?
| Fixed issues  | #14169
| Obsoletes | PR #14262

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -12,6 +12,7 @@ use yii\base\BaseObject;
 use yii\base\InvalidConfigException;
 use yii\base\NotSupportedException;
 use yii\console\Controller;
+use yii\console\ErrorHandler;
 use yii\console\Exception;
 use yii\console\ExitCode;
 use yii\db\MigrationInterface;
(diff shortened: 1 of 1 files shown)


rob006:
Why `stdout`? `stderr` looks more appropriate for errors.

Vovan-VE:
I use `stdout` here because `stderr` is not used neither in `BaseMigrateController` nor in `MigrateController`.

schmunk42:
> May eating all exceptions and errors be a BC break here? Should the changes be moved from BaseMigrationController to MigrationController?

Actually yes, if you rely on the output of a certain stream.

But thinking more about this, I am against such a change, because I see no valid reason why those exceptions should be omitted. If your migrations fail, you're in trouble, why continue in such a case?

If ... I'd add this via an additional param, like `-c` for `--continue` or similar.

Vovan-VE:
The goal is not to continue applying rest migrations after failed one. The goal is just to show summary "applied 7 from 10 migrations" in case of error, which is now showing currently. It can be done in different ways.

samdark:
Won't be merged into 2.0.

Issue #14916, opened 2017-10-05
`Query::each()` key started with 1 instead of 0 (no `indexBy`)


Vovan-VE:
My API in a project produce data with multiple arrays of different descendants of `ActiveRecord`. So I cannot feed just one `DataProviderInterface` to `Json::encode()` (via `Serializer`), because `Json::encode()` itself does not expand `DataProviderInterface` to models under it.

So, I feed to `Json::encode()` array with a number of `Traversable`s somewhere inside, because `Json::encode()` does `foreach` over them.

The problem is that `$query->each()` (without `indexBy`) starts keys with `1` instead of `0`. 
So, `Json::encode()` will produce object `{"1": {...}, "2": {...}, ...}`
instead of array `[{...}, {...}, ...]`

This caused by line https://github.com/yiisoft/yii2/blob/2.0.12/framework/db/BatchQueryResult.php#L121

### What steps will reproduce the problem?

```php
echo Json::encode([
    'foo' => [
        'bar' => Customer::find()
            ->limit(3)
            ->each(),
    ],
]);
```

### What is the expected result?

```json
{"foo":{"bar":[{...},{...},{...}]}}
```

### What do you get instead?

```json
{"foo":{"bar":{"1":{...},"2":{...},"3":{...}}}}
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | any (actually 7.0.22)

samdark:
Why do you need `each()` in this case?

Vovan-VE:
I want pass it as `Traversable` to `Json::encode()` somewhere inside. Current solution is generator:

```php
public function getCustomers()
{
    foreach ($query->each() as $model) {
        yield $model;
    }
}

echo Json::encode([
    'foo' => [
        'bar' => $foo->getCustomers(),
    ],
]);
```

samdark:
Can't you pass `$query->all()` instead?

Vovan-VE:
Yes, but... Stop. `Json::encode()` will do it anyway. Ok, there is no need almost for all of this. The only difference is a number of loops performed.

samdark:
OK. The indexing `each()` does seems to be non-natural but I'm afraid we can't easily change it. Need to check if it's covered by any existing tests or mentioned in docs.

samdark:
If we'll do change it, UPGRADE notes should be written.

Vovan-VE:
[...]

PR #15102, opened 2017-11-06
Fix: Query::each() without `indexBy()` zero based keys

Vovan-VE:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes?
| Tests pass?   | yes
| Fixed issues  | #14916

Existing tests did not cover subject case, so it passed before.

- [ ] Add info to CHANGELOG for appropriate dev version
- [ ] Add notes to UPGRADE

### DIFF
--- a/framework/db/BatchQueryResult.php
+++ b/framework/db/BatchQueryResult.php
@@ -118,7 +118,11 @@ class BatchQueryResult extends BaseObject implements \Iterator
             if ($this->query->indexBy !== null) {
                 $this->_key = key($this->_batch);
             } elseif (key($this->_batch) !== null) {
-                $this->_key++;
+                if (null === $this->_key) {
+                    $this->_key = 0;
+                } else {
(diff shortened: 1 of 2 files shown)


framework/db/BatchQueryResult.php
```diff
@@ -118,7 +118,11 @@ public function next()
             if ($this->query->indexBy !== null) {
                 $this->_key = key($this->_batch);
             } elseif (key($this->_batch) !== null) {
-                $this->_key++;
+                if (null === $this->_key) {
```
berosoboy: I think you could put the same code of line 131 here. (Only for consistency)
`$this->_key = $this->_key === null ? 0 : $this->_key + 1;`

klimov-paul:
Resolved by commit 00afe6a4b8d80c8ee475f831a657f0adad77dbee
Thank you.

Issue #7309, opened 2015-02-16
Add ability to render view from string


RomeroMsk:
I can't find the way to render a view from string (not from file). I want to use template engine (Twig or Smarty) to render texts from templates stored in DB. For example, store email message template in DB table and generate email body from it on the fly.

klimov-paul:
Use old good `eval()`.

nineinchnick:
@klimov-paul why would you ever recommend `eval()`? @RomeroMsk you're asking what to use instead of `render()` in a controllers action?

RomeroMsk:
@klimov-paul, are you serious?
@nineinchnick, yes I need to user `render()` method (or something like `renderString()`) with string content instead of view file path.

nineinchnick:
I've updated my response above, use [renderContent](http://www.yiiframework.com/doc-2.0/yii-base-controller.html#renderContent%28%29-detail).

klimov-paul:
> I've updated my response above, use renderContent.

I can not see how `renderContent()` can help here. It will just pass the PHP code as a plain string into the layout file.

As far as I understand you with to evaludate some string, stored in DB, as a view file content. There is no way for it instead of using `eval()`.
As far as I remember Smarty has built in `eval` function (or plugin), which allows evalution of a string as Smarty template content. As for Twig - I am unsure. For plain PHP view - native `eval()` is only option.

RomeroMsk:
@nineinchnick, can you show an example of how to use this with Twig and not in controller context? Let's say we have `$template = 'Hello, {{ username }}!'` and want to put render result into `$result`, using `['username' => 'Test']` as parameters.

RomeroMsk:
@klimov-paul, I've created this issue to ask for an ability to use stirng content instead of files in ViewRenderer. I don't like your workaround.

klimov-paul:
This is not walk around - this is the only solution. Of course we can instoduce another method in the `View` class, but it will do just the same thing.
For smarty you can use following:
http://www.smarty.net/docs/en/language.function.eval.tpl
For Twig - I am don't know.

I suppose for your case a special tool should be used like following:
https://mustache.github.io/

nineinchnick:
@RomeroMsk I think this does not concern Yii at all, you're asking how to process a Twig template. See [twig docs](http://twig.sensiolabs.org/doc/api.html).

RomeroMsk:
[...]

PR #15116, opened 2017-11-07
Implement rendering of PHP strings

Deele:
Stores string in temporary file, renders and returns result using existing renderPhpFile function.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #6252,#7309

- [x] Code
- [ ] Tests
- [x] Documentation (PHPDoc)
- [ ] Documentation (Guide)
- [ ] Example of use case

The purpose of this feature is to create a way to utilize Yii framework View class to work with string-based PHP view code that are stored in data base or generated dynamically.

This can be used for CMS editable PHP templates for widgets, templated e-mail messages and templated PDF documents (generated from HTML code).

This implementation should be tested how it performs in console environment, where `Yii::$app->view` is a bit different than in web environment.

### DIFF
--- a/framework/base/View.php
+++ b/framework/base/View.php
@@ -349,6 +349,29 @@ class View extends Component
     }
 
     /**
+     * Renders a view string as a PHP script.
+     *
+     * This method treats the view string as a PHP script and evaluates it.
+     * It extracts the given parameters and makes them available in the view string.
+     * The method captures the output of the view string and returns it as a rendered string.
+     *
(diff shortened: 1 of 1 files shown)


mikk150:
what about twig and other renderers?

samdark:
Moved to yiisoft/view.

PR #15119, opened 2017-11-07
Support DI container constructor injection

Slayer911:
Object create directly, without DI injection resolving. I propose resolve this by "Instance::ensure" object creating.
The same bug i find in (but I'm not sure that there is this problem now exist in follow files, since in the new version you use the model::instance() method) :
yiisoft/yii2/grid/DataColumn.php - function getHeaderCellLabel
yiisoft/yii2/data/ActiveDataProvider.php - function setSort

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Tests pass?   | no

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -1179,7 +1179,9 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      */
     public static function instantiate($row)
     {
-        return new static();
+        /** @var static $instance */
+        $instance = Instance::ensure(static::className());
+        return $instance;
     }
 
     /**


samdark:
While it may make sense for widgets, for AR it's could be performance killer. What do you want to achieve with instantiating AR through DI container?

Slayer911:
My AR (Search model for admin panel) class have composite structure. For avoid creating object directly with "new", i try get their by constructor. Composite object use always, that way i don't try make lazy initialization.
At the same time, DI may resolve constructor injection by singleton. It's doesn't make performance slower

samdark:
In case of the forms it's correct but in case of regular ARs performance is definitely affected. I suggest you to override `instantiate()` in your class.

PR #15143, opened 2017-11-11
Fixed DB schema cache did not honor table prefixes (fixes #15117)

### LINKED ISSUE #15117: db/Schema::getTableMetadata() broken

Run migrate/fresh more than once.

Command should run on each call.

### What do you get instead?

On second call an exception is thrown because the migration table gets not created after truncate.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13
| PHP version      |  7.0.18
| Operating system | Debian

The error was introduced in Yii 2.0.13.

The reason is the implementation of db/Schema::getTableMetadata(). The migration table does not exist but it's schema information are still present in file cache.

This is the current implementation:

	if ($refresh || !isset($this->_tableMetadata[$name])) {
		...
	}

	if (!array_key_exists($type, $this->_tableMetadata[$name])) {
		...
	}

And it should look like this:

	if (!isset($this->_tableMetadata[$name])) {
		...
	}

	if ($refresh || !array_key_exists($type, $this->_tableMetadata[$name])) {
		...
	}

sergeymakinen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15117 

> `MigrateController::truncateDatabase()` gets table names from `Schema::getTableSchemas()`, where the `TableSchema::$name` for the migrations table is plain **`migration`** (or `{prefix}migration` if current `Connection::$tablePrefix` is `{prefix}`). The metadata for **this** table name is discarded.
> 
 > But `MigrateController::getMigrationHistory()`, which is called when migrating up, gets the metadata for `MigrateController::$migrationTable`, defined as **`{{%migration}}`** by default. If the metadata for **this** table name is cached, `Schema::getTableMetadata()` will return that regardless of the value of `$refresh` (and the actual schema state). That's what this issue is all about.
> 
> I suppose a solution would involve tuning table prefix resolution for schema caching and/or migrations.
> 
> @U-D13

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -7,6 +7,7 @@ Yii Framework 2 Change Log
 - Bug #15081: Fixed "Undefined offset: 1" in log Target (ischenko)
 - Bug #15086: Fixed jQuery onLoad event handling (alexantr)
 - Bug #15108: Fixed `yii\db\Schema::getSchemaNames()` for MSSQL and added tests for all DBMSes (sergeymakinen)
+- Bug #15117: Fixed DB schema cache did not honor table prefixes (sergeymakinen)
 
 
 2.0.13 November 03, 2017
--- a/framework/db/Schema.php
+++ b/framework/db/Schema.php
@@ -277,12 +277,13 @@ abstract class Schema extends BaseObject
      */
     public function refreshTableSchema($name)
     {
-        unset($this->_tableMetadata[$name]);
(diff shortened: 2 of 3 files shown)


samdark (approved):

rob006 (approved):

samdark:
Thank you for taking care of it.

Issue #13212, opened 2016-12-15
Database exception using DbSession


puku:
### What steps will reproduce the problem?
When I'm trying to login on the website, from time to time I'm receiving an error message.
After reloading of the page everything is ok.
`YII_DEBUG `is enabled, `YII_ENV_DEV` is set to true.

Maybe this issue is linked with [this one](https://github.com/yiisoft/yii2/pull/1812)

[Stackoverflow question](http://stackoverflow.com/questions/41158812/yii2-dbsession-exception)
### What do you get instead?

> Integrity constraint violation – yii\db\IntegrityException
> 
> SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '0' for key 'PRIMARY'
> The SQL being executed was: UPDATE `session` SET `id`=0 WHERE `id`='3u5r7gq6pqh8s0gjrusfqpmni7'
> Error Info: Array
> (
>     [0] => 23000
>     [1] => 1062
>     [2] => Duplicate entry '0' for key 'PRIMARY'
> )
> ↵
> Caused by: PDOException
> 
> SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '0' for key 'PRIMARY'
> 
> in /var/www/develop/vendor/yiisoft/yii2/db/Command.php at line 844

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.10
| PHP version      | 7.0
| Operating system | Ubuntu 14.04

bizley:
I think `DbSession` should check if [$newID](https://github.com/yiisoft/yii2/blob/master/framework/web/DbSession.php#L104) is properly set before trying to [update DB with it](https://github.com/yiisoft/yii2/blob/master/framework/web/DbSession.php#L114).

dynasource:
looks like it. If session_regenerate_id fails to generate an ID, its no use to update the DB with it.

bizley:
Should the old session row be deleted in case the new ID can not be generated? For something like to prevent loosing the old data in race condition case?

I've tested `DbSession` simulating loosing ID and deleting session row in that case - everything looks ok.

yanhuixie:
Same issue with Redis session implementation.
session_regenerate_id(): Session object destruction failed. ID: user (path: )
/vendor/yiisoft/yii2/web/Session.php L282
Yii2 2.0.10

andrewnester:
[...]

Issue #13879, opened 2017-03-30
Upsert support for query builder


samdark:
Currently there's no way to atomically upsert a record i.e. try to insert a record, if uniqueness constraint failed, update it.

Suggested signature:

```php
public function upsert($table, $columns, &$params);
```

- MySQL supports it via `INSERT ... ON DUPLICATE KEY UPDATE`: https://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html
- In PostgreSQL it could be achieved multiple ways: https://dzone.com/articles/performing-upsert-update-or-insert-with-postgresql
- SQLite has `INSERT ON CONFLICT REPLACE`: http://sqlite.org/lang_conflict.html
- MSSQL, Oracle. Complicated but doable via MERGE + some safeguards: http://weblogs.sqlteam.com/dang/archive/2009/01/31/UPSERT-Race-Condition-With-MERGE.aspx and http://michaeljswart.com/2011/09/mythbusting-concurrent-updateinsert-solutions/.

More analysis from PostgreSQL: https://wiki.postgresql.org/wiki/UPSERT

samdark:
Could help solving issues like https://github.com/yiisoft/yii2/issues/11401

Vovan-VE:
Can and will it support things like MySQL's ``VALUES(`col`)``?

```mysql
INSERT INTO `table` (..., `col`)
VALUES (..., 1)
ON DUPLICATE KEY UPDATE `col` = `col` + VALUES(`col`);
```

Does it violate AR concept?
Without this feature stupid MySQL cause stupid problems. One of underlying reasons is that it cannot lock phantom rows. I did try two scenarios to perform `INSERT or UPDATE;` with non-AI PK and both does hit deadlock due to this reason. I want again to throw MySQL into trash can.

samdark:
What's `VALUES(`col`)`?

acs-ferreira:
Something like this:

`INSERT INTO table (a,b,c) VALUES (1,2,3),(4,5,6)
    ON DUPLICATE KEY UPDATE c=VALUES(a)+VALUES(b);`

https://mariadb.com/kb/en/mariadb/values/

Vovan-VE:
Sorry, I want to ask about referencing to old value:

```mysql
ON DUPLICATE KEY UPDATE `col` = `col` + ...
```
Without this ability the subject feature is not so cool. Without this I still need to combine SELECT with INSERT or UPDATE.

[...]

PR #15146, opened 2017-11-11
Fix DbSession::writeSession race condition.

ddinchev:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #13212

It has been previously reported #13212 There is a race condition in `DbSession::writeSession()`. 

```
Integrity constraint violation: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'kupbup8mepuaco3dr5qdretsde' for key 'PRIMARY'
The SQL being executed was: INSERT INTO `YiiSession` (`data`, `id`, `expire`) VALUES ('__flash|a:0:{}', 'kupbup8mepuaco3dr5qdretsde', 1510386856)
```

After updating to `2.0.13`, I get around a dozen of those errors per day (I use MySQL 5.7), on a low traffic website that has less than a thousand visits a day. The session table has less than 1000 rows, the garbage collector is working correctly. I guess prior to `2.0.13` the errors were not exposed through the standard error logger - now this is fixed by #14908 

I'm not entirely sure why would the generated session id collide so often. My solution does the check if session id already exists in the storage in a transaction. Performance impact should be negligible as there should be only one session read/write per request.

### DIFF
--- a/framework/web/DbSession.php
+++ b/framework/web/DbSession.php
@@ -108,28 +108,30 @@ class DbSession extends MultiFieldSession
             return;
         }
 
-        $query = new Query();
-        $row = $query->from($this->sessionTable)
-            ->where(['id' => $oldID])
-            ->createCommand($this->db)
-            ->queryOne();
-        if ($row !== false) {
-            if ($deleteOldSession) {
-                $this->db->createCommand()
-                    ->update($this->sessionTable, ['id' => $newID], ['id' => $oldID])
-                    ->execute();
+        $this->db->transaction(function () use ($deleteOldSession, $oldID, $newID) {
+            $query = new Query();
+            $row = $query->from($this->sessionTable)
+                ->where(['id' => $oldID])
+                ->createCommand($this->db)
+                ->queryOne();
+            if ($row !== false) {
+                if ($deleteOldSession) {
+                    $this->db->createCommand()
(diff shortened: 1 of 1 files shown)


ddinchev:
Based on the DBMS, this might not prevent race conditions. However, I'm not sure if we can write it in DBMS-independent SQL. The mechanisms to lock tables are different in all databases and approaches like `INSERT .. ON DUPLICATE KEY UPDATE` are also DBMS-specific.

samdark:
Related to https://github.com/yiisoft/yii2/issues/13879

rob006:
I'm afraid that this will not change anything. Opening transaction will not prevent other processes from reading this table -  you will still get:
1. Request A opens transaction and reads session table.
2. Request B opens transaction and reads session table.
3. Request A inserts new record to session table.
4. Request B waits till request A commit transaction and tries to save the same record again.

ddinchev:
@rob006 you are correct, I realized that soon after opening the proposed MR. I'm a bit concerned however about that we actually hit this at all. The session id shouldn't collide so often.

njasm:
this still is an issues, even if @ddinchev approach is not the solution, we should have this issue reopened and try to find a solution for it..

This issue concerns me since i'm finding it one or two time a day and still in development environment!

PR #7664, opened 2015-03-11
7593 new AtLeastValidator

sdlins:

PR #15157, opened 2017-11-15
new AtLeastValidator

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7664

Since I don't have permission to change PR #7664, let's finalize it in this PR.


framework/validators/AtLeastOneValidator.php
```diff
@@ -0,0 +1,106 @@
...
+        if ($this->message === null) {
+            $this->message = Yii::t('yii', 'Please fill at least {min} of {attributesList}.');
+        }
+    }
+
+    public function validateAttribute($model, $attribute)
+    {
+        $value = $model->$attribute;
+
+        $attributes = array_merge((array)$attribute, (array)$this->alternativeAttributes);
```
dmirogin: This will consist of duplicate attributes.
~~~php
$a = ['attr1'];
$b = ['attr1', 'attr3'];
$c = array_merge($a, $b); // ['attr1', 'attr1', 'attr3']
~~~

framework/validators/AtLeastOneValidator.php
```diff
@@ -0,0 +1,106 @@
...
+        parent::init();
+        if ($this->isEmpty($this->alternativeAttributes)) {
+            throw new InvalidConfigException('Param "alternativeAttributes" can not be null');
+        }
+        if ($this->message === null) {
+            $this->message = Yii::t('yii', 'Please fill at least {min} of {attributesList}.');
+        }
+    }
+
+    public function validateAttribute($model, $attribute)
```
dmirogin: It needs dockblok
~~~php
/**
      * @inheritdoc
*/
~~~

framework/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Yii Framework 2 Change Log
 - Enh #15135: Automatic completion for help in bash and zsh (Valkeru)
 - Enh #14662: Added support for custom `Content-Type` specification to `yii\web\JsonResponseFormatter` (Kolyunya)
 - Enh #14568: Refactored migration templates to use `safeUp()` and `safeDown()` methods (Kolyunya)
+- Enh #7593: Created a new core validator AtLeastValidator (slinstj)
```
dmirogin: Add your name to participants. 😄

framework/validators/AtLeastOneValidator.php
developeruz: It is ok, because 
``['attr1', 'atLeastOne', 'alternativeAttributes' => ['attr2', 'attr3']],``
means that I need attr1 or any of alternativeAttributes. It doesn't make sense to put attr1 twice in validator

dmirogin: here
~~~php
$this->addError($model, $attribute, $this->message, [
     'attributesList' => $attributesList,
]);
~~~
error can be added twice for one attribute, and i guess we get error message twice too.

developeruz: Ok, I will add test for this part

developeruz: All error assignment is covered by [tests](https://github.com/yiisoft/yii2/pull/15157/files#diff-321dfbe7e92e4e7d446dd2f9d4354220)

framework/validators/AtLeastOneValidator.php
```diff
@@ -0,0 +1,107 @@
...
+use Yii;
+use yii\base\InvalidConfigException;
+
+/**
+ * AtLeastValidator checks if at least one of many attributes are filled.
+ *
+ * In the following example, the `attr1`, `attr2` and `attr3` attributes will
+ * be verified. If at least one of them are not filled, all will receive an error:
+ *
+ * ~~~[php]
```
samdark: It is better to use \`\`\`php.

cebe: not just better, `~~~[php]` does not work as expected.

njasm:
why the rule have this structure ?
```php
return [['attr1', 'atLeastOne', 'validateWith' => ['attr2', 'attr3'], 'errorAttributes' => ['attr2', 'attr3']]];
```
and with an importante notice to use only one attribute in the attribute place?

something in the lines of 
```php
return [['attr1', 'attr2', attr3'], 'atLeastOne', errorAttributes' => [...]];
```
wouldn't be more in line with the current rules parse in the current core validators?

am I missing something?

developeruz:
@njasm where did you find that structure is like this? In code it looks [like this](https://github.com/developeruz/yii2/blob/e590270c2eb1746e329fa5c7522447ee14b50614/framework/validators/AtLeastOneValidator.php#L22):
```
 ['attr1', 'atLeastOne', 'alternativeAttributes' => ['attr2', 'attr3'], 'min' => 2, 'errorAttributes' => ['attr1', 'attr3']]
```
If you need multiple attributes, you also can use 
```
 [['attr1', 'attr2', 'attr3'], 'atLeastOne', 'min' => 2, 'errorAttributes' => ['attr1', 'attr3']]
```
therefore it is in line with the current structure of other validators

njasm:
@developeruz sure, right where you link to.
the Docblock show's the rule structure as mentioned.

The implied question is:
- If the Validator name is "AtLeastOne", I take from granted that no attribute have more weight than any other, nad if this is true, isn't the `alternativeAttributes` property redudant?

you could 
```php
return [['att1', 'att2', ...], 'atLeastOne', ...];
```
without the need of the alternativeAttributes property, taking that you should put all attributes in the first section as an array, right?

developeruz:
@njasm I tried to get rid of `alternativeAttributes` and realized why the author of this PR added this parameter. Only one attribute can be specified in attributes parameter, otherwise, each attribute validates multiple time and got error messages multiple time as @dmirogin  assumed in the comment above. I've added exception throw to avoid such misunderstanding of using the validator.
```php
[['att1', 'att2', ...], 'atLeastOne', ...] // will throw exception
['attr1', 'atLeastOne', 'alternativeAttributes' => ['attr2', 'attr3'], ... ] //is correct configuration
```

njasm:
@developeruz  in the validation logic we could loop through th attributes defined in the validator and 
```php 
$this->getErrors($attribute [all attributes in the validator array]); 
```
and if any of "our" attributes have errors check if it contains an error message equals to our `$this->message`

pseudo-code
```php 
if (in_array($this->message, array_values($ourAttrsErrorMessages))) {
    return; // our validator message is already set! 
}
`
else run the validation logic. does it make sense to you? I even see optimization in this approach.

File: tests/framework/validators/AtLeastOneValidatorTest.php [hunk 5]
```diff
@@ -0,0 +1,142 @@
...
+    {
+        $model = new FakedValidationModel;
+
+        $val = new AtLeastOneValidator(['alternativeAttributes' => 'attr2']);
+        $model->attr1 = null;
+        $model->attr2 = null;
+        $val->validateAttribute($model, 'attr1');
+        $this->assertCount(2, $model->getErrors(), 'Model should have 2 error attributes.');
+        $this->assertCount(1, $model->getErrors('attr1'));
+
```
dynasource: can you split these test scenario's in contained methods?

framework/validators/AtLeastOneValidator.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * @inheritdoc
+     */
+    public function init()
+    {
+        parent::init();
+        if ($this->isEmpty($this->alternativeAttributes)) {
+            throw new InvalidConfigException('Param "alternativeAttributes" can not be null');
+        }
+        if (count($this->attributes) > 1) {
```
dynasource: \yii\validators\Validator::$attributes can be a string according to its phpdoc.
When using `count`, this will fail on PHP 7.2

tests/framework/validators/AtLeastOneValidatorTest.php
```diff
@@ -0,0 +1,142 @@
...
+        $val = new AtLeastOneValidator([
+            'alternativeAttributes' => 'attr2',
+            'errorAttributes'=>'attr1',
+        ]);
+        $model->attr1 = null;
+        $model->attr2 = null;
+        $model->clearErrors();
+        $val->validateAttribute($model, 'attr1');
+        $this->assertCount(1, $model->getErrors(), 'Model should have 1 error attribute.');
```
dynasource: I'd suggest to use assertEquals on $model->getErrors() which makes the test easier to read compared to assertCount

framework/validators/AtLeastOneValidator.php
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+namespace yii\validators;
+
+use Yii;
+use yii\base\InvalidConfigException;
+
+/**
+ * AtLeastValidator checks if at least one of many attributes are filled.
+ *
```
dynasource: AtLeastValidator? or AtLeastOneValidator? and is it about 'requiring?'....checking is a bit vague here

framework/validators/AtLeastOneValidator.php
```diff
@@ -0,0 +1,104 @@
...
+ * The additional ones must go into [[alternativeAttributes]] param.
+ *
+ * The attributes that will receive the error message can be specified into [[errorAttributes]]
+ * param. If none is specified, all envolved attributes will receive the error.
+ *
+ * @author Sidney Lins <slinstj@gmail.com>
+ * @since 2.0.14
+ */
+class AtLeastOneValidator extends Validator
+{
```
dynasource: I'd rename this class to something more specific. Something like `RequireGroupValidator`

dynasource (changes requested):

framework/validators/AtLeastOneValidator.php
SilverFire: I am OK with this naming. It is simple enough

framework/validators/AtLeastOneValidator.php
samdark: Group validator doesn't sound clear to me. Suggests nothing about behavior.

developeruz:
@njasm assign error message isn't the only issue with multiple validations.
```php
[['att1', 'att2', ...], 'atLeastOne', ...] // will run validateAttribute method n-times
```
We don't need to run validateAttribute n-times. The results would be the same as for the first validation. Therefore in terms of performance, it is better to have `alternativeAttributes`. Do you agree with it?

@dynasource, I've made requested changes, could you please review again?

framework/validators/AtLeastOneValidator.php
dynasource: unfortunately it is ambiguous. AtLeast could also point to a Validator which checks if a value is `at least` a certain value....

framework/validators/AtLeastOneValidator.php
dynasource: its says someting about a group of attributes that are required...

njasm:
@developeruz I undestand your argument, but I'm still not convinced of it, and I still do not agree with the current implementation or feature.

Let me tell you that English is not my native language and if I sound too harsh it was never my intention. Please accept my apologies upfront.

1) About performance your are right, but I've never talked about performance nor speed of execution. I've talked about optimization at the design level. (be able to exploit the already infrastructure for validators that are in place, and support the full spectrum of the Validators features [server side, client side] with minimum efford, maintainability and readability.

running n-times for n attributes that are in the rule, is something that I expect, If i didn't want it i wouldn't add the attribute in the first place.

How do you envision the client side validation of this validator? sure it can happen, but i believe we will be putting DRY to the trash can, cos i only see rebinding on change events to the "otherAttributes", something that the framework already gives me.

Second is consistency, allow me to explain:
We're talking about validators, we nowday have validators that validate data and validators that mutate state?! Without many discussion I believe that DevaultValidator had been a better DefaultValueBehavior or or something like that...
With current implementation of current validator I feel we're giving another stab on consistency, this is the only one that cannot have more then one attribute in the attributes section of the rules...weird.

with this implementation, i still believe that 'alternativeAttributes' are redudant, i still believe we'll have a not finished validator cos we don't have client side code for it, and if we have the implementation of the client side, it will be a "workaround" against the already in place infrastructure of yii2, instead of working with it.

my pseudo-code suggestion will run N-Times the number of attributes, but keeps consistency with the syntax in place, in term of performance speed, will not be that slow, we're talking milliseconds here. and if the error is already will return earlier. Also the JS implementation against the server side one would be almost like a one-to-one port, and easily we could get the full feature supported for the validator.

With this, I rest my case and sorry for my long post, And I'll also embrace any descision you and other core developers take. :)

developeruz:
> running n-times for n attributes that are in the rule, is something that I expect, If i didn't want it i wouldn't add the attribute in the first place

@njasm, If I remove `alternativeAttributes` you wouldn't have a chose

About consistency: we already have  [CompareValidator](https://github.com/yiisoft/yii2/blob/fd56b8d4ed767a1ec542fca96e380ad78d9ab55f/framework/validators/CompareValidator.php#L143) that does not accept an array as `attributes`

How about removing the restriction to have an array as `attributes` (add checks to avoid multiple error messages) but keep `alternativeAttributes` as an option to avoid multiple validations?

njasm:
@developeruz thank you for your openness about this issue.

> How about removing the restriction to have an array as attributes (add checks to avoid multiple error messages) but keep alternativeAttributes as an option to avoid multiple validations?

It sound's good to me. We should try to implement that and see how it goes.

developeruz:
Done, @njasm could you please check again how it looks now?

framework/validators/AtLeastValidator.php
```diff
@@ -72,33 +63,42 @@ class AtLeastValidator extends Validator
...
     {
         parent::init();
-        if (empty($this->with)) {
-            throw new InvalidConfigException('Param "with" can not be null');
+        if ($this->isEmpty($this->alternativeAttributes)) {
+            $this->alternativeAttributes = (array)$this->attributes;
+        } else {
+            $this->alternativeAttributes = array_merge((array)$this->attributes, (array)$this->alternativeAttributes);
         }
+        $this->errorAttributes = $this->isEmpty($this->errorAttributes) ? (array)$this->alternativeAttributes : (array)$this->errorAttributes;
```
njasm: according to $errorAtrributes definition, this line should be an array_merge between attributes and alternativeAttributes when errorAttributes is empty.

njasm:
@developeruz  looks good. thanks.

Also, I would suggest to check in model, if `$this->message` is already set to any of the attributes found in the attrbutes array for rules like

```php
return [['attr1', 'attr2', attr3', ..] 'atLeasOne, ...];
```
in this scenario, we did not set any `alternative attributes`, and if the error message is set on any of them, or on the defined `errorAttrbitues` [if set] we could return earlier? (this will probably happen when validating the second or plus attributes of the rules only.

framework/validators/AtLeastValidator.php
developeruz: [`alternativeAttributes`](https://github.com/developeruz/yii2/blob/e30ede98ce1c71c13d31f6acdb79bab33362e968/framework/validators/AtLeastValidator.php#L69) is contain array_merge between `attributes` and `alternativeAttributes` (couple lines above)

developeruz:
To define if the error message is set, I have to get `AttributeLabels`, so I can't skip first `foreach`.  All `errorAttributes` should get an error. So the second `foreach` is also required. @njasm, I did not get how can I return earlier?

njasm:
@developeruz  something along this lines...

```php
foreach ($this->alternativeAttributes as $attribute) {
	$value = $model->$attribute;
	$attributesListLabels[] = '"' . $model->getAttributeLabel($attribute) . '"';
	$filled += !$this->isEmpty($value) ? 1 : 0;

	// we still didn't got to a valid state and
	// our error message is already set on this attribute.
	if ($filled < $this->min && in_array($this->message, $model->getErrors($attribute))) {
		return false;
	}
}

// rest of current implementation...
if ($filled >= $this->min) {
	return true;
}
```
what do you think?

framework/validators/AtLeastValidator.php
njasm: 👍 missed that. my bad.

developeruz:
The problem is that `$model` has an only formatted error message like `Please fill at least 2 of A, B, C`. I have to loop through all attributes to have `A, B, C` for error message before I can check if the model has this error. 
I believe that current implementation is the most optimized one. @njasm, thanks a lot for your help and your commitment to make this PR better 👍

njasm:
@developeruz  👍

samdark (approved):

SilverFire:
Will apply changes.

SilverFire (changes requested):
See https://github.com/developeruz/yii2/pull/1

SilverFire (changes requested):
What about a client-side validation? Should be easy to do

dynasource:
I still dont like the name because of the ambiguity. 'At least' can be confused with validating numbers.

samdark:
Any suggestions to name it better?

rob006:
How about name it to `RequireAny` and use syntax similar to [Apache](https://httpd.apache.org/docs/2.4/mod/mod_authz_core.html#requireany):

```php
return [
    // checks if every category ID is an integer
    ['attributeName', 'any', 'rules' => [
         'attributeName' => ['required'],
         'secondAttributeName' => ['required'],
         'thirdAttributeName' => ['integer', 'min' => 1],
    ]],
]
```

It will be a little more complicated to define "require at least 2 non-empty attributes" but it will be much more flexible - you will be able combine any type of rules into group and create alternative paths to fill form.

SilverFire:
Yes, looks more reusabe.

developeruz:
@SilverFire, client-side doesn't seem to me to be easy to implement. All client-side validators, that we have now, check only one attribute value while this validator has to check all attributes and alternative attributes and add error only to `errorAttributes`. Do you have any suggestion for the client-side validation?

@rob006 I agree that it looks good but it doesn't look similar to how all other validators are set up. 
`` 'thirdAttributeName' => ['integer', 'min' => 1],``
means for me that I could use any of validators inside this `RequireAny` validator? Cool, but not doable at least in 2.0.x

@dynasource, do you like `RequireAny` as for naming  and configuration like: 
 `[['att1', 'att2', ...], 'any', 'min' => 2] ` ?

rob006:
> Cool, but not doable at least in 2.0.x

Why? You should be able to use `DynamicModel` - run validation on separate model and proxy results to real one. `RequireAny` is basically a wrapper for separate `rules()` definitions.

developeruz:
Because if I am able to configure one validator as
```
 ['attributeName', 'any', 'rules' => [
         'attributeName' => ['required'],
         'secondAttributeName' => ['required'],
         'thirdAttributeName' => ['integer', 'min' => 1],
    ]],
```
I would expect that I can do the same with all other
```
['attributeName', 'required', 'rules' => [
        'attributeName' => ['integer', 'min' => 1],
        'attributeName' => ['in', 'range' => range(1, 10, 1)],
    ]],
```
I don't think that this particular validator should have additional complexity in the logic. RequireAny checks only that any set of `min` attributes is not empty. It is the same with 

```
['attributeName', 'required'], 
OR
['secondAttributeName', 'required'],
OR 
['thirdAttributeName', 'required'],
```
and any combination of this rules (min N of them is required). All additional checks can be checked in separate validation rules, as it is for all other validators.

PR #9473, opened 2015-08-19
HTTP Methods can prefix a pattern in yii\web\GroupUrlRule like in yii\web\UrlManager

schojniak:
#7996

Issue #7996, opened 2015-04-06
GroupUrlRule unexpected behavior


akireikin:
This does not work:

```php
$app->get('urlManager')->rules[] = new GroupUrlRule([
    'prefix' => 'file',
    'rules' => [
        'POST document' => 'document/create',
    ],
]);
```

This works:

```php
$app->get('urlManager')->rules[] = new GroupUrlRule([
    'prefix' => 'file',
    'rules' => [
        [
            'pattern' => 'document',
            'verb' => 'POST',
            'route' => 'document/create',
        ],
    ],
]);
```

samdark:
By doesn't work you mean 404 when parsing it or inability to create URL?

mdmunir:
@samdark 
verb pattern not work for `yii\web\GroupUrlRule` (like at `yii\rest\UrlRule`).

schojniak:
I fixed this in #9473 pull request.

PR #15166, opened 2017-11-16
Short syntax for verb in GroupUrlRule

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7996

Duplicate of #9473 because the author is inactive.


framework/web/GroupUrlRule.php
```diff
@@ -90,9 +90,16 @@ protected function createRules()
         $rules = [];
         foreach ($this->rules as $key => $rule) {
             if (!is_array($rule)) {
+                $verbs = 'GET|HEAD|POST|PUT|PATCH|DELETE|OPTIONS';
+                $verb = null;
+                if (preg_match("/^((?:($verbs),)*($verbs))\\s+(.*)$/", $key, $matches)) {
```
samdark: There's no need to capture what's not used:

```
^((?:(?:$verbs),)*(?:$verbs))\s+(.*)$
```

Then verb will be in $matches at index 1 and the rest at 2.

samdark (approved):
Overall looks good. Regex could be tweaked a bit.

framework/web/GroupUrlRule.php
developeruz: Changed, thank you

samdark:
Merged. Thank you!

PR #15177, opened 2017-11-17
additional headers for real user IP

nicdnepr:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no


framework/web/Request.php
```diff
@@ -234,6 +234,9 @@ class Request extends \yii\base\Request
      */
     public $ipHeaders = [
         'X-Forwarded-For',
+        'cf-connecting-ip',
```
cebe: can you give a source for this one, why do you think it is relevant?

framework/web/Request.php
```diff
@@ -234,6 +234,9 @@ class Request extends \yii\base\Request
...
+        'HTTP_X_REAL_IP',
+        'HTTP_X_FORWARDED_FOR',
```
cebe: this is how they show up in `$_SERVER`, but the name of the header field is `X-Forwared-For` and  `X-Real-IP`.
Can you give a source or explaination for `X-Real-IP`?

cebe (changes requested):

framework/web/Request.php
nicdnepr: It's for Cloudflare https://support.cloudflare.com/hc/en-us/articles/200170986-How-does-Cloudflare-handle-HTTP-Request-headers-

framework/web/Request.php
nicdnepr: `HTTP_X_REAL_IP` and `HTTP_X_FORWARDED_FOR` headers from nginx, which in frontend
Not sure if header `HTTP_X_REAL_IP` or `X-REAL-IP` depends on `nginx` settings, in my server nginx+apache I get headers `HTTP_X_REAL_IP` and `HTTP_X_FORWARDED_FOR` with real user IP

framework/web/Request.php
cebe: As I read that page, `cf-connecting-ip` should be preferred over `X-Forwarded-For` because cloudflare will not override an existing `X-Forwarded-For` header but rather just append the value of `cf-connecting-ip`. That means `cf-connecting-ip` should be added before `X-Forwarded-For` in this array. Also `cf-connecting-ip` must be added to the list of `$secureHeaders` property.

nicdnepr: maybe yes, I just want to show another ways how to get real user IP

nicdnepr:
I didn't close this PR, but I opened new PR and this closed automatically

cebe:
seems you removed the commits from the branch. You should have a new branch for every pull request.

nicdnepr:
Need I create new PR for this changes? Is it useful?

cebe:
I think it is useful but a few changes need to be made, you can either create a new PR or create an issue describing what you want.

PR #15199, opened 2017-11-21
Added truncateMiddle method.

Eseperio:
This method truncates a string in the middle of it. This keeps start and end of string visible, so when you have, as an example, numbered properties you can truncate them safely. 

`StringHelper::truncateMiddle('Hello world number 2', 10)` produces `"Hello...ber 2".`

There are no tests because i don´t have learned yet to use them. If some of the team consider interesting this function and want do develop the test i´ll be thankful.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | --

### DIFF
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -141,6 +141,41 @@ class BaseStringHelper
 
         return $string;
     }
+    
+    
+    /**
+     * Truncates a string in the middle. Keeping start and end.
+     * `StringHelper::truncateMiddle('Hello world number 2', 8)` produces "Hell...er 2".
+     *
+     * This method does not support html. It will strip all tags even if length is smaller.
+     *
+     * @param string $string    The string to truncate.
+     * @param int $length       How many characters from original string to include into truncated string.
+     * @param string $separator String to append in the middle of truncated string.
+     * @param string $encoding  The charset to use, defaults to charset currently used by application.
+     * @return string the truncated string.
+     */
+    public static function truncateMiddle($string, $length, $separator = '...', $encoding = null)
+    {
+        if ($encoding === null) {
+            $encoding = Yii::$app ? Yii::$app->charset : 'UTF-8';
+        }
+
+        $string = strip_tags($string);
+
+        $strLen = mb_strlen($string, $encoding);
+
+        if ($strLen > $length) {
+
+            $partLen = floor($length / 2);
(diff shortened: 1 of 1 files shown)


samdark:
Overall idea looks OK to me.

framework/helpers/BaseStringHelper.php
```diff
@@ -141,6 +141,41 @@ public static function truncateWords($string, $count, $suffix = '...', $asHtml =
...
+     * @param string $encoding  The charset to use, defaults to charset currently used by application.
+     * @return string the truncated string.
+     */
+    public static function truncateMiddle($string, $length, $separator = '...', $encoding = null)
+    {
+        if ($encoding === null) {
+            $encoding = Yii::$app ? Yii::$app->charset : 'UTF-8';
+        }
+
+        $string = strip_tags($string);
```
samdark: This doesn't seem to be right. Should work with HTML as well.

Eseperio: I cant find a way to do this without breaking BC, because method `truncateHTML` does not use a flag to determine wether it should be truncated by words or by characters. Instead it uses `$encoding` value. If it is false, then words, if is null or other value it truncate by characters.

samdark: Then it's probably a good idea to introduce a separate method for HTML. Don't do it now though, we should verify that the feature is useful to more people than you and, probably, me.

Eseperio: Ok.

egorsmkv:
To me, too.

samdark:
https://twitter.com/sam_dark/status/934398410958036992

samdark:
OK. Seems to be useful enough. Let's continue...

Eseperio:
Well, then what do you think is the best about html?  Do i create a `truncateMiddleHtml` method or all the logic under the same function and an optional parameter (`asHtml`) like in other methods.?

dynasource:
IMHO, all logic should be within the same function, especially for such a rare method

samdark:
Same as existing truncate methods.

kjusupov:
Such a method would definitely be used in my project to display portion of "sensitive" info (such as full name of the member) on default view and reveal full info on explicit request (and hiding/blurring) the rest.

egorsmkv:
Also, the same method used to displaying an email.

PR #15208, opened 2017-11-23
propose setters

bscheshirwork:
This is example for discussion. Not for merge.

This is not a full solution. 
For example - the message will not change if set value second time
(from not-null value to not-null value)
message still contain old value representation.

Another way
This can be avoid with addition logic 
```php
    public function setMin($min)
    {
        $message = strtr($this->tooSmall, [$this->_min => '{min}']);
        $this->_min = $min;
        $this->init();
        $this->tooSmall = strtr($message, ['{min}' => $this->_min]);
    }
```

Another way - add note to guides and PHPDocs

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes/no
| Breaks BC?    | yes: some people can use current logic
| Tests pass?   | yes/no
| Fixed issues  | #15172

### DIFF
--- a/framework/validators/CompareValidator.php
+++ b/framework/validators/CompareValidator.php
@@ -28,6 +28,8 @@ use yii\helpers\Html;
  * are compared byte by byte. When comparing numbers, make sure to set the [[$type]]
  * to [[TYPE_NUMBER]] to enable numeric comparison.
  *
+ * @property string $operator
+ *
  * @author Qiang Xue <qiang.xue@gmail.com>
  * @since 2.0
  */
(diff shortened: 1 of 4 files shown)


Renkas:
This is a wrong way to handle this issue.

I believe correct way would be to use getters for the `$this->tooSmall` etc properties that hold the error message. And remove the message instantiation from the `init` method. This way problem from the issue goes away and also messages would be instantiated at the last possible moment.

bscheshirwork:
@Renkas
prepare PR for correct way please?

Renkas:
I haven't got the time in near future. If it still needs a PR in the future then I may do so.

framework/validators/CompareValidator.php
```diff
@@ -162,6 +165,23 @@ public function validateAttribute($model, $attribute)
...
+        return $this->_operator;
+    }
+
+    /**
+     * @param string $operator
+     */
+    public function setOperator($operator)
+    {
+        $this->_operator = $operator;
+        $this->init();
```
cebe: calling init() multiple times is not expected (e.g. when overriding a class). We should avoid such behavior.

cebe (changes requested):

framework/validators/CompareValidator.php
bscheshirwork: Yes. I close this pr

because this is a example

PR #15219, opened 2017-11-24
Added `yii\filters\auth\HttpHeaderAuth`

bboure:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes


samdark:
I think we support that already via https://github.com/yiisoft/yii2/blob/master/framework/filters/auth/HttpBearerAuth.php

bboure:
@samdark HttpBearerAuth only accepts the `Authorization` header and requires the `Bearer`  "prefix".
This one allows a different header to be used.
Like this:

````bash
curl --request GET \
  --url http://api.example.com \
  --header 'x-api-key: xxxxxxxxxxxxxxxxxxxxxxx'
````

Not sure if this is standard, but a lot of api providers do this. 
Maybe, instead, we could change `HttpBearerAuth` to allow this kind of scenario ?

What do you think?

**Comment by samdark** [comment]
We can use yours as base for Bearer indeed.

**Comment by bboure** [comment]
@samdark that actually sounds like a good idea. I'll refactor in that direction.

File: framework/filters/auth/HttpHeaderAuth.php [hunk 1]
```diff
@@ -0,0 +1,66 @@
...
+ *     ];
+ * }
+ * ```
+ *
+ * The default implementation of HttpHeaderAuth uses the [[\yii\web\User::loginByAccessToken()|loginByAccessToken()]]
+ * method of the `user` application component and passes the value of the `X-Api-Key` header. This implementation is used
+ * for authenticating API clients.
+ *
+ * @author Benoît Boure <benoit.boure@gmail.com>
+ * @since 2.x.x (TBD)
```
samdark: 2.1.14

framework/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Yii Framework 2 Change Log
 - Enh #15135: Automatic completion for help in bash and zsh (Valkeru)
 - Enh #14662: Added support for custom `Content-Type` specification to `yii\web\JsonResponseFormatter` (Kolyunya)
 - Enh #14568: Refactored migration templates to use `safeUp()` and `safeDown()` methods (Kolyunya)
+- Enh: Added `yii\filters\auth\HttpHeaderAuth`
```
samdark: Issue number and github nickname should be here.

framework/filters/auth/HttpHeaderAuth.php
```diff
@@ -0,0 +1,66 @@
...
+{
+    /**
+     * @var string the HTTP header name
+     */
+    public $header = 'X-Api-Key';
+
+    /**
+     * @var string a pattern to use to extract the HTTP authentication value
+     */
+    public $pattern = null;
```
samdark: No need for `= null`. It is default.

framework/filters/auth/HttpHeaderAuth.php
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @author Benoît Boure <benoit.boure@gmail.com>
+ * @since 2.x.x (TBD)
+ */
+class HttpHeaderAuth extends AuthMethod
+{
+    /**
+     * @var string the HTTP header name
+     */
+    public $header = 'X-Api-Key';
```
samdark: Why exactly this header name?

framework/filters/auth/HttpHeaderAuth.php
```diff
@@ -0,0 +1,66 @@
...
+ *         ],
+ *     ];
+ * }
+ * ```
+ *
+ * The default implementation of HttpHeaderAuth uses the [[\yii\web\User::loginByAccessToken()|loginByAccessToken()]]
+ * method of the `user` application component and passes the value of the `X-Api-Key` header. This implementation is used
+ * for authenticating API clients.
+ *
+ * @author Benoît Boure <benoit.boure@gmail.com>
```
samdark: I think we need to add `@author Qiang Xue <qiang.xue@gmail.com>` here as well since major portion of code was moved from `HttpBearerAuth`.

bboure:
@samdark after thinking it about it, I was wondering if we couldn't just keep HttpBearerAuth and add the `$header` and `$pattern` attributes there ?

samdark:
@bboure because "bearer" implies very specific header name to be used. If it's changed then the auth isn't 
"bearer".

framework/CHANGELOG.md
bboure: There is no issue number, should I create one?

framework/filters/auth/HttpHeaderAuth.php
bboure: no real reason. The "X-" prefix seems pretty standard in custom headers

rob006 (approved):

samdark (approved):

samdark:
Merged. Thank you!

PR #15221, opened 2017-11-25
Console improvements

brandonkelly:
Made a few improvements to Yii's CLI support:

1. camelCased console command options can now optionally be specified in kebab-case. So for example if you have an option called `dumpJson`, now either `--dumpJson` or `--dump-json` will work.

   It was technically already possible for console commands to support kebab-cased options, as the option regex has always allowed `-`s. (For example we made it possible by overriding a command’s `__get()` and `__set()` methods [in a recent project](https://github.com/pixelandtonic/craftcom/blob/c801526a6efe1a185b32930b255fc13b6506127f/src/console/controllers/PackagesController.php#L23-L42).) So to ensure we don’t break BC, the code makes sure that the kebab-cased option wasn’t already in `$options` before converting it to camelCase.

   In addition, any camelCase option names are now converted to kebab-case in the `help` command outputs, e.g. `help command/path` and `help/list-action-options command/path`.

2. Added support for `--<option> <value>` and `-<alias> <value>` syntax, as an alternative to `--<option>=<value>` and `-<alias>=<value>`. Basically when looping through the `$params` in yii\console\Request::resolve(), if it gets to a raw param value that didn't have an option name, it will check whether the last param was missing a value, and if so, it overrides its assumed value (`true`) with the current value.

   Technically this is a change in behavior for yii\console\Request::resolve(), which given the options `--foo bar` would have previously returned the params `['foo' => true, 0 => 'bar']`, however that would have resulted in an `Unknown option` error in yii\console\Controller::runAction(), so I don’t think anyone will consider this a breaking change :)

3. Improved the `help/list-action-options` command output, for options without a description. Previously they would have been output as `--<option>:<option>`, which was confusing; now they will output as just `--<option>`.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -10,6 +10,9 @@ Yii Framework 2 Change Log
 - Enh #15135: Automatic completion for help in bash and zsh (Valkeru)
 - Enh #14662: Added support for custom `Content-Type` specification to `yii\web\JsonResponseFormatter` (Kolyunya)
 - Enh #14568: Refactored migration templates to use `safeUp()` and `safeDown()` methods (Kolyunya)
+- Enh #15221: Added support for specifying `--camelCase` console options in `--kebab-case` (brandonkelly)
+- Enh #15221: Added support for the `--<option> <value>` console option syntax (brandonkelly)
+- Enh #15221: Improved the `help/list-action-options` console command output for command options without a description (brandonkelly)
 
 2.0.13.1 November 14, 2017
 --------------------------
--- a/framework/console/Controller.php
+++ b/framework/console/Controller.php
@@ -12,6 +12,7 @@ use yii\base\Action;
 use yii\base\InlineAction;
 use yii\base\InvalidRouteException;
 use yii\helpers\Console;
+use yii\helpers\Inflector;
 
 /**
  * Controller is the base class of console command classes.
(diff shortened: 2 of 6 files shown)


samdark:
Could you elaborate where `#3` is used?

samdark:
`#2` syntax isn't common. It is usually either `-x y` or `--x=y`. Is it used somewhere else?

samdark:
Doing `#1` sounds good to me.

brandonkelly:
> `#2` syntax isn't common. It is usually either `-x y` or `--x=y`. Is it used somewhere else?

To be clear the PR adds support for _both_ `-x y` and `--x y` (neither were possible before, unless I’m missing something).

Off the top of my head, `curl` supports `--x y`, e.g. `curl <url> --data-binary "@<file/path>"`. The only time I’ve ever needed to use the `--x=y` syntax is with Yii actions.

samdark:
OK. `#2` makes sense as well.

brandonkelly:
> Could you elaborate where `#3` is used?

Like everting else in the `help` controller, I assume just by anyone who wants to look up info about a command? In this case info about the available arguments. Not entirely sure why the command exists in the first place but I came across it when testing that last part of `#1`.

samdark (approved):

brandonkelly:
Added tests @samdark @rob006

rob006 (approved):

dynasource (approved):

samdark:
Merged. Thank you!

samdark:
Merged. Thank you!

PR #15228, opened 2017-11-27
Test case. Validator test. Changes in test case naming. Changes in class who called create.

bscheshirwork:
`testValidateAttributes` instead `testValidate` because `validate` method not to be tested here.  
Called:
`public function validateAttributes($model, $attributes = null)`,  
`public function validateAttribute($model, $attribute)`,  
`protected function validateValue($value)`

`Validator` instead `TestValidator` 
because functional of  `TestValidator` is not used in this case. 

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


samdark:
Merged. Thanks!

Issue #15229, opened 2017-11-27
yii\console\widgets\Table w/o screen


webleaf:
### What steps will reproduce the problem?

```
echo Table::widget([
    'headers'       => ['h1', 'h2', 'h3'],
    'rows'          => [[1,2,3], [1,2,3], [1,2,3]],
]);
```
run script with `yii\console\widgets\Table` via `cron` with `stdout` to `*.log` file like:

```
php /path/to/index.php command > /path/to.log 2>&1
```

also same problem i see, when just run command like `php /path/to/index.php command` in PhpStorm `Command Line Tools Console` (Tools -> Run Command...)

### What is the expected result?
```
╔════╤════╤════╗
║ h1 │ h2 │ h3 ║
╟────┼────┼────╢
║ 1  │ 2  │ 3  ║
╟────┼────┼────╢
║ 1  │ 2  │ 3  ║
╟────┼────┼────╢
║ 1  │ 2  │ 3  ║
╚════╧════╧════╝
```
### What do you get instead?

```
PHP Warning 'yii\base\ErrorException' with message 'str_repeat(): Second argument has to be greater than or equal to 0'

in /Users/user/Documents/webproject/vendor/yiisoft/yii2/console/widgets/Table.php:286

Stack trace:
#0 [internal function]: yii\base\ErrorHandler->handleError(2, 'str_repeat(): S...', '/Users/user/Doc...', 286, Array)
#1 /Users/user/Documents/webproject/vendor/yiisoft/yii2/console/widgets/Table.php(286): str_repeat('\xE2\x95\x90', -1)
#2 /Users/user/Documents/webproject/vendor/yiisoft/yii2/console/widgets/Table.php(179): yii\console\widgets\Table->renderSeparator('\xE2\x95\x94', '\xE2\x95\xA4', '\xE2\x95\x90', '\xE2\x95\x97')
#3 /Users/user/Documents/webproject/vendor/yiisoft/yii2/base/Widget.php(140): yii\console\widgets\Table->run()
#4 /Users/user/Documents/webproject/commands/Controller.php(26): yii\base\Widget::widget(Array)
#5 /Users/user/Documents/webproject/vendor/yiisoft/yii2/base/BaseObject.php(108): app\commands\Controller->init()
#6 /Users/user/Documents/webproject/vendor/yiisoft/yii2/base/Controller.php(86): yii\base\BaseObject->__construct(Array)
[...]

PR #15233, opened 2017-11-27
_screenWidth default value

webleaf:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #15229

### DIFF
--- a/framework/console/widgets/Table.php
+++ b/framework/console/widgets/Table.php
@@ -366,6 +366,7 @@ class Table extends Widget
 
     /**
      * Getting screen width.
+     * If it is not able to determine screen width, default value `123` will be set.
      *
      * @return int screen width
      */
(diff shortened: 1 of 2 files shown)


framework/console/widgets/Table.php
```diff
@@ -376,6 +377,9 @@ protected function getScreenWidth()
             if (isset($size[0])) {
                 $this->_screenWidth = $size[0];
             }
+            else {
+                $this->_screenWidth = 123;
```
samdark: Why exactly 123?

webleaf: https://github.com/yiisoft/yii2/blob/0948109a19a59c63d4f065343fe530421764d768/framework/console/widgets/Table.php#L301
At line 301 it decreases by 3 (i don't know why. For terminal scrollbar?) and became 120. I think 120 is "standard" line length. May be, it could be larger for log files.

samdark: OK.

samdark (approved):
Please add changelog.

samdark:
Merged. Thanks!

PR #15235, opened 2017-11-27
Fixes #15234

### LINKED ISSUE #15234: LinkPager - tag is lost for second disabled element

```
    <?php echo LinkPager::widget([
        'pagination'                    => $pagination,
        'disabledListItemSubTagOptions' => [
            'tag' => 'a',
        ],
        'firstPageLabel' => true,
    ]); ?>
```

Both `firstPageLabel` and `prevPageLabel` should be enclosed in `a` tag.

### What do you get instead?

Only `firstPageLabel` is enclosed in `a` tag,  `prevPageLabel` is enclosed in `span` tag.

bug

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13
| PHP version      |  any
| Operating system | any

This comes from https://github.com/yiisoft/yii2/blob/master/framework/widgets/LinkPager.php#L246

```
$tag = ArrayHelper::remove($this->disabledListItemSubTagOptions, 'tag', 'span');
```

Method modifies `disabledListItemSubTagOptions` property, not taking into account that there could be other disabled elements in set.

SDKiller:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  |no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #15234


samdark (approved):
Please add changelog.

PR #15239, opened 2017-11-28
[cs] removed unused docblocks

TomasVotruba:
As followup to #11397, this is prepare for PHP 7.1 typehints.

I'd like to take this in separated PRs.


framework/base/Model.php
```diff
@@ -620,10 +620,6 @@ public function addError($attribute, $error = '')
 
     /**
      * Adds a list of errors.
```
sergeymakinen: It seems unnecessary to remove this.

samdark:
Doesn't look good. We're describing not only hints but meaning of properties.

PR #15244, opened 2017-11-29
change order of autoload if phpunit run from outside

bscheshirwork:
Preamble: 

I try to create test environment for yii2 on local machine.

I want to use difference/fresh/git@dev version of PHPUnit
I look at configs `tests/data/config-docker.php`, `tests/docker-compose*.yml`
I prepare common composition
```yml
# Local run unit tests
# for docker-compose CLI like this manual https://blog.jetbrains.com/phpstorm/2016/11/docker-remote-interpreters/
# run directly:
# docker-compose -f ../yii2-phpunit-docker-composition.yml run --rm --entrypoint bash php
# and run `composer install` before first time use
version: '2'
services:
  php:
    image: bscheshir/codeception:php7.1.12-fpm-yii2 #contain phpunit
    volumes:
      - ./yii2/tests/data/config-docker.php:/var/www/html/tests/data/config.php # use named connection for db
      - ./yii2:/var/www/html #src and tests shared to container
      - ~/.composer/cache:/root/.composer/cache
    environment:
      TZ: Europe/Moscow
      XDEBUG_CONFIG: "remote_host=192.168.0.83 remote_port=9002 remote_enable=On"
      PHP_IDE_CONFIG: "serverName=codeception"
    depends_on:
      - pgsql
      - mysql
# example for pgsql
# also add ./tests/data/config.local.php
# <?php
# $config['databases']['pgsql']['dsn'] = 'pgsql:host=pgsql;port=5432;dbname=yii2advanced';
# $config['databases']['pgsql']['username'] = 'yii2advanced';
# $config['databases']['pgsql']['password'] = 'yii2advanced';
  pgsql:
    image: postgres:10.1
    environment:
      TZ: Europe/Moscow
      POSTGRES_PASSWORD: postgres
      POSTGRES_DB: yiitest
      POSTGRES_USER: postgres
    networks:
      default:
        aliases:
         - postgres
  mysql:
    image: mysql:8.0.3
    entrypoint: ['/entrypoint.sh', '--character-set-server=utf8', '--collation-server=utf8_general_ci']
    environment:
      TZ: Europe/Moscow
      MYSQL_ROOT_PASSWORD: secret
      MYSQL_DATABASE: yiitest
      MYSQL_USER: travis
      MYSQL_PASSWORD: travis
```

I install dependencies
```sh
docker-compose -f ../yii2-phpunit-docker-composition.yml run --rm --entrypoint composer php install
```

I try to run tests from my installation of PHPUnit
```sh
docker-compose -f ../yii2-phpunit-docker-composition.yml run --rm --entrypoint bash php
root@13acef9fb08c:/var/www/html# /repo/vendor/bin/phpunit 
PHPUnit 6.2.4 by Sebastian Bergmann and contributors.
...
```

But I have [error](https://github.com/sebastianbergmann/phpunit/issues/2878) with another composer autoload. Some classes of PHPUnit catch from yii2 composer autoloader. 

Proposition:

Check a PHPUnit composer path and inject your autoloader as second autoloader if necessary

In this way we can run both of installation

Outside
```sh
/repo/vendor/bin/phpunit 
```

or build-in
```sh
vendor/bin/phpunit 
```

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no


samdark:
@schmunk42 would you please take a look?

schmunk42:
@bscheshirwork Could you try installing Yii's dependencies with `--no-dev` to see if this would solve your problem.

I had similar issues when I had a global installation of Yii on my base images back then.
I'd avoid having duplicated classes, even if you can order autoloading - you may run into issues which are really hard to debug, since you might get errors from code you do not expect to run.

Another solution might be to have multiple `composer.json` files for your test setup, so you could switch versions easily. You can specify the composer.json to use with the ENV variable [`COMPOSER`](https://getcomposer.org/doc/03-cli.md#composer).

bscheshirwork:
Ok, `--no-dev` is a solution, thanks. I try it now.

Issue #15200, opened 2017-11-22
Controllers in sub folders is not visable in list commands


johndoejdg:
### What steps will reproduce the problem?
- cd to project folder 
- Create test folder: mkdir ./commands/test
- Create test controller in that folder ./commands/test/TestController.php:
```
<?php

namespace app\commands\test;

use yii\console\Controller;

class TestController extends Controller
{
    public function actionDo()
    {

    }
}
```
- Run yii command to see all commands: ./yii

### What is the expected result?
See test command in list
Working autocomplete
 
### What do you get instead?
Test command do not exist in list of command but callable
./yii test/test/do
And autocomplete does not suggest commands in subfolders

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      |  2.0.13.1
| PHP version      |  7.1.11
| Operating system |  Linux 4.13.12-1

dmirogin:
I think worth trying controllerMap

PR #15249, opened 2017-11-29
Fixed: Controllers in sub folders is not visible in list commands

idsulik:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #15200

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.14 under development
 ------------------------
 
-- Bug #15249: Fixed: Controllers in sub folders is not visible in list commands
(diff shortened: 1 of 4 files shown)


samdark (approved):

dmirogin:
Just a question, controllerMap won't resolve this issue ?

samdark:
Yes, it will resolve the issue as well, I guess. Do you suggest we need to build dynamically when scanning for controllers?

dmirogin:
I think it would be better, but if we already have worked solution in this pull request than i ok with that.

samdark:
Merged. Thank you!

Sarke:
Why was this merged instead of #13902 ?

samdark:
Because there were no unit tests in #13902.

Issue #15254, opened 2017-11-29
Tests fail on Postgres >9.5


schmunk42:
### What steps will reproduce the problem?

Run tests against different PgSQL versions

### What is the expected result?

postgres:9.3 (OK)
- https://gitlab.com/yiisoft/yii2/-/jobs/42420411

postgres:9.4 (OK)
- https://gitlab.com/yiisoft/yii2/-/jobs/42426874

### What do you get instead?

postgres:9.5 (fail)
- https://gitlab.com/yiisoft/yii2/-/jobs/42430723

postgres:9 [9.6] (fail)
- https://gitlab.com/yiisoft/yii2/-/jobs/42417285

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | master
| PHP version      |  7.1
| Operating system | Debian (Docker)

I don't use PgSQL in projects, I've just noticed that tests failed on 10.1 also and narrowed it down to different versions; you can click on "Reveal variables" to see the actual version which has been tested. Note, TravisCI uses 9.3 atm.

terabytesoftw:
btw, the tests are currently passing successfully.

https://github.com/yiisoft/yii2/actions/runs/25488265170/job/74789386720

PR #15251, opened 2017-11-29
added default database versions for dockerized testing

schmunk42:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

Added default values for db-images for dockerized tests on GitLab.com

PR #15255, opened 2017-11-29
lowered default version for PgSQL to match TravisCI

schmunk42:
in addition to https://github.com/yiisoft/yii2/pull/15251 - would be better to have a default for which tests pass, see also https://github.com/yiisoft/yii2/issues/15254

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


samdark:
Merged. Thanks!

Issue #9137, opened 2015-07-16
Allow vs Access-Control-Allow-Methods [\yii\rest\OptionsAction]


mohamed-nazim:
I am working on angular-ionic mobile app and wired up my first REST Api with Yii2.

While trying to use a `PUT` request, chrome sends pre-flight `OPTIONS` to check if permitted methods include `PUT`. By default Yii2 reponds with an `Allow` header instead of the expected (by Chrome and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Access-Control-Allow-Methods)) `Access-Control-Allow-Methods`. The response contains the `Allow` header but Chrome reports Method `PUT` is not allowed by `Access-Control-Allow-Methods`.

Upon digging further, I found that `yii\rest\OptionsAction` is responsible in preparing the allowed header. I fiddled with the CORS settings for more than an hour in trying to get the expected header but without success.

I ended up creating my own class **extending yii\rest\OptionsAction** simply adding the required header:

``` php
namespace app\components;

use Yii;

class OptionsAction extends \yii\rest\OptionsAction {
  public function run($id = null) {
    parent::run($id);
    Yii::$app->response->headers->set
    ('Access-Control-Allow-Methods',Yii::$app->getResponse()->getHeaders()
      ->get('Allow'));
  }

}
```

then I **override the actions method** in my controller:

``` php
public function actions() {
  $actions = parent::actions();
  $actions['options'] = [
    'class' => 'app\components\OptionsAction',
  ];
  return $actions;
}
```

I propose that the `\yii\rest\OptionsAction` in the core be updated as to add the MDN proposed/browser adopted header `Access-Control-Allow-Methods` instead of (or on top of) the current `Allow` header.

BioSin:
@samdark looks like an old feature request, maybe I can implement it and make pull request? I have the same problem.

samdark:
OK.

cebe:
> maybe I can implement it and make pull request?

sure go ahead if you know what needs to be done to fix it. If not, feel free to ask and we'll help :)

michbr:
Shouldn't that be "Access-Control-Allow-Methods"? (Plural) 

[https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods)

I only ask because I'm interested in getting this fix.

PR #15261, opened 2017-11-30
Added `Access-Control-Allow-Method` header for the OPTIONS request 

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #9137

### DIFF
--- a/framework/rest/OptionsAction.php
+++ b/framework/rest/OptionsAction.php
@@ -40,5 +40,6 @@ class OptionsAction extends \yii\base\Action
         }
         $options = $id === null ? $this->collectionOptions : $this->resourceOptions;
         Yii::$app->getResponse()->getHeaders()->set('Allow', implode(', ', $options));
+        Yii::$app->getResponse()->getHeaders()->set('Access-Control-Allow-Method', implode(', ', $options));
     }
 }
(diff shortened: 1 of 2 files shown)


samdark (approved):

samdark:
Merged. Thank you!

framework/rest/OptionsAction.php
```diff
@@ -40,5 +40,6 @@ public function run($id = null)
         }
         $options = $id === null ? $this->collectionOptions : $this->resourceOptions;
         Yii::$app->getResponse()->getHeaders()->set('Allow', implode(', ', $options));
+        Yii::$app->getResponse()->getHeaders()->set('Access-Control-Allow-Method', implode(', ', $options));
```
antonkot: Should be plural: Methods instead of Method

SilverFire: @antonkot good catch. Would you like to submit a PR with fix?

antonkot: sure

antonkot: @SilverFire Here it is: #15869

Issue #15265, opened 2017-12-01
PostgreSQL > 10.0 is not pass tests with default value of timestamp CURRENT_TIMESTAMP


bscheshirwork:
### What steps will reproduce the problem?
run PHPUnit test with [this](https://github.com/yiisoft/yii2/pull/15244#issue-277698421) config

```yml
  pgsql:
    image: postgres:10.1
```
and `postgres:10.0`

run tests 
```sh
vendor/bin/phpunit --filter testColumnSchema tests/framework/db/pgsql/SchemaTest.php
```

### What is the expected result?
no F in output

### What do you get instead?
```sh
$ docker-compose -f ../yii2-phpunit-docker-composition.yml run --rm --entrypoint bash php
root@6dc97aa7b2cd:/var/www/html# vendor/bin/phpunit --filter testColumnSchema tests/framework/db/pgsql/SchemaTest.php
PHPUnit 4.8.34 by Sebastian Bergmann and contributors.

F.

Time: 545 ms, Memory: 6.00MB

There was 1 failure:

1) yiiunit\framework\db\pgsql\SchemaTest::testColumnSchema
defaultValue of column ts_default is expected to be an object but it is not.
Failed asserting that 'CURRENT_TIMESTAMP' is of type "object".

/var/www/html/tests/framework/db/SchemaTest.php:485
/var/www/html/vendor/phpunit/phpunit/phpunit:52

FAILURES!
Tests: 2, Assertions: 142, Failures: 1.
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14-dev
| PHP version      | 7.1.12
| Operating system | Debian (Docker)

schmunk42:
@bscheshirwork Could you verify this again, I can't see this bug anymore with `10.1` - see also https://gitlab.com/yiisoft/yii2/-/jobs/49522413

bscheshirwork:
@schmunk42 @samdark 
 on master I can see previous errors.
```sh
vendor/bin/phpunit --filter testColumnSchema tests/framework/db/pgsql/SchemaTest.php
PHPUnit 4.8.34 by Sebastian Bergmann and contributors.

F.

Time: 661 ms, Memory: 6.00MB

There was 1 failure:

1) yiiunit\framework\db\pgsql\SchemaTest::testColumnSchema
defaultValue of column smallint_col does not match.
Failed asserting that '1' is identical to 1.

/var/www/html/tests/framework/db/SchemaTest.php:488
[...]

Issue #15247, opened 2017-11-29
Tests. 7. defaultValue of column smallint_col does not match.


bscheshirwork:
### What steps will reproduce the problem?
run PHPUnit test with [this](https://github.com/yiisoft/yii2/pull/15244#issue-277698421) config

```yml
  pgsql:
    image: postgres:10.1
```
### What is the expected result?
no F in output

### What do you get instead?
7 fails.

about 7th:
```sh
7) yiiunit\framework\db\pgsql\SchemaTest::testColumnSchema
defaultValue of column smallint_col does not match.
Failed asserting that '1' is identical to 1.

/var/www/html/tests/framework/db/SchemaTest.php:488
/var/www/html/vendor/phpunit/phpunit/phpunit:52
```
https://github.com/yiisoft/yii2/blob/6f896ce9173fb7de3a898b7c03439917828ae342/tests/framework/db/SchemaTest.php#L277-L288

https://github.com/yiisoft/yii2/blob/6f896ce9173fb7de3a898b7c03439917828ae342/tests/framework/db/SchemaTest.php#L488
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13@dev
| PHP version      | 2.1.12
| Operating system |Debian

schmunk42:
Relates to https://github.com/yiisoft/yii2/issues/15254

bscheshirwork:
```yml

```
  pgsql:
    image: postgres:9.5
    environment:
      TZ: Europe/Moscow
      POSTGRES_PASSWORD: postgres
      POSTGRES_DB: yiitest
      POSTGRES_USER: postgres
    networks:
      default:
        aliases:
         - postgres

```sh
root@b2734e1154ec:/var/www/html# vendor/bin/phpunit --filter testColumnSchema tests/framework/db/pgsql/SchemaTest.php
PHPUnit 4.8.34 by Sebastian Bergmann and contributors.

F.

Time: 712 ms, Memory: 6.00MB

There was 1 failure:

1) yiiunit\framework\db\pgsql\SchemaTest::testColumnSchema
defaultValue of column smallint_col does not match.
Failed asserting that '1' is identical to 1.

/var/www/html/tests/framework/db/SchemaTest.php:488
/var/www/html/vendor/phpunit/phpunit/phpunit:52
```

bscheshirwork:
9.4 is OK
```yml
  pgsql:
    image: postgres:9.4
    environment:
[...]

PR #15266, opened 2017-12-01
Tests: avoid pgsql bugs of 9.5 and 10.0

bscheshirwork:
This is only compare with actual data from database schema.

All of other usercase, who related to this pgsql errors is broken.

note: Wait for fix it in source and set top versions who have errors in next PR's.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15265 #15247


bscheshirwork:
Also I send info about this errors to pgsql-bugs

samdark:
Would you please post links to posgresql issue tracker bugs here?

samdark:
Should not be merged. It's hiding the issue, not fixing it.

polobo:
Here's a link to the thread on the PostgreSQL online mailing list browser - there is no dedicated bug tracker.

https://www.postgresql.org/message-id/flat/20180124222007.GQ17109%40momjian.us#20180124222007.GQ17109@momjian.us

No one choose to research this report - and whether its a bug or simple compatibility breakage would depend on findings - partly because the reported complaint has not been reduced to something that is demonstrably PostgreSQL-output related.

What would help move this forward in PostgreSQL would be for someone to write up a self-contained script generates data question and then run that script against 9.4, 9.5, 9.6, and 10.  Post the results of each along with the script to the pgsql-bugs@lists.postgresql.org (or replying to the linked thread if you happen to have it in your inbox).  At least 9.4/9.5 where the smallint problem reported seems to have appeared.  At that point the differences in output can either be explained or fixed depending on what caused it in the first place.

David J.

polobo:
FYI - I've done a limited version example of what I think you are talking about (smallint, 9.3 vs 9.6) and posted it to the original bug report thread.

bscheshirwork:
@polobo Thanks!

schmunk42:
From the answer of Tom Lane in the linked [thread](https://www.postgresql.org/message-id/flat/20180124222007.GQ17109%40momjian.us#20180124222007.GQ17109@momjian.us) above:

> Oh, right, this was an intentional change awhile back.
> [...]
> (digs...) It was changed in this 9.5-era commit:

So, I don't think this will be "fixed" in upstream in Postgres.

What can we do about it? Should Yii convert that?

---

Did something change in the core about that, btw?! I can not reproduce the failing `smallint` test locally at the moment.

samdark:
@bscheshirwork would be great if you'll re-check if the issue in still in master?

schmunk42:
> At least 9.4/9.5 where the smallint problem reported seems to have appeared.

@polobo You meant **dis**appeared, right? 😄

schmunk42:
### Postgres test status

All tested against https://github.com/yiisoft/yii2/commit/eb57d4191cbc5a2d8cf5a28345bc81a13e488fa1

#### Broken (only "smallint" bug)
- 9.5.0 - https://gitlab.com/yiisoft/yii2/-/jobs/49501874
- 9.5.1 - https://gitlab.com/yiisoft/yii2/-/jobs/49501843
- 9.5 (9.5.10) - https://gitlab.com/yiisoft/yii2/-/jobs/49503441
- 9.6 (9.6.6) - https://gitlab.com/yiisoft/yii2/-/jobs/49501860
- 10 - https://gitlab.com/yiisoft/yii2/-/jobs/49522348
- 10.1 - https://gitlab.com/yiisoft/yii2/-/jobs/49522413

#### OK
- 9.4 - https://gitlab.com/yiisoft/yii2/-/jobs/49501855

The error with "float_col2" seems to have disappeared.

I am really lost on this issue :(

polobo:
Short answer: The current behavior, since 9.5, is the correct behavior. Those single quotes that now appear are going to stay.  9.4 and earlier (lacking the isngle quotes) was wrong - but won't be fixed in that (or earlier) realeases.  If your code must support both outputs you will need to decide on the best way to do that for your project.

schmunk42:
Isn't the question (or the test) about how we "see" these values on the PHP side?

I.e. Mysql has a [type map](https://github.com/yiisoft/yii2/blob/65f6b59/framework/db/mysql/Schema.php#L41-L70), which explicitly defines `'smallint' => self::TYPE_SMALLINT,` - do we also need that for Postgres?

I do not use Postgres, but from my understanding I'd expect a `smallint` column to be an `integer` in PHP.

polobo:
It boils down to why you are choosing to parse the default expression in the first place instead of just passing through whatever the database returns.  I suppose trimming off the cast would make sense but take the value being cast as-is.  IOW, the failing test should probably be dropped altogether.  But alas I know PostgreSQL but not yii2 so there may be someone feature that depends on the specifics.

polobo:
To make it clear, your are not really parsing a returned value of a smallint, you are parsing the expression that someone happened to attach to a create table command.  In many ways all that is is a string expression whose evaluated result is cast to the column type.  And the client has little means to evaluate the provided expression, only the server can do that.  For the client the expression should just be documentation.  And you should not need to parse documentation.

Issue #14648, opened 2017-08-15
Simple text / plain text model errorSummary


bscheshirwork:
Is there a simple way to log a string for model->errors like  
`yii\helpers\Html::errorSummary($models)`  
without html tags?

i.e.:
```php
$errorsMessages = [];
foreach ($model->getErrors() ?? [] as $attribute => $errors) {
    $errorsMessages[] =  $model->getAttributeLabel($attribute) . ': ' . implode(', ', $errors);
}
$errorsMessage = implode('; ', $errorsMessages);
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | dev
| PHP version      | 7.1.8
| Operating system | alpine

githubjeka:
```php
Yii::error(json_encode($model->errors));
```
:(

Need something more beautiful...

bscheshirwork:
any else?

githubjeka:
https://github.com/yiisoft/yii2/issues/7988
https://github.com/yiisoft/yii2/issues/3087
https://github.com/yiisoft/yii2/pull/12195

bscheshirwork:
see f3776def0f022355a25a03244aa022a3a7cb9efd

I think we need customization of line separator or something like template in table widgets

Issue #7988, opened 2015-04-05
Console model errorSumary()


manquer:
`\yii\helpers\Html::errorSummary()` does a good job of parsing error data and rendering it for Html rendering,

however there is no similar functionality for console applications, i.e. formatted error Output from model errors in  console application , It would useful addition to the Console helper tools

cebe:
can you explain in more detail how you use model validation on console?

manquer:
Typically I write bulk import /file processing operations in my console actions. 

It is useful to do this using console application as for large uploads / bulk imports  it is better to do the processing( when it make take  few minutes or more etc) asynchronously rather than while the request is made . 

1. In this scenario Initial request on the web is made , the object is queued for processing and then console action does the actually processing ( via scheduled cron / long running worker process listening to a queue etc ). 

2. Validations are performed when save() ic called as is usual. 

3. All the errors are piped to file and/or sent as an email to the uploading user. Here un-formatted or text formatted output is better than only HTML formatted output.

Another slightly different use case is throwing validation errors in REST actions, HTML validated output just comes in the way of parsing the errors. Array output is required ( Although models->getErrors() is basically an array an custom function has to be written anyway to parse a group of models )

Currently I use couple of hack helpers functions extending HTML::errorSummary to handle this, However it would be nice to have other errorSummary formats like for un-formatted text, formatted for stderr, and  json output etc

PR #15268, opened 2017-12-01
\yii\helpers\Html::errorSummary() output formats: json and plain text

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7988, #14648


framework/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Yii Framework 2 Change Log
 - Bug #15194: Fixed `yii\db\QueryBuilder::insert()` to preserve passed params when building a `INSERT INTO ... SELECT` query for MSSQL, PostgreSQL and SQLite (sergeymakinen)
 - Bug #15229: Fixed `yii\console\widgets\Table` default value for `getScreenWidth()`, when `Console::getScreenSize()` can't determine screen size (webleaf)
 - Bug #15234: Fixed `\yii\widgets\LinkPager` removed `tag` from `disabledListItemSubTagOptions` (SDKiller)
+- Enh #7988: Added `\yii\helpers\Html::errorSummary()` output formats: json and plain text (developeruz)
```
samdark: It's a bit weird that plaintext and JSON formats are in HTML helper...

samdark: How about putting them into console and json helpers?

framework/helpers/BaseHtml.php
```diff
@@ -1199,24 +1199,61 @@ public static function activeHint($model, $attribute, $options = [])
...
     public static function errorSummary($models, $options = [])
     {
-        $header = isset($options['header']) ? $options['header'] : '<p>' . Yii::t('yii', 'Please fix the following errors:') . '</p>';
         $footer = ArrayHelper::remove($options, 'footer', '');
         $encode = ArrayHelper::remove($options, 'encode', true);
         $showAllErrors = ArrayHelper::remove($options, 'showAllErrors', false);
+        $format = ArrayHelper::remove($options, 'format', 'html');
+        if (!isset($options['header'])) {
+            $defaultHeaderMessage = Yii::t('yii', 'Please fix the following errors:');
+            $header = $format == 'text' ? $defaultHeaderMessage . "\n" : '<p>' . $defaultHeaderMessage . '</p>';
```
samdark: `==` → `===`

framework/helpers/BaseHtml.php
```diff
@@ -1199,24 +1199,61 @@ public static function activeHint($model, $attribute, $options = [])
...
+        } else {
+            $header = $options['header'];
+        }
         unset($options['header']);
+        $lines = self::collectErrors($models, $encode, $showAllErrors);
+
+        if ($format == 'json') {
```
samdark: `==` → `===`

framework/helpers/BaseHtml.php
```diff
@@ -1199,24 +1199,61 @@ public static function activeHint($model, $attribute, $options = [])
...
+            return json_encode($lines);
+        }
 
+        if ($format == 'text') {
```
samdark: `==` → `===`

framework/CHANGELOG.md
developeruz: Yes, you are right 👍 I've added `\yii\helpers\Console::errorSummary()` and `\yii\helpers\Json::errorSummary()`. I moved duplicated part to the `collectErrors` method, but it is quite complicated and I have no idea how to make it simple. Overall it solved 3 codeclimate issues in `errorSummary` and moved complexity to `collectErrors` method

framework/helpers/BaseConsole.php
```diff
@@ -1051,4 +1052,54 @@ public static function endProgress($remove = false, $keepPrefix = true)
...
+    {
+        $showAllErrors = ArrayHelper::remove($options, 'showAllErrors', false);
+        $lines = self::collectErrors($models, false, $showAllErrors);
+
+        return implode("\n", $lines);
+    }
+
+    /**
+     * Return array of the validation errors
+     * @param $models
```
samdark: phpdoc should be completed.

framework/helpers/BaseHtml.php
```diff
@@ -1216,26 +1216,7 @@ public static function errorSummary($models, $options = [])
...
-                        $lines[] = $line;
-                    }
-                    if (!$showAllErrors) {
-                        break;
-                    }
-                }
-            }
-        }
-
+        $lines = Console::collectErrors($models, $encode, $showAllErrors);
```
samdark: Looks wrong. Html helper should not use Console.

framework/helpers/BaseJson.php
```diff
@@ -179,4 +180,23 @@ protected static function processData($data, &$expressions, $expPrefix)
...
+     * - showAllErrors: boolean, if set to true every error message for each attribute will be shown otherwise
+     *   only the first error message for each attribute will be shown. Defaults to `false`.
+     *
+     * @return string the generated error summary
+     * @since 2.0.14
+     */
+    public static function errorSummary($models, $options = [])
+    {
+        $showAllErrors = ArrayHelper::remove($options, 'showAllErrors', false);
+        $lines = Console::collectErrors($models, false, $showAllErrors);
```
samdark: JSON helper should not use console.

samdark (changes requested):
Html and JSON should not use console.

framework/helpers/BaseHtml.php
developeruz: Could you please, help me to figure out where can I move it? If I put the same `collectErrors` into each of helpers, codeclimate detects 3 times copypaste. It seems we don't have a class that all `Html`, `Console` and `Json` are shared together. Should I create one?

framework/helpers/BaseConsole.php
```diff
@@ -1051,4 +1052,54 @@ public static function endProgress($remove = false, $keepPrefix = true)
...
+     * @param $encode
+     * @param $showAllErrors
+     * @return array
+     */
```
Renkas: Shouldnät this method also have `@since` tag?

framework/helpers/BaseConsole.php
developeruz: Done

framework/helpers/BaseHtml.php
samdark: @yiisoft/core-developers, @yiisoft/reviewers any ideas?

MysteryDragon: `collectErrors()` really more related to `Html` than to `Console`
Maybe it's really better to create some... trait, for example?
And basically not use `$encode`, because it relates only to `Html` and use `Html` functional.
And extend it in `Html`, adding `$encode` as third optional parameter with `true` by default.

Maybe it will looks strangely, but will be more right in design terms.
Or say "get off, codeclimate!", and use same code. It will be two times in Console and Json. It's better then to write interface for them, and... it's again story about "right design/simplicity" balance.

samdark: I'd duplicate it.

developeruz: Done, I duplicated `collectErrors()` for each Helper

tests/framework/helpers/JsonTest.php
```diff
@@ -7,6 +7,7 @@
 
 namespace yiiunit\framework\helpers;
 
+use yii\base\DynamicModel;
 use yii\base\Model;
```
MysteryDragon: Now using `Model` is pointless.

framework/helpers/BaseConsole.php
```diff
@@ -1051,4 +1052,54 @@ public static function endProgress($remove = false, $keepPrefix = true)
...
+     *
+     * @return string the generated error summary
+     * @since 2.0.14
+     */
+    public static function errorSummary($models, $options = [])
+    {
+        $showAllErrors = ArrayHelper::remove($options, 'showAllErrors', false);
+        $lines = self::collectErrors($models, $showAllErrors);
+
+        return implode("\n", $lines);
```
MysteryDragon: Isn't it better to use `PHP_EOL` instead of `"\n"`?

samdark (approved):

framework/helpers/BaseConsole.php
developeruz: Yes, you are right 👍

framework/helpers/BaseJson.php
```diff
@@ -179,4 +180,54 @@ protected static function processData($data, &$expressions, $expPrefix)
...
+
+    /**
+     * Return array of the validation errors
+     * @param Model|Model[] $models the model(s) whose validation errors are to be displayed.
+     * @param $showAllErrors boolean, if set to true every error message for each attribute will be shown otherwise
+     * only the first error message for each attribute will be shown.
+     * @return array of the validation errors
+     * @since 2.0.14
+     */
+    private static function collectErrors($models, $showAllErrors)
```
dmirogin: this realisation is equal to BaseConsole::collecErrors. Time for generalization is came 😸 .

dmirogin (changes requested):

framework/helpers/BaseJson.php
MysteryDragon: @dmirogin yeah, it's debatable thing: https://github.com/yiisoft/yii2/pull/15268#pullrequestreview-80677027

dmirogin: i don't like this way. Better to create new class for that.

samdark: @dmirogin what class name/placement do you suggest?

dmirogin: something like this
~~~php
<?php

class ErrorsCollector
{
    /**
     * @var array Model[]
     */
    private $models;
    /**
     * @var bool
     */
    private $showAllErrors;

    public function __construct($models, $showAllErrors = false)
    {
        $this->models = $this->ensureModelsIsArray($models);
        $this->showAllErrors = $showAllErrors;
    }

    /**
     * @return array
     */
    public function getLines()
    {
        $lines = [];
        
        foreach ($this->getModelsErrors() as $error) {
            if (!in_array($error, $lines, true)) {
                $lines[] = $error;
            }
            if (!$this->showAllErrors) {
                break;
            }
        }
        
        return $lines;
    }

    /**
     * @return array
     */
    public function getModelsErrors()
    {
        $modelErrors = [];
        foreach ($this->models as $model) {
            /* @var $model Model */
            foreach ($model->getErrors() as $errors) {
                $modelErrors[] = $errors;
            }
        }
        
        return $modelErrors;
    }

    /**
     * @param Model[]|Model $models
     * @return array
     */
    private function ensureModelsIsArray($models)
    {
        if (!is_array($models)) {
            return [$models];
        }

        return $models;
    }
}

~~~

samdark: Makes sense.

developeruz: @dmirogin, I disagree that we need another separate class that will work with Model errors. It should be Model responsibility.

dmirogin: Ok, now we have 3 identical methods in 3 different classes. Maybe worth consider trait for that ?

developeruz: As you see, codeclimate ok with this 2 identical (they are similar but not the same (c) samdark) methods in 2 different classes, because these methods now are just a couple of lines. And overall in the code base, there are a hundred of places with such "identical" methods. 
@MysteryDragon, @saidbakr  need your opinion as well.

dmirogin (approved):

framework/helpers/BaseJson.php
samdark: In this particular case these methods aren't same as I said before. I don't think it worth extracting two foreach-es in a separate class.

samdark: Well, it is OK either way...

samdark:
Merged. Thank you!

PR #15271, opened 2017-12-01
Created IpHelper

SilverFire:
Replaces #14043


framework/helpers/BaseIpHelper.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace yii\helpers;
+
+/**
+ * Class BaseIpHelper
+ *
+ * @author Dmytro Naumenko <d.naumenko.a@gmail.com>
```
dmirogin: since tag needed

framework/helpers/BaseIpHelper.php
```diff
@@ -0,0 +1,97 @@
...
+        $parts = explode('/', $range);
+        $net = array_shift($parts);
+        $range_cidr = array_shift($parts);
+
+        $netVersion = static::getIpVersion($net);
+        if ($ipVersion !== $netVersion) {
+            return false;
+        }
+        if ($range_cidr === null) {
+            $range_cidr = $netVersion === 4 ? self::IPV4_ADDRESS_LENGTH : self::IPV6_ADDRESS_LENGTH;
```
dmirogin:
```php
$netVersion === 4
```
4 -> self::IPV4

dmirogin (approved):

framework/validators/IpValidator.php
```diff
@@ -354,14 +355,14 @@ private function validateSubnet($ip)
             return [$this->message, []];
         }
 
-        if ($this->getIpVersion($ip) == 6) {
+        if ($this->getIpVersion($ip) === IpHelper::IPV6) {
             if ($cidr !== null) {
-                if ($cidr > static::IPV6_ADDRESS_LENGTH || $cidr < 0) {
```
samdark: Should constants be deprecated?

samdark (approved):

framework/validators/IpValidator.php
SilverFire: done

sergeymakinen: I think a deprecation is more than enough. BC break in this case is overkill.

SilverFire: It's a small break and I don't think a lot of people relied on internal constants of IpValidator

sergeymakinen: I do agree about internals. But the same applies to many framework areas (for example, `yii\db`) and it's a dangerous road I would not step onto.
IMHO, if we make an important feature everyone asks for years or fixing a long story bug, it's okay. But when we add a new helper, there's no need to rush. :)

framework/helpers/BaseIpHelper.php
```diff
@@ -0,0 +1,115 @@
...
+     * ```
+     *
+     * @see https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing
+     *
+     * @param string $ip the valid IPv4 or IPv6 address
+     * @param int|string $cidr the CIDR of $ip
+     * @param string $range subnet in CIDR format e.g. `10.0.0.0/8` or `2001:af::/64`
+     * @return bool whether IP address with $cidr is in range
+     */
+    public static function inRange($ip, $cidr, $range)
```
Renkas: Why have separate cidr parameter for IP but expect CIDR to be inline on range?
I would expect the same logic for both.

Or am I missing somethng obvious here?

SilverFire: Yeah, good point :) It was ok as internal API of IpValidator, but should be re-designed for public API

SilverFire:
Ready to merge

samdark:
Merged. Thank you!

PR #15276, opened 2017-12-02
Add upsert support

sergeymakinen:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | not yet
| Fixed issues  | #13879

Another little PR from me. The PR implements an upsert support for all DBMSes. `ConstraintFinderInterface` was "backported" from a coming migration table column/constraint builder.

TODO:

- [x] Implement the feature
- [x] Add QB tests
- [x] Add basic `Command` tests
- [x] Finish `Command` tests for PostgreSQL >= 9.5 & CUBRID
- [ ] Add concurrency `Command` tests
- [x] Add multi command `Command::execute()` support for SQLite (#15357)
- [x] Add transaction support to `Command` to reliably test SQLite, Oracle and pre-9.5 PostgreSQL (#15415)
- [x] Add retry support to `Command` to reliably test Oracle and pre-9.5 PostgreSQL (#15415)

These "add support" entries require their own PRs I think. Mostly they are not too hard/too long to be implemented. But I need a feedback to start working on them.


framework/db/QueryBuilder.php
```diff
@@ -305,6 +315,115 @@ public function batchInsert($table, $columns, $rows)
...
+     * @throws NotSupportedException if this is not supported by the underlying DBMS.
+     * @since 2.0.14
+     */
+    public function upsert($table, $insertColumns, $updateColumns, &$params)
+    {
+        throw new NotSupportedException($this->db->getDriverName() . ' does not support upsert statements.');
+    }
+
+    /**
+     * @param string $table
```
samdark: Need description.

samdark:
Travis seems to be unhappy about tests.

samdark:
Overall approach looks good. Will review in more detail later.

sergeymakinen:
@samdark Travis will be happy as soon as we implement multi-command `execute` for SQLite's `Command`. With the SQLite's `SqlTokenizer` from 2.0.13 it's trivial.

samdark (approved):
Looks good to me overall. No changes from me.

dmirogin (approved):
So cool! Great work!

samdark:
@sergeymakinen what's left to be done before we merge it?

framework/db/QueryBuilder.php
```diff
@@ -305,6 +315,115 @@ public function batchInsert($table, $columns, $rows)
...
+     * ```
+     *
+     * The method will properly escape the table and column names.
+     *
+     * @param string $table the table that new rows will be inserted into/updated in.
+     * @param array|Query $insertColumns the column data (name => value) to be inserted into the table or instance
+     * of [[Query]] to perform `INSERT INTO ... SELECT` SQL statement.
+     * @param array|bool $updateColumns the column data (name => value) to be updated if they already exist.
+     * If `true` is passed, the column data will be updated to match the insert column data.
+     * If `false` is passed, no update will be performed if the column data already exists.
```
rob006: Use case: 

I have a table with 20 columns. On upsert, if record exists, I want to update all column except `created_at`. How to do this with this signature? I need to repeat all columns (except one) with own expressions?

framework/db/QueryBuilder.php
```diff
@@ -147,38 +147,51 @@ public function build($query, $params = [])
 
     /**
      * Creates an INSERT SQL statement.
-     *
```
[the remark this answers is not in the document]
sergeymakinen: Got it, thanks!

framework/db/QueryBuilder.php
sergeymakinen: Then you have to pass an array of `Expression`s with `VALUES(column)`/`EXCLUDED.column` depending on a DBMS.
I forgot to note this syntax, good catch. :)

sergeymakinen:
@samdark Well, everything which is on TODO:
- multi-statement `execute` for SQLite (easy, I'm going to make a PR on tue/wed)
- transaction/retry support for `Command` (easy by itself)

Also, in order to use transaction/retry in `Command`, it must be DBMS-aware, to choose its optimal configuration. Because we instantiate commands via `createCommand()` it should not be a problem.

PostgreSQL <= 9.5, SQLite & Oracle upsert commands have to be executed in a transaction. Because there's not enough fun, Oracle's ones have to executed in `try/catch` and repeated in case of special exceptions.

If everyone is okay with it, I will finish these tasks on the next week.

framework/db/QueryBuilder.php
rob006: Honestly, a don't really like this syntax. It took me like 20 minutes to get how this works and results were really surprised to me. I would expect that `$updateColumns` would work similar to `select()`:

1. `['first_name', 'last_name']` will update only `first_name` and `last_name` columns:
   ```ON DUPLICATE KEY UPDATE `first_name`=VALUES(`first_name`), `last_name`=VALUES(`last_name`)```.
2. `['first_name', 'old_first_name' => new Expression('first_name')]` will update `first_name` column and put previous value of `first_name` into `old_first_name`:
   ```ON DUPLICATE KEY UPDATE `first_name`=VALUES(`first_name`), `old_first_name`=first_name```.
3. `['new_first_name' => 'first_name']` will update `new_first_name`column with new `first_name` value:
   ```ON DUPLICATE KEY UPDATE `new_first_name`=VALUES(`first_name`)```.
4. `['first_name' => new Expression('"rob006"')]` will always set `rob006` as `first_name` if record exist:
    ```ON DUPLICATE KEY UPDATE `first_name`="rob006"```.

This should simplify every non-standard (when you don't need just update all columns with inserted values) case and can be used in the same way in `upsertMultiple()`.

sergeymakinen: TBH, I really like this idea. As you could see from the current implementation, it will require a "bit" of magic but it's definitely possible.
I'm going to try to implement it as soon as everything described in https://github.com/yiisoft/yii2/pull/15276#issuecomment-350549841 land.

sergeymakinen:
So… I hope to finish remaining tasks on new year holidays.

samdark:
@sergeymakinen what's missing here except a merge with master?

SilverFire:
Do not have permissions to push. Created a PR in @sergeymakinen repo: https://github.com/sergeymakinen/yii2/pull/2

Will wait for tests

samdark:
You can push/PR against Yii repo as well, I think.

SilverFire:
Continue in #15623.

PR #15282, opened 2017-12-04
Extracted ViewInterface from yii\web\View yiisoft/yii2#14997

njasm:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14997

### DIFF
--- a/framework/web/ViewInterface.php
+++ b/framework/web/ViewInterface.php
@@ -1,0 +1,229 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\web;
+use yii\base\InvalidConfigException;
+
+/**
+ * ViewInterface defines a view object in the MVC pattern.
+ *
+ * View provides a set of methods (e.g. [[render()]]) for rendering purpose.
+ *
+ * View is configured as an application component in [[\yii\base\Application]] by default.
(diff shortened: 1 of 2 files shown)


framework/web/ViewInterface.php
```diff
@@ -0,0 +1,229 @@
...
+     * @return string the rendering result
+     * @see render()
+     */
+    public function renderAjax($view, $params = [], $context = null);
+
+    /**
+     * Registers the asset manager being used by this view object.
+     * @return \yii\web\AssetManager the asset manager. Defaults to the "assetManager" application component.
+     */
+    public function getAssetManager();
```
samdark: Doesn't look like it belongs to the same interface...

framework/web/ViewInterface.php
```diff
@@ -0,0 +1,229 @@
...
+
+    /**
+     * Sets the asset manager.
+     * @param \yii\web\AssetManager $value the asset manager
+     */
+    public function setAssetManager($value);
```
samdark: Doesn't look like it belongs to the same interface...

njasm:
this pull is not finished yet.

We do have a base view that I still didn't understand why, the only usage (or reference to it) is the `yi\web\View`that extend it.

I believe the main idea was to have a base View where the Web and Console controller could have they're own view implementations [plus the base methods that are used by the framework that probably are not in the context of Web Controller or Console], don't know if this is true. need to investigate further.

samdark:
Yes, your assumption is true.

PR #15288, opened 2017-12-04
Extracted FormatterInterface from yii\i18n\Formatter yiisoft/yii2#14997

njasm:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #14997

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -45,7 +45,7 @@ use yii\helpers\HtmlPurifier;
  * @author Carsten Brandt <mail@cebe.cc>
  * @since 2.0
  */
-class Formatter extends Component
+class Formatter extends Component implements FormatterInterface
 {
     /**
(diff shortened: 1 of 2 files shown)


samdark:
Not sure it makes a lot of practical sense to extract this interface but overall it looks OK...

samdark:
@cebe need you opinion.

njasm:
@samdark not extracting all yii's core/components interfaces, even if they don't make any pratical sense at first will invalidate the possibility in #14997 of autowiring capabilities.. I believe.

cebe:
I don't see why you need an interface to enable autowiring. If there is no other practical sense in using the interface than to enable DI, we should fix DI to work with classes directly. An interface that requires to implement > 20 methods does not make sense.

PR #15294, opened 2017-12-05
Middleware

antgubarev:
It is a draft of middleware


docs/guide/middleware.md
```diff
@@ -0,0 +1,90 @@
...
+        ],        
+        function (\yii\web\Request $request, \yii\web\Response $response) {
+            $response->addHeader("heaader", "custom");
+        }
+```
+
+```php
+namespace app\middleware;
+
+use yii\web\MiddlewareInterface;
```
samdark: https://github.com/php-fig/fig-standards/tree/master/proposed/http-handlers I think it makes sense to make it compatible with PSR interfaces (it's still not accepted but at least it's something commonly used).

docs/guide/middleware.md
```diff
@@ -0,0 +1,90 @@
...
+    'rules' => [
+        [
+            'pattern' => '/',
+            'route' => 'site/index',
+            'middleware' => [
+                [
+                    'middleware' => function ($request, \yii\web\Response $response) {
+                        echo "router middleware!";
+                    },
+                    'priority' => 5,
```
samdark: Is it necessary?

docs/guide/middleware.md
```diff
@@ -0,0 +1,90 @@
...
+'urlManager' => [
+    'enablePrettyUrl' => true,
+    'showScriptName' => false,
+    'rules' => [
+        [
+            'pattern' => '/',
+            'route' => 'site/index',
+            'middleware' => [
+                [
+                    'middleware' => function ($request, \yii\web\Response $response) {
```
samdark: Config syntax looks a bit weird with `middleware.middleware`

amarox: I agree. But i has not decided yet how to call it

docs/guide/middleware.md
amarox: Other frameworks has this feature but it is cosily

docs/guide/middleware.md
amarox: fixed

docs/guide/middleware.md
amarox: Need ServerRequestInterface. Now yii\web\Request implemented by RequestInterface.

amarox:
please move to 2.1 milestone

framework/web/Application.php
```diff
@@ -55,7 +55,72 @@ class Application extends \yii\base\Application
...
+     * @throws MiddlewareException
+     */
+    public function addMiddleware($middleware)
+    {
+        if (is_callable($middleware)) {
+            $middleware = [
+                'handle' => $middleware
+            ];
+        } elseif (is_string($middleware)) {
+            if (!class_exists($middleware)) {
```
SilverFire: I'd remove this check as it will immediately trigger autoloading and may affect performance.

framework/web/Application.php
```diff
@@ -55,7 +55,72 @@ class Application extends \yii\base\Application
...
+                throw new MiddlewareException("Class {$middleware} not found");
+            }
+            $middleware = [
+                'handle' => $middleware
+            ];
+        } elseif (is_array($middleware) && !isset($middleware['handle'])) {
+            throw new MiddlewareException("Param `middleware` must be set");
```
SilverFire: middleware -> handle

framework/web/Application.php
```diff
@@ -118,6 +188,64 @@ public function handleRequest($request)
...
+        usort($this->middlewareStack, function ($a, $b) {
+            if ($a['priority'] == $b['priority']) {
+                return 0;
+            }
+            return $a['priority'] > $b['priority'] ? -1 : 1;
+        });
+
+        foreach ($this->middlewareStack as $middleware) {
+
+            if (in_array($action, (array)$middleware['except'])) {
```
SilverFire: Could you add `true` as a 3-rd argument?

framework/web/Application.php
```diff
@@ -118,6 +188,64 @@ public function handleRequest($request)
...
+                continue;
+            }
+            if (count($middleware['only'])>0 && !in_array($action, $middleware['only'])) {
```
SilverFire: Will be a notice in case `only` is not set.
isset + is_array before counting is required

SilverFire: Could you add `true` as a 3-rd argument of `in_array`?

framework/web/Application.php
```diff
@@ -118,6 +188,64 @@ public function handleRequest($request)
...
+                continue;
+            }
+
+            if (is_callable($middleware['handle'])) {
+                Yii::debug("Execute middleware closure");
+                $middleware['handle']($this->getRequest(), $this->getResponse());
+            } elseif (is_string($middleware['handle'])) {
+                $object = new $middleware['handle'];
```
SilverFire: Probably, DI should be used here

framework/web/Application.php
```diff
@@ -55,7 +55,72 @@ class Application extends \yii\base\Application
...
     public $controller;
+    /**
+     * Stack of middleware
+     * @var array
+     */
+    protected $middlewareStack = [];
 
+    /**
+     * Add middleware to stack
+     * @param $middleware callable|string|array
```
SilverFire: + `\Closure`

framework/web/Application.php
```diff
@@ -55,7 +55,72 @@ class Application extends \yii\base\Application
      * @var Controller the currently active controller instance
      */
     public $controller;
+    /**
+     * Stack of middleware
+     * @var array
+     */
+    protected $middlewareStack = [];
```
SilverFire: I'd probably prefer having a separate class to store middleware stack

SilverFire (changes requested):
Good beginning, thank you for the first step.

framework/web/Application.php
amarox: "Will be a notice in case only is not set.". 
No. All keys always exists. See https://github.com/yiisoft/yii2/pull/15294/files#diff-d21ce18380a9785b6cee4643f50873f1R97

framework/web/Application.php
amarox: I thought about it. I wanted to make a collection, and MiddlewareItem class. But decided that this is memory. Should i make it?

SilverFire: For my opinion yes, but it's a point to discuss. PHP 7.1 stores classes in memory as efficient as arrays, so it should not be an issue.

framework/web/Application.php
```diff
@@ -118,6 +188,64 @@ public function handleRequest($request)
...
+     * @param $action string|null
+     * @throws MiddlewareException
+     */
+    protected function callMiddlewareStack($action)
+    {
+        usort($this->middlewareStack, function ($a, $b) {
+            if ($a['priority'] == $b['priority']) {
+                return 0;
+            }
+            return $a['priority'] > $b['priority'] ? -1 : 1;
```
MysteryDragon: I know it's only draft, but perhaps spaceship operator will be nice here due to php7.1+ requirements?

klimov-paul:
I do not like such functionality to be mandatory part of the core as it introduces extra composer dependency, which is not mandatory to be used. Keeping in mind issue #8452 I would like to avoid it.

It would be better to extract it into an Application Behavior using `Application::EVENT_BEFORE_REQUEST` and `Application::EVENT_AFTER_REQUEST` events.
In this case functionality can be extracted into a separated package, which can be used at will.

samdark:
`http-interop/http-middleware` contains PSR interface only, not implementation.

klimov-paul:
> http-interop/http-middleware contains PSR interface only, not implementation.

So what? Extra package is still an extra package.

As far as I remember official PSR packages have 'psr' namespaces, e.g. `psr/http-message` not the 'http-interop' like `http-interop/http-middleware`.

amarox:
@klimov-paul behaviors is disputable feature... May be tomorrow behaviors will be removed. But middleware for many frameworks (and micro) is part of the core.

samdark:
Yes. PSR for middlewares isn't yet released. Here it is (called handlers now): https://github.com/php-fig/fig-standards/tree/master/proposed/http-handlers

framework/composer.json
```diff
@@ -75,7 +75,8 @@
         "cebe/markdown": "~1.0.0 | ~1.1.0",
         "bower-asset/jquery": "3.2.*@stable | 3.1.*@stable | 2.2.*@stable | 2.1.*@stable | 1.11.*@stable | 1.12.*@stable",
         "bower-asset/punycode": "1.3.*",
-        "bower-asset/yii2-pjax": "~2.0.1"
+        "bower-asset/yii2-pjax": "~2.0.1",
+        "http-interop/http-middleware": "0.5"
```
klimov-paul: `http-interop/http-middleware` is abandoned at packagist.org. According to the upgrade note `http-interop/http-server-middleware` should be used instead.

klimov-paul (changes requested):

docs/guide/middleware.md
klimov-paul: Then it should be applied.

klimov-paul (changes requested):

klimov-paul:
Relates to #15438

klimov-paul:
Alternative solution proposed at #15459

samdark:
Solved by https://github.com/yiisoft/yii2/pull/15416

PR #15308, opened 2017-12-07
Bug #5341: HasMany via two relations

shirase:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #5341


framework/db/ActiveRelationTrait.php
```diff
@@ -278,7 +279,14 @@ public function populateRelation($name, &$primaryModels)
             $buckets = $this->indexBuckets($buckets, $this->indexBy);
         }
 
-        $link = array_values(isset($viaQuery) ? $viaQuery->link : $this->link);
+        $link = array_values($this->link);
+        if (isset($viaQuery)) {
+            $deepViaQuery = $viaQuery;
+            while ($deepViaQuery->via) {
+                $deepViaQuery = is_array($deepViaQuery->via) ? $deepViaQuery->via[1] : $deepViaQuery->via;
```
dmirogin:
~~~php
is_array($deepViaQuery->via)
~~~
doesn't garantee that there is element with index 1

framework/db/ActiveRelationTrait.php
```diff
@@ -380,21 +388,41 @@ private function populateInverseRelation(&$primaryModels, $models, $primaryName,
...
                 $key1 = $this->getModelKey($viaModel, $viaLinkKeys);
                 $key2 = $this->getModelKey($viaModel, $linkValues);
                 $map[$key2][$key1] = true;
             }
+
+            $viaQuery->viaMap = $map;
+
+            $viaVia = $viaQuery->via;
+            while ($viaVia) {
+                $viaViaQuery = is_array($viaVia) ? $viaVia[1] : $viaVia;
```
dmirogin: same

framework/db/ActiveRelationTrait.php
```diff
@@ -380,21 +388,41 @@ private function populateInverseRelation(&$primaryModels, $models, $primaryName,
...
+                $viaMap = $viaViaQuery->viaMap;
+                $oldMap = $map;
+                $map = [];
+                foreach ($oldMap as $key1 => $keys) {
```
dmirogin: Don't like this way, key1, key2, key3, a lot of foreaches, can we do it better ?

dmirogin (dismissed):

framework/db/ActiveRelationTrait.php
shirase: Unnecessary checks. Format of `ActiveRelationTrait::via` property is fixed `[$relationName, $relation]` and must have index 1 if its type is Array. I can throw exception, but need add that check everywhere.

framework/db/ActiveRelationTrait.php
shirase: I'll try to change

framework/db/ActiveRelationTrait.php
```diff
@@ -423,6 +442,20 @@ private function buildBuckets($models, $link, $viaModels = null, $viaLink = null
         return $buckets;
     }
 
+    /**
+     * @param array $map
```
samdark: Need description

samdark (approved):
Looks good to me.

dmirogin (approved):

samdark:
Merged. Thank you!

Faryshta:
Need another PR to update guides.

samdark:
@Faryshta what should be updated? It is a bug that was fixed, not a new feature.

Issue #15307, opened 2017-12-07
Not checked filterModel before registerJs in GridView


ice2038:
I am use GridView on AMP pages https://www.ampproject.org/ru/docs/reference/spec (not scripts, list allowed tags, inline style, magic UI).

Example:
```php
$dataProvider = new ArrayDataProvider([
        'allModels'  => $allModels,
        'pagination' => false,
]);

echo \yii\grid\GridView::widget([
                'dataProvider' => $dataProvider,
                'layout' => '{items}',
                'formatter' => ['class' => 'yii\i18n\Formatter','nullDisplay' => ''],
                'columns' => $columns,
]);
```

I don't used filters, but 
```html
<script src="https://site.ru/minify/{many nums}.js?v=1507556035"></script>
<script type="text/javascript">jQuery(function ($) {
jQuery('#w0').yiiGridView({"filterUrl":"\/controller\/action\/arg?","filterSelector":"#w0-filters input, #w0-filters select"});
});</script>
```

Kill him
\yii\grid\GridView::run()
```php
/**
 * Runs the widget.
 */
public function run()
{
    if ($this->filterModel !== null) {
        $id = $this->options['id'];

        $options = Json::htmlEncode($this->getClientOptions());

        $view = $this->getView();
        GridViewAsset::register($view);
        $view->registerJs("jQuery('#$id').yiiGridView($options);");
    }

    parent::run();
}
```

Do I need to make a pulrequest or not?

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14-dev
| PHP version      | 7.0.21-1
| Operating system | Ubuntu 16.04

samdark:
Sounds OK. Please do.

PR #15309, opened 2017-12-07
Bug #15307: Add check filterModel before registerJs in `yii\grid\GridView`

ice2038:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15307


cebe:
this is not correct, gridview.js also handles the checkbox column. Not only filters.

Issue #9727, opened 2015-09-19
Add clearErrors parameter to ValidateMultiple


hector-del-rio:
Currently only validate() method has it but it would make sense that you can also set it in validateMultiple(). See https://github.com/yiisoft/yii2/pull/9728

PR #15314, opened 2017-12-08
Add clearErrors parameter to ValidateMultiple

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | probably yes, because method signature is changed
| Tests pass?   | yes
| Fixed issues  | #9727

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -889,15 +889,16 @@ class Model extends Component implements StaticInstanceInterface, IteratorAggreg
      * @param array $attributeNames list of attribute names that should be validated.
      * If this parameter is empty, it means any attribute listed in the applicable
      * validation rules should be validated.
+     * @param bool $clearErrors whether to call [[clearErrors()]] before performing validation. Default true
      * @return bool whether all models are valid. False will be returned if one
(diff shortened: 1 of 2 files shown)


samdark:
Should be pull requested against 2.1 since it's method signature change.

developeruz:
Changed to be against 2.1

samdark (approved):

framework/base/Model.php
```diff
@@ -887,17 +887,18 @@ public static function loadMultiple($models, $data, $formName = null)
      * be of the same or different types.
      * @param array $models the models to be validated
      * @param array $attributeNames list of attribute names that should be validated.
+     * @param bool $clearErrors whether to call [[clearErrors()]] before performing validation. Default true
```
MysteryDragon: This line must be moved after full `$attributeNames` description.

MysteryDragon (changes requested):

MysteryDragon (approved):

samdark:
Merged. Thank you!

Issue #15302, opened 2017-12-06
DbCache to a PostgreSQL database


edwards-sj:
### What steps will reproduce the problem?

Enable `DbCache` to a `PostgreSQL` database
Turn on `enableSchemaCache` in the database connection
 
### What is the expected result?

Yii2 will use the `cache` table to cache the results of its schema description queries. 

### What do you get instead?

`yii\base\ErrorException`
`unserialize() expects parameter 1 to be string, resource given`

### Additional info

The cache database uses `bytea` to store the values, when loading these from a query you get back a stream rather than a value.

The `getValue` method of `DbCache` uses `queryScalar` which automatically checks for a stream and calls `stream_get_contents`.

The `getValues` method does a `queryAll` which does not do the check, which means the return values are streams rather than the underlying data.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13.1
| PHP version      | 7.1.7
| Operating system | Ubuntu

schmunk42:
@edwards-sj With which Postgres version did you test this?

bscheshirwork:
@edwards-sj try with 9.5...

edwards-sj:
I'm using PostgreSQL 10.0

PR #15316, opened 2017-12-08
DbCache,  getValues now behaves the same as getValue with regards to streams

edwards-sj:
Changed yii\caching\DbCache so that getValues now behaves the same as getValue with regards to streams

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15302


dmirogin (approved):

samdark (approved):

edwards-sj:
The code was already being tested by the `yiiunit\framework\caching\DbCacheTest` although admittedly only for mysql.

I've added `yiiunit\framework\caching\PgSQLCacheTest` which runs it for PostgreSQL as well, this will fail without the fix in place.

samdark:
Merged. Thank you!

PR #15318, opened 2017-12-08
Fix session_name(): Cannot change session name when session is active

bscheshirwork:
https://github.com/php/php-src/commit/7f196e3

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | [yii2-app-advanced #326](https://github.com/yiisoft/yii2-app-advanced/issues/326)


bscheshirwork:
I got BC broken in child 

`yiisoft\yii2-redis\Session`

bscheshirwork:
I can understand what the reason configure session name throw DI...

bscheshirwork:
I will add check on "standard storage". 
Disabled for other. 

What about add this check for all engine-dependencies functions of `web\Sessions`?

bscheshirwork:
related #15523

samdark:
Should be fixed like https://github.com/yiisoft/yii2/blob/master/framework/web/Session.php#L421

bscheshirwork:
Question: the unfrozen name is correct?

framework/web/Session.php
```diff
@@ -322,7 +322,12 @@ public function getName()
      */
     public function setName($value)
     {
+        if ($this->getUseCustomStorage()) {
```
samdark: I don't think there's a need for it.

tests/framework/web/session/SessionTest.php
```diff
@@ -70,4 +70,22 @@ public function testParamsAfterSessionStart()
...
+     * Test set name. Also check set name twice and after open
+     */
+    public function testSetName()
+    {
+        $session = new Session();
+        $session->setName('oldName');
+
+        $this->assertEquals('oldName', $session->getName());
+
+        $session->open();
```
samdark: Why do we need to call `open` here?

bscheshirwork: > check set name ... after open

samdark: Right.

framework/web/Session.php
samdark: So why it's here?

samdark:
Merged. Thanks!

bscheshirwork:
> Removed unneeded code [skip ci]

```diff
-        if ($this->getUseCustomStorage()) {
-            return;
-        }
```
@samdark 
without this "unneeded code" the nested components like [redis session](https://github.com/yiisoft/yii2-redis/blob/master/src/Session.php) can use standard 

https://github.com/yiisoft/yii2/blob/86664eb50336f71e8dfcdf3871a33197cbbaa659/framework/web/Session.php#L945

this is (may be) BC broken

bscheshirwork:
for example: we change storage to the `redis` in the config file. We can't change name from config file. We got a error.

samdark:
Don't we need the same check then for setId, setSavePath, regenerateID, setName, setUseCookies etc.?

bscheshirwork:
IMHO yes. For any non-redefine functional in children. Now. In future.
If in user code (in surce or in config) this functional will called = got a error.

But may be this is responsibility of a end user?

samdark:
Yes. That could be done but in a separate pull request.

PR #15319, opened 2017-12-08
Update BaseHtml.php

dizews:
added ability to show multiple errors on field

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no


dizews:
I have a file it has a some validation rules. I need to show all errors of this file

samdark:
You can use error summary for it. The code looks like it solves a single case and there are no tests. No go for 2.0.

PR #15327, opened 2017-12-10
Bulgarian relative date translation fix.

ddinchev:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Translation typo fix - there is a missing space for relative time for years (it's correct for any other time intervals like hours, months, etc). For example it would be `преди 4години` (incorrect) instead of `преди 4 години` (correct).


samdark:
Merged. Thanks!

PR #15331, opened 2017-12-11
fix to validate when `$requiredValue` is `null` and validation is strict

albertborsos:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | When `RequiredValidator::requiredValue` is `null` and `RequiredValidator::strict` is `true` and the value in `$value` is `null` then the validation failed.

### DIFF
--- a/framework/validators/RequiredValidator.php
+++ b/framework/validators/RequiredValidator.php
@@ -68,7 +68,7 @@ class RequiredValidator extends Validator
     protected function validateValue($value)
     {
         if ($this->requiredValue === null) {
-            if ($this->strict && $value !== null || !$this->strict && !$this->isEmpty(is_string($value) ? trim($value) : $value)) {
+            if ($this->strict && $value === null || !$this->strict && !$this->isEmpty(is_string($value) ? trim($value) : $value)) {
                 return null;
             }
         } elseif (!$this->strict && $value == $this->requiredValue || $this->strict && $value === $this->requiredValue) {
--- a/framework/validators/RequiredValidator.php
+++ b/framework/validators/RequiredValidator.php
@@ -59,6 +59,10 @@ class RequiredValidator extends Validator
         if ($this->message === null) {
             $this->message = $this->requiredValue === null ? Yii::t('yii', '{attribute} cannot be blank.')
                 : Yii::t('yii', '{attribute} must be "{requiredValue}".');
+
+            if ($this->requiredValue === null && $this->strict) {
+                $this->message = Yii::t('yii', '{attribute} must be blank.');
+            }
         }
     }
 
--- a/tests/framework/validators/RequiredValidatorTest.php
+++ b/tests/framework/validators/RequiredValidatorTest.php
@@ -31,6 +31,11 @@ class RequiredValidatorTest extends TestCase
         $this->assertFalse($val->validate([]));
         $this->assertTrue($val->validate('not empty'));
         $this->assertTrue($val->validate(['with', 'elements']));
+        $val->strict = true;
+        $this->assertTrue($val->validate(null));
+        $this->assertFalse($val->validate([]));
+        $this->assertFalse($val->validate('should fail'));
+        $this->assertFalse($val->validate(['should', 'fail', 'with', 'elements']));
     }
 
     public function testValidateValueWithValue()


MysteryDragon:
> strict: whether to check data types when validating a value. Defaults to false. When requiredValue is not set, if this property is true, the validator will check if the input value is **not** strictly null
Looks like it was expected behavior, or am I wrong?

albertborsos:
I see this in the docblock over the `$strict` property.

> Note that when [[requiredValue]] is null, if this property is true, the validator will check
> if the attribute value is null;

MysteryDragon:
Yeah, definitely misunderstanding between docs and phpdoc, and the question is which case is more logical. My point of view that docs case if more correct, because changing behavior so extreme by side parameter... A little weird for me.

albertborsos:
I have a record with attribute `parent_id` with `null` value. This record is in a release, so the values of the record's attributes cannot be change.

I think this is the correct rule to validate this case on update:

```php
[['parent_id], 'required', 'requiredValue' => $this->parent_id, 'strict' => true, 'when' => function () {
    return $this->isReleased();
}]
```

But if you have a better validator for this scenario, I can accept it :)

albertborsos:
> `strict`: [...] When `requiredValue` is not set, if this property is `true`, the validator will check if the input value is not strictly `null`; [...]. When `requiredValue` is set, the comparison between the input and `requiredValue` will also check data types if this property is true.

@MysteryDragon You are right with this in the docs. But there is a forgotten case when I set the `requiredValue` to `null` :)

MysteryDragon:
@albertborsos how about `compare` validator in this case?

albertborsos:
@MysteryDragon then, I think I have found an another bug.

in the docs of CompareValidator it said: 

> `compareValue`: a constant value that the input value should be compared with. When both of this property and `compareAttribute` are specified, this property will take precedence.

And for this validator:
```php
[['parent_id'], 'compare', 'compareValue' => $this->_model->parent_id, 'when' => function () {
        return $this->_model-> isReleased();
}],
```

I get his error message:

>   [yii\base\UnknownPropertyException] Getting unknown property: app\forms\UpdatePageForm::parent_id_repeat  

But it should not assert with the `parent_id_repeat` attribute, because `compareValue` is set.

MysteryDragon:
@albertborsos and this is strange. Because I took a look of source code of this validator, and it looks fine.
And also try this simple test:
```
<?php

namespace app\models;

class TestModel extends \yii\base\Model
{
    public $targetAttribute;

    public $relatedModel;

    public function init()
    {
        $this->relatedModel = new \yii\base\DynamicModel(['relatedAttribute' => 15, 'isReleased' => true]);
        parent::init();
    }

    public function rules()
    {
        return [
            [['targetAttribute'], 'compare', 'compareValue' => $this->relatedModel->relatedAttribute, 'when' => function () {
                return $this->relatedModel->isReleased;
            }],
        ];
    }
}
```
```
<?php

namespace app\controllers;

class TestController extends \yii\web\Controller
{
    public function actionIndex()
    {
        $model = new \app\models\TestModel(['targetAttribute' => 15]);
        echo $model->validate() ? 1 : 0, "<br>";
        $model->targetAttribute = 14;
        echo $model->validate() ? 1 : 0, "<br>";
    }
}
```
And no problem. Actual Yii 2.0.13 version, but it seems like it doesn't matter so much 'cause this code was untouched since 2014.

albertborsos:
@MysteryDragon set `relatedAttribute` to `null`

MysteryDragon:
@albertborsos yeah, you right, I forgot about your `null` already :D 
So, logical consequence...
But my point of view that if this situation must be covered by some validator, then `compare` validator is matching more for your case. This attributes of `require` validator already have some evil-magic logic, it's not good. And you actually comparing values.
And I think that it mustn't 'cause `null` is meaning "nothing", and many validators repel on that. It looks right for me. And this situation may be covered by custom validator, for example.

albertborsos:
You are right, compare validator is more relevant in this case thn the required validator. But I think the compare validator should cover the `null` case too. Maybe with a `strict` property.

MysteryDragon:
It's really controversial issue as for me.
If maintainers also counts that it's wrong, maybe it's better to mark somehow that user attribute values is different thing. As for me it may cause BC break and needs to be applied for all validators. Just my point of view.

MysteryDragon:
@albertborsos It's really controversial issue as for me.
If maintainers also counts that it's wrong, maybe it's better to mark somehow that user attribute values is different thing. As for me it may cause BC break and needs to be applied for all validators. Just my point of view.

nkovacs:
> Note that when [[requiredValue]] is null, if this property is true, the validator will check
if the attribute value is null;

I think you're misunderstanding the documentation (it is worded poorly). If strict is false, the attribute will be considered empty (i.e. it will *fail* the validation) if `isEmpty()` returns true. If strict is true, the attribute will be considered empty (fail the validation) if it is === null.

samdark:
@nkovacs would you help improving wording?

cebe:
> I have a record with attribute `parent_id` with `null` value. This record is in a release, so the values of the record's attributes cannot be change.

if the value can not be changed, it should not be active in the current scenario...

PR #15337, opened 2017-12-11
Adds dbmanager cache into an examle

OndrejVasicek:
When I used the RBAC first time, I didn't know about the dbmanager caching. I was relaying on the examples only. Then I saw many queries in my logs and started searching to eventually find the cache property.

I think there isn't a situation, when someone doesn't use caching for dbmanager rbac, so i'ts very helpful to put it in the example above.

Hope I'm correct.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

### DIFF
--- a/docs/guide/security-authorization.md
+++ b/docs/guide/security-authorization.md
@@ -225,6 +225,7 @@ return [
     'components' => [
         'authManager' => [
             'class' => 'yii\rbac\DbManager',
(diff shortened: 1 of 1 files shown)


samdark:
Merged. Thanks!

Issue #15339, opened 2017-12-12
Data (rest) filter: \yii2\rest\IndexAction::prepareDataProvider() produced wrong SORT if $requestParams not set


bscheshirwork:
### What steps will reproduce the problem?

```php
class SomeRestController extends \yii\rest\ActiveController
{
...
    public function actions()
    {
        return [
            'index' => [
                'class' => 'yii\rest\IndexAction',
                'modelClass' => $this->modelClass,
                'dataFilter' => [
                    'class' => \yii\data\ActiveDataFilter::class,
                    'attributeMap' => [
                        'redefineId' => 'originId',
                    ],
                    'searchModel' => function () {
                        return (new \yii\base\DynamicModel(['id' => null, 'redefineId' => null,  'name' => null]))
                            ->addRule('id', 'integer')
                            ->addRule('formId', 'redefineId')
                            ->addRule('name', 'trim')
                            ->addRule('name', 'string');
                    },
                ],
            ],
...
```
send query like
```php
                /** @var \yii\httpclient\Response $response */
                $response = $client->createRequest()
                    ->setMethod('get')
                    ->setData(['filter' => ['redefineId' => $this->id]])
                    ->setUrl('some-rest/index')
                    ->send();
```

### What is the expected result?
result set will be order by PK
### What do you get instead?
order by first field in the filter `where` condition 

in
https://github.com/yiisoft/yii2/blob/d44d5540b1f58b92ac1f9c6ac1ffc0036970026a/framework/rest/IndexAction.php#L89-L129

See line 116 and line 126.

And another strange: In 119-128 we pass `$requestParams` with 'filter' key to 'pagination' and 'sort'

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14-dev 6345766
| PHP version      |  7.2.0
| Operating system | alpine3.6

klimov-paul:
Unable to reproduce.
Ordering is controller by `yii\data\Sort` instance, which is a part of `DataProvider`.
Unless `Sort::$sortParam` is not present inside the `$requestParams` no order will be applied.
Check your configuration for the `Sort` instance, e.g. `Sort::$defaultOrder` or something.

Also I can not understand how following part of your report code snippent can function:

```
->addRule('formId', 'redefineId')
```

as the second argument of the `DynamicModel::addRule()` should be a validator name - not attribute.

bscheshirwork:
>Also I can not understand how following part of your report code snippent can function:
[...]

PR #15343, opened 2017-12-12
Add order by pk if data filter applied

bscheshirwork:
> note: `$this->dataFilter` can be change in `call_user_func($this->prepareDataProvider, $this, $filter);`

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes/no change behavior around sort with filter
| Tests pass?   | yes/no
| Fixed issues  | #15339


bscheshirwork:
@klimov-paul what about this? It is fixed correctly?

klimov-paul:
This is obviously incorrect as it enforces data ordering per each GET request, which produces extra computing at DB level.

PR #15348, opened 2017-12-12
Implemented custom data types support in DBMSs

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #7640, #15377

This is the alternative implementation of PR #14819, #14836

TODO:

- [x] Code idea
- [x] Discuss and approve it @yiisoft/reviewers @yiisoft/core-developers 
- [x] Add popular types for PgSQL, MySQL
- [x] Enhance tests coverage
- [x] Write docs
- [x] Merge


SilverFire:
In comparison to previous implementations, this one provides DBMS-agnostic classes of expressions (`ArrayExpression`, `JsonExpression`) that are simply a DTO.

Expression building is moved builders that implement `ExpressionBuilderInterface` and are DBMS-specific. I would like to implement wider JSON support in DB abstraction layer based on this PR.

Dear reviewers, please ignore imperfect PHPDocs. They are, but we need to agree design first. Thank you

framework/db/ExpressionBuilder.php
```diff
@@ -0,0 +1,24 @@
...
+    use ExpressionBuilderTrait;
+
+    /**
+     * {@inheritdoc}
+     * @param Expression|ExpressionInterface $expression the expression to be built
+     */
+    public function build(ExpressionInterface $expression, &$params = [])
+    {
+        $params = array_merge($params, $expression->params);
+        return $expression->__toString();
```
samdark: I'd explicitly add a method to the interface instead of using magical `__toString`.

SilverFire: This expression builder is specific for `yii\db\Expression`. Which kind of method would you add to the interface?

framework/db/pgsql/JsonExpressionBuilder.php
```diff
@@ -0,0 +1,64 @@
...
+namespace yii\db\pgsql;
+
+use yii\db\ExpressionBuilderInterface;
+use yii\db\ExpressionBuilderTrait;
+use yii\db\ExpressionInterface;
+use yii\db\JsonExpression;
+use yii\db\Query;
+
+/**
+ * Class JsonExpressionBuilder builds [[JsonExpression]] for PostgreSQL DBMS.
```
samdark: We'll be able to add another builder for MySQL, correct?

samdark:
Love the approach! I'm all for it.

framework/db/pgsql/JsonExpressionBuilder.php
SilverFire: Right

framework/db/ExpressionBuilder.php
samdark: Nevermind. It's fine.

framework/db/ExpressionBuilderTrait.php
```diff
@@ -0,0 +1,28 @@
...
+namespace yii\db;
+
+/**
+ * Trait ExpressionBuilderTrait provides common constructor for classes that
+ * should implement [[ExpressionBuilderInterface]]
+ *
+ * @author Dmytro Naumenko <d.naumenko.a@gmail.com>
+ * @since 2.0.14
+ */
+trait ExpressionBuilderTrait
```
zuozp8: i see no usecase where you need splitted interface and trait, i would merge ExpressionBuilderTrait and ExpressionBuilderInterface into abstract class

framework/db/QueryBuilder.php
```diff
@@ -145,6 +166,43 @@ public function build($query, $params = [])
...
+     * @see ExpressionBuilderInterface
+     * @see expressionBuilders
+     * @since 2.0.14
+     * @throws InvalidParamException when $expression building is not supported by this QueryBuilder.
+     */
+    public function buildExpression(ExpressionInterface $expression, &$params = [])
+    {
+        $className = get_class($expression);
+        if (!isset($this->expressionBuilders[$className])) {
+            foreach ($this->expressionBuilders as $expressionClass => $builderClass) {
```
zuozp8: i would rename `$builderClass` into `$builderClassOrObject`

framework/db/QueryBuilder.php
```diff
@@ -145,6 +166,43 @@ public function build($query, $params = [])
...
+                if (is_subclass_of($expression, $expressionClass)) {
+                    $this->expressionBuilders[$className] = $builderClass;
```
zuozp8: it's a bit counter intuitive that behavior depends on the order of execution
imagine:
we have `$this->expressionBuilders = ['…\ParentExpression' => '…\ConcreteBuilder']
we have `class ChildExpression extends …\ParentExpression`

if we call `->buildExpression($childExpression)` and then `->buildExpression($parentExpression)` then we create separate builder instances for each class

if we call `->buildExpression($parentExpression)` and then `->buildExpression($childExpression)` then we create one instance that is shared between classes

framework/db/ExpressionBuilderTrait.php
SilverFire: Disagreed. I can build my own extending tree and implement class wherever I want.
Abstract class forces me to add it in extending tree root and makes reusing harder.

framework/db/QueryBuilder.php
zuozp8: if we have
`$this->expressionBuilders = ['…\GrandParentExpression' => '…\Builder1', '…\ParentExpression' => '…\Builder2']`
and we pass instance of `…\ChildExpression` to be built, then we would use Builder1, but it seems Builder 2 would be more suitable

it may be problem as setExpressionBuilders is only appending to the list, it's worth documenting at least

SilverFire: Yes, that my be confusing. Of course we can go deeper and find the most suitable builder with `class_parents` and `class_impements`, but it will affect performance and increase code complexity. Do we need it?

I'm OK with documenting this behavior.

framework/db/ExpressionInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * The database abstraction layer of Yii framework supports objects that implement this
+ * interface and will use [[ExpressionBuilderInterface]] to build them.
+ *
+ * The default implementation is a class [[Expression]].
+ *
+ * @author Dmytro Naumenko <d.naumenko.a@gmail.com>
+ * @since 2.0.14
+ */
+interface ExpressionInterface
```
zuozp8: how about calling it QueryFragmentInterface and make Query implement in the future?

It would make us able to use subqueries everywhere (everywhere where we have expression support) and simplify the code. Otherwise we'll never support subqueries in non-obvious places like inside `order by`

framework/db/ExpressionBuilderTrait.php
zuozp8: you're right, separation will be needed for example in https://github.com/yiisoft/yii2/pull/15348/files#r156621010

framework/db/ExpressionInterface.php
SilverFire: Not sure about naming, but in general it's a good idea originally suggested by @hiqsol [in previous PR](https://github.com/yiisoft/yii2/pull/14836#issuecomment-330166721)

framework/db/QueryBuilder.php
hiqsol: Considering `array_merge` appends builders to the end of list I think builders should be processed in reverse order which allows to solve the problem.

framework/db/ExpressionInterface.php
SilverFire: Will try to adopt it in next changes iteration

klimov-paul:
I do not like the idea - it seems to overcomplicated for the purpose.
If you wish to grad develoepr ability to configure particular condition building it is better to be done via [QueryBuilder::$conditionBuilders](https://github.com/yiisoft/yii2/blob/master/framework/db/QueryBuilder.php#L50-LL68). All we need to be is add an ability to specify particular builder as a PHP callback (or a object matching certain interface if you can not live without it) with following signature:

```php
function (QueryBuilder $queryBuilder, string $operator, array $operands, array &$params) : string {
    return 'SQL';
}
```
Particular condition builder may accept different data types for the `$operands` including `db\Expression` or any other complex object.

Obviously `QueryBuilder` will need an additional public setter for the `$conditionBuilders`, like following:

```php
public function setConditionBuilders(array $conditionBuilders)
{
    $this->conditionBuilders = array_merge($this->conditionBuilders, $conditionBuilders);
}
```

Such approach will be much more flexible, allowing of handling both custom data types and operators.

> Implemented custom data types support in DBMSs

Such statement is incorrect regardless to this PR, as well @SilverFire's late attempts of solving original issue, since it covers only query condition composition, while Array DB type support require handling of the array data at the `INSERT` and `UPDATE` statement as well.
In order to grand support for `ARRAY` or `JSON` DB data types, there should be granted ability to pass  PHP arrays as a column value for the `insert()` and `update()` functions.
If you wish to solve the issue using custom TYPE-based classes - they should be implemented at the level of `yii\db\ColumnSchema` instead of  database expressions.

klimov-paul:
Having both `QueryBuilder::$conditionBuilders` and `QueryBuilder::$expressionBuilders` intoduces 2 different appraoches fot doing the same thing. If it is an object interface for condition composition you want - `QueryBuilder::$conditionBuilders` should be entirely removed in favor of `QueryBuilder::$expressionBuilders` and any complex operators should be composed via custom `db\Expressions` classes.
E.g. instead of following code:

```php
$rows = (new Query())
    ->from('item')
    ->where(['in', 'id', [1, 2, 3]])
    ->andWhere(['between', 'date', '2017-09-01', '2017-10-01'])
    ->all();
```

the following code should be used:

```php
$rows = (new Query())
    ->from('item')
    ->where(new InExpression('id', [1, 2, 4]))
    ->adnWhere(new BetweenExpression('date', '2017-09-01', '2017-10-01'))
    ->all();
```

SilverFire:
Generally you suggestion it makes sense, but in wise-versa logical direction.

Expression – is a Data Transfer Object (DTO) for dataset, that can be somehow compiled to some SQL statement (operators, string, array, JSON, etc)
Condition – is an aggregate of multiple Expressions that can be compiled to SQL condition.

In means that we can create create condition classes (`SimpleCondition`, `InCondition`, `ExistsCondition`) that implement `ExpressionInterface` and create builders for each of them (or for logical groups of them). This kind of refactoring should also reduce complexity of `QueryBuilder` and make each expression more testable.

SilverFire:
Ha, you've posted nearly the same while I composed my message.

> E.g. instead of following code:
> ```php
> $rows = (new Query())
>     ->from('item')
>     ->where(['in', 'id', [1, 2, 3]])
>     ->andWhere(['between', 'date', '2017-09-01', '2017-10-01'])
>     ->all();
> ```
>
> the following code should be used:
> 
> ```php
> $rows = (new Query())
>     ->from('item')
>     ->where(new InExpression('id', [1, 2, 4]))
>     ->adnWhere(new BetweenExpression('date', '2017-09-01', '2017-10-01'))
>     ->all();
> ```

In order to keep BC we can transform the first into the second automatically.

sergeymakinen:
In general, @klimov-paul already expressed my initial suggestions for the PR.

I'll review it in detail a bit later but I love the idea and I really like the way it's implemented. @SilverFire, keep it rolling! 👍

framework/base/Controller.php
```diff
@@ -5,6 +5,8 @@
  * @license http://www.yiiframework.com/license/
  */
 
+// TODO titti
```
SilverFire: Nevermind :)

SilverFire:
Added another bunch of effort in this PR.

- Implemented JSON and Array support for PgSQL (in ActiveRecord layer as well)
- Added `PdoValue` class to deprecate array-based (`[$value, \PDO::PARAM_LOG]`) type casting

Future iterations plan: 

- JSON for MySQL
- Merge `expressionBuilders` and `conditionBuilders` as discussed above with @klimov-paul 

Sorry for long delays between commits. AR & DBAL is one of the most complex parts in Yii so it's not easy to apply changes there and stay concentrated for a long time.

docs/guide/db-query-builder.md
```diff
@@ -758,3 +758,15 @@ $unbufferedDb->close();
 ```
 
 > Note: unbuffered query uses less memory on the PHP-side, but can increase the load on the MySQL server. It is recommended to design your own code with your production practice for extra massive data, [for example, divide the range for integer keys, loop them with Unbuffered Queries](https://github.com/yiisoft/yii2/issues/8420#issuecomment-296109257).
+
+## DBMS specific features <span id="dbms-specific-features"></span>
+
+### PostgreSQL arrays
+
+TBD
```
samdark: Should not be merged w/o it done.

docs/guide/db-query-builder.md
```diff
@@ -758,3 +758,15 @@ $unbufferedDb->close();
...
+
+
+### PostgreSQL `ILIKE`
+
+TBD
```
samdark: Should not be merged w/o it done.

framework/UPGRADE.md
```diff
@@ -56,6 +56,11 @@ Upgrade from Yii 2.0.13
 * Constants `IPV6_ADDRESS_LENGTH`, `IPV4_ADDRESS_LENGTH` were moved from `yii\validators\IpValidator` to `yii\helpers\IpHelper`.
   If your application relies on these constants, make sure to update your code to follow the changes.
 
+* `yii\db\ExpressionInterface` has been introduced to represent a wider range of SQL expressions. In case you check for
+  `instanceof yii\db\Expression` in your code, you might consider changing that to checking for the interface and use the newly
+  introduced methods to retrieve the expression content.
+  
+* TODO: `['value', \PDO::PARAM_STR]` syntax deprecation.
```
samdark: Should not be merged w/o it done.

framework/db/ArrayExpression.php
```diff
@@ -0,0 +1,83 @@
+<?php
+
+namespace yii\db;
+
+/**
+ * Class ArrayExpression represents an array SQL expression.
+ *
+ * Expressions of this type can be used for example in conditions, like:
```
samdark: Delete "for example".

framework/db/ArrayExpression.php
```diff
@@ -0,0 +1,83 @@
...
+     * SQL error will be raised.
+     */
+    protected $type;
+    /**
+     * @var array|QueryInterface|mixed the array content. Either represented as an array of values or a [[Query]] that
+     * returns these values. A single value will be considered as an array containing one element.
+     */
+    protected $value;
+    /**
+     * @var int
```
samdark: phpdoc needed.

framework/db/ArrayExpression.php
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * ArrayExpression constructor.
+     *
+     * @param array|QueryInterface|mixed $value the array content. Either represented as an array of values or a Query that
+     * returns these values. A single value will be considered as an array containing one element.
+     * @param string|null $type the type of the array elements. Defaults to `null` which means the type is
+     * not explicitly specified. In case when type is not specified explicitly and DBMS can not guess it from the context,
+     * SQL error will be raised.
+     * @param int $dimension
```
samdark: phpdoc needed.

framework/db/ColumnSchema.php
```diff
@@ -114,12 +115,18 @@ public function dbTypecast($value)
...
             return null;
         }
-        if ($value === null || gettype($value) === $this->phpType || $value instanceof Expression || $value instanceof Query) {
+        if ($value === null || gettype($value) === $this->phpType || $value instanceof ExpressionInterface || $value instanceof Query) {
             return $value;
         }
+        if (is_array($value) && count($value) === 2 && isset($value[1]) && in_array($value[1], $this->getPdoParamTypes(), true)) {
+            return new PdoValue($value[0], $value[1]);
+            // Backward compatibility thing, but too unreliable. Array [1,2] will pass into this condition and transform
+            // into PdoValue class. However, arrays did not come this method earlier so should not be a big deal. TODO: consider.
```
samdark: > However, arrays did not come this method earlier

Are you sure about that? Was it impossible to supply an array here?

framework/db/ColumnSchema.php
```diff
@@ -143,4 +150,9 @@ protected function typecast($value)
 
         return $value;
     }
+
+    private function getPdoParamTypes()
```
samdark: phpdoc

framework/db/Command.php
```diff
@@ -337,8 +337,8 @@ public function bindValue($name, $value, $dataType = null)
      * @param array $values the values to be bound. This must be given in terms of an associative
      * array with array keys being the parameter names, and array values the corresponding parameter values,
      * e.g. `[':name' => 'John', ':age' => 25]`. By default, the PDO type of each value is determined
-     * by its PHP type. You may explicitly specify the PDO type by using an array: `[value, type]`,
-     * e.g. `[':name' => 'John', ':profile' => [$profile, \PDO::PARAM_LOB]]`.
+     * by its PHP type. You may explicitly specify the PDO type by using a PdoValue class: `new PdoValue(value, type)`,
```
samdark: Full namespace needed for PdoValue and `[[   ]]`.

framework/db/PdoValue.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace yii\db;
+
+/**
+ * Class PdoValue represents a $value that should be bound to PDO with defined $type and
+ * should not be transformed before binding.
+ *
+ * TODO: usage example description, update docs in framework to follow this change.
```
samdark: Should be done before merge.

framework/db/mysql/JsonExpressionBuilder.php
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * {@inheritdoc}
+     * @param JsonExpression|ExpressionInterface $expression the expression to be built
+     */
+    public function build(ExpressionInterface $expression, &$params = [])
+    {
+        $value = $expression->getValue();
+
+        if ($value instanceof Query) {
+            list ($sql, $params) = $this->queryBuilder->build($value, $params);
```
samdark: Is it OK that params aren't used?

framework/db/pgsql/ArrayExpressionBuilder.php
```diff
@@ -0,0 +1,134 @@
...
+use yii\db\ArrayExpression;
+use yii\db\ExpressionBuilderInterface;
+use yii\db\ExpressionBuilderTrait;
+use yii\db\ExpressionInterface;
+use yii\db\JsonExpression;
+use yii\db\Query;
+
+/**
+ * Class ArrayExpressionBuilder builds [[ArrayExpression]] for PostgreSQL DBMS.
+ * TODO: tests
```
samdark: Should be done before merging.

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+namespace yii\db\pgsql;
+
+/**
+ * TODO: phpdoc, tests
```
samdark: Should be done before merging.

framework/db/pgsql/ColumnSchema.php
```diff
@@ -0,0 +1,102 @@
...
+                return (bool) $value;
+            case Schema::TYPE_JSON:
+                return json_decode($value, true);
+        }
+
+        return parent::phpTypecast($value);
+    }
+
+    /**
+     * TODO: phpdoc
```
samdark: Should be done before merging.

framework/db/pgsql/JsonExpressionBuilder.php
```diff
@@ -0,0 +1,55 @@
...
+     * {@inheritdoc}
+     * @param JsonExpression|ExpressionInterface $expression the expression to be built
+     */
+    public function build(ExpressionInterface $expression, &$params = [])
+    {
+        $value = $expression->getValue();
+
+        if ($value instanceof Query) {
+            list ($sql, $params) = $this->queryBuilder->build($value, $params);
+            return "($sql)" . $this->getTypecast($expression);
```
samdark: It is good that params aren't used?

framework/db/ExpressionBuilderInterface.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * Method builds the raw SQL from the $expression that will not be additionally
+     * escaped or quoted.
+     *
+     * @param ExpressionInterface $expression the expression to be built.
+     * @param array $params the binding parameters.
+     * @return string the raw SQL that will not be additionally escaped or quoted.
+     */
+    public function build(ExpressionInterface $expression, &$params = []);
```
samdark: Does it make sense to declare params as `array &$params = []`?

framework/db/PdoValue.php
```diff
@@ -0,0 +1,51 @@
...
+     * @param $value
+     * @param $type
+     */
+    public function __construct($value, $type)
+    {
+        $this->value = $value;
+        $this->type = $type;
+    }
+
+    public function getValue()
```
samdark: phpdoc

framework/db/PdoValue.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        return $this->value;
+    }
+
+    public function getType()
```
samdark: phpdoc

framework/db/QueryBuilder.php
```diff
@@ -1584,12 +1626,8 @@ public function buildSimpleCondition($operator, $operands, &$params)
 
         if ($value === null) {
             return "$column $operator NULL";
-        } elseif ($value instanceof Expression) {
-            foreach ($value->params as $n => $v) {
-                $params[$n] = $v;
-            }
-
-            return "$column $operator {$value->expression}";
+        } elseif ($value instanceof ExpressionInterface) {
```
samdark: Could be just `if`.

framework/db/QueryBuilder.php
```diff
@@ -1584,12 +1626,8 @@ public function buildSimpleCondition($operator, $operands, &$params)
...
+            return "$column $operator {$this->buildExpression($value, $params)}";
         } elseif ($value instanceof Query) {
```
samdark: Could be just `if`.

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * Convert array from PostgreSQL to PHP
+     *
+     * @param string $value string to be converted
+     * @return array|null
+     */
+    public function parse($value)
+    {
+        if (is_null($value)) {
```
samdark: `$value === null`

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        if ($value === '{}') {
+            return [];
+        }
+
+        return $this->parseArray($value);
+    }
+
+    private function parseArray($value, &$i = 0)
```
samdark: phpdoc

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $result = [];
+        for(++$i; $i < strlen($value); ++$i) {
```
samdark: 1. `$value` could be in UTF so `strlen` won't work.
2. Formatting.
3. `strlen` should not be called in condition part of `for`. Either in pre-condition or out of cycle.

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,86 @@
...
+                case '{':
+                    $result[] = $this->parseArray($value, $i);
+                    break;
+                case '}':
+                    break 2;
+                case $this->delimiter:
+                    if (empty($result)) { // `{}` case
+                        $result[] = null;
+                    }
+                    if (in_array($value[$i + 1], [$this->delimiter, '}'])) { // `{,}` case
```
samdark: Specify `true` as 3rd argument.

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,86 @@
...
+                    break;
+                default:
+                    $result[] = $this->parseString($value, $i);
+            }
+        }
+
+        return $result;
+    }
+
+    private function parseString($value, &$i)
```
samdark: phpdoc

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $isQuoted = $value[$i] === '"';
+        $stringEndChars = $isQuoted ? ['"'] : [$this->delimiter, '}'];
+        $result = '';
+        for ($i += $isQuoted ? 1 : 0; $i < strlen($value); ++$i) {
```
samdark: 1. `$value` could be in UTF so `strlen` won't work.
2. `strlen` should not be called in condition part of `for`. Either in pre-condition or out of cycle.

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,86 @@
...
+            if (in_array($value[$i], ['\\', '"'], true) && in_array($value[$i + 1], [$value[$i], '"'], true)) {
+                ++$i;
+            } elseif (in_array($value[$i], $stringEndChars)) {
```
samdark: Specify `true` as 3rd argument.

framework/db/pgsql/ColumnSchema.php
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            return $value;
+        }
+
+        return $this->phpTypecastValue($value);
+    }
+
+    /**
+     * TODO: phpdoc
```
samdark: Should be done before merging.

samdark (changes requested):

framework/db/pgsql/JsonExpressionBuilder.php
SilverFire: This variable is passed by reference, so it is OK

framework/db/mysql/JsonExpressionBuilder.php
SilverFire: They are passed be reference

framework/db/ColumnSchema.php
SilverFire: It was used in `dbTypecast()` and `phpTypecast()` methods only.

Here are the complete list of usages:

**dbTypecast()**

 - `QueryBuilder::insert()` prefixed with `!is_array()`
https://github.com/yiisoft/yii2/blob/2ba98c6d5befb667fa4c476951f01a53fee07e02/framework/db/QueryBuilder.php#L195-L197

- `QueryBuilder::batchInsert()` wrapped in `!is_array()` condition
https://github.com/yiisoft/yii2/blob/2ba98c6d5befb667fa4c476951f01a53fee07e02/framework/db/QueryBuilder.php#L279-L281

- `QueryBuilder::update()` prefixed with `!is_array()`
https://github.com/yiisoft/yii2/blob/2ba98c6d5befb667fa4c476951f01a53fee07e02/framework/db/QueryBuilder.php#L344-L346

- MySQL `QueryBuilder::insert()` prefixed with `!is_array()`
https://github.com/yiisoft/yii2/blob/2ba98c6d5befb667fa4c476951f01a53fee07e02/framework/db/mysql/QueryBuilder.php#L259-L251

- Oracle `QueryBuilder::insert()` prefixed with `!is_array()`
https://github.com/yiisoft/yii2/blob/2ba98c6d5befb667fa4c476951f01a53fee07e02/framework/db/oci/QueryBuilder.php#L212-L215

- Oracle `QueryBuilder::batchInsert()` wrapped in `!is_array()` condition
https://github.com/yiisoft/yii2/blob/2ba98c6d5befb667fa4c476951f01a53fee07e02/framework/db/oci/QueryBuilder.php#L268

- PgSQL `QueryBuilder::batchInsert()` wrapped in `!is_array()` condition
https://github.com/yiisoft/yii2/blob/master/framework/db/pgsql/QueryBuilder.php#L305-L307

- SQLite `QueryBuilder::batchInsert()` wrapped in `!is_array()` condition
https://github.com/yiisoft/yii2/blob/master/framework/db/sqlite/QueryBuilder.php#L101-L103

**phpTypecast()**

- `ActiveRecord::populateRecord()` – potentially dangerous
https://github.com/yiisoft/yii2/blob/5a8c3d537ba14dbff02b2adee9072d381b9eb6b7/framework/db/ActiveRecord.php#L418-L427

- `ActiveRecord::insertInternal()` – have you ever seen primary key represented as array of arrays?
https://github.com/yiisoft/yii2/blob/5a8c3d537ba14dbff02b2adee9072d381b9eb6b7/framework/db/ActiveRecord.php#L515-L517

- Also used to typecast default values: [1](https://github.com/yiisoft/yii2/blob/1501c659aca7c50fa5e4f6fbe5ca549acc946c86/framework/db/cubrid/Schema.php#L335), [2](https://github.com/yiisoft/yii2/blob/4e176ca3259083a8c441c70182bc10e96cbfec12/framework/db/mssql/Schema.php#L386), [3](https://github.com/yiisoft/yii2/blob/f9fd4da8de6c0272cd8956e94da1656b41f1416c/framework/db/mysql/Schema.php#L303), [4](https://github.com/yiisoft/yii2/blob/185209957e11ba6a427c1aa418dfcba7734e8b7b/framework/db/oci/Schema.php#L416), [5](https://github.com/yiisoft/yii2/blob/284dae6a55e44590d495831b9761f1ca50a8bc76/framework/db/pgsql/Schema.php#L565), [6](https://github.com/yiisoft/yii2/blob/2d672b67223b8d930f0f444b113d0a13fbd551c0/framework/db/sqlite/Schema.php#L364), but should not be harmful.

Considering that tests are not broken, I tend to say that the change is not a serious breaking.

samdark: Yes. There could be composite PKs. These contain multiple columns.

SilverFire: No, composite PK is array of strings: `['id', 'version']`, not array of arrays `[['id'], ['version']]`

samdark: Then it should be fine.

framework/db/pgsql/ArrayParser.php
SilverFire: 1. Added test to prove it works OK. Can you suggest an example, when it does not? https://github.com/yiisoft/yii2/pull/15348/commits/8f4c56912f90da3a6b5dbe8faeb00ab5de48faff
2. Done

framework/db/ExpressionBuilderInterface.php
SilverFire: Not sure

samdark: I think it does.

framework/db/conditions/BetweenCondition.php
```diff
@@ -0,0 +1,92 @@
...
+     * {@inheritdoc}
+     * @throws InvalidParamException if wrong number of operands have been given.
+     */
+    public static function fromArrayDefinition($operator, $operands)
+    {
+        if (!isset($operands[0], $operands[1], $operands[2])) {
+            throw new InvalidParamException("Operator '$operator' requires three operands.");
+        }
+
+        return (new static($operands[0], $operator, $operands[1], $operands[2]));
```
samdark: No need for extra `(` and `)` wrapping whole `new` clause.

framework/db/conditions/BetweenConditionBuilder.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace yii\db\conditions;
+
+use yii\db\ExpressionBuilderInterface;
+use yii\db\ExpressionBuilderTrait;
+use yii\db\ExpressionInterface;
+use yii\db\Query;
```
samdark: Not used. Remove.

framework/db/conditions/ConditionInterface.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
samdark: Standard Yii header missing.

framework/db/conditions/BetweenConditionBuilder.php
```diff
@@ -0,0 +1,58 @@
+<?php
```
samdark: Standard Yii header missing.

framework/db/conditions/BetweenCondition.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
```
samdark: Standard Yii header missing.

framework/db/conditions/AndCondition.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
```
samdark: Standard Yii header missing.

framework/db/conditions/ConjunctionCondition.php
```diff
@@ -0,0 +1,47 @@
...
+     * @return string
+     */
+    abstract public function getOperator();
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function fromArrayDefinition($operator, $operands)
+    {
+        return (new static($operands));
```
samdark: No need for `(` and `)`.

framework/db/conditions/ExistsCondition.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public static function fromArrayDefinition($operator, $operands)
+    {
+        if (!isset($operands[0]) || !$operands[0] instanceof Query) {
+            throw new InvalidParamException('Subquery for EXISTS operator must be a Query object.');
+        }
+
+        return (new static($operator, $operands[0]));
```
samdark: No need for `(` and `)`.

framework/db/conditions/HashCondition.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace yii\db\conditions;
+
+use yii\base\InvalidParamException;
```
samdark: Not used. Remove.

framework/db/conditions/InCondition.php
```diff
@@ -0,0 +1,84 @@
...
+     * {@inheritdoc}
+     * @throws InvalidParamException if wrong number of operands have been given.
+     */
+    public static function fromArrayDefinition($operator, $operands)
+    {
+        if (!isset($operands[0], $operands[1])) {
+            throw new InvalidParamException("Operator '$operator' requires two operands.");
+        }
+
+        return (new static($operands[0], $operator, $operands[1]));
```
samdark: `(` and `)` aren't needed.

framework/db/conditions/InConditionBuilder.php
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        if (!is_array($values) && !$values instanceof \Traversable) {
+            // ensure values is an array
+            $values = (array) $values;
+        }
+
+        if ($column instanceof \Traversable || ((is_array($column) || $column instanceof \Countable) && count($column) > 1)) {
+            return $this->buildCompositeInCondition($operator, $column, $values, $params);
+        } elseif (is_array($column)) {
```
samdark: No need for else since previously there was a return.

framework/db/conditions/LikeCondition.php
```diff
@@ -0,0 +1,71 @@
...
+    protected $escapingReplacements;
+
+    /**
+     * @param string $column the column name.
+     * @param string $operator the operator to use (e.g. `LIKE`, `NOT LIKE`, `OR LIKE` or `OR NOT LIKE`)
+     * @param string[]|string $value single value or an array of values that $column should be compared with.
+     *   If it is an empty array the generated expression will  be a `false` value if operator is `LIKE` or `OR LIKE`
+     *   and empty if operator is `NOT LIKE` or `OR NOT LIKE`.
+     */
+    public function __construct($column, $operator, $value)
```
samdark: Is it for phpdoc only?

framework/db/conditions/LikeCondition.php
```diff
@@ -0,0 +1,71 @@
...
+     * {@inheritdoc}
+     * @throws InvalidParamException if wrong number of operands have been given.
+     */
+    public static function fromArrayDefinition($operator, $operands)
+    {
+        if (!isset($operands[0], $operands[1])) {
+            throw new InvalidParamException("Operator '$operator' requires two operands.");
+        }
+
+        $condition = (new static($operands[0], $operator, $operands[1]));
```
samdark: No need for `(` and `)`.

framework/db/conditions/LikeConditionBuilder.php
```diff
@@ -0,0 +1,96 @@
...
+
+        $parts = [];
+        foreach ($values as $value) {
+            if ($value instanceof ExpressionInterface) {
+                $phName = $this->queryBuilder->buildExpression($value, $params);
+            } else {
+                $phName = $this->queryBuilder->bindParam(empty($escape) ? $value : ('%' . strtr($value, $escape) . '%'), $params);
+            }
+            $escapeSql = '';
+            if ($this->escapeCharacter !== null) {
```
samdark: Could be moved out of foreach.

framework/db/conditions/NotCondition.php
```diff
@@ -0,0 +1,50 @@
...
+     * {@inheritdoc}
+     * @throws InvalidParamException if wrong number of operands have been given.
+     */
+    public static function fromArrayDefinition($operator, $operands)
+    {
+        if (count($operands) !== 1) {
+            throw new InvalidParamException("Operator '$operator' requires exactly one operand.");
+        }
+
+        return (new static(array_shift($operands)));
```
samdark: `(` and `)` aren't needed.

framework/db/conditions/SimpleCondition.php
```diff
@@ -0,0 +1,78 @@
...
+     * {@inheritdoc}
+     * @throws InvalidParamException if wrong number of operands have been given.
+     */
+    public static function fromArrayDefinition($operator, $operands)
+    {
+        if (count($operands) !== 2) {
+            throw new InvalidParamException("Operator '$operator' requires two operands.");
+        }
+
+        return (new static($operands[0], $operator, $operands[1]));
```
samdark: No need for `(` and `)`.

framework/db/cubrid/QueryBuilder.php
```diff
@@ -46,18 +46,14 @@ class QueryBuilder extends \yii\db\QueryBuilder
...
-     * @inheritdoc
+     * {@inheritdoc}
      */
-    protected $likeEscapingReplacements = [
-        '%' => '!%',
-        '_' => '!_',
-        '!' => '!!',
-    ];
-
+    protected function defaultExpressionBuilders()
```
samdark: phpdoc missing.

framework/db/mssql/QueryBuilder.php
```diff
@@ -46,16 +46,15 @@ class QueryBuilder extends \yii\db\QueryBuilder
...
      */
-    protected $likeEscapingReplacements = [
-        '%' => '[%]',
-        '_' => '[_]',
-        '[' => '[[]',
-        ']' => '[]]',
-        '\\' => '[\\]',
-    ];
-
+    protected function defaultExpressionBuilders()
```
samdark: phpdoc missing.

framework/db/oci/conditions/LikeConditionBuilder.php
```diff
@@ -0,0 +1,39 @@
...
+     * we need to choose replacement value based on [[\yii\db\Schema::quoteValue()]].
+     * @inheritdoc
+     */
+    protected $escapingReplacements = [
+        '%' => '!%',
+        '_' => '!_',
+        '!' => '!!',
+    ];
+
+    public function build(ExpressionInterface $condition, &$params = [])
```
samdark: phpdoc missing.

framework/db/pgsql/ArrayParser.php
```diff
@@ -0,0 +1,103 @@
...
+ * @author Sergei Tigrov <rrr-r@ya.ru>
+ * @author Dmytro Naumenko <d.naumenko.a@gmail.com>
+ * @since 2.0.14
+ */
+class ArrayParser
+{
+    /**
+     * @var string Character used in array
+     */
+    protected $delimiter = ',';
```
samdark: I think it should be private.

samdark (approved):
Overall looks solid. Some minor things added as comments.

framework/db/conditions/AndCondition.php
SilverFire: It will be added during release process automatically

framework/db/conditions/BetweenCondition.php
SilverFire: It will be added during release process automatically

framework/db/conditions/BetweenConditionBuilder.php
SilverFire: It will be added during release process automatically

framework/db/conditions/ConditionInterface.php
SilverFire: It will be added during release process automatically

framework/db/conditions/InConditionBuilder.php
SilverFire: It will be added during release process automatically

framework/db/conditions/LikeCondition.php
SilverFire: Yes

framework/db/cubrid/QueryBuilder.php
SilverFire: it is not indeed

framework/db/mssql/QueryBuilder.php
SilverFire: it is not indeed

framework/UPGRADE.md
```diff
@@ -58,6 +58,14 @@ Upgrade from Yii 2.0.13
 
 * `yii\base\Security::compareString()` is now throwing `yii\base\InvalidParamException` in case non-strings are compared.
 
+* `yii\db\ExpressionInterface` has been introduced to represent a wider range of SQL expressions. In case you check for
+  `instanceof yii\db\Expression` in your code, you might consider changing that to checking for the interface and use the newly
+  introduced methods to retrieve the expression content.
+
+* TODO: `['value', \PDO::PARAM_STR]` syntax deprecation.
```
samdark: Should be done before merge.

framework/UPGRADE.md
```diff
@@ -58,6 +58,14 @@ Upgrade from Yii 2.0.13
...
+
+* TODO: `yii\db\QueryBuilder::conditionBuilders` is no longer used.
```
samdark: Should be done before merge.

samdark (changes requested):

samdark (approved):

samdark:
Merged! That is a huge improvement over query builder architecture and I'm glad that Yii has it now. Will allow to easily extend QB syntax in any direction possible. Thank you very much for working on it and finishing it.

qwerty199369:
Excellent! Can't wait to use it! 👍

Blizzke:
Was there ever a deprecation period regarding using arrays for bindValues? 
Because the switch to `PdoValue` without having some kind of grace period broke `yii2-audit`.

SilverFire:
Yes, it is deprecated since 2.0.14 and will be completely removed in 2.1.
Until that time arrays are silently converted to `PdoValue` objects.

Blizzke:
I assume that you are referring to regular inserts only then?

I have a [bug report](https://github.com/bedezign/yii2-audit/issues/238) on the yii2-audit component. It stopped working since 2.0.14. 
We traced it back to the changes in `bindValues()`.

Our component directly calls bindValues() for batch insertion (it is a public function after all). 
The call [can be found here](https://github.com/bedezign/yii2-audit/blob/master/src/models/AuditEntry.php#L127) and has worked for the last 2 years but fails since this PR.

As it is now, there is no grace period for code directly calling `bindValues()` since the `is_array` part was completely replaced:

screen shot 2018-03-02 at 09 27 45

SilverFire:
This break was not intended. Could you create a separate issue? We can probably fix it in 2.0.14.2

Issue #15353, opened 2017-12-13
Remove side effect of ActiveQuery::getTablesUsedInFrom() introduced in 2.0.13


terales:
Hi @cebe, thanks for your hard work on the framework!

#### In 2.0.12 ([link to source](https://github.com/yiisoft/yii2/blob/2.0.12/framework/db/ActiveQuery.php#L804-L805)):
```php
<?php
if (empty($this->from)) {
    $tableNames = [$this->getPrimaryTableName()];

// …
// cleaning up tables, no $this->from assignment in method
// …
return $cleanedUpTableNames;

```

#### In 2.0.13+ (introduced in https://github.com/yiisoft/yii2/commit/173108b5b05a6d0c4222dd46360b9b59be5abc57):
https://github.com/yiisoft/yii2/blob/173108b5b05a6d0c4222dd46360b9b59be5abc57/framework/db/ActiveQuery.php#L797-L802

### Should I prepare a PR to Yii repo to remove introduced side effect?

What was your motivation to assign a `from` property in `getTablesUsedInFrom` method?
Does it fix any bugs?

This side effect broke our application. We would rewrite it to have a better design, but I wanted to inform you about this change in case it was unconscious change.

- - -

#### What steps will reproduce the problem?

1. Implement `SomeActiveQuery` to be used by all models in the app. This class has `prepare` method which adds additional conditions to all queries with empty `ActiveQuery->from` property.
1. Run `yii2/validators/UniqueValidator`.

#### What is the expected result?

The Query has the additional condition.

#### What do you get instead?

The Query has no expected additional condition.

#### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13, 2.0.13.1
| PHP version      | 7.1
| Operating system | Windows 10 x64

terales:
Failing test: https://travis-ci.org/yiisoft/yii2/jobs/315987955#L688-L706

PR #15354, opened 2017-12-13
Remove side effect of ActiveQuery::getTablesUsedInFrom() introduced in 2.0.13

terales:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15353


samdark:
Tests aren't passing.

terales:
@samdark thanks for attention,

This was done by intention to help @cebe easier check what's going on with this comment: https://github.com/yiisoft/yii2/issues/15353#issuecomment-351456253

I'm waiting for clarification from @cebe. If this side-effect was introduced accidentally than I would make tests pass, but if there is some deep idea than I would close PR and remove branch.

samdark:
Ah... let me take a look. @cebe is very busy now.

samdark:
Oh, yes. This side effect wasn't intended, I think.

terales:
Sure!

terales:
Checked Codeclimat failures: they are connected to the extracting table names clearing in another method. I think refactoring of this function should be in another issue if you feel that it should be refactored.

samdark:
It is fine as is.

framework/db/Query.php
```diff
@@ -474,7 +474,17 @@ public function getTablesUsedInFrom()
...
 
-        // Clean up table names and aliases
+        return $this->cleanUpTableNames($tableNames);
+    }
+
+    /**
+     * Clean up table names and aliases
+     * Both aliases and names are enclosed into {{ and }}.
+     * @param array $tableNames non-empty array
+     * @return string[] table names indexed by aliases
```
samdark: Need since tag.

samdark (approved):
All good. Need changelog and one since tag in phpdoc. Otherwise good to merge.

terales:
Done!

Thanks a lot for your hard work on Yii in general and with pushing modular approach into the core in 2.1!

samdark:
Merged. Thank you!

PR #15357, opened 2017-12-14
Add multi statement support for SQLite’s Command

sergeymakinen:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | see #15276 

It's required to perform upsert statements. Currently it's only supports named placeholders. Position placeholders will lead to errors. I think it can be adjusted later.

This PR has to be merged before #15276. `$commandMap` will get more custom `Command` classes in #15276.


framework/db/sqlite/Command.php
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Splits the specified SQL code into individual SQL statements and returns them
+     * or `false` if there's a single statement.
+     * @param string $sql
+     * @param array $params
+     * @return string[]|false
+     */
+    private function splitStatements($sql, $params)
+    {
+        $semicolonIndex = strpos($sql, ';');
```
samdark: There could be semicolon in string content:

INSERT INTO post (content) VALUES ('Hello; world!');

samdark:
Overall looks good but special cases with `;` in strings should be tested.

framework/db/sqlite/Command.php
sergeymakinen: It's okay because it is a "fail fast" condition to avoid parsing at all.

samdark (approved):

framework/db/Connection.php
```diff
@@ -687,12 +712,17 @@ protected function initConnection()
      */
     public function createCommand($sql = null, $params = [])
     {
+        $driver = $this->getDriverName();
+        $config = ['class' => 'yii\db\Command'];
+        if ($this->commandClass !== $config['class']) {
```
samdark: Looks suspicious. Have you meant to do this check after `if (isset($this->commandMap[$driver])) {`?

samdark:
Merged. Thank you very much!

PR #15385, opened 2017-12-19
Allow $variations to be a string for PageCache & FragmentCache

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Replacement for #9253

### DIFF
--- a/framework/filters/PageCache.php
+++ b/framework/filters/PageCache.php
@@ -88,7 +88,7 @@ class PageCache extends ActionFilter
      */
     public $dependency;
     /**
-     * @var array list of factors that would cause the variation of the content being cached.
+     * @var string[]|string list of factors that would cause the variation of the content being cached.
      * Each factor is a string representing a variation (e.g. the language, a GET parameter).
      * The following variation setting will cause the content to be cached in different versions
      * according to the current application language:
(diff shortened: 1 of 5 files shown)


framework/filters/PageCache.php
```diff
@@ -323,12 +323,6 @@ protected function calculateCacheKey()
...
             $key[] = Yii::$app->requestedRoute;
         }
-        if (is_array($this->variations)) {
-            foreach ($this->variations as $value) {
-                $key[] = $value;
-            }
-        }
-
-        return $key;
+        return array_merge($key, (array)$this->variations);
```
samdark: These are not equivalent. `array_merge` keeps keys while `foreach` that was there does not.

samdark: Are you sure it's OK w/ keys?

developeruz: It should be ok. [it is used as a key for cache](https://github.com/yiisoft/yii2/blob/b0aa7a73de8f972cb4e02e6ae6d3896b0f1a9a11/framework/caching/CacheInterface.php#L97)
> $key a key identifying the value to be cached. This can be a simple string or a complex data structure consisting of factors representing the key
 
If someone introduced array with keys for `variation` it should IMO affect as a factor

samdark (approved):

samdark:
Merged. Thank you!

PR #15387, opened 2017-12-19
Fixed wrong description for param $table on addCommentOnTable()

InsaneSkull:
Migration addCommentOnTable() had wrong description for param $table.


samdark:
Merged. Thanks!

PR #15388, opened 2017-12-19
How to disable form validation in REST API [skip ci]

bscheshirwork:
note: `enableCookieValidation: true` can rise error if `cookieValidationKey` is not set
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no


samdark:
As far as I remember, it works out of the box w/o configuring components...

samdark:
https://github.com/yiisoft/yii2/blob/master/framework/rest/Controller.php#L42

bscheshirwork:
So... Yes. I can't find `enableCsrfValidation` in rest controller, my bad.

But I have error with `enableCookieValidation` "out of the box".

`enableCookieValidation` must set to 'false' too...

Issue #14462, opened 2017-07-16
Error: Call to undefined function yii\base\fnmatch()


tonygeli:
### What steps will reproduce the problem?
Error: Call to undefined function yii\base\fnmatch()

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.0.0
| Operating system | Linux

cebe:
http://php.net/manual/en/function.fnmatch.php

>  Warning: For now, this function is not available on non-POSIX compliant systems except Windows. 

Can you give more details about what system you are on?

WalkingSun:
I get a same error message.
PHP version          :       5.6.31
Operating system :  CentOS 6.6
Yii version            ;   2.0.11.2

cebe:
https://unix.stackexchange.com/questions/305639/centos-6-7-and-fnmatch-in-php#comment537107_305639 ?

Issue #12408, opened 2016-08-31
a bug  about   the   function   fnmatch


donblack01:
### What steps will reproduce the problem?
   on  the  centos7  install  yii2 2.0.9   

when  I  installed yii2   I  access    http：//localhost/basic/web    show  this problem like  this ：’Call to undefined function yii\base\fnmatch()   in /usr/local/nginx/html/basic/vendor/yiisoft/yii2/base/ActionFilter.php line 150‘  ,
I refer to the PHP manual  this function     support to  php7  so  i do  not  kown  how  to  solve it

### What do you get instead?
  no  

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.9
| PHP version      |    7.0.10
| Operating system | centos7

donblack01:
![9pa57 la073nbj b6a fc](https://cloud.githubusercontent.com/assets/21120018/18123154/3349907e-6f9f-11e6-9ad9-c0163dc7f5ac.png)

klimov-paul:
I am afraid there is nothing I can help you with. This is definetely not Yii issue.

`ActionFilter` wildcard support is covered by unit test:
https://github.com/yiisoft/yii2/blob/master/tests/framework/base/ActionFilterTest.php#L141

which is correctly passing on PHP 7:
https://travis-ci.org/yiisoft/yii2/jobs/156473399

php.net docs clearly says `fnmatch()` function is a part of the core and should be available at PHP7.

This must be problem with your environment.

Alex-Code:
According to a comment here http://unix.stackexchange.com/questions/305639/centos-6-7-and-fnmatch-in-php#comment537104_305639 this function should be available in the php packages provided by CentOS but may not if you'd compiled it yourself.

PR #15389, opened 2017-12-19
Add `StringHelper::matchWildcard()`

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12408, #14462

Add `StringHelper::matchWildcard()` replacing `fnmatch()`, which causing relatively many issues.

### DIFF
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -381,7 +381,7 @@ class BaseStringHelper
      */
     public static function matchWildcard($pattern, $string, $options = [])
     {
-        if ($pattern === '*') {
+        if ($pattern === '*' && empty($options['filePath'])) {
             return true;
         }
 
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -374,6 +374,7 @@ class BaseStringHelper
      * @param array $options options for matching. Valid options are:
      *
      * - caseSensitive: bool, whether pattern should be case sensitive. Defaults to `true`.
+     * - escape: bool, whether backslash escaping is enabled. Defaults to `true`.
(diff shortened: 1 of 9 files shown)


dmirogin (approved):
Cool! LGTM

samdark:
Overall looks OK but complementation support is missing while it is [there in fnmatch](https://linux.die.net/man/7/glob):

`[!ab]` = NOT `a` AND NOT `b`.

Other test cases from docs should be checked as well.

samdark:
Added more tests.

klimov-paul:
@samdark, some of the tests you added fails for native `fnmatch()`:

```php
$this->assertTrue(fnmatch('\*\?', '*?'));
$this->assertTrue(fnmatch('[-1]', '1--'));
$this->assertFalse(fnmatch('*', '.dotenv', FNM_PATHNAME));
```

I have replaced them, other are fixed.

framework/helpers/BaseStringHelper.php
```diff
@@ -374,6 +374,7 @@ public static function floatToString($number)
      * @param array $options options for matching. Valid options are:
      *
      * - caseSensitive: bool, whether pattern should be case sensitive. Defaults to `true`.
+     * - escape: bool, whether backslash escaping is enabled. Defaults to `true`.
```
samdark: Why disabling it?

samdark (approved):

SilverFire (approved):

framework/helpers/BaseStringHelper.php
klimov-paul: This option matches `FNM_NOESCAPE` flag of `fnmatch()`.
Its disabling may be useful while processing patterns for strings with backslashes such as file path for Windows. So you can write `'begin\*\end'` instead of `'begin\\\\*\\\\end'`.

PR #15390, opened 2017-12-19
Fixes #15046

### LINKED ISSUE #15046: web/Response::sendHeaders does not warn if headers are already send
I experienced problems with CSRF similar to #541 #7115 . CSRF cookie just was not sent in response to requests.

It turned out that one of my files echoed some content before `View::beginPage()` call.

To find it, I had to debug through all the application up to Response::sendHeaders(), where I could find the root cause of the issue dumping `$file` and `$line` from `headers_sent($file, $line)`.

The question is: why Yii does not warn/throw an error if headers were already sent prior to `Response::send()`? This leads to silent and hard-to-debug problems.

I think simple check before `$this->sendHeaders()` in `Response::send()`, that will throw an Exception if `headers_sent)` is true, will solve this issue.

I can make a PR for this if this solutuion is ok.

dmirogin:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | I think worth mentioning this in upgrade
| Tests pass?   | yes
| Fixed issues  | #15046


framework/web/HeadersAlreadySentException.php
```diff
@@ -0,0 +1,33 @@
...
+class HeadersAlreadySentException extends Exception
+{
+    /**
+     * Create instance of exception depending on debug mode.
+     *
+     * @param string $file
+     * @param int $line
+     * @return HeadersAlreadySentException
+     */
+    public static function make($file, $line)
```
samdark: Why not using constructor?

dmirogin: I dont want to change constructor params.

samdark: Why?

dmirogin: Just prefer this style. But in this case it's redundant.

dmirogin:
I don't know how to tests it, because cli SAPI doesn't support headers access.

samdark:
Right. Let's skip it this time.

samdark:
Merged. Thanks!

framework/web/Response.php
```diff
@@ -362,8 +362,8 @@ public function clear()
      */
     protected function sendHeaders()
     {
-        if (headers_sent()) {
-            return;
+        if (headers_sent($file, $line)) {
+            throw new HeadersAlreadySentException($file, $line);
```
lubosdz: This is a serious BC break - disallows setting flexibly custom headers anywhere in the code, e.g. when downloading files.

samdark: You were not able to set them anyway. It was just silently ignored.

lubosdz: We used to set various custom headers directly e.g. 
`header('Content-Type: application/octet-stream', false);`
Now everybody who used to set inline headers will get Exception. With already 14th branch release there might be heaps of code to be refactored :-( It would be better to leave such a BC change for 2.1.X. I am not sure what you mean `not able to set them anyway` - it was possible via `$response->getHeaders()->add()`, right? But we did not use it coz it's sometime easier to write native PHP headers with second `replace` option.

samdark: If there's now an exception then these headers weren't working well before.

Issue #15356, opened 2017-12-13
Multiple bugs in yii\db\Query::getTablesUsedInFrom()


PowerGamer1:
```PHP
// BUG1: Invalid result when "from" is \yii\db\Expression:
$query = (new \yii\db\Query)->from(new \yii\db\Expression('(SELECT id from user) x'));
var_dump($query->getTablesUsedInFrom()); // array(1) { ["{{id from user) x}}"]=> string(11) "{{(SELECT}}" }

// BUG2: PHP runtime error when "from" is \yii\db\Query with alias:
$query = (new \yii\db\Query)->from(['x' => (new \yii\db\Query)->from('user')]);
$query->getTablesUsedInFrom(); // Object of class yii\db\Query could not be converted to string

// BUG3: Alias and table name are not enclosed into {{ and }}:
$query = (new \yii\db\Query)->from(['x' => 'tbl']);
var_dump($query->getTablesUsedInFrom()); // array(1) { ["x"]=> string(3) "tbl" }
```
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13.1
| PHP version      |  7.0.25
| Operating system | Windows

vladis84:
Hello.
BUG2 expected array `['{{u}}' => '{{user}}']`

Example 2
```php
$subQuery =  new \yii\db\Query;
$subQuery ->from(['user', 'orders']);

$query = new \yii\db\Query;
$query->from(['x' => $subQuery]);

$query->getTablesUsedInFrom();
```
expected array `['{{x}}' => ['{{user}}', '{{orders}}']]`

Am i right?

samdark:
@vladis84 2nd.

samdark:
@vladis84 neither since there's no aliased table used.

PowerGamer1:
I don't like how `getTablesUsedInFrom()` processes `yii\db\Expression` now. By definition `yii\db\Expression` must be treated as a black box, not analyzed or processed in any way and inserted into generated SQL as is. I think `getTablesUsedInFrom()` should just throw exception (`throw new InvalidConfigException(gettype($this->from) . ' in $from is not supported.');`) when `$this->from` is a `\yii\db\Expression`. This won't break any existing code btw, because so far neither `\yii\db\Query` nor `getTablesUsedInFrom()` worked correctly at all when `from` was `yii\db\Expression`.

samdark:
Table names aren't used in the code generation at all.

[...]

PR #15395, opened 2017-12-20
Bug #15356: Fixed multiple bugs in `yii\db\Query::getTablesUsedInFrom()`

vladis84:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15356

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -470,6 +470,8 @@ class Query extends Component implements QueryInterface
             $tableNames = $this->from;
         } elseif (is_string($this->from)) {
             $tableNames = preg_split('/\s*,\s*/', trim($this->from), -1, PREG_SPLIT_NO_EMPTY);
+        } elseif ($this->from instanceof Expression) {
+            return (string) $this->from;
         } else {
             throw new InvalidConfigException(gettype($this->from) . ' in $from is not supported.');
         }
@@ -477,6 +479,11 @@ class Query extends Component implements QueryInterface
         // Clean up table names and aliases
         $cleanedUpTableNames = [];
         foreach ($tableNames as $alias => $tableName) {
+            if ($tableName instanceof self) {
+                $cleanedUpTableNames += $tableName->getTablesUsedInFrom();
+                
+                continue;
+            }
             if (!is_string($alias)) {
                 $pattern = <<<PATTERN
 ~
@@ -514,12 +521,6 @@ PATTERN;
                         } else {
                             $tableName = $alias = $matches[1];
                         }
-                        if (strncmp($alias, '{{', 2) !== 0) {
-                            $alias = '{{' . $alias . '}}';
-                        }
-                        if (strncmp($tableName, '{{', 2) !== 0) {
-                            $tableName = '{{' . $tableName . '}}';
-                        }
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


tests/framework/db/QueryTest.php
```diff
@@ -583,4 +583,110 @@ public function testMultipleLikeConditions()
...
+        $this->assertEquals($expected, $tables);
+    }
+
+    public function testGetTablesUsedInFromIsQueryFrom()
+    {
+        $query = new Query;
+        $subQuery = new Query;
+        $subQuery->from('user');
+        $query->from(['x' => $subQuery]);
+        $expected = ['{{user}}' => '{{user}}'];
```
vladis84: I'm not sure what's right here.

tests/framework/db/QueryTest.php
```diff
@@ -583,4 +583,109 @@ public function testMultipleLikeConditions()
             '%ba',
         ]));
     }
+
+    public function testGetTablesUsedInFromEmptyFrom()
```
samdark: Tests should be moved into `GetTablesAliasTestTrait`.

tests/framework/db/QueryTest.php
samdark:
```php
'x' => $subQuery
```

You've specified `x` as alias explicitly and table name there doesn't make sense since it's sub-query.

tests/framework/db/QueryTest.php
```diff
@@ -583,4 +583,109 @@ public function testMultipleLikeConditions()
...
+
+    public function testGetTablesUsedInFromIsExpressionFrom()
+    {
+        $query = new Query;
+        $expression = new \yii\db\Expression('(SELECT id from user) x');
+        $query->from($expression);
+
+        $tables = $query->getTablesUsedInFrom();
+
+        $this->assertEquals('(SELECT id from user) x', $tables);
```
samdark: That's not correct. It should return an array of strings, not string.

tests/framework/db/QueryTest.php
```diff
@@ -583,4 +583,109 @@ public function testMultipleLikeConditions()
...
+    }
+
+    public function testGetTablesUsedInFromIsQueryStringFrom()
+    {
+        $query = new Query;
+        $subQuery = "(SELECT * FROM `user` WHERE `active` = 1)";
+        $query->from(['activeusers' => $subQuery]);
+        $expected = ['{{activeusers}}' => '{{(SELECT * FROM user WHERE active = 1)}}'];
```
samdark: That looks suspicious.

samdark (changes requested):

framework/db/Query.php
```diff
@@ -470,13 +470,20 @@ public function getTablesUsedInFrom()
             $tableNames = $this->from;
         } elseif (is_string($this->from)) {
             $tableNames = preg_split('/\s*,\s*/', trim($this->from), -1, PREG_SPLIT_NO_EMPTY);
+        } elseif ($this->from instanceof Expression) {
+            return (string) $this->from;
```
samdark: Doesn't looks correct. `getTablesUsedInFrom()` return array of strings, not a string.

samdark:
I'll take a look at it today.

samdark:
Incorporated your commits into new pull request.

PR #15396, opened 2017-12-20
Fixed multiple bugs in `yii\db\Query::getTablesUsedInFrom()`

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15356. Replaces #15395


framework/db/Query.php
```diff
@@ -466,18 +466,29 @@ public function getTablesUsedInFrom()
...
         if (empty($this->from)) {
             return [];
-        } elseif (is_array($this->from)) {
+        }
+
+        if (is_array($this->from)) {
             $tableNames = $this->from;
         } elseif (is_string($this->from)) {
             $tableNames = preg_split('/\s*,\s*/', trim($this->from), -1, PREG_SPLIT_NO_EMPTY);
+        } elseif ($this->from instanceof Expression) {
```
SilverFire: Should it be replaced with `ExpressionInterface` after merging that PR?

framework/db/Query.php
```diff
@@ -466,18 +466,29 @@ public function getTablesUsedInFrom()
...
 
         // Clean up table names and aliases
         $cleanedUpTableNames = [];
         foreach ($tableNames as $alias => $tableName) {
-            if (!is_string($alias)) {
+            $tableNameString = null;
+            if (is_string($tableName)) {
+                $tableNameString = $tableName;
+            } elseif ($tableName instanceof Expression) {
+                $tableNameString = $tableName->expression;
```
SilverFire: Again, in terms of that PR, this code support only `Expression` class, or any `ExpressionInterface`?

framework/db/Query.php
```diff
@@ -507,32 +520,42 @@ public function getTablesUsedInFrom()
...
+                        $tableNameString = $alias = $matches[1];
                     }
                 }
             }
 
-            $tableName = str_replace(["'", '"', '`', '[', ']'], '', $tableName);
-            $alias = str_replace(["'", '"', '`', '[', ']'], '', $alias);
-
-            $cleanedUpTableNames[$alias] = $tableName;
+            if ($tableName instanceof Expression) {
```
SilverFire: Same here

SilverFire (approved):
Good otherwise

framework/db/Query.php
samdark: Yes.

framework/db/Query.php
samdark: `ExpressionInterface` isn't in master yet.

samdark: I think in this case it's about class.

samdark:
Thank you @vladis84 for starting it.

PR #15398, opened 2017-12-21
Copy query cache from `yii\db\Command` to `yii\db\Query::createCommand()` and `yii\db\ActiveQuery::createCommand()`.

hubeiwei:
Q | A
-- | --
Is bugfix? | no
New feature? | yes
Breaks BC? | no
Tests pass? | yes
Fixed issues | -


hubeiwei:
Command will check the cache in the query, Query should also be able to use the cache

```
(new Query())->cache(7200)->all();
User::find()->cache(7200)->all();
```

samdark:
I think http://www.yiiframework.com/doc-2.0/guide-caching-data.html#query-caching is enough as a way to achieve it.

@yiisoft/reviewers, @yiisoft/core-developers need opinions.

rob006:
@samdark Is there any reason to not implementing it? IMO it may be useful and will improve consistency.

samdark:
Two reasons:

1. Another way to do the same thing.
2. We have to support it.

hubeiwei:
`yii\db\Connection: :enableQueryCache` defaults to true, `Command` will first check the cache before query, I think you can take advantage of this.

rob006:
@samdark `yii\db\Command` already has such method - it is quite weird that low-level DB component (`Command`) has such shortcut and high-level DB component (`Query`) does not.

samdark:
@rob006 the idea of having it only at connection level was essentially that it can wrap anything and there's no need for more ways to handle it.

rob006:
> the idea of having it only at connection level

The idea seems to be dropped, since we have:

https://github.com/yiisoft/yii2/blob/e9bec0d24653bf5d330f677f53aafd8c4a7feb1a/framework/db/Command.php#L123

samdark:
While I personally don't like the idea much, if you think it's fine I'm OK merging it.

@cebe, @SilverFire, @klimov-paul thoughts?

SilverFire:
We already have 2 ways to cache a query:

1. 
```php
$db->cache(function($db) {
    return User::find()->all();
});
```

2.
```php
$cache->getOrSet(['meaningful-key'], function () {
    return User::find()->all();
})
```

But the suggested one can be a shorter and more convenient. 

I'd suggest to move properties `queryCacheDuration` and `queryCacheDependency` together with `cache()`, `noCache()` and `hasCache()` methods to the `CacheableQueryTrait`, that can be used in `Query` and `Command`.

SilverFire (changes requested):
See comment above. Otherwise good to go

hubeiwei:
@SilverFire `Command::cache()` have default cache duration, through `$this->db->queryCacheDuration`, but `Query` no `$db` this member variables. if want to use `CacheableQueryTrait`, I think to give it up, need opinions.

rob006:
@SilverFire Why did you merge this without any tests? I'm pretty sure that this will not work as expected: `MyModel::find()->cache()->one($myCustomDb)`. Default `queryCacheDuration` will be taken from `Yii::$app->db` instead of `$myCustomDb`. 

Also `hasCache()` is misleading, since it does not say whether the query has cache, it only says that cache is enabled. I'm also not sure why it is public or whether implementation is correct (does it make sense do enable cache with `queryCacheDuration === null` but `queryCacheDependency !== null`?).

SilverFire:
It was accidentally, I'm now working on fix

SilverFire:
See 5bd6ed

SilverFire:
See 5bd6ed5

rob006:
I would keep `cache()` API consistent with `Command::cache()` and use `$duration === null` as "use `Connection::$queryCacheDuration`". We could just add private `Query::$isCacheEnabled` and pass `$queryCacheDuration` and `$queryCacheDependency` directly to command - it will simplify the code and API.

PR #15400, opened 2017-12-22
fix ArrayAccessTrait offsetExists

yinggaozhen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | None


klimov-paul:
Change is incorrect as `ArrayAccess::offsetExists()` is invoked only for `isset()` or `empty()` operator, both of which returns false on `null`.

PR #15404, opened 2017-12-22
Added event wildcard matching

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #3250, #14978

This solution created in attempt to keep good performance.
In case no wildcard even specified no significant performance degradation will appear.

### DIFF
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -494,6 +494,14 @@ class Component extends BaseObject
      *
      * where `$event` is an [[Event]] object which includes parameters associated with the event.
      *
+     * Since 2.0.14 you can specify event name as a wildcard pattern:
+     *
+     * ```php
+     * $component->on('event.group.*', function ($event) {
+     *     Yii::trace($event->name . ' is triggered.');
+     * });
+     * ```
+     *
      * @param string $name the event name
      * @param callable $handler the event handler
      * @param mixed $data the data to be passed to the event handler when the event is triggered.
@@ -525,7 +533,12 @@ class Component extends BaseObject
 
     /**
      * Detaches an existing event handler from this component.
+     *
      * This method is the opposite of [[on()]].
+     *
+     * Note: in case wildcard pattern is passed for event name, only the handlers registered with this
+     * wildcard will be removed, while handlers registered with plain names matching this wildcard will remain.
+     *
      * @param string $name event name
      * @param callable $handler the event handler to be removed.
      * If it is null, all handlers attached to the named event will be removed.
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -465,7 +465,7 @@ class Component extends BaseObject
         $this->ensureBehaviors();
 
         foreach ($this->_eventWildcards as $wildcard => $handlers) {
-            if (StringHelper::matchWildcard($wildcard, $name) && !empty($handlers)) {
+            if (!empty($handlers) && StringHelper::matchWildcard($wildcard, $name)) {
                 return true;
             }
         }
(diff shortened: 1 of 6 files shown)


samdark:
Love the approach and the code looks good at the first glance. Going to take a break and re-read it several times.

framework/base/Component.php
```diff
@@ -456,6 +463,13 @@ public function behaviors()
     public function hasEventHandlers($name)
     {
         $this->ensureBehaviors();
+
+        foreach ($this->_eventWildcards as $wildcard => $handlers) {
+            if (StringHelper::matchWildcard($wildcard, $name) && !empty($handlers)) {
```
samdark: Condition could be reverted. empty check costs less.

framework/base/Event.php
```diff
@@ -110,23 +137,40 @@ public static function on($class, $name, $handler, $data = null, $append = true)
     public static function off($class, $name, $handler = null)
     {
         $class = ltrim($class, '\\');
-        if (empty(self::$_events[$name][$class])) {
+        if (empty(self::$_events[$name][$class]) && empty(self::$_eventWildcards[$name][$class])) {
             return false;
         }
         if ($handler === null) {
             unset(self::$_events[$name][$class]);
```
samdark: Could be written as `unset(self::$_events[$name][$class], self::$_eventWildcards[$name][$class])`.

framework/base/Event.php
```diff
@@ -153,9 +198,10 @@ public static function offAll()
      */
     public static function hasHandlers($class, $name)
     {
-        if (empty(self::$_events[$name])) {
+        if (empty(self::$_events[$name]) && empty(self::$_eventWildcards)) {
```
samdark: Conditions could be inverted. Checking single variable costs less then accessing array index.

framework/base/Event.php
```diff
@@ -187,9 +251,20 @@ class_implements($class, true)
...
+        if (!empty(self::$_eventWildcards)) {
+            foreach (self::$_eventWildcards as $nameWildcard => $classHandlers) {
+                if (!StringHelper::matchWildcard($nameWildcard, $name)) {
+                    continue;
+                }
+                $wildcardEventHandlers = array_merge($wildcardEventHandlers, $classHandlers);
+            }
+        }
+
+        if (empty(self::$_events[$name]) && empty($wildcardEventHandlers)) {
```
samdark: Conditions could be inverted. Checking single variable costs less then accessing array index.

samdark (approved):
Overall looks great. Haven't found much things to improve.

framework/base/Component.php
klimov-paul: optimized

framework/base/Event.php
klimov-paul: optimized

framework/base/Event.php
klimov-paul: optimized

framework/base/Event.php
klimov-paul: I would better not do it: arguments are long, such refactoring will create a long statement, which is hard to comprehend.

SilverFire (approved):

dmirogin (approved):

sergeymakinen (approved):

samdark:
That was good. Thank you, Paul and thanks everyone for reviews.

macklus:
thanks on that ;-)

Issue #15226, opened 2017-11-26
Auto generate placeholder


nicdnepr:
If I want use placeholder with same label attribute need use code
```
<?= $form->field($model, 'customer_name')->textInput(['placeholder' => $model->getAttributeLabel('customer_name')])->label(false); ?>
```
Why no options to auto generate placeholder from AttributeLabel?

developeruz:
Ups, @vladis84 it seems we did the same work :) I believe that this feature should be applied to all types of `Html::activeInput` like, for example, password field.

vladis84:
You're right about that

SilverFire:
Resolved by #15405

PR #15406, opened 2017-12-24
Auto generate placeholder for `Html::activeInput`

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15226

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1313,6 +1313,9 @@ class BaseHtml
         if (!array_key_exists('id', $options)) {
             $options['id'] = static::getInputId($model, $attribute);
         }
+        if (!array_key_exists('placeholder', $options) && $type!='hidden') {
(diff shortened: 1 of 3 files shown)


framework/helpers/BaseHtml.php
```diff
@@ -1313,6 +1313,9 @@ public static function activeInput($type, $model, $attribute, $options = [])
         if (!array_key_exists('id', $options)) {
             $options['id'] = static::getInputId($model, $attribute);
         }
+        if (!array_key_exists('placeholder', $options) && $type!='hidden') {
```
vladis84: I think that the placeholder should be indicated.
Example ['placeholder' => true].

vladis84:
I don't see generate placeholder for textarea.

framework/helpers/BaseHtml.php
developeruz: Agree

Issue #15391, opened 2017-12-20
Proposal: RequiredValidator "not" setting


vercotux:
Adding a simple `$not` boolean property to [RequiredValidator](http://www.yiiframework.com/doc-2.0/yii-validators-requiredvalidator.html) would inverse the check to only pass validation when the attribute *is* empty. Same principle as is already used with [RegularExpressionValidator::$not](http://www.yiiframework.com/doc-2.0/yii-validators-regularexpressionvalidator.html#$not-detail)

This is useful when doing conditional validation where you do not want an attribute to be set under certain conditions.

samdark:
Would you please provide an example?

vercotux:
Sure thing.

In your model:

```php
public function rules() {
    return [
        ['thing', 'required'],
        ['thing', 'in',
            'range' => ['car', 'bicycle', 'boat'],
        ],
        ['wheel_size', 'required',
            'not' => true,
            'when' => function($model) {
                return $model->thing == 'boat';
            },
            'message' => "Boats have no wheels, silly!",
        ],
    ];
}
```

Now you can still do this

```php
(new MyModel([
    'thing'  => 'car',
]))->validate();

(new MyModel([
    'thing'  => 'car',
    'wheel_size' => 10,
]))->validate();

(new MyModel([
    'thing'  => 'boat',
]))->validate();
```

But you will get stopped by validation if you try to do this:

```php
(new MyModel([
    'thing'  => 'boat',
    'wheel_size' => 10,
]))->validate();
```

samdark:
Would you like to help with a pull request (code + tests)?

PR #15409, opened 2017-12-24
Added `not` parameter to the `yii\validators\RequiredValidator`

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15391

I am not sure about naming because "not required" doesn't mean "prohibited". Also, I need some help with client-side validation.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -53,9 +53,6 @@ Yii Framework 2 Change Log
 - Enh #15360: Refactored `BaseConsole::updateProgress()` (developeruz)
 - Enh #15415: Added transaction/retry support for `yii\db\Command` (sergeymakinen)
 - Enh #15391: Added `not` parameter to the `yii\validators\RequiredValidator` (developeruz)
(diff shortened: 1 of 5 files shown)


developeruz:
I've added client-side validation and JS test

samdark (approved):

rob006:
How does it differ from `CompareValidator`?

developeruz:
@rob006, as I know, there is no way to compare with null, so it impossible check on empty or not empty value with `CompareValidator`

rob006:
@developeruz You can always compare to empty string. 

Anyway, using `RequireValidator` for prohibiting some values is really confusing. :/

samdark:
Hmm... yes, it actually is confusing semantically. @developeruz are you OK with closing the pull request and the issue as "won't fix"?

developeruz:
Yes, sure

PR #15416, opened 2017-12-26
`Psr\Http\Message\ServerRequestInterface` applied for `yii\web\Request`

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #11328, support for #15294

Fixing `yii\web\Request` to implement `Psr\Http\Message\ServerRequestInterface` instead of just `Psr\Http\Message\RequestInterface`


framework/web/Request.php
```diff
@@ -685,44 +697,55 @@ public function getBodyParams()
...
     /**
      * Sets the request body parameters.
      * @param array $values the request body parameters (name-value pairs)
-     * @see getBodyParam()
-     * @see getBodyParams()
+     * @see getParsedBodyParam()
+     * @see getParsedBody()
      */
-    public function setBodyParams($values)
+    public function setParsedBody($values)
```
samdark: Do we still need to have `setParsedBody()` as public?

framework/web/Request.php
```diff
@@ -817,6 +854,82 @@ public function getQueryParam($name, $defaultValue = null)
         return isset($params[$name]) ? $params[$name] : $defaultValue;
     }
 
+    /**
+     * Sets the data related to the incoming request environment.
+     * @param array $serverParams server parameters.
+     * @since 2.1.0
+     */
+    public function setServerParams(array $serverParams)
```
samdark: There are `with*` methods for everything but `serverParams`. I think it makes sense to make it uniform having only `with*` public methods.

framework/web/Request.php
```diff
@@ -817,6 +854,82 @@ public function getQueryParam($name, $defaultValue = null)
...
+        }
+        return $params[$name];
+    }
+
+    /**
+     * Specifies cookies.
+     * @param array $cookies array of key/value pairs representing cookies.
+     * @since 2.1.0
+     */
+    public function setCookieParams(array $cookies)
```
samdark: Do we need it public?

SilverFire (approved):

framework/web/Request.php
```diff
@@ -1753,6 +1871,17 @@ public function setUploadedFiles($uploadedFiles)
         $this->_uploadedFiles = $uploadedFiles;
```
samdark: Do we need it public?

framework/web/Request.php
```diff
@@ -1999,6 +2128,86 @@ private function validateCsrfTokenInternal($clientSuppliedToken, $trueToken)
...
+        if ($this->_attributes === null) {
+            $this->_attributes = $this->defaultAttributes();
+        }
+        return $this->_attributes;
+    }
+
+    /**
+     * @param array $attributes attributes derived from the request.
+     */
+    public function setAttributes(array $attributes)
```
samdark: Do we need it public?

framework/web/Request.php
```diff
@@ -1999,6 +2128,86 @@ private function validateCsrfTokenInternal($clientSuppliedToken, $trueToken)
...
+        $newInstance->setAttributes($attributes);
+        return $newInstance;
+    }
+
+    /**
+     * Returns default server request attributes to be used in case they are not explicitly set.
+     * @return array attributes derived from the request.
+     * @since 2.1.0
+     */
+    protected function defaultAttributes()
```
samdark: What's the usecase?

framework/web/Request.php
klimov-paul: I would rather keep it, since it was there before.
Besides having such public method can be useful while writing unit tests.

framework/web/Request.php
klimov-paul: According to PSR request attributes are determined by particular `Request` class logic.
Having separated method allows developer to define internal attributes while extending `yii\web\Request` without necessity of re-defining private field for thier actual storage.

framework/web/Request.php
samdark: > since it was there before.

That's not a good reason for a BC breaking release. As far as I understand PSR, whole idea is to make Request immutable i.e. no setters.

framework/web/Request.php
klimov-paul: All `with*` methods are dictated by PSR. Since PSR does not provide `withServerParams()` method, I see no reason for implementing it. If you find this inconsistent you should report it ot PHPFig.

Usage of `with*` methods is uneffective, since it causes object cloning. While configuring single `Request` instance via `with*` methods you will have it cloned 5-10 times.
Yii behaviors and filters would be unable to function via immutable setters as they may need to adjust passing `Request` or `Response` without being able to re-set thier origin.

Also keep in mind that while being cloned every instance of `yii\base\Component`, which `Request` is, looses any event handlers or behaviors (filters) attached to it.

framework/web/Request.php
klimov-paul: See https://github.com/yiisoft/yii2/pull/15416#discussion_r158818756

framework/web/Request.php
klimov-paul: See https://github.com/yiisoft/yii2/pull/15416#discussion_r158818756

framework/web/Request.php
klimov-paul: See https://github.com/yiisoft/yii2/pull/15416#discussion_r158818756

framework/web/Request.php
klimov-paul: Also keep in mind that removing setters in favor of `with*` methods will make impossible to configure `Request` instance via DI, even with the recent features from `yiisoft/di`. For example:

```php
[
    '__class' => 'yii\web\Request',
    'withCookieParams' => [....], // have no effect: calls `$request->withCookieParams()`, but looses returned new instance
]
```

samdark: The whole idea of the PSR is that Request object should be created once (passing everything needed in constructor) and then not being modified again except cloning via `with`.

klimov-paul: If you sure in which way it should be - feel free to make modifications.

samdark: Do we have any request behaviors currently?

samdark: Losing event handlers could be a problem. Need to check code again before I suggest anything...

samdark: Currently Request doesn't trigger anything and there are no known behaviors for it.

samdark:
Currently Request serves many responsibilities:

1. Filling itself with `$_GET`, `$_POST`, `$_SERVER` etc. considering trusted hosts and trusted headers.
2. Storing data obtained and giving it to end user via convenient methods.
3. Resolving request into route via `resolve()` method.
4. Generating and setting cookies/tokens for CSRF. Verifying CSRF tokens.

The idea of PSR-7 is immutability and it makes sense in request/response paradigm. After request is received and PSR-7 request object is created, it's set in stone. Attempt to modify it usually means a possible bug.

It makes sens to have multiple classes:

1. PSR-7 Request which is data container that serves purpose `2`: storing data and having convenient methods for its retrieval. It could be initialized via constructor. There are multiple advantages in having it separate. It's more lightweight. Middlewares don't need to re-parse request same as everything else. Also it will be very convenient in tests.
2. Request builder that has methods/settings to prepare data for PSR-7 Request.
3. Resolving request data into route isn't request responsibility so should be moved into router (application).
4. CSRF protection should be converted into middleware.

klimov-paul:
PR merged as an intermediate solution.
Further PSR-7 immutability disscussion should be handled at #15437

samdark:
OK. Thanks.

PR #15427, opened 2017-12-28
Update QueryBuilder.php

asmythe:
documentation is wrong ... smallint() doesn't work, but smallinteger() does

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | na
| Fixed issues  | none

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -815,7 +815,7 @@ class QueryBuilder extends \yii\base\BaseObject
      * - `char`: char type, will be converted into "char(1)"
      * - `string`: string type, will be converted into "varchar(255)"
      * - `text`: a long string type, will be converted into "text"
-     * - `smallint`: a small integer type, will be converted into "smallint(6)"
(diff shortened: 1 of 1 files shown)


SilverFire:
We have a test for this case.

https://github.com/yiisoft/yii2/blob/399dbce0cadbd7631f726841ece3ecf6a830444d/tests/framework/db/QueryBuilderTest.php#L1019


Passing `SMALLINT` constant produces expected result:

https://github.com/yiisoft/yii2/blob/399dbce0cadbd7631f726841ece3ecf6a830444d/tests/framework/db/QueryBuilderTest.php#L652

`SMALLINT` constant is a `smallint` string.

https://github.com/yiisoft/yii2/blob/399dbce0cadbd7631f726841ece3ecf6a830444d/framework/db/Schema.php#L51


I don't see an issue here.

samdark:
I think the change is correct cause we're listing stings there in docs, not constants.

samdark:
Reverted. @SilverFire is correct.

Issue #15417, opened 2017-12-26
validator - minFiles


alexander-ronin:
Yii2 validator has `maxFiles ` property. But there is no `minFiles` in documentation.
I guess it's strange and  `minFiles` should be added too, which will checks for minimum quantity of uploaded files.
Thanks for attention.

samdark:
What's the use case?

alexander-ronin:
In form users must to upload at least 2 images.

samdark:
OK. Not against adding it but since 2.0.14 will be the last feature-release and there are many things we want to get it, the only chance is if you'll prepare a pull request. Else it will be added into 2.1.

alexander-ronin:
Got it, will wait 2.1.
Thanks )

vladis84:
Hello. 
I will do it.

vladis84:
I think it's right to specify the interval.
Example 
```php
['min' => 2, 'max' => 4];
['min' => 2, 'max' => 0]; // 0 - unlimited
```

FeR-S:
Is this fixed in version 2.0.15.1?

samdark:
2.0.14

PR #15434, opened 2017-12-30
FileValidator added minFiles property

vladis84:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes/no
| Tests pass?   | yes
| Fixed issues  | #15417

### DIFF
--- a/framework/validators/FileValidator.php
+++ b/framework/validators/FileValidator.php
@@ -80,6 +80,16 @@ class FileValidator extends Validator
      * @see tooMany for the customized message when too many files are uploaded.
      */
     public $maxFiles = 1;
+
+    /**
+     * @var int the minimum file count the given attribute can hold.
+     * Defaults to 1, meaning single file upload. By defining a higher number,
+     * multiple uploads become possible. Setting it to `0` means there is no limit on
+     * the number of files that can be uploaded simultaneously.
+     *
+     * @see tooFew for the customized message when too few files are uploaded.
+     */
+    public $minFiles = 1;
     /**
      * @var string the error message used when a file is not uploaded correctly.
      */
@@ -120,6 +130,16 @@ class FileValidator extends Validator
      * - {limit}: the value of [[maxFiles]]
      */
     public $tooMany;
+
+    /**
+     * @var string the error message used if the count of multiple uploads less that minFiles.
+     * You may use the following tokens in the message:
+     *
+     * - {attribute}: the attribute name
+     * - {limit}: the value of [[minFiles]]
+     */
+    public $tooFew;
+
     /**
      * @var string the error message used when the uploaded file has an extension name
      * that is not listed in [[extensions]]. You may use the following tokens in the message:
@@ -156,6 +176,9 @@ class FileValidator extends Validator
         if ($this->tooMany === null) {
             $this->tooMany = Yii::t('yii', 'You can upload at most {limit, number} {limit, plural, one{file} other{files}}.');
         }
+        if ($this->tooFew === null) {
+            $this->tooFew = Yii::t('yii', 'You can upload at less than {limit, number} {limit, plural, one{file} other{files}}.');
+        }
         if ($this->wrongExtension === null) {
             $this->wrongExtension = Yii::t('yii', 'Only files with these extensions are allowed: {extensions}.');
         }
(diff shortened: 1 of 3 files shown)


framework/validators/FileValidator.php
```diff
@@ -80,6 +80,16 @@ class FileValidator extends Validator
...
+
+    /**
+     * @var int the minimum file count the given attribute can hold.
+     * Defaults to 1, meaning single file upload. By defining a higher number,
+     * multiple uploads become possible. Setting it to `0` means there is no limit on
+     * the number of files that can be uploaded simultaneously.
+     *
+     * @see tooFew for the customized message when too few files are uploaded.
+     */
+    public $minFiles = 1;
```
samdark: I don't think it should be 1 by default since it will break BC by disallowing not uploading a file.

framework/validators/FileValidator.php
```diff
@@ -80,6 +80,16 @@ class FileValidator extends Validator
...
      */
     public $maxFiles = 1;
+
+    /**
+     * @var int the minimum file count the given attribute can hold.
+     * Defaults to 1, meaning single file upload. By defining a higher number,
+     * multiple uploads become possible. Setting it to `0` means there is no limit on
+     * the number of files that can be uploaded simultaneously.
+     *
+     * @see tooFew for the customized message when too few files are uploaded.
```
samdark: `@since 2.0.14`

framework/validators/FileValidator.php
```diff
@@ -120,6 +130,16 @@ class FileValidator extends Validator
...
     public $tooMany;
+
+    /**
+     * @var string the error message used if the count of multiple uploads less that minFiles.
+     * You may use the following tokens in the message:
+     *
+     * - {attribute}: the attribute name
+     * - {limit}: the value of [[minFiles]]
+     */
+    public $tooFew;
```
samdark: `@since 2.0.14`

samdark (changes requested):

framework/validators/FileValidator.php
vladis84: Fix it

framework/validators/FileValidator.php
```diff
@@ -80,6 +80,15 @@ class FileValidator extends Validator
      * @see tooMany for the customized message when too many files are uploaded.
      */
     public $maxFiles = 1;
+
+    /**
+     * @var int the minimum file count the given attribute can hold.
+     * Defaults to 1, meaning single file upload. By defining a higher number,
```
samdark: Defaults to 0. Higher value means at least that number of files should be uploaded.

vladis84: I fixed it

framework/validators/FileValidator.php
```diff
@@ -156,6 +177,9 @@ public function init()
         if ($this->tooMany === null) {
             $this->tooMany = Yii::t('yii', 'You can upload at most {limit, number} {limit, plural, one{file} other{files}}.');
         }
+        if ($this->tooFew === null) {
+            $this->tooFew = Yii::t('yii', 'You can upload at less than {limit, number} {limit, plural, one{file} other{files}}.');
```
SilverFire: `You can upload at less than {limit, number} {limit, plural, one{file} other{files}}.` -> `You should upload at least {limit, number} {limit, plural, one{file} other{files}}.`

SilverFire (changes requested):
Good otherwise

framework/validators/FileValidator.php
vladis84: I fixed it

developeruz (approved):

framework/validators/FileValidator.php
```diff
@@ -185,30 +209,36 @@ public function init()
...
+            $filesCount = count($files);
+            if ($this->maxFiles && $filesCount > $this->maxFiles) {
                 $this->addError($model, $attribute, $this->tooMany, ['limit' => $this->maxFiles]);
-            } else {
-                foreach ($files as $file) {
-                    $result = $this->validateValue($file);
-                    if (!empty($result)) {
-                        $this->addError($model, $attribute, $result[0], $result[1]);
-                    }
+            }
```
developeruz: don't you need `return` here and on line 236?

vladis84: Validation errors are not overwritten.

samdark (approved):

samdark:
Merged. Thank you!

PR #15435, opened 2017-12-30
Fix code issues

vkhramtsov:
| Q             | A
| ------------- | ---
| Is bugfix?    | possible
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | fix some  scrutinizer warnings, remove obsolete require

### DIFF
--- a/framework/Yii.php
+++ b/framework/Yii.php
@@ -5,8 +5,6 @@
  * @license http://www.yiiframework.com/license/
  */
 
-require __DIR__ . '/BaseYii.php';
-
 /**
  * Yii is a helper class serving common framework functionalities.
  *
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1249,7 +1249,7 @@ class BaseHtml
         }
 
         if ($encode) {
-            for ($i = 0; $i < count($lines); $i++) {
+            for ($i = 0, $linesCount = count($lines); $i < $linesCount; $i++) {
                 $lines[$i] = Html::encode($lines[$i]);
             }
         }
--- a/framework/validators/ExistValidator.php
+++ b/framework/validators/ExistValidator.php
@@ -92,7 +92,7 @@ class ExistValidator extends Validator
     {
         $targetAttribute = $this->targetAttribute === null ? $attribute : $this->targetAttribute;
         $params = $this->prepareConditions($targetAttribute, $model, $attribute);
-        $conditions[] = $this->targetAttributeJunction == 'or' ? 'or' : 'and';
+        $conditions = [ $this->targetAttributeJunction == 'or' ? 'or' : 'and' ];
 
         if (!$this->allowArray) {
             foreach ($params as $key => $value) {
--- a/framework/validators/ExistValidator.php
+++ b/framework/validators/ExistValidator.php
@@ -92,7 +92,7 @@ class ExistValidator extends Validator
     {
         $targetAttribute = $this->targetAttribute === null ? $attribute : $this->targetAttribute;
         $params = $this->prepareConditions($targetAttribute, $model, $attribute);
-        $conditions = [ $this->targetAttributeJunction == 'or' ? 'or' : 'and' ];
+        $conditions = [$this->targetAttributeJunction == 'or' ? 'or' : 'and'];
 
         if (!$this->allowArray) {
             foreach ($params as $key => $value) {
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -121,7 +121,7 @@ class UniqueValidator extends Validator
         $targetClass = $this->getTargetClass($model);
         $targetAttribute = $this->targetAttribute === null ? $attribute : $this->targetAttribute;
         $rawConditions = $this->prepareConditions($targetAttribute, $model, $attribute);
-        $conditions[] = $this->targetAttributeJunction === 'or' ? 'or' : 'and';
+        $conditions = [ $this->targetAttributeJunction === 'or' ? 'or' : 'and' ] ;
 
         foreach ($rawConditions as $key => $value) {
             if (is_array($value)) {
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -121,7 +121,7 @@ class UniqueValidator extends Validator
         $targetClass = $this->getTargetClass($model);
         $targetAttribute = $this->targetAttribute === null ? $attribute : $this->targetAttribute;
         $rawConditions = $this->prepareConditions($targetAttribute, $model, $attribute);
-        $conditions = [ $this->targetAttributeJunction === 'or' ? 'or' : 'and' ] ;
+        $conditions = [$this->targetAttributeJunction === 'or' ? 'or' : 'and'];
 
         foreach ($rawConditions as $key => $value) {
             if (is_array($value)) {
--- a/framework/console/widgets/Table.php
+++ b/framework/console/widgets/Table.php
@@ -228,6 +228,7 @@ class Table extends Widget
 
         $buffer = '';
         $arrayPointer = [];
+        $finalChunk = [];
         for ($i = 0, ($max = $this->calculateRowHeight($row)) ?: $max = 1; $i < $max; $i++) {
             $buffer .= $spanLeft . ' ';
             foreach ($size as $index => $cellSize) {
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -1013,6 +1013,7 @@ class Formatter extends Component
             $isNegative = $interval->invert;
         }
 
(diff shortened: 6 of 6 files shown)


vkhramtsov:
Hello,

I have one additional question: which code checking service do you like more?

framework/Yii.php
```diff
@@ -5,8 +5,6 @@
  * @license http://www.yiiframework.com/license/
  */
 
-require __DIR__ . '/BaseYii.php';
```
samdark: Why?

framework/validators/ExistValidator.php
```diff
@@ -92,7 +92,7 @@ public function validateAttribute($model, $attribute)
     {
         $targetAttribute = $this->targetAttribute === null ? $attribute : $this->targetAttribute;
         $params = $this->prepareConditions($targetAttribute, $model, $attribute);
-        $conditions[] = $this->targetAttributeJunction == 'or' ? 'or' : 'and';
+        $conditions = [ $this->targetAttributeJunction == 'or' ? 'or' : 'and' ];
```
samdark: Why?

framework/validators/UniqueValidator.php
```diff
@@ -121,7 +121,7 @@ public function validateAttribute($model, $attribute)
         $targetClass = $this->getTargetClass($model);
         $targetAttribute = $this->targetAttribute === null ? $attribute : $this->targetAttribute;
         $rawConditions = $this->prepareConditions($targetAttribute, $model, $attribute);
-        $conditions[] = $this->targetAttributeJunction === 'or' ? 'or' : 'and';
+        $conditions = [ $this->targetAttributeJunction === 'or' ? 'or' : 'and' ] ;
```
samdark: Why?

framework/helpers/BaseHtml.php
```diff
@@ -1249,7 +1249,7 @@ private static function collectErrors($models, $encode, $showAllErrors)
         }
 
         if ($encode) {
-            for ($i = 0; $i < count($lines); $i++) {
+            for ($i = 0, $linesCount = count($lines); $i < $linesCount; $i++) {
```
samdark: This one is OK.

samdark (changes requested):

framework/Yii.php
vkhramtsov: It looks like we include autoload in [web application](https://github.com/yiisoft/yii2-app-basic/blob/master/web/index.php) and [console application](https://github.com/yiisoft/yii2/blob/master/framework/yii) before including this file. I think that `require` here not required.

framework/validators/ExistValidator.php
vkhramtsov: Just want to fix [scrunitizer issue](https://scrutinizer-ci.com/g/yiisoft/yii2/issues/master/files/framework/validators/ExistValidator.php?selectedLabels%5B0%5D=3&selectedLabels%5B1%5D=10&orderField=path&order=asc&honorSelectedPaths=0)

framework/Yii.php
samdark: Yes but since `Yii` and `BaseYii` files aren't in any namespace, autoload won't handle them. It's done in 2.1 already and I don't think it's a good idea to touch it in 2.0.

framework/validators/UniqueValidator.php
vkhramtsov: Just want to fix [scrunitizer issue](https://scrutinizer-ci.com/g/yiisoft/yii2/issues/master/files/framework/validators/UniqueValidator.php?selectedLabels%5B0%5D=3&selectedLabels%5B1%5D=10&orderField=path&order=asc&honorSelectedPaths=0)

samdark:
> I have one additional question: which code checking service do you like more?

What do you mean?

framework/validators/ExistValidator.php
samdark: Ah. Right. Looks good. One thing is that `[` and `]` should not have extra spaces.

vkhramtsov:
> What do you mean?
I want to start work with this framework. From my point of view, best way to start with something -- fix some issues, write some new code and so on. I want to start from short and easy tasks, like fixing some minor code issues. Good source of tasks like this is code quality tools like scrunitizer or codeclimate. What do you think, which system I can use as source of issues?
Thank you

framework/validators/UniqueValidator.php
samdark: OK.

samdark:
Both. I'm using Codeclimate more.

framework/Yii.php
samdark: Please revert this one.

samdark (changes requested):
Looks good except removal of BaseYii require. Please revert it and I'll merge a pull request.

framework/Yii.php
vkhramtsov: Sorry, but I see that `BaseYii` is in `yii` namespace (https://github.com/yiisoft/yii2/blob/master/framework/BaseYii.php). Looks like I missed something. Would you be so kind to explain this  little bit more?

samdark: You're actually right :) Just tried it. Thanks for being persistent.

framework/validators/ExistValidator.php
vkhramtsov: Extra spaces removed

framework/validators/UniqueValidator.php
vkhramtsov: Extra spaces removed

vkhramtsov:
Extra spaces removed. `require` not reverted.
Should I squash these commits into one before merging?

framework/Yii.php
rob006: This is incorrect IMO. Autoloader is registered bellow, so you can't rely on it at this point - it works only because you have composer's autoloader which takes care about loading `BaseYii` class. So now you need to another autoloader to init Yii autoloader...

framework/validators/UniqueValidator.php
```diff
@@ -121,7 +121,7 @@ public function validateAttribute($model, $attribute)
         $targetClass = $this->getTargetClass($model);
         $targetAttribute = $this->targetAttribute === null ? $attribute : $this->targetAttribute;
         $rawConditions = $this->prepareConditions($targetAttribute, $model, $attribute);
-        $conditions[] = $this->targetAttributeJunction === 'or' ? 'or' : 'and';
+        $conditions = [$this->targetAttributeJunction === 'or' ? 'or' : 'and'] ;
```
rob006: Unnecessary space before `;`.

framework/Yii.php
vkhramtsov: @rob006 , I see definition of `psr-4` autoloader for `yii` namespace in composer.json [here](https://github.com/yiisoft/yii2/blob/master/framework/composer.json#L79) and [here](https://github.com/yiisoft/yii2/blob/master/composer.json#L99). About another autoloader to init `Yii` autoloader -- from my point of view it is ok, because we should use as less code as possible to create functionality, which we need 😄 .

framework/validators/UniqueValidator.php
vkhramtsov: Fixed

framework/Yii.php
samdark: Oh, right right. Yii 2.0 can work w/o Composer and some developers actually use that... should not be removed in 2.0 because of that...

samdark:
> Should I squash these commits into one before merging?

It isn't necessary.

framework/Yii.php
vkhramtsov: @samdark I see that application include autoloader in framework [console script](https://github.com/yiisoft/yii2/blob/master/framework/yii#L13) and in [web frontend script](https://github.com/yiisoft/yii2-app-basic/blob/master/web/index.php#L7). Are you speaking about users, who implement their own frontend and console scripts? In this case, of course, I must return  `require` back.

And one additional question: which branch are you have the latest code? I think that it is better to fix anything in latest code instead of porting fixes from old branches. For example in branch 2.1 `require` in [`Yii.php`](https://github.com/yiisoft/yii2/blob/2.1/framework/Yii.php#L8) and `require` for [`Yii.php`](https://github.com/yiisoft/yii2/blob/2.1/framework/yii#L27) should be removed, because `Yii.php` loaded via [composer autoload](https://github.com/yiisoft/yii2/blob/2.1/framework/composer.json#L82).

Thank you

samdark: There are two up to date branches: master (2.0) and 2.1. See http://www.yiiframework.com/news/156/yii-2-1-and-yii-2-0-feature-freeze/

samdark: Composer is included in zip distribution so I think that's OK to keep removal of these lines.

samdark (approved):

samdark:
Merged. Thank you!

Issue #14355, opened 2017-06-27
Can't send empty array argument in console command


dangarhow:
### What steps will reproduce the problem?
In a console command, when declaring an action, you can set an array as an expected value:

```php
public class FooController extends yii\console\Console
{
public function actionFoo(array $bar)
```

But there is no way to pass an empty array:

yii foo/foo '' => $bar=['']
yii foo/foo null => $bar=['null']
yii foo/foo [] => $bar=['[]']
yii foo/foo 0 => $bar=[0]

This is because in the bindParams, it always does a preg_split:

```php
public function bindActionParams($action, $params)
    {
        if ($action instanceof InlineAction) {
            $method = new \ReflectionMethod($this, $action->actionMethod);
        } else {
            $method = new \ReflectionMethod($action, 'run');
        }

        $args = array_values($params);

        $missing = [];
        foreach ($method->getParameters() as $i => $param) {
            if ($param->isArray() && isset($args[$i])) {
                $args[$i] = preg_split('/\s*,\s*/', $args[$i]);
            }
```

### What is the expected result?
It makes sense that if you pass an empty string or [] it should interpret it as an empty array

### What do you get instead?
The argument is always an array with one element as described above

### Additional info
It would be a rather small change in the bind params but I it could have a big impact. If the change is approved I would do the pull request

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version      | 7.1
| Operating system | Ubuntu 16

jlorente:
What about a default empty array value?

```
public class FooController extends yii\console\Console {
    public function actionFoo(array $bar = []) {
    }
}
```

dangarhow:
@jlorente that only works if its one argument. Lets say you have two optional arguments, both arrays, if you want to send an empty array on the first and pass a value for the second, you can't:

public function actionFoo(array $bar = [], array $bar2 = [])

jlorente:
Ok, I undestand your problem.

PR #15445, opened 2018-01-02
Added ability to pass an empty array as a parameter in console command

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14355


samdark (approved):

rob006:
Does it work on real console call? When you will get a empty string as a param on running some command from console?

developeruz:
@rob006, it won't work with spaces as an empty string, but it works with ``''``. I've checked on real console call.

rob006 (approved):

samdark:
Thank you!

PR #15448, opened 2018-01-03
"ezyang/htmlpurifier" package has been made optional

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #8452 (patrially)

"html" data format is not mandatory to be used inside every application - it is necessary only in case web user is allowed to enter an HTML code, which should be shown as HTML code.
It is much more common to simply secure any web user input passing it to `htmlspecialchars()`, e.g. `Html::encode()` instead.

"ezyang/htmlpurifier" package has been made optional.
`HtmlPurifier` unlikely to be moved into a separated package since it produces tight coupling with `Formatter::asHtml()`, which can not be simply removed or dropped as developer may extend `Formatter` class providing his own implementation.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -39,6 +39,7 @@ Yii Framework 2 Change Log
 - Chg #14178: Removed HHVM-specific code (samdark)
 - Enh #14671: use `random_int()` instead of `mt_rand()` to generate cryptographically secure pseudo-random integers (yyxx9988)
 - Chg #14761: Removed Yii autoloader in favor of Composer's PSR-4 implementation (samdark)
+- Chg: Package "ezyang/htmlpurifier" has been made optional and does installed via Composer by default (klimov-paul)
 
 2.0.14 under development
 ------------------------
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -39,7 +39,7 @@ Yii Framework 2 Change Log
 - Chg #14178: Removed HHVM-specific code (samdark)
 - Enh #14671: use `random_int()` instead of `mt_rand()` to generate cryptographically secure pseudo-random integers (yyxx9988)
 - Chg #14761: Removed Yii autoloader in favor of Composer's PSR-4 implementation (samdark)
-- Chg: Package "ezyang/htmlpurifier" has been made optional and is not installed by default (klimov-paul)
+- Chg #15448: Package "ezyang/htmlpurifier" has been made optional and is not installed by default (klimov-paul)
 
 2.0.14 under development
 ------------------------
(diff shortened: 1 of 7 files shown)


framework/CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ Yii Framework 2 Change Log
 - Chg #14178: Removed HHVM-specific code (samdark)
 - Enh #14671: use `random_int()` instead of `mt_rand()` to generate cryptographically secure pseudo-random integers (yyxx9988)
 - Chg #14761: Removed Yii autoloader in favor of Composer's PSR-4 implementation (samdark)
+- Chg: Package "ezyang/htmlpurifier" has been made optional and does installed via Composer by default (klimov-paul)
```
sergeymakinen:
```diff
- and does installed via Composer by default
+ and it's not installed by default
```

framework/UPGRADE.md
```diff
@@ -154,6 +154,9 @@ Upgrade from Yii 2.0.x
   `yii\validators\FileValidator::buildMimeTypeRegexp()` have been made `public`. Make sure you use correct
   access level specification in case you override these methods.
 * Default script position for the `yii\web\View::registerJs()` changed to `View::POS_END`.
+* Package "ezyang/htmlpurifier" has been made optional and does installed via Composer by default. If you need
```
sergeymakinen: Same as in the changelog.

framework/composer.json
```diff
@@ -71,12 +71,14 @@
...
         "psr/simple-cache": "~1.0.0",
         "psr/http-message": "~1.0.0",
-        "ezyang/htmlpurifier": "~4.6",
         "cebe/markdown": "~1.0.0 | ~1.1.0",
         "bower-asset/jquery": "3.2.*@stable | 3.1.*@stable | 2.2.*@stable | 2.1.*@stable | 1.11.*@stable | 1.12.*@stable",
         "bower-asset/punycode": "1.3.*",
         "bower-asset/yii2-pjax": "~2.0.1"
     },
+    "suggest": {
+        "ezyang/htmlpurifier": "required at `yii\\helpers\\HtmlPurifier` for 'html' data format support (e.g. `yii\\i18n\\Formatter:asHtml()`)"
```
sergeymakinen: Does Symfony's Console under Composer support/highlight grave accents? If it doesn't, there's no need to use it in the description.

sergeymakinen (changes requested):
LGTM, but some strings have to be corrected.

sergeymakinen (approved):

rob006:
`BaseStringHelper` also uses `HTMLPurifier_*` classes. 

Also note that using `suggest` section gives not real constraints for this dependency. User can install `htmlpurifier` with wrong version and then only runtime bugs will let him know about it.

framework/CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ Yii Framework 2 Change Log
 - Chg #14178: Removed HHVM-specific code (samdark)
 - Enh #14671: use `random_int()` instead of `mt_rand()` to generate cryptographically secure pseudo-random integers (yyxx9988)
 - Chg #14761: Removed Yii autoloader in favor of Composer's PSR-4 implementation (samdark)
+- Chg: Package "ezyang/htmlpurifier" has been made optional and is not installed by default (klimov-paul)
```
samdark: Need issue number.

samdark (approved):

SilverFire:
Will review this evening

SilverFire (approved):

PR #15449, opened 2018-01-03
Event names changed matching 'dot-separated string' format

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | 

Extracted from disscussion at #14349
Event names have been changed to 'dot-separated string' format, which is recommended by PSR.
Such format facilitates usage of wildcards for event handler specification.


framework/base/Application.php
```diff
@@ -52,11 +52,11 @@
     /**
      * @event Event an event raised before the application starts to handle a request.
      */
-    const EVENT_BEFORE_REQUEST = 'beforeRequest';
+    const EVENT_BEFORE_REQUEST = 'request.process.before';
```
samdark: The constant doesn't match the value now. Shouldn't we adjust it as well?

samdark (changes requested):
The change looks logical but constant names do not match values anymore so I think constant names should be adjusted.

klimov-paul:
> constant names do not match values anymore so I think constant names should be adjusted

That is debatable.
I think using dot separated string as an event name should be treated in the same way as class namespace. E.g. the word, which stands to the left, should cover more general entity then the one to the right. Thus event `BaseActiveRecord::BEFORE_INSERT` becomes 'db.ar.insert.before'.
This applows wildcard matching to be easily applied from the left to the right. For example:

- 'db.*'
- 'db.ar.*'
- 'db.ar.insert.*'

However, such notation is not good for the spelling, as it puts words into unnatural order.

I would prefer to keep constant names as they are. Still it would be good to perform some kind of survey for it.

framework/base/Application.php
samdark: Why there's `process`? It is not in the constant name or class name...

framework/base/Application.php
```diff
@@ -52,11 +52,11 @@
...
     /**
      * @event Event an event raised after the application successfully handles a request (before the response is sent out).
      */
-    const EVENT_AFTER_REQUEST = 'afterRequest';
+    const EVENT_AFTER_REQUEST = 'request.process.after';
```
samdark: Why there's `process`? It is not in the constant name or class name...

framework/base/Controller.php
```diff
@@ -32,11 +32,11 @@ class Controller extends Component implements ViewContextInterface
      * @event ActionEvent an event raised right before executing a controller action.
      * You may set [[ActionEvent::isValid]] to be false to cancel the action execution.
      */
-    const EVENT_BEFORE_ACTION = 'beforeAction';
+    const EVENT_BEFORE_ACTION = 'action.run.before';
```
samdark: I don't think there's need for `run` part i.e. `action.before` is enough.

framework/base/Module.php
```diff
@@ -45,11 +45,11 @@ class Module extends ServiceLocator
      * @event ActionEvent an event raised before executing a controller action.
      * You may set [[ActionEvent::isValid]] to be `false` to cancel the action execution.
      */
-    const EVENT_BEFORE_ACTION = 'beforeAction';
+    const EVENT_BEFORE_ACTION = 'action.run.before';
```
samdark: I don't think there's need for `run` part i.e. `action.before` is enough.

framework/base/Module.php
```diff
@@ -45,11 +45,11 @@ class Module extends ServiceLocator
...
     /**
      * @event ActionEvent an event raised after executing a controller action.
      */
-    const EVENT_AFTER_ACTION = 'afterAction';
+    const EVENT_AFTER_ACTION = 'action.run.after';
```
samdark: I don't think there's need for `run` part i.e. `action.before` is enough.

framework/db/Connection.php
```diff
@@ -137,19 +137,19 @@ class Connection extends Component
     /**
      * @event Event an event that is triggered after a DB connection is established
      */
-    const EVENT_AFTER_OPEN = 'afterOpen';
+    const EVENT_AFTER_OPEN = 'db.open.after';
```
samdark: `db.connection.open`? `db.connection.after`? `db.connection.afteropen`?

samdark:
@klimov-paul reverse notation indeed makes sense for wildcard matching but there are inconsistencies. I think we need some clear rules about how to name events...

klimov-paul:
> I think we need some clear rules about how to name events

I have no objections. Which will these rules be?

samdark:
The following rules are clear:

1. Event name should separate its parts by dots.
2. Parts should go from general to to specific.

We can go more strict on naming i.e.:

```
<namespace>/<lowercase_class_name>/<event>
```

In this case event names will become:

```php
ActiveRecord::EVENT_BEFORE_INSERT = 'yii.db.activerecord.insert.before'
ActiveQuery::EVENT_INIT = 'yii.db.activerecord.activequery.init';
Connection::EVENT_AFTER_OPEN = 'yii.db.connection.afteropen';
```

samdark:
Instead of `lowercase_class_name` we can use just class name w/o lowercasing it.

klimov-paul:
I see no reason to put `yii.` into the event name. In general it is simply a garbage, which makes event name longer. Someone may say it allows to catch all core events using `yii.*`, but it is not exactly correct. For example `yii2-mongodb` can be considered to be not Yii core, however it triggers event `BaseActiveRecord::BEFORE_INSERT` in its own code.

Using class name as a part of event is also not very practical. For example `BaseActiveRecord::BEFORE_INSERT` should be `yii.db.baseactiverecord.insert.before` insetad of `yii.db.activerecord.insert.before` then.
Also keep in mind that names of the events `BEFORE_ACTION` for `Controller` and `Module` are the same, which is crucial for the action filters to function.

samdark:
Putting `yii` in may help to prevent possible collisions with application events.

I agree that strict rules do not work well in some cases but at least they could serve as a base to start from.

klimov-paul:
As for the rules: it seems crucial that each event has a unique name. It is important to PSR event paradigm, which assumes there is a single event dispatcher, which does not segregate events by their sender. This means `BaseActiveRecord::EVENT_INIT` and `Widget::EVENT_INIT` can not share the same name.

samdark:
Yes. Let's add these to https://github.com/yiisoft/yii2/blob/2.1/docs/internals/design-decisions.md

klimov-paul:
> Putting yii in may help to prevent possible collisions with application events.

That is debatable. Particular application may inject extra events in the regular core flow. In particular I may want to handle DB connection close event in my application, thus I declare my own Connection class, which triggers new event. In this case such event should be named `db.connect.after` or `yii.db.connect.after`. The second looks weird in the scope of custom application.

My opinion: the first word in event name should describe some general term of application workflow, e.g. 'request', 'db', 'view' and so on. It should not matter from which library (yii core or not) particular implementation comes.

framework/db/ActiveQuery.php
```diff
@@ -78,7 +78,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
     /**
      * @event Event an event that is triggered when the query is initialized via [[init()]].
      */
-    const EVENT_INIT = 'init';
+    const EVENT_INIT = 'db.ar.query.init';
```
samdark: It is specific to active query. Should it be reflected in event name?

framework/db/Connection.php
```diff
@@ -137,19 +137,19 @@ class Connection extends Component
     /**
      * @event Event an event that is triggered after a DB connection is established
      */
-    const EVENT_AFTER_OPEN = 'afterOpen';
+    const EVENT_AFTER_OPEN = 'db.connect.after';
```
samdark: Should it be `db.connection.afteropen`?

framework/widgets/ActiveForm.php
```diff
@@ -30,12 +30,12 @@ class ActiveForm extends Widget
      * @event ActiveFieldEvent an event raised right before rendering an ActiveField.
      * @since 2.1.0
      */
-    const EVENT_BEFORE_FIELD_RENDER = 'beforeFieldRender';
+    const EVENT_BEFORE_FIELD_RENDER = 'widget.form.field.render.before';
```
samdark: Should it be reflected that it's activeform?

framework/widgets/ActiveForm.php
```diff
@@ -30,12 +30,12 @@ class ActiveForm extends Widget
...
     /**
      * @event ActionEvent an event raised right after rendering an ActiveField.
      * @since 2.1.0
      */
-    const EVENT_AFTER_FIELD_RENDER = 'afterFieldRender';
+    const EVENT_AFTER_FIELD_RENDER = 'widget.form.field.render.after';
```
samdark: Should it be reflected that it's activeform?

samdark (changes requested):
Last pack of possible name improvements. The rest of event names are looking good.

framework/db/Connection.php
klimov-paul: That is hard to tell: at the first edition this was `db.connection.open.after` assuming that it is possible for `db.connection.open.before` to exist.
I would rather avoid having chain part consisting of two words like 'afteropen' - word are better to be separated.

framework/widgets/ActiveForm.php
klimov-paul: I would prefer usage of 'form' and 'field' instead of 'activeform' and 'activefield'.
`ActiveForm` is just the name of the widget we use, which reflects it is a "smart" form. But for the event having 'active' inside its name looks redundant. E.g. "not smart" form is unlikely to trigger event - otherwise it becomes "smart", isn't it?

samdark: Makes sense.

framework/db/Connection.php
samdark: `db.connection.open.after` is a better name.

framework/widgets/ActiveForm.php
klimov-paul: done

framework/db/Connection.php
klimov-paul: done

samdark (approved):

framework/widgets/ActiveForm.php
schmunk42: Should event names which are in classes, which are likely to be moved to separate repos after 2.0.14, contain a special naming convention?

samdark: Not really.

schmunk42 (approved):
LGTM

Issue #14903, opened 2017-10-03
Yii2 controller and dash


abolfazl-ziaratban:
this bug has occurred when i use a controller with dash.

```php
class SiteController extends yii\web\Controller{
    function actionIndex(){
        return $this->render('index');
    }
}
```

goto http://example.com/web/site-
or
goto http://example.com/web/site------

error
The view file does not exist: ....\views\site-\index.php
or
The view file does not exist: ....\views\site------\index.php

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12

samdark:
Verified. Controller is executed while it should not.

PR #15453, opened 2018-01-04
Route with extra dashes is executed controller while it should not

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14903

### DIFF
--- a/framework/base/Module.php
+++ b/framework/base/Module.php
@@ -626,14 +626,13 @@ class Module extends ServiceLocator
             $className = substr($id, $pos + 1);
         }
 
-        if (!preg_match('%^[a-z][a-z0-9\\-_]*$%', $className)) {
-            return null;
-        }
-        if ($prefix !== '' && !preg_match('%^[a-z0-9_/]+$%i', $prefix)) {
+        if ($this->isIncorrectClassNameOrPrefix($className, $prefix)) {
             return null;
         }
 
-        $className = str_replace(' ', '', ucwords(str_replace('-', ' ', $className))) . 'Controller';
+        $className = preg_replace_callback('%-([a-z0-9_])%i', function ($matches) {
+                return ucfirst($matches[1]);
+            }, ucfirst($className)) . 'Controller';
         $className = ltrim($this->controllerNamespace . '\\' . str_replace('/', '\\', $prefix) . $className, '\\');
         if (strpos($className, '-') !== false || !class_exists($className)) {
             return null;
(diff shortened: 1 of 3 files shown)


framework/base/Module.php
```diff
@@ -649,6 +648,23 @@ public function createControllerByID($id)
...
     }
 
+    /**
+     * @param $className
+     * @param $prefix
+     * @return bool
+     */
+    private function incorrectClassNameOrPrefix($className, $prefix)
+    {
+        if (!preg_match('%^[a-z][a-z0-9\\-_]*$%', $className)) {
```
vladis84: ClassName can be in format CamelCase?

framework/base/Module.php
```diff
@@ -649,6 +648,23 @@ public function createControllerByID($id)
         return null;
     }
 
+    /**
+     * @param $className
```
vladis84: @param $className string

tests/framework/base/ModuleTest.php
```diff
@@ -120,6 +120,33 @@ public function testServiceLocatorTraversal()
         $this->assertTrue($child->has('test'));
         $this->assertFalse($parent->has('test'));
     }
+
+    public function testCreateControllerByID()
```
vladis84: I recommend in one test one condition.

framework/base/Module.php
developeruz: No, it is part of route and Yii2 splits CamelCase to dashed names.

framework/base/Module.php
```diff
@@ -626,14 +626,13 @@ public function createControllerByID($id)
             $className = substr($id, $pos + 1);
         }
 
-        if (!preg_match('%^[a-z][a-z0-9\\-_]*$%', $className)) {
-            return null;
-        }
-        if ($prefix !== '' && !preg_match('%^[a-z0-9_/]+$%i', $prefix)) {
+        if($this->incorrectClassNameOrPrefix($className, $prefix)) {
```
samdark: `isIncorrectClassNameOrPrefix` is a better name. Also please fix formatting: `if<space>(`

framework/base/Module.php
```diff
@@ -649,6 +648,23 @@ public function createControllerByID($id)
         return null;
     }
 
+    /**
+     * @param $className
+     * @param $prefix
```
samdark: string

framework/base/Module.php
samdark: phpdoc is needed.

framework/base/Module.php
samdark: phpdoc is needed.

samdark (changes requested):
Overall looks good. Please fix minor things.

framework/base/Module.php
developeruz: Done

samdark:
Merged. Thank you!

Issue #15462, opened 2018-01-06
accessChecker configuration error


dicrtarasov:
### What steps will reproduce the problem?
1) configure user component like this:
```php
'components' => [
        'user' => [
            'accessChecker' => 'app\helpers\AccessChecker',
            // ...  other params
        ],
```
### What is the expected result?
accessChecker object must instantinate in User

### What do you get instead?
`PHP Fatal Error – yii\base\ErrorException
Call to a member function checkAccess() on string
in /home/clevertoy/app/vendor/yiisoft/yii2/web/User.php line 712:`

```php
709:    if (($accessChecker = $this->getAccessChecker()) === null) {
710:        return false;
711:    }
712:    $access = $accessChecker->checkAccess($this->getId(), $permissionName, $params);
```
$this->getAccessChecker() must return object instead of string

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13.1
| PHP version      |  7.0.24
| Operating system | Debian 8

developeruz:
Confirmed. It works if one setup `accessChecker` from the code like 
```php
 $user->accessChecker = new MyAccessChecker();
```
but for the config it is impossible to give an object.

PR #15475, opened 2018-01-08
Fixed `accessChecker` configuration error

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15462

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -165,6 +165,9 @@ class User extends Component
         if ($this->enableAutoLogin && !isset($this->identityCookie['name'])) {
             throw new InvalidConfigException('User::identityCookie must contain the "name" element.');
         }
+        if (!empty($this->accessChecker) && is_string($this->accessChecker)) {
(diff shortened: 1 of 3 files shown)


framework/web/User.php
```diff
@@ -165,6 +165,9 @@ public function init()
         if ($this->enableAutoLogin && !isset($this->identityCookie['name'])) {
             throw new InvalidConfigException('User::identityCookie must contain the "name" element.');
         }
+        if (!empty($this->accessChecker) && is_string($this->accessChecker)) {
+            $this->accessChecker = new $this->accessChecker;
```
samdark: I think it's better to use `Yii::createObject()` here.

developeruz: Changed

samdark:
Merged. Thank you!

PR #15480, opened 2018-01-10
Update InvalidParamException.php

tvitcom:
prevent raise fatal error on case throw new \yii\base\InvalidParamException

| Q             | A
| ------------- | ---
| Is bugfix?    | yes


SilverFire:
Not reproduced.

Issue #15456, opened 2018-01-04
Entities double encoded in labels of ActiveForm fields


bverdeyen:
### What steps will reproduce the problem?

* Create a model with a field
* In the method attributeLabels, set the label of this field to "Carte d&#039;identité" (**obviously this field was in fact translated in database in the message table**)
```php
namespace common\models;
use yii\base\Model;
class FakeModel extends Model {

	var $field;

	public function attributeLabels()
    {
        return [
            'field' => 'Carte d&#039;identité',         
        ];
    }
}
```

* Try to display an input for this field

```php
use yii\widgets\ActiveForm;
use yii\widgets\DetailView;

$form = ActiveForm::begin();
echo $form->field(new common\models\FakeModel, 'field');
ActiveForm::end();

// also displaying a detail view to compare the behaviours
echo DetailView::widget(['model' => new common\models\FakeModel]);
```
### What is the expected result?

The label of the input should display "Carte d'identité", like for the DetailView widget also present in the example

### What do you get instead?

The label has the entity for the apostrophe re-encoded and thus not correctly displayed.

The DetailView widget has it correctly.

![capture d ecran 2018-01-04 a 14 39 46](https://user-images.githubusercontent.com/14311526/34565995-30aae020-f15d-11e7-84a0-55893cac1766.png)

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11.2
| PHP version   | 5.6.12
| Operating system | MacOs High Sierra Version 10.13.2

PR #15484, opened 2018-01-10
Modified BaseHtml to not double encode entities in label (Issue #15456)

bverdeyen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #15456

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,6 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.14 under development
 ------------------------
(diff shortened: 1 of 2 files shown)


rob006:
I think that #15456 should be fixed in the exact opposite way. `DetailView` should encode labels by default, same as `GridView`, to make this behavior consistent. Returning html-encoded values in `attributeLabels()` doesn't have much sense for me.

Issue #14971, opened 2017-10-15
Make MSSQL and Oracle tests run on Travis


klimov-paul:
See https://github.com/yiisoft/yii2-httpclient/issues/126

According to the specified criteria any functionality, which has no proof to be running via Travis build can not be declared as supported one.

Thus either unit tests for MSSQL and Oracle should be made running on Travis or related libraries should be removed along with the docs mentioning there is a support for MSSQL and Oracle.

rob006:
If you have spare time for unit test fix - open a PR for it.

klimov-paul:
Sorry, I do not have any spare time - you should handle this on your own.

samdark:
It was worked on and I agree that we should finish it. We're running [gitlab CI](https://gitlab.com/yiisoft/yii2/pipelines) and there are containers prepared for MSSQL and Oracle. We need to ensure these are executed on a regular basis.

samdark:
@rob006 this one is rather complicated. @schmunk42 has details.

schmunk42:
Sorry, I missed this conversation. Did anyone see a MSSQL test setup on Travis?

See also https://github.com/yiisoft/yii2/pull/15485#issuecomment-356760976 - MSSQL needs 3,5 GB of **free** RAM, on GitLab I had to create a custom runner with ~5 GB RAM to make them work.

samdark:
MSSQL part done. Thanks, @darkdef!

PR #15485, opened 2018-01-10
Fix messy .gitlab-ci.yml file

mikk150:
Gitlab CI definition file is kind of messy, I cleaned it up a bit with not using docker-compose

screenshot from 2018-01-10 20-30-50

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    |no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,8 +30,10 @@ stages:
     POSTGRES_USER: postgres
     POSTGRES_PASSWORD: postgres
     
-    
-    
+.caching-shared: &caching-shared
+  services:
+    - name: redis
+    - name: memcached   
 
 test:5.6:
   <<: *test-shared
@@ -39,13 +41,11 @@ test:5.6:
   script:
     - php vendor/bin/phpunit -v --exclude caching,db,data
 
-
 test:7.0:
   <<: *test-shared
   image: yiisoftware/yii2-php:7.0-apache
   script:
     - php vendor/bin/phpunit -v --exclude caching,db,data
-
 
 test:7.1:
   <<: *test-shared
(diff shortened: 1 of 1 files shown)


mikk150:
I may even be able to add mssql tests #14971

I tried, but have no idea how to make default DB - always got
```
Something wrong when preparing database: SQLSTATE[42000]: [unixODBC][Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Cannot open database "yii2test" requested by the login. The login failed.
```

schmunk42:
> Gitlab CI definition file is kind of messy

"messy" in which way, could you elaborate this a bit more?
While I like certain parts of this PR, ie. using YML templates, I think it very different from the current implementation, which has some very useful features...

> I cleaned it up a bit with not using docker-compose

Using `docker-compose` was a main idea behind developing this test setup, since you can reproduce it easy locally. 
Moreover you can pick any DB version just by using a variable in a trigger, which is really helpful for spotting incompatibilities with ie. newer DB versions.

> I may even be able to add mssql tests #14971

I doubt those will run on a standard (free) runner, since you need at least 3,5 GB of RAM for the MSSQL image - at least the last time I tried, the free runners were not providing that amount of memory.

> I tried, but have no idea how to make default DB - always got

I had a hard time figuring that out, since you need a separate image with `sqlcmd` to create the db, etc... https://github.com/yiisoft/yii2/blob/master/.gitlab-ci.yml#L98 

Overall I do not really see an improvement in this PR.

mikk150:
> Using docker-compose was a main idea behind developing this test setup, since you can reproduce it easily locally.

$ gitlab-ci-multi-runner exec docker test:mssql:7.0

> Moreover you can pick any DB version just by using a variable in a trigger, which is really helpful for spotting incompatibilities with ie. newer DB versions.

When you have issue with particular SQL version, shouldn't you add those specific tests with that specific SQL version to testing suite

> I doubt those will run on a standard (free) runner, since you need at least 3,5 GB of RAM for the MSSQL image - at least the last time I tried, the free runners were not providing that amount of memory.

Did run... Tho had [errors](https://gitlab.com/mikk150/yii2/-/jobs/47485049) (Not running out of memory tho)

schmunk42:
> $ gitlab-ci-multi-runner exec docker test:mssql:7.0

That actually sounds like a nice feature, but what do you need to execute tests locally? Did you install a runner binary locally?

> When you have issue with particular SQL version, shouldn't you add those specific tests with that specific SQL version to testing suite

There can be so many combinations of versions considering PHP, DB, Cache, Yii Framework.

Travis is really good for testing the above combinations (matrix) you described, GitLab is pretty strong when using triggers with custom variables and also for running partial sets of tests, just by pushing to a branch. Also parallelizing tests are done very well by GitLab.

I tried to use the strengths of GitLab with the current test-setup, not just doing the same like on Travis.

> Did run... Tho had errors (Not running out of memory tho)

The `services` part is also nice, since it looks pretty reliable for running on GitLab. On the other hand, how do you bring up a testing stack locally, keep it running for running single tests in when fixing issues etc.?

samdark:
Doesn't seem to be valid/active anymore. Closing.

PR #15491, opened 2018-01-12
translate plural format for MS

xicond:
I'm working on ms, th, zh, id localization
and found these ms plural translation is not correct by my team

translate plural format for MS
@samdark 

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | translate plural format for MS


samdark:
Thanks!

PR #15507, opened 2018-01-16
Fix #11389 Refactor Event Interface

### LINKED ISSUE #11389: Suggetion about event.
Reopen old issue #3243. Maybe for 2.1

Event as jquery like
### 1. Trigger event using `name` or `object` (break BC? yes/no).

Currently, to send event object, it must be as as second parameter of `trigger()`. But, event object itself contain name property. Better first parameter can be event name or event object.

``` php
// use
Yii::$app->trigger('myEvent');
// or use
$myEvent = new Event('myEvent');
Yii::$app->trigger($myEvent);
```
### 2. Use `return false;` or `$event->stopPropagation()` to stop handling (break BC? no).

Currently to stop event handling we use `$event->handled = true`. Prefer using `return false`. 

``` php
Yii::$app->on('myEvent', function($event){
    //...
    if ($condition){
        return false;
        // or $event->stopPropagation();
        // instead of $event->handled = true;
    }
});

Yii::$app->trigger('myEvent');
```
### 3. Add extra parameter when trigger event (break BC? no).

``` php
Yii::$app->on('myEvent', function($event, $param1, $param2){
    // do with $param1 and $param2
});

Yii::$app->trigger('myEvent',['param1', 'param2']);
```

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #11389 

Interface of the `Event` class refactored inspired by [Event Manager PSR](https://github.com/php-fig/fig-standards/blob/master/proposed/event-manager.md)


framework/base/Controller.php
```diff
@@ -270,8 +270,8 @@ public function createAction($id)
      */
     public function beforeAction($action)
     {
-        $event = new ActionEvent($action);
-        $this->trigger(self::EVENT_BEFORE_ACTION, $event);
+        $event = new ActionEvent($action, ['name' => self::EVENT_BEFORE_ACTION]);
```
samdark: What do you think about using classes for events same as we do for exceptions:

```php
class BeforeActionEvent extends Event
{
    const NAME = 'request.action.before';

    public function getName()
    {
        return self::NAME;
    }
}
```

It would look like:

```php
$event = new BeforeActionEvent($action);
```

samdark (approved):
Good direction overall.

framework/base/Controller.php
klimov-paul: Mostly such classes are redundant.
We already have specific event classes for some cases like `ActionEvent` or `ModelEvent`.
I think introduction of the separated class just because it provides default event name is a redundant overcomplication.

samdark: OK.

SilverFire (approved):

dmirogin:
need change since from 2.1 to 3.0

dmirogin (approved):
lgtm
but need change since from 2.1 to 3.0

samdark:
Merged. Thank you very much!

PR #15509, opened 2018-01-16
Fixed #14135 `yii\web\Request::getBodyParam()` crashes on object body

### LINKED ISSUE #14135: \yii\web\Request::getBodyParam crashes when \yii\web\JsonParser is configured to return objects
When the `JsonParser` is configured to return objects via `asArray = false` `Request::getBodyParam()` will crash as it expects to get an array from the `getBodyParams`. By returning an object the `JsonParser` also breaks its own interface. But returning an object could be quite handy sometimes, so I think that the `\yii\web\RequestParserInterface::parse` interface should be extended to incorporate this. Thoughts?

| Yii version      | 2.0.11.2

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14135

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -582,6 +582,15 @@ class Request extends \yii\base\Request
     {
         $params = $this->getBodyParams();
 
+        if (is_object($params)) {
+            // unable to use `ArrayHelper::getValue()` due to different dots in key logic and lack of exception handling
+            try {
+                return $params->{$name};
+            } catch (\Exception $e) {
+                return $defaultValue;
+            }
+        }
+
         return isset($params[$name]) ? $params[$name] : $defaultValue;
     }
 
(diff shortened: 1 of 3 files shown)


framework/web/Request.php
```diff
@@ -582,6 +582,15 @@ public function getBodyParam($name, $defaultValue = null)
     {
         $params = $this->getBodyParams();
 
+        if (is_object($params)) {
+            // unable to use `ArrayHelper::getValue()` due to different dots in key logic and lack of exception handling
+            try {
+                return $params->{$name};
+            } catch (\Exception $e) {
+                return $defaultValue;
```
samdark: Is that expected behavior?

klimov-paul: See https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseArrayHelper.php#L206-L209

I can not find any other way to handle unexisting object property here.

samdark: OK.

samdark (approved):

samdark:
Merged. Thank you!

PR #15516, opened 2018-01-18
improved dockerized tests

schmunk42:
- added ENV variable for using custom runtime folder
- create assets directory before using asset-manager

| Q             | A
| ------------- | ---
| Is bugfix?    | yes (when using host-volumes for debugging tests)
| New feature?  | yes (ENV var)
| Breaks BC?    | no
| Tests pass?   | yes(!)

When running and/or debugging tests from a host-volume with Docker, several tests fail due to disk-sync issues, therefore I added a variable to use a custom runtime folder, if not set it behaves like before.

Although the asset directory is available as empty folder from the repo, I also had issues with it in the above setup, when it was not create before the asset-manager was used.

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -23,5 +23,10 @@ require_once __DIR__ . '/../framework/Yii.php';
 
 Yii::setAlias('@yiiunit', __DIR__);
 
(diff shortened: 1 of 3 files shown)


samdark (approved):

samdark:
Merged. Thanks!

Issue #14973, opened 2017-10-15
Base phpunit classes should be extracted to yii2-phpunit package


rob006:
### What steps will reproduce the problem?

We have a compatibility layer for phpunit: https://github.com/yiisoft/yii2/blob/master/tests/compatibility.php

This code was copied between many (all?) of yii2 official extensions. But this code relies on assumption, that phpunit 4 does not have `PHPUnit\Framework\TestCase` class. This was changed in phpunit 4.8.35: https://github.com/sebastianbergmann/phpunit/commit/a55235c033c2dc18db631cc6b804844f6d37615a
So now some tests does not work on phpunit 4.8.35 or 4.8.36 and you need to fix it in every extension. Obviously this is a monkey job and nobody wants to do it. So we have some lazy/temporary workarounds:
* Lock phpunit to 4.8.34: https://github.com/yiisoft/yii2/commit/758f01934f180b4a4a47d36e1be103b1f979d76b#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L84
* Just disable tests on PHP 5.4, which requires phpunit 4: https://github.com/yiisoft/yii2-httpclient/commit/e61daa63eca37f21daca1f9f79888ffa3ba5836d

### What is the expected result?

DRY. There should be one package with phpunit-related classes. All potential improvements or fixes should be done there, not in each package separately.

samdark:
Agree. https://github.com/yiisoft/yii2-phpunit

samdark:
Not going to do it for 2.0.

PR #15517, opened 2018-01-18
updated phpunit and php-cs-fixer, removed php 5.4 from platform

### LINKED ISSUE #15512: Update phpunit dev-dependency

Run `phpunit`

No errors/warnings from PHPunit.

### What do you get instead?

```
PHP Deprecated:  The each() function is deprecated. This message will be suppressed on further calls in /yii2/vendor/phpunit/phpunit/src/Util/Getopt.php on line 38
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13
| PHP version      | 7.2
| Operating system | Debian Linux

See also https://github.com/sebastianbergmann/phpunit/issues/2887#issuecomment-348459163

Are there special reasons why this is pinned to `4.8.34`? The `platform` in `composer.json` also needs to be updated, I think. Is it `5.4`, because this is the current minimum requirement?

How do you make tests run on Travis due to this (by disabling) deprecated warnings?

schmunk42:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | ?
| Tests pass?   | that's what this PR is for

Fixes #15512

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,9 +81,9 @@
         "bower-asset/yii2-pjax": "~2.0.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "4.8.34",
+        "phpunit/phpunit": "6.5.5",
         "cebe/indent": "~1.0.2",
-        "friendsofphp/php-cs-fixer": "~2.2.3"
+        "friendsofphp/php-cs-fixer": "~2.10.0"
     },
     "repositories": [
         {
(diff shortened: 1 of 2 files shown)


rob006:
Old phpunit and php-cs-fixer is used for specific reason - we need to support PHP 5.4. We need to fix #14973 first, then we could use constraints like `"phpunit/phpunit": "^4.8.34 | ^6.5.5"` and run `composer update` on travis.

schmunk42:
I'll close this one for now.

PR #15527, opened 2018-01-19
WIP: Add JSON validator

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #15524

### DIFF
--- a/framework/helpers/BaseJson.php
+++ b/framework/helpers/BaseJson.php
@@ -110,28 +110,28 @@ class BaseJson
     }
 
     /**
-     * Validates the given data and returns a boolean indicating whether
-     * it is a valid JSON string or not.
+     * Validates the given data and returns an integer indicating whether
+     * any errors occurred while validation or not.
      *
      * @param mixed $data data to validate.
-     * @return boolean `true` if given data is a valid JSON string, `false` otherwise.
+     * @return int one of the constants returned from the `json_last_error()`.
+     * `JSON_ERROR_NONE` is returned if provided data is a valid JSON string.
      *
+     * @see http://php.net/manual/en/function.json-last-error.php
      * @since 2.0.14
      */
     public static function validate($data)
     {
-        $isValid = false;
+        $error = JSON_ERROR_NONE;
 
-        if (is_string($data) === true) {
+        if (is_string($data) === false) {
+            $error = JSON_ERROR_SYNTAX;
+        } else {
             json_decode($data);
-
-            $lastError = json_last_error();
-            if ($lastError === JSON_ERROR_NONE) {
-                $isValid = true;
-            }
+            $error = json_last_error();
         }
 
-        return $isValid;
+        return $error;
     }
 
     /**
--- a/framework/helpers/BaseJson.php
+++ b/framework/helpers/BaseJson.php
@@ -110,28 +110,32 @@ class BaseJson
     }
 
     /**
-     * Validates the given data and returns an integer indicating whether
-     * any errors occurred while validation or not.
+     * Validates the given data and returns a boolean indicating whether
+     * it is a valid JSON string or not.
      *
      * @param mixed $data data to validate.
-     * @return int one of the constants returned from the `json_last_error()`.
-     * `JSON_ERROR_NONE` is returned if provided data is a valid JSON string.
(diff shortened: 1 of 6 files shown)


framework/helpers/BaseJson.php
```diff
@@ -109,6 +109,31 @@ public static function decode($json, $asArray = true)
...
+     * @return boolean `true` if given data is a valid JSON string, `false` otherwise.
+     *
+     * @since 2.0.14
+     */
+    public static function validate($data)
+    {
+        $isValid = false;
+
+        if (is_string($data) === true) {
+            json_decode($data);
```
SilverFire: I'd add the STFU operator just for the case.

SilverFire (approved):

framework/helpers/BaseJson.php
samdark: Nope. It doesn't raise any errors.

framework/helpers/BaseJson.php
```diff
@@ -109,6 +109,31 @@ public static function decode($json, $asArray = true)
...
+
+            $lastError = json_last_error();
+            if ($lastError === JSON_ERROR_NONE) {
```
samdark: This info is valuable so I'd think twice before returning just true/false and losing it.

samdark: http://www.beliefmedia.com/code/php-snippets/check-valid-json

klimov-paul:
I am afraid that simple syntax check is not enough for JSON validation.
It should support matching particular internal structure as well. For example: there should be a way to determine whether some JSON can suite as a configuration for Composer.

samdark:
Yep. That as well: https://github.com/Seldaek/jsonlint

samdark:
And https://github.com/thephpleague/json-guard

samdark:
Composer itself uses https://github.com/Seldaek/json-schema

framework/helpers/BaseJson.php
Kolyunya: @samdark I refactored the method so that now it returns actual validation errors.

Kolyunya:
@klimov-paul @alekson described a use-case when he only needs to know if provided data is a valid JSON or not. We may add schema validation later on. These are two independent features.

framework/helpers/BaseJson.php
Kolyunya: @samdark another option is to return the error via the output parameter. In fact I like this option more.

rob006:
Why did you add this to helper instead of create separate validator class?

Kolyunya:
I've added this method to a helper for a convenience of use. In some scenarios you want to validate a value w/o instantiating a validator class. I think this method doesn't lower the helper cohesion.

I've also added a separate validator which makes use of a new helper method. But it's more like a proof of concept since I feel like the whole PR is going to be rejected.

samdark:
I'm leaning towards rejecting it... While it may be useful, validating just JSON syntax doesn't sounds like a common requirement while validating schema doesn't sound like we should implement it (there are many good libraries available).

Kolyunya:
@samdark what do you think about implementing a JSON schema validator based on existing solutions? This way we will provide a convenient tool out of box which works together with Yii forms w/o reimplementing complex logic.

samdark:
What kind of Yii bindings do you propose? Validator?

Kolyunya:
@samdark yes, I meant a validator.

samdark:
I don't think we want to add another dependency to the core. btw., it's already available as extension: https://github.com/dstotijn/yii2-json-schema-validator


@Kolyunya thanks for contribution but it won't be accepted. Of course, that's me speaking. If other core team members disagree, we'll reopen it.

alekson:
Жаль, хотелось простой валидатор на уровне numeric, string и т.д. - json, но который проверит что строка без ошибок конвертится в json.  Актуально для небольшого конфига, где не нужна сложная json schema.

Сейчас пользуюсь - https://github.com/romi45/yii2-json-validator, для  такой простой задачи не оч хотелось тянуть зависимость.

PR #15528, opened 2018-01-20
incorrect float to string conversion in yii\helper\BaseStringHelper

kminooie:
on my system,  (string) $number in line
   return str_replace(',', '.', (string) $number);
is converted to scientific notation, so it returns this 1.51643E+9 instead of 1516425820.449680. 

By the way this result is being used in \yii\log\Target::getTime(), this is obviously not the expected output of this code, yet I can not find anything in my php configuration/environment that would explain this. Although, it seems that I am only encountering this when I am using cli. 

I am seeing this on fedora ( latest php) and debian 9, running:

$ php --version
PHP 7.1.12-2+0~20171207160618.12+stretch~1.gbp5c91f3 (cli) (built: Dec  7 2017 16:06:20) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.12-2+0~20171207160618.12+stretch~1.gbp5c91f3, Copyright (c) 1999-2017, by Zend Technologies

it seems number_format() is a more foolproof and setting independent way of accomplishing this conversion

| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

### DIFF
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -361,7 +361,7 @@ class BaseStringHelper
      */
     public static function floatToString($number)
     {
-        return sprintf("%F", $number);
+        return sprintf('%F', $number);
     }
 
     /**
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -361,9 +361,7 @@ class BaseStringHelper
      */
     public static function floatToString($number)
     {
-        // . and , are the only decimal separators known in ICU data,
-        // so its safe to call str_replace here
-        return str_replace(',', '.', (string) $number);
+        return number_format($number, ini_get('precision'), '.', '');
     }
 
     /**
--- a/framework/helpers/BaseStringHelper.php
+++ b/framework/helpers/BaseStringHelper.php
@@ -361,7 +361,7 @@ class BaseStringHelper
      */
     public static function floatToString($number)
     {
-        return number_format($number, ini_get('precision'), '.', '');
+        return sprintf("%F", $number);
     }
 
     /**


rob006:
`sprintf()` may be a better option than `number_format()`.

kminooie:
@rob006 I agree. %F ( capital f ) is "non-local aware" and in a little test that I just did, on my system, about 30% faster.

samdark:
Seems to fail tests.

SilverFire:
@kminooie would you like to continue work on this PR?

s1lver:
I added
```php 
date_default_timezone_set('UTC');
```
for test testFormatMessage().
It`s ok.

default

rob006:
We could leave `floatToString()` unchanged and use `sprintf()` in `Target::getTime()` directly - it should fix this issue and simplify the code a bit (and generate more predictable results IMO).

s1lver:
I agree. It's not about `floatToString()`, this method performs its function well.

cebe:
sorry, I thought this would be solved by #16876, but it seems different.

s1lver:
Something I don’t like to handle the `timestamp` this way, and I don’t see much sense in using` sprintf('%F', $parts[1])`.

```php
var_dump($timestamp);
```
float(1508160390.6083)

```php
var_dump(sprintf('%F', $parts[1]));
```
string(11) "6083.000000"

```php
var_dump(round(sprintf('%F', $parts[1])));
```
float(6083)

Therefore, I suggest just doing so. since microseconds are not processed, we simply attach them to the received date (without processing).

```php
$date = new \DateTime();
$date->setTimestamp($parts[0]);
dateTime = $date->format('Y-m-d H:i:s');

var_dump($dateTime);
```

string(19) "2017-10-16 13:26:30"


```php
$resultDate = $dateTime.'.'.$parts[1];
var_dump($resultDate);
```

string(24) "2017-10-16 13:26:30.6083"

So we will provide a predictable result.

s1lver:
On the other hand, you can do nothing at all. Since the test passes
```php
$timestampScientific  = 1.51643E+9;
$expectedWithoutMicro = '2018-01-20 06:33:20 [info][application] message';
$formatted = $target->formatMessage([$text, $level, $category, $timestampScientific]);
$this->assertSame($expectedWithoutMicro, $formatted);
```

rob006:
@kminooie Can you take a look at #16883?

PR #15532, opened 2018-01-21
Fixes #14806: GridView option $placeFooterAfterBody

### LINKED ISSUE #14806: GridView table tfoot should come after tbody
The current order is thead, tfoot, tbody (as it was used to be done) in the GridView table element, but the new HTML 5.1 recommendation changes this: https://www.w3.org/TR/html51/tabular-data.html#tabular-data: " [...] followed optionally by a thead element, followed by either zero or more tbody elements or one or more tr elements, followed optionally by a tfoot element [...] ".
This change would also produce nicer mouse selection behaviour.

The fix would be very simple, one would have to swap two lines in yii2\grid\GridView.php, in the function renderItems().

Create a GridView with 'showFooter' => true and with some data.

In the DOM tree, the order of the table children should be: thead, tbody, tfoot. (Of course with colgroup/caption etc if set).

### What do you get instead?
The tfoot element comes before tbody. This also changes the order of selected elements (try selecting the table content with the mouse and copy-pasting it into Excel, the footer will come as the second row).

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.1.1
| Operating system | Windows 7

terehru:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14806

` <?= GridView::widget([ 

        'showFooter' => true,
        // Set new option, check <tfoot> position in DOM
        'placeFooterAfterBody' => true,
        'columns' => [
            'id',
            ],
        ],
    ]); ?>
`

### DIFF
--- a/framework/grid/GridView.php
+++ b/framework/grid/GridView.php
@@ -351,12 +351,15 @@ class GridView extends BaseListView
         $tableHeader = $this->showHeader ? $this->renderTableHeader() : false;
         $tableBody = $this->renderTableBody();
 
+        $tableFooter = false;
+        $tableFooterAfterBody = false;
+        
         if ($this->showFooter) {
-	        $tableFooter = !$this->placeFooterAfterBody ? $this->renderTableFooter() : false;
-	        $tableFooterAfterBody = $this->placeFooterAfterBody ? $this->renderTableFooter() : false;
-        } else {
-            $tableFooter = false;
-	        $tableFooterAfterBody = false;
+            if ($this->placeFooterAfterBody) {
+                $tableFooterAfterBody = $this->renderTableFooter();
+            } else {
+                $tableFooter = $this->renderTableFooter();
+            }	        
         }
 
         $content = array_filter([
--- a/framework/grid/GridView.php
+++ b/framework/grid/GridView.php
@@ -349,8 +349,15 @@ class GridView extends BaseListView
         $columnGroup = $this->renderColumnGroup();
         $tableHeader = $this->showHeader ? $this->renderTableHeader() : false;
         $tableBody = $this->renderTableBody();
-	    $tableFooter = $this->showFooter ? (!$this->placeFooterAfterBody ? $this->renderTableFooter() : false) : false;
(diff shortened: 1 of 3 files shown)


framework/grid/GridView.php
```diff
@@ -127,6 +127,10 @@ class GridView extends BaseListView
      * @var bool whether to show the footer section of the grid table.
      */
     public $showFooter = false;
+	/**
+	 * @var bool whether to place footer after body in DOM if $showFooter is true
```
samdark: Please add `@since 2.0.14`.

framework/grid/GridView.php
```diff
@@ -345,13 +349,15 @@ public function renderItems()
         $columnGroup = $this->renderColumnGroup();
         $tableHeader = $this->showHeader ? $this->renderTableHeader() : false;
         $tableBody = $this->renderTableBody();
-        $tableFooter = $this->showFooter ? $this->renderTableFooter() : false;
+	    $tableFooter = $this->showFooter ? (!$this->placeFooterAfterBody ? $this->renderTableFooter() : false) : false;
```
samdark: It's better not to use short conditions in case there are many. Let's have it as proper `if`-s.

framework/grid/GridView.php
```diff
@@ -345,13 +349,15 @@ public function renderItems()
...
+	    $tableFooterAfterBody = $this->showFooter ? ($this->placeFooterAfterBody ? $this->renderTableFooter() : false) : false;
```
samdark: Same here.

samdark (changes requested):
See comments.

+ Need a line for CHANGELOG.

Otherwise the idea is fine and tests are OK.

framework/grid/GridView.php
terehru: added below, but changes for this line still pending

samdark (approved):

samdark:
Merged. Thank you!

PR #15541, opened 2018-01-22
Fixed `ActiveQuery::findWith` behavior that violates container configuration

prowwid:
PR for issue [#15540](https://github.com/yiisoft/yii2/issues/15540)

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15540

### DIFF
--- a/framework/db/ActiveQueryTrait.php
+++ b/framework/db/ActiveQueryTrait.php
@@ -162,7 +162,7 @@ trait ActiveQueryTrait
     {
         $primaryModel = reset($models);
         if (!$primaryModel instanceof ActiveRecordInterface) {
-            $primaryModel = new $this->modelClass();
+            $primaryModel = \Yii::createObject($this->modelClass);
         }
         $relations = $this->normalizeRelations($primaryModel, $with);
         /* @var $relation ActiveQuery */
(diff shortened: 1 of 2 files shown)


framework/db/ActiveQueryTrait.php
```diff
@@ -162,7 +162,7 @@ public function findWith($with, &$models)
     {
         $primaryModel = reset($models);
         if (!$primaryModel instanceof ActiveRecordInterface) {
-            $primaryModel = new $this->modelClass();
+            $primaryModel = \Yii::createObject($this->modelClass);
```
cebe: This should be solved using `StaticInstanceInterface` instead. In case `$this->modelClass` implements  `yii\base\StaticInstanceInterface` (can be checked with http://php.net/manual/en/function.is-subclass-of.php ), call static instance method. If it does not implement the interface, use the current code with `new`.

cebe (changes requested):

cebe:
See also #13779

klimov-paul:
Resolved by commit b9cbc56330c16f21c1fd8ea35a7017bb45b7ee49

Issue #15476, opened 2018-01-08
ActiveForm / ActiveField validation class on input element


nadar:
As discussed here: https://github.com/yiisoft/yii2-bootstrap/issues/228 it would make sense to implement this feature in the core ActiveForm/ActiveField as it could be a behavior for other CSS Frameworks as well (If bootstrap does, others will maybe follow).

### What steps will reproduce the problem?

New Bootstrap 4 Framework requires form validation on the Input element since beta 2. An example:

```
<div class="col-md-3 mb-3">
      <labe>Zip</label>
      <input type="text" class="form-control is-invalid" placeholder="Zip">
      <div class="invalid-feedback">Please provide a valid zip.</div>
</div>
```

### What is the expected result?

An option to configure which element should contain the validation class in ActiveForm / ActiveField.

### What do you get instead?

Its not possible yet.

### Additional info

https://github.com/yiisoft/yii2-bootstrap/issues/228

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13.1

PR #15565, opened 2018-01-26
Added ability to specify validation state element

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #15476

TODO:

- [x] Tests
- [x] JavaScript

### DIFF
--- a/framework/widgets/ActiveField.php
+++ b/framework/widgets/ActiveField.php
@@ -241,10 +241,10 @@ class ActiveField extends Component
         if ($this->model->isAttributeRequired($attribute)) {
             $class[] = $this->form->requiredCssClass;
         }
+        $options['class'] = implode(' ', $class);
         if ($this->form->validationStateOn === ActiveForm::VALIDATION_STATE_ON_CONTAINER) {
             $this->addErrorClassIfNeeded($options);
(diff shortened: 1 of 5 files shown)


framework/widgets/ActiveField.php
```diff
@@ -363,6 +363,11 @@ public function hint($content, $options = [])
     public function input($type, $options = [])
     {
         $options = array_merge($this->inputOptions, $options);
+
+        if ($this->form->validationStateOn === ActiveForm::VALIDATION_STATE_ON_INPUT && $this->model->hasErrors($this->attribute)) {
+            $options['class'][] = $this->form->errorCssClass;
+        }
```
nadar:
```
} else {
    $options['class'][] = $this->form->validCssClass;
}
```

This would provide the option to set `is-valid` and on error `is-invalid`. 

https://getbootstrap.com/docs/4.0/components/forms/#server-side

samdark: That's `successCssClass` and it's already there. There's no need to add this class initially.

PR #15571, opened 2018-01-29
[minor] SCA: minor code tweaks

kalessil:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | n/a

### DIFF
--- a/build/controllers/ReleaseController.php
+++ b/build/controllers/ReleaseController.php
@@ -109,7 +109,7 @@ class ReleaseController extends Controller
             foreach ($items as $item) {
                 $this->stdout("fetching tags for $item...");
                 if ($item === 'framework') {
-                    $this->gitFetchTags("{$this->basePath}");
+                    $this->gitFetchTags((string)$this->basePath);
                 } elseif (strncmp('app-', $item, 4) === 0) {
                     $this->gitFetchTags("{$this->basePath}/apps/" . substr($item, 4));
                 } else {
@@ -874,7 +874,7 @@ class ReleaseController extends Controller
                 $state = 'end';
             }
             // add continued lines to the last item to keep them together
-            if (!empty(${$state}) && trim($line !== '') && strpos($line, '- ') !== 0) {
+            if (!empty(${$state}) && trim($line) !== '' && strncmp($line, '- ', 2) !== 0) {
                 end(${$state});
                 ${$state}[key(${$state})] .= "\n" . $line;
             } else {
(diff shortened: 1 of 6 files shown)


build/controllers/ReleaseController.php
```diff
@@ -874,7 +874,7 @@ protected function splitChangelog($file, $version)
                 $state = 'end';
             }
             // add continued lines to the last item to keep them together
-            if (!empty(${$state}) && trim($line !== '') && strpos($line, '- ') !== 0) {
+            if (!empty(${$state}) && trim($line) !== '' && strncmp($line, '- ', 2) !== 0) {
```
samdark: This one looks serious.

tests/framework/db/CommandTest.php
```diff
@@ -307,7 +307,7 @@ public function testBatchInsert()
 
     public function testBatchInsertWithYield()
     {
-        if (version_compare(PHP_VERSION, '5.5', '<')) {
+        if (PHP_VERSION_ID < 50500) {
```
samdark: Why? It's less readable.

samdark (approved):

tests/framework/db/CommandTest.php
kalessil: Optimizer friendly, in case of the constant it can drop the branch when generation opcodes. One thing I'm not sure if it should be prefixed with '\' for this or not. if I find the link where this described, I'll poke you at Twitter.

build/controllers/ReleaseController.php
kalessil: A little bit, but due to weak typing this worked. I just noticed that we can drop `trim($line) !== ''` here at all, should I do it?

kalessil:
@samdark : will it be welcome a PR covering this case https://github.com/symfony/symfony/pull/25854/files (it's 10-20 functions like that) ?

samdark:
Yes.

kalessil:
Super, thank you. I'll create a separate one today (evening or so).

samdark:
Thank you for code improvements.

tests/framework/db/CommandTest.php
samdark: That's actually OK and makes sense but in this case it's up to code style (performance difference doesn't matter much).

kalessil:
👍

PR #15572, opened 2018-01-29
Update View.php

gouguoyin:
修复使用其他模板引擎时自定义模板后缀不是php时的BUG


samdark:
As far as I remember, that's intentional so we can fall-back to default PHP views provided by framework.

SilverFire:
Sorry, the PR will not be accepted

PR #15577, opened 2018-01-30
normalizePath for windows network shares that start with two backslashes

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes, I hope so
| Fixed issues  | #13034


samdark:
Merged. Thank you!

PR #15590, opened 2018-02-02
Fix BaseInflector::transliterate bug when a transliterator rule is incompatible with ICU version 

beroso:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | Fix Response::sendFile when a transliterator error happened

### DIFF
--- a/framework/helpers/BaseInflector.php
+++ b/framework/helpers/BaseInflector.php
@@ -504,7 +504,11 @@ class BaseInflector
                 $transliterator = static::$transliterator;
             }
 
-            if ($transliterator instanceof \Transliterator || $transliterator = \Transliterator::create($transliterator)) {
+            if (!$transliterator instanceof \Transliterator) {
+                $transliterator = \Transliterator::create($transliterator);
+            }
+
+            if ($transliterator !== null) {
                 return $transliterator->transliterate($string);
             }
         }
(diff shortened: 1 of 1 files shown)


berosoboy:
I came across with a bug when I tried to send file to browser, using yii\web\Response::sendFile method.
After deployed to production server, the action raised a ERR_EMPTY_RESPONSE on browser.
The error occurs when the sendFile method tried to transliterate the filename. My production server has ICU version 4.4.1, and is a shared host, so I can't upgrade it right now.

The proposed PR fix this use case, and I think it make the method more safe anyway.

framework/helpers/BaseInflector.php
```diff
@@ -504,7 +504,13 @@ public static function transliterate($string, $transliterator = null)
...
 
-            return transliterator_transliterate($transliterator, $string);
+            if (!$transliterator instanceof \Transliterator) {
+                $transliterator = \Transliterator::create($transliterator);
+            }
+
+            if ($transliterator !== null) {
+                return $transliterator->transliterate($string);
+            }
         }
```
cebe: this would mean to fall back to a method that may work completely different from the transliteration rule that is configured. We should probably throw an exception or at least warn about it. This is somewhat related to https://github.com/yiisoft/yii-core/issues/38#issuecomment-427301785 you basically want to disable intl if it does not work on your environment.

samdark:
Closing since it's better to solve by allowing to disable intl explicitly: https://github.com/yiisoft/i18n/issues/14

PR #15591, opened 2018-02-03
Log targets export exceptions

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14296

This fix adds condition check for log targets exporting and in case of failure exception is thrown.
Exception is caught by dispatcher and the warning can be passed to the next log target if there are any left.


samdark:
Merged. Thank you!

PR #15598, opened 2018-02-04
Fixes #14593: Added `yii\db\Command::EVENT_AFTER_EXECUTE` event that is triggered after command is executed

### LINKED ISSUE #14593: [ENH] Add new event  to 'yii\db\Command' class
I think it will be useful for different situations.

```php
class Command extends Component
{

const EVENT_AFTER_EXECUTE = 'afterExecute';

public function execute()
    {
        $sql = $this->getSql();
        list($profile, $rawSql) = $this->logQuery(__METHOD__);

        if ($sql == '') {
            return 0;
        }

        $this->prepare(false);

        try {
            $profile and Yii::beginProfile($rawSql, __METHOD__);

            $this->pdoStatement->execute();
            $n = $this->pdoStatement->rowCount();

            $profile and Yii::endProfile($rawSql, __METHOD__);

	    $this->trigger(self::EVENT_AFTER_EXECUTE, new AfterExecuteEvent([
		'command' => $this,
	    ]));

            $this->refreshTableSchema();

            return $n;
        } catch (\Exception $e) {
            $profile and Yii::endProfile($rawSql, __METHOD__);
            throw $this->db->getSchema()->convertException($e, $rawSql ?: $this->getRawSql());
        }
    }

}
```

As for me I use this event for automatic generation raw .sql files with migrations.

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14593

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.14 under development
 ------------------------
 
+- Enh #14593: Added `yii\db\Command::EVENT_AFTER_EXECUTE` event that is triggered after command is executed (ARHImed82, samdark)
 - Enh #15047: `yii\db\Query::select()` and `yii\db\Query::addSelect()` now check for duplicate column names (wapmorgan)
 - Enh #14643: Added `yii\web\ErrorAction::$layout` property to conveniently set layout from error action config (swods, cebe, samdark)
 - Enh #13465: Added `yii\helpers\FileHelper::findDirectory()` method (ArsSirek, developeruz)
--- a/framework/db/AfterExecuteEvent.php
+++ b/framework/db/AfterExecuteEvent.php
@@ -1,0 +1,24 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\db;
+
+use yii\base\Event;
+
+/**
+ * AfterExecuteEvent represents the information available in [[Command::EVENT_AFTER_EXECUTE]].
+ *
+ * @author Alexander Makarov <sam@rmcreative.ru>
+ * @since 2.0.14
+ */
+class AfterExecuteEvent extends Event
+{
+    /**
+     * @var Command command that was executed
+     */
+    public $command;
+}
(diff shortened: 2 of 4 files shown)


klimov-paul:
This will degrade the performance without much cause.

samdark:
This is for things like https://github.com/yiisoft/yii2-shell/issues/14 or manual replication or alike tasks that are to intercept SQL queries made.

samdark:
Usually there are < 100 SQL requests per page. Triggering same amount of event should not hurt performance much.

samdark:
If you have a way to test it in one of your current projects and if it will display significant performance degradation, I'll revert it right away.

klimov-paul:
DAO was meant to be light-weight layer for the SQL execution, with the max performance.
If you wish sophisticated layer including events around SQL you should wrap DAO for that.
ActiveRecord already triggers the events around insert or update command.

Personally I do not like bringing extra functionality with performance degradation just to support rarely used feature for console shell.

klimov-paul:
If we start adding events per each data manipulation - we will end in performance hell.
For example what about having `Command::EVENT_BEFORE_EXECUTE`?
What about events for `Cache` compoent on getting data from the cache storage?
What about events on opening or closing session?
What about events on retrieving translations in `I18N`?

If start adding events at every place - there will be no end it them and that will cause the significant performance impact in the end.

klimov-paul:
Either way, there should be consistency about event introduction. Either events should be rare and stick to high-level sophisticated layers, or they should appear everywhere and stick to lower level components.

samdark:
https://github.com/yiisoft/yii2/commit/c7e55450dd2c92ad5150fedc90e26c93adaca1ba

GrayWolfy:
Hey, guys! What do I need to remark.
I wanted to instrument my yii2 app via OpenTelemetry. 
And that's how I found this conversation. I wanted to make simillar instrumentation library to this [Auto-Instrument For Laravel](https://github.com/opentelemetry-php/contrib-auto-laravel)
For example I wanted to catch requests to database, and as it seems to me there is no easy path to do it. Me personally had to decorate Command and Connection classes. Connection to make decorated Command. And Command to decorate internalExecute to trigger an event for i could catch it by my observer and create Jaeger spans. Is there is an easy way to do it without it in yii2?  Cause I didn't find it.

PR #15622, opened 2018-02-08
Fix #15621

### LINKED ISSUE #15621: yii\web\User::getIdentity() memoizes `null` if exception is thrown & caught

1. Write some code that will cause an exception to be thrown somewhere in the course of `yii\web\User::renewAuthStatus()`.
2. Call `yii\web\User::getIdentity()` within a try/catch, catching the exception and letting the app continue running.
3. Call `yii\web\User::getIdentity()` again, but not within a try/catch.

The exception would be thrown again; this time bubbling up to the browser.

### What do you get instead?

`null` is returned, as if no user is logged in.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13.1
| PHP version      | 7.0.26
| Operating system | Ubuntu 16.04.3 LTS

brandonkelly:
Let yii\web\User::$_identity stay set to false if an exception is thrown as a result of calling renewAuthStatus()

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15621

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -187,8 +187,16 @@ class User extends Component
     {
         if ($this->_identity === false) {
             if ($this->enableSession && $autoRenew) {
-                $this->_identity = null;
-                $this->renewAuthStatus();
+                try {
+                    $this->_identity = null;
+                    $this->renewAuthStatus();
+                } catch (\Exception $e) {
+                    $this->_identity = false;
+                    throw $e;
+                } catch (\Throwable $e) {
+                    $this->_identity = false;
+                    throw $e;
+                }
             } else {
                 return null;
             }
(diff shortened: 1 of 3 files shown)


framework/web/User.php
```diff
@@ -187,8 +187,16 @@ public function getIdentity($autoRenew = true)
     {
         if ($this->_identity === false) {
             if ($this->enableSession && $autoRenew) {
-                $this->_identity = null;
-                $this->renewAuthStatus();
+                try {
+                    $this->_identity = null;
+                    $this->renewAuthStatus();
+                } catch (\Exception $e) {
+                    $this->_identity = false;
```
samdark: Method return is specified as `IdentityInterface|null`. Returning `false` breaks this contract.

framework/web/User.php
```diff
@@ -187,8 +187,16 @@ public function getIdentity($autoRenew = true)
...
+                    throw $e;
+                } catch (\Throwable $e) {
+                    $this->_identity = false;
```
samdark: Method return is specified as `IdentityInterface|null`. Returning `false` breaks this contract.

samdark (commented):
Looks correct. Please add a unit test and a line for changelog.

SilverFire (approved):

samdark:
Merged. Thank you!

PR #15633, opened 2018-02-10
Deprecate some things going away/changing in 2.1

brandonkelly:
Obviously not all-inclusive, but any isolated changes in 2.1 that were easy to replicate in a non-breaking way, are now also in 2.0, with `@deprecated` tags on the deprecated classes/methods. This may help ease the migration to 2.1 a bit, with zero cost to the 2.0 codebase complexity.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


framework/BaseYii.php
```diff
@@ -380,21 +381,33 @@ public static function setLogger($logger)
...
 
     /**
-     * Logs a trace message.
+     * Logs a debug message.
      * Trace messages are logged mainly for development purpose to see
      * the execution work flow of some code. This method will only log
      * a message when the application is in debug mode.
      * @param string|array $message the message to be logged. This can be a simple string or a more
      * complex data structure, such as array.
      * @param string $category the category of the message.
```
samdark: `@since` needed.

framework/base/InvalidParamException.php
```diff
@@ -12,6 +12,7 @@
  *
  * @author Qiang Xue <qiang.xue@gmail.com>
  * @since 2.0
+ * @deprecated 2.0.14 Use [[InvalidArgumentException]]
  */
 class InvalidParamException extends \BadMethodCallException
```
samdark: Should now extend `InvalidArgumentException` else it would break backwards compatibility.

samdark (changes requested):

framework/base/InvalidParamException.php
samdark: Exceptions are often checked with `instanceof`.

rob006: `InvalidArgumentException` should extend `InvalidParamException` to keep BC - only then `catch (InvalidParamException $e)` will work as before.

framework/web/ErrorHandler.php
```diff
@@ -482,7 +482,7 @@ public function argumentsToString($args)
      */
     public function getExceptionName($exception)
     {
-        if ($exception instanceof \yii\base\Exception || $exception instanceof \yii\base\InvalidCallException || $exception instanceof \yii\base\InvalidParamException || $exception instanceof \yii\base\UnknownMethodException) {
+        if ($exception instanceof \yii\base\Exception || $exception instanceof \yii\base\InvalidCallException || $exception instanceof \yii\base\InvalidArgumentException || $exception instanceof \yii\base\UnknownMethodException) {
```
rob006: I think we should leave it unchanged. We should also handle old `InvalidParamException` in this way - if `InvalidArgumentException` will extend `InvalidParamException` then `instanceof InvalidParamException` will handle both.

This also applies to many try-catch blocks, which may catch user exceptions.

rob006 (changes requested):
Docs should be adjusted.

framework/base/InvalidParamException.php
brandonkelly: Doh, not sure how I forgot to do that – it was the plan.

framework/web/ErrorHandler.php
brandonkelly: No it’s the other way around. `InvalidParamException` must extend `InvalidArgumentException`. Otherwise you would always be dealing with a deprecated class (either directly or via its parent).

samdark (approved):

brandonkelly:
@rob006 Regarding docs – considering 2.0 requires PHP 5.4+ and `::class` wasn't introduced until PHP 5.5, seems like we leave the `::className()` references in-place in the docs. What do you think?

brandonkelly:
(Same reason I didn’t actually change any `::className()` references in the code to `::class`.)

framework/web/ErrorHandler.php
rob006: > Otherwise you would always be dealing with a deprecated class (either directly or via its parent).

This is the cost of BC. You can't change thrown exception in this way, because all `instanceof` or `catch` will stop working correctly.

rob006:
Personally I would deprecate `className()` in 2.1 and remove in 2.2. One additional method will not hurt anyone.

framework/web/ErrorHandler.php
brandonkelly: OK I see what you’re saying. I just reverted most of the `InvalidArgumentException` changes; now that class exists as a subclass of `InvalidParamException`, and `InvalidParamException` has been deprecated, so apps can choose to throw `InvalidArgumentException`s in their own code now, but Yii itself still only throws/catches/checks instanceof `YiiParamException` until 2.1.

rob006: You misunderstood me. I still think that we should throw new `InvalidArgumentException` everywhere. This will not break BC since all `$exception instanceof InvalidParamException` and `catch (InvalidParamException $exception)` will work as before. But we need to use old `InvalidParamException` for `instanceof` and `catch` in framework, to make sure that old apps and extensions, which may throw old exception, will work as before. In this way users may start using  new exceptions everywhere (even for `instanceof` and `catch` for exceptions thrown by framework), and old code will still work.

brandonkelly: @rob006 Gotcha - OK now everything throws `InvalidArgumentException`s, but anything that catches `InvalidParamExceptions` or checks for `instanceof InvalidParamException` has been left alone.

Also I’ve left `@throws InvalidParamException` tags alone in `yii\db\conditions\ConditionInterface`, `yii\rbac\CheckAccessInterface`, and `yii\rbac\MessageInterface`, since changing them in an interface felt wrong. Let me know if you think that wasn’t the right call.

rob006 (approved):

samdark:
Merged. Thank you!

framework/base/InvalidParamException.php
bologer: @samdark  is it okay that `InvalidParamException` is now deprecated and suggests to use `InvalidArgumentException` in the doc block and `InvalidArgumentException` is extending  `InvalidParamException` already deprecated exception type?

brandonkelly: @bologer that was discussed here: https://github.com/yiisoft/yii2/pull/15633/commits/7f4bd69dfb07214425297878e0b886a9ea364244#r167403068

brandonkelly: @bologer that was discussed here: https://github.com/yiisoft/yii2/pull/15633/commits/7f4bd69dfb07214425297878e0b886a9ea364244#diff-7a2fbae94c890a0d999b38f8a82fe81eR485

PR #15640, opened 2018-02-11
Failing tests for 'like' escaping #15630

rob006:
Proper tests for #15630.

/cc @samdark


samdark:
Thanks for test, it's now here: https://github.com/yiisoft/yii2/pull/15642

PR #15641, opened 2018-02-11
Added BetweenConditionBuilder

### LINKED ISSUE #11611: Query builder between 2 columns

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | Replaces #12678, Fixes #11611

### DIFF
--- a/framework/db/conditions/BetweenColumnsCondition.php
+++ b/framework/db/conditions/BetweenColumnsCondition.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace yii\db\conditions;
+
+use yii\base\InvalidArgumentException;
+use yii\db\ExpressionInterface;
+use yii\db\Query;
+
+/**
+ * Class BetweenColumnCondition represents a `BETWEEN` condition where
+ * values is between two columns. For example:
+ *
+ * ```php
+ * new BetweenColumnsCondition(42, 'BETWEEN', 'min_value', 'max_value')
+ * // Will be build to:
+ * // 42 BETWEEN min_value AND max_value
+ * ```
+ *
+ * And a more complex example:
+ *
+ * ```php
+ * new BetweenColumnsCondition(
+ *    new Expression('NOW()'),
+ *    'NOT BETWEEN',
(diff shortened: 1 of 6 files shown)


framework/db/conditions/BetweenColumnsCondition.php
```diff
@@ -0,0 +1,115 @@
...
+ *
+ * @author Dmytro Naumenko <d.naumenko.a@gmail.com>
+ * @since 2.0.14
+ */
+class BetweenColumnsCondition implements ConditionInterface
+{
+    /**
+     * @var string $operator the operator to use (e.g. `BETWEEN` or `NOT BETWEEN`)
+     */
+    protected $operator;
```
samdark: Do we need these to be protected instead of private?

samdark (approved):

framework/db/conditions/BetweenColumnsCondition.php
SilverFire: Well, to be honest all properties is conditions may be private. They are not intended to be reused, except abstract class `ConjunctionCondition`. I can make all 'em private with a separate PR

samdark: Why separate PR? Let's make em private, I think.

SilverFire: Ok

PR #15642, opened 2018-02-11
15630 fixed like escaping

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15630

### DIFF
--- a/framework/db/conditions/InCondition.php
+++ b/framework/db/conditions/InCondition.php
@@ -5,7 +5,7 @@ namespace yii\db\conditions;
 use yii\base\InvalidArgumentException;
 
 /**
- * Class LikeCondition represents `IN` condition.
+ * Class InCondition represents `IN` condition.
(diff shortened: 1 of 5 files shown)


SilverFire (approved):

SilverFire:
The test fails

samdark:
Yes. But the issue is fixed. It just seems that LIKE conditions are database-specific so test itself should be adjusted.

sergeymakinen:
I would like to review it tomorrow if it’s possible.

rob006:
I simplified test case, this should be enough to test #15630.

samdark:
Accidentally squashed your tests as my commit. Hope you don't mind...

samdark:
Thanks!

PR #15643, opened 2018-02-11
Implement ExpressionInterface in Query

### LINKED ISSUE #10855: Boolean subquery support
Currently to generate the condition,

``` sql
 AND (parent_id = 1 OR (SELECT 1 FROM `table` WHERE id = `table`.parent_id AND parent_id = 1))
```

I have to hack `QueryBuilder` a bit,

``` php
->andWhere(['', 'parent_id = 1 OR', (new Query)->select([new Expression(1)])->from('table')->where('`parent_id` = 1 AND `id` = `table`.`parent_id`')])
```

Would feel a bit cleaner if you could do the below.

``` php
->andWhere(['or', 'parent_id = 1', (new Query)->select([new Expression(1)])->from('table')->where('`parent_id` = 1 AND `id` = `table`.`parent_id`')])
```

### LINKED ISSUE #11218: query builder cannot use subquery in where with '< >' operators
consider this query:

``` sql
SELECT comment.*
FROM comment
WHERE
(
    SELECT COUNT(*) FROM comment AS c
    WHERE c.postId = comment.postId
    AND c.timestamp >= comment.timestamp
    ORDERY BY c.timestamp DESC
) <= 5
AND postId IN (42,43,44,35,8)
ORDER BY timestamp DESC
```

since where( '<=' ... only accepts a column name and a specific value it appears that query builder cannot build this query.

based on how query builder works, i would expect the following to run the above query:

``` php
$subQuery = (new \yii\db\Query())
    ->select( ['COUNT(*)'] )
    ->from( Comment::tableName().' c2' )
    ->where( [
        'and',
        'c2.postId = comment.postId',
        'c2.timestamp >= comment.timestamp' ] )
    ->orderBy( ['c2.timestamp' => SORT_DESC] );

return $query
    ->addSelect( ['comment.*'] )
    ->where( ['<=', $subQuery, 5] )
    ->andWhere( ['postId' => array_keys( $posts )] )
    ->orderBy( ['timestamp' => SORT_DESC] );
```

### LINKED ISSUE #14732: \yii\db\Query::where(): Unable to compare two subqueries with operator format

```php
$query->where(['>', $subquery1, $subquery2]);
```

A condition consisting of two subqueries with the operator between them, such as:

`WHERE (SELECT COUNT(*) FROM table1 ...) > (SELECT COUNT(*) FROM table2 ...)`

### What do you get instead?

>strpos() expects parameter 1 to be string, object given

In QueryBuilder line 1460

### LINKED ISSUE #14810: Allow a second operand of LIKE to be a subquery
Currently subquery instead of a value are supported in many places but not LIKE operator. For example:
```PHP
SomeModel::find()->where(['=', 'field', OtherModel::find()->select('field2')->where([...])]);
```
works, while the same with `LIKE` instead of `=` does not:
```PHP
SomeModel::find()->where(['LIKE', 'field', OtherModel::find()->select('field2')->where([...])]);
```
Support for subqueries must be added for LIKE operator too.

SilverFire:
This changes makes possible to build `Query` object using `queryBuilder->buildExpression()` and use unified interface.

After this changes, it is possible to use Query in conditions, where `Expression` was allowed before. Maybe later we could move logic from the `QueryBuilder::build()` method away to `QueryExpressionBuilder::build()` in order to make `QueryBuilder` class less massive.

Also this PR changes visibility for properties inside `Condition` classes from `protected` to `private`. Thinking about marking conditions with keyword `final` to discourage their extending and force composition over inheritance.

Fixes #14732, fixes #11218, fixes #14810, fixes #10855

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -47,7 +47,7 @@ use yii\base\InvalidConfigException;
  * @author Carsten Brandt <mail@cebe.cc>
  * @since 2.0
  */
-class Query extends Component implements QueryInterface
+class Query extends Component implements QueryInterface, ExpressionInterface
 {
     use QueryTrait;
 
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -161,6 +161,7 @@ class QueryBuilder extends \yii\base\BaseObject
     protected function defaultExpressionBuilders()
     {
         return [
+            'yii\db\Query' => 'yii\db\QueryExpressionBuilder',
             'yii\db\PdoValue' => 'yii\db\PdoValueBuilder',
             'yii\db\Expression' => 'yii\db\ExpressionBuilder',
             'yii\db\conditions\ConjunctionCondition' => 'yii\db\conditions\ConjunctionConditionBuilder',
@@ -284,6 +285,10 @@ class QueryBuilder extends \yii\base\BaseObject
             if (!isset($this->expressionBuilders[$className])) {
                 throw new InvalidArgumentException('Expression of class ' . $className . ' can not be built in ' . get_class($this));
             }
+        }
+
+        if ($this->expressionBuilders[$className] === __CLASS__) {
+            return $this;
         }
 
         if (!is_object($this->expressionBuilders[$className])) {
--- a/framework/db/QueryExpressionBuilder.php
+++ b/framework/db/QueryExpressionBuilder.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace yii\db;
+
+/**
+ * Class QueryExpressionBuilder is used internally to build [[Query]] object
+ * using unified [[QueryBuilder]] expression building interface.
+ *
+ * @author Dmytro Naumenko <d.naumenko.a@gmail.com>
+ * @since 2.0.14
+ */
+class QueryExpressionBuilder implements ExpressionBuilderInterface
+{
+    use ExpressionBuilderTrait;
+
+    /**
+     * Method builds the raw SQL from the $expression that will not be additionally
+     * escaped or quoted.
+     *
+     * @param ExpressionInterface|Query $expression the expression to be built.
+     * @param array $params the binding parameters.
+     * @return string the raw SQL that will not be additionally escaped or quoted.
+     */
+    public function build(ExpressionInterface $expression, array &$params = [])
+    {
+        list($sql, $params) = $this->queryBuilder->build($expression, $params);
+
+        return "($sql)";
+    }
+}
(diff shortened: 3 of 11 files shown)


samdark:
The idea is OK.

rob006:
> Thinking about marking conditions with keyword final to discourage their extending and force composition over inheritance.

I don't think that this is a good reason to make it final. It is developer decision whether to prefer composition over inheritance. It makes sense if it will simplify keeping BC, but it will make code less flexible and I'm not sure if it is worth it.

rob006:
This probably will fix  #14732, #11218 and #14810.

SilverFire:
Good point. I'd stick to it.

SilverFire:
It also should fix #10855

SilverFire:
Fixed tests. Ready for review and merge

rob006 (approved):

samdark (approved):

samdark:
Looks great. Seems we've messed up HHVM again :(

SilverFire:
Thanks for fast response

SilverFire:
Added CHANGELOG in 3c966f991

brandonkelly:
Shouldn’t ExpressionInterface require a `getParams()` method, rather than [assuming](https://github.com/yiisoft/yii2/blob/d16586334d7bea226a67aa8db28982848b5c92dd/framework/db/ExpressionBuilder.php#L21) that its implementations will define a `params` property?

At least in other cases where a property is assumed, there is a a Trait to go along with the interface, that the interface recommends implementing classes should use. (For example, ActiveQueryInterface [instructs](https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveQueryInterface.php#L17) implementing classes to use ActiveQueryTrait.)

samdark:
@SilverFire I think it should require `getParams()`...

brandonkelly:
While at it, maybe requiring `getExpression()` wouldn’t hurt too, so the builder isn’t relying on `__toString()`.

SilverFire:
> rather than [assuming](https://github.com/yiisoft/yii2/blob/d16586334d7bea226a67aa8db28982848b5c92dd/framework/db/ExpressionBuilder.php#L21) that its implementations will define a `params` property?

The `ExpressionBuilder` is a very specialized class that knows how to build the `yii\db\Expression` object. This fact it is explicitly hinted [in PHPDoc for `$expression` parameter](https://github.com/yiisoft/yii2/blob/d16586334d7bea226a67aa8db28982848b5c92dd/framework/db/ExpressionBuilder.php#L17).
Both `getParams()` method and `params` property are described in `yii\db\Expression`.

> @SilverFire I think it should require getParams()...

So the answer is: no, the `ExpressionInterface` should require nothing. Developers can create their own expression classes, specialized builders and connect them together in `QueryBuilder::expressionBuilders` property.

SilverFire:
Btw, fixed wrong documentation https://github.com/yiisoft/yii2/commit/5da39b4a7e3a903f8fdf3dbfe1ec37d99bf25871

PR #15660, opened 2018-02-17
Update UPGRADE.md

antonmarin:
No Yii::debug() method in 2.0.13.1

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | 
| Fixed issues  | Error in documentation

### DIFF
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -82,7 +82,6 @@ Upgrade from Yii 2.0.13
 
   - Replace `::className()` calls with `::class` (if you’re running PHP 5.5+).
   - Replace usages of `yii\base\InvalidParamException` with `yii\base\InvalidArgumentException`.
-  - Replace calls to `Yii::trace()` with `Yii::debug()`.
   - Remove calls to `yii\BaseYii::powered()`.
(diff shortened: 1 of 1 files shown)


rob006:
> No Yii::debug() method in 2.0.13.1

These are instructions for upgrading *to* 2.0.14, and it will have such method: https://github.com/yiisoft/yii2/blob/78d54bc54148d2834a680501c2dcdf0054bad3cf/framework/BaseYii.php#L382-L397

antonmarin:
But its in section fron 2.0.12

rob006:
No, it's not.

antonmarin:
Really. Sorry. Why section "from 2.0.13" exists without 2.0.14? I upgraded to 2.0.13.1 and thought that i should check that section too

rob006:
You should check `UPGRADE.md` from release, which you're using. There is no such section in 2.0.13.1: https://github.com/yiisoft/yii2/blob/2.0.13.1/framework/UPGRADE.md
`UPGRADE.md` from `master` branch is for current `master` status - you should use these instructions only i you're using this branch.

antonmarin:
I see now, thanks

Issue #15662, opened 2018-02-18
Incorrect FileTarget behavior when target is disabled


alexeevdv:
### What steps will reproduce the problem?
Add FileTarget to the logger targets and set `enabled` attribute to `false`

### What is the expected result?
No output is produced

### What do you get instead?
Directory for logFile is created.

### Additional info
Directory is created in `init()` method. Seems like `export()` would be more apropriate place for this.

| Q                | A
| ---------------- | ---
| Yii version      | `2.0.13`
| PHP version      | `5.*, 7.*`
| Operating system |`GNU/Linux`

SilverFire:
Would you like to create a Pull Request?

SilverFire:
Merged, thank you

PR #15663, opened 2018-02-18
\yii\log\FileTarget directory creation is moved to export() method

alexeevdv:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15662

### DIFF
--- a/framework/log/FileTarget.php
+++ b/framework/log/FileTarget.php
@@ -72,7 +72,6 @@ class FileTarget extends Target
      */
     public $rotateByCopy = true;
 
-
     /**
      * Initializes the route.
      * This method is invoked after the route is created by the route manager.
--- a/framework/log/FileTarget.php
+++ b/framework/log/FileTarget.php
@@ -85,10 +85,6 @@ class FileTarget extends Target
         } else {
             $this->logFile = Yii::getAlias($this->logFile);
         }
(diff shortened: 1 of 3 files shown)


SilverFire (approved):

framework/log/FileTarget.php
```diff
@@ -105,6 +101,11 @@ public function init()
      */
     public function export()
     {
+        $logPath = dirname($this->logFile);
+        if (!is_dir($logPath)) {
+            FileHelper::createDirectory($logPath, $this->dirMode, true);
+        }
```
cebe: `export()` can be triggered multiple times in the application flow, its probably better to do it in `init()` but only if `enabled` is true.

cebe: thinking more about it, my suggestion will make it much more complex as the path should still be created after enabling a disabled log target. Looks good to me as is.

cebe (approved):

SilverFire:
Btw, `is_dir` could be changed to `file_exists` as it works a bit faster

alexeevdv:
`is_dir` check is not needed actually. Since it is performed inside `FileHelper::createDirectory` anyway

samdark (approved):

PR #15664, opened 2018-02-18
Check db driver have useMaster or not for calling.

jafaripur:
Breaking change in mongodb or another driver.
In checking unique validation in mongodb validator error occurred in calling useMaster but in mongodb this is not exist and this is breaking change in older driver.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -142,7 +142,7 @@ class UniqueValidator extends Validator
 
         $modelExists = false;
 
-        if ($this->forceMasterDb) {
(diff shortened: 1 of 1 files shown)


SilverFire (approved):

SilverFire:
I'd rather add this method to `yii2-mongodb` and make in dumb.
https://github.com/yiisoft/yii2-mongodb/blob/master/src/Connection.php

@klimov-paul

samdark:
I'd merge it. Safer this way and doesn't cost that much.

samdark:
Thanks!

Issue #15661, opened 2018-02-17
Object of class yii\db\JsonExpression could not be converted to string using batchInsert


khvalov:
### What steps will reproduce the problem?

postgres field contains jsonb type, trying to make batch Insert using: createCommand()->batchInsert()->execute();

### What is the expected result?
Insert fields into DB table

### What do you get instead?
Error message: 

HP Recoverable Error – yii\base\ErrorException
Object of class yii\db\JsonExpression could not be converted to string
1. in /private/var/www/zzz/www/vendor/yiisoft/yii2/db/pgsql/QueryBuilder.php at line 466
                                        $value = 'FALSE';
                } elseif ($value === null) {
                    $value = 'NULL';
                }
                $vs[] = $value;
            }
            $values[] = '(' . implode(', ', $vs) . ')';
        }
        if (empty($values)) {
            return '';
        }
 
        foreach ($columns as $i => $name) {
            $columns[$i] = $schema->quoteColumnName($name);
        }
 
             
### Additional info

looks like _tostring pattern have to be implemented

SilverFire:
Will handle

khvalov:
Same issue with Command::update() method.

samdark:
@khvalov please open new issue. Thanks.

PR #15667, opened 2018-02-18
Fix issue with supports JsonExpresion

khvalov:
Fix for bug https://github.com/yiisoft/yii2/issues/15661 when trying to build query contains JSON(B) object

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | 15661


SilverFire:
It's not so easy to fix :)

https://github.com/yiisoft/yii2/pull/15668

PR #15679, opened 2018-02-19
Update Response.php

pabloneruda1:
set_time_limit() has been disabled for security reasons on some shared hosting

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -417,7 +417,7 @@ class Response extends \yii\base\Response
             return;
         }
 
-        set_time_limit(0); // Reset time limit for big files
+        @set_time_limit(0); // Reset time limit for big files
         $chunkSize = 8 * 1024 * 1024; // 8MB per chunk
 
         if (is_array($this->stream)) {


framework/web/Response.php
```diff
@@ -417,7 +417,7 @@ protected function sendContent()
             return;
         }
 
-        set_time_limit(0); // Reset time limit for big files
+        @set_time_limit(0); // Reset time limit for big files
```
lubosdz: How can setting unlimited script timeout cause error? Supressing errors/warnings globally is not good idea.

samdark:
I don't see a need for the change. Please explain why it is necessary.

berosoboy:
set_time_limit can be disabled in .ini **disable_functions** config.
I suggest to test if function exists before:
```
if (function_exists('set_time_limit')) {
    set_time_limit(0);
}
```

samdark:
Any function could be disabled via `disable_functions`.

berosoboy:
Understood.

pabloneruda1:
Set_time_limit have a significant impact on server performance. Only a few functions affect server performance like this. Maybe it's even the only one. It is logical to disallow it on shared hosting, unlike others functions = I can not use yii\web\Response::sendFile() on shared hosting.
My opinion is that set_time_limit() should not be used here at all. It should be the decision of the developer himself in the controller for a specific case - not globally.

cebe:
if the function is not available on your system you could work around it by creating it:

```php
if (!function_exists('set_time_limit')) {
    function set_time_limit() {};
}
```

brandonkelly:
@cebe We just tried that workaround (craftcms/cms@9ce7f9cc), but servers that disable `set_time_limit()` are now getting a different error:

```
FastCGI: server "/var/run/php71-cgi/php-cgi" stderr: PHP message: PHP Fatal error:  Cannot redeclare set_time_limit() in <FilePath>
```

Which is consistent with [this comment](http://php.net/manual/en/function.function-exists.php#67947) on php.net (emphasis added):

> function_exists will return false for functions disabled with the disable_functions ini directive. **However those functions are still declared so trying to define them yourself will fail.**

So that’s not actually a solution to this.

@samdark:

> Any function could be disabled via disable_functions.

Sure but it seems like disabling `set_time_limit()` specifically is relatively common, if both your users and ours are running into issues with it.

Can you please reconsider? 🙏

lubosdz:
I agree with @pabloneruda1 

> My opinion is that set_time_limit() should not be used here at all. It should be the decision of the developer himself in the controller for a specific case - not globally.

brandonkelly:
I would be fine with that too :)

samdark:
I'm OK with adding function_exists check.

samdark:
Closein since there is issue about it https://github.com/yiisoft/yii2/issues/16531 as nd the pull request can't be merged as is.

PR #15680, opened 2018-02-19
Fixed code style for better readability

alexraputa:
| Q                | A
| ---------------- | ---
| Is bugfix?       | no
| New feature?     | no
| Breaks BC?       | no
| Tests pass?      | yes


SilverFire:
What fixer have you used for this change

alexraputa:
@SilverFire, I used `php-cs-fixer` and hands for some rules :)

SilverFire:
Could you post a fixer config you've applied?

alexraputa:
Do another pull request for it?

SilverFire:
Yes, because we already have one: https://github.com/yiisoft/yii2/blob/master/cs/src/YiiConfig.php

tests/framework/helpers/FileHelperTest.php
```diff
@@ -4,7 +4,6 @@
  * @copyright Copyright (c) 2008 Yii Software LLC
  * @license http://www.yiiframework.com/license/
  */
-
 use yii\helpers\FileHelper;
```
rob006: Why remove this line?

rob006 (changes requested):
Changes in requirement checker should be reverted - it should work with really old PHP versions, it is better to not touch it at all. :P

tests/framework/helpers/FileHelperTest.php
alexraputa: Maybe after updating to the latest version of the `php-cs-fixer`.

PR #15704, opened 2018-02-20
Automatically expand relations in \yii\rest\Serializer

michaelarnauts:
Since #14219 got merged, the foundation for this patch was accepted. Now, the only remaining thing for a clean solution for rest calls is this.

This patch allows, for example, the following in a `\yii\rest\Controller` (or a `\yii\web\Controller` with a `\yii\rest\Serializer` serializer).

```php
    public function actionGetUser(int $id)
    {
        return User::find($id)
            ->with(['services'])
            ->one();
    }

    public function actionGetAllUsers()
    {
        return User::find()
            ->with(['services'])
            ->all();
    }
```

Output to frontend (`/get-user/1`)
```json
{
  "id": 1,
  "name": "Mike",
  "services": [
    {
      "id": 1,
      "name": "Service 1"
    },
    {
      "id": 2,
      "name": "Service 2"
    }
  ]
}
```

Output to frontend (`/get-all-users`)
```json
[
  {
    "id": 1,
    "name": "Mike",
    "services": [
      {
        "id": 1,
        "name": "Service 1"
      },
      {
        "id": 2,
        "name": "Service 2"
      }
    ]
  },
  {
    "id": 2,
    "name": "Bob",
    "services": [
      {
        "id": 1,
        "name": "Service 1"
      },
      {
        "id": 4,
        "name": "Service 4"
      }
    ]
  }
]
```

The output of this function will be passed to the `\yii\rest\Serializer` and it will automatically include the `services` in the JSON output, since you've loaded it with `with()`.

Currently, this is not possible in yii2. You have to revert to using `asArray()` on the query, but that messes up the types of the attributes, since they are not mapped to the model. You also lose the ability to call functions on the objects before sending them to the frontend.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/framework/rest/Serializer.php
+++ b/framework/rest/Serializer.php
@@ -13,6 +13,7 @@ use yii\base\Component;
 use yii\base\Model;
 use yii\data\DataProviderInterface;
 use yii\data\Pagination;
+use yii\db\BaseActiveRecord;
 use yii\helpers\ArrayHelper;
 use yii\web\Link;
 use yii\web\Request;
@@ -118,6 +119,14 @@ class Serializer extends Component
      * @since 2.0.10
      */
     public $preserveKeys = false;
+    /**
+     * @var bool whether to automatically expand relations when serializing models.
+     * Set this to `true` to use the loaded relations of the passed model in serialization. The default is to only
+     * use the attributes of the model, relations are ignored unless explicitly specified in the `expandParam` parameter
+     * of the request.
+     * @since 2.1
+     */
+    public $autoNestedExpand = false;
 
 
     /**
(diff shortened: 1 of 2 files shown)


bboure:
@michaelarnauts Your PR makes me realize something. And I think this is also a bit your concern here too.
`expand` when passed as a parameter lazy loads all the relations (unless you eager loaded them in advance as in your example).
We should eager load them, especially when dealing with collections like in your second example.

samdark:
Won't make it into 2.0.

PR #15713, opened 2018-02-21
Detects circular array references in VarDumper::dumpInternal

### LINKED ISSUE #15506: VarDumper::dump($GLOBALS) recurses into $GLOBALS

    echo VarDumper::dump($GLOBALS)

A dump with `*RECURSION*` shown when accessing `$GLOBALS['GLOBALS']`, like `var_dump()`

### What do you get instead?

A dump of 10 levels (default depth).

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13
| PHP version      | 7.1
| Operating system | Alpine (Linux)

beroso:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes (except hhvm)
| Fixed issues  | #15506

### DIFF
--- a/framework/helpers/BaseVarDumper.php
+++ b/framework/helpers/BaseVarDumper.php
@@ -65,7 +65,7 @@ class BaseVarDumper
      * @param mixed $var variable to be dumped
      * @param int $level depth level
      */
-    private static function dumpInternal($var, $level)
+    private static function dumpInternal($var, $level, $seenArrays = [])
     {
         switch (gettype($var)) {
             case 'boolean':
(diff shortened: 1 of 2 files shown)


klimov-paul:
This does not seem to be appripriate fix as having 2 identical arrays by with different keys will case the latest to be ignored.
Also this does not fix possible recursion around object fields.

berosoboy:
@klimov-paul Thanks for your feedback. I will try to improve it

PR #15718, opened 2018-02-21
Added more rules for 'cs-fixer'

alexraputa:
Related to #15680.

| Q                | A
| ---------------- | ---
| Is bugfix?       | no
| New feature?     | no
| Breaks BC?       | no
| Tests pass?      | yes


cs/src/YiiConfig.php
```diff
@@ -44,7 +44,14 @@ public function __construct($name = 'yii-cs-config')
             'dir_constant' => true,
             'ereg_to_preg' => true,
             'function_typehint_space' => true,
-            'hash_to_slash_comment' => true,
```
alexraputa: DEPRECATED: use `single_line_comment_style` instead.

cs/src/YiiConfig.php
```diff
@@ -64,7 +77,7 @@ public function __construct($name = 'yii-cs-config')
             'no_empty_comment' => true,
             'no_empty_phpdoc' => true,
             'no_empty_statement' => true,
-            'no_extra_consecutive_blank_lines' => [
```
alexraputa: DEPRECATED: use `no_extra_blank_lines` instead.

cs/src/YiiConfig.php
```diff
@@ -82,7 +95,7 @@ public function __construct($name = 'yii-cs-config')
             'no_leading_namespace_whitespace' => true,
             'no_mixed_echo_print' => true,
             'no_multiline_whitespace_around_double_arrow' => true,
-            'no_multiline_whitespace_before_semicolons' => true,
```
alexraputa: DEPRECATED: use `multiline_whitespace_before_semicolons` instead.

cs/src/YiiConfig.php
```diff
@@ -98,20 +109,20 @@ public function __construct($name = 'yii-cs-config')
...
-//                    'constant_private',
-//                    'property_public',
-//                    'property_protected',
-//                    'property_private',
-//                    'construct',
-//                    'destruct',
-//                    'magic',
-//                ],
-//            ],
+            'ordered_class_elements' => [
```
SilverFire: I think thing config part was under discussion because in some classes we have private properties near methods, that use them for caching etc.

alexraputa: Some classes already use this rule. See [Component](https://github.com/yiisoft/yii2/blob/master/framework/base/Component.php).
It will be more consistency.

rob006: See https://github.com/yiisoft/yii2/pull/14679

alexraputa: Ok, I see. Is it possible to exclude some rules?
```
'ordered_class_elements' => [
    'order' => [
        'use_trait',
        'constant_public',
        'constant_protected',
        'constant_private',
       //'property_public',
       //'property_protected',
       //'property_private',
        'construct',
        'destruct',
        'magic',
    ],
],
```
@rob006, @SilverFire, What do you think about it?

alexraputa: No, it's https://github.com/yiisoft/yii2/pull/15718#discussion_r169567238 wrong. Then I can delete this rule, because in this case the properties are behind magic methods.

rob006 (changes requested):
1. This config should work with php-cs-fixer 2.2 - some of these deprecations require 2.3+. 
2. #15680 has a lot more changes than this config will do. Could you just run fixer with this config, so we could see what will be result of these changes?

alexraputa:
1. We can update the dependency.
2. Into the #15680 I made many things with my hands. :)

rob006:
> We can update the dependency.

We can't - 2.2 is the last line that supports PHP 5.4.

alexraputa:
@rob006, I understood. Can we rebase this pull request into the 2.1 branch? Or is it better to close it?

samdark:
Rebased but not sure how good is diff now...

alexraputa:
@samdark, We can remove this https://github.com/yiisoft/yii2/pull/15718/files#diff-61f406df6747f6dfb2816e2799551983R66 rule. Then `php-cs-fixer` will not remove the extra blank line after methods.

samdark:
There are conflicts now also so these should either be resolved or PR should be closed and re-opened for 2.1.

alexraputa:
I''ll try to resolve it.

rob006:
I don't see any sense in merging it to 2.1. This config should be extracted to separate package and used as a dependency. 
I would merge #15680 and close this. I don't think that changing rules names is worth of bumping requirements.

alexraputa:
It seems like weird after merge. I close it.
If it's needed then I'll open a new pull request.

PR #15745, opened 2018-02-22
Updating the DocBlock for the upsert() Function

PioneerGeek:
Changing the return type and description for the DocBlock of the upsert() function.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none


samdark:
Original is correct.

nastron:
Hmm, this doesn't look like an object, at the end of the implementation of the function that gets executed: `return $insertSql . ' ON DUPLICATE KEY UPDATE ' . implode(', ', $updates);`.

PR #15751, opened 2018-02-23
Update QueryBuilder.php

priitparis:
This fix update issue with  ArrayExpression. If you did update record, ArrayExpression was not populated as array but like ArrayExpression. 
Issue was with postgresSql and array column.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes


SilverFire:
Duplicates 15732

PR #15777, opened 2018-02-26
Make cache key more unique (more scoped)

uaoleg:
For example DB Schema model also uses CLASS constant as basis for cache key
https://github.com/yiisoft/yii2-framework/blob/master/db/Schema.php#L713

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/rbac/DbManager.php
+++ b/framework/rbac/DbManager.php
@@ -86,7 +86,7 @@ class DbManager extends BaseManager
      * @see cache
      * @since 2.0.3
      */
-    public $cacheKey = 'rbac';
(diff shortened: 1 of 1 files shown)


lav45:
```php
class Foo
{
	public $cacheKey = __CLASS__;
}

class Baz extends Foo {}

$model = new Baz;

var_dump($model->cacheKey); // string(3) "Foo"
```

samdark:
Won't make it into 2.0. https://github.com/yiisoft/rbac-db/issues/1

PR #15781, opened 2018-02-26
Update Connection.php

antonmarin:
How to disable load balancing

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | 
| Fixed issues  |

### DIFF
--- a/framework/db/Connection.php
+++ b/framework/db/Connection.php
@@ -328,9 +328,9 @@ class Connection extends Component
      */
     public $enableSavepoint = true;
     /**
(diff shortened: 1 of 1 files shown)


antonmarin:
@SilverFire, how should i test docs?

SilverFire:
There are no tests confirming that setting `null` will work as expected. It will be good to test that it works as is documented

antonmarin:
Okay. Understand now

SilverFire:
Merged with tests and docs adjustment: b642c36019790e45e916493318458938165066d0

PR #15802, opened 2018-02-27
Added try catch for yii2/framework/di/Container

ghost:
This can cause unhandled exception.
In my case it was unhandled exception that cause application error without any message with debug enabled:
```
ReflectionException {
  message: "Class baibaratsky\yii\rollbar\Rollbar does not exist",
  code: -1,
  file: "vendor/yiisoft/yii2/di/Container.php",
  line: 429
}
```

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/framework/di/Container.php
+++ b/framework/di/Container.php
@@ -429,7 +429,7 @@ class Container extends Component
         
         try {
             $reflection = new ReflectionClass($class);
-        } catch (ReflectionException $e) {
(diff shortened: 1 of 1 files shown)


developeruz:
Replaced by #16440

PR #15811, opened 2018-03-01
renderDynamic break

xicond:
renderDynamic break while params inside contain string with quote
coz when the string concat then evaluate, it will break the closing quote

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


xicond:
@samdark I branched out from 2.1, will come out to 2.0.14.2 ?

samdark (approved):

samdark:
Please add a line for changelog. Thanks.

xicond:
@samdark added

samdark:
Merged. Thank you!

Issue #15814, opened 2018-03-01
[Regression] [yii\test\ActiveFixture] throws [yii\base\InvalidConfigException] after update to 2.0.14


mikk150:
After upgrading to 2.0.14 when you have Fixture without fixture datafile, it throws now `yii\base\InvalidConfigException`

### What steps will reproduce the problem?
Create fixture without data file and try to load it
### What is the expected result?
It should truncate the table
### What do you get instead?
It throws `yii\base\InvalidConfigException` now instead

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14
| PHP version      |  5.6.33
| Operating system | Ubuntu 16.04

leandrogehlen:
@SilverFire, Do you can remove this exception?

SilverFire:
It seems that nobody have checked this issue yet. First we need to find out why this exception was introduced and whether the reported behavior is expected

leandrogehlen:
I have built this feature, and the exception was added to avoid setting invalid path for fixture data file.
The behavior is all right.

The problem is the BC. Now, it's necessary to define the `property $dataFile = false;` to avoid the execption.

samdark:
@leandrogehlen let's remove it.

leandrogehlen:
@samdark , I was removing the exception, but i found this code:
https://github.com/yiisoft/yii2/blob/master/framework/test/FileFixtureTrait.php#L43.

I have built a test for it: #15840, but  i can't found the problem.

PR #15840, opened 2018-03-05
Added tests for #15814

leandrogehlen:
Added tests for truncate table via fixture

### DIFF
--- a/framework/test/FileFixtureTrait.php
+++ b/framework/test/FileFixtureTrait.php
@@ -35,10 +35,11 @@ trait FileFixtureTrait
      * The file should return the data array that will be stored in [[data]] after inserting into the database.
      *
      * @param string $file the data file path
+     * @param bool $throwException whether to throw exception if fixture data file does not exist.
      * @return array the data to be put into the database
      * @throws InvalidConfigException if the specified data file does not exist.
      */
-    protected function loadData($file)
+    protected function loadData($file, $throwException = true)
     {
         if ($file === false || $file === null) {
             return [];
@@ -53,7 +54,11 @@ trait FileFixtureTrait
             return require $file;
         }
 
-        throw new InvalidConfigException("Fixture data file does not exist: {$file}");
+        if ($throwException) {
+            throw new InvalidConfigException("Fixture data file does not exist: {$file}");
+        }
+
+        return [];
     }
 
 }
--- a/framework/test/FileFixtureTrait.php
+++ b/framework/test/FileFixtureTrait.php
@@ -41,21 +41,19 @@ trait FileFixtureTrait
      */
     protected function loadData($file, $throwException = true)
     {
-        if ($file === false || $file === null) {
-            return [];
-        }
+        if ($file) {
+            if (basename($file) === $file && $this->dataDirectory !== null) {
+                $file = $this->dataDirectory . '/' . $file;
+            }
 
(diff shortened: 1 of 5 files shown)


leandrogehlen:
@samdark, the `ArrayFixtureTest::testWrongDataFileException` is a test to validate this exception.

It seems a little strange that's `ActiveFixture` to have a different behavior of `ArrayFixture`.

Do you want remove this exception for  ArrayFixture too or remove this excption only for ActiveFixture ?

samdark:
For now I want it to be compatible to pre-2.0.14.

samdark (approved):

framework/test/FileFixtureTrait.php
```diff
@@ -35,25 +35,26 @@
...
+     * @param bool $throwException whether to throw exception if fixture data file does not exist.
      * @return array the data to be put into the database
      * @throws InvalidConfigException if the specified data file does not exist.
      */
-    protected function loadData($file)
+    protected function loadData($file, $throwException = true)
     {
-        if ($file === false || $file === null) {
-            return [];
+        if ($file) {
```
alexraputa: @leandrogehlen, @samdark How about this structure of the method?

```php
if (!$file) {
    return [];
}

if (basename($file) === $file && $this->dataDirectory !== null) {
    $file = $this->dataDirectory . '/' . $file;
}

$file = Yii::getAlias($file);
if (is_file($file)) {
    return require $file;
} elseif ($throwException) {
    throw new InvalidConfigException("Fixture data file does not exist: {$file}");
}
```

rugabarbo:
```php
if (!$file) {
    return [];
}

if (basename($file) === $file && $this->dataDirectory !== null) {
    $file = $this->dataDirectory . '/' . $file;
}

$file = Yii::getAlias($file);
if (is_file($file)) {
    return require $file;
}

if ($throwException) {
    throw new InvalidConfigException("Fixture data file does not exist: {$file}");
}
```

leandrogehlen: I think this not works, when file does not exists

leandrogehlen: must be :

```php
if (!$file) {
    return [];
}

if (basename($file) === $file && $this->dataDirectory !== null) {
    $file = $this->dataDirectory . '/' . $file;
}

$file = Yii::getAlias($file);
if (is_file($file)) {
    return require $file;
} elseif ($throwException) {
    throw new InvalidConfigException("Fixture data file does not exist: {$file}");
} else {
   return [];
}

Comment (alexraputa) [reply framework/test/FileFixtureTrait.php]: I think these `else` statement and `return []` are unnecessary. The test should be green after the implementation of my example.

```php
if (is_file($file)) {
    return require $file;
} elseif ($throwException) {
    throw new InvalidConfigException("Fixture data file does not exist: {$file}");
} else {
   return [];
}
```

leandrogehlen:
@samdark, maybe something like:

```php
$result = [];

if (basename($file) === $file && $this->dataDirectory !== null) {
    $file = $this->dataDirectory . '/' . $file;
}

$file = Yii::getAlias($file);
if (is_file($file)) {
    $result = require $file;
} elseif ($throwException) {
    throw new InvalidConfigException("Fixture data file does not exist: {$file}");
}
return $result;

samdark:
Merged. Thank you!

PR #15842, opened 2018-03-06
Update AccessRule.php

frontbear:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

### DIFF
--- a/framework/filters/AccessRule.php
+++ b/framework/filters/AccessRule.php
@@ -172,7 +172,7 @@ class AccessRule extends Component
             && $this->matchController($action->controller)
             && $this->matchCustom($action)
         ) {
-            return $this->allow ? true : false;
+            return $this->allow;
         }
 
         return null;


frontbear:
according to https://github.com/frontbear/yii2/blob/master/framework/filters/AccessRule.php#L28, code `? true : false`  is not necessary

berosoboy:
Don't do this. `AccessRule::$allow` can be `null`, and the null return has a proper meaning in the `allows` method.

berosoboy:
I think that change could be dangerous.
The current implementation ensure the boolean return;
For example, if `AccessRule::$allow` is `null`, it will cause a improper behavior, because null return has a different meaning in the `allows` method.

frontbear:
@berosoboy i see, thanks.

PR #15844, opened 2018-03-06
`grid\ActionColumn` refactored allowing buttons specification and override via array configuration

klimov-paul:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | 

Flexibility of the `ActionColumn` extended:

- allow configuration `ActionColumn::$buttons` via arrays.
- allow override default buttons using arrays, merging configuration
- allow to re-define button icon rendering and defaul buttons via class extension

Usage of Bootstrap glyphicons removed in favor of more neutral construction.
In case this PR accepted, a specific `yii\bootstrap\grid\ActionColumn` class should be created at `yii2-bootstrap`, providing Bootstrap styled icon rendering.

### DIFF
--- a/framework/grid/ActionColumn.php
+++ b/framework/grid/ActionColumn.php
@@ -8,6 +8,8 @@
 namespace yii\grid;
 
 use Yii;
+use yii\base\InvalidConfigException;
+use yii\helpers\ArrayHelper;
 use yii\helpers\Html;
 use yii\helpers\Url;
 
@@ -86,7 +88,8 @@ class ActionColumn extends Column
      * ```
      */
     public $buttons = [];
-    /** @var array visibility conditions for each button. The array keys are the button names (without curly brackets),
+    /**
+     * @var array visibility conditions for each button. The array keys are the button names (without curly brackets),
      * and the values are the boolean true/false or the anonymous function. When the button name is not specified in
      * this array it will be shown by default.
      * The callbacks must use the following signature:
@@ -104,6 +107,10 @@ class ActionColumn extends Column
      *     'update' => \Yii::$app->user->can('update'),
      * ],
      * ```
+     *
+     * Note that visibility of the particular button can be also controlled via 'visible' key in its array configuration.
+     * Usage of [[visibleButtons]] make sense mostly for buttons specified via `Closure`.
+     *
      * @since 2.0.7
      */
     public $visibleButtons = [];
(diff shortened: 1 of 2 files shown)


framework/grid/ActionColumn.php
```diff
@@ -213,20 +206,103 @@ protected function renderDataCellContent($model, $key, $index)
...
+    {
+        if (isset($this->visibleButtons[$name])) {
+            $isVisible = $this->visibleButtons[$name] instanceof \Closure
+                ? call_user_func($this->visibleButtons[$name], $model, $key, $index)
+                : $this->visibleButtons[$name];
+        } else {
+            $isVisible = true;
+        }
+
+        if (!$isVisible || !isset($this->buttons[$name])) {
```
samdark: Isn't it better to error at the very start of the method in case there is non-existing button used in template?

klimov-paul: We have never done such a thing.
If you are sure code behavior should be changed in this way - I will add an exception.

samdark: That is case for incorrect templates only as far as I can see. I think it is better to error in this case, otherwise typos won't be hard to spot and fix.

samdark: Let's do it.

schmunk42 (approved):
LGTM overall

buttflattery:
Will this update allow us to extend the ActionColumn Class and override the buttons overall, for example, i want to use material icons for the `edit`, `update` and `delete` links in the `GridView` overall in my site.
i currently have to override the buttons in every grid used in my application like below 

```
[
    'class' => 'yii\grid\ActionColumn',
    'template' => '{view}{update}{delete}',
    'buttons' => [
        'view' => function ($url, $model, $key) {
            return Html::a('<i class="material-icons">visibility</i>', $url);
        },
        'update' => function ($url, $model, $key) {
            return Html::a('<i class="material-icons">edit</i>', ['/lead/update', 'id' => $model->lead_id]);
        },
        'delete' => function($url, $model, $key) {
            return Html::a('<i class="material-icons">delete</i>', ['/lead/update', 'id' => $model->lead_id]);
        }
    ]
],
```

i tried to extend the class `yii\grid\ActionColumn` so that i could override the default buttons, but noticed that the `initDefaultButtons()` and `initDefaultButton()` both are declared as `protected` so what looks like currently that i can do this and have to override in every `GridView` used.

samdark:
Won't make it into 2.0.

Issue #15845, opened 2018-03-07
Allow to set undefined properties in yii\base\Object 


mikehaertl:
I have to come back to my closed issue #12021. It seems that this is still not fixed in Yii 2.1.

My description and use case there is still valid.

The outcome of #12021 was, that this should be implemented via a trait. But it seems this never happened in 2.1.

**tl;dr**: I want to use the magic getter/setter methods from `yii\base\BaseObject` to be available in my custom classes. But I don't want that this changes PHP's default behavior, where you can also set undefined properties on a class.

``` php
<?php
class X
{
}

$x = new X;
$x->demo = 'x';   // No error, PHP allows to have dynamic properties

echo $x->wtf;  // Error, because 'wtf' was not set yet.
```

Yii changes this default behavior of PHP:

``` php
<?php
class X extends yii\base\BaseObject
{
}

$x = new X;

// This will throw an error:
$x->demo = 'x';
```

berosoboy:
You can create your custom Object class that extends yii\base\Object and overrides the __set with your proposed code:

```php
    public function __set($name, $value)
    {
        $setter = 'set' . $name;
        if (method_exists($this, $setter)) {
            $this->$setter($value);
        } elseif (method_exists($this, 'get' . $name)) {
            throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
        } else {
            // Instead of throwing exception:
            $this->$name = $value;
        }
    }
```

And your custom classes can extends your custom Object;

samdark:
OK. Let's have another look at it. Yii indeed enhances default PHP objects behavior adding several features:

- Getters/setters.
- Strict errors when non-existing property is set.

On top if it, Component adds:

- Events.
- Behaviors.

Ideally we want to divide these into traits and be able to use any combination of these traits for any object.

The problem is that same `__get()`, `__set()` and other magic methods are used by all these features therefore I don't immediately see a good way to achieve that.

samdark:
We could probably add a specialized trait that could be configured with an array of handlers for the purpose but I do not see any way for it to be as performant as current solution.

mikehaertl:
I don't know a perfect way either. Each thinkable solution has shortcomings. Still, my current favourite is having a (*optional!!*) class property like:

```
public $__strictGetSet = false;
```

*Only if defined* this would disable checks for existing properties. The default would be the same behavior we have now.

mikehaertl:
[...]

PR #15856, opened 2018-03-08
Issue #15845 Implement LooseSetInterface

mikehaertl:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | let's see ...
| Fixed issues  | #15845

### DIFF
--- a/framework/base/BaseObject.php
+++ b/framework/base/BaseObject.php
@@ -124,10 +124,11 @@ class BaseObject implements Configurable
         $getter = 'get' . $name;
         if (method_exists($this, $getter)) {
             return $this->$getter();
-        }        
+        }
         if (method_exists($this, 'set' . $name)) {
             throw new InvalidCallException('Getting write-only property: ' . get_class($this) . '::' . $name);
         }
+
         throw new UnknownPropertyException('Getting unknown property: ' . get_class($this) . '::' . $name);
     }
 
--- a/framework/base/BaseObject.php
+++ b/framework/base/BaseObject.php
@@ -124,7 +124,9 @@ class BaseObject implements Configurable
         $getter = 'get' . $name;
         if (method_exists($this, $getter)) {
             return $this->$getter();
-        } elseif (method_exists($this, 'set' . $name)) {
+        }
+        
+        if (method_exists($this, 'set' . $name)) {
             throw new InvalidCallException('Getting write-only property: ' . get_class($this) . '::' . $name);
         } else {
             throw new UnknownPropertyException('Getting unknown property: ' . get_class($this) . '::' . $name);
--- a/framework/base/BaseObject.php
+++ b/framework/base/BaseObject.php
@@ -124,13 +124,11 @@ class BaseObject implements Configurable
         $getter = 'get' . $name;
         if (method_exists($this, $getter)) {
             return $this->$getter();
-        }
-        
+        }        
         if (method_exists($this, 'set' . $name)) {
             throw new InvalidCallException('Getting write-only property: ' . get_class($this) . '::' . $name);
-        } else {
-            throw new UnknownPropertyException('Getting unknown property: ' . get_class($this) . '::' . $name);
         }
+        throw new UnknownPropertyException('Getting unknown property: ' . get_class($this) . '::' . $name);
     }
 
     /**
--- a/framework/base/BaseObject.php
+++ b/framework/base/BaseObject.php
@@ -149,7 +149,7 @@ class BaseObject implements Configurable
             $this->$setter($value);
         } elseif (method_exists($this, 'get' . $name)) {
             throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
-        } elseif ($this instanceof LooseGetSetInterface) {
+        } elseif ($this instanceof LooseSetInterface) {
             $this->$name = $value;
         } else {
             throw new UnknownPropertyException('Setting unknown property: ' . get_class($this) . '::' . $name);
--- a/framework/base/BaseObject.php
+++ b/framework/base/BaseObject.php
@@ -124,12 +124,11 @@ class BaseObject implements Configurable
         $getter = 'get' . $name;
         if (method_exists($this, $getter)) {
             return $this->$getter();
-        }
-        if (method_exists($this, 'set' . $name)) {
+        } elseif (method_exists($this, 'set' . $name)) {
             throw new InvalidCallException('Getting write-only property: ' . get_class($this) . '::' . $name);
+        } else {
+            throw new UnknownPropertyException('Getting unknown property: ' . get_class($this) . '::' . $name);
         }
-
-        throw new UnknownPropertyException('Getting unknown property: ' . get_class($this) . '::' . $name);
     }
 
     /**
@@ -150,6 +149,8 @@ class BaseObject implements Configurable
             $this->$setter($value);
         } elseif (method_exists($this, 'get' . $name)) {
             throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
+        } elseif ($this instanceof LooseGetSetInterface) {
+            $this->$name = $value;
         } else {
             throw new UnknownPropertyException('Setting unknown property: ' . get_class($this) . '::' . $name);
         }
--- a/framework/base/LooseSetInterface.php
+++ b/framework/base/LooseSetInterface.php
@@ -1,0 +1,23 @@
+<?php
+/**
+ * @link http://www.yiiframework.com/
+ * @copyright Copyright (c) 2008 Yii Software LLC
+ * @license http://www.yiiframework.com/license/
+ */
+
+namespace yii\base;
+
+/**
+ * LooseSetInterface is an empty interface which can be attached to any class
+ * that extends from [[BaseObject]] to disable strict checks of existing
+ * properties. If a class implements this interface, object instances behave
+ * like PHP objects and allow to set and read-back non-existing properties.
+ *
+ * Note though, that PHP still throws an error if a property is read which was
+ * not set before.
+ *
+ * @since 2.1
+ */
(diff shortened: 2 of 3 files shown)


SilverFire (changes requested):
...

SilverFire:
As the interface does not affect properties getting, I'd rename it to `LooseSetInterface`. Don't rush to apply changes, let's wait for more reviews

framework/base/BaseObject.php
```diff
@@ -150,6 +149,8 @@ public function __set($name, $value)
             $this->$setter($value);
         } elseif (method_exists($this, 'get' . $name)) {
```
SilverFire: Is it expected that existing getter will prevent property setting?

SilverFire:
// Don't amend your commit after review has been stared. It will mess everything

mikehaertl:
>  Don't amend your commit after review has been stared. It will mess everything

Sorry. I thought the [WIP] makes it obvious that I'm still changing things. But will no longer ammend now.

framework/base/BaseObject.php
mikehaertl: If someone only writes a getter and ommits a setter I'd say it's a clear decision make by the developer. So in that case the error about writing a read-only property should still be thrown.

SilverFire:
We usually treat `[WIP]` as "_Don't try to merge it now_"

mikehaertl:
Noted. I only had the problem that the tests still didn't pass because of stupid little typos here and there (couldn't test locally). And I didn't want to clutter up commit history with those tiny little fix commits.

SilverFire:
That's not a problem, we usually squash commits to one when merging long PRs

sergeymakinen:
Just a thought: a name like `LooselySettableInterface` makes easier to understand the interface alters an object `__set` behavior. At least for me.

mikehaertl:
@sergeymakinen Hmm, I see your point but wouldn't `LoosenSetterInterface` be even better then?

sergeymakinen:
@mikehaertl Maybe. I just prefer `-able` names. :)

framework/base/BaseObject.php
```diff
@@ -124,12 +124,11 @@ public function __get($name)
         $getter = 'get' . $name;
         if (method_exists($this, $getter)) {
             return $this->$getter();
-        }
-        if (method_exists($this, 'set' . $name)) {
+        } elseif (method_exists($this, 'set' . $name)) {
```
samdark: No need to change it to `elseif` since there was return.

framework/base/BaseObject.php
```diff
@@ -124,12 +124,11 @@ public function __get($name)
...
             throw new InvalidCallException('Getting write-only property: ' . get_class($this) . '::' . $name);
+        } else {
```
samdark: No need to change it.

framework/base/BaseObject.php
```diff
@@ -150,6 +149,8 @@ public function __set($name, $value)
             $this->$setter($value);
         } elseif (method_exists($this, 'get' . $name)) {
             throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
+        } elseif ($this instanceof LooseSetInterface) {
```
samdark: No need for `else` here. `if` on the next line is enough.

samdark (approved):
Overall it looks good to me. A compromise solution.

framework/base/BaseObject.php
mikehaertl: Shouldn't `__get()` and `__set()` use the same coding style for `if` / `elseif`? This wasn't the case when I started so I thought, let's unify the style:

```
    public function __get($name)
    {
        $getter = 'get' . $name;
        if (method_exists($this, $getter)) {
            return $this->$getter();
        }
        if (method_exists($this, 'set' . $name)) {
            throw new InvalidCallException('Getting write-only property: ' . get_class($this) . '::' . $name);
        }
        throw new UnknownPropertyException('Getting unknown property: ' . get_class($this) . '::' . $name);
    }

```

```
    public function __set($name, $value)
    {
        $setter = 'set' . $name;
        if (method_exists($this, $setter)) {
            $this->$setter($value);
        } elseif (method_exists($this, 'get' . $name)) {
            throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
        } else {
            throw new UnknownPropertyException('Setting unknown property: ' . get_class($this) . '::' . $name);
        }
    }
```

framework/base/BaseObject.php
mikehaertl: See comment above

framework/base/BaseObject.php
mikehaertl: See comment above.

framework/base/BaseObject.php
mikehaertl: Also please note PSR-2: https://www.php-fig.org/psr/psr-2/#51-if-elseif-else

> The keyword elseif SHOULD be used instead of else if so that all control
keywords look like single words.

samdark: I didn't mean `else if`, I mean `if`. `else` isn't needed at all in this case. Yes, it's good to unify it removing useless constructs.

mikehaertl: Ok, I understand that. I'm pretty sure that some PSR recommended to *always* use a `else` / `elseif` block even if the block before ended with a `return`, as the latter is easy to miss and the code is more readable.

But whatever - I'll undo everything.

framework/base/BaseObject.php
mikehaertl: Actually I'm confused now. The current code already does not follow your ideas. The `else` here is not required:

```php
    public function __set($name, $value)
    {
        $setter = 'set' . $name;
        if (method_exists($this, $setter)) {
            $this->$setter($value);
        } elseif (method_exists($this, 'get' . $name)) {
            throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
        } else {
            throw new UnknownPropertyException('Setting unknown property: ' . get_class($this) . '::' . $name);
        }
    }
```

So please paste the code block the way you want it. I don't have time for more discussions.

mikehaertl: Ok, nevermind - it is required. TBH I hate wasting time on this.

mikehaertl: I give up at this point. If I remove the `elseif` as you want, then when a setter exists both, the setter and `$this->$name = $value` gets executed:

```php
$setter = 'set' . $name;
if (method_exists($this, $setter)) {
    $this->$setter($value);
} elseif (method_exists($this, 'get' . $name)) {
    throw new InvalidCallException('Setting read-only property: ' . get_class($this) . '::' . $name);
} 
if ($this instanceof LooseSetInterface) {
    $this->$name = $value;
} else {
    throw new UnknownPropertyException('Setting unknown property: ' . get_class($this) . '::' . $name);
}
```

framework/base/BaseObject.php
samdark: Nope. No PSR recommends that.

framework/base/BaseObject.php
samdark: Ah, right. I've got confused. Leave it as is, it's not that important and I can adjust it later.

framework/base/LooseSetInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * that extends from [[BaseObject]] to disable strict checks of existing
+ * properties. If a class implements this interface, object instances behave
+ * like PHP objects and allow to set and read-back non-existing properties.
+ *
+ * Note though, that PHP still throws an error if a property is read which was
+ * not set before.
+ *
+ * @since 2.1
+ */
+interface LooseSetInterface
```
rob006: `DynamicPropertiesInterface`?

rob006:
`Component` also needs to be adjusted.

What about `canSetProperty()`? Now it returns `false` even if property can be set.

schmunk42 (approved):
I think the name (LooseSetInterface) is fine, compared to the other proposals.

LGTM overall, especially the extensive tests

sergeymakinen (approved):

framework/base/BaseObject.php
SilverFire: Yes, and it's BC safe.

framework/base/LooseSetInterface.php
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * LooseSetInterface is an empty interface which can be attached to any class
+ * that extends from [[BaseObject]] to disable strict checks of existing
+ * properties. If a class implements this interface, object instances behave
+ * like PHP objects and allow to set and read-back non-existing properties.
+ *
+ * Note though, that PHP still throws an error if a property is read which was
+ * not set before.
+ *
+ * @since 2.1
```
SilverFire: Add `@author`

tests/framework/base/LooseSetInterfaceTest.php
```diff
@@ -0,0 +1,206 @@
...
+ */
+
+namespace yiiunit\framework\base;
+
+use yii\base\BaseObject;
+use yii\base\LooseSetInterface;
+use yiiunit\TestCase;
+
+/**
+ * @group base
```
SilverFire: Add 
"@author"

SilverFire (approved):
Good

dmirogin (approved):

mikehaertl:
@rob006 `canSetProperty()` should return `true` after you've set a property. We could argue that it should always return `true` when this interface is attached. We could also return `false` if there's a getter but no setter. I don't know what's best here.

rob006 (changes requested):
This is incomplete, see https://github.com/yiisoft/yii2/pull/15856#issuecomment-371951345

samdark:
Won't make it into 2.0.

Issue #15858, opened 2018-03-08
Broken BaseHtml::collectErrors with showAllErrors = true in 2.0.14.1


FlorinRo:
In a scenario where you have the same error message on multiple fields and want to display all of them BaseHtml::collectErrors will crash with 'Undefined offset' error at line 1260
https://github.com/yiisoft/yii2/blob/e63a459a18924bc1672c473c1bbeadfc1b502879/framework/helpers/BaseHtml.php#L1260

Example:
`$model->getErrors():
Array
(
    [field1] => Array
        (
            [0] => Message1
            [1] => Message2
        )
    [field2] => Array
        (
            [0] => Message1
        )
)
$model->getErrorSummary():
Array
(
    [0] => Message1
    [1] => Message1
    [2] => Message2
)`

`echo Html::errorSummary($model,['showAllErrors'=>true]);`
will throw error

Its because of line 1255:
https://github.com/yiisoft/yii2/blob/e63a459a18924bc1672c473c1bbeadfc1b502879/framework/helpers/BaseHtml.php#L1255
array_unique will preserve keys and Model::getErrorSummary produces an array where the same error message is grouped on consecutive positions. So $lines ends up like this
`Array
(
    [0] => Message1
    [2] => Message2
)`

Simple fix: `$lines = array_values($lines);` after foreach

Here is the relevant stack of the ErrorException:
`
[yii\base\ErrorException:8] yii\base\ErrorException: Undefined offset: 1 in /vendor/yiisoft/yii2/helpers/BaseHtml.php:1260
Stack trace:
#0 /vendor/yiisoft/yii2/helpers/BaseHtml.php(1260): yii\base\ErrorHandler->handleError(8, 'Undefined offse...', '...', 1260, Array)
#1 /vendor/yiisoft/yii2/helpers/BaseHtml.php(1226): yii\helpers\BaseHtml::collectErrors(Array, true, true)
#2 /vendor/yiisoft/yii2/widgets/ActiveForm.php(305): yii\helpers\BaseHtml::errorSummary(Object(...), Array)
`

SilverFire:
Thank you very much for your report. Would you like to help us and try to submit a pull request with fix?

FlorinRo:
ups, it seams someone already fixed this,
thx :)

FabrizioCaldarelli:
Sorry @FlorinRo, I was working on the core so I thought to publish the pull request.

SilverFire:
Fixed by #15859

PR #15859, opened 2018-03-08
15858 broken basehtml collecterrors with showallerrors true

fcaldarelli:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15858


SilverFire (approved):

SilverFire:
@FabrizioCaldarelli I have simplified code and PHPDocs a bit

FabrizioCaldarelli:
@SilverFire You can remove also line with "$lines = array_values($lines);" because after you are not considering $lines array indexes.

I don't like much using variabile by reference ( " foreach ($lines as &$line) {" ), because if you make some changes and forget to apply "&", the result will change, but code is more concise.

SilverFire:
> You can remove also line with `$lines = array_values($lines)`

No, it works as it should giving consistent sequence on function output

> because if you make some changes and forget to apply "&", the result will change, but code is more concise.

Yes, it is possible. That's why it's good to have tests :)

FabrizioCaldarelli:
> No, it works as it should giving consistent sequence on function output

The sequence order is mantained even if there are some index missing. Isn't it?
Infact $line index is neither mentioned nor used inside foreach cycle.

SilverFire:
Sorry, I didn't get you question.

FabrizioCaldarelli:
For example, if $lines array had 2 elements (as test https://github.com/yiisoft/yii2/pull/15859/commits/52263077c057c71147aacdbe7aeaf87a86a3d540#diff-580fa7c1d99b2db708afd866e9d71cb9R1325), with 0 and 2 as indexes values, applying array_values ( row 1260 ) is indifferent, because at next foreach cycle with don't care about indexes.

SilverFire:
Foreach in this method will not be affected, but developer will get array with non-seq indices, that may crash his code

FabrizioCaldarelli:
@SilverFire You are right.

Issue #15850, opened 2018-03-07
AssetManager unnecessarily checks for writable basePath when not configured to publish any assets


monzol:
### What steps will reproduce the problem?
- Configure AssetManager to use assets already in web accessible folder or external assets from CDN
- Set permission on assets folder to read-only (755) (this is a standard security measure on our production servers)

### What is the expected result?
Use the configured assets without publishing or checking for writable publish path

### What do you get instead?
Invalid Configuration – yii\base\InvalidConfigException
The directory is not writable by the Web process: /var/www/htdocs/admin/assets
1. in /var/www/includes/vendor/yiisoft/yii2/web/AssetManager.php at line 215

### Additional info
My AssetManager config:
```
return [
    // Map file names to external resources
    'assetMap' => [
        'jquery.js'         => "//ajax.googleapis.com/ajax/libs/jquery/2.2.4/jquery.js",
        'bootstrap.css'     => "//maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.css",
        'bootstrap.js'      => "//maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.js",
    ],

    // Do not publish Yii assets and dependencies
    'bundles' => [
        'yii\web\YiiAsset'                   => false,
        'yii\web\JqueryAsset'                => false,
        'yii\bootstrap\BootstrapAsset'       => false,
        'yii\bootstrap\BootstrapPluginAsset' => false,
    ],
];
```
My AppAsset class:
```
class AppAsset extends AssetBundle
{
    public $basePath = false;
    public $baseUrl = '@web/assets';

    public $css = [
        'bootstrap.css',
    ];

    public $js = [
        'jquery.js',
        'bootstrap.js',
        'js/yii.js',
        'js/yii.activeForm.js',
        'js/yii.captcha.js',
        'js/yii.gridView.js',
        'js/yii.validation.js',
    ];

    public $depends = [ ];
}
```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14.1
| PHP version      | 5.6.31
| Operating system | Linux CentOS 7

The problem is caused by the AssetManager::init() method where the basePath is checked if it's an existing and writable directory regardless of the need for such check. 

Could this check be moved from the init() method to somewhere where it's more relevant and only perform the check when asset publishing is involved?

[...]

PR #15877, opened 2018-03-11
#15850: Fixed AssetManager to validate basePath only when neccessary

monzol:
Fixed AssetManager to only validate basePath when there are files to be published

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #15850

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.14.2 under development
 ------------------------
 
+- Bug #15850: Fixed `yii\web\AssetManager` to only validate `yii\web\AssetManager::$basePath` when there are files to be published (monzol)
 - Bug #15858: Fixed `Undefined offset` error calling `yii\helpers\Html::errorSummary()` with the same error messages for different model attributes (FabrizioCaldarelli, silverfire)
 - Bug #15783: Regenerate CSRF token only when logging in directly (samdark)
 - Bug #15801: Fixed `has-error` CSS class assignment in `yii\widgets\ActiveField` when attribute name is prefixed with tabular index (FabrizioCaldarelli)
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -203,19 +203,15 @@ class AssetManager extends Component
 
     /**
      * Initializes the component.
-     * @throws InvalidConfigException if [[basePath]] is invalid
      */
     public function init()
     {
         parent::init();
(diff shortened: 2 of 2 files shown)


rob006:
This may decrease performance - multiple `publish()` calls will run this check multiple times. I would rather document workaround from #15850 or just recommend to use some DumbAssetManager.

SilverFire:
Both `is_dir` and `is_writable` use realpath cache so I guess it will affect performance significantly

samdark:
Will or will not?

rob006:
Performance is not the only problem. This PR hides the fact, that component is improperly configured. First `publish()` call will give you error anyway, we should not rely on assumption, that it will not be used.

samdark:
@rob006 exceptions are still there. The difference is that they're triggered on publishing that, considering original issue, seems to be OK.

samdark:
The solution is not correct though: `$basePath` is used in `getAssetUrl()`, `getAssetPath()`, `publishFile()` additionally to `publish()`.

rob006:
@samdark I don't think that #15850 is a valid approach. OP disabled some bundles, and assumes that there will be no other bundles in his app. But this is a very fragile assumption - on some page may be used some extension, which will register additional bundle, which will trigger an error. After this PR such configuration errors will be harder to find, because you need to go to this particular page to realize, that your asset manager has invalid configuration. 

If use case from #15850 requires an asset manager which does not publish anything, there should be an asset manager which does not publish anything. Delaying configuration errors is not a solution, it just hides problem instead solving it.

samdark:
How would you solve it then?

monzol:
@rob006 Then how about leaving the $basePath validity check inside init() but only perform it if `$basePath !== false`? So it could be documented that setting $basePath to false will disable file and directory publishing and you need to manually publish the required assets or configure using external assets. 
This would require adding some extra tests for `$basePath !== false` in a few places in `AssetManager`, but that would probably be lighter than is_dir and is_writable.

rob006:
@samdark One of:
a) Create `DumbAssetManager` which will not publish anything, only log warning when someone is trying to publish something.
b) Add a switch to `AssetManager` to disable asset publishing (https://github.com/yiisoft/yii2/pull/15877#issuecomment-377824564).
c) Mark as WONTFIX since workaround is easy (#15850) and solution **a)** could be done by extension.

I prefer **c)** :P

samdark:
Both `b` and `c` are fine...

PR #15884, opened 2018-03-12
Feature improve base object and component

vkhramtsov:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no

Call class methods where possible. Improve code coverage

### DIFF
--- a/tests/framework/base/BaseObjectTest.php
+++ b/tests/framework/base/BaseObjectTest.php
@@ -9,6 +9,8 @@ namespace yiiunit\framework\base;
 
 use yii\base\BaseObject;
 use yii\base\InvalidCallException;
+use yii\base\UnknownMethodException;
+use yii\base\UnknownPropertyException;
 use yiiunit\TestCase;
 
 /**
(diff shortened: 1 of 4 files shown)


samdark:
As far as I remember, it was made like that intentionally in order to keep performance at acceptable levels.

vkhramtsov:
Waiting for your decision. Currently you are using php 7.1, so it works faster than before

vkhramtsov:
@samdark Any news here? Also I would like to mention that first build fails because of other tests, not mine ;)

samdark:
The tests are OK. Cannot decide on it without performance measurements.

tests/framework/base/BaseObjectTest.php
```diff
@@ -64,25 +65,31 @@ public function testCanSetProperty()
         $this->assertFalse($this->object->canSetProperty('Content'));
     }
 
+    /**
+     * @expectedException \yii\base\UnknownPropertyException
+     */
```
rob006: Do not use annotations for expected exception: https://thephp.cc/news/2016/02/questioning-phpunit-best-practices

vkhramtsov: OK, fixed. I suggest to add this to contribution guide

PR #15941, opened 2018-03-21
`yii\db\ActiveRecord::findOne()` now accepts column names prefixed with table name

### LINKED ISSUE #15931: FindOne error when it using with an associative array or innerJoin
I have a ActiveQuery class that override one() and all() methods to basically add a innerJoin and where clause at every search of that model and the model that joins has a same column name of primary key (at this case `id`). So i need to use findOne() method to get only one AR element searching by primary key or another column.

Retrieves one element of AR after search.

### What do you get instead?
When i try to use `findOne([tableName.columnName => $id])` tells me that `tableName.columnName` is not a column name.
`Key "cota.id" is not a column name and can not be used as a filter`

When i try to use findOne($id) throws a IntegrityException with the message that `id`column is ambiguos at sql.
``The SQL being executed was: SELECT `cota`.* FROM `cota` INNER JOIN `secretaria` ON secretaria.id = cota.fk_secretaria WHERE `id`='1'``

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15
| PHP version      | 7.0
| Operating system | Ubuntu

cebe:
fixes #15931

will tag this as 2.0.15.1 if accepted and merge into master manually afterwards.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15931

### DIFF
--- a/framework/db/ActiveRecord.php
+++ b/framework/db/ActiveRecord.php
@@ -209,8 +209,9 @@ class ActiveRecord extends BaseActiveRecord
         $result = [];
         // valid column names are table column names or column names prefixed with table name
         $columnNames = static::getTableSchema()->getColumnNames();
-        $columnNames = array_merge($columnNames, array_map(function($columnName) {
-            return static::tableName() . ".$columnName";
+        $tableName = static::tableName();
+        $columnNames = array_merge($columnNames, array_map(function($columnName) use ($tableName) {
+            return "$tableName.$columnName";
         }, $columnNames));
         foreach ($condition as $key => $value) {
             if (is_string($key) && !in_array($key, $columnNames, true)) {
--- a/framework/db/ActiveRecord.php
+++ b/framework/db/ActiveRecord.php
@@ -207,7 +207,11 @@ class ActiveRecord extends BaseActiveRecord
     protected static function filterCondition(array $condition)
     {
         $result = [];
+        // valid column names are table column names or column names prefixed with table name
         $columnNames = static::getTableSchema()->getColumnNames();
+        $columnNames = array_merge($columnNames, array_map(function($columnName) {
+            return static::tableName() . ".$columnName";
+        }, $columnNames));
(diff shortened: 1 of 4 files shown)


framework/db/ActiveRecord.php
```diff
@@ -207,7 +207,11 @@ protected static function findByCondition($condition)
     protected static function filterCondition(array $condition)
     {
         $result = [];
+        // valid column names are table column names or column names prefixed with table name
         $columnNames = static::getTableSchema()->getColumnNames();
+        $columnNames = array_merge($columnNames, array_map(function($columnName) {
+            return static::tableName() . ".$columnName";
```
SilverFire: `static::tableName()` could be cached in variable outside anonymous function for performance reasons. Micro-optimization, but anyway

SilverFire (changes requested):
Good otherwise

cebe:
here is the patch to apply to 2.0.13. https://github.com/yiisoft/yii2/commit/c70308a92adbb1d36eb16f4d6bb3409a01fa80c8 please comment on the commit, there is no pull request.

framework/db/ActiveRecord.php
cebe: fixed.

cebe:
> continuous-integration/travis-ci/push — The Travis CI build failed

something is wrong on hhvm but it seems unrelated to these changes.

framework/db/ActiveRecord.php
```diff
@@ -207,7 +207,12 @@ protected static function findByCondition($condition)
     protected static function filterCondition(array $condition)
     {
         $result = [];
+        // valid column names are table column names or column names prefixed with table name
```
samdark: both column names prefixed with table name and not prefixed with table name are valid

PR #15943, opened 2018-03-22
BaseActiveRecord change to store changed attribute to shrink the memory

tsingsun:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no

BaseActiveRecord keep the old attribute,i think keep changed attribute will be better,it also trace the changed attribute and shrink the memory useage

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -92,7 +92,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      * @var array|null old attribute values indexed by attribute names.
      * This is `null` if the record [[isNewRecord|is new]].
      */
-    private $_oldAttributes;
+    private $_changeAttributes;
     /**
      * @var array related models indexed by the relation names
      */
@@ -317,6 +317,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
             ) {
                 $this->resetDependentRelations($name);
             }
+            $this->markAttributeDirty($name);
             $this->_attributes[$name] = $value;
         } else {
             parent::__set($name, $value);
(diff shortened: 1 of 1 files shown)


samdark:
Seems to break tests: https://travis-ci.org/yiisoft/yii2/jobs/356687083

kushalhalder:
How much memory are we saving with this?

tsingsun:
@samdark the PR is used in my project,i will check failed tests

tsingsun:
@kushalhalder sorry,i hava not an exact number,but BaseActiveRecord keep current state's data,backup origin data from database source to $_oldAttributes,when fetch from source,two copies data keep in memory.the PR don't persist $_oldAttributes

framework/db/BaseActiveRecord.php
```diff
@@ -92,7 +92,7 @@
      * @var array|null old attribute values indexed by attribute names.
      * This is `null` if the record [[isNewRecord|is new]].
      */
-    private $_oldAttributes;
+    private $_changeAttributes;
```
Renkas: Why not name it `$_changedAttributes`?

framework/db/BaseActiveRecord.php
```diff
@@ -92,7 +92,7 @@
      * @var array|null old attribute values indexed by attribute names.
```
Renkas: I think you should also change the documentation part if you change the data this variable contains

cebe:
> sorry,i hava not an exact number,but BaseActiveRecord keep current state's data,backup origin data from database source to $_oldAttributes,when fetch from source,two copies data keep in memory.the PR don't persist $_oldAttributes

dependent on how php handles these internally, they could be pointers to the same data which are changed when data is written (copy on write). We should not make assumptions about this but verify how much memory you safe.

tsingsun:
@cebe give the fate,great!  this pull request will be close

PR #15956, opened 2018-03-23
yii\base\Object finally dropped

rugabarbo:
The documentation was also checked. There are no entries for cleaning.


samdark:
Thanks!

FabrizioCaldarelli:
If before there was an Object class, could be useful again to have a class that extends BaseObject?

Obviously not named Object, but YiiObject for example.

rob006:
@FabrizioCaldarelli What for? What is wrong with using `BaseObject` directly?

FabrizioCaldarelli:
Before there was an Object class, is it right? Now removed because Object is a php private name.

My question is : if before there was this class, I think that it was useful to something. So could it be usuful restore it with other name (ex. YiiObject).

rob006:
It was already restored - as a `BaseObject`. It is explained in `Object` description. See also https://github.com/yiisoft/yii2/issues/7936.

Issue #15957, opened 2018-03-23
[2.1] Finally remove deprecated stuff


rugabarbo:
Roadmap Wiki section: https://github.com/yiisoft/yii2/wiki/Roadmap#cleanup – (BTW: will be good to link it with this issue).

For each code marked as `@deprecated` need to do:
* rewrite dependencies
* remove documentation occurrences
* remove code

Аlready done: 
* 800574871419b99442d8be73a51e7f8e9e0d090f
* 4e15d3375633ae92eb0e91fbb0f6050de61b5d5c

samdark:
Linked to this issue from wiki.

rugabarbo:
All deprecations removed (last PR #15963 is waiting for the @SilverFire review). Some things can reappear after merges (2.0.x >> 2.1.x). That's why I'll wait for other 2.1 commits (and merges), then will search deprecated things again. 

Don't close this task.

SilverFire:
Merged #15963

Issue #15716, opened 2018-02-21
Array field in PostgreSQL return ArrayExpression instead of array


tQuant:
There is no point to return ArrayExpression for array fields when loading ActiveRecord from database.
ArrayExpression do not have any usefull information per value. Data type and dimension can be obtained from the table scheme.

### What steps will reproduce the problem?
Create an array postgresql field. Create ActiveRecord model. Load model from database.

### What is the expected result?
Field contain array

### What do you get instead?
Field contain ArrayExpression

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14
| PHP version      | 5.6
| Operating system | Debian 9

SilverFire:
Thank you for opening this issue.

On the first iterations of Arrays support, the AR model's fields got arrays after data population. From some point of discussions, we've decided to create `ArrayExpression` object after data population to keep things more consistent, but I don't remember the exact use cases that pushed me to change the implementation. I'm not really sticked to the current one, so we can discuss and change it, if it turns out it does not make much sense.

We already have two unexpected behaviors related to this decision. See https://github.com/yiisoft/yii2/issues/15710

@yiisoft/core-developers

soulhunter1987:
Yes, I already discovered two issues with this approach. And I think there might be much more.
But as i remember, one of the reasons of choosing such approach was backward incompatibility with query builder and type casting bacause DB query methods accept field values as an array when implicit type casting is required...

klimov-paul:
> From some point of discussions, we've decided to create ArrayExpression object after data population to keep things more consistent, but I don't remember the exact use cases that pushed me to change the implementation. 

This is the matter of lazy loading.
When fetching row with ARRAY or JSON from the database it is unknown whether those fields will be in use anywhere. Thus it is better to leave parsing for the later stage - to the moment it will be in need first time.

Having large JSON or ARRAY data may provide significant impact while being parsing over hunderds of records fetched from DB.

Also not every complex DB type can be converted to an array. For example, while using PostGis extension, its specific type 'geometry' can not be represented as mere array - it should be a specific complex object.

Still, this matter is debatable.

zhdanovartur:
Great feature, but I think ArrayExpression is not a PATCH update, because it broke a lot of model getters, used before for Postgres array fields.

SilverFire:
[...]

PR #15963, opened 2018-03-23
Issue #15957: dropped deprecated ColumnSchema properties

rugabarbo:
See also #15716

### DIFF
--- a/framework/db/mysql/ColumnSchema.php
+++ b/framework/db/mysql/ColumnSchema.php
@@ -19,17 +19,6 @@ use yii\db\JsonExpression;
 class ColumnSchema extends \yii\db\ColumnSchema
 {
     /**
-     * @var bool whether the column schema should OMIT using JSON support feature.
-     * You can use this property to make upgrade to Yii 2.0.14 easier.
-     * Default to `false`, meaning JSON support is enabled.
-     *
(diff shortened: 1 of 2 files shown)


samdark (approved):

SilverFire (approved):

SilverFire:
Thank you!

framework/db/mysql/ColumnSchema.php
```diff
@@ -18,17 +18,6 @@
  */
 class ColumnSchema extends \yii\db\ColumnSchema
 {
-    /**
-     * @var bool whether the column schema should OMIT using JSON support feature.
-     * You can use this property to make upgrade to Yii 2.0.14 easier.
-     * Default to `false`, meaning JSON support is enabled.
-     *
-     * @since 2.0.14.1
-     * @deprecated Since 2.0.14.1 and will be removed in 2.1.
```
rob006: @SilverFire How this is possible to add and deprecate property in the same release? :D

samdark: Why not? :)

SilverFire: Because we can :)

rugabarbo: When you can

PR #15971, opened 2018-03-24
Fix master travis builds

rugabarbo:


samdark:
1. HHVM isn't easy to fix.
2. It's release command that writes the line you've fixed. Should be adjusted as well.

rugabarbo:
> It's release command that writes the line you've fixed. Should be adjusted as well.

Release command is OK: https://github.com/yiisoft/yii2/blob/afe278b6e084a652937aa8aba57f15e2b4318f83/build/controllers/ReleaseController.php#L834

Looks like somebody added this line manually.

rugabarbo:
52ffaa1b25f09f34add73085adb4fd6a0f4f7139
I think @cebe added this line manually while merging.

rugabarbo:
HHVM fixed (see build).

rugabarbo:
Looks like one of HHVM bugs like this: https://github.com/facebook/hhvm/issues/8150 – issue is about **3.25.0** version. Yii2 travis use **3.25.1** version – I think it's affected too. 

HHVM can't open `STDIN` stream, so then can't close it because of it isn't a valid stream resource.

Removing of `fclose` is hack for ignore this bug, I don't think think we must dig it more deeper. We can add fopen result checks (as you do), but It's stupid to check every time whether fopen of STDIN worked properly because of it just bug of HHVM.

tests/framework/validators/NumberValidatorTest.php
```diff
@@ -309,6 +309,12 @@ public function testValidateResource()
         $model->attr_number = $fp;
         $val->validateAttribute($model, 'attr_number');
         $this->assertTrue($model->hasErrors('attr_number'));
+        
+        // the check is here for HHVM that
+        // was losing handler for unknown reason
+        if (is_resource($fp)) {
+            fclose($fp);
+        }
```
rugabarbo: IMO if you want the most correct solution, you must skip this test for HHVM because of idea of the test for HHVM is broken (`$fp` isn't valid resource and the test need a valid resource).

rugabarbo: In general, this call of `fclose` has no any effect in any case. `$fp` is a local var, so a resource with no more references to it is detected automatically, and it is freed by the garbage collector. Isn't it?

samdark: Skipping test completely on HHVM sounds bad for me because stable version works well with it.

samdark:
Thank you for fixing it. Make life much easier.

PR #15976, opened 2018-03-25
Issue #15957: dropped deprecated UniqueValidator::$comboNotUnique

rugabarbo:
See also PR #12305

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -82,13 +82,6 @@ class UniqueValidator extends Validator
      */
     public $message;
     /**
-     * @var string
-     * @since 2.0.9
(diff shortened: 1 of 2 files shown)


sdscbr:
what's the change

SilverFire:
Should we mention each deprecation removing in UPGRADE.md?

rugabarbo:
@SilverFire it seems to me that there will be enough mention that all `@deprecated` things will be deleted: https://github.com/yiisoft/yii2/blame/2.1/framework/CHANGELOG.md#L27 

Also we can give a hint about the possibility of searching for the string '@deprecated'.

samdark:
No, we should not.

PR #16002, opened 2018-03-28
Fixes #14660: fixed synchronous DbCache::set() with the same key

### LINKED ISSUE #14660: Setting the same cache entry twice during the same second makes the 2nd set fail.
When setting the exact same cache entry during the same second, the 2nd `set()` returns false, making it seem like the set has failed, but it didn't. 

The `yii\caching\DbCache` his `setValue()` function first tries an UPDATE to update the cache, and if that fails, it tries an INSERT via the `addValue()` function. This seems fine, except that the UPDATE query returns the number of rows changed, and by default MySQL returns 0 changed rows when the UPDATE is identical to the existing value. I'm not sure what other databases do.
This causes `setValue()` to think the UPDATE has failed and tries a `addValue()` call, what throws an DUPLICATE KEY Exception since the data is already there. This is handled with a `return false` causing the calling functions to think the cache set has failed.

A fix is to set the `PDO::MYSQL_ATTR_FOUND_ROWS` attribute to `true`. This causes MySQL to return the number of found rows instead the number of changed rows in a UPDATE call.

I'm not sure if this is the correct behavior, so a better way could be to change `setValue()` to see the difference between an existing identical cache entry, or a non-existing cache entry.

### Workaround
```php
return [
    ...
    'components' => [
        ...
        'db' => [
            'dsn' => 'mysql:host=*hostname*;dbname=*database*',
            'username' => '*username*',
            'password' => '*password*',
            'attributes' => [
                PDO::MYSQL_ATTR_FOUND_ROWS => true,
            ],
        ],
    ],
];
```

```php
if (!Yii::$app->cache->set('key', 'value', 86400))) {
  throw new Exception("O my god, the cache set has failed. (1)");
}

if (!Yii::$app->cache->set('key', 'value', 86400))) {
  throw new Exception("O my god, the cache set has failed. (2)");
}
```
Returns:
O my god, the cache set has failed. (2)

```php
if (!Yii::$app->cache->set('key', 'value', 86400))) {
  throw new Exception("O my god, the cache set has failed. (1)");
}

sleep(1); // sleep a second, the expiry value will change

if (!Yii::$app->cache->set('key', 'value', 86400))) {
  throw new Exception("O my god, the cache set has failed. (2)");
}
```
Returns nothing.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.0

rugabarbo:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14660

### DIFF
--- a/framework/caching/DbCache.php
+++ b/framework/caching/DbCache.php
@@ -205,6 +205,8 @@ class DbCache extends Cache
 
             return true;
         } catch (\Exception $e) {
+            Yii::warning("Unable to update or insert cache data: {$e->getMessage()}", __METHOD__);
+
             return false;
         }
     }
@@ -234,6 +236,8 @@ class DbCache extends Cache
 
             return true;
         } catch (\Exception $e) {
+            Yii::warning("Unable to insert cache data: {$e->getMessage()}", __METHOD__);
+
             return false;
         }
     }
--- a/framework/caching/DbCache.php
+++ b/framework/caching/DbCache.php
@@ -192,22 +192,21 @@ class DbCache extends Cache
      */
     protected function setValue($key, $value, $duration)
     {
-        $result = $this->db->noCache(function (Connection $db) use ($key, $value, $duration) {
-            $command = $db->createCommand()
-                ->update($this->cacheTable, [
+        try {
+            $this->db->noCache(function (Connection $db) use ($key, $value, $duration) {
+                $db->createCommand()->upsert($this->cacheTable, [
+                    'id' => $key,
                     'expire' => $duration > 0 ? $duration + time() : 0,
                     'data' => new PdoValue($value, \PDO::PARAM_LOB),
-                ], ['id' => $key]);
-            return $command->execute();
-        });
+                ])->execute();
+            });
 
-        if ($result) {
             $this->gc();
 
             return true;
+        } catch (\Exception $e) {
+            return false;
         }
-
-        return $this->addValue($key, $value, $duration);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


rugabarbo:
I suggest to ignore codeclimate issue. Should not be doing such refactoring because the code will lose its flexibility and clarity.

samdark (approved):

framework/caching/DbCache.php
```diff
@@ -192,22 +192,21 @@ protected function getValues($keys)
...
-            return $command->execute();
-        });
+                ])->execute();
+            });
 
-        if ($result) {
             $this->gc();
 
             return true;
+        } catch (\Exception $e) {
```
rob006: What is the reason for hiding this exception?

rugabarbo: I implemented abstract method: https://github.com/yiisoft/yii2/blob/master/framework/caching/Cache.php#L400-L410

As you see it must return boolean and not throw any exception.

But this is a good question: should we hide an exception in the case of DB layer failures?

@samdark, @rob006 what do you think about this?

rob006: We could at least log some errors. Same for `addValue()`.

rugabarbo: Like here? https://github.com/yiisoft/yii2/blob/d880148fd09f827aefee2f418294a72a38579bfb/framework/caching/FileCache.php#L156

samdark:
Merged. Thank you!

PR #16022, opened 2018-04-01
Fix UniqueValidator for PostgreSQL

lav45:
Checks the uniqueness of keys in `jsonb` field.

This add-on allows you to use special characters `[[` , `]]` in the `targetAttribute` settings

```php
class News extends ActiveRecord
{
    public function rules()
    {
        return [
            [['title'], 'unique',
                'targetAttribute' => [
                    'title' => "[[data]]->>'title'",
                ]
            ],
        ];
    }
}
```

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  |


rugabarbo:
Maybe it makes sense to do this for all databases? Not only for PgSQL. MySQL supports JSON-fields too.

lav45:
@requilence You mean move the test out of `yiiunit\framework\db\pgsql\UniqueValidatorTest::testPrepareParams()` => `yiiunit\framework\validators\UniqueValidatorTest::testPrepareParams()` ?
The implementation will work for all databases. Just an error was seen on Postgresql

rugabarbo:
Now I understand. Thanks for the explanation.

cebe (approved):

samdark (approved):
@lav45 would you please add a line for changelog?

lav45:
@samdark done)

samdark:
Merged. Thank you!

PR #16038, opened 2018-04-04
method not allowed exception message translation zh-CN

oraluben:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | None

Translate for zh-CN


cebe:
We have the convention to not translate exception messages:

> When to translate messages messages should be translated when these are displayed to non-tech end user and make sense to him. HTTP status messages, exceptions about the code etc. should not be translated. Console messages are always in English because of encoding and codepage handling difficulties.
> https://github.com/yiisoft/yii2/blob/master/docs/internals/design-decisions.md

samdark:
Push your commits into the same branch please.

PR #16047, opened 2018-04-05
bug fix #16044

### LINKED ISSUE #16044: Cors filter is broken since 2.14
We have some action responding to the OPTIONS calls. 
It is replying with the Access-Control-Allow-Methods set to available verbs. That is required for CORS requests to be done. 

We see the response header.

### What do you get instead?
We have 200 status and no headers.

### Additional info
That is caused by a wrong condition in a Cors filter. 
https://github.com/yiisoft/yii2/blob/master/framework/filters/Cors.php#L109
We never get our options action running.
image

| Q                | A
| ---------------- | ---
| Yii version      | 2.15.1
| PHP version      | 7.1
| Operating system | Linux

ldar:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no
| Fixed issues  |  #16044

### DIFF
--- a/framework/filters/Cors.php
+++ b/framework/filters/Cors.php
@@ -194,6 +194,7 @@ class Cors extends ActionFilter
      */
     protected function prepareAllowHeaders($type, $requestHeaders, &$responseHeaders)
     {
+        $requestHeaders['Access-Control-Request-Headers']='authorization';
         $requestHeaderField = 'Access-Control-Request-' . $type;
         $responseHeaderField = 'Access-Control-Allow-' . $type;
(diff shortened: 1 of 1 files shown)


chenxigua:
dsafdsaf

chenxigua:
good

samdark:
I don't see it anywhere in the spec. What does that change?

ldar:
Sorry. This is my incompetence. My problem can be solved in the configuration
            'cors' => [
                ...
                'Access-Control-Request-Headers' => ['authorization']
            ],

PR #16052, opened 2018-04-06
Fixes #15117 `yii\db\Schema::getTableMetadata` cache refreshing

### LINKED ISSUE #15117: db/Schema::getTableMetadata() broken

boboldehampsink:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15117

Before, setting `$refresh` still tried to pull from the cache. With this fix it doesn't and makes sure it loads from table again.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -8,6 +8,7 @@ Yii Framework 2 Change Log
 - Bug #16010: Fixed `yii\filters\ContentNegotiator` behavior when GET parameters contain an array (rugabarbo)
 - Bug #14660: Fixed `yii\caching\DbCache` concurrency issue when set values with the same key (rugabarbo)
 - Bug #15988: Fixed bash completion (alekciy)
+- Bug #15117: Fixed `yii\db\Schema::getTableMetadata` cache refreshing (boboldehampsink)
 
 2.0.15.1 March 21, 2018
 -----------------------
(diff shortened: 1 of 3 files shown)


cebe (approved):
change looks okay, but its a tricky situation that should be verified by a test.

boboldehampsink:
I agree. Looking at the current schema cache tests they don't make any sense to me? Where in these tests is the schema actually changed? 

https://github.com/yiisoft/yii2/blob/master/tests/framework/db/SchemaTest.php#L108-L135

samdark (approved):

samdark:
Merged. Thank you!

PR #15985, opened 2018-03-26
Callback execution with mutex synchronization

zhuravljov:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

It can be able to execute callback code with mutex synchronization. For example:

```php
$result = $mutex->sync($mutexName, 10, function () {
    // business logic execution with synchronization
    return $result;
}));
```

PR #16054, opened 2018-04-07
Callback execution with mutex synchronization

zhuravljov:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

It can be able to execute callback code with mutex synchronization. For example:

```php
$result = $mutex->sync($mutexName, 10, function () {
    // business logic execution with synchronization
    return $result;
}));
```

Moved from #15985.


samdark (approved):
Add a changelog line and we'll merge it.

zhuravljov:
Done.

Issue #16065, opened 2018-04-09
date()


sashsvamir:
### What steps will reproduce the problem?
1. Add follow code at header of `framework/yii`:
```php
echo date_default_timezone_get();
echo date('Y-m-d H:i:s', time());
// ...
```

2. Create action (to any controller):
```php
public function actionDate() {
    echo date_default_timezone_get();
    echo date('Y-m-d H:i:s', time());
}
```

3. Run above action

### What is the expected result?
```
Europe/Moscow
2018-04-09 08:18:40
Europe/Moscow
2018-04-09 08:18:40
```

### What do you get instead?
```
Europe/Moscow
2018-04-09 08:18:40
UTC
2018-04-09 05:18:40
```

### Additional info

Also i try run `php -r 'phpinfo();` where i get:
```
Europe/Moscow
2018-04-09 08:18:40
```

Note: both yii app and terminal executed from same php and using the same `php.ini` config file, there no diffs at php configs.

Also i'm compared `phpinfo()` output from yii app and from terminal command `php -r 'phpinfo();'`, there found that yii app use default timezone as `UTC`, but php running from terminal using `Europe/Moscow`. There full diff of `phpinfo()`:
```diff
--- Untitled	Mon Apr  9 07:29:42 2018
+++ Untitled2	Mon Apr  9 07:29:42 2018
@@ -209,7 +209,7 @@
 timelib version => 2017.05beta9
 "Olson" Timezone Database Version => 0.system
 Timezone Database => internal
-Default timezone => UTC
+Default timezone => Europe/Moscow
 
 Directive => Local Value => Master Value
 date.default_latitude => 31.7667 => 31.7667
@@ -962,7 +962,7 @@
 LESSOPEN => | /usr/bin/lesspipe %s
 J2REDIR => /usr/lib/jvm/java-8-oracle/jre
 LESSCLOSE => /usr/bin/lesspipe %s %s
-_ => ./yii
+_ => /usr/bin/php
 OLDPWD => /home/sash
 
 PHP Variables
@@ -994,22 +994,21 @@
 $_SERVER['LESSOPEN'] => | /usr/bin/lesspipe %s
 $_SERVER['J2REDIR'] => /usr/lib/jvm/java-8-oracle/jre
 $_SERVER['LESSCLOSE'] => /usr/bin/lesspipe %s %s
-$_SERVER['_'] => ./yii
+$_SERVER['_'] => /usr/bin/php
[...]

PR #16070, opened 2018-04-09
fix timezone

sashsvamir:
In php > 7.0 was removing warning on empty setting `date.timezone`, so now we don't need set this on Application::preInit(), because `date_default_timezone_get()` always return timezone.

#16065


cebe (approved):
looks good, a CHANGELOG entry is needed.

cebe:
Thank you!

PR #16075, opened 2018-04-10
blind fix for oracle InCondition builder

### LINKED ISSUE #16073: Getting error with relation records when we have more 1000 parent records
When model **A** have more 1000 records and we need to get **relation** records:
```php
A::find()->with(['relation'])->limit(1001)->all();
```
Relation records

### What do you get instead?
Error in ./vendor/yiisoft/yii2/db/QueryBuilder.php at line 1544
```php
strtoupper() expects parameter 1 to be string, object given
```
First elements isn`t string. All elements is objects ("in condition"):
```php
$operator = strtoupper(array_shift($condition));
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15
| PHP version      | 7.0.25
| DB | Oracle OCI

cebe:
condition should be an array like `['OR', ['IN', 'col', [...]], ['IN', 'col', [...]], ...]` instead of `['OR', [  ['IN', 'col', [...]], ['IN', 'col', [...]]  ]]`

This should have already broken the test at https://github.com/yiisoft/yii2/blob/ac687aab292088a1a3525f10930e7a1f0403173b/tests/framework/db/oci/QueryBuilderTest.php#L135
but as we are not running oracle tests on travis it got unnoticed.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  |no
| Breaks BC?    | no
| Tests pass?   | **someone with an oracle environment should check this.**
| Fixed issues  | fixes #16073

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -9,6 +9,8 @@ Yii Framework 2 Change Log
 - Bug #14660: Fixed `yii\caching\DbCache` concurrency issue when set values with the same key (rugabarbo)
 - Bug #15988: Fixed bash completion (alekciy)
 - Bug #15117: Fixed `yii\db\Schema::getTableMetadata` cache refreshing (boboldehampsink)
+- Bug #16073: Fixed regression in Oracle `IN` condition builder for more than 1000 items (cebe)
+
 
 2.0.15.1 March 21, 2018
 -----------------------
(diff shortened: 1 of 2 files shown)


samdark (approved):

samdark:
Merged. Confirmed by issue reporter.

cebe:
applied the patch to yii2-oracle  as well: https://github.com/yiisoft/yii2-oracle/commit/8fd52cd400c06c3a5add188393a7bcac976be5da

Issue #15875, opened 2018-03-11
afterSave for new models flushes unsaved data


yujin1st:
After #13567 AR now flushed relational data, so there is new issue:

### What steps will reproduce the problem?
1. Create simple app with Order and OrderItem models.
2. Load data in controller, create relational OrderItem records and populate relation 'items'
```php
// In Order 
public function load($data, $formName = null) {
  if (!parent::load($data, $formName)) return false;
  $this->loadItems($data);
  return true;
}

/// loading and populating data
private function loadItems($data) {
  if (!isset($data['items'])) {
    return;
  }
  $items = $this->items;
  foreach ($data['items'] as $info) {
    $itemId = $info['itemId'];
    if (isset($items[$itemId])) {
      $items[$itemId]->load($info, '');
    } else {
      $item = new OrderItem([]);
      $item->load($info, '');
      $items[$itemId] = $item;
    }
  }
  $this->populateRelation('items',$items);
}


```

3. Validate, do something in beforeSave method and event
3.1 Validate in Order model
3.2 For example warehouse module can check item quantity before order creating in event
4. Save items in afterSave
```php
public function afterSave($insert, $changedAttributes) {
  $this->saveItems();
  parent::afterSave($insert, $changedAttributes);
}

```

5. Do something with new items in afterSave method and event
 
### What is the expected result?
All works fine, order data saved, relational data saved too

### What do you get instead?
After #13567 items relation with unfinished data is flushed, as Order model changes it's id, so you lose data

### Additional info
It simple workflow allows to validate and process relational form data as ActiveRecord models with all it features and controllers remains thing. And you could use scenarios for complex relations, so models can be easily tested.

I'm afraid this issue will be closed due contradiction with updated framework design, but i have a multiply applications that follows this workflow, so i don't know what to do ;(((

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14

samdark:
@markmarco16, @rob006, @Kolyunya thoughts?

cebe:
which operation causes the reset of relational data? A one-many relation has no attribute in the order that should cause resetting the relation when modified.

samdark:
Indeed, that should not be related to #13567 as there are no relation attributes in the model you modify... @yujin1st are you sure your problem caused by #13567?

yujin1st:
I'm sure. Debugging leads to this code:

```php
ActiveRecord class:

protected function insertInternal($attributes = null){
//...
//lines 518
[...]

PR #16085, opened 2018-04-11
Bug: afterSave for new models flushes unsaved data

shirase:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15875


samdark:
Merged. Thank you!

framework/db/BaseActiveRecord.php
```diff
@@ -458,6 +458,12 @@ protected function createRelationQuery($class, $link, $multiple)
      */
     public function populateRelation($name, $records)
     {
+        foreach ($this->_relationsDependencies as $attribute => $relationNames) {
+            if (isset($relationNames[$name])) {
+                unset($this->_relationsDependencies[$attribute][$name]);
+            }
+        }
```
cebe: doesn't this revert the whole idea of what `_relationsDependencies` is doing?

cebe: the tests in https://github.com/yiisoft/yii2/pull/13618/files#diff-5c2b59d39a2ea1dc5443fb98d1325d11 do not cover eager loading. Pretty sure this gets broken by this change.

shirase: Eager loading can have bug. Needs to be tested.

shirase: relationsDependencies logic, do not works with eager loading initially, my pr do not change that behavior.

PR #16097, opened 2018-04-13
added missing line break before header

w3lifer:


samdark:
It is fine as is: https://www.yiiframework.com/doc/guide/2.0/en/caching-fragment#duration

w3lifer:
@samdark, yes, but it's not like the rest of the headers. I just parse the files (for my own purposes) and discovered it.

PR #16113, opened 2018-04-17
Update tutorial-yii-as-micro-framework.md

bodilych:


cebe:
Thank you!

bologer:
@cebe could you please also update parts like "сообщения with ID 1" to "сообщения с ID 1", "with" part is not translated.

cebe:
fixed in 7d20e11 thanks.

PR #16119, opened 2018-04-17
Fixed writing tags that dont close.

sharkbit:
Adjusted the closing of tags, and added /> to close tags that are Void Elements.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | didn't feel like making an issue very simple fix

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -147,8 +147,8 @@ class BaseHtml
         if ($name === null || $name === false) {
             return $content;
         }
-        $html = "<$name" . static::renderTagAttributes($options) . '>';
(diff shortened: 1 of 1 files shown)


Thoulah:
I vote strongy against this

sharkbit:
Well you give a great reason for such a strong disagreement.

SilverFire:
While the change is not bad itself, it will break a lot of tests in framework and in developer's code, so the necessity is disputable.

samdark:
It isn't necessary in HTML5. http://blog.teamtreehouse.com/to-close-or-not-to-close-tags-in-html5

PR #16124, opened 2018-04-18
Fix action name in commentaries (list -> index) [skip ci]

filipe1309:
The action name referenced in the commentary (list) was different from the action name used in the example (index)

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | -

yii2_httpcache_issue


samdark:
Merged. Thank you!

PR #16128, opened 2018-04-19
DI inject class to action params.

borodulin:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

It allows you to inject any class, which can be constructed with arguments known to Yii.
The idea is taken from Codeception.

### DIFF
--- a/framework/console/Controller.php
+++ b/framework/console/Controller.php
@@ -176,6 +176,8 @@ class Controller extends \yii\base\Controller
             if (!isset($args[$i])) {
                 if ($param->isDefaultValueAvailable()) {
                     $args[$i] = $param->getDefaultValue();
+                } elseif ($param->getClass()) {
+                    $args[$i] = Yii::$container->get($param->getClass()->name);
                 } else {
                     $missing[] = $param->getName();
                 }
--- a/framework/console/Controller.php
+++ b/framework/console/Controller.php
@@ -176,8 +176,8 @@ class Controller extends \yii\base\Controller
             if (!isset($args[$i])) {
                 if ($param->isDefaultValueAvailable()) {
                     $args[$i] = $param->getDefaultValue();
-                } elseif ($param->getClass()) {
-                    $args[$i] = Yii::$container->get($param->getClass()->name);
+                } elseif ($class = $param->getClass()) {
+                    $args[$i] = Yii::$container->get($class->name);
                 } else {
                     $missing[] = $param->getName();
                 }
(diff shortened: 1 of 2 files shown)


samdark:
It won't be merged into master but could be merged into 2.1. The feature itself, while definitely handy, is a bit controversial from correctness of DI standpoint.

samdark:
Let's do a little poll. Add 👍 if you like the feature to be merged and 👎 if not.

cebe:
This has been discussed in https://github.com/yiisoft/yii2/issues/9476#issuecomment-179749039 already.

borodulin:
@cebe When specifying a variable class, the only way to use such a variable is to inject a dependency through DI. In another case, the use of such an action with such variables will be impossible. This feature opens up the use of new development patterns and does not break anything old. If you do not like these patterns, do not use them. But why you decided everything for others - I do not understand.

SilverFire:
Constructor is a good place for dependencies because it highlights things that are required by this class and Yii already supports injection in controller constructor. Injection to methods makes dependencies less remarkable and blows actions definition. 

@borodulin why don't you like to inject in class?

borodulin:
case 1. Use of models as input parameters.
Now this approach is proposed:
```
  public function actionCreate()
  {
    $model = new MyModel();
    $model->load(...);
  }
```
This construction can be replaced by the following:
```
  public function actionCreate(MyModel $model)
  {
  }
```
case 2: Use of services as input parameters;
```
  public function actionProcess(MyService $myService)
  {
    $myService->doSomething();
  }
```
Of course I can use the constructor, but then I need to declare a private variable for the class. And if I only use one method, then it becomes inconvenient to add / remove services.
You can combine services and models, then it becomes simple:
```
  public function actionProcess(MyModel $model, MyService $service)
  {
    return $service->process($model);
  }
```

SilverFire:
> case 1. Use of models as input parameters.

This is a very special use case that can have different behaviors and require a lot of settings to: 

- Process GET, POST or other request methods
- Either extract existing model from DB or create new
- Handle errors
- Support tabular input (bulk operations)
- Manipulate scenarios

And the list, obviously, is not complete.


> case 2: Use of services as input parameters;

As I've said above, it should be done in constructor.

> And if I only use one method, then it becomes inconvenient to add / remove services

In case you have the only action right now, you will probably have more action later. Controllers rarely stay with a single action because they are designed to group a bunch of actions related by some criteria (model, operation, etc).

___

Thank you for your suggestions, but the topic has been already discussed in #9476 and I don't see new arguments to change the decision.

PR #16146, opened 2018-04-22
`\yii\helpers\BaseUrl` => `\yii\web\UrlManager`

lav45:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #16017


framework/helpers/BaseHtml.php
```diff
@@ -236,7 +236,7 @@ public static function cssFile($url, $options = [])
         if (!isset($options['rel'])) {
             $options['rel'] = 'stylesheet';
         }
-        $options['href'] = Url::to($url);
+        $options['href'] = Yii::$app->getUrlManager()->to($url);
```
SilverFire: While moving logic from helper to class is a good change, I think methods should be renamed because now they don't describe what are they doing indeed.

_"Url manager to $url"_ &ndash; whaat?

When it was `Url::to()` it was pretty clear what does the `to()` method does. Now it's not like that :) I don't have a good idea now and discussion is opened.

SilverFire (changes requested):

rob006:
This looks like regression to me. Previously we were able to use different `UrlManager` instance for most of the URLs-related operations, now this is hardcoded to `Yii::$app->getUrlManager()`.

framework/helpers/BaseHtml.php
Deele:
```
Yii::$app->getUrlManager()->urlTo($url)
Yii::$app->urlManager->urlTo($url)
$um = Yii::$app->urlManager;
$um->urlTo($url)
```

| Old | New |
|-----------------|-----|
| `Url::toRoute($route, false)` | `Yii::$app->urlManager->urlToRoute($route, false)`  |
| `Url::normalizeRoute($route)` | `Yii::$app->urlManager->normalizeRoute($route)`  |
| `Url::to('', false)` | `Yii::$app->urlManager->to('', false)`  |
| `Url::ensureScheme($url, $scheme)` | `Yii::$app->urlManager->ensureUrlScheme($url, $scheme)`  |
| `Url::base(false)` | `Yii::$app->urlManager->baseUrl(false)`  |
| `Url::remember('', null)` | `Yii::$app->urlManager->rememberUrl('', null)`  |
| `Url::previous(null)` | `Yii::$app->urlManager->previousUrl(null)`  |
| `Url::canonical()` | `Yii::$app->urlManager->canonicalUrl()`  |
| `Url::home(false)` | `Yii::$app->urlManager->homeUrl(false)`  |
| `Url::isRelative($url)` | `Yii::$app->urlManager->isRelativeUrl($url)`  |
| `Url::current([], false)` | `Yii::$app->urlManager->currentUrl([], false)`  |

How about using getter syntax for methods that are usually used without arguments?

| Old | New |
|-----------------|-----|
| `Url::base(false)` | `Yii::$app->urlManager->getBaseUrl(false)`  |
| `Url::base(false)` | `Yii::$app->urlManager->baseUrl`  |
| `Url::previous(null)` | `Yii::$app->urlManager->getPreviousUrl(null)`  |
| `Url::previous(null)` | `Yii::$app->urlManager->previousUrl`  |
| `Url::canonical()` | `Yii::$app->urlManager->getCanonicalUrl()`  |
| `Url::canonical()` | `Yii::$app->urlManager->canonicalUrl`  |
| `Url::home(false)` | `Yii::$app->urlManager->getHomeUrl(false)`  |
| `Url::home(false)` | `Yii::$app->urlManager->homeUrl`  |
| `Url::current([], false)` | `Yii::$app->urlManager->getCurrentUrl([], false)`  |
| `Url::current([], false)` | `Yii::$app->urlManager->currentUrl`  |

Deele:
What about retaining old helper, just adding url manager argument to allow backward compat?

Something like this:

```
protected static function getUrlManager($name = null)
{
    if (is_null($name)) {
        return static::$urlManager ?: Yii::$app->getUrlManager();
    }
    $urlManager = Yii::$app->{$name};
    if (!($urlManager instanceof UrlManager)) {
        throw new InvalidArgumentException('Invalid UrlManager');
    }
}

public static function to($url = '', $scheme = false, $urlManager = null)
{
    return static::getUrlManager($urlManager = null)->to($url, $scheme);
}
```

framework/helpers/BaseHtml.php
lav45: @SilverFire,  how about this renaming methods?

| old | new |
|---|---|
| `to()` | `createUrlTo()` |
| `toRoute()` | `createUrlToRoute()` |
| `base($scheme = false)` + `getBaseUrl()` | `getBaseUrl($scheme = false)` |
| `home($scheme = false)` + `getHomeUrl()` | `getHomeUrl($scheme = false)` |

rob006:
@lav45 But I'm no longer able to configure URL manager used by some widgets. Previously I was able to use `Url::$urlManager = Yii::$app->frontendUrlManager` and `ActionColumn` will generate URL using `Yii::$app->frontendUrlManager`. Now it is hardcoded to always use `Yii::$app->urlManager `.

samdark:
Yes. That was not consider and isn't good :(

lav45:
@rob006 We lead no one bothers to add а method `ActionColumn::setUrlManager()` and you will be able to override default urlManager

samdark:
@rob006 ?

rob006:
I like the idea of passing UrlManager instance to widgets/methods, but:

1. There is still many places where `Yii::$app->getUrlManager()` is hardcoded.
2. I think we should use `Instance::ensure()` to process `urlManager` configuration. Or event use tree traversal to inherit `urlManger` from application -> module -> controller -> widget.

lav45:
@SilverFire, what do you think about it?

rob006 (changes requested):
I'm against this at current state. As long as I would like to get rid of using global state in `Url::$urlManager`, this PR creates more issues than actually fixes. It creates much less convenient and readable syntax in many places (`Url::toRoute($params)` vs `$this->urlManager->createUrlToRoute($params)`) and it makes much harder to create widget/components with flexible URLs handling. Previously you can just use `Url` helper directly. If someone wants to use different URL manager, he could always set correct URL manager as `Url::$urlManager` before calling my widget:

```php
// in backend
Url::$urlManager = Yii::$app->frontedUrlManager
// use some frontend stuff (like generate 3rd-party widget) which will use frontend UrlManager
$content = SomeWidget::widget();
Url::$urlManager = Yii::$app->urlManager
// backend UrlManager will be used again at this point
```

Right now it is impossible - I need to provide property for URL manager in my widgets/methods, use inconvenient syntax for generating URLs and propagate URL manager settings to all children widgets (assuming that they're supporting this). I'm pretty sure that almost nobody will do this and most of the extensions will use hardcoded `Yii::$app->urlManager`, mostly by using `Url` helper like before (which now has hardcoded `Yii::$app->urlManager` as a URL manager) or any other helper (which also uses hardcoded `Yii::$app->urlManager`). This is regression for #10610 and probably big step back.

Also I don't like moving helper logic to `UrlManager`.  It makes this class much heavier (920 LOC vs old 640 LOC) and I don't think that `UrlManager` should be responsible for normalization of different formats before actual URL generation. URL generation is already complicated, we should keep `UrlManager` and `UrlRule` classes as small as possible.

Deele:
@rob006 What about my [proposal](https://github.com/yiisoft/yii2/pull/16146#issuecomment-388352767) earlier? I already told that this would break backwards compatability and by introducing `name` argument to those function, we would solve those issues in verbose way.

`Url::to('site/index', false, 'frontedUrlManager')`

Specifically for widgets, I think that changes to URL generation should be done by widget maintainer, by implementing widget class attribute or constant that allows to specify which url manager to use and implement use of it in every place within widget.

PS: I see this issue ideologically related to `Yii::t` where there is no consistent way for widget maintainers to create categories of messages, thus it is hard to manage which translation component configuration for it to use. Some widget maintainers fix it with widget class attribute, constant or method, that returns which i18n or which default i18n component configuration to use.

rob006:
@Deele I don't think it will make current situation noticeably better - you still would need to specify URL manager explicitly every time and provide some settings in your widgets. 

We would need something like https://github.com/yiisoft/yii2/pull/15342 to achieve this in convenient way - static helpers could use current widget/controller/module to get URL manager (or any other component) by traversing parents. So if in action `/module1/module2/my-controller/some-action` I will use `GridView` with `ActionColumn`, `Url::to()` will search URL manager in this order:
1. `ActionColumn::get(UrlManager::class)`
2. `GridView::get(UrlManager::class)`
3. `MyControllerController::get(UrlManager::class)`
4. `Module2Module::get(UrlManager::class)`
5. `Module1Module::get(UrlManager::class)`
6. `Yii::$app->get(UrlManager::class)`

You could configure URL manager at any level, for example in controller - then all content generated in this controller should use configured URL manager.

PR #16156, opened 2018-04-24
fix code completion for `UploadedFile::getInstance()` - may return null

k-timoshenko:
Method `UploadedFile::getInstance()` may return `null` but it doesn't mentioned in PHPDoc correctly.
Changes are the same with `UploadedFile::getInstanceByName()`

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Fixed issues  | more strict return type declaration


samdark:
Merged. Thanks!

PR #16161, opened 2018-04-25
Minor refactor

243083df:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -


samdark:
Doesn't seem to be obvious change to me. Would you please elaborate a bit?

rob006:
@samdark You don't need to check if value is set if you want to unset it - unsetting non-existent element will not give any warning, so you're avoiding unnecessary `if` in this way.

rob006 (approved):

samdark:
Merged. Thank you!

PR #16162, opened 2018-04-25
Added `Yii::getEnv`

leandrogehlen:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Helper to retrieve values from enviroment variables

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -95,6 +95,40 @@ class BaseYii
     }
 
     /**
+    * Returns the value of an environment variable.
+    * @param string $key the var name
+    * @param mixed $default the default result value
+    * @return mixed
+    */
+   public static function getEnv($key, $default = null)
+   {
+       $value = getenv($key);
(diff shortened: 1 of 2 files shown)


samdark:
Is it better than any existing dotenv libraries?

leandrogehlen:
The problem is get the value from environment variables.
These libraries just load the environmnet variables from file

samdark:
Nope. These usually do both i.e. environment variables + files with files priority.

leandrogehlen:
I guess that https://github.com/vlucas/phpdotenv is the most popular and used in laravel.
And i didn't find this feature.

Laravel has implemented the same feature
https://github.com/laravel/framework/blob/5.6/src/Illuminate/Support/helpers.php#L604

samdark:
https://github.com/yiisoft/yii2-app/issues/1

PR #16168, opened 2018-04-26
initialize the property defination as array

lichunqiang:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

### DIFF
--- a/framework/caching/ArrayCache.php
+++ b/framework/caching/ArrayCache.php
@@ -24,7 +24,7 @@ namespace yii\caching;
  */
 class ArrayCache extends Cache
 {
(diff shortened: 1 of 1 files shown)


lichunqiang:
If we do not initialize with the correct type, the `$_cache` is `null`

So, there is some reasons:

1. there will be have a typecast
2. handling `null` as array is allowed by language, but not the principle


Just a suggestion. :grinning:

rob006 (approved):

PR #16185, opened 2018-04-28
fix #16184 use self instead of static

### LINKED ISSUE #16184: static::$stack misused in yii\base\Widget of begin() and end functions
When re-declare the public static property  `$stack` in the sub class of `yii\base\Widget`,
the exception of 
```php
            throw new InvalidCallException('Expecting end() of ' . get_class($widget) . ', found ' . get_called_class());
```
would not trigger correctly due to `static` keyword before `$stack`.
This happens when two widgets begin and end not correctly.
e.g.
Both WidgetA and WidgetB are instance of yii\base\widget.
ether WidgetA or WidgetB redeclared the $stack;

```php
<?php WidgetA::begin()?>
<?php WidgetB::begin()?>
<?php WidgetA::end()?>
<?php WidgetB::end()?>
```
Even though the public static property $stack redeclared in the widget,
the exception could also be thrown.

### What do you get instead?
when $stack is pushed and popped, use self::$stack instead of static$stack

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15
| PHP version      | 7.1
| Operating system | CentOs

yanggs07:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 16184

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -13,7 +13,7 @@ Yii Framework 2 Change Log
 - Bug #16073: Fixed regression in Oracle `IN` condition builder for more than 1000 items (cebe)
 - Bug #16120: FileCache: rebuild cache file before touch when different file owner (Slamdunk)
 - Bug #16091: Make `yii\test\InitDbFixture` work with non-SQL DBMS (cebe)
-- Bug #16184: fix `yii\base\Widget` $stack use `self` scope instead of `static`(yanggs07)
+- Bug #16184: Fixed `yii\base\Widget` to access `stack` property with `self` instead of `static` (yanggs07)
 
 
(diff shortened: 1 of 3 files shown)


SilverFire:
Could you add a unit test to check the original problem reported in #16184?

yanggs07:
@SilverFire thx for replying my pr.

it would take me some time to learn how to write test case for this.

I would be as soon as possible.

yanggs07:
@SilverFire unit test added.

SilverFire (approved):

samdark:
Merged. Thank you!

PR #16191, opened 2018-04-29
Fixed `yii\helpers\Inflector::camel2words()` to work with UTF-8

SilverFire:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | this one

### DIFF
--- a/framework/helpers/BaseInflector.php
+++ b/framework/helpers/BaseInflector.php
@@ -369,13 +369,13 @@ class BaseInflector
      */
     public static function camel2words($name, $ucwords = true)
     {
-        $label = strtolower(trim(str_replace([
+        $label = mb_strtolower(trim(str_replace([
             '-',
             '_',
             '.',
-        ], ' ', preg_replace('/(?<![A-Z])[A-Z]/', ' \0', $name))));
+        ], ' ', preg_replace('/(\p{Lu})/u', ' \0', $name))));
 
-        return $ucwords ? ucwords($label) : $label;
+        return $ucwords ? mb_convert_case($label, MB_CASE_TITLE) : $label;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


samdark (approved):

SilverFire:
There are many uses of non-unicode-safe functions (strtolower, ucwords, ucfirst) in inflector.
I think it makes sense to make them consistently unicode safe.

framework/helpers/BaseInflector.php
```diff
@@ -369,13 +369,13 @@ public static function camelize($word)
...
-        $label = strtolower(trim(str_replace([
+        $label = mb_strtolower(trim(str_replace([
             '-',
             '_',
             '.',
-        ], ' ', preg_replace('/(?<![A-Z])[A-Z]/', ' \0', $name))));
+        ], ' ', preg_replace('/(\p{Lu})/u', ' \0', $name))));
 
-        return $ucwords ? ucwords($label) : $label;
+        return $ucwords ? mb_convert_case($label, MB_CASE_TITLE, 'UTF-8') : $label;
```
samdark: Yii::$app->charset?

SilverFire:
Updated PR

framework/helpers/BaseInflector.php
```diff
@@ -599,4 +600,27 @@ public static function sentence(array $words, $twoWordsConnector = null, $lastWo
...
+    }
+
+    /**
+     * The same as built-in `ucfirst`, but unicode-safe
+     *
+     * @param $string
+     * @param $encoding
+     * @return string
+     */
+    private static function mb_ucfirst($string, $encoding)
```
samdark: Could be renamed to `ucfirst`, made public and moved to `StringHelper`.

samdark: But overall I'm OK leaving it private. Parameter phpdoc types should be specified.

PR #16197, opened 2018-04-29
Fix #16068: wrong return value in `yii\web\CookieCollection::has`

### LINKED ISSUE #16068: Cookie method has returns wrong value when using in response
Add cookie

```
Yii::$app->response->cookies->add(new Cookie([
                'name' => 'test',
                'value' => 1,
            ]));
```

When checking if the CookieCollection has the current cookie, it should return true.
`$value = Yii::$app->response->cookies->has('test')`

### What do you get instead?
`$value` is actually false.

I suppose its because in the Yii code `expire` is checked and the cookie has `expire` 0 in my case. But it's a valid value, it means 'until the browser is closed'.

```
return isset($this->_cookies[$name]) && $this->_cookies[$name]->value !== ''
    && ($this->_cookies[$name]->expire === null || $this->_cookies[$name]->expire >= time());
```

### Suggested solution
I’d updated the code like this, but I'm really not sure if it's enough and won't break anything.
```
return isset($this->_cookies[$name]) && $this->_cookies[$name]->value !== ''
    && ($this->_cookies[$name]->expire === null || $this->_cookies[$name]->expire === 0 || $this->_cookies[$name]->expire >= time());
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.1.4
| Operating system | Win10

OndrejVasicek:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  |no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #16068 

Please check the typos. Thanks.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #16068: Fixed `yii\web\CookieCollection::has` when an expiration param is set to 'until the browser is closed' (OndrejVasicek)
 - Bug #16006: Handle case when `X-Forwarded-Host` header have multiple hosts separated with a comma (pgaultier)
 - Bug #16010: Fixed `yii\filters\ContentNegotiator` behavior when GET parameters contain an array (rugabarbo)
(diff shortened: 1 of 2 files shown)


SilverFire (approved):

samdark (approved):

samdark:
Merged. Thank you!

PR #16199, opened 2018-05-01
doc: added some extra information for postgresql users.

cristobalvega:
It looks like if a user of PostgreSQL have his schema on public, Yii doesn't recognize his db. Maybe this will help to other Yii2 users
> Note: If you are using PostgreSQL, make sure you have the schema on public.


SilverFire:
Hello. Thank you for the suggestion, but I neither understand the described problem, nor understand the outcome of the note you've added.
Maybe, you could supply an example?

samdark:
It's possible to use schema different from public.

PR #16243, opened 2018-05-07
Fix #16183 when `yii\helpers\BaseFileHelper` sometimes returned wrong value

### LINKED ISSUE #16183: BaseFileHelper::unlink() returns wrong value
Use BaseFileHelper::unlink(), pass non-existent file.

Return false.

### What do you get instead?
True.

### Additional info
Here is the simple code

```php
    public static function unlink($path)
    {
        $isWindows = DIRECTORY_SEPARATOR === '\\';

        if (!$isWindows) {
            return unlink($path);
        }

        if (is_link($path) && is_dir($path)) {
            return rmdir($path);
        }

        try {
            return unlink($path);
        } catch (ErrorException $e) {
            // last resort measure for Windows
            $lines = [];
            exec('DEL /F/Q ' . escapeshellarg($path), $lines, $deleteError);
            return $deleteError !== 0;
        }
    }
```

I'm on Win 10, so I pass the first condition. I also pass the second one and skip into the try section.
In the catch part, the $deleteError has value 1, which I suppose is an error.
And i think the last line is a bug. It returns true whenever the value is different than 0.
So my file isn't deleted and still the whole method returns true.

It should be like this
`return $deleteError == 0;`

Or my understanding of $deleteError is wrong, but then again - why it returns true when no file was deleted?

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.?
| PHP version      | 7.1.4
| Operating system | Win10

OndrejVasicek:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | 
| Fixed issues  | #16183


samdark (approved):

samdark:
Merged. Thanks!

PR #16263, opened 2018-05-11
Fix IE11 compatibility mode redirect

Alex-Bond:
In IE11 when you turn on compatibility mode Yii stops redirecting the user to login page.
Problem is that accepted types contains ```*/*``` but is positioned as the last type, not first.
Technically there is 2 option to fix that:
1. Proposed change when we just adding ```*/*``` to acceptableRedirectTypes and system will try to find it at any position
2. We change L758 and move the ```*/*``` check to separate loop where Yii will check ```*/*``` at any position in the array.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Fixed issues  | none

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -147,7 +147,7 @@ class User extends Component
      * @var array MIME types for which this component should redirect to the [[loginUrl]].
      * @since 2.0.8
      */
-    public $acceptableRedirectTypes = ['text/html', 'application/xhtml+xml'];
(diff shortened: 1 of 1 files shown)


samdark:
https://travis-ci.org/yiisoft/yii2/jobs/377539542

samdark:
@Alex-Bond any plans to finish the pull request? Currently it fails tests.

Alex-Bond:
Oh... I will try to look at it by the end of the week.

PR #16270, opened 2018-05-13
Use Null coalesce operator where possible

SilverFire:
Related to #11397


cs/src/YiiConfig.php
```diff
@@ -149,6 +149,7 @@ public function __construct($name = 'yii-cs-config')
             'standardize_not_equals' => true,
             'ternary_operator_spaces' => true,
             'trailing_comma_in_multiline_array' => true,
+            'ternary_to_null_coalescing' => true,
```
Thoulah: For alphabetic order, please move this one line up.

SilverFire:
Merged in 1cc5f156359403a3877acb7055830be2431f9011

PR #16284, opened 2018-05-15
Fix an error when the controller outputs the _file_ variable to the template

suyar:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    |no
| Tests pass?   | no
| description  | When my controller calls $this->render('view', ['_file_'=>'any']);   `$this->render('view', ['_file_'=>'any']);` time error


samdark:
The change doesn't make sense to me :(

PR #16290, opened 2018-05-16
Switch to "config.asset-installer-paths" option as "extra" is depreca…

yuniorsk:
…ted + fixed headers in docs


samdark:
Merged. Thank you!

Issue #16252, opened 2018-05-09
DynamicModel should support check dynamic property.


vuongxuongminh:
### What steps will reproduce the problem?

Create an object of DynamicModel with dynamic attributes. Need to check attribute exist because it is `dynamic`, i use method `hasProperty` to check but currently it is not support.

### What is the expected result?

`hasProperty`, `canGetProperty`, `canSetProperty` methods should be return `TRUE` when check.

### What do you get instead?

`hasProperty`, `canGetProperty`, `canSetProperty` methods return `FALSE` when check.

### Additional info

Should we override  `canGetProperty`, `canSetProperty` to check an attribute dynamic exist?

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15
| PHP version      |  7.1
| Operating system | Windows 10

samdark:
Yes. It sounds like a bug.

PR #16326, opened 2018-05-28
Fixed #16104 bug: use schema name from table path in PostgreSQL QueryBuilder::dropIndex()

### LINKED ISSUE #16104: Incorrect SQL query generated by dropIndex in migration

1) In database PortgreSQL create schema myproject
2) In schema create table mytable
3) In table create fields id:PKint and category_id:int
4) create index on field category_id with name myproject_categoryid_index

5) Write migration with next code

`$this->dropIndex(‘myproject_categoryid_index’, 'myproject.mytable');`

6) run php yii migrate

must be generated SQL with schema
so it must be like this
`DROP INDEX myproject.myproject_categoryid_index RESTRICT`

and migration executed succesful

### What do you get instead?

`$this->dropIndex(‘myproject_categoryid_index’, 'myproject.mytable');`
this line of code generate SQL:
`DROP INDEX myproject_categoryid_index`

without schema!

and when executed get error:
` > drop index myproject_categoryid_index on myproject.mytable ...Exception: SQLSTATE[42704]: Undefined object: 7 ОШИБКА:  индекс "myproject_categoryid_index" не существует`

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1?
| PHP version      | 7.2
| Operating system | Debian

wapmorgan:
Fixed PostgreSQL `QueryBuilder::dropIndex()` if index is used in non-default schema, e.g:
There's an index `idx_news_id` in table `site.news`.

## Current behavior

Now in **master**:
```php
News::getDb()->getQueryBuilder()->dropIndex('idx_news_id', 'site.news')
```
will generate
```sql
DROP INDEX "idx_news_id"
```
which fails.

## After applying this PR
The same code will generate right query:
```sql
DROP INDEX "site"."idx_news_id"
```

**Note**: if schema is already defined in index name (e.g. **site.idx_news_id**), then no modifications will be performed.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #16104


wapmorgan:
Need to fix usage with `{{%...}}`..

wapmorgan:
I think to add this logic:
- If table is `schema.table`, then just add `schema` part to an index.
- If table is `{{schema.table}}`, then strip off curly braces and add `schema` part to an index.
- If table is `{{schema.%table}}`, then do the same.
- If table is `{{%schema.table}}`, which is really crazy, it will modfy index `idx` to `{{%schema.idx}}`.
- If table is `{{%table}}` (and `tablePrefix` contains schema definition, which unwraps it to `schema.table`), then modify index `idx` like `{{%idx}}`, which should lead to `schema.idx`. _Does anyone use it this way? Anyway, It will work_

Is there any other cases?

wapmorgan:
Build fails due to line in changelog from #16252: it has trailing space at the end of line.

samdark:
That's fine. Will fix that separately.

samdark:
Merged. Thank you!

Issue #16327, opened 2018-05-28
getComposer() yii\BaseYii::createObject(null) BaseMailer.php


cjtterabyte:
Yii Version: 2.1
PHP Version: 7.1.17
Operating system: Centos 7.1.

```
Problem Compose Function:

    ->compose(
        ['html' => 'passwordResetToken-html', 'text' => 'passwordResetToken-text'],
        ['user' => $user]
    )

Error Function getComposer():

    public function getComposer()
    {
        if (!is_object($this->_composer) || $this->_composer instanceof \Closure) {
            if (is_array($this->_composer) && !isset($this->_composer['__class'])) {
                $this->_composer['__class'] = Composer::class;
            }
            $this->_composer = Yii::createObject($this->_composer); ---> ERROR = OBJECT NULL.
        }
        return $this->_composer;
    }

   Error Yii2.1.  yii\BaseYii::createObject(null)
```

https://github.com/yiisoft/yii2-swiftmailer/issues/65

cjtterabyte:
Hi @samdark After checking and debugging to understand the new philosophy BaseMailer->Composer-> Template, I discovered the silly mistake, here is the same.

```
Error Code:

    public function getComposer()
    {
        if (!is_object($this->_composer) || $this->_composer instanceof \Closure) {
            if (is_array($this->_composer) && !isset($this->_composer['__class'])) { --->ERROR CODE
                $this->_composer['__class'] = Composer::class;
            }
            $this->_composer = Yii::createObject($this->_composer);
        }
        return $this->_composer;
    }

Fix Code:

    public function getComposer()
    {
        if (!is_object($this->_composer) || $this->_composer instanceof \Closure) {
            if (!is_array($this->_composer) && !isset($this->_composer['__class'])) { --->FIX CODE
                $this->_composer['__class'] = Composer::class;
            }
            $this->_composer = Yii::createObject($this->_composer);
        }
        return $this->_composer;
    }
```

PR #16344, opened 2018-05-31
Fix getComposer() yii\BaseYii::createObject(null) BaseMailer.php #16327

tecnoterabyte:
Fixed (#16327):  getComposer() yii\BaseYii::createObject(null) BaseMailer.php #16327

### DIFF
--- a/framework/mail/BaseMailer.php
+++ b/framework/mail/BaseMailer.php
@@ -94,7 +94,7 @@ abstract class BaseMailer extends Component implements MailerInterface
     public function getComposer()
     {
         if (!is_object($this->_composer) || $this->_composer instanceof \Closure) {
-            if (is_array($this->_composer) && !isset($this->_composer['__class'])) {
+            if (!is_array($this->_composer) && !isset($this->_composer['__class'])) {
                 $this->_composer['__class'] = Composer::class;
             }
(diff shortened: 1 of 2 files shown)


framework/mail/BaseMailer.php
```diff
@@ -94,7 +94,7 @@
     public function getComposer()
     {
         if (!is_object($this->_composer) || $this->_composer instanceof \Closure) {
-            if (is_array($this->_composer) && !isset($this->_composer['__class'])) {
+            if (!is_array($this->_composer) && !isset($this->_composer['__class'])) {
```
rob006: Can you explain it a little more? This check does not make any sense right now - if `$this->_composer` is not an array, then `$this->_composer['__class']` will never exist. And it will never set default class for array without `__class` key, which seems to be the whole point of this piece of code.

cjtterabyte: Hi @rob006 you are absolutely right, reviewing all the code and testing, it is expected that $this->_ composer is an array, and you only have to assign the class in case it does not exist, then we just initialize the variable _composer as an array.

samdark (approved):

samdark:
Merged. Thanks.

PR #16357, opened 2018-06-03
Fixes #16192: Extracted `getCacheKey()` protected method from `queryInternal()` protected method

### LINKED ISSUE #16192: The QueryInternal method in \yii\db\Command
Hello.
I have a two questions related to the `queryInternal` method in `\yii\db\Command`.
1) Is there any reason for `logQuery` to be `private`? This is the only private method in the whole class. We need to redefine it for some testing purposes and in this case we must copy whole `queryInternal` method just to be able to override `logQuery`.
2) Can the cache key generation be extracted to a separate protected method from the `queryInternal`? We have some key in the persistent cache that is changed when any migrations is applied. We need to copy whole `queryInternal` method once again to be able to add that key to the cache key generation.

drlibra:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16192


samdark:
Merged. Thanks!

PR #16361, opened 2018-06-04
Ability to customize database query if condition is true

mg-code:
Sometimes it is necessary to customize query if some condition is true.

Now you can do this by using this function:
```
$model = Product::find()
    ->when(\Yii::$app->user->identity, function (Query $query, User $user) {
        $query->orderBy($user->getProductOrder());
    }
    ->all();
```
You can pass additional parameters:
```
$model = Product::find()
    ->when(\Yii::$app->user->identity, function (Query $query, User $user) use ($defaultOrder) {
        if ($orderBy = $user->getProductOrder()) {
            $query->orderBy($orderBy);
        } else {
            $query->orderBy($defaultOrder);
       }
    }
    ->all();
```

You can pass callback that will be applied if value is false:
```
$user = Yii::$app->user->identity;
$model = Product::find()
    ->when($user ? $user->getProductOrder() : false, function (Query $query, $productOrder) {
          $query->orderBy($productOrder);
    }, function (Query $query) use ($defaultOrder) {
          $query->orderBy($defaultOrder);
    })
    ->all();
```

What is your opinion on this? 
If everything is ok, I will create tests and changelog entry.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no


mg-code:
@samdark @cebe what are your thoughts on this?

rob006:
I would be easier and more practical to introduce something like `public function touch(Closure $callback)` which just run `$callback` and returns query. Every required condition would be inside of `$callback` so it should be more flexible (you can use multiple `elseif`) and readable (you don't introduce another syntax for `if` statement).

mg-code:
We can implement both. when can use touch function.

Using `when` code is much compact. E.g.:
```
Item::find()
    ->when(Yii::$app->request->get('maxPrice'), function ($query, $maxPrice) {
        $query->andWhere('price <= :maxPrice', [':maxPrice' => (int) $maxPrice]);
    })
```
vs
```
Item::find()
    ->touch(function ($query) {
        if ($maxPrice = Yii::$app->request->get('maxPrice')) {
            $query->andWhere('price <= :maxPrice', [':maxPrice' => (int) $maxPrice]);
        }
    })
```

rob006:
> Using when code is much compact. E.g.:

And much less readable. If you want to compact and unreadable syntax, you can always use ugly oneliner:

```php
Item::find()
    ->touch(function ($query) { if ($maxPrice = Yii::$app->request->get('maxPrice')) { $query->andWhere('price <= :maxPrice', [':maxPrice' => (int) $maxPrice]); } })
```

I'm against reinventing syntax for `if` and creating 2 different ways to do the same thing.

mg-code:
I really don't see any reason why it is less readable. Can you describe why?
Then we can just continue write old style without any callbacks.
```
$query = Item::find();
if ($maxPrice = Yii::$app->request->get('maxPrice')) {
    $query->andWhere('price <= :maxPrice', [':maxPrice' => (int) $maxPrice]);
}
$result = $query->all();
```

Laravel has such feature for a long time. I did not hear any complains from Laravel developers.
https://laravel.com/docs/5.6/queries#conditional-clauses

P.s. Laravel solution works a little bit different. In my solution you can't override return value of when.

rob006:
> I really don't see any reason why it is less readable. Can you describe why?

Because I need to learn a new overcomplicated syntax for `if`. Understanding this is a nightmare:

```php
$model = Product::find()
    ->when($user ? $user->getProductOrder() : false, function (Query $query, $productOrder) {
          $query->orderBy($productOrder);
    }, function (Query $query) use ($defaultOrder) {
          $query->orderBy($defaultOrder);
    })
    ->all();
```

Compare to:

```php
$model = Product::find()
	->when(function (Query $query) use ($user, $defaultOrder) {
		if ($user) {
			$query->orderBy($user->getProductOrder());
		} else {
			$query->orderBy($defaultOrder);
		}
	})
	->all();
```

Or more compact:

```php
$model = Product::find()
	->when(function (Query $query) use ($user, $defaultOrder) {
		$user ? $query->orderBy($user->getProductOrder()) : $query->orderBy($defaultOrder);
	})
	->all();
```
___ 

Honestly, I would rather prefer some `pipe()` method on higher level (`Component`?) to simplify chaining on any object:

```php
[
	'newModel' => (new MyModel())
		->pipe(function (MyModel $model) {
			$model->name = 'rob006';
			$model->generateSlug();
			return $model;
		}),
	'modelId' => (new MyModel())
		->pipe(function (MyModel $model) {
			$model->name = 'rob006';
			$model->generateSlug();
			$model->save()
			return $model;
		})
		->id,
];
```

mg-code:
I agree that my first examples were too complicated and there is no point from them if we have `touch` function.

We could remove "else" callback from `when` function and use this function only for such cases:
```
->when(Yii::$app->request->get('maxPrice'), function ($query, $maxPrice) {
     $query->andWhere('price <= :maxPrice', [':maxPrice' => (int) $maxPrice]);
})
```

For everything else we could use `touch` function as you described. It will be more clean solution.
`pipe` would be nice replacement for `touch` solution.

What do you think on this?

samdark:
1. Looks quite similar to how it is now:

```php
$query = Product::find();
$user = \Yii::$app->user->identity;
if ($user) {
    $query->orderBy($user->getProductOrder());
}
$models = $query->all();
```


```php
$query = Product::find();
$user = \Yii::$app->user->identity;
if ($user) {
    if ($orderBy = $user->getProductOrder()) {
            $query->orderBy($orderBy);
        } else {
            $query->orderBy($defaultOrder);
       }
}
$models = $query->all();
```

```php
$user = Yii::$app->user->identity;
$query = Product::find()->orderBy($defaultOrder);

$productOrder = $user ? $user->getProductOrder() : false;
if ($productOrder) {
    $query->orderBy($productOrder);
}

$models = $query->all();
```

2. You variant has a side effect of both checking condition and passing a variable to function. That, while it looks cool, would certainly cause issues.

Considering two points above I prefer not to have such syntax in the framework.

samdark:
If you still want it you can add it to your models in a base custom query.

rob006:
@samdark 
> Looks quite similar to how it is now:

Right now you need to create temporary variables and in many cases it is annoying and breaks flow. `Query` uses fluent setters everywhere - I'm pretty sure that there is a reason for that. The same reason applies to this FR.

> You variant has a side effect of both checking condition and passing a variable to function.

What side effects?

samdark:
I've meant that often the check and parameter you pass are totally different. For example, if "if the type of the order is corporate, order by corporation ID".

samdark:
Also in many cases it's `if` and `else` and there's no good fluent syntax for that.

rob006:
> Also in many cases it's if and else and there's no good fluent syntax for that.

Yes, it is, I proposed it [here](https://github.com/yiisoft/yii2/pull/16361#issuecomment-394968584) with examples [here](https://github.com/yiisoft/yii2/pull/16361#issuecomment-394998506).

I've already implemented something similar in my Yii 1.1 project for `CDbCriteria` and `CActiveRecord` and it works great and it is very convenient. I'm not sure what i s the problem here.

samdark:
Yes, it your variant it looks good.

mg-code:
@samdark I can define this function inside model query class, but what about DAO?

@rob006 were right, that assigning a variable sometimes is very annoying.
With which solution of those 3 we will implement?
1. `touch` inside QueryTrait?
2. `touch` and `when` (with parameter validation) inside QueryTrait?
3. `pipe` function inside high order component?

Imho, if we do not implement `when` solution, then `pipe` would be better choice.

samdark:
@mg-code it's good to have it in query builder as well.

cebe:
> Right now you need to create temporary variables and in many cases it is annoying and breaks flow. Query uses fluent setters everywhere - I'm pretty sure that there is a reason for that. The same reason applies to this FR.

therefor the code is more readable. Imo hiding changes to the object behind function calls that check the condition and make changes inside (not visible from the code you look at) introduces a lot of unnecessary complexity.

Fluent interface is for simple queryies so you can write select()->where()->all(), if your query gets more complex it is important that you can see on which condition which part is added. hiding that in a function call does not make sense to me.

mg-code:
I am for @rob006 solution. I already implemented such solution in trait: https://github.com/mg-code/yii2-helpers/blob/master/ActiveQueryHelperTrait.php#L21

@cebe IMHO with such solution code is more readable.


```
$model = Order::find()
    ->touch(function(OrderQuery $query)) {
        if ($someCondition) {
            ....
        }
    })
    ->mine()
    ->one();
```
than

```
$query = Order::find();
if ($someCondition) {
    $query->....
}
$model = $query
    ->mine()
    ->one();
```

samdark:
Won't make it into 2.0.

Issue #14289, opened 2017-06-13
resetSequence fail with an oci database


vinpel:
### What steps will reproduce the problem?
In a codeception unit test :
```php
  public function _before()
  {
  $this->tester->haveFixtures([
      'table_track' => [
        'class' => TableTrackFixture::className(),
      ],
    ]);
  }
```
### What is the expected result?
The goal is to clean table before doing my test.
The code work perfectly fine in a mySQL database.

### What do you get instead?
```bash
1) BDDTrackImportTest: My big test
 Test  tests/unit/BDDTrackImportTest.php:myBigTest
                                                                                                                                                                                                               
  [yii\db\Exception] SQLSTATE[HY000]: General error: 911 OCIStmtExecute: ORA-00911: caract▒re non valide
 (/tmp/pear/download/PDO_OCI-1.0/oci_statement.c:142)
The SQL being executed was: DROP SEQUENCE "table_track_SEQ";CREATE SEQUENCE "table_track_SEQ" START WITH 1 INCREMENT BY 1 NOMAXVALUE NOCACHE
                                                                                                                                                                                                               
/385e665a-0314-47e9-a3d8-90c6fe83712a/vendor/yiisoft/yii2/db/Command.php:856
/385e665a-0314-47e9-a3d8-90c6fe83712a/vendor/yiisoft/yii2/test/ActiveFixture.php:118
/385e665a-0314-47e9-a3d8-90c6fe83712a/vendor/yiisoft/yii2/test/ActiveFixture.php:78
/385e665a-0314-47e9-a3d8-90c6fe83712a/tests/fixtures/TableTrack.php:18
/385e665a-0314-47e9-a3d8-90c6fe83712a/vendor/yiisoft/yii2/test/FixtureTrait.php:95
/385e665a-0314-47e9-a3d8-90c6fe83712a/vendor/codeception/base/src/Codeception/Module/Yii2.php:301
/385e665a-0314-47e9-a3d8-90c6fe83712a/vendor/codeception/base/src/Codeception/Step.php:260
/385e665a-0314-47e9-a3d8-90c6fe83712a/vendor/codeception/base/src/Codeception/Scenario.php:72
/385e665a-0314-47e9-a3d8-90c6fe83712a/tests/_support/_generated/UnitTesterActions.php:41
/385e665a-0314-47e9-a3d8-90c6fe83712a/tests/unit/components/BDDTrackImportTest.php:90
[...]

PR #16362, opened 2018-06-04
Fix define command::executeResetSequence() to work with Oracle

CedricYii:
Introduce command::executeResetSequence() to handle OCI need for several
queries to reset a sequence.
Preserve backward compatibility with other databases.
Check if resetSequence() should be written as deprecated.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes for travis
| Fixed issues  | #14289

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Enh #14289: Added `yii\db\Command::executeResetSequence()` to work with Oracle (CedricYii)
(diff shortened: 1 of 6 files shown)


sergeymakinen (approved):
LGTM. `@since` should be added.

samdark:
@sergeymakinen should there be tests?

sergeymakinen:
@samdark Oh. I have noticed there's an existing test but forgot it should be altered. Good catch!

I will run tests as soon as they are ready.

sergeymakinen (changes requested):
Need updated tests for Oracle.

samdark:
Merged. Thank you!

PR #16367, opened 2018-06-05
Fixes #16364: integrity constraint violation when saving empty string

### LINKED ISSUE #16364: yii\db\ColumnSchema::typecast() converts empty string to null even if column doesn't allow null

1. Create a MySQL database table with a `text` column.
2. Run this:

   ```php
   Yii::$app->db->createCommand()->insert('table_name', [
       'column_name' => '' // empty string
   ])->execute();
   ```

The row should be inserted.

### What do you get instead?

A yii\db\IntegrityException exception is thrown because the empty string [gets converted to `null`](https://github.com/yiisoft/yii2/blob/01636ae45a51b3b5d4c2796b05c16e27678adb36/framework/db/ColumnSchema.php#L128) by yii\db\ColumnSchema::typecast(), despite the ColumnSchema’s `$allowNull` property being set to `false`.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1

brandonkelly:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16364


samdark:
Test results are interesting...

PR #16368, opened 2018-06-06
kB instead of KB

diegotibi:
When kilobyte is referred to 1000 bytes the abbreviation should be written with lower case "k" as kB. Upper case K is usually intended to refer to 1024 multiplier
See this page: https://en.wikipedia.org/wiki/Kilobyte

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none


Thoulah:
They already have the difference between kB en KiB. You change is in the else part of `if ($this->sizeFormatBase == 1024)`

Thoulah:
I think you should additionally edit all translation files as well, or this change will break other languages.

diegotibi:
Yes, I know. I've just proposed to change **K**B in **k**B. KiB (Kibibyte, 2^10 or 1024 prefix) is written correctly.

Thoulah:
And documentation on line https://github.com/diegotibi/yii2/blob/cd74431f24383db1767fc855d46f9f0377718580/framework/i18n/Formatter.php#L1308

ghost (approved):

samdark:
Please add a line for changelog. The change itself looks OK.

samdark (approved):

samdark:
Merged. Thanks!

Issue #15953, opened 2018-03-23
use component::off method bug


Choate:
### What steps will reproduce the problem?
```php
$component = new Component();
$component->off(ActiveRecord::EVENT_AFTER_INSERT, $handler);
```
### What is the expected result?
remote event
### What do you get instead?
PHP Warning – yii\base\ErrorException
Invalid argument supplied for foreach()

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14
| PHP version      | 7.1.11
| Operating system | Windows 7

Fixed

```php
$removed = false;
foreach ($this->_eventWildcards[$name] as $i => $event)
```
to

```php
$removed = false;
if (isset($this->_eventWildcards[$name])) {
  $removed = false;
  foreach ($this->_eventWildcards[$name] as $i => $event)
}
```

rugabarbo:
This case is covered by tests: https://github.com/yiisoft/yii2/blob/master/tests/framework/base/ComponentTest.php#L181-L200

Which file and line of code caused this warning?

Choate:
@rugabarbo 

I need to trigger an AfterInsert or AfterUpdate event in ActiveRecord, but I don't want it to fire repeatedly.

I use the following code:
```php
Class User extends ActiveRecord {
     public function login() {
         $event = new UserLoginEvent();
         $event->sender = $this;
         $handler = [$event, 'handler'];
         if ($this->getIsNewRecord()) {
             $this->off(self::EVENT_AFTER_INSERT, $handler);
             $this->on(self::EVENT_AFTER_INSERT, $handler);
         } else {
             $this->off(self::EVENT_AFTER_UPDATE, $handler);
             $this->on(self::EVENT_AFTER_UPDATE, $handler);
         }
     }
}
```

This method works fine when `eventWildcards` is not added.
I read the source code and found that the behavior of `eventWildcards` and `events` was not consistent.
Therefore, this warning will be generated.

ilgiz-badamshin:
Also error occurred when one handler was attached to event and second not attached handler tried detach
```
Event::on(MyClass::class, 'myEvent', [
    FirstHandler::class,
    'handle'
]);
Event::off(MyClass::class, 'myEvent', [
    SecondHandler::class,
    'handle'
]);
```
`yii\base\ErrorException: Undefined index: myEvent in yiisoft/yii2/base/Event.php:167`

SilverFire:
[...]

PR #16377, opened 2018-06-08
Fix Event off method undefined index error

razvanphp:
This fatal error happens when one tries to delete (off) an event that matches the class & event name but has different event handlers. The method should return false instead.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none


dmirogin (approved):

samdark:
Merged. Thank you!

ilgiz-badamshin:
Also should fix: https://github.com/yiisoft/yii2/issues/15953#issuecomment-378560092

PR #16406, opened 2018-06-14
Fix #16332: protected function can still be called outside the class

### LINKED ISSUE #16332: protected function can still be called outside the class
Declare a class then extends `BaseObject`

```
class Sample extends \yii\base\BaseObject {
    private $_mode = 'mode';
  
    protected function getMode()
    {
        return $this->_mode;
    }
}
```
Now create new instance
```
$sample = new Sample();
echo $sample->mode;

// output:
// mode
```
Though you can't call the `getMode()` method directly but still I think this is an issue.

I think this is because how the `__get($name)` magic function works. When it checks if a method exists or not, it will return true because it has access to that method.
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 5.6

jwlewisiii:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16332


samdark:
We'll definitely need upgrade note for this one because https://github.com/yiisoft/yii2/issues/16332#issuecomment-397522464

rob006:
This needs some performance tests, using reflection for getters/setters may affect performance significantly.

rob006:
Actually this should affect almost every method in `BaseObject` and related methods in `Component`. This change done only for `__set()` and `__get()` creates inconsistence in multiple places (`canSetProperty()` returns `true` for protected setter, but `__set()` will refuse to set property via this setter). This PR should break multiple tests. We need much more tests for this. :D

samdark:
Tests actually should be made for current master and, preferably, in a separate pull request...

Issue #16329, opened 2018-05-28
Active override input(): options lost


MacGyer:
### What steps will reproduce the problem?
- create a custom ActiveField class which extends `yii\widgets\ActiveField`
- create a new method which uses `parent::input()` to render the field
- use `Html::addCssClass()` in this method to add custom CSS classes (as an example to utilise the `$options`)

### What is the expected result?
In this case, `yii\widgets\ActiveField()` has the `$inputOptions = ['class' => 'form-control']`. The extending class' `emailInput()` method adds the email CSS class.

The custom CSS classes shall be merged with the existing, thus giving the field both CSS classes: `form-control` and `email`.

### What do you get instead?
The parent class' options (`yii\widgets\ActiveField::$inputOptions`) get overriden by child class.

This happens in https://github.com/yiisoft/yii2/blob/master/framework/widgets/ActiveField.php#L365.
A quick test with array_merge_recursive() or ArrayHelper::merge() led to the expected result.

### Code / Screens
```
class CustomActiveField extends \yii\widgets\ActiveField
{
    /**
     * @param array $options
     * @return ActiveField
     */
    public function emailInput($options = [])
    {
        Html::addCssClass($options, ['input' => 'email']);
        return parent::input('email', $options);
    }
}
```

![image](https://user-images.githubusercontent.com/2939625/40625519-48e6623c-62b3-11e8-86d2-0dccae97cf45.png)

![image](https://user-images.githubusercontent.com/2939625/40625572-aee81e4a-62b3-11e8-8026-426f5692bc20.png)

### Additional info
As I cannot overview all implications or do not know whether the override IS actually expected behaviour, I did not create a pull request with a fix.
Can do if you like.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15
| PHP version      | 7.1
| Operating system | Ubuntu 1604

rustamwin:
https://github.com/yiisoft/yii2/blob/6590feef0db1e994a42ffdc3536bccbe2f737ee4/framework/helpers/BaseHtml.php#L1953-L1954
use 
```
    public function emailInput($options = [])
    {
        Html::addCssClass($options, 'email');
        return parent::input('email', $options);
    }
```

MacGyer:
Thanks for getting back and pointing out.

But using the string variant of adding the CSS class leads to the exact same result. 

`Html::addCssClass($options, 'email');`

The original options from $this->inputOptions are not merged with $options.

Another project, same case:
[...]

PR #16421, opened 2018-06-18
fix ActiveField input to merge options from child classes #16329

MacGyer:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | tba
| Fixed issues  | #16329


rob006:
This probably breaks BC. And using `array_merge_recursive()` for merging configs is incorrect - it will give incorrect result if you try to override some settings. `array_replace_recursive()` or `ArrayHelper::merge()` should be used in this case.

samdark:
Can't be applied as is as @rob006 mentioned. Closing.

PR #16430, opened 2018-06-22
Fixes #16247: clone behaviors when cloning components

### LINKED ISSUE #16247: Cloning objects should also clone their behaviors
Currently, Component::__clone() will null-out the `$_behaviors` array:

https://github.com/yiisoft/yii2/blob/3d96a45f6b2ded171c607e20b9f3e39a618ab9cb/framework/base/Component.php#L307-L312

I can see why this is better than leaving `$_behaviors` alone (those behaviors contain a reference to the original component instance), but I think it would be better if those behaviors were cloned to the new instance as well.

```php
public function __clone()
{
    $this->_events = [];
    $this-> _eventWildcards = [];

    if ($this->_behaviors !== null) {
        $behaviors = array_merge($this->_behaviors);
        $this->_behaviors = null;
        foreach ($behaviors as $name => $behavior) {
            $this->attachBehavior($name, clone $behavior);
        }
}
```

That way any property values set on the behavior will be maintained on the cloned instance, just like normal properties.

```php
$obj = new MyClassWithBehavior();
$obj->myBehaviorProperty = 'bar';
$obj2 = clone $obj;
echo $obj2->myBehaviorProperty; // should output 'bar'
```

(related: craftcms/cms#2857)

brandonkelly:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #16247

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -52,6 +52,7 @@ Yii Framework 2 Change Log
 - Enh #16054: Callback execution with mutex synchronization (zhuravljov)
 - Enh #16126: Allows to configure `Connection::dsn` by config array (leandrogehlen)
 - Chg #11397: `yii\i18n\MessageFormatter` polyfills and `yii\i18n\MessageFormatter::parse()` method were removed resulting in performance boost. See UPGRADE for compatibility notes (samdark)
+- Chg #16247: Cloning components will now clone their behaviors as well (brandonkelly)
 
 2.0.14.2 under development
 ------------------------
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -302,13 +302,20 @@ class Component extends BaseObject
 
     /**
      * This method is called after the object is created by cloning an existing one.
-     * It removes all behaviors because they are attached to the old object.
+     * It clones all behaviors as well, and attaches them to the new object.
      */
     public function __clone()
     {
         $this->_events = [];
         $this->_eventWildcards = [];
-        $this->_behaviors = null;
+
+        if ($this->_behaviors !== null) {
+            $behaviors = $this->_behaviors;
+            $this->_behaviors = null;
+            foreach ($behaviors as $name => $behavior) {
+                $this->attachBehavior($name, clone $behavior);
+            }
+        }
     }
 
     /**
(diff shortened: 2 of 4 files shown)


samdark:
Merged. Thanks!

rob006:
Why events are not cloned in the same way?

samdark:
Should be, I guess. Want to make a pull request?

brandonkelly:
Events don’t seem as obvious as behaviors. Cloning an object is basically a shortcut to defining a new object with the same class and setting the same properties on it - and if behaviors are ment to act as extensions of the class, that definitely seems like it should be cloned too. But events are intentionally registered either at the class level (already will work on a clone) or for a specific instance. Maybe it's not intended that the events should be set on the clone as well though.

samdark:
Hmm... right. That's controversial.

rob006:
> Events don’t seem as obvious as behaviors.

I don't see any difference. It looks inconsistent to me, especially that events attached from behaviors are cloned, but events attached manually are not. 

But AFAIK there will be some changes in events handling in 3.0 so it may be irrelevant after all.

brandonkelly:
> especially that events attached from behaviors are cloned

Events added from behaviors are essentially the class dynamically defining its own business logic. The non-behavior equivelant would have been for the owner class to just call some internal method alongside firing the event.

I’m not saying it _definitely_ doesn’t make sense for events to be cloned as well; just that it’s not always going to be expected behavior. For example, if I did this:

```php
$obj = new MyObject();
$obj->on('afterDoSomething', function() {
    // ...
});

$obj2 = clone $obj;
$obj2->doSomething();
```

I don’t think expected behavior would be that the `afterDoSomething` handler would get called for `$obj2`. If you disagree, maybe it’s worth opening this up as a new issue for debate there. This PR is  probably the wrong place to discuss it :)

rob006:
@brandonkelly Let's just agree to disagree. For me current behavior is inconsistent and I don't see any reason why dynamically attached behaviors and its events should be handled differently than dynamically attached events. But this is not important for me, I almost never use this feature anyway. I just wanted to point it now, because fixing this inconsistency may require BC break (you cannot just clone all events because cloned behaviors already attaches its events, so you will get duplicated events).

samdark:
Let's not clone events for now.

PR #16436, opened 2018-06-25
Update DbManager.php

sobhan93:
For extendes and override DbManager methods like getAssignments method.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  |


dmirogin:
This PR break a BC, because people can define this method in their own realizations.

samdark:
Closing because of no use case description for a week.

Issue #16425, opened 2018-06-20
Html::a data-confirm => false for disabled confirm dialog


mariovials:
### What steps will reproduce the problem?
```
Html::a("Text", ["url/to"], [
  'data' => ['confirm' => false ]
]);
```

### What is the expected result?
Don't should by generated a confirm dialog

### What do you get instead?
But, it generates an empty confirm dialog

With this fix, we could type code cleaner and fast, put code in the same tag.
Now we must tye two tag for disabled the confirm dialog

Alex-Code:
If I've understood correctly this could be because the check only looks at whether `confirm` is set not at it's actual value so having `false` will still trigger the confirm dialog.
https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.js#L475

mariovials:
Yes, i suggest that evaluate the value before trigger confirm dialog, to improve the coding.

This would make it possible to write` 'confirm' => $confirmation`
if `$confirmation` is `false `(or `null `may be), do not trigger confirm dialog.

To make this, currently I have write twice, one for each situation

rob006:
Doesn't this do what you want?

```php
Html::a("Text", ["url/to"], [
  'data' => ['confirm' => null]
]);
```

mariovials:
Yes, just not generate a confirm dialog when confirm like false.
Either way, it would be nice if it worked with `''` (an empty string), `false `or `null`.

PR #16438, opened 2018-06-25
Data confirm

alex-code:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| Fixed issues  | #16425

Extra checks to the `data-confirm` attribute.


samdark:
They aren't too different.

framework/assets/yii.js
```diff
@@ -472,7 +472,7 @@ window.yii = (function ($) {
                 return true;
             }
 
-            if (message !== undefined) {
+            if (message !== undefined && message !== false && message !== '') {
```
rob006: How message could be `false`?

samdark: In case message is set to `false`. See https://api.jquery.com/data/#data2. We're using `data()` to get the value. It converts values automatically.

samdark (approved):
All good but need a line for changelog and tests.

Issue #16442, opened 2018-06-26
Signature of 'filter' closure for 'unique'/'exist' validator: add model and attribute


bscheshirwork:
### What steps will reproduce the problem?
I wish check unique by model-specific condition.
I try to use this rule
```php
...
    public function rules() {
        return [
...
            [
                ['flags'],  // one unique combination by each type
                'unique',
                'skipOnError' => true,
                'targetAttribute' => ['type'],
                'when' => function (ActiveRecord $model, string $attribute) {
                    return (bool) $model->$attribute; // the rule is applied if one or more sign is on
                },
                'filter' => function (ModelQuery $query, ActiveRecord $model, string $attribute) {
                    $query->isSign($model->$attribute);
                },
                'message' => yii::t('validators', 'The flag must be one unique combination by type'),
            ],
```
but 'filter' function can accept only `query` and not accept current `model` and `attribute`

### What is the expected result?
pass to closure `model` and `attribute`

### What do you get instead?
not implemented
```php
    /**
     * Prepares a query by applying filtering conditions defined in $conditions method property
     * and [[filter]] class property.
     *
     * @param ActiveRecordInterface $targetClass the name of the ActiveRecord class that should be used to validate
     * the uniqueness of the current attribute value.
     * @param array $conditions conditions, compatible with [[\yii\db\Query::where()|Query::where()]] key-value format
     *
     * @return ActiveQueryInterface|ActiveQuery
     */
    private function prepareQuery($targetClass, $conditions)
    {
        $query = $targetClass::find();
        $query->andWhere($conditions);
        if ($this->filter instanceof \Closure) {
            call_user_func($this->filter, $query);
        } elseif ($this->filter !== null) {
            $query->andWhere($this->filter);
        }

        return $query;
    }

```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | dev
| PHP version      | 7.2.6
| Operating system |alpine

samdark:
Won't be implemented in 2.0.

PR #16444, opened 2018-06-26
change signature of 'filter'

bscheshirwork:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes/no
| Tests pass?   | yes/no
| Fixed issues  | #16442


samdark:
Won't be implemented in 2.0.

OndrejVasicek:
Is there a reason why the PR wasn't allowed? This PR seems like a very handy update.
Otherwise one have to create a custom validation rule to achieve the same. Which isn't such a clean solution.

Issue #16278, opened 2018-05-14
console migrate/frersh fail on db view


bobonov:
Running migrate/fresh will fail on dropping view since the command run drop table instead of drop view
The command assume that all the items in schema are table:
Line 312 in yii2/console/controllers/MigrateController.php
```php
        // Then drop the tables:
        foreach ($schemas as $schema) {
            $db->createCommand()->dropTable($schema->name)->execute();
            $this->stdout("Table {$schema->name} dropped.\n");
        }
```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.2.3
| Operating system | Ubuntu 18.04

PR #16446, opened 2018-06-27
Fixed drop existing views when console `migrate/fresh` command runs

developeruz:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16278

### DIFF
--- a/framework/console/controllers/MigrateController.php
+++ b/framework/console/controllers/MigrateController.php
@@ -310,8 +310,17 @@ class MigrateController extends BaseMigrateController
 
         // Then drop the tables:
         foreach ($schemas as $schema) {
-            $db->createCommand()->dropTable($schema->name)->execute();
-            $this->stdout("Table {$schema->name} dropped.\n");
+            try {
+                $db->createCommand()->dropTable($schema->name)->execute();
+                $this->stdout("Table {$schema->name} dropped.\n");
+            } catch (\Exception $e) {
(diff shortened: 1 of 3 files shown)


framework/console/controllers/MigrateController.php
```diff
@@ -310,8 +310,17 @@ protected function truncateDatabase()
 
         // Then drop the tables:
         foreach ($schemas as $schema) {
-            $db->createCommand()->dropTable($schema->name)->execute();
-            $this->stdout("Table {$schema->name} dropped.\n");
+            try {
+                $db->createCommand()->dropTable($schema->name)->execute();
+                $this->stdout("Table {$schema->name} dropped.\n");
+            } catch (\Exception $e) {
+                if (strpos($e->getMessage(), 'DROP VIEW to delete view') !== false) {
```
samdark: Have you checked if the message is valid for all databases we support?

samdark (changes requested):

framework/console/controllers/MigrateController.php
samdark: We run PostgreSQL, SQLite and MySQL. We don't run Oracle and MSSQL.

samdark:
Merged. Thank you!

PR #16457, opened 2018-06-29
Fixes #16456: OptimisticLockBehavior omitting web HTML forms

### LINKED ISSUE #16456: OptimisticLockBehavior doesnt work.
https://github.com/yiisoft/yii2/blob/master/framework/behaviors/OptimisticLockBehavior.php#L128

As it stand OptimisticLockBehavior expected the submission of the lock attribute like so 
```php
echo Html::hiddenInput('lock', $value); // <input name="lock" type="hidden" />
```
shouldn't it expect the following by default

```php
echo   $form->field($model, 'lock')->hiddenInput(); // <input name="ModelName[lock]" type="hidden" />
```

This is how i think the  `getValue($event)` should behave.
```php
    /**
     * {@inheritdoc}
     *
     * In case of `null`, value will be parsed from [[\yii\web\Request::getBodyParam()|getBodyParam()]] or set to 0.
     */
    {
        if ($this->value === null) {
            $lock = $this->getLockAttribute();
            $scope = $this->owner->formName();
            $params = Yii::$app->getRequest()->getBodyParam($scope);
            $input = isset($params[$lock]) ? $params[$lock] : null;

            $isValid = $input && (new NumberValidator())->validate($input);
            return $isValid ? $input : 0;
        }
        return parent::getValue($event);
    }
```

tunecino:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no (target feature is tagged 2.0.16 and yet to be released)
| Tests pass?   | yes
| Fixed issues  | #16456


samdark:
Merged. Thank you!

PR #16458, opened 2018-06-29
Added Changed renderSelectOptions method of the BaseHtml,make the dropDownList component of the yii framework support adding several attributes to the select option.

presidentma:
Changed renderSelectOptions method of the BaseHtml,make the dropDownList component of the yii framework support adding several attributes to the select option.The array $option key use itemAttr,and value of the itemAttr  is an array.For example,Html::dropDownList('dropDown',null,$userArr,['class'=>'form-control','itemAttr'=>['id'=>$itemId]]).

| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1859,6 +1859,11 @@ class BaseHtml
             } else {
                 $attrs = isset($options[$key]) ? $options[$key] : [];
                 $attrs['value'] = (string) $key;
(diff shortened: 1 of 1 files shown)


samdark:
Closing since initial problem was not clear.

PR #16495, opened 2018-07-06
Added method call support in DI and Yii::configure

hiqsol:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Cannot leave `Yii::configure()` untouched to have consistent behavior for all objects.

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -563,7 +563,11 @@ class BaseYii
     public static function configure($object, $properties)
     {
         foreach ($properties as $name => $value) {
-            $object->$name = $value;
+            if (substr($name, -2) === '()') {
+                call_user_func_array([$object, substr($name, 0, -2)], $value);
+            } else {
+                $object->$name = $value;
+            }
         }
 
         return $object;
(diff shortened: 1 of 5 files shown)


samdark (approved):
Looks good. I think we need to update docs.

framework/BaseYii.php
```diff
@@ -563,7 +563,11 @@ public static function t($category, $message, $params = [], $language = null)
     public static function configure($object, $properties)
     {
         foreach ($properties as $name => $value) {
-            $object->$name = $value;
+            if (substr($name, -2) === '()') {
```
SilverFire: mb_substr should be used

framework/di/Container.php
```diff
@@ -402,7 +402,11 @@ protected function build($class, $params, $config)
 
         $object = $reflection->newInstanceArgs($dependencies);
         foreach ($config as $name => $value) {
-            $object->$name = $value;
+            if (substr($name, -2) === '()') {
```
SilverFire: mb_substr should be used

SilverFire (changes requested):
It will be good to add test for calling not existing or private method

samdark:
Merged. Thank you very much!

PR #16496, opened 2018-07-06
WIP: Active Record attribute getters and setters

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | https://github.com/yiisoft/active-record/issues/4

This PR is still work in progress. I will remove `WIP` when PR is final.

I have several questions to discuss before finishing it:

1. Is it better to have `getName`, `setName` or `getNameAttribute` and `setNameAttribute` style getters and setter. I personally like to have simple getters and setters (without any suffix), as it is more intuitive for programmers (they are like usual getters and setters), though I think it can be confusing and may add place for mistakes and difficulty debugging (maybe).
2. If you check `setAttribute` setter, you will see it returns the value. Is it better to just set the value there without returning it, as it is just setter, the body will become something like `$this->attributes['name'] = 'something';`. I did with returning value, as it is more intuitive for me that setter mutates the value and returns it, but if we come closer to programming setters, the setter must have `void` type and just do it's job.
3. I have already mentioned in original issue (and referenced stackoverflow) that `method_exists` has very small performance hit, but I agree that caching would be better anyways. The thing is there can be 2 approaches for caching. Caching getter and setter names (actually storing some map, that there is getter and/or setter for attribute) and accessing it without doing further `method_exists` checks or the second method would be to cache only setters and for getters to cache actual value returned by it. It will also save on any calculations done in getter, but it has one downside. If we consider getter simply mutator for the specific attribute (I mean getter is working only with that attribute value), it will work ok. but if the getter is calculating final value based on other attributes we can have stale cache when other attribute changes between 2 calls to getter.
4. Do I need to add any docs? Are we doing docs for 3.0 already?

Thanks for your attention.

### DIFF
--- a/framework/db/ActiveQueryInterface.php
+++ b/framework/db/ActiveQueryInterface.php
@@ -19,6 +19,8 @@ namespace yii\db;
  * @author Qiang Xue <qiang.xue@gmail.com>
  * @author Carsten Brandt <mail@cebe.cc>
  * @since 2.0
+ *
+ * @property $link
  */
 interface ActiveQueryInterface extends QueryInterface
 {
--- a/framework/db/ActiveRecordInterface.php
+++ b/framework/db/ActiveRecordInterface.php
@@ -294,12 +294,12 @@ interface ActiveRecordInterface extends StaticInstanceInterface
      *
      * @param array $attributes attribute values (name-value pairs) to be saved for the record.
      * Unlike [[update()]] these are not going to be validated.
-     * @param array $condition the condition that matches the records that should get updated.
+     * @param string|array $condition the condition that matches the records that should get updated.
      * Please refer to [[QueryInterface::where()]] on how to specify this parameter.
      * An empty condition will match all records.
      * @return int the number of rows updated
      */
-    public static function updateAll($attributes, $condition = null);
+    public static function updateAll($attributes, $condition = '');
 
     /**
      * Deletes records using the provided conditions.
--- a/tests/data/ar/Order.php
+++ b/tests/data/ar/Order.php
@@ -217,9 +217,9 @@ class Order extends ActiveRecord
         ];
     }
 
-    public function getTrackingNumber()
+    public function getTrackingNumber($value)
     {
-        return '#' . $this->tracking_number;
+        return '#' . $value;
     }
 
     public function setTrackingNumber($value)
--- a/tests/data/ar/Order.php
+++ b/tests/data/ar/Order.php
@@ -14,6 +14,7 @@ namespace yiiunit\data\ar;
  * @property int $customer_id
  * @property int $created_at
  * @property string $total
+ * @property string $tracking_number
  */
 class Order extends ActiveRecord
 {
@@ -214,5 +215,15 @@ class Order extends ActiveRecord
         return [
             0 => 'customer_id',
         ];
+    }
+
+    public function getTrackingNumber()
+    {
+        return '#' . $this->tracking_number;
+    }
+
+    public function setTrackingNumber($value)
+    {
+        return str_pad($value, 16, '0', STR_PAD_LEFT);
     }
 }
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -16,6 +16,7 @@ use yii\base\ModelEvent;
 use yii\base\NotSupportedException;
 use yii\base\UnknownMethodException;
 use yii\helpers\ArrayHelper;
+use yii\helpers\Inflector;
 
 /**
  * ActiveRecord is the base class for classes representing relational data in terms of objects.
@@ -105,6 +106,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
     /**
      * {@inheritdoc}
      * @return static|null ActiveRecord instance matching the condition, or `null` if nothing matches.
+     * @throws InvalidConfigException
      */
     public static function findOne($condition)
     {
@@ -114,6 +116,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
     /**
      * {@inheritdoc}
      * @return static[] an array of ActiveRecord instances, or an empty array if nothing matches.
+     * @throws InvalidConfigException
      */
     public static function findAll($condition)
     {
@@ -222,7 +225,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      * 1. Create a column to store the version number of each row. The column type should be `BIGINT DEFAULT 0`.
      *    Override this method to return the name of this column.
      * 2. Ensure the version value is submitted and loaded to your model before any update or delete.
-     *    Or add [[\yii\behaviors\OptimisticLockBehavior|OptimisticLockBehavior]] to your model 
+     *    Or add [[\yii\behaviors\OptimisticLockBehavior|OptimisticLockBehavior]] to your model
      *    class in order to automate the process.
      * 3. In the Web form that collects the user input, add a hidden field that stores
      *    the lock version of the recording being updated.
@@ -280,10 +283,16 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      * @throws \yii\base\InvalidArgumentException if relation name is wrong
      * @return mixed property value
      * @see getAttribute()
+     * @throws \yii\base\UnknownPropertyException
      */
     public function __get($name)
     {
         if (isset($this->_attributes[$name]) || array_key_exists($name, $this->_attributes)) {
+            $getter = 'get' . Inflector::variablize($name);
+            if (method_exists($this, $getter)) {
+                // read attribute, e.g. getName()
+                return $this->$getter();
+            }
             return $this->_attributes[$name];
         }
 
@@ -308,6 +317,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
      * This method is overridden so that AR attributes can be accessed like properties.
      * @param string $name property name
      * @param mixed $value property value
+     * @throws \yii\base\UnknownPropertyException
      */
     public function __set($name, $value)
     {
@@ -317,6 +327,11 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
                 && (!array_key_exists($name, $this->_attributes) || $this->_attributes[$name] !== $value)
             ) {
                 $this->resetDependentRelations($name);
+            }
+            $setter = 'set' . Inflector::variablize($name);
+            if (method_exists($this, $setter)) {
+                // set attribute, e.g. setName()
+                $value = $this->$setter();
             }
             $this->_attributes[$name] = $value;
         } else {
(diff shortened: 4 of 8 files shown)


armpogart:
@Inkognitoo Sorry I don't get why you think it won't work. The tests pass.

P.S. Если что я по русски тоже понимаю ))

MattRh:
Why did you add docs changes into this PR?

MattRh:
And in current state I think we have to add getter caching with ability to flush this cache and rename it from `Class::getName()` to `Class::prepareName()` as this may be more convinient

MattRh:
As for your doubts about caching variants - `Caching getter and setter names (actually storing some map, that there is getter and/or setter for attribute) and accessing it without doing further method_exists` - would be absolute useless, as by that you will cache only names in hashmap. I think this may be even slower than `method_exists` as this function searches this function name in function hashmap(https://github.com/php/php-src/blob/091d77f28a78a700ce0e72457606e01ed696de83/Zend/zend_inheritance.c#L1379). This is worthless to create another table

rob006:
Can you remove all unrelated changes from this PR? Most of PHPDoc changes is incorrect.

armpogart:
@Inkognitoo Ну да, геттеры не принимают атрибуты. Но это так не будет работать тут, так как в геттере ты будешь опять работать с атрибутом ($this->attribute) и как бы опять этот же геттер будет вызываться и так до бесконечности :) Это работает так же в ларавеле.

@MattRh `Class::getName()` won't work, you need to pass actual value to getter to use it there, or you will call it again from getter `$this->attribute` will get recursive calls.... I don't get it why to rename it to `prepareName`, it is counterintuitive to have getter with such name))

@rob006 Yep, sure. It's my IDE. PHPStorm did this automatically. Also I think we need separate issue to remove all `Exception`s for transactions as with minimum required php version raised `Exception` will never be thrown there, only `Throwable`.

Still waiting for some discussion on the proposed questions. @MattRh arguments on getter name caching are valid. I see php source code that there is no a reason to do it. So the final question for caching would be, do we need a way to cache return value from getter and way to invalidate that cache?

rob006:
If I define `getTrackingNumber()` and `setTrackingNumber()` for my `tracking_number` attribute, how can I access `tracking_number` attribute directly without involving setters/getters?

armpogart:
@rob006 You can still access it from attributes array `$this->attributes['tracking_number']` and in majority cases I think you won't need to access it directly if you have defined the getter (of course there will be still reasons to access it, but more often no).

armpogart:
@rob006 You can still access it from attributes array `$this->getAttribute('tracking_number')` and in majority cases I think you won't need to access it directly if you have defined the getter (of course there will be still reasons to access it, but more often no).

rob006:
Honestly, I don't like this approach:

1. You cannot disable it.
2. It is inflexible, because it forces some naming conversions...
3. ...which conflicts with current convention of virtual attributes provided by getter/setters - getter requires value as a parameter and setter does not set anything. This is not only inconsistent, but it is also a huge BC break - I often use pairs of `some_data` column with encoded JSON and `getSomeData()`/`setSomeData()` methods to provide virtual attribute `someData` with decoded JSON. 
 This breaks by code (which already does what you want) in multiple places. 
4. It has performance implications - `method_exists()` may be cheap, but `Inflector::camelize()` is not.
5. It does not fix most annoying limitation of magic getters/setters - you cannot do `$this->someData['someKey'] = 'someValue'` if `someData` is provided by setter/getter.

First 4 points may be fixed by some map of getters/setter:

```php
protected function mutators() {
	return [
		'some_attribute' => [
			'set' => function ($value) {
				return json_encode($value);
			},
			'get' => function ($value) {
				return json_decode($value);
			},
		],
	];
}
```

1. Witch empty `mutators()` you don't get any undesired behavior.
2. No `method_exists()` or `Inflector::camelize()` calls, only `isset()` on `mutators()` result.
3. No conventions which reserves some method names and conflicts with existing conventions.
4. It could be integrated with solution, which fixes 5.

Inkognitoo:
@rob006
1. Just don't use it. But if you use it and still need original attribute, get it through `getAttribute($attribute_name)`
2. It's super flexible because you can mutate data before get it or write to db. You think it's can bring some bracking change issue and it is. But if you have attribute `name` and you use getter `getName()` for something that is not attribute name exactli, it's very strange, I think )
3. You right, no comment, but it's just your case
4. Maybe we need some little native function for this
5. And it's normal, I think

rob006:
@Inkognitoo 

1. I cannot "not use it" - this solution will use my getters/setters in a wrong way even if I don't want to. 
2. I don't have getters/setters for virtual attributes with the same name as regular attributes (it will not work anyway) - `some_data` != `someData`.

I proposed alternative approach to achieve the same effect without these drawbacks - it will be more productive to comment implementation proposal instead of "defending" the idea of mutators (which was not "attacked" by me... yet :P).

Inkognitoo:
I don't think we need some over engendering here
We just need simple getters and setters
@rob006 is we are can overthrow your issues if will use get/set<name>Attribute() for custom get/set instead just get/set<name>?
For example how it's look like in laravel: https://laravel.com/docs/5.6/eloquent-mutators

rob006:
> We just need simple getters and setters

We already have simple getters and setters. This new convention conflicts with existing one.

rob006:
BTW: it could be implemented as a trait:

	public function __get($name) {
		if ($this->hasAttribute($name)) {
			$getter = 'get' . Inflector::camelize($name);
			if (method_exists($this, $getter)) {
				// read attribute, e.g. getName()
				return $this->$getter(parent::__get($name));
			}
		}
		
		return parent::__get($name);
	}

It still will be inflexible, inconsistent and slow, but at least it will be optional.

framework/db/ActiveRecordInterface.php
```diff
@@ -294,12 +294,12 @@ public static function findAll($condition);
...
      * @param array $attributes attribute values (name-value pairs) to be saved for the record.
      * Unlike [[update()]] these are not going to be validated.
-     * @param array $condition the condition that matches the records that should get updated.
+     * @param string|array $condition the condition that matches the records that should get updated.
      * Please refer to [[QueryInterface::where()]] on how to specify this parameter.
      * An empty condition will match all records.
      * @return int the number of rows updated
      */
-    public static function updateAll($attributes, $condition = null);
+    public static function updateAll($attributes, $condition = '');
```
cebe: this change is incorrect, `ActiveRecordInterface` is independet from SQL DB so the correct value for no condition is `null`.

tests/data/ar/Order.php
```diff
@@ -215,4 +216,14 @@ public function activeAttributes()
             0 => 'customer_id',
         ];
     }
+
+    public function getTrackingNumber($value)
```
cebe: if we ever add something like this, the naming should not be `getX()` or `setX()` because such naming is reserved for the properties feature in Yii BaseObject.

framework/db/ActiveQueryInterface.php
```diff
@@ -19,6 +19,8 @@
  * @author Qiang Xue <qiang.xue@gmail.com>
  * @author Carsten Brandt <mail@cebe.cc>
  * @since 2.0
+ *
+ * @property $link
```
dmirogin: Property for interface ? wtf ?

armpogart: Yep, just for autocompletion.

tests/data/ar/Order.php
armpogart: @cebe I'm going to rewrite a little bit this pull request, after some more discussion in linked issue.

framework/db/ActiveRecordInterface.php
armpogart: @cebe But in implementation of the interface you have `$condition = ''`, and IDE is raising warning there. Maybe it's better not to have default value in interface at all and just document possible values.

rob006: > IDE is raising warning there

IDE is wrong, changing default value of parameter works perfectly fine in PHP.

framework/db/ActiveQueryInterface.php
rob006: Interfaces does not have properties, this change is incorrect.

armpogart:
@rob006 

> We already have simple getters and setters. This new convention conflicts with existing one

Nope, I don'k so. If I understand you correctly by existing one you mean [data transformation](https://www.yiiframework.com/doc/guide/2.0/en/db-active-record#data-transformation). It is clearly documented as a way to do data transformation and I see it that way.

The difference here would be, that if I generally need value that is in db and in some cases I need some formatting or transformation, I will surely use the existing technique. But if I know that I will always need db data somehow formatted or transformed than I will use the technique used in this pull request. Also there is no existing way to have setter except doing that in some `before` event, the downsides of which were already mentioned.

armpogart:
@rob006 @cebe 

> BTW: it could be implemented as a trait:
> It still will be inflexible, inconsistent and slow, but at least it will be optional.

I still see it as a core functionality which will be helpful for beginners and newcomers. Yeah, I won't have any problems to use a trait, or use my own `ActiveRecord` as a parent for my models which will implement this functionality on it's own, but for beginners and newcomers only core functionality matters at first. I'm sure majority of beginners don't even check what extensions are there that can help them (even core extensions, if they are not documented in the guide).

For the trait: yeah it could be done in the trait, but it will it's own problems and many of them. If you really want it to be optional and if you don't want to have the performance hit (even though I think it is negligible), we can hide it under some flag and deactivate it by default (clearly documenting this functionality). What do you think about this approach?

P.S. I will refactor a little bit this PR in coming days and remove all those changes that are not related to the functionality.

armpogart:
@rob006 And for the performance part, I can also do some benchmarking to see whether it is slow and really performance hit, or some negligible change.

framework/db/ActiveQueryInterface.php
armpogart: Yeah, I know, I just did it for autocompletion and IDE was raising warning. I will remove it.

framework/db/ActiveRecordInterface.php
armpogart: @rob006 It works fine, but I see it as a violation of contract between the class and interface. If the default parameter is changed in majority of places what is the purpose of having it in interface in the first place?

rob006: > If the default parameter is changed in majority of places what is the purpose of having it in interface in the first place?

To define parameter which is optional, as a part of the interface. Default value is a implementation detail and it is not part of the contract.

rob006:
> Also there is no existing way to have setter

You can define setter like this:

```php
public function setAttributeFormatted($value) {
    $this->attribute = $value;
}
```

And use it:

```php
$model->attributeFormatted = $someValue;
```

The only difference is that attributes have precedence over setters/getters (so you cannot use getter/setter with the same name as attribute) and it does not automatically convert snake_case attributes to camelCase getters/setters.

samdark:
Can't be applied in its current form because it will introduce BC breaks.

PR #16520, opened 2018-07-15
Update start-installation.md

teo1978:
In the Yii2 Guide, where the Advanced Project Template is mentioned, the link should be to the topic about it in the guide, rather than to an .md file in a github repository.


samdark:
Merged. Thanks!

PR #16523, opened 2018-07-16
Fixed #15826 : JavaScript compareValidator for attributes not in rules

### LINKED ISSUE #15826: compareValidator javascript validator only works with attributes in rules
since commit [link](https://github.com/yiisoft/yii2/commit/c009b0611472359c6f2e8a4a38b954c2bc46f612) the javascript for compareValidator does not work anymore, because the targetAttribute is only found when it's in the rules section. 

```
public function rules()
    {
        return [
            ['password', 'compare'],
        ];
    }
```
password validates correctly to the password_repeat attribute.

### What do you get instead?
It always does not validate and gives an error while both values are the same, because the password_repeat attribute was not found in the client validation method. 

### Suggestion
Adding password_repeat as safe attribute was also not working. Adding another validator linke StringValidator for password_repeat fixed the validation. Like:
```
public function rules()
    {
        return [
            ['password', 'compare'],
            [['password_repeat'], 'string', 'length' => [6, 40]],
        ];
    }
```

My suggestion is to add a warning or notice to the validator, that the target attribute must be an attribute in the rules section. 

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14.1
| PHP version      | 7.2.3
| Operating system | Linux

mgrechanik:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15826

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #15826: Fixed JavaScript compareValidator in `yii.validation.js` for attributes not in rules (mgrechanik)
 - Enh #16365: Added $filterOnFocusOut option for GridView (s1lver)
 - Enh #14289: Added `yii\db\Command::executeResetSequence()` to work with Oracle (CedricYii)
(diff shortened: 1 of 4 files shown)


samdark:
Thanks!

Issue #15791, opened 2018-02-27
data param attribute does not allow setting key with name submit


buttflattery:
If we try to assign the `key` with the name `submit` when setting the `data-params` attribute, clicking the link will do nothing, maybe because it is reserved and set automatically, because as soon as i change the key name from `submit` to `Submit` it starts working, but it is not documented anywhere in the docs under the [renderTagAttributes()](http://www.yiiframework.com/doc-2.0/yii-helpers-basehtml.html#renderTagAttributes()-detail)

### What steps will reproduce the problem?
```
<?=Html::a('Submit', ['save', 'id' => $model->id], [
    'class' => 'btn btn-primary',
    'data' => [
        'method' => 'POST',
        'params' => ['submit' => 'my-value'],
    ],
]
)?>
```
### What is the expected result?
clicking the link should submit the page to the desired link 

### What do you get instead?
nothing happens when clicking on the link 

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.14?
| PHP version      | 5.6
| Operating system |Ubuntu

ieatbytes:
its actually yii.js issue, for a temporary fix you can override yii.js handleAction method but before doing that i will request you fetch the fresh master release and check. If you still have that issue please create a ticket.

s1lver:
Yii version 2.0.15.1

There is a problem. And it is not at the same time. Rather, it is the expected behavior.

But, there are many questions in general about the appropriateness of using. It would not occur to me to use this way, there are other methods for these purposes.

The matter is that form is transferred from the transferred parameters. Naturally, with the transfer '**submit**', the formation of '**input**':
```html
<input name="submit">
```

s1lver:
the solution is simple. Let's check if we have "submit", if there is, then we call the trigger on the first element (This will be our form). otherwise we call as usual.

```js
if (("submit" in params)) {
    $form[0].trigger('submit');
} else {
    $form.trigger('submit');
};
```

s1lver:
I think that the correction does not make sense and the topic can be closed.

> Be careful when naming form elements such as submit buttons. According to the [jQuery documentation](https://api.jquery.com/submit/) there are some reserved names that can cause conflicts:
 Forms and their child elements should not use input names or ids that conflict with properties of a form, such as submit, length, or method. Name conflicts can cause confusing failures. For a complete list of rules and to check your markup for these problems, see [DOMLint](http://kangax.github.io/domlint/).


[...]

PR #16557, opened 2018-07-24
Fix submit form (#15791)

s1lver:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #15791

Simple bug fix


samdark:
Why waiting? Can't you do it as well?

s1lver:
Ok I'll add.
I was confused by the fact that there are tests for yii.js, but they were not started at the commit.

s1lver:
I tested this change. Need another option (

rustamwin:
How about checking with `hasOwnProperty`

PR #16568, opened 2018-07-28
Added ability to disable transliteration in Inflector::slug

gevorgmansuryan:
In some cases needed disable transliteration for non english characters. Changed preg_replace to support non english characters if transliteration is disabled, and strtolower to mb_strtolower.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/helpers/BaseInflector.php
+++ b/framework/helpers/BaseInflector.php
@@ -473,7 +473,7 @@ class BaseInflector
      * @param string $string An arbitrary string to convert
      * @param string $replacement The replacement to use for spaces
      * @param bool $lowercase whether to return the string in lowercase or not. Defaults to `true`.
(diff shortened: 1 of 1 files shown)


samdark:
Which cases are that?

gevorgmansuryan:
@samdark for example needed keep original language in uri (based on project specs) for example in multilang apps and not specify language in uri.
For example:
en: `/news/simple-article`
ru: `/новости/простая-статья` instead of `/новости/prostaya-statya`

gevorgmansuryan:
Also it will increase chances of page match in Search Engines, coz they take a look at uri also.

samdark:
OK. Then enhancement makes sense. Will be merged in 3.x.

samdark:
https://github.com/yiisoft/inflector/issues/2

PR #16569, opened 2018-07-28
Исправление для #15889

lesha724:
Добавлена возможность изменить/дополнить метод setActivePlaceholder после переопределиния базового класса Html

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #15889

### DIFF
--- a/tests/framework/helpers/HtmlTest.php
+++ b/tests/framework/helpers/HtmlTest.php
@@ -1729,6 +1729,34 @@ HTML;
         $this->assertContains('placeholder="Name"', $html);
     }
 
+    public function testOverrideSetActivePlaceholder()
+    {
+        $model = new HtmlTestModel();
+
+        $html = MyHtml::activeTextInput($model, 'name', ['placeholder' => true]);
+
+        $this->assertContains('placeholder="My placeholder: Name"', $html);
+    }
+}
+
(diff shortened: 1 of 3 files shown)


samdark (approved):
Please add changelog. Thanks!

samdark:
Merged. Thanks!

tests/framework/helpers/HtmlTest.php
```diff
@@ -1729,6 +1729,34 @@ public function testActiveTextInput_placeholderFillFromModelTabular()
...
+
+        $this->assertContains('placeholder="My placeholder: Name"', $html);
+    }
+}
+
+/**
+ * Class MyHtml
+ * @package yiiunit\framework\helpers
+ */
+class MyHtml extends Html{
```
rob006: Why two classes in single file?

lesha724: Because this class was used only once. If you want I can transfer it to a separate file.

rob006: It breaks PSR-1. I'm for fixing it (including classes existing before this PR).

Issue #14901, opened 2017-10-02
Client trim validation filter get wrong value with radioList (checkboxList)


degibons:
Model: 
```php
public function rules()
    {
        return [
            [
                ['myAttr'],
                'trim',
            ],
            [
                ['myAttr'],
                'required',
            ],
       ];
}
```

View (ActiveFormWidget):
```
form.field(model, 'myAttr').radioList(items)
```

Generated (with yiiActiveForm) javascript validation part:
```
... "validate":function (attribute, value, messages, deferred, $form) {value = yii.validation.trim($form, attribute, []);yii.validation.required(value, messages, {"message":"Required"}); ...
```

Trim function in `yii.validation.js` (`trim: function ($form, attribute, options)`) for radioList gets the value from the container `div` which is always empty string. So all remaining client validators get empty value. In my case requered validator (following after trim) always gets the error.

s1lver:
@degibons The error is not reproduced.
The code is generated:

```js
jQuery('#w0').yiiActiveForm([{
    "id":"test-myattr",  
    "name":"myAttr",
    "container":".field-test-myattr",
    "input":"#test-myattr",
    "validate":function (attribute, value, messages, deferred, $form) {
        yii.validation.required(value, messages, {"message":"Необходимо заполнить «My Attr»."});
        value = yii.validation.trim($form, attribute, []);
    }
}], []);
```
But, at the same time, all client validators continue to work correctly...

degibons:
@s1lver 
Forgot to mention, you need to predetermine the value of the radioList attribute.
Model:
```
public $attr = 1;

public function getItems() {
        return [
            1 => 'one',
            2 => 'two',
            3 => 'three',
        ];
}

public function rules() {
        return [
            [
                ['attr'],
                'trim',
            ],
            [
                ['attr'],
                'required',
            ],
        ];
}
```

View:
```
<?php $form = ActiveForm::begin(); ?>
<?= $form->field($model, 'attr')->radioList($model->getItems()) ?>
```

s1lver:
Really. In this case, whatever value is not selected, the test will fail.

```js
jQuery('#w0').yiiActiveForm([
    {
        "id":"test-attr",
        "name": "attr",
        "container":".field-test-attr",
        "input":"#test-attr",
        "validate":function (attribute, value, messages, deferred, $form) {
            value = yii.validation.trim($form, attribute, []);
[...]

PR #16599, opened 2018-08-03
Fixed trim validation (#14901)

s1lver:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #14901

Option without loss, possible, compatibility. But, I would get rid of **$form**

### DIFF
--- a/framework/assets/yii.validation.js
+++ b/framework/assets/yii.validation.js
@@ -236,9 +236,15 @@ yii.validation = (function ($) {
             }
         },
 
-        trim: function ($form, attribute, options) {
+        trim: function ($form, attribute, options, value) {
             var $input = $form.find(attribute.input);
-            var value = $input.val();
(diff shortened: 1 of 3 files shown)


framework/assets/yii.validation.js
```diff
@@ -236,9 +236,15 @@ yii.validation = (function ($) {
             }
         },
 
-        trim: function ($form, attribute, options) {
+        trim: function ($form, attribute, options, value) {
             var $input = $form.find(attribute.input);
-            var value = $input.val();
+            var isCheckAble = $input.find('[type=radio]').is('[type=radio]') || $input.find('[type=checkbox]').is('[type=checkbox]');
+
+            if (!isCheckAble) {
```
samdark: Negation isn't needed here. Will fix separately.

s1lver: Seems to understand what is at stake. I will now make a change

s1lver: I saw the fix

PR #16292, opened 2018-05-16
Fix class namespace

khvalov:
Fix for error:  exception 'yii\base\ErrorException' with message 'Class 'yii\filters\Exception' not found' in /www/vendor/yiisoft/yii2/filters/Cors.php:171

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

PR #16602, opened 2018-08-03
Fix namespace issue

SergeAx:
Causes an error when `YII_DEBUG` is `true`, `$cors['Origin']` contains `'*'` and `$cors['Access-Control-Allow-Credentials']` is `true`

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | --
| Fixed issues  | --

### DIFF
--- a/framework/filters/Cors.php
+++ b/framework/filters/Cors.php
@@ -167,7 +167,7 @@ class Cors extends ActionFilter
                 // Per CORS standard (https://fetch.spec.whatwg.org), wildcard origins shouldn't be used together with credentials
                 if (isset($this->cors['Access-Control-Allow-Credentials']) && $this->cors['Access-Control-Allow-Credentials']) {
                     if (YII_DEBUG) {
-                        throw new Exception("Allowing credentials for wildcard origins is insecure. Please specify more restrictive origins or set 'credentials' to false in your CORS configuration.");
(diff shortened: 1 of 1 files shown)


samdark:
Merged. Thanks!

framework/filters/Cors.php
```diff
@@ -167,7 +167,7 @@ public function prepareHeaders($requestHeaders)
                 // Per CORS standard (https://fetch.spec.whatwg.org), wildcard origins shouldn't be used together with credentials
                 if (isset($this->cors['Access-Control-Allow-Credentials']) && $this->cors['Access-Control-Allow-Credentials']) {
                     if (YII_DEBUG) {
-                        throw new Exception("Allowing credentials for wildcard origins is insecure. Please specify more restrictive origins or set 'credentials' to false in your CORS configuration.");
+                        throw new \Exception("Allowing credentials for wildcard origins is insecure. Please specify more restrictive origins or set 'credentials' to false in your CORS configuration.");
```
rob006: It does not look right - `\Exception` is too generic. It should be `InvalidConfigException` or at least `yii\base\Exception`.

rob006:
This was already fixed in master: https://github.com/yiisoft/yii2/pull/16292

Why this change is target to 2.0.16 but merged to 3.0? Both versions should use the same fix with `InvalidConfigException` instead `\Exception`.

Issue #16603, opened 2018-08-03
FileMutex not able to release locks when running on Windows-hosted Unix VM


brandonkelly:
### What steps will reproduce the problem?

1. Create a Unix-based VM on a Windows host box, using VirtualBox (e.g. with Vagrant or Docker).
2. Share a Windows directory with the VM
3. Install Yii in that shared directory
4. Write a script that creates a lock with the FileMutex class and (attempts to) releases it.

### What is the expected result?

The lock should be released.

### What do you get instead?

The lock doesn’t get released, because PHP thinks it is running in a Unix environment, and so the Unix fork of logic here is run, when the Windows fork should have run:

https://github.com/yiisoft/yii2/blob/001e4c5aed5c8e3ab93fe530bc96e80d83852208/framework/mutex/FileMutex.php#L152-L164

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.2
| Operating system | Unix via Windows host

PR #16604, opened 2018-08-03
Allow FileMutex to be explicitly told it is operating on a Windows filesystem

brandonkelly:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16603

### DIFF
--- a/framework/mutex/FileMutex.php
+++ b/framework/mutex/FileMutex.php
@@ -63,6 +63,7 @@ class FileMutex extends Mutex
      * @var bool whether file handling should assume a Windows file system.
      * This value will determine how [[releaseLock()]] goes about deleting the lock file.
      * If not set, it will be determined by checking the DIRECTORY_SEPARATOR constant.
+     * @since 2.0.16
      */
     public $isWindows;
 
--- a/framework/mutex/FileMutex.php
+++ b/framework/mutex/FileMutex.php
@@ -59,6 +59,12 @@ class FileMutex extends Mutex
      * but read-only for other users.
      */
     public $dirMode = 0775;
+    /**
+     * @var bool whether file handling should assume a Windows file system.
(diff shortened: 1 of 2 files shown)


rob006:
There is 6 other places where `DIRECTORY_SEPARATOR === '\\'` is used for Windows detection. Shouldn't we have some generic solution for this instead for patching this in one place?

brandonkelly:
Only IO operations are going to be potentially affected by #16603 (not BaseConsole::streamSupportsAnsiColors(), for example).

Even if there are similar issues elsewhere, I think it’s important that FileMutex has a localized `isWindows` setting, because its base path is configurable. You could have a Yii application where only the base mutex directory is on a Windows share, but the rest of the app is not.

So this seems like a good starting point, and if another issue arrises later, we could look into an application config setting that determines how IO operations should be handled. All that would change in FileMutex at that point is how the default `isWindows` value is determined:

```php
// old:
$this->isWindows = DIRECTORY_SEPARATOR === '\\';

// new:
$this->isWindows = Yii::$app->ioIsWindows;
```

(or something like that)

framework/mutex/FileMutex.php
```diff
@@ -59,6 +59,12 @@ class FileMutex extends Mutex
      * but read-only for other users.
      */
     public $dirMode = 0775;
+    /**
+     * @var bool whether file handling should assume a Windows file system.
+     * This value will determine how [[releaseLock()]] goes about deleting the lock file.
+     * If not set, it will be determined by checking the DIRECTORY_SEPARATOR constant.
```
samdark: Need `@since`.

brandonkelly: @samdark done

brandonkelly:
Awesome, thanks!

PR #16610, opened 2018-08-05
Fixed ErrorException xdebug trace

Izumi-kun:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


arzzen (approved):

samdark:
Closing since use case was not clarified.

samdark:
@arzzen you've approved it. Probably you know something about it?

Izumi-kun:
1. Use xdebug
2. Try to handle php error
3. In exception trace array you wish to see place with error as you see it without xdebug

```
    public function appLogic()
    {
        0 / 0;
    }

    public function actionIndex()
    {
        try {
            $this->appLogic();
        } catch (Exception $e) {
            return $e->getTraceAsString();
        }
        return 'page';
    }
```

samdark:
The fix is correct.

Initially we were handing fatals directly without first converting them to exceptions: https://github.com/yiisoft/yii2/commit/2d5db951. Because of that trace was different and we had to cut some items.

samdark:
Merged. Thank you!

PR #16629, opened 2018-08-10
ActiveField - merge options passed to active field with inputOptions

paweljankowiak06:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no

In every field except radio and checkbox options from `inputOptions` property were merged with those passed as argument, now that have same behavior as others.


samdark:
Merged. Thank you!

SilverFire:
Hello, @paweljankowiak06 

Sorry, but I had to revert the merge commit as it breaks inputs layout for projects that use Bootstrap for form styling. See the following demo: https://jsbin.com/nazegicetu/edit?html,css,output

paweljankowiak06:
Hello, @SilverFire 
I think this PR is fixing incorrect behavior of passed attributes not being merged with those existing to radio and checkbox functions. In my opinion bootstrap implementation should be adjusted to those changes. What do you think?

SilverFire:
Yes, this behavior is not consistent, but the change we've merged is a regression. Maybe we should really adjust Bootstrap implementation, but probably someone already relies on current behavior, so mind that it's a minor BC break

Issue #16636, opened 2018-08-13
ActiveForm - disabled field client side validation on radio buttons


omzy83:
### What steps will reproduce the problem?

```
<?php $form = ActiveForm::begin([
    'id' => 'register-form',
]); ?>

<fieldset disabled>
    <?php echo $form->field($model, 'name'); ?>

    <?php echo $form->field($model, 'active')->radioList([1 => 'Yes', 0 => 'No']); ?>
</fieldset>

<?php echo Html::submitButton('Submit'); ?>

<?php ActiveForm::end(); ?>
```

### What is the expected result?

It should not validate any of the input fields, since the `fieldset` has `disabled` attribute.

### What do you get instead?

The text input is not validated but the `radioList()` is validated (error message appears upon form submit - this is when using client side validation).

Same issue occurs for `checkboxList()`.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.2
| Operating system | Windows 10

omzy83:
Also I think it will still attempt validation because of the hidden input field generated by `radioList()`.

omzy83:
I have submitted a PR but it has a request for unit tests. Would appreciate any help on this :)

s1lver:
Your code does not disable validation for disabled fields. Manual testing and self-test does not show disabling validation.

omzy83:
@s1lver are you sure?

Please can you run a test as follows:

```
<?php $form = ActiveForm::begin([
    'id' => 'register-form',
]); ?>

<fieldset>
    <?php echo $form->field($model, 'name'); ?>
</fieldset>

<fieldset disabled>
    <?php echo $form->field($model, 'active')->radioList([1 => 'Yes', 0 => 'No']); ?>
</fieldset>

<?php echo Html::submitButton('Submit'); ?>

<?php ActiveForm::end(); ?>
```

In your model rules, set `name` and `active` as required. Then submit the form without entering any information. With my patch applied, there should be no error generated for the `active` field.

Without the patch, you will see the error is generated for the `active` field.

If it still doesn't work for you, please can you let me exactly how you are testing it?

s1lver:
Yii 2.0.15.1
PHP 7.2.0

It's like that.
- First I created an empty form with your example
- Make sure that the disabled items are checked
- Then I applied your patch
- I wrote the test, the test does not pass
- I checked the test several times
- Then I checked the validation manually - there are no changes with the patch
[...]

PR #16637, opened 2018-08-13
Fix radioList / checkboxList disabled field validation

omzy:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16636


samdark:
Doesn't seem to work as @s1lver found out.

PR #16644, opened 2018-08-17
Issue 16639 postgres load indexes

carono:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #16639


samdark:
Are you sure that sub-query won't affect schema reading performance significantly?

carono:
I will answer this question a little later, I will make measurements of productivity. I'll also try to make a single request

carono:
you're right, there's still something to think about about optimizing the subquery.

tables: 100
columns in each table: 20
indexes in each table: 15
queries: 10000

Current query: 17sec
With sub query: 49sec

I will have to remove the subquery, but I do not yet know how =)

samdark:
Closing since in its current form the pull request affects performance significantly.

Issue #16649, opened 2018-08-17
yii\filters\PageCache - closure support for 'variations' and 'enabled'


C4RoCKeT:
When using the yii\rest\activeController, I wanted to vary the cache based on the user. While defining the behaviors, the user identity is not yet available and thus, Yii::$app->user->id always returns null.

Support of Closures would be highly appreciated.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1

samdark:
Won't make it into 2.0.

PR #16650, opened 2018-08-17
Allow usage of Closures in yii\filters\PageCache for enabled and variations

C4RoCKeT:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Fixed issues  | #16649

### DIFF
--- a/framework/filters/PageCache.php
+++ b/framework/filters/PageCache.php
@@ -98,20 +98,23 @@ class PageCache extends ActionFilter implements DynamicContentAwareInterface
      */
     public $dependency;
     /**
-     * @var string[]|string list of factors that would cause the variation of the content being cached.
+     * @var string[]|string|\Closure[] list of factors that would cause the variation of the content being cached.
      * Each factor is a string representing a variation (e.g. the language, a GET parameter).
      * The following variation setting will cause the content to be cached in different versions
-     * according to the current application language:
+     * according to the current application language and user id.:
      *
      * ```php
      * [
+     *     function (PageCache $pageCache) {
+     *        return Yii::$app->user->id;
+     *     },
      *     Yii::$app->language,
      * ]
      * ```
      */
     public $variations;
     /**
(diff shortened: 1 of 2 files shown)


framework/filters/PageCache.php
```diff
@@ -155,7 +158,7 @@ public function init()
      */
     public function beforeAction($action)
     {
-        if (!$this->enabled) {
+        if (($this->enabled instanceof \Closure && !call_user_func($this->enabled, $this)) || !$this->enabled) {
```
rob006: This looks incorrect. If closure will return `false `, it will be used again in `!$this->enabled` (which will always evaluate to `false`). Although it may work, it looks weird and relies on a bit of luck on type juggling.

framework/filters/PageCache.php
```diff
@@ -298,6 +301,11 @@ protected function calculateCacheKey()
         if ($this->varyByRoute) {
             $key[] = Yii::$app->requestedRoute;
         }
+        foreach ($this->variations as &$variation) {
+            if ($variation instanceof \Closure) {
+                $variation = call_user_func($variation, $this);
+            }
+        }
```
rob006: I don't think that we should modify `$this->variations` directly. This will create non-obvious side effect.

framework/filters/PageCache.php
C4RoCKeT: `!call_user_func($this->enabled, $this)` so, if the closure returns `false`, it will become `true` in the statement and will not trigger the evaluation of the second statement, thus never checking `!$this->enabled`.

rob006: And what will happen if closure return `true`?

C4RoCKeT: Ah snap, you are right. Didn't think that one through.

`if($this->enabled instanceof \Closure) {
    $this->enabled = call_user_func($this->enabled, $this);
}
if (!$this->enabled) {`

This probably would be a cleaner solution.

PR #16663, opened 2018-08-23
Fixed bug where component behaviors could be double-attached

brandonkelly:
This fixes a 3.0 bug introduced in #16430.

If a component's behaviors() method returns any behaviors without a name, then cloning it would result in the behavior getting attached twice.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -311,9 +311,9 @@ class Component extends BaseObject
 
         if ($this->_behaviors !== null) {
             $behaviors = $this->_behaviors;
-            $this->_behaviors = null;
+            $this->_behaviors = [];
             foreach ($behaviors as $name => $behavior) {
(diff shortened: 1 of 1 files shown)


samdark:
The change most probably will be lost since we've switched to https://github.com/yiisoft/yii-core etc.

brandonkelly:
@samdark Ah I had missed the new repos.

Just submitted yiisoft/yii-core#17 with the same fix.

samdark:
Nope.

rob006:
It is more real than moving every change manually. Especially if you want to do this nicely, with keeping history and authorship.

samdark:
There are many changes made to the code itself so commits are mostly not merge-able.

PR #16671, opened 2018-08-28
Log level changed in yii\db\Connection::open()

chernyshev:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

Logger in yii\db\Connection::open() method should use the same log level as in yii\db\Connection::close()


samdark:
Agree but should it be debug or info?

rob006:
It should be either `debug()` everywhere, including `Command::logQuery()` or this logging should always respect `Connection::$enableLogging`. Right now it is pretty annoying, that you can't disable logging this in any way.

PR #16677, opened 2018-08-30
to correct formater dd/MM/yyyy in formatter config

VolneiMonteiroRibeiro:
problem with config 'formatter' 'dateFormat' => 'php:d/m/Y', 'datetimeFormat'=>'php:d/m/Y H:i:s' changes month for day.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -829,6 +829,10 @@ class Formatter extends Component
                 return $checkDateTimeInfo ? [$timestamp, false, true] : $timestamp;
             } elseif (($timestamp = DateTime::createFromFormat('Y-m-d H:i:s', $value, new DateTimeZone($this->defaultTimeZone))) !== false) { // try Y-m-d H:i:s format (support invalid dates like 2012-13-01 12:63:12)
                 return $checkDateTimeInfo ? [$timestamp, true, true] : $timestamp;
+            } elseif (($timestamp = DateTime::createFromFormat($this->getFormatDatePhp($this->dateFormat,$checkDateTimeInfo), $value, new DateTimeZone($this->defaultTimeZone))) !== false) {
+                return $checkDateTimeInfo ? [$timestamp, false, true] : $timestamp;
+            } elseif (($timestamp = DateTime::createFromFormat($this->getFormatDatePhp($this->timeFormat,$checkDateTimeInfo), $value, new DateTimeZone($this->defaultTimeZone))) !== false) {
+                return $checkDateTimeInfo ? [$timestamp, true, true] : $timestamp;
             }
             // finally try to create a DateTime object with the value
             if ($checkDateTimeInfo) {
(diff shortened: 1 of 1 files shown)


cebe:
What you change here is the function that reads the value that is to be formatted. This should be a machine readable and unambiguous format. dd/mm/yyyy does not fulfil this criteria as there is no clear distinction from mm/dd/yyyy. You must use an unambiguous format for storing dates in the database.

VolneiMonteiroRibeiro:
no cebe, my storing dates in the database in this format, i think that my pull request rashly

VolneiMonteiroRibeiro:
cebe, yii no apply my config formatDate=>'dd/MM/yyyy' interpreting wrong way,when load post! i see this as a bug. if the date load from my db, "yyyy-MM-dd H:i:s" is interpreted correct.
you can open my pull request, for a better analysis, or while we dictate?

rob006:
@VolneiMonteiroRibeiro You could start from writing a unit test for this bug/feature - right now it is not clear what you're trying to achieve with this PR.

cebe:
@VolneiMonteiroRibeiro your change is incorrect, and I do not see a good way to improve it to a proper solution, that's why I closed the pull request. You are using the property that configures the output format for parsing the input format. This does not make sense and changes the way the formatter works completely.

If we'd want to introduce a configuration for the input format we should discuss that in a separate issue. However I do not think there are reasonable storage formats to add so that we'd need such a configuration. We currently have support for storage of datetime values as unixtimestamp (integer) or in the  `yyyy-mm-dd`, which correspond to the format MySQL datetime fields are using. I am not aware of other formats being used commonly.

VolneiMonteiroRibeiro:
Showing "01/05/2018"
If 
formatter->timeZone = 'America/Sao_Paulo'; \Yii::$app->formatter->defaultTimeZone = 'America/Sao_Paulo'; \Yii::$app->formatter->dateFormat = 'dd/MM/yyyy'; echo \Yii::$app->formatter->asTimestamp('05/01/2018');
Is wrong, agree?

rob006:
I agree with @cebe - you're configuring formatter **output** format, and expecting that it will be used for parsing **input**. If you have dates in such format, you need to normalize it yourself.

VolneiMonteiroRibeiro:
@rob006 
if i want use the method asTimestamp?
this config stay in web.php:formatter->timeZone = 'America/Sao_Paulo'; \Yii::$app->formatter->defaultTimeZone = 'America/Sao_Paulo'; \Yii::$app->formatter->dateFormat = 'dd/MM/yyyy'; 

if i use \Yii::$app->formatter->asTimestamp('05/01/2018'); wrong out.
/**
     * Formats a date, time or datetime in a float number as UNIX timestamp (seconds since 01-01-1970).
     * @param int|string|DateTime $value the value to be formatted. The following
     * types of value are supported:
     *
     * - an integer representing a UNIX timestamp
     * - a string that can be [parsed to create a DateTime object](http://php.net/manual/en/datetime.formats.php).
     *   The timestamp is assumed to be in [[defaultTimeZone]] unless a time zone is explicitly given.
     * - a PHP [DateTime](http://php.net/manual/en/class.datetime.php) object
     *
     * @return string the formatted result.
     */
    public function asTimestamp($value)

VolneiMonteiroRibeiro:
**- a string that can be parsed to create a DateTime object.**
dont parse correct my string! it is contradictory with the comment, how can it be correct? my friends!

cebe:
> a string that **can** be parsed to create a DateTime object.

your string is ambiguous so it **can not** be parsed correctly. Use another format. The Yii formatter accepts unix timestamps (integer) or YYYY-MM-DD.

VolneiMonteiroRibeiro:
Stop the ego @cebe analise the situactions.... you do not correct, its ugly, ope  my requst

Issue #16524, opened 2018-07-16
Wrong order of parameters in @var in view files


springimport:
> ### @var
> You may use the @var tag to document the “Type” of properties, sometimes called class variables.
> ### Syntax
> @var [“Type”] [$element_name] [<description>]
> ### Examples
> ```php
> class Foo
> {
>   /**
>    * @var string $name        Should contain a description
>    * @var string $description Should contain a description
>    */
>   protected $name, $description;
> }
> ```

[phpDocumentor](http://docs.phpdoc.org/references/phpdoc/tags/var.html)

### What steps will reproduce the problem?
Open any view file in framework.

### What is the expected result?
```php
/* @var \yii\web\View $this */
/* @var \yii\gii\generators\module\Generator $generator */
```

### What do you get instead?
```php
/* @var $this yii\web\View */
/* @var $generator yii\gii\generators\module\Generator */
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1

samdark:
Because it's not phpdoc, it's vardoc.

springimport:
@samdark Okay, but what about moving to phpdoc syntax?

rob006:
@springimport  At least some time ago vardoc has better IDE support. And it is still not clear whether using phpdoc in such way is valid - phpdoc should proceed structural element, which does not exist in this case.

springimport:
@rob006
`\yii\bootstrap\Nav`, phpdoc:
```php
/** @var Widget $dropdownClass */
```
`yii\db\ActiveQuery`, vardoc:
```php
/* @var $modelClass ActiveRecord */
```
`yii\console\controllers\HelpController`, ??:
```php
/** @var $controller Controller */
```

samdark:
As @rob006 said, neither is phpdoc. Phpdoc can't be applied to inline variables according to its specifications.

springimport:
@samdark I think it will be good to have single format in yii.

samdark:
Yes. Should be `/* @var $modelClass ActiveRecord */` but technically vardoc allows both. If you have some free time to burn, feel free to submit a pull request.

PR #16691, opened 2018-09-08
Refactor phpdoc to vardoc #16524

iproger:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16524

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -170,7 +170,7 @@ class UniqueValidator extends Validator
      */
     private function modelExists($targetClass, $conditions, $model)
     {
-        /** @var ActiveRecordInterface $targetClass $query */
+        /* @var $targetClass $query ActiveRecordInterface */
         $query = $this->prepareQuery($targetClass, $conditions);
 
         if (!$model instanceof ActiveRecordInterface
@@ -263,7 +263,7 @@ class UniqueValidator extends Validator
             return $conditions;
         }
 
-        /** @var ActiveRecord $targetModelClass */
+        /* @var $targetModelClass ActiveRecord */
         return $this->applyTableAlias($targetModelClass::find(), $conditions);
     }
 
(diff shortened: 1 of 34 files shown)


framework/validators/UniqueValidator.php
```diff
@@ -170,7 +170,7 @@ private function getTargetClass($model)
      */
     private function modelExists($targetClass, $conditions, $model)
     {
-        /** @var ActiveRecordInterface $targetClass $query */
+        /* @var $targetClass $query ActiveRecordInterface */
```
rob006: This should be split into two blocks, separate for `$targetClass` and `$query`?

iproger: I can remove $targetClass from vardoc and add to @param string|ActiveRecordInterface $targetClass.

rob006: That would be incorrect - `$targetClass` needs to be a string.

iproger: No, look at 131 and 137 lines.
```php
$db = $targetClass::getDb();

...

$db->useMaster(function () use ($targetClass, $conditions, $model, &$modelExists) {
    $modelExists = $this->modelExists($targetClass, $conditions, $model);
});
```

$targetClass can be string or ActiveRecordInterface.

rob006: No, it cannot:

```php
get_class($model) !== $targetClass
```

This will always return `true` if you pass object instead of class name as string.

iproger: I don't see a problem if modelExists() docblock will contain 2 types. But we can separate the annotation into 2 lines or just remove it.

rob006: > I don't see a problem if modelExists() docblock will contain 2 types.

The problem is pretty obvious for me - method will not work correctly if you pass `ActiveRecordInterface` instance as `$targetClass`, so adding `ActiveRecordInterface` to phpdoc is misleading.

samdark:
Closing since it can't me merged as is.

PR #16702, opened 2018-09-10
Facility for DataProvider-oriented custom template rendering through REST APIs

codespede:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

Facility for DataProvider-oriented custom template rendering through REST which features the following:
- Pagination Headers will be present in the Response Headers just as how it will be present when returning a DataProvider - which helps the client to navigate through the pages.
- Allows pagination, sorting, filtering and all other operations supported by DataProvider.

### Use Cases
 - Suppose they want to deliver the data in CSV like below:
```
title,image,content
ABC,abc.jpg,Content of ABC
MNO,mno.jpg,Content of MNO
XYZ,xyz.jpg,Content of XYZ
```
 - Suppose one has to render the API response in a specific format like below:
```
-begin-
--title=ABC
--image=abc.jpg
--content=Content of ABC
---
--title=MNO
--image=mno.jpg
--content=Content of MNO
---
--title=XYZ
--image=xyz.jpg
--content=Content of XYZ
-end-
```
 - Or in any situation where the user has to deliver custom formatted data through the API.

In all the above cases, the data can be paginated, sorted and filtered.

One can use this by simply returning the TemplateRenderer object as shown in the below code:
```
public function actionTest(){
    $dataProvider = new ActiveDataProvider(['query' => Model::find()->where($condition)])
    return new \yii\rest\TemplateRenderer([
            'parentView' => '/path/to/parent-view', //path to the parent or wrapper view file
            'itemView' => '/path/to/item-view', //path to the item view file
        ]);
}
```

### DIFF
--- a/framework/rest/Serializer.php
+++ b/framework/rest/Serializer.php
@@ -137,7 +137,7 @@ class Serializer extends Component
      * Serializes the given data into a format that can be easily turned into other formats.
      * This method mainly converts the objects of recognized types into array representation.
      * It will not do conversion for unknown object types or non-object data.
-     * The default implementation will handle [[Model]] and [[DataProviderInterface]].
+     * The default implementation will handle [[Model]], [[DataProviderInterface]] and [[TemplateRenderer]].
      * You may override this method to support more object types.
      * @param mixed $data the data to be serialized.
      * @return mixed the converted data.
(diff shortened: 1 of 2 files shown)


samdark:
Interesting concept but shouldn't we render on client instead? `ResponseFormatterInterface` already exists for the purpose of custom serialization/rendering of API data.

codespede:
Thought this will be a more easier and quicker solution for the developers because, it is ready-to-use as the new format is ready by requiring only the parent and child template and they don't have to create a new ResponseFormatter class for it.

samdark:
I prefer not having two different solutions for a single purpose in this case.

PR #16709, opened 2018-09-13
Fix __unset() method description [skip ci]

CedricYii:
Fix description as this inherited __unset() method does not set property to null but really unset an attribute or a relation.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -343,10 +343,10 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
     }
 
     /**
-     * Sets a component property to be null.
+     * Unsets a component property.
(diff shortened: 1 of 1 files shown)


rob006:
It does not unset property - for example you can still access attribute after unsetting it:

```php
$model = new Item();
unset($model->name);

var_dump($model->name);
```
It will return `null`, so it works like property still exist and its value is `null`. So at least for attributes `unset($model->attribute)` works like `$model->attribute = null`.

CedricYii:
That's true. But it is a generated value only. If you do a 
````php
var_dump($model);
````
you see that _attributes[name] does not exist anymore.

How would you word it? Unset a attribute value?

Issue #14368, opened 2017-06-29
W3C Validation Errors on ActifeField radioList


nagyt234:
I've added to the latest basic template with yii 2.0.12 to the LoginForm model a simple required attribute:
```
...
    public $gender;
...
    [['username', 'password', 'gender'], 'required'],
```
and a radioList imput field to the login form (I've removed the fieldsConfig option from ActiveForm):
```
...
   <?php $form = ActiveForm::begin([
        'id' => 'login-form',
        'layout' => 'horizontal',
    ]); ?>
...
    <?= $form->field($model, 'gender')->label('Gender')->radioList([1 => 'male', 2 => 'female']) ?>
...
```
This simple input generates two HTML validation error using W3C validation:
```
Error: Element div is missing one or more of the following attributes: aria-expanded, aria-valuemax, aria-valuemin, aria-valuenow, role.
From line 44, column 56; to line 44, column 103
  value=""><div id="loginform-gender" aria-required="true"><div c
Attributes for element div:
  Global attributes
```
```
Error: The for attribute of the label element must refer to a non-hidden form control.
From line 42, column 1; to line 42, column 61
equired">↩<label class="control-label col-sm-3" for="loginform-gender">Gender
```
Note that the second error comes up only if the `label` method is explicitly using.

I would expect, that simple forms and fields generates HTML without validation errors.

Note also, that there are also 2 warning,associated with the navigation and the form:
```
Warning: The navigation role is unnecessary for element nav.
From line 16, column 5; to line 16, column 82
rap">↩    <nav id="w0" class="navbar-inverse navbar-fixed-top navbar" role="navigation"><div c
```
```
Warning: The form role is unnecessary for element form.
From line 31, column 5; to line 31, column 111
</p>↩↩    <form id="login-form" class="form-horizontal" action="/index.php?r=site%2Flogin" method="post" role="form">↩<inpu
```
### Additional info
| Yii version      | 2.0.12
| PHP version      |  7.1.5
| Operating system | Ubuntu 16.04

berosoboy:
This issue is mainly related to yii2-bootstrap.

1. To solve the `aria-required="true"` problem, the method `\yii2\bootstrap\ActiveField::radioList()` could add the` role="radiogroup"` attribute. I can submit a PR if that's ok:
```php
    public function radioList($items, $options = [])
    {
        if ($this->model->isAttributeRequired($this->attribute) && !isset($options['role'])) {
            $options['role'] = 'radiogroup';
[...]

PR #16713, opened 2018-09-13
Added role attributes to the input options (#14368)

s1lver:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #14368

I think it makes sense to add not only to RadioList. What do you think?


samdark:
Merged. Thank you!

PR #16719, opened 2018-09-16
Added info in nested modules docs

SOHELAHMED7:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | -
| Fixed issues  | no

Added info in nested modules docs regarding retrieval of all  parent modules of a module


SOHELAHMED7:
I thought it would be handy.
Or if not necesaary there, I can add it in Users note section at the bottom of the page

samdark:
I don't think these facts are important to state in the guide.

PR #16722, opened 2018-09-17
Pressing "Enter" key  does not depend on the option "filterOnFocusOut"

s1lver:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #16716


s1lver:
hmm, there is already a test for this behavior

samdark:
If there's already a test and it was passing then either the test is wrong or there's nothing to fix.

s1lver:
I will check it

s1lver:
Added tests for three behaviors

default

samdark (approved):
Looks good to me. Please add a line for changelog.

samdark:
Merged. Thanks!

PR #16732, opened 2018-09-21
Fix submit form

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes/no
| Breaks BC?    | yes/no
| Tests pass?   | yes/no
| Fixed issues  | #15791

### DIFF
--- a/framework/assets/yii.js
+++ b/framework/assets/yii.js
@@ -57,7 +57,7 @@ window.yii = (function ($) {
          * The selector for clickable elements that need to support confirmation and form submission.
          */
         clickableSelector: 'a, button, input[type="submit"], input[type="button"], input[type="reset"], ' +
-        'input[type="image"]',
+            'input[type="image"]',
         /**
          * The selector for changeable elements that need to support confirmation and form submission.
          */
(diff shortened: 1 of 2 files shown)


rustamwin:
#16557

framework/assets/yii.js
```diff
@@ -162,6 +162,10 @@ window.yii = (function ($) {
                 pjaxContainer,
                 pjaxOptions = {};
 
+            if (params !== undefined && params.hasOwnProperty('submit')) {
+                console.error("Property 'submit' of object #<HTMLFormElement> is not a function.");
```
samdark: This error doesn't give you any clue on what to do to fix it thus it's not giving end user much.

rustamwin: Any ideas?

samdark: 1. The error should say `Parameter name "submit" conflicts with a same named form property. Please use another name.`.
2. As it was described in original issue, the same is true for other elements:

> Forms and their child elements should not use input names or ids that conflict with properties of a form, such as submit, length, or method.

So it may make sense to expand condition.

samdark (approved):

samdark:
Merged. Thank you!

PR #16735, opened 2018-09-23
Remove redundant changelog line [skip ci]

rob006:
After #16734 there are two lines for the same change.


samdark:
Merged. Thanks!

PR #16756, opened 2018-10-01
[CS Fix] Removes redundant else statement in BaseObject

greeflas:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

Just small fix of CS in BaseObject.


rob006:
It is not redundant, first branch of this `if` does not throw or return anything.

Issue #16763, opened 2018-10-02
Unable to load dynamic library pdo_sqlsrv.so


marcelodeandrade:
### What steps will reproduce the problem?
Running tests with docker: 
`cd yii2/tests/mssql`
`docker-compose run --rm php vendor/bin/phpunit -v --group mssql`
### What is the expected result?
Load `pdo_sqlsrv.so`
### What do you get instead?
```
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php/20151012/pdo_sqlsrv.so' - /usr/lib/php/20151012/pdo_sqlsrv.so: undefined symbol: php_pdo_register_driver in Unknown on line 0
PHPUnit 4.8.34 by Sebastian Bergmann and contributors.

Runtime:	PHP 7.0.32-0ubuntu0.16.04.1 with Xdebug 2.4.0
Configuration:	/project/phpunit.xml.dist

```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16-dev
| PHP version      | 7.2.10
| Operating system | ubuntu 18.04.1
| MSSQL version | 2017
| MSSQL driver | sqlsrv

PR #16764, opened 2018-10-02
 Fix Unable to load dynamic library pdo_sqlsrv.so

marcelodeandrade:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16763

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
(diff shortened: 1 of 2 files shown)


framework/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #16763: Fix Unable to load dynamic library pdo_sqlsrv.so (marcelodeandrade)
```
rob006: Please do not add changelog entries about tests/documentation fixes - changelog is about changes in framework, anything else is only a noise from end user perspective.

Issue #14759, opened 2017-09-05
Error return when using JSON_FORMAT with NULL value


xpader:
Invoke `return $this->asJson(null);` in controller
or
`
$response->format = Response::FORMAT_JSON;
$response->data = null;
`

will output nothing.

In standard json, the null value must be output a 'null' string.
But when I set output format to json, and value is null, it output nothing.
I think the right output is a 'null' string, hope fixed.

cebe:
this is expected according to the docs of `Response::$data`: https://github.com/yiisoft/yii2/blob/9b01ca275f891c245c2e840ed93c87f715c04e70/framework/web/Response.php#L125-L126 what is the use case of returning null from a controller?

rob006:
@cebe I think that use case is irrelevant here - empty string is invalid JSON. "null response" should return `null` encoded as JSON.

```
>> JSON.parse('')
SyntaxError: JSON.parse: unexpected end of data at line 1 column 1 of the JSON data
>> JSON.parse('null')
null
```

https://3v4l.org/R4mRJ

tom--:
Might this not be a breaking change?

Strict correctness of the JSON encoding isn't necessarily the requirement if Yii, according to its documentation, specifies that a PHP `null` means "don't output anything". People might be using it.

samdark:
Hmm... you're right. It may be used.

rob006:
> People might be using it.

For what? Empty JSON is invalid response. You can almost always say "we can not fix this bug, because people may rely on it".

xpader:
@tom-- But people has use `\Yii::$app->response->format = Response::FORMAT_JSON;`
That mean wan't output as json format. If output nothing, that format must not be JSON.

cebe:
the problem here is the duplicate meaning of `null`, it could be a valid JSON value OR mean that property `data` should not be converted to populate property `content`.

cebe:
To fix this for `Controller::asJson()` we could add something like

```php
$response->format = Response::FORMAT_JSON;
if ($data === null) {
    $response->data = null;
    $response->content = 'null';
} else {
    $response->data = $data;
}
```

misantron:
@cebe I don't think we should modify `content` directly in `Controller::asJson()`.
The better solution is to change `JsonResponseFormatter` behavior like this:
```
protected function formatJson($response)
    {
        if ($response->data !== null) {
            $options = $this->encodeOptions;
            if ($this->prettyPrint) {
[...]

PR #16772, opened 2018-10-05
Json response with null data

misantron:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14759


samdark (approved):
Looks good to me. Please add a line for changelog.

framework/web/JsonResponseFormatter.php
```diff
@@ -117,6 +117,8 @@ protected function formatJson($response)
                 $options |= JSON_PRETTY_PRINT;
             }
             $response->content = Json::encode($response->data, $options);
+        } else {
+            $response->content = 'null';
```
cebe: content should not be overwritten, if data is null, your change contradicts the documentation...

cebe: this seems to be fixed by https://github.com/yiisoft/yii2/commit/3c091b802a03730c3591eae1d51cceebf88aadde

PR #16774, opened 2018-10-06
Fixes #16766: ContentNegotiator now adds Vary response headers

### LINKED ISSUE #16766: ContentNegotiator should set Vary header

Use ContentNegotiator and HttpCache behaviors together in a controller.

Related to yiisoft-contrib/yiiframework.com#403

Get cached data stored in browser cache for each Content-Type variant.

### What do you get instead?

Cached data does not vary based on requested content type.

Browser caches and other HTTP proxies do not seem to store different variants of the HTTP body for different Content-Types by default. Adding the `Vary: Accept` header fixes this.

To solve this problem the `ContentNegotiator` should set the `Vary: Accept` header by default.

### Additional info

https://stackoverflow.com/questions/1975416/what-is-the-function-of-the-vary-accept-http-header

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15
| PHP version      | all

koteq:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no (it shouldn't)
| Tests pass?   | yes
| Fixed issues  | #16766


samdark (approved):
Looks good. Please add a line for changelog. Thanks.

samdark:
Merged. Thank you!

Issue #16424, opened 2018-06-20
Transaction Class Rollback Problem


lenyy:
Hello guys,

So by inspecting your [Transaction](https://github.com/yiisoft/yii2/blob/master/framework/db/Transaction.php) in the methods begin and rollback, im not sure if what you have is a bug or by design of the framework.
But in this piece of code from the method [begin](https://github.com/yiisoft/yii2/blob/master/framework/db/Transaction.php#L114)
```php
$schema = $this->db->getSchema();
if ($schema->supportsSavepoint()) {
            Yii::debug('Set savepoint ' . $this->_level, __METHOD__);
            $schema->createSavepoint('LEVEL' . $this->_level);
        } else {
            Yii::info('Transaction not started: nested transaction not supported', __METHOD__);
        }
```
you don't throw an Exception, but in the piece of code from [rollback](https://github.com/yiisoft/yii2/blob/master/framework/db/Transaction.php#L175)
```php
$schema = $this->db->getSchema();
        if ($schema->supportsSavepoint()) {
            Yii::debug('Roll back to savepoint ' . $this->_level, __METHOD__);
            $schema->rollBackSavepoint('LEVEL' . $this->_level);
        } else {
            Yii::info('Transaction not rolled back: nested transaction not supported', __METHOD__);
            // throw an exception to fail the outer transaction
            throw new Exception('Roll back failed: nested transaction not supported.');
        }
```
you throw an Exception if the schema doesn't support save point.
The first Question is:
Shouldn't the begin method throw the exception if SavePoint is not enabled?

Well because this leads to unexpected behavior by the IDE when i try to catch the exceptions that are thrown by the transaction like this 
![screenshot_1](https://user-images.githubusercontent.com/10136403/41656942-8a9bde22-7489-11e8-8aaf-f7059aa2529b.png)

To me this feels really weird because we caught  an exception we should rollback but we need to have another try catch to catch the exception that the rollback might throw.

The solution would be to throw that exception in the method [begin](https://github.com/yiisoft/yii2/blob/master/framework/db/Transaction.php#L175) because we should not start a transaction that doesn't support SavePoint, at least by following your logic.

Any help you need just say.
Cheers Pedro Rosa

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | Master
| PHP version      |  Doesn't Matter
| Operating system | Doesn't Matter

samdark:
Yes. It should throw as well, I think.

lenyy:
[...]

PR #16807, opened 2018-10-20
Savepoint unsupported detection in transaction begin

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | should
| Fixed issues  | #16424

I agree with @lenyy that throwing exception in rollback is not needed since it will be thrown in `Transaction::begin()`.


lenyy:
wooww n1 @bizley thx dude.
Great job and now you are going to save me a lot of unnecessary try catch blocks :joy:
Cheers to you and @samdark for making it quick with the pull request

cebe:
Which database systems would be affected by this? I assume this could result in a similar number of issue reports as the new exception thrown when headers where already sent if it affects a lot of systems.

cebe:
> Savepoints are defined in the SQL standard and are supported by all established SQL relational databases, including PostgreSQL, Oracle Database, Microsoft SQL Server, MySQL, DB2, SQLite (since 3.6.8), Firebird, H2 Database Engine, and Informix (since version 11.50xC3). 
> https://en.wikipedia.org/wiki/Savepoint

so it seems that is a rare case :)

bizley:
Yes, and on top of it exception only is raised when nested transaction is started so single level ones are still fine.

PR #16827, opened 2018-10-26
Fix command::executeResetSequence() quote table and column names

CedricYii:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Keeps quoting logic of the framework for table and column names

### DIFF
--- a/framework/db/oci/QueryBuilder.php
+++ b/framework/db/oci/QueryBuilder.php
@@ -157,15 +157,18 @@ EOD;
             // use master connection to get the biggest PK value
             $value = $this->db->useMaster(function (Connection $db) use ($tableSchema) {
                 return $db->createCommand(
-                    'SELECT MAX("' . $tableSchema->primaryKey[0] . '") FROM "'. $tableSchema->name . '"'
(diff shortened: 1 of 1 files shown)


CedricYii:
Quoting Table and Column Names

When writing database-agnostic code, properly quoting table and column names is often a headache because different databases have different name quoting rules. To overcome this problem, you may use the following quoting syntax introduced by Yii:

    [[column name]]: enclose a column name to be quoted in double square brackets;
    {{table name}}: enclose a table name to be quoted in double curly brackets.

Yii DAO will automatically convert such constructs into the corresponding quoted column or table names using the DBMS specific syntax.
https://www.yiiframework.com/doc/guide/2.0/en/db-dao#quoting-table-and-column-names

rob006:
But this is query builder for specific database, why do you want to use database-agnostic  approach in database-specific context?

samdark:
As @rob006 said, that's specific database type you've changed it in. Do you intend to solve a problem you have? If so, which one?

samdark:
Closing since initial problem was not clear.

PR #16830, opened 2018-10-26
Fixed issue where console table widget throws division by zero

### LINKED ISSUE #15552: \yii\console\widgets\Table::widget problem with screenwidth
use console widget table when screen size __can't__ be determined
```php
print '<pre>';
print \yii\console\widgets\Table::widget([
          'headers' => ['a','b'],
          'rows'    => [['a','b']]
        ]);
```

image

### What do you get instead?
image

### Additional info
yii2\console\widgets\Table.php::getScreenWidth() doesn't assign <b>$this->_screenWidth</b> when __Console::getScreenSize()__ return false.
for you'r information on my system : 
```bash
stty -a 2>&1
```
produce :
```bash
speed 38400 baud; rows 55; columns 207; line = 0;
intr = ^C; quit = ^\; erase = ^?; kill = ^U; eof = ^D; eol = <undef>; eol2 = <undef>; swtch = <undef>; start = ^Q; stop = ^S; susp = ^Z; rprnt = ^R; werase = ^W; lnext = ^V; flush = ^O; min = 1; time = 0;
-parenb -parodd -cmspar cs8 -hupcl -cstopb cread -clocal -crtscts
-ignbrk -brkint -ignpar -parmrk -inpck -istrip -inlcr -igncr icrnl ixon -ixoff -iuclc -ixany -imaxbel -iutf8
opost -olcuc -ocrnl onlcr -onocr -onlret -ofill -ofdel nl0 cr0 tab0 bs0 vt0 ff0
isig icanon iexten echo echoe echok -echonl -noflsh -xcase -tostop -echoprt echoctl echoke
```

so when __yii2\console\widgets\Table.php: calculateRowsSize()__  do :
```php
      $screenWidth = $this->getScreenWidth() - 3;
```
it give a negative value for __$screenWidth__ and produce the php error.

#### Set a default value for screenWidth:
```php
    /**
     * Getting screen width.
     *
     * @return int screen width
     */
    protected function getScreenWidth()
    {
        if (!$this->_screenWidth) {
            $size = Console::getScreenSize();
            if (isset($size[0])) {
                $this->_screenWidth = $size[0];
            }else{
              $this->_screenWidth=1000;
            }
        }

        return $this->_screenWidth;
    }
```
Will produce :
image

#### Only Fix the php error
```php
    /**
     * Calculate the size of rows to draw anchor of columns in console.
     *
     * @see \yii\console\widgets\Table::render()
     */
    protected function calculateRowsSize()
    {
        $this->_columnWidths = $columns = [];
        $totalWidth = 0;
        $screenWidth = $this->getScreenWidth() - 3;
        if ($screenWidth < 0){
          $screenWidth=0;
        }

```
will produce : 
image

ugly but no error

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13
| PHP version      | 7.1
| Operating system | Alpine Linux

nadar:
Fixed issue where console table widget throws division by zero exception when unable to determine screen height/width. closes #15552 

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15552


nadar:
As described in the issue, its actually very hard to indicate the real problem, so at least this fix won't change the core behavior but ensures there is no division by zero exception. Maybe there will be further problems when using in dockerized environments, but we will see. If there are any further problems, i will create a new issue and send a new PR.

cebe:
The issue is about not detecting the screen width. Why do you add a fallback height?

we could add another check to `getScreenWidth()` to return default value if screen width is <= 0

cebe:
Seems this error occurs if it is not possible to render a table in the given screen width, so it depends on the number of columns as well as the width of the screen.

```php
    public function testTableTooSmallScreen()
    {
        $table = new Table();

        // test fullwidth chars
        // @see https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms
        $expected = <<<EXPECTED
╔══════════════╗
║ testcontent1 ║
║ testcontent2 ║
║ testcontent3 ║
╚══════════════╝

EXPECTED;

        $this->assertEqualsWithoutLE($expected, $table
            ->setRows([
                [Console::ansiFormat('testcontent1testcontent2testcontent3', [Console::FG_BLUE])],
            ])->setScreenWidth(2)->run()
        );
    }
```

this test fails with 

```
1) yiiunit\framework\console\TableTest::testTableTooSmallScreen
str_repeat(): Second argument has to be greater than or equal to 0

/home/cebe/dev/yiisoft/yii2/framework/console/widgets/Table.php:320
/home/cebe/dev/yiisoft/yii2/framework/console/widgets/Table.php:183
/home/cebe/dev/yiisoft/yii2/tests/framework/console/widgets/TableTest.php:345
/home/cebe/dev/yiisoft/yii2/vendor/phpunit/phpunit/phpunit:52
```

cebe:
could be fixed by applying this patch:

```diff
diff --git a/framework/console/widgets/Table.php b/framework/console/widgets/Table.php
index 1c68883..917f0cb 100644
--- a/framework/console/widgets/Table.php
+++ b/framework/console/widgets/Table.php
@@ -363,6 +363,9 @@ class Table extends Widget
         if ($totalWidth > $screenWidth) {
             foreach ($this->_columnWidths as $j => $width) {
                 $this->_columnWidths[$j] = (int) ($width * $relativeWidth);
+                if ($this->_columnWidths[$j] < 3) {
+                    $this->_columnWidths[$j] = 3;
+                }
                 if ($j === count($this->_columnWidths)) {
                     $this->_columnWidths = $totalWidth;
                 }
```

to `calculateRowsSize()` to ensure column has at least 1 char + 2 whitespace width to be able to render content.

nadar:
> The issue is about not detecting the screen width. Why do you add a fallback height?

Cause i was asking if you would accept a PR which at least fixes this problem with division by zero exception. But yes indeed, i agree with you this won't fix the actual problem. I was also unsure if the proposed fix from @vinpel mentioned in the issue was already done in the code (https://github.com/yiisoft/yii2/issues/15552#issue-291201959) (Which looks the same to me here: https://github.com/yiisoft/yii2/blob/master/framework/console/widgets/Table.php#L389-L395) - maybe i miss understood something.

Now we need also a test which throws the division by zero exception, so i will fix both things accordingly.

ps: For me it was also very hard to understand the table widget code, cause its accessing private properties in different part of the class, which means "it assumes" given methods has been run. Maybe i could also refractor this code parts (https://github.com/yiisoft/yii2/blob/master/framework/console/widgets/Table.php#L306-L308)

nadar:
@cebe indeed, its really just a problem about screen size. If the table is longer then the window, the exception is thrown, this is quite often the case.

schmunk42:
I am also affected by this(?) division by zero bug.

Strange thing is, that it just occurs on a specific range of window widths, meaning that smaller and larger widths are working.

---

Example data & code to create a test:

```
array(3) {
  [0]=>
  array(3) {
    ["id"]=>
    string(1) "1"
    ["model"]=>
    string(27) "hrzg\papi\adapters\Basecamp"
    ["config_json"]=>
    string(104) "{
    "username": "api-user-**@****.de",
    "password": "*************",
    "user_id": "1****00"
}"
  }
  [1]=>
  array(3) {
    ["id"]=>
    string(1) "2"
    ["model"]=>
    string(25) "hrzg\papi\adapters\GitLab"
    ["config_json"]=>
    string(84) "{
    "url": "https://git.****.**/api/v4/",
    "token": "********************"
}"
  }
  [2]=>
  array(3) {
    ["id"]=>
    string(1) "3"
    ["model"]=>
    string(25) "hrzg\papi\adapters\GitHub"
    ["config_json"]=>
    string(123) "{
    "token": "****************************************",
    "users": ["*********"],
    "organizations": ["*****"]
}"
  }
}
```

```
        $headers = array_keys($data[0]);
        $table = \yii\console\widgets\Table::widget(['headers' => $headers, 'rows' => $data]);
        $this->stdout($table);
```

Widths working: 129, 62
Widths broken (division by zero): 92, 103

samdark:
@schmunk42 do you want to take over it?

samdark:
Nope, he won't have time.

samdark:
The solution presented here won't fix the problem as it should be fixed as @cebe mentioned.

PR #16836, opened 2018-10-27
Fix MySQL lock name length limit

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

> MySQL 5.7.5 and later enforces a maximum length on lock names of 64 characters. Previously, no limit was enforced.

See https://github.com/symfony/symfony/pull/27250


samdark:
Merged. Thank you!

rob006:
> Would it be safer to only hash if length is too long?

That still would be BC break if someone uses MySQL <5.7.5 and has lock name longer than 64 characters - it will be hashed even if there is no need for it.

IMO such condition will add unnecessary complicity and inconsistency. `hashLockName()` is protected, if someone relies on lock name, he could always override it.

PR #16837, opened 2018-10-27
Add `$message` parameter to `TestCase::assertEqualsWithoutLE()`

rob006:
This allows to set custom error message in `assertEqualsWithoutLE()`

This third parameter is already used in some tests, even if it was not supported until now :P:

https://github.com/yiisoft/yii2/blob/ba0a3c00ca36a26f2f34b09d31230e4d83bd831e/tests/framework/mail/BaseMailerTest.php#L273-L274


samdark:
Merged. Thanks!

PR #16838, opened 2018-10-27
[Bug] MysqlMutexTest and PgsqlMutexTest allows to acquire the same lock multiple times

rob006:
I'm not sure if this is a bug, but it is quite surprising and a little scary that I can acquire the same lock multiple times if I use the same component (and the same DB connection).

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -78,6 +78,7 @@ Yii Framework 2 Change Log
 - Bug #15204: `yii\helpers\BaseInflector::slug()` is not removing substrings matching provided replacement from given string anymore (bizley)
 - Bug #16101: Fixed Error Handler to clear registered meta tags, link tags, css/js scripts and files in error view (bizley)
 - Bug #16836: Fix `yii\mutex\MysqlMutex` to handle locks with names longer than 64 characters (rob006)
+- Bug #16838: `yii\mutex\Mutex::acquire()` no longer returns `true` if lock is already acquired by the same component in the same process (rob006)
 - Enh #16839: Increase frequency of lock tries for `yii\mutex\FileMutex::acquireLock()` when $timeout is provided (rob006)
 - Enh #16839: Add support for `$timeout` in  `yii\mutex\PgsqlMutex::acquire()` (rob006)
 - Bug #16828: `yii\console\controllers\MessageController::translator` recognized object' methods and functions calls as identical sets of tokens (erickskrauch)
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -62,6 +62,17 @@ Upgrade from Yii 2.0.15
   ```json
   "cebe/markdown": "~1.1.0",
   ```
+  
+* `yii\mutex\Mutex::acquire()` no longer returns `true` if lock is already acquired by the same component in the same process.
+  Make sure that you're not trying to acquire the same lock multiple times in a way that may create infinite loops, for example:
+    
+  ```php
+  if (Yii::$app->mutex->acquire('test')) {
+       while (!Yii::$app->mutex->acquire('test')) {
+           // `Yii::$app->mutex->acquire('test')` will always return `false` here, since lock is already acquired
+      }
+  }
+  ```
 
 
 Upgrade from Yii 2.0.14
--- a/framework/mutex/Mutex.php
+++ b/framework/mutex/Mutex.php
@@ -69,7 +69,7 @@ abstract class Mutex extends Component
      */
     public function acquire($name, $timeout = 0)
     {
-        if ($this->acquireLock($name, $timeout)) {
+        if (!in_array($name, $this->_locks, true) && $this->acquireLock($name, $timeout)) {
             $this->_locks[] = $name;
 
             return true;
--- a/tests/framework/mutex/MutexTestTrait.php
+++ b/tests/framework/mutex/MutexTestTrait.php
@@ -46,10 +46,11 @@ trait MutexTestTrait
 
         $this->assertTrue($mutexOne->acquire($mutexName));
         $this->assertFalse($mutexTwo->acquire($mutexName));
-
-        $mutexOne->release($mutexName);
+        $this->assertTrue($mutexOne->release($mutexName));
+        $this->assertFalse($mutexTwo->release($mutexName));
 
         $this->assertTrue($mutexTwo->acquire($mutexName));
+        $this->assertTrue($mutexTwo->release($mutexName));
     }
 
     /**
--- a/tests/framework/mutex/MutexTestTrait.php
+++ b/tests/framework/mutex/MutexTestTrait.php
@@ -41,8 +41,8 @@ trait MutexTestTrait
      */
     public function testThatMutexLockIsWorking($mutexName)
     {
-        $mutexOne = $this->createMutex($mutexName);
-        $mutexTwo = $this->createMutex($mutexName);
+        $mutexOne = $this->createMutex();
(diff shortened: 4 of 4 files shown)


cebe:
How can the same component run in parallel with itself? Mutex is used to sync different processes. Could you explain how you get to use the mutex parallelly in one process?

rob006:
You can get it with recursion or cyclic dependency (like processing `A` requires processing `B` which requires processing `A`) - not so hard with complicated and event-based systems. But the main problem here is that number of successful `acquire()`  will not match number of successful  `release()`:

```php
if (Yii::$app->mutex->acquire('test')) {
     // do something
     if (Yii::$app->mutex->acquire('test')) {
         // do something
         Yii::$app->mutex->release('test');
    }
    // I don't have  active lock here!

    Yii::$app->mutex->release('test'); // false
}
```

samdark:
@rob006 the mutex was originally made to deal with multi-threaded software where multiple threads could access a shared resource. There was no intention to block access within the same thread. Since there are no threads in PHP, we're talking about processes i.e. current behavior is correct.

Your case looks like reentrant mutex...

rob006:
@samdark So this is intended to work in this way and I can acquire the same lock multiple times in one process, and framework will not complain about it? Is there any reason why it works in this way, so I need additional layer to protect me from acquiring the same lock multiple times (whatever it was the result of bug in code or misinterpretation how mutex works)?

samdark:
> So this is intended to work in this way and I can acquire the same lock multiple times in one process, and framework will not complain about it?

Yes. The description of the class explicitly states that "The Mutex component allows mutual execution of concurrent processes in order to prevent "race conditions"".

> Is there any reason why it works in this way

To follow [mutex definition](https://en.wikipedia.org/wiki/Lock_(computer_science)) precisely. If it won't work that way it would be not so good idea to call it either Lock or Mutex.

samdark:
Closing since current behavior is correct.

rob006:
> To follow [mutex definition](https://en.wikipedia.org/wiki/Lock_(computer_science)) precisely.

Can you quote precisely what part of this definition says that it should work that way (locks are *always* bound to process/thread)?

samdark:
> synchronization mechanism for enforcing limits on access to a resource in an environment where there are many threads of execution

There is a single thread of execution in a single PHP process.

cebe:
>  // I don't have active lock here!

I understand the problem this causes in such situation, we should probably throw an exception if you try to aquire the lock twice.

rob006:
> I understand the problem this causes in such situation, we should probably throw an exception if you try to aquire the lock twice.

I think it would be an overkill and this does not deserve anything more than a warning. Personally, I would not emit any warning and just return `false`. A warning would make sense if we return `true` in this case (since it is not "success" but more like "success, however...").

___

I made some research how Symfony and Laravel handles such case and:

## Symfony

In Symfony lock is an object, so to create two lock you're creating two `Lock` objects and tries to call `acquire()` on both. I tested only `SemaphoreStore` but it does not allow to acquire two separate locks with the same key (but you can call `acquire()` multiple times on one object, so you can ensure that already acquired lock is still acquired).

```php
use Symfony\Component\Lock\Factory;
use Symfony\Component\Lock\Store\SemaphoreStore;

$store = new SemaphoreStore();
$factory = new Factory($store);
$lock1 = $factory->createLock('my-lock');
$lock2 = $factory->createLock('my-lock');

var_dump($lock1->acquire(), $lock1->acquire());
var_dump($lock2->acquire(), $lock2->acquire());
```

Result:

```
robert@robert-laptop:~/Poligon/symfony-mutex/public$ php index.php 
/home/robert/Poligon/symfony-mutex/public/index.php:18:
bool(true)
/home/robert/Poligon/symfony-mutex/public/index.php:18:
bool(true)
/home/robert/Poligon/symfony-mutex/public/index.php:19:
bool(false)
/home/robert/Poligon/symfony-mutex/public/index.php:19:
bool(false)
```

## Laravel 

I found only this: https://github.com/laravel/framework/blob/5.5/src/Illuminate/Console/Scheduling/CacheMutex.php - it uses different naming and instead of `acquire()` there is `create()` - you can create lock only once even if you're doing it in one process.

IMO both implementations works differently than current Yii implementation.

___

And besides all of that I think that current implementation is against Yii values:

1. **practice over theory** - implementation blindly follows definition from Wikipedia (which is not clear in this matter and may be interpreted differently) instead of relying on real use-case.
2. **explicitness** - current implementation return success in situation which is at least controversial (and should not happen with current assumptions).

cebe:
trying to aquire the same lock twice in one process is wrong, it can result in a dead lock when the process is blocking itself. Lock can not be released in other part of the program if it is waiting for release in another part. There is no situation where it makes sense to use the same lock twice in the same thread, so failing hard makes sense to me.

rob006:
> There is no situation where it makes sense to use the same lock twice in the same thread

It makes sense if you want to check if some business process is not in progress - we already solving similar problem on initializing components to avoid infinite loops when component A tries to initialize component B which tries to initialize component A. Throwing exception will make mutex component less flexible (and inconsistent with other popular implementations) just because you decided that it does not make sense. 

> it can result in a dead lock when the process is blocking itself

Current mutex implementation does not support infinite blocking - you will get deadlock only when you implement infinite loop yourself. And framework is not able to prevent deadlock anyway - even if we leave old implementation or throw exception, you still can get deadlock if two separate processes tries to acquire the same two lock in different order.

samdark:
> check if some business process is not in progress

A variable should be enough for it. No need to lock anything with files or database.

samdark:
OK. That makes sense. Now, thinking about applying changes, are you sure the change is backwards compatible?

rob006:
> are you sure the change is backwards compatible?

No, this PR changes existing behavior and every such change may be considered as BC break (even if you're changing incorrect behavior to correct - there could always be people who rely on some bugs). 

Right now I can image only one case when someone may intentionally use current implementation in this way: using `acquire()` to ensure that lock is still active (like in long running process, which could disconnect with database). But it should not break anything - you should expect that `acquire()` may return `false` in this case. The only problem is if someone really put this is infinite loop without any break condition (`while(!$mutex->acquire()) {}`). I'm not sure how popular is such approach, maybe mentioning it in `UPGRADE.md` will be enough?

slow-loris:
This behavior broke my use case:

Method "moveStock":
- Acquires lock
- Does some stock calculations
- Move stock
- Release lock

Method "moveBinLocation":
- Acquires lock
- Does some stock calculations
- Calls method moveStock for every stock on bin location
- Release lock

This worked fine until this update.

I just want to make sure, that no other process can modify stock util all stock operations are done.
How can I archive this?

rob006:
@slow-loris Use different lock names for `moveBinLocation` and `moveStock`.

slow-loris:
> @slow-loris Use different lock names for `moveBinLocation` and `moveStock`.

That's not possible.
If someone moves the whole bin location, no single stock operation should be possible.
If someone wants to perform an elementary stock operation he has to wait until the batch processing (`moveBinLocation`) has finished.

So my use case is: Lock everything that has something to do with a stock operation for the running process.

rob006:
You can extract `moveStockInternal()` which will move stock without acquiring lock and `moveStock()` which will acquire lock and call `moveStockInternal()`. Then `moveBinLocation()` will call `moveStockInternal()` - you can avoid nested locks in this way.

Note that nested locks never worked properly - `moveStock()` may release lock acquired in `moveBinLocation()` and `moveBinLocation()` will continue execution without active lock (see https://github.com/yiisoft/yii2/pull/16838#issuecomment-433699914).

PR #16839, opened 2018-10-27
Increase frequency of lock tries when $timeout is used in FileMutex

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

This changes frequency of lock tries for `FileMutex::acquireLock()` when `$timeout` is provided (and make it configurable). Previously it was once try per second, now it is 20 tries per second. 

See https://github.com/yiisoft/yii2-queue/pull/244#issuecomment-406853925


samdark:
Merged. Thanks!

PR #16844, opened 2018-10-29
Update yii.gridView.js

pvsaintpe:
If there are many filters in the GridView, then when changing any filter, all filters are sent by the GET method, we get the error: Url is too long. 
Please use the changes I suggested to make the URLs shorter. 
Thank.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Tests pass?   | yes

### DIFF
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -124,7 +124,9 @@
                 if (!(this.name in data)) {
                     data[this.name] = [];
                 }
-                data[this.name].push(this.value);
+                if (this.value) {
+                    data[this.name].push(this.value);
+                }
             });
 
             var namesInFilter = Object.keys(data);
@@ -135,11 +137,15 @@
                         value = [value];
                     }
                     if (!(name in data)) {
-                        data[name] = value;
+                        if (value) {
+                            data[name] = value;
+                        }
                     } else {
                         $.each(value, function (i, val) {
                             if ($.inArray(val, data[name])) {
-                                data[name].push(val);
+                                if (val) {
+                                    data[name].push(val);
+                                }
                             }
                         });
                     }


rob006:
This will significantly change current behavior and definitely break some apps. Right now if default value of field is for example `'default'` and someone changes this to an empty string (so he wants to disable default filtering), this empty string is sent in GET and `load()` will replace default value with an empty string. After this change empty inputs will be ignored, so there is no way to clear default filtering.

samdark:
@rob006 correct. @pvsaintpe sorry, but since the change is backwards compatibility breaking, we can not accept it.

pvsaintpe:
> This will significantly change current behavior and definitely break some apps. Right now if default value of field is for example `'default'` and someone changes this to an empty string (so he wants to disable default filtering), this empty string is sent in GET and `load()` will replace default value with an empty string. After this change empty inputs will be ignored, so there is no way to clear default filtering.

Yes, but how to solve a problem with grids where there are a lot of columns, and long names of Search-models, if there are more than 15 columns in the grid, and the model name is long, we get the error Url is too long and the application is broken.

rob006:
Did you tried to hide model name from URL by overriding `formName()` in search model to return empty string?

pvsaintpe:
> Did you tried to hide model name from URL by overriding `formName()` in search model to return empty string?

Yes, I'm tried.

But my client, for whom the application is being developed, wants grids in which there are more than 20-30 columns with horizontal scrolling and this does not solve the problem if I exclude the form name from the URL by redefining formName (), all the same we get an error with "Url is too long. "

It is in my application that only those filters that are changed should be sent to the URL. And if it were possible to make an option in yii.gridView.js, which would be thrown in js from yii \ grid \ GridView then it would be possible to use your components more flexibly.

PR #16846, opened 2018-10-30
Update GridView.php

pvsaintpe:
If you do not want to make changes to yii.gridView.js, let at least be able to override GridViewAsset and connect your version of yii.gridView.js there. 

Because in our application there are no default filters that need to be rewritten when filtering, but the problem with long URLs after parameter substitution when using filtering is very relevant for us. 

Thank you in advance!

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Tests pass?   | yes
| Fixed issues  | https://github.com/yiisoft/yii2/pull/16844

### DIFF
--- a/framework/grid/GridView.php
+++ b/framework/grid/GridView.php
@@ -265,6 +265,13 @@ class GridView extends BaseListView
      */
     public $layout = "{summary}\n{items}\n{pager}";
 
+    /*
+     * To be able to override the base Asset for the GridView to your own.
+     * @see https://github.com/yiisoft/yii2/pull/16844
+     * @var string Class for GridViewAsset without leading slash
+     * Defaults to 'yii\grid\GridViewAsset'.
+     */
+    public $gridViewAssetClass;
 
     /**
      * Initializes the grid view.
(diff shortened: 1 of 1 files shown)


framework/grid/GridView.php
```diff
@@ -294,10 +301,16 @@ public function init()
...
         $view = $this->getView();
-        GridViewAsset::register($view);
         $id = $this->options['id'];
         $options = Json::htmlEncode(array_merge($this->getClientOptions(), ['filterOnFocusOut' => $this->filterOnFocusOut]));
-        $view->registerJs("jQuery('#$id').yiiGridView($options);");
+        $baseGridViewAsset = 'yii\grid\GridViewAsset';
+        if (!($gridViewAsset = $this->gridViewAssetClass)) {
+            $gridViewAsset = $baseGridViewAsset;
+        }
+        if (is_subclass_of($gridViewAsset, $baseGridViewAsset) || $gridViewAsset == $baseGridViewAsset) {
```
samdark: instanceof could be used here instead.

samdark:
Another way to do it is to extract registering assets into `protected registerAssets()`.

framework/grid/GridView.php
```diff
@@ -265,6 +265,13 @@ class GridView extends BaseListView
      */
     public $layout = "{summary}\n{items}\n{pager}";
 
+    /**
+     * To be able to override the base Asset for the GridView to your own.
+     * @see https://github.com/yiisoft/yii2/pull/16844
+     * @var string Class for GridViewAsset without leading slash
+     * Defaults to 'yii\grid\GridViewAsset'.
+     */
+    public $gridViewAssetClass;
```
cebe: why use null as the default and not `'yii\grid\GridViewAsset'`?

cebe:
you could instead of configuring this set the asset to false on assetmanager, or override the scripts loaded as described here:

https://www.yiiframework.com/doc/guide/2.0/en/structure-assets#customizing-asset-bundles

there should be no need to configure gridview asset class in this place.

samdark:
@pvsaintpe are you interested to finish it?

samdark:
Closing since it seems there's no plan finishing it.

PR #16847, opened 2018-10-30
Memory overflows in cli mode

hughcube:
Memory overflows in cli mode

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes

### DIFF
--- a/framework/caching/ArrayCache.php
+++ b/framework/caching/ArrayCache.php
@@ -93,6 +93,7 @@ class ArrayCache extends Cache
     protected function deleteValue($key)
     {
         unset($this->_cache[$key]);
+        $this->gc();
 
         return true;
     }
--- a/framework/caching/ArrayCache.php
+++ b/framework/caching/ArrayCache.php
@@ -93,7 +93,6 @@ class ArrayCache extends Cache
     protected function deleteValue($key)
     {
         unset($this->_cache[$key]);
-        $this->gc();
 
         return true;
     }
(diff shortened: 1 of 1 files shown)


framework/caching/ArrayCache.php
```diff
@@ -84,6 +104,26 @@ protected function deleteValue($key)
...
+     * Removes the expired data values.
+     * @param bool $force whether to enforce the garbage collection regardless of [[gcProbability]].
+     * Defaults to false, meaning the actual deletion happens with the probability as specified by [[gcProbability]].
+     */
+    public function gc($force = false)
+    {
+        if ($force || mt_rand(0, 1000000) < $this->gcProbability){
+            foreach($this->_cache as $key => $item){
+                if (!isset($item[0], $item[1]) || (0 !== $item[0] && $item[1] < microtime(true))){
+                    $this->deleteValue($key);
```
rob006: You have potential infinite recursion here: `gc()` calls `deleteValue()` which calls `gc()`.

rob006 (dismissed):

framework/caching/ArrayCache.php
hughcube: Has been fixed

rob006:
Honestly, I'm not sure about the whole concept. AFAIK `ArrayCache` was designed as super fast dummy cache, and adding GC (especially enabled by default) in 99.99% cases will give you nothing except additional overhead. Especially that it is triggered more often than in other cache backends.

samdark:
@hughcube are you running daemons?

hughcube:
> are you running daemons?

Yes, otherwise there won't be this demand

cebe:
Could be disabled by default. Also GC value depends a lot on the number of calls to the cache, it could have a time limit also, which would prevent a GC to run more than x times per time frame.

samdark:
That seems to be out of scope of the common framework usage. Thanks for contribution, @hughcube but won't be merged.

PR #16848, opened 2018-10-30
Update yii.gridView.js

pvsaintpe:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | https://github.com/yiisoft/yii2/pull/16844

### DIFF
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -25,7 +25,7 @@
         filterUrl: undefined,
         filterSelector: undefined,
         filterOnFocusOut: true,
-        filterAllowedSubmitEmpty: true
+        filterAllowedEmpty: true
     };
 
     var gridData = {};
@@ -120,12 +120,13 @@
         applyFilter: function () {
             var $grid = $(this);
             var settings = gridData[$grid.attr('id')].settings;
+            var allowedEmpty = $(settings.filterAllowedEmpty) || true;
             var data = {};
             $.each($(settings.filterSelector).serializeArray(), function () {
                 if (!(this.name in data)) {
                     data[this.name] = [];
                 }
-                if (settings.filterAllowedSubmitEmpty || (!settings.filterAllowedSubmitEmpty && this.value)) {
+                if (allowedEmpty || (!allowedEmpty && this.value)) {
                     data[this.name].push(this.value);
                 }
             });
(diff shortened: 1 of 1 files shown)


pvsaintpe:
Please, comment this Node.js test: https://travis-ci.org/yiisoft/yii2/jobs/448366748

pvsaintpe:
Your test showed 5 errors, but this cannot be considered a mistake.

I added a new parameter to the gridView, with which you can allow or prohibit sending empty filters, thus preserving the performance of old applications (where you need to be able to rewrite the default filters) and simultaneously solve the problem with long URLs (sending only empty filters).

For example,
filterAllowedEmpty = true
? PostSearch [name] = & PostSearch [subject] = subject & PostSearch [author] =

filterAllowedEmpty = false
? PostSearch [subject] = subject

I propose to slightly change your Node.js test in parts of the last 5 cases, which, when entering my parameter, drop with an error.

rob006:
Wouldn't be more practical to use POST in such case?

pvsaintpe:
> Not if you need to link to these filtered pages.

We considered options with POST, but at one moment it became inconvenient, since There are no parameters in the address bar, and the customer’s desire was to "change" some values in the address bar and to have links to filtered grids.

Please understand me, I am not trying to push through a non-working code, on the contrary, I want to use your original framework and receive updates.

I can solve the problem myself, do a fork, but this is the right way.

I would be grateful if you could use my pool request.

rob006:
My point is that this is only a partial solution - you still may reach limit if filters will be used and this is still extremely ineffective for multiselects (or any kind of arrays).

samdark:
@rob006 is correct. You can set values to most of your fields and the problem is back. The only value I see in the pull request is that it may produce a bit better-looking URLs.

s1lver:
I think, nevertheless, it is necessary to give the opportunity to adjust this behavior in the parameters of the widget.

Now I use a partial URL to send a link, so I did not encounter such a problem. In this case, the columns in the table can reach up to 20-30.

samdark:
Yes. I think it makes sense.

pvsaintpe:
Thank you for your understanding, I have never written module tests for Yii2, if you gave me examples, I would be happy to write these tests for you, thanks in advance.

s1lver:
Tests can be found here `yii2\tests`
You need to at least do a JS test. You can run as in a browser, but you will need to be further configured, and with the help of Node.js in your IDE or Node.js console. I prefer to run and debug from PHPShorm.
 
File for tests

- yii2\tests\js\tests\yii.gridView.test.js
- yii2\tests\framework\grid\GridViewTest.php

 You can also read the README in the yii2\tests\README.md folder.

s1lver:
@pvsaintpe you're not working on it anymore? If not, I could finish by adding tests and entering optionality.

pvsaintpe:
@s1lver, 

Travis has passed the test (backward compatibility is not broken), and after talking to @samdark, I’m waiting for action on your part, thanks!

arzzen (approved):

terabytesoftw:
Thank you for the contribution and for the discussion here.

I’m closing this PR because it has been inactive for a long time, still lacks the required unit tests, and the contributor source branch/fork is no longer available (`head.repo` is missing), so this PR cannot be updated or merged as-is.

PR #16863, opened 2018-11-04
Remove template method in test case

samnela:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

If it is okay with you, I can remove the other templates in the tests


samdark:
I don't think it makes sense to remove calling a parent implementation.

Issue #14950, opened 2017-10-10
asDecimal() return wrong results


hirenbhut93:
Different results return from `intl` and  `number_format()`

### What steps will reproduce the problem?

Yii::$app->formatter->asDecimal('18446744073709551615');

### What is the expected result?

When i have used php function `number_format()` that time return below output. so i think is a expected or not ?.  not confirm

Results : 18,446,744,073,709,551,616

### What do you get instead?
// convert using `intl`
Results :  18,446,744,073,709,600,000

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.12
| PHP version      | 7.1
| Operating system | Ubuntu 16.04

SilverFire:
Verified

zhukovra:
@SilverFire @samdark would not be `asDecimal` completely rewritten without using native functions like `number_format()` and `intl` extension?

**Test**

    $value = '123456789012345678901234567890';
    $this->assertSame('123,456,789,012,345,678,901,234,567,890', $this->formatter->asDecimal($value, 0));

**Output**

    Failed asserting that two strings are identical.
    Expected :123,456,789,012,345,678,901,234,567,890
    Actual   :123,456,789,012,345,677,877,719,597,056

PHP internal casting big integers to float leads to badly mistaken result.

samdark:
Probably a good idea.

PR #16879, opened 2018-11-11
Formatter fallback methods for very big numbers

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14950

This one is tricky and I'm not sure if you want it that way.

I couldn't find any reliable way to check when normalized value are mispresented because of internal casting, floating point numbers presentation or precision configuration of PHP but my tests show that comparing normalized value to original one as strings gives a pretty good hint that final output may be broken.

Unfortunately internal casting of very big numbers can't be fixed even with fallback methods since the original value is already corrupted when passed to the method. That is why I've added hint in phpdoc to pass numbers as strings in case of very big ones.

I've added more tests for other formatters to see if these are subjects to the same issue and yes, `integer`, `percent` and `currency` needed to be fixed as well.

Strictly `intl` formatters are basically fine except `ordinal` where I've added phpdoc warning for v. big numbers (it should not be used with them anyway).


samdark (changes requested):

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+     * to the defined decimal digits.
+     *
+     * @param string|int|float $value the value to be formatted.
+     * @param int $decimals the number of digits after the decimal point. The default value is `2`.
+     * @return string the formatted result.
+     * @see decimalSeparator
+     * @see thousandSeparator
+     * @since 2.0.16
+     */
+    protected function asDecimalStringFallback($value, $decimals = null)
```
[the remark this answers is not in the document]
bizley: These extensions are not available by default AFAIK hence the strings operations.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+    {
+        if (empty($value)) {
+            $value = 0;
+        }
+
+        $value = (string) $value;
+
+        if (strpos($value, '.') !== false) {
```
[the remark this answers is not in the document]
bizley: It's true but numeric string with comma decimal point will throw exception in `normalizeNumericValue()` before getting here.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+    }
+
+    /**
+     * Fallback for formatting value as an integer number by removing any decimal digits without rounding.
+     *
+     * @param string|int|float $value the value to be formatted.
+     * @return string the formatted result.
+     * @since 2.0.16
+     */
+    protected function asIntegerStringFallback($value)
```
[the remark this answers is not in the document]
bizley: Like in `StringHelper::truncate()`? Sort of but not exact same.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+            $integerPart = substr($value, 0, strrpos($value, '.'));
+            $fractionalPart = substr($value, strrpos($value, '.') + 1);
+        } else {
+            $integerPart = $value;
+            $fractionalPart = null;
+        }
+
+        $decimalOutput = '';
+
+        if ($decimals === null) {
```
samdark: Isn't it better to do `$decimals = 2` in the method signature?

framework/i18n/Formatter.php
samdark: `strpos($value, '.')` should not be executed three times. Assign a variable and reuse it.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+            }
+
+            if ($fractionalPart === null) {
+                $fractionalPart = str_repeat('0', $decimals);
+            } elseif (strlen($fractionalPart) > $decimals) {
+                $cursor = $decimals;
+
+                if ((int) substr($fractionalPart, $cursor, 1) >= 5) {
+                    while (--$cursor >= 0) {
+                        $round = false;
```
samdark: Isn't it better to call it `$carry = 0` instead of `$round = false`? Because you are doing `+1` addition here.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+                        }
+
+                        $fractionalPart = substr($fractionalPart, 0, $cursor) . $newDigit . substr($fractionalPart, $cursor + 1);
+
+                        if (!$round) {
+                            break;
+                        }
+                    }
+                    if ($round) {
+                        $roundIntegerPart = true;
```
samdark: Additional variable and this condition could be avoided by defining `$round = false` or `$carry = 0` before `if ($fractionalPart === null) {`.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+                $fractionalPart = str_pad($fractionalPart, $decimals, '0');
+            }
+
+            $decimalOutput .= $decimalSeparator . $fractionalPart;
+        }
+
+        if ($roundIntegerPart || ($decimals === 0 && $fractionalPart !== null && (int) substr($fractionalPart, 0, 1) >= 5)) {
+            $cursor = 0;
+            while (++$cursor <= strlen($integerPart)) {
+                $round = false;
```
samdark: That looks like addition is happening. Worth using `$carry`.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+            } elseif (strlen($fractionalPart) < $decimals) {
+                $fractionalPart = str_pad($fractionalPart, $decimals, '0');
+            }
+
+            $decimalOutput .= $decimalSeparator . $fractionalPart;
+        }
+
+        if ($roundIntegerPart || ($decimals === 0 && $fractionalPart !== null && (int) substr($fractionalPart, 0, 1) >= 5)) {
+            $cursor = 0;
+            while (++$cursor <= strlen($integerPart)) {
```
samdark: `strlen($integerPart)` should be moved out of condition.

samdark: Could be moved right to the beginning of the method.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+     */
+    protected function asIntegerStringFallback($value)
+    {
+        if (empty($value)) {
+            $value = 0;
+        }
+
+        $value = (string) $value;
+
+        if (strpos($value, '.') !== false) {
```
samdark: `strpos($value, '.')` could be extracted into variable not to call it multiple times.

framework/i18n/Formatter.php
```diff
@@ -1741,4 +1778,220 @@ protected function createNumberFormatter($style, $decimals = null, $options = []
...
+            $value = 0;
+        }
+
+        if ($decimals === null) {
+            $decimals = 0;
+        }
+
+        $value = (string) $value;
+
+        if (strpos($value, '.') !== false) {
```
samdark: `strpos($value, '.')` should be extracted in variable not to call it multiple times.

samdark (changes requested):

framework/i18n/Formatter.php
bizley: Not sure. `null` can be still passed so I left condition there but changed the signature as well.

framework/i18n/Formatter.php
bizley: Right. Variabled.

framework/i18n/Formatter.php
bizley: You are right. Rewritten with `$carry`.

framework/i18n/Formatter.php
bizley: Done.

framework/i18n/Formatter.php
bizley: Also done.

framework/i18n/Formatter.php
bizley: Moved before `while` to avoid calling it if not necessary in method. Last `if` needs to check it again anyway because of `$carry` adding 1.

framework/i18n/Formatter.php
bizley: Variabled.

samdark:
Merged. Thank you!

PR #16882, opened 2018-11-12
 Added test with scientific timestamp

s1lver:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  |no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15528


rob006:
This test does not make any sense and it is unrelated to #15528. You're just testing PHP parser here, since there is no real difference between `1.51643E+9` and `1516430000`, both are parsed to the same value.

s1lver:
@rob006 So I also want to show that there is no difference. I do not know how to express a thought.
In the comments to the topic, I tried to explain.

In short, any change to this function does not make sense.

rob006:
@s1lver I think you don't understand the problem. In #15528 for some weird reason `1516425820.449680` is represented as `1.51643E+9` when you typecast float to string. Even if you can't reproduce it, it only prove that *you* can't reproduce it - float can be represented in scientific notation and `Target::getTime()` should be prepared for such possibility.

s1lver:
@rob006 I understood you

PR #16903, opened 2018-11-22
fixed NumberValidator so it does not validate boolean as true(as number)

annechko:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #16903: Fixed 'yii\validators\NumberValidator' method 'isNotNumber' returns false for true/false value (annechko)
 - Bug #16648: Html::strtolower() was corrupting UTF-8 strings (Kolyunya)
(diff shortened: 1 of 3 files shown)


annechko:
For example, as standalone valdator - it will count boolean as valid number
```php
    $val = new NumberValidator(['integerOnly' => true]);
    $val->validate(false); // true
    $val->validate(true); // true
```
also in models
```php
    /*
    [
        ['field', 'integer'],
    ]
    */
    $model->load(['field' => true], '');
    $model->validate(); // true

    $model->load(['field' => false], '');
    $model->validate(); // true  
```
(but false/true are not numbers so I guess it should not be valid)

cebe (approved):
Looks good to me. Thanks!

cebe:
Please add an entry in the CHANGELOG

samdark:
Merged. Thank you very much!

annechko:
Thank you!
It's my first ever contribution to Open Source! 
(Sorry, can't help but post this)
image

samdark:
Welcome to our world :)

PR #16905, opened 2018-11-23
ActiveRecord:findByCondition alias fix

rikosage:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | I couldn't run it
| Fixed issues  | no

`ActiveRecord::findOne($id)` method doesn't includes table alias. If alias is defined in `find` or `init` methods, `ActiveRecord::findByCondition` returns error.

### DIFF
--- a/framework/db/ActiveRecord.php
+++ b/framework/db/ActiveRecord.php
@@ -174,26 +174,7 @@ class ActiveRecord extends BaseActiveRecord
         $query = static::find();
 
         if (!ArrayHelper::isAssociative($condition)) {
-            // query by primary key
-            $primaryKey = static::primaryKey();
-
-            if (!isset($primaryKey[0])) {
-                throw new InvalidConfigException('"' . get_called_class() . '" must have a primary key.');
-            }
-
-            $pk = $primaryKey[0];
-            if (!empty($query->join) || !empty($query->joinWith)) {
-                // AR-model table name or table alias if defined
-                $tableOrAlias = static::tableName();
-                foreach (array_keys($query->from ?: []) as $alias) {
-                    if (is_string($alias)) {
-                        $tableOrAlias = $alias;
-                        break;
-                    }
-                }
-
-                $pk = $tableOrAlias . '.' . $pk;
-            }
+            $pk = self::primaryKeyFieldWithAlias($query);
             // if condition is scalar, search for a single primary key, if it is array, search for multiple primary key values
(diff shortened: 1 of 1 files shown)


framework/db/ActiveRecord.php
```diff
@@ -179,7 +179,7 @@ protected static function findByCondition($condition)
             if (isset($primaryKey[0])) {
                 $pk = $primaryKey[0];
                 if (!empty($query->join) || !empty($query->joinWith)) {
-                    $pk = static::tableName() . '.' . $pk;
+                    $pk = ($query->from ? key($query->from) : static::tableName()) . '.' . $pk;
```
cebe: this is not correct, `$query->from` could be `['sometable']` where the key would be 0.
This should use similar approach as `ActiveQuery::getTableNameAndAlias()` but that method is protected right now.

rikosage: Should I create new method like `getAlias()`? Or I can just change `protected` to `public`, but it's doubtful. What's better solution?

cebe: ActiveQuery should not have public methods other than those building the query, so a private function for getting the alias should be used. It's okay to duplicate some code from ActiveQuery. I don't see a better way to solve it right now.

rikosage:
I've reduced cyclomatic comlexity. 
Should I use Reflection for testing new private method?

samdark:
Isn't it possible to test it w/o reflection?

framework/db/ActiveRecord.php
```diff
@@ -174,25 +174,47 @@ protected static function findByCondition($condition)
...
+
+        if (!isset($primaryKey[0])) {
+            throw new InvalidConfigException('"' . get_called_class() . '" must have a primary key.');
+        }
+
+        if (empty($query->join) && empty($query->joinWith)) {
+            return $primaryKey[0];
+        }
+
+        $tableOrAlias = static::tableName();
```
GHopperMSK: In this conditions `$query->from` returns or null or single-element array. That is why we can just use
```suggestion
        $tableOrAlias = (!$query->from || isset($query->from[0])) ? static::tableName() : key($query->from);
```
without foreach loop, `array_keys` and even `is_string` funtions.

samdark:
Closed in favor of https://github.com/yiisoft/yii2/pull/17127

PR #16918, opened 2018-11-27
made private vars protected in console table

schmunk42:
| Q             | A
| ------------- | ---
| Is bugfix?    | -
| New feature?  | -
| Breaks BC?    | no
| Tests pass?   | see Travis

Otherwise it is practically impossible to extend this widget.

I wanted to create a slightly different table rendering, example:

```php
class SimpleTable extends Table
{
    /**
     * @var array table chars
     */
    protected  $_chars = [
        self::CHAR_TOP => '═',
        self::CHAR_TOP_MID => '╤',
        self::CHAR_TOP_LEFT => '╔',
        self::CHAR_TOP_RIGHT => '╗',
        self::CHAR_BOTTOM => '═',
        self::CHAR_BOTTOM_MID => '╧',
        self::CHAR_BOTTOM_LEFT => '╚',
        self::CHAR_BOTTOM_RIGHT => '╝',

        self::CHAR_LEFT => '>',

        self::CHAR_LEFT_MID => '╟',

        self::CHAR_MID => null,

        self::CHAR_MID_MID => '┼',
        self::CHAR_RIGHT => '',
        self::CHAR_RIGHT_MID => '╢',
        self::CHAR_MIDDLE => '',
    ];

    /**
     * @return string the rendered simple table
     */
    public function run()
    {
        $this->calculateRowsSize();
        $buffer = '';
        // Header
        $buffer .= $this->renderRow($this->_headers,
            $this->_chars[self::CHAR_LEFT],
            $this->_chars[self::CHAR_MIDDLE],
            $this->_chars[self::CHAR_RIGHT]
        );

        // Content
        foreach ($this->_rows as $row) {
            $buffer .= $this->renderRow($row,
                $this->_chars[self::CHAR_LEFT],
                $this->_chars[self::CHAR_MIDDLE],
                $this->_chars[self::CHAR_RIGHT]);
        }

        return $buffer;
    }
}
```


framework/console/widgets/Table.php
```diff
@@ -69,15 +69,15 @@ class Table extends Widget
     /**
      * @var array table headers
      */
-    private $_headers = [];
+    protected $_headers = [];
```
samdark: protected variables should not start with `_`

samdark (changes requested):

framework/console/widgets/Table.php
schmunk42: So just removing the `_` everywhere would be fine?

rob006: > protected variables should not start with `_`

`_` prefix is for avoiding conflicts between regular properties and virtual properties from getters and setters. Visibility should not matter here (this would be against PSR-2).

cebe:
Making all at once protected does not make sense to me. E.g. `_screenWidth` has a setter and getter which you can override. I don't see why it should be protected.

Can you explain how you want to extend the class?

schmunk42:
@cebe Have a look at the code in my initial posting.

I actually just wanted to override the run method to discard the separation lines in the table to make it more condensed, but I need to access at least `$this->_rows` which is private.

But you're right about getters and setters.

samdark:
@schmunk42 do you still need these changes?

schmunk42:
Actually yes.

samdark:
Adjusted/merged.

schmunk42:
Thank you!

PR #16924, opened 2018-11-29
Wrong test removed

bizley:
I'm sorry but spellout formatter, just like ordinal one does not work well with very big numbers.

I really don't know why I've put this test here and decided it's fine and dandy.

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -1296,6 +1296,8 @@ class Formatter extends Component
      * @return string the formatted result.
      * @throws InvalidArgumentException if the input value is not numeric or the formatting failed.
      * @throws InvalidConfigException when the [PHP intl extension](http://php.net/manual/en/book.intl.php) is not available.
+     *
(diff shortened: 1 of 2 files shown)


samdark:
Merged. Thanks!

framework/i18n/Formatter.php
```diff
@@ -1296,6 +1296,8 @@ public function asCurrency($value, $currency = null, $options = [], $textOptions
      * @return string the formatted result.
      * @throws InvalidArgumentException if the input value is not numeric or the formatting failed.
      * @throws InvalidConfigException when the [PHP intl extension](http://php.net/manual/en/book.intl.php) is not available.
+     *
+     * This formatter does not work well with very big numbers.
      */
```
cebe: this text will now be part of the `@throws` tag ;)

cebe: fixed in https://github.com/yiisoft/yii2/commit/fe07177dddeaf4b8d9c1dcbc103ba3d6f572cb10

Issue #16910, opened 2018-11-23
Parameter sort in extraction of messages config not working?


AndersonDan:
### What steps will reproduce the problem?

Create the config file to extract translation messages:

```php
return [
    'color' => null,
    'interactive' => true,
    'help' => null,
    'messagePath' => '@backend/messages',
    'languages' => [
        'en',
        'pt-BR',
    ],
    'translator' => 'Yii::t',
    'sort' => false, // <---------------- here
    'overwrite' => true,
    'removeUnused' => false,
    'markUnused' => false,
    'except' => [
        '.svn',
        '.git',
        '.gitignore',
        '.gitkeep',
        '.hgignore',
        '.hgkeep',
        '/messages',
        '/BaseYii.php',
    ],
    'only' => [
        '*.php',
    ],
    'format' => 'php',
    'catalog' => 'messages',
    'ignoreCategories' => [
        'yii*',
    ],
    'phpFileHeader' => '',
    'phpDocBlock' => null,
];
```

### What is the expected result?
New messages extracted at the end of the generated translation files.

### What do you get instead?
Messages are added to the existing file and are sorted in alphabetical order.

### Additional info

| Q                | A
| --------------------- | ---
| Yii version               | 2.0.15.1?
| PHP version            |  7.2
| Operating system   | Debian GNU/Linux 9 (stretch)

Groonya:
Not really. Option `$sort` in MessageController just controls in what order messages should be inserted into translation target (file or db). It's very important for files because you can open any file and see ordered messages. But it is useless for a database. When you do `SELECT * FROM table ` in most cases dbms doesn't guarantee the order - you have to specify the order. 
So main question here is for whom we need to order messages in the database?
For user? No, he must specify the order if he wants to see ordered messages.
For framework? No, it's not important for it how messages are ordered.

PR #16927, opened 2018-12-02
Fixed sorting on extracting messages to PO and PHP files #16910.

aivchen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | partially
| Fixed issues  | #16910.

I've written test and fixed messages extracting for PHP and PO files only. But the test doesn't pass for Db (Do we really need to sort extracted messages in db?)


samdark:
There's no big need for sorting DB messages but since general config supports it, it's fine to implement it. Should be ORDER BY, right?

Groonya:
Not really. Option $sort in MessageController just controls in what order messages should be inserted into translation target (file or db). It's very important for files because you can open any file and see ordered messages. But it is useless for a database. When you do SELECT * FROM table in most cases dbms doesn't guarantee the order - you have to specify the order.
So main question here is for whom we need to order messages in the database?
For user? No, he must specify the order if he wants to see ordered messages.
For framework? No, it's not important for it how messages are ordered.

samdark:
Well, yes. There's no need to order for database but that breaks current inheritance design...

Groonya:
What do you mean? 
If you are talking about tests, then we can skip the test for database?
Just want to be sure I'm thinking in a right way.

samdark:
Yes, I think we can do it. Would you please mark test as skipped?

samdark:
Merged. Thank you!

Issue #16580, opened 2018-07-30
MessageController (message/extract) does not delete files with removeUnused=true


Vovan-VE:
### What steps will reproduce the problem?

1. PhpMessageSource
2. Files `messages/<language>/category/subcategory.php`:
```php
<?php
return [
    'Message' => 'Translation',
];
```
3. Source code to extract messages from does not contain anymore any messages the category `Yii::t('category/subcategory', ...)`;
4. `./yii message/extract message/config.php` so `removeUnused` is `true`

### What is the expected result?

Since there are no more messages in the category `category/subcategory`, I extect a deletion of files `messages/<language>/category/subcategory.php` (with directories becoming empty) or at least see any notice about such files or categories.

### What do you get instead?

Neither the files was deleted, nor notices are displayed.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | *
| Operating system | *

samdark:
Indeed, deleting files isn't implemented.

PR #16930, opened 2018-12-02
#16580: Check for unused php message files in MessageController if $r…

aivchen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes/no (I'm not sure)
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16580 

I think files deleting is dangerous because we don't know what categories already exists (user can put his own files in messages dir), so I just implemented a notice.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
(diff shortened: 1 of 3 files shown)


samdark:
The option is called `removeUnused` so it should remove, not check...

Groonya:
So we should remove all other php files in messages directory?

PR #16935, opened 2018-12-04
#16580: Delete unused php message files in MessageController if $remo…

aivchen:
…veUnused option is on

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16580

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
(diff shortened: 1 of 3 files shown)


samdark:
Merged. Thank you!

samdark:
Merged. Thank you!

albertborsos:
This PR breaks the message file generation. It removes existing message files even if they are in use. After it removed the message files, then I set `removeUnused` to `false` and then it recovers the deleted message files (of course with empty values). So they are exist in the source code.

Last working commit hash is `b4adada51`

PR #16936, opened 2018-12-05
[doc] Fix Apache URL rewriting

bizley:
This new rule must be first or `[L]` must be added to `RewriteRule . index.php` because it breaks URL rewriting otherwise.

https://stackoverflow.com/questions/53627613/yii2-rules-in-routing-section


samdark:
Merged. Thanks!

Issue #16891, opened 2018-11-16
Pagination::totalCount  initialized incorrectly


taobig:
### What steps will reproduce the problem?
```
        $dataProvider = new ActiveDataProvider([
            'query' => $query,
            'totalCount' => $totalCount,//the value is difference from $query->count()
            'pagination' => [
            ],
        ]);
        echo $dataProvider->pagination->page;//always 0
```

### What is the expected result?
 $dataProvider->pagination->page == $_GET['page']
### What do you get instead?
$dataProvider->pagination->page;//always 0
### Additional info
correct code：
```
        $dataProvider = new ActiveDataProvider([
            'query' => $query,
            'totalCount' => $totalCount,//the value is difference from $query->count()
            'pagination' => [
                totalCount' => $totalCount,// add this line
            ],
        ]);
```
or
```
        $dataProvider = new ActiveDataProvider([
            'query' => $query,
            'totalCount' => $totalCount,//the value is difference from $query->count()
            'pagination' => [
            ],
        ]);
        $dataProvider->getModels();// add this line
```
I found yii\data\Pagination::getPageCount()  to use `totalCount` without initialized.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.2
| Operating system |

samdark:
Yes. That's how it is currently. Any idea on how to fix it?

samdark:
@GHopperMSK would you please check regression that's caused by solving this task? https://github.com/yiisoft/yii2-elasticsearch/issues/201 Thanks!

samdark:
Fixed in in ElasticSearch. Probably a good idea to release it first.

PR #16951, opened 2018-12-08
Fix Pagination::totalCount initialized incorrectly #16891

taobig:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | maybe? Pagination::totalCount default value turn 0 to null.
| Tests pass?   | yes
| Fixed issues  | #16891


samdark:
Seems to break tests: https://travis-ci.org/yiisoft/yii2/jobs/465231209

framework/data/Pagination.php
```diff
@@ -126,7 +126,7 @@ class Pagination extends BaseObject implements Linkable
     /**
      * @var int total number of items.
      */
-    public $totalCount = 0;
+    public $totalCount = -1;
```
samdark: `-1` is a special value that decreases overall clarity of the class. We can change it to `null` instead but that would require manually initializing `totalCount` for the case where it's `0`.

framework/data/BaseDataProvider.php
```diff
@@ -165,7 +168,14 @@ public function getTotalCount()
     {
         if ($this->getPagination() === false) {
             return $this->getCount();
-        } elseif ($this->_totalCount === null) {
+        }
+
+        return $this->getTotalCountWithPagination();
+    }
+
+    private function getTotalCountWithPagination()
```
samdark: Extracting this method doesn't add much clarity. Especially the method name.

Also there's not much to reuse. I'd revert it and, instead, add two lines to `getPagination()` directly.

samdark (changes requested):

framework/data/BaseDataProvider.php
taobig: Yes, we can use `$this->_pagination->totalCount = $this->prepareTotalCount();` instead of `$this->_pagination->totalCount = $this->getTotalCountWithPagination();`. But we will calc totalCount twice, one is calculating Pagination.totalCount, the other is  calculating DataProvider.totalCount.

taobig: @samdark I made it, as you said.

PR #16956, opened 2018-12-09
Check if assets path is writable on publish (#15850)

aivchen:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15850


framework/web/AssetManager.php
```diff
@@ -211,8 +211,6 @@ public function init()
         $this->basePath = Yii::getAlias($this->basePath);
         if (!is_dir($this->basePath)) {
```
alexeevdv: What about this check? Shouldn't it be moved too? It seems like there is no reason to check `basePath` if nothing will be published there

Groonya: May be it should be moved, but the issue is about checking $basePath for writability only.

Also I've found one problem in test.
To emulate read only dir I run chmod($path, 0555) in the test. But after that developer can have a problem to switch to another branch. 
Can you suggest something? Thanks.

alexeevdv: You can create this folder in `setUp` and delete it in `tearDown` method

alexeevdv: Or you can `chmod` it back in `tearDown`

Groonya: Done, thanks.

samdark:
Merged. Thank you!

PR #16963, opened 2018-12-13
Fixing arabic guide wip issue

aneeshikmat:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes


docs/guide-ar/start-installation.md
```diff
@@ -87,12 +89,12 @@ composer create-project --prefer-dist yiisoft/yii2-app-basic basic
...
-<li> قم بتعديل ملف `config / web.php` عن طريق إدخال secret key ل` cookieValidationKey`
+<li> قم بتعديل ملف <code>config/web.php</code> عن طريق إدخال secret key ل<code> cookieValidationKey</code>
 (يتم ذلك تلقائيًا إذا قمت بتثبيت ال Yii باستخدام Composer): </li>
 </ol>
 
    ```php
-   // !!! insert a secret key in the following (if it is empty) - this is required by cookie validation
+  // !!! insert a secret key in the following (if it is empty) - this is required by cookie validation
    'cookieValidationKey' => 'enter your secret key here',
    ```
```
cebe: this code block should not be intended. in markdown that renders fine, but as you add HTML above it will not work anymore.

docs/guide-ar/start-installation.md
```diff
@@ -41,8 +41,10 @@ sudo mv composer.phar /usr/local/bin/composer
...
 <p dir="rtl">
     في هذا الدليل ، نفترض أنك قمت بتثبيت ال composer على مستوى جميع المشاريع (<a href="https://getcomposer.org/doc/00-intro.md#globally">globally</a>)  بحيث تكون أوامر ال composer متاحة لجميع المشاريع من أي مكان. أما إذا كنت تستخدم ال composer.phar لمسار محدد فقط(local directory)،  فيجب عليك ضبط  الأومر وفقًا لذلك.
+</p>
 
-إذا كان ال composer مثبتًا من قبل، فتأكد من استخدام إصدار حديث. يمكنك تحديث ال composer عن طريق تنفيذ الأمر التالي `composer self-update`
+<p dir="rtl">
+إذا كان ال composer مثبتًا من قبل، فتأكد من استخدام إصدار حديث. يمكنك تحديث ال composer عن طريق تنفيذ الأمر التالي <code>composer self-update</code>
 </p>
 
 <blockquote><p dir="rtl">
```
cebe: btw, for blockquotes that start with `Info:`, `Warning:` etc, you can add a CSS class to make them look better.
See https://github.com/yiisoft/yii2/blob/master/docs/documentation_style_guide.md#blocks for how this works in general. But it does not apply to your case as you are using HTML, not markdown, so you would need to add the class to the element:

- `<blockquote class="info">`
- `<blockquote class="warning">`
- `<blockquote class="note">`
- `<blockquote class="tip">`

docs/guide-ar/start-installation.md
aneeshikmat: Yes, that's true, I set it back to default now ..I think its change when I remove spaces between slashes.

docs/guide-ar/start-installation.md
aneeshikmat: I updated all blockqiute now by adding needed class.

samdark:
@aneeshikmat is it finished?

aneeshikmat:
@samdark I think yes, I resolved conversation issue.

samdark:
Merged. Thank you!

aneeshikmat:
Thanks

Issue #14272, opened 2017-06-08
Suggestion: Html::clearButton() for search and filter forms


robsch:
In [8606](8606) someone else has already thought that `Html::resetButton()` would work differently. I just stumbled over this too and I wondered if it would be possible to introduce something like `Html::clearButton()`. Although I know that it is the standard behaviour, the reset button of a form is quite useless, and it would be very helpful if there would be a way to clear a form and possibly submit it with the same click.

I think it is a very common requirement for search/filter inputs. And it would be also nice if gii could generate forms with such buttons. Could be also introduced for GridView filter inputs.

What do you think?

samdark:
While it makes sense I'm not sure such JavaScript should be in the framework core.

@yiisoft/core-developers opinions?

arogachev:
The way I usually use to reset a form is the button with a link to initial url.

For example, if you have applied filter something like this:

```
/enterprises/?region=1&district=1&settlement=9
```

You can use this link and after click on it the form will be reset:

```html
<a class="btn btn-default" href="/enterprises/">Reset</a>
```

So it does not require any additional JS.

klimov-paul:
I agreed with @arogachev : it should be a simple link. We already have `Url::canonical()`, which may help in such link composition, however automatic generation of such URL is too risky.
I do not consider putting such functionality into `Html` helper is valid. However filter reset button can be added to `GridView` widget as it already have a `filterUrl` field for this.

samdark:
Agree. Let's a add it to gridview and leave Html helper as is.

robsch:
Fine. Your're right, Html::clearButton() is not necessary if a button with an appropiate URL can be realized easily (current URL, just without the filter parameters). But the GridView button would be quite useful! Especially if gii could generate it automatically.

Just a question: is this GridView clear function then "pjax-able"?

samdark:
Depends on implementation.

rrmontuan:
Are we talking about creating a clearButton method in the GridView class or modifying the **_search** template in yii2-gii extension?

samdark:
GridView.

PR #16978, opened 2018-12-18
#14272 - Added a resetButton method for GridView

rrmontuan:
#14272 

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | Added resetButton Method for GridView

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 Change Log
 
 2.0.16 under development
 ------------------------
-
+- Enh #14272: added `yii\grid\GridView::resetButton()` method to generate a reset button for GridView (rrmontuan)
 - Bug #15850: check basePath is writable on publish in AssetManager (Groonya)
 - Bug #16910: Fix messages sorting on extract (Groonya)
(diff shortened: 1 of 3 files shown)


machour:
Hi @rrmontuan 

What is this reset button supposed to do? I don't see any logic behind it. It just seems to generate a `<a>` which I could do using `yii\helpers\Html`.

s1lver:
Hi!

@machour 
The discussion dealt with the addition of a button without implementation. Because automatic generation is too risky. Implementation can be added independently. If I understood correctly.

@rrmontuan 
I think it's worth using the button naming method.

```php
Yii::t();
```

rrmontuan:
> @machour
> The discussion dealt with the addition of a button without implementation. Because automatic generation is too risky. Implementation can be added independently. If I understood correctly.

Yeah @s1lver, that's why they asked for this implementation. 

> @rrmontuan
> I think it's worth using the button naming method.
> 
> ```
> Yii::t();
> ```

You're right, I'm going to adjust the method to use Yii::t() as you mentioned.

samdark:
Not going to be merged into 2.0.x. Original issue moved to 3.0.

PR #16988, opened 2018-12-26
Fixes #16028: valid serialization of complex cache keys that contain non-UTF sequences

### LINKED ISSUE #16028: \yii\caching\Cache::buildKey can generate identical cache keys for different $key attribute
So method is written like this:

    public function buildKey($key)
    {
        if (is_string($key)) {
            $key = ctype_alnum($key) && StringHelper::byteLength($key) <= 32 ? $key : md5($key);
        } else {
            $key = md5(json_encode($key));
        }

        return $this->keyPrefix . $key;
    }

The problem is in passing `json_encode` output directly to `md5` function.
It can return `false` in case of error. In my case it was an SQL query with special symbols, key generated from Yii2 components itself, when i used query caching. So it returned the same md5 key and caused multiple errors (same result for different queries)

Maybe it is better to throw an exception if output of `json_encode` is `false` in this case?
And i guess i need to create another ticket for `\yii\db\Command::queryInternal`?

rugabarbo:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16028

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #16028: Fix serialization of complex cache keys that contain non-UTF sequences (rugabarbo)
 - Bug #15850: check basePath is writable on publish in AssetManager (Groonya)
 - Bug #16910: Fix messages sorting on extract (Groonya)
 - Bug #16969: Fix `yii\filters\PageCache` incorrectly storing empty data in some cases (sammousa)
--- a/framework/caching/Cache.php
+++ b/framework/caching/Cache.php
@@ -79,6 +79,19 @@ abstract class Cache extends Component implements CacheInterface
      */
     public $defaultDuration = 0;
 
+    /**
+     * @var bool whether [igbinary serialization](http://pecl.php.net/package/igbinary) is available or not.
+     */
+    private $_igbinaryAvailable = false;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function init()
+    {
+        parent::init();
+        $this->_igbinaryAvailable = \extension_loaded('igbinary');
+    }
 
     /**
      * Builds a normalized cache key from a given key.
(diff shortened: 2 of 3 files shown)


samdark:
Merged. Thank you!

tyler-ham:
I am (slightly) of the opinion that this should have been considered potentially BC breaking, though I can understand how the potential impact was deemed to be small enough not to rise to the level of "breaking".

Changing the serialization of keys causes the new keys to no longer match entries that already exist in the cache. In terms of `get` operations, it is as if much of the cache was flushed. However, the old entries remain as orphans in the cache store until the whole cache is flushed (or the entries are evicted if the entries have expirations and the cache store supports automatic eviction).

I will work around it by implementing a `Cache` subclass to maintain the original key serialization algorithm or migrate entries to the new algorithm over time.

Perhaps key serialization should be configurable the way value serialization/deserialization is? Then developers could choose whichever method performs best for their use case (and could easily replace the default algorithm without having to write one or more Cache subclasses).

rugabarbo:
@tyler-hav, I do not understand what the problem is. Why you need to maintain the original key serialization? What exactly has stopped working correctly in your application?

tyler-ham:
Maintaining support for the original key serialization is a temporary solution to allow the keys to change slowly over time to avoid a sudden performance degradation due to a cold cache. To be clear, I don't believe that is something that should be implemented in the framework. The main reason for my comment is to document my experience with this change in case anyone else comes across it and is looking for a workaround to avoid essentially flushing most of the cache all at once in a production system after installing an updated Yii framework.

Secondarily, while it may not rise to the level of breaking backward-compatibility (as it does not change any function signatures, names, etc.), I would suggest making a note of changes like this that could have an impact on performance or resource usage, even if only temporarily, in the `UPGRADE.md` documentation.

samdark:
@tyler-ham noted for Yii 3.

PR #16993, opened 2018-12-28
Fix invalid count() and query Copy for inherited Query classes

sartor:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

It is impossible now to inherit db/Query class without copy-pasting of queryInternal method just to work with inherited class (instead of parent).

The only change is keyword _static_ instead of _self_

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -463,7 +463,7 @@ class Query extends Component implements QueryInterface, ExpressionInterface
             return $command->queryScalar();
         }
 
-        $command = (new self())
+        $command = (new static())
             ->select([$selectExpression])
             ->from(['c' => $this])
             ->createCommand($db);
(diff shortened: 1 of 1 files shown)


samdark:
https://travis-ci.org/yiisoft/yii2/jobs/473006535

machour:
@sartor it seems that this PR is breaking the tests:

```
Missing argument 1 for yii\db\ActiveQuery::__construct(), called in /home/travis/build/yiisoft/yii2/framework/db/Query.php on line 466 and defined
```

could you have a look?

sartor:
Looks like it can't be fixed such way, unfortunately.

samdark:
OK. Closing this pull request then. If you have more ideas on how to approach it, please do. Thanks.

PR #16995, opened 2018-12-29
Fixes #16081: fixed composite IN using just one column

### LINKED ISSUE #16081: Illegal offset type in isset or empty in InConditionBuilder
Set one column as array to in condition:
```
MyModel::find()
->andWhere(['in', ['id'], [['id' => 1, 'name' => 'Name1'], ['id' => 2, 'name' => 'Name2']]])
->all();
```
Return my models where id in (1, 2)

### What do you get instead?
Exception: Illegal offset type in isset or empty
```
yiisoft/yii2/db/conditions/InConditionBuilder.php  93
 $value = isset($value[$column]) ? $value[$column] : null;
```

### Some researches
In InConditionBuilder build function check count($column), normalize it by reset($column) and call buildValues.
```
public function build(ExpressionInterface $expression, array &$params = [])
...
if ($column instanceof \Traversable || ((is_array($column) || $column instanceof \Countable) && count($column) > 1)) {
    return $this->buildCompositeInCondition($operator, $column, $values, $params);
}

if (is_array($column)) {
    $column = reset($column);
}

$sqlValues = $this->buildValues($expression, $values, $params);
```
But in function buildValues $column is not normalized and equals to  `[ 0 => 'id']`
```
protected function buildValues(ConditionInterface $condition, $values, &$params)
{
    $sqlValues = [];
    $column = $condition->getColumn();
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.2
| Operating system | Ubuntu

rugabarbo:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16081

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #16081: Fixed composite IN using just one column (rugabarbo)
 - Bug #15850: check basePath is writable on publish in AssetManager (Groonya)
 - Bug #16910: Fix messages sorting on extract (Groonya)
 - Bug #16969: Fix `yii\filters\PageCache` incorrectly storing empty data in some cases (sammousa)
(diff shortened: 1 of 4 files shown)


samdark:
Merged. Thank you!

Issue #16966, opened 2018-12-14
Support ArrayExpression in ActiveRelationTrait


tQuant:
### What steps will reproduce the problem?
Create two tables in Postgresql
```
CREATE TABLE city
(
  id SERIAL PRIMARY KEY NOT NULL,
  name VARCHAR
)
CREATE TABLE group
(
  id SERIAL PRIMARY KEY NOT NULL,
  city_ids INTEGER[]
)
```
Create models with MANY relation

```
class City extends \yii\db\ActiveRecord {
    public static function tableName() {
        return 'city';
    }
}
class Group extends \yii\db\ActiveRecord {
    public static function tableName() {
        return 'group';
    }

    public function getCities() {
        return $this->hasMany(City::class, ['id' => 'city_ids']);
    }
}
```

Fill tables, load group and access relation
`$cities = $group->cities`
### What is the expected result?
$cities contain list of City models

### What do you get instead?
Error: Illegal offset type in isset or empty in yiisoft/yii2/db/ArrayExpression.php line 112

### Additional info
If you set
\yii\db\pgsql\ColumnSchema::$deserializeArrayColumnToArrayExpression = false
then it works fine

It can be fixed by adding check for ArrayExpression in ActiveRelationTrait
![fix](https://user-images.githubusercontent.com/7368485/49990862-ff6d3f00-ffa0-11e8-9f93-6fbeb7e72656.png)

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 5.6
| Operating system | Debian 8

GHopperMSK:
@samdark I confirm the problem and offered solution works fine. If there isn't any objections, I can do the job.

Maybe we should throw an Exception when `$value->getDimension() !== 1`?

samdark:
No objections.

PR #16998, opened 2018-12-31
Add ArrayExpression support in related tables

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | Fixed error which obtained during accessing to related data Array type. You can find user case in #16966

### DIFF
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -525,7 +525,7 @@ trait ActiveRelationTrait
                 if (($value = $model[$attribute]) !== null) {
                     if (is_array($value)) {
                         $values = array_merge($values, $value);
-                    } elseif($value instanceof ArrayExpression && $value->getDimension() === 1) {
+                    } elseif ($value instanceof ArrayExpression && $value->getDimension() === 1) {
                         $values = array_merge($values, $value->getValue());
                     } else {
                         $values[] = $value;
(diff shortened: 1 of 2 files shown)


framework/db/ActiveRelationTrait.php
```diff
@@ -525,6 +525,8 @@ private function filterByModels($models)
                 if (($value = $model[$attribute]) !== null) {
                     if (is_array($value)) {
                         $values = array_merge($values, $value);
+                    } elseif(($value instanceof ArrayExpression) && ($value->getDimension() === 1)) {
```
rob006:
```suggestion
                    } elseif ($value instanceof ArrayExpression && $value->getDimension() === 1) {
```

GHopperMSK: No, unfortunately it doesn't help

rob006:
I would ignore this Codeclimate complaining - there is no point to create code less readable code for humans only to satisfy SCA tool.

GHopperMSK:
> I would ignore this Codeclimate complaining - there is no point to create code less readable code for humans only to satisfy SCA tool.

In this case I think I should undo last commit due to it isn't such readable for human as previous one.

SilverFire (approved):
Thank you for a pull request. It looks perfectly good to me

Issue #16941, opened 2018-12-06
migration generation with prefix


Isitar:
is there a reason why the migration templates (createTableMigration.php, addColumnMigration.php etc.) do not have the table prefix syntax `{{%tablename}}` ?

### What steps will reproduce the problem?
php yii migrate/create add_xxx_column_to_yyy_table --fields="xxx:integer"
### What is the expected result?
```
$this->addColumn('{{%yyy}}', 'xxx', $this->integer());
```
### What do you get instead?
```
$this->addColumn('yyy', 'xxx', $this->integer());
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.2
| Operating system |Windows

rob006:
There is a switch for enabling prefix in generated migrations.

https://github.com/yiisoft/yii2/blob/379a2002dd8af037d86e1db1a6a1607f0775478a/framework/console/controllers/MigrateController.php#L115

 The question is why it is disabled by default and why we even need it - does adding prefix everywhere has any drawbacks?

rob006:
I'm for changing default value of `$useTablePrefix` to `true`. We could also mark this property as deprecated, but I'm not sure if it makes any sense.

PR #16999, opened 2019-01-01
16941 migration with prefix

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16941 Database prefix was added to migration templates.


rob006:
Prefix is already handled in:

https://github.com/yiisoft/yii2/blob/379a2002dd8af037d86e1db1a6a1607f0775478a/framework/console/controllers/MigrateController.php#L467-L474

PR #17003, opened 2019-01-02
Fixes AssetBundleTest for vagrant ENV

rugabarbo:
1. There is no need to use setUp/tearDown just for one TestCase method.
2. Previous implementation of testBasePathIsWritableOnPublish() doesn't work in vagrant environment using shared files because of unreliable chmod.

So, this implementation is more versatile and optimized.


samdark:
Merged. Thanks!

Issue #16991, opened 2018-12-27
Usage of utf8_encode() in \yii\web\Request::resolvePathInfo() makes Yii depend on the php-xml extension


MysteryDragon:
### What steps will reproduce the problem?
Just try to route on something like `/test%F2%FB` without mentioned extension.

### What is the expected result?
Well, 404

### What do you get instead?
Ta-daaa, 500

### Additional info

https://github.com/yiisoft/yii2/blob/572e5f6655f279df574fe4937406ec04a6852cbc/framework/web/Request.php#L902-L926

It happens because `utf8_encode()` moved in core only in PHP 7.2.

I was intrigued, so I open https://yiipowered.com, and start to test :)
First catch: https://craftcms.com/test%F2%FB

| Q                | A
| ---------------- | ---
| Yii version      | any (catched in 2.0.14)
| PHP version      | up to 7.1 (catched in 7.1)
| Operating system | catched in CentOS 7 with PHP from remi repo

P.S.: Perhaps I did a lot of grammar mistakes in my sentences. Sorry.)

samdark:
We can instead use something like https://github.com/symfony/polyfill-php72/blob/master/Php72.php#L24

GHopperMSK:
Clear installation. Why don't I get 500?
![screenshot_20190102_234029](https://user-images.githubusercontent.com/19670887/50611585-0b855600-0ee8-11e9-8298-a2fe96394559.png)

samdark:
Clear installation of PHP from package manager?

PR #17005, opened 2019-01-03
Add Php72 helper

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | Static method `Php72::utf8_encode()` was added and used to solve #16991 bug.


MysteryDragon:
I don't understand why not to simply require that polyfill. Not so much code in that class, requirements good for Yii 2. Doesn't make sense to introduce one more helper for that.

MysteryDragon (commented):
I don't understand why not to simply require that polyfill. Not so much code in that class, requirements good for Yii 2. Doesn't make sense to introduce one more helper for that, and copy-paste that code.

rob006:
> Doesn't make sense to introduce one more helper for that, and copy-paste that code.

I would say that it does not make sense to require the whole polyfill only for one method used in one place. It is highly unlikely that this code will be ever changed, so there is no much hassle with maintaining copy-pasted code.

However I would either move it to string helper or to `Request` as private method.

Dzhuneyt (approved):

samdark:
Moving to Request as private method makes sense to me.

framework/web/Request.php
```diff
@@ -922,7 +922,7 @@ protected function resolvePathInfo()
             | \xF4[\x80-\x8F][\x80-\xBF]{2}      # plane 16
             )*$%xs', $pathInfo)
         ) {
-            $pathInfo = utf8_encode($pathInfo);
+            $pathInfo =  $this->utf8Encode($pathInfo);
```
MysteryDragon: One extra unnecessary space, I guess.

framework/web/Request.php
```diff
@@ -944,6 +944,25 @@ protected function resolvePathInfo()
         return (string) $pathInfo;
     }
 
+    /**
+     * Encodes an ISO-8859-1 string to UTF-8
+     * @param string $s
+     * @return string the UTF-8 translation of <i>s</i>.
```
MysteryDragon: I suggest to point source class by `@see` tag; also, I'm not sure about `<i>` - maybe markdown will work better?

framework/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #16991: Replace `utf8_encode()` by `Php72::utf8_encode()` (GHopperMSK)
```
MysteryDragon: Is this reference correct? Or better `Request::utf8Encode()`?

Dzhuneyt:
```suggestion
- Bug #16991: Replace `utf8_encode()` by `Request::utf8_encode()` (GHopperMSK)
```

rob006: This is private method, it should not be mentioned in changelog, since it cannot be used. ```Removing usage of `utf8_encode()` from `Request::resolvePathInfo()` ``` should be enough.

samdark:
Merged. Thank you!

frontbear:
@samdark  this PR should NOT be merged. why not just `composer require symfony/polyfill-php72` ?

PR #17006, opened 2019-01-03
Fixes CommandTest::testCreateView() cleanup logic

rugabarbo:
View must be deleted before deleting the tables on which it depends, otherwise we will get errors when rerunning the tests in the same environment.

For example:
```
1) yiiunit\framework\db\mysql\CommandTest::testCreateView
yii\db\Exception: SQLSTATE[HY000]: General error: 1356 View 'yii2_test.testCreateView' references invalid table(s) or column(s) or function(s) or definer/invoker of view lack rights to use them
The SQL being executed was: SHOW FULL COLUMNS FROM `testCreateView`
```


samdark:
Merged. Thanks!

Issue #17014, opened 2019-01-04
Query Builder - Wrong quoting when using decimal value in select


tsanchev:
### What steps will reproduce the problem?
```
...
(new Query())
            ->select([
                'overtime' => '1.3',
            ])
...

```
or
```
...
(new Query())
            ->select([
                "'1.3' AS `overtime`",
            ])
...

```
### What is the expected result?

the generated query to be:

```
SELECT '1.3' AS `overtime` ...
```

### What do you get instead?

the generated query is:

```
SELECT `1`.`3` AS `overtime` ...
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16-dev
| PHP version      | 7.2.10

rob006:
I don't think that this is a bug - if you want avoid quoting you should use `Expression` - this is explained in `select()` documentation. See https://github.com/yiisoft/yii2/pull/17017/files#r245482467

samdark:
Second case isn't a bug as @rob006 noted.

@tsanchev what's the problem with 

```
(new Query())
            ->select([
                'overtime' => '1.3',
            ])
```

?

tsanchev:
@samdark With both i've got the same result, but I guess i should use Expression so it's not a bug

PR #17017, opened 2019-01-04
17014 quoting fix

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17014

### DIFF
--- a/tests/framework/db/QueryBuilderTest.php
+++ b/tests/framework/db/QueryBuilderTest.php
@@ -1633,6 +1633,7 @@ abstract class QueryBuilderTest extends DatabaseTestCase
         $query = (new Query())
             ->select([
                 'ID' => 't.id',
+                'numeric' => '1.3',
                 'gsm.username as GSM',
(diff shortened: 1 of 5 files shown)


samdark:
Seems to fail tests: https://travis-ci.org/yiisoft/yii2/jobs/475521053

tests/framework/db/QueryBuilderTest.php
```diff
@@ -1633,6 +1633,7 @@ public function testComplexSelect()
         $query = (new Query())
             ->select([
                 'ID' => 't.id',
+                'numeric' => '1.3',
```
rob006: This seems to be incorrect. Array values should be column names, not values - this creates inconsistency how this array is interpreted. Not mention that `1.3` may be valid column name. 

If you want something like `1.3 as numeric` you should use `Expression`.

PR #17019, opened 2019-01-05
Fix of Oracle tests for proper skipping

rugabarbo:
PHPUnit must get the data that will be passed to the tests, before executing them. That's why you can't prevent the dataProvider to be executed using `$this->markTestSkipped()` inside a test. Instead, you must use a skipping inside data provider when the test cannot be executed.

Now Oracle tests are skipped without fails when Oracle isn't installed:
```
$ vendor/bin/phpunit --filter=BuildLikeCondition --group=oci --verbose

...

There was 1 skipped test:

1) yiiunit\framework\db\oci\QueryBuilderTest::testBuildLikeCondition
Test for yiiunit\framework\db\oci\QueryBuilderTest::testBuildLikeCondition skipped by data provider
Could not execute Connection::quoteValue() method: Connection::dsn cannot be empty.
```


samdark:
Merged. Thanks!

PR #17021, opened 2019-01-06
Fix to do not delete used message files with nested category names

albertborsos:
| Q             | A
| ------------- | ---
| Is bugfix?    |yes, only in 2.0.16-dev
| New feature?  |no
| Breaks BC?    |no
| Tests pass?   | yes
| Fixed issues  | #16580 #16935


samdark:
Merged. Thanks!

PR #17022, opened 2019-01-06
Fixes #15482: AR::find()->with() missing data when using string identifiers for relations

### LINKED ISSUE #15482: [yii\db\ActiveRelationTrait] ActiveRecord Relation Query Missing Some Data
https://github.com/yiisoft/yii2/blob/1e0157f0acbcbb5440d3f7a1569ede48593d9098/framework/db/ActiveRelationTrait.php#L524

Why here use `array_unique($values, SORT_REGULAR)` instead of `array_unique($values)`? it cause some problem in my case. 

Two tables are designed as below:

Table A:

id  | name
---| -------
1   | 001
2   | 0001

Table B：

id  | name
---| ------
1   | 001
2   | 0001

In my case, i use the relation `hasOne` in the Table A ActiveRecord like below:

```php
public function getB() {
    $this->hasOne(TableB::className(), ['name' => 'name'])
}
```

and for a better query performance, i use function `with('b')` to reduce query times. But from the db log, i got the sql:

```sql
select * from B where name = '001'
```

not the expectant one:

```sql
select * from B where name in ('001', '0001')
```

I don`t know if this is a bug or It is intentionally.

rugabarbo:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15482


rob006:
Is it possible to have objects (like `yii\db\Expression`) in `$values`?

rugabarbo:
HHVM tests failed with error:
> Array to string conversion

Example: https://travis-ci.org/yiisoft/yii2/jobs/476034932

samdark:
HHVM doesn't matter much as it's a dead-end branch of PHP.

rugabarbo:
> Is it possible to have objects (like yii\db\Expression) in $values?

@rob006 thank you for your comment. I checked it. Indeed, each `$value` can be not only a scalar, but also an array of scalars, and `Expression` object. I implemented a separation (8969de03), as a result of which the filtering of unique values applies only to scalar types. As for the rest of the data types, I propose to omit the verification of their uniqueness, because this may reduce performance. The most frequent case is handling scalars. 

> HHVM doesn't matter much as it's a dead-end branch of PHP.

HHVM doesn't matter much, but it gave me a hint that the array can contain values of different types, including arrays ;)

samdark (approved):

samdark:
Merged. Thank you!

PR #17024, opened 2019-01-08
Fixed random fails of RetryAcquireTraitTest

rugabarbo:
Property `RetryAcquireTrait::$retryDelay` sets the minimum delay, but not the exact delay. In fact, delay may be longer. Because of this, sometimes less than 20 blocking attempts occur. Checking the exact number of locks sometimes leads to the following errors:
```
$ vendor/bin/phpunit --filter=testRetryAcquire

F

You should really fix these slow tests (>500ms)...
 1. 1050ms to run yiiunit\framework\mutex\RetryAcquireTraitTest:testRetryAcquire

Time: 4.08 seconds, Memory: 44.00MB

There was 1 failure:

1) yiiunit\framework\mutex\RetryAcquireTraitTest::testRetryAcquire
Failed asserting that 19 is identical to 20.

/yii2/tests/framework/mutex/RetryAcquireTraitTest.php:36
/yii2/vendor/phpunit/phpunit/phpunit:52

FAILURES!
Tests: 1, Assertions: 3, Failures: 1.
```

I reworked the test so that it checks duration of delays, but not the exact number of blocking attempts.

### DIFF
--- a/tests/framework/mutex/RetryAcquireTraitTest.php
+++ b/tests/framework/mutex/RetryAcquireTraitTest.php
@@ -33,7 +33,17 @@ class RetryAcquireTraitTest extends TestCase
         $this->assertTrue($mutexOne->acquire($mutexName));
         $this->assertFalse($mutexTwo->acquire($mutexName, 1));
 
-        $this->assertSame(20, $mutexTwo->attemptsCounter);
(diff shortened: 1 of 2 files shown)


samdark:
Merged. Thanks!

PR #17025, opened 2019-01-08
Fast fix of nightly build

rugabarbo:
This is just a quick fix. It would be correct to use for the test a component that doesn't require graphic extensions.

### DIFF
--- a/tests/framework/di/ServiceLocatorTest.php
+++ b/tests/framework/di/ServiceLocatorTest.php
@@ -92,6 +92,10 @@ class ServiceLocatorTest extends TestCase
      */
     public function testModulePropertyIsset()
     {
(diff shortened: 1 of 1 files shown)


rob006:
I'm for using different component for this test - it is a test for `ServiceLocator`, it not necessary needs to be captcha.

samdark:
@rob006 idea sounds better. @rugabarbo could you adjust it?

rugabarbo:
Already working on it. Strange, but with other components the test failed. We may have found another error :)

rugabarbo:
Understood. I tried to use properties of components that are not declared. 

**Fixed.**

samdark:
Merged. Thanks!

PR #17053, opened 2019-01-20
ActiveRecord::findOne() now accepts quoted table and column names

### LINKED ISSUE #15931: FindOne error when it using with an associative array or innerJoin

SilverFire:
Fixes #15931


samdark:
Merged. Thank you very much!

Issue #17049, opened 2019-01-18
Asset bundling broken with 2.0.16-dev


schmunk42:
### What steps will reproduce the problem?

Bundle assets with `yii` command

### What is the expected result?

Bundled assets

### What do you get instead?

```
Creating output bundle 'all':
  Compressing JavaScript files...
Parse error at /app/src/../web/bundles/common-temp.js.tmp:14339,20
                let $input = findInput($form, this);
                    ^
ERROR: Unexpected token: name ($input)
    at JS_Parse_Error.get (eval at <anonymous> (/usr/lib/node_modules/uglify-js/tools/node.js:20:1), <anonymous>:71:23)
    at fatal (/usr/lib/node_modules/uglify-js/bin/uglifyjs:291:52)
    at run (/usr/lib/node_modules/uglify-js/bin/uglifyjs:235:9)
    at Object.<anonymous> (/usr/lib/node_modules/uglify-js/bin/uglifyjs:160:5)
    at Module._compile (module.js:577:32)
    at Object.Module._extensions..js (module.js:586:10)
    at Module.load (module.js:494:32)
    at tryModuleLoad (module.js:453:12)
    at Function.Module._load (module.js:445:3)
    at Module.runMain (module.js:611:10)
Error: Unable to compress JavaScript files into '/app/src/../web/bundles/common-temp.js'.
2019-01-18 12:04:29 [-][-][-][error][yii\console\Exception] yii\console\Exception: Unable to compress JavaScript files into '/app/src/../web/bundles/common-temp.js'. in /app/vendor/yiisoft/yii2-dev/framework/console/controllers/AssetController.php:527
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16-dev
| PHP version      | 7.2
| Operating system | Debian (Docker)

I ran this as an integration test for 2.0.16 - see also https://github.com/orgs/yiisoft/projects/14

While this is an error in an external JS-tool, we still should pick another variable name here. Introduced in https://github.com/yiisoft/yii2/commit/c59df914c1f544889eafc4d8aa64132ae1337102

machour:
Just ran into this as well.

machour:
I think that the problem is that the code uses `let` instead of `var`.

samdark:
@schmunk42 please recheck.

schmunk42:
:+1: Build is :green_heart: https://travis-ci.org/dmstr/phd5-app/builds/482395361

PR #17055, opened 2019-01-21
Use var instead of let

machour:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17049 

Introduced in c59df91


samdark:
Merged. Thank you!

Issue #16268, opened 2018-05-13
Rare bug-case: timestamp is not appending for some specific AssetBundle


MysteryDragon:
Well, it's funny bug which I firstly faced about 2 years ago :)
But when I just decided that "it's just not working well for now, will wait and continue use my reinvented wheel".
But second time I faced it few months ago, and finally found a reason. But for now I don't have so many time, and this was delayed again.
Now let me at least tell about that. :)

### What steps will reproduce the problem?
1. Create asset bundle without specifiyng `basePath` value (just omit that property).
2. Set something simple like:
```
'css' => [
    'css/site.css',
],
```
3. Enable `appendTimestamp` in AssetManager.

### What is the expected result?
Well, why not! Resources registered without troubles, so appending timestamp must work well.
So, it will be like `/css/site.css?v=1526200000`

### What do you get instead?
Nope, not for this case.

### Reasons

1. Here `basePath` sets as `null`:
https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L314

2. There aren't such file at system root directory (`/<filename>`) of course:
https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L322

### Proposals

IMO if `basePath`/`baseUrl` property definitions in asset bundle class are *actually* required then there must be thrown some exception when there aren't any value during initialization.
Or there may be announced some default values (`basePath` = `@webroot`, `baseUrl` = `@web`?).

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | >=2.0.3, as I can see
| PHP version      | any
| Operating system | any

samdark:
@MysteryDragon would you please check if https://github.com/yiisoft/yii2/commit/96cd8bcb687d66ae86fba01ac6a3e9c54687d616 fixed it?

MysteryDragon:
@samdark will try tomorrow

MysteryDragon:
@samdark it seems that it works well now!

PR #17060, opened 2019-01-22
fix timestamp appending if basePath is not defined

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16268

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #16268: Fixed asset timestamp appending when basePath is not defined (GHopperMSK)
 - Bug #15482: AR::find()->with() missing data when using string identifiers for relations (rugabarbo)
 - Bug #15167: Fixed loading of default value `current_timestamp()` for MariaDB >= 10.2.3 (rugabarbo, bloodrain777, Skinka)
 - Bug #16253: Fixed empty checkboxlist validation (GHopperMSK)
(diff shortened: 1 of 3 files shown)


GHopperMSK:
It looks like @webroot alias doesn't exists in console environment.

rob006:
It does not look right to me. AFAIK #16268 is about configuration errors which are ignored instead of throwing an exception. Blind guess of correct path in case of misconfiguration does not look like correct solution...

GHopperMSK:
What do you mean "blind guess"? All I have done is just check if I able to use absolute path or only relative one to get the file time.

rob006:
Using relative path is a blind guess - result may differ depending on entry script position / current working directory and may give random results in some edge cases.

GHopperMSK:
We don't have to specify `basePath` or `baseUrl`:

https://github.com/yiisoft/yii2/blob/5544d31bb3b7b0959ef51f7c78c7e06c8f5b354f/framework/web/AssetBundle.php#L184-L186

In case relative path gives us "random result" we just don't get timestamp postfix in asset URL. But in common case we always get exactly what we expected.

GHopperMSK:
As an option, I can throw an exception if `$timestamp` equal to `false`.

rob006:
> We don't have to specify `basePath` or `baseUrl`:
> https://github.com/yiisoft/yii2/blob/5544d31bb3b7b0959ef51f7c78c7e06c8f5b354f/framework/web/AssetBundle.php#L184-L186

In this case `basePath` and `baseUrl` are inherited from asset manager. This is not the same as using relative paths.

GHopperMSK:
As you wish! Let me know about your decision. It is not a problem to completely remake the job, but I need understand what I should to do.

samdark:
Alike fix is in master now. https://github.com/yiisoft/yii2/commit/96cd8bcb687d66ae86fba01ac6a3e9c54687d616

Issue #16158, opened 2018-04-25
yii.activeForm.js multiple select validation problem


procin:
When you have a multiple dropDownList like:

`<?= $form->field($model, 'bla')->dropDownList(['test' => 'test', 'foo' => 'foo'], ['multiple' => 'multiple']) ?>`

And the field is validated (it is for example required)
If you go to the view containing the form, and another field is validated (changed of blurred) then this field is also validated.
This is due the getValue function in yii.activeForm.js
This function gets the value for the field as array like ["test", "foo"] (when both selected)
But in the function "validateAttribute" also in yii.activeForm.js, the old value and new value are compared like:

`if (this.value !== getValue($form, this)) {`

This will always match when arrays are compared in javascript, so the field will be always validated even when there are no changes.

SilverFire:
Thank you for the reported issue. Would you like to help up to fix this issue by submitting a pull request? We have [contributors guide](https://github.com/yiisoft/yii2/blob/master/docs/internals/git-workflow.md) that is a good starting point. If you need any tip – let us know.

procin:
Thanks for your invitation! I'm reasonably new with Yii2, but very impressed by the quality of the code and the way the documentation is made.
So I want to make compliments to the complete team. Maybe after I have some more experience with Yii2, I will be of real use for the community.

samdark:
Why waiting? Try it. We'll correct the pull request to make sure quality is fine.

PR #17065, opened 2019-01-23
16158 fix multiple select validation

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16158


GHopperMSK:
I am not sure we should follow current recommendation of codeclimate. It will increase the methods complexity. What do you think?

samdark:
I think in this case it is better to ignore it.

framework/assets/yii.activeForm.js
```diff
@@ -565,6 +565,70 @@
...
+        if (Array.isArray(val1)) {
+            return isArraysEqual(val1, val2);
+        }
+
+        // simple types
+        return val1 === val2;
+    };
+
+    var isObjectsEqual = function(obj1, obj2) {
+        if (!(obj1 instanceof Object)) {
```
samdark:
```suggestion
        if (!(obj1 instanceof Object) || !(obj2 instanceof Object)) {
```

framework/assets/yii.activeForm.js
```diff
@@ -565,6 +565,70 @@
...
+            if (obj1[keys1[i]] !== obj2[keys1[i]]) {
+                return false;
+            }
+        }
+
+        return true;
+    };
+
+    var isArraysEqual = function(arr1, arr2) {
+        if (!Array.isArray(arr1)) {
```
samdark:
```suggestion
        if (!Array.isArray(arr1) || !Array.isArray(arr2)) {
```

framework/assets/yii.activeForm.js
```diff
@@ -565,6 +565,70 @@
...
+     * @returns boolean
+     */
+    var isEqual = function(val1, val2) {
+        // objects
+        if (val1 instanceof Object) {
+            return isObjectsEqual(val1, val2)
+        }
+
+        // arrays
+        if (Array.isArray(val1)) {
```
samdark: Arrays are objects as well. Is special case necessary?

samdark (changes requested):

framework/assets/yii.activeForm.js
GHopperMSK: Yes, we do need the special case due to different approach to compare them.

samdark:
Merged. Thank you!

PR #17073, opened 2019-01-25
introduce prepareQuery() method

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | https://github.com/yiisoft/active-record/issues/63

Usage:
```
        $provider = new ActiveDataProvider([
            'query' => City::find(),
            'pagination' => [
                'pageSize' => 2,
            ],
        ]);

        var_dump($provider->prepareQuery()->createCommand()->getRawSql());
```
Now we can easily get final SQL-query which will be passed to DB for current page data obtaining.
All changes where made in `ActiveDataProvider` class only due to they are useless in other data sources.


framework/data/ActiveDataProvider.php
```diff
@@ -97,23 +97,36 @@ public function init()
...
     protected function prepareModels()
+    {
+        $query = $this->prepareQuery();
+
+        return $query ? $query->all($this->db) : [];
+    }
+
+    /**
+     * Prepares the sql-query that will get the data for current page.
+     * @return array|QueryInterface
```
rob006: It probably should be `QueryInterface|null`. Or even just `QueryInterface` if you use `emulateExecution()` instead of returning `null`/`false`.

GHopperMSK: I like this idea, but I think we also should add method `getEmulateExecution` for testing purposes.

rob006: Why?

GHopperMSK: https://github.com/yiisoft/yii2/blob/5544d31bb3b7b0959ef51f7c78c7e06c8f5b354f/tests/framework/db/UnqueryableQueryMock.php#L13-L30

GHopperMSK: It throws the exception in any case. The only way to pass this test is not to call this methods, but we can't achieve it with current realization of emulateExecution

GHopperMSK:
Who can fix Travis?

samdark:
As enhancement it was planned for 3.x versions...

samdark:
Closing since issue was moved to https://github.com/yiisoft/active-record/issues/63

Issue #17071, opened 2019-01-25
ArrayExpression and ActiveRecord::getDirtyAttributes


tQuant:
### What steps will reproduce the problem?
Load model with non empty array field.

$mail = MyTable::findOne(1);
$mail->array_field->offsetSet(0, 'qwe');
$mail->update(false);

where "$mail->array_field" is ArrayExpression

### What is the expected result?
First item in "array_field" in the database changed to "qwe"

### What do you get instead?
"array_field" in the database has not changed at all

### Additional info
The reason is that in the "getDirtyAttributes" method, the old and the new values are compared through "==="
Which returns **true** if comparing the same instance of the same class.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 5.6
| Operating system | debian 8

Issue #5375, opened 2014-10-06
\yii\base\Model::loadMultiple + save() generate redundant queries


gephaest:
I was trying to use Multiple models with single form and found strange behavior.
For example: I've form with many load models for edit. I'm changing one row (one model) and saving form. Yii in theory it should execute **one** UPDATE query, but it update **all** models (if page include 100 models will be 100 UPDATES).
![screenshot_34](https://cloud.githubusercontent.com/assets/1038708/4524244/b332d8e0-4d41-11e4-999c-7410d613f617.png)

My code similar to [this](http://www.yiiframework.com/doc-2.0/guide-input-forms.html#handling-multiple-models-with-a-single-form) (Handling multiple models with a single form section).

I think Yii can be smarter and update only changed rows.

Erik-r:
@samdark @cebe The issue is failing dirty attributes in AR has been discussed in #5911 and #2790 already.

I personally think it's not done by a documentation. The framework should be able to handle it correctly out of the box because all necessary information is available for the framework. See my thoughts:

Dirty attributes works only if numeric attributes are type casted with rules and strings, texts, dates, times, datetimes are nulled if form delivers an empty string. So each developer has to write rules for all attributes. This is the proposal of samdark. My comment: a) the framework has all knowledge about the attribute type in database. b) Even if you document this most developers would forget this rules.

Often it's not an issue if one record is stored with same values. But then Yii doesn't need dirty attributes. In the above case the storage is inefficient. In case of my app I use an inherited ActiveRecord class which stores data temporal (see http://en.wikipedia.org/wiki/Temporal_database). if a user push save button without a change and if dirty attributes fails the same record is written again with new valid from date.

As qiangxue wrote in #2790 the comparison of new and old attributes can't be done by unprecise comparator (0 != null for numeric values) what is correct from my opinion. Therefore I have overridden the method `getDirtyAttributes`. I use the column type of table schema to decide if the comparison is strict or not. smallint with one character (= bool) is handled unstrict. I know it's an unbeautiful hack, but it works. see part of code:

```php
        $attributes = [];
        $checkAttributes = $this->getAttributes();
        if ($this->getOldAttributes() === null) {
            foreach ($checkAttributes as $name => $value) {
                if (isset($names[$name])) {
                    $attributes[$name] = $value;
                }
            }
        } else {
            $schema = static::getTableSchema();
            $unpreciseTypes = ['void', 'string', 'date', 'datetime', 'time', 'text'];
[...]

PR #17077, opened 2019-01-27
17071 fix BaseActiveRecord::populateRecord() method

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17071, #15265

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
+- Bug #17071: `BaseActiveRecord::populateRecord()` method now clones values instead of coping if they aren't scalar values (GHopperMSK)
 - Bug #15482: AR::find()->with() missing data when using string identifiers for relations (rugabarbo)
 - Bug #15167: Fixed loading of default value `current_timestamp()` for MariaDB >= 10.2.3 (rugabarbo, bloodrain777, Skinka)
 - Bug #16253: Fixed empty checkboxlist validation (GHopperMSK)
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.16 under development
 ------------------------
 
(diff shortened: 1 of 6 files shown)


rob006:
This will break `getDirtyAttributes()` and `isAttributeChanged()` behavior - `===` will return `false` if you compare 2 cloned object (even if they're identical).

GHopperMSK:
> This will break `getDirtyAttributes()` and `isAttributeChanged()` behavior - `===` will return `false` if you compare 2 cloned object (even if they're identical).

I think we should introduce `isEqual` method, which compares two values and returns boolean. If the method obtain objects, it will compare them member by member.

rob006:
> I think we should introduce `isEqual` method

Related: #5375

GHopperMSK:
Maybe we should keep `isEqual` method somewhere in `Helper`? It can be useful in many cases!

GHopperMSK:
I decided to get rid of `identical` parameter in `isAttributeChanged` method. Possibly, it isn't a good decision, but I didn't find usage this parameter within whole the source.

GHopperMSK:
[Manual](http://php.net/manual/en/language.oop5.object-comparison.php) tells us

> When using the comparison operator (==), object variables are compared in a simple manner, namely: Two object instances are equal if they have the same attributes and values (values are compared with ==), and are instances of the same class

Why don't we compare them by native method '$obj1 == $obj2'?

rob006:
Because it will use `==` also for comparing properties, while we need to use `===` in this case.

GHopperMSK:
@rob006 , would you like to comment the current statue? Also, I need a clue to deal with Cognitive Complexity.

terabytesoftw:
Closing this PR as stale/outdated. The underlying issue #17071 may still need a focused fix, but this PR is no longer suitable for merge: it is not mergeable, mixes unrelated #15265 changes already fixed later, removes the public `$identical` argument from `isAttributeChanged()` causing a BC break, and the object comparison approach was not accepted in review.

Issue #14553, opened 2017-07-28
Routing Redirect


Knight-Yoshi:
Would you consider adding to urlManager 301/302 redirect support for the rules. This way it doesn't have to go to a controller/action of something like a deprecated path to be redirected by that action.
```
[
   'pattern' => 'some/old/route/',
   'redirect' => 'some/new/route/', // could accept parameters like how the route prop does
   'statusCode' => 301 // optional, but defaults to 301
]
```

dynasource:
agree. This is useful to have

cebe:
we have UrlNormalizer for that. Could probably be extended to make it easier to support complete rewrite of Urls. On the new website we currently have the following, which could benefit from such a feature:

- https://github.com/yiisoft-contrib/yiiframework.com/blob/d4374342766a031ca1dde6fb170f8e6f0c1328ac/config/urls.php#L90-L95
- https://github.com/yiisoft-contrib/yiiframework.com/blob/d4374342766a031ca1dde6fb170f8e6f0c1328ac/controllers/SiteController.php#L40-L67

samdark:
Agree that it could be useful.

PR #17091, opened 2019-02-01
routing redirect

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #14553

I would like to implement this feature on Yii2 and only after that transfer the code to Yii3 branch.

### DIFF
--- a/framework/web/UrlRule.php
+++ b/framework/web/UrlRule.php
@@ -163,10 +163,6 @@ class UrlRule extends BaseObject implements UrlRuleInterface
      */
     private $_routeRule;
     /**
-     * @var string the regex for matching the redirect part. This is used in generating URL.
-     */
-    private $_redirectRule;
-    /**
      * @var array list of regex for matching parameters. This is used in generating URL.
      */
     private $_paramRules = [];
@@ -174,10 +170,6 @@ class UrlRule extends BaseObject implements UrlRuleInterface
      * @var array list of parameters used in the route.
      */
     private $_routeParams = [];
-    /**
-     * @var array list of parameters used in the redirect.
-     */
-    private $_redirectParams = [];
 
     /**
      * @return string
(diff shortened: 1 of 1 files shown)


rob006:
I don't think that redirection should be done by rule - redirecting is `UrlNormalizer` job.

GHopperMSK:
I thought about this, but every time I go through
https://github.com/yiisoft/yii2/blob/fe3ae964cdd439e7f96a99cf9d625cf7d8ca2f2c/framework/web/UrlRule.php#L440-L443
I get `$normalized` equal to false.
Here is my `web.php` config:
```
...
'urlManager' => [
    'enablePrettyUrl' => true,
    'showScriptName' => false,
    'rules' => [
        [
            'pattern' => 'posts/<year:\d{4}>/<category>',
            'redirect' => 'redirect_destination/<category>/<year>',
            'statusCode' => 301,
        ],
    ],
],
...
```

samdark:
Sorry but enhancements aren't accepted into 2.0.

GHopperMSK:
All I wanted is just find out the solution and after that implement it over v3.0 I even didn't make any changes in `CHANGELOG.md` file due to I knew it will not be merged.

Now I have only one question - do we need this feature implementation with configuration as was described in #14553? As I can understand, the configuration doesn't imply `UrlNormalizer` existence?

rob006:
Considering https://github.com/yiisoft/yii-web/issues/5 and https://github.com/yiisoft/yii-core/issues/165 the whole idea may be obsolete - all you need is parse-only rule with simple handler for redirection.

rob006:
And from `UrlNormalizer` perspective it may be  fixed in some more generic way:

1. Parse request URL.
2. Try create URL using current route and GET data.
3. If current URL is different that generated one - redirect.

In this way URL will be always normalized to one version. If you want to support some old URLs, all you need is to put old rule after new one - new rule will have precedence for creating URLs and old URL will redirect to new URL.

GHopperMSK:
Thank you. It is clear now.

PR #17092, opened 2019-02-01
Added `yii\widgets\ActiveForm::validateEmptySelects` and the same att…

tafid:
…ribute to `yii.activeForm.js` which determines whether to validate SELECT without OPTION

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17092


hiqsol (approved):

samdark:
Thanks for your contribution but enhancements aren't accepted for 2.0.x. We can do the change in 3.0 though.

SilverFire (changes requested):
Considering it as bugfix. Please, add tests

Issue #17093, opened 2019-02-01
Wrong DataProvider->totalCount


anditza:
Wrong DataProvider->totalCount 

### What steps will reproduce the problem?
```php
$query = MyModel::find();
$dataProvider = new ActiveDataProvider(['query' => $query);
$dataProvider->pagination->pageSize = 30;
//    Now $dataProvider->totalCount is set because of BaseDataProvider::getPagination()
$query->andFilterWhere(['number' => $this->number]);
//    From now on totalCount is NOT updated to consider this condition and any other conditions added before feeding it to a GridView and rendering.
```

### What is the expected result?
totalCount should be set only when DataProvider gets to be used to fetch and render data.
Up to version 12.0.15 DataProvider pagination could be invoked and set before setting all its query conditions and afterwards totalCount was calculated correctly as expected.

### What do you get instead?
DataProvider totalCount is wrong - total number of rows in the table!! - doesn't care about query conditions added after invoking $dataProvider->pagination.

### Additional info
Related issue: https://github.com/yiisoft/yii2/issues/16891

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16
| PHP version      | 7.*
| Operating system |Windows/Linux

samdark:
What is the AR storage you are using?

GHopperMSK:
Is here any reason you don't use config to set `pageSize` value?

I think we should block any changes in `query` or `pagination` after `ActiveDataProvider` was initialized. Otherwise we have to recalculate it parameters every time. Or introduce an events system, which to be able to catch such changes and adjust only needed parameters.

anditza:
> What is the AR storage you are using?

MySQL

> Is here any reason you don't use config to set `pageSize` value?

Supposing you refer to app config, the reason is `pageSize` is not fixed but different for different `DataProvider`s.

> I think we should block any changes in `query` or `pagination` after `ActiveDataProvider` was initialized. Otherwise we have to recalculate it parameters every time. Or introduce an events system, which to be able to catch such changes and adjust only needed parameters.

...or delay calculating and setting `totalCount` until it is really needed. Why not?

GHopperMSK:
> Supposing you refer to app config, the reason is `pageSize` is not fixed but different for different `DataProvider`s.

No proplem:
```
$dataProvider = new ActiveDataProvider([
     'query' => $query,
     'pagination' => [
            'pageSize' => $YOUR_DYNAMIC_VALUE_HERE,
     ],
);
```

[...]

PR #17095, opened 2019-02-01
fix pagination

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17093


samdark:
Seems to be outdated.

PR #17097, opened 2019-02-03
Updates pt-BR translation of structure-views

sdlins:
Updates and fixes several grammatical, didactic, conceptual and functional errors.

PS: As I have time I will fix other pages because there are really so many errors. However, we know that it was done voluntarily by well minded persons.


samdark:
Merged. Thanks!

PR #17110, opened 2019-02-06
fix mongo greedy load on hasMany

lotos2512:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    |no
| Tests pass?   | yes
| Fixed issues  | fix mongo greedy load on hasMany

fix mongo greedy download on hasMany if relation one-to-many in mongo 

like 
```php
        class A {
              public function getNews()
             {
                 return $this->hasMany(B::class, ['firms' => '_id']);
             }
        }

        class B {
        }
      $result = (new A())->with(['news'])->all();
```
object a 

```json
{
   "_id" :  ObjectId("582596079a794727a6aa734t")
}
```
object b 
```json
{
   "_id" :  ObjectId("582596079a794727a6aa734f")
    "firms" : [ 
        ObjectId("582596079a794727a6aa734f"), 
        ObjectId("582596079a794727a6aa734t"), 
        ObjectId("582595eb9a794727a6aa70a1"), 
        ObjectId("582596349a794727a6aa76b4"), 
        ObjectId("582596d39a794727a6aa8125")
    ],
}
```

### DIFF
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -434,8 +434,16 @@ trait ActiveRelationTrait
         }
 
         if ($checkMultiple && !$this->multiple) {
-            foreach ($buckets as $i => $bucket) {
-                $buckets[$i] = reset($bucket);
+            $countKeys = count($linkKeys);
+            foreach ($models as $model) {
+                if($countKeys == 1 && is_array($model->{$linkKeys[0]})) {
+                    foreach ($model->{$linkKeys[0]} as $key) {
+                        $buckets[(string) $key][] = $model;
+                    }
+                } else {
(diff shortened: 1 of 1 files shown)


lotos2512:
@samdark may be method buildBuckets revrite as protected and fix it in  ActiveQuerymongo in mongodb extension

terabytesoftw:
Closing this PR as stale and not suitable for merge.

The reported MongoDB eager-loading use case may still be worth investigating, but this implementation does not fix the described `hasMany` case: the changed code only runs when the relation is not multiple. It also changes the existing `hasOne` bucketing behavior by no longer reducing buckets to a single related model.

PR #17117, opened 2019-02-08
Fixes #17089: fixed caching of related records when via() using with callable

### LINKED ISSUE #17089: Callable in model relation via junction table stop working on 2.0.16

Define two relation in some db active record model via junction table with callable:
```php
class Foo extends Article
{
    const
        FILE_RELATION_MAIN_IMAGE = 0,
        FILE_RELATION_LOGO = 1;

    public function getMainImage()
    {
        return $this->hasOne(File::class, ['file_id' => 'file_id'])->via('articleToFiles', function (ActiveQuery $query) {
            return $query->where(['relation' => self::FILE_RELATION_MAIN_IMAGE]);
        });
    }

    public function getLogoImage()
    {
        return $this->hasOne(File::class, ['file_id' => 'file_id'])->via('articleToFiles', function (ActiveQuery $query) {
            return $query->where(['relation' => self::FILE_RELATION_LOGO]);
        });
    }
}

class Article extends ActiveRecord
{
    public function getArticleToFiles()
    {
        return $this->hasMany(ArticleToFile::class, ['article_id' => 'article_id'])->inverseOf('article');
    }
}

class ArticleToFile extends ActiveRecord {}

class File extends ActiveRecord {}
```

Database tables has:
`article`: have article_id 1
`file`: have file_id 1 and 2
`articleToFile`: have two records:
```
article_id	file_id		relation
1		1		0
1		2		1
```
Call `$foo->mainImage->file_id` return 1
Call `$foo->logoImage->file_id` return 2

### What do you get instead?

Both (and any next file relation call) return `1` (first result).
When reverse calls (logoImage, then mainImage) got double `2`.

Works after rollback to 2.0.15.1.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16
| PHP version      | 7.2.x
| Operating system | KDE Neon

rugabarbo:
Closes #17116 (alternative PR).

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17089

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -6,7 +6,7 @@ Yii Framework 2 Change Log
 
 - Bug #17094: Fixed response on 204 status. Now it is empty (GHopperMSK)
 - Bug #17098: Fixed message/extract when using message params returned from method calls (rugabarbo)
-
+- Bug #17089: Fixed caching of related records when `via()` using with callable (rugabarbo)
 
 2.0.16 January 30, 2019
 -----------------------
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -170,16 +170,20 @@ class ActiveQuery extends Query implements ActiveQueryInterface
             } elseif (is_array($this->via)) {
                 // via relation
                 /* @var $viaQuery ActiveQuery */
-                list($viaName, $viaQuery) = $this->via;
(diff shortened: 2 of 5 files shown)


samdark (approved):

samdark:
Merged. Thank you!

PR #17087, opened 2019-01-30
fix tablename hardcode

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17075

The tests passed success, except one, witch I already fixed in #17077

Issue #17075, opened 2019-01-26
ActiveRecord findOne cannot recognize Model alias when using joinWith together


razonyang:
### What steps will reproduce the problem?

```sql
CREATE TABLE `user` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `username` varchar(100) COLLATE utf8mb4_unicode_ci NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;

CREATE TABLE `user_profile` (
  `user_id` int(11) NOT NULL,
  `name` varchar(64) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `address` varchar(128) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT ''
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

```
class User extends ActiveRecord
{
    public static function find()
    {
        return parent::find()
            ->alias('u')
            ->joinWith('profile');
    }

    public function getProfile()
    {
        return $this->hasOne(UserProfile::className(), ['user_id' => 'id']);
    }
}

User::findOne(1);
```

### What is the expected result?

No error

### What do you get instead?

```sql
Exception 'yii\db\Exception' with message 'SQLSTATE[42S22]: Column not found: 1054 Unknown column 'user.id' in 'where clause'
The SQL being executed was: SELECT `u`.* FROM `user` `u` LEFT JOIN `user_profile` ON `u`.`id` = `user_profile`.`user_id` WHERE `user`.`id`=1'
```

But it works if I specify the primary key without alias:

```php
User::findOne(['id' => 1]);
```

And if with alias that same as find():

```
User::findOne(['u.id' => 1]);
```

Got:
```
Exception 'yii\base\InvalidArgumentException' with message 'Key "u.id" is not a column name and can not be used as a filter'
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15
| PHP version      |  7.2
| Operating system | Linux

rob006:
Can you share generated SQL and `getProfile()` relation definition?

razonyang:
@rob006 I've updated the issue:)

PR #17127, opened 2019-02-13
`yii\db\ActiveRecord::findOne()` now accepts table aliases

albertborsos:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  |no
| Breaks BC?    |no
| Tests pass?   | yes
| Fixed issues  | #17075, #17087, #16905

This bug exists since `2.0.15`. I saw that there is already a PR for this issue, but that PR did not work for me.


samdark:
Do you mean #16905?

albertborsos:
Yes, #16905

framework/db/ActiveRecord.php
```diff
@@ -199,34 +200,56 @@ protected static function findByCondition($condition)
...
      * @param array $condition condition to filter.
+     * @param array $aliases
      * @return array filtered condition.
      * @throws InvalidArgumentException in case array contains unsafe values.
+     * @throws InvalidConfigException
      * @since 2.0.15
      * @internal
      */
-    protected static function filterCondition(array $condition)
+    protected static function filterCondition(array $condition, array $aliases)
```
samdark: Changing signature this way would break backwards compatibility. How about `$aliases = []`?

framework/db/ActiveRecord.php
```diff
@@ -199,34 +200,56 @@ protected static function findByCondition($condition)
...
+
+    /**
+     * Valid column names are table column names or column names prefixed with table name or table alias
+     *
+     * @param Connection $db
+     * @param array $aliases
+     * @return array
+     * @throws InvalidConfigException
+     */
+    public static function validFilterColumnNames($db, array $aliases)
```
samdark: Can we make it protected?

samdark: The name could be `filterValidColumnNames`.

samdark (changes requested):

framework/db/ActiveRecord.php
albertborsos: of course

framework/db/ActiveRecord.php
albertborsos: I've  just made the changes. Thanks for the review!

samdark (approved):

framework/db/ActiveRecord.php
```diff
@@ -199,34 +200,56 @@ protected static function findByCondition($condition)
...
      * @param array $condition condition to filter.
+     * @param array $aliases
      * @return array filtered condition.
      * @throws InvalidArgumentException in case array contains unsafe values.
+     * @throws InvalidConfigException
      * @since 2.0.15
      * @internal
      */
-    protected static function filterCondition(array $condition)
+    protected static function filterCondition(array $condition, array $aliases = [])
```
rob006: This breaks BC since it changes method signature.

rob006: > This breaks BC since it changes method signature.

Hmm, method is marked as internal, so it is probably fine...

framework/db/ActiveRecord.php
```diff
@@ -199,34 +200,56 @@ protected static function findByCondition($condition)
...
+     * @param array $aliases
+     * @return array
+     * @throws InvalidConfigException
+     */
+    protected static function filterValidColumnNames($db, array $aliases)
+    {
         $columnNames = [];
         $tableName = static::tableName();
         $quotedTableName = $db->quoteTableName($tableName);
+        $tableAliases = array_keys(ArrayHelper::removeValue($aliases, $tableName, []));
```
rob006: Why does it work in this way? Why not `array_unique(array_keys($aliases))`?

albertborsos: If I remember well, then it is to prevent to check twice for the table name and for the alias, if these two are the same.

rob006: I don't get it. AFAIK `$aliases` could be array like `['t1' => 'table1', 't2' => 'table2']`. Assuming that we're in `Table1` model, this will ignore `t2` alias - why?

albertborsos: You are right, I messed this up. I've just refactored it, based on that aliases are also could be `['t1' => 'table1', 't2' => 'table1']`.

Also I skip the alias if it is the same as the table name eg.

framework/db/ActiveRecord.php
```diff
@@ -199,34 +200,58 @@ protected static function findByCondition($condition)
...
+     * @return array
+     * @throws InvalidConfigException
+     */
+    protected static function filterValidColumnNames($db, array $aliases)
+    {
         $columnNames = [];
         $tableName = static::tableName();
         $quotedTableName = $db->quoteTableName($tableName);
+        $tableAliases = array_keys($aliases);
+        unset($tableAliases[$tableName]);
```
rob006: This probably will not work - `$tableAliases` is now integer-indexed array, `$tableAliases[$tableName]` probably never exist.

Is there any reason we're not using regexp for this validation? Whitelist approach looks like a source of countless edge cases:
1. `$this->from` is not always associative, it could be `['table1', 'table2']` and then `array_keys($aliases)` will return array of integers. Does it make sense to support `0.id` as column name?
2. Right now every additional table name is also ignored, so for `['t1' => 'table1', 't2' => 'table2']` column `table2.id` will be reported as incorrect.

albertborsos: I completely rewrote the alias handling and I wrote tests for it too.

samdark:
@albertborsos doesn't seem to work with PHP 5.4...

albertborsos:
Previously it failed on 5.5 too.. I will fix it soon.

rob006 (approved):
Looks fine for me. But I would consider making these new methods private - they should not be used in descendants anyway.

samdark:
Merged. Thank you!

albertborsos:
You are welcome. Thanks for the suggestions.

Issue #17098, opened 2019-02-03
message/extract ignores Yii::t calls when the message is using params returned from method calls


fl0v:
The update to yii 2.0.16 contains commit [5208ef761ed0552d1a4144589b1b7dff4c6fe3e7](https://github.com/yiisoft/yii2/commit/5208ef761ed0552d1a4144589b1b7dff4c6fe3e7) which tries to fix  issue [#16828](https://github.com/yiisoft/yii2/issues/16828).

The code at row [559](https://github.com/yiisoft/yii2/commit/5208ef761ed0552d1a4144589b1b7dff4c6fe3e7?diff=unified#diff-c4a5086d29e4d4b9309010d8f4a58e01R559) creates a bug regarding messages using params passed as a result from another method call:
```php
Yii::t('app', 'Message with param from method call {val}', [
    'val' => \Yii::$app->formatter->asDecimal(23, 4),
]);
```
The above message and all others following in the same file are ignored.

### What steps will reproduce the problem?

Having the following code in your application:
```php
echo PHP_EOL, Yii::t('app', '1. Simple message');
echo PHP_EOL, Yii::t('app', '2. Message with simple param {val}', [
    'val' => 'today',
]);
echo PHP_EOL, Yii::t('app', '3. Message with param from function call {val}', [
    'val' => date('Y-m-d'),
]);

// the next call creates the bug:
echo PHP_EOL, Yii::t('app', '4. Message with param from method call {val}', [
    'val' => \Yii::$app->formatter->asDecimal(23, 4),
]);

// after the bug:
echo PHP_EOL, Yii::t('app', '5. Simple message');
echo PHP_EOL, Yii::t('app', '6. Message with simple param {val}', [
    'val' => 'today',
]);
echo PHP_EOL, Yii::t('app', '7. Message with param from function call {val}', [
    'val' => date('Y-m-d'),
]);
```

Run message/extract with default translator:
```bash
./yii message/extract --source-path=@app --except='/vendor' --languages='en-US' --message-path=@app/messages --mark-unused=1 --overwrite=1 --remove-unused=0 --sort=1 --translator=Yii::t
```

### What is the expected result?

In messages/en-us/app.php:
```php
return [
    '1. Simple message' => '',
    '2. Message with simple param {val}' => '',
    '3. Message with param from function call {val}' => '',
    '4. Message with param from method call {val}' => '',
    '5. Simple message' => '',
    '6. Message with simple param {val}' => '',
[...]

PR #17128, opened 2019-02-13
Additional test case for #17098 (comment 463324939)

rugabarbo:
Additional test case for https://github.com/yiisoft/yii2/issues/17098#issuecomment-463324939.


samdark:
It doesn't seem to add new case so doesn't worth having it.

samdark:
@rugabarbo thank you for checking!

Issue #12182, opened 2016-08-17
Message context for Yii::t()


Deele:
I feel that `Yii::t()` lacks ability to provide context of textual value and when translations start to count in couple hundreds, it starts to be quite hard to understand what was meant with translation.

If we look at popular translation systems like "Crowdin" ([example](https://crowdin.com/page/in-action)), they allow to provide message context to support translators.

Currently, only way to provide some helpful information using `Yii::t()` is to write something in "key" area, that in turn defies purpose of "source language", for example, if you write a message "Propose (button in issue tasklist table header)" until I change that translation to "Propose", that message still be visible.

I am guessing, that there is performance issue with long keys, and length limitations does not allow long context descriptions.

For backwards compatability, I would like to suggest 5th optional attribute to `Yii::t($category, $message, $params = [], $language = null, $context = null)` (but this is not very convenient) that would allow to provide some sort of context description.

Maybe, we could convert `Yii::t()` to allow passing in first attribute array of attributes, like `Yii::t(['category' => 'message.category', 'message' => 'message text', 'params' => [], 'language' => null, 'context' => null])` or even allow shorthand for `category` and `message` with numeric keys `Yii::t(['message.category', 'message text', 'params' => [], 'language' => null, 'context' => null])`.

With `\yii\i18n\DbMessageSource`, they would be stored in separate table `source_message__context`, where we would store source message ID and context description with 255 characters long string.

With `\yii\i18n\PhpMessageSource`, they would be stored in separate file, right next to current message files, in `{category}_context.php` format, where we would have associative array with source message category as a key, and array with context descriptions, as a value.

I have no experience with `\yii\i18n\GettextMessageSource`, but I see in their documentation, that they already have an option to provide context descriptions. But I would like to see some suggestions for this solution.

This feature would allow to display all of those context description together with message in interface, to help translator to translate messages and understand context of it.

samdark:
Getext context is what we know as a category, not an example of usage: https://www.gnu.org/software/gettext/manual/html_node/Contexts.html

klimov-paul:
I am against such feature.
Placing some descriptive information around any call of `Yii::t()` sounds like overkill to me.

Besides how we can control same pair of 'category' and 'message' will have same 'context' at all its invocations? For example:

```php
Yii::t('app', 'Some message', [], 'en', 'Some context');
// ...
Yii::t('app', 'Some message', [], 'en', 'Another context!!!');
```

[...]

PR #17137, opened 2019-02-17
Issue #12182 support context for translations `Yii::t`

FilipBenco:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | https://github.com/yiisoft/i18n/issues/4 (it was moved form yii2 before I was able to finish it.. :/ )

THIS IS DRAFTED PULL REQUEST. Please look at this, and tell mi four opinion, or what I should improve. I would be very happy, if this feature became part of Yii framework. Thanks!

Message command is able to extract message context from `Yii:t()` calls containing `/* */` comments. To allow message context extraction, you need to set `extractContext` paramter to `true`.

After integrating this into Yii2, I will then port this also to upcoming Yii3.

## Reasoning
Our applications run in multiple markets and thus have to support multiple languages. Currently it is quite impossible, to just use extract command and send generated files to our translators, because the have no knowledge, where and how is some message used in our application. Mainly if for one English sentence there might be multiple Slovak translations, depending on context and vice versa. 
Also, translators have no idea, in which format or what kind of values can have parameters, used in messages.
Much deeper discussion about this topic can be found here: [yiisoft/yii-core/issues/70](https://github.com/yiisoft/yii-core/issues/70)

## Goal & Inspirations
To be able to add some context to translation messages, that can be used also in external localisation management tools like [Crowdin](https://crowdin.com/).
Similar approach is used for example in wordpress: https://codex.wordpress.org/I18n_for_WordPress_Developers

## Features
- Extract context for messages and params
- Locate file, from which was context for given message extracted. In case, message is used in multiple files, all files with given context are located and context is extracted.

### Main Message context
Message description should be specified as comment `/* */`, be inside `Yii::t()` call and follow directly after message and before params.
`Yii:t('category', 'message' /* mesasge description */)`
`Yii:t('category', 'message with {param}' /* mesasge description */, ['param' => 'parameter'])`

### Message Parameter context
Parameter description should be specified as comment `/* */`, be inside `Yii::t()` call.
For named parameters, it should follow directly after parameter name (key).
`Yii:t('category', 'message with {param}', ['param' /* param description */ => $param])`
For unnamed parameters, it should be before parameter value:
`Yii:t('category', 'message with {0}', [/* param description */ $param])`

## How extracted context is saved
```php
<?php // views/site/index.php
echo Yii:t(
    'category',
    'message with {param} and {anotherParam}' /* message description */,
    ['param' /* parameter description */ => $value, 'anotherParam' => $anotherValue]
);
```
### PHPMessageSource
Result is saved normally into `category.php` message file. Context is saved in comments above extracted message. 
```php
<?php // messages/en_US/category.php
return [
/**
 * @file views/site/index.php
 * @description message description
 * @param {param} parameter description
 * @param {anotherParam}
 */
'message with {param} and {anotherParam}'  => ''
];
```
### POMessageSource
Result is saved in PO file. Context is saved in `#.` gettext paramters, which refers to extracted comments and `#:` which refers to message reference. For more info about PO file format: [PO Format manual](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html)
```gettext
#: views/site/index.php
#. message description
#. {param} parameter description
#. {anotherParam}
msgctxt "category"
msgid "message with {param} and {anotherParam}"
msgstr ""

```
### DBMessageSource
not yet implemented. I think, I will store it in custom table, containing context for each occurrence of some message, but I am open to discussion.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -7,7 +7,7 @@ Yii Framework 2 Change Log
 - Bug #17094: Fixed response on 204 status. Now it is empty (GHopperMSK)
 - Bug #17098: Fixed message/extract when using message params returned from method calls (rugabarbo)
 - Bug #17089: Fixed caching of related records when `via()` using with callable (rugabarbo)
-- Enh #12182: Extract context for translations (FilipBenco)
+- Enh #12182: Extract context for translations from `Yii::t` calls (FilipBenco)
 
 2.0.16 January 30, 2019
 -----------------------
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -7,6 +7,7 @@ Yii Framework 2 Change Log
 - Bug #17094: Fixed response on 204 status. Now it is empty (GHopperMSK)
 - Bug #17098: Fixed message/extract when using message params returned from method calls (rugabarbo)
 - Bug #17089: Fixed caching of related records when `via()` using with callable (rugabarbo)
+- Enh #12182: Extract context for translations (FilipBenco)
 
 2.0.16 January 30, 2019
 -----------------------
--- a/framework/console/controllers/MessageController.php
+++ b/framework/console/controllers/MessageController.php
@@ -635,7 +635,7 @@ EOD;
                                         if (isset($translatorToken[0])
                                             && $translatorToken[0] === T_COMMENT
                                         ) {
-                                            $message['context'] = trim(mb_substr($translatorToken[1], 3, -3));
+                                            $context['context'] = trim(mb_substr($translatorToken[1], 3, -3));
                                             break;
                                         }
                                     }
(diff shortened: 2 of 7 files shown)


rob006:
Why comments are placed after message/param? I would say that preferred syntax for comments would be:

```php
echo Yii:t(
    'category',
    /* message description */
    'message with {param} and {anotherParam}',
    [
        /* parameter description */
        'param'  => $value, 
        'anotherParam' => $anotherValue,
    ]
);
````

samdark:
@FilipBenco thanks for contributing. The issue was moved to i18n package of Yii 3. The pull request isn't going to be merged into Yii 2 but would serve as an example on how to solve the issue.

Issue #17103, opened 2019-02-05
Count for relation with via() fails in 2.0.16


ddinchev:
Some of my tests started failing after upgrading to Yii 2.0.16. Same work OK on 2.0.15.1. Case 2 is the problematic one.

### What steps will reproduce the problem?

```
class Vehicle extends ActiveRecord {
    public function getEntries(): ActiveQuery
    {
        return $this->hasMany(Entry::class, ['VehicleID' => 'VehicleID'])->inverseOf('vehicle');
    }

    public function getFuelingCount(): int
    {
        return $this->hasMany(Fueling::class, ['EntryID' => 'EntryID'])->via('entries')->count('FuelingID');
    }
}


// CASE 1 - OK
$vehicle = Vehicle::findOne(1);
// insert 3 Fueling::class related to $vehicle
$vehicle->fuelingsCount; // 3
// When debugging this, it seems to produce a RAW SQL like `SELECT COUNT(FuelingID) FROM Fueling WHERE EntryID IN (1,2,3)` 


// CASE 2 - NOT OK
$vehicle = Vehicle::findOne(1);
$vehicle->fuelingsCount; // 0
// insert 3 Fueling::class related to $vehicle
$vehicle->fuelingsCount; // 0 <- this would be fine in Yii2 2.0.15.1
// The RAW query is `SELECT COUNT(FuelingID) FROM Fueling WHERE 1=0`

```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16
| PHP version      | 7.2.24
| Operating system | Ubuntu 16.04
| MySQL      | 8.0

samdark:
Is it similar to https://github.com/yiisoft/yii2/issues/17089?

ddinchev:
Seems to be the same issue. I'll try to revert [this commit](https://github.com/yiisoft/yii2/commit/5ee7fabbf948f3a759f5cfd9f76c72eef17694c8) to confirm that this is the one causing the bug.

rugabarbo:
How **exactly** do you insert 3 `Fueling::class` related to `$vehicle` in your cases?

ddinchev:
One particular test is this.

```php
class VehicleTest extends Unit
{
    use Specify;

    /**
     * @var Vehicle
     */
    private $vehicle;

    /**
     * @var \tests\common\UnitTester;
     */
    protected $tester;

    public function _before()
    {
        $this->vehicle = $this->tester->have(Vehicle::class);
    }

[...]

PR #17143, opened 2019-02-19
Issue #17103: test case

rugabarbo:
Just a failing test case for #17103


rugabarbo:
See https://github.com/yiisoft/yii2/issues/17103#issuecomment-466469692.
Closed.

Issue #17133, opened 2019-02-15
Console controller optionAliases are not always displayed in help text


joorloohuis:
### What steps will reproduce the problem?

In a console controller, add an option that will render hyphenated, e.g. `numMessages`. Also add an alias for this option:
```
public function optionAliases()
{
    return array_merge([
        'n' => 'numMessages',
    ], parent::optionAliases());
}
```

### What is the expected result?

When rendering the help for the command, I expect the option to be rendered as
```
--num-messages, -n: text to go with option
```

### What do you get instead?

```
--num-messages: text to go with option
```

So  no option alias is rendered. It turns out that I need to specify the alias to the _rendered_ value of the option:
```
public function optionAliases()
{
    return array_merge([
        'n' => 'num-messages',
    ], parent::optionAliases());
}
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16
| PHP version      | 7.2.15-1+ubuntu18.04.1+deb.sury.org+1
| Operating system | Ubuntu 18.04.1

PR #17149, opened 2019-02-20
fix aliases rendering

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17133 

We can now use both notations in aliases definition.
```
public function optionAliases()
{
    return array_merge([
        'n' => 'numMessages',
    ], parent::optionAliases());
}
```

or

```
public function optionAliases()
{
    return array_merge([
        'n' => 'num-messages',
    ], parent::optionAliases());
}
```


GHopperMSK:
Should  add test for this case?

samdark:
Won't hurt, yes.

GHopperMSK:
I couldn't suppress console output by `ob_start()/ob_get_contents()` statements. That is why I introduced `FakeHelpControllerWithoutOutput` controller. Maybe it isn't an optimal solution.

samdark (approved):

samdark:
Merged. Thank you!

PR #17151, opened 2019-02-21
Fixes #17150: BaseInflector::camel2words() bug with ALLCAPS words

### LINKED ISSUE #17150: BaseInflector camel2words behavour changed in 2.0.16

echo Inflector::camel2words("LEAD");

It should echo "Lead"

### What do you get instead?

It echos "L E A D"

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16
| PHP version      | 7.1.25
| Operating system | Ubuntu

brandonkelly:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17150

### DIFF
--- a/framework/helpers/BaseInflector.php
+++ b/framework/helpers/BaseInflector.php
@@ -373,7 +373,7 @@ class BaseInflector
             '-',
             '_',
             '.',
-        ], ' ', preg_replace('/(?<!\p{Lu})(\p{Lu})/u', ' \0', $name))), self::encoding());
+        ], ' ', preg_replace('/(?<!\p{Lu})(\p{Lu})|(\p{Lu})(?=\p{Ll})/u', ' \0', $name))), self::encoding());
 
         return $ucwords ? StringHelper::mb_ucwords($label, self::encoding()) : $label;
     }
--- a/framework/helpers/BaseInflector.php
+++ b/framework/helpers/BaseInflector.php
@@ -373,7 +373,7 @@ class BaseInflector
             '-',
             '_',
             '.',
-        ], ' ', preg_replace('/(\p{Lu})/u', ' \0', $name))), self::encoding());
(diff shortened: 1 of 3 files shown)


brandonkelly:
The Travis test failures were not introduced by this PR, fwiw.

framework/helpers/BaseInflector.php
```diff
@@ -373,7 +373,7 @@ public static function camel2words($name, $ucwords = true)
             '-',
             '_',
             '.',
-        ], ' ', preg_replace('/(\p{Lu})/u', ' \0', $name))), self::encoding());
+        ], ' ', preg_replace('/(?<!\p{Lu})(\p{Lu})|(\p{Lu})(?=\p{Ll})/u', ' \0', $name))), self::encoding());
```
cebe: what is the or branch of the regex for? This part: `(\p{Lu})(?=\p{Ll})`

Original regex (in 2.0.15 was `/(?<![A-Z])[A-Z]/`), making that unicode aware would be just: `/(?<!\p{Lu})\p{Lu}/`

brandonkelly: @cebe It looks for capital letters that are followed by a lowercase letter. Without it, `ITest` would be returned as `Itest` (as it did pre-2.0.16), but that would break this new test added in 2.0.16:

https://github.com/yiisoft/yii2/blob/ed64d6588630f9a1df92041fd5ca4087e786f3fb/tests/framework/helpers/InflectorTest.php#L108

(`Іце Дійсно Так!` would be returned instead of `І Це Дійсно Так!`.)

cebe: Got it, thanks.

SilverFire (approved):

cebe:
I added a few more tests to the test case.

cebe (approved):

samdark (approved):

samdark:
Merged. Thank you very much!

Issue #15216, opened 2017-11-24
[ENH] Add traceline to ErrorHandler


pana1990:
This one https://github.com/yiisoft/yii2-debug#open-files-in-ide but in the errorHandler
 view 
I want to click here(whatever line) and open file in my ide : 

![screenshot_1](https://user-images.githubusercontent.com/6630197/33221059-2ccaa3e4-d14d-11e7-8521-1d827b4e0a4b.png)

I can handle it :)

samdark:
Let's do it.

asamats:
Do somebody use this a function? I want to use. I love this function.
But, who tested it? I've just tried to use but I can't. First - We can't to click on a link. Second - The UI like "doesn't matter what I'll show".
Please, can we open the issue and fix it?

Thanks.

samdark:
@asamats did you configure it as described?

asamats:
@samdark Yes, I did. I configured my OS. From debug toolbar, I can open my IDE. From the error log, I also can but it's not easy because the link looks so strange (Chrome). Yes, if we look at the HTML code it looks correct.
![screenshot](https://user-images.githubusercontent.com/3476245/53109996-24ef6980-354b-11e9-9677-c40d57d6b82e.png)
1) Files should go one by one.
2) Links must be clickable.

PR #17152, opened 2019-02-21
Bug #15216: Fix a link for open files in IDE

asamats:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #15216


samdark:
@asamats would you please add a line for changelog?

asamats:
@samdark Yes, I can and must. Thanks.

framework/views/errorHandler/callStackItem.php
```diff
@@ -9,13 +9,28 @@
...
 /* @var $args array */
 /* @var $handler \yii\web\ErrorHandler */
+$html = <<<HTML
+IDE
+<svg class="icon icon--new-window" focusable="false" aria-hidden="true" width="16" height="16">
+    <use href="#new-window"></use>
+</svg>
+HTML;
 ?>
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" hidden>
```
samdark: What is this SVG for? I don't quite understand how does that work...

samdark (changes requested):

framework/views/errorHandler/callStackItem.php
asamats: I added an icon for an external link. It's just UI/UX thing.

It's "link" by id #new-window. I did this way to save a little bit of data/DOM size.
Info: https://www.w3.org/TR/SVG/linking.html

Excuse me, if I've not understood your question.

samdark:
Doesn't work well to me. The icon is there but there's unexpected spacing introduced in Chrome-based browsers:

c1

It's more visible if scrolling down a bit:

c2

framework/views/errorHandler/callStackItem.php
```diff
@@ -9,13 +9,28 @@
...
+        </g>
+    </symbol>
+</svg>
 <li class="<?= ($index === 1 || !$handler->isCoreFile($file)) ? 'application' : '' ?> call-stack-item"
     data-line="<?= (int) ($line - $begin) ?>">
     <div class="element-wrap">
         <div class="element">
             <span class="item-number"><?= (int) $index ?>.</span>
             <span class="text"><?= $file !== null ? 'in ' . $handler->htmlEncode($file) : '' ?></span>
+            <span> - </span>
```
alexantr: Need `&ndash;` like in `<span class="call">` below

asamats: That's not the problem. The problem is svg element. It has height, that is why we have this problem.
Problem solution:
1) svg must be invisible
2) svg must location higher

alexantr: Remove svg at all or use base64 png like links in the header

samdark: SVG is fine if it behaves well.

samdark (approved):

samdark:
Merged. Thanks!

PR #17168, opened 2019-02-26
Update to https protocol for php.net links

MysteryDragon:
Old links are currently unavailable (for me, at least).
However there are same content at http://php.net, and there are many links to http://php.net in PHPDoc (and only one link to https://secure.php.net, that's why I changed link to http://php.net).


rob006:
We should probably use https URLs.

MysteryDragon:
I can update URLs for all links (for consistency), or just for this one, if you want to.)
I also vote for https.

samdark:
Please update all links to https. Thanks.

MysteryDragon:
@samdark @rob006 well, I have done some changes for **php.net**.
I broke them into commits for more comfortable review. (Pay attention on https://github.com/yiisoft/yii2/pull/17168/commits/350de2bc7162724269c99346493971ed1feeb4b3, https://github.com/yiisoft/yii2/pull/17168/commits/0254ed0dda429eeb66b45f5a36d07e487ecb0e6f, https://github.com/yiisoft/yii2/pull/17168/commits/35ee28478449dcd10e1c44d6fcb120405492a952 .)

I can continue for other links which I will manage to find, or we can say that it's enough for one PR.)

samdark:
Merged. Thank you!

rob006:
https://secure.php.net is going to be deprecated and removed, I would suggest to use https://php.net instead: https://externals.io/message/104533

MysteryDragon:
Hmmm, interesting news... Well, I could prepare PR for that, it's easy to change again. It's logical, despite the fact that for now https://php.net redirects to https://secure.php.net

rob006:
Now we could wait till this change actually happen. There is no big difference in changing URLs today or in two months, and plans like to change, maybe it will never happen and URLs should stay as is. :)

PR #17175, opened 2019-03-01
Introduced foreign values for fixture data

mklemarczyk:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no

I suggest to introduce format into fixture data:
```['ActiveRecordModelName', 'ForeignModelDataAlias', 'ModelProperty']```

It will be beneficial to get foreign data from dependent fixtures, especially the generated data like id, updated_at etc.

Example of use:
```['app\models\User', 'disabled_user', 'id']```

It was made to work, so feedback is welcome.

Unfortunately it is not possible to implement such feature without changes in Yii2 itself.

### DIFF
--- a/framework/test/ActiveFixture.php
+++ b/framework/test/ActiveFixture.php
@@ -76,6 +76,19 @@ class ActiveFixture extends BaseActiveFixture
         $this->data = [];
         $table = $this->getTableSchema();
         foreach ($this->getData() as $alias => $row) {
+            foreach ($row as $key => $value) {
+                if (is_array($value)) {
+                    list($modelClass, $modelAlias, $property) = $value;
+                    if (!key_exists($modelClass, $this->dependInstances)) {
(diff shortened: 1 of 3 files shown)


samdark:
@mklemarczyk sorry but enhancements aren't accepted into 2.0 branch anymore.

samdark:
Thank you for contributing anyway.

mklemarczyk:
Hi @samdark,
Will it be possible to add this change into 2.1?
Fixtures are missing right now 2 importants things in my opinion.
1) Related fixtures (like suggested change) to be able link fixture data,
2) Filling data in junction table, model can have viaTable relation, but there is no way to put data into it using Fixtures without creating model only for Fixture use.

Best regards,
Maciek

samdark:
Could be added to 3.0.

Issue #15037, opened 2017-10-28
Session Write Failed


Knight-Yoshi:
### What steps will reproduce the problem?
Couldn't say this was the first time it happened and it happened only once on page reload - reloading a second time had no issue.

### What is the expected result?
Session writes to the DB

### What do you get instead?
Session write failed

### Additional info
[...]

Issue #13740, opened 2017-03-09
session_write_close(): Failed to write session data (user). Please verify that the current setting of session.save_path is correct (/var/lib/php/session)


miradnan:
### What steps will reproduce the problem?
I had 2.0.8 and I upgraded to 2.0.11
### What is the expected result?
It should allow me to browse the app instead it started throwing error
### What do you get instead?
session_write_close(): Failed to write session data (user). Please verify that the current setting of session.save_path is correct (/var/lib/php/session)

<img width="1418" alt="screen shot 2017-03-09 at 9 02 40 pm" src="https://cloud.githubusercontent.com/assets/397415/23757239/3b7fc274-050c-11e7-9627-c08d51dfdb63.png">

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11
| PHP version      | 7.0.15
| Operating system | Ubuntu 14

samdark:
Do you have permissions allowing your webserver user to access `/var/lib/php/session`?

miradnan:
Yes. I gave it 777 even though it doesn't work. I have 2.0.8 working on the same machine and PHP config without any issues.

samdark:
Well, "failed to write session data" means exactly what it's saying i.e. file could not be written by the user running PHP process. 2.0.8 was ignoring the fact and continue running with empty session. Current version throws this error to warn developer that it's not OK and session, in fact, isn't written.

You should check for permissions again and make sure that the user (usually `www-data`) can actually write to that directory.

miradnan:
Alex,

I checked the folder permissions before I posted this as an issue. 
My sessions are stored in MySQL database. sessions doesn't even have to write to the folder in the first place. They should directly read and write in MySQL. I can share my session config if needed pls let me know 


Thanks,
Mir Adnan

samdark:
Hmm...

> My sessions are stored in MySQL database.

But

> session_write_close(): Failed to write session data (user). Please verify that the current setting of session.save_path is correct (/var/lib/php/session)

That doesn't seem to make sense. Do you have right config file loaded?

miradnan:
Yes. The same config was working absolutely fine with 2.0.8. I upgraded and this issue came up.

Here is my sessions config.
<img width="361" alt="screen shot 2017-03-10 at 9 24 33 am" src="https://cloud.githubusercontent.com/assets/397415/23781450/684d5854-0573-11e7-9945-229af99731b2.png">

File common\db\DbSession contains
[...]

PR #17188, opened 2019-03-05
Handle DB session callback custom fields before session closed

### LINKED ISSUE #9438: error saving sessions
hi all, im using the new version of the yii2 2.0.6 but i have this problem

i get this Error

[16-Aug-2015 19:18:46 UTC] PHP Fatal error:  Uncaught exception 'yii\base\ErrorException' with message 'Unknown: Failed to write session data (user). Please verify that the current setting of session.save_path is correct (/tmp)' in Unknown:0
Stack trace:
#0 [internal function]: yii\base\ErrorHandler->handleError(2, 'Unknown: Failed...', 'Unknown', 0, Array)
#1 {main}

  thrown in Unknown on line 0

and is because i use this configuration

``` php
'session' => [
            'class' => 'yii\web\DbSession',
            'readCallback' => function ($fields) {
                    return [
                        'expireDate' => Yii::$app->formatter->asDate($fields['expire']),
                    ];
                },
            'writeCallback' => function ($session) {
                    return [
                        'user_id' => Yii::$app->user->id,
                        'ip' => $_SERVER['REMOTE_ADDR'],
                        'is_trusted' => $session->get('is_trusted', false),
                        'controllerCurrent' => Yii::$app->requestedAction->controller->id,
                        'actionCurrent' => Yii::$app->requestedAction->id,
                        'pageCurrent' => Yii::$app->id,
                        'is_guest' => Yii::$app->user->isGuest
                    ];
                }
        ]
```

in the common.php

in the localhost work Perfectly but in the remote i get the error

any have idea how can solve this ?

lubosdz:
Attempt to fix #9438, #13740, #15037.
This fixes error `Unkown: Failed to write session data (user)`.
Explanation of cause [here](https://github.com/yiisoft/yii2/issues/9438#issuecomment-469625968).

Please review - perhaps somebody might get better idea how to populate fields before closing the session without corrupting the data integrity .. (e.g. if a session variable stored into session field changed between being populated and closing the session). Personally I am not happy with this commit, but so far haven't found better solution :-|


samdark:
Doesn't seem to pass tests: https://travis-ci.org/yiisoft/yii2/jobs/501995713

lubosdz:
This is really hard to fix without BC break. 
The implementation is faulty since the very beginning.

Method `MultiFieldSession::composeFields()` attempts to do two things at the same time - populate custom fields in custom session handler after session is closed (inactive). This results into permanent errors when somebody tries to read some variable from session into custom field, typically `user ID`.

I tried 3 attempts to move preparation of custom fields, but there is always some unit test failing. I could update unit tests but it might break backward compatibility if somebody used this *feature* to set custom session content. 

Any idea how to fix ? - either introduce minor BC break or leave faulty implementation ? :-)

Also - it there guide how to run locally all framework tests? Before pushing the change I would like to run specific tests locally.

samdark:
The tests could be executed with `./vendor/bin/phpunit`. As the argument you may pass the name of the test class.

samdark:
> if somebody used this feature to set custom session content

That is possible.

lubosdz:
Fail on codeclimate `writeSession accesses the super-global variable $_SESSION` - it's used in many methods. 

Final fix works, but IMO modifying session data directly via callback should not be allowed, just like it's not possible for FileSession. Users mainly use custom writeCallback fields for tracking who's online or IDs-related stuff etc. It's kinda undocumented dirty feature specific only to DB session :-)

samdark:
We may consider disallowing it for 3.0 (sessions will be handled differently there anyway).

framework/web/DbSession.php
```diff
@@ -76,6 +76,10 @@ class DbSession extends MultiFieldSession
      */
     public $sessionTable = '{{%session}}';
 
+    /**
+    * @var array Session fields to be written into session table columns
```
samdark: Need `@since 2.0.17`.

framework/web/DbSession.php
```diff
@@ -136,6 +140,18 @@ public function regenerateID($deleteOldSession = false)
         }
     }
 
+    /**
+     * Ends the current session and store session data.
+     */
+    public function close()
```
samdark: Need `@since 2.0.17`.

framework/web/DbSession.php
```diff
@@ -169,14 +185,28 @@ public function writeSession($id, $data)
...
         try {
-            $fields = $this->composeFields($id, $data);
-            $fields = $this->typecastFields($fields);
-            $this->db->createCommand()->upsert($this->sessionTable, $fields)->execute();
+            // ensure backwards compatability (fixed #9438)
+            if($this->writeCallback && !$this->fields){
+                $this->fields = $this->composeFields();
+            }
+            // ensure data consistency
+            if(!isset($this->fields['data'])){
```
samdark: Code formatting is wrong.

1. Need spaces `if (`
2. Need spaces `) {`

samdark: 3. Need spaces `} else {`

samdark (changes requested):
Looks good except minor things.

framework/web/DbSession.php
lubosdz: Done

framework/web/MultiFieldSession.php
```diff
@@ -89,30 +89,19 @@ public function getUseCustomStorage()
...
-            $fields = array_merge(
-                $fields,
-                call_user_func($this->writeCallback, $this)
-            );
-            if (!is_string($fields['data'])) {
-                $_SESSION = $fields['data'];
-                $fields['data'] = session_encode();
-            }
+        $fields = $this->writeCallback ? call_user_func($this->writeCallback, $this) : [];
+        if($id !== null){
```
samdark: Formatting here is wrong as well.

samdark (commented):
Is it possible to add unit tests for it?

framework/web/MultiFieldSession.php
lubosdz: Sorry, missed that one, now it's fixed.

lubosdz:
Unit tests for sessions are already there:
https://github.com/yiisoft/yii2/blob/master/tests/framework/web/session/AbstractDbSessionTest.php#L136
There is no new feature to test.
Or what should be added?

samdark:
A test that failed before these changes and isn't failing with these changes.

lubosdz:
There was no failing test before this fix.
Travis was failing, because my fix was not compatible with original implementation.
Now the fix is backwards compatible so no failing tests now - just like before the fix.

lubosdz:
OK, I think it would make sense to add test for new method `close()` with user ID pulled out from session - to ensure custom fields do not interfere with inactive session anymore.

lubosdz:
Whooops .. I did not intent to merge anything, 
I just clicked Updated branch button .. and it merged. 
Please check if it's OK or squash commits, whatever needed .. 
sorry, I am not really familiar with github administration, I use gitlab or bitbucket.

samdark:
It's fine to merge master into pull request branch. No issues with it.

framework/web/DbSession.php
```diff
@@ -169,14 +187,28 @@ public function writeSession($id, $data)
...
         // https://secure.php.net/manual/en/function.session-set-save-handler.php#refsect1-function.session-set-save-handler-notes
         try {
-            $fields = $this->composeFields($id, $data);
-            $fields = $this->typecastFields($fields);
-            $this->db->createCommand()->upsert($this->sessionTable, $fields)->execute();
+            // ensure backwards compatability (fixed #9438)
+            if ($this->writeCallback && !$this->fields) {
+                $this->fields = $this->composeFields();
+            }
+            // ensure data consistency
```
samdark: Would you please elaborate what happens in this block?

tests/framework/web/session/AbstractDbSessionTest.php
```diff
@@ -147,6 +148,35 @@ public function testWriteCustomField()
...
 
+    /**
+     * @depends testReadWrite
+     */
+    public function testWriteCustomFieldWithUserId()
+    {
+        $session = new DbSession();
+        $session->open();
+
+        $savedUserId = mt_rand(111, 999);
```
samdark: Is it necessary for it to be random?

framework/web/DbSession.php
lubosdz: This is to ensure backwards compatibility & that unit tests won't fail.
Basically it forces populating custom fields on directly calling custom session handler `$session->writeSession()`.

Normally, custom fields are populated [just before closing the session](https://github.com/yiisoft/yii2/pull/17188/files#diff-6f6a191d70c6d117086ec1873e231af5R152).

But original unit tests would not call dedicated method `$session->close()` but instead directly custom handler method `$session->writeSession(sessId, sessDataString)` (because the session has not been opened), which IMO is not quite correct. 

I did not want to touch original unit tests, therefore this line must have been added as a work around solution. If direct injection of session data will drop in Yii3 these lines may be removed, or unit test should be rewritten to open/close session rather than directly calling custom handlers.

tests/framework/web/session/AbstractDbSessionTest.php
lubosdz: No, it just did development for me slightly easier to check that test locally passed during debug. Shall I make it static, e.g. 123 ?

samdark: Keeping things simple is better. Yes.

lubosdz: OK, done.

samdark (approved):

sergeymakinen (approved):
LGTM. Hacky but should work.

samdark:
Merged. Thank you!

Issue #17185, opened 2019-03-04
AssetManager appendTimestamp to manually published files


crisps-for-breakfast:
I'm attempting to use the AssetManager to manually publish some static files (neither CSS or JS), but when I call `AssetManager::publish()` it does not append the timestamp. I have set `$appendTimestamp = true`

Semantically, I would expect a function called publish, which returns a URL, to give a final URL with all amendments

Furthermore, both the parameter name used to append the timestamp, 'v', and the code itself, are hardcoded deep in the `AssetManager::getAssetUrl()`, so it's not extensible

It appears that lots of people are hitting this issue. Could you at the least open this up for extension by separating the code into a protected or public function, like `AssetManager::appendTimestamp($url)`

 It would also be nice to have a `public $appendTimestampParam = 'v'` so that this can be configured too

*I love your framework, but I often run into situations where code is not separated into functions or variables. Extending the framework often requires copying copying whole chunks of code into overridden functions, or just replacing files wholesale*

samdark:
How do you call `AssetManager::publish()` exactly?

crisps-for-breakfast:
Here is my makeshift function that publishes a file using the asset manager

    /**
     * Publish file using asset manager and return URL
     *
     * @param string $filename File to publish
     * @return string URL of published file
     *
     * @throws InvalidConfigException
     */
    public function publish($filename) {

        $assetManager = Yii::$app->assetManager;
        $sourceFilename = Yii::getAlias($filename);

        // Publish file, I was hoping this would append the timestamp
        list($publishedFilename,$url) = Yii::$app->assetManager->publish($sourceFilename);

        // Append timestamp if necessary, code from AssetManager::getAssetUrl()
        if (
            $assetManager->appendTimestamp &&
            ($timestamp = @filemtime($publishedFilename)) > 0
        ) {
            $url .= '?v=' . $timestamp;
        }

        return $url;
    }

Here is my relevant AssetManager config:

  from common

        'assetManager' => [
            // Cache bust newer assets
            'appendTimestamp' => true,
        ],

  from frontend

        'assetManager' => [
            'basePath' => '@frontend-assets-dir', // frontend/web/assets
            'baseUrl' => '@frontend-assets-url', //  example.com/assets
        ],

samdark:
@crisps-for-breakfast try code from master. Was your issue fixed?

crisps-for-breakfast:
That got it! Thank you so much :)

PR #17189, opened 2019-03-05
fix timestamp

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17185

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.17 under development
 ------------------------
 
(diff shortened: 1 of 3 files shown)


s1lver:
@samdark @GHopperMSK 
```php
@filemtime($dstFile)
```
I don't think that's good. It's better to get an error than to guess what goes wrong

samdark:
@s1lver yes but it's in the same way that other timestamps added in the same class.

s1lver:
really,but don't keep doing that

samdark:
Won't do in Yii 3.

PR #17196, opened 2019-03-07
Integrate some section with example or suggestion

minitia:
For me, it wasn't easy to understand where register event handlers

### DIFF
--- a/docs/guide/concept-events.md
+++ b/docs/guide/concept-events.md
@@ -73,6 +73,7 @@ public function init(){
   parent::init(); // DON'T Forget to call the parent method.
 }
 ```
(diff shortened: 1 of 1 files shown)


minitia:
I was wrong with the link of advanced template, and I like to add a reference to the stackoverflow page where I found the example:
https://stackoverflow.com/questions/28575636/how-to-use-events-in-yii2

samdark:
Thanks for contribution but these changes won't be accepted:

1. Events section is general purpose. Events could be attached to anything so model example looks odd.
2. Events could be registered anywhere so I'm not sure why it's important to emphasize on bootstrap specifically.

machour:
I think that this is a signal that events documentation is not clear enough for new comers.

samdark:
Likely. @minitia I think for your examples the wiki fits quite well: https://www.yiiframework.com/wiki. Would you mind creating an article there?

minitia:
I've done here
https://www.yiiframework.com/wiki/2548/events-registration-examples
thank you guys

Issue #17174, opened 2019-02-28
unlink and link should also use the via on conditions of the relation


Bhoft:
Hi 
I have a ManyMany relation between 2 Tables e.g. ReviewerInvites and Submissions.
I have an allocation of "multiple" reviewer Types e.g. individual review, consensus review.

Therefore i have an additional column "reviewer_type" in the junction Table.
and defined constants like :

```php
    REVIEW_TYPE_INDIVIDUAL = 1
    REVIEW_TYPE_CONSENSUS = 2
```

The Junction Table has the following columns, 
submission_id, reviewer_invite_id, reviewer_type 

I could now have the relation in my Submission Model like this:

```php
    public function getConsensusReviewer()
    {
        return $this->hasMany(ReviewerAllocation::className(), ['submission_id' => 'id'])
            ->andOnCondition(['reviewer_type' => self::REVIEW_TYPE_CONSENSUS]);
    }

    public function function getConsensusReviewers()
    {
        return $this->hasMany(Reviewer::className(), ['id' => 'reviewer_id'])
             ->via('consensusReviewer');
    }
```

the same for the individual reviewers

```php
    public function getIndividualInvites()
    {
        return $this->hasMany(ReviewerInvite::className(), ['id' => 'reviewer_invite_id'])
            ->via('individualReviewer');
    }
    public function getIndividualReviewer()
    {
        return $this->hasMany(ReviewerAllocation::className(), ['submission_id' => 'id'])
            ->andOnCondition(['reviewer_type' => self::REVIEW_TYPE_INVIVIDUAL]);
    }
```

Everything works except "link()" and "unlink()".

For "link()" i could use the "extraColumns" attribute to set the reviewer_type.
But when "unlink()" is used the On condition isn't used.

The SQL always deletes me entries for both reviewer_types:
e.g.

```sql
    DELETE FROM `submission_reviewer_invite` WHERE (`submission_id`=70) AND (`reviewer_invite_id`=47) 
```

instead of:

```sql
    DELETE FROM `submission_reviewer_invite` WHERE (`submission_id`=70) AND (`reviewer_invite_id`=47) AND (`reviewer_type`=2)
```

I also tried to use viaTable instead of via and set the onCondition directly.

```php
    public function getIndividualInvites()
    {
        return $this->hasMany(ReviewerInvite::className(), ['id' => 'reviewer_invite_id'])
	  		->viaTable('submission_reviewer_invite', ['submission_id' => 'id'], function ($query) {
[...]

PR #17197, opened 2019-03-08
update BaseActiveRecord

Bhoft:
Update BaseActiveRecord so that link() and unlink() supports viaRelation on conditions.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #17174

### DIFF
--- a/framework/db/BaseActiveRecord.php
+++ b/framework/db/BaseActiveRecord.php
@@ -1319,7 +1319,7 @@ abstract class BaseActiveRecord extends Model implements ActiveRecordInterface
             }
             
             // add viaRelation On conditions to preset columns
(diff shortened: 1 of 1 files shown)


Bhoft:
I have realized that the On Condition of the ActiveQuery could also be a String. 
My solution only handles the conditions by array. 
I am not sure how a "string" condition could be split up into parts to fill the $columns array.
Maybe its enough to throw an error if its an string. 
So at least the condition via arrays are handled correctly.

bizley:
Closed in favor of #18564

PR #17206, opened 2019-03-14
Update concept-di-container.md

yidas:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

For the first example code for DI guide, I think it's better to use Yii container to define `$container`  for practical usage.

### DIFF
--- a/docs/guide/concept-di-container.md
+++ b/docs/guide/concept-di-container.md
@@ -33,7 +33,7 @@ class Foo
     }
 }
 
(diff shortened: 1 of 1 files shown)


samdark:
While it works, it's not a good practice to reference container using global `Yii::` shortcut...

Issue #17212, opened 2019-03-19
ActiveRecord orderBy params(array)  is invalid


huaiguoguo:
### What steps will reproduce the problem?
![image](https://user-images.githubusercontent.com/1888094/54577338-536f4000-4a36-11e9-94b6-8e4a937ddfc0.png)
![image](https://user-images.githubusercontent.com/1888094/54577344-58cc8a80-4a36-11e9-9d52-10e308caf3d6.png)

![image](https://user-images.githubusercontent.com/1888094/54577408-ae089c00-4a36-11e9-8b44-9ca0bb4b11d8.png)
![image](https://user-images.githubusercontent.com/1888094/54577413-b82a9a80-4a36-11e9-940f-8fd5c5b5476b.png)

ActiveRecord -> orderBy(['created'=>'desc'])  is   invalid
ActiveRecord -> orderBy('created desc')  is   valid

### What is the expected result?
 i hope ActiveRecord -> orderBy(['created'=>'desc'])    will valid

![image](https://user-images.githubusercontent.com/1888094/54577413-b82a9a80-4a36-11e9-940f-8fd5c5b5476b.png)

### What do you get instead?
![image](https://user-images.githubusercontent.com/1888094/54577338-536f4000-4a36-11e9-94b6-8e4a937ddfc0.png)
![image](https://user-images.githubusercontent.com/1888094/54577344-58cc8a80-4a36-11e9-9d52-10e308caf3d6.png)

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16.1
| PHP version      |  7.3.3
| Operating system | win10 64

bizley:
While it's stated in [Guide](https://www.yiiframework.com/doc/guide/2.0/en/db-query-builder#order-by) that for array parameter `SORT_DESC` and `SORT_ASC` should be used it does not state that it's only valid option. Maybe we should mention that or allow "desc" and "asc" strings for array.

huaiguoguo:
> While it's stated in [Guide](https://www.yiiframework.com/doc/guide/2.0/en/db-query-builder#order-by) that for array parameter `SORT_DESC` and `SORT_ASC` should be used it does not state that it's only valid option. Maybe we should mention that or allow "desc" and "asc" strings for array.

i am try this ,  but is invalid too!

huaiguoguo:
[...]

PR #17216, opened 2019-03-20
Issue 17212 fix

setnemo:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17212

### DIFF
--- a/framework/db/QueryTrait.php
+++ b/framework/db/QueryTrait.php
@@ -307,7 +307,8 @@ trait QueryTrait
      * Sets the ORDER BY part of the query.
      * @param string|array|ExpressionInterface $columns the columns (and the directions) to be ordered by.
      * Columns can be specified in either a string (e.g. `"id ASC, name DESC"`) or an array
-     * (e.g. `['id' => SORT_ASC, 'name' => SORT_DESC]`).
+     * (e.g. `['id' => SORT_ASC, 'name' => SORT_DESC]` or `['id' => 'asc', 'name' => 'DESC']` ). Be careful,
+     * such a call `['asc']` creates a query that will sort in ascending order column by name '0'
      *
      * The method will automatically quote the column names unless a column contains some parenthesis
      * (which means the column contains a DB expression).
--- a/framework/db/QueryTrait.php
+++ b/framework/db/QueryTrait.php
@@ -368,7 +368,7 @@ trait QueryTrait
             return [$columns];
         } elseif (is_array($columns)) {
             foreach ($columns as $column => $value) {
-                if (preg_match('#(asc|desc)#i', $value, $matches)) {
+                if (preg_match('#\b(asc|desc)\b#i', $value, $matches)) {
                     $result[$column] = strcasecmp($matches[0], 'desc') ? SORT_ASC : SORT_DESC;
                 } else {
                     $result[$column] = $value == SORT_DESC ? $value : SORT_ASC;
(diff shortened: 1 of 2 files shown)


framework/db/QueryTrait.php
```diff
@@ -363,14 +363,23 @@ public function addOrderBy($columns)
...
     {
+        $result = [];
         if ($columns instanceof ExpressionInterface) {
             return [$columns];
         } elseif (is_array($columns)) {
-            return $columns;
+            foreach ($columns as $column => $value) {
+                if (is_integer($value)) {
+                    $result[$column] = $value;
+                } elseif (preg_match('#(asc|desc)#i', $value, $matches)) {
```
samdark: How about `ORDER BY description` or `ORDER BY ascription`?

framework/db/QueryTrait.php
```diff
@@ -363,14 +363,21 @@ public function addOrderBy($columns)
      */
     protected function normalizeOrderBy($columns)
     {
+        $result = [];
         if ($columns instanceof ExpressionInterface) {
             return [$columns];
         } elseif (is_array($columns)) {
-            return $columns;
+            foreach ($columns as $column => $value) {
+                if (preg_match('#(asc|desc)#i', $value, $matches)) {
```
samdark: How about ORDER BY description or ORDER BY ascription?

samdark: It could search for `" asc"` and `" desc"`. In this case non-regex search could be a bit faster. Or it could use word boundaries i.e. `#\b(asc|desc)\b#i`.

bizley: This is partial match as well (like in "aaaascbbb") - is it intended? Otherwise `#^(asc|desc)$#i`.

omentes: Oh, your right. Maybe add  `^...$` ?
`#^\b(asc|desc)\b$#i`

omentes: Without ^ $ like a trim. I will add your solution. Thanks!

omentes: change to `#\b(asc|desc)\b#i`
Work with space (typo) before and after

omentes:
@samdark What do you think about this?

rob006:
What if my column is named `asc` or `desc`?

omentes:
> What if my column is named `asc` or `desc`?

Everything will be OK, because the code checks value.
For example
`->orderBy(['asc' => 'desc'])`
where 'asc' is key, 'desc' - value

rob006:
What if I do `->orderBy(['desc'])`?

omentes:
> What if I do `->orderBy(['desc'])`?

Great question! The query will sort the column "0" in descending order. I'm not sure that I should handle this behavior. It is necessary to add information about the new functionality to the documentation.

omentes:
> What if I do `->orderBy(['desc'])`?

If I add a key check to a number, it will be impossible to work with columns that have the number-name

rob006:
Personally I'd like to avoid introducing "yet another way of defining sorting" - it does not give any additional value and it is less reliable (it is easy to make a typo - `->orderBy(['id' => 'decs'])`). Especially if it breaks BC and does not play well with edge cases.

omentes:
> Personally I'd like to avoid introducing "yet another way of defining sorting" - it does not give any additional value and it is less reliable (it is easy to make a typo - `->orderBy(['id' => 'decs'])`). Especially if it breaks BC and does not play well with edge cases.

I think using arguments in an associative array with strings (as in other methods) is a logical addition. I was surprised that orderBy() works differently.

samdark:
Thinking more about it, it cannot be considered a bug and enhancements aren't accepted into 2.x. For 3.0 it could be OK to have.

samdark:
Thanks for contribution and sorry that I'm closing the issue.

Issue #17223, opened 2019-03-24
Behavior event doesn't detach


laxity7:
Behavior event is not detach if handler is Closure.

Simple example:

```php
$obj = new class () extends Component
{
    public $foo = 'foo';
};
$obj->attachBehavior('bar', (new class () extends Behavior
{
    public function events()
    {
        return [
            'barEventOnce' => function ($event) {
                echo $this->owner->foo;
                $this->detach(); // like that I intended to do
            },
        ];
    }
}));

$obj->trigger('barEventOnce'); // it's ok, see "foo"
$obj->trigger('barEventOnce'); // exception, because owner is null
```

Actually, I did the behavior for a model, the event in which was supposed to work only once. The handler of an event at me also anonymous function, but an event does not get rid and by a repeated call I receive a mistake.
It turns out that the behavior got rid, and the event does not, because of here such check in the class [Component](https://github.com/yiisoft/yii2/blob/master/framework/base/Component.php#L563)
```
if ($event[0] === $handler) {
```

I replaced Closure with a class function, but was unpleasantly surprised by this behavior. Maybe add this remark to the documentation or fix this error?

Yii version : 2.0.16
PHP version:  5.6/7.1

laxity7:
The problem is not only in behavior, but global, therefore best solution is to compare functions in Component in **off** function, like this:

```php
public function off($name, $handler = null)
{
    $this->ensureBehaviors();
    if (empty($this->_events[$name]) && empty($this->_eventWildcards[$name])) {
        return false;
    }
    if ($handler === null) {
        unset($this->_events[$name], $this->_eventWildcards[$name]);
        return true;
    }

    $removed = false;
    // just an example of a function to compare. Need to improve and move
    $isClosureEqual = function (\Closure $func1, \Closure $func2) {
        $refFunc = new \ReflectionFunction($func1);
        $sign1 = $refFunc->getFileName() . '|' . $refFunc->getStartLine();

        $refFunc = new \ReflectionFunction($func2);
        $sign2 = $refFunc->getFileName() . '|' . $refFunc->getStartLine();

        return $sign1 === $sign2;
    };

    // plain event names
    if (isset($this->_events[$name])) {
        foreach ($this->_events[$name] as $i => $event) {
            if ($handler instanceof \Closure) {
                $handlersEqual = $isClosureEqual($event[0], $handler);
            } else {
                $handlersEqual = $event[0] === $handler;
            }

            if ($handlersEqual) {
[...]

PR #17225, opened 2019-03-25
Fix event detaching when in is a Closure instance

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17223

### DIFF
--- a/framework/base/Behavior.php
+++ b/framework/base/Behavior.php
@@ -85,16 +85,16 @@ class Behavior extends BaseObject
     public function detach()
     {
         if ($this->owner) {
-            foreach ($this->events() as $event => $handler) {
-                if (is_string($handler)) {
-                    $sParam = [$this, $handler];
-                } elseif ($handler instanceof \Closure) {
-                    $sParam = null;
+            foreach ($this->events() as $event => $value) {
+                if (is_string($value)) {
+                    $handler = [$this, $value];
+                } elseif ($value instanceof \Closure) {
+                    $handler = null;
                 } else {
-                    $sParam = $handler;
+                    $handler = $value;
                 }
 
-                $this->owner->off($event, $sParam);
+                $this->owner->off($event, $handler);
             }
             $this->owner = null;
         }
--- a/framework/base/Behavior.php
+++ b/framework/base/Behavior.php
@@ -88,7 +88,7 @@ class Behavior extends BaseObject
             foreach ($this->events() as $event => $handler) {
                 if (is_string($handler)) {
                     $sParam = [$this, $handler];
-                } else if ($handler instanceof \Closure) {
+                } elseif ($handler instanceof \Closure) {
                     $sParam = null;
                 } else {
                     $sParam = $handler;
--- a/framework/base/Behavior.php
+++ b/framework/base/Behavior.php
@@ -86,7 +86,15 @@ class Behavior extends BaseObject
     {
         if ($this->owner) {
             foreach ($this->events() as $event => $handler) {
-                $this->owner->off($event, is_string($handler) ? [$this, $handler] : $handler);
+                if (is_string($handler)) {
+                    $sParam = [$this, $handler];
+                } else if ($handler instanceof \Closure) {
+                    $sParam = null;
+                } else {
+                    $sParam = $handler;
+                }
+
+                $this->owner->off($event, $sParam);
             }
             $this->owner = null;
         }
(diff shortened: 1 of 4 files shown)


framework/base/Behavior.php
```diff
@@ -86,7 +86,15 @@ public function detach()
     {
         if ($this->owner) {
             foreach ($this->events() as $event => $handler) {
-                $this->owner->off($event, is_string($handler) ? [$this, $handler] : $handler);
+                if (is_string($handler)) {
+                    $sParam = [$this, $handler];
```
samdark: What does `sParam` mean?

framework/base/Behavior.php
```diff
@@ -86,7 +86,15 @@ public function detach()
...
+                } else if ($handler instanceof \Closure) {
```
samdark: `else if` → `elseif`

samdark (changes requested):

framework/base/Behavior.php
GHopperMSK: `second param`

samdark: Second parameter of the method is `$handler`. I'd suggest naming the variable like that instead of "sParam".

GHopperMSK:
php below 7.0 doesnt's support `class` statement. That is why the tests broke. 
How can I specify php version for a specific test?

samdark:
Why using anonymous classes? Can't we create a class in a separate file instead?

GHopperMSK:
According to the [problem](https://github.com/yiisoft/yii2/issues/17223), we have to use at least one anonymous class (event handler).

samdark:
Ah. Right. Then the test could be skipped like in https://github.com/yiisoft/yii2/blob/da447fd1bb9778b8b76994f3e3e14f9748c67625/tests/framework/db/mysql/ActiveRecordTest.php#L25

samdark (approved):

GHopperMSK:
Can somebody explain me, how to fix codeclimate issue?

samdark:
What it doesn't like is [cognitive complexity](https://docs.codeclimate.com/docs/cognitive-complexity). Could be possibly fixed by extracting getting a handler into separate method.

tests/framework/base/ComponentTest.php
```diff
@@ -445,6 +445,25 @@ public function testDetachNotAttachedHandler()
...
+     * @see https://github.com/yiisoft/yii2/issues/17223
+     */
+    public function testEventClosureDetachesItself()
+    {
+        if (PHP_VERSION_ID < 70000) {
+            $this->markTestSkipped('Can not be tested on PHP < 7.0');
+            return;
+        }
+
+        $obj = include 'stub/AnonymousComponentClass.php';
```
rob006:
```suggestion
        $obj = require __DIR__ . '/stub/AnonymousComponentClass.php';
```

GHopperMSK: Should we improve this code too?
https://github.com/yiisoft/yii2/blob/caf9a28657eca8e49ead3d79a9b24a2439f5a829/tests/framework/base/ModelTest.php#L497

samdark: Yes.

laxity7:
Are you sure that it will be correct to remove a event completely?
What if the event is ActiveRecord :: EVENT_BEFORE_UPDATE?

framework/base/Behavior.php
```diff
@@ -85,8 +85,16 @@ public function attach($owner)
     public function detach()
     {
         if ($this->owner) {
-            foreach ($this->events() as $event => $handler) {
-                $this->owner->off($event, is_string($handler) ? [$this, $handler] : $handler);
+            foreach ($this->events() as $event => $value) {
+                if (is_string($value)) {
+                    $handler = [$this, $value];
+                } elseif ($value instanceof \Closure) {
+                    $handler = null;
```
laxity7: This will also remove other event handlers.

Are you sure that it will be correct to remove a event completely? 
What if the event is ActiveRecord::EVENT_AFTER_FIND?

GHopperMSK: I thought a lot about this problem, but I didn't find a solution. We just can't distinct one Closure from another one. That is why we can't remove one specific handler from event list.

samdark:
See https://github.com/yiisoft/yii2/issues/17223#issuecomment-479189458

Issue #17220, opened 2019-03-23
#16681 cause error if the $config param not already include 'options' key


webinjaz:
#16681

```php
    public function widget($class, $config = [])
    {
        foreach ($this->inputOptions as $key => $value) {
            if (!isset($config['options'][$key])) {
                $config['options'][$key] = $value;
            }
        }
```

must be 

```php
    public function widget($class, $config = [])
    {
        foreach ($this->inputOptions as $key => $value) {
             if(!isset($config['options']){
                  $config['options'] = [];
              }
            if (!isset($config['options'][$key])) {
                $config['options'][$key] = $value;
            }
        }
```

webinjaz:
i have this code and i get error on it

`					$form->field($model, "xs_$lang")->widget(PhotoInput::className(), [
						'form' => $form,
						'imgWidth' => 200,
						'imgHeight' => 200 * $mod->getPhotoSaveDimension('xs')['height'] / ($mod->getPhotoSaveDimension('xs')['width']?:400),
					]).`

samdark:
https://github.com/yiisoft/yii2/commit/a7f98964151d5894d9862fb0a1bf80040de86b69

samdark:
Reverted. See https://3v4l.org/D988b

s1lver:
That doesn't solve the problem
```php
<?= $form->field($model, 'test')
    ->widget(Test::class)
    ->label(false); ?>
```

```
Unknown Property – yii\base\UnknownPropertyException
Setting unknown property: app\components\test\Test::options
```

samdark:
Please post code of Test class.

rob006:
This foreach probably should be moved under `if (is_subclass_of($class, 'yii\widgets\InputWidget'))` condition - `inputOptions` should be populated only for `InputWidget` instances.

s1lver:
I use to pass in attributes.
```php
public $model;
public $attribute;
```
If you add ```$options```, everything works. But, this worked and until 2.0.17. I'm not sure this fix should be like this

> This foreach probably should be moved under `if (is_subclass_of($class, 'yii\widgets\InputWidget'))` condition - `inputOptions` should be populated only for `InputWidget` instances.

It doesn't work.

rob006:
@s1lver How it could not work if `InputWidget` has `options` property? You will never get `UnknownPropertyException` exception in this case.

s1lver:
@rob006 the property is, because I added it there to avoid errors. But, why this surplus action?

[...]

PR #17229, opened 2019-03-27
Fixed (#17220)

s1lver:
Fixed error if the $config param not already include 'options' key

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #17220

### DIFF
--- a/framework/widgets/ActiveField.php
+++ b/framework/widgets/ActiveField.php
@@ -760,7 +760,6 @@ class ActiveField extends Component
      * @param string $class the widget class name.
      * @param array $config name-value pairs that will be used to initialize the widget.
      * @return $this the field object itself.
-     * @throws \Exception
      */
     public function widget($class, $config = [])
     {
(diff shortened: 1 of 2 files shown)


framework/widgets/ActiveField.php
```diff
@@ -760,19 +760,20 @@ public function radioList($items, $options = [])
      * @param string $class the widget class name.
      * @param array $config name-value pairs that will be used to initialize the widget.
      * @return $this the field object itself.
+     * @throws \Exception
```
rob006: ?

rob006 (approved):

framework/widgets/ActiveField.php
s1lver: PHPDoc comment doesn't contain all necessary @throws tag
```php
$this->parts['{input}'] = $class::widget($config);
```

rob006: It is useless annotation - basically every method may throw `\Exception`, and without any description it does not give you any value. Such noisy annotations are the reason why I disabled this inspection completely.

s1lver: probably, but I will observe the accepted practice

PR #17234, opened 2019-03-28
Fixed bug with integer attributes

### LINKED ISSUE #17233: Validator attribute problem since version 2.0.17
image

We are creating a dynamic model and use the ActiveForm widget afterwards to render the model, this seems to throw the above error since version 2.0.17, with version 2.0.16 it works.

```php
$model = new DynamicModel();
foreach ($organisation->properties as $prop) {
            $model->defineAttribute($prop->id);
            $model->addRule([$prop->id], SafeValidator::class);
            $model->{$prop->id} = ApiOrganisationPropertyValue::getValue($prop, $this->attachedContext);
}
```

Now using ActiveForm for the generate dynamic model throws the warning above (screenshot).

```php
<?= $form->field($model, $vp->id);
```

(i made a check on the `$model` to see if the attribute exists - yes it does. via `$model->hasAttribute($vp->id)`)

No warrnings.

### What do you get instead?

The warnning screenshot

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.17
| PHP version      | 
| Operating system |

nadar:
If an attribute is an integer value, is_string won't work, and a warning will be raised.

This will fix #17233

Maybe we should use 

```php
if (!is_array($attributes)) {

}
```

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | #17233


framework/validators/Validator.php
```diff
@@ -277,7 +277,7 @@ public function getValidationAttributes($attributes = null)
             return $this->getAttributeNames();
         }
 
-        if (is_string($attributes)) {
+        if (is_scalar($attributes)) {
```
samdark: Why do you have int there? It's a list of attribute names according to phpdoc of both parent method and this method.

rob006: @samdark Technically object in PHP can have property like `123`.

nadar: We have generated the model and the form dynamically, so to have an unique identifier we have used the tables id - its not very common indeed, but it can happen though. Therefore just checking `if (!is_array($attributes))` would be safe as well.

samdark: `!is_array` is true for objects and resources so it looks like a bad idea.

samdark: Are `float` properties possible?

samdark: I've put together a test case: https://3v4l.org/dQ5Yv

All the properties are strings...

samdark: So yes, your fix could be applied, I think.

samdark:
@nadar would you please put together a unit test for it and add a line for changelog?

nadar:
@samdark yes of course

framework/validators/Validator.php
rob006: > All the properties are strings...

`var_dump()` will report `string`, but integer properties will be cast to `int` if you try to export them as array: https://3v4l.org/4YZcV

samdark:
@nadar still interested in this one?

nadar:
Yes

nadar:
@samdark I just detected that Validator::getAttributeNames() will convert int values into strings, therefore strict compare of in_array won't work anymore. So, either remove strict compare or take the solution i submitted which wont apply ltrim on int values. What do you think?

samdark:
Non-strict comparison, in my opinion, looks better.

samdark:
Travis tests failed btw.

nadar:
Ok, i removed strict compare. Let me know if you should remove the comment, but it explains why this has ben explizit turned off.

(Yes it failed - PHP 5.4 ftw!)

nadar:
Do you need a changelog entry? Would need to merge upstream first.

samdark:
Yes. Need a changelog entry. Updated branch from master.

samdark:
Merged. Thank you!

PR #17250, opened 2019-04-03
17223 detach behavior event

np25071984:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17223


framework/base/Behavior.php
```diff
@@ -27,6 +27,10 @@ class Behavior extends BaseObject
      */
     public $owner;
 
+    /**
+     * @var Array Attached events handlers
+     */
+    public $_attachedEvents = [];
```
laxity7: Why publicly?

samdark:
Seems this one won't be merged as the issue isn't staightforward at all...

framework/base/Behavior.php
```diff
@@ -85,7 +90,7 @@ public function attach($owner)
     public function detach()
     {
         if ($this->owner) {
-            foreach ($this->events() as $event => $handler) {
+            foreach ($this->_attachedEvents as $event => $handler) {
```
rob006: Maybe we should call `unset($this->_attachedEvents[$event]);` after `$this->owner->off()`? Or reset `$_attachedEvents` at the and of `detach()`?

samdark: Done.

rob006 (approved):

Issue #17268, opened 2019-04-15
Regression in Formatter fallback methods for very small numbers


marcovtwout:
### What steps will reproduce the problem?
```Yii::$app->formatter->asCurrency(1E-10, 'EUR');```

### What is the expected result?
```€0.00```

### What do you get instead?
```EUR 1.0E```

### Additional info
The expected result was achieved before `asCurrencyStringFallback()` was added ( https://github.com/yiisoft/yii2/pull/16879)

Related similar issue: https://github.com/yiisoft/yii2/issues/16934

I applied the following quick workaround in my project:

    /**
     * Quick workaround when formatting very small numbers.
     * @see https://github.com/yiisoft/yii2/issues/16934#issuecomment-482078758
     *
     * @inheritdoc
     */
    public function asCurrency($value, $currency = null, $options = [], $textOptions = [])
    {
        if (abs($value) < 1E-9) {
            $value = 0.00;
        }
        
        return parent::asCurrency($value, $currency, $options, $textOptions);
    }

samdark:
Verified that the issue is there. Would you like to fix it with a pull request?

marcovtwout:
I just found this isolated issue, but perhaps this problem applies to the other formatters as well? Maybe the original author of the fallback @bizley can take a look?

samdark:
I took a look. It's, indeed, the fact that `normalizeNumericStringValue()` doesn't take "power" part into account.

samdark:
Here's a fix. @marcovtwout please check if it works for your cases.

marcovtwout:
@samdark Confirmed that this issue is solved with the referenced fix.

PR #17270, opened 2019-04-15
17268 fix formatter

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17268


GHopperMSK:
We can use `number_format` - it gets both a number and a string and returns a number. So, all we need is single line of code in Formatter class:


```
    public function asCurrency($value, $currency = null, $options = [], $textOptions = [])
    {

        $value = number_format($value, 10);
...
```

samdark:
@GHopperMSK the issue isn't in `asCurrency` only. It's decimals formatting overall.

samdark:
Also `number_format` can not deal with huge numbers.

GHopperMSK:
> Also `number_format` can not deal with huge numbers.

```
$ php -r "var_dump(number_format(1E-10, 100));"
string(102) "0.0000000001000000000000000036432197315497741579165547065599639608990401029586791992187500000000000000"
```

samdark:
It cannot deal with input such as `87,654,321,098,765,440.0000000000`.

GHopperMSK:
> It cannot deal with input such as `87,654,321,098,765,440.0000000000`.

I can't deal with such a number too )

samdark:
Well, current implementation can and there are tests for it.

PR #17283, opened 2019-04-22
Fixes #12763 make framework honor default database schema

### LINKED ISSUE #12763: Postgresql schema ignored
1. Install Yii2 advanced app using composer and run `./init`.
2. Create a Postgresql database.
3. Create a schema in the database (different from the user default schema so it is not in the search path).
4. Edit the database config in common/config/main-local.php as follows: 
```
return [
    'components' => [
        'db' => [
            'dsn' => 'pgsql:host=myserver;dbname=mydb',
            'username' => 'user',
            'password' => 'password',
            'charset' => 'utf8',
            'schemaMap' => [
                'pgsql'=> [
                    'class'=>'yii\db\pgsql\Schema',
                    'defaultSchema' => 'myschema',
                ],
            ],
        ],
        ...
];
```
5. Run migration: `./yii migrate`.

Migration and user tables should exist under `myschema` in the database.

### What do you get instead?
Tables are created under a different schema (the one that I assume is the default for the user - I don't have admin rights to the database).

### Additional info
Using a `controllerMap` directive I can get the `migration` table created under the correct schema (haven't found out how to do that with the `user` table yet):

```
    'controllerMap' => [
        'migrate' => [
            'class' => 'yii\console\controllers\MigrateController',
            'migrationTable' => "myschema.migration",
        ],
    ],

```
But without specifying a `controllerMap` for `user` then that table is still created under the wrong schema.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.9
| PHP version      | 5.5.38 / 7.0.10
| Operating system | MacOs X (El Capitan)

mtangoo:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #12763

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -342,6 +342,7 @@ class QueryBuilder extends \yii\base\BaseObject
      */
     public function insert($table, $columns, &$params)
     {
+        $table = $this->db->schema->normalizeTableName($table);
         list($names, $placeholders, $values, $params) = $this->prepareInsertValues($table, $columns, $params);
         return 'INSERT INTO ' . $this->db->quoteTableName($table)
             . (!empty($names) ? ' (' . implode(', ', $names) . ')' : '')
@@ -448,6 +449,8 @@ class QueryBuilder extends \yii\base\BaseObject
         if (empty($rows)) {
             return '';
         }
+        
+        $table = $this->db->schema->normalizeTableName($table);
 
         $schema = $this->db->getSchema();
         if (($tableSchema = $schema->getTableSchema($table)) !== null) {
@@ -622,6 +625,7 @@ class QueryBuilder extends \yii\base\BaseObject
      */
     public function update($table, $columns, $condition, &$params)
     {
+        $table = $this->db->schema->normalizeTableName($table);
         list($lines, $params) = $this->prepareUpdateSets($table, $columns, $params);
         $sql = 'UPDATE ' . $this->db->quoteTableName($table) . ' SET ' . implode(', ', $lines);
         $where = $this->buildWhere($condition, $params);
@@ -643,7 +647,6 @@ class QueryBuilder extends \yii\base\BaseObject
         $columnSchemas = $tableSchema !== null ? $tableSchema->columns : [];
         $sets = [];
         foreach ($columns as $name => $value) {
-
             $value = isset($columnSchemas[$name]) ? $columnSchemas[$name]->dbTypecast($value) : $value;
             if ($value instanceof ExpressionInterface) {
                 $placeholder = $this->buildExpression($value, $params);
@@ -676,6 +679,7 @@ class QueryBuilder extends \yii\base\BaseObject
      */
     public function delete($table, $condition, &$params)
     {
+        $table = $this->db->schema->normalizeTableName($table);
         $sql = 'DELETE FROM ' . $this->db->quoteTableName($table);
         $where = $this->buildWhere($condition, $params);
 
@@ -718,6 +722,7 @@ class QueryBuilder extends \yii\base\BaseObject
                 $cols[] = "\t" . $type;
             }
         }
+        $table = $this->db->schema->normalizeTableName($table);
         $sql = 'CREATE TABLE ' . $this->db->quoteTableName($table) . " (\n" . implode(",\n", $cols) . "\n)";
 
         return $options === null ? $sql : $sql . ' ' . $options;
@@ -731,6 +736,8 @@ class QueryBuilder extends \yii\base\BaseObject
      */
     public function renameTable($oldName, $newName)
     {
+        $oldName = $this->db->schema->normalizeTableName($oldName);
+        $newName = $this->db->schema->normalizeTableName($newName);
         return 'RENAME TABLE ' . $this->db->quoteTableName($oldName) . ' TO ' . $this->db->quoteTableName($newName);
     }
 
@@ -741,6 +748,7 @@ class QueryBuilder extends \yii\base\BaseObject
      */
     public function dropTable($table)
     {
+        $table = $this->db->schema->normalizeTableName($table);
         return 'DROP TABLE ' . $this->db->quoteTableName($table);
     }
 
@@ -761,6 +769,7 @@ class QueryBuilder extends \yii\base\BaseObject
             $columns[$i] = $this->db->quoteColumnName($col);
         }
 
+        $table = $this->db->schema->normalizeTableName($table);
         return 'ALTER TABLE ' . $this->db->quoteTableName($table) . ' ADD CONSTRAINT '
             . $this->db->quoteColumnName($name) . ' PRIMARY KEY ('
             . implode(', ', $columns) . ')';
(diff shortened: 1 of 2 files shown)


samdark:
Seems to fail tests.

mtangoo:
> Seems to fail tests.

How do I test the framework alone after cloning? I could not get the information.

mtangoo:
@samdark I was investigating that and it fails in the case like this:

```php
$expected = $this->replaceQuotes('SELECT * FROM (SELECT * FROM [[user]] WHERE account_id = :id) [[activeusers]] WHERE abc = :abc');
```

table `[[user]] isn't prefixed with schema name. I have a question on the better way to prefix the schema names so that the tests pass.  I plan to do something like

```php
$schemaName = $this->getQueryBuilder()->getConnection(false, true)->getSchema()->defaultSchema;
$expected = $this->replaceQuotes("SELECT * FROM (SELECT * FROM [[{$schemaName}]].[[user]] WHERE account_id = :id) [[activeusers]] WHERE abc = :abc ");
```
Is this acceptable?

samdark:
Since it was OK before, I assume there's some un-prefixing goes on.

mtangoo:
>Since it was OK before, I assume there's some un-prefixing goes on. Maybe it's better to fix it? 

Yes, I didn't fix prefixing of the table on tests. Above is the solution I wanted you to review before I apply it. I will go ahead and apply it.

samdark:
If you'll fix it in tests only that would mean that behavior is no more compatible with old code. That's not something we want to 2.x.

mtangoo:
> If you'll fix it in tests only that would mean that behavior is no more compatible with old code. That's not something we want to 2.x.

Compatibility is not broken here is why: The new code appends schema name only when it is defined in the configuration so that `tableName` becomes `schema.tableName`. Of course it will break the code where user careless put wrong `defaultSchema` in config (am not sure if this amounts to breaking BC), otherwise from database POV there is no difference.

So what I need to make right now is, the tests should pass if `defaultSchema` isn't defined as well as when it is indeed defined. I hope that will amount to backward compatibility with old behavior. No?

samdark:
Yes. That should be fine.

samdark:
@mtangoo do you have time to finish it?

mtangoo:
@samdark I'm working on it. It should be finished within this week. I have done most of the work before something intefered

mtangoo:
am still fixing this. A lot of code to change but am progressing. I hope I will finish sooner!

brandonkelly:
@mtangoo Have you given up on this?

mtangoo:
@brandonkelly not yet. But the changes are so huge and I have to go manually check, edit and run tests. Unfortunately I have a lot on the table right now. But as soon as things cedes I will revisit and finish. It is a feature that I need!

leandrogehlen:
@mtangoo, some progress about this issue?
Do you need some help?

mtangoo:
@brandonkelly sure, help is appreciated. I will try to work on whatever I have and remaining and update here. Then if you can help finalizing remaining part it will be great!

brandonkelly:
@mtangoo @leandrogehlen I’m not sure this is the right approach (see https://github.com/yiisoft/yii2/issues/12763#issuecomment-522231992).

mtangoo:
@brandonkelly can you make a PR with correct approach? I would be happy to test it tomorrow, since I will have some "free" time

brandonkelly:
I think the current behavior may already be the correct approach; the only issue may be that the docs don’t do a good job explaining what the point of the setting is.

leandrogehlen:
@brandonkelly. I have different applications where each other uses itself schema.
How to define settings to run this migrations  currently whitou this implementation?

brandonkelly:
@leandrogehlen If you have flexibility on the schema names, the easiest way to do that would be to create two separate database users, and create schemas for your applications named after the usernames. Then configure the DSN strings to connect with the respective user accounts, and set `defaultSchema` to those usernames.

leandrogehlen:
@brandonkelly my applications are hosted on aws lightsail and is not allowed to add new users

brandonkelly:
@leandrogehlen In that case then I guess you just need to start naming the schema in every table reference within your app. You could make it configurable by extending `yii\db\Connection` with a custom `$schema` property on it, and configure that instead of `$defaultSchema`, then include its value whenever you are referencing table names. For example:

```php
{
    return '{{' . \Yii::$app->db->schema . '.%your_table_name}}';
}
```

leandrogehlen:
I disagree, the framework must be able to solve it via `defaultSchema` property

brandonkelly:
@leandrogehlen That’s what I assumed `$defaultSchema` was for initially, but in reality it has the opposite purpose – to *remove* the schema from queries when it is not necessary. We should not change it simply because we wanted it to mean something else.

Changing its behavior would also be a breaking change. If a Yii update were to start adding the `$defaultSchema` onto queries automatically, all of a sudden it’s critically important that you have it set to the correct schema name, whereas currently it doesn’t matter. (You can currently run a Yii app on a non-`public` schema, while leaving `$defaultSchema = 'public'`.)

That said, I do think there is a case for adding a *separate* `$schema` DB setting that achieves the behavior you want, perhaps from within `getRawTableName()`.

```php
public function getRawTableName($name)
{
    if (strpos($name, '{{') !== false) {
        $name = preg_replace_callback('/\\{\\{(.*?)\\}\\}/', function($match) {
            $name = $match[1];
            if ($this->schema !== null && strpos($name, '.') === false) {
                $name = "{$this->schema}.{$name}";
            }
            return str_replace('%', $this->db->tablePrefix, $name);
        }, $name);
    }

    return $name;
}
```

mtangoo:
@leandrogehlen @brandonkelly there is one more use case actually that need to be covered namely multi-tenant schema based implementation. The default schema for logged in user is dynamic depending on which user is logged in. In this aspect how do you tackle that?

brandonkelly:
@mtangoo Set the property dynamically, then. And there’s nothing stopping you from setting up multiple DB connection components (one primary one (`db`) that points to a central schema like `public` which contains all user account info, and a secondary one (`userDb`) that is configured with `$schema` set to the logged-in username).

mtangoo:
@brandonkelly woudn't that create new connection each time, for example, the  model `getTable()` get called?

mtangoo:
>That’s what I assumed $defaultSchema was for initially, but in reality it has the opposite purpose – to remove the schema from queries when it is not necessary. We should not change it simply because we wanted it to mean something else.

This PR addresses that, in that the added method checks if default schema is defined and forces the table to use it if it is indeed not using it. for tables that have schema already `schema.table` or when schema is `null` It just ignores the adding of schema.

The only thing that I have failed to finish is, this change requires a lot of fixing for the tests to work again: too many places to change.

brandonkelly:
> woudn't that create new connection each time, for example, the model `getTable()` get called?

No, it would create one connection the first time its `open()` method is called, just like `db`.

> This PR addresses that, in that the added method checks if default schema is defined and forces the table to use it if it is indeed not using it.

That is the problem. `defaultSchema` is set to `'public'` by default, so with this change, suddenly it will start modifying queries in a way that it wasn’t before, leading to backwards compatibility issues. Which is why I think we should do basically the same thing that this PR does, but with a new `$schema` property that is `null` by default.

terabytesoftw:
Closing this PR because it has been inactive since 2019 and the proposed approach changes the semantics of `defaultSchema`, which is `public` by default for PostgreSQL and may introduce backward compatibility issues.

The underlying use case in #12763 may still be valid, but it should be addressed separately, either with documentation around PostgreSQL `search_path` or with a new opt-in setting that is null/disabled by default.

PR #17306, opened 2019-05-08
Add .mjs mimetype

mreinstein:
Google and node.js represent `.mjs`  as ecmascript modules:
* https://developers.google.com/web/fundamentals/primers/modules#mjs
* https://nodejs.org/dist/latest-v10.x/docs/api/esm.html#esm_enabling

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | n/a

### DIFF
--- a/framework/helpers/mimeTypes.php
+++ b/framework/helpers/mimeTypes.php
@@ -448,6 +448,7 @@ return [
     'mime' => 'message/rfc822',
     'mj2' => 'video/mj2',
     'mjp2' => 'video/mj2',
+    'mjs' => 'text/javascript',
     'mk3d' => 'video/x-matroska',
     'mka' => 'audio/x-matroska',
     'mks' => 'video/x-matroska',


samdark:
That would be overwritten by our mime-type update script that takes info from Apache server mime types. It's a good idea to fix it instead: https://github.com/yiisoft/yii2/blob/master/build/controllers/MimeTypeController.php#L33

mreinstein:
@samdark what would i put in `MimeTypeController.php`? I don't see `mjs` in Apache's `mime.types` file, so it's not really mapping from an existing type to another.

samdark:
Yes. It the command can't currently add mime-types that aren't present in apache's mime.types. Could be extended for it, I think.

samdark:
Fixed differently.

PR #17322, opened 2019-05-23
create AssetManager base path dir when init

noname007:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no


Thoulah:
Would this overwrite $this->basePath is that happens to be a file or a link..?

samdark:
Unfortunately, enhancements aren't accepted to Yii 2 anymore. That could be fixed in Yii 3 though.

samdark:
Thanks for contributing anyway.

PR #17324, opened 2019-05-24
Removed unnecessary ternary operator

germanow:
Why is there ternary operator which creating Behavior before calling attachBehavior() if attachBehaviorInternal() does it later? I think it is not necessary. 

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

### DIFF
--- a/framework/base/Component.php
+++ b/framework/base/Component.php
@@ -188,7 +188,7 @@ class Component extends BaseObject
         } elseif (strncmp($name, 'as ', 3) === 0) {
             // as behavior: attach behavior
             $name = trim(substr($name, 3));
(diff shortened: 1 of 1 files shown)


rob006:
This looks valid, but I'm not sure if it is worth to change it. We already had situations where such cleanups created bugs, because code was more tricky than we suspected. Technically it could also break some existing code if someone override `attachBehavior()` and don't expect anything except `Behavior` instance.

samdark:
> Technically it could also break some existing code if someone override attachBehavior() and don't expect anything except Behavior instance.

Right. A valid reason not to change it.

Issue #17336, opened 2019-06-04
This line to match if the class has event is wrong?


lubobill1990:
https://github.com/yiisoft/yii2/blob/2ca8da062cc58f5ec778b560dc9e65ada9306166/framework/base/Event.php#L244

Here, the anticipated condition is that, if the $classWildcard match the $class, then return false.

```
if (!StringHelper::matchWildcard($classWildcard, $class)) {
```
should be 
```
if (StringHelper::matchWildcard($classWildcard, $class)) {
```

PR #17337, opened 2019-06-04
return true when $class matches

lubobill1990:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17336

### DIFF
--- a/framework/base/Event.php
+++ b/framework/base/Event.php
@@ -241,7 +241,7 @@ class Event extends BaseObject
                     continue;
                 }
                 foreach ($classes as $class) {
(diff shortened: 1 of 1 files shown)


bizley:
There were obviously no tests for this line. Could you add some to cover it?

samdark:
Implemented tests with https://github.com/yiisoft/yii2/pull/17338. Also fixed non-match for FQN.

lubobill1990:
Thanks!

PR #17338, opened 2019-06-04
Fixes #17336: Fixed wildcard matching in Event::hasHandlers()

### LINKED ISSUE #17336: This line to match if the class has event is wrong?

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17336


schmunk42 (approved):
LGTM

PR #17344, opened 2019-06-06
Improve performance of select() and addSelect()

brandonkelly:
We just profiled a 25.8s request with Blackfire, and found that 11.5s / 44.5% was spent on `yii\db\Query::getUniqueColumns()` (called by `select()` and `addSelect()`), and its callee, `getUnaliasedColumnsFromSelect()`.

Screen Shot 2019-06-06 at 3 21 25 PM

[View on Blackfire ➡️](https://blackfire.io/profiles/26b68fc8-ccb8-424c-a9b6-06010fbb06e9/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=yii%5Cdb%5CQuery%3A%3AgetUniqueColumns%3F34f7364a45ee985db52f3b3a2c668108&callname=main())

We rewrote `select()` and `addSelect()` to be more efficient, avoiding `getUniqueColumns()` altogether and ensuring uniqueness by normalizing SELECT columns similar to the way `orderBy()` / `normalizeOrderBy()` works. Now aliases are always normalized into the `['alias' => 'column']` syntax, and even un-aliased _column names_ (excluding expressions, sub-selects, and strings with parentheses) are aliased to themselves (`['column' => 'column']`).

The result:  `select()` isn’t even showing up on Blackfire’s radar anymore, and `addSelect()` is only taking up 663ms (17.3x faster), bringing the request time down to 14.1s (1.8x faster).

Screen Shot 2019-06-06 at 3 53 03 PM

[View on Blackfire ➡️](https://blackfire.io/profiles/62b0108f-2492-406b-8352-a6c8bfd8cc14/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=craft%5Cdb%5CQuery%3A%3AaddSelect&callname=main())

We hope that is a performance improvement worth considering :)

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -6,6 +6,7 @@ Yii Framework 2 Change Log
 
 - Bug #17341: Fixed error from yii.activeForm.js in strict mode (mikehaertl)
 - Enh #17345: Improved performance of `yii\db\Connection::quoteColumnName()` (brandonkelly)
+- Enh #17344: Improved performance of `yii\db\Connection::addSelect()` (brandonkelly)
 
 
 2.0.20 June 04, 2019
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -656,8 +656,12 @@ PATTERN;
         foreach ($columns as $columnAlias => $columnDefinition) {
             if (is_string($columnAlias)) {
                 $select[$columnAlias] = $columnDefinition;
-            } else if (is_string($columnDefinition) && preg_match('/^\w+$/', $columnDefinition)) {
-                $select[$columnDefinition] = $columnDefinition;
+            } elseif (is_string($columnDefinition) && strpos($columnDefinition, '(') === false) {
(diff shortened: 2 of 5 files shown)


brandonkelly:
@samdark Sorry for being a little trigger happy on this PR, but it’s good to go now :)

samdark (approved):

samdark:
Merged. Thank you!

samdark:
I've tested on some projects I have without issues.

brandonkelly:
Awesome! The only change we had to make in Craft was due to an `array_splice()` call on `$query->select()`, as the key that was passed in was no longer numeric. Doubt many people are doing things like that with normal queries. Probably worth mentioning that `yii\db\Query::select()` and `addSelect()` now normalize the columns when saving them to `$this->select`, so code that works directly with that property may need to be modified.

PR #17346, opened 2019-06-07
Fixed #17340

### LINKED ISSUE #17340: Problem to inherit yii\db\Query
The problem is that when you want to inherit after yii\db\Query class I have to also overwrite some functions that refer to new self().
Changing self to static should eliminate possible errors that might appear in such cases and it will maintain object consistency. 
https://github.com/yiisoft/yii2/blob/869c8aa22059f3aaa1dbdce0b74b849f0d7cb189/framework/db/Query.php#L466-L469
image

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.3
| PHP version      | 7.3.x
| Operating system |

rskrzypczak:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | #17340

### DIFF
--- a/framework/db/Query.php
+++ b/framework/db/Query.php
@@ -463,7 +463,7 @@ class Query extends Component implements QueryInterface, ExpressionInterface
             return $command->queryScalar();
         }
 
-        $command = (new self())
+        $command = (new static())
             ->select([$selectExpression])
             ->from(['c' => $this])
             ->createCommand($db);
@@ -1282,7 +1282,7 @@ PATTERN;
      */
     public static function create($from)
     {
-        return new self([
+        return new static([
             'where' => $from->where,
             'limit' => $from->limit,
             'offset' => $from->offset,


samdark:
Seems the change breaks things: https://travis-ci.org/yiisoft/yii2/jobs/542600371

samdark:
@rskrzypczak are you going to work on it further?

rob006:
This is not that simple: https://github.com/yiisoft/yii2/pull/14729, https://github.com/yiisoft/yii2/pull/16993.

It could be fixed by introducing another protected method for creating instance used by these two methods, so descendants could override it. Changing this to `static` directly in `Query` would break BC.

samdark:
Yes. Seems it can't be applied in this form. Closing.

PR #17347, opened 2019-06-07
Update path configuration for composer-asset-plugin

michalwoz:
Composer install don't take "installer-paths" on the level of "config". The path which works: "config" => "fxp-asset" => "installer-paths"

Documentation

### DIFF
--- a/docs/guide/structure-assets.md
+++ b/docs/guide/structure-assets.md
@@ -266,9 +266,11 @@ will be placed, if you want to publish them using Yii:
 
 ```json
 "config": {
-    "asset-installer-paths": {
-        "npm-asset-library": "vendor/npm",
-        "bower-asset-library": "vendor/bower"
+    "fxp-asset": {
+        "installer-paths": {
+            "npm-asset-library": "vendor/npm",
+            "bower-asset-library": "vendor/bower"
+        }
     }
 }
 ```


Thoulah:
The default is fine, but only if you have fxp plugin install your change is required. This is already in the [docs](https://github.com/yiisoft/yii2/blob/98872c1e7dc02ae9b1c4668fb34d895481708cd5/docs/guide/structure-assets.md#using-fxpcomposer-asset-plugin).

michalwoz:
> The default is fine, but only if you have fxp plugin install your change is required. This is already in the [docs](https://github.com/yiisoft/yii2/blob/98872c1e7dc02ae9b1c4668fb34d895481708cd5/docs/guide/structure-assets.md#using-fxpcomposer-asset-plugin).

Yes that piece of composer.json regards only the case when you use fxp plugin. But it seems that there were bug in the documentation. There was missing "fxp-asset" key. without proposed fix composer with fxp plugin installs bower to vendor/bower-asset, so either you have to change app config or create symlink vendor/bower -> vendor/bower-asset. With proposed fix composer installed bower to vendor/bower and no other steps are required.

Thoulah:
My bad, sorry

docs/guide/structure-assets.md
```diff
@@ -266,9 +266,11 @@ will be placed, if you want to publish them using Yii:
 
 ```json
 "config": {
-    "asset-installer-paths": {
```
samdark: @SilverFire, @hiqsol does that affect https://asset-packagist.org?

samdark: No, it won't be affected.

PR #17362, opened 2019-06-12
Added type null to @return annotation

epifrin:
Added type null to @return annotation of ActiveRecordInterface::findOne() function.
BaseActiveRecord::findOne() has @return static|null but ActiveRecordInterface::findOne() @return static
[phpstan](https://github.com/phpstan/phpstan) on level 4 returns "If condition is always true." error for next example of code:
```php
if ($user = User::findOne($id)) {
    // some code
}
```
but variable $user can be null

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |


samdark:
Merged. Thanks!

Issue #16796, opened 2018-10-16
Error in sp_updateextendedproperty when adding field comments


sdlins:
### What steps will reproduce the problem?
```php
$this->createTable('MY_TABLE', [
            'name' => $this->string(32)->notNull()
                ->comment('An unique name for this group like group_fisc or group_cash'),
...
```

### Results in
```shell
*** applying m181009_061513_modarq_init
    > create table MY_TABLE ...Exception: SQLSTATE[42000]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]
An invalid parameter or option was specified for procedure 'sp_updateextendedproperty'.
The SQL being executed was: 
sp_updateextendedproperty @name = N'MS_Description', 
@value = 'An unique name for this group like group_fisc or group_cash', 
@level1type = N'Table',  @level1name = [MY_TABLE], 
@level2type = N'Column', @level2name = [name] (C:\Users\...\vendor\yiisoft\yii2\db\Schema.php:664)
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      |  7.1
| SQL Server      |  13.0
| Operating system | Win10 64

tourze:
same problem. has it fixed?

samdark:
@tourze no. The issue isn't closed.

sdlins:
I am tracing this again. 

Till now I found MSSQL has two functions: `sp_updateextendedproperty` and `sp_addextendedproperty`.
I can test wether the description already exists before update or add a description. After `add` the description an `update` doesn't generate any errors.

Hope to make a PR next weeks, according to my free time.

sdlins:
@samdark I would appreciate your help in a question.

This would be the new `mssql\QueryBuilder::addCommentOnTable()` method:
```php
    public function addCommentOnTable($table, $comment)
    {
        $schemaName = $this->db->schema->defaultSchema;
        $comment = $this->db->quoteValue($comment);

        return <<< SQL
        -- Maybe we can simplify this later
        IF
            NOT EXISTS (
                SELECT 1
                FROM fn_listextendedproperty (
                    NULL,
                    'SCHEMA', N'$schemaName',
                    'TABLE', N'$table',
                    null, null
                )
            )
        BEGIN
            EXEC sys.sp_addextendedproperty
                @name = N'MS_description',
                @value = $comment,
                @level0type = N'SCHEMA', @level0name = N'$schemaName',
[...]

PR #17364, opened 2019-06-13
Fixes comments for table or column

sdlins:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #16796

### DIFF
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -260,11 +260,16 @@ class QueryBuilder extends \yii\db\QueryBuilder
      * @param string $column optional. The name of the column to be commented. If empty, the command will add the
      * comment to the table instead. The column name will be properly quoted by the method.
      * @return string the SQL statement for adding comment.
+     * @throws InvalidArgumentException if the table does not exist.
      * @since 2.0.24
      */
     protected function buildAddCommentSql($comment, $table, $column = null)
     {
         $tableSchema = $this->db->schema->getTableSchema($table);
+
+        if ($tableSchema === null) {
+            throw new InvalidArgumentException("Table not found: $table");
+        }
 
         $schemaName = $tableSchema->schemaName ? "N'" . $tableSchema->schemaName . "'": 'SCHEMA_NAME()';
         $tableName = "N" . $this->db->quoteValue($tableSchema->name);
@@ -321,11 +326,16 @@ class QueryBuilder extends \yii\db\QueryBuilder
      * @param string $column optional. The name of the column whose comment will be removed. If empty, the command
      * will remove the comment from the table instead. The column name will be properly quoted by the method.
      * @return string the SQL statement for removing the comment.
+     * @throws InvalidArgumentException if the table does not exist.
      * @since 2.0.24
      */
     protected function buildRemoveCommentSql($table, $column = null)
     {
         $tableSchema = $this->db->schema->getTableSchema($table);
+
+        if ($tableSchema === null) {
+            throw new InvalidArgumentException("Table not found: $table");
+        }
 
         $schemaName = $tableSchema->schemaName ? "N'" . $tableSchema->schemaName . "'": 'SCHEMA_NAME()';
         $tableName = "N" . $this->db->quoteValue($tableSchema->name);
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -250,9 +250,6 @@ class QueryBuilder extends \yii\db\QueryBuilder
         return $command;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     protected function mountCommonSqlForCommentAddition($comment, $table, $column = null)
     {
         $tableSchema = $this->db->schema->getTableSchema($table);
@@ -300,7 +297,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
      */
     public function addCommentOnTable($table, $comment)
     {
-        return $this->mountCommonSqlForCommentAddition($comment, $table) ;
+        return $this->mountCommonSqlForCommentAddition($comment, $table);
     }
 
     /**
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -305,11 +305,39 @@ class QueryBuilder extends \yii\db\QueryBuilder
 
     /**
      * {@inheritdoc}
+     */
+    protected function mountCommonSqlForCommentRemoval($table, $column = null)
+    {
+        $tableSchema = $this->db->schema->getTableSchema($table);
+
+        $schemaName = $tableSchema->schemaName ? "N'" . $tableSchema->schemaName . "'": 'SCHEMA_NAME()';
+        $tableName = "N'{$tableSchema->name}'";
+        $columnName = $column ? "N'$column'" : null;
+
+        return "
+            IF EXISTS (
+                    SELECT 1
+                    FROM fn_listextendedproperty (
+                        N'MS_description',
+                        'SCHEMA', $schemaName,
+                        'TABLE', $tableName,
+                        " . ($column ? "'COLUMN', $columnName " : ' DEFAULT, DEFAULT ') . "
+                    )
+            )
+                EXEC sys.sp_dropextendedproperty
+                    @name = N'MS_description',
+                    @level0type = N'SCHEMA', @level0name = $schemaName,
+                    @level1type = N'TABLE', @level1name = $tableName"
+                    . ($column ? ", @level2type = N'COLUMN', @level2name = $columnName" : '') . ';';
+    }
+
+    /**
+     * {@inheritdoc}
      * @since 2.0.8
      */
     public function dropCommentFromColumn($table, $column)
     {
-        return "sp_dropextendedproperty @name = N'MS_Description', @level1type = N'Table',  @level1name = {$this->db->quoteTableName($table)}, @level2type = N'Column', @level2name = {$this->db->quoteColumnName($column)}";
+        return $this->mountCommonSqlForCommentRemoval($table, $column);
     }
 
     /**
@@ -318,7 +346,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
      */
     public function dropCommentFromTable($table)
     {
-        return "sp_dropextendedproperty @name = N'MS_Description', @level1type = N'Table',  @level1name = {$this->db->quoteTableName($table)}";
+        return $this->mountCommonSqlForCommentRemoval($table);
     }
 
     /**
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -251,15 +251,15 @@ class QueryBuilder extends \yii\db\QueryBuilder
     }
 
      /**
-     * Builds a SQL command for adding or updating a comment to table or column. The command builded will check if a comment
-     * already exists. If so, it will be updated, or it will be added otherwise.
+     * Builds a SQL command for adding or updating a comment to a table or a column. The command built will check if a comment
+     * already exists. If so, it will be updated, otherwise, it will be added.
      *
      * @param string $comment the text of the comment to be added. The comment will be properly quoted by the method.
      * @param string $table the table to be commented or whose column is to be commented. The table name will be
      * properly quoted by the method.
      * @param string $column optional. The name of the column to be commented. If empty, the command will add the
      * comment to the table instead. The column name will be properly quoted by the method.
-     * @return string the SQL statement for adding comment.
+     * @return string the SQL statement for adding a comment.
      * @throws InvalidArgumentException if the table does not exist.
      * @since 2.0.24
      */
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -318,8 +318,8 @@ class QueryBuilder extends \yii\db\QueryBuilder
     }
 
     /**
-     * Builds a SQL command for removing a comment of a table or column. The command builded will check if a comment
-     * already exists before trying to perform the removing.
+     * Builds a SQL command for removing a comment from a table or a column. The command built will check if a comment
+     * already exists before trying to perform the removal.
      *
      * @param string $table the table that will have the comment removed or whose column will have the comment removed.
      * The table name will be properly quoted by the method.
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -250,6 +250,18 @@ class QueryBuilder extends \yii\db\QueryBuilder
         return $command;
     }
 
+     /**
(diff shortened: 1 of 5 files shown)


sdlins:
[WIP]
Still lacking to fix `dropCommentFromColumn|Table()` and create tests. Pulling now to allow the code to be analysed in advance.

sdlins:
@samdark, it is done. Sorry for so many forced pushes.
Let me know if something should be modified.

framework/db/mssql/QueryBuilder.php
```diff
@@ -250,13 +250,48 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
         return $command;
     }
 
+    /**
+     * {@inheritdoc}
```
samdark: Is there such method in parent class?

framework/db/mssql/QueryBuilder.php
```diff
@@ -250,13 +250,48 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
...
+     */
+    protected function mountCommonSqlForCommentAddition($comment, $table, $column = null)
+    {
+        $tableSchema = $this->db->schema->getTableSchema($table);
+
+        $schemaName = $tableSchema->schemaName ? "N'" . $tableSchema->schemaName . "'": 'SCHEMA_NAME()';
+        $tableName = "N'{$tableSchema->name}'";
```
samdark: Should we use `quoteTableName` here?

framework/db/mssql/QueryBuilder.php
```diff
@@ -250,13 +250,48 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
...
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    protected function mountCommonSqlForCommentAddition($comment, $table, $column = null)
+    {
+        $tableSchema = $this->db->schema->getTableSchema($table);
+
+        $schemaName = $tableSchema->schemaName ? "N'" . $tableSchema->schemaName . "'": 'SCHEMA_NAME()';
```
samdark: Should schema name be quoted?

framework/db/mssql/QueryBuilder.php
```diff
@@ -250,13 +250,48 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
...
+        $tableName = "N'{$tableSchema->name}'";
+        $columnName = $column ? "N'$column'" : null;
```
samdark: Should column name be quoted?

framework/db/mssql/QueryBuilder.php
```diff
@@ -265,7 +300,35 @@ public function addCommentOnColumn($table, $column, $comment)
      */
     public function addCommentOnTable($table, $comment)
     {
-        return "sp_updateextendedproperty @name = N'MS_Description', @value = {$this->db->quoteValue($comment)}, @level1type = N'Table',  @level1name = {$this->db->quoteTableName($table)}";
+        return $this->mountCommonSqlForCommentAddition($comment, $table) ;
```
samdark:
```suggestion
        return $this->mountCommonSqlForCommentAddition($comment, $table);
```

framework/db/mssql/QueryBuilder.php
```diff
@@ -265,7 +300,35 @@ public function addCommentOnColumn($table, $column, $comment)
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function mountCommonSqlForCommentRemoval($table, $column = null)
+    {
+        $tableSchema = $this->db->schema->getTableSchema($table);
+
+        $schemaName = $tableSchema->schemaName ? "N'" . $tableSchema->schemaName . "'": 'SCHEMA_NAME()';
```
samdark: Should we quote names here?

samdark (changes requested):

framework/db/mssql/QueryBuilder.php
sdlins: We shouldn't. `fn_listextendedproperty`, `sp_addextendedproperty` and `sp_updateextendedproperty` does not accept quoted names, even althoug this is unbeliveable. I tried it in many different ways/combinations of quotations right in mssql and each trying results in error. This is one of the reasons the code was not working before.

framework/db/mssql/QueryBuilder.php
sdlins: We shouldn't. Same reason.

framework/db/mssql/QueryBuilder.php
sdlins: We shouldn't. Same reason.

framework/db/mssql/QueryBuilder.php
sdlins: Ok.

framework/db/mssql/QueryBuilder.php
sdlins: We shouldn't. Same reason.

framework/db/mssql/QueryBuilder.php
sdlins: No, copy/paste other function. Fixing.

framework/db/mssql/QueryBuilder.php
samdark: That's a bit weird. What would happen if you'll use `'` in a name?

sdlins: I think this happens because those "methods" are, in fact, mssql functions. and as such, they have limitations in what will be received by its params. Function params are only datatypes (varchar, int, etc). But just in case, I will take a bit more time to show you some screenshots and search any workarounds for it on the web. I think this have some impact for mssql users and so it is really necessary to take double care.

sdlins: @samdark, not even `getTableSchema()` works when using a table having a quote in its name:


As you can see in that debug screenshot, the table exists.

We should fix `mssql\Schema::findColumns()` line 394 `$whereSql .= " AND [t1].[table_schema] = '{$table->schemaName}'";` to quote table name.

But quotation really can not be done for `sp_listextendedproperty` or for its alternative, a query in sys.* objects, because both need a varchar data. If we use `[tbl' testes]` or `'[tbl'' testes]'` instead of `tbl'' testes` we get errors:



Using just escaped name works fine:

sdlins: For now, I only can see two options:

1. (Infeasible for Yii 2) Change the design of db objects' interations to make `createCommand()` use `bindParams()`, so that the own PDO would escape those names when necessary; OR

2. Just escape single quotes using `str_replace("'", "''", "tbl' testes")` or something like because mssql use two single quotes to generate one escaped single quote.

I think it would be valid for the most of the situations, but I feel it would break for some complex table names (or columns). Since right now the code breaks for *all* those situations, I can see this not perfect solution as a very enough improvement. 

Let me know what you think.

samdark: Going second way is reasonable.

framework/db/mssql/QueryBuilder.php
```diff
@@ -250,13 +250,45 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
         return $command;
     }
 
+    protected function mountCommonSqlForCommentAddition($comment, $table, $column = null)
```
samdark:
```suggestion
    protected function buildAddCommentSql($comment, $table, $column = null)
```

samdark (changes requested):

sdlins:
@samdark, finished. It is already rebased against master. All tests for `mssql\QueryBuilderTest` and `mssql\SchemaTest` that already was passing in master continue passing in this branch too.

framework/db/mssql/QueryBuilder.php
```diff
@@ -250,13 +250,45 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
...
+                    FROM fn_listextendedproperty (
+                        N'MS_description',
+                        'SCHEMA', $schemaName,
+                        'TABLE', $tableName,
+                        " . ($column ? "'COLUMN', $columnName " : ' DEFAULT, DEFAULT ') . "
+                    )
+            )
+                EXEC sys.sp_addextendedproperty $functionParams
+            ELSE
+                EXEC sys.sp_updateextendedproperty $functionParams
```
alexkart: What is the purpose of having sp_updateextendedproperty? If you need to update comment you can always delete old comment and add new one. Add and delete is enough for all use cases. No need for update I think.

framework/db/mssql/QueryBuilder.php
```diff
@@ -265,7 +297,35 @@ public function addCommentOnColumn($table, $column, $comment)
...
+            IF EXISTS (
+                    SELECT 1
+                    FROM fn_listextendedproperty (
+                        N'MS_description',
+                        'SCHEMA', $schemaName,
+                        'TABLE', $tableName,
+                        " . ($column ? "'COLUMN', $columnName " : ' DEFAULT, DEFAULT ') . "
+                    )
+            )
+                EXEC sys.sp_dropextendedproperty
```
alexkart: Does this fail without checking if comment exists?

framework/db/mssql/QueryBuilder.php
alexkart: AS far as I know it works with quoted table names see tests from this PR https://github.com/yiisoft/yii2/pull/17413/files
@sdlins It appears that I fixed the same issues as you. Could you review my PR and leave you comments as we need to merge them somehow either accept your changes or mine.

framework/db/mssql/QueryBuilder.php
```diff
@@ -250,13 +250,45 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
         return $command;
     }
 
+    protected function buildAddCommentSql($comment, $table, $column = null)
+    {
+        $tableSchema = $this->db->schema->getTableSchema($table);
```
alexkart: Why can’t we just use `Schema::defaultSchema` here?

framework/db/mssql/QueryBuilder.php
```diff
@@ -250,13 +250,45 @@ public function checkIntegrity($check = true, $schema = '', $table = '')
         return $command;
     }
 
+    protected function buildAddCommentSql($comment, $table, $column = null)
```
samdark: Need phpdoc and `@since` in it.

samdark (changes requested):

tests/framework/db/mssql/QueryBuilderTest.php
```diff
@@ -67,36 +67,178 @@ public function testOffset()
...
+        $columnComment = 'A comment for stranger \'field in stranger \'table.';
+        $this->runAddCommentOnColumn($columnComment, $table, $column);
+        $this->runDropCommentFromColumn($table, $column);
+        $result = $this->getCommentsFromColumn($table, $column);
+        $this->assertEquals([], $result);
+    }
+
+    public function testCommentColumn()
+    {
+        $this->markTestSkipped("Testing the behavior, not sql generation anymore.");
```
alexkart: What’s wrong with testing generated SQL? It’s faster than running those queries.

framework/db/mssql/QueryBuilder.php
sdlins: I can revert the question to you and we are at the same point. So checking and updating or droping and inserting we will have at least two commands anyway. But since we can't drop without check before because it fails, we would need 3 queries: check, drop and insert. Or maybe I just could not see what is your point with this question, sorry.

framework/db/mssql/QueryBuilder.php
sdlins: Because the point here is not the schema, but to get the table name *without* schema (line 258). And since I had the table schema I took advantage to use it for schema too.

framework/db/mssql/QueryBuilder.php
sdlins: Yeah, it fails badly!

tests/framework/db/mssql/QueryBuilderTest.php
sdlins: 1) Originally we were testing the sql, the tests were passing but the comments was not working. Yeah, it was a test error, but just in case...
2) In dealing with some MS products that have theirs own ~~strange~~ "particular" way to do the things, It seems safer to test the real thing happening instead of only sql code. By the way, the additional time seems derisory.

sdlins:
@samdark and @alexkart, I think all questions were answered. Let me know If I have forgotten anyone. I will wait you return again with *all* items you think should be changed so that I can make the final code.

samdark:
@sdlins re-reviewed. No more comments from me.

alexkart:
@sdlins what about my comment about quoting? Schema name, table name etc. should be quoted with brackets (‘[‘, ‘]’) as those names can contain dot. Also could write test that adds comment to such table, there should be the table with dots in the name in the db dump.

sdlins:
@alexkart please, see my explanation here https://github.com/yiisoft/yii2/pull/17364#discussion_r298458104.

And no, points will not make sp_* functions to fail, they work fine (tested here with `table.with.point`), but quotes (') in object names would make our query invalid. 

`sp_add...` and `sp_drop...` functions works with quoted or string object names (example: [valid name], 'valid name') but `fn_listextendedproperty` or its alternatives like queries joining `sys.*` tables *does not*. As I explained before, we *really* need to check comment existence before drop (and even before add) because sqlserver will fail otherwise. So, it would not make difference to use `quoteTableName()` only for add and drop when we can't use it for list. See the names I used for tables in my screenshots pointed in the link above and you will see spaces, quotes, brackets, etc in object names without problems.

Please, let me know if you have more questions.

alexkart:
What about PHPDoc and some description of the following methods `\yii\db\mssql\QueryBuilder::buildAddCommentSql`, `\yii\db\mssql\QueryBuilder::buildRemoveCommentSql`?
Other than that, no more comments.

sdlins:
No problem! I will check that soon. Thank you for your insights. :+1:

sdlins:
Will be done till this weekend.

sdlins:
@samdark @alexkart

tests/framework/db/mssql/QueryBuilderTest.php
```diff
@@ -68,36 +68,178 @@ public function testOffset()
...
+    }
+
+    public function testCommentTable()
```
alexkart: Can be removed, I think. @samdark ?

samdark: Then parent test from `yiiunit\framework\db\QueryBuilderTest` execution won't be prevented. Isn't that the intention of having this method?

samdark (approved):

tests/framework/db/mssql/QueryBuilderTest.php
alexkart: Right, yes.

samdark:
Merged. Thank you!

sdlins:
It's always a pleasure to contribute to Yii ecosystem :v:

PR #17381, opened 2019-06-18
Add support for closure controller actions

brandonkelly:
We have a use case where it would be nice if controller actions could be defined by closures from `Controller::actions()`, and the closures could end up getting stored in `InlineAction` instances, the same way validation closures get stored in `InlineValidator` instances.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes


samdark:
While the idea is fine and already available for 3.0, I don't think it is a good idea to introduce enhancements  to 2.0 at this point.

brandonkelly:
Alright no worries, I’ll just add y own InlineAction-type of class.

PR #17390, opened 2019-06-21
Closes #17389. Fix issue in \yii\validators\UniqueValidator

### LINKED ISSUE #17389: UniqueValidator attempts to populate relations when using joinWith
Add a `joinWith` to model `find()` method that has a unique validator.
Trigger the unique validator by saving or the like
Model saves
### What do you get instead?
Query from unique validator attempts to populate relation defined in joinWith and triggers exception with message `Undefined index: `

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.21
| PHP version      | 7.1.26
| Operating system | Alpine 3.8

garthpmurray:
Set with and joinwith to null to prevent relation loading

Signed-off-by: Garth Murray <garthpmurray@gmail.com>

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17389


framework/validators/UniqueValidator.php
```diff
@@ -194,6 +194,15 @@ private function modelExists($targetClass, $conditions, $model)
                 
                 // any with relation can't be loaded because related fields are not selected
                 $query->with = null;
+    
+                if (is_array($query->joinWith)) {
```
samdark: Is there a need to take non-eager-loaded `joinWith` or `with` into account when building a query for uniqueness validation at all?

framework/validators/UniqueValidator.php
```diff
@@ -194,6 +194,15 @@ private function modelExists($targetClass, $conditions, $model)
...
+                    // any joinWiths need to have eagerLoading turned off to prevent related fields being loaded
+                    foreach ($query->joinWith as &$joinWith) {
```
samdark: Would setting `$query->joinWith` to `null` work?

samdark (changes requested):

framework/validators/UniqueValidator.php
garthpmurray: My use case requires a external table holding an account id, I join on this to prevent unique collision within an account but allow duplicates within the table structure itself.

framework/validators/UniqueValidator.php
garthpmurray: Structurally yes, my original push did just that.

However, I think that would produce unexpected results if one was added since `joinWith` modifies the original `find()` query.

samdark (approved):

samdark:
Merged. Thank you!

PR #17396, opened 2019-06-26
Update Action.php. Add 'invoked by ' to the debug message.

alexkart:
Add 'invoked by <controller>' to the debug message when `\yii\base\Action` is used.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

When `InlineAction` is used debug message looks like this:
```
Running action: User\Account\App\Controller\InfoController::actionPhoto()
```
But for `\yii\base\Action` it logs only action class name and doesn't log controller class name, it's useful to see the controller name which invoked the action like this:

```
Running action: App\Domain\Web\Actions\ListRecords::run(), invoked by Report\App\Controller\Frontend\LibraryController
```


samdark (approved):

samdark:
Merged. Thanks!

PR #17408, opened 2019-07-02
Closes #17057. MSSQL quoted tables.

### LINKED ISSUE #17057: Dots in the table names

Set up an active record for a table contains dots in the name (Yes, ms sqlserver table names can has a dots like this: [SRV_NAME].[DB_NAME].[SCHEMA_NAME].[TABLE.NAME.WITH.A.LOT.OF.DOTS.]). 

Nothing too bad.

### What do you get instead?

Database Exception.

resolveTableNames method in yii\db\mssql\Schema treats every dot like qualifier delimiter.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      |2.0.15.1
| PHP version      | 7.2
| Operating system |Win srv 2012 r2

alexkart:
Fix issues with table names that contain special characters or keywords in MSSQL.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #17057

### DIFF
--- a/framework/db/Schema.php
+++ b/framework/db/Schema.php
@@ -494,7 +494,7 @@ abstract class Schema extends BaseObject
 
     /**
      * Splits full table name into parts
-     * @param $name string
+     * @param string $name
      * @return array
      * @since 2.0.22
      */
--- a/framework/db/Schema.php
+++ b/framework/db/Schema.php
@@ -494,8 +494,9 @@ abstract class Schema extends BaseObject
 
     /**
      * Splits full table name into parts
-     * @param $name
+     * @param $name string
      * @return array
+     * @since 2.0.22
      */
     protected function getTableNameParts($name)
     {
--- a/framework/db/Schema.php
+++ b/framework/db/Schema.php
@@ -15,7 +15,6 @@ use yii\base\NotSupportedException;
 use yii\caching\Cache;
 use yii\caching\CacheInterface;
 use yii\caching\TagDependency;
-use yii\helpers\StringHelper;
 
 /**
  * Schema is the base class for concrete DBMS-specific schema classes.
@@ -485,12 +484,22 @@ abstract class Schema extends BaseObject
         if (strpos($name, '.') === false) {
             return $this->quoteSimpleTableName($name);
         }
-        $parts = explode('.', $name);
+        $parts = $this->getTableNameParts($name);
         foreach ($parts as $i => $part) {
             $parts[$i] = $this->quoteSimpleTableName($part);
         }
 
         return implode('.', $parts);
+    }
+
+    /**
+     * Splits full table name into parts
+     * @param $name
+     * @return array
+     */
+    protected function getTableNameParts($name)
+    {
+        return explode('.', $name);
     }
 
     /**
@@ -661,7 +670,7 @@ abstract class Schema extends BaseObject
         }
         $message = $e->getMessage() . "\nThe SQL being executed was: $rawSql";
         $errorInfo = $e instanceof \PDOException ? $e->errorInfo : null;
-        return new $exceptionClass($message, $errorInfo, (int) $e->getCode(), $e);
+        return new $exceptionClass($message, $errorInfo, (int)$e->getCode(), $e);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


framework/db/Schema.php
```diff
@@ -485,14 +484,24 @@ public function quoteTableName($name)
...
         foreach ($parts as $i => $part) {
             $parts[$i] = $this->quoteSimpleTableName($part);
         }
 
         return implode('.', $parts);
     }
 
+    /**
+     * Splits full table name into parts
+     * @param $name
```
samdark:
```suggestion
     * @param string $name
```

framework/db/Schema.php
```diff
@@ -485,14 +484,24 @@ public function quoteTableName($name)
...
+     * @return array
```
samdark: Need `@since`.

framework/db/mssql/Schema.php
```diff
@@ -126,6 +126,24 @@ protected function resolveTableName($name)
         return $resolvedName;
     }
 
+    /**
+     * {@inheritDoc}
+     * @param $name
```
samdark:
```suggestion
     * @param string $name
```

framework/db/mssql/Schema.php
```diff
@@ -126,6 +126,24 @@ protected function resolveTableName($name)
...
+     * @return array|mixed
```
samdark: Why would it return mixed?

framework/db/mssql/Schema.php
```diff
@@ -126,6 +126,24 @@ protected function resolveTableName($name)
...
+     */
+    protected function getTableNameParts($name)
+    {
+        $parts = [$name];
+        preg_match_all('/([^.\[\]]+)|\[([^\[\]]+)\]/', $name, $matches);
+        if (isset($matches[0]) && is_array($matches[0]) && !empty($matches[0])) {
+            $parts = $matches[0];
+        }
+
+        $parts = str_replace(['[', ']'], '', $parts);
```
samdark: If the `if` above we assume that `$parts` is an array. Then we're doing `str_replace` on it. That would cause error.

samdark (changes requested):

framework/db/mssql/Schema.php
alexkart: Why? `str_replace` works with arrays.

samdark: You're right. PHP standard library is a mess :)

framework/db/mssql/Schema.php
```diff
@@ -126,6 +126,24 @@ protected function resolveTableName($name)
         return $resolvedName;
     }
 
+    /**
+     * {@inheritDoc}
+     * @param $name string
+     * @return array
+     */
+    protected function getTableNameParts($name)
```
samdark: Need `@since`.

samdark (approved):
Other than missing `@since` tag it's all good. Great job!

framework/db/Schema.php
```diff
@@ -494,8 +494,9 @@ public function quoteTableName($name)
 
     /**
      * Splits full table name into parts
-     * @param $name
+     * @param $name string
```
Thoulah: `@param string $name`
https://docs.phpdoc.org/references/phpdoc/tags/param.html

framework/db/mssql/Schema.php
```diff
@@ -128,8 +128,8 @@ protected function resolveTableName($name)
 
     /**
      * {@inheritDoc}
-     * @param $name
-     * @return array|mixed
+     * @param $name string
```
Thoulah: `@param string $name`

samdark:
Merged. Thank you!

alexkart:
@samdark are `framework/db/mssql/` tests run on CI? I've found one more place where we need to handle table names with dots. But this case should be caught by this test `framework\db\mssql\SchemaTest:testGetTableSchemas`.

samdark:
I think they currently aren't run.

alexkart:
> I think they currently aren't run.

I see. BTW, there are lots of broken MSSQL tests.

samdark:
Yes :( If you have ideas on how to run these via travis, let me know.

alexkart:
Something like this? https://github.com/biolab/orange3/commit/ea8205a20acd6a74dfd7c829abed7f0bd00c7184

samdark:
Yes, that could work. Does travis allow running non-docker and docker in the same build?

PR #17413, opened 2019-07-05
MSSQL: fix tests

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #15880

PR #17418, opened 2019-07-09
MSSQL: fix tests

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17413, #17418

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.23 under development
 ------------------------
 
-- Bug #17413: Fixed MSSQL tests: `QueryBuilderTest::testGetColumnType()`, `QueryBuilderTest::testResetSequence()` (alexkart)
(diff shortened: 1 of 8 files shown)


alexkart:
Trying to fix `\yiiunit\framework\db\ActiveRecordTest::testDefaultValues`.
But the question is should we fix the test to match current implementation which is currently incorrect or fix implementation? If we fix implementation it may break the existing code as someone could rely on it using incorrect default values.
The issue is that default values are stored in such format in MSSQL `('value')`.
af3a717d124f4a046795bc0d11dfd2a3

Has the format changed since the test was written or it was incorrect from the beginning?

samdark:
The test may have been incorrect .

alexkart:
OK, I'll move this change to a separate pull request then.

samdark:
Merged. Thank you!

PR #17419, opened 2019-07-09
fix to call static method of an interface defined in DI Container

albertborsos:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | 

Context:
I use the DI container to use ARs through repositories. Every AR related methods are exists only in the repositories.

In a form model, i would like to validate a uniqueness of an attribute:

```php
    public function rules()
    {
        return [
            [['name'], 'unique', 'targetClass' => \app\models\mysql\Singer::class, 'targetAttribute' => 'name'],
        ];
    }
```
But I do not want to burn in the mysql AR class to my rules, so I would like to use the `dataModelClass` interface method of the repository which is in use:

```php
    public function rules()
    {
        return [
            [['name'], 'unique', 'targetClass' => Yii::createObject([SingerRepositoryInterface::class, 'dataModelName']), 'targetAttribute' => 'name'],
        ];
    }
```

Now it is only possible with a hack:

```php
    public function rules()
    {
        $repository = \Yii::createObject(SingerRepositoryInterface::class);
        $targetClass = call_user_func([get_class($repository), 'dataModelClass']);
        return [
            [['name'], 'unique', 'targetClass' => $targetClass, 'targetAttribute' => 'name'],
        ];
    }
```

But I extended the `BaseYii::createObject()` method, so now it is working as the desired way.

What do you think?


samdark:
Do you have generic interface for all your repositories? It looks like you can do `SingerRepository::dataModelClass()` directly.

samdark:
https://travis-ci.org/yiisoft/yii2/jobs/556359451#L1221

albertborsos:
Of course I use a common `RepositoryInterface` I just did not want to complicate the example.
Btw here are my repositories: https://github.com/albertborsos/yii2-ddd/tree/repository/src/repositories.

samdark:
Are there multiple implementations for SingerRepositoryInterface?

albertborsos:
Yes, there is. I have different implementations for different databases. For example I have to write different query expressions for mysql and pgsql queries.

samdark:
Discussed this one internally in the core team and decided not to have it. The use case is quite specific while it touches very fundamental part of the framework.

samdark:
Thanks for contributing anyway.

PR #17426, opened 2019-07-11
MSSQL: fix tests

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |


tests/framework/db/CommandTest.php
```diff
@@ -472,7 +472,11 @@ public function testNoTablenameReplacement()
                 'address' => 'Some {{%weird}} address',
             ]
         )->execute();
-        $customerId = $db->getLastInsertID();
+        if ($this->driverName === 'pgsql') {
```
samdark: How did it work before?

alexkart: It inserted id explicitly but MSSQL doesn't allow inserting of identity columns.

samdark: How's that connected to pgsql? It was called without arguments before and tests for PostgreSQL passed.

alexkart: No, previously `'id' => 43,` was inserted and then used in select queries. I've removed `id` from insert statement and added `$db->getLastInsertID()`.

samdark: Ah. OK then.

samdark:
Merged. Thank you very much!

PR #17430, opened 2019-07-12
Added type null to @return annotation of IdentityInterface::findIdentity* methods

k-timoshenko:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌


samdark:
Merged. Thanks!

PR #17433, opened 2019-07-15
Closes #10023. Fix MSSQL "There are no more rows" exception

### LINKED ISSUE #10023: Query each method does not recognize if there are no more rows
Like mentioned in: #9962 there is a problem with the recognition of the end of a resultset.

Example:
There is a table with 10 entries

``` php
foreach ($query->each() as $key => $value) {
    echo $key."\n";
}
```

after it selects the 10th row it throws an exception:

```
Exception 'PDOException' with message 'SQLSTATE[IMSSP]:
There are no more rows in the active result set.
Since this result set is not scrollable, no more data may be retrieved.'

in vendor\yiisoft\yii2\db\DataReader.php:111

Stack trace:
#0 vendor\yiisoft\yii2\db\DataReader.php(111): PDOStatement->fetch()
#1 vendor\yiisoft\yii2\db\BatchQueryResult.php(143): yii\db\DataReader->read()
#2 vendor\yiisoft\yii2\db\BatchQueryResult.php(112): yii\db\BatchQueryResult->fetchData()
#3 commands\TestController.php(26): yii\db\BatchQueryResult->next()
```

alexkart:
when using `each()` and `batch()`

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #10023

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -7,6 +7,7 @@ Yii Framework 2 Change Log
 - Bug #17413, #17418, #17426, #17431: Fixed MSSQL tests (alexkart)
 - Bug #17420: Fixed loading of column default values for MSSQL (alexkart)
 - Bug #17395: Fixed issues with actions that contain underscores in their names (alexkart)
+- Bug #10023: Fixed MSSQL "There are no more rows in the active result set" exception when using `each()` and `batch()` (alexkart)
 
 
 2.0.22 July 02, 2019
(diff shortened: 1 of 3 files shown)


alexkart:
Should I fix codeclimate issue?

samdark:
If you can. That's not mandatory in this case.

samdark (approved):

samdark:
Merged. Thank you!

PR #17436, opened 2019-07-16
Container::get() merge by ArrayHelper::merge()

DrDeath72:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #14821

### DIFF
--- a/framework/di/Container.php
+++ b/framework/di/Container.php
@@ -121,7 +121,11 @@ class Container extends Component
      */
     private $_dependencies = [];
 
-
+    /*
+     * @var boolean use ArrayHelper::merge() instead array_merge when merge class definition
+     */
+    public $mergeDefinitionRecursive = false;
+    
     /**
      * Returns an instance of the requested class.
      *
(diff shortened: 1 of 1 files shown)


samdark:
But that would cause other issues described in the original issue if enabled.

samdark:
https://github.com/yiisoft/yii2/issues/14821#issuecomment-329738899

samdark:
i.e. replacing values won't work.

DrDeath72:
> But that would cause other issues described in the original issue if enabled.
which one?

DrDeath72:
> i.e. replacing values won't work.

definition can't replace value from config?

samdark:
See the test from the comment I've referenced to above.

DrDeath72:
> See the test from the comment I've referenced above.

ArrayHelper::merge don't make array when key exists as array_merge_recursive

terabytesoftw:
Closing this PR due to long inactivity and because the proposed behavior conflicts with the documented DI container semantics: runtime config must override container definitions, and arrays are replaced rather than recursively merged.

The linked issue #14821 is already closed, and this PR would introduce new public API and special definition keys without tests, documentation, or upgrade notes. A new PR can still be opened if there is a clearly documented, fully tested, backward-compatible opt-in design.

PR #17447, opened 2019-07-18
Closes #17219. MSSQL: fix quoting of table names with spaces

### LINKED ISSUE #17219: Incorrect table name(extra square brackets injected) when using SQL server table with two nonconsecutive spaces after overriding the tablename function
1)Create a table in sql server named [PPPPPPPP$Internet Customer Cart] with an id column.
2)Create a table in SQL server named [PPPPPPPP$Order] with an id column
3)Create a join table in SQL server named [PPPPPPPP$AtoB] with columns
cartId
orderId
4)Create a model named CartItem
5)Create a model named Order

6)Override the CartItem tablename function like so:
    {
		return '[PARTWERX$Internet Customer Cart]'; 
    }

7)Try to create a getItems function in the Order table so that you can write
$order->items. It is impossible because of the bug. You will see something like:

SQLSTATE[42000]: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]Incorrect syntax near the keyword 'IN'.
The SQL being executed was: SELECT * FROM [PPPPPPP$Internet Customer [Cart]] WHERE [id] IN ('....', '...')

Notice the extra square brackets around the Cart. There should only be one square bracket around the entire table name.

I expect that no extra square brackets appear.

### What do you get instead?
Extra square brackets appear.

### Additional info
Tested in latest version 2.0.16
| Q                | A
| ---------------- | ---
| Yii version      | 2.0.?
| PHP version      | 
| Operating system |

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17219

### DIFF
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -1739,7 +1739,7 @@ class QueryBuilder extends \yii\base\BaseObject
      * @param $table
      * @return bool|array
      */
-    private function extractAlias($table)
+    protected function extractAlias($table)
     {
         if (preg_match('/^(.*?)(?i:\s+as|)\s+([^ ]+)$/', $table, $matches)) {
             return $matches;
--- a/framework/db/QueryBuilder.php
+++ b/framework/db/QueryBuilder.php
@@ -1738,6 +1738,7 @@ class QueryBuilder extends \yii\base\BaseObject
      * Extracts table alias if there is one or returns false
      * @param $table
      * @return bool|array
+     * @since 2.0.24
      */
     protected function extractAlias($table)
     {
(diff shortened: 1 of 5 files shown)


framework/db/QueryBuilder.php
```diff
@@ -1733,4 +1733,18 @@ public function bindParam($value, &$params)
 
         return $phName;
     }
+
+    /**
+     * Extracts table alias if there is one or returns false
+     * @param $table
+     * @return bool|array
+     */
+    protected function extractAlias($table)
```
samdark: Need "since" tag because method visibility is `protected` thus it can be used now.

samdark (approved):

samdark:
Merged. Thank you!

PR #17452, opened 2019-07-20
add another condition to isEmpty

undefinedor:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌

An undefined variable should means empty  in the method named isEmpty of QueryTrait

### DIFF
--- a/framework/db/QueryTrait.php
+++ b/framework/db/QueryTrait.php
@@ -290,6 +290,7 @@ trait QueryTrait
      *
      * The value is considered "empty", if one of the following conditions is satisfied:
      *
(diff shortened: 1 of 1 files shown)


undefinedor:
this pr is to sloved the following case :

```php

$query = new \yii\db\Query();

$query->andFilterWhere([
    'myCondition'=>$myCondtionValue //  This undefined variable cause a warning
]);
```

undefinedor:
my mistake

Issue #17437, opened 2019-07-16
Migration templates do not work with namespace naming convention


Isitar:
### What steps will reproduce the problem?
`php yii migrate/create 'common\migrations\CreateNewsTable' --fields="hello:int"`
### What is the expected result?
a migration with the create_table template
### What do you get instead?
an empty migration

### Additional info
it doesnt work because the preg_match does not work for this naming convention (line 392 onwards in MigrateController)

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.22
| PHP version      | 7.3 
| Operating system | Windows

bizley:
I'm on it.

Isitar:
@bizley do not forget the template files itself.

PR #17453, opened 2019-07-20
Making migration generator generate namespaced migrations

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17437 

This allows generator to understand camelcased commands.

1. Namespaced class name is generated according to namespaced format `M<YYMMDDHHMMSS><Name>` where `Name` is camelized version of what is provided for generator.
2. Table names provided for generator are underscored in migration (so `User` is `user`, `MyData` is `my_data`, and `BOB` is `b_o_b` unless the name is preceded with underscore (for BC) then name case is kept as-is.
3. If table name should be suffixed with underscore you must use two underscores for generator.
4. Unintentional change but I don't want to fight it - now you can run generator with any mix of cases and/or underscores separating command words like `create_userTable`, `CreateUsertabLE` and so on. In guide I'm using official format every time.
5. Guide is updated for this (also PL version is added).

### DIFF
--- a/framework/console/controllers/MigrateController.php
+++ b/framework/console/controllers/MigrateController.php
@@ -13,6 +13,7 @@ use yii\db\Query;
 use yii\di\Instance;
 use yii\helpers\ArrayHelper;
 use yii\helpers\Console;
+use yii\helpers\Inflector;
 
 /**
  * Manages application migrations.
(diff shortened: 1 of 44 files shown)


samdark:
Tests do not pass.

bizley:
Fixed now ;)

framework/console/controllers/MigrateController.php
```diff
@@ -375,6 +376,26 @@ protected function getMigrationNameLimit()
...
     }
 
+    /**
+     * Normalizes table name for generator.
+     * When name is preceded with underscore name case is kept - otherwise it's converted from camelcase to underscored.
+     * Last underscore is always trimmed so if there should be underscore at the end of name use two of them.
+     * @param string $name
+     * @return string
+     */
+    private function normalizedTableName($name)
```
samdark:
```suggestion
    private function normalizeTableName($name)
```

samdark (approved):

samdark:
All good except minor method naming issue.

framework/console/controllers/MigrateController.php
bizley: Right, thanks! Commited as suggested

samdark:
Merged. Thank you!

PR #17471, opened 2019-07-29
ArrayHelper check if object key exist

upiksaleh:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ❌


samdark:
It is expected to be as it is:

> // this is expected to fail if the property does not exist, or __get() is not implemented
> // it is not reliably possible to check whether a property is accessible beforehand

PR #17477, opened 2019-07-30
Fix #17469: Fixed updating `Yii` logger instance when setting new logger via configuration

### LINKED ISSUE #17469: logger not working if it's set from config
Inside any controller, put this:
```php
public function actionIndex()
{
        Yii::info('Hello World');
        return $this->render('index');
}
```

inside the config:
```php
'log' => [
            'traceLevel' => YII_DEBUG ? 3 : 0,
            'logger' => Yii::createObject('yii\log\Logger'),
            'targets' => [
                [
                    'class' => 'yii\log\FileTarget',
                    'levels' => ['error', 'warning', 'info'],
                ],
            ],
],
```

Log inside the `runtime/logs/app.log`

### What do you get instead?
Nothing in `runtime/logs/app.log`

### Additional info
It'll work if  `'logger' => Yii::createObject('yii\log\Logger'),` is removed

I have found location from where it's misbehaving.
**Class** : `yii\log\Logger` 
```php
public function __construct($config = [])
    {
        // ensure logger gets set before any other config option
        if (isset($config['logger'])) {
            $this->setLogger($config['logger']);
            unset($config['logger']);
        }
        // connect logger and dispatcher
        $this->getLogger();

        parent::__construct($config);
    }

```
and
```php
/**
     * Gets the connected logger.
     * If not set, [[\Yii::getLogger()]] will be used.
     * @property Logger the logger. If not set, [[\Yii::getLogger()]] will be used.
     * @return Logger the logger.
     */
    public function getLogger()
    {
        if ($this->_logger === null) {
            $this->setLogger(Yii::getLogger());
        }

        return $this->_logger;
    }

    /**
     * Sets the connected logger.
     * @param Logger|string|array $value the logger to be used. This can either be a logger instance
     * or a configuration that will be used to create one using [[Yii::createObject()]].
     */
    public function setLogger($value)
    {
        if (is_string($value) || is_array($value)) {
            $value = Yii::createObject($value);
        }
        $this->_logger = $value;
        $this->_logger->dispatcher = $this;
    }
```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.23
| PHP version      | 7.3.7-1+ubuntu16.04.1+deb.sury.org+1
| Operating system | ubuntu16.04.1+deb.sury.org+1

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17469

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -11,6 +11,7 @@ Yii Framework 2 Change Log
 - Bug #16796: Fixed addition and removal of table and column comments in MSSQL (sdlins)
 - Bug #17449: Fixed order of SQL column build syntax for MySQL migration (choken)
 - Bug #17437: Fixed generating namespaced migrations (bizley)
+- Bug #17469: Fixed updating `Yii` logger instance when setting new logger via configuration (samdark)
 
 
 2.0.23 July 16, 2019
--- a/framework/log/Dispatcher.php
+++ b/framework/log/Dispatcher.php
@@ -131,6 +131,7 @@ class Dispatcher extends Component
         }
         $this->_logger = $value;
         $this->_logger->dispatcher = $this;
+        Yii::setLogger($this->_logger);
     }
 
     /**
--- a/tests/framework/log/DispatcherTest.php
+++ b/tests/framework/log/DispatcherTest.php
@@ -88,6 +88,16 @@ namespace yiiunit\framework\log {
             $this->assertEquals(42, $dispatcher->getLogger()->traceLevel);
         }
 
+        public function testSetNewLoggerUpdatesGlobalLogger()
+        {
+            $dispatcher = new Dispatcher();
+            $this->assertSame(Yii::getLogger(), $dispatcher->getLogger());
+
+            $newLogger = new Logger();
+            $dispatcher->setLogger($newLogger);
+            $this->assertSame(Yii::getLogger(), $dispatcher->getLogger());
+        }
+
         /**
          * @covers \yii\log\Dispatcher::setLogger()
          */


rob006:
I have mixed feelings about this. It looks like ugly voodoo magic that assumes that you always have only one dispatcher and it is `Yii::$app->log` (which is really naive assumption). Shouldn't we at list check if `$this === Yii::$app->log` before implicitly assigning static global logger.

samdark:
That will return `false` most of the time since logger is initialized early and then cached statically forever. That is indeed, a bit of voodoo magic but that is what service locator brings us :(

rob006:
> a bit of voodoo magic but that is what service locator brings us :(

Not it's not. This is the result of ugly shortcut which creates twisted relation between `Yii::getLogger()` and `Yii::$app->log->logger`. This PR just make it way more magic (by introducing BC break BTW - this is really ugly and unexpected side effect). You could have service locator without assuming that one class could be instantiated only once.

samdark:
Yes but in this case it was already working that way (when getting it) but without the ability to configure it properly. That part was fixed.

rob006:
There is a difference between using `Yii::getLogger()` as a fallback, and between implicitly setting global logger when you use setter on **local object**. Now if you use in your code `new Dispatcher(['logger' => $myFakeLogger]);` it will magically break your app by overriding global logger.

samdark:
Right. Is there a better way to solve original issue?

rob006:
I'm not sure. But I would rather do not touch it at all than introduce BC break that makes this part of framework even more magic. This issue is known from at least 3 years (#12163) with known (`Yii::setLogger(new MyLogger())` before app initialization) and consistent (you should use `Yii::setLogger()` to change logger returned by `Yii::getLogger()`) workaround, I'm not sure it is a good time to make such changes.

schmunk42:
I agree with @rob006, please no BC breaks in 2.x

samdark:
OK. Will be reverted next release.

PR #17478, opened 2019-07-30
Fixed docs for Module::createController method

yuniorsk:
Fixes outdated documentation of given method. Steps 2 and 3 were switched.


samdark:
Good catch. Thank you!

Issue #8225, opened 2015-04-29
Ajax Validation with checkboxList only triggered on first select


thejahweh:
Hi I have to check that in a checkboxList are min 2 and max 3 checkboxes selected. I had written a simple inline validator for that case, which is not the problem. The Ajax validation is enabled and works for the other fields. If I select the first checkbox in the checkboxList it sends the ajax request and gives me the error, that I have to select min 2 and max 3 checkboxes, which is not that nice but also not really a problem.

The problem is that if the next checkbox is selected nothing happens, it does not send any ajax request and the error message doesn't disappear.

thejahweh:
I have located the problem at `yii.activeForm.js` in function `getValue` at line ~596.

replace 
```javascript
            if (!$realInput.length) {
                $realInput = $form.find('input[type=hidden][name="' + $input.prop('name') + '"]');
            }
```
with
```javascript
            if (!$realInput.length) {
                var vals = [];
                $realInput.each(function(i) {
                    vals.push(i.val());
                })
                return vals;
            }
```

The problem is just because the `getValue` function does not support arrays, but if you can select more than one checkox it is an array which should be supported.

cebe:
thanks for investigation on this.

execut:
@samdark I tried to write a test, but I didn’t have enough js unit testing experience. Please tell me how to test this bug? The getValue method is not covered by tests. Need to do it? After that, add and test this error?

samdark:
Yes.

PR #17480, opened 2019-07-31
Bug #8225: Fixed ajax Validation with checkboxList only on first select

execut:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #8225


execut:
I tried to write a test, but I didn’t have enough jsunit testing experience. It is interesting to see how it should be implemented. I also did not find tests for the method, just to add it for this bug

arzzen (approved):

samdark:
@execut so how about tests?

execut:
> @execut so how about tests?
I will try this week

execut:
> @execut so how about tests?

Done. Is everything right?

samdark (approved):

samdark:
@execut looks fine to me.

samdark:
Merged. Thank you!

PR #17488, opened 2019-08-01
Fix minor typos in fixture guide

mikehaertl:
* Fix typo with `grabFixture()` (remove second argument)
* Use lowercase for filename placeholders to better match real filenames
* Add some missing articles for better readability

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -

### DIFF
--- a/docs/guide/test-fixtures.md
+++ b/docs/guide/test-fixtures.md
@@ -47,8 +47,8 @@ class UserFixture extends ActiveFixture
 > - Elasticsearch: [[yii\elasticsearch\ActiveFixture]] (since version 2.0.2)
 
 
-The fixture data for an `ActiveFixture` fixture is usually provided in a file located at `FixturePath/data/TableName.php`,
-where `FixturePath` stands for the directory containing the fixture class file, and `TableName`
+The fixture data for an `ActiveFixture` fixture is usually provided in a file located at `fixturepath/data/tablename.php`,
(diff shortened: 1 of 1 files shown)


docs/guide/test-fixtures.md
```diff
@@ -154,7 +154,7 @@ You may also assign an alias to a fixture. In the above example, the `UserProfil
 In the test methods, you may then access a fixture object using its alias in `grabFixture()` method. For example,
 
 ```php
-$profile = $I->grabFixture('profiles', 'user1');
+$profile = $I->grabFixture('profiles');
```
samdark: I think it was correct. `user1` is the index in the fixtures array so there we are grabbing a profile for user 1.

samdark (changes requested):

docs/guide/test-fixtures.md
mikehaertl: Well, check the full paragraph:

> For example,
>     $profile = $I->grabFixture('profiles', 'user1');
> will return the UserProfileFixture object.

That's not correct IMO, because you get a `User` instance and not the `UserProfileFixture` itself if you call it with the `user1` argument, don't you?

samdark: Yes, you are correct. Change is good.

samdark:
Merged. Thank you!

PR #17489, opened 2019-08-01
Closes #17486. Fix error when using `batch()` without `$db` parameter…

### LINKED ISSUE #17486: Trying to get property 'driverName' of non-object
```php
$query = MssqlModel::findBySql($sql);
foreach ($query->batch()) {
// some code...
}
```

Iterate all MssqlModel objects.

### What do you get instead?
Exception after first iteration:

```
PHP Notice 'yii\base\ErrorException' with message 'Trying to get property 'driverName' of non-object'

in /app/vendor/yiisoft/yii2/db/BatchQueryResult.php:168
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.24
| PHP version      | 7.4
| Operating system | Debian GNU/Linux

### Additional information
In `/vendor/yiisoft/yii2/db/Query.php:187`: `$db` is null by default and `Query` has no `$db` param to set correct value.
```php
    public function batch($batchSize = 100, $db = null)
    {
        return Yii::createObject([
            'class' => BatchQueryResult::className(),
            'query' => $this,
            'batchSize' => $batchSize,
            'db' => $db,
            'each' => false,
        ]);
    }
```

This is workaround:
```php
foreach ($query->batch(100, MssqlModel::getDb()) {
// some code
}
 ```

alexkart:
… with MSSQL

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17486

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.25 under development
 ------------------------
 
-- no changes in this release.
+- Bug #17486: Fixed error when using `batch()` without `$db` parameter with MSSQL (alexkart)
 
 
 2.0.24 July 30, 2019
--- a/framework/db/BatchQueryResult.php
+++ b/framework/db/BatchQueryResult.php
@@ -151,8 +151,8 @@ class BatchQueryResult extends BaseObject implements \Iterator
 
     /**
      * Reads and collects rows for batch
-     * @since 2.0.23
      * @return array
+     * @since 2.0.23
      */
     protected function getRows()
     {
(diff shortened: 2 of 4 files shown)


samdark (approved):

samdark:
Merged. Thank you!

michalwoz:
@alexkart "if $db is not passed to batch it uses connection from the active record model."

$this->getDbDriverName() still can return null, when you don't use any model:
`
$query = (new Query)->from($tbl->name);
    foreach ($query->each() as $row) {}
`
Why the "db" application component is not used?

workaround is to pass Yii::$app->db to each() function

PR #17501, opened 2019-08-08
fix flushing cache from console

Spell6inder:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | -

When using following application configurations console cache can't find cache component

```php
return [
    //...
    'container' => [
        //...
        'singletons' => [
             //...
            'cache-main' => [
                'class' => \yii\caching\FileCache::class,
            ],
            //...
        ],
        //...
    ],
    //...
    'components' => [
        //...
        'cache' => 'cache-main',
        //...
    ],
    //...
];
```

But in this case, `\yii\web\AssetManager` should be configured...

### DIFF
--- a/framework/console/controllers/CacheController.php
+++ b/framework/console/controllers/CacheController.php
@@ -267,8 +267,10 @@ class CacheController extends Controller
 
             if ($component instanceof CacheInterface) {
                 $caches[$name] = get_class($component);
-            } elseif (is_array($component) && isset($component['class']) && $this->isCacheClass($component['class'])) {
-                $caches[$name] = $component['class'];
+            } elseif (is_array($component) && isset($component['class'])) {
+                if($this->isCacheClass($component['class'])){
+                    $caches[$name] = $component['class'];
+                }
             } elseif (is_string($component) && $this->isCacheClass($component)) {
                 $caches[$name] = $component;
             } elseif (Yii::$app->has($name)) {
--- a/framework/console/controllers/CacheController.php
+++ b/framework/console/controllers/CacheController.php
@@ -271,7 +271,7 @@ class CacheController extends Controller
                 $caches[$name] = $component['class'];
             } elseif (is_string($component) && $this->isCacheClass($component)) {
                 $caches[$name] = $component;
-            } elseif ($component instanceof \Closure) {
+            } elseif (Yii::$app->has($name)) {
                 $cache = Yii::$app->get($name);
                 if ($this->isCacheClass($cache)) {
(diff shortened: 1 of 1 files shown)


framework/console/controllers/CacheController.php
```diff
@@ -267,11 +267,13 @@ private function findCaches(array $cachesNames = [])
 
             if ($component instanceof CacheInterface) {
                 $caches[$name] = get_class($component);
-            } elseif (is_array($component) && isset($component['class']) && $this->isCacheClass($component['class'])) {
-                $caches[$name] = $component['class'];
+            } elseif (is_array($component) && isset($component['class'])) {
```
samdark: That is basically the same as it was before, right?

samdark:
How's that related to AssetManager ?

framework/console/controllers/CacheController.php
Spell6inder: yes, but if `$component['class']` is not cache don't try next case because we know that is exactly not cache

samdark: `&&` already works like that. If condition on the left fails, condition on the right is never executed.

Spell6inder:
> How's that related to AssetManager?

In my case i have frontend, backend and console apps assetManagers for frontend and backend are different. And i not configure assetManager for console, but 'assetManager' is core component. If i try `Yii::$app->get($name)` (`$name = 'assetManager'`) got error because console app havn`t webroot.

framework/console/controllers/CacheController.php
Spell6inder: I add this then got error with AssetManager after previos commit

Spell6inder: if `&& $this->isCacheClass($component['class'])`  in that `if` we got `false` and going to next `if` 
In result can be error on`Yii::$app->get($name)`

samdark:
Seems this change make obtaining component normally from `components` fail: https://travis-ci.org/yiisoft/yii2/jobs/569707906

terabytesoftw:
Thanks for the contribution. I rechecked this against current master. The reported scenario is still valid: CacheController does not discover a cache component configured as a DI alias such as `'cache' => 'cache-main'`.

However, this PR cannot be merged as-is. The current approach may instantiate unrelated non-cache components while scanning, which can break existing configurations/tests, and the PR does not include the required regression tests. Since it has been inactive since 2019, i am closing it.

PR #17516, opened 2019-08-19
fix populates the model with input data

aaronlei:
When you use load() as follows

```php
         $post = Yii::$app->request->post();
         $model = new DealerOrder();
         $model->load($post,'DealerOrder');//1. Specify formName
         $model->load($post);//2. Do not specify formName
         $model->load(['DealerOrder'=>$post]);//3. Use form submission in the browser
         $model->load($post,'');//4. Give formName an empty string
```

You will find that neither the first nor the second can successfully set properties for the model. The third and fourth styles can successfully set properties for the model.

After I modified load(), the first and second styles were also supported.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ❌


samdark:
The change breaks tests: https://travis-ci.org/yiisoft/yii2/jobs/576631041

rob006:
This is serious BC break which may result in loading incorrect data to model. I'm against such magic - there is already `formName()` which could be set to empty string and achieve the same goal at model level.

PR #17518, opened 2019-08-20
Closes #17504. Fix upsert 

### LINKED ISSUE #17504: yii\db\mysql\QueryBuilder::upsert produce invalid SQL on upsert if all columns in table are unique
Lets take a SQL table with 2 columns, which is composite primary key:

```sql
CREATE TABLE `blog_categ` (
  `categ_id` int(10) unsigned NOT NULL,
  `blog_id` int(10) unsigned NOT NULL,
  PRIMARY KEY (`categ_id`,`blog_id`),
)
```
Executing model upsert:
```php
static::getDb()->createCommand()->upsert(static::tableName(), $values, true)->execute();
```
Produce incorrect SQL error with empty UPDATE list:
```sql
INSERT INTO `blog_categ` (`categ_id`, `blog_id`) VALUES (2244, 139) ON DUPLICATE KEY UPDATE
```
The correct SQL will be:
```sql
INSERT INTO `blog_categ` (`categ_id`, `blog_id`) VALUES (2244, 139) ON DUPLICATE KEY UPDATE `categ_id`=`categ_id`
```
If all columns are unique, then QueryBuilder::prepareUpsertColumns return empty list of $updateNames and produce incorrect sql.  So, when $updateNames is empty, the behavior must be the same as when $updateColumns = false.

Small fix to yii\db\mysql\QueryBuilder::upsert($table, $insertColumns, $updateColumns, &$params)
```php
        --- } elseif ($updateColumns === false) {
        +++ } elseif ($updateColumns === false || empty($updateNames)) { 
```
Please, add this fix.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.24
| PHP version      | 
| Operating system |

alexkart:
Fix upsert when `$updateColumns = true` but there is no columns to update in the tale.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17504

### DIFF
--- a/framework/db/cubrid/QueryBuilder.php
+++ b/framework/db/cubrid/QueryBuilder.php
@@ -70,6 +70,10 @@ class QueryBuilder extends \yii\db\QueryBuilder
         if (empty($uniqueNames)) {
             return $this->insert($table, $insertColumns, $params);
         }
+        if ($updateNames === []) {
+            // there is no columns to update
+            $updateColumns = false;
+        }
 
         $onCondition = ['or'];
         $quotedTableName = $this->db->quoteTableName($table);
--- a/framework/db/cubrid/QueryBuilder.php
+++ b/framework/db/cubrid/QueryBuilder.php
@@ -71,7 +71,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
             return $this->insert($table, $insertColumns, $params);
         }
         if ($updateNames === []) {
-            // there is no columns to update
+            // there are no columns to update
             $updateColumns = false;
         }
 
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -454,6 +454,10 @@ class QueryBuilder extends \yii\db\QueryBuilder
         if (empty($uniqueNames)) {
             return $this->insert($table, $insertColumns, $params);
         }
+        if ($updateNames === []) {
+            // there is no columns to update
+            $updateColumns = false;
+        }
 
         $onCondition = ['or'];
         $quotedTableName = $this->db->quoteTableName($table);
--- a/framework/db/mssql/QueryBuilder.php
+++ b/framework/db/mssql/QueryBuilder.php
@@ -455,7 +455,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
             return $this->insert($table, $insertColumns, $params);
         }
         if ($updateNames === []) {
-            // there is no columns to update
+            // there are no columns to update
             $updateColumns = false;
         }
 
--- a/framework/db/mysql/QueryBuilder.php
+++ b/framework/db/mysql/QueryBuilder.php
@@ -278,6 +278,10 @@ class QueryBuilder extends \yii\db\QueryBuilder
         if (empty($uniqueNames)) {
             return $insertSql;
         }
+        if ($updateNames === []) {
+            // there is no columns to update
+            $updateColumns = false;
+        }
 
         if ($updateColumns === true) {
             $updateColumns = [];
--- a/framework/db/mysql/QueryBuilder.php
+++ b/framework/db/mysql/QueryBuilder.php
@@ -279,7 +279,7 @@ class QueryBuilder extends \yii\db\QueryBuilder
             return $insertSql;
         }
         if ($updateNames === []) {
-            // there is no columns to update
+            // there are no columns to update
             $updateColumns = false;
         }
 
--- a/framework/db/oci/QueryBuilder.php
+++ b/framework/db/oci/QueryBuilder.php
@@ -218,6 +218,10 @@ EOD;
         if (empty($uniqueNames)) {
             return $this->insert($table, $insertColumns, $params);
         }
+        if ($updateNames === []) {
+            // there is no columns to update
+            $updateColumns = false;
+        }
 
         $onCondition = ['or'];
         $quotedTableName = $this->db->quoteTableName($table);
--- a/framework/db/oci/QueryBuilder.php
+++ b/framework/db/oci/QueryBuilder.php
@@ -219,7 +219,7 @@ EOD;
             return $this->insert($table, $insertColumns, $params);
         }
         if ($updateNames === []) {
-            // there is no columns to update
+            // there are no columns to update
             $updateColumns = false;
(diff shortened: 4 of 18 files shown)


framework/db/cubrid/QueryBuilder.php
```diff
@@ -70,6 +70,10 @@ public function upsert($table, $insertColumns, $updateColumns, &$params)
         if (empty($uniqueNames)) {
             return $this->insert($table, $insertColumns, $params);
         }
+        if ($updateNames === []) {
+            // there is no columns to update
```
samdark:
```suggestion
            // there are no columns to update
```

framework/db/mssql/QueryBuilder.php
```diff
@@ -454,6 +454,10 @@ public function upsert($table, $insertColumns, $updateColumns, &$params)
         if (empty($uniqueNames)) {
             return $this->insert($table, $insertColumns, $params);
         }
+        if ($updateNames === []) {
+            // there is no columns to update
```
samdark:
```suggestion
            // there are no columns to update
```

framework/db/mysql/QueryBuilder.php
```diff
@@ -278,6 +278,10 @@ public function upsert($table, $insertColumns, $updateColumns, &$params)
         if (empty($uniqueNames)) {
             return $insertSql;
         }
+        if ($updateNames === []) {
+            // there is no columns to update
```
samdark:
```suggestion
            // there are no columns to update
```

framework/db/oci/QueryBuilder.php
```diff
@@ -218,6 +218,10 @@ public function upsert($table, $insertColumns, $updateColumns, &$params)
         if (empty($uniqueNames)) {
             return $this->insert($table, $insertColumns, $params);
         }
+        if ($updateNames === []) {
+            // there is no columns to update
```
samdark:
```suggestion
            // there are no columns to update
```

framework/db/pgsql/QueryBuilder.php
```diff
@@ -338,6 +338,10 @@ private function newUpsert($table, $insertColumns, $updateColumns, &$params)
         if (empty($uniqueNames)) {
             return $insertSql;
         }
+        if ($updateNames === []) {
+            // there is no columns to update
```
samdark:
```suggestion
            // there are no columns to update
```

framework/db/pgsql/QueryBuilder.php
```diff
@@ -368,6 +372,10 @@ private function oldUpsert($table, $insertColumns, $updateColumns, &$params)
         if (empty($uniqueNames)) {
             return $this->insert($table, $insertColumns, $params);
         }
+        if ($updateNames === []) {
+            // there is no columns to update
```
samdark:
```suggestion
            // there are no columns to update
```

framework/db/sqlite/QueryBuilder.php
```diff
@@ -74,6 +74,10 @@ public function upsert($table, $insertColumns, $updateColumns, &$params)
         if (empty($uniqueNames)) {
             return $this->insert($table, $insertColumns, $params);
         }
+        if ($updateNames === []) {
+            // there is no columns to update
```
samdark:
```suggestion
            // there are no columns to update
```

tests/framework/db/oci/QueryBuilderTest.php
```diff
@@ -239,6 +239,10 @@ public function upsertProvider()
         foreach ($concreteData as $testName => $data) {
             $newData[$testName] = array_replace($newData[$testName], $data);
         }
+
+        // skip test
+        unset($newData['no columns to update']);
```
samdark: Why skipping it? It was changed in OCI builder.

tests/framework/db/cubrid/QueryBuilderTest.php
```diff
@@ -107,6 +107,10 @@ public function upsertProvider()
         foreach ($concreteData as $testName => $data) {
             $newData[$testName] = array_replace($newData[$testName], $data);
         }
+
+        // skip test
+        unset($newData['no columns to update']);
```
samdark: Why skipping it? It was changed in the CUBRID builder.

tests/framework/db/oci/QueryBuilderTest.php
alexkart: Yes, but I don't have oci to run tests with it.

tests/framework/db/cubrid/QueryBuilderTest.php
alexkart: Yes, but I don't have cubrid to run tests with it.

tests/framework/db/oci/QueryBuilderTest.php
samdark: Travis has it.

tests/framework/db/cubrid/QueryBuilderTest.php
samdark: OK but it should be removed before merge.

tests/framework/db/oci/QueryBuilderTest.php
alexkart: I think it doesn't https://github.com/yiisoft/yii2/blob/master/.travis.yml#L31

tests/framework/db/cubrid/QueryBuilderTest.php
alexkart: In this case, we need to add the expected sql query, but I don't know what the correct query is and whether it matches the actual query without running the test.

tests/framework/db/oci/QueryBuilderTest.php
samdark: Oh, bad... Anyway, that's not the reason to exclude it.

tests/framework/db/cubrid/QueryBuilderTest.php
samdark: As for CUBRID, that is acceptable :)

samdark:
Thank you!

PR #17519, opened 2019-08-20
Fix #17507: Fixed regular expression escaping in `Controller::createAction()`

### LINKED ISSUE #17507: Yii is returning 404 for two words actions and making dashes fail
Yii is refusing dash character (`-`) in action ids, but it is accepting backslashes (`\`) on the other hand.

I think this is due to a possible mistake in the pattern to validate an id in method `Controller::createAction()`, line 228, as follows:

https://github.com/yiisoft/yii2/blob/efe8d48f156e9e76994a2322cdd0fb4d68bd8f6a/framework/base/Controller.php#L227-L229

The double backslashes in the pattern are matching characters in the range from `\` (93) to `_` (95), which includes `\`, `]`, `^` and `_` instead of only `-` and `_`, which I suppose was the initial intention. 

Can someone please confirm me if that is a bug or am I lacking any information?

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17507

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -5,7 +5,7 @@ Yii Framework 2 Change Log
 ------------------------
 
 - Bug #17511: Fixed IPv6 subnets matching in `IpHelper::inRange()` (kamarton)
-
+- Bug #17507: Fixed regular expression escaping in `Controller::createAction()` (samdark)
 
 2.0.25 August 13, 2019
 ----------------------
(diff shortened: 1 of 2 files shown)


samdark:
Closed in favor of #17523

PR #17527, opened 2019-08-24
.gitignore: add NPM package-lock.json file

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | 
| Tests pass?   | ✔️
| Fixed issues  | -

I use PHPStorm. If installing NPM, then creating this file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ phpunit.phar
 # NPM packages
 /node_modules
 .env
+.package-lock.json


samdark:
Why do we need to ignore it?

kamarton:
It is not at all certain that it created content same things for two developers. If you get into PR, you will have an conflict and you will not be mergeable. 
I'm afraid of being accidentally committing to any pull request.

kamarton:
It was a shame that I did not create an issue for this.

rob006:
IMO it should be committed to repository to ensure that everyone uses the same dependencies. Same story as with `composer.lock`.

kamarton:
I accept the arguments.

PR #17534, opened 2019-08-27
Fix #17522: `DbManager::isEmptyUserId()` is now protected

### LINKED ISSUE #17522: Make yii2/rbac/DbManager more flexible
Adding additional info to either of the auth item tables.
Being able to customize queries that are used in DbManager for certain [ManagerInterface](https://github.com/yiisoft/yii2/blob/master/framework/rbac/ManagerInterface.php) methods, or maybe adding property for additional constraints for retrieving assignments (i.e. https://github.com/yiisoft/yii2/blob/master/framework/rbac/DbManager.php#L708), in a form of https://www.yiiframework.com/doc/api/2.0/yii-db-queryinterface#where()-detail

The default implemetation has nothing wrong with it, but i just need to add extra constraints.
In my case, i need to share rbac info between multiple apps, so having a particular rbac assignment would work in only one of the systems but not all.
### What do you get instead?
Extension of class and overwriting methods used for querying rbac data (in this case assignments) from database. And [some methods](https://github.com/yiisoft/yii2/blob/master/framework/rbac/DbManager.php#L1050) are private for DbManager

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.25
| PHP version      | 7.3.8-1+ubuntu18.04.1+deb.sury.org+1
| Operating system | Linux  4.15.0-58-generic Ubuntu SMP Tue Aug 6 11:12:41 UTC 2019 x86_64

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️ (fixes extensibility blocker)
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17522

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -9,6 +9,7 @@ Yii Framework 2 Change Log
 - Bug #17507: Fixed regular expression escaping and simpler condition in `Controller::createAction()` (kamarton)
 - Bug #16305: Fix `FileValidator` mime-type validation failure because of case sensitivity (kamarton)
 - Bug #17355: Fixed Pjax after request event bug (kamarton)
+- Bug #17522: `DbManager::isEmptyUserId()` is now protected (samdark)
 
 
 2.0.25 August 13, 2019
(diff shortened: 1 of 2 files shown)


samdark:
Merged. Thank you!

Issue #17449, opened 2019-07-19
Comment on a column in a migration is at the wrong place, causing a syntax error


Manu311:
### What steps will reproduce the problem?
Creating a table-column with a "check" and a "comment".
Either by putting it inside of the initial creation or by adding it afterwards.
This is in a default yii-migration-class:
```php
$this->createTable('something', [ 'test' => "longtext NULL DEFAULT NULL COMMENT 'this works' CHECK(json_valid(`test`))" ]);
$this->addCommentOnColumn('something', 'test', 'this breaks');
$this->createTable('something2', [ 'test' => $this->json()->comment('this breaks') ]);
//Created Type: longtext NULL DEFAULT NULL CHECK(json_valid(`test`)) COMMENT 'this breaks'
```

### What is the expected result?
The both not working commands should create MySQL similar to the working example included above.
In the end the column should have the check and the comment.

### What do you get instead?
A Syntax error from MariaDb (and probably MySql if I used that). According to the MySql-reference, comment should be way earlier than check_constraint_definition:
https://dev.mysql.com/doc/refman/8.0/en/create-table.html

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16.1
| PHP version      | 7.3
| Operating system | Linux

samdark:
Do you want to help by working on a pull request?

choken:
I tried to make an pull request, but the test build script failed due to some expected  mysql-5.5 depency.
https://github.com/yiisoft/yii2/pull/17454

samdark:
> Btw, sorry for the double commit. Should I merge the two commits to one?

No, that's fine.

Manu311:
Sadly this doesn't work.
For some reason comments (in mysql) are added via additional alter-commands after table-creation.

I've fixed this problem locally and will create a Pull-Request that work's for me.

YokiToki:
Also `CHECK` statement goes after `AFTER` statement
```php
$this->addColumn('table', 'column', $this->json()->null()->after('previous_column'));
```

mtangoo:
Just stumbled upon extension that sets comment on migration and then Query tries to do it MySQL way. In PostgreSQL the comment is attached later in different statement.

Instead of 
```CREATE TABLE "oauth2_client" (
	"id" serial NOT NULL PRIMARY KEY,
	"some_column" int8 COMMENT "Some comment"
    ....
)```

it is
```CREATE TABLE "oauth2_client" (
[...]

PR #17537, opened 2019-08-28
Fix 17449: Moved Check of column-definition behind Comment when adding a comment to a column

Manu311:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | 
| Fixed issues  | #17449


Manu311:
After trying a lot (for the last 5 hours ....), I figured it's not that easy to write unit-tests for the issue.
The CHECK-feature before MySql 8 was just ignored and removed. So I can't check if it would work with the docker-container.
I tried updating to percona:8 but nothing worked after I just tried that and even if I did fix it, it wouldn't run on CI-tests anyways.

samdark:
So it's causing syntax error when using MySQL 8+ only?

Manu311:
When I switched to MySql 8 the first line of my test which was sth like `$db = $this->getConnection()->pdo` already failed.
I'm not near my code right now, but I could test that once again (tomorrow probably).

Manu311:
The issue was actually not related to yii but rather to an incompatible configuration.
When I try to access the connection, I get:
```
Something wrong when preparing database: SQLSTATE[HY000] [2054] The server requested authentication method unkown to the client.
```

samdark:
Yes, that is known issue: https://stackoverflow.com/questions/52364415/php-with-mysql-8-0-error-the-server-requested-authentication-method-unknown-to

Manu311:
I fixed my mysql-login and tried again. However now this test won't work because with that mysql-version the "bug" apparently doesn't happen.

I have the following code which runs fine in the tests (with `percona:8.0`) but fails in my local system inside a migration (`Mariadb 10.4.6`).
The difference is this (both without my patch):
```
//percona/test:
.string(101) "ALTER TABLE `issue_17449` CHANGE `test_column` `test_column` json DEFAULT NULL COMMENT 'Some comment'"

//mariadb/migration
string(182) "ALTER TABLE `issue_17449` CHANGE `test_column` `test_column` longtext CHARACTER SET utf8mb4 COLLATE utf8mb4_bin DEFAULT NULL  CHECK (json_valid(`test_column`)) COMMENT 'Some comment'"
```
Even if I replace the table creation with the actual DDL of my migration, it will only change to
```
ALTER TABLE `issue_17449` CHANGE `test_column` `test_column` longtext CHARACTER SET utf8mb4 COLLATE utf8mb4_bin COMMENT 'Some comment'
```
which doesn't contain the `check` and therefor can't break.
For some reason the checks are all moved at the end of the table definition while for my (current) mariadb version they are not.

I haven't found a way to reproduce this issue in tests. Percona apparently behaves differently than MariaDb.
If you want to put more than one mysql-database-provider into the tests, I guess that would be the only way to actually verify that this is working.

Manu311:
I added the test and skipped it if the `CHECK` part is not part of the Column-part which is nessesary for the bug to happen.
I locally replaced the percona-container with my mariadb-container and was able to verify the issue without my patch and it being fixed with my patch.

samdark:
So does that mean that it is MariaDB specific?

Manu311:
I haven't tried vanilla MySql. I just tried MariaDb and Percona.
I haven't actually heard about Percona before, so I don't know how close they are to the original MySql, but afaik MariaDb is pretty close.
However I found the reason for the issue (the order) in the official MySql-Docs, so MariaDb is within the specifications.
What Percona does is somewhat different, apparently it does take the (perfectly valid) "CREATE TABLE"-Statement and move "CHECK"s out of the column-definitions and puts them at the end of the table-definition instead.
I assume the behavior of the table itself is identical, but the "CHANGE column" commands are different. Since for Percona it does not contain the CHECK anymore (since that has been moved), so the issue doesn't affect a MySql-backend that manipulates your Table-DDL in this way.

samdark:
Percona should be close to original MySQL.

samdark (approved):

samdark:
Verified and merged. Thank you!

Issue #17434, opened 2019-07-15
Redirecting problem in Internet Explorer 11 (IE11)


art40100:
### What steps will reproduce the problem?
Performing an ajax request and doing a redirect from it.
(Issue when using Internet Explorer 11 (IE11))
Related (same issue) #9670 
### What is the expected result?
A successful redirect to given URL in function Response::redirect()
### What do you get instead?
No redirect!
![jokeie11](https://user-images.githubusercontent.com/33236770/61212503-293f3d80-a70b-11e9-963f-9feb6e4ad551.PNG)

### Additional info
Using 308 statusCode (default is 302) solves the problem in general but it is probably not the right way to do this.
`return Yii::$app->getResponse()->redirect(['aaa/view', 'id' => $this->lid], 308);`

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.22
| PHP version      |  5.6.40
| Operating system | W10x64

art40100:
Quick and dirty solution - extending the Response class and overriding redirect() 
```
class CResponse extends \yii\web\Response
{
    /**
     * @inheritdoc
     */
    public function redirect($url, $statusCode = 302, $checkAjax = true)
    {
        $browser = Yii::$app->browser;
        $browserName = $browser->getName();
        if (Yii::$app->getRequest()->isAjax 
            && ($browserName === $browser::IE)
        ) {
            return parent::redirect($url, 308);
        } else {
            return parent::redirect($url, $statusCode, $checkAjax);
        }
    }
}
```

samdark:
@art40100 what's  `Yii::$app->browser;`?

alex-code:
This should work in JS to use the frameworks redirect handling.
```js
var ua = window.navigator.userAgent;
if (ua.indexOf('MSIE ') > 0 || ua.indexOf('Trident/') > 0 || ua.indexOf('Edge/') > 0) {
  $(document).ajaxSend(function(event, jqXHR, ajaxOptions) {
      jqXHR.setRequestHeader('X-Ie-Redirect-Compatibility', 'true');
  });
}
```

art40100:
> @art40100 what's `Yii::$app->browser`?

in config/main.php
```php
'components' => [
    'browser' => [
        'class' => 'Sinergi\BrowserDetector\Browser',
    ]
];
```
A library installed in my project via composer (https://packagist.org/packages/sinergi/browser-detector)

samdark:
[...]

PR #17555, opened 2019-09-10
Fix regular expression illegal character #17434

kamarton:
Removing a accidentally typed character that causes a regular expression to fails (not match) in some environments.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17434


samdark:
Merged. Thank you!

Issue #17560, opened 2019-09-16
need ExistValidator filter as callable


DrDeath72:
now only string|array|\Closure

samdark:
What's your use case?

DrDeath72:
Big composite DynamicModel with cache. Many "cache killer" place in code where only \Closure allowed :(

samdark:
So what do you want instead? `[MyClass::class, 'mystaticmethod']`?

DrDeath72:
yep... flex \Closure or callable in all palces)

DrDeath72:
RangeValidator range same...

kamarton:
It is advisable to store only attributes in the cache, objects are not recommended.

DrDeath72:
ArrayCache good for reequest time big object, but useless because many \Closure
https://github.com/yiisoft/yii2/search?q=instanceof+/Closure&unscoped_q=instanceof+/Closure

DrDeath72:
good flexible one
https://github.com/yiisoft/yii2/blob/77d70bfbe6c3ce1d9988057b936793a5689e1b76/framework/behaviors/BlameableBehavior.php#L126

samdark:
> yep... flexible \Closure or callable in all places)

So how exactly would you use it?

DrDeath72:
create many objects from "configs" and cache it after build instead of fill private props

DrDeath72:
https://github.com/yiisoft/yii2/pull/17561

PR #17561, opened 2019-09-16
callable in validators

DrDeath72:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   |  ✔️
| Fixed issues  | https://github.com/yiisoft/yii2/issues/17560

### DIFF
--- a/framework/validators/ExistValidator.php
+++ b/framework/validators/ExistValidator.php
@@ -71,7 +71,7 @@ class ExistValidator extends Validator
     /**
      * @var string|array|callable additional filter to be applied to the DB query used to check the existence of the attribute value.
      * This can be a string or an array representing the additional query condition (refer to [[\yii\db\Query::where()]]
-     * on the format of query condition), or an anonymous function with the signature `function ($query)`, where `$query`
+      * on the format of query condition), or an callable with the signature `function ($query)`, where `$query`
      * is the [[\yii\db\Query|Query]] object that you can modify in the function.
      */
     public $filter;
--- a/framework/validators/ExistValidator.php
+++ b/framework/validators/ExistValidator.php
@@ -69,7 +69,7 @@ class ExistValidator extends Validator
      */
     public $targetRelation;
     /**
-     * @var string|array|\Closure additional filter to be applied to the DB query used to check the existence of the attribute value.
+     * @var string|array|\Closure|callable additional filter to be applied to the DB query used to check the existence of the attribute value.
      * This can be a string or an array representing the additional query condition (refer to [[\yii\db\Query::where()]]
      * on the format of query condition), or an anonymous function with the signature `function ($query)`, where `$query`
      * is the [[\yii\db\Query|Query]] object that you can modify in the function.
@@ -125,7 +125,7 @@ class ExistValidator extends Validator
         /** @var ActiveQuery $relationQuery */
         $relationQuery = $model->{'get' . ucfirst($this->targetRelation)}();
 
-        if ($this->filter instanceof \Closure) {
+        if ($this->filter instanceof \Closure || (is_array($this->filter) && is_callable($this->filter))) {
             call_user_func($this->filter, $relationQuery);
         } elseif ($this->filter !== null) {
             $relationQuery->andWhere($this->filter);
@@ -295,7 +295,7 @@ class ExistValidator extends Validator
     {
         /* @var $targetClass \yii\db\ActiveRecordInterface */
         $query = $targetClass::find()->andWhere($condition);
-        if ($this->filter instanceof \Closure) {
+        if ($this->filter instanceof \Closure || (is_array($this->filter) && is_callable($this->filter))) {
             call_user_func($this->filter, $query);
         } elseif ($this->filter !== null) {
             $query->andWhere($this->filter);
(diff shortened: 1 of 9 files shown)


framework/validators/ExistValidator.php
```diff
@@ -69,7 +69,7 @@ class ExistValidator extends Validator
      */
     public $targetRelation;
     /**
-     * @var string|array|\Closure additional filter to be applied to the DB query used to check the existence of the attribute value.
+     * @var string|array|\Closure|callable additional filter to be applied to the DB query used to check the existence of the attribute value.
```
samdark:
```suggestion
     * @var string|array|callable additional filter to be applied to the DB query used to check the existence of the attribute value.
```

framework/validators/ExistValidator.php
```diff
@@ -69,7 +69,7 @@ class ExistValidator extends Validator
...
      * This can be a string or an array representing the additional query condition (refer to [[\yii\db\Query::where()]]
      * on the format of query condition), or an anonymous function with the signature `function ($query)`, where `$query`
```
samdark:
```suggestion
      * on the format of query condition), or an callable with the signature `function ($query)`, where `$query`
```

framework/validators/RangeValidator.php
```diff
@@ -24,7 +24,7 @@
 class RangeValidator extends Validator
 {
     /**
-     * @var array|\Traversable|\Closure a list of valid values that the attribute value should be among or an anonymous function that returns
+     * @var array|\Traversable|\Closure|callable a list of valid values that the attribute value should be among or an anonymous function that returns
```
samdark:
```suggestion
     * @var array|\Traversable|callable a list of valid values that the attribute value should be among or a callable that returns
```

framework/validators/UniqueValidator.php
```diff
@@ -56,7 +56,7 @@ class UniqueValidator extends Validator
      */
     public $targetAttribute;
     /**
-     * @var string|array|\Closure additional filter to be applied to the DB query used to check the uniqueness of the attribute value.
+     * @var string|array|\Closure|callable additional filter to be applied to the DB query used to check the uniqueness of the attribute value.
```
samdark:
```suggestion
     * @var string|array|callable additional filter to be applied to the DB query used to check the uniqueness of the attribute value.
```

framework/validators/UniqueValidator.php
```diff
@@ -56,7 +56,7 @@ class UniqueValidator extends Validator
...
      * This can be a string or an array representing the additional query condition (refer to [[\yii\db\Query::where()]]
      * on the format of query condition), or an anonymous function with the signature `function ($query)`, where `$query`
```
samdark:
```suggestion
     * on the format of query condition), or a callable with the signature `function ($query)`, where `$query`
```

samdark (changes requested):

alex-code:
Is
```php
$this->filter instanceof \Closure
```
still needed with `is_callable`?
Could this be simplified to 
```php
if (is_callable($this->filter)) {
```

DrDeath72:
i think no, do it like in other place :)

samdark (approved):

rob006:
:-1: from me. This will create ambiguous where array could be interpreted either as callable or value - this may lead to really weird bugs and even security issues. For example `['Foo', 'bar']` in `RangeValidator::$range` may be interpreted as list of allowed values. But if you add `Foo` class with `bar` method, it will magically call this method and use result as list of allowed values.

kamarton:
>  from me. This will create ambiguous where array could be interpreted either as callable or value - this may lead to really weird bugs and even security issues. For example `['Foo', 'bar']` in `RangeValidator::$range` may be interpreted as list of allowed values. But if you add `Foo` class with `bar` method, it will magically call this method and use result as list of allowed values.

I completely agree.

rob006:
Just for the record - the use case seems to be valid, but implementation is plain wrong. I would personally disallow `is_callable` in every place where non-callables are also allowed - this is just asking for troubles and it is really easy to shoot yourself in the foot by simple mistake. #17560 probably could be fixed by introducing some kind of `CallableExpression` which will wrap callable array/string, so they will never be confused with plain array/string.

alex-code:
If you're able to use PHP7.1+ there's https://www.php.net/manual/en/closure.fromcallable.php
e.g.
```php
class foo {
    public static function bar() {
        return 'foo-bar';
    }
}

$closure = Closure::fromCallable(['foo', 'bar']);
echo call_user_func($closure);
```

Should work with the existing checks then.

DrDeath72:
Closure not cacheable

kamarton:
Create your own validators and implement the necessary logic in them. Caching validators object is wrong idea.

terabytesoftw:
Thank you for the contribution.

After reviewing the discussion again, I am closing this PR because the proposed approach is not safe to merge. These validator properties already accept arrays as values or query conditions, so interpreting callable arrays specially would make behavior ambiguous and may introduce backward compatibility issues.

The use case described in #17560 may still be valid as a feature request, but it needs a different explicit design rather than changing existing array semantics implicitly.

Closing this PR as outdated and not acceptable in its current form.

Issue #17521, opened 2019-08-21
filterHeaders() should consider user IP


koxu1996:
### What steps will reproduce the problem?

Let's suppose following architecture:
~~~
 Client   --->  Load Balancer  --->  Yii2 App
50.0.0.1          30.0.0.1           10.0.0.1
~~~

_Load Balancer_ is configured to replace `X-Forwarded-For` with reliable value (as advised in https://github.com/yiisoft/yii2/issues/16122#issuecomment-382419713) and _Yii2 App_ is configured to trust headers from *30.0.0.1*.

Example scenario:
 1. _Client_ connects via HTTP**S**
 2. Load balancer replaces *X-Forwarded-For* with user real IP (50.0.0.1) and appends own IP (30.0.0.1). It also sets *X-Forwarded-Proto*, so final headers look like:
    ~~~
    REMOTE_ADDR: 50.0.0.1
    X-Forwarded-For: 50.0.0.1,30.0.0.1
    X-Forwarded-Proto: https
    ~~~

### What is the expected result?

Headers should be trusted and `getIsSecureConnection()` should return `true`.

### What do you get instead?

Headers are not trusted, because [filterHeaders](https://github.com/yiisoft/yii2/blob/master/framework/web/Request.php#L300) uses `getRemoteIP` instead of `getUserIP`.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.23
| PHP version      | PHP 7.3.8
| Operating system | Arch Linux x64 5.2.9-arch1-1-ARCH

nadar:
We do have the same problem, if filterHeaders() would take the load balancer proxy as trusted source this would just work out of the box: https://github.com/yiisoft/yii2/blob/master/framework/web/Request.php#L300

Define REMOTE_ADDR (which is taken to determine trustedHosts) on a constantly changing infrastructure like AWS, Google Cloud is almost not possible, or do i miss something? So we ended up with override `secureHeaders` with an empty array, but this does not look like a good solution.

```php
'request' => [
      'secureHeaders' => [],
],
```

@koxu1996 taking `getUserIP()` won't work as far as i can see, as headers are already resolved: https://github.com/yiisoft/yii2/blob/cdd40b8dfb3a2f29a72fb2e07ca0e9718f15494c/framework/web/Request.php#L341

koxu1996:
I overrided `secureHeaders` as well.

@nadar You are right, it would result in mutual recursion.

[...]

PR #17569, opened 2019-09-25
test for trusted host + isSecureConnection #17521

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17521 

`Request::$userIpIsAlwaysRemoteIp` added for resolve the problem.

### DIFF
--- a/tests/framework/web/RequestTest.php
+++ b/tests/framework/web/RequestTest.php
@@ -742,15 +742,78 @@ class RequestTest extends TestCase
         $this->assertSame($expectedMethod, $request->getMethod());
     }
 
-    public function testTrustedHostAndSecureConnection()
+    /**
+     * @dataProvider trustedHostsDataProvider
+     */
+    public function testTrustedHosts($remoteAddress, $xForwardedFor, $xForwardedProto, $trustedHosts, $userIpIsRemoteIp, $expectedUserIp, $expectedRemoteAddress, $expectedISecureConnection)
     {
-        $_SERVER['REMOTE_ADDR'] = '30.0.0.1';
-        $_SERVER['HTTP_X_FORWARDED_FOR'] = '50.0.0.1,30.0.0.1';
-        $_SERVER['HTTP_X_FORWARDED_PROTO'] = 'https';
+        $_SERVER['REMOTE_ADDR'] = $remoteAddress;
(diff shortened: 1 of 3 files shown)


tests/framework/web/RequestTest.php
```diff
@@ -741,4 +741,16 @@ public function testRequestMethodCanNotBeDowngraded($requestMethod, $requestOver
 
         $this->assertSame($expectedMethod, $request->getMethod());
     }
+
+    public function testTrustedHostAndSecureConnection()
+    {
+        $_SERVER['REMOTE_ADDR'] = '30.0.0.1';
```
koxu1996: Try setting `REMOTE_ADDR` to client IP (50.0.0.1) and you will see that it does not work as expected.

kamarton:
```
 Client   --->  Load Balancer  --->  Yii2 App
50.0.0.1          30.0.0.1           10.0.0.1
```

The `REMOTE_ADDR` cannot be `50.0.0.1`.

koxu1996: You are wrong, because it's totally possible - that's my real case. If you want to reproduce it then you can use _load balancer_ for _nginx_ servers with `ngx_http_realip` module.

koxu1996:
IMO this PR does not make sense: you can just call `getRemoteIp()`, so `$userIpIsAlwaysRemoteIp `property is useless.

tests/framework/web/RequestTest.php
kamarton: That's what I mean, that's exactly what PR added.

kamarton:
> IMO this PR does not make sense: you can just call getRemoteIp(), so $userIpIsAlwaysRemoteIp property is useless.

No. A previous network endpoint has already completed the user's IP resolution, so the value of `$ _SERVER ['REMOTE_ADDR']` must be accepted for the user's IP address. `getUserIP()` must indicate this, otherwise yii will not work properly.

samdark:
Thanks for contribution but I don't see it as a correct way to solve the issue. See original issue comments.

Issue #17533, opened 2019-08-27
yiiunit\framework\mutex\RetryAcquireTraitTest::testRetryAcquire test fail by ci.appveyor.com


kamarton:
### What steps will reproduce the problem?

https://ci.appveyor.com/project/samdark/yii2/builds/26985116#L161

### What is the expected result?

accept 49.97

### What do you get instead?

```bash
1) yiiunit\framework\mutex\RetryAcquireTraitTest::testRetryAcquire
Failed asserting that 49.96800422668457 is equal to 50 or is greater than 50.
```

### Additional info

Maybe the float vs integer is the problem.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.26-dev
| PHP version      | ?
| Operating system | windows

samdark:
That is an indication that it fails with certain conditions. Any idea on what conditions these may be?

kamarton:
At first glance, it cannot be less than 50.0, unless it is time traveler.

samdark:
That is the issue with concurrency issues. They are very similar to time travelers.

kamarton:
There is a bigger difference:
```
1) yiiunit\framework\mutex\RetryAcquireTraitTest::testRetryAcquire
Failed asserting that 48.61307144165039 is equal to 50 or is greater than 50.
C:\projects\yii2\tests\framework\mutex\RetryAcquireTraitTest.php:45
```

samdark:
I don't think it is correct to ignore it in tests. I've reproduced it. Trying to understand what actually happens.

samdark:
OMG, seems windows `usleep()` precision is awful: https://ci.appveyor.com/project/samdark/yii2/builds/27850232. It finished 2ms earlier than specified.

samdark:
Aha. That is normal on Windows: https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep?redirectedfrom=MSDN

> If dwMilliseconds is less than the resolution of the system clock, the thread may sleep for less than the specified length of time.

PR #17574, opened 2019-09-26
Request::getUserIP() secure fix #17573

### LINKED ISSUE #17573: Request::getUserIP() not secure with trustedHost and ip headers

Processing of `Request::getUserIP()` is not secure for `trustedHost`, for example from `X-Forwarded-For`.

network:
```
client ------ > valid network endpoint (e.g. LB) -----> server (---> PHP)
1.1.1.1             2.2.2.2                        3.3.3.3
```

simple way:
- client not send X-Forwarded-For
- LB add client IP to X-Forwarded-For (`1.1.1.1`)
- PHP see `X-Forwarded-For: 1.1.1.1` / `Remote address: 2.2.2.2`

hard way:
- client send X-Forwarded-For. This is due to additional network endpoints or malicious injection. (`127.0.0.1, 8.8.8.8`)
- LB add client IP (`127.0.0.1, 8.8.8.8,1.1.1.1`)
- PHP see `X-Forwarded-For: 127.0.0.1, 8.8.8.8,1.1.1.1` / `Remote address: 2.2.2.2`

If `trustedHost` is set to 2.2.2.2 then the `getUserIp()` is `1.1.1.1` (the address of after the last valid endpoint)

### What do you get instead?

If `trustedHost` is set to 2.2.2.2 then the `getUserIp()` is `127.0.0.1`.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.x
| PHP version      | -
| Operating system | -

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17573


framework/web/Request.php
```diff
@@ -1131,15 +1142,55 @@ public function getUserAgent()
      */
     public function getUserIP()
     {
-        foreach ($this->ipHeaders as $ipHeader) {
+        foreach($this->getTrustedIpHeaders() as $ipHeader) {
             if ($this->headers->has($ipHeader)) {
-                return trim(explode(',', $this->headers->get($ipHeader))[0]);
+                $ip = $this->getUserIpFromIpHeader($this->headers->get($ipHeader));
+                if(isset($ip)) {
```
samdark:
```suggestion
                if ($ip !== null) {
```

framework/web/Request.php
```diff
@@ -1131,15 +1142,55 @@ public function getUserAgent()
...
+     * @param string $ips comma separated IP list
+     * @return string|null
+     * @see $getUserHost
+     * @see $ipHeader
+     * @see $trustedHeaders
+     * @since 2.0.28
+     */
+    protected function getUserIpFromIpHeader($ips)
+    {
+        $ips = preg_split('/\s*,\s*/', trim($ips));
```
samdark:
```suggestion
        $ips = preg_split('/\s*,\s*/', trim($ips), -1, PREG_SPLIT_NO_EMPTY);
```

framework/web/Request.php
```diff
@@ -1131,15 +1142,55 @@ public function getUserAgent()
...
+        $resultIp = null;
+        $stop = false;
+        foreach ($ips as $ip) {
+            foreach ($this->trustedHosts as $cidr => $cidrOrHeaders) {
+                if (!is_array($cidrOrHeaders)) {
+                    $cidr = $cidrOrHeaders;
+                }
+                $validator->setRanges($cidr);
+                $resultIp = $ip;
+                if ($validator->validate($ip)) {
```
samdark: I think we don't need `$stop` at all:

```php
if (!$validator->validate($ip)) {
    break 2;
}
```

samdark (changes requested):

framework/web/Request.php
kamarton: `$stop` is required because it is within the cycle that it will be overwritten. If it find a IP address in a subsequent iteration of the loop, `$stop=false` and the internal loop will end.

I simplified the code.

rob006: Maybe renaming this variable (`$isValid`?) will help - right now it is little hard to understand what is happening here, because `$stop` does not really explaining why it stops.

framework/web/Request.php
```diff
@@ -1131,15 +1142,53 @@ public function getUserAgent()
...
+                    $stop = false;
+                    break;
+                }
+            }
+            if ($stop) {
+                break;
+            }
+        }
+        return $resultIp;
```
rob006: This looks suspicious. `$resultIp` is set before `$validator->validate($ip)`, so at this point it could contain invalid IP?

kamarton: validation added

framework/web/Request.php
kamarton: variable is renamed

framework/web/Request.php
rob006: Why `$resultIp = $ip;` is outside of `$validator->validate($ip)` condition?

kamarton: I moved it to the main cycle. 

> Why $resultIp = $ip; is outside of $validator->validate($ip) condition?

After the last trusted IP address is the usable IP address. The result may be invalid IP or even nonexistent. Therefore, at the very end of the method, it validates as a plus.

kamarton:
#17533

kamarton:
@sam002 @rob006 The code is a bit redesigned to make it much clearer. Additional tests have been added.

samdark (approved):

samdark:
Merged. Thanks!

PR #17575, opened 2019-09-26
activeListInput-unselect

alex-code:
Allow `['unselect' => false]` to omit hidden input.

e.g.
```php
Html::activeRadioList($model, 'attribute', [
  'Value 1' => 1,
  'Value 2' => 2
], [
  'unselect' => false
])
```

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.28 under development
 ------------------------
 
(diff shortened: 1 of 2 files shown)


rob006:
You should use `'unselect' => null` if you want to omit hidden field.

alex-code:
My bad, should've looked further to see the `isset` :)

PR #17577, opened 2019-09-27
Removing repeated description about inline actions

LSViana:
The same content is already described at line 230 ("Inline actions take less effort...")

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Documentation revision | ✔️
| Fixed issues  | No issues related


samdark:
It's not duplicate. It is telling when to use inline actions and when to use standalone actions.

PR #17579, opened 2019-09-28
Added guestRequired() User method.

wrux:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | I added a method to redirect away from guest pages for authenticated users. 

While working on a login template on CraftCMS I noticed that there is a method for `loginRequired()` but not `guestRequired()`. This is useful to redirect away from login/registration pages.

### DIFF
--- a/framework/web/User.php
+++ b/framework/web/User.php
@@ -92,19 +92,7 @@ class User extends Component
      * If this property is `null`, a 403 HTTP exception will be raised when [[loginRequired()]] is called.
      */
     public $loginUrl = ['site/login'];
-    /**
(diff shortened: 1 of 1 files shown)


samdark:
Thanks for contribution but pull request won't be merged:

1. Usually it's checked via access control in controller.
2. It's not common to do it other way.
3. Yii 2.0 (framework itself) is generally not accepting enhancements (except very important ones such as security-related). We're currently focused on Yii 3 and Yii 2 bugfixing.

PR #17580, opened 2019-09-29
Super slow schema queries with MySQL 8.0.16 fix #17444

### LINKED ISSUE #17444: Super slow schema queries with MySQL 8.0.16
After updating to MySQL 8, I have noticed how super slow Yii2 has become.
After using the Yii Debugger, I have found the following 6 queries consuming more than 95% of total loading time: 

The only way I was able to fix the issue is by adding
`'enableSchemaCache' => true,` to the database component and enabling caching `class' => 'yii\caching\FileCache',`
But since I'm using my machine for development I don't want to enable caching.

Note: I tried the same code with MySQL version 5 and it was working fine. Seems an issue with MySQL 8 and Yii2, I'm not sure.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.22
| PHP version      | 7.2.20
| Operating system | macOS Mojave 10.14.5

kamarton:
**work in progress**

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | WIP
| Fixed issues  | #17444

Fixed with a simple solution: >99.99% of the use, no required loading of constraints. It loads on first use.

- [ ] Serialization of 'Closure' is not allowed
  - prepare for cache, callable change to false
- [x] fix drivers
- [x] local unit test

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.28 under development
 ------------------------
 
-- no changes in this release.
+- Bug #17444: fixed super slow schema queries with MySQL 8.0.16 (kamarton) 
 
 
 2.0.27 September 18, 2019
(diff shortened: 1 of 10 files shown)


kamarton:
BC, Cannot be solved under 2.0.x.

PR #17582, opened 2019-09-30
update db-dao

sihar:
> Info: In Yii 2.1 and later, the DAO supports for CUBRID, Oracle and MSSQL are no longer provided as the built-in
  core components of the framework. They have to be installed as the separated [extensions](structure-extensions.md).
  There are [yiisoft/yii2-oracle](https://www.yiiframework.com/extension/yiisoft/yii2-oracle) and
  [yiisoft/yii2-mssql](https://www.yiiframework.com/extension/yiisoft/yii2-mssql) in the 
  [official extensions](https://www.yiiframework.com/extensions/official).

I suggest remove this part, since yii2 only have version 2.0 and oracle extension still provided as the built-in
  core components of the framework (I have test insert data using dao in oracle) CMIIW


samdark:
Good catch. Thanks!

PR #17583, opened 2019-09-30
Changed ActiveField __toString method to catch all php 7 error types.

diego-mathis:
When developing code, error message related to ActiveField are very obscure. The actual error is not shown. All what is given, is that __toString is not allowed to throw an error.

\Throwable is part of php 7+, so I left the \Exception in for older php.
According to 3v4l.org this seems to work.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   |  ✔️
| Fixed issues  | -

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.28 under development
 ------------------------
 
-- no changes in this release.
+- Bug #17853: Fix errors in ActiveField to be properly caught when PHP 7 is used (My6UoT9)
 
 
 2.0.27 September 18, 2019
--- a/framework/widgets/ActiveField.php
+++ b/framework/widgets/ActiveField.php
@@ -177,6 +177,9 @@ class ActiveField extends Component
         } catch (\Exception $e) {
             ErrorHandler::convertExceptionToError($e);
             return '';
+        } catch (\Throwable $e) {
+            ErrorHandler::convertExceptionToError($e);
+            return '';
         }
     }
 


samdark:
@My6UoT9 do you have a good way to reproduce the issue this pull request is fixing?

My6UoT9:
I tried to make a small example, but seems its not so easy. The code which had the issue, is complex, as it does extend ActiveForm and ActiveField - and I am not allowed to share that.
So far, all my simple tries to cause an error in the view or in some widget where handled correctly - and I saw a proper error msg, not sure how to make a proof of this problem. 
Will try again tomorrow.

As info in cakephp they made something smiliar, https://github.com/cakephp/cakephp/issues/9043

My6UoT9:
Ok this is a total artificial test, but its shows that the error information is almost compeletely lost.
A CRUD which uses the ActiveField is required for this "test".

In the `public function render($content = null)` (`vendor/yiisoft/yii2/widgets/ActiveField.php`), add `throw \CompileError;` as first line in the function body. 
Then just load a page which uses ActiveField, 
You should get with (YII_DEBUG enabled) 
```
PHP Fatal Error – yii\base\ErrorException
Method yii\widgets\ActiveField::__toString() must not throw an exception, caught Error: Undefined constant 'CompileError'
...
...
...
```

Now if we add my PR, the error shows as:
```
PHP User Error – yii\base\ErrorException
Exception 'Error' with message 'Undefined constant 'CompileError''

in C:\laragon\www\fire\vendor\yiisoft\yii2\widgets\ActiveField.php:206
...
...
...
```

samdark:
Merged. Thank you!

PR #17586, opened 2019-10-01
Fix #17585: @calendar param included in Formatter Yii:t() calls

### LINKED ISSUE #17585: `@calendar` locale param passed to Yii::t() calls from Formatter
Yii 2.0.7 introduced the ability to format dates with non-Gregorian calendars (cdb7d84fea5c502b17bffa639d69108cead27a3f). The main way the docs specify to do that is by setting the `Formatter` class’s new `calendar` property to `IntlDateFormatter::TRADITIONAL`, and appending `@calendar=<calendar_type>` to the `locale` property.

However, the `locale` property is passed to several `Yii::t()` calls within the `Formatter` class, which is only expecting a language string, not `<language>@calendar=<calendar_type>`.

https://github.com/yiisoft/yii2/blob/940eb363ab02999ececef47733bf6e2e8d86c761/framework/i18n/Formatter.php#L389

These calls will end up causing Yii to re-load all `yii` translations messages with a different translation key that includes the `@calendar` param, and if the language happens to be more than two letters (e.g. `xx-YY`), the correct translations won’t even get loaded; `yii\i18n\PhpMessageSource::loadMessages()` won’t be able to find `xx-YY@calendar=<calendar_type>/yii.php` so it will fall back to `xx/yii.php` instead (if it even exists – not the case with `nb-NO`, `zh-CN`, or `zh-TW`).

Create a `yii\i18n\Formatter` instance with a locale ID that has more than two characters, and a custom calendar type.

```php
$formatter = new Formatter([
    'locale' => 'zh-CN@calendar=persian',
    'calendar' => IntlDateFormatter::TRADITIONAL,
]);
```

The formatter’s internal `Yii::t()` calls should ignore the `@calendar` param on the `locale` property.

### What do you get instead?

`Yii::t()` calls include the `@calendar` param in their `$language` argument.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.27
| PHP version      | N/A
| Operating system | N/A

brandonkelly:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17585

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.28 under development
 ------------------------
 
-- no changes in this release.
+- Bug #17585: Fix `yii\i18n\Formatter` including the `@calendar` locale param in `Yii::t()` calls (brandonkelly)
 
 
 2.0.27 September 18, 2019
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -97,6 +97,13 @@ class Formatter extends Component
      */
     public $locale;
     /**
+     * @var string the language code (e.g. `en-US`, `en`) that is used to translate internal messages.
+     * If not set, [[locale]] will be used (without the `@calendar` param, if included).
+     *
+     * @since 2.0.28
+     */
+    public $language;
+    /**
      * @var string the time zone to use for formatting time and date values.
      *
      * This can be any value that may be passed to [date_default_timezone_set()](https://secure.php.net/manual/en/function.date-default-timezone-set.php)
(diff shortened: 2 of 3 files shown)


samdark:
There seems to be another way to solve the issue via getting language string from locale. Something like what's done for Yii 3: https://github.com/yiisoft/i18n/blob/master/src/Locale.php#L86. This way we can avoid having additional parameter.

brandonkelly:
I could change the visibility of `public $language` to `protected` or `private` so it’s not introducing an additional configurable param, if that’s preferred. Seems like implementing this `getBCP47Regex()` thing is a little overkill though.

samdark:
Actually, the approach is very similar. The difference is how language is obtained from locale.

samdark (approved):

samdark:
Merged. Thank you!

brandonkelly:
Awesome, thanks!

PR #17588, opened 2019-10-01
No Array_merge In Loops

rumours86:
https://github.com/dseguy/clearPHP/blob/master/rules/no-array_merge-in-loop.md

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | ❌

### DIFF
--- a/framework/base/Event.php
+++ b/framework/base/Event.php
@@ -270,11 +270,7 @@ class Event extends BaseObject
         }
         $wildcardEventHandlers = [];
         if ($classHandlersList) {
-            if (version_compare(PHP_VERSION, '5.6', '>=')) {
-                $wildcardEventHandlers = array_merge(...$classHandlersList);
-            } else {
-                $wildcardEventHandlers = call_user_func_array('array_merge', $classHandlersList);
-            }
+            $wildcardEventHandlers = call_user_func_array('array_merge', $classHandlersList);
         }
 
         if (empty(self::$_events[$name]) && empty($wildcardEventHandlers)) {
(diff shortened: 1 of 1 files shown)


samdark:
Seems tests are failing after the change.

rob006:
I'm unsure if it will improve performance overall. AFAIK in most cases `$_eventWildcards` will be empty or contain one element, and in this case old implementation was more efficient.

rumours86:
@samdark  не знаю как исправить codeclimate

samdark:
OK. Now it passes tests. Is there any real performance benefit?

rumours86:
у нас была ошибка
Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 20480 bytes) in vendor/yiisoft/yii2/base/Event.php on line 292

я увидел мержи в цикле и решил их поправить, тем более пхпшторм подсвечивает их красным

samdark:
Was the error you were getting resolved by the change you've made?

rumours86:
no

samdark:
In order to merge it we need to prove that it improves something. I think performance test is needed.

rumours86:
https://github.com/dseguy/clearPHP/blob/master/rules/no-array_merge-in-loop.md

mikehaertl:
Isn't this premature optimization? The benefit looks so extremely small that it can be neglected in almost every thinkable case.

My6UoT9:
When u dont use wildcards it might spend a little more time for this, but if you use wildcards - this patch is surely better.
Does not look like premature optimization, running code in a loop - which does not belong in the loop.

rob006:
> When u dont use wildcards it might spend a little more time for this, but if you use wildcards - this patch is surely better.

If you care about performance, you don't use wildcards at all (or at least no more than one or two). This `array_merge()` micro-optimization is unnoticeable when you compare this with overhead created by actual event processing by wildcard handlers.

rumours86:
i use 
$result = call_user_func_array('array_merge', $source);
because yii2 support php5.4

samdark:
Theoretically it might improve performance but I'd not merge it without a benchmark.

terabytesoftw:
Any update on this topic, it looks like we can close it.

PR #17589, opened 2019-10-02
update db-active-record.md

sihar:
I have tried example script working in transaction using active record (first way). It doesn't work like transaction using db dao because method save only return boolean and doesn't return throw exception when fail. I have not try the second way example script, I will try it later. Thanks.

### DIFF
--- a/docs/guide/db-active-record.md
+++ b/docs/guide/db-active-record.md
@@ -667,7 +667,14 @@ $customer = Customer::findOne(123);
 
 Customer::getDb()->transaction(function($db) use ($customer) {
     $customer->id = 200;
-    $customer->save();
+    if(!$customer->save())
+        throw new Exception('customer data saving failed!');
+        
+    $order = new Order;
+    $order->name = 'zzz';
+    if(!$order->save())
+        throw new Exception('order data saving failed!');
+        
     // ...other DB operations...
 });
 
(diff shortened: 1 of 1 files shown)


rob006:
`save()` returns `false` only if validation doesn't pass - you should validate model before starting transaction (and probably handle this more gently than throwing an exception). 

`save()` may still throw an exception on SQL error.

sihar:
> `save()` returns `false` only if validation doesn't pass - you should validate model before starting transaction (and probably handle this more gently than throwing an exception).
> 
> `save()` may still throw an exception on SQL error.

Thanks @rob006  for your response before. I suggest this because I want scenario if one data saving fail to saved then all data saving will rollback. I don't know how to handle this besides throwing an exception. May you give example how to handle transaction besides this one?

rob006:
You can just do this **before** starting transaction:

```php
if (!$customer->validate() || !$order->validate()) {
    // handle error, usually display form with validation errors
}
```

sihar:
thanks again @rob006 
I know this pull request closed. But seems I describe wrong about my question, I want to clearify how we trigger method save() in AR so it will throw exception and rollback transaction?
I mean we can write code
```php
if (!$customer->validate() || !$order->validate()) {
    // handle error, usually display form with validation errors
}

$transaction = Customer::getDb()->beginTransaction();
try {
    $customer->id = 200;
    $customer->save();
    // ...other DB operations...
    $transaction->commit();
} catch(\Exception $e) {
    $transaction->rollBack();
    throw $e;
} catch(\Throwable $e) {
    $transaction->rollBack();
    throw $e;
}
```

How we trigger this block try so it will thrown an exception? Should I worry about this?
Or I should trust this code and only fail/rollback if database server goes down?
Is this probably what you mean?

samdark:
> How we trigger this block try so it will thrown an exception?

Database layer throws exception in case of failure to perform a query.

> Should I worry about this?

No.

> Is this probably what you mean?

Yes. If validation passed, there's only database failures to look at and these are handled.

sihar:
many thanks @samdark 
I am relieved now

PR #17594, opened 2019-10-05
already resolved ip #17521

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17521
| Depend on | https://github.com/yiisoft/yii2/pull/17593

If the test is correct (depend on https://github.com/yiisoft/yii2/pull/17593), it is sufficient to supplement the documentation with this case of use.


samdark:
Seems to fail.

kamarton:
> Seems to fail.

Yes, I have noticed that during https://github.com/yiisoft/yii2/pull/17574 PR, `secureHeaders` is used incorrectly to resolve the IP address. 

https://github.com/yiisoft/yii2/blob/23a8d041a6867090d95df9051b294dcc73f50b7e/framework/web/Request.php#L1145

https://github.com/yiisoft/yii2/pull/17593 in PR is patched and uses `ipHeaders`.

Then the idea will work properly.

samdark:
Merged. Thanks!

Issue #17479, opened 2019-07-30
Missing Icons in ActionColumn with Bootstrap 4.3.x


demonking:
With the Version 4 of Bootstrap, the icons were removed from the default library.

https://getbootstrap.com/docs/4.0/extend/icons/

### What steps will reproduce the problem?

AppAsset: 

~~~
            'yii\bootstrap4\BootstrapAsset',
            'yii\bootstrap4\BootstrapPluginAsset',
~~~

GridView: 
~~~
 
   <?= GridView::widget([
        'dataProvider' => $dataProvider,
        'columns' => [
            ['class' => 'yii\grid\SerialColumn'],

            'id',
            'name',
            'description:ntext',
            'image',

            [
                'class' => 'yii\grid\ActionColumn',
            ]
        ],
    ]); ?>


~~~
### What's expected?

Seeing the icons in the last column of the gridview.

### What do you get instead?

The css classes are not present in boostrap, so the icon's as not visible.

### Additional info

| Q                | A
| ---------------| ---
| Yii version   | 2
| PHP version  | 7.2
| Operating system | Linux Debian Stretch

For the Moment I use the awesome font library and a function which creates the Buttons:

~~~
        function createActionButtons($buttons=['view','update','delete']){
                $actionButtons = [];
                $actionButtons['update']= function($url,$model) {
                         return Html::a('<i class="fa fa-edit"></i>', $url, [
                                 'title' => Yii::t('app', 'edit')
                         ]);
                };
                $actionButtons['delete'] = function($url,$model) {
                         return Html::a('<i class="fas fa-trash"></i>', $url, [
                                 'title' => Yii::t('app', 'delete')
                         ]);
                };
                $actionButtons['view'] = function($url,$model) {
                         return Html::a('<i class="fas fa-eye"></i>', $url, [
                                 'title' => Yii::t('app', 'view')
                         ]);
                };
                return array_intersect_key($actionButtons,array_flip($buttons));
        }
~~~

samdark:
Moved.

rob006:
Related: https://github.com/yiisoft/yii-dataview/issues/10

jai3232:
The easiest way is add the glyphicon to the Yii2. There is already yii2-bootstrap4-glyphicons by xtetis.
You can the see sort icon in the GridView as well.

imitronov:
```css
.glyphicon {
    font-family: "Font Awesome 5 Free";
}
.glyphicon-eye-open:before {
[...]

PR #17601, opened 2019-10-07
replace deprecated bootstrap3 glyphicons with fontawesome5 svg-sprites.

HenryVolkmer:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17479

### DIFF
--- a/framework/grid/ActionColumn.php
+++ b/framework/grid/ActionColumn.php
@@ -142,9 +142,9 @@ class ActionColumn extends Column
      */
     protected function initDefaultButtons()
     {
-        $this->initDefaultButton('view', 'eye-open');
-        $this->initDefaultButton('update', 'pencil');
-        $this->initDefaultButton('delete', 'trash', [
+        $this->initDefaultButton('view');
+        $this->initDefaultButton('update');
+        $this->initDefaultButton('delete', [
             'data-confirm' => Yii::t('yii', 'Are you sure you want to delete this item?'),
             'data-method' => 'post',
         ]);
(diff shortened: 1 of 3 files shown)


framework/grid/ActionColumn.php
```diff
@@ -153,33 +153,53 @@ protected function initDefaultButtons()
     /**
      * Initializes the default button rendering callback for single button.
      * @param string $name Button name as it's written in template
-     * @param string $iconName The part of Bootstrap glyphicon class that makes it unique
      * @param array $additionalOptions Array of additional options
      * @since 2.0.11
      */
-    protected function initDefaultButton($name, $iconName, $additionalOptions = [])
+    protected function initDefaultButton($name, $additionalOptions = [])
```
samdark: That could be a problem for classes extending `ActionColumn`. How about leaving $iconName but marking it as `@deprecated` instead?

samdark:
Also it doesn't seem to pass tests...

My6UoT9:
Large tables with lots of svg are going be laggy, it also increases the size. Why not use emoticon, extend it so that a developer can set easily a character and font by himself.

framework/grid/ActionColumn.php
HenryVolkmer: yes, you are right. I make it so.

HenryVolkmer:
> Also it doesn't seem to pass tests...

adjusted tests in todays push.

My6UoT9:
> Attribution is required by MIT, SIL OLF, and CC BY licenses. Downloaded Font Awesome Free files already contain embedded comments with sufficient attribution, so you shouldn't need to do anything additional when using these files normally.

https://github.com/FortAwesome/Font-Awesome#license

HenryVolkmer:
> Large tables with lots of svg are going be laggy, it also increases the size. Why not use emoticon, extend it so that a developer can set easily a character and font by himself.

it think, in 2020 the enduser-hardware have more as enough computing power to render dozens of svg's.
Further, Fontawesome leads the topic in the documentaries (https://fontawesome.com/how-to-use/on-the-web/advanced/svg-sprites).

But in general, i pledge for the these options:

utilize Bootstrap buttons instead of icon(fonts)

- BS4 has no bc's to BS3-syntax
- no additionaly Libs required
- safe till yii2 EOL
Bildschirmfoto von 2019-10-10 10-24-28

*or*

we welcome fontawesome5-free in the yii2 assets

- easy icon configuration via ActionColumn::buttonOptions

*or*

we implement a property ActionColumn::icon[]

the keys are the button-names and the values a string or a callable.
The string (or the result of the callable) are used as icon.
If no icon is configured, the Button-name will be used as fallback.
Togehter with `$buttonOptions = ['class' => 'btn btn-default btn-sm'];` a bootstrap-button would be rendered.

My6UoT9:
> it think, in 2020 the enduser-hardware have more as enough computing power to render dozens of svg's.
That is simply not true. There are is still people paying for each mb they download, no flatrate.
And there will always be devices which have limited computing power.

> utilize Bootstrap buttons instead of icon(fonts)
True, fonts would not be that great.

> Further, Fontawesome leads the topic in the documentaries
https://fontawesome.com/how-to-use/on-the-web/advanced/svg-sprites
On that page they talk about JS + SVG, but that is not even needed.

JS is not required for reusing svg, see here for a example:
https://css-tricks.com/svg-sprites-use-better-icon-fonts/


> adjusted tests in todays push.
it seems like ur push did not end up in github

HenryVolkmer:
> That is simply not true. There are is still people paying for each mb they download, no flatrate.
> And there will always be devices which have limited computing power.

we talking not about the byte-size of the inline-svg, we talking about cpu-costs

> On that page they talk about JS + SVG, but that is not even needed.

no, the documentary is about the useage without JS.
"... goodness of SVG **without** having the need for our SVG + **JS framework** "
 

> > adjusted tests in todays push.
> 
> it seems like ur push did not end up in github

it was late yesterday. Let's first find a passable solution and then write the code.

@samdark whats your thoughts?

HenryVolkmer:
useage of inline-svg is totally wrong here and should be rejected.

My6UoT9:
> useage of inline-svg is totally wrong here and should be rejected.

I dont get it, was't ur proposal with inline-svg ?
https://github.com/yiisoft/yii2/pull/17601/commits/8a27a7fe93d8b6897b6d3b4c51a063621b0689b9
As far as I understand that code, it does insert the svg for each button.

My proposal was simply to use svg sprites to reduce both cpu time and size.

HenryVolkmer:
it was first discussed here https://github.com/yiisoft/yii2/issues/17479 
There was no no-votes. 
Since our discourse, its clearer for me.

have a look (the btn way): https://github.com/HenryVolkmer/yii2/commit/595b6700baca55fe1d525e5171cf5dd863cd34ba

My6UoT9:
Sorry did overlook that issue, should had said something back then.
Only button text is ideal I think, easy to extend.

tests/framework/grid/ActionColumnTest.php
```diff
@@ -41,9 +41,10 @@ public function testRenderDataCell()
             return 'http://test.com';
         };
         $columnContents = $column->renderDataCell(['id' => 1], 1, 0);
-        $viewButton = '<a href="http://test.com" title="View" aria-label="View" data-pjax="0"><span class="glyphicon glyphicon-eye-open"></span></a>';
-        $updateButton = '<a href="http://test.com" title="Update" aria-label="Update" data-pjax="0"><span class="glyphicon glyphicon-pencil"></span></a>';
-        $deleteButton = '<a href="http://test.com" title="Delete" aria-label="Delete" data-pjax="0" data-confirm="Are you sure you want to delete this item?" data-method="post"><span class="glyphicon glyphicon-trash"></span></a>';
+
+        $viewButton = '<a class="btn btn-default btn-sm" href="http://test.com" title="View" aria-label="View" data-pjax="0">View</a>';
```
samdark: `btn btn-default btn-sm` may look weird in gridview.

framework/grid/ActionColumn.php
```diff
@@ -142,9 +142,9 @@ public function init()
      */
     protected function initDefaultButtons()
     {
-        $this->initDefaultButton('view', 'eye-open');
-        $this->initDefaultButton('update', 'pencil');
-        $this->initDefaultButton('delete', 'trash', [
+        $this->initDefaultButton('view');
```
samdark: Does it just remove icons, not replacing them?

rob006:
Using text buttons is huge UX regression. I'm not sure why SVG approach was rejected. The only drawback of this solution is repeating the same SVG multiple times, but there are techniques to avoid this.

HenryVolkmer:
yii2 is coupled with bootstrap3, so we could leave the \yii\grid\ActionColumn as it is.

#17479 is only an issue if yii\bootstrap4 is used. Therefore, my proposal is to provide a `\yii\bootstrap4\ActionColumn` with a property "icons" whichs provides a callback for the icon rendering.

By doing so, the User decides the icon-solution: inline-svg, raster-image, icon-font or no icon at all.

*example with fontawesome:*
```
<?=\yii\grid\GridView::widget([
    'columns' => [
        [
            'class' => 'yii\bootstrap4\ActionColumn',
            'icons' => [
                'view' => function ($model) {
                    return '<i class="fas fa-eye"></i>';
                },
                'update' => function ($model) {
                    return '<i class="fas fa-edit"></i>';
                }
                'delete' => function ($model) {
                    return '<i class="fas fa-trash"></i>';
                }
            ],
        ],
    ]
]);?>
```

terabytesoftw:
Completed via https://github.com/yiisoft/yii2/pull/18573

Issue #17602, opened 2019-10-08
EmailValidator with checkDNS=true throws ErrorException on bad domains


batyrmastyr:
### What steps will reproduce the problem?
validate `sbdfbnd@ya2.ru` using EmailValidator with checkDNS=true
`error_reporting=E_ALL` or alike is needed to convert warning to exception in \yii\base\ErrorHandler::handleError
### What is the expected result?
validation returns false
### What do you get instead?
 PHP Warning – yii\base\ErrorException is thrown by `$mxRecords = dns_get_record($domain . '.', DNS_MX)` and `$mxRecords = dns_get_record($domain . '.', DNS_A);`;
That's due to dns_get_record return false and emits Warning on some domains

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.27 (should be since 2.0.11, issue https://github.com/yiisoft/yii2/issues/12714) 
| PHP version      | 7.3.10
| Operating system | Alpine

Suppose it should be fine to rewrite calls to `dns_get_record` this way:
```
try {
     $mxRecords = dns_get_record($domain . '.', DNS_MX);
} catch (\yii\base\ErrorException $exception) {
     return false;
}

**Comment by kamarton** [comment]
https://mxtoolbox.com/SuperTool.aspx?action=mx%3aya2.ru&run=toolpage#

> Sorry, we couldn't find any name servers for 'ya2.ru'

I don't really understand why `checkdnsrr()` doesn't work.

https://github.com/yiisoft/yii2/blob/5efe0bda8f07fc8a16ddd2f86c585a26c7c18785/framework/validators/EmailValidator.php#L114-L126

I don't really understand what the purpose of the current implementation is to get DNS records. I suggests:

```php
protected function isDNSValid($domain) {
  return checkdnsrr($domain . '.', 'MX') || checkdnsrr($domain . '.', 'A');
}
```

kamarton:
```bash
$ php -r 'var_dump(dns_get_record("ya2.ru" . ".", DNS_MX));'
PHP Warning:  dns_get_record(): A temporary server error occurred. in Command line code on line 1
bool(false)
$ php -r 'var_dump(@dns_get_record("ya2.ru" . ".", DNS_MX));'
bool(false)
```

batyrmastyr:
> what the purpose of the current implementation is to get DNS records

`checkdnsrr` always return `true` in Alpine. Yes, it's php bug and it isn't fixed yet (maybe it isn't reported also).

[...]

PR #17604, opened 2019-10-09
fix yiisoft#17602 suppress ErrorException if dns_get_record returned f…

batyrmastyr:
…alse and emitted Warning that was converted to ErrorException

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17602

### DIFF
--- a/framework/validators/EmailValidator.php
+++ b/framework/validators/EmailValidator.php
@@ -8,8 +8,8 @@
 namespace yii\validators;
 
 use Yii;
+use yii\base\ErrorException;
 use yii\base\InvalidConfigException;
-use yii\base\UserException;
 use yii\helpers\Json;
 use yii\web\JsExpression;
 
@@ -112,29 +112,30 @@ class EmailValidator extends Validator
      */
     protected function isDNSValid($domain)
     {
-        if (checkdnsrr($domain . '.', 'MX')) {
-            try {
-                // dns_get_record can return true and emit Warning that may or may not be converted to UserException
-                $mxRecords = dns_get_record($domain . '.', DNS_MX);
-            } catch (UserException $exception) {
-                return false;
-            }
-
-            if ($mxRecords !== false && count($mxRecords) > 0) {
-                return true;
-            }
+        if ($this->hasDNSRecord($domain, true)) {
+            return true;
         }
 
-        if (checkdnsrr($domain . '.', 'A')) {
-            try {
-                $aRecords = dns_get_record($domain . '.', DNS_A);
-            } catch (UserException $exception) {
-                return false;
-            }
+        return $this->hasDNSRecord($domain, false) === true;
+    }
 
-            if ($aRecords !== false && count($aRecords) > 0) {
-                return true;
-            }
+    private function hasDNSRecord($domain, $mx)
+    {
+        $normalizedDomain = $domain . '.';
+        if (!checkdnsrr($normalizedDomain, ($mx ? 'MX' : 'A'))) {
+            return false;
+        }
+
+        try {
+            // dns_get_record can return false and emit Warning that may or may not be converted to ErrorException
+            $mxRecords = dns_get_record($normalizedDomain, ($mx ? DNS_MX : DNS_A));
+        } catch (ErrorException $exception) {
+            Yii::warning($exception->getMessage(). '. Domain: "' . $domain . '"', __METHOD__);
+            return false;
+        }
+
+        if ($mxRecords !== false && count($mxRecords) > 0) {
+            return true;
         }
 
         return false;
--- a/framework/validators/EmailValidator.php
+++ b/framework/validators/EmailValidator.php
@@ -9,6 +9,7 @@ namespace yii\validators;
 
 use Yii;
 use yii\base\InvalidConfigException;
+use yii\base\UserException;
 use yii\helpers\Json;
 use yii\web\JsExpression;
 
@@ -112,14 +113,25 @@ class EmailValidator extends Validator
     protected function isDNSValid($domain)
     {
         if (checkdnsrr($domain . '.', 'MX')) {
-            $mxRecords = dns_get_record($domain . '.', DNS_MX);
+            try {
+                // dns_get_record can return true and emit Warning that may or may not be converted to UserException
+                $mxRecords = dns_get_record($domain . '.', DNS_MX);
+            } catch (UserException $exception) {
+                return false;
+            }
+
             if ($mxRecords !== false && count($mxRecords) > 0) {
                 return true;
             }
         }
 
         if (checkdnsrr($domain . '.', 'A')) {
-            $aRecords = dns_get_record($domain . '.', DNS_A);
+            try {
+                $aRecords = dns_get_record($domain . '.', DNS_A);
+            } catch (UserException $exception) {
+                return false;
+            }
+
             if ($aRecords !== false && count($aRecords) > 0) {
                 return true;
             }
--- a/framework/validators/EmailValidator.php
+++ b/framework/validators/EmailValidator.php
@@ -112,11 +112,7 @@ class EmailValidator extends Validator
      */
     protected function isDNSValid($domain)
     {
-        if ($this->hasDNSRecord($domain, true)) {
-            return true;
-        }
-
-        return $this->hasDNSRecord($domain, false) === true;
+        return $this->hasDNSRecord($domain, true) || $this->hasDNSRecord($domain, false) === true;
     }
 
     private function hasDNSRecord($domain, $mx)
@@ -130,7 +126,6 @@ class EmailValidator extends Validator
             // dns_get_record can return false and emit Warning that may or may not be converted to ErrorException
             $records = dns_get_record($normalizedDomain, ($mx ? DNS_MX : DNS_A));
         } catch (ErrorException $exception) {
-            Yii::warning($exception->getMessage(). '. Domain: "' . $domain . '"', __METHOD__);
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


batyrmastyr:
Maybe it's better to rewrite it this way?:
```
    protected function isDNSValid($domain)
    {
        $result = $this->checkDNSRecord($domain, true);
        if (is_bool($result)) {
            return $result;
        }

        return $this->checkDNSRecord($domain, false) === true;
    }

    /**
     * @param string $domain
     * @param bool $mx
     * @return bool|null
     */
    private function checkDNSRecord($domain, $mx)
    {
        $normalizedDomain = $domain . '.';
        if (!checkdnsrr($normalizedDomain, ($mx ? 'MX' : 'A'))) {
            return false;
        }

        try {
            // dns_get_record can return true and emit Warning that may or may not be converted to UserException
            $mxRecords = dns_get_record($normalizedDomain, ($mx ? DNS_MX : DNS_A));
        } catch (UserException $exception) {
            return false;
        }

        if ($mxRecords !== false && count($mxRecords) > 0) {
            return true;
        }

        return null;
    }
```

framework/validators/EmailValidator.php
```diff
@@ -111,21 +112,29 @@ protected function validateValue($value)
...
+        $normalizedDomain = $domain . '.';
+        if (!checkdnsrr($normalizedDomain, ($mx ? 'MX' : 'A'))) {
+            return false;
+        }
+
+        try {
+            // dns_get_record can return false and emit Warning that may or may not be converted to ErrorException
+            $records = dns_get_record($normalizedDomain, ($mx ? DNS_MX : DNS_A));
+        } catch (ErrorException $exception) {
+            Yii::warning($exception->getMessage(). '. Domain: "' . $domain . '"', __METHOD__);
```
rob006: Why warning? If this exception will be thrown every time when user provide invalid domain, then it shouldn't be a warning - handling garbage input is purpose of this validator, there is nothing special in this situation.

batyrmastyr: Adding warning/error was a suggestion by [samdark](https://github.com/yiisoft/yii2/issues/17602#issuecomment-539504117) and [kamarton](https://github.com/yiisoft/yii2/issues/17602#issuecomment-539885255). 
At first I was going to avoid emitting warnings but after catching `dns_get_record(): Unable to parse DNS data received.` I'm not sure what's better - to discard error silently or to notify developer.

framework/validators/EmailValidator.php
```diff
@@ -111,21 +112,29 @@ protected function validateValue($value)
...
+        if ($this->hasDNSRecord($domain, true)) {
+            return true;
         }
 
-        if (checkdnsrr($domain . '.', 'A')) {
-            $aRecords = dns_get_record($domain . '.', DNS_A);
-            if ($aRecords !== false && count($aRecords) > 0) {
-                return true;
-            }
+        return $this->hasDNSRecord($domain, false) === true;
```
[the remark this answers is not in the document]
batyrmastyr: You're right.

framework/validators/EmailValidator.php
rob006: Logging could be limited to `Yii::debug()`. But I personally would just ignore it - warning is expected here. Handling this warning is internal logic of this validator, it should not pollute application logs - this is nothing that end user should worry about.

framework/validators/EmailValidator.php
```diff
@@ -111,21 +112,24 @@ protected function validateValue($value)
      */
     protected function isDNSValid($domain)
     {
-        if (checkdnsrr($domain . '.', 'MX')) {
-            $mxRecords = dns_get_record($domain . '.', DNS_MX);
-            if ($mxRecords !== false && count($mxRecords) > 0) {
-                return true;
-            }
+        return $this->hasDNSRecord($domain, true) || $this->hasDNSRecord($domain, false) === true;
```
rob006:
```suggestion
        return $this->hasDNSRecord($domain, true) || $this->hasDNSRecord($domain, false);
```

framework/validators/EmailValidator.php
batyrmastyr: Removed warnings as soon as site owner can do nothing to fix malformed replies from external DNS server.

framework/validators/EmailValidator.php
batyrmastyr: oops

batyrmastyr:
Do you want me to add `$this->assertFalse($validator->validate('test@ya2.ru'));` to tests? There are one more trouble with this domain:   
> You should really fix these slow tests (>500ms)...
>  1. 712ms to run yiiunit\framework\validators\EmailValidatorTest:testValidateValueMx
Timings vary from 500 ms to 20 000 ms (what's the hell with that domain I wonder? 0_o).

samdark:
I guess that one just times out.

batyrmastyr:
@samdark yes, most likely. Unfortunately there is no way to set timeout to php's dns-related functions. Adding external libraries like `spatie/dns` is a bit overkill, as I think.
Are you fine with unpredictable tests time or it's better to leave tests as is?

rob006:
Actually if this `dns_get_record()` check is executed only on alpine and fails only for this domain, then I don't think it is worth to add test for it - Travis will not be able to test this scenario anyway.

rob006 (approved):

framework/validators/EmailValidator.php
```diff
@@ -111,21 +112,24 @@ protected function validateValue($value)
...
     {
-        if (checkdnsrr($domain . '.', 'MX')) {
-            $mxRecords = dns_get_record($domain . '.', DNS_MX);
-            if ($mxRecords !== false && count($mxRecords) > 0) {
-                return true;
-            }
+        return $this->hasDNSRecord($domain, true) || $this->hasDNSRecord($domain, false);
+    }
+
+    private function hasDNSRecord($domain, $mx)
```
samdark:
```suggestion
    private function hasDNSRecord($domain, $isMx)
```

framework/validators/EmailValidator.php
```diff
@@ -111,21 +112,24 @@ protected function validateValue($value)
...
+    {
+        $normalizedDomain = $domain . '.';
+        if (!checkdnsrr($normalizedDomain, ($mx ? 'MX' : 'A'))) {
```
samdark:
```suggestion
        if (!checkdnsrr($normalizedDomain, ($isMx ? 'MX' : 'A'))) {
```

framework/validators/EmailValidator.php
```diff
@@ -111,21 +112,24 @@ protected function validateValue($value)
...
         }
 
-        if (checkdnsrr($domain . '.', 'A')) {
-            $aRecords = dns_get_record($domain . '.', DNS_A);
-            if ($aRecords !== false && count($aRecords) > 0) {
-                return true;
-            }
+        try {
+            // dns_get_record can return false and emit Warning that may or may not be converted to ErrorException
+            $records = dns_get_record($normalizedDomain, ($mx ? DNS_MX : DNS_A));
```
samdark:
```suggestion
            $records = dns_get_record($normalizedDomain, ($isMx ? DNS_MX : DNS_A));
```

samdark (approved):
Looks good but please adjust naming a bit, OK?

framework/validators/EmailValidator.php
batyrmastyr: ok

framework/validators/EmailValidator.php
kamarton: > Logging could be limited to Yii::debug()...

@rob006 I think this is a bug that the developer or hostmasters should be aware of. If they cannot resolve it, they have the option to deactivate this feature. If the error is absorbed, the return value is false, which produces unwanted results without feedback.

The [`dns_get_record()`](https://www.php.net/manual/en/function.dns-get-record.php) documentation does not documented an exception at all, so the presence of an exception indicates an error.

samdark:
Merged. Thanks!

framework/validators/EmailValidator.php
kamarton: I suggests: You can directly transfer a DNS record types to it.
```php
private function hasDnsRecord($domain, $dnsRecordTypes = DNS_MX);
```

kamarton: It might be better to get all the types you want in a single step (one DNS request), including the ipv6 address, e.g. `DNS_A | DNS_AAAA | DNS_MX`.

It might be enough to get the name servers and MX: `DNS_NS | DNS_MX`

framework/validators/EmailValidator.php
rob006: `dns_get_record()` check is workaround for Alpine bug - every other environment will use `checkdnsrr()`, which returns `false` without any warning. Ignoring warning here just makes this behavior consistent on all platforms.

> which produces unwanted results without feedback

Returning `false` is expected result for domain with invalid DNS zone settings - it is unreachable anyway.

batyrmastyr:
@rob006 
> is executed only on alpine and fails only for this domain

Unfortunately, no. In fact, isDNSValid return true only if `dns_get_record` returned some records (on all platforms). So `checkdnsrr` is used only to return false for bad domains on glibc-based systems.
De facto, the main reason to keep `checkdnsrr` is possibility of false-negatives by `dns_get_record` on glibc-based systems.

Detection of glibc vs musl is quite tricky, but is possible to some extent: 
1) Parse phpinfo, extract configure command, search `'./configure'  '--build=x86_64-linux-musl' '--with-config... `. Doesn't work for php 7.0 packages from dotdeb.org because configure command is stripped from output. Expect the same on shared hostings for security reasons.
2) resort to several `exec`'s as in [lovell/detect-libc](https://github.com/lovell/detect-libc/blob/master/lib/detect-libc.js). Not 100% accurate and end up in heavy filesystem scans.

samdark:
That's too complicated. Let's stick with current solution for now.

PR #17606, opened 2019-10-10
Proposal: do not publish dummy asset bundle

onmotion:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | 

**Proposal:**
Change the dummy bundle to a real dummy instance AssetBundle

**The reason**
If used dynamic initializer in some AssetBundle, for example, to initiate js like so
```php
// yii\web\AssetBundle 
public function init()
    {
        $this->sourcePath = '@vendor/webcreate/jquery-ias/src';
        $this->js = [
            'callbacks.js',
        ];
    }
```

If disable whole assetManager bundles `\Yii::$app->assetManager->bundles = false;` expected in accordance with docs *>>[If a value is false, it means the corresponding asset bundle is disabled and getBundle() should return null.](https://www.yiiframework.com/doc/api/2.0/yii-web-assetmanager#$bundles-detail)* But in this case, **$js** will be overridden and assets will still be registered, therefore. Plus it requires unnecessary resource allocation.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -5,7 +5,7 @@ Yii Framework 2 Change Log
 ------------------------
 
 - Bug #17573: `EmailValidator` with `checkDNS=true` throws `ErrorException` on bad domains on Alpine (batyrmastyr)
-
+- Bug #17606: Fix error in `AssetBundle` when a disabled bundle with custom init() was still published
 
 2.0.28 October 08, 2019
 -----------------------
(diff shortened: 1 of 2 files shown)


samdark (changes requested):
That would break dependency resolving for the case when bundles are disabled one by one: https://github.com/yiisoft/yii2/blob/master/framework/web/AssetManager.php#L243

samdark (approved):

samdark:
@onmotion changelog?

samdark:
Merged. Thank you!

PR #17609, opened 2019-10-14
(fix) Widget::getId() generates same identifiers when ajax request

zalatov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | 

---

EN: Widget::getId() generates same identifiers when ajax request, because counter was reset to zero every request. This creates conflicts when JavaScript/jQuery linking to widgets identifiers, bu they already exists on page.

RU: Widget::getId() генерирует одинаковые идентификаторы при AJAX-запросах, так как при каждом запросе счётчик сбрасывается в ноль. Это создаёт конфликты, когда JavaScript/jQuery завязывается на идентификаторы виджетов, но они уже есть на странице.

### DIFF
--- a/framework/base/Widget.php
+++ b/framework/base/Widget.php
@@ -161,7 +161,7 @@ class Widget extends Component implements ViewContextInterface
     public function getId($autoGenerate = true)
     {
         if ($autoGenerate && $this->_id === null) {
-            $this->_id = static::$autoIdPrefix . static::$counter++;
+            $this->_id = static::$autoIdPrefix . substr(md5(random_bytes(5)), 0, 8) . static::$counter++;
         }
 
         return $this->_id;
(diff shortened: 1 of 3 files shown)


samdark:
The issue is known. In case of asynchronous calls you have to set IDs explicitly. We can not merge the pull request into Yii 2 because that would break too many applications relying on sequences in auto-generated IDs (I know it's not a good idea in first place but there are many).

zalatov:
So, if I has ajax-page with hundred widgets (buttons, sliders, etc), I need to specify ids all of that?
Why do not mark this bugfix as breaking BC?

zalatov:
I think that rely to an auto-generated ID is a bad development style.
So, such developers will have a good chance to fix this.

samdark:
> Why do not mark this bugfix as breaking BC?

Because Yii 2.0 doesn't accept breaking changes anymore (for about 2 years) except security or PHP 7 compatibility.

> I think that rely to an auto-generated (!) ID is a bad development style.

It certainly is but that's how it is any many many projects are to be broken if the change is applied.

PR #17610, opened 2019-10-14
(fix) BaseFileHelper::createDirectory() now catch Throwable after cat…

zalatov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | 

---

EN:  
BaseFileHelper::createDirectory() catching mkdir() error via @ and error_get_last() because when it called via __toString method ErrorHandler marks file exists warning as fatal error and PHP exit.

RU:
BaseFileHelper::createDirectory() ловит ошибки mkdir() через @ и error_get_last(), потому что когда функция вызвана из метода __toString, то ErrorHandler отмечает предупреждение file exists как фатальную ошибку, и PHP завершает работу.

### DIFF
--- a/framework/helpers/BaseFileHelper.php
+++ b/framework/helpers/BaseFileHelper.php
@@ -636,13 +636,10 @@ class BaseFileHelper
         if ($recursive && !is_dir($parentDir) && $parentDir !== $path) {
             static::createDirectory($parentDir, $mode, true);
         }
-        try {
-            if (!mkdir($path, $mode)) {
-                return false;
-            }
-        } catch (\Exception $e) {
-            if (!is_dir($path)) {// https://github.com/yiisoft/yii2/issues/9288
-                throw new \yii\base\Exception("Failed to create directory \"$path\": " . $e->getMessage(), $e->getCode(), $e);
+        if (false === @mkdir($path, $mode)) {
+            $error = error_get_last();// https://github.com/yiisoft/yii2/issues/9288
+            if (false === is_dir($path)) {
+                throw new \yii\base\Exception("Failed to create directory \"$path\": " . $error['message'], $error['type']);
             }
         }
         try {


zalatov:
@samdark What unit test is needed for my changes? I only added catch Throwable...

samdark:
Anything that causes this throwable.

zalatov:
I still has error... And trying to catch Exception and Throwable will not work.
```
yii\base\ErrorException: mkdir(): File exists in /home/web/site/vendor/yiisoft/yii2/helpers/BaseFileHelper.php:640
```

This fail when many requests per second.

If I append `@` before mkdir, works fine.
Later I'll do some changes in my request.

ailmanki:
https://github.com/kalessil/phpinspectionsea/blob/master/docs/probable-bugs.md#mkdir-race-condition

rob006:
@zalatov That is weird. `} catch (\Exception $e) {` should catch `yii\base\ErrorException`.

zalatov:
@rob006 yes, and this is confuse me, that exception is not catched. Php is down on this row. PHP 7.2, CentOS

On local Ubuntu 18 this exception is catched and throws yii specific exception in lines below.

ailmanki:
Its probably a race condition.
The error says: "mkdir(): File exists in .....".
So when mkdir runs, the folder exists already. Then the Exception is thrown, which lands in the catch block. There it checks if the folder does not exist, if that is the case it throws an appropriate exception on its own.

Now the folder does exist already. 

copy paster from https://github.com/kalessil/phpinspectionsea/blob/master/docs/probable-bugs.md#mkdir-race-condition
### 'mkdir(...)' race condition
This issue is difficult to reproduce, as any concurrency-related issues are. It appears when several processes are attempting to create a directory which does not yet exist. Specifically, when one process is between is_dir() and mkdir() after another process has already managed to create the directory.

```
    /* before */
    if (!is_dir($directory)) {
        mkdir($directory);
    }
    
    /* after */
    if (! is_dir($directory) && ! mkdir($directory) && ! is_dir($directory)) {
        throw new \RuntimeException(sprintf('Directory "%s" was not created', $directory));
    }
```

samdark:
> That is weird. } catch (\Exception $e) { should catch yii\base\ErrorException.

@rob006 that's PHP 7. https://www.php.net/manual/en/class.error.php is being thrown.

rob006:
@samdark `Error` is thrown only on actual error. Warning works the same as on PHP 5 - it is caught by Yii error handler and converted to `yii\base\ErrorException`.

samdark:
You're right https://3v4l.org/pB6tL

ailmanki:
Oh wow, Warnings are not catched, 
https://stackoverflow.com/questions/1241728/can-i-try-catch-a-warning

samdark:
Well, we're converting warnings to exceptions by default so these should be caught...

samdark:
I've just tried myself the following in controller:

```php
try {
    mkdir(sys_get_temp_dir() . '/test');
    mkdir(sys_get_temp_dir() . '/test');
} catch (\yii\base\ErrorException $e) {
    // do nothing
}
```

Works.

samdark:
@ailmanki can you do it as well?

ailmanki:
@samdark yes works for me too, even the other code works fine, I just add both. In second case it goes int \Exception catch block.

```
try {
    mkdir(sys_get_temp_dir() . '/test',755);
    mkdir(sys_get_temp_dir() . '/test',755);
} catch (\yii\base\ErrorException $e) {
}
try {
    if (!mkdir(sys_get_temp_dir() . '/test',755)) {
    }
} catch (\Exception $e) {

} catch (\Throwable $e) {

}
```

samdark:
No idea. If https://github.com/yiisoft/yii2/pull/17610#issuecomment-545291263 was at the server then error handler was called and it converted warning to exception successfully.

zalatov:
I try to debug on production tomorrow. I think, error handler think that this exception is final and exit (some code exists in error handler method).

Code with @mkdir and get_last_error() fix this.

zalatov:
I catch this error when my class has __toString implementation.
For example, I have class EmailContent($message, $user) which autoconvert to string when send via mailer.
In ErrorHandler:219 has condition for check if function is `__toString`.
This exit PHP code as final error.

rob006:
> In ErrorHandler:219 has condition for check if function is `__toString`.

This looks like a bug in error handler. :/

samdark:
`__toString()` is a special case that would stop being special in PHP 7.4: https://wiki.php.net/rfc/tostring_exceptions

samdark:
But yes, it would not hurt re-checking.

samdark:
@zalatov I've checked your solution and it doesn't work for `__toString()` case:

```
__toString() must not throw an exception, caught yii\base\Exception: Failed to create directory
```

I don't think we can and should fix it within the framework since that would mean not throwing exceptions at all. Everything could be used in `__toString()`.

rob006:
@samdark This still looks like an error handler bug. It tries to be too smart and blindly assumes that no exception will be caught in `__toString()` call, so it breaks its own contract and does not convert warning to exception. This may be a problem for every piece of code which relies on conversion of warnings/notices to exceptions.

samdark:
Because we can't throw in __toString() thus we can't convert anything to exception: https://3v4l.org/vBsLC

rob006:
@samdark We can throw in `__toString()`, you just need to catch exception in order to prevent fatal error: https://3v4l.org/BSD9G 
Right now error handler assumes that nobody will do this (even if it is standard practice in framework itself and Yii relies on converting warnings to exceptions in many places).

samdark:
@rob006 do you have some time to do the fix?

rob006:
I'll take a look at this.

PR #17611, opened 2019-10-14
(fix) Unterminated JS statement in views/ErrorHandler/exception.php

zalatov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | 

---

EN:
Added semicolon after javascript declaration because this can cause error for some browsers.

RU:
Добавлена точка с запятой в присвоении javascript-функции, потому как это может вызывать ошибку в некоторых браузерах.

---

https://stackoverflow.com/questions/11978698/do-we-need-a-semicolon-after-function-declaration
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/function#Syntax

### DIFF
--- a/framework/views/errorHandler/exception.php
+++ b/framework/views/errorHandler/exception.php
@@ -531,8 +531,8 @@ window.onload = function() {
 };
 
     // Highlight lines that have text in them but still support text selection:
-    document.onmousedown = function() { document.getElementsByTagName('body')[0].classList.add('mousedown') };
-    document.onmouseup = function() { document.getElementsByTagName('body')[0].classList.remove('mousedown') };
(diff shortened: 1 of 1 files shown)


samdark (approved):
Looks OK. Would you please add a line for changelog?

samdark:
Also... which browser has errors?

zalatov:
I don't remember what browser is - this fix was mad in our project year ago (or older).

Now I test in online-browser-shots without semicolon - all is ok.
So... this is stylish modification, if you want, you could close this request without acception.

PS: May be this fix was for JSLint, PhpStorm also mark this as warning.

samdark:
PhpStorm doesn't complain about it but let's have it. Thanks.

PR #17616, opened 2019-10-18
Update start-forms.md

BXGo:
'siteController' is a file. Can not be translate to site "控制器".

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | /❌
| Breaks BC?    | /❌
| Tests pass?   | /❌


samdark:
It's good as is. In previous chapters it is explained about controller ids so there should be no confusion.

PR #17617, opened 2019-10-18
Update start-forms.md

BXGo:
'siteController' is a php file. 
It's better to say the name of the file, in case somebody mistake to think it is create an entry action under view/site.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/
| New feature?  | /❌
| Breaks BC?    | /❌
| Tests pass?   | /❌


samdark:
It's good as is. In previous chapters it is explained about controller ids so there should be no confusion.

PR #17625, opened 2019-10-21
Add support for boolean data attributes to Html::renderTagAttributes()

brandonkelly:
Currently the following produce inconsistent results:

```php
Html::renderTagAttributes(['data-foo' => true]);
// output: ' data-foo'

Html::renderTagAttributes(['data' => ['foo' => true]]);
// output: ' data-foo="1"'
```

This PR causes boolean `data` sub-keys to be treated the same way top-level boolean keys are treated: `true` will cause the attribute to be added without a value; `false` will prevent the attribute from being added at all.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -1955,6 +1955,10 @@ class BaseHtml
                     foreach ($value as $n => $v) {
                         if (is_array($v)) {
                             $html .= " $name-$n='" . Json::htmlEncode($v) . "'";
+                        } else if (is_bool($v)) {
+                            if ($v) {
+                                $html .= " $name-$n";
+                            }
                         } else {
                             $html .= " $name-$n=\"" . static::encode($v) . '"';
                         }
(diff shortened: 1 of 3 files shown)


samdark (approved):

brandonkelly (commented):
This will break the `false` test, which should just prevent the attribute from being added at all (like other `false` attributes).

framework/helpers/BaseHtml.php
```diff
@@ -1955,10 +1955,8 @@ public static function renderTagAttributes($attributes)
                     foreach ($value as $n => $v) {
                         if (is_array($v)) {
                             $html .= " $name-$n='" . Json::htmlEncode($v) . "'";
-                        } else if (is_bool($v)) {
-                            if ($v) {
-                                $html .= " $name-$n";
-                            }
+                        } elseif (is_bool($v) && $v) {
+                            $html .= " $name-$n";
```
brandonkelly: This will break the `false` test, which should just prevent the attribute from being added at all (like other `false` attributes).

samdark: Ah, right.

samdark:
Merged. Thanks!

Issue #17572, opened 2019-09-25
fix #6226 symlinking multi threaded  ERROR symlink(): Cannot create symlink, error code(183)


Eloy-Gonzalez:
### What steps will reproduce the problem?
- Try using Webserver Nginx version 1.10.1 on Windows 7 32 Bits  
- Clean folder \app\web\assets and reload browser with your app developed

### What is the expected result?
- Generate ERROR symlink(): Cannot create symlink, error code(183)
### What do you get instead?
Fix error adding symbol "@" on this line 489 and 534 on file \yii\web\assetManager.php

Before::
```php
 protected function publishFile($src)
    {
        $dir = $this->hash($src);
        $fileName = basename($src);
        $dstDir = $this->basePath . DIRECTORY_SEPARATOR . $dir;
        $dstFile = $dstDir . DIRECTORY_SEPARATOR . $fileName;

        if (!is_dir($dstDir)) {
            FileHelper::createDirectory($dstDir, $this->dirMode, true);
        }

        if ($this->linkAssets) {
            if (!is_file($dstFile)) {
                try { // fix #6226 symlinking multi threaded
                    symlink($src, $dstFile);
                } catch (\Exception $e) {
                    if (!is_file($dstFile)) {
                        throw $e;
                    }
                }
            }
        } elseif (@filemtime($dstFile) < @filemtime($src)) {
            copy($src, $dstFile);
            if ($this->fileMode !== null) {
                @chmod($dstFile, $this->fileMode);
            }
        }

        return [$dstFile, $this->baseUrl . "/$dir/$fileName"];
    }
```
After::
```php
 protected function publishFile($src)
    {
        $dir = $this->hash($src);
        $fileName = basename($src);
        $dstDir = $this->basePath . DIRECTORY_SEPARATOR . $dir;
        $dstFile = $dstDir . DIRECTORY_SEPARATOR . $fileName;

        if (!is_dir($dstDir)) {
            FileHelper::createDirectory($dstDir, $this->dirMode, true);
        }

        if ($this->linkAssets) {
            if (!is_file($dstFile)) {
                try { // fix #6226 symlinking multi threaded
                    @symlink($src, $dstFile);
                } catch (\Exception $e) {
                    if (!is_file($dstFile)) {
                        throw $e;
                    }
                }
            }
        } elseif (@filemtime($dstFile) < @filemtime($src)) {
            copy($src, $dstFile);
            if ($this->fileMode !== null) {
[...]

PR #17638, opened 2019-10-24
// ERROR symlink(): Cannot create symlink, error code(183)

### LINKED ISSUE #6226: symlink(): File exists

Eloy-Gonzalez:
// fix #6226 symlinking multi threaded ON LINE 489 in file \yii\web\AssetManager.php
// Fixed and tested on Nginx on Windows 32 Bits

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️ 
| New feature?  |❌
| Breaks BC?    | ❌
| Tests pass?   | ❌

### DIFF
--- a/framework/web/AssetManager.php
+++ b/framework/web/AssetManager.php
@@ -486,7 +486,7 @@ class AssetManager extends Component
         if ($this->linkAssets) {
             if (!is_file($dstFile)) {
                 try { // fix #6226 symlinking multi threaded
-                    symlink($src, $dstFile);
+                    @symlink($src, $dstFile);
                 } catch (\Exception $e) {
                     if (!is_file($dstFile)) {
                         throw $e;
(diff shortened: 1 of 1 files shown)


samdark:
Wasn't it already fixed in https://github.com/yiisoft/yii2/pull/14905?

rob006:
@samdark  it is the same as https://github.com/yiisoft/yii2/issues/17572. But this PR is incorrect IMO - this issue should be fixed by `clearstatcache()` (https://github.com/yiisoft/yii2/issues/17572#issuecomment-536503532).

samdark:
Agree. @Eloy-Gonzalez thanks for contributing. Could you try what @rob006 proposed and check if it works?

PR #17642, opened 2019-10-28
allow `getFirstError` attribute is null.

eecjimmy:
| Q             | A
| ------------- | ---
| Is bugfix?     ❌
| New feature?  ✔️
| Breaks BC?   ✔️
| Tests pass?   ✔️
| Fixed issues  allow `getFirstError` attribute is null.


eecjimmy:
When I use `Model` to process some business logic, sometimes I don't care which `attribute` error occurs, and only need to return an error information, then I can use ` $model->getFirstError()`.Returns an error directly.

```php
$model->getFirstError();
```

@samdark

githubjeka:
> When I use Model to process some business logic, sometimes I don't care which attribute error occurs, and only need to return an error information,

```php
print_r($model->getErrors(), true)
```

samdark:
@eecjimmy thanks for contributing but your use case seems to be quite specific. Moreover, it is a good practice to display all errors at once as @githubjeka mentioned (not with print_r, of course).

PR #17647, opened 2019-10-29
add formatter->asText cut string feature

iridance:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | no

This is a commonly used feature, I found that yii2 is not available, I think it would be helpful to add it.

#### With formatter component
```php
$intro = Yii::$app->formatter->asText($model->description, ['width' => 20]);
```
#### With GridView widget
```php
$widget = GridView::begin([
     'filterModel' => $searchModel,
     'columns' => [
          ['class' => 'yii\grid\SerialColumn'],
          'id',
          [
               'attribute' => 'title',
               'format' => ['text', ['width' => 20]],
          ],
          [
               'attribute' => 'description',
               'format' => ['text', ['width' => 40, 'suffixSymbol' => '---']]
          ],
          'link:url',
          'weight',
     ],
]);
```

Annotation 2019-10-29 204308

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -477,14 +477,35 @@ class Formatter extends Component
     }
 
     /**
+     * Cut the value as short text.
+     *
+     * @param [type] $value the value to be cut.
+     * @param array $options string cut options.
+     * @return string the cut string result.
+     */
+    public function cut($value, $options = [])
+    {
+        $options = array_merge(
+            ['width' => 0, 'suffixSymbol' => '...'],
+            $options
+        );
+        return mb_strimwidth($value, 0, $options['width'], $options['suffixSymbol']);
+    }
+    
+    /**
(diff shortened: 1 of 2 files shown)


samdark:
Thanks for contributing. 

There are multiple issues with the pull request:

1. It seems it doesn't work with older PHP versions: https://travis-ci.org/yiisoft/yii2/jobs/604423016#L976 There is https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseStringHelper.php#L105 for the purpose that is proved to work well with unicode.
2. Yii 2.0 is in feature freeze mode so adding new features, especially ones that break backwards compatibility (change existing public or protected interfaces) isn't desired.

As a workaround, you can either extend formatter adding it. Or use a closure with `StringHelper::truncate()`.

PR #17607, opened 2019-10-11
Added Yii 3 DI compatibility

hiqsol:
- Allowed `__class` and `__construct()` in configurations:
    - `'__class' => SomeClass::class`
    - `'__construct()' => ['ConstructorArg1', 'ConstructorArg2']`
- Allowed wider use of `Instance::of()`:
    - in definitions: `'SomeInterface' => Instance::of('SomeService')`
    - in get: `$container->get(Instance::of('SomeInterface')`
- Allowed static call DI definition:
    - `'SomeInterface' => [SomeFactory::class, 'createMethod']`
- Added more tests

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

PR #17670, opened 2019-11-13
Fix wrong overriden application core component configuration when uses Yii3 style

hyuzea:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | 

If you overrides core component config using `__class` in it, you may get an error message: `setting unknown prop __class`. It because the config at this time become like this
```
[
  '__class' => 'MyRequest',
  'prop1' => '1234',
  'class' => 'yii\web\Request',
]
```


samdark:
Related to #17607

samdark:
Fixed, added changelog line and merged. Thanks for reporting.

Issue #17665, opened 2019-11-12
Support Forwarded header (RFC 7239 section 4)


mikk150:
### What steps will reproduce the problem?
Make request through [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/)
### What is the expected result?
Yii should get user IP, hostname and whether it is HTTP or HTTPS from Forwarded header
### What do you get instead?
Yii thinks that Ingress was requester, it is HTTP always

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.29
| PHP version      | 7.2.24
| Operating system | Docker 7.2-fpm-alpine

https://tools.ietf.org/html/rfc7239#section-4
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

kamarton:
See https://www.yiiframework.com/doc/guide/2.0/en/runtime-requests#trusted-proxies

mikk150:
> See https://www.yiiframework.com/doc/guide/2.0/en/runtime-requests#trusted-proxies

I know that, but it does not help

samdark:
That is definitely something we want in Yii 3. Not sure it could be considered a bug not supporting it...

mikk150:
> That is definitely something we want in Yii 3. Not sure it could be considered a bug not supporting it...

what if I told you that I am already trying to implement it

samdark:
@mikk150 for Yii 3?

mikk150:
> @mikk150 for Yii 3?

No, for Yii2, but it can be copied to that middleware

samdark:
OK. If there will be no BC breaks then we'd consider it for Yii 2 as well.

PR #17676, opened 2019-11-20
Implement RFC 7239 `Forwarded` header parsing

mikk150:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17665


mikk150:
It cannot be implemented 100%, because PHP is stupid, It does not allow to decode multiple same headers. If I do this, it does not work...
```
GET / HTTP/1.1
Forwarded: for=10.0.0.1,for=20.30.40.50;host=yiiframework.com,host=awesome.proxy.com;proto=https,proto=http
Forwarded: for=10.30.20.10;host=second.awesome.proxy.com;proto=http
```

samdark:
Yes. Seems that it's not possible to get multiple same-named headers at all.

mikk150:
Aaand it actually seems that it is doable, as https://tools.ietf.org/html/rfc2616#section-4.2 states that what I assumed was wrong(you can and SHOULD be able to concatenate headers with comma). Apache does it correctly, but FPM is broken for now

samdark:
FPM is the most used setup for PHP nowadays, I think.

mikk150:
> FPM is the most used setup for PHP nowadays, I think.

No, apache rules it(not that its the best), you are forgetting shared hosting

kamarton:
Web server applications are usually able to handle concatenation: https://stackoverflow.com/a/47051518/6856708.

mikk150:
> Web server applications are usually able to handle concatenation: https://stackoverflow.com/a/47051518/6856708.

I still made Bug report for PHP, as it not giving all headers is wrong https://bugs.php.net/bug.php?id=78844

samdark:
Voted.

kamarton:
> I still made Bug report for PHP, as it not giving all headers is wrong https://bugs.php.net/bug.php?id=78844

I like it, but I don't think it will change. Due to the following BC issues:
- Webserver and fastcgi communicate with variables (key + value), for example HTTP_* in $_SERVER variable. As a result, the last of the duplicate headers is the winner.
- Based on the previous point, any development in the case must involve the addition development of web servers.

mikk150:
> > I still made Bug report for PHP, as it not giving all headers is wrong https://bugs.php.net/bug.php?id=78844
> 
> I like it, but I don't think it will change. Due to the following BC issues:
> 
> * Webserver and fastcgi communicate with variables (key + value), for example HTTP_* in $_SERVER variable. As a result, the last of the duplicate headers is the winner.
> * Based on the previous point, any development in the case must involve the addition development of web servers.

which also means current solution (X-forwarded-for) is also broken, so no BC. Also you need to trust proxy in order for this to work anyway. You betta know when you need to enable support for this

kamarton:
> Also you need to trust proxy in order for this to work anyway. You betta know when you need to enable support for this

This is absolutely true. Most systems will concatenate this, but in fact it is not mandatory for `X-Forwarded-For`. I think that the `Forwarded` header is safe to assume (comma joined), and if it is not, then the pre-PHP software should take care of it.

mikk150:
> This is absolutely true. Most systems will concatenate this, but in fact it is not mandatory for `X-Forwarded-For`. I think that the `Forwarded` header is safe to assume (comma joined), and if it is not, then the pre-PHP software should take care of it.

https://tools.ietf.org/html/rfc2616#section-4.2 mandates that

> It MUST be possible to combine the multiple header fields into one
   "field-name: field-value" pair, without changing the semantics of the
   message, by appending each subsequent field-value to the first, each
   separated by a comma.

which means `X-Forwarded-For` behaves exactly the same way as `Forwarded` header

samdark:
@mikk150 please ping me when it's ready for review.

mikk150:
> @mikk150 please ping me when it's ready for review.

@samdark what is your public IP address so I can ping you...

Yes it is ready for code review

framework/web/Request.php
```diff
@@ -1809,4 +1829,61 @@ private function validateCsrfTokenInternal($clientSuppliedToken, $trueToken)
 
         return $security->compareString($security->unmaskToken($clientSuppliedToken), $security->unmaskToken($trueToken));
     }
+
+    /**
+     * Gets first `Forwarded` header value for token
+     *
+     * @param string $token Header token
+     *
+     * @return array
```
samdark: Is it correct?

mikk150: ah, not any more

kamarton (commented):
> https://tools.ietf.org/html/rfc2616#section-4.2 mandates that

PHP has not HTTP input interface, requires web server application. As a result, PHP does not need to support this feature.
Web server applications also support concatenation, as shown [here](https://stackoverflow.com/a/47051518/6856708.).

framework/web/Request.php
```diff
@@ -1809,4 +1829,61 @@ private function validateCsrfTokenInternal($clientSuppliedToken, $trueToken)
...
+     */
+    protected function getForwardedHeaderPart($token)
+    {
+        $token = strtolower($token);
+        if ($decodedHeader = $this->decodedForwardedHeader()) {
+            /**
+             * First one is always correct, because proxy CAN append to last
```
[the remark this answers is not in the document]
samdark: Indeed, the header could be easily spoofed and it depends very much on which one we choose as safe one.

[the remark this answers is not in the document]
mikk150: how could you detect spoofing?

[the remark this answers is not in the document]
samdark: You can not. That's why by default these headers should not be trusted at all. When trusted server is specified we can trust IPs right after it but not before it.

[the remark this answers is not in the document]
kamarton: Going backwards through the IP list, you need to check that it matches `$trustedIps`. The first not match will be the client IP.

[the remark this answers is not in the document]
mikk150: > Going backwards through the IP list, you need to check that it matches `$trustedIps`. The first not match will be the client IP.

So if I have 
```
Forwarded: for=<CLIENT>,for=<NOT_TRUSTED_PROXY>, for=<TRUSTED_PROXY>
```

I would tell that `<NOT_TRUSTED_PROXY>` is client IP? even tho it is technically incorrect?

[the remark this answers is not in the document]
kamarton: Developers/system maintainer need to find a way to keep it trusted (eg find IP ranges), if this is not possible then it is on the edge of the trusted network and should be considered as a client.
eg. a user coming from a TOR network, private VPN, or behind a transparent proxy cannot be accurately identified by any means, so these cases must be ignored.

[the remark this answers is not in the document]
samdark: If external IP is important to you (for example, if you're limiting number of requests) then yes. If it's not so important, you can choose to trust any host. But then anyone can fake his IP by simply passing an HTTP header.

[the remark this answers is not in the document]
mikk150: yep, I just wanted to ask if this is what you want. I will do it, no worries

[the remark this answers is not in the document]
samdark: Yes. That's the only thing we can rely on for tasks like checking IP for security reasons.

framework/web/Request.php
```diff
@@ -1142,6 +1152,16 @@ public function getUserAgent()
      * @since 2.0.28
      */
     protected function getUserIpFromIpHeaders() {
+        if (($ip = $this->getForwardedHeaderPart('for')) !== null) {
```
kamarton: check only, if the `forwarded` header is contains in `$ipHeaders`.

mikk150: it will break `protected function getUserIpFromIpHeader($ips)`

because Forwarded header is more complicated than just list of IPs

kamarton: However, if you ignore `$ipHeaders`, anyone can inject the IP address by submitting the `Forwarded` header.

```http
GET / HTTP/1.1
x-forwarded-for: 8.8.8.8, 9.9.9.9
forwarded: for=127.0.0.1             <---- injected
```

framework/web/Request.php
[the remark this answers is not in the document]
mikk150: this should be done

framework/web/Request.php
kamarton: @samdark maybe if create a separate configuration parameter?
```php
[
  'class' => Request::class,
  'ipListRfc7239' => ['forwarded'],
]
```

mikk150: I could filter out `Forwarded` from `getUserIpFromIpHeader()` but that wouldn't be nice

mikk150: OR I could look every time in `$secureHeaders` array, that `Forwarded` is there
That way user can just say that `Forwarded` is not secure, by removing that from `$secureHeaders`

kamarton: > OR I could look every time in $secureHeaders array, that Forwarded is there

@samdark how about this?

samdark: Yes. That's a good thing to do. It should be configured explicitly.

mikk150: Aaaand we should be fine

samdark (approved):

samdark:
Merged. Thank you!

PR #17679, opened 2019-11-21
Add oci command

243083df:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | 

Fix "ORA-01461: can bind a LONG value only for insert into a LONG column".
This happens if insert big string in CLOB


243083df:
I don't know how to create unit test for this case.

samdark:
How to reproduce the issue then?

243083df:
To reproduce you need to create table in oracle with cloumn wich has CLOB field type, and insert into this column big string (>4000 bytes)

framework/db/oci/Command.php
```diff
@@ -0,0 +1,25 @@
...
+class Command extends \yii\db\Command
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function bindPendingParams()
+    {
+        foreach ($this->_pendingParams as $name => $value) {
+            if($value[1] === \PDO::PARAM_STR) {
+                $clonedValue = clone $value[0];
```
samdark: It looks suspicious. `clone` works with objects only so the statement would fail if I'll do

```php
$command->bindValues(['myParam' => 'myValue']);
```

framework/db/oci/Command.php
```diff
@@ -0,0 +1,25 @@
...
+                $this->pdoStatement->bindParam($name, $clonedValue, $value[1], strlen($clonedValue));
```
samdark: `strlen()` would error with "Warning: strlen() expects parameter 1 to be string, object given" since `$clonedValue` should be an object.

samdark (changes requested):
Thanks for the contribution but it doesn't look like the code in the pull request could work.

samdark:
@243083df please confirm that you've tested the code locally and it fixes the issue.

243083df:
@samdark Tested

243083df:
@samdark Something else?

samdark:
Waiting till someone with Oracle can confirm the fix is fine. I'm currently don't have it in my environment.

vinpel:
I had the same error on the table "log" with long data.
The proposed fix make this error gone

samdark:
@vinpel in order for it to be merged, we need unit tests for the case. Do you have some time for contribution? Would be really helpful since I don't have Oracle environment set up currently.

vinpel:
i can't make it work to insert data into the log file, still the error ORA-01745 ...
LEVEL seem to be a reserved word, but modifiy the file referenced in the commit and seting random name didn't work neither

vinpel:
i added unit test in this pr :
https://github.com/yiisoft/yii2/pull/17854

samdark:
Replaced by #17854

Issue #17661, opened 2019-11-07
using NOT IN as operator with an array as value 


buttflattery:
### What steps will reproduce the problem?

Create a query like below 

```
        $query = new Query();
        return $query->select('*')
            ->from( 'release rls')
            ->where(
                [
                    'AND',
                    ['=', 'rls.list_id', $listId],
                    ['not in', 'rls.subscriber_id', $campaignSubscribers],
                ]
            )
            ->createCommand()->rawSql;
```

### What is the expected result?
In the above query the `$campaignSubscribers` is an array, if that array has some values then all is good but if the array is empty then the query is generated like below .

```
SELECT * FROM `release` `rls` WHERE (`rls`.`list_id` = 1) AND (`rls`.`subscriber_id` <>null)
```
which never yields the correct results and returns empty array although there are rows that should be returned, it should generate the last part as `(rls.subscriber_id IS NOT null)` rather than `(rls.subscriber_id <>null)`

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.27
| PHP version      |  7.2
| Operating system | ubuntu

samdark:
That's certainly incorrect result. Do you have acwsy to fix it? Want to work on a pull request?

strychu:
Could not reproduce this. InConditionBuilder build method returns empty string when values property of passed expression is an empty array.

    ``` $sqlValues = $this->buildValues($expression, $values, $params);
        if (empty($sqlValues)) {
            return $operator === 'IN' ? '0=1' : '';
        } ```

samdark:
@buttflattery is there anything special that is required to reproduce it?

buttflattery:
@samdark i just looked up into my old source code where i had to change the code base and it looks like the reason is not just an empty array but, if the array is like

```
$campaignSubscribers=[null];
```

rather than 

```
$campaignSubscribers=[];
```

i reverted the code and var dumped the array to re-check and if the array has null then it produces the query as mentioned in the issue above.  Not sure now if it should be considered as an issue or not .

strychu:
imo if there's null value in array it should generate IS NOT NULL, now it allows to pass null in array and produces unexpected result

buttflattery:
@strychu i agree, and to my amusement i just test the scenario with 

$campaignSubscribers=[null,null];

and it generates the condition as `(`rls`.`subscriber_id` NOT IN (NULL, NULL))`

buttflattery:
[...]

PR #17697, opened 2019-11-29
InConditionBuilder proper null handling

hstrychalski:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17661


samdark:
Seems to fail tests: https://travis-ci.org/yiisoft/yii2/jobs/618641372

strychu:
I've fixed tests and added some test cases to condition provider

strychu:
Although i see that build fails for php 7.4

samdark:
That's OK. Same in master.

tests/framework/db/QueryBuilderTest.php
```diff
@@ -1178,6 +1178,18 @@ public function conditionProvider()
 
             [['in', 'id', new TraversableObject([1, 2, 3])], '[[id]] IN (:qp0, :qp1, :qp2)', [':qp0' => 1, ':qp1' => 2, ':qp2' => 3]],
 
+            //in using array objects containing null value
+            [['in', 'id', new TraversableObject([1, null])], '[[id]]=:qp0 AND [[id]] IS NULL', [':qp0' => 1]],
```
samdark: That doesn't look right. `in` is about the value is being in the list. "in X [1,2,3]" transforms into `X IN [1,2,3]` that is equal to `X = 1 OR X = 2 OR X = 3`. Here we have `AND` and I think it should be `OR`.

tests/framework/db/QueryBuilderTest.php
```diff
@@ -1178,6 +1178,18 @@ public function conditionProvider()
...
+            [['in', 'id', new TraversableObject([1, 2, null])], '[[id]] IN (:qp0, :qp1) AND [[id]] IS NULL', [':qp0' => 1, ':qp1' => 2]],
```
samdark: Same here. Looks like it should generate `OR`.

tests/framework/db/QueryBuilderTest.php
```diff
@@ -1178,6 +1178,18 @@ public function conditionProvider()
...
+
+            //not in using array object containing null value
+            [['not in', 'id', new TraversableObject([1, null])], '[[id]]<>:qp0 AND [[id]] IS NOT NULL', [':qp0' => 1]],
```
samdark: These ones are correct.

tests/framework/db/QueryBuilderTest.php
```diff
@@ -1178,6 +1178,18 @@ public function conditionProvider()
...
+            [['not in', 'id', new TraversableObject([1, 2, null])], '[[id]] NOT IN (:qp0, :qp1) AND [[id]] IS NOT NULL', [':qp0' => 1, ':qp1' => 2]],
+
+            //in using array object containing only null value
+            [['in', 'id', new TraversableObject([null])], '[[id]] IS NULL', []],
```
samdark: These ones are correct as well.

samdark (changes requested):

samdark (approved):

samdark:
Merged. Thank you!

Issue #17701, opened 2019-12-02
PHP notice when request param is incompatible with action argument type


brandonkelly:
If a controller action argument specifies an `int` or `float` argument and PHP can’t convert the corresponding request param to that type, a PHP notice will occur.

### What steps will reproduce the problem?

1. Create a controller action with an argument that has an `int` type declaration.
    ```php
    public function actionTest(int $foo) {
        return 'hello';
    }
    ```
2. Send a request to that controller action, and set that param to a non-integer value, such as `foo=oops`.

### What is the expected result?

A `BadRequestHttpException` should be thrown.

### What do you get instead?

A PHP notice will occur:

> A non well formed numeric value encountered

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.30
| PHP version      | 7.3.9
| Operating system | Ubuntu 18.04.1

PR #17702, opened 2019-12-02
Throw BadRequestHttpException when request params can't be bound to controller action

brandonkelly:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17701

### DIFF
--- a/framework/web/Controller.php
+++ b/framework/web/Controller.php
@@ -133,24 +133,22 @@ class Controller extends \yii\base\Controller
                     $params[$name] = (array)$params[$name];
                 } elseif (is_array($params[$name])) {
                     $isValid = false;
-                } elseif (PHP_VERSION_ID >= 70000 && ($type = $param->getType()) !== null) {
-                    if ($type->allowsNull() && ($params[$name] === null || $params[$name] === '')) {
-                        $params[$name] = null;
-                    } elseif ($type->isBuiltin()) {
-                        switch ((string)$type) {
-                            case 'bool':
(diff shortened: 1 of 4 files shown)


framework/web/Controller.php
```diff
@@ -128,15 +128,37 @@ public function bindActionParams($action, $params)
...
                 if ($param->isArray()) {
-                    $args[] = $actionParams[$name] = (array) $params[$name];
-                } elseif (!is_array($params[$name])) {
-                    $args[] = $actionParams[$name] = $params[$name];
-                } else {
+                    $params[$name] = (array)$params[$name];
+                } elseif (is_array($params[$name])) {
+                    $isValid = false;
+                } elseif (PHP_VERSION_ID >= 70000 && ($type = $param->getType()) !== null) {
+                    if ($type->allowsNull() && ($params[$name] === null || $params[$name] === '')) {
```
rob006: Are you sure about converting empty string to `null`? This is probably a BC break.

brandonkelly: Yeah you‘re right. Just pushed another commit that stops doing that. Also no longer doing anything at all with `bool` params, since those weren’t causing PHP errors in the first place. So now the only change in behavior this PR brings is that a `BadRequestHttpException` will be thrown in cases where a PHP error was going to occur otherwise.

schmunk42 (approved):

samdark (approved):

samdark:
Merged. Thank you!

brandonkelly:
Awesome, thanks!

Issue #17694, opened 2019-11-28
Javascript in error page


Fr1ar:
Let's assume I want to render a simple view:
```
<?php
$this->registerJs('alert("Hello!")');

// And an unhandled exception is thrown during the view (or layout) rendering:
throw new Exception('Opps');
```

If YII_DEBUG is false than I see the error page rendered by yii\web\ErrorAction, but I also see a "Hello!" message, which means that JS code was included into error page. Is it supposed to be like that?

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.27
| PHP version      | 7.3.11
| Operating system |CentOS 7

alex-code:
Maybe the assets should be cleared before the error view is rendered?
Might cause an issue if any of those assets are needed to render the layout though.

bizley:
This is similar to #16101 which I thought I've fixed a year ago...

alex-code:
@bizley Looks like this occurs if you use the `ErrorAction` option so it's rendered differently.
```php
'errorHandler' => [
  'errorAction' => 'site/error',
],
```

PR #17703, opened 2019-12-02
Fixed Error Handler to clear view tags, scripts, etc. when rendering error action view

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17694 

I've added TestController in `yiiunit\data\controllers` namespace for testing this and at the same time to use it instead of TestController that was in ErrorActionTest


samdark:
Merged. Thank you!

Issue #17698, opened 2019-11-30
Fix PHP 7.4


samdark:
See https://travis-ci.org/yiisoft/yii2/jobs/619055884

samdark:
So we have two problem classes:

1. There's no PHPUnit that is compatible with both php5.4 (which is set as platform target).
2. Real errors.

samdark:
There are two ways fixing it.

1. Patch old PHPUnit like I've tried in #17706. Messy and complicated.
2. Remove `"platform": {"php": "5.4"}` from config, set two possible PHP versions in dependencies, have separate test sets for PHP 7.4 and PHP < 7.4. Leading to maintenance hell.

samdark:
- https://phpunit.de/supported-versions.html
- https://www.php.net/supported-versions.php

samdark:
Another option is to release Yii 2.1.0 for PHP >= 7.4 and leaving Yii 2.0.x for PHP < 7.4.

rob006:
Bumping required PHP version is not a BC break. If bumping requirements is the only reason for next major release, then it is definitely not worth it - we could require PHP 7 in Yii 2.0.31 and leave 2.0.30 as LTS with security fixes only.

terabytesoftw:
I would vote to have a minimum version of php 7.4, such as Yii 2.1 and take advantage of and add the functionalities that bc would do in lower versions of php, and an LTS version in 2.0 that would not accept corrections but only security maintenance. Yii2 has had many corrections and should continue advancing and debugging the code with the new php functionalities as much as possible, those are my two cents.

charescape:
> Another option is to release Yii 2.1.0 for PHP >= 7.4 and leaving Yii 2.0.x for PHP < 7.4.

I quite agree with this option. Yii 2.0.x has done enough compatibility work, it is time to release 2.1.x for PHP 7.4 and later.

samdark:
There is another option: drop everything except 7.2, 7.3 and 7.4 from tests but don't drop support for 5.4 overall.

talentant:
> There is another option: drop everything except 7.2, 7.3 and 7.4 from tests but don't drop support for 5.4 overall.

I don't think thsi is a good idea. If so, how do we ensure that test cases always pass?

samdark:
You mean dropped ones? We can't.

samdark:
😕  So Yii 2 being compliant with 7.4 is tricky.

[...]

PR #17706, opened 2019-12-03
Add PHPUnit patch to avoid __toString() reflection warning on PHP 7.4

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ?
| Fixed issues  | #17698


samdark:
Seems to be leading nowhere :(

Issue #17708, opened 2019-12-04
FormatterNumberTest::testIntlAsScientific fails on GitLab


schmunk42:
### What steps will reproduce the problem?

See tests here: https://gitlab.com/yiisoft/yii2/-/jobs/368388365

### What is the expected result?

No failure or error.

### What do you get instead?

```
There was 1 failure:
573 1) yiiunit\framework\i18n\FormatterNumberTest::testIntlAsScientific
574 Failed asserting that two strings are identical.
575 --- Expected
576 +++ Actual
577 @@ @@
578 -8.76543210987654E16
579 +8.765432109876544E16
580 /project/tests/framework/i18n/FormatterNumberTest.php:531
581 /project/vendor/phpunit/phpunit/phpunit:52
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | master
| PHP version      | 7.1.33
| Operating system | Debian (Docker)

Looks like there's one more digit now?!
Test fails since https://gitlab.com/yiisoft/yii2/commit/c1e9739ed232e207ff6920e90b7351b2b19f69d0

Pipelines: https://gitlab.com/yiisoft/yii2/pipelines

schmunk42:
Just noticed: Test worked on **PHP 7.1.26**

samdark:
So is there a change between that and PHP 7.1.33 that causes one extra digit to appear?

bizley:
Has precision setting been changed in php.ini?

schmunk42:
I triggered tests for all versions in question at https://gitlab.com/yiisoft/yii2-docker/pipelines

I have not changed anything in our files.

schmunk42:
Tests fail for 7.1.30

PHP Changelog says:

Version 7.1.30
30 May 2019
EXIF:
Fixed bug #77988 (heap-buffer-overflow on php_jpg_get16) (CVE-2019-11040).
GD:
Fixed bug #77973 (Uninitialized read in gdImageCreateFromXbm) (CVE-2019-11038).
Iconv:
Fixed bug #78069 (Out-of-bounds read in iconv.c:_php_iconv_mime_decode() due to integer overflow) (CVE-2019-11039).
SQLite:
Fixed bug #77967 (Bypassing open_basedir restrictions via file uris).

schmunk42:
Debian version has been changed:

```
> docker run --rm php:7.1.30-apache cat /etc/os-release 
[...]

PR #17714, opened 2019-12-06
Bug #17708: Explicitly define precision to be 15 for FormatterNumberTest::testIntlAsScientific

diego-mathis:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -7,7 +7,7 @@ Yii Framework 2 Change Log
 - Bug #17865: Fix invalid db component in `m180523_151638_rbac_updates_indexes_without_prefix` (rvkulikov)
 - Bug #17694: Fixed Error Handler to clear registered view tags, scripts, and files when rendering error view through action view (bizley)
 - Bug #17701: Throw `BadRequetHttpException` when request params can’t be bound to `int` and `float` controller action arguments (brandonkelly)
-
+- Bug #17708: Explicitly define precision to be 15 for FormatterNumberTest::testIntlAsScientific (My6UoT9)
 
 2.0.30 November 19, 2019
 ------------------------
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -9,6 +9,7 @@ Yii Framework 2 Change Log
 - Bug #17701: Throw `BadRequetHttpException` when request params can’t be bound to `int` and `float` controller action arguments (brandonkelly)
 - Bug #17708: Explicitly define precision to be 15 for FormatterNumberTest::testIntlAsScientific (My6UoT9)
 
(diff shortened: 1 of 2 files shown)


schmunk42:
```
1) yiiunit\framework\i18n\FormatterNumberTest::testIntlAsScientific
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-8.76543210987654E16
+8.765432109876540E16
```

Fails with the same error, maybe we should use `14` as precision value?

My6UoT9:
That is really strange. Seems its more difficult to fix.
With 14, I get this (I actually thought its 14, but only 15 did pass the test for me):
```
There was 1 failure:

1) yiiunit\framework\i18n\FormatterNumberTest::testIntlAsScientific
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-8.76543210987654E16
+8.7654321098765E16
```
With 15 it works.

schmunk42:
I think we should do the test with, i.e. precision `10` and the number formatter has to format to this length/precision.

schmunk42:
@samdark Maybe there should be an exception, when the precision is higher than the system can handle?

My6UoT9:
There are more problems with this.
Disabling intl extension gives:
```
1) yiiunit\framework\i18n\FormatterNumberTest::testIntlAsScientific
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-1.23E2
+1.230000E+2

E:\fire\yii2-new\tests\framework\i18n\FormatterNumberTest.php:519
E:\fire\yii2-new\vendor\phpunit\phpunit\phpunit:52

2) yiiunit\framework\i18n\FormatterNumberTest::testAsScientific
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-1.23E+2
+1.2E2

E:\fire\yii2-new\tests\framework\i18n\FormatterNumberTest.php:538
```

Then I tried (with intl active):
```
 $this->assertSame('8.765432109876544E16', $this->formatter->asScientific('87654321098765436'));
```
Notice the extra "4" 8.76543210987654**4**E16

That did pass on my windows. And it would pass travis I guess.
But this doesn't feel satisfying. I believe this should be investigated some more.

terabytesoftw:
Closing as obsolete.

The underlying issue #17708 was resolved by #17837, which skipped  `FormatterNumberTest::testIntlAsScientific()` because the output depends on the ICU version. This PR only changes an assertion inside that skipped test, so it no longer has practical effect on the current codebase.

PR #17720, opened 2019-12-08
Fix yii3 di

hyuzea:
Work with:
- Application core components
- Yii clasess that depend on another one, eg:
	- View -> Theme
	- AccessControll -> AccessRule
	- GridView -> DataColumn
	- DataProvider -> Sort, Pagination
	- etc.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️ - with ```vendor/bin/phpunit --exclude-group=db,mail,validators,i18n```
| Fixed issues  | -
| Related          | #17607, #17670

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -345,23 +345,23 @@ class BaseYii
             return static::$container->get($type, $params);
         }
 
-        if (is_array($type) && isset($type['class'])) {
-            $class = $type['class'];
-            unset($type['class']);
-            return static::$container->get($class, $params, $type);
-        }
-
-        if (is_array($type) && isset($type['__class'])) {
-            $class = $type['__class'];
-            unset($type['__class']);
-            return static::$container->get($class, $params, $type);
-        }
-
         if (is_callable($type, true)) {
             return static::$container->invoke($type, $params);
         }
 
         if (is_array($type)) {
+            if (isset($type['__class'])) {
+                $class = $type['__class'];
+                unset($type['__class'], $type['class']);
+                return static::$container->get($class, $params, $type);
+            }
+
+            if (isset($type['class'])) {
+                $class = $type['class'];
+                unset($type['class']);
+                return static::$container->get($class, $params, $type);
+            }
+
             throw new InvalidConfigException('Object configuration must be an array containing a "class" or "__class" element.');
         }
 
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -349,23 +349,23 @@ class BaseYii
             return static::$container->invoke($type, $params);
         }
 
-        if (is_array($type)) {
-            if (isset($type['__class'])) {
-                $class = $type['__class'];
-                unset($type['__class'], $type['class']);
-                return static::$container->get($class, $params, $type);
-            }
-
-            if (isset($type['class'])) {
-                $class = $type['class'];
-                unset($type['class']);
-                return static::$container->get($class, $params, $type);
-            }
-
-            throw new InvalidConfigException('Object configuration must be an array containing a "class" or "__class" element.');
+        if (!is_array($type)) {
+            throw new InvalidConfigException('Unsupported configuration type: ' . gettype($type));
         }
 
-        throw new InvalidConfigException('Unsupported configuration type: ' . gettype($type));
+        if (isset($type['__class'])) {
+            $class = $type['__class'];
+            unset($type['__class'], $type['class']);
+            return static::$container->get($class, $params, $type);
+        }
+
+        if (isset($type['class'])) {
+            $class = $type['class'];
+            unset($type['class']);
+            return static::$container->get($class, $params, $type);
+        }
+
+        throw new InvalidConfigException('Object configuration must be an array containing a "class" or "__class" element.');
     }
 
     private static $_logger;
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,10 +4,10 @@ Yii Framework 2 Change Log
 2.0.31 under development
 ------------------------
 
-- Enh #17720: DI 3 support for application core components and default object configurations (sup-ham)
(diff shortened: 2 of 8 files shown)


tests/framework/BaseYiiTest.php
```diff
@@ -126,6 +128,17 @@ public function testCreateObjectInvalidConfigException()
         Yii::createObject(null);
     }
 
+    public function testDi3CompatibilityCreateObject()
+    {
+        $objek = Yii::createObject([
```
xepozz:
```suggestion
        $object = Yii::createObject([
```

tests/framework/BaseYiiTest.php
```diff
@@ -126,6 +128,17 @@ public function testCreateObjectInvalidConfigException()
...
+            '__class' => View::className(),
+            'theme' => ['__class' => Theme::className()],
+        ]);
+
+        $this->assertInstanceOf(View::className(), $objek);
```
xepozz:
```suggestion
        $this->assertInstanceOf(View::className(), $object);
```

tests/framework/BaseYiiTest.php
```diff
@@ -126,6 +128,17 @@ public function testCreateObjectInvalidConfigException()
...
+        $this->assertInstanceOf(Theme::className(), $objek->theme);
```
xepozz:
```suggestion
        $this->assertInstanceOf(Theme::className(), $object->theme);
```

framework/BaseYii.php
```diff
@@ -345,23 +345,23 @@ public static function createObject($type, array $params = [])
...
-        }
-
         if (is_callable($type, true)) {
             return static::$container->invoke($type, $params);
         }
 
         if (is_array($type)) {
+            if (isset($type['__class'])) {
+                $class = $type['__class'];
+                unset($type['__class'], $type['class']);
```
xepozz: Why need to unset `class` key?

framework/BaseYii.php
```diff
@@ -345,23 +345,23 @@ public static function createObject($type, array $params = [])
...
+                return static::$container->get($class, $params, $type);
+            }
+
+            if (isset($type['class'])) {
+                $class = $type['class'];
+                unset($type['class']);
+                return static::$container->get($class, $params, $type);
+            }
+
```
xepozz: Please use `if (!is_array($type)) {` then throw an exception

framework/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Yii Framework 2 Change Log
 2.0.31 under development
 ------------------------
 
+- Enh #17607: DI 3 support for application core components and default object configurations (sup-ham)
```
xepozz:
```suggestion
- Enh #17720: DI 3 support for application core components and default object configurations (sup-ham)
```

xepozz (changes requested):

framework/BaseYii.php
sup-ham: Lot of Yii component has another component with default config ['class' => DefaultClass]. This will lead to error `setting unknown property class`.

framework/BaseYii.php
sup-ham: `$type` here exactly is an array

xepozz: I mean it needs to flip over the condition from `if (is_array($type)) {` to `if (NOT is_array($type)) {` then throw an exception, that `$type` is not an array and continue to check a properties

viktorprogger: We don't need to change the condition here because this exception is thrown only on these three conditions matching:

- the config **is** an array
- there is no `class` key in it
- there is no `__class` key in it

framework/base/Application.php
```diff
@@ -258,7 +258,7 @@ public function preInit(&$config)
         foreach ($this->coreComponents() as $id => $component) {
             if (!isset($config['components'][$id])) {
                 $config['components'][$id] = $component;
-            } elseif (is_array($config['components'][$id]) && !isset($config['components'][$id]['class']) && !isset($config['components'][$id]['__class'])) {
+            } elseif (is_array($config['components'][$id]) && !isset($config['components'][$id]['class'])) {
```
viktorprogger: And what's the deal with the `__class` key? You made [a change with this key](https://github.com/yiisoft/yii2/pull/17670/files#diff-7aab80b18ec8de0f4e9dcf3168056e40R261) not so far.

tests/framework/BaseYiiTest.php
```diff
@@ -190,3 +203,11 @@ public function testLog()
...
     }
 }
+
+class ThemeMock extends Theme
+{
+}
+
+class ViewMock extends View
+{
+}
```
viktorprogger: I think we should use less complex classes for an example and define configurable fields ourselves. Now this test is dependent not only on DI logic, but on View logic too.

Try something like:

```
class ObjectMock extends BaseObject
{
    public $testField;
}

class SecondMock extends BaseObject
{
}
```

tests/framework/base/ApplicationTest.php
```diff
@@ -78,3 +92,7 @@ public function bootstrap($app)
     {
     }
 }
+
+class ViewMock extends View
+{
+}
```
viktorprogger: The same as in the previous comment. Please, use less complex mocks.

viktorprogger (changes requested):
Thnx, it is a good work. There is just a little change pending in tests and a question.

framework/base/Application.php
sup-ham: [Refactored to ServiceLocator]( https://github.com/yiisoft/yii2/pull/17720/files?short_path=6ed2533#diff-86a8d4b20894b0affd0b4dc8828373b6R202-R206)

viktorprogger (approved):

terabytesoftw:
Approved.

samdark (approved):

samdark:
Merged. Thank you!

PR #17723, opened 2019-12-09
Fixed for PHP 7.4 without tests

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️ (except 7.4)
| Fixed issues  | -

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -819,7 +819,7 @@ class Model extends Component implements StaticInstanceInterface, IteratorAggreg
         }
         $attributes = array_keys(array_flip($scenarios[$scenario]));
         foreach ($attributes as $i => $attribute) {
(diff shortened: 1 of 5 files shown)


framework/base/Model.php
```diff
@@ -819,7 +819,7 @@ public function activeAttributes()
         }
         $attributes = array_keys(array_flip($scenarios[$scenario]));
         foreach ($attributes as $i => $attribute) {
-            if ($attribute[0] === '!') {
+            if (is_string($attribute) && strpos($attribute, '!') === 0) {
```
rob006:
```suggestion
            if (strncmp($attribute, '!', 1) === 0) {
```

?

PR #17725, opened 2019-12-10
Patch PHPUnit and PHPUnit mock objects to update deprecated code

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ?
| Fixed issues  | #17698

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -84,7 +84,8 @@
         "phpunit/phpunit": "4.8.34",
         "cebe/indent": "~1.0.2",
         "friendsofphp/php-cs-fixer": "~2.2.3",
-        "johnkary/phpunit-speedtrap": "^1.0"
+        "johnkary/phpunit-speedtrap": "^1.0",
+        "cweagans/composer-patches": "^1.6"
     },
     "repositories": [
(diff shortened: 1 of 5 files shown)


samdark:
So it worked but PHP 5.4 fails about pbkdf2 which is weird because it works in master: https://travis-ci.com/yiisoft/yii2/jobs/265592886 and no code related to that was changed.

samdark:
Looks like it's caused by dependencies update: https://travis-ci.com/yiisoft/yii2/builds/140580150

samdark:
Found the culprit. It's symfony/polyfill-php55 that replaces `hash_pbkdf2` in case it does not exists. It is one of the transitive dependencies of dev packages. Implementation was correct in version 1.6.0. Current version that gives incorrect value is 1.13.1.

samdark:
This is the commit: https://github.com/symfony/polyfill-php55/commit/42a4c00a347625ac8853c3358c47eeadc7fd4e96

Issue #17037, opened 2019-01-13
MultipartFormDataParser using UploadedFile does not work


Nuffic:
### What steps will reproduce the problem?

1. use `yii\web\MultipartFormDataParser` as a request parser
2. upload file using PATCH method
3. use UploadedFile to instanciate the file

### What is the expected result?
When using UploadedFile::saveAs() file is saved into specified directory.

### What do you get instead?
Since it uses php-s `move_uploaded_file()` or `is_uploaded_file()` which check if php itself dealt with uploaded file (which it didn't), file is not saved.

### Additional info

`yii\web\MultipartFormDataParser` deals with $_FILES itself not php.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1

Relates to #6996 but is not the same.

mihaivarga:
We attempted an upgrade from 2.0.14. For us this seems to be happening with latest #2.0.18 version using POST also. Any advice? File do not get saved, type not identified and files shows error 3 and validation trigger the file type missing. All works fine when we downgrade to 2.0.14.1.

alexkart:
@Nuffic this mentioned in the documentation https://www.yiiframework.com/doc/api/2.0/yii-web-multipartformdataparser

```
Note: although this parser fully emulates regular structure of the $_FILES, related temporary files, which are available via tmp_name key, will not be recognized by PHP as uploaded ones. Thus functions like is_uploaded_file() and move_uploaded_file() will fail on them. This also means yii\web\UploadedFile::saveAs() will fail as well.
```

PR #17729, opened 2019-12-12
Fix file uploads saving in case generated by `MultipartFormDataParser`

hyuzea:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17037


sup-ham:
Also the docs need to be adjusted.

samdark:
@sup-ham which docs?

samdark:
There seems to be trouble with the changes under Windows.

My6UoT9:
Would this work if my filename is "@webroot.jpg" ? Looks like trouble?

framework/web/UploadedFile.php
```diff
@@ -149,25 +162,48 @@ public static function reset()
...
+        return !$deleteTempFile || $this->deleteTempFile();
+    }
+
+    /**
+     * @return bool|int
+     */
+    private function copyTempFile($file)
+    {
+        if (!is_resource($this->_tempResource)) {
+            return is_file($this->tempName) && copy($this->tempName, $file);
```
samdark: That looks suspicious to me. Previously there was `is_uploaded_file` check that ensured that file moved is what came via POST. Now, resource is empty only when the method is POST but the check is missing.

samdark (changes requested):

samdark (approved):

samdark:
Thank you!

Issue #17710, opened 2019-12-05
Memcache expire does't work in special cases


lyt8384:
### What steps will reproduce the problem?
https://github.com/yiisoft/yii2/issues/17063#issue-402056858 , this problems arise for many people.
```php
    $resultSet = Yii::$app->cache->set('test', 'ok', 60);
    var_dump($resultSet);  //ture
    $resultGet = Yii::$app->cache->get('test');
    var_dump($resultGet);  //false
```

Because memcached internally uses a clock based on the monotonic clock, php `time()` uses sytems timestamp. 
If memcached has been running for a long time since startup,In `framework/caching/MemCache.php#L298` `$expire` probably behind memcached internally timestamp, so the problem happened.
see https://github.com/php-memcached-dev/php-memcached/issues/368#issuecomment-359137077
 https://github.com/yiisoft/yii2/blob/f0de56b81e4fb6731304fd070f8d2f72a040e52e/framework/caching/MemCache.php#L298

### What is the expected result? 
I suggest `$expire` use the number of seconds, not Unix timestamp, although this will cause `$expire` may not exceed 30 days, but can be solved this problems

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.24
| PHP version      |  7.3.12
| Operating system | centos 7

samdark:
That changes behavior very significantly so I'm not sure it's a good idea overall.

samdark:
@yiisoft/reviewers what do you think?

rob006:
It shouldn't hurt if we do this only for `$expire` smaller than 30 days.

samdark:
I like @rob006 solution. We can reliably detect the intent in this case. If value is less than 2592001 then it's duration. Otherwise it's timestamp.

PR #17731, opened 2019-12-12
Fix MemCache duration normalization to avoid memcached/system timestamp mismatch

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17710


rob006:
Previously if user pass negative `$duration` it was normalized to 0. Now it is used as is. Does it matter for memcache?

samdark:
It should not but I've added it just in case.

Issue #17635, opened 2019-10-23
I give "unserialize(): Error at offset 0 of 60 bytes" for create permission with rule


bagtim:
```php
 $auth = Yii::$app->authManager;

// add the rule
        $rule = new \app\rbac\AuthorRule;
        $auth->add($rule);

// add "updateOwnPost" permission and bind a rule to it
        $updateOwnPost = $auth->createPermission('updateOwnPost');
        $updateOwnPost->description = 'Update own post';
        $updateOwnPost->ruleName = $rule->name;
        $auth->add($updateOwnPost);
```
Without setted ruleName, permission is good created. But if set ruleName, i have error "unserialize(): Error at offset 0 of 60 bytes"

| Q                | A
| ---------------- | ---
| Microsoft SQL Server| 2017 (RTM) - 14.0.1000.169 (X64) | 
| Yii version      | 2.0.29 |
| PHP version   | 5.6.31  | 
| Operating system | win10 64 |

samdark:
1. Was PHP manager used?
2. What was the name?

samdark:
Do you store RBAC rules in MSSQL?

samdark:
Was table created from migrations or https://github.com/yiisoft/yii2/blob/b39bb71587cb4a4964238bdc6e846535588a3cfc/framework/rbac/migrations/schema-mssql.sql?

samdark:
What was used as the rule name?

bagtim:
> What was used as the rule name?

From docs: 
```
class AuthorRule extends Rule
{
    public $name = 'isAuthor';
...
```
and then from topic post:
```
$rule = new \app\rbac\AuthorRule;
$auth->add($rule);
...
$updateOwnPost->ruleName = $rule->name;
$auth->add($updateOwnPost);
```

samdark:
OK. That's enough info to try reproducing it.

toatall:
I had an analogical problem.
I did so:
1) Change to CONVERT(VARBINARY(**MAX**), ...) in [QueryBuilder::normalizeTableRowData](https://github.com/yiisoft/yii2/blob/master/framework/db/mssql/QueryBuilder.php#L429) because default length is 30 characters ([Link MS SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/functions/cast-and-convert-transact-sql?view=sql-server-ver15))
[...]

PR #17735, opened 2019-12-13
17653 unserialize rbac rule

OlegTDev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌
| Fixed issues  | #17635


samdark:
Seems to cause errors: https://travis-ci.com/yiisoft/yii2/jobs/266838304#L455

toatall:
I found another way. I added an attribute in the config file `console/config/main-local.php`:

`'db' => ['attributes' => [ \PDO::SQLSRV_ATTR_ENCODING => \PDO::SQLSRV_ENCODING_SYSTEM ]]`

samdark:
Hmm... which solution do you think is the best one?

toatall:
I would leave the commit changing varbinary to varbinary(max) -> 6c9f962c1b537c8de886f6463d7a606b37a7e9e0
And for the correct save a binary data use the db attribute `\PDO::SQLSRV_ATTR_ENCODING`

samdark:
So we need both?

toatall:
Yes

samdark:
I think then https://github.com/yiisoft/yii2/pull/17735#issuecomment-565944429 should be added to docs around https://github.com/yiisoft/yii2/blob/master/docs/guide/db-dao.md#creating-db-connections- and to MSSQL Connection phpdoc. Do you have time to handle it?

toatall:
@samdark Ok, but a little later

samdark (approved):

samdark:
Thanks!

PR #17736, opened 2019-12-15
Update Controller.php

aaronlei:
Correct the annotation.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/framework/base/Controller.php
+++ b/framework/base/Controller.php
@@ -214,7 +214,7 @@ class Controller extends Component implements ViewContextInterface
      * where `xyz` is the action ID. If found, an [[InlineAction]] representing that
      * method will be created and returned.
      * @param string $id the action ID.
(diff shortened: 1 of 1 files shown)


framework/base/Controller.php
```diff
@@ -214,7 +214,7 @@ public function bindActionParams($action, $params)
      * where `xyz` is the action ID. If found, an [[InlineAction]] representing that
      * method will be created and returned.
      * @param string $id the action ID.
-     * @return Action|null the newly created action instance. Null if the ID doesn't resolve into any action.
+     * @return object|InlineAction|null the newly created action instance. Null if the ID doesn't resolve into any action.
```
kamarton: Why? if necessary then `Action|InlineAction|null`.

samdark:
Original phpdoc is correct.

PR #17741, opened 2019-12-17
fix #17740: `yii\helpers\BaseInflector::slug()` doesn't replace multip…

### LINKED ISSUE #17740: yii\helpers\BaseInflector::slug() doesn't shrink multiple replacement occurences to single
Inflector::slug(' _ _ remove excess _ _ replacements_', '_');
remove_excess_replacements
### What do you get instead?
remove_excess_____replacements

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16-2.0.30 since #15204
| PHP version      | 7.0.33
| Operating system | Debian 7

batyrmastyr:
…le replacement string occurrences to single one

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17740

### DIFF
--- a/framework/helpers/BaseInflector.php
+++ b/framework/helpers/BaseInflector.php
@@ -485,6 +485,7 @@ class BaseInflector
         }, $parts);
 
         $string = trim(implode($replacement, $replaced), $replacement);
+        $string = preg_replace('#'.preg_quote($replacement).'+#', $replacement, $string);
 
         return $lowercase ? strtolower($string) : $string;
     }
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -15,7 +15,7 @@ Yii Framework 2 Change Log
 - Bug #17723: Fix incorrect type-casting of reflection type to string (samdark)
 - Bug #17725: Ensure we do not use external polyfills for pbkdf2() as these may be implemented incorrectly (samdark)
 - Bug #17710: Fix MemCache duration normalization to avoid memcached/system timestamp mismatch (samdark)
-
+- Bug #17740: `yii\helpers\BaseInflector::slug()` is not replace multiple replacement string occurrences to single one (batyrmastyr)
 
 2.0.30 November 19, 2019
 ------------------------
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -15,7 +15,7 @@ Yii Framework 2 Change Log
 - Bug #17723: Fix incorrect type-casting of reflection type to string (samdark)
 - Bug #17725: Ensure we do not use external polyfills for pbkdf2() as these may be implemented incorrectly (samdark)
 - Bug #17710: Fix MemCache duration normalization to avoid memcached/system timestamp mismatch (samdark)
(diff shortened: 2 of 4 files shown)


framework/helpers/BaseInflector.php
```diff
@@ -485,6 +485,7 @@ public static function slug($string, $replacement = '-', $lowercase = true)
         }, $parts);
 
         $string = trim(implode($replacement, $replaced), $replacement);
+        $string = preg_replace('#'.preg_quote($replacement).'+#', $replacement, $string);
```
kamarton: > ... s a backwards compatibility break for Yii 2.

add an extra parameter to the method (only for yii2). @samdark ?

```php
public static function slug($string, $replacement = '-', $lowercase = true, $clean = false)
```

I would not implement the extra parameter in yii3.

samdark: I'd use extra preg_replace or external wrapper or your own implementation of Inflector instead. That's not something that could be considered Yii 2 bug.

batyrmastyr: It was introduced in Yii 2.0.16 and it doesn't look like intended change, so why it's not a bug? In terms of age - 2.0.0 - 2.0.16 is 4 years, 2.0.16 - now is 1.5 year.

samdark: Hmm... wait, it was doing that before 2.0.16?

samdark: Alright. That's bug.

samdark:
@batyrmastyr would you please add a note about output change for UPGRADE document? Thanks.

batyrmastyr:
@samdark should there be `clean` option with true or false or no option at all?
```Upgrade from Yii 2.0.30:```

- In case `$clean` defaults to `true`.

`yii\helpers\BaseInflector::slug()` now ensures that there is no repeating $replacement occurrences in slug, ex. 'slug-without-repeats' vs 'slug-with---repeats'. In case you relied on behavior, replace `Inflector` with your own implementation.
```
public static function slug($string, $replacement = '-', $lowercase = true, $clean = true)
{
   return parent::slug($string, $replacement, $lowercase, $clean)
}
Yii::$classMap['yii\helpers\Inflector'] = '@your/Inflector.php';
```
- In case `$clean` defaults to `false`. 

If you upgraded from Yii 2.0.16 and expect `yii\helpers\BaseInflector::slug()` to remove repeating $replacement occurrences, ex. 'slug-without-repeats' vs 'slug-with--repeats', replace `Inflector` with your own implementation.
```
public static function slug($string, $replacement = '-', $lowercase = true, $clean = true)
{
   return parent::slug($string, $replacement, $lowercase, $clean)
}
Yii::$classMap['yii\helpers\Inflector'] = '@your/Inflector.php';
```
- In case there is not option at all, there is nothing to write except apologies.

samdark:
No option at all, I think since it's a regression.

samdark:
> In case there is not option at all, there is nothing to write except apologies.

Kind of :) The info should be there about the fact of the change.

samdark (approved):

samdark:
Merged. Thank you!

PR #17743, opened 2019-12-17
Allowed to set custom url parameter name in yii\grid\ActionColumn

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

In my case I had to custom url parameter name in ActionColumn.
I decided to share it with community.


samdark:
While it makes sense, I'd prefer not to add it to Yii 2.0:

1. That was not requested before. Means it is quite specific use-case.
2. Feature freeze.
3. Passing name as "id" is possible.

PR #17761, opened 2019-12-24
Dispatcher::dispatch() target crash php 7.x fix #17749

### LINKED ISSUE #17749: dispatch crash, if logtarget::collect throw \Throwable

1. `$target->collect()` throw `\Throwable`
2. Further targets can no longer run. the developer toolbar cannot save the log file.

https://github.com/yiisoft/yii2/blob/596bcff0b41d6c5dc074ca5c248741f1a87bbe64/framework/log/Dispatcher.php#L188-L199

The other logs work correctly.

### What do you get instead?

After the first error, the other logs are no longer running.

### Additional info

The code below fixes the bug, but require PHP 7.0+ 
```php
                 try { 
                     $target->collect($messages, $final); 
                 } catch (\Throwable $e) { 
```

Separate try-catch block for PHP 7.0+ and PHP <7.0 can solve the problem, but need to consider whether there is a better solution.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.x
| PHP version      | 7.0+
| Operating system | -

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17749


samdark:
Seems to fail tests.

kamarton:
@samdark reviewable

framework/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ Yii Framework 2 Change Log
 ------------------------
 
 - Bug #17744: Fix a bug with setting incorrect `defaultValue` to AR column with `CURRENT_TIMESTAMP(x)` as default expression (MySQL >= 5.6.4) (bizley)
-
+- Bug #17749: Dispatcher fix if target crashed in PHP 7.0+ (kamarton)
```
samdark: Exact version could be determined by pull-requesting tests only and checking which versions fail.

framework/log/Dispatcher.php
```diff
@@ -184,24 +184,41 @@ public function dispatch($messages, $final)
...
         }
     }
+
+    /**
+     * Generate target error message
+     *
+     * @param Target                $target
+     * @param \Throwable|\Exception $throwable
+     * @param string                $method
+     * @return array
```
samdark: Need `@since` tag. Also, alignment should not be done for phpdoc.

framework/CHANGELOG.md
samdark: Would you please do it?

framework/log/Dispatcher.php
kamarton: modified.

framework/CHANGELOG.md
kamarton: sorry, I don't understand what you want (by my poor english)

samdark: 1. Create a new branch from `17749-dispatch-crash-test`.
2. Remove code from it except unit tests.
3. Make another pull request.
4. Wait till travis is finished. It will show you which PHP versions fail and which did not.

kamarton: #17768

samdark: OK. That seems to be 7.0+. Thanks for validating it.

samdark (approved):

samdark:
Thanks!

PR #17763, opened 2019-12-24
Removed BaseActiveRecord::offsetExists to rely on yii/base/Model's im…

### LINKED ISSUE #17717: Unable to check if ActiveRecord own properties isset or empty using array access notation
Created a ActiveRecord model class, for example:

```php
class FinancialTransaction extends \yii\db\ActiveRecord
{
    /**
     * Custom property to be set manually
     */
    public $annotation;

    /**
     * @inheritdoc
     */
    {
        return '{{%account}}';
    }

    // ... 
}
```

And run code like this:

```php
$model = new FinancialTransaction();
$model->annotation = 'Some annotation';

var_dump($model['annotation']);
var_dump(isset($model['annotation']));
var_dump(empty($model['annotation']));

```

```
string(15) "Some annotation"
bool(true)
bool(false)
```
### What do you get instead?

```
string(15) "Some annotation"
bool(false)
bool(true)
```

### Suggestion:

Change implementation of BaseActiveRecor::offsetExists from:
```php
public function offsetExists($offset)
{
    return $this->__isset($offset);
}
```
to:
```php
public function offsetExists($offset)
{
    return isset($this->$offset);
}
```
So it will cover own ActiveRecord properties and fallback to __isset when needed.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.28, 2.0.30
| PHP version      | 7.3, 7.4
| Operating system | Mac OSX 10.15.1

jwlewisiii:
Removed BaseActiveRecord::offsetExists to rely on yii/base/Model's implementation of it which fixes yiisoft#17717

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17717


schmunk42 (changes requested):
You can not remove a `public` method in a patch (nor a minor) release.

jwlewisiii:
> You can not remove a `public` method in a patch (nor a minor) release.

My bad. I didn't realize it. That should be added to the Yii Versioning wiki doc. We can keep the `public` method and change the implementation to call the parents offsetExist.

rob006:
> We can keep the `public` method and change the implementation to call the parents offsetExist.

This is pointless - it will work the same as if you remove this method from child class.

@schmunk42 This method is implemented at `Model` level - can you explain how removing this method from `ActiveRecord` could break BC?

schmunk42:
Sorry, then it is OK

samdark:
There are cases I'd specifically check (if behavior is changed):

```php
class MyActiveRecord extends ActiveRecord
{
    public function setX()
    {
        // ...
    }
}
```

1. `isset($myActiveRecord['non_existing'])`.
2. `isset($myActiveRecord['x'])`
3. `isset($myActiveRecord['incorrectly_defined_relation'])`

samdark:
@jwlewisiii any luck with checking additional cases?

jwlewisiii:
> @jwlewisiii any luck with checking additional cases?

Unfortunately I haven't had a lot of time to focus on it but I'm still working on adding the additional test.

samdark:
@jwlewisiii any update?

PR #17769, opened 2019-12-27
Fix ActiveRelationTrait.php for compatibility 7.4

indrig:
Fix for PHP 7.4

Trying to access array offset on value of type null
When $model is null on Php 7.4:
if (($value = $model[$attribute]) !== null)

/var/www/builds/cms/3.6.0/vendor/yiisoft/yii2/db/ActiveRelationTrait.php line 526

#0 /var/www/builds/cms/3.6.0/vendor/yiisoft/yii2/db/ActiveRelationTrait.php(526): yii\base\ErrorHandler->handleError()
#1 /var/www/builds/cms/3.6.0/vendor/yiisoft/yii2/db/ActiveRelationTrait.php(246): yii\db\ActiveQuery->filterByModels()
#2 /var/www/builds/cms/3.6.0/vendor/yiisoft/yii2/db/ActiveQueryTrait.php(151): yii\db\ActiveQuery->populateRelation()
#3 /var/www/builds/cms/3.6.0/vendor/yiisoft/yii2/db/ActiveQuery.php(224): yii\db\ActiveQuery->findWith()

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  |❌
| Breaks BC?    | ❌
| Tests pass?   | ❌

### DIFF
--- a/framework/db/ActiveRelationTrait.php
+++ b/framework/db/ActiveRelationTrait.php
@@ -523,7 +523,7 @@ trait ActiveRelationTrait
             // single key
             $attribute = reset($this->link);
             foreach ($models as $model) {
-                if (($value = $model[$attribute]) !== null) {
+                if ($model !== null && ($value = $model[$attribute]) !== null) {
                     if (is_array($value)) {
                         $values = array_merge($values, $value);
                     } elseif ($value instanceof ArrayExpression && $value->getDimension() === 1) {


framework/db/ActiveRelationTrait.php
```diff
@@ -523,7 +523,7 @@ private function filterByModels($models)
             // single key
             $attribute = reset($this->link);
             foreach ($models as $model) {
-                if (($value = $model[$attribute]) !== null) {
```
samdark: Looking at the code I don't see anything that would fail on PHP 7 but won't fail on PHP 5. Are you sure it's PHP 7 specific?

Ximich: @samdark, yes, I can confirm it's php 7.4 incompatible change: https://www.php.net/manual/en/migration74.incompatible.php
>>> Array-style access of non-arrays
>>> Trying to use values of type null, bool, int, float or resource as an array (such as $null["key"]) will now generate a notice.



+ #17914

Please, merge this PR.

Ximich: Do I need create another PR or you can merge this one?

rob006: @Ximich How can you get `null` in `$models`?

Ximich: @rob006, in relation with via, where relation attribute has null value.

```
    public function getSomething()
    {
        return $this->hasOne(Something::class, ['id' => 'something_id'])->via('viaAnotherModel');
    }
```

This code works fine in php 7.3.

Ximich: populateRelation() method in ActiveRelationTrait return $viaModels with null value in this case here: https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveRelationTrait.php#L245

rob006: Shouldn't we remove these nulls from `$models` before calling `filterByModels()`? IMO it looks weird that this method needs to handle nulls, while it obviously expecting models.

Ximich: I don't know where's right place to filter null models. @samdark, what do you think: should we filter null `$models` in `populateRelation()` method or check it in `filterByModels()`?

samdark: Filtering sounds like a right thing to do.

PR #17773, opened 2019-12-28
Update input-tabular-input.md

murtraja:
added the loop variable for array population

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | just a small typo, which had me scratching my head for 10 minutes


rob006:
It shouldn't be necessary - `$settings[] = new Setting();` will use next available index for element, which is the same as `$i`.

murtraja:
Wow, didn't know that was possible. Can you give the name of this feature or a link to the corresponding php docs?

rob006:
https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax.modifying

Issue #17623, opened 2019-10-21
Flash is always lost when using DbSession (postgresql)


yakninja:
### What steps will reproduce the problem?

1. Create a new application, basic template
2. Create a database and a session table:
```sql
CREATE TABLE session
(
    id CHAR(64) NOT NULL PRIMARY KEY,
    expire INTEGER,
    data BYTEA
)
```
3. Edit config/db.php:
```php
<?php

return [
    'class' => 'yii\db\Connection',
    'dsn' => 'pgsql:host=127.0.0.1;dbname=dbname',
    'username' => 'dbuser',
    'password' => 'dbpassword',
];
```
4. Edit config/web.php:
```php
<?php
...

$config = [

    'components' => [
...
        'session' => [
            'class' => 'yii\web\DbSession',
        ],
...
    ],
];
...
```
5. Edit controllers/SiteController.php:
```php
...
class SiteController extends Controller
{
...
    public function actionTest1()
    {
        Yii::$app->session->setFlash('foo', 'bar');
        return $this->redirect(['/site/test2']);
    }

    public function actionTest2()
    {
        var_dump(Yii::$app->session->getAllFlashes());
    }
...
}
```
6. `./yii serve`
7. open `http://localhost:8080/index.php?r=site%2Ftest1`

### What is the expected result?

`array(1) { ["foo"]=> string(3) "bar" }`

### What do you get instead?

`array(0) { }`

### Additional info

Works ok if:

- Database is mysql
- `yii\web\Session` is used instead of DbSession

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.26
| PHP version      | 7.3.8
| Operating system |macOS 10.15, Centos 7
| PostgreSQL version | 11.3

samdark:
Is your session written correctly into database table?

yakninja:
Yes. Oddly enough, this happens only to flash, the rest of the sessions work just fine.

The flash data is correctly written to the db as well, you can see it if you comment out the redirect in site/test1.

If you call site/test2 just after that, the flash data disappears BEFORE the controller method is called. So my guess is that it is lost somewhere during the session initialization, but I'm not sure where exactly.

samdark:
And you're sure that if you switch to file sessions it works?

yakninja:
Yes. It works after commenting out `'class' => 'yii\web\DbSession'` line in config as well as after switching to mysql.

HenryVolkmer:
[...]

PR #17775, opened 2019-12-30
Session::updateFlashCounters() was called twice

HenryVolkmer:
one time in Session::init() and a second time in Session::open()
Removing the call in Session::open() fixes the issue.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17623

### DIFF
--- a/framework/web/Session.php
+++ b/framework/web/Session.php
@@ -138,7 +138,6 @@ class Session extends Component implements \IteratorAggregate, \ArrayAccess, \Co
 
         if ($this->getIsActive()) {
             Yii::info('Session started', __METHOD__);
-            $this->updateFlashCounters();
         } else {
             $error = error_get_last();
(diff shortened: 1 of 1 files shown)


rob006:
This does not look right. Call in `Session::init()` is for handling of misconfiguration (you're trying to init session when session is already active - you may have two concurrent session components) - in most cases it will never be called.

samdark:
Indeed, that is weird.

It looks like `init()` is called with session already opened. Then session is closed and after that it is opened again with `open()`.

@HenryVolkmer we need a use case to reproduce the issue. Based on what we see, there's something special about the whole situation. Ideally, start with clean basic application template and try to reproduce it with minimal code changes.

samdark:
Closing since no use-case was provided to reproduce the issue.

Issue #9014, opened 2015-07-05
Add case-insensitive validation in UniqueValidator and ExistValidator (for PostgreSQL)


YasserHassan:
Currently there seems to be no way to validate case-insensitive unique text values using UniqueValidator when working with PostgreSQL. This is particularly important in cases like detecting duplicate user names and emails during user registration. It would be good to provide a `caseSensitive` option as in as in Yii 1.0 to accommodate both case-sensitive and insensitive situations.

Faryshta:
There is a not so simple way to do it already.

Lets say you have a `User` model with an `username` attribute.

```php
public function getUsernameLowercase()
{
    return strtolower($this->username);
}

public function rules()
{
    return ['username', 'unique', 'targetAttribute' => ['usernameLowercase' => 'lower(username)']];
}
```

YasserHassan:
Thank you for the tip @Faryshta. `ExistValidator` has the same behavior too so I'll add it to the issue title.

samdark:
I don't think it makes sense to abstract it.

PR #17776, opened 2019-12-31
Update UniqueValidator.php

RichardPillay:
Added the ability to ignore case in the comparison, to be used when vetting email addresses.

No tests updated or submitted - I'm still learning. Would appreciate  help or advice in getting unit Testing and functional testing working

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ❌

### DIFF
--- a/framework/validators/UniqueValidator.php
+++ b/framework/validators/UniqueValidator.php
@@ -28,6 +28,8 @@ use yii\helpers\Inflector;
  * ['a1', 'unique']
  * // a1 needs to be unique, but column a2 will be used to check the uniqueness of the a1 value
  * ['a1', 'unique', 'targetAttribute' => 'a2']
+ * // email must be unique using the User model. In addition the case is ignored in the comparison: User@Domain is the same as user@domain
+ * ['email', 'unique', 'targetClass' => '\common\models\User', 'ignoreCase' => true, 'message' => 'This email address has already been taken.'],
  * // a1 and a2 need to be unique together, and they both will receive error message
  * [['a1', 'a2'], 'unique', 'targetAttribute' => ['a1', 'a2']]
  * // a1 and a2 need to be unique together, only a1 will receive error message
@@ -55,6 +57,11 @@ class UniqueValidator extends Validator
      * used to validate the uniqueness, while the array keys are the attributes whose values are to be validated.
      */
     public $targetAttribute;
+    /**
+     * @var boolean if set, the uniqueness will be determined in a case-insensitive fashion.
+     *
+     */
+    public $ignoreCase=false;
     /**
      * @var string|array|\Closure additional filter to be applied to the DB query used to check the uniqueness of the attribute value.
      * This can be a string or an array representing the additional query condition (refer to [[\yii\db\Query::where()]]
@@ -133,7 +140,14 @@ class UniqueValidator extends Validator
                 $this->addError($model, $attribute, Yii::t('yii', '{attribute} is invalid.'));
                 return;
             }
-            $conditions[] = [$key => $value];
+            if ($this->ignoreCase)
+            {
+                $newKey         = "lower(" . $key . ")";
+                $newValue       = "lower('" . $value . "')";
+                $conditions[] = $newKey . " = " . $newValue;
+            }
+            else
+                $conditions[] = [$key => $value];
         }
 
         $db = $targetClass::getDb();


samdark:
Thanks for contributing but I don't think that's a good way to solve the problem. `LOWER()` would make index useless for it won't work for any real production database.

RichardPillay:
Run the explain query on a production database. The query is wrapped in select exists(select *), which makes it choose to do a full table scan anyway.


Best Regards,Richard 

    On Tuesday, 31 December 2019, 23:40:08 GMT+11, Alexander Makarov <notifications@github.com> wrote:  
 
 
Thanks for contributing but I don't think that's a good way to solve the problem. LOWER() would make index useless for it won't work for any real production database.

rob006:
This may be tricky: https://eng.getwisdom.io/hacking-github-with-unicode-dotless-i/

RichardPillay:
> This may be tricky: https://eng.getwisdom.io/hacking-github-with-unicode-dotless-i/

Well known problem, unfortunately. With different locales, the user's typed-in email address may map to a different real email address when folded to lower-case. This is not a problem in reality, though, just something to be aware of and to take account of during development.

Email addresses need to folded to lower case using a consistent locale (the server's). It's possible, but not good practice to send the user's typed-in email address to the SMTP server, because that may change over the life of the system. 

With case folding occurring on the server, any mismatch between what is stored and what is sent to the email server is discovered during signup. That way, a user may not be able to sign up, but once signed up, he is always going to get his password reset emails.

Another thing that developers need to be aware of that although it increases server workload and response times, is this needs to be done server side rather than client-side. Doing this in java would always work for the user during signup, but would introduce exactly the same vulnerability you were worried about - the possibility of identity theft.

rob006:
Email was just an example, this problem affects everything, since `UniqueValidator` is generic validator. There should be at least a warning about potential problems caused by using this setting.

rob006:
Duplicates https://github.com/yiisoft/yii2/issues/9014

samdark:
@RichardPillay how about the way that was in #9014?

samdark:
In general, I'd avoid enhancements into 2.0 branch unless very necessary or useful but not currently achievable by other means...

RichardPillay:
> @RichardPillay how about the way that was in #9014?

I'm not exactly sure at the moment. I'd have to run some tests over the weekend to understand the performance difference between the two ways of doing things. My initial impression is that my solution and the one in #9014 were "kludges" in that they don't go far enough to do things properly. For example, on my system, I've got a lowercase index on the email column, but that is not reflected in my pull request.

In my case, doing this only a few days after starting out with Yii, I just followed the code until I found a place where I could alter the code. There may be a better way of doing it or a better place for the alteration.

The other problem is the one that you picked up on - the performance impact. In the case of UniqueValidator, the performance hit is not great (probably none, apart from the case conversion time) because of the way the SQL in the original query causes, in most cases, a full table scan. What the performance impact will be on ExistValidator I have no idea at the moment. 

Maybe what we need to do close this pull-request and I can start a new one in a while to make another email-specific change, with a little more thought behind it. Maybe the 'email' validator keyword can be used to trigger proper case-insensitive validation that is used specifically for email and that does the job without a full table scan.

What I do know is that different ways of doing this without a large performance impact are possible. I'm on Postgres, and it has a "citext" column type that would have achieved this without any code changes. The problem there is that it is not standard and would not work on the other DB types.

Even if we do close this pull-request, what I'm sure of is that the need for this is definitely there, so it deserves some time and effort to do it properly rather than a quick fix.

samdark:
Since there's no issue associated, let's keep this one open for further discussion. I've solved email case in the past with form validation rule that lowercases email address. This way you have lowercase addresses only and there's no need for any special indexes or additional SQL-based lowercasing.

RichardPillay:
That would be good - for most developers, storing lowercase email addresses is the way to go. The problem in this case is that there's a business rule to store and use the email address in the user's preferred manner. By extension, this means the additional checks on lowercase email names to pick up attempts at spoofing as well as accidental or intentional duplication of accounts.
For the moment I'll deal with it using a modified local copy, and when I have time I can try to find a robust solution that can be fed back to the framework. Realistically, that will be Yii3 rather than Yii2.


Best Regards,Richard 

    On Saturday, 11 January 2020, 00:43:05 GMT+11, Alexander Makarov <notifications@github.com> wrote:  
 
 
Since there's no issue associated, let's keep this one open for further discussion. I've solved email case in the past with form validation rule that lowercases email address. This way you have lowercase addresses only and there's no need for any special indexes or additional SQL-based lowercasing.

kamarton:
@RichardPillay add a new column with transformed value (example lower case) to table.

You can use this method to perform any additional transformation (eg. clean gmail addresses `john.smith@gmail.com === johnsmith@gmail.com === j.o.h.n.smith@gmail.com`).

How many database records? 10k 100k 1m? minimal storage overhead and resource requirements.

framework/validators/UniqueValidator.php
```diff
@@ -133,7 +140,14 @@ public function validateAttribute($model, $attribute)
                 $this->addError($model, $attribute, Yii::t('yii', '{attribute} is invalid.'));
                 return;
             }
-            $conditions[] = [$key => $value];
+            if ($this->ignoreCase)
+            {
+                $newKey         = "lower(" . $key . ")";
+                $newValue       = "lower('" . $value . "')";
```
kamarton: allowing to inject, see more https://www.yiiframework.com/doc/api/2.0/yii-db-expression

RichardPillay:
> @RichardPillay add a new column with transformed value (example lower case) to table.
> 
I've gone with the route of having a lower-case index, because of the other benefits it offers. Space-wise, it's not any different  from having a second column holding transformed values, but it offers several other advantages:

- it prevents multiple accounts with the same email address, which is a stated business rule
- it avoids a full table scan when searching for lowercased email addresses. This, though probably does not apply universally across different DBs.


> You can use this method to perform any additional transformation (eg. clean gmail addresses `john.smith@gmail.com === johnsmith@gmail.com === j.o.h.n.smith@gmail.com`).
> 
I had no idea that periods were ignored by GMail. Useful to know, but I wouldn't go putting that into the business rules, since it's not a standard, and Google could change it at any time. As it is, Google does that for GMail personal accounts, but not for GMail premium or enterprise accounts

framework/validators/UniqueValidator.php
RichardPillay: > allowing to inject, see more https://www.yiiframework.com/doc/api/2.0/yii-db-expression

I've only been a Yii user for a couple of weeks now, and didn't know this yet - I'm still going through the documentation. This, though, **is incredibly useful**. Thank's for the suggestion and the code example.

RichardPillay: This seems like a good idea, but I'm having a problem understanding it. The easiest way to get my head around it, is to look at what it's doing in terms of the final sql passed to the database, so I'll do that as soon as I can.

Do you have any advice for **where** the expression should be created? Right there within the ```if ($this->ignoreCase)``` block as I have it, or elsewhere. 

I also want to change the way the query is done at the moment. It currently produces this code:
```SELECT EXISTS(SELECT * FROM "User" WHERE lower("User"."UserEmail")=lower('NewUser@Example.com'))```

It would be better if it produced this code:
```SELECT EXISTS(SELECT lower("User"."UserEmail") FROM "User" WHERE lower("User"."UserEmail")=lower('NewUser@Example.com'))```

While it does not look like much of a change, on a larger dataset, it would cause the planner to choose to do an index scan rather than a table scan, though that would change from DB to DB.

**Comment by kamarton** [comment]
> * it prevents multiple accounts with the same email address, which is a stated business rule
> * it avoids a full table scan when searching for lowercased email addresses. This, though probably does not apply universally across different DBs.

add unique index. for example in mysql:
```sql
CREATE TABLE `mydb`.`email` ( `transformed_email` VARCHAR(256) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL , UNIQUE (`transformed_email`)) ENGINE = InnoDB;

mysql> insert into email set transformed_email='example@example.com';
Query OK, 1 row affected (0.03 sec)

mysql> insert into email set transformed_email='example@example.com';
ERROR 1062 (23000): Duplicate entry 'example@example.com' for key 'transformed_email'

mysql> explain select * from email where transformed_email='example@example.com';
+----+-------------+-------+------------+-------+-------------------+-------------------+---------+-------+------+----------+-------------+
| id | select_type | table | partitions | type  | possible_keys     | key               | key_len | ref   | rows | filtered | Extra       |
+----+-------------+-------+------------+-------+-------------------+-------------------+---------+-------+------+----------+-------------+
|  1 | SIMPLE      | email | NULL       | const | transformed_email | transformed_email | 770     | const |    1 |   100.00 | Using index |
+----+-------------+-------+------------+-------+-------------------+-------------------+---------+-------+------+----------+-------------+
1 row in set, 1 warning (0.02 sec)
```

use index for find, block duplicate value.

> since it's not a standard, and Google could change it at any time

If it is changed (which is unlikely), you will definitely be notified.

> but not for GMail premium or enterprise accounts.

Premium and enterprise accounts can only be usable with your own domain, so the `gmail.com` domain is not affected.

RichardPillay:
> ```sql
> CREATE TABLE `mydb`.`email` ( `transformed_email` VARCHAR(256) CHARACTER SET utf8 COLLATE utf8_bin NOT NULL , UNIQUE (`transformed_email`)) ENGINE = InnoDB;

I'm not sure what you were trying to explain with the mysql example. If it's to point out to me that all RDBMSs have unique indexes or other ways of preventing duplication, I do know that. What I was actually saying was that in Postgres, I'm using a feature where a text column can be stored in mixed case, but it can be indexed and searched in lower-case. This is what I was talking about when I said it is probably not the same in all DBs. As it happens, if I were using MySql, this wouldn't have been an issue for me as MySql has a column collation setting that gives you case-insensitive search by default. No need to create an additional transformed column or another index at all.



> If it is changed (which is unlikely), you will definitely be notified.

Again, not the point at all. The business rule says only one account is to be created with an email address. The rule does not say the system must try to police it or prevent a person from signing up more than once. Nothing in the business rules prevents someone from signing up for one account using AUser@GMail.com and for another using A.User@GMail.com. If somebody wants to deliberately circumvent the checks, that's fine, as long as they're willing to pay the double bill that will result. However, it's hard to justify double-billing someone when it's obvious that AUser@GMail.com and auser@gmail.com are one and the same person and the extra signup was accidental.
There's a clear email standard which states clearly that a period in an email address makes a difference. If Google decides in future, after the project is complete, to change it's policy, this project is not going to be restarted again. However, if by some miracle, the email standard changes, this project and hundreds of thousands of others around the world will have to be revisited.

In any case, it's not really an issue - the business rules or the project scope do not mention GMail-specific addressing at all, so I'm going to pretend I don't know any better and tiptoe quietly out of the room.

kamarton:
> signing up for one account using [AUser@GMail.com](mailto:AUser@GMail.com) and for another using [A.User@GMail.com](mailto:A.User@GMail.com)

https://www.lifewire.com/are-email-addresses-case-sensitive-1171111

>  The username, however, is case sensitive. 

Lowercase sensitivity can also be a problem if a service provider decides to be different in the future, which is the same rule as gmail, which ignores midpoints in the name.

kamarton:
> I'm using a feature where a text column can be stored in mixed case, but it can be indexed and searched in lower-case

this is make sense. https://stackoverflow.com/a/3980083/6856708

> This may be tricky: https://eng.getwisdom.io/hacking-github-with-unicode-dotless-i/

Will it still work correctly in this case? I think it's much safer to implement a higher-level transformation at a higher language level (understand PHP, Java, and more, not DBMS and the like).

kamarton:
@samdark @RichardPillay how about this:

```php
['column', 'unique', 'condition' => function($column, $value) {
  return new Expression('LOWER('.$column.') = LOWER(:value)', [':value' => $value]);
}]
```

RichardPillay:
> > This may be tricky: https://eng.getwisdom.io/hacking-github-with-unicode-dotless-i/
> 
> Will it still work correctly in this case? I think it's much safer to implement a higher-level transformation at a higher language level (understand PHP, Java, and more, not DBMS and the like).

I think part of the problem you face when designing the system is making those sorts of decisions. You have to tread a path, though, because during design you have to make high-level decisions that do not go to a low enough level to be implementation-specific. You have to say **what** to do, but stay away from **how** to do it. 
A portion of the design process spends time examining who the **users** of the system will be and how they relate to the **owners** of the system. During the design, email addresses were identified as a concern, because while there was an email standard, it was well known that SMTP servers are out there that choose to ignore or bend the standard (Microsoft was the one we identified - we didn't know about Google). Also, international character sets lead to some interesting situations, but these were too varied and were so often edge cases that it was decided not to spend any money trying to understand them further. The decision was made very early that the system would skirt around this by storing the email address exactly as the user entered it at signup (or as he later changed it to) and passing it to the SMTP server to handle it as it thought best. This gave rise to the problem where someone could accidentally sign up twice when really what they should have done was recover their account - this is the reason for the case-insensitive search. Intentional double-signups were not seen as an issue. What came out of the high-level design was a number of guidelines for the developers and these were to be used as the starting point for both functional tests and unit tests. Among these are:
- As we discussed before, email addresses must be stored **exactly** as the user typed them.
- Only the stored email address must be used for mailing (this moves responsibility for interperating the unicode onto the server hosting the email domain)
- On password reset:
-- No password recovery must be allowed, only a password reset, which may be the same password as before. (Note how this may not follow best security practices - it's intended to be more palatable to the identified target userbase)
-- The user must be emailed or texted a verification code created at the time of the reset request.
-- Verification codes must never be reused.
-- Verification codes are to be combined with a previously entered answer to one of several questions.

There is much more, but this gives you the idea - all of these say what a developer must produce, not how. However, they do give enough to allow tests to be developed that the developer has to conform to.

samdark:
Thought more about it. I think for now we'll not commit into having lowercase as an option for validator since there are multiple ways it could be achieved, each with its own benefits and shortcomings.

@RichardPillay thanks for raising the topic though.

p.s. gmail (even enterprise accounts) allow you to add anything after `+` in the first part of the email. For example, `sam@rmcreative.ru` is equal to `sam+yii@rmcreative.ru`.

RichardPillay:
> Thought more about it. I think for now we'll not commit into having lowercase as an option for validator since there are multiple ways it could be achieved, each with its own benefits and shortcomings.
> 
Yes, I agree. It seems simple when you first think of it, but then you realise it's not that easy to do it in a way that is portable.

> p.s. gmail (even enterprise accounts) allow you to add anything after `+` in the first part of the email. >For example, `sam@rmcreative.ru` is equal to `sam+yii@rmcreative.ru`.

I found out from this thread - and while it was a surprise, I am surprised at how useful it is for spam control.

PR #17781, opened 2020-01-04
Fix #17771: migrate/fresh was not returning exit code

### LINKED ISSUE #17771: BaseMigrateController::actionFresh doesn't return an exit code
```php
    public function actionFresh()
    {
        if (YII_ENV_PROD) {
            $this->stdout("YII_ENV is set to 'prod'.\nRefreshing migrations is not possible on production systems.\n");
            return ExitCode::OK;
        }

        if ($this->confirm(
            "Are you sure you want to reset the database and start the migration from the beginning?\nAll data will be lost irreversibly!")) {
            $this->truncateDatabase();
            $this->actionUp();
        } else {
            $this->stdout('Action was cancelled by user. Nothing has been performed.');
        }
    }
```
It would be nice if you could add `return` to the line `$this->actionUp();` sentence so that the result of applying the migrations is returned to the calling action.

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17771

### DIFF
--- a/framework/console/controllers/BaseMigrateController.php
+++ b/framework/console/controllers/BaseMigrateController.php
@@ -464,10 +464,11 @@ abstract class BaseMigrateController extends Controller
         if ($this->confirm(
             "Are you sure you want to reset the database and start the migration from the beginning?\nAll data will be lost irreversibly!")) {
             $this->truncateDatabase();
-            $this->actionUp();
-        } else {
-            $this->stdout('Action was cancelled by user. Nothing has been performed.');
+            return $this->actionUp();
         }
+
+        $this->stdout('Action was cancelled by user. Nothing has been performed.');
+        return ExitCode::OK;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


framework/console/controllers/BaseMigrateController.php
```diff
@@ -464,10 +464,11 @@ public function actionFresh()
...
             "Are you sure you want to reset the database and start the migration from the beginning?\nAll data will be lost irreversibly!")) {
             $this->truncateDatabase();
-            $this->actionUp();
-        } else {
-            $this->stdout('Action was cancelled by user. Nothing has been performed.');
+            return $this->actionUp();
         }
+
+        $this->stdout('Action was cancelled by user. Nothing has been performed.');
+        return ExitCode::OK;
```
xepozz: I think it should be non-ok code
For example, `echo "n" | sudo apt upgrade 1>/dev/null; echo $?` prints "1"

samdark: That would be a backwards compatibility break.

PR #17790, opened 2020-01-09
Fix #17760: Fix `JSON::encode()` for `\DateTimeInterface` under PHP 7.4

### LINKED ISSUE #17760: JSON::encode for DateTime
`echo \yii\helpers\Json::encode(['date' => new \DateTime()]);`

`{"date":{"date":"2019-12-23 15:52:27.717576","timezone_type":3,"timezone":"UTC"}}`

### What do you get instead?
`{"date":{}}`

But when I run native php `json_encode`:
`echo json_encode(['date' => new \DateTime()]);`

Returns expected result:
`{"date":{"date":"2019-12-23 15:52:27.717576","timezone_type":3,"timezone":"UTC"}}`

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.31
| PHP version      | 7.4.0 (cli)
| Operating system | Debian 10

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17760

### DIFF
--- a/framework/helpers/BaseJson.php
+++ b/framework/helpers/BaseJson.php
@@ -151,9 +151,17 @@ class BaseJson
                 $expressions['"' . $token . '"'] = $data->expression;
 
                 return $token;
-            } elseif ($data instanceof \JsonSerializable) {
+            }
+
+            if ($data instanceof \JsonSerializable) {
                 return static::processData($data->jsonSerialize(), $expressions, $expPrefix);
-            } elseif ($data instanceof Arrayable) {
+            }
+
+            if ($data instanceof \DateTimeInterface) {
+                return static::processData((array)$data, $expressions, $expPrefix);
+            }
+
+            if ($data instanceof Arrayable) {
                 $data = $data->toArray();
             } elseif ($data instanceof \SimpleXMLElement) {
                 $data = (array) $data;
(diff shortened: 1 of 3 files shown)


framework/helpers/BaseJson.php
```diff
@@ -151,9 +151,17 @@ protected static function processData($data, &$expressions, $expPrefix)
...
                 return $token;
-            } elseif ($data instanceof \JsonSerializable) {
+            }
+
+            if ($data instanceof \JsonSerializable) {
                 return static::processData($data->jsonSerialize(), $expressions, $expPrefix);
-            } elseif ($data instanceof Arrayable) {
+            }
+
+            if ($data instanceof \DateTimeInterface) {
```
xepozz: DateTimeInterface appeared only in 5.5 version (https://www.php.net/manual/en/class.datetimeinterface.php)
So it breaks BC, because Yii2 supports minimal 5.4 version https://github.com/yiisoft/yii2/blob/2.0.31/composer.json#L71

xepozz (changes requested):

framework/helpers/BaseJson.php
rob006: `instanceof` works fine with non-existing types - you don't need to have `DateTimeInterface`  to know that `$data` is not `DateTimeInterface`.

rob006 (approved):

framework/helpers/BaseJson.php
samdark: Since the issue doesn't affect PHP versions lower than 7.4, non-existence of `\DateTimeInterface` is fine.

PR #17798, opened 2020-01-13
Fix LogTarget to avoid creating folder

wapmorgan:
Fix `LogTarget` to avoid creating unnecessary folder in case when `php://stdout` or other `php://` log destination used.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️/❌
| Fixed issues  | none

**Problem**:
When `php://stdout` used as log destination file, modified lines created `php:` folder in current folder.
This condition checks log destination and avoid this directory creation.


framework/log/FileTarget.php
```diff
@@ -101,8 +101,10 @@ public function init()
      */
     public function export()
     {
-        $logPath = dirname($this->logFile);
-        FileHelper::createDirectory($logPath, $this->dirMode, true);
+        if (strncmp($this->logFile, 'php://', 6) !== 0) {
```
samdark: How about other stream types?

sup-ham:
I think FileTarget log should restricted to local file. you can create another eg. StdoutTarget if you want to use php://stdout and also STDOUT constant could be used instead as it always opened

samdark:
Initially it was meant to work with files only...

wapmorgan:
So, may be StreamTarget there would be a good solution?

wapmorgan:
Also, don't want to create another issue.
Another propose is `JsonFileTarget` and `JsonStreamTarget`, that will format messages as json-encoded objects, describing log message, per line.

samdark:
- JSON one looks way too specific to me to be part of the framework. 
- Are there any streams besides stdout that are usable as logger target?

wapmorgan:
Zip, Gzip, custom streams. Also, `stderr` is an option.

samdark:
Should we then check for these in `export()`?

wapmorgan:
I propose check for `://` in `target`, this is not used in local files path.

rob006:
> I propose check for `://` in `target`, this is not used in local files path.

`file://` is valid protocol, and it points to local filesystem.

samdark (approved):

samdark:
👍 Please add a line for CHANGELOG.md. Then I'll merge it.

samdark:
Thank you 👍

PR #17806, opened 2020-01-16
Fix #17805 Submit button with formaction outside of form tag

### LINKED ISSUE #17805: Submit button outside of `<form></form>` and `formaction`
Create a form, and put the submit button outside of the form.

And set the "form" attribute to the id of the form (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form).

Now set the formaction attribute on the button.

On pressing that button, it should submit the form to my formaction.

### What do you get instead?
It is submitted to the action of the form.

### Additional info
It uses `$form.on`, and then filters by `":submit"`, since the submit button is not in the form, it is never going to set `submitObject`.
https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.activeForm.js#L229

My current solution is:
```
$("[form=" + $form.attr("id") + "]:submit").on('mouseup.yiiActiveForm keyup.yiiActiveForm', function () {
    $form.data('yiiActiveForm').submitObject = $(this);
});
```
| Q                | A
| ---------------- | ---
| Yii version      | 2.0.23-dev
| PHP version      |  7.4.1
| Operating system | Win 10

diego-mathis:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | 
| Fixed issues  | #17805


My6UoT9:
This does not work correctly, at least the gii generator is broken with this patch.
Still fixing it, but no eta.

PR #17818, opened 2020-01-21
#17817 - Added helper class, which converts $query->where condition s…

alhimik1986:
…tructure to DataFilter structure.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌


samdark:
Thanks for sharing. That could be useful but since it's adding new feature it doesn't fit into Yii 2 feature freeze. I think it's better to release it as an extension.

PR #17826, opened 2020-01-23
fixed #17708 (intl number formatter)

### LINKED ISSUE #17708: FormatterNumberTest::testIntlAsScientific fails on GitLab

schmunk42:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️/❌ (if you rely on a bug ;))
| Tests pass?   | ✔️ (hopefully for all versions)
| Fixed issues  | #17708

Looks like the intl formatter and framework implementation are off-by one handling digits?!

btw: How is the intl extension disabled in tests.

---

Addon: Format is also different between both implementation `E5` vs. `E+5`, shouldn't that be fixed also or does this depend on a locale?

### DIFF
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -1223,10 +1223,6 @@ class Formatter extends Component
         $value = $this->normalizeNumericValue($value);
 
         if ($this->_intlLoaded) {
-            // fix for https://github.com/yiisoft/yii2/issues/17708
-            if ($decimals !== null) {
-                $decimals++;
-            }
             $f = $this->createNumberFormatter(NumberFormatter::SCIENTIFIC, $decimals, $options, $textOptions);
             if (($result = $f->format($value)) === false) {
                 throw new InvalidArgumentException('Formatting scientific number value failed: ' . $f->getErrorCode() . ' ' . $f->getErrorMessage());
--- a/framework/i18n/Formatter.php
+++ b/framework/i18n/Formatter.php
@@ -1219,9 +1219,14 @@ class Formatter extends Component
         if ($value === null) {
             return $this->nullDisplay;
         }
+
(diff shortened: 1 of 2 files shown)


schmunk42:
I removed the hack, so this is now just some extended tests.

schmunk42:
Oh my...

This `1.20E2` (intl)
https://github.com/yiisoft/yii2/pull/17826/files#diff-6f7bd2d4996479139d8bd3d4950e00d5R518

should actually be the same as `1.23E2` (non intl)
https://github.com/yiisoft/yii2/pull/17826/files#diff-6f7bd2d4996479139d8bd3d4950e00d5R540

but it is not :( 

with ICU greater 60 (with intl) it is `1.2E2` (the reason why I increased the digits in my initial commit)

So something is wrong with NumberFormatter and `intl`.

CC: @samdark

PR #17827, opened 2020-01-24
Add $strict property and adjust validation

adhayward:
Proposed fix for "StringValidator fails on integer #13327"

Adds a $strict property to StringValidator.
If true (default) then it validates as it does currently
If false, then it will use php's is_scalar to convert any scalar values to strings before the existing validation is applied.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #13327


samdark:
Added tests, cleaned up and merged. Thanks!

PR #17828, opened 2020-01-24
Fix yii\web\UploadedFile::saveAs() failing when error value in $_FILES entry is a string

haveyaseen:
This fixes a regression introduced in v2.0.32 (#17729).

`saveAs()` must not check strict equality as `error` can be a string. It returned early if `error === '0'` instead of the integer `0`.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | ❌


haveyaseen:
After rebasing onto master, Travis no longer fails on PHP 7.2 and 7.3.

samdark (approved):

samdark:
Thanks!

Issue #17829, opened 2020-01-24
ArrayHelper::filter deeper than two levels returns all data after 2nd level


rhertogh:
### What steps will reproduce the problem?
Let's say we have the following array:
```php
$array = [
    'A' => [1, 2],
    'B' => [
        'C' => 1,
        'D' => 2,
        'E' => [
            'F' => 1,
            'G' => 2,
            'H' => 3,
        ],
    ],
    'I' => 1,
];
```
If we filter 1 or 2 levels deep, only the specified field will be returned as expected, e.g. 
```php
ArrayHelper::filter($array, ['B.D'])
```
would return:
```php
[
  "B" => [
    "D" => 2, //Note that only D is set at this level (as expected)
  ]
]
```

Now if we filter 3 (or more) levels all data after the 2nd level will be returned, e.g. 
```php
ArrayHelper::filter($array, ['B.E.G']);
```
would return:

```php
[
  "B" => [
    "E" =>[
      "F" => 1, //Not expected
      "G" => 2,
      "H" => 3, //Not expected
    ],
  ],
]
```

### What is the expected result?

Only the specified node should be returned
```php
[
  "B" => [
    "E" =>[
      "G" => 2,
    ],
  ],
]
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.30
| PHP version      | 7.3.13
| Operating system | CentOS

PR #17830, opened 2020-01-24
Arrayhelper filter support more than 2 levels

rhertogh:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️(But only if `ArrayHelper::filter` was used in an unexpected way)
| Tests pass?   | ✔️
| Fixed issues  | #17829


framework/helpers/BaseArrayHelper.php
```diff
@@ -923,41 +923,51 @@ public static function isSubset($needles, $haystack, $strict = false)
...
+                    continue 2; //Jump to next filter
+                }
+                $nodeValue = $nodeValue[$key];
             }
-            if (!array_key_exists($globalKey, $result)) {
-                $result[$globalKey] = [];
+
+            //We've found a value now let's insert it
+            $resultNode = &$result;
+            foreach ($keys as $key) {
```
xepozz: Can you merge this loop with previous?

framework/helpers/BaseArrayHelper.php
```diff
@@ -923,41 +923,51 @@ public static function isSubset($needles, $haystack, $strict = false)
...
 
-        foreach ($forbiddenVars as $var) {
-            list($globalKey, $localKey) = $var;
-            if (array_key_exists($globalKey, $result)) {
-                unset($result[$globalKey][$localKey]);
+        foreach ($excludeFilters as $filter) {
+            $excludeNode = &$result;
+            $keys = explode('.', $filter);
+            $numKeys = count($keys);
+            for ($i = 0; $i < $numKeys; $i++) {
```
xepozz: Can you use `foreach` loop?

xepozz (changes requested):

framework/helpers/BaseArrayHelper.php
rhertogh: Nope, let's say we have a filter 'A.B' and the source does have key A but that level does not have key B in it, the resulting array would be set `[A => []]` while it should return `[]`.

framework/helpers/BaseArrayHelper.php
rhertogh: Now using foreach

framework/helpers/BaseArrayHelper.php
xepozz: Are tests fix this case?

rhertogh: Yes, the test would fail at line 1336 `$this->assertEquals(ArrayHelper::filter($array, ['A.X']), []);`.

samdark (approved):

samdark:
Merged. Thank you!

It would be cool to have it ported to Yii 3 https://github.com/yiisoft/arrays/issues/15

Issue #17726, opened 2019-12-11
Patch PHPUnit Getopt to be compatible with new PHP versions


samdark:
Need to patch PHPUnit 4 /vendor/phpunit/phpunit/src/Util/Getopt.php:38 in order to ensure forward compatibility:

https://travis-ci.com/yiisoft/yii2/jobs/265953981

Could be done similar to https://github.com/yiisoft/yii2/pull/17725

PR #17831, opened 2020-01-25
Apply patch to PHPUnit which replaces the deprecated each method

haveyaseen:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17726


samdark:
The patch seems to be correct. Travis is happy. 

Need to find out why AppVeyor started to fail though...

haveyaseen:
@samdark  In the previous patch there was still an issue with operator precedence:
- `$opt_arg = current($args) === false` should have been
- `($opt_arg = current($args)) === false`

I've now replaced all changes by cleaner versions that are easier to read.

samdark (approved):

samdark:
Thank you for handling this one!

Issue #17797, opened 2020-01-11
Wrong attributes for activeListInput


matiosfree:
Create an active form and add field like:
```php
<?= $form
   ->field($searchForm, 'some_attribute')
    ->checkboxList($options, ['name' => 'new_field_name', 'value' => 1]);
?>

```
### What is the expected result?
"div" tag should not have attributes "name" and "value"

### What do you get instead?
```html
<div id="searchform-some_attribute" name="new_field_name" value="1">
  <label><input type="checkbox" name="new_field_name[]" value="4"> 2835</label>
  <label><input type="checkbox" name="new_field_name[]" value="5"> 3014</label>
  <label><input type="checkbox" name="new_field_name[]" value="9"> 5050</label>
  <label><input type="checkbox" name="new_field_name[]" value="11"> 5730</label>
</div>
```

### Additional info
I think problem is here:
https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1825

Here should be 
```php
$name = ArrayHelper::remove($options, 'name', static::getInputName($model, $attribute));
$selection = ArrayHelper::remove($options, 'value', static::getInputName($model, $attribute));
```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.31

kamarton:
> "div" tag should not have attributes "name" and "value"

Why do I need the `name` and` value` on the `div` tag? Why don't you use `data-name` and` data-value` attributes? see more https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

alex-code:
Maybe `ArrayHelper::remove` should be used to get the options.

samdark:
I think the intent is to have these options applied to inputs, right @matiosfree?

matiosfree:
> 
> 
> I think the intent is to have these options applied to inputs, right @matiosfree?

Yes, but now they are applied to "div" also and that's is wrong.
We must have:
`<div id="searchform-some_attribute">`

matiosfree:
> Why do I need the `name` and` value` on the `div` tag?

What's the problem. "div" should NOT have these 2 attributes. But now it has.

mikehaertl:
@matiosfree I'd say this is an intentional design decision. If you want to set attributes on inputs use  `itemOptions` in your `$options` array instead:

```php
$form->checkboxList($items, ['itemOpions' => ['attrname' => 'attrvalue']]);
```

PR #17834, opened 2020-01-26
Fix for #17797, activeListInput options

alex-code:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17797


samdark:
Seems to fail tests...

alex-code:
My bad, had the wrong function which passed the `HtmlTest` I checked,

mikehaertl:
@samdark As commented in #17797 I'd say that the current behavior is fine. If you want to set attributes for the input elements, you can use `itemOptions`. This change will break BC.

alex-code:
@mikehaertl I think you've misunderstood.
This PR stops invalid attributes being added to the `<div/>` element.

mikehaertl:
@alex-code Right, now I see. It's about the specific attributes `name` and `value`. I think the change is fine then. But there is one edge case: If someone used this to set exactly these attributes on the `<div>` (e.g. for some javascript or whatever) it will no longer work. So IMO it's a minor but acceaptable BC break. Not sure if it's worth an upgrade note.

alex-code:
@mikehaertl Might be worth a note.
@samdark Do you want me to add a note?

rob006:
`name` and `value` attributes are not valid for `div` tag. No one should rely on this, note in changelog should be enough.

mikehaertl:
> name and value attributes are not valid for div tag. No one should rely on this, note in changelog should be enough.

Can you point me to a source for this? Never heard before that `<div name="myname">` should be invalid.

alex-code:
A `<div/>` element is allowed global attributes and (`class`, `id` and `slot`).
https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element
https://html.spec.whatwg.org/multipage/dom.html#global-attributes

mikehaertl:
@alex-code Does this really say, that it MUST NOT have any other attribute? I admit, that I've never heard that before. And I'd assume many other developers also. So why not put an update notice just to make sure? Yii2 is very old now and there is a pletora of apps out there. We should be friendly with our devs and not break their apps - even if the use an attribute that may not be perfect in line with the specs.

rob006:
> Does this really say, that it MUST NOT have any other attribute?

This spec whitelists allowed attributes - if it is not on the list of allowed attributes, then it is not allowed. For example https://validator.w3.org/ reports this as an error.

mikehaertl:
Ok, I'm convinced now, thanks for pointing that out.

samdark (approved):

samdark:
👍

PR #17838, opened 2020-01-29
updated README and added CHANGELOG entry [fix #17802]

### LINKED ISSUE #17802: Try GitHub actions instead of Travis
Travis build currently takes too much time. That's, of course, mainly because our tests aren't fast but time could be cut a bit by switching to GitHub actions.

simialbi:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17802


simialbi:
@samdark seems as you have to remove travis from needed checks because i disabled it by renaming `.travis.yml` to `.travis.yml.bak`

samdark:
I've removed changelog entry since it's not a change in the framework. I'll highlight it in development notes though.

Issue #17839, opened 2020-01-30
$mimeType image/heic


bscheshirwork:
### What steps will reproduce the problem?
try to load image use rule
```php
    const FILE_EXTENSIONS = ['png', 'jpg', 'gif', 'heic',];
    const FILE_MAX_SIZE = 1024 * 1024 * 5; // note: work only if it less than PHP upload_max_filesize
...
            [['file'], 'file', 'skipOnEmpty' => false, 'maxFiles' => static::MAX_FILES, 'extensions' => static::FILE_EXTENSIONS, 'maxSize' => static::FILE_MAX_SIZE],

```
### What is the expected result?
well done for all extensions
### What do you get instead?
heic produce error from backend (RU) "Разрешена загрузка файлов только со следующими расширениями: png, jpg, gif, heic."

### Additional info

try to check
https://github.com/yiisoft/yii2/blob/master/framework/helpers/mimeTypes.php
I can't find 'image/heic' - this value I can see in debug for $mimeType

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.31-dev PHP 7.4.2
| PHP version      | 7.4.2
| Operating system |alpine

bscheshirwork:
So... I think nobody use 'heic' in real web. Image pass as jpeg from device to form.

samdark:
Yes. Likely you're right.

lemarweb:
Hello from 2025.

I have the same issue with Yii 2.0.53-dev.
I tried mimeTypes **image/heif**, **image/heic** and extensions **heif**, **heic**. Same result...

There is a mobile app with Yii2 backend. User may to upload HEIC image from Gallery (iOS), but backend cannot validate it with FileValidator. Of course, it can be converted on the client side, but we want to get the original file.

It would be great to add heif/heic.

PR #17840, opened 2020-01-30
add mime type heic for image/heic

bscheshirwork:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️ ( I not found test for get mime from file)
| Fixed issues  |  #17839


samdark:
It won't work this way. Will be overwritten since we're taking mimes list from Apache server source code. That's how you can add your own:

https://github.com/yiisoft/yii2/blob/master/build/controllers/MimeTypeController.php#L51

bscheshirwork:
Agree

PR #17848, opened 2020-02-03
Fix #7622: Allow `yii\data\ArrayDataProvider` to control the sort flags

### LINKED ISSUE #7622: ArrayDataProvider is missing a property to control the sort flags for sortModels()
Could add a property

```
    /**
     * @var int sort flags to be used by [[sortModels()]]. This will be passed as the `` parameter
     * to [[ArrayHelper::multisort()]].
     */
    public $sortFlags = SORT_REGULAR;
```

or handle it via Sort class defined in `$attributes` property.

askobara:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #7622

### DIFF
--- a/framework/data/Sort.php
+++ b/framework/data/Sort.php
@@ -186,6 +186,12 @@ class Sort extends BaseObject
      */
     public $urlManager;
 
+    /**
+     * @var int Allow to control a value of the fourth parameter which will be
+     * passed to [[ArrayHelper::multisort()]]
+     * @since 2.0.33
+     */
+    public $sortFlags = SORT_REGULAR;
 
     /**
      * Normalizes the [[attributes]] property.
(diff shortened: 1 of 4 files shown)


samdark:
OK. That one is looking clear and doesn't break anything.

framework/data/Sort.php
```diff
@@ -186,6 +186,11 @@ class Sort extends BaseObject
      */
     public $urlManager;
 
+    /**
+     * @var int Allow to control a value of the fourth parameter which will be
+     * passed to [[ArrayHelper::multisort()]]
```
samdark: Need `@since` tag.

samdark (approved):

schmunk42 (approved):

framework/data/Sort.php
askobara: done

samdark:
Thanks!

PR #17854, opened 2020-02-11
Bug #17679: Fix oracle exception  "ORA-01461"

vinpel:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | added test for issue #17679


vinpel:
a lot of other tests for oracle don't seem to work on my environment, .

With oracle, without proper quote an object string name is interpreted as the name in UPPERCASE.

Sample : 
yii2\tests\framework\db\ConnectionTest.php  line 147, table and field name are not quoted :
it should be : 
```php
 $this->assertEquals(1, $connection->createCommand("SELECT COUNT(*) FROM [[profile]] WHERE description = 'test transaction';")->queryScalar());
```

is this my environnement or the oracle tests need to be fixed ?

samdark:
> is this my environnement or the oracle tests need to be fixed ?

Likely that's tests.

framework/db/Command.php
```diff
@@ -91,7 +91,7 @@ class Command extends Component
     /**
      * @var array pending parameters to be bound to the current PDO statement.
      */
-    private $_pendingParams = [];
+    protected $_pendingParams = [];
```
samdark: - Protected properties do not use `_` in the name.
- `@since` annotation should be added because it would be part of API.

tests/framework/db/oci/CommandTest.php
```diff
@@ -42,4 +43,36 @@ public function batchInsertSqlProvider()
...
+        }
+
+        $db->createCommand()->createTable('longstring', ['message' => Schema::TYPE_TEXT])->execute();
+        $long_data = str_pad('-', 4001, '-=', STR_PAD_LEFT);
+        try {
+            $db->createCommand()->insert('longstring', [
+                'message' => $long_data,
+            ])->execute();
+
+        } catch (\yii\db\Exception $exception) {
```
samdark: Looks suspicious. Do we expect that another exception would be triggered?

framework/db/oci/Command.php
```diff
@@ -0,0 +1,22 @@
...
+class Command extends \yii\db\Command
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function bindPendingParams()
+    {
+        foreach ($this->_pendingParams as $name => $value) {
+            if (\PDO::PARAM_STR === $value[1]) {
+                $this->pdoStatement->bindParam($name, $value[0], $value[1], strlen($value[0]));
```
samdark: It's not the same as in original pull request: https://github.com/yiisoft/yii2/pull/17679/files#diff-a4b7a5107e6519248e41616a4ea9641aR16. There it's ensured that `$value[0]` won't be modified as far as I can see it.

tests/framework/db/oci/CommandTest.php
vinpel: the error can be another things like an error during creation of the table, database gone or other stuff.

I made it that way so it's more clear what goes wrong

framework/db/oci/Command.php
vinpel: In the original code it's passed like that :

https://github.com/yiisoft/yii2/blob/7ec7fd11eecfa086cec4f828392a6cdaa2afbf28/framework/db/Command.php#L311-L320

samdark: I mean the code from another pull request.

tests/framework/db/oci/CommandTest.php
samdark: But is it throwing exception during the test?

framework/db/oci/Command.php
243083df: @vinpel Yes, but since we change method bindValue to bindParam, and bindParam accept value by reference. And ever if we know that this method not change value, but in future it can according to method signature. So, to avoid UB in future i copy variable.

tests/framework/db/oci/CommandTest.php
vinpel: If the query complete successfuly no exception at all
if the test fail, with the ora-01461 a specific message is throw
if the test fail for another reason the test will throwi it normal exception.


The test is specific to string > 4 000, if it fail i want to know if it's because of this test or another reason.

Is my logic good ?

samdark: Do you mean that the test is unstable and can randomly fail?

vinpel: no the test is reproductible and stable, no exception is leveled during this test because @243083df proposed fix work well

Before the fix this test has leveled an exception, who is catched by :

```php
 $this->assertNotContains('ORA-01461', substr($exception->getMessage(),0,100))
```
if this test fail i just wanted to make it the way i made it normaly you see immediatly if it's caused by the tested scenario or another stuff.

it's better if i made a " $this->fail($e->$exception->getMessage() " in the expcetion part ?

samdark: If the exception is never thrown under normal circumstances, I think that we should not add `try-catch` at all. Its presence makes reader wonder why it's there.

samdark (approved):

samdark:
Thank you, @vinpel, @243083df.

PR #17858, opened 2020-02-13
Fix #17667: Fix `CREATE INDEX` failure on sqlite when specifying schema

### LINKED ISSUE #17667: sqlite3: CREATE INDEX with schema.table syntax fails
# What steps will reproduce the problem?
Create two sqlite3 databases named `otherdb` and `mydb` and create a connection for yii2 to use the sqlite3 `mydb` database.

Then, create a migration like this:

```php
	public function safeUp()
	{
                 $this->db->createCommand("attach 'otherdb.db' as otherdb");
		$this->createTable('{{otherdb.users}}', [
			'id' => $this->primaryKey(),
			'email' => $this->string(100)->null(),
                 ]);
		$this->createIndex('my_index', 
                      '{{otherdb.users}}',
                      'email' );
        }
```

# What is the expected result?

An index is created in the `otherdb` database

# What do you get instead?

The following error:

```
Failed to prepare SQL: CREATE INDEX `my_index` ON `otherdb`.`users` (`email`) (.../vendor/yiisoft/yii2/db/Command.php:264)
```

# Cause

The correct syntax for the CREATE INDEX in this case is:
```
CREATE INDEX otherdb.`my_index` ON `users` (`email`)
```

# Solution:

Create a new function in `yii\db\sqlite\QueryBuilder`:

```php
    public function createIndex($name, $table, $columns, $unique = false)
    {
		$sql = ($unique ? 'CREATE UNIQUE INDEX ' : 'CREATE INDEX ')
            . $this->db->quoteTableName($name) . ' ON '
            . $this->db->quoteTableName($table)
            . ' (' . $this->buildColumns($columns) . ')';
        $sql = preg_replace_callback(
            '/(`.*`) ON (\{\{(%?)([\w\-]+)\}\}\.\{\{((%?)[\w\-]+)\\}\\})|(`.*`) ON (\{\{(%?)([\w\-]+)\.([\w\-]+)\\}\\})/',
            function ($matches) {
                if (!empty($matches[1])) {
					return $matches[4].".".$matches[1]
					 . " ON {{" .$matches[3].$matches[5] . "}}";
                } else if( !empty($matches[7]) ) {
					return $matches[10].".".$matches[7]
					 . " ON {{" .$matches[9].$matches[11] . "}}";
                }
            },
            $sql
        );
		return $sql;
    }
```

This problem and a lot more with sqlite3 is solved using the extension `yii2-sqlite3-full-support`.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.30-dev
| PHP version      |  7.3
| Operating system | Linux Debian 10

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ?
| Fixed issues  | #17667


xepozz:
What about to make some test cases?

samdark:
It is even failing existing test cases :)

PR #17867, opened 2020-02-17
WIP: Schema builder mysql [fix #11945]

### LINKED ISSUE #11945: Schema builder (MySQL): append in the wrong place

Create a migration like this:

```
        $this->createTable('table', [
            'name' => $this->string(50)->notNull()->append('CHARACTER SET ascii COLLATE ascii_general_ci')->comment('Property name'),
            'value' => $this->text()->comment('Property value')
        ], 'ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci');
```

The SQL command is execute successfully.
### What do you get instead?

An error is generated by MySQL because of the wrong order in the column definition:

```
Exception 'yii\db\Exception' with message 'SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'CHARACTER SET ascii COLLATE ascii_general_ci,
        `value` text COMMENT 'Property va' at line 2
The SQL being executed was: CREATE TABLE `table` (
        `name` varchar(50) NOT NULL COMMENT 'Property name' CHARACTER SET ascii COLLATE ascii_general_ci,
        `value` text COMMENT 'Property value'
) ENGINE=MyISAM DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci'
```

This is happening because the `append()` function literally appends the given string to the column definition (maybe I'm doing something wrong?). I don't know if it's legal in other RDBMS, but in MySQL it's not.
### Additional info

| Q | A |
| --- | --- |
| Yii version | 2.0.9 |
| PHP version | 7.0.7 |
| Operating system | Latest Debian testing |
| MySQL version | 5.6.30 |

Great software and thanks for your help.

simialbi:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #11945


samdark:
Thank you!

Issue #17875, opened 2020-02-19
Saving of uploaded files does not works.


eLFuvo:
After upgrading yii2 from 2.0.30 to 2.0.32 normal uploading files (by POST method) does not works.
I have get a error: 
> copy(/var/www/site.com/tmp/phpQwOV1W): failed to open stream: No such file or directory

An error occurs in this method:
```php
class UploadedFile extends BaseObject
{
...
  protected function copyTempFile($targetFile)
  {
   if (!is_resource($this->_tempResource)) {
 ->     return $this->isUploadedFile($this->tempName) && copy($this->tempName, $targetFile);
  }
```
I have changed `copy` to `move_uploaded_file` and it works fine

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.32
| PHP version      | 7.1.8
| Operating system | Linux

samdark:
Related to https://github.com/yiisoft/yii2/pull/17729. @sup-ham would you please take a look?

samdark:
@eLFuvo would you please check if https://github.com/yiisoft/yii2/pull/17877 fixes the issue?

eLFuvo:
I have checked this fixes in real-world project. 
Works properly.

Thanks for your support!

PR #17877, opened 2020-02-19
Revert `move_uploaded_file()` function for saving uploaded files when POST request

hyuzea:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17875


sup-ham:
Should we delete these unused methods?

samdark:
Yes.

samdark:
Thank you!

rob006:
This change should be mentioned in `UPGRADE.md`.

samdark:
@rob006 this one is actually makes everything compatible again.

rob006:
@samdark It removes protected methods, which is against [BC policy](https://github.com/yiisoft/yii2/blob/master/docs/internals/bc.md).

It also reverts fix for issue that is marked as fixed, and it is not clear from changelog whether this issue is still fixed, or not.

samdark:
@rob006 you're correct about removal of protected methods but the fix isn't removed.

samdark:
1f8ca7b7b153a5590e2e4928e2d28604c3331ccf

Issue #17878, opened 2020-02-19
\yii\web\Request does not detect cross-origin Ajax requests


dicrtarasov:
\yii\web\Request:getIsAjax()
```php
public function getIsAjax()
{
    return $this->headers->get('X-Requested-With') === 'XMLHttpRequest';
}
```

But, for the cross-origin ajax requests, send by browser `XMLHttpRequest` (or modern `fetch API`)  dows not contains `XMLHttpRequest` headers.  

How to reproduce:
1) configure Cors filter to allow accepting cross-origin requests.
For Apache:
```
Header setifempty Access-Control-Allow-Origin "*"
Header setifempty Access-Control-Allow-Methods "GET"
```
For Yii:
```php
public function behaviors()
{
    return ['cors' => \yii\filters\Cors::class];
}
```

2) send cross-origin request from browser console to different your site:
```
fetch('http://mysite.ru')
```
Request headers:
```
GET / HTTP/1.1
Host: mysite.ru
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache
Sec-Fetch-Dest: empty
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.106 Safari/537.36
DNT: 1
Accept: */*
Origin: http://localhost:3000
Sec-Fetch-Site: cross-site
Sec-Fetch-Mode: cors
Referer: http://localhost:3000
Accept-Encoding: gzip, deflate, br
Accept-Language: ru,en;q=0.9,uk;q=0.8,und;q=0.7
```

There is no longer the usual `XMLHttpRequest` header, but, there are 2 new headers:
`Sec-Fetch-Mode: cors` and 
`Sec-Fetch-Site: cross-site`

`Yii::$app->request->isAjax` give `false`..... I propose to add `Sec-Fetch-Mode` header in `getIsAjax`.

dicrtarasov:
Because of Request::getIsAjax() is false for cors request, so authentication behavior different - it send redirects to login page for ajax requests :(

samdark:
Good idea. Would you like to do a pull request?

dicrtarasov:
I found a workaroud. Unlike the regular ajax-regular, where setting a X-Requested-With by hand is prohibited, in cross-origin requests - it's allowed!
```
fetch(url, {
    method: 'GET',
    headers: {'X-Requested-With': 'XMLHttpRequest'}
})
[...]

PR #17880, opened 2020-02-19
jsAjax for CORS

dicrtarasov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | https://github.com/yiisoft/yii2/issues/17878

### DIFF
--- a/framework/web/Request.php
+++ b/framework/web/Request.php
@@ -478,7 +478,10 @@ class Request extends \yii\base\Request
      */
     public function getIsAjax()
     {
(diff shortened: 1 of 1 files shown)


samdark:
I've re-read the spec. These headers may be omitted and are always omitted for non-HTTPS so actually your workaround would be way more reliable.

I won't merge this pull request. The workaround is to be documented instead.

PR #17881, opened 2020-02-19
queryScalar() fix

brandonkelly:
If an exception ends up getting thrown from `$this->createCommand()` within `Query::queryScalar()`, then the query’s `select`, `orderBy`, `limit`, and `offset` params will never get reverted to their original state. `select` will still be set to `[$selectExpression]`, and the others will be `null`.

With this PR, any throwables/exceptions that are thrown by `createCommand()` will be caught, and then re-thrown once those params have been reverted to their original state.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -


samdark (approved):

samdark:
👍

brandonkelly:
Thanks!

PR #17882, opened 2020-02-20
detect cors Ajax requests

dicrtarasov:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | #17878

Cross domain requests contains `Origin` header, which is always differ from Request::hostInfo


samdark:
Would that work for non-CORS fetch()?

dicrtarasov:
Yes, because the old check of `X-Requested-Withrequest` remained in place as before.

samdark:
But fetch doesn't send `X-Requested-With`...

dicrtarasov:
Oh, yes, agree ..... checked just now.
`fetch` does not send any headers that may distinguish it request from ordinal non-ajax.
But ... at least we can detect cross domain requests by the difference between `Origin` and `Request::hostInfo`. It's already a little better than nothing at all :)

samdark:
Added tests, merged. Thanks!

Issue #17884, opened 2020-02-21
Console Table renders '0' as empty string


mikehaertl:
# What steps will reproduce the problem?

```
echo Table::widget([
    'rows' => [
        ['Total', '0'],
    ],
]);
```
### What is the expected result?

```
╔══════════════════════╤══════╗
║ Total                │    0 ║
╚══════════════════════╧══════╝
```
### What do you get instead?

```
╔══════════════════════╤══════╗
║ Total                │      ║
╚══════════════════════╧══════╝
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.32
| PHP version      |  7.3.11
| Operating system | Debian

### Suggested fix

The rewrite of row values in `setRows()` should be modified:

```php
    public function setRows(array $rows)
    {
        $this->rows = array_map(function($row) {
            return array_map(function($value) {
                return $value ?: ' ';
            }, array_values($row));
        }, $rows);
        return $this;
    }
```

It should better use this in the inner `array_map()`:

```php
return empty($value) && $value !== '0' ? ' ' : $value;
```

lubosdz:
perhaps simpler: `trim($value) === '' ? ' ' : $value`

PR #17885, opened 2020-02-21
Issue #17884 Fix 0 values in console table

mikehaertl:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | (unsure)
| Tests pass?   | ✔️
| Fixed issues  | #17884

### DIFF
--- a/framework/console/widgets/Table.php
+++ b/framework/console/widgets/Table.php
@@ -133,7 +133,7 @@ class Table extends Widget
     {
         $this->rows = array_map(function($row) {
             return array_map(function($value) {
-                return $value || $value === 0 || $value === '0' ? $value : ' ';
+                return empty($value) && !is_numeric($value) ? ' ' : $value;
             }, array_values($row));
         }, $rows);
         return $this;
--- a/framework/console/widgets/Table.php
+++ b/framework/console/widgets/Table.php
@@ -133,7 +133,7 @@ class Table extends Widget
     {
         $this->rows = array_map(function($row) {
             return array_map(function($value) {
-                return $value ?: ' ';
+                return $value || $value === 0 || $value === '0' ? $value : ' ';
             }, array_values($row));
         }, $rows);
         return $this;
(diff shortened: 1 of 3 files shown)


samdark (approved):
👍 Please add a line for changelog.

framework/console/widgets/Table.php
```diff
@@ -133,7 +133,7 @@ public function setRows(array $rows)
     {
         $this->rows = array_map(function($row) {
             return array_map(function($value) {
-                return $value ?: ' ';
+                return $value || $value === 0 || $value === '0' ? $value : ' ';
```
rob006: Wouldn't be better to use `strlen($value) > 0`? Current condition does not support `0.0`, and there may be even more edge cases.

mikehaertl: It's one of these decade old weirdnesses of PHP that we don't have a simple way to differentiate `0`,  `0.00`, `0.000` etc. from other "falsey" values like `false`, `null`, `[]` and `''`. 

Given that `$value` can also be an array here, `strlen()` alone is not enough. I'll update like follows (better suggestion welcome):

```php
return $value || $value !== [] && (string) $value !== '' ? $value : ' ';
```

I also considered this alternative but both are somewhat ugly:

```php
return $value || !is_array($value) && strlen($value) !==0 ? $value : ' ';
```

mikehaertl:
@samdark Tests failing now at some other unrelated place. Could you restart them? I may have to tweak the expected string though (no time to set up tests locally, sorry)

framework/console/widgets/Table.php
rob006: What is the purpose of this line in the first place? Why empty cells should be normalized to one space?

mikehaertl: I didn't look why this is required - but that is what it did before, so I didn't change it. Probably the table would not rendered correctly if there isn't at least 1 single space character even if the cell is empty.

rob006: Right now `$value || ` condition seems to be redundant.

mikehaertl: Final version:

```php
return empty($value) && !is_numeric($value) ? ' ' : $value;
```

samdark:
Thank you!

Issue #16721, opened 2018-09-17
Allow configuration of the cache used by the url manager


hofrob:
### What steps will reproduce the problem?
* use a yii application in production with url manager cache and extensive use of the cache in general (where you can't just flush everything) 
* endpoints (actions/urls) get added/removed through the development process
* try to access the new urls
* 404 because url rules are cached

### What is the expected result?
* one should be able to flush the url manager cache only
* it should be possible to config the url manager cache (duration, dependency)

### Additional info
https://github.com/yiisoft/yii2/blob/master/framework/web/UrlManager.php#L278

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
 PHP version      | 7.2.9
| Operating system | docker/alpine

paweljankowiak06:
Same could be said about rbac DbManager. There is no possibility to clear cache only for texts categories.

hofrob:
Now that I think about it.

In yii\rbac\DbManager you could just use the same cache with a different config:

https://github.com/yiisoft/yii2/blob/master/framework/rbac/DbManager.php#L64

The cache property accepts an array as config for the cache.

And in yii\web\UrlManager you can use a dedicated cache application component. Still confusing that there are two different approaches to the same problem.

rob006:
> And in yii\web\UrlManager you can use a dedicated cache application component. Still confusing that there are two different approaches to the same problem.

This actually should be fixed - `Instance::ensure()` should be consistently used in such cases.

We could also make `$cacheKey` public, so cache could be revalidated by changing this value. But this will break BC, so probably not an option in 2.0.

paweljankowiak06:
There are few places where such behavior is implemented. Maybe some common trait would be a good solution.

samdark:
About original issue. You can specify separate cache instance there:

- https://github.com/yiisoft/yii2/blob/master/framework/rbac/DbManager.php#L83
- https://github.com/yiisoft/yii2/blob/master/framework/web/UrlManager.php#L132

So I don't see a problem with flushing this separate instance.
 
@mahkali what for?

mahkali:
@samdark flushing the cache on change would indeed be more efficient as we don't require to check a dependency on every request.

PR #17890, opened 2020-02-22
Use `Instance::ensure()` to initialize `UrlManager::$cache`

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ?
| Tests pass?   | ✔️
| Fixed issues  | #16721

This allows to pass configuration component for `UrlManager::$cache` (like in many other similar cases) so you could create custom cache instance just for `UrlManger` without a need to declare it as global component (see https://github.com/yiisoft/yii2/issues/16721#issuecomment-422143911 and https://github.com/yiisoft/yii2/issues/16721#issuecomment-422315982).

Technically this could be a BC break - previously misconfiguration (for example typo in component name) was silently ignored. Now it will throw an exception if configured cache component cannot be initialized. IMO this is a bugfix - configuration errors should not be hidden in this way, especially in such cases where lack of cache affects performance and it is easy to miss it.


samdark (approved):

samdark:
👍

Issue #17889, opened 2020-02-22
Serializer should recursively serialize Models


patacca:
### What steps will reproduce the problem?
Consider the following test
You can put it inside test/framework/rest/SerializerTest.php

```php
class TestModel5 extends Model
{
    public static $fields = ['field3', 'field4'];
    public static $extraFields = [];

    public $field3 = 'test3';
    public $field4 = 'test4';
    public $extraField4 = 'testExtra2';

    public function fields()
    {
        $fields = static::$fields;
        $fields['testDataProvider'] = function() {
            return $this->getTestDataProvider();
        };
        $fields['testDataProviderArray'] = function() {
            return [$this->getTestDataProvider(), 'testKey' => $this->getTestDataProvider()];
        };
        return $fields;
    }

    public function extraFields()
    {
        return static::$extraFields;
    }

    public function getTestDataProvider()
    {
        return new ArrayDataProvider([
            'allModels' => [
                ['id' => 1, 'username' => 'Bob'],
                ['id' => 2, 'username' => 'Tom'],
            ],
            'pagination' => [
                'route' => '/',
                'pageSize' => 1,
                'page' => 1,
            ],
        ]);
    }
}

public function testSerializeModelComplex()
{
	$serializer = new Serializer();
	$serializer->preserveKeys = true;
	$model = new TestModel5();
	
	$this->assertEquals([
		'field3' => 'test3',
		'field4' => 'test4',
		'testDataProvider' => [1 => ['id' => 2, 'username' => 'Tom']],
		'testDataProviderArray' => [
			[1 => ['id' => 2, 'username' => 'Tom']],
			'testKey' => [1 => ['id' => 2, 'username' => 'Tom']]
		]
	],
	$serializer->serialize($model));
}

```
### What is the expected result?

It should pass the test

### What do you get instead?

It doesn't pass the test

### Additional info

The problem is that inside `Serializer::serializeModel` the model is converted to an array without considering that some of its elements might not be directly convertable to an array. `ArrayableTrait::toArray` in fact consider only these types `Arrayable`, `array` and `JsonSerializable` and for the others it just call `ArrayHelper::toArray`. If we pass a `DataProviderInterface`, as in the above test, it will not be serialized

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.33
| PHP version      |  7.4.1
| Operating system | Arch Linux (Linux 5.4.10-arch1-1 09/01/2020 x86_64)

xepozz:
[...]

PR #17891, opened 2020-02-22
yii\base\ArrayableTrait::toArray serializes objects through Serializer

patacca:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17889 

It fixed the issue but I don't know if it is ok to use Serializer in this class

### DIFF
--- a/framework/base/ArrayableTrait.php
+++ b/framework/base/ArrayableTrait.php
@@ -11,6 +11,7 @@ use Yii;
 use yii\helpers\ArrayHelper;
 use yii\web\Link;
 use yii\web\Linkable;
+use yii\rest\Serializer;
 
 /**
  * ArrayableTrait provides a common implementation of the [[Arrayable]] interface.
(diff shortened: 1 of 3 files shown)


framework/base/ArrayableTrait.php
```diff
@@ -11,6 +11,7 @@
 use yii\helpers\ArrayHelper;
 use yii\web\Link;
 use yii\web\Linkable;
+use yii\rest\Serializer;
```
samdark: Overall that doesn't look too good. Is there another way to solve it?

patacca: See this https://github.com/yiisoft/yii2/issues/17889#issuecomment-590565449

framework/base/ArrayableTrait.php
```diff
@@ -121,6 +122,8 @@ public function extraFields()
      */
     public function toArray(array $fields = [], array $expand = [], $recursive = true)
     {
+        $serializer = new Serializer();
```
romkatsu: Duplication. Maybe it's better to put it in the method?

```php
private function createSerializer() {
   $serializer = new Serializer();
   $serializer->preserveKeys = true;
   return $serializer;
}

```

romkatsu (changes requested):

framework/base/ArrayableTrait.php
patacca: I think this isn't going to be merged. Objects serializable should just be treated as non-serializable
See this https://github.com/yiisoft/yii2/issues/17889#issuecomment-591518449

PR #17897, opened 2020-02-26
fix: #13749 Yii opens db connection even when hits query cache

shushenghong:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️ remove $rawsql parameter from yii\db\Command::getCacheKey()
| Tests pass?   | ✔️
| Fixed issues  | #13749 Yii opens db connection even when hits query cache

### DIFF
--- a/framework/db/Command.php
+++ b/framework/db/Command.php
@@ -1148,7 +1148,6 @@ class Command extends Component
                 }
             }
         }
-        $rawSql = $rawSql ?: $this->getRawSql();
 
         $this->prepare(true);
 
--- a/framework/db/Command.php
+++ b/framework/db/Command.php
@@ -1140,8 +1140,7 @@ class Command extends Component
             if (is_array($info)) {
                 /* @var $cache \yii\caching\CacheInterface */
                 $cache = $info[0];
-                $rawSql = $rawSql ?: $this->getRawSql();
-                $cacheKey = $this->getCacheKey($method, $fetchMode, $rawSql);
+                $cacheKey = $this->getCacheKey($method, $fetchMode);
                 $result = $cache->get($cacheKey);
                 if (is_array($result) && isset($result[0])) {
                     Yii::debug('Query result served from cache', 'yii\db\Command::query');
@@ -1191,15 +1190,17 @@ class Command extends Component
      * @return array the cache key
      * @since 2.0.16
      */
-    protected function getCacheKey($method, $fetchMode, $rawSql)
+    protected function getCacheKey($method, $fetchMode)
     {
+        ksort($this->params);
         return [
             __CLASS__,
             $method,
             $fetchMode,
             $this->db->dsn,
             $this->db->username,
-            $rawSql,
+            $this->getSql(),
+            json_encode($this->params),
         ];
     }
 
--- a/framework/db/Command.php
+++ b/framework/db/Command.php
@@ -1140,7 +1140,7 @@ class Command extends Component
             if (is_array($info)) {
                 /* @var $cache \yii\caching\CacheInterface */
                 $cache = $info[0];
-                $cacheKey = $this->getCacheKey($method, $fetchMode, $rawSql);
+                $cacheKey = $this->getCacheKey($method, $fetchMode);
                 $result = $cache->get($cacheKey);
                 if (is_array($result) && isset($result[0])) {
                     Yii::debug('Query result served from cache', 'yii\db\Command::query');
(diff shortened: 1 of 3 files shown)


framework/db/Command.php
```diff
@@ -1140,8 +1140,7 @@ protected function queryInternal($method, $fetchMode = null)
             if (is_array($info)) {
                 /* @var $cache \yii\caching\CacheInterface */
                 $cache = $info[0];
-                $rawSql = $rawSql ?: $this->getRawSql();
-                $cacheKey = $this->getCacheKey($method, $fetchMode, $rawSql);
+                $cacheKey = $this->getCacheKey($method, $fetchMode);
```
xepozz: I think the line `$rawSql = $rawSql ?: $this->getRawSql();` shouldn't be removed, because if `$rawSql` is not empty we no need to recalculate it again.

xepozz (changes requested):
Could you please write testcase?

framework/db/mysql/QueryBuilder.php
```diff
@@ -390,7 +390,15 @@ private function getColumnDefinition($table, $column)
      */
     private function supportsFractionalSeconds()
     {
-        $version = $this->db->getSlavePdo()->getAttribute(\PDO::ATTR_SERVER_VERSION);
+        // use cache to prevent open mysql connection
+        // https://github.com/yiisoft/yii2/issues/13749#issuecomment-481657224
+        $key = [__METHOD__, $this->db->dsn];
```
rob006: Shouldn't we also include `$this->db->username` here, same as in `Command::getCacheKey()`?

framework/db/mysql/QueryBuilder.php
```diff
@@ -390,7 +390,15 @@ private function getColumnDefinition($table, $column)
...
+        $version = isset(\Yii::$app->cache) ? \Yii::$app->cache->get($key) : null;
+        if( !$version ) {
+            $version = $this->db->getSlavePdo()->getAttribute(\PDO::ATTR_SERVER_VERSION);
+            isset(\Yii::$app->cache) ? \Yii::$app->cache->set($key, $version) : null;
```
rob006: Please use regular `if` here - ternary operator is unreadable and hides complexity.

framework/db/Command.php
shushenghong: getRawSql is the place which cause db connection (quoteValue), but you are right, if there is no cache, rawsql maybe empty, and the code after cache may cause error

framework/db/Command.php
```diff
@@ -1191,15 +1190,17 @@ protected function queryInternal($method, $fetchMode = null)
      * @return array the cache key
      * @since 2.0.16
      */
-    protected function getCacheKey($method, $fetchMode, $rawSql)
+    protected function getCacheKey($method, $fetchMode)
```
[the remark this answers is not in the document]
shushenghong: yes, thank you for your advice

framework/db/Command.php
```diff
@@ -1191,15 +1190,17 @@ protected function queryInternal($method, $fetchMode = null)
...
     {
+        ksort($this->params);
```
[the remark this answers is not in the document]
shushenghong: ok

framework/db/mysql/QueryBuilder.php
shushenghong: i think mysql server version is only related with dsn

framework/db/mysql/QueryBuilder.php
```diff
@@ -390,7 +390,15 @@ private function getColumnDefinition($table, $column)
      */
     private function supportsFractionalSeconds()
     {
-        $version = $this->db->getSlavePdo()->getAttribute(\PDO::ATTR_SERVER_VERSION);
+        // use cache to prevent open mysql connection
+        // https://github.com/yiisoft/yii2/issues/13749#issuecomment-481657224
+        $key = [__METHOD__, $this->db->dsn];
+        $version = isset(\Yii::$app->cache) ? \Yii::$app->cache->get($key) : null;
```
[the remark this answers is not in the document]
shushenghong: yes, change to schema cache maybe more clear

framework/db/Command.php
shushenghong: > getRawSql is the place which cause db connection (quoteValue), but you are right, if there is no cache, rawsql maybe empty, and the code after cache may cause error

after research the code, rawsql is only used for log, if profile enabled it will be set, so there is no problem

samdark:
Merged and adjusted. Thanks.

shushenghong:
@samdark 
https://github.com/yiisoft/yii2/blob/fbdf4645109039552e165096bb78dc144ad52259/framework/db/Command.php#L1144
need change add $rawSql parameter, since you add backwards compatibility to getCacheKey method.

atomrus1993:
bug!
https://github.com/yiisoft/yii2/runs/483600374#step:15:31

Issue #16092, opened 2018-04-12
ActiveRecord: innerJoinWith('foo')->joinWith('foo.bar') = duplicate join on 'foo'


jurchiks:
Have the following relations defined (some obvious things ommitted for brevity):

```php
/** @property int $bar_id */
class Foo extends ActiveRecord {
    public function getBar() {
        return $this->hasOne(Bar::class, ['id' => 'bar_id']);
    }
}

/**
 * @property int $id
 * @property int $baz_id
 */
class Bar extends ActiveRecord {
    public function getBaz() {
        return $this->hasOne(Baz::class, ['id' => 'baz_id']);
    }
}

/** @property int $id */
class Baz extends ActiveRecord {}
```
Execute a query like this:

```php
Foo::find()
    ->innerJoinWith('bar')
    ->joinWith('bar.baz')
    ->all();
```

### What is the expected result?

```php
SELECT foo.*
FROM foo
INNER JOIN bar ON bar.id = foo.bar_id
LEFT JOIN baz ON baz.id = bar.baz_id
```

### What do you get instead?

```php
SELECT foo.*
FROM foo
INNER JOIN bar ON bar.id = foo.bar_id
LEFT JOIN bar ON bar.id = foo.bar_id <----- the problem
LEFT JOIN baz ON baz.id = bar.baz_id
```

I did not find a solution for this in the docs, nor did I find existing bugs that describe this. Perhaps this is already solved in some way that is simply not documented, but if so, then it is still an issue as this seemingly intuitive code just doesn't work.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.0.28
| Operating system | Windows 7 Enterprise

jurchiks:
This works fine if you write `innerJoinWith('bar', 'bar.baz')` or its equivalent chained call, so there are some checks for duplicates, but they apparently consider the LEFT/INNER keyword, which they really shoudn't.
The only part that should signify uniqueness is the name (or alias) of the join, i.e. in this case it should be "bar" and "baz", no matter if they're LEFT or INNER.
If you were to write `innerJoinWith('bar aliasOne', 'bar.baz aliasTwo')` then the uniqueness should only check "aliasOne" and "aliasTwo", as that's what the database would check. You can make a hundred joins on the same table with LEFT/INNER/FULL joins, but as long as their aliases (or names, if no alias is specified) are unique, it doesn't matter.

[...]

PR #17900, opened 2020-02-27
Fix 16092: Fix duplicate joins in usage of joinWith

germanow:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #16092

### DIFF
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -486,6 +486,16 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         }
         $this->join = array_values($uniqueJoins);
 
+        // https://github.com/yiisoft/yii2/issues/16092
+        $uniqueJoinsByTableName = [];
+        for ($i = 0; $i < count($this->join); $i++) {
+            $tableName = $this->join[$i][1];
+            if (!array_key_exists($tableName, $uniqueJoinsByTableName)) {
+                $uniqueJoinsByTableName[$tableName] = $this->join[$i];
+            }
+        }
+        $this->join = array_values($uniqueJoinsByTableName);
+
         if (!empty($join)) {
             // append explicit join to joinWith()
             // https://github.com/yiisoft/yii2/issues/2880
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -489,7 +489,7 @@ class ActiveQuery extends Query implements ActiveQueryInterface
         // https://github.com/yiisoft/yii2/issues/16092
         $uniqueJoinsByTableName = [];
         for ($i = 0; $i < count($this->join); $i++) {
-            $tableName = $this->join[$i][1];
+            $tableName = serialize($this->join[$i][1]);
             if (!array_key_exists($tableName, $uniqueJoinsByTableName)) {
                 $uniqueJoinsByTableName[$tableName] = $this->join[$i];
             }
--- a/framework/db/ActiveQuery.php
+++ b/framework/db/ActiveQuery.php
@@ -488,10 +488,10 @@ class ActiveQuery extends Query implements ActiveQueryInterface
 
         // https://github.com/yiisoft/yii2/issues/16092
         $uniqueJoinsByTableName = [];
-        for ($i = 0; $i < count($this->join); $i++) {
-            $tableName = serialize($this->join[$i][1]);
+        foreach ($this->join as $config) {
+            $tableName = serialize($config[1]);
             if (!array_key_exists($tableName, $uniqueJoinsByTableName)) {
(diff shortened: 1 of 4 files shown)


tests/framework/db/ActiveQueryTest.php
```diff
@@ -136,6 +136,26 @@ public function testInnerJoinWith()
         ], $result->joinWith);
     }
 
+    public function testBuildJoinWithRemoveDuplicateJoinByTableName()
+    {
+        $query = new ActiveQuery(Customer::className());
+        $query->innerJoinWith('orders')
+            ->joinWith('orders.orderItems');
```
xepozz: What about closure as a `on` statement?
https://www.yiiframework.com/doc/guide/2.0/en/db-active-record#relation-table-aliases

xepozz (changes requested):

tests/framework/db/ActiveQueryTest.php
germanow: Do you mean this case?
```php
$query = new ActiveQuery(Customer::className());
$query->innerJoinWith('orders')
    ->joinWith('orders', function($q) {
        $q->joinWith('orderItems');
    });
```
Works well.

xepozz:
```php
$query->joinWith([
    'orders' => function ($q) {
        $q->from(['o' => Order::tableName()]);
    },
])
```
and
```php
$query->joinWith([
    'orders' => function ($q) use ($value) {
        $q->where(['key' => $value]);
    },
])
```

germanow: I added tests for it to ActiveRecordTest. Why build failed after that? Failure in `FixtureControllerTest`: 
```
1) yiiunit\framework\console\controllers\FixtureControllerTest::testLoadActiveFixtureSequence
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'yiiunit\data\console\controllers\fixtures\SecondIndependentActiveFixture'
-    1 => 'yiiunit\data\console\controllers\fixtures\FirstIndependentActiveFixture'
+    0 => 'yiiunit\data\console\controllers\fixtures\FirstIndependentActiveFixture'
+    1 => 'yiiunit\data\console\controllers\fixtures\SecondIndependentActiveFixture'
     2 => 'yiiunit\data\console\controll...ixture'
 )

/home/runner/work/yii2/yii2/tests/framework/console/controllers/FixtureControllerTest.php:244
/home/runner/work/yii2/yii2/vendor/phpunit/phpunit/phpunit:52
```

samdark: That test was incorrect. I've fixed it.

tests/framework/db/ActiveRecordTest.php
```diff
@@ -561,6 +561,134 @@ public function testJoinWith()
         $this->assertEquals(2, $orders[0]->items[0]->category->id);
     }
 
+    /**
+     * @depends testJoinWith
+     */
+    public function testJoinWithDuplicate()
```
xepozz: Could you please use `@dataProvider` if it's possible or split this long method to small methods if not?

framework/db/ActiveQuery.php
```diff
@@ -486,6 +486,16 @@ private function buildJoinWith()
         }
         $this->join = array_values($uniqueJoins);
 
+        // https://github.com/yiisoft/yii2/issues/16092
+        $uniqueJoinsByTableName = [];
+        for ($i = 0; $i < count($this->join); $i++) {
```
xepozz: I think here `foreach` will be best look. What do you think?

xepozz (changes requested):
Everything looks good

tests/framework/db/ActiveRecordTest.php
germanow: I'll split this method

xepozz: This method contains eight test cases

xepozz (approved):

samdark (approved):

samdark:
👍

framework/db/ActiveQuery.php
```diff
@@ -486,6 +486,16 @@ private function buildJoinWith()
         }
         $this->join = array_values($uniqueJoins);
 
+        // https://github.com/yiisoft/yii2/issues/16092
+        $uniqueJoinsByTableName = [];
```
jurchiks: Does this work for aliased tables or only table names?
It should work primarily on aliases, and only use table name if an alias is not present.

germanow: It works for both aliased tables and table names.

PR #17910, opened 2020-03-05
Reset the DB schema, transaction, and driver name on connection close

brandonkelly:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17909

PR #17935, opened 2020-03-25
Clear a couple more Connection caches on close()

brandonkelly:
This is a continuation of #17910. If a DB connection is closed, it should be prepared for the possibility that it will be reopened with a different driver set, in which case any cached quoted table/column names should be cleared out.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |


samdark:
👍

PR #17939, opened 2020-03-26
Fixed yii\web\Response event classes

brandonkelly:
The event constants in `yii\web\Response` claim to use a `ResponseEvent` class that doesn’t exist. Fixed to reference `yii\base\Event` instead. (Fully qualified class name because there’s no other reference to `yii\base\Event` in that class, so importing it would just end up getting automatically removed by PhpStorm code reformatting.)

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |


samdark:
Thanks!

PR #17944, opened 2020-03-29
Fix #17810 EachValidator crashing with uninitialized typed properties

### LINKED ISSUE #17810: EachValidator doesn't work with typed properties
Set property type, e.g.
```php
public array $dummy;
```

Validate it using each validator
```php
[
   'dummy',
   'each',
   'rule' => [
       'boolean',
   ],
],
```

No errors

### What do you get instead?
TypeError
https://github.com/yiisoft/yii2/blob/master/framework/validators/EachValidator.php#L143

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.31
| PHP version      | 7.4.1

ricardomm85:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |  #17810 

When EachValidator is used in a uninitialized typed property (PHP 7.4) it generates a fatal error.
"Fatal error: Uncaught Error: Typed property XXX must not be accessed before initialization..."

Keep in mind that uninitialized is not the same as null.
Just added a check ( isset() ) before performing EachValidator logic.

### DIFF
--- a/framework/validators/EachValidator.php
+++ b/framework/validators/EachValidator.php
@@ -8,6 +8,7 @@
 namespace yii\validators;
 
 use Yii;
+use yii\base\DynamicModel;
 use yii\base\InvalidConfigException;
 use yii\base\Model;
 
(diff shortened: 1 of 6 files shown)


ricardomm85:
Sorry haven't check the test correctly

ricardomm85:
ok, all test passed correctly

framework/validators/EachValidator.php
```diff
@@ -128,6 +128,11 @@ private function createEmbeddedValidator($model)
      */
     public function validateAttribute($model, $attribute)
     {
+        // PHP 7.4 Typed properties not initialized
+        if(property_exists($model, $attribute) && !isset($model->$attribute)) {
```
bizley:
```suggestion
        if (property_exists($model, $attribute) && !isset($model->$attribute)) {
```

bizley: Also, I suggest to throw exception here, like InvalidArgument or something, because it will silently validate positively everything instead.

bizley (changes requested):

framework/validators/EachValidator.php
ricardomm85: You mean throwing some thing  like this?


```suggestion
        if (property_exists($model, $attribute) && !isset($model->$attribute)) {
            throw new InvalidArgumentException("Uninitialized property. Please set '$attribute' to null, 0, [], ...");
        }
```

bizley: Yes, just without suggesting anything about the setting.
OR maybe better idea - when uninitialized property is detected how about setting it to `null` explicitly. This will work as a certain BC fix and in most cases keep the original workflow. I'm not sure about the drawbacks of it though (apart from obvious).

ricardomm85: Just tried the idea but it fails because in the test we have 
`public array $arrayTypedProperty;`

It don't accepts null as a value (?array)

I think going for **InvalidArgument** is the best approach.

```suggestion
        if (property_exists($model, $attribute) && !isset($model->$attribute)) {
            throw new InvalidArgumentException('Uninitialized property.');
        }
```

bizley: `is_array()` should work here and setting it to `[]`.

ricardomm85: Sorry I'm a bit lost.

The problem is, when we have an uninitialized typed property, no matter the type (array, string, Interface, ...), we cannot access it. So cannot perform is_array() on it. Neither we can set it to null because we don't know if the property is nullable (?array, ?string, ?Interface, ...)

Following with you idea of setting to null, what about that?

```suggestion
        if (property_exists($model, $attribute) && !isset($model->$attribute)) {
            $value = null;
        } else {
            $value = $model->$attribute;
        }
```
With the current workflow it will add an error ('{value} is invalid')

bizley: Unless someone got better idea.

ricardomm85:
Digging deeper I realize that it's not just the EachValidator failing to access an uninitialized typed property. It will happen with the rest of Validators too. Whenever we try to access  **$model->$attribute** 

🤔

ricardomm85:
Crazy idea. Check on Validator::init() if all the typed properties are set. 

Draft code (not checked):
Screenshot from 2020-03-30 16-19-24

rob006:
What is the point of converting specific `TypeError` to more generic `InvalidArgumentException`? Both will essentially work in the same way from user perspective - you need to have try-catch for every `validate()` call.

bizley:
> 
> 
> What is the point of converting specific `TypeError` to more generic `InvalidArgumentException`? Both will essentially work in the same way from user perspective - you need to have try-catch for every `validate()` call.

For me it's what you see on screen:
Fatal Error = OMG, Yii crashed
vs
Yii Exception = Oh, I f*cked up, need to fix

And yes, this is basically the same but with InvalidArgument exception we can additionally hint that this is happening because of validation attempt.

rob006:
@bizley For me it looks like "we're aware of that problem, but instead fixing it, we'll put a label on it". If we want to do anything special for incomplete objects (with uninitialized properties), we should rather add validation error for them.

BTW: this PR is unrelated to #17810, since #17810 is about reusing model property for storing value of different type during validation (if you have `public array $dummy;` where `$dummy` is array of booleans, validator will try to assign these booleans one by one to `$dummy`).

bizley:
> 
> 
> [...] we should rather add validation error for them.
> 

Love that idea, this is the way to go.

ricardomm85:
@rob006 my bad I misunderstood the issue #17810  
I leave aside the "uninitialized typed prop"

Let me redo the test and fix the error ;)

samdark:
Ready for another review?

ricardomm85:
> Ready for another review?

Yes, please! @rob006 @bizley

bizley:
LGTM. What about this @rob006's idea to add validation error for uninitialized typed properties in PHP7.4?

ricardomm85:
@bizley  I've have some doubts on how to tackle this. 

Those variables are spread all around the validators, so we should do something on the Validator::init(), correct?

I'm not sure how to prevent the validations, without throwing an exception...

Maybe we should open another issue to discuss this and when we reach an agreement, prepare another PR

bizley:
@ricardomm85 Yes, let's.

bizley (approved):

samdark (approved):

samdark:
👍

PR #17948, opened 2020-03-30
Ignore set_time_limit() errors

brandonkelly:
We have a customer whose hosting environment has a custom PHP module installed, which doesn’t allow `set_time_limit(0)` to be called, even though the function exists. Ignoring the error with `@` has proven to work for them.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -430,9 +430,8 @@ class Response extends \yii\base\Response
             return;
         }
 
-        if (function_exists('set_time_limit')) {
-            @set_time_limit(0); // Reset time limit for big files
-        } else {
+        // Try to reset time limit for big files
+        if (!function_exists('set_time_limit') || !@set_time_limit(0)) {
             Yii::warning('set_time_limit() is not available', __METHOD__);
         }
 
--- a/framework/web/Response.php
+++ b/framework/web/Response.php
@@ -431,7 +431,7 @@ class Response extends \yii\base\Response
         }
 
         if (function_exists('set_time_limit')) {
-            set_time_limit(0); // Reset time limit for big files
+            @set_time_limit(0); // Reset time limit for big files
         } else {
             Yii::warning('set_time_limit() is not available', __METHOD__);
         }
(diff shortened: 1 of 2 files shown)


samdark:
That is a bit controversial. Any function could be "checked" like that. We can not add checks everywhere. In addition, there is a standard `disable_functions` and there is already a check for it.

samdark:
Since it's very non-standard environment, I wonder if there is a workaround for it?

framework/web/Response.php
```diff
@@ -431,7 +431,7 @@ protected function sendContent()
         }
 
         if (function_exists('set_time_limit')) {
-            set_time_limit(0); // Reset time limit for big files
+            @set_time_limit(0); // Reset time limit for big files
         } else {
             Yii::warning('set_time_limit() is not available', __METHOD__);
         }
```
rob006: I suggest:

```php
        // try to reset time limit for big files
        if (!@set_time_limit(0)) {
            Yii::warning('set_time_limit() is not available', __METHOD__);
        }
```

angrybrad:
For some context, it's a custom PHP extension called "phplimiter" on a shared hosting environment.

It _does_ allow `set_time_limit` to be called, but it has a range restriction on what the values could be - in this case, between 1-180.

Since `0` is out of that range, a Yii `ErrorException` is raised.

samdark:
@rob006 have you encountered such restrictions as well?

rob006:
> @rob006 have you encountered such restrictions as well?

No, but since we already have check for unavailable `set_time_limit()`, we could extend it to more cases.

framework/web/Response.php
brandonkelly: Good point, should be a warning. Done now (07e8a74d219ff6334cf75a6915e18208bcad4337).

brandonkelly:
@rob006 All good here?

rob006 (approved):

samdark (approved):

samdark:
Merged. Thanks!

Issue #17942, opened 2020-03-28
Db cache memory exausted 


mauriziocingolani:
I updated to Yii 2.0.33 (then 2.0.34) and all my applications broke.
I'm getting "Allowed memory size of 268435456 bytes exhausted (tried to allocate 20480 bytes)" from line 132 of yiisoft/Yii2/caching/DbCache.php.
Here's my cache and db component configuration:

`return [
'components' => [
	'cache' => [
            'class' => 'yii\caching\DbCache',
            'cacheTable' =>  'YiiCache',
	],
        'db' => [
...,
    'enableSchemaCache' => true, 
    'schemaCacheDuration' => 3600 * 24,
        ]
];`

Flushing the cache doesn't fix the problem.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.33
| PHP version    | 7.0.33  
| Operating system| linux

alex-code:
What DB are you using?

mauriziocingolani:
@alex-code MySQL/MariaDB/Percona

mauriziocingolani:
Another server gave me a little more information on this error:

`Fatal error: Uncaught yii\base\ErrorException: Unknown: Cannot call session save handler in a recursive manner in Unknown:0 Stack trace: #0 [internal function]: yii\base\ErrorHandler->handleError(2, 'Unknown: Cannot...', 'Unknown', 0, NULL) #1 {main} thrown in Unknown on line 0`

samdark:
Interesting. It is reproducible with basic application template by enabling DB cache?

mauriziocingolani:
@samdark Yes it is. Cache table:

```CREATE TABLE YiiCache (
  id char(128) NOT NULL,
  expire int(11) DEFAULT NULL,
  data longblob DEFAULT NULL
) ENGINE=MyISAM DEFAULT CHARSET=latin1;```

Db configuration:

```return [
    'class' => 'yii\db\Connection',
    'dsn' => 'mysql:host=localhost;dbname=...',
    'username' => '...',
    'password' => '...',
    'charset' => 'utf8',
    // Schema cache options (for production environment)
    //'enableSchemaCache' => true,
    'schemaCacheDuration' => 60,
    'schemaCache' => 'cache',
];```

Cache component:
```        'cache' => [
            'class' => 'yii\caching\DbCache',
            'cacheTable' => 'YiiCache'
        ],```

[...]

PR #17949, opened 2020-03-30
Fix for #17942

alex-code:
Fix for #17942.
Don't use cache if it's `DbSchema` to avoid loop.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17942

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 Change Log
 2.0.35 under development
 ------------------------
 
-- no changes in this release.
+- Bug #17942: Fix for `DbCache` loop in MySQL `QueryBuilder` (alex-code)
 
 
 2.0.34 March 26, 2020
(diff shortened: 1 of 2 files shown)


xepozz (approved):

samdark (approved):

samdark:
Thank you!

razorsharpshady:
> Fix for #17942.
> Don't use cache if it's `DbSchema` to avoid loop.
> 
> Q	A
> Is bugfix?	
> Breaks BC?	
> Tests pass?	
> Fixed issues	#17942

Does this fix mean to never set **enableSchemaCache** to **true**?

alex-code:
To avoid a loop this fix just means this function `QueryBuilder::supportsFractionalSeconds()` can't use the cache if it's an instance of `DbCache`.
It doesn't affect anything else.

PR #17957, opened 2020-04-02
Fix nested forms issue with yii.gridView.js

Ziggizag:
Appending dynamically created filter form to <body> rather than to $grid element results in leaner structure free of nested form issue in case GridView widget with active filters is positioned inside another form. This change has no other impact on GridView performance.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -

### DIFF
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -160,7 +160,7 @@
                 'class': 'gridview-filter-form',
                 style: 'display:none',
                 'data-pjax': ''
-            }).appendTo($grid);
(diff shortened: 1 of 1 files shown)


samdark:
Seems tests are not alright: https://github.com/yiisoft/yii2/pull/17957/checks?check_run_id=556131958

Ziggizag:
I see - pJax issue :/

Issue #17941, opened 2020-03-28
Conflicting controller creation (mapped, namespaced and under sub_module)!!!


djidji01:
### What steps will reproduce the problem?
I think the issue #13433 should be revisited. instead of architectural or configuration problem, its looks the issue resides on method `yii\base\Module::createController` at this level:

`// module and controller map take precedence

        if (isset($this->controllerMap[$id])) {
            $controller = Yii::createObject($this->controllerMap[$id], [$id, $this]);
            return [$controller, $route];
        }`

  
### What is the expected result?

[$controller, $route] with `$route` runnable by the `$controller` after evaluating all possibilities or `false`.

### What do you get instead?

it return a `$controller` and `$route` irrespective of whether the `$route` is valid or not for the `$controller` which stops evaluation of any other possibility (e.g:  `$module` with `$id` or sub-namespaced (subdirectory) `$controller` which may run the `$route` successfully without throwing exception.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.32?
| PHP version      | 7.4.2
| Operating system | windows 10 64Bit

samdark:
How would you solve it?

djidji01:
> 
> 
> How would you solve it?

I was trying to modify a bit the algorythm of the mentioned method as bellow which is working till now!!!!:

```
public function createController($route)
    {
        if ($route === '') {
            $route = $this->defaultRoute;
        }
        // double slashes or leading/ending slashes may cause substr problem
        $route = trim($route, '/');
        if (strpos($route, '//') !== false) {
            return false;
        }

        $prettyUrl=Yii::$app->getUrlManager()->enablePrettyUrl;

      // Every segment of the $route will be countable
        $segments=explode('/', $route);
        $id='';
        $i=0;
        $count=count($segments);
        while ($i<= ($count-1)) {

            // $id keeps growing to n segments as long as no controller creatde for n-1 segment
            $id=$i===0?$segments[$i]:($id.'/'.$segments[$i]);

            // $actionId set to n+1 segment of current iteration or empty string (this will be equivalent to $defaultAction)
            $actionId= isset($segments[$i+1])?$segments[$i+1]:'';

           // this is to introduce a way of prettyfiying url for named arguments for $action method parameters when no rule defined in the urlManager.
            $params=$prettyUrl?array_slice($segments,$i+2):[];

           // Priority to Mapped controller ( note that here $id may contains '/' as stated in the issue #13433
[...]

Issue #13433, opened 2017-01-23
Unable to map controller when in subdirectory of a module


nkostadinov:
### What steps will reproduce the problem?

Create an "identity" module and a controller inside "controllers/api/OauthController". Try to map the controller to something else e.g.

```
    'modules' => [
        'identity' => [
            'class' => IdentityModule::className(),
            'controllerMap' => [
                'api/oauth' => '\app\controllers\OauthController',
            ],
        ],
    ],
```

### What is the expected result?

When requesting "/identity/api/oauth" to call the action from the mapped controller

### What do you get instead?

The original controller is executed. It is impossible to map a controller inside a subfolder 

| Yii version      | 2.0.11-dev
| PHP version      | 7.0.0
| Operating system | Windows

TerraSkye:
the key of the controller  in the config cannot be api/oauth. 
normaly u would define it as ```oauth``` 
and if you want the url to be /identity/api/oauth. u would solve that through the url rules.

SilverFire:
@TerraSkye is right.

nkostadinov:
Well I don`t quite agree. I see no problem in defining the key of the controller including the path since its the only way to specify this controller like a namespace. Specifying only the name cannot solve the cases where I have controllers with the same name in different folders. It seems a little inconsistent to be able to only map controllers in the root of controllers folder.
What @TerraSkye proposed is just a workaround which is not the same as mapping the controller. Why is there mapping of controllers if we can solve this via url rules ?
May be I do not understand how this is gonna work, could you provide example configuration ? Just let me clarify that i have logic in the module which need to be kept, and I want to access $this->module in my controller pointing to the right module in this case 'identity'.

TerraSkye:
your case would be a submodule. api in your identity module.

nkostadinov:
If the framework allows controllers in subfolders there should be a way to map that. that's my point and I'm aware of all the other workarounds. I think it should just be modified to allow 'path/controller' in the map section just like it tries to find it in createController.

TerraSkye:
But thej you will never know when to redirect to it.

I have a module named moduleA
A contoller in it :
api/auth

ModuleA has a submodule api. With a controller auth. 

Wich one gets used.?

In my opinion. The controllers should be flat as a key value manner.

nkostadinov:
[...]

PR #17959, opened 2020-04-07
Issues #17941, #13433 Conflicting controller creation (mapped, namespaced and under sub_module)!!! proposed changes

djidji01:
Here is my proposed final changes on issues #17941, #13433 mentioned above:

- The Module::createController method will return [$controller, $actionId, $args] which are already validate for running to gather or false.
- The Module::createController will be used for both routing and url creation respectively for requested url and designated route if no urlManager rule specified.
- Named parameter will be prettified in the order declared in the action method and if parameter is required will have to be declared in the route for url creation and in the url for parsing rather than getting them form $_GET.
- Caching and Performance ??!!

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌/❌
| New feature?  | ❌/❌
| Breaks BC?    | ❌/❌
| Tests pass?   | ❌/❌
| Fixed issues  | #17941, #13433

### DIFF
--- a/framework/BaseYii.php
+++ b/framework/BaseYii.php
@@ -93,7 +93,7 @@ class BaseYii
      */
     public static function getVersion()
     {
-        return '2.0.34';
+        return '2.0.35-dev';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


djidji01:
I do not know why these red crosses of test not passing although with basic project every thing is working as I am expecting (specifically when working with gii and debug modules as I am expecting). may be not knowing how to test!!!

PR #17965, opened 2020-04-09
Fix #17964: Fix `UrlValidator` for url like `http://localhost`

### LINKED ISSUE #17964: UrlValidator returns false for http://localhost

```php
use yii\validators\UrlValidator;

// returns false
(new UrlValidator)->validate('http://localhost');
```

true

### What do you get instead?

false

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      | 7.4
| Operating system | Linux Mint 19.3

germanow:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17964


rob006:
I would be careful with that - allowing localhost may create security issues (SSRF).

samdark:
See https://github.com/yiisoft/validator/issues/4

Issue #17816, opened 2020-01-20
MSSQL drop column error: "The object 'DF__yyyyy__xxxxx__151B244E' is dependent on column"


michalwoz:
# What steps will reproduce the problem?
Dropping one of table columns in migration on MSSQL. table has many columns, not the onle one being dropped.

### What is the expected result?
column drop

### What do you get instead?
exception:
```
drop column xxxxx from table {{%yyyyy}} ...Exception: SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]The object 'DF__yyyyy__xxxxx__151B244E' is dependent on column 'xxxxx'.
The SQL being executed was: ALTER TABLE [yyyyy] DROP COLUMN [xxxxx] (app\vendor\yiisoft\yii2\db\Schema.php:674)
```

This issue is the same as https://github.com/yiisoft/yii2/issues/17035 but I think , that it was incorrectly classified as the attempt of removing single/last column in table. The exception occurs, not because it is the only column, but because there exists DF constrain.

Ugly workaround is to find constrain name first, than remove it and finally drop the column:
```php
    $migration->execute(Yii::$app->db->getQueryBuilder()->dropDefaultValue( // on MSSQL you need to drop default constrain before dropping column
        Yii::$app->db->createCommand("select dobj.name as def_name from sys.columns col 
            left outer join sys.objects dobj on dobj.object_id = col.default_object_id and dobj.type = 'D' 
            where col.object_id = object_id(N'dbo.$table') and col.name = '$column' and dobj.name is not null"
        )->queryScalar(), $table));
    $migration->dropColumn($table, $column);
```
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.25
| PHP version      | 
| Operating system |

samdark:
DF constrain = default value?

michalwoz:
Yes, this constrain is automatically created by MSSQL for each column which has default value.

simialbi:
It's not yii related. I always solved it like this:
```php
if ($this->getDb()->driverName === 'sqlsrv' || $this->getDb()->driverName === 'mssql' || $this->getDb()->driverName === 'dblib') {
	foreach (array_keys($fields) as $field) {
		$sql = <<<SQL
DECLARE @ObjectName NVARCHAR(100);
SELECT @ObjectName = OBJECT_NAME([default_object_id]) FROM SYS.COLUMNS
[...]

Issue #16481, opened 2018-07-04
DbManager::removeAll() MSSQL error.


sdlins:
I have this migration for RBAC:

```php
    public function safeUp()
    {
        /** @var \yii\rbac\DbManager $rbac */
        $rbac = \Yii::$app->get('authManagerAlteracaoDeCronograma');

        $rbac->removeAll(); // <-- results in error
         ...
    }
```
Line `$rbac->removeAll();` results in:

```shell
SQLSTATE[21000]: [Microsoft][ODBC Driver 13 for SQL Server][SQL Server]Subquery returned more than 1 value. This is not permitted when the subquery follows =, !=, <, <= , >, >= or when the subquery is used as an expression.
The SQL being executed was: DELETE FROM [ALTERACAOCRONO_RBAC_ITEM] (C:\...\vendor\yiisoft\yii2\db\Schema.php:664)
#0 C:\...\vendor\yiisoft\yii2\db\Command.php(1263): yii\db\Schema->convertException(Object(PDOException), 'DELETE FROM [AL...')
#1 C:\...\vendor\yiisoft\yii2\db\Command.php(1075): yii\db\Command->internalExecute('DELETE FROM [AL...')
#2 C:\...\vendor\yiisoft\yii2\rbac\DbManager.php(901): yii\db\Command->execute()
#3 C:\...\common\migrations\alteracao-de-cronograma\m180704_153354_rbac_permissoes_e_regramento.php(18): yii\rbac\DbManager->removeAll()
#4 C:\...\vendor\yiisoft\yii2\db\Migration.php(114): m180704_153354_rbac_permissoes_e_regramento->safeUp()
#5 C:\...\vendor\yiisoft\yii2\console\controllers\BaseMigrateController.php(725): yii\db\Migration->up()
#6 C:\...\vendor\yiisoft\yii2\console\controllers\BaseMigrateController.php(199): yii\console\controllers\BaseMigrateController->migrateUp('m180704_153354_...')
#7 [internal function]: yii\console\controllers\BaseMigrateController->actionUp(0)
#8 C:\...\vendor\yiisoft\yii2\base\InlineAction.php(57): call_user_func_array(Array, Array)
#9 C:\...\vendor\yiisoft\yii2\base\Controller.php(157): yii\base\InlineAction->runWithParams(Array)
#10 C:\...\vendor\yiisoft\yii2\console\Controller.php(148): yii\base\Controller->runAction('', Array)
[...]

PR #17966, opened 2020-04-09
split original trigger into two distinct ones.

achretien:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️/❌
| Fixed issues  | #16481 

I add the findForeignKeyName method because in SQL Server the name of the foreign key is generated  when not named at creation (same story than #17816)


samdark:
Merged. Thank you!

PR #17967, opened 2020-04-09
Document [alias=>table] syntax for join methods

brandonkelly:
I just realized that query `join` methods support the same `['alias' => 'table_name']` syntax that `from()` supports, but the documentation didn’t mention that.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |


samdark:
Thank you!

PR #17969, opened 2020-04-12
Update tutorial-yii-integration.md

ahmadasjad:
Detailed instruction to use both versions or gradually move from yii1 to yii2

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/docs/guide/tutorial-yii-integration.md
+++ b/docs/guide/tutorial-yii-integration.md
@@ -177,3 +177,5 @@ That's all! Now in any part of your code, you can use `Yii::$app` to access the 
 echo get_class(Yii::app()); // outputs 'CWebApplication'
 echo get_class(Yii::$app);  // outputs 'yii\web\Application'
 ```
+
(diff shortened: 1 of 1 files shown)


samdark:
There is a section about it: https://www.yiiframework.com/doc/guide/2.0/en/intro-upgrade-from-v1

It could be expanded. An external link isn't a good idea.

ahmadasjad:
> An external link isn't a good idea

I agree with your statement 

> It could be expanded.

This link is not only about the doc how to move, but it also has the library to run both versions simultaneously. I was able to develop this library with lots of effort, chich made may able to understand the entire architecture of both versions. This library will save lots of time for others.

samdark:
That's fine. How about posting info about it to https://yiifeed.com/ and https://www.yiiframework.com/wiki?

ahmadasjad:
Ok
I'll do the same way

PR #17970, opened 2020-04-12
#17941 fix conficting controller creation

djidji01:
This is the final proposal on the #13433 and #17941 fixing conflicting controller creation. now all test passes
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17941,


framework/base/Module.php
```diff
@@ -571,32 +571,43 @@ public function createController($route)
...
+            return $module->createController($moduleRoute);
         }
 
-        $controller = $this->createControllerByID($id);
-        if ($controller === null && $route !== '') {
-            $controller = $this->createControllerByID($id . '/' . $route);
-            $route = '';
+        $controller = $this->createControllerByID($id . '/' . $route);
+        if ($controller !== null ) {
+            $id=$id.'/'.$route;
```
samdark, line 607: New id isn't used anywhere.

framework/base/Module.php
```diff
@@ -571,32 +571,43 @@ public function createController($route)
...
         }
 
-        if (strpos($route, '/') !== false) {
-            list($id, $route) = explode('/', $route, 2);
-        } else {
-            $id = $route;
-            $route = '';
+        // module and controller map take precedence
+        $routeInfo=pathinfo($route);
+        $id =$routeInfo['dirname']!=='.'?$routeInfo['dirname']:$routeInfo['basename'];
```
samdark, line 576: Why is there a dot in a route?

framework/base/Module.php
```diff
@@ -571,32 +571,43 @@ public function createController($route)
             return false;
         }
 
-        if (strpos($route, '/') !== false) {
-            list($id, $route) = explode('/', $route, 2);
-        } else {
-            $id = $route;
-            $route = '';
+        // module and controller map take precedence
+        $routeInfo=pathinfo($route);
```
samdark, line 575: Route here is `moduleId/controllerId/actionId`.

- moduleId and actionId are `[a-z-]+`.
- controllerId is `([a-z0-9_/]+/)?[a-z-]+`.

So for `moduleId/prefix/controllerId/actionId` it gives 

```
 'dirname' =>
  string(28) "moduleId/prefix/controllerId"
  'basename' =>
  string(8) "actionId"
  'filename' =>
  string(8) "actionId"
```

framework/base/Module.php
```diff
@@ -571,32 +571,43 @@ public function createController($route)
...
+        $id =$routeInfo['dirname']!=='.'?$routeInfo['dirname']:$routeInfo['basename'];
+        $route = $routeInfo['dirname']!=='.'?$routeInfo['basename']:'';
+        if (isset($this->controllerMap[$id.'/'.$route])) {
+            $controllerClass= $this->controllerMap[$id.'/'.$route];
+            $id=$id.'/'.$route;
```
samdark, line 580: Doesn't `$route` already contain ID?

samdark (changes requested):
It's not clear what the code does currently. Is it possible to create a unit test that fails with master but passes with this code?

framework/base/Module.php
djidji01: Yes, absolutely.

framework/base/Module.php
djidji01: Because if `$route` is an empty string `""`  the `pathinfo($route)['dirname']` returns `"."`

framework/base/Module.php
djidji01: Yeah sure it was not necessary necessary to initialize it. but at that step the controller ID is that` $id.'/'.$route` string and `actionId` ($route) is an empty string

framework/base/Module.php
djidji01: No. Though the parameter `$route` has always been  given as `moduleId/controllerId/actionId`, the returned  variable `$route` in `[$controller, $route]` has been always  the `actionId`. I am sorry it was also challenging for me with these variable naming.

In fact this is where the issue #13433 is solved. 

Thanks

djidji01:
> It's not clear what the code does currently. Is it possible to create a unit test that fails with master but passes with this code?

In fact this is where the issue #13433 is solved. The code was failing because misspelled  `enh ` I think.

Thanks

samdark:
@djidji01 yes, I know which issue is that but I'd like to have a unit tests to:

1. Prove it needs to be fixed.
2. Ensure it won't be broken in the future. Especially considering that the code isn't straightforward.

framework/base/Module.php
samdark: That certainly looks very magical. Can we solve it without `pathinfo()`?

djidji01: Absolutely yes.
```
 if (($pos = strrpos($route, '/')) !== false) {
            $id = substr($route, 0, $pos);
            $route = substr($route, $pos + 1);
        }else{
            $id = $route;
            $route = '';
}
```
give completely same result as: 

```
$routeInfo=pathinfo($route);
$id =$routeInfo['dirname']!=='.'?$routeInfo['dirname']:$routeInfo['basename'];
$route = $routeInfo['dirname']!=='.'?$routeInfo['basename']:'';
```
but I thought calling only one function `pathinfo()`  might perform better than calling `strrpos()` and `substr()` twice!!?

terabytesoftw:
Closing this PR because the source branch/repository is no longer available, the requested tests were not added, and the current patch changes controller resolution precedence beyond the proposed feature. The underlying feature can still be discussed in #17941, but it should be addressed with a fresh PR that includes regression tests and an explicit routing precedence design.

Issue #17975, opened 2020-04-16
typo in yii2/i18n/migrations/schema-mysql.sql


auerswald:
unused translation-sources cannot be deleted via the console, if the message tables are created manually with the file `yii2/i18n/migrations/schema-mysql.sql`.

see https://www.yiiframework.com/doc/api/2.0/yii-i18n-dbmessagesource

> If you don't want to use migration and need SQL instead, files for all databases are in migrations directory.

### What steps will reproduce the problem?
`./yii message message.php -r`

### What is the expected result?
no errors 🙃

### What do you get instead?

_"Cannot delete or update a parent row: a foreign key constraint fails (\`message\`, CONSTRAINT \`fk_message_source_message\` FOREIGN KEY (\`id\`) REFERENCES \`source_message\` (\`id\`) ON UPDATE CASCADE)"_

#### wrong: ####
ALTER TABLE \`message\` ADD CONSTRAINT \`fk_message_source_message\` FOREIGN KEY (\`id\`) REFERENCES \`source_message\` (\`id\`) **ON UPDATE CASCADE ON DELETE RESTRICT**;

#### fix: ####
ALTER TABLE \`message\` ADD CONSTRAINT \`fk_message_source_message\` FOREIGN KEY (\`id\`) REFERENCES \`source_message\` (\`id\`) **ON UPDATE RESTRICT ON DELETE CASCADE**;

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.x
| PHP version      | 7.4
| Operating system | *nix

samdark:
@auerswald do you have some time for a pull request?

PR #17976, opened 2020-04-16
Update schema-mysql.sql

auerswald:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| Fixed issues  | #17975


samdark (approved):
Please add a line for changelog. Thanks.

auerswald:
how and where? i am not so familiar with github 😎

samdark:
https://github.com/yiisoft/yii2/blob/master/framework/CHANGELOG.md this one. Just edit/push it to the same branch.

rob006:
What about migration and schemes for other DBMS?
Why `ON UPDATE RESTRICT`? Shouldn't we use `CASCADE` everywhere?

auerswald:
> What about migration and schemes for other DBMS?

these files are partly also wrong. I have only tested it with mysql.
but that should be taken care of by someone who knows more about it.

> Why `ON UPDATE RESTRICT`? Shouldn't we use `CASCADE` everywhere?

"ON UPDATE" and "ON DELETE" refers to the _id_ of the table _message_source_. it is not intended that this _id_ is changed in the program. but if a record in the table _message_source_ is to be deleted, all entries in the translation table _message_ should also be **optionally** deleted.

this happens if the property _'removeUnused'_ in the message-config is _true_ or the console parameter _'-r'_ or _'--remove-unused'_ is used.

btw... **the migration script creates the tables correctly.** the error occurs only if you have created the messages tables manually using the sql files (this is mentioned in the api documentation).

cebe (approved):

cebe:
> What about migration and schemes for other DBMS?

made the change here: https://github.com/yiisoft/yii2/pull/17978

samdark:
Merged in https://github.com/yiisoft/yii2/commit/d07cbdaab3e467d1a8875249d17db04af8e79864. Thanks!

PR #17980, opened 2020-04-17
Add method to load default values from default validator rule

kr0lik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/framework/base/Model.php
+++ b/framework/base/Model.php
@@ -14,6 +14,7 @@ use IteratorAggregate;
 use ReflectionClass;
 use Yii;
 use yii\helpers\Inflector;
+use yii\validators\DefaultValueValidator;
 use yii\validators\RequiredValidator;
 use yii\validators\Validator;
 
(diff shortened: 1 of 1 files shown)


kr0lik:
In this function you can set default values to attributes in rules. It can be used, for example, to fill out forms. This function is more convenient since you can specify values for some fields depending on the scenario.

rob006:
You could override `loadDefaultValues()` for that - it seems to be more intuitive than using validation rules for filling object with default data.

kr0lik:
Why validation rules for filling object with default values less intuitive then `loadDefaultValues`, sets only default values from database schema. In addition, it sets values such as current_timestamp or now(), instead of a date, for all cases, even when it is not needed. This is more useful for forms, its not need overwrite or use tricks for attributes

rob006:
`loadDefaultValues()` loads default values. If you don't like how it works, you can always overwrite it for your needs. 

Validation in Yii 2 already does too much. It not only validates data, but also mutates validated data. Adding another responsibility to validation rules is a step in wrong direction. It may work fine in your project, but IMO this is not a pattern that should be promoted in core framework.

kr0lik:
I do not propose to overload the validation rules, you already did it, I just suggest using what exist is more powerful. If there is functionality for creating default rules, why not load them into a clean form? You have defaultValidation. But now for correct use it you need add default values in form or in model by hands for any attributes, instead use one function. You can forgott somethig and in form on submit will be clear string and this validation will not work, but must. Yes i can overwrite `loadDefaultValues` but this is no correct. But It is will be trick. I can use trait with this function. But i think it must be a part of defaultValue functional.

kr0lik:
May be i somethgid dont understand. But 'loadDefaultValues' sets values that write for database validation. And in some cases this values incorrect for use in models. If you write DefaultValidationRules you need to use it. this function give a chance to use it quick. Yii is great framework for quick assembly some projects. Let stay it.

samdark:
@kr0lik first of all, thanks for proposal. That wasn't easy to decide on it, it is both hady for some cases and not really a good practice for majority of cases so the decision is not to merge it. Sorry.

Issue #17972, opened 2020-04-13
Prettified URL with 'enablePrettyUrl'=>false should result in error!!!


djidji01:
### What steps will reproduce the problem?
request to www.myapp.edu/courses/biology url resolve in a response despite that `'enablePrettyUrl'=>false`!!!

### What is the expected result?

Should get an error!

### What do you get instead?

result of site/index route

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.35-dev?
| PHP version      | 7.4
| Operating system | windows 10

Thoulah:
Confirmed, shows the default website instead of 404.

rob006:
You should probably remove rewrite rules from `.htaccess`.

djidji01:
My proposal would be to add a check on pathInfo at this level:

https://github.com/yiisoft/yii2/blob/a66fac3bb1700c612f1f0b80413051c82b506398/framework/web/UrlManager.php#L363
```
        try {
            if ($request->getPathInfo()!=='') {
                return false;
            }
        } catch (\Throwable $th) {
            return false;
        }catch (\Exception $e)
        {
            return false;
        }
```

samdark:
I've read the conversation in https://github.com/yiisoft/yii2/pull/17983 and it seems that despite your proposal to be logically fine, there are cases when it may break existing applications so I prefer not to solve it in 2.0 releases.

Thanks for discussion and a pull request though and sorry these were not merged.

PR #17983, opened 2020-04-19
Pretty formated url route to show 404 when 'enablePrettyUrl'=>false

djidji01:
#17972: Pretty formatted url route to show 404 when 'enablePrettyUrl'=>false, instead of showing home page.
| Q             | A
| ------------- | ---
| Is bugfix?    | 
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17972

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -9,6 +9,7 @@ Yii Framework 2 Change Log
 - Bug #17942: Fix for `DbCache` loop in MySQL `QueryBuilder` (alex-code)
 - Bug #17960: Fix unsigned primary key type mapping for SQLite (bizley)
 - Enh #17758: `Query::withQuery()` can be used for CTE (sartor)
+- Enh #17972: Pretty formated url route to show 404 when 'enablePrettyUrl'=>false (Djibril)
 
 
 2.0.34 March 26, 2020
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -360,7 +360,16 @@ class UrlManager extends Component
 
             return [$pathInfo, []];
         }
-
+        try {
+            if ($request->getPathInfo()!=='') {
+                return false;
+            }
+        } catch (\Throwable $th) {
+            return false;
+        }catch (\Exception $e)
+        {
+            return false;
+        }
         Yii::debug('Pretty URL not enabled. Using default URL parsing logic.', __METHOD__);
         $route = $request->getQueryParam($this->routeParam, '');
         if (is_array($route)) {
--- a/framework/web/UrlManager.php
+++ b/framework/web/UrlManager.php
@@ -365,10 +365,10 @@ class UrlManager extends Component
                 return false;
             }
         } catch (\Throwable $th) {
-            return false;
+            
         }catch (\Exception $e)
         {
-            return false;
+            
         }
         Yii::debug('Pretty URL not enabled. Using default URL parsing logic.', __METHOD__);
         $route = $request->getQueryParam($this->routeParam, '');
--- a/tests/framework/web/UrlManagerTest.php
+++ b/tests/framework/web/UrlManagerTest.php
@@ -253,6 +253,13 @@ class UrlManagerTest extends TestCase
         $result = $manager->parseRequest($request);
         $this->assertEquals(['site/index', []], $result);
         $this->assertEquals(5, $request->getQueryParam('id'));
+
+        // With path info not '';
+        $request->setPathInfo('pathInfo');
+        $request->setQueryParams([$routeParam => 'site/index', 'id' => 5]);
+        $result = $manager->parseRequest($request);
+        $this->assertFalse($result);
+        $this->assertEquals(5, $request->getQueryParam('id'));
     }
 
     public function testSetBaseUrl()


djidji01:
Hi @samdark  
should I delete this branch? I supposed it was merged but not? was it rejected?!

djidji01:
Or should I fix the codeclimate test fail?!!

samdark:
It was not rejected. I did not have a chance to check it yet.

xepozz:
Would be great to cover yours changes with test. Can you do it?

samdark:
The idea itself is fine but I wonder if anyone may rely on this working as it is. @yiisoft/reviewers any case when this may be true?

rob006:
I'm pretty sure it will break for cases where someone uses "non-pretty" URLs by default, with few exceptions configured by mod rewrite (like `RewriteRule ^/news/$ index.php?r=default/news`).

djidji01:
> 
> 
> I'm pretty sure it will break for cases where someone uses "non-pretty" URLs by default, with few exceptions configured by mod rewrite (like `RewriteRule ^/news/$ index.php?r=default/news`).

@rob006 I don't see any break yet I am using "non-pretty" without any mod rewrite configured? My I have a bit of detail?

djidji01:
> 
> 
> Would be great to cover yours changes with test. Can you do it?



> 
> 
> Would be great to cover yours changes with test. Can you do it?

Test added @xepozz. thanks

rob006:
Now, if you have `enablePrettyUrl` set to `false`, and enabled mod rewrite with rule `RewriteRule ^/news/$ index.php?r=default/news`, URL like `https://example.com/news/` will work fine. After your patch it will return 404, since `$request->getPathInfo()` is not empty in this case.

djidji01:
In fact URL like `https://example.com/news/ ` has to return 404 when `enablePrettyUrl `set to `false`! hasn't is? @xepozz

djidji01:
This because, if `enablePrettyUrl `set to `false ` whatever yo add after entry script `index.php` is only misleading and has no effect on request handling. Hence URL like `https://example.com/whatever/whatever/` return the same as `https://example.com/news` and` https://example.com/whatever/whatever?r=default/news`!!!!!

rob006:
> In fact URL like `https://example.com/news/ ` has to return 404 when `enablePrettyUrl `set to `false`! hasn't is?

If you have `RewriteRule ^/news/$ index.php?r=default/news` rewrite rule, then `https://example.com/news/` should be treated as `https://example.com/index.php?r=default/news`. 

> This because, if `enablePrettyUrl `set to `false ` whatever yo add after entry script `index.php` is only misleading and has no effect on request handling. Hence URL like `https://example.com/whatever/whatever/` return the same as `https://example.com/news` and` https://example.com/whatever/whatever?r=default/news`!!!!!

This is true only if you redirect all requests for non-existent files to `index.php`. I'm not sure why you're doing it, if you don't want to use pretty URLs.

___

Some kind of strict parsing could be interesting to handle cases like #17972, but:
1. This is BC break, and it probably should be hidden behind some switch (which needs to be disabled by default).
2. There are other cases where `UrlManager` does not use strict validation for parsing URLs. For example even if you set `showScriptName` to `false`, URLs like `https://example.com/index.php/news/` will still work fine.

djidji01:
> This is true only if you redirect all requests for non-existent files to index.php. I'm not sure why you're doing it, if you don't want to use pretty URLs 

Ahh sorry I can see. I was not seeing that the rule is redirecting `^/news/$`!!!! 
Otherwise, I should say the same!! Why one should make that rewrite rule if he don't want to use pretty URLs and why Yii should allow that yet it expect the URL to be not pretty

> Some kind of strict parsing could be interesting to handle cases like #17972, but:

Shouldn't this be at Yii level instead of server config??!!
> 
>     1. This is BC break, and it probably should be hidden behind some switch (which needs to be disabled by default).
Sure, Thanks
> 
>     2. There are other cases where `UrlManager` does not use strict validation for parsing URLs. For example even if you set `showScriptName` to `false`, URLs like `https://example.com/index.php/news/` will still work fine.
This should be another issue as well, because it is not a correct result either. set `showScriptName` to `false` should enforce to hide the entry script in any case!!

Thanks

rob006:
> Otherwise, I should say the same!! Why one should make that rewrite rule if he don't want to use pretty URLs and why Yii should allow that yet it expect the URL to be not pretty

Because I want only this one URL be handled in this way. This is pretty standard procedure if you want to migrate some existing service to Yii step by step. I can start with replacing only `/news/` by new Yii-based app (by redirecting traffic from this URL using rewrite rule) and "steal" URLs one by one until the whole app is migrated and all "exception rules" can be replaced by standard `RewriteRule . index.php [L]` rule.

djidji01:
In this case, The Yii-based app to which the migration is to should have enablePrettyUrl set to true as well. Otherwise I wonder what my be the result of standard `RewriteRule . index.php [L]` rule after complete migration!!!

rob006:
> In this case, The Yii-based app to which the migration is to should have enablePrettyUrl set to true as well.

Probably, but this is not required in this case (only a few URLs handled by rewrite rules), and `enablePrettyUrl` is disabled by default, why should anyone bother to change it, if it is not needed?

djidji01:
> why should anyone bother to change it, if it is not needed?

Everyone should bother to change it because it is inconsistent to the correct logic reasoning. This should be needed indeed and required in that case!!!

> there are probably better ways to handle such scenario

Sure. Always there is a better way to handle any current scenario. Probably one will come with it or maybe you can propose one please? Or my be consider 

> valid cases where this PR at current state will break BC

I would say that this PR was to try and add consistency to the framework logic and any case that may be broken is completely external to Yii it self, a hack.

rob006:
> Everyone should bother to change it because it is inconsistent to the correct logic reasoning. This should be needed indeed and required in that case!!!

I could say the same about your case from #17972. If you don't want to use pretty URLs, then don't enable mod rewrite and don't redirect all requests to Yii. Then `www.myapp.edu/courses/biology ` will be handled as 404 at webserver level.

> Everyone should bother to change it because it is inconsistent to the correct logic reasoning.

That depends what your "logic" is. For me logical assumption is that I don't need to enable features I don't use, and with `enablePrettyUrl` disabled, `UrlManager` should not care about path info, only check what is in `$_GET['r']`. So current behavior is correct and consistent, since URL manager does not care that much about duplicated URLs while parsing:
1. Treats `https://example.com/index.php/news` and `https://example.com/news` the same regardless of `showScriptName` setting.
2. Supports multiple URLs to default actions: `https://example.com/news/index` and `https://example.com/news` works the same.
3. Obviously it does not remove redundant and unnecessary query parameters (it is impossible to guess which params are redundant), so `https://example.com/news` and `https://example.com/news?foo=bar` will work the same too.

Why should I suspect that for your case there should be an exception and UrlManager should care about these duplicated URLs? Especially that usually `enablePrettyUrl` is disabled when someone does not care how URLs look like.

djidji01:
> If you don't want to use pretty URLs, then don't enable mod rewrite and don't redirect all requests to Yii

URL formatting and switch between format  at Yii level has never been dependent on mod-rewrite enabled or not I think. Also mod-rewrite is not only formatting URLs. Finally the purpose of redirecting  all request to entryScript is not for URL formatting.

> Especially that usually enablePrettyUrl is disabled when someone does not care how URLs look like.

I get the point: If one do not want to use prettyURL in Yii way, disable it and format URL as He want  Yii will not care !!!  My idea is that Yii should care about this #17972 !!!

Thanks.

PR #17992, opened 2020-04-23
Improve `yii\db\Connection` master and slave failover

### LINKED ISSUE #17991: Improve MySQL/MariaDB master load balancing on error condition

I think this has been brought up in #11306 already but it was not described clearly.

Setup:
- MySQL or MariaDB cluster with multiple masters
- or a single master with one or more slaves setup

Problem:
The connection may take too long to recover after a failure.

Current behavior of Yii is to cache the status of unavailable master servers. In case all masters are unavailable this is cached too and no attept is made to try connecting.

It will take `serverRetryInterval` seconds (which is default to 10min) to get a connection back.

During that time the exception message `None of the master DB servers is available.` is shown without any attempt to connect.

When no master is available, Yii should ignore unavailable server cache and try to connect anyway, ideally updating the cache status marking a server as available again after successful connection.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.x
| PHP version      | all
| Operating system | all

cebe:
no connection attempt was made when all servers are marked as
unavailable.

if server status cache is enabled and no server is available,
we now ignore the cache and try to connect anyway.

This helps to recover from short downtime much quicker.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | fixes #17991


samdark (approved):

samdark:
Merged. Thanks!

Issue #17999, opened 2020-04-27
Missing case in yiiunit\framework\base\SecurityTest::testGenerateRandomKey


DeryabinSergey:
### What steps will reproduce the problem?

On PHP version 7.1 and hi missing test case with data set 4 (array(false, true, true)) skipping, but need to check

### What is the expected result?

In `yii\base\Security` 

```
        if ($this->_useLibreSSL === null) {
            $this->_useLibreSSL = defined('OPENSSL_VERSION_TEXT')
                && preg_match('{^LibreSSL (\d\d?)\.(\d\d?)\.(\d\d?)$}', OPENSSL_VERSION_TEXT, $matches)
                && (10000 * $matches[1]) + (100 * $matches[2]) + $matches[3] >= 20105;
        }
```

make checking LibreSSL in public static method.

In tests `yiiunit\framework\base\SecurityTest::testGenerateRandomKey with data set 4 (array(false, true, true))` make a new case to checking LibreSSL and make test if ok and skip when using old library

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.34
| PHP version      | 7.4.5
| MySQL version      | 8.0.19
| Operating system | debian 10

samdark:
Alright. Since you're into it, how about a pull request?

DeryabinSergey:
Almost one in In `yii\base\Security` checking windows by

```
        if (function_exists('openssl_random_pseudo_bytes')
            && ($this->_useLibreSSL
            || (
                DIRECTORY_SEPARATOR !== '/'
                && substr_compare(PHP_OS, 'win', 0, 3, true) === 0
            ))
        ) {
```
and

```
        // If not on Windows, try to open a random device.
        if ($this->_randomFile === null && DIRECTORY_SEPARATOR === '/') {
```

Maybe do static method?

```
    public static function isWindows()
    {
        return 
            DIRECTORY_SEPARATOR !== '/'
            && substr_compare(PHP_OS, 'win', 0, 3, true) === 0;
    }
```

DeryabinSergey:
@samdark tell me, for what need class [ExposedSecurity.php](https://github.com/yiisoft/yii2/blob/master/tests/framework/base/ExposedSecurity.php) used only in [SecurityTest](https://github.com/yiisoft/yii2/blob/master/tests/framework/base/SecurityTest.php#L87)

samdark:
@DeryabinSergey it is for test only.

PR #18003, opened 2020-04-28
17999 missing test case in security test

DeryabinSergey:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17999


framework/base/Security.php
```diff
@@ -468,22 +499,10 @@ public function generateRandomKey($length = 32)
...
-
         // Since 5.4.0, openssl_random_pseudo_bytes() reads from CryptGenRandom on Windows instead
         // of using OpenSSL library. LibreSSL is OK everywhere but don't use OpenSSL on non-Windows.
         if (function_exists('openssl_random_pseudo_bytes')
-            && ($this->_useLibreSSL
-            || (
-                DIRECTORY_SEPARATOR !== '/'
-                && substr_compare(PHP_OS, 'win', 0, 3, true) === 0
-            ))
+            && ($this->_useLibreSSL || $this->isWindows())
```
samdark, line 505: That would pass tests because `getUseLibreSSL()` is called but will fail in real usage since property won't be initialized.

framework/base/Security.php
```diff
@@ -90,7 +90,39 @@ class Security extends Component
...
+        }
 
+        return $this->_useLibreSSL;
+    }
+
+    /**
+     * Check server operation system - windows or not
+     * @return boolean true if used windows on server
+     */
+    public static function isWindows()
```
samdark, line 120: I don't think it is a good idea to make it part of the public API.

framework/base/Security.php
```diff
@@ -90,7 +90,39 @@ class Security extends Component
...
+     * @var boolean The recent (> 2.1.5) LibreSSL RNGs are faster and likely better than /dev/urandom.
+     */
+    private $_useLibreSSL;
+
+
+    /**
+     * Check that openssl used LibreSSL  version >= 2.1.5
+     * @return boolean that used LibreSSL version 2.1.5 or hi
+     */
+    public function getUseLibreSSL()
```
samdark, line 103: I don't think it is a good idea to make it part of the public API.

samdark (changes requested):

framework/base/Security.php
DeryabinSergey: What should i Use in project in this case for coding style?

`$this->useLibreSSL` or `$this->getUseLibreSSL()`

framework/base/Security.php
DeryabinSergey: ExposedSecurity.php is a good idea for this case? Make method protected in Security and public in ExposedSecurity.php

DeryabinSergey:
Make new PR?

samdark:
You can push into the same branch. This PR will be updated automatically.

framework/base/Security.php
samdark: Yes.

framework/base/Security.php
samdark: Why would you need that in the project?

framework/base/Security.php
DeryabinSergey: No, only this. Which call do you prefer

samdark: ?

PR #18005, opened 2020-04-28
Fixes #18000 PK value of AR is missing after save

### LINKED ISSUE #18000: Oracle ActiveRecord primary keys get null after Save()
Create a model 
```php
$myModel= new ModelClass();
$myModel->KEY_COLUMN_NAME = 'foo';
$myModel->save();

// Get the key after insert
$pk = $myModel->KEY_COLUMN_NAME; //should be foo

```

The key attribute of `$myModel` should be `foo`

### What do you get instead?
It returns `null`

### Additional info

I found that the function ` insert($table, $columns)` in `db/oci/Schema.php`:
The `returnParams` is passed by reference when `bindParam` in `foreach` loop
After executing the statement, the `value` in `$returnParams[$phName]` is still null

I tried manually create a variable and bind it to the statement, it works fine.
```php
$command->pdoStatement->bindParam(':qp6',  $myVar, $value['dataType'], $value['size']);
if (!$command->execute()) {
            return false;
        }
// $myVar is updated and is the Pk value

```
Form the db log, the oracle sql is using returning into the param
```sql
INSERT INTO "SCHEMA"."TABLE" ("KEY_COLUMN_NAME", "A_COLUMN") VALUES ('foo',  'value') RETURNING "KEY_COLUMN_NAME" INTO :qp6
```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.34
| PHP version      | 7.4.3
| Operating system | Win 10

mankwok:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | #18000

I wrote a test case `testMultiplePrimaryKeyAfterSave` for testing the fix.

Howevever, when I clone the yii repo and run `composer install`
It throws exception for applying patches for phpunit:
`error: cannot apply patch fix PHP 7 compalibility (./tests/phpunit_mock_objects.patch)!`

Anyone can help?


samdark (approved):

samdark:
Thank you!

PR #18039, opened 2020-05-15
Minor refactoring

SOHELAHMED7:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | -
| Fixed issues  | -

~~As `Exception` class implements `Throwable` interface catching exceptions of both is not necessary~~

Initially I intended to remove catching of both `\Exception` and `\Throwable` but after I came to know \Throwble is not supported in PHP 5, I removed original refactoring and added a comment. So code is not changed

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "57627068914aee26c8205110ee14f928",
+    "content-hash": "957f707aecdad25f76b97a214ac67e54",
     "packages": [
         {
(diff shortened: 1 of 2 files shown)


rob006:
> As `Exception` class implements `Throwable` interface catching exceptions of both is not necessary

It is necessary on PHP 5, where `Throwable` is not available.

SOHELAHMED7:
Modifying my PR

SOHELAHMED7:
Fixed by 5280acb40

SOHELAHMED7:
@rob006 I have edited PR description

samdark:
@SOHELAHMED7 thanks for trying to improve the code but I don't think adding the comment makes it better.

PR #18049, opened 2020-05-19
Remove yii\web\Request::getScriptFile() & setScriptFile()

brandonkelly:
I noticed that `yii\web\Request::getScriptFile()` and `setScriptFile()` were doing the exact same thing as `yii\base\Request`, just in slightly different syntaxes.

The one notable exception is that `yii\base\Request::setScriptFile()` is parsing the passed-in path for aliases, and validating that the path is a file – which seem like things that `yii\web\Request` would benefit from.

I do wonder if it would be better for `getScriptFile()` to set `$this->_scriptFile` directly though, rather than going through `setScriptFile()`, to avoid that extra overhead, since `$_SERVER['SCRIPT_FILENAME']` can probably be trusted.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | n/a


brandonkelly:
Ok this is actually revealing some bugs with the tests, due to the added is_file() verification that `yii\web\Request::setScriptPath()` now inherits from `yii\base\Request`.

The following `scriptFile` definitions are actually invalid:

https://github.com/yiisoft/yii2/blob/60a39654dd2d425a24c64d6c1d641a0754b015bc/tests/TestCase.php#L83

https://github.com/yiisoft/yii2/blob/60a39654dd2d425a24c64d6c1d641a0754b015bc/tests/framework/web/ControllerTest.php#L53

https://github.com/yiisoft/yii2/blob/60a39654dd2d425a24c64d6c1d641a0754b015bc/tests/framework/web/ControllerTest.php#L128

https://github.com/yiisoft/yii2/blob/60a39654dd2d425a24c64d6c1d641a0754b015bc/tests/framework/web/UserTest.php#L210

I guess this is why `yii\web\Request::setScriptFile()` doesn’t currently have any validation. So just removed the `is_file()` check from `yii\base\Request::setScriptFile()`. No sense on enforcing it for console requests but not web requests, right?

rob006:
`yii\base\Request::setScriptPath()` uses `realpath()` for path normalization. Wouldn't that change behavior in case of using symlinks for webroot?

brandonkelly:
@rob006 Still calling `realpath()` because path normalization seemed like a good thing – which `yii\web\Request::setScriptPath()` was **not** doing.

Since `realpath()` can return `false` if the file doesn’t exist, I’ve just added one more commit that will only use the `realpath()` result if it doesn’t return `false`. I think that is the best of both worlds:

- If the path is valid, normalize it with `realpath()`
- If the path is invalid, just save it as-is. (It’s not like normalization is going to help you here anyway.)

rob006:
> Still calling `realpath()` because path normalization seemed like a good thing – which `yii\web\Request::setScriptPath()` was **not** doing.

Whether this is a good thing is debatable, I would expect that `yii\web\Request::getScriptPath()` will return the same value as in `$_SERVER['SCRIPT_FILENAME']`. But this is definitely a BC break, and should be considered in this context too.

brandonkelly:
Alternatively we could just change `setScriptPath()` to set `$this->_scriptPath()` directly, as `yii\web\Request::setScriptPath()` has been doing all along, and have the BC break happen for console requests, which are the only things that currently get normalized.

Either way, I think they should behave consistently.

samdark:
While consistency matters, this pull request changes behavior thus it is a backwards compatibility break. Let's leave it as it was.

Issue #18051, opened 2020-05-20
UnknownMethodException DynamicModel - each custom validator


1Luc1:
After update the Yii2 version to **2.0.35** from 2.0.34 the _each_ validator doesn't work with _custom validation function._ I suspect #18012 to have cause this problem (or #17944?).

### What steps will reproduce the problem?
Set up custom validation function with each within a model class:
```
public $staff_ids;

public function rules()
{
    return [
        ...
        ['staff_ids', 'each', 'rule' => ['string']],
        [['staff_ids'], 'each', 'rule' => ['validateStaffIds']],
        ...
    ];
}

public function validateStaffIds($attribute, $params, $validator)
{
   ...
}
```

### What is the expected result?
For each id within the staff_ids the custom validater validateStaffIds is called.

### What do you get instead?
Error:
`yii\base\UnknownMethodException: Calling unknown method: yii\base\DynamicModel::validateStaffIds() in /srv/http/ig-net/vendor/yiisoft/yii2/base/Component.php:300`

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.35
| PHP version      | v7.4.5
| Operating system | Manjaro 5.4.39
| Webserver | Nginx 1.18

samdark:
@ricardomm85, @bizley would you please take a look?

samdark:
That should be #17944 when the model was replaced with dynamic model.

bizley:
I think I have a solution. Let me test it a bit before PR.

bizley:
Looks like it works, let me know what you think.

samdark:
@1Luc1 please check if it is OK with master version.

1Luc1:
With v2.0.34 i used the attribute within the costume validator as follow (continuing the example from above):

```
public function validateStaffIds($attribute, $params, $validator)
{
   ...
   if (is_numeric($this->$attribute)) {
      ...
   }
}
```
Where this custom validator function is called for each element within _staff_ids_.

With v2.0.36-dev it doesn't work like that anymore and i have to change it as follow to at least don't get any errors:
```
public function validateStaffIds($attribute, $params, $validator)
{
   ...
   if (is_numeric($this->$attribute[0])) {
      ...
   }
}
```
As i can see from debug logs, _validateStaffIds_ is only called once and `$this->$attribute` has the same values as _staff_ids_.

I don't know if this is indented to work like that? If so, i have to run through all values with a loop within the custom validator function.

[...]

PR #18052, opened 2020-05-20
Adding support for custom rule in EachValidator

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | should
| Fixed issues  | #18051 

This changes a bit how the InlineValidator is created in case of model-method-based inline validators. Instead of just passing a method's name now there is an array with an instance of Model in question and the method's name. This works thanks to the fact that `call_user_func` can take such array as first parameter so there should be no BC change.

Since EachValidator creates a new instance of DynamicModel internally this change allows inline validators defined in the original model to be used.

I haven't checked all possible drawbacks here. Definitely now method that is inline validator cannot have limited visibility but I haven't checked if it could be the case before this change.

Please take a look and see if this can be the solution to the issue. If so, I will add changelog.

### DIFF
--- a/framework/CHANGELOG.md
+++ b/framework/CHANGELOG.md
@@ -8,6 +8,7 @@ Yii Framework 2 Change Log
 - Enh #18019: Allow jQuery 3.5.0 to be installed (wouter90)
 - Bug #18026: Fix `ArrayHelper::getValue()` did not work with `ArrayAccess` objects (mikk150)
 - Enh #18048: Use `Instance::ensure()` to set `User::$accessChecker` (lav45)
+- Bug #18051: Fix missing support for custom validation method in EachValidator (bizley)
 
 
 2.0.35 May 02, 2020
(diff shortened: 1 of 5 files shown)


samdark:
> I haven't checked all possible drawbacks here. Definitely now method that is inline validator cannot have limited visibility but I haven't checked if it could be the case before this change.

I think it was allowed before.

bizley:
As I can see the validator is doing the same thing when `$type` is just a string - it uses `[$model, $type]` for `call_user_func` but then the $model is the original one. So I think limited visibility methods were no good previously as well.

SilverFire:
Hm, I also thought it always worked

SilverFire (approved):

samdark (approved):
👍 Please add a line for CHANGELOG.md.

bizley:
Added :heavy_check_mark:

samdark:
👍

PR #18062, opened 2020-05-24
Fixing validating array values separately

bizley:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | #18051


bizley:
This unfortunately requires now BC change for behavior of EachValidator with InlineValidator as a rule.

I have updated signature of inline validator with 4th parameter `current`:

```php
/**
 * @param mixed $current the currently validated value of attribute
 */
function ($attribute, $params, $validator, $current)
```

And now `$attribute` refers to original model's attribute and not its temporary counterpart:
  
```php
public $array_attribute = ['first', 'second'];

public function rules()
{
    return [
        ['array_attribute', 'each', 'rule' => ['customValidatingMethod']],
    ];
}
  
public function customValidatingMethod($attribute, $params, $validator, $current)
{
    // $attribute === 'array_attribute' (as before)
  
    // now: $this->$attribute === ['first', 'second'] (on every iteration)
    // previously:
    // $this->$attribute === 'first' (on first iteration)
    // $this->$attribute === 'second' (on second iteration)
  
    // use now $current instead
    // $current === 'first' (on first iteration)
    // $current === 'second' (on second iteration)
}
```

To be honest I don't like this. It's a lot just to allow typed property for array attribute. Maybe you can come up with better solution.

samdark:
@bizley any better ideas since the pull request?

bizley:
Unfortunately not.

samdark (approved):

samdark:
Thank you, @bizley!

Issue #17852, opened 2020-02-09
Error in console/Controller.php?


WeeSee:
In console/Controller.php we can read:

```php
/**
     * Prints a string to STDOUT.
     *
     * You may optionally format the string with ANSI codes by
     * passing additional parameters using the constants defined in [[\yii\helpers\Console]].
     *
     * Example:
     *
     * ```
     * $this->stdout('This will be red and underlined.', Console::FG_RED, Console::UNDERLINE);
     * ```
     *
     * @param string $string the string to print
     * @return int|bool Number of bytes printed or false on error
     */
    public function stdout($string)
```

The function arguments do not match the docs above. 

The [Yii2 tutorial](https://www.yiiframework.com/doc/guide/2.0/en/tutorial-console#formatting-and-colors) is also showing two arguments.

Is this an error or do I miss something?

simialbi:
As you can see some lines among, the function arguments are loaded via `func_get_args`:
```php
  /**
     * Prints a string to STDOUT.
     *
     * You may optionally format the string with ANSI codes by
     * passing additional parameters using the constants defined in [[\yii\helpers\Console]].
     *
     * Example:
     *
     * ```
     * $this->stdout('This will be red and underlined.', Console::FG_RED, Console::UNDERLINE);
     * ```
     *
     * @param string $string the string to print
     * @return int|bool Number of bytes printed or false on error
     */
    public function stdout($string)
    {
        if ($this->isColorEnabled()) {
            $args = func_get_args();
            array_shift($args);
```

So no, there is no error.

WeeSee:
Wow. I didn't see this. 

What's the reason not to supply additional parameters as in [PHP's ```json_encode `` function](https://www.php.net/manual/de/function.json-encode.php), e.g. as bit defined flags or a options array?

Using this way documentation may be misleading and function definitions do not give a full understanding of how to use the function.

simialbi:
It's more like [PHP's `array_merge` function](https://www.php.net/manual/de/function.array-merge). It can have 0-∞ additional styling parameters (after `$string` parameter). But you're right, the documentation might be more clear about this (maybe like [BaseArrayHelper](https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseArrayHelper.php#L114)).

PR #18076, opened 2020-05-31
resolves issue #17852

ahmadasjad:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17852

It was also possible to add `...`(three dot) inside the function param, but I skipped as it's available from **5.6** and we are supporting **5.4**


samdark:
Thanks!

PR #18081, opened 2020-06-04
Fix for PDO_DBLIB (MSSQL). 

darkdef:
Set flag ANSI_NULL_DFLT_ON to ON for current connect to DB.

This flag modifies the behavior of the session to override default nullability of new columns.
By default, for some versions of dblib + freetds, this flag is off.
ANSI recommends default behavior - NULL for new fields without default setted null/not null

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️/❌
| Fixed issues  | ❌


samdark (approved):

samdark:
Thanks!

Issue #18001, opened 2020-04-27
Can't get table metadata for tables with '(' in their names


trombipeti:
Disclaimer: I know it is a horrible idea to have table names like this, but I have to work with a DB that already contains some and I'd like for my code to work nonetheless.

### What steps will reproduce the problem?
Create a table in the database of the app with parenthesis in the table name, so like:
``CREATE TABLE `example_table(0)` (
 `ID` int(11) NOT NULL AUTO_INCREMENT,
 PRIMARY KEY (`ID`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8``
Try to run this line or any code that involves querying the tables schema.
`Yii::$app->db->getTableSchema("example_table(0)", true);`
### What is the expected result?
I expect to get the table metadata returned, as it is with any other tables without the parenthesis.

### What do you get instead?
`SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '(0)' at line 1
The SQL being executed was: SHOW FULL COLUMNS FROM example_table(0)`
As you can see, the table name isn't quoted correctly. Looking at the code makes me believe that this issue is present independent of the DB driver.
https://github.com/yiisoft/yii2/blob/dd657f8d64dc24b842f5b65a0e8748cd78bf6554/framework/db/Schema.php#L481-L483
Moreover, it doesn't help if I quote the table name myself, because the quoting characters are removed when building the query. I _think_ this only happens when the MySQL driver is used.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.34
| PHP version      | 7.4.5
| Operating system | Linux

samdark:
Yes, it could be considered a bug. A very rare one. Fixing it is not a priority but we'll accept a pull request with tests and a fix.

trombipeti:
Ok, thanks!
Any background on why it is like this? The '{{' part I get but I couldn't figure out why having a '(' in the table name should prevent it from being quoted.

samdark:
> Any background on why it is like this?

We haven't thought of such names.

> but I couldn't figure out why having a '(' in the table name should prevent it from being quoted

I think that's about SQL for create/alter table or column. `(` is used there to specify size of the column.

mbunyan:
[...]

PR #18088, opened 2020-06-07
Brackets checking removed from the return condition.

floor12:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  |❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | #18001

Let watch how the tests falls.


samdark:
Seems to fail tests. At least table quoting ones.

floor12:
I updated the code to check brackets in first and last position in table name string. When all tests passed I also added new test to check quoting table name with brackets.

samdark (approved):
Please add a line for CHANGELOG.md

floor12:
@samdark Sorry for my stupid question, but its my first contribute in a such big project. So, my question is about CHANGELOG. I made all this changes from the last stable release 2.0.35  (because master branch had test fails at the moment of forking). So, my version of CHANGELOG hasn't list of 2.0.36 changes. And im not sure that it will be merged properly if I just add new line. So, I should merge master into my branch, add the line with my changes, and make new pull request, o something else?
Thanks.

framework/db/Schema.php
```diff
@@ -478,7 +478,11 @@ public function quoteValue($str)
      */
     public function quoteTableName($name)
     {
-        if (strpos($name, '(') !== false || strpos($name, '{{') !== false) {
+
+        if (strpos($name, '(') === 0 && strpos($name, ')') === strlen($name) - 1) {
```
rob006, line 482:
```suggestion
        if (strncmp($name, '(', 1) === 0 && substr($name, -1) === ')') {
```

floor12: @rob006 I made a few speed tests and found that your variant works slowly for 5-15%. I was surprised but it's true.

samdark:
@floor12 I'll handle it, don't worry.

framework/db/Schema.php
rob006: Hmm, `strpos()` indeed could be more efficient in case of short strings.

samdark:
Merged. Thank you.

PR #18089, opened 2020-06-08
Fix ImageValidator's maxHeight doc

hamtiko:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | ❌

I found a mistake on `maxHeight` property doc in `ImageValidator` class. In the doc it is mentioned to `@see overWidth ...` but should be `@see overHeight ...`.


samdark:
Thanks!

PR #18097, opened 2020-06-11
Fix #18096: Fix InlineValidator with anonymous inline function not wo…

### LINKED ISSUE #18096: EachValidator & InlineValidator don't work together
https://github.com/yiisoft/yii2/issues/17810 introduced a new way of using EachValidator: it uses a new instance of DynamicModel to run the validations on. This breaks the InlineValidators that'd use $this->$attribute in an inline anonymous function.

Have a model with an array attribute.
Create a rule in your model with EachValidator, and as a rule set an inline function, causing an InlineValidator instance to be created. In this inline anonymous function, try checking if the members of the array are valid by accessing $this->$attribute.

$this should be the actual object on which the function was called, $this->$attribute should be one member of the array in each call.

### What do you get instead?
$this is bound to the original Model class, and $this->$attribute is the array attribute.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.35
| PHP version      | 7.4
| Operating system | Windows

trombipeti:
…rking well from EachValidator

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #18096

### DIFF
--- a/framework/validators/InlineValidator.php
+++ b/framework/validators/InlineValidator.php
@@ -68,7 +68,10 @@ class InlineValidator extends Validator
         $method = $this->method;
         if (is_string($method)) {
             $method = [$model, $method];
+        } else if ($method instanceof \Closure) {
+            $method = $this->method->bindTo($model);
         }
+
         call_user_func($method, $attribute, $this->params, $this);
     }
 
@@ -81,6 +84,8 @@ class InlineValidator extends Validator
             $method = $this->clientValidate;
             if (is_string($method)) {
                 $method = [$model, $method];
+            } else if ($method instanceof \Closure) {
+                $method = $this->method->bindTo($model);
             }
 
             return call_user_func($method, $attribute, $this->params, $this);
--- a/framework/validators/InlineValidator.php
+++ b/framework/validators/InlineValidator.php
@@ -68,7 +68,7 @@ class InlineValidator extends Validator
         $method = $this->method;
         if (is_string($method)) {
             $method = [$model, $method];
-        } else if ($method instanceof \Closure) {
+        } elseif ($method instanceof \Closure) {
             $method = $this->method->bindTo($model);
         }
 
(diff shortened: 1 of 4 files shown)


samdark:
Seems to break tests on pre PHP 7.4.

trombipeti:
Sorry, fixed.

framework/validators/InlineValidator.php
```diff
@@ -81,6 +84,8 @@ public function clientValidateAttribute($model, $attribute, $view)
             $method = $this->clientValidate;
             if (is_string($method)) {
                 $method = [$model, $method];
+            } else if ($method instanceof \Closure) {
```
bizley, line 87:
```suggestion
            } elseif ($method instanceof \Closure) {
```

tests/data/validators/models/ValidatorTestEachAndInlineMethodModel.php
```diff
@@ -0,0 +1,26 @@
...
+use yii\base\Model;
+
+class ValidatorTestEachAndInlineMethodModel extends Model
+{
+    public $arrayProperty = [true, false];
+
+    public function rules()
+    {
+        return [
+            ['arrayProperty', 'each', 'rule' => [function($attribute, $params, $validator) {
```
bizley, line 19:
```suggestion
            ['arrayProperty', 'each', 'rule' => [function ($attribute, $params, $validator) {
```

tests/data/validators/models/ValidatorTestEachAndInlineMethodModel.php
```diff
@@ -0,0 +1,26 @@
...
+                if(is_array($this->$attribute)) {
```
bizley, line 20:
```suggestion
                if (is_array($this->$attribute)) {
```

bizley (changes requested):
Some code style changes only. The rest looks good

samdark (approved):

samdark:
Thank you!

brandonkelly:
This change introduced a change in behavior that should be documented: #18175

PR #18100, opened 2020-06-12
Use inheritDoc to replace outdated docblock

ghost:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | https://github.com/yiisoft/yii2/pull/18098/files#r439294563

@samdark Perhaps {inheritDoc} should be used where applicable on all child classes of Session, but this is partly a design choice you'd have to make.

### DIFF
--- a/framework/web/DbSession.php
+++ b/framework/web/DbSession.php
@@ -95,9 +95,7 @@ class DbSession extends MultiFieldSession
     }
 
     /**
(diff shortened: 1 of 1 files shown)


samdark:
Thanks!

PR #18102, opened 2020-06-13
master/slave → primary/replica

brandonkelly:
There is a growing call for the tech industry to move away from “master”/“slave” terminology in the interest of inclusion, especially toward the Black community.

In that spirit, this PR deprecates those terms as they relate to read/write splitting, in favor of “primary” and “replica” – which I’d argue are not just more inclusive terms, but also more semantically correct. (The replica servers are not merely “taking orders” from their “master” server; they are replicating it.)

I’ve managed to do this in a non-breaking way, by keeping old master/slave methods around as aliases for the new ones, and adding magic getter/setter methods for the old master/slave properties. I’ve also introduced a couple traits to document the now-deprecated properties, based on https://stackoverflow.com/a/51890536/1688568.

Code that demonstrates how PhpStorm recognizes yii\db\Connection::$master as deprecated

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/docs/guide-es/db-dao.md
+++ b/docs/guide-es/db-dao.md
@@ -439,8 +439,8 @@ try {
 
 Muchos DBMS soportan [replicación de bases de datos](http://en.wikipedia.org/wiki/Replication_(computing)#Database_replication) para tener
 una mejor disponibilidad de la base de datos y un mejor tiempo de respuesta del servidor. Con la replicación de bases
-de datos, los datos están replicados en los llamados *servidores maestros* (master servers) y *servidores esclavos*
-(slave servers). Todas las escrituras y actualizaciones deben hacerse en el servidor maestro, mientras que las lecturas
+de datos, los datos están replicados en los llamados *servidores maestros* (primary servers) y *servidores esclavos*
+(replica servers). Todas las escrituras y actualizaciones deben hacerse en el servidor maestro, mientras que las lecturas
 se efectuarán en los servidores esclavos.
 
 Para aprovechar las ventajas de la replicación de la base de datos y lograr una división de lecuta-escritura, se puede configurar
@@ -451,13 +451,13 @@ el componente [[yii\db\Connection]] como se muestra a continuación:
     'class' => 'yii\db\Connection',
 
     // configuración para el maestro
-    'dsn' => 'dsn for master server',
-    'username' => 'master',
+    'dsn' => 'dsn for primary server',
+    'username' => 'primary',
     'password' => '',
 
     // configuración para los esclavos
-    'slaveConfig' => [
-        'username' => 'slave',
+    'replicaConfig' => [
+        'username' => 'replica',
         'password' => '',
         'attributes' => [
             // utiliza un tiempo de espera de conexión más pequeña
@@ -466,11 +466,11 @@ el componente [[yii\db\Connection]] como se muestra a continuación:
     ],
 
     // listado de configuraciones de esclavos
-    'slaves' => [
-        ['dsn' => 'dsn for slave server 1'],
-        ['dsn' => 'dsn for slave server 2'],
-        ['dsn' => 'dsn for slave server 3'],
-        ['dsn' => 'dsn for slave server 4'],
+    'replicas' => [
+        ['dsn' => 'dsn for replica server 1'],
+        ['dsn' => 'dsn for replica server 2'],
+        ['dsn' => 'dsn for replica server 3'],
+        ['dsn' => 'dsn for replica server 4'],
     ],
 ]
 ```
@@ -492,7 +492,7 @@ $db->createCommand("UPDATE user SET username='demo' WHERE id=1")->execute();
 
 > Info: Las consultas realizadas llamando a [[yii\db\Command::execute()]] se consideran consultas de escritura,
   mientras que todas las demás se ejecutan mediante alguno de los métodos "query" de [[yii\db\Command]] son consultas
-  de lectura. Se puede obtener la conexión de esclavo activa mediante `$db->slave`.
+  de lectura. Se puede obtener la conexión de esclavo activa mediante `$db->replica`.
 
 El componente `Connection` soporta el balanceo de carga y la conmutación de errores entre esclavos. Cuando se realiza
 una consulta de lectura por primera vez, el componente `Connection` elegirá un esclavo aleatorio e intentará realizar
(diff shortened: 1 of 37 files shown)


rob006:
Master-slave is common naming pattern in this case. You may consider new term more inclusive and correct, but i is also more confusing if you want to read and refer to any existing documentation. 
https://mariadb.com/resources/blog/database-master-slave-replication-in-the-cloud/
https://en.wikipedia.org/wiki/Master/slave_(technology)

Also, I don't think that "replica" is correct term here and it does not mean the same as "slave" in "master/slave" model.

brandonkelly:
I chose those terms because SQL Server and RDS use them:

- [SQL Server - Configure replication](https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/configure-replication-for-always-on-availability-groups-sql-server?view=sql-server-ver15):

  > …If set to 1 Windows authentication is used to connect to the current **primary**. If set to 0, SQL Server authentication is used with the specified `@login` and `@password` values. The login and password specified must be valid at each secondary **replica** for the validation stored procedure to successfully connect to that **replica**…

- [RDS - Working with Read Replicas](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_ReadRepl.html)

  > …Business reporting or data warehousing scenarios where you might want business reporting queries to run against a read **replica**, rather than your **primary**, production DB instance…

Also:

- [box - How We Learned to Stop Worrying and Read from Replicas](https://medium.com/box-tech-blog/how-we-learned-to-stop-worrying-and-read-from-replicas-58cc43973638)
- [Pantheon - Scaling WordPress with MySQL Replicas and HyperDB](https://pantheon.io/docs/hyperdb)

“Master”/“slave” may still be considered a convention, but it’s on its way out, and “replica” is the new conventional term.

brandonkelly:
Also, [from the very Wikipedia article you linked to](https://en.wikipedia.org/wiki/Master/slave_(technology)#Terminology_concerns):

> ## Terminology concerns
>
> The terminology has sometimes been replaced with something different because the terms "master" and "slave" refer to the practice of slavery.
>
> "Main" & "Secondary" has been proposed, enabling reusable interpretation of the remaining acronym labels for connecting.
>
> One alternative for databases is "primary" and "replica", which is used in the documentation from IBM, Microsoft, Engine Yard, Amazon Web Services/Amazon Relational Database Service, and ACM as well as in Python, Django, Drupal, CouchDB, Redis and MediaWiki (which still uses "master").

Thoulah:
Are you gonna rename whitelist / blacklist too?

onebrightlight:
> Are you gonna rename whitelist / blacklist too?

That sounds like a great suggestion too, tbh.

brandonkelly:
It crossed my mind but Yii doesn’t have explicit “whitelist”/“blacklist” concepts. The words only show up in a couple comments/docs. Can look into updating those in a separate PR, but this felt more pressing as a starting point.

samdark (approved):
Looks like backwards compatibility is fine.

brandonkelly:
Thanks @samdark!

lubosdz:
I'd never think before that `whitelist/blacklist` or `master/slave` in framework regard has anything to do with xenophoby. Now I do. Every developer knows those are common historical terms used around the IT world for years. This PR does not add any value, and IMHO is too rushy to reflex last 2 week's social events somewhere in America and is a BC break. Framework as a tech stuff should stay away from such a things. Or at least it should be discussed & agreed by a larger Yii community.

samdark:
It is not a BC break, only soft deprecation. I've checked it. If it actually causes any issues with existing apps, please report it.

I agree that the PR doesn't add value and may cause a minor inconvenience for people who do not have historical and cultural background similar to US and thus these people, including myself, do not find these terms personally offensive.

I agree that in ideal world technical terms should not offend anyone because the context is different but in reality it's not like that. And that's not only about last 2 weeks but about cultural differences overall.

If new terms would have been any worse than previous ones, I'd likely criticize such pull request but primary/replica terms are alright in terms of understanding and are nowadays widely used as @brandonkelly highlighted.

rob006:
@samdark This PR removes existing and documented properties - of course it breaks BC and it is against [documented BC promise](https://github.com/yiisoft/yii2/blob/master/docs/internals/bc.md#classes-1). It will break my code if:
1. I extend one of changed classes and override one of removed properties to change default value (for example set default value for `Connection::$slaveConfig`) - my property will be ignored after this change.
2. I extend any of changed classes and implement one of getter/setter methods for different purposes (for example `Connection::getSlaveConfig()`) - it will either change behavior or throw exception if method signatures does not match.

brandonkelly:
Almost every change could theoretically result in some unintended break. Doesn’t mean it’s a practical concern. And in this case the “break” would be that read queries start pointing to the primary DB server, until you notice that you’re overriding a now-deprecated property. So pretty low risk I’d say.

rob006:
> And in this case the “break” would be that read queries start pointing to the primary DB server, until you notice that you’re overriding a now-deprecated property.

That depends on the case, it could bring any effect, from nothing to fatal errors as result of inconsistent configuration. But "read queries start pointing to the primary DB server" is probably the worst of them, because it will be easy to miss it on testing/dev environment and could completely kill your production master server by flood of read queries (previously handled by swarm of read replicas). And finding the cause of this won't be so obvious.

> So pretty low risk I’d say.

I'd say that there is not much to gain by this change either, at least from technical perspective. Yii 2 is in maintenance mode, and this is cosmetic change with potential BC break implications. I don't believe that any similar PR (but not politically affected) would be merged, especially so quickly.

BTW: this renaming still could be done and keep BC at the same time by leaving old properties and adding only PC-compliant setters and getters to provide new interface to set primary/replica settings.

brandonkelly:
> BTW: this renaming still could be done and keep BC at the same time by leaving old properties and adding only PC-compliant setters and getters to provide new interface to set primary/replica settings.

That’s a great point. Would be a bit simpler to implement (I can remove those docs-only trait classes), and the docs can still point people to the new “properties”. I’ll push that up in a new PR.

brandonkelly:
…and done: #18106

brandonkelly:
This change ended up getting reverted due to backwards compatibility concerns (see #18106).

I’ve made similar changes to Craft CMS via our own DB connection class that extends `yii\db\Connection` (craftcms/cms@656aea8352b682b0eb408a42f1c3c507f51b9f52).

Issue #18090, opened 2020-06-09
GridView with several CheckboxColumns and multiselect doesn't work


hohnacker:
### What steps will reproduce the problem?

> echo GridView::widget(
        [
            'dataProvider' => $yourDataProvider,
            'columns'      => [
                [
                    'class'  => CheckboxColumn::class,
                    'name'   => 'checkboxone',
                    'header' => Html::checkBox(
                        'checkboxone_all',
                        false,
                        [
                            'class' => 'select-on-check-all',
                            'label' => 'labelone'
                        ]
                    ),
                ],
                [
                    'class'  => CheckboxColumn::class,
                    'name'   => 'checkboxtwo',
                    'header' => Html::checkBox(
                        'checkboxtwo_all',
                        false,
                        [
                            'class' => 'select-on-check-all',
                            'label' => 'labeltwo',
                        ]
                    ),
                ],
            ]
        ]
    ); ?>

### What is the expected result?

"Check all" functionality per Column Header

### What do you get instead?

If I click on the first header checkbox no checkbox is checked automatically.
The last columns work. It seems that always the last one wins.

![image](https://user-images.githubusercontent.com/33871826/84131174-83906d80-aa44-11ea-9e65-c223bf7cb436.png)

I think the issue is in https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.gridView.js #194
**setSelectionColumn** is called twice for each column.
**initEventHandler** is called for both columns, but both with the same type name
"**checkAllRows**".
If we change checkAllRows with the column spezific name it would work.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.35
| PHP version    |  7.3 
| Windows  | 10

floor12:
I will check it today.

floor12:
Yes, this problem is confirmed by my experiments.
The problem exists in `yii.gridView.js` file in `initEventHandler` function. This function and and structure `gridEventHandlers` object designed to store only one type of event for one grid view. Theoretically this can create other problems besides this problem with checkboxes. I already started work to fix this issue.

samdark:
See https://github.com/yiisoft/yii2/pull/18103#issuecomment-645476633

PR #18103, opened 2020-06-14
The gridEventHandlers object keys generation updated.

floor12:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | #18090

I checked the problem described in issue #18090 and found that the problem was in rewriting element in `gridEventHandlers` in `yii.gridView.js` file in `initEventHandler` function. 
Looks like the CheckboxColumn was designed to use only once in one GridView Widget. 
After a few different tries to fix this issue and not fail a lot of test, I found this solution, with fails inly one JS test. 
Now all seems to be work good, so I think we need to update *setSelectionColumn method with repeated calls* JS test (line 675 in `yii.gridView.test.js`), because it was designed to work only with one checkbox column. I'm not a big specialist in JS test, so some help needed.

### DIFF
--- a/framework/assets/yii.gridView.js
+++ b/framework/assets/yii.gridView.js
@@ -14,7 +14,7 @@
         if (methods[method]) {
             return methods[method].apply(this, Array.prototype.slice.call(arguments, 1));
         } else if (typeof method === 'object' || !method) {
-                return methods.init.apply(this, arguments);
+            return methods.init.apply(this, arguments);
         } else {
             $.error('Method ' + method + ' does not exist in jQuery.yiiGridView');
             return false;
@@ -253,6 +253,7 @@
             gridEventHandlers[id] = {};
         }
         $(document).on(event, selector, callback);
-        gridEventHandlers[id][type] = {event: event, selector: selector};
+        var typeKey = (type === 'filter') ? type : type + selector;
+        gridEventHandlers[id][typeKey] = {event: event, selector: selector};
     }
 })(window.jQuery);


samdark:
@floor12 Seems it breaks another case. See tests: https://github.com/yiisoft/yii2/pull/18103/checks?check_run_id=770426107#step:8:1385

floor12:
Yes, this is I was talking about. This test, like the checkboxField,  was designed to use only once per gridView widget. I think we need to update this test and add one another. Im not a big specialist in js testing. It would be great if someone helped in this matter. If there is no specialist in JS-Testing, I will have to dive into this question.

samdark:
Yes. Test should be updated and no, there's no specialist about it :(

floor12:
Ok. I will take a look to these tests with more attention.

floor12:
Having studied the problem again, as well as tests and documentation on this issue, I came to the conclusion that this described behaviour is normal. 
Most likely, we need to edit the documentation, adding information that the grid is designed to use only one column with checkboxes using `checkboxField` column widget.
The solution I proposed turned out to be not entirely correct. In my case the `yiiGridView.setSelectionColumn()` method works fine with any number of columns, but GridView also has the `yiiGridView::getSelectedRows()` method, which has no parameters and returns the selected columns. However, to work with several columns, its signature must be changed by adding an optional parameter to clarify which column selects rows needs to be returned.
In order not to lose backward compatibility, this parameter must be optional.
Tests for this will have to write decently. )
I would change the ticket label from a bug to something else ... Because its not a bug.
The question remains, is there really a need for such modernization? )

samdark:
> The question remains, is there really a need for such modernization? )

It is not a pressing matter for sure. If you have more time, you're welcome but I'd spent it on docs / tests instead.

samdark:
Closing the pull request since we agreed that this code isn't for the master branch.

PR #18104, opened 2020-06-14
blacklist → exclude

brandonkelly:
In the same vein as #18102, we should avoid using “whitelist” and “blacklist” terms.

Yii doesn’t have any concepts or API docs that use those terms, so this PR only touches a couple things in the docs.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/docs/guide/rest-resources.md
+++ b/docs/guide/rest-resources.md
@@ -82,7 +82,7 @@ public function fields()
 }
 
 // filter out some fields, best used when you want to inherit the parent implementation
-// and blacklist some sensitive fields.
+// and block some sensitive fields.
 public function fields()
 {
     $fields = parent::fields();
--- a/docs/guide/rest-resources.md
+++ b/docs/guide/rest-resources.md
@@ -82,7 +82,7 @@ public function fields()
 }
 
 // filter out some fields, best used when you want to inherit the parent implementation
-// and block some sensitive fields.
+// and exclude some sensitive fields.
 public function fields()
 {
     $fields = parent::fields();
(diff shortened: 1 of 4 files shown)


docs/guide/rest-resources.md
```diff
@@ -82,7 +82,7 @@ public function fields()
 }
 
 // filter out some fields, best used when you want to inherit the parent implementation
-// and blacklist some sensitive fields.
+// and block some sensitive fields.
```
samdark, line 85:
```suggestion
// and exclude some sensitive fields.
```

docs/guide/runtime-logging.md
```diff
@@ -142,8 +142,8 @@ the pattern `yii\db\*`.
 If you do not specify the [[yii\log\Target::categories|categories]] property, it means the target will process
 messages of *any* category.
 
-Besides whitelisting the categories by the [[yii\log\Target::categories|categories]] property, you may also
-blacklist certain categories by the [[yii\log\Target::except|except]] property. If the category of a message
+In addition to specifying allowed categories using the [[yii\log\Target::categories|categories]] property, you may also
+block certain categories by the [[yii\log\Target::except|except]] property. If the category of a message
```
samdark, line 146:
```suggestion
exclude certain categories by the [[yii\log\Target::except|except]] property. If the category of a message
```

docs/guide/structure-models.md
```diff
@@ -478,7 +478,7 @@ public function fields()
 }
 
 // filter out some fields, best used when you want to inherit the parent implementation
-// and blacklist some sensitive fields.
+// and block some sensitive fields.
```
samdark, line 481:
```suggestion
// and exclude some sensitive fields.
```

samdark (changes requested):

brandonkelly:
@samdark good call, just noticed that PHPUnit is going with “exclude” as well - https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-9.3.md#930---2020-08-07

samdark (approved):

PR #18106, opened 2020-06-15
master/slave → primary/replica (take 2)

brandonkelly:
This PR reimplements the changes made in #18102 in a definitely non-breaking way, as suggested by @rob006 (https://github.com/yiisoft/yii2/pull/18102#issuecomment-643847335).

- The old master/slave properties are back, but deprecated.
- The new primary/replica properties are now implemented by magic getter/setter methods, and documented with `@property` tags.
- All Yii code is now referencing the deprecated properties/methods, to ensure that any overridden properties/methods are still used.

You can view the combined changes made by #18102 and this PR here: https://github.com/brandonkelly/yii2/compare/9a3093442c8f7a39d1fa5972b58127266d8504f7...61ead1d890b589db98834205cc92f5b01667c1db

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️


samdark (approved):

schmunk42:
This change is against good coding practices, since now you have two names for things which are **exactly** the same.

For Yii2 you can not entirely deprecate it, since it would require a new major version. Also, we are in feature/code-freeze state for Yii2, but are adding useless new properties.

To my understanding a slave and a replica are two different things, even for db-servers.

PS: How do I call a primary master and secondary master btw... primary primary and secondary primary?

PS2: And, what's the new name for the *master* branch, btw?

brandonkelly:
@schmunk42 See #18102 for context. 

> This change is against good coding practices, since now you have two names for things which are **exactly** the same.

It’s not a bad practice as long as it’s clear which one _should_ be used, which is precisely why you the old properties/methods are now deprecated.

> For Yii2 you can not entirely deprecate it, since it would require a new major version. Also, we are in feature/code-freeze state for Yii2, but are adding useless new properties.

Only breaking changes warrant new major versions. Deprecations are not breaking changes; they are warnings of future breaking changes. If this is going to be a breaking change in Yii 3, then the earlier it can be deprecated in Yii 2, the better, so people have more time to prepare.

> To my understanding a slave and a replica are two different things, even for db-servers.
> 
> PS: How do I call a primary master and secondary master btw... primary primary and secondary primary?

We are going with modern naming conventions here – see https://github.com/yiisoft/yii2/pull/18102#issuecomment-643662745 and the following comment.

> PS2: And, what's the new name for the _master_ branch, btw?

GitHub is [switching to `main` by default](https://thenextweb.com/dd/2020/06/15/github-plans-to-replace-racially-insensitive-terms-like-master-and-whitelist/), but that is obviously not something you can change via a PR, nor would it have any effect on the installable Yii package, so I’ll leave it up to the Yii team to decide whether they care to make that change.

schmunk42:
Please also remove all **white**spaces from the codebase.

The next thing will be to rename `parent` to `super` I guess, and inherited classes will need to check their privilege. :roll_eyes: 

---

We can think about changes like that for Yii 3, if they make sense, but for Yii 2 it's just actionism.

My 2 cents.

brandonkelly:
It’s a pretty insignificant change that moves Yii along with modern conventions & makes the code/docs more inclusive, with zero to low impact on your project depending on your level of OCD. Seemed like an obvious v2 change to me. ¯\\_(ツ)\_/¯

rob006:
> If this is going to be a breaking change in Yii 3, then the earlier it can be deprecated in Yii 2, the better, so people have more time to prepare.

That could make sense if renaming was done for Yii 3, and then deprecations are added to Yii 2 to make migration smoother (assuming that there will be sane migration path from Yii 2 to Yii 3, which I highly doubt). But Yii 3 still uses "non-modern" terminology, and it probably will be another fresh start without migrations path (from Yii 2) in mind. So at this point this is deprecation for the sake of deprecation - it only brings more work for people to adjust their code to new naming, without any benefits from this change.

brandonkelly:
I want to get to that. One thing at a time.

rob006:
> I want to get to that.

You should start with that. How can you predict how these deprecations will simplify migration to Yii 3 if you don't know how these interfaces will look like in Yii 3. Most of deprecated properties does not exist in Yii 3.

rob006:
BTW:

> It’s not a bad practice as long as it’s clear which one _should_ be used, which is precisely why the old properties/methods are now deprecated.

This is not that simple. If I need to override one of the changed methods, should I override original deprecated method (`getSlave()`), or new method (`getReplica()`)? Old one is deprecated, so at first glance I should override the new method. But if I do that, old code and extensions which still use the old method will ignore my override - I need to override deprecated method to make sure everything works, which is counter-intuitive. 
Similarly with properties - if I want to redefine property with new default value, I need to use deprecated property name. If I define property using new name (for example `public $enableReplicas = false;`) it will be ignored, since this property is never used directly (and it will also make related getter and setter useless). If I really need to use new interface to set default value for such property, I need to override constructor, which is much more tricky.

Redundant interfaces **are** confusing and create additional surface for mistakes. It is "necessary evil" if you want to provide migration path to new version with BC break, but if there is no migration path, "necessary" does not apply and only "evil" is left.

schmunk42:
@samdark please read the above posting from @rob006 and reconsider this merge

brandonkelly:
You’re the one that suggested we go this route!

samdark:
Seems both current and previous approaches to renaming are technically problematic, not fully backwards compatible and likely cause issues :(

samdark:
@brandonkelly thanks for trying it but I'm going to revert these changes.

We can use new terms in Yii 3 where there's no backwards compatibility yet. In case there will be a way to do renaming in Yii 2 without causing issues, I'll gladly merge it.

@rob006 thanks for re-checking technical side of things.

samdark:
Created a task in Yii 3 Trello board not to forget it: https://trello.com/c/GiGjG2y5/69-update-terminology

brandonkelly:
Alright, no worries. We will just make the same changes in our own extension of yii\db\Connection, and provide our own read-write splitting docs. Thanks for being responsive.

rob006:
> You’re the one that suggested we go this route!

I'm not sure if you're talking to me, but I was against this renaming in the first place. This PR is much better than the first one, since it is more BC and less magic, but it is still much worse than not touching anything and leaving old master-slave terminology.

PR #18109, opened 2020-06-17
Revert #18102 since renaming is technically problematic

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | reverts #18102

See https://github.com/yiisoft/yii2/pull/18106#issuecomment-645051180 for reasons.


rob006:
Changelog entry should be also removed.

PR #18114, opened 2020-06-18
Fix mismatch between interface and model

olleharstedt:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️/❌

### DIFF
--- a/framework/db/ActiveRecord.php
+++ b/framework/db/ActiveRecord.php
@@ -319,7 +319,7 @@ class ActiveRecord extends BaseActiveRecord
      * For a large set of models you might consider using [[ActiveQuery::each()]] to keep memory usage within limits.
      *
      * @param array $attributes attribute values (name-value pairs) to be saved into the table
-     * @param string|array $condition the conditions that will be put in the WHERE part of the UPDATE SQL.
+     * @param string|array|null $condition the conditions that will be put in the WHERE part of the UPDATE SQL.
      * Please refer to [[Query::where()]] on how to specify this parameter.
      * @param array $params the parameters (name => value) to be bound to the query.
      * @return int the number of rows updated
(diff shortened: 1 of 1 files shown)


framework/db/ActiveRecord.php
```diff
@@ -319,7 +319,7 @@ public function refresh()
      * For a large set of models you might consider using [[ActiveQuery::each()]] to keep memory usage within limits.
      *
      * @param array $attributes attribute values (name-value pairs) to be saved into the table
-     * @param string|array $condition the conditions that will be put in the WHERE part of the UPDATE SQL.
+     * @param string|array|null $condition the conditions that will be put in the WHERE part of the UPDATE SQL.
```
samdark, line 322: Seems to be incorrect. Everywhere further in the calls it's string|array.

framework/db/ActiveRecord.php
```diff
@@ -772,7 +772,7 @@ protected function deleteInternal()
      * Returns a value indicating whether the given active record is the same as the current one.
      * The comparison is made by comparing the table names and the primary key values of the two active records.
      * If one of the records [[isNewRecord|is new]] they are also considered not equal.
-     * @param ActiveRecord $record record to compare to
+     * @param ActiveRecordInterface $record record to compare to
```
samdark, line 775: That's not correct since there's `tableName()` call and it is not in `ActiveRecordInterface`.

samdark (changes requested):

samdark:
Closing since both changes aren't correct. Thanks for contributing anyway.

olleharstedt:
These changes are based on the Psalm analyzer. I can try and reproduce it.

PR #18119, opened 2020-06-18
Upgrade note for #18083

brandonkelly:
Adds a note to UPGRADE.md, to warn people that they should make sure their controllers’ `init()` methods are calling `parent::init()`.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/framework/UPGRADE.md
+++ b/framework/UPGRADE.md
@@ -51,6 +51,12 @@ if you want to upgrade from version A to version C and there is
 version B between A and C, you need to follow the instructions
 for both A and B.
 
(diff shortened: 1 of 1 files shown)


samdark:
👍

lubosdz:
This should be top-mentioned in the upgrade composer notes for 2.0.36.
After upgrading from 2.0.35 our sites started breaking down.
Took some time to find changes in BaseController, commit and then this note.
I bet I am not the only one ..

PR #18121, opened 2020-06-19
Fix #18047: Console Table Widget width issue with colorized input

### LINKED ISSUE #18047: Console Table Widget width issue with colorized input
Use colorized input data in Table Widget rows.

Width of columns should be calculated without console colorization markers.
[Correct Table](https://github.com/erkib/yii2-issues/blob/master/correct_table_widget_output.png)

### What do you get instead?
Messed up widths.
[Incorrect Table](https://github.com/erkib/yii2-issues/blob/master/incorrect_table_widget_output.png)

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.35
| PHP version      | 7.4.5
| Operating system | OS X 10.15.4

cheeseq:
Made an escaping character method, which invokes when calculating string width. Can someone have a look pls?

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #18047

### DIFF
--- a/framework/console/widgets/Table.php
+++ b/framework/console/widgets/Table.php
@@ -271,7 +271,7 @@ class Table extends Widget
                     $alreadyPrintedCells[$index] = true;
                 }
                 $chunk = $prefix . $chunk;
-                $repeat = $cellSize - mb_strwidth($this->escapeCharacters($chunk), Yii::$app->charset) - 1;
+                $repeat = $cellSize - Console::ansiStrwidth($chunk) - 1;
                 $buffer .= $chunk;
                 if ($repeat >= 0) {
                     $buffer .= str_repeat(' ', $repeat);
@@ -334,12 +334,9 @@ class Table extends Widget
         foreach ($columns as $column) {
             $columnWidth = max(array_map(function ($val) {
                 if (is_array($val)) {
-                    $encodings = array_fill(0, count($val), Yii::$app->charset);
-                    return max(array_map(function ($listItem) {
-                        return mb_strwidth($this->escapeCharacters($listItem), Yii::$app->charset);
-                    }, $val, $encodings)) + mb_strwidth($this->listPrefix, Yii::$app->charset);
+                    return max(array_map('yii\helpers\Console::ansiStrwidth', $val)) + Console::ansiStrwidth($this->listPrefix);
                 }
-                return mb_strwidth($this->escapeCharacters($val), Yii::$app->charset);
+                return Console::ansiStrwidth($val);
             }, $column)) + 2;
             $this->columnWidths[] = $columnWidth;
             $totalWidth += $columnWidth;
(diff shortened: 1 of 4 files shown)


Thoulah:
Why not use `Console::ansiStrlen` and/or `Console::stripAnsiFormat`?

cheeseq:
> Why not use `Console::ansiStrlen` and/or `Console::stripAnsiFormat`?

Indeed, thanks for the tip

cheeseq:
Used `Console::stripAnsiFormat` and added tests.
Not sure about `testColorizedInputStripsANSIMarkersInternally()` because it uses reflection to access protected property. I can remove this one if need.

framework/console/widgets/Table.php
```diff
@@ -241,36 +241,37 @@ protected function renderRow(array $row, $spanLeft, $spanMiddle, $spanRight)
...
                     }
                 } else {
-                    $chunk = mb_substr($cell, ($cellSize * $i) - ($i * 2), $cellSize - 2, Yii::$app->charset);
+                    if(!isset($alreadyPrintedCells[$index]))
+                        $chunk = $cell;
+                    $alreadyPrintedCells[$index] = true;
                 }
                 $chunk = $prefix . $chunk;
-                $repeat = $cellSize - mb_strwidth($chunk, Yii::$app->charset) - 1;
+                $repeat = $cellSize - mb_strwidth($this->escapeCharacters($chunk), Yii::$app->charset) - 1;
```
Thoulah, line 274: I am not sure about the necessity of `Yii::$app->charset`, but I think it is preferred to either do
`$repeat = $cellSize - mb_strwidth(Console::stripAnsiFormat($chunk), Yii::$app->charset) - 1;`
or
`$repeat = $cellSize - Console::ansiStrlen($chunk) - 1;`
and get rid of the function below altogether as it serves no additional purpose. (after changing all instance of such code of course, I will not mark them all)

ghost (changes requested):

framework/console/widgets/Table.php
cheeseq: I'm not 100% sure about `Console::ansiStrlen()` method, because it uses `mb_strlen()` inside (not `mb_strwidth()` which is used all across `Table` widget for some reason)
What about my own `escapeCharacters()` method, well initially i thought about stripping only such characters, that may have effect on string width, but since we used `Console::stripAnsiFormat()`, this probably doesn't make sense. Will do

Thoulah: Agreed on `Console::ansiStrlen()`, I failed to see the usage of `mb_strwidth` instead of `mb_strlen` :unamused:

Perhaps it is a good idea to introduce a `Console::ansiStrwidth`. What is your idea about that @samdark ?

Thoulah:
```
    /**
     * Returns the width of the string without ANSI color codes.
     * @param string $string the string to measure
     * @return int the width of the string not counting ANSI format characters
     * @since 2.0.36
     */
    public static function ansiStrwidth($string)
    {
        return mb_strwidth(static::stripAnsiFormat($string), Yii::$app->charset);
    }
```

Thoulah:
@samdark Yes or no on a new function `Console::ansiStrwidth`?

framework/console/widgets/Table.php
samdark: I'm fine with introducing it.

ghost (approved):

samdark (approved):

samdark:
Merged. Thank you!

Issue #18131, opened 2020-06-23
[BUG] Method ActiveRecord :: find


rafaelt88:
Is it my impression or is it really a terrible Framework bug?

When you use ActiveRecord :: findOne ('1') or ActiveRecord :: findOne ('a1') or ActiveRecord :: findOne ('1a').

Perform the search for the first numeric portion, which should not be if my primary key is 1.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.35
| PHP version      | 7.4.2
| Operating system | Windows 7 x64

rafaelt88:
![image](https://user-images.githubusercontent.com/4501606/85374540-f1d43600-b502-11ea-8cf7-3fe6a8f60ad5.png)
![image](https://user-images.githubusercontent.com/4501606/85374596-044e6f80-b503-11ea-8894-c68b027db660.png)
![image](https://user-images.githubusercontent.com/4501606/85374640-13352200-b503-11ea-9124-758f5370eade.png)

rafaelt88:
@samdark

samdark:
Let's check it. I've put together a simple test. See the pull request.

alex-code:
I'm on MySQL 8 and running the queries below all return the same row (Queries run direct, outside of framework).
```sql
SELECT * FROM `item` WHERE `id`='1';
SELECT * FROM `item` WHERE `id`='1a';
SELECT * FROM `item` WHERE `id`='1a1';
```

It looks like MySQL sees the column as an `INT` and only uses the first valid numerical part.

rafaelt88:
Ok, but with my version MariaDB 10.4.11 fail.
I'm use xampp 7.4.6 exactly.

samdark:
So overall we're not doing any special validation on the framework side and do not plan to add it. It all depends on how underlying database behaves and how it is configured.

My6UoT9:
@rafaelt88 

read up on "Strict SQL Mode",  https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sql-mode-strict

> Strict mode controls how MySQL handles invalid or missing values in data-change statements such as INSERT or UPDATE. A value can be invalid for several reasons. For example, it might have the wrong data type for the column, or it might be out of range.

rafaelt88:
Yes it's error SQL, sorry. First time that its occured.

rafaelt88:
Can anybody help me...
[...]

PR #18132, opened 2020-06-23
A test for #18131

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #18131


rafaelt88:
@samdark 
I have installed MySQL 8.0.20 & MySQL 5.7.29, exactly the same problem occurs.
I don't think it's configuration since Yii1 runs me perfectly.
I am migrating a system from Yii 1 and this error has seriously doubted me.

samdark:
@rafaelt88 as was noted in https://github.com/yiisoft/yii2/issues/18131#issuecomment-648123359, that's MySQL itself that behaves like that. As far as I know, there is nothing different in Yii 1.1 about it so it should behave in the same way.

Issue #18113, opened 2020-06-18
Incorrect default value for BIT columns that allow NULL


trombipeti:
### What steps will reproduce the problem?
Create a mysql table with a column of type "BIT(1) NULL DEFAULT NULL", and get its schema info from Yii. (here size doesn't matter, so any BIT(N) column is ok).

### What is the expected result?
The BIT(N) column's defaultValue shoud be null.

### What do you get instead?
The column's defaultValue is 0.

### Additional info
Seems like https://github.com/yiisoft/yii2/commit/651e6a3feb0b89405adccdb1db688f6c37106d63 is the original commit, which seems to affect cubrid, mysql, pgsql. I have only tried it with mysql though.

I can make a fix & a pull request but I can only test it with mysql.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.35
| PHP version      | 7.4.7
| Operating system | Arch Linux

samdark:
So is that about schema introspection only?

samdark:
It is a good idea to start with a unit test that fails.

trombipeti:
Ok, I have some progress, creating a failing test was trickier than I thought, I had to use assertSame() instead of assertEquals() to compare the default value (0 and null).

samdark:
@trombipeti You can do a pull request so we can view test runs together but the test looks alright along with how it was executed. I think we can proceed with the actual fix. Good job 👍

trombipeti:
Seems like this only affects mysql, do you agree?

samdark:
1. cubrid tests aren't currently running in CI.
2. You've stated that it may affect pgsql but it's not checked in tests.

trombipeti:
I looked quickly in the source and it looked like pgsql wasn't affected, but the newly created unit tests prove otherwise.

I've (hopefully) implemented the fix for all 3 DBMSes.

terabytesoftw:
@samdark @rob006 @bizley I can add the tests for each driver and fix it where it should, this way we would have the bit column working correctly.

terabytesoftw:
Solved in `22.0` branch.

PR #18133, opened 2020-06-25
Fix for #18113

trombipeti:
| Q             | A
| ------------- | ---
| Is bugfix?    | Yes
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌
| Fixed issues  | #18113

Add tests & hopefully a fix for distinguishing default NULL for 'bit' columns.


trombipeti:
Seems like e820324 isn't necessary, e.g. curbid is working fine. It might still make some sense to keep the test for it.

trombipeti:
pgsql still looks bad, I can't run any tests locally though. Could someone take a look at it?

Issue #18094, opened 2020-06-10
Composite file extension validation is not supported


saidbakr:
### What steps will reproduce the problem?
 Try to validate composite file extension such as `tar.xz`

### What is the expected result?
 File name `file.bar.xz` should not be uploaded only `file.tar.xz` should be allowed.

### What do you get instead?
 Any file ended with `.xz` is uploaded.

### Additional info
 [Checkout this question on Stack Overflow](https://stackoverflow.com/questions/62268584/yii2-file-validation-tar-xz?noredirect=1#comment110192522_62268584)
| Q                | A
| ---------------- | ---
| Yii version      | 2.0.?
| PHP version      | 7
| Operating system |

saidbakr:
@samdark I don't, exactly, understand what you have suggested. However, I could understand that it should allow the developer to examine whatever extension he want. In other words, any file name extension that the server's operating system allow as a valid file name.
```
public function rules()
    {
        return [
            [['imageFile'], 'file', 'skipOnEmpty' => false, 'extensions' => 'png, jpg, lib.inc, fox.box.lux'],
        ];
    }
```
@rob006 You have mixed between file type or content type and file extension. In the regarded Wikipedia at the third sentences of the first paragraph, the abstract definition of the file extension. What you have regarded has no any conflict with what I have said or with this definition.  

> **A filename extension is typically delimited from the filename with a full stop (period), but in some systems it is separated with spaces.** 

I think that `pathinfo` deals with the file name as string data. In other words, if a plain text file has been created and the name has been changed from `my.txt` to `my.png`, there is no way in the PHP to determine the real content type of that file.

rob006:
> What you have regarded has no any conflict with what I have said or with this definition.

My interpretation has no conflict either (`mysite.index` is file name and `html` is extension - they're separated by dot), and it match further clarification for file names with multiple dots. You're just focusing on single ambiguous sentence (it does not say which dot, you just blindly assumed that it is first) and ignoring rest of the article.

> I think that `pathinfo` deals with the file name as string data. In other words, if a plain text file has been created and the name has been changed from `my.txt` to `my.png`, there is no way in the PHP to determine the real content type of that file.

[...]

PR #18137, opened 2020-06-27
Support for composite file extension validation

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️/❌
| Fixed issues  | #18094 

Support for composite file extension validation (#18094)


samdark:
@saidbakr would you please test it?

saidbakr:
> @saidbakr would you please test it?

I will test as soon as possible today.

saidbakr:
> @saidbakr would you please test it?

Please, could you tell me how to update Yii to the latest development version. I use basic app and I have tried, `composer update` and `composer update yii2`. Is there any possible changes to the dependencies in `composer.json`? 

```
{
    "name": "yiisoft/yii2-app-basic",
    "description": "Yii 2 Basic Project Template",
    "keywords": ["yii2", "framework", "basic", "project template"],
    "homepage": "http://www.yiiframework.com/",
    "type": "project",
    "license": "BSD-3-Clause",
    "support": {
        "issues": "https://github.com/yiisoft/yii2/issues?state=open",
        "forum": "http://www.yiiframework.com/forum/",
        "wiki": "http://www.yiiframework.com/wiki/",
        "irc": "irc://irc.freenode.net/yii",
        "source": "https://github.com/yiisoft/yii2"
    },
    "minimum-stability": "stable",
    "require": {
        "php": ">=5.6.0",
        "yiisoft/yii2": "~2.0.14",
        "yiisoft/yii2-bootstrap": "~2.0.0",
        "yiisoft/yii2-swiftmailer": "~2.0.0 || ~2.1.0"
    },
    "require-dev": {
        "yiisoft/yii2-debug": "~2.1.0",
        "yiisoft/yii2-gii": "~2.1.0",
        "yiisoft/yii2-faker": "~2.0.0",
        "codeception/codeception": "^4.0",
        "codeception/verify": "~0.5.0 || ~1.1.0",
        "codeception/specify": "~0.4.6",
        "symfony/browser-kit": ">=2.7 <=4.2.4",
        "codeception/module-filesystem": "^1.0.0",
        "codeception/module-yii2": "^1.0.0",
        "codeception/module-asserts": "^1.0.0"
    },....
```

My6UoT9:
@saidbakr u can try:
`"yiisoft/yii2": "dev-master",` instead of `"yiisoft/yii2": "~2.0.14",`
then it should get the latest version, but I did not test that.

darkdef:
```
"yiisoft/yii2": "dev-master"
```
It's valid, but this changes is not merged into dev-master

saidbakr:
I have updated it to `Yii version 2.0.36-dev` by changing the dependency in `composer.json` for Yii2 from  `"yiisoft/yii2": "~2.0.14"` to `"yiisoft/yii2": "2.0.x-dev"` and then running `composer update yiisoft/yii2`. I think it should be updated to the latest development version.

Now, the issue, as my understand, still the same. The following is the validation rule of `ContactForm` model:
```
class ContactForm extends Model
{
    public $name;
    public $email;
    public $subject;
    public $body;
    public $filen;
    public $verifyCode;


    /**
     * @return array the validation rules.
     */
    public function rules()
    {
        return [
            
            [['filen'], 'file', 'extensions' => 'br.fox, png'],
            
        ];
    }
```
When I try to upload file named `a.br.fox` the validation returns error
image

samdark:
@saidbakr no, it's not dev version. It is a fork.

saidbakr:
> @saidbakr no, it's not dev version. It is a fork. Either you can try specifying it or do a hot-patch manually by overwriting files in vendor (just for testing).

I have copied the two files to there paths in vendor/yiisoft, changed the server port `php -S localhost:7075` from `7070` and clearing all cache by deleting `runtime/cache` and `web/assets` contents.
The client-side JavaScript validation works fine, but after submit the server-side validation returns the same error

samdark:
At least something :)

samdark:
@saidbakr would you please share your validator config?

saidbakr:
> @saidbakr would you please share your validator config?

It is in a previous comment. However, it is here again:
```
/**
     * @return array the validation rules.
     */
    public function rules()
    {
        return [
            
            [['filen'], 'file', 'extensions' => 'br.fox, png'],
            
        ];
    }
```

samdark:
Would you please add:

```php
'checkExtensionByMimeType' => false
```

as we've discussed in the issue?

saidbakr:
> Would you please add:
> 
> ```
> 'checkExtensionByMimeType' => false
> ```
> 
> as we've discussed in the issue?

**Oh no, I am sorry, I forgot it!**
I added it and it works fine now. The file has been uploaded successfully.
```
/**
     * @return array the validation rules.
     */
    public function rules()
    {
        return [
            
            [['filen'], 'file', 'extensions' => 'br.fox, png', 'checkExtensionByMimeType' => false],
            
        ];
    }
```
However, we may have to make an Ajax validation to reflect the absence of `checkExtensionByMimeType` in the client-side. In other words, the JavaScript validation before submit should has any way to check the mimetype from the server before submit too.

samdark:
@saidbakr that's not the same thing:

```php
/**
     * @var bool whether to check file type (extension) with mime-type. If extension produced by
     * file mime-type check differs from uploaded file extension, the file will be considered as invalid.
     */
    public $checkExtensionByMimeType = true;
```

Setting it to `false` doesn't mean that mime type won't be checked in case you specify `$mimeTypes`. Actually, you should specify it.

samdark:
> However, we may have to make an Ajax validation to reflect the absence of checkExtensionByMimeType in the client-side.

Do you mean AJAX or client validation?

saidbakr:
> > However, we may have to make an Ajax validation to reflect the absence of checkExtensionByMimeType in the client-side.
> 
> Do you mean AJAX or client validation?

I don't think that JavaScript as a client side could able to check mimetype. So I regarded Ajax. The current implementation tells that JavaScript just check the extension as a string only.

samdark:
> I don't think that JavaScript as a client side could able to check mimetype

It is currently doing that :)

saidbakr:
> > I don't think that JavaScript as a client side could able to check mimetype
> 
> It is currently doing that :) https://github.com/yiisoft/yii2/blob/master/framework/assets/yii.validation.js#L419

Very good, the last update of `yii.validation.js` you have mentioned is doing it in both sides. However, the error message, I think, should be improved to be more descriptive. i.e I have the following message:

> Only files with these **extensions** are allowed: br.fox, png.

I think in the case of `checkExtensionByMimeType => true` the error message should be something like:

> Only files with these **mime types** are allowed: br.fox, png.

saidbakr:
**Side Note**
In such cases of validation rules that may depend on any client-side scripting, the developer should be mentioned to clear `web/assets` contents, or preventing assets caching from the configuration file under components:
```
'assetManager' => [
            'class' => 'yii\web\AssetManager',
            'forceCopy' => true, //Prevent caching assets during development
            'linkAssets' => true, //Prevent caching assets during development
]
```
Without clearing `web/assets`, the change of `checkExtensionByMimeType` from true to false or vice versa will not be reflected on the client-side directly.

samdark:
`checkExtensionByMimeType` does not affect how frontend works at all. What it does when turned on is:

1. For an extension a common mime type is determined based on config https://github.com/yiisoft/yii2/blob/master/framework/helpers/mimeTypes.php.
2. Extension is checked. If it's valid then mime-type from point 1 is checked in the file.

In case it's turned off, mime-type isn't automatically checked but you can still specify it via `mimeTypes`.

samdark:
@rob006 would you please take a look at the code?

framework/validators/FileValidator.php
```diff
@@ -412,7 +413,11 @@ protected function validateExtension($file)
             }
         }
 
-        if (!in_array($extension, $this->extensions, true)) {
+        if (is_array($this->extensions) and count($this->extensions)) {
+            foreach ($this->extensions as $ext) {
+                if (StringHelper::endsWith($file->name, $ext, false))
+                    return true;
+            }
```
rob006, line 420:
```suggestion
        if (!empty($this->extensions)) {
            foreach ((array) $this->extensions as $ext) {
                if (StringHelper::endsWith($file->name, ".$ext", false))
                    return true;
            }
```

rob006 (changes requested):
Tests are missing.

saidbakr:
> `checkExtensionByMimeType` does not affect how frontend works at all. What it does when turned on is:
> 
>     1. For an extension a common mime type is determined based on config https://github.com/yiisoft/yii2/blob/master/framework/helpers/mimeTypes.php.
> 
>     2. Extension is checked. If it's valid then mime-type from point 1 is checked in the file.
> 
> 
> In case it's turned off, mime-type isn't automatically checked but you can still specify it via `mimeTypes`.

Let me explain the issue in other words:

**1** Starting with validation below in the model:
```
return [
            [['filen'], 'file', 'extensions' => 'br.fox, png', 'checkExtensionByMimeType' => true],
]
```

**2** Trying to upload file named `a.br.fox`, there is no any error rendered before submit but after submit, an error is returned and the file does not be uploaded.
```

**Comment by samdark** [comment]
That is expected. We don't have a list of mime types on the client so leaving that to server validation.

File: framework/validators/FileValidator.php (hunk 1 above)
Comment (rob006) [reply framework/validators/FileValidator.php]: This actually could be simplified to:

```php
        foreach ($this->extensions as $ext) {
            if (StringHelper::endsWith($file->name, ".$ext", false)) {
                return true;
            }
        }
```
since `$extensions` is normalized to array in `init()`.

framework/assets/yii.validation.js
```diff
@@ -408,10 +408,18 @@ yii.validation = (function ($) {
...
     function validateFile(file, messages, options) {
         if (options.extensions && options.extensions.length > 0) {
-            var index = file.name.lastIndexOf('.');
-            var ext = !~index ? '' : file.name.substr(index + 1, file.name.length).toLowerCase();
+            var found = false;
+            var filename = file.name.toLowerCase();
 
-            if (!~options.extensions.indexOf(ext)) {
+            for (var index=0; index < options.extensions.length; index++) {
+                var ext = options.extensions[index].toLowerCase();
```
rob006, line 415: You should include dot here (so you will search '.html` for `html` extension) - right now it will validate `indexhtml` for `html` extension.

tests/framework/validators/FileValidatorTest.php
```diff
@@ -452,6 +452,25 @@ public function testValidateAttributeType()
         $this->assertNotFalse(stripos(current($m->getErrors('attr_exe')), 'Only files with these extensions '));
     }
 
+    public function testValidateAttributeDoubleType()
+    {
+        $val = new FileValidator([
+            'extensions' => 'tar.gz, tar.xz',
+            'checkExtensionByMimeType' => false,
```
rob006, line 459: How about tests if `checkExtensionByMimeType` is enabled?

darkdef: mime type for ".gz" - not exists in Yii2 mimeTypes.php

framework/assets/yii.validation.js
darkdef: fixed

tests/framework/validators/FileValidatorTest.php
darkdef: ok, i'm add test for other extension from issue - tar.xz

tests/framework/validators/FileValidatorTest.php
```diff
@@ -496,6 +522,7 @@ public function validMimeTypes()
             ['test.txt', 'text/*', 'txt'],
             ['test.xml', '*/xml', 'xml'],
             ['test.odt', 'application/vnd*', 'odt'],
+            ['test.tar.xz', 'application/x-xz', 'tar.xz'],
```
samdark, line 525:
```suggestion
            ['test.tar.xz', 'application/x-xz', 'xz'],
```

samdark: Mime types are specified for topmost extension.

darkdef: In FileValidator.php:validateExtension in row 401 get extension from real filename.
After - checks mimetype from uploaded file, get all variants of extensions for this mimetype and compare this variants with real extension of file.
For same filename.tar.xz - real extension from row 401 - xz

samdark (approved):

samdark (approved):

samdark:
Thanks!

Issue #18134, opened 2020-06-26
Illegal offset type in isset or empty


srnden:
### What steps will reproduce the problem?
Hi, i'm trying add WHERE LIKE condition to jsonb column.
```php
$query = MyModel::find();
$query->andFilterWhere(['ilike', new Expression("lang_content -> '" . $lang->id . "' ->> 'title'"), $string]);
$result = $query->all();
```
### What is the expected result?
Executed query and returns all results as an array.
### What do you get instead?
```
PHP Warning 'yii\base\ErrorException' with message 'Illegal offset type in isset or empty' 

in vendor/yiisoft/yii2/db/Connection.php:930

Stack trace:
#0 vendor/yiisoft/yii2/db/Connection.php(930): yii\web\ErrorHandler->handleError()
#1 vendor/yiisoft/yii2/db/conditions/LikeConditionBuilder.php(70): yii\db\Connection->quoteColumnName()
#2 vendor/yiisoft/yii2/db/QueryBuilder.php(288): yii\db\conditions\LikeConditionBuilder->build()
#3 vendor/yiisoft/yii2/db/QueryBuilder.php(1580): yii\db\pgsql\QueryBuilder->buildExpression()
#4 vendor/yiisoft/yii2/db/QueryBuilder.php(1357): yii\db\pgsql\QueryBuilder->buildCondition()
#5 vendor/yiisoft/yii2/db/QueryBuilder.php(235): yii\db\pgsql\QueryBuilder->buildWhere()
#6 vendor/yiisoft/yii2/db/ActiveQuery.php(328): yii\db\pgsql\QueryBuilder->build()
#7 vendor/yiisoft/yii2/db/Query.php(478): yii\db\ActiveQuery->createCommand()
...
#27 backend/web/index.php(17): backend\components\web\Application->run()
#28 {main}
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.35
| PHP version      | 7.1
| Operating system | Ubuntu 16.04
| DB | PostgreSQL 9.6

srnden:
https://github.com/yiisoft/yii2/issues/17473#issuecomment-649589526

PR #17490, opened 2019-08-01
#17473 fix simple condition build

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #17473

Issue #17473, opened 2019-07-29
Illegal offset type in isset or empty


vitovtnet:
### What steps will reproduce the problem?
Query class

```
class OrderQuery extends ActiveQuery
{
 public function beforeOpeningDate(\DateTime $openingDate)
    {
        $this->andFilterWhere(['<=', new Expression('operation_date') , $openingDate->format('Y-m-d')]);
        return $this;
    }
}

```

`Order::find()->beforeOpeningDate(new \DateTime())->all();
`

And result is

```
yii\base\ErrorException
Illegal offset type in isset or empty
```

If I use query builder andFilterWhere without Expression -  everything is working fine

`
$this->andFilterWhere(['<=', operation_date , $openingDate->format('Y-m-d')]);
`
### What is the expected result?

QueryBuilde understanding  Expression 

### Additional info

Trace
```

     * @param string $name column name
     * @return string the properly quoted column name
     */
    public function quoteColumnName($name)
    {
        if (isset($this->_quotedColumnNames[$name])) {
            return $this->_quotedColumnNames[$name];
        }
        return $this->_quotedColumnNames[$name] = $this->getSchema()->quoteColumnName($name);
    }
```

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.23
| PHP version      |  7.2
| Operating system | Ubuntu 18.04

alexkart:
@vitovtnet is it allowed to pass `Expression` as `operand1` in operator format condition?
And it's not clear what you want to achieve using this?
As you said `$this->andFilterWhere(['<=', operation_date , $openingDate->format('Y-m-d')]);` works correct and it is correct condition.

vitovtnet:
@alexkart Expression help to use SQL in RAW format without any quotas.

`$this->andFilterWhere(['<=', operation_date , $openingDate->format('Y-m-d')]);`

SQL: ...WHERE `operation_date`<='2019-05-05'

But if I want to use, for example DATE_FORMAT fucntion or DAYS function or any other SQL function I will write my query like

`$this->andFilterWhere(['<=', new Expression("DATE_FORMAT('peation_date', '....')" , $openingDate->format('Y-m-d')]);`

vitovtnet:
Also, I checked all mo code which written 05.04.2018 (more than a year ago), I'm everywhere used a construction like

```
 if (!empty($this->endDate)) {
            $endDate = new \DateTime($this->endDate);

            $query->andFilterWhere([
                '<=',
[...]

PR #18142, opened 2020-06-29
Db expression

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️/❌
| Tests pass?   | ✔️
| Fixed issues  | #18134 

 Expression is not will be quoted


framework/db/Connection.php
```diff
@@ -925,11 +925,16 @@ public function quoteTableName($name)
      * If the column name contains prefix, the prefix will also be properly quoted.
      * If the column name is already quoted or contains special characters including '(', '[[' and '{{',
      * then this method will do nothing.
+     * If the column name is Exresion - return as is
```
samdark, line 928:
```suggestion
     * If the column name is an instance of `ExpressionInterface` it is returned as is.
```

framework/db/Connection.php
```diff
@@ -925,11 +925,16 @@ public function quoteTableName($name)
...
      * @param string $name column name
      * @return string the properly quoted column name
      */
     public function quoteColumnName($name)
     {
+        // Expression is not will be quoted (https://github.com/yiisoft/yii2/issues/18134)
```
samdark, line 934:
```suggestion
```

framework/db/Connection.php
```diff
@@ -925,11 +925,16 @@ public function quoteTableName($name)
      * If the column name contains prefix, the prefix will also be properly quoted.
      * If the column name is already quoted or contains special characters including '(', '[[' and '{{',
      * then this method will do nothing.
+     * If the column name is Exresion - return as is
      * @param string $name column name
```
samdark, line 929:
```suggestion
     * @param string $name|ExpressionInterface column name
```

tests/framework/db/QueryBuilderTest.php
```diff
@@ -2523,6 +2523,17 @@ public function testBuildLikeCondition($condition, $expected, $expectedParams)
         $this->assertEquals($expectedParams, $params);
     }
 
+    /**
+     * @see https://github.com/yiisoft/yii2/issues/18134
+     */
+    public function testIssue18134()
```
samdark, line 2529:
```suggestion
    public function testExpressionIsNotQuotedInColumnName()
```

samdark (changes requested):
Overall it's fine. Need some minor things fixed and a line for CHANGELOG.md

samdark (approved):

samdark:
Thanks!

rob006 (commented):
I don't think that this is right place to fix it. Expressions should be handled at `LikeConditionBuilder::build()` level, there is no need to add special conditions in `Connection::quoteColumnName()`.

See https://github.com/yiisoft/yii2/pull/17490

darkdef:
But what about the case when the Expression will be used instead of the column name in other cases?
This issue (for example): https://github.com/yiisoft/yii2/issues/17473

framework/db/Connection.php
```diff
@@ -925,11 +925,15 @@ public function quoteTableName($name)
      * If the column name contains prefix, the prefix will also be properly quoted.
      * If the column name is already quoted or contains special characters including '(', '[[' and '{{',
      * then this method will do nothing.
-     * @param string $name column name
+     * If the column name is an instance of `ExpressionInterface` it is returned as is.
+     * @param string $name|ExpressionInterface column name
```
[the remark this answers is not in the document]
samdark: Why?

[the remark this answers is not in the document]
darkdef: @param $name string|ExpressionInterface column name

rob006:
@darkdef What cases? #17473 is already fixed by #17490 in a way I proposed (by fixing `SimpleConditionBuilder::build()`). It just does not make sense to pass non-string to `Connection::quoteTableName()`, so code should be adjusted to not do that. This PR makes `Connection::quoteTableName()` interface more twisted only to silently handle bugs in other places.

samdark:
Yes, makes sense. Reverted. Should be fixed another way.

darkdef:
ok. I understood

framework/db/Connection.php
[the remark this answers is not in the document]
rob006: Valid syntax is:

```
@param string|ExpressionInterface $name column name
```

But this should not be changed in the first place.

PR #18147, opened 2020-07-01
Fixed parameters binding for mysql when prepare emulation is off

rskrzypczak:
The same as here: https://github.com/yiisoft/yii2/commit/2201bf14e239477b52cfb1f331e0c7b92d98e2cb
When PDO::ATTR_EMULATE_PREPARES => false in DB config this uses native preparation which means that you cannot use the same parameter twice.

It only works when $emulatePrepare = true, but it's usually false, so it doesn't work :)

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -


samdark:
@rskrzypczak would you please add a line to CHANGELOG.md? Thanks.

samdark:
Thanks!

PR #18151, opened 2020-07-03
Added `Mutex::isAcquired()` to check if lock is currently acquired

rhertogh:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | -


rhertogh:
The purpose is to support checking if a mutex is acquired. We need a `isMutexAcquired($name)` function which would return `true` or `false`. I could also include it in this pull request but since Yii2 doesn't accept new features I left it out of the PR.
If the `isMutexAcquired($name)` function would be accepted in the PR I would be happy to include it with some test coverage.

samdark:
What's the use case for checking if mutex is acquired without acquiring it?

rhertogh:
It would be used in a model's save function to ensure it is only called when the mutex is active. The mutex itself would be acquired in different places and wraps different models.

samdark:
I'd better add a method for this case. Better than exposing a property.

rhertogh:
ok, I'll implement the method.

framework/mutex/Mutex.php
```diff
@@ -97,6 +97,16 @@ public function release($name)
         return false;
     }
 
+    /**
+     * Checks if a lock is currently acquired
+     *
+     * @param string $name of the lock to check
+     * @return bool Returns true if currently acquired
```
samdark, line 104: Need `@since 2.0.36`

samdark (approved):

framework/mutex/Mutex.php
rhertogh: fixed

samdark:
Thanks!

PR #18184, opened 2020-07-17
ActiveQuery is not getting connection from a model

ziaratban:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ❌

default db connection not send from `$modelClass` in `one()` and `all()` methods.
so i create a `finalConObj()` protected method and use in across `ActiveQuery` class.


ziaratban:
please see [docs](https://www.yiiframework.com/doc/api/2.0/yii-db-activequery#one()-detail)
>$db : The DB connection used to create the DB command. If null, the DB connection returned by **$modelClass** will be used.

in [ActiveQuery.php#L304](https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveQuery.php#L304)
`$db` send direct to `Query` class and according to [this docs](https://www.yiiframework.com/doc/api/2.0/yii-db-activequery#one()-detail) is mistake.

**Let me give you an example**
```php
Class Table1 extend ActiveRecord{

}

Class Table2 extend ActiveRecord{
    public static function getDB(){
        return new Connection([..]); //connect to data center A
    }
}

$myOtherCon = new Connection([...]);
$myRecord1 = Table1::find()->one($myOtherCon); // in this line yii2 use $myOtherCon
$myRecord2 = Table1::find()->one(); // in this line yii2 use yii::$app->db

$myRecord3 = Table2::find()->one(); // in this line yii2 must be use Table2::getDB() but use yii::$app->db
```

framework/db/ActiveQuery.php
```diff
@@ -122,6 +122,16 @@ public function init()
         $this->trigger(self::EVENT_INIT);
     }
 
+    /**
+     * returns connection object from `modelClass` if `$db` is null otherwise returns `$db`
+     * @param null|Connection your custom connection object
+     * @return Connection returns a connection object base on input
+    */
+    protected function finalConObj($db){
```
samdark, line 130:
```suggestion
    protected function ensureConnection($db)
    {
```

bizley:
Requires UPGRADE note since it's BC breaking.

framework/db/ActiveQuery.php
ziaratban: `finalConObj` is `final connection object`. like the sentence `final decision`

bizley:
@ziaratban are you planning to keep working on this?

ziaratban:
@samdark @bizley 
> The issue is valid. Would you please add some unit tests for it?

i don't work with `phpunit`. so my codes is not efficient. (but little in `yii2-mongodb`)
> @ziaratban are you planning to keep working on this?

you can continue and i do not have any problem. 
👍

bizley:
I'll take it from here then.

bizley:
After preparing tests for this #18286 I can confirm that ActiveQuery is using model's connection in both `one()` and `all()` methods. `createCommand()` from `Query` is overwritten in `ActiveQuery` and it takes care of selecting proper connection. No changes are required.

ziaratban:
> 
> 
> After preparing tests for this #18286 I can confirm that ActiveQuery is using model's connection in both `one()` and `all()` methods. `createCommand()` from `Query` is overwritten in `ActiveQuery` and it takes care of selecting proper connection. No changes are required.

no, i can not confirm.

please see this graphical debug :
image

bizley:
You are showing `createCommand()` from `Query` here but `ActiveQuery` is using its own version as I said.

ziaratban:
yes `ActiveQuery` is using its own version but in this line([ActiveQuery.php:304](https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveQuery.php#L304)) we are calling the method `one()` of the query class , so we are using the `createCommand()` method of the `Query` class.

bizley:
No, it doesn't work like that.

ziaratban:
> 
> 
> No, it doesn't work like that.

yes you right.

this is very amazing for me.
http://sandbox.onlinephpfunctions.com/code/a7f8c2d8d417d3522013a28604551b067e9f6804

Issue #15501, opened 2018-01-15
Exceptions are not documented for save()


turkelt0n:
### What is the expected result?

A phpdoc for BaseActiveRecord::save() that suggest the types of exceptions which can occur during execution.

### What do you get instead?

No indication of any exception

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11

samdark:
Yes, nice to have.

samdark:
Checked it more. See https://github.com/yiisoft/yii2/pull/18186#issuecomment-660708196

turkelt0n:
> usually nothing could be done about these exceptions, there's no need to declare it in the phpdoc

I agree that there is no way for the ActiveRecord to act upon these exceptions but this does not imply that there is no reposibility to declare that they can happen. 

Also as i understand the AR should be (somewhat) agnostic to the underlying db-technology. So my expectation would be that the save-method would behave in the same manner if something fails. For me it doesnt matter if they throw an exception or return false, as long as this is consitent for all db-technology and i can safely exchange them without touching any other code

samdark:
> but this does not imply that there is no responsibility to declare that they can happen.

I think it's the case or else we'll have to declare `\Throwable` everywhere since errors/warnings/notices may happen and Yii converts these to exceptions.

samdark:
> So my expectation would be that the save-method would behave in the same manner if something fails.

That's the case. Validation error results in `false`. Database problem results in an exception.

PR #18186, opened 2020-07-17
Add thrown exception phpdoc in BaseActiveRecord save method

SOHELAHMED7:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | -
| Fixed issues  | #15501


samdark:
That's not strictly correct. It returns false on save failure. It may or may not throw an exception if underlying layers do but since  usually nothing could be done about these exceptions, there's no need to declare it in the phpdoc.

PR #18196, opened 2020-07-22
DbManager invalidate assignment cache only

alex-code:
Addition of method `invalidateAssignmentCache` to invalidate assignment cache for specified user.

My use case may be a little specific but I've got a WebSocket server that needs to clear cached assignments when the user disconnects but doesn't need to clear the other cached items.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


samdark:
I'd say that is very specific. How about allowing current code to be adjusted instead of implementing the method? i.e. private -> protected/

alex-code:
That works for me.
Just need a way to free memory for users who are no longer connected.

Wan't me to update this PR or make a new one?

samdark:
Update this one.

alex-code:
Did you want a `@since` noting the change from `private` to `protected`?

samdark:
Yes, change of scope to higher visibility should be tagged with "since".

samdark (approved):

bizley (approved):

Issue #18210, opened 2020-08-03
wrong checkboxList items checked


johnny-silverhand:
### What steps will reproduce the problem?
`echo \yii\helpers\Html::checkboxList('test', ['5.1'], ['5' => '5', '5.1' => '5.1', '5.10' => '5.10']);`

### What is the expected result?
only one checkbox are checked  (key 5.1)

### What do you get instead?
two checkboxes are checked (5.1 & 5.10)

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.36
| PHP version      | 7.4
| Operating system | Linux

johnny-silverhand:
The same behavior is observed in default selection on a dropdown and radio list

alex-code:
If your `$selection` is a string it'll work,
e.g.
```php
echo \yii\helpers\Html::checkboxList('test', '5.1', ['5' => '5', '5.1' => '5.1', '5.10' => '5.10']);
```

When `$selection` is an array is does a different type of comparison (https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L971).

I think we'd need to change the comparison type to strict on the `ArrayHelper::isIn` call or add a new option to allow it's setting.

johnny-silverhand:
Thanks for the answer. Unfortunately, changing $selection to string doesn't work for me. I can have multiple default values.

alex-code:
@samdark Do you think we should make the `ArrayHelper::isIn` checks strict by default for:
`Html::checkboxList`
`Html::radioList`
`Html::renderSelectOptions`

Or have an extra option?

samdark:
@alex-code are there any possible breaks that may be caused by it?

alex-code:
There could be with numeric keys and selections.
```php
echo \yii\helpers\Html::checkboxList('test', 5.1, ['5' => '5', '5.1' => '5.1', '5.10' => '5.10']);
```
This wouldn't match with a strict check.
Maybe an extra option is better.

samdark:
Yes. Agree.

PR #18215, opened 2020-08-05
Fix 18210

alex-code:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #18210

Allow strict comparison for multi select inputs.
`in_array` does loose checking so can have unexpected results.

```php
$a = ['1.1', '1.10'];

foreach ($a as $v) {
    var_dump(in_array($v, ['1.1']));
}
// Returns, `true`, `true`.
foreach ($a as $v) {
    var_dump(in_array($v, ['1.1'], true));
}
// Returns, `true`, `false`.
```

### DIFF
--- a/framework/helpers/BaseHtml.php
+++ b/framework/helpers/BaseHtml.php
@@ -819,6 +819,7 @@ class BaseHtml
      *   Defaults to false.
      * - encode: bool, whether to encode option prompt and option value characters.
      *   Defaults to `true`. This option is available since 2.0.3.
+     * - strict: boolean, if `$selection` is an array and this value is true a strict comparison will be performed on `$items` keys. Defaults to false.
      *
      * The rest of the options will be rendered as the attributes of the resulting tag. The values will
      * be HTML-encoded using [[encode()]]. If a value is null, the corresponding attribute will not be rendered.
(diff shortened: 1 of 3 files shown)


framework/helpers/BaseHtml.php
```diff
@@ -819,6 +819,7 @@ protected static function booleanInput($type, $name, $checked = false, $options
      *   Defaults to false.
      * - encode: bool, whether to encode option prompt and option value characters.
      *   Defaults to `true`. This option is available since 2.0.3.
+     * - strict: boolean, if `$selection` is an array and this value is true a strict comparison will be performed on `$items` keys. Defaults to false.
```
samdark, line 822: Need to add "since 2.0.37"

framework/helpers/BaseHtml.php
```diff
@@ -931,6 +932,7 @@ public static function listBox($name, $selection = null, $items = [], $options =
      *   This option is available since version 2.0.16.
      * - encode: boolean, whether to HTML-encode the checkbox labels. Defaults to true.
      *   This option is ignored if `item` option is set.
+     * - strict: boolean, if `$selection` is an array and this value is true a strict comparison will be performed on `$items` keys. Defaults to false.
```
samdark, line 935: Need to add "since 2.0.37"

samdark (approved):

framework/helpers/BaseHtml.php
alex-code: I don't think `booleanInput` is effected by this.

alex-code: Never mind, it's showing above as `booleanInput` but I've added since to the `dropDownList` method.

samdark (approved):

samdark:
Thank you!

Issue #18217, opened 2020-08-05
Chained Exceptions are not rendered in console apps


mikehaertl:
### What steps will reproduce the problem?

Throw a chained exeption in a console app:

```
try {
    $some->foo();
} catch (Exception $e) {
    throw new MyException("My action failed", 0, $e);
}
```

### What is the expected result?

All chained exceptions should be rendered.

### What do you get instead?

Only the top most exception is rendered.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.36
| PHP version      |  any
| Operating system | any

PR #18218, opened 2020-08-05
Issue #18217 Render chained exceptions in console ErrorHandler

mikehaertl:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | unsure
| Tests pass?   | ✔️
| Fixed issues  | #18217

### DIFF
--- a/framework/console/ErrorHandler.php
+++ b/framework/console/ErrorHandler.php
@@ -65,7 +65,7 @@ class ErrorHandler extends \yii\base\ErrorHandler
         } else {
             echo $message . "\n";
         }
-        if (($previous = $exception->getPrevious()) !== null) {
+        if (YII_DEBUG && ($previous = $exception->getPrevious()) !== null) {
             $causedBy = "\n" . $this->formatMessage('Caused by: ', [Console::BOLD]);
             if (PHP_SAPI === 'cli') {
                 Console::stderr($causedBy);
--- a/framework/console/ErrorHandler.php
+++ b/framework/console/ErrorHandler.php
@@ -65,6 +65,15 @@ class ErrorHandler extends \yii\base\ErrorHandler
         } else {
             echo $message . "\n";
         }
+        if (($previous = $exception->getPrevious()) !== null) {
+            $causedBy = "\n" . $this->formatMessage('Caused by: ', [Console::BOLD]);
+            if (PHP_SAPI === 'cli') {
+                Console::stderr($causedBy);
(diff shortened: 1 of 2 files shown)


framework/console/ErrorHandler.php
```diff
@@ -65,6 +65,15 @@ protected function renderException($exception)
...
         }
+        if (($previous = $exception->getPrevious()) !== null) {
+            $causedMessage = "\n" . $this->formatMessage('Caused by:', [Console::BOLD, Console::FG_GREEN]) . "\n";
+            if (PHP_SAPI === 'cli') {
+                Console::stderr($causedMessage . "\n");
+            } else {
+                echo $causedMessage . "\n";
+            }
+            $this->renderException($previous);
+        }
```
mikehaertl, line 76: @samdark It's probably better to only do this in debug mode, right? If you agree I'll change it.

One thing thogh: I tried to keep "Caused by:" in the same line like the next "Exception 'XyzException' ..." message, but i could not figure out, where the linebreak is added. If you know and agree, that this would look better then please let me know.

mikehaertl: After more thinking: Chained exceptions should always be rendered, not only in debug mode. They will already look different in debug/non-debug mode. So the logic is fine.

Just not sure about the formatting. Comments welcome.

mikehaertl: Updated to improve formatting:

chained_exception

mikehaertl:
@samdark Why 2.0.38? Would be cool if this change made it into 2.0.37.

samdark:
@mikehaertl if I'll have time to review it before 2.0.37 then yes. As for now, I'm not sure about it. Want to tag it this week.

framework/console/ErrorHandler.php
samdark: Yes, it's better to do it in debug only. In non-debug mode we aren't printing too many details on purpose.

samdark:
Overall looks alright. Can be merged into current release in case `YII_DEBUG` comment is fixed and CHANGELOG is added.

framework/console/ErrorHandler.php
mikehaertl: Ok, changed it.

samdark (approved):

samdark:
Thanks!

PR #18233, opened 2020-08-16
Add PHP 8 support

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | -
| Fixed issues  | -


terabytesoftw:
It can be added separately from the main build and fixed until it passes and we move it to the main build.

samdark:
@terabytesoftw it passes. That's CodeClimate warning that make it look like it does not.

PR #18238, opened 2020-08-18
allow rate limiter to work with user closure

### LINKED ISSUE #18236: yii\filters\RateLimiter wont work with another user component.
When using a custom user component (in my case `partnerUser`) the rate limiting won't work as the recived `user` value is null.

Example of behaviors setup:

```php
        'class' => CompositeAuth::class,
        'authMethods' => [
            QueryParamAuth::class,
            HttpBearerAuth::class,
        ],
        'user' => Yii::$app->partnerUser,
];

$behaviors['rateLimiter'] = [
        'class' => RateLimiter::class,
        'user' => Yii::$app->partnerUser->identity,
];
```

Since the `'user' => Yii::$app->partnerUser->identity,` will receive null when setting up rateLimiter due to the login process is not done while setting up behaviors. So i think the `user` property in RateLimiter is useless.

The only solution i came up to solve my problem and also make `user` work is another check for a user service, similar to what is done when no user is specified (https://github.com/yiisoft/yii2/blob/master/framework/filters/RateLimiter.php#L84).

Solution to make `'user' => 'partnerUser'` work:

```php
// if the user is defined as service name
if (!is_object($this->user)) { 
    $this->user = Instance::ensure($this->user)->getIdentity();
}
``` 

https://github.com/yiisoft/yii2/blob/master/framework/filters/RateLimiter.php#L86

If this is an acceptable solution, i would do a PR. If there is another solution for my problem with custom user components let me know.

**update**:

Another proposed solution might be to accept closures:

```php
if ($this->user instanceof Closure) {
    $this->user = call_user_func($this->user);
}
```

nadar:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes & no
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | this closes #18236


framework/filters/RateLimiter.php
```diff
@@ -48,8 +49,14 @@ class RateLimiter extends ActionFilter
      */
     public $errorMessage = 'Rate limit exceeded.';
     /**
-     * @var RateLimitInterface the user object that implements the RateLimitInterface.
-     * If not set, it will take the value of `Yii::$app->user->getIdentity(false)`.
+     * @var RateLimitInterface|closure the user object that implements the RateLimitInterface. If not set, it will take the value of `Yii::$app->user->getIdentity(false)`.
```
samdark, line 52:
```suggestion
     * @var RateLimitInterface|Closure the user object that implements the RateLimitInterface. If not set, it will take the value of `Yii::$app->user->getIdentity(false)`.
```

framework/filters/RateLimiter.php
```diff
@@ -84,6 +91,10 @@ public function beforeAction($action)
             $this->user = Yii::$app->getUser()->getIdentity(false);
         }
 
+        if ($this->user instanceof Closure) {
+            $this->user = call_user_func($this->user, $action);
```
samdark, line 95:
```suggestion
            $this->user = $this->user($action);
```

samdark (approved):
Looks OK overall but please fix minor things I've pointed to.

framework/filters/RateLimiter.php
nadar: @samdark are you sure? `call_user_func` is commonly used in a lot of Yii framework code in this siutations.

https://github.com/yiisoft/yii2/blob/da447fd1bb9778b8b76994f3e3e14f9748c67625/framework/widgets/ListView.php#L148
https://github.com/yiisoft/yii2/blob/da447fd1bb9778b8b76994f3e3e14f9748c67625/framework/validators/RangeValidator.php#L97

But if you like, i can change it.

samdark: OK, let it be `call_user_func`...

samdark (approved):

Issue #18248, opened 2020-08-21
Multiple stack traces for chained exceptions in console app


mikehaertl:
### What steps will reproduce the problem?

Throw a chained exception in a console app

### What is the expected result?

You should only get one stack trace at the end. The same as the web error handler renders it.

### What do you get instead?

Multiple stack traces, one for each exception.

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.37
| PHP version      |  any
| Operating system | any

PR #18249, opened 2020-08-21
Issue #18248 Render only one stack trace for chained exceptions on console

mikehaertl:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  |❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #18248

### DIFF
--- a/framework/console/ErrorHandler.php
+++ b/framework/console/ErrorHandler.php
@@ -29,6 +29,7 @@ class ErrorHandler extends \yii\base\ErrorHandler
      */
     protected function renderException($exception)
     {
+        $previous = $exception->getPrevious();
         if ($exception instanceof UnknownCommandException) {
             // display message and suggest alternatives in case of unknown command
             $message = $this->formatMessage($exception->getName() . ': ') . $exception->command;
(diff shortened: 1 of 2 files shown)


framework/console/ErrorHandler.php
```diff
@@ -55,7 +56,9 @@ protected function renderException($exception)
             if ($exception instanceof \yii\db\Exception && !empty($exception->errorInfo)) {
                 $message .= "\n" . $this->formatMessage("Error Info:\n", [Console::BOLD]) . print_r($exception->errorInfo, true);
             }
-            $message .= "\n" . $this->formatMessage("Stack trace:\n", [Console::BOLD]) . $exception->getTraceAsString();
+            if ($previous === null) {
+                $message .= "\n" . $this->formatMessage("Stack trace:\n", [Console::BOLD]) . $exception->getTraceAsString();
```
mikehaertl, line 60: @samdark There's a little difference to the web error handler. It renders the call stack for the top/outer most exception. Whereas here we render the stack trace for the inner most (originating) exception. Actually I find this more helpful. The stack trace usually includes the outer methods in its trace anyway.

samdark (approved):

samdark (approved):

samdark:
Thanks!

PR #18265, opened 2020-09-01
fix error how try get value by index from int

Ladone:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌ (I'm not know this thermin)
| Tests pass?   | ✔️
| Fixed issues  | https://github.com/yiisoft/yii2/issues/18269

When safeScenario get int value, in condition case try get 0 index from int
```
-  if ($attribute[0] !== '!' && !in_array('!' . $attribute, $scenarios[$scenario])) {
+  if (strncmp($attribute, '!', 1) !== 0 && !in_array('!' . $attribute, $scenarios[$scenario])) {```

### DIFF
--- a/tests/framework/base/ModelTest.php
+++ b/tests/framework/base/ModelTest.php
@@ -175,7 +175,7 @@ class ModelTest extends TestCase
         $this->assertTrue($speaker->isAttributeSafe('firstName'));
     }
 
-    public function testSafeAttributes(): void
+    public function testSafeAttributes()
     {
         $model = new RulesModel();
         $model->rules = [
--- a/tests/framework/base/ModelTest.php
+++ b/tests/framework/base/ModelTest.php
@@ -175,6 +175,18 @@ class ModelTest extends TestCase
         $this->assertTrue($speaker->isAttributeSafe('firstName'));
     }
 
+    public function testSafeAttributes(): void
+    {
+        $model = new RulesModel();
+        $model->rules = [
+            [
+                [123456], 'safe',
+            ]
+        ];
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Ladone** [comment]
@samdark how I can remove label `request for unit tests`?

**Comment by bizley** [comment]
We kindly request adding test(s) supporting the change, that is why the label is here.

But how come `$attribute` is not string there? Scenarios are suppose to be arrays of attribute names that are active in given scenario.

**Comment by Ladone** [comment]
@bizley yii2 have extension how have name Dynamic Form, how create from array object instance Form, and if user set int values they values fall in to safeAttributes when called method getScenario.

**Comment by bizley** [comment]
Ok, fair enough. This PR is just, especially that `strncmp` is already used in the next method below that. Please add unit test covering the case.

**Comment by Ladone** [comment]
@bizley, Unit Test has been added. Thanks for help. I'm waiting to next steps :)

File: tests/framework/base/ModelTest.php [hunk 1]
```diff
@@ -175,6 +175,18 @@ public function testIsAttributeSafe()
         $this->assertTrue($speaker->isAttributeSafe('firstName'));
     }
 
+    public function testSafeAttributes()
```
bizley, line 178:
```suggestion
    public function testIsAttributeSafeForIntegerAttribute()
```

bizley (approved):

samdark:
Would you please add a line for CHANGELOG.md?

Ladone:
@samdark CHANGELOG.MD has been updated

samdark:
Thanks!

